PR #1751, opened 2023-08-16
Add `-as-parameter` option

lukemaurer:
This allows compiling an .mli as a _parameter module_ rather than a normal compilation unit. A parameter module defines a module _type_ rather than a module, so it cannot be referred to directly from another module. A forthcoming PR will add the `-parameter P` option which adds the parameter module `P` as a parameter to the current module, which then allows references to `P` in the module. Further PRs will deal with how to actually use a module that takes parameters.

For the moment, `-parameter` is unimplemented, so any reference to a parameter module is an error.


ocaml/typing/env.ml
```diff
@@ -2652,28 +2652,26 @@ let persistent_structures_of_dir dir =
   |> String.Set.of_seq
 
 (* Save a signature to a file *)
-let save_signature_with_transform cmi_transform ~alerts sg modname filename =
+let save_signature_with_transform cmi_transform ~alerts sg modname cu filename =
```
lpw25, line 2655: `cu` seems an odd name for a `Cmi_format.kind` value

tools/flambda_backend_objinfo.ml
```diff
@@ -115,8 +115,15 @@ let print_cma_infos (lib : Cmo_format.library) =
...
-let print_cmi_infos name crcs =
+let print_cmi_infos name crcs kind =
+  let open Cmi_format in
   printf "Unit name: %a\n" Compilation_unit.output name;
+  let is_param =
+    match kind with
+    | Normal -> false
+    | Parameter -> true
+  in
+  printf "Is parameter: %s\n" (if is_param then "YES" else "no");
```
lpw25, line 126: (Same comment as the similar code in `tools/objinfo.ml`)

lpw25 (changes requested):
Looks good.

One thing that I think is missing is an error for if you try to compile `foo.ml` when `foo.mli` has been compiled as a parameter. Even if such an error does already exist it needs a test.

ocaml/tools/objinfo.ml
```diff
@@ -110,8 +110,15 @@ let print_cma_infos (lib : Cmo_format.library) =
...
-let print_cmi_infos name crcs =
+let print_cmi_infos name crcs kind =
+  let open Cmi_format in
   printf "Unit name: %a\n" Compilation_unit.output name;
+  let is_param =
+    match kind with
+    | Normal -> false
+    | Parameter -> true
+  in
+  printf "Is parameter: %s\n" (if is_param then "YES" else "no");
```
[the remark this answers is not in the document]
lukemaurer: It's in there, in `print_cmo_infos` and `print_cma_infos` for instance.

ocaml/typing/env.ml
lukemaurer: Renamed to `kind`

tools/flambda_backend_objinfo.ml
lukemaurer: See previous. (There should really be a `print_yes_or_no` function somewhere in the file for this.)

ocaml/typing/persistent_env.ml
```diff
@@ -392,6 +411,9 @@ let is_imported {imported_units; _} s =
 let is_imported_opaque {imported_opaque_units; _} s =
   CU.Name.Set.mem s !imported_opaque_units
 
+let is_parameter_unit penv s =
+  is_registered_parameter_import penv s
```
lpw25, line 415: Why are there two names for the same function? Are they different in later patches?

ocaml/typing/persistent_env.mli
```diff
@@ -73,6 +73,12 @@ val find_in_cache : 'a t -> Compilation_unit.Name.t -> 'a option
 val check : 'a t -> (Persistent_signature.t -> 'a)
   -> loc:Location.t -> Compilation_unit.Name.t -> unit
 
+(* Lets it be known that the given module is a parameter and thus is expected
+   to have been compiled as such. It may or may not be a parameter to _this_
+   module (see [register_parameter]). Raises an exception if the module has
```
lpw25, line 78: Note that `register_parameter` does not appear to exist. Is it added later? If it is then you can probably just leave this comment, but I wanted to make sure that it will exist.

ocaml/typing/typemod.ml
```diff
@@ -3634,6 +3649,17 @@ let report_error ~loc _env = function
       Location.errorf ~loc
         "@[Top-level module bindings must have layout value, but@ \
          %s has layout@ %a.@]" id Sort.format sort
+  | Cannot_implement_parameter path ->
+      Location.errorf ~loc
+        "@[Interface %s@ found for this unit is flagged as a parameter.@ \
+         It cannot be implemented directly. Use -as-argument-for instead.@]"
```
lpw25, line 3655: I don't think the suggestion is helpful here. I think making an argument for the parameter is unlikely to be what the user was trying to do.

ocaml/typing/persistent_env.ml
```diff
@@ -58,6 +58,7 @@ type can_load_cmis =
 
 type pers_struct = {
   ps_name: CU.t;
+  ps_is_param: bool;
```
lpw25, line 61: Why is this a bool rather than a `cmi_kind`?

ocaml/testsuite/tests/templates/basic/bad_param_impl.ml
```diff
@@ -0,0 +1,18 @@
...
+** ocamlc.byte
+flags = "-as-parameter"
+module = "bad_param_impl.mli"
+*** ocamlc.byte
+flags = ""
+module = "bad_param_impl.ml"
+ocamlc_byte_exit_status = "2"
+compiler_output = "bad_param_impl.output"
+**** skip
+reason = "error currently broken"
```
lpw25, line 15: Why is this error currently broken? It seems like it should work

lpw25 (approved):
Looks good. A few comments.

ocaml/typing/persistent_env.ml
lukemaurer: Looks like I lost track of what the function was supposed to do. The doc comment in the .mli is just wrong now. I should probably just delete this function and add `is_registered_parameter_import` to the .mli.

ocaml/typing/persistent_env.mli
lukemaurer: Actually, looks like I changed the name later on. Fixing the comment both here (to add "forthcoming") and in #1846 (to fix the name).

ocaml/typing/typemod.ml
lukemaurer: Changed the suggestion to "Recompile it without `-as-parameter`."

ocaml/typing/persistent_env.ml
lukemaurer: In later patches, `Cmi_format` goes into a correct-by-construction style: the `cmi_arg_for` field only makes sense for `Normal` modules, so for instance as of #1841, `cmi_arg_for` is a field of the `Normal` constructor of `cmi_kind`. For `Persistent_env`'s internal data structures, this seemed much less valuable, so `is_param` stays as a boolean.

(If we change #1841 and #1846 to keep `cmi_kind` a pure two-value enum, I'm happy to change `is_param`'s type to `cmi_kind` and rename it, probably to `cmi_kind` since the word "kind" is so overloaded (maybe I should change `cmi_kind` as well?).)

ocaml/testsuite/tests/templates/basic/bad_param_impl.ml
lukemaurer: Hmm ... I had thought this gets fixed later, but it looks like not. The issue is that the code in `typemod.ml` that's supposed to raise `Cannot_implement_parameter` can never actually raise: if it's declared as a parameter unit, the call to `Env.read_signature` fails. Basically, `persistent_env.ml` is currently too good at catching inconsistencies. It can't tell the difference between "this shouldn't be a parameter unit because it appeared in the source and it's not a `-parameter`" and "this shouldn't be a parameter because we're implementing it."

Basically, in order to report the error nicely to the user, either:

1. `Persistent_env` needs to do less error checking; or
2. `Persistent_env` needs to be given more context so that it can produce the correct error; or
3. pather than raising a fatal error itself, `Persistent_env` needs to return some error value that `Typemod` can interpret itself; or
4. in this particular case, I suppose `Persistent_env` could notice that the `.cmi` that shouldn't be a parameter happens to have the current module's name, and so it could assume that it should raise `Cannot_implement_parameter` instead of `Illegal_import_of_parameter`, but that feels a bit sketchy and I'm not sure it's always going to work out that nicely.

ocaml/typing/persistent_env.ml
lukemaurer: Done.

ocaml/testsuite/tests/templates/basic/bad_param_impl.ml
lpw25: I think you should take option 4. The code already looks up the current module name a few lines before the error is emitted.

ocaml/typing/typemod.ml
```diff
@@ -3686,10 +3692,19 @@ let report_error ~loc _env = function
...
+           @[In this strengthened module type, the type of %a@ \
+             does not match the underlying type@]@ \
+           %t@]"
+        longident lid main
   | Cannot_implement_parameter path ->
       Location.errorf ~loc
         "@[Interface %s@ found for this unit is flagged as a parameter.@ \
-         It cannot be implemented directly. Use -as-argument-for instead.@]"
+         It cannot be implemented directly. Recompile it without@ \
+         -as-parameter.@]"
```
lpw25, line 3707: I think it would be better not to give a suggestion here. It's hard to guess what the user was actually trying to do, and the error is already pretty clear.

lpw25 (approved):
This is ready to merge, but we should wait for the OCaml 5 branch to get merged in first.

lukemaurer:
The only test failures at this point are threading stuff, which seems like it's probably flaky?

lukemaurer:
> The only test failures at this point are threading stuff, which seems like it's probably flaky?

For the record, this was the breakage addressed by PR #2054. Rebasing fixed it. Merging

ocaml/typing/typemod.ml
```diff
@@ -3338,6 +3343,8 @@ let type_implementation sourcefile outputprefix modulename initial_env ast =
           signature = simple_sg
         } (* result is ignored by Compile.implementation *)
       end else begin
+        if !Clflags.as_parameter then
+          error Cannot_compile_implementation_as_parameter;
```
riaqn, line 3347: I think currently if you pass `-i -as-parameter`, the `-as_parameter` is silently ignored? We should print some error along the line of "-i and -as-parameter" cannot be used together.

riaqn: This is mentioned in https://github.com/ocaml-flambda/flambda-backend/pull/1846#issuecomment-2192194104 and will be addressed there.

PR #1752, opened 2023-08-16
Use raise-catch control flow to exit guarded right-hand sides

rajgodse:
We spent a lot of time doing complicated things to deal with passing pattern-guarded right-hand sides to the pattern match compiler. The main problem was that at the time that pattern-match compilation begins, the fallthrough code is not known. Our best prior try involved representing the right-hand sides as continuations (#1749).

However, as suggested by @lpw25, we can do something a lot simpler. Namely, we may translate a pattern-guarded rhs
```
... when e match (
    | P1 -> rhs1
    ...
    | Pn -> rhsn
)
```

as though it were
```
match e with
| P1 -> rhs1
...
| Pn -> rhsn
| _ -> Lstaticraise (i, [])
```
where `i` is some fresh exception label.

Then, when the code for jumping to the remaining cases (call it `patch`) becomes known, we may wrap the translated pattern-guarded rhs (call it `rhs`) in the following way:
```
Lstaticcatch (rhs, (i, []), patch, return_kind)
```

Naively, this seems to add multiple jumps when one would suffice. When the guarded rhs wishes to fall through, it would seem that it first raises `i`, which is caught and then raises some `i'`, caught by the handler for the next case.

However, these two jumps are optimized into one in `simplif.ml`, as confirmed by examining some dumped lambda. This greatly simplifies the logic around pattern guard translation while also not sacrificing the quality of the produced code.

### DIFF
--- a/ocaml/lambda/matching.ml
+++ b/ocaml/lambda/matching.ml
@@ -130,24 +130,22 @@ let may_compat = MayCompat.compat
 and may_compats = MayCompat.compats
 
 type rhs =
-  | Guarded of { action: lambda; exn_label: static_label }
-  (* Guarded rhs's must allow for fallthrough if the guard fails. This
-     fallthrough is facilitated by raising with label [exn_label].
-  *)
+  | Guarded of { rhs_with_fallthrough : lambda; fallthrough : static_label }
+  (* [rhs_with_fallthrough] raises exn [fallthrough] if the guard fails. *)
   | Unguarded of lambda
 
-let lambda_of_rhs = function
-  | Unguarded action | Guarded { action; _ } -> action
-
 let map_rhs ~f = function
-  | Unguarded action -> Unguarded (f action)
-  | Guarded { action; exn_label } -> Guarded { action = f action; exn_label }
+  | Unguarded rhs -> Unguarded (f rhs)
+  | Guarded { rhs_with_fallthrough; fallthrough } ->
+      Guarded { rhs_with_fallthrough = f rhs_with_fallthrough; fallthrough }
 
 let mk_unguarded_rhs action = Unguarded action
 let mk_guarded_rhs ~patch_guarded =
   let fresh_label = next_raise_count () in
-  let with_exit = patch_guarded ~patch:(Lstaticraise (fresh_label, [])) in
-  Guarded { action = with_exit; exn_label = fresh_label }
+  let rhs_with_fallthrough =
+    patch_guarded ~patch:(Lstaticraise (fresh_label, []))
+  in
+  Guarded { rhs_with_fallthrough; fallthrough = fresh_label }
 
 let is_guarded = function
   | Guarded _ -> true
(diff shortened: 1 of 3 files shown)


ocaml/lambda/matching.ml
```diff
@@ -129,65 +129,30 @@ let may_compat = MayCompat.compat
...
+  | Guarded of { action: lambda; exn_label: static_label }
+  (* Guarded rhs's must allow for fallthrough if the guard fails. This
+     fallthrough is facilitated by raising with label [exn_label].
   *)
   | Unguarded of lambda
 
-let mk_boolean_guarded_rhs ~patch_guarded ~free_variables =
-  Guarded { patch_guarded; free_variables = Precomputed free_variables }
+let lambda_of_rhs = function
+  | Unguarded action | Guarded { action; _ } -> action
```
ncik-roberts, line 140: I'd actually prefer to get rid of this function entirely and push any pattern-matching out to clients. This client is the one who has enough context to decide whether it's appropriate to treat the `action` as the "whole lambda term".

E.g. what if a client wanted to collect "unhandled static raise labels"? They'd get the wrong answer if they used `lambda_of_rhs`.

It's obvious that collecting free variables is ok to do, but that is clearest to see by inlining the match at the point we collect the free variables.

ocaml/lambda/matching.ml
```diff
@@ -129,65 +129,30 @@ let may_compat = MayCompat.compat
...
-     When translating a guarded rhs, the code to execute on fallthrough must
-     be "patched" in. Because the fallthrough code is translated after the
-     rhs is created, guarded actions rhs's a function [patch_guarded], which
-     generates a lambda term for the action from the fallthrough code.
-
-     One translation optimization requires us to check the free variables of an
-     rhs. When it is simple to do so, we precompute these free variables for use
-     in the optimization. Otherwise, we leave them uncomputed and do not apply
-     the optimization.
+  | Guarded of { action: lambda; exn_label: static_label }
```
ncik-roberts, line 133: Can we name `action` something else? I'd like to avoid introducing "action" as a term. Maybe the name could suggest a stronger relationship with `arg_label`, something like:

```ocaml
{ rhs_with_fallthrough : lambda; fallthrough : static_label }
```

The comment could simply say "`rhs_with_fallthrough` raises exn `fallthrough` if the guard fails."

ocaml/lambda/matching.ml
```diff
@@ -1162,14 +1127,9 @@ let same_actions = function
...
-    match act1, act2 with
-    | Unguarded act1, Unguarded act2 ->
-        (match make_key act1, make_key act2 with
-         | Some key1, Some key2 -> key1 = key2
-         | None, _ | _, None -> false)
-    (* CR-soon rgodse: Investigate the callsites of this function to determine
-       if Guarded rhs's should ever be deemed the same. *)
-    | Guarded _, _ | _, Guarded _ -> false
+    match make_key (lambda_of_rhs act1), make_key (lambda_of_rhs act2) with
+    | Some key1, Some key2 -> key1 = key2
```
ncik-roberts, line 1131: I would think it's safest/clearest to classify the actions as the same only if they're both guarded RHSes with the same key *and* the same exn label.

ocaml/lambda/matching.ml
```diff
@@ -3424,14 +3352,19 @@ let rec compile_match ~scopes value_kind repr partial ctx
...
+       | Guarded { action; exn_label } ->
+           let lambda, total =
+             compile_match ~scopes value_kind None partial ctx
+               { m with cases = rem }
+           in
+           let rec catch_fallthrough = function
+             | Llet (str, k, id, lam, body) ->
+                 Llet (str, k, id, lam, catch_fallthrough body)
+             | Levent (lam, ev) -> Levent (catch_fallthrough lam, ev)
+             | lam ->  Lstaticcatch (lam, (exn_label, []), lambda, value_kind)
```
ncik-roberts, line 3365: Why is this necessary? Naively, I'd have guessed that inserting `Lstaticcatch` around the whole expression is always ok.

ncik-roberts (commented):
I agree that this is a wonderful simplification. I left a few minor comments.

ocaml/lambda/matching.ml
rajgodse: The result of this operation is passed to `event_branch`, which gets mad when `repr` is not `None` and `lam` is not some series of `Llet`s pointing to an `Levent` (as in this case, `repr` is meant to be stitched into the `Levent`).

I think the better way to handle this is to change `event_branch` to traverse the bodies of `Lstaticcatch`s when looking for the appropriate `Levent`, which is what I've done.

ocaml/lambda/matching.ml
rajgodse: Done!

ncik-roberts (approved):
Looks good to me, thanks!

PR #1757, opened 2023-08-18
Add simd-regalloc flag to x86 backend

TheNumbat:
This flag allows us to enable the SIMD-related register allocation changes without the full SIMD language extension.
When the extension is enabled, the register allocation features are automatically enabled as well.

### DIFF
--- a/backend/amd64/proc.ml
+++ b/backend/amd64/proc.ml
@@ -23,7 +23,7 @@ open Cmm
 open Reg
 open Mach
 
-let simd_regalloc_disabled () = not (Language_extension.is_enabled SIMD)
+let simd_regalloc_disabled () = not (Language_extension.is_enabled SIMD || !simd_regalloc)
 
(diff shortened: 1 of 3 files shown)


backend/amd64/proc.ml
```diff
@@ -23,7 +23,7 @@ open Cmm
 open Reg
 open Mach
 
-let simd_regalloc_disabled () = not (Language_extension.is_enabled SIMD)
+let simd_regalloc_disabled () = not (Language_extension.is_enabled SIMD || !simd_regalloc)
```
xclerc, line 26: I am not sure it is the right condition;
if the langage extension is on, disabled
will be false independently of the arch-
specific parameter.

TheNumbat: That is the intent - I think users should only need to enable the extension

xclerc (approved):

PR #1759, opened 2023-08-18
Port upstream #12368 about abstract environments

goldfirere:
My real goal is in accepting

```
type 'a t = 'a foo foo
and 'a foo = 'a;;
```

which is rejected today. But this at least improves the error message. I will continue to tackle the real problem when I return from next week's vacation.

The actual content here is just a pretty straightforward porting of https://github.com/ocaml/ocaml/pull/12368. I would say only minimal review is required, given that this was reviewed upstream.


lpw25 (approved):

mshinwell:
@goldfirere ready for rebasing and merging, by the look of it?

PR #1761, opened 2023-08-21
Fix modes annotation ghost location

riaqn:
This PR fixes the handling of mode annotation locations info, especially when they are on the LHS of `let` bindings. For example:
```
let local_ unique_ x = body in
```
This will be parsed as
```
let x = [%extension.local] [%extension.unique] body in
```
where the two extensions themselves will have non-ghost locations set, but both 
```
[%extension.unique] body
```
and 
```
 [%extension.local] [%extension.unique] body
```
will have ghost locations copied from `body`.


ocaml/parsing/parser.mly
```diff
@@ -221,40 +232,41 @@ let wrap_exp_once exp loc =
...
+    The extension node will always have non-ghost (but filled-in) location.
+ *)
+let mkexp_with_mode ?loc (mode, kwd_loc) exp =
+  match mode with
+  | Local -> mkexp_stack exp ?loc ~kwd_loc
+  | Unique -> mkexp_unique exp ?loc ~kwd_loc
+  | Once -> mkexp_once exp ?loc ~kwd_loc
 
-let mkexp_with_modes flags exp loc =
+let mkexp_with_modes modes exp =
```
ncik-roberts, line 245: I wonder if we should name this `ghexp_with_modes` to more strongly suggest that the location minted is a ghost location. This matches the naming scheme used for a few other bindings in the file.

ocaml/parsing/parser.mly
```diff
@@ -162,25 +162,36 @@ let once_attr loc =
...
+  in
   Exp.mk ~loc (Pexp_apply(unique_extension (make_loc kwd_loc), [Nolabel, exp]))
 
-let mkexp_once ~loc ~kwd_loc exp =
-  let loc = make_loc loc in
+let mkexp_once ?loc ~kwd_loc exp =
+  let loc =
+    match loc with
+    | None -> Location.ghostify exp.pexp_loc
+    | Some loc -> make_loc loc
```
ncik-roberts, line 193: Perhaps we can move this repeated block of code from `mkexp_once/mkexp_unique/mkexp_stack` into `mkexp_with_mode`. Then, we can make the location argument to `mkexp_once/mkexp_unique/mkexp_stack` required again (which I think only affects one additional callsite &mdash; a call to `mkexp_stack` that supplies the location anyway).

ocaml/testsuite/tests/typing-unique/unique.ml
```diff
@@ -212,9 +212,9 @@ val x : string = "foo"
 (* Top-level must be many *)
 let once_ foo = "foo"
 [%%expect{|
-Line 1, characters 4-21:
+Line 1, characters 16-21:
 1 | let once_ foo = "foo"
-        ^^^^^^^^^^^^^^^^^
+                    ^^^^^
```
ncik-roberts, line 217: This error message was a bit clearer when the ghost location spanned from the start of `once_` to the end of `"foo"`. Can we make `ghexp_with_modes` continue to take a location argument? Then, `ghexp_with_modes` can call `ghostify` on that location before passing it to `mkexp_with_mode` &mdash; if we do this, I'm pretty sure that we can avoid having the `?loc` argument be optional in `mkexp_with_mode` and the functions it calls.

ocaml/parsing/parser.mly
ncik-roberts: Please see my other comments before addressing this comment. It may no longer be needed.

ncik-roberts (commented):
Thanks for looking at this &mdash; I just have a few minor comments.

ocaml/parsing/parser.mly
riaqn: Yes, I believe this is resolved because no longer needed.

ocaml/testsuite/tests/typing-unique/unique.ml
riaqn: The error message is restored now. Also I found that the nameing convention is either `mkexp*` where `make_loc` is called, or `ghexp*` where `ghost_loc` is called - there is no binding where the callsite passes a location that already knows ghost-ness. So to make everything more consistent, I also changed `mkexp_with_mode` to `ghexp_with_mode` that always calls `ghost_loc` - I don't think it makes much difference, and probably less confusing anyway.

riaqn:
@ncik-roberts thanks for the commit - I think it looks good and ready to go (once CI is green)

ncik-roberts (approved):
Zesen and I chatted and I convinced him (mostly) that we should keep a non-ghost version of `mkexp_with_mode`. This is a slight inconsistency in naming &mdash; usually, the `mkexp_*` function is the one that calls `make_loc`, not the caller of said function, but this inconsistency should be ironed out with Antal's PR.

TL;DR: LGTM.

PR #1714, opened 2023-08-09
zero alloc: assume that works with inlining

gretay-js:
This PR propagates `[@zero_alloc assume]` annotation from a function to its body, piggybacking on `Debuginfo.t`. The PR starts with a bit of refactoring to make it easier to add a field to `Debuginfo.t`. I'll be happy to separate the refactoring commit into another PR. 

The propagation happens in `middle_end` during translation from `Lambda` and boils down to adding a new argument to `Debuginfo.from_location` and to the "environment" of translation. 
This PR a draft because it looks too complicated. The change is separate for each middle end, and I am not sure if I have all the places updated correctly. Is there a better way? Would it be better to add another pass on Lambda dedicated to propagating this information? 

Tested on small examples manually with all 3 middle ends. TODO: test on larger examples.

PR #1762, opened 2023-08-21
Zero alloc: assume that works with inlining - propagate via Scoped_location

gretay-js:
A new version of #1714. Instead of changing the middle-end translation from `Lambda`, this PR changes `Translcore` to parse `[@zero_alloc assume]` function attributes a little earler than before and propagate it as part of the `scopes` to the body of the function. Other changes are:

- Add field `assume_zero_alloc` to `Debuginfo.Scoped_location.scopes.Cons` 
- Propagate `assume_zero_alloc` from `Scoped_location.t` to Debuginfo.t`
- Use `assume_zero_alloc` in `Checkmach`
- Remove warning 198 and related tests
- Add tests

Best reviewed commit by commit.
Still needs testing on large examples.

### DIFF
--- a/ocaml/lambda/debuginfo.ml
+++ b/ocaml/lambda/debuginfo.ml
@@ -166,7 +166,67 @@ type item = {
   dinfo_scopes: Scoped_location.scopes;
 }
 
-type t = { dbg : item list; assume_zero_alloc : bool }
+module Dbg = struct
+ type t = item list
+
+  (* CR-someday afrisch: FWIW, the current compare function does not seem very
+     good, since it reverses the two lists. I don't know how long the lists are,
+     nor if the specific currently implemented ordering is useful in other
+     contexts, but if one wants to use Map, a more efficient comparison should
+     be considered. *)
+  let compare dbg1 dbg2 =
+    let rec loop ds1 ds2 =
+      match ds1, ds2 with
+      | [], [] -> 0
+      | _ :: _, [] -> 1
+      | [], _ :: _ -> -1
+      | d1 :: ds1, d2 :: ds2 ->
+        let c = String.compare d1.dinfo_file d2.dinfo_file in
+        if c <> 0 then c else
+          let c = compare d1.dinfo_line d2.dinfo_line in
+          if c <> 0 then c else
+            let c = compare d1.dinfo_char_end d2.dinfo_char_end in
+            if c <> 0 then c else
+              let c = compare d1.dinfo_char_start d2.dinfo_char_start in
+              if c <> 0 then c else
+                let c = compare d1.dinfo_start_bol d2.dinfo_start_bol in
+                if c <> 0 then c else
+                  let c = compare d1.dinfo_end_bol d2.dinfo_end_bol in
+                  if c <> 0 then c else
+                    let c = compare d1.dinfo_end_line d2.dinfo_end_line in
+                    if c <> 0 then c else
+                      loop ds1 ds2
+    in
+    loop (List.rev dbg1) (List.rev dbg2)
+
+  let is_none dbg =
+    match dbg with
+    | [] -> true
+    | _ :: _ -> false
+
+  let hash dbg =
+    List.fold_left (fun hash item -> Hashtbl.hash (hash, item)) 0 dbg
+
+  let to_string dbg =
+    match dbg with
+    | [] -> ""
+    | ds ->
+      let items =
+        List.map
+          (fun d ->
+             Printf.sprintf "%s:%d,%d-%d"
+               d.dinfo_file d.dinfo_line d.dinfo_char_start d.dinfo_char_end)
+          ds
+      in
+      "{" ^ String.concat ";" items ^ "}"
+
+  let to_list t = t
+
+  let length t = List.length t
+
+end
+
+type t = { dbg : Dbg.t; assume_zero_alloc : bool }
 
 type alloc_dbginfo_item =
   { alloc_words : int;
@@ -175,27 +235,8 @@ type alloc_dbginfo = alloc_dbginfo_item list
 
 let none = { dbg = []; assume_zero_alloc = false }
 
-let is_none { dbg; assume_zero_alloc } =
-  if assume_zero_alloc then false else
-  match dbg with
-  | [] -> true
-  | _ :: _ -> false
-
-let to_string dbg =
-  match dbg with
-  | [] -> ""
-  | ds ->
-    let items =
-      List.map
-        (fun d ->
-           Printf.sprintf "%s:%d,%d-%d"
-             d.dinfo_file d.dinfo_line d.dinfo_char_start d.dinfo_char_end)
-        ds
-    in
-    "{" ^ String.concat ";" items ^ "}"
-
 let to_string { dbg; assume_zero_alloc; } =
-  let s = to_string dbg in
+  let s = Dbg.to_string dbg in
   if assume_zero_alloc then s^"(assume zero_alloc)" else s
 
 let item_from_location ~scopes loc =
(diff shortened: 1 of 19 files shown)


ocaml/lambda/debuginfo.ml
```diff
@@ -143,21 +166,22 @@ type item = {
...
   { alloc_words : int;
     alloc_dbg : t }
 type alloc_dbginfo = alloc_dbginfo_item list
 
-let none = { dbg = []; }
+let none = { dbg = []; assume_zero_alloc = false }
 
-let is_none { dbg } =
+let is_none { dbg; assume_zero_alloc } =
+  if assume_zero_alloc then false else
```
xclerc, line 179: Can you elaborate on why this change is needed?

ocaml/lambda/debuginfo.ml
```diff
@@ -247,8 +277,14 @@ let compare { dbg = dbg1; } { dbg = dbg2; } =
   in
   loop (List.rev dbg1) (List.rev dbg2)
 
-let hash { dbg; } =
-  List.fold_left (fun hash item -> Hashtbl.hash (hash, item)) 0 dbg
+let compare { dbg = dbg1; assume_zero_alloc = a1; }
+      { dbg = dbg2; assume_zero_alloc = a2; } =
```
xclerc, line 281: I think this may cause a loss of sharing
in `Emitaux`. If it is to be fixed, the
change should probably be in `Emitaux`.

Ekdohibs:
What is the expected semantics of `[@zero_alloc assume]`, precisely?

I'm slightly worried that this might cause some allocations to be seen as "inside a zero-alloc assume block" instead of outside because of things like CSE: consider for instance this code.
```ocaml
let[@zero_alloc assume][@inline always] f x = (x, x)

let[@zero_alloc check] g y =
  let (a, b) = f y in
  (y, y)
```
Here, `f` is a function marked `[@zero_alloc assume]` because although it does an allocation, we assume this allocation is only ever used to return two values, and it will always be used in contexts like `let (a, b) = f x` (which is the case here).
However, when `g` is simplified, flambda sees that `(y, y)` and the allocation in `f` are the same, immutable block; and thus reuses the allocation inside of `f`, which would have otherwise disappeared. However, this means that the zero-alloc check will then see an allocation inside of `g`, but thinks that this allocation was in fact inside of `f`, thus passing the check!

As I don't know the expected semantics of `[@zero_alloc assume]` with inlining, this might be benign or even expected, but I think it's important to think about the fact that optimisations have the potential to influence where the backend thinks an allocation comes from.

gretay-js:
@Ekdohibs 
> What is the expected semantics of `[@zero_alloc assume]`, precisely?

I don't have a formal definition. The intended behavior of the check is to report all allocations except the ones that **must** originate from a function annotated with `[@zero_alloc assume]`. Conservatively approximating this property means that if we cannot prove that an allocation originates only from "assume", we should report it. I hope that in most cases we can compute a precise enough approximation, but not sure how to phrase this. 

In your example (thank you for it!), I expect the check to fail, because execution of `g` allocates a block that represents`f`. Unfortunately, the check passes with flambda2 currently and this is "unsound" according to the above intuition.

Optimizations like CSE that may reuse a block should be careful about how they propagate the "assume_zero_alloc" field. I don't know how to enforce this in practice. In your example, we should take the conjunction of the values of `assume_zero_alloc` fields of the two debuginfo. Can you direct me to the right code in flambda2 please to fix it? 

`Comballoc` should do the same, I'll try to come up with a test and fix the code for it, it is currently not looking at the `dbginfo` field of the combined `Ialloc` instructions.

ocaml/lambda/debuginfo.ml
gretay-js: You are right, in `Emitaux` we should ignore the `assume` field.

Ekdohibs:
Concerning flambda2, I'm not even sure where this form of CSE comes from: there's `commmon_subexpression_elimination.ml` which does some of the CSE, but we're in a situation where the types are enough to do the CSE, so it probably comes from there. Maybe @lthls would know more about it? I expect the fix to be something like adding a "count-for-zero-alloc" property on immutable blocks.

On another point, I'm not sure relying on `Debuginfo.t` is really the best, as it might be forgotten in some points (although this would only cause false positives), couldn't we just set this information on allocating primitives instead?

However, if we say that it must come from the zero-alloc, I this this means that the following will fail zero-alloc check:
```ocaml
let[@inline always][@zero_alloc assume] f x = (x, x)

let[@zero_alloc check] g x =
  let (a, b) = (x, x) in
  f x
```
Indeed, in this case, CSE will happen as well, and this the allocation could have come from inside the `zero_alloc check` region, it will be reported, even if it corresponds to the allocation inside of `f`. This is a false positive, so it is less of a problem, but we might want to think how we can improve this.

lpw25:
> Here, f is a function marked [@zero_alloc assume] because although it does an allocation, we assume this allocation is only ever used to return two values, and it will always be used in contexts like let (a, b) = f x (which is the case here).

That's not really the intended use case of `assume`, and it's not clear to me that giving no error here is the wrong semantics. `assume` is for marking functions that do not allocate (or at least do not allocate down non-error paths), but where the analysis cannot tell that they do not allocate. Here the user essentially said that they don't care about that pair allocation. Since that is the only allocation that happens then the function is zero alloc.

Ekdohibs:
While this is true, I believe the point still stands. I gave this example because it was easy to do; but if `f` only contained something like `let (a, b) = (x, x) in ...` but this was not simplified before inlining it (this is probably hard to do while keeping `f` zero-alloc, but on the other hand, both the check and flambda using complicated analyses, we can expect there will be differences in precision in a lot of places), then the same problem might arise.

On the other hand, I now see that the CSE can have dangerous behaviours, such as moving allocations before branches in some cases. For instance, consider the following example.
```ocaml
let[@inline always] f x = (x, x)

let g u y =
  let w = (y, y) in
  let (a, b) = w in
  if u then f y else (a, 0)
```
Here, the allocation for `w` can be straightforwardly eliminated, and all allocations happen under the conditional. However, with CSE, the allocation in the `f y` path sees it is identical to `w` and keeps it alive instead. This causes the allocation to happen even in the other branch!

With the zero-alloc check, this means that an allocation on the exception-only path (even in another function, for instance in `f`) can keep alive an allocation that should be eliminated; and the zero-alloc check needs to be able to detect this; this might also be something we want to think about concerning the implementation of the CSE.

gretay-js:
> ```ocaml
> let[@zero_alloc assume][@inline always] f x = (x, x)
> 
> let[@zero_alloc] g u y =
>   let w = (y, y) in
>   let (a, b) = w in
>   if u then f y else (a, 0)
> ```

For this example, the check fails pointing to  `let w = (y, y) in ..` as the source of allocation, but actually the code does not allocate anything before the conditional:

```
camlTest7__g_2_5_code(R/0[%rax] R/1[%rbx]) {test7.ml:5,19-96} assert zero_alloc
  u/45 := R/0[%rax]
  y/46 := R/1[%rbx]
  if u/45 ==s 1 then
    V/48 := alloc 24{test7.ml:8,21-31}
    [V/48 + -8] := 2048 (init)
    val[V/48] := y/46 (init)
    I/49 := y/46
    I/49 := I/49 - y/46{test7.ml:8,25-30}
    I/50 := I/49
    I/50 := I/50 + 1{test7.ml:8,25-30}
    val[V/48 + 8] := I/50 (init)
    R/0[%rax] := V/48
    return R/0[%rax]
  else
    V/47 := alloc 24{test7.ml:6,10-16}
    [V/47 + -8] := 2048 (init)
    val[V/47] := y/46 (init)
    val[V/47 + 8] := y/46 (init)
    R/0[%rax] := V/47
    return R/0[%rax]
  endif
```

The allocation that comes from the true branch of the source condition does not have "assume" tag in Cmm. I guess CSE in flambda2 "lost" it and the check is conservative in this case (unlike the previous example, where CSE in flambda2 happened to keep "assume", leading to an unsound result). In this example the error report happens to be useful, but in general `let w = (y,y)` may be due to inlining and won't be easy for the user to fix this false alarm.

Ekdohibs:
Oh, it seems the allocation was moved after the conditional by `to_cmm` (it is before the conditional in the simplify output). In any case, there are complexities we need to think about here, as I wouldn't be surprised if to_cm couldn't sink the allocation past the conditional in some cases.

gretay-js:
> With the zero-alloc check, this means that an allocation on the exception-only path (even in another function, for instance in f) can keep alive an allocation that should be eliminated; and the zero-alloc check needs to be able to detect this; this might also be something we want to think about concerning the implementation of the CSE.

I think Mark's PR#1767 addresses it (please correct me if not) and with PR#1767, zero_alloc check will be conservative even in the presence of CSE in the middle-end and won't affect generated code. I tested on small examples above and they gave error messages as expected (sounds, sometimes overly conservative due to CSE).  I'll add these tests.

ocaml/lambda/debuginfo.ml
gretay-js: I interpreted `is_none t` to return true if and only if  `t` is equal to `Debuginfo.none`, i.e., taking into account all fields of the `Debuginfo.t` record.  Did you expect `is_none`  to work only on the list of items? 
I updated the interface `debuginfo.mli` to make it clear when functions operate on the list of items (i.e.,g `dbg` field of `Debuginfo.t`) rather than the entire record. It is used only during Emit. I think it addresses your comment below as well about sharing `Debuginfo` in `Emitaux` below. Please let me know what you think about this version.

xclerc (approved):

ccasin:
Ah - I guess the bit about changing the interface doesn't make sense because the scopes need to be updated before translating the function.  Sorry for the noise.

mshinwell (approved):

PR #1499, opened 2023-06-14
128-bit SIMD vector primitive type

TheNumbat:
Adds a new primitive type (`vec128`) representing a boxed 128-bit SIMD vector.
Currently, values of this type are opaque, so can only be passed among functions/stored in boxed data structures.
They may also be passed unboxed to external C calls that expect `__m128` parameters/returns.

Features to add in future PRs:
- primitive type per specific packed type + typed intrinsics for SSE operations
- otherlibs module providing the user facing API
- bigstring / bigarray / load from float array support
- vec128array of unboxed vec128s (for float32 / float64 only)
- vec128 layout + unboxed 128 bit types

Later, the same for 256-bit and 512-bit operations.

Current omissions:
- No arm64 support
- No upstream compiler backend support
- Does not assure vectors stored in memory are 16-byte aligned, except stack spills/stack arguments for extcalls

PR #1763, opened 2023-08-23
Refactor stack classes

xclerc:
This pull request refactors the handling of
stack classes. Since the introduction of
vector types (#1499), we distinguish
between register and stack classes.

Having both encoded as mere ints is
error-prone, so this pull request introduces
a proper and abstract type for stack
classes (a similar refactoring will be
performed on register classes, but it is
slightly more complex).

This pull request introduces the following
modules:

- `Stack_class_utils` which defines what
  stack classes and tables with stack classes
  as keys are;
- `Stack_class` which defines backend-dependent
  stack classes.

As of 23c812461b7d1293b267ffe18b452ebe74cd6cca, tables are implemented
as hash tables instead of arrays. If this causes
a performance or marshalling issue, it is
easy to use a bare array as the underlying
representation of the table.

### DIFF
--- a/backend/.ocamlformat-enable
+++ b/backend/.ocamlformat-enable
@@ -6,6 +6,9 @@ checks.ml
 checks.mli
 checkmach.ml
 checkmach.mli
+stack_class.mli
+stack_class_utils.ml
+stack_class_utils.mli
 cfg/**/*.ml
 cfg/**/*.mli
 asm_targets/**/*.ml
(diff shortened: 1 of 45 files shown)


gretay-js (approved):

xclerc:
Superseded by https://github.com/ocaml-flambda/flambda-backend/pull/3784.

PR #1765, opened 2023-08-23
Disable split preprocessing in flambda2 classic mode

xclerc:
(As per title.)


mshinwell:
Hmm, I'm wondering if we should try to avoid the backend depending on flambda2.  Should we set this (in the flambda2 code) as an option on the `fundecl` records or something?

mshinwell (approved):
Xavier and I discussed this, in fact this should be ok

PR #1767, opened 2023-08-24
Merging of Debuginfo.t across CSEd occurrences

mshinwell:
This should allow fixing of the problem discussed in https://github.com/ocaml-flambda/flambda-backend/pull/1762.  It allows the `Debuginfo.t` value, occurring on the right-hand side of a let whose bound variable was used to substitute out CSEd expressions, to contain data merged from all of the CSEd occurrences (and the original one).


lthls:
It all looks fine, except for the part in `debuginfo.ml`: we're not actually merging anything, so I don't think it solves the actual problem.
@mshinwell Did you intend this to be a proper fix or were you planning to do something more clever ?

lthls:
Ah, I see. The Flambda 2 parts seem correct, so I think we can merge this.

lthls (approved):

Gbury (approved):

PR #1509, opened 2023-06-16
Factor out kernel of `Language_extension` used by Jane Syntax

ncik-roberts:
Reviewer: @antalsz. (Talk to me when you start reviewing this &mdash; I can show you how this will be imported into `ppxlib_jane`)

---

**Problem:** We want to import `Jane_syntax` into the new `ppxlib_jane` library we're putting into public release. This implies transitively importing all dependencies of `Jane_syntax` (that aren't already in upstream compilerlibs) into public release.

**Extra problem:** We want the implementation of some of these imports to differ between the compiler and `ppxlib_jane`. In particular, we'd like `Language_extension.is_enabled` and friends to always return true in `ppxlib_jane`. (We have a separate mechanism in ppxlib that will guard against Jane Street syntax elements making it into ppxes that don't intend for this &mdash; that's what erasability vs. non-erasability is for.) If `Language_extension.is_enabled` returned false in `ppxlib_jane`, it would fail to interpret Jane Street syntax.

**Solution:** Import `Jane_syntax` into the new `ppxlib_jane` library, but with a `Language_extension`-shaped hole that we fill in with a custom implementation where `is_enabled` always returns true.

**How this PR enables this solution:** We'd really like to import as much as possible directly from this repo into the `ppxlib_jane` library. So, this PR factors out a `Language_extension_kernel` module that can be imported directly.

The "`Language_extension`-shaped hole" in `Jane_syntax` is a module `Language_extension : Language_extension_kernel_for_jane_syntax`.

   * In the compiler, this hole is filled by the module `utils/language_extension.ml`
   * In `ppxlib_jane`, this hole is filled by an implementation in `ppxlib_jane/src/language_extension.ml`

Both of these `Language_extension` modules include `Language_extension_kernel`, so the amount of code we need to add in `ppxlib_jane` is very small.

PR #1768, opened 2023-08-24
Regulate access to [Language_extension] from within Jane Syntax

ncik-roberts:
Along the same lines as #1509: avoid accessing non-exported constructors from `Language_extension_kernel`. The need for this PR was flagged during the import process into `ppxlib_jane`.

**Reviewer:** @antalsz

### DIFF
--- a/ocaml/parsing/jane_syntax.ml
+++ b/ocaml/parsing/jane_syntax.ml
@@ -2,6 +2,19 @@ open Asttypes
 open Parsetree
 open Jane_syntax_parsing
 
+(** We carefully regulate which bindings we import from [Language_extension]
+    to ensure that we can import this file into the Jane Street internal
+    repo with no changes.
+*)
+module Language_extension = struct
+  include Language_extension_kernel
+  include (
+    Language_extension
+    : Language_extension_kernel.Language_extension_for_jane_syntax)
+end
+
+module _ = Language_extension
+
 (****************************************)
 (* Helpers used just within this module *)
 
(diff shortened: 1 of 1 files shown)


ocaml/parsing/jane_syntax.ml
```diff
@@ -2,6 +2,19 @@ open Asttypes
...
+    repo with no changes.
+*)
+module Language_extension = struct
+  include Language_extension_kernel
+  include (
+    Language_extension
+    : Language_extension_kernel.Language_extension_for_jane_syntax)
+end
+
+module _ = Language_extension
```
antalsz, line 16: Why do we need this?

antalsz (commented):
This mostly looks good, but I'm confused about one piece of it

ocaml/parsing/jane_syntax.ml
ncik-roberts: This suppresses an "unused module" warning if we don't actually use `Language_extension` (which we currently do, but I don't want that warning to pop up if the use site gets deleted in an unrelated change).

antalsz: Ah, I was confused because we currently do.  Can you add that in a comment, then?  Besides that, LGTM

ncik-roberts:
```suggestion
(* Suppress the unused module warning so it's easy to keep around the
   shadowing even if we delete use sites of the module. *)
module _ = Language_extension
```

antalsz (approved):
LGTM, let's merge when CI completes.

PR #1777, opened 2023-08-29
Create a new `Jane_asttypes` module for our new types

antalsz:
It causes some importing trouble when `Jane_syntax` depends on parts of `Asttypes` that aren't in the upstream compiler; in general, just as we don't want to modify the parse tree, we don't want to modify `Asttypes`, since it's effectively part of the parsetree.

Moving some types into `Jane_syntax` causes thorny dependency knots, though, so this PR creates a parallel `Jane_asttypes` module, which contains our extensions but lives at the same point higher up the dependency tree that `Asttypes` does.

Reviewer: Anybody, really; to avoid paralysis, I'll nominate @ncik-roberts or @goldfirere, but anybody else should feel free to take this intead.


ncik-roberts (approved):
Looks good to me. Probably worth double-checking that we don't need to move anything else over, but otherwise I'd say ship it.

antalsz:
It's a shame that I don't see a way to automate these checks

antalsz:
For instance, here's another missed move; we can't reference new functions like `Location.ghostify` in `jane_syntax.ml` and friends.

antalsz:
Alright, this finally shakes all the dependencies out, and provides some motivation.  Mergeable?

ocaml/parsing/jane_asttypes.mli
```diff
@@ -0,0 +1,39 @@
...
+(*   Copyright 2023 Jane Street Group LLC                                 *)
+(*                                                                        *)
+(*   All rights reserved.  This file is distributed under the terms of    *)
+(*   the GNU Lesser General Public License version 2.1, with the          *)
+(*   special exception on linking described in the file LICENSE.          *)
+(*                                                                        *)
+(**************************************************************************)
+
+(** Auxiliary Jane Street extensions to AST types used by parsetree and
+    typedtree.
```
ncik-roberts, line 16: Should we refer to the "Note" in `jane_syntax.mli`?

ocaml/parsing/jane_syntax.ml
```diff
@@ -1291,15 +1309,17 @@ module Layouts = struct
...
+       want this file to be compatible with the upstream compiler; ; see Note
+       [Buildable with upstream] in jane_syntax.mli for details. *)
     let report_error ~loc = function
-      | Unexpected_wrapped_type typ ->
+      | Unexpected_wrapped_type _typ ->
         Location.errorf ~loc
-          "Layout attribute on wrong core type:@;%a"
-          (Printast.core_type 0) typ
-      | Unexpected_wrapped_ext ext ->
+          "Layout attribute on wrong core type (unprintable)"
```
ncik-roberts, line 1319: Can we remove these "(unprintable)"s? I don't think they clarify.

ncik-roberts (approved):
I looked at the more recent commits and confirm that they look good.

antalsz:
Looks like we need an approving review from @mshinwell or @xclerc, because this touches the `dune` files

mshinwell (approved):
Only dune files

PR #1758, opened 2023-08-18
Fix `make debug`, yet again

goldfirere:
Going back to the dune setup moved debug_printers. This commit chases it down.

PR #1778, opened 2023-08-29
Fix ocamldebug source path for debug printers

ncik-roberts:
Change the `.ocamldebug` file generated by `make debug` to end in:
```
source _build/main/ocaml/tools/debug_printers
```

instead of:

```
source ocaml/tools/debug_printers
```

This is needed because the build step for `ocaml/tools/debug_printers` invokes dune, and dune puts its output in `_build/main`.

If I don't make this change, then running `ocamldebug-ocaml` from emacs generates confusing output and then crashes:

```
Executing file .ocamldebug
Directories:  _build/main/ocaml/.ocamlcommon.objs/byte .
  /home/nroberts/local/ocaml/fresh-flambda-backend-clone/nroberts-out/lib/ocaml
  _build/main/backend/asm_targets/.asm_targets.objs/byte
...hundreds of lines of similar output...
Source file not found.

OCaml-Debugger exited abnormally with code 2
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -183,4 +183,4 @@ ocamlopt:
 
 .ocamldebug: install
 	find _build/main -name '*.cmo' -type f -printf 'directory %h\n' | sort -u > .ocamldebug
-	echo "source ocaml/tools/debug_printers" >> .ocamldebug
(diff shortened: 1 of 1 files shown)


goldfirere (approved):

goldfirere:
Dup of #1758, but slightly more general.

PR #1779, opened 2023-08-29
Loosen ownership of build files

antalsz:
Having all changes to the `dune` (etc.) files go through @mshinwell and @xclerc is very restrictive, and doesn't seem necessary.  This PR drops that requirement, and leaves them owned by whoever owns the directory they reside in.

This PR was inspired by this requirement delaying #1777.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -8,9 +8,6 @@
 
 autoconf-aux            @mshinwell @xclerc
 configure.ac            @mshinwell @xclerc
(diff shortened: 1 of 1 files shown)


goldfirere:
I'm in favor of this change. Or, if there are particular requirements @mshinwell and @xclerc know to check for in these files, what are they? Then others can also review.

mshinwell:
Yeah I think this is ok.

mshinwell (approved):

PR #1781, opened 2023-08-30
Enable peephole optimization by default

xclerc:
(As per title.)


gretay-js (approved):
Overhead is [statically bounded](https://github.com/ocaml-flambda/flambda-backend/blob/895eb0c417c76dd37d5a077e152236d04a671cff/backend/peephole/peephole_optimize.ml#L6C28-L6C28).

PR #1784, opened 2023-08-30
Fix a bug in the handling of regions with exclaves

stedolan:
There is a well-named field in `Typedtree.Texp_function` called `region`: it is true if this function has a region.

There is a badly-named field in `Lambda.lfunction` called `region`: it is true if this function cannot allocate locals in its caller's region.

These two ideas used to coincide, but they don't since the addition of exclaves, which allow a function with a region (`Texp_function{region = true}`) to allocate locals in its caller's region inside an `exclave_`. This means it is incorrect to directly initialise the `region` field of `lfunction` with that of `Texp_function`.

This patch fixes the issue by computing the `region` field of lfunction by examining the body. It also weakens the invariants slightly on lfunction: it is no longer the case that `region=false` implies `nlocals>0`. (This only mattered to inhibit a function-merging optimisation in Simplif that is not valid in this case, but that optimisation is already inhibited by `region=false` alone).

### DIFF
--- a/ocaml/lambda/translcore.ml
+++ b/ocaml/lambda/translcore.ml
@@ -127,14 +127,22 @@ let transl_apply_position position =
     if Config.stack_allocation then Rc_close_at_apply
     else Rc_normal
 
+let iter_exclaves ~f lam =
+  let rec loop ~depth = function
+    | Lregion (body, _layout) ->
+      loop ~depth:(depth + 1) body
+    | Lexclave excl when depth = 0 -> f excl
+    | Lexclave excl ->
+      loop ~depth:(depth - 1) excl
+    | lam ->
+      shallow_iter lam
+        ~tail:(loop ~depth)
+        ~non_tail:ignore (* Exclaves are in tail position *)
+  in
+  loop ~depth:0 lam
+
 let may_allocate_in_region lam =
-  (* loop_region raises, if the lambda might allocate in parent region *)
-  let rec loop_region lam =
-    shallow_iter ~tail:(function
-      | Lexclave body -> loop body
-      | lam -> loop_region lam
-    ) ~non_tail:(fun lam -> loop_region lam) lam
-  and loop = function
+  let rec loop = function
     | Lvar _ | Lmutvar _ | Lconst _ -> ()
 
     | Lfunction {mode=Alloc_heap} -> ()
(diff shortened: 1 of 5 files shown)


ncik-roberts:
It's worth mentioning that we'd like to rename the badly-named `region` field in `Lambda.lfunction`, but we'll do so after I finish n-ary lambdas, a PR with which such a rename would conflict substantially. Stephen suggests (eventually):

  * add a type `Lambda.exclave_status = May_allocate_in_caller | Never_allocates_in_caller`
  * use this type in `Lambda.lfunction.region` instead of bool
  * use this type in `Lambda.Lapply.ap_mode` instead of `alloc_mode`

ocaml/lambda/translcore.ml
```diff
@@ -127,14 +127,22 @@ let transl_apply_position position =
...
+      loop ~depth:(depth + 1) body
+    | Lexclave excl when depth = 0 -> f excl
+    | Lexclave excl ->
+      loop ~depth:(depth - 1) excl
+    | lam ->
+      shallow_iter lam
+        ~tail:(loop ~depth)
+        ~non_tail:ignore (* Exclaves are in tail position *)
+  in
+  loop ~depth:0 lam
```
lpw25, line 142: I think this part is no different than the original code other than supporting nested `Lregion`. Note that nested regions are not currently supported and would break in a few places in lambda, so this is isn't strictly necessary.

ocaml/lambda/translcore.ml
```diff
@@ -1343,7 +1348,9 @@ and transl_function ~scopes e alloc_mode param arg_mode arg_sort return_sort
   let attr = default_function_attribute in
   let loc = of_location ~scopes e.exp_loc in
   let body = if region then maybe_region_layout return body else body in
-  let lam = lfunction ~kind ~params ~return ~body ~attr ~loc ~mode ~region in
+  let may_locally_allocate_in_parent = may_allocate_in_region body in
```
lpw25, line 1351: This feels a bit wasteful in the case where the function has a region and does no local allocations: we end up looking at all it's allocations twice. Is there a simple way to only do one traversal here?

mshinwell:
I believe this has been merged as part of another PR.

PR #1788, opened 2023-08-30
Backport #10364

ncik-roberts:
This backports ocaml/ocaml#10364 on top of flambda-backend. I'm backporting this because it makes backport syntactic arity slightly easier.

I don't understand the code deeply but backporting it was fairly straightforward and tests changed in the same way, with some easily-explainable exceptions:
  * Tests for internal features like layouts: The error printed in principal mode for these is now the same as the error in non-principal mode.
  * `ocaml/testsuite/tests/typing-poly/poly.ml`: this PR just removes the diff that was inserted in https://github.com/ocaml-flambda/ocaml-jst/pull/14. (It again changes the error in principal mode to be the same as the error in non-principal mode.) This also seems fine.

Run this command from the cloned  to get the diff between this PR's diff and the upstream PR's diff:

```
patdiff <(curl https://patch-diff.githubusercontent.com/raw/ocaml-flambda/flambda-backend/pull/1788.diff) <(curl https://patch-diff.githubusercontent.com/raw/ocaml/ocaml/pull/10364.diff)
```

The two thorny-looking sections of the diff-of-diffs are the testsuite changes (already explained) and the fact that the body of `type_unknown_args` is different, but that is easy enough to review on its own.

### DIFF
--- a/ocaml/typing/typecore.ml
+++ b/ocaml/typing/typecore.ml
@@ -678,13 +678,19 @@ let extract_option_type env ty =
     Tconstr(path, [ty], _) when Path.same path Predef.path_option -> ty
   | _ -> assert false
 
+let protect_expansion env ty =
+  if Env.has_local_constraints env then generic_instance ty else ty
+
 type record_extraction_result =
   | Record_type of Path.t * Path.t * Types.label_declaration list * record_representation
   | Not_a_record_type
   | Maybe_a_record_type
 
+let extract_concrete_typedecl_protected env ty =
+  extract_concrete_typedecl env (protect_expansion env ty)
+
 let extract_concrete_record env ty =
-  match extract_concrete_typedecl env ty with
+  match extract_concrete_typedecl_protected env ty with
   | Typedecl(p0, p, {type_kind=Type_record (fields, repres)}) ->
     Record_type (p0, p, fields, repres)
   | Has_no_typedecl | Typedecl(_, _, _) -> Not_a_record_type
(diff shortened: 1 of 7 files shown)


ocaml/typing/typecore.ml
```diff
@@ -3342,6 +3348,12 @@ let collect_unknown_apply_args env funct ty_fun mode_fun rev_args sargs ret_tvar
 let collect_apply_args env funct ignore_labels ty_fun ty_fun0 mode_fun sargs ret_tvar =
   let warned = ref false in
   let rec loop ty_fun ty_fun0 mode_fun rev_args sargs =
+    let type_unknown_args () =
+      (* We're not looking at a *known* function type anymore, or there are no
+         arguments left. *)
+      collect_unknown_apply_args env funct ty_fun0 mode_fun rev_args sargs ret_tvar
+    in
+    if sargs = [] then type_unknown_args () else
```
riaqn, line 3356: I think this line is not needed, because in the immediately following match, `sargs` is matched as well. In that case, `type_unknown_args` is used only once - maybe we don't need a function for it?

riaqn (approved):
LGTM

ocaml/typing/typecore.ml
ncik-roberts: The idea is to skip the `expand_head` if the list is empty, which is extra work and changes which programs are accepted. I did this only because the upstream patch does the same; this is actually responsible for the diff in `ocaml/testsuite/tests/typing-gadts/ambivalent_apply.ml` that also shows up upstream.

riaqn: That makes sense - thanks.

PR #1791, opened 2023-09-01
Optimise Switch expressions using lookup tables

mshinwell:
`Switch` expressions where all of the arms have the same destination, with a single constant argument, are easily constructed.  The first `match` here produces one such:
```
type t = A | B | C | D | E
type i = F | G | H | I | J

let f i g =
  let t =
    match i with
    | F -> B
    | G -> C
    | H -> A
    | I -> D
    | J -> E
  in
  match t with
  | A -> g 10
  | B -> g 20
  | C -> g 30
  | D -> g 40
  | E -> g 50
```
With a large number of arms the code generated for these expressions is poor, with excessively long jump tables taking a lot of code size.  This patch simply uses a lookup table in these cases instead.

It is true that the particular example above and others like it could be resolved by a full match-in-match optimisation.  However I think this table optimisation is still valuable as it applies in many cases where the alternative would be a lot of continuation specialisation, which is likely difficult and expensive to do.


Gbury (changes requested):

middle_end/flambda2/simplify/simplify_switch_expr.ml
```diff
@@ -254,14 +254,96 @@ let rebuild_switch ~arms ~condition_dbg ~scrutinee ~scrutinee_ty
...
+                 actions *)
+              match expected_dest with
+              | Some expected_dest
+                when not
+                       (Continuation.equal
+                          (Apply_cont.continuation dest)
+                          expected_dest) ->
+                None
+              | Some _ | None -> (
+                if (* Discriminants must be 0..(num_arms-1) *)
```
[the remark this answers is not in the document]
mshinwell: Hmm, I actually thought other varieties of switches were possible in flambda2, which is why I added the check in this form.

middle_end/flambda2/simplify/simplify_switch_expr.ml
```diff
@@ -254,14 +254,96 @@ let rebuild_switch ~arms ~condition_dbg ~scrutinee ~scrutinee_ty
...
+      match dest_and_args_rev_and_expected_discr with
+      | None | Some (None, _, _) | Some (_, [], _) -> None
+      | Some (Some dest, args_rev, _) -> (
+        let args = List.rev args_rev in
+        assert (List.compare_length_with args 1 >= 0);
+        (* For the moment just do this for things that can be put in scannable
+           blocks. *)
+        match Reg_width_const.descr (List.hd args) with
+        (* All arguments must be of the same kind *)
+        | Naked_immediate _ ->
```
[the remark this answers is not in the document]
mshinwell: I don't think any tagging needs doing here -- these are `Targetint_31_63.t` values.

middle_end/flambda2/simplify/simplify_switch_expr.ml
```diff
@@ -280,6 +362,108 @@ let rebuild_switch ~arms ~condition_dbg ~scrutinee ~scrutinee_ty
...
+            let block_sym =
+              let var = Variable.create "switch_block" in
+              Symbol.create
+                (Compilation_unit.get_current_exn ())
+                (Linkage_name.of_string (Variable.unique_name var))
+            in
+            let uacc =
+              let fields =
+                List.map
+                  (fun const -> Field_of_static_block.Tagged_immediate const)
```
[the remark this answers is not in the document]
mshinwell: I don't think I agree, these `Targetint_31_63.t` values always contain the plain native representation of the integer concerned.  (See e.g. `to_cmm_static.ml` line 32.)

middle_end/flambda2/simplify/simplify_switch_expr.ml
```diff
@@ -210,15 +205,169 @@ let find_cse_simple dacc required_names prim =
...
+              let expected_discr = TI.add TI.one expected_discr in
+              Some (Some dest', const :: args_rev, expected_discr))))
+  in
+  match TI.Map.fold check_arm arms (Some (None, [], TI.zero)) with
+  | None | Some (None, _, _) | Some (_, [], _) -> None
+  | Some (Some dest, args_rev, _) -> (
+    let args = List.rev args_rev in
+    assert (List.compare_length_with args 1 >= 0);
+    (* For the moment just do this for things that can be put in scannable
+       blocks. *)
```
Gbury, line 249:
```suggestion
       blocks (after adding some tag/untag operations if needed). *)
```

middle_end/flambda2/simplify/simplify_switch_expr.ml
```diff
@@ -210,15 +205,169 @@ let find_cse_simple dacc required_names prim =
...
+      (NO.remove_var free_names_of_body ~var:final_arg_var)
+  in
+  let increase_in_code_size =
+    (* Very likely negative. *)
+    Code_size.( - )
+      (Code_size.( + )
+         (Code_size.apply_cont apply_cont)
+         (match must_untag_lookup_table_result with
+         | Must_untag -> Code_size.prim untag_arg_prim
+         | Leave_as_tagged_immediate -> Code_size.zero))
```
Gbury, line 355: Shouldn't you also add the cost of the block load  ?

middle_end/flambda2/terms/code_size.ml
```diff
@@ -423,4 +425,4 @@ let of_int t = t
 
 let to_int t = t
 
-let evaluate ~args:_ t = float_of_int t
+let evaluate ~args:_ t = float_of_int (max 0 t)
```
Gbury, line 428: won't this ignore negative code size changes ?

Gbury (changes requested):
All is good except for my comment about the code size change (and the missing block load size).

middle_end/flambda2/simplify/simplify_switch_expr.ml
mshinwell: Fixed

middle_end/flambda2/terms/code_size.ml
mshinwell: Fixed

middle_end/flambda2/simplify/simplify_switch_expr.ml
mshinwell: Fixed

Gbury (approved):

PR #1793, opened 2023-09-02
Fix incorrect sort assumption in lambda for `bop_exp`s in letops

ccasin:
This fixes a bug where we incorrectly assumed that the `bop_exp` part of a letop must have layout value.  The `bop_exp` is the argument of a letop or andop binding, which doesn't need to be value (though the binder itself does, if there is at least one andop, which is probably the source of the original confusion).

Concretely in:
```ocaml
let* x = e1 in
... x ...
```
We were incorrectly assuming `e1` is a value, in transl.  It needn't be.  In this case `x` also need not be a value.
 
In:
```ocaml
let* x = e1
and* y = e2 in
... x ... y ...
```
We were incorrectly assuming `e1` and `e2` are values.  They also need not be, though in this case `x` and `y` must be values (because we put them in a tuple in lambda).

I've updated the `letop` tests to reflect this.  Rather than just replacing the old `void` versions with new `float64` versions, I kept both, because void still behaves differently (the float examples can be translated, while the void examples hit the lambda layout sanity checks, which seems worth testing still).

Review: @goldfirere , probably.


ccasin:
Annoyingly, this seems to hit an (unrelated?) issue in the legacy build version of `cmm_helpers`:
```
/home/ccasinghino/tmp/foo.o: In function `camlFoo__18':
:(.data+0xd0): undefined reference to `caml_tuplify2_RF'
collect2: error: ld returned 1 exit status
File "caml_startup", line 1:
Error: Error during linking (exit code 1)
```
I've pushed a commit that "fixes" it by copying over what is done in the other version of `cmm_helpers`, but I'm not sure this is the right fix (it's not obvious to me why we need different versions of `caml_tuplify` for different return sorts anyway).  @mshinwell, it looks like you made the corresponding change in `backend`, can you advise?

Here's a program that triggers the issue with the legacy build (must be compiled with `-extension layouts_alpha`):
```ocaml
module Float_u = struct
  let of_float = Stdlib__Float_u.of_float
  let to_float = Stdlib__Float_u.to_float
  let ( + ) = Stdlib__Float_u.add
  let ( - ) = Stdlib__Float_u.sub
end

let ( let* ) x (f : _ -> float#) = f x
let ( and* ) x y = Float_u.(x, 1.2)
let _ =
  let _result =
    let* x = 42.0
    and* y = Float_u.of_float 3.3 in
    Float_u.of_float (x +. y)
  in
  ()
```

Ekdohibs:
The change to the legacy cmm_helpers is needed, since the generation of the names of the tuplify functions will otherwise not match the names given to the functions (potentially even causing duplicate symbols).

Concerning whether the return layout is needed for caml_curry/tuplify/apply, since it always corresponds to tail calls, it should a priori not be needed. However, cmm has no concept of tail calls, so we need to give a return layout for the calls at the end, preventing us from making a single function (besides, with unboxed tuples and results being passed on the stack, it might even be the case that the calling convention could change in those cases; I am not sure).

ocaml/typing/typecore.ml
```diff
@@ -7680,19 +7684,22 @@ and type_let
   let new_env = add_module_variables new_env mvs in
   (l, new_env)
 
-and type_andops env sarg sands expected_ty =
-  let rec loop env let_sarg rev_sands expected_ty =
+and type_andops env sarg sands expected_sort expected_ty =
+  let rec loop env let_sarg rev_sands expected_sort expected_ty =
```
goldfirere, line 7688:
```
(* Pass arguments to [loop] to avoid allocating closure; [env] and [let_sarg] get passed down unchanged *)
```

goldfirere (approved):

PR #1795, opened 2023-09-04
Add tag to Block_approximation

mshinwell:
This fixes a bug where all block types reconstructed for Simplify from the classic mode approximations would be given tag zero.


lthls (approved):
I was wondering whether it would be better to use `Tag.t ` or `Tag.Scannable.t`.
Since the types use `Tag.t`, I think `Tag.t` is simpler, but at some point we should probably switch both of them to `Tag.Scannable.t`
Everything looks correct apart from this detail.

Issue #1782, opened 2023-08-30
Simplify terminator when destination is known


xclerc:
In some circumstances, we know the destination of
a terminator. So for instance the following:
```
[mov const, reg] --> [] --[reg==const']--> [block1]
               ... --^ `--[reg!=const']--> [block2]
```
could be rewritten to:
```
[mov const, reg] --> [blocki]
... --> [] --[reg==const']--^ [block1]
          `--[reg!=const']--> [block2]
```

PR #1809, opened 2023-09-06
Simplify terminator integer condition

xclerc:
We can sometimes statically determine the
destination of a conditional terminator, and
thus change it to a mere jump.

This pull request does so if the following case:
- a first block ends with an `Iconst_int`
  instruction followed by an `Always` terminator;
- the next block has an empty body and a
  terminator which is an integer condition 
  over the same register used by the
  `Iconst_int` instruction;

by changing the destination of the `Always` terminator.
(ASCII art available at #1782)

Fixes https://github.com/ocaml-flambda/flambda-backend/issues/1782

### DIFF
--- a/backend/cfg/cfgize.ml
+++ b/backend/cfg/cfgize.ml
@@ -910,10 +910,10 @@ let fundecl :
      terminator simplification. *)
   Profile.record ~accumulate:true "optimizations"
     (fun () ->
-       if simplify_terminators then begin
-         Merge_straightline_blocks.run cfg_with_layout;
-         Simplify_terminator.run cfg
-       end;
+      if simplify_terminators
+      then (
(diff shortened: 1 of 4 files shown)


Ekdohibs:
This looks good, but could we extend it a bit to the case where the block has multiple assignments? Something like a block ending in `reg1 = 0, reg2 = 0` followed by a test on `reg1` (currently only a test on `reg2` would be optimized), or even a block containing `reg1 = 0` then no further modification to `reg1`?

xclerc:
I agree in principle, but would be tempted to
merge as-in and register your suggestion in
an issue. Does that sound OK?

mshinwell:
That sounds like the right course of action to me - these PRs are overdue for merging as-is.

Ekdohibs (approved):

PR #1811, opened 2023-09-06
Support mode crossing at identifiers

riaqn:
This PR adds mode crossing at `Texp_ident`. One reason of this is to remedy the counter-intuitive behaviour of `type_let`, which checks the LHS before looking at RHS. See the added test for an example why this is needed. 

Kindly request review from @stedolan or @lpw25

### DIFF
--- a/ocaml/testsuite/tests/typing-local/crossing.ml
+++ b/ocaml/testsuite/tests/typing-local/crossing.ml
@@ -345,7 +345,7 @@ val f : local_ M.t -> M.t = <fun>
 val f : local_ t2 -> t2 = <fun>
 |}]
 
-(* This test needs the snapshotting in is_always_global to prevent a type error
+(* This test needs the snapshotting in [is_immediate] to prevent a type error
    from the use of the gadt equation in the inner scope. *)
 type _ t_gadt = Int : int t_gadt
 type 'a t_rec = { fld : 'a }
--- a/ocaml/testsuite/tests/typing-local/crossing.ml
+++ b/ocaml/testsuite/tests/typing-local/crossing.ml
@@ -437,3 +437,14 @@ Line 1, characters 12-39:
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 Error: Type int -> local_ int is not a subtype of local_ int -> int
 |}]
+
+(* Mode crossing at identifiers - in the following, x and y are added to the
+environment at mode local, but they cross to global when they are refered to
+again. Note that ref is polymorphic and thus doesn't trigger crossing. *)
+let foo () =
+  let x, y = local_ (42, 24) in
+  let _ = ref x, ref y in
+  ()
+[%%expect{|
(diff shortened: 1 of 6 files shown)


goldfirere:
I know this is a tiny PR that I have no business reviewing, but given my modal kinds work, I thought I would take a peek at this work on mode crossing. Looking into this all, I have a number of questions / concerns:

1. `mode_cross_to_min` calls `mode_cross` calls `is_always_global`. But `mode_cross` does *not* consider linearity or uniqueness. This seems wrong. For example:

```ocaml
module M : sig
  type t : immediate (* we can say this now! but ocamlformat can't format it yet, so none of our friends can use it :( *)
end = struct
  type t = int
  val create : unit -> once_ t
end

let stop_counting : once_ t -> t = fun x -> x
```

The author of `M` wants to have non-duplicable tokens. But `stop_counting` can get around the restriction.

I suppose I can't use this to actually cause trouble because the soundness of our system relies only on tracking onceness of closures, which are never immediate. But it still makes me uncomfortable. Though maybe it's the same discomfort that arises with locals that I'm trying to solve with modal kinds. Still, there's a delicate dependence on the applicability of `is_always_global` here. For example, if we allowed `Gel.t` to mode-cross, we'd be in deep trouble here.

2. I'm surprised that identifiers are a special case for mode crossing. I would just expect this to work without adding any code. Just guessing here, but it's probably like generalization: in theory you can do it anywhere, but in practice we do it just at `let`. And there are only a few places that `mode_cross_to_min` (or its dual `expect_mode_cross`) is called. What's the theory that tells us exactly where these calls should be made? I mean, I can imagine it, but is it written down?

riaqn:
- I think `is_always_global` needs to be renamed after the introduction of the new axes (uniqueness and linearity). It should probably just be called `is_immediate64`. I think the problem you illustrated here is less about the difference between the axes - I feel that all three axes we have now are about memory/reference safety, wrt which immediates should behave very similarly (i.e., it crosses all three axes).

- I think the issue here is that we want some types to not cross modes even though they are immediate. My impression is that we will have the modal kind system to do that? For example, we will have a kind that doesn't cross any mode, and `M.t` while internally being of `immediate`, will be exposed as of that kind? For now, though, I think the author of `module M` can try to not expose the fact that `M.t` is immediate.

goldfirere:
OK. That all makes sense. Part of the challenge is that I view these modes as primarily features meant to aid programmers in expressing their abstractions, where unexpected mode crossing is an abstraction violation. But this is probably premature, in that today the modes are really about expressing certain constraints on memory.

Your last bullet is helpful. Maybe add that as a comment in the PR?

riaqn:
Mind to take a look again? @goldfirere

goldfirere (approved):
In my gut, I'm still a little unhappy that `immediate` means "crosses all the modes", but I understand this is what we're going with. When we announce this stuff more widely, we should make this point very clear.

PR #1816, opened 2023-09-08
Add a test for an interaction between omitted mli and overeager heap allocation of argument

ncik-roberts:
Add tests showing whether function arguments are stack allocated in the presence/absence of an mli for the caller. This documents a behavior that I couldn't otherwise find documentation for.

The only possibly-surprising case is this one:

```ocaml
(* in test.ml, with no test.mli *)
type t = { a : int; b : int }
let consume { a; b } = a + b
let run a b = consume { a; b }
(* allocates! *)
let _ = run 3 4
```

That's because:
  * in the no-mli case, the types of bindings in the inferred signature are straightforwardly shared from the types inferred for the implementation;
  * further, in the no-mli case, any unconstrained mode variable in the signature is equated to the legacy mode.

These have the net effect of constraining mode variables in the *implementation* to the legacy mode as well, which means that callers (such as `run`) will heap allocate objects that could be stack allocated.

I don't attempt to fix this in this patch, but how could we? As suggested by @stedolan , we could improve this use case by copying the types (with fresh mode variables constrained to the legacy mode) instead of sharing them. The copied version's mode variables could then be related to the original version's via submoding. Then, internal callers like `run` could stack allocate the argument to `consume`. (It's not essential to support this use case &mdash; users should prefer writing mlis anyway.)

### DIFF
--- a/ocaml/jane/doc/local-reference.md
+++ b/ocaml/jane/doc/local-reference.md
@@ -79,6 +79,35 @@ function types, see below) then inference will resolve them according to what
 appears in the `.mli`. If there is no `.mli` file, then inference will always
 choose `global` for anything that can be accessed from another file.
 
+Local annotations (or the lack thereof) in the mli don't affect inference
+within the ml. In the below example, the `~foo` parameter is inferred to
+be local internally to `A`, so `foo:(Some x)` can be constructed locally.
+
+```ocaml
+(* in a.mli *)
+val f1 : foo:local_ int option -> unit
+val f2 : int -> unit
+
+(* in a.ml *)
+let f1 ~foo:_ = ()
+let f2 x = f1 ~foo:(Some x) (* [Some x] is stack allocated *)
+```
+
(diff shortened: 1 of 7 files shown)


goldfirere (approved):
All the changes here look good to me. But there is something missing: documentation. The PR description says that this documents the behavior, but it doesn't quite. I think we've agreed that the best place to do so is in our internal documentation system, but I'd also favor putting a note in `ocaml/jane/doc`, which is an in-repo probably-out-of-date reformulation of what's in our internal documentation system. (Ideally we'd have an automated way of keeping this in sync, but I understand how that might be hard.)

Approving in the meantime, because I think this request for documentation goes beyond the usual standard.

ncik-roberts:
I pushed some (intended-to-be) user-facing documentation to `ocaml/jane/doc`. @goldfirere, would you mind taking a look before I import it internally?

goldfirere (approved):

PR #1819, opened 2023-09-11
Fixed off by one in cached generated function file

poechsel:
Fix an off-by-one between generation of the cached functions file and filtering of these functions

In the `cached-generic-functions.o` caml_curry are generated using a `Seq.init (Lambda.max_arity()) (fun n -> ...caml_curry{n}...)` - making the largest one to be pre-generated be `caml_curry125`.
However, when linking, all `caml_curry{n}` with `n <= Lambda.max_arity()` are considered to be part of the
pre-generated object file. This mean that `caml_curry126` was believe to be pre-generated, which unfortunately wasn't the case.


gretay-js (approved):
The fix looks fine to me but I wonder if we should have some tests for it.

PR #1820, opened 2023-09-11
Fix uncaught Unify exception in filter_arrow

alanechang:
The included testcase current results in an uncaught exception:

```
Fatal error: exception Ctype.Unify_trace(_)
Raised at Ctype.raise_trace_for in file "ocaml/typing/ctype.ml", line 99, characters 16-41
Called from Ctype.filter_arrow in file "ocaml/typing/ctype.ml", line 3861, characters 6-51
Called from Typecore.type_function in file "ocaml/typing/typecore.ml", line 6133, characters 8-60
Called from Builtin_attributes.warning_scope in file "ocaml/parsing/builtin_attributes.ml", line 415, characters 14-18
Re-raised at Builtin_attributes.warning_scope in file "ocaml/parsing/builtin_attributes.ml", line 420, characters 4-13
Called from Typecore.type_expect in file "ocaml/typing/typecore.ml", line 4476, characters 4-181
Called from Typecore.type_cases.(fun) in file "ocaml/typing/typecore.ml", line 7235, characters 10-106
...
```

This PR seeks to fix this by wrapping the error as a proper `Filter_arrow_failed` exception before raising.

### DIFF
--- a/ocaml/testsuite/tests/typing-layouts/basics_beta.ml
+++ b/ocaml/testsuite/tests/typing-layouts/basics_beta.ml
@@ -295,7 +295,7 @@ Error: Signature mismatch:
          val x : ('a : immediate). 'a
        is not included in
          val x : string
-       The type string is not compatible with the type string
+       The type ('a : immediate) is not compatible with the type string
        string has layout value, which is not a sublayout of immediate.
 |}];;
 
@@ -332,7 +332,8 @@ Error: Signature mismatch:
          val x : ('a : immediate). 'a t
        is not included in
          val x : string
-       The type string t = string is not compatible with the type string
+       The type 'a t = ('a : immediate) is not compatible with the type
+         string
        string has layout value, which is not a sublayout of immediate.
 |}]
 
--- a/ocaml/testsuite/tests/typing-layouts/basics_beta.ml
+++ b/ocaml/testsuite/tests/typing-layouts/basics_beta.ml
@@ -545,3 +545,26 @@ Line 3, characters 15-40:
                    ^^^^^^^^^^^^^^^^^^^^^^^^^
 Error: Type 'a has layout value, which is not a sublayout of immediate.
 |}]
+
+(****************************************************)
+(* Test 35: check bad layout error in filter_arrow *)
+
+module M35_1 : sig
+  val x : string
+end = struct
+  type ('a : immediate) t = 'a
+
+  let f : 'a t -> 'a = fun x y z -> x
+
+  let x = f (assert false)
+
+end;;
+
+[%%expect {|
+Line 6, characters 23-37:
+6 |   let f : 'a t -> 'a = fun x y z -> x
+                           ^^^^^^^^^^^^^^
+Error: This expression has type 'a -> 'b
+       but an expression was expected of type 'a -> 'b
+       'a -> 'b has layout value, which is not a sublayout of immediate.
+|}]
--- a/ocaml/testsuite/tests/typing-layouts/basics_beta.ml
+++ b/ocaml/testsuite/tests/typing-layouts/basics_beta.ml
@@ -549,21 +549,15 @@ Error: Type 'a has layout value, which is not a sublayout of immediate.
 (****************************************************)
 (* Test 35: check bad layout error in filter_arrow *)
 
-module M35_1 : sig
-  val x : string
-end = struct
-  type ('a : immediate) t = 'a
 
-  let f : 'a t -> 'a = fun x y z -> x
-
-  let x = f (assert false)
-
-end;;
+type ('a : immediate) t35 = 'a
+let f : 'a t35 -> 'a = fun x y -> ()
 
 [%%expect {|
-Line 6, characters 23-37:
-6 |   let f : 'a t -> 'a = fun x y z -> x
-                           ^^^^^^^^^^^^^^
+type ('a : immediate) t35 = 'a
+Line 2, characters 23-36:
+2 | let f : 'a t35 -> 'a = fun x y -> ()
+                           ^^^^^^^^^^^^^
 Error: This expression has type 'a -> 'b
        but an expression was expected of type 'a -> 'b
        'a -> 'b has layout value, which is not a sublayout of immediate.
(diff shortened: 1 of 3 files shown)


ocaml/testsuite/tests/typing-layouts/basics_beta.ml
```diff
@@ -545,3 +545,26 @@ Line 3, characters 15-40:
...
+
+  let x = f (assert false)
+
+end;;
+
+[%%expect {|
+Line 6, characters 23-37:
+6 |   let f : 'a t -> 'a = fun x y z -> x
+                           ^^^^^^^^^^^^^^
+Error: This expression has type 'a -> 'b
```
goldfirere, line 567: This error message is pretty bad. I suppose it's better than tthrowing an exception. But can we do better before all the layout histories work is merged?

ocaml/typing/ctype.ml
```diff
@@ -3858,7 +3858,17 @@ let filter_arrow env t l ~force_tpoly =
     Tvar { layout } ->
       let t', arrow_desc = function_type (get_level t) in
       link_type t t';
-      constrain_type_layout_exn env Unify t' layout;
+
+      begin match constrain_type_layout env t' layout with
```
goldfirere, line 3862: I would prefer this before the call to `link_type`. It probably doesn't matter, but I would always want `link_type` to link two types with compatible layouts.

goldfirere (commented):
Quick work at figuring out where the problem is! A few suggestions below.

ocaml/typing/ctype.ml
ccasin: I'd also drop the leading and trailing whitespace to match the style of the compiler.

ocaml/testsuite/tests/typing-layouts/basics_beta.ml
```diff
@@ -545,3 +545,26 @@ Line 3, characters 15-40:
                    ^^^^^^^^^^^^^^^^^^^^^^^^^
 Error: Type 'a has layout value, which is not a sublayout of immediate.
 |}]
+
+(****************************************************)
+(* Test 35: check bad layout error in filter_arrow *)
+
+module M35_1 : sig
```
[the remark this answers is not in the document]
alanechang: Simplified the test case by quite a bit. Leaving only `t` and `f`. I tested it with the unpatched version and verified that it still triggers the bug.

ocaml/typing/ctype.ml
alanechang: Changed the call order and removed the extra whitespaces.

There are two other cases of `link_type` being called before `constrain_type_layout_exn`:

https://github.com/ocaml-flambda/flambda-backend/blob/5f80067709edffb71ad1bd833dfd15c562c0e684/ocaml/typing/ctype.ml#L4374
https://github.com/ocaml-flambda/flambda-backend/blob/5f80067709edffb71ad1bd833dfd15c562c0e684/ocaml/typing/ctype.ml#L4390

Should we change those too?

ocaml/testsuite/tests/typing-layouts/basics_beta.ml
ccasin: Does dropping the `Diff` element of the trace list make the error any better?  If not, one could consider adding a new `filter_arrow_failure` case.  The revised one may be good enough for now, though.

alanechang: Dropping the `Diff` element results in a similar error but with the first two lines removed:

```
> --- basics_beta.ml	2023-09-11 16:14:09.400891238 -0400
> +++ basics_beta.ml.corrected.corrected	2023-09-11 16:44:50.991568189 -0400
> @@ -557,7 +557,6 @@
>  Line 2, characters 19-31:
>  2 | let f35 : 'a t35 = fun () -> ()
>                         ^^^^^^^^^^^^
> -Error: This expression has type 'b -> 'c
> -       but an expression was expected of type ('a : immediate)
> +Error:
>         'a -> 'b has layout value, which is not a sublayout of immediate.
>  |}]
> 
> )
```

ccasin: This looks like an improvement to me, since the problem with the old error was, in part, that it was talking about a difference between two types that didn't obviously correspond to something meaningful.  But I wonder what Richard thinks.

ocaml/typing/ctype.ml
goldfirere: Yes, probably. Would be curious to know whether that changes error messages. My desire for the reordering is purely because it just makes more sense to me in the other order, but if it improves errors, that would be great!

ocaml/testsuite/tests/typing-layouts/basics_beta.ml
```diff
@@ -545,3 +545,19 @@ Line 3, characters 15-40:
...
+let f35 : 'a t35 = fun () -> ()
+
+[%%expect {|
+type ('a : immediate) t35 = 'a
+Line 2, characters 19-31:
+2 | let f35 : 'a t35 = fun () -> ()
+                       ^^^^^^^^^^^^
+Error: This expression has type 'b -> 'c
+       but an expression was expected of type ('a : immediate)
+       'a -> 'b has layout value, which is not a sublayout of immediate.
```
goldfirere, line 562: Yikes. How did the `'b -> 'c` become `'a -> 'b`?

ocaml/testsuite/tests/typing-layouts/basics_beta.ml
goldfirere: Yes, I think this smaller error is better. Though there is a weird extra newline after the `Error:` that might be worth figuring out. (Sometimes tracking down these weird inconsistencies reveals something interesting. Sometimes it's just an extra `\n` though.)

ocaml/typing/ctype.ml
goldfirere: Ooh. Reordering made a nice improvement to error messages. Nice.

goldfirere (approved):

PR #1821, opened 2023-09-12
Finish moving `any` to `layouts_beta`

ccasin:
As discussed in #1812.  This PR builds is on top of that one, to avoid conflicts in test files.  Look at just the last commit.


goldfirere (commented):
Given the discussion yesterday, I was expecting to see the test cases removed from the alpha tests, instead of duplicated. What do you think here?

Everything else looks good.

ccasin:
> Given the discussion yesterday, I was expecting to see the test cases removed from the alpha tests, instead of duplicated. What do you think here?
> 
> Everything else looks good.

I guess I am surprised at that.  I thought the plan was to leave existing tests alone but not add new duplication in the future.

goldfirere:
But this commit *does* add duplication, doesn't it? I understood us to say that we wouldn't go through already-existing duplication and strip it out (other than perhaps optionally).

ccasin:
> But this commit _does_ add duplication, doesn't it? I understood us to say that we wouldn't go through already-existing duplication and strip it out (other than perhaps optionally).

Yes, poorly phrased on my part.

The thing I am sure we agreed on was not bothering to add new tests to multiple files where adding them to one offers most of the benefit.  That's nice, because it's less work and smaller diffs, even if it means we're not running some tests we might want.  Here you're suggesting (marginally) more work and larger diffs with the effect of removing tests we might want, and plan to run again one day when we can combine the files.  I'm less excited about that.

goldfirere (approved):

PR #1825, opened 2023-09-13
Remove duplicated creation_reasons from layout history error messages

alanechang:
This PR aims to improve layout history error messages by removing duplicated `creation_reason`s. For example:

```
module M11_2 = struct
  let foo x = VV (x # getvoid)
end;;
[%%expect{|
Line 2, characters 17-30:
2 |   let foo x = VV (x # getvoid)
                     ^^^^^^^^^^^^^
       but an expression was expected of type t_void
       But the layout of t_void must be a sublayout of value, because all of the following:
           an object field
           an object field
|}];;
```

will turn into

```
...
       but an expression was expected of type t_void
         an object field.
|}];;
```

This is currently implemented using polymorphic equal checks. An alternative will be to define `compare`/`equal` functions for all `*_creation_reason` types. Would love some input on whether it will be worth it to do that.

### DIFF
--- a/ocaml/typing/layouts.ml
+++ b/ocaml/typing/layouts.ml
@@ -814,7 +814,7 @@ module Layout = struct
       | Sublayout ->
         fprintf ppf "sublayout check"
 
-    (* CR layouts v2.9: An older implementation of format_flattened_history existed
+    (* CR layouts: An older implementation of format_flattened_history existed
        which displays more information not limited to one layout and one creation_reason
        around commit 66a832d70bf61d9af3b0ec6f781dcf0a188b324d in main.
 
--- a/ocaml/typing/layouts.ml
+++ b/ocaml/typing/layouts.ml
@@ -823,6 +823,7 @@ module Layout = struct
 
        INVARIANT: the creation_reasons within a list all are reasons for
        the layout they are paired with.
+       INVARIANT: the creation_reasons do not store duplicates
        INVARIANT: L is a sublayout of all the Li in a flattened_history.
        INVARIANT: If Li and Lj are stored in different entries in a
        flattened_history, then not (Li <= Lj) and not (Lj <= Li).
@@ -846,9 +847,11 @@ module Layout = struct
         let layout_desc = get_internal layout in
         let rec go acc = function
           | ((key, value) as row) :: rest ->
+            let is_unseen reason = List.fold_left (fun acc r -> acc && not (r = reason)) true value in
             begin match sub_desc layout_desc key with
             | Sub -> go acc rest
-            | Equal -> (key, reason :: value) :: acc @ rest
+            | Equal when is_unseen reason -> (key, reason :: value) :: acc @ rest
+            | Equal -> row :: acc @ rest
             | Not_sub -> go (row :: acc) rest
             end
           | [] -> (layout_desc, [reason]) :: acc
--- a/ocaml/typing/layouts.ml
+++ b/ocaml/typing/layouts.ml
@@ -963,7 +963,7 @@ module Layout = struct
     if flattened_histories
     then begin match sub_desc (get_internal lhs.layout) (get_internal rhs.layout) with
       | Sub -> lhs.history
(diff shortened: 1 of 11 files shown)


goldfirere (commented):
Upon reading this, I think there is a better implementation strategy:

When `flattened_histories` is true, just keep the most recent reason. Very simple! Basically, we'd have an invariant that says if `flattened_histories` is true, then a `history` is always a `Creation`. (That's probably a bad constructor name.) When interacting two existing layouts, we'd want to retain the "better" reason (the one associated with a sublayout if the two layouts are not equal, or the non-`Concrete_reason` if they are, or an arbitrary one otherwise). Then printing becomes a breeze. And we can always set `flattened_histories` to false and get the tree back. This approach seems easier to implement (no awkward tree search) and faster.

Do you see any problems with this?

Sorry for not coming up with the simplified plan sooner!

alanechang:
That sounds like a great idea! We can save the cost of building up the tree in the first place.

I made an attempt at implementing that.

ocaml/typing/layouts.ml
```diff
@@ -1015,8 +960,18 @@ module Layout = struct
   let equate = equate_or_equal ~allow_mutation:true
 
   let combine_histories reason lhs rhs =
-    Interact { reason; lhs_layout = lhs.layout; lhs_history = lhs.history;
-               rhs_layout = rhs.layout; rhs_history = rhs.history }
+    if flattened_histories
+    then begin match sub_desc (get_internal lhs.layout) (get_internal rhs.layout) with
+      | Sub -> lhs.history
+      | Not_sub -> rhs.history
```
goldfirere, line 966:
```suggestion
      | Not_sub -> rhs.history  (* CR layouts: this will be wrong if we ever have a non-trivial meet in the layout lattice *)
```

ocaml/typing/layouts.ml
```diff
@@ -1015,8 +960,18 @@ module Layout = struct
...
+      | Equal -> begin match lhs.history, rhs.history with
+        (* Prefer other creation_reasons over Concrete_creation *)
+        | h, Creation (Concrete_creation _)
+        | Creation (Concrete_creation _), h -> h
+        | h, _ -> h
```
goldfirere, line 971: Might be interesting to compare error messages with `h, _` vs `_, h`. Not sure what to expect there. Is one better than the other?

goldfirere (approved):
Looks great. It might be worth checking to see whether combining histories should be left-biased or right-biased. Give it a try and just eyeball which looks better -- no need to be super scientific here. But if one clearly is better and you think you know why, it would be good to record it.

ocaml/typing/layouts.ml
alanechang: Did a quick test and the difference between the two are quite minimal.

Going from left-biased to right-biased resulted in these two types of diff across the test cases:

```diff
--- a/ocaml/testsuite/tests/typing-layouts/modules_beta.ml
+++ b/ocaml/testsuite/tests/typing-layouts/modules_beta.ml
@@ -161,7 +161,7 @@ Error: This expression has type string but an expression was expected of type
        The layout of string is value, because
          it equals the primitive value type string.
        But the layout of string must be a sublayout of immediate, because
-         of the annotation on 'a in the declaration of the type t.
+         of the annotation on 'a in the declaration of the type s2'.
 |}]
```

```diff
--- a/ocaml/testsuite/tests/typing-layouts-float64/parsing.ml
+++ b/ocaml/testsuite/tests/typing-layouts-float64/parsing.ml
@@ -105,7 +105,7 @@ Error: This type float# should be an instance of type ('a : value)
        The layout of float# is float64, because
          it equals the primitive value type float#.
        But the layout of float# must be a sublayout of value, because
-         a term-level argument to a class constructor.
+         a type argument defaulted to have layout value.
 |}];;
```

The left-biased one looks to be a bit better.

goldfirere: OK. Let's go with that then.

PR #1827, opened 2023-09-14
tmc: Remove close-on-apply flag when producing a call in non-tail position

lukemaurer:
The close-on-apply flag is invalid on a function call in non-tail position, and recent flambda1 changes cause us to notice this.


mshinwell:
@lpw25 @stedolan could one of you review please?

lpw25 (approved):
LGTM

PR #1829, opened 2023-09-14
Make -extension immutable_arrays on by default

goldfirere:
Now that they're available in foundation, I see little reason to keep these guarded behind a flag.


ccasin (approved):
Yep, looks good.

PR #1826, opened 2023-09-14
Cap max arity of caml_apply

poechsel:
Functions can have a max arity of `Lambda.max_arity ()`. It is thus a waste to generate `caml_applyN` with `N > max_arity ()` as the same thing could be achieved by calling `caml_apply128` enough times until we need to call `caml_applyN` with `N < max_arity ()`

This was tested using the following example:
```
[...]

PR #1830, opened 2023-09-15
Split generated functions cache according to arity

poechsel:
Superseeds #1755

Once #1826 goes in this will split the generated function cache in one object file per generated function and arity. Overall this creates 200 different object files. The generated archive is ~20mb in size.

This is possible since after #1826 we don't need to generate any caml_apply with an arity greater than 126 - so it cuts down on the number of items quite a lot

### DIFF
--- a/backend/cmm_helpers.ml
+++ b/backend/cmm_helpers.ml
@@ -3076,7 +3076,7 @@ module Generic_fns_tbl = struct
     let considered_as_small_threshold = 20
 
     let is_curry (kind, arity, result) =
-      (* For now we don't cache generic functions involving unboxed typs *)
+      (* For now we don't cache generic functions involving unboxed types *)
       if not (only_concerns_values ~arity ~result)
       then false
       else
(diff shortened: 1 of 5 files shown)


backend/cmm_helpers.ml
```diff
@@ -3051,54 +3051,69 @@ module Generic_fns_tbl = struct
...
-      then false
+      then Evict
       else
         match kind with
         | Lambda.Tupled ->
           let l = len_arity arity in
-          2 <= l && l <= considered_as_small_threshold
+          if 2 <= l && l <= considered_as_small_threshold then keep l else Evict
         | Lambda.Curried { nlocal } ->
           let l = len_arity arity in
```
Ekdohibs, line 3076: Although it does not originally come from this PR, it seems to me that this will cause a function with 2 value parameters, and say, 100 unboxed float parameters, to be considered as `Small` if at any point we start pre-generating generic functions for other layouts than `value`.

poechsel: That's right - I've made it a bit more clear that we only cache generated functions involving boxed values

mshinwell (approved):

PR #1833, opened 2023-09-18
Fix bug in `Clambda_primitives.result_layout`.

ccasin:
This bug results an assert failure in flambda1 for some uses of externals that return non-values.  I have not checked whether the other middle ends are miscompiling these programs or avoid this bug.

(Note that, obviously, `Clambda_primitives.result_layout` should not be a fragile match.  But that change is already being made in the `float#` records PR.)


mshinwell (approved):

mshinwell:
Looks fine, but please check the other middle ends too.

PR #1834, opened 2023-09-18
Improve layout history error message: add path field to Type_argument value_creation_reason

alanechang:
Currently, layout error messages related to type arguments are not very informative:

```
type t = float# list;;
[%%expect {|
Line 1, characters 9-15:
1 | type t = float# list;;
             ^^^^^^
Error: This type float# should be an instance of type ('a : value)
         it equals the primitive value type float#.
         a type argument defaults to layout value.
```

This PR aims to attach a `Path.t` field to the `Type_argument` variant of `value_creation_reason`. This field will store the parent type that the type argument is passed to.

The error above will thus be improved:

```
...
         a type argument of list defaults to layout value.
```


ocaml/testsuite/tests/typing-layouts-float64/parsing.ml
```diff
@@ -47,7 +47,7 @@ Error: This type float# should be an instance of type ('a : value)
        The layout of float# is float64, because
          it equals the primitive value type float#.
        But the layout of float# must be a sublayout of value, because
-         a type argument defaults to layout value.
+         a type argument of list defaults to layout value.
```
goldfirere, line 50: Actually, on further thought, something is off here.

`list` is imported. We don't really care, here, whether its argument's layout was defaulted or annotated. I realize this is an orthogonal change to the one you're aiming for here, but I think I'd rather this line read

```
the type argument of imported type list has layout value
```

That is, we just state that it has layout value. The "imported" part is meant to convey that this fact isn't up for debate or inference. But maybe that part is less clear.

On the other hand, the change you made would be great for

```
type 'a t
val f : 'a -> 'a t
let errors = f (Float_u.of_float 3.14)
```

where the fact that defaulting happened is relevant -- precisely because the type is local.

How to achieve this? Not quite sure. You might want a `Imported_type_argument` constructor for one of the reason types (it would apply to any layout, not just values). I know that the layouts for the arguments of `list`, `option`, and `array` all are hard-coded in `predef.ml`. You could use `Imported_type_argument` there. Harder would be sorting this out for proper imported types. The current treatment (in subst.ml) that changes all the reasons to `Imported` upon writing to a .cmi file at the end of type-checking would be hard to rewrite in this way. Perhaps better would be to do it upon *loading* a cmi file... which might just work by adding some logic to `Env.store_type` to update the reasons there. (That would also likely update the reasons for predef types, so you might be able to get away with putting dummy reasons in predef. Worth experimenting.)

ocaml/typing/ctype.ml
```diff
@@ -3828,7 +3828,7 @@ let filter_arrow env t l ~force_tpoly =
               (* CR layouts v5: Change the Layout.value when option can
                  hold non-values. *)
               (Tconstr(Predef.path_option,
-                       [newvar2 level (Layout.value ~why:Type_argument)],
+                       [newvar2 level (Layout.value ~why:(Type_argument Predef.path_option))],
```
goldfirere, line 3831: This, too, should probably be `Imported`.

goldfirere (commented):
I'm surprised no changed were necessary in typedecl. That suggests that we're not updating the reason appropriately when that module defaults.

Also, there were no tests for local types. Please add some. Including a test where the type in question is actually shadowed; this will test that the path printer is working intelligently.

ocaml/testsuite/tests/typing-layouts-float64/parsing.ml
```diff
@@ -47,7 +47,7 @@ Error: This type float# should be an instance of type ('a : value)
        The layout of float# is float64, because
          it equals the primitive value type float#.
```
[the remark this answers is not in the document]
alanechang: https://github.com/ocaml-flambda/flambda-backend/blob/0efcea4f79eb1387f9c3072545c618d3bdccd509/ocaml/typing/predef.ml#L279-L280

Is the fix here to create a new ident for `float64` and change `Primitive ident_unboxed_float` to something like `Primitive ident_float64`?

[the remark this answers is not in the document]
goldfirere: No, I think it's just to change the printing of the `Primitive` constructor of `float64_creation_reason`.

ocaml/testsuite/tests/typing-layouts-err-msg/test.ml
```diff
@@ -0,0 +1,75 @@
...
+type t_void : void
+Line 2, characters 10-16:
+2 | type t = (t_void, t_void) A.t2
+              ^^^^^^
+Error: This type t_void should be an instance of type ('a : value)
+       The layout of t_void is void, because
+         of the annotation on the declaration of the type t_void.
+       But the layout of t_void must be a sublayout of value, because
+         the type argument of imported type A.t2 has layout value.
+|}]
```
goldfirere, line 75: Would be good also to look at the errors if `A.t` is referred through different modules. Like have

```
module B = A
... B.t ...
```

to see what gets reported.

ocaml/testsuite/tests/typing-layouts-float64/parsing.ml
```diff
@@ -45,9 +45,9 @@ Line 1, characters 9-15:
              ^^^^^^
 Error: This type float# should be an instance of type ('a : value)
        The layout of float# is float64, because
-         it equals the primitive value type float#.
+         it equals the primitive float64 type float#.
```
goldfirere, line 48: Upon reading this again, this message is a little silly. Of course `float# = float#`. Maybe

```
  it is a primitive float64 type.
```

Would be better. We'd have to make sure that

```
type myfloat = float#
```

would get the old message, though (with "equals").

I think this might be possible by comparing the path against the path for `float#`, as exported from predef.

ocaml/testsuite/tests/typing-layouts/parsing_alpha.compilers.reference
```diff
@@ -7,7 +7,7 @@ Error: This type ('a : value) should be an instance of type ('a0 : void)
        The layout of 'a is void, because
          of the annotation on 'a in the declaration of the type t0.
        But the layout of 'a must overlap with value, because
-         a type argument defaults to layout value.
+         the type argument of imported type list has layout value.
```
goldfirere, line 10: Actually, after reading all these messages, I think we can drop "imported". And "type". So this would become

```suggestion
         the type argument of list has layout value.
```

Shorter. Simpler.

ocaml/typing/ctype.ml
```diff
@@ -3828,7 +3828,8 @@ let filter_arrow env t l ~force_tpoly =
               (* CR layouts v5: Change the Layout.value when option can
                  hold non-values. *)
               (Tconstr(Predef.path_option,
-                       [newvar2 level (Layout.value ~why:Type_argument)],
+                       [newvar2 level (Layout.value ~why:(Imported_type_argument
```
goldfirere, line 3831: This layout has to be used a bunch. Probably best to export it from predef, maybe as `option_argument_layout`. That will also make it easy to revise when we allow `type ('a : any) option`, which won't be all that far in the future.

ocaml/typing/layouts.ml
```diff
@@ -744,7 +752,12 @@ module Layout = struct
       | Extensible_variant -> fprintf ppf "it's an extensible variant"
       | Primitive id ->
         fprintf ppf "it equals the primitive value type %s" (Ident.name id)
-      | Type_argument -> fprintf ppf "a type argument defaults to layout value"
+      | Type_argument parent_path ->
+        fprintf ppf "the type argument of %a has layout value"
+          !printtyp_path parent_path
+      | Imported_type_argument parent_path ->
```
goldfirere, line 758: Interesting. My simplification above suggests we should recombine `Type_argument` and `Imported_type_argument`. I can't think of a good reason why not, so let's just do that.

ocaml/typing/typetexp.ml
```diff
@@ -677,6 +677,20 @@ and transl_type_aux env policy mode styp =
       in
       List.iter2
         (fun (sty, cty) ty' ->
+           let get_reason layout = match Layout.get layout with
+             | Const Value -> Layout.(Value_creation (Imported_type_argument path))
+             (* CR layouts: Add more cases here when type params of imported types can
```
goldfirere, line 682: This is already true: we can import definitions with non-value layout arguments.

But I think this doesn't need to match on the layout at all. Instead of putting `Imported_type_argument` into `value_creation_reason`, put it elsewhere to avoid the need to worry about which layout the argument has.

ocaml/typing/typetexp.ml
```diff
@@ -677,6 +677,20 @@ and transl_type_aux env policy mode styp =
...
+                have non-value layout *)
+             | _ -> Layout.Imported
+           in
+           begin match Types.get_desc ty' with
+           | Tvar {layout; _} when Layout.has_imported_history layout ->
+             (* In case of a Tvar with imported layout history, we can improve
+                the layout reason using the in scope [path] to the parent type. *)
+             let reason = get_reason layout in
+             Types.set_var_layout ty' (Layout.update_reason layout reason)
```
goldfirere, line 691: We've been worried about performance here. If you've done some experiments that suggest why performance is OK, summarize the results in a comment near here.

Ekdohibs (approved):
For changes to `chamelon/`.

ocaml/testsuite/tests/typing-layouts-err-msg/test.ml
```diff
@@ -0,0 +1,75 @@
...
+type t_void : void
+Line 2, characters 36-37:
+2 | let f (x : t_void): ('a, 'b) A.t2 = x
+                                        ^
+Error: This expression has type t_void but an expression was expected of type
+         ('a, 'b) A.t2 = ('a : value)
+       The layout of t_void is void, because
+         of the annotation on the declaration of the type t_void.
+       But the layout of t_void must be a sublayout of value, because
+         the type argument of imported type A.t2 has layout value.
```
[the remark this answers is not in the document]
alanechang: Implemented this by tracking the arity and argument position within Type_argument

ocaml/typing/typetexp.ml
alanechang: Added `Imported_type_argument` as just a `creation_reason` to eliminate the need for the match.

This makes `Imported_type_argument` and `Type_argument` different again (one is a `creation_reason` and the other is a `value_creation_reason`).

ocaml/testsuite/tests/typing-layouts-float64/parsing.ml
alanechang: This seems a little difficult to implement. The path stored in the `Primitive` creation_reason is going to be `ident_unboxed_float` even in the case of `type myfloat = float#`. To enable this path check, `myfloat` will have to be threaded through all the layout history format functions to get to `format_float64_creation_reason`. There will also be a circular dependence between Predef and Layouts if we want the path of `float#` from Predef.

Not sure if all that will be worth it?

ocaml/testsuite/tests/typing-layouts-err-msg/test.ml
```diff
@@ -0,0 +1,180 @@
...
+
+[%%expect{|
+Line 1, characters 37-43:
+1 | type t = (t_value, t_value, t_value, t_void, t_void) A.t5
+                                         ^^^^^^
+Error: This type t_void should be an instance of type ('a : value)
+       The layout of t_void is void, because
+         of the annotation on the declaration of the type t_void.
+       But the layout of t_void must be a sublayout of value, because
+         the type argument at position 4 of A.t5 has this layout.
```
goldfirere, line 141: If you want to use numbers instead of words at this point, how about

```suggestion
         the 4th type argument of A.t5 has this layout.
```

ocaml/testsuite/tests/typing-layouts-err-msg/test.ml
```diff
@@ -0,0 +1,180 @@
...
+[%%expect{|
+Line 1, characters 26-27:
+1 | let f2 (x: t_void) = A.f2 x
+                              ^
+Error: This expression has type t_void but an expression was expected of type
+         ('a : value)
+       The layout of t_void is void, because
+         of the annotation on the declaration of the type t_void.
+       But the layout of t_void must be a sublayout of value, because
+         it's imported from another compilation unit.
```
goldfirere, line 179: This still isn't great, but we can return to it later.

ocaml/testsuite/tests/typing-layouts-float64/parsing.ml
goldfirere: Fine. Maybe just change `equals` to `is` and then we'll call it good enough. For `float#` the error message will b slightly pedantic. For synonyms, it will be useful.

ocaml/typing/layouts.ml
```diff
@@ -806,6 +823,15 @@ module Layout = struct
          format_concrete_layout_reason ppf concrete
       | Imported ->
          fprintf ppf "it's imported from another compilation unit"
+      | Imported_type_argument {parent_path; position; arity} ->
+        (match arity, position with
+         | 1, _ -> fprintf ppf "the type argument of %a has this layout"
```
goldfirere, line 828: Definitely worth making a function to abstract over this pattern. Happily, I think the function can just retun the string between "the" and "type argument". In the case of arity 1, that string will be empty.

ocaml/typing/layouts.ml
```diff
@@ -262,7 +262,7 @@ module Layout = struct
     | Boxed_variant
     | Extensible_variant
     | Primitive of Ident.t
-    | Type_argument
+    | Type_argument of {parent_path: Path.t; position: int; arity: int}
```
goldfirere, line 265: Worth clarifying that `position` is 1-indexed.

ocaml/typing/predef.mli
```diff
@@ -84,6 +84,8 @@ val ident_cons : Ident.t
 val ident_none : Ident.t
 val ident_some : Ident.t
 
+val option_argument_layout : Layouts.Layout.t
```
goldfirere, line 87: Please put a little comment describing what this is.

ocaml/typing/typecore.ml
```diff
@@ -7992,7 +7998,7 @@ and type_comprehension_expr
         Predef.type_list,
         (fun tcomp -> Texp_list_comprehension tcomp),
         comp,
-        Layout.Type_argument
+        (Layout.Type_argument {parent_path=Predef.path_list;position=1;arity=1})
```
goldfirere, line 8001: Just barely, I think it's worth making a `list_argument_layout` in parallel with the one for option. It would be used here and in predef.

goldfirere (approved):
Just a few small tweaks. I've Approved. Feel free to merge when you've addressed the concerns.

PR #1835, opened 2023-09-18
Adjust the location on `as` pattern vars for better errors/warnings

ccasin:
This program has an unused pattern variable:
```ocaml
type t =
  { x : int
  ; y : int
  }

let sum ({ x; y } as t) = x + y
```
Previously, it gave this warning:
```
File "/home/ccasinghino/tmp/milo.ml", line 6, characters 8-23:
6 | let sum ({ x; y } as t) = x + y
            ^^^^^^^^^^^^^^^
Warning 26 [unused-var]: unused variable t.
```
But the location here is unhelpful, and it's worse if the pattern is bigger.

After this PR, we instead give:
```
File "/home/ccasinghino/tmp/milo.ml", line 6, characters 21-22:
6 | let sum ({ x; y } as t) = x + y
                         ^
Warning 26 [unused-var]: unused variable t.
```
Better!

This location shows up once in the testsuite for a different error. I think it's an improvement there too.

### DIFF
--- a/ocaml/testsuite/tests/typing-gadts/or_patterns.ml
+++ b/ocaml/testsuite/tests/typing-gadts/or_patterns.ml
@@ -217,9 +217,9 @@ let simple_merged_annotated_return (type a) (t : a t) (a : a) =
 ;;
 
 [%%expect{|
-Line 3, characters 12-20:
+Line 3, characters 18-19:
 3 |   | IntLit, (3 as x)
-                ^^^^^^^^
+                      ^
 Error: This pattern matches values of type a
        This instance of a is ambiguous:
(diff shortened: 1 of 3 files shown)


ncik-roberts (approved):
This seems like a clear improvement. Should it be upstreamed? (If you think it should but don't want to post it upstream yet for some reason, would you mind tagging it as `to-upsteam`?)

ccasin:
> This seems like a clear improvement. Should it be upstreamed? (If you think it should but don't want to post it upstream yet for some reason, would you mind tagging it as `to-upsteam`?)

Yep - will upstream, and should have tagged it.  Thanks!

PR #1841, opened 2023-09-20
Add `-as-argument-for` parameter

lukemaurer:
**Depends on #1764.**

Add the ability to compile a module declaring that it is suitable to pass as the argument to a module taking a given parameter. Essentially the module is now checked against two different signatures (its own .mli and the parameter, which of course must be a supertype).


lukemaurer:
This probably needs reworking. The implementation currently turns every argument module into a two-field record, the first field of which is the usual module structure and the second field of which is the record with the exact type for the module. This isn't great: everything that accesses a module needs to go through an indirection. That's not as bad as it sounds because (a) we optimize and (b) the mapping of `Address.t` to Lambda code is defined in one place, but there's still enough annoying complexity that I'm inclined to revisit it.

Mark suggested eons ago that we could take the coercion from the full module to its argument type and write the Closure code for it to a special field of the record (we'd just need to write a lambda that applies `Translmod.apply_coercion` to its argument). All that's needed then is a way to tell other modules where that field is. (Having a secret extra field already happens for objects so I assume that's doable.)

PR #1844, opened 2023-09-20
Fix mutable blocks being assumed as escaping if inside an unused Make_block

Ekdohibs:
It was the only primitive affected, as it was both tracked for mutable unboxing and had arguments that escaped if and only it the result was used.


Gbury (approved):

mshinwell (approved):

mshinwell:
Example that was failing:
```
type t = { mutable a : int; }

let[@inline] bar () =
  let t = { a = 42 } in
  t, t.a

let foo () =
  let _t, x = bar () in
  x
```

PR #1848, opened 2023-09-20
Expose `Pprintast.tyvar`

ccasin:
Carl requested this, and the function [is already exposed upstream](https://github.com/ocaml/ocaml/blob/8e303850d904d749d505cc0a8d834ee3527a6e41/parsing/pprintast.mli#L58)

Review: @antalsz


antalsz (approved):
LGTM

PR #1849, opened 2023-09-20
Backport of upstream -H implementation

ccasin:
~~(This is a draft because it will be reviewed upstream first.  Backporting it for internal testing in the meantime.)~~

This is ready for review.  It has already been reviewed and merged [upstream](https://github.com/ocaml/ocaml/pull/12246), so it does not need a ton of scrutiny.  I have integrated all the feedback from that review.  Additionally, a version of this based on 4.14.1 has been tested with a version of our internal build system modified to use `-H`, and works fine.

@lpw25 Could you take a quick look at the files not under `ocaml/`, which weren't covered by the upstream review (there are just a few lines of diff) and then approve.


ccasin:
This is now ready to be merged.  But it will require changes in the tree, so I'm waiting for the OCaml 5 minus work to cool down to avoid making that harder.

PR #1851, opened 2023-09-21
Simplify integer comparisons that use "compare"

lthls:
This PR provides support for simplifying the common pattern `(compare x y) op 0` (where `op` is a comparison operator) into `x op y`. It is only implemented for the integer types as it is incorrect on floats (for non-regular values) and not obviously correct on non-numerical types.

I have only tested it on a simple example, other examples that should be simplified are welcome.

### DIFF
--- a/middle_end/flambda2/simplify/comparison_result.ml
+++ b/middle_end/flambda2/simplify/comparison_result.ml
@@ -14,58 +14,65 @@
 
 module P = Flambda_primitive
 
-type comparison_kind =
-  | Int of { kind : Flambda_kind.Standard_int.t; sign : P.signed_or_unsigned }
-  | Float
+type tagged_or_untagged =
+  | Tagged
+  | Untagged
 
 type t =
-{ lhs : Simple.t;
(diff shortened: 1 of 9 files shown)


mshinwell:
I'll review this.

lthls:
I've run the tests provided in ocaml/ocaml#12181 and they pass.

I've also tested the following function:
```ocaml
let rec split (x : int) l =
  match l with
  | [] -> [], [], []
  | y :: l ->
    let lt, eq, gt = split x l in
    let c = Stdlib.compare y x in
    if c < 0 then y :: lt, eq, gt
    else if c > 0 then lt, eq, y :: gt
    else lt, y :: eq, gt
```
With this PR, the call to `compare` disappears and the two comparisons with 0 are replaced by comparisons on `x` and `y`. In an ideal world where we have a three-way comparison operator exposed, this might not be optimal (as we could compare once and just check the relevant flags later), but since we don't have that then I think it's a good idea to perform these transformations.

mshinwell (approved):

PR #1853, opened 2023-09-21
Ask user to add `exclave_` instead of `local_`

riaqn:
Since we plan to gradually transition from tail-poition `local_` to `exclave_`, we should also change the error message to reflect that transition. 


ncik-roberts (approved):
Looks good to me, thanks for the quick turnaround.

PR #1855, opened 2023-09-22
Improve layout history error message: additional info on value layout defaulting

alanechang:
When a concrete layout defaults to value layout, the error message can be a bit confusing. This PR adds an additional message on these cases to explicitly highlight that the value layout is the result of defaulting.

### DIFF
--- a/ocaml/testsuite/tests/typing-layouts-float64/basics_alpha.ml
+++ b/ocaml/testsuite/tests/typing-layouts-float64/basics_alpha.ml
@@ -405,7 +405,7 @@ Error: This expression has type t_float64
        The layout of t_float64 is float64, because
          of the annotation on the declaration of the type t_float64.
        But the layout of t_float64 must be a sublayout of value, because
-         it's used as a function result.
+         it's used as a function result (defaulted to layout value).
 |}];;
 
 let x8_2 = id_value (make_t_float64_id ());;
@@ -418,7 +418,7 @@ Error: This expression has type 'a t_float64_id = ('a : float64)
        The layout of 'a t_float64_id is float64, because
          of the annotation on 'a in the declaration of the type t_float64_id.
        But the layout of 'a t_float64_id must overlap with value, because
-         it's used as a function result.
+         it's used as a function result (defaulted to layout value).
 |}];;
 
 let x8_3 = id_value (make_floatu ());;
@@ -431,7 +431,7 @@ Error: This expression has type float# but an expression was expected of type
        The layout of float# is float64, because
          it is the primitive float64 type float#.
        But the layout of float# must be a sublayout of value, because
-         it's used as a function result.
+         it's used as a function result (defaulted to layout value).
 |}];;
 
 (*************************************)
--- a/ocaml/testsuite/tests/typing-layouts-float64/basics_alpha.ml
+++ b/ocaml/testsuite/tests/typing-layouts-float64/basics_alpha.ml
@@ -405,7 +405,7 @@ Error: This expression has type t_float64
        The layout of t_float64 is float64, because
          of the annotation on the declaration of the type t_float64.
        But the layout of t_float64 must be a sublayout of value, because
-         it's used as a function result (defaulted to layout value).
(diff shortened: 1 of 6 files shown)


ocaml/testsuite/tests/typing-layouts-float64/basics_alpha.ml
```diff
@@ -405,7 +405,7 @@ Error: This expression has type t_float64
        The layout of t_float64 is float64, because
          of the annotation on the declaration of the type t_float64.
        But the layout of t_float64 must be a sublayout of value, because
-         it's used as a function result.
+         it's used as a function result (defaulted to layout value).
```
goldfirere, line 408: I guess the defaulting information is an improvement over the status quo, but this still should probably mention the function `id_value` somewhere.

goldfirere, line 408: I think I'd prefer `, defaulted to layout value` over the parenthetical.

ocaml/testsuite/tests/typing-layouts/basics_alpha.ml
```diff
@@ -1801,3 +1801,21 @@ Error: This type signature for foo33 is not a value type.
 (* Test 34: Layout clash in polymorphic record type *)
 
 (* tested elsewhere *)
+
+(****************************************************)
+(* Test 35: unannotated type parameter defaults to layout value *)
```
goldfirere, line 1806: When adding a new test to "alpha", please add stubs to "beta" and the regular test.

ocaml/testsuite/tests/typing-layouts/basics_alpha.ml
```diff
@@ -1801,3 +1801,21 @@ Error: This type signature for foo33 is not a value type.
...
+type 'a t35 = 'a
+Line 2, characters 30-31:
+2 | let f35 (x: t_void): 'a t35 = x
+                                  ^
+Error: This expression has type t_void but an expression was expected of type
+         'a t35 = ('a : value)
+       The layout of t_void is void, because
+         of the annotation on the declaration of the type t_void.
+       But the layout of t_void must be a sublayout of value, because
+         it appears as an unannotated type parameter (defaulted to layout value).
```
goldfirere, line 1820: "appears" isn't quite right. "instantiates" would work there, but maybe not in other places the error might come up? OK to fix this in a later PR.

ocaml/typing/layouts.ml
```diff
@@ -841,7 +841,12 @@ module Layout = struct
         format_desc lay;
       begin match t.history with
       | Creation reason ->
-        fprintf ppf ", because@ %a" format_creation_reason reason
+        fprintf ppf ", because@ %a" format_creation_reason reason;
+        begin match reason, lay with
+        | Concrete_creation _, Const Value ->
```
goldfirere, line 846: Why only deal with `Value` here? What about other possible sorts?

ocaml/testsuite/tests/typing-layouts/basics_alpha.ml
alanechang: "instantiates" seems to work better in all places where the error comes up. Went ahead and made the change.

ocaml/testsuite/tests/typing-layouts/basics_alpha.ml
```diff
@@ -1738,7 +1741,7 @@ Error: This expression has type t_void but an expression was expected of type
        The layout of t_void is void, because
          of the annotation on the declaration of the type t_void.
        But the layout of t_void must be a sublayout of value, because
-         it's used as a function argument.
+         it's used as a function argument, defaulted to layout value.
```
goldfirere, line 1744: I'm worried that this message might be wrong in a closely related scenario. Try changing the sig for `f` to be `val f : 'a -> ('a : value)`. Does the output still say "defaulted"? If it doesn't, then yes let's merge.

goldfirere (approved):

PR #1857, opened 2023-09-22
Add room for more runtime tags

TheNumbat:


mshinwell:
Closing this as a different approach is being used now.

PR #1861, opened 2023-09-25
Revert "Only execute caml-apply-split-max-arity on linux"

poechsel:
Reverts ocaml-flambda/flambda-backend#1837


mshinwell (approved):
(expected to break CI)

PR #1864, opened 2023-09-25
Add hint for `#float`

ccasin:
This adds a hint if you forget where the hash goes and type `#float` instead of `float#`:
```ocaml
# type t = #float;;
Error: Unbound class type float
Hint: Did you mean float#?
```
Our first user requested this.  I'm not totally sure we want it, or that this is the best way to do it, but it seems reasonable.

Review: @goldfirere


goldfirere (approved):
I think this is good, but we should improve it to handle the general case. I'll post more on the internal ticket.

PR #1867, opened 2023-09-25
Centralise warning configuration

lukemaurer:
There's a lot of warning boilerplate in our Dune files (thankfully not in our source files anymore). This makes it very easy to leave off warning flags, and in fact it turns out we've been compiling all of `backend/` and `tools/` with only default warnings on (which is to say, practically no warnings enabled).

I've chosen to keep this PR to Dune files, so I haven't fixed any warnings, but it did turn up several that should be easy to fix.

Principles at play here:

1. All `flags` clauses should include `:standard`, and none should include `-w +a` except the topmost declaration possible. (The exception is `ocaml/dune`, which has to be self-sufficient.) Also, `-principal` should be set early on and `-no-principal` used to override it.

2. All modules that already had reasonable warnings set should be compiled with exactly the same warnings enabled as before. Further simplifications are probably possible here, since there are quite a few random-looking differences that may not have been intended.

3. All modules that were being compiled with the default warnings now explicitly disable enough warnings that they currently compile. I've left comments for the warnings that we probably want to fix.

CI can't really tell us what warnings should be enabled but aren't, so reviewers please make sure I haven't left anything important out.

### DIFF
--- a/backend/asm_targets/dune
+++ b/backend/asm_targets/dune
@@ -3,6 +3,7 @@
 (library
   (name asm_targets)
   (wrapped true)
-  (flags (:standard -principal))
+  ; FIXME Fix warning 27
+  (flags (:standard -w -27-70))
   (libraries ocamlcommon dwarf_flags)
 )
--- a/backend/debug/dwarf/dwarf_flags/dune
+++ b/backend/debug/dwarf/dwarf_flags/dune
@@ -3,7 +3,6 @@
 (library
   (name dwarf_flags)
   (wrapped true)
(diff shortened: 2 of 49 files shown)


mshinwell:
@dkalinichenko-js could you maybe look at this?

dkalinichenko-js:
I'll look at it.

dkalinichenko-js:
I've rebased this PR on the recent main branch here: https://github.com/ocaml-flambda/flambda-backend/pull/2617.

PR #1868, opened 2023-09-26
Use a forest rather than a tree for dominators

xclerc:
Because some CFG blocks may be unreachable,
it is not always possible to build a dominator tree.
Since there are possibly several roots, this pull
request changes the representation to a forest.

(When a function would induce several roots, the
build would fail with an `assert false` being
raised by `Cfg_dominators.compute_doms`.)

### DIFF
--- a/backend/cfg/cfg_dominators.ml
+++ b/backend/cfg/cfg_dominators.ml
@@ -48,15 +48,12 @@ let invariant_doms : Cfg.t -> doms -> unit =
     cfg.blocks;
   (* Check that (i) the immediate dominator is a strict dominator, and (ii)
      there is no other intermediate strict dominator - except for the entry
-     block. *)
+     block and unreachable blocks, which all have (by convention) themselves as
+     immediate dominators. *)
   Label.Tbl.iter
     (fun n idom_n ->
-      if Label.equal n cfg.entry_label
+      if not (Label.equal n idom_n)
       then (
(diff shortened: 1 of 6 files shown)


backend/cfg/cfg_dominators.ml
```diff
@@ -48,15 +48,12 @@ let invariant_doms : Cfg.t -> doms -> unit =
     cfg.blocks;
   (* Check that (i) the immediate dominator is a strict dominator, and (ii)
      there is no other intermediate strict dominator - except for the entry
-     block. *)
+     block and unreachable blocks, which all have (by convention) themselves as
+     immediate dominators. *)
   Label.Tbl.iter
     (fun n idom_n ->
-      if Label.equal n cfg.entry_label
+      if not (Label.equal n idom_n)
```
gretay-js, line 55: Have you considered strengthing this condition to check that `n` has no predecessors?

gretay-js (approved):

backend/cfg/cfg_dominators.ml
xclerc: I expect the condition you suggest
to be true as of today, but I wonder
whether it is guaranteed to hold.
(I am thinking about an isolated loop.)

gretay-js: I am not sure if the dominator and dominator tree computation works with isolated loops, I don't think we can have them.

PR #1365, opened 2023-05-08
Introduce an API to swap the runtime lock for a different lock.

stedolan:
This patch adds a new API to the systhreads module, allowing the runtime lock (i.e. the lock released during blocking sections) to be swapped out for a different lock with a custom implementation.

The motivation is to allow efficient interop with the global interpreter lock of Python (by unifying the OCaml and Python locks, so no synchronisation need be done when moving from one language to another). However, there is nothing Python-specific in the API created here.

PR #1869, opened 2023-09-26
Replace thread save/restore API with switch

stedolan:
This is an optimisation to the API introduced by #1365 : instead of providing separate save/restore operations to be performed before releasing and after re-acquiring the lock, we provide a single "switch" operation to be performed after re-acquiring the lock, which both saves the old and restores the new thread.

The point of this is that we can optimise the case where the old and new threads are the same. This is particularly common in Python, which has to do a lot of save/restore pairs for safety. (I don't think this change particularly helps or harms anything else)


lpw25 (approved):
LGTM

mshinwell:
This contains a write-after-free bug if a thread is stopped; `curr_thread` may then be invalid.

PR #1870, opened 2023-09-26
Improve debuginfo for for-loops (and conditionals)

mshinwell:
This started out as an attempt to improve the debuginfo on `for`-loops in flambda2, which currently don't have any location information, and about which there have (quite reasonably) been complaints.

For example in the following fragment, there is no debuginfo on either the conditional immediately prior to the first iteration or on the conditional at each iteration!  This means that, for example, this latter conditional will end up adopting the last debuginfo that happens to come from the body of the loop.
```
let[@inline] poke buf ~pos c = Bigarray.Array1.unsafe_set buf pos c

let fill buf len c =
  for i = 0 to len do
    poke buf ~pos:i c
  done
```

This patch fixes this.  In order to do that, I've had to fix many places in the backend where debuginfo was just getting dropped, although the changes are almost all individually trivial.  In addition I added some extra debug info propagation for spills and reloads.  So in fact, this should pretty much improve debuginfo across the board.  It is only intended to work with flambda2.

This also fixes what appears to be a bug where we were failing to add the inlined debuginfo during `To_cmm`'s processing of `Switch` expressions.

Linearized code after this PR for the above example:
```
camlFor_loc__fill_1_3_code: {for_loc.ml:3,9-71}
  prologue
  buf/45[%rax] := R/0[%rax]
  len/46[%rbx] := R/1[%rbx]
  c/47[%rdi] := R/2[%rdi]
  if len/46[%rbx] <s 1 goto L107 {for_loc.ml:4,2-50}   <-- important debuginfo
  spilled-c/58[s[i:1]] := c/47[%rdi] (spill)
  spilled-len/56[s[i:3]] := len/46[%rbx] (spill)
  spilled-buf/59[s[i:0]] := buf/45[%rax] (spill)
  I/50[%rsi] := 1
  I/51[%rsi] := I/50[%rsi]
  i/49[%rsi] := I/51[%rsi]
  spilled-i/57[s[i:2]] := i/49[%rsi] (spill)
  L108:
  buf/60[%rdi] := spilled-buf/59[s[i:0]] (reload)
  R/2[%rdi] := buf/60[%rdi]
  R/3[%rsi] := i/49[%rsi]
  c/61[%rdx] := spilled-c/58[s[i:1]] (reload)
  R/4[%rdx] := c/61[%rdx]
  {spilled-len/56[s[i:3]] spilled-i/57[s[i:2]] spilled-c/58[s[i:1]]* spilled-buf/59[s[i:0]]*}
  R/0[%rax] := extcall "caml_ba_set_1" R/2[%rdi] R/3[%rsi] R/4[%rdx] {for_loc.ml:5,4-21;for_loc.ml:1,31-67}
  i/49[%rsi] := spilled-i/57[s[i:2]] (reload) {for_loc.ml:4,2-50}   <-- important debuginfo
  len/63[%rax] := spilled-len/56[s[i:3]] (reload) {for_loc.ml:4,2-50}
  if i/49[%rsi] ==s len/63[%rax] goto L112 {for_loc.ml:4,2-50}
  I/54[%rsi] := i/49[%rsi]
  I/54[%rsi] := I/54[%rsi] + 2 {for_loc.ml:4,2-50}
  I/55[%rsi] := I/54[%rsi]
  i/49[%rsi] := I/55[%rsi]
  spilled-i/57[s[i:2]] := i/49[%rsi] (spill)
  goto L108
  L112:
  I/53[%rax] := 1
  R/0[%rax] := I/53[%rax]
  reload retaddr
  return R/0[%rax]
  L107:
  I/48[%rax] := 1
  R/0[%rax] := I/48[%rax]
  reload retaddr
  return R/0[%rax]
```

### DIFF
--- a/middle_end/flambda2/from_lambda/closure_conversion_aux.mli
+++ b/middle_end/flambda2/from_lambda/closure_conversion_aux.mli
@@ -74,8 +74,11 @@ module IR : sig
 
   type switch =
     { numconsts : int;
-      consts : (int * Continuation.t * trap_action option * simple list) list;
-      failaction : (Continuation.t * trap_action option * simple list) option
+      consts :
+        (int * Continuation.t * Debuginfo.t * trap_action option * simple list)
+        list;
+      failaction :
+        (Continuation.t * Debuginfo.t * trap_action option * simple list) option
     }
 
   val print_simple : Format.formatter -> simple -> unit
--- a/middle_end/flambda2/from_lambda/closure_conversion_aux.mli
+++ b/middle_end/flambda2/from_lambda/closure_conversion_aux.mli
@@ -74,6 +74,7 @@ module IR : sig
 
   type switch =
     { numconsts : int;
+      (* CR mshinwell: use record types *)
       consts :
         (int * Continuation.t * Debuginfo.t * trap_action option * simple list)
         list;
--- a/middle_end/flambda2/from_lambda/lambda_to_flambda.ml
+++ b/middle_end/flambda2/from_lambda/lambda_to_flambda.ml
@@ -167,6 +167,39 @@ let compile_staticfail acc env ccenv ~(continuation : Continuation.t) ~args :
         acc ccenv)
     acc ccenv
 
+let rec try_to_find_location (lam : L.lambda) =
+  (* This is very much best-effort and may overshoot, but will still likely be
+     better than nothing. *)
+  match lam with
+  | Lprim (_, _, loc)
+  | Lfunction { loc; _ }
+  | Lapply { ap_loc = loc; _ }
+  | Lfor { for_loc = loc; _ }
+  | Lswitch (_, _, loc, _)
+  | Lstringswitch (_, _, _, loc, _)
+  | Lsend (_, _, _, _, _, _, loc, _)
+  | Levent (_, { lev_loc = loc; _ }) ->
+    loc
+  | Llet (_, _, _, lam, _)
+  | Lmutlet (_, _, lam, _)
+  | Lletrec ((_, lam) :: _, _)
+  | Lifthenelse (lam, _, _, _)
+  | Lstaticcatch (lam, _, _, _)
+  | Lstaticraise (_, lam :: _)
+  | Lwhile { wh_cond = lam; _ }
+  | Lsequence (lam, _)
+  | Lassign (_, lam)
+  | Lifused (_, lam)
+  | Lregion (lam, _)
+  | Lexclave lam
+  | Ltrywith (lam, _, _, _) ->
+    try_to_find_location lam
+  | Lvar _ | Lmutvar _ | Lconst _ | Lletrec _ | Lstaticraise (_, []) ->
+    Debuginfo.Scoped_location.Loc_unknown
+
+let try_to_find_debuginfo lam =
+  Debuginfo.from_location (try_to_find_location lam)
+
 let switch_for_if_then_else ~cond ~ifso ~ifnot ~kind =
   let switch : Lambda.lambda_switch =
     { sw_numconsts = 2;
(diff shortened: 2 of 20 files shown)


gretay-js (approved):
For backend only.  Do we need to fix the CFG pipeline too? 

I also wonder if this PR addresses any [other](https://github.com/ocaml-flambda/flambda-backend/blob/95b514abbf83ce01dc06808e286d3b761e8f3a90/middle_end/flambda2/to_cmm/to_cmm_expr.ml#L689) [CRs](https://github.com/ocaml-flambda/flambda-backend/blob/95b514abbf83ce01dc06808e286d3b761e8f3a90/middle_end/flambda2/to_cmm/to_cmm_expr.ml#L435-L439) about `Debuginfo.none`.

Gbury (changes requested):
This seems (at least for the parts I can review) good except for my comment on the if/then/else debug which would at least deserve a comment with an explanation of why it is correct/desirable to copy debug info between conditional branches.

middle_end/flambda2/to_cmm/to_cmm_expr.ml
```diff
@@ -945,6 +949,13 @@ and switch env res switch =
           (Backend_var.Set.union else_free_vars then_free_vars)
       in
       let cmm, free_vars =
+        let then_dbg, else_dbg =
+          if Debuginfo.is_none then_dbg
+          then else_dbg, else_dbg
+          else if Debuginfo.is_none else_dbg
+          then then_dbg, then_dbg
+          else then_dbg, else_dbg
```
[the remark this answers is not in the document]
mshinwell: Ah, I was experimenting with this, but I don't think it needs to be in this PR.  Have removed it.

mshinwell:
@gretay-js I don't think any CFG fixes are needed for the moment, but there may be some needed later along the lines of the ones I used to have in `Linearize` (to make sure that any inserted branches get the correct debug info).

middle_end/flambda2/from_lambda/lambda_to_flambda.ml
```diff
@@ -167,6 +167,39 @@ let compile_staticfail acc env ccenv ~(continuation : Continuation.t) ~args :
...
+  | Lwhile { wh_cond = lam; _ }
+  | Lsequence (lam, _)
+  | Lassign (_, lam)
+  | Lifused (_, lam)
+  | Lregion (lam, _)
+  | Lexclave lam
+  | Ltrywith (lam, _, _, _) ->
+    try_to_find_location lam
+  | Lvar _ | Lmutvar _ | Lconst _ | Lletrec _ | Lstaticraise (_, []) ->
+    Debuginfo.Scoped_location.Loc_unknown
```
Gbury, line 198: out of curiosity, did you try to change this to an exception/error, just to see if this case is hit often or not ?

middle_end/flambda2/from_lambda/closure_conversion_aux.mli
```diff
@@ -74,8 +74,11 @@ module IR : sig
 
   type switch =
     { numconsts : int;
-      consts : (int * Continuation.t * trap_action option * simple list) list;
-      failaction : (Continuation.t * trap_action option * simple list) option
+      consts :
+        (int * Continuation.t * Debuginfo.t * trap_action option * simple list)
+        list;
+      failaction :
+        (Continuation.t * Debuginfo.t * trap_action option * simple list) option
```
Gbury, line 81: at one point these tuples will deserve dedicated record types, maybe a CR is enough for now

Gbury (approved):

middle_end/flambda2/from_lambda/lambda_to_flambda.ml
mshinwell: I did not.

middle_end/flambda2/from_lambda/closure_conversion_aux.mli
mshinwell: I've added a CR.

PR #1873, opened 2023-09-27
Syntactic support for instance names as identifiers

lukemaurer:
**Depends on #1872.**

Currently the only support is as a handwritten extension like

```
Module(Name1)(Value1)(Name2)(Value2) [@jane.non_erasable.instances]
```

which isn't great, but initially we plan to support creating and referring to instances only using Dune, so this syntax will only see the light of day in source files generated by Dune and not meant for human consumption.


ocaml/parsing/jane_syntax.mli
```diff
@@ -276,6 +276,19 @@ module Layouts : sig
     (Jkind.annotation * Parsetree.attributes) option
 end
 
+module Instances : sig
+  (** The name of an instance module. Gets converted to [Global.Name.t] in the
+      flambda-backend compiler. *)
+  type instance =
+    { head : string;
+      args : (instance * instance) list
```
riaqn, line 284: Can we try to distinguish the parameter and value in `instance * instance`?

In particular, the parameter cannot have arguments.

ocaml/utils/compilation_unit.ml
```diff
@@ -408,7 +408,7 @@ end = struct
             })
           arguments
       in
-      of_full (Global (Global_module.Name.create head arguments))
+      of_full (Global (Global_module.Name.create_exn head arguments))
```
riaqn, line 411: Should this handle errors properly? Are we assuming that currently only the build system can generate the syntax, and it's their responsibility to do it correctly?

ocaml/parsing/ast_iterator.ml
```diff
@@ -403,9 +404,24 @@ end
 module M = struct
   (* Value expressions for the module language *)
 
-  let iter sub {pmod_loc = loc; pmod_desc = desc; pmod_attributes = attrs} =
+  module I = Jane_syntax.Instances
+
+  let iter_instance _sub : I.instance -> _ = function
+    | _ -> () (* Currently treating these as atomic? *)
```
riaqn, line 410: I imagine in the future we will want to iterate through it. Can you add a CR?

ocaml/parsing/ast_mapper.ml
```diff
@@ -479,12 +481,30 @@ end
 
 
 module M = struct
+  module I = Jane_syntax.Instances
+
   (* Value expressions for the module language *)
+  let map_instance _sub : I.instance -> I.instance = function
+    | i -> i (* Currently treating these as atomic? *)
```
riaqn, line 488: Similarly, please make a CR. :-)

ocaml/parsing/pprintast.ml
```diff
@@ -1637,7 +1637,9 @@ and module_expr ctxt f x =
   if x.pmod_attributes <> [] then
     pp f "((%a)%a)" (module_expr ctxt) {x with pmod_attributes=[]}
       (attributes ctxt) x.pmod_attributes
-  else match x.pmod_desc with
+  else match Jane_syntax.Module_expr.of_ast x with
+    | Some ext -> extension_module_expr ctxt f ext
+    | None -> match x.pmod_desc with
```
riaqn, line 1642: Could you add a test to `pprintast_unconditional.ml`? In particular, the print should contain `[@jane.non_erasable.instancee]` because pprintast is supposed to roundtrip with parser.

ocaml/typing/global_module.mli
```diff
@@ -1,18 +1,27 @@
...
-  and argument = {
-    param : t;
-    value : t;
-  }
+  and argument = (t, t) Argument.t
 
   include Identifiable.S with type t := t
 
-  val create : string -> argument list -> t
+  val create : string -> argument list -> (t, (t, t) duplicate) Result.t
```
riaqn, line 22: It seems that we either trust the build system to generate well-formed instances, or `args` is empty. Then we don't need to expose this function?

ocaml/typing/typemod.ml
```diff
@@ -102,6 +102,7 @@ type error =
       old_arg_type : Global_module.Name.t option;
       old_source_file : Misc.filepath;
     }
+  | Duplicate_parameter_name of Global_module.Name.t
```
riaqn, line 105: This error is funny: it's only used by a `Misc.fatal_errorf`. This PR seems to assume that the build system won't generate wrong instances (as the duplication check is not complete anyway). Therefore, we don't need this error.

riaqn (approved):
LGTM - some small comments.

We will move the syntax to `parsetree` in a future PR.

ocaml/utils/compilation_unit.ml
lukemaurer: This is the general constructor for `Compilation_unit.t`s, so any problem with the arguments is going to be an internal error rather than bad user input. User input gets turned into a `Global_module.Name.t` to begin with, then goes through `of_global_name` rather than this code.

ocaml/typing/global_module.mli
lukemaurer: I'd rather not bake in the assumption that the build system is writing these. It's still source code and CLI arguments, which we generally treat with nice error messages. (We do have the awful syntax on the assumption that the build system is writing it, but that's much easier to change later.)

The no-argument case is frequent enough to make a separate function, though. Adding `create_no_args : string -> t`. (I might have called it `of_string` but it's not actually the inverse of `to_string`.)

ocaml/typing/typemod.ml
lukemaurer: That's only true because the code that calls it is a stub for unimplemented code. The next PR will have real code in `type_module_extension_aux` and it should produce type errors as usual. Not sure what you mean about the duplication check not being strict enough? I can add the check to `Global_module.create` if that would help.

lukemaurer: (decided to add the check to `Global_module.create`)

ocaml/typing/global_module.mli
riaqn: I like `create_no_args`. Resolving.

ocaml/typing/typemod.ml
riaqn: Thank you for making the check complete.

I think if you rewrite `type_module_extension_aux` to:
```
      ignore (alias, sttn, env, smod);
     let modname = instance_name ~loc:smod.pmod_loc env glob in
      Misc.fatal_errorf "@[<hv>Unimplemented: instance identifier@ %a@]"
        Global_module.Name.print modname
```
it would be more clearer. That is, we want to separate "instance names are checked" and "the proper type checking hasn't been implemented". Previously I was confused because the two are together.

PR #1875, opened 2023-09-27
Rename "layout" to "kynd"

goldfirere:
In the brave new modal kinds world, types are classified by *kinds*, not *layouts*. A layout is, instead, just one component of a kind.

The word kind is overloaded in the source. It's already in use in the type-checker, as part of a type declaration. To avoid confusion (and for easy recognition when editing code), this patch does not use `kind` in the source, but instead uses `kynd`. Perhaps one day we will convince upstream to rename type declaration kinds to something else, and then we can replace `kynd` with `kind`. But until that day, `kynd` it is.

(There was a discussion some time ago about this, and `kynd` won out over `kkind`, which was my original attempt.)

Two other changes in this PR:
* I restructured (slightly) the module structure in kynd.ml (originally layouts.ml).
* I clarified an error message by a tiny bit as I swooped by.


goldfirere:
Changed to use jkind, by popular request

ocaml/lambda/matching.ml
```diff
@@ -100,25 +99,25 @@ open Printpat
 module Scoped_location = Debuginfo.Scoped_location
 
 type error =
-    Non_value_layout of Layout.Violation.t
-  | Illegal_record_field of Layout.const
+    Non_value_jkind of Jkind.Violation.t
```
ccasin, line 102: I'm surprised to see this error change.  I think the error is really about layouts.  We might add other similar errors in the future for the other parts of `Jkind`, if there are particularly experimental ones, but I don't think this would grow to encompass them.

ocaml/lambda/matching.ml
```diff
@@ -100,25 +99,25 @@ open Printpat
...
 
 exception Error of Location.t * error
 
 let dbg = false
 
 (* CR layouts v5: When we're ready to allow non-values, these can be deleted or
    changed to check for void. *)
-let layout_must_be_value loc layout =
-  match Layout.(sub layout (value ~why:V1_safety_check)) with
+let jkind_must_be_value loc jkind =
```
ccasin, line 111: Similarly I think this check is really about layouts.

ocaml/lambda/translcore.ml
```diff
@@ -30,8 +29,8 @@ open Debuginfo.Scoped_location
 type error =
     Free_super_var
   | Unreachable_reached
-  | Bad_probe_layout of Ident.t
-  | Illegal_record_field of Sort.const
+  | Bad_probe_jkind of Ident.t
```
ccasin, line 32: Another error I think is really about layouts.  (That is, I think we wouldn't have trouble compiling probes that capture things where the other bits of the jkind are unusual).

ocaml/typing/typeopt.ml
```diff
@@ -16,18 +16,17 @@
...
-open Layouts
 open Types
 open Asttypes
 open Typedtree
 open Lambda
 
 type error =
-    Non_value_layout of type_expr * Layout.Violation.t option
-  | Non_value_sort of Sort.t * type_expr
+    Non_value_jkind of type_expr * Jkind.Violation.t option
```
ccasin, line 25: As in the `transl` files, my intuition is that this error is properly about layouts.

ocaml/typing/typeopt.ml
```diff
@@ -197,20 +196,20 @@ let bigarray_type_kind_and_layout env typ =
   | _ ->
       (Pbigarray_unknown, Pbigarray_unknown_layout)
 
-let value_kind_of_value_layout layout =
-  match Layout.get_default_value layout with
+let value_kind_of_value_jkind jkind =
```
ccasin, line 199: Probably this function really takes a layout argument and shouldn't be renamed.

ocaml/typing/typeopt.ml
```diff
@@ -372,27 +371,27 @@ let rec value_kind env ~loc ~visited ~depth ~num_nodes_visited ty
...
+        List.fold_left_map (fun num_nodes_visited field ->
+          let num_nodes_visited = num_nodes_visited + 1 in
+          (* CR layouts v5 - this is fine because voids are not allowed in
+             tuples.  When they are, we probably need to add a list of jkinds
+             to Ttuple, as in variant_representation and record_representation
+             *)
+          value_kind env ~loc ~visited ~depth ~num_nodes_visited field)
+          num_nodes_visited fields
+      in
+      num_nodes_visited,
```
ccasin, line 394: Stray formatting changes?

ocaml/typing/typeopt.ml
```diff
@@ -425,10 +424,10 @@ and value_kind_variant env ~loc ~visited ~depth ~num_nodes_visited
           List.fold_left_map
             (fun num_nodes_visited (ty, _) ->
                let num_nodes_visited = num_nodes_visited + 1 in
-               (* CR layouts v5: when we add other layouts, we'll need to check
+               (* CR layouts v5: when we add other jkinds, we'll need to check
```
ccasin, line 427: I think these comments are really about layouts.  (Though probably there are a lot of such comments, and it's not worth fixing them all)

ccasin (approved):
To record some out-of-band discussion:  Richard and I disagree a bit about the ideal structure of this PR.  Particularly in the new `Jkind.ml`, I think it would be clearer to leave the `Jkind.Layout` module unchanged, and jkinds should be added as a new thing `Jkind.t` that will today be an alias of `Jkind.Layout.t` but grow some new fields over time.  Richard prefers it the way he's done it here.

I don't think it's worth holding this up to argue further, though, so I'm approving.  To mitigate one major downside of Richard's approach, he's going to break this into two PRs.  The first just renames the file, and the second does the rest of these changes.  That way the history will record the file name change as a move.

I followed the suggestion review procedure.  I focused my spot check on files around the edges of the typechecker, and have some minor suggestions in those files about things I think should keep their old name (but feel free to disagree).

ocaml/testsuite/tests/typing-layouts/datatypes.ml
```diff
@@ -3,7 +3,7 @@
    flags = "-extension layouts"
 *)
 
-(* Tests for layouts in algebraic datatypes *)
+(* Tests for jkinds in algebraic datatypes *)
```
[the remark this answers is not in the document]
goldfirere: Almost all of the additions to type-checking are about kinds, I think, not layouts. The only real exception is around the checks to make sure that a type is representable (that is, not `any`). But any check for, say, compatbility between a declared classifier on a type and some other type is really a kind check, not just a layout check.

I went through the testsuite files and moved some of the comments back to layout.

ocaml/lambda/matching.ml
goldfirere: Yes, I agree here. I've reverted this change.

ocaml/lambda/matching.ml
goldfirere: This is debatable. The thing we care about is whether the layout is value, as you say. But the argument to this function must remain a jkind, because that's what we have to pass it. That all might change, but building up that distinction is going to happen in future PRs; I do not want to implement that in this one. I'll see if I can find a better name.

ccasin: (I'm convinced, no need to find a better name.)

ocaml/lambda/translcore.ml
goldfirere: OK

ocaml/typing/typeopt.ml
goldfirere: Disagree here. Types are classified by jkinds, not layouts, and the arguments to this function are type classifiers.

ocaml/typing/typeopt.ml
goldfirere: Yup. Reverted.

ocaml/typing/typeopt.ml
goldfirere: Yeah -- I actually tend to agree with you here. I've looked for uses of jkind in this file and changed as I saw appropriate.

goldfirere:
This PR has now been restructured to depend on #1886. This means git can recognize that what was previously layouts.ml is now jkind.ml.

I have resolved a few Discussions above, as Chris elsewhere invited me to do. I think once CI passes on #1886, we can merge that, and then this.

goldfirere:
The current state here is a little inconsistent: With this patch, the type checker loses its notion of "layout". Instead, it just has jkinds. But the modal kinds work will introduce a new concept called a layout. And so @ccasin wanted some comments to jump to their end state of saying "layout" instead of "jkind", even though I claim that the latter would be more consistent. Using "layout" means fewer changes, and the inconsistency will be temporary. An alternative approach would be to add the (re)introduction of layouts into this PR, but I thought having this be just a renaming -- no new logic! -- would be better. I think there's no perfect answer here.

PR #1876, opened 2023-09-27
Enable ocamlformat for Jane Syntax / language extensions code

ncik-roberts:
Enable ocamlformat for these files:
  * `ocaml/parsing/jane_syntax.{ml,mli}`
  * `ocaml/parsing/jane_asttypes.mli`
  * `ocaml/parsing/jane_syntax_parsing.{ml,mli}`
  * `ocaml/utils/language_extension.{ml,mli}`

Keep the same settings as other places ocamlformat is used in the repo, except disable comment wrapping (because it mangles lots of the existing comments in `jane_syntax.ml`).

This PR makes it so we can make changes to these files without carefully hand-formatting the resulting code. These files don't exist upstream yet (or probably ever, for these particular ones) so this doesn't introduce more surface area for merge conflicts.

You run `make fmt` to format all files in the repo. CI will complain if you forget to do this.

**Review:** probably worth glancing to make sure this doesn't make things appreciably worse, but this one should be pretty straightforward. Maybe @goldfirere can do that glancing?

### DIFF
--- a/ocaml/parsing/.ocamlformat
+++ b/ocaml/parsing/.ocamlformat
@@ -1,0 +1,17 @@
+# Please make a pull request to change this file.
+disable=true
+# There is an .ocamlformat-enable file in this directory.
+# Keep the remainder of this file in sync with other .ocamlformat files in this repo.
+assignment-operator=begin-line
+cases-exp-indent=2
+doc-comments=before
+dock-collection-brackets=false
+if-then-else=keyword-first
+module-item-spacing=sparse
+parens-tuple=multi-line-only
+sequence-blank-line=compact
+space-around-lists=false
+space-around-variants=false
+type-decl=sparse
+wrap-comments=false
+version=0.24.1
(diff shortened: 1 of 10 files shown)


goldfirere:
This is fine, I guess?

When a project universally uses ocamlformat and your editor knows this and automatically keeps your code formatted, that's very nice. But now we'll have to remember which files are formatted and which ones aren't. Our editors won't know this, and so the formatting won't be automatic. Instead, we'll get a CI failure, likely after we've task-switched upon pushing. And it won't necessarily save us from manually formatting, because we have manual formatting as a habit (because most of the code doesn't get formatted automatically).

So, basically, for me, it comes down to: is there a way our editors can slurp in these settings and automatically format the files that should be formatted? If yes, then I'm in strong support. If no, I think the downsides outweigh the upsides.

That said, if I'm in the minority here, I won't block this.

ncik-roberts:
All `make fmt` does is run `ocamlformat -i` from the root of the repo. So there will certainly be a way to rig up an editor to automatically format these files. Happy to block this PR until I’ve produced a witness to that assertion.

ncik-roberts:
I've sent Richard (what I hope to be) a witness for that assertion.

goldfirere:
I have not tested the emacs code Nick sent, but it is very plausible. Happy to Approve this and then iterate on editor support.

goldfirere (approved):

PR #1682, opened 2023-08-02
128-bit Array Load/Store

TheNumbat:
Adds 128-bit load/store for strings, bytes, and bigstrings. Tested in `tests/simd/arrays.ml`.

Four types of operations are supported: unaligned, unsafe unaligned, aligned, and unsafe aligned. 

The safe aligned operations are only exposed for bigstrings, as strings/bytes can be moved by the GC. To support the bigstring check, I've also added a new built-in exception  to the runtime, `Pervasives.array_align_error`. A new instruction, `Ccheckalign`/`Icheckalign`, may raise this exception.

This PR does not add a vec128 bigarray, as this would make the stdlib bigarray module need `-extension simd`.
Additionally, 128-bit operations are not supported in `ocaml/asmcomp`, 32-bit mode, or bytecode mode.

PR #1878, opened 2023-09-28
Refactor x86-specific float64 intrinsics

TheNumbat:
(Merges into #1682).

Moves x86-specific 64-bit float operations into the SIMD selection path. 
These operations were already SSE2/SSE4.1 instructions; emitted code is identical.

This PR does not refactor `Ifloatarithmem` and `Ifloatsqrtf`, which have the additional constraint of always loading from a non-stack-spill memory location.

Also updates the name of the SSE4.2 CRC32 intrinsics to be consistent with the others.
These changes will hence require updating the ocaml_intrinsics library with the new operation names.


backend/amd64/simd.ml
```diff
@@ -443,15 +453,19 @@ let equal_operation_sse41 l r =
   | Multi_sad_u8 l, Multi_sad_u8 r
     when l = r ->
     true
-  | (Round_f64 l, Round_f64 r | Round_f32 l, Round_f32 r) when l = r -> true
+  | Round_sf64 l, Round_sf64 r
+  | Round_f64 l, Round_f64 r
+  | Round_f32 l, Round_f32 r
+    when l = r ->
```
xclerc, line 459: (Not really related to this pull request,
but we should probably avoid polymorphic
equality when defining such functions.)

xclerc, line 459: (Not really related to this pull request,
but we should probably avoid polymorphic
equality when defining such functions.)

xclerc (approved):

backend/amd64/simd.ml
TheNumbat: Fixed in the SIMD PR

backend/amd64/simd_selection.ml
```diff
@@ -400,21 +408,22 @@ let register_behavior_sse = function
...
+  | SLL_i64 | SRL_i16 | SRL_i32 | SRL_i64 | SRA_i16 | SRA_i32 | Avg_unsigned_i8
   | Avg_unsigned_i16 | SAD_unsigned_i8 | Shuffle_64 _ | Interleave_high_8
   | Interleave_high_16 | Interleave_high_64 | Interleave_low_8
   | Interleave_low_16 | Interleave_low_64 | I16_to_i8 | I32_to_i16
   | I16_to_unsigned_i8 | I32_to_unsigned_i16 ->
     R_RM_to_fst
   | Shuffle_high_16 _ | Shuffle_low_16 _ | I32_to_f64 | I32_to_f32 | F64_to_i32
-  | F64_to_f32 | F32_to_i32 | F32_to_f64 ->
+  | Cast_scalar_f64_i64 | F64_to_f32 | F32_to_i32 | F32_to_f64 | Sqrt_scalar_f64
+    ->
```
gretay-js, line 426: Previously, `sqrt` was constraint to registers as arg and res. It can take arg in memory so your change is fine, but please make sure to have a specific test for this.

gretay-js (commented):
There are a couple of changes in this PR that are not strictly refactoring and would change the generated code. I would prefer to have them split out into a separate PR, particularly `sqrt`.

I would have preferred to leave `sqrt` alone where it was because it's already this way upstream (until we refactor memory operations too) but since you have already implemented the chance, let's keep it.

backend/amd64/emit.mlp
```diff
@@ -1506,21 +1504,8 @@ let emit_instr fallthrough i =
       I.bswap (res32 i 0);
   | Lop(Ispecific(Ibswap { bitwidth = Sixtyfour })) ->
       I.bswap (res i 0)
-  | Lop(Ispecific Isqrtf) ->
-      if arg i 0 <> res i 0 then
-        I.xorpd (res i 0) (res i 0); (* avoid partial register stall *)
```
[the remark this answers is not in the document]
TheNumbat: Hmm, I was thinking this will no longer be necessary when we start using VEX encoded operations (it no longer preserves the destination's upper bits), but that's not a good reason to remove it now...added back

backend/amd64/emit.mlp
```diff
@@ -1506,21 +1504,8 @@ let emit_instr fallthrough i =
...
-      I.sqrtsd (arg i 0) (res i 0)
   | Lop(Ispecific(Ifloatsqrtf addr)) ->
-      I.xorpd (res i 0) (res i 0); (* avoid partial register stall *)
```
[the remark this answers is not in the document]
TheNumbat: Same as above - added back

backend/amd64/simd_selection.ml
TheNumbat: What's the best way to test this? The operand won't get spilled unless there is particular register pressure, but even in that case it seems to prefer loading the operand to `xmm0` and then doing `sqrtsd xmm0,xmm0`

gretay-js: not sure, @xclerc  do you have a suggestion? 

if it's hard to test, perhaps we should keep is as `R_to_R` in `register_behavior` in this PR, to make progress, and make this change in a separate PR (would be a good idea anyway)?

TheNumbat: Switched to `R_to_R`

gretay-js (approved):

PR #1741, opened 2023-08-14
Regalloc/split: move definitions (resp. destructions) down (resp. up)

xclerc:
As an optimization, after computing the
"definitions at beginning" and "destruction
at end" sets from destruction points, move
definitions (resp. destructions) down
(resp. up) when possible. This reduces the
live range of registers and creates more
opportunities for the optimization removing
definitions and destructions of unused
registers. Effectively, it means that a reload
happening for the sole purpose or re-spilling
the value (in straightline) will disappear.
On the compiler distribution, the combination
of #1739 and this very pull request removes
1.9% of the spills and 2.6% or the reloads.

PR #1879, opened 2023-09-29
Regalloc: disable spill/reload motion between blocks

xclerc:
The code motion introduced by https://github.com/ocaml-flambda/flambda-backend/pull/1741 breaks
an invariant used later when computing and
propagating the substitutions for the blocks.

This pull request comments out the call to the
new code until we push a new handling of
the substitutions compatible with the code
motion.


gretay-js (approved):

xclerc:
(The new handling I was referring to above is in https://github.com/ocaml-flambda/flambda-backend/pull/1902.)

PR #1880, opened 2023-09-29
Remove var_constraint

goldfirere:
This was a small simplification that I spotted.

Note that this patch targets the `rae/layout-to-kynd` branch to avoid conflicts. My hope is that `rae/layout-to-kynd` lands first, at which point I will retarget this PR.


Ekdohibs:
I think this needs rebasing now that #1875 has been merged.

goldfirere:
Rebased and retargeted

ncik-roberts (approved):
I reviewed this PR as my PR added this code in #1474. I agree that the post-condition holds as documented, and I can't recall why I structured the code the way I did. (An error message in my PR clearly states the same post-condition &mdash; maybe I was concerned about needing to keep track of this post-condition across module boundaries, but this does not seem like a compelling argument...) Looks good to merge to me.

PR #1881, opened 2023-09-29
ocamlformat in Jane-Street only files in typing

goldfirere:
Follow on from #1876.

(This is pointing to #1880, to avoid merge conflicts. First, let's merge #1875, then #1880, then this. I'll do the git-wrangling.)

### DIFF
--- a/ocaml/typing/.ocamlformat
+++ b/ocaml/typing/.ocamlformat
@@ -1,0 +1,21 @@
+# Please make a pull request to change this file.
+disable=true
+# There is an .ocamlformat-enable file in this directory.
+# Keep the remainder of this file in sync with other .ocamlformat files in this repo.
+assignment-operator=begin-line
+cases-exp-indent=2
+doc-comments=before
+dock-collection-brackets=false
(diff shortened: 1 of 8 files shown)


riaqn:
While we are at it, could you also add `lambda/translmode.ml*` as well?

ncik-roberts (approved):
Looks good to me.

ocaml/typing/.ocamlformat
```diff
@@ -0,0 +1,21 @@
+# Please make a pull request to change this file.
```
ncik-roberts, line 1: Please resolve this comment when the PRs you mentioned are merged. Then, you can merge this PR.

ncik-roberts: (never mind, it looks like this doesn't block merging. so I'll just resolve this now.)

goldfirere:
#1880 should be merged before this one to make reviewing that PR easier. If #1880 gets sufficiently stalled, I'll just reshuffle this one to go underneath that one.

PR #1882, opened 2023-09-29
ocamlformat in JS-only files in lambda

goldfirere:
Further opportunities for automatic formatting, following #1876. Suggested by @riaqn.

(In contrast to #1881, this one points directly at `main`, as the conflicts with #1875 will be small.)


mshinwell:
@goldfirere this seems fine, can you rebase and reformat?

PR #1883, opened 2023-09-29
Added a hook on imported units

poechsel:
When writing tools connecting to the compiler it is often useful to know which external compilation units are needed, and from which source file they were generated from.
This PR adds a new compiler hook to retrieve that information. Most of the code is straightforward but the two following needed to happen to have it working:
- `Checkmach` depends on `Compilenv` so to avoid a cyclic dependency I moved types definitions out from `Checkmach`
- The `cmx` format is now slightly altered. The absolute path to the filename (after taking into account any build prefix) is now stored after the `crc` as the last field. It is stored away from the unit info to avoid being part of the crc computation

### DIFF
--- a/backend/checkmach_types.ml
+++ b/backend/checkmach_types.ml
@@ -1,0 +1,143 @@
+(**********************************************************************************
+ *                             MIT License                                        *
+ *                                                                                *
+ *                                                                                *
+ * Copyright (c) 2022-2022 Jane Street Group LLC                                  *
+ *                                                                                *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy   *
+ * of this software and associated documentation files (the "Software"), to deal  *
+ * in the Software without restriction, including without limitation the rights   *
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell      *
+ * copies of the Software, and to permit persons to whom the Software is          *
+ * furnished to do so, subject to the following conditions:                       *
+ *                                                                                *
+ * The above copyright notice and this permission notice shall be included in all *
+ * copies or substantial portions of the Software.                                *
+ *                                                                                *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR     *
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,       *
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE    *
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER         *
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,  *
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE  *
+ * SOFTWARE.                                                                      *
+ *                                                                                *
+ **********************************************************************************)
+[@@@ocaml.warning "+a-30-40-41-42"]
+
+module Witness = struct
+  type kind =
+    | Alloc of
+        { bytes : int;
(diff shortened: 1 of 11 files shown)


backend/checkmach_types.ml
```diff
@@ -0,0 +1,143 @@
+(**********************************************************************************
+ *                             MIT License                                        *
+ *                                                                                *
+ *                                                                                *
+ * Copyright (c) 2022-2022 Jane Street Group LLC                                  *
```
gretay-js, line 5: update or remove

backend/checkmach_types.ml
```diff
@@ -0,0 +1,143 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,       *
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE    *
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER         *
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,  *
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE  *
+ * SOFTWARE.                                                                      *
+ *                                                                                *
+ **********************************************************************************)
+[@@@ocaml.warning "+a-30-40-41-42"]
+
```
gretay-js, line 27: This file is only concerned with witnesses. It would be better to rename it `witness` or `witnesses` specifically and adjust the types within it to be "flatter".  We can also use it when the analysis is ported to CFG, so I'd would avoid `checkmach` in the name (maybe add a comment that it is currently used in Checkmach). Just putting it here while I am thinking about it. I'm not suggesting to make these changes in the current PR, we can do this later (although it would require updating ocaml-ir again).

driver/compiler_hooks.mli
```diff
@@ -46,7 +46,9 @@ type _ pass =
   | Cmm : Cmm.phrase list pass
 
   | Inlining_tree : Flambda2_simplify_shared.Inlining_report.Inlining_tree.t pass
-  | Check_allocations : Checkmach.iter_witnesses pass
+  | Check_allocations : Checkmach_types.iter_witnesses pass
+
+  | Imported_compilation_unit : (Compilation_unit.t * string) pass
```
gretay-js, line 51: What is the meaning of the `string` component?

tools/flambda_backend_objinfo.ml
```diff
@@ -341,6 +341,7 @@ let dump_obj_by_kind filename ic obj_kind =
     let first_section_offset = pos_in ic in
     seek_in ic (first_section_offset + uir.uir_sections_length);
     let crc = Digest.input ic in
+    let _filename = (input_value ic : unit_infos_associated_source) in
```
gretay-js, line 344: Why not printing it here?

gretay-js (commented):
There is a change to CMX file format, so this PR needs to be reviewed by someone from the middle-end and will require magic number bump. 

I am not convinced that absolute path is needed. Why is it not enough to pass `CU.Name.to_string` or `filename` computed in `Compilenv.get_unit_info`? We don't have directory names in error messages. I don't see why this will be different.

~I am worried that including absolute path in `.cmx` will have bad effect on build cache / build reproducibility, but I am not sure what exactly they rely on.~

The backend changes are fine. Ideally, moving checkmach types out would be in a separate PR.

gretay-js:
@poechsel pointed out offline that Location.absolute_path doesn't always return the absolute path, but something else that is build-cache friendly.

gretay-js:
> When writing tools connecting to the compiler it is often useful to know which external compilation units are needed, and from which source file they were generated from.

This seems to be useful for inlining decision visualization: when looking at a function, we want to find the file that contains the source code of all the callees that were inlined. It looks like something that merlin already knows (e.g., when doing go to definition, how does it do it)? It may be useful to have the path the `.cmx` anyway.

mshinwell:
@poechsel should this PR be closed, at least for the moment?

mshinwell:
I'm going to close this for the moment.

PR #1885, opened 2023-09-29
Undo sort changes during snapshot backtrack

alanechang:
This PR records layout sort changes in snapshot history and add support for undoing these changes during backtrack.

### DIFF
--- a/ocaml/testsuite/tests/typing-layouts/annots.ml
+++ b/ocaml/testsuite/tests/typing-layouts/annots.ml
@@ -434,3 +434,6 @@ Line 1, characters 21-30:
 Error: Layout immediate is more experimental than allowed by -extension layouts.
        You must enable -extension layouts_beta to use this feature.
 |}]
+
+(************************************************************************)
+(* Test 11: Annotation on type parameter within constructor declaration *)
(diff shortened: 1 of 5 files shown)


ocaml/typing/layouts.ml
```diff
@@ -30,6 +30,14 @@ module Sort = struct
     | Const of const
   and var = t option ref
 
+
+  (* recording changes *)
```
ccasin, line 34:
```suggestion
  (* recording changes to sorts, for use with `Types.{snapshot, backtrack}` *)
```

ocaml/typing/layouts.mli
```diff
@@ -68,6 +68,13 @@ module Sort : sig
       it is set to [value] first. *)
   val get_default_value : t -> const
 
+  type change
```
ccasin, line 71: Probably these should each get a comment (or group them together without newlines and put a comment above)

ocaml/typing/layouts.ml
```diff
@@ -30,6 +30,14 @@ module Sort = struct
     | Const of const
   and var = t option ref
 
+
```
ccasin, line 33: extra newline

ccasin (approved):
Looks good.  I discussed testing with Alan.  Hard to add a meaningful test now, but he's going to add a CR to revisit after histories are merged (or possibly add the instigating example as test with a CR to check it then).

PR #1886, opened 2023-09-29
Rename layouts.ml to jkind.ml

goldfirere:
This is a preparatory step before #1875, which is the main payload. That PR makes changes within the jkind.ml, and it means that git doesn't recognize the filename change. This PR will immediately be followed up by #1875, and so does not need close scrutiny. The key thing is just that the file name change is recognized by git.


ccasin (approved):

ccasin:
Richard and I agree he should merge this, once it passes CI, without waiting for code owner approval for the remaining files.  This seems fine because the changes are just a renaming in the type type system, and is warranted because of the high vulnerability to merge conflicts.  He will do the same for #1875 .

PR #1888, opened 2023-10-02
Support underscore partial application

riaqn:
Adds support for underscore partial application. We had first attempt at https://github.com/ocaml-flambda/ocaml-jst/pull/52 but decided that native type checking is better than syntax unfolding.

Its interaction with labels and omitted/optional arguments are subtle; see tests for examples. But in short, it works as follows:
1. For an application, collect arguments as before. Unlabelled underscores are treated as positional arguments, and labelled underscore are treated as labelled arguments.
2. Then, for each underscore, put a `fun x -> ...` around the application, in the order of the function type (not of the application). Replace the underscore with `x`.

For example, `foo x _ y _ z` can be understood as `fun u v -> foo x u y v z`.
Note an alternative interpretation:
```
let f = foo x in 
fun u -> 
let g = f u y in 
fun v -> 
g v z
```
which is if we take underscore simply as `Omitted` arguments - but underscores are not `Omitted` arguments.


ncik-roberts:
I took a brief look at this because I was curious and had two quick comments.

1. Should there be a test to witness the fact that non-underscored arguments are evaluated once, at the callsite with underscores, and not each time the function is called? I see the `"world" ^ "hello"` test and wonder if we should add something with effects, where the repeated work obviously can't be hoisted out by some optimization pass. Something like [1]
2. Shouldn't there be a language extension for this?

[1]
```
# let f x y = x + y;;
val f : int -> int -> int = <fun>
# let g = f _ (print_endline "evaluated"; 4);;
evaluated
val g : int -> int = <fun>
# g 6;;
- : int = 10
```

riaqn:
@ncik-roberts that's good point - I addressed [1]. I will do [2] tomorrow - although I'm not sure how should we proceed with the upcoming jane-syntax PR.

ocaml/testsuite/tests/typing-misc/underscore.ml
```diff
@@ -0,0 +1,93 @@
...
+Line 1, characters 15-16:
+1 | let g = f ~foo:_
+                   ^
+Error: Underscore argument cannot be wrapped in Some.
+|}]
+
+(* Underscore can still be supplied as optional argument raw *)
+let g = f ?foo:_
+[%%expect{|
+val g : ?foo:'_weak1 -> unit -> string = <fun>
```
Ekdohibs, line 80: It feels strange this is a weak type variable, is this expected?

riaqn: yeah this is bug - in some cases it will cause crash - fixing.

riaqn:
Ready for review after fixing a bug reported by @ncik-roberts  - do you want to review this, Nick?

ocaml/typing/typecore.ml
```diff
@@ -3339,9 +3347,16 @@ let collect_unknown_apply_args env funct ty_fun mode_fun rev_args sargs ret_tvar
                               (expand_head env funct.exp_type)))
         in
         let arg =
-          Unknown_arg { sarg; ty_arg_mono; mode_fun; mode_arg; sort_arg }
+          match sarg.pexp_desc with
+          | Pexp_extension({txt = "extension.dummy"; loc}, PStr []) ->
+              if not (Language_extension.is_enabled Dummy_arguments) then
+                raise (Typetexp.Error (loc, env,
+                  Unsupported_extension Dummy_arguments));
+              Dummy {ty_arg; mode_fun; mode_arg; sort_arg; level = lv}
```
ncik-roberts, line 3355: I would prefer for the language extension check to be moved into the Jane Syntax module along with the literal extension string, but I don't feel too strgonly about this.

ocaml/typing/typecore.ml
```diff
@@ -3392,6 +3417,11 @@ let collect_apply_args env funct ignore_labels ty_fun ty_fun0 mode_fun sargs ret
             (Eliminated_optional_arg
                { mode_fun; ty_arg; mode_arg; sort_arg; level = lv })
         in
+        let omit_arg () =
+          may_warn funct.exp_loc
+            (Warnings.Non_principal_labels "commuted an argument");
+          Omitted { mode_fun; ty_arg; mode_arg; level = lv; sort_arg }
+        in
```
ncik-roberts, line 3424: Why factor this out? It seems like it's just used once.

ocaml/typing/typecore.ml
```diff
@@ -3428,58 +3458,93 @@ let collect_apply_args env funct ignore_labels ty_fun ty_fun0 mode_fun sargs ret
...
                       marg (mode_partial_application expected_mode);
                     Value.regional_to_local_alloc marg)
                  open_args
              in
              let closed_args = new_closed_args @ closed_args in
              let open_args = [] in
-             let mode_closed_args = List.map Alloc.close_over closed_args in
-             let mode_partial_fun = Alloc.partial_apply mode_fun in
+             (* note that [mode_fun] is closed_over, not partial_apply *)
+             let all_closed = mode_fun :: closed_args @ underscore_args in
```
ncik-roberts, line 3502: Is there a test that we can write (perhaps involving unique/once, which you mentioned distinguish between `partial_apply` and `close_over`) that will help us avoid regressing in the future?

ncik-roberts (commented):
This may seem like a lot of comments, but they're very minor across the board. I need to finish reviewing `type_omitted_parameters`, and I haven't started reviewing `Translcore`, but here are some comments in the meantime.

riaqn:
@ncik-roberts thanks for the review - they are all great points. I'm gonna wait for the rest of the review and fix all in one go.

ocaml/typing/typecore.ml
```diff
@@ -3428,58 +3458,93 @@ let collect_apply_args env funct ignore_labels ty_fun ty_fun0 mode_fun sargs ret
...
              register_allocation_mode mode_closure;
              let arg = Omitted { mode_closure; mode_arg; mode_ret; sort_arg } in
              let args = (lbl, arg) :: args in
-             (ty_ret, mode_closure, open_args, closed_args, args))
-      (ty_ret, mode_ret, [], [], []) (List.rev args)
+             (ty_ret, mode_closure, open_args, closed_args, underscore_args, args))
+      (ty_ret, mode_ret, [], [], [], []) (List.rev args)
+  in
+  let closed_args =
+    if List.length underscore_args = 0 then closed_args
```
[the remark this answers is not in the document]
ncik-roberts: Along the same lines, I think it might be clearer to structure this as something like:

```ocaml
match underscore_args with
| [] -> ty_ret, mode_ret, args
| _ :: _ ->
  (* complicated logic including closing over of open args and [type_underscore_parameters] *)
```

instead of:

```ocaml
let closed_args =
  if List.length underscore_args = 0 then closed_args
  else (* close over open args *)
in
(* [type_underscore_parameters] *)
```

One concrete reason for this: `closed_args` only makes sense as "closed" if there are any underscore args

ocaml/typing/typecore.ml
```diff
@@ -3428,58 +3458,93 @@ let collect_apply_args env funct ignore_labels ty_fun ty_fun0 mode_fun sargs ret
...
+      List.map
+        (fun (marg, loc) ->
+          submode ~loc ~env ~reason:Other
+            marg (mode_partial_application expected_mode);
+          Value.regional_to_local_alloc marg)
+        open_args
+    in
+    funct_mode :: new_closed_args @ closed_args
+  in
+  let rec type_underscore_parameters closed_args = function
```
ncik-roberts, line 3525: What do you think about hoisting this out to a top-level function? That makes it clear which local variables it closes over. (In this case there are only two, the expected ones.) feel free to push back if you would rather not.

ocaml/typing/typecore.ml
```diff
@@ -3428,58 +3458,93 @@ let collect_apply_args env funct ignore_labels ty_fun ty_fun0 mode_fun sargs ret
...
+        in
+        register_allocation_mode mode_closure;
+        let arg = Dummy {mode_closure; mode_arg; mode_ret; sort_arg } in
+        ty_ret, mode_closure, (lbl, arg) :: args
+    | (lbl, Omitted r) :: rest ->
+      let ty_ret, mode_ret, args = type_underscore_parameters closed_args rest in
+      ty_ret, mode_ret, (lbl, Omitted r) :: args
+    | (lbl, Arg r) :: rest ->
+      let ty_ret, mode_ret, args = type_underscore_parameters closed_args rest in
+      ty_ret, mode_ret, (lbl, Arg r) :: args
```
ncik-roberts, line 3545: You can change this to one case with an or-pattern:

```suggestion
    | (lbl, (Omitted _ | Arg _ as arg)) :: rest ->
      let ty_ret, mode_ret, args = type_underscore_parameters closed_args rest in
      ty_ret, mode_ret, (lbl, arg) :: args
```

`as` patterns can generalize types that aren't mentioned by the constructors that appear in them.

ocaml/typing/typecore.ml
```diff
@@ -3428,58 +3458,93 @@ let collect_apply_args env funct ignore_labels ty_fun ty_fun0 mode_fun sargs ret
...
+  let closed_args =
+    if List.length underscore_args = 0 then closed_args
+    else
+    let new_closed_args =
+      List.map
+        (fun (marg, loc) ->
+          submode ~loc ~env ~reason:Other
+            marg (mode_partial_application expected_mode);
+          Value.regional_to_local_alloc marg)
+        open_args
```
ncik-roberts, line 3521: This appears twice -- should we factor it out to a helper function? Maybe name it `close_over_arg`?

ncik-roberts (commented):
Done with typecore! Only one of my comments is potentially a bug and not just a nitpick. Now looking at translation.

ocaml/lambda/translcore.ml
```diff
@@ -1177,24 +1204,78 @@ and transl_apply ~scopes
...
-  let args =
-    List.map
-      (fun (_, arg) ->
-         match arg with
-         | Omitted _ as arg -> arg
-         | Arg (exp, sort_arg) ->
-           Arg (transl_exp ~scopes sort_arg exp, layout_exp sort_arg exp))
-      sargs
+
+  let rec build_apply_underscore lam rev_args loc prot = function
```
ncik-roberts, line 1212: change to "dummy" here

ncik-roberts: This looks like a big block of code that's very similar to `build_apply` but with subtle differences. I'd really like to avoid that. (Also `prot` is just an optimization right? I don't think the old code has it. But I agree we'll run into a lot more so might as well avoid quadratic behavior.)

I realize why you've done this &mdash; you want to put the dummy parameters first in the lambda function &mdash; but can't we do this in a pre-processing step where we:

1. build up the entire arg list in the correct order
2. reorder the original arg list to put the dummy arguments first
3. call `build_apply` where the eventual body is an application to the correctly-ordered arg list (not the reordered arg list)

There might be some reason that this doesn't work, but if it does work, I think we can avoid duplicating the logic in `build_apply` and here.

ncik-roberts (commented):
Done with my first pass at review. This one is mostly minor suggestions + one larger suggestion for translation.

ocaml/typing/untypeast.ml
```diff
@@ -515,6 +515,7 @@ let expression sub exp =
         Pexp_apply (sub.expr sub exp,
           List.fold_right (fun (label, arg) list ->
               match arg with
+              | Dummy _ -> list
```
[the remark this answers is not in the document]
riaqn: Yeah I cut some corners because I think this file is almost never used. Will fix - thanks.

ocaml/typing/typecore.ml
```diff
@@ -3569,6 +3632,21 @@ let rec is_nonexpansive exp =
...
+       | Dummy _ -> true
+       | _ -> false
+       ) args
+     in
+     if has_underscore then true
+     else begin
+       match args with
+       | (_, Omitted _) :: _ -> true
+       | _ -> false
+     end
```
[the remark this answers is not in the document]
riaqn: Thanks for for suggestion. Will adopt.

ocaml/testsuite/tests/typing-local/local.ml
```diff
@@ -2723,3 +2724,71 @@ Line 2, characters 33-58:
...
+                         ^^^
+Error: This local value escapes its region
+  Hint: Cannot return local value without an "exclave_" annotation
+  Hint: This is a partial application
+        Specifying 1 more argument may make the value non-local
+|}]
+
+let foo (local_ f) = local_ f _
+[%%expect{|
+val foo : local_ ('a -> 'b) -> 'a -> 'b = <fun>
```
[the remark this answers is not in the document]
riaqn: If explicit, this should print 
`local_ ('a -> 'a) -> local_ ('a -> 'b)`
which is equivalent to the implicit printing here.

ocaml/testsuite/tests/typing-misc/underscore.ml
```diff
@@ -0,0 +1,154 @@
...
+|}]
+
+(* Dummy can still be supplied as optional argument raw *)
+let g = f ?foo:_
+[%%expect{|
+val g : ?foo:'a -> unit -> string = <fun>
+|}]
+
+(* Poly bug reported by nick roberts *)
+let f g = g _ _ _
```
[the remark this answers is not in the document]
riaqn: I simplified your example a bit because I think the core issue was about polymorphism rather than mode stuff - but I'm happy to add that back. What do you think?

ocaml/testsuite/tests/typing-misc/underscore.ml
```diff
@@ -0,0 +1,154 @@
...
+|}]
+
+let h'_ref = f_ref h_ref
+[%%expect{|
+Line 1, characters 19-24:
+1 | let h'_ref = f_ref h_ref
+                       ^^^^^
+Error: This argument has type 'b -> 'b which is less general than
+         'a. 'a -> 'a
+|}]
```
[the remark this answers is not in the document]
riaqn: I think a major use case of this has been remedied by https://github.com/ocaml-flambda/flambda-backend/pull/1756

Also we implement `_` such that `f _` is translated to `fun x@m -> f m` where `m` is simply copied from `f : a@m -> ...`. So dummy arguments can't be used for the purpose you said.

ocaml/typing/typecore.ml
riaqn: It seems that the other two cases are factored out (`eliminate_arg` and `use_arg`), so I factored this out for symmetry.

ocaml/typing/typecore.ml
riaqn: In your example of `g`, when we look at `Omitted y`, the function applying to `y` is `f x`, and thus `mode_fun` describes the mode of `f x`. 

In your example of `f`, we will look at `Omitted x`, the function applying to `x` happens to be `f`, and thus `mode_fun` describes the mode of `f`.

riaqn: As for `close_over` vs `partial_apply`, consider a function
```
let unique_ f ~x ~y = ... 
```
then we have 
```
f ~y
```
unfolds to
```
fun ~x -> f ~x ~y 
```
And because `f` is `unique`, this function closing over it should be `once`.

But of course I'm wrong; this closing over of `f` is result of unfolding and has specific shape. In particular, it doesn't use any capability provided by the uniqueness of `f`. That is, we can think this function closes over weakened-to-`shared` version of `f` and doesn't need to be `once`. 

I will fix this.

ocaml/typing/typecore.ml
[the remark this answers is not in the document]
riaqn: That makes great sense, thanks.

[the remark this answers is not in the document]
riaqn: Sadly we have to traverse all `args` again using `type_underscore_parameters` anyway, because polyrmorphism type issues:
```
Error: This expression has type
         (Asttypes.arg_label *
          (Typedtree.expression * 'a, Typedtree.omitted_parameter,
           untyped_dummy_param)
          Typedtree.arg_or_omitted)
         list
       but an expression was expected of type
         (Asttypes.arg_label *
          (Typedtree.expression * 'a, Typedtree.omitted_parameter,
           Typedtree.dummy_parameter)
          Typedtree.arg_or_omitted)
         list
       Type untyped_dummy_param is not compatible with type
         Typedtree.dummy_parameter 
```
(The list would not contain any `dummy`, but the type would mismatch)

ocaml/typing/typecore.ml
riaqn: I have no preference, so will do that.

ocaml/lambda/translcore.ml
riaqn: I feel that `Omitted` and `Dummy` are different enough that it's easier to process them in two passes (like what we have now). But I've mechanically factor out the common part of the two functions to reduce duplication.

ocaml/parsing/jane_syntax.ml
```diff
@@ -1767,3 +1767,25 @@ module Extension_constructor = struct
       (* See Note [Outer attributes at end] *)
       { ext_ctor with pext_attributes = ext_ctor.pext_attributes @ attrs }
 end
+
+module Dummy_arguments = struct
+  type expression = Dummy_argument
+
+  let embedded_name = "jane.non_erasable.dummy_arguments"
```
riaqn, line 1774: I would love to use `Embedded_name.to_string` but it's not exposed. Its comment says "not expose" which I suppose has a reason behind it.

ocaml/testsuite/tests/typing-misc/underscore.ml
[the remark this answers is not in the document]
ncik-roberts: OK.

ocaml/testsuite/tests/typing-misc/underscore.ml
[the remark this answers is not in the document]
ncik-roberts: #1756 seems useful for 1-argument functions.

Hmm, I think this may be problematic. I'll ask in the modes channel.

ocaml/typing/typecore.ml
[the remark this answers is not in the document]
ncik-roberts: That makes a lot of sense -- but could we turn the call in that case into `List.map (function | Omitted | Present as x -> x | Dummy -> Misc.fatal_error ...)`? That's total pseudocode but you know what I mean. Please resolve upon addressing.

ocaml/parsing/jane_syntax.ml
ncik-roberts: Could we define this constant in `jane_syntax_parsing.ml`? That way we don't have to expose `Embedded_name.to_string` and don't duplicate its implementation (kind of).

ocaml/typing/typecore.ml
ncik-roberts: We discussed this and agree the new code is right. I'll suggest a comment in review in a moment.

ocaml/testsuite/tests/typing-misc/underscore.ml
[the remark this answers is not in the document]
ncik-roberts: We're going to do this in another PR.

ocaml/lambda/translcore.ml
```diff
@@ -1124,77 +1126,126 @@ and transl_apply ~scopes
...
+          let id = Ident.create_local name in
+          defs := (id, layout, lam) :: !defs;
+          (Lvar id, layout)
+    in
+    let body = f protect in
+    List.fold_right
+      (fun (id, layout, lam) body -> Llet(Strict, layout, id, lam, body))
+      !defs body
+  in
+  let lfunction_helper ~mode_closure ~mode_arg ~mode_ret loc
```
ncik-roberts, line 1144: I was about to suggest to call this `build_unary_function_for_elided_argument` but then that made me realize... this only builds unary functions! That's not good. We need `f _ _` to be a two-argument function for efficiency.

ncik-roberts (commented):
One change requested.

ocaml/lambda/translcore.ml
ncik-roberts: We might take a different approach in light of my most recent comment (the function we build shouldn't be a series of nested unary functions for dummy arguments) -- perhaps the code sharing will be dramatically less and we won't need `lfunction_helper`.

ocaml/parsing/jane_syntax.ml
riaqn: My impression of `Jane_syntax_parsing` is that it's not specific to any extension or syntax - so probably shouldn't have a `string` specific to dummy arguments? Maybe it's not dangerous to expose `Embedded_name.to_string`?

ocaml/lambda/translcore.ml
riaqn: Yes that makes sense. I will rewrite this part into mutli-arguments function.

ocaml/parsing/jane_syntax.ml
ncik-roberts: I don't have a strong preference between either approach.

riaqn:
This is a reminder to myself that we should include some docs in the PR.

PR #1889, opened 2023-10-02
Improve layout history error message: hide layout history reason of various generalized vars

alanechang:
This PR erases the layout history reason of certain generalized type variables. Motivation behind this is when explaining the reason of an error, the implementation details of some other value/type defined far away is not particularly interesting nor helpful to the user.

To illustrate, this PR changes the error message of this example:

```ocaml
module _ = struct
  module M : sig
    val f : 'a -> 'a
  end = struct
    let f x = x
  end
  let g (x : t_void) = M.f x
end
```

in such a way:

```diff
Line 8, characters 27-28:
8 |   let g (x : t_void) = M.f x
                               ^
Error: This expression has type t_void but an expression was expected of type
         ('a : value)
       The layout of t_void is void, because
         of the annotation on the declaration of the type t_void.
       But the layout of t_void must be a sublayout of value, because
-        it's used as a function argument, defaulted to layout value.
+        of definition of f at Line 3, characters 4-20.
```

This also puts the error reporting of layouts more in line with what's being done for types.

In terms of implementation details, this layout reason erasing behavior is added at three places:
- for let statements within `type_let`
- for val/external declarations within `transl_value_decl`
- for only the type params of type declarations within `transl_type_decl`

### DIFF
--- a/ocaml/testsuite/tests/typing-layouts-err-msg/error_messages.ml
+++ b/ocaml/testsuite/tests/typing-layouts-err-msg/error_messages.ml
@@ -22,7 +22,7 @@ Error: This expression has type int -> int
        The layout of int -> int is value, because
          it's a function type.
        But the layout of int -> int must be a sublayout of void, because
-         of the annotation on 'a in the declaration of the type t.
+         of definition of t at Line 1, characters 0-22.
 |}]
 
 
@@ -40,7 +40,7 @@ Error: This expression has type [ `A | `B ]
        The layout of [ `A | `B ] is immediate, because
          it's an enumeration variant (all constructors are constant).
        But the layout of [ `A | `B ] must be a sublayout of void, because
-         of the annotation on 'a in the declaration of the type t.
+         of definition of t at Line 1, characters 0-22.
 |}]
 
 
@@ -58,7 +58,7 @@ Error: This expression has type [ `A of int | `B ]
        The layout of [ `A of int | `B ] is value, because
          it's a polymorphic variant.
        But the layout of [ `A of int | `B ] must be a sublayout of void, because
-         of the annotation on 'a in the declaration of the type t.
+         of definition of t at Line 1, characters 0-22.
 |}]
 
 
@@ -78,7 +78,7 @@ Error: This expression has type v but an expression was expected of type
        The layout of v is immediate, because
          it's an enumeration variant (all constructors are constant).
        But the layout of v must be a sublayout of void, because
-         of the annotation on 'a in the declaration of the type t.
+         of definition of t at Line 1, characters 0-22.
 |}]
 
 (* Extensible_variant *)
@@ -97,7 +97,7 @@ Error: This expression has type attr but an expression was expected of type
        The layout of attr is value, because
          it's an extensible variant.
        But the layout of attr must be a sublayout of void, because
-         of the annotation on 'a in the declaration of the type t.
+         of definition of t at Line 1, characters 0-22.
 |}]
 
 (* First_class_module *)
@@ -118,7 +118,7 @@ Error: This expression has type (module X_int)
        The layout of (module X_int) is value, because
          it's a first-class module type.
        But the layout of (module X_int) must be a sublayout of void, because
-         of the annotation on 'a in the declaration of the type t.
+         of definition of t at Line 1, characters 0-22.
 |}]
 
 (* Match *)
--- a/ocaml/testsuite/tests/typing-layouts-err-msg/error_messages.ml
+++ b/ocaml/testsuite/tests/typing-layouts-err-msg/error_messages.ml
@@ -76,7 +76,7 @@ Line 3, characters 21-22:
 Error: This expression has type v but an expression was expected of type
          'a t = ('a : void)
        The layout of v is immediate, because
-         it's an enumeration variant (all constructors are constant).
+         of the definition of v at Line 2, characters 0-10.
        But the layout of v must be a sublayout of void, because
          of the definition of t at Line 1, characters 0-22.
 |}]
@@ -95,7 +95,7 @@ Line 3, characters 24-25:
 Error: This expression has type attr but an expression was expected of type
          'a t = ('a : void)
        The layout of attr is value, because
-         it's an extensible variant.
+         of the definition of attr at Line 2, characters 0-14.
        But the layout of attr must be a sublayout of void, because
          of the definition of t at Line 1, characters 0-22.
 |}]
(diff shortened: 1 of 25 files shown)


ocaml/testsuite/tests/typing-layouts-err-msg/test.ml
```diff
@@ -178,3 +178,80 @@ Error: This expression has type t_void but an expression was expected of type
...
+9 | end
+Error: Signature mismatch:
+       Modules do not match: sig type 'a t = 'a end is not included in S2
+       Type declarations do not match:
+         type 'a t = 'a
+       is not included in
+         type ('a : void) t = 'a
+       The type ('a : value) is not equal to the type ('a0 : void)
+       because their layouts are different.
+       The layout of 'a is value, because
```
goldfirere, line 253: I think in a module mismatch error, the histories are always going to be boring. Can we leave them off here?

ocaml/testsuite/tests/typing-layouts/modules_alpha.ml
```diff
@@ -351,7 +351,7 @@ Error: This type M4'.s should be an instance of type ('a : void)
        The layout of M4'.s is immediate, because
          of the annotation on the declaration of the type t.
```
goldfirere, line 352: I'm surprised here. I would expect a reference to the definition of `s`, above. I think there may be a call to generalize that you're missing.

ocaml/testsuite/tests/typing-layouts-err-msg/error_messages.ml
```diff
@@ -22,7 +22,7 @@ Error: This expression has type int -> int
        The layout of int -> int is value, because
          it's a function type.
        But the layout of int -> int must be a sublayout of void, because
-         of the annotation on 'a in the declaration of the type t.
+         of definition of t at Line 1, characters 0-22.
```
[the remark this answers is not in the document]
alanechang: Can be done by adding a `print_loc_in_lowercase` function to `location.ml`:

https://github.com/ocaml-flambda/flambda-backend/blob/d45c7e4b3f6813eef980fd906c2f80d2d389a634/ocaml/parsing/location.ml#L280-L281

Should such a change be made?

ocaml/testsuite/tests/typing-layouts-err-msg/test.ml
alanechang: That's a good idea, but don't see a straightforward way of implementing it. There has to be some kind of global state to toggle history on/off or it has to be threaded through all the recursive functions in `includemod_errorprint.ml`.

ocaml/testsuite/tests/typing-layouts-err-msg/error_messages.ml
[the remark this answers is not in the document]
goldfirere: Yes, this sounds worthwhile.

ocaml/testsuite/tests/typing-layouts-err-msg/test.ml
goldfirere: OK.

ocaml/testsuite/tests/typing-layouts-err-msg/error_messages.ml
```diff
@@ -76,28 +76,26 @@ Line 3, characters 21-22:
 Error: This expression has type v but an expression was expected of type
          'a t = ('a : void)
        The layout of v is immediate, because
-         it's an enumeration variant (all constructors are constant).
+         of the definition of v at line 2, characters 0-10.
```
goldfirere, line 79: This is not a particularly helpful message, precisely because the type about which we are reporting is just a plain old `Tconstr`. Is there a way to detect such cases and avoid printing the history here? Doesn't have to be in this PR, but perhaps note this down as a future possible improvement.

ocaml/testsuite/tests/typing-layouts-err-msg/error_messages.ml
```diff
@@ -76,28 +76,26 @@ Line 3, characters 21-22:
...
        But the layout of v must be a sublayout of void, because
-         of the annotation on 'a in the declaration of the type t.
+         of the definition of t at line 1, characters 0-22.
 |}]
 
 (* Extensible_variant *)
 type ('a: void) t = 'a
 type attr = ..
-let f (x: attr): 'a t = x
+and w = attr t
```
goldfirere, line 87: Why change this test?

ocaml/typing/typedecl.ml
```diff
@@ -1167,6 +1167,25 @@ let update_decl_layout env dpath decl =
...
+let update_decls_layout_reason decls =
+  List.map
+    (fun (id, decl) ->
+       let reason =  Layout.(Generalized (Some id, decl.type_loc)) in
+       List.iter
+         (fun ty -> Ctype.update_generalized_ty_layout_reason ty reason)
+         decl.type_params;
+       Btype.iter_type_expr_kind
+         (fun ty -> Ctype.update_generalized_ty_layout_reason ty reason) decl.type_kind;
+       begin match decl.type_manifest with
```
goldfirere, line 1179: Use `Option.iter` here.

goldfirere (approved):

ocaml/testsuite/tests/typing-layouts-err-msg/error_messages.ml
alanechang: Motivation is to keep the `it's an extensible variant` error message.

With the original test, the layout reason gets hidden:
```
Error: This expression has type attr but an expression was expected of type
         'a t = ('a : void)
       The layout of attr is value, because
         of the definition of attr at file "./debug_test.ml", line 246, characters 0-14.
       But the layout of attr must be a sublayout of void, because
         of the definition of t at file "./debug_test.ml", line 245, characters 0-22.
```

ocaml/testsuite/tests/typing-layouts-err-msg/error_messages.ml
alanechang: Opened a jira ticket to record this

ocaml/testsuite/tests/typing-layouts/modules_alpha.ml
goldfirere: Yes, that's unfortunate (that is, the hiding of more information). But it's internally consistent. And I claim that the right way to tell the reason for the layout of a well-typed definition is to use merlin or some other editor integration. So: OK.

PR #1890, opened 2023-10-02
Remove Jkind.of_sort

goldfirere:
We shouldn't ever really have to convert a sort back into a layout or kind. And it turns out that almost all uses of `of_sort` worked on a freshly created sort variable. Because converting a sort into a layout/kind is suspect, this refactor makes it harder to do so. Sadly, in one case, the conversion really is the easiest way to construct the right error message, so I kept that one usage of `of_sort`.

This is just a refactoring; no change in functionality, so no testcase changes.


ccasin (approved):
Great!

I've been meaning to go back and factor out the "create sort var, create type var" pattern since #1793.  Thanks for doing it!

PR #1891, opened 2023-10-02
Add a missing check for jane syntax

ccasin:
This adds a missing check for jane syntax in the parser where we inspect an expression to see if it looks like a big array index.  

I don't think this can really cause a problem today, because the underlying expression would need to be an int or a tuple to do something surprising.  But as it happens both unboxed constants and labeled tuples are coming soon, and it will be a problem then.


ncik-roberts:
This is one of those places where shadowing `Bigarray` can change what function is called at runtime, right? Seems like it:

```ocaml
# let f x = x.{3};;
val f : ('a, 'b, 'c) Bigarray.Array1.t -> 'a = <fun>
# module Bigarray = struct end;;
module Bigarray : sig end
# let f x = x.{3};;
Error: Unbound module Bigarray.Array1
```

I'm not claiming this is used, but if it is, then there are reasons that people might want to pass (say) an unboxed float as the first index. Should we default to the `[ expr ]` behavior instead of raising an error in the parser?

ccasin:
Yes, that seems better - will do tomorrow.  And I was wrong that unboxed ints are problematic here - they would be a type error with the old behavior, but not miscompilation or anything.

ccasin:
@ncik-roberts OK, I've made this a much smaller PR that just does the thing you suggest.  I haven't added a test, because there aren't any current examples where this changes the behavior of the compiler.  But I will add a test in the labeled tuples branch, where I am copying this change.

ncik-roberts (approved):
Seems right to me, in light of labeled tuples incoming.

PR #1895, opened 2023-10-03
Abbreviate module types when printing error messages

rleshchinskiy:
Module types can become huge and pretty-printing them in error messages seems to be very slow. In one case, we produced a ~400k lines error message which took about 4 minutes to generate.

This PR caps the overall number of signature elements we print and the depth to which we unfold module types. In my somewhat limited testing, this seemed to be the best heuristic I tried, although it is far from perfect. We can adjust later based on user feedback.

For the one file in question, this reduced the compile time to ~13s.

### DIFF
--- a/ocaml/testsuite/tests/typing-modules/functors.ml
+++ b/ocaml/testsuite/tests/typing-modules/functors.ml
@@ -921,37 +921,14 @@ Error: Signature mismatch:
          sig
            module B :
              sig
-               module C :
-                 sig
-                   module D :
-                     sig
-                       module E :
-                         sig
-                           module F :
-                             functor (X : sig type x end)
-                               (Y : sig type y' end) (W : sig type w end) ->
-                               sig end
-                         end
-                     end
-                 end
+               module C : sig module D : sig module E : sig ... end end end
              end
          end
        is not included in
          sig
            module B :
              sig
-               module C :
-                 sig
-                   module D :
-                     sig
-                       module E :
-                         sig
-                           module F :
-                             sig type x end -> sig type y end ->
-                               sig type z end -> sig type w end -> sig end
-                         end
-                     end
-                 end
+               module C : sig module D : sig module E : sig ... end end end
              end
          end
        In module B:
@@ -966,8 +943,8 @@ Error: Signature mismatch:
                    module E :
                      sig
                        module F :
-                         sig type x end -> sig type y end ->
-                           sig type z end -> sig type w end -> sig end
+                         sig ... end -> sig ... end -> sig ... end ->
+                           sig ... end -> sig end
                      end
                  end
              end
--- a/ocaml/testsuite/tests/typing-modules/illegal_permutation.ml
+++ b/ocaml/testsuite/tests/typing-modules/illegal_permutation.ml
@@ -502,22 +502,7 @@ Error: Signature mismatch:
                      sig
                        module C :
                          functor (X : sig end) (Y : sig end)
-                           (Z : sig
-                                  module D :
-                                    sig
-                                      module E :
-                                        sig
-                                          module F :
-                                            functor (X : sig end)
-                                              (Arg : sig
-                                                       val two : int
-                                                       val one : int
-                                                     end)
-                                              -> sig end
-                                        end
-                                    end
-                                end)
-                           -> sig end
+                           (Z : sig ... end) -> sig end
                      end
                  end
              end
@@ -532,22 +517,7 @@ Error: Signature mismatch:
                      sig
                        module C :
                          functor (X : sig end) (Y : sig end)
-                           (Z : sig
-                                  module D :
-                                    sig
-                                      module E :
-                                        sig
-                                          module F :
-                                            functor (X : sig end)
-                                              (Arg : sig
-                                                       val one : int
-                                                       val two : int
-                                                     end)
-                                              -> sig end
-                                        end
-                                    end
-                                end)
-                           -> sig end
+                           (Z : sig ... end) -> sig end
                      end
                  end
              end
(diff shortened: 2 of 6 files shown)


goldfirere (commented):
Took a look at the testsuite changes and thought it best to review those before diving into implementation. In short: I'm worried this change makes some messages un-actionable, and so I'm unconvinced.

Maybe a better way to reduce the compilation time is to speed up the pretty printer? I know that's not so easy, but even 13s is too long for a file, and so this patch doesn't really solve the problem.

ocaml/testsuite/tests/typing-modules/functors.ml
```diff
@@ -966,8 +943,8 @@ Error: Signature mismatch:
                    module E :
                      sig
                        module F :
-                         sig type x end -> sig type y end ->
-                           sig type z end -> sig type w end -> sig end
+                         sig ... end -> sig ... end -> sig ... end ->
```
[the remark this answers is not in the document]
rleshchinskiy: I don't really understand the objection. All these message have the form "These types don't match because these types don't match because these types don't match etc." Really, the only actionable part of the error message is the last one (way further below) which doesn't change with this PR. What useful information does the fully unfolded type of `C` provide in this particular spot?

ocaml/testsuite/tests/typing-modules/illegal_permutation.ml
```diff
@@ -502,22 +502,7 @@ Error: Signature mismatch:
...
-                                              (Arg : sig
-                                                       val two : int
-                                                       val one : int
-                                                     end)
-                                              -> sig end
-                                        end
-                                    end
-                                end)
-                           -> sig end
+                           (Z : sig ... end) -> sig end
```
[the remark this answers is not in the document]
rleshchinskiy: Again, could you please explain why. I find the original error message basically unreadable.

rleshchinskiy:
> I know that's not so easy, but even 13s is too long for a file, and so this patch doesn't really solve the problem.

The 13s aren't spent in the pretty-printer. I agree that it might be a problem but it's an entirely different (and a lot less pressing) one.

ocaml/typing/printtyp.ml
```diff
@@ -2065,17 +2065,68 @@ let expand_module_type =
...
+    | Some t ->
+        let n = List.length sg in
+        let k = min t.width n in
+        t.width <- t.width - n;
+        Some k, (k < n)
+    | None ->
+        None, false
+end
+
+let rec tree_of_modtype ?abbrev = function
```
goldfirere, line 2120: It just comes down to taste, but I tend to lean away from optional parameters, unless the omitted form is much more common than the explicit form -- not the case here.

ocaml/typing/printtyp.ml
```diff
@@ -2102,30 +2153,52 @@ and tree_of_functor_parameter = function
...
+      wrap_env (fun env -> env)(fun sg ->
+        let max_items, trimmed = Abbrev.items abbrev sg in
+        let tree_groups = tree_of_signature_rec ?abbrev ?max_items !printing_env sg in
+        let items = List.concat_map (fun (_env,l) -> List.map snd l) tree_groups in
+        if trimmed then items @ [Osig_ellipsis] else items
+        ) sg
+    )
 
-and tree_of_signature_rec env' sg =
+and tree_of_signature_rec ?abbrev ?max_items env' sg =
```
goldfirere, line 2171: Ditto here -- I think this would be clearer with `~max_items`.

ocaml/typing/printtyp.ml
```diff
@@ -2102,30 +2153,52 @@ and tree_of_functor_parameter = function
...
+                | n, x :: xs -> take (n-1) (x :: acc) xs
+              in
+              let n, group_trees = take n [] group_trees in
+              Some n, group_trees
+        in
+        max_items, (env, group_trees)
+  in
   set_printing_env env';
-  List.map collect_trees_of_rec_group structured
+  snd (List.fold_left_map collect_trees_of_rec_group max_items structured)
```
goldfirere, line 2197: It looks to me like this could just use `Abbrev.items` to do its work, no? That is, instead of peeling off some allocation into `max_items` just to let that dwindle while iterating, peel off the items you need as you need them. That strikes me as simpler. Or maybe I'm missing something here?

ocaml/typing/printtyp.ml
```diff
@@ -2102,30 +2153,52 @@ and tree_of_functor_parameter = function
...
       in
-      Some (name, tree_of_modtype ~ellipsis:false ty_arg), env
+      Some (name, tree_of_modtype ?abbrev ty_arg), env
 
-and tree_of_signature sg =
-  wrap_env (fun env -> env)(fun sg ->
-      let tree_groups = tree_of_signature_rec !printing_env sg in
-      List.concat_map (fun (_env,l) -> List.map snd l) tree_groups
-    ) sg
+and tree_of_signature ?abbrev = function
```
goldfirere, line 2158: This area could have a brief comment outlining the algorithm. After staring for a little while, I figured it out, but (fairly simple though it is) it wasn't immediately apparent to me.

goldfirere (approved):
A few style comments. Otherwise looks good. I think this is a step in the right direction.

ocaml/typing/printtyp.ml
rleshchinskiy: There are quite a few calls further down in the module which would have to be changed. I lean towards making diffs as small as possible to make upstream merging easier but happy to change this if you feel more strongly about this than I do.

ocaml/typing/printtyp.ml
rleshchinskiy: Again, there is a call (just one) further down which would have to be changed. Happy to do that if you think it's worthwhile.

ocaml/typing/printtyp.ml
rleshchinskiy: IIUC, what you suggest would expand things depth first. So given:
```
sig
  module M : sig <big> end
  <small>
end
```
we currently get
```
sig
  module M: sig <partial> ... end
  <small>
end
```
because we "pre-allocate" the items before descending. But if we didn't, we'd get
```
sig
  module M : sig <big> end
  ...
end
```
IMO the former is better (but shout if you disagree!). I added a comment.

ocaml/typing/printtyp.ml
goldfirere: No, I don't feel strongly.

ocaml/typing/printtyp.ml
goldfirere: OK

PR #1897, opened 2023-10-03
Runtime helpers for 128-bit vectors

TheNumbat:
Adds runtime definitions for loading/storing/copying boxed 128-bit SIMD vectors.
Should make writing C bindings using SIMD vectors slightly less error-prone, since users need to remember to use unaligned load/store intrinsics.


mshinwell:
I'll review this.

mshinwell (approved):

PR #1899, opened 2023-10-03
Make `assert false` behave as local_ or not, depending on what's better (+ 2 bugfixes)

goldfirere:
In work meant to support unboxed types, I recently updated a ppx to produce `if false then (* stuff to improve behavior of warnings *); assert false else (* the goods *)`. But when `(* the goods *)` has a `local_` in tail position, the compiler complains that the `assert false` branch doesn't have a `local_` in tail position. This is a bit silly, really: if `assert false` has magical typing, it may as well have some magical moding too. (Well, this isn't really about moding, exactly, but just go with it.)

So I added this cleverness to `assert false`.

Along the way, a few other things happened:
* I had to refactor `is_local_returning_function` to get this to work. Because there's a largish chunk of code there with one well-defined entry point, I wrapped it and friends into a module.
* I noticed that the treatment for `function` wasn't quite right: it just combined all the cases without checking that all the cases either all have are all don't have a `local_` annotation. I wasn't able to get this to misbehave, really, just to produce a bad error message. In any case, fixed now.
* I also noticed that a newtype with a kind annotation got in the way of the look for a `local_`. That's just wrong. Fixed now.

If we want the bugfixes without the new `assert false` feature, I can move the fixes out. But it was more efficient to do this all at once.

### DIFF
--- a/ocaml/testsuite/tests/typing-local/local-layouts.ml
+++ b/ocaml/testsuite/tests/typing-local/local-layouts.ml
@@ -8,10 +8,5 @@ let bar _t (type a : value) = local_ 2
 
 [%%expect{|
 val foo : 'a -> local_ int = <fun>
-Line 2, characters 30-38:
-2 | let bar _t (type a : value) = local_ 2
-                                  ^^^^^^^^
-Error: This local value escapes its region
-  Hint: Cannot return local value without an "exclave_" annotation
+val bar : 'a -> local_ int = <fun>
 |}]
-(* that's just wrong; the layout annotation shouldn't affect locality *)
--- a/ocaml/testsuite/tests/typing-local/local-layouts.ml
+++ b/ocaml/testsuite/tests/typing-local/local-layouts.ml
@@ -1,0 +1,17 @@
+(* TEST
+   * expect
+   flags = "-extension layouts_beta"
+*)
+
+let foo _t (type a) = local_ 1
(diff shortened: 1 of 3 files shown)


riaqn:
Thanks for the fixes, I like the refactoring as well.

WRT the special case of `assert false`, I wonder if it's really special. The `local_` in function's tail position indicates that the function won't have a region. The expression following `local_`, whether it is `global` or `local`, is irrelavent. 

I think really `local_ assert false` is what we are going to write here. But to make our life easier, I also don't see anything particularly wrong, if we take an expression to be local-returning, if any of the branches are local-returning.

goldfirere:
The challenge is that it's a ppx writing the `assert false`. And I'd rather not have to write a check in the ppx looking for other `local_` in tail position to determine whether to put one before the `assert false`. It's true that we could just check for `local_` in one tail position and call that enough, but I think that's harder to reason about -- I like the current design requiring it everywhere. It's just that `assert false` is already special (`assert false : 'a`, whereas `assert anything_else : unit`) and so this just makes it more special.

riaqn:
`assert false` can be typed as any `'a` because it never returns. `assert exp` can also be moded as any mode because it either never returns, or returns `unit` which crosses modes. However, the tail-position `local_` has nothing to do with any of those. 

But I think in the end it doesn't matter that much - approving.

riaqn (approved):

PR #1900, opened 2023-10-03
Add boxed_float variant to all float operations

TheNumbat:
Renames `Lambda.Pfloatval` to `Pboxedfloatval of boxed_float` and uses the `boxed_float` variant everywhere that previously assumed float64. 

The following PR adds Pfloat32 to this variant.


mshinwell:
Is it possible to pull out "Renames Lambda.Pfloatval to Pboxedfloatval of boxed_float and uses the boxed_float variant everywhere that previously assumed float64" into a separate PR, to reduce the size of this one?

TheNumbat:
I could, but I don't think it would help much - almost all of the diff here is the `boxed_float` change

ocaml/testsuite/tests/translprim/comparison_table.heap.reference
```diff
@@ -96,161 +103,195 @@
        (function {nlocal = 0} x[nativeint] y[nativeint] : int
          (Nativeint.>= x y))
      eta_gen_cmp =
-       (function {nlocal = 0} prim prim stub (caml_compare prim prim))
+       (function {nlocal = 0} prim prim stub ignore assert all zero_alloc
+         : int (caml_compare prim prim))
      eta_int_cmp =
        (function {nlocal = 0} prim[int] prim[int] stub
-         (compare_ints prim prim))
+         ignore assert all zero_alloc : int (compare_ints prim prim))
```
ccasin, line 110: What is the source of this diff?

ocaml/testsuite/tests/translprim/module_coercion.compilers.no-flat.reference
```diff
@@ -31,13 +31,20 @@
...
-        (function {nlocal = 0} prim[float] prim[float] stub (==. prim prim))
-        (function {nlocal = 0} prim[float] prim[float] stub (!=. prim prim))
-        (function {nlocal = 0} prim[float] prim[float] stub (<. prim prim))
-        (function {nlocal = 0} prim[float] prim[float] stub (>. prim prim))
-        (function {nlocal = 0} prim[float] prim[float] stub (<=. prim prim))
-        (function {nlocal = 0} prim[float] prim[float] stub (>=. prim prim)))
+          ignore assert all zero_alloc : int
+          (compare_floats float prim prim))
+        (function {nlocal = 0} prim[float] prim[float] stub
+          ignore assert all zero_alloc : int (Float.==. prim prim))
```
ccasin, line 37: More mysterious (to me) printing diffs.

ocaml/typing/primitive.ml
```diff
@@ -323,6 +325,11 @@ let equal_boxed_integer bi1 bi2 =
   | (Pnativeint | Pint32 | Pint64), _ ->
     false
 
+let equal_boxed_float f1 f2 =
+  match f1, f2 with
+  | Pfloat64, Pfloat64 ->
+    true
```
ccasin, line 331: I suspect this fits on the previous line.

ccasin (approved):
I read `ocaml/typing`, `ocaml/lambda` and `ocaml/testsuite`.  Looks basically fine - some minor questions / comments.

ocaml/lambda/printlambda.ml
```diff
@@ -102,7 +105,7 @@ let variant_kind print_contents ppf ~consts ~non_consts =
 let rec value_kind ppf = function
   | Pgenval -> ()
   | Pintval -> fprintf ppf "[int]"
-  | Pfloatval -> fprintf ppf "[float]"
+  | Pboxedfloatval Pfloat64 -> fprintf ppf "[float]"
```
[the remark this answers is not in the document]
TheNumbat: They should

ocaml/lambda/printlambda.ml
```diff
@@ -376,19 +386,19 @@ let primitive ppf = function
...
-  | Pdivfloat m -> fprintf ppf "/.%s" (alloc_kind m)
-  | Pfloatcomp(cmp) -> float_comparison ppf cmp
+  | Pintoffloat Pfloat64 -> fprintf ppf "int_of_float"
+  | Pfloatofint (Pfloat64,m) -> fprintf ppf "float_of_int%s" (alloc_kind m)
+  | Pabsfloat (bf,m) -> print_boxed_float "absf" ppf bf m
+  | Pnegfloat (bf,m) -> print_boxed_float "negf" ppf bf m
+  | Paddfloat (bf,m) -> print_boxed_float "addf" ppf bf m
+  | Psubfloat (bf,m) -> print_boxed_float "subf" ppf bf m
+  | Pmulfloat (bf,m) -> print_boxed_float "mulf" ppf bf m
+  | Pdivfloat (bf,m) -> print_boxed_float "divf" ppf bf m
```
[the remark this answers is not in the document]
TheNumbat: I preferred these names because they're now prefixed with `Float.`, like the Int64/Int32/etc functions. Removed the trailing `f`s, and also the dots from `float_comparison`.

ocaml/lambda/printlambda.ml
```diff
@@ -511,8 +521,8 @@ let primitive ppf = function
   | Pprobe_is_enabled {name} -> fprintf ppf "probe_is_enabled[%s]" name
   | Pobj_dup -> fprintf ppf "obj_dup"
   | Pobj_magic _ -> fprintf ppf "obj_magic"
-  | Punbox_float -> fprintf ppf "unbox_float"
-  | Pbox_float m -> fprintf ppf "box_float%s" (alloc_kind m)
+  | Punbox_float _ -> fprintf ppf "unbox_float"
+  | Pbox_float (bf,m) -> fprintf ppf "box_%s%s" (boxed_float_name bf) (alloc_kind m)
```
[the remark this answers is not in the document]
TheNumbat: Yes

ocaml/lambda/printlambda.ml
```diff
@@ -241,17 +251,17 @@ let integer_comparison ppf = function
...
+  | CFeq -> "==."
+  | CFneq -> "!=."
+  | CFlt -> "<."
+  | CFnlt -> "!<."
+  | CFle -> "<=."
+  | CFnle -> "!<=."
+  | CFgt -> ">."
+  | CFngt -> "!>."
+  | CFge -> ">=."
+  | CFnge -> "!>=."
```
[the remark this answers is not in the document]
TheNumbat: Hmm, tried that but it didn't seem any better

ocaml/lambda/translprim.ml
```diff
@@ -660,55 +660,55 @@ let comparison_primitive comparison comparison_kind =
   match comparison, comparison_kind with
   | Equal, Compare_generic -> Pccall caml_equal
   | Equal, Compare_ints -> Pintcomp Ceq
-  | Equal, Compare_floats -> Pfloatcomp CFeq
+  | Equal, Compare_floats -> Pfloatcomp (Pfloat64, CFeq)
```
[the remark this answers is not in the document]
TheNumbat: Yeah, the next PR adds `Compare_float32s`. Possibly that should use a flag instead, but the `comparison_kind` already splits out `Compare_int32s` and `Compare_int64s`.

ocaml/testsuite/tests/translprim/comparison_table.heap.reference
TheNumbat: It appears this was changed a while back: https://github.com/ocaml-flambda/flambda-backend/blame/main/ocaml/lambda/lambda.ml#L738

But somehow this test output was never regenerated?

ocaml/testsuite/tests/translprim/module_coercion.compilers.no-flat.reference
TheNumbat: See above

ocaml/testsuite/tests/translprim/comparison_table.heap.reference
ccasin: OK, I dug into this a little more.  The two files that changed are for the no-stack-allocation and no-flat-float-array configurations, which we don't actually build or test anymore with the full testsuite, but which can get run if you build an individual test.  Presumably you regenerated the output of these tests for other reasons, and that regenerated it for all configurations, even those we don't use.  Seems fine.

mshinwell:
One thing I would suggest doing here is looking through the diff (in an editor where the diff updates live) and ensuring that every match on the affected primitives says `Pfloat64` rather than `_`.  This should help with the follow-up PR to ensure that all places get updated -- one can just grep for `Pfloat64`.  It isn't possible to rely on warning 4 because it's often disabled, and there are a lot of underscore pattern matches.

In this PR many of the places that need changing should have been caught by virtue of a type error when a primitive went from zero to one argument.  Unfortunately I think for the ones that already had an argument, an existing underscore match could potentially conceal a place that needs editing.  However I think it's worth finding all of those in this PR and doing the above, so that the next one is easier.

TheNumbat:
Ok, I've converted all boxed_float wildcards to `Pfloat64`. I think most of them are just going to get changed back in #1913, though.

TheNumbat:
For some reason the merge caused a huge spurious diff in `.github/workflows/build.yml` - there are no actual changes from main.

mshinwell:
I haven't read `testsuite/` and `typing/` (but @ccasin has), everything else looks ok.

mshinwell (approved):

PR #1901, opened 2023-10-04
Remove no_flat_float_array from the compiler config

poechsel:
The compiler can't build when this feature is on.

This should be turned back on once we get it to work again


mshinwell:
This isn't needed now, #1894 fixes the `no_flat_float_array` build.

PR #1902, opened 2023-10-05
Register allocation: more flexible handling of substitutions

xclerc:
As noted in https://github.com/ocaml-flambda/flambda-backend/pull/1879, moving spills/reloads
from one block to another leads to invalid
results. Not because the moves were
problematic per se, but because a later
phase computing the substitutions for
the basic blocks relied on the fact that
reloads (and phi) were "all-or-nothing",
i.e. all live names (or none) would be
reloaded at a given point.

This pull request changes the way the
substitutions are computed. We now
iterate on the dominators tree, and
keep bindings for the various registers
as we progress from the root to the
leaves. A new name is introduced when
we see a definition, and old names
are forgotten when we see a phi
or a destruction.

### DIFF
--- a/backend/regalloc/regalloc_split.ml
+++ b/backend/regalloc/regalloc_split.ml
@@ -5,76 +5,76 @@ open! Regalloc_split_utils
 module DLL = Flambda_backend_utils.Doubly_linked_list
 module State = Regalloc_split_state
 
-(* [propagate_substitution state cfg substs start subst] propagates the
-   substitution [subst], from block [start] until new ones are defined.
-
-   In practice, it means that we stop propagating when we encounter a node with
-   either "definitions at beginning" or a "phi" because: - when there are new
-   definitions, they follow a destruction point, which means that every live
-   register will get a new name (thus replacing the binding in the current
-   substitution); - when there is a phi, it means that every register in the
-   substitution we are currently propagating is either no longer live, or needs
-   a phi move (thus replacing the binding in the current substitution).
-
-   The effect of the propagation is the addition of [subst] to the relevant
-   blocks in [substs]. *)
-let propagate_substitution :
+(* [remove_from_bindings state label ~field ~extract bindings] removes registers
+   in a set from [bindings]. The set is determined by (i) calling [field] on
+   [state] in order to get a map, (ii) retrieving the data associated with
+   [label] from this map, and (iii) calling [extract] on the data. *)
+let[@inline] remove_from_bindings :
     State.t ->
-    Cfg.t ->
-    Substitution.map ->
-    Cfg.basic_block ->
-    Substitution.t ->
-    unit =
- fun state cfg substs start subst ->
-  if split_debug then log ~indent:1 "propagate_substitution";
-  let visited = ref Label.Set.empty in
-  let rec iter (block : Cfg.basic_block) : unit =
-    if not (Label.Set.mem block.start !visited)
-    then (
-      if split_debug then log ~indent:2 "iter %d" block.start;
-      visited := Label.Set.add block.start !visited;
-      if Label.Tbl.mem substs block.start
-      then fatal "block %d already has a substitution" block.start;
-      Label.Tbl.replace substs block.start subst;
-      match destruction_point_at_end block with
-      | Some Destruction_on_all_paths -> ()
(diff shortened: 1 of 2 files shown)


backend/regalloc/regalloc_split.ml
```diff
@@ -5,93 +5,107 @@ open! Regalloc_split_utils
...
- fun state cfg_with_infos ->
-  if split_debug then log ~indent:0 "compute_substitutions";
-  let cfg = Cfg_with_infos.cfg cfg_with_infos in
-  let substs = Label.Tbl.create (Label.Tbl.length cfg.blocks) in
-  let compute_substitution_for_block label =
-    if split_debug then log ~indent:1 "visiting block %d" label;
+  if split_debug then log ~indent:2 "removing from definitions";
+  let bindings =
+    remove_from_bindings state label ~field:State.definitions_at_beginning
+      ~extract:Fun.id bindings
```
gretay-js, line 55: In what situation does this line have any effect on bindings? I mean, bindings are a result of a reload in a preceeding block, is it possible to have a reload again, without a spill between them on the same path  (that would have killed the binding) and without a the current block havig a phi too? This is safe to keep, just trying to confirm my understanding of what this code is doing.

backend/regalloc/regalloc_split.ml
```diff
@@ -5,93 +5,107 @@ open! Regalloc_split_utils
...
+  Label.Tbl.replace substs label subst;
+  (* Second, add the bindings to the substitution. *)
+  if split_debug then log ~indent:2 "adding from bindings";
+  Reg.Map.iter
+    (fun old_reg new_reg ->
+      if split_debug
+      then log ~indent:3 "%a -> %a" Printmach.reg old_reg Printmach.reg new_reg;
+      Reg.Tbl.replace subst old_reg new_reg)
+    bindings;
+  (* Third, add the definitions to the substitution. *)
```
gretay-js, line 67:
```suggestion
  (* Third, add the definitions to the substitution and bindings. *)
```

gretay-js (approved):

backend/regalloc/regalloc_split.ml
xclerc: Yes, it it possible to have two reloads in a row.
_i.e._ with neither a spill or a phi. Spills may
disappear since we delete dominated spills of
constant values. If that happens on straight-line
code, there could be no phi as well.

backend/regalloc/regalloc_split_state.ml
```diff
@@ -631,6 +621,19 @@ let make cfg_with_infos ~next_instruction_id =
...
+      (fun _ set -> not (Reg.Set.is_empty set))
+      definitions_at_beginning
+  in
+  let destructions_at_end =
+    Label.Map.filter
+      (fun _ (_, set) -> not (Reg.Set.is_empty set))
+      destructions_at_end
+  in
+  let phi_at_beginning =
+    compute_phis cfg_with_infos ~destructions_at_end ~definitions_at_beginning
```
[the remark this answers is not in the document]
xclerc: The previous version was safe but suboptimal
(generating too many phi, caused by definitions
and destructions that would later be deleted).
Since we had to reason again about names for
this PR, it felt reasonable to treat that minor issue
as well.

backend/regalloc/regalloc_split_state.ml
```diff
@@ -631,6 +621,19 @@ let make cfg_with_infos ~next_instruction_id =
...
+  let definitions_at_beginning =
+    Label.Map.filter
+      (fun _ set -> not (Reg.Set.is_empty set))
+      definitions_at_beginning
+  in
+  let destructions_at_end =
+    Label.Map.filter
+      (fun _ (_, set) -> not (Reg.Set.is_empty set))
+      destructions_at_end
+  in
```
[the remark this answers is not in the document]
xclerc: They don't, unfortunately, but I will do what
I ended up doing in the other file with the function
named `remove_from_bindings`.

backend/regalloc/regalloc_split.ml
gretay-js: ah, yes, the constant spills. Worth a comment.

PR #1905, opened 2023-10-05
Add `-instantiate`

lukemaurer:
This is the last piece of compiler support for parameterised libraries: the ability to take a parameterised module and a set of arguments and produce an instance module. The `-instantiate` option is similar to `-pack` in that it specifies a special compilation mode that takes in .cmo or .cmx files and produces a new one from them. It takes the module to instantiate and the modules to provide as instance arguments.

One bit of awkwardness is that the `-o` option is required, but (up to directory) there's no actual choice: the output module must have the correct form to be usable from other modules so that .cmx lookup can function. This would be somewhat awkward to fix, and it's not worth doing so at the moment since we don't intend for humans to be running `-instantiate` by hand.


testsuite/tests/templates/basic/bad_instantiate_repeated_param.reference
```diff
@@ -0,0 +1,4 @@
+File "_none_", line 1:
+Error: Float_list_element and Int_list_element
+       were both compiled with -as-argument-for List_element.
+       Only one argument may be given for each parameter.
```
riaqn, line 4: > Error: Cannot supply both Float_list_element and Int_list_element
       as arguments for the parameter List_element.
    Hint: Both were compiled with `-as-argument-for List_element`.

testsuite/tests/templates/basic/bad_instantiate_missing_arg.reference
```diff
@@ -0,0 +1,2 @@
+File "_none_", line 1:
+Error: No argument given for parameter Semigroup
```
riaqn, line 2: > Error: Instantiation of Int_list_element is missing an argument for parameter Semigroup.

testsuite/tests/templates/basic/bad_instantiate_no_such_param.reference
```diff
@@ -0,0 +1,4 @@
+File "_none_", line 1:
+Error: Mismatched argument: Int_list_element
+       was compiled with -as-argument-for List_element
+       but Monoid_of_semigroup was not compiled with -parameter List_element
```
riaqn, line 4: > Error: `Int_list_element` is an argument for parameter List_element, which is not a parameter of Monoid_of_semigroup.
> Hint: `int_list_element.cmi` was compiled with `-as-argument-for List_element`
> Hint: parameters of Monoid_of_semigroup: XXX

testsuite/tests/templates/basic/bad_instantiate_not_arg.reference
```diff
@@ -0,0 +1,2 @@
+File "_none_", line 1:
+Error: Argument must be compiled with -as-argument-for: Widget
```
riaqn, line 2: > Error: Module `Widget` cannot be used as argument for instantiation of `List_monoid`.
> Hint: Compile `Widget.cmi` with `-as-argument-for Foo` where `Foo` is a parameter of `List_monoid`.
> Hint: parameters of `List_monoid`: XXX

driver/instantiator.ml
```diff
@@ -0,0 +1,184 @@
...
+      error (Repeated_parameter { param = key; arg1; arg2 })
+  in
+  let compilation_unit = CU.create_instance base_compilation_unit arg_pairs in
+  let expected_output_prefix = CU.base_filename compilation_unit in
+  let output_prefix = Filename.remove_extension targetcm in
+  let output_prefix_basename = Filename.basename output_prefix in
+  if not (String.equal output_prefix_basename expected_output_prefix)
+  then begin
+    (* The module will only work if given the correct filename *)
+    error (Incorrect_target_filename
```
riaqn, line 66: I'm unfamiliar with how compilation units usually works. In vanilla ocaml, I can compile `a.ml` into `b.cmi/cmo`? Does that interfere with future code refer to `module A`?

testsuite/tests/templates/basic/test.ml
```diff
@@ -32,11 +40,14 @@
    ref_indirect.ml \
    ref_indirect.cmo.ocamlobjinfo.reference \
    ref_indirect.cmx.ocamlobjinfo.reference \
+   run.ml \
```
riaqn, line 43: is this file intentionally empty?

riaqn (commented):
Reviewed all tests

typing/persistent_env.ml
```diff
@@ -864,11 +864,32 @@ let local_ident penv modname =
...
   persistent_structures
-  |> Hashtbl.to_seq_keys
+  |> Hashtbl.to_seq_values
   |> Seq.filter_map
-       (fun name -> local_ident penv name |> Option.map (fun id -> name, id))
+        (fun ps ->
+           match ps.ps_binding with
+           | Runtime_parameter local_ident ->
+               Some (ps.ps_name_info.pn_global, local_ident)
+           | Constant _ -> None)
```
riaqn, line 892: Might be better to move this change to the previous PR. Also, the function `local_ident` is no longer needed?

typing/env.mli
```diff
@@ -360,6 +360,14 @@ val bound_cltype: string -> t -> bool
 
 val make_copy_of_types: t -> (t -> t)
 
+(* Resolution of globals *)
+
+(* [global_of_instance_compilation_unit cu] checks that a compilation unit is a
+   complete instantiation - that is, that all of its parameters are filled by
+   arguments, and all of those arguments' parameters are filled, and so on -
+   and converts it into a global. *)
+val global_of_instance_compilation_unit : Compilation_unit.t -> Global_module.t
```
riaqn, line 369: I think this change is related to code-generation and should stay in this PR. The rest of changes in this file should go to the previous PR.

utils/compilation_unit.mli
```diff
@@ -114,12 +114,18 @@ type argument =
...
+    is an argument for each parameter and each argument is itself a complete
+    instantiation. This ensures the name determines a compile-time constant, and
+    then the [t] returned here is the module corresponding to that constant.
+
+    The exception to the above is that a global name with _no_ arguments makes
+    sense even if the named module takes parameters, only in this case the [t]
+    refers to the module with the instantiating functor. This is only relevant
+    in two cases: when we're compiling the parameterised module itself, and
+    when we're instantiating it. *)
 val of_global_name : Global_module.Name.t -> t
```
riaqn, line 129: `Name.t` is user-syntax; this directly interacting with `CU` is probably not good.

Can you leave a CR in `compilation_unit.mli`:

> Should make a type `Global_module.Complete.t` which represents a complete instantiation. It's isomorphic to `Global_module.Name.t`. Have a function: `val is_complete: Global_module.t -> Global_module.Complete.t option`. Then, CU should only interact with this new type.

tools/objinfo.ml
```diff
@@ -229,15 +243,23 @@ let print_cms_infos cms =
...
     (match cms.cms_sourcefile with None -> "(none)" | Some f -> f)
 
-let print_general_infos print_name name crc defines iter_cmi iter_cmx =
+let print_general_infos print_name name crc defines arg_descr mbf
+    iter_cmi iter_cmx =
   printf "Name: %a\n" print_name name;
   printf "CRC of implementation: %s\n" (string_of_crc crc);
   printf "Globals defined:\n";
   List.iter print_name_line defines;
+  let () =
```
riaqn, line 252: Maybe abstract this logic into a function and call `Option.iter`, similar to the code below.

lambda/lambda.mli
```diff
@@ -762,9 +762,31 @@ and lambda_event_kind =
...
+  | Rp_argument_block of Global_module.t  (* The argument block of a module
+                                             compiled with [-as-argument-for] *)
+  | Rp_dependency of Global_module.t      (* A parameterised module (not itself a
+                                             parameter) that this module depends
+                                             on *)
+  | Rp_unit                               (* The unit value (only used when
+                                             there are no other parameters) *)
+
+type module_block_format =
+  | Mb_record of { mb_size : int }      (* A block with [mb_size] fields *)
```
riaqn, line 776: Maybe just `Mb_normal` and `Mb_parameterized`.  I feel "record" carries too much connotation (and also this doesn't have field names, so not really record)

lambda/lambda.mli
```diff
@@ -762,9 +762,31 @@ and lambda_event_kind =
   | Lev_function
   | Lev_pseudo
 
+(* Descriptor for a parameter that this module takes at runtime. *)
+type runtime_param_descr =
+  | Rp_argument_block of Global_module.t  (* The argument block of a module
+                                             compiled with [-as-argument-for] *)
+  | Rp_dependency of Global_module.t      (* A parameterised module (not itself a
+                                             parameter) that this module depends
+                                             on *)
```
riaqn, line 771: > `Rp_dependency x` is a runtime parameter that takes the module `x`, which is not completely instantiated; in particular, `x.hidden_args` is non-empty

driver/optcompile.ml
```diff
@@ -70,31 +89,106 @@ let emit unix i =
     (Unit_info.prefix i.target)
     ~progname:(Unit_info.source_file i.target)
 
-let implementation unix ~(flambda2 : flambda2) ~start_from ~source_file
-    ~output_prefix ~keep_symbol_tables =
-  let backend info ({ structure; coercion; _ } : Typedtree.implementation) =
+type starting_point =
```
riaqn, line 92: Since this is not really "starting_point" (`-instantiate` is incompatible with `-stop-after`), let's get a new name. How about:
```
type implementation_mode =
  | Normal
  | Emit
  | Instantiate of ..
```

driver/optcompile.ml
```diff
@@ -70,31 +89,106 @@ let emit unix i =
...
+      ~source_file ~output_prefix ~keep_symbol_tables
+      ~(compilation_unit : Compile_common.compilation_unit_or_inferred) =
+  let transl_style : Translmod.compilation_unit_style =
+    if Config.flambda || Config.flambda2 then Plain_block
+    else Set_individual_fields
+  in
+  let pipeline : Asmgen.pipeline =
+    Direct_to_cmm (flambda2 ~keep_symbol_tables)
+  in
+  let backend info ({ structure; coercion; argument_interface; _ }
```
riaqn, line 118: This function can be pushed into the `Parsing` branch.

driver/optcompile.ml
```diff
@@ -70,31 +89,106 @@ let emit unix i =
...
     in
-    let pipeline : Asmgen.pipeline =
-      Direct_to_cmm (flambda2 ~keep_symbol_tables)
+    let typed = structure, coercion, argument_coercion in
+    let as_arg_for =
+      !Clflags.as_argument_for
+      |> Option.map (fun param ->
+           (* Currently, parameters don't have parameters, so we assume the argument
+              list is empty *)
+           Global_module.Name.create_exn param [])
```
riaqn, line 133: Can call `create_no_args` instead.

testsuite/tests/templates/basic/test.ml
```diff
@@ -530,58 +754,286 @@
...
+           module = "";
+           flags = "-instantiate -as-argument-for Monoid";
+           program = "monoid_utils-Monoid_of_semigroup--String_semigroup.cmx";
+           all_modules = "monoid_utils.cmx monoid_of_semigroup-String_semigroup.cmx";
            ocamlopt.byte;
 
-           flags = "-parameter Semigroup -parameter List_element -w -misplaced-attribute";
-           module = "main_multi_arg.mli";
+           module = "";
+           flags = "-instantiate -as-argument-for Category";
```
riaqn, line 872: Note that `category_of_monoid.cmx` was already compiled with `-as-argument-for Category`. And here we say `-as-argument-for Category` again for the instance - do we check for some coherence?

Looking at `Instantiator.instantiate`, we simply take the `ui_arg_descr` of `Category_of_monoid.cmx`, and ignore the other one?

driver/compile.ml
```diff
@@ -57,21 +80,79 @@ let emit_bytecode i (bytecode, required_globals) =
...
+    (* FIXME delete; should not be necessary {[
+    (* Consider the names of arguments to be parameters for the purposes of the
+       subset rule - that is, a module we import can refer to our arguments as
+       parameters. *)
+    List.iter
+      (fun (param, _value) ->
+         let import = Compilation_unit.Name.of_head_of_global_name param in
+         Env.register_parameter_import import)
+      global_name.args;
+    ]} *)
```
riaqn, line 133: maybe just delete?

driver/compile.ml
```diff
@@ -57,21 +80,79 @@ let emit_bytecode i (bytecode, required_globals) =
...
        bytecode
        |> Profile.(record ~accumulate:true generate)
-         (Emitcode.to_file oc i.module_name cmo ~required_globals);
+         (Emitcode.to_file oc i.module_name cmo ~required_globals
+            ~module_block_format ~arg_descr);
     )
 
-let implementation ~start_from ~source_file ~output_prefix
-    ~keep_symbol_tables:_ =
+type starting_point =
```
riaqn, line 87: (See my comment in `optcompile.ml`)

riaqn (commented):
More comments.

driver/instantiator.ml
lukemaurer: Yes, you can, and it does: other modules have to say `B`. (We actually use this all the time to compile libraries.) We don't have that degree of freedom with `-instantiate` since .cmx lookup has to work.

testsuite/tests/templates/basic/test.ml
lukemaurer: Oh heh no, I must've had some idea for what it was going to be and changed my mind. Deleting

utils/compilation_unit.mli
lukemaurer: Makes sense to avoid mingling with user syntax, but for that we can just use `Global_module.t` instead, especially once the other caller can be covered by `Parameter.t` (once that's a thing). We can simply assert that the `Global_module.t` is complete. I'm pretty certain `Global_module.Complete.t` would only ever be used as an input to this function, since the whole point of a complete global is that it corresponds to a `CU.t`.

testsuite/tests/templates/basic/test.ml
lukemaurer: Oh, good point. We probably don't need to say `-as-argument-for Category` here. (Possibly it should even be an error?)

lambda/lambda.mli
```diff
@@ -762,9 +762,31 @@ and lambda_event_kind =
   | Lev_function
   | Lev_pseudo
 
+(* Descriptor for a parameter that this module takes at runtime. *)
+type runtime_param_descr =
+  | Rp_argument_block of Global_module.t  (* The argument block of a module
+                                             compiled with [-as-argument-for] *)
```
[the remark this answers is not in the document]
lukemaurer: The thing it actually takes _is_ the argument block, though. It's a crucial operational detail (and actually the only reason we need a separate constructor here). I'd rather mention the main module block in the `Rp_dependency` comment than not mention it in the first constructor.

lambda/lambda.mli
```diff
@@ -762,9 +762,31 @@ and lambda_event_kind =
   | Lev_function
   | Lev_pseudo
 
+(* Descriptor for a parameter that this module takes at runtime. *)
+type runtime_param_descr =
```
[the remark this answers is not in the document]
lukemaurer: Probably didn't need the comment before either :-) Done

lambda/lambda.mli
```diff
@@ -762,9 +762,31 @@ and lambda_event_kind =
...
+                                        (* A block with exactly one field:
+                                           a function taking [mb_runtime_params] and
+                                           returning a block with
+                                           [mb_returned_size] fields *)
+
 type program =
   { compilation_unit : Compilation_unit.t;
-    main_module_block_size : int;
+    module_block_format : module_block_format;
+    arg_block_field : int option;       (* Unnamed field with argument block
```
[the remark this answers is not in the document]
lukemaurer: I've renamed `module_block_format` to `main_module_block_format` since it wasn't clear that it only describes the main module block. `arg_block_field` describes where we put the argument block, which indeed only exists when someone passes `-as-argument-for`

lambda/lambda.mli
lukemaurer: I feel strongly that this is the place where we've passed through the nice typechecker view of the world and into the middle end, and these types should reflect that: we're talking about what the actual lambda terms consist of, not what was going on in the modules. I've added a comment describing the correspondence, though. I see your point about the word _record_, but I don't see a better word than _block_ (and it doesn't help much to say the block is a block). _Normal_ carries no information. Maybe `Mb_struct`? (Then the other might be `Mb_wrapped_functor`.)

lambda/lambda.mli
lukemaurer: Good idea to mention the invariant, but it's actually not quite that (adding a “see `is_complete`” instead)

driver/optcompile.ml
lukemaurer: It's exactly a starting point, though? It's the allowed values for the `~start_from` parameter to `implementation`, plus one more place we can start from. (Actually, I have no idea why `-instantiate` is deemed incompatible with `-stop-after`. I'll take that out.)

lukemaurer: (It's not in fact incompatible, it was just a string that said so)

typing/persistent_env.ml
lukemaurer: Deleting `local_ident`.

utils/compilation_unit.mli
lukemaurer: Adding a CR saying more or less this (it should take a [Global_module.t] and check that it's complete, and there should also be `of_parameter`)

lambda/lambda.mli
```diff
@@ -1040,3 +1064,13 @@ val simple_prim_on_values
 -> arity:int
 -> alloc:bool
 -> external_call_description
+
+(* Info for a compilation unit that implements a parameter (i.e., is an argument
+   for that parameter) *)
+
+type arg_descr =
```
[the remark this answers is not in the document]
lukemaurer: It's not just the argument to `-as-argument-for`, though (that would be the `arg_param` field, which I think is a more helpful name; I'll add a comment about the connection)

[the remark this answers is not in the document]
lukemaurer: Oh, leaving it as a non-documentation comment since it would be the only doc comment around, which only raises questions

testsuite/tests/templates/basic/test.ml
riaqn: Yeah if we don't need it, we shouldn't allow it.

driver/optcompile.ml
```diff
@@ -53,8 +60,20 @@ let compile i typed ~transl_style ~unix ~pipeline =
...
              ~ppf_dump:i.ppf_dump
              program);
+           let arg_descr =
+             make_arg_descr ~param:as_arg_for
+               ~arg_block_field:program.arg_block_field
+           in
            Compilenv.save_unit_info
-             (Unit_info.Artifact.filename (Unit_info.cmx i.target)))
+             (Unit_info.Artifact.filename (Unit_info.cmx i.target))
+             ~module_block_format:program.module_block_format
```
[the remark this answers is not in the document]
lukemaurer: I'm not sure how that helps? We would need separate functions `set_main_module_block_format` and `set_arg_descr` that would do nothing but secretly be parameters to this function. Even `set_export_info` is frankly rather suspect. Global state is bad

ocaml/testsuite/tests/templates/basic/main.ml
```diff
@@ -1,3 +1,7 @@
+module Monoid_utils_of_string =
```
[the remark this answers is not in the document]
lukemaurer: Good point. (In fact, in the long term we should really make `-parameter` and such attributes instead of (or in addition to?) CLI parameters.)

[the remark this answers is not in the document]
riaqn: +100 to make `-parameter` attribute.

testsuite/tests/templates/basic/bad_instantiate_not_arg.reference
lukemaurer: Leaving out the "for instantiation of" bit since it makes it sound like it's not compatible with `List_monoid` in particular, when actually it can't be used as an argument for instantiating _anything_

riaqn: Good point. Then maybe:
> Error: Module Widget cannot be used as argument because it was not compiled as such.
Hint: Compile Widget.cmi with -as-argument-for followed by one of parameters of List_monoid: XXX, YYY, ZZZ

lambda/lambda.mli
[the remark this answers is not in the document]
riaqn: > It's not just the argument to -as-argument-for, though

Do you mean there is an extra field `arg_block_field`? But that's like an implementation-level details about "in what way is this module an argument for `arg_param`".

testsuite/tests/templates/basic/bad_instantiate_not_arg.reference
lukemaurer: Oh, also, listing out the parameters here is a good idea but it turns out to be very awkward; I'll add a CR

driver/optcompile.ml
riaqn: What bothers me is that the name `type starting_point` sounds like just an alias for `Clflags.Compiler_pass.t` (in the sense that it specifies a point in the compiler pass), but in fact it also contains `Instantiate` which is not a point in the compiler pass, but a mode.

lambda/lambda.mli
riaqn: I take your point that lambda is middle-end and the names should sound so. However, given that the type name still involves "module", I think we can still mention "parameterized".

"Normal" indicates that it's the common case - note that `cmi_format.kind` already have two constructors "Normal" and "Parameter".

`Mb_struct` combined with the name "module" give the impression that the fields correspond to values in a `struct .. end` (is that true?)

lambda/translmod.ml
```diff
@@ -919,40 +928,136 @@ let required_globals ~flambda body =
...
   required
 
+let add_arg_block_to_module_block primary_block_lam size restr =
+  let primary_block_id = Ident.create_local "*primary-block*" in
+  let arg_block_id = Ident.create_local "*arg-block*" in
+  let arg_block_lam =
+    apply_coercion Loc_unknown Strict restr (Lvar primary_block_id)
+  in
+  let get_field i = Lprim (mod_field i, [Lvar primary_block_id], Loc_unknown) in
+  let all_fields = List.init size get_field @ [Lvar arg_block_id] in
```
riaqn, line 938: Looking at the code here, it seems that "arg_block" is a misnomer? It should be called "the field in the main block that represents -as-argument-for". I.e., it's a field in a block, not a block.

lambda/translmod.ml
```diff
@@ -919,40 +928,136 @@ let required_globals ~flambda body =
...
 (* Compile an implementation *)
 
 type compilation_unit_style =
   | Plain_block
   | Set_global_to_block
   | Set_individual_fields
 
-let transl_implementation_plain_block compilation_unit (str, cc) =
+let has_parameters () =
+  !Clflags.parameters <> []
```
riaqn, line 1004: I'm surprised that we need this. Can this information be passed to functions instead?

riaqn (commented):
More comments; I need to take a closer look at instantiator and translmod and asmpackager.

driver/compile.ml
```diff
@@ -57,21 +80,79 @@ let emit_bytecode i (bytecode, required_globals) =
...
+  | Parsing -> Parsing
+  | _ -> Misc.fatal_errorf "Cannot start from %s"
+           (Clflags.Compiler_pass.to_string start_from)
+
+let implementation0 ~start_from ~source_file ~output_prefix
+    ~keep_symbol_tables:_
+    ~(compilation_unit : Compile_common.compilation_unit_or_inferred) =
   let backend info typed =
-    let bytecode = to_bytecode info typed in
+    let as_arg_for =
```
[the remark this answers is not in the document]
lukemaurer: I'd rather keep this file as textually close to `optcompile.ml` as possible for easier maintenance. (They should, of course, be refactored.)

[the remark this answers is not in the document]
lukemaurer: Oh heh nevermind, I see the same comment applies there, too.

testsuite/tests/templates/basic/test.ml
lukemaurer: Making a CR; doing it right would be more effort than we need for something we don't intend people to write yet

lukemaurer: (actually, I'll just do it but make it ugly for the moment)

lambda/translmod.mli
```diff
@@ -25,7 +25,7 @@ type compilation_unit_style =
   | Set_individual_fields (* Closure *)
 
 val transl_implementation:
-      Compilation_unit.t -> structure * module_coercion
+      Compilation_unit.t -> structure * module_coercion * module_coercion option
```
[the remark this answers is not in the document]
lukemaurer: It should really be `Typedtree.implementation` rather than a tuple appearing out of nowhere, but that turns out to be untenable for _unbelievably_ bad reasons. Suffice it to say that making this API not shitty is beyond the scope of this PR. Adding a comment instead.

lambda/lambda.mli
lukemaurer: I wouldn't mind a better name in `Cmi_format.kind` as well, if we can think of a good retronym for “module that's not a parameter module.” Tempting to say _static module_ but that's not quite right: a parameterized module isn't quite static (it varies by what arguments you give it). The salient property is that the module has one specific definition (possibly an open term), but _defined module_ seems weird.

Yeah, the fields will be exactly the values as they would be in a `struct .. end` with an inferred signature (there are sometimes also a few extra fields, like the argument block, but the basic picture of “stuff stored like a `struct`” is accurate).

driver/optcompile.ml
lukemaurer: It's meant to be a generalization: sometimes we start from a compiler pass but sometimes we don't. (It might make sense to have one of the constructors be `Pass of Clflags.Compiler_pass.t` but I wanted to keep the constructors the same to keep as much of the `match` the same as upstream.)

lambda/translmod.ml
lukemaurer: I'm not sure what you mean: `arg_block_lam` is exactly the Lambda code that builds the argument block, `arg_block_field` is the field in which we put the argument block, and `arg_block_id` is a local identifier where we bind the argument block. It very much is a block (`apply_coercion` will return either a reference to a block or a `Pmakeblock` expression).

lambda/translmod.ml
lukemaurer: Yes, or at any rate we can ask `Env` for it. Fixing. (I think at some point the code must have been less uniform in handling parameters so sometimes `Env.parameters` didn't have the right information. The tests are passing still so that time must be in the past.)

lambda/translmod.ml
```diff
@@ -919,40 +919,134 @@ let required_globals ~flambda body =
...
+        body, Mb_record { mb_size = size }
+    | true ->
+        let mb_runtime_params, runtime_param_idents =
+          match Env.runtime_parameters () with
+          | [] ->
+              (* We didn't end up using any of the parameters, but this is still a
+                 parameterised module, so it must still be implemented as a function that
+                 produces a distinct value for each instance. *)
+              let unit_ident = Ident.create_local "*unit*" in
+              [ Rp_unit ], [ unit_ident ]
```
riaqn, line 1023: Does this correspond to the case where a module is compiled with `-parameter X` but never mentions `X`? I find it strange that this module can be later instantiated without `X`. I think such module should still require `X` for instantiation.

lambda/translmod.ml
```diff
@@ -919,40 +919,134 @@ let required_globals ~flambda body =
...
+          | globals ->
+              List.map
+                (fun (global, _) ->
+                   if Env.is_imported_parameter (Global_module.to_name global)
```
riaqn, line 1027: If a module registers `X` as parameter, but never uses it, it will be considered as `Rp_dependency` here?

lambda/translmod.ml
```diff
@@ -919,40 +919,134 @@ let required_globals ~flambda body =
...
+  let new_size = size + 1 in
+  Llet(Strict, layout_module, primary_block_id, primary_block_lam,
+       Llet(Strict, layout_module, arg_block_id, arg_block_lam,
+            Lprim(Pmakeblock(0, Immutable, None, alloc_heap),
+                  all_fields,
+                  Loc_unknown))),
+  new_size,
+  Some arg_block_field
+
+let add_parameters lam params =
```
riaqn, line 940: this special-purpose function is unsuitable as a top-level function, especially there is another function below called "has_parameters". At least call it `add_runtime_parameters`

riaqn (commented):
Looked at all code

lambda/lambda.mli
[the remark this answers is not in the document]
riaqn: I think the constructor names need to have some coherent contrast. So I think they should either be "parameter" and "dependency", or they should be "argument_field" and "whole_block". Currently we have one from each naming scheme.

lambda/lambda.mli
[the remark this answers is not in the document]
riaqn: If it's not essential that the argument field is itself a block, I suggest we just call it "argument_field".

driver/instantiator.ml
```diff
@@ -0,0 +1,235 @@
...
+
+type unit_info = {
+  ui_unit : CU.t;
+  ui_arg_descr : Lambda.arg_descr option;
+  ui_format : Lambda.main_module_block_format;
+}
+
+let instantiate
+      ~src ~args targetcm ~expected_extension ~read_unit_info ~compile =
+  let unit_infos = read_unit_info src in
```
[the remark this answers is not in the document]
lukemaurer: Oh heh true, I think it got inherited from `Cmx_format.unit_infos` but the singular is much better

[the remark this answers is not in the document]
lukemaurer: Actually, let me try calling them `base_unit_info` and `base_compilation_unit` (making the one on L48 unnecessary). `compilation_unit` is used below there for the CU that we're building

lambda/translmod.ml
riaqn: As discussed, I'm less worried that this doesn't affect type checking; but I still think having `Unit` in the middle-end complicates things.

driver/instantiator.ml
```diff
@@ -0,0 +1,235 @@
...
+         { param = CU.of_global_name param; value })
+      arg_info
+  in
+  let arg_map : (CU.t * int) Global_module.Name.Map.t =
+    match Global_module.Name.Map.of_list_checked arg_info with
+    | Ok map -> map
+    | Error (Duplicate { key; value1 = (arg1, _); value2 = (arg2, _) }) ->
+      error (Repeated_parameter { param = key; arg1; arg2 })
+  in
+  let compilation_unit = CU.create_instance base_compilation_unit arg_pairs in
```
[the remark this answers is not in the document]
lukemaurer: `output_prefix` is part of the API we're calling into, actually (it's used at the end of the file), and we do use `output_prefix` here. I've renamed them to be a bit more consistent and avoid the word _prefix_, but we do actually make use of the intermediate variable here.

lambda/lambda.mli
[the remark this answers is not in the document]
lukemaurer: Right. But that's why I hesitate to name the whole thing after the CLI flag: it _isn't_ just the value of the flag.

lambda/lambda.mli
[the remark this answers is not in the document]
lukemaurer: Makes sense, yeah, changing `Rp_dependency` to `Rp_main_module_block`

lambda/lambda.mli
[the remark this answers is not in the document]
lukemaurer: Its block-ness isn't essential here, but the _concept_ of “argument block” is. (I think you suggested elsewhere calling it `field_idx` so I did add the `_idx`.)

lambda/lambda.mli
lukemaurer: Ah, maybe I should think of a better name there. The issue is that actually it's _always_ a struct (changing that expectation might be possible but seems more disruptive than necessary), so in the case that we want to store a functor, we instead have to make a struct with just the functor in it (stuffing it in a one-element struct is what I meant by “wrapped”). Maybe I should just call it `Mb_functor` but I'd prefer that `lambda.mli` tell as few lies as possible, if we can manage

lambda/translmod.ml
lukemaurer: If it never uses it, it won't be returned by `Env.runtime_parameters ()`. Worth spelling that out, I'll add a comment.

lukemaurer: Actually, this is just needlessly confusing. Changed it to call `Env.is_parameter_unit` instead.

lambda/translmod.ml
lukemaurer: Yeah, that's a better name for sure, changing it

lambda/translmod.ml
lukemaurer: As elsewhere, I think “argument block” is an important concept that should be called out

lambda/lambda.mli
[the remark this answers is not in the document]
riaqn: I think it's fine for it to contain some middle-end details. `Lambda.lfunction` has plenty of such things, and we still call it function, because the correspondance to front-end is still useful.

lambda/lambda.mli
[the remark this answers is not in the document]
riaqn: My complaints against mentioning block here is that there is already main module block. So, is argument block parallel to the main module block? No, it's in fact a block inside a block. I think we should at least have comments saying it's the index INSIDE the main module block.

lambda/translmod.ml
riaqn: We discussed later, and because runtime parameters are determined by usage, including real parameters (which I would expected to be determined by registeration), it's difficult to do the way I want. As such, I'm resolving now.

I continue to think that real parameters should be treated seperately from dependencies and determined by registeration.

lambda/translmod.ml
riaqn: OK then how about `arg_block_idx`? I see you've adopted `_idx` elsewhere.

lukemaurer: Hmm! Yeah, it's not clear that the word _field_ is doing anything there, and _idx_ is good.

lambda/lambda.mli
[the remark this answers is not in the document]
lukemaurer: Ah, it's a sub-block of the main module block. That could certainly stand to be clearer. Adding comments.

riaqn (approved):

PR #1912, opened 2023-10-06
Propagate Uids for local variables and function parameters

mshinwell:
This is split out of #1852 and was mostly written by @tonowak .  It attaches `Uid.t` values, as used by the existing `Shape` infrastructure in the type checker, to at least some of the bound names in `Lambda`.  After that they make their way through Flambda 2, using a new `Flambda_uid` module (which I added) to keep track of situations where the corresponding variables are actually unboxed products.  They are then attached to `Backend_var.t` values which gets them into the DWARF emission code.

The remaining code in #1852 will then use these uids to look up the "shapes" of OCaml types, in order that they may be encoded into DWARF types.  The shape resolution code should in due course be able to deal with all of the complicated cases (for functors etc.), avoiding the need for something like the separate `libmonda` library on the debugger side.

I think this is ready for an initial look through at a high level from maybe two reviewers, one for the front end and one for Flambda 2 - I will come and ask people.  I'd like to get this in fairly soon as the potential for conflicts is high.

### DIFF
--- a/ocaml/lambda/translclass.ml
+++ b/ocaml/lambda/translclass.ml
@@ -67,7 +67,9 @@ let lapply ap =
       Lapply ap
 
 let lparam name layout : Lambda.lparam =
-  { name; layout;
+  { name;
+    var_uid = Uid.internal_not_actually_unique (* CR tnowak: verify *);
+    layout;
     attributes = Lambda.default_param_attribute; mode = alloc_heap }
 
 let mkappl (func, args, layout) =
@@ -112,7 +114,9 @@ let transl_val tbl create name =
 let transl_vals tbl create strict vals rem =
   List.fold_right
     (fun (name, id) rem ->
-      Llet(strict, layout_int, id, transl_val tbl create name, rem))
+      (* CR tnowak: unsure, verify *)
+      let uid = Uid.internal_not_actually_unique in
+      Llet(strict, layout_int, id, uid, transl_val tbl create name, rem))
     vals rem
 
 let meths_super tbl meths inh_meths =
@@ -128,7 +132,8 @@ let meths_super tbl meths inh_meths =
 let bind_super tbl (vals, meths) cl_init =
   transl_vals tbl false StrictOpt vals
     (List.fold_right (fun (_nm, id, def) rem ->
-         Llet(StrictOpt, layout_meth, id, def, rem))
+         let uid = Uid.internal_not_actually_unique in
+         Llet(StrictOpt, layout_meth, id, uid, def, rem))
        meths cl_init)
 
 let create_object cl obj init =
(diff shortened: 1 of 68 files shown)


mshinwell:
One other thing: there is a plan to also propagate information about which variables are currently in scope, but that's going to come separately (and is not yet written).

ocaml/lambda/translclass.ml
```diff
@@ -67,7 +67,9 @@ let lapply ap =
       Lapply ap
 
 let lparam name layout : Lambda.lparam =
-  { name; layout;
+  { name;
```
ccasin, line 70: I think at many call-sites it would be possible to provide a good uid with what is already available or by threading through a little more info from the typechecker.

This is also true for many other `internal_not_actually_unique` uses in this file.  I haven't bothered to label them all, because it's not obvious to me how much value there is in improving debugging for the class system.

ocaml/lambda/translcore.ml
```diff
@@ -1133,7 +1147,8 @@ and transl_apply ~scopes
             Lvar _ | Lconst _ -> (lam, layout)
           | _ ->
               let id = Ident.create_local name in
-              defs := (id, layout, lam) :: !defs;
+              let uid = Uid.internal_not_actually_unique in
+              defs := (id, uid, layout, lam) :: !defs;
```
ccasin, line 1151: Seems like we don't really need to add `uid` to this tuple of they are all going to be `internal_not_actually_unique`.

ocaml/lambda/translcore.ml
```diff
@@ -1242,6 +1269,8 @@ and transl_curried_function
         let arg_mode = transl_alloc_mode arg_mode in
         let params = {
           name = param ;
+          var_uid = Ident.Tbl.find_opt (uids_of_vars cases) param
+                    |> Option.value ~default:Uid.internal_not_actually_unique ;
```
ccasin, line 1273: This seems expensive and unnecessary.  I think where `param` is calculated in `typecore`, the uid is available.  That is, `Typecore.name_pattern` has access to the relevant uid in the case it doesn't use the default, and we should just rememeber that uid as part of the `Texp_function` rather than doing all this work here.

ocaml/lambda/translcore.ml
```diff
@@ -1341,6 +1372,8 @@ and transl_function0
     let arg_mode = transl_alloc_mode arg_mode in
     ((Curried {nlocal},
       [{ name = param;
+         var_uid = Ident.Tbl.find_opt (uids_of_vars cases) param
+            |> Option.value ~default:Uid.internal_not_actually_unique;
```
ccasin, line 1376: Similarly, we can remember the uid instead of searching for it here, and then delete `uids_of_vars`.

ccasin (commented):
I've taken a first pass over this - I read `ocaml/typing` and `ocaml/lambda`

My mental model of what is going on here is that we're adding `Uid.t`s in various places which will be used by the debugger to try to point to a relevant identifier in the source program.  So, in any place where `Uid.internal_not_actually_unique` appears, it's possible that we're missing nice debugging information, but there's no possibility of a serious bug (and the debugging info could be improved later). So, for most of those, I've applied limited scrutiny to the question of whether a better uid might be available (and I confess my enthusiasm for thinking hard about it waned as a I read).  If I've misunderstood and these deserve more scrutiny, let me know.

In addition to my comments below, one small overall remark: There are many very long lines in this PR, and a pass to get them all down to 80 characters would be worthwhile.  Maybe we should turn on some kind of checking for this, as they have upstream.

PR #1913, opened 2023-10-06
Add float32 frontend definitions 

TheNumbat:
Merges into #1900.

Adds `Pfloat32` to the `boxed_float` variant, `float32` ~and `float32array`~ types to the frontend, and threads these representations through the compiler up until the middle-ends. 

Does not yet add parser support for float32 literals.


tests/float32/basic.ml
```diff
@@ -0,0 +1,5 @@
+open Stdlib
+
+type f32 = float32
+
+type f32array = float32array
```
ccasin, line 5: Do we expect float32 to work correctly in the `ocaml/` compiler?  I think yes, in which case I think it's fine to add tests only to `ocaml/testsuite`.

ocaml/testsuite/tests/typing-float32/types.ml
```diff
@@ -0,0 +1,13 @@
...
+
+type t = float32;;
+[%%expect{|
+type t = float32
+|}];;
+
+type t = float32array;;
+[%%expect{|
+type t = float32array
+|}];;
```
ccasin, line 13: I think we should have more tests .  In particular, we should be checking that the compiler does indeed blow up loudly if you try to compile a program with a float32.

ocaml/typing/untypeast.ml
```diff
@@ -138,6 +138,7 @@ let constant = function
   | Const_int64 i -> Pconst_integer (Int64.to_string i, Some 'L')
   | Const_nativeint i -> Pconst_integer (Nativeint.to_string i, Some 'n')
   | Const_float f -> Pconst_float (f, None)
+  | Const_float32 f -> Pconst_float (f, Some 's')
```
ccasin, line 141: I think in C this is `f`, right?  I can't recall if we discussed what the best choice here is.

ocaml/typing/typecore.ml
```diff
@@ -8263,6 +8264,8 @@ let report_literal_type_constraint expected_type const =
       Some 'L'
     else if Path.same expected_type Predef.path_nativeint then
       Some 'n'
+    else if Path.same expected_type Predef.path_float32 then
+      Some 's'
```
ccasin, line 8268: Same question about suffix as in untypeast

ocaml/lambda/lambda.ml
```diff
@@ -1596,7 +1598,10 @@ let primitive_result_layout (p : primitive) =
   | Pbigarrayref (_, _, kind, _) ->
       begin match kind with
       | Pbigarray_unknown -> layout_any_value
-      | Pbigarray_float32 | Pbigarray_float64 -> layout_boxed_float Pfloat64
+      | Pbigarray_float32 ->
+        (* CR mslater: (float32) bigarrays *)
+        layout_boxed_float Pfloat64
```
ccasin, line 1603: Is `Pfloat64` correct here?

ccasin (commented):
I read the `ocaml/typing` and `ocaml/lambda` bits.  Looks fine, except that we should have good tests showing that you get a loud error if you try to use one of these.

tests/float32/basic.ml
mshinwell: Hmm, I thought we'd decided that the unboxed types support wouldn't be available in `ocaml/`?

TheNumbat: float32 won't work in the `ocaml/` compiler. I'm not sure if there's a good way to disable the primitive types only in the upstream build, though.

ocaml/typing/untypeast.ml
TheNumbat: We can't use `f` because it conflicts with hex float literals, so we decided to go with `s` for single and later on `h` for half.

ocaml/lambda/lambda.ml
TheNumbat: Yeah - I just broke out the case so I could add the CR. Later on I will convert the float32 bigarray kind to actually give you back float32s.

ocaml/testsuite/tests/typing-float32/types.ml
TheNumbat: (Discussed offline)

tests/float32/basic.ml
ccasin: It seems fine not to support it.  I wish we had a way to test that it errors, but I don't see an easy way to.

ocaml/typing/untypeast.ml
ccasin: Right, thanks.

ocaml/testsuite/tests/typing-float32/types.ml
ccasin: Thanks!

ccasin (approved):
Approving for `ocaml/typing` and `ocaml/lambda` (and tests)

ocaml/typing/predef.mli
```diff
@@ -102,6 +102,11 @@ val build_initial_env:
 val add_simd_extension_types :
   (Ident.t -> type_declaration -> 'a -> 'a) -> 'a -> 'a
 
+(* Add small number types to an environment.  This is separate from [build_initial_env]
+   because we'd like to only do it if the simd extension is on. *)
```
ccasin, line 106: I think you mean the small number extension.

ccasin (approved):
Minor typo, but this looks good - thanks for changing to an extension.

TheNumbat:
@ccasin should I update these PRs to also add the float kind to `Pufloatfield`? It's still assuming float64.

mshinwell:
I've read everything except `lambda/` and `typing/` which @ccasin has read.

mshinwell (approved):

PR #1382, opened 2023-05-11
Refactor primitive handling in Lambda_to_flambda

mshinwell:
This allows us to remove the `primitive_result_kind` function together with some duplicate code (calling `transform_primitive`).

PR #1915, opened 2023-10-09
Fix breaking of tail recursion in classic mode

mshinwell:
Patch from @lthls 

Tail recursion was being broken in cases involving `Psequand` and `Psequor` in classic mode.  For example:
```
let rec f1 x =
  x && f1 x

let rec f2 x y z =
  x && y && z && f2 x y z

let rec f3 x =
  x || f3 x

let rec f4 x y z =
  x || y || z || f4 x y z
```


mshinwell:
The problem might have been introduced by https://github.com/ocaml-flambda/flambda-backend/pull/1382

PR #888, opened 2022-10-12
Adjust use of kinds during CPS conversion

mshinwell:
Easy one first: add a missing kind propagation in the `Lstaticcatch` case.

Harder: switch to a code fragment / idea suggested independently by @lpw25 and @chambart , which will produce fewer `Let_cont`s, which is good.  This new code will cause the value kind/subkind information on `Llet` to be dropped in certain cases (note: not the `Lprim` case).  However it seems like this information shouldn't be needed anyway and @lpw25 would like to stop providing it in Lambda, since it's expensive to compute.  If it's essential for a particular primitive we can add it to that primitive.  In general, though, we should have the kind/subkind information from the "inputs": function parameters, results of function applications, etc -- up to the bound in `Typeopt.value_kind` at least.

PR #1916, opened 2023-10-09
Fix CR in Lambda_to_flambda about extraneous continuations

mshinwell:
This is a first attempt at doing this.


mshinwell:
Leaving this for the moment, related to #888

PR #1919, opened 2023-10-10
Shorten symbol names in classic mode

mshinwell:
This helps reduce string table size in executables.

### DIFF
--- a/middle_end/flambda2/identifiers/int_ids.ml
+++ b/middle_end/flambda2/identifiers/int_ids.ml
@@ -714,7 +714,11 @@ module Code_id = struct
       !previous_name_stamp
     in
     let linkage_name =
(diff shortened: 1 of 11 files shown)


middle_end/flambda2/identifiers/int_ids.ml
```diff
@@ -714,7 +714,11 @@ module Code_id = struct
       !previous_name_stamp
     in
     let linkage_name =
-      let name = Printf.sprintf "%s_%d_code" name name_stamp in
+      let name =
+        if Flambda_features.Expert.shorten_symbol_names ()
+        then Printf.sprintf "%s_%d" name name_stamp
```
Ekdohibs, line 719: Couldn't this cause a conflict with the name of the set of closures itself?

PR #1923, opened 2023-10-10
Avoid partial register stall in roundsd

TheNumbat:
As with `sqrtsd` (see https://github.com/ocaml/ocaml/pull/9041), `roundsd` should pre-clear the destination register to prevent false dependencies from flowing through its upper 64 bits.


gretay-js (approved):
- This is more subtle than I expected: the use of `<>` was okay for `sqrt` because we only had one register class, but now we really need to get types of registers right for this to work. 

- Is `roundsd` the only operation that may lead to a partial register stall situation similar to `sqrt`? There are a few other new unary operations with register behavior `RM_R` and `R_to_R` that were added recently (`cvtss2sd`, `cvtsi2sd`). I am not suggesting that we emit `xorpd` for all of them, it would depend on performance vs size trade-off.  It would be nice to track these decisions somewhere (even as a comment in code or on this PR). It's probably too much work at this point, but we can add a function `should_avoid_partial_register_stall simd_instruction` so that when we add new instructions, it is one of the things we consider.

TheNumbat:
`roundsd` maps a scalar float register to a scalar float register, so it should still always be comparing within the same register class, right? Agreed that we now have to think about comparisons between scalar and vector registers, though.

I think `cvtss2sd`, `cvtsi2sd`, and `cvtsd2ss` can also cause stalls, since they're all unary operations that preserve the high bits of the destination. These are only used for casting between int/float/float32, though, so I'm not sure if it matters. (I'm also not sure this matters in general with the current register allocator and CPUs, either.)

PR #1924, opened 2023-10-11
Improve layout history error message: better test coverage

alanechang:
Add tests to cover all possible layout reason messages


goldfirere:
Delegate review to @ncik-roberts.

alanechang:
Note to reviewer: This PR contains very little change to actual code (only removes a unused constructor `With_constraint`) and mainly focuses on adding tests. The new tests aim to cover every creation_reason constructor in `layouts.ml` and have their corresponding error message displayed. This goal is not fully achieved and it was difficult to generate code snippets to reach certain constructors. In those cases, either a comment is left explaining why this constructor is believed to be unreachable or a CR is attached.

How to review: For each reason constructor (noted in the comment above the test case), if an error message is present, check it's relevant to the constructor specified and the message content is helpful. If no error message is present, suggest potential test case to reach that constructor (don't think too hard on this, only do it if it's easy) or check the associate comment in place is reasonable.

ncik-roberts:
Reassigning review to Richard as discussed.

ocaml/testsuite/tests/typing-layouts-err-msg/any.ml
```diff
@@ -0,0 +1,16 @@
...
+(* Any layout errors *)
+
+(* Missing_cmi *)
+(* Wildcard *)
+(* Unification_var *)
+(* Initial_typedecl_env *)
+(* Dummy_layout *)
+
+(* Type_expression_call *)
+(* Everything is a sublayout of any. Won't throw error *)
```
goldfirere, line 16: I think it might be possible to make a `Missing_cmi` error, though the others probably not. But not because everything is a sublayout of `any`. If we had `t : any` and you wrote `t list`, then we'd check `any < value`, which fails, and it would print out a message.

I would say that the text associated with each of these (and any other messages we can't produce) should request folks to contact the Jane Street compilers team.

ocaml/testsuite/tests/typing-layouts-err-msg/concrete.ml
```diff
@@ -0,0 +1,151 @@
...
+type t_void : void
+Line 5, characters 15-37:
+5 | let () = match (assert false : t_any) with _ -> ()
+                   ^^^^^^^^^^^^^^^^^^^^^^
+Error: This expression has type t_any but an expression was expected of type
+         ('a : '_representable_layout_1)
+       The layout of t_any is any, because
+         of the definition of t_any at line 1, characters 0-16.
+       But the layout of t_any must be a sublayout of '_representable_layout_1, because
+         it's matched against a pattern.
```
goldfirere, line 26:
```suggestion
         a value of this type is matched against a pattern.
```

ocaml/testsuite/tests/typing-layouts-err-msg/concrete.ml
```diff
@@ -0,0 +1,151 @@
...
+
+[%%expect{|
+Line 1, characters 9-19:
+1 | type t = A of t_any
+             ^^^^^^^^^^
+Error: Constructor argument types must have a representable layout.
+       The layout of t_any is any, because
+         of the definition of t_any at line 1, characters 0-16.
+       But the layout of t_any must be a sublayout of '_representable_layout_2, because
+         it's used as constructor field 0.
```
goldfirere, line 40: "field 0" isn't very user friendly. "first constructor field" is good, but probably more work than necessary. So I think I'd just go with

```suggestion
         it's used as the type of a constructor field.
```

ocaml/testsuite/tests/typing-layouts-err-msg/concrete.ml
```diff
@@ -0,0 +1,151 @@
...
+
+[%%expect{|
+Line 1, characters 10-18:
+1 | type t = {a: t_any}
+              ^^^^^^^^
+Error: Record element types must have a representable layout.
+       The layout of t_any is any, because
+         of the definition of t_any at line 1, characters 0-16.
+       But the layout of t_any must be a sublayout of '_representable_layout_3, because
+         it's used in the declaration of the record field "a/279".
```
goldfirere, line 54:
```suggestion
         it is the type of record field a.
```

Note that I've dropped the quotes and the numerical suffix. (Actually, I think error mesages should do a better job of distinguishing message from quoted code, but that would be a major overhaul.) To get rid of the suffix, you'll have to use a different printer for Idents, I think.

ocaml/testsuite/tests/typing-layouts-err-msg/concrete.ml
```diff
@@ -0,0 +1,151 @@
...
+[%%expect{|
+Line 2, characters 9-14:
+2 | and t2 = t_any t
+             ^^^^^
+Error: This type t_any should be an instance of type
+         ('a : '_representable_layout_4)
+       The layout of t_any is any, because
+         of the definition of t_any at line 1, characters 0-16.
+       But the layout of t_any must be a sublayout of '_representable_layout_4, because
+         it instantiates an unannotated type parameter.
```
goldfirere, line 70:
```suggestion
         it instantiates an unannotated type parameter of t.
```

Not necessary if it's hard

ocaml/testsuite/tests/typing-layouts-err-msg/concrete.ml
```diff
@@ -0,0 +1,151 @@
...
+Line 1, characters 4-12:
+1 | let x: t_any = assert false
+        ^^^^^^^^
+Error: This pattern matches values of type t_any
+       but a pattern was expected which matches values of type
+         ('a : '_representable_layout_5)
+       The layout of t_any is any, because
+         of the definition of t_any at line 1, characters 0-16.
+       But the layout of t_any must be a sublayout of '_representable_layout_5, because
+         it's bound by a `let`.
```
goldfirere, line 92:
```suggestion
         it's the type of a variable bound by a `let`.
```

ocaml/testsuite/tests/typing-layouts-err-msg/concrete.ml
```diff
@@ -0,0 +1,151 @@
...
+|}]
+
+(* Structure_item_expression *)
+(* See [concrete_struct_item_expr.ml] *)
+
+(* V1_safety_check *)
+(* See [concrete_v1_check.ml] *)
+
+(* External_argument *)
+(* external eq : t_any -> 'a -> bool = "%equal" *)
```
goldfirere, line 133: Keep this test in anyway. `Function_argument` won't last much longer.

ocaml/testsuite/tests/typing-layouts-err-msg/concrete.ml
```diff
@@ -0,0 +1,151 @@
...
+(* Shadowed by Function_argument *)
+
+(* External_result *)
+(* external eq : 'a -> 'a -> t_any = "%equal" *)
```
goldfirere, line 137: Keep this test in anyway. `Function_result` won't last much longer.

ocaml/testsuite/tests/typing-layouts-err-msg/concrete_struct_item_expr.compilers.reference
```diff
@@ -0,0 +1,11 @@
+type t_any : any
+Line 1, characters 0-22:
+1 | (assert false : t_any);;
+    ^^^^^^^^^^^^^^^^^^^^^^
+Error: This expression has type t_any but an expression was expected of type
+         ('a : '_representable_layout_1)
+       The layout of t_any is any, because
+         of the definition of t_any at line 6, characters 0-16.
+       But the layout of t_any must be a sublayout of '_representable_layout_1, because
+         it's used in an expression in a structure.
```
goldfirere, line 10:
```suggestion
         it's used as the type of an expression in a structure.
```

ocaml/testsuite/tests/typing-layouts-err-msg/concrete_v1_check.ml
```diff
@@ -0,0 +1,9 @@
+(* TEST
+   flags = "-extension layouts_alpha"
+   * toplevel
+*)
+
+type t_void : void;;
+(assert false : t_void);;
+(* CR layouts v2.9: This error message is bad, but it should be gone when merged back into
```
goldfirere, line 8: Why is this going away? It will eventually, but not for months.

ocaml/testsuite/tests/typing-layouts-err-msg/concrete_v1_check.compilers.reference
```diff
@@ -0,0 +1,9 @@
+type t_void : void
+File "_none_", line 1:
+Error: Non-value detected in [translmod]:
+       Please report this error to the Jane Street compilers team.
+       The layout of t_void is void, because
+         it's part of the v1 safety check, defaulted to layout void.
```
goldfirere, line 6: I'm a little alarmed at "defaulted". We should never be defaulting to void anymore. (We did for a short time about a year ago.)

ocaml/testsuite/tests/typing-layouts-err-msg/debug_printer.ml
```diff
@@ -0,0 +1,8 @@
+(* TEST
+   flags = "-extension layouts_alpha"
+   * toplevel
+*)
+type ('a : float64) t = 'a
+let f ppf (x : 'a t) = ();;
+#install_printer f;;
+(* Always shows the "??? has the wrong type for a printing function" message on exception *)
```
goldfirere, line 8: OK -- then update the text for `Debug_printer_argument` to tell people to give us a ring.

ocaml/testsuite/tests/typing-layouts-err-msg/immediate.ml
```diff
@@ -0,0 +1,66 @@
+(* TEST
+   flags = "-extension layouts_alpha"
+   * expect
+*)
+
+(***************************)
+(* Immediate layout errors *)
+
+(* Empty_record *)
+(* Records with all void fields are not yet supported *)
```
goldfirere, line 10: Make the test anyway. The error will be about void elements in a record, but then it will magically become the right error later.

ocaml/testsuite/tests/typing-layouts-err-msg/immediate.ml
```diff
@@ -0,0 +1,66 @@
...
+                                 ^
+Error: This expression has type [ `A | `B ]
+       but an expression was expected of type 'a t = ('a : void)
+       The layout of [ `A | `B ] is immediate, because
+         it's an enumeration variant (all constructors are constant).
+       But the layout of [ `A | `B ] must be a sublayout of void, because
+         of the definition of t at line 1, characters 0-22.
+|}]
+
+(* Gc_ignorable_check *)
```
goldfirere, line 62: These should go. Could you please add a `(* CR layouts v2.8: Remove Gc_ignorable_check after the check uses modal kinds *)` to the declaration of the constructors, please?

ocaml/testsuite/tests/typing-layouts-err-msg/immediate.ml
```diff
@@ -0,0 +1,66 @@
...
+(* Only used within [Result.is_ok] and not thrown as an exception *)
+
+(* Value_kind *)
+(* Not used anywhere *)
```
goldfirere, line 66: Then please remove.

ocaml/testsuite/tests/typing-layouts-err-msg/immediate64.ml
```diff
@@ -0,0 +1,16 @@
+(* TEST
+   flags = "-extension layouts_alpha"
+   * expect
+*)
+
+(*****************************)
+(* Immediate64 layout errors *)
+
+(* Local_mode_cross_check *)
```
goldfirere, line 9: This should also go away with v2.8. Please leave a CR at the declaration site.

ocaml/testsuite/tests/typing-layouts-err-msg/immediate64.ml
```diff
@@ -0,0 +1,16 @@
...
+(* Only used within [Result.is_ok] and not thrown as an exception *)
+
+(* Gc_ignorable_check *)
+(* Only used within [Result.is_ok] and not thrown as an exception *)
```
goldfirere, line 13: This one, too, will go away with v2.8.

ocaml/testsuite/tests/typing-layouts-err-msg/immediate64.ml
```diff
@@ -0,0 +1,16 @@
...
+
+(* Separability_check *)
```
goldfirere, line 15: Maybe this goes away or maybe not. But I see there's already a CR v2.8 nearby. Good. This is fine then. (Resolve upon reading.)

ocaml/testsuite/tests/typing-layouts-err-msg/probe.reference
```diff
@@ -0,0 +1,10 @@
+File "probe.ml", line 12, characters 6-7:
+12 |   let f () = x in
+           ^
+Warning 26 [unused-var]: unused variable f.
+File "probe.ml", lines 11-14, characters 20-2:
+11 | ....................[%probe "a" (
+12 |   let f () = x in
+13 |   ()
+14 | )]
+Error: Variables in probe handlers must have layout value, but x in this handler does not.
```
goldfirere, line 10: This is very sad, as the comments in translcore describe. Just update the layout-histories text to tell a reader to call for help.

ocaml/testsuite/tests/typing-layouts-err-msg/value.ml
```diff
@@ -0,0 +1,433 @@
+(* TEST
+   flags = "-extension layouts_alpha"
+   * expect
+*)
+
```
goldfirere, line 5: I did not review these error messages for quality. Just completeness. Please add a CR to the top of this file saying we have to review for quality.

ocaml/testsuite/tests/typing-layouts-err-msg/value.ml
```diff
@@ -0,0 +1,433 @@
...
+                                   ^
+Error: This expression has type int -> int
+       but an expression was expected of type 'a t = ('a : void)
+       The layout of int -> int is value, because
+         it's a function type.
+       But the layout of int -> int must be a sublayout of void, because
+         of the definition of t at line 1, characters 0-22.
+|}]
+
+(* Tfield *)
```
goldfirere, line 262: `Tfield` and `Tnil` are very strange. As the message suggests, they shouldn't actually happen. Maybe just clarify the text to call for help, but we don't need to add tests here.

ocaml/testsuite/tests/typing-layouts-err-msg/value.ml
```diff
@@ -0,0 +1,433 @@
...
+let f : t -> ('a : void) = fun x -> x
+[%%expect{|
+type t
+Line 2, characters 36-37:
+2 | let f : t -> ('a : void) = fun x -> x
+                                        ^
+Error: This expression has type t but an expression was expected of type
+         ('a : void)
+       The layout of t is value, because
+         of the definition of t at line 1, characters 0-6.
```
goldfirere, line 330: This isn't the message we're looking for. I think you'll need a mutually recursive group of type declarations.

ocaml/testsuite/tests/typing-layouts-err-msg/value.ml
```diff
@@ -0,0 +1,433 @@
...
+Error: This expression has type 'b * 'c
+       but an expression was expected of type ('a : void)
+       The layout of 'a * 'b is value, because
+         it's a tuple type.
+       But the layout of 'a * 'b must be a sublayout of void, because
+         of the annotation on the type variable 'a.
+|}];;
+
+(* Row_variable *)
+(* CR layouts v2.9: add test *)
```
goldfirere, line 228: I couldn't get this to print either. Just update the text to call for help.

ocaml/testsuite/tests/typing-layouts-err-msg/value.ml
```diff
@@ -0,0 +1,433 @@
...
+                        ^^
+Error: m1 must have a type of layout value because it is captured by an object.
+       The layout of t_float64 is float64, because
+         of the definition of t_float64 at line 5, characters 0-24.
+       But the layout of t_float64 must be a sublayout of value, because
+         it's captured in an object.
+|}];;
+
+(* Unknown *)
+(* CR layouts v2.9: add test *)
```
goldfirere, line 433: If we could add a test, it wouldn't be Unknown! I think we can just skip for now.

ocaml/testsuite/tests/typing-layouts-err-msg/void.ml
```diff
@@ -0,0 +1,10 @@
+(* TEST
+   flags = "-extension layouts_alpha"
+   * expect
+*)
+
+(**********************)
+(* Void layout errors *)
+
+(* V1_safety_check *)
+(* Not used in the codebase *)
```
goldfirere, line 10: Then remove it.

ocaml/testsuite/tests/typing-layouts-err-msg/concrete_v1_check.ml
alanechang: Just the layout history part will be going away when merged.

The check has been refactor in `main`: 
https://github.com/ocaml-flambda/flambda-backend/blob/177760ad8f2b96538c30db223ec467493a5803b2/ocaml/lambda/translmod.ml#L56-L58
It's not longer creating a sort var just for the error message like before: 
https://github.com/ocaml-flambda/flambda-backend/blob/bc38ff3587c2c6d33f1d3e24f9eea79715849945/ocaml/lambda/translmod.ml#L57-L65
In the future, the error message will just be:
```
Error: Non-value sort void detected in [translmod] in type t_void:
       Please report this error to the Jane Street compilers team.
```
and the reason constructor in question can be deleted.

ocaml/testsuite/tests/typing-layouts-err-msg/concrete_v1_check.compilers.reference
alanechang: See the comment on `concrete_v1_check.ml`

ocaml/testsuite/tests/typing-layouts-err-msg/concrete_v1_check.ml
goldfirere: OK -- makes sense. Then perhaps good to leave a CR reminding us to delete the constructor.

ocaml/testsuite/tests/typing-layouts-err-msg/concrete.ml
alanechang: Was able to implement this.

alanechang:
This is ready for review again. Main changes that were made are:

- some extra logic to get the missing cmi layout reason in more places (changes `typing-missing-cmi-indirections/test.compilers.reference`)

https://github.com/ocaml-flambda/flambda-backend/blob/9a0d3a4dc653955d3108d56dd0e076276c06fd07/ocaml/typing/layouts.ml#L937-L946

> This also means backporting the `score_reason` function from #1943

- adding a `Path.t` parameter to `Unannotated_type_parameter`
- attaching a "Please notify the Jane Street compilers group" message to layout reasons that we believe to be unreachable
- new test for `Missing_cmi`, fixed `Default_type_layout` test
- various wording changes

goldfirere (approved):

goldfirere:
I have reviewed @alanechang's new commits. I thought the approach in one involving missing cmis wasn't great, and so took a different approach. Alan, could you review my three commits above? (I ended up doing this myself, starting last week and not realizing I would be unable to finish.) Then we can merge! Thanks.

goldfirere:
I've reviewed the last two commits from @alanechang, which include a fix for a bug I introduced. (Thanks, Alan!) Is this ready to merge now?

alanechang:
@goldfirere The changes look good. Just an issue around setting `missing_cmi` in `Violation.t`. I pushed two new commits to both fix the issue and add tests.

alanechang:
Yes, it's ready to be merged now.

PR #1925, opened 2023-10-11
Eliminate need for parens around labeled tuple expressions

ccasin:
This is (hopefully) the first in a series of three small PRs to remove the need for parens around labeled tuple expressions, patterns, and types.  This one does expressions.

I wrote a long comment (maybe overkill?) explaining how the parsing works, so I won't re-explain it here.  I'm open to suggestions about better solutions (there's almost certainly a cleaner way).

Note that this PR appears to move where in the grammar tuples go, but in fact it's just moving them back to where they were before the labeled tuples work started (and making them a special case of labeled tuples).

No new tests - but the existing tests have been updated to drop the parens in all the places where it seems sensible, and it works.


ocaml/parsing/parser.mly
```diff
@@ -3359,7 +3438,7 @@ reversed_ltuple_component_comma_list:
 %inline ltuple_component_comma_list:
   | rev(reversed_ltuple_component_comma_list) { $1 }
 ;
-
+*)
```
riaqn, line 3441: If we no longer need these old rules, can we remove them?

ocaml/parsing/parser.mly
```diff
@@ -3312,14 +3314,91 @@ fun_param_as_list:
...
+     { None, $1 }
+  | LABEL simple_expr %prec below_HASH
+     { Some $1, $2 }
+  | TILDE label = LIDENT
+     { let loc = $loc(label) in
+       Some label, mkexpvar ~loc label }
+  | TILDE LPAREN label = LIDENT ty = type_constraint RPAREN %prec below_HASH
+      { Some label,
+        mkexp_constraint
+          ~loc:($startpos($2), $endpos) (mkexpvar ~loc:$loc(label) label) ty }
```
riaqn, line 3362: It seems that some rules has `%prec` while some doesn't. Is it because in those cases the default precedence (determinited by the rightmost terminal symbol) already gives the right behaviour? But it seems that `LIDENT` is higher than `below_HASH`?

riaqn (commented):
Some questions. 

Also some side note:  `parser.mly` already spells `Labelled` while this PR spells `Labeled` - but I guess that's fine.

goldfirere (commented):
For what it's worth, I read the long comment in the parser and don't immediately see a better solution.

ocaml/parsing/parser.mly
```diff
@@ -2860,20 +2858,24 @@ fun_expr:
...
+      }
   | mkrhs(constr_longident) simple_expr %prec below_HASH
-      { Pexp_construct($1, Some $2) }
+      { mkexp ~loc:$sloc (Pexp_construct($1, Some $2)) }
   | name_tag simple_expr %prec below_HASH
-      { Pexp_variant($1, Some $2) }
+      { mkexp ~loc:$sloc (Pexp_variant($1, Some $2)) }
   | e1 = fun_expr op = op(infix_operator) e2 = expr
-      { mkinfix e1 op e2 }
+      { mkexp ~loc:$sloc (mkinfix e1 op e2) }
```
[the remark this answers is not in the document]
ccasin: That's exactly right - this is just a small refactor, with a corresponding change in the definition of `fun_expr` where what used to be `| mkexp(expr_)` became `| expr_`.

ccasin:
@riaqn I think this is ready for another look.

Regarding "labeled" vs "labelled", I had a quick look through the compiler.  It seems for labeled function arguments, "labelled" is used for internal constructors and function, but the external interface is inconsistent, with "labeled" being used more often in comments and the manual.

Since it's already inconsistent, my instinct is to leave this alone.  If I were going to change it, I'd argue for keeping the american spelling in the external interface/documentation, as that seems to be what is (mostly) done for labeled arguments, and changing only internals.

riaqn (approved):

PR #1926, opened 2023-10-12
Middle and backend support for unboxed float/int32/int64/nativeint arrays

mshinwell:
This adds the arrays of the following element layouts to `Lambda` array kinds:
- unboxed `float`
- unboxed `int32`
- unboxed `int64`
- unboxed `nativeint`

It then threads this through the middle end and backend.  It should support all the usual things e.g. static allocation.  There are still two things missing **[update: now both fixed - see below]**:
- implementations at the Cmm level for the 32-bit arrays, which need packing
- it uses `Abstract_tag` instead of `Custom_tag` just for the moment.

There are some places in flambda2 where we could probably now factor out some code.


ocaml/typing/rec_check.ml
```diff
@@ -526,6 +526,9 @@ let array_mode exp = match Typeopt.array_kind exp with
   | Lambda.Paddrarray | Lambda.Pintarray ->
     (* non-generic, non-float arrays act as constructors *)
     Guard
+  | Lambda.Punboxedfloatarray | Lambda.Punboxedintarray _ ->
+    (* XXX mshinwell: is this correct? *)
```
goldfirere, line 530: Yes. Looks good to me.

goldfirere (approved):
Only file owned by me.

goldfirere:
Who is going to review this? @mshinwell and I discussed yesterday that 1) there is no way to test the new codepaths in this PR and 2) we should merge anyway. Tests will come when we add frontend support. But first some review would be good, I think.

mshinwell:
@Gbury is reviewing

ocaml/lambda/lambda.mli
```diff
@@ -279,6 +285,8 @@ and block_shape =
 and boxed_integer = Primitive.boxed_integer =
     Pnativeint | Pint32 | Pint64
 
+and unboxed_integer = boxed_integer
```
Gbury, line 288: maybe it's worth re-exporting the constructors here for more clarity, else this type declaration is a bit confusing.

ocaml/lambda/lambda.ml
```diff
@@ -283,22 +283,30 @@ and block_shape =
...
   | Paddrarray_set of modify_mode
   | Pintarray_set
   | Pfloatarray_set
+  | Punboxedfloatarray_set
+  | Punboxedintarray_set of unboxed_integer
 
 and boxed_integer = Primitive.boxed_integer =
     Pnativeint | Pint32 | Pint64
 
+and unboxed_integer = boxed_integer
```
Gbury, line 308: same remark: re-exporting the constructors would make the definition clearer

middle_end/flambda2/to_cmm/to_cmm_static.ml
```diff
@@ -49,6 +49,18 @@ let rec static_block_updates symb env res acc i = function
...
       static_block_updates symb env res acc (i + 1) r)
 
+let rec static_unboxed_int_array_updates symb env res acc i = function
+  | [] -> env, res, acc
+  | sv :: r -> (
+    match (sv : _ Or_variable.t) with
+    | Const _ -> static_unboxed_int_array_updates symb env res acc (i + 1) r
+    | Var (var, dbg) ->
+      let env, res, acc =
+        C.make_update env res dbg Word_int ~symbol:(C.symbol ~dbg symb) var
```
Gbury, line 59: This `Word_int` seems wrong and unprecise (but mostly wrong, e.g. for int32). You should probably take a `Cmm.memory_chunk` as argument of the `static_unboxed_int_array_updates` and pass it instead of `Word_int`.

middle_end/flambda2/to_cmm/to_cmm_static.ml
```diff
@@ -185,6 +197,37 @@ let static_const0 env res ~updates (bound_static : Bound_static.Pattern.t)
     let float_array = C.emit_float_array_constant sym static_fields in
     let env, res, e = static_float_array_updates sym env res updates 0 fields in
     env, R.update_data res float_array, e
+  | Block_like s, Immutable_int64_array fields ->
+    (* XXX mshinwell: as elsewhere, use custom blocks *)
```
Gbury, line 201: This needs to be done and not stay a `XXX` comment: the primitives operating on unboxed arrays will use the helpers from `cmm` which assume that the representation is a custom block, so this current code will very likely produce segfaults and/or memory corruption.
If that code is supposed to be unreachable, then a fatal error would be preferable.

middle_end/flambda2/to_cmm/to_cmm_static.ml
```diff
@@ -185,6 +197,37 @@ let static_const0 env res ~updates (bound_static : Bound_static.Pattern.t)
...
+        fields
+    in
+    let block = C.emit_block sym header static_fields in
+    let env, res, updates =
+      static_unboxed_int_array_updates sym env res updates 0 fields
+    in
+    env, R.set_data res block, updates
+  | Block_like s, Immutable_nativeint_array fields ->
+    let sym = R.symbol res s in
+    let header = C.black_block_header Obj.abstract_tag (List.length fields) in
```
Gbury, line 217: same remark, you need to use the custom block representation here

Gbury (changes requested):

Gbury:
Closing this one since it is superseded by #2185

PR #1927, opened 2023-10-12
Turn `Poll_and_jump` terminator into `Poll` basic instruction

xclerc:
(as per title)

### DIFF
--- a/backend/cfg/cfg_to_linear_desc.ml
+++ b/backend/cfg/cfg_to_linear_desc.ml
@@ -52,5 +52,6 @@ let from_basic (basic : basic) : Linear.instruction_desc =
         Iname_for_debugger
           { ident; which_parameter; provenance; is_assignment; regs }
       | Dls_get -> Idls_get
+      | Poll -> Ipoll { return_label = None }
     in
     Lop op
--- a/backend/amd64/proc.ml
+++ b/backend/amd64/proc.ml
@@ -470,6 +470,7 @@ let destroyed_at_basic (basic : Cfg_intf.S.basic) =
     [| rax |]
   | Op (Specific (Irdtsc | Irdpmc)) ->
     [| rax; rdx |]
+  | Op Poll -> destroyed_at_alloc_or_poll
   | Op (Move | Spill | Reload
        | Const_int _ | Const_float _ | Const_symbol _ | Const_vec128 _
        | Stackoffset _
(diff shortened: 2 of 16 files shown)


backend/cfg/cfg_to_linear_desc.ml
```diff
@@ -52,5 +52,6 @@ let from_basic (basic : basic) : Linear.instruction_desc =
         Iname_for_debugger
           { ident; which_parameter; provenance; is_assignment; regs }
       | Dls_get -> Idls_get
+      | Poll -> Ipoll { return_label = None }
```
gretay-js, line 55: `Linearize` has [an optimization for back-edges](https://github.com/ocaml-flambda/flambda-backend/blob/74131893886b362d8a7676e4ba49905afae41238/backend/linearize.ml#L191-L195) for this case. I am not sure if we are getting it on `main` with `Cfg`, depends on how good our `cfgize` and fallthrough elimination passes are, but this PR is not going to have it. It may be  hard to fix as part of this PR, but it would be nice to at least have a test and open an issue to track this.

backend/cfg/linear_to_cfg.ml
```diff
@@ -603,10 +603,8 @@ let rec create_blocks (t : t) (i : L.instruction) (block : C.basic_block)
       DLL.add_end block.body (create_instruction t desc i ~stack_offset);
       let stack_offset = stack_offset + bytes in
       create_blocks t i.next block ~stack_offset ~traps
-    | Ipoll { return_label = None } ->
-      terminator_fallthrough (fun return_label -> Poll_and_jump return_label)
-    | Ipoll { return_label = Some return_label } ->
-      terminator (C.Poll_and_jump return_label)
+    | Ipoll { return_label = None } -> basic C.Poll
+    | Ipoll { return_label = Some _return_label } -> assert false (* XXX *)
```
gretay-js, line 607: this is fine for now, this pass will be goin soon #2020 .

gretay-js (approved):
`Poll` cannot raise after #1804 and #2007, so it can be `basic` instruction in `Cfg`.

backend/cfg/cfg_to_linear_desc.ml
xclerc: (I have created https://github.com/ocaml-flambda/flambda-backend/issues/2228 to track it.)

PR #1932, opened 2023-10-16
Add --enable-dev configure option etc.

mshinwell:
This adds a new configure option `--enable-dev` which is intended for day-to-day development, not for building production compilers.  It sets `-opaque` for the whole build except the stdlib.  When configuring for Flambda 2 it also uses classic mode for the main compiler build (this is not set for the boot compiler build as it is assumed that is a non-Flambda 2 compiler).

This PR also uses dune's `dev` build profile for the boot compiler build.  This uses higher warning levels than we had previously and so there are a number of small fixes scattered around for that.

In addition the optimization levels specified in `dune` files have been cleaned up for Flambda 2 builds.  Whether in dev mode or not, the stdlib is built at `-O3` (any compilation time difference will be minimal).  When not in dev mode, the main compiler build operates at `-O3` throughout, removing the need for explicit occurrences of `-O3` in `dune` files.  The latter is indeed important, as such occurrences would override the classic mode setting.


backend/debug/dwarf/dwarf_flags/dune
```diff
@@ -4,6 +4,6 @@
   (name dwarf_flags)
   (wrapped true)
   (flags (:standard -principal))
-  (ocamlopt_flags (:standard -O3))
+  (ocamlopt_flags (:standard))
```
lukemaurer, line 7: We should just delete the whole `ocamlopt_flags` clause (here and anywhere it's just `(:standard)` now).

mshinwell: We should, but I think maybe in a follow-up PR (just makes it very clear what the changes are here)

lukemaurer: Sure, makes sense

lukemaurer (approved):

stedolan:
I'm not convinced by this. The version just merged into the `merge5` branch of ocaml-jst uses `dev` unconditionally for the boot compiler and `release` unconditionally for the installed compiler, which seems simpler than a configure option.

mshinwell:
As I was saying to @stedolan elsewhere, there is a significant difference with `--enable-dev` on this PR.  For a normal `make`:
```
real	3m59.993s
user	13m32.561s
sys	1m28.581s

--enable-dev
real	1m4.069s
user	6m9.862s
sys     1m15.180s
```
Note that at the moment (prior to this PR) only a subset of the compiler code is built at `-O3`.  Hopefully this change will improve the performance of the whole compiler, I will see if I can get some numbers.

TheNumbat (approved):

PR #1746, opened 2023-08-15
Split `Import_info.t` into `Import_info.{Intf,Impl}.t`

lukemaurer:
**Depends on #1933.**

This improves clarity, since every `Import.t` is statically either an interface import or an implementation import and they don't support quite the same operations. In particular, every implementation import (being an import of a .cmo/x file) must have a fully-specified compilation unit, whereas an interface import might only have a name (with unknown pack prefix).

Produced by reverting commit 4bd754402c2078fae30dbc15c60b2a3a1a4798ee from the 'instance-typing' branch.

PR #1933, opened 2023-10-16
Copy `Import_info` as `Import_info.Intf` and `Import_info.Impl`

lukemaurer:
PR #1746 will change the APIs and implementations to reflect the differences in the datatypes. (Currently, the major difference is that an `Impl.t` always has a full compilation unit, but parameterised libraries will add more differences.)


lukemaurer:
> Converting to draft, as this work may be getting rolled into #1753 .

(or at any rate this PR's direction may be influenced by that one)

lukemaurer:
Superseded by #2552.

PR #1934, opened 2023-10-16
Improve parsing for labeled tuple patterns

ccasin:
review: @riaqn


ocaml/parsing/parser.mly
```diff
@@ -3499,17 +3499,79 @@ pattern_no_exn:
...
+  | mkpat(
+      self COLONCOLON error
         { expecting $loc($3) "pattern" }
     | self BAR pattern
         { Ppat_or($1, $3) }
     | self BAR error
-        { expecting $loc($3) "pattern" }
-  ) { $1 }
+        { expecting $loc($3) "pattern" }  ) { $1 }
+;
```
riaqn, line 3519: some nit pick: seems like we can reduce the diff here?

riaqn, line 3519: some nit pick: seems like we can reduce the diff here?

ocaml/parsing/parser.mly
```diff
@@ -3614,64 +3672,11 @@ simple_delimited_pattern:
...
-      { $1, Open }
-  // Partial patterns are allowed to be length-one
-  | ltuple_component_pat COMMA DOTDOT
-      { [$1], Open }
-
 %inline pattern_semi_list:
   ps = separated_or_terminated_nonempty_list(SEMI, pattern)
     { ps }
 ;
+
```
riaqn, line 3679: extra line

riaqn (approved):
some minor points, otherwise fine to merge.

ccasin:
@riaqn nice catches, thanks.  Will merge when CI passes.

ocaml/parsing/parser.mly
```diff
@@ -3499,9 +3499,18 @@ pattern_no_exn:
...
+      { let closed, pats = $1 in
+        if    closed = Closed
+           && List.for_all (fun (l,_) -> Option.is_none l) pats
+        then
+          mkpat ~loc:$sloc (Ppat_tuple(List.rev_map snd pats))
+        else
+          ppat_lttuple $sloc (List.rev pats) closed
+      }
+  | mkpat(
+      self COLONCOLON error
```
riaqn, line 3513: Sorry I wasn't clear - I meant that currently there are two `mkpat` (but there was only one) - I think if we keep it that way, we can minimize the diff to only removing `pattern_comma_list` and adding `reversed_labeled_tuple_pattern`.

ccasin: Right - I think I see, and agree that is better.  Does it look right now?

riaqn: yeah I think it's perfect now - thanks for fullfilling my nits.

PR #1936, opened 2023-10-16
Zero alloc: add payload "opt" and "-zero-alloc-check-opt" flag

gretay-js:
Support `[@zero_alloc opt]` annotation. The intended use is to annotate functions that are "zero-alloc"  in an optimized build but may allocate in a non-optimized build. This annotation is checked when `-zero-alloc-check-opt` compilation flag is set (off by default).  

Using both `[@zero_alloc][@zero_alloc opt]` on the same function is not allowed. Instead, optimized build should set both `-zero-alloc-check` and `-zero-alloc-check-opt` flags. If a function is expected to be `zero_alloc` in all builds, it should be annotated with `[@zero_alloc]` only, without `opt`.   

Payload `opt` can be combined with `strict` payload but not with `assume`.

The new annotation is resolved on the way out of `Lambda`, so that backend and middle-end do not need to know about `opt`. It is hard to resolve this information earlier because of the way function attributes are handled using two separate calls to `Translattribute.add_function_attributes`.

### DIFF
--- a/backend/checkmach.ml
+++ b/backend/checkmach.ml
@@ -1162,7 +1162,13 @@ end
 module Spec_zero_alloc : Spec = struct
   let property = Cmm.Zero_alloc
 
-  let enabled () = !Clflags.zero_alloc_check || !Clflags.zero_alloc_check_opt
+  let enabled () =
+    (* Checkmach no longer distinguishes between opt and default checks. *)
+    match !Clflags.zero_alloc_check with
+    | No_check -> false
+    | Check_default -> true
+    | Check_all -> true
+    | Check_opt_only -> true
 
   (* Compact the mapping from function name to Value.t to reduce size of Checks
      in cmx and memory consumption Compilenv. Different components have
--- a/backend/checkmach.ml
+++ b/backend/checkmach.ml
@@ -1162,7 +1162,7 @@ end
 module Spec_zero_alloc : Spec = struct
   let property = Cmm.Zero_alloc
 
-  let enabled () = !Clflags.zero_alloc_check
+  let enabled () = !Clflags.zero_alloc_check || !Clflags.zero_alloc_check_opt
 
   (* Compact the mapping from function name to Value.t to reduce size of Checks
      in cmx and memory consumption Compilenv. Different components have
--- a/backend/cmm_helpers.ml
+++ b/backend/cmm_helpers.ml
@@ -4329,8 +4329,10 @@ let transl_attrib : Lambda.check_attribute -> Cmm.codegen_option list = function
           never_returns_normally;
           loc
         } ]
-  | Check { property; strict; loc } ->
-    [Check { property = transl_property property; strict; loc }]
+  | Check { property; strict; loc; opt } ->
+    if Lambda.is_check_enabled ~opt property
+    then [Check { property = transl_property property; strict; loc }]
+    else []
 
 let kind_of_layout (layout : Lambda.layout) =
   match layout with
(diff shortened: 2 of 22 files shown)


xclerc (approved):

ccasin (approved):
Looks fine, but a couple minor comments:

I think the interface design here is slightly surprising, in that if I pass `-zero-alloc-check-opt` but not `-zero-alloc-check`, the non-`opt` functions don't get checked.  This seems like an easy mistake to make and accidentally not check most of the functions silently.  I think I'd prefer it if either `-zero-alloc-check-opt` impled `-zero-alloc-check`, or it issued a warning if passed without `-zero-alloc-check`.  I don't feel very strongly about this, though, since in reality the build system will pass these flags, so it's fine to merge as is if you disagree.

Also, the parsing in `parse_property_attribute` is starting to get a little unpleasant.  I think it's not worth doing some thing about now, but will be worth rationalizing a bit if more modifiers are added.

gretay-js:
> Looks fine, but a couple minor comments:
> 
> I think the interface design here is slightly surprising, in that if I pass `-zero-alloc-check-opt` but not `-zero-alloc-check`, the non-`opt` functions don't get checked. This seems like an easy mistake to make and accidentally not check most of the functions silently. I think I'd prefer it if either `-zero-alloc-check-opt` impled `-zero-alloc-check`, or it issued a warning if passed without `-zero-alloc-check`. I don't feel very strongly about this, though, since in reality the build system will pass these flags, so it's fine to merge as is if you disagree.
 
I agree but the reason I made them separate flags is that it is possible to pass `-no-zero-alloc-check` or `-no-zero-alloc-check-opt` which turns off only one of them and end up in a state that is not covered..
 
> Also, the parsing in `parse_property_attribute` is starting to get a little unpleasant. I think it's not worth doing some thing about now, but will be worth rationalizing a bit if more modifiers are added.

what do you propose there?

xclerc:
I think I broadly agree with Chris. My first thought
after reading the pull request was that there is
probably no strong reason to have the command-line
interface follow closely the annotation. If this is true,
a more explicit command-line interface could be
a single `-zero-alloc-check {all|all-but-opt|...}` or
 `-zero-alloc-check {standard-build|optimized-build}`.

gretay-js:
@ccasin @xclerc  I modified `-zero-alloc-check` flag to take arguments `{ default | all | none | opt }`. 
`opt` is useful for tests.

ccasin (approved):
I took a look at the new commits and agree this is an improvement.  Thanks for the nice suggestion, @xclerc

xclerc:
@ccasin does it address your concern about a misuse
of `opt` from the command-line? I know I tend to be
verbose, but maybe `opt-excluding-default` instead of
`opt`?

mshinwell (approved):
check_attribute.ml only

ccasin:
> @ccasin does it address your concern about a misuse of `opt` from the command-line? I know I tend to be verbose, but maybe `opt-excluding-default` instead of `opt`?

Good point.  I was kind of imagining Greta added this only for testing and maybe we wouldn't tell people about it.  But maybe a name like `-opt-only`, `-opt-excluding-default`, or `-opt-debug`, or ... would be clearer.

gretay-js:
I added detailed help message that attempts to clarify how to use this flag. I think that even the longer names like `opt-only` and `opt-excluding-default` may be confusing because the flag can be passed twice and the expected behavior may be additive (even though it is not for other flags).

xclerc:
But with the new set of options, is there still a
need for the flag to be passed twice?

gretay-js:
> But with the new set of options, is there still a need for the flag to be passed twice?
no but it may be passed twice, e.g., once from a top level default and another from the ocamlopt_flags field in a dune file.

xclerc:
Sure, but the semantics is not additive, right?

gretay-js:
yes, the semantics is not additive.

xclerc:
Sorry, I don't see the confusion you alluded to.
There are few flags with an additive semantics,
so I would expect most users to expect replacement
semantics.

PR #1937, opened 2023-10-17
Remove flambda-backend copies of Closure, Flambda 1, Clambda, Cmmgen, etc.

mshinwell:
Unfortunately since we still have the copy from upstream in `ocaml/`, we can't get rid of the "`transl_store`" compilation model in `Translmod`.


lthls (approved):
Nothing obviously wrong, but I haven't tried to find other places that might be missing from this patch.

mshinwell:
We're going to wait another couple of weeks before merging this.

mshinwell:
This should now remove all of Closure, Flambda 1, `Clambda`, `Cmmgen` and the unused functions from `Cmm_helpers` (it's possible that one or two might have slipped through the net for the latter, but the vast majority I think have been removed).

The structured constant code that used to be in `Clambda`, which is required by `Cmmgen_state` (that is still used), has now gone directly into `Cmmgen_state`.  The constructors are now called `Const_...` instead of `Uconst_...`.

mshinwell:
I've asked @Gbury if he can review this

mshinwell:
A follow-up patch can do the slightly more intricate work of removing `Cregion`, non-delayed trap handlers, etc. that are never used by Flambda 2.

Gbury (approved):

mshinwell:
This PR makes the `--enable-middle-end` configure flag optional.  The only non-`flambda2` value permitted, if it is supplied, is `upstream-closure`.
The test exclusion list has been updated as discussed with @Gbury

PR #1943, opened 2023-10-17
Support for [@error_message] attribute on layout annotations

alanechang:
With this new attribute, it's now possible to attach custom error messages to layout constraints:

```ocaml
let f x =
  ignore ((x : (_ : value)[@error_message "Custom message"]));
  Float_u.to_float x
[%%expect{|
Line 3, characters 19-20:
3 |   Float_u.to_float x
                       ^
       but an expression was expected of type float#
         it is the primitive float64 type float#.
         the [@error_message] attribute: Custom message.
|}]
```

Main motivation here is for ppx authors to be able to signal to clients when their ppx doesn't work with unboxed types.

The current parser doesn't have support for attaching attributes to universal variable layout annotations. Thus, `attrs` are ignored in the `Ltyp_poly` case of `transl_type_aux_jst_layout`.

### DIFF
--- a/ocaml/testsuite/tests/typing-layouts/error_message_attr.ml
+++ b/ocaml/testsuite/tests/typing-layouts/error_message_attr.ml
@@ -1,0 +1,156 @@
+(* TEST
+   flags = "-extension layouts_alpha"
+ * expect
+*)
+
+module Float_u = Stdlib__Float_u
+
+[%%expect{|
+module Float_u = Stdlib__Float_u
+|}]
+
+(* Needs a string payload *)
+
+let f (v: float#): ((_ : value)[@error_message]) = v
+[%%expect{|
+Line 1, characters 31-47:
+1 | let f (v: float#): ((_ : value)[@error_message]) = v
+                                   ^^^^^^^^^^^^^^^^
+Warning 47 [attribute-payload]: illegal payload for attribute 'error_message'.
+error_message attribute expects a string argument
+Line 1, characters 51-52:
+1 | let f (v: float#): ((_ : value)[@error_message]) = v
+                                                       ^
+Error: This expression has type float# but an expression was expected of type
+         ('a : value)
+       The layout of float# is float64, because
+         it is the primitive float64 type float#.
+       But the layout of float# must be a sublayout of value, because
+         of the annotation on the wildcard _ at line 1, characters 20-31.
+|}]
+
+let f (v: float#): ((_ : value)[@error_message 1]) = v
+[%%expect{|
+Line 1, characters 31-49:
+1 | let f (v: float#): ((_ : value)[@error_message 1]) = v
+                                   ^^^^^^^^^^^^^^^^^^
+Warning 47 [attribute-payload]: illegal payload for attribute 'error_message'.
+error_message attribute expects a string argument
+Line 1, characters 53-54:
+1 | let f (v: float#): ((_ : value)[@error_message 1]) = v
+                                                         ^
+Error: This expression has type float# but an expression was expected of type
+         ('a : value)
+       The layout of float# is float64, because
+         it is the primitive float64 type float#.
+       But the layout of float# must be a sublayout of value, because
+         of the annotation on the wildcard _ at line 1, characters 20-31.
+|}]
+
+(* Ltyp_var { name = None; layout } case *)
+let f (v: float#): ((_ : value)[@error_message "Custom message"]) = v
+[%%expect{|
+Line 1, characters 68-69:
+1 | let f (v: float#): ((_ : value)[@error_message "Custom message"]) = v
+                                                                        ^
+Error: This expression has type float# but an expression was expected of type
+         ('a : value)
+       The layout of float# is float64, because
+         it is the primitive float64 type float#.
+       But the layout of float# must be a sublayout of value, because
+         of the annotation on the wildcard _ at line 1, characters 20-31.
+         Custom message
+|}]
+
+let f x =
+  ignore ((x : (_ : value)[@error_message "Custom message"]));
+  Float_u.to_float x
+[%%expect{|
+Line 3, characters 19-20:
+3 |   Float_u.to_float x
+                       ^
+Error: This expression has type ('a : value)
+       but an expression was expected of type float#
+       The layout of float# is float64, because
+         it is the primitive float64 type float#.
+       But the layout of float# must be a sublayout of value, because
+         of the annotation on the wildcard _ at line 2, characters 15-26.
+         Custom message
+|}]
+
+(* Ltyp_var { name = Some name; layout } case *)
+module type a = sig
+  type ('a : float64) t = 'a
+  val f : (('a : value)[@error_message "Custom message"]) -> 'a t
+end
+
+[%%expect{|
+Line 3, characters 61-63:
+3 |   val f : (('a : value)[@error_message "Custom message"]) -> 'a t
+                                                                 ^^
+Error: This type ('a : value) should be an instance of type ('b : float64)
+       The layout of 'a is value, because
+         of the annotation on the type variable 'a.
+         Custom message
+       But the layout of 'a must overlap with float64, because
+         of the definition of t at line 2, characters 2-28.
+|}]
+
+
+(* Ltyp_alias { aliased_type; name; layout } case *)
+
+(* First call to [layout_of_annotation] in [transl_type_alias] *)
+module type a = sig
+  type t : float64
+  val f : 'a -> t -> (t as ('a : value)[@error_message "Custom message"])
+end
+[%%expect{|
+Line 3, characters 33-38:
+3 |   val f : 'a -> t -> (t as ('a : value)[@error_message "Custom message"])
+                                     ^^^^^
+Error: Bad layout annotation:
+         The layout of t is float64, because
+           of the definition of t at line 2, characters 2-18.
+         But the layout of t must be a sublayout of value, because
+           of the annotation on the type variable 'a.
+           Custom message
+|}]
+
+(* Second call to [layout_of_annotation] in the Not_found case
+   of [transl_type_alias] *)
+module type a = sig
+  type t : float64
+  val f : t -> (t as ('a : value)[@error_message "Custom message"])
+end
+[%%expect{|
+Line 3, characters 16-33:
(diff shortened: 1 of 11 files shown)


ocaml/testsuite/tests/typing-layouts/error_message_attr.ml
```diff
@@ -0,0 +1,118 @@
...
+[%%expect{|
+Line 1, characters 68-69:
+1 | let f (v: float#): ((_ : value)[@error_message "Custom message"]) = v
+                                                                        ^
+Error: This expression has type float# but an expression was expected of type
+         ('a : value)
+       The layout of float# is float64, because
+         it is the primitive float64 type float#.
+       But the layout of float# must be a sublayout of value, because
+         the [@error_message] attribute: Custom message.
```
goldfirere, line 23: Hmm.... this isn't exactly right. It's note the `[@error_message]` that's causing the result to be a `value`.

Maybe something like

```
  of a layout constraint at line 1, columns xyz.
  Custom message
```

This makes it so that the constraint is what forces the layout, and just appends `Custom message` below. That seems to give the most flexibility to custom message authors.

ocaml/testsuite/tests/typing-layouts/error_message_attr.ml
```diff
@@ -0,0 +1,118 @@
...
+Line 3, characters 26-31:
+3 |   val f : t -> (t as (_ : value)[@error_message "Custom message"])
+                              ^^^^^
+Error: Bad layout annotation:
+         The layout of t/2 is float64, because
+           of the definition of t at line 2, characters 2-18.
+         But the layout of t/2 must be a sublayout of value, because
+           of the annotation on the wildcard _ at line 3, characters 26-31.
+|}]
+(* CR layouts: Bug in Jane_syntax is duplicating the attrs here *)
```
goldfirere, line 116: Crumbs. But actually I think the new approach to extract the attributes will fix this. Maybe. Try that and resolve this comment if it does actually fix the problem.

ocaml/testsuite/tests/typing-layouts/error_message_attr.ml
```diff
@@ -0,0 +1,118 @@
...
+
+(* Currently it's not possible to attach attributes to Ltyp_poly *)
```
goldfirere, line 118: Yes. I think this is fine.

ocaml/testsuite/tests/warnings/w53.ml
```diff
@@ -355,3 +355,21 @@ module TestNewtypeAttr = struct
...
+end
+
+module TestErrorMessageStruct = struct
+  type 'a t1 = 'a [@@error_message ""] (* rejected *)
+  type s1 = Foo1 [@error_message ""] (* rejected *)
+  let x : int = 42 [@@error_message ""] (* rejected *)
+
+  external y : (int [@error_message ""]) -> (int [@error_message ""]) = "x" "y" (* rejected *)
+  external z : int -> int = "x" "y" [@@error_message ""] (* rejected *)
+end
```
goldfirere, line 375: After redoing the code in Builtin_attributes, a repeated `[@error_message ...]` should trigger warning 53, so add a test here.

ocaml/typing/layouts.ml
```diff
@@ -323,6 +323,7 @@ module Layout = struct
     | Type_wildcard of Location.t
 
   type creation_reason =
+    | Err_msg_attr of string
```
goldfirere, line 326: Often better to avoid abbreviations: code is read *much* more than it is written.

ocaml/typing/layouts.ml
```diff
@@ -993,16 +996,21 @@ module Layout = struct
 
   let equate = equate_or_equal ~allow_mutation:true
 
+  let score_reason reason =
```
goldfirere, line 999: Good function. Please add a comment describing what it does and why it's a good idea.

goldfirere, line 999: Style nitpick: I think it's nicer to use `= function` instead of naming `reason` and then matching on it.

ocaml/typing/layouts.ml
```diff
@@ -993,16 +996,21 @@ module Layout = struct
...
+    | Creation (Concrete_creation _) -> -1
+    | _ -> 0
+
   let combine_histories reason lhs rhs =
     if flattened_histories
     then begin match sub_desc (get_internal lhs.layout) (get_internal rhs.layout) with
       | Sub -> lhs.history
       | Not_sub -> rhs.history  (* CR layouts: this will be wrong if we ever have a non-trivial meet in the layout lattice *)
-      | Equal -> begin match lhs.history, rhs.history with
+      | Equal -> begin match score_reason lhs.history, score_reason rhs.history with
```
goldfirere, line 1010: I think this is just `if score_reason lhs.history >= score_reason rhs.history`?

ocaml/typing/typetexp.ml
```diff
@@ -541,6 +541,12 @@ let check_arg_type styp =
     | _ -> ()
   end
 
+let layout_of_annotation_or_attr annot_ctx attrs layout =
+  let error_message_attr_opt = Builtin_attributes.error_message_attr attrs in
+  match error_message_attr_opt with
+  | Some msg -> Layout.of_const ~why:(Err_msg_attr msg) layout.txt
```
goldfirere, line 547: Careful: this skips the extension check in `of_annotation`. Might be best to make a new constructor of `annotation_context`:

```ocaml
| With_error_message of string * annotation_context
```

that would wrap the actual context (like `Type_variable`).

ocaml/typing/typetexp.ml
```diff
@@ -541,6 +541,12 @@ let check_arg_type styp =
     | _ -> ()
   end
 
+let layout_of_annotation_or_attr annot_ctx attrs layout =
```
goldfirere, line 544: The `_or_attr` here confused me: it's saying that the layout comes from the attribute, but that's not quite accurate. I would just go with `layout_of_annotation`.

ocaml/typing/typetexp.ml
```diff
@@ -1020,10 +1026,10 @@ and transl_type_alias env policy mode alias_loc styp name_opt layout_annot_opt =
       with Not_found ->
         if !Clflags.principal then begin_def ();
         let layout =
-          Layout.(of_annotation_option_default
-            ~default:(any ~why:Dummy_layout)
-            ~context:(Type_variable alias)
-            layout_annot_opt)
+          Option.fold ~none:(Layout.any ~why:Dummy_layout)
```
goldfirere, line 1029: I think this would be easier to read with a `match layout_annot_opt with ...` construction instead of `Option.fold`.

ocaml/testsuite/tests/warnings/w53.ml
alanechang: The layout annotation feature is behind `-extension layouts`. Not sure if we want to add that flag to this file. Added tests to `error_message_attr_w53.ml` with a CR about moving it to `w53.ml`.

ocaml/testsuite/tests/typing-layouts/error_message_attr.ml
alanechang: The new approach does fix this. Something to note is that since the duplicated attributes are not registered, you also don't get the `warning 53` misplaced-attribute error from them.

goldfirere: Right. That's exactly what I was hoping for. Sometimes, two wrongs _do_ make a right. :)

ocaml/testsuite/tests/warnings/w53.ml
goldfirere: Great.

ocaml/testsuite/tests/typing-layouts/error_message_attr.ml
```diff
@@ -20,7 +58,8 @@ Error: This expression has type float# but an expression was expected of type
        The layout of float# is float64, because
          it is the primitive float64 type float#.
        But the layout of float# must be a sublayout of value, because
-         the [@error_message] attribute: Custom message.
+         of the annotation on the wildcard _ at line 1, characters 20-31
+         Custom message.
```
goldfirere, line 62: Could we get the `.` to appear on the previous line? I see how that might be hard. If it is hard, then it's OK to keep it here, but also include it on the previous line (which should not be hard).

Why I want to get rid of the final dot: maybe the custom message is actually a question like "Did you mean to use match%optional_u?"

ocaml/testsuite/tests/typing-layouts/error_message_attr.ml
```diff
@@ -50,15 +90,16 @@ Line 3, characters 61-63:
                                                                  ^^
 Error: This type ('a : value) should be an instance of type ('b : float64)
        The layout of 'a is value, because
-         the [@error_message] attribute: Custom message.
+         of the annotation on the type variable 'a
+         Custom message.
```
goldfirere, line 94: People tend to read the beginning and end of error messages. Putting the custom message in the middle seems like it might get lost. But I think fixing this would be hard. If you agree, then let's not for now. But maybe you see an easy way of pulling this out.

goldfirere (approved):

ocaml/testsuite/tests/typing-layouts/error_message_attr.ml
alanechang: By having a special match expression for `With_error_message` in `format_flattened_history`, we can have the message after the `.`

ocaml/testsuite/tests/typing-layouts/error_message_attr.ml
alanechang: I agree it does seem hard to fix this. Will have to add something at every `format_history` call site (two places at the moment) to check for `With_error_message`.

goldfirere:
Went to merge this. It looks like it needs a manual rebase...

alanechang:
Just rebased it. The PR is currently merging into [aec/layout-msg-test-coverage](https://github.com/ocaml-flambda/flambda-backend/tree/aec/layout-msg-test-coverage). Might be easier to get #1924 merged before this one. Or I can rebase it again on top of [layout-histories](https://github.com/ocaml-flambda/flambda-backend/tree/layout-histories).

PR #1947, opened 2023-10-18
Erase callees in classic mode

Ekdohibs:


mshinwell:
@lthls would you be able to review this?

lthls:
Eventually, yes. I'm a bit busy so it might have to wait until after next week.

lthls:
I've reviewed the PR. I'm not very fond of the current state:
- C calls have a callee, even though it's the main case where having a callee doesn't make sense;
- The handling in Simplify is bad in several places

I'm willing to approve that if it's urgent, as it does not seem to introduce bugs in the current state. But ideally I'd like to take a bit of time to at least clean up `simplify_apply_expr.ml` a little bit.

lthls (approved):
I've added a commit to clean up the handling of no-callee applications in Simplify.
I'm approving, but I think we should follow up with additional PRs to fix some related parts:
- C calls shouldn't have a callee
- `simplify_direct_full_application` should read `params_arity`, `result_arity` and `result_types` from its callee instead of taking parameters. This is particularly important for `result_types` as we currently use the wrong ones (the less precise ones).

lthls:
The MacOS CI failure doesn't look related, but I don't know if it's something that can be fixed by a restart or needs to be properly investigated.

mshinwell:
@lthls if not producing PRs straight away, please leave CRs for the above

lthls:
There's already a CR in `Apply_expr.mli` about not using callees for C calls. I will add one for `simplify_direct_full_application`.

PR #1948, opened 2023-10-18
scripts/merge-ocaml-jst-with-conflicts

mshinwell:
This should import `ocaml-jst` even if there are conflicts.


mshinwell:
I think we can skip review on this.

PR #1949, opened 2023-10-18
OCaml 5 minus

mshinwell:
This is (will be...) 5.1 running on the 4.14 runtime, although the 5.1 runtime is present in the tree.

We'll update this regularly using the script in https://github.com/ocaml-flambda/flambda-backend/pull/1948


mshinwell:
PRs fixing flambda-backend for OCaml 5 minus should use this PR's branch (`ocaml-5-minus`) as their base.

lukemaurer:
Mark and I discussed 00550a34224aacd71408a1985877036db2d4000d and decided we should instead just rename `hash` to `seeded_hash` and keep the same definition (ignoring the seed), since that appears to be all that the version 5 stdlib changed.

ocaml/testsuite/tests/typing-gadts/test.ml
```diff
@@ -689,9 +689,9 @@ let f : type a b. (a,b) eq -> (<m : a; ..> as 'c) -> (<m : b; ..> as 'c) =
 ;; (* fail *)
 [%%expect{|
 type (_, _) eq = Eq : ('a, 'a) eq
-Line 3, characters 18-72:
-3 | let f : type a b. (a,b) eq -> (<m : a; ..> as 'c) -> (<m : b; ..> as 'c) =
-                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Lines 3-4, characters 4-15:
+3 | ....f : type a b. (a,b) eq -> (<m : a; ..> as 'c) -> (<m : b; ..> as 'c) =
+4 |   fun Eq o -> o
```
ncik-roberts: Note that this change, and the handful of others in this commit that have to do with location, were due to a stale parser file that is fixed in https://github.com/ocaml-flambda/flambda-backend/pull/1949/commits/9a71a4e2a2b35ee187cdf9bb9dec1ae386610d44. That commit undoes these test changes.

xclerc (approved):

PR #1951, opened 2023-10-19
Port OCaml 5 `asmcomp/` changes

xclerc:
As per title.

Based on #1959 (only 51d125bc17c8e73e978d47a6439350c6f024436d is new),
and builds `ocamloptcomp.cmxa` for `amd64`
(`arm64` not tested).


backend/amd64/proc.ml
```diff
@@ -60,12 +60,12 @@ let win64 = Arch.win64
    Under Unix:
      rdi, rsi, rdx, rcx, r8, r9: C function arguments
      xmm0 - xmm7: C function arguments
-     rbx, rbp, r12-r15 are preserved by C
+     r12-r15 are preserved by C
```
gretay-js, line 63: Where is this diff coming from? Why are `rbx` and `rbp` no longer preserved by C?

mshinwell: Hmm I looked at this on ocaml-jst, but I can't find where my diff has gone to.  Will update

backend/amd64/arch.mli
```diff
@@ -0,0 +1,81 @@
...
+(**************************************************************************)
+
+(* Machine-specific command-line options *)
+
+val command_line_options : (string * Arg.spec * string) list
+
+(* Specific operations for the AMD64 processor *)
+
+type addressing_mode =
+    Ibased of string * int              (* symbol + displ *)
```
gretay-js, line 24:
```suggestion
    Ibased of string * sym_global * int              (* symbol + displ *)
```

backend/amd64/proc.ml
mshinwell: This is what's needed: https://github.com/ocaml-flambda/ocaml-jst/pull/243

backend/amd64/arch.mli
gretay-js: we need to update the [dune](https://github.com/ocaml-flambda/flambda-backend/blob/51d125bc17c8e73e978d47a6439350c6f024436d/backend/dune#L21C2-L21C49) file to copy `ARCH/arch.mli` over to the parent directory.

backend/amd64/proc.ml
gretay-js: Thanks! What is the plan for the diff between BACKPORT comment and the corresponding code in ocaml-jst#243 that doesn't use rbx and rbp?

gretay-js (approved):

mshinwell:
Necessary changesets included in PR1949

PR #1952, opened 2023-10-19
Record *old* value of sort during backtracking

goldfirere:
This can't easily be tested, but this bug surfaced in my upcoming modal kinds patch. Thought it would be easier to review separately.

Review suggestion: @alanechang

### DIFF
--- a/ocaml/typing/typedecl.ml
+++ b/ocaml/typing/typedecl.ml
@@ -73,6 +73,7 @@ type error =
   | Deep_unbox_or_untag_attribute of native_repr_kind
   | Jkind_mismatch_of_type of type_expr * Jkind.Violation.t
   | Jkind_mismatch_of_path of Path.t * Jkind.Violation.t
+  | Jkind_mismatch_in_check_constraints of type_expr * Jkind.Violation.t
   | Jkind_sort of
       { kloc : jkind_sort_loc
       ; typ : type_expr
@@ -850,7 +851,7 @@ let rec check_constraints_rec env loc visited ty =
         | Unification_failure err ->
           raise (Error(loc, Constraint_failed (env, err)))
         | Jkind_mismatch { original_jkind; inferred_jkind; ty } ->
-          raise (Error(loc, Jkind_mismatch_of_type (ty,
+          raise (Error(loc, Jkind_mismatch_in_check_constraints (ty,
                               (Jkind.Violation.of_ (Not_a_subjkind
                                  (original_jkind, inferred_jkind))))))
         | All_good -> ()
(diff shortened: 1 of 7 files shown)


goldfirere:
This is ready for review now.

ocaml/typing/typedecl.ml
```diff
@@ -1548,8 +1548,6 @@ let transl_type_decl env rec_flag sdecl_list =
         raise (Error (loc, Type_clash (new_env, err))))
       checks)
     delayed_jkind_checks;
-  (* Check that constraints are enforced *)
-  List.iter2 (check_constraints new_env) sdecl_list decls;
   (* Check that all type variables are closed; this also defaults any remaining
```
alanechang, line 1551: Should this comment be updated to include something like `Defaulting also has to occur prior to check_constraints to prevent soundness bugs`?

ocaml/typing/typedecl.ml
```diff
@@ -2406,11 +2406,30 @@ let report_error ppf = function
...
+          "@[<v>Layout inference of mutually recursive datatypes is not@ \
+           clever enough to propagate layouts through variables in different@ \
+           declarations. It is also not clever enough to produce a good error@ \
+           message, so we'll say this instead:@ %a@ \
+           The fix will be to add a layout annotation on a variable, likely@ \
+           in the declaration including the location of this error.@]"
+          (Jkind.Violation.report_with_offender
+             ~offender:(fun ppf -> Printtyp.type_expr ppf ty)) violation
+      | None ->
       fprintf ppf "@[<v>Constraints are not satisfied in this type.@ ";
```
alanechang, line 2427: Should this be indented?

alanechang (approved):
Bug fix and the new error message look good! Left some minor suggestions.

ocaml/typing/typedecl.ml
goldfirere: Good point. Done.

goldfirere:
@alanechang please review the last two commits.

I understand we can't merge this until the OCaml 5 merge happens, which will hopefully be later this week.

ocaml/typing/typedecl.ml
```diff
@@ -2378,6 +2384,18 @@ let explain_unbound_single ppf tv ty =
         "case" (fun (lab,_) -> "`" ^ lab ^ " of ")
   | _ -> trivial ty
 
+let report_jkind_mismatch_in_check_constraints ppf ty violation =
+  fprintf ppf
+    "@[<v>Layout mismatch in final type declaration consistency check.@ \
+     This is most often caused by the fact that type inference is not@ \
+     clever enough to propagate layouts through variables in different@ \
+     declarations. It is also not clever enough to produce a good error@ \
+     message, so we'll say this instead:@  @[%a@]@ \
```
alanechang, line 2393:
```suggestion
     message, so we'll say this instead:@;<1 2>@[%a@]@ \
```
> `<1 2>` means one space if it fits on the same line, two if on a new line

Use break hint for indentation?

alanechang (approved):
The new changes look good and the error messages should be an improvement. Left one comment regarding indentation.

goldfirere:
I have not added @alanechang to CODEOWNERS because of #1985. But if I had, then his approval would be sufficient for merging this. Thus I am just going to override the settings to merge.

PR #1953, opened 2023-10-19
Change error message for non-value class lets

goldfirere:
The old way of producing this error required `of_sort_for_error`, which causes trouble in the modal kinds world. This kills off that terrible function.

Review: @ccasin


ccasin (approved):
LGTM

PR #1954, opened 2023-10-20
Support for [@error_message] attribute on Pexp_constraint

alanechang:
This PR follows #1943 and adds support for the `[@error_message]` attribute on `Pexp_constraint`.

It enables custom error messages on type mismatches:

```ocaml
let f (x : bool) = (x : int)[@error_message "custom message"]
[%%expect{|
Line 1, characters 20-21:
1 | let f (x : bool) = (x : int)[@error_message "custom message"]
                        ^
Error: This expression has type bool but an expression was expected of type
         int
       because [@error_message]: custom message
|}]
```

It's implemented using the explanation system already in place within `typecore.ml`. See added tests for its exact behavior.

### DIFF
--- a/ocaml/testsuite/tests/typing-layouts/error_message_attr.ml
+++ b/ocaml/testsuite/tests/typing-layouts/error_message_attr.ml
@@ -154,3 +154,130 @@ Error: Bad layout annotation:
 |}]
 
 (* Currently it's not possible to attach attributes to Ltyp_poly *)
+
+(* *************************************************************** *)
+(* Tests for [@error_message] applied to [Pexp_constraint].
+   Seperate implementation from when it's applied to layout annotations. *)
+
+(* Needs a string body *)
+let f (x : bool) = (x : int)[@error_message]
+[%%expect{|
+Line 1, characters 28-44:
+1 | let f (x : bool) = (x : int)[@error_message]
+                                ^^^^^^^^^^^^^^^^
+Warning 47 [attribute-payload]: illegal payload for attribute 'error_message'.
+Error_message attribute expects a string argument
+Line 1, characters 20-21:
+1 | let f (x : bool) = (x : int)[@error_message]
+                        ^
+Error: This expression has type bool but an expression was expected of type
+         int
+|}]
+
+(* Can only be applied once *)
+let f (x : bool) = (x : int)[@error_message "A"][@error_message "B"]
+[%%expect{|
+Line 1, characters 48-68:
+1 | let f (x : bool) = (x : int)[@error_message "A"][@error_message "B"]
+                                                    ^^^^^^^^^^^^^^^^^^^^
+Warning 47 [attribute-payload]: illegal payload for attribute 'error_message'.
+More than one error_message attribute present. All of them will be ignored.
+Line 1, characters 20-21:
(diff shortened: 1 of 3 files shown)


ncik-roberts:
I volunteered to review.

ncik-roberts:
Reassigning review to Richard as discussed.

ocaml/testsuite/tests/typing-layouts/error_message_attr.ml
```diff
@@ -116,3 +116,130 @@ Error: Bad layout annotation:
...
+
+(* Simple test case *)
+let f (x : bool) = (x : int)[@error_message "custom message"]
+[%%expect{|
+Line 1, characters 20-21:
+1 | let f (x : bool) = (x : int)[@error_message "custom message"]
+                        ^
+Error: This expression has type bool but an expression was expected of type
+         int
+       because [@error_message]: custom message
```
goldfirere, line 162: I would just have `custom message` here. Gives the user full control.

ocaml/testsuite/tests/typing-layouts/error_message_attr.ml
```diff
@@ -116,3 +116,130 @@ Error: Bad layout annotation:
...
+|}]
+
+(* Doesn't work when the type mismatch happens later. This differ from
```
goldfirere, line 165:
```suggestion
(* Doesn't work when the type mismatch happens later. This differs from
```

ocaml/testsuite/tests/typing-layouts/error_message_attr.ml
```diff
@@ -116,3 +116,130 @@ Error: Bad layout annotation:
...
+
+(* Can be used to enforce layouts but not great *)
+let f (x : string) = (x : (_ : immediate))[@error_message "custom message"]
+[%%expect{|
+Line 1, characters 22-23:
+1 | let f (x : string) = (x : (_ : immediate))[@error_message "custom message"]
+                          ^
+Error: This expression has type string but an expression was expected of type
+         ('a : immediate)
+       because [@error_message]: custom message
```
goldfirere, line 196: As with layouts, it would be cool to put this at the end. But that might be too hard.

ocaml/typing/typecore.ml
```diff
@@ -8367,6 +8372,8 @@ let report_type_expected_explanation expl ppf =
       because "a range-based for iterator stop index in a comprehension"
   | Comprehension_when ->
       because "a when-clause in a comprehension"
+  | Error_message_attr msg ->
+      fprintf ppf "@ because [@error_message]: %s" msg
```
goldfirere, line 8376: If you agree with my suggestion not to include the `because`, then use a separator other than `@ `; some of the others force a newline (I forget which).

goldfirere (approved):

ocaml/testsuite/tests/typing-layouts/error_message_attr.ml
alanechang: I agree. Looked a bit more into this and don't see a simple way of moving the message to the end.

goldfirere:
This is ready to merge. With the unusual merge target, I'm leaving it for you to merge @alanechang -- don't want to cause git trouble.

PR #1955, opened 2023-10-20
Handle empty cases (fixes bug from #1899)

goldfirere:
This fixes a bug I introduced in #1899 where I failed to handle empty cases. The fix is straightforward.

The first commit removes a `[@warning ...]` I added in #1899 but forgot to remove.

The second commit adds the test case and fixes the bug. The test case is a direct copy from @ncik-roberts's #1817, which had to also test empty cases. I confirmed that the test case there indeed tests this behavior.

Review: @riaqn


riaqn (approved):
Looks good - thanks.

PR #1956, opened 2023-10-23
Speedup Shape.local_reduce

poechsel:
`Shape.local_reduce` was taking ~18s (compared to ~8s of type checking) on one file of our codebase. The file in question was large, but not extremely large (~4k lines), but was instancing large functors several times.

After spending a bit of time in pprof I've came up with two optimizations:
- Defining equal / compare functions manually to not call `caml_compare`. The equality function for shape descriptions is still checking physically equality as it is an important property of shapes
- Lazily computing hashes of shapes and using them to prune shapes that are unequal. This change is a little invasive and will require code changes to shape users

Performance detail:
| What | Run 1 | Run 2 | Run 3 |
| :---         |     ---:      |          ---: |          ---: |
| Baseline   | 17.75s     | 17.38s    | 17.6s |
| manual equal    | 3.87s       | 3.83s      | 3.89s |
| manual equal  + hash  | .23s       | .23s      | .23s |

### DIFF
--- a/ocaml/typing/shape.mli
+++ b/ocaml/typing/shape.mli
@@ -65,7 +65,7 @@ module Item : sig
 end
 
 type var = Ident.t
-type t = { uid: Uid.t option; desc: desc }
+type t = { hash: int; uid: Uid.t option; desc: desc }
 and desc =
   | Var of var
   | Abs of var * t
@@ -82,12 +82,12 @@ val print : Format.formatter -> t -> unit
 val for_unnamed_functor_param : var
 val fresh_var : ?name:string -> Uid.t -> var * t
 
-val var : Uid.t -> Ident.t -> t
+val var : ?uid:Uid.t -> Ident.t -> t
 val abs : ?uid:Uid.t -> var -> t -> t
 val app : ?uid:Uid.t -> t -> arg:t -> t
 val str : ?uid:Uid.t -> t Item.Map.t -> t
 val proj : ?uid:Uid.t -> t -> Item.t -> t
-val leaf : Uid.t -> t
+val leaf : ?uid:Uid.t -> unit -> t
 
 val decompose_abs : t -> (var * t) option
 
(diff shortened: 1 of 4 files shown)


Ekdohibs (approved):
Seems good; I'm wondering if using the hash stored in terms could help improve the has function used in`ReduceMemoTable`; but it probably doesn't change anything.

ocaml/typing/shape.mli
```diff
@@ -82,12 +82,12 @@ val print : Format.formatter -> t -> unit
 val for_unnamed_functor_param : var
 val fresh_var : ?name:string -> Uid.t -> var * t
 
-val var : Uid.t -> Ident.t -> t
```
lpw25, line 85: This change doesn't seem necessary and makes the interface worse, making it much more likely to forget to include a uid.

ocaml/typing/shape.mli
```diff
@@ -82,12 +82,12 @@ val print : Format.formatter -> t -> unit
...
+val var : ?uid:Uid.t -> Ident.t -> t
 val abs : ?uid:Uid.t -> var -> t -> t
 val app : ?uid:Uid.t -> t -> arg:t -> t
 val str : ?uid:Uid.t -> t Item.Map.t -> t
 val proj : ?uid:Uid.t -> t -> Item.t -> t
-val leaf : Uid.t -> t
```
lpw25, line 90: Same as above.

ocaml/typing/shape.mli
```diff
@@ -65,7 +65,7 @@ module Item : sig
 end
 
 type var = Ident.t
-type t = { uid: Uid.t option; desc: desc }
```
lpw25, line 68: The new hash field essentially has an invariant on it that is ensured by the smart constructors. That probably means that this type should be made `private` if it can be.

lpw25 (commented):
Just had a quick look at the API changes, and had a couple of comments. I'll review the rest of the change once those are addressed.

PR #1957, opened 2023-10-24
Fix boolean functions tail call position bug

riaqn:
Fixes a bug for boolean operators `&&` and `||`. Currently the `expected_mode.position` is not set correctly for the RHS of the two operators.

See added tests.

Request review from @lpw25

### DIFF
--- a/ocaml/testsuite/tests/typing-local/local.ml
+++ b/ocaml/testsuite/tests/typing-local/local.ml
@@ -2048,12 +2048,28 @@ Error: This local value escapes its region
   Hint: This argument cannot be local, because this is a tail call
 |}]
 
-(* if RHS returns local, the call allocate in caller region *)
+(* boolean operator when at tail of function makes the function local-returning
+   if its RHS is local-returning *)
 let foo () = exclave_ let local_ _x = "hello" in true
 let testboo3 () =  true && (foo ())
 [%%expect{|
 val foo : unit -> local_ bool = <fun>
 val testboo3 : unit -> local_ bool = <fun>
+|}]
+
+(* Test from Nathanaëlle Courant.
+  User can define strange AND, and its arguments should not cross modes. *)
+external strange_and : 'a option -> 'a option -> bool = "%sequand"
+
+let testboo4 () =
+  let local_ x = Some "hello" in
+  strange_and x x
+[%%expect{|
+external strange_and : 'a option -> 'a option -> bool = "%sequand"
+Line 5, characters 14-15:
+5 |   strange_and x x
+                  ^
+Error: This value escapes its region
 |}]
 
 (* mode-crossing using unary + *)
--- a/ocaml/testsuite/tests/typing-local/local.ml
+++ b/ocaml/testsuite/tests/typing-local/local.ml
@@ -2048,6 +2048,14 @@ Error: This local value escapes its region
   Hint: This argument cannot be local, because this is a tail call
 |}]
 
+(* if RHS returns local, the call allocate in caller region *)
+let foo () = exclave_ let local_ _x = "hello" in true
+let testboo3 () =  true && (foo ())
+[%%expect{|
+val foo : unit -> local_ bool = <fun>
+val testboo3 : unit -> local_ bool = <fun>
+|}]
+
 (* mode-crossing using unary + *)
 let promote (local_ x) = +x
 [%%expect{|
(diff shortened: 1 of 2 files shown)


ocaml/typing/typecore.ml
```diff
@@ -504,17 +505,17 @@ let mode_exact mode =
...
       Val_prim {Primitive.prim_name = ("%sequor"|"%sequand")}},
                 Id_prim _, _), 1, Tail ->
-     (* The second argument to (&&) and (||) is in
-        tail position if the call is *)
-      (* vmode is wrong; fine because of mode crossing on boolean *)
-     mode_return vmode
+     (* The second argument to (&&) and (||) is a boolean and crosses modes, so
+        we expect [mode_max]. It is also at the same function tail position if
+        the call is *)
+     mode_max_with_position (RTail (Option.get position_and_mode.region_mode, FTail))
```
lpw25, line 518: I'd prefer we didn't make assumptions about the type of `&&` and `||` only their runtime behaviour. This matches what we do elsewhere for known primitives. Also, if we add more modes later that `bool` doesn't cross then this assumption will be wrong. So I'd rather we tried to give the "correct" mode and let mode crossing handle weakening it based on type.

lpw25 (commented):
I'd like to slightly adjust the mode computation. Then this is good to go.

ocaml/typing/typecore.ml
Ekdohibs: Besides, although it likely doesn't happen anywhere, it could be technically possible to declare `&&` with type `bool -> 'a option -> 'a option`; if might be worth thinking what would happen in this case.

riaqn: That makes sense. 

Just to confirm: there are other similar code.

- the lower and upper bound of `Pexp_loop` are int and checked with `mode_max`. Apparantly your arguments apply to this as well and we should fix this. 

- the body of a loop is checked with `mode_max` - I guess this is fine because the value if any are discarded anyway.

riaqn: @Ekdohibs I think here it's specifically about `%sequor` and `%sequand`; I.e., the primitives. The meaning of those are quite fixed.

Ekdohibs: @riaqn yes, I meant declaring `%sequand` with that type, sorry for the confusion! If I'm not mistaken, the runtime semantics of `%sequand` would be compatible with that type.
Concerning `Pexp_loop`, it is different, because it is a language construct with a special syntax, and not a primitive (declared with `external ( && ) : bool -> bool -> bool = "%sequand"`)

Ekdohibs: The main difference is that primtives are declared with a type, which we trust in the rest of the compiler (this allows us to define `Obj.magic`!), while language constructs have a special syntax, are not user-declarable and so their types are always known in the rest of the compiler.

lpw25: I think there are two separate concerns, and mixing them is causing some confusion in this thread.

1. The relationship between types and `max_mode` may change in the future, which may indeed mean we should adjust the code related to `Pexp_loop`.
2.  We don't actually know what type the `%sequand` primitive is declared with, and as @Ekdohibs points out there are even some non-standard choices that are probably safe to use right now, which means that we should try not to rely on that assumption.

Ekdohibs: Ah, thanks for clearing that out! I understand the part about `Pexp_loop` better now.

riaqn: The mode crossing bit is fixed now.

For the other thing,  Leo and me discussed and decides that they don't need fixing.

ocaml/testsuite/tests/typing-local/local.ml
```diff
@@ -2048,14 +2048,30 @@ Error: This local value escapes its region
...
 let foo () = exclave_ let local_ _x = "hello" in true
 let testboo3 () =  true && (foo ())
 [%%expect{|
 val foo : unit -> local_ bool = <fun>
 val testboo3 : unit -> local_ bool = <fun>
 |}]
 
+(* Test from Nathanaëlle Courant.
+  User can define strange AND, and its arguments should not cross modes. *)
+external strange_and : 'a option -> 'a option -> bool = "%sequand"
```
lpw25, line 2062: I think this should be `bool -> 'a option -> 'a option`

ocaml/testsuite/tests/typing-local/local.ml
```diff
@@ -2048,14 +2048,30 @@ Error: This local value escapes its region
...
+
+let testboo4 () =
+  let local_ x = Some "hello" in
+  strange_and x x
```
lpw25, line 2066: It would be better to only check the mode crossing on the second argument, since that is the one with the weird behaviour.

lpw25 (approved):
The test isn't quite right, but the rest is good.

PR #1960, opened 2023-10-24
Error message: add hint for unboxed types

alanechang:
This PR builds on top of #1864 and adds a hint to the error message whenever `#x` is encountered and `x#` is a valid type in the current environment.

Review: @goldfirere

### DIFF
--- a/ocaml/testsuite/tests/typing-layouts-float64/parsing.ml
+++ b/ocaml/testsuite/tests/typing-layouts-float64/parsing.ml
@@ -255,10 +255,11 @@ Error: The type constructor float# expects 0 argument(s),
 (* Hint for #float *)
 type t = #float;;
 [%%expect {|
-Line 1, characters 10-15:
+Line 1, characters 9-15:
 1 | type t = #float;;
-              ^^^^^
-Error: Did you mean float#?
+             ^^^^^^
+Error: float is neither a polymorphic variant nor a class type.
+       Did you mean the unboxed type float#?
 |}]
 
 (* Hint should not show up in this case *)
--- a/ocaml/testsuite/tests/typing-layouts-float64/parsing.ml
+++ b/ocaml/testsuite/tests/typing-layouts-float64/parsing.ml
@@ -258,7 +258,7 @@ type t = #float;;
 Line 1, characters 10-15:
 1 | type t = #float;;
               ^^^^^
-Error: Unbound class type float
+Error: Did you mean float#?
 |}]
 
 (* Hint should not show up in this case *)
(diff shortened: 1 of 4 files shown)


ocaml/typing/env.ml
```diff
@@ -3889,8 +3889,8 @@ let report_lookup_error _loc env ppf = function
   | Unbound_cltype lid ->
       fprintf ppf "Unbound class type %a" !print_longident lid;
       begin match lid with
-      | Lident "float" ->
-        Misc.did_you_mean ppf (fun () -> ["float#"])
+      | Lident name when List.exists (String.equal (name^"#")) (extract_types None env) ->
```
goldfirere, line 3892: Not a new bug in this patch, but this extra check is a regression from existing behavior. With upstream OCaml, if I say

```
class type floot = object end
class type c = float
```

I'm helpfully told I might have meant `floot`. With our HEAD, I'm asked if I meant `float#`, which actually makes no sense in a class type.

Better would be to remove this code (probably by reverting the earlier patch, which was done in haste) and instead add some logic to the `Ptyp_class` case in Typetexp. That proceeds by first trying to find a type that is a variant. If it's not a variant, then we look up `#t`. If that fails, too, we should see if `t#` is in scope and suggest it. If not, then fail with the existing `Env.lookup_cltype` behavior.

alanechang: As suggested, I reverted the old patch and moved the unboxed type checking logic into the `Ptyp_class` case inside Typetexp. The regression is also added as an extra test.

https://github.com/ocaml-flambda/flambda-backend/blob/fc4d1be345b958cb289b5b02976dc714f15d9065/ocaml/testsuite/tests/typing-layouts-float64/parsing.ml#L254-L274

Compare to before, we lose the `Unbound class type float` line in the error message. Should that information be preserved?

ocaml/testsuite/tests/typing-layouts-float64/parsing.ml
```diff
@@ -258,6 +258,17 @@ type t = #float;;
 Line 1, characters 10-15:
 1 | type t = #float;;
               ^^^^^
+Error: Did you mean float#?
```
goldfirere, line 261: Here's my dream message:

```
Line 1, characters 9-15:
1 | type t = #float;;
             ^^^^^^
Error: float is neither a polymorphic variant nor a class type.
       Did you mean the unboxed type float#?
```

Note that I adjusted the location to include the `#`, because we want to replace `#float` with `float#`, not just `float` with `float#`.

I think this ideal should be easy to cook up.

alanechang: Yep, I'm able to generate this exact message!

goldfirere (approved):

PR #1961, opened 2023-10-24
Break out of loop in `Printtyp.best_type_path` when depth gets big

lukemaurer:
The code in `Printtyp.best_type_path` assumes that if the path is not in the printing env, it must be the case that adding more to the printing depth will eventually succeed in putting it there. However, if something has gone wrong, the path might just be missing from the printing env altogether, in which case the loop continues searching to arbitrary depths. If _two_ things have gone wrong, there might be a self-reference lurking in the environment, meaning the search can go to arbitrary depth without running out of paths to traverse.

We can cut off both scenarios by breaking out of the loop as soon as the printing depth is as big as the best candidate so far, since from that point increasing the depth cannot help.

This change also returns the original path in cases where the normalised one is longer. This may cut down on cases where paths with double underscores are seen in error messages.


goldfirere:
Went to review this. Does it solve the problem? How does it affect error messages? (Maybe this wasn't meant to be reviewed yet?)

lukemaurer:
> Went to review this. Does it solve the problem? How does it affect error messages? (Maybe this wasn't meant to be reviewed yet?)

Oh sorry, should've converted to draft earlier when I realized it broke a test. This wasn't intended to have an impact on error messages, but it turns out it causes ambiguous types to be printed sometimes when the unambiguous path would be longer. (Current plan is to have ambiguity be penalized similarly to double underscores, so that we still have a useful cutoff in most cases.)

ocaml/typing/printtyp.ml
```diff
@@ -845,7 +845,14 @@ let is_unambiguous path env =
       List.for_all (fun p -> lid_of_path p = id) rem &&
       Path.same p (fst (Env.find_type_by_name id env))
 
-let rec get_best_path r =
+let ambiguity_penalty path env =
+  if is_unambiguous path env then 0 else 10
+
+let path_size path env =
+  let l, s = path_size path in
```
lpw25, line 852: I think it would be worth moving `path_size` and `penalty` down to join these two functions.

ocaml/typing/printtyp.ml
```diff
@@ -845,7 +845,14 @@ let is_unambiguous path env =
       List.for_all (fun p -> lid_of_path p = id) rem &&
       Path.same p (fst (Env.find_type_by_name id env))
 
-let rec get_best_path r =
+let ambiguity_penalty path env =
+  if is_unambiguous path env then 0 else 10
```
lpw25, line 849: It would be good to define a single value to hold the `10` and use it here and in the other penalty calculation.

lpw25 (approved):
A couple of small suggestions, but this is good to go.

ocaml/typing/printtyp.ml
lukemaurer: Done.

ocaml/typing/printtyp.ml
lukemaurer: Done (called it `penalty_size`).

lpw25 (approved):

ocaml/typing/printtyp.ml
```diff
@@ -860,7 +842,34 @@ let is_unambiguous path env =
...
+let path_size path env =
+  let rec size = function
+      Pident id ->
+        name_penalty (Ident.name id), -Ident.scope id
+    | Pdot (p, _) | Pextra_ty (p, Pcstr_ty _) ->
+        let (l, b) = size p in (1+l, b)
+    | Papply (p1, p2) ->
+        let (l, b) = size p1 in
+        (l + fst (size p2), b)
+    | Pextra_ty (p, _) -> path_size p
```
lpw25, line 867: Bug from the rebase: this should be `size`.

lukemaurer: Whoops, thanks. That's what I get for rebasing without building.

PR #1962, opened 2023-10-24
Iterate over all persistent modules in `Env.iter_env`

lukemaurer:
Currently, `Env.iter_env` doesn't involve `Persistent_env` unless something in the `Env.t` is bound to a persistent module. This risks missing possibilities for path shortening, since most persistent modules are only stored in `Persistent_env`.


lpw25 (approved):
LGTM

PR #1964, opened 2023-10-25
Modal kinds (v0)

goldfirere:
This is a rewrite of jkind.ml that implements the start of modal kinds. It is meant to be a purely internal refactor of jkind, but the old-style mode-cross check sadly has become effectful and needs to be protected with backtracking. This can be avoided, but it would make the code in jkind uglier (by checking externality first and then aborting before looking at layouts).

There are two commits. The first is a tiny renaming. The second is the payload.

For review, I recommend just reading the new jkind.ml. Trying to interpret the diff will leave you sad.

I have tried to keep the layout-histories portions of jkind.ml essentially untouched, to make merging with @alanechang's work easier.

Next steps after this is merged:
- Incorporate other mode axes into the scheme
- Use the modal kinds to do the actual mode-crossing check, instead of looking for `immediate64`.
- Remove the compatibility layer at the bottom of jkind.ml
- Expose modal kinds to users via the syntax we have agreed on.

Review suggestion: @ccasin


goldfirere:
This PR targets #1952, as that sorts out a little inference puzzle that is incompatible with modal kinds, as @ccasin and I have discussed. @alanechang is queued up to review #1952.

ocaml/typing/ctype.ml
```diff
@@ -2130,7 +2130,14 @@ let is_immediate64 env ty =
     Btype.backtrack snap;
     result
   else
-    perform_check ()
+    (* CR layouts v2.8: Remove the backtracking once mode crossing is
+       implemented correctly; it's needed for now because checking whether
+       a jkind is immediate (rightly) sets the sort to be Value. It worked
+       previous to this patch because the subjkind check failed earlier. *)
```
ccasin, line 2136: (I haven't reviewed in detail yet - just a question about this first.)

Can you say more about what the "correct" implementation of mode crossing check will look like?

I think a thing that was true before this PR is that a failing jkind check never changed jkind state.  It could admittedly change some other state, necessitating the backtracking in the previous case.  But the fact about jkinds does still seem like a nice property to have for reasoning about the code.  You say here that it's "right" for a failing check to set the sort variable, but that's not obvious to me - I could imagine it either way, and the other way seems to give that nice reasoning principle.  Probably I'm missing something though?

You mention that it would make the code in `Jkind` uglier to avoid this.  If I understand correctly, it's just a matter of making `Sub_result.combine` take an list of closures and checking them in the given order?  Probably we'll want to generalize `combine` to take a list anyway as jkinds grow, and having that be a list of closures doesn't seem like such a big price to pay.

goldfirere: The "correct" implementation will just be a check against externality, not involving the layout at all.

The reason I say it's appropriate for the existing check to set the sort is that the existing check looks to see whether a kind is a subtype of immediate. Following the general design here, every check tries to mutate variables to make the check succeed. The only way for a kind to be a subtype of immediate is for the sort to be `value`. So when we're checking to see whether a kind is a subtype of immediate, we'll set the sort to be `value` if it's a sort variable. The higher-level point is that we shouldn't even be making that check if we don't want to set the sort variable.

As for the failing check not mutating things: that's not true elsewhere. If I try to unify `'a * int` with `char * bool`, I'll link `'a` with `char` before failing. And I think that's just fine. The situation here is very similar.

ccasin: Your comment about "the general design here" isn't obvious to me.  Until now, failing checks didn't mutate sort variables.  So, rather than saying every layout check tries to mutate sort variables to make the check succeed, you could equally well say every check only mutates variables if it will make the check succeed.  And the latter seems nice, even if it's not true for other similar checks in the typechecker.  So the word "rightly" surprised me, because I think either way would work fine.

But, anyway, I don't think this matters so much, particularly since the current implementation is temporary.  I'll try to review this in earnest by early next week.

ocaml/typing/jkind.ml
```diff
@@ -261,6 +274,238 @@ end
...
+      | Sort s -> fprintf ppf "Sort %a" Sort.Debug_printers.t s
+  end
+end
+
+(* Whether or not a type is external to the garbage collector *)
+module Externality = struct
+  type t =
+    | External (* not managed by the garbage collector *)
+    | External64 (* not managed by the garbage collector on 64-bit systems *)
+    | Internal (* managed by the garbage collector *)
```
Ekdohibs, line 358: So, the way I understand this, we use externality `External` for  for unboxed floats; shouldn't we instead have a different externality for types that *must not* be managed by the garbage collector at opposed to types than can or cannot be managed by it?

ocaml/typing/ctype.ml
lpw25: FWIW I think the right position on mutations in such checks is that *consistent* state changes should be preserved. `'a * int` mutating to have `'a` linked with `char` is an example of such a change. An example of an inconsistent state change is when you unify `string t` with `float t`, the first thing that does is mutate the `string t` node to be a link to the `float t` node (which is needed to handle recursive types), but once the inner unification has failed it is important to undo that change and leave things in a consistent state.

The reason to leave consistent state changes in place rather than undo them is that it improves error messages. When you try and make two things equal you want the algorithm to make them as equal as they can be, that way when you print the two things, the only differences between them are the ones that can't be rectified.

(I actually think it would be good to change type unification to do more state changes, so that e.g. unifying `'a * int * 'b` with `string * string * string` would set both `'a` and `'b` to `string`).

ccasin: (FWIW, I am pretty convinced by the argument about better errors that mutating here is indeed preferable, as Richard's original comment suggested)

goldfirere: @ccasin is correct about the alternative interpretation of the old implementation -- that we only mutated sort variables if the check would eventually succeed. It was not how I saw things (as my comments suggest!), but it's absolutely true. In any case, I think we agree that this particular choice matters little and that it will be cleaned up soon.

ocaml/typing/jkind.ml
goldfirere: Missed this somehow; sorry.

This story will likely evolve. I think we'll probably never consult the externality of a non-`value` type. So maybe this ends up somehow being folded in with the `Value` sort. I have not experimented with this. First, I'd like to use externality to affect compilation (that is, check the externality of a value to decide whether to compile it like an `immediate`) and then revisit the possibility of reshuffling this.

In more direct answer to your question: if the sort of a type is not `value`, then it must not be managed by the garbage collector. All such things should also be `External`, but arranging for this in general is hard and unnecessary (because I think we consult `External` only for `value`s).

ocaml/typing/jkind.ml
```diff
@@ -261,6 +274,238 @@ end
...
+    | Const of Const.t
+    | Var of Sort.var (* all modes will be [max] *)
+
+  let format ppf =
+    let open Format in
+    function
+    | Const c -> fprintf ppf "%s" (Const.to_string c)
+    | Var v -> fprintf ppf "%s" (Sort.var_name v)
+
+  (* considers sort variables < Any, but otherwise just checks for equality.
```
ccasin, line 440: I think this comment is slightly out of date wrt externality, which makes it no longer just an equality check.

ocaml/typing/jkind.ml
```diff
@@ -261,6 +274,238 @@ end
...
+     Pre-condition: no filled-in sort variables. *)
+  let sub d1 d2 : Sub_result.t =
+    match d1, d2 with
+    | Const c1, Const c2 -> Const.sub c1 c2
+    | Var _, Const { layout = Any; externality = Internal } -> Sub
+    | Var v1, Var v2 -> if v1 == v2 then Equal else Not_sub
+    | Const _, Var _ | Var _, Const _ -> Not_sub
+end
+
+module Jkind_ = struct
```
ccasin, line 451: Why `Jkind_` rather than `Jkind`?

ocaml/typing/jkind.ml
```diff
@@ -404,65 +638,29 @@ let fresh_jkind jkind ~why = { jkind; history = Creation why }
...
 
-let string_of_const : const -> _ = function
-  | Any -> "any"
-  | Value -> "value"
-  | Void -> "void"
-  | Immediate64 -> "immediate64"
-  | Immediate -> "immediate"
-  | Float64 -> "float64"
+let immediate64 ~why =
+  fresh_jkind Jkind_.immediate64 ~why:(Immediate64_creation why)
```
ccasin, line 659: It's not new code here, but should we be memoizing these?  I don't see any state.

ocaml/typing/jkind.ml
```diff
@@ -1009,18 +1164,11 @@ end
...
-    | (Equal_mutated_first | Equal_mutated_second) when not allow_mutation ->
-      Misc.fatal_errorf "Jkind.equal: Performed unexpected mutation"
-    | Unequal -> false
-    | Equal_no_mutation | Equal_mutated_first | Equal_mutated_second -> true)
-  | (Any | Immediate64 | Immediate | Sort _), _ -> false
+let equate_or_equal ~allow_mutation
+    { jkind = { layout = lay1; externality = ext1 }; history = _ }
+    { jkind = { layout = lay2; externality = ext2 }; history = _ } =
+  Layout.equate_or_equal ~allow_mutation lay1 lay2
+  && Externality.equal ext1 ext2
```
ccasin, line 1171: Should there be `Jkind_.equate_or_equal` that this just calls?

ccasin (approved):
Sorry for taking a couple weeks to get to this.  Looks fine - starting to come together.  Approving because I didn't spot any bugs/problems, but I do have some questions - see below.

ocaml/typing/jkind.ml
ccasin: (for common creation reasons, I mean.  Maybe not worth doing?)

ocaml/typing/jkind.ml
goldfirere: > Should it be a static error to construct a kind with a layout other than value that is not always external?

I originally thought it should be, but this is hard to arrange, because we sometimes have to decide externality before we know what sort a type will have. If we had externality variables, then all would be well, but let's not. So I think externality just is meaningless for non-`value`s.

@Ekdohibs is right, also, once we get to product layouts. For a product of a `value` and some other layout, the externality field will still be relevant.

ocaml/typing/jkind.ml
```diff
@@ -261,6 +274,238 @@ end
...
+    open Format
+
+    let t ppf = function
+      | Any -> fprintf ppf "Any"
+      | Sort s -> fprintf ppf "Sort %a" Sort.Debug_printers.t s
+  end
+end
+
+(* Whether or not a type is external to the garbage collector *)
+module Externality = struct
```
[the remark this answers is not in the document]
goldfirere: I've added that CR; you're right that it should be with the other modes. But between the mode solver work and the fact that I wanted to keep this PR self-contained, it was easier to put it here for now.

A mode in a modal kind is an upper bound. That is, suppose expression `e` has a type whose kind includes the mode `m`. Then if `e` is at a mode greater than `m` along `m`'s axis, we can just say `e` is at mode `m`. Since the current `Externality.t` type is used only for mode crossing -- and not for the mode of an expression -- constructors like the ones you're describing make sense. But the goal will be to use the same structure for existing modes, in which case the current constructor names will line up nicely.

Does that help?

ocaml/typing/jkind.ml
goldfirere: I've clarified the comment. Better?

ocaml/typing/jkind.ml
goldfirere: Because I didn't want `Jkind.t` to mean one thing inside this file and a different thing everywhere else. I thought of `Jkind_internal` or something like that, but I didn't like that much more either. Open to suggestions.

ocaml/typing/jkind.ml
goldfirere: It's a good question. I've added a CR. One challenge is that we're still figuring out what fields we need on the reasons. Alan has added a bunch. So I didn't want to figure this out until that part is done. But when it all settled, we might get a nice little speed boost by doing some memoization.

ocaml/typing/jkind.ml
goldfirere: Yes. Good idea.

goldfirere:
Will rebase after #1952 lands.

goldfirere:
Will rebase after #2003 lands.

ocaml/typing/jkind.ml
ccasin: Is the long term plan to keep this module, or to one day combine the two types (so the history is just another field of what is currently called `Jkind_.t`?  (And we're not doing that today to avoid annoying merge conflicts with the errors work).

If so, I think this doesn't matter so much.  And I think that would be a fine design (maybe better - fewer boxes to allocate and pointers to follow).

If not, I prefer a name like `Jkind_internal` or just `Internal`.  Another option would be to name this module `Jkind_desc` or the type `jkind_desc`, by analogy with other parts of the type system.

goldfirere: My thought was to keep this separation indefinitely. It allows us to talk about the meaty parts of a jkind, without regard to the history. Though I agree it's unfortunately a little allocate-y. If you think it's worthwhile, I can make a ticket about possibly eliminiating the internal type later.

I'm quite happy with `Jkind_desc` -- that's definitely better than what I've done. I will make that change.

ocaml/typing/jkind.ml
goldfirere: I think there's nothing to action here now. Resolving.

goldfirere:
Will rebase after #1907 lands.

PR #1965, opened 2023-10-26
Fix error when mutably unboxing a block containing another mutably unboxed block

Ekdohibs:
This was previously failing on:

```ocaml
[@@@ocaml.flambda_o3]

type t = { mutable v : float }

type t2 = { mutable y : float ; z : t }

module C = struct
  type 'a t =
    | T : { t : 'c ; get : 'c -> 'a } -> 'a t

  let[@inline] get (T { t; get }) = (get [@inlined hint]) t
end

let make_c =
  fun [@inline] t -> C.T { t; get = fun [@inline] t -> t.v }

let get t = C.get (make_c t)

let () =
  let t = { y = 1.; z = { v = 1. } } in
  for _ = 1 to 100 do
    let (_ : _) =
      get t.z
    in
    ()
  done
```

### DIFF
--- a/middle_end/flambda2/simplify/flow/mutable_unboxing.ml
+++ b/middle_end/flambda2/simplify/flow/mutable_unboxing.ml
@@ -264,7 +264,7 @@ let continuations_using_blocks ~non_escaping_blocks ~dom
       in
       if (not (Variable.Set.is_empty used)) && Flambda_features.dump_flow ()
       then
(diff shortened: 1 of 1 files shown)


Gbury (approved):

Gbury:
Note: the macOS CI failure seems unrelated to the current PR and instead looks to be caused by `afl`

PR #1967, opened 2023-10-27
Minor fixes for PR#1966

xclerc:


mshinwell:
"Missed change in selectgen.ml" and "SIMD files" are approved

mshinwell:
Necessary changesets included in PR1949

PR #1968, opened 2023-10-27
More precise layout for array patterns

alanechang:
In preparation for arrays of unboxed types, this PR addresses an existing CR and picks a more accurate `Lambda.layout` when matching array patterns.

Review suggestion: @Ekdohibs

### DIFF
--- a/ocaml/lambda/matching.ml
+++ b/ocaml/lambda/matching.ml
@@ -2197,18 +2197,22 @@ let get_expr_args_array ~scopes kind head (arg, _mut, _sort, _layout) rem =
     if pos >= len then
       rem
     else
-      (* CR ncourant: could do better than layout_field using kind *)
+      (* TODO: The resulting float should be allocated to at the mode of the
+         array pattern, once that's available *)
+      let prim = Parrayrefu Lambda.(array_ref_kind alloc_heap kind) in
+      let result_layout =
+        (* only care about layout on native *)
+        if !Clflags.native_code then primitive_result_layout prim else layout_field
+      in
       ( Lprim
-          (* TODO: The resulting float should be allocated to at the mode of the
-             array pattern, once that's available *)
-          (Parrayrefu Lambda.(array_ref_kind alloc_heap kind),
+          (prim,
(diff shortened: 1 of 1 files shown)


ocaml/lambda/matching.ml
```diff
@@ -2197,18 +2197,22 @@ let get_expr_args_array ~scopes kind head (arg, _mut, _sort, _layout) rem =
     if pos >= len then
       rem
     else
-      (* CR ncourant: could do better than layout_field using kind *)
+      (* TODO: The resulting float should be allocated to at the mode of the
+         array pattern, once that's available *)
+      let prim = Parrayrefu Lambda.(array_ref_kind alloc_heap kind) in
+      let result_layout =
+        (* only care about layout on native *)
+        if !Clflags.native_code then primitive_result_layout prim else layout_field
```
Ekdohibs, line 2205: Even if we do not care about it, it is probably simpler to simply use the native case in all cases (especially since this is seldom called).
We could expose a function in `lambda.ml` getting a layout from an `array_ref_kind`, but the code is fine like that for now.

Ekdohibs (approved):
Looks good, although I'd say you should compute the layout in all cases, except if you have actual examples where it is important not to compute it in bytecode mode.

ocaml/lambda/matching.ml
alanechang: Main reason for the if check is the assertion in `primitive_result_layout`:

https://github.com/ocaml-flambda/flambda-backend/blob/2001e134945a0e542f146d9f43a2a7112f57b77e/ocaml/lambda/lambda.ml#L1534

Is this line something that can be removed? Then we can call `primitive_result_layout` directly on both bytecode and native.

Ekdohibs: Oh, good catch! I think it's better to keep that line (for the case of C calls), so we might want to export a function of type `array_ref_kind -> layout` in lambda and use it in both places.

alanechang: Sounds good! I made the change suggested.

PR #1969, opened 2023-10-27
Abstract over instruction prefix in binary emitter

TheNumbat:
This refactors `emit_mod_rm_reg` to work with arbitrary instruction prefixes. It is preparation for adding two VEX-encoded instructions in #1970.

There should be no changes to functionality.


backend/x86_binary_emitter.ml
```diff
@@ -420,36 +420,38 @@ let buf_sym b sym offset =
       record_reloc b (Buffer.length b.buf) (Relocation.Kind.DIR32 (lbl, offset));
       buf_int32L b 0L
 
-let emit_mod_rm_reg b rex opcodes rm reg =
+let emit_prefix_modrm b opcodes rm reg ~prefix =
   match rm with
   | Reg32 rm ->
       let rm = rd_of_reg64 rm in
-      emit_rex b (rex lor rexr_reg reg lor rexb_rm rm);
+      prefix b ~rex:0 ~rexr:(rexr_reg reg) ~rexb:(rexb_rm rm) ~rexx:0;
```
xclerc, line 427: It is not immediately obvious to me why
"replacing" `rex` with `0` in the `lor` expresion
(not below) is correct. Could you add a
comment?

TheNumbat:
The binary emitter is tested via the `tests/simd` tests, but they're definitely not comprehensive outside of the SIMD instructions. It does exercise this particular code, though.

xclerc (approved):

PR #1970, opened 2023-10-27
BMI2 Intrinsics

TheNumbat:
I needed these instructions when writing a SIMD sexp parser, so here are their intrinsics.

Note that PEXT/PDEP are VEX encoded. 

Basic tests in `tests/simd/scalar_ops.ml`; also tested via usage in the parser.


tests/simd/scalar_ops.ml
```diff
@@ -29,3 +29,18 @@ module Int64x2 = struct
...
+  type t = int64
+
+  external bit_deposit : t -> t -> t = "caml_vec128_unreachable" "caml_bmi2_int64_deposit_bits"
+      [@@noalloc] [@@unboxed] [@@builtin]
+  external bit_extract : t -> t -> t = "caml_vec128_unreachable" "caml_bmi2_int64_extract_bits"
+      [@@noalloc] [@@unboxed] [@@builtin]
+
+  let () =
+    eq (bit_deposit 3L 4L)     (bit_extract 3L 4L)     0x4L 0x0L;
+    eq (bit_deposit 235L 522L) (bit_extract 235L 522L) 0xAL 0x3L;
```
xclerc, line 44: (The way the test is expressed feels a
bit odd: on each line the tests of
`bit_deposit` and `bit_extract` are
actually independent.)

backend/x86_binary_emitter.ml
```diff
@@ -1151,6 +1151,40 @@ let emit_movmskpd b dst src =
       emit_mod_rm_reg b 0 [ 0x0f; 0x50 ] rm (rd_of_reg64 reg)
   | _ -> assert false
 
+let emit_vex3 buf ~rexr ~rexx ~rexb ~vexm ~vexw ~vexv ~vexl ~vexp =
```
xclerc, line 1154: (Same remark as in the target branch:
if we have no tests for this code, it seems
very easily for both the developer and
the reviewer to let a typo slip.)

xclerc (commented):
I have already said so on a past pull
request, but I am still afraid the way
we handle instrinsics/primitives does
not scale.

That will not happen on this pull
request, but we should try and find
a better way to encode the constraints
of the instructions.

tests/simd/scalar_ops.ml
TheNumbat: True - split them up

backend/amd64/simd_reload.ml
```diff
@@ -38,6 +38,13 @@ let reload_operation makereg op arg res =
     let arg0 = if stackp arg.(0) then makereg arg.(0) else arg.(0) in
     let res0 = if stackp res.(0) then makereg res.(0) else res.(0) in
     [| arg0 |], [| res0 |]
+  | R_RM_to_R ->
+    (* First argument and result must be in registers. *)
+    let arg0 = if stackp arg.(0) then makereg arg.(0) else arg.(0) in
+    let res0 = if stackp res.(0) then makereg res.(0) else res.(0) in
+    let arg = Array.copy arg in
+    Array.set arg 0 arg0;
+    arg, [| res0 |]
```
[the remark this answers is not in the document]
TheNumbat: Sure, updated all occurances.

backend/x86_binary_emitter.ml
TheNumbat: This is tested by compiling scalar_ops.ml with the binary emitter

xclerc (approved):

PR #1971, opened 2023-10-27
Add Jkind.sort to Tpat_array

alanechang:
This PR follows after #1968 and adds a `Jkind.sort` to `Tpat_array` with the goal of eventually supporting pattern matching on arrays of unboxed types.

For review: Most of the file changes are just adding the extra field in pattern matches. The interesting changes are:
- the places using the sort in `iter_pattern_full ` of `typedtree.ml`, `get_expr_args_array` in `matching.ml`
- the place where the sort is set in `type_pat_array` of `typecore.ml`
- some changes in `parmatch.ml` are a bit tricky - mostly ignored the sort var under the assumption that the expression is already type checked

### DIFF
--- a/ocaml/typing/printtyped.ml
+++ b/ocaml/typing/printtyped.ml
@@ -290,8 +290,9 @@ and pattern : type k . _ -> _ -> k general_pattern -> unit = fun i ppf x ->
   | Tpat_record (l, _c) ->
       line i ppf "Tpat_record\n";
       list i longident_x_pattern ppf l;
-  | Tpat_array (am, _arg_sort, l) ->
+  | Tpat_array (am, arg_sort, l) ->
       line i ppf "Tpat_array %a\n" fmt_mutable_flag am;
+      line i ppf "%a\n" Jkind.Sort.format arg_sort;
       list i pattern ppf l;
   | Tpat_lazy p ->
       line i ppf "Tpat_lazy\n";
--- a/ocaml/typing/printtyped.ml
+++ b/ocaml/typing/printtyped.ml
@@ -290,7 +290,7 @@ and pattern : type k . _ -> _ -> k general_pattern -> unit = fun i ppf x ->
   | Tpat_record (l, _c) ->
       line i ppf "Tpat_record\n";
       list i longident_x_pattern ppf l;
-  | Tpat_array (am, l) ->
+  | Tpat_array (am, _arg_sort, l) ->
       line i ppf "Tpat_array %a\n" fmt_mutable_flag am;
       list i pattern ppf l;
   | Tpat_lazy p ->
--- a/ocaml/typing/typecore.ml
+++ b/ocaml/typing/typecore.ml
@@ -3065,8 +3065,9 @@ let rec check_counter_example_pat
       in
       map_fold_cont type_label_pat fields
         (fun fields -> mkp k (Tpat_record (fields, closed)))
-  | Tpat_array (mut, _, tpl) ->
+  | Tpat_array (mut, original_arg_sort, tpl) ->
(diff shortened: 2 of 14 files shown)


ocaml/typing/typedtree.ml
```diff
@@ -909,12 +909,13 @@ let iter_pattern_full ~both_sides_of_or f sort pat =
...
             (loop f) (Jkind.sort_of_jkind lbl.lbl_jkind) pat)
             lbl_pat_list
+      | Tpat_array (_, arg_sort, patl) ->
+        List.iter (loop f arg_sort) patl
       (* Cases where the inner things must be value: *)
       | Tpat_variant (_, pat, _) -> Option.iter (loop f Jkind.Sort.value) pat
       | Tpat_tuple patl -> List.iter (loop f Jkind.Sort.value) patl
         (* CR layouts v5: tuple case to change when we allow non-values in
            tuples *)
-      | Tpat_array (_, patl) -> List.iter (loop f Jkind.Sort.value) patl
```
goldfirere, line 917: Why move this? That will cause more challenge when merging later.

ocaml/typing/typecore.ml
```diff
@@ -1464,11 +1464,13 @@ let solve_Ppat_array ~refine loc env mutability expected_ty =
     | Mutable -> Predef.type_array
   in
   (* CR layouts v4: in the future we'll have arrays of other jkinds *)
```
goldfirere, line 1466: Update this CR to explain that the code below (which is a bit silly: the sort will always be `Sort.value`) will get generalized once we generalize `array`.

ocaml/typing/printtyped.ml
```diff
@@ -305,7 +305,7 @@ and pattern : type k . _ -> _ -> k general_pattern -> unit = fun i ppf x ->
   | Tpat_record (l, _c) ->
       line i ppf "Tpat_record\n";
       list i longident_x_pattern ppf l;
-  | Tpat_array (am, l) ->
+  | Tpat_array (am, _arg_sort, l) ->
```
goldfirere, line 308: Print the new field. This is used for debugging, so we don't want to hide this value.

goldfirere (approved):

ocaml/typing/typedtree.ml
alanechang: Moved it to try to keep the `(* Cases where the inner things must be value: *)` and `(* Cases where we compute the sort of the inner thing from the pattern *)` comments true.

I agree it may not be worth it and can introduce merge conflicts. Reverted the change in a new commit

alanechang:
Rebased this PR over the ocaml 5 minus changes resulting in one extra place that had to be changed. I singled out that change in the last commit. Would be helpful to have that reviewed @goldfirere

ocaml/typing/typecore.ml
```diff
@@ -3065,10 +3065,10 @@ let rec check_counter_example_pat
       in
       map_fold_cont type_label_pat fields
         (fun fields -> mkp k (Tpat_record (fields, closed)))
-  | Tpat_array (mut, tpl) ->
-      let ty_elt = solve_Ppat_array ~refine loc env mut expected_ty in
+  | Tpat_array (mut, _, tpl) ->
+      let ty_elt, arg_sort = solve_Ppat_array ~refine loc env mut expected_ty in
```
goldfirere, line 3069: I don't know this function well. I think it would be safest to assert that the returned sort is the same as the original sort. That way, in case this is wrong, we'll know right away.

Ekdohibs (approved):
For changes in chamelon.

ocaml/typing/typecore.ml
alanechang: Sounds good! I added an assert with call to `Jkind.Sort.equate`.

PR #1972, opened 2023-10-30
CLMUL Intrinsic

TheNumbat:


gretay-js (approved):
Not for now, but we should consider auto-generating [tests/simd/dune](https://github.com/ocaml-flambda/flambda-backend/pull/1972/files#diff-73526202af4eb1792f6b4945a265a248eb31d0b4e9e9528fac6e50f2ba5501e7), as it gets bigger.

PR #1974, opened 2023-10-31
Fix more warnings (runtime)

xclerc:
(This fixes all but two warnings,
even when building with clang.)

@mshinwell


xclerc:
0cd17d5d577f39edf4802479ad722d8c480b1154 _fixes_ the two remaining
warnings by silencing them; I think
it is OK because the offending
warning is disabled only in `prims.c`
(which is generated).

mshinwell (approved):

PR #1975, opened 2023-11-01
Add OCaml 4 feature macros

ncik-roberts:
Add feature macros defining which OCaml 4 features our runtime has. We should add to this freely, I'm just adding ones that come to mind.

This is needed because C code often uses conditional compilation to decide which bindings in the runtime to access. Usually that code just inspects `OCAML_VERSION` in `caml/version.h`. We've bumped that version past OCaml 5, but don't yet offer all OCaml 5-related bindings in the runtime, so we need to guard access to the new bindings based on additional preprocessor checks. The new macros in this PR are meant to be used in such checks.


ncik-roberts:
Holding off for the moment

mshinwell (approved):

PR #1976, opened 2023-11-01
Prevent use after free bug after a thread has stopped

lpw25:
#1869 introduced a use-after-free bug. When a thread is stopped or unregisters it frees it's associated `caml_thread_struct`. If this is the one currently stored in `curr_thread` then the next call to `caml_thread_switch_runtime_state` will write to the freed memory.

The PR fixes the issue by setting `curr_thread` to `NULL` when the referenced value is going to be freed, and guarding the writes in `caml_thread_save_runtime_state` by a check that `curr_thread` is not `NULL`.

This adjusts the invariant on `curr_thread`. When we hold the run-time lock, `curr_thread` can only be `NULL` if we have not yet initialized the threading system. However, when we do not hold the run-time lock we may see `curr_thread` as `NULL` if the most recently executing thread has been stopped. I've checked that all the uses of `curr_thread` outside of `caml_thread_save_runtime_state` are when we already hold the runtime lock and so don't need to check for `NULL`.

The one exception to this is `caml_thread_reinitialize` which can run in any state since it will run whenever `fork` is caleld. However, that code already seems completely broken in the case where the `fork` is called whilst we don't hold the runtime lock, so that needs more work to fix it properly anyway. I've left it alone for now as I don't think #1869 made things much worse there.

### DIFF
--- a/ocaml/otherlibs/systhreads/st_stubs.c
+++ b/ocaml/otherlibs/systhreads/st_stubs.c
@@ -247,27 +247,29 @@ static void thread_save_runtime_state(void)
 {
   if (Caml_state->in_minor_collection)
     caml_fatal_error("Thread switch from inside minor GC");
+  if (curr_thread) {
 #ifdef NATIVE_CODE
-  curr_thread->top_of_stack = Caml_state->top_of_stack;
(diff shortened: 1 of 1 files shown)


stedolan (approved):
This looks correct to me.

PR #1981, opened 2023-11-02
Fix error message regression: use manifest jkind less

alanechang:
With this example, we get a confusing error message about layouts:

```
utop # type a = unit ;;
--
type a = unit
utop # type b = a = { a : int };;
Line 1, characters 0-24:
Error: Type b has layout value, which is not a sublayout of immediate.
```

The original error message is more helpful:

```
2 | type b = a = { a : int }
    ^^^^^^^^^^^^^^^^^^^^^^^^
Error: This variant or record definition does not match that of type a
       Their kinds differ. 
```

This PR seeks to have the old error message printed rather than the new one.

To achieve this, it restricts the jkind inference using the type manifest in `transl_declaration` to only when it's a `Ttype_abstract`. There's a call to `check_coherence` later than should help ensure the type jkind does not differ from the manifest jkind.

### DIFF
--- a/ocaml/testsuite/tests/typing-misc/records.ml
+++ b/ocaml/testsuite/tests/typing-misc/records.ml
@@ -300,3 +300,14 @@ Line 6, characters 13-23:
 Error: Unbound module Coq__10
 Hint: Did you mean Coq__11?
 |}]
+
+type a = unit
+type b = a = { a : int }
+[%%expect{|
+type a = unit
+Line 2, characters 0-24:
+2 | type b = a = { a : int }
+    ^^^^^^^^^^^^^^^^^^^^^^^^
+Error: This variant or record definition does not match that of type a
+       The original is abstract, but this is a record.
+|}]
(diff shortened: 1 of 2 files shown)


goldfirere:
I've just pushed a further simplification, as suggested by the CR I wrote near here some time ago. It also fixes a small bug about which type was entered into the `final_env`.

Could you review my commit please? I'll review yours.

goldfirere (approved):
Commits from @alanechang

alanechang (commented):
Changes in the latest commit look good.

This seems to fix:

```
type 'p pair : any = 'a * 'a constraint 'p = < m : 'a >
let foo : < m0 : 'a. < m : 'a> pair > -> < m : 'a. <m : 'a> pair > = fun x -> x
```

Error message is now:

```
2 | let foo : < m0 : 'a. < m : 'a> pair > -> < m : 'a. <m : 'a> pair > = fun x -> x
                                                                                  ^
Error: This expression has type < m0 : 'a. < m : 'a > pair >
       but an expression was expected of type < m : 'a. < m : 'a > pair >
       The second object type has no method m0
```

Should I add this example in as a test case?

goldfirere:
Not necessary: it's redundant with existing tests. It was just simplified for the purposes of the discussion earlier. Thanks for the review. Will merge now.

PR #1982, opened 2023-11-03
Make submode_vv_faster

poechsel:
A large source file mostly composed of  sequential operations over integers was taking ~300s to typecheck:
```
004.506s -pp
  000.254s parsing
    000.254s other
  289.355s typing
    289.014s infer
    000.003s check_sig
    000.002s save_cmi
    000.135s save_cmt
    000.201s other
  001.445s transl
  017.338s generate
    010.840s flambda2
      004.204s lambda_to_flambda
      006.636s flambda_to_cmm
    003.874s compile_phrases
      000.744s selection
      000.035s checkmach
        000.034s check zero_alloc
        000.001s other
      000.052s comballoc
      000.635s cse
      002.199s regalloc
        002.198s default
          000.257s liveness
          000.070s deadcode
          000.678s spill
          000.192s split
          000.816s regalloc
          000.181s mach to linear
            000.142s cfgize
              000.005s register_preds
              000.063s optimizations
              000.074s other
            000.038s cfg_to_linear
            000.001s other
          000.004s other
        000.001s other
      000.202s emit_fundecl
        000.202s emit
        000.001s other
      000.006s other
    000.394s write_asm
    000.651s compact
    001.106s assemble
    000.473s other
  000.004s other
000.023s other
```
After inspection with perf it seemed that ~80% of the time was spend in the call to `List.memq` within `Mode.submode_vv`.

Reordering the layout of fields in `'a var` to make `mvid` be the first element confirmed these numbers as it was enough to make typechecking go down to ~200s.
Going further and replacing the `list` by a `map` reduced typechecking time even further to a mere `7.4s`.

This PR makes the assumption that equality of vars can be reduced to equality of their unique identifier. If this is not the case then I think we can still use a Map but it will get trickier as vars are parametrized.

### DIFF
--- a/ocaml/typing/mode.ml
+++ b/ocaml/typing/mode.ml
@@ -13,19 +13,25 @@
 (*                                                                        *)
 (**************************************************************************)
 
+module VarMap = Map.Make (struct
+  type t = int
+
+  let compare = Int.compare
+end)
+
 type 'a var =
-  { mutable upper : 'a;
+  { mvid : int;
+    mutable upper : 'a;
     mutable lower : 'a;
-    mutable vlower : 'a var list;
-    mutable mark : bool;
-    mvid : int
+    mutable vlower : 'a var VarMap.t;
+    mutable mark : bool
   }
 
 type changes =
   | Cnil : changes
   | Cupper : 'a var * 'a * changes -> changes
   | Clower : 'a var * 'a * changes -> changes
-  | Cvlower : 'a var * 'a var list * changes -> changes
+  | Cvlower : 'a var * 'a var VarMap.t * changes -> changes
 
 let set_lower ~log v lower =
   log := Clower (v, v.lower, !log);
@@ -144,7 +150,12 @@ module Solver (L : Lattice) : Solver with type const := L.t = struct
 
   let fresh () =
     incr next_id;
-    { upper = L.max; lower = L.min; vlower = []; mvid = !next_id; mark = false }
+    { upper = L.max;
+      lower = L.min;
+      vlower = VarMap.empty;
+      mvid = !next_id;
+      mark = false
+    }
 
   exception NotSubmode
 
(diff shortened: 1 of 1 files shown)


lpw25:
I would mostly expect this to make most type-checking slower. I think these lists mostly shouldn't grow that large, so I'd be interested to see what pattern of code is managing to produce huge ones.

riaqn:
The assumption is correct: physical equality of `var` can be tested by comparing their `mvid`.

I remember discussing using `Set` instead of `List` with @lpw25 , and we were uncertain about the overhead of `Set`. How does this PR compare against the `main` branch for normal code? (i.e., codes that doesn't benefit extremely from `VarMap`).

Speaking of which, why do we use `Map` instead of `Set`? Or are they equivalent?

poechsel:
I don't have performance numbers yet for a full tree build but will have next week. Here the `Map` is needed instead of the `Set` has you still need to iterate over the different elements and a set can't have a `'a t` has the key.

lpw25:
You can still use a `Set` by just defining a comparison function of vars that just uses their id. Not sure it would meaningfully affect the performance, but is probably makes the code easier to understand.

lpw25:
Oh, sorry I misunderstood your point. You can still do it though, it "just" needs you to use recursive modules. It's almost [the classic example of them from the manual](https://v2.ocaml.org/manual/recursivemodules.html).

lpw25:
No, that's still not enough. You can do it with the `Set`s in base/core using things like `Base.Comparator.Make1`, but it might not be doable using the stdlib's set.

mshinwell:
@poechsel what is happening with this PR?

riaqn:
I think we might need to do similar things for the current newer mode solver, but we need to do some benchmark first to justify the  overhead. Also it's not gonna happen anytime soon.

riaqn:
Superseded by #3160

PR #1985, opened 2023-11-03
Remove ownership of the ocaml/ directory

goldfirere:
As title.

We seem to have converged on the idea of killing off CODEOWNERS, but I wasn't sure quite how far to take it. Please add more changes to this branch!

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -11,9 +11,6 @@ configure.ac            @mshinwell @xclerc
 
 flambda_backend.opam    @mshinwell @lthls
 
-HACKING.md      @mshinwell @goldfirere
-README.md       @mshinwell @goldfirere
-
 build_ocaml_compiler.sexp       @poechsel
 
 backend/                        @gretay-js @xclerc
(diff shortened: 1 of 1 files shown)


ncik-roberts:
I propose a more drastic removal (done in my latest commit). All that's left is:
  * `chamelon/`
  * `middle_end/`
  * `backend/`
  * `external/memtrace/`
  * Some misc top-level files: codeowners, an opam file, `build_ocaml_compiler.sexp`

I leave the ones where I think there's arguably a larger benefit to automatically blocking review on people / automatically notifying them when the files change.

Anything we should add back in?

lpw25 (approved):
I'm in favour of this change, but I wonder if it should be even more aggresive: @gretay-js and @mshinwell do you still want CODEOWNERS on for all of `backend` and `middle_end`, or should we remove those entries too?

mshinwell:
I think I'm happy to remove it from `middle_end/` at least for the moment.

goldfirere:
I've removed ownership from `middle_end` as per Mark's suggestion. Waiting for a signal from @gretay-js before pulling the trigger here.

gretay-js:
I don't think that we need exlicit code ownership in `backend`.

PR #1987, opened 2023-11-03
Fix Stdlib.Domain upstream build dependencies

TheNumbat:
Adds domain to stdlibmodules and regenerates dependencies. Also adds it as a dependency of the top level.


ncik-roberts (approved):
Looks good to me.

PR #1989, opened 2023-11-03
Make max arity local test less churny

ncik-roberts:
* Pass `-dno-unique-ids` to suppress a frequently-changing ID.
  * Make the output the same between flambda and closure by removing the need to make blocks (which was affecting the outputted code).
  * Now that the output is the same, just convert the tests to expect tests.

Reviewer: @goldfirere as this test was written to confirm the in-flight syntactic arity PR doesn't affect its output


mshinwell:
Thanks, this will be a great help

goldfirere (changes requested):
Does this still test what you want it to test? I'm not very familiar with this output format, but there seems to be less information here than previously.

ncik-roberts:
Yes, that's a silly mistake of mine. Expect tests run as bytecode and so have a huge max_arity. We need these tests to run as native code.

I've restored the flambda version of this test. I haven't restored the closure version as it doesn't contribute much more information and it's onerous to update test output for multiple backends.

Now I'd describe this PR as:
- making the test's output change less frequently
- when the tests' output does change, require only changing one middle end's output, not two

goldfirere:
This still seems a bit wrong. The `nlocal` values printed out don't change between tests, though they did in the old test. I don't see why that would happen, but it seems to be.

ncik-roberts:
Yes, it's because the functions are now locally allocated. I'll undo a change I made that was only needed to get flambda/non-flambda output to agree, anyway. (I deleted the non-flambda version of this test.)

ncik-roberts:
OK, this is ready for review again. Here's what I ran to check that the output before/after this PR is reasonable:

```
patdiff <(git show main:ocaml/testsuite/tests/syntactic-arity/max_arity_locals.flambda.reference | sed 's|/[0-9]\+||g' | tr -d '\n')  ocaml/testsuite/tests/syntactic-arity/max_arity_locals.compilers.reference 
```

This exits with no diff. The `sed` just removes the unique stamps, and the `tr` is to convince patdiff to try harder to ignore whitespace.

goldfirere (approved):

PR #1990, opened 2023-11-03
`--enable-warn-error` by default

ccasin:
This is implemented by passing `--enable-warn-error` to the subdirectory configure scripts if the user provided neither `--enable-warn-error` nor `--disable-warn-error`.  If they did provide either of those, it's already passed through.


mshinwell (approved):
The configure.ac change is ok

ccasin:
Subsumed by #2140

PR #1991, opened 2023-11-03
Fix "make bootstrap" and bootstrap

mshinwell:
Some new lambdas in `Target_system` prevent that code seeing the architecture variable being set to `none` during bootstrapping (new change in OCaml 5).
There is then a bootstrap, done with `make bootstrap`, included in this PR.  (The bootstrap compilers imported from PR1949 had been produced using dune.)


Gbury (approved):
I'm not sure I completely understand why we need this change, but at least it's correct and safe.

ncik-roberts:
CI passed for the relevant target (upstream build) so I merged.

PR #1992, opened 2023-11-03
Run CI only on push

TheNumbat:
We've been running CI both upon pushing a commit and refreshing a PR. Now it only runs on push. It should still be running the same set of jobs.


ncik-roberts (approved):
I think we should merge this as-is (before CI passes) to cut down on the deluge of CI jobs scheduled. If we run into issues later, we can revert this PR.

ncik-roberts:
And we can't merge, because of codeowners.

PR #1993, opened 2023-11-03
Fix arch.mli

TheNumbat:
The 5 merge added `arch.mli`, so the new `clmul_support` bool needs to be exposed there.


ncik-roberts (approved):

goldfirere:
This is a fixup for #1972, which failed in CI only after merging. Merging now to try to clear out the queue.

PR #1994, opened 2023-11-03
Remove C11 _Atomic typedefs

TheNumbat:
We currently build C stubs using runtime4/ with -std=gnu99, and `_Atomic` is not supported until C11.
When updating to the 5.0 runtime, we will need to update our C standard.


ncik-roberts (approved):
LGTM.

PR #1995, opened 2023-11-03
Ignore dynlink mlis

TheNumbat:
These are generated by the build; this .gitignores them.
Also removes `pull_request` from the other CI workflows.


ncik-roberts:
I'm surprised that we don't also want to delete dynlink mls. We deleted them originally in https://github.com/ocaml-flambda/flambda-backend/commits/main/ocaml/otherlibs/dynlink/byte/dynlink.ml. Maybe Mark has more context.

PR #1705, opened 2023-08-08
Add [Obj.uniquely_reachable_words]

apilatjs:
This function takes in a list of objects and for each one computes the number of words of memory that can be reached from that object, but no others in the list.

PR #2000, opened 2023-11-06
Retainer profiling for 5.x runtime

lukemaurer:
Move retainer profiling (#1705) to the OCaml 5 runtime. Checked against @apilatjs's upstream branch (https://github.com/apilatjs/ocaml/tree/retainer-profiling) so hopefully we can pull it in for multicore support in the future without much trouble.


mshinwell:
I added one CR relating to `add_to_long_value` which will need fixing in due course for domains.

mshinwell (approved):

lukemaurer:
> I added one CR relating to `add_to_long_value` which will need fixing in due course for domains.

I suspect it can't really be fixed for domains: currently it's only used in an array that's never shared with another domain, and if that changes (i.e., if we parallelize retainer profiling), we'll want to use atomics or something anyway.

mshinwell:
Right, I was thinking of using atomics.

PR #2004, opened 2023-11-07
Configure flag to use the 5.x runtime; make the 5.x runtime build

mshinwell:
This provides `--enable-runtime5` to `configure`.  The build will then use the 5.x runtime.

`make runtime-stdlib` (and `make hacking`) should work.

We will need to check that all runtime changes at the `last-4.14` tag are correctly ported over to the `runtime/` directory, which is the 5.x runtime.


mshinwell:
I've added a new CI job for `--enable-runtime5`.

There is a non-build-system change in this PR: the field that used to be called `exception_pointer` in the domain state table has been renamed to `exn_handler`, to match 5.x (although this seems a worse name, it is not a pointer to exception handling code).  This allows the emitters to build without conditional compilation being required for 4.x / 5.x.

mshinwell:
One minor thing: I removed "+runtime4" from the version number as it was proving problematic to have this dynamically determined based on a `configure` option.

TheNumbat (approved):
Looks good, `make runtime-stdlib` and `make hacking` work for me. It should run the rest of the CI if you merge in main now.

ocaml/runtime/caml/exec.h
```diff
@@ -60,7 +60,7 @@ struct exec_trailer {
 
 /* Magic number for this release */
 
-#define EXEC_MAGIC "Caml1999X033"
+#define EXEC_MAGIC "Caml1999X520"
```
xclerc, line 63: I wonder whether we should use a different
magic number since this runtime has additional
opcodes.

mshinwell: Agreed - have made a note elsewhere to deal with this.

PR #2005, opened 2023-11-07
Use bundled menhirlib and parser

Ekdohibs:
This avoid relying on having `menhir.20210419` installed when compiling.


mshinwell:
@stedolan any thoughts on this?

lthls:
Could we make progress on that ? I can't compile main without this patch.

Ekdohibs:
It now includes an autoconf switch: with `--with-menhir`, the behaviour previous to this PR will be preserved, while without it (or with `--without-menhir`), we use the pregenerated parsers.

PR #2008, opened 2023-11-07
Run CI on pull request + push to main

TheNumbat:
Running CI on push means it won't run on PRs from forks. This makes it run on all PRs + pushes to main. Pushes to branches without a corresponding PR will no longer run CI.


mshinwell (approved):

mshinwell:
Looks good, merging now so we get CI back to normal.

PR #2009, opened 2023-11-07
Labeled Tuples

ccasin:
(This code was written primarily by @rtjoa)

This adds support for labeled tuples.  See the tests for examples.  A couple notes about some choices we made:

- Reordering is supported only for patterns - expressions and types are never reordered.  Tuple patterns are only reordered if the type of the pattern is already known, making this considerably more restricted than how records work, but we think it's a good compromise for this feature.  Partial matching with `..` is also supported.

- In the typed tree, we've extended the existing tuples with labels rather than adding new constructors.  There are obvious tradeoffs (the main downside being that we're taking up space for labels that is unnecessary for legacy tuples).  This choice could be re-evaluated in the future.

Review:  This code has basically already been reviewed.  I reviewed Ryan's PRs into the labeled tuples branch as they were written.  I made a few additional PRs to fix up the parsing later, and @riaqn has reviewed these.  The main remaining thing to do is for a jane syntax expert to look over whether its use here is sensible/idiomatic - that code had to change a fair bit when I rebased this recently.  We previously decided @ncik-roberts would have a look at that.


ccasin:
(One additional note: It seems a few recent PRs should have updated the dependencies and bootstrapped, but didn't, which is why there are some unrelated changes to the `.depend` files here)

ccasin:
@Ekdohibs I believe this will need your sign-off for changes to `chamelon`.  Note that you already had a look at these changes when I merged them into the branch this PR comes from in #1929

ocaml/parsing/jane_syntax.mli
```diff
@@ -199,6 +199,63 @@ module N_ary_functions : sig
...
+        Invariant:
+        - If Closed, [n >= 2] and there is at least one label.
+        - If Open, [n >= 1]
+      *)
+
+  val typ_of :
+    loc:Location.t ->
+    attrs:Parsetree.attributes ->
+    core_type ->
+    Parsetree.core_type
```
ncik-roberts, line 247: The way that this (and `expr_of` and `pat_of`) handles attributes is inconsistent with the rest of Jane Syntax, and I think incorrect. I think this will duplicate attributes on labeled tuples in some cases, though I haven't been able to reproduce it in a test so I'm not 100% confident. I expect we should remove the `~attrs` argument to be consistent with e.g. `N_ary_functions.typ_of`.

ncik-roberts (approved):
The use of Jane Syntax looks good to me.

I checked two things:
  * Does the use of Jane Syntax look technically correct? (Yes, with a small comment.)
  * Did this require invasive refactors of `typecore.ml`, etc., as is typical for uses of Jane Syntax in its full glory?

Happily, I found that the answer to the second bullet point is apparently "no". The logic inlined into the `Pexp_tuple`, `Ppat_tuple`, etc. cases seems small enough that it doesn't look that bad to pull them out to a function.

If the answer to the second bullet point hadn't been "no", I may have recommend avoiding adding a new `Lexp_tuple` (etc.) constructor in favor of

ocaml/parsing/jane_syntax.mli
ccasin: Ah, nice catch.  I've made this change.

ncik-roberts:
Looks good to me. We should probably wait til the dust clears with OCaml 5 minus before merging.

Ekdohibs (approved):
For changes to `chamelon/`.

PR #2010, opened 2023-11-07
Proper compile errors for bad immediates

TheNumbat:
The compiler now reports a proper error message when SIMD instructions find an invalid immediate operand.

### DIFF
--- a/backend/amd64/simd_selection.ml
+++ b/backend/amd64/simd_selection.ml
@@ -23,14 +23,6 @@ type error = Bad_immediate of string
 
 exception Error of error
 
-let () =
-  let report_error ppf = function
-    | Bad_immediate msg -> Format.pp_print_string ppf msg
-  in
-  Location.register_error_of_exn (function
-    | Error err -> Some (Location.error_of_printer_file report_error err)
-    | _ -> None)
-
 let bad_immediate fmt =
   Format.kasprintf (fun msg -> raise (Error (Bad_immediate msg))) fmt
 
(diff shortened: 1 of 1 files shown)


backend/amd64/simd_selection.ml
```diff
@@ -19,6 +19,21 @@
...
+
+exception Error of error
+
+let () =
+  let report_error ppf = function
+    | Bad_immediate msg -> Format.pp_print_string ppf msg
+  in
+  Location.register_error_of_exn (function
+    | Error err -> Some (Location.error_of_printer_file report_error err)
+    | _ -> None)
```
ccasin, line 32: As a minor style point: This stuff usually appears at the end of the file, with `report_error` as a stand-alone definition.  I don't think there's any technical reason for this, but I suggest matching what is done in [other](https://github.com/ocaml-flambda/flambda-backend/blob/439417b90da18cf88ca6550e93ca2be2dba836dc/backend/asmgen.ml#L559) [files](https://github.com/ocaml-flambda/flambda-backend/blob/439417b90da18cf88ca6550e93ca2be2dba836dc/backend/asmlibrarian.ml#L119) in the [compiler](https://github.com/ocaml-flambda/flambda-backend/blob/439417b90da18cf88ca6550e93ca2be2dba836dc/ocaml/typing/typeclass.ml#L2092) just so that people looking for this code will find it in the expected place.

TheNumbat: Ok, moved to end of file

PR #1094, opened 2023-01-31
Restructure LIBDIR: Move Dynlink, Str and Unix to sub-directories (port part of ocaml/PR11198)

gretay-js:
This is a simpler verison of #895. 
Only changes the layout of `ocaml/otherlibs` and the corresponding build files.
Does not add missing libraries automatically and no alerts. No changes to `Load_path.init`.

PR #2012, opened 2023-11-08
Disable Compmisc.auto_include: do not add the directory, do not alert.

gretay-js:
Recover the behavior we had before ocaml-5 merge: do not automatically include directories from compiler distribution in the  load path and do not alert about missing directories. Instead, we will get `Error: Unbound module Unix` for example if `Unix` is not explicitly listed as a dependency of a library that directly depends on it. 

(See also https://github.com/ocaml-flambda/flambda-backend/pull/1094)


gretay-js:
Forgot to say that the new flag needs to be in `Clflags` becuase it is used in `ocaml/` subdirectory.

xclerc (approved):

mshinwell (approved):

PR #2014, opened 2023-11-08
Update Clambda_primitives.result_layout

alanechang:
Update `Clambda_primitives.result_layout` to be more like `Lambda.primitive_result_layout`. The lambda implementation is copied over and changed to fit the clambda type.

This PR fixes the flambda1 crash on:

```ocaml
module Float_u = Stdlib__Float_u
let of_int = Float_u.of_int
let () =
  let a = if Sys.opaque_identity true then of_int 1 else assert false in
  ()
```
> Added as a new testcase


mshinwell (approved):
Middle end parts only

PR #2006, opened 2023-11-07
Add 5.x systhreads code

mshinwell:
Based on #2004

We're considering `otherlibs/systhreads/` to be part of the runtime.  This PR puts the code in place and makes roughly the adjustments required for the dune build system.  The make build can be done later, once we're in a position to build this PR at all.

PR #2015, opened 2023-11-08
Add 5.x systhreads code

TheNumbat:
Copy of #2006 + fix because I can't push to @mshinwell's fork.


ccasin (approved):
Confirmed this is the previous PR + one commit, and that commit looks reasonable.

PR #2016, opened 2023-11-08
Add some native code tests for immediate/immediate64

ncik-roberts:
We (@ccasin , @goldfirere ) realized we didn't have many native code tests for types with immediate/immediate64 layouts. This PR adds a few small tests. The structure is adapted from `ocaml/testsuite/tests/typing-layouts-float64/unboxed_floats.ml`. The test added in this PR is much smaller than the `unboxed_floats.ml` test, and that's because immediates have been around for a while and need less testing. The test just contributes some higher-order uses of the new `immediate` and `immediate64` annotations, especially around tricky things like the write barrier.


goldfirere:
This looks good to me, but I actually think @mshinwell or designee would be the best person to look at this. Those folks know where the skeletons are buried.

ccasin (approved):
I concur that these tests looks good (thanks Nick!) and that it would be nice to see if Mark has any ideas about additional tests.  But it's fine to merge this and ask that question separately.

mshinwell:
Immediates in closures would seem like the other thing to test.

mshinwell:
(but those tests can be added separately as @ccasin said)

PR #938, opened 2022-11-02
Add is_last flag to closinfo words

mshinwell:
This is one of the ideas (the more space-efficient one) from #924.  Done very quickly so might be wrong!

PR #2017, opened 2023-11-08
Successful build with --enable-runtime5

TheNumbat:
Fixes runtime and stdlib-related build/link failures when building with `--enable-runtime5`.
- Properly installs runtime5/caml headers
- Disables stack allocation when runtime5 is enabled
- Adds a compile time constant indicating if we're configured with runtime5 or not.
- Conditionally implements `caml_with_async_exns` as a try/catch
- Copies implementation of `caml_obj_make_forward`
- Copies implementation of `caml_flambda2_invalid`
- Implements `caml_sys_const_naked_pointers_checked` as false
- Conditionally uses `pthread_sigmask` instead of `caml_sigmask_hook`
- Bootstraps compiler on runtime4

Compilation now succeeds, but running the ocaml tools segfaults. Will fix in a following PR.


Ekdohibs:
A possible cause of the segfault is that you need to backport the changes of https://github.com/ocaml-flambda/flambda-backend/pull/938 as well. On the other hand, the changes were needed only for the compactor, which is currently disabled; another possibility would simply be to generate the closinfo fields as they were prior to #938 when compiling for runtime5.

mshinwell:
I'm reviewing this.

mshinwell (approved):

PR #2020, opened 2023-11-09
Remove Linearize Pass

gretay-js:
Removes `linearize` pass and the corresponding CI job.

Also removes `Linear_to_cfg` pass and the related `Trap_stack` module, because we do not need to go from `Linear` IR back to `Cfg` any more (even for FDO). 

This PR is a draft for now. It is intended to follow shortly after #2013 and free up `-ocamlcfg` flag for something else, so I'm not removing the flag in this PR.


xclerc (approved):

gretay-js:
Converted to draft so we don't merge it by mistake. Teh plan is to merge it a bit later, after testing https://github.com/ocaml-flambda/flambda-backend/pull/2013 some more.

PR #2023, opened 2023-11-09
`tast_{iterator,mapper}` were missing locs in jkind annotations

ccasin:
This fixes a small bug in the ocaml 5 merge that I noticed while doing the merlin merge.

Review: @ncik-roberts ?


ncik-roberts (approved):
Good catch.

ncik-roberts:
Though it will now need to map over the second component of the tuple in light of #1907.

ccasin:
> Though it will now need to map over the second component of the tuple in light of #1907.

Oops!  (Somehow I was testing a different branch than I was working on.)

Perhaps take a quick second look, as the type of the iterator also had to change.

ncik-roberts (approved):
Seems like a good approach to me.

PR #2024, opened 2023-11-09
Upstream build with runtime5

TheNumbat:
- Building upstream with runtime5 requires running the boot compiler with a runtime4 `ocamlrun`
   - Therefore we check in `ocamlrun` in `boot/` so it's possible to run the boot compiler without runtime4.
   - Running `make coldstart` now only overwrites `boot/ocamlrun` if you're building runtime4.
   - Once the compiler itself can run on runtime5, we can get rid of this complication.


lpw25:
This seems something of a pain. I wonder if we should aim to be able to choose between runtimes at compile time rather than configure time, probably using the `-runtime-variant` flag. Then we'd always build both and so could run the bootstrap compiler -- or indeed all the compilers -- using the 4 runtime.

TheNumbat:
We're now planning on completing a working runtime5 compiler via the dune build and then using that as a 5 boot compiler.

PR #2025, opened 2023-11-10
Implement unboxed float literals

alanechang:


alanechang:
This PR adds support for unboxed float literals in the type system. It does so by redefining the `constant` type in the typedtree.

What this PR does not do is change the definition of `constant` in lambda. Instead, it relies on the `Pbox_float` and `Punbox_float` primitives when translating from typedtree to lambda. Changes to the lambda `constant` type can be its own separate PR.

Ekdohibs:
This looks quite complicated to me, wouldn't it be simpler to just extend the `constant` type? It seems to me in the current version, lambda uses `Parsetree.constant` inside `Const_base` where we would want the extended `Typedtree.constant`. It seems that most of the changes you introduced elsewhere wouldn't be needed in that case.

alanechang:
@Ekdohibs I agree using the extended `Typedtree.constant` type in lambda would be more ideal but that change doesn't seem exactly trivial: `declare_const` in flambda2 converts `Lambda.structured_constant` to `Field_of_static_block.t` which assumes things to have kind Value and I'm unsure if we can add unboxed float there (also, do we have to change flambda1&closure as well to keep them from breaking?).

And with that change, we still can't remove the workaround in `lambda/matching.ml` unless we introduce new lambda primitives for unboxed float comparisons, and the changes in `typing/` will need stay to deal with converting `Parsetree.constant` to `Typedtree.constant`.

ocaml/typing/untypeast.ml
```diff
@@ -130,14 +130,29 @@ let rec extract_letop_patterns n pat =
...
+  | Const_char c -> Pexp_constant (Pconst_char c)
+  | Const_string (s,loc,d) -> Pexp_constant (Pconst_string (s,loc,d))
+  | Const_int i -> Pexp_constant (Pconst_integer (Int.to_string i, None))
+  | Const_int32 i -> Pexp_constant (Pconst_integer (Int32.to_string i, Some 'l'))
+  | Const_int64 i -> Pexp_constant (Pconst_integer (Int64.to_string i, Some 'L'))
+  | Const_nativeint i -> Pexp_constant (Pconst_integer (Nativeint.to_string i, Some 'n'))
+  | Const_float f -> Pexp_constant (Pconst_float (f,None))
+  | Const_unboxed_float f ->
+    Jane_syntax.Layouts.expr_of ~loc (Lexp_constant (Float (f, None)))
+    |> add_jane_syntax_attributes
```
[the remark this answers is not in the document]
ncik-roberts: (you could even imagine exposing this as `constant`, reversing the decision to drop it from the mli, though I doubt this is too important)

ocaml/typing/typecore.ml
```diff
@@ -3021,7 +3008,6 @@ let rec check_counter_example_pat
       end
   | Tpat_alias (p, _, _, _, _) -> check_rec ~info p expected_ty k
   | Tpat_constant cst ->
-      let cst = constant_or_raise !env loc (Untypeast.constant cst) in
```
ncik-roberts, line 3024: Why is it ok to delete this line?

ocaml/testsuite/tests/typing-layouts/literals_native.ml
```diff
@@ -0,0 +1,61 @@
...
+
+  let ( + ) = add
+  let ( - ) = sub
+  let ( * ) = mul
+  let ( / ) = div
+  let ( ** ) = pow
+  let ( > ) x y = (compare x y) > 0
+end
+
+let test s f = Format.printf "%s: %f\n" s (Float_u.to_float f); Format.print_flush ()
```
ncik-roberts, line 21:
```suggestion
let test s f = Format.printf "%s: %f\n%!" s (Float_u.to_float f)
```

ocaml/testsuite/tests/typing-layouts/literals.ml
```diff
@@ -3,87 +3,156 @@
...
+    val max : float# -> float# -> float#
+    val min_num : t -> t -> t
+    val max_num : t -> t -> t
+    val ( + ) : float# -> float# -> float#
+    val ( - ) : float# -> float# -> float#
+    val ( * ) : float# -> float# -> float#
+    val ( / ) : float# -> float# -> float#
+    val ( ** ) : float# -> float# -> float#
+    val ( > ) : t -> t -> bool
+  end
```
ncik-roberts, line 100: Anything that can be done to avoid printing this? It seems sad to need to edit the output of this test every time the signature of `Float_u` changes.

Maybe we can just define `module Float_u = Stdlib__Float_u` and then define the additional operators separately? Or maybe wrapping this declaration with `open struct ... end` will suppress the printing?

ocaml/testsuite/tests/typing-layouts-float64/alloc.ml
```diff
@@ -167,3 +167,18 @@ let _ =
   let _ = measure_alloc_value (fun () -> cse_test false r) in
   let allocs = get_exact_allocations () in
   Printf.printf "CSE test (0 bytes):\n  allocated bytes: %.2f\n" allocs
+
+let[@inline never] literal_test () =
+  let open Float_u in
+  (#1. + #2.) * (#3. - #4.) / (#3. ** #1.)
```
ncik-roberts, line 173: I would be more convinced about zero-allocness if some of these floats were passed as args.

ocaml/typing/typedtree.ml
```diff
@@ -18,6 +18,16 @@
...
 
+type constant =
+    Const_int of int
+  | Const_char of char
+  | Const_string of string * Location.t * string option
+  | Const_float of string
+  | Const_unboxed_float of string
+  | Const_int32 of int32
+  | Const_int64 of int64
+  | Const_nativeint of nativeint
```
ncik-roberts, line 29: Is it much churnier to prefix these names with a `T`? I'd prefer that as it would then distinguish these constructors from the constructors with the same names in `Asttypes`.

ncik-roberts (commented):
The approach seems fine to me -- I'm holding off on approving as we're discussing this in a group tomorrow. Just a few cosmetic comments.

ocaml/testsuite/tests/typing-layouts/literals.ml
alanechang: I was able to get rid of this output.

ocaml/typing/typecore.ml
alanechang: Although I'm relatively confident that the constant here is always going to be well-formed (since we typed it once already) and we will just be getting the same thing back again from `constant_or_raise`, I did add the logic back in after the suggested refactor of `Untypeast.constant` which makes it possible.

ocaml/typing/untypeast.ml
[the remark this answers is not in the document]
alanechang: Thanks for the suggestion! I rewrote the `constant` function to share the common logic. I also added back the function in the mli just to use it in typecore (see other comment below).

ocaml/typing/typedtree.mli
```diff
@@ -23,6 +23,16 @@
...
 
+type constant =
+    Const_int of int
+  | Const_char of char
+  | Const_string of string * Location.t * string option
+  | Const_float of string
+  | Const_unboxed_float of string
+  | Const_int32 of int32
+  | Const_int64 of int64
+  | Const_nativeint of nativeint
```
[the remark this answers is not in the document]
alanechang: I added a short comment here to describe the context and its intended purpose.

ocaml/testsuite/tests/typing-layouts-float64/alloc.ml
alanechang: Made that change and it now passes in a few unboxed floats as arguments

ocaml/typing/typedtree.ml
alanechang: I believe the idea is to eventually share this type with lambda and completely replace the usage of `Asttypes.constant`. In which case, I'm not sure if renaming it now and changing it back shortly after is worth the effort.

ncik-roberts (approved):
Thanks, this change looks good to me. As discussed offline, we don't think that this will rule out all possible allocations with using unboxed literals, but this is OK for `layouts_alpha`.

PR #2031, opened 2023-11-14
Disable zstd to avoid dependency

mshinwell:
This was introduced for compressed marshalling, but it produces a dependency on zstd for every OCaml-produced executable, which we don't want at the moment.


mshinwell:
I tested this by building on Linux and running ldd on ocamlopt.opt

TheNumbat (approved):

PR #2032, opened 2023-11-15
Add timestamp to GC log

riaqn:
This forward ports https://github.com/ocaml-flambda/ocaml-jst/commit/b16493b317a8851789a00c4195ffa3ce1107d0e8


lpw25:
This PR now seems broken. I was just about to submit a review when it jumped from 1 commit to 31.

riaqn:
@lpw25 I addressed the comment and it's ready for review again.

lpw25 (approved):

stedolan:
This PR needs fixing: there are two logging functions in the runtime now (`caml_gc_log` and `caml_gc_message`), and this only adds timestamps to one of them. (I don't know why there are two)

PR #2035, opened 2023-11-15
add `caml_get_header` to runtime5

riaqn:
Forward port the runtime part of https://github.com/ocaml-flambda/ocaml-jst/commit/5be3cb82b572ff2b72cb8640acef32add8a25de3

Note that as of now, runtime5 doesn't have stack allocation, so `lib-obj/get_header.ml` will fail as some object are on heap (while expected to be on stack)

### DIFF
--- a/ocaml/runtime/obj.c
+++ b/ocaml/runtime/obj.c
@@ -73,6 +73,13 @@ CAMLprim value caml_obj_make_forward(value blk, value fwd)
   return Val_unit;
 }
 
+CAMLprim value caml_get_header(value blk)
+{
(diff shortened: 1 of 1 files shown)


mshinwell:
@riaqn can you fix the test case so that it passes when stack allocation is disabled too?  There are some examples in `typing-local/` as to how to do this, using the `stack-allocation` ocamltest predicate.

riaqn:
I suggest postpone this PR until proper local allocation is supported in runtime; because currently `is_stack` cannot be defined, becuase I don't know what would be the color for stack objects.

PR #2036, opened 2023-11-15
Fix `~region` in list comprehensions

ccasin:
@Ekdohibs spotted this suspicious code over in #1871, and I agree it looks like a bug.

The list comprehensions code builds up intermediate list results as local lists, and then at the end constructs a global list to return to the user (see long comment at the top of `transl_list_comprehension.ml`).  This code, `translate_bindings` constructs a function that builds such a local list (if you trace it's implementation you'll see that the body of this function is eventually constructed using `rev_list_snoc_local`, which explicitly constructs a local block).  Since it intends to return this locally allocated list to its caller, it should not have a region.  But it does, and this PR fixes that.

This could be a rather bad bug, but I weakly believe it's not actually observable because there's no opportunity for the stack to get smashed after the function constructed by `translate_bindings` is called and before we use its result to build the final global list.  So, I haven't added a new test.

Review: @goldfirere, @riaqn, @lpw25 or @stedolan

### DIFF
--- a/ocaml/lambda/transl_list_comprehension.ml
+++ b/ocaml/lambda/transl_list_comprehension.ml
@@ -259,7 +259,7 @@ let rec translate_bindings
           ~attr:default_function_attribute
           ~loc
           ~mode:alloc_local
-          ~region:true (* One region per iterator, like for loops *)
+          ~region:false
           ~body:(add_bindings body)
       in
       let result =


ccasin:
Nick and I looked a little harder at trying to make a bug out of this.  I mistakenly thought this `~region` bool was determining whether this function gets a region, but that doesn't seem to be the case.  I'm now not sure what it's used for, and whether this is actually a bug.

Ekdohibs:
It looks like it is used only in the middle-end to determine the modes for over-applications, and is an optimization on simply passing the mode: `region = true` means the function does not allocate in the parent region, and thus in the case of an over-application, we know that we do not have to cleanup the region afterwards.
In our case we probably can't trigger the bug, since the function returns a list and therefore can't be over-applied, although I'm not 100% sure the `region` flag is used only for that.

goldfirere (approved):

PR #2038, opened 2023-11-16
Backport lazy from OCaml 5 to runtime4

xclerc:


backend/cmm_helpers.ml
```diff
@@ -826,9 +826,7 @@ let get_header ptr dbg =
      data race on headers. This saves performance with ThreadSanitizer
      instrumentation by avoiding to instrument header loads. *)
   Cop
-    ( (* BACKPORT BEGIN mk_load_immut Word_int, *)
-      mk_load_mut Word_int,
-      (* BACKPORT END *)
+    ( mk_load_immut Word_int,
```
Ekdohibs, line 829: I believe this is safe, but we should probably completely remove the `obj_set_tag` and `obj_truncate` functions (currently they are not declared in `obj.ml`, but remain available for anyone to declare as external).

ocaml/runtime4/obj.c
```diff
@@ -27,22 +28,34 @@
...
+    return 1000;   /* int_tag */
   }else if ((long) arg & (sizeof (value) - 1)){
-    return Val_int (1002);   /* unaligned_tag */
+    return 1002;   /* unaligned_tag */
   }else if (Is_in_value_area (arg)){
-    return Val_int(Tag_val(arg));
+    /* The acquire load ensures that reading the field of a Forward_tag
+       block in stdlib/camlinternalLazy.ml:force_gen has the necessary
+       synchronization. */
+    hd = (header_t)atomic_load_acquire(Hp_atomic_val(arg));
```
Ekdohibs, line 47: I'm not sure the atomic_load here is necessary, since we have a single domain at all times, and thus there should be no synchronization needed. Maybe someone more knowledgeable about this could give their opinion?

Ekdohibs (approved):
This looks good; I was a bit worried about the interaction of flambda2 with the code of camlinternalLazy which does not prevent inlining and mutates tags while flambda believes they are immutable, but since we never re-read the tag, this looks safe.

lthls:
> This looks good; I was a bit worried about the interaction of flambda2 with the code of camlinternalLazy which does not prevent inlining and mutates tags while flambda believes they are immutable, but since we never re-read the tag, this looks safe.

The invariant should be: we never read the tag of a possibly-lazy value from ocamlopt-generated code. Tag reads that are allowed to be lazy tags (by the type system) should always go through `caml_obj_tag`, which is opaque to the compiler.

ocaml/runtime4/obj.c
mshinwell: This sounds correct to me.

backend/cmm_helpers.ml
mshinwell: I'd like to think about this a bit more before we merge (@xclerc let's discuss)

ocaml/lambda/matching.ml
```diff
@@ -1963,6 +1960,12 @@ let call_force_lazy_block ?(inlined = Default_inlined) varg loc ~pos =
 
 let lazy_forward_field = Lambda.Pfield (0, Pointer, Reads_vary)
 
+(* CR xclerc for xclerc: temporary workaround. *)
+module Obj = struct
+  include Obj
+  let forcing_tag = 244
+end
```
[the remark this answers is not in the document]
xclerc: (See https://github.com/ocaml-flambda/flambda-backend/pull/2063)

mshinwell (approved):
I read the changesets since the previous approval

PR #2042, opened 2023-11-16
(SIMD) Port #1935

riaqn:
This PR ports the runtime part of #1935


lpw25:
This seems to have many more commits that intended

riaqn:
This PR probably depends on #2044 to make sense.

TheNumbat (approved):

TheNumbat:
This does need to merge into #2044 - also looks like it needs rebasing to fix the CI

PR #797, opened 2022-08-30
Long frames in frametable

gretay-js:
Current frametable format restricts stack frame size to 16 bit for compactness. In some corner cases (e.g., owl library's autogenerated file) when compiling with `-linscan`, this limit can be exceeded (e.g., see #791).  Usually the right way to handle this error is to restructure the code, but sometimes it is hard to do quickly and a temporary workaround is needed. 
This PR proposes a way to emit long frametable enteries as an extension of the current format. 
Add a compiler flag -allow-long-frames to turn on the new functionality.

PR #2048, opened 2023-11-20
port #797 (long frames in frametable)

riaqn:
This ports #797. 

I checked that all direct access of `->live_ofs` and `->num_live` are safe. Remaining accesses are indirect via the `frame_end_of_live_ofs` which are aware of long format. Similar for `->frame_data`.


gretay-js (changes requested):

riaqn:
@gretay-js Thanks for the review - can you look again?

gretay-js (approved):

PR #2050, opened 2023-11-20
Port runtime of #1682

riaqn:
This ports the runtime part of #1682


TheNumbat (approved):
Looks good (other than the failing any.ml test).

PR #2051, opened 2023-11-20
OCaml 5: Version includes "minus"

ncik-roberts:
Pushing this so @mshinwell can look. I don't we should merge this because of the comment in `ocaml/build-aux/ocaml_version.m4`:

```
# The OCAML__VERSION below must be in the format specified in stdlib/sys.mli
```

which comment:

```
val ocaml_version : string
(** [ocaml_version] is the version of OCaml.
    It is a string of the form
      ["major.minor[.patchlevel][(+|~)additional-info]"],
    where [major], [minor], and [patchlevel] are integers, and
    [additional-info] is an arbitrary string.
    The [[.patchlevel]] part was absent before version 3.08.0 and
    became mandatory from 3.08.0 onwards.
    The [[(+|~)additional-info]] part may be absent. *)
```


mshinwell:
We've decided not to do this and will fix some JS-internal tooling instead.

PR #154, opened 2021-08-10
Add -mbranches-within-32B to major GC compilation (where available)

stedolan:
The major GC performs better with this flag on Intel processors

This is a workaround for an Intel CPU bug: <https://www.intel.co.uk/content/www/uk/en/support/articles/000055650/processors.html>

PR #2055, opened 2023-11-20
Ports runtime of #154 (Add -mbranches-within-32B to major GC compilation)

riaqn:
This ports runtime of #154. 

Request review from @stedolan  who authored the original PR.

### DIFF
--- a/ocaml/Makefile
+++ b/ocaml/Makefile
@@ -1015,7 +1015,7 @@ $(DEPDIR)/$(RUNTIME_DIR)/%.npic.$(D): OC_CPPFLAGS += $(OC_NATIVE_CPPFLAGS)
 # The major GC performs better with this flag on Intel processors
 # This is a workaround for an Intel CPU bug:
 # https://www.intel.co.uk/content/www/uk/en/support/articles/000055650/processors.html
-$(RUNTIME_DIR)/major_gc.%: OC_CFLAGS += $(INTEL_JCC_BUG_CFLAGS)
+$(RUNTIME_DIR)/major_gc.$(O): OC_CFLAGS += $(INTEL_JCC_BUG_CFLAGS)
 
 ## Compilation of runtime C files
 
--- a/ocaml/Makefile
+++ b/ocaml/Makefile
@@ -1016,7 +1016,6 @@ $(DEPDIR)/$(RUNTIME_DIR)/%.npic.$(D): OC_CPPFLAGS += $(OC_NATIVE_CPPFLAGS)
 # This is a workaround for an Intel CPU bug:
 # https://www.intel.co.uk/content/www/uk/en/support/articles/000055650/processors.html
 $(RUNTIME_DIR)/major_gc.%: OC_CFLAGS += $(INTEL_JCC_BUG_CFLAGS)
-$(DEPDIR)/$(RUNTIME_DIR)/major_gc.%: OC_CPPFLAGS += $(INTEL_JCC_BUG_CFLAGS)
 
 ## Compilation of runtime C files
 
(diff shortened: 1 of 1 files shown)


ocaml/Makefile
```diff
@@ -1011,6 +1011,11 @@ $(RUNTIME_DIR)/%.npic.$(O): OC_CFLAGS += $(OC_NATIVE_CFLAGS) $(SHAREDLIB_CFLAGS)
 $(RUNTIME_DIR)/%.npic.$(O): OC_CPPFLAGS += $(OC_NATIVE_CPPFLAGS)
 $(DEPDIR)/$(RUNTIME_DIR)/%.npic.$(D): OC_CPPFLAGS += $(OC_NATIVE_CPPFLAGS)
 
+# The major GC performs better with this flag on Intel processors
+# This is a workaround for an Intel CPU bug:
+# https://www.intel.co.uk/content/www/uk/en/support/articles/000055650/processors.html
+$(RUNTIME_DIR)/major_gc.%: OC_CFLAGS += $(INTEL_JCC_BUG_CFLAGS)
+
```
riaqn, line 1018: Please check: I'm not sure if `$(DEPDIR)/$(RUNTIME_DIR)/major_gc.%` needs to be involved? If they only care about `CPPFLAGS`, then maybe not?

stedolan:
To verify that this is done correctly, please have a look at the assembly of `do_some_marking` in `major_gc.c`. It should have branches aligned to 32 bytes by inserting `nops`.

riaqn:
@stedolan As you sugggested earlier that `dune` based build will also be impacted by this, so I just ran `make` in the top directory, and find the following `.o` files:
```
./_build/runtime_stdlib/ocaml/runtime/major_gc.c
./_build/runtime_stdlib/ocaml/runtime/major_gc.nd.o
./_build/runtime_stdlib/ocaml/runtime/major_gc.ni.o
./_build/runtime_stdlib/ocaml/runtime/major_gc.npic.o
./_build/runtime_stdlib/ocaml/runtime/major_gc.b.o
./_build/runtime_stdlib/ocaml/runtime/major_gc.bd.o
./_build/runtime_stdlib/ocaml/runtime/major_gc.bi.o
./_build/runtime_stdlib/ocaml/runtime/major_gc.bpic.o
./_build/runtime_stdlib/ocaml/runtime/major_gc.n.o
```
I then run `objdump -d` on those - and find the addresses jumped to are not aligned to 32 bytes:
```
55:   72 59                   jb     b0 <do_some_marking+0xb0>
  57:   48 85 ed                test   %rbp,%rbp
  5a:   7e 0d                   jle    69 <do_some_marking+0x69>
  5c:   49 8b 47 08             mov    0x8(%r15),%rax
  60:   48 85 c0                test   %rax,%rax
  63:   0f 85 f7 00 00 00       jne    160 <do_some_marking+0x160>
  69:   4d 85 f6                test   %r14,%r14
  6c:   0f 85 de 00 00 00       jne    150 <do_some_marking+0x150>
```

riaqn:
UPDATE: we verified that `INTEL_JCC_BUG_CFLAGS` is properly passed, but the flag is empty on my local machine because my `as` doesn't supports it.

ocaml/Makefile
riaqn: I believe this is not needed - I asked the above question because I saw lines such as:
```
$(DEPDIR)/$(RUNTIME_DIR)/%.npic.$(D): OC_CPPFLAGS += $(OC_NATIVE_CPPFLAGS)
```
but `$(D)` seems to be dependency descriptions (and thus need pre-processor flags correctly set).

The flag we intend to add has nothing to do with dependency or pre-processor, only with assembler.

xclerc (approved):

PR #2058, opened 2023-11-21
Fixes for runtest on runtime5

mshinwell:
- Fixes to checkmach related to symbol separators
- Likewise for flexpect...
- ...but also flexpect is now disabled when stack allocation is off.  This is only the case whilst locals is being ported to runtime5, and it's complicated to support otherwise.

### DIFF
--- a/middle_end/flambda2/tests/tools/flexpect.ml
+++ b/middle_end/flambda2/tests/tools/flexpect.ml
@@ -118,6 +118,10 @@ let run_mdflx_file filename : Outcome.t =
     Error
 
 let _ =
+  if not Config.stack_allocation
+  then (
+    Printf.printf "flexpect not supported when stack allocation disabled";
+    exit 0);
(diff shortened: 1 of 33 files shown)


mshinwell:
@xclerc to review please

mshinwell:
Some other minor fixes have been added

middle_end/flambda2/tests/tools/flexpect.ml
```diff
@@ -118,6 +118,11 @@ let run_mdflx_file filename : Outcome.t =
     Error
 
 let _ =
+  Symbol0.force_runtime4_symbols ();
+  if not Config.stack_allocation
```
xclerc, line 122: Maybe leave a CR to revisit?
(assuming that `Config.runtime5`
will be removed soon, while
`Config.stack_allocation` will stay)

xclerc (approved):

PR #2060, opened 2023-11-21
Implement 5.x atomic primitives in Flambda 2 and use them in the stdlib

mshinwell:
I spoke with @stedolan about the setting of the effects/coeffects judgements for these.


lthls (approved):
Looks correct, but the new `Cmm_helpers` functions seem to be redundant with the matching code in `Cmmgen`. It would be nice to avoid duplication (maybe you could add a CR and leave that for another PR ?).

mshinwell:
We're about to delete Cmmgen though, so not sure it's worth doing anything?

lthls:
My solution would have been to patch `Cmmgen` to use the new helper functions; not doing anything works too.

PR #2061, opened 2023-11-21
Refactor Cfgize Poptrap insertion

gretay-js:
Record handler label for `Itrywith` in `Cfgize.state.iends_with_poptrap`. The label is not yet used in this PR.
In preparation for deleting dead push/pop trap instructions.

### DIFF
--- a/backend/cfg/cfgize.ml
+++ b/backend/cfg/cfgize.ml
@@ -32,9 +32,11 @@ module State : sig
 
   val get_next_instruction_id : t -> int
 
-  val add_iend_with_poptrap : t -> Mach.instruction -> unit
+  val add_iend_with_poptrap : t -> Mach.instruction -> Label.t -> unit
 
   val is_iend_with_poptrap : t -> Mach.instruction -> bool
+
+  val get_iend_with_poptrap : t -> Mach.instruction -> Label.t option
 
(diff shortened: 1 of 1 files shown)


backend/cfg/cfgize.ml
```diff
@@ -628,7 +633,7 @@ let rec add_blocks :
       terminate_block ~trap_actions:[]
         (copy_instruction_no_reg state last ~desc:(Cfg.Always label_body));
       let next, add_next_block = prepare_next_block () in
```
gretay-js, line 635: @xclerc  something strange is going on here. 

- `add_iend_with_poptrap` is recorded for both delayed and regular `Itrywith`
- it is always converted to `Pop_generic` (this is fine because we don't yet use the argument of `Cmm.Pop`)
- `prepare_next_block` modifies state, so moving `add_iend_with_poptrap` earlier (to distinguish the two kinds) breaks the build.
- is it possible to mistakenly generate two `Poptrap` instructions with `flambda2`? one from this `iend_with_pop` and the other from `trap_actions` that come from `Iexit` and `Ireturn`.  I'm thinking some

xclerc (approved):

mshinwell:
This one shouldn't be merged on its own.  @gretay-js will look

PR #2062, opened 2023-11-21
Punboxed_float_comp lambda primitive

alanechang:


alanechang:
This PR follows #2025 and adds a new `Punboxed_float_comp` lambda primitive for comparing unboxed floats. The new primitive is implemented in closure, flambda1, flambda2, but the simplif step only works in flambda2 and not in the other two middle ends.

Currently, the only usage of this new primitive is in `lambda/matching.ml` for translating match expressions.

The new `transl_constant` function added to `lambda` in this PR will be removed in the subsequent PR #2071.

This is now ready for review.

PR #2063, opened 2023-11-22
Add `Runtimetags` module

xclerc:
This pull requests extracts the parts of https://github.com/ocaml-flambda/flambda-backend/pull/1857 useful for https://github.com/ocaml-flambda/flambda-backend/pull/2038,
_i.e._ the module with the tag definitions.


lpw25 (approved):
LGTM. Perhaps it could use a test to check that these are in sync between `Obj` and `Runtimetags` in compiler-libs, although these don't change often or lightly, so it might also be fine without if that proves too awkward for some reason.

PR #2066, opened 2023-11-22
Support for locals in the 5.x runtime

stedolan:
Mostly ported over with the necessary changes from 4.14

The nontrivial change is that the GC colours are different: now,  stack allocations are always `NOT_MARKABLE`. This colour is shared with externals and statically allocated data, so more work needs to be done to determine whether a block is local.


mshinwell:
I'm going to review this

mshinwell (approved):
We should discuss the use of the top bit for the mark flag, since that seems like it could interfere with any uses of the reserved header space (e.g. for mixed blocks).  Fine for now though, and everything else looked correct to me.

mshinwell:
I'm investigating the pr902 test failure

PR #2068, opened 2023-11-22
Ports upstream #12439 (Fix custom block promotion)

riaqn:
This trivially ports https://github.com/ocaml/ocaml/pull/12439 . No tests so correctness is unknown.


TheNumbat (approved):
(Diff is the same, but I also can't comment on correctness)

PR #2070, opened 2023-11-22
Records mixing immediates, floats, and float64s

ccasin:
This adds support for records containing any mix of floats, float64s, and immediates.  The records are represented with a block that has abstract tag.

Before this can be merged, it is going to need some contributions from flambda2 people (and also a careful review from backend/runtime people).  I've left `XXX layouts` in the relevant places.  Additionally, the current implementation is only correct for 64-bit systems.  I think that's fine, though?

This was slightly trickier than I expected it to be.  The main complexities are:
- A new bytecode instruction is needed for creating these abstract records.  Unboxed floats are boxed in bytecode, but bytecode does have the float record optimization.  So when constructing such a record, we check each element to see if it is boxed, in which case it must be a `float`, and unbox it.  
- I had hoped to get away without adding new primitives for mutation and projection.  We can very nearly get away with just using the existing primitives appropriate for the type of the particular field - in fact this seems to work fine on closure and flambda1 - but in flambda2 there are assumptions in various places that, e.g., when projecting a naked float field, the block in question is a flat float block.  Given that, it seemed easier to add new primitives than attempt to generalize the existing ones.

@mshinwell to take an initial look for the middle-end and points beyond.
@goldfirere to review or delegate for the front-end


ocaml/middle_end/printclambda_primitives.ml
```diff
@@ -167,6 +175,16 @@ let primitive ppf (prim:Clambda_primitives.primitive) =
...
       fprintf ppf "setufloatfield%s %i" init n
+  | Psetabstractfield (n, shape, init) ->
+      let init =
+        match init with
+        | Heap_initialization -> "(heap-init)"
+        | Root_initialization -> "(root-init)"
+        | Assignment Modify_heap -> ""
+        | Assignment Modify_maybe_stack -> "(maybe-stack)"
+      in
+      fprintf ppf "setufloatfield%s %i %a"
```
ncik-roberts, line 186:
```suggestion
      fprintf ppf "setabstractfield%s %i %a"
```

PR #2071, opened 2023-11-22
Change lambda constant type to support native unboxed float

alanechang:


alanechang:
This PR follows #2062 and changes the `constant` type inside of `Lambda.structured_constant` to allow unboxed floats in `Const_base`. This simplifies the translation of unboxed float literals to lambda and not longer requires the use of `Punbox_float`.

The flambda2 middle end is updated to support this change, but closure and flambda1 will now crash when encountering unboxed float literals.

This is ready for review.

mshinwell (approved):
A few minor things to fix, but ok to merge without re-review once those have been fixed.

PR #2072, opened 2023-11-23
Remove Config.runtime5 check in dll.ml; guard the one in symbol.ml

mshinwell:
Two minor changes working towards the goal of not having any `Config.runtime5` conditionals in the bytecode compiler.

1. Harmonize caml_dynlink_open_lib interface between 4 and 5 runtimes, by adjusting the 5 runtime.
2. Check in `ocaml/utils/symbol.ml` that the `Config.runtime5` check is only executed in the native code compiler.

### DIFF
--- a/ocaml/runtime/dynlink.c
+++ b/ocaml/runtime/dynlink.c
@@ -235,7 +235,9 @@ void caml_free_shared_libs(void)
 
 #define Handle_val(v) (*((void **) (v)))
 
-CAMLprim value caml_dynlink_open_lib(value filename)
+/* The mode argument is here for compatibility with runtime4. */
+/* CR ocaml 5 runtime: Remove [mode] when all-runtime5. */
+CAMLprim value caml_dynlink_open_lib(value mode, value filename)
 {
   void * handle;
   value result;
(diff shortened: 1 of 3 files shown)


ocaml/runtime/dynlink.c
```diff
@@ -235,7 +235,9 @@ void caml_free_shared_libs(void)
 
 #define Handle_val(v) (*((void **) (v)))
 
-CAMLprim value caml_dynlink_open_lib(value filename)
+/* The mode argument is here for compatibility with runtime4. */
+/* CR ocaml 5 runtime: Remove [mode] when all-runtime5. */
+CAMLprim value caml_dynlink_open_lib(value mode, value filename)
```
xclerc, line 240: (I am susprised you do not get a
warning, but we can deal with
that later.)

xclerc (approved):

ocaml/runtime/dynlink.c
mshinwell: Oh yes - whoops

mshinwell: Ah sorry I was misunderstanding your comment.  I think that particular warning is off.

xclerc (approved):

PR #2075, opened 2023-11-23
add `in_minor_collection` and port #1743 (Add some checks that the minor GC does not recurse)

riaqn:
This ports #1743.

The original PR relies on `Caml_state->in_minor_collection` which doesn't exist yet (neither does the upstream). So we add that field as well. The field is updated right after `caml_minor_gc_begin_hook` is called, and again right before `caml_minor_gc_end_hook` is called. I hope that's sensible.


xclerc (approved):
(We could add the `#define caml_in_minor_collection ...` of
`minor_gc.h`, but that can also happen in another pull request.)

PR #2076, opened 2023-11-23
Port "condvar" bugfix to runtime 5

xclerc:


stedolan (approved):
I think the custom condvar implementation has a performance problem when a single thread rapidly takes and releases the systhreads lock, but we've lived with it in ocaml 4 for quite a while so we can use it as is in ocaml 5 for now as well.

stedolan:
There's an issue in this patch: the runtime build now produces warnings of this form:
```
runtime/domain.c:206:5: note: (near initialization for ‘all_domains_cond.cond.counter’)
runtime/domain.c:206:5: warning: braces around scalar initializer
     CAML_PLAT_COND_INITIALIZER(&all_domains_lock);
     ^~~~~~~~~~~~~~~~~~~~~~~~~~
```
The trouble is that CAML_PLAT_COND_INITIALIZER is still using `PTHREAD_COND_INITIALIZER`, which needs to be updated for the custom condvar implementation.

PR #1411, opened 2023-05-24
Extend caml_locking_scheme with callbacks for thread start/stop

stedolan:
Adds two more callbacks to the structure introduced by #1365

PR #2077, opened 2023-11-23
Ports custom runtime locking

riaqn:
This ports a series of PRs related to custom runtime locking: #1365, #1411, #1429 

This does not port #1869 or the follow-up bug fix #1976. Note that #1869 is reverted by #1984.

- A major difficulty in this porting PR is that the default runtime lock `st_masterlock`  tries to acquire the `domain_lock` as well. Our API allows user to replace the `st_masterlock` with their own, which probably won't acquire `domain_lock`. This is only safe if there is no backup-thread in the current domain, which is ensured by inserting an `abort` in `backup_thread_func`. This would prevent mutliple-domain.
- I also removed acquiring `domain_lock` from `st_masterlock` just for consistency. As a result, `domain_lock` is rendered pretty much redundent. I didn't remove it for minimal diff and future reference - let me know if I should. 
- Note that `domain_lock` has a further security check, which is that the domain runtime state pointer `caml_state` is a thread-local variable and only not-null for the current active thread. This check has to be removed. because a thread could acquire the lock in python (and thus not setting `caml_state`), then enter ocaml assuming it has the lock (and thus `caml_state` not null). I believe our current approach should ensure that `caml_state` is always not-null for all systhreads. 
- Diff that will have to change in the future with multi-domain are marked with "single-domain hack", so it's easier to find and fix in the future.
- Comments starting with `CR zqian` are intended to initiate review conversion, and will be removed right before merging.
- Note that in `save_runtime_state` we talk about `Active_thread` instead of `This_thread`. I think it doesn't make much difference for this PR, but will be needed when we want to port #1869.

Request review from @stedolan

### DIFF
--- a/ocaml/otherlibs/systhreads/st_stubs.c
+++ b/ocaml/otherlibs/systhreads/st_stubs.c
@@ -241,9 +241,7 @@ static void caml_thread_scan_roots(
 
 static void save_runtime_state(void)
 {
-  /* CR zqian: we save to [active_thread] instead of [this_thread]. I believe
-  they are equivalent here, but I think use [active_thread] is easier to
-  understand, also follows the systhreads4 behavior. */
+  CAMLassert(Active_thread == This_thread);
   caml_thread_t th = Active_thread;
   CAMLassert(th != NULL);
   th->current_stack = Caml_state->current_stack;
@@ -472,7 +470,7 @@ static void caml_thread_reinitialize(void)
   Active_thread->next = Active_thread;
   Active_thread->prev = Active_thread;
 
-  // Single-domain hack: systhreads doesn't maintain domain lock
+  // CR ocaml 5 domains: systhreads doesn't maintain domain lock
   /* Within the child, the domain_lock needs to be reset and acquired. */
   // caml_reset_domain_lock();
   // caml_acquire_domain_lock();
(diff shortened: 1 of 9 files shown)


ocaml/otherlibs/systhreads/st_stubs.c
```diff
@@ -194,27 +243,35 @@ static void save_runtime_state(void)
...
-  this_thread->exn_handler = Caml_state->exn_handler;
-  this_thread->async_exn_handler = Caml_state->async_exn_handler;
-  this_thread->local_roots = Caml_state->local_roots;
-  this_thread->local_arenas = caml_get_local_arenas(Caml_state);
-  this_thread->backtrace_pos = Caml_state->backtrace_pos;
-  this_thread->backtrace_buffer = Caml_state->backtrace_buffer;
-  this_thread->backtrace_last_exn = Caml_state->backtrace_last_exn;
+  /* CR zqian: we save to [active_thread] instead of [this_thread]. I believe
+  they are equivalent here, but I think use [active_thread] is easier to
+  understand, also follows the systhreads4 behavior. */
```
stedolan, line 248: Yes, Active_thread makes more sense to me too, but I agree they should be the same here: to call `save_runtime_state`, the current thread should be active. Maybe add a `CAMLassert` that they're equal?

PR #2078, opened 2023-11-23
Port upstream's #12193 (compaction)

xclerc:


mshinwell:
I'm reviewing

mshinwell:
There was a mistake in the definition of `POOL_BLOCKS` (`shared_heap.c`) and also indentation mismatching upstream in `pool_free`.  As far as I could see, everything else is ok - I used a diff of diffs.

mshinwell (approved):

PR #2081, opened 2023-11-24
zero alloc: better error message for allocations due to flat float array

gretay-js:
Zero alloc check fails (correctly) on the following example, because of spurious allocations related to flat float array.
We generate them on a path that is not feasible, but the compiler does not have enough type information to eliminate it. 
Users find these failures puzzling. This is a workaround to track such allocations and report them slightly differently.

Approach:
- add field `from_flat_float_array:bool` to `Debuginfo.t`
- use this field in `Checkmach` to record that an `IAlloc` instruction that it originates from flat float array boxing a value.
- this field is set only in flambda2 mode.
- it may not cover all allocations associated with flat float array
- it may be lost during existing or future flambda2 optimizations
- it is easy to identify relevant allocations during lambda to flambda2 translation, but debug info is not readily accessible at that layer of abstraction. i've attached this metadata as an argument of `Boxable_number.Naked_float` and then at `to_cmm` it can be placed on the relevant `Debuginfo.t` for allocation instructions.

This is most certainly not the right way to track these allocations and may not work on more complex examples. Feedback from flambda2 devs will be much appreciated. 

```
type t = Bigstring.t array
let[@zero_alloc] set t ~table pos word =  Bigstring.set_int64 t.(table) pos word
let[@zero_alloc] get t ~table pos = exclave_ Bigstring.get_int64 t.(table) pos
```

### DIFF
--- a/backend/checkmach.ml
+++ b/backend/checkmach.ml
@@ -32,7 +32,7 @@ module Witness = struct
     | Alloc of
         { bytes : int;
           dbginfo : Debuginfo.alloc_dbginfo;
-          from_flat_float_array : bool;
+          from_flat_float_array : bool
         }
     | Indirect_call
     | Indirect_tailcall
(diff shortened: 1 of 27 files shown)


mshinwell:
We decided to hold off on this pending further discussions.

gretay-js:
This PR doesn't address the poor code generation problem and makes some horrible hacks in the middle end.  #2202 is better.

PR #1420, opened 2023-05-26
Local immutable arrays

antalsz:
This is the migration of ocaml-flambda/ocaml-jst#156 to flambda-backend; most of the review should therefore already be done.

PR #2084, opened 2023-11-24
Ports #1420 (local immutable array)

riaqn:
This trivially ports the runtime part of #1420. Note that some changes are already introduced by previous porting PRs.

### DIFF
--- a/ocaml/runtime/array.c
+++ b/ocaml/runtime/array.c
@@ -73,12 +73,40 @@ CAMLprim value caml_floatarray_get(value array, value index)
   return res;
 }
 
+/* [ floatarray -> int -> local_ float ] */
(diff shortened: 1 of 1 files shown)


stedolan:
There are some tests for this - do they pass now?

riaqn:
Sorry I should mention - yes all tests under `lib-array` and `typing-local` passed.

stedolan:
The patch here looks good, but there are a couple of `CR ocaml 5 runtime` at the bottom of `array.c` that should be looked at as well. (I'm not sure whether there are tests for those)

riaqn:
Those `CR ocaml 5 runtime` are taken care of.

TheNumbat (approved):

PR #2087, opened 2023-11-27
Zero alloc: change format of error messages

gretay-js:
Do not use "sub-locations". Refactor the way error messages are printed. 

Instead of printing one error message per function with a list of allocation witnesses, print each witness as a separate error message. This makes editor integeration easier, in particular, no adjustments are needed to click on the source location of witnesses.

### DIFF
--- a/backend/checkmach.ml
+++ b/backend/checkmach.ml
@@ -62,7 +62,7 @@ module Witness = struct
   let print_kind ppf kind =
     let open Format in
     match kind with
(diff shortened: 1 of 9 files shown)


gretay-js:
This is not ready yet.

gretay-js:
ready to review after a bit more testing with vscode and emacs.

xclerc (approved):

PR #2091, opened 2023-11-27
Improve layout history error message: review new messages

alanechang:
This PR removes all the `error messages below here are unreviewed` CRs. Only merge it after the new error messages are in fact reviewed.

### DIFF
--- a/ocaml/testsuite/tests/typing-immediate/immediate.ml
+++ b/ocaml/testsuite/tests/typing-immediate/immediate.ml
@@ -161,7 +161,7 @@ Line 2, characters 2-41:
 2 |   type t = Foo of int | Bar [@@immediate]
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 Error: The layout of type t is value, because
-         it's a boxed variant.
+         it's a boxed variant type.
        But the layout of type t must be a sublayout of immediate, because
          of the annotation on the declaration of the type t.
 |}];;
@@ -175,7 +175,7 @@ Line 2, characters 2-38:
 2 |   type t = { foo : int } [@@immediate]
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 Error: The layout of type t is value, because
-         it's a boxed record.
+         it's a boxed record type.
        But the layout of type t must be a sublayout of immediate, because
          of the annotation on the declaration of the type t/2.
 |}];;
--- a/ocaml/testsuite/tests/typing-layouts-err-msg/value.ml
+++ b/ocaml/testsuite/tests/typing-layouts-err-msg/value.ml
@@ -38,7 +38,7 @@ Error: Variables bound in a class must have layout value.
        The layout of v is float64, because
          it's the type of a variable bound by a `let`, defaulted to layout float64.
        But the layout of v must be a sublayout of value, because
-         it's let-bound in a class expression.
+         it's the type of a let-bound variable in a class expression.
 |}];;
 (* CR layouts v2.9: The part about defaulting here is incorrect.
    It's due to the logic in Pcl_let using sorts directly instead of
@@ -54,7 +54,7 @@ Error: Tuple element types must have layout value.
        The layout of t_any is any, because
          of the definition of t_any at line 1, characters 0-18.
        But the layout of t_any must be a sublayout of value, because
-         it's a tuple element.
+         it's the type of a tuple element.
 |}];;
 
 (* Probe *)
@@ -80,7 +80,7 @@ Error: In this `with' constraint, the new definition of t
        is not included in
          type t : immediate
        The layout of the first is value, because
-         it's used as an element in a first-class module.
+         it's a type declaration in a first-class module.
        But the layout of the first must be a sublayout of immediate, because
          of the definition of t at line 2, characters 2-20.
 |}];;
@@ -95,7 +95,7 @@ Error: The object type must have layout value.
        The layout of 'a is void, because
          of the annotation on the type variable 'a.
        But the layout of 'a must overlap with value, because
-         it's an object.
+         it's the type of an object.
 |}];;
 
 (* Instance_variable *)
@@ -113,7 +113,7 @@ Error: Variables bound in a class must have layout value.
        The layout of baz is void, because
          of the definition of t_void at line 6, characters 0-19.
        But the layout of baz must be a sublayout of value, because
-         it's an instance variable.
+         it's the type of an instance variable.
 |}];;
 
 (* Object_field *)
@@ -127,7 +127,7 @@ Error: This expression has type ('a : value)
        The layout of t_void is void, because
          of the definition of t_void at line 6, characters 0-19.
        But the layout of t_void must be a sublayout of value, because
-         it's an object field.
+         it's the type of an object field.
 |}];;
 
 (* Class_field *)
@@ -143,7 +143,7 @@ Error: Variables bound in a class must have layout value.
        The layout of bar is void, because
          of the definition of t_void at line 6, characters 0-19.
        But the layout of bar must be a sublayout of value, because
-         it's an class field.
+         it's the type of a class field.
 |}];;
 
 (* Boxed_record *)
@@ -153,7 +153,7 @@ Line 1, characters 0-26:
 1 | type r : void = {a:string}
     ^^^^^^^^^^^^^^^^^^^^^^^^^^
 Error: The layout of type r is value, because
-         it's a boxed record.
+         it's a boxed record type.
        But the layout of type r must be a sublayout of void, because
          of the annotation on the declaration of the type r.
 |}];;
@@ -165,7 +165,7 @@ Line 1, characters 0-28:
 1 | type v : void = A of t_value
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 Error: The layout of type v is value, because
-         it's a boxed variant.
+         it's a boxed variant type.
        But the layout of type v must be a sublayout of void, because
          of the annotation on the declaration of the type v.
 |}];;
@@ -177,7 +177,7 @@ Line 1, characters 0-21:
 1 | type attr : void = ..
     ^^^^^^^^^^^^^^^^^^^^^
 Error: The layout of type attr is value, because
-         it's an extensible variant.
+         it's an extensible variant type.
        But the layout of type attr must be a sublayout of void, because
          of the annotation on the declaration of the type attr.
 |}]
@@ -238,7 +238,7 @@ Line 2, characters 36-37:
 Error: This expression has type [ `A of int | `B ]
        but an expression was expected of type 'a t = ('a : void)
        The layout of [ `A of int | `B ] is value, because
-         it's a polymorphic variant.
+         it's a polymorphic variant type.
        But the layout of [ `A of int | `B ] must be a sublayout of void, because
          of the definition of t at line 1, characters 0-22.
 |}]
@@ -313,7 +313,7 @@ Error: This expression has type t_float64
        The layout of t_float64 is float64, because
          of the definition of t_float64 at line 5, characters 0-24.
        But the layout of t_float64 must be a sublayout of value, because
-         it's a field of a polymorphic variant.
+         it's the type of the field of a polymorphic variant.
 |}];;
 
 (* Default_type_layout *)
@@ -347,7 +347,7 @@ Error: This expression has type t_float64
        The layout of t_float64 is float64, because
          of the definition of t_float64 at line 5, characters 0-24.
        But the layout of t_float64 must be a sublayout of value, because
-         it's an array element.
+         it's the type of an array element.
 |}];;
 
 (* Lazy_expression *)
@@ -361,7 +361,7 @@ Error: This expression has type t_float64
        The layout of t_float64 is float64, because
          of the definition of t_float64 at line 5, characters 0-24.
        But the layout of t_float64 must be a sublayout of value, because
-         it's a lazy expression.
+         it's the type of a lazy expression.
 |}];;
 
 (* Class_argument *)
@@ -389,7 +389,7 @@ Error: This type for x is not a value type.
        The layout of type t_void is void, because
          of the definition of t_void at line 6, characters 0-19.
        But the layout of type t_void must be a sublayout of value, because
-         it's stored in a module structure.
+         it's the type of something stored in a module structure.
 |}];;
 
 (* Debug_printer_argument *)
@@ -423,7 +423,7 @@ Error: m1 must have a type of layout value because it is captured by an object.
        The layout of t_float64 is float64, because
          of the definition of t_float64 at line 5, characters 0-24.
        But the layout of t_float64 must be a sublayout of value, because
-         it's captured in an object.
+         it's the type of a variable captured in an object.
 |}];;
 
 (* Unknown *)
--- a/ocaml/testsuite/tests/typing-layouts-err-msg/value.ml
+++ b/ocaml/testsuite/tests/typing-layouts-err-msg/value.ml
@@ -364,7 +364,27 @@ Error: This expression has type t_float64
          it's the type of a lazy expression.
 |}];;
 
-(* Class_argument *)
+(* Class_type_argument *)
+module M = struct
+  type ('a : void) t
+
+  class virtual ['a] foo =
+    object
+      val virtual baz : 'a t
+    end
+end
+[%%expect{|
+Line 6, characters 24-26:
+6 |       val virtual baz : 'a t
+                            ^^
+Error: This type ('a : void) should be an instance of type ('a0 : value)
+       The layout of 'a is value, because
+         it's a type argument to a class constructor.
+       But the layout of 'a must overlap with void, because
+         of the definition of t at line 2, characters 2-20.
+|}];;
+
+(* Class_term_argument *)
 class foo (x : t_float64) =
   object end;;
 [%%expect{|
@@ -376,7 +396,7 @@ Error: This pattern matches values of type t_float64
        The layout of t_float64 is float64, because
          of the definition of t_float64 at line 5, characters 0-24.
        But the layout of t_float64 must be a sublayout of value, because
-         it's a type argument to a class constructor.
+         it's the type of a term-level argument to a class constructor.
 |}];;
 
 (* Structure_element *)
--- a/ocaml/testsuite/tests/typing-layouts-err-msg/value.ml
+++ b/ocaml/testsuite/tests/typing-layouts-err-msg/value.ml
@@ -376,7 +376,7 @@ Error: This pattern matches values of type t_float64
        The layout of t_float64 is float64, because
          of the definition of t_float64 at line 5, characters 0-24.
        But the layout of t_float64 must be a sublayout of value, because
-         it's a term-level argument to a class constructor.
+         it's a type argument to a class constructor.
 |}];;
 
 (* Structure_element *)
--- a/ocaml/testsuite/tests/typing-layouts-err-msg/value.ml
+++ b/ocaml/testsuite/tests/typing-layouts-err-msg/value.ml
@@ -299,7 +299,7 @@ Error: This expression has type ('a : value)
        The layout of 'b is void, because
          of the annotation on the type variable 'b.
        But the layout of 'b must overlap with value, because
-         it's an unannotated universal variable.
+         it is or unifies with an unannotated universal variable.
 |}];;
 
 (* Polymorphic_variant_field *)
--- a/ocaml/testsuite/tests/typing-layouts-err-msg/value.ml
+++ b/ocaml/testsuite/tests/typing-layouts-err-msg/value.ml
@@ -92,9 +92,9 @@ Line 1, characters 36-37:
 1 | let f: ('a : void) -> 'b = fun x -> x # baz
                                         ^
 Error: The object type must have layout value.
-       The layout of 'a is void, because
+       The layout of the type of this expression is void, because
          of the annotation on the type variable 'a.
-       But the layout of 'a must overlap with value, because
+       But the layout of the type of this expression must overlap with value, because
          it's the type of an object.
 |}];;
 
--- a/ocaml/testsuite/tests/typing-layouts-err-msg/value.ml
+++ b/ocaml/testsuite/tests/typing-layouts-err-msg/value.ml
@@ -385,10 +385,10 @@ module type S = sig val x : t_void end
 Line 1, characters 28-34:
 1 | module type S = sig val x : t_void end
                                 ^^^^^^
-Error: This type signature for x is not a value type.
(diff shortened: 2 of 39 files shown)


alanechang:
@goldfirere

ocaml/testsuite/tests/typing-immediate/immediate.ml
```diff
@@ -1,7 +1,6 @@
 (* TEST
    * expect
 *)
-(* CR layouts v2.9: all error messages below here are unreviewed *)
```
goldfirere, line 4: * Line 149: I would expect to see a message about the `[@@immediate]` attribute, not just a general blame on the definition of `t`. (Same problem on line 192 and 250.) In contrast, the blame on the definition is fine on line 216 because the definition is indeed well-typed.
* Line 177: Why the `/2`?
* Line 221: There is no test of the error produced here.

The rest of this file has been checked.
*

goldfirere (commented):
Took a quick look this morning on the train. I think auditing the messages will be good. I will coordinate this either tomorrow or Friday. But in the meantime, I left some comments on the one file. Maybe make a new PR just to clean up that one file?

ocaml/testsuite/tests/typing-immediate/immediate.ml
alanechang: See #2144

ocaml/testsuite/tests/typing-layouts-err-msg/value.ml
```diff
@@ -2,7 +2,6 @@
    flags = "-extension layouts_alpha"
    * expect
 *)
-(* CR layouts v2.9: all error messages below here are not reviewed for quality *)
```
ccasin, line 5: The error message beginning on line 37 reads:
```
Error: Variables bound in a class must have layout value.
       The layout of v is float64, because
         it's the type of a variable bound by a `let`, defaulted to layout float64.
       But the layout of v must be a sublayout of value, because
         it's let-bound in a class expression.
```
The reason given for `v` says that the type of a variable was "defaulted to layout float64", but this doesn't make sense - when we default, we default to value.

ccasin: Oops, I see there's a CR just below on that point.  But perhaps until we can address that CR we should pick some other word than "defaulted" for whatever is happening here.

ocaml/testsuite/tests/typing-missing-cmi/test.ml
```diff
@@ -19,5 +19,3 @@ module = "main.ml"
 ocamlc_byte_exit_status = "2"
 ******** check-ocamlc.byte-output
 *)
-
-(* CR layouts v2.9: error message here is unreviewed *)
```
ncik-roberts, line 23: `test.compilers.reference` says:

```
File "main.ml", line 1, characters 8-11:
1 | let _ = A.a = B.b
            ^^^
Error: This expression has type M.a but an expression was expected of type
         ('a : value)
       The layout of M.a is any, because
         the .cmi file for M.a is missing.
       But the layout of M.a must be a sublayout of value, because
         of layout requirements from an imported definition.
```
I don't know what "layout requirements from an imported definition" means. Is it possible to be more specific as to which definition? (Or which kind of definition?) Otherwise, I find this to be more confusing than helpful.

ocaml/testsuite/tests/typing-layouts/void_alpha.ml
```diff
@@ -2,8 +2,6 @@
    flags = "-extension layouts_alpha"
    * expect
 *)
-(* CR layouts v2.9: all error messages below here are unreviewed *)
```
alanechang, line 5: 78 - refer to `void_rec` instead of `t_void` perhaps
333- potential place for layout history

ocaml/testsuite/tests/typing-layouts-err-msg/value.ml
ccasin, line 5: On line 83, "it's used as an element in a first-class module" reads funny to me - I'm not sure what an element of a first class module is, but it sounds like something we'd say about an expression, not a type declaration.  Maybe just "because it is a type declaration in a first-class module"?

ccasin, line 5: The error for the example beginning on line 94 is confusing to me:
```ocaml
let f: ('a : void) -> 'b = fun x -> x # baz
[%%expect{|
Line 1, characters 36-37:
1 | let f: ('a : void) -> 'b = fun x -> x # baz
                                        ^
Error: Methods must have layout value.
       The layout of this expression is void, because
         of the annotation on the type variable 'a.
       But the layout of this expression must overlap with value, because
         it's an object.
|}];;
```
The confusing bit is the first line of this error: Where is there a method here that doesn't have layout value?  The error should be about the layout of the object `x`.

ocaml/testsuite/tests/typing-missing-cmi-indirections/test.ml
```diff
@@ -14,5 +14,3 @@ module = "client.ml"
 ocamlc_byte_exit_status = "2"
 ****** check-ocamlc.byte-output
 *)
-
-(* CR layouts v2.9: error message here is unreviewed *)
```
ncik-roberts, line 18: I don't like this error message because it doesn't give any suggestion of how these two comments connect:
  * "the layout of alias must be a sublayout of value"
  * "because the type argument of list has layout value"

I think this could be improved if the error message gave a location for the second bullet point, namely the RHS of `and indirect_usage = alias list`.

============

More context

`test.compilers.reference`:

```
File "client.ml", line 2, characters 0-19:
2 | and alias = missing
    ^^^^^^^^^^^^^^^^^^^
Error: 
       The layout of alias is any, because
         the .cmi file for Missing.t is missing.
       But the layout of alias must be a sublayout of value, because
         the type argument of list has layout value.
       No .cmi file found containing Missing.t.
       Hint: Adding "missing" to your dependencies might help.
```

The recursive group is:

```
type missing = Present.t
and alias = missing
and indirect_usage = alias list
```

ocaml/testsuite/tests/typing-missing-cmi-2/test.ml
```diff
@@ -13,5 +13,3 @@ module = "baz.ml"
 ocamlc_byte_exit_status = "2"
 ****** check-ocamlc.byte-output
 *)
-
-(* CR layouts v2.9: error message here is unreviewed *)
```
ncik-roberts, line 17: `ocaml/testsuite/tests/typing-missing-cmi-2/test.compilers.reference`

This one is good, and I don't have a real issue with it. A nit:

```
Error: This expression has type 'a Foo.t
       but an expression was expected of type ('b : '_representable_layout_1)
       [...]
       But the layout of 'a Foo.t must be a sublayout of '_representable_layout_1, because
         it's the type of a variable bound by a `let`.
```

The first two lines read well. This doesn't read as well: `But the layout of 'a Foo.t must be a sublayout of '_representable_layout_1`.

This might read better: `But the layout of 'a Foo.t must be representable.`

This is just a nit, I think this message is fine.

ncik-roberts (commented):
Feedback on missing cmis-related errors from review.

ocaml/testsuite/tests/typing-missing-cmi/test.ml
ncik-roberts: Also, I think this file is stale and is not checked by the test suite:

```
ocaml/testsuite/tests/typing-missing-cmi/main.ml.reference
```

ocaml/testsuite/tests/typing-layouts-err-msg/value.ml
ccasin, line 5: The error for the example on line 292 looks off:
```ocaml
let f: 'a. 'a -> ('b : void) = fun x -> x
[%%expect{|
Line 1, characters 40-41:
1 | let f: 'a. 'a -> ('b : void) = fun x -> x
                                            ^
       but an expression was expected of type ('b : void)
       The layout of 'b is void, because
         of the annotation on the type variable 'b.
       But the layout of 'b must overlap with value, because
         it's an unannotated universal variable.
|}];;
```
In particular, the last line says "it's an unannotated universal variable".  That's true about `'a`, but not about `'b`, which is the referent of the "it's".

ocaml/testsuite/tests/typing-layouts-missing-cmi/c.ml
```diff
@@ -37,8 +37,6 @@ type bar = B.b_value value_arg
 type boz = B.b_imm value_arg
 |}];;
 
-(* CR layouts v2.9: the error message below is unreviewed *)
-
```
ncik-roberts, line 41: LGTM. (Feel free to resolve.)

ocaml/testsuite/tests/typing-layouts/modules_beta.ml
```diff
@@ -2,7 +2,6 @@
    flags = "-extension layouts_beta"
    * expect
 *)
-(* CR layouts v2.9: all error messages below here are unreviewed *)
```
alanechang, line 5: 378 - priorize `it's used as an element in a first-class module.` less than other reasons?

ocaml/testsuite/tests/typing-layouts-missing-cmi/function_arg.ml
```diff
@@ -10,8 +10,6 @@ module = "function_b.ml"
 script = "rm -f function_a.cmi"
 ***** expect
 *)
-(* CR layouts v2.9: all error messages below here are unreviewed *)
```
ncik-roberts, line 13:
```
       But the layout of Function_a.t must be a sublayout of '_representable_layout_1, because
         it's used as a function argument.
```
I left the same comment elsewhere, but this might be more intelligible as `But the layout of Function_a.t must be representable, because...`

These cases read especially oddly, because `'_representable_layout_1` is only introduced in this phrase and isn't referred to elsewhere, so there's no benefit to be had in naming it.

ocaml/testsuite/tests/typing-layouts-err-msg/value.ml
ccasin, line 5: Nits regarding the last line of the error for the example beginning on line 306:
```
let f (x : t_float64) = `A x;;
[%%expect{|
Line 1, characters 27-28:
1 | let f (x : t_float64) = `A x;;
                               ^
Error: This expression has type t_float64
       but an expression was expected of type ('a : value)
       The layout of t_float64 is float64, because
         of the definition of t_float64 at line 5, characters 0-24.
       But the layout of t_float64 must be a sublayout of value, because
         it's a field of a polymorphic variant.
|}];;
```
Polymorphic variants have at most one argument (unlike normal variants), so I would say "the field" rather than "a field".  Also, maybe better to say "the type of the field" than "the field"?

ocaml/testsuite/tests/typing-layouts-missing-cmi/function_arg.ml
ncik-roberts, line 13: It's also odd that the number in the name of the _representable_layout variable keeps ticking up in this file. Usually the "scope" of such an invented variable is just for the current error message -- we don't need for it to be unique across multiple definitions.

ocaml/testsuite/tests/typing-layouts-err-msg/value.ml
ccasin, line 5: Similarly, some of the error messages say things like "it's an array element", "it's a lazy expression" or "it's a polymorphic variant".  But they are talking about types, not terms - should they say "it's the type of ..." instead?  Certainly we use this kind of shorthand all the time, so I don't feel particularly strongly.

ocaml/testsuite/tests/typing-layouts-missing-cmi/function_arg.ml
ncik-roberts, line 13:
```
[%%expect{|
Line 1, characters 28-56:
1 | let f2 : Function_b.fun_t = fun ~arg1:_ ~arg2 () -> arg2
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error: Function arguments and returns must be representable.
       The layout of Function_a.t is any, because
         the .cmi file for Function_a.t is missing.
       But the layout of Function_a.t must be a sublayout of '_representable_layout_3, because
         it's used as a function argument.
       No .cmi file found containing Function_a.t.
       Hint: Adding "function_a" to your dependencies might help.
|}]
```

In this error message and others, the combination of the location and error message is confusing.
  * The location identifies the function as a whole
  * The error message zeroes in on `Function_a.t`, as it's the type of one of the arguments to the function.

The underlined location is the ambient referent. If the main subject of the error message *isn't* the ambient referent, then I think there should be a sentence that identifies the subject.

In this case, I'd like something like this as the second sentence of the error message:

```
This function has an argument of type Function_a.t.
```

ocaml/testsuite/tests/typing-layouts-err-msg/value.ml
ccasin: And here's another case where I think there's a kind of category error:
```
1 | module type S = sig val x : t_void end
                                ^^^^^^
Error: This type signature for x is not a value type.
       The layout of x is void, because
         of the definition of t_void at line 6, characters 0-19.
       But the layout of x must be a sublayout of value, because
         it's stored in a module structure.
```
We say "the layout of `x` is void", but `x` isn't a type.  Again we use this short hand all the time, but I guess I lean in the direction that errors should be precise.

ocaml/testsuite/tests/typing-layouts-missing-cmi/function_arg.ml
ncik-roberts: (The same comment goes for `f2` and onward. `f0` and `f1` are fine as the referent is the underlined piece.)

ncik-roberts (commented):
The rest of the missing cmi error messages.

ocaml/testsuite/tests/typing-layouts/annots_beta.ml
```diff
@@ -2,7 +2,6 @@
    flags = "-extension layouts_beta"
    * expect
 *)
-(* CR layouts v2.9: all error messages below here are unreviewed *)
```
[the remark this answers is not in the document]
alanechang: Line 63: done

Line 140: added history to this error message type

Line 260: filed ticket for this

Line 284: done

Line 315: done

Line 513: histories are now correct. filed ticket about the header bit

ocaml/testsuite/tests/typing-missing-cmi-2/test.ml
alanechang: suggested change implemented

ocaml/testsuite/tests/typing-layouts-err-msg/value.ml
alanechang: Updated the first line to say "The object type must have layout value."

ocaml/testsuite/tests/typing-layouts-float64/basics_alpha.ml
```diff
@@ -2,7 +2,6 @@
    flags = "-extension layouts_alpha"
    * expect
 *)
-(* CR layouts v2.9: all error messages below here are unreviewed *)
```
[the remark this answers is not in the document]
alanechang: Line 57:  done
Line 284: done
Line 362: done
Line 571: filed ticket for this
Line 625: done
Line 636: filed ticket. this is easier to fix after merging back to main because a corner case around `Pcl_let`
Line 639: done
Several: rephrased all messages to refer to types
Line 660: same as 636
Line 752: we implemented something for type arguments of imported types, but not for imported functions. filed a ticket for this

ocaml/testsuite/tests/typing-layouts-float64/parsing.ml
```diff
@@ -2,7 +2,6 @@
    flags = "-extension layouts_alpha"
    * expect
 *)
-(* CR layouts v2.9: all error messages below here are unreviewed *)
```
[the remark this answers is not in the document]
alanechang: fixed

ocaml/testsuite/tests/typing-layouts-err-msg/value.ml
alanechang: This is changed in `main` to throw a different type of error. The incorrect part about `defaulted to layout float64` should disappear when rebased

ocaml/testsuite/tests/typing-layouts/void_alpha.ml
alanechang: filed ticket

ocaml/testsuite/tests/typing-layouts-err-msg/value.ml
alanechang: fixed

ocaml/testsuite/tests/typing-missing-cmi/test.ml
alanechang: Stale file removed. Filed ticket for the `of layout requirements from an imported definition` problem. There's a fix to have better messages for type parameters of imported types but not yet for imported functions.

ocaml/testsuite/tests/typing-layouts-err-msg/value.ml
alanechang: fixed

ocaml/testsuite/tests/typing-layouts-missing-cmi/function_arg.ml
alanechang: filed ticket

ocaml/testsuite/tests/typing-layouts-err-msg/value.ml
alanechang: I changed it to say `it is or unifies with an unannotated universal variable`. An alternative would be to produce different messages depending on the placement, but the complexity doesn't seem worth it since it's just for this one message type.

ocaml/testsuite/tests/typing-missing-cmi-indirections/test.ml
alanechang: filed ticket

ocaml/testsuite/tests/typing-layouts-missing-cmi/function_arg.ml
alanechang: filed ticket

alanechang:
I went through all the feedback addressing the ones that are straightforward to fix and filing tickets for all the rest.

It would be helpful to do a quick review of the changes in this PR. Then we can rebase and merge the `layout-histories` branch back into `main` before tackling rest of the tickets.

goldfirere:
@goldfirere please review

ocaml/testsuite/tests/typing-layouts-err-msg/value.ml
goldfirere: Fine by me.

ocaml/testsuite/tests/typing-layouts/annots_beta.ml
```diff
@@ -516,7 +516,7 @@ Error: This definition has type 'b -> 'b which is less general than
        The layout of 'a is value, because
          of the annotation on the universal variable 'a.
        But the layout of 'a must be a sublayout of immediate, because
-         of the annotation on the universal variable 'a.
+         of the definition of f_imm at line 1, characters 4-9.
```
goldfirere, line 519: This line is definitely an improvement. Good. But the first line of this error (which GitHub won't allow a comment on) is still not great, without having `'b : immediate` somewhere nearby. Looking at the open tickets around error messages, I don't think there's one for this infelicity. Could you add one?

ocaml/testsuite/tests/typing-layouts-err-msg/value.ml
```diff
@@ -377,7 +396,7 @@ Error: This pattern matches values of type t_float64
        The layout of t_float64 is float64, because
          of the definition of t_float64 at line 5, characters 0-24.
        But the layout of t_float64 must be a sublayout of value, because
-         it's a term-level argument to a class constructor.
+         it's the type of a term-level argument to a class constructor.
```
goldfirere, line 399: This is still not yet fixed.

ocaml/testsuite/tests/typing-layouts-err-msg/value.ml
```diff
@@ -91,10 +91,10 @@ let f: ('a : void) -> 'b = fun x -> x # baz
 Line 1, characters 36-37:
 1 | let f: ('a : void) -> 'b = fun x -> x # baz
                                         ^
-Error: Methods must have layout value.
-       The layout of this expression is void, because
+Error: The object type must have layout value.
```
goldfirere, line 94: I think this would be better with "Object types", not "The object type".

goldfirere (commented):
Took another quick pass here. We're getting very close!

ocaml/testsuite/tests/typing-layouts-err-msg/value.ml
alanechang: In a [previous comment](https://github.com/ocaml-flambda/flambda-backend/pull/2091#discussion_r1431890173):

> But this one really is a term-level argument. I think the problem is that Class_argument is used in too many places. Some are term-level, and some are type-level.

For the test case in question here:

```ocaml
class foo (x : t_float64) =
  object end;;
```

Should `x` be considered as a term-level argument?

ocaml/testsuite/tests/typing-layouts/annots_beta.ml
alanechang: Opened a ticket for this.

ocaml/testsuite/tests/typing-layouts-err-msg/value.ml
goldfirere: Yes -- `x` is very much a term level argument. Type arguments go before the class name (and in brackets); term arguments go afterwards.

alanechang: Then is the error message `it's the type of a term-level argument to a class constructor` here correct since it's referring to the type of `x`?

The message for type arguments has been updated so that in examples like the one below, a different message is displayed:

```ocaml
class ['a] c12_4 = object
  method x : 'a t_float64_id -> 'a t_float64_id = assert false
end;;
[%%expect{|
Line 2, characters 13-15:
2 |   method x : 'a t_float64_id -> 'a t_float64_id = assert false
                 ^^
Error: This type ('a : float64) should be an instance of type ('a0 : value)
       The layout of 'a is value, because
         it's a type argument to a class constructor.
       But the layout of 'a must overlap with float64, because
         of the definition of t_float64_id at line 2, characters 0-37.
|}];;
```

goldfirere: Maybe I just read too quickly previously? Yes this looks fine.

goldfirere (approved):

PR #2098, opened 2023-11-29
Wider scope of `in_minor_collection`

riaqn:
This fixes the interaction between #2068 and #2075 by increasing the scope of `in_minor_collection`, so now it covers a wider range, including the custom block finalizer.

Reqeust review from @stedolan

### DIFF
--- a/ocaml/testsuite/tests/regression/pr2098/stub.c
+++ b/ocaml/testsuite/tests/regression/pr2098/stub.c
@@ -1,7 +1,6 @@
-#include <caml/mlvalues.h>
+#define CAML_INTERNALS
+
 #include <caml/custom.h>
-#include <caml/alloc.h>
-#include <caml/signals.h>
 
 static void caml_test_finalize(value v)
 {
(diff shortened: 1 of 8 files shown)


riaqn:
I had to disable the regression test as it's not passing for the `closure` CI, but passing on my local machine with `--enable-middle-end=closure` - can't figure out why.

ocaml/testsuite/tests/regression/pr2098/stub.c
```diff
@@ -0,0 +1,23 @@
+#include <caml/custom.h>
+
+static void caml_test_finalize(value v)
+{
+  if (Caml_state_field(in_minor_collection))
+    caml_fatal_error("Thread switch from inside minor GC");
```
stedolan, line 6: I was expecting this to be something like:
```
caml_enter_blocking_section();
caml_leave_blocking_section();
```
so that we test the error message from the runtime, instead of duplicating it in a test. I see one of the other versions of this file worked like this. Did you work out what the flakiness was?

stedolan (approved):

PR #2099, opened 2023-11-29
Propagate never-returning functions, except when annotated with [@opaque]

Ekdohibs:
When a function never returns, this information is propagated and used, even if the function is not inlined. This behaviour can be disabled by an `[@opaque]` attribute on the function, which automatically implies `[@inline never]` and `[@local never]` as well.

### DIFF
--- a/ocaml/testsuite/tests/backtrace/pr6920_why_at.ml
+++ b/ocaml/testsuite/tests/backtrace/pr6920_why_at.ml
@@ -5,7 +5,7 @@
    exit_status = "2"
 *)
 
-let why : unit -> unit = fun () -> raise Exit [@@inline never]
+let why : unit -> unit = fun () -> raise Exit [@@inline never][@@opaque]
 let f () =
   why @@ ();
   ignore (3 + 2);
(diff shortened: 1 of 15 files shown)


ocaml/testsuite/tests/backtrace/pr6920_why_at.ml
```diff
@@ -5,7 +5,7 @@
    exit_status = "2"
 *)
 
-let why : unit -> unit = fun () -> raise Exit [@@inline never]
+let why : unit -> unit = fun () -> raise Exit [@@inline never][@@opaque]
```
lthls, line 8:
```suggestion
let why : unit -> unit = fun () -> raise Exit [@@opaque]
```

ocaml/testsuite/tests/backtrace/pr6920_why_swallow.ml
```diff
@@ -5,7 +5,7 @@
    exit_status = "2"
 *)
 
-let why : unit -> unit = fun () -> raise Exit [@@inline never]
+let why : unit -> unit = fun () -> raise Exit [@@inline never][@@opaque]
```
lthls, line 8:
```suggestion
let why : unit -> unit = fun () -> raise Exit [@@opaque]
```

lthls (approved):
Looks good.
I've left suggestions that I think would make the tests a bit cleaner (if they work as I intend them to).

mshinwell:
I haven't read the details, but LGTM in principle.  OK to merge once the suggestions addressed.

PR #376, opened 2021-11-09
Cause a C warning when CAMLreturn is missing in C stubs.

stedolan:
Not for the first time, a missing CAMLreturn in a C stub recently caused several people to spend a long time debugging. This patch tries to make that situation a bit easier to debug, by triggering an unused variable warning when it occurs. GCC error output looks like this (when I broke something in otherlibs/str):
```
In file included from strstubs.c:20:
strstubs.c: In function ‘re_replacement_text’:
../../runtime/caml/memory.h:300:7: error: unused variable ‘caml__missing_CAMLreturn’ [-Werror=unused-variable]
  300 |   int caml__missing_CAMLreturn = 0; \
      |       ^~~~~~~~~~~~~~~~~~~~~~~~
../../runtime/caml/memory.h:312:3: note: in expansion of macro ‘CAMLparam0’
  312 |   CAMLparam0 (); \
      |   ^~~~~~~~~~
strstubs.c:483:3: note: in expansion of macro ‘CAMLparam3’
  483 |   CAMLparam3(repl, groups, orig);
      |   ^~~~~~~~~~
cc1: all warnings being treated as errors
```
That's not great but beats debugging mysterious nontermination with gdb. (cc @mshinwell)

PR #2102, opened 2023-11-29
Port "Cause a C warning when CAMLreturn is missing in C stubs" to the 5 runtime

mshinwell:
Was originally #376


TheNumbat (approved):
Can confirm the change is the same as #376. (I assume we are no longer using Caml_state_field in runtime5)

PR #2103, opened 2023-11-29
Fix a couple of build errors

mshinwell:
1. caused by a conflict between the async exns and compactor patches
2. silence a warning about an unused copy of `sync_check_error`


TheNumbat (approved):

stedolan:
Using `Caml_inline` instead of `static` should silence the warning too without so much ifdeffery

PR #2104, opened 2023-11-29
Quick temporary fix for rec mod layouts segfault

ccasin:
Currently, this program segfaults:

```ocaml
module type S = sig
  val f : ('a : float64). 'a -> 'a
end

module rec M : S = M

let _ = M.f (Stdlib__Float_u.of_float 4.0)
```

(Credit to @alanechang for noticing this).

The segfault happens in the code for initializing the elements of recursive modules, which assumes the first argument to any such function like `f` will have a type with layout value.  This PR adds a layout check in the place where this initialization is set up.

We should really have a fix that allows this, and are discussing that elsewhere, but I wanted to get up this quick fix since it is a live segfault.

I'm also kind of sad about this quick fix - it seems weird to do this in transl rather than typechecking.  But I guess I actually feel that way about the recmod safety check entirely, so I'm not making it worse here.

### DIFF
--- a/ocaml/lambda/translmod.ml
+++ b/ocaml/lambda/translmod.ml
@@ -31,6 +31,7 @@ type unsafe_component =
   | Unsafe_functor
   | Unsafe_non_function
   | Unsafe_typext
+  | Unsafe_non_value_arg
 
 type unsafe_info =
   | Unsafe of { reason:unsafe_component; loc:Location.t; subid:Ident.t }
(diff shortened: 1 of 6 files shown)


goldfirere:
Doing this in transl is pretty terrible, but I completely agree that this fix doesn't make the problem worse.

I'm a little worried about the perf impact, especially with the call to `correct_levels`. My hunch is that we can avoid that 99% of the time, by having `check_type_jkind` (or one of its helpers) take an extra argument saying whether the levels need correcting. Then it corrects levels before doing expansion. But almost all the time, we won't need to expand and then we can skip ever correcting the levels. Does this sound plausible?

ccasin:
> I'm a little worried about the perf impact, especially with the call to `correct_levels`. My hunch is that we can avoid that 99% of the time, by having `check_type_jkind` (or one of its helpers) take an extra argument saying whether the levels need correcting. Then it corrects levels before doing expansion. But almost all the time, we won't need to expand and then we can skip ever correcting the levels. Does this sound plausible?

I worried about this briefly, but thought recursive modules are rare enough that it's not worth doing something more complicated.  Could be wrong about that though!  As for your idea: maybe, but how would we know that the levels might not need correcting?

Also I don't actually have an example demonstrating correct_levels is necessary here (the test suite passes without it) but my I believe it should be possible to make one.

PR #2107, opened 2023-11-30
Move layout restriction from -> to fun

alanechang:


alanechang:
This PR removes the restriction on `->` that forces both its argument and result types to have concrete layouts (representable by sort variables). Instead, it adds the restriction on function expressions that forces the arguments and return values to have concrete layouts. Result of this is that layout `any` can now occur in more places inside of type signatures.

This PR is ready for review, but it should not be merged in the current state: a follow-up PR is planned to rework how type variable initialization/defaulting works.

goldfirere:
Added a few more tests. One of them is slightly problematic, as I'll comment in my review.

ocaml/testsuite/tests/typing-layouts/basics_alpha.ml
```diff
@@ -1337,7 +1347,7 @@ let q () =
   assert false
 
 [%%expect{|
-val ( let* ) : 'a -> ('b -> t_float64) -> unit = <fun>
+val ( let* ) : 'a ('b : any). 'a -> ('b -> t_float64) -> unit = <fun>
```
goldfirere, line 1350: Please add comments toward the end of the file for the newly added tests, just to keep the test numbers lined up. No need to copy the actual tests.

ocaml/testsuite/tests/typing-layouts/basics.ml
```diff
@@ -1463,7 +1714,399 @@ Line 1, characters 28-40:
...
+       Types for method foo are incompatible
+|}]
+
+(*************************************************************)
+(* Test 39: Inference of functions that don't bind arguments *)
+
+let rec f () : 'a -> 'a = f ()
+
+[%%expect{|
+val f : unit -> 'a -> 'a = <fun>
```
goldfirere, line 2059: I expected this to have `('a : any)`. If there's some late-stage defaulting process that forces this to have `'a : value`, that's fine, but I'm skeptical.

ocaml/testsuite/tests/typing-layouts/basics.ml
```diff
@@ -1463,7 +1714,399 @@ Line 1, characters 28-40:
...
+|}]
+
+module M = struct
+  let rec f () : 'a -> 'a = f ()
+end
+
+[%%expect{|
+module M : sig val f : unit -> 'a -> 'a end
```
goldfirere, line 2067: I'm a little less bothered here; finishing off a module might reasonable default variables. But inferring `'a : any` here would be fine (and perhaps a small improvement).

ocaml/testsuite/tests/typing-layouts/basics.ml
```diff
@@ -1463,7 +1714,399 @@ Line 1, characters 28-40:
...
+|}]
+
+let rec f () : 'a -> 'a = f ()
+let g : ('a : any). unit -> 'a -> 'a = f
+
+[%%expect{|
+val f : unit -> 'a -> 'a = <fun>
+Line 2, characters 39-40:
+2 | let g : ('a : any). unit -> 'a -> 'a = f
```
goldfirere, line 2076: I'm a little sad this doesn't work. Not _very_ sad. But I hope we can get this without too much trouble.

ocaml/testsuite/tests/typing-layouts/basics.ml
```diff
@@ -1463,7 +1714,399 @@ Line 1, characters 28-40:
...
+|}]
+
+module M : sig
+  val f : ('a : any). unit -> 'a -> 'a
+end = struct
+  let rec f () : 'a -> 'a = f ()
+end
+
+[%%expect{|
+Lines 3-5, characters 6-3:
```
goldfirere, line 2090: This makes me properly sad. Again, if there's some reason we can't get this, fine. But I'd like to try.

ocaml/testsuite/tests/typing-layouts/basics.ml
```diff
@@ -1463,7 +1714,399 @@ Line 1, characters 28-40:
...
+         val f : ('a : any). unit -> 'a -> 'a
+       The type unit -> ('a -> 'a) is not compatible with the type
+         unit -> 'b -> 'b
+       'a has layout any, which is not representable.
+|}]
+
+let rec f : ('a : any). unit -> 'a -> 'a = fun () -> f ()
+
+[%%expect{|
+val f : ('a : any). unit -> 'a -> 'a = <fun>
```
goldfirere, line 2111: If _this_ didn't work, we'd have a major problem. But it does work, and it means we can actually define functions that deal with `any`. Good. (Just Resolve -- no action necessary.)

goldfirere (commented):
I've thoroughly checked the tests, including adding lots of new ones. There are still some improvements to be found in the implementation (as exposed in the tests). (Though mostly my new tests worked out well!) So I will wait to review the implementation until the tests get one more round.

ocaml/testsuite/tests/typing-layouts/basics.ml
alanechang: Fixed this. It's a single line change in `approx_type`.

ocaml/testsuite/tests/typing-layouts/basics.ml
```diff
@@ -29,31 +29,137 @@ Error: Layout void is more experimental than allowed by the enabled layouts exte
...
+
+module type S1 = sig
+  type t : any
+
+  type 'a s = 'a -> int constraint 'a = t
+
+  type q = t s
+end;;
+[%%expect{|
+module type S1 =
```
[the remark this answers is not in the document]
alanechang: Filed ticket to track this. Will work on it in a separate PR

ocaml/testsuite/tests/typing-layouts/annots.ml
```diff
@@ -242,14 +242,13 @@ val f : 'a -> 'a = <fun>
...
-Line 1, characters 8-28:
+Line 1, characters 31-41:
 1 | let f : ('a : any). 'a -> 'a = fun x -> x
-            ^^^^^^^^^^^^^^^^^^^^
-Error: The universal type variable 'a was declared to have
-       layout any, but was inferred to have a representable layout.
+                                   ^^^^^^^^^^
+Error: This definition has type 'b -> 'b which is less general than
+         ('a : any). 'a -> 'a
+       'a has layout any, which is not representable.
```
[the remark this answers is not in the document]
alanechang: Added CR. This does look like a case where layout-histories can help.

ocaml/testsuite/tests/typing-layouts/annots.ml
```diff
@@ -401,14 +400,12 @@ val f : ('a : float64). 'a -> 'a = <fun>
...
 ;;
 [%%expect {|
-Line 1, characters 24-30:
+Line 1, characters 33-43:
 1 | let f : type (a : any). a -> a = fun x -> x
-                            ^^^^^^
-Error: The universal type variable 'a was declared to have
-       layout any, but was inferred to have a representable layout.
+                                     ^^^^^^^^^^
+Error: Function arguments and returns must be representable.
```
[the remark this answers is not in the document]
alanechang: The difference seems to be caused by things getting typed in different orders. I did not see a straightforward way to get this error for the example on L242. Hopefully layout-histories is enough to improve the error message for that case. If not, we can investigate this further.

goldfirere:
Note to self: I have reviewed commits up to (and including) the one with message "fix test" (899010b).

ocaml/typing/ctype.ml
```diff
@@ -4428,7 +4391,7 @@ let rec moregen inst_nongen variance type_pairs env t1 t2 =
         moregen_occur env (get_level t1) t2;
         update_scope_for Moregen (get_scope t1) t2;
         occur_for Moregen env t1 t2;
-        constrain_type_jkind_exn env Moregen t2 jkind;
+        check_type_jkind_exn env Moregen t2 jkind;
```
goldfirere, line 4394:
```suggestion
        (* use [check], not [constrain], here because [constrain] would be like
        instantiating [t2], which we do not wish to do *)
        check_type_jkind_exn env Moregen t2 jkind;
```

goldfirere (approved):
Very nice. There were a few cases in here that took me quite a while to figure out (because the situations can be complex!) -- but you did absolutely the right thing. Nicely done.

ocaml/typing/ctype.ml
```diff
@@ -3856,24 +3842,14 @@ exception Filter_arrow_failed of filter_arrow_failure
...
-       function types that arise from inference, and the check in
-       [Typetexp.transl_type_aux] handles function types explicitly written in
-       the source program.  When we decide to drop that restriction, we can
-       allow both to be any.  Separately, the relevant checks on function
-       arguments should happen when functions are constructed, not their
-       types. *)
-    let k_arg, arg_sort = Jkind.of_new_sort_var ~why:Function_argument in
-    let k_res, ret_sort = Jkind.of_new_sort_var ~why:Function_result in
+    let k_arg = Jkind.any ~why:Inside_of_Tarrow in
+    let k_res = Jkind.any ~why:Inside_of_Tarrow in
```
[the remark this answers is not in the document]
alanechang: If we prioritize other reasons over `any` creation reasons, then this likely won't even be visible to the user. I'm leaning more towards keeping it as is for now.

PR #2109, opened 2023-11-30
Re-enable some tests that it seems should work now

mshinwell:
These were disabled, but it seems like they should work now.


goldfirere (approved):
I don't really know the context here, but enabling more (passing) tests seems like it can only be a good thing.

PR #1750, opened 2023-08-16
Unboxed ints (`int32`, `int64`, and `nativeint`)

antalsz:
This adds unboxed integer support in the same vein as unboxed float support:

* `int32#`, `int64#`, and `nativeint#` types
* `bits32`, `bits64`, and `word` layouts
* `Stdlib__Int32_u`, `Stdlib__Int64_u`, and `Stdlib__Nativeint_u` modules

There's one standing issue: unboxed integers are still sometimes allocating with the closure and flambda middle-ends, but not the flambda2 middle-end, due to some inling failure.  I'll be investigating that, but the rest should be reviewable.

Reviwers: @goldfirere or @ccasin

PR #2113, opened 2023-11-30
Unboxed ints (int32, int64, and nativeint)

alanechang:
This is #1750 rebased over the tip of main.


alanechang:
This PR should be ready for review.

The backend changes have been simplified by using the existing `typ_int` Cmm type instead of introducing new ones. New tests are added at the end of `unboxed_{bits32s,bits64,nativeint}_alpha.ml` to check for correct behavior of if-then-else expressions where one branch is `assert false` (not really sure how to best detect failure here, will a bug always crash the program?).

Conversion functions between unboxed int32, int64, and nativeint are not included in this PR. They will be worked on and added in a followup PR (tracked by a ticket).

Have not add a case match in the parser to have `int32#` and `float#` behind different maturity levels following the spirit of having less logic in the parser. Consequence would be that the type `int32#` can be accessed with `layouts` but it can't compile without `layouts_alpha` due to the checks in `typeopt`. Wonder if this is acceptable.

**How to review**

Might be helpful to look through the comments left on #1750 and check that they have been addressed here. 

Reviewers: @ccasin @goldfirere @mshinwell

goldfirere:
@ccasin, are you OK to lead the review here?

ccasin:
> @ccasin, are you OK to lead the review here?

Yes, will do

backend/cmmgen.ml
```diff
@@ -415,12 +415,16 @@ let exttype_of_sort (s : Jkind.Sort.const) =
...
+  | Word -> XInt
+  | Bits32 -> XInt32
+  | Bits64 -> XInt64
   | Void -> Misc.fatal_error "Cmmgen.exttype_of_sort: void encountered"
 
 let machtype_of_sort (s : Jkind.Sort.const) =
   match s with
   | Value -> typ_val
   | Float64 -> typ_float
+  | Word | Bits32 | Bits64 -> typ_int
```
ccasin, line 427: I can't leave a comment in the right place, but I'm wondering if, since we've decided it's fine to use `typ_int` for all these cases, we should delete `Cmm_helpers.typ_int64`.  It is only used in one place (`To_cmm_shared.machtype_of_kind`) and I think that use can now be replaced with `typ_int`.  @mshinwell?

ccasin (changes requested):
This was already reviewed fairly thoroughly in its previous incarnation, so to review this version I've taken the following steps:
- I read each of the new commits individually.
- I went back and looked through the outstanding comments on the original PR, checked they were addressed, and resolved the ones that were.

In addition to the new comments I've left here, there are two comments on the original PR that haven't been addressed and should be:
- Richard suggested adding some more conversion functions in the new stdlib modules.
- Everything added by this PR should be gated behind `layouts_beta`, not `layouts_alpha`.  We've decided `layouts_alpha` is only for highly experimental features that we want to leave in the deployed compiler for whatever reason.  When making this change, various tests added by this PR should also change to use `beta`, and the `alpha` and `beta` versions of those tests can be combined into a single file - see how this is done now for unboxed floats (though unlike unboxed floats, we won't put this in the unqualified `layouts` extension just yet, so there will still be two versions).

goldfirere:
@ccasin and I just had a chat about the merits of releasing into beta or just straight to layouts. We settled on this plan:

* This should all go into `layouts_beta`.
* After we roll the compiler, we'll need to write a little library that people can use to interact with the new types. This will have some tests.
* After that library is written and passes its testsuite, we'll have enough confidence to move to `layouts`.

Meta-point: every feature that goes into `beta` should have a criterion (preferably written down somewhere) that will cause it to go into `layouts`.

ccasin (approved):

backend/cmmgen.ml
alanechang: Latest commit in this PR removes both `Cmm_helpers.typ_int64` and `Cmm_helpers.Extended_machtype.typ_int64` as discussed in our meeting, and there seems to already be an assert in `cmm_helpers.ml` checking `size_int = 8`:

https://github.com/ocaml-flambda/flambda-backend/blob/dad7b3a289307c83943aec3a91693e6f781599ff/backend/cmm_helpers.ml#L1481

ccasin:
(This is ready to be merged, but let's wait until tomorrow to make our lives easier of minus-4 needs to be retagged)

PR #2114, opened 2023-12-01
Fix the build system when using dune 3.11 and above

alanechang:
There's an issue with the build scripts when using dune version 3.11 and above (noted by #1931). The root cause seems to be that the newer versions of dune clear the install directory before building. I made a change to copy the `runtime_stdlib` install folder to another location and that's able to fix the problem.


goldfirere:
Who would be a good reviewer here? Stabbing in the dark, but maybe @poechsel ?

goldfirere:
Maybe complain on Slack looking for a reviewer here?

alanechang:
Think there's a wish to better understand why this change happened. I did open an issue on dune https://github.com/ocaml/dune/issues/10038 but did not get any response yet

mshinwell:
I've asked @dkalinichenko-js if she could look at this.

dkalinichenko-js (approved):
LGTM. I tested that the changes work with the latest Dune (3.13.0) on my machine and updated the docs. I will create a separate PR that moves our CI to Dune 3.13.0.

PR #2116, opened 2023-12-04
Replace Closure and Flambda 1 CI jobs by a Flambda 2 + debug runtime5 CI job

mshinwell:
As subject.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,27 +13,16 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - name: closure
-            config: --enable-middle-end=closure
-            os: ubuntu-latest
-            build_ocamlparam: ''
-            check_arch: true
-            use_runtime: d
-            ocamlrunparam: "v=0,V=1"
-
-          - name: flambda1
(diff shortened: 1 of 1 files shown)


riaqn:
Now that #2118 is merged, this PR should pass CI after rebase.

.github/workflows/build.yml
```diff
@@ -13,27 +13,16 @@ jobs:
...
+          - name: flambda2_runtime5
+            config: --enable-middle-end=flambda2 --enable-runtime5
             os: ubuntu-latest
-            build_ocamlparam: ''
 
-          - name: flambda2_runtime5
+          - name: flambda2_debug_runtime5
             config: --enable-middle-end=flambda2 --enable-runtime5
             os: ubuntu-latest
+            build_ocamlparam: ''
```
riaqn, line 23: Do we need to specify this line? the macos target doesn't have this line.

riaqn (approved):
LGTM with a minor comment that can be ignored.

.github/workflows/build.yml
mshinwell: It's maybe better for clarity, I copied it from the one below.

PR #2126, opened 2023-12-05
Correct order of printed quantified type vars

ncik-roberts:
Print quantified type vars in the order they appear in the provided types, not the reverse of that order. Quantified type vars are just printed for types schemes quantified over non-value layouts.

This will improve merlin's output in [this PR](https://github.com/janestreet/merlin-jst/pull/32).


alanechang (approved):
The changes here look good!

PR #2128, opened 2023-12-06
`Cfg_dominators`: tentative fix for dead code handling

xclerc:
Pull request #1868, which was meant to support
functions whose dominators structure was a forest
rather than a tree, is incorrect and in (very)
rare cases leads to an infinite loop.

The reason why we wanted to support forests is
because we can end up with a CFG containing
dead code (*i.e.* blocks which cannot be reached
from the entry point). If all reachable blocks
on one hand, and all unreachable blocks on the
other hand live in unconnected sub-graphs, the
changes in #1868 are fine.

However, if a block has a predecessor which
is reachable and another one which is not, this
can result in a infinite loop in the `intersect`
function. This function essentially starts from
two blocks and moves "upwards" until it finds a
common ancestor. Such an ancestor is
unfortunately not guaranteed to exist in
presence of dead code.

This pull request fixes the issue by first computing
the "components" of the CFG and, by then
essentially running the main part of the algorithm
"within" components. A component is the set of
blocks which can be reached from a given block
with no predecessors. The first component is the
one whose starting block is the entry point of the
CFG.

Inside a component we know a common ancestor
exists. Working "inside" components means that
when a block would be reached from two (or more)
origins, we arbitrarily pick one. Since we start from
the entry point, we know that all reachable blocks
are in the same component. We may arbitrarily cut
unreachable portions of the graphs into components,
but that does not matter since it is about dead code.

note: the above relies on the fact that unreachable
components have a "pseudo-entry", *i.e.* a node
with no predecessors. It is currently true because
all unreachable blocks will eventually an exception
handler as their origin, but we may want to not rely
on this property. (If the property does no hold, a
fatal error will be raised.)

### DIFF
--- a/backend/cfg/cfg_dominators.ml
+++ b/backend/cfg/cfg_dominators.ml
@@ -72,35 +72,106 @@ let invariant_doms : Cfg.t -> doms -> unit =
           doms))
     doms
 
-(* CR-soon xclerc for xclerc: factor out with the function in
-   `Regalloc_ls_utils`; the only difference is the point where `f` is called. *)
-let iter_blocks_dfs : Cfg.t -> f:(Cfg.basic_block -> unit) -> unit =
- fun cfg ~f ->
-  let marked = ref Label.Set.empty in
-  let rec iter (label : Label.t) : unit =
-    if not (Label.Set.mem label !marked)
-    then (
-      marked := Label.Set.add label !marked;
-      let block = Cfg.get_block_exn cfg label in
-      Label.Set.iter
-        (fun succ_label -> iter succ_label)
-        (Cfg.successor_labels ~normal:true ~exn:true block);
-      f block)
-  in
-  iter cfg.entry_label;
-  (* note: some block may not have been seen since we currently cannot remove
-     all non-reachable blocks. *)
-  if Label.Set.cardinal !marked <> Label.Tbl.length cfg.blocks
-  then
-    Cfg.iter_blocks cfg ~f:(fun label block ->
-        if not (Label.Set.mem label !marked) then f block)
+module DFSUtils : sig
+  type component
+  (* Identifier of a "component", which is the set of blocks reachable from a
+     given (pseudo-)entry (i.e. a block with no predecessors). *)
 
-(* CR-soon xclerc for xclerc: we could compute the stack once, and use
-   `Stack.copy` when we need to iterate. *)
-let reverse_post_order : Cfg.t -> f:(Cfg.basic_block -> unit) -> unit =
- fun cfg ~f ->
+  val equal_component : component -> component -> bool
+
+  type components
+  (* Map from block labels to component identifier. *)
+
+  val find_component : components -> Label.t -> component
+  (* [find_component components label] return the component identifier for
+     [label] in [components] - cannot raise. *)
+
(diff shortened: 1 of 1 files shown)


backend/cfg/cfg_dominators.ml
```diff
@@ -72,49 +72,121 @@ let invariant_doms : Cfg.t -> doms -> unit =
...
+   fun cfg ~components ->
+    try
+      Cfg.iter_blocks cfg ~f:(fun label block ->
+          if (not (Label.Tbl.mem components label))
+             && Label.Set.is_empty block.predecessors
+          then raise (Found label));
+      Misc.fatal_error "did not find a block with no predecessors"
+    with Found label -> label
+
+  let rec iter_blocks_pseudo_entries :
```
gretay-js, line 141: This `iter_blocks_pseudo_entries` is more complicated than I expected. We can find all blocks with no predecessors  using `Cfg.iter_blocks`, and then just trace from there, using `iter_blocks_from`.  (This won't find unreachable cycles but neigher the current code does). Am I missing something?

xclerc:
> Are you referring to unreachable cycles? I don't see how they would be caught (L268 does not apply I think).

Yes, I think it is impossible to have "only" an
unreachable cycle. There will be an exception
handler, with no parents, that would lead to
such a cycle.

backend/cfg/cfg_dominators.ml
xclerc: I don't really follow; how does that differ
from what is implemented?

xclerc: (We are not finding unreachable cycles,
because the assumption is that they do
not exist. But if they do, the program will
fail with `Misc.fatal_error "did not find
a block with no predecessors"`.)

xclerc:
> Compared to adding a fake root, this does not have the property that any two nodes have a dominator in common. Do we rely on this property?

This property was already no longer true
after #1868, since we are (arguably in
corner cases) building a forest.

backend/cfg/cfg_dominators.ml
gretay-js: > I don't really follow; how does that differ from what is implemented?

current version finds pseudo_entry blocks one at a time, and iterates from it.  what I was describing is to first find them all (put in a temporary set), and then iterate from them - I think it is simpler and no need for a recursive function.

> (We are not finding unreachable cycles,
because the assumption is that they do
not exist. But if they do, the program will
fail with Misc.fatal_error "did not find a block with no predecessors".)

aha, that's what I was missing! and perhaps that's why you used this more complicated recursion, but I would still prefer the above and a check at the end if not all labels have components, then fail. i think it would work too and would be simpler to understand because of the separate concerns.

xclerc: ah, thanks for the clarification, I thought
it was more obvious correct by kind of
peeling what remains component per
component but maybe not.

gretay-js (approved):

backend/cfg/cfg_dominators.ml
mshinwell: @gretay-js Let's go with the current version for now, as this is really blocking.  If we want to restructure the recursion, then that could be considered for a future PR.

PR #2131, opened 2023-12-06
Optimize generated cmm for negated comparisons

### LINKED ISSUE #2125: poor code generation for boolean negation in conditions of if statements
We are missing some optimizations for tagging/untagging of boolean expressions.

For example,
```
let if_good (a:int) (b:int) =
  if a <> b then
    Sys.opaque_identity ()

let if_bad (a:int) (b:int) =
  if not (a = b) then
    Sys.opaque_identity ()
```
Produced this code:
```
camlD__if_good_0_4_code:
	cmpq	%rbx, %rax
	je	.L103
	movl	$1, %eax
	ret
	.align	4
.L103:
	movl	$1, %eax
        ret

camlD__if_bad_1_5_code:
	cmpq	%rbx, %rax
	setne	%al
	movzbq	%al, %rax
	salq	$1, %rax
	sarq	$1, %rax
	testq	%rax, %rax
	je	.L107
	movl	$1, %eax
	ret
	.align	4
.L107:
	movl	$1, %eax
	ret
```
when compiled with flambda2 -O3. Here is the corresponding cmm:
```
(function{d.ml:1,12-73} camlD__if_good_0_4_code (a/388: int b/389: int)
 (if (!= a/388 b/389) (opaque 1) 1))

(function{d.ml:6,11-77} camlD__if_bad_1_5_code (a/393: int b/394: int)
 (if (>>s (<< (!= a/393 b/394) 1) 1) (opaque 1) 1))
```
Both functions produced the "good" code with flambda1.

Gbury:
Closes #2125 

See the added comment for more explanation.

### DIFF
--- a/backend/cmm_helpers.ml
+++ b/backend/cmm_helpers.ml
@@ -290,6 +290,14 @@ let rec mul_int c1 c2 dbg =
     add_const (mul_int c (Cconst_int (k, dbg)) dbg) (n * k) dbg
   | c1, c2 -> Cop (Cmuli, [c1; c2], dbg)
 
+(* identify cmm operations whose result is guaranteed to be small enough that it
+   is correct to ignore their high bit. *)
+let guaranteed_to_be_small_int = function
+  | Cop ((Ccmpi _ | Ccmpf _), _, _) ->
+    (* integer/float comparisons return either [1] or [0]. *)
+    true
+  | _ -> false
+
 let ignore_low_bit_int = function
   | Cop
       ( Caddi,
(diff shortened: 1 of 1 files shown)


Gbury:
This code in this PR has been completely changed to move the point where we optimize away pairs of `untag of tag` into `cmm_helpers`, when in some very specific situations, we can know that a cmm expression will aways fit on 63 bits (or rather n-1 bits), and therefore pairs of `untag of tag` (or rather `asr of lsl`) operations can be removed.

Gbury:
@lthls this is ready for review

PR #2133, opened 2023-12-07
Add caml_hot__code{begin,end} symbols in runtime5

mshinwell:
This allows C `extern` declarations to be used to get hold of these symbols under runtime5.  The runtime5 symbols involving "`.`" characters cannot be accessed in this manner, as the full stop is illegal in C identifiers.


gretay-js (approved):
We may need to emit similar symbols for compilation unit begin/end.

mshinwell:
@gretay-js not that I've seen at present, but will bear that in mind.

PR #2134, opened 2023-12-08
Turn `Alloc` terminator into basic instruction

xclerc:
(as per title)

### DIFF
--- a/backend/cfg/cfg_equivalence.ml
+++ b/backend/cfg/cfg_equivalence.ml
@@ -286,15 +286,6 @@ let check_operation : location -> Cfg.operation -> Cfg.operation -> unit =
     ()
   | Dls_get, Dls_get -> ()
   | Poll, Poll -> ()
-  | _ -> different location "operation"
- [@@ocaml.warning "-4"]
-
-let check_prim_call_operation :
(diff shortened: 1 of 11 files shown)


backend/cfg/cfg_equivalence.ml
```diff
@@ -307,6 +298,15 @@ let check_prim_call_operation :
          && Lambda.eq_mode expected_mode result_mode ->
     (* CR xclerc for xclerc: also check debug info *)
```
gretay-js, line 299: Why is it a CR? Isn't it simply `Debuginfo.compare expected_dbginfo result_dbginfo` ?

gretay-js (approved):
Alloc annot raise after https://github.com/ocaml-flambda/flambda-backend/issues/1804 and https://github.com/ocaml-flambda/flambda-backend/pull/2007, so it can be basic instruction in Cfg. Separately, we may need to adjust FDO hanlding of execution counters for `gc_call` blocks generated for Alloc. Similarly for Poll instructions in #1927 .

backend/cfg/cfg_equivalence.ml
xclerc: (This is dead code, which has probably not bee
run for quite some time - if anything, I would
rather delete this module.)

PR #2137, opened 2023-12-08
Make local GC scanning respect header reserved bits

ncik-roberts:
Prior to this PR, the bit used by the GC when it scanned the local stack would clobber the top reserved header bits (if any bits were reserved). This PR shifts that bit down by the number of reserved header bits to avoid this clobbering.

This is anticipation of using reserved header bits for mixed blocks.


ncik-roberts:
We should not merge this. We should instead use #2141.

PR #2138, opened 2023-12-08
Take the local scanning bit into account for Max_wosize

ncik-roberts:
Halve `Max_wosize` to account for the fact that we use a bit for tracking locals/externals during GC. This ensures that (e.g.) array creation operations will raise if you try to create an array whose size would need to use this bit.


ncik-roberts:
No longer needed due to #2141

PR #2139, opened 2023-12-08
Implement unboxed int literals

alanechang:


alanechang:
This PR implements unboxed int literals in the same way as unboxed float literals (#2025, #2062, #2071).

It adds:

- new constant constructors for `Const_unboxed_int32`, `Const_unboxed_int64`, and `Const_unboxed_nativeint`
- support for these new constructors in the frontend + middle/backend
- a new `Punboxed_int_comp` lambda primitive used to compile match statements with unboxed ints

## How to review

It might be easier to review the lambda primitive change separate from the rest ([unboxed literal without the primitive](https://github.com/ocaml-flambda/flambda-backend/pull/2139/files/63b217fe23714f6e2eda5f505bed25df9fbaa586), [commit adding the new primitive](https://github.com/ocaml-flambda/flambda-backend/pull/2139/commits/5201f64e26aad1a84a8d4efdff7e3c670edbf5e2)).

ocaml/typing/printpat.ml
```diff
@@ -24,22 +24,27 @@ let is_cons = function
 | {cstr_name = "::"} -> true
 | _ -> false
 
+let fmt_unboxed_const_str s =
+  match String.split_on_char '-' s with
+  | [""; s] -> "-#" ^ s
+  | [s] -> "#" ^ s
+  | _ -> assert false
```
ncik-roberts, line 31: This is not expressly related to this PR, but I don't see analogous code in the parsetree's `pprintast.ml`. That makes me think that `-dsource` won't appropriately round trip. Light testing with the top-level seems to confirm.

Perhaps a good regression test would be to add an example to `ocaml/testsuite/tests/parsetree/source_jane_street.ml`. That's become a bit more comprehensive since our first go at unboxed literals.

ocaml/testsuite/tests/typing-layouts/literals.ml
```diff
@@ -7,150 +7,142 @@
...
 Error: Unknown modifier 'P' for literal #0.P
 |}]
 
+let () = test_int32 "unknown_floating_point_suffix" (#0g)
+[%%expect{|
+Line 1, characters 52-57:
+1 | let () = test_int32 "unknown_floating_point_suffix" (#0g)
+                                                        ^^^^^
+Error: Unknown modifier 'g' for literal #0g
+|}]
```
ncik-roberts, line 144: `unknown_int_suffix`? (I'd make a GitHub suggestion but I'm sure I'd get the column numbers wrong...)

ocaml/testsuite/tests/typing-layouts-word/alloc.ml
```diff
@@ -98,3 +99,12 @@ end
...
 let () = Collatz_boxed.go ()
+
+let[@inline never] literal_test x y =
+  let open Nativeint_u in
+  let[@inline never] f x y = (#1n + x) * (y - #4n) in
+  match x with
+  | #2n ->  (f x y) / (#3n % #10n)
+  | _ -> #0n
+
+let _ = measure_alloc "literals (should be -1): %nd" (fun () -> literal_test #2n #3n)
```
ncik-roberts, line 110: It would be good to add a test (somewhere) using a literal that's out-of-bounds of a normal int63.

ocaml/lambda/lambda.ml
```diff
@@ -217,6 +217,7 @@ type primitive =
   | Plsrbint of boxed_integer * alloc_mode
   | Pasrbint of boxed_integer * alloc_mode
   | Pbintcomp of boxed_integer * integer_comparison
+  | Punboxed_int_comp of boxed_integer * integer_comparison
```
ncik-roberts, line 220: It's a bit surprising that a constructor with "unboxed" in the name would carry a `boxed_integer` payload. Is this correct? If so, do you think it's worth commenting this?

FWIW, I'd be in favor of introducing a new type `unboxed_integer` that has the same constructors as what's currently in `boxed_integer`, if that's easy. It could even be a private alias.

ncik-roberts (commented):
I reviewed the `typing` and `lambda` directories, as well as the tests. It would be good for a middle-end person to glance at `lambda` and the tests, too.

The front-end changes broadly look good.

ocaml/lambda/lambda.ml
mshinwell: I think that already exists in the unboxed arrays PR actually.

mshinwell: I thought I'd added such a type though, maybe not...

Gbury: actually, you did (see https://github.com/ocaml-flambda/flambda-backend/pull/2185/files#diff-4e8efb33bb39fe1c7f2c33b3747200c58d90d8c8ba21956fb7ecc52403b9b2d8R329 ), but as an alias of `boxed_integer` which is a bit weird the first time one reads the type definition.

ocaml/typing/typecore.ml
```diff
@@ -643,7 +657,13 @@ let unboxed_constant : Jane_syntax.Layouts.constant -> (Typedtree.constant, erro
   = function
   | Float (f, None) -> Ok (Const_unboxed_float f)
   | Float (x, Some c) -> Error (Unknown_literal ("#" ^ x, c))
-  | Integer (_, _) -> Error Unboxed_int_literals_not_supported
+  | Integer (i, suffix) ->
+    begin match constant_integer i ~name:("#"^i) ~suffix with
```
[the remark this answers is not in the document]
alanechang: Good point! I defined a new error type and moved the `Unknown_literal` construction out of `constant_integer`

Gbury (approved):

ocaml/lambda/lambda.ml
alanechang: For compatibility with the other PR, I adopted the same approach here:

- defined `unboxed_integer` as an alias to `boxed_integer`
- and changed `Punboxed_int_comp` to take `unboxed_integer`

alanechang:
@ncik-roberts This is ready for another look. I fixed the `-dsource` round trip issue, added tests, and addressed the other comments. The new commits are the ones after the last force push.

ocaml/lambda/lambda.ml
ncik-roberts: OK. I think it would be a bit better to mint another type with the same constructors, but I'm happy to leave this as-is until after the other PR is merged as well. (This admits that we may never fix this, but I don't think the state is too bad.)

ncik-roberts (approved):
Approving front-end changes. Thanks for the fixes.

PR #2140, opened 2023-12-08
Enable warn-error by default

ncik-roberts:
Change the default for the `warn-error` configure flag:
  * Previously, it was disabled by default and had to be turned on with `--enable-warn-error`
  * Now, it is enabled by default and has to be turned off with `--disable-warn-error`

I think it's a good idea to need to fix warnings in the development process. This would have caught e.g. https://github.com/ocaml-flambda/flambda-backend/pull/2136.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -57,7 +57,7 @@ jobs:
             build_ocamlparam: _,ocamlcfg=0
 
           - name: flambda2_macos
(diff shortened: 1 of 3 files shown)


ncik-roberts:
Ping @mshinwell , hopefully this is quick as it matches what we discussed internally.

mshinwell (approved):
Sorry, missed this.

PR #2141, opened 2023-12-11
Switch to using GARBAGE during locals marking instead of the high bit

stedolan:
The `runtime5` support for marking locals currently flips the high bit of the header temporarily during marking. @lpw25 points out that we could use the GC color `GARBAGE` instead, as it is not otherwise reachable. (The GC colors change periodically in runtime5, but cannot change during scanning, so this usage is safe).


lpw25 (approved):
LGTM

PR #2142, opened 2023-12-11
Fix simplification of never_return applications

Ekdohibs:
This case was not possible before, and therefore unimplemented.

### DIFF
--- a/middle_end/flambda2/simplify/expr_builder.ml
+++ b/middle_end/flambda2/simplify/expr_builder.ml
@@ -830,9 +830,14 @@ let rewrite_fixed_arity_apply uacc ~use_id arity apply =
     in
     uacc, RE.create_apply (UA.are_rebuilding_terms uacc) apply
   in
-  match Apply.continuation apply with
-  | Never_returns -> make_apply apply
-  | Return cont ->
+  match use_id, Apply.continuation apply with
+  | _, Never_returns -> make_apply apply
+  | None, Return _ ->
+    Misc.fatal_errorf
+      "Expr_builder.rewrite_fixed_arity_apply: got no use_id but the apply \
+       could return:@ %a@."
+      Apply.print apply
+  | Some use_id, Return cont ->
     rewrite_fixed_arity_continuation uacc cont ~use_id arity
       ~around:(fun uacc return_cont ->
         let exn_cont =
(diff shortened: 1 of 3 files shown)


mshinwell:
@Gbury will review

middle_end/flambda2/simplify/simplify_apply_expr.ml
```diff
@@ -854,19 +854,21 @@ let rebuild_function_call_where_callee's_type_unavailable apply call_kind
...
-    EB.rewrite_fixed_arity_apply uacc ~use_id (Apply.return_arity apply) apply
+    match use_id with
+    | Some use_id ->
+      EB.rewrite_fixed_arity_apply uacc ~use_id (Apply.return_arity apply) apply
+    | None ->
+      let uacc =
+        UA.add_free_names uacc (Apply.free_names apply)
+        |> UA.notify_added ~code_size:(Code_size.apply apply)
+      in
+      uacc, RE.create_apply (UA.are_rebuilding_terms uacc) apply
```
Gbury, line 865: `Expr_builder.rewrite_fixed_arity_apply` already handles non-returning calls (by doing exactly what this code does), so I'd prefer if we could avoid the code deduplication: why not change `Expr_builder.rewrite_fixed_arity_apply` to take a optional `use_id` as argument ?

Gbury (approved):
The code looks correct.

It'd be better if we could avoid the code duplication between `EB.rewrite_fixed_arity_apply` and `rebuild_function_call_where_callee's_type_unavailable` though...

middle_end/flambda2/simplify/expr_builder.ml
```diff
@@ -830,9 +830,14 @@ let rewrite_fixed_arity_apply uacc ~use_id arity apply =
...
     uacc, RE.create_apply (UA.are_rebuilding_terms uacc) apply
   in
-  match Apply.continuation apply with
-  | Never_returns -> make_apply apply
-  | Return cont ->
+  match use_id, Apply.continuation apply with
+  | _, Never_returns -> make_apply apply
+  | None, Return _ ->
+    Misc.fatal_errorf
+      "Expr_builder.rewrite_fixed_arity_apply: got no use_id but the apply \
```
Gbury, line 837:
```suggestion
      "Expr_builder.rewrite_fixed_arity_apply: got no use_id for the return continuation but the apply \
```

PR #2144, opened 2023-12-12
Layout history: mark `typing-immediate/immediate.ml` as reviewed

alanechang:
Review: @goldfirere

### DIFF
--- a/ocaml/testsuite/tests/typing-immediate/immediate.ml
+++ b/ocaml/testsuite/tests/typing-immediate/immediate.ml
@@ -1,7 +1,6 @@
 (* TEST
    * expect
 *)
-(* CR layouts v2.9: all error messages below here are unreviewed *)
 
 module type S = sig type t [@@immediate] end;;
 module F (M : S) : S = M;;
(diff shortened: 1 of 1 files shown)


ocaml/testsuite/tests/typing-immediate/immediate.ml
```diff
@@ -149,6 +148,9 @@ Error: The layout of type string is value, because
        But the layout of type string must be a sublayout of immediate, because
          of the definition of t at line 2, characters 2-31.
 |}];;
+(* CR layouts v2.9: The "of the definition of t ..." part is not great and it
```
goldfirere, line 151: Are there also internal tickets about these? Just leaving CRs makes it possible we'll forget.

goldfirere (approved):

ocaml/testsuite/tests/typing-immediate/immediate.ml
alanechang: Made tickets to track both CRs

PR #2145, opened 2023-12-12
Fix ast iteration/mapping for layout type declarations

ncik-roberts:
Demonstrate and fix a bug where type declarations with layouts were causing the compiler to flag warning 53. That's because AST mapping/iteration had a bug: Jane Syntax attributes were being directly exposed to the `attribute` mapper/iterator rather than being hidden.

This PR should be reviewed commit-by-commit. Both commits pass all tests.
  * The first commit fixes a small pretty-printing bug, adds a regression test that runs a no-op ppx driver, and commits the result of the failing test.
  * The second test fixes the mapping/iteration issue, thus fixing the regression test from the previous commit.

How can we prevent this in the future? We should remember to add new syntactical features to this test that exercises them.

### DIFF
--- a/ocaml/parsing/ast_iterator.ml
+++ b/ocaml/parsing/ast_iterator.ml
@@ -178,12 +178,19 @@ module T = struct
     | Ptyp_extension x -> sub.extension sub x
 
   let iter_type_declaration sub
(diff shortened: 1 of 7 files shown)


ccasin (approved):
LGTM

PR #2146, opened 2023-12-12
API for constructing a labeled tuple AST is more permissive

ncik-roberts:
Previously, it was an error to use the Jane Syntax API to create a "labeled" tuple expression/pattern/type where all labels were actually None. Now, the Jane Syntax API will create a normal, non-Jane-Syntax-embedded tuple in this case, only using the embedding for labeled tuples if any label is present.

This simplifies the callers of this Jane Syntax API. Previously, they'd have to do the check. (And, if they got it wrong, there would be no immediate feedback. If you're lucky, you're a PPX and the error would be caught by the AST invariant check.)

This was motivated by needing to write this check many times in merlin. I'll import this PR into merlin once this is reviewed.

Implementation notes:
  * I kill the constructor wrappers `Lttyp_tuple`, etc. I claim this makes it a little more sensible for (e.g.) `Jane_syntax.Labeled_tuples.typ_of` to not actually do Jane Syntax embedding. (Also, I don't see a benefit to the indirection -- if there are more labeled-tuples related constructs later, we can do this refactoring then.)
  * I get rid of some tests for AST invariant checks, as there's no longer a way to construct a label-less labeled tuple. (That seems like a good thing, though.) I keep the checks, though!


ccasin (approved):
A clear improvement, thanks.

Several other places in jane syntax follow this pattern of having a single somewhat unnecessary constructor (e.g., immutable array expressions/patterns, and layout patterns).  I think they could similarly be removed, but don't think this PR needs to do so.

PR #2150, opened 2023-12-13
Add `ocamlyacc` as dependency of `.mly` -> `.ml` pattern rule

lukemaurer:
This missing dependency breaks `make depend`, which depends on `lex/parser.ml` but not the `ocamlyacc` that is needed to actually build it.


lthls:
If it works, we should consider sending a PR upstream too (I've been bit a few times when dealing with conflicts in `.depend`).

PR #2151, opened 2023-12-13
zero_alloc: relaxed interpretation of [@zero_alloc assume]

gretay-js:
Interpretation of `assume` should be `relaxed` to match the "assert" annotations without explicit `strict`.  

Currently, `assume` is always interpreted as `strict` when it is propagated on expressions in `lambda` (for inlining). This can cause zero alloc check to fail when a callee annotated with `zero_alloc assume` stops being inlined. For example, 
```ocaml
let[@zero_alloc assume] bar x =
  if x > 0 then failwith (Printf.sprintf "BOO %d!" x);
  (x+1,x)
  
let[@zero_alloc strict] foo x = bar x
```
the check of `foo` passes when `bar`  is inlined but fails when `bar` is not inlined, even though inlining doesn't cause allocations to be comletely optimized away. This happens because `assume` is propagate incorrectly. This PR fixes the bug.

A follow-up PR will also propagate `strict` and `never_returns_normally`, in preparation for assume on calls.

### DIFF
--- a/backend/checkmach.ml
+++ b/backend/checkmach.ml
@@ -251,7 +251,7 @@ module Value : sig
 
   val safe : t
 
-  val relaxed : t
+  val relaxed : Witnesses.t -> t
 
   val print : witnesses:bool -> Format.formatter -> t -> unit
 
@@ -316,8 +316,8 @@ end = struct
 
   let top w = { nor = V.Top w; exn = V.Top w; div = V.Top w }
 
-  let relaxed =
-    { nor = V.Safe; exn = V.Top Witnesses.empty; div = V.Top Witnesses.empty }
+  let relaxed w =
+    { nor = V.Safe; exn = V.Top w; div = V.Top w }
 
   let print ~witnesses ppf { nor; exn; div } =
     let pp = V.print ~witnesses in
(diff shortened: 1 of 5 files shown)


backend/checkmach.ml
```diff
@@ -1077,17 +1076,16 @@ end = struct
       transform_top t ~next ~exn w ("external call to " ^ func) dbg
     | Ispecific s ->
       let effect =
+        let w = create_witnesses t Arch_specific dbg in
         if Debuginfo.assume_zero_alloc dbg
         then
           (* Conservatively assume that operation can return normally. *)
           let nor = V.Safe in
-          let exn = if Arch.operation_can_raise s then V.Safe else V.Bot in
+          let exn = if Arch.operation_can_raise s then V.Top w else V.Bot in
```
xclerc, line 1084: I am not sure I understand why this
change is needed; could you elaborate?

gretay-js: The interpretation of "assume" should be "relaxed", which means `exn` component is allowed to be `Top`.

xclerc (approved):

PR #2152, opened 2023-12-13
Zero alloc fix assume payload parsing

gretay-js:
The following annotation is incorrectly rejected with a misplaced attribute warning:
`[@zero_alloc assume never_returns_normally strict]`. 

This PR fixes it. 

The list of permitted payload keywords passed to `Translattribute.parse_ids_payload` must be sorted alphabetically.

### DIFF
--- a/ocaml/lambda/translattribute.ml
+++ b/ocaml/lambda/translattribute.ml
@@ -136,6 +136,8 @@ let get_ids_from_exp exp =
   |> Result.map List.rev
 
 
+(* [parse_ids_payload] requires that each element in [cases]
+   the first component (string list) is alphabetically sorted. *)
 let parse_ids_payload txt loc ~default ~empty cases payload =
   let[@local] warn () =
     let ( %> ) f g x = g (f x) in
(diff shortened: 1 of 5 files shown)


ccasin:
I think we should do something to avoid this kind of error in the future.  At a minimum `parse_ids_payload` should get a comment documenting this invariant, but alternatively:
- `parse_ids_payload` could sort the lists in `cases` itself
- `parse_ids_payload` could be rewritten to eliminate this string list equality completely.  For example it could parse the individual words into a record of boolean fields indicating which are present, which could then be interpreted and checked for validity by a stand-alone function.

gretay-js:
> I think we should do something to avoid this kind of error in the future. At a minimum parse_ids_payload should get a comment documenting this invariant [...]

@ccasin 
I've added a comment for now. I considered sorting the cases in [parse_ids_payload] but it seems wasteful to sort  them over and over again, every time `parse_ids_payload` is called (even though the list is short). Currently, `parse_property_attribute` is the only caller of `parse_ids_payload` and this code may change for `zero_alloc` on signatures, so it doesn't seem worth changing at the moment.

ccasin (approved):

PR #2148, opened 2023-12-13
zero_alloc: refactor Translattribute.assume_zero_alloc slightly

gretay-js:
No functionality change. Refactoring in preparation for `assume` on calls. Add explict [with_warnigns] argument.

PR #2153, opened 2023-12-13
Zero alloc misplaced attr warning

gretay-js:
~On top of #2148.~

The following examples should trigger a "misplaced attribute" warning: 
```ocaml
let[@zero_alloc assume] foo =  let x = 42 in fun z -> z + x
let[@zero_alloc] bar = let x = 42 in fun z -> z + x
```

@ccasin noticed that they are do not. Looks like this used to work in the past but stop working after #1714 when we introduced "look-ahead" for assume annotations. This PR fixes it. I expect it to be a temporary fix, until support for `zero_alloc` on interfaces is available.

### DIFF
--- a/ocaml/lambda/translattribute.ml
+++ b/ocaml/lambda/translattribute.ml
@@ -54,9 +54,9 @@ let is_opaque_attribute =
   [ ["opaque"; "ocaml.opaque"], true ]
 
 
-let find_attribute p attributes =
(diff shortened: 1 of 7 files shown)


ccasin (approved):
LGTM

I am always slightly sad to see an optional argument added.  But I think this will go away once we do the rework for zero_alloc in signatures, so it seems fine as a temporary solution to the misplaced attributes problem.

PR #2154, opened 2023-12-13
Fix segfault when signal handlers invoked in non-OCaml threads on runtime5

mshinwell:
At present, `caml_record_signal` (which is the genuine C signal handler installed by the OCaml runtime) attempts to access a TLS variable to get the current domain's info structure.  This will segfault if the signal handler is not running on an OCaml thread.

Upstream [PR11307](https://github.com/ocaml/ocaml/pull/11307) fixes this as part of a long sequence of commits improving various matters (which is also split across PRs).  This PR ports the minimal change to just fix this bug, as discussed with @stedolan .

I have added a test which segfaults prior to this PR.  The remainder of this PR is verbatim cherry-picked (rev `59029b99420f0f871f56cccc63fd6c29e146e91b` from upstream).

### DIFF
--- a/ocaml/runtime/caml/domain.h
+++ b/ocaml/runtime/caml/domain.h
@@ -66,6 +66,7 @@ void caml_handle_gc_interrupt(void);
 void caml_handle_incoming_interrupts(void);
 
 CAMLextern void caml_interrupt_self(void);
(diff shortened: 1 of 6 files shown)


TheNumbat (approved):
Diff is the same, fix makes sense

PR #2155, opened 2023-12-13
Issue misplaced attribute warning earlier

gretay-js:
Call `Builtin_attributes.warn_unused` earlier, before `Simplify_lambda`, instead of calling it after code generation.  This is sound and accurate because all attributes should have been registered during Parsing and consumed before or during the Translation to Lambda (i.e., there are no calls to [Builtin_attributes.mark_*] and [Builtin_attributes.filter*] etc after Translmod).

Split out [Builtin_attributes.warn_unchecked_property] to be called after backend.

The motivation for this change is to report misplaced attribute warnings earlier, even if there are user-facing compilation errors in the backend during zero_alloc check (i.e., the check fails, so compilation fails).

Previously, there was no need to report warnings for misplaced attributes if there were any compilation errors, because there were no user facing checks that depended on attributes.

With zero_alloc check, it is possible to get errors during code generation that would prevent the misplaced attribute warnings from being printed. At the same time, zero_alloc check is controlled using "zero_alloc" attributes. This makes it hard for users to understand zero alloc error messages caused by misplaced attributes, as for example in the following case:

```ocaml
let[@inline never] bar x = (x,x)
let[@zero_alloc] foo x =  ((bar x)[@zero_alloc assume])
```
Here, `@zero_alloc assume` is not consumed, it's only allowed only on function definition at the moment.

### DIFF
--- a/driver/optcompile.ml
+++ b/driver/optcompile.ml
@@ -39,6 +39,7 @@ let compile i typed ~transl_style ~unix ~pipeline =
   |> Compiler_hooks.execute_and_pipe Compiler_hooks.Raw_lambda
   |> Profile.(record generate)
    (fun program ->
+      Builtin_attributes.warn_unused ();
       let code = Simplif.simplify_lambda program.Lambda.code in
       { program with Lambda.code }
       |> print_if i.ppf_dump Clflags.dump_lambda Printlambda.program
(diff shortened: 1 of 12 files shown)


ccasin (approved):
LGTM

Is it the case that we expect `test_assume_on_call.ml` to compile without error after the support for zero alloc on applications is in?  Maybe leave a comment in it to that effect, if so.

gretay-js:
> Is it the case that we expect test_assume_on_call.ml to compile without error after the support for zero alloc on applications is in? Maybe leave a comment in it to that effect, if so.

yes. The PR is coming up in a moment. I've added a comment anyway, as the PR may take a while to review.

PR #2160, opened 2023-12-13
Reinitialize IO mutexes in systhreads5

TheNumbat:
See 5: https://github.com/ocaml/ocaml/blob/trunk/otherlibs/systhreads/st_stubs.c#L410 
See 4: https://github.com/ocaml-flambda/flambda-backend/blob/main/ocaml/otherlibs/systhreads4/st_stubs.c#L524

Systhreads4 destroys all channel mutexes upon a fork, since other threads may have locked them at the time of the fork. 
We also need to do this in 5, though support for forking with threads is only best-effort and applications really shouldn't do it.

Also, I noticed that `caml_seek_in` and `caml_ml_close_channel` in `io.c` both call `caml_enter_blocking_section_no_pending` without first calling `check_pending`. Is this wrong? Once we enter the blocking section, another thread can start running and process pending signals while this channel is locked. But the point of `check_pending` is to ensure that the channel is not locked while signal handlers/finalizers run.

### DIFF
--- a/ocaml/otherlibs/systhreads/st_stubs.c
+++ b/ocaml/otherlibs/systhreads/st_stubs.c
@@ -463,6 +463,7 @@ static void caml_thread_remove_and_free(caml_thread_t th)
 
 static void caml_thread_reinitialize(void)
 {
+  struct channel * chan;
   caml_thread_t th, next;
 
   th = Active_thread->next;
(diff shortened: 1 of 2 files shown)


mshinwell (approved):
The change looks ok.  I think you're probably right about `check_pending` but I'm not certain I understand all the intricacies of this code - let's discuss elsewhere.

PR #2162, opened 2023-12-14
Fix split_default_wrapper when default value could allocate in region

Ekdohibs:


mshinwell (approved):

mshinwell:
Prior to this PR, functions with local allocations in the default values of optional arguments could cause a compile-time failure in flambda2.

PR #2163, opened 2023-12-14
Port `Comballoc` to work on CFG values

xclerc:
This pull request, as the title suggests,
is merely a port of the `Comballoc`
optimization pass to CFG values.

However, the code is significantly
different, taking advantage of the
representation of basic blocks and
the ability to change / insert
instructions quite easily.

Since this code has a lot of potential
in terms of miscompilation, we should
not merge it before further testing.

### DIFF
--- a/backend/asmgen.ml
+++ b/backend/asmgen.ml
@@ -293,9 +293,10 @@ let compile_fundecl ~ppf_dump ~funcnames fd_cmm =
       | true ->
         (* Will happen after `Cfgize`. *)
         (match register_allocator fd with
-         | Upstream -> ()
+         | Upstream ->
+           fatal_error "-cfg-cse-optimize should only be used with a CFG register allocator"
          | GI | IRC | LS ->
-           fatal_error "-cfg-cse-optimize should only be used with a CFG register allocator");
+           ());
         fd)
   ++ Profile.record ~accumulate:true "regalloc" (fun (fd : Mach.fundecl) ->
     match register_allocator fd with
--- a/backend/asmgen.ml
+++ b/backend/asmgen.ml
@@ -279,7 +279,14 @@ let compile_fundecl ~ppf_dump ~funcnames fd_cmm =
   ++ Compiler_hooks.execute_and_pipe Compiler_hooks.Mach_polling
   ++ Profile.record ~accumulate:true "checkmach"
        (Checkmach.fundecl ~future_funcnames:funcnames ppf_dump)
-  ++ Profile.record ~accumulate:true "comballoc" Comballoc.fundecl
+  ++ pass_dump_if ppf_dump dump_combine "Before allocation combining"
+  ++ (fun fd ->
+      match register_allocator fd with
+      | Upstream ->
+        Profile.record ~accumulate:true "comballoc" Comballoc.fundecl fd
+      | GI | IRC | LS ->
+        (* Will happen after `Cfgize`. *)
+        fd)
   ++ Compiler_hooks.execute_and_pipe Compiler_hooks.Mach_combine
   ++ pass_dump_if ppf_dump dump_combine "After allocation combining"
   ++  (fun fd ->
@@ -299,6 +306,7 @@ let compile_fundecl ~ppf_dump ~funcnames fd_cmm =
         let cfg =
           fd
           ++ Profile.record ~accumulate:true "cfgize" cfgize
+          ++ Profile.record ~accumulate:true "cfg_comballoc" Cfg_comballoc.run
           ++ (fun cfg_with_layout ->
               match !Flambda_backend_flags.cfg_cse_optimize with
               | false -> cfg_with_layout
--- a/backend/asmgen.ml
+++ b/backend/asmgen.ml
@@ -308,7 +308,7 @@ let compile_fundecl ~ppf_dump ~funcnames fd_cmm =
                 cfg_with_layout
                 ++ Profile.record ~accumulate:true "cfg_comballoc" Cfg_comballoc.run
                 ++ Compiler_hooks.execute_and_pipe Compiler_hooks.Cfg_combine
-                ++ Profile.record ~accumulate:true "cse" CSE.cfg_with_layout
+                ++ Profile.record ~accumulate:true "cfg_cse" CSE.cfg_with_layout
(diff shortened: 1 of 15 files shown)


gretay-js (approved):
I am slightly concerned about modying instruction description in place without chaning instruction ids. It is currently safe (e.g., liveness is not yet precomputed, we only replace Alloc with a pure instruction not introducing enew ffects), but may lead to tricky bugs due to e.g., caching based on instruction ids or instructions themselves. Do we do it in other places?

Possible future improvement: when allocations of different modes are interleaved within a block, they can be combined into just two allocations (one of each mode), for example: `H1;L1;H2;L2` would be combined into `H12;L12`.

backend/asmgen.ml
```diff
@@ -279,18 +279,20 @@ let compile_fundecl ~ppf_dump ~funcnames fd_cmm =
...
+        ++ Profile.record ~accumulate:true "comballoc" Comballoc.fundecl
+        ++ Compiler_hooks.execute_and_pipe Compiler_hooks.Mach_combine
+        ++ pass_dump_if ppf_dump dump_combine "After allocation combining"
         ++ Profile.record ~accumulate:true "cse" CSE.fundecl
         ++ Compiler_hooks.execute_and_pipe Compiler_hooks.Mach_cse
         ++ pass_dump_if ppf_dump dump_cse "After CSE"
-     | true ->
-       fd)
+      | true ->
+        (* Will happen after `Cfgize`. *)
```
gretay-js, line 294: Is there a check that if `cfg_cse_optimize` is `true` than `regalloc` is not `Upstream`?

xclerc: I was indeed wondering whether we wanted
to error out in such a case.

backend/cfg/cfg_comballoc.ml
```diff
@@ -0,0 +1,197 @@
...
+(* [find_compatible_allocations cell ~curr_mode ~curr_size] returns the
+   allocations compatible with mode [curr_mode] and total size [curr_size]. *)
+let find_compatible_allocations :
+    cell option ->
+    curr_mode:Lambda.locality_mode ->
+    curr_size:int ->
+    compatible_allocations =
+ fun cell ~curr_mode ~curr_size ->
+  let rec loop (allocations : allocation list) (cell : cell option)
+      ~(curr_mode : Lambda.alloc_mode) ~(curr_size : int) :
```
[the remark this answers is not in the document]
xclerc: Yes (and it felt a bit clearer to treat the
mode and size the same), although I
am certainly not making consistent
choices in this are over the pull requests.

backend/cfg/cfg_comballoc.ml
```diff
@@ -0,0 +1,197 @@
...
+        | Lambda.Alloc_local ->
+          { allocations = List.rev allocations; next_cell = Some cell }
+        | Lambda.Alloc_heap ->
+          loop allocations (DLL.next cell) ~curr_mode ~curr_size)
+      | Op (Intop (Icheckbound | Icheckalign _))
+      | Op (Intop_imm ((Icheckbound | Icheckalign _), _))
+      | Op Poll ->
+        { allocations = List.rev allocations; next_cell = Some cell }
+      | Reloadretaddr | Poptrap | Prologue | Pushtrap _ ->
+        (* CR-soon xclerc for xclerc: is it too conservative? *)
```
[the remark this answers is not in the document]
xclerc: I have left a note about it in the CR, and will
perform further testing once merged (as
you wrote, correctness is not at stake), in
particular to see whether that makes us
miss a large number of "combinations".

xclerc:
I have tested the code a bit more, and
on the stdlib both pipelines do exactly
the same work. I will try with larger
code bases.

backend/asmgen.ml
gretay-js: I don't see any reason not to, unless you want this combination for testing/benchmarking for now... though I can't see how it would be useful even for that.

backend/cfg/cfg_comballoc.ml
```diff
@@ -0,0 +1,197 @@
...
+    (match compatible_allocs.allocations with
+    | [] -> ()
+    | other_allocations ->
+      let first_allocation_instr = DLL.value cell in
+      let first_allocation_res0 = first_allocation_instr.res.(0) in
+      (* First, replace the "other" allocations with a reference to the result
+         of the previous allocation and compute the total size. *)
+      let total_size_of_other_allocations, dbginfo_of_other_allocations, _ =
+        List.fold_left other_allocations ~init:(0, [], first_allocation_res0)
+          ~f:(fun (size, dbginfos, prev_res0) other_allocation ->
```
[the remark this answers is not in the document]
xclerc: I think I prefer the current version, separating
the selection of the allocation from their
combination. As you wrote, the slight inefficiency
does not matter in practice since these list are
very small.

backend/asmgen.ml
xclerc: That provides a way to disable the
optimization on the mach pipeline,
so that can be used to get some
kind of baseline but indeed outside
of debugging it does not feel
particularly useful. Updated to
error out.

backend/comballoc.ml
```diff
@@ -28,7 +28,32 @@ type allocation_state =
...
+    match !Flambda_backend_flags.dump_comballoc_path with
+    | "" -> None
+    | dir ->
+      let file = Sys.argv |> Array.to_list |> String.concat " " |> Digest.string |> Digest.to_hex in
+      let path = Filename.concat dir file in
+      let chan = open_out path in
+      let ppf = Format.formatter_of_out_channel chan in
+      at_exit (fun () -> close_out_noerr chan);
+      Some ppf)
+
```
gretay-js, line 41: We can use the existing `ppf_dump` in `Asmgen` and just pass it to `Comballoc`.  I don't think we need the new flag `-dcomballoc-path`. The printing can be enabled using the existing `-dcombine` flag, it will just print a few extra lines. I don't think anything relies on the format of the output of this command.  To print to a file, we already have `-dump-into-file` flag that would print to `foo.cmx.dump` for example when the compilation command had `-o foo.cmx`

gretay-js (approved):

PR #2169, opened 2023-12-14
Improve codegen for int intrinsics

TheNumbat:
The x86 backend now emits `inc`/`dec` even if the speed-over-size flag is enabled (these instructions are not actually slower, and they are smaller). This improves the popcnt intrinsic.

Additionally, there is now a BMI architecture flag. When it is set, the count leading/trailing zeroes intrinsics generate `lzcnt` and `tzcnt` instead of sequences using `bsr`/`bsf`. 

Added tests for ctz/clz/popcnt for int/int32/int64 in `tests/simd/scalar_ops.ml`.

### DIFF
--- a/backend/amd64/arch.ml
+++ b/backend/amd64/arch.ml
@@ -37,6 +37,9 @@ let sse42_support = ref true
 (* Carry-less multiplication (Westmere+) *)
 let clmul_support = ref true
 
+(* Bit manipulation (Haswell+) *)
+let bmi_support = ref true
+
 (* Bit manipulation 2 (Haswell+) *)
 let bmi2_support = ref true
 
@@ -87,6 +90,10 @@ let command_line_options =
       " Enable CLMUL intrinsics (default)";
     "-fno-clmul", Arg.Clear clmul_support,
       " Disable CLMUL intrinsics";
+    "-fbmi", Arg.Set bmi_support,
+      " Enable BMI intrinsics (default)";
+    "-fno-bmi", Arg.Clear bmi_support,
+      " Disable BMI intrinsics";
     "-fbmi2", Arg.Set bmi2_support,
       " Enable BMI2 intrinsics (default)";
     "-fno-bmi2", Arg.Clear bmi2_support,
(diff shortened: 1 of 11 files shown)


TheNumbat:
Since `tzcnt`/`lzcnt` are interpreted as `bsr`/`bsf` on unsupported processors (instead of faulting), we'll probably want to add a runtime check for support when bmi_support is enabled. I'll work on adding this for all of the isa extensions in another PR.

backend/amd64/emit.mlp
```diff
@@ -1481,9 +1481,9 @@ let emit_instr fallthrough i =
       instr_for_intop op (arg i 1) (res i 0)
   | Lop(Iintop_imm(Iadd, n)) when i.arg.(0).loc <> i.res.(0).loc ->
       I.lea (mem64 NONE n (arg64 i 0)) (res i 0)
-  | Lop(Iintop_imm(Iadd, 1) | Iintop_imm(Isub, -1)) when not !fastcode_flag ->
+  | Lop(Iintop_imm(Iadd, 1) | Iintop_imm(Isub, -1)) ->
       I.inc (res i 0)
-  | Lop(Iintop_imm(Iadd, -1) | Iintop_imm(Isub, 1)) when not !fastcode_flag ->
+  | Lop(Iintop_imm(Iadd, -1) | Iintop_imm(Isub, 1)) ->
```
gretay-js, line 1486: I think the change to inc/dec is independent of lzcnt/tzcnt and if so - it would be better to have a separate PR for it.

TheNumbat: #2211

backend/amd64/arch.ml
```diff
@@ -37,6 +37,9 @@ let sse42_support = ref true
 (* Carry-less multiplication (Westmere+) *)
 let clmul_support = ref true
 
+(* Bit manipulation (Haswell+) *)
```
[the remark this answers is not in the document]
TheNumbat: Added, and will work on a PR adding an architecture check to the runtime.

gretay-js (approved):

PR #2171, opened 2023-12-15
Remove Cregion, Ctail and Regular Ctrywith handlers

mshinwell:
This removes complexity from the Cmm language (in `backend/` only, not `ocaml/asmcomp/`) which is no longer needed after the removal of Closure and Flambda 1:
- `Cregion` expressions
- `Ctail` expressions
- The `Regular` case of `Ctrywith` handlers.

### DIFF
--- a/backend/cmm.ml
+++ b/backend/cmm.ml
@@ -172,7 +172,6 @@ type trap_action =
   | Pop of pop_action
 
 type trywith_kind =
-  | Regular
   | Delayed of trywith_shared_label
 
 type bswap_bitwidth = Sixteen | Thirtytwo | Sixtyfour
@@ -297,8 +296,6 @@ type expression =
   | Cexit of exit_label * expression list * trap_action list
   | Ctrywith of expression * trywith_kind * Backend_var.With_provenance.t
       * expression * Debuginfo.t * kind_for_unboxing
-  | Cregion of expression
-  | Ctail of expression
 
 type property =
   | Zero_alloc
@@ -369,8 +366,6 @@ let iter_shallow_tail f = function
       true
   | Cexit _ | Cop (Craise _, _, _) ->
       true
-  | Cregion _
-  | Ctail _
   | Cconst_int _
   | Cconst_natint _
   | Cconst_float _
@@ -414,8 +409,6 @@ let map_shallow_tail ?kind f = function
               Option.value kind ~default:kind_before)
   | Cexit _ | Cop (Craise _, _, _) as cmm ->
       cmm
-  | Cregion _
-  | Ctail _
   | Cconst_int _
   | Cconst_natint _
   | Cconst_float _
@@ -428,8 +421,6 @@ let map_shallow_tail ?kind f = function
 
 let map_tail ?kind f =
   let rec loop = function
-    | Cregion _
-    | Ctail _
     | Cconst_int _
     | Cconst_natint _
     | Cconst_float _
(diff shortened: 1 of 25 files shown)


backend/debug/available_regs.ml
```diff
@@ -61,18 +61,11 @@ let augment_availability_at_raise avail =
   | Specific_trap (label, _) -> augment_availability_at_exit label avail
 
 let setup_avail_at_raise kind =
```
xclerc, line 63: (We could probably delete both
`setup_avail_at_raise` and
`restore_avail_at_raise`, and
then also `avail_at_raise`.)

backend/selectgen.ml
```diff
@@ -119,7 +78,6 @@ let env_find_static_exception id env =
 
 let env_enter_trywith env kind =
   match kind with
-  | Regular -> { env with trap_stack = Generic_trap env.trap_stack; }
   | Delayed id -> let env, _ = env_add_static_exception id [] env in env
```
xclerc, line 81: (Can we delete the `Generic_trap` constructor?
I think it is no longer produced, only matched upon.)

xclerc (commented):
(Only read 8367c61b11be99986f3d40695ca86fe0ffcce472.)

backend/cmm.ml
```diff
@@ -172,7 +172,6 @@ type trap_action =
   | Pop of pop_action
 
 type trywith_kind =
-  | Regular
   | Delayed of trywith_shared_label
```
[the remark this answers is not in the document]
mshinwell: Fixed

backend/selectgen.ml
mshinwell: Ah yes, deleting that now

backend/debug/available_regs.ml
mshinwell: Doing that now

backend/selectgen.ml
```diff
@@ -1637,9 +1521,7 @@ method emit_tail (env:environment) exp =
      end
   | Cphantom_let (_var, _defining_expr, body) ->
       self#emit_tail env body
-  | Cop((Capply(ty, ((Rc_close_at_apply | Rc_normal) as pos))) as op,
-        args, dbg)
-       when self#tail_call_possible env pos ->
```
lthls, line 1642: The test `backtrace/names` fails because we ignore the position here: `f () [@nontail]` is translated to a tail call, disappearing from the backtrace

backend/selectgen.ml
```diff
@@ -1335,55 +1240,44 @@ method emit_expr_aux (env:environment) exp ~bound_name :
...
-      | Regular -> with_handler env e2
       | Delayed lbl ->
         begin match env_find_static_exception lbl env_body with
         | { traps_ref = { contents = Reachable ts; }; _} ->
           with_handler (env_set_trap_stack env ts) e2
         | { traps_ref = { contents = Unreachable; }; _ } ->
-          let unreachable =
+          let dummy_constant =
+            match r1 with
+            | None -> Ctuple []
```
lthls, line 1256: This line causes the error in the cmm tests.
The body doesn't return, but it might be joined with a branch that does, and we can't guess the type.
We need to find a simple expression that is non-returning; I'll try to come up with something.

backend/cmm.mli
```diff
@@ -154,14 +154,12 @@ type trap_action =
   (** Remove the last handler from the trap stack. *)
 
 type trywith_kind =
```
gretay-js, line 156: Why not just removing `trywith_kind`?

mshinwell-js: It wasn't clear to me this was straightforward, but we should maybe try.

mshinwell: I've done this now.

mshinwell:
I've asked @lthls if he can review the commits after @xclerc 's one.

lthls (approved):
The last commits look correct, alhough I didn't look too closely at the parsecmm part (hopefully the CI will catch obvious bugs).

PR #2172, opened 2023-12-15
Move unboxed float literals to stable

alanechang:
#2062 and #2071 added proper middle/backend support for unboxed float literals which motivates moving it out of alpha.

### DIFF
--- a/ocaml/testsuite/tests/typing-layouts-float64/unboxed_floats.ml
+++ b/ocaml/testsuite/tests/typing-layouts-float64/unboxed_floats.ml
@@ -1,22 +1,23 @@
 (* TEST
+   * flambda2
    reference = "${test_source_directory}/unboxed_floats.reference"
-   * native
+   ** native
      flags = "-extension layouts_alpha"
-   * bytecode
+   ** bytecode
      flags = "-extension layouts_alpha"
-   * native
+   ** native
      flags = "-extension layouts_beta"
-   * bytecode
+   ** bytecode
      flags = "-extension layouts_beta"
-   * native
+   ** native
      flags = "-extension layouts"
-   * bytecode
+   ** bytecode
      flags = "-extension layouts"
-   * setup-ocamlc.byte-build-env
+   ** setup-ocamlc.byte-build-env
      ocamlc_byte_exit_status = "2"
-   ** ocamlc.byte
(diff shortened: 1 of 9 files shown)


goldfirere (approved):

goldfirere:
I can sometimes be rather gung-ho about changes like this, so would also like a second opinion from @ccasin.

ccasin:
> I can sometimes be rather gung-ho about changes like this, so would also like a second opinion from @ccasin.

Seems fine, but maybe now would be a good moment to update `ocaml/testsuite/tests/typing-layouts-float64/unboxed-floats.ml` to use unboxed literals rather than `of_float 3.14` everywhere, for a little bonus confidence in their dynamic behavior?

alanechang:
I updated all tests in `ocaml/testsuite/tests/typing-layouts-float64/` to use unboxed float literals. 

`unboxed_floats_disabled.compilers.reference` is the only test output that changed. Also had to mark all the tests with `* flambda2` due to it being the only middle end with unboxed literal support.

ocaml/testsuite/tests/typing-layouts-float64/unboxed_floats.ml
```diff
@@ -1,22 +1,23 @@
 (* TEST
+   * flambda2
```
goldfirere, line 2: Why was this change necessary? I remember hearing that all the backends should be working.

goldfirere (commented):
Just one question. If that's resolved, we should merge.

ocaml/testsuite/tests/typing-layouts-float64/unboxed_floats.ml
ccasin: That was true before literals, but I assume this is necessary because the literals themselves have only been implemented for flambda2?  I am sad about that, but not sad enough to slow this down.

alanechang: Yes, unboxed literals are currently only supported on flambda2.

The first implementation of the feature was compatible with all middle/backends when it compiles to calling unbox on a boxed constant. But when #2062 and #2071 changed the implementation to compile directly to an unboxed constant, closure and flambda1 support were dropped.

Issue #2156, opened 2023-12-13
bad folding of non-constant arithmetic into array indexing


ahh:
With flambda2 backend we see this:
```
let[@cold] get_sum (t : int array) a b = Array.unsafe_get t (a + b)
```
compile to:
```
add    %rdi,%rax
shl    $0x2,%rax
mov    -0x8(%rbx,%rax,1),%rax
```

rather than folding the shl into the scale of the `mov`.  By comparison:
```
let[@cold] get_sum_const (t : int array) a = Array.unsafe_get t (a + 35)
```
generates the nicer:
```
mov    0x114(%rbx,%rax,4),%rax
```

Combining tthe two:
```
let[@cold] get_sum_and_const (t : int array) a b = Array.unsafe_get t (a + b + 35)
```
we get:
```
add    %rdi,%rax
shl    $0x2,%rax
mov    0x110(%rbx,%rax,1),%rax
```

which again should be folded.

Gbury:
#2173 should solve this problem (at least for amd64).

Gbury:
Closing this since #2173 has been merged.

PR #2173, opened 2023-12-15
Improve constant folding during memory addressing

Gbury:
The default case for addressing a cmm expr of the form `(e * 2|4|8)` was to treat it as a single expression (it happens when `e` is for instance an addition of non-constant expressions). We change the default to instead be a scale memory addressing, allowing more optimisations for memory addressing to take place.

Solves #2156


mshinwell:
@xclerc could you review this?

xclerc (approved):

PR #2180, opened 2023-12-19
Add region in Call_kind for C calls

mshinwell:
The region was missing in the recently-refactored `Call_kind` for C calls, which means that locally-allocating C calls might not be in the correct region.  A new test has been added; in the existing test, allocations from the test itself rather than the C runtime code would cause the region concerned to be held live.

### DIFF
--- a/middle_end/flambda2/terms/call_kind.ml
+++ b/middle_end/flambda2/terms/call_kind.ml
@@ -83,7 +83,7 @@ let [@ocamlformat "disable"] print ppf t =
       Simple.print obj
       Method_kind.print kind
       Alloc_mode.For_allocations.print alloc_mode
-  | C_call { needs_caml_c_call; is_c_builtin;alloc_mode } ->
+  | C_call { needs_caml_c_call; is_c_builtin; alloc_mode } ->
     fprintf ppf "@[<hov 1>(C@ \
         @[<hov 1>(needs_caml_c_call@ %b)@]@ \
         @[<hov 1>(is_c_builtin@ %b)@]@ \
--- a/middle_end/flambda2/terms/call_kind.ml
+++ b/middle_end/flambda2/terms/call_kind.ml
@@ -60,8 +60,9 @@ type t =
         alloc_mode : Alloc_mode.For_allocations.t
       }
   | C_call of
-      { alloc : bool;
-        is_c_builtin : bool
+      { needs_caml_c_call : bool;
+        is_c_builtin : bool;
+        alloc_mode : Alloc_mode.For_allocations.t
       }
 
 let [@ocamlformat "disable"] print ppf t =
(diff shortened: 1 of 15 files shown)


mshinwell:
@lthls @Ekdohibs Please approve if you are around!

mshinwell:
Have squashed into single changeset, this should be ready.

ncik-roberts:
This example fails:

```
external foo : unit -> ('a[@local_opt]) = "foo" [@@noalloc]

let foo () = foo ()
```

with this error message:

```
>> Fatal error: Unexpected free my_region in code with heap result mode:

(apply
 (((Some *extern*.foo)〈k11〉《k10》(0)) (args_arity 𝕍)
  (return_arity 𝕍)
  (call_kind
   (C (alloc false) (is_c_builtin false)
    (alloc_mode (Local (region my_region/17N))))) (dbg test.ml:7,13--19)
  (inline Default_inlined)
  (inlining_state
   (depth 0, arguments
    ((max_inlining_depth 109) (max_rec_depth 0) (call_cost 1.875000)
     (alloc_cost 2.625000) (prim_cost 1.125000) (branch_cost 1.875000)
     (indirect_call_cost 1.500000) (poly_compare_cost 3.750000)
     (small_function_size 100) (large_function_size 500)
     (threshold 100.000000)))) (probe ()) (position Normal)))
```

(The stack trace is quite a bit longer, so I haven't included it, but it should be easy enough to reproduce this locally.)

Ekdohibs:
This is due to a clash between `primitive_may_allocate` in `lambda.ml` which says that `[@@noalloc]` primitives never allocate (even locally), and the code added here in `closure_conversion.ml` which only checks `prim_native_repr_res` and not `prim_alloc` as well. I'm not sure which of these is the correct one, however.

Other than that, this PR looks good; we might even want  to put `alloc_mode` out of `call_kind` if C calls can also create allocations, as there is no point of keeping it inside in that case.

ocaml/lambda/lambda.ml
```diff
@@ -1458,12 +1460,13 @@ let primitive_may_allocate : primitive -> alloc_mode option = function
...
+    if p.prim_alloc then Some alloc_heap else None
   | Pccall p ->
-     if not p.prim_alloc then None
-     else begin match p.prim_native_repr_res with
-       | (Prim_local|Prim_poly), _ -> Some alloc_local
-       | Prim_global, _ -> Some alloc_heap
-     end
+    begin match p.prim_native_repr_res with
+    | (Prim_local | Prim_poly), _ -> Some alloc_local
+    | Prim_global, _ -> Some alloc_heap
```
Ekdohibs, line 1468: This prevents the optimisation of non-local-returning noalloc C calls, I think there should be a noalloc check inside this line.

mshinwell: Fixed (code now shared with Lambda)

middle_end/flambda2/terms/call_kind.ml
```diff
@@ -82,13 +83,15 @@ let [@ocamlformat "disable"] print ppf t =
       Simple.print obj
       Method_kind.print kind
       Alloc_mode.For_allocations.print alloc_mode
-  | C_call { alloc; is_c_builtin; } ->
+  | C_call { needs_caml_c_call; is_c_builtin;alloc_mode } ->
```
Ekdohibs, line 86: Missing space

ocaml/lambda/lambda.ml
```diff
@@ -1443,6 +1443,16 @@ let mod_field ?(read_semantics=Reads_agree) pos =
 let mod_setfield pos =
   Psetfield (pos, Pointer, Root_initialization)
 
+let alloc_mode_of_primitive_description (p : Primitive.description) =
+  if not Config.stack_allocation then
+    if p.prim_alloc then Some alloc_heap else None
+  else
+    match p.prim_native_repr_res with
+    | (Prim_local | Prim_poly), _ -> Some alloc_local
+    | Prim_global, _ -> Some alloc_heap
```
Ekdohibs, line 1452: I think we should replace this by `if p.prim_alloc then Some alloc_heap else None`

Ekdohibs (approved):
Mostly good, see comments

ocaml/lambda/lambda.ml
mshinwell: Fixed

middle_end/flambda2/terms/call_kind.ml
mshinwell: Fixed

PR #2186, opened 2023-12-20
Remove unnecessary constructors from export_info handling

mshinwell:
This follows on from the Closure / Flambda 1 removal.


xclerc (approved):
(Not strictly necessary, but it might be
prudent to bump some magic numbers.)

mshinwell:
Indeed - will do that prior to the next tag.

PR #2190, opened 2023-12-21
Treat Prim_poly primitives as Alloc_heap in Lambda

mshinwell:
Following on from https://github.com/ocaml-flambda/flambda-backend/pull/2189#issuecomment-1865985338

Test case from @ncik-roberts taken from #2189 .  I checked that without the change to `Lambda.alloc_mode_of_primitive_description` in this PR, but with the remainder of this PR applied, the test fails.

### DIFF
--- a/ocaml/lambda/printlambda.ml
+++ b/ocaml/lambda/printlambda.ml
@@ -84,13 +84,14 @@ let array_set_kind ppf k =
   | Pintarray_set -> fprintf ppf "int"
   | Pfloatarray_set -> fprintf ppf "float"
 
-let alloc_mode = function
+let alloc_mode_if_local = function
   | Alloc_heap -> ""
   | Alloc_local -> "local"
 
-let alloc_mode' = function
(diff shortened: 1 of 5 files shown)


ocaml/lambda/printlambda.ml
```diff
@@ -88,6 +88,10 @@ let alloc_mode = function
   | Alloc_heap -> ""
   | Alloc_local -> "local"
 
+let alloc_mode' = function
```
lpw25, line 91: Would you mind renaming this to `alloc_mode` and then renaming the one above to something like `alloc_mode_if_local`? I really dislike `'` names.

lpw25 (approved):
LGTM. One naming suggestion

ocaml/lambda/printlambda.ml
mshinwell: Fixed

mshinwell:
I've cherry-picked this to `main`

PR #2191, opened 2023-12-22
Modify Caml_state->backtrace_last_exn properly

mshinwell:
`Caml_state->backtrace_last_exn` cannot be updated using direct assignment, as it is registered as a generational global root.


xclerc (approved):

mshinwell:
https://github.com/ocaml/ocaml/pull/12861

PR #2192, opened 2023-12-22
Fix assertion in caml_ephe_clean

mshinwell:
This assertion wasn't taking into account the possibility of an ephemeron pointing at static data.


TheNumbat (approved):

mshinwell:
https://github.com/ocaml/ocaml/pull/12860

PR #2193, opened 2023-12-22
Remove unused cextcall function

goldfirere:
Reviewer: @mshinwell

I came across this function and discovered it was unused. Interestingly, after this is gone, the `prim_effects` and `prim_coeffects` fields of `Primitive.description` are now never used. I didn't remove them, too, as that seemed too invasive.


gretay-js (approved):
That's a good catch, thanks! It's fine to remove this dead code. We need to fix Flambda2 to propagate effects and coeffects of primitives, - there are CRs in the code about it, I added https://github.com/ocaml-flambda/flambda-backend/issues/2198 to keep track of this task.

PR #2194, opened 2023-12-26
Move code between Printast and Pprintast

goldfirere:
This reverts commit https://github.com/ocaml-flambda/flambda-backend/commit/bdac10c4886aa4b4ff857274aa63f7bb86ef98e0.

That commit was part of https://github.com/ocaml-flambda/flambda-backend/pull/1417, but ended up getting squashed
in a rebase. Why revert? Because it's hard to upstream this
change, due to the new dependency of the tyvar printer on the
lexer. This can be disentangled, but doing so would be disruptive
for little gain. So I'm just undoing this unforced little cleanup,
which isn't so important anyway.

Review: @alanechang. Recommended review plan: confirm that this patch is just a little shuffling around of definitions and that CI passes.


alanechang (approved):
The changes look good. Can confirm it's only moving around some definitions and CI builds.

PR #2196, opened 2023-12-27
Implement layout histories

alanechang:
This is the `layout-histories` branch rebased over the tip of `main`.

This is done by:
- rebasing on top of the [commit right before the layout to jkind rename](https://github.com/ocaml-flambda/flambda-backend/commit/88f75c4221b11accd361e060473979e4df97fe7c) - merge conflict resolution can be found [here](https://github.com/alanechang/flambda-backend/commit/559b0f41578bfc9b8bd7cb97676f2bc4f0fd0959)
-  rebasing another time over just the jkind rename commits [1](https://github.com/ocaml-flambda/flambda-backend/commit/93be0a1ed2c5a314c9f7992509059e27635ab8b8) [2](https://github.com/ocaml-flambda/flambda-backend/commit/d2c9b3ec180e8e79ad424a437409eba74b544224)
- finally rebasing again over the tip of main

The merge conflicts from the last step are kept and will be resolved via a PR to the `layout-histories-rebased` branch.

PR #2197, opened 2023-12-27
Layout histories: resolve rebase conflicts

alanechang:
This PR resolves the merge conflicts from rebasing `layout-histories` over the tip of `main`. It should be merged before #2196.

### DIFF
--- a/ocaml/testsuite/tests/typing-immediate/immediate.ml
+++ b/ocaml/testsuite/tests/typing-immediate/immediate.ml
@@ -177,10 +177,8 @@ Line 2, characters 2-38:
 Error: The layout of type t is value, because
          it's a boxed record type.
        But the layout of type t must be a sublayout of immediate, because
-         of the annotation on the declaration of the type t/2.
+         of the annotation on the declaration of the type t.
 |}];;
-(* CR layouts v2.9: Investigate why the "/2" is here and check if it's only limited
-   to expect tests. *)
 
 (* Not guaranteed that t is immediate, so this is an invalid declaration *)
 module C = struct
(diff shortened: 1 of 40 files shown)


ocaml/testsuite/tests/typing-immediate/immediate.ml
```diff
@@ -177,10 +177,8 @@ Line 2, characters 2-38:
 Error: The layout of type t is value, because
          it's a boxed record type.
        But the layout of type t must be a sublayout of immediate, because
-         of the annotation on the declaration of the type t/2.
+         of the annotation on the declaration of the type t.
 |}];;
-(* CR layouts v2.9: Investigate why the "/2" is here and check if it's only limited
```
goldfirere, line 182: I love it when delayed problems just melt away. :) (Resolve after reading.)

ocaml/testsuite/tests/typing-layouts-float64/basics.ml
```diff
@@ -266,7 +287,10 @@ Error: Layout mismatch in final type declaration consistency check.
        clever enough to propagate layouts through variables in different
        declarations. It is also not clever enough to produce a good error
        message, so we'll say this instead:
-         'a has layout float64, which does not overlap with value.
+         The layout of 'a is float64, because
```
goldfirere, line 290: Actually, that message is somewhat decent! The type inference isn't clever, but at least the error is. Could reword the message. But humility is a virtue, and so I don't actually think we need to change anything. But I'm pleased. (Resolve after reading.)

goldfirere (approved):

Issue #2198, opened 2023-12-28
Propagate effects and coeffects of primitives through flamdba2


gretay-js:
Externals can be annotated with `[@@no_effects]` and `[@@no_coeffects]`. This information was propagated through closure and flambda1 and used in selection.

PR #2199, opened 2023-12-29
Propagate effect and coeffect attributes on C calls through flambda2

gretay-js:
Address https://github.com/ocaml-flambda/flambda-backend/issues/2198/

This PR only propagates the attributes to the backend (similarly to the way it was in closure and flambda1) so they can be used in selection, but does not make this information available to optimizations in the middle-end.


mshinwell (approved):

mshinwell:
Just needs `make fmt` then ok to merge.

PR #2200, opened 2023-12-29
Disable some SIMD tests when running under Rosetta2

xclerc:
This pull request disables the SIMD tests
failing when using Rosetta2 (determined
by running the test suite on an M2 box).

Rosetta2 is macOS' emulation layer that
allows one to run `amd64` binaries on
Apple Silicon boxes.

### DIFF
--- a/tests/simd/test_helpers.ml
+++ b/tests/simd/test_helpers.ml
@@ -1,0 +1,28 @@
+type rosetta2_status =
+  | Unknown
+  | Off
+  | On
+
+let get_rosetta2_status () =
+  let file_path = Filename.temp_file "ocaml-rosetta2" "sysctl" in
+  let command = Printf.sprintf "/usr/sbin/sysctl -n sysctl.proc_translated > %S 2> /dev/null" file_path in
+  let exit_code = Sys.command command in
+  match exit_code with
+  | 0 ->
+    begin try
+        let file_chan = open_in file_path in
+        let first_line = input_line file_chan in
+        match first_line with
+        | "0" -> Off
+        | "1" -> On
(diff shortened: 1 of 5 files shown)


tests/simd/test_helpers.ml
```diff
@@ -0,0 +1,28 @@
...
+        | "1" -> On
+        | _ -> Unknown
+      with _ ->
+        Unknown
+    end
+  | _ ->
+    Unknown
+
+let run_if_not_under_rosetta2 ~f =
+  match get_rosetta2_status () with
```
mshinwell, line 26: Maybe we should print the status each time?

mshinwell (approved):
One minor comment

xclerc:
Note: the tests in `scalar_ops.ml` fail with `SIGILL`
while the ones in `ops.ml` seems to fail because of
different rounding.

tests/simd/test_helpers.ml
xclerc: That would be useful, but wouldn't that
make the test not reproducible?

xclerc:
The test suite is failing on the CI (not finding
the new module) but is OK locally. Did the
semantics of dune's `modules` stanza change
at some point?

xclerc:
Actually, I have the same dune version locally (3.6.1).

xclerc:
Please ignore the dune-related comments
above; the issue was an auxiliary target
which is not built on non-macOS boxes.

tests/simd/test_helpers.ml
mshinwell: Ah, true.

PR #2201, opened 2023-12-29
Remove checkbound and checkalign

gretay-js:
Checkbound and Checkalign instructions are not generated by flambda2. After removal of flambda1 and closures, these instructions can be deleted from the backend.


TheNumbat (approved):
Looks good. I've removed the checkbound instructions from backend/arm64. 

There's also code for handling bounds/align check exceptions in the runtime, but I assume we need to keep that around since the upstream build still uses them.

PR #2202, opened 2023-12-29
Add not_float

goldfirere:
This adds a new type to the initial environment, `not_float` that is, well, not `float`.

See documentation in predef.mli, repeated here:

```
(** The [not_float] type is an abstract type that is known not to
    be [float]. [Typeopt.classify] is the only place that looks for it.
    Recommended use:
    {[
      module M : sig
        type !'a t = private 'a not_float
        val create : unit -> t
        val use : t -> ...
      end = struct
        type !'a t = private 'a not_float  (* private, even in the struct *)
        external create : unit -> t = "some_bit_of_magic"
        external use : t -> ... = "more_magic"
      end
    ]}
    Now, everyone knows that [M.t] is not a float, meaning that the
    [M.t array] usages will not need to dynamically check for floats,
    and will definitely not allocate.
    Why is this safe? Because there is no way to use a [not_float].
    The only reason to parameterize [not_float] is to satisfy injectivity
    checks.
*)
```

Some questions:

1. Is this a good idea? @ccasin is a little worried that someone might forget the `private` in the struct. I think it's probably OK, but agree that this mistake is possible.
2. What to do about public release? The easiest would be to just add `type !'a not_float` to somewhere (where?) in the public-release ecosystem. That way, code using `not_float` will continue to compile, but won't have the magical not-float behavior.
3. I've magically made the parameter to `not_float` bivariant and injective. This is doubly impossible: OCaml doesn't support abstract bivariant types, and an injective parameter is certainly not bivariant. An alternative is to give `not_float` two parameters, one covariant and one contravariant (both injective), but that feels like overkill. Another alternative is to keep the parameter invariant -- only the injectivity is needed for `Bigarray` (I added the variance just because I was there). Is what I've done OK?
4. Is the test added in the right way?


Gbury:
Is there a reason not to add an annotation instead of the predefined type ? I can think of a few potential reasons, such as lack of propagation of annotations in the typechecker, but it'd be good to check that the idea has been explored and the reasons why it wouldn't work.

goldfirere:
@Gbury A good question. I think this would basically amount to something quite similar to upstream's `[@@immediate]` annotations. ("upstream's" because we here have layouts now.) There's a fair amount of fuss with those, including checking the annotation in structures, carefulness around substitutions and `with` constraints, and propagation (as you mention). The approach written here is less powerful -- it won't work if the implementation of a type is internal to OCaml instead of magical -- but extremely lightweight. If it's insufficient, I think we would favor making a `not_float` sublayout of `value` over doing the annotation. @ccasin and I estimate that the work would be similar, and making `not_float` a layout integrates better with existing technology.

mshinwell:
@goldfirere what's the plan for this one (still seems useful)?

goldfirere:
Yes, this is still useful. Conversation on Slack, though, let to the idea of giving more user control here. Instead of just having one `not_float` type, we would give the users the ability to define external types -- types whose semantics are given by the compiler or by C code, say. I have this mostly complete in https://github.com/goldfirere/flambda-backend/tree/rae/external-types, but:
* There is still more work to do there (I forget the details now, though I don't remember it being thorny)
* There is still some design debate to be had
* It seemed lower priority than other work on my plate (e.g. the mode solver and modal kinds)

I think it would take maybe a cumulative day more of work to get that out. If I hear enough murmurs that the time investment is worthwhile, then I'll gladly put in the day.

Another possibility is just to merge this patch. But actually I think it would be better to consider the more general approach taken in https://github.com/goldfirere/flambda-backend/tree/rae/external-types

goldfirere:
We've decided not to do this, in favor of separability tracked in kinds.

PR #2203, opened 2024-01-02
Fix caml_alloc_shr_check_gc for tags >= No_scan_tag

mshinwell:
The assertions in the new `caml_alloc_shr_check_gc` function don't match those in `caml_alloc`, which causes debug runtime5 builds to fail under flambda2 (and for there to be unnecessary block initialization).


xclerc (approved):

mshinwell:
I'm not certain the upstream middle ends can hit this (to be checked), but I think we should upstream anyway.

PR #2208, opened 2024-01-04
Ensure the minor heap is actually empty before reallocating it

mshinwell:
The `caml_domain_external_interrupt_hook` strikes again.  When `caml_set_minor_heap_size` is called, the minor heap is emptied by calling `caml_minor_collection`, before the existing minor heap for the domain concerned is freed and reallocated.  Except that `caml_minor_collection` doesn't necessarily leave the heap empty, because of the hook.

To fix this, in this PR we use `caml_empty_minor_heaps_once`, which doesn't call the hook.  The test case that was failing is a bit intermittent but I have reasonable confidence this fix has solved the problem.

I added an assertion at the end of `caml_stw_empty_minor_heap_no_major_slice` to show that the minor heap is indeed empty at the end of that function, since this isn't that obvious.


mshinwell:
Upstream PR: https://github.com/ocaml/ocaml/pull/12879

mshinwell:
I'm going to merge this, as it was approved upstream.

PR #2210, opened 2024-01-04
Simplify csel builtin when condition is known

mshinwell:
So for example the following two functions:
```
external select : bool -> 'a -> 'a -> 'a = "caml_csel_value"
[@@noalloc] [@@no_effects] [@@no_coeffects] [@@builtin]

let csel x y z =
  let a = select true x y in
  let b = select false x y in
  select z a b

let csel2 x y z =
  select z x y
```
both now compile to:
```
        movq    %rax, %rsi
        movq    %rbx, %rax
        cmpq    $1, %rdi
        cmovne  %rsi, %rax
        ret
```
Previously for the first one we produced:
```
        movq    %rax, %rsi
        movq    %rbx, %rax
        movl    $1, %ebx
        movq    %rax, %rdx
        testq   %rbx, %rbx
        cmovne  %rsi, %rdx
        xorl    %ebx, %ebx
        testq   %rbx, %rbx
        cmovne  %rsi, %rax
        cmpq    $1, %rdi
        cmovne  %rdx, %rax
```
@TheNumbat could you review?

### DIFF
--- a/backend/cmm_builtins.ml
+++ b/backend/cmm_builtins.ml
@@ -490,7 +490,11 @@ let transl_builtin name args dbg typ_res =
     let op = Ccsel typ_res in
     let cond, ifso, ifnot = three_args name args in
     if_operation_supported op ~f:(fun () ->
-        Cop (op, [test_bool dbg cond; ifso; ifnot], dbg))
+        let cond = test_bool dbg cond in
+        match cond with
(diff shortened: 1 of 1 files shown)


TheNumbat (approved):
How `csel` deals with tagged bools is a bit confusing, but I think it's correct. Might warrant a comment.

PR #25, opened 2021-05-10
Amd64 emit tweaks

xclerc:
I was instructed to transfer/revive https://github.com/ocaml/ocaml/pull/1490 here.
It is noteworthy, though, that the most interesting tweak was
merged through https://github.com/ocaml/ocaml/pull/8990. The remaining tweaks are:

- a slightly shorter code when `and`-ing a register and a
  known (32-bit) constant; 
- a slightly shorter code when `xor`-ing a register with itself;
- a (conditional) use of `add`/`sub` instead of `inc`/`dec` because
  the latter may introduce unnecessary data dependencies.

PR #2211, opened 2024-01-05
Always use x86 inc/dec for add/sub with 1

TheNumbat:
The x86 backend now emits inc/dec instead of add/sub with 1, even if the speed-over-size flag is enabled. 
The encoding of inc/dec is smaller than the full add/sub, so this reduces code size. 
In particular, the popcount intrinsic codegen is improved.

In general, inc/dec can be slower than add/sub, as it doesn't reset all flags (so there can be a data dependency on cf).
They're also extra slow on [recent Intel efficiency cores](https://stackoverflow.com/questions/36510095/inc-instruction-vs-add-1-does-it-matter), but this doesn't apply to most CPUs, so reduced code size should be a win.

### DIFF
--- a/backend/amd64/emit.mlp
+++ b/backend/amd64/emit.mlp
@@ -1443,9 +1443,9 @@ let emit_instr fallthrough i =
       instr_for_intop op (arg i 1) (res i 0)
   | Lop(Iintop_imm(Iadd, n)) when i.arg.(0).loc <> i.res.(0).loc ->
       I.lea (mem64 NONE n (arg64 i 0)) (res i 0)
-  | Lop(Iintop_imm(Iadd, 1) | Iintop_imm(Isub, -1)) when not !fastcode_flag ->
+  | Lop(Iintop_imm(Iadd, 1) | Iintop_imm(Isub, -1)) ->
       I.inc (res i 0)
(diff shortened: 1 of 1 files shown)


gretay-js (commented):
The condition was introduced in https://github.com/ocaml-flambda/flambda-backend/pull/25 and discussed upstream in https://github.com/ocaml/ocaml/pull/1490#issuecomment-357528533 but the change hasn't been merged upstream, so this PR would make codegen similar to upstream again.

What motivated this change now? You mentioned `popcnt` - do you an example and does the pattern come up in other places? I don't object to this PR, just curious. Also, wonder if @xclerc originally had a use case for preferring `sub/add` and if it is still relevant.

TheNumbat:
Popcount is hardcoded to generate `popcnt` + `add 0xffffffffffffffff`, so it's now `popcnt` + `dec` instead. This change was suggested by ahunter.
I don't think any of the other intrinsics specifically generate an add/sub with 1, but any operation that compiles to an add/sub with int constant 1 will become inc/dec as well.

TheNumbat:
@xclerc what do you think here?

mshinwell:
@xclerc ping

xclerc (approved):

PR #2205, opened 2024-01-03
Enable Stdlib.Effect + add Flambda 2 support

mshinwell:
- Uncomments the contents of `Stdlib.Effect`
- Adds flambda2 support for effect operations.  These are encoded as a new variety of `Apply_expr`, rather than primitives, since they may raise exceptions
- Adds the necessary `Cmm_helpers` support needed by flambda2
- Uncomments the upstream code in `Cmmgen` for good measure.

PR #2215, opened 2024-01-09
Locals x effects

mshinwell:
This provides support for local allocations in the presence of algebraic effects.

There aren't presently any clever optimisations, although care is taken to ensure that the code sequences generated by the OCaml compiler for local allocations remain unchanged.  This is done by arranging for some of the per-stack locals state to be present at the root of `Caml_state` in addition to being stored in the `stack_info` structures.

Based on #2205.


mshinwell:
I'm also going to put in this PR the re-enabling of tests disabled with the `CR ocaml 5 effects` marker.

lpw25:
Could you or @stedolan add some description to this PR of how this works, and some indication of what invariants it relies on to work? It's unclear whether this supports e.g. pointers to locals on a parent's stack.

mshinwell:
> Could you or @stedolan add some description to this PR of how this works, and some indication of what invariants it relies on to work? It's unclear whether this supports e.g. pointers to locals on a parent's stack.

Sure, I'll probably do this as comments in the code.

mshinwell:
Superceded by #3648

PR #2216, opened 2024-01-09
fixes #2032

riaqn:
This fixes a BUG introduced by #2032 

Request review from @stedolan


stedolan (approved):
Looks good.

This will need another fix, now or later: it does two separate prints to stderr, so multi-domain programs risk interleaving their timestamps and logs. (The fmtbuf in caml_gc_log fixes this)

riaqn:
That's fair - I will make a separate PR for it. Merging for now.

PR #2217, opened 2024-01-09
Add new C primitives for custom block sizes

goldfirere:
EDIT: Reach out to me on Slack. There may be an easier fix for the actual problem than the approach here. This code might still be useful to someone sometime, but I think it's less pressing.

A use case has arisen where a colleague wants the GC to know about the quantity of memory allocated deep within some C that's called from OCaml. Custom blocks have this ability: you can tell a custom block how much off-heap memory is being used, which tells the GC to speed up if memory is being used too quickly. But in this particular use case, it's awkward to allocate the custom block deep within the C code, because there are no other managed-heap values lying around, and so the code around the allocation doesn't do the usual `CAMLparam` and `CAMLreturn` stuff needed to keep the GC happy.

It turns out that the final pointer (to the unmanaged memory) returned to OCaml is stored as the contents of a `nativeint`. A `nativeint` is a custom block. So @ncik-roberts and I thought it would be cool just to tell the GC that the `nativeint` has this extra off-heap memory associated with it.

This PR adds two new C primitives to the OCaml runtime, `caml_set_custom_size` and `caml_set_custom_size_mem` that take an already-allocated custom block and records the extra memory usage. The `external` declarations to use these are given in comments in C. I have not added `external` declarations to the stdlib, because doing so has a poor public-release story. After we merge this and roll, I would expect we would add these primitives to a suitable non-public-release library. If/when we upstream this (we should), I would expect the functions to be added to `Gc`.

Open question: does this work for heap values that do not have the `Custom_tag`? That would be nice.

Review request: @stedolan

### DIFF
--- a/ocaml/runtime/custom.c
+++ b/ocaml/runtime/custom.c
@@ -32,6 +32,96 @@ uintnat caml_custom_major_ratio = Custom_major_ratio_def;
 uintnat caml_custom_minor_ratio = Custom_minor_ratio_def;
 uintnat caml_custom_minor_max_bsz = Custom_minor_max_bsz_def;
 
+static void caml_set_minor_custom_size(value custom_obj,
+                                       mlsize_t mem,
+                                       mlsize_t max_major,
+                                       mlsize_t mem_minor,
+                                       mlsize_t max_minor)
+{
+  CAMLparam1(custom_obj);
+
+  if (mem > mem_minor) {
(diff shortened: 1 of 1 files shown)


goldfirere:
This needs tests, but I don't know the most effective way to do so. I'd be happy for someone else to contribute tests here, or to point me to a place where existing tests have the right flavor to exercise the GC in the right way.

goldfirere:
I think we've decided not to do this.

PR #2218, opened 2024-01-09
Cleanup simd regalloc flag

TheNumbat:
Deletes the `-fsimd-regalloc` amd64 backend flag, as we no longer need it for testing.
SIMD registers are now enabled whenever the SIMD language extension is enabled.

### DIFF
--- a/backend/amd64/emit.mlp
+++ b/backend/amd64/emit.mlp
@@ -41,8 +41,6 @@ let _label s = D.label ~typ:QWORD s
 
 (* Override proc.ml *)
 
-let simd_frontend_disabled () = not (Language_extension.is_enabled SIMD)
-
 let int_reg_name =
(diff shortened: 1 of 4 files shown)


backend/amd64/emit.mlp
```diff
@@ -41,7 +41,9 @@ let _label s = D.label ~typ:QWORD s
 
 (* Override proc.ml *)
 
-let simd_frontend_disabled () = not (Language_extension.is_enabled SIMD)
+let assert_simd () =
```
gretay-js, line 44: Nice! Could this go into `Arch` to avoid another copy of it in `proc.ml` ?
```suggestion
let assert_simd_extension_enabled () =
```

TheNumbat: Sure, moved + renamed

gretay-js (approved):

PR #2221, opened 2024-01-10
Mark symbols as having visibility STV_PROTECTED

poechsel:
Because of cross module inlining we generally don't want OCaml symbols to be preempt-able. This means that when doing dynamic linking / using dynlink we don't want OCaml symbols to be override by another implementation.
As this is not the case in C, linkers are delaying symbol resolution under the last minute. However, the `STV_PROTECTED` symbol visibility allow to control that behavior, and allows linkers to resolve more relocations when building shared libraries.

More information about STV_PROTECTED is available [here](https://docs.oracle.com/cd/E19683-01/816-7529/chapter6-79797/index.html)
In practice the linker will behave as-if `-bsymbolic` was passed to it and limited to OCaml symbols.

### DIFF
--- a/backend/internal_assembler/symbol_entry.ml
+++ b/backend/internal_assembler/symbol_entry.ml
@@ -88,11 +88,11 @@ let create_symbol (symbol : X86_binary_emitter.symbol) symbol_table sections
     | Some s -> failwith ("Unknown symbol type" ^ s)
     | None -> 0
   in
-  let global_and_protected = Bool.to_int symbol.sy_global_and_protected in
+  let global = Bool.to_int symbol.sy_global in
   let symbol_entry =
     { st_name = String_table.current_length string_table;
-      st_info = (global_and_protected lsl 4) lor bind;
-      st_other = if symbol.sy_global_and_protected then 0x3 else 0;
+      st_info = (global lsl 4) lor bind;
+      st_other = if symbol.sy_protected then 0x3 else 0;
       st_shndx =
         Section_table.get_sec_idx sections
           (X86_proc.Section_name.of_string symbol.sy_sec.sec_name);
--- a/backend/internal_assembler/symbol_entry.ml
+++ b/backend/internal_assembler/symbol_entry.ml
@@ -88,11 +88,11 @@ let create_symbol (symbol : X86_binary_emitter.symbol) symbol_table sections
     | Some s -> failwith ("Unknown symbol type" ^ s)
     | None -> 0
   in
-  let global = Bool.to_int symbol.sy_global in
+  let global_and_protected = Bool.to_int symbol.sy_global_and_protected in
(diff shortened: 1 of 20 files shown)


backend/x86_ast.mli
```diff
@@ -374,7 +374,7 @@ type asm_line =
   | Byte of constant
   | Bytes of string
   | Comment of string
-  | Global of string
+  | Global of { protected: bool; sym: string }
```
gretay-js, line 377: It would be better to have `Protected` directive as a separate varient, not part of `Global`, similarly to the way we have `Hidden` separately.

backend/internal_assembler/symbol_entry.ml
```diff
@@ -92,7 +92,7 @@ let create_symbol (symbol : X86_binary_emitter.symbol) symbol_table sections
   let symbol_entry =
     { st_name = String_table.current_length string_table;
       st_info = (global lsl 4) lor bind;
-      st_other = 0;
+      st_other = if symbol.sy_protected then 0x3 else 0;
```
gretay-js, line 95:
```suggestion
      st_other = if symbol.sy_protected then 3 else 0;
```

backend/x86_gas.ml
```diff
@@ -433,7 +433,10 @@ let print_line b = function
         bprintf b "\t.ascii\t\"%s\""
           (string_of_substring_literal i (String.length s - i) s)
   | Comment s -> bprintf b "\t\t\t\t/* %s */" s
-  | Global s -> bprintf b "\t.globl\t%s" s;
+  | Global s ->
+    bprintf b "\t.globl\t%s" s
```
gretay-js, line 437: unrelated diff (whitespace only)
```suggestion
  | Global s -> bprintf b "\t.globl\t%s" s
```

ocaml/driver/main_args.ml
```diff
@@ -296,6 +296,15 @@ let mk_no_probes f =
     "-no-probes", Arg.Unit f, " Ignore [%%probe ..]"
 ;;
 
+let mk_symbol_visibility_protected f =
```
gretay-js, line 299: why is it in `Main_args` and not in `Flambda_backend_args`?

ocaml/driver/main_args.ml
```diff
@@ -296,6 +296,15 @@ let mk_no_probes f =
...
+    "-symbol-visibility-protected", Arg.Unit f,
+    " Emit symbols with visibility STV_PROTECTED"
```
gretay-js, line 301:
```suggestion
    "-symbol-visibility-protected", Arg.Unit f,
    " Emit global symbols with visibility STV_PROTECTED on supported systems"
```

gretay-js (approved):

PR #2222, opened 2024-01-10
Respect return modes for @@unboxed primitives

mshinwell:
The translation into Flambda 2 is currently ignoring the return modes for primitives that are annotated `[@@unboxed]`, which can lead to allocations being on the heap instead of the local stack.

### DIFF
--- a/middle_end/flambda2/from_lambda/closure_conversion.ml
+++ b/middle_end/flambda2/from_lambda/closure_conversion.ml
@@ -492,19 +492,24 @@ let close_c_call acc env ~loc ~let_bound_ids_with_kinds
              (fun var -> Named.create_simple (Simple.var var))
              let_bound_vars))
   in
+  let alloc_mode =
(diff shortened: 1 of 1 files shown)


lthls (approved):
There are still some subtle bugs hiding around all of that.
Some primitives have mode `Prim_poly`, which means they could be returning either a local or heap value. With the current code, we assume that they always return a heap value.
So for example:
```ocaml
external get_float : unit -> float[@local_opt] = "my_primitive" [@@unboxed]
```
If I understand correctly, the C side only has to return a naked float, then the OCaml compiler should introduce the proper boxing primitive depending on the inferred mode at the use point.
I believe that currently we do not compute this mode, but assume it must be `Heap`.

This PR does fix the case where the return mode is declared as local, so I think it's fine to merge as it is, but I would welcome some additional opinions on the `Prim_poly` case.

PR #2225, opened 2024-01-11
Greedy register allocator (POC)

xclerc:
This pull request adds a third register
allocator to the CFG pipeline. The new
allocator is loosely based on LLVM's
greedy allocator.

At this point, it is little more than
a mere proof-of-concept. Like LLVM's
greedy allocator, it first builds the
intervals for uses (recycling code from
the linscan allocator) and then uses a
priority queue to drive the attempts to
assign the temporaries to hardware
registers. Unlike LLVM's greedy allocator,
when it cannot assign a temporary, it
will not try to split the temporary,
but will instead always spill (recycling
code from the IRC allocator).

The focus for this first pull request
has been on correctness, and the new
allocator can already be used to build
(with the validator on):

- the compiler distribution, including
  its test suite (the pull request adds
  a new CI job showing that);
- several large-ish applications with
  a number of dependencies, like jenga.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -76,6 +76,12 @@ jobs:
             build_ocamlparam: '_,w=-46,regalloc=ls,regalloc-param=SPLIT_LIVE_RANGES:on,regalloc-param=LS_ORDER:layout,regalloc-validate=1'
             check_arch: true
 
+          - name: gi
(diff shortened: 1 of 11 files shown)


TheNumbat (approved):
Looks reasonable (no bad top level side effects etc.) so approving - more review to come.

PR #2230, opened 2024-01-15
Tweak sets of destroyed registers

xclerc:
As noted by @mshinwell, `rbx` is clobbered
by an external call only if the call is not
allocating. This pull request changes the sets
of clobbered registers to reflect that.

note: this change increases the number of
allocations (by the compiler); it seems OK
for testing but should be fixed in due time.


xclerc:
(That patch makes no sense - closing.)

PR #2233, opened 2024-01-17
Small bug fixes for middle/backend unboxed array support

alanechang:
This PR contains fixes for three bugs found when adding frontend support:

1. array length output needs to be tagged
2. ref/set operations receive tagged ints as inputs. the process of skipping the custom field needs to be updated
3. correct symbol names for the custom ops

Each fix is its own commit. Might be easier to review one commit at a time.


mshinwell (approved):
These were all my bugs, thanks for fixing.

PR #2234, opened 2024-01-18
Fix layout annotation encoding to work better with ppxlib 

alanechang:
The value declaration `val f1 : ('a : float64) ('b : immediate) 'c. 'a -> 'b -> 'c` has both poly vars with and without layout annotations. It's currently parsed as

```
signature_item (//toplevel//[2,16+6]..[2,16+65])
              Psig_value
              value_description "f1" (//toplevel//[2,16+10]..[2,16+12]) (//toplevel//[2,16+6]..[2,16+65])
                core_type (//toplevel//[2,16+15]..[2,16+65])
                  attribute "jane.erasable.layouts.poly"
                    [
                      structure_item (_none_[0,0+-1]..[0,0+-1]) ghost
                        Pstr_eval
                        expression (//toplevel//[2,16+15]..[2,16+65]) ghost
                          Pexp_ident "float64" (//toplevel//[2,16+21]..[2,16+28])
                      structure_item (_none_[0,0+-1]..[0,0+-1]) ghost
                        Pstr_eval
                        expression (//toplevel//[2,16+15]..[2,16+65]) ghost
                          Pexp_ident "immediate" (//toplevel//[2,16+36]..[2,16+45])
                      structure_item (_none_[0,0+-1]..[0,0+-1]) ghost
                        Pstr_attribute "none"
                        []
                    ]
                  attribute "jane.erasable.layouts"
                    []
                  Ptyp_poly 'a 'b 'c
                  core_type (//toplevel//[2,16+51]..[2,16+65])
                    ...
                []
```

Notably, jane syntax uses an attribute with name `"none"` to encode the fact that `'c` doesn't have a layout annotation.

The problem arises when ppxlib tries to process the program and fails:
```
File ".../_none_", line 1:
Error: Attribute `none' was not used
```

This PR attempts to fix this error by changing the attribute name from `none` to `jane.none` since ppxlib already ignores unused attributes that starts with `jane.`

### DIFF
--- a/ocaml/parsing/jane_syntax.ml
+++ b/ocaml/parsing/jane_syntax.ml
@@ -211,7 +211,7 @@ module Make_payload_protocol_of_stringable (Stringable : Stringable) :
     let structure_item_of_none =
       { pstr_desc =
           Pstr_attribute
-            { attr_name = Location.mknoloc "none";
+            { attr_name = Location.mknoloc "jane.none";
               attr_payload = PStr [];
               attr_loc = Location.none
             };
@@ -278,7 +278,8 @@ module Make_payload_protocol_of_stringable (Stringable : Stringable) :
         | _ -> raise Unexpected
 
       let is_none_structure_item = function
-        | { pstr_desc = Pstr_attribute { attr_name = { txt = "none" } } } ->
+        | { pstr_desc = Pstr_attribute { attr_name = { txt = "jane.none" } } }
+          ->
           true
         | _ -> false
 
--- a/ocaml/testsuite/tests/parsetree/source_jane_street.ml
+++ b/ocaml/testsuite/tests/parsetree/source_jane_street.ml
@@ -9,6 +9,7 @@
 let f (type a : immediate) (x : a) = x;;
 let f (type (a : immediate)) (x : a) = x;;
 let f (type (a : immediate) (b : immediate)) (x : a) = x;;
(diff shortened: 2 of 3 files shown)


ncik-roberts (approved):
The approach seems good.

My only thought is about testing. It would be nice if we ran a battery of tests on a battery examples of Jane Street syntax to ensure that the attributes are of the expected form. Right now, the closest thing we have to that battery of examples is `ocaml/testsuite/tests/parsetree/source_jane_street.ml` &mdash; I think there's a decent chance we'll remember to extend it with future examples, too. Could you spend a small amount of time seeing whether it's easy to modify `ocaml/testsuite/tests/parsetree/test.ml` to add this check? I think you could do something like "all attributes start either with `test` or with `jane`". I'm open to pushback.

alanechang:
Added an additional test for `ocaml/testsuite/tests/parsetree/source_jane_street.ml` to catch this kind of problems. It checks the attribute/extension names all begin with `extension.` (still used for locals), `jane.`, and `test.` . Had to do it with `Printast` because `Ast_mapper` decodes jane syntax by default which hides away the attributes used for its encoding and we will have to duplicate a fair bit of code to fit it to this use case.

ncik-roberts:
I discussed with Alan whether it would be better to use a patched `Ast_iterator` to do the check, and we decided ultimately that it wasn't worth the extra effort (as we could silently forget to patch the iterator method for a relevant syntactic construct, so it doesn't buy us much).

PR #2235, opened 2024-01-19
Parsing new mode syntax

riaqn:
This depends on https://github.com/ocaml-flambda/flambda-backend/pull/2231

This PR supports parsing new syntax in addition to the old syntax. Printing is not changed and still in old syntax.

```
let x @ local = ... in 
let foo () = @local
  "hello"
in
type r = string @ local -> string @local
type r = { x @ global : string}
```


riaqn:
Note to myself: we should modify `at_mode_expr` and `atat_mode_expr` to improve the error message.

riaqn:
This needs to be rebased on https://github.com/ocaml-flambda/flambda-backend/pull/2318

ocaml/testsuite/tests/typing-modes/modes.ml
```diff
@@ -0,0 +1,59 @@
...
+        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Error: This value escapes its region
+|}]
+
+let foo @ foo = "hello"
+[%%expect{|
+Line 1, characters 10-13:
+1 | let foo @ foo = "hello"
+              ^^^
+Error: Unrecognized mode foo for coercion.
```
ncik-roberts, line 55: Suggestion: remove "for coercion"

ocaml/parsing/parser.mly
```diff
@@ -3085,15 +3091,27 @@ let_binding_body_no_punning:
...
       { ($1, $3, None, []) }
-  | simple_pattern_not_ident COLON core_type EQUAL seq_expr
-      { ($1, $5, Some(Pvc_constraint { locally_abstract_univars=[]; typ=$3 }), []) }
+  | simple_pattern_not_ident modes = at_mode_expr_nonempty EQUAL seq_expr
+      {
+        let modes_ghost = Mode.ghostify modes in
+        let exp = mkexp_with_modes ~ghost:true ~loc:$sloc modes_ghost $4 in
+        ($1, exp, None, let_binding_mode_attrs modes)
+      }
+  | simple_pattern_not_ident COLON core_type modes=atat_mode_expr EQUAL seq_expr
```
ncik-roberts, line 3139: Did you consider making a single production for "either `at_mode_expr` OR `COLON core_type atat_mode_expr`"? That would allow you to merge these two rules, potentially.

ocaml/parsing/parser.mly
```diff
@@ -4263,6 +4341,45 @@ strict_function_or_labeled_tuple_type:
    | { Mode.empty }
    | mode_expr_legacy_nonempty {$1}
 ;
+
+/* New mode annotation, followed by AT or ATAT and thus enjoys its own namespace */
```
ncik-roberts, line 4345:
```suggestion
/* New mode annotation, introduced by AT or ATAT */
```

ocaml/parsing/parser.mly
```diff
@@ -4263,6 +4298,30 @@ strict_function_or_labeled_tuple_type:
...
+  | AT mode_expr {$2}
+;
+
+%inline atat_mode_expr_nonempty:
+  | ATAT mode_expr {$2}
+
+%inline atat_mode_expr:
+  | { Mode.empty }
+  | atat_mode_expr_nonempty {$1}
+;
```
ncik-roberts, line 4381: Nit: I would find it more understandable to have `atat_mode_expr`, `at_mode_expr`, and `optional_atat_mode_expr`. This makes it easier for me to understand which clients expect `@@` and `@` to necessarily show up in the syntax.

ocaml/testsuite/tests/typing-modes/modes.ml
```diff
@@ -0,0 +1,59 @@
+(* TEST
+   * expect *)
```
ncik-roberts, line 2: I think there should be a test that shows what happens if you specify the local mode in both syntaxes at once. My expectation is that the type-checker will complain that locality was specified twice.

ncik-roberts (commented):
Review for first commit.

ocaml/typing/typemode.ml
```diff
@@ -40,23 +41,24 @@ let transl_mode_annots modes =
...
+      let acc : Global_flag.t =
         match txt with
         | "global" -> (
           Jane_syntax_parsing.assert_extension_enabled ~loc Mode ();
           match acc with
-          | Unrestricted -> Global_flag.Global
+          | Unrestricted -> Global
+          (* Duplicated modality is not an error, just silly and thus a warning.
+             As we introduce more modalities, it might be in general difficult
+             to detect all redundant modalities, but we should do our best. *)
```
ncik-roberts, line 55: I don't understand why duplicated mode is an error but duplicated modality is a warning. I find consistency to be compelling here. Is there something I'm not considering?

ocaml/parsing/parser.mly
```diff
@@ -150,6 +150,8 @@ let mk_attr ~loc name payload =
   Builtin_attributes.(register_attr Parser name);
   Attr.mk ~loc name payload
 
+(* CR zqian: someday we should put attributes on the expression, instead of
+extension node. *)
```
ncik-roberts, line 154: Why?

ocaml/testsuite/tests/typing-modes/modes.ml
```diff
@@ -232,3 +232,56 @@ Warning 250 [redundant-modality]: This global modality is redundant.
...
+1 | let foo ?x:(local_ (x,y) : 'a.'a->'a @@ unique once) () = ()
+                       ^^^^^^^^^^^^^^^^^
+Error: Optional parameters cannot be polymorphic
+|}]
+
+(* note: Legacy mode syntax is not parsed for patterns *)
+
+(* CR zqian: currently all patterns, except those directly as function
+   arguments, DO NOT pick up modes during type checking. This should be fixed in
+   another PR. Here, we test that they at least parse. *)
```
ncik-roberts, line 278: I'm not inclined to support this syntax until it means something. Can we reject this in type-checking? If not, can we just delete the parse rules for now?

ocaml/parsing/parser.mly
```diff
@@ -173,11 +175,22 @@ guaranteed to be used. *)
...
   let modes = Mode.ghostify modes in
+  
+  (* [pat] might already contain mode annotations (mostly due to mixing legacy
+  and new mode syntax), which is OK. Conceptually, we should create a new
+  [Ppat_mode_constraint] wrapping around the [pat]. But we are not allowed to
+  change parsetree. So the second best is probably to use [Ppat_alias] with mode
+  annotation wrapping around [pat].
+
+  Here, we cheat a bit, and simply merge with the exising mode annotation.
+  *)
```
ncik-roberts, line 186: This is complicated. I notice you just added the new syntax into `let_pattern` and `label_let_pattern`. Can we just return a tuple from these productions (where the second component is modes)? That would allow us to combine the mode expressions with the legacy modes by concatentation, rather than the current approach of blindly merging if the structure looks plausibly OK.

ocaml/testsuite/tests/typing-modes/syntax-error.compilers.reference
```diff
@@ -0,0 +1,123 @@
...
+2 | let foo = ("hello" @ );;
+                         ^
+Error: Syntax error
+val foo : bar:string -> string = <fun>
+Line 4, characters 19-20:
+4 |   foo ~(bar : _ @@ )
+                       ^
+Error: Syntax error
+Line 4, characters 14-15:
+4 |   foo ~(bar @ )
```
ncik-roberts, line 43: Many of these errors (e.g. all of these ones) do not in fact show improved error messages. Do you know why that is?

ncik-roberts (commented):
Finished.

I was surprised to not see val bindings in this PR. Was that intentional?

riaqn:
> I was surprised to not see val bindings in this PR. Was that intentional?
Yeah, because modalities on val bindings are not implemented in the type system yet, so it's hard to test.

lpw25:
I'd recommend adding all the syntax now in one go, because there is a reasonable overhead to each time we add syntax due to e.g. ocamlformat. You can just error out in the type-checker if someone uses the syntax, and that's enough to test that the new syntax is being parsed correctly.

ocaml/testsuite/tests/typing-modes/modes.ml
ncik-roberts: Per Leo's comment, it seems better to reject in type-checking rather than deleting the parse rules.

ocaml/testsuite/tests/typing-modes/modes.ml
```diff
@@ -57,3 +58,89 @@ Error: Unrecognized mode foo for coercion.
...
+               ^^^^^^^
+Error: This expression has type string but an expression was expected of type
+         'a list
+|}]
+
+let foo ~bar = bar ^ "hello"
+
+let x =
+  let bar = "world" in
+  foo ~(bar : _ @@ unique)
```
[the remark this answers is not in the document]
riaqn: I guess the concern is that this little-used part might bit-rot over time without being detected?

I did it simply for consistency. Basically, where you used to write `: ty`, you can also write `@ modes` or `: ty @@ modes`. This is implemented replacing all occurrences of `type_constraint` with `constraint_`. So I think in terms of maintainability it should be fine.

ocaml/testsuite/tests/typing-modes/modes.ml
```diff
@@ -57,3 +58,89 @@ Error: Unrecognized mode foo for coercion.
...
+Line 3, characters 8-11:
+3 |   foo ~(bar @ unique)
+            ^^^
+Error: Found a shared value where a unique value was expected
+|}]
+
+type r = {a : string; b : string}
+
+let r = {a : _ @@ unique = "hello";
+         b : _ @@ once = "world"}
```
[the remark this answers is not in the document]
riaqn: Indeed this is not in the design document. The reason I did this is similar to the one below: because it's more consistent and it's simple.

In term of confusion with modalities - modalities can only show up in type declarations - do we think the users will know that?

ocaml/parsing/parser.mly
```diff
@@ -4208,24 +4208,24 @@ function_type:
 strict_function_or_labeled_tuple_type:
   | mktyp(
       label = arg_label
-      arg_modes = mode_expr_legacy
-      domain = extra_rhs(param_type)
+      domain_with_modes = with_at_mode_expr(extra_rhs(param_type))
       MINUSGREATER
       codomain = strict_function_or_labeled_tuple_type
-        { Ptyp_arrow(label, mktyp_with_modes arg_modes domain , codomain) }
+        { let (domain, _), arg_modes = domain_with_modes in
```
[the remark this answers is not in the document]
riaqn: That's good point. I will take the second approach.

ocaml/testsuite/tests/typing-modes/modes.ml
riaqn: Unfortunately, `let x @ local local = ...` is parsed as mode coercion `let x = [%jane.modes local local] ...` which lacks duplication check. I just made some changes so it's detected.

ocaml/typing/typemode.ml
riaqn: I understand `local local` as assembling axes: Both `local` specify something in the locality axis, and you can't specify the locality axis twice. Sure, we can specially treat the case where the two specifications match, but I doubt it's worth the extra effort.

Modalities on the other hand are mappings. `global global` means `global (global (m))` where `m` is the mode of the record. It's silly but still sensible.

ocaml/parsing/parser.mly
riaqn: Usually I would try to factor out this, but is it still worth it when it's used only once? (in contrast to `constraint_` which is usually multiple times)

ocaml/parsing/parser.mly
ncik-roberts: I don't think future developers will wonder that. The fact that people can write mode aliases should make that clear.

ocaml/typing/typemode.ml
ncik-roberts: I don't think we should allow duplicate modalities until there exists a modality for which it is meaningful, and even then we should only allow it for the meaningful modalities.

ocaml/parsing/parser.mly
ncik-roberts: I think you should do whatever leads to simpler-feeling code. That might mean factoring out the action into a helper function, factoring out the two rules into a production that can share the same action, or leaving things as-is.

riaqn:
The commit "modality in val is parsed but not type checked" supports parsing modalities in value descriptions. Unfortunately, this requires some refactoring in `pprintast.ml` which make the commit larger than ideal. Apologies for the review inconvenience.

ocaml/parsing/parser.mly
riaqn: Since we put modes in `ppat_attributes` to means "this pattern should be of this mode", and expressions are dual to patterns, we should put modes in `pexp_attributes` to mean "this expression should be of this modes".

But I also understand this is a big change and would need some discussion.

ocaml/testsuite/tests/typing-modes/modes.ml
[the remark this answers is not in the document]
riaqn: As discussed on slack,  I will remove those parsing for now.

ocaml/testsuite/tests/typing-modes/syntax-error.compilers.reference
riaqn: OK I think it's because of inlining. I removed some inlining and it's giving better messages.

ocaml/typing/typemode.ml
riaqn: Some modalities are useful to duplicate. For example, the `regional` modality would be useful under https://github.com/ocaml-flambda/flambda-backend/pull/1815 : if a record is at region `n`, a field with `regional regional` would be at region `n-2`. 

The following is me thinking out loud, that @lpw25 might have opinions on:

Now, if a user writes `type r = {x : string @@ global regional regional}`, that gives field's mode `regional (regional (global (m)))`, where `m` is the record's mode. That should be equivalent to `global (m)`.

If a user writes `type r = {x  : string @@ regional regional global}`, that gives field's mode `global (regional (regional (m)))`, which is still equivalent to `global (m)`. 

Well, I guess my gist is that, modalities are mappings and could have rich behaviours. As such, redundant modality detection is likely to be ad-hoc and incomplete. Therefore, it's better to issue warnings instead of errors. Also, those programs are not "wrong" anyway.

riaqn: Well, on a second thought, maybe we should restrict user to specify only one modality per axis? That should be easy to do, and not too restrictive. The aforementioned `regional regional` will be replaced by `regional(2)`.

If we agree to this restriction, I will be OK with taking violation of it as errors.

lpw25: I agree that there are modalities that are not idempotent and pairs of modalities that do not commute. Possibly we should avoid supporting such modalities of course, because they are difficult to think about for the user. I agree that warnings for redundent modalities is the way to go -- although I think we don't really need that in the MVP.

Whether redundent modes should be a warning or an error depends on the details of our full mode expression language. In some approaches it is essentially a kinding error to repeat a mode, in other approaches it is merely redundant. I'm leaning towards an approach that makes it a kinding error, but I also think we can just do nothing for the MVP and address this when we add mode polymorphism.

ocaml/parsing/parser.mly
ncik-roberts: I think I'm just arguing to delete this CR as the point is not that important. If it ever becomes important to make the attribute encoding the same, we can make that change then.

riaqn:
@ncik-roberts I think it's ready for review now. The last missing piece is the syntax:
```
let foo a b @ local = "hello"
let foo a b : _ @@ local = "hello"
```
which I think requires refactoring `Jane_syntax.N_ary_functions`; in particular, we will need to extend 
```  
type function_constraint =
    { mode_annotations : Mode_expr.t;
      type_constraint : type_constraint
    }
```
with another field `mode_constraint` which is the mode of the return value (where `mode_annotations` is the mode of the function). I'm slightly worried that this is too big of a change and might break things. What do you think? Is it OK to leave it out for now?

ocaml/typing/typecore.ml
```diff
@@ -2795,7 +2800,7 @@ and type_pat_aux
   | Ppat_constraint(sp_constrained, sty) ->
       (* Pretend separate = true *)
       let cty, ty, expected_ty' =
-        let type_modes = alloc_mode_from_pat_attrs sp in
+        let type_modes, sty = alloc_mode_from_typ_attrs sty in
```
ncik-roberts, line 2803: I fear that putting the attributes on the type is incompatible with a small detail of #2335: in that PR, we start tracking the mode in `let local_ x : int -> int = ...` on the type in "value binding elaboration" in the type-checker. So I'd like to take a look at this after you resolve conflicts and before we merge into `main`.

ncik-roberts (commented):
Looks good, but there's something I want to take a look at after you pull in the changes from `main`.

ncik-roberts (approved):
Actually my comment was a false alarm, I think this looks fine. Though I'll be around and I am very happy to take a look at the post-conflict-fixing result, especially because I expect some conflicts with #2335

ocaml/typing/typedecl.ml
```diff
@@ -2387,6 +2387,9 @@ let error_if_containing_unexpected_jkind prim env cty ty =
 
 (* Translate a value declaration *)
 let transl_value_decl env loc valdecl =
+  match Jane_syntax.Mode_expr.maybe_of_attrs valdecl.pval_type.ptyp_attributes with
+  | Some _, _ -> Misc.fatal_error "Modalities not supported in value declarations"
```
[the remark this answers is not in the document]
riaqn: Yeah technically the user can hit this (as demonstrated by the test cases). But since we will very soon support this, I just felt that defining an error is too heavy.

(You comment below suggests that you are no longer concerned by this? If so, I'm resolving)

riaqn:
@ncik-roberts 
It's now rebased and ready for another review. During rebasing, here is the list of non-trivial conflicts:
(All line numbers are in the after-conflict-fixing files)
- `typecore.ml`, L4962-4972
- `pprintast.ml`, probably the whole file.

ocaml/typing/typedecl.ml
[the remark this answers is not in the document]
ncik-roberts: Ah, sorry, my "never mind" comment was with respect to another comment.

It's only like 3 more lines of code to do a proper error message, I think we just should do it. Stack traces like this make people think they should report bugs to us.

ocaml/typing/typecore.ml
```diff
@@ -4951,64 +4960,62 @@ let may_lower_contravariant_then_generalize env exp =
   if maybe_expansive exp then lower_contravariant env exp.exp_type;
   generalize exp.exp_type
 
+(* This added mode attribute is read and removed by
+    [alloc_mode_from_pexp_constraint_typ_attrs] or
+    [alloc_mode_from_pexp_constraint_typ_attrs]. *)
```
ncik-roberts, line 4965:
```suggestion
    [alloc_mode_from_pexp_constraint_typ_attrs] or
    [alloc_mode_from_ppat_constraint_typ_attrs]. *)
```

ncik-roberts (approved):
Looks good to me! Please merge after adding the error messages.

ocaml/typing/typecore.ml
riaqn: Yeah, not only they are compatible, both changes are driven by the same idea: the said modes is only for the interpretation of the type in `Pexp_constraint` or `Ppat_constraint`, so the modes should be in `ptyp_attributes` instead of `pexp_attributes` or `ppat_attributes`.

ocaml/typing/typemode.ml
riaqn: For the record, I asked around, and all warnings are enabled by default in our internal build, unless specified otherwise. So this should be fine.

PR #2237, opened 2024-01-19
Fix ordering of masterlock operations

mshinwell:
These orderings didn't match runtime4, and on the `Async_unix` tests seem to bring performance much closer to runtime4.  The fix in `st_masterlock_release` seems maybe the most important.

### DIFF
--- a/ocaml/otherlibs/systhreads/st_pthreads.h
+++ b/ocaml/otherlibs/systhreads/st_pthreads.h
@@ -173,8 +173,8 @@ static void st_masterlock_release(st_masterlock * m)
   m->busy = 0;
   // CR ocaml 5 domains: we assume no backup thread
   // st_bt_lock_release(m);
-  custom_condvar_signal(&m->is_free);
   pthread_mutex_unlock(&m->lock);
+  custom_condvar_signal(&m->is_free);
 
   return;
 }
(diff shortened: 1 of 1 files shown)


ocaml/otherlibs/systhreads/st_pthreads.h
```diff
@@ -173,8 +173,8 @@ static void st_masterlock_release(st_masterlock * m)
   m->busy = 0;
   // CR ocaml 5 domains: we assume no backup thread
   // st_bt_lock_release(m);
-  custom_condvar_signal(&m->is_free);
   pthread_mutex_unlock(&m->lock);
+  custom_condvar_signal(&m->is_free);
```
stedolan, line 177: This change looks great: in a 2-thread case it means that the signal will fire _after_ the owning thread has released the lock, so there's no risk of the woken thread immediately sleeping again to wait for the lock.

ocaml/otherlibs/systhreads/st_pthreads.h
```diff
@@ -204,8 +204,8 @@ Caml_inline void st_thread_yield(st_masterlock * m)
   }
 
   m->busy = 0;
-  atomic_fetch_add(&m->waiters, +1);
   custom_condvar_signal(&m->is_free);
+  atomic_fetch_add(&m->waiters, +1);
```
stedolan, line 208: This change looks bad: in a 2-thread case it means that there's _more_ work between the owning thread signalling and waiting, _increasing_ the chances that the woken thread immediately sleeps again to wait for the lock.

stedolan (changes requested):

ocaml/otherlibs/systhreads/st_pthreads.h
mshinwell: Indeed, that was my thinking.

stedolan (approved):

PR #2239, opened 2024-01-23
Fix newtype loc ghost

alanechang:
Prior to this PR, `let f (type a b c) = ();;` produces the following parsetree:

```
...
          pattern (//toplevel//[1,0+4]..[1,0+5])
            Ppat_var "f" (//toplevel//[1,0+4]..[1,0+5])
          expression (//toplevel//[1,0+6]..[1,0+23]) ghost
            Pexp_newtype "a"
            expression (//toplevel//[1,0+6]..[1,0+23])
              Pexp_newtype "b"
              expression (//toplevel//[1,0+6]..[1,0+23])
                Pexp_newtype "c"
                expression (//toplevel//[1,0+21]..[1,0+23])
                  Pexp_construct "()" (//toplevel//[1,0+21]..[1,0+23])
                  None
      ]
  ]
```

Here the newtype expressions after the first one are not marked as ghost. This differs from upstream and leads to `let f (type (a : immediate) (b : immediate)) = ()` failing the ppxlib location overlap check.

This PR fixes this by marking all `Pexp_newtype` expressions above as ghost.


ocaml/testsuite/tests/parsetree/locations_test.ml
```diff
@@ -132,3 +132,7 @@ let g y =
   let f ~y = y + 1 in
   f ~(y:int)
 ;;
+
+(* Newtype *)
+let f (type a b c d e) = ()
+;;
```
ncik-roberts, line 138: Could you add a test using the `fun` syntax as well?

ocaml/parsing/parser.mly
```diff
@@ -858,7 +858,7 @@ let mkghost_newtype_function_body newtypes body_constraint body ~loc =
         let body = Exp.mk (mkexp_desc_constraint body type_constraint) ~loc in
         exp_with_modes loc mode_annotations body
   in
-  mk_newtypes ~loc newtypes wrapped_body
+  mk_newtypes ~loc:(ghost_loc loc) newtypes wrapped_body
```
ncik-roberts, line 861: As discussed, I don't think the outermost node should be ghost if the user used the `fun` syntax.

ncik-roberts (commented):
A quick comment we discussed in person as well.

ocaml/parsing/parser.mly
alanechang: Fixed this by changing the `mk_newtypes` function.

It also affects the parsing of `let a : type a b c. unit = ()` which we do differently from upstream to support layout annotation. Now the outermost newtype expression is not marked as ghost compared to before.

We can't return just the `pexp_desc` like the upstream implementation due to layout annotation being encoded as attributes.

ocaml/testsuite/tests/parsetree/locations_test.ml
alanechang: test case added

alanechang:
This needs another bootstrap before merging

ocaml/parsing/parser.mly
```diff
@@ -634,14 +634,21 @@ let pat_of_label lbl =
...
   in
-  List.fold_right mk_one newtypes exp
+  (* This is doing a fold right while making sure only
+     the outermost expression has non-ghost location *)
+  let rec inner acc = function
+    | x :: ((_ :: _) as xs) -> inner (mk_one (ghost_loc loc) x acc) xs
+    | [x] -> mk_one (make_loc loc) x acc
+    | [] -> acc
+  in
+  inner exp (List.rev newtypes)
```
ncik-roberts, line 651: Perhaps it would be simpler to construct all nested nodes with a ghost expression, and then set the location on the outermost node to non-ghost.

ocaml/parsing/parser.mly
ncik-roberts: To make sure I'm interpreting you right: the fact you mention (that the newtype node in `let a : type a b c. unit = ()` is not ghost) is not something introduced by this PR, right?

alanechang: My bad that part is not correct. To clearify, the parsing of `let a : type a b c. unit = ()` is already different from upstream before this PR. In our original implementation before this PR, It's parsed as:

```
Ptop_def
  [
    structure_item (//toplevel//[1,0+0]..[1,0+29])
      Pstr_value Nonrec
      [
        <def>
          pattern (//toplevel//[1,0+4]..[1,0+24]) ghost
            Ppat_constraint
            pattern (//toplevel//[1,0+4]..[1,0+5])
              Ppat_var "a" (//toplevel//[1,0+4]..[1,0+5])
            core_type (//toplevel//[1,0+20]..[1,0+24]) ghost
              Ptyp_poly 'a 'b 'c
              core_type (//toplevel//[1,0+20]..[1,0+24])
                Ptyp_constr "unit" (//toplevel//[1,0+20]..[1,0+24])
                []
          expression (//toplevel//[1,0+4]..[1,0+29])
            Pexp_newtype "a"
            expression (//toplevel//[1,0+4]..[1,0+29])
              Pexp_newtype "b"
              expression (//toplevel//[1,0+4]..[1,0+29])
                Pexp_newtype "c"
                expression (//toplevel//[1,0+4]..[1,0+29])
                  Pexp_constraint
                  expression (//toplevel//[1,0+27]..[1,0+29])
                    Pexp_construct "()" (//toplevel//[1,0+27]..[1,0+29])
                    None
                  core_type (//toplevel//[1,0+20]..[1,0+24])
                    Ptyp_constr "unit" (//toplevel//[1,0+20]..[1,0+24])
                    []
      ]
  ]
```

After this PR, it will be parsed as:

```
Ptop_def
  [
    structure_item (//toplevel//[1,0+0]..[1,0+29])
      Pstr_value Nonrec
      [
        <def>
          pattern (//toplevel//[1,0+4]..[1,0+24]) ghost
            Ppat_constraint
            pattern (//toplevel//[1,0+4]..[1,0+5])
              Ppat_var "a" (//toplevel//[1,0+4]..[1,0+5])
            core_type (//toplevel//[1,0+20]..[1,0+24]) ghost
              Ptyp_poly 'a 'b 'c
              core_type (//toplevel//[1,0+20]..[1,0+24])
                Ptyp_constr "unit" (//toplevel//[1,0+20]..[1,0+24])
                []
          expression (//toplevel//[1,0+4]..[1,0+29])
            Pexp_newtype "a"
            expression (//toplevel//[1,0+4]..[1,0+29]) ghost
              Pexp_newtype "b"
              expression (//toplevel//[1,0+4]..[1,0+29]) ghost
                Pexp_newtype "c"
                expression (//toplevel//[1,0+4]..[1,0+29])
                  Pexp_constraint
                  expression (//toplevel//[1,0+27]..[1,0+29])
                    Pexp_construct "()" (//toplevel//[1,0+27]..[1,0+29])
                    None
                  core_type (//toplevel//[1,0+20]..[1,0+24])
                    Ptyp_constr "unit" (//toplevel//[1,0+20]..[1,0+24])
                    []
      ]
  ]
```

Difference is that the original version doesn't have ghost on any of the `Pexp_newtype` expressions and now we put  ghost on all of them except the outermost one. This feels like an improvement to me. Let me know if you disagree.

ncik-roberts: That sounds ok to me.

ocaml/parsing/parser.mly
alanechang: Yeah, I was hoping not to manipulate the location directly but the code above is not great either. I made the suggested change.

ncik-roberts (approved):
LGTM. Thanks for fixing this.

PR #2240, opened 2024-01-24
Add missing case for `Ltyp_poly` in Pprintast

alanechang:
`module type S = sig val f1 : ('a : float64) ('b : immediate) 'c. 'a -> 'b -> 'c end;;` currently causes an infinite loop in `Pprintast`.

This is due to a missing case match for `Ltyp_poly`. This PR adds the missing case, removes the use of wildcard match cases, and provides regression tests.


ncik-roberts (approved):
I didn't scrutinize the code too closely, but I checked that its changes are localized to Jane Street syntax and that the test cases look good.

PR #2241, opened 2024-01-25
Turn the (most mature) layouts extension on by default

ccasin:
This also updates various tests accordingly.


ocaml/testsuite/tests/typing-layouts-float64/unboxed_floats.ml
```diff
@@ -10,16 +10,13 @@
    ** bytecode
      flags = "-extension layouts_beta"
    ** native
-     flags = "-extension layouts"
    ** bytecode
-     flags = "-extension layouts"
    ** setup-ocamlc.byte-build-env
      ocamlc_byte_exit_status = "2"
    *** ocamlc.byte
+     flags = "-disable-all-extensions"
```
goldfirere, line 17: Can we also test with `-only-erasable-extensions`? That's what's used in public release, and should stop layouts from working there.

goldfirere (approved):

ocaml/testsuite/tests/typing-layouts-float64/unboxed_floats.ml
ccasin: The layouts extension is erasable.

(I don't actually know if we have the tooling in place to do the erasing, but if not, I think that bug is independent from the change in this PR.)

ocaml/testsuite/tests/typing-layouts-bits32/basics.ml
```diff
@@ -4,7 +4,6 @@
    * expect
    flags = "-extension layouts_beta"
    * expect
-   flags = "-extension layouts"
```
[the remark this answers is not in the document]
ccasin: I did test this and it works, but I've now changed the tests to put the no-flags version first to make it clearer.

ocaml/testsuite/tests/parsing/pr6604.compilers.reference
```diff
@@ -1,4 +1,4 @@
 File "pr6604.ml", line 9, characters 0-2:
 9 | #1
     ^^
-Error: Syntax error: line number directive not expected.
+Error: Syntax error: unboxed integer literal with type-specifying suffix expected.
```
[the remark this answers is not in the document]
ccasin: Unfortunately, this error is issued not only at the beginning of lines, but wherever this appears.

I think adding a hint about line number directives is confusing for the case where someone just forgets a suffix on their unboxed int literal (which, in the not-too-distant future will hopefully be much more common than line number directives).

It might conceivably be possible to distinguish the beginning of line case, but I think not without some lexer trickery, and propose it is not worth investing further time in right now.

I have updated the error along the lines you suggest, though.

ocaml/testsuite/tests/typing-layouts-float64/unboxed_floats.ml
goldfirere: That sounds bad. I agree it's separate from this PR. But I'll raise the issue elsewhere.

goldfirere:
As discussed internally, we should mark `-extension layouts` as not erasable. Maybe just do that in this PR? Seems germane.

ccasin:
> As discussed internally, we should mark `-extension layouts` as not erasable. Maybe just do that in this PR? Seems germane.

Done.  We're good to press the button once CI is green again.

PR #2242, opened 2024-01-25
Add my_region check upon set of closures creation in classic mode

mshinwell:
This adds a check that there is no use of the implicit `my_region` parameter if a function's return mode is Heap, when in classic mode.  This mirrors the existing check performed in Simplify.  This check is helpful for catching errors in code generated by the frontend, where failure to do so may lead to a local stack leak.

I tested this check using the example from #2236.

The second commit declutters the printing by eliding the printing of the `my_region` parameter if the return mode is Heap.


Ekdohibs (approved):
First commit is good. I don't like much the approach of the second commit, which introduces a dependency between the code metadata and the printing of `function_params_and_body`, but it works; a better solution would be to properly account for the `my_region` parameter by replacing it with an option.

mshinwell:
OK let's just do the first commit for now then.

PR #2244, opened 2024-01-26
Refine assertion in caml_modify_local

mshinwell:
It is unfortunately the case that one can end up with `caml_modify_local` calls on blocks outside of the OCaml heap.  This trips an assertion when using the debug runtime.  For the moment I'd like to disable this assertion, since the semantics of the function are correct for external values, and decide later how we should proceed in the medium term.


stedolan:
I'm not convinced that this change is correct - generally, using this function on out-of-heap values is an error.

I think we could safely weaken the assertion to the following, though, which would probably be enough for your purposes:

        CAMLassert(caml_is_stack(obj) || (!Is_block(val) && !Is_block(Field(obj, i)));

stedolan (approved):

PR #2245, opened 2024-01-26
Printing new mode syntax

riaqn:
This is based on https://github.com/ocaml-flambda/flambda-backend/pull/2235

This PR switches the printing to the new mode syntax as well. 

Currently still a draft, because our way of encoding makes the round trip "parsetree -> .ml -> parsetree" not easy to achieve. In particular: `exp : _ @ modes` is parsed as `Pexp_apply (Pexp_extension "jane.erasable.mode" modes) (Pexp_constraint (exp, Ptyp_any)`, which is then printed as `(exp : _) : _ @ modes`, which is re-parsed differently. I can workaround this by some extra check, but I'd like a more principled approach.


riaqn:
Outdated and no longer relevant.

PR #2246, opened 2024-01-26
Adjust delayed layout checks in typedecl to fix a kind inference bug.

ccasin:
The first commit adds a buggy test and the second fixes it.

This is the conclusion of a long discussion in slack, which I won't attempt to fully recreate here.  I tried to capture the most relevant parts in comments.


ocaml/testsuite/tests/typing-layouts/basics.ml
```diff
@@ -2331,3 +2331,39 @@ Error: This expression has type t_float64
...
+type t1 = string t2 as (_ : immediate)
+and 'a t2 = 'a
+
+[%%expect{|
+Line 2, characters 0-14:
+2 | and 'a t2 = 'a
+    ^^^^^^^^^^^^^^
+Error:
+       The layout of 'a t2 is value, because
+         it instantiates an unannotated type parameter of t2, defaulted to layout value.
```
goldfirere, line 2348: This doesn't look right to me: I don't see an instantiation of this sort.

Really, the error message should mention `string` somewhere, as that's the root of the problem. (Although actually with the conservative approach, maybe this definition would fail even with `int t2` there.)

ocaml/testsuite/tests/typing-layouts/basics.ml
```diff
@@ -2331,3 +2331,39 @@ Error: This expression has type t_float64
...
+   parameter was just value (a bug). *)
+type 'a t1 = 'a t2 as (_ : immediate)
+and 'a t2 = 'a
+
+[%%expect{|
+Line 2, characters 0-14:
+2 | and 'a t2 = 'a
+    ^^^^^^^^^^^^^^
+Error:
+       The layout of 'a t2 is value, because
```
goldfirere, line 2365: This message would be better to admit that the compiler is not clever enough to accept the definition. Don't know how hard that would be to arrange.

ocaml/typing/typedecl.ml
```diff
@@ -1757,7 +1757,15 @@ let transl_type_decl env rec_flag sdecl_list =
...
   List.iter (fun (checks,loc) ->
     List.iter (fun (ty,jkind) ->
-      match Ctype.constrain_type_jkind new_env ty jkind with
+      (* The use [check_type_jkind] rather than [constrain_type_jkind] here is
+         conservative. It ensures that the delayed checks don't succeed by
+         mutating type variables from the [temp_env] in a way that won't be
+         reflected in the final type decls and may be incompatible with them.
+         An alternative would be to beef up [check_constraints] and really make
+         sure we re-check any kind constraint that could arise from translating
+         the typedecl RHSs. See Test 41 in [tests/typing-layouts/basics.ml] for
```
goldfirere, line 1766:
```suggestion
         the typedecl RHSs, for example by looking at Typedtree instead of just what's in the type environment. See Test 41 in [tests/typing-layouts/basics.ml] for
```

ocaml/typing/typedecl.ml
```diff
@@ -1757,7 +1757,15 @@ let transl_type_decl env rec_flag sdecl_list =
...
+         a subtle example. *)
+      match Ctype.check_type_jkind new_env ty jkind with
```
goldfirere, line 1768: To improve error messages, it might be better to use `constraint_type_jkind` here but then check whether any mutation happened. That shouldn't be too hard: look at `ty` before the call. If it's a variable, remember its kind. Then check again after the call. That way, we can report a message that complains about how our inference algorithm is not really as clever as it could be.

ocaml/testsuite/tests/typing-layouts/basics.ml
ccasin: I agree this error message isn't ideal, but it is already issued in other similar circumances, e.g., [here](https://github.com/ocaml-flambda/flambda-backend/blob/4f2310021bcf28bccdada7d6f2048446cd5a60b2/ocaml/testsuite/tests/typing-layouts-float64/basics.ml#L293).

I think it would be a mistake for the error to mention `string` exactly because the same problem occurs with `int` (I've added that as a test).

I've adjusted this to print a message similar to the `check_constraint` message - see below.  (Though I'm also not totally happy with that since in fact this program is genuinely incorrect and can't be fixed with an annotation.)

ocaml/testsuite/tests/typing-layouts/basics.ml
ccasin: Arranged - see below.

ocaml/typing/typedecl.ml
ccasin: I think this is not so simple.  We might need to expand `ty` and deal with `@@unboxed` types in the mutually recursive group to see it is a variable.  And doing that expansion properly in the middle of recursive typedecl checking is just why `constrain_type_jkind` is so complicated.

So I did it a different way: if `check_type_jkind` fails, I try `constrain_type_jkind`.  If that succeeds, we know we're in the case where this PR changes the behavior. I print an error similar to the one issued for the other case in typedecl where the compiler is insufficiently clever.  I'm not sure this relatively rare case justifies the added complexity in typedecl, and as noted above this is sometimes not only the compiler being stupid, but see what you think.  (I'm not worried about the cost of checking this twice since it only happens on the error path)

ocaml/testsuite/tests/typing-layouts/basics.ml
```diff
@@ -2331,3 +2331,72 @@ Error: This expression has type t_float64
...
+
+(* This example must be rejected. *)
+type t1 = string t2 as (_ : immediate)
+and 'a t2 = 'a
+
+[%%expect{|
+Line 2, characters 0-14:
+2 | and 'a t2 = 'a
+    ^^^^^^^^^^^^^^
+Error: Layout mismatch in delayed kind checks for mutually recursive groups.
```
goldfirere, line 2346: "delayed kind checks" seems to leak some internal details. I think I'd prefer

```suggestion
Error: Layout mismatch in when checking consistency of mutually recursive groups.
```

ocaml/testsuite/tests/typing-layouts/basics.ml
```diff
@@ -2331,3 +2331,72 @@ Error: This expression has type t_float64
...
+       This is most often caused by the fact that type inference is not
+       clever enough to propagate layouts through variables in different
+       declarations. It is also not clever enough to produce a good error
+       message, so we'll say this instead:
+         The layout of 'a t2 is value, because
+           it instantiates an unannotated type parameter of t2, defaulted to layout value.
+         But the layout of 'a t2 must be a sublayout of immediate, because
+           of the annotation on the wildcard _ at line 1, characters 28-37.
+       The fix will likely be to add a layout annotation on a parameter to
```
goldfirere, line 2355: Good point about there not being a fix here. Maybe

```suggestion
       A good next step is to add a layout annotation on a parameter to
```

?

ocaml/typing/typedecl.ml
goldfirere: Yes, good points about the difficulty. Happy with your approach.

goldfirere (approved):

PR #2247, opened 2024-01-26
128-bit load/store primitives for GC'd arrays

TheNumbat:
Adds 128-bit vector load/store primitives for GC-managed arrays. All operations are unaligned, since GC blocks are only 8 byte aligned. The safe versions of the primitives perform a bounds check.

- `float64x2` to/from `float array` (requires flat float arrays), `floatarray`, and `float# array`
- `int64x2` to/from `int array` (or any array of immediates), `int64# array`, and `nativeint# array`. 
     - Note that loading from an immediate array returns the raw tagged values, and storing writes potentially untagged values. The safe operations presented to users will automatically tag/untag the values in OCaml. I didn't add tag/untag to the primitives because it uses SIMD intrinsics that are not exposed to the middle end.
- `int32x4` to/from `int32# array`

Tests were added for `float array`, `floatarray`, and `int array`, but we need frontend unboxed array support (#2238) before testing the unboxed variants. The post-frontend codepaths are identical, so these should just work.

After I implement unboxed vector types, I'll also update these primitives to support them.


middle_end/flambda2/simplify/simplify_binary_primitive.ml
```diff
@@ -967,6 +967,31 @@ let simplify_array_load (array_kind : P.Array_kind.t) mutability dacc
     let dacc = DA.add_variable dacc result_var ty in
     SPR.create named ~try_reify:false dacc
 
+let simplify_array_vector_load (vec_kind : Lambda.boxed_vector)
```
TheNumbat, line 970: Not sure if this function is necessary, since vector loads should only ever operate on monomorphic arrays.

middle_end/flambda2/simplify/simplify_ternary_primitive.ml
```diff
@@ -55,9 +55,19 @@ let simplify_array_set (array_set_kind : P.Array_set_kind.t) dacc ~original_term
       | Naked_nativeints -> Naked_nativeints
     in
     let named =
-      Named.create_prim
-        (Ternary (Array_set array_set_kind, array, index, new_value))
-        dbg
+      match vec_kind with
```
TheNumbat, line 58: Same here

middle_end/flambda2/terms/flambda_primitive.ml
```diff
@@ -1295,10 +1317,11 @@ type binary_primitive =
   | Bigarray_get_alignment of int
   | Atomic_exchange
   | Atomic_fetch_and_add
+  | Array_vector_load of Lambda.boxed_vector * Array_kind.t * Mutability.t
```
Gbury, line 1320: Why is this a separate primitive instead of having the `Array_load` primitive take a `array_accessor_width` payload similar to the string/bigstring case ?

In any case, it seems that having both the `boxed_vector` and the `Array_kind.t` is redundant (and possibly dangerous if/when the two disagree), and since the arrya kind can be computed from the `boxed_vector` I'd say to only keep that one.

middle_end/flambda2/to_cmm/to_cmm_primitive.ml
```diff
@@ -151,6 +151,38 @@ let array_load ~dbg (kind : P.Array_kind.t) ~arr ~index =
   | Naked_floats -> C.unboxed_float_array_ref arr index dbg
   | Naked_int32s -> C.unboxed_int32_array_ref arr index dbg
 
+let array_vector_load ~dbg (vec_kind : Lambda.boxed_vector)
+    (array_kind : P.Array_kind.t) ~arr ~index =
+  P.Array_kind.check_vector_access_kind array_kind vec_kind;
+  let offset =
```
Gbury, line 157: maybe rename that to a more explicit name, such as `element_width_log2` ? Also, it took a bit of time for @lthls and I to understand the semantics of these array operations, it'd be nice if they were documented somewhere (for instance in the primitive file, or in a separate doc somewhere)

middle_end/flambda2/terms/flambda_primitive.ml
```diff
@@ -1512,11 +1549,12 @@ type ternary_primitive =
   | Bytes_or_bigstring_set of bytes_like_value * string_accessor_width
   | Bigarray_set of num_dimensions * Bigarray_kind.t * Bigarray_layout.t
   | Atomic_compare_and_set
+  | Array_vector_set of Lambda.boxed_vector * Array_set_kind.t
```
Gbury, line 1552: Same remark as above, do we really need the `Array_set_kind.t` here, or can can we always recompute it from the `boxed_vector` info ?

middle_end/flambda2/to_cmm/to_cmm_primitive.ml
```diff
@@ -151,6 +151,38 @@ let array_load ~dbg (kind : P.Array_kind.t) ~arr ~index =
...
+  C.unaligned_load_128 arr
+    (C.lsl_int (C.untag_int index dbg) (Cconst_int (offset, dbg)) dbg)
+    dbg
+
+let array_vector_set ~dbg (vec_kind : Lambda.boxed_vector)
+    (array_kind : P.Array_set_kind.t) ~arr ~index ~new_value =
+  P.Array_kind.check_vector_access_kind
+    (P.Array_set_kind.array_kind array_kind)
+    vec_kind;
+  let offset =
```
Gbury, line 174: same variable renaming as above.

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
```diff
@@ -544,6 +544,62 @@ let bytes_like_set_safe ~dbg ~size_int ~access_size kind bytes index new_value =
...
+      assert false
+  in
+  let width = H.Simple (Simple.const_int (Targetint_31_63.of_int width)) in
+  [ H.Binary
+      ( Int_comp (Tagged_immediate, Yielding_bool (Lt Unsigned)),
+        index,
+        Prim
+          (Binary
+             ( Int_arith (I.Tagged_immediate, Sub),
+               Prim (Unary (Array_length (Array_kind array_kind), array)),
```
Gbury, line 567: you're missing a `max_with_zero` call here to avoid underflow of the array length computation, similar to what is done in the `actual_max_length_for_string_like_access` function

middle_end/flambda2/simplify/simplify_binary_primitive.ml
Gbury: well even in that case, it can be useful to remove dead code branches based on the `Bottom` case.

Gbury (changes requested):

ocaml/lambda/lambda.ml
```diff
@@ -1675,6 +1709,14 @@ let primitive_result_layout (p : primitive) =
       layout_boxedint Pint64
   | Pstring_load_128 _ | Pbytes_load_128 _ | Pbigstring_load_128 _ ->
       layout_boxed_vector (Pvec128 Int8x16)
+  | Pfloatarray_load_128 _ | Pfloat_array_load_128 _
+  | Punboxed_float_array_load_128 _ ->
+    layout_boxed_vector (Pvec128 Float64x2)
+  | Pint_array_load_128 _ | Punboxed_int64_array_load_128 _
+  | Punboxed_nativeint_array_load_128 _ ->
+    layout_boxed_vector (Pvec128 Int64x2)
```
Gbury, line 1717: This line probably deserves a comment/CR/assertion that a `nativeint` has size `64`

middle_end/flambda2/parser/fexpr.ml
```diff
@@ -353,12 +353,14 @@ type binop =
   | Infix of infix_binop
   | String_or_bigstring_load of string_like_value * string_accessor_width
   | Bigarray_get_alignment of int
+  | Array_vector_load of Lambda.boxed_vector * array_kind * mutability
```
Gbury, line 356: this also needs changing if we decide to change how `128`-bits loads are represented in flambda2 primitives

middle_end/flambda2/terms/code_size.ml
```diff
@@ -363,6 +363,7 @@ let binary_prim_size prim =
   match (prim : Flambda_primitive.binary_primitive) with
   | Block_load (kind, _) -> block_load kind
   | Array_load (kind, _mut) -> array_load kind
+  | Array_vector_load (_vec, kind, _mut) -> array_load kind
```
Gbury, line 366: not sure it should be the same size as an array load, I'd think that the `128`-bits load instructions take more space than regular word-sized ones ?

middle_end/flambda2/terms/code_size.ml
```diff
@@ -385,6 +386,7 @@ let ternary_prim_size prim =
   match (prim : Flambda_primitive.ternary_primitive) with
   | Block_set (block_access, init) -> block_set block_access init
   | Array_set kind -> array_set kind
+  | Array_vector_set (_vec, kind) -> array_set kind
```
Gbury, line 389: same remark here, the actual instructions for 128-bits might be bigger than the ones regularly used for arrays.

middle_end/flambda2/terms/code_size.ml
TheNumbat: The encoding of `movupd` (m128<->xmm) is the same size as `movq` (m64<->xmm)

They're actually both a bit larger than `mov` on integer regs, but it doesn't look like code_size represents size at that granularity

ocaml/lambda/lambda.ml
TheNumbat: Is it possible to check `Arch.size_int` in lambda? I didn't see an obvious way. 
In any case, 128-bit operations are only supported in the x86_64 backend, so this should always be true.

mshinwell: Probably not, but you could check this in `lambda_to_flambda_primitives.ml`, with the `Arch.size_int` value having been passed through from `flambda2.ml`.  (The latter file, together with the `To_cmm` files, have access to `ocamloptcomp` and hence `Arch`, whereas the remainder of Flambda 2 does not.)

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
TheNumbat: Fixed

middle_end/flambda2/to_cmm/to_cmm_primitive.ml
TheNumbat: Renamed

ocaml/lambda/lambda.ml
TheNumbat: Added

middle_end/flambda2/terms/flambda_primitive.ml
TheNumbat: Now merged into `Array_load` / `Array_set`

middle_end/flambda2/parser/fexpr.ml
TheNumbat: (Refactored)

middle_end/flambda2/terms/code_size.ml
TheNumbat: (Same remark)

middle_end/flambda2/terms/code_size.ml
Gbury: Ok, i'm good with that explanation, :)

middle_end/flambda2/to_cmm/to_cmm_primitive.ml
mshinwell: Did this question about documentation get resolved?

Gbury: I don't think so. It'd probably deserves a file in the `docs` folder.

TheNumbat: I've added `middle_end/flambda2/docs/simd.md`, that sufficient for now?

Gbury (approved):

middle_end/flambda2/to_cmm/to_cmm_primitive.ml
Gbury: that should be good for now

PR #2248, opened 2024-01-29
Allocate pools for the shared heap in batches rather than one at a time

stedolan:
This patch changes the major heap allocator to request memory for new pools in large batches rather than one pool at a time. It currently grows the heap by 15% of the current size each time, with a minimum of 8 new pools.

The 15% number is used because that's the default in OCaml 4 - in a future patch, I'll restore the plumbing from OCaml 4 to make this configurable using the `major_heap_increment` / `i` parameter.

In programs with rapidly-growing heaps, this speeds things up by reducing the amount of time spent kernel-side in `mmap` calls. In a [synthetic benchmark](https://gist.github.com/stedolan/06fb028a6d6b3be38a67433d4f9ec7d5), which allocates more than it collects (i.e. `promotion_rate` high, `collection_rate` below 100) and grows the heap to about a gigabyte, this patch improves performance:
```
runtime4:
            time (s): 3.98
    heap before (mb): 109
     heap after (mb): 1558
  heap growth (mb/s): 364
              minors: 445
              majors: 8

runtime5:
            time (s): 4.04
    heap before (mb): 106
     heap after (mb): 1579
  heap growth (mb/s): 365
              minors: 445
              majors: 8

runtime5+this:
            time (s): 3.71
    heap before (mb): 106
     heap after (mb): 1579
  heap growth (mb/s): 397
              minors: 445
              majors: 8
```

The time difference is mostly explained by spending less time in the kernel (i.e. system time improves much more than user).

The large allocations also seem to improve TLB performance on this benchmark. Here are the TLB walk counts from a server-class Intel Skylake processor:

```
runtime4:
           417,452      dtlb_load_misses.walk_completed_4k                                   
           179,418      dtlb_store_misses.walk_completed_4k                                   
            15,186      dtlb_load_misses.walk_completed_2m_4m                                   
             1,597      dtlb_store_misses.walk_completed_2m_4m                                   

runtime5:
         1,714,417      dtlb_load_misses.walk_completed_4k                                   
         2,648,033      dtlb_store_misses.walk_completed_4k                                   
            41,662      dtlb_load_misses.walk_completed_2m_4m                                   
             2,221      dtlb_store_misses.walk_completed_2m_4m                                   

runtime5+this:
           303,186      dtlb_load_misses.walk_completed_4k                                   
           156,856      dtlb_store_misses.walk_completed_4k                                   
            23,515      dtlb_load_misses.walk_completed_2m_4m                                   
             1,603      dtlb_store_misses.walk_completed_2m_4m                                   
```

However, the current implementation of compaction frees pools in an arbitrary pattern. This will make holes in the large allocation blocks, so some of these benefits may disappear in programs which compact. (The benchmark measured here does not).

### DIFF
--- a/ocaml/runtime/shared_heap.c
+++ b/ocaml/runtime/shared_heap.c
@@ -74,8 +74,10 @@ static struct {
   uintnat fresh_pools;
   char* next_fresh_pool;
 
-  /* Counts pools in some domain's heap, or in the global list below.
-     Does not count free or fresh pools above */
+  /* Count of all pools in use across all domains and the global lists below.
+
+     Does not include unused pools ('free' above) or freshly allocated pools
+     ('next_fresh_pool' above). */
   uintnat active_pools;
 
   /* these only contain swept memory of terminated domains*/
--- a/ocaml/runtime/shared_heap.c
+++ b/ocaml/runtime/shared_heap.c
@@ -70,6 +70,14 @@ static struct {
   caml_plat_mutex lock;
   pool* free;
 
+  /* Mapped but not yet active pools */
+  uintnat fresh_pools;
+  char* next_fresh_pool;
+
+  /* Counts pools in some domain's heap, or in the global list below.
+     Does not count free or fresh pools above */
+  uintnat active_pools;
+
   /* these only contain swept memory of terminated domains*/
   struct heap_stats stats;
   pool* global_avail_pools[NUM_SIZECLASSES];
@@ -78,6 +86,9 @@ static struct {
 } pool_freelist = {
   CAML_PLAT_MUTEX_INITIALIZER,
   NULL,
+  0,
+  NULL,
+  0,
   { 0, },
   { 0, },
   { 0, },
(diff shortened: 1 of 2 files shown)


ocaml/runtime/shared_heap.c
```diff
@@ -70,6 +70,14 @@ static struct {
   caml_plat_mutex lock;
   pool* free;
 
+  /* Mapped but not yet active pools */
+  uintnat fresh_pools;
+  char* next_fresh_pool;
+
+  /* Counts pools in some domain's heap, or in the global list below.
```
sadiqj, line 77: It took me a little while to understand this comment. I think something like "Count of all pools that are in-use across all domains and the global free list"

sadiqj (commented):
Other that the minor comment nit, this looks good.

We will need to test what the behaviour is with compaction under very fragmented heaps. The previous behaviour meant the kernel filled free pool-sized holes in the mappings and the total number of mappings didn't increase - this won't happen here due to the batching.

kayceesrk:
> We will need to test what the behaviour is with compaction under very fragmented heaps. 

Do we have a benchmark for this? It would be useful to include this in Sandmark (for posterity).

mshinwell (approved):

tmcgilchrist:
@stedolan @sadiqj Is there a benchmark for this? I can add it to sandmark if there is.

PR #2250, opened 2024-01-30
Additional bug fixes for middle/backend unboxed array support

alanechang:
Fixes a few more bugs discovered when writing tests in #2238 (the tests there should have pretty good coverage over the various corner cases at this point)

1. When building an unboxed int32 array dynamically, the element stored in the lower 4 bytes needs to be truncated before the `Cor` operation
2. Static unboxed int32 arrays need to be constructed differently from unboxed int64/nativeint arrays
3. Use correct symbol names for the custom ops in `to_cmm_static.ml`

### DIFF
--- a/backend/cmm_helpers.ml
+++ b/backend/cmm_helpers.ml
@@ -3689,7 +3689,12 @@ let make_unboxed_int32_array_payload dbg unboxed_int32_list =
     | [] -> Even, List.rev acc
     | a :: [] -> Odd, List.rev (a :: acc)
     | a :: b :: r ->
-      let i = Cop (Cor, [a; Cop (Clsl, [b; Cconst_int (32, dbg)], dbg)], dbg) in
+      let i =
+        Cop
+          ( Cor,
+            [zero_extend_32 dbg a; Cop (Clsl, [b; Cconst_int (32, dbg)], dbg)],
+            dbg )
+      in
       aux (i :: acc) r
   in
   aux [] unboxed_int32_list
--- a/backend/cmm_helpers.ml
+++ b/backend/cmm_helpers.ml
@@ -3692,7 +3692,10 @@ let make_unboxed_int32_array_payload dbg unboxed_int32_list =
       let i =
         Cop
           ( Cor,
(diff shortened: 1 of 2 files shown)


backend/cmm_helpers.ml
```diff
@@ -3689,7 +3689,12 @@ let make_unboxed_int32_array_payload dbg unboxed_int32_list =
     | [] -> Even, List.rev acc
     | a :: [] -> Odd, List.rev (a :: acc)
     | a :: b :: r ->
-      let i = Cop (Cor, [a; Cop (Clsl, [b; Cconst_int (32, dbg)], dbg)], dbg) in
+      let i =
+        Cop
+          ( Cor,
+            [zero_extend_32 dbg a; Cop (Clsl, [b; Cconst_int (32, dbg)], dbg)],
```
Gbury, line 3695: I think this deserves a comment in the code, explaining that by default, `a` will be sign-extended (as are all Int32s), and we need it to be zero_extended instead for the `or` to be correct.

middle_end/flambda2/to_cmm/to_cmm_static.ml
```diff
@@ -141,11 +141,25 @@ let immutable_unboxed_int_array env res updates update_kind ~symbol ~elts
...
+      | Int32 ->
+        let rec aux acc = function
+          | [] -> List.rev acc
+          | a :: [] -> List.rev (a :: acc)
+          | a :: b :: r ->
+            let i = Int64.(add (logand a 0xffffffffL) (shift_left b 32)) in
+            aux (i :: acc) r
+        in
+        aux [] int64_of_elts
+      | Int64_or_nativeint -> int64_of_elts
```
Gbury, line 158: This code should be split into a separate function.

Gbury (commented):
The fixes are good. I left a few style comments.

alanechang:
Thanks for the feedback! I made the changes as suggested.

Gbury (approved):

mshinwell (approved):

PR #2252, opened 2024-01-30
Determine unboxed int32 array length using custom_ops index

TheNumbat:
If the runtime stores the even and odd-length custom_ops structures in an array, the code to measure an `int32# array` can directly compute the index of the custom_ops pointer instead of comparing and branching.

Creating an odd-length array takes very slightly more work, since it has to add an offset to the base custom_op symbol, but this should be more than offset by speeding up the length computation (which is used in bounds-checked accesses).


backend/cmm_helpers.ml
```diff
@@ -891,13 +891,25 @@ let array_indexing ?typ log2size ptr ofs dbg =
    cross-compiling for 64-bit on a 32-bit host *)
 let int ~dbg i = natint_const_untagged dbg (Nativeint.of_int i)
 
-let custom_ops_unboxed_int32_odd_array =
-  Cconst_symbol
-    (Cmm.global_symbol "caml_unboxed_int32_odd_array_ops", Debuginfo.none)
+(* custom_ops_size_log2 must be kept in sync with runtime/caml/custom.h *)
```
mshinwell, line 894: Can we find a programmatic way of enforcing this?

mshinwell (changes requested):

backend/cmm_helpers.ml
lthls: Why do we store the log anyway ? We could store the size in words or bytes directly and do multiplications instead of shifts.

TheNumbat: I used a shift because I wasn't confident a div/mul would get optimized out post-cmm, but I'll check if that's actually the case.

Not sure of the best way to enforce the sizes match, though - maybe adding custom_size=64 to config.h and _Static_asserting in custom.h?

TheNumbat: The code-gen for `div_int` is indeed worse:
```asm
        movq	caml_unboxed_int32_array_ops@GOTPCREL(%rip), %rax
	subq	%rax, %rsi
	movq	%rsi, %rdx
	sarq	$5, %rdx
	shrq	$58, %rdx
	addq	%rdx, %rsi
	sarq	$6, %rsi
```
vs
```asm
	movq	caml_unboxed_int32_array_ops@GOTPCREL(%rip), %rsi
	subq	%rsi, %rax
	shrq	$6, %rax
```

lthls: Why does the division matter ? I can't see any divisions added by your PR.

TheNumbat: Oh, sorry, I thought you were talking about emitting `lsr_int` vs `mul_int` or `div_int`. I've updated the PR to define the custom ops size in bytes instead.

lthls: Ah no, I was indeed thinking of using multiplication and division, but failed to remember that in my second comment. I shouldn't post late in the evening... Thanks for using the size though, it makes things clearer.

mshinwell:
@TheNumbat I wouldn't make this too fancy for the moment, as I think the use of the `Custom_tag` blocks was supposed to be a stop-gap (if we make the header info more flexible, we can probably describe the arrays there)...

TheNumbat:
Yeah, this is done for now

ocaml/runtime4/caml/custom.h
```diff
@@ -39,6 +39,7 @@ struct custom_operations {
   int (*compare_ext)(value v1, value v2);
   const struct custom_fixed_length* fixed_length;
 };
+_Static_assert(sizeof(struct custom_operations) == CUSTOM_OPS_STRUCT_SIZE);
```
alanechang, line 42:
```suggestion
_Static_assert(sizeof(struct custom_operations) == CUSTOM_OPS_STRUCT_SIZE, "...");
```

alanechang (approved):
The changes look good to me. I can confirm the the tests in #2238 are still able to pass. Left a minor note about `_Static_assert`

Gbury (approved):
I'm not very competent about the changes to the C code or the asm changes, but this should be good.

mshinwell (approved):

PR #2253, opened 2024-01-31
Fix the package hack

ccasin:
This is a copy of upstream ocaml/ocaml#12930, with adjustments for layouts.  This is not super complicated, but I think the interesting parts are all to do with first class modules, which are unchanged from the upstream version.  So I suggest either reviewing the upstream PR first, or waiting until an external upstream review.  I was initially going to wait for that upstream review to materialize before posting here, but the issue is reported frequently enough that it may make sense to move faster internally.

The changes to the upstream version for layouts are fairly minimal.  There were some interesting questions to do with kind inference created by this change, but those were resolved in #2246.  The last commit in the upstream version is unnecessary here because layout checking doesn't trust the stored layout in a type declaration (it would also be unnecessary upstream if ocaml/ocaml#11841 were merged).


ccasin:
Review suggestion: Look at the main commit in the upstream version, comparing the new code path for package with type constraints to the old one, convincing yourself that nothing missing from the old one matters thanks to the restrictions on packages.  Ideally, leave a comment saying you confirmed this upstream, to move that along.  Then, come back here and look at the small additional diffs related to layouts.

goldfirere:
Just to keep folks up to date: I've reviewed the upstream PR, which is still in the process of settling. I'll return to this after it does. But there is indeed slow-moving action forward on this.

ccasin:
@goldfirere I have integrated the feedback from upstream review here.  I would just skim this - it's nearly identical to the upstream version - paying attention to changes for layouts, of which I think there are just three:
- An added layout check in the `With_type_package` case of `Typemod.merge_constraint`
- The lack of an immediacy computation in `transl_package_constraint` (because the layout can be computed by clients.  At some point I did have a layout computation here anyway - you suggested it was unnecessary, which is true, though I wonder if it would be better to add it back to avoid repeating layout checks - but either way seems fine)
- The removal of the `Package_hack` layout creation reason.

ocaml/testsuite/tests/typing-modules/package_constraint.ml
```diff
@@ -0,0 +1,332 @@
+(* TEST
```
goldfirere, line 1: This file is different than the version upstream. Maybe separate out our local changes (just by putting them at the bottom) to make merging (and reviewing) easier?

goldfirere (approved):
There are changes in printtyp.ml upstream, but not here. Intentional?

ocaml/testsuite/tests/typing-modules/package_constraint.ml
ccasin: Done (well, mostly - one addition makes a lot of sense where it is, and I think doesn't meaningfuly impact future merges becuase of how many nearby errors also change in our version).

PR #2257, opened 2024-02-01
Adjustments of sharing to improve unboxing

mshinwell:
The following example (which I must admit I thought used to work, but maybe a mistake was made back then) fails to unbox because sharing of `Lambda` terms causes the final `match` to be collapsed into a single `Pfield`.  This means that, even though there is a continuation available for unboxing in Flambda 2, we fail to unbox as the "reconstruction" `Switch` is absent:
```
type t =
  | C
  | D
  | E

type s =
  | A of int
  | B of int

let foo c a b =
  let m =
    match c with
    | C -> A a
    | D -> B b
    | E -> B (b + 1)
  in
  match m with
  | A x -> x
  | B y -> y
```

@lthls and I discussed this.  The proposed solution is to:
- stop sharing `Pfield` at the `Lambda` level
- allow sharing of `Pfield` (and indeed other operations) at the Cmm level.

Another way might have been to adjust the Flambda 2 unboxing code to have a variable per field (not per tag+field combination).  However this seems worse, since it loses precision in some cases (e.g. it would fail to propagate that the final return of `x` should actually just return `a`, not some unboxing parameter), and is also a more complicated change.  (Care would need to be taken in case the same field indexes were of different kinds for different tags, etc.)

This PR illustrates the use of a new function `Cmm_helpers.simple_and_equal_exprs` to implement the above strategy.  We started adjusting the `Cmm_helpers.transl_switch_clambda` code too, but wanted to discuss further before finishing this part.

Some further examples to try, following on from the above fragment:
```
let bar m =
  match m with
  | A x -> x
  | B y -> y

let baz c a b =
  let m =
    match c with
    | C -> A a
    | D -> B b
    | E -> B (b + 1)
  in
  match Sys.opaque_identity m with
  | A x -> x
  | B y -> y
```

This change should probably be put behind a flag.

@Gbury We await your opinions!

### DIFF
--- a/backend/cmm_helpers.ml
+++ b/backend/cmm_helpers.ml
@@ -2047,19 +2047,65 @@ end
 module StoreExpForSwitch = Switch.CtxStore (struct
   type t = expression
 
-  type key = int option * int
+  type key_info =
+    | Exit of int
+    | Load of
+        { block : Backend_var.t;
+          byte_offset : int;
+          memory_chunk : memory_chunk;
+          mutability : Asttypes.mutable_flag;
+          is_atomic : bool
+        }
+
+  type key = key_info option * int
 
   type context = int
 
   let make_key index expr =
     let continuation =
-      match expr with Cexit (Lbl i, [], []) -> Some i | _ -> None
+      match expr with
+      | Cexit (Lbl i, [], []) -> Some (Exit i)
+      | Cop (Cload { memory_chunk; mutability; is_atomic }, [Cvar block], _) ->
+        Some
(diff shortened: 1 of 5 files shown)


mshinwell:
Handing over to @gbury

chambart:
In the first example, it would be possible for the unboxed version of the code to be equivalent in both branches.
Right now unboxing generates a different variable for each field of each case. But if we were to share those variables in certain predictable cases, we could make that kind of example shareable in cmm in unboxed version.

PR #2259, opened 2024-02-02
Improved meet for naked immediates (is_int and get_tag)

lthls:
A first attempt to improve the meet when more than one `Get_tag`/`Is_int` type is involved.
As noted in the comment, I'm also considering more comprehensive changes to get even better results, but I'll probably wait until I have at least one motivating example before making significant changes the type grammar.

### DIFF
--- a/middle_end/flambda2/types/expand_head.ml
+++ b/middle_end/flambda2/types/expand_head.ml
@@ -375,6 +375,42 @@ let is_bottom env t = ET.is_bottom (expand_head env t)
 
 let is_unknown env t = ET.is_unknown (expand_head env t)
 
+let[@warning "-4"] reduce_is_int env t : _ Or_unknown_or_bottom.t =
+  match ET.descr (expand_head env t) with
+  | Unknown -> Unknown
+  | Bottom -> Bottom
+  | Ok (Value (Variant { immediates; blocks; _ })) ->
+    Ok
+      (Targetint_31_63.Set.union
+         (match immediates with
+         | Unknown -> Targetint_31_63.Set.singleton Targetint_31_63.one
+         | Known ty ->
+           if TG.is_obviously_bottom ty
+           then Targetint_31_63.Set.empty
+           else Targetint_31_63.Set.singleton Targetint_31_63.one)
+         (match blocks with
+         | Unknown -> Targetint_31_63.Set.singleton Targetint_31_63.zero
+         | Known blocks ->
+           if TG.Row_like_for_blocks.is_bottom blocks
+           then Targetint_31_63.Set.empty
+           else Targetint_31_63.Set.singleton Targetint_31_63.zero))
+  | Ok _ -> Unknown
+
+let[@warning "-4"] reduce_get_tag env t : _ Or_unknown_or_bottom.t =
+  match ET.descr (expand_head env t) with
+  | Unknown -> Unknown
+  | Bottom -> Bottom
+  | Ok (Value (Variant { blocks = Known blocks; _ })) -> (
+    match blocks.other_tags with
+    | Bottom ->
+      Ok
+        (Tag.Map.fold
+           (fun tag _ imms ->
+             Targetint_31_63.Set.add (Tag.to_targetint_31_63 tag) imms)
+           blocks.known_tags Targetint_31_63.Set.empty)
+    | Ok _ -> Unknown)
+  | Ok _ -> Unknown
+
 let missing_kind env free_names =
   Name_occurrences.fold_variables free_names ~init:false
     ~f:(fun missing_kind var ->
--- a/middle_end/flambda2/types/expand_head.mli
+++ b/middle_end/flambda2/types/expand_head.mli
@@ -97,6 +97,12 @@ val is_bottom : Typing_env.t -> Type_grammar.t -> bool
 
 val is_unknown : Typing_env.t -> Type_grammar.t -> bool
 
+val reduce_is_int :
+  Typing_env.t -> Type_grammar.t -> Targetint_31_63.Set.t Or_unknown_or_bottom.t
+
+val reduce_get_tag :
+  Typing_env.t -> Type_grammar.t -> Targetint_31_63.Set.t Or_unknown_or_bottom.t
+
 type to_erase =
   | Everything_not_in of Typing_env.t
   | All_variables_except of Variable.Set.t
--- a/middle_end/flambda2/types/meet_and_join_new.ml
+++ b/middle_end/flambda2/types/meet_and_join_new.ml
@@ -783,21 +783,22 @@ and meet_head_of_kind_naked_immediate env (t1 : TG.head_of_kind_naked_immediate)
     | Left -> Ok (Right_input, env)
     | Right -> Ok (Left_input, env)
   in
-  let meet_with_shape ~rebuild ty shape side =
+  let meet_with_shape env ~rebuild ty shape side =
     map_result ~f:rebuild
       (match side with Left -> meet env ty shape | Right -> meet env shape ty)
   in
-  let is_int_immediate ~is_int_ty ~immediates ~is_int_side =
+  let is_int_immediate env ~is_int_ty ~immediates ~is_int_side =
     let rebuild = TG.Head_of_kind_naked_immediate.create_is_int in
     match I.Set.mem I.zero immediates, I.Set.mem I.one immediates with
     | false, false -> Bottom
     | true, true -> keep_side is_int_side
     | true, false ->
-      meet_with_shape ~rebuild is_int_ty MTC.any_block is_int_side
+      meet_with_shape env ~rebuild is_int_ty MTC.any_block is_int_side
     | false, true ->
-      meet_with_shape ~rebuild is_int_ty MTC.any_tagged_immediate is_int_side
+      meet_with_shape env ~rebuild is_int_ty MTC.any_tagged_immediate
+        is_int_side
   in
-  let get_tag_immediate ~get_tag_ty ~immediates ~get_tag_side =
+  let get_tag_immediate env ~get_tag_ty ~immediates ~get_tag_side =
     if I.Set.is_empty immediates
     then keep_other_side get_tag_side
     else
@@ -816,7 +817,7 @@ and meet_head_of_kind_naked_immediate env (t1 : TG.head_of_kind_naked_immediate)
           MTC.blocks_with_these_tags tags (Alloc_mode.For_types.unknown ())
         with
         | Known shape ->
-          meet_with_shape
+          meet_with_shape env
             ~rebuild:TG.Head_of_kind_naked_immediate.create_get_tag get_tag_ty
             shape get_tag_side
         | Unknown -> keep_side get_tag_side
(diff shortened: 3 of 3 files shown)


middle_end/flambda2/types/grammar/type_grammar.ml
```diff
@@ -3093,24 +3144,30 @@ end
...
     if Targetint_31_63.Set.is_empty imms
     then
       Misc.fatal_error
         "Head_of_kind_naked_immediates.create_naked_immediates_non_empty";
-    Naked_immediates imms
+    { immediates = Known imms; relations = RelationSet.empty }
 
-  let create_is_int ty = Is_int ty
+  let create_is_int name =
+    { immediates = Unknown; relations = RelationSet.singleton (Is_int name) }
```
Ekdohibs, line 3167: Shouldn't immediates be the set {#0, #1} here?

lthls: Yes, that would work, although I'm not sure if that is useful. If you look at the provers for instance then we do more clever things it this is still unknown.

middle_end/flambda2/types/grammar/more_type_creators.ml
```diff
@@ -352,3 +352,27 @@ let unknown_types_from_arity arity =
...
+    | false, false -> Bottom
+    | true, false -> Ok (name, any_block)
+    | false, true -> Ok (name, any_tagged_immediate))
+  | Get_tag name -> (
+    let tags =
+      I.Set.fold
+        (fun tag tags ->
+          match Tag.create_from_targetint tag with
+          | Some tag -> Tag.Set.add tag tags
+          | None -> tags
```
Gbury, line 372: Shouldn't we produce `bottom` in that case ? From what I understand, we are trying to reconstruct a shape from information about a `Gat_tag`, and in this case, we have just seen a tag that is outside the bounds of tags, which should not happen except in dead branches ?

Gbury (commented):
All good except for the one comment/question.

middle_end/flambda2/types/grammar/more_type_creators.ml
lthls: Returning Bottom for the whole type would be unsound, because you could end up with Bottom here just because you've lost precision elsewhere.
Not adding the tag to the set already treats the corresponding case as unreachable.

Gbury: Indeed, I misunderstood.

Gbury (approved):

mshinwell:
@lthls please rebase

lthls:
Rebased

middle_end/flambda2/types/expand_head.mli
```diff
@@ -102,6 +102,12 @@ val is_bottom : Typing_env.t -> Type_grammar.t -> bool
 
 val is_unknown : Typing_env.t -> Type_grammar.t -> bool
 
+val reduce_is_int :
+  Typing_env.t -> Type_grammar.t -> Targetint_31_63.Set.t Or_unknown_or_bottom.t
+
+val reduce_get_tag :
+  Typing_env.t -> Type_grammar.t -> Targetint_31_63.Set.t Or_unknown_or_bottom.t
+
```
chambart, line 110:
```suggestion
val reduce_is_int :
  Typing_env.t -> block_type:Type_grammar.t -> Targetint_31_63.Set.t Or_unknown_or_bottom.t

val reduce_get_tag :
  Typing_env.t -> block_type:Type_grammar.t -> Targetint_31_63.Set.t Or_unknown_or_bottom.t

```
I would add a label to those arguments to make it obvious what is expected. Otherwise we could imagine that the type agument might be the integer returned by `is_int` or `get_type`.

middle_end/flambda2/types/expand_head.ml
```diff
@@ -402,6 +402,42 @@ let is_bottom env t = ET.is_bottom (expand_head env t)
...
+           if TG.is_obviously_bottom ty
+           then Targetint_31_63.Set.empty
+           else Targetint_31_63.Set.singleton Targetint_31_63.one)
+         (match blocks with
+         | Unknown -> Targetint_31_63.Set.singleton Targetint_31_63.zero
+         | Known blocks ->
+           if TG.Row_like_for_blocks.is_bottom blocks
+           then Targetint_31_63.Set.empty
+           else Targetint_31_63.Set.singleton Targetint_31_63.zero))
+  | Ok _ -> Unknown
```
chambart, line 424: We can be more certain than `Unknown` in that case. But would it matter ?

middle_end/flambda2/types/expand_head.ml
```diff
@@ -402,6 +402,42 @@ let is_bottom env t = ET.is_bottom (expand_head env t)
...
+  | Ok (Value (Variant { blocks = Known blocks; _ })) -> (
+    match blocks.other_tags with
+    | Bottom ->
+      Ok
+        (Tag.Map.fold
+           (fun tag _ imms ->
+             Targetint_31_63.Set.add (Tag.to_targetint_31_63 tag) imms)
+           blocks.known_tags Targetint_31_63.Set.empty)
+    | Ok _ -> Unknown)
+  | Ok _ -> Unknown
```
chambart, line 439: Same as above, we could know more than `Unknown`

middle_end/flambda2/types/grammar/type_grammar.ml
```diff
@@ -2177,15 +2218,20 @@ and project_head_of_kind_value ~to_project ~expand head =
...
+  let relations' =
+    RelationSet.filter
+      (function
+        | Is_int name | Get_tag name ->
+          Name.pattern_match name
+            ~symbol:(fun _ -> true)
+            ~var:(fun var -> not (Variable.Set.mem var to_project)))
+      relations
+  in
+  if RelationSet.cardinal relations' = RelationSet.cardinal relations
```
chambart, line 2232: Physical equality for Set.filter is guaranteed when nothing is filtered

middle_end/flambda2/types/grammar/type_grammar.ml
```diff
@@ -104,9 +118,9 @@ and head_of_kind_value =
    if we don't actually use the extensions, while the second version would be
    particularly useful if we switch several times on the same scrutinee. *)
 and head_of_kind_naked_immediate =
-  | Naked_immediates of Targetint_31_63.Set.t
-  | Is_int of t
-  | Get_tag of t
+  { immediates : Targetint_31_63.Set.t Or_unknown.t;
+    relations : RelationSet.t
+  }
```
chambart, line 123:
```suggestion
  | Naked_immediates of Targetint_31_63.Set.t
  | Relations of RelationSet.t
  | Both of
  { immediates : Targetint_31_63.Set.t Or_unknown.t;
    relations : RelationSet.t
  }
```
For compaction, as most of the time there would be only one of the two parts and to avoid an indirection for through Ok in the common case, this change might be benefitial (bu tedious to apply).

middle_end/flambda2/simplify/common_subexpression_elimination.ml
```diff
@@ -264,13 +264,19 @@ let join_one_cse_equation ~cse_at_each_use prim bound_to_map
...
         | Unary (Is_int { variant_only = true }, scrutinee) ->
-          Name.Map.add (Name.var var)
-            (T.is_int_for_scrutinee ~scrutinee)
-            extra_equations
+          Simple.pattern_match scrutinee
+            ~name:(fun scrutinee ~coercion:_ ->
+              Name.Map.add (Name.var var)
+                (T.is_int_for_scrutinee ~scrutinee)
+                extra_equations)
+            ~const:(fun _ -> extra_equations)
```
chambart, line 272: Should we add a `= 1` equation in that case ?

chambart: After discussion that should be a fatal error instead. Same thing for get_tag

middle_end/flambda2/types/provers.ml
```diff
@@ -157,28 +157,55 @@ let prove_get_tag env t = as_property (prove_get_tag_generic env t)
...
+    let refine_with_is_int (proof : _ generic_proof) name =
+      let is_int_set =
+        let scrutinee_ty = TE.find env name (Some Flambda_kind.value) in
+        match prove_is_int_generic ~variant_only:true env scrutinee_ty with
+        | Proved true -> Targetint_31_63.Set.singleton Targetint_31_63.bool_true
+        | Proved false ->
+          Targetint_31_63.Set.singleton Targetint_31_63.bool_false
+        | Unknown ->
+          Targetint_31_63.Set.of_list
+            [Targetint_31_63.bool_false; Targetint_31_63.bool_true]
```
chambart, line 184: Would this be worthwhile to declare that value at toplevel to avoid recomputing it. This might occur quite often

middle_end/flambda2/types/grammar/type_grammar.mli
```diff
@@ -29,6 +29,12 @@ module Block_size : sig
   val inter : t -> t -> t
 end
 
+type relation = private
+  | Is_int of Name.t
+  | Get_tag of Name.t
```
chambart, line 34: That name might be a bit too generic. `int_returning_relation` maybe ?

chambart: Another possibility is to put them in a naked_immediate submodule that would make it obvious

middle_end/flambda2/types/meet_and_join_new.ml
```diff
@@ -780,79 +786,46 @@ and meet_head_of_kind_naked_immediate env (t1 : TG.head_of_kind_naked_immediate)
...
+       passed to the fold function *)
+    match extract_value new_value t1 t2 with
+    | { immediates = Unknown; relations = _ } ->
+      (* No reduction to do *)
+      result
+    | { immediates = Known is; relations } ->
+      (* Update the result by reduction, following the relations *)
+      let reduce relation (result : _ meet_result) : _ meet_result =
+        match result with
+        | Bottom -> Bottom
```
chambart, line 816: Here it is `match result with Bottom -> Bottom` and higher, it is `match result with Bottom -> result`
It should be coherent I think. I tend to prefer `Bottom -> Bottom`

middle_end/flambda2/types/expand_head.mli
chambart: By the way those functions are not used. They should be called from the meet.

middle_end/flambda2/types/provers.ml
```diff
@@ -157,28 +157,55 @@ let prove_get_tag env t = as_property (prove_get_tag_generic env t)
 
 let prove_naked_immediates_generic env t : Targetint_31_63.Set.t generic_proof =
   match expand_head env t with
-  | Naked_immediate (Ok (Naked_immediates is)) ->
+  | Naked_immediate (Ok { immediates = Known is; relations = _ }) ->
+    (* No attempt at reduction *)
```
chambart, line 161: This might be a bit too fidly not to reduce in that case. It would be simpler to just always reduce

chambart (changes requested):
Globally OK

PR #2260, opened 2024-02-02
Zero alloc: slightly better error messages

gretay-js:
Instead of
```
Error: Unexpected direct tailcall camlTest_misplaced_assume.bar_0_2_code
```
print 
```
Error: direct tailcall camlTest_misplaced_assume.bar_0_2_code may allocate
```

### DIFF
--- a/backend/checkmach.ml
+++ b/backend/checkmach.ml
@@ -70,7 +70,7 @@ module Witness = struct
       fprintf ppf "direct tailcall %s" callee
     | Missing_summary { callee } -> fprintf ppf "missing summary for %s" callee
     | Forward_call { callee } ->
-      fprintf ppf "foward call or tailcall (conservatively handled) %s" callee
+      fprintf ppf "forward call or tailcall (conservatively handled) %s" callee
     | Extcall { callee } -> fprintf ppf "external call to %s" callee
     | Arch_specific -> fprintf ppf "arch specific operation"
     | Probe { name; handler_code_sym } ->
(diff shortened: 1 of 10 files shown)


gretay-js:
> This is a clear improvement, and it's fine if you merge as is. But I think it can be even better.
> 
> 1. I think these errors are slightly confusing because they emphasize (by putting first) the type of call (in your example, "direct tailcall").  But of course it doesn't really matter that this is a direct tailcall.  The relevant thing is that the function being called may allocate.  So this error leads a naive user to wonder: Can I fix this by changing the way I call this function? Is it a problem that it is a tailcall?  Would an indirect call be better?
>    So I think it would be even better to just print something like "Called function camlTest_misplaced_assume.bar_0_2_code may allocate."  If you want to include the type of call, a parenthentical after this information would be better, or some other way or indicating it's not the most important thing.

Fixed.

> 2. "may allocate" is a little vague.  Under what circumstances does `get_alloc_dbginfo` return `None`?  

Ah, that was just a shortcut to identify that the witness kind is `Alloc` and avoid printing "allocation .. may allocate". I refactored the printing code to make this a bit more obvious.

> Can we give the user a hint?  E.g., does this suggest we didn't find the function's definition, and are there build settings they could adjust that might help?

There is a separate  "missing summary" message for the case of not find the function's definition. When CMXs are not available we just end up with an indirect call, not a missing summary. I feel that giving a "hint" about build system settings is out of scope.  We can't tell if it's because of missing CMX or genuinely higher order function argument. 
We print "may allocate" conservatively - the nature of the analysis makes it hard to distinguish between "must allocate" and imprecision of the analysis (e.g., co-related control flow). We can try tracking "must allocate" separately, but I haven't seen an example yet where it would be helpful. Let's leave further error message improvements for later.

ccasin:
Thanks, looks great!

PR #2262, opened 2024-02-06
Warn if Dune is too new

lukemaurer:
Make `./configure` complain if Dune (as found either in the `PATH` or via `--with-dune`) has a version greater than we know to work (currently 3.8). As a bonus, error out if dune can't be executed (previously this would keep going in a doomed state).


riaqn (approved):
LGTM

PR #2266, opened 2024-02-08
Weak symbol support in x86 binary emitter

TheNumbat:
Adds support for weak symbols. For use (and testing) in #2220.


gretay-js (approved):
Looks fine for the limited use of  `.weak` directives in the backend (no relocations or expression over weak symbols, used once, always defined using `.label`, defined once, etc).

PR #2268, opened 2024-02-08
Request ISA extension checks from C stubs

TheNumbat:
User-defined C bindings will now include `isa.h` by default (via `mlvalues.h`). 

Based on the C compiler flags (e.g. gcc defines `__SSE3__` when SSE3 is enabled), this file defines weak symbols corresponding to each enabled extension. These symbols cause the CPUID check (see #2220)  to also check for any extensions used in C stubs.

### DIFF
--- a/ocaml/runtime/caml/isa.h
+++ b/ocaml/runtime/caml/isa.h
@@ -17,6 +17,92 @@
 
 #include "misc.h"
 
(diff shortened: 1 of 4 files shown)


gretay-js (commented):
This looks fine to me, but I'd like someone familiar with the runtime to make the final call about this one.

mshinwell (approved):

PR #2275, opened 2024-02-12
Implicit source position merge conflicts

Enoumy:
# Implicit Source Positions Conflict resolution
    
This feature solely fixes merge conflicts that the implicit source positions project had after not having been rebased since the summer.

Sadly, I mistakenly committed a new test! I accidentally committed a test that adds support for `let+` and `let*` into this branch instead of a separate branch, and then I accidentally committed some more changes after that making it trickier to split... Please let me know if I should only keep this feature do "merge conflict resolution".

Testing
-------
- I ran `make test`, and it passes.
- `make install` works, I migrated bonsai to start using this and it works!
- Please let me know if there is other additional testing/ci I should perform. Thanks!

For reviewers
------------------
Please let me know if the commit messages/general git "etiquette" conventions are followed/if there is something I should change.
- Weird things include that my first draft of this pr had many commits with changes that were later undone, so I ended up squashing the commits. Please let me know if this is reasonable/if you would like me to split off the commits.


ocaml/testsuite/tests/typing-implicit-source-positions/let_operators.ml
```diff
@@ -0,0 +1,62 @@
+(* TEST
```
goldfirere, line 1: This is a good idea embodied in this file. It would be cool if all of the examples here (both with source positions and with ordinary optionals) worked. Do you think we'd use this feature within Jane Street? I don't think it would be hard to make work, but not quite trivial either.

goldfirere (approved):

ocaml/testsuite/tests/typing-implicit-source-positions/let_operators.ml
Enoumy: > Do you think we'd use this feature within Jane Street?

Ooh good question! I think it would not be used within Jane Street as `ppx_js_style` forbids using let operators in favor of `ppx_let`.

goldfirere: OK. Let's not do this, then. Maybe when upstreaming, we can take a stab. Instead, leave comments (plain ones -- no `CR`) saying that the examples would be good to accept, but we're not going to for now.

ocaml/testsuite/tests/typing-implicit-source-positions/let_operators.ml
```diff
@@ -0,0 +1,52 @@
...
+   * expect
+*)
+
+let ( let+ ) ~(src_pos : [%src_pos]) a f = f (src_pos, a);;
+[%%expect{|
+val ( let+ ) : src_pos:[%src_pos] -> 'a -> (lexing_position * 'a -> 'b) -> 'b =
+  <fun>
+|}]
+
+(* CR src_pos: Add support for implicit position parameters and (also maybe optional
```
goldfirere, line 11:
```suggestion
(* Would be nice to add support for implicit position parameters and (also maybe optional
```

I think we should leave off the `CR`. This isn't something we'll come back to as part of this project, and we'd be content even upstreaming the code as-is (if upstream is content).

goldfirere (approved):

ocaml/testsuite/tests/typing-implicit-source-positions/let_operators.ml
Enoumy: I have updated the comment! Thanks!

PR #2280, opened 2024-02-13
Interpret `[@inlined]` as `[@unrolled 1]` on recursive functions

lukemaurer:
This keeps things in line with expectations if a function `outer` calls a higher-order parameter `f` with `[@inlined hint]` and someone calls `outer` with `f` set to a recursive function. Previously, this could cause the interpreter to loop (or blow up, at any rate) because `[@inlined]` bypassed most limitations on inlining of recursive functions. Using `[@unrolled 1]` honours the intent that the call site be inlined but makes sure things don't get out of hand.

Specifically, the existing behaviour is to treat any `[@inlined]` as very nearly the final authority: it is only ever ignored if either we're already unrolling or we've hit the max overall inlining depth (counting _everything_ being inlined). In the case of a recursive function, this is fine if the call is external but disastrous in any code that's mutually recursive with the function.

With this patch, after we find the function that we would be inlining, we consult its metadata to find whether it's recursive, and if so, we interpret any `[@inlined]` the same as `[@unrolled 1]`. For an external call, this should have basically the same effect as `[@inlined]`. For a recursive call, we'll inline only one copy of the body.

Caveats:

* This relies on the function being correctly identified as recursive in the metadata, which isn't 100% but the known cases hitting this problem all involve a conventional `let rec`.

* There could be some sort of second-order effect from the fact that the inlined body's recursive calls will carry rec info saying they're fully unrolled.

### DIFF
--- a/middle_end/flambda2/simplify/inlining/call_site_inlining_decision.ml
+++ b/middle_end/flambda2/simplify/inlining/call_site_inlining_decision.ml
@@ -232,12 +232,30 @@ let make_decision dacc ~simplify_expr ~function_type ~apply ~return_arity :
            ramifications of treating unknown-ness as an observable property this
            way. Are we relying on monotonicity somewhere? *)
         let apply_inlining_state = Apply.inlining_state apply in
+        let recursive =
+          Code_metadata.recursive
+            (Code_or_metadata.code_metadata code_or_metadata)
+        in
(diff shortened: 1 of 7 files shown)


lthls (approved):
The code looks good.
The test output looks reasonable at first glance, though I haven't looked too closely.

Gbury (approved):

PR #2281, opened 2024-02-13
Add flag limiting registers available for allocation

TheNumbat:
The `-flimit-regalloc=N` flag specifies that only the first `N` hardware registers should be used for normal register allocation. 
Pre-allocated registers (C calls, etc.) are not effected.

Currently only the greedy allocator respects the flag, as it is intended for testing GI.

### DIFF
--- a/backend/amd64/arch.ml
+++ b/backend/amd64/arch.ml
@@ -95,6 +95,9 @@ let trap_notes = ref true
 (* Emit extension symbols for CPUID startup check  *)
 let arch_check_symbols = ref true
 
+(* Limit hardware registers available for allocation *)
+let limit_regalloc = ref Int.max_int
+
 (* Machine-specific command-line options *)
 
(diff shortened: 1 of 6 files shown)


lthls:
I'm dubious about the change to the calling conventions. If you link together two files compiled with different limits you will get runtime errors. So it should either be a configuration flag, or restricted to the allocator itself.
If it's just intended as a hack for testing, you can ignore this comment.

TheNumbat:
Yeah, this is a hack for testing, but it might be useful to keep around for future debugging. Should it still be a configuration option, or do you think naming it `unsafe-limit-regs` (or something like that) would be fine?

mshinwell:
If it's just a hack, having a runtime flag is probably easier.

TheNumbat:
After experimenting more in #2284, I think this is probably too hacky to merge into main - we can close it

PR #2284, opened 2024-02-13
[testing] greedy allocator

TheNumbat:
(not to merge)


TheNumbat:
(Appears to be working)

PR #1271, opened 2023-03-29
Functions with unboxed parameters

Ekdohibs:
This is an extension of #1248, with support for unboxed parameters as well as unboxed returns. Note that, for now, parameters will be unboxed if and only if there is an `[@unboxed]` annotation on the function, in which case the return is unboxed as well, if possible. Once we get #1257, we can unbox them on a per-parameter basis.

PR #2285, opened 2024-02-13
Fix build (float32 renaming conflict)

TheNumbat:
Updates #1271 to compile with the recently added `Pfloat32` case.


ncik-roberts (approved):
LGTM, let's merge after CI passes.

PR #2286, opened 2024-02-13
Mark `layouts` extension as erasable and limit the use of immediate/immediate64

alanechang:
This PR seeks to make two changes:

- Mark the `layouts` extension as erasable again after #2241
- Restrict the use of `immediate`/`immediate64` when the `-only-erasable-extensions` flag is set:

This is done to maintain the promise: *programs that typecheck under `-only-erasable-extensions` can be erased/transformed into an upstream compatible version*.

Programs such as:

```
let f (x : (_ : immediate)) = x;;
```

and soon (after #2253):

```
module type S = sig
  type t: immediate64
end

module type K = sig
  val f : 'a -> (module S with type t = 'a) -> 'a
end
```

can't be expressed in the upstream compiler.

To catch these programs and reject them under `-only-erasable-extensions`, there's now a check when a type is generalized to see that it doesn't contain any generic type variables with jkind `immediate` or `immediate64`.

Alternative implementation here would be to limit the contexts where `: immediate` layout annotations are allowed, but that wouldn't cover the first class module example above and additional checks would still be necessary.

---

There's the suggestion to rename the `-only-erasable-extensions` flag to `-upstream-compatible`. It seems easier to make that change in a follow-up PR rather than in this one.

### DIFF
--- a/ocaml/testsuite/tests/typing-layouts/erasable_annot.ml
+++ b/ocaml/testsuite/tests/typing-layouts/erasable_annot.ml
@@ -1,6 +1,6 @@
 (* TEST
    * expect
-   flags = "-extension layouts -only-erasable-extensions"
+   flags = "-only-erasable-extensions"
 *)
 
 (* Upstream compatible usages of immediate/immediate64 are allowed *)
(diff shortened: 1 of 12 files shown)


ocaml/testsuite/tests/typing-layouts/erasable_annot.ml
```diff
@@ -0,0 +1,241 @@
+(* TEST
+   * expect
+   flags = "-extension layouts -only-erasable-extensions"
```
goldfirere, line 3: Shouldn't need `-extension layouts` I think.

ocaml/testsuite/tests/typing-layouts/erasable_annot.ml
```diff
@@ -0,0 +1,241 @@
+(* TEST
```
goldfirere, line 1: Nice tests here. Please also add for
* `... int as (_ : immediate) ...` (should be accepted, because that annotation does nothing)
* `type 'a id = 'a` and then `... ('a id as (_ : immediate)) ...` (should be rejected, as it will affect `'a`)

ocaml/typing/ctype.mli
```diff
@@ -555,7 +555,12 @@ val check_type_jkind :
 val constrain_type_jkind :
   Env.t -> type_expr -> Jkind.t -> (unit, Jkind.Violation.t) result
 
-(* Update the jkind reason of all generalized type vars inside the given [type_expr]
+(* This function should get called after a type is generalized.
+
+   It does two thing:
```
goldfirere, line 560:
```suggestion
   It does two things:
```

goldfirere (approved):
Looks great. This was as easy as I was hoping for it to be. :)

ocaml/testsuite/tests/typing-layouts/erasable_annot.ml
alanechang: Yep, removed the flag

ocaml/testsuite/tests/typing-layouts/erasable_annot.ml
alanechang: Added both test cases

PR #2287, opened 2024-02-14
Disable function sorting

gretay-js:
Temporarily disable sorting of function symbols by debug info. Emit function symbols in topological order. 

Currently, the layout of function symbols is determined by their debug info, so that functions in the final assembly file appear in the same order as in the source file. This is convenient for debugging and stable layout.  However, it can makes some backend passes overly conservative. Currently, this affects only `checkmach` and `polling` that rely on "future" function names. For example, `zero_alloc` check fails on a common code pattern `raise_s [%message ...]`. 

```
let[@zero_alloc] raise_invalid_index ~index ~name =
  if index < 0 then
  raise_s
    [%message
      "Negative" (index : int) (name : string)]
```
results in the following code after [ppx](https://github.com/janestreet/ppx_sexp_message) :
```
let raise_invalid_index ~index ~name =
  if index < 0 then
  raise_s
    (let ppx_sexp_message () =
       Sexp.List
         [Sexp.Atom "Negative";
          Sexp.List
            [Sexp.Atom "index"; Sexp.of_int index];
          Sexp.List
            [Sexp.Atom "name"; Sexp.Atom name]]
     [@@ocaml.inline never][@@ocaml.local never][@@ocaml.specialise never]
     in
     ((ppx_sexp_message ())[@nontail ]))
[@@ocaml.inline never][@@ocaml.local never][@@ocaml.specialise never]
```
and `raise_invalid_index` is emitted before `ppx_sexp_message` even though they are not mutually recursive. 

A proper fix is more involved. The backend translates functions in layout order, fully translating each function from `Cmm` all the way to `assembly` before moving on to the next function.  We can change the backend to translate the functions in topological order and then emit them in source order. This means holding on to `Mach.fundecl` longer and careful handling any mutable state. Separately, an improvenent to zero_alloc analysis to handle mutually recursive functions will also fix this.

### DIFF
--- a/driver/flambda_backend_args.ml
+++ b/driver/flambda_backend_args.ml
@@ -114,6 +114,13 @@ let mk_checkmach_details_cutoff f =
      | No_details -> 0
      | At_most n -> n)
 
+let mk_function_layout f =
+  let layouts = Flambda_backend_flags.Function_layout.(List.map to_string all) in
+  let default = Flambda_backend_flags.Function_layout.(to_string default) in
+  "-function-layout", Arg.Symbol (layouts, f),
+  (Printf.sprintf " Order of functions in the generated assembly (default: %s)"
+     default)
+
 let mk_disable_poll_insertion f =
   "-disable-poll-insertion", Arg.Unit f, " Do not insert poll points"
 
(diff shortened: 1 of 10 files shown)


gretay-js:
Sorting was introduced here
https://github.com/ocaml-flambda/ocaml/pull/74

mshinwell (approved):

PR #2289, opened 2024-02-14
Restore asmcomp tests

lthls:
This PR restores the asmcomp tests.
The first commit makes them run again, the second one disables tests that we don't expect to work, and the last one fixes the `optargs` test that we do expect to work.

The extra copy of `testsuite/tests/asmcomp` that I've removed has only a subset of the files, and the contents of those files (both inputs and references) are exactly the same as the ones in the main testsuite folder.


lthls:
The test failures in classic mode are a bit annoying. Should I disable these tests for now (in classic mode), or should I try to get them to pass ?
I guess that `optargs` at least would be nice to have in classic mode, but I don't know yet if it's going to be easy to fix.

ocaml/testsuite/tests/asmcomp/is_static_flambda.ml
```diff
@@ -1,12 +1,17 @@
 (* TEST
    modules = "is_in_static_data.c is_static_flambda_dep.ml"
    * flambda
-   ** naked_pointers
-   *** native
+   ** runtime4
+   *** naked_pointers
+   **** native
+   flags = "-flambda2-join-points"
```
Gbury, line 7: This is redundant with the `force O3` line just below

Gbury (approved):

mshinwell:
@lthls just a small tweak to do, as discussed + rebase

PR #2291, opened 2024-02-14
Abort if greedy introduces too many temporaries

TheNumbat:
Adds a second termination condition: if a round introduces more than 10 times the number of temporaries we started with.

### DIFF
--- a/backend/regalloc/regalloc_gi.ml
+++ b/backend/regalloc/regalloc_gi.ml
@@ -109,15 +109,28 @@ let make_hardware_registers_and_prio_queue (cfg_with_infos : Cfg_with_infos.t) :
 (* CR xclerc for xclerc: try to find a reasonable threshold. *)
 let max_rounds = 32
 
+let max_temp_multiplier = 10
+
 (* CR xclerc for xclerc: the `round` parameter is temporary; this is an hybrid
    version of "greedy" using the `rewrite` function from IRC when it needs to
    spill. *)
-let rec main : round:int -> flat:bool -> State.t -> Cfg_with_infos.t -> unit =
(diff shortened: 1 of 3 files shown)


backend/regalloc/regalloc_gi.ml
```diff
@@ -109,15 +109,28 @@ let make_hardware_registers_and_prio_queue (cfg_with_infos : Cfg_with_infos.t) :
...
+
 (* CR xclerc for xclerc: the `round` parameter is temporary; this is an hybrid
    version of "greedy" using the `rewrite` function from IRC when it needs to
    spill. *)
-let rec main : round:int -> flat:bool -> State.t -> Cfg_with_infos.t -> unit =
- fun ~round ~flat state cfg_with_infos ->
+let rec main :
+    round:int ->
+    flat:bool ->
+    temporaries:int ->
```
xclerc, line 120: I think it might be better to
have `temporaries` stored
in `State.t`.

TheNumbat: Done

ocaml/runtime/caml/isa.h
```diff
@@ -66,7 +66,7 @@ CAML_REQUIRE_ARCH_PREFETCHW
 #endif
 
 #ifdef __PREFETCHWT1__
-CAML_REQUIRE_ARCH_PREFETCHWT1
+// CAML_REQUIRE_ARCH_PREFETCHWT1
```
xclerc: Is it part of this pull request on purpose?

xclerc: (looks like I was commenting on something not in the diff... sorry for the noise)

xclerc (approved):

PR #2292, opened 2024-02-15
Inline stubs more aggressively

lthls:
This PR is intended to fix the `optargs` test, which ensures that recursive functions with default parameters do not allocate options when calling themselves recursively.
The test itself is still inactive, but will be restored by #2289, which this commit is cherry-picked from.

### DIFF
--- a/middle_end/flambda2/simplify/inlining/call_site_inlining_decision.ml
+++ b/middle_end/flambda2/simplify/inlining/call_site_inlining_decision.ml
@@ -142,14 +142,22 @@ let argument_types_useful dacc apply =
       (Apply.args apply)
 
 let might_inline dacc ~apply ~code_or_metadata ~function_type ~simplify_expr
-    ~return_arity : Call_site_inlining_decision_type.t =
+    ~return_arity ~rec_depth_exceeded : Call_site_inlining_decision_type.t =
   let denv = DA.denv dacc in
   let env_prohibits_inlining = not (DE.can_inline denv) in
   let decision =
     Code_or_metadata.code_metadata code_or_metadata
     |> Code_metadata.inlining_decision
   in
-  if Function_decl_inlining_decision_type.must_be_inlined decision
+  if Function_decl_inlining_decision_type.is_stub decision
+  then
+    Definition_says_inline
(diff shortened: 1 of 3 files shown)


Gbury (approved):
Should probably be merged after #2280 to avoid potential infinite loops ?

middle_end/flambda2/simplify/inlining/call_site_inlining_decision.ml
```diff
@@ -142,14 +142,22 @@ let argument_types_useful dacc apply =
...
-    ~return_arity : Call_site_inlining_decision_type.t =
+    ~return_arity ~rec_depth_exceeded : Call_site_inlining_decision_type.t =
   let denv = DA.denv dacc in
   let env_prohibits_inlining = not (DE.can_inline denv) in
   let decision =
     Code_or_metadata.code_metadata code_or_metadata
     |> Code_metadata.inlining_decision
   in
-  if Function_decl_inlining_decision_type.must_be_inlined decision
+  if Function_decl_inlining_decision_type.is_stub decision
```
lukemaurer, line 152: Since we already have the code metadata in hand, is there a reason to say `is_stub decision` here rather than just check whether the function is a stub? It seems much clearer for the policy to be “inline stubs even if it would exceed the max recursion depth” than “inline those functions for which the definition-site inlining decision was true for reasons of being a stub, even if it would exceed the max recursion depth.”

lukemaurer (changes requested):
Just to check my understanding: The effect here is to cause a stub to be inlined even if doing so may exceed the max recursion depth? (It's a sound policy, though for somewhat subtle reasons.)

lukemaurer:
> Should probably be merged after #2280 to avoid potential infinite loops ?

I believe they're largely orthogonal, actually, since this only kicks in if there isn't an `[@inlined]` annotation on the call site to the stub. The only way there would be an infinite loop is if it's possible for stubs to form a loop (possibly a single self-recursive one) that goes through stubs alone. _Is_ that possible? (If so, I don't think #2280 helps unless an `[@inlined]` is involved.)

PR #2295, opened 2024-02-16
Continuation lifting

Gbury:
# Introduction

This Pull Request introduces continuation lifting. Basically, consider a term of the form `letk k x = letk k' y = ... in ...`, the idea is to transform it into a term of the form `letk k' y x = ... in letk k x = ....`, where `k'` has been lifted out of its original parent `k`. The ultimate goal of this transformation is then to allow us to specialise and duplicate `k` without needing to tuplicate `k'`. This will be particularly important for the upcoming match-in-match optimization.

The utlimate goal of continuation lifting is to make continuation duplication/specialization easier. Considering that the purpose of continuation specialization is to have more precise types, this means that continuation specialization should be done on the way down, and therefore that continuation lifting should also be done on the way down, in a way that allows the future specialization code to work properly. This is not that easy, and for the context, we have tried multiple designs that would not work before arriving at the current one.

# Design overview

Before talking about lifting properly, we need to establish how specialization will work, and then show how lifting integrates  with it. Specialization is split into two parts: first is the decision to specialize, which needs enough information to make an appropriate decision, and then is the actual specialization, which needs to occur on the way down.

The goal for the decision of whether to specialize a continuation `k` is to wait until we have seen `k`'s uses, as well as its handler, but just before we start going down the handlers of the continuation defined inside of `k`. This matches very well with the current traversal order of Simplify. In particular, this will enable specialization of continuations whose handler ends with a switch, and where all (or at least some) call sites have enough information to reduce the switch to a single arm.

Once the decision to specialize a continuation `k` is done, we then enter a  `lifting mode` where the goal is to lift all the continuations `k'` that are defined in the handler of `k` (to ensure those are not duplicated when we duplicate the handler of `k`). We then store all of the non-simplified handlers of lifted continuations in the `dacc`. Once we reach the `down_to_up` for `k`, we then pop all of the lifted_continuations from the `dacc`, and then proceed to simplify their handlers, before then rebuilding all of the let-conts. This delaying may seem useless, as currently, I don't think there is any work done between the moment we store lifted continuations in the `dacc`, and when we reach the `down_to_up`, however in the future, specialization of continuation will duplicate the handler of `k` between these two points.

# Technical details

## Additional arguments

Lifting continuation effectively changes the scope in which they are bound. In particular, when lifting `k'` out of `k`, before lifting `k'` has access to the parameters of `k` as well as any variable let-bound in `k` before `k'` is bound, however, after lifting these parameters/variables are not in scope anymore. To solve this issue the lifting process needs to add parameters to `k'`. Unfortunately, in order for these parameters to be correctly seen (and the typing env to contain the correct equation for them), we cannot re-use the existing `extra_params_and_args` mechanism which only works on the upwards pass. In addition, when lifting multiple continuations, say `k'` and then `k''` out of `k`, we need to be careful to record enough information to rewrite the calls to `k'` inside of `k''`. To that effect, a notion of `lifted continuation parameter` is added to record all of these new parameters, and index them so that the corresponding arguments can be computed as needed. We also keep track of all variables defined in the current scope, so that we know which parameters to add to lifted continuations.

As a result of having additional parameters, some of these can actually be regions and/or rec_infos. This is not a problem as the only change needed for these is in `to_cmm`: parameters and arguments of kind rec_infos are eliminated, and regions are just integers.

## Denvs

As said just above, the lifting process changes the scope of a continuation, which means one must be careful about crafting the correct `denv` before simplifying a lifted continuations. Indeed some elements of the `denv` follow the lexical scope, and for these we must use the values that are true at the point where we will insert the continuation, instead of the values from the point where the continuation was originally defined. Most of those decisions can be seen in the `DE.denv_for_lifted_continuations` function.

## Typing env

The typing env has a notion of scope, which is used when joining typing envs, and which is increased before going down a continuation. Say we are at scope `n` before a let-cont, before this PR, we explore the body of the let-cont at level `n+1`, and then join and explore the handler at level `n` (unless I missed something, cc @lthls ). With that strategy, one needs one level for each `let-cont`. Unfortunately with continuation lifting, when we increase the scope, but we would need to increase the scope by the number of continuations lifted, which we have no way of knowing at that point, so another strategy is needed.

The new strategy is explained in a comment in `simplify_let_cont`, and it is the following:
- suppose we are at scope `n` and we see a let-cont
- we increase scope twice before inspecting the body, so the scope is `n+2`
- we use a denv at fork of scope `n+ 1` to perform the join, so that the join correctly detects the difference between the body of the let cont, and the env for the handler of the continuation
- we thus get an env with scope `n+1` after the join, but we actually bump the scope before entering the continuation's handler. The `bump` operation actually moves everything that was at level `n +1` and it puts it at level `n+2` (this is actually a very cheap operation on typing envs). Note that it is very different from increasing the scope.
- that way, for any lifted continuation, we get in the accumulator an env at scope `n + 2`, which we join at level `n+1`, and then bump the scope before exploring the lifted handler.

# Lifting criterion and testing

Remember that this PR only implements continuation lifting. Without continuation specialization, there are no optimizations that require or benefit from continuation lifting, so there technically is no reason to do any lifting. That being said, in order to test that lifting works as intended, this PR currently implements a dummy criterion that acts as if every continuation ending with a switch was specialized, and therefore lifts any continuation `k'` defined inside another continuation `k`, when `k` ends with a switch. This is more than we'll ever likely want to, and it indeed appears to slow down the compiler noticeably. We nevertheless intend to keep this dummy criterion, at least for now, so that proper testing of the lifting process alone can be carried out. It would also be interesting to collect data about the potential slowdown incurred when lifting way too many continuations as this does.

There might be some slight code cleanup to do (some unrelated patchs may have found their way into this branch), but this PR can be reviewed, and it would be nice to have it tested. (there is currently a failure on `unroll4.ml`, but we think this is a problem related to fexpr's handling of symbols, which already needed a fix in this PR).


Gbury:
CI is green, this is ready for review and testing @mshinwell

lthls:
The chamelon changes look independent, would it make sense to extract them into a separate PR that we can merge quickly ?

Ekdohibs:
Note that an equivalent version of the chamelon changes has already been done (and pushed to main) so we can remove those changes from this PR.

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
```diff
@@ -919,63 +1074,99 @@ let sort_handlers data handlers =
...
-    ~arg_types_by_use_id =
+    ~lifted_params ~arg_types_by_use_id =
   (* In the recursive case, [params] are actually the invariant params, as we
      prepare a common dacc for all handlers. *)
   let join_result =
+    (* We add to the continuation uses (and therefore to the handler env) the
+       params added by the eventual lifting of continuations. However, we do not
+       want those lifted params to be unboxed: it should not be beneficial
+       (since the lifted params should already include unbxoed versions). *)
+    let lifted_params, uses =
```
chambart, line 1157: This might not be true for variables bound before the continuation, but that are not argument an argument of the parent continuation.

middle_end/flambda2/types/env/typing_env.ml
```diff
@@ -146,10 +154,12 @@ let [@ocamlformat "disable"] print ppf
     let levels =
       current_level :: prev_levels
     in
+    (*
     let levels =
       List.filter (fun level -> not (One_level.is_empty level))
         levels
     in
+    *)
```
chambart, line 162: Why ? Should this be really removed ?

middle_end/flambda2/to_cmm/to_cmm_shared.mli
```diff
@@ -16,12 +16,18 @@
     this module, unlike the ones in [Cmm_helpers], depend on Flambda 2 data
     types. *)
 
+val actual_params :
+  (Backend_var.With_provenance.t * Cmm.machtype To_cmm_env.param_type) list ->
+  (Backend_var.With_provenance.t * Cmm.machtype) list
+
+val actual_args : 'a list -> _ To_cmm_env.param_type list -> 'a list
```
chambart, line 23: maybe `remove_skip_params`

middle_end/flambda2/simplify/flow/flow.mli
```diff
@@ -19,6 +19,9 @@ module Acc : sig
   (** printing *)
   val print : Format.formatter -> t -> unit
 
+  (** TMP *)
```
chambart, line 22: TMP ?

middle_end/flambda2/simplify/env/downwards_env.ml
```diff
@@ -577,3 +604,63 @@ let with_code_age_relation code_age_relation t =
...
+       fields *)
+    round = denv.round;
+    propagating_float_consts = denv.propagating_float_consts;
+    unit_toplevel_return_continuation = denv.unit_toplevel_return_continuation;
+    unit_toplevel_exn_continuation = denv.unit_toplevel_exn_continuation;
+    do_not_rebuild_terms = denv.do_not_rebuild_terms;
+    closure_info = denv.closure_info;
+    get_imported_code = denv.get_imported_code;
+    loopify_state = denv.loopify_state
+  }
```
chambart, line 666: I would prefer if the order of fields would be all the denv.stuff before then those from denv_for_join.

middle_end/flambda2/simplify/are_lifting_conts.ml
```diff
@@ -0,0 +1,46 @@
...
+  | Not_lifting
+  | Analyzing of
+      { continuation : Continuation.t;
+        uses : Continuation_uses.t
+      }
+  | Lifting_out_of of { continuation : Continuation.t }
+
+let no_lifting : t = Not_lifting
+
+let think_about_lifting_out_of continuation uses =
```
chambart, line 26: the function name should be the same as the constructor name

middle_end/flambda2/simplify/common_subexpression_elimination.ml
```diff
@@ -105,18 +105,35 @@ end
...
+          | param :: params -> (
+            match
+              TE.get_canonical_simple_exn env_at_use param
+                ~min_name_mode:NM.normal ~name_mode_of_existing_simple:NM.normal
+            with
+            | exception Not_found -> find_param simple params
+            | arg ->
+              if Simple.equal arg simple
+              then Some param
+              else find_param simple params)
```
chambart, line 132: Isn't that List.find_opt ?

middle_end/flambda2/simplify/env/downwards_env.ml
```diff
@@ -46,7 +46,12 @@ type t =
     get_imported_code : unit -> Exported_code.t;
     all_code : Code.t Code_id.Map.t;
     inlining_history_tracker : Inlining_history.Tracker.t;
-    loopify_state : Loopify_state.t
+    loopify_state : Loopify_state.t;
+    continuation_stack : Continuation.t list;
+    variables_defined_in_current_continuation : Lifted_cont_params.t;
+    number_of_continuations_defined_in_current_continuation : int
+        (* CR gbury: maybe rename Lifted_cont_params -> Indexed_bound_parameters
+           ? *)
```
chambart, line 54: I think that this CR might be obsolete

middle_end/flambda2/simplify/env/lifted_cont.ml
```diff
@@ -0,0 +1,98 @@
+(**************************************************************************)
+(*                                                                        *)
+(*                                 OCaml                                  *)
+(*                                                                        *)
+(*                        Guillaume Bury, OCamlPro                        *)
+(*                                                                        *)
+(*   Copyright 2023--2023 OCamlPro SAS                                    *)
+(*   Copyright 2023--2023 Jane Street Group LLC                           *)
```
chambart, line 8: --2024

middle_end/flambda2/simplify/lifted_cont_params.ml
```diff
@@ -0,0 +1,102 @@
...
+(*   Copyright 2023--2023 OCamlPro SAS                                    *)
+(*   Copyright 2023--2023 Jane Street Group LLC                           *)
+(*                                                                        *)
+(*   All rights reserved.  This file is distributed under the terms of    *)
+(*   the GNU Lesser General Public License version 2.1, with the          *)
+(*   special exception on linking described in the file LICENSE.          *)
+(*                                                                        *)
+(**************************************************************************)
+
+(* This is mainly to have indexable unique ids *)
```
chambart, line 16: The comment is not required: the signature tell it.

middle_end/flambda2/simplify/common_subexpression_elimination.ml
Ekdohibs: Fixed in #2628

middle_end/flambda2/simplify/flow/flow.mli
Ekdohibs: Deleted

middle_end/flambda2/to_cmm/to_cmm_shared.mli
Ekdohibs: Fixed in #2627.

middle_end/flambda2/simplify/simplify_switch_expr.ml
```diff
@@ -597,6 +597,34 @@ let simplify_switch0 dacc switch ~down_to_up =
   let condition_dbg =
     DE.add_inlined_debuginfo (DA.denv dacc) (Switch.condition_dbg switch)
   in
+  let dacc =
+    match DA.are_lifting_conts dacc with
+    | Lifting_out_of _ -> Misc.fatal_errorf "this should not happen"
```
chambart, line 602: Hum, either `assert false` or a better message.

middle_end/flambda2/simplify/simplify_switch_expr.ml
```diff
@@ -597,6 +597,34 @@ let simplify_switch0 dacc switch ~down_to_up =
...
+      let cost =
+        DE.number_of_continuations_defined_in_current_continuation denv
+        * Lifted_cont_params.length
+            (DE.variables_defined_in_current_continuation denv)
+      in
+      if DA.get_continuation_lifting_budget dacc < cost
+      then dacc
+      else
+        (* Format.eprintf "Analyzing inside of continuation %a@."
+           Continuation.print continuation; *)
```
chambart, line 619:
```suggestion
```

middle_end/flambda2/simplify/simplify_switch_expr.ml
```diff
@@ -597,6 +597,34 @@ let simplify_switch0 dacc switch ~down_to_up =
...
+        (* TODO/FIXME: implement an actual criterion for when to lift
+           continuations. Currently for testing, we lift any continuation that
+           occurs in a handler that ends with a switch. *)
+        DA.with_are_lifting_conts
+          (DA.decrease_continuation_lifting_budget dacc cost)
+          (Are_lifting_conts.lift_continuations_out_of continuation)
+  in
+  let dacc = DA.register_cont_lifting_params_of_current_continuation dacc in
```
chambart, line 627: Maybe do that only when `Analysing` and not when `Not_lifting`

chambart: I got my answer: NO

middle_end/flambda2/simplify/simplify_apply_expr.ml
```diff
@@ -900,21 +902,25 @@ let simplify_function_call_where_callee's_type_unavailable dacc apply
         (T.unknown_types_from_arity
            (Exn_continuation.arity (Apply.exn_continuation apply)))
   in
-  let call_kind =
+  let call_kind, apply =
```
chambart, line 905: Should this be reverted ?

chambart: Or splitted in another PR

middle_end/flambda2/simplify/simplify_let_expr.ml
```diff
@@ -328,6 +328,21 @@ let simplify_let0 ~simplify_expr ~simplify_function_body dacc let_expr
     let dacc =
       DA.add_to_lifted_constant_accumulator dacc prior_lifted_constants
     in
+    let dacc =
+      List.fold_left
+        (fun dacc (binding : Expr_builder.binding_to_place) ->
+          Bound_pattern.fold_all_bound_vars binding.let_bound ~init:dacc
+            ~f:(fun dacc bound_var ->
+              let v = Bound_var.var bound_var in
+              DA.map_denv dacc ~f:(fun denv ->
```
chambart, line 337: Maybe do the `DA.map_denv` outside of the fold to avoid allocating too many dacc

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
```diff
@@ -1155,40 +1369,52 @@ let simplify_recursive_handlers ~rebuild_body ~invariant_params ~invariant_epa
...
-      after_downwards_traversal_of_body_and_handlers data ~down_to_up dacc
+      k dacc data
     | Some uses ->
+      let dacc =
+        DA.with_are_lifting_conts dacc
+          (Are_lifting_conts.think_about_lifting_out_of cont uses)
+      in
+      (* let _uses = Continuation_uses.get_uses uses in Format.eprintf "USES =
+         %a@." (Format.pp_print_list (fun ff use -> One_continuation_use.print
+         ff use; DE.print ff (One_continuation_use.env_at_use use) ) ) _uses; *)
```
chambart, line 1417: :grin:

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
```diff
@@ -1205,15 +1431,25 @@ let after_downwards_traversal_of_body ~simplify_expr
...
+        prepare_dacc_for_handlers dacc ~env_at_fork:denv ~params ~lifted_params
           ~consts_lifted_during_body ~is_recursive:false
           (Continuation.sort cont)
           (if is_exn_handler then Some cont else None)
           (Continuation_uses.get_uses uses)
           ~arg_types_by_use_id:(Continuation_uses.get_arg_types_by_use_id uses)
       in
+      (* Format.eprintf "Decisions for %a:@\n%a@\ndacc:@\n%a@\n@."
+         Continuation.print cont Unbox_continuation_params.Decisions.print
+         unbox_decisions TE.print (DE.typing_env (DA.denv dacc)); *)
```
chambart, line 1443: :grin:

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
chambart, line 1417:
```suggestion
```

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
chambart, line 1443:
```suggestion
```

middle_end/flambda2/simplify/simplify_apply_expr.ml
Ekdohibs: Now in #2629

middle_end/flambda2/simplify/unboxing/unboxing_epa.mli
```diff
@@ -35,3 +35,6 @@ val add_extra_params_and_args :
   Continuation_extra_params_and_args.t ->
   Unboxing_types.decision ->
   Continuation_extra_params_and_args.t
+
+val compute_extra_params_in_unspecified_order :
+  Bound_parameters.t -> Unboxing_types.decision -> Bound_parameters.t
```
chambart, line 40: From the type and the name, it's not obvious what this function does with it's first argument. I propose to rename it to `add_extra_params_in_unspecified_order`

middle_end/flambda2/simplify/unboxing/unboxing_epa.ml
```diff
@@ -479,3 +479,81 @@ let add_extra_params_and_args extra_params_and_args decision =
       EPA.add extra_params_and_args ~extra_param ~extra_args:epa.args
   in
   aux extra_params_and_args decision
+
+(* Extracting just the extra params *)
+(* ******************************** *)
+
+let compute_extra_params_in_unspecified_order extra_params decision =
```
chambart, line 486: `extra_params` here don't have to be that. I think that calling that variable `acc` would make it more obvious that this is more 'neutral' to the function.

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
```diff
@@ -44,32 +44,36 @@ open! Simplify_import
 (* For each stage, the information received by this stage is of type
    stage_data. *)
 
-type one_recursive_handler =
+type one_recursive_handler = Lifted_cont.one_recursive_handler =
```
chambart, line 47: Does this need to be redefined here ? Is it any better ?

chambart: Same for the rest of that type

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
```diff
@@ -88,16 +92,24 @@ type handler_after_downwards_traversal =
...
     at_unit_toplevel : bool;
+    lifted_params : Lifted_cont_params.t;
     invariant_params : Bound_parameters.t;
     invariant_extra_params_and_args : EPA.t;
     handlers : handler_after_downwards_traversal Continuation.Map.t
   }
 
+type after_downwards_traversal_of_body_and_handlers_data =
+  | Lifted_out of after_downwards_traversal_of_body_and_handlers_lifted
+  | Rebuild of after_downwards_traversal_of_body_and_handlers_rebuild
```
chambart, line 111: A bit strange. One is a verb and the other isn't

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
```diff
@@ -919,63 +1074,99 @@ let sort_handlers data handlers =
...
-  down_to_up dacc ~rebuild:(prepare_to_rebuild_handlers data)
+     handlers. *)
+  match data with
+  | Lifted_out { rebuild_body } ->
+    (* In this case, we want to lift the continuations being defined *)
+    let data : prepare_to_rebuild_handlers_data = Lifted_out { rebuild_body } in
+    down_to_up dacc ~rebuild:(prepare_to_rebuild_handlers data)
+  | Rebuild data ->
+    let dacc, down_to_up =
+      (* Format.eprintf "LIFTOUT %a@." DA.print dacc; *)
```
chambart:
```suggestion
```

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
```diff
@@ -833,9 +981,11 @@ let get_uses (data : after_downwards_traversal_of_body_and_handlers_data) cont =
       Continuation.print cont
   | Some cont -> cont
 
-let create_handler_to_rebuild
-    (data : after_downwards_traversal_of_body_and_handlers_data) cont
+let create_handler_to_rebuild cont_lifting_params
+    (data : after_downwards_traversal_of_body_and_handlers_rebuild) cont
     (handler : handler_after_downwards_traversal) =
+  (* See comment at the top of
+     [after_downwards_travsersal_of_body_and_handlers]. *)
```
chambart, line 983: The comment is not at the top of `after_downwards_travsersal_of_body_and_handlers` now

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
```diff
@@ -1290,58 +1593,36 @@ let simplify_let_cont0 ~simplify_expr dacc (data : simplify_let_cont_data)
...
-          (fun handler ->
-            let is_cold = CH.is_cold handler in
-            CH.pattern_match handler ~f:(fun params ~handler ->
-                { params; handler; is_cold }))
-          handlers
-      in
-      { body; handlers = Recursive { invariant_params; continuation_handlers } }
-  in
-  simplify_let_cont0 ~simplify_expr dacc data ~down_to_up
+  let body, handlers = split_let_cont let_cont in
```
chambart, line 1612: Just above the line: the comment is obsolete. It seems it was before this PR.

chambart: `split_let_cont` is not large and the name is not very informative. This code would probably be better with it inlined

chambart: Maybe not inline, rather find a better nam.

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
```diff
@@ -1290,58 +1581,36 @@ let simplify_let_cont0 ~simplify_expr dacc (data : simplify_let_cont_data)
...
-    DA.with_denv dacc (DE.increment_continuation_scope denv_before_body)
+  (* About scopes: supposing we are at scope 'n' before the let-cont, we will:
+
+     - use scope 'n + 2' to inspect the body
+
+     - use scope 'n + 1' to perform the join of continuation uses
+
+     - bump the scope of the env after join to get an env at scope 'n + 2'
+
+     - recursively for lifted continuations, whose join will take place at level
```
chambart, line 1592: This is a strange sentence starting with 'recursively'

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
```diff
@@ -1274,14 +1505,74 @@ let after_downwards_traversal_of_body ~simplify_expr
...
+      Lifted_out { rebuild_body }
+    in
+    after_downwards_traversal_of_body_and_handlers ~simplify_expr ~denv_for_join
+      data ~down_to_up dacc
+  | Not_lifting | Analyzing _ ->
+    simplify_handlers ~simplify_expr data dacc ~rebuild_body (fun dacc data ->
+        after_downwards_traversal_of_body_and_handlers ~simplify_expr
+          ~denv_for_join
+          (Rebuild data : after_downwards_traversal_of_body_and_handlers_data)
+          ~down_to_up dacc)
```
chambart, line 1548: For readability purpose, I would rather have simplify handler build his own continuation rather than building it here. It just need `denv_for_join` and `down_to_up` as additionnal arguments

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
```diff
@@ -1155,40 +1363,49 @@ let simplify_recursive_handlers ~rebuild_body ~invariant_params ~invariant_epa
...
-    (data : after_downwards_traversal_of_body_data) ~down_to_up dacc
-    ~rebuild:rebuild_body =
-  (* At this point, we have done the downwards traversal of the body. We prepare
-     to loop over all the handlers defined by the let cont. *)
+and simplify_handlers ~simplify_expr ~rebuild_body
+    (data : after_downwards_traversal_of_body_data) dacc k =
+  (* In this case we have decided not to lift the continuation being let-bound
+     outside of its context. So the remaining thing to do is setup the dacc, and
+     later decide whether to lift any continuation the is defined inside the
+     continuation being currentl let-bound. *)
```
chambart, line 1375:
```suggestion
     later decide whether to lift the continuations defined inside the
     continuation being currently let-bound. *)
```

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
```diff
@@ -1030,15 +1219,25 @@ let prepare_dacc_for_handlers dacc ~env_at_fork ~params ~is_recursive
...
-    ~invariant_params ~params cont dacc handler k =
+and simplify_handler ~simplify_expr ~is_recursive ~is_exn_handler ~lifted_params
+    ~invariant_params ~params ~all_extra_params cont dacc handler k =
+  let all_params = Bound_parameters.append invariant_params params in
+  let dacc = DA.map_denv dacc ~f:(DE.enter_continuation cont lifted_params) in
+  let dacc =
+    DA.map_denv dacc ~f:(fun denv ->
+        List.fold_left
+          (fun denv bp ->
+            DE.add_variable_defined_in_current_continuation denv bp)
```
chambart, line 1230:
```suggestion
        List.fold_left DE.add_variable_defined_in_current_continuation
```

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
```diff
@@ -1030,15 +1219,25 @@ let prepare_dacc_for_handlers dacc ~env_at_fork ~params ~is_recursive
     is_exn_handler,
     join_result.extra_params_and_args )
 
-let simplify_handler ~simplify_expr ~is_recursive ~is_exn_handler
-    ~invariant_params ~params cont dacc handler k =
+and simplify_handler ~simplify_expr ~is_recursive ~is_exn_handler ~lifted_params
+    ~invariant_params ~params ~all_extra_params cont dacc handler k =
+  let all_params = Bound_parameters.append invariant_params params in
```
chambart, line 1224: `all_params` don't contains the extra_params. I don't have a good idea for the name, but it's a bit confusing

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
```diff
@@ -1235,17 +1459,22 @@ let after_downwards_traversal_of_body ~simplify_expr
...
               invariant_extra_params_and_args = EPA.empty;
               handlers = Continuation.Map.singleton cont rebuild;
               at_unit_toplevel
             }
           in
-          after_downwards_traversal_of_body_and_handlers data ~down_to_up dacc))
-  | Recursive { continuation_handlers; invariant_params } ->
+          k dacc data))
+  | Recursive { continuation_handlers; invariant_params; lifted_params } ->
+    let dacc = DA.with_are_lifting_conts dacc Are_lifting_conts.no_lifting in
```
chambart, line 1477: This would need a comment to justify that we don't lift inside a recursive continuation.

chambart: And maybe this would be a good way to test some seldomly used other codes. ;)

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
```diff
@@ -919,63 +1069,98 @@ let sort_handlers data handlers =
       group :: inner)
     [] sorted_handlers_from_the_inside_to_the_outside
 
-let after_downwards_traversal_of_body_and_handlers
-    (data : after_downwards_traversal_of_body_and_handlers_data) ~down_to_up
-    dacc =
+let rec after_downwards_traversal_of_body_and_handlers ~simplify_expr
```
chambart, line 1072: All those functions are now recursive. I think it is not required. That's good for review that they didn't move. But reordering them and removing the rec might be better afterwards.
```suggestion
let rec after_downwards_traversal_of_body_and_handlers ~simplify_expr
(* CR pchambart: Those functions don't need to be recursive, it would be nice to fix that later *)
```

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
```diff
@@ -159,6 +177,125 @@ type rebuild_let_cont_data =
...
+  then Bound_parameters.empty, uses
+  else
+    let lifted_params =
+      Lifted_cont_params.bound_parameters callee_lifted_params
+    in
+    let callee_lifted_param_kinds =
+      List.map
+        (fun bound_param ->
+          K.With_subkind.kind (Bound_parameter.kind bound_param))
+        (Bound_parameters.to_list lifted_params)
```
chambart, line 246: I would prefer to have a `Bound_parameters.kinds` function doing this.

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
```diff
@@ -1274,14 +1470,68 @@ let after_downwards_traversal_of_body ~simplify_expr
...
+          ~denv_for_join data ~down_to_up dacc)
+
+and down_to_up_for_lifted_continuations ~simplify_expr ~denv_for_join
+    lifted_conts ~down_to_up =
+  match lifted_conts with
+  | [] -> down_to_up
+  | (denv, handlers) :: other_lifted_handlers ->
+    (* At this point, we are lifting a continuation k' with handler [handlers],
+       out of a continuation k, and: - [denv_for_join] is the denv just before
+       the letk for k - [denv] is the denv just before the letk for k'
```
chambart, line 1516: Formating problem ?
```suggestion
    (* At this point, we are lifting a continuation k' with handler [handlers],
       out of a continuation k, and:
        - [denv_for_join] is the denv just before the letk for k
        - [denv] is the denv just before the letk for k'
```

chambart:
There are many instances of
```ocaml
let dacc = DA.register_cont_lifting_params_of_current_continuation dacc in
```
This is a bit dangerous because this requires to add that at each new case of a tail expression (for instance new cases of Invalid)

To avoid that, we would like to register those at the continuation definition.

Gbury:
Another note: we should add our super continuation lifting testcase to the testsuite.

middle_end/flambda2/simplify/lifted_cont_params.mli
```diff
@@ -0,0 +1,94 @@
...
+(** Add a new parameter *)
+val new_param : t -> Bound_parameter.t -> t
+
+(** Rename all new parameters, and returns the corresponding renaming. *)
+val rename : t -> t * Renaming.t
+
+(** Given a unique id, and a stack of the parent continuations lifted cont params,
+    this function lookups the first occurrence of a new param that maps to the
+    same unique id. *)
+val find_arg : Id.t -> t list -> Simple.t
```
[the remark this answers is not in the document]
Gbury: this code was removed with recent simplifications, so no problem anymore

middle_end/flambda2/simplify/env/downwards_env.mli
```diff
@@ -212,3 +215,17 @@ val loopify_state : t -> Loopify_state.t
 val set_loopify_state : Loopify_state.t -> t -> t
 
 val with_code_age_relation : Code_age_relation.t -> t -> t
+
+val continuation_stack : t -> Continuation.t list
+
+val enter_continuation : Continuation.t -> Lifted_cont_params.t -> t -> t
```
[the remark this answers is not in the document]
Gbury: this code disappeared as well

Gbury:
I have pushed new simplifications to the code, in particular concerning the computation of the extra params and args, the handling of the join, and how and where we store the lifted cont params.

This is therefore good for a final review (I expect some comments about some names added by recent commits, where I sometimes didn't have much inspiration), cc @lthls @NathanReb @chambart

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
```diff
@@ -1492,7 +1483,8 @@ let simplify_let_cont0 ~(simplify_expr : _ Simplify_common.expr_simplifier) dacc
   let denv_for_join = DE.increment_continuation_scope denv_before_body in
   let denv_for_body = DE.increment_continuation_scope denv_for_join in
   let denv_for_body =
-    DE.add_continuation_defined_in_current_continuation denv_for_body
+    let lifting_cost = Lifted_cont_params.length (DE.variables_defined_in_current_continuation denv_for_body) in
```
Ekdohibs: This has potentially quadratic cost, are we worried enough to cache the length in `Lifted_cont_params`?

middle_end/flambda2/simplify/env/downwards_env.ml
```diff
@@ -47,14 +47,15 @@ type t =
...
     loopify_state : Loopify_state.t;
-    continuation_stack : Continuation.t list;
-    variables_defined_in_current_continuation : Lifted_cont_params.t;
-    variables_defined_in_current_continuation' : Lifted_cont_params.t;
     lifted : Variable.Set.t;
-    number_of_continuations_defined_in_current_continuation : int
-        (* this last field makes it possible to count the number of parameters
-           added by the lifting process, and therefore estimate the cost of
-           lifting continuations. *)
+    defined_variables : Lifted_cont_params.t list;
```
lthls, line 51: Maybe `defined_variables_by_scope`, as that reflects a bit better why we're organising them into a list. Also, the comment should explain why we're storing `Lifted_cont_params.t` (basically because it's only used for lifting so that's the more convenient type).

middle_end/flambda2/simplify/env/downwards_env.ml
```diff
@@ -47,14 +47,15 @@ type t =
     all_code : Code.t Code_id.Map.t;
     inlining_history_tracker : Inlining_history.Tracker.t;
     loopify_state : Loopify_state.t;
-    continuation_stack : Continuation.t list;
-    variables_defined_in_current_continuation : Lifted_cont_params.t;
-    variables_defined_in_current_continuation' : Lifted_cont_params.t;
     lifted : Variable.Set.t;
```
lthls, line 50: This field provides a fast access to the set of variables defined in the current continuation (the head of `defined_variables_by_scope` below). It should be moved down in the record and documented as such.

middle_end/flambda2/simplify/join_points.ml
```diff
@@ -24,6 +24,33 @@ type result =
     escapes : bool
   }
 
+let introduce_extra_params_in_use_env epa (env_at_use, use_id, kind) =
```
lthls, line 27: Do you think you could make this work with `env_at_use` being a `DE.t` here ? There's already a CR at line 75 and I think now might be a good moment to address it.

lthls (approved):
I've reviewed everything except `simplify_let_cont_expr.ml`.
I've left various comments but none of them are strictly necessary.

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
```diff
@@ -967,18 +1041,34 @@ let after_downwards_traversal_of_body_and_handlers
...
+  let lifted_cont_extra_params_and_args =
+    extra_params_and_args_for_lifting lifted_params uses
+  in
   let join_result =
+    (* We add to the continuation uses (and therefore to the handler env) the
+       params added by the eventual lifting of continuations. *)
+    (* CR gbury: Some of those params need not be unboxed (the ones that come
+       from continuation original parameters), but others may benefit from being
+       unboxed, for example variables that are let-bound in a continuations's
+       handler. We sould look into whether it's actually useful, and try
```
Ekdohibs, line 1062:
```suggestion
       handler. We should look into whether it's actually useful, and try
```

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
```diff
@@ -967,18 +1041,34 @@ let after_downwards_traversal_of_body_and_handlers
...
+       enabling unboxing for thos (while avoiding re-unboxing parameters that
```
Ekdohibs, line 1063:
```suggestion
       enabling unboxing for those (while avoiding re-unboxing parameters that
```

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
Ekdohibs: Maybe `all_normal_params`? Or we can just revert this part of the diff, now that `all_params` is only used once after refactorings.

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
Gbury: comment added

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
```diff
@@ -1131,13 +1225,14 @@ let simplify_recursive_handlers ~rebuild_body ~invariant_params ~invariant_epa
...
         { rebuild_body;
           cont_uses_env = cont_uses_env_so_far;
+          lifted_params;
           invariant_params;
           invariant_extra_params_and_args = invariant_epa;
           handlers = simplified_handlers;
           at_unit_toplevel = false
         }
       in
-      after_downwards_traversal_of_body_and_handlers data ~down_to_up dacc
```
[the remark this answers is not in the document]
Gbury: done

middle_end/flambda2/simplify/join_points.ml
Gbury: done

middle_end/flambda2/simplify/env/downwards_env.mli
```diff
@@ -213,19 +213,15 @@ val set_loopify_state : Loopify_state.t -> t -> t
 
 val with_code_age_relation : Code_age_relation.t -> t -> t
 
-val continuation_stack : t -> Continuation.t list
+val defined_variables : t -> Lifted_cont_params.t list
 
-val reset_lifted_params : Lifted_cont_params.t -> t -> t
-
-val enter_continuation : Continuation.t -> Lifted_cont_params.t -> t -> t
+val enter_continuation_handler_and_reset_lifted_params :
```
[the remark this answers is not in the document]
Gbury: renamed

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
```diff
@@ -1162,36 +1257,45 @@ let simplify_recursive_handlers ~rebuild_body ~invariant_params ~invariant_epa
...
+          lifted_params = Lifted_cont_params.empty;
           invariant_params = Bound_parameters.empty;
           invariant_extra_params_and_args = EPA.empty;
           handlers = Continuation.Map.empty;
           at_unit_toplevel = false
         }
       in
-      after_downwards_traversal_of_body_and_handlers data ~down_to_up dacc
+      after_downwards_traversal_of_body_and_handlers data dacc
+        ~simplify_expr ~denv_for_join ~down_to_up
```
Ekdohibs, line 1293: I'm wondering whether we should include `denv_for_join` in `data` here (I don't have a clear answer though).

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
Gbury: I added a length cache in `Lifted_cont_params`

middle_end/flambda2/simplify/simplified_named.ml
```diff
@@ -83,3 +83,20 @@ let print ppf { named; _ } = Named.print ppf (to_named named)
 let cost_metrics { cost_metrics; _ } = cost_metrics
 
 let update_cost_metrics cost_metrics t = { t with cost_metrics }
+
+let kind ~tenv { named; _ } =
```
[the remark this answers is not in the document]
Gbury: Actually, there is one case where a useful subkind is returned: in the const case, when the constant is a tagged immediate, the appropriate kind with subkind is returned.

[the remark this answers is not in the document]
lthls: It looks like this function isn't even used, so just remove it

Gbury:
The current CI failure is due to a bug in the typing env that leads to some missing aliases after a join (probably related/introduced to #2705 ), @lthls is investigating/writing a fix.

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
Gbury: I've done what @Ekdohibs suggested

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
Gbury: I manually inlined  `split_let_cont`. I think that at one point, `split_let_cont` was used at multiple points in the file, but that has now stopped being the case, so indeed there is no reason to keep it as a separate function.

middle_end/flambda2/simplify/inlining/call_site_inlining_decision.ml
```diff
@@ -48,6 +48,7 @@ let speculative_inlining dacc ~apply ~function_type ~simplify_expr ~return_arity
     DA.map_denv dacc ~f:(fun denv ->
         DE.set_do_not_rebuild_terms_and_disable_inlining denv)
   in
+  let dacc = DA.with_are_lifting_conts dacc Are_lifting_conts.no_lifting in
```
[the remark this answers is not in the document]
Gbury: done

Gbury:
Can we merge this ? @mshinwell

Gbury:
rebased

Gbury:
This has been reviewed and the only failing test fixed, so I think we can merge this.

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
```diff
@@ -1483,8 +1464,10 @@ let simplify_let_cont0 ~(simplify_expr : _ Simplify_common.expr_simplifier) dacc
 
      - bump the scope of the env after join to get an env at scope 'n + 2'
 
-     - recursively for lifted continuations, whose join will take place at level
-     'n + 1' (which is the level for denv_for_join) *)
+     - we repeat this process for each lifted continuation, whose join will take
+     place at level 'n + 1' (which is the level for denv_for_join), after which
+     the scope will be bumped to `n +2`, so that further lifted continuations
```
Ekdohibs, line 1469:
```suggestion
     the scope will be bumped to `n + 2`, so that further lifted continuations
```

Ekdohibs (approved):

PR #2296, opened 2024-02-16
Define Hd_val separately on x86-64

mshinwell:
The runtime5 definition of `Hd_val` has been shown to disturb inlining heuristics on at least one C compiler.  On x86-64, I believe the relaxed atomic definition compiles to the same thing as the runtime4 version of this macro, so this patch uses the latter for runtime5.

I checked manually that the effects on `caml_string_compare` previously seen with runtime5 (in particular part of it being moved out into another function) did not appear after this patch.


stedolan:
Is there more of an explanation of what's going on here? Generally, changing an atomic access to a nonatomic one can break things, so I'm very skeptical of this change.

PR #2297, opened 2024-02-16
Remove Product lattice

goldfirere:
This refactoring was generally agreed upon with @riaqn. Though it has proved unnecessary, I did this in the course of a mistaken route through modal kinds. Yet having done it, we may as well keep it, given that we wanted to do this regardless.

This is just a refactoring. Should have no other effect.

Review: @riaqn.

### DIFF
--- a/ocaml/typing/mode.ml
+++ b/ocaml/typing/mode.ml
@@ -46,75 +46,6 @@ module Global_flag = struct
     | Global, Global | Unrestricted, Unrestricted -> 0
 end
 
-(* This module is too general and should be specialized in the future.
-   https://github.com/ocaml-flambda/flambda-backend/pull/1760#discussion_r1468531786
-*)
-module Product = struct
-  type ('a0, 'a1) t = 'a0 * 'a1
-
-  (* type aware indexing into a tuple *)
-  type ('a0, 'a1, 'a) axis =
-    | Axis0 : ('a0, 'a1, 'a0) axis
-    | Axis1 : ('a0, 'a1, 'a1) axis
-
-  let print_axis : type a0 a1 a. Format.formatter -> (a0, a1, a) axis -> unit =
-   fun ppf -> function
-    | Axis0 -> Format.fprintf ppf "0"
-    | Axis1 -> Format.fprintf ppf "1"
-
-  let proj (type a0 a1 a) : (a0, a1, a) axis -> a0 * a1 -> a = function
-    | Axis0 -> fun (x, _) -> x
-    | Axis1 -> fun (_, x) -> x
-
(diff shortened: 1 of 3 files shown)


ocaml/typing/solver_intf.mli
```diff
@@ -178,7 +178,7 @@ module type Lattices_mono = sig
     ('a0, 'a1) Misc.eq option
 
   (** Print morphism *)
-  val print_morph : 'b obj -> Format.formatter -> ('a, 'b, 'd) morph -> unit
+  val print_morph : Format.formatter -> ('a, 'b, 'd) morph -> unit
```
riaqn, line 181: `morph` alone is not sufficient to uniquely identifies a morphism - it needs to be accompanied by the destination target.

For example, current `Const_min` is printed as just `const_min` - but I think it would be much more informative if we print `const_min_locality` instead.

Even if our `mode.ml` implementation doesn't need `dst`, we should keep the original interface of `solver_intf.mli` which is more correct.

ocaml/typing/mode.ml
```diff
@@ -349,22 +296,28 @@ module Lattices_mono = struct
...
-        -> ('a, ('a0, 'a1) Product.t, 'l * disallowed) morph
-        (** Maps to minimum product except the given axis *)
+    | Proj_areality :
+        'a and_comonads obj
+        -> ('a and_comonads, 'a, 'l * 'r) morph
+        (** Project the areality component from the comonads *)
+    | Proj_linearity :
+        'a and_comonads obj
+        -> ('a and_comonads, Linearity.t, 'l * 'r) morph
+        (** Project the linearity component from the comonads *)
```
riaqn, line 306: `Proj` is splitted into two, which is less abstraction. This makes `maybe_compose` more complex and will get worse as we add more axes. I'm refactoring this and I think it can work.

riaqn: I have pushed the refactoring - see if you like it :smiley:

ocaml/typing/solver_intf.mli
goldfirere: Nope -- not after perf gain. After simplicity. But I think you're right about keeping the interface with the obj. I'll make that change.

ocaml/typing/mode.ml
goldfirere: I do! There's a few edits I want to try out, but I think this is the right move.

goldfirere: My edits all ended in disaster. (I was trying to get rid of the `assert false` you added. But it was not to be.) Let's just go with this as is.

riaqn (approved):
LGTM. Let's just wait the CI to green.

PR #2213, opened 2024-01-08
The reaper

Ekdohibs:
This PR, written with @chambart, adds a first version of a global pass executed after simplify, used to perform some cleanup work made possible when we look at a compilation unit globally.

While the code is not production-ready yet and can be improved (and the code produced can be further improved as well), it should be ready for a first round of testing.

PR #2299, opened 2024-02-19
Allow equations on parameters in result types

lthls:
This PR increases the amount of information we store in result types.
Previously we would only store the equations on the result variables (made suitable for any environment), with this patch we also add equations on parameters.
This change can be observed on the following examples:
```ocaml
type t = A of int
let f (A x) = x

let swap p = (snd p, fst p)
```
Without this PR the result type for `f` is any immediate, and for `swap` it is a tuple with arbitrary contents.
With this PR the relations between parameters and results are fully expressed in the result types.

This is not complete, of course: if equations on intermediary variables are needed, they are still discarded at the moment.
This can be seen on the following example:
``` ocaml
let g x =
  let tmp = (x, x) in
  tmp, tmp
```
Here we will likely get a result type of the form `∃tmp. result = (tmp, tmp)`, but without equations on `tmp`.

Whether this PR is actually useful is practice is not completely clear, but for #2213 we were looking at what happens in corner cases and this PR makes it easier to generate tricky examples.

### DIFF
--- a/middle_end/flambda2/simplify/simplify_set_of_closures.ml
+++ b/middle_end/flambda2/simplify/simplify_set_of_closures.ml
@@ -276,9 +276,11 @@ let compute_result_types ~is_a_functor ~is_opaque ~return_cont_uses
         ~is_recursive:false ~params:return_cont_params ~env_at_fork
         ~consts_lifted_during_body:lifted_consts_this_function
     in
+    let bound_params_and_results =
+      Bound_parameters.append params return_cont_params
+    in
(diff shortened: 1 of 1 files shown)


chambart (approved):
This is quite sensible and the risk of size increase for the result type is quite limited.
And there are some very reasonable examples where this is quite naturally useful.

chambart:
For instance
```ocaml
module F(X) = struct
  let compare = X.compare
  let f x y =
    compare x y
end
```

PR #2300, opened 2024-02-20
Remove `exact` from `expected_mode`

riaqn:
As title.

### DIFF
--- a/ocaml/testsuite/tests/typing-local/exclave.ml
+++ b/ocaml/testsuite/tests/typing-local/exclave.ml
@@ -214,6 +214,27 @@ Error: This function or one of its parameters escape their region
        when it is partially applied.
 |}]
 
+let f () =
+  exclave_ (
+    (fun x -> function | "a" -> () | _ -> ()) : (string -> string -> unit)
+  )
+[%%expect{|
+Line 3, characters 4-45:
+3 |     (fun x -> function | "a" -> () | _ -> ()) : (string -> string -> unit)
+        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Error: This function or one of its parameters escape their region
+       when it is partially applied.
+|}]
+
+(* For n-ary functions, inner functions are not constrained *)
+let f () =
+  exclave_ (
+    (fun x -> fun y -> ()) : (string -> string -> unit)
+  )
+[%%expect{|
+val f : unit -> local_ (string -> (string -> unit)) = <fun>
+|}]
+
 let f : local_ string -> string =
   fun x -> exclave_ s
 [%%expect{|
--- a/ocaml/testsuite/tests/typing-local/exclave.ml
+++ b/ocaml/testsuite/tests/typing-local/exclave.ml
@@ -226,7 +226,7 @@ Error: This function or one of its parameters escape their region
        when it is partially applied.
 |}]
 
-(* For n-ary functions, inner functions are not constrained *)
+(* For nested functions, inner functions are not constrained *)
 let f () =
   exclave_ (
     (fun x -> fun y -> ()) : (string -> string -> unit)
--- a/ocaml/testsuite/tests/typing-local/local.ml
+++ b/ocaml/testsuite/tests/typing-local/local.ml
@@ -2767,6 +2767,15 @@ Error: This function or one of its parameters escape their region
        when it is partially applied.
 |}];;
 
+let f () = ((fun x -> function | 0 -> x | y -> x + y) : (local_ int -> (int -> int)));;
+[%%expect{|
+Line 1, characters 12-53:
+1 | let f () = ((fun x -> function | 0 -> x | y -> x + y) : (local_ int -> (int -> int)));;
+                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Error: This function or one of its parameters escape their region
+       when it is partially applied.
+|}];;
+
 (* ok if curried *)
 let f () = ((fun x -> (fun y -> x + y) [@extension.curry])
             : (local_ int -> (int -> int)));;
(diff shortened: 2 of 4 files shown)


ocaml/testsuite/tests/typing-local/local.ml
```diff
@@ -2767,6 +2767,15 @@ Error: This function or one of its parameters escape their region
        when it is partially applied.
 |}];;
 
+let f () = ((fun x -> function | 0 -> x | y -> x + y) : (local_ int -> (int -> int)));;
```
goldfirere, line 2770: I think I expect this to be accepted, because of mode-crossing. Maybe we can't mode-cross in the middle of an n-ary function? But I don't see why.

ocaml/testsuite/tests/typing-local/exclave.ml
```diff
@@ -214,6 +214,27 @@ Error: This function or one of its parameters escape their region
...
+  )
+[%%expect{|
+Line 3, characters 4-45:
+3 |     (fun x -> function | "a" -> () | _ -> ()) : (string -> string -> unit)
+        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Error: This function or one of its parameters escape their region
+       when it is partially applied.
+|}]
+
+(* For n-ary functions, inner functions are not constrained *)
```
goldfirere, line 229:
```suggestion
(* For nested functions, inner functions are not constrained *)
```

ocaml/testsuite/tests/typing-local/local.ml
```diff
@@ -2784,6 +2793,26 @@ Error: This function or one of its parameters escape their region
...
+let f () = local_ ((fun x -> function | 0 -> x | y -> x + y) : (_ -> _));;
+[%%expect{|
+Line 1, characters 19-60:
+1 | let f () = local_ ((fun x -> function | 0 -> x | y -> x + y) : (_ -> _));;
+                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Error: This function or one of its parameters escape their region
+       when it is partially applied.
+|}];;
+
+(* For n-ary functions, inner functions are not constrained *)
```
goldfirere, line 2805:
```suggestion
(* For nested functions, inner functions are not constrained *)
```

ocaml/testsuite/tests/typing-unique/unique.ml
```diff
@@ -561,6 +561,21 @@ Error: This function when partially applied returns a once value,
...
+let curry : unique_ box -> (unique_ box -> unique_ box) = fun b1 -> function | b2 -> b2
+[%%expect{|
+Line 1, characters 58-87:
+1 | let curry : unique_ box -> (unique_ box -> unique_ box) = fun b1 -> function | b2 -> b2
+                                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Error: This function when partially applied returns a once value,
+       but expected to be many.
+|}]
+
+(* For n-ary functions, inner functions are not constrained *)
```
goldfirere, line 573:
```suggestion
(* For nested functions, inner functions are not constrained *)
```

ocaml/testsuite/tests/typing-unique/unique.ml
```diff
@@ -561,6 +561,21 @@ Error: This function when partially applied returns a once value,
...
+let no_curry : unique_ box -> (unique_ box -> unique_ box) = fun b1 -> fun b2 -> b2
```
goldfirere, line 574: This presumably works only because of submoding. That is, `fun b1 -> fun b2 -> b1` would be rejected. Maybe add another test?

ocaml/typing/typecore.ml
```diff
@@ -4822,29 +4807,12 @@ let split_function_ty
...
-      begin match
-        Alloc.submode (Alloc.partial_apply alloc_mode) inner_alloc_mode
-      with
-      | Ok () -> ()
-      | Error e ->
-        raise (Error(loc_fun, env, Uncurried_function_escapes e))
-      end;
-      mode_exact (alloc_as_value inner_alloc_mode) inner_alloc_mode,
-      More_args {partial_mode = Alloc.disallow_right inner_alloc_mode}
+      let ret_value_mode = alloc_as_value ret_mode in
```
goldfirere, line 4814: This line is identical in both branches of the `if`. Maybe move above the `if`?

goldfirere (approved):

ocaml/typing/typecore.ml
```diff
@@ -6665,10 +6665,11 @@ and type_function
                   suffix_contains_gadt
               in
               let curry =
-                if is_final_val_param then
+                match fun_alloc_mode with
+                | None ->
+                  assert(is_final_val_param);
```
riaqn, line 6670: Isn't this just replacing option.get exception with assert exception?

ocaml/testsuite/tests/typing-local/exclave.ml
riaqn: I personally think "n-ary" is slightly better than "nested", as it points to a specific thing in our system, whereas "nested" could be vague: `fun a b -> ...` is still a nested function to some people.

ocaml/testsuite/tests/typing-local/local.ml
riaqn: Good point - when typing a function, the `arg_mode` is not mode crossed. I can fix that in a separate PR. Meanwhile, for the tests here I will use `string`. 

Speaking of this, it's really easy to forget spots to cross modes. Maybe we should make `expected_mode.submode` takes a `type_expr` and handle mode crossing in a single spot.

ocaml/testsuite/tests/typing-unique/unique.ml
riaqn: Yeah the example you gave would be rejected. But I think the test is confusing, since the thing we are testing is not related to what the function returns. I have made them all return `unit`.

ocaml/typing/typecore.ml
goldfirere: Yes, but `Option.get` is an anti-pattern, to me. That is, we have `option` so that we don't use `Option.get`.

riaqn: I think the root anti-pattern here is that `is_final_val_param` and `fun_alloc_mode` have duplicated information. And this anti-pattern manifests by either `Option.get` or `assert` - neither is better. But `is_final_val_param` wins slightly for its suggestive name.

riaqn: Well, maybe we can just remove the `assert`? Then I think it looks fine.

goldfirere: I agree about the root anti-pattern here. I looked into fixing that but saw no easy way. I would personally keep the `assert`, because it helps the reader understand what's going on. Another possibility is to revert to the original code here but add a comment saying that the `Option.get` is safe because `is_final_val_param` is false.

I don't actually feel that strongly here. I will Resolve this conversation. Happy to merge with whatever here, if you want to make a change.

ocaml/testsuite/tests/typing-unique/unique.ml
```diff
@@ -542,38 +542,38 @@ type box = { x : int }
 type box = { x : int; }
 |}]
 
-let curry (unique_ b1 : box) (unique_ b2 : box) = b1
+let curry (unique_ b1 : box) (unique_ b2 : box) = ()
```
goldfirere, line 545: Why the changes here? These seem to test a little less than before.

ocaml/typing/typecore.ml
riaqn: Thanks - I added comment and a symmetric `assert`.

ocaml/testsuite/tests/typing-unique/unique.ml
riaqn: I suppose the thing we are testing with this series of `curry` is just the logic around uncurried functions. If we say `b1` or `b2`, it would also test the logic around `Env.lock_mode` and introduce noise. The latter is covered by other tests.

PR #2303, opened 2024-02-20
Add type-directed disambiguation for `local_ expr` expressions

ncik-roberts:
Make it so that type-directed disambiguation works for `local_ expr` expressions. The way we do this is by changing the way `is_inferred` is computed for `local_ expr` expressions, which helps `type_argument` decide to pass the expected type inward rather than trying to infer the type of a function argument.

The relevant comment above `is_inferred`:
```ocaml
(* If [is_inferred e] is true, [e] will be typechecked without using
   the "expected type" provided by the context. *)
```

In metaquot syntax:
  * Prior to this PR, `is_inferred ([%expr local_ [%e e]]) = true` for any `e`.
  * Now, `is_inferred ([%expr local_ [%e e]]) = is_inferred e`

In particular, we now make `is_inferred (local_ expr)` return `false` in some circumstances.

I think the prior state of the world was not intentional. `Pexp_apply` (the choice of encoding for local expressions) happened to map to `true`.

One related error message changes, for constructs like `let pat : _ -> _ -> _ = local_ (fun p1 p2 -> body)`. The new error message seems fine. It's the same error message you get as `let pat = local_ ((fun p1 p2 -> body) : _ -> _ -> _)`, and the error message doesn't appear to be any less relevant in the new place it shows up.

Review:
  * I suggest @riaqn 
  * You can look at the first commit for a test that demonstrates the behavior prior to the code change.

### DIFF
--- a/ocaml/testsuite/tests/typing-local/local.ml
+++ b/ocaml/testsuite/tests/typing-local/local.ml
@@ -2806,11 +2806,11 @@ let foo () =
   let _bar : int -> int -> int = local_ (fun x y -> x + y) in
   ()
 [%%expect{|
-Line 2, characters 33-58:
+Line 2, characters 40-58:
 2 |   let _bar : int -> int -> int = local_ (fun x y -> x + y) in
-                                     ^^^^^^^^^^^^^^^^^^^^^^^^^
-Error: This expression has type int -> local_ (int -> int)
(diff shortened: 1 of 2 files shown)


goldfirere (approved):

ncik-roberts:
@goldfirere reviewed instead of @riaqn&mdash;thanks!

PR #2304, opened 2024-02-20
Speed up compile time for dev builds

alanechang:
Currently running `make compiler` in a fresh clone of the repo takes more than 2mins.

This PR seeks to introduces a `--enable-dev-build` autoconf flag that changes the dune build profile for the second stage from `main` to `dev`. This results in a large speed up in terms of compile time. After this change, `make compiler` takes less than a minute and the concurrent jobs count consistently hovers at a higher value.

`make runtest_upstream` is also faster. Result on my box:

```
--- before ---
real	2m52.009s
user	15m24.579s
sys	2m36.192s
--- after ---
real	1m22.176s
user	14m30.707s
sys	2m35.240s
```

### DIFF
--- a/Makefile.config.in
+++ b/Makefile.config.in
@@ -3,3 +3,4 @@ dune = @dune@
 middle_end = @middle_end@
 coverage = @coverage@
 legacy_layout = @legacy_layout@
+main_build_profile = @main_build_profile@
--- a/configure.ac
+++ b/configure.ac
@@ -66,10 +66,11 @@ AC_ARG_ENABLE([legacy-library-layout],
   [legacy_layout=yes],
   [legacy_layout=no])
 
-AC_ARG_ENABLE([dev-build],
-  [AS_HELP_STRING([--enable-dev-build],
+AC_ARG_ENABLE([dev],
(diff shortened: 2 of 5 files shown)


goldfirere:
This all sounds great! But what are the downsides of a dev build? Does the entire testsuite still pass? Is the resulting compiler slower? There is presumably something worse about it. :)

alanechang:
The entire testsuite still passes. `dev` profile treats a few more warnings as errors so I had to fix a few files (see first commit).

The resulting compiler is probably a bit slower but I don't know enough about dune to comment on the impact. That's why I also measured the compile + test time with `make runtest_upstream` from a clean repo and it seems to be a win. Think it's probably fine to use this flag for development and switch back to the `main` profile for release builds.

goldfirere:
All very promising. It would still be good to have someone knowledgeable about what's going on here to weigh in before we blast ahead. Will probably need to wait until tomorrow given the hour. I'll post on Slack.

mshinwell:
This is ok, but I think `--enable-dev` would be a better flag name.

mshinwell (approved):

PR #2305, opened 2024-02-21
Closing over argument crosses modes

riaqn:


ocaml/typing/typecore.ml
```diff
@@ -6676,6 +6676,10 @@ and type_function
                   Final_arg
                 | Some fun_alloc_mode ->
                   assert(not is_final_val_param);
+                  (* If the argument cross modes, then the inner closure won't
+                     contain a pointer to it, and thus its mode not constrained.
+                     *)
+                  if not (mode_cross env ty_arg) then
```
goldfirere, line 6682: I don't love this, for two reasons:

1. It's unclear when the `if` ends. Maybe use parentheses to show that this applies only to the `close_over` and not the `partial_apply`?
2. It won't work well with modal kinds, where there may be different effects on the different axes. Instead, would it be correct to call `mode_cross_to_min` on `arg_mode` instead? That seems better to me. And it looks like maybe you tried this, given the changes to the type of `mode_cross_to_min`.

riaqn: Good point about parentheses. Yes ocaml parsing always confuses me.

For the second point, I also attempted to use `mode_cross_to_min` first, before realising it's not correct.

`mode_cross_to_min` achieved mode crossing, only if it's *directly on the LHS* of submode. But here, it's not.

In particular, if we set `arg_mode` to `unique` (the lowest mode), it would force `fun_alloc_mode` to be `once` (the highest mode).

Instead, a better way to understand this is: because `ty_arg` is an immediate, so the closure doesn't *really* close over it.

A better interface of `close_over` should take the `ty_arg` and returns `min` if it's immediate, but that would require us to solve some module dependency issue which might not worth it.

Do you think we should amend the comment, explain why we don't use `mode_cross_to_min`?

riaqn: Huh! In fact I now even have a better idea: we should have a `mode_cross_to_min_max` that pushes left comonadic fragment to min, and right monadic fragment to max. This would also match the left/right status of `close_over`.

(There is a symmetric `mode_cross_to_max_min` that pushes right comonadic fragment to max, and left monadic fragment to min, but not useful to us for now).

riaqn: I have pushed the change - see if you like it :smile: 

I also have reversed my changes to `mode_cross_to_min`. I made that change to make it more general, but just realised that the weaker version is safer: basically, we want `mode_cross_to_min` to return a left-only mode, so it can only be on the LHS of submode. I realised this when replying to you above.

goldfirere: Nice how all the types work out there, but I still don't think this is quite right -- at least not with respect to modal kinds.

A modal kind defines an upper bound for each mode. (For immediate, this upper bound is the minimum along each axis. But I think the general case sometimes uses a middle mode as an upper bound.) For a moment I wondered whether we actually want a lower bound for monadic modes, but that doesn't make any sense, really.

I think it's instructive to think about this: `let h = fun (x @ unique) y z -> y ^ z`. (No mode crossing here.) What should the type `h` be? I claim: `'a @ unique -> (string -> string -> string)`. Note the parens: we do _not_ need to make the `string -> string` bit `once`. This is because the unique `x` is unused in the function body. It's a bit like

```ocaml
let g y =
  let x @ unique = ... in
  let f z = y ^ z in
  f
```

We could imagine saying that `f` closes over `x` uniquely; doing so is actually sound. But we don't so this, because `x` is not used in the body of `f`. It's the same with my `h` example above.

Another pair of examples:

```
let k (x @ unique) y z = [x; x; y; z]
let m y =
  let x @ unique = ... in
  let n z = [x; x; y z] in
  n
```

In the second case, `n` closes over `x`... but not uniquely! Similarly, neither does the partial application `k x`. So I claim that the inferred type for `k` should be `'a @ unique -> ('a -> 'a -> 'a list)`, without any `once`.

With this idea for capture in place, we can see a different spot to implement mode-crossing for function parameters: in the occurrence check. That is, an occurrence of a variable at a mode is interesting only if the upper bound for that mode is strictly greater than the measured mode. Instantiating that cryptic sentence tells us that if `x`'s upper bound is `unique`, then `x` cannot occur uniquely -- and thus closing over `x` will not induce a `once`. This already seems to be the case, somehow. When I infer the type for

```
let q _ =
  let use_uniquely (unique_ _u) = () in
  let unique_ x : int = 5 in
  let f z = [z; use_uniquely x] in
  f
```

I get `'a -> unit -> unit list`, but I was expecting `'a -> (unit -> unit list) @ once`.

Does this make any sense?

riaqn: For your last example, it works like that because there is mode crossing in `type_ident`. For your example, it works like this:
- During the `type_ident` on `x`, we first call `Env.lookup_value`, which returns a `mode` which is constrained by the mode of `f`. The constraint is done in `Env.closure_mode`. Basically, if `mode` needs to be `unique`, then `f` needs to be `once`.
- However, we immediately call `mode_cross_to_min` on `mode`, which gives `min` that's not constrained by anything.
- This `min` will be returned as the mode of `x`. It's later required by `use_uniquely` to be `unique`, which is trivial and doesn't involve the mode of `f` at all.

riaqn: For the example `let h = fun (x @ unique) y z -> y ^ z`, my impression is that `h x` will be a closure who runtime representation will contain pointers to `x` and `h`, regardless of whether `x` is used in `h`. In that light, `h x` is like a record containing these two fields (with potential modalities on them if that makes things easier). Therefore, some constraints still need to hold. For example, if either of `x` or `h` is local, the closure `h x` cannot be global - GC might blow up otherwise. Also, `h x` being a closure containing a pointer to `unique x` (even not using it) should be `once`. It sounds dangerous to say "x is not used hence no constraint". Also, I'm pretty sure examples like `h` are quite rare and not worthwhile to fine-tuning our mode constraint.

Another thing to consider is `Alloc.Const.close_over`: When user write `unique_ string -> string -> string`, they almost certainly mean `unique_ string -> once_ (string -> string)`, because the function they are gonna write very likely *do* close over the `unique_ string`. This means `Alloc.Const.close_over` and `Alloc.close_over` will diverge in their behaviour (if we adopt the changes in the paragraph above), which is fine, but we should probably rename `Alloc.close_over` to something else, because it's not really closing over.

goldfirere: Thanks for these explanations.

I agree that we needn't fine-tune for unused parameters. That analysis wasn't really because I care about unused parameters, but rather because I was trying to figure out what the right approach is, and I found it helpful to think about that case.

In the end, the reason I'm pushing back here is because I think the approach here doesn't work in the general modal-kinds world: there will be no `mode_cross` function. Instead, there will be an upper bound for the uniqueness modal axis (among other axes, but they're not important here). So we need to have some operation that works like this (everything considering only the uniqueness axis):

* Let `m` be the mode of the argument.
* Let `b` the upper bound as given in the argument's kind.
* If `b <= m`:
  * Then: ignore `m` in `close_over`; equivalently, set `m` to `max`
  * Else: leave `m` exactly as it is

So we need to be able to say `if b <= m then max else m` (where `b` is a constant).

Hmm.... I had been writing all that thinking that the bit at the end there would be non-monotonic and therefore troublesome. But actually it *is* monotonic, if I'm reading it correctly. Does that morphism look familiar to you? Is it a left or a right? If that all checks out, then it will be possible to proceed along roughly the road you've paved here.

(Separately, I completely agree about `Alloc.Const.close_over`. Even if we did my fine-grained usage analysis, I don't think that should change.)

riaqn: Yes you are exactly right.

#2302 introduces `Meet_with` which is a left-morphism, and therefore it has a right adjoint `Imply`, whose concrete definition, well you guessed it: https://github.com/ocaml-flambda/flambda-backend/blob/9b74d7ec3999b7df0f3e8d34906f297c6db9c43d/ocaml/typing/mode.ml#L111

Say `b` is the upper bound of the kind.

So basically, `alloc_mode_cross_max_min` will apply `Meet_with b` on comonadic axes, and `Imply b` on monadic axes. 

It all works out :100:

goldfirere (approved):

ocaml/typing/typecore.ml
goldfirere: OK. I'm convinced.

But can we merge #2310 first? That's a big patch and annoying to rebase. This will be much easier! :)

riaqn: Sure - #2302 will need rebasing anyway because of recent refactoring, so it can wait a bit.

goldfirere: This doesn't actually depend on #2302, because we can take advantage of the fact that all our axes have only 2 elements, just like I've done elsewhere in #2310. But this is also pretty low priority anyway. So maybe it's easiest to hold this until after #2302 lands, and then we don't have to rewrite this the hacky 2-element way.

riaqn: Then I guess the plan would be:
- merge #2310 
- merge #2302
- merge this one.

PR #2306, opened 2024-02-21
Allow `[@unboxed]` on non-value layouts

alanechang:
Allow `[@unboxed]` on non-value layouts for better upstream compatiablity: after erasing layout annotations, external c stubs should still do the right thing.

The `[@unboxed]` attribute is optional by default, but becomes mandatory when `-only-erasable-extensions` is provided.

Type variables marked `[@layout_poly]` don't need and can't have `[@unboxed]`.

~~Merge after #2229~~ merged

### DIFF
--- a/ocaml/testsuite/tests/typing-layouts-bits32/basics.ml
+++ b/ocaml/testsuite/tests/typing-layouts-bits32/basics.ml
@@ -485,17 +485,17 @@ Error: The native code version of the primitive is mandatory
 
 external f10_6 : (int32#[@unboxed]) -> bool -> string  = "foo" "bar";;
 [%%expect{|
-external f10_6 : int32# -> bool -> string = "foo" "bar"
+external f10_6 : (int32# [@unboxed]) -> bool -> string = "foo" "bar"
 |}];;
 
 external f10_7 : string -> (int32#[@unboxed])  = "foo" "bar";;
 [%%expect{|
-external f10_7 : string -> int32# = "foo" "bar"
+external f10_7 : string -> (int32# [@unboxed]) = "foo" "bar"
 |}];;
 
 external f10_8 : int32 -> int32#  = "foo" "bar" [@@unboxed];;
 [%%expect{|
-external f10_8 : (int32 [@unboxed]) -> int32# = "foo" "bar"
+external f10_8 : int32 -> int32# = "foo" "bar" [@@unboxed]
 |}];;
 
 external f10_9 : (int32#[@untagged]) -> bool -> string  = "foo" "bar";;
--- a/ocaml/testsuite/tests/typing-layouts-bits32/basics.ml
+++ b/ocaml/testsuite/tests/typing-layouts-bits32/basics.ml
@@ -461,7 +461,8 @@ val f9_3 : unit -> int32# t_bits32_id = <fun>
    for uses the typechecker should reject.  In particular
    - if using a non-value layout in an external, you must supply separate
      bytecode and native code implementations,
-   - unboxed types can't be unboxed more.
+   - [@unboxed] is allowed on unboxed types but has no effect. Same is not
+     true for [@untagged].
 *)
 
 external f10_1 : int -> bool -> int32# = "foo";;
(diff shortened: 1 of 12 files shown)


Gbury:
I'm not very fond of accepting `[@unboxed]` annotations on non-value layouts (because the annotation is effectively ignored). Would it be too hard to add the adequate `@[unboxed]` annotations when one erases the layouts ?

alanechang:
It would be difficult to add the adequate `[@unboxed]` annotations when we erase layouts due to examples like:

```
type t = float#
external f : t -> t = ...
```

Without access to type information, ocamlformat can't tell that `t` needs `[@unboxed]`.

Gbury:
Ok, that seems reasonable then.

mshinwell:
I'm not really convinced about this either, will discuss.

alanechang:
This PR is ready for review.

It allows `[@unboxed]` on external declaration argument/return types with non-value layouts.

The attribute would normally be optional, but when in upstream compatible mode (with `-only-erasable-extensions`):

1. `[@unboxed]` becomes required
2. the type with non-value layout must be an alias of `float#`, `int32#`, `int64#`, or `nativeint#` such that after all the unboxed types get erased, the program would be compatible with the upstream compiler.

Request review from @goldfirere

ocaml/testsuite/tests/typing-layouts-bits32/basics.ml
```diff
@@ -484,29 +485,33 @@ Error: The native code version of the primitive is mandatory
 
 external f10_6 : (int32#[@unboxed]) -> bool -> string  = "foo" "bar";;
 [%%expect{|
-Line 1, characters 18-24:
-1 | external f10_6 : (int32#[@unboxed]) -> bool -> string  = "foo" "bar";;
-                      ^^^^^^
-Error: Don't know how to unbox this type.
-       Only float, int32, int64, nativeint, and vector primitives can be unboxed.
+external f10_6 : int32# -> bool -> string = "foo" "bar"
```
goldfirere, line 488: I have a sinking feeling that dropping the attribute in the pretty-printer will come back to bite us. I can't tell you where, but it's just a hunch.

ocaml/testsuite/tests/typing-layouts/erasable_annot.ml
```diff
@@ -263,3 +263,125 @@ module M :
...
+  type t = float#
+end = struct
+  type t = float#
+end
+
+external f_3 : M2.t -> M2.t = "%identity" [@@unboxed];;
+[%%expect{|
+module M2 : sig type t = float# end
+external f_3 : M2.t -> M2.t = "%identity"
+|}];;
```
goldfirere, line 387: `[@unboxed]` works with `private` types. We should test that here, too. That is, I would expect this to work:

```ocaml
module M : sig
  type t = private float#
end = struct
  type t = float#
end

external f_4 : M.t -> M.t = "%identity" [@@unboxed]
```

ocaml/typing/typedecl.ml
```diff
@@ -2170,6 +2172,20 @@ let get_native_repr_attribute attrs ~global_repr =
   | _, Some { Location.loc }, _ ->
     raise (Error (loc, Multiple_native_repr_attributes))
 
+let is_upstream_compatible_non_value_unbox env ty =
```
goldfirere, line 2175: When we have a better treatment for unboxed type names (which should store the base name and the `#` separately), this should change. Please leave a CR (v2.5).

ocaml/typing/typedecl.ml
```diff
@@ -2248,19 +2265,46 @@ let make_native_repr env core_type ty ~global_repr ~is_layout_poly ~why =
...
-  | Native_repr_attr_present kind ->
+  | Native_repr_attr_present kind, (Poly | Sort Value)
+  | Native_repr_attr_present (Untagged as kind), Sort _ ->
     begin match native_repr_of_type env kind ty with
     | None ->
       raise (Error (core_type.ptyp_loc, Cannot_unbox_or_untag_type kind))
     | Some repr -> repr
     end
+  | Native_repr_attr_present Unboxed, (Sort sort) ->
+    (* We allow [@unboxed] on non-value sorts. *)
```
goldfirere, line 2298: Please add a few sentences describing the design here. It's not really a perfect place to put that text, but someone reading this will be curious as to why this is, and writing it here will be helpful. Thanks.

goldfirere (approved):

ocaml/testsuite/tests/typing-layouts-bits32/basics.ml
alanechang: I updated it to print the `[@unboxed]` attributes. It has the slightly unfortunate consequence of printing the attribute on all non-value types even when it's not annotated:

```
external f : float# -> float# = "%identity";;
[%%expect{|
external f : float# -> float# = "%identity" [@@unboxed]
|}]
```

Think this is mostly fine and better than not printing the attribute when explicitly annotated.

This issue is limited to `Printtyp` and does not affect `Pprintast` (used for `-dsource`). To fix it, we need to add a new `Unboxed_same_as_ocaml_repr` constructor which doesn't seem to be worth it.

ocaml/testsuite/tests/typing-layouts/erasable_annot.ml
```diff
@@ -263,3 +263,125 @@ module M :
...
+
+
+external[@layout_poly] id : ('a : any). 'a -> 'a = "%identity" [@@unboxed]
+[%%expect{|
+Line 1, characters 40-42:
+1 | external[@layout_poly] id : ('a : any). 'a -> 'a = "%identity" [@@unboxed]
+                                            ^^
+Error: Don't know how to unbox this type.
+       Only float, int32, int64, nativeint, and vector primitives can be unboxed.
```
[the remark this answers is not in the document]
alanechang: Good point! I updated the error message to what's suggested

ocaml/testsuite/tests/typing-layouts/erasable_annot.ml
alanechang: Added this as a test and it works

ocaml/typing/typedecl.ml
alanechang: added CR

ocaml/typing/typedecl.ml
alanechang: Added a short comment for why we want to allow `[@unboxed]` here

ocaml/testsuite/tests/typing-layouts-bits32/basics.ml
goldfirere: As discussed, let's print only when `-only_erasable_extensions` is set.

alanechang: Made the change in the latest commit

goldfirere (approved):

goldfirere:
I found it confusing that `is_unboxed` is used _only_ for printing. This is important now that it checks for `-only_erasable_extensions` (which is otherwise very unexpected). So I moved the function into `print`, to make this more glaringly obvious.

@alanechang please review my last commit and then merge if happy.

PR #2307, opened 2024-02-21
[draft ]Implicit source positions object system support

Enoumy:
[DRAFT] feature.

This feature adds support for "implicit source positions" (`[%src_pos]` arguments) to OCaml's object system.

Testing
----------
- New tests added to the `typing-implicit-source-positions` test directory
- Additional "drive-by" testing from the "conflict resolution" in the parent feature to sanity check that the unique_ keyword interacts nicely with the src_pos argument.

For reviewers
------------------
There were some place in typeclass.ml that are a bit code duplicate-ey between how src_pos arguments and optional arguments are erased. In places where it was trivial/localized to de-duplicated I de-duplicated, but in typeclass.ml there were a couple of places where making things use a `match` would've involved larger changes and I am unsure if we prefer to make smaller, self contained changes to avoid merge conflicts with upstream OCaml. Please let me know if this is reasonable/isn't.

### DIFF
--- a/ocaml/typing/typeclass.ml
+++ b/ocaml/typing/typeclass.ml
@@ -1328,6 +1328,10 @@ and class_expr_aux cl_num val_env met_env virt self_scope scl =
                    Jkind.Sort.value
                   )
             in
+            let eliminate_position_arg () =
+              let arg = Typecore.src_pos (Location.ghostify scl.pcl_loc) [] val_env in
+              Arg (arg, Jkind.Sort.value)
+            in
             let remaining_sargs, arg =
               if ignore_labels then begin
                 match sargs with
@@ -1343,6 +1347,12 @@ and class_expr_aux cl_num val_env met_env virt self_scope scl =
                              remaining_sargs)
                     then
                       (sargs, eliminate_optional_arg ())
+                    else if
+                      Btype.is_position l &&
+                      not (List.exists (fun (l, _) -> name = Btype.label_name l)
+                             remaining_sargs)
+                    then
+                      (sargs, eliminate_position_arg ())
                     else
                       raise(Error(sarg.pexp_loc, val_env, Apply_wrong_label l'))
               end else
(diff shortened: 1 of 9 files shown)


ocaml/typing/typeclass.ml
```diff
@@ -1327,21 +1337,64 @@ and class_expr_aux cl_num val_env met_env virt self_scope scl =
                    Jkind.Sort.value
                   )
             in
+            let eliminate_position_arg () =
+              (* XXX jrodri: I _think_ this is correct, but I wanted to sanity check.
```
goldfirere, line 1341: Yes, this looks right to me.

ocaml/typing/typeclass.ml
```diff
@@ -1327,21 +1337,64 @@ and class_expr_aux cl_num val_env met_env virt self_scope scl =
...
+              let arg = Typecore.src_pos (Location.ghostify scl.pcl_loc) [] val_env in
+              Arg (arg, Jkind.Sort.value)
+            in
             let remaining_sargs, arg =
               if ignore_labels then begin
                 match sargs with
                 | [] -> assert false
                 | (l', sarg) :: remaining_sargs ->
+                    let label_is_absent_in_remaining_args =
+                      lazy (not (List.exists (fun (l, _) -> name = Btype.label_name l) remaining_sargs))
```
goldfirere, line 1388: Instead of using `lazy`, I think it's conceptually simpler just to make `label_is_absent_in_remaining_args` a function that takes `unit`. The function is only ever called at most once.

ocaml/typing/typeclass.ml
```diff
@@ -1353,9 +1406,12 @@ and class_expr_aux cl_num val_env met_env virt self_scope scl =
                            (Printtyp.string_of_label l));
                     remaining_sargs, use_arg sarg l'
                 | None ->
+                    let is_erased = lazy (List.mem_assoc Nolabel sargs) in
```
goldfirere, line 1409: Ditto here -- no need for `lazy`.

goldfirere (approved):

ocaml/typing/typeclass.ml
Enoumy: Changed! + also waited for CI to pass again which it did!

ocaml/typing/typeclass.ml
Enoumy: Changed!

goldfirere (approved):

PR #2309, opened 2024-02-22
Improve code generation for range checks, etc.

mshinwell:
Current code generation for code such as:
```
let f x bound a b =
  if x >= 0 && x < bound then a else b
```
is not great, and we'd like to fix it right away.  We even reify a boolean:
```
camlRange__f_0_1_code:
  cmpq $1, %rax
  jl .L106
  cmpq %rbx, %rax
  setl %al
  movzbq %al, %rax
  jmp .L102
.align 4
.L106:
  xorl %eax, %eax
.L102:
  testq %rax, %rax
  je .L109
  movq %rdi, %rax
  ret
.align 4
.L109:
  movq %rsi, %rax
  ret
```
This patch aims to catch the relevant pattern during Cmm generation, which I think is probably the right place to do this, being after inlining and constant propagation in particular.  It is replaced by a single unsigned comparison like we do for array bounds checks.  This yields:
```
camlRange.f_0_1_code:
  movq %rax, %rdx
  movq %rsi, %rax
  cmpq %rbx, %rdx
  jae .L103
  movq %rdi, %rax
  ret
.align 4
.L103:
  ret
```
I haven't yet written the Cmm pattern for when the arguments are in the other order.  @Gbury @xclerc if you're happy with this, I will do that.


lthls:
It's fine to do this in Cmm (it kind of fits with the other peephole optimisations), but isn't this something we would normally do in a Selection pass ?

mshinwell:
I think if it were architecture-dependent, then probably yes, but this one seems like it should apply to all architectures?

lthls:
Selection is responsible for transforming conditional expressions into tests, so I think it would be a good fit.

mshinwell:
Not sure, I'll have to look at the code some more.  One point is that the output will need selection again, e.g. to introduce `cmov` in the future.

mshinwell:
Except of course this transformation can only apply when the bound is positive, so this isn't right yet.

mshinwell:
This should now be correct.  It also exposes the `isout` primitive which can be used as a kind of user assertion that a value is positive, in order to generate an efficient range check.

Gbury:
So the new optimisation now only triggers when the bound is a compile-time constant, is that correct ?

mshinwell:
It needs to be a non-negative compile time constant, but yes.

mshinwell:
I might try to turn that into a proper test case.

backend/cmm_helpers.ml
```diff
@@ -3345,7 +3345,120 @@ let trap_return arg trap_actions =
...
+          ifso,
+          dbg_ifso,
+          ifnot,
+          dbg_ifnot,
+          kind )
+    in
+    match bound with
+    | Cconst_int (n, _) when n >= 0 -> optimize ()
+    | Cconst_natint (n, _) when n >= 0n -> optimize ()
+    | _ -> default ())
```
Gbury, line 3404: This deserves a comment explaining what is being optimized, and why it's correct.

backend/cmm_helpers.ml
```diff
@@ -3345,7 +3345,120 @@ let trap_return arg trap_actions =
 
 let create_ccatch ~rec_flag ~handlers ~body =
   let rec_flag = if rec_flag then Cmm.Recursive else Cmm.Nonrecursive in
-  Cmm.Ccatch (rec_flag, handlers, body, Any)
+  let[@inline] default () = Cmm.Ccatch (rec_flag, handlers, body, Any) in
+  match body, handlers with
+  | ( Cifthenelse
+        ( Cop (Ccmpi Cge, [Cvar x; Cconst_int (1, _)], _),
```
Gbury, line 3351: Since the comparison here is with `1` (i.e. tagged `0`), the optimization is only correct if we are sure that `x` can only hold a tagged integer, else an innocent comparison of unboxed integers between `1` and `n` (with `n` a positive compile-time constant), might be optimised and the semantics changed.

How do you ensure that `x` is necessarily a tagged integer here ?

rc.ml
```diff
@@ -0,0 +1,34 @@
...
+  let a = 10000 in
+  if x < a && x >= 0 then a else b
+
+let opt_strict_l x b =
+  let a = 10000 in
+  if 0 <= x && a > x then a else b
+
+let opt_strict_l_rev x b =
+  let a = 10000 in
+  if a > x && 0 <= x then a else b
```
Gbury, line 34: Considering my comment above concerning correctness, it'd be nice to have this as a proper test, including comparison of Int64s (and others) between `1` and `1000`, with runtime assertions, to at least check that the semantics are preserved.

Gbury (changes requested):
I'm not sure that the code is actually correct (see my comment).

backend/cmm_helpers.ml
Gbury: One solution proposed by @lthls is to instead do the folliwng translation:
`1 <= x {op} n` --> `x - 1 {unsigned_op} n - 1`

xclerc: (I have not spent enough time on it, but
it looks like there is a condition on `n`;
`min_int` would be problematic for instance.
But maybe I am confused between tagged
and untagged values.)

Gbury: @xclerc I'm not sure I follow: in this case, `n` is a compile-time constant literal, that must be positive for the optimisation to trigger, so it can never be `min_int` (and `n - 1` is also a literal that can be computed at compile-time)

xclerc: OK, I missed the condition about
`n` being positive.

Gbury:
@mshinwell ping (there is a soundness issue to resolve if you want this to be merged, see the discussion above)

PR #2311, opened 2024-02-26
Fix SIMD register destruction 

TheNumbat:
The various `Proc` functions need to know about the SIMD operations that return a value in a specific register.
This may have been benign because these operations always use the specific register as the result, but we should fix it regardless.

This was a default-match issue; we added the `_to_xmm0` and `_to_rcx` SIMD register behaviors without being forced to update `Proc`.

The `register_behavior` functions were moved to a separate file to break a circular dependency with `Proc`.


backend/arm64/simd_proc.ml
```diff
@@ -0,0 +1,17 @@
...
+(*                                                                        *)
+(*   Copyright 2024 Jane Street Group LLC                                 *)
+(*                                                                        *)
+(*   All rights reserved.  This file is distributed under the terms of    *)
+(*   the GNU Lesser General Public License version 2.1, with the          *)
+(*   special exception on linking described in the file LICENSE.          *)
+(*                                                                        *)
+(**************************************************************************)
+
+[@@@ocaml.warning "+a-4-30-40-41-42"]
```
xclerc, line 15:
```suggestion
[@@@ocaml.warning "+a-30-40-41-42"]
```
(Arguably useless, for consistency.)

backend/amd64/simd_proc.ml
```diff
@@ -0,0 +1,121 @@
...
+(*                                                                        *)
+(*   Copyright 2024 Jane Street Group LLC                                 *)
+(*                                                                        *)
+(*   All rights reserved.  This file is distributed under the terms of    *)
+(*   the GNU Lesser General Public License version 2.1, with the          *)
+(*   special exception on linking described in the file LICENSE.          *)
+(*                                                                        *)
+(**************************************************************************)
+
+[@@@ocaml.warning "+a-4-30-40-41-42"]
```
xclerc, line 15:
```suggestion
[@@@ocaml.warning "+a-30-40-41-42"]
```
Given the nature of the bug, I would rather
not disable this warning.

backend/amd64/proc.ml
```diff
@@ -465,14 +480,28 @@ let destroyed_at_basic (basic : Cfg_intf.S.basic) =
...
-    destroy_xmm15 ()
+    destroy_xmm 15
   | Op(Intop(Imulh _ | Icomp _) | Intop_imm((Icomp _), _)) ->
     [| rax |]
   | Op (Specific (Irdtsc | Irdpmc)) ->
     [| rax; rdx |]
   | Op Poll -> destroyed_at_alloc_or_poll
   | Op (Alloc _) ->
     destroyed_at_alloc_or_poll
+  | Op (Specific (Isimd op)) ->
```
xclerc, line 491: Should probably be factored out
with lines 424-437.

backend/amd64/simd_proc.ml
mshinwell: (also I think this is an old warning stanza from an old version of flambda2, we can probably remove it now since the build system now sets reasonable warnings)

backend/arm64/simd_proc.ml
TheNumbat: I've added back all the unnecessarily disabled warnings

backend/amd64/simd_proc.ml
TheNumbat: I've added back all the unnecessarily disabled warnings

xclerc (approved):

PR #2313, opened 2024-02-27
Fix assertion in Emit

gretay-js:
Assertions in `Emit` related to `simd` are too strong. It is sufficient to check for `Reg.same_loc` . Note that register class does not matter at this point after register allcoation and also `Reg.typ` is only used in `Emit.move`.

Checking registers for equality using `=` in `Emit` fails when building with FDO because various fields of `Reg` related to register allocation (`spill_cost` for example) are not initialized when starting compilation from `Emit`.


TheNumbat (commented):
Also note that the comparisons with `phys_xmm0v` no longer assert that the reg type is `Vec128`. (But that isn't necessary for the emitted code to be correct.)

PR #2314, opened 2024-02-27
Remove file-level attributes disabling warning 4 in Flambda 2

mshinwell:
This removes the file-level attributes in the main part of the Flambda 2 codebase that disable the fragile-match warning.  In the majority of cases I have made patterns exhaustive rather than just pushing the attributes down onto particular matches.  This has cleaned the provers up somewhat; even if they are all a bit verbose, they are at least now consistent.

I made a tiny semantic change in `Closure_conversion` along the way to propagate information from a `Value_int` approximation.

Based on #2312

### DIFF
--- a/middle_end/flambda2/from_lambda/closure_conversion.ml
+++ b/middle_end/flambda2/from_lambda/closure_conversion.ml
@@ -14,8 +14,6 @@
 (*                                                                        *)
 (**************************************************************************)
 
-[@@@ocaml.warning "-fragile-match"]
-
 open! Flambda
 module BP = Bound_parameter
 module IR = Closure_conversion_aux.IR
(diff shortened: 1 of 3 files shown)


middle_end/flambda2/from_lambda/closure_conversion.ml
```diff
@@ -3086,8 +3105,10 @@ let wrap_final_module_block acc env ~program ~prog_return_cont
     let module_block_simple =
       let simple_var = Simple.var module_block_var in
       match find_value_approximation env simple_var with
-      | Value_approximation.Value_symbol s -> Simple.symbol s
-      | _ -> simple_var
+      | Value_symbol s -> Simple.symbol s
+      | Value_int i -> Simple.const_int i
```
lthls, line 3109: I'm not sure about this addition. We're creating a simple for the main module block, if it is an integer we have a serious problem.

lthls (commented):
I think the PR is correct, but I'm not very enthusiastic. In my opinion it makes the code harder to read because it's less obvious that the expanded case is a default case. Plus a few instances are really hard to read (a match in a when clause, for instance.)

mshinwell:
We will leave out the changes to `provers.ml` but @lthls will write some "extraction" functions which can be used to rewrite the provers in another way, so the warning can then be enabled.

PR #2315, opened 2024-02-27
Add `caml_*_of_string_unboxed` runtime primitives for boxed ints

alanechang:
Add  `caml_*_of_string_unboxed` runtime primitives for nativeint, int32, and int64. These new primitives return unboxed values and don't cause any allocations in the success case.

### DIFF
--- a/ocaml/runtime/ints.c
+++ b/ocaml/runtime/ints.c
@@ -334,9 +334,14 @@ CAMLprim value caml_int32_format(value fmt, value arg)
   return caml_alloc_sprintf(format_string, Int32_val(arg));
 }
 
+int32_t caml_int32_of_string_unboxed(value s)
+{
+  return (int32_t) parse_intnat(s, 32, INT32_ERRMSG);
+}
+
 CAMLprim value caml_int32_of_string(value s)
 {
-  return caml_copy_int32((int32_t) parse_intnat(s, 32, INT32_ERRMSG));
+  return caml_copy_int32(caml_int32_of_string_unboxed(s));
 }
 
 int32_t caml_int32_bits_of_float_unboxed(double d)
@@ -582,7 +587,7 @@ CAMLprim value caml_int64_format(value fmt, value arg)
   return caml_alloc_sprintf(format_string, Int64_val(arg));
 }
 
-CAMLprim value caml_int64_of_string(value s)
+int64_t caml_int64_of_string_unboxed(value s)
 {
   const char * p;
   uint64_t res, threshold;
@@ -616,7 +621,12 @@ CAMLprim value caml_int64_of_string(value s)
     }
   }
   if (sign < 0) res = - res;
-  return caml_copy_int64(res);
+  return res;
+}
+
+CAMLprim value caml_int64_of_string(value s)
+{
+  return caml_copy_int64(caml_int64_of_string_unboxed(s));
 }
 
 int64_t caml_int64_bits_of_float_unboxed(double d)
@@ -845,7 +855,12 @@ CAMLprim value caml_nativeint_format(value fmt, value arg)
   return caml_alloc_sprintf(format_string, Nativeint_val(arg));
 }
 
+intnat caml_nativeint_of_string_unboxed(value s)
+{
+  return parse_intnat(s, 8 * sizeof(value), INTNAT_ERRMSG);
+}
+
 CAMLprim value caml_nativeint_of_string(value s)
 {
-  return caml_copy_nativeint(parse_intnat(s, 8 * sizeof(value), INTNAT_ERRMSG));
+  return caml_copy_nativeint(caml_nativeint_of_string_unboxed(s));
 }
--- a/ocaml/runtime/ints.c
+++ b/ocaml/runtime/ints.c
@@ -334,7 +334,7 @@ CAMLprim value caml_int32_format(value fmt, value arg)
   return caml_alloc_sprintf(format_string, Int32_val(arg));
 }
 
-int32_t caml_int32_of_string_unboxed(value s)
+CAMLprim int32_t caml_int32_of_string_unboxed(value s)
 {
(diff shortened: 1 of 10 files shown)


mshinwell:
I'll review this.
It seems like we should try to upstream this change to minimize stdlib divergence.

ocaml/runtime/ints.c
```diff
@@ -334,9 +334,14 @@ CAMLprim value caml_int32_format(value fmt, value arg)
   return caml_alloc_sprintf(format_string, Int32_val(arg));
 }
 
+int32_t caml_int32_of_string_unboxed(value s)
```
mshinwell, line 337: This should also have `CAMLprim` or the bootstrap for the upstream build won't see the primitive (I think).

ocaml/runtime/ints.c
```diff
@@ -582,7 +587,7 @@ CAMLprim value caml_int64_format(value fmt, value arg)
   return caml_alloc_sprintf(format_string, Int64_val(arg));
 }
 
-CAMLprim value caml_int64_of_string(value s)
+int64_t caml_int64_of_string_unboxed(value s)
```
mshinwell, line 590: Same comment re. `CAMLprim`

ocaml/runtime/ints.c
```diff
@@ -845,7 +855,12 @@ CAMLprim value caml_nativeint_format(value fmt, value arg)
   return caml_alloc_sprintf(format_string, Nativeint_val(arg));
 }
 
+intnat caml_nativeint_of_string_unboxed(value s)
```
mshinwell, line 858: Same again

ocaml/runtime4/ints.c
```diff
@@ -334,9 +334,14 @@ CAMLprim value caml_int32_format(value fmt, value arg)
   return caml_alloc_sprintf(format_string, Int32_val(arg));
 }
 
+int32_t caml_int32_of_string_unboxed(value s)
```
mshinwell, line 337: Same comments as per runtime5

alanechang:
Sounds good. I will try opening a similar PR in the upstream repo.

mshinwell (changes requested):

ocaml/runtime/ints.c
alanechang: Added `CAMLprim`

alanechang:
Some marshalling tests are failing in oclassic and upstream builds:

https://github.com/ocaml-flambda/flambda-backend/blob/bfb237d84d391ddb221d295b00d5d8a048dd2f3c/ocaml/testsuite/tests/lib-marshal/intext.ml#L154-L158
> ^ Test 39 here

mshinwell:
I might have to look in more detail tomorrow, but this has presumably affected the sharing of values which the test was relying on.  It's probably because the boxing has moved to the OCaml side, I guess.

alanechang:
Yeah, this is probably resulting in two allocations now rather than one:

https://github.com/ocaml-flambda/flambda-backend/blob/bfb237d84d391ddb221d295b00d5d8a048dd2f3c/ocaml/testsuite/tests/lib-marshal/intext.ml#L73-L74

Playing with a toy example seems to support that:

```ocaml
let open_out_bin name =
  open_out_gen [Open_wronly; Open_creat; Open_trunc; Open_binary] 0o666 name

let f filename flags =
  let oc = open_out_bin filename in
  let i = Int64.of_string "123456789123456" in
  Marshal.to_channel oc (i,i) flags;
  ()
```
before:
```
(function{./debug_test.ml:36,6-147} camlDebug_test__f_1_1
     (filename/394: val flags/395: val)
 (let
   (oc/397
      (app{./debug_test.ml:37,11-32;./debug_test.ml:34,2-76}
        G:"camlStdlib__open_out_gen_19_111_code" G:"camlDebug_test__s4" 877
        filename/394 val)
    i/398
      (extcall "caml_int64_of_string"{./debug_test.ml:38,10-43}
        L:"camlDebug_test__s6" int->val)
    apply_result/403
      (extcall "caml_output_value"{./debug_test.ml:39,2-35;marshal.ml:27,0-93}
        oc/397 (alloc{./debug_test.ml:39,24-29} 2048 i/398 i/398) flags/395
        int,int,int->val))
   1))
```
after:
```
(function{./debug_test.ml:36,6-147} camlDebug_test__f_1_1
     (filename/394: val flags/395: val)
 (let
   (oc/397
      (app{./debug_test.ml:37,11-32;./debug_test.ml:34,2-76}
        G:"camlStdlib__open_out_gen_19_111_code" G:"camlDebug_test__s4" 877
        filename/394 val)
    i/398
      (extcall "caml_int64_of_string_unboxed"{./debug_test.ml:38,10-43}
        L:"camlDebug_test__s6" int->int)
    apply_result/404
      (extcall "caml_output_value"{./debug_test.ml:39,2-35;marshal.ml:27,0-93}
        oc/397
        (alloc{./debug_test.ml:39,24-29} 2048
          (alloc{./debug_test.ml:38,10-43} 2303 G:"caml_int64_ops" i/398)
          (alloc{./debug_test.ml:38,10-43} 2303 G:"caml_int64_ops" i/398))
        flags/395 int,int,int->val))
   1))
```
> both are flambda2 with -Oclassic

mshinwell:
@Gbury I think this might be because classic mode unboxing is duplicating a boxing down two branches - what do you think?

lthls:
That's the intended behaviour. Closure worked the same way, and unboxing in classic mode was designed to mimic it.

Gbury:
As @lthls says, this is the intended behavior in classic mode, made to mimic Closure.

lthls:
You can fix the test by using `let i = Sys.opaque_identity (Int64.of_string "123456789123456")`. This should work both for classic mode and the upstream build.

mshinwell (approved):

PR #2316, opened 2024-02-27
Document assumptions about FDO in the backend

gretay-js:
Split-compilation that we use for FDO makes some assumptions that are not easy to see (for example, #2313). This PR proposes a comment to at least document some of them.

### DIFF
--- a/backend/asmgen.mli
+++ b/backend/asmgen.mli
@@ -39,6 +39,27 @@ val compile_implementation
   -> Lambda.program
   -> unit
 
+(** [compile_implementation_linear] reads Linear IR from [progname] file
+    produced by previous compilation stages (for example,
+    using "ocamlopt -save-ir-after" and "ocamlfdo opt")
+    and continues compilatin from [Emit].
+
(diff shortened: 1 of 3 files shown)


backend/asmgen.mli
```diff
@@ -39,6 +39,27 @@ val compile_implementation
   -> Lambda.program
   -> unit
 
+(** [compile_implementation_linear] reads Linear IR from [progname] file
+    produced by previous compilation stages (for example,
+    using "ocamlopt -save-ir-after" and "ocamlfdo opt")
+    and continues compilatin from [Emit].
+
+    Correctness: carefully consider any use of [Config], [Clflags],
```
mshinwell, line 47: I think there should be another comment at the top of both emitters pointing at this one.

xclerc:
I am not positive this should be part of this comment
(because it is arguably more about the implementation
of `ocamlfdo`) but when "resuming" compilation, it
seems at least risky to use `Reg.{create,clone}` and
possibly others.

mshinwell (approved):

PR #2317, opened 2024-02-28
Fix for naked pointer problem in unsafe_get_global_value

mshinwell:
There is a use in the native toplevel of `Dynlink.unsafe_get_global_value` on code pointers.  This is not allowed because it causes a naked pointer to be visible to the GC at one (and only one) allocation site, namely the allocation of `Some` in that function.

All that the native toplevel needs is a way of testing the existence of code pointer symbols, which is provided by this patch using another function.  In bytecode, the new function probably isn't useable for these symbols, but it is never called with them.  New comments have been added accordingly.

Joint work with @xclerc, who will review.

### DIFF
--- a/native_toplevel/opttoploop.ml
+++ b/native_toplevel/opttoploop.ml
@@ -67,7 +67,7 @@ let global_symbol comp_unit =
   | Some obj -> obj
 
 let need_symbol sym =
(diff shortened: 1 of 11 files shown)


xclerc (approved):

lthls:
Should we propose this patch upstream too ?

mshinwell:
Yes I will do that.

PR #2318, opened 2024-02-28
Improve mode syntax locations

riaqn:
This is a follow-up on https://github.com/ocaml-flambda/flambda-backend/pull/2231, and improves the location info of mode syntax encoding, mostly to make our internal tools happy.


ocaml/parsing/parser.mly
```diff
@@ -153,7 +153,7 @@ let mk_attr ~loc name payload =
 (* For modes-related attributes, no need to call [register_attr] because they
 result from native syntax which is only parsed at proper places that are
 guaranteed to be used. *)
-let mkexp_with_modes ?(ghost=false) loc modes exp =
+let mkexp_with_modes ?(ghost=false) ~loc modes exp =
```
ncik-roberts, line 156: I'm confused why `mkpat_with_modes` and `mktyp_with_modes` always call `Mode.ghostify`, but `mkexp_with_modes` makes its caller make the choice whether to call `Mode.ghostify`.
  
  * Can we always call `Mode.ghostify` in `mkexp_with_modes` as well?
  * Even simpler, can we just always make modes as ghost?

ocaml/parsing/parser.mly
```diff
@@ -170,13 +170,21 @@ let mkexp_with_modes ?(ghost=false) loc modes exp =
       Exp.apply ~loc ext [Nolabel, exp]
 
 let mkpat_with_modes modes pat =
+  (* Modes doesn't belong to the pattern; in particular, it should not be
+     mentioned in a pattern error message. Therfore, it's not covered by the
+     pattern's location. We mark it ghost for location well-nestedness. *)
```
ncik-roberts, line 175: I would remove the first two sentences of this comment and the below comment. They don't strike me as useful. The third sentence is most useful. (I meant mention that it's ppxlib's well-nestedness checks.)

ocaml/parsing/parser.mly
riaqn: In `mkexp_with_modes ?ghost ~loc modes exp`, the `ghost` doesn't control the ghost-ness of `modes`, but the whole application (applying `[%jane.erasable.mode modes]` to `exp`). The caller knows better if that should be ghost:
- If you write `local_ exp`, then I think the whole application should be non-ghost, and all modes should be non-ghost as well.
- If you write `let local_ a = exp`, it's parsed as:
```
let [@jane.erasable.mode "local"] a = [%jane.erasable.mode "local"] exp
```
The `pvb_attribute` will be non-ghost as well as all the modes in it. On the RHS, the whole application should be ghost, because it doesn't exist; All the modes should be ghost to avoid clash with the modes on the LHS.

ncik-roberts: OK, I'm convinced enough.

ncik-roberts (approved):
Part of the complication here is that the modes in `local_ x -> local_ y` are associated with the LHS/RHS types as opposed to with the arrow. I think this will become a lot simpler in a future world where we can just modify the parsetree for arrow (if that day ever comes).

ocaml/parsing/parser.mly
riaqn: thanks - resolving.

PR #2319, opened 2024-02-28
Use "unboxable" as the name of the best-effort unboxing attribute

mshinwell:
We have eventually decided to call this `[@unboxable]`, after much deliberation, since it is best-effort.


Gbury:
You'll also need to update the doc at `middle_end/flambda2/docs/unboxed_params.md`

mshinwell:
Ah thanks, have done that (and added a couple more sentences).

Gbury (approved):

PR #2321, opened 2024-02-28
Add tests for function parameter/return unboxing

Gbury:
This PR adds some tests following #1271 .

These tests actually show that #2319 missed one spot when it renamed the attribute.

Also, it appears that the location of the warning is not always exact: indeed, as can be show in the compilers reference for the added tests, we have two warnings as expected, but both quote the same location, which is less than ideal since this means we cannot distinguish the warning for the annotation on the parameter from the warning on the annotation of the function return, cc @Ekdohibs .

### DIFF
--- a/ocaml/testsuite/tests/unboxed-params/test.ml
+++ b/ocaml/testsuite/tests/unboxed-params/test.ml
@@ -1,0 +1,89 @@
+(* TEST_BELOW *)
+
+(* About testing for allocations.
+
+   The call to `Gc.counters` itself allocates a bit (about 10 words: a
+   3-tuple plus some floats), so we will never have zero allocation. It is a
+   bit fragile to check the exact number of words allocated, so instead we
+   check that the number of allocated minor words is "not a lot more" than
+   what `Gc.counters` allocates by using a reasonable threshold.
+
+   To clearly differentiate the allocations from `Gc.counters` from those
+   that are (potentially) in a function `f`, we run `f` in a loop enough times
+   so that if it had any allocation, the number of minor words allocated would
+   be a lot bigger than the threshold we check for. *)
+let test_allocs test_name f arg1 arg2 =
+  let (minor, promoted, major) = Gc.counters () in
+  for i = 1 to 1_000 do
+    if not (f arg1 arg2) then failwith "incorrect result"
+  done;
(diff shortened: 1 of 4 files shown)


middle_end/flambda2/from_lambda/closure_conversion.ml
```diff
@@ -2366,7 +2384,15 @@ let close_functions acc external_env ~current_region function_declarations =
...
             }
         in
-        Function_slot.Map.add function_slot approx approx_map)
+        let approx_map =
+          Function_slot.Map.add function_slot approx approx_map
+        in
+        match Function_decl.calling_convention decl with
+        | Normal_calling_convention -> approx_map
+        | Unboxed_calling_convention (_, _, unboxed_function_slot) ->
+          (* CR mshinwell: add a proper approximation here *)
```
lthls, line 2393: If we don't have a proper approximation, we won't be able to inline the stubs, so unboxing is never going to be beneficial (in classic mode).

Gbury: Unboxing is now disabled in classic mode. If/when we decide to enable it, we'll need to revisit this CR.

mshinwell:
This isn't ready yet, @mshinwell to try reproducing the failure on `env.ml` in classic mode.

Gbury:
@mshinwell ping ?

ocaml/testsuite/tests/unboxed-params/test.ml
```diff
@@ -0,0 +1,92 @@
...
+   fragile to check the exact number of words allocated, so instead we check
+   that the number of allocated minor words is "not a lot more" than what
+   `Gc.counters` allocates by using a reasonable threshold.
+
+   To clearly differentiate the allocations from `Gc.counters` from those that
+   are (potentially) in a function `f`, we run `f` in a loop enough times so
+   that if it had any allocation, the number of minor words allocated would be
+   a lot bigger than the threshold we check for. *)
+let test_allocs test_name f arg1 arg2 =
+  let (minor, promoted, major) = Gc.counters () in
```
lthls, line 19: `Gc.minor_words` looks more interesting (and doesn't allocate in native mode)

PR #2323, opened 2024-02-29
Ignore bootstrapping artifacts

goldfirere:
Bootstrapping produces three files. I would like them ignored. But it would be great if someone knew why these are created and whether ignoring is the right behavior.


mshinwell (approved):
This looks correct, I'm unsure why these files weren't ignored previously.

PR #2325, opened 2024-02-29
Improve interaction between `Jane_syntax` and modes

riaqn:
This fixes an issue about the interaction between `Jane_syntax` and modes. In particular, when you write:
```
type r = local_ (x : string * y : string) -> string
```
the labelled tuple will have two attributes: one about modes, one about labeled tuples.

`T.iter` will recognise the labeled tuple attribute first, and iterate all the remaining attributes straightaway. In particular, that means the `attribute` iterator in `ast_invariants.ml` will register the attribute, which we don't want for the mode attribute.

Indeed, the mode attribute is "outer"/"looser" than the labeled tuple attributes, and should be recognized first.

### DIFF
--- a/ocaml/parsing/ast_iterator.ml
+++ b/ocaml/parsing/ast_iterator.ml
@@ -158,16 +158,16 @@ module T = struct
   let iter sub ({ptyp_desc = desc; ptyp_loc = loc; ptyp_attributes = attrs}
                   as typ) =
     sub.location sub loc;
-    match Jane_syntax.Core_type.of_ast typ with
-    | Some (jtyp, attrs) ->
-        sub.attributes sub attrs;
-        sub.typ_jane_syntax sub jtyp
-    | None ->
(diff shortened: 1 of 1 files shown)


ncik-roberts (approved):
*sigh* Jane Syntax attribute ordering strikes again.

For types and patterns, this is clearly fine. `local_` doesn't really belong to the type/pattern and instead belongs to some outer context. So `local_` should be interpreted first.

For expressions I'm less convinced. `local_ expr` is just a valid expression We'll have to be careful if we ever introduce another construct that is marked by a single attribute, as we'll need to get the ordering right. But that seems like a general problem that we'll have to consider when adding another such expression, and the problem is not limited to AST iteration, so this PR seems fine.

riaqn:
Sorry about the confusion - `local_ exp` is encoded by an extension node applied to `exp`, so it should be unambiguous. I don't know why this PR changed the ordering for expressions as well - but also it won't break it.

But indeed I was thinking about parsing `local_ exp` as some attribute on `exp` sometime in the future - it now sounds like a bad idea.

PR #2326, opened 2024-02-29
Tolerate `global` in `transl_mode_annots`

riaqn:
Our internal tools (PPXes) might leak modalities on both record fields and constructor arguments to the LHS/RHS of arrow types, which is incorrect. This PR is a dirty fix that tolerate that behaviour. 

The commit "ignore global in alloc_mode" should be reverted ASAP.


ncik-roberts (approved):
We discussed this and I agree this is:
  * equivalent to what we did in minus-10
  * useful for getting minus-11 out
Let's wait for your internal build to pass first.

goldfirere:
Is there an internal ticket to track the better fix?

PR #2327, opened 2024-02-29
Make `Sys.opaque_identity` work on different layouts

alanechang:
Use the new `[@layout_poly]` attribute on `Sys.opaque_identity`.

Wonder if `Obj.magic` should get the same treatment.

### DIFF
--- a/ocaml/stdlib/sys.mli
+++ b/ocaml/stdlib/sys.mli
@@ -415,7 +415,7 @@ val runtime_warnings_enabled: unit -> bool
 
 (** {1 Optimization} *)
 
-external opaque_identity : 'a -> 'a = "%opaque"
+external[@layout_poly] opaque_identity : ('a : any). 'a -> 'a = "%opaque"
 (** For the purposes of optimization, [opaque_identity] behaves like an
     unknown (and thus possibly side-effecting) function.
 
--- a/ocaml/stdlib/sys.ml.in
+++ b/ocaml/stdlib/sys.ml.in
@@ -159,7 +159,7 @@ let ocaml_release = {
 
 (* Optimization *)
 
-external opaque_identity : 'a -> 'a = "%opaque"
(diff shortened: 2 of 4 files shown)


Ekdohibs:
`Obj.magic` should be properly handled on different layouts inside the compiler, but relies on the fact that the input layout is the same as the output layout. This should be guaranteed by `[@layout_poly]` if I'm not mistaken, but does not take into account wrong declarations like `external wrong : (int * int) -> int32# = "%obj_magic"`, for instance.

alanechang:
Yes, `[@layout_poly]` does guarantee all the layout polymorphic type variables get the same layout such that:

```
external[@layout_poly] magic : ('a : any) ('b : any). 'a -> 'b = "%obj_magic"

let f1 : int32# -> int32# = magic;; (* ok *)
let f2 : float# -> float# = magic;; (* ok *)
let f3 : float# -> int32# = magic;; (* error *)
```

In the same pr, we also added some basic layout checks that will reject bad declarations:

```
# external wrong : (int * int) -> int32# = "%obj_magic";;
Error: The primitive [%obj_magic] is used in an invalid declaration.
       The declaration contains argument/return types with the wrong layout.
```

ocaml/stdlib/sys.mli
```diff
@@ -415,7 +415,7 @@ val runtime_warnings_enabled: unit -> bool
 
 (** {1 Optimization} *)
 
-external opaque_identity : 'a -> 'a = "%opaque"
+external[@layout_poly] opaque_identity : ('a : any). 'a -> 'a = "%opaque"
```
goldfirere, line 418: Is it worth adding a test for @Ekdohibs's question? That is, it might be good to test that a use of `opaque_identity` that crosses layouts is rejected.

goldfirere (approved):

PR #2328, opened 2024-03-01
Variant unboxing for parameters and returns

lthls:
This PR adds support for variant unboxing on top of #1271.

I don't think we should merge it, as currently the code doesn't simplify away as well as we want to (even with Simplify). We will likely need some match-in-match support, and maybe some better support for poison-like values.
Also, we should probably restrict it to a subset of variant types for which it is reasonable.

But it should still work at preventing allocations:
```ocaml
let[@inline never] f (x[@unboxable] : _ result) =
  match x with Error _ -> 0 | Ok n -> n

let g x err =
  f (Ok x) + f (Error err)
```
Here, `g` will not allocate anything, but the unboxed version of `f` will do three switches instead of one. One is there because this PR always adds a switch on `is_int`, which for `result` should always return `false`, and the two other ones are equivalent but cannot be simplified yet.

### DIFF
--- a/middle_end/flambda2/from_lambda/closure_conversion.ml
+++ b/middle_end/flambda2/from_lambda/closure_conversion.ml
@@ -1429,67 +1429,81 @@ let close_switch acc env ~condition_dbg scrutinee (sw : IR.switch) :
         untag ~body
 
 type unboxing_kind_with_variables =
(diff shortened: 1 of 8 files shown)


mshinwell:
Let's restrict this to reasonable types, and put it behind a flag for the moment, since we need match-in-match to avoid code being worse in the case where the unboxed function gets inlined.

PR #2329, opened 2024-03-01
`pprintast` recognize modalities on constructor arguments

riaqn:
With regression tests added.


ncik-roberts (approved):
Looks good to me.

PR #2330, opened 2024-03-01
`Ast_mapper` and `Ast_iterator` should traverse location inside mode expressions

alanechang:
This PR tries to fix two issues:

1. Before this PR, `Ast_mapper` maps over mode expressions as normal attributes all except for ones in `N_ary.function_constraint` which just get ignored (see broken test that triggers this). This PR aims to make this more consistent and adds a `modes` mapper function that applies to all mode expressions. The default mapper will also be responsible for descending into location fields.

2. `Ast_iterator` also ignores modes inside `N_ary.function_constraint` and the default iterator function doesn't descend into location fields. This PR changes this and bring it in line with the new mapper.

Request review from @riaqn 

Extra context: I'm working on modal kind syntax where we want to embed mode expressions in kind annotations. The implementation is similar to how `N_ary_functions` carries mode expressions. That's how I originally ran into the issue with ast_mapper.

### DIFF
--- a/ocaml/parsing/ast_iterator.ml
+++ b/ocaml/parsing/ast_iterator.ml
@@ -518,7 +518,8 @@ module E = struct
   let iter_function_constraint sub : N_ary.function_constraint -> _ =
     (* Enable warning 9 to ensure that the record pattern doesn't miss any
        field. *)
-    fun[@ocaml.warning "+9"] { mode_annotations = _; type_constraint } ->
+    fun[@ocaml.warning "+9"] { mode_annotations; type_constraint } ->
+      sub.modes sub mode_annotations;
       match type_constraint with
       | Pconstraint ty ->
           sub.typ sub ty
@@ -773,6 +774,8 @@ module CE = struct
     sub.attributes sub pci_attributes
 end
 
+module ME = Jane_syntax.Mode_expr
+
 (* Now, a generic AST mapper, to be extended to cover all kinds and
    cases of the OCaml grammar.  The default behavior of the mapper is
    the identity. *)
@@ -967,8 +970,12 @@ let default_iterator =
       this.location this a.attr_loc
     );
     attributes = (fun this l -> List.iter (this.attribute this) l);
-    (* CR zqian: should go into the modes and at least iterate the locations. *)
-    modes = (fun _this _m -> ());
+    (* Location inside a mode expression needs to be traversed. *)
+    modes = (fun this m ->
+      let iter_const sub : ME.Const.t -> _ =
+        fun m -> iter_loc sub (m : ME.Const.t :> _ Location.loc)
+      in
+      iter_loc_txt this (fun sub -> List.iter (iter_const sub)) m);
     payload =
       (fun this -> function
          | PStr x -> this.structure this x
--- a/ocaml/parsing/ast_iterator.ml
+++ b/ocaml/parsing/ast_iterator.ml
@@ -774,8 +774,6 @@ module CE = struct
     sub.attributes sub pci_attributes
 end
 
-module ME = Jane_syntax.Mode_expr
-
 (* Now, a generic AST mapper, to be extended to cover all kinds and
    cases of the OCaml grammar.  The default behavior of the mapper is
    the identity. *)
@@ -972,8 +970,9 @@ let default_iterator =
     attributes = (fun this l -> List.iter (this.attribute this) l);
     (* Location inside a mode expression needs to be traversed. *)
     modes = (fun this m ->
-      let iter_const sub : ME.Const.t -> _ =
-        fun m -> iter_loc sub (m : ME.Const.t :> _ Location.loc)
+      let open Jane_syntax.Mode_expr in
+      let iter_const sub : Const.t -> _ =
+        fun m -> iter_loc sub (m : Const.t :> _ Location.loc)
       in
       iter_loc_txt this (fun sub -> List.iter (iter_const sub)) m);
     payload =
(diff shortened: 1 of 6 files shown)


ocaml/parsing/ast_iterator.ml
```diff
@@ -969,8 +972,12 @@ let default_iterator =
...
     );
     attributes = (fun this l -> List.iter (this.attribute this) l);
-    (* [ast_iterator] should not know about the structure of mode expressions *)
-    modes = (fun _this _m -> ());
+    (* Location inside a mode expression needs to be traversed. *)
+    modes = (fun this m ->
+      let iter_const sub : ME.Const.t -> _ =
+        fun m -> iter_loc sub (m : ME.Const.t :> _ Location.loc)
+      in
+      iter_loc_txt this (fun sub -> List.iter (iter_const sub)) m);
```
riaqn, line 980: Is it possible to move this logic to the client code of `ast_iterator`? I think in the future `Mode_expr.t` will be more complex than a list (it will be a real "expression" with mode operators join, meet, etc.) and it would occur strange to me if the `ast_iterator` needs to know about that.

ocaml/parsing/ast_mapper.ml
```diff
@@ -1059,7 +1062,13 @@ let default_mapper =
...
       }
     );
-    attributes = (fun this l -> List.map (this.attribute this) l);
+    attributes = (fun this l ->
+      List.filter_map (fun attr ->
+        let m, _ = ME.maybe_of_attrs [attr] in
+        match m with
+        | Some m -> this.modes this m |> ME.attr_of
+        | None -> Some (this.attribute this attr)) l);
+
```
riaqn, line 1071: Here we are detecting modes directly under attributes. It might be more principled if we detect it at places that they can actually occur. Such contexts can be found by search `Mode_expr.maybe_of_attrs` in `ast_iterator.ml`.

In addition, mode coercion expression (`local_ expr`) encodes modes not in attributes but in extension node. I created https://github.com/ocaml-flambda/flambda-backend/pull/2331 to make it easier to handle.

lpw25:
> I'm working on modal kind syntax where we want to embed mode expressions in kind annotations. 

I don't think we actually want to embed mode expressiosn in kind annotations. The mode expressions in kind annotations are going to be restricted to the trivial case of a sequence of mode names. Those are the only mode expressions we have right now, but when we add mode polymorphism that will no longer be the case. It will save us work in the long run if we use distinct types for the modes in kind annotations now, rather than needing to uncouple to two uses later.

riaqn:
> The mode expressions in kind annotations are going to be restricted to the trivial case of a sequence of mode names. 

Yeah I agree, and now I'm confused, it seems we already have that? https://github.com/ocaml-flambda/flambda-backend/blob/79d6d7a5462f2f9f036fe8dc3c6c57cdf10c9ee6/ocaml/typing/jkind.ml#L525

ocaml/parsing/ast_mapper.ml
riaqn: This should be unblocked now that https://github.com/ocaml-flambda/flambda-backend/pull/2335 is merged.

alanechang: Great!

I still find it a bit tricky to implement `ast_mapper` correctly here.

I agree what you suggested is a better approach (the current implementation also makes mode attributes visible when someone overrides the `attributes` function which is not ideal).

If we take the suggested approach similar to `ast_iterator`:
- look for mode expressions at typ/pat/value_bindings sites
- call `sub.attributes` on non-mode attributes and `sub.modes` on mode expressions
- combine everything back together into an attribute list

Then there's a concern around ordering when we recombine the attributes. Do we maintain an invariant about whether mode attributes should always come first/last when parsing (suspect this is already the case but is it something we want to be explicit about)?

riaqn:
@alanechang I think what you had was almost finished. I rebased and it's now ready for your review.

ocaml/parsing/ast_mapper.ml
```diff
@@ -646,7 +645,7 @@ module E = struct
   let map_modes_exp sub : Modes.expression -> Modes.expression = function
     (* CR modes: One day mappers might want to see the modes *)
```
alanechang, line 646: Can we remove this CR now?

alanechang (commented):
@riaqn The changes look good. Left a few minor comments/observations.

 (can't approve my own PR, so feel free to approve and merge when you think it's ready!)

ocaml/testsuite/tests/typing-modes/modes.ml
```diff
@@ -337,8 +337,8 @@ module type S = sig
   val x : string -> string @ local @@ foo bar
 end
 [%%expect{|
-Line 337, characters 38-45:
-337 |   val x : string -> string @ local @@ foo bar
-                                            ^^^^^^^
+Line 2, characters 38-45:
+2 |   val x : string -> string @ local @@ foo bar
```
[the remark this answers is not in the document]
riaqn: Yeah exactly. ( I didn't notice the location is wrong before this PR)

ocaml/testsuite/tests/parsetree/modes_ast_mapper.reference
```diff
@@ -0,0 +1,13 @@
+local [File "_none_", line 1]
+------------------------------
+local [File "_none_", line 1]
+unique [File "_none_", line 1, characters 4-11]
+unique [File "_none_", line 1, characters 4-11]
+------------------------------
+local [File "_none_", line 1, characters 29-30]
+local [File "_none_", line 1, characters 4-10]
+local [File "_none_", line 1, characters 4-10]
```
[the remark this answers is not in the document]
riaqn: Yeah, one in `Pexp_extention`, one in `pvb_attributes`, and one in `N_ary.mode_annotations`.

ocaml/parsing/ast_mapper.ml
riaqn: Good point.

riaqn (approved):

PR #2332, opened 2024-03-04
Disable stack checks for generated functions

xclerc:
This pull request disables the stack checks for
"generated" functions (_i.e._ `caml_apply`,
`caml_currify`, `caml_send`, and `caml_tuplify`)
by introducing a "codegen" option disabling
the check for a given function.

### DIFF
--- a/backend/amd64/emit.mlp
+++ b/backend/amd64/emit.mlp
@@ -1864,7 +1864,7 @@ let fundecl fundecl =
   let handle_overflow_and_max_frame_size =
     (* CR mshinwell: this should be conditionalized on a specific
        "stack checks enabled" config option, so we can backport to 4.x *)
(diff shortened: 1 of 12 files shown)


mshinwell:
In particular, relying on the red zone for `caml_apply` prevents double (repeated) stack checks on indirect calls.

xclerc:
(note: the checks are kept for the
runtime5/debug job, because the
initial stack is tiny.)

stedolan:
I don't understand how it can be safe to rely on the red zone in caml_apply - in the non-tail-call case, caml_apply surely consumes some stack?

stedolan:
I think shrink-wrapping would help here. I mostly doubt that combining the checks would help, because you're introducing a new load (to a location not in L1 dcache!) and making code bigger.

PR #2334, opened 2024-03-04
Move the `Uniqueness` axis into the `Monadic` fragment

riaqn:
This PR introduces the object representing the monadic fragment to the mode system, and moves the uniqueness axis into the monadic fragment. 

Currently the monadic fragment contains only the uniqueness axis, which means this PR is not observable externally. However, we will very soon add more axes (`contention`) to the monadic fragment.

Request review from @goldfirere  - I think it's better to review per-commit.

### DIFF
--- a/ocaml/typing/mode.ml
+++ b/ocaml/typing/mode.ml
@@ -12,6 +12,9 @@
 (*                                                                        *)
 (**************************************************************************)
 
+(* warn on fragile matches *)
+[@@@warning "+4"]
+
 open Solver
 open Solver_intf
 open Mode_intf
@@ -304,7 +307,7 @@ module Lattices = struct
     | Areality, Areality -> Some Refl
     | Linearity, Linearity -> Some Refl
     | Uniqueness, Uniqueness -> Some Refl
-    | _ -> None
+    | (Areality | Linearity | Uniqueness), _ -> None
 
   let proj : type t r. (t, r) axis -> t -> r =
    fun ax t ->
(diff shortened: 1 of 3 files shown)


ocaml/typing/mode.ml
```diff
@@ -1220,29 +1312,67 @@ module Comonadic_with_locality = struct
...
+    S.Negative.via_monotone Obj.obj
+      (C.lift SUniqueness (Const_min Uniqueness_op))
+      (S.Negative.disallow_left m)
 
+  let set_uniqueness_min m =
+    S.Negative.via_monotone Obj.obj
+      (C.lift SUniqueness (Const_max Uniqueness_op))
+      (S.Negative.disallow_right m)
+
+  let zap_to_legacy = zap_to_ceil
```
goldfirere, line 1354: Might be nice to write the explicit tuple here; as it is, this might be easy to forget to update. The legacy mode for contention is *uncontended*.

goldfirere (approved):
Please review my comments, too.

ocaml/typing/mode.ml
```diff
@@ -1362,9 +1360,8 @@ module Monadic = struct
     match submode m0 m1 with
     | Ok () -> Ok ()
     | Error { left = uni0, (); right = uni1, () } ->
-      if Uniqueness.Const.le uni0 uni1
-      then assert false
-      else Error (`Uniqueness { left = uni0; right = uni1 })
+      assert (not (Uniqueness.Const.le uni0 uni1));
+      Error (`Uniqueness { left = uni0; right = uni1 })
```
riaqn, line 1364: The style in which it was originally written has a particular reason: as we add more axes (contention), it will be a chain of if-then-else. The original code is the special case where the chain length = 1. This minimizes the diff of the future PR introducing contention.

ocaml/typing/mode.ml
```diff
@@ -1220,29 +1312,67 @@ module Comonadic_with_locality = struct
...
 
   type equate_error = equate_step * error
 
-  let max_with_uniqueness m = S.Negative.disallow_left m
+  let uniqueness m =
+    S.Negative.via_monotone Uniqueness.Obj.obj (Proj (Obj.obj, Uniqueness)) m
 
-  let min_with_uniqueness m = S.Negative.disallow_right m
+  let max_with_uniqueness m =
+    S.Negative.via_monotone Obj.obj (Min_with Uniqueness)
```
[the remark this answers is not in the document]
riaqn: Yeah I agree this is confusing. The whole monadic fragment is inverted. This `Min_with` will invoke `Monadic_op.min` which invokes `Monadic.max` which works out fine. In fact, if you replace this with `Max_with` you get type error about left/right status (which did prevent me from writing the bug, nice).

We discussed previously that `Solver_polarized` abstracts away some of the inversion logic, but apparently not this one here. To be able to write `Max_with` here, we can simply define:
```
val opposite : ('a, 'b, 'l * 'r) morph -> ('a, 'b, 'r * 'l) morph
```
which maps `Min_with` to `Max_with` and vice versa. That might clarify things a bit?

riaqn:
Thanks for the changes. Good call about `prod_comonadic_obj` and `prod_monadic_obj`.

ocaml/typing/mode.ml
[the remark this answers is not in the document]
goldfirere: I'm not sure how you'd implement `opposite`. But in any case, I think it's OK just to leave a comment.

ocaml/typing/mode.ml
goldfirere: Yes, makes sense. Please revert (I'm off to a meeting in a moment), and then merge when CI is green. Thanks.

PR #2335, opened 2024-03-04
Full blown Jane Syntax for mode exprs

ncik-roberts:
Build on Zesen's work in #2331: Use full-blown Jane Syntax for encoding mode coercion expressions rather than another function off to the side.

Usually, full-blown Jane Syntax's costs outweigh its benefits. But, for mode coercions, it's lightweight to add it. So we get to reap the benefit: hooking into the exhaustive match.

We should probably use full-blown Jane Syntax for `stack_` when we add that as a keyword.

The main cost was the time it took to write the PR you're reading. It took about 1.5 hours, where most of that was debugging issues that my initial half-hour attempt introduced. But we have the PR now, so why not merge it.

### Tests
* `source_jane_syntax.ml` shook out bugs in `Ast_mapper` and `Pprintast`, and gives me more confidence we got that right.
* the existing test suite seems sufficient to me.

### DIFF
--- a/ocaml/parsing/jane_syntax.ml
+++ b/ocaml/parsing/jane_syntax.ml
@@ -516,15 +516,17 @@ module Modes = struct
 
   let extension_name = Mode_expr.attribute_or_extension_name
 
-  let of_expr { pexp_desc; pexp_attributes; _ } =
+  let of_expr ({ pexp_desc; pexp_attributes; _ } as expr) =
     match pexp_desc with
     | Pexp_apply
         ( { pexp_desc = Pexp_extension ({ txt; _ }, payload); pexp_loc; _ },
           [(Nolabel, body)] )
       when txt = extension_name ->
       let modes = Mode_expr.of_payload ~loc:pexp_loc payload in
-      Some (Coerce (modes, body), pexp_attributes)
-    | _ -> None
+      Coerce (modes, body), pexp_attributes
+    | _ ->
+      Misc.fatal_errorf "Improperly encoded modes expression: %a"
+        (Printast.expression 0) expr
 
   let expr_of ~loc (Coerce (modes, body)) =
     match Mode_expr.payload_of modes with
@@ -1939,10 +1941,9 @@ module Expression = struct
     | Language_extension Labeled_tuples ->
       let expr, attrs = Labeled_tuples.of_expr expr in
       Some (Jexp_tuple expr, attrs)
-    | Language_extension Mode -> (
-      match Modes.of_expr expr with
-      | Some (expr, attrs) -> Some (Jexp_modes expr, attrs)
-      | None -> None)
+    | Language_extension Mode ->
+      let expr, attrs = Modes.of_expr expr in
+      Some (Jexp_modes expr, attrs)
     | _ -> None
 
   let of_ast = Expression.make_of_ast ~of_ast_internal
--- a/ocaml/parsing/jane_syntax.ml
+++ b/ocaml/parsing/jane_syntax.ml
@@ -459,8 +459,6 @@ module Mode_expr = struct
 
   let attribute_name = attribute_or_extension_name
 
-  let extension_name = attribute_or_extension_name
-
   let payload_of { txt; _ } =
     match txt with
     | [] -> None
@@ -508,26 +506,36 @@ module Mode_expr = struct
     let loc = { loc with loc_ghost = true } in
     let txt = List.map Const.ghostify txt in
     { loc; txt }
+end
 
-  let coerce_of_expr { pexp_desc; _ } =
+(** Some mode-related constructs *)
+module Modes = struct
+  let feature : Feature.t = Language_extension Mode
+
+  type nonrec expression = Coerce of Mode_expr.t * expression
+
+  let extension_name = Mode_expr.attribute_or_extension_name
+
+  let of_expr { pexp_desc; pexp_attributes; _ } =
     match pexp_desc with
     | Pexp_apply
         ( { pexp_desc = Pexp_extension ({ txt; _ }, payload); pexp_loc; _ },
           [(Nolabel, body)] )
       when txt = extension_name ->
-      let modes = of_payload ~loc:pexp_loc payload in
-      Some (modes, body)
+      let modes = Mode_expr.of_payload ~loc:pexp_loc payload in
+      Some (Coerce (modes, body), pexp_attributes)
     | _ -> None
 
-  let expr_of_coerce ~loc modes body =
-    match payload_of modes with
+  let expr_of ~loc (Coerce (modes, body)) =
+    match Mode_expr.payload_of modes with
     | None -> body
     | Some payload ->
       let ext =
         Ast_helper.Exp.extension ~loc:modes.loc
           (Location.mknoloc extension_name, payload)
       in
-      Ast_helper.Exp.apply ~loc ext [Nolabel, body]
+      Expression.make_entire_jane_syntax ~loc feature (fun () ->
+          Ast_helper.Exp.apply ~loc ext [Nolabel, body])
 end
 
 (** List and array comprehensions *)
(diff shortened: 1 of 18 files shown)


ocaml/parsing/jane_syntax.ml
```diff
@@ -508,26 +506,36 @@ module Mode_expr = struct
     let loc = { loc with loc_ghost = true } in
     let txt = List.map Const.ghostify txt in
     { loc; txt }
+end
+
+(** Some mode-related constructs *)
+module Modes = struct
```
riaqn, line 512: Would it be better if we merge `module Mode_expr` and `module Modes`? I guess you separated them because the latter uses full-blown jane_syntax but the former doesn't - but that's also not a strong reason?

ocaml/parsing/jane_syntax.ml
```diff
@@ -1930,6 +1939,10 @@ module Expression = struct
     | Language_extension Labeled_tuples ->
       let expr, attrs = Labeled_tuples.of_expr expr in
       Some (Jexp_tuple expr, attrs)
+    | Language_extension Mode -> (
+      match Modes.of_expr expr with
+      | Some (expr, attrs) -> Some (Jexp_modes expr, attrs)
+      | None -> None)
```
riaqn, line 1945: Is it correct/better to `Misc.fatal_error` in the `None` case?

ocaml/parsing/pprintast.ml
```diff
@@ -1586,15 +1579,35 @@ and binding ctxt f {pvb_pat=p; pvb_expr=x; pvb_constraint = ct; _} =
...
+          *)
+          let mode_names (modes : Jane_syntax.Mode_expr.t) =
+            List.map Location.get_txt (modes.txt :> string loc list)
+          in
+          if
+            List.equal String.equal
+              (mode_names modes_on_binding)
+              (mode_names modes_on_expr)
+          then {x with pvb_expr = sbody}
+          else x
```
riaqn, line 1606: Is it correct/better to `Misc.fatal_error` in the case when `modes_on_bindings` is not equal to `modes_on_expr`? I guess PPXes can generate such programs, hence we cannot `Misc.fatal_error`. Moreover, such programs are still sensible to type check.

(Just thinking out loud; please resolve if you agree)

ocaml/typing/printtyped.ml
```diff
@@ -378,6 +378,11 @@ and expression_extra i ppf x attrs =
   | Texp_newtype (s, lay) ->
       line i ppf "Texp_newtype %a\n" (typevar_jkind ~print_quote:false) (s, lay);
       attributes i ppf attrs;
+  | Texp_mode_coerce modes ->
+      let modes = (modes :> string Location.loc list Location.loc) in
+      line i ppf "Texp_mode_coerce \"%s\"\n"
```
riaqn, line 383: Nit: do we need the quotes around modes? I think the types in `Texp_coerce` are not wrapped in quotes. I think this file mostly uses newlines and indentation for disambiguation.

ocaml/typing/typedtree.mli
```diff
@@ -209,6 +209,12 @@ and exp_extra =
         (** Used for method bodies. *)
   | Texp_newtype of string * Jkind.annotation option
         (** fun (type t : immediate) ->  *)
+  | Texp_mode_coerce of Jane_syntax.Mode_expr.t
+        (** local_ E *)
+
+(* CR modes: Consider fusing [Texp_mode_coerce] and [Texp_constraint] when
+   the syntax changes.
+*)
```
riaqn, line 217: I imagine mode coercion and type coercion in principle don't need to show up together, and therefore should be separated here.

(Sure, our current syntax only supports `exp : ty @@ modes` which means modes can only occur accompanied by types, but I think we shouldn't let concrete syntax affect the representation of typed tree.)

ocaml/typing/typecore.ml
```diff
@@ -8664,6 +8632,43 @@ and type_expect_jane_syntax
...
+          Misc.fatal_errorf "Unrecognized mode %s - should not parse" s
+        ) expected_mode modes.txt
+    in
+    let exp =
+      type_expect env expected_mode sbody (mk_expected ty_expected ?explanation)
+    in
+    {exp with
+     (* CR modes: We should consider not overriding [exp_loc] here -- that would
+        be more consistent to the typing of [Pexp_constraint].
+     *)
```
riaqn, line 8669: That's good point - I wonder what's preventing us? If it's just the change of error messages, I think that's fine, and we should just do it (i.e., not override `exp_loc`)?

riaqn (approved):

ocaml/parsing/jane_syntax.ml
ncik-roberts: Let's keep them separate. `Mode_expr.t` means something entirely different than `Modes.expression`. I think putting them too close together is confusing.

ocaml/parsing/pprintast.ml
ncik-roberts: I don't think we should crash in `Pprintast`. I agree with your  conclusion.

ocaml/typing/typedtree.mli
ncik-roberts: Let's make this decision (and delete this CR) once we have the new syntax.

ocaml/typing/typecore.ml
ncik-roberts: This does change a lot of error messages. I don't think we should do it in this PR, one of whose benefits is that it's clearly a no-op to tests.

ocaml/typing/printtyped.ml
ncik-roberts: I think there are plenty of instances where this file uses quotes, usually to disambiguate whether something is an identifier (quotes) vs. a Typedtree field/constructor name (no quotes). Modes are identifiers.

It's probably a bit better to wrap each mode in quotes, I've done that now.

ocaml/parsing/jane_syntax.ml
riaqn: Thanks Nick - if it ends up taking too much time, let's just leave it as it is.

ncik-roberts: The latest commit allow us to use `Misc.fatal_error` here. Resolve this once you convince yourself it is correct.

goldfirere:
> On the whole, I lean a bit more negatively and think it would have been a better choice for me to abandon this PR early. I don't think that we realize much more immediate benefit from this PR compared to Zesen's original PR — if we were writing lots of new matches over the entire parsetree, I might feel differently, but I don't think we're writing lots.

Interesting. This suggests, to me, that we should make a plan not to expand full-blown jane syntax at all, because I would think that this case is an ideal candidate.... and if jane syntax isn't good enough here, I don't think it would ever be. This is different than the previous plan I had in my head, which was to avoid restructuring jane syntax, but still use it when an application had the right shape.

ncik-roberts:
Hm, I don't want to argue that the end state (more things in full-blown Jane Syntax) is not good. I think the end state of this PR has lower technical debt interest payments than the start state. It's just that, in order to achieve it, we had to pay off all of the technical debt principal at once, which is a concentrated time cost.

riaqn:
thanks - I think it's good to go now.

(I pushed another minor commit, which is to rename `alloc_mode` to `type_mode`. Basically, this `type_mode` is not used for mode checking, but only passed to `Typetexp` for better interpretation of arrow types written by the user).

The following discussion should not block the PR:

I see your last commit moves the `type_mode` from `pexp_attributes` to `ptyp_attributes`. I think it's more proper - because as I said above, the `type_mode` is not about the pattern, but about the interpretation of the user-written type. Coincidently, on the dual side, in https://github.com/ocaml-flambda/flambda-backend/pull/2235 I was also moving the `type_mode` from `ppat_attributes` to `ptyp_attributes` - this is needed because I want to `Misc.fatal_error` for modes in `ppat_attributes`, but hit the special case of `Ppat_constraint`.

PR #2337, opened 2024-03-06
Index arrays with unboxed ints

alanechang:


ocaml/bytecomp/bytegen.ml
```diff
@@ -504,30 +504,30 @@ let comp_primitive stack_info p sz args =
   (* In bytecode, nothing is ever actually stack-allocated, so we ignore the
      array modes (allocation for [Parrayref{s,u}], modification for
      [Parrayset{s,u}]). *)
-  | Parrayrefs (Pgenarray_ref _) -> Kccall("caml_array_get", 2)
-  | Parrayrefs (Pfloatarray_ref _) -> Kccall("caml_floatarray_get", 2)
-  | Parrayrefs (Paddrarray_ref | Pintarray_ref) ->
+  | Parrayrefs (Pgenarray_ref _, _index_kind) -> Kccall("caml_array_get", 2)
```
Gbury, line 507: Shouldn't these at least assert that `index_kind` is a tagged int ? If it's not the case, I'd expect the bytecode interpreter to crash/misinterpret the instruction.

Gbury:
Small question: why is this wanted ? From what I can tell, the main/only change of this PR is to allow internaly to represent array operations whose index are not tagged integers, but I didn't see any use of that new feature.

ocaml/bytecomp/bytegen.ml
alanechang: Yeah, the PR is still very much work-in-progress. This needs to be fixed

alanechang:
The PR allows for `external get : 'a array -> int64# -> 'a = "%array_safe_get"` and ideally implements it in a way that doesn't produce extra instructions than the existing implementaiton for tagged ints

ocaml/bytecomp/bytegen.ml
alanechang: Added some new runtime functions for this to work properly.

alanechang:
@Gbury This is ready for some review. Thanks!

Gbury (changes requested):

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
```diff
@@ -745,16 +757,29 @@ let bigarray_set ~dbg ~unsafe kind layout b indexes value =
...
+      ( standard_int_of_boxed_integer bint,
+        H.Prim
+          (Unary
+             ( Num_conv
+                 { src = Tagged_immediate;
+                   dst = standard_int_or_float_of_boxed_integer bint
+                 },
+               arr_len_as_tagged_imm )) )
+  in
+  [H.Binary (Int_comp (comp_kind, Yielding_bool (Lt Unsigned)), index, arr_len)]
```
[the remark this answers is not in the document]
alanechang: There were two main reasons for converting the length:

1. Given a large negative 64-bit int such as `(-#9223372036854775807L)` which is represented by `0b1000000000000000000000000000000000000000000000000000000000000001`, we want the bounds check to fail. If we went with the approach of taggging the index before the length check, we get `0b11` as the tagged index and would return the element at index 1 without any errors.

2. The generated code is slightly better with the current implementation. If we tag the index before the length check, we end up with an additional `lea    rbx,[rbx+rbx*1+0x1]` instruction. Whereas, the current version can often produce code with the same efficiency as the tagged version:

Example I tried:

```
external get : 'a array -> int64# -> 'a = "%array_safe_get"
let do_access (arr : int array) idx =
  (get arr idx)
```


`int64#` index:
```
  41aa90:	48 8b 78 f8          	mov    rdi,QWORD PTR [rax-0x8]
  41aa94:	48 c1 ef 0a          	shr    rdi,0xa
  41aa98:	48 39 fb             	cmp    rbx,rdi
  41aa9b:	73 07                	jae    41aaa4 <camlDebug_test__do_access_0_1_code+0x14>
  41aa9d:	48 8b 04 d8          	mov    rax,QWORD PTR [rax+rbx*8]
  41aaa1:	c3                   	ret    
  41aaa2:	66 90                	xchg   ax,ax
  41aaa4:	48 8d 05 a5 a8 03 00 	lea    rax,[rip+0x3a8a5]        # 455350 <camlDebug_test__do_access_1+0x20>
  41aaab:	49 8b 66 10          	mov    rsp,QWORD PTR [r14+0x10]
  41aaaf:	41 8f 46 10          	pop    QWORD PTR [r14+0x10]
  41aab3:	41 5b                	pop    r11
  41aab5:	41 ff e3             	jmp    r11
  41aab8:	0f 1f 84 00 00 00 00 	nop    DWORD PTR [rax+rax*1+0x0]
  41aabf:	00 
...
```

tagged int for comparison:

```
  41aa10:	48 8b 78 f8          	mov    rdi,QWORD PTR [rax-0x8]
  41aa14:	48 c1 ef 09          	shr    rdi,0x9
  41aa18:	48 39 fb             	cmp    rbx,rdi
  41aa1b:	73 07                	jae    41aa24 <camlDebug_test__do_access_0_1_code+0x14>
  41aa1d:	48 8b 44 98 fc       	mov    rax,QWORD PTR [rax+rbx*4-0x4]
  41aa22:	c3                   	ret    
  41aa23:	90                   	nop
  41aa24:	48 8d 05 25 a9 03 00 	lea    rax,[rip+0x3a925]        # 455350 <camlDebug_test__do_access_1+0x20>
  41aa2b:	49 8b 66 10          	mov    rsp,QWORD PTR [r14+0x10]
  41aa2f:	41 8f 46 10          	pop    QWORD PTR [r14+0x10]
  41aa33:	41 5b                	pop    r11
  41aa35:	41 ff e3             	jmp    r11
  41aa38:	0f 1f 84 00 00 00 00 	nop    DWORD PTR [rax+rax*1+0x0]
  41aa3f:	00 
...
```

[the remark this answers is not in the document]
Gbury: Okay, that's indeed good reasons. Maybe you could put a comment to that effect at that point in the code ?

Gbury (approved):

Gbury (approved):

PR #2338, opened 2024-03-06
Remove a check for values in `with type` constraints

goldfirere:
The first commit adds a failing test, and then the second commit fixes the bug. Except that we then hit the package hack, so this is in a race with #2253. But I'm pretty sure the bug fixed here is independent of the package hack.

Suggested review: @alanechang

### DIFF
--- a/ocaml/typing/typemod.ml
+++ b/ocaml/typing/typemod.ml
@@ -3272,7 +3272,6 @@ let type_package env m p fl =
   in
   List.iter
     (fun (n, ty) ->
-       (* CR layouts v5: relax value requirement. *)
       try Ctype.unify env ty
             (Ctype.newvar (Jkind.any ~why:Dummy_jkind))
       with Ctype.Unify _ ->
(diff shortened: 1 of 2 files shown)


ocaml/typing/typemod.ml
```diff
@@ -3274,7 +3274,7 @@ let type_package env m p fl =
     (fun (n, ty) ->
        (* CR layouts v5: relax value requirement. *)
```
alanechang, line 3275: Remove this CR?

alanechang (approved):
The changes look good.

---

Interesting how:

```ocaml
module type S = sig
  type t : any
end

module C = struct
  type t : any
end

type t_any : any
let x = (module C : S with type t = t_any)
```

will get the good layout history error. Suspect in the test case provided, `'a` is constrained to `value` for one of the package hack checks but errors at this check in `type_package`. Did not investiagte further since all of these checks are going away soon.

ocaml/typing/typemod.ml
goldfirere: D'oh. :facepalm: 

Thanks

PR #2346, opened 2024-03-08
Implicit source positions directory locations

Enoumy:
Adds a `-directory` cli flag to the compiler
-------------------------------------------------------
We would like for the implicit source position argument to include the "directory" and not just the basename of the file, for example:

```ocaml
(* in app/foo/main.ml*)
let f ~(src_pos : [%src_pos]) () = src_pos

let pos : string = (f ()).pos_fname
```
In the above example the value of `pos` would be `"main.ml"` (just the basename), when we would instead like it to be `"app/foo/main.ml"` where `"app/foo"` is the directory relative to the "root" of the build system (e.g. relative to the location of `dune-project`)

This feature adds a `-directory` flag the the build system will provide when it calls the compiler. This feature also uses this flag for the implicit source position arguments.

Testing
---------
I have added a test `ocaml/testsuite/tests/typing-implicit-source-positions/location_with_directory.ml` that tests that the implicit source position argument uses the directory location. Please let me know if this testing is sufficient, if I should add other tests. Thanks!


ocaml/driver/main_args.ml
```diff
@@ -1630,6 +1638,7 @@ struct
     mk_no_alias_deps F._no_alias_deps;
     mk_app_funct F._app_funct;
     mk_no_app_funct F._no_app_funct;
+    mk_directory F._directory;
```
goldfirere, line 1641: I don't think we need this in ocamldoc, which is basically deprecated.

goldfirere (commented):
Code looks good. Just want an opinion from a few others before merging.

goldfirere (approved):
Turns out @Enoumy was already on the "asking around" part. Let's merge.

ocaml/driver/main_args.ml
Enoumy: Removed!

PR #2347, opened 2024-03-08
Parsing rule for modal kinds

alanechang:
Adds only the parsing rules for modal kinds. This includes the new kind annotation syntax and a new `kind_abbrev_` construct.

`Misc.fatal_error` gets called when parsing these rules.


alanechang:
@goldfirere This PR is ready for some review.

goldfirere (approved):

PR #2348, opened 2024-03-11
Delay the start of marking with a new GC phase

stedolan:
During the new GC phase, marking does not occur. This means that new allocations can be UNMARKED, and caml_modify need not darken the values it sees.

Darkening fewer values means that memory allocated during this phase can be reused more quickly, which reduces heap usage.

For now, this phase continues until sweeping is done. The implementation is currently single-domain only, and the upcoming multi-domain version will not necessarily guarantee that sweeping is completed on all domains before marking begins: the important condition is that marking has not begun until the new phase ends, not that sweeping has ended. So, the new phase is called `Phase_sweep_main`, and the existing phase keeps its name `Phase_sweep_and_mark_main`.

Ephemerons allocated during `Phase_sweep_main` require some special handling: they are added to the `todo` list like ephemerons allocated during the previous cycle, since they are not known to be marked.

### DIFF
--- a/ocaml/runtime/domain.c
+++ b/ocaml/runtime/domain.c
@@ -1784,11 +1784,11 @@ static void handover_finalisers(caml_domain_state* domain_state)
 
   if (f->todo_head != NULL || f->first.size != 0 || f->last.size != 0) {
     /* have some final structures */
-    if (caml_gc_phase != Phase_sweep_and_mark_main) {
+    if (caml_gc_phase != Phase_sweep_main) {
       /* Force a major GC cycle to simplify constraints for
        * handing over finalisers. */
       caml_finish_major_cycle(0);
-      CAMLassert(caml_gc_phase == Phase_sweep_and_mark_main);
+      CAMLassert(caml_gc_phase == Phase_sweep_main);
     }
     caml_add_orphaned_finalisers (f);
     /* Create a dummy final info */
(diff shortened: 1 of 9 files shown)


ocaml/runtime/domain.c
```diff
@@ -1784,11 +1784,11 @@ static void handover_finalisers(caml_domain_state* domain_state)
 
   if (f->todo_head != NULL || f->first.size != 0 || f->last.size != 0) {
     /* have some final structures */
-    if (caml_gc_phase != Phase_sweep_and_mark_main) {
```
lpw25, line 1787: I think this surprises me. I'd expect something that used to be ok during marking to still be ok during marking.

lpw25 (approved):
This looks correct to me. A couple of small points.

ocaml/runtime/domain.c
stedolan: Yeah, that seems fine. The important thing here is that we're not in one of the ephemeron processing subphases.

ocaml/runtime/domain.c
```diff
@@ -1784,7 +1784,8 @@ static void handover_finalisers(caml_domain_state* domain_state)
 
   if (f->todo_head != NULL || f->first.size != 0 || f->last.size != 0) {
     /* have some final structures */
-    if (caml_gc_phase != Phase_sweep_main) {
+    if (caml_gc_phase != Phase_sweep_main &&
+        caml_gc_phase != Phase_sweep_and_mark_main) {
```
lpw25, line 1788: This means that `caml_add_orphaned_finalisers` can be called in `Phase_sweep_and_mark_main` so it's assertion also needs updating. I've done that in #2358.

PR #2350, opened 2024-03-11
Modal kind jane syntax encoding

alanechang:


alanechang:
This PR adds jane syntax encoding for the new modal kind syntax. This includes a rework of the existing layout annotation encoding plus additional logic to handle `kind_abbrev_`.

Attempts to use the new syntax will still result in fatal_error, but now from the type checker instead of the parser.

Tests are mainly added to `pprintast_unconditional.ml` which exercises the encoding, decoding, and printing logic. 

Requesting review from @goldfirere

goldfirere (approved):
@liam923 please review my last commit. Then, if it all looks good, rebase against `main` and merge. Thanks!

PR #2353, opened 2024-03-12
Zero alloc: don't loose witnesses

gretay-js:
There is a mismatch in handling witnesses: `V.lessequal` ignores them and `V.join` does not.  As a result, witnesses discovered in the last iteration of a fixed point computation might be lost, if there are already other witnesses for the same component.  
The analysis uses `V.lessequal new_value value` to detect fixed points and will not join the `new_value` to the preiously computed value if the only difference is in witnesses. 

This is not a soundness problem in the sense that the check will not miss allocations, but it might not return all allocations as witnesses. We do not guarantee that all witnesses will be reported, but it is a nice property to have. We rely on the witnesses produced by the analysis to visualize allocations and not only to check annotations. Having consistent output also makes testing easier. 

This PR fixes it "properly" by taking witnesses into account in `V.lessequal` and refactoring some other witness related checks to make it clear that we are allowed to ignore witnesses of "expected" values that come from annotations, and not the computed values. The downside of this approach is that the fixpoint can take longer: the height of the abstract domain now depends on the size of the input file in the worst case, instead of being constant height of `3`. I think that the bad case cannot happen in practice because of the order or propagation of witnesses along all the control flow paths within a loop before checking for a fixed point. 

An alternative is to call `join` even when `lessequal` is true. It works too but feels like a hack.

For example, the output of existing test `fail19.ml` changes depending on the order in which functions are emitted in the file: `-function-layout topological` includes `probe` witness and with the `source` layout it doesn't. This happens because the fixpoint for "forward defined" functions discovered a new witness but the value was already `Top`, so the new witness is not recorded. I haven't found a way to reproduce this on main yet.

On top of #2290.

### DIFF
--- a/backend/checkmach.ml
+++ b/backend/checkmach.ml
@@ -23,7 +23,7 @@
  * SOFTWARE.                                                                      *
  *                                                                                *
  **********************************************************************************)
-[@@@ocaml.warning "+a-30-40-41-42-32"]
+[@@@ocaml.warning "+a-30-40-41-42"]
 
 module String = Misc.Stdlib.String
 
@@ -651,8 +651,6 @@ module Unit_info : sig
     Mach.instruction option ->
     unit
 
-  val update : t -> string -> Value.t -> unit
-
   val iter : t -> f:(Func_info.t -> unit) -> unit
 
   val should_keep_witnesses : bool -> bool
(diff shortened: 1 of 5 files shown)


xclerc (approved):
Do we have an idea of the size of the
witness sets in practice? With the
comparision over the domain now
possibly involving an inclusion test,
the cost could increase significantly.

gretay-js:
> Do we have an idea of the size of the witness sets in practice? With the comparision over the domain now possibly involving an inclusion test, the cost could increase significantly.

I am collecting some statistics on this RP and it's baseline #2290.

gretay-js:
I messed up the repo/branch in this PR, sorry. I am closing this in favor of #2354 to reduce the confusion. We can continue the discussion there.

PR #2356, opened 2024-03-13
Zero alloc: flag to disable precise analysis of forward functions

gretay-js:
Add compiler flag `-disable-precise-checkmach` that turns on conservative zero_alloc interpretation of foward calls and mutually recursive functions. 

This is intended to be a temporary measure while rolling #2290 and #2354 in the case we encounter a pathological case in which computing `zero_alloc` summaries is very expensive,  especially on functions that are not even expected to be `zero_alloc`. The flag will allow us to temporarily disable the fixpoint computation of summaries on a per file or library basis. It makes `zero_alloc` analysis cheaper but also less precise, so some annotations may fail the check with this flag. Unlike `-disable-checkmach` that turns off the computation of summaries competely, `disable-precise-checkmach`  will compute fixpoints within a function as usual, and will be able to check most annotations. 

On top of #2341.


tests/backend/checkmach/fail26.ml
```diff
@@ -0,0 +1,63 @@
+(* all functions fail the check when -disable-checkmach is passed. *)
```
xclerc, line 1: (Is it the same contents as
`fail25.ml`? if so, should we
have only one file processed
with different sets of flags?)

xclerc (approved):

tests/backend/checkmach/fail26.ml
gretay-js: yes, the files are identical but the little `gen_dune.ml` script doesn't support "same file with different flags" yet, as it would need a different output name.

backend/checkmach.ml
```diff
@@ -849,11 +849,19 @@ end = struct
           unresolved v "self-call init"
         | Some approx -> unresolved approx "self-call approx"
       else
-        ((* Call is defined later in the current compilation unit. Summary of
-            this callee is not yet computed, conservatively return Top. Won't be
-            able to prove any recursive functions as non-allocating. *)
-         t.unresolved <- true;
-         unresolved Value.safe)
+        let res =
+          if (* Call is defined later in the current compilation unit. Summary
```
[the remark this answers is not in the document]
gretay-js: Fixed. Also moved this up to cover "self-recursive" calls.

tests/backend/checkmach/fail26.ml
gretay-js: fixed.

xclerc (approved):

PR #2357, opened 2024-03-13
Classic mode: partially lift the restrictions on inlining functions containing closures

lthls:
This PR proposes to lift two restrictions on inlining of functions in classic mode.
Before this PR, every function whose body contains a function (in Lambda) would be unavailable for inlining.
This PR contains two independent commits lifting this restriction is specific cases:
- The first commit makes functions eligible again if their Flambda2 version does not contain any closure allocation. Typically, constant functions will not prevent inlining of the surrounding context anymore.
- The second commit makes an explicit `[@inline always]` attribute override the restriction. This makes us diverge from Closure, but unlike Closure we don't duplicate the code of the subfunctions when inlining the closure so there isn't any danger in doing that.

I can work on a way to improve the error messages when the restriction prevents us from inlining a function too, but it is a bit more invasive.

### DIFF
--- a/middle_end/flambda2/from_lambda/closure_conversion.ml
+++ b/middle_end/flambda2/from_lambda/closure_conversion.ml
@@ -2230,7 +2230,9 @@ let close_one_function acc ~code_id ~external_env ~by_function_slot
     else meta
   in
   let acc = Acc.add_code ~code_id ~code acc in
-  let acc = Acc.with_seen_a_function acc true in
+  let acc =
+    if has_lifted_closure then acc else Acc.with_seen_a_function acc true
+  in
   ( acc,
     ( Function_slot.Map.add function_slot approx by_function_slot,
       function_code_ids ) )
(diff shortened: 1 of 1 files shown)


Gbury (approved):

lpw25:
I guess `[@inline available]` should probably override this behaviour too. If the user asked for something, and we know how to do it, we might as well give them what they want.

lthls:
> I guess `[@inline available]` should probably override this behaviour too. If the user asked for something, and we know how to do it, we might as well give them what they want.

Done. There's still `[@unroll]` left; I don't think we've seen much use of it so it probably doesn't matter, but we could probably handle it as `[@inline]`.

lpw25:
It wasn't clear to me whether classic mode can handle `[@unroll]` if it can then I agree it should override the other behaviour.

lthls:
It can inline recursive functions, but not in their own body (and only once, of course). So we could interpret such an attribute as `[@inline always]`, or ignore it, but we cannot give it any unrolling semantics.

lpw25:
As you say, it probably doesn't matter, but I guess `[@inline always]` is closer to the user's intention than ignoring it.

mshinwell:
It's not clear to me that the existing semantics of unrolling on this PR is bad; in some ways it's simpler - "unrolling is not supported in classic mode" rather than having some other definition of what it means to unroll.  I'd like to merge this now and we can revisit in the future if necessary.

PR #2359, opened 2024-03-14
Fix Unbound module Stdlib__Nativeint_u

poechsel:
Building `make world.opt` in `ocaml/` would fail due to an unbound module Stdlib__Nativeint_u


alanechang (approved):
Thanks for catching this issue! It only shows up reliably when you do `make -j1 world` (having more concurrent jobs can make the error go away sometimes).

The changes look good and make the comment in the file true again:

```
# Basenames of the source files for the standard library (i.e. unprefixed and
# with lowercase first letters). These must be listed in dependency order.
```

PR #2360, opened 2024-03-14
Fix case where local functions optimisation breaks tail calls

lpw25:
The following code will currently stack overflow rather than loop forever:
```ocaml
let[@inline never] ignore_local (local_ x) = ()

let rec foo p a b =
  let bar x y =
    foo p x y
  in
  ignore_local (Some a);
  if p then
    bar a b
  else
    bar b a

let () = foo true 1 2
```
due to the "local functions" optimisation moving the tailcalls inside of the region.

To fix this issue, this PR adds support in lambda for `Lstaticcatch` constructs in the tail of a region, where the handler is executed after the region has closed. This is achieved by adding  a `pop_region` flag to `Lstaticcatch` and updating the cps conversion code to handle it appropriately.

Once this feature is available the fix to the "local functions" optimisation is trivial.

### DIFF
--- a/middle_end/flambda2/from_lambda/lambda_to_flambda.ml
+++ b/middle_end/flambda2/from_lambda/lambda_to_flambda.ml
@@ -1041,14 +1041,12 @@ let rec cps acc env ccenv (lam : L.lambda) (k : cps_continuation)
     maybe_insert_let_cont "staticcatch_result" layout k acc env ccenv
       (fun acc env ccenv k ->
         let pop_region =
-          match r with
-          | Popped_region -> true
-          | Same_region -> false
+          match r with Popped_region -> true | Same_region -> false
         in
         let continuation = Continuation.create () in
         let { Env.body_env; handler_env; extra_params } =
-          Env.add_static_exn_continuation env static_exn
-            ~pop_region continuation
+          Env.add_static_exn_continuation env static_exn ~pop_region
+            continuation
         in
         let recursive : Asttypes.rec_flag =
           if Env.is_static_exn_recursive env static_exn
--- a/middle_end/flambda2/from_lambda/lambda_to_flambda.ml
+++ b/middle_end/flambda2/from_lambda/lambda_to_flambda.ml
@@ -127,8 +127,6 @@ let compile_staticfail acc env ccenv ~(continuation : Continuation.t) ~args :
        allocation region"
       Continuation.print continuation;
   let rec add_end_regions acc ~region_stack_now =
-    (* This can maybe only be exercised right now using "match with exception",
-       since that causes jumps out of try-regions (but not normal regions). *)
     (* CR pchambart: This closes all the regions between region_stack_now and
        region_stack_at_handler, but closing only the last one should be
        sufficient. *)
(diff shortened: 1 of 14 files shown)


mshinwell:
@lpw25 maybe my comment about printing on the follow-up PR should have been here

mshinwell:
@Ekdohibs would you be able to review this?

Ekdohibs:
It looks like the tests are failing, probably due to printing changes. Other than that, the code looks good, but I'm not sure about one thing: what is the difference between the region pop happening on `Lstaticcatch` and adding an `Lexclave` either at the beginning of the handler or on `Lstaticraise`?

mshinwell:
I've pinged Leo separately on this point.

mshinwell:
(also this PR was broken for the upstream build, but that has gone away now)

lpw25:
> what is the difference between the region pop happening on Lstaticcatch and adding an Lexclave either at the beginning of the handler or on Lstaticraise?

We want to be able to do `Lstaticraise` to a handler from inside an `Lexclave`. Currently all `Lstaticcatch` handlers are forced to be in the current region and so that is an error. This PR makes it possible to annotate an `Lstaticcatch` handler as being in the parent region so that this is allowed.

ocaml/lambda/simplif.ml
```diff
@@ -156,18 +156,24 @@ let simplify_exits lam =
   | Lstaticraise (i,ls) ->
       incr_exit i 1 try_depth;
       List.iter (count ~try_depth) ls
-  | Lstaticcatch (l1,(i,[]),Lstaticraise (j,[]), _) ->
+  | Lstaticcatch (l1,(i,[]),Lstaticraise (j,[]), _, _) ->
```
Ekdohibs, line 159: It seems to me this can make a pop_region disappear?

middle_end/flambda2/from_lambda/lambda_to_flambda.ml
```diff
@@ -1043,12 +1044,18 @@ let rec cps acc env ccenv (lam : L.lambda) (k : cps_continuation)
         compile_staticfail acc env ccenv ~continuation
           ~args:(List.flatten args @ extra_args))
       k_exn
-  | Lstaticcatch (body, (static_exn, args), handler, layout) ->
+  | Lstaticcatch (body, (static_exn, args), handler, r, layout) ->
     maybe_insert_let_cont "staticcatch_result" layout k acc env ccenv
       (fun acc env ccenv k ->
+        let pop_region =
+          match r with
+          | Popped_region -> true
```
Ekdohibs, line 1052: So in the `pop_region` case, it looks like the difference with what would happen if we had `Lstaticcatch (_, _, Lexclave _, _)` is that we don't add an `End_region`, isn't that a problem?

lpw25: No, that's the point: the region will already have been popped when we enter the handler. Perhaps this would all be clearer if I named various things `parent_region` rather than `pop_region`?

ocaml/lambda/simplif.ml
lpw25: Not as long as the original lambda was well-formed. If this `Lstaticcatch` is in the parent region then it can only jump to other handlers that are also in the parent region.

The key point here is that `pop_region` is an annotation describing what region the handler is in, not an instruction to insert an `End_region` construct. As said above, I think perhaps a rename would make this clearer.

lpw25:
> a Lstaticraise must have the same region stack as its handler

It's slightly more relaxed than that: an `Lstaticraise` must have the same region or a deeper region as its handler -- we will automatically add `End_region`s if you try to jump to an shallower region. The thing we can't support is jumping to a deeper region because there is no way to re-enter a region once you have left it.

>  what about the case where we might want to jump to a handler through two levels of Lexclave?

I think that we currently don't currently generate nesting multiple `Lexclave` which is what you would need to have before you needed to have a handler that was two levels up. I'm not actually sure if nested `Lexclave` would compile correctly or not, but I'm pretty sure we can't currently generate them.

Ekdohibs (approved):

PR #2361, opened 2024-03-14
Try to share code to pop regions in local functions optimisation

lpw25:
This PR is an improvement on top of #2360 that shares the code to pop the region in the common case where all the calls to the local function are `rc_close_on_apply` calls for the region in which the function was defined. It is on top of #2360, so only the second commit should be reviewed here.


mshinwell:
@lpw25 maybe something wrong with the Lambda printing code, looking at the test failures?

lpw25:
I've rebased this onto main now that #2360 has been merged.

Ekdohibs (approved):

PR #2362, opened 2024-03-14
float32 flambda2 support

TheNumbat:
Implements `float32` numbers in flambda2. 

Uses `float` to represent `float32` constants throughout the compiler, but the flambda2 numeric type (`Float32_by_bit_pattern`) is an actual 32-bit float represented by an int32.
This PR includes all necessary infrastructure for unboxed float32s, but there is no support for `float32# array` yet.

Tests to come in a following PR that implements the missing cmm operations.


mshinwell:
@TheNumbat I was expecting to see C functions being used for the arithmetic operations on `Float32_by_bit_pattern.t`.  How can we rigorously justify that the OCaml versions give the correct results?  (cc @stedolan with whom I discussed this briefly some time ago.)

TheNumbat:
Stephen mentioned division might differ due to being specified to 0.5 ulp, but I haven't found a definitive source for that.
I think I should just implement the real float32 operations in C. Do you think doing so in `Float32_by_bit_pattern` is sufficient, or should I also mint a custom `float32` type to use elsewhere in the compiler? 
(It is, at least, true that converting float->double->float preserves the value...but not necessarily the specific NaN bit pattern?)

mshinwell:
I spoke to @stedolan last night and it sounds like emulation of float32 division using float64 might produce different results, because of the two roundings involved.  I think it's probably best if we use C stubs here.

stedolan:
Longer explanation:

Finite Float32 and Float64 values are defined as a particular subset of the reals. Arithmetic on them is specified as doing real arithmetic (to infinite precision), and then rounding according to a "rounding rule" (OCaml only supports the default one, round-to-nearest, ties-to-even).

A result is "correctly rounded" if it is the true real result rounded according to the rounding rule (i.e. the nearest float to the actual answer). A result is "faithfully rounded" if it is one of the two consecutive floats whose interval contains the true result (i.e. rounded up or rounded down, not necessarily to the nearest). Correct rounding has an error of at most 0.5ulp (since if the error was >0.5ulp it rounds the other way), and faithful rounding has an error of at most 1ulp.

Doing arithmetic on float32 by converting to float64, doing arithmetic there, and then converting back has the effect of doing a double rounding: first the value is rounded to a float64 and then to a float32.

Every float32 is a float64, so double-rounding is faithful: we cannot skip over a float32 by first rounding to a float64 because there are no float32s between two float64s. Many float library functions (e.g. trig functions, log, etc.) only guarantee faithful rounding (if even!), so nothing is lost by going via float64.

However, the IEEE754 spec requires that "elementary arithmetic" (addition, subtraction, multiplication, division, square root and FMA) is correctly rounded, not merely faithfully rounded. Double rounding does not in general preserve correct rounding. For instance, this is the cause of floating point inaccuracies using the x87 FPU, where double arithmetic sometimes goes via 80-bit registers and double-rounds.

As it happens, it is tricky to construct an counterexample where 32->64bit double rounding breaks things. The standard counterexamples (e.g. in the Handbook of Floating-point Arithmetic) are, I think, impossible to construct using addition, subtraction or multiplication of float32, since float64 has a mantissa with more than twice the precision of float64. (I think it can be done for division, though, and I have no idea about sqrt or fma).

So, I think the thing to do here is to use C stubs that do explicit float32 arithmetic for the six elementary arithmetic operations. (Possibly you can skip these for three of them, but that requires more thought than I've given it).

mshinwell:
@TheNumbat I think just doing it for `Float_by_bit_pattern` should be sufficient.

Gbury (commented):
- I haven't reviewed the C code for the float 32 computation as I'm not very expert about that
- Some parts of the code do not handle 32-bits float arrays (with some `assert false` and associated comment), while other parts of the code do try to handle these at least partially: this is slightly misleading/confusing wen reading the code and trying to understand what is being supported or not.

middle_end/flambda2/simplify/simplify_extcall.ml
```diff
@@ -95,6 +95,10 @@ let simplify_comparison ~dbg ~dacc ~cont ~tagged_prim ~float_prim
   | Proved (Boxed (_, Naked_float, _)), Proved (Boxed (_, Naked_float, _)) ->
     simplify_comparison_of_boxed_numbers ~dbg dacc cont a b ~kind:Naked_float
       ~cmp_prim:float_prim
+  | Proved (Boxed (_, Naked_float32, _)), Proved (Boxed (_, Naked_float32, _))
+    ->
+    simplify_comparison_of_boxed_numbers ~dbg dacc cont a b ~kind:Naked_float32
+      ~cmp_prim:float_prim
```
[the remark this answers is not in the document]
TheNumbat: I think this is fine - the following PR parameterizes float_comp over the width 
https://github.com/ocaml-flambda/flambda-backend/pull/2384/files#diff-0985b96a2d567e584a9e7ee680993fdae7f33a8566d8568921557be695200b9d

middle_end/flambda2/simplify/lifting/reification.ml
```diff
@@ -50,11 +51,15 @@ let create_static_const dacc dbg (to_lift : T.to_lift) : RSC.t =
       if is_unique then Immutable_unique else Immutable
     in
     RSC.create_block art tag mut ~fields
+  | Boxed_float32 f -> RSC.create_boxed_float32 art (Const f)
   | Boxed_float f -> RSC.create_boxed_float art (Const f)
   | Boxed_int32 i -> RSC.create_boxed_int32 art (Const i)
   | Boxed_int64 i -> RSC.create_boxed_int64 art (Const i)
   | Boxed_nativeint i -> RSC.create_boxed_nativeint art (Const i)
   | Boxed_vec128 v -> RSC.create_boxed_vec128 art (Const v)
+  | Immutable_float32_array { fields } ->
```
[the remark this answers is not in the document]
TheNumbat: Made this one an assert as well - I will add float32# and float32# array after these PRs get merged

middle_end/flambda2/to_cmm/to_cmm_static.ml
```diff
@@ -295,6 +307,13 @@ let static_const0 env res ~updates (bound_static : Bound_static.Pattern.t)
     let header = C.black_block_header 0 0 in
     let block = C.emit_block sym header [] in
     env, R.set_data res block, updates
+  | Block_like s, Empty_array Naked_float32s ->
+    let block =
+      C.emit_block (R.symbol res s)
+        (C.black_custom_header ~size:1)
+        [C.symbol_address (Cmm.global_symbol "caml_unboxed_float32_array_ops")]
```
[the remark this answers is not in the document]
TheNumbat: Yeah, this PR shouldn't have any float32 array handling. Removed here

external/float32/float32_stubs.c
```diff
@@ -0,0 +1,332 @@
+/**************************************************************************/
+/*                                                                        */
+/*                                 OCaml                                  */
+/*                                                                        */
+/*                       Pierre Chambart, OCamlPro                        */
+/*           Mark Shinwell and Leo White, Jane Street Europe              */
+/*                        Max Slater, Jane Street                         */
```
xclerc, line 7: This should probably retain the copyright
from upstream.

external/float32/float32_stubs.c
```diff
@@ -0,0 +1,332 @@
...
+}
+
+value compiler_float32_to_float_boxed(value i)
+{
+    return caml_copy_double(compiler_float32_to_float(Int32_val(i)));
+}
+
+static int compiler_float32_of_hex(const char * s, const char * end, float * res)
+{
+  /* See caml_float_of_hex */
```
xclerc, line 191: There is a fair amount of code duplication;
if it is not possible/desirable to factor out,
we should probably have a short comment
explaining why.

xclerc (commented):
(I have only read the files under
`backend/` and `external/`.)

external/float32/float32_stubs.c
TheNumbat: Done

external/float32/float32_stubs.c
TheNumbat: Added a comment; we can't use a modified runtime in the flambda compiler build

TheNumbat: (The near-identical versions in float32.c are added in a later PR)

TheNumbat: (The near-identical versions in float32.c are added in a later PR)

Gbury (approved):

PR #2363, opened 2024-03-14
Use `Meet_with` for mode crossing

riaqn:
This PR uses the `Meet_with` from #2302 to do mode crossing, so now mode crossing supports non-trivial `upper_bounds` of modal kinds.

In that process, some cleaning up of the mode system is required. In particular, to express `alloc_mode_cross_to_max_min`, we need to split the `upper_bound : Alloc.Const.t` into `Monadic.Const.t` and `Comonadic.Const.t`. Previously they are exposed in `check_const` as unlabelled tuples, which is bad. They are now opaque, because the user never needs to look into the axes.

### DIFF
--- a/ocaml/typing/mode.ml
+++ b/ocaml/typing/mode.ml
@@ -1259,6 +1259,7 @@ module Comonadic_with_regionality = struct
   end
 
   include Common (Obj)
+  open Obj
 
   type error =
     [ `Regionality of Regionality.error
@@ -1308,6 +1309,11 @@ module Comonadic_with_regionality = struct
       (Meet_with (C.max_with Obj.obj Linearity c))
       (S.Positive.disallow_right m)
 
+  let meet_with c m =
+    Solver.via_monotone obj (Meet_with c) (Solver.disallow_right m)
+
+  let imply c m = Solver.via_monotone obj (Imply c) (Solver.disallow_left m)
+
   let zap_to_legacy m =
     let regionality = regionality m |> Regionality.zap_to_legacy in
     let linearity = linearity m |> Linearity.zap_to_legacy in
@@ -1351,6 +1357,7 @@ module Comonadic_with_locality = struct
   end
 
   include Common (Obj)
+  open Obj
 
   type error =
     [ `Locality of Locality.error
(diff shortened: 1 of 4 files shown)


ocaml/typing/typecore.ml
```diff
@@ -802,76 +802,35 @@ let has_poly_constraint spat =
...
-    if Linearity.Const.le upper_bounds.linearity Linearity.Const.min
-    then Value.join_with_linearity Linearity.Const.max mode
-    else mode
-  in
-  let mode =
-    if Uniqueness.Const.le upper_bounds.uniqueness Uniqueness.Const.min
-    then Value.join_with_uniqueness Uniqueness.Const.max mode
-    else mode
-  in
-  { expected_mode with mode; strictly_local }
```
riaqn, line 874: We don't need to update `strictly_local`, because `strictly_local` is only read when type checking functions, and functions don't cross locality.

ocaml/typing/mode.ml
```diff
@@ -1259,6 +1259,7 @@ module Comonadic_with_regionality = struct
   end
 
   include Common (Obj)
+  open Obj
```
goldfirere, line 1262: If we're going to `open Obj`, then best to get rid of the various `Obj.` uses below.

ocaml/typing/mode.ml
```diff
@@ -1351,6 +1357,7 @@ module Comonadic_with_locality = struct
...
```
goldfirere, line 1360: Ditto here.

ocaml/typing/mode.ml
```diff
@@ -1443,6 +1455,7 @@ module Monadic = struct
...
```
goldfirere, line 1458: And here.

ocaml/typing/typecore.ml
goldfirere: I agree. Please put this comment in the code, though, where it is easier for our future selves to find. Thanks.

goldfirere (approved):

ocaml/typing/mode.ml
riaqn: Yeah, I was uncertain about this: this would introduce changes unrelated to this PR; those changes can be done in the future lazily.

riaqn: During a meeting today we briefly discussed this, and think it's fine to leave it as is.

PR #2364, opened 2024-03-15
[Implicit Source Positions] - Better Error Messages

Enoumy:
This feature is a catchall feature for lingering changes for the implicit source positions project. This feature contains small changes like:

- Renaming `[%src_pos]` -> `[%call_pos]`
- Addressing a couple of `CR src_pos`'es on better error messages when the implicit source position argument is misused.
- Addressing a CR src_pos that was missed when adding support for the class system a CR src_pos on ocamldoc was missed.
- Deleted an already addressed CR src_pos.

For reviewers
------------------
There are two CR-questions that I left regarding previously existing CR src_pos as I had some questions on them. I left them in the following files:
- ocaml/typing/typetexp.ml:712
- ocaml/typing/typecore.ml:7586

Testing
---------
[X] - Existing CI passes
[X] - Updated tests with the better error messages.

### DIFF
--- a/ocaml/typing/typecore.ml
+++ b/ocaml/typing/typecore.ml
@@ -9875,14 +9875,21 @@ let report_error ~loc env = function
         | Nolabel, _ | _, Nolabel -> true
         | _                       -> false
       in
+      let maybe_positional_argument_hint = 
+        match got, expected with
+        | Labelled _, Position _ ->
+          "\nHint: Consider explicitly annotating the label with '[%call_pos]'"
+        | _ -> ""
+      in
       Location.errorf ~loc
         "@[<v>@[<2>This function should have type@ %a%t@]@,\
-         @[but its first argument is %s@ instead of %s%s@]@]"
+         @[but its first argument is %s@ instead of %s%s@]%s@]"
         Printtyp.type_expr expected_type
         (report_type_expected_explanation_opt explanation)
         (label ~long:true got)
         (if second_long then "being " else "")
         (label ~long:second_long expected)
+        maybe_positional_argument_hint
   | Scoping_let_module(id, ty) ->
       Location.errorf ~loc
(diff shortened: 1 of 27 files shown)


ocaml/typing/typecore.ml
```diff
@@ -7583,7 +7583,9 @@ and type_apply_arg env ~app_loc ~funct ~index ~position_and_mode ~partial_app (l
           (lbl, Arg (arg, Mode.Value.legacy, sort_arg))
       | Position _ ->
           let arg = src_pos (Location.ghostify app_loc) [] env in
-          (* CR src_pos: Confirm that global value mode is correct *)
+          (* XXX jrodri for goldfirere: I think using [Mode.Value.legacy] here is correct,
```
goldfirere, line 7586: Yes, it's correct, by analogy with optional arguments.

ocaml/typing/typetexp.ml
```diff
@@ -709,9 +709,13 @@ and transl_type_aux env ~row_context ~aliased ~policy mode styp =
           let l = transl_label l (Some arg) in
           let arg_cty =
             if Btype.is_position l then
+              (* XXX jrodri for goldfirere: Hmm, I think that I do not understand the
```
goldfirere, line 712: Don't worry about it -- not worth trying to recreate whatever simplification I thought might be possible last summer. There are better ways to spend our time. :)

goldfirere (approved):

ocaml/typing/typecore.ml
Enoumy: Cool sounds good. I have removed the CR question. Thanks!

ocaml/typing/typetexp.ml
Enoumy: Cool that sounds reasonable. I've removed the cr question.

PR #2366, opened 2024-03-16
Remove `mode_subcomponent`

riaqn:
Currently we have `mode_subcomponent` which calls `value_regional_to_global`. This is because the container is an allocation, so if the container is at mode `regional`, the container should be allocated at `global`, and the subcomponents therein should follow that. This can be de-coupled into two steps:
- First, because the container is allocation, the "real" expected_mode of the container is the one refined via `value_regional_to_global`.
- Second, subcomponents should follow that "real" expected_mode. 

- The first step is already handled by `register_allocation`. Indeed, there are many occurrences of successive calls to both `register_allocation` and `mode_subcomponent`, which is not nice.
- The second step is a trivial one : just call `mode_default expected_mode.mode`.

And therefore we are able to remove `mode_subcomponent`.


ocaml/typing/typecore.ml
```diff
@@ -2176,17 +2176,14 @@ let map_fold_cont f xs k =
   List.fold_right (fun x k ys -> f x (fun y -> k (y :: ys)))
     xs (fun ys -> k (List.rev ys)) []
 
-let type_label_a_list loc closed env usage type_lbl_a expected_type lid_a_list =
+let disambiguate_sort_lid_a_list loc closed env usage expected_type lid_a_list =
   let lbl_a_list =
     disambiguate_lid_a_list loc closed env usage expected_type lid_a_list
```
lpw25, line 2181: This is the only use of `disambiguate_lid_a_list`. Maybe just push the sorting code into `disambiguate_lid_a_list` and have get rid of `disambiguate_sort_lid_a_list`.

ocaml/typing/typedtree.mli
```diff
@@ -333,10 +340,9 @@ and expression_desc =
             in which case it does not need allocation.
           *)
   | Texp_field of expression * Longident.t loc * Types.label_description *
-      unique_use * Mode.Alloc.r option
-    (** [alloc_mode] is the allocation mode of the result; available ONLY
-        only when getting a (float) field from a [Record_float] record
-      *)
+      texp_field_float
```
lpw25, line 343: I'd rather leave `unique_use` as a separate field. It's not obvious why you don't need the `unique_use` for float projections, and it feels like it is conflating separate concerns: whether the projection allocates and whether it counts as a unique use. It doesn't count as a unique use essentially because unboxed floats cross the uniqueness modes, but you could in principle make other types that did boxing on projection where that wasn't true.

lpw25 (approved):
Looks good. A few suggestions.

ocaml/typing/typedtree.mli
riaqn: Is there some example where we need `unique_use` for float projections?

> but you could in principle make other types that did boxing on projection where that wasn't true.

I'm not sure what you mean - I thought projecting out of a float record involves boxing the float?

lpw25: It does. These are two independent properties of the `float` type:

1. Records consisting of only `float` store their fields unboxed and perform boxing on projection.
2. The unboxed form of the type (i.e. `float#`)does not care about uniqueness or linearity.

Currently, the only type with the first property also has the second, but there is no inherent reason for that to be the case. One could easily add an abstract unboxed type that didn’t cross those modes and where records consisting of just the boxed form of that type were stored unboxed and got boxed on projection.

We’re not actually going to add such a type, because our planned support for unboxed types is nicer than these weird magic boxing rules, but the type’s hypothetical existence shows that the two properties are independent and so it is more natural not to assume that 1 implies 2 in the definition of `Texp_field`.

riaqn: Take example `(string * string)#` which is an unboxed tuple of strings. It does not cross uniqueness/linearity because `string` does not. Now if we have a record type
```
type r = {x : (string * string); y : (string * string)}
```
and let's say it will be implemented as a record of two `(string * string)#`. Projecting `r.x` will box the unboxed tuple (described by `alloc_mode`), and we need to check that the mode of `r.x` is less than `alloc_mode`, subject to mode crossing of `(string * string) #`.

Now, the question left is how the UA should treat the newly allocated boxed `(string * string)`. It's not a child of `r`, certainly. Moreover, our UA is not rich enough to express "it shares parts with `r`". Therefore, similar to how we treat `{foo with bar}` where `Kept` fields are marked as used (instead of linked to the new record), here we need to mark `r.x` as used (instead of linked to the boxed value). Therefore, the proper representation is probably:
```
type texp_field_box = 
   | Require_boxing of Alloc.r * unique_use
   | Not_require_boxing of unique_use
```
One might argue to factor out the common `unique_use`, and indeed both represent "the usage of `r.x`, whether as a direct projection, or as an argument to boxing". But on the other hand, they require different treatments in UA, so it's safer to have them under separate constructors leading to different treatments.

riaqn:
OK this is ready for another review now. (please review the last two commits).

PR #2367, opened 2024-03-18
Fix debug printing of simd instructions with one argument

gretay-js:
Fix assertion failure in `Printmach`  when printing some unary `simd` instructions. 
Also adds a test for printing ~and sneaks in a few missing file directives~.

For a proper fix, we should refactor printing to get the number of arguments from `register_behavior` before we add any more simd instructions to avoid similar bugs in the future.


gretay-js:
Looks like file directives don't work with `ocamlformat`.

TheNumbat (approved):

PR #2368, opened 2024-03-18
Add unboxed bigstring primitives to `prim_has_valid_reprs`

alanechang:
Support unboxed bigstring primitives added by https://github.com/ocaml-flambda/flambda-backend/pull/2243 in `prim_has_valid_reprs`

Trying to use these primitives before this PR results in:

```
Error: The primitive [%caml_bigstring_set64u#] is used in an invalid declaration.
       The declaration contains argument/return types with the
       wrong layout.
```

Request review from @ccasin


goldfirere:
@ccasin is there a place where those extra tests are tracked (e.g. an internal ticket)? It would be easy to lose this.

ccasin:
Ah, after writing that I looked through existing tests for similar primitives and concluded this was sufficient.

PR #2369, opened 2024-03-18
Implement `mutable()` logic for record and array

riaqn:
This PR implements the logic for `mutable()` logic for record and array, without supporting the syntax `mutable(modes)`.

- The logic of `mutable` in classes are not changed at all, since everything in OO is strictly constrained to legacy anyway.
- Keep the logic that `mutable` implies `global` modality, but this is now implemented in `typecore.ml` only, instead of desugaring. This simplifies printing which is something we will do soon anyway.
- Because `global` modality implies `shared` and `many` modality, which overrides the `mutable()` logic, so this PR has little observable effects. The only two changes are demonstrated by the tests.

### DIFF
--- a/ocaml/typing/includecore.ml
+++ b/ocaml/typing/includecore.ml
@@ -549,10 +549,10 @@ module Record_diffing = struct
           | Mutable _, Immutable -> Some First
           | Immutable, Mutable _ -> Some Second
           | Mutable m1, Mutable m2 ->
-              if Misc.eq_from_le Mode.Alloc.Const.le m1 m2 then None
+              if Mode.Alloc.Comonadic.Const.eq m1 m2 then None
               else
                 Misc.fatal_errorf "Unexpected mutable(%a) = mutable(%a)"
-                  Mode.Alloc.Const.print m1 Mode.Alloc.Const.print m2
+                  Mode.Alloc.Comonadic.Const.print m1 Mode.Alloc.Comonadic.Const.print m2
         in
         begin match mut with
         | Some mut -> Some (Mutability mut)
--- a/ocaml/typing/includecore.ml
+++ b/ocaml/typing/includecore.ml
@@ -543,11 +543,20 @@ module Record_diffing = struct
   let compare_labels env params1 params2
         (ld1 : Types.label_declaration)
         (ld2 : Types.label_declaration) =
-        if ld1.ld_mutable <> ld2.ld_mutable
-        then
-          let ord = if ld1.ld_mutable = Asttypes.Mutable then First else Second in
-          Some (Mutability ord)
-        else begin
+        let mut =
+          match ld1.ld_mutable, ld2.ld_mutable with
+          | Immutable, Immutable -> None
+          | Mutable _, Immutable -> Some First
+          | Immutable, Mutable _ -> Some Second
+          | Mutable m1, Mutable m2 ->
+              if Misc.eq_from_le Mode.Alloc.Const.le m1 m2 then None
+              else
+                Misc.fatal_errorf "Unexpected mutable(%a) = mutable(%a)"
+                  Mode.Alloc.Const.print m1 Mode.Alloc.Const.print m2
+        in
+        begin match mut with
+        | Some mut -> Some (Mutability mut)
+        | None ->
           match compare_global_flags ld1.ld_global ld2.ld_global with
           | None ->
             let tl1 = params1 @ [ld1.ld_type] in
--- a/ocaml/typing/includecore.ml
+++ b/ocaml/typing/includecore.ml
@@ -549,10 +549,12 @@ module Record_diffing = struct
           | Mutable _, Immutable -> Some First
           | Immutable, Mutable _ -> Some Second
           | Mutable m1, Mutable m2 ->
-              if Mode.Alloc.Comonadic.Const.eq m1 m2 then None
-              else
-                Misc.fatal_errorf "Unexpected mutable(%a) = mutable(%a)"
-                  Mode.Alloc.Comonadic.Const.print m1 Mode.Alloc.Comonadic.Const.print m2
+            let open Mode.Alloc.Comonadic.Const in
+            (if not (eq m1 legacy) then
+              Misc.fatal_errorf "Unexpected mutable(%a)" print m1);
+            (if not (eq m2 legacy) then
+              Misc.fatal_errorf "Unexpected mutable(%a)" print m2);
+            None
         in
         begin match mut with
         | Some mut -> Some (Mutability mut)
--- a/ocaml/typing/mode_intf.mli
+++ b/ocaml/typing/mode_intf.mli
@@ -236,14 +236,10 @@ module type S = sig
         uniqueness : 'c
       }
 
-    module Const : sig
-      include
-        Lattice
-          with type t =
-            (Regionality.Const.t, Linearity.Const.t, Uniqueness.Const.t) modes
-
-      val split : t -> (Monadic.Const.t, Comonadic.Const.t) monadic_comonadic
-    end
+    module Const :
+      Lattice
+        with type t =
+          (Regionality.Const.t, Linearity.Const.t, Uniqueness.Const.t) modes
 
     type error =
       [ `Regionality of Regionality.error
@@ -335,11 +331,18 @@ module type S = sig
     end
 
     module Comonadic : sig
+      module Const : sig
+        include Lattice
+
+        val eq : t -> t -> bool
+      end
+
       include
         Common
           with type error =
             [ `Locality of Locality.error
             | `Linearity of Linearity.error ]
+           and module Const := Const
 
       val meet_with : Const.t -> ('l * 'r) t -> ('l * disallowed) t
     end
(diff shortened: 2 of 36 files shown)


riaqn:
On a second thought - no need to base on #2344 - the benefit is little. Will rebase without it.

riaqn:
This is now rebased and ready for review.

Ekdohibs (approved):
For changes in chamelon.

ocaml/typing/printtyped.ml
```diff
@@ -76,11 +76,21 @@ let fmt_constant f x =
...
 
+let fmt_mutable_mode_flag f (x : Types.mutable_flag) =
+  match x with
+  | Immutable -> fprintf f "Immutable"
+  | Mutable m ->
+    if Misc.eq_from_le Mode.Alloc.Const.le m Mode.Alloc.Const.legacy
+    then fprintf f "Mutable"
+    else
+      Misc.fatal_errorf "Unexpected mutable(%a)"
+        Mode.Alloc.Const.print m
```
lpw25, line 92: I think you should just print the mode rather than erroring here. This is a debug printer anyway, it's probably better if it can print invalid values.

ocaml/typing/printtyp.ml
```diff
@@ -1570,13 +1570,21 @@ let param_jkind ty =
...
   in
-  (Ident.name l.ld_id, gom, tree_of_typexp Type l.ld_type)
+  let mut =
+    match l.ld_mutable with
+    | Immutable -> false
+    | Mutable m ->
+        if Misc.eq_from_le Alloc.Const.le m Alloc.Const.legacy then
+          true
+        else
+          Misc.fatal_errorf "Unexpected mutable(%a)" Alloc.Const.print m
```
lpw25, line 1585: The printers in this file tend to print something like `"<foo>"` rather than erroring for cases that make semantic sense but have no syntax. That's slightly better when we use these printers for debugging the compiler.

ocaml/typing/outcometree.mli
```diff
@@ -87,7 +82,7 @@ type out_type =
   | Otyp_constr of out_ident * out_type list
   | Otyp_manifest of out_type * out_type
   | Otyp_object of { fields: (string * out_type) list; open_row:bool}
-  | Otyp_record of (string * out_mutable_or_global * out_type) list
+  | Otyp_record of (string * bool * out_type * out_global) list
```
lpw25, line 85: How about `Asttypes.mutable_flag` rather than `bool`?

ocaml/typing/mode_intf.mli
```diff
@@ -236,10 +236,14 @@ module type S = sig
...
-      Lattice
-        with type t =
-          (Regionality.Const.t, Linearity.Const.t, Uniqueness.Const.t) modes
+    module Const : sig
+      include
+        Lattice
+          with type t =
+            (Regionality.Const.t, Linearity.Const.t, Uniqueness.Const.t) modes
+
+      val split : t -> (Monadic.Const.t, Comonadic.Const.t) monadic_comonadic
```
lpw25, line 245: Maybe no longer used after the discussed change

ocaml/typing/includecore.ml
```diff
@@ -543,11 +543,20 @@ module Record_diffing = struct
...
+        let mut =
+          match ld1.ld_mutable, ld2.ld_mutable with
+          | Immutable, Immutable -> None
+          | Mutable _, Immutable -> Some First
+          | Immutable, Mutable _ -> Some Second
+          | Mutable m1, Mutable m2 ->
+              if Misc.eq_from_le Mode.Alloc.Const.le m1 m2 then None
+              else
+                Misc.fatal_errorf "Unexpected mutable(%a) = mutable(%a)"
+                  Mode.Alloc.Const.print m1 Mode.Alloc.Const.print m2
```
lpw25, line 555: If you're going to error anyway, it's probably better to just error if either of the modes isn't legacy, rather than letting it succeed in some invalid cases but not others.

ocaml/typing/typecore.ml
```diff
@@ -867,6 +867,62 @@ let apply_mode_annots ~loc ~env (m : Alloc.Const.Option.t) mode =
     | Error (s, e) -> error (s, `Linearity e)
     ) m.linearity
 
+(** Takes [m0] which is the parameter on mutable, and [m] which is a left mode
+    on the record being accessed, returns the left mode of the projected mutable
+    field. *)
+let project_mutable m0 m =
```
lpw25, line 873: Once the monadic case is removed, this function is no longer needed.

ocaml/typing/typecore.ml
```diff
@@ -867,6 +867,62 @@ let apply_mode_annots ~loc ~env (m : Alloc.Const.Option.t) mode =
...
+  | Ok () -> ()
+  | Error _ -> Misc.fatal_error
+      "mutable defaults to Comonadic.legacy, \
+      which is min, so this call cannot fail."
+  );
+  m |> Value.disallow_left
+    (* CR zqian: decouple mutable and global *)
+    |> modality_box_right Global
+
+let _construct_maybe_mutable mut m =
```
lpw25, line 912: Unused function should be removed

ocaml/typing/typecore.ml
```diff
@@ -5495,9 +5548,18 @@ and type_expect_
...
-        List.map (type_label_exp true env argument_mode loc ty_record) lbl_a_list
+      let type_label_exp ((_, label, _) as x) =
+        let argument_mode =
+          match label.lbl_mut with
+          | Immutable -> argument_mode
+          | Mutable m0 ->
+              (* This can't be unboxed record, safe to drop other fields in
+              [argument_mode] *)
+              mode_default (construct_mutable m0 argument_mode.mode)
+        in
```
lpw25, line 5559: Probably worth lifting this out as `mode_field` or something

lpw25 (commented):
Reviewing this PR made us realize this feature is only useful on comonadic axes. Once it's been adjusted to take account of that it should be good to go.

ocaml/typing/typedtree.mli
```diff
@@ -163,7 +163,7 @@ and 'k pattern_desc =
             Invariant: n > 0
          *)
   | Tpat_array :
-      mutable_flag * Jkind.sort * value general_pattern list -> value pattern_desc
```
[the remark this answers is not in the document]
riaqn: I think middle-end or back-end might want to know the `m0` for arrays, or at least it's more correct to think that they do. In that case, it's better to make it available the typedtree, instead of make middle-end know that `m0` is always legacy for arrays.

ocaml/typing/typedecl.ml
```diff
@@ -362,11 +362,14 @@ let transl_labels ~new_var_jkind env univars closed lbls =
     Builtin_attributes.warning_scope attrs
       (fun () ->
          let gbl =
-           match mut with
-           | Mutable -> Mode.Global_flag.Global
-           | Immutable -> Typemode.transl_global_flags
```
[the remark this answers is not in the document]
riaqn: I think you mean that some consumers of the `ld_modalities` field would want to check if it contains `Global`. But I think that's bad behaviour anyway: modalities is just a sequence of functions that should be applied; analysis based on the presence/absence of certain modalities is bound to be unreliable. Similarly, `ld_mutable` is another mapping of modes.

The current approach is to bake the behaviour of `Global` into `Mutable`, that doesn't sound fundamentally wrong.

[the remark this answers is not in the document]
lpw25: `ld_mutable` isn't a mapping it is a constant mode, and only some consumers of `ld_modalities` should also consume `ld_mutable` -- e.g. field projection doesn't need to read it. It is important to put the right information in the right places.

Currently, each consumer of `ld_modalities` is having to do an additional check of `ld_mutable` in order to find out the actual modality of the field. That is needlessly error prone. It also means that the semantics of `ld_modalities` is unclear: it's some subset of the modalities on the field that varies based on the syntax used by the user.

As an example of a misbehaviour caused by this confusion, I think that the following is currently incorrectly rejected with this PR:
```ocaml
module M : sig
  type t = { mutable foo : string }
end = struct
  type t = { mutable global foo : string }
end
```
Obviously we are going to change the meaning of `mutable` later so that this will fail, but I think it is accepted currently and this PR accidentally breaks it.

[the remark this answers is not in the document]
riaqn: Yes good point that `ld_mutable` is better understood as a constant that can be easily compared with each other. To preserve its simplicity, we should avoid adding modality logic to it. I will fix that.

ocaml/typing/typedtree.mli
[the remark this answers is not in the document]
lpw25: Yeah, ok, fair enough

ocaml/typing/typecore.ml
lpw25: I think it is clearer to get rid of them so that it is obvious that projections do not care about the mutability

riaqn: After we introduce syncness/contention, `project_mutable` is good place to check that `m` is uncontended.

PR #2371, opened 2024-03-19
Fix topological iterator in control_flow_graph.ml

Ekdohibs:
Previously, the topological iteration order in `control_flow_graph.ml` was useless, as as soon there was a loop, everything that depended on the loop was recomputed, whether it was in the loop or not.


mshinwell:
@Gbury could you review?

Gbury (approved):

PR #2372, opened 2024-03-19
Stack check as a linear/CFG instruction

xclerc:
This pull request gives an explicit representation
to stack checks on the linear and CFG IRs. The
semantics is unaffected. This change is only a
preliminary step to be able to control where
the check is emitted, rather than always emit it
at the top of a function.

### DIFF
--- a/backend/amd64/emit.mlp
+++ b/backend/amd64/emit.mlp
@@ -1240,7 +1240,7 @@ let emit_simd_instr op i =
     I.pcmpistri (X86_dsl.int n) (arg i 1) (arg i 0); I.set E (res8 i 0); I.movzx (res8 i 0) (res i 0)
 
 (* Emit an instruction *)
-let emit_instr ~first fallthrough i =
+let emit_instr ~first ~fallthrough i =
   emit_debug_info_linear i;
   match i.desc with
   | Lend -> ()
@@ -1836,11 +1836,11 @@ let emit_instr ~first fallthrough i =
       sc_max_frame_size_in_bytes = max_frame_size_bytes;
     }
 
-let rec emit_all first fallthrough i =
+let rec emit_all ~first ~fallthrough i =
   match i.desc with
   | Lend -> ()
   | _ ->
-      emit_instr ~first fallthrough i;
+      emit_instr ~first ~fallthrough i;
       emit_all ~first:false (Linear.has_fallthrough i.desc) i.next
 
 let all_functions = ref []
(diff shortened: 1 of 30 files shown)


backend/amd64/stack_check.ml
```diff
@@ -0,0 +1,57 @@
...
+        + (if fp then 8 else 0))
+      in
+      Misc.align sz 16
+    end else
+      stack_offset + 8
+
+let linear
+  : Linear.fundecl -> Linear.fundecl
+  = fun fundecl ->
+   (* CR mshinwell: this should be conditionalized on a specific
```
mshinwell, line 30: I think this CR is irrelevant now and can be deleted

backend/amd64/stack_check.ml
```diff
@@ -0,0 +1,57 @@
...
+        contains_nontail_calls || max_frame_size >= stack_threshold_size
+      in
+      if insert_stack_check then
+        let fun_body =
+          Linear.instr_cons
+            (Lstackcheck { max_frame_size_bytes = max_frame_size; save_registers = false})
+            [||]
+            [||]
+            ~available_before:None
+            ~available_across:None
```
mshinwell, line 52: Let's try to set these availability values correctly (presumably based on the first instruction of `fundecl.fun_body`?)

backend/amd64/emit.mlp
```diff
@@ -473,6 +463,36 @@ let emit_call_safety_errors () =
...
+}
+
+let stack_realloc = ref (None : stack_realloc option)
+
+let clear_stack_realloc () =
+  stack_realloc := None
+
+let emit_stack_realloc () =
+  begin match !stack_realloc with
+  | None -> ();
```
mshinwell, line 480: Spurious semicolon

backend/amd64/emit.mlp
```diff
@@ -473,6 +463,36 @@ let emit_call_safety_errors () =
     emit_call (Cmm.global_symbol "caml_ml_array_align_error")
   end
 
+(* Stack reallocation *)
+type stack_realloc = {
+  sc_label : Label.t; (* Label of the reallocation code. *)
+  sc_return : Label.t; (* Label to return to after reallocation. *)
+  sc_max_frame_size : int; (* Size for reallocation. *)
```
mshinwell, line 470: maybe put "in bytes" here

mshinwell (changes requested):
Looks good, just minor comments.

backend/cfg/cfg_intf.ml
```diff
@@ -170,6 +170,13 @@ module S = struct
     | Pushtrap of { lbl_handler : Label.t }
     | Poptrap
     | Prologue
+    | Stack_check of
+        { max_frame_size_bytes : int;
+          save_registers : bool
```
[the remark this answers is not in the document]
xclerc: Indeed, `save_registers` is simply "not the first instruction".

mshinwell (approved):

PR #2373, opened 2024-03-19
Move down the stack check if possible

xclerc:
This pull request changes the logic behind the
insertion of the stack check. Instead of always
emitting it at the top of the function, we try to
"push it down" so that it is executed only when
necessary.

In order to determine where to emit the check, we:

- determine the set of CFG block that must only
  be executed after the check;
- for each node of the dominator tree compute
  the number of nodes in the subtree that need
  the check;
- select the block for the stack check as the root
  of the smallest subtree that contains all block
  that must only be executed after the check.

The new beahviour is controlled by a command-line
flag, namely `-cfg-stack-checks`, and disabled by
default.


xclerc:
(As noted offline by @stedolan, it would
make sense to also move the prologue.)

mshinwell (approved):
This is ok to merge once the Specific_can_raise match has been made exhaustive, and maybe "is a non-tail call" moved into Cfg itself, as discussed elsewhere.

PR #2374, opened 2024-03-19
Add missing morphism reductions

riaqn:
This PR adds reduction for the morphisms introduced by #2302 . This was causing #2339 to raise.

In #2302 I said it passed the #2339 test, which was true. However, #2363 makes more use of these morphisms which reveals the lack of some reduction.

### DIFF
--- a/ocaml/typing/mode.ml
+++ b/ocaml/typing/mode.ml
@@ -883,13 +883,24 @@ module Lattices_mono = struct
       type a b c d.
       c obj -> (b, c, d) morph -> (a, b, d) morph -> (a, c, d) morph option =
    fun dst m0 m1 ->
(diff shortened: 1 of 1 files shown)


ocaml/typing/mode.ml
```diff
@@ -883,13 +883,24 @@ module Lattices_mono = struct
       type a b c d.
       c obj -> (b, c, d) morph -> (a, b, d) morph -> (a, c, d) morph option =
    fun dst m0 m1 ->
+    let mid = src dst m0 in
```
lpw25, line 886: I'd be sligtly happier if you pushed this into the `when` clauses.

lpw25 (approved):
LGTM

PR #2376, opened 2024-03-19
probes: add padding after the last semaphore

gretay-js:
Reading/writing a semaphore using ptrace requires accessing 8 bytes of memory, but a semaphore is only 4 bytes. As a result, accessing the last semaphore can fail (e.g., if it's at the end of a page or a segment).  To avoid it, align the end of the ".probes" section to 8 bytes.

### DIFF
--- a/backend/amd64/emit.mlp
+++ b/backend/amd64/emit.mlp
@@ -2309,7 +2309,11 @@ let emit_probe_notes0 () =
       D.word (const 0); (* for systemtap probes *)
       D.word (const (Bool.to_int enabled_at_init)); (* for ocaml probes *)
       add_def_symbol label)
-    !probe_semaphores
+    !probe_semaphores;
+  (* ptrace peek/poke to read/write a semaphores accesses 8 bytes,
+     but the size of the semaphore is only 4 bytes.
+     Add padding to avoid illegal accesses when reading the last semaphore. *)
+  D.align ~data:true 8
 
 let emit_probe_notes () =
   match !probes with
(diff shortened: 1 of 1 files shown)


backend/amd64/emit.mlp
```diff
@@ -2309,7 +2309,11 @@ let emit_probe_notes0 () =
       D.word (const 0); (* for systemtap probes *)
       D.word (const (Bool.to_int enabled_at_init)); (* for ocaml probes *)
       add_def_symbol label)
-    !probe_semaphores
+    !probe_semaphores;
+  (* ptrace peek/poke to read/write a semaphores accesses 8 bytes,
```
TheNumbat, line 2313:
```suggestion
  (* Accessing semaphores with ptrace peek/poke reads/writes 8 bytes,
```

TheNumbat (approved):

TheNumbat:
(Discussed offline, this also needs to add a padding dword after the probes section)

gretay-js:
> (Discussed offline, this also needs to add a padding dword after the probes section)

done https://github.com/ocaml-flambda/flambda-backend/pull/2376/commits/fbd15f4e37bd4981803569a9cf3b7ca381a572a5

gretay-js:
This patch not quite right: a semaphore symbol might have different sizes in different compilation units (depending on whether the symbol is last in the unit or not). The problem will be fixed in ocaml-probes library directly by always using process_vm_read/write to access semaphores.

PR #2379, opened 2024-03-19
float32 backend plumbing/casts

TheNumbat:
Threads `float32` support through cmm / mach / emitters.

Adds cast operations between float, float32, and int. Merges the `floatofint` and `intoffloat` operations into the `scalar_cast` type. Behavior should be preserved.

Adds tests for number conversions, but the bulk of the testing will come in the next PR, which implements float32 arithemtic operations and runtime functions.


backend/cmm.mli
```diff
@@ -165,7 +165,8 @@ type memory_chunk =
   | Thirtytwo_signed
   | Word_int                           (* integer or pointer outside heap *)
   | Word_val                           (* pointer inside heap or encoded int *)
-  | Single
+  | Storage_single                     (* F64 in registers, F32 in memory. *)
+  | Real_single                        (* F32 in both registers and memory. *)
```
gretay-js, line 169: Why do we need a different `memory_chunk` if the memory representation is the same?

TheNumbat: These get propagated all the way to `emit.mlp`, which maps `Storage_single` (same as the previous `Single`) to `cvtss2sd` and `Real_single` to `movss`.

PR #2380, opened 2024-03-19
Basic mixed blocks for `float#` in runtime 5

ncik-roberts:
Support records with `float#` fields in addition to value fields, and represent the unboxed fields as flat at runtime. This PR only adds support for the OCaml 5 runtime, not the OCaml 4 runtime.

Terminology:
  * The record declaration with mixed value and unboxed fields is called a "mixed record".
  * The runtime representation of a record with `float#` and value fields is called a "mixed block".

The native code compiler and bytecode compiler support mixed records. Native code represents mixed record values as mixed blocks; bytecode represents mixed record values as normal blocks.

A mixed record must consist of a **value prefix** (a sequence of fields whose layout is value) followed by a **flat suffix** (a sequence of fields that are either immediates or unboxed floats). This restriction mirrors the mixed block runtime representation, where the 8 top bits of the header are reserved to encode the length of the flat suffix. The length of the flat suffix is read by the GC and other polymorphic traversals, both to detect whether the block is a mixed block and to decide how to recursively descend into the fields.

We use the "reserved header bits" configuration setting to steal 8 bits for the tag.

## What is supported in this PR

* Mixed records where the only non-value fields are `float#`, compiled as mixed blocks in native code.
* Records whose fields are all `float` or `float#`. These are encoded as mixed blocks with scannable prefix 0, and all fields are stored flat.

## What is not supported in this PR (that we plan to support soon)

* Cleanup for reserving header bits.
* Support in the OCaml 4 runtime.
* Inline record arguments to constructors.
* `int64#`, `nativeint#`, and `int32#` fields

## What is not supported in this PR (that we don't have immediate plans to build support for)

* Polymorphic comparison, hash, and marshaling

## How to review

Review directory-by-directory, not commit-by-commit.

  * @ccasin, please review `ocaml/typing/`, `ocaml/utils/`, and `ocaml/lambda/`. I would also appreciate your input on all of the tests, not just the ones relevant to type-checking.
  * @TheNumbat, please review `middle_end/flambda2` and `backend/cmm_helpers.ml`.
  * I will figure out a good person for the runtime and bytecode changes.

Many tests are auto-generated. I check in a small example of the auto-generated test for review.


ncik-roberts:
The tests are failing because you need to pass `--enable-reserved-header-bits=8` to enable mixed block support. The full invocation is:

```
./configure --enable-reserved-header-bits=8 --enable-runtime5 --disable-naked-pointers
```

stedolan:
> Records whose fields are all float or float#. These are encoded as mixed blocks with scannable prefix 0, and all fields are stored flat.

> What is not supported in this PR: Polymorphic comparison, hash, and marshaling

Not supporting polymorphic compare for code using the new unboxed types seems fine. However, not supporting polymorphic compare for records containing `float` (rather than `float#`) is a regression. Does this PR break `Complex.one = Complex.one`?

> The tests are failing because you need to pass --enable-reserved-header-bits=8 to enable mixed block support

If the compiler doesn't work any more without reserved header bits, should we remove this as a configure option?

ncik-roberts:
> Not supporting polymorphic compare for code using the new unboxed types seems fine. However, not supporting polymorphic compare for records containing float (rather than float#) is a regression.

Sorry, my summary was misleading. All-`float` records continue to be treated exactly the same.

This comment from my original summary is wrong:

> Records whose fields are all float or float#. These are encoded as mixed blocks with scannable prefix 0, and all fields are stored flat.

It should instead be about "records whose fields are all float or float#, and at least one field is float#".

ncik-roberts:
> If the compiler doesn't work any more without reserved header bits, should we remove this as a configure option?

That seems like the best option to me. I'm working on this in another PR, and we should merge that at the same time as this.

ocaml/runtime/obj.c
```diff
@@ -161,11 +161,22 @@ CAMLprim value caml_obj_with_tag(value new_tag_v, value arg)
...
        and some of the "values" being copied are actually code pointers.
        That's because the new "value" does not point to the minor heap. */
-    for (i = 0; i < sz; i++) caml_initialize(&Field(res, i), Field(arg, i));
+    for (i = 0; i < scannable_sz; i++) {
+      caml_initialize(&Field(res, i), Field(arg, i));
+    }
+
+    if (scannable_sz < sz) {
+      memcpy(Bp_val(res) + scannable_sz,
+             Bp_val(arg) + scannable_sz,
```
stedolan, line 176: I think those `Bp_val` should be `Op_val`, since `scannable_sz` is in words, not bytes.

ocaml/runtime/minor_gc.c
```diff
@@ -167,6 +167,9 @@ static value alloc_shared(caml_domain_state* d,
                                     reserved);
   d->allocated_words += Whsize_wosize(wosize);
   if (mem == NULL) {
+    // CR nroberts: remove
+    printf("HEADER_RESERVED_BITS: %d\n", HEADER_RESERVED_BITS);
+    printf("wosize: %lu, tag: %u, reserved: %lu\n", wosize, tag, reserved);
```
stedolan, line 172: Leftover debug code.

stedolan:
When playing with this, I got a segfault when trying to print a value using the native toplevel (`ocamlnat`):
```
# type mx = { a : string; b : float# } ;;
type mx = { a : string; b : float#; }
# let x = { a = "a"; b = #1.0 };;
Segmentation fault
```
I think you need to update the `Type_record` case of `tree_of_val` in `ocaml/toplevel/genprintval.ml`.

ocaml/runtime/alloc.c
```diff
@@ -320,6 +332,28 @@ CAMLprim value caml_alloc_dummy_float (value size)
...
+  mlsize_t wosize = Long_val(size);
+  mlsize_t scannable_wosize = Long_val(scannable_size);
+#ifdef NATIVECODE
+  /* The below code runs for bytecode and native code, and critically assumes
+     that a double record field can be stored in one word. That's true both for
+     32-bit and 64-bit bytecode (as a double record field in a mixed record is
+     always boxed), and for 64-bit native code (as the double record field is
+     stored flat, taking up 1 word).
+   */
+  CAMLassert(Double_wosize == 1);
```
stedolan, line 350: `CAML_STATIC_ASSERT`, even?

ocaml/runtime/minor_gc.c
```diff
@@ -315,6 +318,16 @@ static void oldify_one (void* st_v, value v, volatile value *p)
         st->todo_list = v;
       } else {
         CAMLassert (sz == 1);
```
stedolan, line 320: The logic here took me a while to figure out. I don't think it's wrong, but I think it might be simpler to copy the non-scannable fields directly in oldify_one, as is currently done for entirely non-scannable blocks (where `tag >= No_scan-tag`).

That is, I think it would be simpler to directly copy the fields in `[scannable_sz, sz)`, and do the `sz > 1` tail-call test on `scannable_sz` instead (handling the `0` case with a return as you do here). That way, I think you can delete the new loop in `oldify_mopup`.

stedolan (changes requested):
Looks good! I only read the runtime bits.

ocaml/typing/typedecl.ml
```diff
@@ -1113,40 +1171,119 @@ let update_decl_jkind env dpath decl =
...
+                  | Float_element -> Float
+                  | Flat_float64_element -> Float64
+                  | Element_without_runtime_component -> Imm
+                  | Flat_imm_element | Value_element ->
+                      Misc.fatal_error "Expected only floats and float64s")
+                classifications
+              |> Array.of_list
+            in
+            assert_mixed_record_support loc ~value_prefix_len:0;
+            Record_mixed { value_prefix_len = 0; flat_suffix }
```
Ekdohibs, line 1221: Why do we use a mixed record there instead of a flat float record with `tag = 252`?

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
```diff
@@ -1395,6 +1429,21 @@ let convert_lprim ~big_endian (prim : L.primitive) (args : Simple.t list list)
     [ Ternary
         (Block_set (block_access, init_or_assign), block, Simple field, value)
     ]
+  | ( Psetmixedfield (field, shape, initialization_or_assignment),
+      [[block]; [value]] ) ->
+    let imm = Targetint_31_63.of_int field in
+    check_non_negative_imm imm "Psetufloatfield";
```
TheNumbat, line 1435:
```suggestion
    check_non_negative_imm imm "Psetmixedfield";
```

middle_end/flambda2/simplify/simplify_variadic_primitive.ml
```diff
@@ -56,19 +107,28 @@ let simplify_make_block ~original_prim ~field_kind tag ~shape
...
-        T.immutable_block ~is_unique:true tag ~field_kind alloc_mode ~fields
-      | Mutable -> T.mutable_block alloc_mode
+      match block_shape with
+      | Mixed _ ->
+        (* CR mixed blocks: I don't really know what I'm doing here. In
+           particular for [ty] I've picked [T.any_block] because exisiting more
+           specific types for blocks all want the fields to have the same kind.
+           I don't even know of [any_block] is correct -- perhaps it's meant to
+           be normal blocks below no scan tag. *)
+        T.any_block
```
TheNumbat, line 117: Going by the following code, looks like you can construct the same fields list and use the match on mutable_or_immutable for both cases?

middle_end/flambda2/terms/flambda_primitive.mli
```diff
@@ -101,6 +115,7 @@ module Duplicate_block_kind : sig
           length : Targetint_31_63.t
         }
     | Naked_floats of { length : Targetint_31_63.t }
+    | Mixed
```
TheNumbat, line 118: Either add the tag/shape here or note that the fields are not necessary because they're only used for printing

middle_end/flambda2/to_cmm/to_cmm_primitive.ml
```diff
@@ -100,6 +100,12 @@ let block_load ~dbg (kind : P.Block_access_kind.t) (mutability : Mutability.t)
   | Values { field_kind = Immediate; _ } ->
     C.get_field_computed Immediate mutability ~block ~index dbg
   | Naked_floats _ -> C.unboxed_float_array_ref block index dbg
+  | Mixed { field_kind; _ } -> (
+    (* CR mixed blocks: a backend person should confirm I can just use these and
```
TheNumbat, line 104: This and the below CR should be fine, at least until we allow packed int32#s, etc.

Note `unboxed_float_array_ref` always generates a mutable load, which is overly conservative. (It doesn't matter much)

TheNumbat (commented):
I read `backend/` and `middle_end/flambda2`; looks good.
I'm not sure about the CRs in `middle_end/flambda2/simplify`, so an flambda2 author should also take a look.

middle_end/flambda2/simplify/simplify_variadic_primitive.ml
lthls: Yes, `T.any_block` is meant for blocks with regular constructors. Until we have support for mixed blocks in the Flambda2 type system, you should use `T.any_value`.

ocaml/typing/typedecl.ml
ncik-roberts: First, why even support this case at all? The thinking is that this gives users a way of gradually migrating their all-float records to all-unboxed-floats without sacrificing the flat float optimization at any point.

Next, what tag to use?
  * Downside of current approach: the user loses polymorphic operations. But this is fine for Jane Street code.
  * Benefit of current approach: this could allow us to store boxed floats as flat even in inline record arguments to constructors. But actually I don't plan on supporting this: The argument for flat floats in mixed records doesn’t apply, as no existing code can possibly rely on the flat float optimization in constructor arguments.

So it does seem a bit better to use `Double_array_tag` here. The main reason I didn't do that in this PR is that @ccasin tells me it was hard to teach the middle-end how to do projections correctly here (some require boxing and others don't), and would require a small refactor.

But I plan on looking into this (in another PR) and suggest going with this approach for now.

ncik-roberts:
> When playing with this, I got a segfault when trying to print a value using the native toplevel (`ocamlnat`):
> 
> ```
> # type mx = { a : string; b : float# } ;;
> type mx = { a : string; b : float#; }
> # let x = { a = "a"; b = #1.0 };;
> Segmentation fault
> ```
> 
> I think you need to update the `Type_record` case of `tree_of_val` in `ocaml/toplevel/genprintval.ml`.

@stedolan, fixing this (now-fixed) bug revealed to me another bug: `caml_hash` now can raise, and so external declarations thereof shouldn't be annotated as `[@@noalloc]`. I'm following up internally about alternative approaches for hashing.

ocaml/runtime/obj.c
ncik-roberts: Indeed.

ocaml/runtime/compare.c
```diff
@@ -207,6 +207,11 @@ static intnat do_compare_val(struct compare_stack* stk,
           if (t1 != t2)
               return (intnat)t1 - (intnat)t2;
       }
+      if ( Is_mixed_block_reserved(Reserved_val(v1))
+        || Is_mixed_block_reserved(Reserved_val(v2))) {
+        compare_free_stack(stk);
+        caml_invalid_argument("compare: mixed block value");
+      }
```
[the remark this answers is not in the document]
ncik-roberts: This mirrors what we do for other non-comparable things, like abstract blocks and functions. It's admittedly strange but it's consistent.

ccasin (commented):
I've read typing, lambda, and the tests for typing (but not the runtime tests).

This looks good!  Mainly minor comments/suggestions.

ocaml/typing/typedecl.ml
```diff
@@ -1,4 +1,4 @@
-(**************************************************************************)
+(******************************u*******************************************)
```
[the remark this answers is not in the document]
TheNumbat: no u

ocaml/runtime/minor_gc.c
ncik-roberts: Done! I like this structure more, too.

backend/cmm_helpers.ml
```diff
@@ -1354,33 +1444,61 @@ let make_alloc_generic ~mode set_fn dbg tag wordsize args =
...
+  let set_fn idx arr ofs newval dbg =
+    if idx < value_prefix_len
+    then addr_array_init arr ofs newval dbg
+    else
+      match flat_suffix.(idx - value_prefix_len) with
+      | Imm -> int_array_set arr ofs newval dbg
+      | Float | Float64 -> float_array_set arr ofs newval dbg
+  in
+  let size =
+    let values, floats = Lambda.count_mixed_block_values_and_floats shape in
```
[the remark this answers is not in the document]
ncik-roberts: Yes, immediates are counted in the value section.

I actually don't think that we should call this scannable/non-scannable. The difference we're getting at here is the difference in sizes between values/floats, not whether or not the GC scans them.

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
ncik-roberts: Accepted (manually).

middle_end/flambda2/simplify/simplify_variadic_primitive.ml
ncik-roberts: @TheNumbat: I'm not sure I understand your suggestion. I'm taking Vincent's suggestion to use `T.any_value` here; let me know if it seems like I'm missing something.

lthls: By the way, do you have a dedicated primitive for projecting regular values out of mixed blocks or do you use the existing `Pfield` primitive ? If you have a dedicated primitive then we can postpone any changes to the Flambda2 type system, if you reuse the existing primitive we will likely have to do at least some quick fixes to avoid wrongly confusing a mixed block and a regular block.

ocaml/testsuite/tests/translprim/comparison_table.heap.reference
```diff
@@ -114,11 +114,7 @@
          ignore assert all zero_alloc : int (compare_ints prim prim))
      eta_float_cmp =
        (function {nlocal = 0} prim[float] prim[float] stub
-<<<<<<< HEAD
-         ignore assert all zero_alloc : int (compare_floats float prim prim))
-=======
          ignore assert all zero_alloc : int (compare_floats prim prim))
->>>>>>> main
```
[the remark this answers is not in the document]
ncik-roberts: Marking resolved as the change I made (and have now reverted) is irrelevant to this PR.

ocaml/typing/typeopt.ml
```diff
@@ -586,6 +595,9 @@ and value_kind_record env ~loc ~visited ~depth ~num_nodes_visited
             [0, fields]
           | Record_inlined (Extension _, _) ->
             [0, fields]
+          | Record_mixed _ ->
+            (* CR mixed blocks v1: Tag should not be 0. *)
```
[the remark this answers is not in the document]
ncik-roberts: The comment `CR mixed blocks v1` indicates that we'll need to make a change before v1. V1 will include non-tag-0 mixed blocks.

ocaml/typing/typedecl.ml
```diff
@@ -3169,9 +3306,39 @@ let report_error ppf = function
...
+      match error with
+      | Flat_field_expected { boxed_lbl; non_value_lbl } ->
+          fprintf ppf
+            "@[Expected all flat fields after non-value field, %s,@]@,\
+            \ @[but found boxed field, %s.@]"
+            (Ident.name non_value_lbl)
+            (Ident.name boxed_lbl)
+      | Runtime_support_not_enabled ->
+          fprintf ppf
+            "@[This OCaml runtime doesn't support mixed records. Contact\
```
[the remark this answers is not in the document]
ncik-roberts: I added a test for this now.

ocaml/typing/typedecl.ml
```diff
@@ -1113,40 +1171,119 @@ let update_decl_jkind env dpath decl =
...
+      List.iter
+        (fun (_lbl, repr) ->
+           match repr with
+           | Float_element -> element_reprs.floats <- true
+           | Flat_imm_element -> element_reprs.imms <- true
+           | Flat_float64_element -> element_reprs.float64s <- true
+           | Value_element -> element_reprs.values <- true
+           | Element_without_runtime_component -> ())
+        classifications;
       let rep =
```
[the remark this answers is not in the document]
ncik-roberts: I did the naive refactor (pull out the function and turn its free variables into parameters) and I don't think it's more readable. I did a smaller refactor (just pull out `find_flat_suffix`) and think that's much more readable. So that's what I did. Thanks for the suggestion.

middle_end/flambda2/simplify/simplify_variadic_primitive.ml
ncik-roberts: Per discussion, I'm going to change this PR to use a dedicated primitive for projecting a regular value out of a mixed block.

ncik-roberts: I now use the mixed-block version of the primitive for getting/setting fields in the value prefix as well.

ocaml/testsuite/tests/typing-layouts-float64/unboxed_floats_alpha.ml
```diff
@@ -0,0 +1,214 @@
...
+  let ( ** ) = pow
+  let ( > ) x y = (compare x y) > 0
+end
+
+let print_floatu prefix x = Printf.printf "%s: %.2f\n" prefix (Float_u.to_float x)
+let print_float prefix x = Printf.printf "%s: %.2f\n" prefix x
+let print_int prefix x = Printf.printf "%s: %d\n" prefix x
+
+(***********************************************)
+(* Test 13: basic (float# + immediate) records *)
```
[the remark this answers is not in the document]
ncik-roberts: I split out two sets of tests:
  * Some tests under `ocaml/testsuite/tests/typing-layouts-float64` of all-float-and-float# mixed records
  * Some tests under `ocaml/testsuite/tests/mixed-blocks/test_mixed_blocks.ml` that test normal mixed records. Along your suggestion, I included a `string` field to more plainly suggest that the prior `float` fields are flat.

ocaml/testsuite/tests/typing-layouts-float64/basics.ml
```diff
@@ -195,31 +193,50 @@ Error: This type ('b : value) should be an instance of type ('a : float64)
          it's the type of a tuple element.
 |}]
 
-(******************************************************************************)
-(* Test 5: Can't be put in structures in typedecls, except all-float records. *)
+(****************************************************************************)
+(* Test 5: Can't be put in structures in typedecls, except certain records. *)
 
+(* all-float64 records are allowed, as are any combination of immediate, float,
+   and float64 *)
```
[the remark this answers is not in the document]
ncik-roberts: Thanks, this was a dangling comment from an earlier version of this PR. It's now corrected.

ocaml/testsuite/tests/typing-layouts-float64/basics_alpha.ml
```diff
@@ -0,0 +1,904 @@
...
+|}];;
+
+type ('a : float64) t5_13 = {x : 'a; y : float#};;
+[%%expect{|
+type ('a : float64) t5_13 = { x : 'a; y : float#; }
+|}];;
+
+type 'a t5_14 = {x : 'a; y : float#};;
+[%%expect{|
+type 'a t5_14 = { x : 'a; y : float#; }
```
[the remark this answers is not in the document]
ncik-roberts: I've left this test in `typing-layouts-float64`, and added the other suggested tests in `typing-layouts/mixed_records.ml` (which I moved from `mixed-blocks/mixed_records.ml` and added a few more tests to).

ocaml/lambda/translcore.ml
```diff
@@ -568,6 +568,26 @@ and transl_exp0 ~in_new_scope ~scopes sort e =
...
+          if lbl.lbl_num < value_prefix_len then
+            Lprim (Pfield (lbl.lbl_pos, maybe_pointer e, sem), [targ],
+                   of_location ~scopes e.exp_loc)
+          else
+            let flat_projection =
+              match flat_suffix.(lbl.lbl_num - value_prefix_len) with
+              | Imm -> Projection_imm
+              | Float64 -> Projection_float64
+              | Float ->
+                 (match float with
```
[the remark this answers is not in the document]
ncik-roberts: Yeah, this was buggy and accidentally omitted from the tests. `typecore.ml` now supports it (and the auto-generated tests now exercise this too).

ocaml/toplevel/genprintval.ml
```diff
@@ -519,17 +540,27 @@ module Make(O : OBJ)(EVP : EVALPATH with type valu = O.t) = struct
...
-                    if O.tag obj = O.double_array_tag then
-                      O.repr (O.double_field obj pos)
-                    else
-                      O.field obj pos
-                  in
-                  nest tree_of_val (depth - 1) fld ty_arg
-                end
+                else match rep with
+                  | Outval_record_boxed -> tree_of_val (depth - 1) obj ty_arg
+                  | Outval_record_unboxed ->
```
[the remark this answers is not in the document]
ncik-roberts: Indeed, that was segfaulting in a bunch of tests and is now fixed.

ocaml/lambda/lambda.mli
```diff
@@ -744,6 +761,15 @@ val transl_value_path: scoped_location -> Env.t -> Path.t -> lambda
...
 val transl_class_path: scoped_location -> Env.t -> Path.t -> lambda
 
+val transl_mixed_record_shape: Types.mixed_record_shape -> mixed_block_shape
+val count_mixed_block_values_and_floats : mixed_block_shape -> int * int
+
+type mixed_block_element =
+  | Value_prefix
+  | Flat_suffix of flat_element
+
+val get_mixed_block_element : mixed_block_shape -> int -> mixed_block_element
```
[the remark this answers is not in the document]
ncik-roberts: Added a comment.

ocaml/typing/types.mli
```diff
@@ -821,6 +836,14 @@ val bound_value_identifiers: signature -> Ident.t list
 
 val signature_item_id : signature_item -> Ident.t
 
+val count_mixed_record_values_and_floats : mixed_record_shape -> int * int
+
+type mixed_record_element =
+  | Value_prefix
+  | Flat_suffix of flat_element
+
+val get_mixed_record_element : mixed_record_shape -> int -> mixed_record_element
```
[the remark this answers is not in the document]
ncik-roberts: accepted

ocaml/typing/typedecl.ml
```diff
@@ -1113,40 +1171,119 @@ let update_decl_jkind env dpath decl =
...
+              |> Array.of_list
+            in
+            assert_mixed_record_support loc ~value_prefix_len:0;
+            Record_mixed { value_prefix_len = 0; flat_suffix }
+        | { values = true ; imms = _    ; floats = _    ; float64s = true  }
+        | { values = _    ; imms = true ; floats = _    ; float64s = true  } ->
+            let rec find_flat_suffix classifications =
+              match classifications with
+              | [] -> Misc.fatal_error "Expected at least one Float64"
+              | (non_value_ld, c) :: classifications ->
```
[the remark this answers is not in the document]
ncik-roberts: Indeed. Consider it `ld`.

ocaml/typing/typedecl.ml
```diff
@@ -1097,9 +1133,31 @@ let update_constructor_arguments_jkinds env loc cd_args jkinds =
...
+       if [imms] is [true].)
+    *)
+    type element_reprs =
+      {  mutable values : bool; (* also includes [imm64] *)
+         mutable imms : bool;
+         mutable floats: bool;
+         mutable float64s : bool }
+
+    type element_repr =
+      | Flat_imm_element
```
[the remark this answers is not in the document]
ncik-roberts: Indeed, `flat` is redundant. I've renamed this to `Imm_element`.

ocaml/testsuite/tests/typing-layouts-float64/basics_alpha.ml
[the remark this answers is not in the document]
ncik-roberts: I ended up moving more of these tests into `mixed_records.ml`, which now has a twin `mixed_records_alpha.ml`. I minted new files instead of reusing `basics.ml` because the output for some tests is different in runtime 4. (To sidestep this, `mixed_records*.ml` is currently just run for runtime 5.)

ocaml/runtime/minor_gc.c
```diff
@@ -308,13 +308,20 @@ static void oldify_one (void* st_v, value v, volatile value *p)
     st->live_bytes += Bhsize_hd(hd);
     result = alloc_shared(st->domain, sz, tag, Reserved_hd(hd));
     field0 = Field(v, 0);
+    mlsize_t scannable_sz = Scannable_wosize_hd(hd);
+    /* Copy the non-scannable suffix of fields. */
+    for (i = scannable_sz; i < sz; i++) {
+      Field(result, i) = Field(v, i);
+    }
```
stedolan, line 315: For multicore performance, this loop should be below the `if (try_update_object_header ...)`.

With multiple domains, if a block is reachable from multiple domains during minor GC then both will race to promote it, and `try_update_object_header` will return true only on the domain that wins the race. So, moving the loop down helps in two ways:

  - you are more likely to win the race if you don't stop to do a memcpy before crossing the finish line
  - if you lose the race, you don't need to do the copy at all

ocaml/runtime/minor_gc.c
```diff
@@ -411,13 +421,17 @@ static void oldify_mopup (struct oldify_state* st, int do_ephemerons)
     new_v = Field(v, 0);                 /* Follow forward pointer. */
     st->todo_list = Field (new_v, 1);    /* Remove from list. */
 
-    f = Field(new_v, 0);
-    CAMLassert (!Is_debug_tag(f));
-    if (Is_block (f) && Is_young(f)) {
-      oldify_one (st, f, Op_val (new_v));
+    mlsize_t scannable_wosize = Scannable_wosize_val(new_v);
+
+    if (scannable_wosize != 0) {
```
stedolan, line 426: `v` comes from `todo_list`, and blocks are now added to `todo_list` only if `scannable_wosize > 1`, so I think this condition is redundant.

stedolan (changes requested):
Runtime stuff looks good! Two more small comments for `minor_gc.c`, but the rest lgtm.

ocaml/runtime/minor_gc.c
ncik-roberts: I accepted this suggestion.

middle_end/flambda2/simplify/simplify_variadic_primitive.ml
```diff
@@ -108,13 +108,7 @@ let simplify_make_block ~original_prim tag ~(block_shape : Block_shape.t)
...
     let ty =
       match block_shape with
-      | Mixed _ ->
-        (* CR mixed blocks: I don't really know what I'm doing here. In
-           particular for [ty] I've picked [T.any_block] because exisiting more
-           specific types for blocks all want the fields to have the same kind.
-           I don't even know of [any_block] is correct -- perhaps it's meant to
-           be normal blocks below no scan tag. *)
-        T.any_block
+      | Mixed _ -> T.any_block
```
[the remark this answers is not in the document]
ncik-roberts: I misunderstood @lthls 's earlier comment. I think this should be `any_value`, so I'll change that.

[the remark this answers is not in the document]
lthls: Yes, I confirm that `any_value` is the expected type for mixed block for the moment.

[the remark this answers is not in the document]
ncik-roberts: None of the tests were failing with the old code. Any suggestions for a test that could show the issue?

ocaml/runtime/minor_gc.c
ncik-roberts: Indeed. I changed this to a `CAMLassert`. The interplay between `oldify_one` and `oldify_mopup` has confused me once again so I tried to write an explanatory comment on the assert.

ocaml/runtime/minor_gc.c
ncik-roberts: (My naive attempt to move the `memcpy` down introduced a bug, because I forgot that `try_update_object_header` clobbers the 0th field of `v`. I've just pushed a fix.)

ocaml/typing/rec_check.ml
```diff
@@ -646,6 +646,7 @@ let rec expression : Typedtree.expression -> term_judg =
         let field_mode = match rep with
           | Record_float | Record_ufloat -> Dereference
           | Record_unboxed | Record_inlined (_,Variant_unboxed) -> Return
+          | Record_mixed _
```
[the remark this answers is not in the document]
ccasin: For posterity: Nick ended up pushing a more fine-grained version of this that treats the flat suffix as `Dereference` and the value prefix as `Guard`.  I've reviewed this.

ccasin (approved):
Looks good now, thanks!

middle_end/flambda2/simplify/simplify_variadic_primitive.ml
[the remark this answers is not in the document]
ncik-roberts: (resolving this for now as I'm not blocked on such a test.)

ncik-roberts:
@stedolan I believe this is waiting for your approval as you requested changes earlier. Also, the latest commit includes one runtime change: the implementation of polymorphic hash for mixed blocks. Mind looking at that (and also at whether the other conversations are resolved to your liking)?

ocaml/runtime/hash.c
```diff
@@ -282,10 +283,15 @@ CAMLprim value caml_hash(value count, value limit, value seed, value obj)
...
         h = caml_hash_mix_uint32(h, Cleanhd_hd(Hd_val(v)));
         /* Copy fields into queue, not exceeding the total size [sz] */
-        for (i = 0, len = Wosize_val(v); i < len; i++) {
+        for (i = 0, len = Scannable_wosize_val(v); i < len; i++) {
           if (wr >= sz) break;
           queue[wr++] = Field(v, i);
         }
+        /* Mix the flat suffix */
+        for (i = Scannable_wosize_val(v), len = Wosize_val(v); i < len; i++) {
+          h = caml_hash_mix_intnat(h, Field(v, i));
```
xclerc, line 292: (I don't think float values should be mixed
like that, `caml_hash_mix_double` is
performing some kind of normalization.)

ncik-roberts: In this PR, we could call `caml_hash_mix_double`. But in future PRs, there will be flat elements that are, say, `int64#` (or two `float32#`s packed together, or 4 `float16#s`, or a `float32#` followed by 2 `float16#s`). I'm worried there's not going to be a way to normalize these correctly.

* If we just normalize -0 to 0, then I think we could do that fairly easily even with arbitrary mixing of small floats.
* But I'm worried about NaN normalization. (Is that even needed? Two nans are never equal, so it's vacuously true that `nan1 = nan2` implies `hash(nan1) = hash(nan2)`.

xclerc: Sure, I think it is just safe to ignore that part of the block.

ncik-roberts: After some discussion:
  * We actually don't need to do any normalization. That's because mixed blocks can never "compare as equal" &mdash; polymorphic compare raises &mdash; so there's no risk of defying the hash invariant.
  * But this is all very tricky. I'm just going to return a constant for now on mixed blocks and revisit this later.

ncik-roberts: I'd like to change this to raise before moving mixed blocks out of `layouts_alpha.` But I'd like to do this in a separate PR to more easily assess performance impact of making `caml_hash` to not be `[@@noalloc]`. @stedolan would you mind reviewing the remaining runtime changes and approving if you think they look good (modulo the hash matter)?

stedolan (approved):

ocaml/runtime/minor_gc.c
stedolan: Well spotted on the bug, I hadn't thought of that!

PR #2382, opened 2024-03-20
Fix bug (infinite loop) in dominance frontier

xclerc:
This pull request fixes a bug in the computation
of the dominance frontier, where the current
code can loop for ever.

Like in https://github.com/ocaml-flambda/flambda-backend/pull/2128, the bug is related to dead code:
when computing the dominance frontier, we
repeatedly move a _runner_ from a block's
predecessor to its immediate dominator until
we reach the dominator of the starting block.

The problem is that the predecessor may in
some very rare cases not be in the same
"component" (See https://github.com/ocaml-flambda/flambda-backend/pull/2128) as the starting block,
which leads to an infinite loop. The fix is to
also stop when we reach a block which is its
own immediate predecessor (_i.e._ a dead block
with no predecessors).

### DIFF
--- a/backend/cfg/cfg_dominators.ml
+++ b/backend/cfg/cfg_dominators.ml
@@ -309,7 +309,10 @@ let invariant_dominance_frontiers : Cfg.t -> doms -> dominance_frontiers -> unit
         frontier_labels)
     dominance_frontiers
 
-(* See Figure 5 in the cited article. *)
+(* See Figure 5 in the cited article, the only difference is that because of
+   dead code, there is a second condition to the while loop: the loop stops if
+   it reaches a block which is its own immediate dominator (this means we have
+   reached a "dead" entry point). *)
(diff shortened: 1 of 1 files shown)


backend/cfg/cfg_dominators.ml
```diff
@@ -325,14 +328,21 @@ let compute_dominance_frontiers (cfg : Cfg.t) (doms : doms) :
       | _ :: _ :: _ ->
         List.iter predecessor_labels ~f:(fun predecessor_label ->
             let runner = ref predecessor_label in
-            while not (Label.equal !runner (Label.Tbl.find doms label)) do
+            let continue = ref true in
+            while
+              (not (Label.equal !runner (Label.Tbl.find doms label)))
```
gretay-js, line 333: Why are we not using `_idom` instead of `Label.Tbl.find doms label` here?  (I know it's not related to the fix in this PR.)

gretay-js (approved):

PR #2383, opened 2024-03-20
Mode error messages should use the solver error

riaqn:
The mode solver returns `{left; right}` if `submode a b` fails, where the `left` is lowest mode `a` can be, and `right` is the highest mode `b` can be.  And because the submoding fails, we are guaranteed that `left` is not less or equal than `right`. This should be used to print mode error messages.

Currently mode error messages discards that information, and proceed with the assumption that there are only two levels in each axis.

### DIFF
--- a/ocaml/typing/typecore.ml
+++ b/ocaml/typing/typecore.ml
@@ -10007,24 +10007,22 @@ let report_error ~loc env = function
         "@[This application is complete, but surplus arguments were provided afterwards.@ \
          When passing or calling a local value, extra arguments are passed in a separate application.@]"
   | Param_mode_mismatch (s, mkind) ->
-      let error_of_equate f (step, {Solver.left; Solver.right}) =
-        let left, right =
+      let print_error f (step, {Solver.left; Solver.right}) =
+        let actual, expected =
           match (step : equate_step) with
           | Left_le_right -> left, right
           | Right_le_left -> right, left
         in
-        (fun ppf -> f ppf left), (fun ppf -> f ppf right)
-      in
-      let left, right =
-        match mkind with
-        | `Locality e -> error_of_equate Locality.Const.print (s, e)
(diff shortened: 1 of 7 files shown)


ocaml/typing/typecore.ml
```diff
@@ -10004,25 +10006,34 @@ let report_error ~loc env = function
...
+        | `Locality e -> error_of_equate Locality.Const.print (s, e)
+        | `Uniqueness e -> error_of_equate Uniqueness.Const.print (s, e)
+        | `Linearity e -> error_of_equate Linearity.Const.print (s, e)
       in
       Location.errorf ~loc
-        "@[This function has a %s parameter, but was expected to have type:@ %a@]"
-        mkind Printtyp.type_expr ty
+        "@[This function takes a %t parameter,@ \
+         but was expected to take a %t parameter.@]"
+        left right
```
ncik-roberts, line 10027: Suggestion: run this code in `error_of_equate` (perhaps also renaming `error_of_equate`) so you don't need to capture `left`/`right` in closures just to hide their type.

ocaml/typing/typecore.ml
```diff
@@ -10004,25 +10006,34 @@ let report_error ~loc env = function
...
-  | Param_mode_mismatch (ty, (_, mkind)) ->
-      let mkind =
+  | Param_mode_mismatch (s, mkind) ->
+      let error_of_equate f (step, {Solver.left; Solver.right}) =
+        let left, right =
+          match (step : equate_step) with
+          | Left_le_right -> left, right
+          | Right_le_left -> right, left
+        in
+        (fun ppf -> f ppf left), (fun ppf -> f ppf right)
```
ncik-roberts, line 10016: It's a bit weird that `left` and `right` can switch names here. Shouldn't they be called `expected` and `actual` after you match on `step`? That's how they're used in the error message.

ncik-roberts (approved):
The new error message indeed looks like an improvement.

PR #2385, opened 2024-03-21
float32 backend operations

TheNumbat:
Implements float32 arithmetic operations in the backend and runtime.
- Implements float32 operations in cmm
- Adds a `float_width` parameter to the float arithmetic instructions and threads this through mach/cfg/etc
- Adds float32 instructions to the x86 emitter / ast / dsl / binary emitter
- Implements float32 custom_ops in the runtime + arithmetic operations for bytecode
- Adds tests for float32 arithmetic, comparisons, hash, marshalling


ocaml/runtime/float32.c
```diff
@@ -0,0 +1,158 @@
...
+
+static int float32_cmp(value v1, value v2)
+{
+  return caml_float32_compare_unboxed(Float32_val(v1), Float32_val(v2));
+}
+
+static intnat float32_hash(value v)
+{
+  float f = Float32_val(v);
+  return float32_as_int32(f);
```
xclerc, line 57: That does not feel right; for instance,
it means that `-0.0` and `+0.0` will
have different hash values.

TheNumbat: Updated to be consistent with `caml_hash_mix_float`

backend/cfg/cfg_equivalence.ml
```diff
@@ -442,8 +445,9 @@ let check_terminator_instruction :
       Truth_test { ifso = ifso2; ifnot = ifnot2 } ) ->
     State.add_to_explore state ifso1 ifso2;
     State.add_to_explore state ifnot1 ifnot2
-  | ( Float_test { lt = lt1; eq = eq1; gt = gt1; uo = uo1 },
-      Float_test { lt = lt2; eq = eq2; gt = gt2; uo = uo2 } ) ->
+  | ( Float_test { width = w1; lt = lt1; eq = eq1; gt = gt1; uo = uo1 },
+      Float_test { width = w2; lt = lt2; eq = eq2; gt = gt2; uo = uo2 } )
+    when Stdlib.compare w1 w2 = 0 ->
```
xclerc, line 450:
```suggestion
    when Cmm.equal_float_width w1 w2 ->
```

backend/regalloc/regalloc_validate.ml
```diff
@@ -539,8 +539,9 @@ end = struct
         Truth_test { ifso = ifso2; ifnot = ifnot2 } ) ->
       compare_label ifso1 ifso2;
       compare_label ifnot1 ifnot2
-    | ( Float_test { lt = lt1; eq = eq1; gt = gt1; uo = uo1 },
-        Float_test { lt = lt2; eq = eq2; gt = gt2; uo = uo2 } ) ->
+    | ( Float_test { width = w1; lt = lt1; eq = eq1; gt = gt1; uo = uo1 },
+        Float_test { width = w2; lt = lt2; eq = eq2; gt = gt2; uo = uo2 } )
+      when Stdlib.compare w1 w2 = 0 ->
```
xclerc, line 544:
```suggestion
      when Cmm.equal_float_width w1 w2 ->
```

backend/amd64/emit.mlp
```diff
@@ -587,47 +596,87 @@ let emit_float_test cmp i ~(taken:X86_ast.condition -> unit) =
...
-  | CFge ->
+  | Float64, CFge ->
       I.comisd (arg i 1) (arg i 0);(* swap compare *)
       taken AE                     (* branch taken if x>=y *)
-  | CFnge ->
+  | Float64, CFnge ->
       I.comisd (arg i 1) (arg i 0);(* swap compare *)
       taken B                      (* taken if unordered or x<y i.e. !(x>=y) *)
+  | Float32, CFeq when arg i 1 = arg i 0 ->
+      I.ucomisd (arg i 1) (arg i 0);
```
xclerc, line 641: Shouldn't that be `I.ucomiss`?

backend/amd64/emit.mlp
```diff
@@ -587,47 +596,87 @@ let emit_float_test cmp i ~(taken:X86_ast.condition -> unit) =
...
+      taken NP
+  | Float32, CFeq ->
+      let next = new_label() in
+      I.ucomisd (arg i 1) (arg i 0);
```
xclerc, line 645: Shouldn't that be `I.ucomiss`?

backend/amd64/emit.mlp
```diff
@@ -587,47 +596,87 @@ let emit_float_test cmp i ~(taken:X86_ast.condition -> unit) =
...
+      I.jp (label next);           (* skip if unordered *)
+      taken E;                     (* branch taken if x=y *)
+      def_label next
+  | Float32, CFneq when arg i 1 = arg i 0 ->
```
xclerc, line 649: It looks like a copy/paste from the
`Float64` cases, only changing the
`I.ucomisd` call to `I.ucomiss`. If it
is the case, it might be worth
factoring out.

tests/small_numbers/float32.ml
```diff
@@ -0,0 +1,267 @@
...
+      assert (h = Poly.hash bits)
+  )
+;;
+
+let () = (* Marshal *)
+  CFloat32.check_float32s (fun l r ->
+    let b = Poly.to_bytes (Sys.opaque_identity l, Sys.opaque_identity r) [] in
+    let (readl, readr) : float32 * float32 = Poly.from_bytes_unsafe b 0 in
+    assert (Float32.compare l readl = 0);
+    assert (Float32.compare r readr = 0)
```
xclerc, line 265: (I wonder whether this should use the bit patterns.)

backend/amd64/emit.mlp
```diff
@@ -1582,9 +1642,12 @@ let emit_instr fallthrough i =
       I.bswap (res32 i 0);
   | Lop(Ispecific(Ibswap { bitwidth = Sixtyfour })) ->
       I.bswap (res i 0)
-  | Lop(Ispecific(Ifloatsqrtf addr)) ->
+  | Lop(Ispecific(Ifloatsqrtf (Float64, addr))) ->
       I.xorpd (res i 0) (res i 0); (* avoid partial register stall *)
       I.sqrtsd (addressing addr REAL8 i 0) (res i 0)
+  | Lop(Ispecific(Ifloatsqrtf (Float32, addr))) ->
+      I.xorps (res i 0) (res i 0); (* avoid partial register stall *)
+      I.sqrtss (addressing addr REAL8 i 0) (res i 0)
```
xclerc, line 1650: Not sure why it is the same as the
`Float64` case; shouldn't we address
a 32-bit value?
(Also, is this actually generated?)

backend/x86_binary_emitter.ml
```diff
@@ -663,6 +663,12 @@ let emit_andpd b dst src =
       emit_mod_rm_reg b 0 [ 0x0f; 0x54 ] rm (rd_of_regf reg)
   | _ -> assert false
 
+let emit_andps b dst src =
```
xclerc, line 666: The only differences between the existing
instructions and the new ones are the
prefixes; should we factor out the common
parts?

tests/small_numbers/float32.ml
TheNumbat: Yep

backend/amd64/emit.mlp
TheNumbat: This should use `REAL4`, yeah. 
It's not currently being generated because I didn't add an override in selection. 
I've just made this a failure case; I will add an SSE2 instruction `Sqrt_scalar_f32` in another PR.

backend/x86_binary_emitter.ml
TheNumbat: Factored

TheNumbat:
I will add a separate float32 `machtype_component` as discussed, so that should cover testing mixing 32/64 in the backend.

backend/x86_binary_emitter.ml
```diff
@@ -634,10 +634,12 @@ let emit_movq b ~dst ~src =
     emit_mod_rm_reg b no_rex [ 0x0F; 0x7E ] rm (rd_of_regf reg)
   | _ -> assert false
 
-let emit_movsd b dst src =
+let emit_mov_float ~width b dst src =
   match (dst, src) with
   | Regf reg, ((Regf _ | Mem _ | Mem64_RIP _) as rm) ->
-      buf_int8 b 0xF2;
+      (match width with
```
xclerc, line 640: I would probably factor out the
code emitting the prefixes, but
I reckon it is not systematically
done in existing code.

backend/x86_binary_emitter.ml
```diff
@@ -1531,12 +1541,14 @@ let imm8_of_float_condition = function
   | NLEf -> 0x06
   | ORDf -> 0x07
 
-let emit_cmpsd b ~condition ~dst ~src =
+let emit_cmp_float ~width b ~condition ~dst ~src =
   match (dst, src) with
   | (Regf reg, ((Regf _ | Mem _ | Mem64_RIP _) as rm)) ->
     (* CMPSD xmm1, xmm2/m64, imm8 *)
```
xclerc, line 1547: (The comment should probably be tweaked.)

backend/x86_binary_emitter.ml
```diff
@@ -1836,9 +1848,9 @@ let imm arg = match arg with Imm n -> Int64.to_int n | _ -> assert false
 
 let assemble_instr b loc = function
   | ADD (src, dst) -> emit_ADD b dst src
-  | ADDSD (src, dst) -> emit_addsd b dst src
+  | ADDSD (src, dst) -> emit_add_float ~width:Cmm.Float64 b dst src
```
xclerc, line 1851: Not for this PR, but I wonder
whether the size should actually
be part of the instruction (_i.e._
have only one contructor for
`addsd` and `addss`).

backend/x86_binary_emitter.ml
TheNumbat: Yeah, I didn't do that because the choice of prefix depends on the instruction (though only either 0x66/nothing or 0xF3/0xF2 here)

backend/x86_binary_emitter.ml
TheNumbat: Fixed

TheNumbat:
I merged the new `machtype_component` from #2391 and added it to the unboxed external args test here

xclerc (approved):

mshinwell:
`4ffa571e1e9fab9c94cfc95cc345be7a8c687af7 ` is ok

PR #2387, opened 2024-03-22
Use extension node for encoding labeled tuples

ncik-roberts:
Encode labeled tuple expressions `~x, ~y` as:

```ocaml
([%jane.labeled_tuple] (x, y)) [@jane.labeled_tuple.x.y]
```

instead of merely:

```ocaml
(x, y) [@jane.labeled_tuple.x.y]
```

The intention of this is to make it less likely that tools that consume the OCaml parsetree look past the attribute and treat a labeled tuple as a normal tuple.

I didn't add any additional tests, but existing ones continue to pass.


ccasin (approved):
Looks good to me, fine to merge as is.

But, one question: Why not play a similar game with patterns and types?

PR #2388, opened 2024-03-22
Better atomicity of mode system

riaqn:
Currently, `equate` is implemented as two consecutive `submode`, who are themselves atomic. However, the `equate` itself is not atomic: if the first `submode` succeeds but the second one fails, the mutations by the first `submode` is not reverted. While this hasn't caused any issues, it's better to fix it.


ocaml/typing/solver.ml
```diff
@@ -484,88 +485,70 @@ module Solver_mono (C : Lattices_mono) = struct
...
+  let submode (type a r l) (obj : a C.obj) (a : (a, allowed * r) mode)
+      (b : (a, l * allowed) mode) ~log =
     let submode_cc ~log:_ obj left right =
       if C.le obj left right then Ok () else Error { left; right }
     in
     let submode_mvc ~log obj v right =
       Result.map_error
         (fun left -> { left; right })
-        (submode_mvc ~log obj v right)
+        (submode_mvc ~log:(Some log) obj v right)
```
goldfirere, line 496: I'm a little worried about how often we might allocate these `Some` nodes. Maybe it makes sense to have `submode` take an optional log just to avoid the repeated allocation? Not sure.

ocaml/typing/solver.ml
```diff
@@ -595,7 +578,7 @@ module Solver_mono (C : Lattices_mono) = struct
           | Amode b -> loop (C.join obj a b) mvs xs
           (* some minor optimization: if [a] is lower than [mlower mv], we
               should keep the latter instead. This helps to fail early in
-             [submode_try] *)
+             [submode_log] *)
```
goldfirere, line 581: There's no function of this name.

ocaml/typing/solver.ml
```diff
@@ -620,45 +603,46 @@ module Solver_mono (C : Lattices_mono) = struct
...
     in
     loop (C.max obj) [] l
 
-  let zap_to_floor_morphvar ~commit obj mv =
-    let log, lower = zap_to_floor_morphvar_aux obj mv in
-    if commit then !append_changes log else undo_changes !log;
+  let zap_to_floor_morphvar obj mv ~log =
+    let log_, lower = zap_to_floor_morphvar_aux obj mv in
+    (match log with
+    | None -> undo_changes log_
```
goldfirere, line 617: This surprises me. I would expect the `log` argument to be `None` when I don't care about logging.... but this treatment means that `None` means "be pure". I would make this more explicit. Maybe I just mean that I like the old `commit` approach.

ocaml/typing/mode.ml
```diff
@@ -1115,11 +1118,29 @@ module type Obj = sig
...
+  match op ~log with
+  | Ok _ as x ->
+    !append_changes log;
+    x
+  | Error _ as x ->
+    S.undo_changes !log;
+    x
+  [@@inline]
+
+let atomically' op =
```
goldfirere, line 1132: Maybe `append_local_log`? There's nothing particularly atomic here.

goldfirere (approved):
This seems a little marginal: there's not a known problem that this fixes. Though I suppose it's no worse than we had before.

But even better than this, I think, is an explicit notion of mode equality in the solver, where equality is not the composition of two submode checks. I thought we had discussed this possibility.

riaqn:
> But even better than this, I think, is an explicit notion of mode equality in the solver, where equality is not the composition of two submode checks. I thought we had discussed this possibility.

That's right. There is another thing this PR applies to: submode on two `Value` modes consists of submoding the comonadic fragment and the monadic fragment. If the second submoding fails, the first successful submoding is not reverted. While this hasn't caused an issue, it's conceptually simpler to just say "all submoding operations exposed by `mode.mli` is atomic".

ocaml/typing/solver.ml
riaqn: good point - this replaces the `commit` argument which has clearer meaning. I will rename the argument with comment.

ocaml/typing/solver.ml
riaqn: Good point about the allocation. I lifted the allocation to the top level of `submode`, which means just one allocation per `Solver.submode`. Note that `Mode.Value.submode` consists of two `Solver.submode`; so if we lift the allocation to `Mode.Value.submode` we can reduce the allocation further.

That would mean `solver.submode` as you said takes optional log. But I think solver not logging is just wrong and we would be exposing the unsafe capability that the user should never use.

ocaml/typing/mode.ml
riaqn: Hmm I do hope to mean that `atomically'` is a special case of `atomically` that always succeeds. Do you have suggestions?

ocaml/typing/solver.ml
riaqn: Well on a second thought, considering that `mode.ml` is the only user of `solver.mli`, both of which are maintained by the same set of people, it might be fine to allow `solver.mli` be a bit "primitive" and expose bare metals.

ocaml/typing/mode.ml
goldfirere: But if it always succeeds, then there's nothing interestingly atomic about it. It's needed just because some functions work with a local log, instead of appending to the global one. That's why I suggested `append_local_log` or maybe `with_local_log`.

riaqn: I have renamed them to `with_log` and `try_with_log` - that should incorporate both of ours ideas.

PR #2389, opened 2024-03-22
Unconditionally reserve the top header byte in runtime 4 and runtime 5

ncik-roberts:
We plan on using the top header byte to encode mixed blocks. This PR:

  * Reserves the header byte in runtime 5
  * Removes the ability to configure reserved bits in runtime 5
  * Restores the ability to independently configure reserved bits (called "profinfo_width") in runtime 4. We don't enable this by default; this is intentional.

This PR can and should be merged prior to #2380.

### DIFF
--- a/ocaml/utils/config.mli
+++ b/ocaml/utils/config.mli
@@ -237,9 +237,9 @@ val reserved_header_bits : int
    regardless of whether we're in runtime 4 or runtime 5.
 
    In runtime 5, this corresponds to the HEADER_RESERVED_BITS C preprocessor
-   macro.  In runtime 4, this corresponds to the PROFINFO_WIDTH C preprocessor
-   macro (and is controlled by the --with-profinfo and --enable-profinfo-width
-   configure arguments), and is only set when WITH_PROFINFO is also true.
+   macro. In runtime 4, this corresponds to the PROFINFO_WIDTH C preprocessor
+   macro. Both of these are unconditionally set to a constant by the configure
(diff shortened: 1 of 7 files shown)


ncik-roberts:
Based on a conversation with @mshinwell I've merged #2397 into this branch. (But #2397 was unreviewed.) So review in this PR should be for both runtime 4 and runtime 5.

ocaml/utils/config.mli
```diff
@@ -233,7 +233,14 @@ val with_cmm_invariants : bool
 (** Whether the invariants checks for Cmm are enabled *)
 
 val reserved_header_bits : int
-(** How many bits of a block's header are reserved *)
+(** How many bits of a block's header are reserved. This is correct
+   regardless of whether we're in runtime 4 or runtime 5.
+
+   In runtime 5, this corresponds to the HEADER_RESERVED_BITS C preprocessor
+   macro.  In runtime 4, this corresponds to the PROFINFO_WIDTH C preprocessor
+   macro (and is controlled by the --with-profinfo and --enable-profinfo-width
```
mshinwell, line 241: The part in parenthesis should be removed now, I think?

mshinwell (changes requested):

ocaml/utils/config.mli
ncik-roberts: Yes, thanks.

mshinwell (approved):

PR #42, opened 2021-05-28
Remove Cmm.memory_chunk.Double_u

gretay-js:
All current targets treat Double_u the same as Double. 
Double_u was [introduced](https://github.com/ocaml/ocaml/commit/c02c1359461e27ba91ccd8751f5855c50f9929f7) for sparc and mips, which treated it differently from Double.

This PR only makes changes in `ocaml` directory for now.

PR #2390, opened 2024-03-22
Fix a bug in `rec_check.ml` for arrays of unboxed things

ccasin:
Before this PR, this program segfaults in bytecode and hits a middle-end assertion failure on flambda2:
```ocaml
let _ =
  let rec x = [| x |]; #42.0 in
  ();;
```
After this PR, it is correctly rejected by the typechecker.  The issue was introduced for arrays of unboxed types in #2185 and has a one line fix.  I can't imagine anyone would ever hit this in practice, I just noticed it while reading the mixed blocks PR (which also touches this file).

See upstream ocaml/ocaml#6939 for why we reject this - the same is done for normal float arrays.  While the program technically meets the criterea for being a sensible recursive definition, supporting it is hard.

### DIFF
--- a/ocaml/testsuite/tests/typing-layouts-arrays/basics.ml
+++ b/ocaml/testsuite/tests/typing-layouts-arrays/basics.ml
@@ -254,3 +254,29 @@ Error: This expression has type ('a : float64)
        But the layout of int32# must be a sublayout of float64, because
          of the definition of arr at line 6, characters 12-16.
 |}]
+
+(*********************)
+(* Test 7: rec check *)
+
+(* See upstream PR #6939 *)
+
+let _ =
+  let[@warning "-10"] rec x = [| x |]; #42.0 in
+  ();;
+[%%expect{|
+Line 2, characters 30-44:
+2 |   let[@warning "-10"] rec x = [| x |]; #42.0 in
+                                  ^^^^^^^^^^^^^^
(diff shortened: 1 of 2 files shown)


lthls:
The compilation of recursive values assumes that every recursive variable is of kind Value. Your fix still allows things like `let rec x = #42.0` which should be rejected.
Is it possible to catch this during type-checking ? That would provide better error messages too.

goldfirere:
While a self-recursive non-value type seems hard to imagine, a mutually-recursive one can be ok:

```ocaml
let rec r = { x } and x = #{ r }
```

seems OK to me. I think detecting the difference between self-recursion and mutual recursion will be hard, and we should just let the rec_check module handle this for us.

goldfirere (approved):

lthls:
I meant that the code that translates recursive definitions has no support for non-value bindings unless they are not actually recursive (as far as I know). So your example would get accepted and wrongly compiled in some way.

goldfirere:
We don't have unboxed structures yet, so we're safe for now. But, yes, we will need to address this more fully soon.

ccasin:
Just to add: I agree with @lthls that it makes sense to ban all such recursive definitions elsewhere in the typechecker, at least until we add some way to support them.  But I still think it made sense to merge this narrower PR as it fixes a clear bug in `rec_check`.

PR #2392, opened 2024-03-25
Fix `Mode.zap_to_floor` BUG

riaqn:
(Based on #2388 , for easier rebasing).

Given `f v`, where `v` is a mode variable and `f` is a morphism, applying `zap_to_floor` will first apply `zap_to_floor` on `v` to get `a`, and then return `f a`.

This is wrong; in particular, if `f` is projection, we would be zapping the whole product to floor instead of just one axis. This wasn't triggering any issue, because for existing axes we want to zap them in the same direction anyway. This is no longer true as we introduce more axes with different legacy direction.

This isn't a soundness issue, as all mode constraints are still met.

### DIFF
--- a/ocaml/testsuite/tests/formatting/test_locations.dlocations.ocamlc.reference
+++ b/ocaml/testsuite/tests/formatting/test_locations.dlocations.ocamlc.reference
@@ -88,14 +88,14 @@ let rec fib = function | 0 | 1 -> 1 | n -> (fib (n - 1)) + (fib (n - 2))
       <def>
         pattern (test_locations.ml[17,534+8]..test_locations.ml[17,534+11])
           Tpat_var "fib"
-          value_mode global,many,shared
+          value_mode meet_local,once(0[global,many,global,many]),join_shared(1[shared,shared])
         expression (test_locations.ml[17,534+14]..test_locations.ml[19,572+34])
           Texp_function
           region true
-          alloc_mode global,many,shared
+          alloc_mode map_comonadic(regional_to_global)(6[global,many,global,many]),id(7[shared,shared])
           []
           Tfunction_cases (test_locations.ml[17,534+14]..test_locations.ml[19,572+34])
-          alloc_mode global,many,shared
+          alloc_mode id(2[global,many,global,many]),id(3[shared,shared])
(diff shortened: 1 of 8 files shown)


ocaml/testsuite/tests/formatting/test_locations.dlocations.ocamlc.reference
```diff
@@ -88,14 +88,14 @@ let rec fib = function | 0 | 1 -> 1 | n -> (fib (n - 1)) + (fib (n - 2))
       <def>
         pattern (test_locations.ml[17,534+8]..test_locations.ml[17,534+11])
           Tpat_var "fib"
-          value_mode global,many,shared
+          value_mode meet_local,once(0[global,many,global,many]),join_shared(1[shared,shared])
```
goldfirere, line 91: The changes in this file seem worrying to me. That is, the change proposed here means variables will live for longer than they did previously, but I'm not sure of the upside. I understand we want to be able to set the legacy modes separately, but it's not obvious to me that the step in this PR is necessary.

goldfirere (commented):
The implementation looks good, and is a nice little simplification. I'm just worried about the performance impact -- especially because I think we don't need to make this change in order to have varying legacy axes (but I could be wrong).

ocaml/testsuite/tests/formatting/test_locations.dlocations.ocamlc.reference
riaqn: This is caused by `check_const` now requiring the mode to be allowed on both LHS and RHS, because it calls `zap_to_floor_morphvar` which requires that. the mode on patterns is left-only so cannot be `check_const`, and therefore we are printing them raw.

I think this doesn't affect anything on the production path - `check_const` is only called for tests and debug printing.

goldfirere: Not quite: `check_const` is also used to extract the mode from arrow types in `external` declarations. But I agree that we needn't worry about that case. OK -- I'm much happier with this patch now.

goldfirere (approved):

PR #2396, opened 2024-03-25
Refactor arrow modes printing

riaqn:
This PR refactors the logic around modes in `printtyp.ml` and `oprint.ml`, in particular the part about currying.

Currently, `oprint.ml` has its own copy of handling currying, legacy modes and such, which duplicates code elsewhere. 

Instead, we should make `printtyp.ml` worry about these (using the existing code from `mode.ml` and `ctype.ml`), and `outcometree.mli` should reflect the final printing, and `oprint.ml` should verbatim print the given outcometree

### DIFF
--- a/ocaml/testsuite/tests/typing-implicit-source-positions/invalid_usages.ml
+++ b/ocaml/testsuite/tests/typing-implicit-source-positions/invalid_usages.ml
@@ -44,7 +44,7 @@ Line 1, characters 14-15:
                   ^
 Error: This expression has type call_pos:[%call_pos] -> unit -> unit
        but an expression was expected of type
-         call_pos:Lexing.position -> (unit -> 'a)
+         call_pos:Lexing.position -> unit -> 'a
 |}]
 
 let h ?(call_pos:[%call_pos]) () = ()
--- a/ocaml/typing/ctype.ml
+++ b/ocaml/typing/ctype.ml
@@ -1588,6 +1588,17 @@ let with_locality locality m =
   Alloc.submode_exn (Alloc.meet_with (Comonadic Areality) Locality.Const.min m) m';
   m'
 
+let curry_mode alloc arg : Alloc.Const.t =
+  let acc =
+    Alloc.Const.join
+      (Alloc.Const.close_over arg)
+      (Alloc.Const.partial_apply alloc)
+  in
+  (* Arrow types cross uniqueness axis. Therefore, when user writes an
+  A -> B -> C (to be used as constraint on something), we should make
+  (B -> C) shared. A proper way to do this is via modal kinds. *)
+  {acc with uniqueness=Uniqueness.Const.Shared}
+
 let rec instance_prim_locals locals mvar macc finalret ty =
   match locals, get_desc ty with
   | l :: locals, Tarrow ((lbl,marg,mret),arg,ret,commu) ->
--- a/ocaml/typing/ctype.ml
+++ b/ocaml/typing/ctype.ml
@@ -1594,9 +1594,19 @@ let curry_mode alloc arg : Alloc.Const.t =
       (Alloc.Const.close_over arg)
       (Alloc.Const.partial_apply alloc)
   in
-  (* Arrow types cross uniqueness axis. Therefore, when user writes an
-  A -> B -> C (to be used as constraint on something), we should make
-  (B -> C) shared. A proper way to do this is via modal kinds. *)
+  (* For A -> B -> C, we always interpret (B -> C) to be of shared. This is the
+    legacy mode which helps with legacy compatibility. Arrow types cross
+    uniqueness so we are not losing too much expressvity here. One
+    counter-example is:
+
+    let g : (A -> B -> C) = ...
+    let f (g : A -> unique_ (B -> C)) = ...
+
+    And [f g] would not work, as mode crossing doesn't work deeply into arrows.
+    Our answer to this issue is that, the author of f shouldn't ask B -> C to be
+    unique_. Instead, they should leave it as default which is shared, and mode
+    crossing it to unique at the location where B -> C is a real value (instead
+    of the return of a function). *)
   {acc with uniqueness=Uniqueness.Const.Shared}
 
 let rec instance_prim_locals locals mvar macc finalret ty =
--- a/ocaml/typing/oprint.ml
+++ b/ocaml/typing/oprint.ml
@@ -338,14 +338,17 @@ let pr_var_jkinds =
   disallowed in parsing of this file, but non-legacy modes might still pop
   up. For example, the current file might cite values from other files that
   mention non-legacy modes *)
-let print_out_legacy_axis f ppf = function
-  | None -> ()
-  | Some m -> Format.fprintf ppf "%a_ " f m
+let print_out_mode ppf = function
+  | Omd_local -> fprintf ppf "local_"
+  | Omd_unique -> fprintf ppf "unique_"
+  | Omd_once -> fprintf ppf "once_"
 
-let print_out_modes ppf (m : Mode.Alloc.Const.Option.t) =
-  print_out_legacy_axis Mode.Locality.Const.print ppf m.locality;
-  print_out_legacy_axis Mode.Linearity.Const.print ppf m.linearity;
-  print_out_legacy_axis Mode.Uniqueness.Const.print ppf m.uniqueness
+let print_out_mode_space ppf m =
+  print_out_mode ppf m;
+  pp_print_space ppf ()
+
+let print_out_modes ppf l =
+  pp_print_list print_out_mode_space ppf l
 
 (* Labeled tuples with the first element labeled sometimes require parens. *)
 let is_initially_labeled_tuple ty =
@@ -377,12 +380,16 @@ let rec print_out_type_0 ppf =
    - Or, there is at least one mode to print.
  *)
 and print_out_type_mode ~arg mode ppf ty =
-  let mode = Format.asprintf "%a" print_out_modes mode in
+  let has_modes =
+    match mode with
+    | [] -> false
+    | _ -> true
+  in
   let parens =
     is_initially_labeled_tuple ty
-    && (arg || String.length mode > 0)
+    && (arg || has_modes)
   in
-  pp_print_string ppf mode;
+  print_out_modes ppf mode;
   if parens then
     pp_print_char ppf '(';
   print_out_type_2 ppf ty;
--- a/ocaml/typing/oprint.ml
+++ b/ocaml/typing/oprint.ml
@@ -332,108 +332,20 @@ let pr_var_jkind ppf (v, l) = match l with
 let pr_var_jkinds =
   print_list pr_var_jkind (fun ppf -> fprintf ppf "@ ")
 
-let join_locality lm1 lm2 =
-  match lm1, lm2 with
-  | Olm_local, _ -> Olm_local
-  | _, Olm_local -> Olm_local
-  | Olm_unknown, _ -> Olm_unknown
-  | _, Olm_unknown -> Olm_unknown
-  | Olm_global, Olm_global -> Olm_global
+(* NON-LEGACY MODES
+  Here, we are printing mode annotations even if the mode extension is
+  disabled. Mode extension being disabled means mode annotations are
+  disallowed in parsing of this file, but non-legacy modes might still pop
+  up. For example, the current file might cite values from other files that
+  mention non-legacy modes *)
+let print_out_legacy_axis f ppf = function
+  | None -> ()
+  | Some m -> Format.fprintf ppf "%a_ " f m
 
-let join_uniqueness u1 u2 =
-  match u1, u2 with
-  | Oum_shared, _ -> Oum_shared
-  | _, Oum_shared -> Oum_shared
-  | Oum_unknown, _ -> Oum_unknown
-  | _, Oum_unknown -> Oum_unknown
-  | Oum_unique, Oum_unique -> Oum_unique
-
-let join_linearity l1 l2 =
-  match l1, l2 with
-  | Olinm_once, _
-  | _, Olinm_once -> Olinm_once
-  | Olinm_unknown, _
-  | _, Olinm_unknown -> Olinm_unknown
-  | Olinm_many, Olinm_many -> Olinm_many
-
-let uniqueness_to_linearity = function
-  | Oum_unique -> Olinm_once
-  | Oum_shared -> Olinm_many
-  | Oum_unknown -> Olinm_unknown
-
-let join_modes m1 m2 =
-  { oam_locality = join_locality m1.oam_locality m2.oam_locality;
-    oam_uniqueness = join_uniqueness m1.oam_uniqueness m2.oam_uniqueness;
-    oam_linearity = join_linearity m1.oam_linearity m2.oam_linearity }
-
-let default_mode =
-  { oam_locality = Olm_global;
-    oam_uniqueness = Oum_shared;
-    oam_linearity = Olinm_many; }
-
-let close_over arg_mode =
-  let oam_locality = arg_mode.oam_locality in
-  let oam_uniqueness = Oum_shared in
-  let oam_linearity =
-    join_linearity
-      arg_mode.oam_linearity
-      (uniqueness_to_linearity arg_mode.oam_uniqueness)
-  in
-  { oam_locality;
-    oam_uniqueness;
-    oam_linearity }
-
-let partial_apply alloc_mode =
-  let oam_locality = alloc_mode.oam_locality in
-  let oam_uniqueness = Oum_shared in
-  let oam_linearity = alloc_mode.oam_linearity in
-  { oam_locality; oam_uniqueness; oam_linearity }
-
-(* Following functions are used to check if the return mode can omitted in the
-   case of currying *)
-let locality_agree expected real =
-  match expected, real with
-  (* If expected and real matches, can omit *)
-  | Olm_local, Olm_local | Olm_global, Olm_global -> true
-  (* If the real mode is unknown, we'd rather not put extra parentheses, because
-     we wouldn't put "unknown" around the parentheses either, which would make
-     the printing even less precise *)
-  | _, Olm_unknown -> true
-  (* In all other cases (the real mode is known), we will print the mode to be
-     safe*)
-  | _, _ -> false
-
-let uniqueness_agree expected real =
-  match expected, real with
-  | Oum_unique, Oum_unique | Oum_shared, Oum_shared -> true
-  | _, Oum_unknown -> true
-  | _, _ -> false
-
-let linearity_agree expected real =
-  match expected, real with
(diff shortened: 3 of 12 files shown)


ocaml/testsuite/tests/typing-implicit-source-positions/invalid_usages.ml
```diff
@@ -44,7 +44,7 @@ Line 1, characters 14-15:
                   ^
 Error: This expression has type call_pos:[%call_pos] -> unit -> unit
        but an expression was expected of type
-         call_pos:Lexing.position -> (unit -> 'a)
+         call_pos:Lexing.position -> unit -> 'a
```
riaqn, line 47: It seems that this PR improves the error message here: the root cause here is type mismatch (`[%call_pos]` vs `Lexing.position`), hence the extra parens is just confusing and shouldn't be printed.

ocaml/typing/ctype.ml
```diff
@@ -1588,6 +1588,17 @@ let with_locality locality m =
...
 
+let curry_mode alloc arg : Alloc.Const.t =
+  let acc =
+    Alloc.Const.join
+      (Alloc.Const.close_over arg)
+      (Alloc.Const.partial_apply alloc)
+  in
+  (* Arrow types cross uniqueness axis. Therefore, when user writes an
+  A -> B -> C (to be used as constraint on something), we should make
+  (B -> C) shared. A proper way to do this is via modal kinds. *)
```
goldfirere, line 1599: I would expect we'd want to make `B -> C` unique, not shared.

ocaml/typing/oprint.ml
```diff
@@ -445,55 +360,43 @@ let string_of_gbl_space = function
...
  *)
 and print_out_type_mode ~arg mode ppf ty =
-  let is_local = is_local mode in
-  let is_unique = is_unique mode in
-  let is_once = is_once mode in
+  let has_modes =
+    match mode with
+    | [] -> false
+    | _ -> true
+  in
```
goldfirere, line 387:
```suggestion
  let has_modes = not (List.is_empty mode) in
```

ocaml/typing/printtyp.ml
```diff
@@ -1281,7 +1257,20 @@ let outcome_label : Types.arg_label -> Outcometree.arg_label = function
   | Optional l -> Optional l
   | Position l -> Position l
 
-let rec tree_of_typexp mode ty =
+let tree_of_mode mode l =
```
goldfirere, line 1260:
```suggestion
(* If the [mode] is [Some m], returns a list containing the corresponding [out_mode] from
   the list [l].
   Precondition: If [mode] is [Some m], [m] is included as a first element in [l]. *)
let tree_of_mode (mode : 'mode option) (l : ('mode * out_mode) list) : out_mode list =
```

ocaml/typing/printtyp.ml
```diff
@@ -1281,7 +1257,20 @@ let outcome_label : Types.arg_label -> Outcometree.arg_label = function
...
+  match mode with
+  | None -> []
+  (* should never raise *)
+  | Some x -> [List.assoc x l]
+
+let tree_of_modes modes =
+  let diff = Mode.Alloc.Const.diff modes Mode.Alloc.Const.legacy in
+  tree_of_mode diff.locality [Mode.Locality.Const.Local, Omd_local] @
```
goldfirere, line 1268:
```suggestion
  (* The lists passed to [tree_of_mode] must contain mappings for all non-legacy modes. *)
  tree_of_mode diff.locality [Mode.Locality.Const.Local, Omd_local] @
```

ocaml/typing/printtyp.ml
```diff
@@ -1300,27 +1289,31 @@ let rec tree_of_typexp mode ty =
           if !print_labels || is_omittable l then outcome_label l
           else Nolabel
         in
+        (* [marg] will contain undetermined axes. It would be imprecise if we
+          don't print anything for those axes, since user would interpret that
+         as legacy. The best we can do is to zap to legacy and if they do land
+         at legacy, we will be able to omit the printing. *)
+        let am = Alloc.zap_to_legacy marg in
```
goldfirere, line 1296: I find this really scary. It means that printing a type can change inference. If we print during debugging, say, we're now altering the course of type inference. I think we shouldn't do this.

Instead, what if we printed e.g. `local?` in this scenario? Obviously that's not parseable. But it's more honest than anything we could say. Or it could be `'_locality` maybe, invoking the weak polymorphism around the value restriction.

ocaml/typing/printtyp.ml
```diff
@@ -1281,7 +1257,20 @@ let outcome_label : Types.arg_label -> Outcometree.arg_label = function
...
+  (* should never raise *)
+  | Some x -> [List.assoc x l]
+
+let tree_of_modes modes =
+  let diff = Mode.Alloc.Const.diff modes Mode.Alloc.Const.legacy in
+  tree_of_mode diff.locality [Mode.Locality.Const.Local, Omd_local] @
+  tree_of_mode diff.linearity [Mode.Linearity.Const.Once, Omd_once] @
+  tree_of_mode diff.uniqueness [Mode.Uniqueness.Const.Unique, Omd_unique]
+
+(* [alloc_mode] is only used if [ty] is arrow-shaped. *)
```
goldfirere, line 1272: But what is it? The mode expected by the context or something? That is, when I'm calling `tree_of_typexp`, how do I know what to pass as `alloc_mode`?

ocaml/typing/printtyp.ml
goldfirere, line 1296:
```suggestion
        let arg_mode = Alloc.zap_to_legacy marg in
```

ocaml/typing/printtyp.ml
```diff
@@ -1420,26 +1413,47 @@ and tree_of_qtvs qtvs =
...
 and tree_of_typ_gf (ty, gf) =
   let gf =
     match gf with
     | Global_flag.Global -> Ogf_global
     | Global_flag.Unrestricted -> Ogf_unrestricted
   in
-  (tree_of_typexp Type ty, gf)
+  (tree_of_typexp Type Alloc.Const.legacy ty, gf)
+
+and tree_of_ret_typ mode acc_mode (m, ty) =
```
goldfirere, line 1437: There are three mode parameters here. Please write a comment (maybe with an example) explaining what they all are. Probably also good to use labels. As it is, I don't understand this function well enough to be able to review well. (Don't resolve, so that I come back and review.)

goldfirere (commented):
I think this is a great direction of travel -- just one little hang-up, as you'll see in the comments.

ocaml/typing/oprint.ml
riaqn: For some reason `List.is_empty` is not available here - do you know why?

ocaml/typing/ctype.ml
riaqn: In fact neither is correct. We don't know in which variance this `B->C` will be, so we don't know which direction to push.

For example: `let f : A -> B -> C = ...`, in this case we want to push to `unique` for a stronger type.
Another example: `let f (g : A -> B -> C) = ...` in this case we want to push to `shared`, so that `f` is more permissive in its argument.

The current approach of making it `shared` just happens to work, because `shared` is legacy and we haven't used the axis much.

To properly support this, we need mode crossing on arrow types.

ocaml/typing/printtyp.ml
riaqn: After the type annotation, I believe the function is clear to read. The function is only 3 lines, after all.

ocaml/typing/printtyp.ml
riaqn: The first `mode` is unrelated to our mode system. I added some comments; together with the comments inside the function, I hope it's now clearer.

ocaml/typing/printtyp.ml
riaqn: > If we print during debugging,

Would it be better to use `Printtyped` for debugging? That one is more informative.

If we really need to call `Pprinttyp` for debugging, I suppose `Printtyp` can revert the changes (using log) after printing. 

> Instead, what if we printed e.g. local? in this scenario? 

Very often the root error is argument/return type mismatch rather than mode mismatch, in which case we want to deemphasize the modes. `local?` would be too conspicuous.

riaqn: Huh, so there is actually calling to `Btype.backtrack` in `printtyp.ml` - should we do that here too?

ocaml/typing/oprint.ml
goldfirere: Blargh. It's in 5.1. I actually tried to upgrade my system compiler to 5.1.1 just for this, but it ran into trouble, and I have no appetite for solving the trouble. Just keep what you have.

ocaml/typing/ctype.ml
goldfirere: I think what you're saying could maybe work? But probably not, because somewhere we'll have a `'a` that gets discovered to be `t1 -> t2` but it won't be unique.

But actually I think we do have subtyping for function arguments:

```ocaml
let f (g : string -> local_ (string -> string)) = g "hi" "bye"

let h s1 s2 = "foo"

let x = f h
```

This works, because global < local. So I think shared is fine here. But this has nothing to do with mode-crossing. The comment here is about mode-crossing, so that's maybe where I'm confused.

ocaml/typing/printtyp.ml
goldfirere: Backtracking would be OK.... but sometimes the problem will be the modes, and it seems willfully annoying for us to unconditionally suppress the modes. We could imagine choosing whether or not to print this based on whether or not modes are the issue in the error (the original call into this module likely knows this). Have you tried the `local?` approach? How much does it clutter real output?

And in answer to your first question, I think people (I am one of these people) will want to use pretty-printed typed during debugging.

My guess is that the answer will be to have the original caller say whether or not modes should be printed.

ocaml/typing/printtyp.ml
goldfirere: Maybe forgot to push? I don't see the comments.

ocaml/typing/printtyp.ml
lpw25: I think we should continue the current approach of not printing a mode when it is unconstrained. We already know that the mode in question can’t be the cause of a type error, since that would require it be constrained in some way. I can’t see how printing anything else here won’t just add unhelpful information to a huge percentage of error messages.

We also shouldn’t have type printing mutate anything. That’s bound to cause trouble, as Richard suggests.

ocaml/typing/oprint.ml
riaqn: Yeah I realised this last night. I also tried to update my system compiler to 5 but ran into trouble.

ocaml/typing/ctype.ml
riaqn: Huh indeed. I had the wrong impression.

Now consider the example:
```
let g : (A -> B -> C) = ...
```
if we say `B -> C` is `shared`, we will be producing the `g` weaker than it needs to be. And now if we have:
```
let f (g : A -> unique_ (B -> C)) = ...
```
`f g` will not work, because mode crossing doesn't work deeply into arrows. But I suspect this won't cause much trouble in practice, because the author of `f` shouldn't ask `B -> C` to be `unique_`. Instead, they should leave it as default which is `shared`, and mode crossing it to `unique` at the location where `B -> C` is a real value (instead of the return of a function).

ocaml/typing/printtyp.ml
riaqn: >  but sometimes the problem will be the modes, and it seems willfully annoying for us to unconditionally suppress the modes.

Yes those cases are also fine. Because our approach here is to suppress modes as much as allowed by the constraints; In the case of mode mismatch, `mode0` and `mode1` will prints differently, even though both are suppressed.

riaqn: OK I've pushed a commit that does backtracking.

ocaml/typing/ctype.ml
goldfirere: Yes I agree with all that. We could perhaps add mode-crossing considerations to work in functions (I think this would fit into the current approach to arrow subtyping), but we probably won't ever have to.

goldfirere: I still don't quite see what this conversation has to do with the line of code / comment highlighted, though...

ocaml/typing/printtyp.ml
goldfirere: What if there's an equality check between a mode constrained between `unique` and `exclusive` and one that is known to be `aliased`? I agree the problem can't arise on two-mode axes, but that won't last forever. All that said, it's likely reasonable to delay this until it happens in practice. But when it does, if we suppress this printing, the person who gets the error will likely lose hours of time getting mad at us.

riaqn: The mode constrained between `unique` and `exclusive` will be zapped towards legacy, which is `aliased`. So it will be printed as `exclusive`.

The other one will be printed as `aliased`.

ocaml/typing/ctype.ml
riaqn: Yes indeed. The new commit has rewritten this comment.

ocaml/typing/printtyp.ml
lpw25: My preferred solution for 3-valued axes is not actually to zap to legacy in the printer. Rather I would ensure that the solver, when there is a failure still makes the constraint as true as possible. So if `m` cannot be made less than `n` it should zap `m` to its lower bound and zap `n` to its upper bound. That way any modes causing an error are constant and the printer will show exactly the useful information.

This follows a general principle that constraint solvers should make constraints as true as they can, so that error printing highlights only the genuine differences.

Zapping to legacy will sometimes have this effect — as in your example — but will sometimes go the other way and make modes look further apart than they really are. So I’d rather the printer just continued to ignore non-constant modes.

Once we have the unzipped syntax, I’d like to go further and try avoiding printing any modes when the error is about types, and any types when the error is about modes. Not sure if that will work for every error, but I think it’s the right direction to go.

riaqn: To rephrase what you said:
- When submoding fails in `ctype.ml`, we should make the modes into constants such that they are close to each other.
- In printer, we still want to call `zap_to_legacy` on modes, because there will be undetermined modes, and we want to de-emphasize those.

Yes I agree with that. I wonder at which level should we implement "leave submode error in the closest state". I looked at `solver.ml` briefly, and it would be too much effort to re-work the pre/post-conditions of those mutually recursive functions.  Considering that the errors are not on the production path so performance doesn't matter, it's easiest to wrap each call to `submode/equate` in `ctype.ml` with a further `submode` that does the constraint in case of error. This is easy because `submode a b` already gives in the case of error `{left; right}` where `left` is the lowest `a` can be, and `right` is the highest `b` can be. 

> So I’d rather the printer just continued to ignore non-constant modes.

We can't just ignore non-constant modes - a mode constrained between `unique` and `exclusive` still needs to be printed, for otherwise the user would assume `shared`.

lpw25: It should be in `solver.ml` on the error path. It is a concern of the mode solver not `ctype.ml`. You shouldn’t need to change any recursive functions, just add another layer of wrapping if you need it. The old solver already needed a layer of wrapping for converting exceptions to results, I would put this code in there if it still exists.

> We can't just ignore non-constant modes - a mode constrained between unique and exclusive still needs to be printed

I disagree. Whatever you print there will confuse people because it either won’t be accurate or it won’t be valid syntax. If you print nothing users just won’t see it and we already no it is not the part of the type the user needs to see because it’s not relevant to the error.

I think the right model is not: “`a -> b` says that the function takes a global value and returns a global value”, but instead “`a -> b` does not specify what modes the argument and return value have. Combined with a rule that says if we don’t know the mode for something we default it to legacy. I think this model makes what is going on with the unzipped syntax much clearer. It also allows error messages to just leave out modes that aren’t important (just as the type printer should, but currently doesn’t, replace unimportant parts of types with `_`).

Perhaps it would help if we changed the meaning of `(e : a -> b)` to not constrain the modes of the function. So users would write `( e : a @ . -> b @ .)` if they really wanted to constrain the modes.

> we still want to call zap_to_legacy on modes

For clarity, I would prefer to not print such modes rather than zap them to legacy. I don’t mind zapping them to legacy for now, since for two-mode axes that’s very close to the same thing.

riaqn: Yes, if we say `a -> b` doesn't specify any modes at all, then I'm happy to omit printing modes when they are not the root error. BTW this probably needs overhauling `ctype.ml` to pinpoint the exact mismatched mode; but I guess that will be subsumed by the PR of unzipped syntax.

In signatures `a -> b` will still default to legacy.

I also wonder how much of user's code relies on `a -> b` meaning legacy on LHS&RHS. I guess it should be mostly fine, because we are making more programs type check.

ocaml/typing/printtyp.ml
```diff
@@ -1434,11 +1444,20 @@ and tree_of_typ_gf (ty, gf) =
...
+(** We are on the RHS of an arrow type, where [ty] is the return type, and [m]
+    is the return mode. This function decides the printed modes on [ty].
+    - If [ty] is another arrow type, [acc_mode] is the mode that has accumulated
+      from the currying, and thus the mode that the user would interpret as on
+      [ty] if it doesn't have parens around it.
+    - If [ty] is not an arrow type, [acc_mode] is meaningless.
+
+    NB: This function might mutate states; the caller is responsible for
+    reverting them. *)
+and tree_of_ret_typ_unsafe mode acc_mode (m, ty) =
```
goldfirere, line 1456: I really dislike using `unsafe` to mean so many different things. (That is, different uses of it mean different things.) How about `effectful`? `mutating`? `impure`? `with_effects`? `should_be_backtracked`? Any of these would be better than `unsafe` for me.

goldfirere (approved):

ocaml/typing/printtyp.ml
```diff
@@ -1257,19 +1257,25 @@ let outcome_label : Types.arg_label -> Outcometree.arg_label = function
   | Optional l -> Optional l
   | Position l -> Position l
 
-let tree_of_mode mode l =
-  match mode with
-  | None -> []
-  (* should never raise *)
-  | Some x -> [List.assoc x l]
+(** [tree_of_mode m l] finds the outcome node in [l] that corresponds to [m].
+Raise if not found. *)
```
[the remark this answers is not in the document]
riaqn: Hmm not sure what you mean? Surely the only caller of this function will ensure that it never raises, but it's fair to describe the function as "might rasie"?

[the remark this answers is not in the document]
goldfirere: Yes, you're right. I took a quick look at the body, saw an `Option.map`, and made a false assumption.

ocaml/typing/printtyp.ml
riaqn: Good point - changed to `mutating`.

PR #2398, opened 2024-03-26
Add portability and contention

riaqn:
This is based on #2369 #2395 #2396 #2392   - we should merge those first. I will then rebase this PR on top of those. Currently only the last commit belongs to this PR.

This is a minimal PR that adds syncness/contention axes that interacts with mutations.

### DIFF
--- a/ocaml/stdlib/array.ml
+++ b/ocaml/stdlib/array.ml
@@ -23,7 +23,7 @@ type 'a t = 'a array
 
 (* Array operations *)
 
-external length : 'a array -> int = "%array_length"
+external length : 'a array @ contended -> int = "%array_length"
 external get: 'a array -> int -> 'a = "%array_safe_get"
 external set: 'a array -> int -> 'a -> unit = "%array_safe_set"
 external unsafe_get: 'a array -> int -> 'a = "%array_unsafe_get"
@@ -41,7 +41,7 @@ external create_float: int -> float array = "caml_make_float_vect"
 
 module Floatarray = struct
   external create : int -> floatarray = "caml_floatarray_create"
-  external length : floatarray -> int = "%floatarray_length"
+  external length : floatarray @ contended -> int = "%floatarray_length"
   external get : floatarray -> int -> float = "%floatarray_safe_get"
   external set : floatarray -> int -> float -> unit = "%floatarray_safe_set"
   external unsafe_get : floatarray -> int -> float = "%floatarray_unsafe_get"
--- a/ocaml/stdlib/array.ml
+++ b/ocaml/stdlib/array.ml
@@ -23,7 +23,7 @@ type 'a t = 'a array
 
 (* Array operations *)
 
-external length : 'a array @ contended -> int = "%array_length"
+external length : 'a array -> int = "%array_length"
 external get: 'a array -> int -> 'a = "%array_safe_get"
 external set: 'a array -> int -> 'a -> unit = "%array_safe_set"
 external unsafe_get: 'a array -> int -> 'a = "%array_unsafe_get"
@@ -41,7 +41,7 @@ external create_float: int -> float array = "caml_make_float_vect"
 
 module Floatarray = struct
   external create : int -> floatarray = "caml_floatarray_create"
-  external length : floatarray @ contended -> int = "%floatarray_length"
+  external length : floatarray -> int = "%floatarray_length"
   external get : floatarray -> int -> float = "%floatarray_safe_get"
   external set : floatarray -> int -> float -> unit = "%floatarray_safe_set"
   external unsafe_get : floatarray -> int -> float = "%floatarray_unsafe_get"
--- a/ocaml/testsuite/tests/typing-modes/class.ml
+++ b/ocaml/testsuite/tests/typing-modes/class.ml
@@ -76,12 +76,12 @@ Error: This value is nonportable but expected to be portable.
 class cla = object
     val mutable x = "hello"
 
-    method foo = x <- (world : _ @@ local)
+    method foo = x <- ("world" : _ @@ local)
 end
 [%%expect{|
-Line 4, characters 22-42:
-4 |     method foo = x <- (world : _ @@ local)
-                          ^^^^^^^^^^^^^^^^^^^^
+Line 4, characters 22-44:
+4 |     method foo = x <- ("world" : _ @@ local)
+                          ^^^^^^^^^^^^^^^^^^^^^^
 Error: This value escapes its region.
 |}]
 
--- a/ocaml/testsuite/tests/typing-modes/class.ml
+++ b/ocaml/testsuite/tests/typing-modes/class.ml
@@ -1,0 +1,149 @@
+(* TEST
+    flags += "-extension unique";
+    expect;
+*)
+
+(* This file tests that classes/objects are sound wrt modes. *)
+
+let unique_use : 'a @ unique -> unit = fun _ -> ()
+
+let portable_use : 'a @ portable -> unit = fun _ -> ()
+[%%expect{|
+val unique_use : unique_ 'a -> unit = <fun>
+val portable_use : 'a @ portable -> unit = <fun>
+|}]
+
+(* There is a closure_lock of legacy around a class. We test for comonadic and
+   monadic axes separately. *)
+
+(* class cannot refer to external local things *)
+let foo () =
+    let local_ s = "hello" in
+    let module M = struct
+    class cla = object
+        val k = s
+    end
+    end in ()
+[%%expect{|
+Line 5, characters 16-17:
+5 |         val k = s
+                    ^
+Error: The value s is local, so cannot be used inside a class.
+|}]
+
+(* class can refer to external unique things, but only as shared. *)
+let foo () =
+    let unique_ s = "hello" in
+    let module M = struct
+    class cla = object
+        val k = unique_use s
+    end
+    end in ()
+[%%expect{|
+Line 5, characters 27-28:
+5 |         val k = unique_use s
+                               ^
+Error: This value is shared but expected to be unique.
+  Hint: This identifier cannot be used uniquely,
+  because it is defined in a class.
+|}]
+
+(* instance variables need to be defined as legacy *)
+class cla = object
+    val x = ("world" : _ @@ local)
+end
+[%%expect{|
+Line 2, characters 12-34:
+2 |     val x = ("world" : _ @@ local)
+                ^^^^^^^^^^^^^^^^^^^^^^
+Error: This value escapes its region.
+|}]
+
+(* instance variables are available as legacy to methods *)
+class cla = object
+    val x = ("world" : _ @@ portable)
+
+    method foo = portable_use x
+end
+[%%expect{|
+Line 4, characters 30-31:
+4 |     method foo = portable_use x
+                                  ^
+Error: This value is nonportable but expected to be portable.
+|}]
+
+class cla = object
+    method m = "hello"
+end
+
+(* For Pexp_send, the object needs to be legacy *)
+let foo () =
+    let local_ obj = new cla in
+    obj#m
+[%%expect{|
+class cla : object method m : string end
+Line 8, characters 4-7:
+8 |     obj#m
+        ^^^
+Error: This value escapes its region.
+|}]
+
+(* methods are available as legacy *)
+let foo @ portable =
+    let obj = new cla in
+    obj#m
+(* CR zqian: this should fail. *)
+[%%expect{|
+val foo : string = "hello"
+|}]
+
+(* for methods, arguments can be of any modes *)
+class cla = object
+    method foo (x : string) = portable_use x
+end
+[%%expect{|
+class cla : object method foo : string @ portable -> unit end
+|}]
+
+(* the argument mode is soundly required during application *)
+let foo () =
+    let x @ nonportable = "hello" in
+    let o = new cla in
+    o#foo x
+[%%expect{|
+Line 4, characters 10-11:
+4 |     o#foo x
+              ^
+Error: This value is nonportable but expected to be portable.
+|}]
+
+
+(* Closing over classes affects closure's mode *)
+let (foo @ portable) () =
+    new cla
+(* CR zqian: this should fail. *)
+[%%expect{|
+val foo : unit -> cla = <fun>
+|}]
+
+module type SC = sig
+    class cla : object end
+end
+[%%expect{|
+module type SC = sig class cla : object  end end
+|}]
+
+let (foo @ portable) () =
+    let m = (module struct class cla = object end end : SC) in
+    let module M = (val m) in
+    new M.cla
+[%%expect{|
+val foo : unit -> <  > = <fun>
+|}]
+
+(* objects are always legacy *)
+let obj @ portable = new cla
+(* CR zqian: this should fail. *)
+[%%expect{|
+val obj : cla = <obj>
+|}]
--- a/ocaml/testsuite/tests/typing-modes/class.ml
+++ b/ocaml/testsuite/tests/typing-modes/class.ml
@@ -89,12 +89,12 @@ Error: This value escapes its region.
 |}]
 
 (* methods are available as legacy *)
-let foo @ portable =
+let u =
     let obj = new cla in
-    obj#m
+    portable_use obj#m
 (* CR zqian: this should fail. *)
 [%%expect{|
-val foo : string = "hello"
+val u : unit = ()
 |}]
 
 (* for methods, arguments can be of any modes *)
@@ -119,11 +119,12 @@ Error: This value is nonportable but expected to be portable.
 
 
 (* Closing over classes affects closure's mode *)
-let (foo @ portable) () =
-    new cla
+let u =
+    let foo () = new cla in
+    portable_use foo
 (* CR zqian: this should fail. *)
 [%%expect{|
-val foo : unit -> cla = <fun>
+val u : unit = ()
 |}]
 
 module type SC = sig
@@ -133,17 +134,22 @@ end
 module type SC = sig class cla : object  end end
 |}]
 
-let (foo @ portable) () =
-    let m = (module struct class cla = object end end : SC) in
-    let module M = (val m) in
-    new M.cla
+let u =
+    let foo () =
+        let m = (module struct class cla = object end end : SC) in
+        let module M = (val m) in
+        new M.cla
+    in
+    portable_use foo
 [%%expect{|
-val foo : unit -> <  > = <fun>
+val u : unit = ()
 |}]
 
 (* objects are always legacy *)
-let obj @ portable = new cla
+let u =
+    let obj = new cla in
+    portable_use obj
 (* CR zqian: this should fail. *)
 [%%expect{|
-val obj : cla = <obj>
+val u : unit = ()
 |}]
--- a/ocaml/testsuite/tests/typing-modes/class.ml
+++ b/ocaml/testsuite/tests/typing-modes/class.ml
@@ -72,6 +72,19 @@ Line 4, characters 30-31:
 Error: This value is nonportable but expected to be portable.
 |}]
 
+(* values written to instance variables need to be legacy *)
+class cla = object
+    val mutable x = "hello"
+
+    method foo = x <- (world : _ @@ local)
+end
+[%%expect{|
+Line 4, characters 22-42:
+4 |     method foo = x <- (world : _ @@ local)
+                          ^^^^^^^^^^^^^^^^^^^^
+Error: This value escapes its region.
+|}]
+
 class cla = object
     method m = "hello"
 end
--- a/ocaml/testsuite/tests/typing-modes/lazy.ml
+++ b/ocaml/testsuite/tests/typing-modes/lazy.ml
@@ -3,10 +3,12 @@
 *)
 
 (* lazy expression is legacy *)
-let x @ portable = lazy "hello"
+let u =
+    let _x @ portable = lazy "hello" in
+    ()
 (* CR zqian: this should fail. *)
 [%%expect{|
-val x : string lazy_t = lazy "hello"
+val u : unit = ()
 |}]
 
 (* lazy body is legacy *)
--- a/ocaml/testsuite/tests/typing-modes/lazy.ml
+++ b/ocaml/testsuite/tests/typing-modes/lazy.ml
@@ -1,0 +1,21 @@
+(* TEST
+    expect;
+*)
+
+(* lazy expression is legacy *)
+let x @ portable = lazy "hello"
+(* CR zqian: this should fail. *)
+[%%expect{|
+val x : string lazy_t = lazy "hello"
+|}]
+
+(* lazy body is legacy *)
+let x = lazy ("hello" : _ @@ local)
+[%%expect{|
+Line 1, characters 13-35:
+1 | let x = lazy ("hello" : _ @@ local)
+                 ^^^^^^^^^^^^^^^^^^^^^^
+Error: This value escapes its region.
+|}]
+
+(* Forcing lazy gives legacy, but that's in stdlib and not compiler *)
--- a/ocaml/testsuite/tests/typing-modes/module.ml
+++ b/ocaml/testsuite/tests/typing-modes/module.ml
@@ -14,17 +14,36 @@ module M = struct
     type 'a t = int
     let x _ = ()
 end
+module F (X : S) = struct
+    let x = X.x
+end
 [%%expect{|
 val portable_use : 'a @ portable -> unit = <fun>
 module type S = sig val x : 'a -> unit end
 module type SL = sig type 'a t end
 module M : sig type 'a t = int val x : 'a -> unit end
+module F : functor (X : S) -> sig val x : 'a -> unit end
 |}]
 
 (* Closing over modules affects closure's modes *)
 let u =
     let foo () =
         let _ = (module M : S) in
+        ()
+    in
+    portable_use foo
+(* CR zqian: This should fail *)
+[%%expect{|
+val u : unit = ()
+|}]
+
+let u =
+    let foo () =
+        let module X = struct
+            let x _ = ()
+        end
+        in
+        let module R = F(X) in
         ()
     in
     portable_use foo
--- a/ocaml/testsuite/tests/typing-modes/module.ml
+++ b/ocaml/testsuite/tests/typing-modes/module.ml
@@ -6,18 +6,19 @@
 
 let portable_use : 'a @ portable -> unit = fun _ -> ()
 
-module type S = sig val x : string end
+module type S = sig val x : 'a -> unit end
 
-module type Empty = sig end
+module type SL = sig type 'a t end
 
 module M = struct
-    let x = "string"
+    type 'a t = int
+    let x _ = ()
 end
 [%%expect{|
 val portable_use : 'a @ portable -> unit = <fun>
-module type S = sig val x : string end
-module type Empty = sig end
-module M : sig val x : string end
+module type S = sig val x : 'a -> unit end
+module type SL = sig type 'a t end
+module M : sig type 'a t = int val x : 'a -> unit end
 |}]
 
 (* Closing over modules affects closure's modes *)
@@ -36,7 +37,7 @@ val u : unit = ()
 separately. *)
 let u =
     let foo () =
-        let _ = (module List : Empty) in
+        let _ = (module List : SL) in
         ()
     in
     portable_use foo
@@ -75,7 +76,7 @@ val u : unit = ()
 
 let u =
     let foo () =
-        let m = (module struct let x = "hello" end : S) in
+        let m = (module struct let x _ = () end : S) in
         let module M = (val m) in
         M.x
     in
@@ -85,15 +86,15 @@ val u : unit = ()
 |}]
 
 (* first class modules are produced at legacy *)
-let x = ((module M : Empty) : _ @@ portable)
+let x = ((module M : SL) : _ @@ portable)
 (* CR zqian: this should fail *)
 [%%expect{|
-val x : (module Empty) = <module>
+val x : (module SL) = <module>
 |}]
 
 (* first class modules are consumed at legacy *)
 let foo () =
-    let m @ local = (module M : Empty) in
+    let m @ local = (module M : SL) in
     let module M = (val m) in
     ()
 [%%expect{|
--- a/ocaml/testsuite/tests/typing-modes/module.ml
+++ b/ocaml/testsuite/tests/typing-modes/module.ml
@@ -4,6 +4,8 @@
 
 (* This file tests that modules are sound wrt modes. *)
 
+let portable_use : 'a @ portable -> unit = fun _ -> ()
+
 module type S = sig val x : string end
 
 module type Empty = sig end
@@ -12,28 +14,35 @@ module M = struct
     let x = "string"
 end
 [%%expect{|
+val portable_use : 'a @ portable -> unit = <fun>
 module type S = sig val x : string end
 module type Empty = sig end
 module M : sig val x : string end
 |}]
 
 (* Closing over modules affects closure's modes *)
-let (foo @ portable) () =
-    let _ = (module M : S) in
-    ()
+let u =
+    let foo () =
+        let _ = (module M : S) in
+        ()
+    in
+    portable_use foo
 (* CR zqian: This should fail *)
 [%%expect{|
-val foo : unit -> unit = <fun>
+val u : unit = ()
 |}]
 
 (* File-level modules are looked up differently and need to be tested
 separately. *)
-let (foo @ portable) () =
-    let _ = (module List : Empty) in
-    ()
+let u =
+    let foo () =
+        let _ = (module List : Empty) in
+        ()
+    in
+    portable_use foo
 (* CR zqian: this should fail *)
 [%%expect{|
-val foo : unit -> unit = <fun>
+val u : unit = ()
 |}]
 
 (* Values in modules are defined as legacy *)
@@ -48,24 +57,31 @@ Error: This value escapes its region.
 |}]
 
 (* Values from modules are available as legacy *)
-let (foo @ portable) () = M.x
+let u =
+    let foo () = M.x in
+    portable_use foo
 (* CR zqian: this should fail *)
 [%%expect{|
-val foo : unit -> string = <fun>
+val u : unit = ()
 |}]
 
-let (foo @ portable) () = List.length
+let u =
+    let foo () = List.length in
+    portable_use foo
 (* CR zqian: this should fail *)
 [%%expect{|
-val foo : unit -> 'a list -> int = <fun>
+val u : unit = ()
 |}]
 
-let (foo @ portable) () =
-    let m = (module struct let x = "hello" end : S) in
-    let module M = (val m) in
-    M.x
+let u =
+    let foo () =
+        let m = (module struct let x = "hello" end : S) in
+        let module M = (val m) in
+        M.x
+    in
+    portable_use foo
 [%%expect{|
-val foo : unit -> string = <fun>
+val u : unit = ()
 |}]
 
 (* first class modules are produced at legacy *)
--- a/ocaml/testsuite/tests/typing-modes/module.ml
+++ b/ocaml/testsuite/tests/typing-modes/module.ml
@@ -1,0 +1,88 @@
+(* TEST
+   expect;
+*)
+
+(* This file tests that modules are sound wrt modes. *)
+
+module type S = sig val x : string end
+
+module type Empty = sig end
+
+module M = struct
+    let x = "string"
+end
+[%%expect{|
+module type S = sig val x : string end
+module type Empty = sig end
+module M : sig val x : string end
+|}]
+
+(* Closing over modules affects closure's modes *)
+let (foo @ portable) () =
+    let _ = (module M : S) in
+    ()
+(* CR zqian: This should fail *)
+[%%expect{|
+val foo : unit -> unit = <fun>
+|}]
+
+(* File-level modules are looked up differently and need to be tested
+separately. *)
+let (foo @ portable) () =
+    let _ = (module List : Empty) in
+    ()
+(* CR zqian: this should fail *)
+[%%expect{|
+val foo : unit -> unit = <fun>
+|}]
+
+(* Values in modules are defined as legacy *)
+module M = struct
+    let x = local_ "hello"
+end
+[%%expect{|
+Line 2, characters 8-9:
+2 |     let x = local_ "hello"
+            ^
+Error: This value escapes its region.
+|}]
+
+(* Values from modules are available as legacy *)
+let (foo @ portable) () = M.x
+(* CR zqian: this should fail *)
+[%%expect{|
+val foo : unit -> string = <fun>
+|}]
+
+let (foo @ portable) () = List.length
+(* CR zqian: this should fail *)
+[%%expect{|
+val foo : unit -> 'a list -> int = <fun>
+|}]
+
+let (foo @ portable) () =
+    let m = (module struct let x = "hello" end : S) in
+    let module M = (val m) in
+    M.x
+[%%expect{|
+val foo : unit -> string = <fun>
+|}]
+
+(* first class modules are produced at legacy *)
+let x = ((module M : Empty) : _ @@ portable)
+(* CR zqian: this should fail *)
+[%%expect{|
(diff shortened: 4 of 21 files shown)


lpw25:
In recent discussions, we've switched from `sync` to `portable` and `unsync` to `nonportable`, using "portability" for the axis as a whole. I think you should make that change to this PR.

goldfirere:
Is this ready for review?x

riaqn:
Yes it is, thanks!

goldfirere (commented):
I reviewed the tests a few days ago, but I've now decided to prioritize this after finishing the design doc for kind polymorphism, as the latter is holding up @liam923's ability to implement that feature. Hope to return around the end of the week, but someone else (e.g. @lpw25) might get to it first.

ocaml/testsuite/tests/typing-modes/portable-contend.ml
```diff
@@ -0,0 +1,249 @@
...
+Line 1, characters 26-27:
+1 | let foo (r @ contended) = r.a
+                              ^
+Error: Found a contended value where a uncontended value was expected
+|}]
+
+(* reading immutable field from contended record is fine *)
+let foo (r @ contended) = r.b
+[%%expect{|
+val foo : r @ contended -> string @ contended = <fun>
```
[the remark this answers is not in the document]
riaqn: Yes, a better test should use a generic `'a` instead of string.

ocaml/stdlib/array.ml
```diff
@@ -23,7 +23,7 @@ type 'a t = 'a array
 
 (* Array operations *)
 
-external length : 'a array -> int = "%array_length"
+external length : 'a array @ contended -> int = "%array_length"
```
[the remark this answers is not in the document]
riaqn: `a -> b` without modes means legacy modes, which is `uncontended`. My mental model of `length` is that it only reads  an immutable section in the metadata of the array, hence doesn't require `uncontended` access to the array.

> Are there other similar functions that should change?

There might - but it's sound to forget that. We can lazily fix those when needed. Most data types cross `contention` anyway.

ocaml/testsuite/tests/typing-modes/portable-contend.ml
```diff
@@ -0,0 +1,249 @@
...
+[%%expect{|
+Line 1, characters 4-27:
+1 | let r @ contended = "hello"
+        ^^^^^^^^^^^^^^^^^^^^^^^
+Error: Found a contended value where a uncontended value was expected
+|}]
+
+let x @ portable = "world"
+
+let y @ portable = x
```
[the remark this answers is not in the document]
riaqn: Good point. What's happening here is that, we are at top-level, which enables a special behaviour:
- When type checking a module NOT top-level, we type check all the values without mode restriction, and in the end requiring them to be submode of `legacy`. That is, if we put this example inside a `module`, there won't be any error at all, because both `x` and `y` will be `portable` which is submode of `nonportable` as required.
- When type checking at top-level, we type check all the values at `legacy`. That means the `x` in the example will be required to be `nonportable`, and made available in the environment at `nonportable`, which causes the error.

I'm not 100% sure about why this top-level behaviour, some comment about that in `typemod.ml` (search for `force_toplevel`).

ocaml/testsuite/tests/typing-modes/portable-contend.ml
```diff
@@ -0,0 +1,249 @@
...
+Error: Found a nonportable value where a portable value was expected
+|}]
+
+(* Closing over reading immutable field is OK *)
+let foo () =
+    let r @ portable = {a = "foo"; b = "bar"} in
+    let bar () = let _ = r.b in () in
+    let _ @ portable = bar in
+    ()
+(* CR zqian: currently mutable(legacy) means all records constructed are nonportable,
```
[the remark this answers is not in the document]
riaqn: Currently `mutable` means `mutable(legacy)` which is `mutable(global, nonportable)`, which means upon constructing a mutable record, we take into account writing `global, nonportable` values into the record in the future. That means the constructed record must be higher than `global, nonportable`.

ocaml/testsuite/tests/typing-modes/portable-contend.ml
```diff
@@ -0,0 +1,249 @@
...
+[%%expect{|
+Line 2, characters 23-45:
+2 |     let r @ portable = {a = "foo"; b = "bar"} in
+                           ^^^^^^^^^^^^^^^^^^^^^^
+Error: Found a nonportable value where a portable value was expected
+|}]
+
+
+(* TESTING arrays *)
+(* reading/writing to array requires uncontended *)
```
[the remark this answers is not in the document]
riaqn: It is boring now indeed as you said. But it will become less boring in the future as we mode cross most data types on `contention`, and these tests ensures that we haven't done too much to the point of unsoundness.

ocaml/testsuite/tests/typing-modes/portable-contend.ml
```diff
@@ -0,0 +1,249 @@
...
+let foo (r @ contended) =
+    match r with
+    | [| x; y |] -> ()
+[%%expect{|
+Line 3, characters 6-16:
+3 |     | [| x; y |] -> ()
+          ^^^^^^^^^^
+Error: Found a contended value where a uncontended value was expected
+|}]
+
```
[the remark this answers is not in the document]
riaqn: immutable arrays are discussed below. I'm resolving this conversion in favour of that one.

ocaml/testsuite/tests/typing-modes/portable-contend.ml
```diff
@@ -0,0 +1,249 @@
...
+
```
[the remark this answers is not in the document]
riaqn: Yes that's true. Currently we have `mutable(uncontended)` which mean we expect new field value to be `uncontended`.

ocaml/testsuite/tests/typing-modes/portable-contend.ml
```diff
@@ -0,0 +1,249 @@
...
+[%%expect{|
+val foo : int -> int @ contended -> unit = <fun>
+|}]
+
+(* TESTING immutable array *)
+module Iarray = Stdlib__Iarray
+
+let foo (r @ contended) = Iarray.get r 42
+(* CR zqian: The following should pass. let's not change [iarray.mli] to accept
+contended iarray; instead, let the modal kind system to mode cross iarray. *)
```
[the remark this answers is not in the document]
riaqn: Good point. I imagine `iarray` will implicitly carry a `contended` modality, which means `'a iarray` will cross `contention`.
It's element will always be `contended`. If the element type is `string`, it will cross to `uncontended`; if the element type is `int array`, it won't.

ocaml/testsuite/tests/typing-modes/portable-contend.ml
```diff
@@ -0,0 +1,249 @@
...
+
+
+(* Closing over measuring length doesn't force nonportable *)
+let foo () =
+    let r @ portable = [| "hello"; "world" |] in
+    let bar () = Array.length r in
+    let _ @ portable = bar in
+    ()
+(* CR zqian: The follows fails, because currently all records are constructed as
+    nonportable. Once we support mutable() syntax, we can test this. *)
```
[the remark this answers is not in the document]
riaqn: This test wants to show that closing over `contended` access to an `uncontended` value does not force `nonportable`. But we are not able to show that, because `r` is `nonportable` which  `bar` also closes over and are forced `nonportable`.

Array carrys an implicit `mutable(nonportable)`, and as a result, they can only be constructed as `nonportable`.

ocaml/testsuite/tests/typing-modes/portable-contend.ml
```diff
@@ -0,0 +1,249 @@
...
+
+(* Reading/writing mutable field from contended record is rejected. Also note
+    that the mutation error precedes type error. *)
+let foo (r @ contended) = r.a <- 42
+[%%expect{|
+type r = { mutable a : string; b : string; }
+Line 7, characters 26-27:
+7 | let foo (r @ contended) = r.a <- 42
+                              ^
+Error: Found a contended value where a uncontended value was expected
```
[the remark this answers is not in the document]
riaqn: I have added hint, which is not exactly what you suggested. I'm resolving this conversation - please comment on the new tests if needed. Thanks!

ocaml/testsuite/tests/typing-modes/portable-contend.ml
```diff
@@ -0,0 +1,249 @@
...
+val foo : r @ contended -> string @ contended = <fun>
+|}]
+
+(* Force top level to be uncontended and nonportable *)
+let r @ contended = "hello"
+[%%expect{|
+Line 1, characters 4-27:
+1 | let r @ contended = "hello"
+        ^^^^^^^^^^^^^^^^^^^^^^^
+Error: Found a contended value where a uncontended value was expected
```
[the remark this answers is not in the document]
riaqn: This is a trickier case. There are two related error messages we can give:
- Top-level values must be legacy. I agree that this should be hinted, but the current mode hinting framework (or the lack of) make this hard. I plan to improve that in a not-so-distant future, based on which this should be trivial to do.

- Module contents must be legacy. We will very soon allow module contents to carry modalities, and for monadic axes, there is simply no constraint between the module and it's contents (i.e. module can be `contended` while contents `uncontended`, and vice versa), so this error will go away. There will be errors when the value is `contended` but user didn't write `contended` modality in the module type, but the error message for that is another problem out of the scope of this PR.

ocaml/stdlib/array.ml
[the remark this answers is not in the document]
goldfirere: I'd rather not touch the stdlib here. If a test needs this, then I would just redefine it in the context of the test.

ocaml/testsuite/tests/typing-modes/portable-contend.ml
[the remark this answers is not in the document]
goldfirere: What about `bytes`? It's like string, but mutable, and so it won't cross contention.

ocaml/testsuite/tests/typing-modes/portable-contend.ml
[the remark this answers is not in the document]
goldfirere: That code seems like it has more to do with locality than modes in general. It looks like this can be fixed.

ocaml/testsuite/tests/typing-modes/portable-contend.ml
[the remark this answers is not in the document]
goldfirere: Yes, of course.

ocaml/testsuite/tests/typing-modes/portable-contend.ml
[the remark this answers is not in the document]
goldfirere: Yes.... but we should still test that we can't put contended values into an array, I think. It's possible that the check for arrays is different than, say, the check for mutable fields.

ocaml/testsuite/tests/typing-modes/portable-contend.ml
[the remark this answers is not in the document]
goldfirere: Why? If I have an `iarray` of some record with mutable fields, why should they suddenly become contended? I just want `iarray` to mode-cross contention whenever the elements do -- just like other immutable data. We don't have this yet, because we don't have kind polymorphism. Soon.

ocaml/testsuite/tests/typing-modes/portable-contend.ml
```diff
@@ -29,6 +29,21 @@ Error: This value is contended but expected to be uncontended.
   this record needs to be uncontended.
 |}]
 
+let foo (r @ contended) = {r with a = "hello"}
+[%%expect{|
+val foo : r @ contended -> r @ contended = <fun>
```
goldfirere, line 34: The output shouldn't be contended once `string` crosses modes.

ocaml/typing/mode.ml
```diff
@@ -1314,6 +1446,25 @@ module Uniqueness = struct
   let zap_to_legacy = zap_to_ceil
 end
 
+module Contention = struct
+  module Const = C.Contention
+
+  module Obj = struct
+    type const = Const.t
+
+    (* the negation of Uniqueness_op gives us the proper uniqueness *)
```
goldfirere, line 1455:
```suggestion
    (* the negation of Contention_op gives us the proper uniqueness *)
```

ocaml/typing/mode.ml
```diff
@@ -1370,11 +1521,14 @@ module Comonadic_with_regionality = struct
   let submode_log m0 m1 ~log : _ result =
     match submode_log m0 m1 ~log with
     | Ok () -> Ok ()
-    | Error { left = reg0, lin0; right = reg1, lin1 } ->
+    | Error { left = reg0, lin0, syn0; right = reg1, lin1, syn1 } ->
```
goldfirere, line 1524:
```suggestion
    | Error { left = reg0, lin0, port0; right = reg1, lin1, port1 } ->
```
and elsewhere

goldfirere (commented):
Almost done! Still have typecore and typemode to do, but that's it. Will happen tomorrow morning.

ocaml/stdlib/array.ml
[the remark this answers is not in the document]
riaqn: May I ask the conservation about stdlib? Note that we already modified stdlib with `[@local_opt]` anyway. Also with the removal of the upstream build, it matters even less now. So it's nice to ship with the compiler a stdlib that's "enriched".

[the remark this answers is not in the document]
goldfirere: I can think of a few reasons to leave this alone:

1. It could potentially pose problems for code that must work both within Jane Street and outside of JS, including stuff we open-source, and stuff we import from the outside world. The two stdlibs will be different, and so it's conceivable that code that works in one place will fail in the other.
2. Updating just a few, arbitrary functions in the stdlib to have modes isn't actually helpful. If e.g. Base can't rely on a fully updated stdlib, the folks updating Base will have to consult each function individually, anyway.
3. It makes upstreaming harder. I expect a nice debate to crop up when we start upstreaming this stuff -- especially around libraries, where the design space is much larger. By not touching the stdlib (at all), we can have the debate about language features separate from library features.

So, mostly I see costs to making changes to the stdlib, while the only benefit is that a few tests can avoid writing some `external`s.

An alternative I would be happy with is a module in the right otherlib (by "right", I mean the otherlib that corresponds to the relevant extension level, which for this is `stable`) which could shadow the stdlib module and is fully updated for modes. Maybe it would be OK even not fully updated. But our new otherlibs situation allows us to have functions in the compiler that work with our new features without touching the stdlib.

ocaml/testsuite/tests/typing-modes/portable-contend.ml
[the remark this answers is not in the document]
riaqn: Yes in retrospect, if the kind system is clever enough to infer that, that would be the best.

ocaml/testsuite/tests/typing-modes/portable-contend.ml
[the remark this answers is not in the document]
riaqn: I think the reason for the special treatment is simply that top-level phrases are type-checked loosely and differently, and it's  safe and simple to assume everything is legacy.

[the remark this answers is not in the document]
riaqn: But maybe we shouldn't test top-level at all - it's rarely used. I will put both `x` and `y` into a module, which is worth testing.

ocaml/typing/typecore.ml
goldfirere, line 1: (Not really about typecore, but this needs to become a Discussion so that we do not forget about it.)

I think there needs to be more attention paid to classes, sadly.

* In experiments, I was able to put a `nonportable` thing in an instance variable in a class `c`. But then calling `new c` can produce a portable object. This is bad.
* I see nothing here that checks for portability/contention around mutable instance variables. This could be used to sneak a nonportable thing across a thread barrier.

goldfirere, line 1: I think `lazy` always needs to be nonportable. Otherwise, I could create a `lazy (print_endline "hello")`, send that to some other thread, and then evaluate it twice (because whether or not the thunk has been evaluated is a mutable cell).

goldfirere, line 1: I think first-class modules always need to be nonportable, because modules are nonportable. Right now, I could write

```
module type S = sig val set : int -> unit end
let global_ref = ref 0
module M = struct let set x = global_ref.contents <- x end

let () =
  let x @ portable = (module M : S) in
  ()
```

which seems bad.

goldfirere: Specifically, I expected to see some code around `Pexp_new`, `Pexp_setinstvar`, `Pexp_override`, and `Pexp_object`.

goldfirere (commented):
Review complete, finally.

ocaml/testsuite/tests/typing-modes/portable-contend.ml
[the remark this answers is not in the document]
goldfirere: Is this testing the top-level of the REPL? Or is this treatment true of type-level definitions in any file? I think it's probably the former (and then I agree -- though we should document), but I wanted to double-check.

ocaml/testsuite/tests/typing-modes/portable-contend.ml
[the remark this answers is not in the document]
goldfirere: Both, actually. Because the mode of the array must be the same as the mode of its contents, we shouldn't be able to put any contended thing into an array. Unless I've misunderstood.

ocaml/typing/typecore.ml
riaqn: I believe some of the concerns are addressed by https://github.com/ocaml-flambda/flambda-backend/pull/2431
The code there are orthogonal to the code here, hence a separate PR. But the fixes are only observable for the portability/contention axes, and therefore is based on this PR.

riaqn: This is a general problem wrt lazy vs. modes; let's fix it in a separate PR after this PR.

riaqn: This is a general problem wrt FCM vs. modes; let's fix it in a separate PR after this PR.

ocaml/testsuite/tests/typing-modes/portable-contend.ml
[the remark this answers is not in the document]
riaqn: Yes I believe those `expect` tests are backed by top-level of REPL, not file-level modules.

I suggest that we test both, because some people do use REPL sometimes.

ocaml/typing/typecore.ml
```diff
@@ -5572,7 +5605,7 @@ and type_expect_
           None, expected_mode
       in
       let type_label_exp ((_, label, _) as x) =
-        check_construct_mutability label.lbl_mut argument_mode;
+        check_construct_mutability ~loc ~env label.lbl_mut argument_mode;
```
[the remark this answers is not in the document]
riaqn: Me and @lpw25 had a discussion again last week, the outcome of which has been incorprated to the document. To properly implement that would require #2344. Also, I think "mode vs. mutation" is orthogonal to this PR and should be fixed separately.

Of course those things are still in the future, and the question is whether or not it's safer/better to make the changes you suggested **_now_**, to which I think it's no. Currently there is just one modality `Global` which secretly entails `Many` and `Shared`. It would be messy to base any type-checking optimization on top of that.

ocaml/typing/typecore.ml
goldfirere: I think we decided that this PR would add test cases we expect to change for this and other cases. Then later PRs would fix the tests. Does that sound right?

riaqn: Yes that's exactly right. I will do that.

ocaml/typing/typecore.ml
```diff
@@ -5634,8 +5667,9 @@ and type_expect_
                   unify_exp_types loc env ty_arg1 ty_arg2;
                   with_explanation (fun () ->
                     unify_exp_types loc env (instance ty_expected) ty_res2);
+                  check_project_mutability ~loc:exp.exp_loc ~env lbl.lbl_mut mode;
                   let mode = modality_unbox_left lbl.lbl_global mode in
-                  check_construct_mutability lbl.lbl_mut argument_mode;
+                  check_construct_mutability ~loc ~env lbl.lbl_mut argument_mode;
```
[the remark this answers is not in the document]
riaqn: This is similar to the resolved conversation above, so I will resolve this.

ocaml/testsuite/tests/typing-modes/portable-contend.ml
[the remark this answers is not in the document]
riaqn: Mutating array is only via stdlib, which we are not gonna modify. Do we still test it?

ocaml/stdlib/array.ml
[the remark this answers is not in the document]
riaqn: OK if we already think `otherlibs` is the place for enriched libraries, then I shouldn't touch stdlib.
I will leave the work in `otherlibs` to future PRs.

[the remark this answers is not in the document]
riaqn: I will also remove those tests that relies on the changed stdlib, because they are trivial, and not worth having `external` for.

ocaml/testsuite/tests/typing-modes/portable-contend.ml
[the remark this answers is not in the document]
riaqn: Excellent choice sir. Fixed.

ocaml/testsuite/tests/typing-modes/portable-contend.ml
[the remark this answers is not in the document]
riaqn: In the end, I keep those tests unchanged. Because the fact that "values in a module is always legacy" is tested separately by `module.ml` already.

ocaml/testsuite/tests/typing-modes/class.ml
```diff
@@ -0,0 +1,149 @@
+(* TEST
```
goldfirere, line 1: I was expecting to see some tests around the mutability mode of instance variables. It's definitely always going to be the legacy mode. But we still need to be careful: we must check that anything written into an instance variable is uncontended, because -- unusually -- we can write to a *contended* instance variable. So I think a corner case exists here that doesn't exist elsewhere. Getting it wrong would allow us to smuggle a contended value into an uncontended one by storing it in an instance variable -- bad.

ocaml/testsuite/tests/typing-modes/module.ml
```diff
@@ -0,0 +1,88 @@
...
+    let x = "string"
+end
+[%%expect{|
+module type S = sig val x : string end
+module type Empty = sig end
+module M : sig val x : string end
+|}]
+
+(* Closing over modules affects closure's modes *)
+let (foo @ portable) () =
```
goldfirere, line 21: As above, I think this test would be more effective in a local variable.

ocaml/testsuite/tests/typing-modes/module.ml
```diff
@@ -0,0 +1,88 @@
...
+    let _ = (module M : S) in
+    ()
+(* CR zqian: This should fail *)
+[%%expect{|
+val foo : unit -> unit = <fun>
+|}]
+
+(* File-level modules are looked up differently and need to be tested
+separately. *)
+let (foo @ portable) () =
```
goldfirere, line 31: Also should be a local variable

ocaml/testsuite/tests/typing-modes/module.ml
```diff
@@ -0,0 +1,88 @@
...
+    let _ = (module List : Empty) in
```
goldfirere, line 32: There's a tiny chance this will be interpreted like `unit` and mode-cross. (It would actually be sound!) So maybe use a module type other than `Empty`.

ocaml/testsuite/tests/typing-modes/module.ml
```diff
@@ -0,0 +1,88 @@
...
+    ()
+(* CR zqian: this should fail *)
+[%%expect{|
+val foo : unit -> unit = <fun>
+|}]
+
+(* Values in modules are defined as legacy *)
+module M = struct
+    let x = local_ "hello"
```
goldfirere, line 41: Similar tests for `once` and `contended`?

ocaml/testsuite/tests/typing-modes/module.ml
```diff
@@ -0,0 +1,88 @@
...
+end
+[%%expect{|
+Line 2, characters 8-9:
+2 |     let x = local_ "hello"
+            ^
+Error: This value escapes its region.
+|}]
+
+(* Values from modules are available as legacy *)
+let (foo @ portable) () = M.x
```
goldfirere, line 51: Please put in a local variable

goldfirere, line 51: This test will pass once `string` mode-crosses portability. Maybe make the module element a function to protect against that and keep the test robust?

ocaml/testsuite/tests/typing-modes/module.ml
```diff
@@ -0,0 +1,88 @@
...
+(* CR zqian: this should fail *)
+[%%expect{|
+val foo : unit -> string = <fun>
+|}]
+
+let (foo @ portable) () = List.length
```
goldfirere, line 57: Please put in a local variable

goldfirere, line 57: `List.length` will itself become portable someday. So it might be better to use a function which definitely won't be portable, like `print_endline` (?).

ocaml/testsuite/tests/typing-modes/module.ml
```diff
@@ -0,0 +1,88 @@
...
+(* CR zqian: this should fail *)
+[%%expect{|
+val foo : unit -> 'a list -> int = <fun>
+|}]
+
+let (foo @ portable) () =
+    let m = (module struct let x = "hello" end : S) in
```
goldfirere, line 64: Maybe put a non-portable function in there? That module might mode-cross someday, making this test not do the thing you want it to.

ocaml/testsuite/tests/typing-modes/module.ml
```diff
@@ -0,0 +1,88 @@
...
+val foo : unit -> string = <fun>
+|}]
+
+let (foo @ portable) () = List.length
+(* CR zqian: this should fail *)
+[%%expect{|
+val foo : unit -> 'a list -> int = <fun>
+|}]
+
+let (foo @ portable) () =
```
goldfirere, line 63: Please put in a local variable

ocaml/testsuite/tests/typing-modes/module.ml
```diff
@@ -0,0 +1,88 @@
...
+    let m = (module struct let x = "hello" end : S) in
+    let module M = (val m) in
+    M.x
+[%%expect{|
+val foo : unit -> string = <fun>
+|}]
+
+(* first class modules are produced at legacy *)
+let x = ((module M : Empty) : _ @@ portable)
```
goldfirere, line 72: Similar test for `unique`?

ocaml/testsuite/tests/typing-modes/module.ml
```diff
@@ -0,0 +1,88 @@
...
+(* CR zqian: this should fail *)
+[%%expect{|
+val x : (module Empty) = <module>
+|}]
+
+(* first class modules are consumed at legacy *)
+let foo () =
+    let m @ local = (module M : Empty) in
+    let module M = (val m) in
```
goldfirere, line 81: Similar tests for `once` and `uncontended`?

ocaml/testsuite/tests/typing-modes/module.ml
```diff
@@ -0,0 +1,88 @@
+(* TEST
```
goldfirere, line 1: I think there should also be some tests around functor application: functor application can cause mutation, and applying a functor in a function (even without first-class modules) should cause the function to close over both the functor and the argument, causing modes to change.

ocaml/testsuite/tests/typing-modes/lazy.ml
```diff
@@ -0,0 +1,21 @@
+(* TEST
+    expect;
+*)
+
+(* lazy expression is legacy *)
+let x @ portable = lazy "hello"
```
goldfirere, line 6: As with other `portable` tests, please put this test in a local variable

goldfirere, line 6: Similar test for `unique`?

ocaml/testsuite/tests/typing-modes/lazy.ml
```diff
@@ -0,0 +1,21 @@
...
+(* CR zqian: this should fail. *)
+[%%expect{|
+val x : string lazy_t = lazy "hello"
+|}]
+
+(* lazy body is legacy *)
+let x = lazy ("hello" : _ @@ local)
```
goldfirere, line 13: Similar tests for `once` and `contended`?

goldfirere (commented):
Review of recent commits

riaqn:
@goldfirere To reply to the series of the comments above:
The mode checking (existing ones and ones that I will add in future PRs) either treats all mode axes uniformly, or distinguish between monadic and comonadic axes. I believe tests should try to exhaust all branches in the type checker, and there is no benefit in testing the same code path mutliple times.

ocaml/testsuite/tests/typing-modes/class.ml
```diff
@@ -0,0 +1,149 @@
...
+[%%expect{|
+Line 4, characters 10-11:
+4 |     o#foo x
+              ^
+Error: This value is nonportable but expected to be portable.
+|}]
+
+
+(* Closing over classes affects closure's mode *)
+let (foo @ portable) () =
```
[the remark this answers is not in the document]
riaqn: The mode checking logic of top-level is "outer" than the `(foo @ portable)` logic, so the test works as intended. That being said, let me know if you still prefer a local definition.

ocaml/testsuite/tests/typing-modes/module.ml
riaqn: I think the test here expresses that all values in modules are legacy(nonportable), even ones that morally are portable. If we use `print_endline`, it gives the impression smarter than the mode checker actually is.

goldfirere:
I hear what you're saying about duplicate tests, but I'm not yet totally convinced. The tests I highlighted care about whether the legacy mode is the top or the bottom. That is, some tests only make sense when legacy is top, and others make sense only when legacy is bottom. What's surprising is that the legacy = top/bottom distinction doesn't line up with the monadic/comonadic distinction. We in fact have all the possibilities:

* Comonadic, legacy = top: portability
* Comonadic, legacy = bottom: locality, linearity
* Monadic, legacy = top: uniqueness
* Monadic, legacy = bottom: contention

I don't know the code for fixing all these tests. But much code around this cares pretty deeply about the monadic-vs-comonadic distinction, and so I think it's probably a good idea to test one monadic axis and one comonadic axis for each scenario. So I agree that we don't need to test both locality and linearity (though I suggest doing so above) in legacy=bottom scenarios. But we should test both locality (say) and contention, then.

riaqn:
In general, some axes have their legacy mode set in such a way that certain checks will always pass, and therefore boring to test. I've picked axes so that none of the tests are boring.

The thing that do care about the distinction is "closing over module/classes". For those, I've picked one non-boring axis from monadic, and one non-boring from comonadic, to show the constraints are working (by seeing a type error). 

Most checks don't care about the distinction between monadic and comonadic axes, and therefore there is no need to check twice of the same thing.

goldfirere:
OK. I'll back down on that point.

ocaml/testsuite/tests/typing-modes/class.ml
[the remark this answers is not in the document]
riaqn: In my current mental model, I try not to talk about the modes of "instance variable". Sure, one can comtemplate the runtime representation of an object, which probably contains pointers to instance variables, based on which the meaning of "contended instance variables" can be derived.

But we are approaching this more coarsely: assuming the upstream ocaml (with everything legacy) is correct, we think our class/module is _probably_ also correct, if 
- we don't change the type checking of class/module; don't track their modes.
- When they have to interact with modes, it must be legacy.

riaqn: Similar to above: the type checker doesn't even talk about the `mutable(m0)` of instance variables.

riaqn: Similar to above, the type checker doesn't talk about the modes of instance variables.

But you are right that we should test that it's disallowed to write non-legacy value into a instance variable.

ocaml/testsuite/tests/typing-modes/portable-contend.ml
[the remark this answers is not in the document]
riaqn: I have checked and they are correct.

ocaml/testsuite/tests/typing-modes/module.ml
riaqn: I agree with what you said, but think it's better to decompose:
- Functor application can cause mutation, this is reflected by the fact that functors (like all modules) are `nonportable`.
- Closing over `F(M)` will constrain modes. But really from the perspective of the implementation, we are closing over two modules `F` and `M`. 

In summary, the logic needed is already there for regular modules. That being said, I will still add the tests as integration tests.

riaqn: Well in fact I suggest we don't test this at all. If the test involves `F(X)`, we can't tell if we are testing `F` or `X`. If the test is about `F` alone, then it doesn't really seem like a new test to me. That is, `F` alone is obviously treated the same as `X` alone.

ocaml/testsuite/tests/typing-modes/portable-contend.ml
```diff
@@ -0,0 +1,249 @@
...
+[%%expect{|
+Line 4, characters 23-26:
+4 |     let _ @ portable = bar in
+                           ^^^
+Error: Found a nonportable value where a portable value was expected
+|}]
+
+(* Closing over reading immutable field is OK *)
+let foo () =
+    let r @ portable = {a = "foo"; b = "bar"} in
```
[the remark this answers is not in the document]
riaqn: Note the CR below - this test should pass but fails currently. The `@ portable` is added so it fails early, to make the point made in the CR clearer.

ocaml/testsuite/tests/typing-modes/class.ml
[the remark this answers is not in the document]
goldfirere: (Addressed in the comment below. Resolving this one.)

goldfirere: Yes, I think that would capture the kind of thing I'm worried about here. We need to make sure everything that comes out of a class is legacy, *and* we need to make sure that everything that goes in is legacy, too.

ocaml/testsuite/tests/typing-modes/module.ml
goldfirere: There's something not quite right here. Suppose I have

```ocaml
let mylength list = List.length list
```

Does this close over `List`? If so, then because all modules are nonportable, `mylength` should be nonportable, too. Yet I don't think that's the case -- nor should it be. When we're projecting out of a module (as we almost always do), we can just look at the projection's modality (soon, we'll allow `List.length` to be portable). So, how do we have a function close over a module? We could use a first-class module to capture an entire module... but I don't want to test first-class modules here. It seems to me the only other way is to do a functor application, with `let module M = F(X) in ...`. It's true that this also says something about `X`, but I don't see a way around that.

My bottom line: something is different about functor application than just projecting from a module, because functor application is actually effectful. And I think that capturing that effect should somehow be reflected in our tests.

ocaml/testsuite/tests/typing-modes/portable-contend.ml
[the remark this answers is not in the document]
goldfirere: Are there checked-in tests that reflect this?

ocaml/testsuite/tests/typing-modes/module.ml
riaqn: Good question. @lpw25 and I had a previous discussion about this, and think `mylength` doesn't close over `List`. See the section "Closing over values in modules" in the internal document "modes and modules".

ocaml/testsuite/tests/typing-modes/portable-contend.ml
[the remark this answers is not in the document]
riaqn: The topic in question is about modes vs. mutation in general, which is implemented by #2369 . However, in that PR, mutation implies global many shared (for compatibility), which makes the system boring and no extra tests were added.

This PR introduces new axes which makes the `mutable` logic testable. I will add tests.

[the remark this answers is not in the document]
riaqn: In fact would you mind I create a JIRA ticket to track this instead of doing this in this PR? I don't want unrelated tests in this PR.

ocaml/testsuite/tests/typing-modes/module.ml
goldfirere: Right. I agree. But then there must be some test that can witness side effects that happen during functor application. This is why I wanted a test looking at `F(X)`, because I think that having a functor application in the middle of a function might affect modes differently than projecting values.

ocaml/testsuite/tests/typing-modes/portable-contend.ml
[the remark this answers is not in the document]
goldfirere: The test I’m looking for — that you can’t put contended things into an array — seems related to this patch, which introduces the contention axis. But I’m also fine with delaying this piece with an internal ticket.

ocaml/testsuite/tests/typing-modes/module.ml
riaqn: The fact that `F(X)` might cause mutation is modelled by `F` being nonportable. Functor application is not essential here: a function that captures this `F` without applying it will still be `nonportable`.

And it's same logic (in #2431 ) that handles closing over modules, whether it's a structure or a functor. Tests for the former already exists; I don't mind adding the latter just to be sure.

goldfirere: Yes, I agree that closing over `F` is the real issue. The problem is that the test that closes over `F` requires using a first-class module, which has its own set of mode restrictions, etc. My desire for functor application was a way of closing over `F` without invoking first-class modules.

riaqn: The problem with `F(X)` is that this also closes over `X`, which is treated similarly to `F`. So even if the closure's mode is forced as expected, we don't know if it's because of `F` or `X`. (Hence my second comment above).

goldfirere: I continue to think this test is worthwhile. I agree that we can't test exactly the thing we'd like -- closing over `F` without first-class modules or any other modules -- but a test that covers too much (both `F` and `X`) is better than none at all. And `X` can just be `struct end`! Or not even that: if `F` is a generative functor, we don't even need the `X`. (Though actually I think we should test both generative and applicative functors.)

riaqn: I suppose I can define `X` inside the closure, then it's not closed over `X`.

ocaml/testsuite/tests/typing-modes/portable-contend.ml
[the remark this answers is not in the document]
riaqn: Is this resolved?

ocaml/testsuite/tests/typing-modes/class.ml
```diff
@@ -72,6 +72,19 @@ Line 4, characters 30-31:
 Error: This value is nonportable but expected to be portable.
 |}]
 
+(* values written to instance variables need to be legacy *)
+class cla = object
+    val mutable x = "hello"
+
+    method foo = x <- (world : _ @@ local)
```
goldfirere, line 79: What’s `world`? Maybe it doesn’t matter for the test, but it’s a bit strange to use an out-of-scope variable here.

goldfirere (approved):

ocaml/testsuite/tests/typing-modes/class.ml
riaqn: Yeah it looks strange. I mean it kind of show that the mode is constrained before diving down the sub-expression. But yes it's confusing and fixed.

PR #1559, opened 2023-07-12
Zero alloc: warnings instead of errors and scoped attributes (DRAFT)

gretay-js:
It works as intended, I think, but `Lambda.check_attribute` and `Translattrib.add_check_attribute` needs cleaning, which will also hopefully reduce the diff in middle-end. Making a PR to discuss how to split this into manageable chunks for review.

PR #2400, opened 2024-03-27
Parse `zero_alloc` attributes into the typed tree

ccasin:
This adds `zero_alloc` attribute information to the typed tree, moving the responsibility for parsing said attributes from lambda to the typechecker.  This is purely refactoring - no behavior change is intended.

It is the first of three PRs that will add support for `zero_alloc` in signatures.  The second PR will allow writing `zero_alloc` on functions in signatures and enforce it in module inclusion checks (so, functions in modules with that signature must be labeled `zero_alloc` as well).  The third will allow this information to be used in `zero_alloc` checking of code that calls these functions.  There is an internal design doc for this feature I will provide reviewers.

This is split into four commits:
- The first adds more extensive warning 53 (misplaced attribute) tests for `zero_alloc`, and moves the existing tests to the upstream testsuite (since w53 does not rely on the middle-end or backend).  `zero_alloc` is a tricky case for w53 because, both before and after this PR, the compiler phase where we look for the attribute is far removed from the compiler phase where we actually do the `zero_alloc` checks.  This commit just documents the existing w53 behavior more thoroughly so we can be more confident it is not accidentally changed in future commits and future PRs.
- The second moves some definitions and files around so they will be accessible in typechecking.  In particular, we move `assume_info.ml` from `lambda` to `utils`, and move functionality for zero_alloc attribute parsing from `lambda/translattribute.ml` to `parsing/builtin_attribute.ml`.  Some infrastructure for attribute payload parsing more generally is also moved, since it is now used both in `Builtin_attribute` (for `zero_alloc`) and `Translattribute` (for other function attributes).
- The third commit does some interface simplification of the functions that process attributes, made possible by the moves in the previous commit.  This commit could be dropped if we thought a minimal change were better than a clean interface, but I think it's a clear improvement.
- The fourth commit does the move of the actual responsibility for parsing `zero_alloc` attributes, adding them directly to the typed tree rather than just hanging on to them as attributes.

`zero_alloc` attributes can appear in three places:
1) function applications (only `zero_alloc assume`)
2) anonymous function definitions
3) let bindings of functions

Moving the first and second case to the typed tree is very straightforward - they go on `Texp_apply` and `Texp_function` nodes, respectively.  For the last one, one could imagine recording the information either on the `value_binding` or on the `Texp_fucntion` node under it.

I've done the latter.  This choices makes the typed tree and its translation to lambda more uniform - we don't need to move `zero_alloc` from value bindings to functions during `transl` anymore, because it's done at typechecking time.  However, it does mean that when I implement the second PR in this series, there will be a bit more work to do, because we'll need to look at the body of the term in a `value_binding` to determine whether it is annotated with `zero_alloc`.  I've already implemented this and it's slightly unpleasant but ultimately works fine.

**Review**: @gretay-js should take a look at the new misplaced attributes test and confirm this PR is on the roadmap in the design doc, but I think a front-end dev should review the bulk of the refactor.  I think @ncik-roberts is a good candidate for this PR series - careful reasoning about function arity is important for the next two parts, and it probably makes sense for one reviewer to do all three.  I suggest reviewing commit by commit, as this both moves code around and changes it, but in separate commits.


ocaml/parsing/builtin_attributes.mli
```diff
@@ -199,3 +206,62 @@ val jkind : Parsetree.attributes -> jkind_attribute Location.loc option
...
+
+type check_attribute =
+  | Default_check
+  | Ignore_assert_all of property
+  | Check of { property: property;
+               strict: bool;
+               (* [strict=true] property holds on all paths.
+                  [strict=false] if the function returns normally,
+                  then the property holds (but property violations on
+                  exceptional returns or divering loops are ignored).
```
xclerc, line 251:
```suggestion
                  exceptional returns or diverging loops are ignored).
```

ocaml/typing/typedtree.mli
```diff
@@ -271,7 +273,8 @@ and expression_desc =
           saturated with n arguments.
       *)
   | Texp_apply of
-      expression * (arg_label * apply_arg) list * apply_position * Mode.Locality.l
+      expression * (arg_label * apply_arg) list * apply_position *
+        Mode.Locality.l * Assume_info.t
```
gretay-js, line 277: Instead of propagating two separate types `Builtin_attributes.check_attribute` and `Assume_info.t`  here, we can extend `Assume_info.t` to propagate `Check` info not only `Assume`. Is there a reason not to in this PR? I'm guessing that a subsequent PR will convert the `zero_alloc` field of  `Text_function` from `Builtin_attributes.checkattribute` to `Assume_info.t` to check inclusion anyway.

ocaml/lambda/translcore.ml
```diff
@@ -1549,16 +1547,16 @@ and transl_scoped_exp ~scopes sort expr =
...
 and transl_bound_exp ~scopes ~in_structure pat sort expr loc attrs =
-  let should_introduce_scope =
+  let should_introduce_scope, zero_alloc =
     match expr.exp_desc with
-    | Texp_function _ -> true
-    | _ when in_structure -> true
-    | _ -> false in
+    | Texp_function { zero_alloc } -> true, zero_alloc
+    | _ when in_structure -> true, Default_check
+    | _ -> false, Default_check in
```
gretay-js, line 1554: This is not obviously the same behavior: in `| _ when in_structure -> true` case we could have a non-empty "assume" that would have been propagated. Maybe it was a bug before? Or not feasible? 

Independently, it's a bit strange to use `check_allowed` here in transcore, because it is a parsing concern. We can convert to `Assume_info.t` earlier, in typecore.

gretay-js (changes requested):
- I like the move of attribute and payload parsing from `translattribute` to `Builtin_attributes`. 
 
- I think we should merge `assume_info.ml` into `zero_alloc_utils.ml` if there are no cyclic deps, now that they are in the same folder. I considered doing it in #2165 but kept `assume_info` next to `debuginfo`.  See comments about  `transl_bound_exp` and the new fields of `typedtree.ml`. It will also be easier to add "zero_alloc assert" info on expressions later. 

- recording `zero_alloc` information on `Texp_function` for `let` bindings of functions:

> makes the typed tree and its translation to lambda more uniform - we don't need to move zero_alloc from value bindings to functions during transl anymore, because it's done at typechecking time. However, it does mean that when I implement the second PR in this series, there will be a bit more work to do, because we'll need to look at the body of the term in a value_binding to determine whether it is annotated with zero_alloc. I've already implemented this and it's slightly unpleasant but ultimately works fine.

Previously, `Translcore.transl_function` had to "look-ahead" for `fun` so it can propagate `assume_zero_alloc` to the appropriate scopes, report the right warnings (and report them only once). That's why we had `with_warnings` and `mark_unused:false`. I'm glad to see it is no longer needed. Can you explain how it is handled now?

ocaml/testsuite/tests/warnings/w53.ml
```diff
@@ -473,3 +473,61 @@ module TestLayoutPolyStruct = struct
...
+
+  class[@zero_alloc] foo _y = (* rejected *)
+    let[@inline never][@zero_alloc] f x = (x, x) in (* accepted *)
+    (fun[@zero_alloc] z -> (* rejected *)
+    object
+      val[@zero_alloc] bar = (4, 5) (* rejected *)
+
+      method[@zero_alloc] baz x = (f (z+10), x+1) (* rejected *)
+    end)
+
```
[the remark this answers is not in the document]
ncik-roberts: I'm not excited about minting 2 more files for misplaced attribute tests. (There's already 2, for ppxes vs no-ppxes.)
* Eventually, could `[@@@zero_alloc all]` be scoped to the submodule it appears in, like `[@@@ocaml.warning ...]` attributes?
* For the time being, maybe we could just put the `[@@@zero_alloc all]` test last.

ocaml/parsing/builtin_attributes.mli
```diff
@@ -199,3 +206,52 @@ val jkind : Parsetree.attributes -> jkind_attribute Location.loc option
...
+               opt: bool;
+               loc: Location.t;
+             }
+  | Assume of { property: property;
+                strict: bool;
+                loc: Location.t;
+                never_returns_normally: bool;
+              }
+
+val is_check_enabled : opt:bool -> property -> bool
```
ncik-roberts, line 262: "check" and "property" are two names for one thing, used interchangeably in several of these bindings. I find both names to be too vague.

I wonder if this signature's names are just too general. There's a variant, `property`, with a single constructor `Zero_alloc`. Could we get rid of that variant and change these names to say `_zero_alloc`, or alternatively just make a `Zero_alloc` submodule?

It's not clear to me whether we will ever want to extend `property`. Maybe we do and my inclination is just wrong here.

ncik-roberts: (And I recognize you're just moving bindings around in this commit. But rehoming them is a fine enough excuse for me to bikeshed their names.)

ocaml/parsing/builtin_attributes.mli
```diff
@@ -199,3 +206,52 @@ val jkind : Parsetree.attributes -> jkind_attribute Location.loc option
...
+
+val get_property_attribute : Parsetree.attributes -> property -> check_attribute
+
+val get_assume_zero_alloc :
+  with_warnings:bool -> Parsetree.attributes -> Assume_info.t
```
ncik-roberts: Suggestion: `Assume_info` could be `Zero_alloc_assume_info`.

ocaml/parsing/builtin_attributes.mli
```diff
@@ -207,24 +207,34 @@ val jkind : Parsetree.attributes -> jkind_attribute Location.loc option
...
+    Parsetree.payload -> (bool option, unit) Result.t
+
+(** [parse_id_payload] is a helper for parsing information from an identifier
+    attribute payload. If the given payload consists of a single identifier,
+    that identifier is looked up in the association list.  The result is
+    returned, if it exists.  The [empty] value is used if the payload is empty.
+    Otherwise the [default] value is used an a warning is issued. *)
+val parse_id_payload :
+  string -> Location.t -> default:'a -> empty:'a -> (string * 'a) list ->
+  Parsetree.payload -> 'a
```
ncik-roberts, line 237: `get_int_payload` and `get_optional_bool_payload` have clear and simple documentation and types. I think we should keep them in `builtin_attributes.ml`. `parse_id_payload` is weirdly complex, and it seems less compelling to me to hoist it out to a more fundamental attribute-parsing library like `builtin_attributes.ml`. Could we keep it in `translattribute.ml`, closer to its use sites?

ocaml/parsing/builtin_attributes.mli
ccasin: I agree the attempt at generality here leads to code that is more confusing than it could be, and am open to eliminating the premise that there is more than one `property`, at least until there actually is.  But I presume @gretay-js made it this way for a reason, perhaps she can weigh in.

ocaml/parsing/builtin_attributes.mli
ccasin: `parse_ids_payload` (with an `s`!) is needed in this file for `parse_property_attribute`.  I moved `parse_id_payload` (without an `s`) because it's very similar to the `ids` version and needs similar helper functions that I otherwise have to expose in the `mli` here for only this purpose.  It seemed better to me to put the very similar `id` and `ids` versions together and avoid exposing the helpers, even though I agree `id` is a little weird in the mli.

I'm not super happy with this, but there's an explanation of the tradeoffs, anyway.  If you still think it's better to move it, I will.

ncik-roberts: Augh, I fold. I don't think it's better to move it. I will make some suggestions though:
  * "identifier attribute payload" is an unfamiliar term. Maybe "attribute whose payload is an identifier"?
  * `parse_optional_id_payload`? It is `optional`, that's what `empty` is for.
  * Get rid of `default` and instead return `Error ()`. `default` and `empty` are just too easy to conceptually confuse at the moment.

ccasin:
> recording zero_alloc information on Texp_function for let bindings of functions:
> 
> > makes the typed tree and its translation to lambda more uniform - we don't need to move zero_alloc from value bindings to functions during transl anymore, because it's done at typechecking time. However, it does mean that when I implement the second PR in this series, there will be a bit more work to do, because we'll need to look at the body of the term in a value_binding to determine whether it is annotated with zero_alloc. I've already implemented this and it's slightly unpleasant but ultimately works fine.
> 
> Previously, Translcore.transl_function had to "look-ahead" for fun so it can propagate assume_zero_alloc to the appropriate scopes, report the right warnings (and report them only once). That's why we had with_warnings and mark_unused:false. I'm glad to see it is no longer needed. Can you explain how it is handled now?

Previously, we used `~with_warnings:false` in two places to check if we needed to update the scope with assume info for something that might be a function, but before we had reached `Translattribute` where the main check for the attribute happened.  And it was important not to mark the attribute used in these cases, because the w53 logic relied on that happening in `Translattribute`.

I think there are two questions you might have:

1) How does lambda find the assume info and update the scopes now, without running into the same issue of masking w53

Now the parsing of this attribute happens way earlier, in typing.  By the time we reach the places where we used to set `~with_warnings:false`, we can simply look at the typed tree rather than touching the attributes at all.  You can check both of them to see where this information is available.  (You've noted elsewhere I might have the logic in one of those spots a little wrong - I'll respond to that separately there.)

2) In the second PR, this information is needed at the level of the let-binding to make it part of the module type, but in this PR it's only being stored on the function.  How do we get the information when translating the let binding without accidentally masking w53 if the attribute is put on non-function let bindings?

There is already a function in `typedtree.ml` which is used to traverse patterns and accumulate information about the pattern variables after typechecking a let-binding, called [`let_bound_idents_with_modes_and_sorts`](https://github.com/ocaml-flambda/flambda-backend/blob/b498eadc5b927307daaa5890a49dd087f22803bb/ocaml/typing/typedtree.ml#L1018).  I've made this function do more, "enhancing" it to [`let_bound_idents_with_modes_sorts_and_checks`](https://github.com/ccasin/flambda-backend/blob/658b6c2f71d8a08bbae514d1adbb65d8f90e6ee2/ocaml/typing/typedtree.ml#L1011).  It turns out this function operates on fully typechecked let bindings and can look at the bound term, see if it's a function, and then look at the already-parsed zero_alloc info on that function.  This is pretty gross, just because of the way the code is structured, but it does seem to work.

ocaml/typing/typedtree.mli
ncik-roberts: On a separate note, I would find this clearer if `Texp_apply` carried `Assume_info.t option` instead. It seems like `Assume_info.t` can sneakily represent "None" &mdash; that seems kind of odd, and I'd like to get rid of `Assume_info.is_none` if that's possible.

ocaml/parsing/builtin_attributes.mli
```diff
@@ -263,5 +263,5 @@ val is_check_enabled : opt:bool -> property -> bool
 
 val get_property_attribute : Parsetree.attributes -> property -> check_attribute
 
-val get_assume_zero_alloc :
-  with_warnings:bool -> Parsetree.attributes -> Assume_info.t
+val assume_zero_alloc :
+  check_allowed:bool -> check_attribute -> Assume_info.t
```
ncik-roberts, line 267: Suggestion: `is_check_allowed`. I misinterpreted "check" as a verb at first.

ocaml/typing/typecore.ml
```diff
@@ -5308,9 +5342,17 @@ and type_expect_
       let (args, ty_res, ap_mode, pm) =
         type_application env loc expected_mode pm funct funct_mode sargs rt
       in
+      let assume_zero_alloc =
+        let open Builtin_attributes in
```
ncik-roberts, line 5346: Greta left a comment to this same effect elsewhere: let's avoid `open` here.

ocaml/typing/typedtree.mli
gretay-js: What is the advantage of your proposal? Does `Texp_apply` need to know what `Assume_info.t` is and whether it is `None` or not? `Assume_info.is_none` is used by `Debuginfo.is_none`. Explicitly tracking `option` here would make it harder to change the representation.

ocaml/testsuite/tests/warnings/w53.ml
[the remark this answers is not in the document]
gretay-js: > I'm not excited about minting 2 more files for misplaced attribute tests. (There's already 2, for ppxes vs no-ppxes.)
> 
> * Eventually, could `[@@@zero_alloc all]` be scoped to the submodule it appears in, like `[@@@ocaml.warning ...]` attributes?

yes, we had at go at it a while ago https://github.com/ocaml-flambda/flambda-backend/pull/1559 , it's still something the user want, and it maybe be easier to implement now that the parsing of these attributes on functions is earlier.

> * For the time being, maybe we could just put the `[@@@zero_alloc all]` test last.

It doesn't work like this, unfortunately, in the current `main`.  No matter where it is in the file, its effect is for the entire file (because its effect is applied later, after parsing).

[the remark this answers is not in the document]
ncik-roberts: Ah, I see. In that case I'll retire my objection about separate files.

ocaml/parsing/builtin_attributes.mli
gretay-js: Yeah, we had other properties at the beginning (like check that a functio nhas no caml_apply).  None of them was so I'm happy to rename everything to `Zero_alloc_checker` etc and eliminate unnecessary but it should be done in a separate PR across the entire code not just frontent. I'm happy to make such a  PR.

> "check" and "property" are two names for one thing, used interchangeably in several of these bindings. I find both names to be too vague.

The intented meaning of "property"  and "check" were different, but I agree that "check" is vague and would love to hear your suggestion of a better name!

ocaml/parsing/builtin_attributes.mli
gretay-js: how about `Zero_alloc_info` as we are considering extending it for "assert" not just "assume" soon?

ocaml/parsing/builtin_attributes.mli
ncik-roberts: Luckily the name `zero_alloc_checker` as you suggest sounds great to me :) Thanks for the context, I agree the original decision made sense but revisiting it now sounds good to me.

gretay-js: Here is a draft PR for the proposed refactoring https://github.com/ocaml-flambda/flambda-backend/pull/2416. It will conflict with this one, so we should probably wait.

ocaml/testsuite/tests/warnings/w53.ml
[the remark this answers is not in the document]
ccasin: I am a bit confused by the above conversation.  `zero_alloc all` changes the meaning of `zero_alloc` attributes, but I do not think it changes whether or not we expect w53 to fire for them.

I added a test that demonstrates that.  It is just a copy of the existing w53 tests for zero_alloc with `[@@@zero_alloc all]` added to the top, showing that the same things do/don't result in w53.

Along the way, I figured out how to share reference files for the normal and marshalled versions of these tests, so you'll see I also deleted `w53_marshalled.compilers.reference`.  That test now uses the same reference file as the normal `w53` test.  And the normal/marshalled versions of this new test share one new reference file.

ocaml/typing/typedtree.mli
ccasin: To Greta's suggestion: The reason some places of the typed tree have `check_attribute` and other places have `assume_info` is just that these types are what is needed by the existing support for `zero_alloc` after typechecking.  For example, lambda's `function_attribute` type wants a `check_attribute`, and the support for putting assumes on scopes wants a `Assume_info.t`.  We could definitely consider trying to clean up those interfaces to agree, but that starts to touch later parts of the compiler, and I propose putting that off for a future PR that isn't focused on the front-end.

In general I already feel bad that this PR does some amount of moving files around and some amount of refactoring - it makes the history nicer when moves and changes happen in separate commits.  I have left a CR in the `Assume_info` module (now part of `Zero_alloc_utils` as requested, see below) to revisit this.

To Nick's suggestion: For the same reason, I don't really want to make changes to the `Assume_info.t` type in this PR, which is meant to be about front-end refactoring.  Such changes would reach further into other parts of the compiler.  I have left a comment in `typed_tree.mli` explaining this field - maybe that, along with the CR mentioned above, is sufficient for now?

ocaml/testsuite/tests/warnings/w53.ml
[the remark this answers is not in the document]
ncik-roberts: I didn’t understand that `all` didn’t affect w53 for other zero-alloc attributes. (You could imagine it doing so.) In any case, your approach sounds good, and thanks also for getting rid of the extra reference file.

ocaml/typing/typedtree.mli
ncik-roberts: Happy to defer!

ccasin:
> I think we should merge assume_info.ml into zero_alloc_utils.ml if there are no cyclic deps, now that they are in the same folder. I considered doing it in https://github.com/ocaml-flambda/flambda-backend/pull/2165 but kept assume_info next to debuginfo. See comments about transl_bound_exp and the new fields of typedtree.ml. It will also be easier to add "zero_alloc assert" info on expressions later.

I have done the merge, which is a good idea.  But suggest putting off bigger refactors for later - see other comments.

ocaml/parsing/builtin_attributes.mli
ccasin: Done, thanks.

ocaml/parsing/builtin_attributes.mli
ccasin: It is now `Zero_alloc_utils.Assume_info`.

gretay-js (approved):

ocaml/testsuite/tests/warnings/w53.ml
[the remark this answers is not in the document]
gretay-js: The new tests look fine to me.

ocaml/typing/typedtree.mli
gretay-js: sounds good, thanks!

ocaml/lambda/translcore.ml
ccasin: Well spotted.  To summarize a discussion had in person, I think this is fine, though we can make an additional simplification.

This isn't a bug, because: In the old world, lambda needs to look for attributes in two places to see whether to extend the scope - here we looked for attributes on the let binding and in `transl_function` we looked for attributes on the function.   The effect of the old code here was to add assume info on let bindings into the scope.  The front-end now moves that info into the `Texp_function` (or gives an error if the bound thing is not a function). So it's fine to only look for it in the `Texp_function` case.

But actually, we don't even need to look for it in the `Texp_function` case, because in the new world lambda doesn't need to worry about `zero_alloc` attributes on lets - they are moved to the let-bound function in type checking.  So in  the common case where you have a let-bound function, I was needlessly calculating the `assume_info` and updating the scope twice.  I am pushing a commit that just uses `Assume_info.none` here, because it will be updated in transl_function.

ccasin:
@ncik-roberts This is now fully ready for another look.  Sorry for updating it so incrementally - busy couple weeks.

@Ekdohibs I think this will require your approval for the changes to `chamelon` - could you take a quick look?

Ekdohibs (approved):
For changes to `chamelon/`

ncik-roberts (approved):
Looks good to me. (There's one outstanding conversation but it looks like you're working that out with Greta.)

PR #2402, opened 2024-03-28
Take into account witnesses when computing [lessequal] on values

gretay-js:
otherwise we lose witnesses sometimes. 

Take 3 of #2353.  
(I messed up something when merging on the chain of PRs and this PR closed on me, can't reopen it as it's base branch is gone.)

### DIFF
--- a/backend/checkmach.ml
+++ b/backend/checkmach.ml
@@ -88,6 +88,8 @@ module Witnesses : sig
 
   val meet : t -> t -> t
 
+  val lessequal : t -> t -> bool
+
   val create : Witness.kind -> Debuginfo.t -> t
 
   val print : Format.formatter -> t -> unit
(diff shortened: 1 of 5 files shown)


gretay-js:
@xclerc to address your question from  #2353 :
> Do we have an idea of the size of the
witness sets in practice? With the
comparision over the domain now
possibly involving an inclusion test,
the cost could increase significantly.

I tested it on a large code base.  The overall build time was not significantly different (within noise). The number of fixed point iterations on most files did not change.

Even with `-function-layout source`, there was only a score of files (out of thousands) with more than 3 fixed point iterations, the maximum number of fixed point iterations was 6 (observed on 3 files). The inner fixed point in dataflow analysis had at most 2 iterations. The special case of "self-recursive" functions had at most 3 fixed point iterations.

PR #2404, opened 2024-04-01
Disable stack checks by default

xclerc:
This pull requests disable stack checks
by default. They can be enabled by
passing `--enable-stack_checks` to
the configuration script.

Without stack checks, there are no
reallocations when the stack would
need to grow (as stack overflow
exception is raised instead).

As a consequence, two runtime
parameters, namely `i` and `j` are
introduced to control respectively
the size of the stack for a main
thread and an auxiliary thread.


mshinwell:
Looks like the wrong version of autoconf was used to regenerate `ocaml/configure`.

mshinwell:
I don't understand the changes in the testsuite: a couple of tests have been marked as runtime5-only, but it seems like these should work on runtime4 still?

ocaml/runtime/startup_aux.c
```diff
@@ -48,11 +50,25 @@ static void init_startup_params(void)
...
+    }
+  }
+
   params.init_percent_free = Percent_free_def;
   params.init_minor_heap_wsz = Minor_heap_def;
   params.init_custom_major_ratio = Custom_major_ratio_def;
   params.init_custom_minor_ratio = Custom_minor_ratio_def;
   params.init_custom_minor_max_bsz = Custom_minor_max_bsz_def;
+  params.init_main_stack_wsz = init_main_stack_wsz;
+  params.init_thread_stack_wsz = init_main_stack_wsz / 4;
```
mshinwell, line 71: Why was 4 chosen?  We should add a comment.

ocaml/runtime/startup_aux.c
```diff
@@ -48,11 +50,25 @@ static void init_startup_params(void)
   char_os * cds_file;
 #endif
 
+  uintnat init_main_stack_wsz;
+  struct rlimit rlimit;
+  if (getrlimit(RLIMIT_STACK, &rlimit)) {
+    init_main_stack_wsz = Wsize_bsize(Stack_init_bsize);
```
mshinwell, line 56: I think `Stack_init_bsize` might be too small, we should discuss and choose a larger number I think.

ocaml/runtime/signals_nat.c
```diff
@@ -85,3 +85,39 @@ void caml_garbage_collection(void)
...
+    act.sa_handler = SIG_DFL;
+    act.sa_flags = 0;
+    sigemptyset(&act.sa_mask);
+    sigaction(SIGSEGV, &act, NULL);
+  }
+}
+
+void caml_init_nat_signals(void)
+{
+    struct sigaction act;
```
mshinwell, line 118: (should be 2 space indentation)

ocaml/runtime/startup_aux.c
mshinwell: We should also check this new number against `Max_stack_def`.

ocaml/runtime/fiber.c
```diff
@@ -49,13 +49,16 @@
 
 static _Atomic int64_t fiber_id = 0;
 
-uintnat caml_get_init_stack_wsize (void)
+uintnat caml_get_init_stack_wsize (bool is_main_thread)
 {
-  uintnat default_stack_wsize = Wsize_bsize(Stack_init_bsize);
+  uintnat init_stack_wsize =
```
mshinwell, line 54: As discussed, we should use the normal OCaml 5 logic when stack checks are enabled (same everywhere in this PR).

ocaml/runtime/fiber.c
```diff
@@ -803,6 +833,17 @@ void caml_free_stack (struct stack_info* stack)
 
   CAMLassert(stack->magic == 42);
   CAMLassert(cache != NULL);
+
+#ifndef USE_MMAP_MAP_STACK
+#ifdef NATIVE_CODE
+  // CR xclerc for xclerc: check the value returned by mprotect?
```
mshinwell, line 839: I would probably skip that.

ocaml/runtime/fiber.c
```diff
@@ -803,6 +833,17 @@ void caml_free_stack (struct stack_info* stack)
...
+  int page_size = getpagesize();
+  mprotect((void *) ((char *) stack + page_size),
```
mshinwell, line 841: Maybe we could add some macros for taking the `struct stack_info` and returning the lower and upper bounds of the guard page?  There are other occurrences where this would clean things up.

ocaml/runtime/fiber.c
```diff
@@ -111,7 +114,34 @@ Caml_inline struct stack_info* alloc_for_stack (mlsize_t wosize)
   si->size = len;
   return si;
 #else
+#ifdef NATIVE_CODE
+  size_t bsize = sizeof(value) * wosize;
```
mshinwell, line 118: I think the canonical way of doing this in the runtime is `Bsize_wsize(wosize)`.

ocaml/runtime/fiber.c
```diff
@@ -111,7 +114,34 @@ Caml_inline struct stack_info* alloc_for_stack (mlsize_t wosize)
...
+  }
+  block->size = len;
+  int protect = mprotect(((char*) block) + page_size, page_size, PROT_NONE);
+  if (protect) {
+    free(block);
+    return NULL;
+  }
+  return block;
+#else
+  size_t len = sizeof(struct stack_info) +
```
mshinwell, line 139: We should make sure we're happy with the `Stack_threshold_ptr` computation, or ensure it isn't used if the value is wrong.

ocaml/runtime/fiber.c
```diff
@@ -111,7 +114,34 @@ Caml_inline struct stack_info* alloc_for_stack (mlsize_t wosize)
   si->size = len;
   return si;
 #else
+#ifdef NATIVE_CODE
```
mshinwell, line 117: I'm unclear why this `NATIVE_CODE` conditional is here.

mshinwell (changes requested):

mshinwell:
As discussed we also need more tests (including for threads).

ocaml/runtime/signals_nat.c
```diff
@@ -85,3 +85,39 @@ void caml_garbage_collection(void)
...
+DECLARE_SIGNAL_HANDLER(segv_handler)
+{
+  struct sigaction act;
+  struct stack_info *block = Caml_state->current_stack;
+  char* fault_addr = info->si_addr;
+  int page_size = getpagesize();
+  char* protected_low = (char *) block + page_size;
+  char* protected_high = protected_low + page_size;
+  if ((fault_addr >= protected_low) && (fault_addr < protected_high)) {
+    caml_raise_stack_overflow();
```
stedolan, line 107: This looks suspicious to me - you're calling the C code to raise an exception, but you're probably on the OCaml stack. Have you tried catching this exception with Sys.with_async_exns? Does it work?

Also, I don't think you're guaranteed to hit the fault at function entry, since it doesn't look like functions touch their whole stackframe early. That may mean that something goes wrong when trying to compute a backtrace here.

Returning from a signal handler like this is dicey and platform-dependent. See the ifdef'd code in OCaml 4 for details.

ocaml/runtime/fiber.c
```diff
@@ -803,6 +833,16 @@ void caml_free_stack (struct stack_info* stack)
...
+
+#ifndef USE_MMAP_MAP_STACK
+#ifdef NATIVE_CODE
+  int page_size = getpagesize();
+  mprotect((void *) ((char *) stack + page_size),
+           page_size,
+           PROT_READ | PROT_WRITE | PROT_EXEC);
+#endif
+#endif
+
```
stedolan, line 845: Trying to set a page `PROT_WRITE | PROT_EXEC` will fail on many OSes.

ocaml/runtime/fiber.c
```diff
@@ -111,7 +114,34 @@ Caml_inline struct stack_info* alloc_for_stack (mlsize_t wosize)
...
+               bsize +
+               8 /* for alignment to 16-bytes, needed for arm64 */ +
+               sizeof(struct stack_handler);
+  struct stack_info* block;
+  int memalign_res = posix_memalign((void **) &block, page_size, len);
+  if (memalign_res) {
+    return NULL;
+  }
+  block->size = len;
+  int protect = mprotect(((char*) block) + page_size, page_size, PROT_NONE);
```
stedolan, line 132: It's probably fine, but there's definitely something a bit weird about mprotecting memory that you got from the allocator. Have you considered using mmap directly instead? For instance, by extending the USE_MMAP_MAP_STACK case above with a guard page?

stedolan: Mostly, it struck me as impolite!

Concretely, it'll kill transparent hugepages if the allocator gives you a region in the middle of a large allocated area that had THP optimisation applied by the kernel.

ocaml/runtime/fiber.c
xclerc: (Added `#ifdef`s where necessary.)

ocaml/runtime/fiber.c
xclerc: Discussed off-line, that macro is used only
in bytecode, where the changes in this pull
request do not apply.

ocaml/runtime/startup_aux.c
xclerc: Added the logic to cap the value to `Max_stack_def`.

ocaml/runtime/signals_nat.c
xclerc: Changed to the OCaml 4 code.

mshinwell (approved):

mshinwell:
Two things to check based on discussion with @lpw25:
- Make sure a test that causes a stack overflow does so after approximately the same number of iterations under both runtime4, and runtime5 + this patch
- Look at `/proc/<pid>/smaps` to try to see if huge page usage is similar on runtime4, and runtime5 + this patch.

xclerc:
I have manually check both items; in particular
huge parges are being used if we comment
out the call to `mprotect`.

mshinwell:
(you meant `madvise`, but yes!)

xclerc:
Indeed, sorry, I meant `madvise`,
and I should have added that
huge pages are then used if we
request a stack of 1 Gb.

PR #2405, opened 2024-04-01
Improve error message around `any` in functions

goldfirere:
As title.

Inspired by feedback from a colleague.


ccasin (approved):
Thanks!

PR #2406, opened 2024-04-01
Fix and test vector load/store for arrays of unboxed ints

TheNumbat:
The array load/store functions for `int32# array`/`int64# array`/`nativeint# array` need to skip the custom_ops pointer - see tests.

### DIFF
--- a/middle_end/flambda2/to_cmm/to_cmm_primitive.ml
+++ b/middle_end/flambda2/to_cmm/to_cmm_primitive.ml
@@ -142,17 +142,24 @@ let array_length ~dbg arr (kind : P.Array_kind.t) =
        though the contents are of word width. *)
     C.unboxed_int64_or_nativeint_array_length arr dbg
 
-let array_load_128 ~dbg arr index element_width_log2 =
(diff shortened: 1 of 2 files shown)


middle_end/flambda2/to_cmm/to_cmm_primitive.ml
```diff
@@ -142,19 +142,25 @@ let array_length ~dbg arr (kind : P.Array_kind.t) =
...
-       (Cconst_int (element_width_log2, dbg))
-       dbg)
-    new_value dbg
+let array_load_128 ~dbg ~element_width_log2 ~has_custom_ops arr index =
+  let index =
+    C.lsl_int (C.untag_int index dbg) (Cconst_int (element_width_log2, dbg)) dbg
+  in
+  let index =
+    (* Skip custom_ops pointer *)
+    if has_custom_ops then C.add_int index (Cconst_int (8, dbg)) dbg else index
```
mshinwell, line 151: This should probably use `Arch.size_addr` instead of `8`.

mshinwell (changes requested):

mshinwell (approved):

PR #2408, opened 2024-04-02
Move int<->float casts into scalar cast instruction

TheNumbat:
Preparation for #2391 - moves `floatofint` and `intoffloat` instructions into `scalar_cast`. This will reduce code duplication when adding float32 casts.


xclerc (approved):
Not sure about the `of` _vs_ `to`
choice in the constructor name,
but unfortunately I don't think
we have an universally-enforced
convention.

PR #2417, opened 2024-04-08
Mode cross unboxed float

riaqn:
This PR makes jkind to mode cross unboxed float.

During the process I found another issue: it seems that the kind information is not used when check for mode crossing:

```
module Immediate : sig
  val id : ('a : immediate). 'a -> 'a
end = ...

let imm_escape () = Immediate.id (local_ 42) [@nontail]
[%%expect{|
val imm_escape : unit -> int = <fun>
|}, Principal{|
Line 1, characters 33-44:
1 | let imm_escape () = Immediate.id (local_ 42) [@nontail]
                                     ^^^^^^^^^^^
Error: This value escapes its region
|}]
```
The argument to `Immediate.id` is principally known to be `immediate` and thus mode cross. Note that this is on immediate instead of unboxed float, and thus an issue orthogonal to this PR. Therefore, I only left a few comments without fixing them.


ocaml/testsuite/tests/typing-modal-kinds/basics.ml
```diff
@@ -493,29 +494,50 @@ Line 1, characters 51-52:
 
 |}]
 
-let float_u_unshare = let x : float# = #3.14 in Float_u.ignore x; Float_u.unique x
+(* CR layouts: The following should pass, even in principal mode, because the
```
goldfirere, line 497:
```suggestion
(* CR layouts v2.8: The following should pass, even in principal mode, because the
```

ocaml/testsuite/tests/typing-modal-kinds/expected_mode.ml
```diff
@@ -115,10 +115,13 @@ Line 1, characters 56-57:
 Error: This value escapes its region
 |}]
 
+(* CR layouts: The following should pass, even in principal mode. *)
```
goldfirere, line 118:
```suggestion
(* CR layouts v2.8: The following should pass, even in principal mode. *)
```

goldfirere (approved):
Odd about the failures in principality. But I disagree with the characterization in the ticket description. The kinds are definitely being used, because things work for `Hidden_float`. My suspicion is that there is something off about the principality calculation. Thanks for flagging these, regardless.

PR #2418, opened 2024-04-08
Catch and handle Bottom equations in the typing environment

lthls:
The exception `Bottom_equation` is raised when an inconsistency is discovered in a typing environment and there is no `Or_bottom` wrapper to handle it properly.
This can happen for legitimate reasons (the typing env is not always in a fully-reduced state, and in theory even adding unrelated equations could cause us to perform a bit of reduction and discover Bottom). However, currently (with the advanced meet) we let it crash the compiler, in order to get a better sense of what cases can trigger this. This PR is one attempt to properly catch the exception and handle it.

My first attempts changed the type of all functions that could throw the exception to return `Or_bottom.t` instead. It proved very painful to do: a lot of places in `Simplify` rely on such functions, and not all of them can do something obviously correct in the case where `Bottom` is returned.
So this proposal hides everything in the typing environment: an additional `has_bottom` flag tracks whether we have introduced a bottom equation, and every operation that can return a Bottom value (`meet`, `find`, ...) will immediately return `Bottom` if the environment has this flag set.

The next step is to start exposing this flag in the `Flambda2_types` interface in some way, to allow `Simplify` to make clear when it is expecting its environment to be non-bottom. I'll work on that on the same branch for now and update this PR when it's done, unless we decide to merge this first (in which case I'll make a follow-up PR).

### DIFF
--- a/middle_end/flambda2/types/env/typing_env.ml
+++ b/middle_end/flambda2/types/env/typing_env.ml
@@ -106,7 +106,9 @@ type t =
        list *)
     current_level : One_level.t;
     next_binding_time : Binding_time.t;
-    min_binding_time : Binding_time.t (* Earlier variables have mode In_types *)
+    min_binding_time : Binding_time.t;
+        (* Earlier variables have mode In_types *)
+    has_bottom : bool
   }
 
 and serializable =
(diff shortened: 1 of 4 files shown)


chambart (commented):
Ok for now. The change to the `typing_env` type (with or without bottom) can be in this PR or a next one.

Ekdohibs (approved):

PR #2419, opened 2024-04-08
Fix incorrect use of AC_ARG_ENABLE

stedolan:
The two last arguments to `AC_ARG_ENABLE` are not 'if enabled' and 'if disabled', they are 'if specified' and 'if unspecified'. Getting this wrong in `configure.ac` means that currently `--enable-runtime5` and `--disable-runtime5` do the same thing, enabling runtime5.

This patch fixes these uses, so that `--enable-foo` and `--disable-foo` become opposites.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -48,22 +48,22 @@ AC_ARG_ENABLE([middle-end],
 AC_ARG_ENABLE([runtime5],
   [AS_HELP_STRING([--enable-runtime5],
     [Use the OCaml 5 runtime])],
-  [runtime5_arg=--enable-runtime5],
-  [runtime5_arg=])
+  [],
+  [enable_runtime5=no])
 
 AC_ARG_ENABLE([coverage],
   [AS_HELP_STRING([--enable-coverage],
     [Run compiler tests instrumented to output coverage data using bisect_ppx
      (WARNING: Cannot build an installable compiler with this flag enabled.
      Mainly intended for use in CI.)])],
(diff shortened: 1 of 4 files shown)


configure.ac
```diff
@@ -48,45 +48,52 @@ AC_ARG_ENABLE([middle-end],
...
+  [enable_dev=no])
 
 AC_SUBST([prefix])
 AC_SUBST([middle_end])
 AC_SUBST([dune])
+
+coverage="$enable_coverage"
 AC_SUBST([coverage])
+
+legacy_layout="$legacy_layout"
```
mshinwell, line 84: Same here, in the quotes?

configure.ac
```diff
@@ -48,45 +48,52 @@ AC_ARG_ENABLE([middle-end],
...
+  [],
+  [enable_coverage=no])
 
 AC_ARG_ENABLE([legacy-library-layout],
   [AS_HELP_STRING([--enable-legacy-library-layout],
     [Install libraries unix, str, dynlink and bigarray in the toplevel
      ocaml library directory (Same as upstream OCaml < 5.0)])],
-  [legacy_layout=yes],
+  [],
   [legacy_layout=no])
```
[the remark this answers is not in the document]
stedolan: Thanks, fixed

configure.ac
stedolan: Thanks, fixed.

mshinwell (approved):

PR #2422, opened 2024-04-08
Implement mixed blocks in runtime 4

ncik-roberts:
Copy the changes from #2380 into runtime 4. I did this as follows:

- Look at the diff under `ocaml/runtime` in #2380 and adapt all such changes for runtime 4.
- Grep for certain signposts and read the surrounding code:
  * Signpost 1: `No_scan_tag`. This indicates that the runtime system may be scanning the fields in some way, and has already carved out an exception for `No_scan_tag` and therefore might also need an exception for mixed blocks.
  * Signpost 2: `Closure_tag`. Similar to the previous item, but additionally catches the cases where each of the tags above the `No_scan_tag` threshold are matched on individually. `Closure_tag`s are handled specially because they can have code pointers.
  * Signpost 3: `Wosize_*`. These often need to be changed to `Scannable_wosize`.

Most of the changes come from "looking at the diff under `ocaml/runtime`".

Testing: I enable the mixed block tests from #2380 for runtime 4 as well.


ncik-roberts:
(A test is failing &mdash; sorry, this wasn't happening prior to opening the PR. Looking into that...)

ocaml/runtime4/weak.c
```diff
@@ -398,27 +398,35 @@ CAMLprim value caml_ephe_get_data (value ar)
 
 static void copy_value(value src, value dst)
 {
-  mlsize_t sz, i;
-  sz = Wosize_val(src);
+  mlsize_t scan_from, scan_to;
   if (Tag_val (src) >= No_scan_tag) {
     /* Direct copy */
-    memcpy (Bp_val (dst), Bp_val (src), Bsize_wsize (sz));
+    // CR mixed blocks: bsize val?
```
stedolan, line 404: `Bosize_val` (but `Bsize_wsize (Wosize_val (..))` is fine too)

ocaml/runtime4/major_gc.c
```diff
@@ -670,7 +672,7 @@ Caml_noinline static intnat do_some_marking
         continue;
       }
       scan = Op_val(block);
-      obj_end = scan + Wosize_hd(hd);
+      obj_end = scan + Scannable_wosize_hd(hd);
```
stedolan, line 675: This needs to account for the work skipped or the GC will run at the wrong rate. `work` needs to decrease by 1 for each word of heap that is encountered, whether scanned or not. So I think something like `work -= wosize - scannable_sz` needs to happen here. (Possibly this is wrong in runtime5 too? Haven't checked)

stedolan (changes requested):

ocaml/runtime4/major_gc.c
ncik-roberts: I believe this handles it in runtime 5: https://github.com/ocaml-flambda/flambda-backend/blob/5c5fb708dd41f5ee9fe2cebec3b1212cb850cf48/ocaml/runtime/major_gc.c#L924. It looks like I dropped it when porting the diff.

I've now added it -- this is ready for another review.

stedolan: Looks good!

stedolan (approved):
Looks good. I think there are conflicts from the ocamltest syntax change, but looks ready to merge otherwise.

PR #2416, opened 2024-04-05
Zero alloc: refactor to remove "property" and improve naming

gretay-js:
No change in functionality, just rename and refactor. 

- Remove `Lambda.property`: we only have one property to check at the moment: "zero_alloc". Some but not all the code is trying to be more general, but the terminology is vague and confusing. 
- Rename "Checkmach" to `zero_alloc_checker` as a consequence of the above and in preparation to cfg-based analysis in a separate PR.

PR #2424, opened 2024-04-09
Zero alloc: fix bug interaction between "opt" and top-level "all"

gretay-js:
The following example fails the check:
```
[@@@zero_alloc all]
let[@zero_alloc opt] foo x = (x,x)
```
but it is expected to pass the check by default (and fail with `-check-zero-alloc all` compiler flag). 

This PR fixes the bug and adds a test to show the expected behavior. 

The bug is that `Clflags.zero_alloc_check_assert_all` is checked too late.  The cause of the bug is that "opt" annotations are removed by the middle-end (`from_lambda` transformation) if `Lambda.is_check_enabled ~opt` is false, but then in the backend, if `Clflags.zero_alloc_check_assert_all` is true, we add "assert zero_alloc"  annotations on the same function, because it doesn't have "ignore" or "assume". The fix is to check the two flags in one place in the middle-end during `from_lambda` conversion.  The reason that we don't check it all in the backend is the interaction with the dead code elimination in flambda: it's better to remove unnecessary "asserts" earlier. We can probably move this check even earlier after #2400 is merged, but that would be a separate PR for refactoring only.

As a result of the fix, we can eliminate the now-unused `Ignore_all` constructor from the middle-end and `Cmm`.

~On top of #2416.~  Best review commit-by-commit.

### DIFF
--- a/backend/cfg/cfg.ml
+++ b/backend/cfg/cfg.ml
@@ -57,9 +57,7 @@ let rec of_cmm_codegen_option : Cmm.codegen_option list -> codegen_option list =
     match hd with
     | No_CSE -> No_CSE :: of_cmm_codegen_option tl
     | Reduce_code_size -> Reduce_code_size :: of_cmm_codegen_option tl
-    | Use_linscan_regalloc | Assume _ | Check _
-      ->
-      of_cmm_codegen_option tl)
+    | Use_linscan_regalloc | Assume _ | Check _ -> of_cmm_codegen_option tl)
 
 type t =
   { blocks : basic_block Label.Tbl.t;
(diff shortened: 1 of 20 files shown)


xclerc (commented):
I have arguably not really paid
attention to the semantics of
`[@@@zero_alloc all]`, but I find
it a bit confusing. It looks like
it applies to the whole file, while
I would expect it to apply "from
that point". Is that correct?

gretay-js:
> It looks like 
> it applies to the whole file, while
> I would expect it to apply "from
> that point". Is that correct?

yes, `[@@@zero_alloc all]` currently applies to the whole file.  it's not ideal, but it was easier (because the flag is only used in the backend). one way to improve it would be to keep track of the scope of this attribute similarly to the way we keep track of the warnings scopes.

PR #2425, opened 2024-04-09
Protect calling not_nolabel_function, just as it used to be

goldfirere:
This fixes a regression detected in the rollout.

There remains a bug in the handling of optional parameters, but this bug is not new. I have also submitted a bugfix upstream at https://github.com/ocaml/ocaml/pull/13088, but there's always the chance this bugfix will cause problems for us here. Let's just wait until we pick up the upstream bugfix in the fullness of time.


ccasin (approved):
LGTM

PR #2429, opened 2024-04-10
Hash of mixed blocks raises an exception

ncik-roberts:
Raise an exception when hashing a mixed block.

After extensive discussion we decided this is the best way of disarming footguns.
  * Raising reliably is, well, reliable. You don't sometimes get a good hash and sometimes get a bad hash.
  * It's not clear how, or whether, to properly hash the unscannable suffix.
  * Raising makes it less likely that users will hit bugs due to different hashing behaviour in bytecode vs. native code &mdash; raising in native code will reliably flag that something is up.

The main downside of this PR is that it requires the removal of the `[@@noalloc]` attribute on the `caml_hash` external used by `Hashtbl.hash`. There are two concerns you might have:
  * Safety. Will we remember to remove `[@@noalloc]` on all other externals bound to `caml_hash`?
  * Performance. `[@@noalloc]` removes some overhead by changing the calling convention based on the assumption that the external won't allocate or raise. Does the overhead matter?

To mitigate the safety concern, I've renamed `caml_hash` to `caml_hash_exn`. That will force us to visit and think about every external that's currently bound to `caml_hash`.

My prior would be that the performance concern is not substantial. Most data structures that use polymorphic hash also use polymorphic compare, which is already not `[@@noalloc]`. I'll going to do some internal benchmarking of the compiler itself, which uses polymorphic hash extensively, to validate this hypothesis.


ocaml/stdlib/char.ml
```diff
@@ -64,7 +64,10 @@ type t = char
 let compare c1 c2 = code c1 - code c2
 let equal (c1: t) (c2: t) = compare c1 c2 = 0
 
+(* [caml_hash_exn] doesn't raise on booleans, so it's safe for
```
xclerc, line 67:
```suggestion
(* [caml_hash_exn] doesn't raise on chars, so it's safe for
```

ocaml/testsuite/tests/mixed-blocks/hash.ml
```diff
@@ -18,13 +18,20 @@ all_modules = "hash.ml"
...
-   revisiting -- the simplest way to accomplish this may well be to
-   support mixed blocks in bytecode.
- *)
+   normal blocks in bytecode. Hash raises on mixed blocks (but not
+   on normal blocks).
+
+   We could consider making native hash return a value instead. But,
+   if it's a different value than bytecode, it's important that users
+   can't marshal mixed block values from native code to bytecode.
+   (Otherwise you could marshal a hashtable, breaking its invariants.)
```
xclerc, line 27: (Not really about this PR, just to be
sure I understand.)
One can still marshal from bytecode
to native, and the unmarshaled value
will contain regular blocks rather than
mixed ones, which will pressumably
lead to unexpected results. Is that
correct?

xclerc (approved):
Renaming the primitive may make it
more difficult to use this compiler as
a drop-in replacement for upstream's,
but I think the benefits outlined above
outweigh this concern.

ocaml/testsuite/tests/mixed-blocks/hash.ml
xclerc: Incidentally, that probably means we
should never use mixed blocks in the
compiler (or at least `file_format/`
and dependencies).

ncik-roberts:
There is a good chance I'll revert the renaming aspect of this PR after we've had the chance to check that nobody uses `caml_hash`. (I don't think there's a strong reason to name it `caml_hash_exn` in perpetuity &mdash; just when we change it to raise.)

Your point about marshaling from bytecode to native code is good, I hadn't thought of that. I'll open a follow-up PR that sets canary bits in bytecode for the runtime repr of mixed blocks.

mshinwell:
I'm unsure why we would revert the renaming.  It seems perfectly possible for future packages to appear with direct bindings to `caml_hash`, which might then be broken.

PR #2430, opened 2024-04-11
Shorter sequence for stack checks

xclerc:
This pull request tweaks the instruction sequence
emitted to make it shorter, and also avoid the
consumption of a register (the later does not
matter here, but may simplify another pull request).

The original sequence was:

```
          I.lea (mem64 NONE (-(max_frame_size + threshold_offset)) RSP) r10;
          I.cmp (domain_field Domainstate.Domain_current_stack) r10;
          I.jb (label overflow);
```

and is now:

```
      I.cmp (domain_field Domainstate.Domain_current_stack_bound) rsp;
      I.jb (label overflow);
```

which is possible because the new sequence is
emitted after `rsp` is updated, and because a
newly-introduced domain field (namely
`current_stack_bound`) contains the value that
was computed by the old sequence.

(original suggestion by @stedolan, errors are mine)

### DIFF
--- a/backend/amd64/emit.mlp
+++ b/backend/amd64/emit.mlp
@@ -1196,6 +1196,19 @@ let emit_simd_instr op i =
   | SSE42 (Cmpistrz n) ->
     I.pcmpistri (X86_dsl.int n) (arg i 1) (arg i 0); I.set E (res8 i 0); I.movzx (res8 i 0) (res i 0)
 
+(* CR xclerc for xclerc: consider for a proper type (possibly when rebasing
+   over the branch moving checks down) *)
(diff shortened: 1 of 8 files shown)


mshinwell:
I'll review this.

stedolan:
I think this needs to do something slightly more complicated than currently, because it looks like it is assuming that the max_frame_size is equal to the amount subtracted from `rsp` in the function prologue. This is nearly but not quite true - if the function contains e.g. a call with stack arguments, I think these are pushed to the stack later on in the function and not included in the original frame.

I don't think such cases come up much, so it would probably be fine to detect when frame_size differs from max_frame_size and use the old logic in such cases.

PR #2431, opened 2024-04-11
Fix mode locks for classes and modules

riaqn:
This PR is based on #2398, because we need the syncness/contention axes for testing (the uniqueness/linearity legacy setting cannot test this). It is however orthogonal to #2398 and can be reviewed standalone. After approval, this PR should **not** be merged until #2398 is merged.

This PR fixes the following things:
- Accessing modules/classes inside a closure_lock/shared_lock/escaping_lock should trigger locks too.
- `new cla` should always be `legacy`, where `cla` is a class.
- `obj # meth` should always be `legacy`, where `obj` is an object, and `meth` is a method.

In addition, it does the following clean up:
- `unboxed_lock` is enforced during `walk_locks`, instead of generating a boolean `must_lock`.
- `Env.add_value` takes mandatory `mode` instead of optional - it's dangerous to just assume value is `legacy` when unspecified.

### DIFF
--- a/ocaml/typing/env.ml
+++ b/ocaml/typing/env.ml
@@ -325,13 +325,18 @@ type shared_context =
   | Probe
   | Lazy
 
-type value_lock =
+type lock =
   | Escape_lock of escaping_context
   | Share_lock of shared_context
   | Closure_lock of closure_context option * Mode.Value.Comonadic.r
   | Region_lock
   | Exclave_lock
   | Unboxed_lock (* to prevent capture of terms with non-value types *)
+
+type lock_item =
+  | Value
+  | Module
+  | Class
 
 module IdTbl =
   struct
@@ -374,7 +379,7 @@ module IdTbl =
         }
 
       | Lock of {
-          mode: 'lock;
+          lock: 'lock;
           next: ('lock, 'a, 'b) t;
         }
 
@@ -407,8 +412,8 @@ module IdTbl =
       | _ ->
           assert false
 
-    let add_lock mode next =
-      { current = Ident.empty; layer = Lock {mode; next} }
+    let add_lock lock next =
+      { current = Ident.empty; layer = Lock {lock; next} }
 
     let map f next =
       {
@@ -416,21 +421,24 @@ module IdTbl =
         layer = Map {f; next}
       }
 
-    let rec find_same id tbl =
+    let rec find_same_without_locks id tbl =
       try Ident.find_same id tbl.current
       with Not_found as exn ->
         begin match tbl.layer with
-        | Open {next; _} -> find_same id next
-        | Map {f; next} -> f (find_same id next)
-        | Lock {mode=_; next} -> find_same id next
+        | Open {next; _} -> find_same_without_locks id next
+        | Map {f; next} -> f (find_same_without_locks id next)
+        | Lock {lock=_; next} -> find_same_without_locks id next
         | Nothing -> raise exn
         end
 
-    let rec find_name_and_locks wrap ~mark name tbl macc =
+    let find_same id (tbl : (empty, _, _) t) =
+      find_same_without_locks id tbl
+
+    let rec find_name_and_locks wrap ~mark name tbl macc : _ Result.t =
       try
         let (id, desc) = Ident.find_name name tbl.current in
-        Pident id, macc, desc
-      with Not_found as exn ->
+        Ok (Pident id, macc, desc)
+      with Not_found ->
         begin match tbl.layer with
         | Open {using; root; next; components} ->
             begin try
@@ -438,33 +446,38 @@ module IdTbl =
               let res = Pdot (root, name), macc, descr in
               if mark then begin match using with
               | None -> ()
-              | Some f -> begin
-                  match find_name_and_locks wrap ~mark:false name next macc with
-                  | exception Not_found -> f name None
-                  | _, _, descr' -> f name (Some (descr', descr))
+              | Some f -> begin match
+                  (find_name_and_locks wrap ~mark:false name next macc : _ Result.t)
+                  with
+                  | Error _ -> f name None
+                  | Ok (_, _, descr') -> f name (Some (descr', descr))
                 end
               end;
-              res
+              Ok res
             with Not_found ->
               find_name_and_locks wrap ~mark name next macc
             end
         | Map {f; next} ->
-            let (p, macc, desc) =
-              find_name_and_locks wrap ~mark name next macc in
-            p, macc, f desc
-        | Lock {mode; next} ->
-            find_name_and_locks wrap ~mark name next (mode :: macc)
+            find_name_and_locks wrap ~mark name next macc
+            |> Result.map (fun (p, macc, desc) -> p, macc, f desc)
+        | Lock {lock; next} ->
+            find_name_and_locks wrap ~mark name next (lock :: macc)
         | Nothing ->
-            raise exn
+            Error macc
         end
 
-    let find_name_and_modes wrap ~mark name tbl =
+    let find_name_and_locks wrap ~mark name tbl =
       find_name_and_locks wrap ~mark name tbl []
 
+    (** Find item by name whose accesses are not affected by locks, and thus
+        shouldn't encounter any locks. *)
     let find_name wrap ~mark name tbl =
-      let (id, ([] : empty list), desc) =
-        find_name_and_modes wrap ~mark name tbl in
-      id, desc
+      match (find_name_and_locks wrap ~mark name tbl
+        : (_ * empty list * _, empty list) Result.t) with
+      | Ok (id, [], desc) -> id, desc
+      | Ok (_, _ :: _, _) -> .
+      | Error [] -> raise Not_found
+      | Error (_ :: _) -> .
 
     let rec find_all wrap name tbl =
       List.map
@@ -482,7 +495,7 @@ module IdTbl =
       | Map {f; next} ->
           List.map (fun (p, desc) -> (p, f desc))
             (find_all wrap name next)
-      | Lock {mode=_;next} ->
+      | Lock {lock=_;next} ->
           find_all wrap name next
 
     let rec find_all_idents name tbl () =
@@ -499,7 +512,7 @@ module IdTbl =
             else
               find_all_idents name next ()
         | Map {next; _ } -> find_all_idents name next ()
-        | Lock {mode=_;next} ->
+        | Lock {lock=_;next} ->
             find_all_idents name next ()
       in
       Seq.append current next ()
@@ -524,7 +537,7 @@ module IdTbl =
           |> fold_name wrap
                (fun name (path, desc) -> f name (path, g desc))
                next
-      | Lock {mode=_; next} ->
+      | Lock {lock=_; next} ->
           fold_name wrap f next acc
 
     let rec local_keys tbl acc =
@@ -547,7 +560,7 @@ module IdTbl =
           iter wrap f next
       | Map {f=g; next} ->
           iter wrap (fun id (path, desc) -> f id (path, g desc)) next
-      | Lock {mode=_; next} ->
+      | Lock {lock=_; next} ->
           iter wrap f next
       | Nothing -> ()
 
@@ -555,7 +568,7 @@ module IdTbl =
       let keys2 = local_keys tbl2 [] in
       List.filter
         (fun id ->
-           try ignore (find_same id tbl1); false
+           try ignore (find_same_without_locks id tbl1); false
            with Not_found -> true)
         keys2
 
@@ -570,13 +583,13 @@ type type_descriptions = type_descr_kind
 let in_signature_flag = 0x01
 
 type t = {
-  values: (value_lock, value_entry, value_data) IdTbl.t;
+  values: (lock, value_entry, value_data) IdTbl.t;
   constrs: constructor_data TycompTbl.t;
   labels: label_data TycompTbl.t;
   types: (empty, type_data, type_data) IdTbl.t;
-  modules: (empty, module_entry, module_data) IdTbl.t;
+  modules: (lock, module_entry, module_data) IdTbl.t;
   modtypes: (empty, modtype_data, modtype_data) IdTbl.t;
-  classes: (empty, class_data, class_data) IdTbl.t;
+  classes: (lock, class_data, class_data) IdTbl.t;
   cltypes: (empty, cltype_data, cltype_data) IdTbl.t;
   functor_args: unit Ident.tbl;
   summary: summary;
@@ -683,6 +696,10 @@ and cltype_data =
   { cltda_declaration : class_type_declaration;
     cltda_shape : Shape.t }
 
+let mda_mode = Mode.Value.legacy |> Mode.Value.disallow_right
+
+let clda_mode = Mode.Value.legacy |> Mode.Value.disallow_right
+
 let empty_structure =
   Structure_comps {
     comp_values = NameMap.empty;
@@ -718,10 +735,11 @@ type lookup_error =
   | Generative_used_as_applicative of Longident.t
   | Illegal_reference_to_recursive_module
   | Cannot_scrape_alias of Longident.t * Path.t
-  | Local_value_escaping of Longident.t * escaping_context
-  | Once_value_used_in of Longident.t * shared_context
-  | Value_used_in_closure of Longident.t * Mode.Value.Comonadic.error * closure_context option
-  | Local_value_used_in_exclave of Longident.t
+  | Local_value_escaping of lock_item * Longident.t * escaping_context
+  | Once_value_used_in of lock_item * Longident.t * shared_context
+  | Value_used_in_closure of lock_item * Longident.t *
+      Mode.Value.Comonadic.error * closure_context option
+  | Local_value_used_in_exclave of lock_item * Longident.t
   | Non_value_used_in_object of Longident.t * type_expr * Jkind.Violation.t
 
 type error =
@@ -735,6 +753,16 @@ let error err = raise (Error err)
 
 let lookup_error loc env err =
   error (Lookup_error(loc, env, err))
+
+type actual_mode = {
+  mode : Mode.Value.l;
+  context : shared_context option
+}
+
+let mode_default mode = {
+  mode;
+  context = None
+}
 
 let same_constr = ref (fun _ _ _ -> assert false)
 
@@ -831,14 +859,14 @@ let check_functor_application =
   (* to be filled by Includemod *)
   ref ((fun ~errors:_ ~loc:_
          ~lid_whole_app:_  ~f0_path:_ ~args:_
-         ~arg_path:_ ~arg_mty:_ ~param_mty:_
+         ~arg_path:_ ~arg_mty:_ ~arg_mode:_ ~param_mty:_
          _env
          -> assert false) :
          errors:bool -> loc:Location.t ->
        lid_whole_app:Longident.t ->
-       f0_path:Path.t -> args:(Path.t * Types.module_type) list ->
-       arg_path:Path.t -> arg_mty:module_type -> param_mty:module_type ->
-       t -> unit)
+       f0_path:Path.t -> args:(Path.t * Types.module_type * Mode.Value.l) list ->
+       arg_path:Path.t -> arg_mty:module_type -> arg_mode:Mode.Value.l ->
+       param_mty:module_type -> t -> unit)
 
 let scrape_alias =
   (* to be filled with Mtype.scrape_alias *)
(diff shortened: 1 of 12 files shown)


riaqn:
The latest commit "better lock safety" changes `find_same` to `find_same_and_locks`, because this function could return `value_data` which includes a `vda_mode`, and returning that without locks is just assuming no locks and dangerous.

Instead, we make it return `locks`. We drop the locks at the point where it's clear there is no mode involved. For example, for `module_data` and `class_data`, we drop the locks instantly since they don't contain modes. For `value_data`, we drop the locks at the same point where we drop `vda_mode`.

ocaml/typing/env.ml
```diff
@@ -2400,7 +2426,8 @@ let proj_shape map mod_shape item =
       Shape.Map.add map item shape, Some shape
 
 module Add_signature(T : Types.Wrapped)(M : sig
-  val add_value: ?shape:Shape.t -> Ident.t -> T.value_description -> t -> t
+  val add_value: ?shape:Shape.t -> mode:(Mode.allowed * 'r0) Mode.Value.t -> Ident.t -> T.value_description
```
lpw25, line 2437: Another line that is too long

ocaml/typing/env.ml
```diff
@@ -3298,26 +3378,53 @@ let lookup_all_dot_constructors ~errors ~use ~loc usage l s env =
...
-  | Lident s ->
-      if !Clflags.transparent_modules && not load then
-        fst (lookup_ident_module Don't_load ~errors ~use ~loc s env)
-      else
-        fst (lookup_ident_module Load ~errors ~use ~loc s env)
-  | Ldot(l, s) -> fst (lookup_dot_module ~errors ~use ~loc l s env)
-  | Lapply _ as lid ->
-      let path_f, _comp_f, path_arg = lookup_apply ~errors ~use ~loc lid env in
-      Papply(path_f, path_arg)
+  let path, locks =
```
lpw25, line 3389: It would be better if this were `let path, locks, _ =`, rather than doing that in each branch (assuming there is a suitable dummy value to use in the `Lapply` case).

ocaml/typing/env.ml
```diff
@@ -3298,26 +3378,53 @@ let lookup_all_dot_constructors ~errors ~use ~loc usage l s env =
...
+          path, locks
+    | Ldot(l, s) ->
+        let path, locks, _ = lookup_dot_module ~errors ~use ~loc l s env in
+        path, locks
+    | Lapply _ as lid ->
+        let path_f, _comp_f, path_arg = lookup_apply ~errors ~use ~loc lid env in
+        Papply(path_f, path_arg), []
+  in
+  if use then begin
+    let mda_mode = Mode.Value.legacy |> Mode.Value.disallow_right in
```
lpw25, line 3410: I assume this gets replaced in a later patch. Probably better to define it next to the type definition for the type with the `mda_` fields.

ocaml/typing/env.ml
```diff
@@ -3189,44 +3261,52 @@ and lookup_apply ~errors ~use ~loc lid0 env =
...
+        path, locks, md
+    | Lapply _ as lid ->
+        let path_f, comp_f, path_arg = lookup_apply ~errors ~use ~loc lid env in
+        let md = md (modtype_of_functor_appl comp_f path_f path_arg) in
+        Papply(path_f, path_arg), [], md
+  in
+  if use then begin
+    let mda_mode = Mode.Value.legacy |> Mode.Value.disallow_right in
+    let vmode = walk_locks ~errors ~loc ~env ~item:Module ~lid mda_mode None locks in
+    submode_legacy ~errors ~loc ~env ~item:Module ~lid vmode
```
lpw25, line 3290: I think this should be done by the callers of this function. We don't yet know what the module is being used for, so we don't know what mode is needed.

ocaml/typing/env.ml
```diff
@@ -3153,9 +3220,14 @@ and lookup_all_args ~errors ~use ~loc lid0 env =
...
   let f0_lid, args0 = lookup_all_args ~errors ~use ~loc lid0 env in
   let args_for_errors = List.map (fun (_,p,mty) -> (p,mty)) args0 in
-  let f0_path, f0_comp =
+  let f0_path, f0_locks, f0_comp =
     lookup_module_components ~errors ~use ~loc f0_lid env
   in
+  if use then begin
+    let f0_mode = Mode.Value.legacy |> Mode.Value.disallow_right in
+    let f0_vmode = walk_locks ~errors ~loc ~env ~item:Module ~lid:f0_lid f0_mode None f0_locks in
+    submode_legacy ~errors ~loc ~env ~item:Module ~lid:f0_lid f0_vmode
```
lpw25, line 3237: I don't think we should require that a functor be legacy to be applied. As with ordinary functions, we should allow functors to be applied at any mode and ensure at their definitions that this will be safe.

ocaml/typing/env.ml
```diff
@@ -3339,10 +3446,18 @@ let lookup_modtype ~errors ~use ~loc lid env =
...
-  | Ldot(l, s) -> lookup_dot_class ~errors ~use ~loc l s env
-  | Lapply _ -> assert false
+  let path, locks, cld =
+    match lid with
+    | Lident s -> lookup_ident_class ~errors ~use ~loc s env
+    | Ldot(l, s) -> lookup_dot_class ~errors ~use ~loc l s env
+    | Lapply _ -> assert false
+  in
+  if use then begin
+    let clda_mode = Mode.Value.legacy |> Mode.Value.disallow_right in
```
lpw25, line 3464: This would be better defined near the definition of the `class_data` type.

ocaml/typing/env.ml
```diff
@@ -3339,10 +3446,18 @@ let lookup_modtype ~errors ~use ~loc lid env =
...
+    let vmode = walk_locks ~errors ~loc ~env ~item:Class ~lid clda_mode None locks in
+    submode_legacy ~errors ~loc ~env ~item:Class ~lid vmode
```
lpw25, line 3466: I think it would be better to constrain the mode at the call sites of `lookup_class`.

lpw25 (commented):
Mostly looks good, a few small suggestions

ocaml/typing/env.mli
```diff
@@ -240,16 +245,15 @@ val lookup_error: Location.t -> t -> lookup_error -> 'a
...
+  (** Explains why [mode] is high. *)
+}
 
-    TODO: A better approach is passing down the expected mode to this function
-    as argument, so that sub-moding error is triggered at the place where error
-    hints are immediately available. *)
 val lookup_value:
   ?use:bool -> loc:Location.t -> Longident.t -> t ->
-  Path.t * value_description * Mode.Value.l * shared_context option
+  Path.t * value_description * actual_mode
```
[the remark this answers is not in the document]
riaqn: Similar to `expected_mode` which contains an `closure_context` explaining why the `mode.r` is low; here `actual_mode` contains a `shared_context` explaining why the `mode.l` is high.

I imagine in a not-so-distance future, we will improve mode error hints by enriching both `expected_mode` and `actual_mode`.

ocaml/typing/env.ml
```diff
@@ -452,32 +457,35 @@ module IdTbl =
...
+        shouldn't encounter any locks. *)
     let find_name wrap ~mark name tbl =
-      let (id, ([] : empty list), desc) =
-        find_name_and_modes wrap ~mark name tbl in
-      id, desc
+      match find_name_and_locks wrap ~mark name tbl with
+      | Ok (id, [], desc) -> id, desc
+      | Ok (_, _ :: _, _) -> assert false
+      | Error [] -> raise Not_found
+      | Error (_ :: _) -> assert false
```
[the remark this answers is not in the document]
riaqn: They are indeed wrong. This function should only work for `lock_type = empty`, which statically rules out non-empty lists.

This change is reflected by the commit "[more type safety](https://github.com/ocaml-flambda/flambda-backend/pull/2431/commits/9f2997dd41e335da761fd85ffb54807746a9cc76)"

ocaml/typing/env.ml
riaqn: functors can be partially applied, meaning there are some mode constraints between the partial parameter, the functor, and the resulted module. That sounds like some efforts required out of the scope this PR. If we just assume all modules are legacy whatsoever, then we don't need to worry about those.

ocaml/typing/env.ml
riaqn: Similar to the comment below, I think for the scope of this PR, it's quick and safe to just say all modules are always legacy.

ocaml/typing/env.mli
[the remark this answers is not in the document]
lpw25: I think the fact that all the callers immediately separate the information means this is not currently a good abstraction. If it becomes one in the not-so-distant future we can add it back then.

ocaml/typing/env.ml
lpw25: I find it hard to think about the correctness of the current code. This check might be too strict for some of the uses. If we just pull it out to the right places then it will be clear.

ocaml/typing/env.ml
lpw25: Rather than having to work out the soundness of a different set of typing rules for functors, I think it will be easier to use the same typing rules as for functions. Note that the "close over" morphism, applied to the legacy mode produces the legacy mode. This means that restricting all functor parameters to legacy and all functor returns to legacy is sound by our normal rules, including correctly handling partial application, so there is no need to add any mode constraints.

lpw25: I assume that you meant something like:
```ocaml
let foo @ portable =
  fun () -> 
    let module X = struct ... end in
    let module Z = F(X) in
    ...
```
Just as with ordinary functions, this should be rejected because `F` is not `portable` not because of anything to do with `contended`.

ocaml/typing/env.ml
riaqn: Do you have an example where this might be too strict?

ocaml/testsuite/tests/typing-modes/legacy.ml
```diff
@@ -0,0 +1,116 @@
+(* TEST
+   * expect
+*)
+
+(* This file tests various legacy constraints. *)
```
goldfirere, line 5:
```suggestion
(* This file tests various legacy constructs. *)
```

ocaml/testsuite/tests/typing-modes/legacy.ml
```diff
@@ -0,0 +1,116 @@
+(* TEST
```
goldfirere, line 1: Maybe it belongs here, maybe in a different one: should there be a test that tries to create a new constructor of an extensible type within a portable function? I think that probably should not be allowed, because adding a new constructor of an extensible type causes a mutation of a hidden global variable (and could potentially cause a race).

goldfirere (commented):
Took a quick glance out of curiosity, and thought of a new edge case we might want to consider (though possibly not in this patch).

ocaml/testsuite/tests/typing-modes/legacy.ml
riaqn: Usually such things are modelled by the function closing over a `uncontended` variable. But here, the variable being mutated is implicit, so there would require speical treatment.

Any chance that this global state is already made multithread-safe by ocaml runtime? That's likely considering that exception is a built-in feature and ocaml5 should take care of that.

lthls: The underlying global variable is accessed using atomic primitives in OCaml 5.x. So I think it does not count as a race.

goldfirere: I’d love @lpw25 ’s opinion here, as I don’t know that we have a written-down definition of “race”. (Maybe we do. I just don’t know where to find it.) That is, the extant primitives may prevent direct corruption, but there is still the possibility of unintended inter-thread communication via this channel. Is it possible that some function adds 10 constructors to an extensible type and then assumes that the tags are sequential? Probably not, but it’s conceivable to me.

Basically, my question boils down to: does “race-free” mean that the only inter-thread communication channels are our primitives? or does it only mean that there will be no data corruption as might be caused by tearing non-atomic structures?

ocaml/testsuite/tests/typing-modes/legacy.ml
riaqn: (This PR will rebase, where this `legacy.ml` is splitted into `module.ml` `class.ml` etc., and the commented wording doesn't exist. Therefore, I'm resolving.)

riaqn:
@lpw25 The PR is ready for anothe look now. Please start from the commit "not expose actual_mode".

Then I will rebase and merge.

There are several other soundness bugs I need to fix, but I will do those in other PRs. The current PR is blocking #2685, so should be merged ASAP.

ocaml/testsuite/tests/typing-modes/legacy.ml
riaqn: This is out of the scope of this PR (I want to merge this PR ASAP to avoid blocking others). Let's move this discussion to other venues.

ocaml/typing/env.ml
```diff
@@ -425,25 +425,33 @@ module IdTbl =
...
 
     let map f next =
       {
         current = Ident.empty;
         layer = Map {f; next}
       }
 
-    let rec find_same id tbl =
-      try Ident.find_same id tbl.current
+    let rec find_same_and_locks id tbl macc =
```
lpw25, line 437: I think this is a bad change. No caller of this function ever users the locks, and it would be suspicious if they did. The `find_same` functions are used primarily for looking up types and type-level components from things -- they should never care about locks. Their performance is also important since type lookup is one of the hottest parts of the compiler.

ocaml/jane/doc/proposals/data-race-freedom.md
```diff
@@ -94,11 +94,11 @@ annotation.
...
 which have the following modes ordered by increasing strictness.
 
-| Locality | Sync   | Uniqueness | Linearity  | Readonly  |
-| -------- | ------ | ---------- | ---------- | --------- |
-| Local    | Unsync | Shared     | Once       | Readonly  |
-| Global   | Sync   | Exclusive  | Separate   | Readwrite |
-|          |        | Unique     | Many       |           |
+| Locality | Portable   | Uniqueness | Linearity  | Readonly  |
+| -------- | ---------- | ---------- | ---------- | --------- |
+| Local    | Unportable | Shared     | Once       | Readonly  |
```
lpw25, line 99: This should say `Nonportable` not `Unportable`

ocaml/stdlib/array.mli
```diff
@@ -33,7 +33,7 @@ open! Stdlib
 type 'a t = 'a array
 (** An alias for the type of arrays. *)
 
-external length : 'a array -> int = "%array_length"
```
lpw25, line 36: I'm not sure we should be making this change. I think it might break attempts to build opam, or even just things in our tree. We mostly only ever use `[@local_opt]` or similar in the stdlib. Probably we should add `[@contended_opt]` or similar and do the same -- although not in this PR.

ocaml/typing/env.ml
```diff
@@ -749,6 +772,16 @@ let error err = raise (Error err)
 let lookup_error loc env err =
   error (Lookup_error(loc, env, err))
 
+type actual_mode = {
```
lpw25, line 775: No longer used

ocaml/typing/env.ml
```diff
@@ -3145,40 +3211,47 @@ and lookup_all_args ~errors ~use ~loc lid0 env =
...
-  let args_for_errors = List.map (fun (_,p,mty) -> (p,mty)) args0 in
-  let f0_path, f0_comp =
+  let args_for_errors = List.map (fun (_,p,mty,vmode) -> (p,mty,vmode.mode)) args0 in
+  let f0_path, f0_locks, f0_comp =
     lookup_module_components ~errors ~use ~loc f0_lid env
   in
-  let check_one_apply ~errors ~loc ~f_lid ~f_comp ~arg_path ~arg_mty env =
+  if use then begin
+    walk_locks ~errors ~loc ~env ~item:Module ~lid:f0_lid mda_mode None f0_locks
+    |> ignore
```
lpw25, line 3229: I think that `lookup_apply` is only ever used when looking up types and module types. At the moment I think you will error for this:
```ocaml
module M = struct end
module F (M : sig end) = struct type t end
let use_portable (f @ portable) = f ()
let _ = use_portable (fun () -> let module _ = struct type t = F(M).t end in ())
```
I think you can just not walk the locks here.

lpw25 (commented):
Some suggestions and a bug

ocaml/testsuite/tests/typing-modes/legacy.ml
```diff
@@ -26,7 +26,7 @@ let (foo @ portable) () =
 Line 2, characters 20-21:
 2 |     let _ = (module M : S) in
                         ^
-Error: The module M is unportable, so cannot be used inside a closure that is portable.
+Error: Modules are unportable, so cannot be used inside a closure that is portable.
```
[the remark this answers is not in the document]
riaqn: This will be fixed after rebasing (I didn't rebase for incremental reviewing)

ocaml/jane/doc/proposals/data-race-freedom.md
riaqn: This was fixed in #2634

ocaml/testsuite/tests/typing-modes/legacy.ml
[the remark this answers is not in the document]
riaqn: Ugh I thought this was fixed in a previous PR, so I didn't change in this PR because I thought we just need a rebase. But turned out `env.ml` has its own copy of mode printing. I will fix this; in the long run we should have a single copy of mode-printing.

ocaml/typing/env.ml
riaqn: The name nor the type of this function suggests that it's only used for looking up types. So I think it's more "correct" to not ignore locks.

In terms of performance: when this function is used for looking up types, the `macc` will stay as `[] : empty list`, which I think is an acceptable cost for safety. The list being `[]` is guaranteed by `empty` being uninhabited.

This is also similar to `find_name` below.

ocaml/typing/env.ml
riaqn: Yes you are right. I looked at `parser.mly` and `Lapply` can only be generated for "compile-time" module application. For runtime module application (where a pointer to the functor is needed), `Pmod_apply` is used instead.

I was under the wrong impression that we can reuse the `~use` argument is to distinguish between the two usages. This is also why in a previous discussion I insisted that `Env.lookup_* ~use:true` is a "real" usage and should just walk the locks and enforcing legacy. This is now obviously wrong for modules.

ocaml/typing/env.ml
riaqn: I still find it useful to have this type internal to `env.ml`, even just to reduce the number of arguments.

ocaml/typing/env.ml
lpw25: It does somewhat tell you that, because it is looking for an `Ident.t` not a string. You are mostly only going to do that to lookup the definitions of types.

> the macc is statically guaranteed to stay at [] : empty list

That is only true if you are looking up something like `foo`, not if you are looking up `Foo.t` which is the more common case.

If you prefer, you can rename this function `find_same_without_locks` to make it clear. That is in fact clearer than `find_same_and_locks |> fst`. You can continue to expose it as `find_same` for the case where the type ensures that there are no locks.

ocaml/typing/env.ml
lpw25: Ok, if you really want to keep this type then probably the thing to do is revert the change I requested so it's still in the interface. Then rename it something like `identifier_mode` so it has a name with a clearer semantic meaning and then add versions of `submode` and `mode_cross_left` that operate directly on it.

That will fix my criticism of the earlier code. Previously the pair was only ever immediately deconstructed by the callers of `lookup_value`, which implies that it wasn't being used as an actual abstraction. If they instead start to treat it like a real value then it is being used as an actual abstraction.

ocaml/stdlib/array.mli
riaqn: This change *was* in #2398 but removed due to similar concerns. The current PR will not contain this change after rebase.

riaqn: I think the options were:
- Not do anything, because our internal base/core will have the proper interface anyway.
- Define those changed functions in `otherlibs`.

ocaml/typing/env.ml
riaqn: I agree with all what you said.

ocaml/typing/env.ml
riaqn: Yes I also thought about adding new version of `submode` and `mode_cross_left`, but felt they will be overhauled soon with a future refactoring of mode errors. So I didn't do that which left the abstraction not fully utilized.

I still find the simple packing useful - that being said, I will remove the packing to get the PR merged soon.

riaqn: Put it in another way: we already have many functions taking and returning this `actual_mode` - is that sort of abstraction? Then, I think partial abstraction is better than no abstraction.

ocaml/typing/env.ml
riaqn: Well in the example, the look-up of `M` must also not cause walking locks. The proper way is to add `~lock` to `lookup_module` and siblings. I believe module is the only thing that needs it - for other categories, `~use:true` means needing walking locks.

riaqn: Ok there are some other cases where a reference to a module should not trigger locks. This is fixed and tested in the commit "some ref to module should not trigger locks".

lpw25 (approved):
LGTM

ocaml/typing/env.ml
riaqn: By "two lines long" I guess you mean to define them as wrappers of the existing functions in `typecore.ml`, which I have done now in the commit "more abstraction for actual_mode".

In a future PR I will move `actual_mode`, `expected_mode` and related functions to a single module that sits between `env/typecore` and `mode.ml`.

Issue #2433, opened 2024-04-11
Build instructions in HACKING.adoc fail with Menhir-related errors


OlivierNicole:
The current instructions in HACKING.adoc do not seem to work. I followed them closely, and after the configure step, doing `make _install` yields errors about the menhir executable not being in PATH. Installing menhir and trying again gives what seems to be a version error:
```
RUNTIME_DIR=runtime4 /home/olivier/.opam/4.14.2/bin/dune build --root=. --workspace=duneconf/boot.ws  ocaml/main_native.exe boot_ocamlopt.exe ocaml/tools/ocamlmklib.exe ocaml/tools/ocamldep.exe tools/flambda_backend_objinfo.exe
File "middle_end/flambda2/parser/flambda_parser.ml", line 5, characters 2-54:
5 |   CamlinternalMenhirLib.StaticVersion.require_20210419
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error: Unbound value CamlinternalMenhirLib.StaticVersion.require_20210419
File "ocaml/parser.pp.ml", line 5, characters 2-42:
5 |   MenhirLib.StaticVersion.require_20210419
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error: Unbound value MenhirLib.StaticVersion.require_20210419
make: *** [ocaml/Makefile.common-jst:66: boot-compiler] Error 1
```
And this, even if I pin Menhir to version 20210419.

goldfirere:
I think I've hit that. Clean out your build products with `git clean -Xdf` and try again, now that you've pinned menhir.

OlivierNicole:
Thanks, that did work. The mention that (a specific version of) menhir is needed is missing though, I’ll open a PR about that.

OlivierNicole:
Fixed by #2435.

PR #2435, opened 2024-04-12
Add missing mention of menhir in build instructions

OlivierNicole:
See #2433.


lthls:
#2005 would have avoided the dependency on a specific menhir, but it's still open.

PR #2442, opened 2024-04-15
Enable locals for arm64

mshinwell:
The majority of this work was kindly done by @anmolsahoo25 in #1691.  I've only taken the locals part here and only what is required to make flambda-backend compilers build with locals enabled, not the upstream ones.  I have made a few improvements and fixes.

I have not yet tested the runtime5 implementation of `caml_call_local_realloc`, that can be done later once the stack checks etc have been enabled for arm64.  The `typing-local/` directory passes on runtime4 with this PR on arm64.


xclerc (approved):
(Actually read it while I was reviewing #2447)

PR #2447, opened 2024-04-15
Stack checks for arm64

mshinwell:
This starts with #2442, #2443 and #2446.  The remaining `BACKPORT` markers (except for one about DLS which will be dealt with in a separate PR) are removed from the arm64 emitter.  I've then attempted to enable stack checks by wiring in the new `Stack_check` infrastructure.


xclerc (approved):
I am low confidence, but did not spot any
issues.

(Not right now, but when there are `arm64`
runner for the CI, I think we should consider
having a job testing `arm64` with the debug
runtime.)

PR #2453, opened 2024-04-16
Better handling of multiple simultaneous async exns

stedolan:
If multiple asynchronous exceptions arise between the same pair of safepoints, `Sys.with_async_exns` will currently handle _neither_. The problem is that when attempting to raise the first, a second poll point occurs and the second exception is raised asynchronously instead, causing a fatal error (unless there is a second enclosing `with_async_exns`).

The fix here is to drain the queue of pending actions, by running pending actions until they stop raising asynchronous exceptions.


mshinwell:
Looks good, can we have a runtime5 implementation too?

stedolan:
> Looks good, can we have a runtime5 implementation too?

Done, forgot this was currently supported there too.

mshinwell (approved):

PR #2459, opened 2024-04-16
Zero alloc: propagate assume on applications of externals

gretay-js:
Handle "assume" on calls to external functions and primitives similarly to the way we handle assume on calls to ocaml functions.  

The change is simple but it would affect all primitives, including builtins - not sure if that's right.

Added a test.


ocaml/lambda/translcore.ml
```diff
@@ -430,8 +430,12 @@ and transl_exp0 ~in_new_scope ~scopes sort e =
         else Rc_normal
       in
       let lam =
+        let loc =
+          map_scopes (update_assume_zero_alloc ~assume_zero_alloc)
+            (of_location ~scopes e.exp_loc)
+        in
```
ccasin, line 436: I think this is right, but it's a little complicated, so to check my understanding:

This is particularly interesting in the case that `extra_args` is not empty.  I think in that case we are over-applying a primitive.  So, for example: `(Sys.opaque_identity[@zero_alloc]) (fun x -> x) 42`.  This case of translcore is first compiling the primitive application, then applying that to the rest of the args.  Here you are updating the scopes for the first part, and the below call to `transl_apply` in this case will do so for the rest application.  I suppose this matches the user's intent to treat the whole application as zero_alloc.

ccasin (commented):
> The change is simple but it would affect all primitives, including builtins - not sure if that's right.

I think this is all fine, see my comment in the code.  But I think we should have some tests for some builtins documenting the behavior.  Here is one I was playing with:
```ocaml
external revapply : 'a -> ('a -> 'b) -> 'b = "%revapply"

let[@inline never] id x = x

let[@zero_alloc] overapplied g x = (revapply[@zero_alloc assume]) id g x
```
This works (the zero_alloc check passes).  Which I think is what we want, even if `g` may allocate, since the user has added the assume.  But see what you think.

ocaml/lambda/translcore.ml
gretay-js: I don't really understand what this code does, so thank you for the details. At the high level, yes, the intended meaning in the case of an overapplication is that "assume" applies to the rest of the arguments as well. I added your example and a couple of variations that show this.

gretay-js:
Added a few tests for % builtins.

ccasin (approved):

xclerc (approved):

PR #2461, opened 2024-04-17
Statmemprof support for runtime5

stedolan:
This patch imports the following upstream PRs:

  - #12381 Multicore statmemprof API
  - #12382 Basic thread and domain structures for statmemprof
  - #12817 Fix assertion failure (bugfix for #12382)
  - #12824 Fix two assertion failures introduced by recent merge (bugfix for #12382)
  - #12383 Improve backtrace abstractions inside runtime
  - #12923 Statmemprof resurrected
  - #13068 Minor post-merge improvements to statmemprof

Additionally, there are three extra small commits not imported from upstream:
  - to make the new statmemprof testsuite pass on flambda-backend (some robustness fixes that will be uptreamed, and updates to handle flambda-backend's nonstandard async exn setup)
  - to make the new statmemprof testsuite pass on runtime4 (mostly, by disabling the bits that test new 5-only features)
  - to avoid allocations during bytecode callbacks, which ensures that with_async_exns in flambda-backend does not allocate in bytecode


mshinwell:
I'll do the review.

mshinwell:
I think I've reached a reasonable level of confidence on this.  I have a couple of questions but they are more about the workings of memprof than the safety of merging this patch when memprof is off.

mshinwell (approved):

PR #2463, opened 2024-04-17
New strategy for safepoint invocation

mshinwell:
This uses a load or store from a protected memory page to trigger entry into the runtime, instead of the existing safepoint mechanism.  The code sequence is consequentially rather shorter.  When waking up domains, we first wait, hoping that each domain will hit an allocation point (or some other means of entering the runtime, but not a safe point).  If that fails, the memory protection is applied to the safepoint trigger page.

This is currently only a sketch and needs testing, not yet ready for review.


mshinwell:
Now based on #2464

mshinwell:
This might be in a working state...

mshinwell:
I neglected to mention yesterday that some pieces of code here were taken from #2404 written by @xclerc

mshinwell:
Closing because @NickBarnes is going to take this over.

PR #2464, opened 2024-04-17
Enable Domain module fully on runtime5

mshinwell:
This enables the `Domain` module fully on runtime5.  One consequence of this is that the domain-local state (`Domain.DLS` ) module changes under runtime5 to the upstream runtime5 implementation, rather than what we are using for runtime4.

Not yet ready for review, will mark as draft for now.  (The diff must be viewed in a whitespace-insensitive mode, though.)

Based on #2205 to avoid conflicts

### DIFF
--- a/ocaml/middle_end/semantics_of_primitives.ml
+++ b/ocaml/middle_end/semantics_of_primitives.ml
@@ -160,7 +160,7 @@ let for_primitive (prim : Clambda_primitives.primitive) =
   | Pget_header _ -> No_effects, No_coeffects
   | Pdls_get ->
       (* only read *)
-      No_effects, No_coeffects
+      No_effects, Has_coeffects
 
 type return_type =
   | Float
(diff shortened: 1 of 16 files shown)


mshinwell:
@Ekdohibs could you please review this?  For the stdlib `Domain` module, probably best to compare against 5.1.1 upstream.

ocaml/middle_end/semantics_of_primitives.ml
```diff
@@ -160,7 +160,7 @@ let for_primitive (prim : Clambda_primitives.primitive) =
   | Pget_header _ -> No_effects, No_coeffects
   | Pdls_get ->
       (* only read *)
-      No_effects, No_coeffects
```
Ekdohibs, line 163: Why was this changed?

mshinwell: Hmm, well I think this operation has coeffects - it's a memory load?

Ekdohibs: I checked upstream and it's `No_coeffects` there; I think it's correct because while it's a memory load, it's never to a location that is written to.

mshinwell:
Adding the `to_upstream` tag for the change in `ocaml/middle_end/semantics_of_primitives.ml`

PR #2466, opened 2024-04-17
Improve SIMD constant error messages

TheNumbat:
Passing a non-constant argument to any of the SIMD constant constructor externals now raises a proper error instead of `Misc.fatal_errorf`.

### DIFF
--- a/backend/cmm_builtins.ml
+++ b/backend/cmm_builtins.ml
@@ -200,6 +200,8 @@ let rec const_int_args n args name =
   match n, args with
   | 0, [] -> []
   | n, Cconst_int (i, _) :: args -> i :: const_int_args (n - 1) args name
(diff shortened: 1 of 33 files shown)


ccasin:
Looks fine, but please add a test showing the output

ccasin (approved):

ccasin:
Hmm, I clicked the approve button, but actually looking closer I'm not in love with the way these errors have bad locations.  Maybe you should just print an error explicitly to avoid claiming the error happens on line 1?

TheNumbat:
It's already generating an invalid line number, but it seems the error printer always converts it to `1`: https://github.com/ocaml-flambda/flambda-backend/blob/main/ocaml/parsing/location.ml#L301

PR #2469, opened 2024-04-18
Remove lies about mixed block shape in lambda and flambda

ncik-roberts:
Two changes related to tracking mixed block shapes **more honestly**. Previously, in lambda and flambda, we claimed in some ways that the fields of a mixed blocks were values.

The way we remedy that lie is different in lambda vs. flambda:
  * In lambda, we fix the inaccuracy by **accurately tracking the layout of fields** in the flat suffix of mixed blocks. I modify the `Pvariant` value kind to track block mixedness.
  * In flambda, we fix the inaccuracy by **purposefully being vaguer** about the information we track. See `middle_end/flambda2/kinds/flambda_kind.ml`: We decide that the shape of the record as a whole is `any_value` in the event that any constructor of the variant is mixed. (For mixed records, this just means that the record itself is mixed.)

We'll eventually need to fix flambda2 by being more accurate, so that optimizations can apply.

Note that both of these give us a framework for being more accurate about flat float records (including ones with all-unboxed-float fields), but I don't do that in this PR.

To test, I add a test that crashes in the present `main` branch. (A case where the flambda middle end attempts to apply the unboxing optimization in `main`, but after this PR it doesn't attempt to optimize.)


ccasin (approved):
Looks good to me.

I read the middle-end changes too, and they seem fine, but probably @mshinwell or a person he names should confirm using `any_value` in this context won't lead to any problems.

Ekdohibs (approved):
This looks good to me. I was a bit unsure at the beginning of `constructor_shape` instead of using `layout list`, but I feel it is actually better that way.  In the future we might want to add `Constructor_float_block` as an additional constructor and remove all the special-casing that exists for float arrays; and use constructor shapes in flambda2 as well, but this can wait for another PR.
The use of `any_value` currently in flambda2 is correct but will limit optimization opportunities on mixed blocks; this won't affect code that does not use mixed blocks however, so this seems good.

PR #2470, opened 2024-04-18
`[@zero_alloc]` in signatures (part 2 of 3)

ccasin:
This PR allows writing `[@zero_alloc]` in signatures.  Any implementing module is required to have a `[@zero_alloc]` attribute on the corresponding definition.  This is the second of three PRs in the "zero_alloc in signatures" series.  The final PR will allow other compilation units to make use of the information this PR adds to signatures.

What does `[@zero_alloc]` in a signature mean about a corresponding value in a module? One wants it to mean that the implementation must be zero alloc. Of course, the typechecker is too early to check that property accurately. So really what is means, from the perspective of the typechecker, is that the corresponding value is annotated such that it will be checked for zero_alloc in the backend. Which works out to the same thing for the user.

The annotations need not match exactly. For example, if the value is annotated such that it will be checked "more strictly" than the signature implies, that's fine. Details in a comment at the relevant place in the code.

The main complication is arity. The way we will use this information in the final PR relies on our ability to determine when a function is fully applied, even if all we know about it is the `value_description` we got from a `cmi` file. As such, we record the arity that we understand the function to have, based on the number of arrows in its type, in the `value_description`, and check that against the syntactic arity of the function.  The user can override the arity we would calculate from the type (useful, for example, when the type is an alias).

The actual implementation is straightforward. We make `zero_alloc` part of value descriptions in module types (not part of function types proper).  There are two bits that are a little complex mainly due to constraints of the existing code:
*  In typemod, where we actually need to find out whether a pattern variable that we're making a `value_description` for has a `zero_alloc` annotation. I made a slightly gross (in my mind) change to `Typedtree.let_bound_idents_with_modes_and_sorts` to support this, which seemed the only solution short of a substantial refactoring of how pattern variables are treated here.
* The parsing of the optional `(arity n)` in `zero_alloc` payloads is unpleasant. We considered a few other ways to write this, but this was the only one that seemed parseable at all. The existing code was already unpleasant (these payloads are interpreted as function applications by the ocaml parser and then re-interpreted by `builtin_attributes`).  If the reviewers insist that now is the moment to burn it down and try an entirely different approach, I'll relent.

Review: @ncik-roberts to review the core of the typechecker changes, and the design.  @gretay-js also to review the design.  @Ekdohibs will need to sign off for the chamelon changes.

There are two commits.  The first just makes it so there is only one place we translate zero_alloc attributes to the abstract domain, rather than having two versions that could get out of sync.  @gretay-js does this conflict with your other in-flight PRs?  We could undo it if so.  The second has the bulk of the changes.

### DIFF
--- a/ocaml/parsing/builtin_attributes.ml
+++ b/ocaml/parsing/builtin_attributes.ml
@@ -685,12 +685,14 @@ type check_attribute =
   | Check of { property: property;
                strict: bool;
                opt: bool;
+               arity: int;
                loc: Location.t;
              }
   | Assume of { property: property;
                 strict: bool;
-                loc: Location.t;
                 never_returns_normally: bool;
+                arity: int;
+                loc: Location.t;
               }
 
 let is_check_enabled ~opt property =
@@ -808,32 +810,113 @@ let parse_ids_payload txt loc ~default ~empty cases payload =
       | Some r -> r
       | None -> warn ()
 
-let parse_property_attribute attr property =
+(* Looks for `(arity n)` in payload. If present, this returns `n` and an updated
+   payload with `(arity n)` removed *)
+let parse_arity payload =
+  let open Parsetree in
+  let is_arity exp =
+    match exp.pexp_desc with
+    | Pexp_apply (exp_fun, [Nolabel, exp_arg]) ->
+      begin match exp_fun.pexp_desc, exp_arg.pexp_desc with
+      | Pexp_ident {txt = Lident "arity"; _},
+        Pexp_constant (Pconst_integer (s,None)) -> Some (int_of_string s)
+      | _ -> None
+      end
+    | _ -> None
+  in
+  let rec find_arity_in_args acc args =
+    (* Scan a list of arguments for one that is the arity clause.  If found,
+       return the arity and the rest of the args. *)
+    match args with
+    | [] | ((Labelled _ | Optional _), _) :: _ ->
+      (* If the payload contains any labeled or optional args, it will be
+         rejected later, and we just leave it alone. *)
+      None
+    | (Nolabel, exp1) as arg1 :: args ->
+      begin match is_arity exp1 with
+      | Some n -> Some (n, (List.rev acc) @ args)
+      | None -> find_arity_in_args (arg1 :: acc) args
+      end
+  in
+  match payload with
+  | PStr [{pstr_desc = Pstr_eval (exp, []); pstr_loc}] ->
+    let new_payload pexp_desc =
+      let new_exp =
+        { pexp_desc; pexp_loc = exp.pexp_loc; pexp_loc_stack = [];
+          pexp_attributes = [] }
+      in
+      PStr [{pstr_desc = Pstr_eval (new_exp, []); pstr_loc}]
+    in
+    begin match is_arity exp with
+    | Some n ->
+      (* payload is (arity n) by itself *)
+      Some (n, PStr [])
+    | None ->
+      begin match exp.pexp_desc with
+      | Pexp_apply (exp, args) ->
+        begin match find_arity_in_args [] ((Nolabel, exp) :: args) with
+        | Some (n, (Nolabel, exp1) :: args) ->
+          (* payload includes (arity n) somewhere *)
+          let new_exp =
+            if List.length args = 0
+            then exp1.pexp_desc
+            else Pexp_apply (exp1, args)
+          in
+          Some (n, new_payload new_exp)
+        | Some (_, ((Labelled _ | Optional _), _) :: _) -> None
+        | Some (_, []) ->
+          Misc.fatal_error "Builtin_attributes.parse_arity: illegal application"
+        | None -> None
+        end
+      | _ -> None
+      end
+    end
+  | _ -> None
+
+let parse_property_attribute ~is_arity_allowed ~default_arity attr property =
   match attr with
   | None -> Default_check
-  | Some {Parsetree.attr_name = {txt; loc}; attr_payload = payload}->
+  | Some {Parsetree.attr_name = {txt; loc}; attr_payload = payload} ->
+      let payload, arity =
+        match parse_arity payload with
+        | None -> payload, default_arity
+        | Some (user_arity, payload) ->
+          if is_arity_allowed then
+            payload, user_arity
+          else
+            (warn_payload loc txt
+               "The \"arity\" field is only supported on \"zero_alloc\" in \
+                signatures";
+             payload, default_arity)
+      in
       parse_ids_payload txt loc
         ~default:Default_check
-        ~empty:(Check { property; strict = false; opt = false; loc; } )
+        ~empty:(Check { property; strict = false; opt = false; arity; loc; } )
         [
-          ["assume"],
-          Assume { property; strict = false; never_returns_normally = false; loc; };
-          ["strict"], Check { property; strict = true; opt = false; loc; };
-          ["opt"], Check { property; strict = false; opt = true; loc; };
-          ["opt"; "strict"; ], Check { property; strict = true; opt = true; loc; };
-          ["assume"; "strict"],
-          Assume { property; strict = true; never_returns_normally = false; loc; };
-          ["assume"; "never_returns_normally"],
-          Assume { property; strict = false; never_returns_normally = true; loc; };
-          ["assume"; "never_returns_normally"; "strict"],
-          Assume { property; strict = true; never_returns_normally = true; loc; };
-          ["ignore"], Ignore_assert_all property
+          (["assume"],
+           Assume { property; strict = false; never_returns_normally = false;
+                    arity; loc; });
+          (["strict"],
+           Check { property; strict = true; opt = false; arity; loc; });
+          (["opt"], Check { property; strict = false; opt = true; arity; loc; });
+          (["opt"; "strict"; ],
+           Check { property; strict = true; opt = true; arity; loc; });
+          (["assume"; "strict"],
+           Assume { property; strict = true; never_returns_normally = false;
+                    arity; loc; });
+          (["assume"; "never_returns_normally"],
+           Assume { property; strict = false; never_returns_normally = true;
+                    arity; loc; });
+          (["assume"; "never_returns_normally"; "strict"],
+           Assume { property; strict = true; never_returns_normally = true;
+                    arity; loc; });
+          (["ignore"], Ignore_assert_all property)
         ]
         payload
 
-let get_property_attribute l p =
+let get_property_attribute ~is_arity_allowed ~default_arity l p =
   let attr = find_attribute (is_property_attribute p) l in
-  let res = parse_property_attribute attr p in
+  let res = parse_property_attribute ~is_arity_allowed ~default_arity attr p in
   (match attr, res with
    | None, Default_check -> ()
    | _, Default_check -> ()
--- a/ocaml/parsing/builtin_attributes.ml
+++ b/ocaml/parsing/builtin_attributes.ml
@@ -753,16 +753,23 @@ let get_id_from_exp =
   | { pexp_desc = Pexp_ident { txt = Longident.Lident id } } -> Result.Ok id
   | _ -> Result.Error ()
 
-let get_ids_from_exp exp =
+let get_id_or_constant_from_exp =
+  let open Parsetree in
+  function
+  | { pexp_desc = Pexp_ident { txt = Longident.Lident id } } -> Result.Ok id
+  | { pexp_desc = Pexp_constant (Pconst_integer (s,None)) } -> Result.Ok s
+  | _ -> Result.Error ()
+
+let get_ids_and_constants_from_exp exp =
   let open Parsetree in
   (match exp with
    | { pexp_desc = Pexp_apply (exp, args) } ->
-     get_id_from_exp exp ::
+     get_id_or_constant_from_exp exp ::
      List.map (function
-       | (Asttypes.Nolabel, arg) -> get_id_from_exp arg
+       | (Asttypes.Nolabel, arg) -> get_id_or_constant_from_exp arg
        | (_, _) -> Result.Error ())
        args
-   | _ -> [get_id_from_exp exp])
+   | _ -> [get_id_or_constant_from_exp exp])
   |> List.fold_left (fun acc r ->
     match acc, r with
     | Result.Ok ids, Ok id -> Result.Ok (id::ids)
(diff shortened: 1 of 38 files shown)


ocaml/testsuite/tests/typing-zero-alloc/signatures.ml
```diff
@@ -0,0 +1,821 @@
...
+module type S8_4 =
+  sig val f : int -> int [@@zero_alloc strict opt (arity 42)] end
+module type S8_5 =
+  sig val f : int -> int [@@zero_alloc strict opt (arity 42)] end
+module type S8_6 =
+  sig val f : int -> int [@@zero_alloc strict opt (arity 42)] end
+|}]
+
+(****************************************************)
+(* Test 9: (arity n) in structures gives warning 47 *)
```
ncik-roberts, line 764: Why? It would be kind of nice to check the arity in the impl. Less difference between the language of zero_alloc for intfs vs. impls.

ocaml/testsuite/tests/typing-zero-alloc/signatures.ml
```diff
@@ -0,0 +1,821 @@
+(* TEST
```
ncik-roberts, line 1: Other maybe interesting test cases:
  * `module type of` a structure that uses zero_alloc? I'm prepared for the answer to be "the attributes are dropped in the inferred signature" but even that feels worth recording in a test.
  * A test with a functor where the zero-allocness of the impl depends on the zero-allocness of functions in the param's signature? I'm unsure if this is supported yet. Regardless another test file might be a better home than this one.

ncik-roberts (commented):
Just review of the tests so far.

ocaml/testsuite/tests/typing-zero-alloc/signatures.ml
```diff
@@ -0,0 +1,821 @@
...
+
+   These tests are just about what is allowed and not allowed by the
+   typechecker.  The implementation of the actual `[@zero_alloc]` backend checks
+   (including how the annotations in signatures affect those checks) are tested
+   in the `tests/backend/checkmach` directory at the root of the project.
+*)
+
+(*******************************************)
+(* Test 1: Allowed and disallowed payloads *)
+module type S1_1 = sig
```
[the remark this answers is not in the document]
ccasin: I agree the numbers are a bit annoying.  On the other hand, I strongly dislike when the test programs use the same name repeatedly - I find it makes them hard to navigate, and have occasionally spent time staring at the wrong test.  And I'm bothered less by the numbers getting a bit wonky, since in my mind they exist just to be unique.

In any case, here it's reasonably easy to come up with mostly unique names, which seems obviously better, so I've done that.  (But I'd like to stop short of promising never to do this again if it proves harder to generate good names in a future case)

ocaml/testsuite/tests/typing-zero-alloc/signatures.ml
```diff
@@ -0,0 +1,821 @@
...
+6 |   let f x = x
+7 | end
+Error: Signature mismatch:
+       Modules do not match: sig val f : 'a -> 'a end is not included in S3_1
+       Values do not match:
+         val f : 'a -> 'a
+       is not included in
+         val f : 'a -> 'a [@@zero_alloc]
+       The former provides a weaker "zero_alloc" guarantee than the latter
+|}]
```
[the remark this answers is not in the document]
ccasin: :)

It was easy enough to add a hint for the case where the attribute is entirely absent in the implementation, so I've done that.  Good idea.

Because the module inclusion check happens somewhat distantly from the attribute parsing on a more abstract version of the information in the attribute, it would be harder to provide exactly the payload the user needs to write, or to give a hint in cases where the attribute is present but just wrong. But I think that's fine - the new hint improves the case most likely to send users to Rust.

ocaml/testsuite/tests/typing-zero-alloc/signatures.ml
```diff
@@ -0,0 +1,821 @@
...
+       Values do not match:
+         val f : 'a -> 'a
+       is not included in
+         val f : 'a -> 'a [@@zero_alloc opt]
+       The former provides a weaker "zero_alloc" guarantee than the latter
+|}]
+
+module type S3_3 = sig
+  val[@zero_alloc strict] f : 'a -> 'a
+end
```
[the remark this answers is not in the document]
ccasin: Well spotted.  Added.

ocaml/testsuite/tests/typing-zero-alloc/signatures.ml
```diff
@@ -0,0 +1,821 @@
...
+module type S4_1 = sig
+  val[@zero_alloc] x : int
+end
+[%%expect{|
+Line 2, characters 2-26:
+2 |   val[@zero_alloc] x : int
+      ^^^^^^^^^^^^^^^^^^^^^^^^
+Error: In signatures, zero_alloc is only supported on function declarations.
+       Found no arrows in this declaration's type.
+       Hint: You can add "(arity n)" to specify the arity of an alias.
```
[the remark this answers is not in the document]
ccasin: Added

ocaml/testsuite/tests/typing-zero-alloc/signatures.ml
```diff
@@ -0,0 +1,821 @@
...
+Line 2, characters 2-34:
+2 |   val[@zero_alloc] f : t5_two_args
+      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Error: In signatures, zero_alloc is only supported on function declarations.
+       Found no arrows in this declaration's type.
+       Hint: You can add "(arity n)" to specify the arity of an alias.
+|}]
+
+module type S5_3 = sig
+  val[@zero_alloc (arity 2)] f : t5_two_args
```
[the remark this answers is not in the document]
ccasin: Well, the fact that ocamlformat would turn good programs into bad programs seems like a pretty compelling reason! I've adopted your suggested syntax.

ocaml/testsuite/tests/typing-zero-alloc/signatures.ml
ccasin: In signatures, writing arity overrides the arity we would otherwise infer.  In structures, it could only be a check.  To me, it seemed potentially confusing to have it mean two different things in the two places, and therefore better to just disallow it.

Also, it feels odd to give users a way to check the arity of their funtions, but only if they are zero alloc.  (I already feel this way a bit about the signatures, but there at least there's a clear reason you might need it in combination with zero_alloc.)  Maybe we should just add a separate attribute for this, if we think it is useful?

ocaml/testsuite/tests/typing-zero-alloc/signatures.ml
ccasin: Right, I meant to add `module type of` tests, and have now done so. It works well. The fact that you can't sneak `assume` or `assume never_returns_normally` into a signature with it motivated parts of the details of the current implementation.

As for the functor case: There is currently no way for the signatures to influence the zero alloc checking of implementations, beyond requiring that such a check might happen.  So I don't think we can do your example.  But I will add lots of tests like this in the next PR.

Ekdohibs (approved):
For `chamelon/` changes.

ccasin:
I've responded to the initial feedback on tests and rebased this.  I also pushed a small fix and test that checks the interaction between these attributes in signatures and warning 199, which was broken (warning 199 would always fire, since there is no function to be checked).

ocaml/testsuite/tests/typing-zero-alloc/signatures.ml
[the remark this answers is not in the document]
ncik-roberts: Thanks!

To be clear, I don't actually think that ocamlformat would have turned good programs into bad programs. `[@zero_alloc (arity 2)]` parses exactly the same as `[@zero_alloc arity 2]`. It's just that the old style required inconsistent parenthesization, depending on whether `arity 2` was the first or later-than-first argument, so is harder to describe in e.g. error messages.

ocaml/testsuite/tests/typing-zero-alloc/signatures.ml
ncik-roberts: Fair enough. If we ever add such an attribute, we can remove `arity` from `zero_alloc` and have `zero_alloc` look for a separate `[@@arity n]` attribute instead, even in signatures.

ocaml/typing/typedtree.ml
```diff
@@ -1035,13 +1035,20 @@ let rev_let_bound_idents_full bindings =
...
+    Ident.Tbl.add modes_and_sorts id
+      (sloc.loc, mode, sort, Builtin_attributes.Default_check)
   in
   List.iter (fun vb ->
-    iter_pattern_full ~both_sides_of_or:true f vb.vb_sort vb.vb_pat)
+    iter_pattern_full ~both_sides_of_or:true f vb.vb_sort vb.vb_pat;
+     match vb.vb_pat.pat_desc, vb.vb_expr.exp_desc with
+     | Tpat_var (id, _, _, _), Texp_function fn ->
+       let (loc, mode, sort, _) = Ident.Tbl.find modes_and_sorts id in
+       Ident.Tbl.replace modes_and_sorts id (loc, mode, sort, fn.zero_alloc)
```
ncik-roberts, line 1049: This `replace` makes me nervous. It modifies the most recently `add`-ed binding, and there may be multiple bindings added by `iter_pattern_full` (as evidenced by the `Ident.Tbl.find_all` below). When can there be multiple entries for the same ident?

ocaml/typing/typemod.ml
```diff
@@ -2807,15 +2807,30 @@ and type_structure ?(toplevel = None) funct_body anchor env sstr =
...
+                match[@warning "+9"] modes with
+                | [(_, _, _, (Default_check | Ignore_assert_all _))] ->
+                  Default_check
+                | [(_, _, _, (Check _ as zero_alloc))] -> zero_alloc
+                | [(_, _, _, Assume { property; strict; arity; loc;
+                                      never_returns_normally = _ })] ->
+                  Check { strict; property; arity; loc; opt = false }
+                | _ -> Default_check
+              in
+              let (first_loc, _, _, _) = List.hd modes in
```
ncik-roberts, line 2824: I am struggling to review this because I don't know what it means for `modes` to contain more than one element. See my other comment about multiple bindings in `let_bound_idents_with_modes_sorts_and_checks`.

ocaml/parsing/builtin_attributes.ml
```diff
@@ -808,43 +810,116 @@ let parse_ids_payload txt loc ~default ~empty cases payload =
       | Some r -> r
       | None -> warn ()
 
-let parse_property_attribute attr property =
+(* Looks for `arity n` in payload. If present, this returns `n` and an updated
+   payload with `arity n` removed *)
+let parse_arity payload =
```
ncik-roberts, line 815: A lot of the grossness in this function seems to be the reconstruction of the new `payload` value to pass on to `parse_ids_payload`.

Suggestion: `parse_ids_payload` should take a function `string list -> 'a` as argument instead of an assoc list `(string list * 'a) list`. Then, the callsite of `parse_ids_payload` in `parse_property_attribute` has the flexibility to inspect the string list and filter out the arity-related pieces before attempting to match against the keys in the current assoc list.

The idea with this suggestion is that we can drop the code in `parse_arity` that reconstructs the payload, and instead cut it down to the `find_arity_in_args` core.

I recognize that this would mean refactoring `get_ids_from_exp` to also look for constants, or something like that. But this doesn't seem so bad to me.

ncik-roberts (commented):
Minor comments. I see you've requested review from @gretay-js, so I will leave it to her to click "Approve".

ocaml/parsing/builtin_attributes.ml
ccasin: > I recognize that this would mean refactoring get_ids_from_exp to also look for constants, or something like that. But this doesn't seem so bad to me.

I would love to clean this up.  Are you proposing that `get_ids_from_exp` should just ignore constants entirely?  I think then we wouldn't warn if the payload has random other constants in it, or integers that aren't preceded by "arity".  And it seems like making it look explicitly for the "arity followed by int constant" pattern adds a similar level of complexity as we're eliminating in the other place.

ocaml/typing/typedtree.ml
```diff
@@ -1035,13 +1035,20 @@ let rev_let_bound_idents_full bindings =
   List.iter (fun vb -> iter_bound_idents add vb.vb_pat) bindings;
   !idents_full
 
-let let_bound_idents_with_modes_and_sorts bindings =
+let let_bound_idents_with_modes_sorts_and_checks bindings =
   let modes_and_sorts = Ident.Tbl.create 3 in
```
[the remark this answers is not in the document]
ccasin: I renamed the variable.  I hesitate to try to come up with a good name for a record because we already have `let_bound_idents_full` (which exists upstream, so I don't particularly want to rename it) and having `_full` and `_augmented` variants seems confusing.

ocaml/typing/typedecl.ml
```diff
@@ -2669,8 +2671,29 @@ let transl_value_decl env loc valdecl =
...
+      in
+      let zero_alloc =
+        Builtin_attributes.get_property_attribute ~in_signature:true
+          ~default_arity valdecl.pval_attributes Zero_alloc
+      in
+      begin match zero_alloc with
+      | Default_check -> ()
+      | Check za ->
+        if za.arity = 0 then
+          raise (Error(valdecl.pval_loc, Zero_alloc_attr_bad_arity))
```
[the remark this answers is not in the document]
ccasin: I adjusted this so there are two different errors:
* The above error now triggers only when your type has no arrows and the arity isn't overridden
* There's a new error for the case where you explicitly write a bad arity (<= 0)

ccasin:
Per discussion on slack, I've adjusted this to ban `opt` in signatures for now.

If you write `opt` on a value binding in a structure, we simply do not record the zero_alloc attribute in the module type (this is important so that you can't get `opt` in a signature via `module type of`).  As a consequence, I've adjuted hint suggesting to add an attribute if none is present to specify you need a non-opt attribute.

ocaml/typing/typedtree.ml
ncik-roberts: Thanks! Should `let_bound_idents_with_modes_sorts_and_checks` return a single check (rather than a list of checks)? That should simplify the client code, and it's easy enough to validate the assertion that "the list is either all Default_check or a singleton of something else" just by looking at this function.

ocaml/typing/typemod.ml
ncik-roberts: Thanks. I suggested a simplification in the other thread.

ocaml/parsing/builtin_attributes.ml
ncik-roberts: No, I'm suggesting `get_ids_and_constants_from_exp` that returns both ids and constants (both as strings).

> And it seems like making it look explicitly for the "arity followed by int constant" pattern adds a similar level of complexity as we're eliminating in the other place.

I think it's appreciably less as the suggestion allows us to avoid forging a `Pexp_apply` node / payload.

gretay-js (approved):
For the changes in `checkmach.ml` and `zero_alloc_utils.ml`.

ocaml/parsing/builtin_attributes.ml
ncik-roberts: (Btw I’m willing to be told that this isn’t worth it — feel free to resolve this and merge.)

ocaml/typing/typedtree.ml
ccasin: I agree, and have made this change.  I had plans of supporting more complicated patterns at some point, but it's totally unnecessary and anyway we can make this more complicated then if it ever happens.

ocaml/parsing/builtin_attributes.ml
ccasin: This didn't turn out to be quite so easy - there are problems with when we know the location and arity, needed to compute the lookup table (or parsing function) and warning message.  But, anyway, I did a slightly larger refactor somewhat along these lines that leads to much clearer code (and the need to forge this payload).

PR #2471, opened 2024-04-18
Fix function return layout bug

ncik-roberts:
#1817 introduced a bug where `translcore` computed a function's return layout according only to its first case if it was written with the `function` syntax. This PR fixes that bug by computing the return layout the same way it was computed prior to #1817.

For example, the function `get` would get a return layout of `Pvalue Pintval` even though a float is returned out of the second case:

```ocaml
type _ value =
  | Int : int value
  | Float : float value

let get (type a) : a value -> a = function
  | Int -> 0
  | Float -> 0.
```

You can turn this into a flambda2 middle-end crash. (I haven't been able to turn it into a miscompilation, but I theorize this is possible.) This PR adds the crashing test as a regression test.

The same bug exists upstream; I'll post the corresponding PR there after this is reviewed.

### DIFF
--- a/ocaml/lambda/translcore.ml
+++ b/ocaml/lambda/translcore.ml
@@ -1288,20 +1288,8 @@ and transl_apply ~scopes
    [trans_curried_function]).
 *)
 and transl_function_without_attributes
-    ~scopes ~return_sort ~return_mode ~mode ~region loc repr params body =
-  let return_layout =
-    match body with
-    | Tfunction_body exp ->
-        layout_exp return_sort exp
-    | Tfunction_cases { fc_cases = { c_rhs; _ } :: _ } ->
-        layout_exp return_sort c_rhs
-    | Tfunction_cases { fc_cases = [] } ->
-        (* ppxes can generate empty function cases, which compiles to
-           a function that always raises Match_failure. We try less
-           hard to calculate a detailed layout that the middle-end can
-           use for optimizations. *)
-        layout_of_sort loc return_sort
-  in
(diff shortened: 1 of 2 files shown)


ncik-roberts:
As discussed, my original attempt at a fix was clearly and painfully wrong (hence the test failures): I was calling `function_return_layout` on a function type that may stand for a multi-argument function, in which case the returned layout was always `Pgenval`.

I've pushed an updated attempt:
  * I store the environment "just before" the `function` cases in the typedtree. That is, the environment under all of the parameters except for the last one that's scrutizined by the `function` cases.
  * I also store the return type.
  * `translcore` gets the return layout from that env/type.

ncik-roberts:
@goldfirere Could you please provide another round of review given the changes?

ncik-roberts:
@Ekdohibs please review the trivial changes to chamelon.

(This is waiting on Richard's review too but I suspect the changes to chamelon will be affected at most in uninteresting ways by his review.)

Ekdohibs (approved):
For `chamelon/` changes.

goldfirere:
As I've commnuicated to Nick directly, I'm still a little surprised at the shape of this patch and believe that there may be a better way. We had scheduled a meeting to discuss, but I'm now unable to make that meeting. Let's merge this, and Nick and I will (briefly) revisit when I'm back.

(I would merge myself, but perhaps there's been some development I'm not yet read up on, and so I will leave it to Nick to merge.)

ncik-roberts:
I'm merging now. I'll wait to upstream this fix until Richard and I have had a chance to meet. I will open an issue upstream, though, in case this impacts the release cycle at all.

PR #2474, opened 2024-04-19
Move zero alloc annotations into a separate file

liam923:
Merlin now cares about the `-zero-alloc-check` flag, which means it needs the `Clflags.Annotations` module. This is currently located in the `clflags.ml` file. However, this file doesn't get merged into Merlin. This makes me worried that a change to the module could quietly break Merlin. This PR does step 1 of resolving this, which is separating out the module into a separate file. Step 2 (which I'm not sure if it's necessary or will be done automatically) is updating Merlin to merge this new file when running the import script.

### DIFF
--- a/ocaml/utils/zero_alloc_annotations.ml
+++ b/ocaml/utils/zero_alloc_annotations.ml
@@ -1,3 +1,30 @@
+(******************************************************************************
+ *                             flambda-backend                                *
+ *                         Greta Yorsh, Jane Street                           *
+ * -------------------------------------------------------------------------- *
+ *                               MIT License                                  *
+ *                                                                            *
+ * Copyright (c) 2024 Jane Street Group LLC                                   *
+ * opensource-contacts@janestreet.com                                         *
+ *                                                                            *
+ * Permission is hereby granted, free of charge, to any person obtaining a    *
+ * copy of this software and associated documentation files (the "Software"), *
+ * to deal in the Software without restriction, including without limitation  *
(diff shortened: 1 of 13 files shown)


mshinwell:
Could the new files live in a better directory than `utils/`, perhaps?

ccasin:
> Could the new files live in a better directory than `utils/`, perhaps?

What directory would you prefer, for a file depended by clflags (itself in utils)?

ocaml/utils/zero_alloc_annotations.ml
```diff
@@ -0,0 +1,29 @@
+type t = Check_default | Check_all | Check_opt_only | No_check
```
goldfirere, line 1: Maybe they're silly, but I like the file headers in the rest of the compiler. At a minimum, they're useful in identifying Jane Street as the source. Not sure what name should go in, though.

goldfirere (approved):
Approved, pending the discussion above with @ccasin and @mshinwell.

ocaml/utils/zero_alloc_annotations.ml
mshinwell: https://github.com/ocaml-flambda/flambda-backend/blob/main/backend/cfg/cfg_stack_checks.mli is an example of a suitable header (JS-only files are supposed to be MIT-licensed, although this isn't consistently reflected across the tree yet).  You don't need to name a particular author, though it's probably @gretay-js

ccasin (approved):

PR #2476, opened 2024-04-19
Represent mixed records as "faux mixed blocks" in bytecode

ncik-roberts:
The problem this PR tries to solve is that, currently, you can create a mixed record value in bytecode, marshal it over to native code, and get a segfault when you try to read out the fields of the flat suffix. The problem is that mixed records have a different representation in bytecode vs. native code, but there were no guardrails preventing you from marshaling values from bytecode to native code. (Conversely, there were always such guardrails from native code to bytecode).

This PR fixes things by changing the representation of mixed records in bytecode from standard uniform blocks to "faux mixed blocks". The "faux" is merely that the header is tagged as being a mixed block, but all fields are stored boxed.

**"Faux mixed blocks" raise if you try to marshal, hash, or compare them.** That's just like normal mixed blocks.

The implementation of this PR is mostly in macros. I vary the definition of mixed-block-related macros for native code vs bytecode.

I request @stedolan's review so I can feel confident about the macro wrangling.


stedolan (approved):
LGTM (but there's some CI issue)

ncik-roberts:
@mshinwell glanced at the last commit, which fixes the CI issue that Stephen mentioned. So I'm going to go ahead and merge this.

PR #2477, opened 2024-04-19
Refactor otherlibs build

TheNumbat:
This refactors how the alpha/beta/stable otherlibs are built.

- They are no longer included in the upstream build, since they will use extensions that the upstream compiler doesn't support. Hence they do not need a Makefile-based build.
- They are now compiled with their corresponding extension universe enabled.
- The jst makefile now copies all ml/mli files from the otherlibs dirs, not just alpha*, etc.
- The tests for exported modules have been moved from `ocaml/testsuite/` to `testsuite/`, so will only run in flambda-backend builds. The upstream compatible versions (only `modules.ml` and `upstream_compatible.ml`) are still under `ocaml/testsuite`.
- Added tests for alpha/beta/upstream_compatible as well as stable.
- Enabled ocamlformat for these libs


ocaml/Makefile.common-jst
```diff
@@ -256,9 +259,9 @@ install_for_test: _install
          $(cpl) stdlib/dynlink/dynlink* otherlibs/dynlink; \
          $(cpl) stdlib/str/{lib,}str* otherlibs/str; \
          $(cpl) stdlib/upstream_compatible/{lib,}upstream_compatible* otherlibs/upstream_compatible; \
-         $(cpl) stdlib/stable/{lib,}stable* otherlibs/stable; \
-         $(cpl) stdlib/beta/{lib,}beta* otherlibs/beta; \
-         $(cpl) stdlib/alpha/{lib,}alpha* otherlibs/alpha; \
+         $(cpl) stdlib/stable/{lib,}* otherlibs/stable; \
+         $(cpl) stdlib/beta/{lib,}* otherlibs/beta; \
+         $(cpl) stdlib/alpha/{lib,}* otherlibs/alpha; \
```
dkalinichenko-js, line 264:
```suggestion
         $(cpl) stdlib/stable/* otherlibs/stable; \
         $(cpl) stdlib/beta/* otherlibs/beta; \
         $(cpl) stdlib/alpha/* otherlibs/alpha; \
```

ocaml/Makefile.common-jst
```diff
@@ -256,9 +259,9 @@ install_for_test: _install
          $(cpl) stdlib/dynlink/dynlink* otherlibs/dynlink; \
          $(cpl) stdlib/str/{lib,}str* otherlibs/str; \
          $(cpl) stdlib/upstream_compatible/{lib,}upstream_compatible* otherlibs/upstream_compatible; \
```
dkalinichenko-js, line 261:
```suggestion
         $(cpl) stdlib/upstream_compatible/* otherlibs/upstream_compatible; \
```

dkalinichenko-js:
> * They are no longer included in the upstream build, since they will use extensions that the upstream compiler doesn't support. Hence they do not need a Makefile-based build.
> * The tests for exported modules have been moved from `ocaml/testsuite/` to `testsuite/`, so will only run in flambda-backend builds. The upstream compatible versions (only `modules.ml` and `upstream_compatible.ml`) are still under `ocaml/testsuite`.

As far as I know, the makefile-based build inside this repository is used not by the upstream compiler, but by our compiler. `ocaml/testsuite/` also contains numerous tests that only work inside our compiler, such as layouts tests. So I'm confused why would we move the otherlibs tests to `testsuite/`.

TheNumbat:
Some time ago we switched flambda-backend builds to use dune whenever available - I confirmed the lib-extensions tests still run under `make test` here, so it seems fine.

The layouts tests actually work in the upstream compiler build because they only require frontend features present in `ocaml/`. As of this PR it's not yet necessary to move the otherlibs tests, but it will be when I add the `Float32` module, because it requires backend features that are not present in `ocaml/`.

dkalinichenko-js:
I think you should move `upstream_compatible` to be flambda-only then too. `upstream_compatible` does not actually mean "builds as is with the upstream compiler", instead it's "something we provide an upstream-compatible substitute shim for, even if it degrades performance". So, for example, `Float_u` can be flambda-only, yet still be in `upstream_compatible` if we have a module with the same interface (modulo extension erasure stuff) and an upstream-compatible implementation, even if that implementation just uses normal `float`.

TheNumbat:
Sounds good, upstream_compatible is now treated the same as the others

ocaml/testsuite/tests/lib-extensions/modules.ml
dkalinichenko-js, line 1: Should we delete his file?

dkalinichenko-js (approved):

ocaml/testsuite/tests/lib-extensions/modules.ml
TheNumbat: I just left it there to assure that make sets up the folder structure properly

dkalinichenko-js:
Should we merge this?

PR #2478, opened 2024-04-22
Clean up mode error printing

riaqn:
This PR cleans up the following things:
- Have a generic `print_axis` that works on any axis, so that `typecore.ml` does't need to mention specific axes when possible.
- Add some missing periods.
- Rephrase the error messages so articles "a" "an" is not followed by mode constants.
- Break long lines.

### DIFF
--- a/ocaml/typing/typecore.ml
+++ b/ocaml/typing/typecore.ml
@@ -9462,17 +9462,19 @@ let escaping_hint (failure_reason : Value.error) submode_reason
          global, then exclave_ won't solve the problem. *)
       [ Location.msg
           "@[Hint: Cannot return a local value without an@ \
-           \"exclave_\" annotation@]" ]
+           \"exclave_\" annotation.@]" ]
     | _, Return -> []
     | _, Tailcall_argument ->
       [ Location.msg
-          "@[Hint: This argument cannot be local, because it is an argument in a tail call@]" ]
+          "@[Hint: This argument cannot be local,@ \
+           because it is an argument in a tail call.@]" ]
     | _, Tailcall_function ->
       [ Location.msg
-          "@[Hint: This function cannot be local, because it is the function in a tail call@]" ]
+          "@[Hint: This function cannot be local,@ \
+           because it is the function in a tail call.@]" ]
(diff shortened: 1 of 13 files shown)


ocaml/typing/typecore.ml
```diff
@@ -10155,35 +10154,29 @@ let report_error ~loc env = function
...
-      end
+  | Param_mode_mismatch (s, Error (ax, {left; right})) ->
+      let actual, expected =
+        match s with
+        | Left_le_right -> left, right
+        | Right_le_left -> right, left
+      in
+      Location.errorf ~loc
+        "@[This function takes a parameter at %a,@ \
+        but was expected to take a parameter at %a.@]"
```
stedolan, line 10165: I think `at %a` isn't great, because of this error message (from one of the test files):
```
Error: This function when partially applied returns a value at once,
       but expected to be at many.
```
"returns a value at once" means "returns a value very soon", not "returns a value whose mode is `once`".

How about `which is %a` instead of `at %a`? (Other messages use modes with "to be", and I think it reads fine)

ocaml/typing/typecore.ml
```diff
@@ -10155,35 +10154,29 @@ let report_error ~loc env = function
...
+          Location.errorf ~loc
+            "This function or one of its parameters escape their region@ \
+            when it is partially applied."
       | Error (Monadic Uniqueness, _) -> assert false
       | Error (Comonadic Linearity, {left; right}) ->
-          Location.errorf ~loc "This function when partially applied returns a %a value,@ \
-          but expected to be %a." Linearity.Const.print left Linearity.Const.print right
+          Location.errorf ~loc
+            "This function when partially applied returns a value at %a,@ \
+              but expected to be at %a."
```
stedolan, line 10177: (same comment about `at %a`)

ocaml/typing/typecore.ml
riaqn: Yeah `at once` has the advantage of close to the actual syntax `@ once`, but yeah it could be confusing for some mode constant names.

stedolan (approved):

PR #2479, opened 2024-04-22
Add partial adjoints of `join_with` and `meet_with`

riaqn:
This PR adds the partial left adjoint of `meet_with`, which we call `Restrict_below`, and the partial right adjoint of `join_with`, which we call `Restrict_above`.

- Recall that our mode solver supports certain form of partial adjunction (See solver_intf.mli), which this PR satisfies.
- This simplifies the definition of modalities and (in the future) mode crossing.

To justify the morphism composition reduction added:
- `Restrict_below _` composed with any `m` is just `m`. If the rest of the type checker is sound, `Restrict_below` will never be applied to values out of its defined range, and will act just as identity.
- `Imply c0 (Meet_with c1 x)` reduces to `Imply c0 x`, if `c0 <= c1`. Proof:
By adjunction, `Imply c0 (Meet_with c1 x)` is characterised as the biggest `y` such that `Meet_with c0 y <= Meet_with c1 x` holds. This reduces to `Meet_with c0 y <= c1` and `Meet_with c0 y <= x`. The former is trivial given `c0 <= c1`. Therefore only the latter matters, which is equivalent to `y <= Imply c0 x`. That is, we want the biggest `y` such that `y <= Imply c0 x` holds, which gives `Imply c0 x`.

The other reduction is dual.

### DIFF
--- a/ocaml/typing/mode.ml
+++ b/ocaml/typing/mode.ml
@@ -504,14 +504,20 @@ module Lattices_mono = struct
 
   type ('a, 'b, 'd) morph =
     | Id : ('a, 'a, 'd) morph  (** identity morphism *)
-    | Meet_with : 'a -> ('a, 'a, 'd * disallowed) morph
+    | Meet_with : 'a -> ('a, 'a, 'l * 'r) morph
         (** Meet the input with the parameter *)
+    | Restrict_below : 'a -> ('a, 'a, 'l * disallowed) morph
+        (** [Restrict_below c] is identity for input lower than [c], and
+        undefined otherwise. This is the partial left joint of [Meet_with]. *)
     | Imply : 'a -> ('a, 'a, disallowed * 'd) morph
         (** The right adjoint of [Meet_with] *)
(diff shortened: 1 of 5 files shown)


riaqn:
This PR has now switched to simply using `Id` for both partial adjoints. The reasoning is as follows:
Inside the solver, the result of `left_adjoint f` is only called on `a` such that `a <= f b`. If `f` is `Meet_with c`, and `a <= Meet_with c b`, we can know that `a <= b`. I.e., `f` can be reduced to `Id`, and the left adjoint of `Id` is just `Id`.

riaqn:
The last commit restricts the kind of "partial adjunction" the solver can handle.
- Previously, we allow any convex sublattices, which can be supported but requires some extra steps (which the implementation was missing and thus not meeting the specification).
- Now, we only support sublattices in either direction: either downward closed or upward closed. This should be sufficient for all partial adjunctions we have in mind.

lpw25 (approved):
LGTM

PR #2480, opened 2024-04-23
`Non_null_value` layout

dkalinichenko-js:
Implement the `non_null_value` layout, which is a subtype of `value`. There aren't yet any types with that layout. In the next PR, I will change most types to have the `non_null_value` layout when `layouts_alpha` is enabled.

Add tests for `non_null_vaue`. I did not add `non_null_value` to every test group in `basic_alpha.ml` and `modules_alpha.ml`, but it should be enough to test parsing and subtyping.


ocaml/typing/jkind.mli
```diff
@@ -177,12 +177,14 @@ end
 type sort = Sort.t
 
 (* The layout of a type describes its memory layout. A layout is either the
-   indeterminate [Any] or a sort, which is a concrete memory layout. *)
+   indeterminate [Any], a sort, which is a concrete memory layout, or
+   [Non_null_value], which is the same as [Value] but also can't be null. *)
```
ccasin, line 181: I think I would say "...which is a sublayout of the sort [Value] describing types that do not allow the concrete value null".  And here or somewhere else it should be noted that this is the layout of classic ocaml values.

ccasin (approved):
This looks good.

I'd like to see more explicit testing of the sublayout relationship (forgive me if I missed it).  I understand this is implicitly tested ~everywhere, but including some tests like the ones we discussed on slack, either in basics or in a new file, would serve as useful explicit documentation of the expected relationship.

ocaml/typing/jkind.mli
dkalinichenko-js: Done.

ccasin:
New tests look good - thanks!

PR #2481, opened 2024-04-23
Use correct debuginfo values for Ialloc

mshinwell:
The store instructions emitted by `Selectgen` for the `Ialloc` code sequence seem to have the wrong debuginfo values at present.  I think they should have the debuginfo of the allocation, not that of the initializer for the relevant field, which might be in a completely different source location.


mshinwell:
@xclerc could you review please?

xclerc (approved):

PR #2482, opened 2024-04-23
Support getting paths from -libloc flags

Forestryks:


ocaml/utils/load_path.ml
```diff
@@ -175,6 +203,10 @@ let init ~auto_include ~visible ~hidden =
   reset ();
   visible_dirs := List.rev_map (Dir.create ~hidden:false) visible;
   hidden_dirs := List.rev_map (Dir.create ~hidden:true) hidden;
+  List.iter (fun (libloc : Clflags.Libloc.t) ->
```
Forestryks, line 206: I'm not sure if that's the place where we want to get libloc from command line arguments. It looks like a better approach would be to pass it as argument to `init`, but it's not clear to me in which of `Load_path.init` calls we should include `libloc`

ocaml/utils/load_path.ml
```diff
@@ -18,35 +18,39 @@ module Dir : sig
   type t
 
   val path : t -> string
-  val files : t -> string list
+  val files : t -> (string * string) list
```
poechsel, line 21: The meaning of `(string * string)` is not obvious here. Maybe define a small type for it?

ocaml/utils/load_path.ml
poechsel: I'm fine with doing it here

ocaml/driver/main_args.ml
```diff
@@ -165,6 +165,10 @@ let mk_H f =
   "<dir>  Add <dir> to the list of \"hidden\" include directories\n\
  \     (Like -I, but the program can not directly reference these dependencies)"
 
+let mk_libloc f =
+  "-libloc", Arg.String f, "<dir>:<libs>:<hidden_libs>  Location of .libloc dir as well \
```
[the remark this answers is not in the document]
poechsel: After speaking about it offline I'm fine with having one single argument but think the documentation could be improved to explain that a bit more about the what libloc is doing, that is:
- It's a way to control how the compiler is looking up libraries
- dir is expected to contain one file per library named `<library-name>/cmi-cmx` with one path per line
- libs and hidden_libs is a comma-separated list of library names

poechsel (approved):

PR #2485, opened 2024-04-23
Asm_directives for arm64; move frame_size etc into Proc for amd64+arm64

mshinwell:
Implements `Asm_directives` for arm64 targets and also a few other things required for DWARF inlined frames (but not specific to debugging support).  I will discuss this with @xclerc in due course.

Includes #2483

### DIFF
--- a/backend/arm64/emit.mlp
+++ b/backend/arm64/emit.mlp
@@ -100,32 +100,24 @@ let prologue_required = ref false
 let contains_calls = ref false
 
 let initial_stack_offset () =
-  Stack_check.initial_stack_offset ~contains_calls:!contains_calls
+  Proc.initial_stack_offset ~contains_calls:!contains_calls
     ~num_stack_slots
 
 let frame_size () =
-  Stack_check.frame_size
+  Proc.frame_size
     ~stack_offset:!stack_offset
-    ~frame_required:(frame_required
-      ~fun_contains_calls:!contains_calls
-      ~fun_num_stack_slots:num_stack_slots)
     ~contains_calls:!contains_calls
     ~num_stack_slots
 
-let slot_offset loc cl =
-  match loc with
-    Incoming n ->
-      assert (n >= 0);
-      frame_size() + n
-  | Local n ->
-      !stack_offset +
-      (if cl = 0
-       then n * 8
-       else num_stack_slots.(0) * 8 + n * 8)
-  | Outgoing n ->
-      assert (n >= 0);
-      n
-  | Domainstate _ -> assert false (* not a satck slot *)
+let slot_offset loc stack_class =
+  let offset =
+    Proc.slot_offset loc ~stack_class ~stack_offset:!stack_offset
+      ~fun_contains_calls:!contains_calls ~fun_num_stack_slots:num_stack_slots
+  in
+  match offset with
+  | Bytes_relative_to_stack_pointer n -> n
+  | Bytes_relative_to_domainstate_pointer _ ->
+    Misc.fatal_errorf "Not a stack slot"
 
 (* Output a stack reference *)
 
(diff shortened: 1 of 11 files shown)


backend/arm64/emit.mlp
```diff
@@ -1268,6 +1268,125 @@ let data l =
...
+      let loc ~file_num ~line ~col ?discriminator () =
+        ignore discriminator;
+        emit_line (Printf.sprintf ".loc %d %d %d" file_num line col)
+
+      let comment str =
+        emit_line (Printf.sprintf "; %s" str)
+
+      let label ?data_type str =
+        let _ = data_type in
+        emit_line (Printf.sprintf "%s:" str)
```
xclerc, line 1326: Doesn't that need a prefix?

backend/arm64/emit.mlp
```diff
@@ -1268,6 +1268,125 @@ let data l =
...
+      let long const =
+        emit_line
+          (Printf.sprintf "\t.long %s" (string_of_constant const))
+
+      let qword const =
+        emit_line
+          (Printf.sprintf "\t.quad %s" (string_of_constant const))
+
+      let bytes str =
+        emit_line (Printf.sprintf "\t.ascii %S" str)
```
xclerc, line 1373: Is there a risk the `str` could be
really large? (I vaguely remember
an assembler whose behaviour
was quite bad -quadratic?)

backend/arm64/emit.mlp
mshinwell: Don't think so, that gets provided by the caller.

backend/arm64/emit.mlp
mshinwell: Not sure if there is a problem here, but this matches what the emitter code does further down the file, so I think it's ok.

backend/arm64/proc.ml
```diff
@@ -405,17 +405,46 @@ let frame_required ~fun_contains_calls ~fun_num_stack_slots =
...
+let compute_initial_stack_offset ~fun_contains_calls ~fun_num_stack_slots =
+    8 * fun_num_stack_slots.(0) +
+    8 * fun_num_stack_slots.(1) +
+    (if fun_contains_calls then 8 else 0)
+
+let frame_size ~stack_offset ~fun_contains_calls ~fun_num_stack_slots =
+  let sz =
+    stack_offset +
+    compute_initial_stack_offset ~fun_contains_calls ~fun_num_stack_slots
+  in Misc.align sz 16
```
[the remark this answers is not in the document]
mshinwell: This should now be fixed.  For both amd64 and arm64 this code is now in `Proc`.  One thing to check: it looks like there was an unused `initial_stack_offset : int` in `proc.ml` for both of these architectures -- @xclerc could you confirm this was the case?  I've replaced it with the above function now.

backend/arm64/proc.ml
```diff
@@ -405,17 +405,46 @@ let frame_required ~fun_contains_calls ~fun_num_stack_slots =
...
+        frame_size ~stack_offset ~fun_contains_calls
+          ~fun_num_stack_slots
+      in
+      Bytes_relative_to_stack_pointer (frame_size + n)
+  | Local n ->
+      let offset =
+        stack_offset +
+        (if stack_class = 0
+        then n * 8
+        else fun_num_stack_slots.(0) * 8 + n * 8)
```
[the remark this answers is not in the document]
mshinwell: Fixed.

backend/arm64/emit.mlp
xclerc: (I vaguely remember we had to split
long string to reduce the time it takes
for the assembly to produce a `o` file.)

backend/arm64/proc.ml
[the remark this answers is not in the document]
xclerc: (Discussed off-line.)

PR #2487, opened 2024-04-24
Zero alloc: fix join of Transform with the same variables

gretay-js:
On top of #2428. Only the last commit is new.

This is a bug fix, as @xclerc noted in review  of #2428: Join of two `Transform.t` with the same variables but different witnesses should result in one Transform with joined witnesses. The fix is to change the representation of `Join`: instead of  a set of `Transform.t` , use a map from a set of variables to `Transform.t`.  

Current implementation is naive because it effectively keeps two sets of variables for each `Transform.t`: one as a key in the map and the other as part of the representation of `Transform.t`.

### DIFF
--- a/backend/checkmach.ml
+++ b/backend/checkmach.ml
@@ -416,7 +416,7 @@ end = struct
     let replace_witnesses t w =
       match t with
       | Args vars -> Args (Vars.replace_witnesses vars w)
-      | Args_with_top { w; vars } ->
+      | Args_with_top { w = _; vars } ->
         Args_with_top { w; vars = Vars.replace_witnesses vars w }
 
     let print ~witnesses ppf t =
--- a/backend/checkmach.ml
+++ b/backend/checkmach.ml
@@ -296,7 +296,10 @@ end = struct
 
     let same_vars = Var.Map.equal (fun _ _ -> (* ignore witnesses *) true)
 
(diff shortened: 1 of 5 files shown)


backend/checkmach.ml
```diff
@@ -439,6 +440,73 @@ end = struct
...
+
+    let empty = M.empty
+
+    let get_key tr = tr |> Transform.get_vars |> Vars.get_vars
+
+    let add tr t =
+      let res = M.add (get_key tr) tr t in
+      maybe_widen res
+
+    let compare = M.compare Transform.compare
```
xclerc, line 493: (Maybe eta-expand to avoid the closure)

gretay-js: done. thanks for repeating this comment, i missed this instance previously as it was on another branch.

backend/checkmach.ml
```diff
@@ -290,6 +296,8 @@ end = struct
 
     let same_vars = Var.Map.equal (fun _ _ -> (* ignore witnesses *) true)
 
+    let get_vars t = t |> Var.Map.to_seq |> Seq.map fst |> Var.Set.of_seq
```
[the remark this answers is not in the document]
gretay-js: I added a CR in the code.

xclerc (approved):

gretay-js:

PR #1472, opened 2023-06-09
Zero alloc bounded join of witnesses

gretay-js:
On top of https://github.com/ocaml-flambda/flambda-backend/pull/1471. Only last two commits are new. Best review commit by commit. 

The analysis keeps track of all witness instructions that lead to `Top` at function entry, even if the flag `-checkmach-details-cutoff n` limits the number of witnesses to be included in error message to some `n > 0`. This PR uses `n` as the apper bound on the size of the set of witnesses to be propagated. 

There are two commits. The first is a simple change to  `join` to implement the above functionality. 
The second commit is refactoring of `Witnesses.t` type to keep track of whether join was in fact bounded or not, and propagate this information to function entry. This information is then used to print "..."  at the end of error messages so the user knows when list of witnesses is partial. I am not entirely sure that the "dots" are worth the extra memory overhead of boolean field to track them. I'd be fine without it but wanted to get feedback first, or maybe there is a cheaper way to track it. I thought about tracking it per function but it breaks the abstractions.

PR #2489, opened 2024-04-24
Zero alloc: bounded join of witnesses

gretay-js:
On top of #2487.  Last two commits are new.

Heuristics to limit the number of witnesses tracked by the analysis.  This PR does not change precision of the analysis.

The first commit is the same as the first commit of https://github.com/ocaml-flambda/flambda-backend/pull/1472:  keep at most `n` witnesses where `n` is based on `-checkmach-details-cutoff n` flag with default 20.

The second commit limits the number of elements of a symbolic representation of  `Join` that contains `Top`. It can result in more or less witnesses depending on how the variables are resolved, but it is a reasonable limit based on `n` above and easy to compute.

Note that witnesses are not tracked by default for functions that are not annotated with `[@zero_alloc ..]` assertions (unless `-checkmach-details-cutoff` is set to track all witnesses, which is used for displaying allocations in an editor).  The motivation of this PR is to ensure that adding an "assert" on a function does not increase compilation overhead unnecessarily.

### DIFF
--- a/backend/checkmach.ml
+++ b/backend/checkmach.ml
@@ -416,7 +416,7 @@ end = struct
     let replace_witnesses t w =
       match t with
       | Args vars -> Args (Vars.replace_witnesses vars w)
-      | Args_with_top { w; vars } ->
+      | Args_with_top { w = _; vars } ->
         Args_with_top { w; vars = Vars.replace_witnesses vars w }
 
     let print ~witnesses ppf t =
--- a/backend/checkmach.ml
+++ b/backend/checkmach.ml
@@ -296,7 +296,10 @@ end = struct
 
     let same_vars = Var.Map.equal (fun _ _ -> (* ignore witnesses *) true)
 
-    let get_vars t = t |> Var.Map.to_seq |> Seq.map fst |> Var.Set.of_seq
+    let get_vars t =
+      (* CR-someday gyorsh: If this is called often, it may be better to store
+         the result of [get_vars t] in [t] instead of recomputing it. *)
+      t |> Var.Map.to_seq |> Seq.map fst |> Var.Set.of_seq
 
     let has_witnesses vars =
       Var.Map.exists (fun _ w -> not (Witnesses.is_empty w)) vars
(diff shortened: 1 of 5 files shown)


backend/checkmach.ml
```diff
@@ -110,10 +125,19 @@ module Witnesses : sig
...
-  (* CR gyorsh: consider using Flambda_backend_flags.checkmach_details_cutoff to
-     limit the size of this set. The downside is that it won't get tested as
-     much. Only keep witnesses for functions that need checking. *)
-  let join = union
+  let print ppf t = Format.pp_print_seq Witness.print ppf (to_seq t)
+
+  let cutoff t ~n = take_first_n t n ~fold ~remove ~cardinal
+
+  let join t1 t2 =
+    let res = union t1 t2 in
```
xclerc, line 139: That could be changed in a follow-up, but
it feels a bit weird to first compute the
union unconstrained to then remove a
number of values if the union happens to
be too large. Would repeated additions be
bad in terms of complexity?

xclerc: (If it is important to select the first `n` values
according to the order, we could iterate
over a couple of sequences until we reach
the threshold.)

gretay-js: Added a CR.

backend/checkmach.ml
```diff
@@ -77,6 +77,21 @@ module Witness = struct
     Format.fprintf ppf "%a {%a}@," print_kind kind Debuginfo.print_compact dbg
 end
 
+(* CR gyorsh: avoid duplicating somehwat tricky logic. is there a nicer way
```
xclerc, line 80: Maybe the sequence is the good abstraction,
there is `Seq.take` to the function would only
need to take `to_seq` and `of_seq` to provide
the same functionality.

xclerc (approved):

backend/checkmach.ml
gretay-js: good idea! we still need `cardinal` to avoid reallocating on inputs shorter than `n` (I'm not sure Seq is smart enough to avoid it).

xclerc (approved):

gretay-js:

PR #2490, opened 2024-04-24
Zero alloc: widening

gretay-js:
On top of #2489. Only the last 2 commits are new.

Symbolic summaries in "dnf-like" normal form can sometimes explode. To avoid it, this PR abstract join of more than `n` elements to Top. The first commit adds a temporary compilation flag `-checkmach-join` to control `n`. The second commit actual implements this widening and is currently a bit of a hack to bail out early in the middle of `distribute_transform_over_join` and other operations that increase the number of elements in a join.

### DIFF
--- a/driver/flambda_backend_flags.ml
+++ b/driver/flambda_backend_flags.ml
@@ -46,6 +46,14 @@ type checkmach_details_cutoff =
 let default_checkmach_details_cutoff = At_most 20
 let checkmach_details_cutoff = ref default_checkmach_details_cutoff
                                        (* -checkmach-details-cutoff n *)
+type checkmach_join =
+  | Keep_all
+  | Widen of int  (* n > 0 *)
+  | Error of int (* n > 0 *)
+
+let default_checkmach_join = Widen 100
+let checkmach_join = ref default_checkmach_join
+                              (* -checkmach-precise-join-threshold n *)
 module Function_layout = struct
   type t =
     | Topological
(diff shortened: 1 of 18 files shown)


driver/flambda_backend_flags.ml
```diff
@@ -44,6 +44,14 @@ type checkmach_details_cutoff =
 let default_checkmach_details_cutoff = At_most 20
 let checkmach_details_cutoff = ref default_checkmach_details_cutoff
                                        (* -checkmach-details-cutoff n *)
+type checkmach_join =
+  | Keep_all
+  | Widen of int  (* n > 0 *)
+  | Error of int (* n > 0 *)
```
xclerc, line 50: `n < 0`? (See https://github.com/ocaml-flambda/flambda-backend/pull/2490/commits/c941ead98a9287947a9665d0fecaefa8484eec4c#diff-39b6a9debf276ee2e6de92522cd7ae1c715822b544c6c38f6a890841eaee49ddR952)

driver/flambda_backend_flags.mli
```diff
@@ -47,6 +47,14 @@ type checkmach_details_cutoff =
 val checkmach_details_cutoff : checkmach_details_cutoff ref
 val default_checkmach_details_cutoff : checkmach_details_cutoff
 
+type checkmach_join =
+  | Keep_all
+  | Widen of int  (* n > 0 *)
+  | Error of int (* n > 0 *)
```
xclerc, line 53: (Same question as in `.ml`.)

xclerc:
(If it is not too cumbersome, a couple
of tests using the various settings would
be useful.)

driver/flambda_backend_flags.ml
gretay-js: Fixed. Parsing of args was wrong in one place, the intent is for the argument to be positive (that's how it is used in `Checkmach`).

xclerc (approved):

gretay-js:

PR #2491, opened 2024-04-24
Zero alloc: resolve dependencies earlier

gretay-js:
On top of #2490. 

This PR resolves some dependencies earlier instead of waiting until the end of the compilation unit. 
It resolves dependencies caused by "forward referenced" functions and "self" recursive functions that do not have any other unresolved dependencies. It doesn't resolve mutually recursive functions early. 

Main changes:
- generalize `Value.apply` to be "partial" : when some but all dependencies are resolved, we can use it to simplify a symbolic summary, even if it does not become resolved as a result.
- a module to track unresolved dependencies: this is a bit easier than in the previous versions of the analysis, because we can just inspect a symbolic summary to find what it depends on.
- refactor `fixpoint` function slightly so we can use it for both the final fixpoint and for resolving "self" recursive functions.

### DIFF
--- a/backend/checkmach.ml
+++ b/backend/checkmach.ml
@@ -416,7 +416,7 @@ end = struct
     let replace_witnesses t w =
       match t with
       | Args vars -> Args (Vars.replace_witnesses vars w)
-      | Args_with_top { w; vars } ->
+      | Args_with_top { w = _; vars } ->
         Args_with_top { w; vars = Vars.replace_witnesses vars w }
 
     let print ~witnesses ppf t =
--- a/backend/checkmach.ml
+++ b/backend/checkmach.ml
@@ -523,13 +523,16 @@ end = struct
     exception Widen
 
     let maybe_widen t =
-      (* CR-soon gyorsh: Implement widening to Top instead of error. *)
-      if M.cardinal t > 1000
-      then
-        Misc.fatal_errorf
-          "Join with %d paths is too big, use -disable-precise-checkmach"
-          (M.cardinal t)
-      else t
+      match !Flambda_backend_flags.checkmach_join with
+      | Keep_all -> t
+      | Widen n -> if M.cardinal t > n then raise Widen else t
+      | Error n ->
+        if M.cardinal t > n
+        then
+          Misc.fatal_errorf
+            "Join with %d paths is too big, use -disable-precise-checkmach"
+            (M.cardinal t)
+        else t
 
     let empty = M.empty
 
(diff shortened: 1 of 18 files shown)


backend/checkmach.ml
```diff
@@ -1250,38 +1250,40 @@ end = struct
...
+  let apply t ~env =
+    let env var w =
+      match env var with
+      | None -> unresolved w var
+      | Some v -> replace_witnesses w v
+    in
+    let rec aux t =
+      match t with
+      | Bot | Safe | Top _ -> t
+      | Var { var; witnesses } -> env var witnesses
```
xclerc, line 1262: (If unresolved, will reallocate while we could return `t`.)

backend/checkmach.ml
```diff
@@ -2039,6 +2039,10 @@ end = struct
     val map : t -> f:(Func_info.t -> Value.t -> Value.t) -> t
 
     val print : msg:string -> Format.formatter -> t -> unit
+
+    (** initialize [env] with Bot for all functions on normal and exceptional
+       return, and Safe for diverage component conservatively. *)
```
xclerc, line 2177: ("diverage" -> "diverge")

backend/checkmach.ml
gretay-js: Slightly improved `aux` to not reallocate if there is no change. Left a CR for further optimizations mentioned in your other comment.

backend/checkmach.ml
```diff
@@ -2254,18 +2246,93 @@ end = struct
...
+     not resolve mutually recursive loops. *)
+  let rec propagate ~callee unit_info unresolved_deps ppf =
+    let callee_info = Unit_info.find_exn unit_info callee in
+    if Value.is_resolved callee_info.value
+       && Unresolved_dependencies.contains ~callee unresolved_deps
+    then (
+      let callers =
+        Unresolved_dependencies.get_callers ~callee unresolved_deps
+      in
+      assert (not (String.Set.is_empty callers));
```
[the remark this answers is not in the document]
gretay-js: That was my worry to and that's why the assert is here, this condition is not trivial. I think it holds because we keep `unresolved_deps` in sync with `unit_info` in line 2321.

driver/flambda_backend_args.ml
```diff
@@ -126,6 +126,17 @@ let mk_checkmach_details_cutoff f =
      | No_details -> 0
      | At_most n -> n)
 
+
+let mk_checkmach_join f =
+  "-checkmach-join", Arg.Int f,
+  Printf.sprintf " How many abstract paths before losing precision \
+                  (default %d, negative to fail instead of widenning, \
```
xclerc, line 133:
```suggestion
                  (default %d, negative to fail instead of widening, \
```

backend/checkmach.ml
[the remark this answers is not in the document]
xclerc: (Discussed off-line.)

xclerc (approved):

gretay-js:

PR #2497, opened 2024-04-25
`[%call_pos]` application via an optional argument results in an error

Enoumy:
This feature explicitly disallows `[%call_pos]` arguments being applied via an optional label. For example:

```ocaml
let f ~(here : [%call_pos]) () = here

let location = f ?here:None ()
let location = f ?here:(Some (f ())) ()
(* the above segments results in an warning and later a type error stating that `here` is not optional. *)
```

This feature explicitly returns an error for this situation. (instead of seeing a confusing warning, and then eventually seeing a type error).

Please let me know if the approach of me adding new error variants for classtype.ml and typecore.ml seem reasonable + also please let me know if the error message is reasonable/clear for users.

## Original feature description (for context - other readers):

This feature fixes a bug where `[%call_pos]` arguments are not applied in a semi-common situation:

```ocaml
let f ~(here : [%call_pos]) () = here

let location = f ?here:None ()
let location = f ?here:(Some (f ())) ()
(* the above segments results in an error stating that `here` is not optional. *)
```
This feature fixes the error in the above situation. The fix involves expading the above scenarios into something like:

```ocaml
let location = f ?here:EXPR ()

(* becomes: *)
let location = f ~here:(match EXPR with None -> (* location of application*) | Some x -> x) 
```
This sadly does add a branch at runtime anytime an argument is applied like this. I am unsure if there is a better way of doing this. Please let me know if you have any concerns with this/alternate approaches. Thanks!

### DIFF
--- a/ocaml/testsuite/tests/typing-implicit-source-positions/application_via_option.ml
+++ b/ocaml/testsuite/tests/typing-implicit-source-positions/application_via_option.ml
@@ -1,0 +1,118 @@
+(* TEST_BELOW
+Fille
+*)
+
+let f = fun ~(call_pos:[%call_pos]) () -> call_pos
+[%%expect {|
+val f : call_pos:[%call_pos] -> unit -> lexing_position = <fun>
+|}]
+
+let _ = f ?call_pos:None ();
+[%%expect {|
+Line 1, characters 20-24:
+1 | let _ = f ?call_pos:None ();
+                        ^^^^
+Warning 43 [nonoptional-label]: the label call_pos is not optional.
+
+Line 1, characters 20-24:
+1 | let _ = f ?call_pos:None ();
+                        ^^^^
+Error: This expression should not be a constructor, the expected type is
+       lexing_position
+|}]
+
+let _ =
+  let pos = f () in
+  f ?call_pos:(Some pos) ();
+[%%expect {|
+Line 3, characters 14-24:
+3 |   f ?call_pos:(Some pos) ();
+                  ^^^^^^^^^^
+Warning 43 [nonoptional-label]: the label call_pos is not optional.
+
+Line 3, characters 14-24:
+3 |   f ?call_pos:(Some pos) ();
+                  ^^^^^^^^^^
+Error: This expression should not be a constructor, the expected type is
+       lexing_position
+|}]
+
+let ( >>| ) ~(call_pos : [%call_pos]) a b = a + b, call_pos ;;
+[%%expect {|
+val ( >>| ) : call_pos:[%call_pos] -> int -> int -> int * lexing_position =
+  <fun>
+|}]
+
+let _ =  ( >>| ) ?call_pos:None 1 2 ;;
+[%%expect {|
+Line 1, characters 27-31:
+1 | let _ =  ( >>| ) ?call_pos:None 1 2 ;;
+                               ^^^^
+Warning 43 [nonoptional-label]: the label call_pos is not optional.
+
+Line 1, characters 27-31:
+1 | let _ =  ( >>| ) ?call_pos:None 1 2 ;;
+                               ^^^^
+Error: This expression should not be a constructor, the expected type is
+       lexing_position
+|}]
+
+let _ =
+  let pos = f () in
+  ( >>| ) ?call_pos:(Some pos) 1 2
+;;
+[%%expect {|
+Line 3, characters 20-30:
+3 |   ( >>| ) ?call_pos:(Some pos) 1 2
+                        ^^^^^^^^^^
+Warning 43 [nonoptional-label]: the label call_pos is not optional.
+
+Line 3, characters 20-30:
+3 |   ( >>| ) ?call_pos:(Some pos) 1 2
+                        ^^^^^^^^^^
+Error: This expression should not be a constructor, the expected type is
+       lexing_position
+|}]
+
+class c ~(call_pos : [%call_pos]) () = object 
+  method call_pos = call_pos
+end
+[%%expect {|
+class c :
+  call_pos:[%call_pos] ->
+  unit -> object method call_pos : lexing_position end
+|}]
+
+let _ = new c ?call_pos:None ();;
+[%%expect {|
+Line 1, characters 24-28:
+1 | let _ = new c ?call_pos:None ();;
+                            ^^^^
+Warning 43 [nonoptional-label]: the label call_pos is not optional.
+
+Line 1, characters 24-28:
+1 | let _ = new c ?call_pos:None ();;
+                            ^^^^
+Error: This expression should not be a constructor, the expected type is
+       lexing_position
+|}]
+
+let _ = 
+  let pos = f () in
+  new c ?call_pos:(Some pos) ();;
+[%%expect {|
+Line 3, characters 18-28:
+3 |   new c ?call_pos:(Some pos) ();;
+                      ^^^^^^^^^^
+Warning 43 [nonoptional-label]: the label call_pos is not optional.
+
+Line 3, characters 18-28:
+3 |   new c ?call_pos:(Some pos) ();;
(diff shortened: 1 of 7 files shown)


lpw25:
These arguments aren’t options so I don’t see why they should support taking options as arguments. Do you have an example of why you want this?

Enoumy:
@lpw25 Good question! I think this would be useful for situations where a function that does not use `[%call_pos]` wants to call a `[%call_pos]` function optionally. For example:

```
let with_call_pos : here:[%call_pos] -> unit -> 'a = fun ~(here : [%call_pos]) () -> ...

let without_call_pos ?here () = with_call_pos ?here ()
```

currently it `without_call_pos` would need to either match on `here` and then call `with_call_pos` with and without `~here`, or `without_call_pos` is forced to also use `[%call_pos]`.

ocaml/typing/btype.ml
```diff
@@ -597,6 +597,25 @@ let label_name = function
   | Optional s
   | Position s -> s
 
+let optionally_apply_call_pos ~(arg : Parsetree.expression) ~application_loc =
+  let loc = { arg.pexp_loc with loc_ghost = true } in
+  let none_case =
+    Ast_helper.Exp.case
+      (Ast_helper.Pat.construct ~loc { loc; txt = Lident "None" } None)
+      (let loc = application_loc in
+       Ast_helper.Exp.extension ~loc ({ loc; txt = "src_pos" }, PStr []))
```
ncik-roberts, line 606: It's a bit strange that we're using `src_pos` here. IIUC that's different than the `call_pos` we'd get from omitting the argument entirely.

ncik-roberts:
A concrete reason why we want this PR is that it makes it easier to migrate functions with `?here:Source_code_position.t` arguments to instead take `?here:[%call_pos]` without updating any callsites.

An alternative to this PR would be to disallow `?call_pos` arguments at call sites. But I prefer the shape of approach in this PR.

lpw25:
I think that there is no new behaviour that matches the old behaviour of a `?here:Source_code_position.t` parameter and so such call sites should be updated. If people want to match on an optional parameter then they should write a match explicitly. Automatically replacing options with default values in certain contexts is too much unexpected magic for too little gain.

ncik-roberts:
I can see that. @Enoumy could you adapt this PR (or post another one) that just rejects this case with an error message?

goldfirere:
I haven't looked at usages in our codebase, but I tend to agree that these arguments are not optional arguments and shouldn't be available to be passed with `?`. (I have also not read the contents of this PR -- just reacting to the comments.)

Enoumy:
Ack, will change the feature into a draft while I update the feature.

Enoumy:
Cool, I have made this situation return an explicit error in this PR. CI also just passed.

I have re-opened this feature for review. Please let me know if the approach I took was reasonable/not. Thanks!

ocaml/testsuite/tests/typing-implicit-source-positions/application_via_option.ml
```diff
@@ -0,0 +1,119 @@
...
+val f : call_pos:[%call_pos] -> unit -> lexing_position = <fun>
+|}]
+
+let _ = f ?call_pos:None ();
+[%%expect {|
+Line 1, characters 20-24:
+1 | let _ = f ?call_pos:None ();
+                        ^^^^
+Error: the [%call_pos] label 'call_pos' is not optional. It
+       cannot be applied with '?'.
```
goldfirere, line 16: I find this phrasing a little confusing. Might I suggest

```suggestion
Error: the argument labeled 'call_pos' is a [%call_pos] argument, filled in
       automatically if omitted. It cannot be passed with '?'.
```

ocaml/testsuite/tests/typing-implicit-source-positions/application_via_option.ml
```diff
@@ -0,0 +1,119 @@
+(* TEST_BELOW
+Fille
```
goldfirere, line 2: ?

ocaml/typing/typeclass.ml
```diff
@@ -1359,10 +1360,15 @@ and class_expr_aux cl_num val_env met_env virt self_scope scl =
...
                 | Some (l', sarg, _, remaining_sargs) ->
-                    if not optional && Btype.is_optional l' then
-                      Location.prerr_warning sarg.pexp_loc
-                        (Warnings.Nonoptional_label
-                           (Printtyp.string_of_label l));
+                    if not optional && Btype.is_optional l' then (
+                      if Btype.is_position l then
+                        raise
+                          (Error
+                             (sarg.pexp_loc, val_env, Nonoptional_call_pos_label (Printtyp.string_of_label l)))
```
goldfirere, line 1367: The convention in the compiler is to keep lines to 80 characters.

And it might be a small improvement to bind `Printtyp.string_of_label l` to a variable before the inner `if`.

ocaml/typing/typecore.ml
```diff
@@ -3646,9 +3647,16 @@ let collect_apply_args env funct ignore_labels ty_fun ty_fun0 mode_fun sargs ret
...
                 end;
-                if not optional && is_optional l' then
-                  Location.prerr_warning sarg.pexp_loc
-                    (Warnings.Nonoptional_label (Printtyp.string_of_label l));
+                if not optional && is_optional l' then (
+                  if is_position l
+                  then
+                    raise
+                      (Error
+                         (sarg.pexp_loc, env, Nonoptional_call_pos_label (Printtyp.string_of_label l)))
```
goldfirere, line 3655: Ditto comment from typeclass.

goldfirere (approved):
Just a few small suggestions.

ocaml/testsuite/tests/typing-implicit-source-positions/application_via_option.ml
Enoumy: Is you `?` on "Fille"? I used `implicit_argument.ml` (within this same directory) as a template for the test. The file was changed to use "Fille" in `a838fa1f9` (from looking at the git blame). Some files in this same directory were changed to use `Fille` and others use  `expect`.

I've mostly treated the ocaml compilers test suite's expect test system as a black box that has comments - I did not look into what "Fille" does/can't find much by grepping for it inside of `ocaml/ocamltest`. I am happy to change them to use `expect` if you would like.

ocaml/testsuite/tests/typing-implicit-source-positions/application_via_option.ml
Enoumy: I like it! I've changed the error message + accepted this corrected output.

ocaml/typing/typeclass.ml
Enoumy: Changed! (committed+pushed, `make ci` built locally, but still waiting for ci to finish on github actions)

ocaml/typing/typecore.ml
Enoumy: Changed! (still waiting for CI to pass)

ocaml/typing/btype.ml
Enoumy: Oh sorry, I've realized this conversation is still open. I don't think this comment is still relevant.

The `scr_pos` there was constructed with location of the function application, so they would both have the same location as ommitting the argument entirely.

ocaml/testsuite/tests/typing-implicit-source-positions/application_via_option.ml
goldfirere: I have no idea what's going on with this. But if that's what's done elsewhere, who am I to argue? I just couldn't tell if this was misspelled "file" or French "girl" or something else. Let's move on, regardless.

goldfirere (approved):

PR #2498, opened 2024-04-25
Use a fixed macos version

ccasin:
Hopefully this resolves the build errors we're seeing on macos in CI.  We should figure out some other fix that allows us to use the latest macos, but this will get us unstuck for now.


ncik-roberts (approved):
LGTM. I confirmed the runner OS is the same as earlier PRs that passed CI:

```
Current runner version: '2.315.0'
Operating System
  macOS
  12.7.4
  21H1123
```

PR #2499, opened 2024-04-25
DWARF inlined frames: middle end

mshinwell:
This PR makes fairly minor changes in/around the middle end to support generation of DWARF inlined frame information.

In `Debuginfo.t` two new fields are added, which enable the tracking of different instances of inlining, together with the ability to obtain the code ID (function symbol) corresponding to any particular inlined frame.  (The type of the latter, which is used for reliable referencing of DWARF abstract instance DIEs from concrete inlined instance DIEs, should really be `Code_id.t` but that is not available in `Debuginfo`.)  There are further details in the comment on the new `Inlined_debuginfo` module, which in due course should probably be extended to provide a new type of debuginfo for the middle end, replacing `Debuginfo.t`.

The existing infrastructure for generating `Debuginfo.t` values with inlined frames in Flambda 2 has been refactored and made to take advantage of the `Inlined_debuginfo` module.  I will do some testing before merging to check that this hasn't altered the debuginfo propagation.

cc @poechsel and @xclerc with whom this has been discussed in person.  @Gbury maybe you could cast an independent eye over the Flambda 2 parts.

### DIFF
--- a/middle_end/flambda2/simplify/env/downwards_env.ml
+++ b/middle_end/flambda2/simplify/env/downwards_env.ml
@@ -31,7 +31,7 @@ type get_imported_code = unit -> Exported_code.t
 type t =
   { round : int;
     typing_env : TE.t;
-    inlined_debuginfo : Debuginfo.t;
+    inlined_debuginfo : Inlined_debuginfo.t;
     can_inline : bool;
     inlining_state : Inlining_state.t;
     propagating_float_consts : bool;
@@ -48,11 +48,6 @@ type t =
     inlining_history_tracker : Inlining_history.Tracker.t;
     loopify_state : Loopify_state.t
   }
-
-let print_debuginfo ppf dbg =
-  if Debuginfo.is_none dbg
-  then Format.pp_print_string ppf "None"
-  else Debuginfo.print_compact ppf dbg
 
 let [@ocamlformat "disable"] print ppf { round; typing_env;
                 inlined_debuginfo; can_inline;
@@ -84,7 +79,7 @@ let [@ocamlformat "disable"] print ppf { round; typing_env;
       )@]"
     round
     TE.print typing_env
-    print_debuginfo inlined_debuginfo
+    Inlined_debuginfo.print inlined_debuginfo
     can_inline
     Inlining_state.print inlining_state
     propagating_float_consts
@@ -112,7 +107,7 @@ let create ~round ~(resolver : resolver)
   in
   { round;
     typing_env;
-    inlined_debuginfo = Debuginfo.none;
+    inlined_debuginfo = Inlined_debuginfo.none;
     can_inline = true;
     inlining_state = Inlining_state.default ~round;
     propagating_float_consts;
@@ -193,7 +188,7 @@ let enter_set_of_closures
     } =
   { round;
     typing_env = TE.closure_env typing_env;
-    inlined_debuginfo = Debuginfo.none;
+    inlined_debuginfo = Inlined_debuginfo.none;
     can_inline;
     inlining_state;
     propagating_float_consts;
@@ -461,12 +456,6 @@ let define_code t ~code_id ~code =
   let all_code = Code_id.Map.add code_id code t.all_code in
   { t with typing_env; all_code }
 
-let set_inlined_debuginfo t dbg = { t with inlined_debuginfo = dbg }
-
-let get_inlined_debuginfo t = t.inlined_debuginfo
-
-let add_inlined_debuginfo t dbg = Debuginfo.inline t.inlined_debuginfo dbg
-
 let cse t = t.cse
 
 let comparison_results t = t.comparison_results
@@ -519,6 +508,14 @@ let set_inlining_arguments arguments t =
     inlining_state = Inlining_state.with_arguments arguments t.inlining_state
   }
 
+let set_inlined_debuginfo t ~from =
+  { t with inlined_debuginfo = from.inlined_debuginfo }
+
+let set_inlined_debuginfo' t inlined_debuginfo = { t with inlined_debuginfo }
+
+let add_inlined_debuginfo t dbg =
+  Inlined_debuginfo.rewrite t.inlined_debuginfo dbg
+
 let enter_inlined_apply ~called_code ~apply ~was_inline_always t =
   let arguments =
     Inlining_state.arguments t.inlining_state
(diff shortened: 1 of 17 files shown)


middle_end/flambda2/term_basics/inlined_debuginfo.mli
```diff
@@ -0,0 +1,80 @@
...
+    came from the [apply_dbg] passed to [create]).
+
+    In addition, the first (outermost, less deep inlining) debuginfo item
+    arising from the inlined body is annotated with the function symbol from the
+    application term being inlined out.  Inductively, in association with the
+    general propagation of [Debuginfo.t] values by Flambda 2, the consequence is
+    that [Debuginfo.t] values returned from this function end up annotated with
+    function symbols at all points except the outermost frame. (The symbol for
+    that frame will be known by the backend, since it is the function into which
+    everything is ultimately being inlined.)
```
Gbury, line 78: I find this paragraph rather confusing, particularly the last part. After re-reading it a couple times, I think I'm starting to understand, but still, reformulating things might help.

middle_end/flambda2/term_basics/inlined_debuginfo.mli
```diff
@@ -0,0 +1,80 @@
...
+    item corresponding to the inlined body (i.e. every item except those at the
+    head of the list, which are those---typically just one item in fact---that
+    came from the [apply_dbg] passed to [create]).
+
+    In addition, the first (outermost, less deep inlining) debuginfo item
+    arising from the inlined body is annotated with the function symbol from the
+    application term being inlined out.  Inductively, in association with the
+    general propagation of [Debuginfo.t] values by Flambda 2, the consequence is
+    that [Debuginfo.t] values returned from this function end up annotated with
+    function symbols at all points except the outermost frame. (The symbol for
```
Gbury, line 76: doesn't this function specifically sets/overwrite the function symbol of the outermost frame with the one from the `t` value ?

middle_end/flambda2/to_cmm/to_cmm_env.ml
```diff
@@ -255,13 +255,12 @@ let enter_function_body env ~return_continuation ~exn_continuation =
 
 (* Debuginfo *)
 
-let enter_inlined_apply t dbg =
-  let inlined_debuginfo = Debuginfo.inline t.inlined_debuginfo dbg in
```
Gbury, line 259: This is probably the cause of the CI failure: you somehow need to be able to `stack`/`merge` successive `inlined_debuginfos`

Gbury, line 259: This is probably the cause of the CI failure: you somehow need to be able to `stack`/`merge` successive `inlined_debuginfos`

Gbury, line 259: This is probably the cause of the CI failure: you somehow need to be able to `stack`/`merge` successive `inlined_debuginfos`

Gbury (changes requested):

middle_end/flambda2/simplify/simplify_nullary_primitive.ml
```diff
@@ -42,8 +42,7 @@ let simplify_nullary_primitive dacc original_prim (prim : P.nullary_primitive)
     Simplify_primitive_result.create named ~try_reify:false dacc
   | Enter_inlined_apply { dbg } ->
     let dacc =
-      DA.map_denv dacc ~f:(fun denv ->
-          DE.set_inlined_debuginfo denv (DE.add_inlined_debuginfo denv dbg))
+      DA.map_denv dacc ~f:(fun denv -> DE.set_inlined_debuginfo' denv dbg)
```
Gbury, line 45: similarly to what needs to be done in `to_cmm` I think that this code should also be fixed to somehow "stack" the inlined_debuginfos, else this line will lose any previous/less deep inlined debuginfos.

middle_end/flambda2/term_basics/inlined_debuginfo.mli
```diff
@@ -76,5 +81,66 @@ val create : called_code_id:Code_id.t -> apply_dbg:Debuginfo.t -> t
...
+
+    and the following in [g]:
+
+    Paddint/61N =
+     ((+ Popaque/60N 1)
+      example.ml:2,23--26;
+      example.ml:1,19--46[871832529][FS=camlExample.f_0_3_code])
+    Pmulint/62N = (( * 2 Paddint/61N) example.ml:2,19--26)
+
+    and the following in [h]:
```
Gbury, line 103: you called in `foo` instead of `h` (at line 89)

middle_end/flambda2/to_cmm/to_cmm_env.ml
mshinwell: Fixed, or fixing...

middle_end/flambda2/term_basics/inlined_debuginfo.mli
mshinwell: I've added some more text in this comment and below, hopefully ok now.

middle_end/flambda2/simplify/simplify_nullary_primitive.ml
mshinwell: Should be fixed

middle_end/flambda2/simplify/env/downwards_env.ml
```diff
@@ -543,8 +544,12 @@ let enter_inlined_apply ~called_code ~apply ~was_inline_always t =
         in
         Inlining_state.increment_depth t.inlining_state ~by)
   in
+  let inlined_debuginfo =
+    Inlined_debuginfo.create ~called_code_id:(Code.code_id called_code)
+      ~apply_dbg:(Apply.dbg apply)
+  in
```
Gbury, line 550: I'm not sure what happens when inlining from classic mode and simplify interleave, and I suspect we could be forgetting/discarding some debuginfos, particularly here since we simply overwrite the current `inlined_debuginfo`: imagine the current value of `t.inlined_debuginfo` comes from `Enter_inlined_apply` primitive inserted during classic mode, what ensures that we do not forget the inlined debuginfo from those when we overwrite it in this function ?

Maybe just a comment/CR about this should be enough for now.

Gbury (approved):
This PR is good to be merged. I left a comment about some possible subtle interaction between classic mode and simplify, but even with that comment, I think the current PR is an improvement over the current situation.

middle_end/flambda2/simplify/env/downwards_env.ml
mshinwell: As discussed this is likely a pre-existing problem; have added a CR.

PR #2500, opened 2024-04-26
Track external symbols

poechsel:
Export in the cmx / cmxa the set of C symbols the current module / library depends on.

We say that a file `foo.ml` depends on `foo` if it contains at list one statement `external xxx = "foo"`

This will be used internally to track the precise set of symbols that must be included when linking together OCaml binaries

### DIFF
--- a/middle_end/compilenv.ml
+++ b/middle_end/compilenv.ml
@@ -57,7 +57,9 @@ let current_unit =
     ui_generic_fns = { curry_fun = []; apply_fun = []; send_fun = [] };
     ui_force_link = false;
     ui_checks = Checks.create ();
-    ui_export_info = None }
+    ui_export_info = None;
+    ui_external_symbols = [];
+  }
 
 let reset compilation_unit =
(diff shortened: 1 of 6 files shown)


middle_end/compilenv.ml
```diff
@@ -72,7 +74,14 @@ let reset compilation_unit =
...
   Hashtbl.clear exported_constants;
-  current_unit.ui_export_info <- None
+  current_unit.ui_export_info <- None;
+  current_unit.ui_external_symbols <- []
+
+let record_external_symbols () =
+  current_unit.ui_external_symbols <- (List.filter_map (fun prim ->
+      if not (Primitive.native_name_is_external prim) then None
+      else begin Some (Primitive.native_name prim) end)
+      !Translmod.primitive_declarations)
```
mshinwell, line 84: Please reformat and remove redundant `begin` .. `end`

mshinwell (changes requested):
What is the effect on build artifact sizes?

PR #2505, opened 2024-04-29
Add check that variables binding result of primitives are the correct kind

Ekdohibs:
This caught two places where we put invalid kinds, which were thankfully not dangerous.

### DIFF
--- a/middle_end/flambda2/from_lambda/closure_conversion.ml
+++ b/middle_end/flambda2/from_lambda/closure_conversion.ml
@@ -1014,6 +1014,22 @@ let close_let acc env let_bound_ids_with_kinds user_visible defining_expr
            generated. *)
         body acc body_env
       | _ -> (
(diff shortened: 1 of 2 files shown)


mshinwell (approved):

lthls:
I'm not sure I understand what's going on. `Get_tag` and `Is_int` have `K.naked_immediate` as their return kind. So why does this PR introduces a change that reads like they have a `K.tagged_immediate` return kind ?

lthls:
Ah, I found the answer. The Flambda `Get_tag` and `Is_int` primitives return a naked immediate, but their Lambda versions (for `Get_tag`, technically this is the Ilambda version) return a tagged immediate. So an extra `Tag_immediate` is added in both cases (though not at the same place), making the expected kind of the let binding `K.tagged_immediate`. Sorry for the noise.

PR #2508, opened 2024-04-30
Mixed constructor args

ncik-roberts:
Support mixed constructor args.

 * In the front end, this consists of a few pieces:
   * Augmenting `Types.variant_representation` to additionally track the "constructor representation" for each constructor of a boxed variant. This `Types.constructor_representation` is roughly parallel to `Lambda.constructor_representation`, but named slightly differently to suggest that `Types.Constructor_uniform_value` can only be used for *values*, not flat floats.
   * Extending validation in `typedecl.ml` to permit constructor args as a new kind of mixed product (and generalizing the naming)
   * Lots of plumbing.
 * From the middle end onward, this mostly consists of looking for places in mixed blocks code that assume the tag is 0, and instead plumbing through the tag from the middle-end. In prior PRs I marked such places with comments like `CR mixed blocks v1: the tag is wrong`.

The tests are meant to exercise both the typing and the translation. They did catch bugs! (For instance, they motivated #2504, and they unearthed an accidental use of `Pgenval` via `Lambda.layout_field` in `matching.ml`.)

**Review**
* @ccasin Could you please review:
  * The front-end changes, up until `lambda/`?
  * The changes in `bytecomp/`? (I know, I know, but they really are straightforward.)
* @TheNumbat Could you please review the middle-end changes? They are straightforward and just need a sanity check.
* @Ekdohibs Could you please approve the one-line chamelon change?


ncik-roberts:
On recommendation of @Ekdohibs I've removed `layout_field` to ensure that all places that used to assume that fields were values are now updated.

ocaml/testsuite/tests/typing-layouts/mixed_constructor_arguments_alpha.ml
```diff
@@ -0,0 +1,391 @@
...
+[%%expect{|
+type t_ext +=
+    A of float# * float# * int
+  | B of int
+  | C of float# * int
+  | D of float# * int * float#
+  | E of int * float# * int * float#
+|}];;
+
+(* Parameterized types *)
```
ccasin, line 221: We should have a test for

```ocaml
type 'a t = C of float# * 'a
```
Currently rejected, which seems correct.  I think you already have a variation of this where we learn the layout from a mutually recursive group.

ocaml/testsuite/tests/typing-layouts/mixed_constructor_arguments_alpha.ml
```diff
@@ -0,0 +1,391 @@
...
+ 4 |     ptr * ptr * ptr * ptr * ptr * ptr * ptr * ptr *
+ 5 |     ptr * ptr * ptr * ptr * ptr * ptr * ptr * ptr *
+ 6 |     ptr * ptr * ptr * ptr * ptr * ptr * ptr * ptr *
+...
+32 |     ptr * ptr * ptr * ptr * ptr * ptr * ptr * ptr *
+33 |     ptr * ptr * ptr * ptr * ptr * ptr * ptr * ptr *
+34 |     ptr * ptr * ptr * ptr * ptr * ptr * ptr *
+35 |     int * float#
+Error: Mixed constructors may contain at most 254 value fields prior to the flat suffix, but this one contains 255.
+|}];;
```
ccasin, line 391: I'd like to have tests for gadt syntax.  Both along the lines of the tests  you already have, but then also some where the layout is different in the different branches.  Testing quickly, this is allowed:
```ocaml
type ('a : float64) tf
type ('a : value) tv

type ('a : any) t = A : 'a tf -> 'a t
                  | B : 'b tv -> 'a t
```
But this variation on t is rejected:
```ocaml
type ('a : any) t = A : float# * 'a tf -> 'a t
                  | B : 'b tv * float# -> 'a t
```
why?

ccasin (commented):
Have only read the tests - sorry I didn't have more time today.

(I'd also like to think a bit more about the one that fails in the final consistency check, but figured it might be useful to post this partial feedback anyway)

middle_end/flambda2/from_lambda/dissect_letrec.ml
```diff
@@ -247,7 +247,10 @@ let rec prepare_letrec (recursive_set : Ident.Set.t)
             (id, def) :: defs, Lambda.Lvar id :: args)
         args ([], [])
     in
-    let arg_layout = Typeopt.layout_union layout_field layout_block in
+    (* [arg_layout] is more general than any of the possible layouts of [args]:
+       The arguments to [Pmakeblock]/[Pmakearray] are value fields, and the
+       argument to [Pduprecord] is a block. *)
+    let arg_layout = Typeopt.layout_union layout_value_field layout_block in
```
TheNumbat, line 253: Is this change because the fields can now be non-values, or is it a bug fix?

TheNumbat (commented):
The `backend/` and `middle_end/` changes look good (and are mechanical).

middle_end/flambda2/from_lambda/dissect_letrec.ml
ncik-roberts: It's not actually a change: `layout_field` and `layout_value_field` are defined to the same thing. But I carved up `layout_field` into several more specific names (including `layout_value_field`, for fields that are known to be values) in light of the fact that fields can now be non-values.

ocaml/testsuite/tests/mixed-blocks/generated_byte_test.ml
```diff
@@ -356,6 +356,1171 @@ type t296 = { str0 : string; mutable float1 : float; float_u2 : float# }
 type t297 = { mutable float0 : float; mutable str1 : string; mutable str2 : string; mutable str3 : string; mutable float_u4 : float#; mutable float_u5 : float# }
 type t298 = { mutable str0 : string; mutable float1 : float; mutable str2 : string; float_u3 : float#; mutable imm4 : int }
 type t299 = { float0 : float; mutable imm1 : int; mutable float_u2 : float#; mutable imm3 : int }
+type t300 =
```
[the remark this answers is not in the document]
ncik-roberts: (It actually grew from ~5000 lines to ~13,000.)

I think it is easy to control the size of this file. There are knobs in the `test.ml` file that let you say "I want to test fewer combinations". 

> But probably we'll add more things that add more dimensions for it to grow exponentially on later 

I don't think it will be exponential. I think it will be more like adding 5,000 more lines. But again it is easy to reduce this (and easy for me to run the test to gauge whether it's taking too long.

>  Should we consider splitting it up into multiple files or testing fewer combinations?

I don't think so at this point in time.

ocaml/testsuite/tests/typing-layouts/mixed_constructor_arguments_alpha.ml
ncik-roberts: Sounds good, done.

ocaml/testsuite/tests/typing-layouts/mixed_constructor_arguments_alpha.ml
ncik-roberts: I think the second `t` should certainly be rejected. `'a tf` is itself a value (though its argument is float64).

ncik-roberts: I've added some tests, including a version of the test you proposed with `type ('a : float64) tf : float64` instead.

ocaml/testsuite/tests/typing-layouts/mixed_constructor_arguments_alpha.ml
```diff
@@ -0,0 +1,391 @@
+(* TEST
```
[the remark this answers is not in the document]
ncik-roberts: Good point, I've added a test for inlined records. Those just aren't supported yet, I'll do so in a follow-up PR.

That bad error message is evidence of me trying to save work in refactoring error message construction (when everything's just going to get simpler with inline records). That is a wrong error message, though. So I've just done the refactor. Hopefully I'll get to delete some of it in a follow-up PR.

ocaml/testsuite/tests/mixed-blocks/constructor_args.ml
```diff
@@ -0,0 +1,337 @@
...
+    Printf.printf
+      "Test (construct and destruct): %f = %f (%s)\n"
+      sum1
+      sum2
+      (if Float.equal sum1 sum2 then "PASS" else "FAIL")
+  in
+  ()
+
+(**************************)
+(* Test: recursive groups *)
```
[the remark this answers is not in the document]
ncik-roberts: I don't think the error message has anything to do with this PR.  I can also write:

```ocaml
type t = C of float * int

let _ =
  let rec x = C (42.0, x); 3 in
  ();;
```

and get the same crash. #2394 fixes the crash for all cases, both my example and yours.

I do think that the acceptance of this program is a problem, though. Compilation fills in a dummy int `0xbbbb` for the value under construction, which is wrong for mixed blocks. Your program doesn't segfault (post #2394) but I suspect that may just be because the middle-end notices that the constructed block is local to the expression and optimizes it out.

I'll change the rec-check as you suggest.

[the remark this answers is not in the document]
ncik-roberts: That's done, with an appropriate test.

ncik-roberts:
RE: "the one that fails the final consistency check." I don't think you have to spend any time thinking about it. This PR is not "about" that error message, and that error message also existed prior to this PR.

 This test is just straightforwardly copied from the same test for mixed blocks, which was straightforwardly copied from the same test for other datatypes.

E.g. from `typing-layouts/datatypes.ml`:

```
type 'a t8_5 = { x : 'a t8_6; y : string}
and 'a t8_6 = 'a float64_t;;
[%%expect {|
Line 1, characters 21-28:
1 | type 'a t8_5 = { x : 'a t8_6; y : string}
                         ^^^^^^^
Error: Layout mismatch in final type declaration consistency check.
       This is most often caused by the fact that type inference is not
       clever enough to propagate layouts through variables in different
       declarations. It is also not clever enough to produce a good error
       message, so we'll say this instead:
         The layout of 'a is float64, because
           of the definition of float64_t at line 2, characters 0-29.
         But the layout of 'a must overlap with value, because
           it instantiates an unannotated type parameter of t8_5, defaulted to layout value.
       A good next step is to add a layout annotation on a parameter to
       the declaration where this error is reported.
|}]
```

ocaml/testsuite/tests/mixed-blocks/generated_byte_test.ml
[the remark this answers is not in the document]
ccasin: Fair enough!

ocaml/testsuite/tests/typing-layouts/mixed_constructor_arguments_alpha.ml
ccasin: Oops!  Thanks for adding some gadt tests, and sorry for the noise in this example.

(I of course meant to write
```ocaml
type ('a : float64) tf = 'a
type ('a : value) tv = 'a
```
which works as expected)

ccasin:
> RE: "the one that fails the final consistency check." I don't think you have to spend any time thinking about it. This PR is not "about" that error message, and that error message also existed prior to this PR.

Yeah, I'm aware this issue long predates this PR, and won't slow this down for it.

ccasin (approved):
Looks good, only minor suggestions/remarks.

ocaml/typing/typedecl.ml
```diff
@@ -1136,34 +1161,164 @@ let update_constructor_arguments_jkinds env loc cd_args jkinds =
...
+        Element_repr.mixed_product_flat_suffix arg_reprs
+          ~on_flat_field_expected:(fun ~non_value ~boxed ->
+              let violation =
+                Flat_constructor_arg_expected
+                  { non_value_arg = non_value;
+                    boxed_arg = boxed;
+                  }
+              in
+              raise (Error (loc, Illegal_mixed_product violation)))
+    | Cstr_record fields ->
```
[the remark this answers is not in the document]
ncik-roberts: Ah, yes, sorry, this was an artifact of me starting to tackle both inline records / constructor args in the same PR. Comment added.

ocaml/typing/types.mli
```diff
@@ -793,6 +806,7 @@ type constructor_description =
     cstr_arity: int;                    (* Number of arguments *)
     cstr_tag: tag;                      (* Tag for heap blocks *)
     cstr_repr: variant_representation;  (* Repr of the outer variant *)
+    cstr_shape: constructor_representation; (* Repr of the constructor itself *)
```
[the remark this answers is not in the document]
ncik-roberts: I also tried something like this and ran aground. I can't remember if it was the same place as you or somewhere different. In any case, I agree both that this is slightly sad and that it's not immediately clear how (or an immediate win) to improve it, so I've just added a comment to lament.

(Also, I realized another prequisite to eliminating this duplication: the `Void` layout should be tracked differently in the mixed block shape. I'm going to make a follow-up PR that batches that with some other type cleanups that Mark requested.)

Ekdohibs (approved):
For `chamelon` changes.

ncik-roberts:
After discussing with Stephen, it's not immediately clear whether #2504 is a good idea. Instead, I'm going to:
  * fix up the tests to reflect the current behavior (only raise in poly compare if you compare mixed blocks whose tags are the same)
  * merge this as-is to un-nest the PR structure
  * decide offline what the proper polymorphic compare semantics is

PR #2512, opened 2024-04-30
Zero alloc: assume error means never raises

gretay-js:
On top of #2501. 
Change the meaning of `[@zero_alloc assume error]` to "never raises" in addition to "never returns normally".

Add `never_raise` field to `Builtin_primitives.Zero_alloc.Assume` and propagate it to the backend, similarly to the existing `never_returns_normally` field. 

Adds a test.

### DIFF
--- a/tests/backend/checkmach/test_assume_error.ml
+++ b/tests/backend/checkmach/test_assume_error.ml
@@ -152,7 +152,7 @@ module A5 = struct
 end
 
 (* A6: shows that "assume error" is not enough to prove "strict",
-   because the call to [string_of_int] is not known to not diverge.
+   because the call to [string_of_int] may allocate and is not known to not diverge.
 *)
 module A6 = struct
   let[@inline never][@local never] h s = [Random.int s ; Random.int s]
--- a/tests/backend/checkmach/test_assume_error.ml
+++ b/tests/backend/checkmach/test_assume_error.ml
@@ -1,0 +1,225 @@
+(* A1: zero_alloc check fails on [foo] because there is an allocation on the
+   path to exceptional return from the call to [f x].
+   This exception is handled in foo and then [foo] returns normally,
+   making this allocation on a normal return of [foo].
+*)
+module A1 = struct
+  let[@inline never][@local never] h s = [Random.int s ; Random.int s]
+
+  let[@inline never][@local never][@zero_alloc] g x =
+    if x < 0 then Sys.opaque_identity (x+1) else raise (Failure ((string_of_int) x))
+
+  let[@inline never][@local never] handle_exn exn =
+    (match exn with
+     | Failure s -> print_string s; print_newline ()
+     | _ -> failwith "Boo");
+    0
+  ;;
+
+  let[@zero_alloc] f x =
+    match g x with
+    | exception exn ->
+      (handle_exn[@zero_alloc assume never_returns_normally]) exn
+      |> h |> List.hd  (* ignore this allocation *)
+    | answer -> (answer * 2)
+
+  let[@zero_alloc] foo x =
+    try f x with
+    | _ -> 0
+end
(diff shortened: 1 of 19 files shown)


tests/backend/checkmach/test_assume_error.ml
```diff
@@ -0,0 +1,225 @@
...
+      |> h |> List.hd  (* ignore this allocation *)
+    | answer -> (answer * 2)
+
+  let[@zero_alloc] foo x =
+    try
+      let y = f x in
+      (x,y)  (* report this allocation *)
+    with
+    | _ ->
+      Printf.printf "BOO %x" x;  (* ignore this indirect call *)
```
ccasin, line 90: Why do we want to ignore this call?  It doesn't look like it's on a path to exceptional return, or necessarily after the `assume error`.

tests/backend/checkmach/test_assume_error.ml
```diff
@@ -0,0 +1,225 @@
...
+  let[@zero_alloc] foo x =
+    try f x
+    with
+    | _ ->
+      Printf.printf "BOO %x" x;  (* ignore this indirect call *)
+      0
+end
+
+(* A6: shows that "assume error" is not enough to prove "strict",
+   because the call to [string_of_int] is not known to not diverge.
```
ccasin, line 155: Why would knowing `string_of_int` doesn't diverge help us here?  Aren't we still going to end up allocating the string on the path to `handle_exn`, which should be disallowed by the strict check?

tests/backend/checkmach/test_assume_error.ml
```diff
@@ -0,0 +1,225 @@
...
+    match g x with
+    | exception exn ->
+      handle_exn exn
+      |> h |> List.hd  (* ignore this allocation *)
+    | answer -> (answer * 2)
+
+end
+
+(* A7 is the same as A6 but without [string_of_int] and we can prove strict annotation on
+   [f]*)
```
ccasin, line 180: (Similarly I am surprised this works)

ccasin (commented):
I left some comments in the tests.  I think I clearly don't understand the intended meaning of something here.

tests/backend/checkmach/test_assume_error.ml
gretay-js: From `assume error` in `f` , the analysis concludes that `f` cannot raise (because `List.hd` is not reachable), so `printf` line in `foo` is not reachable, as far as the analysis knowns.

tests/backend/checkmach/test_assume_error.ml
gretay-js: The analysis thinks that `string_of_int` can diverge, so  it can't prove that `f` is "zero_alloc strict". I'm trying to show that `[@zero_alloc assume error]` doesn't affect the `div`component of the abstract value. It doesn't matter that we allocate on the path to `handle_exn` because "assume error" causes all allocations before and after it to be ignored, as if we had "exit 1" inside `handle_exn`, rather than a `raise`. (In the analysis is implemented by "assume error" does not return and does not raise.)

ccasin: Thanks.  To check my understanding, the error say "may allocate on a path to diverge".  Is the potential allocation also within string_of_int? I don't see another.

ccasin (approved):
Approving for tests and everything under `ocaml/`.  @xclerc should probably review the tests and other folders (though I can take a stab at it if needed)

tests/backend/checkmach/test_assume_error.ml
gretay-js: yes, the allocation is within `string_of_int`. I updated the comment on the test.

xclerc (approved):

PR #2514, opened 2024-04-30
add `typemode` to `dynlink`

riaqn:
What the title says. Not sure if this suffices.


riaqn:
@goldfirere If `types.ml` calls `typemode`, then we need this for the build to pas

PR #2516, opened 2024-05-01
Stop checking layout enabledness in parser

ncik-roberts:
Fix a bug where extensions were checked in the parser. This is bad because extension-checking relies on mutable state, and there are some workflows where the parser runs in ppxes (which have all extensions enabled) and not the main compiler process (which usually has some extensions disabled).

The two lingering cases:
  * unboxed literals. The error message is now a little worse but this is basically fine as the extension is enabled by default.
  * `float#`. We no longer attempt to do any check that `float#` is permitted. We didn't even try for `int32#`, etc. I resolve the irregularity in favor of deleting the `float#` check, but it would be easy to add a check in `typetexp.ml` for all unboxed types if that seems better. (I don't have context on why the decision was made the way it was for `int32#`.)

Testing: I add a regression test that fails in the first commit.

### DIFF
--- a/ocaml/parsing/parser.mly
+++ b/ocaml/parsing/parser.mly
@@ -862,7 +862,7 @@ module Constant : sig
   type loc := Lexing.position * Lexing.position
 
   val value : Parsetree.constant -> t
-  val unboxed : loc:loc -> Jane_syntax.Layouts.constant -> t
+  val unboxed : Jane_syntax.Layouts.constant -> t
   val to_expression : loc:loc -> t -> expression
   val to_pattern : loc:loc -> t -> pattern
 end = struct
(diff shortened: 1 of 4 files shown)


goldfirere (approved):

goldfirere:
This looks like an improvement. But I'm concerned that we never check for `float#` (or the other unboxed types). Even though we're not concerned about stability properties, we're still concerned about compatibility. This compatibility takes two forms:

1. The ability to open-source internal libraries that will work with the upstream compiler. This is probably OK, because ocamlformat erases unboxed things to boxed ones.
2. The ability to use code from the open-source ecosystem with our compiler. Unboxed types steals a tiny bit of syntax, and so it is very remotely possible that some code we import uses that syntax. So we compile code from the open-source ecosystem without any extensions enabled, even ones proved safe. In the stolen syntax case, this would yield an error rather than just reinterpret the code.

I'm fine with merging this PR as-is -- the problem I'm worried about here is independent -- but while you're in town, could you check to see whether the problem is real? If it is, we should either fix here or probably make an internal ticket. Thanks!

ncik-roberts:
> The ability to open-source internal libraries that will work with the upstream compiler. This is probably OK, because ocamlformat erases unboxed things to boxed ones.

I agree that this is not a concern.

I will log an internal ticket but I think this is not a concern. I'll put more context on the ticket.

PR #2517, opened 2024-05-01
Zero alloc assume on inlined trywith

gretay-js:
On top of #2458. 
Attempts to fix the second issue described in #2458: when a function annotated with `[@zero_alloc assume]` is inlined and the function's body contains `try E with ...` or `match E with | exception ... `, then `E` is annotated with a "strict" assume instead of a relaxed assume. This is a draft because 
1) It only works for "assume" on functions and not on calls.  We propagate "assume" on inlined calls in Flambda2, and I don't know how to implement this heuristics on top of the special handling of inlined debug info in Flambda2.
2) In some rare cases, there is a soundness problem: the check fails when the function is not inlined and passes when the function is inlined, even though there is no difference no difference in allocations. For example, the check on `foo` passes when `bar` is inlined but fails with `bar` is not inlined. 

```
let[@inline never][@zero_alloc] f x = if x = 0 then  raise (Exn x) else x * x
let[@inline never][@zero_alloc] g x n = let y = x+n in if  y > 10 then raise (Exn y) else y
let[@zero_alloc assume] bar x =
    try
      f x
    with
    | Exn n -> n

let[@zero_alloc] foo x =
    try bar x with
    | _ ->
      (Sys.opaque_identity 0)
```

### DIFF
--- a/backend/checkmach.ml
+++ b/backend/checkmach.ml
@@ -833,10 +833,11 @@ end = struct
   let transform_call t ~next ~exn callee w ~desc dbg =
     report t next ~msg:"transform_call next" ~desc dbg;
     report t exn ~msg:"transform_call exn" ~desc dbg;
+    let v = find_callee t callee ~desc dbg w in
     let effect =
       match Metadata.assume_value dbg ~can_raise:true w with
-      | Some v -> v
-      | None -> find_callee t callee ~desc dbg w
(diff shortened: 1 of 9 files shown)


gretay-js:
We are considering a different approach for propagating the scope of `assume` using enter and exit markers or similarly to the way inlined function boundaries are tracked in classic mode.

PR #2519, opened 2024-05-01
Unify float sqrt/bit-cast intrinsics

TheNumbat:
Simd_selection is now used to translate calls to sqrt and bit-cast for both float and float32.
This removes the `Ifloatsqrt` specific instruction; it has been subsumed by the SSE2 sqrt instructions.

To preserve compatibility, the float64 intrinsics are translated even without the `[@@builtin]` annotation.

Additionally, the `move` function in `emit.mlp` should now never be called with registers of different types.
It now asserts this, resolving @gretay-js's CR there. Doing so uncovered a bug in selectgen (see diff).

Existing tests should cover the changes to float64; also added a float32 sqrt-with-mem test. Additional testing for the float32 cases is in #2520.


TheNumbat:
@xclerc I've extracted the flambda2 changes to another PR, you can disregard them

backend/amd64/emit.mlp
```diff
@@ -784,33 +784,21 @@ let emit_global_label s =
 let move (src : Reg.t) (dst : Reg.t) =
   if src.loc <> dst.loc then
     begin match src.typ, src.loc, dst.typ, dst.loc with
-    | (Float | Float32), Reg.Reg _, (Float | Float32), Reg.Reg _
-    | Vec128, _, Vec128, _ ->
-      (* Vec128 stack slots are always aligned. *)
+    | Float, Reg _, Float, Reg _
```
xclerc, line 787: (Was maybe easier to read with the or-pattern.)

backend/amd64/emit.mlp
```diff
@@ -784,33 +784,21 @@ let emit_global_label s =
...
       I.movss (reg src) (reg dst)
-    | Float, _, Int, _ | Int, _, Float, _ ->
-      (* CR-soon gyorsh: this case is used by the bits_of_float/float_of_bits intrinsics.
-         They should instead generate a separate Ispecific and this case should be
-         removed. *)
-      I.movq (reg src) (reg dst)
     | (Int | Val | Addr), _, (Int | Val | Addr), _ ->
       I.mov (reg src) (reg dst)
-    | Vec128, _, _, _ | _, _, Vec128, _ ->
+    | _ ->
```
xclerc, line 797: I think we should avoid a catch-all here.

backend/amd64/selection.ml
```diff
@@ -298,53 +298,19 @@ method! select_operation op args dbg =
...
-      | [Cop(Cload { memory_chunk = Word_int; mutability = mut; is_atomic }, [loc], _dbg)] ->
-        let c = Double in
-        let (addr, arg) = self#select_addressing c loc in
-        Iload { memory_chunk = c;
-                addressing_mode = addr;
-                mutability = mut;
-                is_atomic; }, [arg]
-      | _ -> Imove, args)
+  (* Special cases overriding C implementations (regardless of [@@builtin]). *)
+  | Cextcall { func = ("sqrt" as func); _ }
```
xclerc, line 302: Don't we also want to catch `"sqrtf"`?

backend/amd64/selection.ml
```diff
@@ -298,53 +298,19 @@ method! select_operation op args dbg =
...
-         Ispecific Ipause, args
-      | "caml_load_fence", ([|Val|] | [| |]) ->
-         Ispecific Ilfence, args
-      | "caml_store_fence", ([|Val|] | [| |]) ->
-         Ispecific Isfence, args
-      | "caml_memory_fence", ([|Val|] | [| |]) ->
-         Ispecific Imfence, args
+  | Cextcall { func; builtin = true; _ } ->
+      begin match func with
+      | "caml_rdtsc_unboxed" -> Ispecific Irdtsc, args
```
xclerc, line 308: Why are we no longer matching on the
return type? I am not necessarily
implying it was useful to do so - I am
asking mostly out of curiosity.

xclerc (approved):

backend/amd64/emit.mlp
TheNumbat: Yes, except it didn't catch float<->float32 moves!

backend/amd64/emit.mlp
TheNumbat: Fixed

backend/amd64/selection.ml
TheNumbat: `"sqrtf"` is allowed in `select_operation_sse2` - that means the intrinsic requires `[@@builtin]`, but I think that's fine since it didn't already exist in the stdlib

backend/amd64/selection.ml
TheNumbat: No particular reason; it just made the or pattern cleaner (and we don't check it elsewhere anyway)

xclerc: According to `git blame`, @gretay-js added
these check; Greta are you OK with them
being deleted?

TheNumbat: I'll check with Greta, merging for now

backend/amd64/selection.ml
gretay-js: There is an `external` declaration for `sqrt` in `stdlib` and its compiled directly to `sqrt` call to libc for native code.  If we put `builtin` and require it to go through sse2, wouldn't that version be ignored and the old one used instead? Can you make a test to check that both ways work?

backend/amd64/selection.ml
gretay-js: This was added after early versions of intrinsics had mismatches between the string name of the intrinsic and its intended meaning. It's really easy to get the names wrong or miss `unboxed` annotation on the `external` declarations. Especially when input and output register types differ, as in `float_of_bits` and back above. We added this check as a way to catch some such mismatches. It's a shame that we didn't keep it up as more intrinsics were added but if it's the only place remaining, it's OK to delete. Can we add a TODO somewhere to consider adding such checks.

TheNumbat: Yeah, this is the only place. I'll add a CR to reconsider this in simd_selection

backend/amd64/selection.ml
TheNumbat: I wasn't planning on adding `builtin` to the `sqrt` in stdlib/base/core, since this case still applies even without it.
Is that fine?

gretay-js: @TheNumbat I was referring to your plan on having `sqrt` in `sse2` selection with `builtin`, I understand that you are not adding `@@builtin` in stdlib.  You said above

> I think that's fine since it didn't already exist in the stdlib

but it does exist in stdlib, and it doesn't have `builtin` there, so there will be two places where `sqrt` intrinsics is translated to an instructon.

TheNumbat: (Discussed with Greta offline)

PR #1555, opened 2023-07-11
Delete unused x87 FP operations from amd64 backend

TheNumbat:
We do not use the x87 floating point instructions on modern x86_64 CPUs, so this PR removes them.
This change allows the SIMD PR (#1499) to use the same register type `regf` for representing `XMM` regs.  

The operations are not removed from the upstream backend (`ocaml/backend`).

PR #2523, opened 2024-05-02
Remove leftover non-applicable comments

dustanddreams:
#1555 removed support for 80387 floating-point instructions, since amd64 systems use SSE instructions instead. However, it kept the comments mentioning a historical bug in the AT&T assembler, for which a workaround had to be implemented, while removing the workaround code.

This PR mops up these comments.

Similarly to #1555, the original files in `ocaml/` are kept unchanged.

### DIFF
--- a/backend/x86_dsl.ml
+++ b/backend/x86_dsl.ml
@@ -15,19 +15,9 @@
 
 (** Helpers for Intel code generators *)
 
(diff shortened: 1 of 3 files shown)


mshinwell (approved):
Thanks!

PR #2526, opened 2024-05-02
Disable upstream CI builds

mshinwell:
The PR you've all been waiting for.

We believe we can disable the "upstream" builds now.  Apart from the fact they don't currently work for runtime5, they cost a lot of maintenance time, especially for changes such as those editing `Lambda`.  The reason for keeping them was so that an OPAM build could be done from the `ocaml/` subtree, but thanks to recent work by @chambart (followed up by @avsm), we can ditch them.  It will no longer be necessary to use anything other than the `dune` builds when developing.

The relevant links for the OPAM repo work that enable installation of a full flambda-backend are:
* https://github.com/ocaml-flambda/flambda-backend/blob/main/HACKING.md#testing-the-compiler-built-locally-with-opam-new-method
* https://github.com/ocaml/opam-repository/pull/25757

I currently plan to follow up with PRs to remove some parts of the `ocaml/` subtree that we won't use any more (in particular `asmcomp/` and hopefully all of `middle_end/`).


mshinwell:
(Waiting for @stedolan, to discuss whether removing this is going to make importing from upstream harder.)

stedolan (approved):

PR #2527, opened 2024-05-02
New mode crossing representation

riaqn:
This is based on #2344, because we need to support calculating the mode crossing of "a @@ m" where `m` is modality.

This PR refactors the mode-crossing logic, without any observable changes. Future PRs based on this will enable more mode crossing.


goldfirere:
What does this allow? I looked for the testsuite changes, but didn't spot the interesting bit.

riaqn:
superseded by #3639

PR #2528, opened 2024-05-02
Fix initialization of unboxed numbers in static blocks

TheNumbat:
Flambda2 was incorrectly initializing unboxed static closure slots and static constants.

- Closure slots were all assumed to be `Word_val`, which was interpreted as a pointer and hence passed to setfield. It would then call `caml_initialize` with a gc-unfriendly value, like an unboxed float. This also wouldn't fully copy a vec128.

- Static constant boxed int64s and int64 arrays were incorrectly calling setfield with the unboxed int64 value due to using `Word_int`.
 
- When `Config.runtime5` was set, `make_update` unconditionally used setfield.

- Naked vec128s would be written off the end of the closure block (this won't actually occur until unboxed vectors are exposed).

I found this in #2519, which re-enabled the assertion that we do not move between float & int registers. That was triggered when attempting to pass an unboxed float to `caml_initialize`.

### DIFF
--- a/middle_end/flambda2/to_cmm/to_cmm_shared.ml
+++ b/middle_end/flambda2/to_cmm/to_cmm_shared.ml
@@ -259,7 +259,8 @@ let invalid res ~message =
 
 module Update_kind = struct
   type kind =
-    | Value
+    | Pointer
+    | Immediate
     | Naked_int32
     | Naked_int64
     | Naked_float
@@ -271,7 +272,9 @@ module Update_kind = struct
       stride : int
     }
 
-  let values = { kind = Value; stride = Arch.size_addr }
+  let pointers = { kind = Pointer; stride = Arch.size_addr }
+
+  let tagged_immediates = { kind = Immediate; stride = Arch.size_addr }
 
   let naked_int32s = { kind = Naked_int32; stride = 4 }
 
@@ -296,7 +299,8 @@ let make_update env res dbg (kind : Update_kind.t) ~symbol var ~index
   let cmm =
     let must_use_setfield =
       match kind.kind with
-      | Value -> Some Lambda.Pointer
+      | Pointer -> Some Lambda.Pointer
+      | Immediate -> Some Lambda.Immediate
       | Naked_int32 | Naked_int64 | Naked_float | Naked_float32 | Naked_vec128
         ->
         (* The GC never sees these fields, so we can avoid using
@@ -312,7 +316,8 @@ let make_update env res dbg (kind : Update_kind.t) ~symbol var ~index
     | None ->
       let memory_chunk : Cmm.memory_chunk =
         match kind.kind with
-        | Value -> Misc.fatal_errorf "update_kind requires using setfield."
+        | Pointer | Immediate ->
+          Misc.fatal_errorf "update_kind requires using setfield."
         | Naked_int32 -> Thirtytwo_signed
         | Naked_int64 -> Word_int
         | Naked_float -> Double
--- a/middle_end/flambda2/to_cmm/to_cmm_shared.ml
+++ b/middle_end/flambda2/to_cmm/to_cmm_shared.ml
@@ -283,8 +283,6 @@ module Update_kind = struct
 
   let naked_vec128s = { kind = Naked_vec128; stride = 16 }
 
-  let naked_int32_fields = { kind = Naked_int32; stride = Arch.size_addr }
-
   let naked_float32_fields = { kind = Naked_float32; stride = Arch.size_addr }
 
   let naked_vec128_fields = { kind = Naked_vec128; stride = Arch.size_addr }
--- a/middle_end/flambda2/to_cmm/to_cmm_shared.ml
+++ b/middle_end/flambda2/to_cmm/to_cmm_shared.ml
@@ -286,6 +286,8 @@ module Update_kind = struct
 
   let naked_vec128s = { kind = Naked_vec128; stride = 16 }
 
+  let naked_int32_fields = { kind = Naked_int32; stride = Arch.size_addr }
+
   let naked_float32_fields = { kind = Naked_float32; stride = Arch.size_addr }
 
   let naked_vec128_fields = { kind = Naked_vec128; stride = Arch.size_addr }
--- a/middle_end/flambda2/to_cmm/to_cmm_shared.ml
+++ b/middle_end/flambda2/to_cmm/to_cmm_shared.ml
@@ -300,16 +300,21 @@ let make_update env res dbg (kind : Update_kind.t) ~symbol var ~index
   in
   let cmm =
     let must_use_setfield =
-      match kind.kind with
-      | Pointer -> Some Lambda.Pointer
-      | Immediate -> Some Lambda.Immediate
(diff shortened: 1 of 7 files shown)


TheNumbat:
Note - writing an unboxed int32 into the closure environment actually needs to write 64 bits, because `Project_value_slot` loads `Naked_int32s` with the `Word_int` memory chunk instead of the sign-extending `Thirtytwo_signed`.  This relies on the incoming int32 to already have been sign extended, which seems a bit dangerous.

Perhaps we should instead make `Project_value_slot` sign extend.

mshinwell:
I will discuss this with @TheNumbat

middle_end/flambda2/to_cmm/to_cmm_shared.ml
```diff
@@ -258,48 +257,80 @@ let invalid res ~message =
...
+        | Naked_float -> Double
+        | Naked_float32 -> Single { reg = Float32 }
+        | Naked_vec128 -> Onetwentyeight_unaligned
+      in
+      let addr =
+        if index * kind.stride = 0
+        then symbol
+        else
+          Cmm.(
+            Cop (Cadda, [symbol; Cconst_int (index * kind.stride, dbg)], dbg))
```
Gbury, line 334: I'd rather we keep using `field_address ` as much as possible, or introduce another function with a similar semantics if needed (for exotic/different strides), rather than doing the adress computation manually here.

middle_end/flambda2/to_cmm/to_cmm_shared.mli
```diff
@@ -86,20 +86,34 @@ val bound_parameters :
...
+  val naked_float32s : t
+
+  val naked_float32_fields : t
+
+  val naked_vec128s : t
+
+  val naked_vec128_fields : t
+end
 
 (** Make an update to a statically-allocated block. *)
```
Gbury, line 113: This comment should be updated to mention that this function only works if all elements stored in the block are of the same width. Alternatively, the code should be modified to track separately the position within the block where the update has to take place.

Gbury (changes requested):

middle_end/flambda2/to_cmm/to_cmm_shared.ml
TheNumbat: Yeah, I copied it since `field_address` assumes the stride is the same as the write width. I'll add another helper function

middle_end/flambda2/to_cmm/to_cmm_shared.mli
TheNumbat: Added comments on the update kinds

TheNumbat:
Sounds good; the change to `memory_chunk_of_kind` makes int32 loads sign-extend.

middle_end/flambda2/to_cmm/to_cmm_static.ml
```diff
@@ -206,41 +206,41 @@ let static_const0 env res ~updates (bound_static : Bound_static.Pattern.t)
...
+      static_boxed_number ~kind:C.Update_kind.naked_floats ~env ~symbol ~default
         ~emit:C.emit_float_constant ~transl ~structured v res updates
     in
     env, res, updates
   | Block_like symbol, Boxed_int32 v ->
     let structured i = Cmmgen_state.Const_int32 i in
     let res, env, updates =
-      static_boxed_number ~kind:Word_int ~env ~symbol ~default:0l
-        ~emit:C.emit_int32_constant ~transl:Fun.id ~structured v res updates
+      static_boxed_number ~kind:C.Update_kind.naked_int32s ~env ~symbol
```
TheNumbat, line 225: This means the top 32 bits of the static int32 will be zero rather than sign extended.
I believe this is fine because all int32 loads in `cmm_helpers` are sign-extending.

middle_end/flambda2/to_cmm/to_cmm_shared.ml
```diff
@@ -86,9 +86,8 @@ let memory_chunk_of_kind (kind : Flambda_kind.With_subkind.t) : Cmm.memory_chunk
     | Immediate_array | Unboxed_int32_array | Unboxed_int64_array
     | Unboxed_nativeint_array | Value_array | Generic_array ->
       Word_val)
-  | Naked_number (Naked_int32 | Naked_int64 | Naked_nativeint | Naked_immediate)
-    ->
-    Word_int
+  | Naked_number (Naked_int64 | Naked_nativeint | Naked_immediate) -> Word_int
+  | Naked_number Naked_int32 -> Thirtytwo_signed
```
TheNumbat, line 90: This makes loads via `Project_value_slot` sign extend.

middle_end/flambda2/to_cmm/to_cmm_static.ml
```diff
@@ -44,8 +44,8 @@ let rec static_block_updates symb env res acc i = function
       static_block_updates symb env res acc (i + 1) r
     | Dynamically_computed (var, dbg) ->
       let env, res, acc =
-        C.make_update env res dbg Word_val ~symbol:(C.symbol ~dbg symb) var
-          ~index:i ~prev_updates:acc
+        C.make_update env res dbg C.Update_kind.pointers
```
TheNumbat, line 47: If `var` is a tagged immediate this is overly conservative, but it doesn't look like we can check that here. (This preserves the existing behavior.)

TheNumbat:
Added some more edits / explanations after discussing with @mshinwell

middle_end/flambda2/to_cmm/to_cmm_static.ml
Gbury: I'd rather keep the corner behavior for that, if possible. Else, such a change would need a very careful review of all cmm generation code and related assumptions and comments.

TheNumbat: It's easy enough to keep, but surely the relevant cmm code is correct given `emit_int32_constant` already zeroes the top bits, right?

mshinwell: Guillaume and I discussed.  We think it's safest to ensure that every int32 in an int64-sized slot is sign extended.  I'll push a commit soon which changes emit_int32_constant to sign extend (it's unclear why this wasn't happening before, it might very well just have been an oversight).  This particular case here can use "naked_int32_fields".

mshinwell: Likewise above I'll make the equivalent change for the float32 case.

mshinwell:
Please don't rebase this yet, as I'm still reviewing/editing.

mshinwell (approved):

mshinwell:
@TheNumbat I've finished.  I suggest looking at my new changes commit by commit, but read them all before commenting...

middle_end/flambda2/to_cmm/to_cmm_static.ml
mshinwell: I've also added a comment in `cmm.mli` explaning the conventions/rules.

TheNumbat:
New changes look good, merging

PR #2529, opened 2024-05-03
Backport 13099

ncik-roberts:
Backport ocaml/ocaml#13099 to fix a common case where the compiler raises when producing a type error because it tries to load a cmi.


ncik-roberts:
Some test failures here (but not upstream) suggest a small improvement: just suppress cmi loading for sub-msgs, and don't reset the printing env. I've pushed that improvement. I'm not sure why the same tests passed upstream (and I checked the test inputs are the same). I don't think it's worth spending time on getting these implementations to line up exactly. Instead I've left a comment explaining my sin.

goldfirere (approved):
I don't understand the details of what's going on here, but I don't think it's worth the time to delve in. Any trouble will show up in testing.

ccasin (approved):
This is the same as upstream but for the sin Nick mentions, and I checked that it is not a mortal one.

ccasin:
Ah, Richard beat me to this by like 10 seconds.

PR #2531, opened 2024-05-03
Ban void in mixed products

ncik-roberts:
I'm 90% sure that mixed products with void components would be miscompiled somewhere along the line. This PR records this observation as a typing error rather than letting this be a silent footgun when we eventually get around to supporting void properly.

### DIFF
--- a/ocaml/testsuite/tests/typing-layouts/mixed_constructor_arguments.ml
+++ b/ocaml/testsuite/tests/typing-layouts/mixed_constructor_arguments.ml
@@ -498,3 +498,4 @@ Line 1, characters 31-41:
 Error: Type float# has layout float64.
        Inlined records may not yet contain types of this layout.
 |}]
+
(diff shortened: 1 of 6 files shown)


ccasin (approved):
Approving for `aa86aaa`.  I am a bit confused why the other commit shows up when I look at the diff here, but I'm going to assume some github thing.

As an aside, I'm not sure this is really necessary - the fact that compilation for void will need lots of fixes is not unique to mixed blocks. But it doesn't do any harm to make sure we double check the mixed block cases when we get there.

ncik-roberts:
OK. How would you feel about closing this PR instead? In the context you said, I don't think it adds any extra safety. (Maybe I should change the PR to just add tests instead?)

ccasin:
I think it's fine to merge.  At a minimum it will remind us to think about this case.

ccasin:
I merged this without noticing it wasn't into main.  Oops!  But it's fine, I'm reviewing the other PR too, and this is small.

PR #2532, opened 2024-05-03
Middle end/backend/runtime changes for or_null

mshinwell:
To be described later; this is mainly here so @dkalinichenko-js can work from it.  The `Lambda` interface should be about right.

### DIFF
--- a/ocaml/typing/typeopt.ml
+++ b/ocaml/typing/typeopt.ml
@@ -737,6 +737,8 @@ let rec layout_union l1 l2 =
   match l1, l2 with
   | Pbottom, l
   | l, Pbottom -> l
+  | Pnullable_value value_kind1, Pnullable_value value_kind2 ->
+      Pnullable_value (value_kind_union value_kind1 value_kind2)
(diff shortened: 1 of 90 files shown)


mshinwell:
@dkalinichenko-js I've just added some more commentary in `lambda.mli` which will hopefully be of use.

ocaml/typing/typeopt.ml
```diff
@@ -737,6 +737,8 @@ let rec layout_union l1 l2 =
   match l1, l2 with
   | Pbottom, l
   | l, Pbottom -> l
+  | Pnullable_value value_kind1, Pnullable_value value_kind2 ->
+      Pnullable_value (value_kind_union value_kind1 value_kind2)
```
Ekdohibs, line 741: Shouldn't we also want a join between `Pnullable_value` and `Pvalue` to return something better than `Ptop`? This is necessary if the layouts are not incompatible.

mshinwell:
Superceded by newer PRs

PR #2535, opened 2024-05-06
Bugfix for a statmemprof test

stedolan:
Bytecode keeps named values alive until the end of their scope rather just to their last use. This fact nondeterministically breaks one statmemprof test, which relies on a value being collected after its last use.

Inlining the definition (which is used only once) instead of naming it makes both behave the same.


ncik-roberts (approved):
Thanks for teaching me a new interesting fact about bytecode. The fix looks good to me.

ncik-roberts:
I see this test failed again in a recent CI run: https://github.com/ocaml-flambda/flambda-backend/actions/runs/9080247003/job/24951235898. (Also in runtime 4, also in bytecode, also in the debug runtime.) So perhaps there is more going on here.

PR #2538, opened 2024-05-06
Make jkind and types mutually recursive

liam923:
This is work that is necessary in order to add `layout_of` to jkinds. Since `layout_of` must hold a `type_expr`, the types defined in the `Jkind` and `Types` modules will become mutually recursive. To accomplish this, the types defined in `Jkind` have been pulled into a separate module named `Jkind_types`. They are polymorphic, and in `Jkind` they are redefined, with the type variable instantiated with the type `type_expr`. Thus, `Jkind_types` has no dependencies, `Types` depends only on `Jkind_types`, and `Jkind` depends on `Jkind_types` and `Types`.


ocaml/typing/types.ml
```diff
@@ -738,7 +751,7 @@ let log_change ch =
 
 let () =
   Mode.set_append_changes (fun changes -> log_change (Cmodes !changes));
-  Jkind.Sort.change_log := (fun change -> log_change (Csort change))
+  !jkind_sort_change_log := (fun change -> log_change (Csort change))
```
goldfirere, line 754: This is pretty convoluted -- and I think it's broken.

Convoluted: we have a ref to a ref, with one set in `Jkind.Sort` and then the other one set here. Basically, I think the old setup was because Types depended on Jkind, and now that's been flipped -- but instead of undoing the old treatment, now there's another step.

Broken: I think that `jkind_sort_change_log` won't be set by `Jkind.Sort` before this assignment here happens, meaning that the assignment to `jkind_sort_change_log` in `Jkind.Sort` effectively overwrites this change, and it makes `Jkind.Sort.log_change` into a no-op. I'm a little surprised (but not very) that this doesn't show up in tests.

The fix, I think, is to define the (very simple) sort-related change-log functions here in types, breaking the recursion. Maybe there's another way, but I don't see an easy one that doesn't export the `log_change` function immediately above.

ocaml/typing/types.mli
```diff
@@ -221,6 +217,26 @@ and abbrev_memo =
...
+    jkind.ml should fill this ref with the definition of Jkind.Sort.change_log *)
+val jkind_sort_change_log : (Jkind_types.Sort.change -> unit) ref ref
+
+(** INTERNAL USE ONLY
+    jkind.ml should fill this ref with the definition of Jkind.Sort.undo_change *)
+val jkind_sort_undo_change : (Jkind_types.Sort.change -> unit) ref
+
+(** INTERNAL USE ONLY
+    jkind.ml should fill this ref with the definition of Jkind.equal *)
+val jkind_equal : (jkind -> jkind -> bool) ref
```
goldfirere, line 238: I think, given the comments I've written in types.ml, that this should be the one place a ref like this is needed.

ocaml/typing/primitive.mli
```diff
@@ -136,3 +136,10 @@ type error =
   | Invalid_native_repr_for_primitive of string
 
 exception Error of Location.t * error
+
+(* jkind depends on types defined in this file, but Jkind.Sort.equal_const is required
+   to implement equal_native_repr. When jkind.ml is loaded, it fills this ref with
+   the definition of Jkind.Sort.equal_const *)
+(** INTERNAL USE ONLY
+    jkind.ml should fill this ref with the definition of Jkind.Sort.equal_const *)
+val jkind_sort_equal_const : (Jkind_types.Sort.const -> Jkind_types.Sort.const -> bool) ref
```
goldfirere, line 145: There's not a need to play this game with sorts, because sorts don't depend on types. Just define the needed function in jkind_types.ml.

ocaml/typing/primitive.mli
```diff
@@ -27,7 +27,7 @@ type boxed_vector = Pvec128 of vec128_type
    of a primitive *)
 type native_repr =
   | Repr_poly
-  | Same_as_ocaml_repr of Jkind.Sort.const
+  | Same_as_ocaml_repr of Jkind_types.Sort.const
```
goldfirere, line 30: But the comment in jkind_types.mli says not to depend on it. I understand why you did, here, but maybe that comment should be updated to list the modules that sit in the dependency chain between jkind.ml and jkind_types.ml.

goldfirere (commented):
I've suggested a few changes. Please flip through the changes before actioning any of them, because I suggest killing off jkind_types.ml and then later found a need to keep it around (I think).

ocaml/typing/jkind_types.mli
```diff
@@ -0,0 +1,224 @@
...
+        lhs_jkind : 'type_expr Jkind_desc.t;
+        lhs_history : 'type_expr history;
+        rhs_jkind : 'type_expr Jkind_desc.t;
+        rhs_history : 'type_expr history
+      }
+  | Creation of creation_reason
+
+type 'type_expr t =
+  { jkind : 'type_expr Jkind_desc.t;
+    history : 'type_expr history
```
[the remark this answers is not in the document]
liam923: I remember why I didn't do this now - the consequence of this is types like `Types.type_expr`, etc needing to also be polymorphic over `history`. Although I agree the duplication is a pain-point, I think it might be the lesser of whatever other evil we do to work around this.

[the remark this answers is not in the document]
goldfirere: Blargh.

Yes I agree with you.

I couldn't take the duplication. So I found another way to get rid of it; see commit on the way (which I would appreciate review on).

goldfirere (approved):
I'm happy with this now. @liam923 please review my two commits, and then we can merge.

ocaml/typing/jkind.mli
```diff
@@ -70,7 +70,9 @@ end
     [Any] and subjkinds of other sorts, such as [Immediate]. *)
 type t = Types.type_expr Jkind_types.t
 
-include module type of Jkind_intf.History
+include module type of struct
+  include Jkind_intf.History
+end
```
liam923, line 75: Is there a semantice difference between
```
include module type of Jkind_intf.History
```
and
```
include module type of struct
  include Jkind_intf.History
end
```
?

goldfirere: Sadly, yes.

```
# module M = struct type t end;;
module M : sig type t end
# module type S1 = module type of M;;
module type S1 = sig type t end
# module type S2 = module type of struct include M end;;
module type S2 = sig type t = M.t end
```

Note that `S2` remembers that its type `t` is the same as `M.t`, necessary for our use case.

liam923:
Thanks for these changes, I forgot about the `_intf` pattern. This looks good to me. I'll resolve these conflicts.

goldfirere:
It's possible this fix won't work. Maybe yours did? It looks like it wouldn't, though, because the new jkind_intf.ml file is ml-only. I've spent a little while trying to make it work. I've now timed out and need to move on to other tasks.

My recommendation if this doesn't work is to split the jkind_intf file into two mli-only files (supported in our build system): one for the sort stuff and one for the history stuff. The history has to be separate because we need it all wrapped in its own module for inclusion. Annoying.

liam923:
> It's possible this fix won't work. Maybe yours did?

In what sense are you concerned about this not working - build errors or runtime bugs/errors? I'm interpreting your comment as build-wise. Everything seems to build fine for me and the checks are passing, so I think we're okay.

If you agree, feel free to hit merge.

goldfirere:
I must have gotten myself thinking wrongly about the build somehow. I've reverted some complexification I added. If this builds, I will merge.

PR #2539, opened 2024-05-06
Emit stack checks for probes

xclerc:
This pull request add the emission of
stack checks in probes. Since a probe
will save on the stack all the hardware
registers that happen to be live, it
seems possible to have to push more
data than `stack_threshold_size` allows.


backend/amd64/emit.mlp
```diff
@@ -473,28 +473,41 @@ type stack_realloc = {
...
+       cfi_adjust_cfa_offset 8;
+       (* measured in words *)
+       emit_call (Cmm.global_symbol "caml_call_realloc_stack");
+       I.add (int 8) rsp;
+       cfi_adjust_cfa_offset (-8);
+       I.jmp (label sc_return))
+    !stack_realloc
+
+let emit_stack_check ~size_in_bytes ~save_registers =
+    let overflow = new_label () and ret = new_label () in
```
mshinwell, line 498: This needs de-indenting

backend/amd64/emit.mlp
```diff
@@ -473,28 +473,41 @@ type stack_realloc = {
...
+    if save_registers then I.push r10;
+    I.lea (mem64 NONE (-(size_in_bytes + threshold_offset)) RSP) r10;
+    I.cmp (domain_field Domainstate.Domain_current_stack) r10;
+    if save_registers then I.pop r10;
+    I.jb (label overflow);
+    def_label ret;
+    stack_realloc := {
+      sc_label = overflow;
+      sc_return = ret;
+      sc_max_frame_size_in_bytes = size_in_bytes;
```
mshinwell, line 509: We should probably rename `sc_max_frame_size_in_bytes`.  Could you also double-check that there is no assumption in any user of that field that the size represents the amount needed for the whole function, rather than for an individual check?

mshinwell (approved):

PR #2540, opened 2024-05-06
Support `stack_ exp` syntax

riaqn:
This adds the support for `stack_ exp` syntax, which forces the allocation `exp` to be on stack. If `exp` is not an allocation, an error is issued.

The error message has room for improvement: currently no hints are given if the allocation cannot be on stack. Some solutions:
- Inside the typedtree of `exp`, annotate allocation mode with reasons (inherited from `expected_mode`). This is my preferred approach.
- Add extra field `alloc_stack:bool` to `expected_mode`, and `register_allocation` will read that field and force stack allocation. The function works on `expected_mode` so has the hints . This sounds bad idea.

But maybe it's fine without the hints? Let's see what the reviewer says.

Also the docs in `ocaml/jane/doc` need an overhaul. To begin with, we should distinguish `local_` from "stack allocation". That sounds like a big task that I will do in a future PR.

### DIFF
--- a/ocaml/parsing/parser.mly
+++ b/ocaml/parsing/parser.mly
@@ -1042,6 +1042,7 @@ let unboxed_type sloc lident tys =
 %token HASH_SUFFIX            "# "
 %token <string> HASHOP        "##" (* just an example *)
 %token SIG                    "sig"
+%token STACK                  "stack_"
 %token STAR                   "*"
 %token <string * Location.t * string option>
        STRING                 "\"hello\"" (* just an example *)
@@ -1133,7 +1134,7 @@ The precedences must be listed from low to high.
 %nonassoc BACKQUOTE BANG BEGIN CHAR FALSE FLOAT HASH_FLOAT INT HASH_INT OBJECT
           LBRACE LBRACELESS LBRACKET LBRACKETBAR LBRACKETCOLON LIDENT LPAREN
           NEW PREFIXOP STRING TRUE UIDENT
-          LBRACKETPERCENT QUOTED_STRING_EXPR
+          LBRACKETPERCENT QUOTED_STRING_EXPR STACK
 
 
 /* Entry points */
@@ -2811,6 +2812,8 @@ fun_expr:
 %inline expr_:
   | simple_expr nonempty_llist(labeled_simple_expr)
       { mkexp ~loc:$sloc (Pexp_apply($1, $2)) }
+  | STACK simple_expr
+      { mkexp ~loc:$sloc (Pexp_stack $2) }
   | labeled_tuple %prec below_COMMA
       { pexp_ltuple $sloc $1 }
   | mkrhs(constr_longident) simple_expr %prec below_HASH
--- a/ocaml/parsing/printast.ml
+++ b/ocaml/parsing/printast.ml
@@ -384,6 +384,9 @@ and expression i ppf x =
       payload i ppf arg
   | Pexp_unreachable ->
       line i ppf "Pexp_unreachable"
+  | Pexp_stack e ->
+      line i ppf "Pexp_stack\n";
+      expression i ppf e
 
 and value_description i ppf x =
   line i ppf "value_description %a %a\n" fmt_string_loc
--- a/ocaml/testsuite/tests/typing-modes/stack.ml
+++ b/ocaml/testsuite/tests/typing-modes/stack.ml
@@ -117,3 +117,37 @@ let f () = exclave_ stack_ (3, 5)
 [%%expect{|
 val f : unit -> local_ int * int = <fun>
 |}]
+
+let f () =
+    let g = stack_ (fun x -> x) in
+    g 42
+[%%expect{|
+Line 3, characters 4-5:
+3 |     g 42
+        ^
+Error: This value escapes its region.
+  Hint: This function cannot be local,
+  because it is the function in a tail call.
+|}]
+
+let f () =
+    (stack_ (fun x -> x)) 42
+[%%expect{|
+Line 2, characters 12-24:
+2 |     (stack_ (fun x -> x)) 42
+                ^^^^^^^^^^^^
+Error: This allocation cannot be on the stack.
+  Hint: This function cannot be stack-allocated,
+  because it is the function in a tail call.
+|}]
+
+let f () =
+    List.length (stack_ [1; 2; 3])
+[%%expect{|
+Line 2, characters 24-33:
+2 |     List.length (stack_ [1; 2; 3])
+                            ^^^^^^^^^
+Error: This allocation cannot be on the stack.
+  Hint: This argument cannot be stack-allocated,
+  because it is an argument in a tail call.
+|}]
--- a/ocaml/testsuite/tests/typing-modes/stack.ml
+++ b/ocaml/testsuite/tests/typing-modes/stack.ml
@@ -252,3 +252,41 @@ Line 4, characters 17-31:
                      ^^^^^^^^^^^^^^
 Error: Stack allocating modules is unsupported yet.
 |}]
+
+(* stack_ works shallowly *)
+let f () =
+  let r = ref "hello" in
+  let _ = stack_ (r.contents, r.contents) in
+  r.contents
+[%%expect{|
+val f : unit -> string = <fun>
+|}]
+
+let f () =
+  let r = "hello" in
+  let _ = stack_ (r, r) in
+  r
+[%%expect{|
+val f : unit -> string = <fun>
+|}]
+
+type t = { x : int list; y : int list @@ global }
+
+let mk () =
+  let r = stack_ { x = [1;2;3]; y = [4;5;6] } in
+  r.y
+[%%expect{|
+type t = { x : int list; global_ y : int list; }
+val mk : unit -> int list = <fun>
+|}]
+
+let mk () =
+  let r = stack_ { x = [1;2;3]; y = [4;5;6] } in
+  r.x
+[%%expect{|
+Line 3, characters 2-5:
+3 |   r.x
+      ^^^
+Error: This value escapes its region.
+  Hint: Cannot return a local value without an "exclave_" annotation.
+|}]
--- a/ocaml/testsuite/tests/typing-modes/stack.ml
+++ b/ocaml/testsuite/tests/typing-modes/stack.ml
@@ -1,0 +1,103 @@
+(* TEST
+expect;
+*)
+
+let f = ref (stack_ (fun x -> x))
+[%%expect{|
+Line 1, characters 20-32:
+1 | let f = ref (stack_ (fun x -> x))
+                        ^^^^^^^^^^^^
+Error: This allocation cannot be on stack.
+|}]
+
+let f = ref (stack_ (2, 3))
+[%%expect{|
+Line 1, characters 20-26:
+1 | let f = ref (stack_ (2, 3))
+                        ^^^^^^
+Error: This allocation cannot be on stack.
+|}]
+
+type t = Foo | Bar of int
+
+let f = ref (stack_ Foo)
+[%%expect{|
+type t = Foo | Bar of int
+Line 3, characters 20-23:
+3 | let f = ref (stack_ Foo)
+                        ^^^
+Error: This expression is not an allocation.
+|}]
+
+let f = ref (stack_ (Bar 42))
+[%%expect{|
+Line 1, characters 20-28:
+1 | let f = ref (stack_ (Bar 42))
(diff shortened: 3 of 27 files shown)


ocaml/testsuite/tests/typing-modes/stack.ml
```diff
@@ -0,0 +1,103 @@
+(* TEST
```
goldfirere, line 1: I would love to see some more tests. Here are some ideas:

* Each of these cases where `stack_` is rejected should have a similar case nearby where `stack_` is accepted. Otherwise, we'll never know if `stack_` starts rejecting more things.
* There should be cases testing that `stack_` works deeply. That is, my understanding of the specification of `stack_` is that e.g. `stack_ (K1 (K2 (K3 x)))` is meant to work deeply and should mean that all of `K1`, `K2`, and `K3` are allocated on the stack.... which might not be true if one of them has a global modality. So I would want to see tests both with and without intervening global modalities.
* Is there a way to assert `stack_`-allocation on patterns that allocate? Probably not, and that's probably OK for now.
* I would want to see some tests on lists, because list literals actually encompass many allocations.
* Tests on polymorphic-variant constructors?

As for the lack of reasons/hints: I think this feature would be better with them. Stack-allocation is an advanced feature, and helping users learn it is a good thing. I don't imagine it would be too hard to add, along the lines you suggest in the PR description.

ocaml/typing/typecore.ml
```diff
@@ -6429,6 +6431,25 @@ and type_expect_
            exp_type = instance ty_expected;
            exp_attributes = sexp.pexp_attributes;
            exp_env = env }
+  | Pexp_stack e ->
+      let exp = type_expect env expected_mode e ty_expected_explained in
+      begin match exp.exp_desc with
+      | Texp_function { alloc_mode; _} | Texp_tuple (_, alloc_mode)
+      | Texp_construct (_, _, _, Some alloc_mode)
+      | Texp_variant (_, Some (_, alloc_mode))
```
goldfirere, line 6439: Good: this is here! So there should be a test.

ocaml/typing/typecore.ml
```diff
@@ -6429,6 +6431,25 @@ and type_expect_
...
+      | Texp_record {alloc_mode = Some alloc_mode; _}
+      | Texp_array (_, _, _, alloc_mode)
+      | Texp_field (_, _, _, Boxing (alloc_mode, _)) ->
+        begin match Locality.submode Locality.local
+          (Alloc.proj (Comonadic Areality) alloc_mode) with
+        | Ok () -> ()
+        | Error _ -> raise (Error (exp.exp_loc, env, Cannot_stack_allocate))
+        end
+      | _ ->
```
goldfirere, line 6448: A bunch of these cases *are* allocations -- they're just not allocations that can go on the stack. For example, `new cl` or `lazy`. Might be better to separate out allocations from non-allocations in the error message, so as not to confuse users.

Also, what about first-class modules? Can they be stack-allocated?

ocaml/parsing/parser.mly
```diff
@@ -2811,6 +2812,8 @@ fun_expr:
 %inline expr_:
   | simple_expr nonempty_llist(labeled_simple_expr)
       { mkexp ~loc:$sloc (Pexp_apply($1, $2)) }
+  | STACK simple_expr
```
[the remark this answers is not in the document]
riaqn: Yes I think we want to give the impression that `K x` is the argument to `stack_`. Whereas in `exclave_`, it's more like sequential execution? I.e., `exclave_` deletes the region, and proceeds to whatever follows.

ocaml/testsuite/tests/typing-modes/stack.ml
```diff
@@ -0,0 +1,103 @@
...
+
+type t = Foo | Bar of int
+
+let f = ref (stack_ Foo)
+[%%expect{|
+type t = Foo | Bar of int
+Line 3, characters 20-23:
+3 | let f = ref (stack_ Foo)
+                        ^^^
+Error: This expression is not an allocation.
```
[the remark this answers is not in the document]
riaqn: It's very vague to say "an expression allocates". For example, the expression `foo (Some y)` does allocate.

We want to emphasize that the expression itself *is* an allocation.

ocaml/parsing/printast.ml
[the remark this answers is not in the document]
riaqn: Wildcard pattern strikes again.

ocaml/testsuite/tests/typing-modes/stack.ml
```diff
@@ -0,0 +1,103 @@
+(* TEST
+expect;
+*)
+
+let f = ref (stack_ (fun x -> x))
+[%%expect{|
+Line 1, characters 20-32:
+1 | let f = ref (stack_ (fun x -> x))
```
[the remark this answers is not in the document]
riaqn: Not sure what you mean - surely this `fun x -> x` is an allocation? Or do you mean it's hoisted?

ocaml/typing/typecore.ml
riaqn: Yes good point. Two solutions:
- For those (`new cl` or `lazy`) we will print a different message along the line of "stack-allocating this langauge construct is not supported yet"
- Or we rephrase `Not_allocation` to "This language construct cannot be stack-allocated", to cover both `new cl` and genuine non-allocation. 

I prefer the second one for simplicity. What do you think?

riaqn: first-class modules are similar to `new cl` and `lazy` - they are always heap-allocated.

ocaml/testsuite/tests/typing-modes/stack.ml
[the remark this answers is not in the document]
riaqn: Like, should we expose to the user the fact that `fun x -> x` is not an allocation because of optimizations? I think not.

Also, to detect this in the type checker is difficult - will need to defer to middle-end. But middle-end printing error messages sounds bad.

ocaml/testsuite/tests/typing-modes/stack.ml
riaqn: > There should be cases testing that stack_ works deeply.

`stack_` works shallowly. For example:
```
let _ = ref x in
stack_ (K x)
```
will be accepted. `x` is on heap and `K x` is on stack.

riaqn: > Is there a way to assert stack_-allocation on patterns that allocate?

Not in this PR; but I don't see obvious difficulty. But I think we want to support modes on patterns first.

riaqn: > Tests on polymorphic-variant constructors?

I'm curious about your interest in this - I thought they are similar to regular constructors so I skipped.

riaqn: > I would want to see some tests on lists, because list literals actually encompass many allocations.

This wouldn't be super interesting for the reason I listed above - `stack_` is not deep.

ocaml/testsuite/tests/typing-modes/stack.ml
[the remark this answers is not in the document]
goldfirere: Good point. Then maybe
```suggestion
Error: This expression is not an allocation site.
```

My problem was really that I find the original phrasing to be a type error. To me, an expression is not an allocation — it’s a site for allocation, or perhaps it causes an allocation. I agree with you that many expressions allocate, including some that are not suitable with `stack_`. But maybe “allocation site” will satisfy both of us.

ocaml/testsuite/tests/typing-modes/stack.ml
goldfirere: My understanding is that `stack_` would be deep on nested constructor calls. That is, `stack_ (Some (Some thing))` would ensure that both `Some`s are stack-allocated. Contrast with `stack_ G (G thing)` where we have `type ‘a t = G of ‘a @@ global`; the inner `G` would not be stack allocated.

But maybe my understanding is wrong? Might be good to raise this design issue on the appropriate Slack channel, because we want to agree on what would be most useful to users.

goldfirere: My only interest in polymorphic variants is that there is a specific line of code to make `stack_` work on polymorphic variants, so I thought it would probably be good to test that it works.

ocaml/typing/typecore.ml
goldfirere: How about “This expression cannot be a stack-allocation site” ?

ocaml/testsuite/tests/typing-modes/stack.ml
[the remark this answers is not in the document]
goldfirere: Yes, good point. The constant-check is in transl, but I misremembered thinking it was in the type-checker. And the check I’m thinking of is conservative, because it happens before inlining. So I’ll just resolve.

ocaml/testsuite/tests/typing-modes/stack.ml
goldfirere: Also, we shouldn’t forget about tests that do not produce error messages.

ocaml/typing/typecore.ml
lpw25: Please treat the cases where we do not (yet?) support stack allocation separately from the expressions that are not allocations, and give them specific error messages too, for example:
```
Error: Objects cannot be allocated on the stack
```
You just need a variant type that represents all the things that can't be put on the stack, and it's much clearer to the user.

ocaml/testsuite/tests/typing-modes/stack.ml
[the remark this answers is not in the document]
riaqn: I like "allocation site". This is how ocaml programmers talk, I think.

ocaml/testsuite/tests/typing-modes/stack.ml
riaqn: Yeah I can intuitively see how deep `stack_` can be useful. But in your second example, the `stack_` interact with modality, which I don't like. Because `stack_` operates on the axis `stack/heap` (which currently collides with `local/global`), while the `global` modality operates on the regionality axis.

Instead, we say: `stack_` works deeply no matter what. That means the user has to write `let x = G thing in stack_ (G x)`.

But this also means allowing user to write `stack_ (foo bar)` that doesn't do anything. Maybe we should require "each stack_ should at least do something, or we issue a warning"?

If the above doesn't deter you, let's bring this up to a wider audience.

goldfirere: I thought it would be deep only over nested constructor calls, not passing inside of function application. To be honest, I don’t have a strong sense for what would be best here — I’m just reporting on what I thought was the agreed plan (but I don’t have a reference to where we made that agreement; it’s just what I remember).

And I realize, upon rereading, that my previous post about the global modality was confusing. I mean that `stack_ (G (G thing))` would report an error, while `stack_ (Some (Some thing))` would not report an error. This allows users to witness the global modality (which they may have forgotten).

Maybe the best thing would be for this patch to include a change to the documentation which describes how the feature works. Then we’ll have a specification of what it means and we can then seek wider opinions on whether that specification is the best one.

riaqn: good point, will do.

riaqn: I think I'm gonna skip changing the document, because currently the local allocation document strongly associates `local_` with stack allocation. Adding a section about `stack_` without changing the rest is not coherent.

I suggest that we merge this PR without telling the user. We will find time to overhaul the local allocation document, and then we can announce `stack_` to the user.

goldfirere: Yet I still don't know that I crisply understand the specification of this feature, if there is no documentation. The tests can get at this specification, but no test really explores the "deep" aspect of this that I've been asking about in this thread. For example:

```
type t = { x : int list; y : int list @@ global }
let mk () = exclave_ stack_ { x = [1;2;3]; y = [4;5;6] }
```

I don't know whether this should be accepted or not.

riaqn: That's fair. I will add some tests demonstrating this aspect.

riaqn: Your example will be shadowed by much simpler tests, tho. Because `stack_` works shallowly, so the modalities on fields doesn't matter.

goldfirere (approved):

PR #2541, opened 2024-05-06
Upgrade float32 support to beta

TheNumbat:
Upgrades the maturity of the `small_numbers` extension to beta and moves the float32 libraries from alpha to beta.


TheNumbat:
Fortunately, we've now disabled the upstream build CI : )
(The error is some problem with the upstream-built top level trying to use the float32 printer without the compiler knowing about the float32 type.)

PR #2543, opened 2024-05-07
Zero alloc: propagate assume on partial applications

gretay-js:
This PR proposes to propagate "assume" annotation from a partial application to application to extra arguments.

This came up in a discussion of https://github.com/ocaml-flambda/flambda-backend/pull/2165 (see example below).
It came up again recently in #2459 where  a new test has different analysis results depending on compiler configuration with or without stack allocation, even though the generated code is the same and there is no stack allocation. 

The fix is easy but I am not sure it's the right thing to do.

In the examples below, the generated code for functions `g0-g3` is the same. 
The check pass on `g0` but fails on `g1` below, as expected. 
Should the check fail or pass on `g2` and `g3`?
```
let[@inline never][@local never] f x y = (x,y)

let[@zero_alloc] g0 () =
  let x = (f[@zero_alloc assume]) () () in x

let[@zero_alloc] g1 () =
  let x = ((f[@zero_alloc assume]) ()) () in x

let[@zero_alloc] g2 () =
  let x = ((f ())[@zero_alloc assume]) () in x

let[@zero_alloc] g3 () =
  let x = (((f[@zero_alloc assume]) ())[@zero_alloc assume]) () in x
```

This PR updates the tests for `g2` and `g3` to pass and removes a special case for heap allocation in a test recenlty added by #2459.


gretay-js:
Propagating `assume` in `g2` is unsound because the user didn't say that the first application is `zero_alloc`, so we shouldn't merge this PR.  We can try to address the problem observed in #2459 in some other way. It seems to be a special case of primitives that return `Lapply` and it seems to only happen for `apply` and `revapply`.

PR #2545, opened 2024-05-07
Fix for missing stub inlining in classic mode

mshinwell:
The basic stub inlining strategy in classic mode doesn't match up with Simplify and I'm not sure why.  This PR aims to correct it.

The current situation was seen to be producing confusing results as a result of #2357.  A functor that was not previously inlinable, as it contained functions, became inlinable.  However there was a functor coercion stub (because of the ordering of declarations in the `.mli` file) and so the functor itself was not directly exposed in the interface.  Before the aforementioned PR, the stub just contained a call to the original functor; but after the PR, this call was inlined.  As a result, the stub went way over the inlining threshold, and despite being a stub was not written into the `.cmx` file.  Fixing this is the first part of this PR.

At which point, even if the inlining threshold is raised, the stub is blackholed because of the code that looks at approximations upon `.cmx` loading.  Any stub would be forced to metadata-only, thus preventing inlining.  So this is the second part of the fix.

I am unclear as to why either of these pieces of code were in the state prior to this PR, but maybe there is something I am missing here.  I don't think there should be any performance concerns since we're not descending into the bodies of inlined functions unlike in Simplify.

I have wondered on and off, after intermittent user reports of problems with optional argument wrappers, whether there might be a lingering problem with stub inlining - and this could very well be it.

@lthls could you please look at this asap?

@Keryan-dev do you happen to remember any previous relevant discussions?

### DIFF
--- a/middle_end/flambda2/from_lambda/closure_conversion.ml
+++ b/middle_end/flambda2/from_lambda/closure_conversion.ml
@@ -2186,9 +2186,7 @@ let close_one_function acc ~code_id ~external_env ~by_function_slot
   in
   let inlining_decision =
     if Flambda_features.classic_mode ()
-    then Inlining.definition_inlining_decision inline cost_metrics
-    else if stub
-    then Function_decl_inlining_decision_type.Stub
+    then Inlining.definition_inlining_decision inline cost_metrics ~stub
     else Function_decl_inlining_decision_type.Not_yet_decided
   in
   let loopify : Loopify_attribute.t =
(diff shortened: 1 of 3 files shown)


lthls:
I think these weird behaviours are here because we tried to mimic Closure and its heuristics as much as possible.
Since we've decided that we don't need to stick to that, I think this PR is good.

I am a bit concerned about inlining into stubs though. Closure did it so Classic mode does it, and this PR doesn't prevent it, so we can end up with big functions marked as stubs (as in your example with a functor) which can throw off Simplify's heuristics. Before this PR it wouldn't matter as Simplify would never see the code of these stubs, but now we might end up doing more inlining than we should.

lthls (approved):
I think we can merge even if we're not completely sure of the impact on Simplify's inlining decisions; I don't expect issues in practice for real code.

mshinwell:
Example from @tylerhou  where inlining happens inside a partial application stub in classic mode but not with Simplify:
```
let[@inline never] add a b = a + b
let () = ignore add
let foo k = add 1
let calls_foo k = foo k k
```

PR #2548, opened 2024-05-08
Produce Invalid upon mode mismatch in partial application

mshinwell:
It seems that a long-standing check (which was also present in Flambda 1) is too strong during partial application.  Dead GADT match cases can produce a mode mismatch that is currently causing a fatal error, but seems like it should produce `Invalid` instead.  There is an example in this PR.

This diff should be read with patdiff.

### DIFF
--- a/middle_end/flambda2/terms/flambda.mli
+++ b/middle_end/flambda2/terms/flambda.mli
@@ -106,6 +106,7 @@ module Invalid : sig
     | Apply_cont_of_unreachable_continuation of Continuation.t
     | Defining_expr_of_let of Bound_pattern.t * named
     | Closure_type_was_invalid of Apply_expr.t
+    | Closure_type_was_invalid_classic_mode of Debuginfo.t
     | Calling_local_returning_closure_with_normal_apply of Apply_expr.t
     | Zero_switch_arms
     | Code_not_rebuilt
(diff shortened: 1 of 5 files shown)


middle_end/flambda2/terms/flambda.mli
```diff
@@ -106,6 +106,7 @@ module Invalid : sig
     | Apply_cont_of_unreachable_continuation of Continuation.t
     | Defining_expr_of_let of Bound_pattern.t * named
     | Closure_type_was_invalid of Apply_expr.t
+    | Closure_type_was_invalid_classic_mode of Debuginfo.t
```
lukemaurer, line 109: Could we call this `Closure_type_was_invalid_in_lambda` or some such? This happens equally well with classic mode on or off, I imagine?

mshinwell: No, this one is only used in classic mode.  I would have reused `Closure_type_was_invalid` but we don't have an `Apply_expr.t` to hand.

mshinwell: Ah no, as discussed this is actually used in non-classic mode as well.  I've renamed this constructor, and also added another one, to distinguish (when using Simplify) this case from other cases where closure types are invalid.

lukemaurer (approved):

PR #2554, opened 2024-05-09
Arrays of unboxed float32s

TheNumbat:
Adds support for `float32# array`. They are structured identically to `int32# array`s, except that they are constructed from float32s.
- Dynamically allocating a float32 array pairs the elements using a new cmm instruction `Cpackf32`, which is the equivalent of int32*int32->int64 packing for floats.
- Statically allocated float32 arrays with static data use `Csingle` data items, which are 32 bits wide. Dynamic data uses `make_update` with `naked_float32s`, which packs the updates tightly.
- The final 32 bits of odd length arrays are not initialized. When the array is statically allocated, this means they will be zero.

The tests for `int32# array` have been replicated for `float32# array`; see `typing-layouts-arrays`.


mshinwell:
I will read the middle end and backend parts, hopefully on 20th May.

mshinwell:
actually I'll just read all of this.

TheNumbat:
fyi Chris has already read the (minimal) frontend bits

mshinwell (approved):

PR #2555, opened 2024-05-09
Move unboxed definitions from `Stdlib` to `Stable`

dkalinichenko-js:
Move `Float_u`, `Int32_u`, `Int64_u` and `Nativeint_u` to `Stable` from `Stdlib`. Remove special build rules used for those modules.

Fix various tests to work with the new library layout. This takes up most of the PR. Those changes are pretty mechanical: using `Stable.<module>` instead of `Stdlib__<module>`, including the `stable` otherlib or dynamically loading it in expect tests.

Also, fix importing otherlibs in expect tests. To use the version of standard library in `_runtest`, the expect test runner was enabling `-nostdlib` and opening the required version separately. This approach breaks path resolution for otherlibs imports like `+stable`. Set `OCAMLLIB` to point to the right `stdlib` version instead.


ocaml/testsuite/tests/typing-layouts-float64/basics_beta.ml
```diff
@@ -22,6 +22,11 @@
    convenient example of a concrete [float64] type in some tests, but its
    behavior isn't the primary purpose of this test. *)
 
+#directory "+stable"
+#load "stable.cma"
+[%%expect {|
+|}];;
```
ncik-roberts, line 28: It would be nice if this was done by `include stable;`. Not sure how feasible that is, since ocamltest can be unwieldy.

ocaml/testsuite/tests/mixed-blocks/constructor_args.ml
```diff
@@ -11,10 +12,7 @@
 (*****************************************)
 (* Prelude: Functions on unboxed numbers *)
 
-module Float_u = Stdlib__Float_u
-module Int32_u = Stdlib__Int32_u
-module Int64_u = Stdlib__Int64_u
-module Nativeint_u = Stdlib__Nativeint_u
+open Stable
```
ncik-roberts, line 15: I would prefer to explicitly have `module Float_u = Stable.Float_u` etc.

ocaml/testsuite/tests/mixed-blocks/test_mixed_blocks.ml
```diff
@@ -12,20 +13,7 @@
...
-  let ( * ) = mul
-  let ( / ) = div
-  let ( ** ) = pow
-  let ( > ) x y = (compare x y) > 0
-end
-
-module Int32_u = Stdlib__Int32_u
-module Int64_u = Stdlib__Int64_u
-module Nativeint_u = Stdlib__Nativeint_u
+open Stable
```
ncik-roberts, line 16: I would prefer to explicitly have `module Float_u = Stable.Float_u` etc. That way you get more of a suggestion just from scanning the prelude which kinds of mixed blocks are tested in this file.

ocaml/ocamltest/ocaml_actions.ml
```diff
@@ -794,19 +794,22 @@ let cc =
...
     Ocaml_commands.ocamlrun_expect_test;
     expect_flags;
     flags env;
-    repo_root;
     principal_flag;
     input_file
   ] in
+  let environment = Array.concat [
+    default_ocaml_env;
+    [| (Printf.sprintf "OCAMLLIB=%s" stdlib_dir) |]
```
ncik-roberts, line 809: What about this PR makes this necessary?

ncik-roberts (approved):
Seems straightforward and good to me.

ocaml/ocamltest/ocaml_actions.ml
dkalinichenko-js: Otherwise, `#directory +stable` does not work in expect tests, and we can't load the `Stable` module. But I'm now looking into just fixing `include stable` for expect tests.

ocaml/testsuite/tests/typing-layouts-float64/basics_beta.ml
dkalinichenko-js: Done with https://github.com/ocaml-flambda/flambda-backend/pull/2558.

ncik-roberts:
This looks good to merge after we fix up #2558.

PR #2558, opened 2024-05-10
Fix `include` in expect tests

dkalinichenko-js:
Fix `include` in expect tests, and add a test using it.

Pass the same arguments to the test runner as we do for bytecode toplevel tests.
This includes `-nostdlib` and the necessary directory flags, so we can remove the redundant `-repo-root` argument.

Load object files (anonymous args with `.cmo` or `.cma` extensions) in the expect test runner.

Our test suite uses the expect test runner under `testsuite/tools`, but I also synchronized the one under `ocaml/testsuite/tools`. The latter also passed `~auto_include:Load_path.no_auto_include` to `Compmisc.init_path`, but with my changes this should not be necessary.

### DIFF
--- a/ocaml/testsuite/tests/tool-expect-test/include.ml
+++ b/ocaml/testsuite/tests/tool-expect-test/include.ml
@@ -1,0 +1,12 @@
+(* TEST
+ include unix;
+ libunix;
+ expect;
+*)
+
+(* Test that [include] works in expect tests: the Unix module is not
+   available by default. *)
+Unix.sleepf 0.01;;
+[%%expect {|
+- : unit = ()
+|}]
--- a/ocaml/testsuite/tests/tool-expect-test/include.ml
+++ b/ocaml/testsuite/tests/tool-expect-test/include.ml
@@ -1,5 +1,6 @@
 (* TEST
  include unix;
+ hasunix;
  libunix;
  expect;
 *)
--- a/ocaml/testsuite/tools/expect_test.ml
+++ b/ocaml/testsuite/tools/expect_test.ml
@@ -325,7 +325,7 @@ let read_anonymous_arg fname =
   match !main_file with
   | None -> main_file := Some fname
   | Some _ ->
-    Printf.eprintf "expect_test: multiple input files\n";
+    Printf.eprintf "expect_test: multiple input source files\n";
     exit 2
 
 let main fname =
--- a/ocaml/testsuite/tools/expect_test.ml
+++ b/ocaml/testsuite/tools/expect_test.ml
@@ -313,6 +313,7 @@ let process_expect_file fname =
   let correction = eval_expect_file fname ~file_contents in
   write_corrected ~file:corrected_fname ~file_contents correction
 
+let repo_root = ref None
 let keep_original_error_size = ref false
 let preload_objects = ref []
 let main_file = ref None
@@ -321,11 +322,11 @@ let read_anonymous_arg fname =
   if Filename.check_suffix fname ".cmo"
           || Filename.check_suffix fname ".cma"
   then preload_objects := fname :: !preload_objects
-  else
+else
   match !main_file with
   | None -> main_file := Some fname
   | Some _ ->
-    Printf.eprintf "expect_test: multiple input source files\n";
+    Printf.eprintf "expect_test: multiple input files\n";
     exit 2
 
 let main fname =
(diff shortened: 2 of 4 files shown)


ocaml/testsuite/tools/expect_test.ml
```diff
@@ -324,19 +336,18 @@ let main fname =
...
-    match !repo_root with
-    | None -> ()
-    | Some dir ->
-        (* If we pass [-repo-root], use the stdlib from inside the
-           compiler, not the installed one. We use
-           [Compenv.last_include_dirs] to make sure that the stdlib
-           directory is the last one. *)
-        Clflags.no_std_include := true;
-        Compenv.last_include_dirs := [Filename.concat dir "stdlib"]
-  end;
```
ncik-roberts, line 337: I think we should avoid removing this logic. I understand that it becomes dead code if we always pass `-nostdlib` (i.e. `Clflags.no_std_include = true`). But keeping this code around helps us avoid drift from upstream.

ocaml/testsuite/tools/expect_test.ml
```diff
@@ -324,19 +336,18 @@ let main fname =
...
-  Compmisc.init_path ~auto_include:Load_path.no_auto_include ();
```
ncik-roberts, line 338: I think we should avoid removing `auto_include:Load_path.no_auto_include` here. It looks like it was added for good reason in https://github.com/ocaml/ocaml/pull/11298

ocaml/testsuite/tools/expect_test.ml
```diff
@@ -313,8 +313,20 @@ let process_expect_file fname =
...
+let main_file = ref None
+
+let read_anonymous_arg fname =
+  if Filename.check_suffix fname ".cmo"
+          || Filename.check_suffix fname ".cma"
+  then preload_objects := fname :: !preload_objects
+else
+  match !main_file with
+  | None -> main_file := Some fname
+  | Some _ ->
```
ncik-roberts, line 327: It looks like the expect test runner used to be able to handle multiple files. (It ran `main` on each file.) Now it can only handle one. I'd prefer to do the same thing as upstream here. Maybe we can make the `cmo` and `cma` arguments be specified with a different flag, like `-load`, so that we can handle them differently.

ocaml/testsuite/tools/expect_test.ml
```diff
@@ -313,8 +313,20 @@ let process_expect_file fname =
...
+    Printf.eprintf "expect_test: multiple input files\n";
```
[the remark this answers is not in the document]
ncik-roberts: (may be obviated by my other comment about accepting multiple files.)

ncik-roberts (commented):
Many of my comments are about removing drift with upstream. I tested my suggestions and I believe it's the case that the test you added still passes after my accepting my suggestions.

ncik-roberts:
Also, I left my comments on `ocaml/testsuite/tools/expect_test.ml` but I think the copy of the file under `testsuite` is the one that actually gets used to run expect_tests in our branch.

ocaml/testsuite/tools/expect_test.ml
dkalinichenko-js: Did it? `main` calls `exit 0` after processing a file. Seems like before my change the runner just silently ignored the remaining arguments.

ncik-roberts: Ah, OK, I think you're right. It may still be a simplification to require the cma/cmo arguments to be specified using a different flag instead of doing string munging, but I no longer feel as strongly.

dkalinichenko-js: I copied this particular behavior from the OCaml toplevel, it also just checks the file extension.

ncik-roberts: OK

ocaml/testsuite/tests/tool-expect-test/include.ml
```diff
@@ -0,0 +1,12 @@
+(* TEST
+ include unix;
```
ncik-roberts, line 2:
```suggestion
 include unix;
 hasunix;
```
This directive makes sure that tests that require unix only run on unix environments. (We'll probably want this before upstreaming.)

ncik-roberts (approved):
Thanks, this looks good to me now. I've tagged this as "to upstream" as I think it has the potential to be accepted there as well.

PR #2559, opened 2024-05-10
Make Jkind.t abstract

ncik-roberts:
A tiny follow-up to #2538: I observe that we don't have to expose the definition of `Jkind.t` in its signature.


ncik-roberts:
actually, I must have been holding my local build wrong because CI immediately notices that this doesn't build. Sorry for the noise.

PR #2561, opened 2024-05-13
Use Dune 3.15.2

dkalinichenko-js:
Use Dune 3.15.2 in CI, and update docs to indicate that pinning the Dune version is no longer necessary.


ccasin (approved):
LGTM - thanks!

PR #2562, opened 2024-05-13
Consistently use `type t` in unboxed libs

TheNumbat:
Slightly refactors the `Float32`/`Float32_u`/`Int32_u`/`Int64_u`/`Nativeint_u`/`Float_u` interfaces to consistently use their `type t` alias.

### DIFF
--- a/ocaml/otherlibs/beta/float32.ml
+++ b/ocaml/otherlibs/beta/float32.ml
@@ -18,66 +18,53 @@
 
 type t = float32
 
-external float32_of_bits : int32 -> float32
+external float32_of_bits : int32 -> t
   = "caml_float32_of_bits_bytecode" "caml_float32_of_bits"
   [@@unboxed] [@@noalloc]
 
-external neg : (float32[@local_opt]) -> (float32[@local_opt]) = "%negfloat32"
+external neg : (t[@local_opt]) -> (t[@local_opt]) = "%negfloat32"
 
-external add :
-  (float32[@local_opt]) -> (float32[@local_opt]) -> (float32[@local_opt])
+external add : (t[@local_opt]) -> (t[@local_opt]) -> (t[@local_opt])
   = "%addfloat32"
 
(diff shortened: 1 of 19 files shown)


ncik-roberts (commented):
I'm happy with this change, but let's get someone else (@goldfirere or @ccasin) to take a quick look to confirm (or not) my opinion.

The reason why I suggested this change to Max is:
  * we have something like 5 copies of fairly similar interfaces now (`float32_u`, `float_u`, `float32`; `int64_u`, `int32_u`; there will be more) that include dozens of operations
  * the repetition of the fully-spelt-out predef identifier (e.g.) `float32#` in these mlis arguably impedes readability because it makes it harder to spot interesting divergence, e.g. conversions into another type `float32# -> float#`, or even typos.

The downsides of this PR:
  * there is a strong but not completely consistent convention in the stdlib to use the predef name in mlis unless giving the specification of a binding that's used by some functor in the stdlib. E.g. the spec for `compare` is usually given as `t -> t -> int`. This PR deviates from that.
  * some error messages in tests look worse, referring to `Float32_u.t` instead of `float32#`. There is some evidence that this won't be a problem for real error messages, as I've tested using e.g. `Float32_u.compare` in the toplevel and seeing inferred types and error messages use only the `float32#` name. But I don't have an explanation offhand for why the testsuite looks different.

goldfirere:
Without looking at the code, but only @ncik-roberts comment: I think this is a good change. I don't think there's any real reason we should conform to existing conventions in the stdlib. When we're upstreaming, we'll have to have a number of design changes then, and I don't think there's much value in trying to guess these ahead of time. And while `float32#` is maybe a tiny bit better than `Float32_u.t`, they so quickly map to the same concept internally that I doubt I'd notice a difference in practice. I think the upsides outweigh the down.

ncik-roberts (approved):
Let's merge, then.

PR #2563, opened 2024-05-13
Simplify the `install` stanza of extension universe otherlibs

dkalinichenko-js:
Bump Dune language version to 3.11 to use `glob_files` and `with_prefix`. Simplify the `install` stanza of extension universe otherlibs.

This can be extended to all other `dune` files we have later. I'm sticking with just this for now since this makes renaming otherlibs easier.


TheNumbat (approved):
Looks good, thanks

PR #2565, opened 2024-05-14
Optimized version of dominator forest

xclerc:
This pull request adds a new function,
namely `compute_dominator_forest_opt`,
to compute the dominator forest. The
existing version is renamed to
`compute_dominator_forest_naive`.

Instead of re-filtering the whole collection
of CFG blocks every time a node is
added to the forest, the list of children
for each node is computed in one
pass over the CFG blocks. This does
not make a difference most of the
time, but on pathological cases (_i.e._
huge CFG values), the existing
implementation can be impractically
slow.

The pull request also adds code to
checks whether the two functions
compute equivalent forests. It will
be removed in a follow-up.

### DIFF
--- a/backend/cfg/cfg_dominators.ml
+++ b/backend/cfg/cfg_dominators.ml
@@ -423,7 +423,7 @@ let compute_dominator_forest_opt : Cfg.t -> doms -> dominator_tree list =
       | None -> []
       | Some labels -> labels
     in
-    { label; children = List.rev_map children_labels ~f:build_tree }
+    { label; children = List.map children_labels ~f:build_tree }
   in
   Cfg.iter_blocks cfg ~f:(fun label _block ->
       match Label.Tbl.find_opt doms label with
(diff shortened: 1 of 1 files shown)


backend/cfg/cfg_dominators.ml
```diff
@@ -413,12 +413,65 @@ let compute_dominator_forest : Cfg.t -> doms -> dominator_tree list =
...
+        if Label.equal label immediate_dominator
+        then roots := label :: !roots
+        else
+          let current =
+            match Label.Tbl.find_opt children immediate_dominator with
+            | None -> []
+            | Some children -> children
+          in
+          Label.Tbl.replace children immediate_dominator (label :: current));
+  let res = List.rev_map !roots ~f:build_tree in
```
gretay-js, line 441: Where do we rely on the order of roots and children in the tree? Why do the lists need to be reversed in L441  and L426?

xclerc: We don't, that is actually an unfortunate leftover,
due to a try to reproduce an unrelated bug. Thanks!

gretay-js (approved):

PR #2568, opened 2024-05-14
Zero alloc: rename `Checks` to `Zero_alloc_info`

gretay-js:
- Remove `enabled` field of `Checks.t` : the summaries are computed by default and the field is useless for builds that have at least one dependency built with summaries (which is the common case even for non-optimized builds).
- Rename `check.ml` to `zero_alloc_info.ml` as a follow up for #2416.


backend/zero_alloc_info.ml
```diff
@@ -0,0 +1,52 @@
+module String = Misc.Stdlib.String
```
xclerc, line 1: (For history / `git blame`, it would probably
be better to have this as a rename of
`checks.ml` rather than a deletion/addition.)

xclerc (approved):

backend/zero_alloc_info.ml
gretay-js: `git show` on the first commit does show it as a move.

xclerc: That might be a GitHub UI issue then.

PR #2569, opened 2024-05-14
Use IRC for both the compiler and the test suite

xclerc:
(for testing)


xclerc:
(We should do the same for the other allocators.)

TheNumbat:
Merged via #2573

PR #2571, opened 2024-05-14
Avoid CMOV with CFeq

gretay-js:
Improve code generation for `Icsel` when the condition is float equality. This condition is not directly expressible as a CMOV mnemonic. To work around it, the compile emits a jump instruction (to correctly handled the special case of Nan argument), defeating the purpose of `CMOV`.  This PR adds an ad-hoc pattern during `Selection` to detect this case, negate the condition and swap the arguments of `CMOV`.  

Added a test to exercise this special case. I'm not adding any assembly grepping, because it's too fragile.

### DIFF
--- a/backend/amd64/selection.ml
+++ b/backend/amd64/selection.ml
@@ -346,6 +346,19 @@ method! select_operation op args dbg =
       Ispecific Izextend32, [arg]
     | _ -> super#select_operation op args dbg
     end
+  | Ccsel _ ->
+     begin match args with
+     | [cond; ifso; ifnot] ->
+       let (cond, earg) = self#select_condition cond in
+       (match cond with
+        | Ifloattest (w,CFeq) ->
+          (* CFeq cannot be represented as cmov without a jump.
+             Use Cneq and swap the arguments. *)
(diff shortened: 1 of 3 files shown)


backend/amd64/selection.ml
```diff
@@ -346,6 +346,19 @@ method! select_operation op args dbg =
       Ispecific Izextend32, [arg]
     | _ -> super#select_operation op args dbg
     end
+  | Ccsel _ ->
+     begin match args with
+     | [cond; ifso; ifnot] ->
+       let (cond, earg) = self#select_condition cond in
+       (match cond with
+        | Ifloattest (w,CFeq) ->
+          (* CFeq cannot be represented as cmov without a jump.
```
TheNumbat, line 355: I think this should mention that `Icsel` with `CFneq` will do the move on either not-eq or unordered.

TheNumbat (approved):

backend/amd64/selection.ml
gretay-js: I updated the comment.

PR #2572, opened 2024-05-14
Turn upstream compatibility errors into a warning

dkalinichenko-js:
Turn upstream compatibility errors caused by layout annotations into a warning. Edit tests correspondingly.
Since it's now a warning, it can be disabled if we know the annotation is actually unproblematic (i.e. because the whole
declaration is omitted from public release).

Note that `-extension-universe upstream_compatible` still enables only a limited amount of extensions compared to `stable`,
and using most extensions would still cause errors.

### DIFF
--- a/ocaml/testsuite/tests/typing-layouts/erasable_annot.ml
+++ b/ocaml/testsuite/tests/typing-layouts/erasable_annot.ml
@@ -22,8 +22,11 @@ end;;
 Line 2, characters 2-52:
 2 |   val f_immediate : ('a : immediate). 'a -> 'a -> 'a
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-Error: Usage of layout immediate/immediate64 in f_immediate can't be erased for compatibility with upstream OCaml.
-        This error is produced due to the use of -extension-universe (no_extensions|upstream_compatible).
+Warning 187 [incompatible-with-upstream]: Usage of layout immediate/immediate64 in f_immediate can't be erased
+for compatibility with upstream OCaml. This error is produced due to
+the use of -extension-universe upstream_compatible.
+
+module type S = sig val f_immediate : ('a : immediate). 'a -> 'a -> 'a end
 |}];;
 
 module type S = sig
@@ -33,8 +36,11 @@ end;;
 Line 2, characters 2-48:
 2 |   val f_immediate : ('a : immediate) -> 'a -> 'a
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-Error: Usage of layout immediate/immediate64 in f_immediate can't be erased for compatibility with upstream OCaml.
-        This error is produced due to the use of -extension-universe (no_extensions|upstream_compatible).
+Warning 187 [incompatible-with-upstream]: Usage of layout immediate/immediate64 in f_immediate can't be erased
+for compatibility with upstream OCaml. This error is produced due to
+the use of -extension-universe upstream_compatible.
+
+module type S = sig val f_immediate : ('a : immediate). 'a -> 'a -> 'a end
 |}];;
 
 module type S = sig
@@ -44,8 +50,11 @@ end;;
 Line 2, characters 2-25:
 2 |   type ('a : immediate) t
       ^^^^^^^^^^^^^^^^^^^^^^^
-Error: Usage of layout immediate/immediate64 in t can't be erased for compatibility with upstream OCaml.
-        This error is produced due to the use of -extension-universe (no_extensions|upstream_compatible).
+Warning 187 [incompatible-with-upstream]: Usage of layout immediate/immediate64 in t can't be erased
+for compatibility with upstream OCaml. This error is produced due to
+the use of -extension-universe upstream_compatible.
+
+module type S = sig type ('a : immediate) t end
 |}];;
 
 module type S = sig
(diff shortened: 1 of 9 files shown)


ocaml/testsuite/tests/typing-layouts/erasable_annot.ml
```diff
@@ -55,35 +64,68 @@ end;;
...
 |}];;
 
 let f (type a : immediate): a -> a = fun x -> x
 [%%expect {|
 Line 1, characters 4-5:
 1 | let f (type a : immediate): a -> a = fun x -> x
         ^
-Error: Usage of layout immediate/immediate64 in f can't be erased for compatibility with upstream OCaml.
-        This error is produced due to the use of -extension-universe (no_extensions|upstream_compatible).
+Warning 187 [incompatible-with-upstream]: Usage of layout immediate/immediate64 in f can't be erased
```
goldfirere, line 79: Hmmm.... two warnings. Should probably be just one.

ocaml/utils/warnings.ml
```diff
@@ -34,6 +34,11 @@ type constructor_usage_warning =
   | Not_constructed
   | Only_exported_private
 
+type upstream_compat_warning =
+  | Immediate_erasure of string
```
goldfirere, line 38: Maybe add tiny examples of each here? I know other declarations don't do this, but it's so so helpful to. :)

ocaml/utils/warnings.ml
```diff
@@ -1165,15 +1176,33 @@ let message = function
        but is never applied in TMC position."
   | Tmc_breaks_tailcall ->
       "This call\n\
-       is in tail-modulo-cons position in a TMC function,\n\
-       but the function called is not itself specialized for TMC,\n\
-       so the call will not be transformed into a tail call.\n\
-       Please either mark the called function with the [@tail_mod_cons]\n\
-       attribute, or mark this call with the [@tailcall false] attribute\n\
-       to make its non-tailness explicit."
+      is in tail-modulo-cons position in a TMC function,\n\
```
goldfirere, line 1179: Whitespace changes? Might cause unnecessary merge conflicts. Please undo.

ocaml/testsuite/tests/typing-layouts/erasable_annot.ml
goldfirere, line 1: I was expecting some tests around disabling warnings, at least including the example on the internal ticket. But several more would be good, too, using `[@warning "-187"]`, `[@@warning "-187"]`, and `[@@@warning "-187"]`. Make sure the warning gets turned back on outside the scope of the attribute. You can find some examples in the `warnings` and `typing-warnings` directories in the testsuite.

dkalinichenko-js: Added tests for disabling warnings.

ocaml/testsuite/tests/typing-layouts/erasable_annot.ml
dkalinichenko-js: Fixed this particular issue, but some warnings are still duplicated.

ocaml/testsuite/tests/typing-layouts/erasable_annot.ml
```diff
@@ -392,3 +445,102 @@ external f_4 : M3.t -> M3.t = "%identity" [@@unboxed]
...
+  type ('a : immediate) t = 'a * 'a
+end;;
+[%%expect{|
+Line 5, characters 2-35:
+5 |   type ('a : immediate) t = 'a * 'a
+      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Warning 187 [incompatible-with-upstream]: Usage of layout immediate/immediate64 in t
+can't be erased for compatibility with upstream OCaml.
+
+Line 5, characters 2-35:
```
goldfirere, line 526: Duplicated warning?

goldfirere (commented):
So I guess this *is* harder than I thought...

ocaml/typing/typecore.ml
```diff
@@ -8500,8 +8500,10 @@ and type_let ?check ?check_strict ?(force_toplevel = false)
             Tpat_var (id, _, _, _) -> Some id
           | Tpat_alias(_, id, _, _, _) -> Some id
           | _ -> None in
+        (* We don't need to check upstream compatibility for expressions:
```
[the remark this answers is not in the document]
dkalinichenko-js: It's still caught by the previous check, I added a test case demonstrating this.

ocaml/typing/ctype.mli
```diff
@@ -617,7 +617,7 @@ val check_type_externality : Env.t -> type_expr -> Jkind.Externality.t -> bool
 
    *)
 val check_and_update_generalized_ty_jkind :
-  ?name:Ident.t -> loc:Location.t -> type_expr -> unit
+  ?name:Ident.t -> loc:Location.t -> upstream_compat:bool -> type_expr -> unit
```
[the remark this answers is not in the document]
dkalinichenko-js: Done (in the .mli).

ocaml/typing/typecore.ml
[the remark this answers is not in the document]
goldfirere: Thanks, but it doesn't get to the root of the problem. Though it's now getting pretty contrived, this program is not upstream compatible and yet does not trigger the warning:

```ocaml
module type S = sig
  type t [@@immediate]
end

let x =
  ignore (fun (module _ : S with type t = 'a) (_ : 'a) -> 10);
  15
```

This time, I _did_ test!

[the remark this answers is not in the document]
dkalinichenko-js: Yes, you're right. I roll back the change then.

ocaml/testsuite/tests/typing-layouts/erasable_annot.ml
goldfirere: Might be worth deduplicating the pile of warnings, which would solve various problems here. Or if you'd rather avoid that, we can take a look at this together. The algorithm to check type declarations is pretty twisted.

dkalinichenko-js: Yeah, I'd appreciate that.

ocaml/typing/typecore.ml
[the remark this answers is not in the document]
dkalinichenko-js: Rolled back, with duplicated warnings back too.

ocaml/testsuite/tests/typing-layouts/erasable_annot.ml
dkalinichenko-js: Let's fix this in a separate PR as discussed

ocaml/typing/typecore.ml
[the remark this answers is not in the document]
dkalinichenko-js: Left a CR, will be fixed in a separate PR

PR #2573, opened 2024-05-14
Make IRC coalescing respect registers with multiple names

TheNumbat:
IRC currently assumes it is valid to coalese a register `v`, replacing it with `u`, if the two temps are not adjacent in the interference graph.

This is not correct when the same physical register can have multiple names, which became the case after adding `vec128` and `float32`. If `u` refers to the same physical register as some neighbor of `v`, substituting would clobber it.

This PR hence adds an additional condition that `u` does not map to the same physical register as any neighbors of `v`.

Also adds the following checks:
- emit.mlp now raises upon finding a move between registers of incompatible types, even if it's from a location to itself
- with `debug = true`, both IRC and the upstream coloring allocator will assert that all live sets contain each physical register at most once
- IRC `add_alias` will also raise upon finding a move between registers of incompatible types
- CFG CI builds now also build the tests with the enabled register allocator
- The `mach` reg printer includes the type character for easier debugging

### DIFF
--- a/backend/interf.ml
+++ b/backend/interf.ml
@@ -82,6 +82,7 @@ let build_graph fundecl =
      do not add an interference between them if the source is still live
      afterwards. *)
   let add_interf_move src dst s =
+    if debug then assert (Reg.types_are_compatible src dst);
     Reg.Set.iter (fun r -> if r.stamp <> src.stamp then add_interf dst r) s in
 
   (* Compute interferences *)
@@ -130,6 +131,7 @@ let build_graph fundecl =
       float arguments in integer registers, PR#6227.) *)
 
   let add_pref weight r1 r2 =
+    if debug then assert (Reg.types_are_compatible r1 r2);
     let i = r1.stamp and j = r2.stamp in
     if i <> j
     && r1.loc = Unknown
--- a/backend/interf.ml
+++ b/backend/interf.ml
@@ -16,8 +16,7 @@
 (* Construction of the interference graph.
    Annotate pseudoregs with interference lists and preference lists. *)
 
-(* CR mslater: for testing *)
-let debug = true
+let debug = false
 
 module IntPairSet =
   Hashtbl.Make(struct
--- a/backend/interf.ml
+++ b/backend/interf.ml
@@ -16,6 +16,9 @@
 (* Construction of the interference graph.
    Annotate pseudoregs with interference lists and preference lists. *)
 
+(* CR mslater: for testing *)
+let debug = true
+
 module IntPairSet =
   Hashtbl.Make(struct
     type t = int * int
@@ -84,6 +87,7 @@ let build_graph fundecl =
   (* Compute interferences *)
 
   let rec interf i =
+    if debug then assert (not (Reg.set_has_collisions i.live));
     let destroyed = Proc.destroyed_at_oper i.desc in
     if Array.length destroyed > 0 then add_interf_set destroyed i.live;
     match i.desc with
--- a/backend/interf.ml
+++ b/backend/interf.ml
@@ -16,7 +16,7 @@
 (* Construction of the interference graph.
    Annotate pseudoregs with interference lists and preference lists. *)
 
-let debug = false
+let check_collisions = false
 
 module IntPairSet =
   Hashtbl.Make(struct
@@ -81,13 +81,13 @@ let build_graph fundecl =
      do not add an interference between them if the source is still live
      afterwards. *)
   let add_interf_move src dst s =
-    if debug then assert (Reg.types_are_compatible src dst);
+    assert (Reg.types_are_compatible src dst);
     Reg.Set.iter (fun r -> if r.stamp <> src.stamp then add_interf dst r) s in
 
   (* Compute interferences *)
 
   let rec interf i =
-    if debug then assert (not (Reg.set_has_collisions i.live));
+    if check_collisions then assert (not (Reg.set_has_collisions i.live));
     let destroyed = Proc.destroyed_at_oper i.desc in
     if Array.length destroyed > 0 then add_interf_set destroyed i.live;
     match i.desc with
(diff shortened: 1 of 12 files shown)


xclerc:
(Should probably include https://github.com/ocaml-flambda/flambda-backend/pull/2569.)

TheNumbat:
@xclerc I'll turn off the `debug = true`s after you review

backend/reg.ml
```diff
@@ -299,6 +299,15 @@ let set_of_array v =
            if i >= n then Set.empty else Set.add v.(i) (add_all(i+1))
          in add_all 0
 
+let set_has_collisions s =
+  let phys_regs = Hashtbl.create (Set.cardinal s) in
```
xclerc, line 303: (I don't think that matters that much,
but on pathological cases the cardinal
of `s` could be very large, while the
cardinal of `phys_regs` is bounded by
the total number of hardware registers.)

backend/reg.ml
```diff
@@ -299,6 +299,15 @@ let set_of_array v =
...
+  Set.fold (fun r acc ->
+    match r.loc with
+    | Reg id ->
+      if Hashtbl.mem phys_regs id then true
+      else (Hashtbl.add phys_regs id (); acc)
+    | _ -> acc) s false
```
xclerc, line 309: (Again, does not matter that much for
such a function, but please avoid
catch-all patterns.)

backend/reg.ml
```diff
@@ -321,6 +330,12 @@ let equal_location left right =
   | Stack _, (Unknown | Reg _) ->
     false
 
+let same_phys_reg left right =
+  match left.loc, right.loc with
+  | Reg l, Reg r -> Int.equal l r
+  | Reg _, (Unknown | Stack _)
+  | (Unknown | Stack _), _ -> false
```
xclerc, line 337:
```suggestion
  | (Reg _ | Unknown | Stack _), _ -> false
```

backend/regalloc/regalloc_irc.ml
```diff
@@ -199,7 +201,7 @@ let coalesce : State.t -> unit =
     if irc_debug then log ~indent:2 "case #1/4";
     State.add_coalesced_moves state m;
     add_work_list state u)
-  else if State.is_precolored state v || State.mem_adj_set state u v
+  else if State.is_precolored state v || State.interferes_with_adj state v u
```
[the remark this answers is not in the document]
xclerc: Either here or in the `build` function above.

mshinwell (changes requested):

backend/regalloc/regalloc_irc_state.ml
```diff
@@ -466,11 +470,13 @@ let[@inline] rec find_alias state reg =
   else reg
 
 let[@inline] add_alias _state v u =
-  if not (same_reg_class v u)
+  (* We should never generate moves between registers of different types.
+     Bit-casting operations have specific instructions. *)
+  if not (types_are_compatible v u)
```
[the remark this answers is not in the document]
xclerc: I think this `types_are_compatible` function
should also be used in `Interf.add_pref`.

backend/reg.ml
TheNumbat: Fixed

backend/reg.ml
TheNumbat: True, except we can't refer to `Proc.num_available_registers` here without breaking a dependency cycle.
Changed to min with 32 instead

backend/regalloc/regalloc_irc.ml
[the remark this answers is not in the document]
TheNumbat: Added

backend/regalloc/regalloc_irc_state.ml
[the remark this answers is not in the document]
TheNumbat: It currently doesn't need the conjunction; if the types are compatible the class is the same. 
It wouldn't be a refinement if the architecture gave Int/Val/Addr different classes, but I don't think that will ever happen.

(Also added assertions in `add_pref`)

backend/interf.ml
```diff
@@ -84,6 +87,7 @@ let build_graph fundecl =
   (* Compute interferences *)
 
   let rec interf i =
+    if debug then assert (not (Reg.set_has_collisions i.live));
```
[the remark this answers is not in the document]
TheNumbat: We could do this, but if it needs to change select/reg significantly it should probably be another pr

backend/regalloc/regalloc_irc_state.ml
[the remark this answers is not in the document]
xclerc: Sorry, I may have misunderstood one of your comments off-line.

I am fine either way, but:
- if it does not need to be a refinement, we should have the conjunction;
- if it needs to be a refinement, there should be a comment to that effect near the `types_compatible` function.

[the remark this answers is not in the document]
TheNumbat: Ah ok, added another comment

backend/interf.ml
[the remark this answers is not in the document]
mshinwell: That one should probably be merged first then, if it's in another PR.

backend/interf.ml
```diff
@@ -79,11 +81,13 @@ let build_graph fundecl =
      do not add an interference between them if the source is still live
      afterwards. *)
   let add_interf_move src dst s =
+    if debug then assert (Reg.types_are_compatible src dst);
```
xclerc, line 84: I would be tempted to have this assertion even when `debug` is `false`. It is probably cheap, and would catch more cases when a move between incompatible types has been produced. ("more" meaning `emit.mlp` would catch moves present until the end while here we would also catch bad intentions.)

mshinwell: Please also use `Misc.fatal_errorf` so when an assertion fails, we don't necessarily need to rebuild to get useful debugging info.

xclerc: (Sorry, I meant to attach this comment to the `add_pref` instance, where it makes more sense.)

TheNumbat: Sure, enabled the assertions unconditionally (except the collision check)

TheNumbat: Oh, didn't see the new comments - updated with error messages

backend/interf.ml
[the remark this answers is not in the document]
TheNumbat: (Discussed offline)

xclerc (approved):

PR #2575, opened 2024-05-15
Unboxed product fixes for Lambda_to_flambda

mshinwell:
@ccasin provided the following two test cases (for use with `ccasin/unboxed-tuples`) which were causing errors in `Lambda_to_flambda`:

```
let[@inline never] flip #((x : int), (y : float#)) = #(y, x)
let[@inline never] unflip #(y, x) = #(x, y)

let id p = unflip (flip p)
```

```
type t = { x : string; y : int; z : float# }

let[@inline never] unbox {x; y; z} = #(x, y, z)

let[@inline never] flip #(x, y, z) = #(z, y, x)

let id t =
  let #(z, y, x) = flip (unbox t) in
  {x; y; z}
```

There are two commits to fix these.  I'm pretty sure about the code metadata commit, which fixes a problem whereby unarized `params_arity` values were incorrectly appearing in code metadata, rather than their non-unarized equivalents.

The second commit needs more careful scrutiny, and further testing in the unboxed product world which @ccasin is looking at, but I would like to get it merged if we agree it's not going to break code that does not use unboxed products.

@Ekdohibs could you please look at these?

### DIFF
--- a/middle_end/flambda2/from_lambda/closure_conversion.ml
+++ b/middle_end/flambda2/from_lambda/closure_conversion.ml
@@ -1565,9 +1565,7 @@ let compute_body_of_unboxed_function acc my_region my_closure
             (fun (var, kind) -> Bound_parameter.create var kind)
             vars_with_kinds
           @ main_code_params,
-          List.map
-            (fun (_var, kind) ->
-              Flambda_arity.Component_for_creation.Singleton kind)
+          List.map snd
             vars_with_kinds
           @ main_code_params_arity,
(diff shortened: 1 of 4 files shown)


mshinwell:
(@Ekdohibs is going to finish this)

mshinwell:
Here is another example from @Ekdohibs : 

`let f (#(_, x) : #(int * float)) (y[@unboxable]) = x +. y`

cc @ccasin, this could go in the testsuite too

Ekdohibs (approved):

PR #2576, opened 2024-05-15
Revert "Simplify the `install` stanza of extension universe otherlibs"

mshinwell:
I'd like to revert this for now, to understand if there is another solution which doesn't involve breaking existing build environments and OPAM packages.

Reverts ocaml-flambda/flambda-backend#2563


mshinwell:
This unfortunately doesn't work because there are dependent PRs, I'll see if the dune files can be fixed instead.

PR #2577, opened 2024-05-15
IRC: register synonyms

xclerc:
(This is really an early draft, exploring the design space.)

This pull request tweaks the computation of IRC's
interference graph by ensuring that when we add an
edge involving an hardware register, we actually add
edges to all its synonyms.

A "synonym" is this context refers to the fact that
_e.g._ `xmm0` is represented by several `Reg.t` instance
when SIMD and/or float32 is enabled. There is indeed
one instance per machtype, even though they share
the same hardware location (and thus register class).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -71,6 +71,7 @@ jobs:
             config: --enable-middle-end=flambda2
             os: ubuntu-latest
             build_ocamlparam: '_,w=-46,regalloc=irc,regalloc-param=SPLIT_LIVE_RANGES:on,regalloc-param=IRC_SPILLING_HEURISTICS:flat-uses,regalloc-validate=1'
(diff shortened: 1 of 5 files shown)


xclerc:
(This is an alternative to https://github.com/ocaml-flambda/flambda-backend/pull/2573, at most one of the two should be merged.)

mshinwell:
As discussed this morning, it seems like we should probably have this and #2573.  However, maybe if we merged #2573 first, perhaps this PR could be changed such that it alters the extra interference check added by #2573: in particular, it could turn that check into an assertion.  If this PR is operating correctly, it seems like we should never discover an interference by looking at the equivalence class of registers with the same physical location, as this PR should have caused the adjacency matrix to be populated with the relevant edges.

PR #2584, opened 2024-05-15
Simplify otherlibs install stanza

dkalinichenko-js:
Generate the installation list using a script.

### DIFF
--- a/ocaml/otherlibs/to_install.sh
+++ b/ocaml/otherlibs/to_install.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+name=$1
+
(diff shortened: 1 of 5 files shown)


ocaml/otherlibs/to_install.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
mshinwell, line 1: Shell scripts should have `set -eu` unless there is a good reason.

dkalinichenko-js: Done.

PR #2585, opened 2024-05-15
Refactor `Persistent_env` to separate imports from bound names

lukemaurer:
**Re-created from #1764, which got merged to the wrong base**

With the advent of parameters, we have reasons to refer to global module names besides those names being runtime global variables, or addressable values at all. In addition, we're going to want to load a .cmi but then parameterise it in different ways to get different signatures, each of which will be bound to a separate value.

In the end, then, instead of just `pers_struct`, we'll want three different record types, each with an associated cache:

* `import`, corresponding directly to a .cmi file, keyed by `Compilation_unit.Name.t`
* `pers_name`, corresponding to an `import` with parameters applied to it, keyed by a `Compilation_unit.Name.t` and some parameters
* `pers_struct`, corresponding to a persistent name that is actually bound in the environment, with the same key as `pers_name`

For this PR, I'm leaving out the second cache as nothing about it is relevant yet, but the split is disruptive enough that even the two are worthwhile. Note that `pers_struct` is currently still keyed by `Compilation_unit.Name.t` since we don't yet have the datatypes for parameterised names.

Besides the internal reorganisation of `Persistent_env`, this also changes the API between it and `Env`, offloading some of the work done by the callback, `read_sign_of_cmi`. Significantly, `read_sign_of_cmi` is no longer called _at all_ until something is going to be bound (i.e., become a `pers_struct`). This keeps the `Env.t` completely free of anything it shouldn't know about.


lpw25 (approved):
Approving on the assumption that it is the same code I already approved.

PR #2588, opened 2024-05-16
Move unboxed types modules to `stdlib_upstream_compatible`

dkalinichenko-js:
Move `Float_u`, `Int32_u`, `Int64_u` and `Nativeint_u` to `stdlib_upstream_compatible`.

Silence warnings requiring `[@@unboxed]` annotations on external definitions: since `%box_float` and the like are not available upstream, erasing those would include replacing the intrinsics with `%identity`, which does not require `[@@unboxed]`.

Find and replace uses of `stable` in tests with `stdlib_upstream_compatible` where necessary.


ocaml/otherlibs/stdlib_upstream_compatible/float_u.ml
```diff
@@ -19,9 +19,9 @@ open! Stdlib
 
 type t = float#
 
-external to_float : t -> (float[@local_opt]) = "%box_float"
+external to_float : t -> (float[@local_opt]) = "%box_float" [@@warning "-187"]
```
goldfirere, line 22: Actually, I think we can (confusingly) compile the upstream-compatible otherlib with `stable`. The upstream-compatible otherlib will never actually get compiled with the upstream compiler. So it can use all manner of non-upstream-compatible features quite happily. Of course, its interface needs to be shared with a shim library, but that's a separate concern.

goldfirere (approved):

ocaml/Makefile.common-jst
```diff
@@ -113,7 +113,6 @@ dune_config_targets = \
   $(ocamldir)/duneconf/dirs-to-ignore.inc \
   $(ocamldir)/duneconf/jst-extra.inc \
   dune-project \
-  $(ocamldir)/stdlib/ocaml_compiler_internal_params \
```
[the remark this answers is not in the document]
dkalinichenko-js: I already removed this hack from the tree :)

ocaml/otherlibs/stdlib_upstream_compatible/float_u.ml
dkalinichenko-js: Huh, you're right. I'll still leave it at `upstream_compatible` for now, but I'll think about it if we need to add more stuff.

dkalinichenko-js: Actually, since we want to generate the shim interface by erasing the actual interface, probably best to leave it at `upstream_compatible`.

goldfirere: Up to you as this evolves, but this suggests, to me, just having `[@@@warning "-187"]` right at the top of the .ml files (with a comment explaining why this is OK). No need to do that now, just planting a seed for later. :)

PR #2593, opened 2024-05-20
Specialize Sys.max_array_length for arrays of unboxed numbers

TheNumbat:
We're currently using `Sys.max_array_length` as the max length for all kinds of arrays, but the packed `int32# array` and `float32# array` can support twice as many elements. 

Additionally, unboxed non-float64 arrays need to subtract one from max_wosize to account for the custom ops word.


ocaml/testsuite/tests/typing-layouts-arrays/test_int32_u_array.ml
```diff
@@ -46,6 +46,13 @@ module Int32_u_array0 : Gen_u_array.S0
   type ('a : any) array_t = 'a array
   type element_arg = unit -> element_t
   type t = element_t array
+
+  let max_length =
+    match Sys.backend_type with
+    | Bytecode -> Sys.max_array_length
+    | Native -> Sys.max_unboxed_int32_array_length
+    | Other _ -> assert false
```
mshinwell, line 54: Not sure about this - maybe this conditional should be directly in `Sys`?

mshinwell (changes requested):

ocaml/testsuite/tests/typing-layouts-arrays/test_int32_u_array.ml
TheNumbat: Sure, moved the conditionals to sys

PR #2596, opened 2024-05-20
Refactor backend casts

TheNumbat:
Refactors numeric casts into two new types:
- `reinterpret_cast` compiles to a single move instruction and will be eliminated if the argument and result are assigned the same register
- `static_cast` does some additional logic to preserve the value of the argument (truncate float->int, convert float->float32, sign extend int16x8->int and int8x16->int, convert float32x4->float and float->float32x4)

Reinterpret cast subsumes the `intofvalue` and `valueofint` instructions, which have been deleted. Special cases for these instructions have been preserved.

The bit-cast SSE instructions have also been deleted and replaced with reinterpret casts. The corresponding builtin function recognition has been moved up to `cmm_builtins.ml`, since there are now corresponding cmm instructions.

The float64<->int64 reinterpret casts have also been implemented in the arm64 backend, but no float32 or vec128 operations are supported.

Existing tests should cover all of these instructions.


backend/cfg/cfg.ml
```diff
@@ -543,8 +550,7 @@ let same_location (r1 : Reg.t) (r2 : Reg.t) =
 
 let is_noop_move instr =
   match instr.desc with
-  | Op (Move | Spill | Reload | Vectorcast _) ->
```
TheNumbat, line 546: (Vectorcast intentionally removed for consistency)

backend/reloadgen.ml
```diff
@@ -63,7 +63,7 @@ method reload_operation op arg res =
      res to be stack-allocated, but do something for
      stack-to-stack moves *)
   match op with
-    | Imove | Ireload | Ispill | Iintofvalue | Ivalueofint | Ivectorcast _ ->
```
TheNumbat, line 66: The reinterpret casts between ints and float32/float will never have the same stack class, so this is equivalent

backend/cmm.ml
```diff
@@ -203,17 +203,24 @@ type memory_chunk =
...
+type reinterpret_cast =
+  | Int_of_value
+  | Value_of_int
+  | Float_of_float32
+  | Float32_of_float
+  | Float_of_int64
+  | Int64_of_float
+  | Float32_of_int32
+  | Int32_of_float32
+  | V128_of_v128
```
xclerc, line 215: (I think it is not immediately clear
that the purpose of this one is.)

backend/cmm.mli
```diff
@@ -178,18 +178,28 @@ type memory_chunk =
...
-type scalar_cast =
-  (* CR mslater: move all bit-casts into a reinterpret_cast type *)
-  | Float32_as_float
-  | Float_to_int of float_width
+(* These casts compile to a single move instruction. If the operands are assigned
+   the same physical register, the move will be omitted entirely. *)
+type reinterpret_cast =
+  | Int_of_value
+  | Value_of_int
+  | Float_of_float32 (* The high 32 bits of the float are undefined. *)
```
xclerc, line 186: (I am slightly confused: aren't we using
128-bits registers, and storing only 32-bits
in memory?)

backend/printmach.ml
```diff
@@ -225,22 +225,25 @@ let operation' ?(print_reg = reg) op arg ppf res =
...
-    reg arg.(0)
-  | Iscalarcast Float32_as_float -> fprintf ppf "float32 as float %a" reg arg.(0)
-  | Iscalarcast (Float_of_int Float64) -> fprintf ppf "int->float %a" reg arg.(0)
-  | Iscalarcast (Float_to_int Float64) -> fprintf ppf "float->int %a" reg arg.(0)
-  | Iscalarcast (Float_of_int Float32) -> fprintf ppf "int->float32 %a" reg arg.(0)
-  | Iscalarcast (Float_to_int Float32) -> fprintf ppf "float32->int %a" reg arg.(0)
-  | Iscalarcast (Float_of_float32) -> fprintf ppf "float32->float %a" reg arg.(0)
-  | Iscalarcast (Float_to_float32) -> fprintf ppf "float->float32 %a" reg arg.(0)
-  | Iscalarcast (V128_of_scalar ty) ->
+  | Ireinterpret_cast Int_of_value -> fprintf ppf "value as int %a" reg arg.(0)
```
xclerc, line 228: I think part of the printing logic should
be shared with `Printcmm`.

backend/printmach.ml
```diff
@@ -225,22 +225,25 @@ let operation' ?(print_reg = reg) op arg ppf res =
...
+  | Ireinterpret_cast Value_of_int -> fprintf ppf "int as value %a" reg arg.(0)
+  | Ireinterpret_cast Float32_of_float -> fprintf ppf "float as float32 %a" reg arg.(0)
+  | Ireinterpret_cast Float_of_float32 -> fprintf ppf "float32 as float %a" reg arg.(0)
+  | Ireinterpret_cast Int64_of_float -> fprintf ppf "int64 as float %a" reg arg.(0)
```
xclerc, line 232: It looks like there are inconsistencies
in the strings: for `x_of_y`, we sometimes
have `"x as y"` and sometimes `"y as x"`.

backend/reloadgen.ml
xclerc: I should probably appear as a comment in
the code, perhaps specifying what "int"
means: it does not soundstrictly impossible
to have the same class for unboxed int32
and unbox float.

backend/cfg/cfg.ml
```diff
@@ -284,19 +284,24 @@ let dump_op ppf = function
...
-  | Vectorcast Bits128 -> Format.fprintf ppf "vec128->vec128"
-  | Scalarcast Float32_as_float -> Format.fprintf ppf "float32 as float"
-  | Scalarcast (Float_of_int Float64) -> Format.fprintf ppf "int->float"
-  | Scalarcast (Float_to_int Float64) -> Format.fprintf ppf "float->int"
-  | Scalarcast (Float_of_int Float32) -> Format.fprintf ppf "int->float32"
-  | Scalarcast (Float_to_int Float32) -> Format.fprintf ppf "float32->int"
-  | Scalarcast Float_of_float32 -> Format.fprintf ppf "float32->float"
-  | Scalarcast Float_to_float32 -> Format.fprintf ppf "float->float32"
-  | Scalarcast (V128_to_scalar ty) ->
+  | Reinterpret_cast V128_of_v128 -> Format.fprintf ppf "vec128 as vec128"
```
xclerc, line 287: (Could be factored out with `Printcmm`
and `Printmach`.)

backend/cfg/cfg.ml
```diff
@@ -284,19 +284,24 @@ let dump_op ppf = function
...
+  | Reinterpret_cast Float_of_float32 -> Format.fprintf ppf "float32 as float"
+  | Reinterpret_cast Float_of_int64 -> Format.fprintf ppf "int64 as float"
+  | Reinterpret_cast Int64_of_float -> Format.fprintf ppf "float as int64"
+  | Reinterpret_cast Float32_of_int32 -> Format.fprintf ppf "int32 as float32"
+  | Reinterpret_cast Int32_of_float32 -> Format.fprintf ppf "float32 as int32"
+  | Static_cast (Float_of_int Float64) -> Format.fprintf ppf "int->float"
+  | Static_cast (Int_of_float Float64) -> Format.fprintf ppf "float->int"
+  | Static_cast (Float_of_int Float32) -> Format.fprintf ppf "int->float32"
+  | Static_cast (Int_of_float Float32) -> Format.fprintf ppf "float32->int"
+  | Static_cast Float_of_float32 -> Format.fprintf ppf "float32->float"
```
xclerc, line 300: (It is not really part of this pull request,
but I wonder whether operations that
involve rounding should be considered
not pure.)

backend/amd64/emit.mlp
```diff
@@ -988,6 +988,59 @@ let emit_atomic instr op (size : Cmm.atomic_bitwidth) addr =
     I.set E res8;
     I.movzx res8 res
 
+let emit_reinterpret_cast (cast : Cmm.reinterpret_cast) i =
+  let distinct = i.arg.(0).loc <> i.res.(0).loc in
```
xclerc, line 992: (Not part of this pull request, but we
should probably avoid the polymorphic
comparison - same in `emit_static_cast`.)

backend/cfg/cfg.ml
mshinwell: Actually I'm not sure we should factor this out: in general it's better if IRs don't share types.

xclerc: I think they already do, the constructors
are in `Cmm` (`reinterpret_cast` and
`static_cast`).

backend/reloadgen.ml
TheNumbat: Sorry, that wasn't quite right. I mean currently floats / ints have a different stack class, so this will never apply, but even if we combine them this condition should be applicable - there's no transformation that must be done to the stack slot.
But either way, the amd64 reload intercepts reinterpret casts other than value<->int anyway, since register behavior matters for gpr<->xmm moves.

Added a comment saying this

backend/cmm.ml
TheNumbat: Added a comment

backend/printmach.ml
TheNumbat: Good point, done

backend/cfg/cfg.ml
TheNumbat: Factored out - maybe we should separate the cast types, but it was already like that

backend/cfg/cfg.ml
TheNumbat: Hmm, that might be true for the `Round_current_f64_i64` SIMD instruction, which uses the current rounding mode.
I think it doesn't apply to these, though, because they always truncate to int / pick nearest float regardless of rounding mode.

backend/printmach.ml
TheNumbat: Oops, fixed

backend/cmm.mli
xclerc: Do you mean that "undefined" should
be interpreted here as "ignored by
the conversion"?

mshinwell: Please ensure all this is sufficiently commented (especially in non-target-specific files, so when we come to do e.g. arm64 we remember these points).

backend/cfg/cfg.ml
xclerc: Indeed, you cannot manipulate it
from OCaml code - I am not sure
there are rules about the C bindings,
but I agree that if someone changes
the rounding mode that way, they
should use `opaque` to forbid any
code motion.

TheNumbat: Added a cr-someday

backend/cmm.mli
TheNumbat: Yeah, clarified the comment. I think there aren't any other concerns for arm64 here

xclerc (approved):

PR #2601, opened 2024-05-21
Properly handle regions around recursive definitions

lthls:
Fixes an oversight in #2394: recursive values are not allowed to be allocated on the stack, but their defining expressions can still contain (temporary) stack allocations, in which case a region is inserted around the whole definition.

I think @goldfirere should review, although I can find someone else if needed.


Ekdohibs (approved):

goldfirere (commented):
I don't know enough to confidently review, but I bet @Ekdohibs's review is sufficient.

PR #2602, opened 2024-05-21
Fix frame pointers build under runtime5

mshinwell:
Under runtime5 there is an unused function `rewrite_frame_pointers` which the C compiler complains about, unless stack checks are enabled.  This PR fixes that and changes one of the existing CI jobs to use runtime5 and frame pointers.

@xclerc please review


xclerc (approved):

mshinwell:
We'll merge this as part of #2603

PR #2603, opened 2024-05-21
Fix for stack checks / frame pointers

xclerc:
When stack stacks are disabled, we rely
on a guard page whose access triggers
a segfault. The segfault is caught and
a signal handler is in charge of raising
a proper OCaml exception.

However, when frame pointers are
enabled, the wrapper in `amd64.S`
would push a value onto the stack,
thus triggering another segfault.


ocaml/runtime/amd64.S
```diff
@@ -548,7 +548,10 @@ ENDFUNCTION(G(caml_call_gc))
 FUNCTION(G(caml_raise_stack_overflow_nat))
 CFI_STARTPROC
         CFI_SIGNAL_FRAME
-        ENTER_FUNCTION
+        /* Do not ENTER_FUNCTION:
+           - when frame pointers are disaabled, it is a no-op;
```
xclerc, line 552:
```suggestion
           - when frame pointers are disabled, it is a no-op;
```

mshinwell:
I suggest rebasing onto https://github.com/ocaml-flambda/flambda-backend/pull/2602 to get the CI check.

xclerc:
Rebase, I wanted to have a clean run
before to be sure I wasn't breaking
anything...

mshinwell (approved):

PR #2605, opened 2024-05-21
Fix mixed-block-related segfault in runtime4's `redarken_chunk`

ncik-roberts:
Mixed blocks allowed you to get a segfault in `redarken_chunk` in runtime 4 in code that creates mixed blocks. This PR attempts to fix that.

The root of the problem is that there are two reasons that code used to access the size stored in the header:
  * To know the start of the next object laid out linearly in memory
  * To detect how many words the GC should scan

Prior to mixed blocks, these concepts were one and the same. Mixed blocks force you to distinguish these concepts.
  * `Wosize_val` tells you where the next object is
  * `Scannable_wosize_val` tells you how far the GC can scan

The bug in `redarken_chunk` was because I mistook the surrounding code as accessing the size only for the first reason. Actually, it accessed the size for both reasons:
  * `redarken_chunk` traverses objects laid out linearly in memory, so needs to call `Wosize_val` to know where the next object starts
  * `redarken_chunk` also scans fields, so needs to call `Scannable_wosize_val` to know which fields it is allows to scan

The comment above `mark_stack_prune` provides some context here:
```
/* This function prunes the mark stack if it's about to overflow. It does so
   by building a skiplist of major heap chunks and then iterating through the
   mark stack and setting redarken_start/redarken_end on each chunk to indicate
   the range that requires redarkening. */
```
Then, `redarken_chunk` is the thing that interprets the skiplist created by `mark_stack_prune` to actually re-scan and redarken the necessary heap chunks.

This PR attempts to fix the issue by tracking the two concepts (corresponding to the two reasons) separately. We change the `mark_entry` struct to track both the end of the object and the end of the _scannable portion_ of the object.

How was this missed in the first PR?
---

I think there are two reasons:
    - the runtime 5 code is refactored in such a way that avoids the need for similar bookkeeping. The `mark_stack_prune` function there appears to only need the "how far to scan" sense.
    - it requires some effort to get the runtime to decide it wants to call `redarken_chunk`, so the mixed blocks tests didn't catch it.

The original PR #2422  mentions "signposts" that suggest that changes for mixed blocks are needed. Checking `No_scan_tag` is one of these signposts. This triggered me to do another scan through for the signposts mentioned in that PR, and I caught two more issues **in runtime4**:
  * `Obj.truncate` would call `caml_modify` on a unscannable field. (`Obj.truncate` isn't exposed in the stdlib anymore, but still it's good to fix.)
  * Some heap verification code that runs in debug mode could falsely flag an issue.
 
I fix both of these issues in this PR, but will pull them out into another one if this PR stalls

Review
---

@stedolan , could you please review? (I will solicit suggestions from others to figure out if there's a better way to test this.)

### DIFF
--- a/ocaml/runtime4/caml/major_gc.h
+++ b/ocaml/runtime4/caml/major_gc.h
@@ -22,11 +22,16 @@
 #include "misc.h"
 
 /* An interval of a single object to be scanned.
-   The end pointer must always be one-past-the-end of a heap block,
-   but the start pointer is not necessarily the start of the block */
+   The object end pointer must always be one-past-the-end of a heap block,
+   but the start pointer is not necessarily the start of the block,
+   and the scannable end pointer is not necessarily the same as the
+   object end pointer. (The GC should not scan past the scannable
+   end pointer.)
+*/
 typedef struct {
   value* start;
-  value* end;
+  value* scannable_end;
+  value* object_end;
 } mark_entry;
 
 typedef struct {
--- a/ocaml/runtime4/caml/mlvalues.h
+++ b/ocaml/runtime4/caml/mlvalues.h
@@ -266,6 +266,8 @@ Caml_inline mlsize_t Scannable_wosize_reserved_byte(reserved_t res,
 #define Bhsize_hp(hp) (Bsize_wsize (Whsize_hp (hp)))
 #define Bhsize_hd(hd) (Bsize_wsize (Whsize_hd (hd)))
 
+#define Scannable_wosize_hp(hp) (Scannable_wosize_hd (Hd_hp (hp)))
+
 #define Profinfo_val(val) (Profinfo_hd (Hd_val (val)))
 
 #ifdef ARCH_BIG_ENDIAN
--- a/ocaml/runtime4/major_gc.c
+++ b/ocaml/runtime4/major_gc.c
@@ -384,7 +384,7 @@ static int redarken_chunk(char* heap_chunk, struct mark_stack* stk) {
 
     /* Found a block */
     me.start = Op_hp(hp);
-    me.end = me.start + Wosize_hp(hp);
+    me.end = me.start + Scannable_wosize_hp(hp);
     if (Tag_hp(hp) == Closure_tag) {
       me.start += Start_env_closinfo(Closinfo_val(Val_hp(hp)));
     }
--- a/ocaml/runtime4/major_gc.c
+++ b/ocaml/runtime4/major_gc.c
@@ -166,8 +166,8 @@ static void mark_stack_prune (struct mark_stack* stk)
       if (ch->redarken_first.start > me.start)
         ch->redarken_first = me;
 
-      if (ch->redarken_end < me.end)
-        ch->redarken_end = me.end;
+      if (ch->redarken_end < me.object_end)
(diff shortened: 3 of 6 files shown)


mshinwell:
This looks correct to me.  One thing for future discussion (and maybe a PR) is whether the non-scannable portion of mixed blocks should be accounted for in the work done.  It seems that maybe it should be, for consistency with blocks with tags over `No_scan_tag`, whose size does appear to be accounted for in this manner.

mshinwell (approved):

ncik-roberts:
> It seems that maybe it should be, for consistency with blocks with tags over No_scan_tag, whose size does appear to be accounted for in this manner.

I believe that the non-scannable portion of a mixed block is accounted for in the same way as a block with `No_scan_tag`:
  * [Here's where `No_scan_tag` is accounted for.](https://github.com/ocaml-flambda/flambda-backend/blob/3d3c319d305807d452f609bda8cc4155ad808ffe/ocaml/runtime4/major_gc.c#L686)
  * [Here's where the non-scannable suffix of a mixed block is accounted for.](https://github.com/ocaml-flambda/flambda-backend/blob/3d3c319d305807d452f609bda8cc4155ad808ffe/ocaml/runtime4/major_gc.c#L692)

There's analogous code in runtime5; see [here](https://github.com/ocaml-flambda/flambda-backend/blob/3d3c319d305807d452f609bda8cc4155ad808ffe/ocaml/runtime/major_gc.c#L914-L925).

I had this wrong initially, where the non-scanned portion wasn't accounted for, but Stephen argued for the current accounting in [this review comment](https://github.com/ocaml-flambda/flambda-backend/pull/2422#discussion_r1569152019). 

My sense is that there isn't anything more to discuss.

mshinwell:
Ah hmm, maybe that part is correct indeed.  What about this line though?
https://github.com/ocaml-flambda/flambda-backend/pull/2605/files#diff-cbdcb0d08ee6694a8d52b3ee2c91bed023d83e6e06abc30f000dfe5ff0f7b788R735

ncik-roberts:
That line should be OK, as `scan_end` won't be past the scannable end of the object. See [this link](https://github.com/ocaml-flambda/flambda-backend/pull/2605/files#diff-cbdcb0d08ee6694a8d52b3ee2c91bed023d83e6e06abc30f000dfe5ff0f7b788R716-R722) (it might take a few seconds to load, as diff views sometimes do).

ncik-roberts:
I think that line ends up being OK.

The point of that line is to avoid taking credit for scanning a portion that doesn't get scanned. That line runs when the precheck buffer doesn't have more room, so instead the remaining scanning work is added to the mark stack. The mark stack is read from here:

https://github.com/ocaml-flambda/flambda-backend/pull/2605/files#diff-cbdcb0d08ee6694a8d52b3ee2c91bed023d83e6e06abc30f000dfe5ff0f7b788R711-R713

The concern you might have is that the accounting [here](https://github.com/ocaml-flambda/flambda-backend/blob/3d3c319d305807d452f609bda8cc4155ad808ffe/ocaml/runtime4/major_gc.c#L692) that takes credit for the non-scannable portion of a mixed block will run twice. But in fact this doesn't happen: when future work that's added to the mark stack is popped, the mixed block accounting doesn't run again.

PR #2606, opened 2024-05-22
Typing rules support kind polymorphism

goldfirere:
At long last, I'm at a decent stopping point for writing down the typing rules for the language with kind polymorphism. Here they are.

### DIFF
--- a/ocaml/jane/doc/proposals/kind-inference.md
+++ b/ocaml/jane/doc/proposals/kind-inference.md
@@ -1,0 +1,716 @@
+# Kind inference
+
+This document is a formal description of the kind system Richard is
+proposing for OCaml. It is heavy on symbols and light on text.
+
+We use `{{ ... }}` to denote an optional piece of syntax, and
+`[[ ... ]]` to denote zero or more occurrences of a piece of syntax.
+In `[[ ... ]]ᵢ`, the `ᵢ` is the number of repetitions of the syntax.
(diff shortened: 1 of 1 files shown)


ocaml/jane/doc/proposals/kind-inference.md
```diff
@@ -0,0 +1,827 @@
...
+  (* pre-condition: Γ₂ ⊆ Γ *)
+
+t : tconstr_jkind ∈ Γ₂
+--------------------------- FIND_ABSTRACT
+Γ ⊢find t ∈ Γ₂ ↠ tconstr_jkind
+
+t = λ [[ 'aᵢ : κᵢ ]]. τ ∈ Γ₂
+Γ, [[ 'aᵢ : κᵢ ]] ⊢ τ : κ₀
+------------------------------------- FIND_ABBREV
+Γ ⊢find t ∈ Γ₂ ↠ π [[ 'aᵢ : κᵢ ]]. κ₀
```
goldfirere, line 582: This is wrong. It should use `kind_of`. Otherwise

```ocaml
module M : sig
  type ('a : value) t : value
end = struct
  type ('a : any) t = 'a
end
```

is rejected, because the struct kind is just `any`. It needs to be `kind_of 'a`, in which case all would be well.

lpw25 (approved):
Approving based on a discussion about these rules rather than actually going through and reading them in this PR.

PR #2607, opened 2024-05-23
Flatten `Import_info.t`

lukemaurer:
PR #1753 refactored the `Import_info.t` type into a tagged union of tagged unions, which may be increasing artifact sizes. This flattens the new type so that it should be the same size as the status quo ante (or better).


xclerc (approved):

mshinwell:
@lukemaurer please rebase

lukemaurer:
> @lukemaurer please rebase

Oh, I thought I'd closed this. It was only intended as a stopgap; #2552 accomplishes the same thing but better. (I do need to rebase that.)

PR #2609, opened 2024-05-23
Remove redundant `CU.Name.t` from normal import info

lukemaurer:
**Needs #1753 and #2585 to be un-reverted before it can be merged.**

When we have both a `CU.Name.t` and a `CU.t`, they must always match, so storing the `CU.Name.t` is redundant.


xclerc (approved):

lukemaurer:
This got cherry-picked into #2621 and merged in that form.

PR #2610, opened 2024-05-23
Stack checks: IPA optimization

xclerc:
This pull request implements an inter-procedural
optimization of stack checks. If, for instance,
`f` must include a check and calls `g` that must
also include a check, then `f` can perform the
check for both functions and, when calling `g`,
jump to a label after the check. It means that
the check is still present in `g`, and it is
the responsibility of the caller to decide whether
it wants to skip it or not.

In order for the optimization to be applied across
compilation units, we store in cmx files the
information about the stack check for each function.
As a simplification, the current version restricts
the shrink-wrap optmization to `caml_apply_xyz`
functions.

This is a early draft, to gather feedback; various
elements may need to be optimized, and we should
consider sharing some elements with the zero_alloc
machinery.

### DIFF
--- a/backend/.ocamlformat-enable
+++ b/backend/.ocamlformat-enable
@@ -2,6 +2,8 @@ cmm_helpers.ml
 cmm_helpers.mli
 cmm_builtins.ml
 cmm_builtins.mli
+stack_check_info.ml
+stack_check_info.mli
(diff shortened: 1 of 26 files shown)


mshinwell:
I will review this one.

PR #2615, opened 2024-05-23
Fix `check_and_update_generalized_ty_jkind`

dkalinichenko-js:
Fix duplicated warnings raised by `check_and_update_generalized_ty_jkind` by tracking the warning in the layout history.
Run the immediacy check regardless of the type variable level.

Some complicated examples still don't raise the warning. Document them.

### DIFF
--- a/ocaml/typing/jkind.ml
+++ b/ocaml/typing/jkind.ml
@@ -526,17 +526,22 @@ include Jkind_intf.History
 
 type t = type_expr Jkind_types.t
 
-let fresh_jkind jkind ~why = { jkind; history = Creation why }
+let fresh_jkind jkind ~why =
+  { jkind; history = Creation why; has_warned = false }
 
 (******************************)
 (* constants *)
 
 let any_dummy_jkind =
-  { jkind = Jkind_desc.max; history = Creation (Any_creation Dummy_jkind) }
+  { jkind = Jkind_desc.max;
+    history = Creation (Any_creation Dummy_jkind);
+    has_warned = false
+  }
 
 let value_v1_safety_check =
   { jkind = Jkind_desc.value;
-    history = Creation (Value_creation V1_safety_check)
+    history = Creation (Value_creation V1_safety_check);
+    has_warned = false
   }
 
 (* CR layouts: Should we be doing more memoization here? *)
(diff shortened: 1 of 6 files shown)


ocaml/typing/jkind.mli
```diff
@@ -301,6 +301,10 @@ val has_imported_history : t -> bool
 
 val update_reason : t -> creation_reason -> t
 
+val with_warning : t -> t
```
goldfirere, line 304: Please comment these functions.

ocaml/typing/jkind_types.mli
```diff
@@ -126,6 +126,7 @@ type 'type_expr history =
         rhs_history : 'type_expr history
       }
   | Creation of Jkind_intf.History.creation_reason
+  | Warning of 'type_expr history
```
goldfirere, line 129: Though I know this is the design I described, I don't like it now that I see it implemented.

* The `update_reason` function is painful, having to work under `Warning`
* We never really care about `Warning`s in the middle of a tree.

So I think it would be better to have (leaving out the `'type_expr` parameter)

```ocaml
type history = (* just as it was *)
type t =
  { jkind : Jkind_desc.t;
    history : history;
    has_warned : bool
  }
```

Separating out the warning info -- and making it clear that it's just a bool, not (say) a count of how many times we've warned -- seems better.

Sorry to request a reimplementation here! Designs are sometimes hard to get right.

ocaml/typing/jkind.ml
```diff
@@ -1077,6 +1084,7 @@ end = struct
         fprintf ppf "@[<v 2>  %a@]@;%a@ @[<v 2>  %a@]" in_order lhs_history
           format_interact_reason reason in_order rhs_history
       | Creation c -> format_creation_reason ppf c
+      | Warning h -> fprintf ppf "warning @[<v 2>  %a@]" in_order h
```
goldfirere, line 1087: Interestingly, I don't think this output ever gets generated.

ocaml/typing/jkind.mli
dkalinichenko-js: Done.

ocaml/typing/jkind_types.mli
dkalinichenko-js: Done.

goldfirere (approved):

PR #2618, opened 2024-05-24
Fix issue with gadts and layouts

ccasin:
This fixes two bugs preventing pattern matching on a gadt couldn't from refining layouts in the expected way.

The first issue was in `Ctype.unify1_var`.  This function distinguishes failures that might be recoverable later due to a gadt equation from unrecoverable errors.  It incorrectly put layout checks into the latter category.

The second issue was in `Ctype.unify3`. During pattern unification, this function should allow for refining a `TConstr` with a gadt equation.  But in the case the `TConstr` was being unified with a variable, it did not.  The old behavior was correct before layouts, because in that case the variable can't tell us anything interesting about the constr.

I've added a few tests.  One is a demonstration that the old behavior was a soundness bug (because we would incorrectly refute a pattern match).

### DIFF
--- a/ocaml/typing/ctype.ml
+++ b/ocaml/typing/ctype.ml
@@ -3377,6 +3377,13 @@ let unify3_var env jkind1 t1' t2 t2' =
         record_equation t1' t2';
       end
 
+(* This is used to check whether we should add a gadt equation refining a
+   Tconstr's jkind during pattern unification. *)
+let constr_jkind_refinable env t jkind =
+  match unification_jkind_check env t jkind with
+  | () -> false
+  | exception Unify_trace _ -> true
+
 (*
    1. When unifying two non-abbreviated types, one type is made a link
       to the other. When unifying an abbreviated type with a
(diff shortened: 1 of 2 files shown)


lpw25 (approved):
LGTM

ocaml/typing/ctype.ml
```diff
@@ -3375,6 +3377,18 @@ let unify3_var env jkind1 t1' t2 t2' =
         record_equation t1' t2';
       end
 
+(* This is used to check whether we should add a gadt equation refining a
+   Tconstr's jkind during pattern unification. *)
+let constr_jkind_refinable env t jkind =
+  let snap = Btype.snapshot () in
+  let refinable =
+    match unification_jkind_check env t jkind with
```
goldfirere, line 3385: Using `unification_jkind_check` here seems like overkill. That uses the `lmode` to determine whether to perform the check or delay/skip it. But I think we will always want to perform the checks in a GADT match... delaying wouldn't even really make sense here.

And we also don't need the full power of `constrain_type_jkind`, which deals specially with type variables. `check_type_jkind` should be sufficient, given that the type will always be a `Tconstr`.

So I don't think this is wrong, but it's using a bigger hammer than necessary... and would be wrong if the big hammer were necessary (that is, this would do the wrong thing if either `lmode` weren't `Perform_checks` or if the type is a variable).

Maybe fix up in a future patch?

ccasin: I considered trying to create a simplified version of some of the functions involved here, but judged that the clarity that comes from uniformly using the same function for jkind checks in unification was preferable.

ccasin: In particular I think it's important for correctness that the check here matches the check in unify3_var.  As you note, that one is dealing with more than tconstrs, but this solution seemed the cleanest way to achieve that.

Similarly, the backtracking here is unnecessary work, at least for now, considering these constraints.  But Leo pointed out that it made this more obviously correct.

PR #2621, opened 2024-05-24
Reinstate, fix "Use `CU.Name.t` for name of .cmi; support import info for parameters"

lukemaurer:
Reverts ocaml-flambda/flambda-backend#2619 and adds a fix (originally put up as #2609) for artifact sizes.

### DIFF
--- a/backend/asmlink.ml
+++ b/backend/asmlink.ml
@@ -47,7 +47,7 @@ type unit_link_info = {
 
 (* Consistency check between interfaces and implementations *)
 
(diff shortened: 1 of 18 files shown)


mshinwell:
The artifact sizes are ok here (I manually tested, including an unrevert of #2585).  The commit here corresponding to #2609 needs review.  Then it's ok to merge this and an unrevert of #2585.

lpw25 (approved):
#2609 has now been reviewed, so I think this is good to go.

PR #2623, opened 2024-05-24
Small simplification to the new add_gadt_equation

goldfirere:
After I stared a bit at #2618 -- and I understood why it did what it did -- a seemingly-simpler approach sprung to mind.

In the process, I added a bunch of test cases just to increase my understanding.

Review: @ccasin

### DIFF
--- a/ocaml/typing/ctype.ml
+++ b/ocaml/typing/ctype.ml
@@ -3383,7 +3383,7 @@ let unify3_var env jkind1 t1' t2 t2' =
       if can_generate_equations () then begin
         begin match get_desc t2' with
         | Tconstr(path,[],_)
-          when is_instantiable !env ~for_jkind_eqn:true path ->
+          when is_instantiable !env ~for_jkind_eqn:false path ->
             add_gadt_equation env path t1'
               (* This is necessary because a failed kind-check above
                  might meaningfully refine a type constructor *)
--- a/ocaml/typing/ctype.ml
+++ b/ocaml/typing/ctype.ml
@@ -3353,18 +3353,27 @@ let unify1_var env t1 t2 =
 (* Called from unify3 *)
 let unify3_var env jkind1 t1' t2 t2' =
   occur_for Unify !env t1' t2;
+  (* There are two possible ways forward here. Either the variable [t1']
+     will succeed in unifying with [t2], in which case we're done; or
+     the unification fails, in which case we want to add a GADT equation.
+     But in that second case, we want to pretend that we were making a
+     GADT equation all along, so we don't want any evidence of having
+     attempted the unification (which might, say, fill in a sort variable).
+     We thus backtrack in the GADT equation case. *)
+  let snap = snapshot () in
   match
     occur_univar_for Unify !env t2;
     unification_jkind_check !env t2' jkind1
   with
   | () -> link_type t1' t2
   | exception Unify_trace _ when in_pattern_mode () ->
+      backtrack snap;
       reify env t1';
       reify env t2';
       if can_generate_equations () then begin
         begin match get_desc t2' with
         | Tconstr(path,[],_)
-          when is_instantiable !env ~for_jkind_eqn:false path ->
+          when is_instantiable !env ~for_jkind_eqn:true path ->
             add_gadt_equation env path t1'
               (* This is necessary because a failed kind-check above
                  might meaningfully refine a type constructor *)
--- a/ocaml/typing/ctype.ml
+++ b/ocaml/typing/ctype.ml
@@ -2875,7 +2875,7 @@ let non_aliasable p decl =
 
    2) It restricts gadt equations on pervasives, which a hacky implementation of
    the fact that every type in the initial environment is incompatible with
-   every other type.  But its fine to refine the jkind of these types (and we
+   every other type.  But it's fine to refine the jkind of these types (and we
    want to in some cases, like giving int32 the jkind immediate64 on
    appropriate platforms).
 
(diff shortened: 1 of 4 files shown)


ccasin:
I agree this is cleaner!

While working out what was going on here on Friday, we went back and forth about whether to put the special case in `unify3_var` as you've done here or as separate cases in `unify3` as we did in #2618.  One reason to like that approach, Leo argued, was that it's idiomatic within the style of the unify functions, which often have special cases early to prevent falling through to later cases.  I think this PR shows the cleanliness benefits of this approach outweigh that consideration.

But I do think this has a slightly different behavior than #2618, because it drops the backtracking around the jkind check.  Today, it's hard for a jkind check to change any state when it fails (though maybe not impossible, as we discussed in #1964?).  But when we add product layouts this will definitely be possible (and desirable, for the reasons discussed there).  Since we're here handling that failure with a gadt equation, I think we don't want to keep whatever mutation the failed jkind check may have done, and should keep the backtracking element of #2618.  (I should confess I also missed this consideration in my initial implementation of #2618, and Leo pointed out I should add it during in-person review)

goldfirere:
Good point about the backtracking. But doesn't that concern rise above just the special cases in #2618? That is, that patch adds special cases only for nullary type constructors. Yet the backtracking concern should cover all type constructors. So I think the solution would be to put the backtracking in `unify3_var`, undoing any changes in the `try` block in the case that a GADT equation is added, regardless of the arity of constructor. Bottom line here: if/when the lack of backtracking becomes bugsome, then #2618 will fix the bug only for nullary types, but the bug will exist for non-nullary types, too. My patch here doesn't solve the problem yet, but if we agree that it's worth continuing this patch, I can add the backtrack to `unify3_var`.

Regarding the style in `unify` around special cases: yes I broadly agree. But `unify3_var` already has something of a fall-through case for adding a GADT equation. The new story fits right in that fall-through -- it's just that the fall-through wasn't clever enough around nullary type constructors. Really, `record_equation` should call `add_gadt_equation` for any nullary constructors it gets.... it just so happens that every call other than the one in `unify3_var` already could be shown not to ever see nullary constructors.

ccasin:
> Good point about the backtracking. But doesn't that concern rise above just the special cases in #2618? That is, that patch adds special cases only for nullary type constructors. Yet the backtracking concern should cover all type constructors. So I think the solution would be to put the backtracking in `unify3_var`, undoing any changes in the `try` block in the case that a GADT equation is added, regardless of the arity of constructor. Bottom line here: if/when the lack of backtracking becomes bugsome, then #2618 will fix the bug only for nullary types, but the bug will exist for non-nullary types, too. My patch here doesn't solve the problem yet, but if we agree that it's worth continuing this patch, I can add the backtrack to `unify3_var`.

I think we're in agreement here - with the implementation in this PR you should backtrack in unify3_var whether we add a gadt equation or not.  I think this wasn't necessary previously because we'd always raise a type error if the layout check in `unify3_var` fails and we couldn't add a gadt equation.

As I was thinking about this some more, I realized that this PR also fixes a bug present in #2618.  In particular, that PR didn't completely fix the bug it thought it did - this is still allowed:

```ocaml
type ('a : any) is_value =
  | V : ('a : value) is_value

module A : sig
  type 'a t : any

  val is_value : 'a t is_value
end = struct
  type 'a t = int

  let is_value = V
end

let magic (type a) (x : a A.t is_value) : 'b =
  match x with
  | _ -> .

let not_so_good : 'b = magic A.is_value
```

Perhaps we should add that to the test suite.

It's also the case that some more reasonable lookging programs allowed by #2618 are rejected here, e.g.,:

```ocaml
type ('a : any) is_value =
  | V : ('a : value) is_value

type t : float64

let refute (x : t is_value) =
  match x with
  | _ -> .
```

But I will have to think a bit harder about whether we can get this back in the future.

goldfirere:
OK. I think this is in good shape now.

* I've added your test cases.
* I've added the backtracking.
* I added the tests above. It turns out that getting the second test working was as simple as setting `~for_jkind_eqn:true`.
* But I realized that the last test wouldn't work for non-nullary types, which is sad. You can see the test case in the second commit. I added some specialized code just for that case, and it worked, but I realized that there was still a little money on the table. So I ended up in `mcomp`, which was easy enough to patch up to look at kinds. You'll see it in the last commit. Probably didn't need to go this far today, but it cleared a CR.

ccasin:
Great!

But limitations on learning things about non-nullary type constructors from gadt matches aren't unique to layouts, as is being discussed elsewhere.  Without having read the new version of this in detail yet, I think it would be ideal to leave this PR doing the smaller thing of cleaning up / fixing #2618, and not worry about constructors with arguments, splitting the more complicated changes into their own PR.

(I predict I won't have time to thoroughly review this today regardless, unfortunately)

goldfirere:
There are two different things we might do with non-nullary type constructors:

* Check them in e.g. `mcomp`. This will allow us to say, given `type 'a t : float64` and `K : int32# t2` and `x : 'a t t2`, that a pattern-match on `K` is impossible.
* Remember the intersection jkinds. Given `type 'a t3 : non_null_value` and `K2 : ('a : immediate_or_null) t4`, this would allow us to match on `x : 'a t3 t4` to get `K2` and then use `'a t3` as if it were of jkind `non_null_immediate`. This would be good! But it is hard.

This patch is only about the first of these. I agree that improvements to `mcomp` could be in a separate PR. As of now, they're in a separate commit, and it seems lighterweight just to keep things the way they are. But I could restructure if you think that's best.

ocaml/typing/ctype.ml
```diff
@@ -3353,31 +3363,42 @@ let unify1_var env t1 t2 =
...
+  | () -> link_type t1' t2
   | exception Unify_trace _ when in_pattern_mode () ->
+      backtrack snap;
       reify env t1';
       reify env t2';
       if can_generate_equations () then begin
-        occur_univar ~inj_only:true !env t2';
+        begin match get_desc t2' with
+        | Tconstr(path,[],_)
+          when is_instantiable !env ~for_jkind_eqn:true path ->
```
ccasin, line 3386: I think this should be `for_jkind_eqn:false`.  You mention you needed this, but it doesn't seem to be needed by the current testsuite, and I think it is wrong: it could result in adding a gadt equation for a built-in type of one defined in the current module, which we otherwise don't allow.  As is discussed in the long comment on `is_instantiable`, it's possible that restriction could one day be dropped with some additional work elsewhere, but I don't see any reason for this one call site to `add_gadt_equation` to be different.

ccasin (approved):
Eagerly approving, but a few questions to resolve before merging.

ocaml/typing/ctype.ml
goldfirere: Yes. I forget why I changed that. It _was_ necessary, but maybe is no longer now that mcomp is fixed. Perhaps that's what inspired me to explore mcomp. Not worth the time spelunking through all that, though.

PR #2624, opened 2024-05-24
Combine Value and Alloc modules

goldfirere:
The `Value` and `Alloc` modules in mode.ml seemed very similar. It was not hard to combine them, losing almost 400 lines of code.

Review: @riaqn.

### DIFF
--- a/ocaml/typing/mode.ml
+++ b/ocaml/typing/mode.ml
@@ -1279,6 +1279,8 @@ module Regionality = struct
   let global = of_const Const.Global
 
   let legacy = of_const Const.legacy
+
+  let zap_to_legacy = zap_to_floor
 end
 
 module Linearity = struct
@@ -1336,7 +1338,11 @@ let regional_to_global m =
   S.Positive.via_monotone Locality.Obj.obj C.Regional_to_global m
 
 module Comonadic_with_regionality = struct
-  module Const = C.Comonadic_with_regionality
+  module Const = struct
+    include C.Comonadic_with_regionality
+
+    let eq a b = le a b && le b a
+  end
 
   module Obj = struct
     type const = Const.t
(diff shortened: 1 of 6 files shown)


ocaml/typing/mode.ml
```diff
@@ -1335,68 +1341,19 @@ let locality_as_regionality m =
...
-
-  let join_with ax c m = join_const (C.min_with Obj.obj ax c) m
+module type Areality = sig
+  module Const : C.Areality
 
-  let meet_with ax c m = meet_const (C.max_with Obj.obj ax c) m
+  module Obj : Obj with type const = Const.t and module Solver = S.Positive
 
-  let imply c m = Solver.via_monotone obj (Imply c) (Solver.disallow_left m)
+  val comonadic_with_areality : Const.t C.comonadic_with C.obj
```
riaqn, line 1349: Instead of defining this, I think it's more canonical to use the `comonadic_with_obj` function.

ocaml/typing/mode.ml
```diff
@@ -1533,8 +1491,8 @@ type ('mo, 'como) monadic_comonadic =
     comonadic : 'como
   }
 
-module Value = struct
-  module Comonadic = Comonadic_with_regionality
+module Mode (Areality : Areality) = struct
```
riaqn, line 1494: I don't have a strong opinion, but it might be worthwhile to think of a more descriptive name than `Mode`. Maybe `Value_with`? Similar to `Comonadic_with`.

The `Alloc` mode is a bad name anyway, since it's more about "a dumbed down version of `Value` exposed to the user" than about "allocation".

riaqn (approved):
This is great! Some nits and then it's good to go.

ocaml/typing/mode.ml
goldfirere: Good suggestion.

I'll leave renaming `Alloc` for another PR, but I'd support it.

PR #2625, opened 2024-05-27
Fexpr: symbols are dominator-scoped

Ekdohibs:

PR #2626, opened 2024-05-27
Data flow: fix a fatal error when the exception param is aliased

Ekdohibs:
Like #2625, this was extracted out of #2295, as it fixed a bug that was independent of the PR.


chambart (approved):
Ok

PR #2632, opened 2024-05-28
Revert "Revert "Refactor `Persistent_env` to separate imports from bound names""

lukemaurer:
Reverts ocaml-flambda/flambda-backend#2611


lpw25 (approved):
Approving on the basis of the review of the original patch.

PR #2633, opened 2024-05-29
Non-null objects and misc

dkalinichenko-js:
Mark `lazy_t` as non-null: we have not figured out the runtime interactions between `lazy_t` and `or_null` yet, but this should be our design goal for backward compatibility.

Remove `Class_let_binding` from the list of value creation reasons -- it is never used.

Make objects non-null. To achieve this, I also move `Tfield` and `Tnil` to `non_null_value_creation_reason`.

`generalize_class_signature_spine` in `ctype.ml` had this code for checking class methods:

```
  (* But keep levels correct on the type of self *)
  Meths.iter
    (fun _ (_, _, ty) ->
       unify_var env (newvar (Jkind.value ~why:Object)) ty)
    meths;
  sign.csig_meths <- new_meths
  ```
I had to change `~why:` to `Object_fields` to let objects have nullable fields. I assume `Object_fields` was intended all along.


dkalinichenko-js:
Closing since we're reimplementing nullability.

PR #2634, opened 2024-05-29
Rename modes (only)

goldfirere:
Though our description of modes in our docs is out of date in a number of ways, I at least wanted to implement the renaming, as I'm also updating nearby docs on kinds and `or_null`.

Perhaps @lpw25 can just confirm that my new names are correct.


lpw25 (approved):
Renamings are correct

PR #2635, opened 2024-05-29
Fix `--enable-dev`

dkalinichenko-js:
Fix compilation of `systhreads4` with `--enable-dev` set. The dev build profile includes the `-strict-sequence` flag by default, which we disable with `-no-strict-sequence`.

Tested manually with `--enable-dev`.

Also reformat the dune file to be consistent with VSCode formatting.

### DIFF
--- a/ocaml/otherlibs/systhreads4/byte/dune
+++ b/ocaml/otherlibs/systhreads4/byte/dune
@@ -1,19 +1,21 @@
 (copy_files# ../*.{c,h,ml,mli})
 
 (library
-  (name threads)
-  (modes byte)
-  (wrapped false)
-  ; FIXME Fix warning 27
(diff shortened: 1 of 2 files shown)


ncik-roberts (approved):
Thanks, I tested and it works for me.

As a small note, the formatting changes make the GitHub diff view all but useless. It would have aided review to split those into another PR or to include a command I could run by hand to see the whitespace-sensitive diff. It's actually just:

```diff
-  ; FIXME Fix warning 27
+ ; FIXME Fix warning 27 and -no-strict-sequence
  (flags
-   (:standard -g -bin-annot -safe-string -w -27))
+  (:standard -no-strict-sequence -g -bin-annot -safe-string -w -27))
```
(in both files).

PR #2636, opened 2024-05-29
Add `Obj` functions for accessing the scannable prefix len of a mixed block

ncik-roberts:
This PR makes it possible to update generic OCaml traversals to also work for mixed blocks.

I add enough of an API for people to use this without allocating if they wish.


ncik-roberts:
This isn't quite ready for review as it doesn't handle bytecode correctly. Will reopen soon.

ncik-roberts:
It's now ready for review. I nominate @ccasin. There are runtime changes, but they don't require that much context.

mshinwell:
I can look at the runtime changes.

ocaml/stdlib/obj.mli
```diff
@@ -172,3 +172,35 @@ module Ephemeron: sig
   (** Maximum length of an ephemeron, ie the maximum number of keys an
       ephemeron could contain *)
 end
+
+module Uniform_or_mixed : sig
+  (** Blocks with a nominally scannable tag can still have a suffix of
+      unscanned objects; such a block is "mixed". This contrasts with
+      "uniform" blocks which are either all-scanned or all-unscanned.
+  *)
```
ccasin, line 180: Maybe it's worth noting somewhere in here that these functions can return different results on nativecode and bytecode, which might be a surprising source of test failures to someone who hasn't thought about the bytecode representation?

ccasin (approved):
Looks good.  One small suggestion, which you can feel free to ignore if you disagree.

ocaml/stdlib/obj.mli
ncik-roberts: Added a comment about this, thanks.

PR #2639, opened 2024-05-30
Add a dev-mode CI

ncik-roberts:
Adds a CI action that builds the compiler and runs tests with `--enable-dev`.

This checks that a PR doesn't break our local build configuration.


ncik-roberts:
This adds a new job that takes 7m 21s to run, which is less than any of the other build/run jobs, which are generally above 10m. This seems like a good use of a CI slot to me but I'll let @mshinwell weigh in.

mshinwell (approved):

PR #2641, opened 2024-05-30
Fix `tests/` with `--enable-dev`

TheNumbat:
Configuring with `--enable-dev` turns on strict sequence and various compiler warnings, causing these tests to fail.

### DIFF
--- a/tests/simd/arrays.ml
+++ b/tests/simd/arrays.ml
@@ -1,5 +1,6 @@
-
 open Stdlib
+
+[@@@ocaml.warning "-37-32-35"]
 
 external int8x16_of_int64s : int64 -> int64 -> int8x16 = "" "vec128_of_int64s" [@@noalloc] [@@unboxed]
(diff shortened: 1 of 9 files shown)


tests/simd/arrays.ml
```diff
@@ -1,6 +1,7 @@
-
 open Stdlib
 
+[@@@ocaml.warning "-37-32-35"]
```
ncik-roberts, line 3: * Can we just fix the unused for-loop indexes? That seems relatively less invasive than the other ones.
* I suggest `[@@@ocaml.warning "-unused-constructor" "-unused-value-declaration"]` as it's a bit more self-documenting.

ncik-roberts (approved):

tests/simd/arrays.ml
TheNumbat: Sure, done. Though it seems `[@@@ocaml.warning]` can only take one full name at a time.

ncik-roberts:
Thanks! Let's merge when CI turns green.

PR #2642, opened 2024-05-31
Add [Deleted_code] case for code ids in set of closures.

Ekdohibs:
This is a prerequisite for #2213. It adds a way to specify that some code ids in a set of closures have been deleted, and the closure therefore does not need a code pointer (but the function slot is still used for projections, and the closure still exists). For now, we do not change the layout and simply put a zero instead of the code id; in the future we can make slot_offsets share the slot as much as possible.

### DIFF
--- a/middle_end/flambda2/parser/flambda_to_fexpr.ml
+++ b/middle_end/flambda2/parser/flambda_to_fexpr.ml
@@ -674,7 +674,7 @@ let set_of_closures env sc =
       (Set_of_closures.function_decls sc
       |> Function_declarations.funs_in_order
       |> Function_slot.Lmap.map (function
-           | Function_declarations.Deleted -> Misc.fatal_error "todo"
+           | Function_declarations.Deleted _ -> Misc.fatal_error "todo"
            | Function_declarations.Code_id code_id -> code_id)
       |> Function_slot.Lmap.bindings)
   in
--- a/middle_end/flambda2/parser/flambda_to_fexpr.ml
+++ b/middle_end/flambda2/parser/flambda_to_fexpr.ml
@@ -672,7 +672,11 @@ let set_of_closures env sc =
       (fun (function_slot, fun_decl) ->
         function_declaration env fun_decl function_slot alloc)
       (Set_of_closures.function_decls sc
-      |> Function_declarations.funs_in_order |> Function_slot.Lmap.bindings)
+      |> Function_declarations.funs_in_order
+      |> Function_slot.Lmap.map (function
+           | Function_declarations.Deleted -> Misc.fatal_error "todo"
+           | Function_declarations.Code_id code_id -> code_id)
+      |> Function_slot.Lmap.bindings)
   in
   let elts = value_slots env (Set_of_closures.value_slots sc) in
   let elts = match elts with [] -> None | _ -> Some elts in
--- a/middle_end/flambda2/simplify_shared/slot_offsets.ml
+++ b/middle_end/flambda2/simplify_shared/slot_offsets.ml
@@ -752,7 +752,8 @@ end = struct
 
   (* Create slots (and create the cross-referencing). *)
 
-  let create_function_slot set state get_code_metadata function_slot (code_id : Function_declarations.code_id_in_function_declaration) =
+  let create_function_slot set state get_code_metadata function_slot
+      (code_id : Function_declarations.code_id_in_function_declaration) =
     if Compilation_unit.is_current
          (Function_slot.get_compilation_unit function_slot)
     then (
--- a/middle_end/flambda2/simplify_shared/slot_offsets.ml
+++ b/middle_end/flambda2/simplify_shared/slot_offsets.ml
@@ -752,14 +752,14 @@ end = struct
 
   (* Create slots (and create the cross-referencing). *)
 
-  let create_function_slot set state get_code_metadata function_slot code_id =
+  let create_function_slot set state get_code_metadata function_slot (code_id : Function_declarations.code_id_in_function_declaration) =
     if Compilation_unit.is_current
          (Function_slot.get_compilation_unit function_slot)
     then (
       let size =
         match code_id with
-        | None -> 2
-        | Some code_id ->
+        | Deleted { function_slot_size } -> function_slot_size
+        | Code_id code_id ->
           let code_metadata = get_code_metadata code_id in
           let module CM = Code_metadata in
           let is_tupled = CM.is_tupled code_metadata in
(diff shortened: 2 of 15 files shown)


middle_end/flambda2/simplify_shared/slot_offsets.ml
```diff
@@ -853,15 +889,22 @@ end = struct
...
             Function_slot.Map.find_opt function_slot state.function_slots
           with
           | None ->
             create_function_slot set state get_code_metadata function_slot
               code_id
-          | Some s -> s
+          | Some s ->
+            s.sets <- set :: s.sets;
+            update_set_for_slot s set;
+            s
```
lthls, line 907: Could you explain what this patch does (and the same patches later for value slots) ? At first glance it looks unrelated to the rest of the PR

Ekdohibs: It's indeed unrelated to the rest of the PR, and is a fix for a "how-did-it-ever-work-before" kind of bug: previously, slots were only considered to be part of the first set of closures they were seen in. Currently, it works because two set of closures which have a slot in common have exactly the same slots. Maybe I should open another PR with those changes?

lthls: Do you think you could write an fexpr program that tests this ? We can also wait for the global pass, which will make it easier to write such programs in OCaml directly.

Ekdohibs: Now merged as a separate PR.

middle_end/flambda2/parser/flambda_to_fexpr.ml
```diff
@@ -672,7 +672,11 @@ let set_of_closures env sc =
       (fun (function_slot, fun_decl) ->
         function_declaration env fun_decl function_slot alloc)
       (Set_of_closures.function_decls sc
-      |> Function_declarations.funs_in_order |> Function_slot.Lmap.bindings)
+      |> Function_declarations.funs_in_order
+      |> Function_slot.Lmap.map (function
+           | Function_declarations.Deleted _ -> Misc.fatal_error "todo"
```
Gbury, line 677: is that a todo for another PR or for alter this PR ?

middle_end/flambda2/to_cmm/to_cmm_set_of_closures.ml
```diff
@@ -224,40 +230,85 @@ end = struct
...
+          get_func_decl_params_arity env code_id
+        in
+        let closure_info =
+          C.closure_info' ~arity:(kind, params_ty)
+            ~startenv:(startenv - slot_offset) ~is_last:last_function_slot
+        in
+        (* We build here the **reverse** list of fields for the function slot *)
+        match closure_code_pointers with
+        | Full_application_only ->
+          if size < 2
```
Gbury, line 248: I don't understand this change: particularly, why could it happen that `size = 3` ? and why it is correct for such function slots to exist ?

middle_end/flambda2/to_cmm/to_cmm_set_of_closures.ml
```diff
@@ -203,18 +203,24 @@ end = struct
...
         updates )
-    | Function_slot { size; function_slot; last_function_slot } -> (
-      let code_id = Function_slot.Map.find function_slot decls in
-      let code_symbol =
-        R.symbol_of_code_id res code_id ~currently_in_inlined_body:false
-      in
-      let (kind, params_ty, result_ty), closure_code_pointers, dbg =
-        get_func_decl_params_arity env code_id
-      in
+    | Dummy_function_slot { last_function_slot } ->
```
Gbury, line 206: We're getting at the point in code size (no pun intended) where some of the cases of the matches below could be split into their own function.

Gbury (changes requested):

middle_end/flambda2/parser/flambda_to_fexpr.ml
Ekdohibs: Most likely for another PR, there are many other missing features in fexpr so it would make sense to make a single PR to add them all.

middle_end/flambda2/to_cmm/to_cmm_set_of_closures.ml
Ekdohibs: This was a consequence of a previous version when deleted slot offsets had no size; I've removed it.

middle_end/flambda2/terms/cost_metrics.ml
```diff
@@ -70,13 +70,18 @@ let set_of_closures ~find_code_characteristics set_of_closures =
...
-          find_code_characteristics code_id
-        in
-        ( metrics + cost_metrics,
-          (* CR poechsel: valid until OCaml 4.12, as for named_size *)
-          Stdlib.( + ) num_words (if params_arity <= 1 then 2 else 3) ))
+      (fun _ (code_id : Function_declarations.code_id_in_function_declaration)
+           (metrics, num_words) ->
+        match code_id with
+        | Deleted { function_slot_size } ->
+          metrics, Stdlib.( + ) num_words function_slot_size (* XXX *)
```
lthls, line 77: Why the `(* XXX *)` comment ? This looks as correct as the code below (i.e. slightly wrong because it ignores the infix tag).

middle_end/flambda2/to_cmm/to_cmm_set_of_closures.ml
```diff
@@ -203,18 +203,24 @@ end = struct
...
       let closure_info =
-        C.closure_info' ~arity:(kind, params_ty)
+        C.closure_info'
+          ~arity:(Curried { nlocal = 0 }, [()])
```
lthls, line 209:
```suggestion
        C.closure_info
          ~arity:1
```

middle_end/flambda2/to_cmm/to_cmm_set_of_closures.ml
```diff
@@ -224,40 +230,78 @@ end = struct
...
-            "fill_slot: Function slot %a is of size %d, but it is used to \
-             store code ID %a which is classified as \
-             Full_and_partial_application (so the expected size is 3)"
-            Function_slot.print function_slot size Code_id.print code_id;
+            "fill_slot: Function slot %a is of size %d, but it is said to be \
+             deleted of size %d"
+            Function_slot.print function_slot size function_slot_size;
+        let closure_info =
+          C.closure_info'
+            ~arity:(Curried { nlocal = 0 }, if size = 2 then [()] else [(); ()])
```
lthls, line 296:
```suggestion
          C.closure_info
            ~arity:(if size = 2 then 1 else 2)
```

lthls (approved):
Looks good to me

PR #2643, opened 2024-05-31
Cancel running CI if new commits are pushed to a pull request

Ekdohibs:
So that we don't have to wait too long for the CI to run.
See https://docs.github.com/en/actions/using-jobs/using-concurrency.


mshinwell:
This is probably fine but I'd like to discuss more widely first.

mshinwell (approved):

PR #700, opened 2022-06-24
Review slot offsets

Gbury:
Done with @chambart . Best reviewed commit by commit.

PR #2644, opened 2024-06-03
Remove leftover reference in Slot_offsets

lthls:
Apparently #700 (and #706) removed all assignments to a reference but left the reference itself; this PR removes it (the state itself is mutable, and updated by the relevant functions).


chambart (approved):
:wave: poor reference

PR #2645, opened 2024-06-03
Eliminate fallthrough blocks at the end of Cfgize

gretay-js:
This PR enables an existing pass `Eliminate_fallthrough_blocks` to run at the end of `Cfgize`.  This is minor improvement to code generation (avoid an unnecessary branch) at the cost of increased compilation time. A follow up PR will clean it up to reduce the overhead.

### DIFF
--- a/backend/cfg/cfgize.ml
+++ b/backend/cfg/cfgize.ml
@@ -845,6 +845,8 @@ let fundecl :
      terminator simplification. *)
   Profile.record ~accumulate:true "optimizations"
     (fun () ->
+      if simplify_terminators
+      then Eliminate_fallthrough_blocks.run cfg_with_layout;
       if simplify_terminators then Merge_straightline_blocks.run cfg_with_layout;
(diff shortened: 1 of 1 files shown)


xclerc:
I am a bit surprised it makes a difference
to apply this optimization at this point.
Was the difference observed only with
upstream's allocators?

mshinwell (approved):

gretay-js:
> I am a bit surprised it makes a difference to apply this optimization at this point. Was the difference observed only with upstream's allocators?

yes, this is only useful for the upstream allocators. When using CFG-based register allocation, the example did have extra branches. As discussed offline, `Regalloc_utils.simplify_cfg`  call at the end of the CFG regalloc pipeline already calls `Eliminate_fallthrough_blocks`.  `Cfg_to_linear` tries to avoid extra branches for a given layout but it only checks locally, so won't be able to identify and eliminate fallthrough blocks.

PR #2646, opened 2024-06-03
Mode crossing for unboxed integers

goldfirere:
Mode-crossing uniqueness doesn't work, but it doesn't for `int` either (at least, not as I've tested it). This is an independent bug which I claim should not hold up this PR.


ncik-roberts (approved):
I didn't know how to make this change before, and now I do. (Also, it looks obviously correct, and the tests support that.) Thanks!

riaqn:
Yeah mode-crossing uniqueness will require changes in the uniqueness analysis. This is tracked by #2377  (I won't get back to that for a while).

PR #2647, opened 2024-06-03
Add Assert_mixed_block_layout_v1 macro

ncik-roberts:
Define a macro that user code can write to assert that the surrounding C code is sensitive to the layout of mixed blocks (in a way that may change in a future release).


mshinwell (approved):

mshinwell:
This is fine, do you want to do this for `runtime4` too?  If so consider that pre-approved.

ncik-roberts:
Yes, I forgot to do that. I just pushed the change, and I'll merge when CI passes.

PR #2649, opened 2024-06-04
Fix a bug in slot_offsets when two separate sets of closures have a slot in common, but not all

Ekdohibs:
Extracted from #2642.

### DIFF
--- a/middle_end/flambda2/simplify_shared/slot_offsets.ml
+++ b/middle_end/flambda2/simplify_shared/slot_offsets.ml
@@ -861,7 +861,10 @@ end = struct
           | None ->
             create_function_slot set state get_code_metadata function_slot
               code_id
-          | Some s -> s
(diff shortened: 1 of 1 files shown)


Ekdohibs:
I'm not sure how to add a test, given it's something that isn't generated by the compiler as of now, and this can't be tested by flexpect either.

Ekdohibs:
So, this fixes a bug, but there are other remaining. Indeed, we cannot put space between function slots in set of closures, as `Gc.compact` will scan that area (especially the arity). This will require more changes, the scope of which I am not sure about yet.

Gbury (approved):

Gbury:
I'll look at the problem with consecutive slots as soon as I have the time.

Ekdohibs:
The consecutive slots problem has been addressed in #2674.

PR #2650, opened 2024-06-04
Simple cmov insertion

mshinwell:
This causes conditional moves to be used rather than branches for `Cifthenelse (cond, ifso, ifnot)` where `ifso` and `ifnot` are integer-valued, small and pure Cmm expressions.  If both such expressions also look as if they finish by tagging an integer, the tagging is moved out so it happens after the conditional move.

For example:
```
type t = A | B

let foo t (x : int) (y : int) =
  match t with
  | A -> x > y
  | B -> x < y
```

Before:
```
camlK__foo_0_1_code:
        cmpq    $1, %rax
        jne     .L103
        cmpq    %rdi, %rbx
        setg    %al
        movzbq  %al, %rax
        leaq    1(%rax,%rax), %rax
        ret
        .align  4
.L103:
        cmpq    %rdi, %rbx
        setl    %al
        movzbq  %al, %rax
        leaq    1(%rax,%rax), %rax
        ret
```

With this PR:
```
camlK.foo_0_1_code:
        movq    %rax, %rsi
        cmpq    %rdi, %rbx
        setl    %al
        movzbq  %al, %rdx
        cmpq    %rdi, %rbx
        setg    %al
        movzbq  %al, %rax
        cmpq    $1, %rsi
        cmove   %rax, %rdx
        leaq    1(%rdx,%rdx), %rax
        ret
```

### DIFF
--- a/backend/cmm_helpers.ml
+++ b/backend/cmm_helpers.ml
@@ -3618,8 +3618,79 @@ let sequence x y =
   | _, Ctuple [] -> x
   | _, _ -> Csequence (x, y)
 
+let rec is_small_and_pure_int_expr ~size_bias ~size ~toplevel expr =
+  if size > 12 + size_bias
+  then None
+  else
+    match expr with
+    | Cconst_int _ | Cconst_natint _ | Cconst_symbol _ -> Some (size + 1)
+    | Cvar _ ->
+      (* We can't allow [Cvar] at toplevel as it might not be of kind Value. *)
+      (* CR mshinwell: we could maybe make [ite] take the kind from [To_cmm_expr]. *)
+      if toplevel then None else Some (size + 1)
+    | Cop (op, args, _) ->
+      if not (op_is_small_and_pure_int_expr op)
+      then None
+      else
+        List.fold_left
+          (fun size_opt arg ->
+            match size_opt with
+            | None -> None
+            | Some size -> (
+              match
+                is_small_and_pure_int_expr ~size_bias ~size ~toplevel:false arg
+              with
+              | None -> None
+              | Some size' -> Some (size + size')))
+          (Some size) args
+    | Cconst_float32 _ | Cconst_float _ | Cconst_vec128 _ | Clet _ | Clet_mut _
+    | Cphantom_let _ | Cassign _ | Ctuple _ | Csequence _ | Cifthenelse _
+    | Cswitch _ | Ccatch _ | Cexit _ | Ctrywith _ ->
+      None
+
+and op_is_small_and_pure_int_expr op =
+  match op with
+  | Caddi | Csubi | Cmuli | Cmulhi _ | Cand | Cor | Cxor | Clsl | Clsr | Casr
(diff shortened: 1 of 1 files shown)


backend/cmm_helpers.ml
```diff
@@ -3618,8 +3618,79 @@ let sequence x y =
...
+              with
+              | None -> None
+              | Some size' -> Some (size + size')))
+          (Some size) args
+    | Cconst_float32 _ | Cconst_float _ | Cconst_vec128 _ | Clet _ | Clet_mut _
+    | Cphantom_let _ | Cassign _ | Ctuple _ | Csequence _ | Cifthenelse _
+    | Cswitch _ | Ccatch _ | Cexit _ | Ctrywith _ ->
+      None
+
+and op_is_small_and_pure_int_expr op =
```
xclerc, line 3651:
```suggestion
let op_is_small_and_pure_int_expr op =
```

backend/cmm_helpers.ml
```diff
@@ -3618,8 +3618,79 @@ let sequence x y =
...
+  let can_untag_arms = can_untag then_ && can_untag else_ in
+  let size_bias = if can_untag_arms then -2 else 0 in
+  if is_small_and_pure_int_expr ~size_bias then_
+     && is_small_and_pure_int_expr ~size_bias else_
+  then
+    if can_untag_arms
+    then
+      tag_int
+        (Cop
+           ( Ccsel typ_val,
```
xclerc, line 3688:
```suggestion
           ( Ccsel typ_int,
```

backend/cmm_helpers.ml
```diff
@@ -3618,8 +3618,79 @@ let sequence x y =
   | _, Ctuple [] -> x
   | _, _ -> Csequence (x, y)
 
+let rec is_small_and_pure_int_expr ~size_bias ~size ~toplevel expr =
+  if size > 12 - size_bias
+  then None
+  else
+    match expr with
+    | Cconst_int _ | Cconst_natint _ | Cconst_symbol _ -> Some (size + 1)
```
xclerc, line 3626: (Not completely sure about `Cconst_symbol`,
or maybe it should be considered bigger.)

backend/cmm_helpers.ml
```diff
@@ -3618,8 +3618,79 @@ let sequence x y =
   | _, Ctuple [] -> x
   | _, _ -> Csequence (x, y)
 
+let rec is_small_and_pure_int_expr ~size_bias ~size ~toplevel expr =
+  if size > 12 - size_bias
```
xclerc, line 3622:
```suggestion
  if size + size_bias > 12
```
(Maybe easier to read)

xclerc (commented):
I am not 100% sure how the size
computation should be intepreted:
it feels a bit strange that operations
do not count towards the size, and
that we can return `Some _` for a
value above the constant (it is not
a simple threshold as we have in
other places).

I also wonder whether the rewrite
can make things a bit worse in some
corner cases, like if a spilled variable
needs to be reloaded from the stack.

mshinwell:
clang generates cmov for the above example.  Will investigate the others.

PR #2651, opened 2024-06-04
Do not demote direct calls whose callee is a symbol

mshinwell:
This should help remove unnecessary indirect calls when `.cmx` files become unavailable, as discussed.

cc @gretay-js


mshinwell:
We're going to take a slightly different approach, another PR will be opened.

PR #2653, opened 2024-06-04
Add new `Allocated_*` macros for accessing the size of an object

ncik-roberts:
Add new `Allocated_*` macros for accessing the size of an object, to be used in preference to existing macro names.

Hide the old names if a macro `Hide_upstream_size_macros` is defined.

### Review
Here's a command you can run to check to see whether I've copy-and-pasted the old definitions correctly:

```bash
# runtime 5
$ diff \
  <(git show 40493852e15a0d00efd41b71640e92ae7c093987:ocaml/runtime/caml/mlvalues.h | grep '#define [A-Z].size') \
  <(cat ocaml/runtime/caml/mlvalues.h | grep '#define Allocated_' | sed 's/Allocated_\(.\)/\U\1/g')  \
  --context

# runtime 4
$ diff \
  <(git show 40493852e15a0d00efd41b71640e92ae7c093987:ocaml/runtime4/caml/mlvalues.h | grep '#define [A-Z].size') \
  <(cat ocaml/runtime4/caml/mlvalues.h | grep '#define Allocated_' | sed 's/Allocated_\(.\)/\U\1/g')  \
  --context
```
(and then same thing for `runtime4`)

The first argument to `diff` finds macro definitions with "size" in the name prior to this PR; the second argument finds `Allocated_*` macro definitions in this PR, rewriting them to remove "Allocated_". The diff just identifies some macros `Whsize_wosize`, etc., that are moved around in this PR but not deleted.

### DIFF
--- a/ocaml/runtime/caml/mlvalues.h
+++ b/ocaml/runtime/caml/mlvalues.h
@@ -130,7 +130,7 @@ mixed blocks. In the upstream compiler, R is set with the
 
 #define Tag_hd(hd) ((tag_t) ((hd) & HEADER_TAG_MASK))
 #define Hd_with_tag(hd, tag) (((hd) &~ HEADER_TAG_MASK) | (tag))
-#define Wosize_hd(hd) ((mlsize_t) (((hd) & HEADER_WOSIZE_MASK) \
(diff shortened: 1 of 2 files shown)


ncik-roberts:
@mshinwell This is ready for review. I've made the changes we discussed.

mshinwell (approved):

PR #2655, opened 2024-06-04
Fix `--enable-dev` in runtime 5

ncik-roberts:
Mirrors #2635 , but for `systhreads` instead of only `systhreads4`. Also, add a CI job for checking that `--enable-dev` works with runtime 5. (#2639 was just for runtime 4.)


dkalinichenko-js (approved):
LGTM, thanks for testing this.

PR #2656, opened 2024-06-04
Refactor `check_representable` in `typedecl.ml`

dkalinichenko-js:
Call `check_representable` before type variables in record and variant declarations get generalized. This will allow us to initialize more type variables to `any` in later PRs. `Ctype.type_sort` will now lower their layouts to sort variables instead of raising an error. Note that we now only call `Jkind.Sort.get_default_value` if `allow_unboxed = false`, since we might actually default sort variables now.

Call `check_representable` in `transl_labels` and `transl_types_gf`, instead of  `update_` functions. `allow_unboxed` is set to `false` for variants and records with `[@@unboxed]` annotations, and for inlined records. It is set to `true` otherwise.

We now accept `type ('a : any) any4 = Any4 of 'a`, lowering `'a` to `value`. This is confusing, however @goldfirere says that this is in line with existing OCaml behavior.

### DIFF
--- a/ocaml/typing/typedecl.ml
+++ b/ocaml/typing/typedecl.ml
@@ -381,31 +381,32 @@ let set_private_row env loc p decl =
   in
   set_type_desc rv (Tconstr (p, decl.type_params, ref Mnil))
 
-(* Makes sure a type is representable.  Will lower [any] to a sort variable
-   if [allow_unboxed = true], and to [value] if [allow_unboxed = false]. *)
+(* Makes sure a type is representable. When called with a type variable, will
+   lower [any] to a sort variable if [allow_unboxed = true], and to [value]
+   if [allow_unboxed = false]. *)
 (* CR layouts: Many places where [check_representable] is called in this file
    should be replaced with checks at the places where values of those types are
    constructed.  We've been conservative here in the first version. This is the
    same issue as with arrows. *)
-   let check_representable ~why ~allow_unboxed env loc kloc typ =
-    match Ctype.type_sort ~why env typ with
-    (* CR layouts v5: This is a convenient place to rule out non-value types in
-       structures that don't support them yet. (A callsite passes
-       [~allow_unboxed:true] to indicate that non-value types are allowed.)
-       When we support mixed blocks everywhere, this [check_representable]
-       will have outlived its usefulness and we can delete it.
-    *)
-    (* CR layouts v2.5: This rules out non-value types in [@@unboxed] types. No
-       real need to rule that out - I just haven't had time to write tests for it
-       yet. *)
-    | Ok s -> begin
-      if not allow_unboxed then
-        match Jkind.Sort.get_default_value s with
-        | Void | Value -> ()
-        | Float64 | Float32 | Word | Bits32 | Bits64 as const ->
-          raise (Error (loc, Invalid_jkind_in_block (typ, const, kloc)))
-      end
-    | Error err -> raise (Error (loc,Jkind_sort {kloc; typ; err}))
+let check_representable ~why ~allow_unboxed env loc kloc typ =
+  match Ctype.type_sort ~why env typ with
+  (* CR layouts v5: This is a convenient place to rule out non-value types in
+      structures that don't support them yet. (A callsite passes
+      [~allow_unboxed:true] to indicate that non-value types are allowed.)
+      When we support mixed blocks everywhere, this [check_representable]
+      will have outlived its usefulness and we can delete it.
+  *)
+  (* CR layouts v2.5: This rules out non-value types in [@@unboxed] types. No
+      real need to rule that out - I just haven't had time to write tests for it
+      yet. *)
+  | Ok s -> begin
+    if not allow_unboxed then
+      match Jkind.Sort.get_default_value s with
+      | Void | Value -> ()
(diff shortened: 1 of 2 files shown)


ocaml/typing/typedecl.ml
```diff
@@ -435,24 +463,37 @@ let transl_labels ~new_var_jkind env univars closed lbls =
...
+  let mk idx arg =
     let cty = transl_simple_type ~new_var_jkind env ?univars ~closed Mode.Alloc.Const.legacy arg in
     let gf = Typemode.transl_global_flags
       (Jane_syntax.Mode_expr.of_attrs arg.ptyp_attributes |> fst) in
+    check_representable ~why:(Constructor_declaration idx) ~allow_unboxed
+      env loc kloc cty.ctyp_type;
     (cty, gf)
   in
-  let tyl_gfl = List.map mk tyl in
+  let tyl_gfl = List.mapi mk tyl in
```
goldfirere, line 476: `transl_labels` does the representability check in the second pass, while `transl_types_gf` here does it in the first. Is there a reason for this difference? The change might affect error messages somehow, if there is both an ordinary type error (e.g. `bool int`, because `int` is nullary) and a representability error in different arguments/labels. Probably both functions should do the same thing... and my guess is that doing this in the second pass is better, to prioritize more "normal" type errors.

goldfirere: Might be good to add tests that witness the difference.

ocaml/typing/typedecl.ml
```diff
@@ -435,24 +463,37 @@ let transl_labels ~new_var_jkind env univars closed lbls =
...
+           non-value, see comment in [check_representable]. *)
+        transl_types_gf ~new_var_jkind ~allow_unboxed:(not unboxed)
+        env loc univars closed l (Cstr_tuple { unboxed })
+      in
+      Types.Cstr_tuple flds', Cstr_tuple flds
   | Pcstr_record l ->
-      let lbls, lbls' = transl_labels ~new_var_jkind env univars closed l in
+      let lbls, lbls' =
+        (* CR layouts: we forbid fields of inlined records from being
+           non-value, see comment in [check_representable]. *)
```
goldfirere, line 493:
```suggestion
           non-value, see comment in [check_representable]. When we allow mixed inline records, we still want to disallow non-value types in unboxed records, so this should become `not unboxed`, as in the `Pcstr_tuple` case. *)
```

ocaml/typing/typedecl.ml
```diff
@@ -435,24 +463,37 @@ let transl_labels ~new_var_jkind env univars closed lbls =
...
+  env loc univars closed = function
   | Pcstr_tuple l ->
-      let flds, flds' = transl_types_gf ~new_var_jkind env univars closed l in
-      Types.Cstr_tuple flds',
-      Cstr_tuple flds
+      let flds, flds' =
+        (* CR layouts: we forbid [@@unboxed] variants from being
+           non-value, see comment in [check_representable]. *)
+        transl_types_gf ~new_var_jkind ~allow_unboxed:(not unboxed)
+        env loc univars closed l (Cstr_tuple { unboxed })
```
goldfirere, line 487: Indent this line a bit, just to make it clear that it's not a new function call.

ocaml/typing/typedecl.ml
```diff
@@ -435,24 +463,37 @@ let transl_labels ~new_var_jkind env univars closed lbls =
...
+      in
+      Types.Cstr_tuple flds', Cstr_tuple flds
   | Pcstr_record l ->
-      let lbls, lbls' = transl_labels ~new_var_jkind env univars closed l in
+      let lbls, lbls' =
+        (* CR layouts: we forbid fields of inlined records from being
+           non-value, see comment in [check_representable]. *)
+        transl_labels ~new_var_jkind ~allow_unboxed:false
+        env univars closed l (Inlined_record { unboxed })
```
goldfirere, line 495: Please indent

goldfirere (approved):

ocaml/typing/typedecl.ml
```diff
@@ -380,7 +380,33 @@ let set_private_row env loc p decl =
   in
   set_type_desc rv (Tconstr (p, decl.type_params, ref Mnil))
 
-let transl_labels ~new_var_jkind env univars closed lbls =
+(* Makes sure a type is representable.  Will lower [any] to a sort variable
```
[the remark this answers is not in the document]
dkalinichenko-js: Done.

ocaml/typing/typedecl.ml
```diff
@@ -380,7 +380,33 @@ let set_private_row env loc p decl =
...
+   if [allow_unboxed = true], and to [value] if [allow_unboxed = false]. *)
+(* CR layouts: Many places where [check_representable] is called in this file
+   should be replaced with checks at the places where values of those types are
+   constructed.  We've been conservative here in the first version. This is the
+   same issue as with arrows. *)
+   let check_representable ~why ~allow_unboxed env loc kloc typ =
```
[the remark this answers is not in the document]
dkalinichenko-js: Fixed.

ocaml/typing/typedecl.ml
dkalinichenko-js: Moved the check to the second pass.

PR #2657, opened 2024-06-05
Fix `configure` for non-amd64 (enable stack checks implicitly)

xclerc:
Set both `STACK_CHECKS_ENABLED` and `enable_stack_checks`
when we implicitly enable the checks (non-amd64 platforms).


mshinwell:
This seems good.  The remaining test failures are probably because of the `Beta` library being absent on arm64 until we do float32 for that platform.

mshinwell (approved):

PR #2659, opened 2024-06-05
Remove `non_null_value` layout

dkalinichenko-js:
Remove `non_null_value` layout since we're re-implementing nullability as its own axis, similar to externality.

I first reverted the original commit adding `non_null_value`, and then fixed merge conflicts and other related code in jkinds.


goldfirere (approved):

goldfirere:
Please merge when CI passes.

PR #2662, opened 2024-06-06
Disable partial inlining for runtime/str.c

mshinwell:
Partial inlining seems to be producing worse code on x86-64 for `caml_string_compare` (and maybe other functions too)?  This PR disables it, just for `str.c` in the initial instance.


mshinwell:
(This might only apply to certain versions of `gcc`.  In due course we'll investigate bumping the optimization level to `-O3` which iirc will remove the need for this PR.)

xclerc (approved):

PR #2666, opened 2024-06-06
Implement `Nullability` axis for jkinds

dkalinichenko-js:
Implement `Nullability` axis for jkinds, ordered `Non_null < Or_null`.

Implement two new legacy jkinds: `value_or_null` and `any_non_null`. The first will be used for `'a or_null`, while the second will be the jkind of array type parameters.

Mark sort variables as always `Non_null` for now. This is incorrect, but setting it to `Or_null` requires a lot of changes to the typing algorithm.

Split `Jkind.has_layout_any` from `Jkind.is_any` and handle existing callsites.

Test the new jkinds.

### DIFF
--- a/ocaml/testsuite/tests/typing-layouts-or-null/array.ml
+++ b/ocaml/testsuite/tests/typing-layouts-or-null/array.ml
@@ -1,0 +1,24 @@
+(* TEST
+ flags = "-extension layouts_alpha";
+ expect;
+*)
+
+(* CR layouts v3.0: array type arguments should be [any_non_null]: *)
+
+type t_any : any
+
+type should_fail = t_any array
+
+[%%expect{|
+type t_any : any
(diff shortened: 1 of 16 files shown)


ccasin:
(Sorry it has taken me a few days to get to this - I plan to review today)

ccasin:
I started reading this, and what I've read so far looks reasonable, but I wonder if we should hold off on this PR until @liam923's rework of `jkind.ml` is done.  These two PRs are pretty conflicty, and that one is a lot bigger.  Also he's adding a notion of jkind abbreviation (not yet available for users, but at least in jkind.ml) which will handle the new kinds added here (`value_non_null` and `any_non_null`) well.

dkalinichenko-js:
Not ready for review pending design discussion.

PR #2667, opened 2024-06-06
Add documentation for mixed blocks

ncik-roberts:
Add some documentation for putting unboxed types in structures.

Review here: https://github.com/ocaml-flambda/flambda-backend/blob/664056e91d0526378ecc34e8890702027044a2b6/ocaml/jane/doc/extensions/unboxed-types/index.md

### DIFF
--- a/ocaml/jane/doc/extensions/unboxed-types/index.md
+++ b/ocaml/jane/doc/extensions/unboxed-types/index.md
@@ -352,7 +352,7 @@ to which the flat float record optimization currently applies.
 type t_flat_float =
   { x1 : float;
     x2 : float#;
-    x3 : float#;
+    x3 : float;
   }
 ```
 
--- a/ocaml/jane/doc/extensions/unboxed-types/index.md
+++ b/ocaml/jane/doc/extensions/unboxed-types/index.md
@@ -323,7 +323,7 @@ arguments.
 
 Suppose a record contains any unboxed field `fld` whose layout is not `value` or a
 combination of `value` layouts. Then, the following restriction applies: All
-fields occuring after `fld` in the record must be "flat", i.e. the GC can
+fields occurring after `fld` in the record must be "flat", i.e. the GC can
 skip looking at them. The only options for flat fields are immediates (i.e. things
 represented as ints at runtime) and other unboxed numbers.
 
--- a/ocaml/jane/doc/extensions/unboxed-types/index.md
+++ b/ocaml/jane/doc/extensions/unboxed-types/index.md
@@ -285,3 +285,113 @@ Here's the list of primitives that currently support `[@layout_poly]`:
 * `%array_safe_set`
 * `%array_unsafe_get`
 * `%array_unsafe_set`
+
+# Using unboxed types in structures
+
+Unboxed types can usually be put in structures, though there are some restrictions.
+
+These structures support unboxed type fields, but have some restrictions on field
+orders:
+  * Records
+  * Constructors
+  
+Unboxed numbers can't be put in these structures:
+  * Constructors with inline record fields
+  * Exceptions
+  * Extensible variant constructors
+  * Top-level fields of modules
+  * Tuples
+  
+There aren't fundamental issues with the structures that lack support. They will
+just take some work to implement.
+
+## Restrictions on ordering of fields
+
+The below is written about record fields but equally applies to constructor
+arguments.
+
+Suppose a record contains any unboxed field `fld` whose layout is not `value` or a
+combination of `value` layouts. Then, the following restriction applies: All
+fields occuring after `fld` in the record must be "flat", i.e. the GC can
+skip looking at them. The only options for flat fields are immediates (i.e. things
+represented as ints at runtime) and other unboxed numbers.
+
+The only relaxation of the above restriction is records that consists
+solely of `float` and `float#` fields. Any ordering of `float` and `float#`
+fields is permitted. The "flat float record optimization" applies to any
+such record&mdash;all of the fields are stored flat, even the `float` ones
+that will require boxing upon projection. The ordering restriction is relaxed
+in this case to provide a better migration story for all-`float` records
+to which the flat float record optimization currently applies.
+
+The reason for this restriction has to do with the "mixed block" runtime
+representation. Read on for more detail about that.
+
+## Generic operations aren't supported
+
+Some operations built in to the OCaml runtime aren't supported for structures
+containing unboxed types.
+
+These operations aren't supported:
+  * polymorphic comparison and equality
+  * polymorphic hash
+  * marshaling
+  
+These operations raise an exception at runtime, similar to how polymorphic
+comparison raises when called on a function.
+
+You should use ppx-derived versions of these operations instead.
+
+## Runtime representation: mixed blocks
+
+The general principle guiding the restriction on field ordering in the user
+program: The compiler should not change the user's order of fields
+when deciding the runtime representation.
+
+Abiding by this principle this allows you to predictably write C bindings and
+predict hardware cache performance.
+
+A structure containing mixed blocks is represented at runtime as a "mixed
+block". A mixed block always consists of fields the GC can scan followed by
+fields the GC either could or must skip. The garbage collector must be kept
+informed of which fields of the block it should scan. A portion of the header
+word is reserved to track the length of the prefix of the block that should be
+scanned by the garbage collector.
+
+The ordering constraint on structure fields is a reflection of the same
+ordering restriction in the runtime representation. 
+
+## C bindings for mixed blocks
+
+The implementation of field layout in a mixed block is not finalized. For example, we'd like for int32 fields to be packed efficiently (two to a word) on 64 bit platforms. Currently that's not the case: each one takes up a word.
+
+Users who write C bindings might want to be notified when we change this layout. To ensure that your code will need to be updated when the layout changes, use the `Assert_mixed_block_layout_v#` family of macros. For example,
+
+```
+Assert_mixed_block_layout_v1;
+```
+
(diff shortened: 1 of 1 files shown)


ocaml/jane/doc/extensions/unboxed-types/index.md
```diff
@@ -284,4 +284,144 @@ Here's the list of primitives that currently support `[@layout_poly]`:
...
 * `%array_unsafe_get`
-* `%array_unsafe_set`
\ No newline at end of file
+* `%array_unsafe_set`
+
+# Using unboxed types in structures
+
+Unboxed types can usually be put in structures, though there are some restrictions.
+
+These structures support unboxed type fields, but have some restrictions on field
```
TheNumbat, line 293: I think it's a bit misleading to say these support unboxed type _fields_, since inline records don't work yet.
Maybe just 'these structures may contain unboxed types'?

ocaml/jane/doc/extensions/unboxed-types/index.md
```diff
@@ -284,4 +284,144 @@ Here's the list of primitives that currently support `[@layout_poly]`:
...
+  * Constructors with inline record fields
+  * Exceptions
+  * Extensible variant constructors
+  * Top-level fields of modules
+  * Tuples
+  
+There aren't fundamental issues with the structures that lack support. They will
+just take some work to implement.
+
+Here's an example of a record with some unboxed fields. We call such a record
```
TheNumbat, line 308: one unboxed field, or add another (or clarify that the int can be in the flat suffix)

ocaml/jane/doc/extensions/unboxed-types/index.md
```diff
@@ -284,4 +284,144 @@ Here's the list of primitives that currently support `[@layout_poly]`:
...
+
+```ocaml
+type t =
+  { str : string;
+    i : int;
+    f : float#;
+  }
+```
+
+## Restrictions on ordering of fields
```
TheNumbat, line 319: restrictions on field ordering?

ocaml/jane/doc/extensions/unboxed-types/index.md
```diff
@@ -284,4 +284,144 @@ Here's the list of primitives that currently support `[@layout_poly]`:
...
+
+The below is written about record fields but equally applies to constructor
+arguments.
+
+Suppose a record contains any unboxed field `fld` whose layout is not `value` or a
+combination of `value` layouts. Then, the following restriction applies: All
```
TheNumbat, line 325: Do combinations of `value` exist? Everything that's not unboxed should just be `value` right?

ocaml/jane/doc/extensions/unboxed-types/index.md
```diff
@@ -284,4 +284,144 @@ Here's the list of primitives that currently support `[@layout_poly]`:
...
+```ocaml
+type t_rejected =
+  { f : float#;
+    s : string;
+  }
+  (* Error: Expected all flat fields after non-value field, f,
+            but found boxed field, s. *)
+```
+
+The only relaxation of the above restriction is records that consists
```
TheNumbat, line 342: ...is for records that consist...

ocaml/jane/doc/extensions/unboxed-types/index.md
```diff
@@ -284,4 +284,144 @@ Here's the list of primitives that currently support `[@layout_poly]`:
...
+fields is permitted. The "flat float record optimization" applies to any
+such record&mdash;all of the fields are stored flat, even the `float` ones
+that will require boxing upon projection. The ordering restriction is relaxed
+in this case to provide a better migration story for all-`float` records
+to which the flat float record optimization currently applies.
+
+```ocaml
+type t_flat_float =
+  { x1 : float;
+    x2 : float#;
```
TheNumbat, line 353: May as well make this example have a `float` after the `float#`

ocaml/jane/doc/extensions/unboxed-types/index.md
```diff
@@ -284,4 +284,144 @@ Here's the list of primitives that currently support `[@layout_poly]`:
...
+  * marshaling
+  
+These operations raise an exception at runtime, similar to how polymorphic
+comparison raises when called on a function.
+
+You should use ppx-derived versions of these operations instead.
+
+## Runtime representation: mixed blocks
+
+The general principle guiding the restriction on field ordering in the user
```
TheNumbat, line 377: As a general principle, the compiler should not change the user-specified field ordering.

ocaml/jane/doc/extensions/unboxed-types/index.md
```diff
@@ -284,4 +284,144 @@ Here's the list of primitives that currently support `[@layout_poly]`:
...
+program: The compiler should not change the user's order of fields
+when deciding the runtime representation.
+
+Abiding by this principle this allows you to predictably write C bindings and
+predict hardware cache performance.
+
+A structure containing mixed blocks is represented at runtime as a "mixed
```
TheNumbat, line 384: A structure containing unboxed types, you mean?

ocaml/jane/doc/extensions/unboxed-types/index.md
```diff
@@ -284,4 +284,144 @@ Here's the list of primitives that currently support `[@layout_poly]`:
...
+block". A mixed block always consists of fields the GC can scan followed by
```
TheNumbat, line 385: Fields the gc _must_ scan? Though immediates can be in the prefix, so technically fields the gc can-and-maybe-must scan...

ocaml/jane/doc/extensions/unboxed-types/index.md
```diff
@@ -284,4 +284,144 @@ Here's the list of primitives that currently support `[@layout_poly]`:
...
+
+You should use ppx-derived versions of these operations instead.
+
+## Runtime representation: mixed blocks
+
+The general principle guiding the restriction on field ordering in the user
+program: The compiler should not change the user's order of fields
+when deciding the runtime representation.
+
+Abiding by this principle this allows you to predictably write C bindings and
```
TheNumbat, line 381: Extra this, also maybe don't say predictable / predict twice

ocaml/jane/doc/extensions/unboxed-types/index.md
```diff
@@ -284,4 +284,144 @@ Here's the list of primitives that currently support `[@layout_poly]`:
...
+
+The implementation of field layout in a mixed block is not finalized. For example, we'd like for int32 fields to be packed efficiently (two to a word) on 64 bit platforms. Currently that's not the case: each one takes up a word.
+
+Users who write C bindings might want to be notified when we change this layout. To ensure that your code will need to be updated when the layout changes, use the `Assert_mixed_block_layout_v#` family of macros. For example,
+
+```
+Assert_mixed_block_layout_v1;
+```
+
+Write the above in statement context&mdash;it should be OK to do so at the top-level of a file or within a function.
```
TheNumbat, line 404: ...in a statement context&mdash;either at the top-level...

ocaml/jane/doc/extensions/unboxed-types/index.md
```diff
@@ -284,4 +284,144 @@ Here's the list of primitives that currently support `[@layout_poly]`:
...
+    y : int32#;
+  }
+```
+
+Here is the recommend way to access fields:
+
+```c
+Assert_mixed_block_layout_v1;
+#define Foo_t_x(foo) (((int32_t*)foo)[0])
+#define Foo_t_y(foo) (((int32_t*)foo)[2])
```
TheNumbat, line 421: Casting it to an int32_t array is a bit weird, would it not be better to do something like `(int32_t*)&Field(foo,0)` and `(int32_t*)&Field(foo,1)`?

TheNumbat (changes requested):

ocaml/jane/doc/extensions/unboxed-types/index.md
```diff
@@ -284,4 +284,144 @@ Here's the list of primitives that currently support `[@layout_poly]`:
...
+```
+
+## Restrictions on ordering of fields
+
+The below is written about record fields but equally applies to constructor
+arguments.
+
+Suppose a record contains any unboxed field `fld` whose layout is not `value` or a
+combination of `value` layouts. Then, the following restriction applies: All
+fields occuring after `fld` in the record must be "flat", i.e. the GC can
```
xclerc, line 326:
```suggestion
fields occurring after `fld` in the record must be "flat", i.e. the GC can
```

ocaml/jane/doc/extensions/unboxed-types/index.md
ncik-roberts: I don't want to need to reword this when we have unboxed tuples (imminent) or unboxed sums (someday).

ncik-roberts: I'll push the technicality into a footnote. I brainstormed better phrasings with Chris and came up empty.

ocaml/jane/doc/extensions/unboxed-types/index.md
ncik-roberts: Indeed.

ocaml/jane/doc/extensions/unboxed-types/index.md
ncik-roberts: can-or-must? I tried that wording. It's an odd modal, even in the scheme of modals, so I added a footnote.

ncik-roberts:
@TheNumbat Thanks for the suggestions. Here's the new preview: https://github.com/ocaml-flambda/flambda-backend/blob/2902a4d82d23b3ee755cf74d902b0bd0abc3ca5a/ocaml/jane/doc/extensions/unboxed-types/index.md

TheNumbat (approved):

PR #2669, opened 2024-06-07
Tail-position `local_` no longer means local-returning

riaqn:
This PR disables the behaviour that tail-position `local_` in function body means the function is local-returning.

This contains a bug fix to `tmc.ml` - @mshinwell can you look at it?

### DIFF
--- a/ocaml/jane/doc/extensions/local/intro.md
+++ b/ocaml/jane/doc/extensions/local/intro.md
@@ -130,7 +130,7 @@ mark the local parameter in the other module's mli.
 There are a number of other features that allow more precise control
 over which values are locally allocated, including:
 
-  - **Local closures**:
+  - **Local closures**
 
     ```ocaml
     let local_ f a b c = ...
@@ -141,7 +141,7 @@ over which values are locally allocated, including:
   - **Local-returning functions**
 
     ```ocaml
-    let f a b c = local_
+    let f a = exclave_
       ...
     ```
 
--- a/ocaml/jane/doc/extensions/local/reference.md
+++ b/ocaml/jane/doc/extensions/local/reference.md
@@ -460,11 +460,8 @@ value `x` remains available.
 
 The region around the body of a function prevents local allocations inside that
 function from escaping. Occasionally, it is useful to write a function that
-allows local allocations to escape, which can be done by explicitly marking
-such functions.
-
-This is useful particularly for constructor functions of abstract types. For
-instance, consider this code that uses an `int ref` as a counter:
+allocates and returns value in caller's region. For instance, consider this code
+that uses an `int ref` as a counter:
 
 ```ocaml
 let f () =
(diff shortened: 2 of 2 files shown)


ocaml/lambda/tmc.ml
```diff
@@ -564,10 +564,12 @@ let llets lk vk bindings body =
 let find_candidate = function
   | Lfunction lfun when lfun.attr.tmc_candidate ->
      (* TMC does not make sense for local-returning functions *)
-     if not lfun.region then
+     begin match lfun.ret_mode with
+     | Alloc_local ->
        raise (Error (Debuginfo.Scoped_location.to_location lfun.loc,
-                     Tmc_without_region));
-     Some lfun
+                     Tmc_without_region))
```
mshinwell, line 570: Maybe this exception should be renamed as well?

riaqn: Good point - will do.

ocaml/typing/typecore.ml
```diff
@@ -8883,7 +8753,7 @@ and type_n_ary_function
     re
       { exp_desc =
           Texp_function
-            { params; body; region = region_locked; ret_sort;
+            { params; body; region = true; ret_sort;
```
ncik-roberts, line 8756: Now, `region` is always `true` for user-written functions and always `false` for functions that arise from coercing labeled functions. This seems suspicious to me -- what does coercing labeled functions have to do with regions? The reason I ask is: if we no longer need the `region` field, we can simplify translation somewhat.

ocaml/typing/typecore.ml
```diff
@@ -239,8 +239,6 @@ exception Error_forward of Location.error
 type in_function =
   { ty_fun: type_expected;
     loc_fun: Location.t;
-    (** [region_locked] is whether the function has its own region. *)
-    region_locked: bool;
   }
```
ncik-roberts, line 242: This type is just a tuple `type_expected * Location.t` upstream. Maybe we can change back to that now, to reduce drift?

ncik-roberts (commented):
Generally this looks good, but I question why we still need the `region` field of `Texp_function`. (See comment.)

ocaml/typing/typecore.ml
riaqn: Ahh nice. Will do

ocaml/typing/typecore.ml
riaqn: Yes you are right. I didn't do the simplification because it would involve lots of changes in `translcore` as well; but I guess it has to be done sometime anyway.

ncik-roberts: Let’s do it in a separate PR.

riaqn: Maybe we can remvoe `region` from `Typedtree.Texp_function` in this PR, and do the changes in `translcore` in a separate PR? I agree with your that `Texp_function` still having `region` is annoying.

riaqn: So I did what I said (remove `region` from `Typedtree.Texp_function`, while try to not change `translcore.ml`).

ncik-roberts (approved):
It would be good to push through the removal of the unnecessary `region` argument in `translcore.ml` soon.

riaqn:
Created JIRA ticket 3132 for simplifying `translcore.ml`

ncik-roberts:
The code changes were dropped from this PR before it was merged, so I've edited the title accordingly. Zesen will open another PR with the code changes.

PR #2671, opened 2024-06-07
Fix a "bug" in `Ctype.constrain_type_jkind`

ccasin:
This fixes a bug that I believe is currently unreachable but will soon not be.  Here's the bug:

```ocaml
let constrain_type_jkind ~fixed env ty jkind =
  (* An optimization to avoid doing any work if we're checking against
     any. *)
  if Jkind.is_any jkind then Ok ()
  else constrain_type_jkind ~fixed env ty jkind
```

The issue is that `is_any` only checks whether the layout is any, and the layout isn't the only part of a jkind, so knowing the layout is any is insufficient to skip the call to `constrain_type_jkind`.

I don't think this bug is reachable because we haven't given the user any way to ask for weird jkinds like `any mod ...`, and because the compiler also never constructs such a jkind.

I've fixed this by:
* Splitting `Jkind.is_any` into two functions: `Jkind.is_max` and `Jkind.has_layout_any`.
* Updating all call sites to use the appropriate one.
* Adjusting some documentation and function names accordingly.

### DIFF
--- a/ocaml/typing/ctype.ml
+++ b/ocaml/typing/ctype.ml
@@ -1658,6 +1658,9 @@ let instance_prim_layout (desc : Primitive.description) ty =
   else
   let new_sort_and_jkind = ref None in
   let get_jkind () =
+    (* CR layouts v2.8: This should replace only the layout component of the
+       jkind. It's possible that we might want a primitive that accepts a
+       mode-crossing, layout-polymorphic parameter. *)
     match !new_sort_and_jkind with
     | Some (_, jkind) ->
       jkind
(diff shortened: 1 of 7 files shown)


ocaml/typing/ctype.ml
```diff
@@ -1675,10 +1675,10 @@ let instance_prim_layout (desc : Primitive.description) ty =
          from an outer scope *)
       if level = generic_level && try_mark_node ty then begin
         begin match get_desc ty with
-        | Tvar ({ jkind; _ } as r) when Jkind.is_any jkind ->
+        | Tvar ({ jkind; _ } as r) when Jkind.has_layout_any jkind ->
```
goldfirere, line 1678: Could you leave this CR around the definition of `get_jkind`?

```
(* CR layouts v2.8: This should replace only the layout component of the jkind. It's possible that we might want a primitive that accepts a mode-crossing, layout-polymorphic parameter. *)
```

goldfirere (approved):
LGTM

ocaml/typing/ctype.ml
ccasin: Good idea, thanks, done.

PR #2672, opened 2024-06-10
Document arrays of unboxed numbers

ncik-roberts:
Document arrays of unboxed numbers, both their syntax and their runtime representation.

### DIFF
--- a/ocaml/jane/doc/extensions/unboxed-types/index.md
+++ b/ocaml/jane/doc/extensions/unboxed-types/index.md
@@ -217,7 +217,7 @@ Here by defining module type `S` with layout `any` and using `with` constraints,
 reason about modules with similar shapes but that operate on different layouts. This removes code
 duplication and can aid ppxs in supporting unboxed types.
 
-# `[@layout_poly]` attribute
+# `[@layout_poly]` attribute {#layout-poly}
 
 The attribute enables support for **limited layout polymorphism on external
 `%`-primitives**. This is possible because these primitives are always inlined at every
@@ -270,7 +270,7 @@ let f2 : float# -> float# = magic;; (* ok *)
 let f3 : float# -> int32# = magic;; (* error *)
 ```
 
-This feature is conceptually similar to `[@local_opt]` for modes and would be useful for
+This feature is conceptually similar to `[@local_opt]` for modes and is useful for
 array access primitives.
 
 Here's the list of primitives that currently support `[@layout_poly]`:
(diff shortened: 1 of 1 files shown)


ncik-roberts:
Preview: https://github.com/ocaml-flambda/flambda-backend/blob/27de938c4f17955cdd4c59693cedf477c99a44d2/ocaml/jane/doc/extensions/unboxed-types/index.md

ocaml/jane/doc/extensions/unboxed-types/index.md
```diff
@@ -285,6 +285,51 @@ Here's the list of primitives that currently support `[@layout_poly]`:
...
+
+| Array                            | Tag                | Layout of data                                              |
+|----------------------------------|--------------------|-------------------------------------------------------------|
+| `float# array`                   | `Double_array_tag` | 64 bits per element                                         |
+| `int64# array`                   | `Custom_tag`       | reserved custom block word, followed by 64 bits per element |
+| `float32# array`, `int32# array` | `Custom_tag`       | reserved custom block word, followed by 32 bits per element |
+
+The reserved custom block word is the standard custom block field that stores a pointer to the
+record of custom operations, like polymorphic equality and comparison. (For unboxed 32-bit element types,
+it also tracks whether the array stores an odd or even number of elements.)
```
mshinwell, line 329: Maybe explicitly say that the custom block pointers are different for odd and even length unboxed int32 arrays.

We could also maybe say something about the padding at the end of an odd-length int32 array, in the most significant 32 bits of the final field.  I think we should probably say that (a) this is unspecified and (b) it is not guaranteed that the value will be preserved by the generated code or by the runtime.

ocaml/jane/doc/extensions/unboxed-types/index.md
```diff
@@ -285,6 +285,51 @@ Here's the list of primitives that currently support `[@layout_poly]`:
...
+Array elements are packed according to their width. For example, arrays of
+elements whose layout is `bits32` store two elements per word.
+
+You can use normal array syntax for constructing such an array:
+
+```ocaml
+let array = [| #2l |]
+```
+
+Array operations must be declared with `[@layout_poly]` to be usable with arrays of unboxed elements.
```
mshinwell, line 307: "operations" -> "primitives"?  Might also be worth explicitly noting that the names are the same for these primitives; this might be unexpected.

ocaml/jane/doc/extensions/unboxed-types/index.md
TheNumbat: I agree + the same applies to `float32# array`

ncik-roberts: Reworded along those lines.

ocaml/jane/doc/extensions/unboxed-types/index.md
ncik-roberts: Changed. I linked back to the `[@layout_opt]` section earlier in the doc rather than re-explain it.

TheNumbat (approved):

PR #2674, opened 2024-06-11
Fix problems with GC compaction and non-consecutive function slots

Ekdohibs:
If function slots were not consecutive, the GC would previously incorrectly scan them during compaction while trying to revert infix pointers.
Note: due to an invariant in slot_offsets, this can't happen with the existing compiler.

### DIFF
--- a/middle_end/flambda2/to_cmm/to_cmm_set_of_closures.ml
+++ b/middle_end/flambda2/to_cmm/to_cmm_set_of_closures.ml
@@ -280,8 +280,10 @@ end = struct
         else if starting_offset = slot_offset
         then acc
         else
-          (* The space between slot offsets has to be padded with tagged 0, as
-             it is scanned by the GC during compaction. *)
+          (* The space between slot offsets has to be padded with precisely the
+             value tagged 0, as it is scanned by the GC during compaction. This
+             value can't be confused with either infix headers or inverted
+             pointers, as noted in the comment in compact.c *)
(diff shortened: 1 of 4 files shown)


middle_end/flambda2/to_cmm/to_cmm_set_of_closures.ml
```diff
@@ -280,6 +280,8 @@ end = struct
         else if starting_offset = slot_offset
         then acc
         else
+          (* The space between slot offsets has to be padded with tagged 0, as
+             it is scanned by the GC during compaction. *)
```
Gbury, line 284: I think we should specify here that the constraint is not only that the value put here has to be scannable, but that it has to be a very specific value, and add a link/comment to the relevant part of the compaction C code.

Gbury (approved):

mshinwell:
(The precondition for this bug cannot be satisfied with the current compiler.)

mshinwell:
Does this need a fix for runtime5?

Ekdohibs:
I think this doesn't need a fix for runtime5, but I just noticed `extern.c` might need a fix as well; I'll look.

Gbury (approved):
Looks good, we should also probably make a note somewhere that compaction in runtime5 will need to be adjusted when it is merged ?

lthls (approved):
Approving the latest changes

Ekdohibs:
Compaction in runtime5 has already been merged, and works differently to runtime4 so doesn't need to be changed.

PR #2676, opened 2024-06-11
Support for `mod` syntax

liam923:
Allow users to write `mod` syntax (ex: `type t : value mod global`). This PR also refactors Jkind so that `immediate`/`immediate64` is no longer a layout, but instead shorthand for `value mod ...`.

### DIFF
--- a/ocaml/testsuite/tests/tool-ocamlobjinfo/question.ml
+++ b/ocaml/testsuite/tests/tool-ocamlobjinfo/question.ml
@@ -15,6 +15,7 @@
    program = "-no-code question.cmx";
    ocamlobjinfo;
 
+   runtime5;
    check-program-output;
  }{
    program = "question.cmx";
--- a/ocaml/testsuite/tests/tool-ocamlobjinfo/question.reference
+++ b/ocaml/testsuite/tests/tool-ocamlobjinfo/question.reference
@@ -38,27 +38,27 @@ Original unit: Question
 Typing env:
 ((defined_symbols_without_equations ())
  (code_age_relation
-  {([0m[38;5;169;1mcamlQuestion__answer_0_1_code[0m [0m[38;5;169;1mcamlQuestion__answer_0_0_code[0m)})
+  {([0m[38;5;169;1mcamlQuestion.answer_0_1_code[0m [0m[38;5;169;1mcamlQuestion.answer_0_0_code[0m)})
  (type_equations
   {([0m[38;5;111;1m[0m[38;5;98;1mQuestion.camlQuestion[0m[38;5;111;1m[0m
     (Val
      (Variant
       (blocks
        ((alloc_mode Heap) (known
-         {(tag_0 => (Known 1), ((Val ([0m[38;5;160;1m=[0m [0m[38;5;111;1m[0m[38;5;98;1mQuestion.camlQuestion__answer_1[0m[38;5;111;1m[0m))))})
+         {(tag_0 => (Known 1), ((Val ([0m[38;5;160;1m=[0m [0m[38;5;111;1m[0m[38;5;98;1mQuestion.camlQuestion.answer_1[0m[38;5;111;1m[0m))))})
         (other Bottom))) (tagged_imms (Naked_immediate [0m[38;5;37;1m⊥[0m)))))
-   ([0m[38;5;111;1m[0m[38;5;98;1mQuestion.camlQuestion__answer_1[0m[38;5;111;1m[0m
+   ([0m[38;5;111;1m[0m[38;5;98;1mQuestion.camlQuestion.answer_1[0m[38;5;111;1m[0m
     (Val
      ((alloc_mode Heap) (known
        {([0m[38;5;31;1m(answer/0 ∷ [0m[38;5;37;1m𝕍[0m[38;5;31;1m)[0m
          => (Known ((closures { [0m[38;5;31;1m(answer/0 ∷ [0m[38;5;37;1m𝕍[0m[38;5;31;1m)[0m }) (value_slots { }))),
          ((function_types
            {([0m[38;5;31;1m(answer/0 ∷ [0m[38;5;37;1m𝕍[0m[38;5;31;1m)[0m
-             (Ok (function_type (code_id [0m[38;5;169;1mcamlQuestion__answer_0_1_code[0m)
+             (Ok (function_type (code_id [0m[38;5;169;1mcamlQuestion.answer_0_1_code[0m)
                   (rec_info (Rec_info [0m[38;5;249;1m0[0m)))))})
           (closure_types
            ((function_slot_components_by_index
-             {([0m[38;5;31;1m(answer/0 ∷ [0m[38;5;37;1m𝕍[0m[38;5;31;1m)[0m (Val ([0m[38;5;160;1m=[0m [0m[38;5;111;1m[0m[38;5;98;1mQuestion.camlQuestion__answer_1[0m[38;5;111;1m[0m)))})))
+             {([0m[38;5;31;1m(answer/0 ∷ [0m[38;5;37;1m𝕍[0m[38;5;31;1m)[0m (Val ([0m[38;5;160;1m=[0m [0m[38;5;111;1m[0m[38;5;98;1mQuestion.camlQuestion.answer_1[0m[38;5;111;1m[0m)))})))
           (value_slot_types ((value_slot_components_by_index {})))))})
       (other Bottom))))})
  (aliases


Ekdohibs (approved):
For `chamelon` changes

ocaml/testsuite/tests/typing-layouts/annots.ml
```diff
@@ -79,6 +134,19 @@ Error: This alias is bound to type int list
...
+let x : int list as ('a : value mod global) = [3;4;5]
+[%%expect {|
+Line 1, characters 8-43:
+1 | let x : int list as ('a : value mod global) = [3;4;5]
+            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Error: This alias is bound to type int list
+       but is used as an instance of type ('a : value mod global)
+       The layout of int list is value, because
+         it's a boxed variant type.
+       But the layout of int list must be a sublayout of value, because
```
goldfirere, line 146: This clearly is a poor error message. I know doing better will be hard. And maybe best to do in a followup PR. In the meantime, I wonder how it would work to print a layout when all the modes are top, or print `immediate`/`immediate64` for those cases. Otherwise, print the layout `mod` all the non-top modes.

EDIT: Maybe that's already happening, but this error message just reports layout? Read my other comments before taking action here, as I suggest possibly moving error messages to talk about kinds, not layouts.

ocaml/testsuite/tests/typing-layouts/annots.ml
```diff
@@ -103,20 +171,28 @@ type t = string t2
...
 type t1 = int t2_imm
-type t2' = bool t2_imm
+type t2' = bool t2_imm'
 type ('a : float64) t2_float64
 type (_ : float64) t2_float64'
 type t3 = float# t2_float64
+type ('a : value mod global) t2_global
+type (_ : value mod global) t2_global'
+type ('a : word mod external_ many shared) t2_complex
+type (_ : word mod external_ many shared) t2_complex'
```
goldfirere, line 181: Maybe this should be a warning, `mod`ding by a top.

ocaml/testsuite/tests/typing-layouts/annots.ml
```diff
@@ -166,28 +276,73 @@ Error: This type string should be an instance of type ('a : immediate)
...
 
-let f : 'a t2_imm -> 'a t2_imm = fun x -> x
+type t = string t2_global
+;;
+[%%expect {|
+Line 1, characters 9-15:
+1 | type t = string t2_global
+             ^^^^^^
+Error: This type string should be an instance of type ('a : value mod global)
+       The layout of string is value, because
```
goldfirere, line 286: I also think we're going to have to switch these messages from saying "layout" to "kind". Actually, it's probably best to report layout if that's the interesting bit. But if the layouts are OK, we should print the kinds.

ocaml/testsuite/tests/typing-layouts/annots.ml
```diff
@@ -201,46 +356,101 @@ Error: The universal type variable 'a was declared to have layout value.
...
+|}]
+
+let f : ('a : word). 'a t2_complex -> 'a t2_complex = fun x -> x
+;;
+[%%expect {|
+Line 1, characters 8-51:
+1 | let f : ('a : word). 'a t2_complex -> 'a t2_complex = fun x -> x
+            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Error: The universal type variable 'a was declared to have layout word.
+       But it was inferred to have layout word, because
```
goldfirere, line 377: Another message that we'll need to generalize to work over kinds.

ocaml/testsuite/tests/typing-layouts/annots.ml
```diff
@@ -343,6 +599,19 @@ Error: This field value has type 'b -> 'b which is less general than
 |}]
 (* CR layouts v1.5: that's a pretty awful error message *)
 
+let r = { field = fun (type a : value mod global) (x : a) -> x }
+[%%expect{|
+Line 1, characters 18-62:
+1 | let r = { field = fun (type a : value mod global) (x : a) -> x }
+                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Error: This field value has type 'b -> 'b which is less general than
```
goldfirere, line 607: Another terrible error message

goldfirere (commented):
Reviewed ocaml/testsuite/tests/typing-layouts/annots.ml, just to get things started. More to come.

ocaml/testsuite/tests/typing-layouts/annots.ml
liam923: Yes, I definitely agree that these errors are terrible. My thought was to leave the message as-is if the layouts disagree, but in other cases, say that the kinds disagree and give the full kind.

That being said, I intentionally didn't do that as part of this PR to keep it smaller. But it would make sense to me if an improved error message is a prerequisite to merging.

goldfirere: I figured as much, about your intention. The problem is that merging this will, I think, open the possibility of these bad error messages. One way forward is to leave this PR as-is, with bad error messages, and then make a stacked PR, reviewed separately, that's just about the error messages. We'd merge the stacked PR into this one, and then merge into `main`. Does that sound like it would work?

ocaml/testsuite/tests/typing-layouts/jkinds.ml
```diff
@@ -84,6 +87,372 @@ Uncaught exception: Misc.Fatal_error
...
+|}]
+
+type a : immediate
+type b : value mod global unique many uncontended portable external_ = a
+type c : value mod global unique many uncontended portable external_
+type d : immediate = c
+[%%expect{|
+type a : immediate
+type b = a
+type c : value mod global unique many uncontended portable external_
```
goldfirere, line 284: I'm surprised at this output. It's fine -- it's what the user wrote. But I somehow expected to see `immediate` here.

ocaml/testsuite/tests/typing-layouts/annots.ml
liam923: Yep, that sounds reasonable. We may have to lose our review history on this PR though as a consequence. Since the incoming branch is from my fork, I can't open a PR against it within this repo. Unless GitHub has a feature to do this that I'm not aware of, I'll have to close this PR, create a branch within this repo, and open a PR on that. The alternative would be to have a PR within my fork, but we'd lose CI and imo it'd be bad because then PRs wouldn't all be in one place.

I should probably just start making branches on this repo instead of on my fork to avoid these situations in the future...

ocaml/testsuite/tests/typing-layouts/jkinds.ml
```diff
@@ -84,6 +87,372 @@ Uncaught exception: Misc.Fatal_error
...
+type d : immediate64 = c
+[%%expect{|
+type a : immediate64
+type b = a
+type c : value mod global unique many uncontended portable external64
+type d = c
+|}]
+
+type a : float64
+type b : float64 mod global unique many uncontended portable external_ = a
```
goldfirere, line 300: This is a bug. That is, `float64` should really just be the layout, not with any mode crossing. We've been unwilling to fix the bug until we had syntax to get the mode crossing separately, and now we do! But probably best to fix in a future PR. I've made ticket 3157.

ocaml/testsuite/tests/typing-layouts/jkinds.ml
goldfirere, line 1: I was expecting to see some tests involving mode crossing directly. That is, when I say `type t : value mod global`, do values of type `t` really mode-cross locality? Similarly, it would be nice to test that a private abbreviation for `int` can be made not to mode-cross. Also worth testing a `[@@unboxed]` wrapper around `int` that doesn't mode-cross. There should be at least one test for each axis.

-------------------------------

Another test worth doing is that mode-crossing along externality actually causes the optimizations in the back end. That is, when we see a type mode-crosses to `external_`, we know that it's represented by a non-pointer. This means several nice things:
1. We can put the thing in the non-scanned suffix of a mixed block.
2. Updating a mutable field of that type does not cause a call to `caml_modify` (speak up if you have no idea what I'm talking about here)
3. Maybe various other things?

The bottom line is that `value`s that cross to `external_` should be `Pintval` in typeopt. (And similar for `external64`, but only on native code.)

We should also make sure that `caml_modify` is never called when mutating a non-`value`, though that's probably somewhat afield from this PR.

----------------------------------

It might also be good to test intersection behavior. That is, I might have

```
let f : ('a : value mod global many) -> ('a : value mod uncontended) = fun x -> x
```

and I should get the intersection between those two kinds. Similarly in a type declaration, with something like

```
type ('a : value mod global) t = ('a : value mod external_)
```

---------------------------------

What soundness checks are included in this? That is, is there anything that stops us from saying

```
type t : value mod global = { x : string }
```

? I understand that we want to allow unsound mode-crossing for portability and contention, but I don't think we want the same for locality. There should be tests that show the presence (or absence!) of the soundness checks.

-------------------------------

What do we do for inference? Immutable data can cross portability, linearity, and contention (as long as there are no functions inside -- which can be accomplished for now by having concrete types as fields). Do we infer this mode crossing? I'm OK if we don't, for now, but there should be tests witnessing what we actually do.

-----------------------------

I think there should be a few tests for `[@@unboxed]` types and `private` abbreviations: these are both scenarios where kinds of a nominative type are inherited from the kind of some other type, but can be modified by an explicit kind annotation. I think this should all work (it's not new in this patch), but I'd want to see a few tests showing that, say, the new kind can be a superkind of the old one (but never a subkind!).

---------------------------

Inference is even harder in the presence of modalities. For example, the type

```
type 'a global = { global : 'a @@ global } [@@unboxed]
```

can mode-cross to global. I doubt that's implemented or inferred here, but there should be tests witnessing the current state of play. Any test outputs that should eventually change should be labeled with e.g. `(* CR layouts v2.8: ... *)`, even if that means labeling a lot of tests. We don't want any to leak through. (Ditto for poor messages, actually, though admittedly those are less important.)

-----------------------------

A GADT pattern-match should be able to refine kinds. It would be cool to see that this refinement can cause mode-crossing. Let's talk if you don't know how to proceed here; the water is getting deeper here. (I don't actually necessarily expect that this will work out of the box -- but we should test and leave a CR for fixing later.)

--------------------------

No one has thought about how any of this stuff should interact with the object system. I think it's all probably fine, but there should be a test or two to make sure that objects have kind `value` with no mods. Not sure about object fields... but I bet existing checks cover this.

-----------------

Hopefully that's enough!

goldfirere (commented):
Finished reviewing testsuite changes

ocaml/typing/env.ml
```diff
@@ -2241,6 +2241,8 @@ let add_value_lazy ?check ?shape ?(mode=Mode.Value.allow_right Mode.Value.legacy
   store_value ?check mode id addr desc shape env
 
 let add_type ~check ?shape id info env =
+  (* CR layouts: there should be a safety check for extension universe when the type's
```
goldfirere, line 2244: Why? I think doing this check here is awkward. I'd rather do in jkind when we interpret the user-written syntax. I guess I'm depending on the fact that mode-crossing can't suddenly arise without such an annotation.

ocaml/typing/jkind.ml
```diff
@@ -1443,46 +1755,19 @@ let report_error ~loc = function
...
-type desc =
-  | Const of const
-  | Var of Sort.var
-
-let to_legacy_desc : Desc.t -> desc = function
-  | Const c -> Const (Const.to_legacy_jkind c)
-  | Var v -> Var v
-
-let get t = to_legacy_desc (get t)
+type annotation = Const.t * Jane_syntax.Jkind.annotation
```
goldfirere, line 1771: I think it should be a goal of this work to get rid of this. That is, once we have a decent printing story, we shouldn't have to remember the user-written bit.

ocaml/typing/outcometree.mli
```diff
@@ -111,7 +101,31 @@ type out_ret_mode =
...
+  | Ojkind_user_mod of out_jkind_user * string list
+  | Ojkind_user_with of out_jkind_user * out_type
+  | Ojkind_user_kind_of of out_type
+
+and out_jkind_const = { base : string; modal_bounds : string list }
+
+and out_jkind =
+  | Ojkind_user of out_jkind_user
+  | Ojkind_const of out_jkind_const
+  | Ojkind_var of string
```
goldfirere, line 117: This doesn't quite seem right: `Ojkind_user (Ojkind_user_abbreviation ...)` is the same as `Ojkind_var ...`. I think we shouldn't need to care about whether the user wrote the annotation or not. Let's talk if there is trouble simplifying this.

ocaml/typing/outcometree.mli
```diff
@@ -111,7 +101,31 @@ type out_ret_mode =
   (** The ret type is arrow, and need to print parens around the arrow, with
       modes annotating. *)
 
-type out_type =
+(** Represents a user-written jkind annotation *)
+type out_jkind_user =
```
goldfirere, line 105: I think these should be folded in with `out_jkind`.

ocaml/typing/predef.ml
goldfirere, line 1: I did not review this file. Let's come back to this after the `Jkind.Const.Primitive` module is refactored.

ocaml/typing/printtyp.ml
goldfirere, line 1: I did not review this file. Let's come back to this after the types in `Outcometree` are refactored.

ocaml/typing/printtyped.ml
```diff
@@ -222,7 +222,7 @@ let attributes i ppf l =
   ) l
 
 let jkind_annotation i ppf (jkind, _) =
-  line i ppf "%s" (Jkind.string_of_const jkind)
+  line i ppf "%a" Jkind.Const.format jkind
```
goldfirere, line 225: It's possible that `Const.format` is the wrong thing here. We want this output to be pretty raw. If that function evolves to be too cooked, we should do something different here.

ocaml/typing/subst.ml
```diff
@@ -94,29 +94,9 @@ let with_additional_action (config : additional_action_config) s =
...
-          | Const Void -> void
-          | Const Value -> value
-          | Const Immediate -> immediate
-          | Const Immediate64 -> immediate64
-          | Const Float64 -> float64
-          | Const Float32 -> float32
-          | Const Word -> word
-          | Const Bits32 -> bits32
-          | Const Bits64 -> bits64
+          | Const const -> Jkind.of_const const ~why:Jkind.History.Imported
```
goldfirere, line 99: This loses the memoization in the old implementation. As these types get more complicated, we'll lose sharing opportunities. But I think we should still probably memoize the built-in jkinds.

ocaml/typing/typedecl.ml
```diff
@@ -1249,16 +1255,16 @@ module Element_repr = struct
...
+        Value_element
+      | Some Value, External -> Imm_element
+      | Some Float64, _ -> Unboxed_element Float64
+      | Some Float32, _ -> Unboxed_element Float32
+      | Some Word, _ -> Unboxed_element Word
+      | Some Bits32, _ -> Unboxed_element Bits32
+      | Some Bits64, _ -> Unboxed_element Bits64
+      | Some Void, _ -> Element_without_runtime_component { loc; ty }
+      | None, _ ->
           Misc.fatal_error "Element_repr.classify: unexpected Any"
```
goldfirere, line 1268:
```suggestion
          Misc.fatal_error "Element_repr.classify: unexpected abstract layout"
```

Just for future-proofing! (Users won't be able to declare abstract layouts directly. But the layout that's part of an abstract kind will be an abstract layout.)

ocaml/typing/typeopt.ml
```diff
@@ -234,12 +234,16 @@ let bigarray_type_kind_and_layout env typ =
...
 let value_kind_of_value_jkind jkind =
-  match Jkind.get_default_value jkind with
-  | Value -> Pgenval
-  | Immediate -> Pintval
-  | Immediate64 ->
+  let const_jkind = Jkind.default_to_value_and_get jkind in
+  let externality_upper_bound =
+    Jkind.Const.get_externality_upper_bound const_jkind
+  in
+  (* CR: assert the sort is a value *)
```
goldfirere, line 241: Yes, indeed. But why not just do this? I don't think this needs to be delayed.

ocaml/testsuite/tests/language-extensions/pprintast_unconditional.reference
goldfirere, line 1: Do you know what's going on here? I don't immediately see why this output changes.

ocaml/testsuite/tests/typing-layouts/jkinds.ml
liam923: There's a comment in `ocaml/typing/printtyp.ml` that discusses jkind printing. I attempted to keep this comment true with my changes. This case refers to `C1.1` in that comment, which dictates that the user-written jkind is printed for abstract types with no manifest.

liam923: To clarify though, I wouldn't be sad if that logic changes

goldfirere: Makes sense. I wrote that comment some time ago, just feeling my way through it all. At that point, if we printed an annotation, what we printed was unambiguous. Now, the decision of whether to print and what to print are different. That is, if someone writes `type t : value`, should we print the annotation? I was warming up to say "yes"... but actually I think we shouldn't. We should feel free to update the description and the code as time evolves. Maybe even delete the comment if it becomes obvious in context or something.

ocaml/testsuite/tests/language-extensions/pprintast_unconditional.reference
liam923: It's a pprinting test that I added. Line 210 in `ocaml/testsuite/tests/language-extensions/pprintast_unconditional.ml` causes it:
```
let mode = test "mode" mode Example.mode
```

ocaml/typing/jkind_types.mli
```diff
@@ -79,11 +79,29 @@ module Sort : sig
...
-  type ('type_expr, 'sort) layout =
+  type 'sort layout =
     | Sort of 'sort
     | Any
 
-  type 'type_expr t = ('type_expr, Sort.t) layout
+  module Const : sig
+    type t = Sort.const layout
+
+    module Legacy : sig
```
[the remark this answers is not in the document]
liam923: This PR is in a race condition with #2730. Printing for users is moved off legacy in 2730, and the other cases I plan to resolve in this PR. I'll remove the `Legacy` module in whichever PR is merged second.

ocaml/typing/jkind_types.mli
```diff
@@ -79,11 +79,29 @@ module Sort : sig
 end
 
 module Layout : sig
-  type ('type_expr, 'sort) layout =
+  type 'sort layout =
```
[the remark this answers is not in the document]
liam923: Oops yes I believe you're right. I'll leave it for now though - it might be better for avoiding merge conflicts.

ocaml/typing/jkind_types.ml
```diff
@@ -28,17 +28,63 @@ module Sort = struct
...
-    | Value, Value
-    | Float64, Float64
-    | Float32, Float32
-    | Word, Word
-    | Bits32, Bits32
-    | Bits64, Bits64 ->
-      true
-    | (Void | Value | Float64 | Float32 | Word | Bits32 | Bits64), _ -> false
+  module Const = struct
+    type t = const
```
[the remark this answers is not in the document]
liam923: I had tried to do this before, but ran into an OCaml issue I couldn't resolve. When making this change, I ended up with `Sort` in `jkind_types.mli` looking like:
```ocaml
module Sort : sig
  module Const : sig
    type t =
      | Void
      | Value
      | Float64
      | Float32
      | Word
      | Bits32
      | Bits64
  end

  type t =
    | Var of var
    | Const of Const.t
  and var = t option ref

  include Jkind_intf.Sort with type t := t and type var := var and type Const.t := Const.t
end
```
This gives an error on the `include` statement:
```
Multiple definition of the module name Const.
Names must be unique in a given structure or signature.
```
Because of this, I believe that we cannot define `Sort.t` in terms of an identifier named `Sort.Const.t`. Or at least, without modifying the `Jkind_intf` and `Jkind_types` structure that we have. Creating a `const` seemed less bad to me.

ocaml/typing/jkind.mli
```diff
@@ -115,64 +152,126 @@ end
...
+  (** Gets the legacy layout of a constant jkind. Never does mutation. *)
+  val get_legacy_layout : t -> Layout.Const.Legacy.t
+
+  (** Gets the maximum modes for types of this constant jkind. *)
+  val get_modal_upper_bounds : t -> Mode.Alloc.Const.t
+
+  (** Gets the maximum mode on the externality axis for types of this constant jkind. *)
+  val get_externality_upper_bound : t -> Externality.t
+
+  module Primitive : sig
```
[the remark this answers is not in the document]
liam923: Based on our earlier conversation, I renamed the module to `Built_in`. I think we can leave it here for now, but maybe move it in the near future if it begins to make for sense to be in predef.

ocaml/typing/jkind.mli
```diff
@@ -115,64 +152,126 @@ end
...
+    (** This is the jkind of unboxed 64-bit integers. They have sort Bits64. Does
+    not mode-cross. *)
+    val bits64 : t
 
-(** This jkind is the top of the jkind lattice. All types have jkind [any].
+    (** Get a list of all primitive jkinds *)
+    val get_all : t list
+  end
+
+  module Sort : module type of struct
```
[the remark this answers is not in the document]
liam923: `typedecl.ml:1252`:
```ocaml
let sort = Jkind.Const.(Layout.get_sort (get_layout const_jkind)) in
```
vs
```ocaml
let sort = Jkind.(Layout.Const.get_sort (Const.get_layout const_jkind)) in
```

It makes this line nicer, but I think it's probably just a code smell to be duplicating these modules. I'm going to get rid of it.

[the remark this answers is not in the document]
liam923: Also worth noting that this is the only line that actually uses these aliases.

ocaml/typing/jkind_types.mli
```diff
@@ -131,4 +137,12 @@ type 'type_expr t =
     has_warned : bool
   }
 
-type annotation = const * Jane_syntax.Jkind.annotation
+module Const : sig
```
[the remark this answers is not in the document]
liam923: It's necessary because `Jkind.annotation` depends on it, and `types.mli` depends on `Jkind.annotation`.

ocaml/typing/jkind.mli
```diff
@@ -115,64 +152,126 @@ end
...
-  | Word
-  | Bits32
-  | Bits64
+module Const : sig
+  (** Constant jkinds are used for user-written annotations *)
+  type t = Types.type_expr Jkind_types.Const.t
 
-val const_of_user_written_annotation :
-  context:annotation_context -> Jane_syntax.Jkind.annotation -> const
+  val to_out_jkind_const : t -> Outcometree.out_jkind_const
```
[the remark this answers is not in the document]
liam923: That was the reasoning for it. Well, that's a bit of a lie. `Printtyp` would need to see inside `Jkind.Const.t` to be able to print jkinds. This is something that's not supposed to be exposed, but in reality is because of `jkind_types.ml`.

ocaml/typing/jkind.mli
```diff
@@ -115,64 +152,126 @@ end
...
+  module Sort : module type of struct
+    include Sort.Const
+  end
+
+  module Layout : module type of struct
+    include Layout.Const
+  end
+end
+
+module Primitive : sig
```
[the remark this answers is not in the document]
liam923: Renamed to `Builtin`

ocaml/typing/jkind.mli
```diff
@@ -115,64 +152,126 @@ end
...
-(** We know for sure that values of types of this jkind are always immediate *)
-val immediate : why:immediate_creation_reason -> t
+  (** We know for sure that values of types of this jkind are always immediate *)
+  val immediate : why:History.immediate_creation_reason -> t
 
-(** This is the jkind of unboxed 64-bit floats.  They have sort
+  (** This is the jkind of unboxed 64-bit floats.  They have sort
     Float64. Mode-crosses. *)
-val float64 : why:float64_creation_reason -> t
+  val float64 : why:History.float64_creation_reason -> t
```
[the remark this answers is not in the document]
liam923: Done

ocaml/typing/jkind.mli
```diff
@@ -182,22 +281,27 @@ val add_mode_crossing : t -> t
...
+val of_new_sort_var : why:History.concrete_jkind_reason -> t * sort
 
 (** Create a fresh sort variable, packed into a jkind. *)
-val of_new_sort : why:concrete_jkind_reason -> t
+val of_new_sort : why:History.concrete_jkind_reason -> t
+
+val of_const : why:History.creation_reason -> Const.t -> t
 
-val of_const : why:creation_reason -> const -> t
+val const_of_user_written_annotation :
```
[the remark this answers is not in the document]
liam923: Done

ocaml/typing/jkind.ml
```diff
@@ -491,77 +811,67 @@ let fresh_jkind jkind ~why =
...
    At some points in time, we will not care about the context, and so this
    parameter might effectively be unused.
 *)
 (* CR layouts: When everything is stable, remove this function. *)
-let get_required_layouts_level (context : annotation_context)
-    (jkind : Legacy.const) : Language_extension.maturity =
-  match context, jkind with
+let get_required_layouts_level (context : History.annotation_context)
+    (jkind : Const.t) : Language_extension.maturity =
+  let legacy_layout = Const.get_legacy_layout jkind in
```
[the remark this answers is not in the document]
liam923: Fixed

ocaml/typing/jkind.ml
```diff
@@ -109,7 +34,7 @@ type type_expr = Types.type_expr
 module Layout = struct
   open Jkind_types.Layout
 
-  type nonrec 'sort layout = (type_expr, 'sort) layout
+  type nonrec 'sort layout = 'sort layout
```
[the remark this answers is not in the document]
liam923: Removed

ocaml/typing/jkind.ml
```diff
@@ -127,9 +52,67 @@ module Layout = struct
       | _ when equal c1 c2 -> Equal
       | _, Any -> Less
       | Any, Sort _ | Sort _, Sort _ -> Not_le
+
+    let value = Sort Sort.Const.value
+
+    let void = Sort Sort.Const.void
+
+    let float64 = Sort Sort.Const.float64
```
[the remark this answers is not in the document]
liam923: Removed (and the others too)

ocaml/typing/jkind.ml
```diff
@@ -305,6 +301,338 @@ module Const = struct
...
+        immediate;
+        immediate64;
+        float64;
+        float32;
+        word;
+        bits32;
+        bits64 ]
+  end
+
+  module To_out_jkind_const = struct
```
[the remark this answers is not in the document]
liam923: I added a comment. But also as discussed in person, this will be changed soon (to be in terms of kind abbreviations).

ocaml/typing/env.ml
liam923: Yes I think you're right that this CR is wrong. I can't even remember why I wrote it in the first place.

liam923: And this check is already done when interpreting user syntax.

ocaml/typing/outcometree.mli
liam923: Fixed

ocaml/typing/outcometree.mli
liam923: Fixed

ocaml/typing/subst.ml
liam923: This was fixed in #2717

ocaml/typing/typedecl.ml
liam923: done

ocaml/typing/typeopt.ml
liam923: I meant to come back to it and then forgot to. Fixed.

ocaml/testsuite/tests/typing-layouts/annots.ml
liam923: Done in #2730

ocaml/testsuite/tests/typing-layouts/jkinds.ml
liam923: This was addressed (mostly) in #2717, although I'm still not sure how to do the `caml_modify` test.

liam923:
The first commit, 5d4918a (`Promote tests that remove redundant kind from error message history`), makes the same minor change to error messages in tests in many places. I recommend reviewing that one separately from the rest of this PR so the diff is less noisy.

ocaml/testsuite/tests/typing-layouts/annots.ml
liam923: In my mind, a good improvement that seems reasonable to implement would be:
```
type r_value = { field : 'a. 'a -> 'a }
let r = { field = fun (type a : immediate) (x : a) -> x }

[%%expect{|
type r_value = { field : 'a. 'a -> 'a; }
Line 2, characters 18-55:
2 | let r = { field = fun (type a : immediate) (x : a) -> x }
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error: This field value has type ('b: immediate). 'b -> 'b which is less general than
         'a. 'a -> 'a
       The kind of 'a is value
         because of the definition of r_value at line 1, characters 0-39.
       But the kind of 'a must be a subkind of immediate
         because of the annotation on the abstract type declaration for a.
|}]
```
(Note that `'b -> 'b` is now `('b : immediate). 'b -> 'b`).

Actually, I'm mildly surprised this doesn't already happen. This seems to correspond to case `C3` in `printtyp.ml`.

But I think this is outside the scope of this PR - this error message predates my earlier jkind PR.

ocaml/testsuite/tests/typing-layouts/jkinds.ml
liam923: I've updated C1.1 to not print the jkind when it is equal to `value`, and updated the tests to reflect it. I am mildly concerned that a user might get confused if they write, say, `type t : value mod nonportable`, and the error message only refers to `type t`. If they included the `nonportable`, they probably didn't realize that it's a top and therefore a no-op. And if they don't realize that, not including it in the error could be confusing. But I guess if there's warnings for moding by a top, this is a non-issue.

ocaml/typing/jkind_types.mli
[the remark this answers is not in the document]
liam923: (Note that the future printing improvements should get rid of `Jkind.annotation`, which should in turn get rid of this)

ocaml/typing/jkind.ml
```diff
@@ -305,6 +301,338 @@ module Const = struct
...
+  let mode_crossing layout =
+    { layout;
+      modes_upper_bounds = Modes.min;
+      externality_upper_bound = Externality.min
+    }
+
+  module Primitive = struct
+    type nonrec t =
+      { jkind : t;
+        name : string
```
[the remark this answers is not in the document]
liam923: We decided in-person that this will be addressed in a future PR (either the abbreviation or printing one)

ocaml/typing/jkind.ml
```diff
@@ -305,6 +301,338 @@ module Const = struct
...
+        (* convert_with_base is guaranteed to succeed since the layout matches and the
+           modal bounds are all max *)
+        Option.get out_jkind_verbose
+  end
+
+  let to_out_jkind_const = To_out_jkind_const.convert
+
+  let format ppf jkind =
+    let legacy_layout = get_legacy_layout jkind in
+    let layout_str = Layout.Const.Legacy.to_string legacy_layout in
```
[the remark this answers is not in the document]
liam923: This was resolved in PR #2730

ocaml/typing/jkind.ml
liam923: This should be removed when printing is improved in the near future. I'll add a CR.

ocaml/testsuite/tests/typing-layouts/jkinds.ml
goldfirere: We discussed a bit in person and I believe that grepping for `caml_modify` in intermediate output would make a decent test. @ccasin recommended doing this in the output produced by `-dsel`.

ocaml/typing/printtyped.ml
liam923: I've updated it to be more explicit

ocaml/testsuite/tests/typing-layouts/annots.ml
```diff
@@ -32,17 +32,72 @@ Error: Layout void is more experimental than allowed by the enabled layouts exte
...
+|}]
+
+type t1 : float32 mod internal shared many local
+type t2 : bits64 mod once external64 unique
+type t3 : immediate mod local unique
+type t4 : value mod local local
+type t5 : float64 mod global global
+type t6 : bits32 mod local global
+type t7 : bits64 mod global local
+type t8 : value mod global local many once unique shared internal uncontended contended
```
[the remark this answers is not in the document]
liam923: Done

ocaml/testsuite/tests/typing-layouts/annots.ml
liam923: Based on our conversation, this is going to be left to do in another PR.

ocaml/utils/warnings.ml
```diff
@@ -581,6 +583,10 @@ let descriptions = [
     names = ["unboxing-impossible"];
     description = "The parameter or return value corresponding @unboxed attribute cannot be unboxed.";
     since = since 4 14 };
+  { number = 211;
+    names = ["mod-by-top"];
+    description = "Modifying by the top element of an axis is a no-op.";
```
goldfirere, line 588: I think "kind" should be mentioned here somewhere

ocaml/utils/warnings.ml
```diff
@@ -1217,6 +1223,11 @@ let message = function
       Printf.sprintf
         "This [@unboxed] attribute cannot be used.\n\
          The type of this value does not allow unboxing."
+  | Mod_by_top modifier ->
+      Printf.sprintf
+        "The modifier %s is a top.\n\
```
goldfirere, line 1228: "is the topmost modifier"? I don't really like "is a top".

ocaml/utils/warnings.ml
```diff
@@ -1217,6 +1223,11 @@ let message = function
...
+         Modifying by a top element is a no-op."
```
goldfirere, line 1229: I don't see this string in the testsuite. Add some tests with this warning enabled.

goldfirere (approved):

ocaml/testsuite/tests/typing-layouts-caml-modify/basics.ml
```diff
@@ -0,0 +1,103 @@
...
+(* External64 values result in no caml_modify calls iff the system is 64 bit *)
+let () =
+  let f (type (t : value mod external_)) (x : t) =
+    let foo = Array.make 1 x in
+    foo.(0) <- x
+  in
+  let is_64_bit =
+    match Sys.word_size with
+    | 64 -> true
+    | 32 -> false
```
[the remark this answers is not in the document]
liam923: No

ocaml/utils/warnings.ml
liam923: This is the warning where the site that raises it is commented out because `transl_modifier` is called multiple times on the same input, causing the warning to be repeated. I guess I should do the ugly hashtable approach for now since I haven't gotten around to fixing the duplicate calls.

ocaml/testsuite/tests/typing-layouts-caml-modify/basics.ml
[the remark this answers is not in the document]
goldfirere: OK -- just leave a comment saying this isn't exercised.

ocaml/utils/warnings.ml
goldfirere: Ah, right. No, don't do that. We can just merge this as-is, and then come back and do the right thing. The hashtable approach is not super easy, and it will have its own bugs.

ocaml/testsuite/tests/typing-layouts-caml-modify/replace_caml_modify.c
```diff
@@ -0,0 +1,35 @@
+#include <caml/mlvalues.h>
+#include <stdbool.h>
+
+CAMLextern void __real_caml_modify(value *fp, value v);
+
+static bool called_modify = false;
+
+void (*replace_caml_modify_hook)(void) = NULL;
```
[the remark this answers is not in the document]
liam923: done

ocaml/testsuite/tests/typing-layouts-caml-modify/basics.ml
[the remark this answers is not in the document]
liam923: done

ocaml/testsuite/tests/typing-layouts-caml-modify/basics.ml
```diff
@@ -0,0 +1,103 @@
...
+    let foo = Array.make 1 0 in
+    foo.(0) <- 0)
+
+(* Some type definitions for below tests *)
+
+type 'a boxed_variant = Boxed of 'a
+type 'a unboxed_variant = Unboxed of 'a [@@unboxed]
+type external_variant : value mod external_ = External
+type 'a unboxed_record = { unboxed : 'a } [@@unboxed]
+type 'a external_record = { boxed : 'a }
```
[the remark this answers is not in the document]
liam923: That was a typo. Fixed

ocaml/testsuite/tests/typing-layouts-caml-modify/basics.ml
```diff
@@ -0,0 +1,103 @@
...
+let () =
+  test ~expect_caml_modify_call:false (fun () ->
+    let foo = Array.make 1 0 in
+    foo.(0) <- 0)
+
+(* Some type definitions for below tests *)
+
+type 'a boxed_variant = Boxed of 'a
+type 'a unboxed_variant = Unboxed of 'a [@@unboxed]
+type external_variant : value mod external_ = External
```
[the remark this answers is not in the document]
liam923: Done

ocaml/testsuite/tests/typing-layouts-caml-modify/basics.ml
```diff
@@ -0,0 +1,103 @@
...
+type 'a unboxed_record = { unboxed : 'a } [@@unboxed]
+type 'a external_record = { boxed : 'a }
+
+(* Internal values result in caml_modify calls *)
+let () =
+  let f (type t) (x : t) =
```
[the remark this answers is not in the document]
liam923: Agreed, but sadly this isn't the case. I left a test demonstrating this along with a CR

ocaml/typing/typemodifier.mli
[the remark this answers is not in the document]
liam923: done

ocaml/typing/typemodifier.mli
```diff
@@ -0,0 +1,14 @@
+type modifiers =
```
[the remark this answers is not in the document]
liam923: The absence of a modifier is expressed via `None` rather than `Some max`. While this is not strictly necessary for the only user of this (`jkind.ml`), it makes the interface a little more general.

ocaml/typing/typemodifier.ml
```diff
@@ -0,0 +1,173 @@
...
+  let rec loop acc = function
+    | [] -> acc
+    | (name, axis_pair) :: tl -> loop (Str_map.add name axis_pair acc) tl
+  in
+  loop Str_map.empty alist
+
+(* Raise an error if the user duplicated annotations along an axis, and a warning
+   if they used a top. The warning is only output when annot_type is Modifier *)
+let check_annot (type a) ~(annot_type : annot_type) ~(axis : a axis) ~old
+    ~(new' : a) ~(new_raw : _ Location.loc) =
```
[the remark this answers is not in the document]
liam923: done

ocaml/typing/typemodifier.ml
```diff
@@ -0,0 +1,173 @@
...
+      "non_null", Axis_pair (Nonmodal Nullability, Nullability.Non_null);
+      "internal", Axis_pair (Nonmodal Externality, Externality.Internal);
+      "external64", Axis_pair (Nonmodal Externality, Externality.External64);
+      "external_", Axis_pair (Nonmodal Externality, Externality.External) ]
+  in
+  let rec loop acc = function
+    | [] -> acc
+    | (name, axis_pair) :: tl -> loop (Str_map.add name axis_pair acc) tl
+  in
+  loop Str_map.empty alist
```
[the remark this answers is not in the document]
liam923: done

ocaml/typing/typemodifier.ml
```diff
@@ -0,0 +1,173 @@
...
+      | Portability ->
+        { acc.modal_upper_bounds with portability = Some modifier }
+      | Contention -> { acc.modal_upper_bounds with contention = Some modifier }
+    in
+    { acc with modal_upper_bounds }
+  | Nonmodal Externality -> { acc with externality_upper_bound = Some modifier }
+  | Nonmodal Nullability -> { acc with nullability_upper_bound = Some modifier }
+
+let transl_annots (annot_type : annot_type) annots =
+  let rec loop acc = function
```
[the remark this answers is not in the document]
liam923: done

ocaml/typing/typemodifier.ml
```diff
@@ -0,0 +1,173 @@
...
+      match axis with
+      | Modal Areality -> dprintf "locality"
+      | Modal Linearity -> dprintf "linearity"
+      | Modal Uniqueness -> dprintf "uniqueness"
+      | Modal Portability -> dprintf "portability"
+      | Modal Contention -> dprintf "contention"
+      | Nonmodal Externality -> dprintf "externality"
+      | Nonmodal Nullability -> dprintf "nullability"
+    in
+    fprintf ppf "The %t axis has already been specified." ax
```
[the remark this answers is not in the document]
liam923: done

ocaml/typing/typemodifier.ml
```diff
@@ -0,0 +1,173 @@
...
+  in
+  let empty_modifiers =
+    { modal_upper_bounds = Mode.Alloc.Const.Option.none;
+      externality_upper_bound = None;
+      nullability_upper_bound = None
+    }
+  in
+  loop empty_modifiers annots
+
+let transl_mode_annots annots = (transl_annots Mode annots).modal_upper_bounds
```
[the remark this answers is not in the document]
liam923: done

ocaml/typing/jkind_types.mli
```diff
@@ -147,6 +147,7 @@ type 'type_expr t =
     has_warned : bool
   }
 
+(** CR layouts v2.8: remove this when prining is improved *)
```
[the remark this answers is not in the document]
liam923: fixed

ocaml/typing/jkind_types.mli
```diff
@@ -156,4 +157,5 @@ module Const : sig
     }
 end
 
+(** CR layouts v2.8: remove this when prining is improved *)
```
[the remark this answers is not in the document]
liam923: fixed

ocaml/typing/jkind.mli
```diff
@@ -175,18 +158,16 @@ module Const : sig
...
   val get_modal_upper_bounds : t -> Mode.Alloc.Const.t
 
   (** Gets the maximum mode on the externality axis for types of this constant jkind. *)
   val get_externality_upper_bound : t -> Externality.t
 
-  module Primitive : sig
+  val of_user_written_annotation :
+    context:History.annotation_context -> Jane_syntax.Jkind.annotation -> t
+
+  module Builtin : sig
```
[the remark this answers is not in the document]
liam923: done

ocaml/typing/jkind.mli
```diff
@@ -268,42 +239,16 @@ module Primitive : sig
...
-
-  (** This is the jkind of unboxed 64-bit integers. They have sort Bits64. Does
-    not mode-cross. *)
-  val bits64 : why:History.bits64_creation_reason -> t
 end
 
 (** Take an existing [t] and add an ability to mode-cross along all the axes. *)
 val add_mode_crossing : t -> t
 
+(** Take an existing [t] and add an ability to mode-cross across the nullability axis. *)
```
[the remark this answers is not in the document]
liam923: done

ocaml/typing/jkind.mli
```diff
@@ -332,10 +277,9 @@ val of_new_legacy_sort : why:History.concrete_legacy_creation_reason -> t
 
 val of_const : why:History.creation_reason -> Const.t -> t
 
-val const_of_user_written_annotation :
-  context:History.annotation_context -> Jane_syntax.Jkind.annotation -> Const.t
+(* CR layouts v2.8: remove this when prining is improved *)
```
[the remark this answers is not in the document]
liam923: done

ocaml/utils/warnings.ml
liam923: done

ocaml/utils/warnings.ml
liam923: done

ocaml/testsuite/tests/typing-layouts/layout_poly.ml
```diff
@@ -536,8 +534,6 @@ Error: Attribute [@layout_poly] can only be used on built-in primitives.
 external[@layout_poly] id : ('a : any) ('b : any). 'a -> 'b = "%identity"
 let f (x: float#): int64# = id x
 
-(* CR layouts v2.8: The jkind in the error message is wrong. It should really be
```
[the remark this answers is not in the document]
liam923: Agreed. I put it back

ocaml/testsuite/tests/typing-layouts/layout_poly.ml
```diff
@@ -77,16 +77,14 @@ module M = struct
   let () = Format.printf "%f %s\n" (F.to_float (id' #1.)) (id' "abc")
 end
 
-(* CR layouts v2.8: The jkind in the error message is wrong. It should really be
```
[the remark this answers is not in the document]
liam923: Agreed. I put it back.

ocaml/testsuite/tests/typing-layouts-arrays/basics.ml
```diff
@@ -271,16 +271,14 @@ module M6_2 = struct
   let f2 idx : int32# = get arr idx
 end
 
-(* CR layouts v2.8: The jkind in the error message is wrong. It should really be
-   ('a : layout float64) *)
```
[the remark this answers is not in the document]
liam923: Agreed. I put it back.

ocaml/typing/typemode.ml
```diff
@@ -1,67 +1,11 @@
...
 
-type error =
-  | Duplicated_mode : ('a, 'b) Axis.t -> error
-  | Unrecognized_mode of string
-  | Unrecognized_modality of string
+type error = Unrecognized_modality of string
 
 exception Error of Location.t * error
 
-let transl_mode_annots modes =
```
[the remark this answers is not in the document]
liam923: I'll open another PR to do this.

PR #2679, opened 2024-06-12
Minor cleanup of kind checking in ctype

ccasin:
This simplifies `type_jkind_sub` to just take a jkind to check against, rather than a higher-order parameter telling it what check it should do.  The reason for the previous approach was to allow this function to be used for general kind checking (where the higher order parameter compared two jkinds) or just externality checking (where the higher order parameter compared just the externalities of the kinds).

That complexity is unnecessary - externality checking can be implemented as general kind checking with the appropriate kind (`any mod <externality>`).  And this has benefits: in the short term, it means only one function (`check_type_jkind`) needs to understand the somewhat complicated result type of `type_jkind_sub`.  And in the medium term it will simplify the implementation of product layouts, where `type_jkind_sub` wants to traverse the layout deeply.

(This is orthogonal to #2676 and looking quickly the small jkind edit here doesn't overlap with that PR so it shouldn't cause a conflict)

### DIFF
--- a/ocaml/typing/ctype.ml
+++ b/ocaml/typing/ctype.ml
@@ -2136,10 +2136,11 @@ type type_jkind_sub_result =
   | Missing_cmi of Jkind.t * Path.t
   | Failure of Jkind.t
 
-let type_jkind_sub env ty ~check_sub =
+let type_jkind_sub env ty jkind =
   let shallow_check ty =
     match estimate_type_jkind env ty with
-    | Jkind ty_jkind -> if check_sub ty_jkind then Success else Failure ty_jkind
+    | Jkind ty_jkind ->
+      if Jkind.sub ty_jkind jkind then Success else Failure ty_jkind
     | TyVar (ty_jkind, ty) -> Type_var (ty_jkind, ty)
   in
   (* The "fuel" argument here is used because we're duplicating the loop of
@@ -2165,7 +2166,7 @@ let type_jkind_sub env ty ~check_sub =
           try (Env.find_type p env).type_jkind
           with Not_found -> Jkind.any ~why:(Missing_cmi p)
         in
-        if check_sub jkind_bound
+        if Jkind.sub jkind_bound jkind
         then Success
         else if fuel < 0 then Failure jkind_bound
         else begin match unbox_once env ty with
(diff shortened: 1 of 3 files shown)


ocaml/typing/ctype.ml
```diff
@@ -2221,14 +2221,12 @@ let () =
   Env.constrain_type_jkind := constrain_type_jkind
 
 let check_type_externality env ty ext =
-  let check_sub ty_jkind =
-    Jkind.(Externality.le (get_externality_upper_bound ty_jkind) ext)
+  let upper_bound =
```
goldfirere, line 2224: Probably worth memoizing this. This function appears to be called a lot in transl. (This was true of my old implementation, too.)

ocaml/typing/ctype.ml
```diff
@@ -2221,14 +2221,12 @@ let () =
...
+    Jkind.set_externality_upper_bound (Jkind.any ~why:Dummy_jkind) ext
```
goldfirere, line 2225: This new version will be a little slower, because it has a bunch more checks to make. It probably doesn’t matter, and am happy for you to resolve if you think the benefit of this patch is worth the runtime cost.

goldfirere: On the flip side, the old implementation had to make a closure in the `constrain_type_jkind` case, and now there’s no allocation there. My guess is that we actually care about externality more often than about kinds in general, but I could be wrong there.

goldfirere (approved):

ocaml/typing/ctype.ml
ccasin: I agree there is some cost here. The check is pretty cheap even in the new world, though, and some refactoring here is necessary to support product layouts sensibly.

This PR is an intermediate step - I think when we discussed this a few weeks ago we concluded the cleanest thing may be for `constrain_type_jkind` to handle the modal upper bounds separately from the layout, which may well end up exposing a cheaper way to do this check. I'll leave a CR here to remember to revisit after product layouts.

ocaml/typing/ctype.ml
ccasin: The lifetime of this thing is just the next function call, so I strongly expect this to always be a minor allocation and therefore ~free.  Memoizing things like the core kinds in jkind.ml makes more sense, because their lifetime might be all of typechecking.  But maybe my intuition is wrong here?

ccasin: (The above remark is wrong, we might end up storing the thing in a tyvar as a result of this check.  I will change this,)

ccasin: Actually I believe the first remark (this uses `check_type_jkind`, not `constrain_type_jkind`, so it won't end up in the tyvar) - sorry for the noise, a lot going on today.

goldfirere: Yes, OK.

ocaml/typing/ctype.ml
goldfirere: Yes, OK.

PR #2682, opened 2024-06-12
Ghost regions etc. (new version of "Remove unnecessary regions for %int_as_pointer")

mshinwell:
This is a new version of https://github.com/ocaml-flambda/flambda-backend/pull/2677 using the double region variable and ghost region approach.  This seems to work pretty well, not least because it's shown up two concepts that I think were previously being conflated: these can be seen now as `Alloc_mode.For_applications.t` and `Alloc_mode.For_allocations.t`.


middle_end/flambda2/to_cmm/to_cmm_primitive.ml
```diff
@@ -767,8 +770,10 @@ let unary_primitive env res dbg f arg =
         ~else_dbg:dbg )
   | Is_flat_float_array ->
     None, res, C.eq ~dbg (C.get_tag arg dbg) (C.floatarray_tag dbg)
-  | End_region | End_try_region ->
+  | End_region { ghost = false } | End_try_region { ghost = false } ->
     None, res, C.return_unit dbg (C.endregion ~dbg arg)
+  | End_region { ghost = true } | End_try_region { ghost = true } ->
+    None, res, C.unit ~dbg
```
Gbury, line 776: I'm not sure `unit` is the correct thing here. Considering that it is supposed to return a region, something along the lines of `1` would be better I'd think. That being said, we could also ignore/erase ghost regions at that point if that is the goal.

middle_end/flambda2/to_cmm/to_cmm_primitive.ml
```diff
@@ -658,7 +658,10 @@ let nullary_primitive _env res dbg prim =
        `to_cmm_shared.ml` *)
     let expr = Cmm.Cop (Cprobe_is_enabled { name }, [], dbg) in
     None, res, expr
-  | Begin_region | Begin_try_region -> None, res, C.beginregion ~dbg
+  | Begin_region { ghost = false } | Begin_try_region { ghost = false } ->
+    None, res, C.beginregion ~dbg
+  | Begin_region { ghost = true } | Begin_try_region { ghost = true } ->
+    None, res, C.unit ~dbg
```
Gbury, line 664: I'm not sure unit is the correct thing here. Considering that it is supposed to return a region, something along the lines of 1 would be better I'd think. That being said, we could also ignore/erase ghost regions at that point if that is the goal.

Gbury: I think it's safe, but I'd still expect the cmm expression returned to, at least morally, be something of the correct type/kind for a region. In the case of regions, that means a pointer (usually that points inside the local allocation stack), so I think returning the null pointer / `0` would be better here.

Alternatively, this case could simply raise (at compilation time) if we want to guarantee that these regions are correctly eliminated (but that would probably require changes to `to_cmm_expr` to not try and produce cmm code for these primitives).

middle_end/flambda2/parser/fexpr_to_flambda.ml
```diff
@@ -1098,5 +1110,6 @@ let conv comp_unit (fexpr : Fexpr.flambda_unit) : Flambda_unit.t =
   let env = bind_all_code_ids env fexpr in
   let body = expr env fexpr.body in
   Flambda_unit.create ~return_continuation ~exn_continuation
-    ~toplevel_my_region:toplevel_region ~body ~module_symbol
-    ~used_value_slots:Unknown
+    ~toplevel_my_region:toplevel_region
+    ~toplevel_my_ghost_region:toplevel_region (* XXX *)
```
Ekdohibs, line 1114: This should be fixed

middle_end/flambda2/from_lambda/lambda_to_flambda.ml
```diff
@@ -499,7 +506,9 @@ let restore_continuation_context acc env ccenv cont ~close_current_region_early
   let[@inline] normal_case env acc ccenv =
     match Env.pop_regions_up_to_context env cont with
     | None -> body acc ccenv cont
-    | Some region ->
+    | Some { region; ghost_region = _ } ->
+      (* XXX mshinwell: maybe add a comment as to why [ghost_region] isn't
+         relevant (reviewer please check this is indeed the case) *)
```
Ekdohibs, line 511: I'm not confident about this

middle_end/flambda2/from_lambda/lambda_to_flambda.ml
```diff
@@ -520,18 +529,24 @@ let restore_continuation_context acc env ccenv cont ~close_current_region_early
...
+          (End_region
+             { is_try_region = false; region = ghost_region; ghost = true })
+          ~body:(fun acc ccenv -> normal_case env acc ccenv))
   else normal_case env acc ccenv
 
 let restore_continuation_context_for_switch_arm env cont =
   match Env.pop_regions_up_to_context env cont with
   | None -> cont
-  | Some region ->
+  | Some { region; ghost_region = _ } ->
```
Ekdohibs, line 549: Here as well.

middle_end/flambda2/to_cmm/to_cmm_primitive.ml
mshinwell: I've changed this to use `int 0`, that should be fine for now.

Ekdohibs (approved):

PR #1548, opened 2023-07-07
Syntactic function arity parsing

ncik-roberts:
This is the parsing half of syntactic function arity. Unlike the upcoming typechecking/translation PR, this PR does not change the semantics of any existing OCaml program. But it does place new requirements on ppxes, so we should wait to merge it until some internal PRs for ppxes land.

@goldfirere for review please.

This PR transplants bits of the upstream PR (ocaml/ocaml#12236) to flambda-backend.

At a high level:
- The OCaml parser now keeps track of function arity, encoding this in the parsetree using the Jane Syntax framework.
- The sanity checks that run on ASTs produced by ppxes now require that functions were encoded using the Jane Syntax framework.
- For now, typechecking drops this arity information. (It constructs another parsetree that doesn't have any arity information and typechecks that.) 

### How to review
- Review the first two commits (`Add some locals tests` and `Make Jane Syntax treat attributes as inner-to-outer`) in isolation. See "Things of note".
- Then look at the entire diff of the PR.

This PR does not share much code with the upstream PR, except for some pieces of the parser.

#### Making Jane Syntax treat attributes as inner-to-outer

One of the early commits ensures the following things hold:
  1. Jane Syntax should interpret later attributes before earlier attributes. ("Later" and "earlier" are with respect to the order of attributes in, say, the `pexp_attributes` list.)
  2. When the user writes something like `[: :] [@attr]`, the `[@attr]` attribute should be placed later in the list than any Jane Syntax attribute.

(1.) did not hold previously, but (2.) did. This PR makes both properties hold. It's important for (1.) to hold because, without it, this PR would be incompatible with the way that ppxlib merges attributes. For example, suppose a ppx replaces `[%ppx_extension]` with `fun y z -> y + z`, which is encoded in Jane Syntax as `(fun[@jane.erasable._builtin] y -> fun[@jane.erasable._builtin.param] z -> (y + z)[@jane.erasable._builtin.end])`. Then,

```ocaml
(* source program *)
(fun x -> [%ppx_extension])

(* encoding as Jane Syntax parsetree when handed to the ppx *)
(fun[@jane.erasable._builtin] x -> [%ppx_extension] [@jane.erasable._builtin.end])

(* result of running ppx, keeping in mind that ppxlib merges attributes by placing outer attributes after
   inner ones *)
[...]

PR #2683, opened 2024-06-12
Fix warning 34 (unused type declaration) for locally-abstract types

ncik-roberts:
Fix a bug introduced in #1548: we unintentionally stopped warning on some unused locally-abstract types. Add a test to ensure this doesn't happen again in the future.

The bug exists upstream too.

The source of the bug is that the "unused X" check in `env.ml` does not run if the location is marked as ghost. #1548 makes more `Pexp_newtypes` ghost. (I think they should be marked ghost: the range of text indicated by the location in the source program is not something that can be called a "newtype expression", and this is the usual criteria we use. But the fact of the matter is that the compiler uses ghostiness as a proxy for other properties, and sometimes there is a sharp edge like this one.)

The fix I take is to use the identifier's location (which is never ghost) for purposes of this check, not the `Pexp_newtype`'s location.

**Review**: the first commit adds a regression test with bad old output, and the second commit fixes the bug and updates the test output.

### DIFF
--- a/ocaml/testsuite/tests/typing-warnings/unused_types.ml
+++ b/ocaml/testsuite/tests/typing-warnings/unused_types.ml
@@ -535,14 +535,19 @@ module Unused_field_disable_one_warning : sig end
 
 let u (type unused) = ()
 [%%expect {|
+Line 1, characters 12-18:
+1 | let u (type unused) = ()
+                ^^^^^^
+Warning 34 [unused-type-declaration]: unused type unused.
+
 val u : unit = ()
 |}]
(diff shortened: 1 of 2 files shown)


ccasin (approved):
This looks fine, but Nick is going to test it on the tree before merging to guard against the possibility some ppx is relying on being able to make this location ghost to avoid the warning.

goldfirere (approved):

ncik-roberts:
My testing didn't turn up any ppx that relies on this.

PR #2686, opened 2024-06-13
Improved reinterpret casts for integers and floats

mshinwell:
This first replaces the existing Flambda 2 `Reinterpret_int64_as_float` primitive, which is a no-op modulo a move between register classes, with a more flexible reinterpret operation parameterized as follows:
```
  type t =
    | Tagged_int63_as_unboxed_int64
    | Unboxed_int64_as_tagged_int63
    | Unboxed_int64_as_unboxed_float64
    | Unboxed_float64_as_unboxed_int64
```
These are only supported for 64-bit targets.  The semantics are as follows:
- `Tagged_int63_as_unboxed_int64`: no-op
- `Unboxed_int64_as_tagged_int63`: logical OR with 1, to ensure the bottom bit is set
- `Unboxed_int64_as_unboxed_float64`: move between register classes
- `Unboxed_float64_as_unboxed_int64`: move between register classes

The simplification rules have been enhanced to cope with the above, allowing compile-time evaluation when arguments are known.

We also allow Flambda 2 to intercept calls to `caml_int64_bits_of_float_unboxed` (e.g. from `Int64.bits_of_float`) in the same way as the existing code for `caml_int64_float_of_bits_unboxed`, except that both of these now use the new primitive.

The two primitives involving tagged immediates can be bound as follows:
```
external i63_to_i64 : int -> int64# = "%reinterpret_tagged_int63_as_unboxed_int64"
external i64_to_i63 : int64# -> int = "%reinterpret_unboxed_int64_as_tagged_int63"
```


ccasin (commented):
Looks OK for everything under `ocaml`, modulo questions below.  Needs tests though.

middle_end/flambda2/from_lambda/closure_conversion.ml
```diff
@@ -623,7 +616,15 @@ let close_c_call acc env ~loc ~let_bound_ids_with_kinds
...
         | _, _ ->
           Misc.fatal_errorf "Wrong argument and/or result kind(s) for %s"
-            prim_native_name)
+            prim_native_name
+    in
+    match prim_native_name with
+    | "caml_int64_float_of_bits_unboxed" ->
+      unboxed_int64_to_and_from_unboxed_float ~src_kind:(Unboxed_integer Pint64)
+        ~dst_kind:(Unboxed_float Pfloat64) ~op:Unboxed_int64_as_unboxed_float64
+    | "caml_int64_bits_of_float_unboxed" ->
```
TheNumbat, line 625: It's a bit unfortunate to add another secret builtin. What do you think about making these primitives, or requiring is_c_builtin? 

Ideally, I think we'd assure that %primitives are specialized before the middle end, @@ builtins are optionally specialized after the middle end, and otherwise externals are always C calls.

middle_end/flambda2/simplify/simplify_unary_primitive.ml
```diff
@@ -429,24 +429,110 @@ let simplify_boolean_not dacc ~original_term ~arg:_ ~arg_ty ~result_var =
...
+end)
+
+module Simplify_reinterpret_unboxed_int64_as_tagged_int63 =
+Make_simplify_reinterpret_64_bit_word (struct
+  module Src = Int64
+  module Dst = Targetint_31_63
+
+  let prover = T.meet_naked_int64s
+
+  let convert i = Targetint_31_63.of_int64 (Int64.div i 2L)
```
TheNumbat, line 500: Should this be or with 1?

middle_end/flambda2/simplify/simplify_unary_primitive.ml
```diff
@@ -429,24 +429,110 @@ let simplify_boolean_not dacc ~original_term ~arg:_ ~arg_ty ~result_var =
...
+end)
+
+module Simplify_reinterpret_tagged_int63_as_unboxed_int64 =
+Make_simplify_reinterpret_64_bit_word (struct
+  module Src = Targetint_31_63
+  module Dst = Int64
+
+  let prover = T.meet_equals_tagged_immediates
+
+  let convert i = Int64.add (Int64.mul (Targetint_31_63.to_int64 i) 2L) 1L
```
TheNumbat, line 514: Should this be identity?

middle_end/flambda2/terms/flambda_primitive.ml
```diff
@@ -1042,6 +1042,27 @@ let nullary_classify_for_printing p =
   | Begin_try_region | Enter_inlined_apply _ | Dls_get ->
     Neither
 
+module Reinterpret_64_bit_word = struct
+  type t =
+    | Tagged_int63_as_unboxed_int64
+    | Unboxed_int64_as_tagged_int63
+    | Unboxed_int64_as_unboxed_float64
+    | Unboxed_float64_as_unboxed_int64
```
TheNumbat, line 1050: Should I lift the float32 conversions to this level as well? Currently float32<->int32-bits is a @@ builtin.

ocaml/lambda/lambda.mli
```diff
@@ -262,6 +262,8 @@ type primitive =
   | Pbox_float of boxed_float * alloc_mode
   | Punbox_int of boxed_integer
   | Pbox_int of boxed_integer * alloc_mode
+  | Preinterpret_unboxed_int64_as_tagged_int63 (** Only for 63-bit targets *)
```
TheNumbat, line 265: 64-bit targets

ocaml/lambda/translprim.ml
```diff
@@ -1450,7 +1454,8 @@ let lambda_primitive_needs_event_after = function
   | Punboxed_float_array_set_128 _ | Punboxed_int32_array_set_128 _
   | Punboxed_int64_array_set_128 _ | Punboxed_nativeint_array_set_128 _
   | Prunstack | Pperform | Preperform | Presume
-  | Pbbswap _ | Pobj_dup | Pget_header _ -> true
+  | Pbbswap _ | Pobj_dup | Pget_header _
+  | Preinterpret_tagged_int63_as_unboxed_int64 -> true
```
TheNumbat, line 1458: This is because unboxing is ignored in bytecode, right?
Would be good to update the comment at the top of this function.

TheNumbat (commented):
Looks good; left some comments. Are there tests for the two new %primitives?

ocaml/lambda/translprim.ml
mshinwell: Yeah I think I've made a mistake here looking at Chris's comments, will check - but yes, that is the intention.

middle_end/flambda2/terms/flambda_primitive.ml
mshinwell: I think we probably should, in a follow-up PR.

middle_end/flambda2/from_lambda/closure_conversion.ml
mshinwell: I think I'd prefer to leave the PR as-is, since we already have references to these builtins in (at least) the stdlib, but we could consider changing these to primitives later.

In fact I think some of the `@@builtin`s should probably be recognised by flambda2, but that's a separate discussion.

mshinwell:
I've added a test case, which showed up a bug on negative inputs to the i64 -> i63 conversion!
@TheNumbat please approve + merge

TheNumbat (approved):

PR #2688, opened 2024-06-13
Improve parsing of layout annotations in type parameters

ccasin:
A user pointed out that it's confusing and annoying that we allow you to write a type declaration like:
```ocaml
type ('a, 'b : float64) t
```
But if you want to annotate the parameter on some use of `t`, you need more parens:
```ocaml
val foo : ('a, ('b : float64)) t -> int
```
This allows to skip the parens in the second example.  There are three commits:
- Commit one does some drive by cleanup of the relevant test file - we used to need more copies of this test when the parser cared about the layouts extension level more, but we can combine things nicely now.
- Commit two adds some tests that don't parse, but could.
- Commit three makes them parse.

ocamlformat changes will be needed.  I'll do it.


ocaml/parsing/parser.mly
```diff
@@ -4536,9 +4536,26 @@ atomic_type:
...
+    tys = separated_nontrivial_llist(COMMA, one_type_parameter_of_several)
+    RPAREN
       { tys }
 
+(* Layout annotations on type expressions typically require parens, as in [('a :
+   float64)].  But this is unnecessary when the type expression is used as the
+   parameter of a tconstr with more than one argument, as in [(int, 'b :
+   float64) t].  Further, it would be confusing to require parens there because
+   they are not required in that position in type declarations.  So, we have a
+   special case. *)
```
ncik-roberts, line 4549: Nit: I think this comment is just as informative if you leave off the last two sentences. I don't have a strong preference.

ocaml/parsing/parser.mly
```diff
@@ -4536,9 +4536,26 @@ atomic_type:
...
+%inline one_type_parameter_of_several:
+  | core_type { $1 }
+  | QUOTE id=ident COLON jkind=jkind_annotation
+    { Jane_syntax.Layouts.type_of ~loc:(make_loc $sloc) @@
+      Ltyp_var { name = Some id; jkind } }
+  | UNDERSCORE COLON jkind=jkind_annotation
+    { Jane_syntax.Layouts.type_of ~loc:(make_loc $sloc) @@
+      Ltyp_var { name = None; jkind } }
```
ncik-roberts, line 4557: Could we use `type_param_with_jkind` directly instead? Sharing this code also lets the user put an attribute on the type var (or underscore) which seems nice.

ncik-roberts (approved):
I agree this is nice to support.

ocaml/parsing/parser.mly
ccasin: Done.

ocaml/parsing/parser.mly
ccasin: This is a good idea, and I missed that we should allow attributes here.  Unfortunately, allowing them leads to a bunch of parser conflicts that I don't immediately see how to fix.  I think the most prudent course is to avoid spending further time on this - people can add parens if the need attributes.

PR #2691, opened 2024-06-13
Stop using exceptions for control flow in `camlinternalOO`

ncik-roberts:
We have some evidence that doing this will speed up uses of the object system in JSOO.

I don't see a compelling reason to use exceptions in `ocaml/stdlib/camlinternalOO.ml`, and think using options is fine. I don't think this will affect bytecode or native compilation meaningfully. Perhaps this will add some more short-lived allocations for uses of the object system, but I don't think it's meaningfully more, judging from the surrounding code.

### DIFF
--- a/ocaml/stdlib/camlinternalOO.ml
+++ b/ocaml/stdlib/camlinternalOO.ml
@@ -295,9 +295,7 @@ let new_methods_variables table meths vals =
   res
 
 let get_variable table name =
-  match Vars.find_opt name table.vars with
-  | Some x -> x
-  | None -> assert false
+  try Vars.find name table.vars with Not_found -> assert false
 
 let get_variables table names =
   Array.map (get_variable table) names
--- a/ocaml/stdlib/camlinternalOO.ml
+++ b/ocaml/stdlib/camlinternalOO.ml
@@ -200,8 +200,9 @@ let set_method table label element =
     table.hidden_meths <- (label, element) :: table.hidden_meths
 
 let get_method table label =
-  try List.assoc label table.hidden_meths
-  with Not_found -> table.methods.(label)
+  match List.assoc_opt label table.hidden_meths with
+  | Some x -> x
+  | None -> table.methods.(label)
 
 let to_list arr =
   if arr == Obj.magic 0 then [] else Array.to_list arr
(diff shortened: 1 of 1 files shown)


lthls:
I'm surprised you can observe any difference. All of these functions are called in parts of the code that are evaluated once per class (not even once per class created at runtime, but once per class in the source code). So if you notice a difference with this PR, it means you're spending significantly more time creating classes than actually using them.
I don't mind the PR though, I think options are better than exceptions for these cases anyway.

TyOverby:
@lthls: In applications that are compiled to javascript via Js_of_ocaml, our benchmarks show that throwing and catching an exception can cost as much as 2ms.  In a program that had three-level inheritence chain on a 50-method class, we saw that over one second was spent throwing and catching exceptions in `camlinternalOO.ml` at program startup time.

ocaml/stdlib/camlinternalOO.ml
```diff
@@ -291,7 +295,9 @@ let new_methods_variables table meths vals =
   res
 
 let get_variable table name =
-  try Vars.find name table.vars with Not_found -> assert false
+  match Vars.find_opt name table.vars with
+  | Some x -> x
+  | None -> assert false
```
lthls, line 300: This part of the patch seems less useful, since no exceptions are ever raised. Although setting up the exception handler might already be expensive enough to justify the switch to options.

lthls (approved):
The PR looks correct and useful. I'm not too worried about the extra allocations.

lthls:
> @lthls: In applications that are compiled to javascript via Js_of_ocaml, our benchmarks show that throwing and catching an exception can cost as much as 2ms. In a program that had three levels of inheritence on a 50-method class, we saw that over one second was spent throwing and catching exceptions in `camlinternalOO.ml` at program startup time.

That makes sense, thanks for the context.

chambart:
Have you tried using OCamlClean https://github.com/bvaugon/ocamlclean ? It does partially evaluate bytecode programs. I know that it is supposed to be quite good with class initialisation. It has not been updated recently but you could at least try on your files to see if it reduces the size. Also it might not right away with JSOO.

ocaml/stdlib/camlinternalOO.ml
ncik-roberts: Good point. I don't believe that setting up the handler is the expensive part.

ncik-roberts: (This is now reverted.)

PR #2694, opened 2024-06-14
Advise against new bytecode instructions

ncik-roberts:
Add a comment that urges caution when adding a new bytecode instruction. (It's not impossible &mdash; we've done it &mdash; but it shouldn't be done lightly.)


mshinwell (approved):
The comment is fine but I think you might have broken a horrible script.
OK to merge once that is fixed

ncik-roberts:
A good reminder that one should always wait for CI to pass, even if you've just added a comment.

PR #2697, opened 2024-06-14
Float32 array load/store primitives

TheNumbat:
Adds %primitives for loading/storing float32s to/from strings, bytes, bigstrings, and float32_elt bigarrays. Also covers loading/storing unboxed float32s to/from bigstrings.

The string, bytes, and bigstring primitives have the same semantics as those for int32s; the only difference is that they operate on a float32 register directly. This did not require any compiler backend changes.

The bigarray functions are specialized copies of `caml_ba_ref` / `caml_ba_set` for float32. They require bigarrays with the `float32_elt` element kind, but purposefully ignore the bigarray's 'view' type. This allows the API functions to operate on e.g. `(float, float32_elt, 'c) t` bigarrays.
- This required adding a new bigarray kind in the compiler, `Pbigarray_float32_t` / `Float32_t`. This kind not exposed by the standard library, and is only used by the new primitives.
- When generating calls to generic `caml_ba_get_$N` / `caml_ba_set_$N`, we now need to check whether the kind is `Pbigarray_float32_t`. This was necessary because the default runtime functions will convert `float32_elt`s to doubles.
- Only 1, 2, and 3 dimensional `float32_elt` bigarrays are supported. Generic arrays must still convert through float.

Tests for all new primitives have been added to `float32_lib.ml` and `float32_u_lib.ml`.


mshinwell:
@TheNumbat please add a full PR description

TheNumbat:
Yeah, added, but this still isn't ready for review. Something bad is going on with the cfg backends

backend/x86_gas.ml
```diff
@@ -138,7 +138,7 @@ let print_instr b = function
   | CQO ->  i0 b "cqto"
   | CVTSS2SI (arg1, arg2) -> i2 b "cvtss2si" arg1 arg2
   | CVTSD2SI (arg1, arg2) -> i2 b "cvtsd2si" arg1 arg2
-  | CVTSI2SS (arg1, arg2) -> i2 b "cvtsi2ss" arg1 arg2
+  | CVTSI2SS (arg1, arg2) -> i2 b ("cvtsi2ss" ^ suf arg1) arg1 arg2
```
xclerc, line 141: Shouldn't that imply a similar change
in `x86_masm.ml`?

backend/cfg/cfg_equivalence.ml
```diff
@@ -204,6 +204,9 @@ let check_operation : location -> Cfg.operation -> Cfg.operation -> unit =
     ()
   | Const_float expected, Const_float result when Int64.equal expected result ->
     ()
+  | Const_float32 expected, Const_float32 result
```
xclerc, line 207: (As showed by https://github.com/ocaml-flambda/flambda-backend/pull/2719, this should not have any effects.)

backend/x86_gas.ml
TheNumbat: Discussed (probably not, but we're not actually testing masm...)

ccasin (approved):
Approving for `ocaml/{lambda,typing,otherlibs}`.  As discussed in person / below, I think you should get other reviewers for some of the other bits.

ocaml/runtime/float32.c
```diff
@@ -419,6 +420,136 @@ CAMLprim value caml_modf_float32(value f)
   CAMLreturn (res);
 }
 
+CAMLprim value caml_string_getf32(value str, value index)
```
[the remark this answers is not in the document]
mshinwell: We don't support big-endian systems on flambda-backend, so it's ok for these to fail in some way.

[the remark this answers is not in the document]
TheNumbat: Ok, I've made all of these functions raise on big-endian architectures.
The tests are now run in bytecode mode as well.

mshinwell:
Can we please pull out the fix for `cvtsi2ss` into a separate PR?

TheNumbat:
Sure; #2758 - xclerc has already reviewed the fix

mshinwell:
@ccasin did you read the tests?  If not please do.  I am reading everything else that you didn't.

mshinwell (approved):
This is OK to merge once the tests have been reviewed.

ccasin:
Yes, I read the tests, so this seems fine to merge once CI passes.

PR #2700, opened 2024-06-19
Make `-error_size` directly about length of printed error

riaqn:
Currently the compiler accepts a `-error_size` argument which takes an integer (defaults to 500) as the limit on error message context. Error context longer than that will be printed as `...` instead.

However, the limit is not on the printing, but on the marshaled binary. As we introduce complex data structures to  `value_description`, marshaled size is no longer a good indication of printed size.

This PR replaces that with real printing: we now count the length of the printing and compare that with `error_size`. Indentation is not counted, but spaces and newlines are.

I've decided that with the new metric, `error_size` defaulting to 256 is a good balance point. Some tests are updated as a result.


ocaml/testsuite/tests/typing-modules/inclusion_errors_elision.ml
```diff
@@ -27,7 +27,13 @@ Lines 9-13, characters 15-3:
...
 13 | end
 Error: Signature mismatch:
-       ...
+       Modules do not match:
+         sig
+           module B :
+             sig type a and b and c and d and e and f and g and h end
+         end
+       is not included in
+         S
```
ncik-roberts, line 36: The tests in this file are supposed to show that ellision works as expected, and this test no longer appears to show that. Can you change it to show that?

ocaml/typing/includemod_errorprinter.ml
```diff
@@ -576,7 +570,7 @@ let dwith_context ?loc ctx printer =
   Location.msg ?loc "%a%t" Context.pp (List.rev ctx) printer
 
 let dwith_context_and_elision ?loc ctx printer diff =
-  if is_big (diff.got,diff.expected) then
+  if is_big (printer diff) then
     Location.msg ?loc "..."
   else
     dwith_context ?loc ctx (printer diff)
```
ncik-roberts, line 576: It looks like we're calling `printer diff` twice now. I don't think it's actually expensive -- it's the partial application of a function -- but one could be misled by looking at the code. I suggest something like:

```ocaml
let print_diff ppf = printer diff ppf in
if is_big print_diff then
  ...
else
  ...
```

This better suggests that it's the use sites of `print_diff` that might be expensive, not the creation site.

ocaml/utils/misc.ml
```diff
@@ -1099,6 +1099,29 @@ let output_of_print print =
...
+  let out_indent _ = () in
+  let out_functions : Format.formatter_out_functions = {
+    out_string;
+    out_flush;
+    out_newline;
+    out_spaces;
+    out_indent}
+  in
+  let ppf = Format.formatter_of_out_functions out_functions in
+  try ignore (p ppf); false
```
ncik-roberts, line 1123: It's good practice to type-annotate things you ignore, e.g. `ignore (p ppf : unit)`. I think currently, `is_print_bigger_than` accepts `Format.formatter -> 'a` as its second argument, and that generality could perhaps be a footgun if someone accidentally passes in a two-argument function.

ocaml/utils/misc.ml
```diff
@@ -1099,6 +1099,29 @@ let output_of_print print =
   in
   output
 
+exception LimitExceeded
```
ncik-roberts, line 1102: Consider declaring this exception within `is_print_bigger_than`, to document the fact that it's only used within that function. I'm not concerned about creating a new extensible constructor each time the function is called. You could alternatively do something like `let is_print_bigger_than = let exception Limit_exceeded in fun size p -> ...` but it's a bit less natural.

Also please name it `Limit_exceeded`.

ocaml/utils/misc.mli
```diff
@@ -630,6 +630,10 @@ val output_of_print :
     Note that naively using [Format.formatter_of_out_channel] typechecks but
     doesn't work because it fails to flush the formatter. *)
 
+val is_print_bigger_than: int -> (Format.formatter -> 'a) -> bool
+(** Returns [true] if the printing would be longer than the given integer. Stops
+    early if so. Spaces and newlines are counted, but indentation are not. *)
```
ncik-roberts, line 635:
```suggestion
val is_print_longer_than: int -> (Format.formatter -> 'a) -> bool
(** Returns [true] if the printed string is longer than the given integer. Stops
    early if so. Spaces and newlines are counted, but indentation is not. *)
```

ocaml/testsuite/tests/typing-modules/inclusion_errors_elision.ml
riaqn: The original file contains two tests, both have the eliding logic triggered, and I can't see the difference between the two tests.

The current file has two tests - one triggers and one not. Maybe that's better?

ocaml/utils/misc.ml
riaqn: Yes I was after generalization - but that's probably not needed !

ocaml/typing/includemod_errorprinter.ml
riaqn: Good point - I lift the binding higher so it's clearer.

ocaml/testsuite/tests/typing-modules/inclusion_errors_elision.ml
ncik-roberts: I think the test is meant to exercise a different code path: the one that always prints an `Mt_core` message even in the event of the starting message being elided. See https://github.com/ocaml/ocaml/commit/8218be9e2b24907b8558776a34d12032bcc42496#diff-c6741d931dca5b1ee5f6d02173f0370101a831c62b91f4c7c4c58f78bbf23de2. I think we should have both tests in this file exhibit elision.

ncik-roberts (approved):

ocaml/testsuite/tests/typing-modules/inclusion_errors_elision.ml
riaqn: OK I've added two more tests. Now we have 2x2 tests corresponding to all cases.

PR #2705, opened 2024-06-20
Join: only generate alias types to older variables

lthls:
This ensures that we don't forget to join head types when several parameters are all aliases of each other without a common alias in the environment at fork

This problem starts appearing very frequently after #2628, and was causing a number of allocations to be kept because the absence of a head type made unboxing fail.

Here is an example that doesn't involve #2628:
```ocaml
let f x y z =
  let[@local] k a b =
    a +. b
  in
  if x then
    let arg = y -. z in k arg arg
  else
    let arg = z -. y in k arg arg
```
With this PR, `a` and `b` are unboxed properly and the result contains a single allocation at the end. We still keep both variables though, I'll see if I can fix that too but I think this PR should be merged anyway (assuming it doesn't cause any regressions).

### DIFF
--- a/middle_end/flambda2/types/env/typing_env.ml
+++ b/middle_end/flambda2/types/env/typing_env.ml
@@ -599,7 +599,7 @@ let mem_simple ?min_name_mode t simple =
     ~name:(fun name ~coercion:_ -> mem ?min_name_mode t name)
     ~const:(fun _ -> true)
 
-let alias_is_bound_before t ~bound_name ~alias =
+let alias_is_bound_strictly_earlier t ~bound_name ~alias =
   let time_of_name =
     binding_time_and_mode t bound_name
     |> Binding_time.With_name_mode.binding_time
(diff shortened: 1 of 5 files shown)


middle_end/flambda2/types/env/typing_env.ml
```diff
@@ -599,6 +599,17 @@ let mem_simple ?min_name_mode t simple =
     ~name:(fun name ~coercion:_ -> mem ?min_name_mode t name)
     ~const:(fun _ -> true)
 
+let alias_is_bound_before t ~bound_name ~alias =
```
chambart, line 602: From a french point of view on before, it is not obvious that this is strictly before. Also I'm not sure but should it be 'earlier' rather than 'before' ?
So either `alias_is_bound_strictly_before` or `binding_time_of_alias_is_strictly_earlier`

chambart (approved):
This seems good and simple.

middle_end/flambda2/types/env/typing_env.ml
lthls: Renamed to `alias_is_bound_strictly_earlier`

lthls:
> We still keep both variables though, I'll see if I can fix that too but I think this PR should be merged anyway (assuming it doesn't cause any regressions).

It looks like I mistakenly ran the test without join points, in which case it is expected that we still get two unboxed variables. With join points we only keep a single unboxed variable, as expected.

PR #2708, opened 2024-06-21
Simply Env.lookup for module and class

riaqn:
#2431 enriches modules and classes with modes. However, it over-complicates `Env.lookup_module` and siblings. In particular, it doesn't need to return `closure_context` for mode error hints. This PR removes that.


lukemaurer (approved):
LGTM

PR #2710, opened 2024-06-21
Use float32 for `float32x4` scalar casts

TheNumbat:
Updates the following builtins to operate on 32-bit floats: 
- `caml_float32x4_low_to_float` 
- `caml_float32x4_low_of_float`
- `caml_float32x4_const1`
- `caml_float32x4_const4`

The new builtins take and return unboxed `float32`s. This allows us to remove the special cases in the backend for converting to/from 64-bit floats. 

The SIMD tests have been updated to reflect the new behavior.


backend/cmm_builtins.ml
```diff
@@ -195,6 +195,14 @@ let bigstring_atomic_add size (arg1, arg2, arg3) dbg =
 let bigstring_atomic_sub size (arg1, arg2, arg3) dbg =
   bigstring_atomic_add size (arg1, arg2, neg_int arg3 dbg) dbg
 
+(* Assumes unboxed float32 *)
+let rec const_float32_args n args name =
```
xclerc, line 199: (This is the fourth variant of this function;
it might be worth factoring out.)

backend/amd64/emit.mlp
```diff
@@ -1016,12 +1016,8 @@ let emit_static_cast (cast : Cmm.static_cast) i =
...
   | Scalar_of_v128 Int32x4 -> I.movd (arg i 0) (res32 i 0)
   | V128_of_scalar Int32x4 -> I.movd (arg32 i 0) (res i 0)
-  | V128_of_scalar Float32x4 ->
-    (* CR mslater: (SIMD) remove cvt once we have unboxed float32 *)
-    I.cvtsd2ss (arg i 0) (res i 0)
-  | Scalar_of_v128 Float32x4 ->
-    (* CR mslater: (SIMD) remove cvt once we have unboxed float32 *)
-    I.cvtss2sd (arg i 0) (res i 0)
+  | V128_of_scalar Float32x4 | Scalar_of_v128 Float32x4 ->
+    if distinct then I.movss (arg i 0) (res i 0)
```
xclerc, line 1020: (Arguably not part of this pull request,
but it feels like `distinct` might not be
used uniformly. Could you double
check it is always used when it can
be? For instance, shouldn't it be used
on [line 1016](https://github.com/ocaml-flambda/flambda-backend/pull/2710/files#diff-89a48d52bdba7b3a60e36cfc34c5f2f96f59da176212034bfc80e9743f9057d6R1016)?)

backend/cmm_builtins.ml
TheNumbat: Hmm, I tried combining them but I don't think it's worthwhile, since they have different types/constructors/error messages

backend/amd64/emit.mlp
TheNumbat: I think the usage is correct - the `movq` and `movd`s are always between register classes, so are always distinct

backend/cmm_builtins.ml
xclerc: I agree a patch such as the one below is not
particularly elegant, but won't we have more
copy-paste when we introduce additional
small numbers?

```diff
--- a/backend/cmm_builtins.ml
+++ b/backend/cmm_builtins.ml
@@ -195,38 +195,28 @@ let bigstring_atomic_add size (arg1, arg2, arg3) dbg =
 let bigstring_atomic_sub size (arg1, arg2, arg3) dbg =
   bigstring_atomic_add size (arg1, arg2, neg_int arg3 dbg) dbg

-(* Assumes unboxed float32 *)
-let rec const_float32_args n args name =
+let rec const_args_gen ~extract ~type_name n args name =
   match n, args with
   | 0, [] -> []
-  | n, Cconst_float32 (f, _) :: args ->
-    f :: const_float32_args (n - 1) args name
-  | _ -> bad_immediate "Did not find constant float32 arguments for %s" name
+  | _, [] -> assert false (* TODO *)
+  | n, arg :: args ->
+    begin match extract arg with
+    | Some value -> value :: const_args_gen ~extract ~type_name (n - 1) args name
+    | None -> bad_immediate "Did not find constant %s arguments for %s" type_name name
+    end
+
+(* Assumes unboxed float32 *)
+let const_float32_args = const_args_gen ~extract:(function Cconst_float32 (f, _) -> Some f | _ -> None) ~type_name:"float32"

 (* Assumes unboxed float64 *)
-let rec const_float_args n args name =
-  match n, args with
-  | 0, [] -> []
-  | n, Cconst_float (f, _) :: args -> f :: const_float_args (n - 1) args name
-  | _ -> bad_immediate "Did not find constant float arguments for %s" name
+let const_float_args = const_args_gen ~extract:(function Cconst_float (f, _) -> Some f | _ -> None) ~type_name:"float"

 (* Assumes untagged int or unboxed int32, always representable by int63 *)
-let rec const_int_args n args name =
-  match n, args with
-  | 0, [] -> []
-  | n, Cconst_int (i, _) :: args -> i :: const_int_args (n - 1) args name
-  | _ -> bad_immediate "Did not find constant int arguments for %s" name
+let const_int_args = const_args_gen ~extract:(function Cconst_int (i, _) -> Some i | _ -> None) ~type_name:"int"

 (* Assumes unboxed int64: no tag, comes as Cconst_int when representable by
    int63, otherwise we get Cconst_natint *)
-let rec const_int64_args n args name =
-  match n, args with
-  | 0, [] -> []
-  | n, Cconst_int (i, _) :: args ->
-    Int64.of_int i :: const_int64_args (n - 1) args name
-  | n, Cconst_natint (i, _) :: args ->
-    Int64.of_nativeint i :: const_int64_args (n - 1) args name
-  | _ -> bad_immediate "Did not find constant int64 arguments for %s" name
+let rec const_int64_args = const_args_gen ~extract:(function Cconst_int (i, _) -> Some (Int64.of_int i) | Cconst_natint (i, _) -> Some (Int64.of_nativeint i) | _ -> None) ~type_name:"int64"

 let int64_of_int8 i =
   (* CR mslater: (SIMD) replace once we have unboxed int8 *)

```

TheNumbat: Yeah, that's true - updated to your suggestion

xclerc (approved):

PR #2533, opened 2024-05-03
Support for mixed blocks in Flambda2 types

lthls:
As title says.
This includes support for unboxing, but not lifting/reification yet (although that shouldn't be too complicated to add on top of this PR).

There are a few details that I still find annoying but that I have left for now to avoid losing too much time. The issues are mostly about the various types used to represent information about mixed blocks. I can incorporate suggestions in this PR, but it's also something that we can change later if we want to (it's mostly mechanical).

PR #2712, opened 2024-06-24
Static allocation for mixed blocks etc.

mshinwell:
This is based on #2692 although for review purposes, it should be read from `0d24f6a7bc9fff5bd14bccc1b97304c73b9474ed` as a whole.  It should provide full support for mixed block static allocation in both classic and optimized modes.

One nice thing is as follows: we can ditch the old `Field_of_static_block` type entirely and just use a `Simple` paired with a `Debuginfo`.  In addition we can just have a single constructor for scannable blocks in `Static_const`.

The types have been carefully chosen with regard to the `Scannable` submodules so the encoding should be precise.  (Some of the block shape types introduced in #2533 have been reworked a little to make this all fit together nicely.) The block shapes being carried along with the lists of `Simple`s used to build a statically-allocated block provide all the necessary information required for compilation to Cmm.

Since it was useful for getting the types right I've also added support in classic mode for statically allocating all-float blocks (note these are blocks, e.g. from records, not arrays so the float array optimization isn't relevant here: they are always stored unboxed).  There are no approximations for these however.

In addition there is a commit which introduces a new type to describe elements of a mixed block's flat suffix.  I think we should avoid hanging onto Lambda types into Flambda 2: separate types for separate IRs enable nuances to be encoded more effectively, amongst other things.  Here there is a concrete example: the `Float_boxed` constructor doesn't seem to make sense at Cmm time in the existing Lambda type that is being used for flat suffix elements, since any boxing required would have been expanded on the way into Flambda 2.

I need to do another pass to make sure everything looks ok and tests pass, but I'm putting this up now to gather preliminary feedback.


middle_end/flambda2/to_cmm/to_cmm_static.ml
```diff
@@ -23,34 +23,59 @@ module SC = Static_const
...
+        C.cfloat (Numeric_types.Float_by_bit_pattern.to_float f)
+      | Naked_int32 i -> C.cint32 (Nativeint.of_int32 i)
+      (* XCR nroberts: Could somebody check the below cases? They feel funny to
+         me.
+
+         mshinwell: they seem to be correct, what did you think is wrong? *)
+      | Naked_int64 i -> C.cint (Int64.to_nativeint i)
+      | Naked_nativeint i -> (
+        match Targetint_32_64.repr i with
+        | Int32 i -> C.cint (Nativeint.of_int32 i)
```
Gbury, line 51: This case is surprising considering that most of the code elsewhere assumes that we compile for a `64` bit target (and also that the host is `64` bit, else some conversions could lose some information).

middle_end/flambda2/to_cmm/to_cmm_static.ml
```diff
@@ -206,18 +231,29 @@ let immutable_unboxed_float32_array env res updates ~symbol ~elts =
...
+  | Block_like s, Block (tag, mut, shape, fields) ->
+    (match mut with
+    | Immutable | Immutable_unique -> ()
+    | Mutable ->
+      (* CR mshinwell: actually we could probably permit updates on the flat
+         suffix even now *)
+      Misc.fatal_errorf
+        "Symbol %a: the GC does not currently support mutable fields in \
+         statically-allocated values"
+        Symbol.print s);
```
Gbury, line 243: Is that really true ? The GC already handles cases where dynamically allocated values are stored in the fields of a statically allocated value (these statically-allocated blocks are registered as roots so that the fields containing dynamically allocated values are correctly updated when the GC moves values).

So I'm wondering what prevents mutables fields in statically-allocated values ?

mshinwell: I believe this is still the case, but I don't know the details.  I remember that Damien was going to write, or did write most of, a patch for runtime4 several years ago to lift this restriction but I'm not sure what came of it.  It would require rewriting for runtime5 I would guess.

middle_end/flambda2/to_cmm/to_cmm_primitive.ml
```diff
@@ -106,15 +121,25 @@ let block_load ~dbg (kind : P.Block_access_kind.t) (mutability : Mutability.t)
   | Naked_floats _ -> C.unboxed_float_array_ref block index dbg
   | Mixed { field_kind = Flat_suffix field_kind; _ } -> (
     match field_kind with
-    | Imm -> C.get_field_computed Immediate mutability ~block ~index dbg
-    | Float_boxed | Float64 ->
+    | Value ->
+      (* The flat suffix cannot store scannable values, so this must be an
+         immediate *)
```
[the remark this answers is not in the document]
mshinwell: Oh, yes indeed - the type I wrote today had `Tagged_immediate` for this case, so this must be another one.  I will try to fix this.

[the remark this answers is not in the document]
mshinwell: This should now be fixed.

middle_end/flambda2/kinds/flambda_kind.ml
```diff
@@ -173,102 +163,173 @@ let is_naked_float t =
...
+  let equal
+      { value_prefix_size = value_prefix_size1;
+        flat_suffix = flat_suffix1;
+        field_kinds = _
+      }
+      { value_prefix_size = value_prefix_size2;
+        flat_suffix = flat_suffix2;
+        field_kinds = _
+      } =
+    (* [field_kinds] is uniquely determined by [flat_suffix]. *)
```
lthls, line 235: Could you move this comment to the type definition ? Maybe elaborating a bit more, saying that `field_kinds` is what gets accessed most during `Flambda` but the other two are necessary to compile to Cmm.

mshinwell: Done

middle_end/flambda2/to_cmm/to_cmm_primitive.ml
```diff
@@ -123,10 +123,7 @@ let block_load ~dbg (kind : P.Block_access_kind.t) (mutability : Mutability.t)
     match field_kind with
     | Tagged_immediate ->
       C.get_field_computed Immediate mutability ~block ~index dbg
-    | Naked_float ->
-      (* CR layouts v5.1: We should use the mutability here to generate better
-         code if the load is immutable. *)
-      C.unboxed_float_array_ref block index dbg
+    | Naked_float -> C.unboxed_immutable_float_array_ref block index dbg
```
ncik-roberts: Shouldn't this use `mutability` to detect whether the load is mutable or not? It's possible for a naked float field in a mixed block to be mutable.

middle_end/flambda2/to_cmm/to_cmm_static.ml
```diff
@@ -23,34 +23,59 @@ module SC = Static_const
...
+      | Naked_float32 f ->
+        C.cfloat32 (Numeric_types.Float32_by_bit_pattern.to_float f)
+      | Naked_float f ->
+        C.cfloat (Numeric_types.Float_by_bit_pattern.to_float f)
+      | Naked_int32 i -> C.cint32 (Nativeint.of_int32 i)
+      (* XCR nroberts: Could somebody check the below cases? They feel funny to
+         me.
+
+         mshinwell: they seem to be correct, what did you think is wrong? *)
+      | Naked_int64 i -> C.cint (Int64.to_nativeint i)
```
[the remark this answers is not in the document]
mshinwell: @ncik-roberts ping as to what your concern was here...

mshinwell:
@lthls is going to review this.

middle_end/flambda2/types/reify.ml
```diff
@@ -207,27 +211,45 @@ let reify ~allowed_if_free_vars_defined_in ~var_is_defined_at_toplevel
...
+            (* CR mshinwell: could recognise tagged immediates *)
+            try_canonical_simple ()
           | Some (tag, shape, size, field_types, alloc_mode) -> (
             assert (
               Targetint_31_63.equal size
                 (TG.Product.Int_indexed.width field_types));
-            (* CR mshinwell: Could recognise other things, e.g. tagged
-               immediates and float arrays, supported by [Static_part]. *)
+            let field_types = TG.Product.Int_indexed.components field_types in
+            let field_kinds =
```
lthls, line 222: I think you could build the combined list directly here, instead of passing two lists to `try_to_reify_fields` which will combine them anyway.

middle_end/flambda2/types/reify.ml
```diff
@@ -524,9 +546,10 @@ let reify ~allowed_if_free_vars_defined_in ~var_is_defined_at_toplevel
           let kind = Flambda_kind.With_subkind.kind element_kind in
           match kind with
           | Value -> (
+            let field_kinds = List.map (fun _ -> Flambda_kind.value) fields in
```
lthls, line 549: Same as above (you can build the list of tuples directly)

lthls (approved):
I'm fine with the overall approach and implementation. I've left a number of comments in places where I think the code could be easily improved.

ocaml/testsuite/tests/mixed-blocks/structural_constants.ml
```diff
@@ -0,0 +1,86 @@
+(* TEST
+   flambda2;
+   flags="-extension layouts_beta";
```
lthls, line 3: The new check correctly points out that this test doesn't actually run anything.

lthls:
I've found a few bugs remaining in this PR that will need to be addressed. One is a variant of #2763 (`To_cmm_static.static_field` needs to be updated so that all expressions have the correct size), and I also suspect a bug in the translation of constant mixed blocks, where the code in `Closure_conversion` fails to properly unbox the constants that have kind `Float_boxed`. Hopefully fixing both of those should be enough to make the testsuite go green.

middle_end/flambda2/to_cmm/to_cmm_static.ml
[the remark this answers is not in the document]
mshinwell: We're now using `C.const_static` here, and the code seems correct?

middle_end/flambda2/to_cmm/to_cmm_static.ml
mshinwell: This has been removed now

middle_end/flambda2/kinds/flambda_kind.ml
```diff
@@ -839,6 +839,8 @@ module With_subkind = struct
               match Tag.Scannable.create tag with
               | Some tag ->
                 let shape_and_fields : Block_shape.t * _ =
+                  (* CR mshinwell/vlaviron: In both of these cases it would be
+                     nice to propagate immediacy information. *)
```
lthls, line 843: Actually, I believe that `from_lambda_value_kind` takes care of that, the only remaining case where we lose immediacy is the flat suffix of mixed blocks.
(There are other places where we lose immediacy by computing a `kind` instead of a `kind_with_subkind`, but at least that is explicit.)

mshinwell: See discussion elsewhere

PR #2713, opened 2024-06-24
Fix line number directives in stdlib/otherlibs

ncik-roberts:
Augment the change in https://github.com/ocaml-flambda/ocaml-jst/commit/52e6e784acd6d58a7f25b07c1cd01cab3d4744bd to fix the line number known for the rest of the file.

* That commit originally added lines like `# 1 "stdlib.ml"` to the top of all stdlib files to stabilise the file names included in backtraces.
* But 1 is the wrong number to include here. It should be 2.

This helps merlin.

@stedolan described it better than I could:
> Line directives affect the location for following lines and are typically used in preprocessed / generated code to point tools to the actual source, e.g. if `orig.ml` contains:
> 
>     let foo () = 42
> 
> and `gen.ml` contains:
> 
>     # 1 "orig.ml"
>     let foo () = 42
> 
> then the tools know that the definition on line 2 of `gen.ml` should in fact point to line 1 of orig.ml.
> 
> In our unusual usage, we have a single file `stdlib.ml` containing:
> 
>     # 1 "stdlib.ml"
>     let foo () = 42
> 
> and the tools are correctly interpreting this to say that the definition on line 2 of `stdlib.ml` actually came from line 1 of `stdlib.ml`, which is not what we want.
> 
> So if these annotations are still needed to update filenames, they should say `# 2 "stdlib.ml"`, because that's where the code is. (The line directive refers to the line number of the following line)

I produced this PR with this command to be sure I'm just updating the files from the ocaml-jst commit:
```
$ git diff 7d08e0e1701596da64a5b739dbaf71140d68bb01 52e6e784acd6d58a7f25b07c1cd01cab3d4744bd --stat | awk '{ print $1 }' | xargs -n 1 printf "ocaml/%s\n" | xargs sed -i '1s/# 1 "\(.*\)"$/# 2 "\1"/'
```
I then manually applied the same change to the following files, which are new since that commit:
  * `ocaml/otherlibs/stdlib_upstream_compatible/{nativeint_u,int32_u,float_u.ml,int64_u,nativeint_u}.{ml,mli}`
  * `ocaml/otherlibs/systhreads4/thread.{ml,mli}`
  * `ocaml/stdlib/domain.{ml,mli}`


goldfirere (approved):

ncik-roberts:
Unlike the original commit, this PR involves some changes to test output. (Anything that prints a backtrace that involves a function from the stdlib.) This is expected, as the line numbers in the new files did change with the addition of the line number directive &mdash; it's just that the original PR incorrectly suppressed the changes.

The changes to test output are light: only 8 upstream tests need to be updated. I simply promoted the failing output. I don't expect difficulties with merging changes in the future.

ncik-roberts:
The only test failure is a known flake, so I'm going to merge.

PR #2717, opened 2024-06-24
Support for `mod` syntax - unclean

liam923:
The first commit is the diff of PR #2676. The intention of this PR is to improve testing and error messages without resolving other PR comments. The intention is to get something that works merged sooner to unblock some other people.


ocaml/typing/jkind_types.ml
```diff
@@ -289,43 +277,43 @@ module Sort = struct
     and var ppf v = fprintf ppf "{ contents = %a }" opt_t !v
   end
 
-  let for_function = Const Const.value
+  let for_function = Const Value
```
goldfirere: These RHSs should probably all be `value` instead of reconstructing `Const Value` for each

goldfirere (commented):
Reviewed everything but the tests

goldfirere (approved):
I've gotten through line 900 or so of jkinds.ml. But I'm clicking Approve, based on:
* I've reviewed the previous PR (which forms the first commit of this PR), and found no trouble in the implementation
* The improvements I was looking for were around design, but that's not a great reason to delay this needed PR
* But we thought adding in some tests were good, just to check for soundness trouble. It's looking very promising. And Approving now means that we can merge.

I will finish reviewing later today, but likely after @liam923 leaves for holiday.

goldfirere:
Finished review now, with a few small pushed changes. This is ready for merging once CI passes.

PR #2718, opened 2024-06-24
Fix toplevel printer for arrays of unboxed things

dvulakh:
The toplevel printer incorrectly attempts to treat arrays whose elements don't have layout `value` like normal arrays (using the `size` and `field` functions from `Obj` to find the length of the array and project its elements). This causes undefined behavior. For example, the following reliably segfaults our `utop`:
```ocaml
# let y = [| #0n |];;
```

The first commit exhibits the bug. It shows the output of the toplevel printer on a small selection of arrays of unboxed numbers. (There is no (reliable) segfault, but the arrays are rendered with incorrect numbers of elements).

The second commit prevents the printer from descending into arrays with a custom tag. Instead, the printer renders these as `<abstr array>`.


ncik-roberts (approved):
Looks good to me! Thanks for writing this.

PR #2723, opened 2024-06-25
zero_alloc: cleanup attributes in Lambda

gretay-js:
Remove `Ignore_assert_all` and `opt` from Lambda onwards, now that all the decisions about top-level `[@@@zero_alloc all]` can be made before Lambda (see also https://github.com/ocaml-flambda/flambda-backend/pull/2687). This is refactoring only that should not change behavior. Best reviewed with patdiff.

### DIFF
--- a/middle_end/flambda2/terms/zero_alloc_attribute.ml
+++ b/middle_end/flambda2/terms/zero_alloc_attribute.ml
@@ -10,22 +10,22 @@
 (*                                                                        *)
 (**************************************************************************)
 
-type t =
-  | Default_check
+type t = Lambda.zero_alloc_attribute =
+  | Default_zero_alloc
+  | Check of
+      { strict : bool;
+        loc : Location.t
+      }
   | Assume of
       { strict : bool;
         never_returns_normally : bool;
         never_raises : bool;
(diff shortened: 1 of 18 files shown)


middle_end/flambda2/terms/zero_alloc_attribute.ml
```diff
@@ -35,21 +35,13 @@ let print ppf t =
     Format.fprintf ppf "@[assert_zero_alloc%s@]"
       (if strict then "_strict" else "")
 
-let from_lambda : Lambda.zero_alloc_attribute -> Location.t -> t =
- fun a loc ->
+let from_lambda : Lambda.zero_alloc_attribute -> t =
```
ccasin, line 38: Can we now have `type t = Lambda.zero_alloc_attribute` and make this the identity function rather than building new copies?

ccasin (approved):
Looks good - just one small suggestion.

middle_end/flambda2/terms/zero_alloc_attribute.ml
gretay-js: ah, nice! done.

PR #2724, opened 2024-06-25
Fix insertion of poll instruction (CFG)

xclerc:
This pull request fixes the way poll instructions
are inserted in the CFG pipeline. The previous
version, given the safe/unsafe domain and the
implementation of the dataflow analysis, would
consider every single block unsafe and thus insert
a poll instruction on every single back edge.
The new version inserts a poll instruction only
if there is a path for the header of the loop to
the back edge such that all the blocks part of
that path are unsafe.


xclerc:
I think it is basically ready for review,
it needs a bit more testing - including
on the performance side since the
search for a path between the destination
and the source of a back edge may be
too naive to be efficient in practice.

backend/cfg/cfg_polling.ml
```diff
@@ -188,21 +149,54 @@ let potentially_recursive_tailcall :
...
+let exists_unsafe_path :
+    Cfg.t -> safe_map:bool Label.Tbl.t -> from:Label.t -> to_:Label.t -> bool =
+ fun cfg ~safe_map ~from ~to_ ->
+  let exception Found in
+  try
+    let open_ = ref (Label.Set.singleton from) in
+    let closed = ref Label.Set.empty in
+    while not (Label.Set.is_empty !open_) do
+      let label = Label.Set.choose !open_ in
+      if Label.equal label to_ then raise Found;
```
gretay-js, line 165: I think the current implementation is overly conservative, we should check that `label` is not safe before `raise Found`. 
For example: H->B and B->H where A is the loop header block, A it is not safe, B is safe. The result of `exists_unsafe_path ~from:H ~to_:B` will be true, I think. It's not taking into account safe instructions in `to_` block.

gretay-js (approved):
This is correct and better than the existing cfg_polling.

backend/cfg/cfg_polling.ml
xclerc: As discussed off-line, we will address the
issue in a follow-up pull request.

xclerc: (Should be addressed by https://github.com/ocaml-flambda/flambda-backend/pull/3113.)

PR #2725, opened 2024-06-25
Allow forceful crossing of portability and contention

liam923:
Adds the flag `-allow-illegal-crossing`, which allows nominative types to cross the portability and contention axes when explicitly annotated.

I'm not a big fan of the flag name and am open to suggestions.


ocaml/testsuite/tests/typing-layouts/allow-illegal-crossing.ml
```diff
@@ -0,0 +1,654 @@
...
+   flags = "-allow-illegal-crossing";
+   expect;
+ }
+*)
+
+(******************************************************************************)
+(* Test 1: types can cross the portability and contention axes when annotated *)
+
+type a : value mod nonportable
+type b : value mod portable = private a
```
goldfirere, line 12: I think this should be rejected. If we aren't remembering the kinds of private abbreviations (and we're not), then we should check that the kind annotation is a superkind of the kind of the RHS, and it's not here. Ditto for other examples.

Given that we've now discovered that private abbreviations are not nominative types (and therefore not so interesting in the context of this patch), I'm going to skip the other private-abbreviation tests.

ocaml/testsuite/tests/typing-layouts/allow-illegal-crossing.ml
```diff
@@ -0,0 +1,654 @@
...
+|}]
+
+type t : value mod portable = private string
+[%%expect {|
+type t = private string
+|}]
+
+type t : value mod portable uncontended = { a : int; b : int }
+[%%expect {|
+type t = { a : int; b : int; }
```
goldfirere, line 84: I was expecting this to output the kind that supports mode-crossing.

ocaml/testsuite/tests/typing-layouts/allow-illegal-crossing.ml
```diff
@@ -0,0 +1,654 @@
...
+type a = private Foo of int | Bar
+type b = a = private Foo of int | Bar
+|}]
+
+(********************************************)
+(* Test 2: illegal mode crossing propogates *)
+
+type a : value mod portable uncontended = private string
+type ('a : value mod portable uncontended) b
+type c = a b
```
goldfirere, line 179: I would expect this to be rejected, due to lack of nominative type.

ocaml/testsuite/tests/typing-layouts/allow-illegal-crossing.ml
```diff
@@ -0,0 +1,654 @@
...
+[%%expect {|
+type a = private string
+type ('a : value mod uncontended portable) b
+type c = a b
+|}]
+
+type t : value mod portable uncontended = private string
+let f : ('a : value mod portable uncontended). 'a -> 'a = fun x -> x
+let g (x : t) = f x
```
goldfirere, line 188: I would also expect rejection here.

ocaml/testsuite/tests/typing-layouts/allow-illegal-crossing.ml
```diff
@@ -0,0 +1,654 @@
...
+val x : (int, int) t = {a = 5; b = 5}
+|}]
+
+type t : value mod portable uncontended = Foo of string | Bar of int
+let x : _ as (_ : value mod portable uncontended) = Foo "hello world"
+[%%expect {|
+type t = Foo of string | Bar of int
+val x : t = Foo "hello world"
+|}]
+
```
goldfirere, line 215: I would want to see validation of the testing technique here. Something like

```
let x : _ as (_ : value mod portable uncontended) = "hi"
```

which would be rejected (and would show that actually this property is being correctly checked).

ocaml/testsuite/tests/typing-layouts/allow-illegal-crossing.ml
```diff
@@ -0,0 +1,654 @@
...
+Line 2, characters 0-37:
+2 | type b : value mod global = private a
+    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Error: The layout of type a is value, because
+         of the definition of a at line 1, characters 0-24.
+       But the layout of type a must be a sublayout of value, because
+         of the definition of b at line 2, characters 0-37.
+|}]
+
+type a : value mod once
```
goldfirere, line 232: `mod once` should be a no-op. Leave it out (here and elsewhere where `mod`ding by top).

ocaml/testsuite/tests/typing-layouts/allow-illegal-crossing.ml
```diff
@@ -0,0 +1,654 @@
...
+Line 4, characters 2-42:
+4 |   type t : value mod many = private string
+      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Error: The layout of type string is value, because
+         it is the primitive value type string.
+       But the layout of type string must be a sublayout of value, because
+         of the definition of t at line 4, characters 2-42.
+|}]
+
+type 'a t : value mod unique = private 'a
```
goldfirere, line 300: I'm surprised this is rejected. I would just expect to infer a `value mod unique` kind for `'a`. I think this deserves investigation.

ocaml/testsuite/tests/typing-layouts/allow-illegal-crossing.ml
```diff
@@ -0,0 +1,654 @@
...
+Error: The layout of type A.t is value, because
+         of the definition of t at line 2, characters 2-16.
+       But the layout of type A.t must be a sublayout of value, because
+         of the definition of t at line 9, characters 2-39.
+|}]
+
+(*********************************************************************************)
+(* Test 4: types cannot cross portability and contention axes when not annotated *)
+
+(* CR: should this be allowed? I think no, the user should have to specify at the
```
goldfirere, line 411: Do whatever is easiest. I agree that rejecting is probably better. But accepting this is very much in the spirit of this patch, and it may be hard to reject.

ocaml/testsuite/tests/typing-layouts/allow-illegal-crossing.ml
goldfirere, line 1: I was expecting to see some tests here actually covering mode-crossing. That is, if you declare that a type mode-crosses, does it in practice?

ocaml/typing/typedecl.ml
```diff
@@ -1131,6 +1139,12 @@ let check_coherence env loc dpath decl =
   | { type_kind = Type_abstract _;
       type_manifest = Some ty } ->
     let jkind' = Ctype.type_jkind_purely env ty in
+    let jkind' =
+      match decl.type_private, !Clflags.allow_illegal_crossing with
+      | Private, true ->
```
goldfirere, line 1144: As discussed, I think this case should be dropped.

ocaml/typing/typedecl.ml
```diff
@@ -1098,6 +1098,14 @@ let check_coherence env loc dpath decl =
   match decl with
     { type_kind = (Type_variant _ | Type_record _| Type_open);
       type_manifest = Some ty } ->
+      if !Clflags.allow_illegal_crossing then begin
```
goldfirere, line 1101: What's going on here? This doesn't seem to be related to this patch. I think I'm missing something.

liam923: This prevents illegal crossing for types that have both a manifest and a non-abstract kind. For example, without this, the following is accepted:
```
type a = { foo : string }
type b : value mod portable = a = { foo : string }
```
My understanding is that this should not be accepted.

ocaml/testsuite/tests/typing-layouts/allow-illegal-crossing.ml
liam923: I believe section `Test 2: illegal mode crossing propogates` covers this. Are there some examples of cases you'd like to see added?

ocaml/testsuite/tests/typing-layouts/allow-illegal-crossing.ml
liam923: Good point. This bug appears to have already been present in main, before my changes:
```ocaml
type 'a t : word = 'a
[%%expect {|
Line 1, characters 0-21:
1 | type 'a t : word = 'a
    ^^^^^^^^^^^^^^^^^^^^^
Error: The layout of type 'a is value, because
         of the definition of t at line 1, characters 0-21.
       But the layout of type 'a must be a sublayout of word, because
         of the definition of t at line 1, characters 0-21.
|}]
```
So I agree this needs to get fixed, but I believe it is outside the scope of this PR. For now, I'll remove this test.

liam923:
Add test demonstrating unsoundness

ocaml/testsuite/tests/typing-layouts/allow-illegal-crossing.ml
liam923: I added some tests demonstrating this bug to `jkinds.ml`

ocaml/typing/types.mli
```diff
@@ -532,6 +532,9 @@ type type_declaration =
     type_unboxed_default: bool;
     (* true if the unboxed-ness of this type was chosen by a compiler flag *)
     type_uid: Uid.t;
+    type_has_illegal_crossings: bool;
```
goldfirere, line 535: I see why this is here, but it should get removed once we remove illegal crossings. I think it's worth leaving a CR to do so.

Actually, I bet we'll be able to kill it once we can merge the improvements to printing. If not, let's circle back.

ocaml/typing/typeclass.ml
```diff
@@ -1618,6 +1618,7 @@ let temp_abbrev loc id arity uid =
        type_attributes = []; (* or keep attrs from the class decl? *)
        type_unboxed_default = false;
        type_uid = uid;
+       type_has_illegal_crossings = true;
```
goldfirere, line 1621: true?

ocaml/typing/printtyp.ml
```diff
@@ -1860,8 +1860,8 @@ let tree_of_type_decl id decl =
   in
   (* The algorithm for setting [lay] here is described as Case (C1) in
      Note [When to print jkind annotations] *)
```
goldfirere, line 1862: Ideally, that Note would be updated to match the new algorithm. But perhaps all of this is short-lived and so it doesn't matter much.

goldfirere (approved):

goldfirere (changes requested):
Didn't mean to approve without the mode-crossing tests (which I know you're working on)

ocaml/testsuite/tests/typing-layouts/jkinds.ml
```diff
@@ -1565,3 +1565,104 @@ Error: This expression has type <  > but an expression was expected of type
...
+[%%expect {|
+Line 1, characters 0-41:
+1 | type 'a t : value mod global = { x : 'a }
+    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Error: The layout of type t is value, because
+         it's a boxed record type.
+       But the layout of type t must be a sublayout of value, because
+         of the annotation on the declaration of the type t.
+|}]
+(* CR layouts v2.8: this should be accepted; 'a should be inferred to have kind
```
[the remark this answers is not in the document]
liam923: Good catch, I chose my modal axis poorly for this test. I should have chosen one of `portable`, `uncontended`, or `many` instead of `global` to test the behavior I had meant to. I've updated the test to  reflect this. Do you agree with it now?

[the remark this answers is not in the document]
goldfirere: No, still disagree. I agree that if `'a` crosses `many`, then it's safe for the record to, also. But I don't yet expect to be able to infer that. Maybe once we give an initial kind of `immutable_data with 'a`, which is `value mod many uncontended portable with 'a`, and then check `value mod many uncontended portable with 'a  <:  value mod many`, and then be able to reduce that to `linearity_bound_of('a) <: many`, we could get this. But that's a long way off.

ocaml/typing/printtyp.ml
liam923: Done

ocaml/typing/typeclass.ml
liam923: That was a typo, fixed now

ocaml/testsuite/tests/typing-layouts/jkinds.ml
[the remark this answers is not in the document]
liam923: ack, I've removed the CR from the `value mod many` test.

ocaml/testsuite/tests/typing-layouts/allow-illegal-crossing.ml
liam923: More tests have been added.

liam923:
> Add test demonstrating unsoundness

This has now been done

goldfirere (approved):

PR #2727, opened 2024-06-25
Update doc because local doesn't mean local-returning

riaqn:
This finishes the work in #2669

PR #2728, opened 2024-06-25
Fix library import error in `tests/small_numbers`

dkalinichenko-js:
`beta` -> `stdlib_beta`

Also revert #2669


riaqn:
It's actually #2727 that I'm trying to reverting. It shares the same branch name of #2669, which causes github to merge it automatically?

riaqn (approved):
For reverting #2727

PR #2730, opened 2024-06-25
Jkind Error Messages

liam923:
Improve errror messages for kind mismatches. The new behavior is:
1. If the layouts disagree, the error message will only be about layouts (regardless of whether modes agree), and modes are not included in the error message. Layouts are referred to by names like `float64`, even though this is a white lie, since that's actually the name of a kind.
2. If the layouts agree, then the message will be about kinds. The full jkind is printed out.

For the second case, I'd like to add some sort of hint at the end of the message about which modal axes disagree. But I don't have a great idea yet of what I think that should look like, so I chose to leave that for later work.

I think a reasonable alternative approach would be to completely get rid of case 1. Instead, always make the error message about kinds. When there is a layout disagreement, add a hint to the end of the message saying so. I'm not sure if this would be more or less confusing to users.

The diff is large, but most of it is tests. And most of those changes are the same change, repeated a bazillion times.

### DIFF
--- a/ocaml/otherlibs/dynlink/Makefile
+++ b/ocaml/otherlibs/dynlink/Makefile
@@ -123,6 +123,7 @@ COMPILERLIBS_SOURCES=\
   typing/primitive.ml \
   typing/zero_alloc.ml \
   typing/types.ml \
+  typing/oprint.ml \
   typing/jkind.ml \
   typing/typedtree.ml \
   typing/btype.ml \
--- a/ocaml/testsuite/tests/typing-layouts-or-null/basics.ml
+++ b/ocaml/testsuite/tests/typing-layouts-or-null/basics.ml
@@ -27,10 +27,10 @@ Line 1, characters 6-26:
 Error: This pattern matches values of type t_any_non_null
        but a pattern was expected which matches values of type
          ('a : '_representable_layout_1)
-       The layout of t_any_non_null is any, because
-         of the definition of t_any_non_null at line 2, characters 0-34.
-       But the layout of t_any_non_null must be representable, because
-         we must know concretely how to pass a function argument.
+       The layout of t_any_non_null is any
+         because of the definition of t_any_non_null at line 2, characters 0-34.
+       But the layout of t_any_non_null must be representable
+         because we must know concretely how to pass a function argument.
 |}]
 
 type t = { x : t_any_non_null }
@@ -40,10 +40,10 @@ Line 1, characters 11-29:
 1 | type t = { x : t_any_non_null }
                ^^^^^^^^^^^^^^^^^^
 Error: Record element types must have a representable layout.
-       The layout of t_any_non_null is any, because
-         of the definition of t_any_non_null at line 2, characters 0-34.
-       But the layout of t_any_non_null must be representable, because
-         it is the type of record field x.
+       The layout of t_any_non_null is any
+         because of the definition of t_any_non_null at line 2, characters 0-34.
+       But the layout of t_any_non_null must be representable
+         because it is the type of record field x.
 |}]
 
 module type S1 = sig
@@ -55,10 +55,10 @@ Line 2, characters 10-24:
 2 |   val x : t_any_non_null
               ^^^^^^^^^^^^^^
 Error: This type signature for x is not a value type.
-       The layout of type t_any_non_null is any, because
-         of the definition of t_any_non_null at line 2, characters 0-34.
-       But the layout of type t_any_non_null must be a sublayout of value, because
-         it's the type of something stored in a module structure.
+       The layout of type t_any_non_null is any
+         because of the definition of t_any_non_null at line 2, characters 0-34.
+       But the layout of type t_any_non_null must be a sublayout of value
+         because it's the type of something stored in a module structure.
 |}]
 
 module type S2 = sig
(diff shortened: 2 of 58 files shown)


goldfirere:
As you admit, this is kinda impossible to review. One way to make it much easier is to take the `because`-shuffling and break it out into its own commit, possibly before all of the others. Then I could review the interesting commits.

(I actually tried to do this, but ran into an unexpected error in `hidden_modules/test.ml` that I decided not to investigate further.)

Why was it necessary to move the `because` anyway? I'm worried its new placement will cause more lines to overflow, but I could be wrong. (I agree it flows better in its new placement.)

For hard-to-review PRs, it is very helpful to offer the reviewer the tools they need to effectively review, sometimes even providing little scripts to assist in reducing diffs. I don't think that's necessary here -- some git shuffling should be sufficient.

ocaml/typing/jkind.mli
```diff
@@ -417,6 +420,8 @@ val set_externality_upper_bound : t -> Externality.t -> t
 (*********************************)
 (* pretty printing *)
 
+val print_out_jkind : Format.formatter -> Outcometree.out_jkind -> unit
```
goldfirere, line 423: I'm surprised to see this here, and not in oprint, which is where such functions normally live. Is there a reason to buck convention here? Same for the other new exported function here.

ocaml/typing/jkind.mli
goldfirere, line 1: I thought this PR would end up removing the `Legacy` bits, but I don't see that here. Did I misunderstand something? Is that a followup somehow?

liam923: See https://github.com/ocaml-flambda/flambda-backend/pull/2676#discussion_r1667123424

liam923:
Re: shuffling `because`

I think shuffling the `because` makes the message substantially nicer when there is a long jkind. Ex:
```
Error: The kind of type t is value
         because it's a boxed record type.
       But the kind of type t must be a subkind of
         any mod global many uncontended portable
         because of the annotation on the declaration of the type t.
```
Compare this to:
```
Error: The kind of type t is value, because
         it's a boxed record type.
       But the kind of type t must be a subkind of
         any mod global many uncontended portable, because
         of the annotation on the declaration of the type t.
```

I think the difference between these will be even more pronounced when we start having even longer jkinds that take multiple lines. For example, I think that:
```
But the kind of type t must be a subkind of
  value mod global with 'a
                    protable with 'a
                    uncontended with 'a
                    unique
                    many
                    external
  because of the annotation on the declaration of the type t.
```
is a lot more readable than:
```
But the kind of type t must be a subkind of
  value mod global with 'a
                    protable with 'a
                    uncontended with 'a
                    unique
                    many
                    external, because
  of the annotation on the declaration of the type t.
```

liam923:
@goldfirere I've done some commit shuffling. Now, the two most recent commits encompass most of the test changes:
- d40270c: Make tests say kind when appropriate, and print actual layouts instead of legacy
- c2028fc: Update tests to shuffle because

ocaml/typing/jkind.mli
liam923: At the time I had managed to convince myself that this was necessary to avoid a dependency cycle. But I think that was wrong. I'll move it.

liam923: It's been moved

liam923: But the `Outcometree.out_jkind` is still generated in the `Jkind` module. I'm not sure if that goes against convention? I wanted to keep the logic about printing jkinds in the `Jkind` module.

goldfirere: I think probably generating the `out_jkind` here is OK. But I have a suspicion that there's a good reason for the module structure the way that it is. In any case, the processing done in the moved functions is a bit simple, and so I think it's OK for it to be elsewhere.

goldfirere (approved):

ocaml/otherlibs/dynlink/Makefile
```diff
@@ -123,6 +123,7 @@ COMPILERLIBS_SOURCES=\
   typing/primitive.ml \
   typing/zero_alloc.ml \
   typing/types.ml \
+  typing/oprint.ml \
```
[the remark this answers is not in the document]
liam923: Turns out I didn't need to add this - it builds fine without. But it doesn't seem to hurt to leave in.

ocaml/typing/jkind.ml
```diff
@@ -502,7 +511,10 @@ module Const = struct
   end
 
   module To_out_jkind_const = struct
```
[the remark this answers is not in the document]
liam923: Agreed

ocaml/testsuite/tests/typing-layouts-or-null/basics.ml
```diff
@@ -112,7 +113,8 @@ Error: Record element types must have a representable layout.
        The layout of t_value_or_null is value_or_null, because
```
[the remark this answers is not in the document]
liam923: This is a test whose output was modified in the last three commits that touched a bunch of tests. Based on how GitHub is showing your commend to me, I think you were looking at a diff that doesn't include those changes when you left this comment. Just to make sure we are on the same page, the new error message is:
```
Error: Record element types must have a representable layout.
       The kind of t_value_or_null is value_or_null
         because of the definition of t_value_or_null at line 3, characters 0-36.
       But the kind of t_value_or_null must be a subkind of value
         because it is the type of record field x,
         defaulted to kind value.
```
Excluding the first line, this talks about kinds instead of layout, which is an improvement in my book. But I agree the first line is still bad. I'll see if I can improve it.

The `scannable` idea is interesting, althought I think its name would be confusing if we add a `scannability` axis.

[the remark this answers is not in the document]
liam923: I think this might also be fine to leave for now - at least this case. If I understand correctly, this should be accepted in the future, as `value_or_null` should be considered representable.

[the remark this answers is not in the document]
goldfirere: Yes, thanks for clarifying. I agree this is fine.

PR #2732, opened 2024-06-25
Add (more) layout restrictions on unnamed values in the toplevel

ccasin:
This bans some uses of non-values in the toplevel, particularly `e;;` and `let _ = e;;` when `e`'s type does not have layout value.  Those cases currently hit an exception in the middle end, and have for a while - as unboxed types become more widely used we should give a nicer error, at least.

There are a few ways to arrange for this - I'm not sure I've chosen the cleanest, but eventually we'll eliminate this restriction and this code will go away.

Review: @ncik-roberts perhaps?


ccasin:
Note this PR is conservative: We could allow these things in the bytecode toplevel but not in the nativecode toplevel.  It doesn't seem worth the additional effort to thread that distinction to the right place, though, and in any event we expect these toplevels to behave similarly (there are tests for this).

ocaml/typing/typecore.ml
```diff
@@ -8365,14 +8365,23 @@ and type_let ?check ?check_strict ?(force_toplevel = false)
...
 
+  let var_for_binding (_, _, _, pat) =
+    (* The native code toplevel requires special handling of [let _ = ...],
+       see the special case in [Opttoploop.execute_phrase]. *)
+    match pat.ppat_desc with
+    | Ppat_any when force_toplevel ->
+      newvar (Jkind.Primitive.value ~why:Unnamed_toplevel_expression),
+      Jkind.Sort.value
+    | _ -> new_rep_var ~why:Let_binding ()
+  in
```
ncik-roberts, line 8376: Won't this not catch `let _ : _ = #4.0`? (I thought it might also not catch `let (_ | _) = #4.0`, but that appears to already be unproblematic... probably because `Ppat_constraint` vanishes during type-checking, but `Ppat_or` does not, and `execute_phrase` runs on the result of type-checking.)

ccasin: Good point.  I'd forgotten that even though we now parse the constraints nicely, we mangle them back to the legacy patterns before this point in `type_let`.

And while or-patterns aren't problematic there are a few other forms that are (e.g., `let M.(_) = #3.14`).  Rather than trying to guess which parsetree patterns will be compiled to `Tpat_any`, I've moved this check to typemod after the pattern is translated to typedtree, so we can just check.  Probably should have just done that in the first place.  Now the two cases (`Pstr_eval` and `Pstr_value`) are handled more uniformly.

ocaml/typing/typemod.ml
```diff
@@ -2782,34 +2783,60 @@ and type_structure ?(toplevel = None) funct_body anchor env sstr =
...
           Typecore.type_binding env rec_flag ~force_toplevel sdefs in
         let defs = match rec_flag with
           | Recursive -> Typecore.annotate_recursive_bindings env defs
           | Nonrecursive -> defs
         in
+        if force_toplevel then
+          (* See comment on [force_toplevel] *)
+          List.iter (fun vb ->
+            match vb.vb_pat.pat_desc with
+            | Tpat_any ->
```
goldfirere, line 2831: The goal is to find an expression whose let-binding binds no variables, right? I don't think this quite does it. What about

```
let #3L = Int64_u.add #1L #2L
```

? It's a bit silly, but I think that might tickle this case.

Also, what if the pattern does bind boxed variables, but the overall expression is still unboxed? For example

```
let #( s1, s2 ) = #( "hello", "world" )
```

? I don't know exactly how utop operates or what the requirement is here, but I'm worried that `Tpat_any` doesn't quite cover it.

ccasin: > The goal is to find an expression whose let-binding binds no variables, right?

No, the goal is to find any expression that will undergo the transformation done by `Opttoploop.execute_phrase` (see above comment - though if it's not clear I can edit!).  This is exactly things that become `Tstr_eval` or `Tstr_value` with a `Tany` pattern.

Testing quickly, your example does not hit this case in utop.  It does hit an (apparently unrelated) fatal error in the native code toplevel, which looks to be related to how `Stdlib_upstream_compatible` is linked - I will make a ticket for that and ask @dkalinichenko-js to take a look.

ccasin: (To be clearer: All other cases, including the unboxed tuple one, should already be rejected by the typechecker if they are problematic.  This bug is specifically about that transformation done by the native code toplevel.)

ncik-roberts (approved):

PR #2734, opened 2024-06-26
Update document because `local_` no longer mean local returning

riaqn:
Update document in accordance to #2669 . Revised according to #2727


goldfirere:
What’s the relationship between this PR and #2727? They seem to achieve the same result.

riaqn:
That one is now closed (because I messed up the branch name)

ncik-roberts (approved):

PR #2736, opened 2024-06-26
Rc_close_at_apply should only close the current region early

mshinwell:
The test case here exhibits a problem relating to the handling of `Rc_close_at_apply` in `Lambda_to_flambda`.  It was being misinterpreted as "close all regions up to those required by the return continuation" prior to the call -- whereas it should only affect the current region in this regard.  I've added some explanations in `lambda.mli` which are almost entirely due to @stedolan (thanks also to Stephen for other assistance with this bug).

I'm going to think about how we can adjust the test case so it actually reliably fails prior to this PR, and succeeds after.

This patch needs looking over very carefully, as this is a subtle area.


Ekdohibs (approved):

ocaml/testsuite/tests/typing-local/close_at_apply.ml
```diff
@@ -0,0 +1,44 @@
...
+  = "%addfloat"
+
+let locally_allocate =
+  let f = 1.0 in
+  fun [@inline never] () -> ignore (opaque_identity (f +. 1.0) : float)
+;;
+
+let foo list =
+  (* The Simplif local function optimization will transform [run] and
+     [wrapper] into continuations. *)
```
lthls, line 26: Please annotate such functions with `[@local]` if your test relies on the optimisation

PR #2737, opened 2024-06-26
Implement `-no-approx` and `-no-code` in flambda-backend `ocamlobjinfo`

lukemaurer:
(Edit: Thanks to @lthls with lukemaurer#2, this now has finer-grained behavior for `-no-approx` and `-no-code`.)

The old `-no-approx` and `-no-code` options are currently parsed but ignored. This is inconvenient, since the output for an flambda2 .cmx can be screenfuls long even for relatively small modules.

This PR implements `-no-approx` by just suppressing .cmx export information altogether (roughly equivalent to what the old one did).

Also implements `-no-code` by doing exactly the same thing as `-no-approx`. (Which does indeed mean there's no code.)

### DIFF
--- a/middle_end/flambda2/cmx/flambda_cmx_format.ml
+++ b/middle_end/flambda2/cmx/flambda_cmx_format.ml
@@ -210,27 +210,35 @@ let merge t1_opt t2_opt =
     in
     Some (t1 @ t2, nsections)
 
-let print0 ~sections ppf t =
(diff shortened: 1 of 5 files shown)


lthls (approved):
Good for me.
Note that printing code doesn't really work, as we're not actually loading the code so only the metadata gets printed. This can be fixed later.

PR #2739, opened 2024-06-27
Fix missing restore/save_runtime_state in caml_c_thread_register for runtime5

mshinwell:
The systhreads lock was being taken prior to allocation on the OCaml heap during `caml_c_thread_register`, but the members of `Caml_state` were not being updated to correspond to the new thread prior to such allocation, nor restored afterwards.  This could lead to segfaults.  This PR was soak tested using a repro case that previously segfaulted.

### DIFF
--- a/ocaml/otherlibs/systhreads/st_stubs.c
+++ b/ocaml/otherlibs/systhreads/st_stubs.c
@@ -848,6 +848,11 @@ CAMLexport int caml_c_thread_register(void)
   }
   /* Associate the thread descriptor with the thread */
   st_tls_set(caml_thread_key, (void *) th);
+
+  /* Prepare Caml_state for allocation etc., so its various members correspond
+     to those for the newly-registered thread */
+  restore_runtime_state(th);
+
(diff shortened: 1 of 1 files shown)


stedolan:
The fix looks good to me. It restores some logic that was present in 4: the call to `restore_runtime_state` is a clearer and more efficient version of this sequence from 4:
```
  /* Release the master lock */
  release_runtime_lock();
  /* Now we can re-enter the run-time system and heap-allocate the descriptor */
  caml_leave_blocking_section();
```
because `leave_blocking_section` has the side effect of restoring runtime state, as well as re-acquiring the lock.

I think it's worth comparing this PR to ocaml/ocaml#12834, in both directions:
  - Does this change need to be upstreamed after all, or is the issue fixed by 12834?
  - Does 12834 contain other issues that we should backport here?

mshinwell:
It looks like this bug is fixed by the upstream PR in effectively the same way.  Regarding other changes in that PR and the upstream trunk runtime more widely, we'll discuss separately.

stedolan (approved):

PR #1379, opened 2023-05-11
Flambda2_types: Stop exporting get_alias_exn

lthls:
This function doesn't take a `min_name_mode` parameter, so there's no way for its user to know whether it is safe to use the resulting simple.
Most uses of `get_alias_exn` were due to `Simplify_simple.simplify_simple`: the type returned from this function was always guaranteed to be an alias to a simple with the right name mode. I've updated the function to return the simple directly, which trivially removes a number of calls to `get_alias_exn`. This does introduce a few tuple allocations, but hopefully not that much. (I've considered alternatives, like returning only the simple and not the type, but not all users can easily reconstruct the type as it needs a kind).
There were a few other uses that did not fit this, including one in `simplify_obj_dup` which was wrong, one for unboxing decisions that I replaced by a dedicated function, and a few for simplifying binary primitives where I had to move a few simples around.

PR #2740, opened 2024-06-27
Revert "Do not demote direct calls outside stubs"

xclerc:
Reverts ocaml-flambda/flambda-backend#2663


ccasin (approved):
Per discussion elsewhere, this seems like the best course for now.

xclerc:
(#2663 is causing `Unreachable code because of dead value slots`
errors which I mistakenly attributed to #1379.)

PR #2742, opened 2024-06-27
Infer needed `zero_alloc` checks from signatures

ccasin:
In the "`zero_alloc` in signatures" work, we made it so that to expose a function as zero alloc in your signature, you must also put a `zero_alloc` attribute on it in the structure.  This guarantees the backend check will run on that function, as implied by the signature.  But also it's a real pain to write the attribute in two places, and we can do better: This PR lets the user skip writing the attribute on the implementation and infers that the check is needed from the signature.

There are three commits, the first two of which are refactoring.  I suggest the main reviewer reads commit by commit.
1) The first commit makes a change that was suggested for the original signatures work.  It changes the way we store `zero_alloc` attributes on applications to be the syntactic version of those attributes rather than their values in the semantic domain.  At the same time, it changes when we check whether we can infer `[@zero_alloc assume]` on an application - now we do this during translation to lambda rather than when typechecking the application.  This is helpful in later commits - we are delaying that inference step until it is safe to default `zero_alloc` variables.
2) The second commit moves the logic for checking whether one `zero_alloc` attribute is at least as strong as another out of `Include_core` and into its own file.  It makes sense to have this together with the implementation of `zero_alloc` variables (which comes in the next commit), and I've made it its own commit so that I'm not both moving and changing that code in one commit.
3) The third commit implements inference and adds tests. If a function definition has no `zero_alloc` attribute on it we make a variable representing whether or not it will need to be checked.  That variable can be filled in when we compare the module to its signature.  These variables are somewhat simpler than jkind or type variables because we never need to link two of them.

I suggest @ncik-roberts reviews, as he did for the signatures implementation.  @gretay-js should take a look at the tests, at least.

This PR does not yet make it so that zero_alloc information is carried across aliases (`let f = g`) - there is enough here that I think that should be its own PR.

One minor design note:

### Locations in errors

Today `zero_alloc` errors look like this:

```
File "/home/ccasinghino/tmp/foo.ml", line 1, characters 5-15:
1 | let[@zero_alloc] f x = (x,x)
         ^^^^^^^^^^
Error: Annotation check for zero_alloc failed on function Foo.f (camlFoo__f_0_1_code)

File "/home/ccasinghino/tmp/foo.ml", line 1, characters 23-28:
1 | let[@zero_alloc] f x = (x,x)
                           ^^^^^
Error: allocation of 24 bytes
```

The error has two parts: The first identifies the location of the attribute that said to check something.  The second shows the offending allocation.

After this PR, the attribute that caused the check may be far removed from the implementation, off in a signature somewhere.  Making that show up first in the error seems bad for editor integration - people probably want to navigate to the offending function.  So if the check is an inferred one, I've made the first error point to the function that was checked instead.

I'm not sure that's exactly the right place to point.  Also it would probably be convenient to know what attribute caused this check, somehow, even if that's not going to be the place we most commonly want to navigate.  That information is easy to make available, but I think the right thing is to pass it to the backend and have the backend include it later in the error, and propose putting that off for a future PR.

### DIFF
--- a/ocaml/testsuite/tests/typing-zero-alloc/cmi_test.ml
+++ b/ocaml/testsuite/tests/typing-zero-alloc/cmi_test.ml
@@ -1,0 +1,77 @@
+(* TEST
+   readonly_files = "cmi_test_lib.ml";
+   setup-ocamlc.byte-build-env;
+   module = "cmi_test_lib.ml";
+   ocamlc.byte;
+   flags += "-I ocamlc.byte";
+   expect;
+*)
+
+(* Here we show the signatures of [cmi_test_a] and the modules within it do not
+   have zero_alloc variables - we can't add further zero_alloc constraints. *)
+module M1 : sig
+  val[@zero_alloc] f_unconstrained_variable : int -> int
+end = Cmi_test_lib
+[%%expect{|
+Line 3, characters 6-18:
+3 | end = Cmi_test_lib
+          ^^^^^^^^^^^^
+Error: Signature mismatch:
+       Modules do not match:
+         sig
+           val f_unconstrained_variable : int -> int
+           module M_constrained_variable =
+             Cmi_test_lib.M_constrained_variable
+           module M_no_variable = Cmi_test_lib.M_no_variable
+         end
+       is not included in
+         sig val f_unconstrained_variable : int -> int [@@zero_alloc] end
+       Values do not match:
+         val f_unconstrained_variable : int -> int
+       is not included in
+         val f_unconstrained_variable : int -> int [@@zero_alloc]
+       The former provides a weaker "zero_alloc" guarantee than the latter.
+       Hint: Add a "zero_alloc" attribute to the implementation.
+       File "cmi_test_lib.ml", line 4, characters 4-28: Actual declaration
+|}]
+
+module M2 : sig
+  val[@zero_alloc strict] f : int -> int
+end = Cmi_test_lib.M_constrained_variable
+[%%expect{|
+Line 3, characters 6-41:
+3 | end = Cmi_test_lib.M_constrained_variable
+          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Error: Signature mismatch:
+       Modules do not match:
+         sig val f : int -> int [@@zero_alloc] end
+       is not included in
+         sig val f : int -> int [@@zero_alloc strict] end
+       Values do not match:
+         val f : int -> int [@@zero_alloc]
+       is not included in
+         val f : int -> int [@@zero_alloc strict]
+       The former provides a weaker "zero_alloc" guarantee than the latter.
+       File "cmi_test_lib.ml", line 7, characters 6-7: Actual declaration
+|}]
+
+module M3 : sig
+  val[@zero_alloc] f : int -> int
+end = Cmi_test_lib.M_no_variable
+[%%expect{|
+Line 3, characters 6-32:
+3 | end = Cmi_test_lib.M_no_variable
+          ^^^^^^^^^^^^^^^^^^^^^^^^^^
+Error: Signature mismatch:
+       Modules do not match:
+         sig val f : int -> int end
+       is not included in
+         sig val f : int -> int [@@zero_alloc] end
+       Values do not match:
+         val f : int -> int
+       is not included in
+         val f : int -> int [@@zero_alloc]
+       The former provides a weaker "zero_alloc" guarantee than the latter.
+       Hint: Add a "zero_alloc" attribute to the implementation.
+       File "cmi_test_lib.ml", line 13, characters 2-20: Actual declaration
+|}]
--- a/ocaml/testsuite/tests/typing-zero-alloc/cmi_test.ml
+++ b/ocaml/testsuite/tests/typing-zero-alloc/cmi_test.ml
@@ -3,12 +3,13 @@
    setup-ocamlc.byte-build-env;
    module = "cmi_test_lib.ml";
    ocamlc.byte;
-   flags += "-I ocamlc.byte";
+   flags += "-I ocamlc.byte ocamlc.byte/cmi_test_lib.cmo";
    expect;
 *)
 
 (* Here we show the signatures of [cmi_test_a] and the modules within it do not
-   have zero_alloc variables - we can't add further zero_alloc constraints. *)
+   have zero_alloc variables - we can't add further zero_alloc constraints, but
+   we do get the ones already present in the cmi. *)
 module M1 : sig
   val[@zero_alloc] f_unconstrained_variable : int -> int
 end = Cmi_test_lib
@@ -56,6 +57,13 @@ Error: Signature mismatch:
 |}]
 
 module M3 : sig
+  val[@zero_alloc] f : int -> int
+end = Cmi_test_lib.M_constrained_variable
+        [%%expect{|
+module M3 : sig val f : int -> int [@@zero_alloc] end
+|}]
+
+module M4 : sig
   val[@zero_alloc] f : int -> int
 end = Cmi_test_lib.M_no_variable
 [%%expect{|
--- a/ocaml/testsuite/tests/typing-zero-alloc/signatures.ml
+++ b/ocaml/testsuite/tests/typing-zero-alloc/signatures.ml
@@ -62,6 +62,12 @@ module type S_good_inc_base = sig
   val[@zero_alloc] f : 'a -> 'a
 end
 
+module M_absent : S_good_inc_base = struct
+  (* This one works by inferring the annotation on [f], tested further
+     elsewhere. *)
+  let f x = x
+end
+
 module M_base : S_good_inc_base = struct
   let[@zero_alloc] f x = x
 end
@@ -88,6 +94,7 @@ end
 
 [%%expect{|
 module type S_good_inc_base = sig val f : 'a -> 'a [@@zero_alloc] end
+module M_absent : S_good_inc_base
 module M_base : S_good_inc_base
 module M_assume : S_good_inc_base
 module M_assume_nrn : S_good_inc_base
@@ -98,6 +105,12 @@ module M_assume_strict_nrn : S_good_inc_base
 
 module type S_good_inc_opt = sig
   val[@zero_alloc opt] f : 'a -> 'a
+end
+
+module M_absent : S_good_inc_opt = struct
+  (* This one works by inferring the annotation on [f], tested further
+     elsewhere. *)
+  let f x = x
 end
 
 module M_base : S_good_inc_opt = struct
@@ -134,6 +147,7 @@ end
 
 [%%expect{|
 module type S_good_inc_opt = sig val f : 'a -> 'a [@@zero_alloc opt] end
+module M_absent : S_good_inc_opt
 module M_base : S_good_inc_opt
 module M_opt : S_good_inc_opt
 module M_assume : S_good_inc_opt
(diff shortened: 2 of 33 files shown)


tests/backend/zero_alloc_checker/test_inference.ml
```diff
@@ -0,0 +1,73 @@
...
+   [val[@zero_alloc] f_arity_two : int -> int -> int] we'll get a type error.
+   Here we show that if we give it a signature that passes the typechecker (the
+   mli has [val[@zero_alloc (arity 1)] f_arity_two : int -> int -> int]), it
+   correctly gets checked and gives a zero_alloc backend error. *)
+let f_arity_two x = fun y -> x + y
+
+
+(* Shadowing: the mli marks [f_shadow] zero_alloc.  That check should only apply
+   to the last such function in the file, so we get no error even though there
+   is an earlier function with the same name and type that allocates. *)
```
ncik-roberts, line 32: Can you please write a test in the other direction, too? (You get an error if the shadowing function allocates, and the mli is marked zero-alloc?)

tests/backend/zero_alloc_checker/test_inference.ml
```diff
@@ -0,0 +1,73 @@
...
+  if x = 42 then () else
+    let s = Printf.sprintf "%d\n" x in
+    raise (E s)
+
+(* These are tagged [@zero_alloc opt] in the mli - the latter should fail just
+   when the flag to check opt things is passed. *)
+let f_opt_success x =
+  if x = 42 then () else
+    let s = Printf.sprintf "%d\n" x in
+    raise (E s)
```
ncik-roberts, line 62: Is it easy to write a test of a function that is marked as `[@zero_alloc opt]` in the mli, and that _would_ fail if it was marked as `[@zero_alloc]`? (I.e. a test where the higher level of optimization removes allocation.)

ncik-roberts (commented):
Just a review on tests so far.

tests/backend/zero_alloc_checker/test_inference.ml
ccasin: I think no: For users we describe `opt` things as being checked when compiling with optimization, but actually it's just about whether you pass a flag saying the opt things should be checked (which people tell the build system to do when building with those higher optimization levels).

ccasin:
I've reworked this slightly to offer nicer inference behavior, and rebased/squashed to give it a nice history.  Nothing about the original PR description needs changing.

One interesting note, pointed out by Leo, is that the rules for zero_alloc variables are similar to the rules for modalities on the comonadic (prescriptive) axes.  In typemod I've slightly generalized the infrastructure to support those that was just added in #2685 and renamed the functions to, hopefully, explain their purpose more clearly, make clear that they are doing something similar to `remove_mode_and_kind_variables`, and be clear they deal with both modalities and zero_alloc variables.

ocaml/testsuite/tests/typing-zero-alloc/signatures.ml
```diff
@@ -1002,3 +936,257 @@ module type S_no_assume = sig val f : int -> int * int [@@zero_alloc] end
...
+(* When you take the module type of a module, you get the current state of its
+   variables.  The computed signature itself has no variables, and the original
+   module keeps its variables (but changing them won't change the signature you
+   got the first time, just the signature you'll get if you do [module type of]
+   again). *)
+module M_for_mto = struct
+  let f x = x+1
+end
+
+(* The current state of the variales is "no check" *)
```
ncik-roberts, line 1140:
```suggestion
(* The current state of the variables is "no check" *)
```

tests/backend/zero_alloc_checker/test_inference.mli
```diff
@@ -0,0 +1,25 @@
...
+val[@zero_alloc] f_basic_fail : 'a -> 'a * 'a
+
+val[@zero_alloc strict] f_strict_success : int -> int
+val[@zero_alloc strict] f_strict_fail : int -> unit
+
+val[@zero_alloc opt] f_opt_success : int -> unit
+val[@zero_alloc opt] f_opt_fail : int -> int * int
+
+val[@zero_alloc strict opt] f_strict_opt_success : int -> int
+val[@zero_alloc strict opt] f_strict_opt_fail : int -> unit
```
ncik-roberts, line 25: I think we should add a test demonstrating the "spooky action at a distance" of a module inclusion check causing the zero-alloc check to flag an issue with the original module definition. (You have type-checking tests for something like this.)

e.g.
```ocaml
module type S = functor () -> sig val[@zero_alloc] f : 'a -> 'a option end

module F () = struct
  let f x = Some x
end

module _ : S = F
```

In particular, the error message does not identify why the zero-alloc check is running at all, and I think it would be good to have a test demonstrating the iffy error message (and even better to someday make it better).

ocaml/testsuite/tests/typing-zero-alloc/cmi_test.ml
```diff
@@ -0,0 +1,77 @@
...
+         sig val f : int -> int [@@zero_alloc] end
+       is not included in
+         sig val f : int -> int [@@zero_alloc strict] end
+       Values do not match:
+         val f : int -> int [@@zero_alloc]
+       is not included in
+         val f : int -> int [@@zero_alloc strict]
+       The former provides a weaker "zero_alloc" guarantee than the latter.
+       File "cmi_test_lib.ml", line 7, characters 6-7: Actual declaration
+|}]
```
ncik-roberts, line 56: Consider adding a test showing that `M_constrained_variable` is contained in `val[@zero_alloc] f : int -> int`.

ocaml/typing/typedtree.mli
```diff
@@ -296,10 +296,10 @@ and expression_desc =
...
                         ])
 
-            The [Zero_alloc_utils.Assume_info.t] records the optional
-            [@zero_alloc assume] attribute that may appear on applications.  If
-            that attribute is absent, it is [Assume_info.none].
-          *)
+            The [zero_alloc_attribute] records the optional [@zero_alloc assume]
+            attribute that may appear on applications.  Invariant: it is either
+            [Default_zero_alloc] (if the user wrote no attribute) or [Assume]
+            (if they did). *)
```
ncik-roberts, line 302: Why not mint a new type and remove the failure cases? (I'm currently confused how `Ignore_assert_all` is ruled out by type-checking, and I think minting a new type here makes that easier to see/understand.)

ncik-roberts (commented):
some review

ocaml/typing/zero_alloc.mli
```diff
@@ -0,0 +1,50 @@
...
+   zero-alloc-ness, and with what configuration (strict, opt, etc). It can be a
+   variable which will be filled in when the module the function is in is
+   compared against its signature, allowing to infer zero-alloc checks. *)
+type t
+
+(* [default] corresponds to [Default_zero_alloc], meaning no check will be
+   done. *)
+val default : t
+
+val create : const -> t
```
ncik-roberts, line 26: Suggestion: `create_const`.

ocaml/typing/typedtree.ml
```diff
@@ -1068,32 +1068,34 @@ let let_bound_idents_with_modes_sorts_and_checks bindings =
...
-             if arity > 0 then
-               Check { strict = false;
-                       arity;
-                       loc = Location.none;
-                       opt = false }
+             if !Clflags.zero_alloc_check_assert_all && arity > 0 then
+               Zero_alloc.create (Check { strict = false;
+                                          arity;
+                                          loc = Location.none;
+                                          opt = false })
```
ncik-roberts, line 1084: If you rename `Zero_alloc.create` to `Zero_alloc.create_const` as I suggested elsewhere, you'll be prompted to change the indentation of the labels of `Check`. Let's just move the constructor to a new line to reduce this kind of sensitivity.

ocaml/typing/typedtree.ml
```diff
@@ -1068,32 +1068,34 @@ let let_bound_idents_with_modes_sorts_and_checks bindings =
...
              else
-               Default_zero_alloc
-           | Default_zero_alloc -> Default_zero_alloc
+               fn.zero_alloc
+           | Ignore_assert_all | Check _ | Assume _ -> fn.zero_alloc
          in
          Ident.Map.add id zero_alloc checks
+         (* CR ccasinghino: we could copy the zero-allocness if the vb_expr
+            is an ident. *)
```
ncik-roberts, line 1091: Flagging this: not sure if you intended on addressing this before merging.

ncik-roberts (commented):
Thanks for taking the time to add zero_alloc variables. The behavior seems much more normal to me know. I'll take a quick look again after you respond to my (mostly) nits.

ocaml/typing/includecore.ml
```diff
@@ -167,7 +98,9 @@ let value_descriptions ~loc env name
     loc
     vd1.val_attributes vd2.val_attributes
     name;
-  zero_alloc vd1.val_zero_alloc vd2.val_zero_alloc;
+  begin try Zero_alloc.sub_exn vd1.val_zero_alloc vd2.val_zero_alloc with
```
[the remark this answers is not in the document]
ccasin: I think this is fairly idiomatic for the compiler (e.g., this isn't the only try/with in this function). `sub_exn` is a series of checks any of which may fail independently, and I wanted to avoid writing a bunch of gross nested matches.  I could define a letop for bind in the result monad, or something, but in the typechecker the usual style is just to raise.

(I don't feel particularly strongly though - feel free to push back more)

ocaml/typing/typedtree.ml
ccasin: Sorry - this was a note to myself for the next PR.  I've made the comment clearer.

ccasin:
Thanks for the review.  I think I've addressed all the minor stuff and will play around with the types in typedtree sometime in the next few days.

ocaml/typing/includecore.ml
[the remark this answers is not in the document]
ncik-roberts: I thought all error creation appeared in tail position, except for one `if`-`then` needing to grow an `else`. If that’s true, let’s change it to return an error.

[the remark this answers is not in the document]
ccasin: Probably I'm missing some nice way to refactor this, but it's [these](https://github.com/ccasin/flambda-backend/blob/ccd41d34bd07feff3947fbaf2f96af0756789aac/ocaml/typing/zero_alloc.ml#L112) [two](https://github.com/ccasin/flambda-backend/blob/ccd41d34bd07feff3947fbaf2f96af0756789aac/ocaml/typing/zero_alloc.ml#L118) raises that seemed annoying.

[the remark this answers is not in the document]
ncik-roberts: I missed the second one. Fine, then we should move the try-with into the module.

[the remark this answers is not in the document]
ccasin: Good idea, done.

ocaml/typing/typedtree.mli
ccasin: I've taken a stab at this. It didn't result in any major simplfications, but is probably a bit clearer - let me know what you think.

ccasin:
> Thanks for the changes. I think the type refactoring you did in the latest commit is nice and helps make some non-obvious (or comment-enforced) invariants much clearer. I haven't checked whether additional type refactors would be nice in the same way, but you can make the call.

Thanks.  I did look for additional opportunities to make this kind of change, but didn't find them, because the other places where zero_alloc stuff appears in types or the typedtree need the option to be variables.

gretay-js (approved):

PR #2743, opened 2024-06-28
eliminate Ekdohibs review obligation

ccasin:
Enough of us have some understanding of chamelon now that I think it no longer makes sense to gate every small change to it behind @Ekdohibs review.  Of course, it will still make sense to get expert review for big changes, but we can do that manually as we do for other parts of the repo.

@Ekdohibs and @mshinwell should confirm this seems reasonable before merging.


Ekdohibs (approved):
All of my recent reviews on chamelon were simply approving the changes since they were already good, so I'm good with this.

mshinwell (approved):

PR #2744, opened 2024-06-28
Restore "Flambda2_types: Stop exporting get_alias_exn" (#1379)"

lthls:


xclerc:
(As noted on #2740, #1379 was reverted because
`Unreachable code because of dead value slots`
errors were mistakenly attributed to it.)

xclerc (approved):

PR #2745, opened 2024-06-28
Demote calls to non-simplified code only

lthls:
This is an alternative to #2651 and #2663.
The idea stays similar: we want to stop demoting direct calls in cases where we already know the right function to call but have lost the approximation on the closure for some reason (typically a missing cmx).
But we still need to be careful with code that has never been simplified: this code could be slow, so we don't want to call it if there is an alternative (i.e. going through an indirect call), and keeping it alive would require us to compute things like free names and slot offset constraints that we normally only compute when simplifying. The bugs that caused us to revert #2663 shows that we currently don't compute enough information in Closure_conversion: we do have the constraints on slot offsets generated by the sets of closures in non-simplified code, but we don't see the slot projections so we might incorrectly treat some value slots as unused.
While fixing this issue is possible (by storing more information in the slot_offsets structure), this PR goes in the other direction, making sure that we don't call non-simplified code.

### DIFF
--- a/middle_end/flambda2/simplify/env/downwards_acc.ml
+++ b/middle_end/flambda2/simplify/env/downwards_acc.ml
@@ -29,6 +29,7 @@ type t =
     demoted_exn_handlers : Continuation.Set.t;
     code_ids_to_remember : Code_id.Set.t;
     code_ids_to_never_delete : Code_id.Set.t;
+    code_ids_never_simplified : Code_id.Set.t;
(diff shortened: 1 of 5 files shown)


lthls:
To be discussed with @mshinwell when he is back.

chambart:
I'm not sure I can imagine every implications, but that seems to be a sensible way to fix the bug.

chambart (approved):
That still seems sensible

PR #2749, opened 2024-07-01
Implement `'a or_null`

dkalinichenko-js:
Do not update the jkind in `update_decl_jkind` if the manifest is available; always copy the manifest's jkind in `check_coherence`. This allows us to give types like `'a or_null` non-standard jkinds for their kinds (i.e. `value_or_null` for `Type_variant`).

Pre-define `or_null` when `Layouts` has level `Alpha`. Add tests for `or_null`.

`or_null` will be added to `stdlib_alpha` in the next PR.

### DIFF
--- a/ocaml/testsuite/tests/typing-layouts-or-null/test_or_null.ml
+++ b/ocaml/testsuite/tests/typing-layouts-or-null/test_or_null.ml
@@ -1,0 +1,130 @@
+(* TEST
+ flags = "-extension layouts_alpha";
+ expect;
+*)
+
+type ('a : value) t : value_or_null = 'a or_null =
+  | Null
+  | This of 'a
+
+[%%expect{|
+type 'a t = 'a or_null = Null | This of 'a
+|}]
+
+let to_option (x : 'a or_null) =
+  match x with
+  | Null -> None
+  | This x -> Some x
+
+[%%expect{|
+val to_option : 'a or_null -> 'a option = <fun>
+|}]
+
+let of_option (x : 'a option) =
+  match x with
+  | None -> Null
+  | Some x -> This x
+
+[%%expect{|
+val of_option : 'a option -> 'a t = <fun>
+|}]
+
+let pi = This 3.14
+
+[%%expect{|
+val pi : float t = This 3.14
+|}]
+
+let pi' =
+  let value = This 3.14 in
+  value
+
+[%%expect{|
+val pi' : float t = This 3.14
+|}]
+
+type myrec = { x : int
+         ; y : int or_null
+         }
+
+[%%expect{|
+type myrec = { x : int; y : int or_null; }
+|}]
+
+let fst { x; y } = x
+
+[%%expect{|
+val fst : myrec -> int = <fun>
+|}]
+
+let snd { x; y } = y
+
+[%%expect{|
+val snd : myrec -> int or_null = <fun>
+|}]
+
+let snd' (a : myrec) = a.y
+
+[%%expect{|
+val snd' : myrec -> int or_null = <fun>
+|}]
+
+let mk n = { x = n; y = This n }
+
+[%%expect{|
+val mk : int -> myrec = <fun>
+|}]
+
+let test =
+  let a = mk 4 in
+  let a' = { a with y = Null } in
+  a'.y
+
+[%%expect{|
+val test : int or_null = Null
+|}]
+
+let mytup = (4, This 5)
+
+[%%expect{|
+val mytup : int * int t = (4, This 5)
+|}]
+
+type mytup' = int * int t
+
+[%%expect{|
+type mytup' = int * int t
+|}]
+
+type nested = int or_null or_null
+
+[%%expect{|
+Line 1, characters 14-25:
+1 | type nested = int or_null or_null
+                  ^^^^^^^^^^^
+Error: This type int or_null should be an instance of type ('a : value)
+       The layout of int or_null is value_or_null, because
+         it is the primitive value_or_null type or_null.
+       But the layout of int or_null must be a sublayout of value, because
+         the type argument of or_null has layout value.
+|}]
+
+let should_fail = This (This 5)
+
+[%%expect{|
+Line 1, characters 23-31:
+1 | let should_fail = This (This 5)
+                           ^^^^^^^^
+Error: This expression has type 'a t = 'a or_null
+       but an expression was expected of type ('b : value)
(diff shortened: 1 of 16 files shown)


ocaml/testsuite/tests/typing-layouts-or-null/test_or_null.ml
```diff
@@ -0,0 +1,201 @@
...
+1 | type nested = int or_null or_null
+                  ^^^^^^^^^^^
+Error: This type int or_null should be an instance of type ('a : value)
+       The layout of int or_null is value_or_null, because
+         it is the primitive value_or_null type or_null.
+       But the layout of int or_null must be a sublayout of value, because
+         the type argument of or_null has layout value.
+|}]
+
+let should_fail = This (This 5)
```
goldfirere, line 110: Also worth testing: `This Null`.

ocaml/testsuite/tests/typing-layouts-or-null/test_or_null.ml
```diff
@@ -0,0 +1,201 @@
...
+       The layout of 'a t is value_or_null, because
+         it is the primitive value_or_null type or_null.
+       But the layout of 'a t must be a sublayout of value, because
+         of the definition of t at lines 1-3, characters 0-14.
+|}]
+
+let mk' n = `Foo (This n)
+
+[%%expect{|
+val mk' : 'a -> [> `Foo of 'a t ] = <fun>
```
goldfirere, line 127: Again, I would expect an explicit kind on `'a`.

ocaml/testsuite/tests/typing-layouts-or-null/test_or_null.ml
```diff
@@ -0,0 +1,201 @@
...
+external this : 'a -> 'a or_null = "%identity"
+|}]
+
+external unsafe_get : 'a or_null -> 'a = "%identity"
+
+[%%expect{|
+external unsafe_get : 'a or_null -> 'a = "%identity"
+|}]
+
+let should_fail = [| Null; This 5 |]
```
goldfirere, line 175: Probably should test this in a list, too.

ocaml/testsuite/tests/typing-layouts-or-null/test_or_null.ml
```diff
@@ -0,0 +1,201 @@
...
+Line 1, characters 19-32:
+1 | type should_fail = float or_null array
+                       ^^^^^^^^^^^^^
+Error: This type float or_null should be an instance of type
+         ('a : any_non_null)
+       The layout of float or_null is value_or_null, because
+         it is the primitive value_or_null type or_null.
+       But the layout of float or_null must be a sublayout of any_non_null, because
+         it's the type argument to the array type.
+|}]
```
goldfirere, line 201: Maybe add a test with an `or_null` as a instance variable or a method in an object? I don't really care what the behavior is, but I wouldn't want a panic, say.

ocaml/typing/predef.ml
```diff
@@ -460,6 +472,19 @@ let add_small_number_extension_types add_type env =
        ~jkind:(Jkind.Primitive.float32 ~why:(Primitive ident_unboxed_float32))
        ~jkind_annotation:Jkind.Const.Primitive.float32
 
+let add_or_null add_type env =
+  let add_type1 = mk_add_type1 add_type in
+  env
+  |> add_type1 ident_or_null
+  ~variance:Variance.covariant
+  ~separability:Separability.Sep
```
goldfirere, line 480: But this is a lie: this type is *not* separable. We've been thinking this is OK because we prevent `or_null`s from going into arrays. And I can't think of a problem here. But there should be a loud comment (and perhaps some discussion) about this.

Maybe, really, in our new world, we have the property that a type is separable if the type system thinks it is separable according to the old analysis _and_ the type is `non_null`. That's probably right.

ocaml/typing/typedecl.ml
```diff
@@ -1610,6 +1607,17 @@ let update_decl_jkind env dpath decl =
       type_jkind
   in
 
+  (* Do not override the jkind if the manifest jkind if available. This is
```
goldfirere, line 1610:
```suggestion
  (* Do not override the jkind if the manifest jkind is available. This is
```

ocaml/typing/typedecl.ml
```diff
@@ -1610,6 +1607,17 @@ let update_decl_jkind env dpath decl =
...
+     necessary to support types like ['a or_null], which have a non-standard
+     jkind. [check_coherence] will update the jkind to match the manifest. *)
```
goldfirere, line 1612: I don't see how this comment corresponds to the implementation. It says not to do something if there is a manifest.... but the code does something if there is a manifest. Could you try rewording this?

ocaml/testsuite/tests/typing-layouts-or-null/test_or_null.ml
```diff
@@ -0,0 +1,201 @@
...
+type 'a t = 'a or_null = Null | This of 'a
+|}]
+
+let to_option (x : 'a or_null) =
+  match x with
+  | Null -> None
+  | This x -> Some x
+
+[%%expect{|
+val to_option : 'a or_null -> 'a option = <fun>
```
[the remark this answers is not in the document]
goldfirere: Oops. My mistake. Ignore.

ocaml/testsuite/tests/typing-layouts-or-null/test_or_null.ml
goldfirere: Oops. My mistake. Ignore.

ocaml/testsuite/tests/typing-layouts-or-null/test_or_null.ml
dkalinichenko-js: Done.

ocaml/testsuite/tests/typing-layouts-or-null/test_or_null.ml
dkalinichenko-js: Done.

ocaml/typing/predef.ml
dkalinichenko-js: Added a comment.

ocaml/typing/typedecl.ml
dkalinichenko-js: Edited.

goldfirere (approved):

PR #2750, opened 2024-07-01
Honor `[@error_message]` attribute even when its location is ghost

ncik-roberts:
Make the compiler report the error payload of `[@error_message ...]` even in the event that the constrained location is a ghost. This better enables ppxes to insert `[@error_message ...]` themselves, as ppxes often create ghost locations.

Cross-linking with #1943 and #1954. 

Most of this feature is a regression test. It's kind of finicky to get `[@error_message ...]` to work &mdash; it only works when certain syntactic constructs appears as the payload of the constraint. (E.g., it works for variants and not for identifiers &mdash; the explanation mechanism has an interaction with whether the construct is "inferred", and `[@error_message ...]` inherits that interaction.) So the test is fairly contrived.

### DIFF
--- a/ocaml/testsuite/tests/ppx-error-message/ppx_error_message.ml
+++ b/ocaml/testsuite/tests/ppx-error-message/ppx_error_message.ml
@@ -1,0 +1,53 @@
+open Ast_mapper
+
+(* Assert statically that a string that appears in the source text
+   is alphabetically ordered. This is a bit contrived so that we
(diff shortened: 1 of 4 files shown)


ncik-roberts:
Oh, I forgot to mention: the first commit adds a regression test, and the second commit fixes it and updates the test output.

goldfirere (approved):

PR #2754, opened 2024-07-02
Load/store `float32x4` <-> `float32# array`

TheNumbat:
Adds additional 128-bit load/store primitives for `float32# array`s. 
They operate on `float32x4` SIMD vectors, and are otherwise identical to the primitives for other arrays of unboxed numbers.

Tests have been added in `tests/simd/arrays.ml`.


tests/simd/arrays.ml
```diff
@@ -538,6 +566,73 @@ module Float_arrays = struct
     fail [|#0.0|] 1;
     fail [|#0.0|] (-1)
   ;;
+
+  let () =
+    let unboxed_float32_array = unboxed_float32_array () in
+    let _0123 = f32x4 0.0s 1.0s 2.0s 3.0s in
```
ccasin, line 572: This variable name offends my delicate sensibilities.  As a general rule, reserve the `_` prefix for variables that are unused.  This can be `f32x4_0123`, or any other name of your choice.

middle_end/flambda2/from_lambda/closure_conversion.ml
```diff
@@ -865,13 +865,14 @@ let close_primitive acc env ~let_bound_ids_with_kinds named
...
-      | Pctconst _ | Pbswap16 | Pbbswap _ | Pint_as_pointer _ | Popaque _
-      | Pprobe_is_enabled _ | Pobj_dup | Pobj_magic _ | Punbox_float _
+      | Punboxed_float32_array_load_128 _ | Punboxed_int32_array_load_128 _
+      | Punboxed_int64_array_load_128 _ | Punboxed_nativeint_array_load_128 _
+      | Pfloatarray_set_128 _ | Pfloat_array_set_128 _ | Pint_array_set_128 _
+      | Punboxed_float_array_set_128 _ | Punboxed_float32_array_set_128 _
+      | Punboxed_int32_array_set_128 _ | Punboxed_int64_array_set_128 _
+      | Punboxed_nativeint_array_set_128 _ | Pctconst _ | Pbswap16 | Pbbswap _
+      | Pint_as_pointer _ | Popaque _ | Pprobe_is_enabled _ | Pobj_dup
+      | Pobj_magic _ | Punbox_float _
```
ccasin, line 875: I think it would be better to add the two new constructors on their own lines, or at least leave the stuff starting at `Pctconst _` untouched, just to have a clearer diff and a nicer history.

ccasin (approved):
Looks good, just a couple nits.  An flambda2 person should confirm we expect `array_like_load_128` to work with the `Naked_float32s` array kind.

tests/simd/arrays.ml
TheNumbat: Sure, renamed

middle_end/flambda2/from_lambda/closure_conversion.ml
TheNumbat: This file uses ocamlformat, so I think we're stuck with the layout changes unless we want to disable it for this function

ccasin: Ah, sorry for the noise

mshinwell:
Seems fine.

PR #2755, opened 2024-07-02
ocamltest improvements + make sure all tests actually do something

mshinwell:
The main change is to add a "does this action do something?" flag to all actions defined by ocamltest for the use of test scripts.  Then it is enforced that all branches of a test must do something.  This should avoid what's happening at the moment, which is that a chunk of tests aren't running at all, because of stanzas like:
```
(* TEST
   runtime5;
*)
```
In the absence of any other explicit directives like `bytecode;` or `native;` this actually _removes_ the default bytecode/native compile+run actions, and just leaves a predicate test and nothing else.  The test therefore does nothing.

It is terrible that we only discovered this because of a locals bug (https://github.com/ocaml-flambda/flambda-backend/pull/2736).  There, a new test case could not be made to fail even with a syntax error, which was because in fact the test directives did not describe anything substantive to do.

The "does this action do something?" logic isn't perfect; for example if you are going to compile then run, then arguably the compilation step should not "do something".  However this is already much better than what we have, and I think will catch the fundamental errors.

Two other improvements are included, which can be split into separate PRs if necessary:
1. Introduce an explicit notion of "predicate", so that diagnostics for these ones can be improved.  For example we now get things like the following (this is an example when configured for runtime5):
```
 ... testing 'test_instrumented.ml' with 1 (runtime4)
  => predicate "4.x runtime being used" is not satisfied
```
2. Stop printing dozens of `n/a` lines where test directives are skipped because of a previous predicate which was false.

Finally there are fixes to various tests to make them actually run.  Two of the `runtime_events` ones fail (one because of some build failure where `stubs.c` is apparently not being linked, and another because of something semantic in the test itself) but I've disabled these for now, since we're not using that library.  The fixes for statmemprof tests are from @stedolan with thanks.


ocaml/testsuite/tests/statmemprof/stop_start_in_callback.ml
```diff
@@ -28,7 +31,7 @@ module AllocSet = Set.Make(Int3Tuples)
 (* A combined 7-block 33-word allocation *)
 
 let[@inline never] f33 n =
-  ((n, n, (n, n, n, (n,n,n,n,n))), (n, n, (n, n, n, (n,n,n,n,n))))
+  ((n, n, (n, n, n, (n,n,n,n,n))), (n, n, (n, n, n, (n,n,n,n,0))))
```
xclerc, line 34: (I don't see how the change could be
a problem, but am curious why it was
applied.)

ocaml/testsuite/tests/lib-runtime-events/test_caml.ml
```diff
@@ -1,12 +1,25 @@
 (* TEST
+ {
+   runtime4;
+   skip;
+ }{
```
xclerc, line 5: Is this block still necessary if we
put `runtime5` in the other one?

ocaml/testsuite/tests/backtrace/backtrace_c_exn.ml
```diff
@@ -45,7 +45,13 @@ let () =
      native;
    }
  }{
-   reference = "${test_source_directory}/backtrace_c_exn.byte.reference";
    runtime5;
+   {
+     reference = "${test_source_directory}/backtrace_c_exn.byte.reference";
+     bytecode;
+   }{
+     reference = "${test_source_directory}/backtrace_c_exn.byte.reference";
```
xclerc, line 53: Is `byte` correct on this line?
(if so, `reference` can possibly
be factored out)

ocaml/ocamltest/result.ml
```diff
@@ -15,7 +15,7 @@
 
 (* Definition of test-result related types and functions *)
 
-type status = Pass | Skip | Fail
+type status = Pass | Skip | Fail | Predicate of bool
```
xclerc, line 18: I wonder whether the new cases
could/should be nested under
existing constructors. The `is_pass`
and `is_skip` functions seem to
suggest predicates are ways of
passing / skipping.
That being said, we probably
don't want to spend time on
such questions.

ocaml/Makefile.common-jst
```diff
@@ -60,7 +60,8 @@ boot_targets = \
   $(boot_ocamlopt) \
   $(boot_ocamlmklib) \
   $(boot_ocamldep) \
-  $(boot_ocamlobjinfo)
+  $(boot_ocamlobjinfo) \
+  ocaml/ocamltest/ocamltest.byte
```
xclerc, line 64: (Is it a left-over?)

ocaml/ocamltest/tsl_semantics.ml
```diff
@@ -92,6 +92,19 @@ type test_tree =
     string located list *
     (test_tree list)
 
+let tests_do_something (tests : Tests.t) =
+  List.exists Actions.does_something tests.test_actions
```
xclerc, line 96: (It is probably fine as it stands, but
shouldn't the list end with an action
that does something?)

ocaml/ocamltest/tsl_semantics.ml
```diff
@@ -92,6 +92,19 @@ type test_tree =
...
+
+let rec test_tree_does_something_on_all_branches tree =
+  match tree with
+  | Node (_, tests, _, children) ->
+    let from_children =
+      List.fold_left (fun acc child ->
```
xclerc, line 102: It looks like `List.exists`, but I
think I think I was expecting
a `List.for_all` (with the condition
over `tests` under `for_all`).

ocaml/ocamltest/actions_helpers.ml
```diff
@@ -17,19 +17,30 @@
...
 
-let pass_or_skip test pass_reason skip_reason _log env =
+let predicate test pass_reason _skip_reason _log env =
   let open Result in
   let result =
     if test
-    then pass_with_reason pass_reason
-    else skip_with_reason skip_reason in
+    then predicate_satisfied_with_reason pass_reason
+    else predicate_not_satisfied_with_reason pass_reason in
```
xclerc, line 43:
```suggestion
    else predicate_not_satisfied_with_reason skip_reason in
```

xclerc: (not sure; if not the parameter should
probably be removed)

ocaml/testsuite/tests/flambda/approx_meet.ml
```diff
@@ -1,10 +1,4 @@
-(* TEST
- {
-   flambda;
- }{
-   native;
- }
-*)
+(* TEST *)
```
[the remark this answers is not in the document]
mshinwell: Yes, but we only have one middle end...

ocaml/testsuite/tests/flambda/gpr2239.ml
```diff
@@ -1,10 +1,4 @@
-(* TEST
- {
-   flambda;
- }{
-   native;
- }
-*)
+(* TEST *)
```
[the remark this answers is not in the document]
mshinwell: See above.

ocaml/testsuite/tests/statmemprof/stop_start_in_callback.ml
mshinwell: I think it's to avoid a case where flambda2 can see two block allocations are the same.

ocaml/testsuite/tests/lib-runtime-events/test_caml.ml
mshinwell: Maybe not, but it was there already, so I think this is ok.  We'll come back to the runtime_events tests in due course anyway.

ocaml/Makefile.common-jst
mshinwell: Ah no, I forgot to describe this change.

mshinwell:
One other change here that I forgot about: this builds `ocamltest` as part of `make hacking`, which makes it much easier to work on (e.g. merlin works).  It's very fast to build so I don't think this should be problematic.

ocaml/ocamltest/tsl_semantics.ml
mshinwell: Hmm.  Maybe it should be: if the test has children (which iiuc means another line that follows, after a semicolon) then `test_actions` doesn't need to do something, so long as all of the children do.  However if the test doesn't have children then I think you are right, and we need to check the list.  Does that sound plausible?

ocaml/ocamltest/ocaml_actions.ml
```diff
@@ -612,7 +619,9 @@ let objinfo log env =
 
 let ocamlobjinfo =
   Actions.make ~name:"ocamlobjinfo"
-    ~description:"Run ocamlobjinfo on the program" objinfo
+    ~description:"Run ocamlobjinfo on the program"
+    ~does_something:false
```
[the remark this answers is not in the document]
mshinwell: Fixed

ocaml/testsuite/tests/backtrace/backtrace_c_exn.ml
mshinwell: I think this is right since the CI passes, but I've renamed this file.

ocaml/ocamltest/result.ml
mshinwell: Not a bad idea indeed, but I've just left a CR for the moment.

mshinwell: (We shouldn't spend too long on this as it needs to go upstream, and there may be further discussion there resulting in changes.)

ocaml/testsuite/tests/lib-runtime-events/test_caml.ml
xclerc: (that will indeed equally apply to the commented-out version below)

ocaml/ocamltest/tsl_semantics.ml
xclerc: I think you are right, each possible string
should end with an action doing something,
but contrary to what I implied a string is
the sequence of actions from the root
to a leaf - concatenating the lists of actions
along the path.

ocaml/ocamltest/actions_helpers.ml
mshinwell: Actually we always want the pass-reason here, otherwise you get a double negation and the message printed is wrong.  I'll add a comment, but won't remove the parameter yet to minimize the diff with upstream.

ocaml/stdlib/marshal.ml
```diff
@@ -58,7 +58,7 @@ let data_size buff ofs =
   else data_size_unsafe buff ofs
 let total_size buff ofs = header_size + data_size buff ofs
 
-let from_bytes buff ofs =
+let[@inline never] from_bytes buff ofs =
```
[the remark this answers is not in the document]
mshinwell: It's because the backtrace is being checked by a statmemprof test.  This doesn't seem great but given what the function does, it seems fine I think?

ocaml/ocamltest/tsl_semantics.ml
mshinwell: This should be fixed as discussed

ocaml/ocamltest/tsl_semantics.ml
mshinwell: Should be fixed

xclerc (approved):

ocaml/stdlib/marshal.ml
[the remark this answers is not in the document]
mshinwell: This has been fixed

PR #2756, opened 2024-07-02
Fix missing rewrite ids when unboxing detects invalid apply conts

Gbury:
In some cases (e.g. a do_not_unbox decision, followed by a decision that does unbox, and that discovers that some apply conts are invalids), it could happen that the set of rewrite ids known as being invalid was dropped/reset to empty. That together with the caching of the extra args computation (done through the rewrite ids seen) meant that we could "forget" some rewrite ids.

### DIFF
--- a/middle_end/flambda2/simplify/continuation_extra_params_and_args.ml
+++ b/middle_end/flambda2/simplify/continuation_extra_params_and_args.ml
@@ -95,18 +95,34 @@ let add t ~invalids ~extra_param ~extra_args =
     let extra_params = Bound_parameters.cons extra_param extra_params in
     let extra_args =
       Apply_cont_rewrite_id.Map.merge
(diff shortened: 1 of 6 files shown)


lthls (approved):
The patch looks correct. @Gbury do you have an example that reproduces the bug, that we could add to the testsuite ?

Gbury:
This PR has been updated with
- a fix for a second bug found while investigating the first one
- some more informative error messages
- testcases for the two bugs fixed by this PR

lthls (approved):
The new patch looks corect too. Please put `[@@@flambda_o3]` in the tests to ensure we test them in the right conditions.

Gbury:
@lthls Done (although using `O3` is not necessary for these tests to fail on current `main`).

PR #2758, opened 2024-07-03
Fix int64->float32 cast codegen

TheNumbat:
If `gas` encounters `cvtsi2ss` where the first (integer) operand is a memory location, it will assume that the int is 32 bits wide.
Hence, we need to suffix the instruction name with the width of the integer operand.
This matches the usage of `cvtsi2sd`.

This bug does not apply to the internal assembler, which already cases on the operand width.
We believe it does not apply to the nasm assembler (no suffix was required for `cvtsi2sd`), but nasm isn't tested, and isn't supported in flambda-backend anyway.

(Bug found / fix tested in #2697)


xclerc (approved):
(Already reviewed as part of #2697.)

PR #2762, opened 2024-07-05
flambda2: Don't change coercions when returning from `Aliases.add`

lukemaurer:
`Aliases.add` currently has a somewhat more involved contract than necessary: it always returns (in the `add_result` record) a `canonical_element` with no coercion and an `alias_of_demoted_element` with a correspondingly adjusted coercion. This does not actually help anything - the sole caller doesn't require that `canonical_element` has no coercion. Rather, it requires that the returned values are now aliases up to coercion. This is of course precisely what `Aliases.add` accomplishes, so we can just as happily return the original arguments un-faffed-with.


lthls:
Thanks ! I'll review this next week

lthls (approved):
Looks good.

PR #2763, opened 2024-07-08
Fix padding bug for float32 static values

mshinwell:
The code in flambda2 to dynamically initialize static values containing `float32`s was correctly taking into account that they must be put into word-width slots.  However the code for constant static fields was not, meaning that fields after a `float32` would be at addresses 4 bytes too low.

The test case is in https://github.com/ocaml-flambda/flambda-backend/pull/2698.  @ncik-roberts please add that as a test case in PR2698, since it needs inline mixed blocks.

### DIFF
--- a/backend/cmm_helpers.ml
+++ b/backend/cmm_helpers.ml
@@ -3882,6 +3882,8 @@ let infix_field_address ~dbg ptr n =
 
 let cint i = Cmm.Cint i
 
+let cint32 i = Cmm.Cint32 (Nativeint.of_int32 i)
+
 let cfloat32 f = Cmm.Csingle f
 
 let cfloat f = Cmm.Cdouble f
--- a/middle_end/flambda2/to_cmm/to_cmm_shared.ml
+++ b/middle_end/flambda2/to_cmm/to_cmm_shared.ml
@@ -214,7 +214,9 @@ let const_static cst =
   | Naked_float f -> [cfloat (Numeric_types.Float_by_bit_pattern.to_float f)]
   | Naked_float32 f ->
     (* Statically-allocated float32 values are zero padded. We must explicitly
-       add the padding otherwise subsequent values will be misaligned. *)
+       add the padding otherwise subsequent values will be misaligned. If this
+       code is ever used on big-endian systems (which seems unlikely), this
+       needs checking. *)
     [cfloat32 (Numeric_types.Float32_by_bit_pattern.to_float f); cint32 0l]
   | Naked_int32 i -> [cint (Nativeint.of_int32 i)]
   (* We don't compile flambda-backend in 32-bit mode, so nativeint is 64
--- a/middle_end/flambda2/to_cmm/to_cmm_shared.ml
+++ b/middle_end/flambda2/to_cmm/to_cmm_shared.ml
@@ -213,10 +213,9 @@ let const_static cst =
            (tag_targetint (Targetint_31_63.to_targetint i))) ]
   | Naked_float f -> [cfloat (Numeric_types.Float_by_bit_pattern.to_float f)]
   | Naked_float32 f ->
-    (* Here we are relying on the data section being zero initialized to
-       maintain the invariant that statically-allocated float32 values are zero
-       padded. *)
-    [cfloat32 (Numeric_types.Float32_by_bit_pattern.to_float f)]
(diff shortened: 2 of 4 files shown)


ncik-roberts:
The test case illustrated in the comments of https://github.com/ocaml-flambda/flambda-backend/pull/2698 doesn’t require inline record mixed blocks and can be copied here.

mshinwell:
> The test case illustrated in the comments of #2698 doesn’t require inline record mixed blocks and can be copied here.

Ah, indeed I see now.  I've added it here.

mshinwell:
This bug is probably present in https://github.com/ocaml-flambda/flambda-backend/pull/2712/ but I will fix that (and we can add some test cases there to form more robust checks, immune to closure offset computation changes).

ncik-roberts:
Thanks. I’m planning on writing test cases for #2712 tomorrow.

backend/cmm_helpers.ml
```diff
@@ -3882,6 +3882,8 @@ let infix_field_address ~dbg ptr n =
 
 let cint i = Cmm.Cint i
 
+let cint32 i = Cmm.Cint32 (Nativeint.of_int32 i)
```
lthls, line 3885: #2712 introduces a different `cint32` function. I think this one is better, and #2712 should be changed to use it.

middle_end/flambda2/to_cmm/to_cmm_shared.ml
```diff
@@ -213,10 +213,9 @@ let const_static cst =
...
   | Naked_float f -> [cfloat (Numeric_types.Float_by_bit_pattern.to_float f)]
   | Naked_float32 f ->
-    (* Here we are relying on the data section being zero initialized to
-       maintain the invariant that statically-allocated float32 values are zero
-       padded. *)
-    [cfloat32 (Numeric_types.Float32_by_bit_pattern.to_float f)]
+    (* Statically-allocated float32 values are zero padded. We must explicitly
+       add the padding otherwise subsequent values will be misaligned. *)
+    [cfloat32 (Numeric_types.Float32_by_bit_pattern.to_float f); cint32 0l]
   | Naked_int32 i -> [cint (Nativeint.of_int32 i)]
```
lthls, line 219: I wouldn't mind changing the `Naked_int32` case to use two `cint32` instead of one `cint`. It is more obviously correct and does not depend on endianness.

lthls (approved):
The fix looks correct. I've left a few comments that do not need to delay this PR.

backend/cmm_helpers.ml
mshinwell: I'll do this

middle_end/flambda2/to_cmm/to_cmm_shared.ml
mshinwell: Agreed, but let's do this in a separate PR

Gbury (approved):

middle_end/flambda2/to_cmm/to_cmm_shared.ml
```diff
@@ -213,10 +213,9 @@ let const_static cst =
            (tag_targetint (Targetint_31_63.to_targetint i))) ]
   | Naked_float f -> [cfloat (Numeric_types.Float_by_bit_pattern.to_float f)]
   | Naked_float32 f ->
-    (* Here we are relying on the data section being zero initialized to
-       maintain the invariant that statically-allocated float32 values are zero
-       padded. *)
-    [cfloat32 (Numeric_types.Float32_by_bit_pattern.to_float f)]
+    (* Statically-allocated float32 values are zero padded. We must explicitly
+       add the padding otherwise subsequent values will be misaligned. *)
+    [cfloat32 (Numeric_types.Float32_by_bit_pattern.to_float f); cint32 0l]
```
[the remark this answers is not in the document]
mshinwell: I've added a note, but without committing to behaviour.  For one thing I don't think it's worth the time to think about, as we are very unlikely to run this code on a big-endian system; and in any case, I think it depends on what instruction is used to read the value.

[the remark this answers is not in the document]
lthls: It's not like `int32`, where as long as the padding bits encode the sign properly we can load a 64-bit int directly to avoid sign-extension. Here we will always load exactly the first 4 bytes, so this code is correct independently of the endianness.

PR #2764, opened 2024-07-09
Remove variables from computed signature for `with module` constriants.

ccasin:
This fixes a bug where zero_alloc and modality variables could sneak into value descriptions in signatures in a disallowed way via `with module` constraints.  See the test for an example.  These computed signatures are very much like `module type of` signatures, so we just take the same approach.  

There are two commits: The first adds the test and shows it fails.  The second fixes it.


riaqn (approved):
LGTM (provided that `remove_modality_and_zero_alloc_variables_*` are moved without change)

PR #2767, opened 2024-07-10
Use correct `zap` function for modalities in `with module` constraints

ccasin:
This fixes a bug in #2764, which "fixed" a bug in #2742.

The original bug was that `zero_alloc` and modality variables could creep into signatures in a disallowed way when using a `with module` constraint.  The fix was to make these variables into constants in the computed signature for the constraint.  But in the case of modality variables, I picked a constant that could lead to spurious typing errors, and failed to notice it because we only create these variables if `-extension mode_alpha` is used.

There are three commits:
* The first adds a test demonstrating the issue
* The second fixes it
* The third adds more test documenting the behavior.

(There remain some questions about whether the revised behavior is ideal, and the tests in the third commit document the behavior.)

### DIFF
--- a/ocaml/typing/typemod.ml
+++ b/ocaml/typing/typemod.ml
@@ -803,7 +803,7 @@ let merge_constraint initial_env loc sg lid constr =
         let mty = Mtype.scrape_for_type_of ~remove_aliases sig_env mty in
         let mty =
           remove_modality_and_zero_alloc_variables_mty sig_env
-            ~zap_modality:Mode.Modality.Value.zap_to_floor mty
+            ~zap_modality:Mode.Modality.Value.zap_to_id mty
         in
         let md'' = { md' with md_type = mty } in
         let newmd = Mtype.strengthen_decl ~aliasable:false md'' path in
--- a/ocaml/testsuite/tests/typing-modes/val_modalities.ml
+++ b/ocaml/testsuite/tests/typing-modes/val_modalities.ml
@@ -398,47 +398,9 @@ end = struct
 end
 [%%expect{|
 module type S = sig module M : sig val f : int -> int end end
-Lines 13-19, characters 6-3:
-13 | ......struct
-14 |   module Plain = struct
-15 |     let f x = x+1
-16 |   end
-17 |
-18 |   module type S_plain = S with module M = Plain
-19 | end
-Error: Signature mismatch:
-       Modules do not match:
-         sig
-           module Plain : sig val f : int -> int @@ global many end
-           module type S_plain =
-             sig module M : sig val f : int -> int @@ global many end end
-         end
-       is not included in
-         sig
-           module Plain : sig val f : int -> int end
-           module type S_plain =
-             sig module M : sig val f : int -> int end end
-         end
-       Module type declarations do not match:
-         module type S_plain =
-           sig module M : sig val f : int -> int @@ global many end end
-       does not match
-         module type S_plain = sig module M : sig val f : int -> int end end
-       The second module type is not included in the first
-       At position module type S_plain = <here>
-       Module types do not match:
-         sig module M : sig val f : int -> int end end
-       is not equal to
-         sig module M : sig val f : int -> int @@ global many end end
-       At position module type S_plain = sig module M : <here> end
-       Modules do not match:
-         sig val f : int -> int end
-       is not included in
-         sig val f : int -> int @@ global many end
-       At position module type S_plain = sig module M : <here> end
-       Values do not match:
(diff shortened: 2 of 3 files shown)


ocaml/typing/typemod.ml
```diff
@@ -802,7 +802,7 @@ let merge_constraint initial_env loc sg lid constr =
         let mty = Mtype.scrape_for_type_of ~remove_aliases sig_env mty in
         let mty =
           remove_modality_and_zero_alloc_variables_mty sig_env
-            ~zap_modality:Mode.Modality.Value.zap_to_floor mty
```
goldfirere, line 805: I can't think of any correct use of `zap_to_floor`. And a quick search seems to find this use (which is wrong) and also a use in printing (which I believe is also wrong). @riaqn -- what do you think? I think we always want `zap_to_id`.

ocaml/typing/typemod.ml
```diff
@@ -802,7 +802,7 @@ let merge_constraint initial_env loc sg lid constr =
...
+            ~zap_modality:Mode.Modality.Value.zap_to_id mty
```
goldfirere, line 805: I think maybe this is fine for now, but really the problem is that `with module` shouldn't touch `val`s at all. The manual describes `with module` as affecting only type equalities. If that were the case, this would all be moot.

Or maybe we want `with module` to also affect modalities on `val`s? That actually seems like perhaps the right idea. Is this documented anywhere?

goldfirere (approved):

ocaml/typing/typemod.ml
riaqn: Generating .cmi without .mli or printing back the type of a user-written module in REPL, describes the structure. In these cases, the inferred modalities default to the strongest possible.

ocaml/typing/typemod.ml
riaqn: `with modules` was overlooked when we designed `val_modalities`, for the reason in your first paragraph.

But now i think about this, since `with module` is strenghtening, then it might make sense that `with module` strenghthens the modalities. I will look.

riaqn:
`with module` is able to strenghthen. In the manual:
>  After each constraint has been applied, the resulting signature must be a subtype of the signature before the constraint was applied. Thus, the with operator can only add information on the type components of a signature, but never remove information.

In the following example, both `type t` and `val f` is strengthened:
```
module type S = sig
  module N : sig
    type t
    val f : t -> t
  end
end

module N0 : sig
  type t = int
  val f : 'a -> 'a
end = struct
  type t = int
  let f a = a
end

module type S' = S with module N = N0
```
and I can see that `typemod.patch_item` checks `N0` is stronger than `N`. So I think typemod is doing the right thing here.

riaqn:
Following the last comment, I now think that it makes sense that `with module N = N0` will look inside the type of `N0` (in particular, `val_modalities`) to strenghthen `N` on a per-signature-item basis. Therefore, `val_modalities` which could be "variable" modalities need to be defaulted. And it should be `zap_to_id` just to give the best compatibility - for example:
```
module type S = sig
  module M : sig
    val f : int -> int
  end
end

module M0 = struct
    let f a = a
end

module type S_plain = S with module M = M0
```
If we `zap_to_floor`, then `S_plain.M.f` will have `@@ portable` on it, which might confuse the user.

On the other hand, if the user write `let (f @ portable) a = a`, then `S_plain.M.f` should be `@@ portable`. This is currently not the case for `zap_to_id` and I will fix that shortly with a commit here.

ocaml/typing/typemod.ml
riaqn: Yeah that's right. For monadic axes, the modalities are of the form `join_with_c`. And the identity modality is `join_with_min`.

riaqn:
Apologize for delaying the promised fix. I will do this ASAP.

riaqn:
OK to implement what I suggested would require some extra work. Currently when people write `let foo @ portable = bar`, it get typed checked as `let foo = bar @ portable`. A single mode variable `m` will be used on both LHS and RHS. The `@ portable` constrains `m >= portable` and proceeds to check `bar` at `portable`. As a result, `m` (which will also be the mode of `foo`) is not constrained at all.

To fix this properly, we probably need to modify `vb_pat_constraint` and the type checking of `Ppat_constraint`.

I suggest that for now we merge this PR as it is. I will resolve conflict and push.

ocaml/typing/typemod.ml
riaqn: @goldfirere can I resolve this?

PR #2770, opened 2024-07-11
Remove `Stdlib_stable` float32 modules in arm64 build

TheNumbat:
Reshuffles `Stdlib_stable` so that the `Float32` and `Float32_u` modules are not included on non-amd64 architectures.
This should fix the arm64 build, @mshinwell can you check if it works?


dkalinichenko-js (approved):
LGTM, though I think we can autogenerate the copying build rules. But let's do that later.

PR #2772, opened 2024-07-12
Parse syntax for declaring module type with an underscore

smjleo:
The goal is to allow syntax that looks like
```ocaml
module M : sig
  module type S = sig ... end
end = struct
  module type S = _
end
```
where the usage of the underscore allows the typechecker to infer the type of `S` from the earlier definition.
This PR makes necessary changes to the parser and parsetree (as well as miscellaneous changes to accommodate the new parsetree) to recognise this syntax.

### DIFF
--- a/ocaml/parsing/parser.mly
+++ b/ocaml/parsing/parser.mly
@@ -1891,30 +1891,20 @@ module_type_declaration:
     ext = ext
     attrs1 = attributes
     id = mkrhs(ident)
-    typ = preceded(EQUAL, module_type)?
+    typ = module_type_declaration_type
     attrs2 = post_item_attributes
     {
       let attrs = attrs1 @ attrs2 in
       let loc = make_loc $sloc in
       let docs = symbol_docs $sloc in
-      let pmtd_typ = match typ with
-        | None -> Pmtd_abstract
-        | Some t -> Pmtd_define t
-      in
-      Mtd.mk id pmtd_typ ~attrs ~loc ~docs, ext
+      Mtd.mk id typ ~attrs ~loc ~docs, ext
     }
-  | MODULE TYPE
-    ext = ext
-    attrs1 = attributes
-    id = mkrhs(ident)
-    EQUAL UNDERSCORE
-    attrs2 = post_item_attributes
-    {
-      let attrs = attrs1 @ attrs2 in
-      let loc = make_loc $sloc in
-      let docs = symbol_docs $sloc in
-      Mtd.mk id Pmtd_underscore ~attrs ~loc ~docs, ext
-    }
+;
+
+module_type_declaration_type:
+  | (* empty *)         { Pmtd_abstract }
+  | EQUAL module_type   { Pmtd_define $2 }
+  | EQUAL UNDERSCORE    { Pmtd_underscore }
 ;
 
 (* -------------------------------------------------------------------------- *)
--- a/ocaml/parsing/parser.mly
+++ b/ocaml/parsing/parser.mly
@@ -1887,18 +1887,34 @@ include_maybe_functor:
 
 (* A module type declaration. *)
 module_type_declaration:
-  MODULE TYPE
-  ext = ext
-  attrs1 = attributes
-  id = mkrhs(ident)
-  typ = preceded(EQUAL, module_type)?
-  attrs2 = post_item_attributes
-  {
-    let attrs = attrs1 @ attrs2 in
-    let loc = make_loc $sloc in
-    let docs = symbol_docs $sloc in
-    Mtd.mk id ?typ ~attrs ~loc ~docs, ext
-  }
+  | MODULE TYPE
+    ext = ext
+    attrs1 = attributes
+    id = mkrhs(ident)
+    typ = preceded(EQUAL, module_type)?
+    attrs2 = post_item_attributes
+    {
+      let attrs = attrs1 @ attrs2 in
+      let loc = make_loc $sloc in
+      let docs = symbol_docs $sloc in
+      let pmtd_typ = match typ with
+        | None -> Pmtd_abstract
+        | Some t -> Pmtd_define t
+      in
+      Mtd.mk id pmtd_typ ~attrs ~loc ~docs, ext
+    }
+  | MODULE TYPE
+    ext = ext
+    attrs1 = attributes
+    id = mkrhs(ident)
+    EQUAL UNDERSCORE
+    attrs2 = post_item_attributes
+    {
+      let attrs = attrs1 @ attrs2 in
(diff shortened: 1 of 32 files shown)


ocaml/parsing/parsetree.mli
```diff
@@ -921,16 +921,24 @@ and module_substitution =
...
      pmtd_attributes: attributes;  (** [... [\@\@id1] [\@\@id2]] *)
      pmtd_loc: Location.t;
     }
 (** Values of type [module_type_declaration] represents:
    - [S = MT],
+     when {{!module_type_declaration.pmtd_type}[pmtd_type]} is [Pmtd_define MT].
    - [S] for abstract module type declaration,
-     when {{!module_type_declaration.pmtd_type}[pmtd_type]} is [None].
+     when {{!module_type_declaration.pmtd_type}[pmtd_type]} is [Pmtd_abstract].
+   - [S = _] for inferring module type from definition in the signature,
```
riaqn, line 933: Might be simpler if we just say "inferring module type". Because if we say "in the signature" it involves the distinction between signature and structure, and the fact that underscore is invalid in signature, etc.. But this is `parsetree.mli` whose `module_type_decalration` doesn't know whether it's in a signature or a structure.

ocaml/parsing/parsetree.mli
```diff
@@ -921,16 +921,24 @@ and module_substitution =
...
+     when {{!module_type_declaration.pmtd_type}[pmtd_type]} is [Pmtd_underscore].
 *)
 
+and module_type_declaration_type =
+  | Pmtd_abstract
+  | Pmtd_underscore
+  | Pmtd_define of module_type
+
```
riaqn, line 941: Might be better if we move the above comment to here. Previously it was just an `option`, so the comment was on the field; but now it's a dedicated type, which deserves comment.

ocaml/parsing/pprintast.ml
```diff
@@ -1468,16 +1468,16 @@ and signature_item ctxt f x : unit =
       pp f "@[<hov2>module@ type@ %s%a@]%a"
         s.txt
         (fun f md -> match md with
-           | None -> ()
-           | Some mt ->
+           | Pmtd_abstract | Pmtd_underscore -> ()
```
riaqn, line 1471: `pprintast` is a pretty printer that should reproduce user input (but prettier). So here `Pmtd_underscore` should produce underscore.

ocaml/parsing/pprintast.ml
```diff
@@ -1766,8 +1766,8 @@ and structure_item ctxt f x =
...
```
riaqn, line 1769: Similarly here.

ocaml/typing/typedtree.ml
```diff
@@ -551,12 +551,12 @@ and module_substitution =
...
-     mtd_uid: Uid.t;
-     mtd_type: module_type option;
-     mtd_attributes: attribute list;
-     mtd_loc: Location.t;
+      mtd_id: Ident.t;
+      mtd_name: string loc;
+      mtd_uid: Uid.t;
+      mtd_type: module_type option;
+      mtd_attributes: attribute list;
+      mtd_loc: Location.t;
```
riaqn, line 559: It's better if code style changes are separated from semantics changes. (and in this case, I think those are just extra spaces?)

ocaml/typing/typemod.ml
```diff
@@ -4110,6 +4118,9 @@ let report_error ~loc _env = function
         "This value is %a, but expected to be %a because it is inside a module."
         (Mode.Value.Const.print_axis ax) left
         (Mode.Value.Const.print_axis ax) right
+  | Underscore_not_allowed_in_signature ->
+    Location.errorf ~loc
+      "Inferrence of module types is not allowed within a signature."
```
riaqn, line 4123: We need two extra spaces for indentation.

ocaml/testsuite/tests/module-type-underscore/typing.ml
```diff
@@ -0,0 +1,91 @@
...
+end = struct
+  module type S = _
+end ;;
+[%%expect {|
+Uncaught exception: Failure("underscore not implemented")
+
+|}]
+
+module M : sig
+  module type S = _
```
riaqn, line 20: Should this trigger the error "underscore not allowed in signature"?

riaqn (commented):
Thanks - I reviewed everything except ocamldoc.

riaqn (commented):
I've reviewed the ocamldoc part as well, and they looks sensible.

ocaml/testsuite/tests/module-type-underscore/typing.ml
riaqn: If this type error is hard to catch currently (without proper typing), that means the type error is imprecise, then maybe we don't want this type error at all (i.e., just raise whenever spot an underscore).

ocaml/parsing/parsetree.mli
smjleo: Discussed offline - we think it's better to leave it where it is, since the comment is describing what `module_type_declaration` is (even though it references `module_type_declaration_type`. It seems hard to describe the latter without describing the former.

ocaml/typing/typemod.ml
```diff
@@ -4119,8 +4119,8 @@ let report_error ~loc _env = function
         (Mode.Value.Const.print_axis ax) left
         (Mode.Value.Const.print_axis ax) right
   | Underscore_not_allowed_in_signature ->
-    Location.errorf ~loc
-      "Inferrence of module types is not allowed within a signature."
+      Location.errorf ~loc
+        "Inferrence of module types is not allowed within a signature."
```
riaqn, line 4123: This can be removed, now that they are never triggered.

ocaml/parsing/parser.mly
```diff
@@ -1887,18 +1887,24 @@ include_maybe_functor:
...
+    id = mkrhs(ident)
+    typ = module_type_declaration_type
+    attrs2 = post_item_attributes
+    {
+      let attrs = attrs1 @ attrs2 in
+      let loc = make_loc $sloc in
+      let docs = symbol_docs $sloc in
+      Mtd.mk id typ ~attrs ~loc ~docs, ext
+    }
+;
```
riaqn, line 1902: After refactoring, this rule has only one line of diff. We want to minimizes diffs, so that in the long run the origin of code is easier to track.

riaqn (approved):
Looks good - can merge once the two additional points are addressed.

ocaml/parsing/parser.mly
smjleo: Great point - thank you!

PR #2774, opened 2024-07-12
Port 12669 (remove global state in `schedgen.ml`)

xclerc:
As per title, but we should consider whether we also
want to port missing bits (See `try_nesting`).


xclerc:
Closed in favor of https://github.com/ocaml-flambda/flambda-backend/pull/2789.

PR #2776, opened 2024-07-12
Type shuffle for general higher kinds

jbachurski:
This is the first PR for the higher-kinded polymorphism project.
It has been artificially separated out from my implementation of just the first step from the design doc - 'Arrow kinds', as the appropriate changes of type annotations in the implementation (the split between general `Jkind.t` and the existing `Jkind.Type.t`). However, no new types are actually introduced, and `Jkind` _is_ `Jkind.Type`.

The interface file `jkind.mli` was moved over directly from the implementation of this part, so some interface changes might not be obvious. Comments/docs should have been moved over.
Please note that GitHub's diff on `jkind.ml` is broken (and likely also on `jkind.mli`), as indentation was added for the new module.

In review, particular attention should be kept to:

- Whether coercions (`Jkind.to_type_jkind`, assert this is not an arrow jkind) are safe. Promotions `Jkind.of_type_jkind` should be generally sensible.
- Whether there aren't any places missing where we can assert things should be a `type_jkind` (`Jkind.Type.t`) rather than a general `jkind` (`Jkind.t`).

The intent of the creation of this quasi-PR was to separate out all the path changes (`Jkind -> Jkind.Type`) and type annotation changes (especially in `types.ml`). But it should be noted coercions introduced here (`to_type_jkind`) become significant later on as assertions.


ocaml/typing/jkind_types.mli
```diff
@@ -41,108 +41,124 @@
...
+  end
+
+  type 'type_expr annotation = 'type_expr Const.t * Jane_syntax.Jkind.annotation
+end
 
-type 'type_expr t =
-  { jkind : 'type_expr Jkind_desc.t;
-    history : 'type_expr history;
-    has_warned : bool
+type 'a arrow =
```
goldfirere, line 155: This surprises me. I think it's probably better wrapped in a module like other types. And why make it polymorphic this way?

ocaml/typing/jkind.mli
```diff
@@ -41,267 +41,417 @@
...
    https://github.com/goldfirere/flambda-backend/commit/d802597fbdaaa850e1ed9209a1305c5dcdf71e17
    first, which was reisenberg's attempt to do so. *)
-module Externality : sig
-  type t = Jkind_types.Externality.t =
-    | External (* not managed by the garbage collector *)
-    | External64 (* not managed by the garbage collector on 64-bit systems *)
-    | Internal (* managed by the garbage collector *)
 
-  val le : t -> t -> bool
+exception Unexpected_higher_jkind of string
```
goldfirere, line 45: Does this really need to be exported? That makes me a little unhappy about the interface here. Exporting this subtly suggests that any arbitrary function in this module might throw this exception.

ocaml/typing/jkind.mli
goldfirere, line 1: This file has some reordering. Is that necessary? It's seeing a lot of changes these days, and reordering will cause conflicts.

ocaml/typing/types.mli
```diff
@@ -219,14 +219,15 @@ and abbrev_memo =
...
 
-(* jkind depends on types defined in this file, but Jkind.equal is required
+(* jkind depends on types defined in this file, but Jkind.Type.equal is required
    here. When jkind.ml is loaded, it calls set_jkind_equal to fill a ref to the
    function. *)
 (** INTERNAL USE ONLY
-    jkind.ml should call this with the definition of Jkind.equal *)
-val set_jkind_equal : (jkind -> jkind -> bool) -> unit
+    jkind.ml should call this with the definition of Jkind.Type.equal *)
+val set_jkind_equal : (type_jkind -> type_jkind -> bool) -> unit
```
goldfirere, line 230: Does this work over all jkinds or just type_jkinds? The name of the function and its type disagree.

goldfirere (commented):
I've reviewed .mli files only.

ocaml/typing/jkind_types.mli
jbachurski: Sure, I can wrap it in a module.
It's polymorphic so it can be reused for both `Jkind` and `Jkind.Const` - it got placed here accidentally, I will revert it here and use the module wrap in a further PR.

ocaml/typing/jkind.mli
jbachurski: That's true, it doesn't have to be. It should only be raised by `to_type_jkind` in practice, and is not expected to be caught. I could be using `assert false` but I used this to make bugs more obvious to track down. Also, exporting right now is entirely unnecessary.

ocaml/typing/jkind.mli
jbachurski: I will take a look and try to avoid it where possible.

ocaml/typing/types.mli
jbachurski: In this case it only gets used by  `Types.equal_record_representation` and `Types.equal_variant_representation`, so it should be for the better if this gets renamed to `set_type_jkind_equal` with same types.

ocaml/typing/jkind.mli
jbachurski: I've tried to clean this up now.

ocaml/typing/jkind.mli
```diff
@@ -31,277 +31,422 @@
...
+  val format_history :
+    intro:(Format.formatter -> unit) -> Format.formatter -> t -> unit
+
+  (*********************************)
+  (* debugging *)
+
+  module Debug_printers : sig
+    val t : Format.formatter -> t -> unit
+  end
+end
```
jbachurski, line 336: Above this point, `module Type` exports existing interfaces, with some which are unused ones removed.  The visible diff includes the indentation.
Below, the interface for general `Jkind` is given.
I'd recommend using a diff viewer which does a better job with indentation than the one here.

ocaml/typing/jkind.mli
```diff
@@ -31,277 +31,422 @@
...
+
+  val update_reason : t -> Type.History.creation_reason -> t
 end
 
 (******************************)
 (* constants *)
+(* See documentation of relevant functions in Jkind.Type for details
+   - the functions here wrap them *)
 
 module Const : sig
```
goldfirere, line 360: It seems unnecessary to re-export things this way. But perhaps the problem is that the interface to constants is not really so great? Probably best to get rid of these functions from within the `Const` module within `Type`.

ocaml/typing/jkind.mli
```diff
@@ -406,6 +526,49 @@ val format_history :
     this module. *)
 val set_printtyp_path : (Format.formatter -> Path.t -> unit) -> unit
 
+(******************************)
+(* errors *)
+
+module Violation : sig
```
goldfirere, line 532: I don't think we want both a `Type.Violation` and a `Violation`. Just keep this outer one.

ocaml/typing/typedecl.ml
```diff
@@ -782,7 +782,8 @@ let transl_declaration env sdecl (id, uid) =
       let cty = transl_simple_type ~new_var_jkind:Any env ~closed:no_row Mode.Alloc.Const.legacy sty in
       Some cty, Some cty.ctyp_type
   in
-  let any = Jkind.Primitive.any ~why:Initial_typedecl_env in
+  let any = Jkind.Type.Primitive.any ~why:Initial_typedecl_env in
+  let any' = Jkind.Primitive.any ~why:Initial_typedecl_env in
```
goldfirere, line 786: What's going on here? This seems fishy. If it's necessary, please add an explanatory comment.

ocaml/typing/typedecl.ml
```diff
@@ -2680,9 +2681,11 @@ let make_native_repr env core_type ty ~global_repr ~is_layout_poly ~why =
        - this isn't a tvar from an outer scopes ([TyVarEnv] gets reset before
          transl)
     *)
+    (* FIXME jbachurski: Is this the way this unwrapping should happen?
```
goldfirere, line 2684: Say more? The code near here does not seem to be edited, so I'm not sure what this comment is really about.

ocaml/typing/typedecl.ml
```diff
@@ -3184,7 +3187,7 @@ let transl_package_constraint ~loc ty =
 let abstract_type_decl ~injective ~jkind ~jkind_annotation ~params =
   let arity = List.length params in
   Ctype.with_local_level ~post:generalize_decl begin fun () ->
-    let params = List.map Ctype.newvar params in
+    let params = List.map (fun jkind -> Ctype.newvar jkind) params in
```
goldfirere, line 3190: Why eta-expand here?

goldfirere (commented):
Full review this time.

.vscode/settings.json
```diff
@@ -1,12 +1,12 @@
 {
         "ocaml.sandbox": {
                 "kind": "opam",
-                "switch": "4.14.1"
+                "switch": "flambda-backend"
```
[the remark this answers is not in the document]
jbachurski: This is my bad, but it is surprising how `opam` (and hence the VS Code extension) handles this. It essentially forces a generic switch name (`4.14.1`) for everyone working in the project.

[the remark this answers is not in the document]
riaqn: I think it's better if we remove and ignore this file.

ocaml/typing/jkind_types.mli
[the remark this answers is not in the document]
jbachurski: There was some discussion here - when discussing this with @lpw25 in the first week, the type definition for `Jkind.t` became as follows:
```ocaml
type 'type_expr t =
  | Type of 'type_expr Type.t
  | Arrow of 'type_expr t Arrow.t
```
Notice that this does not keep track of history like it does for `Jkind.Type.t`:
```ocaml
  module Jkind_desc : sig
    type 'type_expr t =
      { layout : Layout.t;
        modes_upper_bounds : Modes.t;
        externality_upper_bound : Externality.t
      }
  end

  type 'type_expr history =
    | Interact of
        { reason : Jkind_intf.History.interact_reason;
          lhs_jkind : 'type_expr Jkind_desc.t;
          lhs_history : 'type_expr history;
          rhs_jkind : 'type_expr Jkind_desc.t;
          rhs_history : 'type_expr history
        }
    | Creation of Jkind_intf.History.creation_reason

  type 'type_expr t =
    { jkind : 'type_expr Jkind_desc.t;
      history : 'type_expr history;
      has_warned : bool
    }
```

I remember coming across some example that was worth bringing up, but perhaps I confused histories with violations. We can discuss this, but in the worst case it is easier if history tracking is implemented later.

ocaml/typing/typedecl.ml
jbachurski: Accidental - it says `Tvar { jkind = Type jkind }` in the implementation, but I later found this was okay, as this pattern match only occurs in a context where this has to be a type jkind.

ocaml/typing/typedecl.ml
jbachurski: As far as I see, the existing `any` was just used as a short-hand, and the alternative would be to add a promotion at each use site. This is related to whether there should be `Jkind.Primitive`s in the first place or just `Jkind.Type.Primitive`s.

ocaml/typing/jkind.mli
jbachurski: Well, I'm not sure - essentially these just export a short-hand of the `Jkind.Type` version. Do you think it'd be better for accessors to always include an explicit promotion, like `Jkind.of_type_jkind Jkind.Type.Primitive.any`?

ocaml/typing/jkind_types.mli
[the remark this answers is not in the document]
jbachurski: We agreed this change should happen - but I think I'd rather delay it for now as it's not too important. 
In particular it seems there'd still need to be a `Jkind.Type` version of a with-history type (say we have `Jkind.Type.t = Jkind.Type.Jkind_desc.t with_history` and `Jkind.t = Jkind.Jkind_desc.t with_history`), because there are still places in the code where we want to just have `Jkind.Type.t` instead of a general `Jkind.t`, and that still needs history tracking.

ocaml/typing/typedecl.ml
goldfirere: What do you mean by "add a promotion"? Oh... I think you mean a conversion to `Jkind.t`. I think we'll need both `any` (which will be the top of the type-kind lattice) and `top` (which will be the top of the kind lattice).

ocaml/typing/jkind_types.mli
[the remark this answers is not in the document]
goldfirere: Maybe. That's a hard argument to refute. But my guess is that, in the end, we will only want histories on proper kinds, not on type-kinds.

Though I suppose not making this change decreases the amount of diff, so it might be OK to wait here. My hunch is that making the change now will be easier in the end, but I can't be sure.

ocaml/typing/jkind.mli
goldfirere: `any` is a special case (because it needs to be split between `any` and `top`). But for the others: yes, I think I'd want an explicit conversion. That will clarify what's going on.

ocaml/typing/typedecl.ml
jbachurski: Yes, we also agreed with Leo `top` would be necessary (from my perspective it seemed the compiler creates too many type variables with the `any` kind where it should be `top`, or generally just assumes things are `any` rather than `top`).

jbachurski: I made an attempt at moving to `top` where appropriate. There are many such places, especially in `ctype`, where this seems necessary.

ocaml/typing/jkind_types.mli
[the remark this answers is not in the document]
jbachurski: I think it's hardly debatable that we will want histories on kinds, so this is something that we want.
But I'd also prefer to delay implementing this for now. I don't think this will make things too difficult, as it seems most of the code unpacking jkinds has been written, and the rest will be interfaces.

ocaml/typing/ctype.ml
```diff
@@ -4173,8 +4173,8 @@ exception Filter_method_failed of filter_method_failure
 (* Used by [filter_method]. *)
 let rec filter_method_field env name ty =
   let method_type ~level =
-      let ty1 = newvar2 level (Jkind.Primitive.value ~why:Object_field) in
-      let ty2 = newvar2 level (Jkind.Primitive.value ~why:Row_variable) in
+      let ty1 = newvar2 level (Jkind.Type.Primitive.value ~why:Object_field |> Jkind.of_type_jkind) in
```
goldfirere, line 4176: `new_type_var2` would be nice to avoid this conversion

ocaml/typing/ctype.ml
```diff
@@ -5719,9 +5719,9 @@ let rec build_subtype env (visited : transient_expr list)
           if deep_occur_list ty tl1 then raise Not_found;
           set_type_desc ty
             (Tvar { name = None;
-                    jkind = Jkind.Primitive.value
+                    jkind = Jkind.Type.Primitive.value
```
goldfirere, line 5722: I'm surprised this doesn't need a conversion

goldfirere (approved):

native_toplevel/opttopdirs.ml
```diff
@@ -138,7 +138,7 @@ let match_printer_type ppf desc typename =
         raise Exit
   in
   Ctype.with_local_level ~post:Ctype.generalize (fun () ->
-    let ty_arg = Ctype.newvar (Jkind.Primitive.value ~why:Debug_printer_argument) in
+    let ty_arg = Ctype.newvar (Jkind.Type.Primitive.value ~why:Debug_printer_argument |> Jkind.of_type_jkind) in
```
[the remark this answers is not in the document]
jbachurski: Might that be a bit confusing since `newvar` really also creates a type variable? Or perhaps a type expression variable, now? Otherwise agreed.

ocaml/debugger4/loadprinter.ml
```diff
@@ -93,7 +93,7 @@ let eval_value_path env path =
 
 let match_printer_type desc make_printer_type =
   Ctype.with_local_level ~post:Ctype.generalize begin fun () ->
-    let ty_arg = Ctype.newvar Jkind.Primitive.(value ~why:Debug_printer_argument) in
+    let ty_arg = Ctype.newvar Jkind.Type.Primitive.(value ~why:Debug_printer_argument) in
```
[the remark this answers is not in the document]
jbachurski: Yes, sorry, similar to the latter example this is an oversight. This type shuffling does not actually abstract over `Jkind.t` vs `Jkind.Type.t`, so this doesn't error, and I only fixed it in a future branch. That's my bad.

ocaml/lambda/translmod.ml
```diff
@@ -300,7 +300,7 @@ let init_shape id modl =
             Tarrow(_,ty_arg,_,_) -> begin
               (* CR layouts: We should allow any representable layout here. It
                  will require reworking [camlinternalMod.init_mod]. *)
-              let jkind = Jkind.Primitive.value ~why:Recmod_fun_arg in
+              let jkind = Jkind.Type.Primitive.value ~why:Recmod_fun_arg |> Jkind.of_type_jkind in
               let ty_arg = Ctype.correct_levels ty_arg in
               match Ctype.check_type_jkind env ty_arg jkind with
```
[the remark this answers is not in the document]
jbachurski: Perhaps, if you think that'd be clearer.

jbachurski:
Thanks for the final reviews @goldfirere !
I won't be around for the rest of today but I'll address your final comments on Monday so we can hopefully get this merged then.

ocaml/typing/ctype.ml
jbachurski: Same as above, oversight due to `Jkind.Type.t = Jkind.t`.

native_toplevel/opttopdirs.ml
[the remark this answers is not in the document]
goldfirere: We could just say that `newvar` creates a "Tvar". I agree this is not great. But I think having a `new_type_var` would still be an improvement.

ocaml/lambda/translmod.ml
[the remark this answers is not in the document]
goldfirere: I say: Things that have kinds are type expressions. Some of these are additionally types. Does that cover it?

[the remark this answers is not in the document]
jbachurski: I added `new_type_var` and `new_type_var2`. I'm not completely sold on changing the interface of `check_type_jkind` and `constraint_type_jkind` - is `tyexpr` a good name? That specific abbreviation does not seem to be in use.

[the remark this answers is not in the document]
goldfirere: I'm also not as convinced about the need for a change here. Keep it as is, if you like, or change later.

ocaml/debugger4/loadprinter.ml
```diff
@@ -93,7 +93,11 @@ let eval_value_path env path =
 
 let match_printer_type desc make_printer_type =
   Ctype.with_local_level ~post:Ctype.generalize begin fun () ->
-    let ty_arg = Ctype.newvar Jkind.Type.Primitive.(value ~why:Debug_printer_argument) in
+    let ty_arg =
+      Ctype.newvar (
```
goldfirere: What about `new_type_var`?

ocaml/typing/jkind_intf.ml
```diff
@@ -253,6 +253,14 @@ module History = struct
 
   type bits64_creation_reason = Primitive of Ident.t
 
+  type top_creation_reason =
```
goldfirere, line 256: Now that these are reasons for `top`, do any stop being used for `any`? I would expect `Missing_cmi` and `Dummy_jkind` at least to stop being used for `any`.

ocaml/debugger4/loadprinter.ml
goldfirere: Changed in later commit

ocaml/typing/jkind_intf.ml
jbachurski: I'd rather not make any changes like this, as I found that filling in `top`s by my judgment doesn't seem to have worked out well in practice, and I sometimes need to move it back to using `any`. Let's do removals like that last.

goldfirere:
OK. Merge when ready.

ocaml/typing/jkind.mli
lpw25: I just suggested to put these back in the next PR. I think that e.g. `Jkind.Primitive.value ~reasion` is noticeably clearer than `Jkind.of_type_jkind (Jkind.Type.Primitive.value ~reason)` and you end up getting this in quite a few places. @goldfirere do you object to putting them back?

goldfirere: Fine. I can accept that I was perhaps overzealous in avoiding repetition here. No objection.

PR #2778, opened 2024-07-15
Allow modalities on primitives

riaqn:
This PR allows modalities on primitives.

A proper design would be:
```
module M : sig
    external length : string -> int @@ portable = "%string_length"
end = struct
    external (length @ portable) : string -> int = "%string_length"
end
```
where the first `portable` is modality, while the second is mode. However, this would involve changing the parsetree (and hence our internal code base), which slows down other projects.

Instead, this PR allows:
```
module M : sig
    external length : string -> int @@ portable = "%string_length"
end = struct
    external length : string -> int @@ portable = "%string_length"
end
```
where both `portable` are modalities. As modules are now fixed to be legacy, the outcome of applying the modalites on the module's mode is predictable to the user. 

We must adopt the proper design above before we allow modules to be non-legacy.

### DIFF
--- a/ocaml/testsuite/tests/typing-modes/modes.ml
+++ b/ocaml/testsuite/tests/typing-modes/modes.ml
@@ -385,18 +385,6 @@ let foo () =
 val foo : unit -> unit = <fun>
 |}]
 
-(* modalities on primitives are parsed but not supported yet. *)
-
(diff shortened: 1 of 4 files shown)


lukemaurer (approved):
LGTM

PR #2779, opened 2024-07-15
Implement basic typing for module type inference with an underscore

smjleo:
This PR allows inference of module types in the case where the module type is already defined with in the signature of the corresponding struct, i.e.
```ocaml
module M = sig
  module type S = sig ... end
end = struct
  module type S = _
end
```

Currently, it can only infer the module type in the case where the `sig` and `struct` are defined together using `Pmod_constraint`. The plan is to support the case where the `sig` and `struct` are defined in corresponding `.mli` and `ml` files, respectively, in a later PR.

Also note the last test case in `tests/module-type-underscore/typing.ml` - the type checker currently gives a confusing error message in the case where the module type being inferred references a type/module/module type present in the signature but not yet present in the struct. This will also be fixed in a future PR.


ocaml/typing/typemod.ml
```diff
@@ -100,10 +100,25 @@ type error =
...
+    - [From_signature]: we are checking a signature.
+    - [No_expected]: we are checking a struct, but we don't have information about
+       its expected signature.
+    - [Expected _]: we are checking a struct, and we do have information about its
+       expected signature (where "expected signature" is used for inferring module types
+       when we see a [module type S = _]). *)
+type expected_modtype_decl =
+  | From_signature
+  | No_expected
+  | Expected of modtype_declaration
```
riaqn, line 121: Might be clearer to refactoring this into:
```
type modtype_decl_context =
   | In_signature
   | In_structure of modtype_declaration option
```
Because in the case of `In_signature` it's not much of an "expected modtype_decl".

riaqn: Also a minor formatting nit: might be clearer if you put the comments next to each constructor. (with `(** *)` which will be recognized as doc generator)

ocaml/typing/typemod.ml
```diff
@@ -1991,20 +2007,34 @@ and transl_signature env (sg : Parsetree.signature) =
...
     )
 
-and transl_modtype_decl env pmtd =
+and transl_modtype_decl ~expected_modtype_decl env pmtd =
   Builtin_attributes.warning_scope pmtd.pmtd_attributes
-    (fun () -> transl_modtype_decl_aux env pmtd)
+    (fun () -> transl_modtype_decl_aux env pmtd ~expected_modtype_decl)
 
-and transl_modtype_decl_aux env
+and transl_modtype_decl_aux ~expected_modtype_decl env
```
riaqn, line 2015: With the above suggested change, this should be renamed to `~context` or similar.

riaqn, line 2015: With the above suggested change, this should be renamed to `~context` or similar.

ocaml/typing/typemod.ml
```diff
@@ -2519,8 +2549,20 @@ and type_module_aux ~alias sttn funct_body anchor env smod =
...
+      let expected_sig = match expected_modtype with
+        | None -> None
+        | Some mty ->
+          (match mty with
+          | Mty_ident _
+          | Mty_functor _
+          | Mty_alias _
+          | Mty_strengthen _ -> None
+          (* For now, we only allow module type inference in cases where the signature
+             is "close" to the struct (i.e. declared in the same place). *)
```
riaqn, line 2562: For things that we need to improve in the near future, let's add a CR comment block.

riaqn:
Half done - will continue tomorrow.

ocaml/typing/typemod.ml
```diff
@@ -3221,38 +3348,40 @@ and type_structure ?(toplevel = None) funct_body anchor env sstr =
                 ])
              classes []),
         shape_map,
-        new_env
+        new_env,
+        subst
     | Pstr_include sincl ->
-        type_str_include ~functor_:false ~loc env shape_map sincl sig_acc
+        let (tstr, sg, map, newenv) = type_str_include ~functor_:false ~loc env shape_map sincl sig_acc in
+        tstr, sg, map, newenv, subst
```
riaqn, line 3463: I think we will support `include` for substitution - if it's too much work, let's add a CR.

riaqn (commented):
Reviewed all except tests.

ocaml/typing/typemod.ml
riaqn: This is the special case, in that this can be called in both signatures and structures. The word "expected" in the type checker has a clear meaning which doesn't include the case when it's called in signatures.

riaqn:
Also as discussed, could you force push to this PR, so that this PR contains the bare minimum typing (blanket substitution) with tests showing wrong outcome (annotated with CR)?

ocaml/testsuite/tests/module-type-underscore/typing.ml
```diff
@@ -824,3 +888,6 @@ end
 
 [%%expect {||}]
 *)
+
+[%%expect{|
+|}]
```
riaqn, line 893: Extra bits?

riaqn (approved):
Looks good!

PR #2780, opened 2024-07-15
Remove Jane Syntax for n-ary functions

ncik-roberts:
Start using a direct encoding of n-ary functions in the parsetree. Delete the Jane Syntax encoding for n-ary functions.

We've changed the function AST beyond the making it n-ary. We also track mode and jkind annotations. So, this PR is not a straightforward backport; we need to tack on our extensions to the code from upstream.

There are a few broad classes of diff in this feature:
  * Straightforward backport. Example: `ast_helper.ml`, `ast_invariants.ml`. Here, we just take the code from upstream.
  * Half-backport, half&ndash;adding our extensions. Example: `printast.ml`. Here, we can mostly take the code from upstream. In some cases, I move some code around in a file (e.g. `ast_mapper.ml`) to get it to match more closely where it lives upstream and thus to reduce drift.
  * Removal of coercion `expr : Mode_expr.Const.t :> _ Location.loc` (and same for jkinds). That's because the parsetree type is no longer private (the parsetree is an mli-only file). I could have minted a new file to keep these types private, but I judged it to not be worth the trouble. The private-ness here was just about making it less likely to mix up strings &mdash; it doesn't guard any invariants.
  * Small changes to the locations in tests. Example: `ocaml/testsuite/tests/typing-gadts/pr7269.ml` (this one matches upstream). Another example: `ocaml/testsuite/tests/parsetree/modes_ast_mapper.reference` (this one is just fine; we've changed the location of the mode annotation in a way that seems more accurate anyway).

When I say that a code is a "backport", I encourage you to compare against the code in https://github.com/ocaml/ocaml.

**Please let @ncik-roberts merge this!** This implies some work in ppxes, and I'd like to be sure we're in a good spot to do that work before committing ourselves to doing it.

**Reviewer:** We're discussing internally.


ocaml/parsing/parser.mly
```diff
@@ -768,31 +767,29 @@ let mkghost_newtype_function_body newtypes body_constraint body ~loc =
...
       match all_params_as_newtypes params with
-      | None -> n_ary_function (params, body_constraint, body) ~loc ~attrs
+      | None ->
+          mkexp_attrs (Pexp_function (params, body_constraint, body)) attrs ~loc
       | Some newtypes ->
           wrap_exp_attrs
             ~loc
             (mkghost_newtype_function_body newtypes body_constraint body_exp
-               ~loc)
+                ~loc)
```
freemagma, line 792: Why the space here?

ocaml/parsing/pprintast.ml
```diff
@@ -864,15 +862,30 @@ and expression ?(jane_syntax_parens = false) ctxt f x =
...
     | Pexp_newtype (lid, e) ->
         pp f "@[<2>fun@;(type@;%s)@;%a@]" lid.txt
           (pp_print_pexp_function ctxt "->") e
-    | Pexp_function l ->
-        pp f "@[<hv>function%a@]" (case_list ctxt) l
+    | Pexp_function (params, constraint_, body) ->
+        begin match params, constraint_ with
+          (* Omit [fun] if there are no params. *)
+          | [], None ->
+              let should_paren =
```
freemagma, line 872: Looks like upstream has another comment here that you've omitted

freemagma (approved):

ocaml/parsing/parser.mly
ncik-roberts: It was unintentional. This code got moved around a lot as I was working on the feature, even though the end state is pretty close to the original `parser.mly`. Now resolved.

ocaml/parsing/pprintast.ml
ncik-roberts: Yep, thanks.

ncik-roberts:
I've deleted the "build Jane Syntax with upstream" CI check &mdash; I added this originally for a reason that no longer applies, now that we're modifying the parsetree.

ncik-roberts:
@freemagma The last commit probably should be reviewed (I noticed the issue in testing)

ocaml/parsing/pprintast.ml
```diff
@@ -1575,7 +1575,7 @@ and payload ctxt f = function
       pp f "?@ "; pattern ctxt f x;
       pp f " when "; expression ctxt f e
 
-and pp_print_pexp_function ctxt sep f x =
+and pp_print_pexp_newtype ctxt sep f x =
```
freemagma, line 1578: To make sure I'm understanding the correctly: I assume there was some sort of jane-syntax related issue when an `x` with description `Pexp_function` is passed here? What exactly was the issue?

ncik-roberts: oh, sorry, thought I wrote this down. the issue is in ppxlib's round trip check: it reparses the output of pprintast and complains if it differs from the original parsetree. It's nice for `Pexp_newtype (a, Pexp_function (...))` to be printed differently than `Pexp_function ([ Pparam_newtype a; ... ], ...)` for that reason.

PR #2781, opened 2024-07-15
Don't infer tail-calls for applications where the called function contains a dot

tylerhou:
Suppose a function `f` allocates locals, and `f` calls `g` in tail
position. If `g` is inferred to be a tail-call, it cannot accept (as
parameters) locals that `f` had allocated as `f`'s region does not fully
enclose `g`'s region.

There are many places where function calls happen to be in tail
position, but do not need to be tail calls. These function calls
currently cannot accept locals due to the above restriction. That
restriction is annoying. E.g. the following function does not compile
(before this change), but we would like it to:

```ocaml
module rec List : sig
  val map : local_ ('a -> 'b) -> 'a list -> 'b list
end = List

let call_local_map_in_tail_pos lst =
  let local_ str = "Hello, world!" in
  let local_ f x = let _ = str in () in
  List.map f lst
```

Error:

```
Line 8, characters 11-12:
8 |   List.map f lst
               ^
Error: This value escapes its region.
  Hint: This argument cannot be local,
  because it is an argument in a tail call.
```

By inferring fewer tail calls, we can make that restriction less
annoying. Our first change is to (heuristically) not infer tail calls
for applications of functions that contain a dot (`.`) in the
identifier, as they are unlikely to require tail call optimization. We
are in the process of developing a better heuristic.


ocaml/typing/typecore.ml
```diff
@@ -5959,7 +5973,7 @@ and type_expect_
       }
   | Pexp_send (e, {txt=met}) ->
       submode ~loc ~env Mode.Value.legacy expected_mode;
-      let pm = position_and_mode env expected_mode sexp in
+      let pm = position_and_mode env expected_mode sexp ~maybe_ident:None in
```
dvulakh, line 5976: I agree we're not worried about changing how locals interact with method dispatch and constructors, so we can keep the default behavior of always producing tail calls. I wonder if we should make it more explicit that's what we're doing, especially since it might become easier to accidentally change the behavior here when we start using more complicated heuristics.

ocaml/testsuite/tests/backtrace/names.ml
```diff
@@ -106,7 +106,9 @@ let inline_object f =
 
 let[@inline never] lazy_ f =
   let x = Sys.opaque_identity (lazy (1 + f ())) in
-  Lazy.force x
+  Lazy.force x [@tail]
+  (* CR less-tco: Force this to be TCO'd so that it does not show up in the stacktrace.
```
dvulakh, line 110: Usually when a CR describes an action, that action is something we want to do but haven't. I'd rephrase the background about the problem and how we fix it to be descriptive, and then add something about returning to the test to see if there's a nicer way to make the outputs consistent.

dvulakh (changes requested):
Thanks for putting this up; it looks good overall. I've left a few comments.

ocaml/testsuite/tests/typing-tailcalls/basic.ml
```diff
@@ -0,0 +1,19 @@
+(* TEST
+ flags = "-dtypedtree -dsel";
```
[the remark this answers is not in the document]
mshinwell: Please can we try to clean the output even further, so the test doesn't require promoting upon minor changes that cause identifier stamps etc to change?

ocaml/testsuite/tests/backtrace/names.ml
tylerhou: Done.

ocaml/typing/typecore.ml
tylerhou: Ack; improved in next feature.

ocaml/testsuite/tests/typing-tailcalls/basic.ml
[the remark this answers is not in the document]
tylerhou: Done.

ocaml/testsuite/tests/typing-tailcalls/locals.ml
```diff
@@ -0,0 +1,68 @@
...
+3 |   M.f str [@tail]
+          ^^^
+Error: This value escapes its region.
+  Hint: This argument cannot be local,
+  because it is an argument in a tail call.
+|}]
+
+let implicit_tail () =
+  let local_ str = "hello" in
+  f str
```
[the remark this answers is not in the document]
tylerhou: Done.

ocaml/testsuite/tests/backtrace/names.ml
```diff
@@ -106,8 +106,16 @@ let inline_object f =
...
-  Lazy.force x
-
+  Lazy.force x [@tail]
+  (* CR less-tco: Somehow make the test output consistent between the native
+     and bytecode compilers without using the above [@tail] annotation.
+
+     As a result of the less-tco project, if there were no [@tail] annotation
+     above, the call to Lazy.force would not be TCO'd. For some reason,
+     the stacktrace without TCO differs between the native and bytecode compiler
+     (the bytecode compiler reports an unknown location for one stack frome).
```
dvulakh, line 116: s/frome/frame

dvulakh (changes requested):

ocaml/testsuite/tests/typing-tailcalls/basic.ml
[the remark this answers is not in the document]
tylerhou: > Please can we try to clean the output even further, so the test doesn't require promoting upon minor changes that cause identifier stamps etc to change?

Capturing `-dlambda` instead of `-dsel` now.

dvulakh (approved):

ocaml/testsuite/tests/typing-tailcalls/basic.reference
dvulakh, line 1: I believe we can delete this file.

dvulakh (changes requested):

ocaml/testsuite/tests/typing-tailcalls/basic.reference
tylerhou: Done.

dvulakh (approved):

dvulakh (approved):
Looks good, thanks! Feel free to merge.

Issue #2123, opened 2023-12-05
Failure of mutable unboxing in conjunction with a for-loop


mshinwell:

Gbury:
Actually, this issue is not completely solved by #2132 . We still need to upgrade the flow analysis to handle dominators which are symbols and constants.

Gbury:
For reference (because I spent some time looking for it) here is a problematic example where unboxing does not work as expected in the presence of a loop.

```ocaml
[@@@ocaml.flambda_o3]

type 'a t = {
  mutable x : unit;
  get : int -> local_ 'a
}

let f j =
  let local_ t = { x = (); get = (fun x -> x) } in
  for i = 1 to j do
    Sys.opaque_identity ((t.get [@inlined hint]) i : int)
  done
```

PR #2782, opened 2024-07-15
Allow Simple dominators in Flow Analaysis (instead of just variables)

Gbury:
This PR adds the ability for the data flow analysis to discover any Simple as a dominator, whereas previously only variables could be detected as dominators. In particular, that means that we can now detect (and add aliases) when a variable is aliased/dominated by a symbol or a constant.

Whenever we add such an alias to a constant, or a symbol (with a known type), we trigger a resimplification, since that opens up new opportunities for `Simplify `. This is in particular necessary for the example below to work.

This (finally) solves #2123 , so that in the following example:
```ocaml
[@@@ocaml.flambda_o3]

type 'a t = {
  mutable x : unit;
  get : int -> local_ 'a
}

let f j =
  let local_ t = { x = (); get = (fun x -> x) } in
  for i = 1 to j do
    Sys.opaque_identity ((t.get [@inlined hint]) i : int)
  done
```
The `t.get` function is indeed inlined inside of the loop. Unfortunately, this required three rounds of simplification (the first one does the mutable unboxing, the second one adds the alias to the symbol for the lifted function, and finally the third one can do the inlining). In the future, we could do it with only two passes if we ran a second alias analysis after the mutable unboxing (but that will require quite a bit of work).

It would be good if we could try and benchmark compilation times with this PR to checks whether the new code triggers too many resimplifications.


middle_end/flambda2/simplify/flow/control_flow_graph.ml
```diff
@@ -162,7 +170,7 @@ let fixpoint t ~init ~f =
   res
 
 let extra_args_for_aliases_overapproximation ~required_names
-    ~(source_info : T.Acc.t) ~unboxed_blocks doms t =
+    ~(source_info : T.Acc.t) ~unboxed_blocks (doms : doms) t =
```
chambart, line 173:
```suggestion
    ~(source_info : T.Acc.t) ~unboxed_blocks (doms : Dominator_graph.alias_map) t =
```

middle_end/flambda2/simplify/flow/control_flow_graph.ml
```diff
@@ -15,13 +15,23 @@
 module T = Flow_types
 module G = Strongly_connected_components.Make (Continuation)
 
+type doms = Dominator_graph.alias_map
+
```
chambart, line 19:
```suggestion
```
This is used only once, there is no point in declaring it

middle_end/flambda2/simplify/flow/dominator_graph.ml
```diff
@@ -13,66 +13,77 @@
...
+  in
+  { t with graph }
+
+let add_node_name t simple name =
+  if Name.Set.mem name t.required_names then add_node_aux t simple else t
+
+let add_node t simple =
+  Simple.pattern_match simple
+    ~name:(fun name ~coercion:_ -> add_node_name t simple name)
+    ~const:(fun _ -> add_node_aux t simple)
```
chambart, line 51:
```suggestion
let add_node t simple =
  Simple.pattern_match simple
    ~name:(fun name ~coercion:_ ->
      if Name.Set.mem name t.required_names then add_node_aux t simple else t)
    ~const:(fun _ -> add_node_aux t simple)
```
Inlined is just simpler

middle_end/flambda2/tests/mlexamples/tests3.flt
```diff
@@ -126,15 +126,18 @@ let code loopify(never) size(65) newer_version_of(foo_0)
...
-         where k2 =
+           cont k2
+         where k3 =
            let prim = %unbox_float apply_result in
            let ifso_result = %array_set float arr.(i) <- prim in
-           cont k (0))
+           cont k2
+         where k2 =
+           let if_then_else_result = 0 in
+           cont k (if_then_else_result))
```
chambart, line 140: It does not change much, but is that expected ?

Gbury: From what I recall, it is expected, because `0` can now be discovered as a dominator, thus we insert the let_binding at the start of the continuation.

Gbury: but you might need to look at the rawflambda to really see/understand what is happening

middle_end/flambda2/simplify/flow/mutable_unboxing.ml
```diff
@@ -682,8 +688,8 @@ let add_to_extra_params_and_args result =
   epa
 
 let did_unbox_a_mutable_block result =
-  Variable.Map.exists
-    (fun _var block_to_unbox ->
+  Simple.Map.exists
+    (fun _sym block_to_unbox ->
```
chambart, line 692:
```suggestion
    (fun _simple block_to_unbox ->
```

middle_end/flambda2/simplify/flow/flow_analysis.ml
```diff
@@ -172,3 +172,44 @@ let did_perform_mutable_unboxing (result : T.Flow_result.t) =
...
        result.mutable_unboxing_result.let_rewrites
+
+let added_useful_alias_in_loop typing_env (acc : T.Acc.t)
+    (result : T.Flow_result.t) =
+  (* In some cases, the alias analyis can find invariant parameters that are
+     aliased to a symbol, and which were not detected during the downwards pass,
+     for instance when the parameters are involved in a loop.
+
+     In such cases, it is pertinent to trigger a resimplification of the current
+     expression (i.e. function body), so that we can rpopagate the information
```
chambart, line 183:
```suggestion
     expression (i.e. function body), so that we can propagate the information
```

middle_end/flambda2/simplify/flow/flow_analysis.ml
```diff
@@ -172,3 +172,44 @@ let did_perform_mutable_unboxing (result : T.Flow_result.t) =
...
+     from that new alias.
+
+     In order to avoid triggering too many resimplifications for no reasons, we
+     restrict this process to trigger only when a (non-invariable) parameter of
```
chambart, line 187:
```suggestion
     restrict this process to trigger only when a (non-invariant) parameter of
```

middle_end/flambda2/simplify/simplify_expr.ml
```diff
@@ -78,6 +78,12 @@ let simplify_toplevel_common dacc simplify ~params ~implicit_params
           then UA.set_resimplify uacc
           else uacc
         in
+        let uacc =
+          if Flow.Analysis.added_useful_alias_in_loop (DA.typing_env dacc)
+               data_flow flow_result
+          then UA.set_resimplify uacc
+          else uacc
+        in
```
chambart, line 86: Wouldn't it be nicer to merge that if with the previous one ?

chambart (approved):
LGTM, only some wondering about the resimplification criteria

Ekdohibs:
I'm wondering how safe this is w.r.t. mutable unboxing: if a mutable block ends up in a symbol, it must be marked as potentially escaping since the block could be modified by directly accessing its symbol.

Gbury:
> I'm wondering how safe this is w.r.t. mutable unboxing: if a mutable block ends up in a symbol, it must be marked as potentially escaping since the block could be modified by directly accessing its symbol.

I think we're safe because of the following reasons:
- mutable unboxing only triggers if the dominator is the mutable allocation
- symbols are lifted at top-level, so mutable unboxing involving a symbol can only trigger at top-level
- at top-level, we can effectively see all uses of the symbol, including if it's stored in the module block, so we should correctly compute whether it escapes or not

Note that if/when we try to not move symbol definitions at top-level (and instead tweak the let-binding in place to indicate it's now a symbol), this reasoning will not hold anymore.

middle_end/flambda2/simplify/flow/flow_analysis.ml
```diff
@@ -172,3 +172,44 @@ let did_perform_mutable_unboxing (result : T.Flow_result.t) =
...
+               ~var:(fun _ ~coercion:_ ->
+                 (* variables may not be bound in the typing env at this
+                    point *)
+                 false)
+               ~symbol:(fun sym ~coercion:_ ->
+                 match
+                   Typing_env.find_or_missing typing_env (Name.symbol sym)
+                 with
+                 | None -> false
+                 | Some ty -> not (Flambda2_types.is_unknown typing_env ty)))
```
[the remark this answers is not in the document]
Gbury: Indeed, however, at the point where we do that check, i.e after the term has been rebuilt, we do  not have access to a (or rather all of the) typing env(s) that would be needed to determine that.

middle_end/flambda2/simplify/flow/flow_analysis.ml
```diff
@@ -172,3 +172,44 @@ let did_perform_mutable_unboxing (result : T.Flow_result.t) =
...
+     for instance when the parameters are involved in a loop.
+
+     In such cases, it is pertinent to trigger a resimplification of the current
+     expression (i.e. function body), so that we can rpopagate the information
+     from that new alias.
+
+     In order to avoid triggering too many resimplifications for no reasons, we
+     restrict this process to trigger only when a (non-invariable) parameter of
+     a recursive continuation becomes aliased to something that has a
+     non-trivial type. *)
```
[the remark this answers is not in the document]
Gbury: I moved it to the mli

Ekdohibs (approved):

PR #2783, opened 2024-07-15
Generalise module type inference to file level

smjleo:
(This PR builds off #2779 - marking as draft until that one is merged)

Previously, the underscore syntax for module type inference only worked when they were defined together in one file:
```ocaml
module M : sig
  module type S = sig ... end
end = struct 
  module type S = _
end
```

This PR generalises this to instances where the `sig` and `struct` are `.mli` and `.ml` files, respectively.

```ocaml
(* foo.mli *)
module type S = sig ... end
```

```ocaml
(* foo.ml *)
module type S = _
```


ocaml/typing/typemod.ml
```diff
@@ -3697,15 +3697,19 @@ let type_implementation ~sourcefile outputprefix modulename initial_env ast =
...
       register_params !Clflags.parameters;
-      let (str, sg, names, shape, finalenv) =
-        Profile.record_call "infer" (fun () ->
-          type_structure ~expected_sig:None initial_env ast) in
-      let uid = Uid.of_compilation_unit_id modulename in
-      let shape = Shape.set_uid_if_none shape uid in
-      if !Clflags.binary_annotations_cms then
-        cms_register_toplevel_struct_attributes ~sourcefile ~uid ast;
-      let simple_sg = Signature_names.simplify finalenv names sg in
+      let process_structure expected_sig =
```
riaqn, line 3700: I think it might be better to call this `type_structure`, which shadows the existing `type_structure`.

riaqn, line 3700: I think it might be better to call this `type_structure`, which shadows the existing `type_structure`. This prevents the rest of the function calls the existing `type_structure`.

ocaml/testsuite/tests/module-type-underscore/file_level.ml
```diff
@@ -0,0 +1,22 @@
+(* TEST
+   setup-ocamlc.byte-build-env;
+   compiler_reference = "${test_source_directory}/file_level.reference";
+   ocamlc_byte_exit_status = "2";
+   ocamlc.byte;
+   check-ocamlc.byte-output;
+  *)
+
+module type S = _
```
riaqn, line 9: Let's add some interesting tests playing with multiple files.

- `include List`, where `List` is a file-level module. Check if it's added to substitution properly.
- `module M = List` in both `.mli` and `.ml`. See if the substitution is added properly.
- `module type S = module type of List` in `.mli`, and then `module type S = _` .

smjleo: I've added a new test with `ocamlc_byte_exit_status = "0"` and an empty `.reference`, so at least we know we aren't getting type errors. Is this sufficient, or is there a nice way to print inferred type info (like in `toplevel` tests)?

riaqn: Pass `-i` to the compiler makes it print the inferred type of `.ml` without looking at `.mli`.

riaqn (approved):

PR #2786, opened 2024-07-16
Add counters to profiler (for CFG spill and reload instructions)

milan-tom:
This PR introduces the ability to measure certain values as counters during passes of the compiler.

Options for API:
1. Existing `Profile.record` function with optional `?counter_f` argument
    - This function is used in many files so changing the signature could potentially mean implementations 
      no longer match interfaces.
2. New `Profile.record_with_counters` function
    - Shows intent better and gives hint about purpose of `couter_f` argument.
    - Allows providing specific documentation for this function.
    - New function so won't break any other files.
    - Intertwines logic for timing/memory profiling with counter profiling (these operate differently as 
      timing and memory are difference measures between the start and end of the stage and have a fixed 
      number of measures whereas counters are purely based on the result of a stage and there may be 
      many counters).
3. Completely separate function
    - No intertwining of logic for timing/memory profiling with counter profiling
    - May lead to code duplication (many aspects in common between these different types of profiling)
    - Harder to print all profile columns for `-dprofile` option (need to manage integration of the two profiling 
      mechanisms)

The current implementation uses option 2.

### DIFF
--- a/backend/asmgen.ml
+++ b/backend/asmgen.ml
@@ -464,12 +464,11 @@ let compile_phrases ~ppf_dump ps =
           if !dump_cmm then fprintf ppf_dump "%a@." Printcmm.phrase p;
           match p with
           | Cfunction fd ->
-            let compile_fundec = compile_fundecl ~ppf_dump ~funcnames in
             (* Only profile if function level granularity selected*)
-            let profiled_compile_fundec = match !profile_granularity with
-            | Function_level -> Profile.record ~accumulate:true fd.fun_name.sym_name compile_fundec
-            | File_level -> compile_fundec
-            in profiled_compile_fundec fd;
+            let profile_wrapper = match !profile_granularity with
+            | Function_level -> Profile.record ~accumulate:true fd.fun_name.sym_name
+            | File_level -> fun x -> x
+            in profile_wrapper (compile_fundecl ~ppf_dump ~funcnames) fd;
             compile ~funcnames:(String.Set.remove fd.fun_name.sym_name funcnames) ps
           | Cdata dl ->
             compile_data dl;
--- a/backend/asmgen.ml
+++ b/backend/asmgen.ml
@@ -466,7 +466,7 @@ let compile_phrases ~ppf_dump ps =
             (* Output function-level profiling if specified by user *)
             if !Clflags.profile_granularity = Profile.Function_level then
               (
-                Printf.printf "%s:\n" fd.fun_name.sym_name;
+                Printf.printf "%s\n" fd.fun_name.sym_name;
                 Profile.print Format.std_formatter !Clflags.profile_columns ~timings_precision:!Clflags.timings_precision;
                 Printf.printf "\n";
               );
--- a/backend/asmgen.ml
+++ b/backend/asmgen.ml
@@ -461,16 +461,11 @@ let compile_phrases ~ppf_dump ps =
           if !dump_cmm then fprintf ppf_dump "%a@." Printcmm.phrase p;
           match p with
           | Cfunction fd ->
-            compile_fundecl ~ppf_dump ~funcnames fd;
-
-            (* Output function-level profiling if specified by user *)
-            if !profile_granularity = Function_level then
-              (
-                Printf.printf "%s\n" fd.fun_name.sym_name;
-                Profile.print Format.std_formatter !Clflags.profile_columns ~timings_precision:!Clflags.timings_precision;
-                Printf.printf "\n";
-              );
-
+            let compile_fundec = compile_fundecl ~ppf_dump ~funcnames in
+            let profiled_compile_fundec = match !profile_granularity with
+            | Function_level -> Profile.record ~accumulate:true fd.fun_name.sym_name compile_fundec
+            | _ -> compile_fundec (* Don't profile if -dfunc-level not selected*)
+            in profiled_compile_fundec fd;
             compile ~funcnames:(String.Set.remove fd.fun_name.sym_name funcnames) ps
           | Cdata dl ->
             compile_data dl;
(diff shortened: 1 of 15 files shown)


xclerc (commented):
I think it would be worthwhile to also
amend `/ocaml/driver/compenv.ml`
(in particular the `read_one_param`
function), so that the new behavior
can also be requested through the
`OCAMLRUNPARAM` environment
variable.

xclerc:
(@gretay-js will also review)

ocaml/utils/profile.ml
```diff
@@ -16,26 +16,53 @@
...
 type file = string
+type granularity = File_level | Function_level
+
+module Int = Misc.Stdlib.Int
+module String = Misc.Stdlib.String
+
+module Counters = struct
+  type t = int String.Map.t
+
+  let get name t = String.Map.find name t
```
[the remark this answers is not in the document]
milan-tom: A default value of 0 seems sensible.

backend/asmgen.ml
```diff
@@ -448,7 +461,11 @@ let compile_phrases ~ppf_dump ps =
           if !dump_cmm then fprintf ppf_dump "%a@." Printcmm.phrase p;
           match p with
           | Cfunction fd ->
-            compile_fundecl ~ppf_dump ~funcnames fd;
+            let compile_fundec = compile_fundecl ~ppf_dump ~funcnames in
+            let profiled_compile_fundec = match !profile_granularity with
+            | Function_level -> Profile.record ~accumulate:true fd.fun_name.sym_name compile_fundec
+            | _ -> compile_fundec (* Don't profile if -dfunc-level not selected*)
```
xclerc, line 467:
```suggestion
            | File_level -> compile_fundec (* Don't profile if -dfunc-level not selected*)
```
(It is often better to avoid catch-all pattern,
to be notified by the compiler when new
constructors are added to a type.)

milan-tom: In this case, would it be okay since this part of the code should only be executed if the granularity is `Function_level`? So, even if new granularities are added theoretically this should not change (unless we add a more fine-grained granularity than `Function_level` which seems unlikely).

milan-tom:
> I think it would be worthwhile to also amend `/ocaml/driver/compenv.ml` (in particular the `read_one_param` function), so that the new behavior can also be requested through the `OCAMLRUNPARAM` environment variable.

Done.

xclerc:
Thanks - there was a typo in my comment,
the environment variable is actually named
`OCAMLPARAM` (no `RUN`).

backend/asmgen.ml
xclerc: Generally, we prefer to have to perform
an obvious code update rather than take
the risk of missing an update. Warning 4
is enabled in most files in the backend,
but is unfortunately still disabled in
`Asmgen`.

milan-tom:
> Thanks - there was a typo in my comment, the environment variable is actually named `OCAMLPARAM` (no `RUN`).

Corrected in commit name.

ocaml/utils/clflags.mli
```diff
@@ -59,6 +59,9 @@ module Libloc : sig
   }
 end
 
+type profile_column = [ `Time | `Alloc | `Top_heap | `Abs_top_heap | `Counters ]
```
gretay-js, line 62: Why did `column` move here from `Profile` ?

backend/asmgen.ml
```diff
@@ -448,7 +464,12 @@ let compile_phrases ~ppf_dump ps =
           if !dump_cmm then fprintf ppf_dump "%a@." Printcmm.phrase p;
           match p with
           | Cfunction fd ->
-            compile_fundecl ~ppf_dump ~funcnames fd;
+            let compile_fundec = compile_fundecl ~ppf_dump ~funcnames in
+            (* Only profile if function level granularity selected*)
+            let profiled_compile_fundec = match !profile_granularity with
+            | Function_level -> Profile.record ~accumulate:true fd.fun_name.sym_name compile_fundec
+            | File_level -> compile_fundec
+            in profiled_compile_fundec fd;
```
gretay-js, line 472: This is a bit hard to read, maybe just improve formatting or don't let bind `profiled_compile_fundec`.

backend/asmgen.ml
```diff
@@ -348,12 +348,28 @@ let compile_fundecl ~ppf_dump ~funcnames fd_cmm =
...
         in
         cfg
-        ++ begin match regalloc with
-          | GI -> Profile.record ~accumulate:true "cfg_gi" Regalloc_gi.run
-          | IRC -> Profile.record ~accumulate:true "cfg_irc" Regalloc_irc.run
-          | LS -> Profile.record ~accumulate:true "cfg_ls" Regalloc_ls.run
-          | Upstream -> assert false
-        end
+        ++
+        (
```
gretay-js, line 352: This part of `Asmgen` is already hard to follow to get an overview of the pipeline. Is it possible to factor out counters into a separate function?

ocaml/utils/clflags.mli
milan-tom: `Profile` needed to know what command line arguments were set to determine whether to compute the counter function or not. So, `column` had to move here to avoid creating a circular dependency between `Profile` and `Clflags`.

backend/asmgen.ml
milan-tom: Replaced with [this](https://github.com/ocaml-flambda/flambda-backend/blob/177ba48b5fa25f618ecedc49e05f7d3790dc4626/backend/asmgen.ml#L468).

xclerc (approved):
If you can confirm that the output
of `-dtimings` is unaffected when
`-dcounters` is not passed, I think
it is good to go.

(@gretay-js also suggested off-line
it would be useful to give an
example of the new output in the
description of the pull request.)

PR #2789, opened 2024-07-16
Delete the modules related to scheduling

xclerc:
Since the backends we support currently
define scheduling as the identity, we can
delete the modules without any change
in terms of performance.


xclerc:
There is one thing that might feel a bit awkward
with the diff as of https://github.com/ocaml-flambda/flambda-backend/pull/2789/commits/8e2137d9d578711ec454ab48278cef3ed2c1d2a0: there is still a refence
to `Compiler_pass.Scheduling`.

@gretay-js do you think it is OK, or should we
also refactor `Compiler_pass`

mshinwell:
@gretay-js please review.  I'm ok with the proposed approach.

gretay-js (approved):
I think we should remove `Compiler_pass.Scheduling` from `Clflags`, and update tools that refer to it (such as build rules for and ir viewers). This is also used in `-stop-after` and `-save-ir` compiler flags, so it may need updating existing tests and testing. It can be done as a follow up. We should also remove `-dschedul` flag.

PR #2790, opened 2024-07-16
Resolve conflicts in ocaml/dune and ocaml/ocamltest/dune

mshinwell:
As subject.  Diff against `flambda-backend/main` to see that this is straightforward.  (We might need to adjust the `ocamltest/dune` file later, but this will do for now.)


mshinwell:
This is blocking and straightforward enough that I will self approve.

PR #2792, opened 2024-07-16
flambda2/Cmm changes for float16 bigarrays

mshinwell:
As subject.  Plus one tiny fix to allow flambda2 to compile.

### DIFF
--- a/backend/cmm_helpers.ml
+++ b/backend/cmm_helpers.ml
@@ -838,6 +838,38 @@ let unbox_vec128 dbg =
       | _ -> Cop (mk_load_immut Onetwentyeight_unaligned, [cmm], dbg))
     | cmm -> Cop (mk_load_immut Onetwentyeight_unaligned, [cmm], dbg))
 
+(* Conversions for 16-bit floats *)
+
+let float_of_float16 dbg c =
+  Cop
+    ( Cextcall
+        { func = "caml_double_of_float16";
+          ty = typ_float;
+          alloc = false;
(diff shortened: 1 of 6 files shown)


backend/cmm_helpers.ml
```diff
@@ -1743,6 +1744,7 @@ let bigarray_elt_size_in_bytes : Lambda.bigarray_kind -> int = function
 
 let bigarray_word_kind : Lambda.bigarray_kind -> memory_chunk = function
   | Pbigarray_unknown -> assert false
+  | Pbigarray_float16 -> Sixteen_unsigned
```
Gbury, line 1747: We need a new `memory_chunk` constructor to ensure that this is loaded as a `float` and not an unsigned 16-bit integer

Gbury (changes requested):

backend/cmm_helpers.ml
lthls: I don't think this is the right solution. If we want to stay as close as possible to upstream here, we should keep the memory as a 16-bit integer but call the appropriate functions (`float_of_float16` and `float16_of_float` from `Cmm_helpers`) to convert between the naked float and the 16-bit integer around loading and storing.

Gbury: That seems like the good solution indeed

mshinwell: Fixed

Gbury (approved):

PR #2794, opened 2024-07-16
Resolve conflicts in runtime/

mshinwell:
This is not yet ready for review.

List of upstream changes involved:
https://github.com/ocaml-flambda/flambda-backend/wiki/List-of-upstream-changes-from-5.1.1-to-5.2.0-split-by-area#runtime


mshinwell:
@stedolan I think this is ready for you to look at.  There are some `XXX` comments for you pointing out a few specific places that need extra checking.  In addition to the diff here, another useful diff to look at for some files (e.g. `major_gc.c` is the diff against `1b2c8e5e55`, which corresponds to the flambda-backend revision at the point of merging in the upstream `5.2.0` tag.

Changes for `systhreads` will be in a separate PR.

mshinwell:
https://github.com/ocaml-flambda/flambda-backend/wiki/Runtime-changes-that-are-specific-to-flambda%E2%80%90backend should be useful when reviewing too.

stedolan:
How much testing has this had?

mshinwell:
This compiles `libcamlrun.a` and `libasmrun.a` but hasn't been tested beyond that yet.

ocaml/runtime/domain.c
```diff
@@ -829,15 +774,9 @@ static void domain_create(uintnat initial_minor_heap_wsize,
...
 
-<<<<<<< HEAD
-  add_to_stw_domains(domain_self);
-||||||| 121bedcfd2
-  caml_reset_young_limit(domain_state);
-  add_to_stw_domains(domain_self);
-=======
-  caml_reset_young_limit(domain_state);
+  // XXX mshinwell: is this line needed?  It was deleted on our side
+  // caml_reset_young_limit(domain_state);
```
stedolan, line 778: It should remain deleted.

It was removed upstream as part of the statmemprof work (12923), and backported here as part of #2461. (These changes didn't make it into 5.2)

ocaml/runtime/domain.c
```diff
@@ -686,22 +647,12 @@ static void domain_create(uintnat initial_minor_heap_wsize,
...
-  caml_memprof_new_domain(parent, domain_state);
-  if (!domain_state->memprof) {
-    goto init_memprof_failure;
-  }
 
->>>>>>> 5.2.0
+  // XXX mshinwell: check if this unique_id line is needed and whether
+  // these two lines should be before caml_memprof_new_domain (as we had on
+  // flambda-backend) or after (as on 5.2, although 5.2 lacks the unique_id
+  // line).
```
stedolan, line 654: The `id` needs to be initialised before `caml_memprof_new_domain`. This code should look like upstream trunk and like current flambda-backend, both of which include the new statmemprof code. (Upstream 5.2.0 lacks this)

ocaml/runtime/gc_ctrl.c
```diff
@@ -281,15 +265,10 @@ static value gc_full_major_exn(int force_compaction)
   /* In general, it can require up to 3 GC cycles for a
      currently-unreachable object to be collected. */
   for (i = 0; i < 3; i++) {
-<<<<<<< HEAD
-    caml_empty_minor_heaps_once();
+    // XXX mshinwell for sdolan: please check conflict diff here.
+    // Maybe we should be taking the upstream version now?  (I think we made
+    // a fix to ensure that Gc.compact () did what people expected)
```
stedolan, line 270: I think we should take upstream 5.2's version.

Both sides have the same fix, but implemented slightly differently. Let's go with upstream's to minimise the diff.

ocaml/runtime/unix.c
```diff
@@ -17,6 +17,9 @@
 
 /* Unix-specific stuff */
 
+// XXX mshinwell for mshinwell: figure out what is going on here
+#include <sys/time.h>
```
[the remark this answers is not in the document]
mshinwell: I've removed this hack and added back the upstream stanza of:
```
#ifdef HAS_GETTIMEOFDAY
#include <sys/time.h>
#endif
```
However I've put this after the include of `caml/config.h`.  I don't see otherwise how `HAS_GETTIMEOFDAY` is going to be defined (it's in `s.h` which is included from `caml/config.h`), and indeed I get a build failure on macOS without this fix.

ocaml/runtime/custom.c
```diff
@@ -112,49 +115,12 @@ CAMLexport value caml_alloc_custom_mem(const struct custom_operations * ops,
...
-       it starts to deallocate dead blocks allocated during the previous cycle.
-       [heap_size / 150] is really [heap_size * (2/3) / 100] (but faster). */
-    caml_heap_size(Caml_state->shared_heap) / 150 * caml_custom_major_ratio;
-  mlsize_t max_minor =
-    Bsize_wsize (Caml_state->minor_heap_wsz) / 100 * caml_custom_minor_ratio;
-  value v = alloc_custom_gen (ops, bsz, mem, max_major, mem_minor, max_minor);
   size_t mem_words = (mem + sizeof(value) - 1) / sizeof(value);
+  value v = alloc_custom_gen (ops, bsz, mem, 0, get_max_minor());
+  // XXX mshinwell for sdolan: please check this conflict carefully
+  // and also, see if it relates to the XXX in bigarray.c
```
stedolan, line 121: This looks fine, but I'd prefer if you put `mem_words` _after_ the call to `alloc_custom_gen`. (This matches how it is in trunk upstream, giving us one less line of diff for next time)

stedolan (changes requested):

ocaml/runtime/major_gc.c
```diff
@@ -606,17 +511,15 @@ static void adopt_orphaned_work (void)
   while (f != NULL) {
     myf = domain_state->final_info;
     CAMLassert (caml_gc_phase == Phase_sweep_and_mark_main);
-    /* Since we are in [Phase_sweep_and_mark_main], the current domain has not
+    /* Since we are in [Phase_main], the current domain has not
        updated its finalisers. */
+    // XXX mshinwell for sdolan: please check this comment and the
+    // assertion I have taken from our side below.  Upstream, the previous
+    // assertion that caml_gc_phase == Phase_sweep_and_mark_main has been
+    // deleted, so I'm unsure what is here is correct.
```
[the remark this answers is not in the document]
mshinwell: I think I've done this, please check again.

ocaml/runtime/platform.c
```diff
@@ -183,26 +172,7 @@ static struct lf_skiplist mmap_blocks = {NULL};
 #endif
 
 void* caml_mem_map(uintnat size, int reserve_only)
->>>>>>> 5.2.0
 {
-<<<<<<< HEAD
```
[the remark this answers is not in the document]
mshinwell: Please check again @stedolan (rev `262ac28e2a` contains the fix).

stedolan (approved):

PR #2795, opened 2024-07-16
Fixes for ocaml/testsuite/

mshinwell:
Some of this was done pretty hastily but I think it's a good start.  We should probably do something like:
- get the tree in a state where tests can be run
- run the tests
- promote everything (there will be many changes caused by double quotes being put around type names etc)
- fix failures (checking the diff of a particular test if it fails)
- check the whole diff


mshinwell:
Update: this now contains all fixes to `ocaml/testsuite/` from `mshinwell/5.2-runtime-wip-main`, where except for a few typing tests everything seems to pass.

mshinwell:
The build system fix (relating to the `lib`) library relates to `Makefile` changes made upstream.  Instead of trying to port those I've simply written dune rules.

mshinwell:
This will need splitting for review, #2976 contains the first chunk (and indeed the majority).

PR #2796, opened 2024-07-16
Make pprintast handle new modes

liam923:
Currently, printing the ast of `type a = { x : b @@ portable }` (or any other modality besides `global`) fails. This PR makes this succeed, while still printing global as a legacy modality. For example, `type a = { x : b @@ global portable }` prints as `type a = { x : global_ b @@ portable }`.

### DIFF
--- a/ocaml/parsing/pprintast.ml
+++ b/ocaml/parsing/pprintast.ml
@@ -310,18 +310,29 @@ let optional_legacy_modes f m =
     legacy_modes f m;
     pp_print_space f ()
 
-let space_modality f {txt = Modality m; _} =
-  pp_print_string f " ";
+(** Some modalities have a preferred legacy syntax *)
+module Split_modalities = struct
+  type legacy_modality = | Legacy_modality of string [@@unboxed]
+  type t =
+    { legacy : legacy_modality loc list
+    ; nonlegacy : modality loc list
+    }
+
+  let split_list modalities =
+    let f { txt = modality; loc } =
+      match modality with
+      | Modality "global" -> Either.Left { txt = (Legacy_modality "global_"); loc }
+      | Modality _ -> Either.Right { txt = modality; loc }
+    in
+    let legacy, nonlegacy = List.partition_map f modalities in
+    { legacy; nonlegacy }
+end
+
+let modality f {txt = Modality m; _} =
   pp_print_string f m
 
-let legacy_modality f m =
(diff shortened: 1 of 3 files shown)


ocaml/testsuite/tests/language-extensions/pprintast_unconditional.ml
```diff
@@ -20,11 +20,13 @@ module Example = struct
 
   let modality_record  = parse module_expr
     "struct \
-      type t = {global_ x : string; global_ y : int} \
```
freemagma, line 23: Can you add a test case without a legacy modality in it? All the constructor args in these examples have `global_` on them

ocaml/parsing/pprintast.ml
```diff
@@ -355,10 +365,13 @@ let maybe_modes_type pty ctxt f c =
   | None -> pty ctxt f c
 
 let modalities_type pty ctxt f pca =
-  match pca.pca_modalities with
+  let modalities = Split_modalities.split_list pca.pca_modalities in
+  match modalities.legacy with
   | [] -> pty ctxt f pca.pca_type
```
freemagma, line 370: I think this will fail to print the nonlegacy modalities if there are no legacy modalities

liam923: Yep you're right

liam923: Fixed

ocaml/testsuite/tests/language-extensions/pprintast_unconditional.ml
liam923: Good point

ocaml/parsing/pprintast.ml
```diff
@@ -355,10 +365,16 @@ let maybe_modes_type pty ctxt f c =
...
-  match pca.pca_modalities with
-  | [] -> pty ctxt f pca.pca_type
-  | m ->
-    pp f "%a %a" legacy_modalities m (pty ctxt) pca.pca_type
+  let modalities = Split_modalities.split_list pca.pca_modalities in
+  let space_if_legacies =
+    match modalities.legacy with
+    | [] -> ""
+    | _ :: _ -> " "
+  in
```
freemagma, line 373: there's already a function `optional_legacy_modalities` that should do the space for you. everything else looks good though, so feel free to merge after fixing

freemagma (approved):

PR #2798, opened 2024-07-16
Add better tail-call optimization heuristic and `-less-tco` flag

tylerhou:
Also add a flag `-less-tco` to turn on/off the new tail-call optimization heuristic. The flag is cleared by default (the same as existing behavior). Diffs in
`ocaml/testsuite/tests/{functors,translprim}` revert changes in previous commits, where (the equivalent of) `-less-tco` was set by default.

When `-less-tco` is on, the new heuristic infers a function application to be tail-call optimized in the following circumstance:

  1. A function `f` is being called in tail position AND
  2. `f` is defined by a `let rec` AND
  3. The call site is within the binding definitions of the `let rec`

Example:

```ocaml
(* These calls are TCO'd since they satisfy the above three criteria. *)
let rec implicit_tail_foo n str =
  if n > 0 then implicit_tail_bar (n - 2) str (* TCO *) else str
and implicit_tail_bar n str =
  implicit_tail_foo (n + 1) str (* TCO *)

(* The call to implicit_tail_foo is not TCO'd because it does not
   satisfy criterion 3. *)
let call_foo = let local_ str = "hello" in implicit_tail_foo 10 str (* no TCO *)
```


tylerhou:
Closing in favor of #2800

PR #2800, opened 2024-07-16
Add better tail-call optimization heuristic and `-always-tco` flag

tylerhou:
Add better tail-call optimization heuristic and `-always-tco` flag

Also add a flag `-always-tco` to turn on/off the new tail-call
optimization heuristic. The flag is set to `true` by default (the same
as existing behavior). Diffs in
`ocaml/testsuite/tests/{functors,translprim}` revert changes in previous
commits, where (the equivalent of) `-always-tco` was cleared (set to
`false`) by default.

When `-always-tco` is not set, the new heuristic infers a function
application to be tail-call optimized in the following circumstance:

  1. A function `f` is being called in tail position AND
  2. `f` is defined by a `let rec` AND
  3. The call site is within the binding definitions of the `let rec`

Example:

```ocaml
(* These calls are TCO'd since they satisfy the above three criteria. *)
let rec implicit_tail_foo n str =
  if n > 0 then implicit_tail_bar (n - 2) str (* TCO *) else str
and implicit_tail_bar n str =
  implicit_tail_foo (n + 1) str (* TCO *)

(* The call to implicit_tail_foo is not TCO'd because it does not
   satisfy criterion 3. *)
let call_foo = let local_ str = "hello" in implicit_tail_foo 10 str (* no TCO *)
```


ocaml/typing/env.ml
```diff
@@ -1496,12 +1498,16 @@ let find_modtype_expansion path env =
 let rec is_functor_arg path env =
   match path with
     Pident id ->
-      begin try Ident.find_same id env.functor_args; true
-      with Not_found -> false
-      end
+      Option.is_some (Ident.find_same_opt id env.functor_args)
```
dvulakh, line 1501: I agree the `_opt` version of the function is an improvement over the raising version, but I would discourage making "cleanup" changes to unrelated parts of the compiler as part of this branch; in general it can make the rebase harder.

ocaml/typing/typecore.ml
```diff
@@ -370,18 +370,37 @@ let position_and_mode_default = {
...
-let should_infer_nontail maybe_ident =
-  match maybe_ident with
-  | Some ident -> begin match ident with
-    | Longident.Ldot _ -> true
-    | _ -> false
+type callee_expr =
+  | Not_a_function_callee
+  | Function_callee of Parsetree.expression
+
+type should_infer_tail_t = Infer_tail | Don't_infer_tail
```
dvulakh, line 377: I appreciate the wariness towards polymorphic variants you express. In this case, however, I think they would be acceptable, since this type is used only within the body of `should_infer_tail`. If you were returning `should_infer_tail_t` instead of a bool, I'd likely encourage keeping the variant: but the name of the function is clear enough that I prefer keeping the bool return and changing to polymorphic variants instead.

ocaml/testsuite/tests/typing-tailcalls/locals.ml
dvulakh, line 1: There are a few other interesting shapes that it might be nice to demo in this file. Nested `let rec`s and mixed structure/expression `let rec`s come to mind.

ocaml/testsuite/tests/typing-tailcalls/locals.ml
```diff
@@ -1,37 +1,42 @@
...
   let f = f
 end
+
 [%%expect {|
+val f : local_ 'a -> unit = <fun>
 module M : sig val f : local_ 'a -> unit end
 |}]
 
+
 let implicit_nontail () =
```
dvulakh, line 18: Can we also show implicit nontail of names bound by a surrounding `pexp_let`? Something with a shape like
```ocaml
let x = let g x = x in g 5;;
```

ocaml/typing/typecore.ml
```diff
@@ -8564,6 +8586,14 @@ and type_let_def_wrap_warnings
     end
     else exp_env
   in
+  let exp_env = if is_recursive then begin
+    (* (less-tco project) Used to infer whether function applications should be tail-call
```
dvulakh, line 8590: Is this a CR to be addressed before merging the branch, or a permanent documentation comment? If the former, please write in the same format as the other CRs you've left. If the latter, I think we can drop the `(less-tco project)` opening.

ocaml/typing/typecore.ml
```diff
@@ -1094,6 +1115,11 @@ let maybe_add_pattern_variables_ghost loc_let env pv =
        end
     ) pv env
 
+let add_fns_defined_by_letrec env pvs =
+  List.fold_right
```
dvulakh, line 1119: Nit: I don't think we care about the difference in semantics between `fold_right` and `fold_left`, so we can probably `fold_left` instead. (When I see a `fold_right`, I generally assume there's some significance to it that it's important I not miss.)

ocaml/typing/typecore.ml
```diff
@@ -8333,7 +8356,6 @@ and type_let ?check ?check_strict ?(force_toplevel = false)
   let spatl = List.map (pat_modes ~force_toplevel rec_mode_var) spatl in
   let attrs_list = List.map (fun (attrs, _, _, _) -> attrs) spatl in
   let is_recursive = (rec_flag = Recursive) in
-
```
dvulakh, line 8336: I assume this deletion is accidental? Even if not, I'd avoid making style changes.

dvulakh (changes requested):

dvulakh (approved):
Looks good, thanks!

dvulakh (approved):
Looks good now, thanks for fixing the test.

PR #2802, opened 2024-07-17
Extend `Jkind` with higher jkinds

jbachurski:
- Implements the interface of `Jkind` and appropriately moves out generalised implementations out of `Jkind.Type` (or removes them when they are no longer necessary).
- Fixes some pattern matches destructing `Type` jkinds, when they couldn't be written before

The diff should be fine, except for `jkind.ml` where some moves aren't particularly clean, but they seemed to make sense to avoid the `Jkind.Type` module containing implementations for `Jkind.t`.


goldfirere (commented):
.mli files (only) look reasonable

ocaml/typing/jkind_types.mli
```diff
@@ -152,8 +152,21 @@ module Type : sig
...
+module Arrow : sig
+  type 'a t =
+    { args : 'a list;
+      result : 'a
+    }
+end
+
+type 'type_expr t =
+  | Type of 'type_expr Type.t
+  | Arrow of 'type_expr t Arrow.t
```
lpw25, line 164: I'd probably just inline the definition of `Arrow.t` here. I don't think we gain much from splitting it out.

ocaml/typing/ctype.ml
```diff
@@ -2104,8 +2104,8 @@ let rec estimate_type_jkind env ty =
   end
   | Tvariant row ->
       if tvariant_not_immediate row
-      then Jkind (Type.Primitive.value ~why:Polymorphic_variant)
-      else Jkind (Type.Primitive.immediate ~why:Immediate_polymorphic_variant)
+      then Jkind (Type.Primitive.value ~why:Polymorphic_variant |> Jkind.of_type_jkind)
```
lpw25, line 2107: In `jkind.ml`, I think it would be worth adding re-exporting the contents of `Type.Primitive` in `Primitive` but wrapped in `Jkind.of_type_jkind`.

ocaml/typing/outcometree.mli
```diff
@@ -108,13 +108,15 @@ type out_jkind_user =
...
   | Ojkind_user_kind_of of out_type
+  | Ojkind_user_arrow of out_jkind_user Jkind_types.Arrow.t
 
 and out_jkind_const = { base : string; modal_bounds : string list }
 
 and out_jkind =
   | Ojkind_user of out_jkind_user
   | Ojkind_const of out_jkind_const
   | Ojkind_var of string
+  | Ojkind_arrow of out_jkind Jkind_types.Arrow.t
```
lpw25, line 119: I think it would be better not to reuse `Arrow.t` here. These types represent the syntax of OCaml and the stuff in `Jkind` is the semantics. It is usually better to keep the types representing these things distinct, as they have different purposes and should generally contain different data. Using the same type tends to lead to either needing to separate them again later, or worse failing to separate them even once the semantic and syntactic data has diverged.

ocaml/typing/oprint.ml
```diff
@@ -328,7 +328,20 @@ let print_jkind_with_modes ppf print_jkind base modes =
           (print_list (fun ppf -> fprintf ppf "%s") (fun ppf -> fprintf ppf "@ "))
           modes
 
-let print_out_jkind ppf = function
+let print_arrow ?is_atom ppf t Jkind_types.Arrow.{ args; result } =
```
lpw25, line 331: No need for `is_atom` to be optional: it is provided at both call sites.

ocaml/typing/printtyp.ml
```diff
@@ -1261,27 +1261,44 @@ let out_jkind_of_user_jkind (jkind : Jane_syntax.Jkind.annotation) =
           modes.txt
       in
       Ojkind_user_mod (base, modes)
-    | With _ | Kind_of _ | Arrow _ -> failwith "XXX unimplemented jkind syntax"
+    | Arrow (args, result) -> Ojkind_user_arrow {
```
lpw25, line 1264: Weird styling. Please move constructor to the next line.

ocaml/typing/printtyp.ml
```diff
@@ -1261,27 +1261,44 @@ let out_jkind_of_user_jkind (jkind : Jane_syntax.Jkind.annotation) =
...
+        args = List.map out_jkind_user_of_user_jkind args;
+        result = out_jkind_user_of_user_jkind result }
+    | With _ | Kind_of _ -> failwith "XXX unimplemented jkind syntax"
   in
   Ojkind_user (out_jkind_user_of_user_jkind jkind.txt)
 
 let out_jkind_of_const_jkind jkind =
   Ojkind_const (Jkind.Type.Const.to_out_jkind_const jkind)
 
+let rec out_jkind_of_jkind jkind = match Jkind.get jkind with
```
lpw25, line 1274: Weird styling. Please move match to next line.

ocaml/typing/printtyp.ml
```diff
@@ -1261,27 +1261,44 @@ let out_jkind_of_user_jkind (jkind : Jane_syntax.Jkind.annotation) =
...
+    | Var v -> (* This handles (X1). *)
+      if !Clflags.verbose_types
+      then Some (Ojkind_var (Jkind.Type.Sort.Var.name v))
+      else None
     end
-  | Var v -> (* This handles (X1). *)
-    if !Clflags.verbose_types
-    then Some (Ojkind_var (Jkind.Type.Sort.Var.name v))
-    else None
+  (* We ignore the rules above for arrows, which should always be printed *)
```
lpw25, line 1300: I agree we need to print the arrow, but we should still avoid printing the sort variables unless `!Clflags.verbose_types` is set. I'd suggest just printing `_` when we have a sort variable within an arrow. Probably we should update the syntax to accept that syntax, although we don't need to do that right now.

ocaml/testsuite/tests/typing-higher-jkinds/equalities.ml
```diff
@@ -0,0 +1,110 @@
+(* TEST
+ flags = "-extension layouts_alpha";
+ expect;
+*)
+
+type t : value => value
+
```
lpw25, line 7: I think these are a little easier to read if you get rid of the blank line before the `expect` block.

ocaml/testsuite/tests/typing-higher-jkinds/equalities.ml
```diff
@@ -0,0 +1,110 @@
...
+         sig type a : value end
+       is not included in
+         sig type a : value => value end
+       Type declarations do not match:
+         type a : value
+       is not included in
+         type a : value => value
+       The layout of the first is value, because
+         of the definition of a at line 4, characters 2-16.
+       But the layout of the first must be a sublayout of ((value) => value) (...??)
```
lpw25, line 50: What is causing the `??` part? Probably worth fixing that sooner rather than later.

ocaml/testsuite/tests/typing-higher-jkinds/equalities.ml
```diff
@@ -0,0 +1,110 @@
...
+         sig type a : value => value end
+       is not included in
+         sig type a : (value, value) => value end
+       Type declarations do not match:
+         type a : value => value
+       is not included in
+         type a : (value, value) => value
+       The layout of the first is ((value) => value) (...??)
+       But the layout of the first must be a sublayout of ((value, value) => value) (...??)
+|}]
```
lpw25, line 110: Probably worth adding some tests in this file that use `immediate` inside of arrows, to check the ordering of things.

ocaml/testsuite/tests/typing-higher-jkinds/applied.ml
```diff
@@ -0,0 +1,104 @@
+(* TEST
```
lpw25, line 1: Probably this should go in the PR for type application.

lpw25 (commented):
Reviewed everything except the file with all the real changes (jkind.ml). Should get to that tomorrow.

ocaml/testsuite/tests/typing-higher-jkinds/equalities.ml
jbachurski: History tracking not implemented for higher jkinds. Fixed a few PRs later down the line.

ocaml/typing/jkind_types.mli
jbachurski: I actually sort of like `Arrow.t`, as it's reused semi-often and it's nicer to have a type definition on-hand. It could be argued that `'type_expr t list * 'type_expr t` is already clear in being an n-ary type constructor signature, though. But if it is to be a record type, I'd prefer to keep it this way.

ocaml/typing/ctype.ml
jbachurski: It was done this way, but I reverted everything after @goldfirere's reviews: https://github.com/ocaml-flambda/flambda-backend/pull/2776#discussion_r1680929984

lpw25: (Discussing this further over on that PR)

jbachurski: If anything, I believe that due to the length of the rebase chain at this point I'd prefer to keep this on tab as a future refactor rather than acting on this right now. I think it's a change that won't get more difficult to make later (it just requires going through existing uses and adding promotions).

ocaml/typing/outcometree.mli
jbachurski: I do agree, but I'm not sure I see why you consider `Jkind_types` to be part of the semantics? My intuition was that `Jkind_types` is syntax, and `Jkind` is for semantics. Thus `Types` (mostly syntax) and `Jkind` (semantics) depend on `Jkind_types` (syntax).

lpw25: >  `Types` (mostly syntax)

`Types` is the semantics of the type language. The syntax of e.g. type expressions lives in `Typedtree`. There is also a second copy of the syntax over in `Outcometree` used for printing.

`Jkind_types` is just the types used in `Jkind`, the split there is just to avoid a dependency cycle.

jbachurski: Is it that you consider `Jkind` and `Jkind_types` notionally the same as they are just there to break the reference cycle?

ocaml/typing/ctype.ml
lpw25: The length of the rebase chain is not a good reason to avoid making changes to a PR that is in review. Now that you've got things set up properly, I think you should be able to merge changes from early PRs into later PRs with a single button press. You'll also only need to do it once at the point where the early PR is approved for merging.

ocaml/typing/outcometree.mli
lpw25: I do consider them as such, and their documentation comments say as much. But even were that not the case the types in `Jkind_types` are still describing the semantic kinds and not the syntax for kinds written by users (or printed in error messages).

ocaml/typing/ctype.ml
jbachurski: That's true. I was more thinking about merge conflicts than the rebase itself.

ocaml/typing/jkind_types.mli
jbachurski: Once it is removed from `out_jkind_*`s, `Arrow.t` is only used for `Jkind_types.t`, `Jkind_types.Const.t`, `Jkind.Desc.t` (`get`) and is passed around in a couple places. So I could remove it. Should it remain a record type or become a product type?

ocaml/typing/printtyp.ml
jbachurski: This sounds fine.
Side question: isn't it become a bit inconsistent that in the type jkind case both `value` and a sort variable don't get printed at all (this is irrespective of this change)?

ocaml/typing/jkind.ml
```diff
@@ -1752,12 +1574,505 @@ module Type = struct
...
+  | _ -> false
+
+let equate = equate_or_equal ~allow_mutation:true
+
+(* This mirrors the choice for Jkind.Type, following the note:
+   (layouts v2.8) allow_mutation is to be set to false *)
+let equal = equate_or_equal ~allow_mutation:true
+
+(* Generalises union and intersection at Arrows, as they are mutually recursive when
+   defining Arrow kind intersection *)
```
lpw25, line 1970: I think this set of functions would be easier to follow if written in a more direct style. Something like:
```ocaml
let rec intersection_or_error ~reason k1 k2 =
  match k1, k2 with
  | Type ty1, Type ty2 -> begin
      match Type.intersection_or_error ~reason ty1 ty2 with
      | Ok ty -> Ok (Type ty)
      | Error _ as err -> err
    end
  | Arrow(args1, res1), Arrow(args1, res1) -> begin
      match union_or_error_list ~reason args1 args2 with
      | Ok args -> begin
          match intersection_or_error ~reason res1 res2 with
           | Ok res -> Ok (Arrow { arg; result })
           | Error _ as err -> err
         end
       | Error _ as err -> err
     end
   | (Type _ | Arrow _), _ -> Error (Violation.of_ (No_intersection (k1, k2)))
   
and intersection_or_error_list ~reason arr1 ks1 arr2 ks2 =
  match ts1, ts2 with
  | [], [] -> Ok []
  | k1 :: rest1, k2 :: rest2 -> begin
      match intersection_or_error ~reason k1 k2 with
      | Ok k -> begin
          match intersection_or_error_list ~reason rest1 rest2 with
          | Ok rest -> Ok (k :: rest)
          | Error _ as err -> err
        end
      | Error _ as err -> err
    end
  | [], _ :: _ -> Error (Violation.of_ (No_intersection (arr1, arr2)))
  | _ :: _, [] -> Error (Violation.of_ (No_intersection (arr1, arr2)))
...
```
This code is very simple and clear and requires much less thought to see it is correct. It be even nicer using binding operators, but I'm not sure anyone has really used those in the compiler yet.

ocaml/typing/jkind.ml
```diff
@@ -1752,12 +1574,505 @@ module Type = struct
...
+let rec check_sub (t : t) (t' : t) : Misc.Le_result.t =
+  match t, t' with
+  | Type ty, Type ty' -> Type.check_sub ty ty'
+  | ( Arrow { args = args1; result = result1 },
+      Arrow { args = args2; result = result2 } ) ->
+    if List.length args1 <> List.length args2
+    then Misc.Le_result.Not_le
+    else
+      Misc.Le_result.combine_list
+        (check_sub result1 result2 :: List.map2 check_sub args2 args1)
```
lpw25, line 2033: How about:
```ocaml
let check_sub_list ts1 ts2 =
  if List.length ts1 <> List.length ts2
  then Misc.Le_result.Not_le
  else begin
    List.fold_left2
      (fun acc t1 t2 -> Misc.Le_result.combine acc (check_sub t1 t2))
      Misc.Le_result.Equal ts1 ts2
  end
```
and then call that in the `Arrow` case.

ocaml/typing/jkind.ml
```diff
@@ -1752,12 +1574,505 @@ module Type = struct
...
+  | _ -> Misc.Le_result.Not_le
+
+let sub t t' = Misc.Le_result.is_le (check_sub t t')
+
+let sub_or_error t t' =
+  if sub t t' then Ok () else Error (Violation.of_ (Not_a_subjkind (t, t')))
+
+let sub_with_history sub super =
+  match check_sub sub super with
```
lpw25, line 2042: Might as well call `sub` instead of `check_sub` here.

lpw25 (commented):
Looks good. Made some suggestions.

ocaml/typing/jkind.ml
```diff
@@ -1752,12 +1574,505 @@ module Type = struct
...
+    if List.for_all Option.is_some args && Option.is_some result
+    then
+      Some
+        (Arrow { args = List.map Option.get args; result = Option.get result })
+    else None
+
+(* of_const is defined above for use in Primitive *)
+
+(* The following are placeholders, which assume general higher-order kinds are
+   not present in annotations and declarations. *)
```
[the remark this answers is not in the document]
jbachurski: Good catch, as that one seems to have been one I shouldn't have kept around in the first place (they do parse higher-kinded annotations already).

ocaml/typing/jkind.ml
jbachurski: I attempted a refactor in a more direct style with some `Result.bind` with let operators (I wasn't happy with the nested pattern matches). I do find myself unsatisfied with the code repetition (2 times 2), so I did this generically in the simpler case of combining lists:

```ocaml
let rec combine_list_or_error ~violation combine ts1 ts2 =
  let ( let* ) = Result.bind in
  match ts1, ts2 with
  | [], [] -> Ok []
  | t1 :: ts1, t2 :: ts2 ->
    let* t = combine t1 t2 in
    let* ts = combine_list_or_error ~violation combine ts1 ts2 in
    Ok (t :: ts)
  | [], _ :: _ | _ :: _, [] -> Error violation
```

Perhaps this could be further simplified by factoring out the `map2`, but perhaps this is fine.
I kept `intersection_or_error` and `union_or_error` separate, though.

lpw25: The let operator is definitely nicer. I'm a little nervous of being the first to add it to the compiler code base (I think), but it's so much nicer let's just do it anyway.

Personally I'd prefer to just write `intersection_list_or_error` and `union_list_or_error` rather than using a higher-order combinator. First-order code is just less effort to read and review, and there isn't much benefit from sharing the shared pieces across the two definitions here. But I'm fine with keeping this version.

jbachurski: There seem to be sporadic uses of `let*` in `ocaml/parsing/builtin_attributes.ml` and `ocaml/utils/diffing.ml`; and of `let+` in `ocaml/lambda/tmc.ml` and `ocaml/lambda/value_rec_compiler.ml`. I'm not sure all of these are actively used, but I do notice CI passes so hopefully that takes the minimum compiler version into account.

ocaml/typing/jkind_types.mli
lpw25: This conversation is marked as resolved, but it looks like the `Arrow.t` type is still here.

jbachurski: Sorry - I assumed that since you said 'it's nice to have names for things' you agreed to have `Arrow.t` exist? Or were you just referring to field names?

lpw25 (approved):

PR #2804, opened 2024-07-17
Type algebra parsing with extended type applications

jbachurski:
- Extends the typed and untyped parse trees with `*typ_app` for (effectively) applying `type_expr list` to a `type_expr`. 
- The existing meaning and parsing of `*typ_constr` is preserved.

Tracking annotations should behave. Due to conflicts the parser accepts type applications of annotated type variables (like `('a : value) ('b : value => value)`), but this could be rejected.

`typetexp.ml` remains unimplemented for now, as it involves implementing logic for unapplied datatypes and type application.

There's a basic test added that could be extended.

### DIFF
--- a/ocaml/parsing/parser.mly
+++ b/ocaml/parsing/parser.mly
@@ -913,19 +913,22 @@ let unboxed_type sloc lident tys =
   let loc = make_loc sloc in
   Ptyp_constr (mkloc lident loc, tys)
 
-let app_typevar_list sloc tvs =
-  (* FIXME jbachurski: The source locations here are not right.
-      Variable annotations are also quietly forgotten. *)
+let app_typevar_list
+    (tvs : ((Lexing.position * Lexing.position) *
+             string with_loc *
+             Jane_syntax.Jkind.annotation option) list) =
   let tvs =
-    List.map (fun (tv, annot) ->
+    List.map (fun (loc, tv, annot) ->
       match annot with
-      | None -> mktyp ~loc:sloc (Ptyp_var tv.txt)
+      | None -> mktyp ~loc (Ptyp_var tv.txt)
       | Some jkind ->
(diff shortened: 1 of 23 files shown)


ocaml/typing/typedtree.mli
```diff
@@ -869,6 +869,7 @@ and core_type_desc =
   | Ttyp_arrow of arg_label * core_type * core_type
   | Ttyp_tuple of (string option * core_type) list
   | Ttyp_constr of Path.t * Longident.t loc * core_type list
+  | Ttyp_app of core_type * core_type list
```
lpw25, line 872: FWIW I'd normally leave this part out until the PR where you fill in the code for `typetexp.ml`. It does no harm to leave it in now though.

ocaml/parsing/parser.mly
```diff
@@ -4194,17 +4212,22 @@ with_type_binder:
 
 /* Polymorphic types */
 
-%inline typevar: (* : string with_loc * jkind_annotation option *)
+%inline typevar: (* : (position * position) * string * jkind_annotation option *)
```
lpw25, line 4215: IIUC this comment should still say `string with_loc` rather than `string`, as I think `mkrhs` adds a location.

ocaml/parsing/parser.mly
```diff
@@ -4495,15 +4518,40 @@ tuple_type:
...
+
+%inline applied_type:
+  | LPAREN ty=core_type RPAREN
+      { ty }
+  | mktyp(
+    QUOTE v=ident
+      { Ptyp_var v }
+  )   { $1 }
+
+%inline atomic_type_not_quote_gen:
```
lpw25, line 4539: Maybe `_inline` instead of `_gen`?

lpw25 (approved):
Looks good

PR #2808, opened 2024-07-17
Fix issues in the tree related to `Or_null`

dkalinichenko-js:
Make `Probe`s and variables `Captured_in_object` `value_or_null`.

The check for probes happens too late to infer the right nullability for the jkind, so accept `or_null` for backwards compatibility. Leave a CR to check that this is sound with null pointers.

The check for captured variables similarly happens after generalizing variables in pattern matches on tuples. It should be fine to capture null pointers in objects, so do it.

### DIFF
--- a/ocaml/testsuite/tests/typing-layouts-or-null/variables.ml
+++ b/ocaml/testsuite/tests/typing-layouts-or-null/variables.ml
@@ -397,3 +397,17 @@ Error: This type t_value_or_null dummy should be an instance of type 'a dummy
        But the kind of t_value_or_null must be a subkind of value
          because of the definition of constrained' at lines 1-2, characters 0-44.
 |}]
+
+(* Copied from the tree, should work. *)
+let should_work s =
+  let (a, b) =
+    Marshal.from_string s 0 in
+  ( object
+      method a = a
+      method b = b
+    end
+  )
+
+[%%expect{|
+val should_work : string -> < a : 'a; b : 'b > = <fun>
+|}]
--- a/ocaml/typing/env.ml
+++ b/ocaml/typing/env.ml
@@ -3094,9 +3094,11 @@ let unboxed_type ~errors ~env ~loc ~lid ty =
   match ty with
   | None -> ()
   | Some ty ->
-    (* Sometimes, this function is called on a generalized type variable
-       when it actually should check the specialized one. This is sound,
-       but incomplete. *)
+    (* The type is the type of a variable in the environment. It thus is likely generic. Despite
+       the fact that instantiated variables work better in [constrain_type_jkind] (because they
+       can be assigned more specific jkinds), we actually want to work on these generic types
+       here. After all, it's the value in the environment that is getting captured by the object,
+       not a specific instance of that variable. *)
     match !constrain_type_jkind env ty Jkind.Primitive.(value_or_null ~why:Captured_in_object) with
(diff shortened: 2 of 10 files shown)


ocaml/testsuite/tests/typing-layouts-or-null/variables.ml
```diff
@@ -397,3 +397,17 @@ Error: This type t_value_or_null dummy should be an instance of type 'a dummy
        But the kind of t_value_or_null must be a subkind of value
          because of the definition of constrained' at lines 1-2, characters 0-44.
 |}]
+
+(* Copied from the tree, should work. *)
+let should_work s =
```
goldfirere, line 402: I'm surprised this didn't work out of the box. It's true that the use of `a` and `b` in the body of the object constrain their kinds, but since `a` and `b` are inferred to have completely unconstrained types, this should be fine.

I don't think changing the capture requirement for objects is the right fix here. (Though perhaps it is expedient to do so to get nightly unbroken.)

dkalinichenko-js: Type variables for `a` and `b` are generalized before the check happens. Perhaps we should rework the check to happen before, but I think for now we should just allow `maybe_null` to fix tree builds.

goldfirere: That still should be OK. We get `a : ('a : value_non_null). 'a` and `b : ('b : value_non_null). 'b`. But then the occurrences of `a` and `b` should get instantiated to be values: instantiations can always be more specific.

Or maybe you're saying that the method variables `a` and `b` get generalized too soon? That could indeed cause a problem.... but I would expect methods to always need to be `value`s, too. So I still think everything should be OK here.

dkalinichenko-js: No, we get `a : ('a : value) . 'a` when we unpack the tuple, since tuple elements can be null. And then `'a` is generalized.

goldfirere: Oops. I mistyped in my comment. I meant that the type variables would be `value_or_null` (or is that `value_maybe_null` these days?). The rest of the comment stands.

dkalinichenko-js: Resolved per conversation.

ocaml/typing/env.ml
```diff
@@ -3094,7 +3094,10 @@ let unboxed_type ~errors ~env ~loc ~lid ty =
   match ty with
   | None -> ()
   | Some ty ->
-    match !constrain_type_jkind env ty Jkind.Primitive.(value ~why:Captured_in_object) with
+    (* Sometimes, this function is called on a generalized type variable
+       when it actually should check the specialized one. This is sound,
+       but incomplete. *)
```
goldfirere, line 3099:
```suggestion
    (* The type is the type of a variable in the environment. It thus is likely generic. Despite
       the fact that instantiated variables work better in [constrain_type_jkind] (because they
       can be assigned more specific jkinds), we actually want to work on these generic types
       here. After all, it's the value in the environment that is getting captured by the object,
       not a specific instance of that variable. *)
```

goldfirere (approved):

PR #2812, opened 2024-07-18
Unsigned comparisons

gretay-js:
The middle-end/backend can already emit unsigned comparisons. This PR exposes them as the following new builtins to the users: `%lessequal_unsigned` `%lessthan_unsigned` `%greaterequal_unsigned` `%greaterthan_unsigned` and `compare_unsigned` for int and boxed/unboxed int32/int64/nativeint, reusing the existing typed-based translation of compare primitives. 

The main changes are in `Lambda` and `Translprim`. The rest of the PR is refactoring to propagate the signed/unsigned information to the middle end.

This is a draft because it's completely untested.

### DIFF
--- a/ocaml/lambda/translprim.ml
+++ b/ocaml/lambda/translprim.ml
@@ -27,6 +27,7 @@ type error =
   | Unknown_builtin_primitive of string
   | Wrong_arity_builtin_primitive of string
   | Invalid_floatarray_glb
+  | Unsigned_primitive_not_supported of string
 
 exception Error of Location.t * error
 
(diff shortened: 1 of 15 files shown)


ocaml/lambda/translprim.ml
```diff
@@ -60,15 +61,15 @@ type comparison =
   | Compare
 
 type comparison_kind =
-  | Compare_generic
-  | Compare_ints
+  | Compare_generic of { signed : bool }
```
TheNumbat, line 64: Polymorphic comparison probably shouldn't have an unsigned option, right?

TheNumbat (commented):
This approach looks good, I can review in more detail after you add tests

ocaml/lambda/translprim.ml
gretay-js: Right. They all start as `Compare_generic` in `lookup_primitive`. I will try to move this `bool` to the `Comparison` constructor of `prim`, it'll be cleaner. Thanks!

PR #2813, opened 2024-07-18
Annotate Cmm.fundecl with only_kept_for_zero_alloc

mshinwell:
It should be the case that the flag is set iff:
- the function is unreachable; and
- the function is annotated with a zero-alloc `Check` attribute; and
- flambda2 would have deleted the function save for the presence of such attribute.

WIP for @gretay-js

### DIFF
--- a/backend/cmm.ml
+++ b/backend/cmm.ml
@@ -336,6 +336,7 @@ type fundecl =
     fun_codegen_options : codegen_option list;
     fun_poll: Lambda.poll_attribute;
     fun_dbg : Debuginfo.t;
+    fun_only_kept_for_zero_alloc : bool;
   }
 
 type data_item =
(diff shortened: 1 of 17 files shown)


mshinwell:
Small example:
```
% cat rec.ml
let[@zero_alloc] rec f x = 1 + g (x * 2)
and[@zero_alloc] g x = 2 + f (x - 3)

let bar x = x + 1
% cat rec.mli
val bar : int -> int
```

mshinwell:
(If we want to go ahead with something like this, @Gbury should look at the dataflow-related part of this PR to make sure it matches the specification above.)

Gbury:
As said some time ago on slack: the current code seems okay (for the data flow part at least), but it's not great to run the whole data_flow analysis twice. Instead I'd prefer if only the reachability analysis part of the flow analysis was run twice.

mshinwell:
@gretay-js what should we do with this?

PR #2814, opened 2024-07-18
Fix conflicts in Unix library stubs

mshinwell:
As subject


ocaml/otherlibs/unix/signals.c
```diff
@@ -84,11 +83,6 @@ CAMLprim value caml_unix_sigprocmask(value vaction, value vset)
 #else
```
stedolan, line 83: The comment is obsolete now

stedolan (changes requested):

ocaml/otherlibs/unix/signals.c
mshinwell: Fixed, please approve

stedolan (approved):

PR #2820, opened 2024-07-18
Fix easy typing conflicts

ncik-roberts:
Fix the easy, low-hanging fruit of typing conflicts.


ncik-roberts:
I've taken our side of `typing/shape_reduce.ml`. It's possible the `ocaml/ocaml` side is just better, but @liam923 is going to think about this as part of the broader effort of bringing things inline with upstream.

goldfirere:
The differences between our shape_reduce and upstream's do not seem all that profound. I imagine Ulysse would be able to be helpful in sorting it out, @liam923.

goldfirere (approved):

PR #2821, opened 2024-07-18
Fix conflicts in `typing/printtyp.ml`

ncik-roberts:
The changes are slightly more than trivial (but only slightly). I've put them in a separate PR from #2820

### DIFF
--- a/ocaml/typing/printtyp.ml
+++ b/ocaml/typing/printtyp.ml
@@ -371,7 +371,6 @@ let pervasives_name namespace name =
           set namespace @@ M.add name (Associated_to_pervasives r) (get namespace);
           r
 
-<<<<<<< HEAD
 (** Lookup for preexisting named item within the current {!printing_env} *)
 let env_ident namespace name =
   if S.mem name !protected then None else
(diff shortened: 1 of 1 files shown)


ocaml/typing/printtyp.ml
```diff
@@ -2428,37 +2290,15 @@ let class_type ppf cty =
   prepare_class_type [] cty;
   !Oprint.out_class_type ppf (tree_of_class_type Type [] cty)
 
-<<<<<<< HEAD
-let tree_of_class_param param var_inj =
```
goldfirere, line 2432: It looks like we're losing some logic here, but I did not look all that closely.

ocaml/typing/printtyp.ml
```diff
@@ -3274,36 +3108,9 @@ let explain mis ppf =
...
-      try
-        ignore(Env.find_type p env : Types.type_declaration)
-      with Not_found ->
-=======
-    begin match Env.find_type p env with
-    | exception Not_found ->
->>>>>>> 5.2.0
-        fprintf ppf
-<<<<<<< HEAD
-          "@,@[<hov>Type %a was considered abstract@ when checking\
```
goldfirere, line 3291: It looks like we're losing this, but I don't think we should.

ocaml/typing/printtyp.ml
ncik-roberts: I think we're not losing any logic here.

ocaml/typing/printtyp.ml
ncik-roberts: I think we retain it below: https://github.com/ocaml-flambda/flambda-backend/pull/2821/files#diff-384fa66d4466a5b8c1e3dd8883e02c0c905aa4377e84be7c5afad25d0aa93cd5L3310-L3317

goldfirere (approved):

PR #2825, opened 2024-07-19
Gc stats fix

tdelvecchio-jsc:
Fixes an issue where `Gc.quick_stat ()` would not update minor words until a gc promotion happened. Also fixes a mis-reporting of minor words on `caml_do_exit`.

Testing: In current HEAD, repeatedly calling `Gc.quick_stat ()` would not update the minor words until you allocate enough to cause a gc, but now it updates on every call to `Gc.quick_stat`, which matches the behavior of the Ocaml 4 runtime.


ncik-roberts:
This seems right to me! @stedolan should also glance.

ocaml/runtime/gc_stats.c
```diff
@@ -57,11 +57,18 @@ void caml_accum_alloc_stats(
...
+    + (double) Wsize_bsize ((uintnat) local->young_end -
+                            (uintnat) local->young_ptr);
+}
+
 void caml_collect_alloc_stats_sample(
   caml_domain_state *local,
   struct alloc_stats *sample)
 {
-  sample->minor_words = local->stat_minor_words;
+  sample->minor_words = caml_minor_words(local);
```
stedolan, line 71: I think this change is in the wrong place, since this only makes sense to do for the current domain (see other comment)

stedolan:
It would be good to get a clear spec (and ideally, tests) for what the intended behaviour actually is in multi-domain programs. There are several functions that compute stats about the GC, including `stat`, `quick_stat`, `counters` and `minor_words`. In the switch to OCaml 5, we needed to pick whether these refer to single-domain or whole-system stats.

The original design was that `stat` and `quick_stat` would be whole-system stats, sampled periodically but not immediately up-to-date (computing exact up-to-date stats is expensive), while `counters` and `minor_words` would be single-domain and up-to-date.

This changed [at some point](https://github.com/ocaml/ocaml/commit/a67b13696e341efa159934d7d02c6aeac12137ef), so now stat and quick_stat are sampled stats for other domains plus immediate stats for the current domain. That patch isn't quite right though, because it gets the wrong value for minor allocations, and I think the intent here is to change that. So, is the idea that after this patch, the minor allocations / promotions fields of quick_stat and counters should change by the same amount if only one domain is allocating?

What sort of code motivated this change? If it was using `quick_stat` to measure allocations of some code, then even with this change it'll still break in a multi-domain program, and maybe it should be changed to use `counters` instead.

ncik-roberts:
It was a single-domain program that motivated this change: Users noticed that `Gc.quick_stat ()` / `Gc.stat ()` weren't picking up new minor allocations. (We were confused for a little while too.)

> That patch isn't quite right though, because it gets the wrong value for minor allocations, and I think the intent here is to change that. So, is the idea that after this patch, the minor allocations / promotions fields of quick_stat and counters should change by the same amount if only one domain is allocating?
Yes.

I'd be in favor of making the necessary changes in this patch to just dereference `young_ptr`/`young_end` for the local domain. I'd hazard that many programs continue to be single domain, and this will ensure that `quick_stat` behaves more as it did in 4.11. What do you think about that? (In my view, the standard is whether we'd be happy to upstream this change.)

stedolan:
Apologies, I missed that more code had been pushed here addressing my comments above! At this stage, let's wait until 5.2 is merged (since it affects a bunch of the same code), but then I'll re-read this and we should merge it.

mshinwell:
@stedolan ping

mshinwell:
@stedolan ping again

PR #2826, opened 2024-07-19
Allow `string` to mode cross

liam923:
Modify the definition of `string` in predef to allow crossing along linearity, contention, and portability.

### DIFF
--- a/ocaml/testsuite/tests/typing-modes/val_modalities.ml
+++ b/ocaml/testsuite/tests/typing-modes/val_modalities.ml
@@ -275,6 +275,7 @@ module Close_over_value :
   end
 |}]
 
+(* CR mode-crossing: this should use mutable record *)
 module Close_over_value_monadic = struct
   module M = struct
     let r @ uncontended = fun x -> x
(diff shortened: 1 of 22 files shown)


goldfirere (approved):

ocaml/testsuite/tests/typing-modes/val_modalities.ml
```diff
@@ -277,7 +277,7 @@ module Close_over_value :
 
 module Close_over_value_monadic = struct
   module M = struct
-    let r @ uncontended = "hello"
+    let r @ uncontended = fun x -> x
```
riaqn, line 280: Might be better to use `int ref`  (because arrows will cross contention)

riaqn (commented):
I reviewed `typing-modes` and `typing-unique`.

ocaml/testsuite/tests/typing-modes/val_modalities.ml
riaqn: Unfortunately the currently mode-checking wrt mutable fields is too strict. Can you add a `(* CR mode-crossing: this should use mutable record *)` and I will fix in later PRs?

liam923: Done

PR #2827, opened 2024-07-19
Fix more cases of `zero_alloc` / modality variables getting into signatures

ccasin:
These cases have to do with destructive substitution and module alias types.  When we expand an `Mty_alias` into its definition, we must make the variables rigid if the new module type can be used to constrain something.  To quote the tests:
```ocaml
module M_outer = struct
  module M_inner = struct
    type t
    let f () = () (* this has a zero_alloc variable (and a modality variable) *)
  end
end

(* In [M_outer_strong], the type of [M_inner] will be
   [Mty_alias M_outer.M_inner] *)
module type M_outer_strong =  sig
  module M_inner = M_outer.M_inner
end

(* As a result of the destructive substition, we'll eliminate the alias.  When
   this happens, we shouldn't keep the zero_alloc variables on [f] *)
module type M_outer_subst = M_outer_strong
  with type M_inner.t := M_outer.M_inner.t

(* If we had kept the zero_alloc or modality variables, this would give an
   error. *)
module M : M_outer_subst = struct
  module M_inner = struct
    let f () = ()
  end
end
```

We should think a little harder about other ways this might happen - in particular other uses of `Typemod.extract_sig` or `Mtype.scrape_alias`.  I've taken a quick scan, and will think about it some more (as the reviewer of this PR might do) but don't want to delay getting this PR in.

This PR has three commits:
* The first one just adds some debug printers I was tired of not having.
* The second adds some tests that fail
* The third fixes the bug and, correspondingly, the tests

### DIFF
--- a/ocaml/testsuite/tests/typing-zero-alloc/signatures.ml
+++ b/ocaml/testsuite/tests/typing-zero-alloc/signatures.ml
@@ -1259,9 +1259,7 @@ module M_outer : sig module M_inner : sig type t val f : unit -> unit end end
 module type M_outer_strong = sig module M_inner = M_outer.M_inner end
 module type M_outer_subst =
   sig module M_inner : sig val f : unit -> unit end end
->> Fatal error: zero_alloc: variable constraint
-Uncaught exception: Misc.Fatal_error
-
+module M : M_outer_subst
 |}]
 
 (********************************************)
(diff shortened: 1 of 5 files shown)


ccasin:
Request review from @riaqn

riaqn (approved):
LGTM.

I have some further questions but I will resolve them via other channels.

PR #2828, opened 2024-07-19
Fix conflicts in `typing/typetexp.ml`

ncik-roberts:
A few things to note about this one:
  * I add `Style.as_inline_code` for the error messages we added, to match upstream.
  * I modify the Jane Syntax-encoded parsetree for type aliases with layout annotations (`Ltyp_alias`) to add a location, to match the change to `Ptyp_alias` in https://github.com/ocaml/ocaml/pull/12639.

Otherwise, this PR generally takes our side of the diff.

### DIFF
--- a/ocaml/typing/typetexp.ml
+++ b/ocaml/typing/typetexp.ml
@@ -193,37 +193,6 @@ end = struct
   let used_variables =
     ref (TyVarMap.empty : (type_expr * Location.t) TyVarMap.t)
 
-<<<<<<< HEAD
-||||||| 121bedcfd2
-  (* These are variables we expect to become univars (they were introduced with
-     e.g. ['a .]), but we need to make sure they don't unify first.  Why not
-     just birth them as univars? Because they might successfully unify with a
-     row variable in the ['a. < m : ty; .. > as 'a] idiom.  They are like the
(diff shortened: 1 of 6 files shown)


ocaml/typing/typetexp.ml
```diff
@@ -715,34 +642,17 @@ let transl_bound_vars : (_, _) Either.t -> _ =
...
 
-let rec transl_type env ~policy ?(aliased=false) ~row_context mode styp =
-  Builtin_attributes.warning_scope styp.ptyp_attributes
-    (fun () -> transl_type_aux env ~policy ~aliased ~row_context mode styp)
-||||||| 121bedcfd2
-let rec transl_type env policy styp =
-  Builtin_attributes.warning_scope styp.ptyp_attributes
-    (fun () -> transl_type_aux env policy styp)
-=======
-(* Forward declaration (set in Typemod.type_open) *)
```
ccasin, line 726: I think you dropped this comment on `type_open`.

ccasin (approved):
Looks good, just one tiny nit.

ocaml/typing/typetexp.ml
ncik-roberts: Indeed, thanks.

PR #2829, opened 2024-07-22
Dump profile information to file if command line flag -dump-into-file set

milan-tom:
Brings `driver/optmaindriver.ml` in line with change to `ocaml/driver/optmaindriver.ml` in #943

### DIFF
--- a/driver/optmaindriver.ml
+++ b/driver/optmaindriver.ml
@@ -188,7 +188,6 @@ let main unix argv ppf ~flambda2 =
       end;
       Profile.print ppf !Clflags.profile_columns ~timings_precision:!Clflags.timings_precision
     in
-    (* If -dump-into-csv is set, do not dump profile information to .dump file or stdout *)
     if !Clflags.dump_into_csv then
       Compmisc.with_ppf_file ~file_prefix:"profile" ~file_extension:".csv" output_profile_csv
     else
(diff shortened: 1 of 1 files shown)


driver/optmaindriver.ml
```diff
@@ -168,29 +166,31 @@ let main unix argv ppf ~flambda2 =
...
+        Format.fprintf ppf "    %i major\n" stats.major_collections;
       end;
-      Profile.print
-        Format.std_formatter
-        !Clflags.profile_columns
-        ~timings_precision:!Clflags.timings_precision;
-    end;
+      Profile.print ppf !Clflags.profile_columns ~timings_precision:!Clflags.timings_precision
+    in
+    (* If -dump-into-csv is set, do not dump profile information to .dump file or stdout *)
```
ncik-roberts, line 191: This comment falls into the trap of stating a fact but not indicating why the fact is important or interesting. I might suggest:
```suggestion
    (* -dump-to-csv means we should always produce a .csv file. *)
```
But this comment is maybe not even worth mentioning, and we should perhaps delete it instead.

ncik-roberts (approved):
Sorry for the delay!

The change looks correct: the output will now be sent to a file in the dump dir when `-dump-dir` is passed.

I was misled by the description, I think. #943 looks to be about printing output to stdout instead of stderr, whereas this PR looks to be about sending output to the dump directory.

If my impression is correct, this looks good to merge.

driver/optmaindriver.ml
milan-tom: It seems that the code already conveys the functionality better than a comment would so I have removed it.

PR #2833, opened 2024-07-22
Don't allocate in bytecode on a labeled tuple match

freemagma:
In `typecore.ml`, labeled tuple lets are turned into matches (See `turn_let_into_match`) in order to help type inference. This causes transformations like:
```
let ~a, ~b = match x with true -> ~a:1, ~b:2 | false -> ~a:2, ~b:1 in a - b
```
into 
```
match (match x with true -> ~a:1, ~b:2 | false -> ~a:2, ~b:1) with ~a, ~b -> a - b
```

While the first snippet looks like it shouldn't allocate a tuple, it actually does after transformation to the second snippet (in bytecode).

To prevent this, I added a special case to `translcore.ml` that effectively reverses `turn_let_into_match`.

### DIFF
--- a/ocaml/lambda/translcore.ml
+++ b/ocaml/lambda/translcore.ml
@@ -488,8 +488,29 @@ and transl_exp0 ~in_new_scope ~scopes sort e =
            ~position ~mode (transl_exp ~scopes Jkind.Sort.for_function funct)
            oargs (of_location ~scopes e.exp_loc))
   | Texp_match(arg, arg_sort, pat_expr_list, partial) ->
-      transl_match ~scopes ~arg_sort ~return_sort:sort e arg pat_expr_list
-        partial
+      let default () =
+        transl_match ~scopes ~arg_sort ~return_sort:sort e arg pat_expr_list
+          partial
+      in
+      (* Test if we can convert this match into a let. This is the effective opposite of
+         [turn_let_into_match] in [typecore.ml] *)
+      (match pat_expr_list, partial with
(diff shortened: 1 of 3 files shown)


ocaml/lambda/translcore.ml
```diff
@@ -488,8 +488,29 @@ and transl_exp0 ~in_new_scope ~scopes sort e =
...
+        transl_match ~scopes ~arg_sort ~return_sort:sort e arg pat_expr_list
+          partial
+      in
+      (* Test if we can convert this match into a let. This is the effective opposite of
+         [turn_let_into_match] in [typecore.ml] *)
+      (match pat_expr_list, partial with
+      | [{c_lhs; c_rhs; c_guard = None}], Total when c_rhs.exp_desc <> Texp_unreachable ->
+        let val_pat, exn_pat = split_pattern c_lhs in
+        (match val_pat, exn_pat with
+          | Some { pat_desc = Tpat_or _; _ }, _ -> default ()
```
ccasin, line 501: Why are or patterns handled specially here?  Do also need to worry about, e.g., an `as` pattern with an or pattern inside of it?

freemagma:
Based on a conversation with @ccasin, this change is probably too risky to be worth the minor optimization. Reversing `turn_let_into_match` also ends up changing the compilation of other code (not just `turn_let_into_match` code), the effects of which are hard to understand. It's probably not worth potentially changing the compilation of lots of matches just to improve this one minor case.

PR #2835, opened 2024-07-22
Update arrow types to cross contention and uniqueness

liam923:


ocaml/testsuite/tests/typing-modes/currying.ml
```diff
@@ -458,6 +458,8 @@ val f : ('a -> 'b -> 'c) -> unique_ 'a -> 'b -> 'c = <fun>
 let f (g @ unique) x =
   g x x [@nontail]
 [%%expect{|
+val f : unique_ ('a -> 'a -> 'b) -> ('a -> 'b) = <fun>
+|}, Principal{|
 val f : unique_ ('a -> 'a -> 'b) -> 'a -> 'b = <fun>
 |}]
```
liam923, line 464: I have no idea what's going on here

liam923:
Some tests are failing on CI but succeeding locally. I need to figure out what's going on.

ocaml/testsuite/tests/typing-modes/currying.ml
goldfirere: I think the observed behavior is correct.

The second type really means `unique_ ('a -> 'a -> 'b) -> once_ ('a -> 'b)`. The `once_` is necessary to deal with the possibility of partial application: `f x` is actually a closure with a pointer to the unique `x`, and such things must be `once`. The first type, with its seemingly-redundant parens, means just what it says. That makes sense, because the function can mode-cross uniqueness, and so we don't actually need the `once`. Yet the reason we want the `once` in the second type is because of principality: we only infer that `g` is a function by its usage, which is not principal. If you put a type signature on `g`, then I think you'd only get the second type.

Bottom line: all is well here.

ocaml/testsuite/tests/typing-modes/portable-contend.ml
```diff
@@ -224,12 +224,12 @@ Error: This expression has type bytes but an expression was expected of type
...
 let foo : 'a @ contended portable -> (string -> string) @ portable @@ nonportable contended = fun a b -> best_bytes ()
-(* CR layouts v2.8: arrows should cross contention. *)
 [%%expect{|
-Line 1, characters 4-118:
+Line 1, characters 105-118:
 1 | let foo : 'a @ contended portable -> (string -> string) @ portable @@ nonportable contended = fun a b -> best_bytes ()
-        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-Error: This value is contended but expected to be uncontended.
+                                                                                                             ^^^^^^^^^^^^^
+Error: This expression has type bytes but an expression was expected of type
```
goldfirere, line 231: Is this what this test was expecting to test? @riaqn

goldfirere (approved):

ocaml/testsuite/tests/typing-modes/portable-contend.ml
riaqn: The typo is mine - could you fix this so this test passes?

ocaml/typing/jkind.ml
```diff
@@ -1229,6 +1229,22 @@ let for_boxed_variant ~all_voids =
   then Primitive.immediate ~why:Enumeration
   else Primitive.value ~why:Boxed_variant
 
+let for_arrow (_arg_type : Types.type_expr) (_ret_type : Types.type_expr) =
+  (* Until [with] syntax is supported, ignore argument and return types *)
```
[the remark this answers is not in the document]
liam923: Duplicate of Richard's comment, see my response above.

ocaml/testsuite/tests/typing-modes/portable-contend.ml
liam923: I fixed the typo here, as well as two other places (immediately before and after this test).

liam923:
> Some tests are failing on CI but succeeding locally. I need to figure out what's going on.

Turns out this was just because I needed to merge in changes from main.

liam923:
This is now ready for re-review @riaqn

riaqn:
Thank you - I realized that some of the tests is no longer useful so I removed them. Feel free to merge once CI green. Thanks!

PR #2837, opened 2024-07-23
Unboxed tuples fixes

Ekdohibs:
Unboxed tuples appearing in free variables of closures were previously failing. This PR can be tested with https://github.com/Ekdohibs/flambda-backend/commit/dcd99956a757e4d4fb45e473e96151f7a01013d9 which sits on top of the unboxed tuples frontend work.
Example code that was previously failing:
```ocaml
type t = #(int * int)

let id_t (x : t) = x

let app f (x : t) = f x

let f y = app (fun _ -> id_t y) y
```

### DIFF
--- a/middle_end/flambda2/from_lambda/lambda_to_flambda.ml
+++ b/middle_end/flambda2/from_lambda/lambda_to_flambda.ml
@@ -1607,6 +1607,8 @@ and cps_function env ~fid ~(recursive : Recursive.t) ?precomputed_free_idents
     | Some ids -> ids
     | None -> Lambda.free_variables body
   in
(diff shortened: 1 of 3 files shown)


ccasin:
I've tested this PR on the unboxed tuples branch (see test 3 [here](https://github.com/ccasin/flambda-backend/blob/unboxed-tuples-rebase/ocaml/testsuite/tests/typing-layouts-products/unboxed_tuples.ml), which passes).  It will need an flambda2 reviewer, but waiting for @mshinwell's return is OK if no one is available.

Ekdohibs:
Note: I added fixes to other problems with unboxed products in lambda_to_flambda.

mshinwell (approved):

PR #2838, opened 2024-07-23
Fix unboxing EPA of wrong kind

Ekdohibs:
It still used a poison const independent of kind.
Was failing on:
```ocaml
type t =
  | A of float#
  | B

let evaluate () =
  let[@inline] get () =
    let p = if Sys.opaque_identity true then A (Sys.opaque_identity #4.0) else B in
    match p with
    | A x -> x
    | B   -> assert false
  in
  let x = get () in
  x
;;
```


Gbury (approved):

ncik-roberts:
Thanks, I am cherry-picking it onto `5.1.1minus-19`. I'd be in favor of adding the test case. Is there a reason to not do this?

ncik-roberts:
I spoke with @Ekdohibs , and we think the test overfits a bit too much to the failed rebase so I will merge.

PR #2839, opened 2024-07-23
Improve error message for unbound path in inferred type

smjleo:
Previously, if the module type being inferred (via the underscore syntax) contains a path that hasn't yet been encountered in the struct, it led to very confusing error messages (and different error messages in slightly different contexts, though it has the same root problem). This PR introduces a check to error out early if we see an unbound path.

### DIFF
--- a/ocaml/typing/typemod.ml
+++ b/ocaml/typing/typemod.ml
@@ -159,18 +159,14 @@ module Sig_map = struct
     in
     List.fold_left add_item map sg
 
-  let find_value name map = String.Map.find_opt name map.values
   let find_type name map = String.Map.find_opt name map.types
   let find_module name map = String.Map.find_opt name map.modules
   let find_module_type name map = String.Map.find_opt name map.module_types
-  let find_class name map = String.Map.find_opt name map.classes
   let find_class_type name map = String.Map.find_opt name map.class_types
 
-  let has_value name map = Option.is_some (find_value name map)
   let has_type name map = Option.is_some (find_type name map)
   let has_module name map = Option.is_some (find_module name map)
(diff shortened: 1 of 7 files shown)


ocaml/typing/typemod.ml
```diff
@@ -1607,6 +1617,75 @@ let mksig desc env loc =
...
+    { Btype.type_iterators with
+      Btype.it_path = (fun path_kind path ->
+        match path with
+        | Pident id ->
+            let name = Ident.name id in
+            let (in_sig, in_str) =
+              begin match path_kind with
+                | Path_value ->
+                    Sig_map.has_value name sig_map,
+                    Sig_map.has_value name str_map
```
riaqn, line 1647: This branch should be unreachable. I.e., a module type won't contain reference to a value. If this happens something is very wrong. Therefore, we should say `Misc.fatal_error` for this.

ocaml/typing/typemod.ml
```diff
@@ -1607,6 +1617,75 @@ let mksig desc env loc =
...
+                    Sig_map.has_type name str_map
+                | Path_module ->
+                    Sig_map.has_module name sig_map,
+                    Sig_map.has_module name str_map
+                | Path_modtype ->
+                    Sig_map.has_module_type name sig_map,
+                    Sig_map.has_module_type name str_map
+                | Path_class ->
+                    Sig_map.has_class name sig_map,
+                    Sig_map.has_class name str_map
```
riaqn, line 1659: Similarly, I suspect that module type can't refer to a class. Let's discuss this offline.

riaqn (commented):
Looks good in general - some minor issues.

riaqn (approved):
LGTM

PR #2840, opened 2024-07-24
Fix initialisation of type parameter jkinds from annotations

jbachurski:
A simple fix for jkind assignment for type variables, which cleans up many error messages and is necessary for higher jkinds to work later down the line.

When type parameters are encountered within `Typetexp.transl_type_var`, their jkind is initialised anew, even when an annotation is available. This PR looks up whether a parameter of the same name had an annotation and uses that instead. 
This is fine as the type variable itself is still initialised as distinct, but we know the jkinds will be consistent across all instances.


lpw25 (approved):
This is sound and improves a lot of error messages.

PR #2841, opened 2024-07-24
Propagate tail call attribute information through to the backend

tylerhou:
Function applications / [tail]call instructions show the annotation when dumping lambda (`-dlambda`), flambda2 (`-dflambda`), cmm (`-dcmm`) mach, (`-dsel`), cfg (`-dcfg`). For example, see https://gist.github.com/tylerhou/9c8bf1728e54b7b8d14f4062cd78fb61


middle_end/flambda2/from_lambda/closure_conversion.ml
```diff
@@ -1755,7 +1757,7 @@ let make_unboxed_function_wrapper acc function_slot ~unarized_params:params
                 ~current_region:my_region))
         Debuginfo.none ~inlined:Inlined_attribute.Default_inlined
         ~inlining_state:(Inlining_state.default ~round:0)
-        ~probe:None ~position:Normal
+        ~probe:None ~tail:Default_tail ~position:Normal
```
tylerhou, line 1760: I don't understand this code well enough to know whether I should propagate some tail annotation from a surrounding scope...

tylerhou: I assume that when we introduce the new type we can put Phantom_function here.

dvulakh: I think I agree that there's no annotation to propagate here.
1. If this call inherited some tailcallness properties from the boxed version, I'd expect `~position` to not be constant.
2. Dropping annotations only makes the analysis more conservative (we might start flagging edges that we forgot were explicitly labelled `[@nontail]`), so either way I don't think we should worry about this too much for now.

If you're still concerned, perhaps leave a CR?

dvulakh (changes requested):
Looks good, thanks! I'll approve after we move the tests into the repo (as discussed).

backend/cfg/cfg_intf.ml
```diff
@@ -191,10 +204,12 @@ module S = struct
     | Switch of Label.t array
     | Return
     | Raise of Lambda.raise_kind
-    | Tailcall_self of { destination : Label.t }
-    | Tailcall_func of func_call_operation
+    | Tailcall_self of destination with_tail_attr
```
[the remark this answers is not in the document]
tylerhou: Added a CR.

middle_end/flambda2/from_lambda/closure_conversion.ml
tylerhou: I'm less concerned now, I just wanted to flag it for closer review. Thanks for taking a look.

tylerhou:
I added the test as discussed, and also fixed the builds on the other register allocators -- which were not failures due to the register allocators themselves. Instead, I broke arm64, which was an easy fix. (The multiarch test is skipped for the normal flambda2 builds for some reason.)

dvulakh (approved):
Looks good, thanks!

PR #2845, opened 2024-07-24
Higher-kinded type application at `Tconstr`

jbachurski:
This shows an attempt of the changes that arise when implementing higher-kinded type application **just** at `Tconstr`. 

See `applied.ml` for some tests.

This also allows use of unapplied datatypes (see `jkind_of_decl_unapplied`).

**Note on instantiation of type constructors.** A common pattern is doing something like - deconstruct `Tconstr (p, tys, _)`, get the `decl : type_declaration` of `p`, then instantiate `p` at arguments initialised based on `decl.type_variance` (or iterate, etc.). The problem is that `tys` might be an `Applied` (non-empty list), white `p` is higher-kinded, in which case `type_variance` might be empty in the declaration (as `type_arity = 0` with `type_jkind` higher-kinded). I tried to address this with getters like `app_variance_of_decl` which explicitly query for what variances _can_ the path (declaration) be applied with.
I'm not sure what should happen for `Tapp` in these cases - but perhaps this only happens for type declarations, the higher-kinded case of which is an extension.

**Note on errors.** Errors are raised in bad applications - there's one bad one which I'm not sure what causes. I remember that adding an explicit jkind check in `estimate_type_jkind`  (further to `List.length tys = List.length args`) actually causes failures building stdlib as some types are preemptively assigned `any`. This causes `tys` to be `any` in a contravariant position (would this work any better with `top`, or do we a 'wildcard' due to the contravariance?), so the subkinding would fail.


ocaml/typing/ctype.ml
```diff
@@ -2091,6 +2110,55 @@ let tvariant_not_immediate row =
...
+  (* FIXME jbachurski: Shouldn't we look at type_variance and type_separability here? *)
+  match decl.type_arity with
+  | 0 -> Some decl.type_jkind
+  | _ when is_datatype_decl decl.type_kind ->
+    Some (Arrow
+    { args = List.map (type_jkind env) decl.type_params
+    ; result = decl.type_jkind })
+  | _ -> None
+
+and type_jkind_for_app (app_jkind : jkind) app_arity tys : jkind option =
```
lpw25, line 2134: I think this would be clearer if `type_jkind_for_app_decl` was replaced by a general `jkind_of_decl_applied`, and the logic for the applied case was all in there directly. `estimate_type_jkind` can then check that the `jkind` is in fact a `type_jkind` itself. (Or possibly it should become `estimate_jkind` and the specialisation to type jkinds pushed even further out).

This function can then be simplified to only handle the arity 0 case for a list of types.

ocaml/typing/ctype.ml
```diff
@@ -2091,6 +2110,55 @@ let tvariant_not_immediate row =
...
+  match tys, app_jkind with
+  | Unapplied, _ when app_arity = 0 -> Some app_jkind
+  | Unapplied, _ -> None
+  | Applied tys, _ when app_arity > 0 ->
+    if List.length tys = app_arity
```
lpw25, line 2139: It's probably fine to just assume the arities are correct here rather than checking them.

ocaml/typing/ctype.ml
```diff
@@ -2143,6 +2205,23 @@ let rec estimate_type_jkind env ty =
   | Tpoly (ty, _) -> estimate_type_jkind env ty
   | Tpackage _ -> Jkind (Type.Primitive.value ~why:First_class_module |> Jkind.of_type_jkind)
 
+and type_jkind env ty =
+  jkind_of_result (estimate_type_jkind env (get_unboxed_type_approximation env ty))
+
+let arity_matches_decl env decl t = match t, decl.type_arity with
```
lpw25, line 2211: I think this now only has one valid call site for this function: the one in `typetexp.ml`. It should possibly be moved in there, renamed `check_arity_matches_decl` and changed to raise the associated error itself directly.

ocaml/typing/ctype.ml
```diff
@@ -5784,7 +5856,7 @@ let rec build_subtype env (visited : transient_expr list)
       if memq_warn tt visited then (t, Unchanged) else
       let visited = tt :: visited in
       begin try
-        let decl = Env.find_type p env in
+        let var = Env.find_type p env |> app_variance_of_decl in
```
lpw25, line 5859: Probably wrong for the unapplied case, causing `map2` to fail below.

ocaml/typing/ctype.ml
```diff
@@ -5951,7 +6023,7 @@ let rec subtype_rec env trace t1 t2 cstrs =
         subtype_rec env trace t1 (expand_abbrev env t2) cstrs
     | (Tconstr(p1, tl1, _), Tconstr(p2, tl2, _)) when Path.same p1 p2 ->
         begin try
-          let decl = Env.find_type p1 env in
+          let var = Env.find_type p1 env |> app_variance_of_decl in
```
lpw25, line 6026: Probably wrong for the `Unapplied` case, causing `fold_left2` to fail below.

lpw25 (commented):
Mostly looks good. Some uses of the `app_foo_of_decl` functions are being used in the `Unapplied` case, which will cause issues. Some code could be simplified.

ocaml/typing/typeopt.ml
```diff
@@ -400,9 +400,10 @@ let rec value_kind env ~loc ~visited ~depth ~num_nodes_visited ty
       let decl =
```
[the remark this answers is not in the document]
jbachurski: Sorry, I don't think I follow. Are you saying this should also handle `Tapp`, or that there should be some `expand_head`?

ocaml/typing/ctype.ml
jbachurski: I think I see why it'd be okay to drop the check, but it'd make me more uncomfortable if it wasn't there. Is there a reason to drop it beside performance?

jbachurski:
I _think_ all the regions with cases of `type_variance` in unapplied cases are marked TODO and have a test case. I'm not sure I know how to fix them at the moment beyond leaving a TODO. (especially [`mcomp_type_decl`](https://github.com/ocaml-flambda/flambda-backend/pull/2844#discussion_r1732105477) is weird, as it has some legal `Unapplied`/`Applied` running already).

ocaml/typing/printtyp.ml
```diff
@@ -2367,7 +2367,7 @@ let dummy =
     type_params = [];
     type_arity = 0;
     type_kind = Type_abstract Abstract_def;
-    type_jkind = Jkind.Primitive.top ~why:Dummy_jkind;
+    type_jkind = Jkind.Type.Primitive.any ~why:Dummy_jkind |> Jkind.of_type_jkind;
```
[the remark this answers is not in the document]
jbachurski: You're right, these two seem fine based on introducing the change down the line. 

I'm not happy with the initial choice of setting `top` prematurely, as this must have left me confused here (likely a rough implementation of `top` left this broken). It was harder to test and review, as the `top`s were in-place before they did anything. Oh well.

ocaml/typing/subst.ml
```diff
@@ -289,7 +289,7 @@ let rec typexp copy_scope s ty =
     let has_fixed_row =
       not (is_Tconstr ty) && is_constr_row ~allow_ident:false tm in
     (* Make a stub *)
-    let jkind = Jkind.Primitive.top ~why:Dummy_jkind in
+    let jkind = Jkind.Type.Primitive.any ~why:Dummy_jkind |> Jkind.of_type_jkind in
```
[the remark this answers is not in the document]
jbachurski: As above.

jbachurski:
Subsumed by #3027

PR #2847, opened 2024-07-24
Fix ocamlformat CI

dvulakh:
The ocamlformat CI doesn't actually detect poorly formatted files in the flambda2 and cfg codebases.


dvulakh:
I thought I saw a commit pass that should not have, but can't replicate, so closing now.

PR #2849, opened 2024-07-25
Polling: more tests

xclerc:
This pull request is adding a couple of
test for the polling insertion phase, and
to do so modifies the way polling errors
are reported.

### DIFF
--- a/ocaml/testsuite/tests/polling/polling.ml
+++ b/ocaml/testsuite/tests/polling/polling.ml
@@ -13,7 +13,7 @@ let[@poll error] loop_alloc x =
 
 let[@poll error] loop_no_alloc x =
   for _ = 0 to x do
-    ()
+    ignore(Sys.opaque_identity(ref 42))
   done
 ;;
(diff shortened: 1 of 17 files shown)


ocaml/testsuite/tests/polling/polling.ml
```diff
@@ -0,0 +1,27 @@
...
+
+let[@poll error] loop_alloc x =
+  for _ = 0 to x do
+    ignore(Sys.opaque_identity(ref 42))
+  done
+;;
+
+let[@poll error] loop_no_alloc x =
+  for _ = 0 to x do
+    ()
```
mshinwell, line 16: One day we might be able to detect that this loop can be deleted, so I would put something like `ignore (Sys.opaque_identity 42)` in the middle of it.

mshinwell (approved):
Just one thing to fix, then please merge

gretay-js:
@xclerc  should we merge this?

PR #2851, opened 2024-07-25
Use map for finding signature item; add classes to substitution

smjleo:
Defines a new `Sig_map` and uses this to find modules, module types, etc instead of iterating through signature (which is costly, especially since we need to do this for every signature item).

### DIFF
--- a/ocaml/typing/typemod.ml
+++ b/ocaml/typing/typemod.ml
@@ -113,6 +113,71 @@ type modtype_decl_context =
   (** We are checking a struct, with a given expected signature (used for inferring
       module types when we see a [module type S = _]) *)
 
+(** Map of items in a signature (similar to an [Env.t] after [Env.add_signature]), but
+    only indexed by the item's name as a string. This avoids us having to iterate through
+    an [expected_sig] to find the declaration we are looking for in [type_str_item]. *)
+module Sig_map = struct
+  type t = {
+    values: (Ident.t * value_description) String.Map.t;
+    types: (Ident.t * type_declaration) String.Map.t;
+    modules: (Ident.t * module_declaration) String.Map.t;
+    module_types: (Ident.t * modtype_declaration) String.Map.t;
+    classes: (Ident.t * class_declaration) String.Map.t;
+    class_types: (Ident.t * class_type_declaration) String.Map.t;
+  }
+
+  let empty = {
+    values = String.Map.empty;
+    types = String.Map.empty;
+    modules = String.Map.empty;
+    module_types = String.Map.empty;
+    classes = String.Map.empty;
+    class_types = String.Map.empty;
+  }
+
+  let add_value id value map =
+    { map with values = String.Map.add (Ident.name id) (id, value) map.values }
+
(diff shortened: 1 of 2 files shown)


ocaml/typing/typemod.ml
```diff
@@ -3354,7 +3419,8 @@ and type_structure ?(toplevel = None) ~expected_sig funct_body anchor env sstr =
              classes []),
         shape_map,
         new_env,
-        subst
+        add_expected_class_type_to_subst expected_sig clty_ids ty_ids subst
+
```
riaqn, line 3423: empty line

ocaml/typing/typemod.ml
```diff
@@ -3298,15 +3362,14 @@ and type_structure ?(toplevel = None) ~expected_sig funct_body anchor env sstr =
...
-          in
           match expected_sig with
           | None -> In_structure None
-          | Some sg -> In_structure (List.find_map get_modtype_decl sg)
+          | Some expected_sig ->
+            begin match Sig_map.find_module_type pmtd.pmtd_name.txt expected_sig with
+              | None -> In_structure None
+              | Some (_, decl) ->
+                  In_structure (Some (Subst.modtype_declaration Keep subst decl))
+            end
```
riaqn, line 3372: minor formatting nit: insert two extra space indentation

riaqn (approved):
Looks good - feel free to merge after fixing the minor formatting nits.

ocaml/typing/typemod.ml
riaqn: Also, might be worthwhile commenting at the definition of `In_structure` that the expected modtype decl has gone through substitution.

ocaml/typing/typemod.ml
```diff
@@ -113,6 +113,71 @@ type modtype_decl_context =
...
+      | Sig_type (id, decl, _, _) -> add_type id decl map
+      | Sig_typext _ -> map
+      | Sig_module (id, _, decl, _, _) -> add_module id decl map
+      | Sig_modtype (id, decl, _) -> add_module_type id decl map
+      | Sig_class (id, decl, _, _) -> add_class id decl map
+      | Sig_class_type (id, decl, _, _) -> add_class_type id decl map
+    in
+    List.fold_left add_item map sg
+
+  let _find_value name map = String.Map.find_opt name map.values
```
riaqn, line 168: Let's remove this function if it's not used. Similar for `_find_class`.

ocaml/typing/typemod.ml
smjleo: Sorry, I'm not sure what you mean about the spacing - where should I add the spaces? The file seems to use a number of conflicting conventions.

riaqn: I meant to add two extra space indendation to the quoted five lines of code.

You are right that the file is not consistent in this regards - and I'm guilty for some of that. Regardless, new code should try to follow the rules.

smjleo: Ah I see now - thanks!

PR #2852, opened 2024-07-25
Propagate original fn application position through the typetree into the backend

tylerhou:
We record the original function application position on the translation from Parsetree to Typetree in the same place where we infer `Nontail`, `Tail`, or `Default` application. `flambda2` also contains a `position` field on its `apply` record; from `flambda2` onwards we call the new record field `original_position` to distinguish it from the application position post optimization.


ocaml/testsuite/tests/typing-tailcalls/propagate_original_position.ml
dvulakh, line 1: Now that we also propagate whether the call was in tail position to begin with, it might be good to add a version of each test where the `foo` call is not syntactically tail (e.g. via `let x = foo () [@tail hint] in x`).

dvulakh (changes requested):
Looks good overall! Looking forward to seeing the graph analysis.

I only double-checked a handful of the incidental changes to surrounding tests, but they looked right.

ocaml/testsuite/tests/typing-tailcalls/propagate_original_position.ml
tylerhou: Done.

dvulakh (approved):

PR #2855, opened 2024-07-26
Make all register allocation parameters optional

xclerc:
As per title.

(Having mandatory parameters is useful for
the developer, ensuring everything is explicit
set, but of course this instead puts the burden
on the user.)

### DIFF
--- a/backend/regalloc/regalloc_gi_utils.ml
+++ b/backend/regalloc/regalloc_gi_utils.ml
@@ -49,6 +49,8 @@ module Priority_heuristics = struct
     | Interval_length
     | Random_for_testing
 
+  let default = Interval_length
+
   let all = [Interval_length; Random_for_testing]
 
   let to_string = function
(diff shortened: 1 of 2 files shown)


backend/regalloc/regalloc_gi_utils.ml
```diff
@@ -86,6 +84,8 @@ module Selection_heuristics = struct
     | Worst_fit
     | Random_for_testing
 
+  let default = First_available
+
   let all = [First_available; Best_fit; Worst_fit; Random_for_testing]
```
milan-tom, line 89: Is there any way we can use `[@@deriving enumerate]` here?

milan-tom (approved):
LGTM

backend/regalloc/regalloc_gi_utils.ml
xclerc: Unfortunately, we cannot use such
ppxes in the compiler.

PR #2856, opened 2024-07-26
Add Analyze_tailcall_graph

tylerhou:
Add the `Analyze_tailcall_graph` module, which stores a graph of function calls and categorizes edges according to their relevance to reducing TCO. We introduce two new flags: `-[no-]cfg-analyze-tailcalls` and `-d[no-]cfg-tailcalls`. `-[no-]cfg-analyze-tailcalls` turns on the analysis pass (which will be updated to emit warnings when newly inferred nontails are in some previously fully tail cycle). `-d[no-]cfg-tailcalls` dumps the function call graph to dot for visualization. Here is an example of such a graph:

image


backend/analyze_tailcall_graph.ml
```diff
@@ -0,0 +1,385 @@
...
+    include T
+
+    module Tbl = Hashtbl.Make (struct
+        include T
+      end)
+  end
+
+  type t =
+    { vertex_by_name : Vertex.t String.Tbl.t
+    ; (* Only need to store the edges / vertices for the Known_fn case. *)
```
tylerhou, line 177: This is out of date

backend/analyze_tailcall_graph.ml
```diff
@@ -0,0 +1,385 @@
...
+      adjacencies : unit Edge.Tbl.t Vertex.Tbl.t
+    }
+
+  (* This is nice so we don't have to check for the empty case when adding an edge from
```
tylerhou, line 181: Mildly out of date as well

backend/analyze_tailcall_graph.ml
```diff
@@ -0,0 +1,416 @@
...
+
+    let to_dot_label t =
+      match t with
+      | Unknown_fn -> "<unknown>"
+      | Known_fn { name; _ } -> name
+    ;;
+
+    module Tbl = Hashtbl.Make (struct
+        include T
+      end)
```
dvulakh, line 156: Why not `Hashtbl.Make (T)`? Same below.

backend/analyze_tailcall_graph.ml
```diff
@@ -0,0 +1,416 @@
...
+        | _ -> false
+      ;;
+
+      let equal e1 e2 =
+        label_equal e1.label e2.label
+        && Vertex.equal e1.from e2.from
+        && Vertex.equal e1.to_ e2.to_
+      ;;
+
+      let hash e1 = Vertex.hash e1.from + Vertex.hash e1.to_
```
dvulakh, line 190: Nit: is there a reason we exclude the label from the hash? It is possible to have two edges with the same source and destination, but different labels.

Also, if we end up making a `hash_label` function, we might want to pull the `label` type and its utilities to a submodule.

ocaml/testsuite/tests/typing-tailcalls/dcfg_tailcalls.ml
dvulakh, line 1: Could we also add a test of two edges with different labels between the same pair of vertices? (i.e. a function `f` that calls `g` in tail position in two different ways)

dvulakh, line 1: If you encountered any examples of your "Unknown" case --- like calls to `tuplify` that would show up in the graph --- could you include one here?

dvulakh (changes requested):

backend/analyze_tailcall_graph.ml
tylerhou: Because of refactoring (previously I only defined hash in this anonymous struct, and then I pulled it out.)

backend/analyze_tailcall_graph.ml
tylerhou: I forgot that there could be multiple edges with different labels.

Added `Label` struct.

backend/analyze_tailcall_graph.ml
```diff
@@ -0,0 +1,416 @@
...
+    match from with
+    | Unknown_fn ->
+      (* An edge already exists from Unknown_fn (added when the vertex was created) *)
+      ()
+    | Known_fn _ ->
+      let edges = Vertex.Tbl.find t.adjacencies from in
+      Edge.Hashset.add edges edge
+  ;;
+
+  let replace ~c ~with_ str = String.split_on_char c str |> String.concat with_
```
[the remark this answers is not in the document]
tylerhou: Deleted -- I previously used it to mangle function names for dot, but we no longer need to that since we assign integer IDs to vertices.

ocaml/testsuite/tests/typing-tailcalls/dcfg_tailcalls.ml
tylerhou: Done.

tylerhou: Wasn't able to add for `tuplify`, but added one for a closure application.

dvulakh (approved):
Looks good, thanks!

backend/analyze_tailcall_graph.ml
dvulakh, line 1: Let's add
```
[@@@ocaml.warning "+a-30-40-41-42"]
```
to the top of this file (we do this in other files that we've added to the backend).

dvulakh (changes requested):

backend/analyze_tailcall_graph.ml
```diff
@@ -0,0 +1,419 @@
...
+      | Some vertex -> vertex
+    in
+    vertex
+  ;;
+
+  let add_edge
+    t
+    ~(from : Vertex.t)
+    ~(to_ : Vertex.t)
+    ~(actual_position : [ `Tail | `Nontail ])
```
dvulakh, line 258: Let's avoid the polymorphic variant here. I think I would prefer we either mint a new variant type or using `is_tail_position:bool`.

backend/analyze_tailcall_graph.ml
```diff
@@ -0,0 +1,419 @@
...
+      | `Unknown_nontail -> Explicit_nontail_edge
+      | `Not_in_tail_position -> Explicit_nontail_edge
+      | `Requested_nontail -> Explicit_nontail_edge
+      | `Requested_tail -> Explicit_tail_edge
+      | `Became_tail_after_optimizations ->
+        (* This is a conservative approximation *)
+        Explicit_tail_edge
+      | `Inferred_tail -> Inferred_tail_edge
+      | `Inferred_nontail -> Inferred_nontail_edge
+    in
```
dvulakh, line 311: I would produce the `Edge.Label.t` directly in the first match, and use comments instead of a polymorphic variant to explain anything that seems tricky.

dvulakh (changes requested):

backend/analyze_tailcall_graph.ml
dvulakh, line 1: Let's also add this file and its interface to `backend/.ocamlformat-enable`.

dvulakh (changes requested):

backend/analyze_tailcall_graph.ml
tylerhou: Added variant type.

backend/analyze_tailcall_graph.ml
tylerhou: Done.

backend/analyze_tailcall_graph.ml
tylerhou: Done.

backend/analyze_tailcall_graph.ml
```diff
@@ -0,0 +1,350 @@
...
+          ~case:"Tail_position Hint_tail, Nontail"
+      | Tail_position Explicit_non_tail, Tail
+      | Not_tail_position Explicit_non_tail, Tail ->
+        impossible_because "[@nontail] was optimized to a tailcall"
+          ~case:"_ Explicit_non_tail, Tail"
+      | Tail_position (Explicit_tail | Hint_tail), Tail ->
+        (* Requested tail *) Explicit_tail_edge
+      | Tail_position Default_tail, Tail -> Inferred_tail_edge
+      | Tail_position Explicit_non_tail, Nontail ->
+        (* Requested tail *) Explicit_nontail_edge
```
dvulakh, line 252: Requested nontail?

dvulakh (approved):
Looks good, thanks!

dvulakh (approved):

PR #2864, opened 2024-07-30
Higher-kinded type application

jbachurski:
The (expected) end of the 'typing higher-kinded type application' feature, implementing the final type inference logic around `Tapp`. The flagship example of what this feature adds is in `monad_cons.ml`:

```ocaml
type ('m : value => value) monad = {
  return : 'a. 'a -> 'a 'm;
  bind : 'a 'b. 'a 'm -> ('a -> 'b 'm) -> 'b 'm
}

let list = {
  return = (fun x -> [x]);
  bind = (fun xs f -> List.concat_map f xs)
}

let prod m x y = m.bind x (fun a -> m.bind y (fun b -> m.return (a, b)))
let result = prod list [-1; 1] [0; 1; 2]
(* val result : (int * int) list = [(-1, 0); (-1, 1); (-1, 2); (1, 0); (1, 1); (1, 2)] *)
```

Crux of the logic changes are in `Ctype` - implementing unification etc. The key invariant is that we take `Tconstr(p, args) = Tapp(Tconstr(p, []), args)`, while attempting not to construct `Tapp(_, [])`.

It is possible that this implementation is incomplete, as there are many cases where `Tapp` might not be handled where it should be.

### DIFF
--- a/ocaml/testsuite/tests/typing-gadts/pr6158.ml
+++ b/ocaml/testsuite/tests/typing-gadts/pr6158.ml
@@ -21,5 +21,5 @@ Line 6, characters 45-49:
 Error: This pattern matches values of type (int s, int s) eq
        but a pattern was expected which matches values of type
          (int s, int t) eq
-       Type int s is not compatible with type int t
+       Type s is not compatible with type t
 |}];;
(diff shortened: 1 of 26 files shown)


jbachurski:
Subsumed by #3027

PR #2868, opened 2024-07-30
Decompose the tail call graph into strongly connected components

tylerhou:
For the purpose of the current analysis, we ignore non-inferred nontail edges (these are edges explicitly marked as `[@nontail]`, and function applications that were not in tail position and also not optimized to be tail calls). The dot output has been changed to show SCCs in clusters, and also highlight clusters that have at least one inferred nontail edge (red edges).

image


dvulakh (changes requested):
Let's also think about the testing & debugging story before merging.

tylerhou:
The only failure was due to timeout, it might be flaky? Rerunning.

dvulakh (approved):
Looks good, thanks!

PR #2869, opened 2024-07-30
Add `[@@or_null_reexport]`

dkalinichenko-js:
Add `[@@or_null_reexport]`, an attribute for re-exporting the constructors of `'a or_null`, because they can't be defined by users.

Between `type 'a t [@@or_null_reexport]` and `type 'a t = 'a or_null [@@or_null_reexport]` this PR chooses the latter to avoid unnecessary work in `typedecl.ml`. With the manifest set by user, we can delay updating `type_kind` until relatively late in typechecking.

Also fix `or_null.mli` to use `'a t` instead of `'a or_null`.

### DIFF
--- a/ocaml/testsuite/tests/typing-layouts-or-null/reexport.ml
+++ b/ocaml/testsuite/tests/typing-layouts-or-null/reexport.ml
@@ -3,13 +3,16 @@
  expect;
 *)
 
-(* CR layouts v3: ['a or_null] can't be re-exported normally,
+(* CR layouts v3.5: ['a or_null] can't be re-exported normally,
    because users can't define their own [Null]-like constructors. *)
 module Or_null = struct
   type ('a : value) t : value_or_null = 'a or_null =
     | Null
     | This of 'a
 end
+
+(* CR layouts v3: this error message is not great, but it will be a
+   different error message in the final PR. *)
 
 [%%expect{|
 Lines 2-4, characters 2-16:
@@ -181,8 +184,23 @@ Line 1, characters 0-30:
 1 | type 'a t [@@or_null_reexport]
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 Error: Invalid reexport declaration.
-       Type t must be defined to be equal to or_null
-       with no explicit representation.
+       Type t must be defined equal to the primitive type or_null.
+|}]
+
+(* CR layouts v3: This would be nice to accept, but it's somewhat complicated
+   to implement. So we won't unless we encounter a use-case. *)
+module M : sig
+  type 'a t [@@or_null_reexport]
+end = struct
+  type 'a t = 'a or_null [@@or_null_reexport]
+end
+
+[%%expect{|
+Line 2, characters 2-32:
+2 |   type 'a t [@@or_null_reexport]
+      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Error: Invalid reexport declaration.
+       Type t must be defined equal to the primitive type or_null.
 |}]
 
 (* [@@or_null_reexport] forbids explicit representation. *)
@@ -194,8 +212,7 @@ Line 1, characters 0-63:
 1 | type 'a t = 'a or_null = Null | This of 'a [@@or_null_reexport]
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 Error: Invalid reexport declaration.
-       Type t must be defined to be equal to or_null
-       with no explicit representation.
+       Type t must not define an explicit representation.
 |}]
 
 (* [@@or_null_reexport] requires the type to be equal to ['a null]. *)
@@ -207,8 +224,7 @@ Line 1, characters 0-42:
 1 | type 'a t = 'a option [@@or_null_reexport]
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 Error: Invalid reexport declaration.
-       Type t must be defined to be equal to or_null
-       with no explicit representation.
+       Type t must be defined equal to the primitive type or_null.
 |}]
 
 (* [@@or_null_reexport] behaves sanely in corner cases. *)
@@ -231,4 +247,40 @@ Error: This variant or record definition does not match that of type
          (int * 'a) or_null
        Their parameters differ:
        The type int * 'a is not equal to the type 'a
+|}]
+
+(* [@@or_null_reexport] expands the type on the right. *)
+
+type 'a t1 = 'a or_null [@@or_null_reexport]
+
+type 'a t2 = 'a t1 [@@or_null_reexport]
+
+[%%expect{|
+type 'a t1 = 'a or_null = Null | This of 'a
+type 'a t2 = 'a t1 = Null | This of 'a
+|}]
+
+(* [@@or_null_reexport] handles shadowing correctly. *)
+
+type 'a or_null : value = Null | This of 'a
+
+type 'a t = 'a or_null [@@or_null_reexport]
+
+(* CR layouts v3: this error message is somewhat confusing. *)
+
+[%%expect{|
+type 'a or_null = Null | This of 'a
+Line 3, characters 0-43:
+3 | type 'a t = 'a or_null [@@or_null_reexport]
+    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Error: Invalid reexport declaration.
+       Type t must be defined equal to the primitive type or_null.
+|}]
+
+(* Misplaced attribute warnings are not printed in toplevel. *)
+
+let[@or_null_reexport] foo = 5
+
+[%%expect{|
+val foo : int = 5
 |}]
--- a/ocaml/testsuite/tests/typing-layouts-or-null/reexport.ml
+++ b/ocaml/testsuite/tests/typing-layouts-or-null/reexport.ml
@@ -284,6 +284,33 @@ Error: In this definition, expected parameter variances are not satisfied.
        but it is injective covariant.
 |}]
 
+(* The type's arity must be exactly the same. *)
+
+type ('a, 'b) t = 'b or_null [@@or_null_reexport]
+
+[%%expect{|
+Line 1, characters 0-49:
+1 | type ('a, 'b) t = 'b or_null [@@or_null_reexport]
+    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Error: This variant or record definition does not match that of type
+         'b or_null
+       They have different arities.
+|}]
+
+(* The type parameter must be actually used. *)
+
+type 'a t = int or_null [@@or_null_reexport]
+[%%expect{|
+Line 1, characters 0-44:
+1 | type 'a t = int or_null [@@or_null_reexport]
+    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Error: This variant or record definition does not match that of type
+         int or_null
+       Their parameters differ:
+       The type int is not equal to the type 'a
+|}]
+
+
 (* [@@or_null_reexport] handles shadowing correctly. *)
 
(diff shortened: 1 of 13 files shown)


ocaml/testsuite/tests/typing-layouts-or-null/reexport.ml
```diff
@@ -0,0 +1,234 @@
...
+
+[%%expect{|
+Lines 2-4, characters 2-16:
+2 | ..type ('a : value) t : value_or_null = 'a or_null =
+3 |     | Null
+4 |     | This of 'a
+Error: The kind of type 'a or_null is value_or_null
+         because it is the primitive value_or_null type or_null.
+       But the kind of type 'a or_null must be a subkind of value
+         because of the definition of t at lines 2-4, characters 2-16.
```
goldfirere, line 22: This is a poor error message. It should say something about the fact that variants are values. Doesn't necessarily have to be fixed in this patch, but if not, leave a CR to return to this.

ocaml/testsuite/tests/typing-layouts-or-null/reexport.ml
```diff
@@ -0,0 +1,234 @@
...
+(* [@@or_null_reexport] re-exports those constructors. *)
+
+module Or_null = struct
+  type ('a : value) t : value_or_null = 'a or_null [@@or_null_reexport]
+end
+let n = Or_null.Null
+let t v = Or_null.This v
+
+[%%expect{|
+module Or_null : sig type 'a t = 'a or_null = Null | This of 'a end
```
goldfirere, line 62: Printing out with `[@@or_null_reexport]` would be cool, but very low priority if it's at all difficult.

ocaml/testsuite/tests/typing-layouts-or-null/reexport.ml
```diff
@@ -0,0 +1,234 @@
...
+1 | type t = int or_null [@@or_null_reexport]
+    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Error: This variant or record definition does not match that of type
+         int or_null
+       They have different arities.
+|}]
+
+(* [@@or_null_reexport] requires a manifest. *)
+
+type 'a t [@@or_null_reexport]
```
goldfirere, line 177: It would be cool to accept this, for the following idiom:

```ocaml
module M : sig
  type 'a t [@@or_null_reexport]
end = struct
  type 'a t = 'a or_null [@@or_null_reexport]
end
```

Now, `M.t` behaves just like `or_null` but is still distinct. Which might be useful. That said, you've described this as hard, so we shouldn't. So all of this is to say: leave a comment here saying this would be nice, but we'll wait for an actual use-case.

ocaml/typing/typedecl.ml
```diff
@@ -1094,6 +1098,54 @@ let check_constraints env sdecl (_, decl) =
       check_constraints_rec env sty.ptyp_loc visited ty
   end
 
+(* CR layouts v3: this is a hack to allow re-exporting the definition
```
goldfirere, line 1101: I think this isn't quite v3, because we don't really have a plan for allowing user-written null constructors. Maybe make a v3.5 for this? I'm not sure we'll ever do it -- would need to study the level of need.

ocaml/typing/typedecl.ml
```diff
@@ -1094,6 +1098,54 @@ let check_constraints env sdecl (_, decl) =
...
+   of ['a or_null], including constructors, even if one can't define
+   ['a or_null] "honestly", with tools available to users.
+
+   Remove when we allow users to define their own null constructors.
+*)
+let reexport_special_types env loc dpath decl =
```
goldfirere, line 1107:
```suggestion
let add_special_type_reexports env loc dpath decl =
```

ocaml/typing/typedecl.ml
```diff
@@ -1094,6 +1098,54 @@ let check_constraints env sdecl (_, decl) =
...
+  in
+  match or_null_reexport with
+  | false -> decl
+  | true ->
+    let fail () =
+      raise (Error (loc, Invalid_reexport
+        { definition = dpath; expected = Predef.path_or_null }))
+    in
+    match decl.type_kind, decl.type_manifest with
+    (* We require users to define ['a t = 'a or_null]. By assumption,
```
goldfirere, line 1119: What does "By assumption" mean here? I think just leave that out.

ocaml/typing/typedecl.ml
```diff
@@ -1094,6 +1098,54 @@ let check_constraints env sdecl (_, decl) =
...
+       they can't define the valid type kind, so it must be left abstract.
+       Manifest must be set to [or_null] so typechecking stays correct. *)
+    | Type_abstract _, Some ty ->
+      let param =
+        match get_desc ty with
+        | Tconstr(path, [param], _)
+          when Path.same path Predef.path_or_null -> param
```
goldfirere, line 1126: This doesn't allow expansion. That is, I don't think I could say

```ocaml
type 'a t1 = 'a or_null [@@or_null_reexport]
type 'a t2 = 'a t1 [@@or_null_reexport]
```

That's a tiny bit not cool, because it means there is no way to re-export `or_null` if `or_null` is shadowed. But I think the solution is easy: just `expand_head` before the `get_desc`. (And add this to the tests. Along with tests for when `or_null` is shadowed.)

If this turns out to be difficult, then it's OK not to do this. But I really do think it will be easy.

ocaml/typing/typedecl.ml
```diff
@@ -1094,6 +1098,54 @@ let check_constraints env sdecl (_, decl) =
...
+        | _ -> fail ()
+      in
+      let or_null_decl = Env.find_type Predef.path_or_null env in
```
goldfirere, line 1129: Maybe I'm wrong, but it seems easier to me just to construct the right kind from scratch instead of taking `or_null`'s kind and altering it. Make a little function do this and exporting it from predef will avoid code duplication.

ocaml/testsuite/tests/typing-layouts-or-null/reexport.ml
goldfirere, line 1: Also please add tests for good and bogus injectivity/variance annotations.

ocaml/typing/typedecl.ml
```diff
@@ -1155,6 +1207,7 @@ let check_kind_coherence env loc dpath decl =
   | _ -> ()
 
 let check_coherence env loc dpath decl =
+  let decl = reexport_special_types env loc dpath decl in
```
goldfirere, line 1210: Why do this in `check_coherence`? I was expecting this to appear in `transl_declaration` directly, but maybe there's a good reason for it to be here.

ocaml/testsuite/tests/typing-layouts-or-null/reexport.ml
goldfirere, line 1: And a test or two with the attribute in the wrong place, just to see if the warning gets printed.

ocaml/testsuite/tests/typing-layouts-or-null/reexport.ml
dkalinichenko-js: Will be possibly done in later patches with `Variant_with_null`.

ocaml/typing/typedecl.ml
dkalinichenko-js: Done.

ocaml/typing/typedecl.ml
dkalinichenko-js: We must re-export `or_null` after `check_constraints`, because `check_constraints` expects `type_kind` to match the one in the parsetree.

ocaml/testsuite/tests/typing-layouts-or-null/reexport.ml
dkalinichenko-js: See `reexport_attr_unused.ml`.

dkalinichenko-js: Done.

ocaml/testsuite/tests/typing-layouts-or-null/reexport.ml
goldfirere: OK. Maybe leave a CR here noting that this output is not what we might want. We can always delete the CR later.

ocaml/typing/typedecl.ml
goldfirere: Firstly, that's terrible. `check_constraints` is pretty sad. But I suppose we don't need to fix it today.

But we can just skip this. An `or_null` re-export doesn't need to check constraints: I don't think there's any way for it to fail. `check_constraints` is needed when you have mutually recursive type declarations where one type declares constraints that the other types might or might not respect. That's just not the case for `or_null` re-exports: we know, for sure, that neither the manifest nor the kind mentions any of the types brought into scope within this same group. So it's all OK.

My suggestion: check for the `[@@or_null_reexport]` attribute in `check_constraints` and just do nothing if that attribute is present.

ocaml/typing/typedecl.ml
```diff
@@ -1098,53 +1099,39 @@ let check_constraints env sdecl (_, decl) =
...
   end
 
-(* CR layouts v3: this is a hack to allow re-exporting the definition
+(* CR layouts v3.5: this is a hack to allow re-exporting the definition
    of ['a or_null], including constructors, even if one can't define
    ['a or_null] "honestly", with tools available to users.
 
    Remove when we allow users to define their own null constructors.
+
+   [reexport_special_types] must happen after [check_constraints]
```
goldfirere, line 1108:
```suggestion
   [add_special_type_reexports] must happen after [check_constraints]
```

ocaml/testsuite/tests/typing-layouts-or-null/reexport.ml
```diff
@@ -232,3 +248,90 @@ Error: This variant or record definition does not match that of type
...
+|}]
+
+
+(* [@@or_null_reexport] handles shadowing correctly. *)
+
+type 'a or_null : value = Null | This of 'a
+
+type 'a t = 'a or_null [@@or_null_reexport]
+
+(* CR layouts v3: this error message is somewhat confusing. *)
```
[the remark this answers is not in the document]
dkalinichenko-js: Did not help :(

ocaml/typing/typedecl.ml
dkalinichenko-js: Done. Regarding shapes, I ended up modifying `shape_map_cstrs` to accept `Types.constructor_declaration`, since it has all necessary information.

ocaml/testsuite/tests/typing-layouts-or-null/reexport.ml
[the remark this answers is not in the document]
goldfirere: Try `Printtyp.type_path` instead, maybe?

ocaml/typing/typedecl.ml
```diff
@@ -1041,6 +1074,11 @@ let check_constraints env sdecl (_, decl) =
     sdecl.ptype_params decl.type_params;
   begin match decl.type_kind with
   | Type_abstract _ -> ()
+  (* We skip this check because with [or_null_reexport] the [type_kind]
+     does not match [sdecl.ptype_kind]. This is sound, since re-exporting
+     can't introduce new variables in the kind. *)
+  | Type_variant _ when
+    Builtin_attributes.has_or_null_reexport decl.type_attributes -> ()
```
goldfirere, line 1081: Annoyingly, this will foul the unused-attribute warning, because this function marks the attribute as used. You could get around this by matching for `Ptype_abstract` along with `Type_variant`. Or make a `check_or_null_reexport` that looks for the attribute but doesn't mark it as used? Or maybe just keep the check in `transl_declaration` that I said to remove, to prevent code from reaching here.

dkalinichenko-js: Let's just keep the check.

ocaml/testsuite/tests/typing-layouts-or-null/reexport.ml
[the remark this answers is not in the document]
dkalinichenko-js: It has the same result.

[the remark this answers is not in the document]
goldfirere: :shrug:

goldfirere (approved):

PR #2872, opened 2024-07-30
Backport upstream PR 13001

liam923:
Backport [PR 13001](https://github.com/ocaml/ocaml/pull/13001) from upstream. See the original PR's description for an explanation of these changes.


ocaml/testsuite/tests/shape-index/index_aliases.reference
```diff
@@ -1,17 +1,34 @@
 Indexed shapes:
```
liam923, line 1: If comparing this file to upstream, you'll notice that the line numbers are off. This is expected; the commend at the start of `index_aliases.ml` is formatted differently between the two files, which is the cause for the difference.

goldfirere (approved):

PR #2873, opened 2024-07-31
Clean up arguments for module type inference boundness checking

smjleo:
Small cosmetic PR to swap some arguments around to pass in the `str_map` before `sig_map`, and also make both named. This makes it more in line with the changed `check_*` functions, which in turn were changed to be more in line with `Includemod`.


riaqn (approved):
LGTM

PR #2876, opened 2024-07-31
Improve the modes in `predef.ml`

riaqn:
This PR improves the modes in `predef.ml`. Please review by commits:

- The first commit adds the interpretation of more mode annotation, to allow tests
- The second commit adds the jkind `mutable_data`.
- The third commit fixes up `predef.ml` in the following ways:
  - Add `~jkind` to allow more mode crossing.
  - Add `~jkind_annotation` when the corresponding `~jkind` is not `value`
  - Add `~param_jkind` to `iarray`, to require elements of `iarray` to be non-null.
  - Also some reformatting to avoid long lines.

### DIFF
--- a/ocaml/testsuite/tests/typing-modal-kinds/basics.ml
+++ b/ocaml/testsuite/tests/typing-modal-kinds/basics.ml
@@ -699,6 +699,25 @@ val take_strong_mutable_data : 'a @ portable -> unit = <fun>
 |}]
 
 (* mode crossing on the right *)
+let ref_immutable_data_right x =
+  take_strong_immutable_data (weaken_immutable_data x : float ref);
+[%%expect{|
+Line 2, characters 30-53:
+2 |   take_strong_immutable_data (weaken_immutable_data x : float ref);
+                                  ^^^^^^^^^^^^^^^^^^^^^^^
+Error: This value is once but expected to be many.
+|}]
+
+let ref_immutable_data_left x =
+  let x : float ref = weaken_immutable_data x in
+  take_strong_immutable_data x
+[%%expect{|
+Line 3, characters 29-30:
+3 |   take_strong_immutable_data x
+                                 ^
+Error: This value is once but expected to be many.
+|}]
+
 let float_immutable_data_right x =
   take_strong_immutable_data (weaken_immutable_data x : float);
 [%%expect{|
--- a/ocaml/testsuite/tests/typing-modal-kinds/basics.ml
+++ b/ocaml/testsuite/tests/typing-modal-kinds/basics.ml
@@ -754,6 +754,12 @@ let float32_immutable_data_right x =
 val float32_immutable_data_right : float32 -> unit = <fun>
 |}]
 
+let int64x2_immutable_data_right x =
+  take_strong_immutable_data (weaken_immutable_data x : int64x2)
+[%%expect{|
+val int64x2_immutable_data_right : int64x2 -> unit = <fun>
+|}]
+
 let floatarray_mutable_data_right x =
   take_strong_mutable_data (weaken_mutable_data x : floatarray)
 [%%expect{|
@@ -807,6 +813,13 @@ let float32_immutable_data_left x =
   take_strong_immutable_data x
 [%%expect{|
 val float32_immutable_data_left : float32 -> unit = <fun>
+|}]
+
+let int64x2_immutable_data_left x =
+  let x : int64x2 = weaken_immutable_data x in
+  take_strong_immutable_data x
+[%%expect{|
+val int64x2_immutable_data_left : int64x2 -> unit = <fun>
 |}]
 
 let floatarray_mutable_data_left x =
--- a/ocaml/testsuite/tests/typing-modal-kinds/basics.ml
+++ b/ocaml/testsuite/tests/typing-modal-kinds/basics.ml
@@ -680,9 +680,6 @@ let foo : (string -> string) -> (string -> string) @ unique
 val foo : (string -> string) -> unique_ (string -> string) = <fun>
 |}]
 
-(* CR layouts: maybe use the helper function below to refactor the tests above,
-   as well as [expected_mode.ml]. *)
-
 let weaken_immutable_data : 'a -> 'a @ contended once nonportable =
   fun a -> a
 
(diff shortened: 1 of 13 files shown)


ocaml/typing/jkind.ml
```diff
@@ -404,6 +404,22 @@ module Const = struct
...
+              { linearity = Linearity.Const.min;
+                contention = Contention.Const.max;
+                portability = Portability.Const.min;
+                uniqueness = Uniqueness.Const.max;
+                areality = Locality.Const.max
+              };
+            externality_upper_bound = Externality.max;
+            nullability_upper_bound = Nullability.Non_null
+          };
+        name = "mutable_data"
```
goldfirere, line 420: Check discussion on Slack about the name here.

ocaml/typing/jkind_intf.ml
```diff
@@ -240,7 +240,9 @@ module History = struct
     | Primitive of Ident.t
     | Immediate_polymorphic_variant
 
-  type immediate64_creation_reason = Separability_check
+  type immediate64_creation_reason =
+    | Separability_check
+    | Primitive of Ident.t
```
goldfirere, line 245: Why do we need this? There aren't any primitive `immediate64`s, I think.

ocaml/typing/predef.ml
```diff
@@ -342,23 +342,35 @@ let build_initial_env add_type add_extension empty_env =
   |> add_type1 ident_iarray
        ~variance:Variance.covariant
        ~separability:Separability.Ind
+       ~param_jkind:(Jkind.Primitive.any_non_null ~why:Array_type_argument)
```
goldfirere, line 345: This looks unrelated to this patch, but instead is a bugfix. Is that right? @dkalinichenko-js, does this look right to you?

ocaml/typing/predef.ml
```diff
@@ -342,23 +342,35 @@ let build_initial_env add_type add_extension empty_env =
...
-  |> add_type ident_char ~jkind:(Jkind.Primitive.immediate ~why:(Primitive ident_char))
+       ~jkind_annotation:Jkind.Const.Primitive.immediate
+  |> add_type ident_char
+      ~jkind:(Jkind.Primitive.immediate ~why:(Primitive ident_char))
       ~jkind_annotation:Jkind.Const.Primitive.immediate
   |> add_type ident_exn
        ~kind:Type_open
        ~jkind:(Jkind.Primitive.value ~why:Extensible_variant)
   |> add_type ident_extension_constructor
   |> add_type ident_float
```
goldfirere, line 359: This could also cross uniqueness. Though it's not hard to imagine allowing in-place updates of unique floating point numbers. Maybe better not to.

ocaml/typing/predef.ml
```diff
@@ -472,6 +489,8 @@ let add_small_number_extension_types add_type env =
   let add_type = mk_add_type add_type in
```
goldfirere, line 489: (This is as close as I can comment) @TheNumbat: Should the SIMD types mode-cross? I would assume that they are well-behaved immutable data.

goldfirere (approved):
Looks good!

ocaml/testsuite/tests/typing-modal-kinds/basics.ml
```diff
@@ -683,3 +679,130 @@ let foo : (string -> string) -> (string -> string) @ unique
 [%%expect{|
 val foo : (string -> string) -> unique_ (string -> string) = <fun>
 |}]
+
+(* CR layouts: maybe use the helper function below to refactor the tests above,
```
goldfirere, line 683: I think these new tests are good, but I don't see a need to refactor the old ones. Diversity in test structure is good.

ocaml/testsuite/tests/typing-modal-kinds/basics.ml
```diff
@@ -683,3 +679,130 @@ let foo : (string -> string) -> (string -> string) @ unique
...
+   as well as [expected_mode.ml]. *)
+
+let weaken_immutable_data : 'a -> 'a @ contended once nonportable =
```
goldfirere, line 686: Perhaps add a test or two that fails (e.g. on `(int -> int) ref`) just to validate the test structure.

goldfirere (commented):
Oops skipped tests on last review

ocaml/typing/jkind_intf.ml
riaqn: Yes, previously I thought `int32` is `immediate64` which caused bugs. I've fixed it but didn't remove  this. I will remove it.

ocaml/typing/predef.ml
riaqn: Yes in-place update can in principle apply to any `immutable_data`.

ocaml/testsuite/tests/typing-modal-kinds/basics.ml
riaqn: Sounds good - should I remove the CR, then?

ocaml/typing/predef.ml
riaqn: Yes that sounds like an interesting question. If this turns out to be more complex than a few lines of code, I suggest we leave this as future work. (This PR is needed to unblock some other projects).

riaqn: Values of SIMD types - are they pointers to a region in memory? Is that memory managed by OCaml?

ocaml/typing/predef.ml
dkalinichenko-js: This allows `iarray`s to accept all layouts instead of just values. The nullability here is correct.

ocaml/testsuite/tests/typing-modal-kinds/basics.ml
goldfirere: Yes, please. Thanks.

ocaml/typing/predef.ml
TheNumbat: Yeah, they should be immutable data. You can treat them the same as floats; the unboxed versions live in a single register and the boxed versions are a pointer to an immutable two-word custom block.

TheNumbat: (Same goes for `float32`.)

riaqn: Thanks - added.

ocaml/typing/jkind.ml
riaqn: I will rename `immutable_data` to `data` in a separate PR. `mutable_data` stays as it is.

PR #2879, opened 2024-08-01
Unboxed tuples

ccasin:
This adds our first composite layout, products, and a type that uses it, unboxed tuples.  There are two commits, the first is some middle-end fixes that will be reviewed in #2837 and can be ignored.

**Known remaining issues**
- The tests include an example of a value `let rec` that crashes closure conversion in the native code compiler. We should do something about this before merging this PR, but I'm still discussing with the middle end team. It's an odd example, because the value rec check today will never reject a let rec that's not actually recursive, and it seems we need to start doing that.
- Errors can be improved! Currently a kind error involving two products does not point to the component of the product that is the problem. Probably histories should become product histories to support that. I think this will be some work and we should plan to merge this PR in its current state (with respect to those errors) but improve them before moving unboxed tuples out of beta.  I will make a ticket in our internal system for this.
- Modes: I've written `(* XXX modes *)` in some places where someone more expert on our mode inference should check what I've done.

**Some notes about the design and implementation**
- I've given sort variables unique ids.  We considered this in the initial layouts PR, but I didn't think we needed them then because sort variables are in types and types have unique ids. But now that we have complex layouts, you can have multiple sort vars in a type, and the unique ids become very useful for debugging.  Many tests change because we now print these ids instead of the old system for sort variables.
- I regrouped the statically allocated values in the sort module into some inner modules. We want statically allocated versions of consts, sorts, and sort options and it was becoming unwieldy/confusing (at least to me) to have them all in the same namespace.  It's possible we should be doing dynamic memoization of products in some of these cases, but I haven't yet.
- Limiting use to beta: We've agreed to limit the use of unboxed products to beta until we've had a chance to do some more real-world testing.  I've done that by doing an extension check at the places where we typecheck the relevant syntax (unboxed tuple expressions, patterns, and type expressions, and unboxed product kind expressions). This of course means you can try to slip in an unboxed product from a library, for which there is a backup check in typeopt, as has been our practice.
- We have discussed the possibility of splitting up the core implementation of `constrain_type_jkind` into multiple functions - one which checks layouts and traverses types deeply to handle products, and another which handles the rest of the kind and can be shallower. I have not done that.  It was not necessary, and while this means the new implementation of `constrain_type_jkind` does some unnecessary checking of the other kind bits when dealing with unboxed products, it allows for (in my view) a pretty clean and understandable implementation of kind checking.  It's possible we should split this up in the future, but I think that will require adjusting how we think about kind errors and histories, as right now they live at the level of the whole kind but the checks wouldn't be operating on whole kinds. Either way, we should probably leave that exploration for a future PR, as this one already makes enough changes to kind checking.

Review: @goldfirere


ocaml/typing/jkind.ml
```diff
@@ -717,6 +837,29 @@ module Const = struct
...
+          nullability_upper_bound
+        } =
+      ( layout :: layouts,
+        Modes.join mode_ub modes_upper_bounds,
+        Externality.join ext_ub externality_upper_bound,
+        Nullability.join null_ub nullability_upper_bound )
+    in
+    let layouts, mode_ub, ext_ub, null_ub =
+      List.fold_left folder
+        ([], Modes.min, Externality.min, Nullability.max)
```
goldfirere, line 854:
```suggestion
        ([], Modes.min, Externality.min, Nullability.min)
```

?

Probably worth adding some tests around this function.... though I suppose the existing mode-crossing tests are the tests. Externality and nullability don't matter for non-values.

ocaml/typing/jkind.ml
```diff
@@ -1005,13 +1165,63 @@ module Jkind_desc = struct
     let bits64 = of_const Const.Primitive.bits64.jkind
   end
 
+  let product jkinds =
+    (* CR layouts 7.1: Here we throw away the history of the component
```
goldfirere, line 1169:
```suggestion
    (* CR layouts v7.1: Here we throw away the history of the component
```

Sorry to nitpick, but it makes grepping easier

ocaml/typing/jkind.ml
```diff
@@ -1005,13 +1165,63 @@ module Jkind_desc = struct
     let bits64 = of_const Const.Primitive.bits64.jkind
   end
 
+  let product jkinds =
```
goldfirere, line 1168: This is very similar to `jkind_of_product_annotations`. Combine?

ocaml/typing/jkind.ml
```diff
@@ -1005,13 +1165,63 @@ module Jkind_desc = struct
...
+          has_warned = _
+        } =
+      ( layout :: layouts,
+        Modes.join mode_ub modes_upper_bounds,
+        Externality.join ext_ub externality_upper_bound,
+        Nullability.join null_ub nullability_upper_bound )
+    in
+    let layouts, mode_ub, ext_ub, null_ub =
+      List.fold_left folder
+        ([], Modes.min, Externality.min, Nullability.max)
```
goldfirere, line 1190:
```suggestion
        ([], Modes.min, Externality.min, Nullability.min)
```

?

ocaml/typing/jkind.ml
```diff
@@ -1005,13 +1165,63 @@ module Jkind_desc = struct
...
+        jkinds
+    in
+    { layout = Product (List.rev layouts);
+      modes_upper_bounds = mode_ub;
+      externality_upper_bound = ext_ub;
+      nullability_upper_bound = null_ub
+    }
+
   (* Post-condition: If the result is [Var v], then [!v] is [None]. *)
```
goldfirere, line 1199:
```suggestion
  (* Post-condition: If the result contains [Var v], then [!v] is [None]. *)
```

ocaml/typing/jkind.ml
```diff
@@ -818,24 +964,38 @@ end
 module Desc = struct
   type t =
```
goldfirere, line 965: This doesn't seem quite right to me: this looks like the `Desc` for product kinds. I think this may want a different structure.

goldfirere (commented):
Nearly done with jkind.ml.

ocaml/typing/jkind.ml
ccasin: I meant to write something about this, sorry.  I agree this is weird, but submit that it was already wrong and this PR makes it no worse.  The comment about modes is plainly false, to start, and vars are layouts but consts are kind, so the model/layout mismatch is already here.  This should be cleaned up, and I vaguely though that was in the scope of the ongoing refractors of this module.

goldfirere: I see what you mean here, but I maintain that the old version was OK. That is, the var case (and its comment) was meant to indicate that the var case can be treated as a kind, but with all of its mode stuff set to `max`. This wasn't captured in the old datatype (perhaps foolishly), but any consumer of it can easily apply that knowledge.

This new world, though, resists this kind of interpretation. So I see the new one as worse than the old one.

ccasin: The reason I say the comment is false is that we interpret all kinds with layout sort var as the Var case here - that is, it is used to represent some vars about which we would not say "all modes will be [max]".  I think, therefore, either the old type was wrong and should have been a record containing modes and a (const layout or var), or the old type was fine but the new one is too (because whether or not we're accurately recording kinds should be remembered at the granularity of individual product components).

goldfirere: Previously, we never really used the `Var` output from `get` (the only way in which `Desc.t` interacts with the outside world). In most cases, `Var` is just a fatal error. The exception is in printing, at which point we don't care about the modes.

That's still true in the new patch, actually.

So maybe here's a middle ground: keep `Desc.t` as it is, but with an invariant that a `Product` contains at least one `Var` or `Product`. (This will require an easy implementation change to guarantee the invariant.) Then consumers have reliable mode stuff for `Const` but not for the other cases -- but those other cases are error cases in everything but the printer, so it's all OK. If that doesn't work out in the future (it looks just fine for this patch), we can think about a deeper restructuring later.

goldfirere (commented):
I was hoping to get further here before my vacation -- and I may return during some downtime if no one beats me to it -- but I wanted to post where I had gotten in case someone else wants to pick up.

I have reviewed:
* All the testsuite changes
* All files alphabetically from typing/jkind.ml to typing/patterns.ml
* typedtree.ml{i}
* types.ml{i}
* utils/misc.ml{i}

lthls:
> * The tests include an example of a value `let rec` that crashes closure conversion in the native code compiler. We should do something about this before merging this PR, but I'm still discussing with the middle end team. It's an odd example, because the value rec check today will never reject a let rec that's not actually recursive, and it seems we need to start doing that.

Its possible that there are places in the compilation of `let rec` where we use `layout_letrec` (which is Value) whereas we should be using the actual layout.
I think we can sort recursive definitions into three cases:
- Actually recursive definitions with non-value layouts should be rejected. The current compilation scheme cannot easily be modified to allow them.
- Definitions that are not actually recursive should be handled properly, which may require patching a few things in `Value_rec_compiler`.
- There is a special case for definitions that are not ultimately recursive but where the definition mentions a recursive variable (`let rec x = let _ = x in ()`). The handling for these is weird (it's the `Constant` case in Value_rec_compiler`), and I would argue that we shouldn't support that for non-value layouts.

The important part, I think, is to ensure that `Value_rec_check` doesn't misclassify as `Static` something that has non-value layout. `Static` implies that it can be pre-allocated using a scheme that only applies for (a subset of) values, so even though some non-value expressions could be supported through the `Constant` hack I think it is better to not support it, at least at the beginning.

ccasin:
> > * The tests include an example of a value `let rec` that crashes closure conversion in the native code compiler. We should do something about this before merging this PR, but I'm still discussing with the middle end team. It's an odd example, because the value rec check today will never reject a let rec that's not actually recursive, and it seems we need to start doing that.
> 
> Its possible that there are places in the compilation of `let rec` where we use `layout_letrec` (which is Value) whereas we should be using the actual layout. I think we can sort recursive definitions into three cases:
> 
> * Actually recursive definitions with non-value layouts should be rejected. The current compilation scheme cannot easily be modified to allow them.
> * Definitions that are not actually recursive should be handled properly, which may require patching a few things in `Value_rec_compiler`.
> * There is a special case for definitions that are not ultimately recursive but where the definition mentions a recursive variable (`let rec x = let _ = x in ()`). The handling for these is weird (it's the `Constant` case in Value_rec_compiler`), and I would argue that we shouldn't support that for non-value layouts.
> 
> The important part, I think, is to ensure that `Value_rec_check` doesn't misclassify as `Static` something that has non-value layout. `Static` implies that it can be pre-allocated using a scheme that only applies for (a subset of) values, so even though some non-value expressions could be supported through the `Constant` hack I think it is better to not support it, at least at the beginning.

Thanks for this summary, @lthls.  In the interest of keeping things simple, I am tempted to make a PR that just bans any pattern variable in a value let-rec from having a non-value layout.  I think that considering the existence of `layout_letrec`, we probably intended to ban these and it's just a missing check.  This would ban a few things that we could actually compile (like let recs that aren't actually recursive), but that seems like a small price to pay for a simple implementation and nice understandable rule.

What do you think?

ocaml/typing/ctype.ml
```diff
@@ -2158,28 +2195,94 @@ let type_jkind_sub env ty jkind =
...
+        estimate, Failure { estimate; bound = jkind; ty }
+      | Some jkinds ->
+        (* Note: here we "duplicate" the fuel, which may seem like
+           cheating. Fuel counts expansions, and its purpose is to guard against
+           infinitely expanding a recursive type. In a wide product, we many
+           need to expand many types shallowly, and that's fine. *)
+        let results = List.map2 (check_one fuel) component_types jkinds in
+        process_product_results fuel [] [] results
+
+  (* Here, [results] is a list of results from attempting to constrain a product
```
goldfirere, line 2252: I got thrown off by this comment, because `results` is only the unprocessed results. The others are in the accumulators. But as stated, I was confused to see something meaningfully happen when `results` is empty (but of course it's just because that happens at the end of the recursion).

ocaml/typing/ctype.ml
```diff
@@ -2115,36 +2116,72 @@ let rec estimate_type_jkind env ty =
...
-  | Failure of Jkind.t
+  | Type_vars of type_jkind_sub_possible_failure list
+  (* [Type_vars] indicates a possible success - the caller should check, and
+     possibly update the variables *)
+  | Missing_cmi of Path.t
+  | Failure of type_jkind_sub_possible_failure
+  (* [Failure] indicates we couldn't verify the constraint.  However, further
+     expansion the type might make it possible to verify. *)
 
 let type_jkind_sub env ty jkind =
```
goldfirere, line 2167: After reading this all, I'm left with the feeling that there is probably a cleaner algorithm underneath this all. I see a few bits that don't quite smell right:
* The refactoring you suggest in the CR in `process_product_results`.
* The fact that the jkind estimate is returned and usually stored in the result.
* The fact that there is a failure-with-the-possibility of improvement, and then there is a hard failure -- yet this algorithm doesn't differentiate them. For example, if `is_nary_product` encounters a `float64`, no amount of expansion will help. It might be nice if a discovery like this just stopped the expansion.

None of these are terribly bad. But my overall sense here is that time spent here will hopefully yield code that is simpler and closer to "obviously no bugs" than "no obvious bugs".

goldfirere (commented):
A little bit more: ctype, btype, and all the files in typing up to and including printtyped.

lthls:
Banning non-value layouts in recursive declarations seems reasonable.
If there is demand to support the non-recursive case, I can implement it (it's a simple matter of passing the layouts to `Value_rec_compiler`).

ocaml/typing/ctype.ml
ccasin: I confess I spent a fair amount of time on this and thought I had ended up with an algorithm that was pretty simple.  These three functions each do, in my mind, a clear and necessary thing - it would help to know what you think a simpler algorithm would look like.  Though I agree some of the comments could be better!

A few thoughts:

> The fact that the jkind estimate is returned and usually stored in the result.

I think these are two different things.  These functions return the jkind estimate for the entire type they are passed.  They also return a result.  That result may indicate we failed to prove several different constraints for several different types, and contains the estimates for those types (which are pieces of the original type, but not the same as the original type).  Of course, at the bottom of the recursion where we have non-composite layouts, the failure information and the estimate must be about the same type.

I did consider whether the estimate is needed at all in `type_jkind_sub_possible_failure`.  In practice, it's only really needed in the `Type_vars` case, but having one type to represent all potentially recoverable failures seems nice.

> The fact that there is a failure-with-the-possibility of improvement, and then there is a hard failure -- yet this algorithm doesn't differentiate them. For example, if is_nary_product encounters a float64, no amount of expansion will help. It might be nice if a discovery like this just stopped the expansion.

This is true!  To be clear, the old algorithm also suffered from this - checking whether something with layout `float64` had layout `value` would always result in fully expanding the type.  Our thought was in error cases like these, performance isn't so important - I am tempted to say that bad perf in the failure case is a good trade for a nice clear algorithm.  But I suppose it is the case that some kind checking failures are not actually errors (like mode crossing checks, and probably some gadt stuff), so I will make a ticket to explore this in the future.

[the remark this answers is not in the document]
ccasin: We won't expand if the kind is `any`.  We only expand in the `Tconstr` case, and that case begins with a cheap check, before expansion, that will succeed in the case of `any`.

ocaml/typing/ctype.ml
```diff
@@ -2158,28 +2195,94 @@ let type_jkind_sub env ty jkind =
      ensure it's a valid argument to [t].  (We believe there are still loops
      like this that can occur, though, and may need a more principled solution
      later).  *)
-  let rec loop ty fuel =
+  let rec expand_and_check_one fuel ty jkind =
```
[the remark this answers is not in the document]
ccasin: Hmm, I this this is a little subtle:
* If the type is not a Tconstr, it doesn't expand at all (there is nothing to expand).
* If the type is a Tconstr, it first does a very cheap check as an optimization, then expands, then does the real check.
I think it's clearest to think about it as expanding then checking - the cheap check is just an optimization, as noted in the comment.  I'll move the comment about that optimization into the Tconstr case to make it a little clearer.

ocaml/lambda/matching.ml
```diff
@@ -106,11 +106,16 @@ exception Error of Location.t * error
 let dbg = false
 
 let jkind_layout_default_to_value_and_check_not_void loc jkind =
+  let rec contains_void : Jkind.Layout.Const.t -> bool = function
+    | Any -> false
+    | Base Void -> true
+    | Base (Value | Float64 | Float32 | Word | Bits32 | Bits64) -> false
+    | Product ts -> List.exists contains_void ts
```
mshinwell, line 113: What happens if `ts` is the empty list?

mshinwell:
I rebased onto `b22fb1a970c763952da11b188a471368747ba50d` which made the flambda2 diffs go away.  I then read the changes to `bytecomp/`, `file_formats/` and `lambda/` in that rebase, all of which are ok.  I left one minor comment.

ocaml/testsuite/tests/typing-layouts-products/unboxed_tuples.ml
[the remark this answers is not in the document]
mshinwell: @lthls is going to look at it

ocaml/testsuite/tests/typing-layouts-products/unboxed_tuples.ml
```diff
@@ -0,0 +1,379 @@
+(* TEST
+ reference = "${test_source_directory}/unboxed_tuples.reference";
+ include stdlib_upstream_compatible;
+ flambda2;
```
mshinwell, line 4: We need to test unboxed tuples at `-Oclassic` and `-O3`.  Unsure how you want to set this up.  It is possible to select these using either attributes e.g. `[@@@ocaml.flambda_oclassic]` or using e.g. `ocamlopt_flags="-Oclassic"` in ocamltest.  These only work on a per-source-file basis though...

mshinwell: The easiest is probably to take the product of the existing ocamltest stanzas with ones setting the level I guess.

ocaml/typing/uniqueness_analysis.ml
```diff
@@ -1100,6 +1100,18 @@ and pattern_match_single pat paths : Ienv.Extension.t * UF.t =
       |> conjuncts_pattern_match
     in
     ext, UF.par uf_read uf_args
+  | Tpat_unboxed_tuple args ->
+    (* XXX modes *)
+    let uf_read = Paths.mark_implicit_borrow_memory_address Read occ paths in
```
riaqn, line 1105: This is subtle. `memory_address` here refers to the box memory of a tuple. Unboxed tuples don't have such box so needs to be handled differently.

This file is not exercised in production currently, so it's fine to be loose. I will push a commit of what I think is correct, and let's call it a day.

riaqn (commented):
I've reviewed the following files:
```
 ocaml/parsing/ast_helper.ml                               |   5 ++++
 ocaml/parsing/ast_helper.mli                              |   7 ++++++
 ocaml/parsing/ast_iterator.ml                             |  24 ++++++++------------
 ocaml/parsing/ast_mapper.ml                               |  27 +++++++++++-----------
 ocaml/parsing/depend.ml                                   |  18 +++++++++------
 ocaml/parsing/jane_syntax.ml                              |   6 +++++
 ocaml/parsing/jane_syntax.mli                             |   1 +
 ocaml/parsing/lexer.mll                                   |   1 +
 ocaml/parsing/parser.mly                                  |  41 ++++++++++++++++++++++++++++++++-
 ocaml/parsing/parsetree.mli                               |  23 +++++++++++++++++++
 ocaml/parsing/pprintast.ml                                |   1 +
 ocaml/parsing/printast.ml                                 |  13 +++++++++++
 ocaml/toplevel/genprintval.ml                             |   4 ++++
 ocaml/typing/subst.ml                                     |   9 ++++++--
 ocaml/typing/tast_iterator.ml                             |   3 +++
 ocaml/typing/tast_mapper.ml                               |   9 ++++++++
 ocaml/typing/typecore.ml                                  | 163 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++--
 ocaml/typing/typedecl.ml                                  |  86 ++++++++++++++++++++++++++++++++++++++++++++++-----------------------
 ocaml/typing/typedecl.mli                                 |   1 +
 ocaml/typing/typedecl_separability.ml                     |   3 +++
 ocaml/typing/typedecl_variance.ml                         |   2 ++
 ocaml/typing/typemod.ml                                   |  10 ++++----
 ocaml/typing/typeopt.ml                                   | 100 ++++++++++++++++++++++++++++++++++++++++++++++++++++----------------------------
 ocaml/typing/typeopt.mli                                  |   4 ++--
 ocaml/typing/typetexp.ml                                  |  17 +++++++++++++-
 ocaml/typing/uniqueness_analysis.ml                       |  14 ++++++++++++
 ocaml/typing/untypeast.ml                                 |  10 ++++++++
 ocaml/typing/value_rec_check.ml                           |   6 +++++
 printer/printast_with_mappings.ml                         |  13 +++++++++++
```
I'm unconfident about `typedecl.ml` and `typeopt.ml`, which we will discuss via other channels.

ocaml/typing/typecore.ml
```diff
@@ -1463,6 +1469,38 @@ let solve_Ppat_tuple ~refine ~alloc_mode loc env args expected_ty =
   unify_pat_types ~refine loc env ty expected_ty;
   ann
 
+(* This assumes the [args] have already been reordered according to the
+   [expected_ty], if needed.  *)
+let solve_Ppat_unboxed_tuple ~refine ~alloc_mode loc env args expected_ty =
+  let arity = List.length args in
+  let arg_modes =
+    (* XXX modes *)
```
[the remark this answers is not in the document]
riaqn: I looked after rebase: this is correct.

ocaml/typing/typecore.ml
```diff
@@ -7674,6 +7784,57 @@ and type_tuple ~loc ~env ~(expected_mode : expected_mode) ~ty_expected
...
+    unify_exp_types loc env to_unify (generic_instance ty_expected));
+
+  let argument_modes =
+    (* XXX modes *)
+    match expected_mode.tuple_modes with
+    (* CR zqian: improve the modes of opened labeled tuple pattern. *)
+    | Some tuple_modes when List.compare_length_with tuple_modes arity = 0 ->
+        tuple_modes
+    | _ ->
+        List.init arity (fun _ -> expected_mode.mode)
```
riaqn, line 7814: Using `expected_mode.mode` while ignoring `tuple_modes` is dangerous. (See the comments at the definition of `type expected_mode`). We should use `as_single_mode expected_mode` here.

Speaking of which, could you copy `tuple_modes.ml` to `unboxed_tuple_modes.ml` and adapt the tests to unboxed tuples? Duplicative type checking logic deserves duplicative tests.

ocaml/typing/typecore.ml
```diff
@@ -7753,6 +7864,54 @@ and type_tuple ~loc ~env ~(expected_mode : expected_mode) ~ty_expected
...
+  in
+  let to_unify = newgenty (Tunboxed_tuple labeled_subtypes) in
+  with_explanation explanation (fun () ->
+    unify_exp_types loc env to_unify (generic_instance ty_expected));
+
+  let argument_modes =
+    (* XXX modes *)
+    if List.compare_length_with expected_mode.tuple_modes arity = 0 then
+      expected_mode.tuple_modes
+    else List.init arity (fun _ -> expected_mode.mode)
```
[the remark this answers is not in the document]
riaqn: Resovling this because outdated.

ocaml/typing/typecore.ml
riaqn: Also as you noted, we should unify the type checking of regular tuple and unboxed tuple, as well as `Texp_tuple` and `Texp_unboxed_tuple`.

ocaml/typing/uniqueness_analysis.ml
riaqn: I pushed what I think is correct (with testing showing its difference to regular tuple). Resolving.

ocaml/typing/ctype.ml
goldfirere: Yes, good points. I don't have a clear idea of what exactly I want here.... just a feeling that there is something yet simpler inside trying to get out. But we don't have to find that simpler thing (if it even exists) now -- I'd be happy to merge this as-is, especially if this is already the result of whittling. And that's true about the failure case being problematic previously.

goldfirere:
Would it be helpful for me to review typedecl and typeopt?

ccasin:
> Would it be helpful for me to review typedecl and typeopt?

I believe @riaqn has reviewed everything remaining, but you are of course welcome to take a look!

I plan to clean this up this week.

ocaml/testsuite/tests/typing-layouts-products/unboxed_tuples.ml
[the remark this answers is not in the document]
lthls: I'd like to see examples testing the interaction with the following features:
- Continuations (the easiest way to test that is to use local functions, e.g. `let[@local] foo x = ... in ...`). For the continuations to stay in Flambda 2, there need to be several uses of the function.
- Function-level unboxing (with the [@unboxable] attribute). No need to support new things, but checking that annotating an unboxed tuple with `[@unboxable]` raises a warning as expected and generates correct code.
- Loopification: something like `let[@loop] rec foo #(x, y) = foo #(y, x)`, or maybe a fibonacci-like function to make it runnable.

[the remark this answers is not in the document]
mshinwell: Function-level unboxing tests can wait, but please leave a comment on the PR with the unboxable tests, which are still being worked on.

ocaml/testsuite/tests/typing-layouts-products/basics.ml
```diff
@@ -0,0 +1,720 @@
...
+Line 3, characters 17-21:
+3 |     let #(x,y) = utup in
+                     ^^^^
+Error: This expression has type ('a : value)
+       but an expression was expected of type #('b * 'c)
+       The layout of #('a * 'b) is '_representable_layout_203 & '_representable_layout_204
+         because it is an unboxed tuple.
+       But the layout of #('a * 'b) must be a sublayout of value
+         because it's the type of a variable captured in an object.
+|}];;
```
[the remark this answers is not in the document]
ccasin: A few of these were already there, but I've added the rest.  (And turned up a couple inference issues in the process - mostly unrelated to these actually features, but still very helpful, thanks).

ocaml/testsuite/tests/typing-layouts-products/basics.ml
```diff
@@ -0,0 +1,720 @@
...
+
+type t_product : value & value
+
+external ext_tuple_arg : #(int * bool) -> int = "foo"
+[%%expect{|
+type t_product : value & value
+Line 3, characters 25-38:
+3 | external ext_tuple_arg : #(int * bool) -> int = "foo"
+                             ^^^^^^^^^^^^^
+Error: Unboxed product layouts are not supported in external declarations
```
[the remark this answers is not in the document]
ccasin: Certainly agree this can be made to work, but I don't think we have a design fleshed out (and don't need one, yet).

ocaml/typing/jkind_intf.ml
```diff
@@ -33,8 +33,14 @@ module type Sort = sig
   (** A sort variable that can be unified during type-checking. *)
   type var
 
+  type const =
+    | Const_base of base
```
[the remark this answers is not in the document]
ccasin: Aha - we can.  We had name conflicts because this type was exposed in the outer module, for reasons Liam explains [here](https://github.com/ocaml-flambda/flambda-backend/pull/2676#discussion_r1646730077), but since Sort.t no longer depends on const after this PR, we can unwind that and delete the `const` type, keeping only the `Const.t` version.  I've done all that.

ocaml/typing/jkind_types.mli
```diff
@@ -54,21 +54,35 @@ module Sort : sig
     | Bits32
     | Bits64
 
+  val to_string_base : base -> string
```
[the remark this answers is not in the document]
ccasin: Since `Sort.t` now depends on `base`, attempts to move it completely into its own module run into the same issue previously encountered with `const`, described in response to the previous question.  And I think half moving it is more confusing than not, so I've just left it.

ocaml/typing/jkind.ml
ccasin: Well spotted - a hold-over from the earlier discussion of whether we should take a join or meet of nullability.  Fixed, and a test added.

ocaml/typing/jkind.ml
ccasin: Sadly they look more similar than they are - they are working on different types with similar names.  I did not find a clean way to combine them.

ocaml/typing/jkind.ml
ccasin: (Indeed - see above comment)

ocaml/typing/jkind.ml
ccasin: I took a [stab](https://github.com/ccasin/flambda-backend/commit/7ea66be9536ddb730f8bf1e170ec0af876e529f7) at this, but it made printing worse - something that used to be printed as `any_non_null & any_non_null` was printed as `any & any` (with `alpha` on, so these are different).  So I reverted it.

I think we'll agree that we don't actually want to print either of those, but rather `(any & any) mod non_null`.  Which sent me down a little printing rabbit hole - I think we'll probably want to do a bigger rethink of how kinds are printed to accommodate products, and have realized I don't understand some of the current choices, so I'm going to make a ticket for now and suggest we revisit in the near future.

ocaml/typing/ctype.ml
```diff
@@ -2115,36 +2116,72 @@ let rec estimate_type_jkind env ty =
...
+  | Tpoly (ty, _) -> estimate_type_jkind_head env ty
   | Tpackage _ -> Jkind (Primitive.value ~why:First_class_module)
 
+(* We parameterize [estimate_type_jkind] and [jkind_of_result] by a function
+   [expand_components] because some callers want expansion of types and others
+   don't. *)
+let rec estimate_type_jkind env ~expand_components ty =
+  jkind_of_result env ~expand_components (estimate_type_jkind_head env ty)
+
+and jkind_of_result env ~expand_components jkind_head =
```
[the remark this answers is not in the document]
ccasin: Is the subtlety you are referring to just [expand_components]?  I have added a comment pointing to the explanation on [estimate_type_jkind], and also given this a slightly better name ("result" referred to a type that no longer exists).  But let me know if that doesn't cover it.

ocaml/lambda/matching.ml
ccasin: The typechecker won't allow that, but I've added a sanity check.

ocaml/testsuite/tests/typing-layouts-products/unboxed_tuples.ml
ccasin: I have done this, but I was surprised to see this on `-Oclassic`:
```
> ========================================
> File "unboxed_tuples.ml", line 74, characters 36-63:
> 74 |   print_ints "Test 1, 14/3 inlined" ((div_mod [@inlined]) 14 3);
>                                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
> Warning 55 [inlining-impossible]: Cannot inline: 
>   the optimizer did not know what function was being applied (is it marked [@inline never]?)
>   (the full inlining stack was: unboxed_tuples.ml:74,36--63)
> ========================================
```
This function is defined just above in the same file, and is not marked `@inline never`.  I've changed the `@inlined` to `@inlined hint` for now, but want to check if this is expected.

lthls: Here is the relevant part of the `-drawflambda` output:
```
camlUnboxed_tuples__div_mod_5_5 =
  ((code_metadata
...
     (inlining_decision (Function_body_too_large 25))
```
The message in the warning is not very helpful, and I think it could be refined to mention the actual reason. I'll make a note to try to improve this.

ccasin: That all makes sense - thanks very much for the explanation!

ocaml/parsing/pprintast.ml
```diff
@@ -398,6 +398,7 @@ and jkind ctxt f k = match (k : Jane_syntax.Jkind.t) with
   | With (t, ty) ->
     pp f "%a with %a" (jkind ctxt) t (core_type ctxt) ty
   | Kind_of ty -> pp f "kind_of_ %a" (core_type ctxt) ty
+  | Product ts -> pp f "(%a)" (list (jkind ctxt) ~sep:"@;&@;") ts
```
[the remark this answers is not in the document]
ccasin: Well spotted.  In fact the handling of unboxed tuples in this file was completely wrong, and I'd forgotten about that test.  I've fixed the printing and added a test case covering unboxed tuple types/patterns/expressions and unboxed product layouts.

ocaml/toplevel/genprintval.ml
```diff
@@ -261,6 +261,7 @@ module Make(O : OBJ)(EVP : EVALPATH with type valu = O.t) = struct
       | Void -> Print_as "<void>"
       | Any | Any_non_null -> Print_as "<any>"
       | Float64 | Float32 | Bits32 | Bits64 | Word -> Print_as "<abstr>"
+      | Product _ -> Print_as "<unboxed product>"
```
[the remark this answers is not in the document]
ccasin: I agree, but I think I'll just make a ticket for now.  Note that all other unboxed types are also not printed well here, so this isn't new.  And I think it's hard to actually reach this case, since you can't make a top-level binding to an unboxed product or put one in a structure.

ocaml/typing/typetexp.ml
```diff
@@ -715,6 +715,20 @@ and transl_type_aux env ~row_context ~aliased ~policy mode styp =
         (List.map (fun t -> (None, t)) stl)
     in
     ctyp desc typ
+  | Ptyp_unboxed_tuple stl ->
+    Jane_syntax_parsing.assert_extension_enabled ~loc Layouts
+      Language_extension.Beta;
+    (* XXX modes *)
+    let tl =
+      List.map
+        (fun (label, t) -> label, transl_type env ~policy ~row_context mode t)
```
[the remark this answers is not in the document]
ccasin: OK, I have changed it to match the boxed tuple version - please check this is what you wanted.

ocaml/typing/typecore.ml
ccasin: Zesen and I discussed this out of band.  I adopted his suggestion to duplicate the tests (though I put them all in the same file to make it easier to compare the boxed and unboxed versions).  This turned up some mode bugs that he fixed, and also some missing cases in `parmatch.ml` which are fixed in [this commit](https://github.com/ocaml-flambda/flambda-backend/commit/0d85b523217c85b66ef8f2df40b9265fe4ee4dd1) (which probably should get a quick glance from someone besides me).

ccasin: Regarding unifying the typing of `Texp_tuple` and `Texp_unboxed_tuple` - this could be done, but I think it would make the code less clear.  They use different constructors everywhere and these constructors have different types (`Texp_tuple` takes a list of pairs while `Texp_unboxed_tuple` takes a list of triples) and need various special cases (e.g., the regular tuple case restricts everything to value, while the unboxed tuple case doesn't).  I tried but found it more confusing then helpful. In the future there will be fewer special cases and we'll probably need a list of triples in both cases, so I propose to wait until then.

ocaml/testsuite/tests/typing-layouts-products/unboxed_tuples.ml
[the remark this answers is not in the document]
ccasin: I've added the continuation and loopification tests in `unboxed_tuples.ml` (probably one of you should take a look and see that you are happy).  I could not find the "PR with the unboxable tests" to leave a comment on.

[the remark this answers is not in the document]
lthls: The PR with unboxable tests is #2321.
I will look at the new tests as soon as I find a bit of time.

ccasin:
This PR is ready for final review and merging.

I have squashed all the changes in response to review down into one commit, mainly to make rebasing easier, and this has been rebased on current main.  Tests suggested in review turned up a few issues, all of which have been fixed except one - see Test 11 in `typing-layouts-products/basics.ml`.  That test is broken due to an issue in jkind checking that is orthogonal to unboxed products, and after discussion will fix it separately.

One open question is `chamelon`.  If CI passes I think we are OK to merge, but probably some changes are needed to support minimizing programs involving unboxed products, and I will discuss that separately with @Ekdohibs.

@riaqn Please take the following steps:
* Look through all the unresolved conversations, and resolve them if you agree the current state of affairs is reasonable to merge.  Some of them will lead to longer conversations, but I think there are no big outstanding questions that should block merging. 
* Glance at the changes to the typechecking files in the commit that integrates the review feedback.  You will see a few fixed bugs (e.g., I added some missing cases in `parmatch.ml`, and made a change to `type_jkind_sub` in `ctype` where previously checking whether a type has layout `any` (which should always succeed) could fail if the type had a product layout).
* Approve and merge.

ccasin:
I have resolved all the outstanding conversations here so we can merge, but @goldfirere you will probably find it useful to go through and read the responses at some point.

mshinwell (approved):

PR #2880, opened 2024-08-01
Compile artifacts for lrgrep; upgrade menhir version to 20231231

harrison6462:
This PR generates the necessary build artifacts to use lrgrep for better syntax error messages. This amounted to adding the --cmly and --inspection flags to the menhir stanza in ocaml/dune when building the parser. Lrgrep also requires menhir >= 20211128, so this updates the pin on menhir to enable that. It also makes sure to install the parser.cmly to the install directory so it's accessible for other projects.

When adding the --inspection flag to compile the parser, menhir will generate the inspection API. Some functions / types in this API will refer to types that are defined inside parser.mly (ex., Constant.t or let_binding), but menhir doesn't include these types in the autogenerated parser.mli file, and it causes the build to fail. To get around this, type declarations are moved into the helper parser_types.ml and parser_types.mli files (this is what lrgrep does).

Also, I'm not sure why flambda_parser.ml updated, or if that update was necessary. I included it in the PR just in case these changes from updating the menhir version are meaningful.


ocaml/dune
```diff
@@ -60,13 +72,19 @@
...
+  (per_module
+   ((action
+     (progn
+      (system
+      ; Copy over the interface/implementation of parser_types to prevent the preprocessed
+      ; autogenerated Menhir files from having unbound types/modules exposed from the
+      ; Menhir inspection interface.
+       "[ $(basename %{input-file} | sed 's/.[^.]*$//').mli = $(basename %{input-file}) ] && cat ocaml/parser_types.mli || cat ocaml/parser_types.ml")
+      (system "echo 'module MenhirLib = CamlinternalMenhirLib'")
+      (run cat %{input-file})))
```
lpw25, line 86: It's not obvious why you need to do any copying. Why doesn't the parser just refer to `Parser_types` as an ordinary module? The copied text includes references to other such modules e.g. `Asttypes`.

ocaml/parsing/parser_types.mli
```diff
@@ -0,0 +1,40 @@
+(* Tracks types that the parser uses that get exposed when compiling via Menhir. This
+   prevents the compiled parser.mli from having undeclared types, since it doesn't
+   copy the implementation section from parser.mly, causing types declared in
+   parser.mly and then exposed by Menhir in function signatures in parser.mli
+   will be unbound. *)
```
lpw25, line 5: This comment is hard to make sense of. I think it should just say that these are types used by the parser, and that they need to be placed in a separate module because when you use `--inspection` the parser's interface refers to the result type of every parsing rule.

lpw25 (commented):
Looks good to me, although it's not clear why you still need to do any copying for `parser_types.ml`. This PR needs performance testing before we merge it, just in case it has made the parser noticeably slower.

ocaml/dune
harrison6462: I just tried this out, and actually Menhir does properly generate the .mli file without this line altogether, so I can actually remove it altogether. (Moving the types to parser_types.ml caused the inspection API interface to refer to the previously-problematic types as Parser_types.t)

ocaml/dune
```diff
@@ -60,6 +72,7 @@
    ; remove -w -67 by adding the camlinternalMenhirLib hack like the Makefile
  ))
  (ocamlopt_flags (:include %{project_root}/ocamlopt_flags.sexp))
+ (preprocessor_deps parser_types.ml parser_types.mli)
```
lpw25, line 75: Is this line still needed?

lpw25 (approved):
Looks like there might be a superfluous line left over. Other than that it is good to go. As discussed offline, happy to merge first and then double check there isn't a performance regression.

ocaml/dune
harrison6462: Oops, I didn't consider that removing the echo obsolesced that.

harrison6462: Oops, I didn't consider that removing the echo obsolesced that.

PR #2881, opened 2024-08-01
Fix conflicts in `typing/env.ml` and `typing/persistent_env.ml`

ncik-roberts:
Main changes:
  * See note below about `Unit_info.Artifact.t`, etc.
  * We style more error messages as inline code to match upstream
  * Sometimes, the two implementations of attribute-marking differ a little. These seem easy enough to resolve, but I don't have a ton of context so may have made the wrong decision in some cases.

### `Unit_info.Artifact.t` and `Compilation_unit.Name.t`

See https://github.com/ocaml-flambda/flambda-backend/wiki/Strategy-for-5.2.0-Unit_info-and-compunit.

There are many case of this form:
  * `flambda-backend` takes a `modname : Compilation_unit.Name.t` and `filename : string` argument
  * upstream takes a `cmi : Unit_info.Artifact.t` argument

We resolve these by taking both `modname : Compilation_unit.Name.t` and `cmi : Unit_info.Artifact.t`, and projecting out filename from `Unit_info.Artifact.t`. We should add our internal notion of `modnam` to `Unit_info.Artifact.t` so we can get rid of this extra argument, but we haven't done this yet.

### DIFF
--- a/ocaml/typing/env.ml
+++ b/ocaml/typing/env.ml
@@ -2196,7 +2196,6 @@ and store_extension ~check ~rebind id addr ext shape env =
       cda_shape = shape }
   in
   Builtin_attributes.mark_alerts_used ext.ext_attributes;
-  Builtin_attributes.mark_alerts_used cstr.cstr_attributes;
   Builtin_attributes.mark_warn_on_literal_pattern_used cstr.cstr_attributes;
   Builtin_attributes.warning_scope ext.ext_attributes (fun () ->
   if check && not loc.Location.loc_ghost &&
@@ -4114,7 +4113,8 @@ let report_lookup_error _loc env ppf = function
            "but modules are not module types"
     end
   | Unbound_cltype lid ->
-      fprintf ppf "Unbound class type %a" !print_longident lid;
+      fprintf ppf "Unbound class type %a"
+        (Style.as_inline_code !print_longident) lid;
       spellcheck ppf extract_cltypes env lid
   | Unbound_instance_variable s ->
       fprintf ppf "Unbound instance variable %a" Style.inline_code s;
@@ -4148,7 +4148,8 @@ let report_lookup_error _loc env ppf = function
         (Style.as_inline_code !print_longident) lid
   | Functor_used_as_structure lid ->
       fprintf ppf "@[The module %a is a functor, \
-                   it cannot have any components@]" !print_longident lid
+                   it cannot have any components@]"
+        (Style.as_inline_code !print_longident) lid
   | Abstract_used_as_structure lid ->
       fprintf ppf "@[The module %a is abstract, \
                    it cannot have any components@]"
--- a/ocaml/typing/env.ml
+++ b/ocaml/typing/env.ml
@@ -772,7 +772,6 @@ let error err = raise (Error err)
 let lookup_error loc env err =
   error (Lookup_error(loc, env, err))
 
-<<<<<<< HEAD
 type actual_mode = {
   mode : Mode.Value.l;
   context : shared_context option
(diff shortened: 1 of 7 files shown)


ocaml/typing/env.ml
```diff
@@ -4294,17 +4114,8 @@ let report_lookup_error _loc env ppf = function
...
   | Unbound_cltype lid ->
-<<<<<<< HEAD
       fprintf ppf "Unbound class type %a" !print_longident lid;
       spellcheck ppf extract_cltypes env lid
-||||||| 121bedcfd2
-      fprintf ppf "Unbound class type %a" !print_longident lid;
-      spellcheck ppf extract_cltypes env lid;
-=======
-      fprintf ppf "Unbound class type %a"
-        (Style.as_inline_code !print_longident) lid;
```
ccasin, line 4305: Is there a reason not to take the `Style.as_inline_code` here?

ocaml/typing/env.ml
```diff
@@ -4353,8 +4164,8 @@ let report_lookup_error _loc env ppf = function
       in
```
ccasin, line 4164: I suspect we want to add a `Style.as_inline_code` in the `Functor_used_as_structure` case above (I can't add a comment there)

ocaml/typing/persistent_env.ml
```diff
@@ -46,17 +46,9 @@ let error err = raise (Error err)
...
-      cmi : Cmi_format.cmi_infos }
-=======
-      cmi : Cmi_format.cmi_infos;
-      visibility : Load_path.visibility }
->>>>>>> 5.2.0
 
-<<<<<<< HEAD
   let load = ref (fun ~allow_hidden ~unit_name ->
     let unit_name = CU.Name.to_string unit_name in
     match Load_path.find_uncap_with_visibility (unit_name ^ ".cmi") with
```
ccasin, line 54: I think it was missed in the merge of `Load_path` that we should take the new name for this function (`find_normalized_with_visibility`).

ocaml/typing/persistent_env.ml
```diff
@@ -864,79 +608,48 @@ let report_error ppf =
...
         filename
-        CU.print unit1
-        CU.print unit2
+        (Style.as_inline_code CU.print) unit1
+        (Style.as_inline_code CU.print) unit2
   | Illegal_import_of_parameter(modname, filename) ->
       fprintf ppf
         "@[<hov>The file %a@ contains the interface of a parameter.@ \
          %a is not declared as a parameter for the current unit (-parameter %a).@]"
         Location.print_filename filename
```
ccasin, line 637: Other uses of `Location.print_filename` are now styled as inline code - should we do this one too for consistency?  (And there's another in the next case)

ccasin (approved):
Looks good - some minor comments.

ocaml/typing/env.ml
ncik-roberts: Nope, that was a mistake. Updating.

ocaml/typing/persistent_env.ml
ncik-roberts: Yes, thanks

ocaml/typing/persistent_env.ml
ncik-roberts: I've done that renaming within `Load_path` (of `uncap`-like names to `normalize`-like names). This should be purely a no-op: `normalize`-ing something is `String.lowercase_ascii`.

ncik-roberts: I'll let you resolve @ccasin if the last commit looks fine.

ccasin: Last commit looks good, thanks for doing the rename!

PR #2886, opened 2024-08-02
Modify cms format for ocaml-index

liam923:
To support project wide occurrences, in the short term, cms files need to contain the loadpath and initial environment, along with the text of an identifier. This is information that can likely be removed in the future, but this will get us to a working initial version faster.

Also, this PR updates the `cms_uid_to_loc` field to carry a location that is more desirable. Merlin uses this field to tie a uid to a definition location: https://github.com/janestreet/merlin-jst/blob/main/src/analysis/locate.ml#L224. When using `cmt`s, Merlin returns the location of the name in the definition, via a utility function `loc_of_decl`. When using cms files, it uses the location stored in `cms_uid_to_loc`. But currently, the location stored in `cms_uid_to_loc` is the location of the entire definition. This PR copies Merlin's `loc_of_decl` (with some slight modifications) and uses the location returned by it for `cms_uid_to_loc`. This will make Merlin's behavior consistent between `cms` and `cmt` files.

### DIFF
--- a/ocaml/file_formats/cms_format.ml
+++ b/ocaml/file_formats/cms_format.ml
@@ -29,7 +29,7 @@ type cms_infos = {
   cms_source_digest : Digest.t option;
   cms_initial_env : Env.t option;
   cms_uid_to_loc : (Shape.Uid.t * string Location.loc) Array.t;
-  cms_uid_to_attributes : (Shape.Uid.t * Parsetree.attributes) Array.t;
+  cms_uid_to_attributes : Parsetree.attributes Shape.Uid.Tbl.t;
   cms_impl_shape : Shape.t option; (* None for mli *)
   cms_ident_occurrences :
     (Longident.t Location.loc * Shape_reduce.result) array
@@ -65,21 +65,11 @@ let register_toplevel_attributes uid ~attributes ~loc =
   toplevel_attributes := (uid, loc, attributes) :: !toplevel_attributes
 
 let uid_tables_of_binary_annots binary_annots =
-  (* We use an array rather than a hashtable, map, or list for better memory efficiency.*)
-  (* Compute the size of the arrays *)
-  let declaration_count = ref 0 in
-  Cmt_format.iter_declarations binary_annots
-    ~f:(fun _ _ -> declaration_count := !declaration_count + 1);
-  let items_count = List.length !toplevel_attributes + !declaration_count in
-  (* Allocate the arrays, filling them with a nonsense value. *)
-  let cms_uid_to_loc = Array.make items_count (Shape.Uid.internal_not_actually_unique, Location.{txt=""; loc=none}) in
-  let cms_uid_to_attributes = Array.make items_count (Shape.Uid.internal_not_actually_unique, []) in
-  (* Fill in the arrays. *)
-  let i = ref 0 in
+  let cms_uid_to_loc = Types.Uid.Tbl.create 42 in
+  let cms_uid_to_attributes = Types.Uid.Tbl.create 42 in
   List.iter (fun (uid, loc, attrs) ->
-    Array.set cms_uid_to_loc !i (uid, loc);
-    Array.set cms_uid_to_attributes !i (uid, attrs);
-    i := !i + 1;)
+    Types.Uid.Tbl.add cms_uid_to_loc uid loc;
+    Types.Uid.Tbl.add cms_uid_to_attributes uid attrs)
     !toplevel_attributes;
   Cmt_format.iter_declarations binary_annots
     ~f:(fun uid decl ->
@@ -99,9 +89,8 @@ let uid_tables_of_binary_annots binary_annots =
         | Class v -> v.ci_attributes
         | Class_type v -> v.ci_attributes
       in
-      Array.set cms_uid_to_loc !i (uid, loc);
-      Array.set cms_uid_to_attributes !i (uid, attrs);
-      i := !i + 1
+      Types.Uid.Tbl.add cms_uid_to_loc uid loc;
+      Types.Uid.Tbl.add cms_uid_to_attributes uid attrs
     );
   cms_uid_to_loc, cms_uid_to_attributes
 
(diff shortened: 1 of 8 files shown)


ocaml/typing/typemod.ml
```diff
@@ -3522,7 +3522,7 @@ let cms_register_toplevel_attributes ~sourcefile ~uid ~f ast =
   match attr with
   | None -> ()
   | Some attr ->
-    Cms_format.register_toplevel_attributes uid
+    Cms_format.register_toplevel_attributes (uid : Uid.t)
```
goldfirere, line 3525: This change looks like a debugging fragment. If it is, please remove.

ocaml/file_formats/cms_format.ml
```diff
@@ -26,8 +26,10 @@ type cms_infos = {
   cms_comments : (string * Location.t) list;
   cms_sourcefile : string option;
   cms_builddir : string;
+  cms_loadpath : Load_path.paths;
```
goldfirere, line 29: As discussed, I'm not sure we need to store this. The build system can likely provide it. Please add a CR saying we might be able to get rid of this later.

ocaml/file_formats/cms_format.ml
```diff
@@ -26,8 +26,10 @@ type cms_infos = {
...
   cms_source_digest : Digest.t option;
-  cms_uid_to_loc : Location.t Shape.Uid.Tbl.t;
+  cms_initial_env : Env.t;
```
goldfirere, line 31: This field also may be storing more than necessary. If I understand correctly, it's always produced by `Typemod.initial_env`, which consults a few flags and then does its work. Instead, we could persist the flags (or get them from the build system).

ocaml/file_formats/cms_format.ml
```diff
@@ -26,8 +26,10 @@ type cms_infos = {
...
+  cms_uid_to_loc : string Location.loc Shape.Uid.Tbl.t;
```
goldfirere, line 32: As discussed, I think it would save space to store a `(Uid.t * string Location.loc) array` (or similar), instead of serializing the table directly. Perhaps that improvement goes in a separate PR, though.

ocaml/file_formats/cmt_format.mli
```diff
@@ -120,6 +120,8 @@ val iter_declarations
     -> f:(Shape.Uid.t -> Typedtree.item_declaration -> unit)
     -> unit
 
+val need_to_clear_env : bool
```
goldfirere, line 123: Comment?

goldfirere (approved):

ocaml/file_formats/cms_format.ml
liam923: This was an easy enough change that I just went ahead and did it.

ocaml/file_formats/cms_format.ml
liam923: Agreed. I added a CR

ocaml/file_formats/cms_format.ml
liam923: Agreed, CR added.

ocaml/file_formats/cmt_format.mli
liam923: Added

ocaml/typing/typemod.ml
liam923: Yes, good catch, thanks

liam923: Removed

ocaml/file_formats/cms_format.ml
```diff
@@ -120,7 +120,7 @@ let save_cms filename modname binary_annots sourcefile initial_env shape =
             cms_source_digest = source_digest;
             cms_initial_env = if Cmt_format.need_to_clear_env
               then Env.keep_only_summary initial_env else initial_env;
-            cms_uid_to_loc;
+            cms_uid_to_loc = cms_uid_to_loc |> Shape.Uid.Tbl.to_list |> Array.of_list;
```
goldfirere, line 123: I'm a tiny bit worried about the performance cost of this indirect conversion. Can we go straight to an array?

goldfirere (approved):

ocaml/file_formats/cms_format.ml
```diff
@@ -71,37 +73,41 @@ let uid_tables_of_binary_annots binary_annots =
...
-            Cmt_format.index_occurrences binary_annots
+            let cms_ident_occurrences = Cmt_format.index_occurrences binary_annots in
+            let cms_initial_env = if Cmt_format.need_to_clear_env
+              then Env.keep_only_summary initial_env else initial_env in
+            cms_ident_occurrences, Some cms_initial_env
           else
-            [| |]
+            [| |], None
         in
         let cms_uid_to_loc, cms_uid_to_attributes =
```
goldfirere, line 112: This `cms_uid_to_loc` is generated here, only to be converted to an array just below. Is it possible to create the array directly, without going via a table first?

ocaml/utils/identifiable.ml
```diff
@@ -226,6 +227,28 @@ module Make_tbl (T : Thing) = struct
 
   let map t f =
     of_map (T_map.map f (to_map t))
+
+  let to_array t =
```
goldfirere, line 231: This is both better and worse than going via a list:
* Worse: it allocates *more* memory, requiring allocation for both the `Some` and the tuple at every iteration.
* Better: The allocated memory is short-lived, making the allocations almost surely in the minor heap and thus quite cheap.

Instead, it seems possible to do this more directly, using `iter` and an `array option ref`. Yes, that's imperative and unhappy. But we're filling an array here, which basically forces us into this stance.

liam923: Sigh, too bad we don't have perceus.

ocaml/file_formats/cms_format.ml
liam923: Do you know whether two `Typedtree.item_declaration`s can define the same `Types.Uid.t`? My understanding is no, in which case this is very doable.

If the answer is yes, then we would need to be slightly careful here about duplicates. We'd either need to dedup here (which would be difficult without building a table) or rely on Merlin to handle duplicates properly.

liam923: It appears that there can indeed be duplicates. For now, I'm going to continue encoding it as a hashtbl

PR #2887, opened 2024-08-02
Fix conflicts in `typing/typemod.ml`

ncik-roberts:
There are several sources of diff:
  * Upstream added `Unit_info.t`; we added `Compilation_unit.Name.t`. For now, we're taking both arguments (see [here](https://github.com/ocaml-flambda/flambda-backend/wiki/Strategy-for-5.2.0-Unit_info-and-compunit)), and we'll add `Compilation_unit.Name.t` to `Unit_info` in the future.
  * Some PRs were realized slightly differently internally vs. externally.
    * Fixing the package hack. ([internal](https://github.com/ocaml-flambda/flambda-backend/pull/2253), [upstream](https://github.com/ocaml/ocaml/pull/12930)). This leads to a slightly nasty diff in `merge_constraint`, which is factored differently internally, but I think it's fine to just take our version.
    * Warning 53 for misplaced attributes ([upstream](https://github.com/ocaml/ocaml/pull/12451)). There appears to be some check about "toplevel" attributes introduced in that PR and later backed out in https://github.com/ocaml/ocaml/pull/13170. @ccasin Should I attempt to backport 13170 as part of this conflict fixing? For now, I've just taken [#12451](https://github.com/ocaml/ocaml/pull/12451), with the thought that we'll pick up [#13170](https://github.com/ocaml/ocaml/pull/13170) in some future merge.
  * Adding more inline-code-formatting to error messages, to match upstream's recent effort to do this.

`Unit_info.t` is the main source of hard-to-understand diff. Let's chat if something looks confusing.

### DIFF
--- a/ocaml/file_formats/cms_format.ml
+++ b/ocaml/file_formats/cms_format.ml
@@ -91,11 +91,13 @@ let uid_tables_of_binary_annots binary_annots =
     );
   cms_uid_to_loc, cms_uid_to_attributes
 
-let save_cms filename modname binary_annots sourcefile shape =
+let save_cms target modname binary_annots shape =
   if (!Clflags.binary_annotations_cms && not !Clflags.print_types) then begin
     Misc.output_to_file_via_temporary
-       ~mode:[Open_binary] filename
+       ~mode:[Open_binary] (Unit_info.Artifact.filename target)
        (fun _temp_file_name oc ->
+
+        let sourcefile = Unit_info.Artifact.source_file target in
         let source_digest = Option.map Digest.file sourcefile in
         let cms_ident_occurrences =
           if !Clflags.store_occurrences then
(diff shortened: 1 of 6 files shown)


ccasin:
> Should I attempt to backport 13170 as part of this conflict fixing? For now, I've just taken https://github.com/ocaml/ocaml/pull/12451, with the thought that we'll pick up https://github.com/ocaml/ocaml/pull/13170 in some future merge.

I suspect we will indeed need 13170 to avoid issuing spurious warnings.

ccasin (approved):
Looks good - didn't spot any problems.  A couple random notes:

> Fixing the package hack. (https://github.com/ocaml-flambda/flambda-backend/pull/2253, https://github.com/ocaml/ocaml/pull/12930). This leads to a slightly nasty diff in merge_constraint, which is factored differently internally, but I think it's fine to just take our version.

I think the package hack fix got merged ~identically upstream and internally, except that the internal version has extra layout checks.  The bad diff in `merge_constraint` seems to mostly be caused by [this refactoring PR](https://github.com/ocaml-flambda/ocaml-jst/pull/107), which maybe we should upstream?

> Should I attempt to backport 13170 as part of this conflict fixing?

As noted above, we'll need this.  I am happy to help backporting it (it's small) but I think we should do it as a separate PR, and I should get through reviewing the existing ones first.

PR #2889, opened 2024-08-02
Abstracting over datatypes with higher kinds

jbachurski:
Implements abstracting over datatypes, i.e. this passes the inclusion check:

```ocaml
module M : sig
  type t : value => value
end = struct
  type 'a t = { foo : 'a }
end
[%%expect {|
module M : sig type t : value => value end
|}]
```

Introduces a `[@@datatype]` attribute for now to denote that a type definition is for a datatype. This is only accepted for abstract types without manifests, as otherwise it is difficult to prove. E.g. we reject this:
```ocaml
type 'a t = 'a * 'a
type 'a s = 'a t [@@datatype]
[%%expect {|
type 'a t = 'a * 'a
Line 2, characters 0-29:
2 | type 'a s = 'a t [@@datatype]
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error: This is not a datatype, but one was expected.
|}]
```
But we still want this to pass:
```ocaml
module M : sig
  type ('a : any) t [@@datatype]  (* datatype_ ('a : any) t *)
end = struct
  type t : any => value = array
end
```
Note that neither `type t = array` (due to some bad jkind defaulting) nor `type 'a t = 'a array` (as it would require implementing a trivial 'proof' of `t` being a datatype) actually work here.

### DIFF
--- a/ocaml/testsuite/tests/typing-higher-jkinds/abstracted.ml
+++ b/ocaml/testsuite/tests/typing-higher-jkinds/abstracted.ml
@@ -1,0 +1,75 @@
+(* TEST
+ flags = "-extension layouts_alpha";
+ expect;
+*)
+
+module M : sig
+  type t : value => value
+end = struct
+  type 'a t = { foo : 'a }
+end
+[%%expect {|
+module M : sig type t : value => value end
+|}]
+
+module M : sig
+  type t : value => value
+end = struct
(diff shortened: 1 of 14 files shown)


jbachurski:
Subsumed by #3035 with a different implementation through `type_noun`s (#2937)

PR #2890, opened 2024-08-02
Bidirectional-style higher kind inference with arity-based defaulting

jbachurski:
Extends kind inference in value declarations, where unannotated type variables can have a higher kind inferred based on:
- Application to some arity of arguments, in which case all arguments are assumed to be zeroth-order and representable. For instance, `('a, 'b) 't` infers `'t : (_, _) => ...`, where `_` are sort vars.
- Application as an argument. If `'a 'b` and `'b` has a checked kind `k => k'`, then `'a : k`.
- Application with an expected result. In the above situation `'b` is defaulted to `... => k'`.

The tricky behaviour here is defaulting to the 'representable space' when a known kind is not a subkind of some kind in the representable space, when the jkind initialisation policy (for unannotated type variables) is to default to sort vars. This seemed to be most consistent. 
The rule is: if the policy is `Sort`, then all non-arrow covariant positions in the kind are intersected with a fresh sort var. In particular, as before, an expected `Any` kind is constrained down to a sort var.

Implements and edits some extra tests as less annotations have to be written now.

### DIFF
--- a/ocaml/typing/typetexp.ml
+++ b/ocaml/typing/typetexp.ml
@@ -92,6 +92,7 @@ type error =
   | Non_sort of
       {vloc : sort_loc; typ : type_expr; err : Jkind.Violation.t}
   | Bad_jkind_annot of type_expr * Jkind.Violation.t
+  | Bad_jkind_inference of type_expr * Jkind.Violation.t
   | Jkind_mismatch_in_application of (type_expr * Jkind.t) * (type_expr * Jkind.t) list * Jkind.Violation.t option
   | Bad_jkind_for_application of type_expr * Jkind.t
   | Did_you_mean_unboxed of Longident.t
@@ -164,14 +165,15 @@ module TyVarEnv : sig
        are in scope. *)
 
   val lookup_local :
-    row_context:type_expr option ref list -> string -> type_expr
-    (* look up a local type variable; throws Not_found if it isn't in scope *)
+    row_context:type_expr option ref list -> string -> type_expr * bool
+    (* look up a local type variable and whether it was bound with a jkind annotation; 
+       throws Not_found if it isn't in scope *)
 
   val lookup_global :
     string -> type_expr
     (* look up a global type variable; throws Not_found if it isn't in scope *)
 
-  val remember_used : string -> type_expr -> Location.t -> unit
+  val remember_used : string -> type_expr -> Location.t -> bool -> unit
     (* remember that a given name is bound to a given type *)
 
   val globalize_used_variables : policy -> Env.t -> unit -> unit
@@ -199,8 +201,9 @@ end = struct
   (* These are variables that have been used in the currently-being-checked
      type, possibly including the variables in [type_variables].
   *)
+  type used_var = { ty: type_expr; loc: Location.t; bound_with_annotation: bool }
   let used_variables =
-    ref (TyVarMap.empty : (type_expr * Location.t) TyVarMap.t)
+    ref (TyVarMap.empty : used_var TyVarMap.t)
 
   (* These are variables that will become univars when we're done with the
      current type. Used to force free variables in method types to become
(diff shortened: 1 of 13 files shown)


ocaml/typing/typetexp.ml
```diff
@@ -786,41 +820,56 @@ and transl_type_aux env ~row_context ~aliased ~policy mode styp =
...
-      Error (
+    let ty =
+      transl_type env ~policy
+        ~jkind_check:(Arity (List.length stl, jkind_check))
+        ~row_context Alloc.Const.legacy st in
+    let ty_jkind = estimate_type_jkind env ty.ctyp_type in
+    let arg_ty_jkinds =
+      match Jkind.get ty_jkind with
+      | Arrow { args; result = _ } -> args
+      | Type _ -> raise (Error (
```
lpw25, line 851: Weird styling. Newline after `->` please.

ocaml/typing/typetexp.ml
```diff
@@ -786,41 +820,56 @@ and transl_type_aux env ~row_context ~aliased ~policy mode styp =
...
-    let app_err violation =
-      let arg_jkinds =
-        List.map (fun arg -> estimate_type_jkind env arg.ctyp_type) arg_tys
-      in
-      Error (
+    let ty =
+      transl_type env ~policy
+        ~jkind_check:(Arity (List.length stl, jkind_check))
+        ~row_context Alloc.Const.legacy st in
+    let ty_jkind = estimate_type_jkind env ty.ctyp_type in
```
lpw25, line 847: Using `estimate_type_jkind` here will result in the inference being order dependent. For instance, consider this example:
```ocaml
type ('f : any => value) any_to_value
val foo : 'f any_to_value -> 'a 'f -> unit
```
This will look at `'f any_to_value` before it looks at `'a 'f` which will result in `'a` have kind `any`. (I'm assuming, I think correctly, for these examples that `val` uses the `any` policy). But if you change it to:
```ocaml
type ('f : any => value) any_to_value
val foo : 'a 'f -> 'f any_to_value -> unit
```
then it will look at `'a 'f` first and default `'f` to have a sort variable for its argument, which will then become the kind of `'a`.

Instead of using `estimate_type_jkind` we should instead only use information that comes directly from the type expression being applied. That means we can use `estimate_type_jkind` for `Typ_constr` but for `Typ_var` we should only be using information coming from annotations on the binding of the variable.

ocaml/typing/typetexp.ml
```diff
@@ -1083,7 +1132,7 @@ and transl_type_var env ~policy ~row_context attrs loc name jkind_annot_opt =
     with Not_found ->
       let jkind =
         try TyVarEnv.lookup_global name |> estimate_type_jkind env
-        with Not_found -> TyVarEnv.new_jkind ~is_named:true policy
+        with Not_found -> TyVarEnv.new_jkind ~is_named:true policy jkind_check
```
lpw25, line 1135: The logic of `new_jkind` is the right logic for how we should be typing variables. However, rather than only using that logic in the `Not_found` case we need to use it whenever the variable being used doesn't have a kind annotation on its binding site. Otherwise the kind inference will be order-dependent. Consider this example:
```ocaml
type ('f : (value => value) => value) third_order
val foo : 'f third_order -> 'a 'f -> int
```
This signature will type-check because the first occurrence of `'f` is in a checking position. But if we swap the argument order:
```ocaml
type ('f : (value => value) => value) third_order
val foo : 'a 'f -> 'f third_order -> int
```
then it will stop type-checking because the first occurrence of `'f` is in inference position and it will assume `'f` has a second-order kind.

Rather than have the relevant logic in `new_jkind`, we should pull it out into its own function for computing the expected kind of a variable. If the variable is `Not_found` then we should use that kind as the initial kind for the variable. If the variable has an annotation on its binding then we don't need to consider the expected kind. If the variable has no annotation, but has already been seen before then we need to check that the previously created type has the expected kind. This will result in the variable getting the intersection of all its expected kinds -- assuming such an intersection exists.

lpw25 (commented):
I've reviewed the typing changes but I haven't yet looked at the test changes.

ocaml/typing/typetexp.ml
```diff
@@ -694,26 +718,27 @@ and transl_type_aux env ~row_context ~aliased ~policy mode styp =
...
       let desc, typ =
-        transl_type_var env ~policy ~row_context
+        transl_type_var env ~policy ~jkind_check ~row_context
           styp.ptyp_attributes styp.ptyp_loc name None
       in
       ctyp desc typ
   | Ptyp_arrow _ ->
       let args, ret, ret_mode = extract_params styp in
       let rec loop acc_mode args =
+        let jkind = Jkind.Type.Primitive.any ~why:Inside_of_Tarrow |> Jkind.of_type_jkind in
```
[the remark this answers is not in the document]
jbachurski: added in summary `higher-kinds` branch

jbachurski:
Subsumed by #3024

PR #2892, opened 2024-08-05
Coalesce duplicate spills and reloads per CFG block

milan-tom:
At present, if register allocation fails to register allocate a variable it is reloaded/spilled from the stack for every use throughout the duration of the whole program (either via creation of a new temporary register for that use or direct use in an instruction if the architecture supports this). This creates situations where even in a single CFG block the same variable may be reloaded and spilled many times. This can be avoided by using a register for that variable for the duration of a single block. This PR aims to do this by optimising the temporaries created after the first round of register allocation (only for this round to avoid infinite loops and the more rounds this optimisation is involved in the longer register allocation takes) to ensure at most one temporary register is used per variable in each CFG block.

# Analysis of improvement (tested on compiler files)
# Files
|        counter         | better (%) | worse (%) | equal (%) |
| ---------------------- | ---------: | --------: | --------: |
| reload                 |      77.30 |      2.04 |     20.66 |
| spill                  |      32.92 |      8.06 |     59.02 |
| block_duplicate_reload |      70.15 |      2.04 |     27.81 |
| block_duplicate_spill  |      37.00 |      0.00 |     63.00 |
# Functions
|        counter         | better (%) | worse (%) | equal (%) |
| ---------------------- | ---------: | --------: | --------: |
| reload                 |      30.02 |      0.93 |     69.05 |
| spill                  |       3.26 |      1.65 |     95.08 |
| block_duplicate_reload |      18.61 |      0.83 |     80.56 |
| block_duplicate_spill  |       3.27 |      0.01 |     96.72 |

NB: Timing information is not analysed here as the files in the compiler are too small to do meaningful analyses of the worst cases.

The files/functions that have worse counts afterwards are due to assigning block temporaries for variables which have live ranges across a block that interfere with other registers which have already been register allocated. Future PR(s) will address this by disabling/tweaking the optimization in this case.

[Full analysis](https://gist.github.com/milan-tom/2a768f7e499526298adb8c0cf9091eb8) (preview below)
comparison

### DIFF
--- a/backend/regalloc/regalloc_irc.ml
+++ b/backend/regalloc/regalloc_irc.ml
@@ -281,8 +281,7 @@ let select_spilling_register_using_heuristics : State.t -> Reg.t =
          the introduced temporaries which, by construct, have very few
          occurrences. *)
       +.
-      if State.mem_introduced_temporaries state reg
-         && State.get_round_num state != 2
+      if State.mem_inst_temporaries state reg
       then 10_000.
       else 0.
     in
@@ -395,13 +394,15 @@ let rewrite :
     reset:bool ->
     bool =
  fun state cfg_with_infos ~spilled_nodes ~reset ->
-  let new_temporaries, block_inserted =
+  let new_inst_temporaries, new_block_temporaries, block_inserted =
     Regalloc_rewrite.rewrite_gen
       (module State)
       (module Utils)
       state cfg_with_infos ~spilled_nodes
   in
-  if State.get_round_num state = 2 || new_temporaries <> []
+  State.add_inst_temporaries_list state new_inst_temporaries;
+  let new_temporaries = new_inst_temporaries @ new_block_temporaries in
+  if new_temporaries <> []
   then Cfg_with_infos.invalidate_liveness cfg_with_infos;
   if block_inserted
   then Cfg_with_infos.invalidate_dominators_and_loop_infos cfg_with_infos;
--- a/backend/regalloc/regalloc_irc.ml
+++ b/backend/regalloc/regalloc_irc.ml
@@ -389,14 +389,16 @@ let rewrite :
     Cfg_with_infos.t ->
     spilled_nodes:Reg.t list ->
     reset:bool ->
+    should_coalesce_temp_spills_and_reloads:bool ->
     bool =
- fun state cfg_with_infos ~spilled_nodes ~reset ->
+ fun state cfg_with_infos ~spilled_nodes ~reset
+     ~should_coalesce_temp_spills_and_reloads ->
   let new_inst_temporaries, new_block_temporaries, block_inserted =
     Regalloc_rewrite.rewrite_gen
       (module State)
       (module Utils)
       state cfg_with_infos ~spilled_nodes
-      ~should_coalesce_temp_spills_and_reloads:(State.get_round_num state = 1)
+      ~should_coalesce_temp_spills_and_reloads
   in
   let new_temporaries = new_block_temporaries @ new_inst_temporaries in
   if new_temporaries <> []
@@ -420,9 +422,8 @@ let rewrite :
    seems to be fine with 4 *)
 let max_rounds = 50
 
-let rec main : State.t -> Cfg_with_infos.t -> unit =
- fun state cfg_with_infos ->
-  let round = State.get_round_num state in
+let rec main : round:int -> State.t -> Cfg_with_infos.t -> unit =
+ fun ~round state cfg_with_infos ->
   if round > max_rounds
   then
     fatal "register allocation was not succesful after %d rounds (%s)"
@@ -494,12 +495,14 @@ let rec main : State.t -> Cfg_with_infos.t -> unit =
     then
       List.iter spilled_nodes ~f:(fun reg ->
           log ~indent:1 "/!\\ register %a needs to be spilled" Printmach.reg reg);
-    match rewrite state cfg_with_infos ~spilled_nodes ~reset:true with
+    match
+      rewrite state cfg_with_infos ~spilled_nodes ~reset:true
+        ~should_coalesce_temp_spills_and_reloads:(round = 1)
+    with
     | false -> if irc_debug then log ~indent:1 "(end of main)"
     | true ->
       State.invariant state;
-      State.incr_round_num state;
-      main state cfg_with_infos)
+      main ~round:(succ round) state cfg_with_infos)
 
 let run : Cfg_with_infos.t -> Cfg_with_infos.t =
  fun cfg_with_infos ->
@@ -529,9 +532,12 @@ let run : Cfg_with_infos.t -> Cfg_with_infos.t =
     List.iter spilled_nodes ~f:(fun reg -> State.add_spilled_nodes state reg);
     (* note: rewrite will remove the `spilling` registers from the "spilled"
        work list and set the field to unknown. *)
-    let (_ : bool) = rewrite state cfg_with_infos ~spilled_nodes ~reset:false in
+    let (_ : bool) =
+      rewrite state cfg_with_infos ~spilled_nodes ~reset:false
+        ~should_coalesce_temp_spills_and_reloads:false
+    in
     ());
-  main state cfg_with_infos;
+  main ~round:1 state cfg_with_infos;
   if irc_debug then log_cfg_with_infos ~indent:1 cfg_with_infos;
   Regalloc_rewrite.postlude
     (module State)
--- a/backend/regalloc/regalloc_irc.ml
+++ b/backend/regalloc/regalloc_irc.ml
@@ -280,7 +280,11 @@ let select_spilling_register_using_heuristics : State.t -> Reg.t =
       (* note: while this magic constant is questionable, it is key to not favor
          the introduced temporaries which, by construct, have very few
          occurrences. *)
-      +. if State.mem_introduced_temporaries state reg then 10_000. else 0.
+      +.
+      if State.mem_introduced_temporaries state reg
+         && State.get_round_num state != 2
+      then 10_000.
+      else 0.
     in
     match State.is_empty_spill_work_list state with
     | true -> fatal "spill_work_list is empty"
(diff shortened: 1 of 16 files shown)


backend/regalloc/regalloc_rewrite.ml
```diff
@@ -37,6 +40,54 @@ type direction =
...
+        | Some prev_inst_cell -> redundant := prev_inst_cell :: !redundant);
+        Reg.Tbl.replace last_spill var inst_cell;
+        match Reg.Tbl.find_opt var_to_temp_for_all var with
+        | None -> Reg.Tbl.add var_to_temp_for_all var temp
+        | Some temp_for_all -> replace temp temp_for_all)
+      | _ -> ()
+    in
+    DLL.iter_cell block.body ~f:update_info_using_inst;
+    List.iter ~f:DLL.delete_curr !redundant;
+    Substitution.apply_block_in_place replacements block;
```
xclerc, line 79: I am not claiming it will make a
measurable difference, but we
could maybe have fast path for
the case the substitution ends up
being empty.

backend/regalloc/regalloc_rewrite.ml
```diff
@@ -37,6 +40,54 @@ type direction =
...
+        | None -> ()
+        | Some prev_inst_cell -> redundant := prev_inst_cell :: !redundant);
+        Reg.Tbl.replace last_spill var inst_cell;
+        match Reg.Tbl.find_opt var_to_temp_for_all var with
+        | None -> Reg.Tbl.add var_to_temp_for_all var temp
+        | Some temp_for_all -> replace temp temp_for_all)
+      | _ -> ()
+    in
+    DLL.iter_cell block.body ~f:update_info_using_inst;
+    List.iter ~f:DLL.delete_curr !redundant;
```
xclerc, line 78: If it is not more difficult to remove
the instructions as we go, that might
be clearer that way. If it is more
difficult, we should just have a
comment explaining why.

xclerc:
(I have not read 96afa85 yet.)

backend/regalloc/regalloc_irc.ml
```diff
@@ -280,7 +280,11 @@ let select_spilling_register_using_heuristics : State.t -> Reg.t =
       (* note: while this magic constant is questionable, it is key to not favor
          the introduced temporaries which, by construct, have very few
          occurrences. *)
-      +. if State.mem_introduced_temporaries state reg then 10_000. else 0.
+      +.
+      if State.mem_introduced_temporaries state reg
+         && State.get_round_num state != 2
+      then 10_000.
+      else 0.
```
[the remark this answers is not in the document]
milan-tom: This has been resolved by separating the new temporaries created by `rewrite_gen` into instruction- and block-level temporaries.

backend/regalloc/regalloc_rewrite.ml
```diff
@@ -48,13 +99,15 @@ let rewrite_gen :
  fun (module State : State with type t = s) (module Utils) state cfg_with_infos
      ~spilled_nodes ->
   if Utils.debug then Utils.log ~indent:1 "rewrite";
+  let should_coalesce_temp_spills_and_reloads = State.get_round_num state = 1 in
```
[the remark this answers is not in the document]
milan-tom: The caller of `rewrite_gen` is now able to specify whether the optimization should be allowed via the Boolean `block_temporaries` parameter passed into `rewrite_gen`.

backend/regalloc/regalloc_rewrite.ml
```diff
@@ -48,13 +99,15 @@ let rewrite_gen :
...
   let block_insertion = ref false in
   let spilled_map : Reg.t Reg.Tbl.t =
     List.fold_left spilled_nodes ~init:(Reg.Tbl.create 17)
       ~f:(fun spilled_map reg ->
         if Utils.debug then assert (Utils.is_spilled reg);
         let spilled = Reg.create reg.Reg.typ in
-        Utils.set_spilled spilled;
+        if not should_coalesce_temp_spills_and_reloads
+        then Utils.set_spilled spilled;
```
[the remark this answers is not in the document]
milan-tom: This has been removed.

backend/regalloc/regalloc_rewrite.ml
```diff
@@ -157,44 +210,46 @@ let rewrite_gen :
           let instr = DLL.value cell in
           if instruction_contains_spilled instr
           then
-            match Regalloc_stack_operands.basic spilled_map instr with
-            | All_spilled_registers_rewritten -> ()
-            | May_still_have_spilled_registers ->
+            if should_coalesce_temp_spills_and_reloads
+               || Regalloc_stack_operands.basic spilled_map instr
+                  = May_still_have_spilled_registers
```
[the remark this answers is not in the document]
milan-tom: Yes, it is necessary to short-circuit here. If the we are coalescing the temporary spills and reloads, we do not want to allow the spilled registers to be directly used in the instruction argument or result arrays.

backend/regalloc/regalloc_rewrite.ml
milan-tom: Changed to delete as we iterate over the instructions in the block

backend/regalloc/regalloc_irc.ml
```diff
@@ -389,16 +389,21 @@ let rewrite :
...
-  let new_temporaries, block_inserted =
+ fun state cfg_with_infos ~spilled_nodes ~reset ~block_temporaries ->
+  let new_inst_temporaries, new_block_temporaries, block_inserted =
     Regalloc_rewrite.rewrite_gen
       (module State)
       (module Utils)
-      state cfg_with_infos ~spilled_nodes
+      state cfg_with_infos ~spilled_nodes ~block_temporaries
   in
+  let new_temporaries = new_block_temporaries @ new_inst_temporaries in
```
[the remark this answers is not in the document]
milan-tom: This was used in reset (in the match statement below this line but I have refactored to eliminate the need to concatenate these two lists.

backend/regalloc/regalloc_rewrite.ml
```diff
@@ -37,16 +38,86 @@ type direction =
...
+   the first used temporary for each variable to all the other temporaries used
+   later for that variable, deleting now redundant reload/spill instructions
+   along the way. This currently does not support the case where spilled nodes
+   are used directly in instructions (if allowed by the ISA) without a new
+   temporary being created. This optimization does not use spills introduced in
+   inserted blocks due to spilling in the terminator to remove spills in the
+   original block. Returns the new block temporaries and removes the now block
+   temporaries and the instruction temporaries for now removed instructions from
+   [new_temporaries]. *)
+let coalesce_temp_spills_and_reloads cfg_with_infos new_temporaries =
```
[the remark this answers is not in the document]
milan-tom: Changed code to only apply optimization on blocks that have been rewritten

backend/regalloc/regalloc_rewrite.ml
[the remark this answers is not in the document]
milan-tom: If `should_coalesce_temp_spills_and_reloads` is true at this point, we will definitely coalesce the temporaries in this block as some instruction must contain a spilled register (same for the condition on the terminator). If the block temporary we create gets register allocated, then that is better than using stack operands to use the memory address directly in the instruction. If the block temporary is spilled, stack operand will apply to it in the next round in the same way it would have done to the original variable. So, I think we are fine to short circuit here and will add a comment explaining why.

backend/regalloc/regalloc_rewrite.ml
```diff
@@ -37,16 +37,81 @@ type direction =
   | Load_after_list of Cfg.basic_instruction_list
   | Store_before_list of Cfg.basic_instruction_list
 
+(* Applies an optimization on the CFG outputted by [rewrite_gen] having one
+   temporary per variable per block rather than one per use of the variable,
+   reducing the number of spills and reloads needed for variables used multiple
+   times in a block. It iterates over each block and builds a substitution from
+   the first used temporary for each variable to all the other temporaries used
+   later for that variable, deleting now redundant reload/spill instructions
+   along the way. This currently does not support the case where spilled nodes
```
TheNumbat, line 46: Does not supported imply that the instruction-level spill is untouched for instructions that can use a spilled operand?

backend/regalloc/regalloc_rewrite.ml
```diff
@@ -37,16 +37,81 @@ type direction =
...
+   are used directly in instructions (if allowed by the ISA) without a new
+   temporary being created. This optimization does not use spills introduced in
+   inserted blocks due to spilling in the terminator to remove spills in the
+   original block. Returns the new block temporaries and removes the now block
+   temporaries and the instruction temporaries for now removed instructions from
+   [new_temporaries]. *)
```
TheNumbat, line 52: I'm not quite sure what the last two sentences mean:
- spills in terminators are untouched because their purpose is to reduce spills in the original block
- new block-temporaries are both removed from the block and returned, and this removes temps that were only used in removed instruction-level spill/reload pairs
Is that right?

TheNumbat (commented):
LGTM, Xavier's comments already cover my thoughts

backend/regalloc/regalloc_rewrite.ml
milan-tom: This optimization is unsound when spilled nodes are used directly in instructions (if allowed by the ISA) without a new temporary being created (hence stack operands are not used in `rewrite_gen` if this optimization is enabled).

I have updated the comment.

backend/regalloc/regalloc_rewrite.ml
milan-tom: Spills in blocks inserted by `rewrite_gen` (due to spills in block terminators) do not share temporaries with the body of the original block and are hence not considered in this optimization.

milan-tom: No new temporaries are created by this optimization. Some temporaries are promoted to block temporaries (and so moved from the list of new instruction temporaries to the list of new block temporaries). Instruction temporaries that are now redundant (due to being replaced by block temporaries) are removed from the list of new instruction temporaries.

milan-tom: I have updated the comment using these explanations to hopefully make it clearer.

xclerc (approved):

milan-tom:
Completed final checks and confirmed changes when optimization is disabled are only in files modified in this PR.

PR #2893, opened 2024-08-05
`new` types

jbachurski:
```ocaml
let x = (42, 1337)
[%%expect {|
val x : int * int = (42, 1337)
|}]

type 'a l = new int * 'a
let x1 = (x :> int l)
[%%expect {|
type 'a l = new int * 'a
val x1 : int l = (42, 1337)
|}]
```

Types created with `new` are recognised as distinct from their manifest, but can be coerced to and from.
Because of this property they can be used as datatypes (they have a higher jkind).


jbachurski:
Unfortunately, because `private_flag` is defined in `Asttypes`, CI fails due to `Jane_syntax` things.

jbachurski:
Subsumed by #3036

PR #2895, opened 2024-08-06
Avoid generating empty `profile.dump` files

milan-tom:
Prevents empty `profile.dump` file being created if `-dump-into-file` passed in without any profile flags.


xclerc (approved):
@gretay-js should probably also have a look.

gretay-js (approved):
#

PR #2901, opened 2024-08-07
Permit layouts to be (polarised) unions of sorts

jbachurski:
- Introducing higher kinds introduces contravariant positions, which in particular means layouts need to also have joins (lattice and not just a meet semi-lattice).
- Implementing joins for layouts per `| Sort s1, Sort s2 -> if Sort.equate s1 s2 then Sort s1 else Any` is incomplete and order-dependent.
- A way forward is to have a case of layout of `Union of Sort.t list`. In general, this is a union of variables possibly lower-bounded by a constant. This is what some functions that want to introspect the current layout need to handle now, where previously they may have only seen that a jkind is non-constant and contains some one sort variable.
- This union representation has to be occasionally normalised (`repr`), with `get`s extracting the maximum amount of information about the current state. There are no errors introduced, as a union of mismatching sorts is just `Any`.
- Where a union is used in a non-positive position (negative for intersections and superkinding, or zero for equating), a fatal error is raised, as such situations are unexpected and means we are mistreating polarisations.

This should lead to a fully legitimate jkind inference rather than an order-dependent heuristic in the case of meeting jkinds over contravariant positions (i.e. joins). This had to be resolved because many seemingly simple examples end up leading to barely-non-trivial joins, so physical equality was not enough as a heuristic.

**TODO:** it would be nice to write down an actual example where order-dependence is exhibited in the previous system.

### DIFF
--- a/ocaml/testsuite/tests/typing-higher-jkinds/folds.ml
+++ b/ocaml/testsuite/tests/typing-higher-jkinds/folds.ml
@@ -95,10 +95,10 @@ val result : int list id = {v = [0; 1; 2]}
 |}]
 
 let leaves (perfect : 'a perfect) =
-  fold ({
+  fold {
     zero = (fun l -> [l]);
     succ = (fun l -> List.concat_map (fun (x, y) -> [x; y]) l)
-  } : list folder) perfect
+  } perfect
 let result = leaves example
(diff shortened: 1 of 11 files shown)


jbachurski:
Subsumed by #3024

PR #2903, opened 2024-08-07
Infer tailcalls for functions bound by fn arguments

tylerhou:
Here is the new tailcall graph. The main difference is that many of the calls to `<unknown>` in the bottom SCC are now blue since we infer them as tailcalls.

image


dvulakh (approved):
Looks good to me, thanks!

PR #2904, opened 2024-08-08
Apply improvements to the kind polymorphism design

goldfirere:
This PR encompasses little improvements suggested by various colleagues (as noted in commit messages). I recommend reviewing commit by commit.

Review requested from @stedolan, who gave some very helpful feedback and has already looked at these rules in detail.

### DIFF
--- a/ocaml/jane/doc/proposals/kind-inference.md
+++ b/ocaml/jane/doc/proposals/kind-inference.md
@@ -681,6 +681,13 @@ rec {
 
 ```
 
+# Properties
+
+Lemma (Borrowing is sound): If `Γ ⊢ σ : κ {best}` and `Γ ⊢ κ ≤ any mod global`,
+then `Γ ⊢ κ ≤ any mod unique`.
+
+Proof: TODO.
+
 # Examples
 
 ```
--- a/ocaml/jane/doc/proposals/kind-inference.md
+++ b/ocaml/jane/doc/proposals/kind-inference.md
@@ -110,22 +110,28 @@ rec {
 
 Ξ ::= locality | onceness | uniqueness | contention | portability | externality
 layout ::= layout_of σ | const_layout
-κ ::= layout; ⟪m_Ξ with field_types⟫  (* syntax with ⟪ and ⟫ defined below *)
+
+  (* syntax with ⟪ and ⟫ defined below *)
+χ ::= layout; ⟪m_Ξ⟫                   (* translation of a [jkind] *)
+κ ::= layout; ⟪m_Ξ with field_types⟫  (* translation of a [jkind_scheme] *)
 (* we may leave out a [with] when there are no types afterwards *)
+(* The syntax for [χ] is a subsyntax for [κ]:
+     In premises, we may write [κ] where the grammar suggests we write [χ].
+     In conclusions, we may write [χ] where the grammar suggests we write [κ]. *)
 
 δ ::= (* left abstract; internal representation of a [type_kind] *)
 
-tconstr_jkind ::= π [[ 'a : κ ]]. κ₀
+tconstr_jkind ::= π [[ 'a : χ ]]. κ₀
   (* That should be a capital π, but that looks too much like ⨅. *)
 
 rec {
   type_binding ::=
-    | 'a : κ
+    | 'a : χ
     | t : tconstr_jkind            (* abstract type *)
-    | t = λ [[ 'a : κ ]]. τ        (* type abbreviation *)
-    | t := λ [[ 'a : κ ]]. τ       (* type substitution in a signature *)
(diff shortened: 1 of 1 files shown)


stedolan:
I haven't read it all yet but the new `SUB` and `MB_*` rules look good to me. It's missing one almost trivial rule, though: I need to be able to compute joins of concrete modes on the right, by something like:
```
Γ ⊢ mbi ≤ modal_bound₂_Ξ ⊔ (mode (m₁_Ξ ⊔ m₂_Ξ))
------------------------------------------------ MB_JOIN_R
Γ ⊢ mbi ≤ modal_bound₂_Ξ ⊔ (mode m₁_Ξ) ⊔ (mode m₂_Ξ) 
```
This rule is actually doing something (barely): you might have mbi less than the join of two concrete modes appearing on the right, but not less than either by itself. So you need to actually compute the join on the right in order to compare it to the left with MB_MODE.

goldfirere:
I think I've addressed the concern above, but please double-check, as the rule I wrote is not exactly the rule you suggested.

Would you mind taking a deeper look at the rest of the changes here to see if we can merge this, @stedolan? Thanks!

ocaml/jane/doc/proposals/kind-inference.md
```diff
@@ -197,22 +200,23 @@ k = κ ∈ Γ
...
-Γ ⊢ kind_of_ τ ↠ layout_of τ; ⟪⊥_Ξ with τ⟫
-
 Γ ⊢ jkind ↠ κ
 ------------------------------------------------- K_MOD
 Γ ⊢ jkind mod modes ↠ lay(κ); ⟪Ξ(κ) ⊓ ⨅ extract(Ξ, modes)⟫
-  (* CR reisenberg: This rule isn't right if κ already has [with] constraints; probably
-     should just forbid that syntactically. Not right: we can't meet out
-     the with-added modes because we don't really know what they are.
-     So we'd have to retain the [mod] clause just like we retain the
-     [with] clause, which is terrible. *)
```
stedolan, line 211: I have the same question, and I don't understand how it was resolved by this change. What does happen if κ already has [with] constraints?

stedolan:
(apologies this took so long to get to)

The new MB_* rules look good to me. I'm curious about what's going on with the deleted CR I commented on above, though: essentially, you have a join operation (`with`) and a meet operation (`mod`), and I would like to understand whether and how they interact.

ocaml/jane/doc/proposals/kind-inference.md
goldfirere: The user-written syntax indeed prevents `with` constraints from ending up in the \kappa here. But the system does not track this. I have split out \chi from \kappa, where \chi does not have `with` constraints. This is a nice improvement. I also may have noticed a few other small bugs that I fixed while taking a pass through.

lpw25:
Maybe this is discussed in the doc already, but perhaps it is worth me pointing out that `mod` is not really a meet. It represents a morphism on modes that is either meet-with-constant for comonadic modes or join-with-constant for monadic modes.

goldfirere:
The doc is silent about mode-crossing (somewhat paradoxically, because it describes a system that's all about mode-crossing). It just defines subkind operations. My last commit reverses the order of elements in the monadic axes and now includes some commentary about this potentially-confusing point.

stedolan:
> It's not a general binary meet, but meet-with-constant is still a meet. In particular, the syntax `k mod observing with σ` seems to be allowed in a jkind_scheme, and I don't fully understand what it means or how it is converted to a modal_bound. (Does the join distribute over the meet?)

Ah, from a closer look at `K_MOD` I see that (a) `mod` is completely evaluated during initial translation to `χ` and (b) the left hand side of `mod` must also be a constant (a `const_layout`, another `mod` expression, or a kind abbrevation `k`, but _not_ `kind_of_`). This certainly simplifies things, but will this simplicity survive the addition of mode/kind polymorphism? (I have been assuming that eventually kind abbreviations `k` will generalise to include kind variables that cannot be trivially expanded).

goldfirere:
I think @lpw25's comment meant to say that `mod` is not *always* a meet. That is, the interpretation of `mod` with monadic axes is not a meet at all, but a `subtract`.

Mode polymorphism does not intersect the design in this doc. Mode polymorphism is a feature of terms, not of types. Maybe someday we'll allow types to be parameterized on modes -- and then perhaps we'll allow the modes to appear in kind-level modal bounds -- but there are no plans to add this, and I don't think we need to make the design robust against that possibility now.

There are additionally no plans for kind variables of the form you're worried about. Perhaps unhelpfully, the ideas in the doc as written are called "kind polymorphism". I clearly need a different name for this suite of features, because this is a common source of confusion. Suggestions welcome.

goldfirere:
Are there remaining open questions here? Or should we merge?

stedolan (approved):

PR #2906, opened 2024-08-08
Unboxed load/store primitives for string-likes

dvulakh:
Add primitives for:
- Load/store
- From: `bigstring`, `string`, and `bytes`
- Index by: `int`, `int32#`, `int64#`, `nativeint#`
- Data width: `int16`, `int32`, `float32`, `int64`, `int128` (both boxed and unboxed when possible)

I add tests of operation over `int128` to the `simd/arrays.ml` test. `tests/typing-layouts/generate_stringlike_indexing.ml` generates a test (`tests/typing-layouts/unboxed_int_stringlike_indexing.ml`) that stresses the remaining primitives.

### DIFF
--- a/ocaml/lambda/lambda.mli
+++ b/ocaml/lambda/lambda.mli
@@ -222,34 +222,53 @@ type primitive =
   (* size of the nth dimension of a Bigarray *)
   | Pbigarraydim of int
   (* load/set 16,32,64,128 bits from a string: (unsafe)*)
-  | Pstring_load_16 of bool
-  | Pstring_load_32 of bool * alloc_mode
-  | Pstring_load_f32 of bool * alloc_mode
-  | Pstring_load_64 of bool * alloc_mode
-  | Pstring_load_128 of { unsafe : bool; mode: alloc_mode }
-  | Pbytes_load_16 of bool
-  | Pbytes_load_32 of bool * alloc_mode
-  | Pbytes_load_f32 of bool * alloc_mode
-  | Pbytes_load_64 of bool * alloc_mode
-  | Pbytes_load_128 of { unsafe : bool; mode: alloc_mode }
-  | Pbytes_set_16 of bool
-  | Pbytes_set_32 of bool
-  | Pbytes_set_f32 of bool
-  | Pbytes_set_64 of bool
-  | Pbytes_set_128 of { unsafe : bool }
+  | Pstring_load_16 of { unsafe : bool; index_kind : array_index_kind }
+  | Pstring_load_32 of { unsafe : bool; index_kind : array_index_kind;
+      mode : alloc_mode; boxed : bool }
+  | Pstring_load_f32 of { unsafe : bool; index_kind : array_index_kind;
+      mode : alloc_mode; boxed : bool }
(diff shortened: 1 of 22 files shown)


ocaml/runtime/bigarray.c
```diff
@@ -868,6 +868,51 @@ CAMLprim value caml_ba_uint8_set64(value vb, value vind, value newval)
...
+  }                                                                            \
+  CAMLprim value caml_ba_uint8_set##width##_indexed_by_##name(value vb,        \
+      value vind, value newval)                                                \
+  {                                                                            \
+    index_type idx = val_func(vind);                                           \
+    if (idx != Long_val(Val_long(idx))) caml_array_bound_error();              \
+    return caml_ba_uint8_set##width(vb, Val_long(idx), newval);                \
+  }
+
+CAMLprim_indexed_by(16, int64, int64_t, Int64_val)
```
mshinwell, line 906: Just in case there are any errors in the script that greps for `CAMLprim`, I think I'd prefer this macro not to have `CAMLprim` in its name.

ocaml/lambda/lambda.mli
```diff
@@ -222,34 +222,53 @@ type primitive =
...
-  | Pbigstring_load_64 of { unsafe : bool; mode: alloc_mode; boxed : bool }
-  | Pbigstring_load_128 of { aligned : bool; unsafe : bool; mode: alloc_mode;
+  | Pbigstring_load_16 of { unsafe : bool; index_kind : array_index_kind }
+  | Pbigstring_load_32 of { unsafe : bool; index_kind : array_index_kind; mode :
+      alloc_mode; boxed : bool }
+  | Pbigstring_load_f32 of { unsafe : bool; index_kind : array_index_kind;
+      mode : alloc_mode; boxed : bool }
+  | Pbigstring_load_64 of { unsafe : bool; index_kind : array_index_kind; mode :
+      alloc_mode; boxed : bool }
+  | Pbigstring_load_128 of { aligned : bool; unsafe : bool; index_kind :
```
mshinwell, line 261: Some of this formatting seems weird.  Maybe take this type declaration and paste it into an autoformatted file, format it, then paste just this section back?

mshinwell:
Looks ok in outline, @TheNumbat will review in detail.

ocaml/lambda/lambda.mli
dvulakh: `ocamlformat` puts each record field onto its own line. I can adopt this format if we prefer, but it doesn't seem to match the rest of the file. So for now I've accepted `ocamlformat`'s output, but combined record fields onto one line when possible (without splitting on `:`).

dvulakh:
I noticed a problem with the code generated by these primitives. For example, `%caml_string_get64u#_indexed_by_int64#` generates:
```
salq $1, %rbx
sarq $1, %rbx
movq (%rax,%rbx), %rax
ret
```
This is because the flambda primitives operate on tagged indices. I'll post a fix in another PR; meanwhile I'm reopening review here.

TheNumbat (approved):
LGTM, adding the primitives is mostly mechanical after the refactoring, and the tests are quite thorough. I will review the improved codegen PR as well.

PR #2913, opened 2024-08-09
Rename float32 round_half_to_even to round_current

TheNumbat:
Float32 will soon be supported in javascript and wasm, where the default rounding mode is not half-to-even.
This PR renames the `round_half_to_even` functions to `round_current`, because they actually round using the current mode.
The comments have been updated as well.


ccasin (approved):
LGTM

PR #2915, opened 2024-08-09
Move float32 back to stable

TheNumbat:
To merge after javascript/wasm support is released.
To fix bytecode float32 constants in stdlib_stable, we will also need to stop compiling otherlibs with the boot compiler.


dkalinichenko-js (commented):
It would be good to add a test that `float32` and `float32#` primitive types are not defined with `-extension-universe upstream_compatible`.

dkalinichenko-js (approved):

TheNumbat:
Sure, added for both small_numbers and simd. The output is slightly different than `test_disabled` because `upstream_compatible` disables both extensions.

PR #2924, opened 2024-08-14
Support modalities on `include` in signatures

riaqn:
This PR adds the support for the syntax:
```
sig
  include S @@ m
end
```
where `m` is a list of modalities. The outcome is that for every `val foo : t @@ m'` in `S`, `val foo : t @@ m' ∘ m` is added to the current signature. `m' ∘ m` means "first apply m, then apply m'`.

Similar support for `include functor`.

Please review by commits.

This contains parsetree changes - **_please don't merge until I finish the internal migration._**

### DIFF
--- a/ocaml/testsuite/tests/typing-modes/val_modalities.ml
+++ b/ocaml/testsuite/tests/typing-modes/val_modalities.ml
@@ -463,3 +463,89 @@ Error: Signature mismatch:
          val f : int -> int @@ portable
        The second is portable and the first is not.
 |}]
+
+(* Including module type with modalities *)
+module type S = sig
+  val foo : 'a -> 'a
+
+  val bar : 'a -> 'a @@ nonportable
+
+  val baz : 'a -> 'a @@ portable
+end
+[%%expect{|
+module type S =
+  sig
+    val foo : 'a -> 'a
+    val bar : 'a -> 'a
+    val baz : 'a -> 'a @@ portable
+  end
+|}]
+
+module type S' = sig
+  include S @@ portable
+end
+[%%expect{|
+module type S' =
+  sig
+    val foo : 'a -> 'a @@ portable
+    val bar : 'a -> 'a @@ portable
+    val baz : 'a -> 'a @@ portable
+  end
+|}]
+
+module type S' = sig
+  include S @@ nonportable
+end
+[%%expect{|
+module type S' =
+  sig
+    val foo : 'a -> 'a
+    val bar : 'a -> 'a
+    val baz : 'a -> 'a @@ portable
+  end
+|}]
+
+(* Include functor module types with modalities *)
+module type S = functor (_ : sig end) -> sig
+  val foo : 'a -> 'a
+
+  val bar : 'a -> 'a @@ nonportable
+
+  val baz : 'a -> 'a @@ portable
+end
+[%%expect{|
+module type S =
+  sig end ->
+    sig
+      val foo : 'a -> 'a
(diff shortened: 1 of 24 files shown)


ocaml/testsuite/tests/typing-modes/val_modalities.ml
```diff
@@ -463,3 +463,89 @@ Error: Signature mismatch:
          val f : int -> int @@ portable
        The second is portable and the first is not.
 |}]
+
+(* Including module type with modalities *)
+module type S = sig
+  val foo : 'a -> 'a
+
+  val bar : 'a -> 'a @@ nonportable
```
riaqn, line 471: Being able to write `@@nonportable` here is confusing. Let's fix this in another PR.

lpw25 (approved):
Some small comments, but this LGTM

goldfirere:
I'm surprised that this is an operation on an `include`, as opposed to an operation on the signature itself. That is, would it work to have `S @@ m` work as part of the module type syntax? (This would be somewhat like the `with type` syntax.)

Also, since this affects the user-facing language, is there documentation to update? (I'm more concerned about keeping the team up-to-date than about the rest of Jane Street -- things are moving fast enough that even within the team we need some form of documetation, I think.)

riaqn:
> That is, would it work to have S @@ m work as part of the module type syntax? 

That can work, but similar to `with type` syntax it should work lazily, which takes more work.

> Also, since this affects the user-facing language, is there documentation to update?

The internal document "modes and modules" had been updated.

riaqn:
This is currently blocked by #2949

goldfirere:
> > That is, would it work to have S @@ m work as part of the module type syntax?
> 
> It's unclear what `S @@ m` means if `S` is a functor type. We can say it applies to the returned signature, but I find that biased.

This same problem exists with `with type` syntax:

```ocaml
module type S = functor (X : sig end) -> sig type t end

module type S2 = S with type t = int
```

produces

```
Error: This module type is not a signature
```

The new syntax could behave similarly.

> 
> Also, such a syntax should be lazy (similar to module strenghthening), which takes some extra efforts.

That sounds reasonable -- I don't know enough to compare the implementation challenges between implementing my suggestion and what you've done. But if the implementation here is meant as a way of getting this done more quickly (great!), would it make sense to make a ticket to expand the treatment? Part of why I'm asking is that I couldn't see myself, say, advocating for the design implemented here when upstreaming.

> 
> > Also, since this affects the user-facing language, is there documentation to update?
> 
> The internal document "modes and modules" had been updated.

Thanks -- it's helpful to see the new section on this feature. But that document also has descriptions of features that are not yet implemented, such as having e.g. `@portable` at the top of an mli file. My rationale for wanting the doucmentation on GitHub is so that we can always try to keep the in-GitHub documentation up to date with the implementation, so that we know what language we expect to be compiling. This doesn't seem to be much work to maintain... but maybe the problem is that there is already a big divergence between what we have implemented and what we have documented.

Just to be clear, the reason I frequently push on this point is that I've gotten caught here a few times, wanting to experiment with a feature but unsure of what is implemented with what syntax. That is, my concern is not theoretical/hygienic, but practical. Still, if there is already a divergence, I do see how it's a lot of work to fix, and would not make sense to address in this PR.

riaqn:
> would it make sense to make a ticket to expand the treatment?

done. Internal ticket 3433.

> But that document also has descriptions of features that are not yet implemented

I've amended that document, so now each section has either DONE or TODO annotated on it.

PR #2925, opened 2024-08-14
Share `transl_modalities` code with `transl_modifiers` and `transl_modes`

liam923:
#2730 introduced `typemodifier.ml`, which shares translation logic between jkinds and modes. But although the modality translation code was similar, it was not abstracted into `typemodifier.ml`. This PR does so.


ocaml/typing/typemode.ml
```diff
@@ -1,10 +1,6 @@
 open Location
 open Mode
 
-type error = Unrecognized_modality of string
-
-exception Error of Location.t * error
-
 let transl_mode_annots modes =
   Typemodifier.transl_mode_annots ~required_mode_maturity:Stable modes
```
riaqn, line 5: I think it's odd to have `transl_mode_annots` in `Typemodifier`. Because there are many places where mode annots is not modifying a type. Most common example is `let x @ local = ..`.

I think `transl_mode_annotes` should stay in `Typemode`.

ocaml/typing/jkind_axis.mli
```diff
@@ -0,0 +1,108 @@
+(**************************************************************************)
```
liam923, line 1: This file is a bit heavyweight for just `typemodifier.ml`. But I am working on another PR that utilizes it within `jkind.ml`.

ocaml/typing/jkind_axis.ml
```diff
@@ -0,0 +1,225 @@
...
+    let less_or_equal a b : Misc.Le_result.t =
+      match le a b, le b a with
+      | true, true -> Equal
+      | true, false -> Less
+      | false, _ -> Not_le
+
+    let equal a b = Misc.Le_result.is_equal (less_or_equal a b)
+  end
+
+  let get (type a) : a t -> (module Axis_s with type t = a) = function
```
riaqn, line 145: I'm slightly concerned about the first-class module here. It's later used by `with` syntax PR in `Bounds.sub`, which is hot-path? This could be costy.

One solution is to push `Axis.t` inside the module. That is, having this in a module:
```
val less_or_equal : t Axis.t -> t -> t -> Misc.Le_result.t
```

riaqn (approved):
Thanks for refactoring - I've a pushed a commit containing minor edits and some CRs (that will be done in the future). Feel free to merge once CI is green, thanks!

ocaml/typing/typemode.ml
```diff
@@ -1,12 +1,122 @@
...
+  | Duplicated_axis : _ Axis.t -> error
+  | Unrecognized_modifier : _ annot_type * string -> error
 
 exception Error of Location.t * error
 
-let transl_mode_annots modes =
-  Typemodifier.transl_mode_annots ~required_mode_maturity:Stable modes
+module Str_map = Map.Make (String)
+
+let modifiers =
```
riaqn: static mapping is better represented as pattern matching, which is faster.

ocaml/typing/typemode.ml
```diff
@@ -92,6 +181,28 @@ let mutable_implied_modalities (mut : Types.mutability) attrs =
     then monadic
     else monadic @ comonadic
 
+let locality_to_regionality : Mode.Locality.Const.t -> Mode.Regionality.Const.t
+    = function
+  | Local -> Local
+  | Global -> Global
+
+let transl_modality ~maturity { txt = Parsetree.Modality modality; loc } =
```
riaqn: This function should probably be moved to its original location (also to minimize diff)

ocaml/typing/typemode.ml
```diff
@@ -1,12 +1,122 @@
...
+
+type modal_or_nonmodal = |
+
+type modelike_annot_type =
+  | Mode
+  | Modality
+
+type 'm annot_type =
+  | Modifier : modal_or_nonmodal annot_type
+  | Mode_like : modelike_annot_type -> modal annot_type
```
riaqn: I took the liberty to collapse `modelike_annot_type` into this type.

riaqn (approved):

PR #2930, opened 2024-08-15
Use untagged indices for string-like access in flambda

dvulakh:
The primitives added in #2906 for string-like access via unboxed indices generate bad code. For example, `%caml_string_get64u#_indexed_by_int64#` generates:
```
salq $1, %rbx
sarq $1, %rbx
movq (%rax,%rbx), %rax
ret
```
This is because the flambda primitives for string-like load and store operate on tagged indices, and the untagging does not happen until conversion to cmm. So an unboxed index is first tagged when the primitive is translated to flambda, and then untagged by cmm. The simplifications in `cmm_helpers.ml` can't rewrite a left shift followed by a right shift, so the tag-untag sequence is not eliminated.

Flambda's array-access primitives also operate on tagged indices, but these indices are shifted further left when they're used, so the simplifications in `cmm_helpers.ml` are able to eliminate the right shift for untagging and combine the left shifts. As a result, `%array_unsafe_get_indexed_by_int64#` already generates the sensible:
```
movq (%rax,%rbx,8), %rax
ret
```

This PR changes the flambda primitives for string-like access to use untagged indices, so `%caml_string_get64u#_indexed_by_int64#` now generates
```
movq (%rax,%rbx), %rax
ret
```

For testing, I skimmed the output of
```bash
#!/bin/bash

fin=/tmp/a.ml
fout=/tmp/a.s

for index in int32 int64 nativeint
do
  for width in int16 int32 'int32#' float32 'float32#' int64 'int64#'
  do
    for container in string bigstring bytes
    do
      access_sigil=`echo $width | sed -E 's/(f?)(loat|int)([0-9]+)(#?)/\1\3u\4/'`
      access_type=`echo $width | sed 's/16//'`
      printf "open Bigarray\n`
             `type bigstring = (char, int8_unsigned_elt, c_layout) Array1.t\n`
             `external get_unsafe`
             ` : $container`
             ` -> ${index}#`
             ` -> $access_type`
             ` = \"%%caml_${container}_get${access_sigil}_indexed_by_${index}#\"\n`
             `external set_unsafe`
             ` : $container`
             ` -> ${index}#`
             ` -> $access_type`
             ` -> unit`
             ` = \"%%caml_${container}_set${access_sigil}_indexed_by_${index}#\"\n`
             `let[@inline never] get s i = get_unsafe s i\n`
             `let[@inline never] set s i v = set_unsafe s i v" > $fin
      $INSTALL_DIR/bin/ocamlopt.opt -S -c -O3 $fin
      echo "==== access $width in $container by ${index}# ===="
      sed -nE '/^caml.*__get_.*code/,/^\s*ret$/p' $fout
      sed -nE '/^caml.*__set_.*code/,/^\s*ret$/p' $fout
    done
  done
done
```
and the various combinations generally look right.

### DIFF
--- a/middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
+++ b/middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
@@ -358,6 +358,15 @@ let convert_index_to_tagged_int index (index_kind : Lambda.array_index_kind) =
              },
            index ))
 
+let convert_index_to_untagged_int index
+    (index_kind : Lambda.array_index_kind) =
+  let src : I_or_f.t =
+    match index_kind with
+    | Ptagged_int_index -> Tagged_immediate
+    | Punboxed_int_index bint -> standard_int_or_float_of_boxed_integer bint
+  in
+  H.Prim (Unary (Num_conv { src; dst = Naked_immediate }, index))
+
 let check_non_negative_imm imm prim_name =
   if not (Targetint_31_63.is_non_negative imm)
   then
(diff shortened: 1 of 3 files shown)


mshinwell:
An flambda2 dev will need to review this, I'll make sure that happens.

mshinwell:
I've asked @Gbury if he can review

Gbury (approved):

PR #2933, opened 2024-08-16
Infer jkinds in poly types instead of defaulting to value

jbachurski:
This is a (very early, mostly scratch) work in progress PR to infer jkinds when variables are explicitly quantified over in types like `'a : 'a -> 'a`, instead of defaulting them to value.
This is particularly nice when it comes to higher jkinds and spares writing explicit annotations that can be easily inferred in bidirectional style (#2890).


ocaml/testsuite/tests/typing-layouts-err-msg/value.ml
```diff
@@ -268,7 +268,8 @@ Error: This expression has type ('a : value)
        The layout of 'b is void
          because of the annotation on the type variable 'b.
        But the layout of 'b must overlap with value
-         because it is or unifies with an unannotated universal variable.
+         because it's a fresh unification variable,
+         defaulted to layout value.
```
lpw25, line 272: I think this error message should probably be improved. The reason of this history should mention that it comes from an unannotated universal variable. I don't know if reasons carry locations, but ideally it would point at the relevant location too.

ocaml/testsuite/tests/typing-layouts/annots.ml
```diff
@@ -877,23 +877,13 @@ module type S = sig
 end
 ;;
 [%%expect {|
-Line 2, characters 10-36:
-2 |   val f : 'a. 'a t2_imm -> 'a t2_imm
-              ^^^^^^^^^^^^^^^^^^^^^^^^^^
-Error: The universal type variable 'a was defaulted to have kind value.
-       But it was inferred to have kind immediate
-         because of the definition of t2_imm at line 1, characters 0-28.
+module type S = sig val f : ('a : immediate). 'a t2_imm -> 'a t2_imm end
```
lpw25, line 880: The title of this test section above is now out of date.

lpw25: Note that we are still printing `('a : immediate). ...` even though `'a. ...` would be fine. I don't know if we do the same for other sources of defaulting -- e.g. type parameters -- but it would be better if we only added annotations when needed. If there is such support for type parameters then you should use it here too.

ocaml/testsuite/tests/typing-layouts/basics.ml
```diff
@@ -670,11 +671,57 @@ Line 3, characters 4-34:
...
+let t6_float64_id : type a. a t6_float64 -> a t6_float64 = fun x -> x
+[%%expect{|
+Line 1, characters 28-29:
+1 | let t6_float64_id : type a. a t6_float64 -> a t6_float64 = fun x -> x
+                                ^
+Error: This type a should be an instance of type ('a : float64)
+       The layout of a is value
+         because it is or unifies with an unannotated universal variable.
+       But the layout of a must be a sublayout of float64
+         because of the definition of t6_float64 at line 1, characters 0-42.
```
lpw25, line 722: This doesn't look like the right error message to me. I think it should say something like "unannotated locally abstract type" rather than "unannotated universal variable".

ocaml/testsuite/tests/typing-layouts/basics.ml
```diff
@@ -654,7 +654,8 @@ Line 2, characters 2-32:
 Error: This definition has type 'b -> unit which is less general than
          'a. 'a -> unit
        The kind of 'a is value
-         because it is or unifies with an unannotated universal variable.
+         because it's a fresh unification variable,
+         defaulted to kind value.
```
lpw25, line 658: As elsewhere, this is not an accurate error message

ocaml/testsuite/tests/typing-layouts/basics.ml
```diff
@@ -670,11 +671,57 @@ Line 3, characters 4-34:
 Error: This method has type 'b -> unit which is less general than
          'a. 'a -> unit
        The kind of 'a is value
-         because it is or unifies with an unannotated universal variable.
+         because it's a fresh unification variable,
+         defaulted to kind value.
```
lpw25, line 675: As elsewhere, this is not an accurate error message

ocaml/testsuite/tests/typing-layouts/basics.ml
```diff
@@ -670,11 +671,57 @@ Line 3, characters 4-34:
...
+Line 2, characters 27-28:
+2 |   fun x -> let Foo y = Foo x in Foo y
+                               ^
+Error: This expression has type ('a : value)
+       but an expression was expected of type ('b : float64)
+       The layout of 'a is float64
+         because of the definition of t6_float64 at line 1, characters 0-42.
+       But the layout of 'a must overlap with value
+         because it's a fresh unification variable,
+         defaulted to layout value.
```
lpw25, line 703: As elsewhere, this is not an accurate error message

lpw25 (commented):
This mostly looks good, but there has been a regression in the quality of the error messages.

PR #2934, opened 2024-08-16
Add notes on fuzzing for Flambda2 types

lthls:
Some notes following a discussion with @chambart.


lthls:
Turned into a wiki page: https://github.com/ocaml-flambda/flambda-backend/wiki/Fuzzing-for-Flambda-2-types

PR #2936, opened 2024-08-16
Build otherlibs (and other installed artifacts) using our stdlib

dkalinichenko-js:
Split the last stage of our compiler build into two: `compiler` and `main`. `compiler` mimics `boot-compiler`, but uses our standard library instead of the upstream version to build the targets. `main` builds the rest of the installation using that compiler, reusing artifacts already built.

This is necessary to compile `Stdlib_alpha.Or_null` in bytecode correctly: the upstream standard library can't marshal null constants. This is also helpful for supporting `float32` in bytecode.

Note that `unix` is still compiled using `boot-compiler`, since `compiler` depends on `unix`.

It's complicated to add a test for this, since `or_null` in bytecode is not implemented yet, but I successfully built my WIP bytecode branch with this change. I also inspected `_build/log` to ensure it uses the right compiler.

cc @TheNumbat

### DIFF
--- a/Makefile.common-jst
+++ b/Makefile.common-jst
@@ -2,9 +2,10 @@
 # Build rules common to ocaml-jst and flambda-backend
 #
 
-ws_boot   = --root=. --workspace=duneconf/boot.ws
-ws_runstd = --root=. --workspace=duneconf/runtime_stdlib.ws
-ws_main   = --root=. --workspace=duneconf/main.ws
+ws_boot     = --root=. --workspace=duneconf/boot.ws
+ws_runstd   = --root=. --workspace=duneconf/runtime_stdlib.ws
+ws_compiler = --root=. --workspace=duneconf/compiler.ws
+ws_main     = --root=. --workspace=duneconf/main.ws
 
 ifeq ($(coverage),yes)
   coverage_dune_flags=--instrument-with bisect_ppx
@@ -38,7 +39,7 @@ define dune_runtime_stdlib_context
     (env-vars ("OCAMLPARAM" "$(BUILD_OCAMLPARAM)"))))))
 endef
 
-define dune_main_context
+define dune_compiler_context
 (lang dune 2.8)
 (context (default
   (name main)
@@ -51,9 +52,21 @@ define dune_main_context
     (env-vars ("OCAMLPARAM" "$(BUILD_OCAMLPARAM)"))))))
 endef
 
+define dune_main_context
+(lang dune 2.8)
+(context (default
+  (name main)
+  (profile $(main_build_profile))
+  (paths
+    (PATH ("$(CURDIR)/_build/compiler_install/bin" :standard))
+    (OCAMLLIB ("$(CURDIR)/_build/runtime_stdlib_install/lib/ocaml_runtime_stdlib")))
+  (env (_
+    (flags (:standard -warn-error +A))
+    (env-vars ("OCAMLPARAM" "$(BUILD_OCAMLPARAM)"))))))
+endef
 
-.DEFAULT_GOAL := compiler
-.PHONY: boot-compiler boot-runtest runtime-stdlib compiler runtest
+.DEFAULT_GOAL := main
+.PHONY: boot-compiler boot-runtest runtime-stdlib compiler main runtest
 
 boot_targets = \
   $(boot_ocamlc) \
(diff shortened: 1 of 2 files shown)


mshinwell:
I'll look at this - there have been some problems with 5.2 which might be related.

dkalinichenko-js:
I'm looking into the CI failures, can't reproduce them on my machine so far.

mshinwell:
@dkalinichenko-js can you please get CI to pass on this PR?

dkalinichenko-js:
Seems like Dune clears the `_build/main` default directory between builds in the CI for some reason, while it doesn't locally. I fixed this by hardlinking the resulting compiler instead of symbolic linking. This is not ideal, since we want Dune to reuse already built artifacts in `_build/main`, but should be sufficient for testing.

mshinwell:
Can we try to do something to avoid hardlinking tricks, which can be fragile and difficult to understand?

mshinwell:
(I think the change in behaviour depends on the Dune version, we've encountered this elsewhere iirc)

dkalinichenko-js:
I'll try that, agree it's fragile and bad.

dkalinichenko-js:
I updated the PR to build actual final targets instead of boot targets, because boot targets don't depend on our version of unix and it's very tricky to make them to. I also removed hardlinking in favor of copying. I'm not quite happy with this, since we perform duplicate work if Dune clears `_build/main` between builds, but if the impact on build times is small this should unblock various bytecode changes.

Makefile.common-jst
```diff
@@ -76,9 +89,29 @@ runtime-stdlib: boot-compiler
 # to a separate directory.
 	$(cpl) -R _build/install/runtime_stdlib _build/runtime_stdlib_install
 # Dune does not believe the compiler can make .cmxs unless the following file exists.
-	@touch _build/runtime_stdlib_install/lib/ocaml_runtime_stdlib/dynlink.cmxa
+# Create [dynlink.cmxa] for 4.14 and [dynlink/dynlink.cmxa] for 5.0
+	mkdir -p _build/runtime_stdlib_install/lib/ocaml_runtime_stdlib/dynlink
+	@touch _build/runtime_stdlib_install/lib/ocaml_runtime_stdlib/{,dynlink/}dynlink.cmxa
```
mshinwell, line 94: Should this change be in a separate PR?

Makefile.common-jst
```diff
@@ -76,9 +89,29 @@ runtime-stdlib: boot-compiler
...
+compiler_targets = \
+  $(ocamlc) \
+  $(ocamlopt) \
+  $(boot_ocamlmklib) \
+  $(boot_ocamldep) \
+  $(boot_ocamlobjinfo)
 
 compiler: runtime-stdlib
+# Build the same target as in [boot-compiler], but with our runtime stdlib.
+# Uses the [main] context to avoid duplicating work.
```
mshinwell, line 105: I think this "Uses the..." comment is in the wrong place.

Makefile.common-jst
```diff
@@ -76,9 +89,29 @@ runtime-stdlib: boot-compiler
...
+	RUNTIME_DIR=$(RUNTIME_DIR) $(dune) build $(ws_compiler) $(compiler_targets)
+
+main: _build/compiler_install compiler
+# Build the rest of the installation using the compiler from the [main] context.
+# The [boot-compiler] is not sufficient, since it marshals bytecode constants
+# incorrectly, using the upstream stdlib. This miscompiles code using new constants
+# like null pointers or float32 values.
+# However, otherwise [boot-compiler] produces identical artifacts to [compiler],
+# and it's safe-ish to reuse the [main] context again.
```
mshinwell, line 114: We need a better argument than "safe-ish".  Which artifacts are required from the context in which the final compiler is built in order to build the otherlibs (minus `unix`)?  I'm wondering if we can use a separate context for this last part.

mshinwell (changes requested):

Makefile.common-jst
dkalinichenko-js: Previously, Dune assumed the 4.14 directory structure for the `boot-compiler`. Now, it also assumes the 5.0 directory structure for the `compiler` we use to build otherlibs, so we need this change to build everything.

Makefile.common-jst
dkalinichenko-js: Moved the entire `dune_compiler_context` to its own directory. This means we don't reuse any artifacts, potentially slowing down build times but ensuring we do not miscompile anything. I believe, since `main` was already cleared between builds in the CI, this change should not introduce any extra delay there.

PR #2941, opened 2024-08-19
Fix domain.c bad merges

mshinwell:
As subject.

### DIFF
--- a/ocaml/runtime/domain.c
+++ b/ocaml/runtime/domain.c
@@ -625,31 +625,15 @@ static void domain_create(uintnat initial_minor_heap_wsize,
   caml_state = domain_state;
 
   domain_state->young_limit = 0;
+
+  domain_state->id = d->id;
+  domain_state->unique_id = s->unique_id;
+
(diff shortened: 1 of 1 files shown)


ocaml/runtime/domain.c
```diff
@@ -770,6 +754,7 @@ static void domain_create(uintnat initial_minor_heap_wsize,
   domain_state->trap_barrier_block = -1;
 #endif
 
+  caml_reset_young_limit(domain_state);
```
stedolan, line 757: This line should not be there. It was removed by f4098874dfd045eb53634f01f3763f3b815ed3aa ("Minor changes to complete statmemprof"), which is not in 5.2 but was backported as part of #2461

stedolan (changes requested):

ocaml/runtime/domain.c
mshinwell: Fixed

stedolan (approved):

PR #2942, opened 2024-08-19
Runtime cherry picks: upstream trunk -> 5.2

mshinwell:
These correspond to the list here:
https://github.com/ocaml-flambda/flambda-backend/wiki/Post%E2%80%905.2-changesets-from-trunk-to-backport-to-5.2.0minus

I think the only significant change to any of these is the flambda2 support for `Sys.poll_actions`, but it is straightforward.

This PR is based on #2941.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -8,6 +8,7 @@
                 "smmintrin.h": "c",
                 "tmmintrin.h": "c",
                 "pmmintrin.h": "c",
-                "*.tbl": "c"
+                "*.tbl": "c",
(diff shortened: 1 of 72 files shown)


stedolan:
Could you explain exactly which PRs / patches are in this? It doesn't seem to match the list in the linked wiki page.

For instance, 13299 is listed in "To backport" but not present here, and 12839 is listed in "To skip" and is present here (at least partially).

mshinwell:
(have discussed with @stedolan separately)

PR13063 was added because it's a dependency of some of the others.

PR #2943, opened 2024-08-19
Require let rec pattern variables to have layout value.

ccasin:
Per discussion on the [unboxed tuples PR](https://github.com/ocaml-flambda/flambda-backend/pull/2879), this requires all pattern variables bound by a `let rec` to have layout value.

This isn't really a change: the existence of `Lambda.layout_letrec` shows we thought things bound by a let-rec had layout value.  But we have been relying on `Value_rec_check` to enforce it, and that will no longer be sufficient in the presence of unboxed tuples (there is a test case in that PR showing this - I will rebase it once this is merged).

@liam923 perhaps you could review?  It's mostly changes to error messages.  This is lower priority than the 5.2 merge or current compiler release.


liam923 (approved):
LGTM. These error messages make a lot more sense than the old ones.

PR #2946, opened 2024-08-20
Run compiler build for profiling in temporary directory

milan-tom:
If the compiler build fails during profiling or the script is terminated during the build, this leaves the build in the repository being developed on in a bad state. This PR fixes this problem by cloning the repository to a temporary directory (another directory within the parent of the directory that the local repository is stored in i.e. a sibling) and then running the build in that temporary directory leaving the original unchanged (aside from changes in the profile file dump directory and the summary CSV being added).

### DIFF
--- a/scripts/profile-compiler-build.sh
+++ b/scripts/profile-compiler-build.sh
@@ -39,7 +39,11 @@ build_compiler() {
   make install
 }
 
-temp_dir=$(mktemp -d -p "$root"/.. -t tmp.XXXXXXX)
+if [ -v TMPDIR ] && [ ! -z "$TMPDIR" ]; then
+  temp_dir=$(mktemp -d)
+else
+  temp_dir=$(mktemp -d -p "$root"/.. -t tmp.XXXXXXX)
+fi
 trap "rm -rf "$temp_dir"" EXIT
 cp -r --reflink=auto "$root" "$temp_dir"
 
--- a/scripts/profile-compiler-build.sh
+++ b/scripts/profile-compiler-build.sh
@@ -5,9 +5,9 @@ set -e -u -o pipefail
 # Works regardless of where script run from
 scripts_dir=$( cd "$( dirname "$0" )" &> /dev/null && pwd )
 
-root=$scripts_dir/..
-dump_dir="$root/_profile"
-summary_path="$root/summary.csv"
+root="$scripts_dir"/..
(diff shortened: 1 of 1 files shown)


scripts/profile-compiler-build.sh
```diff
@@ -1,11 +1,13 @@
 #!/bin/sh
 
-# To be run from the root of the Flambda backend repo
-
 set -e -u -o pipefail
 
-dump_dir="`pwd`/_profile"
-summary_path="`pwd`/summary.csv"
+# Works regardless of where script run from
+scripts_dir=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
```
xclerc, line 6:
```suggestion
scripts_dir=$( cd "$( dirname "$0" )" &> /dev/null && pwd )
```
Is that version enough to achieve what we want?
(I am just worried the current version is less
compatible with various shells.)

scripts/profile-compiler-build.sh
```diff
@@ -1,11 +1,13 @@
...
+
+root=$scripts_dir/..
```
xclerc, line 8:
```suggestion
root="$scripts_dir/.."
```

scripts/profile-compiler-build.sh
```diff
@@ -37,7 +39,14 @@ build_compiler() {
   make install
 }
 
+temp_dir=$(mktemp -d -p $root/.. -t tmp.XXXXXXX)
```
xclerc, line 42: Not sure about the `-p $root/..` parameter;
is relying on `TMPDIR` a problem in practice?

xclerc (commented):
Looks good; only a couple of minor
remarks/questions. I think it is a good
property to not longer depend on
the directory from which the script
is launched.

scripts/profile-compiler-build.sh
milan-tom: Yes, this is sufficient.

scripts/profile-compiler-build.sh
milan-tom: As discussed offline, `TMPDIR` is not always set and may have space issues. So, the code has been changed to make the temp directory in `TMPDIR` if it is set and as a sibling of the repository directory if it is not set.

xclerc (approved):

PR #2949, opened 2024-08-20
Migrate `include_functor` from `jane_syntax` to `parsetree`

riaqn:
This PR migrates the `include_functor` extension from `jane_syntax` to `parsetree`.

**_Please do not merge until I finish the internal migration._**

### DIFF
--- a/ocaml/parsing/parsetree.mli
+++ b/ocaml/parsing/parsetree.mli
@@ -51,6 +51,8 @@ type modalities = modality loc list
 type mode = | Mode of string [@@unboxed]
 type modes = mode loc list
 
+type include_kind = Structure | Functor
+
 (** {1 Extension points} *)
 
 type attribute = {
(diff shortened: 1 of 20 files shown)


ocaml/parsing/parsetree.mli
```diff
@@ -1026,6 +1028,7 @@ and open_declaration = module_expr open_infos
 
 and 'a include_infos =
     {
+     pincl_kind : include_kind;
```
ccasin, line 1031: I think in a perfect world we would probably make `Psig_include_functor` and `Pstr_include_functor` their own things, and then do similarly in typed tree.  That would make it harder to write bugs that conflate `include` with `include functor`.  But considering typed tree already does it this way (my fault) this is fine, just an idle remark.

ccasin (approved):
LGTM

PR #2950, opened 2024-08-20
Make `lib-systhreads/testfork.ml` deterministic

riaqn:
The test has become quite flaky on MacOS.

### DIFF
--- a/ocaml/testsuite/tests/lib-systhreads/testfork.ml
+++ b/ocaml/testsuite/tests/lib-systhreads/testfork.ml
@@ -1,4 +1,5 @@
 (* TEST
+ not-macos;
  include systhreads;
  hassysthreads;
  not-bsd;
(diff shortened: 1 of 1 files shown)


stedolan:
In what way is it flaky, and why is the flakiness limited to MacOS?

From the code, it looks like the messages race between the parent and child. I think it could be made non-racy by:
  - removing the `In parent...` line
  - replacing the `Thread.delay 0.4` line in the parent with `Unix.waitpid` so that it definitely occurs after the child is done

riaqn:
good point - I just assumed it was some intractble concurrency bugs; but it's probably just macOS being slow!

@stedolan could you review the new commit?

xclerc (approved):
I would just change the title of the
pull request, as it will end up being
the commit message when you merge.

PR #2957, opened 2024-08-20
Make `lexing_position` `immutable_data`

liam923:
This PR improves the kind of `lexing_position` to be `immutable_data` since it is a record holding immutable `string`s and `int`s.


riaqn (approved):
LGTM

PR #2961, opened 2024-08-22
Fix soundness bug allowing semi-arbitrary (non-`Tconstr`) manifests for datatypes

jbachurski:
```
# type t = int * int = { foo : string }
type t = int * int = { foo : string; } 
# let magic (t : t) = (t : int * int) ;;
val magic : t -> int * int = <fun> 
# magic { foo = "abc" } ;;
(undefined behaviour in utop - depending on the conversion, either fails an assert here or segfaults)
```

The bug allows the introduction of arbitrary non-Tconstr manifests for datatypes (so the case that used to check normally like `('a, 'b) t = ('a, 'b) s = ...` is still checked correctly), implementing Obj.magic

Seems to have been introduced by #2749

### DIFF
--- a/ocaml/testsuite/tests/typing-misc/datatype_manifests.ml
+++ b/ocaml/testsuite/tests/typing-misc/datatype_manifests.ml
@@ -1,0 +1,13 @@
+(* TEST
+ expect;
+*)
+
+type t = int * int = { foo : string }
+[%%expect{|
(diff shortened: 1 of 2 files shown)


riaqn (approved):
LGTM

lpw25 (approved):
Obviously the correct fix

PR #2966, opened 2024-08-22
Unboxed vec128 string/bytes/bigstring/array load/store

TheNumbat:
(Merges into #2965)

Adds load/store primitives for unboxed simd vectors.

The string/bytes/bigstring primitives are a straightforward extension of the existing infrastructure.
However, the array primitives have been duplicated, since they don't fit the same pattern. If we want versions of these that are indexed by the various unboxed ints, we should similarly factor it out.

The tests for array indexing have been replicated for unboxed vectors in `tests/simd/arrays_u.ml`.


ocaml/lambda/lambda.ml
```diff
@@ -1946,16 +1946,28 @@ let primitive_result_layout (p : primitive) =
...
-    layout_boxed_vector Pvec128
-  | Punboxed_int32_array_load_128 _ ->
-    layout_boxed_vector Pvec128
+  | Pfloatarray_load_128 { boxed = true; _ }
+  | Pfloat_array_load_128 { boxed = true; _ }
+  | Punboxed_float_array_load_128 { boxed = true; _ }
+  | Punboxed_float32_array_load_128 { boxed = true; _ }
+  | Pint_array_load_128 { boxed = true; _ }
+  | Punboxed_int64_array_load_128 { boxed = true; _ }
+  (* 128-bit types are only supported in the x86_64 backend, so we may
```
mshinwell, line 1970: I just noticed this existing comment should be updated; we'll support these for arm64 too.

TheNumbat: I just deleted the comment, since it's not directly relevant here and the condition is checked at `lambda_to_flambda_primitives:1807`

dvulakh (approved):
LGTM

PR #2967, opened 2024-08-23
Improve the modes of legacy `lazy`.

riaqn:
This PR improves the modes of the existing `lazy` construction, without changing its runtime behaviour.

# Some analysis

- Upon cosntruction, the comonadic fragment of the lazy must be higher/weaker than both the thunk and the result.

- When projecting the result, the result will have the same mode as the lazy. Therefore, upon construction we should require the lazy to be higher/weaker than the result.

- The regionality axis is special: I'm not familiar with how the thunk is evaluated and the result allocated, so I conservatively require both to be `global`. That also means a global modality upon projection for better ergonomics.

- The lazy itself might be safe to be stack-allocated, but that involves middle-end changes which I leave to future work. For now, the lazy itself is also heap-allocated.

- Note that the existing legacy behaviour of lazy is a special case of the above.

# Review

Please review by commits.

### DIFF
--- a/ocaml/testsuite/tests/typing-modes/lazy.ml
+++ b/ocaml/testsuite/tests/typing-modes/lazy.ml
@@ -8,7 +8,8 @@ let use_portable : 'a @ portable -> unit = fun _ -> ()
 val use_portable : 'a @ portable -> unit = <fun>
 |}]
 
-(* The thunk and the result are required to be global *)
+(* The thunk and the result are required to be global. This is only because we
+don't support allocating lazy values on the stack. *)
 let foo () =
     lazy (let x @ local = "hello" in x)
 [%%expect{|
--- a/ocaml/testsuite/tests/typing-modes/lazy.ml
+++ b/ocaml/testsuite/tests/typing-modes/lazy.ml
@@ -2,24 +2,102 @@
     expect;
 *)
 
-(* lazy expression is legacy *)
-let u =
-    let _x @ portable = lazy "hello" in
-    ()
+let use_portable : 'a @ portable -> unit = fun _ -> ()
 [%%expect{|
-Line 2, characters 24-36:
-2 |     let _x @ portable = lazy "hello" in
-                            ^^^^^^^^^^^^
+val use_portable : 'a @ portable -> unit = <fun>
+|}]
+
+(* The thunk and the result are required to be global *)
+let foo () =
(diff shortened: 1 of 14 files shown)


riaqn:
Actually hold off a bit - forcing a lazy should require it to be `uncontended`. I will push a commit fix that.

riaqn:
> I think that if you make forcing require contended then you can't actually benefit from a lazy being portable

Good point. 

> It's not needed either because the implementation uses atomic accesses to allow threads to safely race on forcing.

Not sure I understand - the link above says otherwise: "Lazy.force is not concurrency-safe. If you use this module with multiple fibers, systhreads or domains, then you will need to add some locks."

lpw25:
Looks like they adjusted the semantics as part of merging multicore into upstream. This appears to be the relevant discussion https://github.com/ocaml-multicore/ocaml-multicore/issues/750. We should give that a read and decide what we want to do.

anfelor:
As thunks are now guaranteed to be evaluated just once, we can store unique things in them while letting the thunk itself be many. This enables the following design pattern: You can create a computation that uses unique data to create an aliased result and store it in a thunk. Then, whenever the result is actually needed, anyone can force the thunk to get the result of the computation and persist it to all aliases.

ocaml/testsuite/tests/typing-modes/lazy.ml
```diff
@@ -1,25 +1,134 @@
...
+|}]
+
+let foo (local_ x) =
+    lazy (let _ = x in ())
+[%%expect{|
+Line 2, characters 18-19:
+2 |     lazy (let _ = x in ())
+                      ^
+Error: The value x is local, so cannot be used inside a lazy expression.
+|}]
```
lpw25, line 29: Maybe leave a comment saying that the thunk only needs to be global because we don't currently support allocating local values on the stack.

lpw25 (approved):
Looks good. A few small comments.

ocaml/testsuite/tests/typing-modes/lazy.ml
riaqn: Hmm I'm not exactly sure what you mean by that. Do you mean "we don't currently support allocating lazy values on the stack"?

lpw25: Yeah, that's what I meant to write

PR #2968, opened 2024-08-23
Relax modes for `probe`

riaqn:
A prob `[%probe "hello" (a + b + c)]` is compiled (in `translcore.ml`) to:
```
let probe_hello a b c = a + b + c in
probe_hello a b c
```
Closure locks are not needed, because no closing-over is happening. Shared locks are not needed, because nothing is being run more than once. (Surely, this code could be in a loop or something, in which case the loop will add the shared lock as needed).

This PR removes the closure locks/shared locks constraints.

### DIFF
--- a/ocaml/lambda/lambda.mli
+++ b/ocaml/lambda/lambda.mli
@@ -696,7 +696,6 @@ type lambda =
   (* [Lexclave] closes the newest region opened.
      Note that [Lexclave] nesting is currently unsupported. *)
   | Lexclave of lambda
-  (** Exit the current region and proceed to the subexpression. *)
 
 and rec_binding = {
   id : Ident.t;
@@ -715,13 +714,8 @@ and lfunction = private
     loc : scoped_location;
     mode : alloc_mode;     (* alloc mode of the closure itself *)
     ret_mode: alloc_mode;
-    (** The mode of the return value. If [local], the function might allocate in
-        caller's region. *)
-    region : bool;
-    (** Does the function have a region? Even if this is [true], the function
-        can still allocate in caller's region via [Lexclave].
-
-        CR zqian: this field is always true and should be removed. *)
+    region : bool;         (* false if this function may locally
+                              allocate in the caller's region *)
   }
 
 and lambda_while =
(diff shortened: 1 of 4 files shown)


ocaml/lambda/translcore.ml
```diff
@@ -1095,7 +1095,7 @@ and transl_exp0 ~in_new_scope ~scopes sort e =
           (* Might be a module, which are all values.  Otherwise raise. *)
           ignore (Env.find_module_lazy path e.exp_env)
       ) arg_idents;
-      let body = Lambda.rename map lam in
+      let body = Lexclave (Lambda.rename map lam) in
```
lpw25, line 1098: It's probably slightly easier to just remove the call to to `maybe_region_layout` below, since that will just remove the `Lexclave`. I'd probably set `region` to `false` rather than `true` as well, although I don't think it should matter.

ocaml/lambda/lambda.mli
```diff
@@ -714,8 +715,13 @@ and lfunction = private
...
     ret_mode: alloc_mode;
-    region : bool;         (* false if this function may locally
-                              allocate in the caller's region *)
+    (** The mode of the return value. If [local], the function might allocate in
+        caller's region. *)
+    region : bool;
+    (** Does the function have a region? Even if this is [true], the function
+        can still allocate in caller's region via [Lexclave].
+
+        CR zqian: this field is always true and should be removed. *)
```
lpw25, line 724: I think your CR comment is not actually accurate: there are some code paths that set it to `false` and I believe they are still reachable. It should still be removed though.

lpw25 (approved):
LGTM

ocaml/lambda/translcore.ml
riaqn: I talked to @mshinwell and turns out my understanding of `region` is wrong. It's outdated anyway and is removed in #2985 which I will review.

ocaml/lambda/lambda.mli
riaqn: My understanding of `region` is wrong. I will move those comments to #2985 instead.

PR #2970, opened 2024-08-26
Convert to simd

ITO444:
replace instructions in the computation tree with simd instructions


backend/amd64/simd_selection.ml
```diff
@@ -463,3 +463,347 @@ let () =
...
+  let create_const_vec consts =
+    let rec get_list list n =
+      if n = 0
+      then [], list
+      else
+        match list with
+        | [] -> assert false
+        | hd :: tl ->
+          let list1, list2 = get_list tl (n - 1) in
+          hd :: list1, list2
```
xclerc, line 516: (Isn't it `Misc.List.split_at`?)

backend/amd64/simd_selection.ml
```diff
@@ -463,3 +463,347 @@ let () =
...
+
+type vectorized_instruction =
+  { operation : Cfg.operation;
+    arguments : register array;
+    results : register array
+  }
+
+let vector_width = 128
+
+let vectorize_operation width (cfg_ops : Cfg.operation list) :
```
xclerc, line 487: (I think this function is too long
to be easy to understand, but
maybe it is enough to simply
move some nested functions.)

backend/amd64/simd_selection.ml
```diff
@@ -463,3 +463,347 @@ let () =
...
+    | Iadd -> Option.bind add_op (make_default 2 1)
+    | Isub ->
+      let sse_op =
+        match width_type with
+        | W64 -> Sub_i64
+        | W32 -> Sub_i32
+        | W16 -> Sub_i16
+        | W8 -> Sub_i8
+      in
+      Cfg.Specific (Isimd (SSE2 sse_op)) |> make_default 2 1
```
xclerc, line 557: Why is this case different from
the ones for `Iadd` / `Imul`?

xclerc: (asking because if we can make
the cases look similar, it will make
review an maintenance easier.)

xclerc (commented):
(Only a fairly superficial first pass.)

backend/amd64/simd_selection.ml
ITO444: oh, add and mul are reused in lea but not the rest of the functions, should I still pull out all the other instructions like add and mul?

backend/cfg/vectorize.ml
```diff
@@ -178,85 +511,1048 @@ end = struct
...
+    match cell_option with
+    | None -> ()
+    | Some cell ->
+      let instructon = Instruction.Basic (DLL.value cell) in
+      let next_cell = DLL.next cell in
+      if Instruction.Id.Set.exists
+           (Instruction.Id.equal (Instruction.id instructon))
+           good_instructions
+      then DLL.delete_curr cell;
+      remove_scalar_instructions next_cell
```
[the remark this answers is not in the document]
ITO444: done

backend/amd64/simd_selection.ml
```diff
@@ -463,3 +463,347 @@ let () =
...
+  | W8
+  | W16
+  | W32
+  | W64
+
+type register =
+  | New of int
+  | Argument of int
+  | Result of int
+  | Original of int
```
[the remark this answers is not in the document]
ITO444: added comments

backend/amd64/simd_selection.ml
ITO444: what nested functions do you suggest to move out?

backend/amd64/simd_selection.ml
ITO444: I can't seem to find that

ITO444: you mean Misc.split_at?

ITO444: I seem to not be able to call that either

ITO444: ah it's Misc.Stdlib.List.split_at

gretay-js:
superseded by https://github.com/ocaml-flambda/flambda-backend/pull/3365

PR #2973, opened 2024-08-27
Second merge from main

mshinwell:
This is like #2971, but closer to current `HEAD` of `main`; again it matches a merge point on `5.2-runtime-wip-main`.

Based on #2972.

PR #2974, opened 2024-08-27
Backport PR12991

mshinwell:
See https://github.com/mshinwell/flambda-backend/pull/36 .
@ccasin please review (but don't merge).  You only need to look at `e1ef46e893e8bc95291a1adb4c25b54095a61eb8`.
This is based on #2973.

### DIFF
--- a/ocaml/parsing/location.ml
+++ b/ocaml/parsing/location.ml
@@ -656,6 +656,24 @@ let lines_around
   loop ();
   List.rev !lines
 
+(* Get lines from a file *)
+let lines_around_from_file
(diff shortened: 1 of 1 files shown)


ccasin (approved):
I am approving this because I think it does no harm and will unblock us, but I'm skeptical that it is needed.  The upstream PR this backports was closed, not merged, and a fix was made in ppxlib instead - probably we should do the same?

We should discuss with @ncik-roberts next week to see if there's some reason he wanted to do it this way, but this is fine to merge in the mean time if it lets us make progress - we can always back it out later.

PR #2587, opened 2024-05-15
Skip closure non-scannable environment in oldify_mopup

TheNumbat:

PR #2983, opened 2024-08-27
Fix allocation of mixed block closures 

TheNumbat:
Currently, flambda2 allocates closures containing unboxed value slots using `Cmm.make_alloc`. This is problematic in two ways:
- Unboxed vec128 slots are assumed to only take up one word. This can cause the block header length to be shorter than the closinfo startenv index. Due to #2587, this can lead to memory corruption upon promotion.
- Unboxed numbers are written to closures directly allocated on the major heap using `caml_initialize`. If the unboxed numbers look like minor heap pointers, the gc will attempt to promote them. This can also lead to memory corruption.

Additionally, statically allocated closures containing dynamic naked vec128s would not reserve sufficient space for the 128-bit write.

Note these issues are unlikely to arise in practice since we don't yet expose unboxed vec128 types, and closures are only allocated directly on the major heap if they are larger than `max_young_wosize`.

This pr makes the following changes:

flambda2:
- Static value slots now properly use their size in words
- Dynamic closure allocation computes a mixed closure shape based on the slot kinds. Additional size and kind checks have been added here. 
- In the implementation of `Slot_offsets`, the more specific `Value_slot` constructor has been renamed `Scannable_value_slot`

cmm:
- `make_alloc` is now only used for unmixed blocks
- `make_mixed_closure` (analogus to `make_mixed_block`) is now used to allocate closures containing an unboxed prefix

### DIFF
--- a/backend/cmm_helpers.ml
+++ b/backend/cmm_helpers.ml
@@ -1375,6 +1375,18 @@ let setfield_unboxed_float32 arr ofs newval dbg =
          [array_indexing log2_size_addr arr ofs dbg; newval],
          dbg ))
 
+let setfield_unboxed_vec128 arr ofs newval dbg =
+  (* CR layouts v5.1: Properly support big-endian. *)
+  if Arch.big_endian
+  then
+    Misc.fatal_error
+      "Unboxed vec128 fields only supported on little-endian architectures";
+  return_unit dbg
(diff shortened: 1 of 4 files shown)


Gbury:
See #2997 for what is, I think, a better fix.

PR #2985, opened 2024-08-28
Clean up `ret_mode` and `region` in `Lambda.lfunction`

mshinwell:
...and consequentially `contains_no_escaping_local_allocs` in Flambda 2, which was unused.

`Simplif` seemed to be the only place relying on this field.  @riaqn suggested that `ret_mode` could be used instead, which seems like it should be correct, but this part of the diff needs checking carefully.  The rest is straightforward.

### DIFF
--- a/lambda/lambda.ml
+++ b/lambda/lambda.ml
@@ -851,7 +851,7 @@ and lfunction =
     loc: scoped_location;
     mode: locality_mode;
     ret_mode: locality_mode;
-    region: bool; }
+  }
 
 and lambda_while =
   { wh_cond : lambda;
@@ -938,7 +938,7 @@ let max_arity () =
   (* 126 = 127 (the maximal number of parameters supported in C--)
            - 1 (the hidden parameter containing the environment) *)
 
-let lfunction' ~kind ~params ~return ~body ~attr ~loc ~mode ~ret_mode ~region =
+let lfunction' ~kind ~params ~return ~body ~attr ~loc ~mode ~ret_mode =
   assert (List.length params > 0);
   assert (List.length params <= max_arity ());
   (* A curried function type with n parameters has n arrows. Of these,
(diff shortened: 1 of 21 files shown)


ocaml/lambda/simplif.ml
```diff
@@ -586,13 +586,12 @@ let simplify_lets lam =
...
       end
-  | Lfunction{kind=outer_kind; params; return=outer_return; body = l;
-              attr=attr1; loc; ret_mode; mode; region=outer_region} ->
-      begin match outer_kind, outer_region, simplif l with
-        Curried {nlocal=0},
-        true,
+  | Lfunction{kind; params; return=outer_return; body = l;
+              attr=attr1; loc; ret_mode; mode} ->
+      begin match ret_mode, simplif l with
+        Alloc_heap,
```
riaqn, line 592: This looks unsafe. For example:
```
let f (local_ _) =
  fun b -> ...
```
The new logic would compress this into a single function whose `kind` and `params` are incoherent. The old logic would prevent this because `nlocal=1` for `f`. I think checking for `nlocal=0` is good enough for this match.

Moreover, I thought we introduced n-ary lambda, so that the arity users write corresponds to the arity at runtime? Wouldn't this optimization break that?

ocaml/lambda/lambda.mli
```diff
@@ -592,7 +592,9 @@ type loop_attribute =
 type curried_function_kind = { nlocal: int } [@@unboxed]
 (* [nlocal] determines how many arguments may be partially applied
     before the resulting closure must be locally allocated.
-    See [lfunction] for details *)
+    See the implementation of [lfunction] for details *)
+(* CR mshinwell: this comment should presumably say that nlocal = 0 implies
+   that the function cannot allocate in the caller's region *)
```
riaqn, line 597: I will push commit to clean up the comments

riaqn (approved):
LGTM - thanks!

Please hold off merging - I'd like to push commit to clean up the comments.

ocaml/lambda/simplif.ml
mshinwell: This was a mistake, I've asked @riaqn to fix this and push any other changes he would like.

riaqn: I've pushed commits - @mshinwell can you look at them?

mshinwell:
I'd like @lpw25 to read the changes to `Simplif` (not commit-by-commit, read the whole at once) and the changes to the comments.  Otherwise this is ok.  Thanks.

mshinwell:
I will have to look at the discussion again but I think there were some problems with inadequate precision of locality information that caused trouble in https://github.com/ocaml-flambda/flambda-backend/pull/1271.  Maybe this new idea will solve that.  We should revisit that discussion when work recommences on this PR.

riaqn:
This is reopend as requested by #3502

riaqn (approved):
I looked at the changes again, and they seem fine to me.

@mshinwell could you look at my last commit "fix probe handler empty parameters", then we can merge.

PR #2987, opened 2024-08-28
Fixes for ocaml/bytecomp/

mshinwell:
Minor fixes and conflict resolution.
The `caml_get_bytecode_sections` change relates to the note in #2986 .

### DIFF
--- a/ocaml/bytecomp/bytelink.ml
+++ b/ocaml/bytecomp/bytelink.ml
@@ -238,9 +238,17 @@ let link_compunit output_fun currpos_fun inchan file_name compunit =
   if !Clflags.debug && compunit.cu_debug > 0 then begin
     seek_in inchan compunit.cu_debug;
     let debug_event_list : Instruct.debug_event list =
-      Compression.input_value inchan in
+      (* CR ocaml 5 compressed-marshal:
+      Compression.input_value inchan
+      *)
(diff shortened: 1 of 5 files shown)


ocaml/bytecomp/bytelink.ml
```diff
@@ -238,9 +238,17 @@ let link_compunit output_fun currpos_fun inchan file_name compunit =
   if !Clflags.debug && compunit.cu_debug > 0 then begin
     seek_in inchan compunit.cu_debug;
     let debug_event_list : Instruct.debug_event list =
-      Compression.input_value inchan in
+      (* CR ocaml 5 compressed-marshal:
```
xclerc, line 241: (After seeing that a couple of times,
I now wonder whether we could call
`Compression.input_value` from the
various places, and have its implementation
simply be `Marshal.from_channel inchan`.
That would leave us with only one CR.)

xclerc (approved):

ocaml/bytecomp/bytelink.ml
mshinwell: Yeah, we could have, but I'm not sure it's worth changing at this point.  It should be easy to find the uses by grepping for `compressed-marshal` when we eventually fix this.

PR #2988, opened 2024-08-28
Fix yacc build

mshinwell:
This was the easiest way to get the `yacc` build to work, although another possibility is given in the comment.  The code is verbatim from runtime5.


xclerc (approved):

mshinwell:
I meant to say: this is needed because the yacc code now uses macros that are not defined by our 4.x system compilers.

PR #2989, opened 2024-08-28
Driver changes

mshinwell:
Conflict resolution and fixes for `driver/` and `ocaml/driver/`.

I'm unsure about the changes required to `makedepend.ml`; I took what we have on `flambda-backend/main`, but this may not be correct.  @Forestryks probably needs to help.

### DIFF
--- a/ocaml/driver/makedepend.ml
+++ b/ocaml/driver/makedepend.ml
@@ -138,14 +138,12 @@ let find_module_in_load_path name =
         find_in_array a (pos + 1)
     end in
   let rec find_in_path = function
-    | [] -> raise Not_found
-    | (dir, contents) :: rem ->
-        let mem s = List.mem s names || List.mem s unames in
-        match Array.find_opt mem contents with
-        | Some truename ->
-            if dir = Filename.current_dir_name then truename
-            else Filename.concat dir truename
-        | None -> find_in_path rem in
+    [] -> raise Not_found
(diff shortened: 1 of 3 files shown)


ocaml/driver/makedepend.ml
```diff
@@ -138,14 +138,12 @@ let find_module_in_load_path name =
...
-        | Some truename ->
-            if dir = Filename.current_dir_name then truename
-            else Filename.concat dir truename
-        | None -> find_in_path rem in
+    [] -> raise Not_found
+  | (dir, contents) :: rem ->
+      match find_in_array contents 0 with
+        Some truename ->
+          if dir = "." then truename else Filename.concat dir truename
+      | None -> find_in_path rem in
```
Forestryks, line 146: I don't think I touched this part of code, so I don't have more context on this than you. Just from looking at code it looks fine, the only difference being that we also search [stdlib_unames] in addition to [names] and [unames] and for some reason we use "." instead of Filename.current_dir_name (which is defined to be exactly the same on Unix). So changes look reasonable to me, but again, I don't think I have more context here than anybody else.

mshinwell:
I think I've got the resolution correct now, c.f.:
- the diff upstream from `5.1.1..5.2.0`
- the diff from the `5.1.1` upstream version to the `flambda-backend/main` version.

@xclerc please have a look

xclerc (approved):

PR #2994, opened 2024-08-29
Fixes for dumpobj

mshinwell:
One fix because we don't have compressed marshalling; the remainder to match changes in `bytecomp/`, but all straightforward.


lukemaurer (approved):
LGTM

PR #1324, opened 2023-04-20
Extensions in variants

lthls:
Based on #406.
This PR fixes a hole in variant unboxing, which would work as long as there was at least two non-constant constructors (in which case the row-like structure would contain the extension) but fail in the single non-constant constructor case because row-like would not store the extension.

It allows us to remove all allocations in the following code:
```ocaml
module Seq = struct

  type _ t = State : ('s * ('s -> ('a * 's) option)) -> 'a t

  let rec fold_left f acc (State (s, next)) =
    match next s with
    | None -> acc
    | Some (x, s') -> fold_left f (f acc x) (State (s', next))

  let rec map f (State (s, next)) =
    State (s, fun s -> match next s with
        | None -> None | Some (x, s') -> Some (f x, s'))

  let rec unfold f acc =
    State (acc, f)

end

let square x = x * x

let ints lo hi =
  Seq.unfold (fun i -> if i > hi then None else Some (i, i + 1)) lo

let[@inline] sum s =
  Seq.fold_left (+) 0 s

let foo () =
  sum (Seq.map square (ints 0 11))
```
The resulting code is still not optimal, as there remain a number of intermediate branches in the loop, but these branches would be trivially removed once we introduce match-in-match optimisations.

@Gbury and I will be also trying to find good examples to add to the flexpect tests to verify that we don't introduce regressions in the future.

PR #2996, opened 2024-08-29
Support for or_null in Flambda 2 types

lthls:
This PR adds support in Flambda2 types to represent null values.
The external API isn't modified: all previous existing functions create non-null types, and it is not possible to create null types from the API at the moment.

Things to review carefully:
- The provers code has been very tedious to write. I think it is correct, but I suspect that there are cases where the handling of the null case is not great (i.e. the prover could return a more precise result).
- Some APIs need a bit more thought. The external API needs a way to define types for null values, and the internal API of `Type_grammar` needs to be thought about a bit more (currently I've duplicated the `Head_of_kind_value` module with a `Head_of_kind_value_non_null` version, I think things could be factored out).
- The choice of naming for the record fields is up to debate.

### DIFF
--- a/middle_end/flambda2/types/grammar/type_grammar.ml
+++ b/middle_end/flambda2/types/grammar/type_grammar.ml
@@ -38,6 +38,10 @@ module Block_size = struct
   let inter t1 t2 = Targetint_31_63.min t1 t2
 end
 
+type is_null =
+  | Not_null
+  | Maybe_null
+
 (* The grammar of Flambda types. *)
 type t =
   | Value of head_of_kind_value TD.t
@@ -52,6 +56,15 @@ type t =
   | Region of head_of_kind_region TD.t
 
 and head_of_kind_value =
+  { (* CR vlaviron: This Or_unknown_or_bottom is in part redundant with the one
+       introduced by Type_descr, but we need to be able to express things like
+       "unknown but not null" or "bottom but maybe null" (which is the type for
+       the Null constructor) *)
+    non_null : head_of_kind_value_non_null Or_unknown_or_bottom.t;
+    is_null : is_null
+  }
+
+and head_of_kind_value_non_null =
   | Variant of
       { immediates : t Or_unknown.t;
         blocks : row_like_for_blocks Or_unknown.t;
@@ -108,6 +121,7 @@ and head_of_kind_naked_immediate =
   | Naked_immediates of Targetint_31_63.Set.t
   | Is_int of t
   | Get_tag of t
+  | Is_null of t
 
 and head_of_kind_naked_float32 = Float32.Set.t
 
@@ -264,7 +278,14 @@ let rec free_names0 ~follow_value_slots t =
   | Region ty ->
     type_descr_free_names ~free_names_head:free_names_head_of_kind_region ty
 
-and free_names_head_of_kind_value0 ~follow_value_slots head =
+and free_names_head_of_kind_value0 ~follow_value_slots { non_null; is_null = _ }
+    =
+  match non_null with
+  | Unknown | Bottom -> Name_occurrences.empty
+  | Ok non_null ->
+    free_names_head_of_kind_value_non_null ~follow_value_slots non_null
+
+and free_names_head_of_kind_value_non_null ~follow_value_slots head =
   match head with
   | Variant { blocks; immediates; extensions; is_unique = _ } ->
     Name_occurrences.union_list
@@ -306,7 +327,7 @@ and free_names_head_of_kind_value0 ~follow_value_slots head =
 and free_names_head_of_kind_naked_immediate0 ~follow_value_slots head =
   match head with
   | Naked_immediates _ -> Name_occurrences.empty
-  | Is_int ty | Get_tag ty -> free_names0 ~follow_value_slots ty
+  | Is_int ty | Get_tag ty | Is_null ty -> free_names0 ~follow_value_slots ty
 
 and free_names_head_of_kind_naked_float32 _ = Name_occurrences.empty
 
@@ -550,6 +571,18 @@ let rec apply_renaming t renaming =
       if ty == ty' then t else Region ty'
 
 and apply_renaming_head_of_kind_value head renaming =
+  let { non_null; is_null = _ } = head in
+  match non_null with
+  | Unknown | Bottom -> head
+  | Ok non_null ->
+    let non_null' =
+      apply_renaming_head_of_kind_value_non_null non_null renaming
+    in
+    if non_null == non_null'
+    then head
+    else { non_null = Ok non_null'; is_null = head.is_null }
+
+and apply_renaming_head_of_kind_value_non_null head renaming =
   match head with
   | Variant { blocks; immediates; extensions; is_unique } ->
     let immediates' =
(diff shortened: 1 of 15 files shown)


middle_end/flambda2/types/grammar/type_grammar.ml
```diff
@@ -52,6 +52,11 @@ type t =
   | Region of head_of_kind_region TD.t
 
 and head_of_kind_value =
+  { non_null : head_of_kind_value_non_null;
+    is_null : bool Or_unknown.t (* CR vlaviron: define a dedicated type *)
```
chambart, line 56: It might be the right time to fix this CR

middle_end/flambda2/types/meet_and_join_new.ml
```diff
@@ -565,6 +565,25 @@ and meet_expanded_head0 env (descr1 : ET.descr) (descr2 : ET.descr) :
...
+  let meet_is_null env (is_null1 : bool Or_unknown.t)
+      (is_null2 : bool Or_unknown.t) : _ meet_result =
+    match is_null1, is_null2 with
+    | Unknown, Unknown | Known true, Known true | Known false, Known false ->
+      Ok (Both_inputs, env)
+    | Unknown, Known _ -> Ok (Right_input, env)
+    | Known _, Unknown -> Ok (Left_input, env)
+    | Known true, Known false | Known false, Known true -> Bottom
+  in
+  combine_results2 env ~meet_a:meet_head_of_kind_value_non_null
```
Ekdohibs, line 577: I think this is actually wrong: if both sides are nullable, but the meets in the non-null case return `Bottom`, this will return `Bottom` instead of something that is always null.

middle_end/flambda2/types/meet_and_join_old.ml
```diff
@@ -335,7 +335,21 @@ and meet_expanded_head0 env (descr1 : ET.descr) (descr2 : ET.descr) :
     assert false
 
 and meet_head_of_kind_value env (head1 : TG.head_of_kind_value)
-    (head2 : TG.head_of_kind_value) : _ Or_bottom.t =
+    (head2 : TG.head_of_kind_value) =
+  let<* non_null, env_extension =
```
Ekdohibs, line 339: Likewise, this seems incorrect.

lthls: Should be fixed in the new version

middle_end/flambda2/types/meet_and_join_new.ml
lthls: Should be fixed in the new version

middle_end/flambda2/types/grammar/type_grammar.ml
lthls: Done

lthls:
I've pushed a new version. I've kept the old commit to make it easier to review the changes between the old and new version, but in practice I expect that it's better to review the complete PR (the second commit reverts many of the changes from the first).
I will squash before merging anyway.

Ekdohibs (approved):

middle_end/flambda2/types/meet_and_join_new.ml
```diff
@@ -865,6 +865,15 @@ and meet_head_of_kind_naked_immediate env (t1 : TG.head_of_kind_naked_immediate)
     | false, true ->
       meet_with_shape ~rebuild is_int_ty MTC.any_tagged_immediate is_int_side
   in
+  let is_null_immediate ~is_null_ty ~immediates ~is_null_side =
+    let rebuild = TG.Head_of_kind_naked_immediate.create_is_null in
```
Ekdohibs, line 869: In #1324, there's a test about `I.Set.is_empty immediate`  that was added for `is_int` (and already exists for `get_tag`), shouldn't we add it here as well?

middle_end/flambda2/types/meet_and_join_old.ml
```diff
@@ -610,7 +610,28 @@ and meet_head_of_kind_naked_immediate env (t1 : TG.head_of_kind_naked_immediate)
       TG.Head_of_kind_naked_immediate.create_get_tag ty, env_extension
     | Unknown ->
       Ok (TG.Head_of_kind_naked_immediate.create_get_tag ty, TEE.empty))
-  | (Is_int _ | Get_tag _), (Is_int _ | Get_tag _) ->
+  | Is_null ty, Naked_immediates is_null | Naked_immediates is_null, Is_null ty
+    -> (
+    match I.Set.elements is_null with
+    | [] -> Bottom
+    | [is_int] -> (
```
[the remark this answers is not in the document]
lthls: Good catch, there was also an actual error below where one of the branches used `create_is_int` instead of `create_is_null`.

middle_end/flambda2/types/meet_and_join_new.ml
lthls: It didn't make sense before #1324, but I've rebased and added the check.

middle_end/flambda2/types/meet_and_join_new.ml
```diff
@@ -267,6 +267,120 @@ let combine_results2 env ~meet_a ~meet_b ~left_a ~right_a ~left_b ~right_b
...
+      ( Or_bottom.bind env_a ~f:(fun env ->
+            TE.add_env_extension_strict env when_a ~meet_type:(New meet_type)),
+        Or_bottom.bind env_b ~f:(fun env ->
+            TE.add_env_extension_strict env when_b ~meet_type:(New meet_type)) )
+  in
+  let a_result : _ meet_result =
+    match env_a with
+    | Bottom -> Bottom (New_result ())
+    | Ok env -> meet_a env val_a1 val_a2
+  in
```
Ekdohibs, line 314: Not new but writing it here just in case: if `env_a` ends up equal to `Bottom` because of the extensions (say, because there were extensions in `extensions1`'s `when_a`), we will have `a_result` equal to `Bottom (New_result ())` even if `val_a2` was already bottom, losing the `Right_input` information. We could perhaps improve this by passing additional `is_obvisously_bottom_{a,b}` functions, do you think this would be worth the change or not?

middle_end/flambda2/types/provers.ml
```diff
@@ -376,63 +429,51 @@ let prove_variant_like_generic env t : variant_like_proof generic_proof =
...
-let prove_variant_like env t = as_property (prove_variant_like_generic env t)
+let prove_variant_like env t =
+  gen_value_to_proof prove_variant_like_generic_value env t
 
 type boxed_or_tagged_number =
   | Boxed of
       Alloc_mode.For_types.t * Flambda_kind.Boxable_number.t * Type_grammar.t
   | Tagged_immediate
 
 (* CR pchambart: Remove fragile matchs and reuse this function *)
```
Ekdohibs, line 449: This comment could now be removed I think.

middle_end/flambda2/types/meet_and_join_new.ml
lthls: I expect that when one side is obviously Bottom, there will be no associated extensions. We should probably turn this into an invariant check at some point.

middle_end/flambda2/types/provers.ml
lthls: Done

middle_end/flambda2/types/meet_and_join_new.ml
Ekdohibs: Ah yes, there should be no associated extensions coming from that side, and extensions coming from the other side should not be enough to cause a bottom, because they should have been turned into a bottom in that case (I think that's still possible but should be rare enough for it not to matter).

Ekdohibs (approved):

PR #2997, opened 2024-08-29
Fix for allocation of closures containing unboxed values

Gbury:
The current code for allocating dynamic sets of closure did not properly take into account the fact that some of the env fields can now be unbxoed values. In some cases (particularly when the sets are bigger than Max_young_wosize), these unboxed fields require cmm code that sets the corresponding field, and that store operation need to be annotated with the correct memory_chunk.

This commit adds the necessary code to propagate enough information so that we have access to the memory_chunks of every field when allocating a dynmaic set of closures (static closures do not need that, which is helpful since it would be much harder to produce the memory_chunks in those cases).

At the same time, it reworks the current way these allocations are generated: the old code used a `set` function, which in effect was only there to distinguish on what is essentially the memory_chunk of the value being written and use the appropriate cmm expression to write the value. The code is simplified by instead providing the adequate memory_chunk, from which the set function can be deduced.

This fixes the same issue as #2983 , but in a more reasonable manner. It contains slightly more code partly because it also refactors the code for regular mixed block as well.

### DIFF
--- a/backend/cmm_helpers.ml
+++ b/backend/cmm_helpers.ml
@@ -1618,7 +1618,7 @@ let alloc_generic_set_fn block ofs newval memory_chunk dbg =
     if Arch.big_endian
     then
       Misc.fatal_errorf
-        "Fields with memory_chunk %s are not supported on little-endian \
+        "Fields with memory_chunk %s are not supported on big-endian \
          architecture"
         (Printcmm.chunk memory_chunk);
     generic_case ()
@@ -1652,7 +1652,7 @@ let make_alloc_generic ~block_kind ~mode dbg tag wordsize args
             fill_fields (idx + ofs) el ml )
       | _ ->
         Misc.fatal_errorf
-          "To_cmm_helpers.maake_alloc_generic: mismatched list size between \
+          "To_cmm_helpers.make_alloc_generic: mismatched list size between \
            fields and memory chunks"
     in
     let caml_alloc_func, caml_alloc_args =
--- a/backend/cmm_helpers.ml
+++ b/backend/cmm_helpers.ml
@@ -1653,7 +1653,7 @@ let make_alloc_generic ~block_kind ~mode dbg tag wordsize args
       | _ ->
         Misc.fatal_errorf
           "To_cmm_helpers.maake_alloc_generic: mismatched list size between \
-           fiels and memory chunks"
+           fields and memory chunks"
     in
     let caml_alloc_func, caml_alloc_args =
       match Config.runtime5, block_kind with
--- a/backend/cmm_helpers.ml
+++ b/backend/cmm_helpers.ml
@@ -1613,7 +1613,7 @@ let alloc_generic_set_fn block ofs newval memory_chunk dbg =
     then
       Misc.fatal_errorf
         "Fields with memory_chunk %s are not supported on big-endian \
-         architecture"
+         architectures"
         (Printcmm.chunk memory_chunk);
     generic_case ()
   (* Forbidden cases *)
@@ -1646,7 +1646,7 @@ let make_alloc_generic ~block_kind ~mode dbg tag wordsize args
             fill_fields (idx + ofs) el ml )
       | _ ->
         Misc.fatal_errorf
-          "To_cmm_helpers.make_alloc_generic: mismatched list size between \
+          "To_cmm_helpers.make_alloc_generic: mismatched list sizes between \
            fields and memory chunks"
     in
     let caml_alloc_func, caml_alloc_args =
@@ -1727,7 +1727,7 @@ let make_mixed_alloc ~mode dbg ~tag ~value_prefix_size args args_memory_chunks =
             error "mixed blocks"
           | Thirtytwo_unsigned | Thirtytwo_signed | Single _ | Double
           | Onetwentyeight_unaligned | Onetwentyeight_aligned ->
-            error "value prefix of a mixed block"
+            error "the value prefix of a mixed block"
         else
           (* flat suffix part of the block *)
           match memory_chunk with
@@ -1736,7 +1736,7 @@ let make_mixed_alloc ~mode dbg ~tag ~value_prefix_size args args_memory_chunks =
             ok ()
           | Byte_unsigned | Byte_signed | Sixteen_unsigned | Sixteen_signed ->
             error "mixed blocks"
-          | Word_val -> error "flat suffix of a mixed block")
+          | Word_val -> error "the flat suffix of a mixed block")
       0 args_memory_chunks
   in
   make_alloc_generic
--- a/backend/cmm_helpers.ml
+++ b/backend/cmm_helpers.ml
@@ -1568,35 +1568,38 @@ let call_cached_method obj tag cache pos args args_type result (apos, mode) dbg
 
 (* Allocation *)
 
+let memory_chunk_size_in_words_for_mixed_block = function
+  | (Byte_unsigned | Byte_signed | Sixteen_unsigned | Sixteen_signed) as
+    memory_chunk ->
+    Misc.fatal_errorf
+      "Fields with memory chunk %s are not allowed in mixed blocks"
+      (Printcmm.chunk memory_chunk)
+  | Thirtytwo_unsigned | Thirtytwo_signed ->
+    (* Int32s are currently stored using a whole word *)
+    1
+  | Single _ | Double ->
+    (* Float32s are currently stored using a whole word *)
+    if size_float <> size_addr
+    then
+      Misc.fatal_error
+        "Unable to compile mixed blocks on a platform where a float is not the \
+         same width as a value.";
+    1
+  | Word_int | Word_val -> 1
+  | Onetwentyeight_unaligned | Onetwentyeight_aligned -> 2
+
 let mixed_block_size memory_chunks =
   (* CR layouts 5.1: When we pack int32s/float32s more efficiently, this code
      will need to change. *)
   List.fold_left
     (fun acc memory_chunk ->
-      match memory_chunk with
-      | Byte_unsigned | Byte_signed | Sixteen_unsigned | Sixteen_signed ->
-        Misc.fatal_errorf
-          "Fields with memory chunk %s are not allowed in mixed blocks"
-          (Printcmm.chunk memory_chunk)
-      | Thirtytwo_unsigned | Thirtytwo_signed ->
-        (* Int32s are currently stored using a whole word *)
-        acc + 1
-      | Single _ | Double ->
-        (* Float32s are currently stored using a whole word *)
-        if size_float <> size_addr
-        then
-          Misc.fatal_error
-            "Unable to compile mixed blocks on a platform where a float is not \
-             the same width as a value.";
-        acc + 1
-      | Word_int | Word_val -> acc + 1
-      | Onetwentyeight_unaligned | Onetwentyeight_aligned -> acc + 2)
+      acc + memory_chunk_size_in_words_for_mixed_block memory_chunk)
     0 memory_chunks
 
 let alloc_generic_set_fn block ofs newval memory_chunk dbg =
   let generic_case () =
     let addr = array_indexing log2_size_addr block ofs dbg in
-    Cop (Cstore (memory_chunk, Assignment), [addr; newval], dbg)
+    Cop (Cstore (memory_chunk, Initialization), [addr; newval], dbg)
   in
   match (memory_chunk : Cmm.memory_chunk) with
   | Word_val ->
@@ -1637,10 +1640,14 @@ let make_alloc_generic ~scannable_prefix ~mode dbg tag wordsize args
       match args, memory_chunks with
       | [], [] -> Cvar id
       | e1 :: el, m1 :: ml ->
+        let ofs = memory_chunk_size_in_words_for_mixed_block m1 in
         Csequence
           ( alloc_generic_set_fn (Cvar id) (int_const dbg idx) e1 m1 dbg,
-            fill_fields (idx + 1) el ml )
-      | _ -> assert false (* TODO: fatal error, mismatched list lengths *)
+            fill_fields (idx + ofs) el ml )
+      | _ ->
+        Misc.fatal_errorf
+          "To_cmm_helpers.maake_alloc_generic: mismatched list size between \
+           fiels and memory chunks"
     in
     let caml_alloc_func, caml_alloc_args =
       match Config.runtime5, scannable_prefix with
--- a/backend/cmm_helpers.ml
+++ b/backend/cmm_helpers.ml
@@ -1706,9 +1706,17 @@ let make_mixed_alloc ~mode dbg ~tag ~value_prefix_size args args_memory_chunks =
           ofs + memory_chunk_size_in_words_for_mixed_block memory_chunk
         in
         let error situation =
-          Misc.fatal_errorf "Fields with memory chunk %s are not allowed in %s"
+          Misc.fatal_errorf
+            "Fields with memory chunk %s are not allowed in %s.@\n\
+             value_prefix_size: %d@\n\
+             args: @[<v>%a@]@\n\
+             chunks: @[<v>%a@]@."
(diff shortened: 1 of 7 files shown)


backend/cmm_helpers.ml
```diff
@@ -1568,8 +1568,59 @@ let call_cached_method obj tag cache pos args args_type result (apos, mode) dbg
...
+             the same width as a value.";
+        acc + 1
+      | Word_int | Word_val -> acc + 1
+      | Onetwentyeight_unaligned | Onetwentyeight_aligned -> acc + 2)
+    0 memory_chunks
+
+let alloc_generic_set_fn block ofs newval memory_chunk dbg =
+  let generic_case () =
+    let addr = array_indexing log2_size_addr block ofs dbg in
+    Cop (Cstore (memory_chunk, Assignment), [addr; newval], dbg)
```
TheNumbat, line 1599: Not sure if it matters, but this can use `Initialization`

backend/cmm_helpers.ml
```diff
@@ -1582,12 +1633,14 @@ let make_alloc_generic ?(scannable_prefix = Scan_all) ~mode set_fn dbg tag
...
-      | e1 :: el ->
+    let rec fill_fields idx args memory_chunks =
+      match args, memory_chunks with
+      | [], [] -> Cvar id
+      | e1 :: el, m1 :: ml ->
         Csequence
-          ( set_fn idx (Cvar id) (int_const dbg idx) e1 dbg,
-            fill_fields (idx + 1) el )
+          ( alloc_generic_set_fn (Cvar id) (int_const dbg idx) e1 m1 dbg,
+            fill_fields (idx + 1) el ml )
```
TheNumbat, line 1642: I think this needs to offset by the field length in words, not 1

backend/cmm_helpers.ml
```diff
@@ -1615,71 +1668,28 @@ let make_alloc_generic ?(scannable_prefix = Scan_all) ~mode set_fn dbg tag
...
-  type t =
-    | Tagged_immediate
-    | Naked_float
-    | Naked_float32
-    | Naked_int32
-    | Naked_int64_or_nativeint
-end
+let make_closure_alloc ~mode dbg ~tag args args_memory_chunks =
+  let size = mixed_block_size args_memory_chunks in
+  make_alloc_generic ~scannable_prefix:Scan_all ~mode dbg tag size args
```
TheNumbat, line 1686: It's a bit strange to pass `Scan_all` since the arguments can still have an unboxed prefix. Can we rename it?

Gbury: We certainly can change the name of these constructors, maybe something along the lines of `Regular_block` and `Mixed block` ? (contrary to what one might think, a closure, even one containing unboxed values, is a regular block from that point of view, since they use the startenv trick to avoid scanning the unboxed values).

backend/cmm_helpers.ml
Gbury: That's a good point, I'll fix that. On the other, it's quite worrisome that there are no tests that fail, even with a bug like that (I guess that until we have some vec128s, we only have values that use exactly one word...)

backend/cmm_helpers.ml
TheNumbat: Sounds good

backend/cmm_helpers.ml
TheNumbat: Yeah, it seems difficult to get unboxed vec128s into a closure at the moment since all the operations are on the boxed types (until #2965)

backend/cmm_helpers.ml
```diff
@@ -1582,12 +1633,14 @@ let make_alloc_generic ?(scannable_prefix = Scan_all) ~mode set_fn dbg tag
...
+    let rec fill_fields idx args memory_chunks =
+      match args, memory_chunks with
+      | [], [] -> Cvar id
+      | e1 :: el, m1 :: ml ->
         Csequence
-          ( set_fn idx (Cvar id) (int_const dbg idx) e1 dbg,
-            fill_fields (idx + 1) el )
+          ( alloc_generic_set_fn (Cvar id) (int_const dbg idx) e1 m1 dbg,
+            fill_fields (idx + 1) el ml )
+      | _ -> assert false (* TODO: fatal error, mismatched list lengths *)
```
mshinwell, line 1643: This needs fixing.

middle_end/flambda2/to_cmm/to_cmm_set_of_closures.ml
```diff
@@ -547,11 +588,17 @@ let let_static_set_of_closures0 env res closure_symbols
...
     }
   in
-  let l, free_vars, length, env, res, _effs, updates =
+  let l, memory_chunks, free_vars, length, env, res, _effs, updates =
     Static.fill_layout (Some for_static_sets) decls dbg
       ~startenv:layout.startenv value_slots env res Ece.pure ~prev_updates
       layout.slots
   in
+  (match memory_chunks with
+  | None -> ()
```
mshinwell, line 597: I think we should consider removing this `option` type: it isn't clear to me what extra safety it brings, and it causes a lot of clutter.

mshinwell:
I'd prefer to keep any refactoring of the field setting separate from the actual bug fixes here: we have two important fixes that need to get released soon, and I'm concerned about increasing the risk of problems by changing other things at the same time.  Is it possible to split this PR into two in a reasonable manner?  (The change about `Value_slot` naming as above could also be done separately of course.)

backend/cmm_helpers.ml
mshinwell: I think we should contrive to write a test somehow.  Can we introduce a temporary Lambda primitive (or even the real one)?

backend/cmm_helpers.ml
mshinwell: We should probably change the name of the label too, in that case.

backend/cmm_helpers.ml
TheNumbat: My failing test was in #2969; it indeed works with these changes

middle_end/flambda2/to_cmm/to_cmm_set_of_closures.ml
mshinwell: Can the list not just be empty in this case?

backend/cmm_helpers.ml
Gbury: I've done the renaming.

backend/cmm_helpers.ml
Gbury: Done

backend/cmm_helpers.ml
Gbury: Fixed

backend/cmm_helpers.ml
Gbury: Fixed

middle_end/flambda2/to_cmm/to_cmm_set_of_closures.ml
Gbury: I've removed the option

Gbury:
> It looks like the following from #2983 may be missing from this PR:
> 
> * Additional size and kind checks in flambda2 for dynamic closure allocation
> * In the implementation of `Slot_offsets`, the more specific `Value_slot` constructor has been renamed `Scannable_value_slot`

I've cherry-picked the renaming in `Slot_offsets'.

Concerning the additional size and kind checks, I'm not sure they help that much: to be useful, they would have to inspect the actual kind of the cmm expression and compare that against the size of the memory_chunk given (or that it's compatible with the offsets of the `set_fn` function), and that's not easy to do.

If we want some additional checks on the result of slot_offsets and the returned layouts, these checks should be written and applied in `Slot_offsets` rather than in `to_cmm_set_of_closures`.

backend/cmm_helpers.ml
```diff
@@ -1568,37 +1574,97 @@ let call_cached_method obj tag cache pos args args_type result (apos, mode) dbg
...
+        let ofs = memory_chunk_size_in_words_for_mixed_block m1 in
         Csequence
-          ( set_fn idx (Cvar id) (int_const dbg idx) e1 dbg,
-            fill_fields (idx + 1) el )
+          ( alloc_generic_set_fn (Cvar id) (int_const dbg idx) e1 m1 dbg,
+            fill_fields (idx + ofs) el ml )
+      | _ ->
+        Misc.fatal_errorf
+          "To_cmm_helpers.maake_alloc_generic: mismatched list size between \
+           fiels and memory chunks"
```
TheNumbat, line 1656: typo

Gbury: good catch ! fixed

TheNumbat (approved):
Confirmed the unboxed vec128 tests pass with this

backend/cmm_helpers.mli
```diff
@@ -325,24 +325,25 @@ val make_float_alloc :
...
+    The [memory_chunk] list should give the memory_chunk corresponding to
+    each element from the [expression] list. *)
 val make_mixed_alloc :
   mode:Lambda.alloc_mode ->
   Debuginfo.t ->
   tag:int ->
   value_prefix_size:int ->
-  flat_suffix:Flat_suffix_element.t array ->
   expression list ->
+  memory_chunk list ->
```
mshinwell, line 346: We've lost some static checking here for the kinds of the flat suffix, which seems unfortunate.  Can either the `Flat_suffix_element` type be reinstated, or checks added to ensure the memory chunk kinds are appropriate for the suffix?  The first of these seems like a better interface to me.

backend/cmm_helpers.mli
```diff
@@ -325,24 +325,25 @@ val make_float_alloc :
...
-    | Naked_float32
-    | Naked_int32
-    | Naked_int64_or_nativeint
-end
+(** Allocate a closure block. *)
+val make_closure_alloc :
+  mode:Lambda.alloc_mode ->
+  Debuginfo.t ->
+  tag:int ->
+  expression list ->
```
mshinwell, line 333: There should probably be a comment saying what the expression and memory_chunk parameters actually correspond to, since these presumably cover function slots as well, and we don't normally talk about the memory chunk types of those I don't think?

backend/cmm_helpers.ml
```diff
@@ -1568,37 +1574,97 @@ let call_cached_method obj tag cache pos args args_type result (apos, mode) dbg
...
+    (* Values must go through "caml_initialize" *)
+    addr_array_initialize block ofs newval dbg
+  | Word_int -> generic_case ()
+  (* Generic cases that may differ under big endian archs *)
+  | Single _ | Double | Thirtytwo_unsigned | Thirtytwo_signed
+  | Onetwentyeight_unaligned | Onetwentyeight_aligned ->
+    if Arch.big_endian
+    then
+      Misc.fatal_errorf
+        "Fields with memory_chunk %s are not supported on little-endian \
```
mshinwell, line 1621: This should say big-endian

backend/cmm_helpers.ml
```diff
@@ -1568,37 +1574,97 @@ let call_cached_method obj tag cache pos args args_type result (apos, mode) dbg
...
+      | e1 :: el, m1 :: ml ->
+        let ofs = memory_chunk_size_in_words_for_mixed_block m1 in
         Csequence
-          ( set_fn idx (Cvar id) (int_const dbg idx) e1 dbg,
-            fill_fields (idx + 1) el )
+          ( alloc_generic_set_fn (Cvar id) (int_const dbg idx) e1 m1 dbg,
+            fill_fields (idx + ofs) el ml )
+      | _ ->
+        Misc.fatal_errorf
+          "To_cmm_helpers.maake_alloc_generic: mismatched list size between \
```
mshinwell, line 1655: Typo

middle_end/flambda2/to_cmm/to_cmm_primitive.ml
```diff
@@ -88,27 +88,27 @@ let check_alloc_fields = function
...
+        (fun kind acc ->
+          let chunk =
+            To_cmm_shared.memory_chunk_of_kind (K.With_subkind.anything kind)
+          in
+          chunk :: acc)
+        (K.Mixed_block_shape.field_kinds shape)
+        []
+    in
+    let tag = Tag.Scannable.to_int tag in
+    C.make_mixed_alloc ~mode dbg ~tag ~value_prefix_size args args_memory_chunks
```
mshinwell, line 111: It's not clear to me why this diff is needed to actually solve the problem at hand (see other comment about `Flat_suffix_element` too).

mshinwell (changes requested):

backend/cmm_helpers.mli
lthls: I don't mind a bit more clarification, but I think the meaning is obvious: each expression is stored in the block, so the `memory_chunk` corresponds to which store operation we want to use. It used to be `caml_initialize` everywhere, even for function pointers, but now we can use the proper operation (which is `Word_int` for function pointers, arity fields and infix headers).
Note that it doesn't matter for small blocks (<256 words), since we use `Calloc` in that case, letting `Selection` do the work of computing the correct sizes and machtypes.

backend/cmm_helpers.mli
lthls: Are you sure you don't mean for the prefix ? It would make sense to check that everything in the value prefix is either `Word_val` or `Word_int`, but for the suffix we have very few restrictions, and those restrictions are checked in `alloc_generic_set_fn`.

Gbury: I'll add some checks on the memory_chunks for the mixed block case.

backend/cmm_helpers.mli
Gbury: I'll try and add some documentation.

Originally, I though about having an `(expr * memory_chunk) list` but it would have been more costly in most cases (because most of the time, the list of memory_chunks is not used, and you always need the list of expressions anyway). In fine, the actual problem here is that it's not possible (currently) to have a function that gives the `memory_chunk` (or `machtype`) of a cmm expression, and therefore we have to explicitly thread it everywhere every time we end up needing it. This is an unfortunate consequence of the design of cmm currently.

middle_end/flambda2/to_cmm/to_cmm_primitive.ml
Gbury: This part of the diff actually helps ensure that we do not make any error with indexes.

In the new code, allocating a block requires a list of expression, along with the corresponding `memory_chunks`. This gives **all** the information needed to generate the correct writes, especially it gives a very simple and correct way to compute the offset and size of each writes, all the while being simple to generate (because we have the kinds on the flambda side).

Without that change, there is a lot more code that need to compute offsets at various and different points in the code (with a lot of potential errors): if one only provides the kinds of the flat suffix, we then need to infer the size of all of the things before, and to do that, we need some assumptions which may turn out to change at one point, and we can't easily test when those assumptions stop being true (also, it duplicates the computation of indexes and offsets, which is superfluous and error prone, since one needs to make sure that multiple pieces of code compute indexes that are coherent).

backend/cmm_helpers.mli
mshinwell: I meant for the suffix, in particular to make sure that values are not provided.  We should indeed do the converse for the prefix.

backend/cmm_helpers.mli
mshinwell: In particular I think the comment should explicitly say that these lists include the function slots, rather than just the environment.

backend/cmm_helpers.ml
Gbury: fixed

backend/cmm_helpers.ml
Gbury: fixed

backend/cmm_helpers.mli
Gbury: done

backend/cmm_helpers.mli
Gbury: I added the check

mshinwell (approved):

PR #2999, opened 2024-08-30
Move selection-related utility functions to a dedicated module

xclerc:
(As per title.)


mshinwell:
I'd prefer `Selection_utils` not `Select_utils` but as agreed with @xclerc we can rename this later.

mshinwell (approved):

PR #3002, opened 2024-08-30
Remove `get_single_mode_exn`

riaqn:
#2911 assumes that patterns are checked before expressions, and therefore, if the pattern is `Ppat_tuple`, the expression will be checked with expected type `Ttuple`. Therefore, if the expected type of the expression is not `Ttuple`, then the corresponding pattern must be not `Ppat_tuple`.

#2911 will `fatal_error` if this assumption is not met, for extra sanity.

This assumption is obviously wrong for many reasons. This PR removes that assumption.


ccasin (approved):
LGTM

PR #3003, opened 2024-08-30
Support kind `with` semantics

riaqn:
Draft PR. Only to make it more visable and have discussions.


jbachurski:
@riaqn I think the test changes are because the changes from #2840 in Typetexp are missing (if you’re still looking for the cause)

riaqn:
> @riaqn I think the test changes might be because the changes from #2840 in Typetexp are missing (if you’re still looking for the cause)

Thanks - that's precisely why. I removed this to make the compiler build. Could you look at our internal channel? Thanks!

riaqn:
The PR is now vaguely working, things to do:
- Check that all `~type_equal` is passed correctly. Set it to `assert false` if possible for sanity.
- Currently printing of recursive types is not working properly:
```
type r = R of r list [@@unboxed]
let rec a = R [a];;
[%%expect{|
type r = R of r list [@@unboxed]
val a : r = R [<cycle>]
|}];;
Probably related code: search `<cycle>` in `genprintval.ml`, or `reduce_bound` in `jkind.ml`

- Printing `value mod (ty)` instead of `value mod 'a`. Yeah that's a string literal `(ty)`. This is to help debug the previous point.

riaqn:
More development at https://github.com/mshinwell/flambda-backend/pull/41
Closing this now.

riaqn:
development moved back to this PR. Please check the above for things to do.

riaqn:
Done in other PR.

PR #3004, opened 2024-09-02
"fix" a regression in the 5.2 merge

ccasin:
This adds a test that newly fails on the 5.2 branch, and a fix, in two commits.

The test passes upstream and seems fine, so I think this is a regression in the merge.  But I don't think the fix here (restoring a piece of the previous implementation of checking `Pexp_constraint`s in typecore) is correct.  I'm just posting this so @mshinwell  can use it to make more progress and so I can point to it for discussion.

What is happening is that `loosen_arrow_modes` in `type_argument` is updating the scope of the type, and the fact that the previous implementation was throwing away the old type meant we ignored that.  That seems wrong, instead, I think `type_argument` shouldn't need to use the equation for mode checking in this example, and so this program typechecks in 5.1 for the wrong reason.

Will discuss with modes people.

### DIFF
--- a/ocaml/testsuite/tests/typing-modes/gadts.ml
+++ b/ocaml/testsuite/tests/typing-modes/gadts.ml
@@ -1,0 +1,24 @@
+(* TEST
+ expect;
+*)
+
+(* An example where [type_argument] can refine modes using a gadt equation. *)
+type _ t =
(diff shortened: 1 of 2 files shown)


lpw25 (approved):
LGTM

PR #3006, opened 2024-09-03
Fix specialization of float32 bigarray ref/set

TheNumbat:
Currently, the following program calls the C function `caml_ba_float32_ref_1`:

```
open Stdlib.Bigarray

external get : ('a, float32_elt, 'c) Array1.t -> int -> float32
    = "%caml_ba_float32_ref_1"

let f (x : ('a, float32_elt, c_layout) Array1.t) = get x 0
```

The C call is emitted because the compiler isn't specializing the primitive with `c_layout`. That's because it only attempts to specialize when both the kind and the layout are unknown, but the custom float32 primitive sets the kind to float32_t. This leads to worse unboxing behavior.

This PR adjusts `specialize_primitive` to attempt to specialize `Pbigarrayref` and `Pbigarrayset` regardless of whether the kind and layout are known.

Flambda2 after the fix:
```
 (λ〈k14〉《k13》⟅my_region/27N⟆ ⟅my_ghost_region/26N⟆
      (x/29UV ∷ 𝕍) (my_closure/28N ∷ 𝕍) my_depth/25N .
    (prim/30N = ((Bigarray_length 1 x/29UV) test.ml:6,51--58)
     prim/33N = ((<u #0 prim/30N) test.ml:6,51--58)
     (switch prim/33N test.ml:6,51--58
     | 0 ↦ goto k18
     | 1 ↦ goto k19)
     k19 #One:
     prim/37N =
      (((Bigarray_load (num_dimensions 1) (kind Float32_t) (layout C)) x/29UV
        0) test.ml:6,51--58)
     Pbigarrayref/38N = ((Box_float32[Heap] prim/37N) test.ml:6,51--58)
     return k14 Pbigarrayref/38N
```

### DIFF
--- a/ocaml/typing/typeopt.mli
+++ b/ocaml/typing/typeopt.mli
@@ -32,8 +32,9 @@ val array_kind :
   Typedtree.expression -> Jkind.Sort.t -> Lambda.array_kind
 val array_pattern_kind :
   Typedtree.pattern -> Jkind.Sort.t -> Lambda.array_kind
-val bigarray_type_kind_and_layout :
-      Env.t -> Types.type_expr -> Lambda.bigarray_kind * Lambda.bigarray_layout
+val bigarray_specialize_kind_and_layout :
+  Env.t -> kind:Lambda.bigarray_kind -> layout:Lambda.bigarray_layout ->
+  Types.type_expr -> Lambda.bigarray_kind * Lambda.bigarray_layout
 
 (* CR layouts v7: [layout], [function_return_layout], [function2_return_layout],
    and [layout_of_sort] have had location arguments added just to support the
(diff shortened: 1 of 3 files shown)


ocaml/typing/typeopt.mli
```diff
@@ -32,8 +32,9 @@ val array_kind :
   Typedtree.expression -> Jkind.Sort.t -> Lambda.array_kind
 val array_pattern_kind :
   Typedtree.pattern -> Jkind.Sort.t -> Lambda.array_kind
-val bigarray_type_kind_and_layout :
-      Env.t -> Types.type_expr -> Lambda.bigarray_kind * Lambda.bigarray_layout
+val bigarray_specialize_kind_and_layout :
+  Env.t -> kind:Lambda.bigarray_kind -> layout:Lambda.bigarray_layout ->
+  Types.type_expr -> Lambda.bigarray_kind * Lambda.bigarray_layout
```
ncik-roberts, line 37: I suggest adding a comment that explains how `kind` and `layout` relate to the output kind/layout.

TheNumbat: Added

PR #3011, opened 2024-09-04
Disable stack switching on noalloc C calls

mshinwell:
On Intel hardware, stack switching in runtime5 seems to be causing (what we believe is likely to be) bad speculation in the CPU frontend, causing many more (150% compared to runtime4) L1 icache hits and misses despite only slightly elevated retired instructions.  We are still investigating, but a significant mitigation appears to be to disable stack switching on `noalloc` C calls.  With stack checks disabled, this should work fine -- the C code will continue to execute on the OCaml stack, which has a guard page -- although it remains to check what happens if there is a stack overflow, in terms of the detection code.

This change also means that `%rbx` can be released to the register allocator across such calls.

### DIFF
--- a/backend/amd64/emit.mlp
+++ b/backend/amd64/emit.mlp
@@ -1416,7 +1416,11 @@ let emit_instr ~first ~fallthrough i =
           I.mov (domain_field Domainstate.Domain_young_ptr) r15
         end
       end else begin
-        if Config.runtime5 then begin
+        let runtime5_should_stack_switch =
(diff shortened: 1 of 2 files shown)


stedolan:
This is confusing to me, because I don't yet see how the proposed change (avoiding stack switching) can cause the described effect (changes in bad speculation rate).

Have you tried benchmarking using NOPs with the same size and instruction boundaries as the instructions you removed, just to verify that the effect here isn't being caused by code size / alignment changes?

PR #3015, opened 2024-09-05
Mark `Capsule.Mutex.t` as crossing portability and contention

dkalinichenko-js:
Title. Requires `-allow-illegal-crossing`, since the record has a mutable field `poisoned`, but the resulting API is still thread-safe.


riaqn (approved):
Looks good  - thanks!

PR #3016, opened 2024-09-06
Change profile output to simplify parsing for dynamic counters

milan-tom:
Changes the profile pass name output from something such as:
```
gen.ml/generate/compile_phrases/camlBin_prefix_gen__Gen__mem_17_29_code/regalloc/cfg/peephole_optimize_cfg/block 329
```
to:
```
gen.ml/generate/compile_phrases/function=camlBin_prefix_gen__Gen__mem_17_29_code/regalloc/cfg/peephole_optimize_cfg/block=329
```
This change includes `function=` and `block=` keys to make identification of function names and block labels in this hierarchy of pass names easier for parsers of this profile output.

This PR also exports the sanitisation function for fields used in the CSV profile output (most notably function symbol names) such that it can be re-used by external parsers.

### DIFF
--- a/ocaml/utils/profile.ml
+++ b/ocaml/utils/profile.ml
@@ -389,10 +389,8 @@ let column_mapping = [
   `Counters, "counters"
 ]
 
-let chars_to_sanitise_for_csv = [|','; '/'|]
-
 let sanitise_for_csv =
-  String.map (fun c -> if Array.mem c chars_to_sanitise_for_csv then '_' else c)
+  String.map (fun c -> if Char.equal c ',' then '_' else c)
 
 let output_to_csv ppf columns =
   let to_csv cell_strings =
--- a/ocaml/utils/profile.ml
+++ b/ocaml/utils/profile.ml
@@ -389,9 +389,15 @@ let column_mapping = [
   `Counters, "counters"
 ]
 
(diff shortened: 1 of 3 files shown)


ocaml/utils/profile.ml
```diff
@@ -389,9 +389,15 @@ let column_mapping = [
   `Counters, "counters"
 ]
 
+let chars_to_sanitise_for_csv = [|','; '/'|]
```
xclerc, line 392: Out of curiosity, why does `/` need
to be sanitised for CSV?

xclerc (approved):
(Please add a pull request description,
it makes future searches.)

ocaml/utils/profile.ml
milan-tom: I realise now that the name is slightly misleading. In reality, I want to sanitise a function symbol name such that it can be used in both a CSV field and as a filename (although I don't expect a forward slash in the function symbol name in reality). Since this seems dependent on the implementation of dynamic counters (dumping CSVs for each function with the function symbol name in the filename), I don't need to include this here but rather it can be part of the external implementation.

xclerc: As you prefer, but I was really not
suggesting a change should be
made - just being curious.

PR #3020, opened 2024-09-10
Fix bug in iarray accumulators, and add lots of tests

goldfirere:
The first commit actually fixes the bug that was spotted; the second commit adds lots of omitted tests for all the local functions.


goldfirere:
Just to share a little archeology I performed: These tests should probably have been added in https://github.com/ocaml-flambda/ocaml-jst/pull/156. I reviewed that PR in some detail, but somehow completely missed the lack of tests. I'm pleased to say that in the past year, I've become much more attuned to looking for tests, so I'm pretty confident I would not make this same mistake today.

PR #3022, opened 2024-09-10
Fix backup thread & systhreads interaction

stedolan:
The alternative runtime locking scheme remains incompatible with domains, but now programs can use one or the other and will get a reasonable error message if they try to use both.

### DIFF
--- a/ocaml/otherlibs/systhreads/st_stubs.c
+++ b/ocaml/otherlibs/systhreads/st_stubs.c
@@ -56,6 +56,30 @@
 /* Max computation time before rescheduling, in milliseconds */
 #define Thread_timeout 50
 
+/* Currently, the caml_switch_runtime_locking_scheme mechanism is incompatible
+   with multi-domain programs. To give a reasonable-looking error if both are
+   used at once, we track which (if either) of the features has been used */
+static enum {
+  /* State at startup: neither feature has been used.
+     The domain lock remains held by the first systhread of domain 0,
+     and the st_masterlock is used to provide mutual exclusion of threads.
+     No backup thread has started. */
+  LOCKMODE_STARTUP,
+
+  /* State if caml_switch_runtime_locking_scheme is used.
+     The domain lock remains held by the first systhread of domain 0,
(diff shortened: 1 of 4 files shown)


ocaml/otherlibs/systhreads/st_stubs.c
```diff
@@ -56,6 +56,30 @@
...
+     The domain lock remains held by the first systhread of domain 0,
+     and the st_masterlock is used to provide mutual exclusion of threads.
+     No backup thread has started. */
+  LOCKMODE_STARTUP,
+
+  /* State if caml_switch_runtime_locking_scheme is used.
+     The domain lock remains held by the first systhread of domain 0,
+     and a custom locking scheme is used to provide mutual exclusion of threads.
+     No backup thread has started, nor may one ever start. */
+  LOCKMODE_NEW_SCHEME,
```
lpw25, line 73: Maybe `LOCKMODE_CUSTOM_SCHEME`?

ocaml/otherlibs/systhreads/st_stubs.c
```diff
@@ -128,6 +152,9 @@ st_tlskey caml_thread_key;
 
 #define This_thread ((caml_thread_t) st_tls_get(caml_thread_key))
 
+/* Equals This_thread on the main (first) systhread of each domain */
+st_tlskey caml_thread_is_main_key;
```
lpw25, line 156: Why is this a separate TLS key, rather than going in the `caml_thread_struct`?

ocaml/otherlibs/systhreads/st_stubs.c
```diff
@@ -317,6 +344,8 @@ CAMLexport void caml_switch_runtime_locking_scheme(struct caml_locking_scheme* n
 {
   struct caml_locking_scheme* old;
   int dom_id = Caml_state->id;
+  if (caml_domain_is_multicore())
```
lpw25, line 347: Why this check instead of checking the locking scheme is not `DOMAINS`?

lpw25 (approved):
Looks good. Some comments.

PR #3023, opened 2024-09-10
Support for or_null value kinds in Lambda

lthls:
This PR adds the Lambda structures to represent value kinds with or without null.
It does not (yet) contain the code to forward this to the middle-end (which results in compilation errors in Flambda2), nor the code to type-check or_null types.

### DIFF
--- a/ocaml/lambda/printlambda.ml
+++ b/ocaml/lambda/printlambda.ml
@@ -171,37 +171,59 @@ let tag_and_constructor_shape print_value_kind ppf (tag, shape) =
     (constructor_shape print_value_kind)
     shape
 
-let variant_kind print_value_kind ppf ~consts ~non_consts =
-  fprintf ppf "@[<hov 1>[(consts (%a))@ (non_consts (%a))]@]"
+let variant_kind or_null_suffix print_value_kind ppf ~consts ~non_consts =
+  fprintf ppf "@[<hov 1>[(consts (%a))@ (non_consts (%a))%s]@]"
     (Format.pp_print_list ~pp_sep:Format.pp_print_space Format.pp_print_int)
     consts
     (Format.pp_print_list ~pp_sep:Format.pp_print_space
       (tag_and_constructor_shape print_value_kind))
     non_consts
+    or_null_suffix
 
-let rec value_kind ppf = function
+let value_kind print_value_kind_non_null ppf vk =
+  let or_null_suffix =
+    match vk.nullable with
+    | Non_nullable -> ""
+    | Nullable -> " or_null"
+  in
+  print_value_kind_non_null or_null_suffix ppf vk.raw_kind
+
+let rec value_kind_non_null or_null_suffix ppf = function
   | Pgenval -> ()
-  | Pintval -> fprintf ppf "[int]"
-  | Pboxedfloatval bf -> fprintf ppf "[%s]" (boxed_float_name bf)
-  | Parrayval elt_kind -> fprintf ppf "[%sarray]" (array_kind elt_kind)
-  | Pboxedintval bi -> fprintf ppf "[%s]" (boxed_integer_name bi)
-  | Pboxedvectorval bv -> fprintf ppf "[%s]" (boxed_vector_name bv)
+  | Pintval -> fprintf ppf "[int%s]" or_null_suffix
+  | Pboxedfloatval bf ->
+    fprintf ppf "[%s%s]" (boxed_float_name bf) or_null_suffix
+  | Parrayval elt_kind ->
+    fprintf ppf "[%sarray%s]" (array_kind elt_kind) or_null_suffix
+  | Pboxedintval bi ->
+    fprintf ppf "[%s%s]" (boxed_integer_name bi) or_null_suffix
+  | Pboxedvectorval bv ->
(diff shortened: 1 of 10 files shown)


ocaml/lambda/printlambda.ml
```diff
@@ -170,29 +170,41 @@ let variant_kind print_value_kind ppf ~consts ~non_consts =
       (tag_and_constructor_shape print_value_kind))
     non_consts
 
-let rec value_kind ppf = function
+let value_kind print_value_kind_non_null ppf vk =
+  match vk.nullable with
+  | Non_nullable ->
+    print_value_kind_non_null ppf vk.raw_kind
+  | Nullable ->
+    fprintf ppf "%a or_null" print_value_kind_non_null vk.raw_kind
```
Ekdohibs, line 178: Maybe it would be good to have parentheses here, or the list of fields of `Pvariant` will be confusing.

ocaml/lambda/printlambda.ml
```diff
@@ -208,26 +220,26 @@ let layout ppf layout_ = layout' true ppf layout_
...
      fprintf ppf ": %s%sarray@ " smode (array_kind elt_kind)
-  | Pvalue (Pboxedintval bi) -> fprintf ppf ": %s%s@ " smode (boxed_integer_name bi)
-  | Pvalue (Pboxedvectorval (Pvec128 v)) ->
+  | Pvalue { raw_kind = Pboxedintval bi; _ } -> fprintf ppf ": %s%s@ " smode (boxed_integer_name bi)
+  | Pvalue { raw_kind = (Pboxedvectorval (Pvec128 v)); _ } ->
     fprintf ppf ": %s%s@ " smode (vec128_name v)
-  | Pvalue (Pvariant { consts; non_consts; }) ->
-    variant_kind value_kind' ppf ~consts ~non_consts
+  | Pvalue { raw_kind = (Pvariant { consts; non_consts; }); _ } ->
+    variant_kind (value_kind value_kind_non_null') ppf ~consts ~non_consts
```
Ekdohibs, line 234: It feels wrong to exclude the `or_null` information when printing return kinds.

ocaml/lambda/translprim.ml
```diff
@@ -1059,7 +1059,7 @@ let specialize_primitive env loc ty ~has_constant_constructor prim =
                                   Jkind.Sort.for_block_element typ))
           fields
       in
-      let useful = List.exists (fun knd -> knd <> Pgenval) shape in
+      let useful = List.exists (fun knd -> knd.raw_kind <> Pgenval) shape in
```
Ekdohibs, line 1062: Wouldn't `useful` now be `List.exists (fun knd -> knd <> Lambda.generic_value) shape` instead?

lthls: It's not completely clear that knowing that a `Pgenval` is not `Null` is useful, but your suggestion is also reasonable and avoids having to answer this complicated question.

ocaml/lambda/printlambda.ml
lthls: I've changed the printing code to generate better results (or at least I hope so). Please re-review when you can.

ocaml/lambda/printlambda.ml
lthls: This should be fixed with the new code.

ocaml/lambda/printlambda.ml
```diff
@@ -220,39 +230,54 @@ let layout ppf layout_ = layout' true ppf layout_
...
+  | Pvalue { raw_kind; nullable } -> begin
+    let or_null_suffix =
+      match nullable with
+      | Non_nullable -> ""
+      | Nullable -> " or_null"
+    in
+    match raw_kind with
+    | Pgenval when is_heap_mode mode -> ()
+    | Pgenval -> fprintf ppf ": %s@ " smode
+    | Pintval -> fprintf ppf ": int@ "
```
Ekdohibs, line 242: In all these cases we don't print the `or_null` information, is this intentional?

ocaml/lambda/printlambda.ml
```diff
@@ -162,41 +162,51 @@ let tag_and_constructor_shape print_value_kind ppf (tag, shape) =
...
+  let or_null_suffix =
+    match vk.nullable with
+    | Non_nullable -> ""
+    | Nullable -> " or_null"
+  in
+  print_value_kind_non_null or_null_suffix ppf vk.raw_kind
 
-let rec value_kind_non_null ppf = function
+let rec value_kind_non_null or_null_suffix ppf = function
   | Pgenval -> ()
```
Ekdohibs, line 183: Here and in `value_kind_non_null'` we don't print `or_null_suffix` in the `Pgenval` case.

lthls: The `Pgenval` cases are intentional: I couldn't find a good way to print it, and I was hoping that it wouldn't be very important.

ocaml/lambda/printlambda.ml
lthls: The `int` case is a mistake; I'll fix it.

lthls:
This PR is included in #3112, and will be merged there.

PR #3025, opened 2024-09-10
OCaml 5.2.0 staging (not for merging)

mshinwell:
This is the fully-reviewed `merge-5.2` branch plus:
- some more recent merges from `main`
- a few other fixes that need review (I will ask the relevant people individually to look at these).


mshinwell:
Note to self: `ocaml/configure` needs regenerating with autoconf 2.71.

xclerc (commented):
(only read `1b89051`..`52f5c0d`)

We may want to update `ocaml/tools/magic`,
but maybe leaving it as-is is better for
backward compatibility.

mshinwell:
This now up-to-date with `5.1.1minus-23`.

PR #3028, opened 2024-09-11
Fix bug with update_field and mixed blocks

mshinwell:
Surprisingly, the code in `To_cmm_static.update_field` was not being exercised by the testsuite, and contained an error: it was always using update kind `pointers` even when the corresponding fields formed part of a mixed block flat suffix.  In some cases this is luckily caught by a check in the x86-64 backend that detects illegal moves between float and int registers (it was trying to pass a naked float to `caml_initialize` in the original failing test during initialization of an inconstant statically-allocated mixed block).

The relevant testsuite test has been updated and confirmed to fail on current `main` on x86-64.

### DIFF
--- a/middle_end/flambda2/to_cmm/to_cmm_static.ml
+++ b/middle_end/flambda2/to_cmm/to_cmm_static.ml
@@ -35,22 +35,22 @@ let or_variable f default v cont =
   | Const c -> f c cont
   | Var _ -> f default cont
 
-let update_field symb env res acc i field =
+let update_field symb env res acc i update_kind field =
   Simple.pattern_match'
     (Simple.With_debuginfo.simple field)
     ~var:(fun var ~coercion:_ ->
       (* CR mshinwell/mslater: It would be nice to know if [var] is an
          immediate. *)
       let dbg = Simple.With_debuginfo.dbg field in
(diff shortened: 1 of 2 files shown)


TheNumbat (approved):

mshinwell:
I forgot to add: this bug only exists on runtime5.

middle_end/flambda2/to_cmm/to_cmm_static.ml
```diff
@@ -231,7 +231,31 @@ let static_const0 env res ~updates (bound_static : Bound_static.Pattern.t)
...
+      | Value_only -> List.map (fun _ -> UK.pointers) fields
+      | Mixed_record shape ->
+        let value_prefix =
+          List.init (Flambda_kind.Mixed_block_shape.value_prefix_size shape)
+            (fun _ -> UK.pointers)
+        in
+        let flat_suffix =
+          List.map
+            (fun (flat_suffix_elt : Flambda_kind.flat_suffix_element) ->
+              match flat_suffix_elt with
```
Gbury, line 245: this is at least the second occurrence of this function (the first being in the recently merged #2997 ). This should be factored out somewhere, either in this file, or provided in the `Mixed_block_shape` module.

PR #3029, opened 2024-09-11
5.2 multicore markdelay

stedolan:
This patch extends the existing single-domain-only markdelay patch to work with multiple domains, and (in separate commits) re-enables all the parallel tests and adds a new parallel test.

The idea is to add a new phase, `Phase_sweep_main`, prior to `Phase_sweep_and_mark_main`, in which only sweeping and not marking can be done. Since the roots have not yet been marked, it is safe to give new allocations the `UNMARKED` status during this phase, allowing them to be collected a cycle earlier than otherwise. Also, caml_modify need not do darkening during this phase. This was implemented for single-domain programs in #2348, and this PR extends it to multi-domain programs.

Since domains freely share new allocations, all must agree on phase transitions. So, when any domain finishes sweeping, it requests a transition to Phase_sweep_and_mark_main. Instead of being acted upon directly, this sets a flag to trigger the phase transition at the next minor GC. (The phase transition involves marking roots, which must be done with the minor heap empty, and the simplest way to achieve this is to wait until the next point when the minor heap naturally empties)


mshinwell:
(This was automatically closed by Github when `5.2-merge-staging` was merged.  I've re-opened and changed the base to `main`.)

ocaml/runtime/major_gc.c
```diff
@@ -1368,35 +1373,52 @@ static intnat ephe_sweep (caml_domain_state* domain_state, intnat budget)
...
 
-  /* CR ocaml 5 domains (sdolan):
-     Either this transition needs to be synchronised between domains,
-     or a different write barrier needs to be used while some domains
-     have started marking and others have not. */
-  CAMLassert(caml_domain_alone());
-  caml_gc_phase = Phase_sweep_and_mark_main;
+  caml_domain_state* domain = Caml_state;
+
+  /* Ephemerons */
```
NickBarnes, line 1395: Would be clearer if all this ephemerons stuff was in a separate function. AIUI it's required before we start marking, so maybe `prepare_ephemerons_for_marking()` or something.

ocaml/runtime/major_gc.c
```diff
@@ -1368,35 +1373,52 @@ static intnat ephe_sweep (caml_domain_state* domain_state, intnat budget)
...
 
   CAML_EV_BEGIN(EV_MAJOR_MARK_ROOTS);
-  caml_do_roots (&caml_darken, darken_scanning_flags, domain, domain, 0);
   {
     uintnat work_unstarted = WORK_UNSTARTED;
-    if(atomic_compare_exchange_strong(&domain_global_roots_started,
-                                      &work_unstarted,
-                                      WORK_STARTED)){
-        caml_scan_global_roots(&caml_darken, domain);
+    if (atomic_load_relaxed(&domain_global_roots_started) == WORK_UNSTARTED &&
```
NickBarnes, line 1413: The domain which wins this atomic scans the global roots. Needs a comment.

ocaml/runtime/major_gc.c
```diff
@@ -153,7 +153,8 @@ static atomic_uintnat work_counter;
 
 enum global_roots_status{
   WORK_UNSTARTED,
-  WORK_STARTED
+  WORK_STARTED,
+  WORK_COMPLETE
 };
 static atomic_uintnat domain_global_roots_started;
```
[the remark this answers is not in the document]
NickBarnes: Also I think this can be a static local in the appropriate root-scanning function.

ocaml/runtime/major_gc.c
```diff
@@ -2042,9 +2056,14 @@ int caml_mark_stack_is_empty(void)
 }
 #endif
 
-void caml_empty_mark_stack (void)
+static void empty_mark_stack (void)
 {
   while (!Caml_state->marking_done){
+    while (!caml_marking_started()) {
```
NickBarnes, line 2062: Add comment to justify this annoying-but-necessary while loop.

ocaml/runtime/minor_gc.c
```diff
@@ -830,7 +830,13 @@ int caml_do_opportunistic_major_slice
 /* Make sure the minor heap is empty by performing a minor collection
    if needed.
 */
-void caml_empty_minor_heap_setup(caml_domain_state* domain_unused) {
+void caml_empty_minor_heap_setup(caml_domain_state* domain_unused,
+                                 void* mark_requested) {
+  /* Check whether the mark phase has been requested */
```
NickBarnes, line 835: Moar comment!! Moar!!!

NickBarnes (commented):
This is all OK, but definitely needs some more comments in a few places (and possibly some better names for things), as commented.

ocaml/runtime/major_gc.c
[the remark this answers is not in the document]
stedolan: Done

ocaml/runtime/major_gc.c
```diff
@@ -319,7 +323,7 @@ static void ephe_todo_list_emptied (void)
 
   /* Since the todo list is empty, this domain does not need to participate in
    * further ephemeron cycles. */
-  atomic_fetch_add(&ephe_cycle_info.num_domains_todo, -1);
+  atomic_fetch_add_verify_ge0(&ephe_cycle_info.num_domains_todo, -1);
```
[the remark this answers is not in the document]
stedolan: Yeah, agreed.

ocaml/runtime/major_gc.c
stedolan: Done

ocaml/runtime/major_gc.c
stedolan: Done

ocaml/runtime/minor_gc.c
stedolan: Done!!!1

stedolan:
I rebased to current `main` and added another commit with review feedback. There are no functional changes in the new commit - it's entirely shuffling code around, renaming, and adding comments.

ocaml/runtime/major_gc.c
```diff
@@ -1399,36 +1409,35 @@ void caml_mark_roots_stw (int participant_count, caml_domain_state** barrier_par
...
+
+  begin_ephe_marking();
 
   CAML_EV_BEGIN(EV_MAJOR_MARK_ROOTS);
   {
     uintnat work_unstarted = WORK_UNSTARTED;
-    if (atomic_load_relaxed(&domain_global_roots_started) == WORK_UNSTARTED &&
-        atomic_compare_exchange_strong(&domain_global_roots_started,
+    if (atomic_load_relaxed(&global_roots_scanned) == WORK_UNSTARTED &&
+        atomic_compare_exchange_strong(&global_roots_scanned,
```
NickBarnes, line 1436: I rely on your knowledge of the C memory model etc: are we guaranteed that _some_ domain succeeds at this CAS? Or is it possible for them all to fail, in which case we need something else here....

stedolan: Yes, you're guaranteed that if the CAS fails then a value other than `WORK_UNSTARTED` must have been read from `global_roots_scanned`, which means some other domain started successfully.

(This is what the `_strong` suffix means: there's also a `_weak` version that's allowed to fail spuriously, and may be faster on some platforms)

mshinwell (approved):
On behalf of Nick Barnes

PR #3031, opened 2024-09-12
Add directory and file annotation to profile output

milan-tom:
Changes:
- Adds directory (from clflag `directory` if passed in) to file in profile output.
- Adds `file=<FILE_NAME>//` annotation to profile CSV output for `.ml`, `.mli` and linked files.
- Changes summary CSV generation script to include file as separate column.
- Uses `x86_proc.string_of_symbol` to sanitise function names rather than ad-hoc sanitiser

### DIFF
--- a/backend/asmgen.ml
+++ b/backend/asmgen.ml
@@ -573,7 +573,8 @@ let compile_phrases ~ppf_dump ps =
         let profile_wrapper =
           match !profile_granularity with
           | Function_level | Block_level ->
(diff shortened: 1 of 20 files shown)


xclerc (commented):
The expected outputs under `tests/profile/counters/language_extensions`
need to be updated with the new `file=` prefix; otherwise, it is ready to
be merged.

xclerc (approved):

PR #3032, opened 2024-09-12
Bound the maximum GC work caused by custom blocks

stedolan:
The current heuristics for accelerating the GC due to off-heap GC-managed data (bigarrays, etc.) have some pathological behaviour. (The misbehaviour is old, but affects OCaml 5 worse, since runtime5 is better at following bad instructions about how much GC to do than runtime4 was).

This patch doesn't fix the heuristics. That work is underway, but not yet done.

This patch bounds how bad the decisions can be, by limiting the work caused by custom blocks to a fixed multiple of the work that would otherwise be done. In some testing, this has been sufficient to avoid the really bad behaviours.

To make this sort of tweaking of GC parameters easier to experiment with, this PR also adds a new module `Gc.Tweak` allowing such parameters to be easily defined, and edited both at runtime and from `OCAMLRUNPARAM`.


ocaml/stdlib/gc.mli
```diff
@@ -623,3 +623,20 @@ module Memprof :
        called on a profile which has not been stopped.
        *)
 end
+
+
+(** GC Tweaks are unstable and undocumented configurable GC parameters,
+    primarily intended for use by GC developers. *)
```
ncik-roberts, line 629: Could you document the command-line syntax here?

ncik-roberts (approved):
This looks great to me. I've pushed some help text for the case where you misspell the tweak name on the command line. Maybe we should add something similar for users of `Gc.Tweak.{get,set}` but I'm not sure if this is worth it.

ocaml/runtime/gc_ctrl.c
```diff
@@ -417,3 +419,73 @@ CAMLprim value caml_ml_runtime_warnings_enabled(value unit)
...
+  for (int i = 0; i < N_GC_TWEAKS; i++) {
+    if (strlen(gc_tweaks[i].name) == len &&
+        memcmp(gc_tweaks[i].name, name, len) == 0) {
+      return gc_tweaks[i].ptr;
+    }
+  }
+  return NULL;
+}
+
+CAMLprim value caml_gc_tweak_get(value name)
```
ncik-roberts, line 460: Should we add `CAMLparam` to both `caml_gc_tweak_get` and `caml_gc_tweak_set`? I believe the current impls are correct, but it's nice for their correctness to not be sensitive to whether they allocate.

stedolan: Yeah, we absolutely should.

This is a prime example of "code written by people who know too much about the GC", which is a huge source of bugs (if not now, then after the first reasonable-looking refactoring)

ocaml/stdlib/gc.mli
stedolan: Good point, will do.

stedolan:
> I've pushed some help text for the case where you misspell the tweak name on the command line — feel free to change that if you don't like it.

Thanks, looks good! I made one small change: now, when you misspell the name you get a one-line error message asking you to try `Xhelp=1`, and only when you pass that do you get the full list. (I suspect this list may grow long, and I dislike it when tools spam a terminal-overflowing amount of information in response to a typo)

ncik-roberts (approved):
Re-approving in light of the new changes (thanks).

ncik-roberts:
I think this was incidentally closed due to the merging of `merge-5.2-staging`, but should actually remain open and just target `main`.

PR #3038, opened 2024-09-16
Fix `unique_barrier` assertion

riaqn:
This PR fixes an assertion failure in the UA.

The assertion thinks that `unique_barrier ref` can be written to only once, which is wrong. For example:
```
match t with
| Foo x | Bar x -> x.a; f x
```
The `x` here will map to two `Path.t`, one for `Foo x` and one for `Bar x`.  The field projection `x.a` generates a usage tree where both `Path.t` maps to the same unique_barrier. This followed by `f x` causes the unique_barrier being written to twice.

Request review from @anfelor

### DIFF
--- a/ocaml/testsuite/tests/typing-unique/pr3038.ml
+++ b/ocaml/testsuite/tests/typing-unique/pr3038.ml
@@ -1,0 +1,19 @@
+(* TEST
+ readonly_files = "pr3038.ml";
+setup-ocamlc.byte-build-env;
+module = "pr3038.ml";
+ocamlc.byte;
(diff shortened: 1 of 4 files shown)


anfelor (commented):
I agree that the example given is the cause of the bug and the fix LGTM.

goldfirere:
I don't see any tests here -- should there be?

anfelor:
The `unique_barrier` currently has no effect and I would expect to add some tests to my patch that test its behavior once properly implemented. We could add the example above to the testsuite though.

goldfirere:
OK.

PR #2834, opened 2024-07-22
Migrate labeled tuples to the parsetree

freemagma:
Migrates labeled tuples away from jane syntax and into the parsetree

PR #3040, opened 2024-09-16
Migrate labeled tuples from Jane Syntax to the parsetree

rtjoa:
Based off of #2834

### DIFF
--- a/ocaml/parsing/ast_helper.ml
+++ b/ocaml/parsing/ast_helper.ml
@@ -100,7 +100,8 @@ module Typ = struct
             Ptyp_var x
         | Ptyp_arrow (label,core_type,core_type',modes,modes') ->
             Ptyp_arrow(label, loop core_type, loop core_type', modes, modes')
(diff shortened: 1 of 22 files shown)


testsuite/tools/expect_test.ml
```diff
@@ -76,8 +76,8 @@ let match_expect_extension (ext : Parsetree.extension) =
         let normal, principal =
           match e.pexp_desc with
           | Pexp_tuple
-              [ a
-              ; { pexp_desc = Pexp_construct
```
ccasin, line 80: Here also we should probably preserve the old behavior by checking for `None`.

Well, I say the old behavior - I guess the old one ignored labels because they were in attributes, but I think it's fair to say the intent was to match on only unlabeled tuples?  Or let me know if you disagree.

ccasin (approved):
Looks good, some minor questions/comments.

rtjoa (commented):
@ccasin Thanks for the review! Rebased and pushed updates.

PR #3042, opened 2024-09-16
Reorder fields in block header

ahh:
Place wosize above scannable prefix, simplifying expressions like [Array.length].

[Array.length] and other block-size-computing expressions are fairly common in real code, and this shrinks them in code size and also operations executed, a rough diff being before:

      movabs  rbx, 72057594037927935
        mov     rax, qword ptr [rax - 8]
        and     rax, rbx
        shr     rax, 9
        or      rax, 1

and after:

	mov    rdx,QWORD PTR [rax-0x8]
 	shr    rdx,0x11
 	or     rdx,0x1

In addition I think this is just more readable when trying to pore over generated assembly.

Getting the scannable prefix size is, I think, as bad as getting block sizes is now: but the only time we do that is in GC scanning, and we also need lengths, so the total cost should be unch'd.

I tested both runtime4 and runtime5 via `make runtest`, and inspected some amount of generated assembly.

### DIFF
--- a/backend/cmm_helpers.ml
+++ b/backend/cmm_helpers.ml
@@ -74,6 +74,12 @@ type t =
 (* Mixed blocks, that need special header to specify the length of the scannable
    prefix. *)
 
+let addr_size_bits = Arch.size_addr * 8
+
+let header_wosize_bits = addr_size_bits - Config.reserved_header_bits - 10
+
+let header_max_size = (1 lsl header_wosize_bits) - 1
+
 module Mixed_block_support : sig
   val assert_mixed_block_support : unit -> unit
 
@@ -93,16 +99,11 @@ end = struct
      the all-0 pattern, and we must subtract 2 instead. *)
   let max_scannable_prefix = (1 lsl required_reserved_header_bits) - 2
 
-  let max_header =
-    (1 lsl (required_addr_size_bits - required_reserved_header_bits)) - 1
-    |> Nativeint.of_int
-
   let assert_mixed_block_support =
     lazy
       (if not Config.native_compiler
        then Misc.fatal_error "Mixed blocks are only supported in native code";
        let reserved_header_bits = Config.reserved_header_bits in
-       let addr_size_bits = Arch.size_addr * 8 in
        match
          ( reserved_header_bits = required_reserved_header_bits,
            addr_size_bits = required_addr_size_bits )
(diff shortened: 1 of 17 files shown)


Svetlitski:
@ahh FWIW, we _could_ have been producing better assembly for computing Array length before. Compiling under `clang` with `-O3 -march=haswell`, this is currently what it comes up with for `Wosize_val(block)`:

```asm
array_length:                           # @array_length
        mov     eax, 56
        bzhi    rax, qword ptr [rdi - 8], rax
        shr     rax, 10
        ret
```

mshinwell:
@stedolan any opinions?

ncik-roberts:
We could generate better assembly (just a right shift) for arrays _without_ moving around the sections of the header. Currently, the top byte of an array header is always zero: arrays can’t be mixed blocks.

The approach in the present PR is more future proof (e.g. would allow us to decide to start using the top byte of array headers, e.g. skip scanning of int arrays). The downside of the present approach is that it introduces drift from upstream.

ahh:
@ncik-roberts to say the obvious it's presented for Array but there's also cases of wanting block length of non-arrays.

(not to mention that my Next Trick was going to be putting string length low bits into the scannable prefix field, which would be a huge simplification for string.length).  I agree there's upstream divergence but...

stedolan:
I do worry about this, since the OCaml header format has been stable and documented for so long that various bits of code "know" that the length is a 10-bit shift away. (As a random example, here's [code that reimplements Make_header for some reason](https://gitlab.com/gerdstolpmann/lib-ocamlnet3/-/blob/master/code/src/netsys/netsys_c_mem.c?ref_type=heads#L592-593)). We could definitely improve the header format, but it would likely be a bunch of work to track down all the weird code that relies on how it is now.

On the other hand, if we are going to bite this bullet and change the header format, I'd consider something more radical. How about we decide that blocks have 32-bit lengths, limiting the maximum single array to 32GB of data (at 8-byte words)? Then we can use the whole upper word for the length and compute Allocated_wosize_val as a 32-bit load with no bit twiddling, and do bounds checks as a fused `cmp+load` on x86. (And also use the high 3 bits of the next word for the string length mod 8 in the String_tag case, so we can finally have reasonable-looking string bounds checks)

Unrelatedly, the original example is just terrible codegen:

> ```
>   movabs  rbx, 72057594037927935
>     mov     rax, qword ptr [rax - 8]
>     and     rax, rbx
>     shr     rax, 9
>     or      rax, 1
> ```

Which compiler produced that? Is this C runtime code or OCaml-compiled code? Even without fancy BMI2 instructions, you can extract an arbitrary bitfield with `shl; shr` without loading a huge constant and burning another register. (Or, as Nick points out, just `shr` for the array case)

ncik-roberts:
NTS: we need to fix `ocaml/debugger/debugcom.ml` and `ocaml/debugger4/debugcom.ml`, which have `lsr 10`.

stedolan:
The codegen here annoyed me enough to write #3067.

mshinwell:
I think we probably need further discussion before merging this.

PR #3043, opened 2024-09-17
mode crossing in inclusion check

riaqn:
This PR adds the support for mode crossing in inclusion check.

The special case is the locality axis of the return mode of an arrow type, where mode crossing is not allowed. This is because a local-returning function might allocate in the caller's region, which is unsafe to forget. 

This also contains a fix to a test which I think is an improvement.

### DIFF
--- a/ocaml/testsuite/tests/typing-modes/crossing.ml
+++ b/ocaml/testsuite/tests/typing-modes/crossing.ml
@@ -2,13 +2,58 @@
  expect;
 *)
 
+(* mode crossing during inclusion check is according to the written type, not
+the inferred type. *)
+
+(* In this example, the inferred type does not allow mode crossing, but the
+written type does. *)
 module M : sig
     val f : int @ nonportable -> int @ portable
 end = struct
-    let f (_ @ portable) = (42 : _ @@ nonportable)
+    let f (x @ portable) = (x : _ @@ nonportable)
 end
 [%%expect{|
 module M : sig val f : int -> int @ portable end
+|}]
+
+(* In this example, the inferred type allows crossing to portable, but the
+written type does not. *)
+module M : sig
+    val f : unit -> [`A | `B of 'a -> 'a] @ portable
+end = struct
+    let f () = (`A : _ @@ nonportable)
+end
+[%%expect{|
+Lines 3-5, characters 6-3:
+3 | ......struct
+4 |     let f () = (`A : _ @@ nonportable)
+5 | end
+Error: Signature mismatch:
+       Modules do not match:
+         sig val f : unit -> [> `A ] end
+       is not included in
+         sig val f : unit -> [ `A | `B of 'a -> 'a ] @ portable end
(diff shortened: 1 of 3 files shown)


ocaml/typing/ctype.ml
```diff
@@ -4724,12 +4724,46 @@ let relevant_pairs pairs v =
...
+  (* CR layouts v2.8: This should probably check for principality. See similar
+      comment in [mode_cross_left]. *)
+  let jkind = type_jkind_purely env ty in
+  let upper_bounds = Jkind.get_modal_upper_bounds jkind in
+  Alloc.imply upper_bounds mode
+
+let submode_with_cross env ~is_ret ty l r =
+  let r' = mode_cross_right env ty r in
+  let r' =
+    if is_ret then
```
goldfirere, line 4751: This special case needs a loud comment

ocaml/typing/ctype.ml
```diff
@@ -4777,8 +4811,10 @@ let rec moregen inst_nongen variance type_pairs env t1 t2 =
                 || !Clflags.classic && equivalent_with_nolabels l1 l2) ->
               moregen inst_nongen (neg_variance variance) type_pairs env t1 t2;
               moregen inst_nongen variance type_pairs env u1 u2;
-              moregen_alloc_mode (neg_variance variance) a1 a2;
-              moregen_alloc_mode variance r1 r2
+              (* [t2] and [u2] is the user-written interface, which we deem as
```
goldfirere, line 4814: I don't understand this comment -- what is the impact of this choice? It might be good to e.g. have a test that demonstrates the difference in behavior between choosing the `2` types vs the `1` types here. Then we could refer to the test from this comment to illustrate the point.

goldfirere (approved):
Looks good -- requested a little more commentary.

ocaml/testsuite/tests/typing-modes/crossing.ml
```diff
@@ -2,15 +2,60 @@
...
     val f : int @ nonportable -> int @ portable
 end = struct
-    let f (_ @ portable) = (42 : _ @@ nonportable)
+    let f (x @ portable) = (x : _ @@ nonportable)
 end
 [%%expect{|
 module M : sig val f : int -> int @ portable end
 |}]
 
+(* In this example, the inferred type allows crossing to portable, but the
```
goldfirere, line 19: Rejecting this is a bit unfortunate, because it would be sound to accept. When the inferred/written types differ, mode crossing at *either one* is fine, essentially because the mode-cross and subtype operations can compose in either direction. I can even imagine scenarios where we want mode-cross then subtype then mode-cross — which would be fine! (Maybe that doesn’t exist for the current subtype relation; not sure.)

I don’t think we need to implement the mode-cross-at-either-type behavior right now, but maybe just leave a comment here saying that we could do better. If folks complain, we’ll return.

goldfirere (approved):

PR #3044, opened 2024-09-17
%identity-related changes

mshinwell:
This uses `%obj_magic` for `Obj.repr` and `Obj.obj`, which seems consistent with the handling of `Obj.magic`, and would have prevented invalid traps in user code recently.

It also updates the invalid trap message, which was out of date.


lthls (approved):
Fine with me.
It seems a bit weird to remove the recommendation to use `Sys.opaque_identity`, but I agree it wouldn't really fit with the new message.

PR #3048, opened 2024-09-17
Zero alloc: remove "assume" annotation when a call is inlined

gretay-js:
Currently, during translation to "lambda" the compiler adds "zero_alloc assume" annotations on a call if the callee's type is anntoated with "zero_alloc" (see #2506). If the middle-end inlines the callee, this "zero_alloc assume" annotation is propagated onto all primitives in the inlined body of the callee. This is not needed: the inlined body is checked separately against its "zero_alloc" annotation. More improtantly, it can be unsound in the presence of some optimizations, e.g., if an allocation that occurs only on an exceptional return from a function is moved such that it occurs on paths to normal return.

I have an example that demonstrates the unsound behavior. I am still trying to minimize it so I can add it as a test.  

Separately, I still don't fully understand how to fix the optimization: after simplify in the middle-end, the allocation appears on all paths but the translation to_cmm  moves it and is able to eliminate it from all normally returning paths.  Disabling CSE doesn't help. The examples also involves unboxed types and the allocations come from operations on unboxed int64 that go through boxed int64 and are supposed to be compiled away by flambda2.


gretay-js:
I've just added a test that shows the expected behavior with this PR: zero_alloc check fails. Prior to this PR, the check passed, because an allocation was annotated with "assume".  This is not sound because the allocation in the callee (where it is checked separately) appeared on the exceptional path, but after the body of the callee was inlined, the allocation appears on the normal path as well. It looks like the allocation was not optimized away during during translation to cmm. 

@Gbury can you pelase have a look at the example? If this explanation is correct, is there a way to improve the heuristic for cmm unboxing to handle such cases?

This example shows that the compiler doesn't always optimize allocations around Int64_u operations (implemented via boxed int64 operations).

ocaml/lambda/debuginfo.ml
```diff
@@ -342,7 +342,12 @@ let to_location { dbg; assume_zero_alloc=_ } =
 
 let inline { dbg = dbg1; assume_zero_alloc = a1; }
       { dbg = dbg2; assume_zero_alloc = a2; } =
-  { dbg = dbg1 @ dbg2; assume_zero_alloc = ZA.Assume_info.meet a1 a2; }
+  { dbg = dbg1 @ dbg2;
+    assume_zero_alloc =
+      (* Drop "inferred" zero_alloc annotation from a call when
+         the callee is inlined. *)
+      if ZA.Assume_info.is_inferred a1 then a2 else
+      ZA.Assume_info.meet a1 a2; }
```
ccasin, line 350: I think this means the order of arguments to `inline` now matters, but didn't before.  Can we add a comment to the mli clarifying which argument is being inlined into the other?  Alternatively it might also be good to give one of the arguments a label making the order clear (e.g., `~into`?).  That way we'll be forced to look at all call sites and can verify they use the expected order.

I think the way this should work is that we look at the debug info of the function being inlined - if that one has an inferred za assume, we drop it.  I see a callsite in flambda like:
```ocaml
      Debuginfo.inline t.dbg dbg_from_inlined_body
```
Which makes me think we should be checking if the second one is inferred, not the first?  But I'm confused why this didn't break the tests if my understanding is correct.  It also looks like there is at least one caller in dead code under `ocaml/` for the old middle ends that uses the other order.

ccasin (commented):
A question and an annoying style request (sorry)

tests/backend/zero_alloc_checker/test_remove_inferred_assume.ml
```diff
@@ -0,0 +1,88 @@
...
+  let pos = Offset.to_int offset in
+  if pos < 0
+  then
+    raise (Failure "Boo")
+
+
+let[@inline] [@zero_alloc] read_exn offset chunk_length  = exclave_
+  bounds_check ~offset;
+  Offset.to_int offset
+
```
ccasin, line 84: extra whitespace

ocaml/lambda/debuginfo.ml
gretay-js: I think the order mattered before, because of list concatenation in L345. I absolutely agree that the direction is confusing and I added a label (similarly to the one in flambd2).

gretay-js: > I think the way this should work is that we look at the debug info of the function being inlined - if that one has an inferred za assume, we drop it. 

No, we look at the debug info on the call site: if it has "assume_inferred" annotation, then we use the annotation from the inlined body.

ccasin: Thanks, I agree I had this backwards yesterday.

ccasin (approved):
lgtm, but would be good to have @Gbury look at the test and confirm we understand it correctly

Gbury:
I'm looking at it, ^^

Gbury:
Firstly, when running the example from the PR, I get more errors from the compiler than the output commited for the test, notably, an error from the definition of `Offset.to_int` at line 70, which is indeed not `no_alloc` (though that may depend on the exact meaning, since the only allocation is on a `cold` branch that ends in a call to `int64_to_int_failure`). So I'm a bit confused as to why I seem to have a different output than CI does.

Concerning the `read_exn` function, I think again the answer would be PDCE : this is a case where one value is used in two distinct branches, both of which happen to be "cold" branches if I understand correctly (also note that the fact that the same value is used for both branches even though the original code has two boxing operations is because of the CSE performed by flambda).

Trying to tweak `to_cmm` to sink down the allocations in this case seem too complex to me: contrary to `Simplify` , `to_cm` tries to be a bit simpler, and lacks a lot of the information propagation mechanism that would allow for it to reasonably sink down the allocations if and only if they are used only in cold branches.

gretay-js:
> Firstly, when running the example from the PR, I get more errors from the compiler than the output commited for the test, notably, an error from the definition of Offset.to_int at line 70, which is indeed not no_alloc (though that may depend on the exact meaning, since the only allocation is on a cold branch that ends in a call to int64_to_int_failure). So I'm a bit confused as to why I seem to have a different output than CI does.

Does the extra error show up locally with `make runtest` for you ? If it's in a separate build from a command line, maybe a compilation flag is missing? Both `-g` and `-O3` are needed.

> Concerning the read_exn function, I think again the answer would be PDCE : this is a case where one value is used in two distinct branches, both of which happen to be "cold" branches if I understand correctly (also note that the fact that the same value is used for both branches even though the original code has two boxing operations is because of the CSE performed by flambda).

ok, thanks. this matches with what @ccasin discussed. 

> Trying to tweak to_cmm to sink down the allocations in this case seem too complex to me: contrary to Simplify , to_cmm tries to be a bit simpler, and lacks a lot of the information propagation mechanism that would allow for it to reasonably sink down the allocations if and only if they are used only in cold branches.

In the example, to_cmm successfully sinks down allocations in `bounds_check` but not in `read_exn` where the allocation is a result of cse. I thought that the allocation is "pushed down" in `bounds_check` unconditionally (not only to cold branches), and then eliminated by cmm unboxing on "hot" branches. Why doesn't it happen for `read_exn` too?

gretay-js:
A workaround suggested by @mshinwell :
```
let int64_to_int_exn x =
    if int64_is_representable_as_int x then
      int64_to_int_trunc x
    else
      let x = Int64.add (globalize_int64 x) (Sys.opaque_identity 0L) in
      int64_to_int_failure x
```
The extra line in the failure case triggers the missing optimization and the check passes. I added a test that demonstrates it.

Gbury:
> Does the extra error show up locally with `make runtest` for you ? If it's in a separate build from a command line, maybe a compilation flag is missing? Both `-g` and `-O3` are needed.

So I tried again, and it's possible I forgot a `-O3` indeed, sorry for the noise, ^^

> Why doesn't it happen for `read_exn` too? @Gbury

Short answer: because in `bounds_check` the allocation is used exactly once, whereas in `read_exn` it has more than one occurrence.

Actually, `to_cmm` does not really "sink down" allocations, what really happens is that when it encounters a let-binding `let x = def_expr in body `, `to_cmm` can decide to substitute every use of `x` in `body` by `def_expr` (assuming `def_expr` has no side-effects, or it has coeffects but there are no effects between the let-binding and the uses of `x`). In general, this is **not** an optimization, since it may duplicate code, and in particular it can duplicate allocations (which is really the opposite of what we want to do in flambda2). Therefore, this substitution is only done when there is exactly one use of `x`, not under a loop, and with the restriction on effects described before (it also happens for boxed number allocations in classic mode, but that's unrelated).

So, in `bounds_check`, the Flambda2 optimizer is actually able to remove all uses of the boxed integer except for the one in the cold/exn branch, so `to_cmm` will perform the substitution, since at that point, there is only one occurrence of it remaining. However, in the `read_exn` function, the boxed integer is actually used in multliple places, notably because there are two remaining calls to `int64_to_int_failure` in the function body, so `to_cmm` does not try to substitute it since there's not enough information to know whether that would be a good idea or not (note that the decision has to be made when looking at the let-binding, before exploring the body).

Does that clear things up ?

gretay-js:
@Gbury thank you for the explanation!

PR #3049, opened 2024-09-18
Support default modalities for signatures

riaqn:
This PR adds the support for default modalities on signatures, including file-level signatures (`.mli`).

Please see added tests for syntax.

Please do not merge until I finish the internal migration.

### DIFF
--- a/ocamldoc/odoc_sig.ml
+++ b/ocamldoc/odoc_sig.ml
@@ -1628,15 +1628,15 @@ module Analyser =
           Module_type_alias { mta_name = Odoc_env.full_module_name env name ;
                               mta_module = None }
 
-      | Parsetree.Pmty_signature {psig_items; _} ->
+      | Parsetree.Pmty_signature {psg_items; _} ->
           (
-           let psig_items = filter_out_erased_items_from_signature erased psig_items in
+           let psg_items = filter_out_erased_items_from_signature erased psg_items in
            (* we must have a signature in the module type *)
            match sig_module_type with
              Types.Mty_signature signat ->
                let pos_start = Loc.start module_type.Parsetree.pmty_loc in
                let pos_end = Loc.end_ module_type.Parsetree.pmty_loc in
-               let elements = analyse_parsetree env signat current_module_name pos_start pos_end psig_items in
+               let elements = analyse_parsetree env signat current_module_name pos_start pos_end psg_items in
                Module_type_struct elements
            | _ ->
                raise (Failure "Parsetree.Pmty_signature signature but not Types.Mty_signature signat")
@@ -1725,9 +1725,9 @@ module Analyser =
             | _ ->
               raise (Failure "Parsetree.Pmty_alias _ but not Types.Mty_alias _")
            end
-      | Parsetree.Pmty_signature {psig_items; _} ->
+      | Parsetree.Pmty_signature {psg_items; _} ->
           (
-           let psig_items = filter_out_erased_items_from_signature erased psig_items in
+           let psg_items = filter_out_erased_items_from_signature erased psg_items in
            match sig_module_type with
              Types.Mty_signature signat ->
                Module_struct
@@ -1737,7 +1737,7 @@ module Analyser =
                     current_module_name
                     (Loc.start module_type.Parsetree.pmty_loc)
                     (Loc.end_ module_type.Parsetree.pmty_loc)
-                    psig_items
+                    psg_items
                  )
            | _ ->
                (* if we're here something's wrong *)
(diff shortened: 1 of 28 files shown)


ocaml/testsuite/tests/language-extensions/pprintast_unconditional.reference
```diff
@@ -6,26 +6,29 @@ modality_record:
...
       z: bool @@ unique }
   end
 
 modality_cstrarg:
   struct
     type t =
-      | Foo of global_ string * global_ string @@ portable *
-      string @@ portable 
+      | Foo of global_ string * global_ string @@ portable * string
+      @@ portable
```
goldfirere, line 19: This seems like worse pretty-printing than previously.

ocaml/testsuite/tests/typing-modes/val_modalities.ml
```diff
@@ -549,3 +549,63 @@ module type S' =
...
+
+module type T = sig @@ portable
+  include functor S @@ nonportable
+end
+[%%expect{|
+module type T =
+  sig
+    val foo : 'a -> 'a
+    val bar : 'a -> 'a
+    val baz : 'a -> 'a @@ portable
```
goldfirere, line 581: I find this quite surprising. This means, I think, that

```
sig
  val f : ... @@ nonportable
end
```

and

```
sig
  val f : ...
end
```

are different -- even though `nonportable` is the default. Is there a design document somewhere that describes this? I want to understand it better. My general philosophy is that having a default meaning when something is omitted (like the portability modality) is great, but that there should always be a way of writing it explicitly if you want. So is there a `val f : ... @@ .` syntax? But how would that know that you want the blank modality for portability specifically?

ocaml/testsuite/tests/typing-modes/val_modalities.ml
```diff
@@ -549,3 +549,63 @@ module type S' =
 
 (* CR zqian: add tests of recursive modules & include w/ modalties, once
    modules can have modes. *)
+
```
goldfirere, line 552: It might be nice to see how this feature interacts with `include`, not just `include functor`. Also it would be good to see that the `@@ portable` thing in a `sig` is disregarded in the module inclusion check between two module types -- one with `@@ portable` up top and the other with it per-function.

goldfirere (commented):
Review includes tests only. I'd like to nail down the specification before diving into the implementation.

ocaml/testsuite/tests/typing-modes/val_modalities.ml
goldfirere: I think I just misunderstood this -- sorry!

goldfirere:
Sorry to lose track of this, but is this ready for review now?

riaqn:
@goldfirere sorry for the confusion - it's mostly ready, but as I migrate PPXes I sometimes have to slightly modify this PR. I will let you know when it's ready. Thanks!

ocaml/testsuite/tests/parsetree/source_jane_street.ml
```diff
@@ -450,8 +450,8 @@ val f1 : ('a : value_or_null). local_ 'a -> unit @@ global many = <fun>
 val f2 : unit -> unit @@ global many = <fun>
 |}]
 
-module type S = sig
-  module type S0 = sig
+module type S = sig @@ portable contended
+  module type S0 = sig @@ portable contended
```
goldfirere, line 454: Shouldn't this change the output below?

ocaml/testsuite/tests/parsetree/source_jane_street.ml
goldfirere, line 1: Please also add a demonstration of `[@@no_recursive_modalities]`.... or is that no longer part of this patch?

ocaml/testsuite/tests/typing-modes/val_modalities.ml
goldfirere, line 1: There should also be a handful of tests for `[@no_recursive_modalities]` here.

goldfirere (approved):

ocaml/parsing/ast_mapper.ml
```diff
@@ -1388,26 +1394,27 @@ let apply_lazy ~source ~target mapper =
...
-      | _ -> [], ast
+      | _ -> [], psig_items
     in
     PpxContext.restore fields;
-    let ast =
+    let psig_items =
       try
         let mapper = mapper () in
-        mapper.signature mapper ast
+        List.map (mapper.signature_item mapper) psig_items
```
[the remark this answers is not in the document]
riaqn: Yeah that's better. In the long run we should introduce `psg_attributes` to contain the `ocaml.ppx.context` thingy, that should make the logic here shallower.

ocaml/testsuite/tests/parsetree/source_jane_street.ml
riaqn: No, this is just checking the roundtrip-ness of user-syntax print back, so it's working fine.

ocaml/testsuite/tests/parsetree/source_jane_street.ml
riaqn: Yeah it's not part of this patch. It's already at the end of this file.

ocaml/testsuite/tests/typing-modes/val_modalities.ml
riaqn: There are already some, introduced by #3095

goldfirere (approved):
:shipit:

PR #3050, opened 2024-09-18
Add shared mode

ageorg29:
Add `shared` mode between `contended` and `uncontended`

### DIFF
--- a/ocaml/testsuite/tests/typing-modes/portable-contend.ml
+++ b/ocaml/testsuite/tests/typing-modes/portable-contend.ml
@@ -62,20 +62,15 @@ Error: This value is "shared" but expected to be "uncontended".
   this record needs to be uncontended.
 |}]
 
-(* Reading from a mutable field in a shared record is fine *)
-let foo (r @ shared) = {r with a = best_bytes ()}
+(* reading mutable field from shared record is fine *)
+let foo (r @ shared) = r.a
 [%%expect{|
-val foo : r @ shared -> r @ shared = <fun>
+val foo : r @ shared -> bytes @ shared = <fun>
 |}]
 
 let foo (r @ shared) = {r with b = best_bytes ()}
 [%%expect{|
 val foo : r @ shared -> r @ shared = <fun>
-|}]
-
-let foo (r @ shared) = r.a
-[%%expect{|
-val foo : r @ shared -> bytes @ shared = <fun>
 |}]
 
 (* reading immutable field from contended record is fine *)
@@ -88,6 +83,11 @@ val foo : r @ contended -> bytes @ contended = <fun>
 let foo (r @ shared) = r.b
 [%%expect{|
 val foo : r @ shared -> bytes @ shared = <fun>
+|}]
+
+let foo (r @ shared) = {r with a = best_bytes ()}
+[%%expect{|
+val foo : r @ shared -> r @ shared = <fun>
 |}]
 
 (* Force top level to be uncontended and nonportable *)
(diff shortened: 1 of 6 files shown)


ocaml/testsuite/tests/typing-modes/portable-contend.ml
```diff
@@ -46,17 +46,50 @@ let foo (r @ contended) = {r with b = best_bytes ()}
...
+[%%expect{|
+Line 1, characters 23-24:
+1 | let foo (r @ shared) = r.a <- 42
+                           ^
+Error: This value is shared but expected to be uncontended.
+  Hint: In order to write into the mutable fields,
   this record needs to be uncontended.
 |}]
 
+(* Reading from a mutable field in a shared record is fine *)
```
goldfirere, line 65: This comment doesn't seem to go with this test -- I don't see reading of a mutable field.

ocaml/testsuite/tests/typing-modes/portable-contend.ml
```diff
@@ -143,11 +184,38 @@ let foo (r @ contended) =
 Line 3, characters 6-16:
 3 |     | [| x; y |] -> ()
           ^^^^^^^^^^
-Error: This value is contended but expected to be uncontended.
+Error: This value is contended but expected to be shared.
   Hint: In order to read from the mutable fields,
-  this record needs to be uncontended.
+  this record needs to be at least shared.
```
goldfirere, line 189: It's an array, not a record. (This is not a new bug.)

Perhaps include a
```
(* CR modes: Error message should mention array, not record. *)
```

ocaml/testsuite/tests/typing-modes/portable-contend.ml
```diff
@@ -143,11 +184,38 @@ let foo (r @ contended) =
...
+Line 1, characters 33-34:
+1 | let foo (r @ shared) = Array.set r 42 (best_bytes ())
+                                     ^
+Error: This value is shared but expected to be uncontended.
 |}]
 
+(* The signature of Array.get could be generalized to expect shared rather than
+   uncontended, but this would require a change to stdlib. For now the following
+   test fails *)
+let foo (r @ shared) = Array.get r 42
```
goldfirere, line 203:
```suggestion
(* CR modes: Fix this *)
let foo (r @ shared) = Array.get r 42
```

ocaml/testsuite/tests/typing-modes/portable-contend.ml
```diff
@@ -225,15 +301,33 @@ Error: This function when partially applied returns a value which is nonportable
        but expected to be portable.
 |}]
 
+(* closing over shared gives nonportable *)
+let foo : 'a @ shared portable -> (unit -> unit) @ portable = fun a () -> ()
+[%%expect{|
+Line 1, characters 62-76:
+1 | let foo : 'a @ shared portable -> (unit -> unit) @ portable = fun a () -> ()
```
goldfirere, line 308: This surprises me. The returned closure does not use the fact that `a` is shared. So (I think) we can submode the `a` to be `contended` and then close over the contended `a` without make the closure `nonportable`.

ocaml/testsuite/tests/typing-modes/portable-contend.ml
goldfirere, line 1: I didn't see a test that closes over a shared value (that is actually read from) and produces a nonportable closure. Could you please add?

goldfirere (commented):
Looks good! Just a few tweaks in the tests.

ocaml/testsuite/tests/typing-modes/portable-contend.ml
ageorg29: Indeed, I see what you mean. This test has the same behavior as: 
`let foo : 'a @ uncontended portable -> (unit -> unit) @ portable = fun a () -> ()`
which also gives an error, even though `a` (here `uncontended`) is not used.

goldfirere: OK. We don't have to fix this here. But I think this is another good time to add `(* CR modes: This should be allowed *)` to encourage us to return.

goldfirere (approved):

PR #3051, opened 2024-09-18
Fix caml_floatarray_create_local

mshinwell:
Fixes the C stub `caml_floatarray_create_local` to cope with the case where the given length is zero or out of bounds.

Exposes this functionality as `Float.Array.create_local`.  We might want to rename it to `create_stack`, @ncik-roberts is deciding.


mshinwell:
Update: we won't yet expose this in the stdlib (consistent with local normal arrays, and we really need to expose the various accessor functions at local mode too).

ncik-roberts (approved):

PR #3052, opened 2024-09-19
Support loosening monadic return modes in [type_argument]

lpw25:
Currently we allow comonadic return modes and all argument modes to be "weakened" in `type_argument`. That allows code like:
```ocaml
let use_global_local (f : string @ global -> string @ local) = ()
let local_global : string @ local -> string @ global = fun _ -> "hello"
let () = use_global_local local_global
```
However, for historical reasons we don't allow it for monadic return modes. So this doesn't work:
```ocaml
let use_contended_ret (f : unit-> int ref @ contended) = ()
let uncontended_ret : unit -> int ref @ uncontended = fun _ -> ref 0
let () = use_contended_ret uncontended_ret
```
There is no longer any reason for this, so this PR removes the restriction.

### DIFF
--- a/ocaml/testsuite/tests/typing-modes/modes.ml
+++ b/ocaml/testsuite/tests/typing-modes/modes.ml
@@ -376,3 +376,179 @@ Line 2, characters 38-41:
                                           ^^^
 Error: The extension "mode" is disabled and cannot be used
 |}]
+
+
+(*
+ * Modification of return modes in argument position
+ *)
+
+let use_local (f : _ -> _ -> _ @@ local) x y =
+  f x y
+let result = use_local (^) "hello" " world"
+[%%expect{|
+val use_local : local_ ('a -> 'b -> 'c) -> 'a -> 'b -> 'c = <fun>
+val result : string = "hello world"
+|}]
+
+let use_local_ret (f : _ -> _ @ local) x y =
+  let _ = f x in ()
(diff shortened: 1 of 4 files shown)


ocaml/testsuite/tests/typing-modes/modes.ml
```diff
@@ -376,3 +376,179 @@ Line 2, characters 38-41:
...
+
+let use_local (f : _ -> _ -> _ @@ local) x y =
+  f x y
+let result = use_local (^) "hello" " world"
+[%%expect{|
+val use_local : local_ ('a -> 'b -> 'c) -> 'a -> 'b -> 'c = <fun>
+val result : string = "hello world"
+|}]
+
+let use_local_ret (f : _ -> _ @ local) x y =
```
goldfirere, line 393: Is this actually OK to accept, given the extra meaning of `local` in return position? That is, `use_local_ret` will assume `f` allocates in the parent stack frame, but it won't. Our calling convention is OK with this?

goldfirere (approved):

ocaml/testsuite/tests/typing-modes/modes.ml
lpw25: It’s fine, we only need to avoid unexpectedly allocating in a caller. This isn’t new behaviour in this PR though, or even a new test, I just moved the test from elsewhere.

PR #3056, opened 2024-09-20
Remove Lambda.alloc_mode from backend/

anfelor:
This PR separates the definition of `alloc_mode` in Lambda from uses in the backend as discussed this morning with @mshinwell and @goldfirere. In defines two new `alloc_mode` definitions in `cmm.mli` and `cmx_format.mli`, which can not be identical: we can not import `cmm.mli` in `cmx_format.mli` and we do not want to add too many definitions to `cmx_format.mli`. Surprisingly, we never need to convert between those two definitions. This distinction seems to occur already in `alloc_mode.mli`, where `For_applications` is always extracted to the `cmx_format` definition, while `For_allocations` is extracted to the cmm definition. To avoid adding extra dependencies, I moved these conversions to `to_cmm_shared.mli`.

Also note that this PR is against `main`, not `overwriting`, since there are no changes to the semantics.


mshinwell:
I'll review this, I don't think @goldfirere needs to.

backend/cmm.ml
```diff
@@ -222,6 +222,30 @@ type static_cast =
...
+  type t = Heap | Local
+
+  let equal a b =
+    match a, b with
+    | Heap, Heap -> true
+    | Local, Local -> true
+    | Heap, Local -> false
+    | Local, Heap -> false
+
+  let print a =
```
mshinwell, line 235: This should be of type `Format.formatter -> t -> unit`, like so:
```ocaml
let print ppf t =
  match t with
  | Heap -> Format.fprintf ppf "Heap"
  ...
```

backend/cmm.ml
```diff
@@ -222,6 +222,30 @@ type static_cast =
   | V128_of_scalar of Primitive.vec128_type
   | Scalar_of_v128 of Primitive.vec128_type
 
+module Alloc_mode = struct
+  type t = Heap | Local
+
+  let equal a b =
```
mshinwell, line 228: We usually use `t1` and `t2` here, since these are of type `t`.

backend/cfg/cfg_comballoc.ml
```diff
@@ -88,8 +88,8 @@ let find_compatible_allocations :
         else { allocations = List.rev allocations; next_cell = Some cell }
       | Op (Begin_region | End_region) -> (
         match curr_mode with
-        | Lambda.Alloc_local -> return ()
-        | Lambda.Alloc_heap ->
+        | Cmm.Alloc_mode.Local -> return ()
```
mshinwell, line 91: I'm surprised you can't just write `Local` here - can you?

mshinwell: There are various other occurrences of this elsewhere e.g. in `emit.mlp` - please omit the module path if possible.

backend/cmm_helpers.ml
```diff
@@ -256,42 +256,42 @@ let closure_info ~(arity : arity) ~startenv ~is_last =
 
 let alloc_boxedfloat32_header mode dbg =
```
mshinwell, line 257: For these I would just write `let alloc_boxedfloat32_header (mode : Cmm.Alloc_mode.t) dbg = ...` then you can omit all the module paths, which tend to clutter.

backend/cmm_helpers.ml
```diff
@@ -1580,9 +1582,13 @@ let unique_arity_identifier (arity : Cmm.machtype list) =
 let result_layout_suffix result =
   match result with [| Val |] -> "" | _ -> "_R" ^ machtype_identifier result
 
-let send_function_name arity result (mode : Lambda.alloc_mode) =
+let send_function_name arity result (mode : Cmx_format.alloc_mode) =
   let res = result_layout_suffix result in
-  let suff = match mode with Alloc_heap -> "" | Alloc_local -> "L" in
+  let suff =
+    match mode with
+    | Cmx_format.Alloc_heap -> ""
```
mshinwell, line 1589: (Same here)

file_formats/cmx_format.mli
```diff
@@ -35,7 +35,11 @@ open Misc
 type machtype_component = Val | Addr | Int | Float | Vec128 | Float32
 type machtype = machtype_component array
 
-type apply_fn := machtype list * machtype * Lambda.alloc_mode
+type alloc_mode =
```
mshinwell, line 38: Maybe add a comment that this can't use `Cmm` because of a circular dependency.

middle_end/flambda2/to_cmm/to_cmm_shared.ml
```diff
@@ -426,3 +426,15 @@ let extended_machtype_of_return_arity arity =
   | arity ->
     (* Functions returning multiple values *)
     List.map extended_machtype_of_kind arity |> Array.concat
+
+let alloc_mode_to_cmx t =
```
mshinwell, line 430: Maybe `alloc_mode_for_applications_to_cmx` would be clearer, and similarly for the other function below.

mshinwell (changes requested):

anfelor:
I have implemented all comments in my latest two commits.

mshinwell (approved):

PR #3058, opened 2024-09-22
Monads are hard

ccasin:
#3037 introduced a monad implementation with a bug, this fixes it.  (Noticed this while writing tests for some other unboxed products stuff today.)

Two commits: the first adds a test showing how bad typing can result from the bug, and the second has the fix.  The repaired test has a very bad error message, but I don't have time today to dig into that.  Perhaps @goldfirere will have thoughts about improving these errors.

Review: @goldfirere


goldfirere (approved):
Yikes. Good catch.

The error message will be hard to improve without product histories, I think.

PR #3062, opened 2024-09-23
Register separate code fragments

gretay-js:
Add a special behavior for an experiment with prelinking to correctly handle closure marshalling. 
- Registering one big code fragment that covers all code segments in `caml_code_segments` is incorrect when the segments are not contiguous. Register each segment as a separate code fragment. 
- Add a global variable `caml_prelinking_in_use`, off by default. This PR does not affect normal compilation and execution.


stedolan (approved):
Looks good, this is an extremely unobtrusive change!

stedolan (approved):

PR #3065, opened 2024-09-23
Promote the `unique` extension to `Stable`

riaqn:
This PR promotes the `unique` extension to `Stable`, which means it will be enabled by default.

Please do not merge until we finish benchmarking.

### DIFF
--- a/testsuite/tests/typing-modal-kinds/basics.ml
+++ b/testsuite/tests/typing-modal-kinds/basics.ml
@@ -1,5 +1,4 @@
 (* TEST
- flags = "-extension unique";
  expect;
 *)
 
--- a/testsuite/tests/typing-modal-kinds/expected_mode.ml
+++ b/testsuite/tests/typing-modal-kinds/expected_mode.ml
@@ -1,5 +1,4 @@
 (* TEST
- flags = "-extension unique";
  expect;
 *)
 
--- a/testsuite/tests/typing-modes/class.ml
+++ b/testsuite/tests/typing-modes/class.ml
@@ -1,5 +1,4 @@
 (* TEST
-    flags += "-extension unique";
     expect;
 *)
 
--- a/testsuite/tests/typing-modes/currying.ml
+++ b/testsuite/tests/typing-modes/currying.ml
@@ -1,5 +1,4 @@
 (* TEST
- flags += "-extension unique";
  expect;
 *)
 
--- a/testsuite/tests/typing-modes/lazy.ml
+++ b/testsuite/tests/typing-modes/lazy.ml
@@ -1,5 +1,4 @@
 (* TEST
-    flags += "-extension unique";
     expect;
 *)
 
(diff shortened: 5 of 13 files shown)


ocaml/utils/language_extension.ml
goldfirere, line 1: Are there changes necessary in the testsuite? I presume all the existing tests that look at uniqueness turn on "alpha". I think those should be changed.

goldfirere, line 1: What's the documentation plan here? Once something is in `stable`, that should mean, I think, that we're expecting folks to use this in the tree. It will need documentation to do that. Maybe @anfelor should add that?

riaqn: Yeah Anton will develop some features based on uniqueness, upon which we should have some documentation for the users. Before that, tho, I think it's fine to not have documentation, because users won't encounter any type errors wrt uniqueness.

goldfirere: Either people will use the feature or they won't. If they do, they'll need documentation. If they don't, then there is no value added by merging this. To me, I think putting something in stable is a great time to require user-facing documentation.

riaqn: For now, people won't use this feature, therefore we don't need documentation. We still want to merge this PR to stress test the trivial code-path in the uniqueness analysis, which I think is valuable.

goldfirere: Yes, I agree that testing this is valuable, but we can do that without merging: just run nightly (as you suggest). I continue to think that we should expect all "stable" features to be documented.

riaqn: Sounds good to me. I converted this to draft PR.

riaqn: Sorry I want to backtrack - the internal benchmark shows that this doesn't cause failures or performance regression. I think while the benefit of merging this is little, the risk is also little. And I think constantly having to rebase this PR on top of `main` to monitor benchmark is consuming. I'd rather merging this now and forcing us to fix any potential issues.

riaqn: As discussed in other channels, UA will be `stable`, but writing `unique` modes will be `alpha`.

riaqn: Yeah, my impression is that we think @anfelor will push documentation to this PR, then everything can be marked as `stable`. Does that sound right, Anton?

anfelor: Created a new PR for this at #3180

anfelor:
Since opening the PR the following changes have happened:

 - #3180 provided documentation for the uniqueness mode
 - #3196 switched the printing of unique_ and once_ in types to always use the new syntax
 - #3198 added full support for inserting unique_barrier
 - #3160 slightly improved the performance of the mode solver. While this is still in review, I don't think it should block this PR. 

We have not yet benchmarked the impact of #3198. Should we do this?

Other than that, I would expect this PR to be good to merge.

riaqn:
#3198 involves a `zap_to_ceil` call for each record projection, but I wouldn't be worried.

I think it might be better to merge #3160 first?

PR #3066, opened 2024-09-23
Map unique_barrier information to Reads_vary

anfelor:
This PR maps the `unique_barrier` information we get from the uniqueness analysis to the `Reads_vary` field read semantics in Lambda. We might choose to map it to a new variant of field read semantics in the future and I created the `add_barrier_.*` functions in `lambda.ml` to make it easy for us to make that change.

In particular:
 - I added a `unique_barrier ref` to `Texp_field`, `Texp_setfield` and `pattern_data`.
 - I have checked all locations in `lambda/` where `P.*field`s are created to make sure they get the right `unique_barrier` information. The following functions create `P.*field`s with `Reads_agree` but are not touched by this PR:
   - `lambda.ml` `transl_address` : Only used to read module/object/extension/class paths.
   - `lambda.ml` `mod_field`: Only used for modules.
   - `translprim.ml`: `%field0_immut` and `%field1_immut` are mapped to a `Pfield` with `Reads_agree`. I am not entirely sure what this primitive does.
   - `matching.ml`: `combine_constructor` reads the constructor extension tag using `Reads_agree`. `for_optional_arg_default` reads optional arguments using `Reads_agree`. `call_switcher_variant_constr` reads the variant field with a unique barrier of `MayBePushedDown`.
 - In all of the places where I added `unique_barrier` information to `Pfield`, I have also changed the strictness of the binding from `Alias` to `StrictOpt`. That should protect us from the bug in `simplif.ml`.
 - I currently do not add `unique_barrier` information to `Psetfield`. As a result, the `unique_barrier ref` on `Texp_setfield` is currently ignored. The semantics of `Psetfield` might already make overwriting sound, but we should check how `Psetfield` interacts with the proposed `Preusefield` in the middle end. Depending on this discussion, we should either remove the `unique_barrier ref` from `Texp_setfield` or promote this information into lambda and beyond.


mshinwell:
`%field{0,1}_immut` are used to implement `fst` and `snd`, and are only used for immutable blocks.

goldfirere (commented):
Just looked at a few files, and hit a barrier at uniqueness_analysis.ml (I just don't know my way around that file). Given that @anfelor and I have a meeting later today, it seems best just to discuss then instead of doing more review now.

ocaml/typing/typedtree.ml
```diff
@@ -851,6 +858,7 @@ let as_computation_pattern (p : pattern) : computation general_pattern =
     pat_type = p.pat_type;
     pat_env = p.pat_env;
     pat_attributes = [];
+    pat_unique_barrier = fresh_unique_barrier ();
```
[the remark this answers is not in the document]
anfelor: Yes, that seems fine. We want to avoid unrelated patterns having the same barrier, which might lead us to be too conservative. But in this case the returned pattern should genuinely be a copy of the argument. I'll amend this.

ocaml/typing/typedtree.mli
```diff
@@ -356,7 +358,8 @@ and expression_desc =
       fields : ( Types.label_description * record_label_definition ) array;
       representation : Types.record_representation;
       extended_expression : expression option;
-      alloc_mode : alloc_mode option
+      alloc_mode : alloc_mode option;
```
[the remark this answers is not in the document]
anfelor: No, the `alloc_mode` is the mode of the current record, while the `unique_barrier` tracks the uniqueness of the record you are reading from if the `extended_expression` is a path. Furthermore, the `alloc_mode` seems to only use the linearity axis.

ocaml/typing/typedtree.mli
```diff
@@ -356,7 +358,8 @@ and expression_desc =
...
+      unique_barrier : unique_barrier ref
```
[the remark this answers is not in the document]
anfelor: Yes, I'll do that

ocaml/typing/typedtree.mli
```diff
@@ -97,6 +98,7 @@ and 'a pattern_data =
     pat_type: Types.type_expr;
     pat_env: Env.t;
     pat_attributes: attributes;
+    pat_unique_barrier : unique_barrier ref;
```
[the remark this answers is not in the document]
anfelor: Don't we also want to overwrite tuples and constructors eventually?

ocaml/lambda/matching.ml
goldfirere, line 1: I searched for more places we might need to use a unique barrier. I have suspicions about the following spots:

* `combine_constructors`
* `for_optional_arg_default`
* `call_switcher_variant_constr`

I think these should either be updated or get comments for why no change is necessary.

As I may have mentioned, the code here is hard-going. But a mistake could be pretty disastrous, so it's worth hunting. Happy to do this alongside you if you like.

ocaml/lambda/matching.ml
```diff
@@ -1799,10 +1799,12 @@ let get_expr_args_constr ~scopes head (arg, _mut, sort, layout) rem =
   Array.iter (fun jkind ->
       jkind_layout_default_to_value_and_check_not_void head.pat_loc jkind)
     cstr.cstr_arg_jkinds;
+  let ubr = Translmode.transl_unique_barrier (head.pat_unique_barrier) in
```
goldfirere, line 1802: If the unique barrier is stored in individual pattern descriptors instead of in the pattern heaer, this line gets harder. My guess is that, here, storing the unique barrier in the `Construct` constructor of `Paterns.Head.desc` may work (and similarly for other cases).

ocaml/lambda/translcore.ml
```diff
@@ -1792,7 +1793,9 @@ and transl_record ~scopes loc env mode fields repres opt_init_expr =
   let size = Array.length fields in
   (* Determine if there are "enough" fields (only relevant if this is a
      functional-style record update *)
-  let no_init = match opt_init_expr with None -> true | _ -> false in
+  let (no_init, ubr) = match opt_init_expr with
+    | None -> (true, MayBePushedDown)
```
goldfirere, line 1797: This is a bit scary. But it's OK: the `ubr` is used only in the `Kept` case which can happen only when this is `Some`. Still, I think I'd like the extraction of the `ubr` to happen closer to where it's used, so that there doesn't have to be a `MayBePushedDown` here. That means that the `None` case might have an `Misc.fatal_error` or something -- that feels safer than an under-motivated `MayBePushedDown`.

goldfirere (commented):
Finished review, at long last.

ocaml/lambda/translcore.ml
anfelor: I agree. I could turn the `if no_init || ..` statement into a case-split on `opt_init_expr`. This would eliminate the boolean-blindness of `no_init` and make the unique_barrier only available in the first branch where it is actually used.

ocaml/lambda/matching.ml
anfelor: Yes, we should go through this together. I think these are fine as remarked above, but we should gain more confidence.

anfelor:
Just pushed another commit to address the points raised above and in our meeting. In particular,

 - the new `Unique_barrier` now ensures that all unique_barriers that are read in `translcore` are actually handled by the uniqueness analysis first. I have also set us up to implement a pass that ensures that all unique barriers are read. The implementation is less nice now, though, since we require the `ref` again.
 - I have convinced myself that there is no bug if we perform ANF due to the algebra of the uniqueness analysis and added a few comments.

ocaml/typing/typedtree.ml
```diff
@@ -51,15 +51,42 @@ type _ pattern_category =
...
+    | _ -> assert false
 
   let resolve barrier =
-    Uniqueness.zap_to_ceil barrier
+    match !barrier with
+    | Enabled uniq ->
+      let zapped = Uniqueness.zap_to_ceil uniq in
+      barrier := Resolved zapped;
+      zapped
+    | Resolved barrier -> barrier
```
goldfirere, line 85: Does this case ever happen? If not, we don't need `Resolved`.

ocaml/typing/uniqueness_analysis.ml
```diff
@@ -995,10 +1012,13 @@ let conjuncts_pattern_match l =
 let rec pattern_match_tuple pat values =
   match pat.pat_desc with
   | Tpat_or (pat0, pat1, _) ->
+    Unique_barrier.enable pat.pat_unique_barrier;
```
goldfirere, line 1015: Could/should this step happen in `pattern_match`?

ocaml/typing/uniqueness_analysis.ml
```diff
@@ -1020,26 +1040,39 @@ let rec pattern_match_tuple pat values =
 and pattern_match_single pat paths : Ienv.Extension.t * UF.t =
   let loc = pat.pat_loc in
   let occ = Occurrence.mk loc in
+  (* To read from the allocation, we need to borrow its memory cell
```
goldfirere, line 1043: In the `` `NoRead `` case, the unique barrier will never be used again, right? Then we can immediately set it to `Resolved`, I think. (It's not that this is more efficient -- it's that it makes the code easier to understand.)

ocaml/typing/uniqueness_analysis.ml
```diff
@@ -1020,26 +1040,39 @@ let rec pattern_match_tuple pat values =
...
+     and set the unique_barrier. However, we do not read in every case,
+     since the user might want use a wildcard for already-consumed data. *)
+  let borrow_memory_address access =
+    Unique_barrier.enable pat.pat_unique_barrier;
+    match access with
+    | `NoRead -> UF.unused
```
goldfirere, line 1049: Is this the only use of `` `NoRead `` and `` `Read ``? If so, just make two functions: `no_borrow_memory_address` and `borrow_memory_address`, say.

ocaml/typing/uniqueness_analysis.ml
anfelor: No, because `pattern_match` only sees the topmost or-pattern, but `pat0` and `pat1` might be or-patterns themselves

ocaml/typing/typedtree.ml
anfelor: It can now happen after your suggestion below :) It also enables us to write the check that you proposed, where we traverse all unique barriers after generating lambda to check that all were resolved.

goldfirere: Ah, yes, that makes sense. OK.

ocaml/typing/uniqueness_analysis.ml
```diff
@@ -1043,36 +1043,38 @@ and pattern_match_single pat paths : Ienv.Extension.t * UF.t =
...
-  let borrow_memory_address access =
+  let no_borrow_memory_address () =
     Unique_barrier.enable pat.pat_unique_barrier;
-    match access with
-    | `NoRead -> UF.unused
-    | `Read ->
-      Paths.mark_implicit_borrow_memory_address occ
-        (Read pat.pat_unique_barrier) paths
+    let _ = Unique_barrier.resolve pat.pat_unique_barrier in
+    ()
```
goldfirere, line 1049:
```suggestion
    ignore (Unique_barrier.resolve pat.pat_unique_barrier)
```

ocaml/typing/uniqueness_analysis.ml
```diff
@@ -1043,36 +1043,38 @@ and pattern_match_single pat paths : Ienv.Extension.t * UF.t =
...
+  in
+  let borrow_memory_address () =
+    Unique_barrier.enable pat.pat_unique_barrier;
+    Paths.mark_implicit_borrow_memory_address occ (Read pat.pat_unique_barrier)
+      paths
   in
   match pat.pat_desc with
   | Tpat_or (pat0, pat1, _) ->
-    let uf_read = borrow_memory_address `NoRead in
+    let () = no_borrow_memory_address () in
```
goldfirere, line 1058:
```suggestion
    no_borrow_memory_address ();
```

anfelor:
Just pushed a few tests, to be extended once we have proper overwriting. The change in the first two expect-tests compared to a branch without the change in this PR is:

```
>  let proj_unique r =
>    let y = r.y in
>    let r = unique_use r in
>    (r, y)
>  [%%expect{|
>  (let
>    (unique_use/283 = (apply (field_imm 0 (global Toploop!)) "unique_use")
>     proj_unique/291 =
>       (function {nlocal = 0} r/293[(consts ()) (non_consts ([0: *, *]))]
>         [(consts ())
>          (non_consts ([0: [(consts ()) (non_consts ([0: *, *]))], *]))]
>         (let
> -         (y/294 = (field_imm 1 r/293)
> +         (y/294 = (field_mut 1 r/293)
>            r/295 =[(consts ()) (non_consts ([0: *, *]))]
>              (apply unique_use/283 r/293))
>           (makeblock 0 ([(consts ()) (non_consts ([0: *, *]))],*) r/295 y/294))))
>    (apply (field_imm 1 (global Toploop!)) "proj_unique" proj_unique/291))
>  val proj_unique : unique_ record -> record * string = <fun>
>  |}]
>  
>  let match_unique r =
>    match r with
>    | { y } ->
>      let r = unique_use r in
>      (r, y)
>  [%%expect{|
>  (let
>    (unique_use/283 = (apply (field_imm 0 (global Toploop!)) "unique_use")
>     proj_unique/302 =
>       (function {nlocal = 0} r/303[(consts ()) (non_consts ([0: *, *]))]
>         [(consts ())
>          (non_consts ([0: [(consts ()) (non_consts ([0: *, *]))], *]))]
>         (let
> !         (y/304 =o (field_mut 1 r/303)
> !          r/305 =[(consts ()) (non_consts ([0: *, *]))]
> -            (apply unique_use/283 r/303))
> -         (makeblock 0 ([(consts ()) (non_consts ([0: *, *]))],*) r/305
> -           (field_imm 1 r/303)))))
> +            (apply unique_use/283 r/303))
> +         (makeblock 0 ([(consts ()) (non_consts ([0: *, *]))],*) r/305 y/304))))
>    (apply (field_imm 1 (global Toploop!)) "proj_unique" proj_unique/302))
>  val proj_unique : unique_ record -> record * string = <fun>
>  |}]
```

As you can see, the field reads become `field_mut` instead of `field_imm` and we no longer push down the projection in the second example.

ocaml/typing/typedtree.mli
```diff
@@ -96,6 +109,7 @@ and 'a pattern_data =
     pat_type: Types.type_expr;
     pat_env: Env.t;
     pat_attributes: attributes;
+    pat_unique_barrier : Unique_barrier.t;
```
goldfirere, line 112: It took us some work to discover what the semantics of this field is. Maybe capture this in a comment? Would it be right to say

```
(* This tracks whether the scrutinee of the pattern is used uniquely within the body of the pattern match. *)
```

?

goldfirere (approved):

PR #3067, opened 2024-09-24
Use shifts instead of masking to strip header reserved bits

stedolan:
In #3042, @ahh reports poor code for Array.length due to the masking needed to remove header words. For instance, `let f x = Array.length x` is currently compiled as:
```
	movabsq	$72057594037927935, %rbx
	movq	-8(%rax), %rax
	andq	%rbx, %rax
	shrq	$9, %rax
	orq	$1, %rax
	ret
```
Instead of masking, we can use a pair of left and right shifts to extract the relevant bitfield. This generates better code, because the shifts combine well with subsequent shifts, and don't require an extra register (which masking does, because the constant is so large):
```
	movq	-8(%rax), %rax
	salq	$8, %rax
	shrq	$17, %rax
	orq	$1, %rax
	ret
```


ncik-roberts:
I rebased this over #3070; @stedolan could you see if the conflicts are resolved to your satisfaction?

stedolan:
> @stedolan could you see if the conflicts are resolved to your satisfaction?

They are, thanks!

PR #3068, opened 2024-09-24
Flambda 2 support for arrays containing unboxed products

mshinwell:
This provides Flambda 2 support for arrays containing unboxed products (although no support in the type system, so no static allocation for immutable arrays).

Based on #3059.

### DIFF
--- a/backend/cmm_helpers.ml
+++ b/backend/cmm_helpers.ml
@@ -1161,6 +1161,33 @@ let unboxed_mutable_float_array_ref arr ofs dbg =
 let unboxed_immutable_float_array_ref arr ofs dbg =
   Cop (mk_load_immut Double, [array_indexing log2_size_float arr ofs dbg], dbg)
 
+let unboxed_mutable_float32_unboxed_product_array_ref arr ~array_index dbg =
+  bind "arr" arr (fun arr ->
+      bind "index" array_index (fun index ->
+          Cop
+            ( mk_load_mut (Single { reg = Float32 }),
+              [array_indexing log2_size_addr arr index dbg],
+              dbg )))
+
+(* CR mshinwell/mslater: if we're writing zeros to the top 32 bits of float32
+   product fields, should we do the same for mixed block record fields? *)
+
+let unboxed_mutable_float32_unboxed_product_array_set arr ~array_index
+    ~new_value dbg =
+  bind "arr" arr (fun arr ->
+      bind "index" array_index (fun index ->
+          bind "new_value" new_value (fun new_value ->
+              Csequence
+                ( Cop
+                    ( Cstore (Word_int, Assignment),
+                      [ array_indexing log2_size_addr arr index dbg;
+                        Cconst_int (0, dbg) ],
+                      dbg ),
+                  Cop
+                    ( Cstore (Single { reg = Float32 }, Assignment),
+                      [array_indexing log2_size_addr arr index dbg; new_value],
+                      dbg ) ))))
+
 let unboxed_float_array_ref (mutability : Asttypes.mutable_flag) ~block:arr
     ~index:ofs dbg =
   match mutability with
@@ -1286,18 +1313,41 @@ let unboxed_int32_array_ref =
      [Thirtytwo_signed] load. *)
   unboxed_packed_array_ref ~memory_chunk:Thirtytwo_signed ~elements_per_word:2
 
+let unboxed_mutable_int32_unboxed_product_array_ref arr ~array_index dbg =
+  bind "arr" arr (fun arr ->
+      bind "index" array_index (fun index ->
+          sign_extend_32 dbg
+            (Cop
+               ( mk_load_mut Thirtytwo_signed,
+                 [array_indexing log2_size_addr arr index dbg],
+                 dbg ))))
+
+let unboxed_mutable_int32_unboxed_product_array_set arr ~array_index ~new_value
+    dbg =
+  bind "arr" arr (fun arr ->
+      bind "index" array_index (fun index ->
+          bind "new_value" new_value (fun new_value ->
+              let new_value = sign_extend_32 dbg new_value in
+              Cop
+                ( Cstore (Word_int, Assignment),
+                  [array_indexing log2_size_addr arr index dbg; new_value],
+                  dbg ))))
+
 let unboxed_float32_array_ref =
   unboxed_packed_array_ref
     ~memory_chunk:(Single { reg = Float32 })
     ~elements_per_word:2
 
-let unboxed_int64_or_nativeint_array_ref arr index dbg =
+let unboxed_int64_or_nativeint_array_ref ~has_custom_ops arr ~array_index dbg =
   bind "arr" arr (fun arr ->
-      bind "index" index (fun index ->
+      bind "index" array_index (fun index ->
           let index =
-            (* Need to skip the custom_operations field. 2 not 1 since we are
-               manipulating a tagged int. *)
-            add_int index (int ~dbg 2) dbg
+            if has_custom_ops
+            then
+              (* Need to skip the custom_operations field. 2 not 1 since we are
+                 manipulating a tagged int. *)
+              add_int index (int ~dbg 2) dbg
+            else index
           in
           int_array_ref arr index dbg))
 
@@ -1324,13 +1374,17 @@ let unboxed_float32_array_set =
     ~memory_chunk:(Single { reg = Float32 })
     ~elements_per_word:2
 
-let unboxed_int64_or_nativeint_array_set arr ~index ~new_value dbg =
+let unboxed_int64_or_nativeint_array_set ~has_custom_ops arr ~index ~new_value
+    dbg =
   bind "arr" arr (fun arr ->
       bind "index" index (fun index ->
           bind "new_value" new_value (fun new_value ->
               let index =
-                (* See comment in [unboxed_int64_or_nativeint_array_ref]. *)
-                add_int index (int ~dbg 2) dbg
+                if has_custom_ops
+                then
+                  (* See comment in [unboxed_int64_or_nativeint_array_ref]. *)
+                  add_int index (int ~dbg 2) dbg
+                else index
               in
               int_array_set arr index new_value dbg)))
 
--- a/backend/cmm_helpers.mli
+++ b/backend/cmm_helpers.mli
@@ -1034,6 +1034,9 @@ val reperform :
   last_fiber:expression ->
   expression
 
+(* CR mshinwell: change unboxed scalar arrays to use mixed block (or similar)
+   representations rather than custom blocks *)
+
 (** Allocate a block to hold an unboxed float32 array for the given number of
     elements. *)
 val allocate_unboxed_float32_array :
@@ -1076,14 +1079,82 @@ val unboxed_vec128_array_length : expression -> Debuginfo.t -> expression
 val unboxed_float32_array_ref :
   expression -> expression -> Debuginfo.t -> expression
 
+(** Read an unboxed float32 from a 64-bit field in an array represented as
+    a mixed block (with tag zero), as used for unboxed product arrays.
+
+    The float32 is expected to be in the least significant bits of the
+    64-bit field.  The most significant 32 bits of such field are ignored.
+
+    The zero-indexed element number is specified as a tagged immediate.
+*)
+val unboxed_mutable_float32_unboxed_product_array_ref :
+  expression -> array_index:expression -> Debuginfo.t -> expression
+
+(* CR mshinwell/mslater: We could do movss xmm xmm, movsd mem xmm instead of
+   separate writes *)
+
+(** Write an unboxed float32 into a 64-bit field in an array represented as
+    a mixed block (with tag zero), as used for unboxed product arrays.
+
+    The zero-indexed element number is specified as a tagged immediate.
+
+    The float32 will be written to the least significant bits of the
+    64-bit field.  The top 32 bits of the written word will be initialized
+    to zero.  Note that two writes are involved.
+*)
+val unboxed_mutable_float32_unboxed_product_array_set :
+  expression ->
+  array_index:expression ->
+  new_value:expression ->
+  Debuginfo.t ->
+  expression
+
 (** Read from an unboxed int32 array (without bounds check). *)
 val unboxed_int32_array_ref :
   expression -> expression -> Debuginfo.t -> expression
 
+(** Read an unboxed int32 from (the least significant bits of) a 64-bit field
+    in an array represented as a mixed block (with tag zero), as used for
+    unboxed product arrays.
+
+    The zero-indexed element number is specified as a tagged immediate.
+
+    The returned value is always sign extended, but it is not assumed that
+    the 64-bit field in the array contains a sign-extended representation.
+*)
+val unboxed_mutable_int32_unboxed_product_array_ref :
+  expression -> array_index:expression -> Debuginfo.t -> expression
+
+(** Write an unboxed int32 into a 64-bit field in an array represented as
+    a mixed block (with tag zero), as used for unboxed product arrays.
+
+    The zero-indexed element number is specified as a tagged immediate.
+
+    The write is done as a 64-bit write of a sign-extended version of the
+    supplied [new_value].
+*)
+val unboxed_mutable_int32_unboxed_product_array_set :
+  expression ->
+  array_index:expression ->
+  new_value:expression ->
+  Debuginfo.t ->
+  expression
+
 (** Read from an unboxed int64 or unboxed nativeint array (without bounds
-    check). *)
+    check).
+
+    The [has_custom_ops] parameter should be set to [true] unless the array
+    in question is an unboxed product array: these are represented as mixed
+    blocks, not custom blocks.
+
+    The zero-indexed element number is specified as a tagged immediate.
+*)
 val unboxed_int64_or_nativeint_array_ref :
-  expression -> expression -> Debuginfo.t -> expression
+  has_custom_ops:bool ->
+  expression ->
+  array_index:expression ->
+  Debuginfo.t ->
+  expression
 
 (** Update an unboxed float32 array (without bounds check). *)
 val unboxed_float32_array_set :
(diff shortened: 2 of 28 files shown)


mshinwell:
This is now based on #3072.

mshinwell:
(I still need to add the code for duplicate array operations.)

middle_end/flambda2/to_cmm/to_cmm_primitive.ml
```diff
@@ -184,14 +184,26 @@ let make_array ~dbg kind alloc_mode args =
   | Naked_int64s -> C.allocate_unboxed_int64_array ~elements:args mode dbg
   | Naked_nativeints ->
     C.allocate_unboxed_nativeint_array ~elements:args mode dbg
+  | Unboxed_product _ ->
+    if P.Array_kind.must_be_gc_scannable kind
+    then C.make_alloc ~mode dbg ~tag:0 args
+    else
+      let mem_chunks =
+        P.Array_kind.element_kinds kind |> List.map C.memory_chunk_of_kind
```
Gbury, line 192: I don't understand this part of the code: shouldn't the list of elements kinds (which is iiuc the list of kinds of the unarized version of **each** element) somehow be duplicated by the number of products in the array being created ?

mshinwell: Ah yes, this is wrong - fixing.

mshinwell:
This is now ready for an initial pass of review.  It only contains the basic code for handling `Pgcscannableproductarray` and `Pgcignorableproductarray`.  It does not contain the code for reinterpret operations, the dynamic make-array primitive (previously known as `%make_unboxed_tuple_vect`), or the measuring of layouts.  It also does not contain the Flambda 2 type system support for unboxed product arrays.  All of these will be provided in follow-up PRs.

@TheNumbat there are a couple of `XXX`-marked questions for you.

I've moved the dev branch that contains all of the aforementioned functionality to `mshinwell/tuple-arrays-middle-end-dev`.

middle_end/flambda2/terms/flambda_primitive.ml
```diff
@@ -130,22 +135,30 @@ module Array_kind = struct
...
+         scannable and that cannot be scanned:@ %a"
+        print t
 
   let has_custom_ops t =
     match t with
     | Immediates | Values | Naked_floats -> false
     | Naked_float32s | Naked_int32s | Naked_int64s | Naked_nativeints
     | Naked_vec128s ->
       true
+    | Unboxed_product _ -> assert false (* XXX where is this function used? *)
```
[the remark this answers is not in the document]
mshinwell: Have already fixed this one.

backend/cmm_helpers.ml
```diff
@@ -1161,19 +1161,19 @@ let unboxed_mutable_float_array_ref arr ofs dbg =
 let unboxed_immutable_float_array_ref arr ofs dbg =
   Cop (mk_load_immut Double, [array_indexing log2_size_float arr ofs dbg], dbg)
 
-let unboxed_mutable_float32_unboxed_product_array_ref arr index dbg =
+let unboxed_mutable_float32_unboxed_product_array_ref arr ~element_num dbg =
```
Gbury, line 1164: as discussed on slack: maybe `array_index` might be a slightly better label

backend/cmm_helpers.mli
```diff
@@ -1079,15 +1079,29 @@ val unboxed_float32_array_ref :
...
 
 (** Read an unboxed float32 from a 64-bit field in an array represented as
-    a mixed block, as used for unboxed product arrays. *)
+    a mixed block (with tag zero), as used for unboxed product arrays.
+
+    The float32 is expected to be in the least significant bits of the
+    64-bit field.
+
+    The zero-indexed element number is specified as a tagged immediate.
+    This always coincides with the index in words.
```
Gbury, line 1088: I'm not sure we need this last sentence.

mshinwell:
@TheNumbat maybe you could look at the proposed sign extension semantics too?  (see `cmm_helpers.mli`)

backend/cmm_helpers.mli
```diff
@@ -1079,15 +1079,29 @@ val unboxed_float32_array_ref :
...
 
 (** Write an unboxed float32 into a 64-bit field in an array represented as
-    a mixed block, as used for unboxed product arrays. *)
+    a mixed block (with tag zero), as used for unboxed product arrays.
+
+    The zero-indexed element number is specified as a tagged immediate.
+    This always coincides with the index in words.
+
+    The float32 will be written to the least significant bits of the
+    64-bit field.  The top 32 bits of the written word are undefined.
```
Gbury, line 1100: Just a thought: couldn't this behaviour (leaving the top bits undefined) be annoying/problematic in certain situations ? I'm thinking if such an array ever gets stored inside a cmx file for instance (which I'd find reasonable to store some constants of that type), then cmx files that should be equal may actually be different (and have different hashes/digests), which may confuse build systems ?

mshinwell:
I should probably have written the following here as well: scannable unboxed product arrays have tag 0 and look like normal arrays.  Non-scannable ones have tag 0 and use the mixed block header bits in order to prevent scanning.  So there are no custom blocks involved.  There is no packing: an array of `#(int32# * int32#)` has the 32-bit integers aligned on the 64-bit word boundaries.

backend/cmm_helpers.mli
```diff
@@ -1096,17 +1110,33 @@ val unboxed_mutable_float32_unboxed_product_array_set :
 val unboxed_int32_array_ref :
   expression -> expression -> Debuginfo.t -> expression
 
+(* XXX mshinwell: are the sign extension semantics for the following two
+   functions reasonable? *)
```
Gbury, line 1114: I think your current specification is reasonable, but I'd argue that it would be nice to have the property that the int32s are stored sign-extended. This links to my comment above about marshalling of arrays of unboxed float32s: I think it's better to have and enforce a canonical representation in those cases (as long as it's not too costly), to avoid surprising behaviours.

backend/cmm_helpers.mli
```diff
@@ -1117,9 +1147,16 @@ val unboxed_mutable_int32_unboxed_product_array_set :
     The [has_custom_ops] parameter should be set to [true] unless the array
     in question is an unboxed product array: these are represented as mixed
     blocks, not custom blocks.
+
+    The zero-indexed element number is specified as a tagged immediate.
+    This coincides with the index in words iff [has_custom_ops] is [false].
```
Gbury, line 1152: I don't think this last sentence is necessary, or if you want to keep it, then you should also specify what happens when `[has_custom_ops] is [true]`.

Gbury:
@mshinwell does using mixed blocks for arrays of unboxed products that are not scannable means that these arrays have a different maximum length than `Sys.max_array_length` ? Additionally, that maximum length will then be counted in the number of unarized elements from what I understand. Is that documented somewhere ? And will those corner/failure cases raise the same exception as the regular array cases ?

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
```diff
@@ -887,27 +1015,87 @@ let array_access_validity_condition array array_kind index
...
+    | Array_kind Values
+    | Array_kind Naked_floats
+    | Array_kind Naked_float32s
+    | Array_kind Naked_int32s
+    | Array_kind Naked_int64s
+    | Array_kind Naked_nativeints
+    | Array_kind Naked_vec128s
+    | Float_array_opt_dynamic ->
+      1
+    | Array_kind (Unboxed_product kinds) -> List.length kinds
```
Ekdohibs, line 1030: I think this should be the sum of the widths for `kinds`, since they themselves can contain unboxed products.

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
```diff
@@ -1406,7 +1644,14 @@ let convert_lprim ~big_endian (prim : L.primitive) (args : Simple.t list list)
         ( Immediates | Values | Naked_floats | Naked_float32s | Naked_int32s
         | Naked_int64s | Naked_nativeints | Naked_vec128s )
     | Float_array_opt_dynamic ->
-      [prim])
+      [prim]
+    | Array_kind (Unboxed_product kinds) ->
+      (* [Array_length] returns the unarized length (see
+         flambda_primitive.mli). *)
```
Ekdohibs, line 1650: It seems to me this comment says the opposite of the code?

backend/cmm_helpers.ml
```diff
@@ -1161,6 +1161,26 @@ let unboxed_mutable_float_array_ref arr ofs dbg =
 let unboxed_immutable_float_array_ref arr ofs dbg =
   Cop (mk_load_immut Double, [array_indexing log2_size_float arr ofs dbg], dbg)
 
+let unboxed_mutable_float32_unboxed_product_array_ref arr ~element_num dbg =
+  bind "arr" arr (fun arr ->
+      bind "index" element_num (fun index ->
+          Cop
+            (* XXX mshinwell for mslater: is Float32 correct? *)
```
TheNumbat, line 1168: Yes, `Float64` would convert to f64 on load

backend/cmm_helpers.ml
```diff
@@ -1161,6 +1161,26 @@ let unboxed_mutable_float_array_ref arr ofs dbg =
...
+              [array_indexing log2_size_float arr index dbg],
+              dbg )))
+
+let unboxed_mutable_float32_unboxed_product_array_set arr ~element_num
+    ~new_value dbg =
+  bind "arr" arr (fun arr ->
+      bind "index" element_num (fun index ->
+          bind "new_value" new_value (fun new_value ->
+              Cop
+                (* XXX mshinwell for mslater: is Float32 correct? *)
```
TheNumbat, line 1179: (same)

backend/cmm_helpers.ml
```diff
@@ -1161,6 +1161,26 @@ let unboxed_mutable_float_array_ref arr ofs dbg =
 let unboxed_immutable_float_array_ref arr ofs dbg =
   Cop (mk_load_immut Double, [array_indexing log2_size_float arr ofs dbg], dbg)
 
+let unboxed_mutable_float32_unboxed_product_array_ref arr ~element_num dbg =
+  bind "arr" arr (fun arr ->
+      bind "index" element_num (fun index ->
+          Cop
+            (* XXX mshinwell for mslater: is Float32 correct? *)
+            ( mk_load_mut (Single { reg = Float32 }),
+              [array_indexing log2_size_float arr index dbg],
```
TheNumbat, line 1170: Is there a reason this uses `log2_size_float` instead of `log2_size_addr`?

backend/cmm_helpers.ml
```diff
@@ -1161,6 +1161,26 @@ let unboxed_mutable_float_array_ref arr ofs dbg =
...
+
+let unboxed_mutable_float32_unboxed_product_array_set arr ~element_num
+    ~new_value dbg =
+  bind "arr" arr (fun arr ->
+      bind "index" element_num (fun index ->
+          bind "new_value" new_value (fun new_value ->
+              Cop
+                (* XXX mshinwell for mslater: is Float32 correct? *)
+                ( Cstore (Single { reg = Float32 }, Assignment),
+                  [array_indexing log2_size_float arr index dbg; new_value],
```
TheNumbat, line 1181: (same)

backend/cmm_helpers.mli
TheNumbat: I think this is okay because static constants will pad the extra space with zeros - or at least that's what float32 scalar constants are doing already.

backend/cmm_helpers.mli
TheNumbat: I agree given that the store functions always write a sign-extended 64 bit word. But having the same semantics as float32 (undefined high bits) might make it easier to transition to tightly packing 32-bit fields in the future.

backend/cmm_helpers.mli
TheNumbat: Maybe also leave a CR for swapping arrays of unboxed scalars to the mixed block representation

ocaml/lambda/lambda.ml
```diff
@@ -420,6 +425,20 @@ and array_set_kind =
...
   | Punboxedvectorarray_set of unboxed_vector
+  | Pgcscannableproductarray_set of
+      modify_mode * scannable_product_element_kind list
+  | Pgcignorableproductarray_set of ignorable_product_element_kind list
+
+and ignorable_product_element_kind =
+  | Pint_ignorable
+  | Punboxedfloat_ignorable of unboxed_float
+  | Punboxedint_ignorable of unboxed_integer
+  | Pproduct_ignorable of ignorable_product_element_kind list
```
TheNumbat, line 436: Looks like this needs `Punboxedvector_ignorable`

middle_end/flambda2/terms/flambda_primitive.mli
```diff
@@ -368,6 +366,9 @@ type unary_primitive =
   | Is_int of { variant_only : bool }
   | Get_tag
   | Array_length of Array_kind_for_length.t
+      (** The unarized length of an array.  So for an example an array of
+          kind [Unboxed_product [tagged_immediate; tagged_immediate]] always
+          has a length that is a multiple of two. *)
```
TheNumbat, line 371: Does this apply to products containing unboxed vectors?

middle_end/flambda2/to_cmm/to_cmm_primitive.ml
```diff
@@ -297,6 +349,12 @@ let array_load ~dbg (array_kind : P.Array_kind.t)
       Debuginfo.print_compact dbg
   | Values, Naked_vec128s ->
     Misc.fatal_error "Attempted to load a SIMD vector from a value array."
+  | Unboxed_product _, Naked_vec128s ->
+    (* CR mshinwell: should this be supported? *)
```
TheNumbat, line 353: Seems fine to support, just need to make sure the element num computation accounts for vectors taking up >1 word

backend/cmm_helpers.mli
Gbury: I don't think that actually solves the case I was thinking about: in that situation, let's assume we are compiling some `foo.ml` file, and in that file there is an array of unboxed  `float32` that can be lifted. When compiling `foo.ml`, the compiler will lift the array so that it will be a static constant of `Foo`, but during compilation, it is not a static constant, and is likely to be represented as a non-static array of `float32` (created dynamically from the AST of the source expression). That non-static array will then be stored in the cmx files, as part of an flambda2 static let binding, and so it will be marshalled when creating the cmx files.

In that situation, the cmx file will contain the marshalled version of a non-static array of unboxed `float32`. Since the higher bits are left uninitialized, that means that the cms file would in essence be non-deterministic: re-compiling the same file without any changes would yield different cmx files, since the uninitialized bits would likely change beetween invocations of the compiler.

More generally, any other program that tries and marshall dynamic arrays for unboxed `float32` would have very surprising behaviours (the compiler with cmx files being just one instance), and I'm not sure that's a nice situation to have.

TheNumbat: Ah, I see. Currently marshalling arrays of unboxed types is not supported, though - do we plan to add it?

Gbury: As long as we don't marshall these, we should be safe indeed. However, with the reinterpret casts on read (and maybe writes ?) to arrays, there will be situations where these uninitialized bits end up being read (and maybe written to ?) though it would be a programming error to write such code.

backend/cmm_helpers.ml
mshinwell: Yeah I guess this might be clearer with log2_size_addr.

backend/cmm_helpers.ml
mshinwell: Renamed

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
mshinwell: Fixed, and code moved to Flambda_primitive.Array_kind and .Array_kind_for_length

ocaml/lambda/lambda.ml
mshinwell: These aren't going to be supported in the first version.  The representation of the indexing in Flambda 2 needs thinking about carefully.

backend/cmm_helpers.mli
mshinwell: I've decided to make this write function zero the whole word, then write the 32-bit float.  This seems probably more consistent with the behaviour in the int32 case, where the top half has a defined semantics.  The read functions in both cases ignore the top half, though.

mshinwell:
> @mshinwell does using mixed blocks for arrays of unboxed products that are not scannable means that these arrays have a different maximum length than `Sys.max_array_length` ? Additionally, that maximum length will then be counted in the number of unarized elements from what I understand. Is that documented somewhere ? And will those corner/failure cases raise the same exception as the regular array cases ?

I think this only matters for the new array creation function for arrays of unboxed products.  It probably is indeed true that the bounds are slightly different, so I'll just make it take the minimum.  The exception raised will be the same.

backend/cmm_helpers.mli
```diff
@@ -1078,14 +1081,79 @@ val unboxed_vec128_array_length : expression -> Debuginfo.t -> expression
...
+  expression -> array_index:expression -> Debuginfo.t -> expression
+
+(** Write an unboxed float32 into a 64-bit field in an array represented as
+    a mixed block (with tag zero), as used for unboxed product arrays.
+
+    The zero-indexed element number is specified as a tagged immediate.
+
+    The float32 will be written to the least significant bits of the
+    64-bit field.  The top 32 bits of the written word will be initialized
+    to zero.  Note that two writes are involved.
```
TheNumbat, line 1102: We could do `movss xmm xmm`, `movsd mem xmm` instead of separate writes

TheNumbat (commented):
lgtm, but we should test that unboxed products containing vec128s are rejected with a reasonable error message. Also, if we're writing zeros to the top 32 bits of float32 product fields, should we do the same for mixed block record fields?

middle_end/flambda2/to_cmm/to_cmm_primitive.ml
```diff
@@ -194,6 +194,37 @@ let block_set ~dbg (kind : P.Block_access_kind.t) (init : P.Init_or_assign.t)
...
+      P.Array_kind.print kind;
+  let mem_chunks_per_non_unarized_element =
+    Array.of_list mem_chunks_per_non_unarized_element
+  in
+  let mem_chunks =
+    List.mapi
+      (fun i _arg ->
+        let index = i mod num_mem_chunks_per_non_unarized_element in
+        mem_chunks_per_non_unarized_element.(index))
+      args
```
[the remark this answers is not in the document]
mshinwell: Deferring this.

backend/cmm_helpers.mli
mshinwell: Moved to a CR

mshinwell:
@TheNumbat I think the error message is a frontend thing, no?  cc @ccasin

Ekdohibs (approved):

PR #3069, opened 2024-09-24
Refactor multi-word array bounds check code

mshinwell:
This is a small refactoring-only PR which is needed for #3068 .  That PR will make use of the existing logic used for SIMD bounds checks, made more generic in this PR.

(cherry picked from commit e73bed11f0f6ab7fe28ed9f43cda86e344d6dc62)

### DIFF
--- a/middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
+++ b/middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
@@ -620,20 +620,12 @@ let bytes_like_set ~dbg ~unsafe
     check_access ~dbg ~size_int ~access_size ~primitive:unsafe_set bytes
       ~index_kind index
 
-(* Array vector load/store *)
+(* Array bounds checks *)
 
-let array_vector_access_validity_condition array ~size_int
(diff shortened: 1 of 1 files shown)


TheNumbat (approved):
Refactor looks good, but I'm a bit concerned about codegen (see godbolt link)

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
```diff
@@ -620,20 +620,12 @@ let bytes_like_set ~dbg ~unsafe
...
-      Misc.fatal_error
-        "Attempted to load/store a SIMD vector from/to a value array."
-  in
+let multiple_word_array_access_validity_condition array ~size_int
+    array_length_kind index_kind ~width_in_scalars ~index =
   let length_untagged =
-    untag_int (H.Prim (Unary (Array_length (Array_kind array_kind), array)))
+    untag_int (H.Prim (Unary (Array_length array_length_kind, array)))
   in
   let reduced_length_untagged =
```
TheNumbat, line 630: We should special-case `width_in_scalars=1` to be sure normal array accesses won't generate unecessary math here

PR #3071, opened 2024-09-24
Add Sequence and Unboxed_product to primitive helper IR

mshinwell:
#3068 generates unboxed products inside the intermediate primitive helper IR terms in places where they couldn't exist before.  In addition it needs a sequencing construct.  To do that, it uses the support provided in this PR, which adds to `Lambda_to_flambda_primitive_helpers.expr_primitive` the following:
```diff
+|  | Sequence of expr_primitive list
+|  | Unboxed_product of expr_primitive list
 |
 |and simple_or_prim =
 |  | Simple of Simple.t
 |  | Prim of expr_primitive
 |
+|val maybe_create_unboxed_product : expr_primitive list -> expr_primitive
```
One consequence is that the if-then-else construct needs to take a result kind list.

I think this actually overall simplifies matters, as can be seen from the removal of the original `bind_rec` wrapper function.

### DIFF
--- a/middle_end/flambda2/from_lambda/lambda_to_flambda_primitives_helpers.ml
+++ b/middle_end/flambda2/from_lambda/lambda_to_flambda_primitives_helpers.ml
@@ -350,16 +350,12 @@ let rec bind_recs acc exn_cont ~register_const0 (prim : expr_primitive)
       in
       let result_pats =
         List.map
-          (fun result_var ->
-            Bound_var.create result_var Name_mode.normal)
+          (fun result_var -> Bound_var.create result_var Name_mode.normal)
           result_vars
       in
-      let result_simples =
-        List.map Simple.var result_vars
-      in
+      let result_simples = List.map Simple.var result_vars in
       let acc, apply_cont =
-        Apply_cont_with_acc.create acc join_point_cont
-          ~args:result_simples ~dbg
+        Apply_cont_with_acc.create acc join_point_cont ~args:result_simples ~dbg
       in
       let acc, body = Expr_with_acc.create_apply_cont acc apply_cont in
       List.fold_left2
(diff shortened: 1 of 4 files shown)


middle_end/flambda2/from_lambda/lambda_to_flambda_primitives_helpers.ml
```diff
@@ -157,64 +186,86 @@ let expression_for_failure acc exn_cont ~register_const0 primitive dbg
...
     in
     let rec build_cont acc args_to_convert converted_args =
       match args_to_convert with
       | [] -> cont acc converted_args
       | arg :: args_to_convert ->
-        let cont acc arg =
-          build_cont acc args_to_convert (arg :: converted_args)
+        let cont acc args =
+          (* XXX check order of the appending here *)
+          build_cont acc args_to_convert (args @ converted_args)
```
Ekdohibs, line 260: Since we do a `List.rev` _before_ and not _after_, a normal append seems good.

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives_helpers.ml
```diff
@@ -286,32 +342,58 @@ let rec bind_rec acc exn_cont ~register_const0 (prim : expr_primitive)
...
-        (Bound_pattern.singleton ifnot_result_pat)
-        ifnot ~body
+      List.fold_left2
+        (fun (acc, body) ifnot_result_pat ifnot ->
+          Let_with_acc.create acc
+            (Bound_pattern.singleton ifnot_result_pat)
+            ifnot ~body)
+        (acc, body)
+        (List.rev ifnot_result_pats)
+        (List.rev ifnot)
```
Ekdohibs, line 396: All this code is duplicated in the ifso/ifnot cases, could we share the code a bit to simplify it?

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives_helpers.ml
```diff
@@ -324,29 +406,45 @@ let rec bind_rec acc exn_cont ~register_const0 (prim : expr_primitive)
...
+      (cont acc [Named.create_simple Simple.const_unit])
+      (List.rev expr_primitives)
+  | Unboxed_product (expr_primitive :: expr_primitives) ->
+    bind_recs acc exn_cont ~register_const0 expr_primitive dbg
+      (fun acc nameds1 ->
+        bind_recs acc exn_cont ~register_const0
+          (Unboxed_product expr_primitives) dbg (fun acc nameds2 ->
+            cont acc (nameds1 @ nameds2)))
+  | Unboxed_product [] ->
+    Misc.fatal_error "Nullary unboxed products are not permitted here"
```
Ekdohibs, line 433: Couldn't we just do `cont acc []` here?

Ekdohibs (approved):

PR #3072, opened 2024-09-24
Refactoring for unboxed products in arrays

mshinwell:
This elaborate patch should be a no-op.

It refactors various of the array-related Flambda 2 primitive types so that the support for unboxed products in arrays will merge in easily.  There is now an `Array_load_kind` module to go along with the existing `Array_set_kind` and the slightly odd "overlay" of the `array_accessor_width` type (e.g. describing a 128-bit access to an array of 64-bit integers) has been folded into this new module.  `Array_load` and `Array_set` now consistently have the array kind, i.e. the description of the underlying data, together with the corresponding load/set kind, i.e. the description of the access.

This includes PR3069 and PR3070 as separate commits.


middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
```diff
@@ -814,74 +826,114 @@ let bigarray_set ~dbg ~unsafe kind layout b indexes value =
...
+        (Array_load (Naked_nativeints, Naked_nativeints, Mutable), array, index)
+    ]
+
+let array_load_unsafe ~array ~index array_kind array_load_kind ~current_region =
+  array_load_unsafe ~array ~index array_kind array_load_kind ~current_region
+  |> H.maybe_create_unboxed_product
+
+let array_set_unsafe dbg ~array ~index array_kind
+    (array_set_kind : Array_set_kind.t) ~new_values : H.expr_primitive list =
+  let[@inline] normal_case array_set_kind new_values =
```
Ekdohibs, line 883: Not sure the `[@inline]` is useful here, since this is a local function so this should probably be `[@local]`?

middle_end/flambda2/parser/flambda_parser.mly
```diff
@@ -476,23 +476,28 @@ string_accessor_width:
       | 128, Some 'u' -> One_twenty_eight {aligned = false}
       | _, _ -> Misc.fatal_error "invalid string accessor width" }
 
-array_accessor_width:
-  | { Scalar }
-  | KWD_VEC128 { Vec128 }
+(* CR mshinwell: temporary hack with array_load_kind *)
```
Ekdohibs, line 479: This needs to be fixed, since it causes CI failures otherwise.

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
mshinwell: I've tended to use `[@inline]` as sometimes `[@local]` is refused, but it does work in this case, so fine.

middle_end/flambda2/to_cmm/to_cmm_primitive.ml
```diff
@@ -247,27 +247,65 @@ let array_set_128 ~dbg ~element_width_log2 ~has_custom_ops arr index new_value =
...
+  match array_kind, load_kind with
+  | ( ( Values | Immediates | Naked_floats | Naked_int32s | Naked_float32s
+      | Naked_int64s | Naked_nativeints | Naked_vec128s ),
+      Immediates ) ->
+    C.int_array_ref arr
+      (accessor_assumes_non_custom_block dbg index array_kind)
+      dbg
+  | ( ( Values | Immediates | Naked_floats | Naked_int32s | Naked_float32s
+      | Naked_int64s | Naked_nativeints | Naked_vec128s ),
+      (Naked_int64s | Naked_nativeints) ) ->
```
Gbury, line 272: why isn't it an error to do a read of a `Naked_int64` or `Naked_nativeints` on an array of values (or immediates) ? I'd expect the read to agree with the array_kind (or at least be compatible), and then if need be some conversion to be done (if one really wants to do that kinds of reads in the source code).

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
```diff
@@ -836,78 +848,110 @@ let bigarray_set ~dbg ~unsafe kind layout b indexes value =
...
     ~conditions:
       (array_access_validity_condition array array_kind index ~index_kind
-         ~size_int)
+         ~width_in_scalars:1 ~size_int)
     ~dbg
 
-let array_load_unsafe ~array ~index ~(mut : Lambda.mutable_flag)
-    (array_ref_kind : Array_ref_kind.t) ~current_region : H.expr_primitive =
+let array_load_unsafe ~array ~index ~(mut : Lambda.mutable_flag) array_kind
+    (array_ref_kind : Array_ref_kind.t) ~current_region : H.expr_primitive list
```
Ekdohibs, line 864: Why do we take both `array_kind` and `array_ref_kind` here? Couldn't we use `convert_array_ref_kind_to_array_kind`?

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
```diff
@@ -836,78 +848,110 @@ let bigarray_set ~dbg ~unsafe kind layout b indexes value =
...
     | Immutable | Immutable_unique -> Immutable
     | Mutable -> Mutable
   in
   match array_ref_kind with
-  | Immediates -> Binary (Array_load (Immediates, Scalar, mut), array, index)
-  | Values -> Binary (Array_load (Values, Scalar, mut), array, index)
+  | Immediates ->
+    [Binary (Array_load (array_kind, Immediates, mut), array, index)]
+  | Values -> [Binary (Array_load (array_kind, Values, mut), array, index)]
   | Naked_floats_to_be_boxed mode ->
```
Ekdohibs, line 875: General comment: here and in other places, I'm wondering if we shouldn't split `Array_ref_kind.t` into a two-layer type, with `Naked_floats_to_be_boxed` in one case and something like `Normal_case` containing all the rest. I think this would also simplify the code of #3068 which already introduces the distinction, but keeps a full type and a no-floats-to-be-boxed case, instead of layering them.

middle_end/flambda2/kinds/flambda_kind.ml
```diff
@@ -973,6 +973,24 @@ module With_subkind = struct
...
+      | Immediate_array | Value_array | Generic_array | Unboxed_float32_array
+      | Unboxed_int32_array | Unboxed_int64_array | Unboxed_nativeint_array
+      | Unboxed_vec128_array ->
+        true)
+    | Naked_number _ | Region | Rec_info -> false
+
+  let may_be_gc_scannable t =
+    match kind t with
+    | Value -> true
+    | Naked_number _ | Region | Rec_info -> false
```
Ekdohibs, line 993: I think here, and above, we can do a `Misc.fatal_error` on `Rec_info` which is essentially ghost code (but not on regions).

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
mshinwell: I've left a CR about this, will aim to do this once the remaining patches have been merged if it works out.

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
mshinwell: I've left a CR about this, will check once the remaining patches are merged.

Ekdohibs (approved):

Gbury (approved):
For the `to_cmm` parts

PR #3075, opened 2024-09-25
Add Punboxedint64array_reinterpret_{ref,set}

mshinwell:
Provides `Punboxedint64array_reinterpret_ref` and `Punboxedint64array_reinterpret_set` array load and store kinds in `Lambda` that can be used for accessing `Punboxedintarray Pint64` arrays using unboxed products and machine word-indexed addressing, with inbuilt reinterpret casts to/from whatever kinds are desired.

Probably needs a bit more work but this should suffice for testing.

Based on #3068


ccasin:
(We discussed in person and David is also going to add a test to `pprintast_unconditional.ml` to check if we need to modify any printers)

mshinwell:
Superceded by #3366

PR #3076, opened 2024-09-25
Make bind `&` tighter than `mod` in jkind parsing

dvulakh:
Currently,
```ocaml
type t = j & k mod m
```
parses as
```ocaml
type t = j & (k mod m)
```

The crowdsourced concensus is that this is confusing. This PR makes it parse as
```ocaml
type t = (j & k) mod m
```
instead.

Why not require the user to be explicit and write parentheses either way? We'll likely make `ocamlformat` add more parentheses than are necessary, but feel that requiring parentheses in the parser would hurt approachability / discoverability of the new syntax.


ccasin:
(We discussed in person and David is also going to add a test to `pprintast_unconditional.ml` to check if we need to modify any printers)

ccasin (approved):
LGTM

PR #3079, opened 2024-09-25
Convert flambda2 `Block_load` and `Block_store` to use a static field index

TheNumbat:
Refactoring PR; behavior should be identical. There are some non-trivial changes to simplify unary/binary primitive, but as far as I can tell the behavior is the same.

`Block_load` is now a unary primitive and `Block_store` is now a binary primitive. Both contain a statically-known field index.
This will enable us to translate the field index to an offset in words in #3077.

Every place that previously created a `Block_load` or `Block_store` is able to provide a statically-known index, except for three:

- `Pfield_computed` and `Psetfield_computed`. These are now translated to `Array_load` and `Array_set` on a value array. These are only generated for object instance variables and patching blocks in `tmc`, so the addition of an opaque identity seems okay.
- `load_from_block_prim` in `rebuild_switch_with_single_arg_to_same_destination`. This is now translated to an `Array_load` on an immutable value array.

This PR also updates the fexpr serialization for block load/store and correspondingly regenerates the fl2 tests.


mshinwell (approved):
I made a couple of minor edits.

PR #3080, opened 2024-09-26
Simplify immutable array loads when contents are known

mshinwell:
This is useful in itself, but should be applicable to the `Switch` case in #3079 in particular.

Out-of-bounds unsafe accesses to known immutable arrays cause `Invalid` traps to be generated.


middle_end/flambda2/simplify/simplify_binary_primitive.ml
```diff
@@ -1087,23 +1088,44 @@ let simplify_array_load (array_kind : P.Array_kind.t)
...
+    in
+    let[@inline] contents_unknown () =
+      return_given_type (T.unknown (P.result_kind' prim)) ~try_reify:false
+    in
+    match T.meet_is_immutable_array (DA.typing_env dacc) array_ty with
+    | Invalid -> SPR.create_invalid dacc
+    | Need_meet -> contents_unknown ()
+    | Known_result (_kind, fields, _mode) -> (
+      (* We try to read the array if it is known to be immutable, even if the
+         load is marked as mutable. *)
```
lthls, line 1114: I think this is correct, but I consider that using a mutable array load on an immutable array should be invalid. So checking the mutability of the load and only simplifying immutable loads should work in all safe cases and will be less surprising in unsafe cases (immutable arrays unsafely cast as mutable ones).

lthls (approved):
Looks good, although I think a mutability check in `simplify_array_load` would be better.

PR #3082, opened 2024-09-26
Specialize array_get primitives for iarrays

TheNumbat:
Currently, reading from an iarray uses `%array_safe_get` and `%array_unsafe_get`, which are translated to mutable array loads in flambda2.

This PR adds an additional `mutable_flag` to the array get primitives and allows specialization to  update the flag to immutable when the primitive is applied to an `iarray`. 

I've manually confirmed the `iarray` tests now generate immutable flambda2 array loads, but I'm not sure if this should have additional tests.

### DIFF
--- a/middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
+++ b/middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
@@ -834,25 +834,30 @@ let check_array_access ~dbg ~array array_kind ~index ~index_kind ~size_int
          ~size_int)
     ~dbg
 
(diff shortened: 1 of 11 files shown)


goldfirere (commented):
Reviewed outside of middle_end, which I'm not familiar with.

I do think some tests are good: maybe one that can compile a little bit of code and then inspect the lambda that comes out? Getting this wrong could be very bad.

lthls (approved):
Approving the middle-end parts

TheNumbat:
@goldfirere ah, there was already a file for doing this. Added tests.

PR #3084, opened 2024-09-26
Check for the labeled_tuples extension

goldfirere:
While working elsewhere, I noticed that the labeled tuples extension is not actually checked anywhere. The first commit observes this and the next one fixes it.

Review: either @ccasin or @rtjoa.


rtjoa (approved):
Thanks for fixing this! (Out of pedantry, I added a test for labeled tuple types in type annotations, even though it's mostly redundant with the test for type expressions at the top of the file.)

PR #3086, opened 2024-09-26
Document [@error_message]

goldfirere:
As title. We somehow missed this.

@TheNumbat is a good writer and I would be pleased for his feedback here.

### DIFF
--- a/ocaml/jane/doc/extensions/error_message.md
+++ b/ocaml/jane/doc/extensions/error_message.md
@@ -1,0 +1,38 @@
+# The `[@error_message]` attribute
+
+You can put the `[@error_message]` attribute on type or kind constraints
+to add custom text to error messages. This can be useful in the output
+of a ppx to better direct the user to the source of the error.
+
+For example, suppose we have a ppx that translates `[%i_have_a x]` to `Some x`.
+This makes sense only for `value`s. So we might actually translate
+
+```
+let f x = [%i_have_a x]
+```
+
+to
+
+```
+let f x = Some (x : ((_ : value)[@error_message "only works with values"]))
+```
+
+which will say `only works with values` to the user in the error message.
+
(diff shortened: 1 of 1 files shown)


ocaml/jane/doc/extensions/error_message.md
```diff
@@ -0,0 +1,38 @@
+# The `[@error_message]` attribute
+
+You can put the `[@error_message]` attribute on type or kind constraints
+to add custom text to error messages. This can be useful in the output
```
TheNumbat, line 4: Is this only intended for use in ppxes? If so it should say that explicitly

ocaml/jane/doc/extensions/error_message.md
```diff
@@ -0,0 +1,38 @@
...
+of a ppx to better direct the user to the source of the error.
+
+For example, suppose we have a ppx that translates `[%i_have_a x]` to `Some x`.
+This makes sense only for `value`s. So we might actually translate
```
TheNumbat, line 8: > so the ppx might actually translate

ocaml/jane/doc/extensions/error_message.md
```diff
@@ -0,0 +1,38 @@
...
+let f x = [%i_have_a x]
+```
+
+to
+
+```
+let f x = Some (x : ((_ : value)[@error_message "only works with values"]))
+```
+
+which will say `only works with values` to the user in the error message.
```
TheNumbat, line 20: > to the user if `x` does not have kind `value`.

TheNumbat (approved):

ocaml/jane/doc/extensions/error_message.md
goldfirere: I've connected the two sentences more -- see if you like.

PR #3087, opened 2024-09-27
Fix test with backtrace that embeds 17 line numbers from the compiler

lukemaurer:
The expected output for `tool-toplevel/pr6468.ml` has a backtrace with a bunch of frames deep in the bowels of the compiler, wherever someone happened to raise `Not_found` most recently. This is of course always a hazard with `Not_found`. We can avoid it by just declaring and raising our own generic exception instead. (There's still a line in `topeval.ml` but at least that doesn't change nearly as often.)


lthls:
Isn't that just a consequence of declaring `raise` as `%reraise` ? I would be in favour of doing the obvious thing and going back to `%raise`.

lukemaurer:
> Isn't that just a consequence of declaring `raise` as `%reraise` ? I would be in favour of doing the obvious thing and going back to `%raise`.

That sounds reasonable to me, if there's consensus around doing that? In the short term, I just want to quit having to promote that dumb file

riaqn (approved):

PR #3088, opened 2024-09-27
Improve sort variable printing

ccasin:
This PR improves the printing of sort variables to be more stable.

The unboxed tuples PR (#2879) added unique ids to sort variables, which is useful for debugging.  I also changed the printing of sort variables to print these unique ids, rather than generating new numbers.  But this has introduced too much churn in the tests as any change to jkind checking is likely to perturb the number of sort variables that are generated during typechecking of a random program.  So, this PR restores the old strategy (but still keeps the uids, as they are essential when inspecting jkind checking in the debugger).


ncik-roberts (approved):
LGTM

PR #3089, opened 2024-09-27
Basic overwriting syntax

anfelor:
As discussed on slack. The only unexpected part might be the handling of underscores: I try to preserve the old behaviour which throws a syntax error whenever an underscore is found outside of a pattern. When the new extension is not enabled, you still get that syntax error, but now its thrown in `typecore.ml`.


anfelor:
This syntax will not allow us to use the `overwrite x with { field = 1; _ }` syntax that we talked about, since it only allows underscores in positions where an expression could appear. We could either add this syntax as a special case or use `overwrite x with { field = 1 }` or `overwrite x with { x with field = 1 }`.

ocaml/testsuite/tests/typing-unique/overwriting_syntax.ml
```diff
@@ -0,0 +1,124 @@
...
+                     ^^^^^^^^^^^^^^^^^^^^^^^^
+Alert : Overwrite not implemented.
+Uncaught exception: File "ocaml/parsing/location.ml", line 1112, characters 2-8: Assertion failed
+
+|}]
+
+type record = { a : int; b : int }
+
+let overwrite_record = function
+  { a; b } as t -> overwrite_ t with { b = a; a = _ }
```
goldfirere, line 24: I would also want to test things like `overwrite_ t with { b = a; _ }` and `overwrite_ t with { b = a }`. Similarly for the inline record case.

ocaml/testsuite/tests/typing-unique/overwriting_underscore.ml
```diff
@@ -0,0 +1,24 @@
+(* TEST
+   expect;
+*)
+
+(* For backwards-compatibility, we want to keep the same error message when the extension
```
goldfirere, line 5: No need to be backward compatible in this way. That said, the error message is fine as it is.

ocaml/parsing/parsetree.mli
```diff
@@ -462,6 +462,8 @@ and expression_desc =
   | Pexp_extension of extension  (** [[%id]] *)
   | Pexp_unreachable  (** [.] *)
   | Pexp_stack of expression (** stack_ exp *)
+  | Pexp_overwrite of Longident.t loc * expression (** overwrite_ x with exp *)
```
goldfirere, line 465: As discussed in person, I don't think the thing being overwritten needs to be a variable. But it's OK to go ahead with this for now and generalize later; leave a `(* CR uniqueness: ... *)` here to remind ourselves to return.

ocaml/parsing/location.mli
```diff
@@ -395,3 +395,5 @@ val raise_errorf: ?loc:t -> ?sub:msg list ->
 
 val report_exception: formatter -> exn -> unit
 (** Reraise the exception if it is unknown. *)
+
+val todo_overwrite_not_implemented : t -> 'a
```
goldfirere, line 399: Please leave a CR reminding us to remove this.

ocaml/parsing/lexer.mll
```diff
@@ -80,6 +80,7 @@ let keyword_table =
     "once_", ONCE;
     "open", OPEN;
     "or", OR;
+    "overwrite_", OVERWRITE;
```
goldfirere, line 83: Before merging to `main`, we'll need to check that there are no variables named `overwrite_` in `jane`.

goldfirere (approved):
Looks good! The parser changes were remarkably easy; I'm glad. :)

ocaml/testsuite/tests/typing-unique/overwriting_syntax.ml
anfelor: `overwrite_ t with { b = a; _ }` does not work in this PR. The reason is that underscores may only appear in places where expressions may appear.

goldfirere: OK. I suppose we don't need that syntax.

ocaml/parsing/lexer.mll
goldfirere: Is there a CR for this little piece?

ocaml/testsuite/tests/typing-unique/overwriting_underscore.ml
```diff
@@ -11,7 +11,7 @@ let underscore () = _
 Line 1, characters 20-21:
 1 | let underscore () = _
                         ^
-Error: Syntax error: wildcard "_" not expected.
+Error: Syntax error: "wildcard "_"" not expected.
```
goldfirere, line 14: Hrm. That doesn't look right.

ocaml/parsing/lexer.mll
anfelor: There is now :)

ocaml/testsuite/tests/typing-unique/overwriting_underscore.ml
anfelor: This is from the `main` branch. I vaguely remember reading a slack message about this change?

goldfirere: Yes, but I don't think there should be nested quotes.

ocaml/testsuite/tests/typing-unique/overwriting_underscore.ml
```diff
@@ -11,7 +11,7 @@ let underscore () = _
 Line 1, characters 20-21:
 1 | let underscore () = _
                         ^
-Error: Syntax error: "wildcard "_"" not expected.
+Error: Syntax error: "wildcard _" not expected.
```
goldfirere, line 14: I hate to nitpick here, but I was hoping to remove the outer quotes, not the inner ones. This is what happens before your patch on this syntax. If it's hard to remove the outer quotes (I understand how it might be!), then I would just craft a better error message. Normally, the stuff in quotes is stuff from the user's code, but the word `wildcard` doesn't appear in the code.

anfelor: But the outer quotes were introduced by upstream: https://github.com/ocaml/ocaml/commit/f042991a8764d61c2155f737371a444dd2d1850f

anfelor: For that matter, upstream now has nested quotes in their error message, so arguably we should follow that: https://github.com/ocaml/ocaml/blob/9bb7929ce61f5cb4457d294f16dee3ee132bc9ef/parsing/parser.mly#L2449

goldfirere: I don't observe nested quotes in the upstream error message. (Yes the quotes are nested in the source code, but not in the output.)

```
[igm-qws-u22468b] 09:44:27 ~/temp> cat bug.ml
let underscore () = _
[igm-qws-u22468b] 09:44:30 ~/temp> ocaml -version
The OCaml toplevel, version 5.3.0+dev4-2024-09-20
[igm-qws-u22468b] 09:44:34 ~/temp> ocamlc bug.ml
File "bug.ml", line 1, characters 20-21:
1 | let underscore () = _
                        ^
Error: Syntax error: wildcard "_" not expected.
```

anfelor: Ah, is the `wildcard "_"` part highlighted in your shell? I think ocaml might change the display depending on whether highlighting is available or not (and use quotes if not). My code also highlights the `wildcard _` when run in the shell.

anfelor: I have reverted the commit "Remove nested quotes" since it clearly does not do what we want. With the reverted commit, I get the same output on the shell in my fork and upstream.

PR #3090, opened 2024-09-27
Move unboxed vec128 to stable

TheNumbat:
The `vec128` layout is now allowed with the stable `layouts` + `simd` extensions. 
I've ported `ocaml_simd` to use unboxed vectors and updated all tests (including arrays & fields), which should be sufficient to promote to stable.

This PR also updates the tests in the compiler to run with the stable extensions and updates the unboxed types (implemented) documentation.

Note that no part of SIMD is supported in the bytecode backend, so nothing will be added to `Stdlib_stable`, which is jsoo compatible.


ocaml/jane/doc/extensions/unboxed-types/index.md
```diff
@@ -125,6 +131,7 @@ Each numeric type has its own library for working with it: `float_u`,
 modules in the `janestreet_shims` library.)
 
 * Unboxed constants are written with a prepended `#`.
+  There is no literal syntax for unboxed vectors: use the `Ocaml_simd_sse` library instead.
```
ccasin, line 134: Is our plan to public release this library on the bleeding-edge opam repo?  (Maybe make a ticket somewhere to note that plan if so, supposing one doesn't exist).

ocaml/typing/predef.ml
```diff
@@ -568,6 +564,10 @@ let add_simd_beta_extension_types add_type env =
                 Jkind.Const.Builtin.vec128.jkind)
       ~jkind_annotation:Jkind.Const.Builtin.vec128
 
+let add_simd_beta_extension_types add_type env =
+  let _add_type = mk_add_type add_type in
+  env
+
```
ccasin, line 570: Let's just delete this and the corresponding call from `env.ml`.  We can add it back if needed in the future.

ccasin (approved):
Looks good - see a couple minor things below.

ocaml/typing/predef.ml
TheNumbat: Sure, deleted

PR #3091, opened 2024-09-27
Add fields to newtype

liam923:
Merlin currently has a duplicate variant of `Typedtree.exp_extra` called `Texp_newtype'`. It is similar to `Texp_newtype` but contains a few extra fields that Merlin needs. Similarly, it defines a variant that it uses for elements of the `fp_newtypes` list in `Typedtree.function_param`. One case of the variant contains the existing fields, and the other contains the existing fields plus some additional ones.

Merlin relies on being able to deserialize typedtrees to be able to read cmt files. The aforementioned variants are added in such a way that Merlin can still correctly deserialize typedtrees produced by the compiler, but this is prone to breakage.

This PR adds the additional fields that Merlin needs to the type definitions in the compiler. This means that the compiler has to do a _little_ extra work, but its extremely small, so it seems well worth the cost for easier maintainability.

### DIFF
--- a/ocaml/typing/typedtree.mli
+++ b/ocaml/typing/typedtree.mli
@@ -228,7 +228,7 @@ and exp_extra =
          *)
   | Texp_poly of core_type option
         (** Used for method bodies. *)
-  | Texp_newtype of string * Jkind.annotation option
+  | Texp_newtype of Ident.t * string loc * Jkind.annotation option * Uid.t
         (** fun (type t : immediate) ->  *)
   | Texp_stack
         (** stack_ E *)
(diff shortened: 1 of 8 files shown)


ocaml/typing/typedtree.mli
```diff
@@ -228,7 +228,7 @@ and exp_extra =
          *)
   | Texp_poly of core_type option
         (** Used for method bodies. *)
-  | Texp_newtype of string * Jkind.annotation option
+  | Texp_newtype of Ident.t * string loc * Jkind.annotation option * Uid.t
```
goldfirere, line 231: Please add some comments here (and in `fp_newtypes`) explaining that the new fields are actually unused, but they're necessary for lining up with merlin. Otherwise someone might come along and delete them.

goldfirere (approved):

PR #3092, opened 2024-09-27
Allow unboxed products in externals

ccasin:
This allows unboxed products in externals, in the front-end.  It is both a precursor to unboxed-tuples-in-arrays, and has been independently requested by our users.

We can divide externals into four categories for the purposes of this PR, and they have varying levels of support:

1. *Primitives that support layout polymorphism and don't involve arrays*: These already work fine with unboxed products, so this PR simply allows such uses to typecheck.  A test of their behavior has been added next to the existing test for layout poly externals - these tests can be combined once unboxed tuples are out of beta.
2. *Primitives that support layout polymorphism and involve arrays*: These continue to reject unboxed products. The check for this is moved to the array kind calculations in `typeopt`.  These will be supported in a near-future PR.
3. *Primitives that don't support layout polymorphism*: These work on fixed layouts, so they naturally do not support unboxed products. Using them with an unboxed product used to be rejected by typedecl, and is now rejected by the existing primitive layout checks in `primitive.ml`.
4. *C externals*: This PR allows C externals to mention unboxed products, but only when `layouts_alpha` is enabled.  They have been gated behind alpha because it was pointed out that we have chosen a non-standard calling convention for them.  Backend folks will fix this, at which time they can be allowed at the same maturity as other uses of unboxed products. That PR will add test of the dynamic behavior, which would be annoying to do now because the non-standard calling convention makes it hard to write corresponding C stubs.  (cc: @mshinwell and @xclerc)

A few implementation details
* The errors when we reject category 4 above because `alpha` is not enabled are terrible. But this will change in the next PR that allows them in `beta`, so I think it is not worth fixing.
* The `native_repr` and `extern_repr` types are updated to support products, which motivated exposing some convenience definitions of constant sorts from `Jkind.Sort.Const.t`.  I've also exposed the equality `Jkind_types.Sort.Const.t = Jkind.Sort.Const.t` - I think it was just a mistake that it wasn't.
* I realized while writing this that C externals involving unboxed products can not be made upstream compatible with our current regime, unlike other unboxed types.  We'll have to do something about that when unboxed products reach stable - I left a long CR.

Review: @goldfirere , perhaps?


ocaml/typing/typeopt.ml
```diff
@@ -801,6 +808,9 @@ let lazy_val_requires_forward env loc ty =
     Misc.fatal_error "Unboxed value encountered inside lazy expression"
   | Float -> Config.flat_float_array
   | Addr | Int -> false
+  | Product cs ->
+    let kind = Jkind.Sort.Const.Product cs in
+    raise (Error (loc, Unsupported_product_in_lazy kind))
```
goldfirere, line 813: Is this error case tested? I don't remember seeing it.

ocaml/lambda/lambda.ml
```diff
@@ -1871,21 +1871,24 @@ let structured_constant_layout = function
...
 
+let rec layout_of_const_sort (c : Jkind.Sort.Const.t) : layout =
+  match c with
+  | Base Value -> layout_any_value
+  | Base Float64 -> layout_unboxed_float Pfloat64
+  | Base Float32 -> layout_unboxed_float Pfloat32
+  | Base Word -> layout_unboxed_nativeint
+  | Base Bits32 -> layout_unboxed_int32
+  | Base Bits64 -> layout_unboxed_int64
+  | Base Void -> assert false
```
goldfirere, line 1882:
```suggestion
  | Base Void -> layout_unboxed_product []
```

?

Though, actually, I think I want

```
type base_layout =
  | Pvalue of value_kind
  | Punboxed_float of boxed_float
  | Punboxed_int of boxed_integer
  | Punboxed_vector of boxed_vector

type layout =
  | Ptop
  | Punboxed_product of base_layout iarray
  | Pbottom
```

and then to flatten the lists. This would be a big change, I think, and not (at all) for this PR. But could this be a good idea? Maybe we stick `Pvalue` also in `layout` because it's a common case and indirection is annoying... but the singleton list of values would be identical to value.

goldfirere (approved):

ocaml/testsuite/tests/typing-layouts-products/basics.ml
```diff
@@ -600,74 +600,83 @@ Error: This value escapes its region.
...
 [%%expect{|
 type t_product : value & value
-Line 3, characters 25-38:
-3 | external ext_tuple_arg : #(int * bool) -> int = "foo"
-                             ^^^^^^^^^^^^^
-Error: Unboxed product layouts are not supported in external declarations
+Line 3, characters 25-45:
+3 | external ext_tuple_arg : #(int * bool) -> int = "foo" "bar"
+                             ^^^^^^^^^^^^^^^^^^^^
+Error: The primitive [foo] is used in an invalid declaration.
```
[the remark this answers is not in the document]
ccasin: This is indeed the horrible error.

ocaml/typing/typeopt.ml
ccasin: This error should be unreachable if the typechecker is correct (note it gives the scary "Contact us" message).

ocaml/lambda/lambda.ml
ccasin: Re void: Possibly soon! But I think not in this PR - should get some testing.

Re: flattening the layouts.  A good question for @mshinwell.  My instinct is that the job of flattening products is part of "unarization", and that task belongs to the `lambda` -> `flambda` pass, so better to keep the product structure here.

ocaml/testsuite/tests/typing-layouts-products/externals_alpha.ml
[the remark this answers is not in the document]
ccasin: Indeed - this is rejected by an existing check that externals work on representable things.  I've added a test.

ocaml/testsuite/tests/typing-layouts-products/basics.ml
```diff
@@ -600,74 +600,83 @@ Error: This value escapes its region.
...
 |}]
 
 external ext_tuple_arg_with_attr : (#(int * bool) [@unboxed]) -> int = "foo"
 [%%expect{|
 Line 1, characters 36-49:
 1 | external ext_tuple_arg_with_attr : (#(int * bool) [@unboxed]) -> int = "foo"
                                         ^^^^^^^^^^^^^
-Error: Unboxed product layouts are not supported in external declarations
+Error: Don't know how to unbox this type.
+       Only "float", "int32", "int64", "nativeint", vector primitives, and
```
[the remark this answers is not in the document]
ccasin: Well spotted. I've adjusted the error in the way we agreed to on slack.

goldfirere:
Good to merge when CI passes

PR #3094, opened 2024-09-29
Reduce the need for the `pprintast_unconditional` test

ncik-roberts:
Simplify the checklist suggested by GitHub Actions (introduced in #3085) when a PR touches the parser. It's nice to eliminate redundancy from checklists we'll frequently follow.

I do this by encouraging that tests be added just to `source_jane_street.ml` and not also to `pprintast_unconditional.ml`. `pprintast_unconditional` is mostly a worse version of `parsetree/source_jane_street.ml`. This statement isn't 100% true. Here are ways that it is/was false:
 
  * `source_jane_street` didn't used to check that `Pprintast` still works even when all extensions are disabled. **Response:** I make it check that. 
  * `source_jane_street` runs the type-checker in addition to the parser, so can't accept some programs that can be pretty-printed. **Response**: I don't delete the `pprintast_unconditional` test, and capture this nuance in a comment.

### DIFF
--- a/ocaml/testsuite/tests/parsetree/source_jane_street.ml
+++ b/ocaml/testsuite/tests/parsetree/source_jane_street.ml
@@ -4,13 +4,21 @@
    expect;
 *)
 
-(* This file is two tests in one! It's both an expect-test, run in the
-   normal way. It *also* is processed by [test.ml] (in this same directory)
-   to make sure that Pprintast round-trips. *)
-
 (* There should be examples of all Jane Street syntax here. This file can
    thus additionally serve as a cheap-and-cheerful documentation for all of
    our features. *)
+
+(* This file is three tests in one!
+    - It's an expect-test, run in the normal way.
+    - It is processed by [test.ml] (in this same directory) to make sure that Pprintast round-trips.
+    - It is processed by [test_ppx.ml] (in this same directory) to make sure that ppxes run properly
+      on examples of Jane Street syntax.
+*)
+
+(* Use [language-extensions/pprintast_unconditional.ml] to document pieces of syntax that
+   can't be added to this file (e.g. because you haven't yet implemented type-checking for
+   your feature).
+*)
 
 (***********)
 (* Layouts *)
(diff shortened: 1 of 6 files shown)


ocaml/testsuite/tests/parsetree/source_jane_street.ml
```diff
@@ -4,14 +4,22 @@
...
 
+(* This file is three tests in one!
+    - It's an expect-test, run in the normal way.
+    - It is processed by [test.ml] (in this same directory) to make sure that Pprintast round-trips.
+    - It is processed by [test_ppx.ml] (in this same directory) to make sure that ppxes run properly
+      on examples of Jane Street syntax.
+*)
+
+(* Use [language-extensions/pprintast_unconditional.ml] to document pieces of syntax that
+   can't be added to this file (e.g. because you haven't yet implemented type-checking for
```
goldfirere, line 19: Is the problem that test_ppx doesn't like tests that lead to `Misc.fatal_error`? Then we should say that more explicitly. Because features that don't type-check can still be tested fine in an expect-test -- you'll just expect type-checking to fail.

We could probably reasonably make one "unimplemented" error somewhere that any unimplemented features could use instead of `Misc.fatal_error` if that's indeed the problem.

goldfirere (approved):

ocaml/testsuite/tests/parsetree/source_jane_street.ml
ncik-roberts: I thought `test_ppx` didn't like tests that lead to `Misc.fatal_error`, but it turns out that it's fine as long as there is a type error further up in the file.

I've moved the modal kind test into `source_jane_street.ml` and removed the confusing cross-linking of the files.

goldfirere: I think this is an improvement. But now: why do we need `pprintast_unconditional`? Isn't the printing already covered in the new test?

ncik-roberts: I think it's nice to test the individual `Pprintast` entry points somewhere  (`expression`, `tyvar`, `mode`, ...) against our extensions. `source_jane_street.ml` just tests `Pprintast.structure` and `Pprintast.signature`.

PR #3095, opened 2024-09-30
Support recursively applying modalities to signature when include

riaqn:
This PR allows:
```
include [@recursive_modalities] S @@ portable
```
such that not only items directly in `S`, but also items in recursive-submodules of `S` will be applied modalities.

### DIFF
--- a/ocaml/testsuite/tests/typing-modes/val_modalities.ml
+++ b/ocaml/testsuite/tests/typing-modes/val_modalities.ml
@@ -613,8 +613,7 @@ end
 module type T' = sig val baz : 'a -> 'a @@ portable module M : S end
 |}]
 
-(* CR zqian: fix this *)
-(* [MT] is not in the scope of [include S] and thus not affected. *)
+(* submodule whose type is in the signature *)
 module type S = sig
   module type MT = sig
     val foo : 'a -> 'a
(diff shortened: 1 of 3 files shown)


ocaml/testsuite/tests/typing-modes/val_modalities.ml
goldfirere, line 1: Also might be good to confirm that this works correctly on abstract module types, like in

```
module type S = sig
  module type MT
end

module F (X : S) = struct
  module type S2 = sig
    include [@recursive_modalities] X.MT @@ portable
  end
end
```

or more simply

```
module type S = sig
  module type MT
  include [@recursive_modalities] MT @@ portable
end
```

ocaml/typing/typemod.ml
```diff
@@ -990,7 +990,26 @@ let apply_modalities_signature modalities sg =
...
+
+and apply_modalities_module_type env modalities = function
+  | Mty_ident p ->
+      let mtd = Env.find_modtype p env in
+      begin match mtd.mtd_type with
+      | None -> Mty_ident p
+      | Some mty -> apply_modalities_module_type env modalities mty
+      end
+  | Mty_strengthen (mty, p, alias) ->
+      Mty_strengthen (apply_modalities_module_type env modalities mty, p, alias)
```
goldfirere, line 1008: There should probably be a test for this.

ocaml/typing/typemod.ml
```diff
@@ -990,7 +990,26 @@ let apply_modalities_signature modalities sg =
...
+  | Mty_signature sg ->
+      let sg = apply_modalities_signature ~recursive:true env modalities sg in
```
goldfirere, line 1010: There should probably be a test for this case.

ocaml/parsing/builtin_attributes.ml
```diff
@@ -112,6 +112,7 @@ let builtin_attrs =
   ; "layout_poly"
   ; "no_mutable_implied_modalities"
   ; "or_null_reexport"
+  ; "recursive_modalities"
```
goldfirere, line 115: Please add an example to the `parsetree/source_jane_street.ml` file.

goldfirere (approved):

ocaml/testsuite/tests/typing-modes/val_modalities.ml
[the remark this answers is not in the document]
riaqn: It's not abstract module types that's causing this, but that `MT` is not in scope when type checking `S2`. 

I will fix this.

ocaml/typing/typemod.ml
```diff
@@ -990,7 +990,30 @@ let apply_modalities_signature modalities sg =
...
+      Sig_module (id, pres, md, rec_, vis)
+  | item -> item
+  ) sg
+
+and apply_modalities_module_type env modalities = function
+  | Mty_ident p ->
+      let mtd_type =
+        match Env.find_modtype p env with
+        | mtd -> mtd.mtd_type
+        | exception Not_found -> None
```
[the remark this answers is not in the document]
riaqn: Oh I had a wrong impression of how this will all work. I think it should just raise, yes.

ocaml/typing/typemod.ml
```diff
@@ -1694,10 +1726,19 @@ and transl_signature env sg =
       | Structure ->
         Tincl_structure, extract_sig env smty.pmty_loc mty
     in
-    let modalities =
-      Typemode.transl_modalities ~maturity:Alpha Immutable [] modalities
+    let sg, modalities =
+      match modalities with
+      | [] -> sg, Mode.Modality.Value.Const.id
```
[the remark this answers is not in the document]
riaqn: Yeah it would be tiny function, so I will just be lazy and resolve this.

PR #3099, opened 2024-09-30
Fix signature inclusion bug for all-float mixed blocks

ncik-roberts:
A signature can hide the fact that some fields of a record declaration are floats. When this happens, we should be careful that a different representation isn't silently chosen for the signature vs. the structure. Not doing so leads to miscompilation.

### DIFF
--- a/ocaml/typing/includecore.ml
+++ b/ocaml/typing/includecore.ml
@@ -218,6 +218,7 @@ type record_mismatch =
   | Float_representation of position
   | Ufloat_representation of position
   | Mixed_representation of position
+  | Mixed_representation_with_flat_floats of position
 
 type constructor_mismatch =
   | Type of Errortrace.equality_error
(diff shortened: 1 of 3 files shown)


ocaml/typing/includecore.ml
```diff
@@ -700,6 +705,31 @@ module Record_diffing = struct
     else
       Some (diffing loc env params1 params2 l r)
 
+  let find_mismatch_in_mixed_record_representations
+      ({ value_prefix_len = v1; flat_suffix = s1 } : mixed_product_shape)
+      ({ value_prefix_len = v2; flat_suffix = s2 } : mixed_product_shape)
+    =
+    if v1 = v2 then None
+    else
+      (* It's currently possible for the value prefix length to legally differ
```
goldfirere, line 714: Yes.... but this is really OK? Is there a test showing so? It seems very weird to me to have two objects of the same type floating about with different headers in this way. But I don't actually see yet what goes wrong. Does marshalling? Does polymorphic comparison?

goldfirere (approved):
Yikes. Good catch!

ocaml/typing/includecore.ml
ncik-roberts: > Does marshalling? Does polymorphic comparison?

Chris and I talked about this: Both of these are unsupported for mixed blocks (i.e. will raise). So we think it's fine. I'll add a test.

My one concern is the optimizer. I'm discussing with flambda2 devs whether this causes issues. If it does, I'll keep immediates in the value prefix. But we should tackle that separately from this PR.

ncik-roberts: (Will open a follow-up PR for this.)

PR #3101, opened 2024-10-01
Zero alloc: better error widen witness

gretay-js:
When zero_alloc checker applies widening to produce a value that leads to an annotation check failure, the error message doesn't have any witnesses, and this is very confusing to the user. 

This PR prints a nicer message. It's still not very helpful because it points to the entire function instead of a particular construct, but at the point when we perform widening we don't have the relevant debuginfo and it would be akward to pass through (because we would need to put it in the abstract value; widening can happen during join invoked from the dataflow analysis). 

Technically, we add a new `Witness.kind` for widening without any further detail, and then use debug info of the function during error printing. We rarely need this message because most of the time we have enough other witnesses or the function has no `zero_alloc` assert annotation, so we don't need to track its witnesses.

### DIFF
--- a/backend/zero_alloc_checker.ml
+++ b/backend/zero_alloc_checker.ml
@@ -519,6 +519,8 @@ end = struct
     val get_unresolved_names : t -> String.Set.t
 
     exception Widen
+
+    exception Widen_with_witness
   end = struct
     (* Join of two Transform with the same set of vars: merged both sets of Vars
        into one Transform in normal form, without loss of precision or
@@ -532,10 +534,18 @@ end = struct
 
     exception Widen
 
+    exception Widen_with_witness
+
     let maybe_widen t =
       match !Flambda_backend_flags.zero_alloc_checker_join with
       | Keep_all -> t
-      | Widen n -> if M.cardinal t > n then raise Widen else t
+      | Widen n ->
+        if M.cardinal t > n
+        then
+          if M.exists (fun _var tr -> Transform.has_witnesses tr) t
+          then raise Widen_with_witness
+          else raise Widen
+        else t
       | Error n ->
         if M.cardinal t > n
         then
@@ -1106,7 +1116,9 @@ end = struct
   let widen_witness = Witnesses.create Witness.Widen Debuginfo.none
 
   let bounded_join f =
-    try Join (f ()) with Transforms.Widen -> Top widen_witness
+    try Join (f ()) with
+    | Transforms.Widen -> Top Witnesses.empty
+    | Transforms.Widen_with_witness -> Top widen_witness
 
   (* Keep [join] and [lessequal] in sync. *)
   let join t1 t2 =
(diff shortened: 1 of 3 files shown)


gretay-js:
This patch causes `flambda2_dev` ci to get stuck. I'll investigate later.

backend/zero_alloc_checker.ml
```diff
@@ -1638,6 +1653,17 @@ end = struct
...
+        | Widen ->
+          ( Format.dprintf
+              "details are not available. This may be a false alarm due to \
+               conservative analysis.\n\
+               Hint: for more precise results, recompile this function with\n\
+               \"-function-layout topological\" or \"-zero-alloc-checker-join \
+               0\" flags.\n\
+               The \"-zero-alloc-checker-join 0\" flag may substantially \
+               increase compilation time.\n\
+               (widenning applied in function %s%s)" t.fun_name component_msg,
```
xclerc, line 1665:
```suggestion
               (widening applied in function %s%s)" t.fun_name component_msg,
```

backend/zero_alloc_checker.ml
```diff
@@ -2264,7 +2296,12 @@ end = struct
...
-      | Iifthenelse _ | Iswitch _ -> next
+      | Iifthenelse _ ->
+        report t next ~msg:"transform" ~desc:"ifthenelse" i.dbg;
+        next
+      | Iswitch _ ->
+        report t next ~msg:"transform" ~desc:"switch" i.dbg;
+        next
       | Icatch (_rc, _ts, _, _body) ->
         report t next ~msg:"transform" ~desc:"catch" i.dbg;
         next
```
xclerc, line 2307: (Does not really matter, but the `desc` are
a bit inconsistent: `iexit`, `ifthenelse`,
and `try-with`.)

backend/zero_alloc_checker.ml
```diff
@@ -517,6 +519,8 @@ end = struct
     val get_unresolved_names : t -> String.Set.t
 
     exception Widen
+
+    exception Widen_with_witness
```
xclerc, line 523: (Possibly a tiny bit safer to add
a nested value to the existing
exception.)

xclerc (approved):

ccasin (approved):
Agree the new error is an improvement, thanks!

I think in an ideal world we would have a test showing that the actions the hint suggests to get more information actually result in more information for this example.  I took a quick look and spotted an exisitng test for `-zero-alloc-checker-join 0` on this file, but not `-function-layout topological`.

gretay-js (commented):
@xclerc  @ccasin  thanks for the reivew. I addressed the comments and plan to merge when the ci is clean.

> I think in an ideal world we would have a test showing that the actions the hint suggests to get more information actually result in more information for this example. I took a quick look and spotted an exisitng test for -zero-alloc-checker-join 0 on this file, but not -function-layout topological.

I couldn't come up with a small example. I did have an example in real code where passing `-function-layout topological` was enough.

PR #3104, opened 2024-10-02
zero alloc: show extra details in error messages

gretay-js:
This is a draft to help discussion of the proposed changes to `Debuginfo`. The motivation for making these changes is to produce a new format of zero_alloc error messages. The new format is based on user feedback to make it easier to identify and eliminate allocations reported by the check.  

For example, instead of
```
File "test_assume_fail.ml", line 16, characters 5-15:
Error: Annotation check for zero_alloc strict failed on function Test_assume_fail.test48 (camlTest_assume_fail.test48_HIDE_STAMP)

File "test_assume_fail.ml", line 17, characters 2-55:
Error: called function may allocate on a path to exceptional return (direct call camlCamlinternalFormat.make_printf_HIDE_STAMP) (test_assume_fail.ml:17,2--55;test_assume_fail.ml:14,54--77;p
rintf.ml:48,18--43;printf.ml:46,2--31)
```
the error message looks like this
```
File "test_assume_fail.ml", line 16, characters 5-15:
Error: Annotation check for zero_alloc strict failed on function Test_assume_fail.test48 (camlTest_assume_fail.test48_HIDE_STAMP)

File "test_assume_fail.ml", line 17, characters 2-55:
Error: called function may allocate on a path to exceptional return (direct call camlCamlinternalFormat.make_printf_HIDE_STAMP)
inlined from
stdlib/printf.ml:46,2--31[Stdlib.Printf.ksprintf][camlStdlib__Printf__ksprintf_HIDE_STAMP]
stdlib/printf.ml:48,18--43[Stdlib.Printf.sprintf][camlStdlib__Printf__sprintf_HIDE_STAMP]
repo/root/relative/dir/test_assume_fail.ml:14,54--77[Test_assume_fail.test46][camlTest_assume_fail.test46_HIDE_STAMP]
repo/root/relative/dir/test_assume_fail.ml:17,2--55[Test_assume_fail.test48]
```

The PR should eventually be split up (see individual commits). 

- Change `Debuginf.compare` function to not reverse the lists. There is a CR-someday about it already. I don't see where it matters, but if this seems too risky, we can implement custom function in `zero_alloc_checker` directly.
- Don't print `[FS=]` when printing extend debug info. It was added for dwarf but I think it was only used for debugging.
- Add `dinfo_dir` field to record `-directory`. This may increase cmx size but note that the string for the current compilation's directory is shared. Debuginfo in code inlined from other libraries will have some duplication (need to measure). This change does not allocate new strings until actually printing error messages. By the way, there are other places in the compiler that can use `-directory` to improve error messages. Maybe eventually we can get this information from dwarf instead of storing it in cmx.
- Reverse the order of "inlined stack" entries shown in zero_alloc error messages, so that the innermost inlined location is listed first, because it's the code where the offending instruction originated from 
- Print function symbol and scope location for "inlined stack". It comes with a change of format of zero_alloc error messages, otherwise the lines are too long and hard to read. Technically, it also brings a bit of refactoringing in `Debuginfo.print_compact*` functions. 
- The above extra information is disabled by default and can be enabled with `-zero-alloc-checker-details-extra` flag that is intended for experimentation. 
- When printing scoped locations, don't include zero_alloc attribute. It was a mistake to print it in  `Debuginfo.string_of_scopes` by default, it should only be printed for debugging. Added an optional. 
- I've noticed that `Debuginfo.of_items` and `Debuginfo.to_items` functions were added as part of the new dwarf inlined frames support so there is some duplication in the interface of `Debuginfo.Dbg` but I haven't refactored it.


gretay-js:
A few indepedent changes were split out of this PR and should be merged first:
- https://github.com/ocaml-flambda/flambda-backend/pull/3120
- https://github.com/ocaml-flambda/flambda-backend/pull/3119
- https://github.com/ocaml-flambda/flambda-backend/pull/3117
- https://github.com/ocaml-flambda/flambda-backend/pull/3118 (not required for this PR)

gretay-js:
This RP is now ready for review.

driver/oxcaml_flags.ml
```diff
@@ -63,6 +63,7 @@ type zero_alloc_checker_join =
   | Widen of int  (* n > 0 *)
   | Error of int (* n > 0 *)
 
+let zero_alloc_checker_details_extra = ref false
```
ccasin, line 66: I think we should just turn this on by default. The new details are usually useful, and too many people will not know they have to set this.  (Or maybe your plan is to enable it in the build system by default?)

ccasin (approved):
LGTM

Big improvement in errors - thanks!

driver/oxcaml_flags.ml
gretay-js: The editor integration many need to be tweaked to show these errors.

PR #3106, opened 2024-10-02
`stack_ fun` doesn't need parens

ncik-roberts:
This PR makes it so you can ensure a function argument is stack-allocated without wrapping it in another set of parens.

Specifically, you can now write:
```ocaml
List.map xs ~f:(stack_ fun x -> x + y)
```

whereas previously this would require:
```ocaml
List.map xs ~f:(stack_ (fun x -> x + y))
```

The same applies for `stack_ function`.

We haven't decided we definitely want this -- this is just a proof-of-concept to be bikeshed in due course.


goldfirere (approved):
Did we agree that this was a good design? I'm personally in favor, but I'd want @riaqn's opinion.

Should this patch also update the pretty-printer to avoid the parens? (Honest question; not sure it's worth worrying about.)

ncik-roberts:
> We haven't decided we definitely want this -- this is just a proof-of-concept to be bikeshed in due course.

Sorry, maybe I should have opened a draft PR instead... I'll close for now so nobody accidentally merges until after we discuss. (Zesen, let's discuss in person at the scheduled meeting time.)

ncik-roberts:
The outcome of our conversation a few months ago:
  * We want `stack_ fun x -> …` and `stack_ A x` to parse and to be formatted without parens &mdash; **this PR does the parsing for that**.
  * We want some things to parse but to be formatted by ocamlformat:
    * `stack_ (x, y), z`  –formatted to→ `stack_ ((x, y), z)` &mdash; **this PR does not change parsing for tuples, I couldn't easily figure out how to do that**. `stack_ (x, y), z` is still the same as `(stack_ (x, y)), z`.

I'm still interested in pushing this through. How do others feel?

goldfirere:
Would ocamlformat rewrite `stack_ (x, y), z` to `(stack_ (x, y)), z`? I'm a little uncomfortable about the difference between `stack_ A x` and `stack_ (x, y), z`. What about `stack_ (x, y) :: z`?

cc @riaqn

ncik-roberts:
> Would ocamlformat rewrite stack_ (x, y), z to (stack_ (x, y)), z? 
It won't, without changes, but these changes could be done at the same time that ocamlformat is changed to not rewrite `stack_ A x` to `stack_ (A x)` (which, will not necessary to realize some benefit from the present PR, is necessary to realize all the benefit).

> What about stack_ (x, y) :: z?
This is also interpreted as `(stack_ (x, y)) :: z`, and the same changes could be made to ocamlformat to make this clear.

riaqn:
@ncik-roberts Are we saying that `stack_ (x, y), z` is `stack_ ((x, y), z), but `stack_ (x, y) :: z` is `(stack_ (x, y)) :: z`? I find that strange - `,` and `::` are very similiar - they are both infix operators that works as constructors. 

I'm happy to do any of the following:
1. `stack_` works like a function name (the current situation): this has the benefit of being described in a single sentence. Users might complain and we can improve upon that.
2. `stack_` works just like a function name, except when followed by `fun` or `function` or a prefix constructor.
3. `stack_` works like `local_` - this has the benefit that users are already familiar with `local_`.

I thought about some intermediate solution between 2 and 3, that specially treat `;` and `,`, but I suspect it's too much to remember for the users, that we might as well just adopt 3.

ncik-roberts:
> Are we saying that stack_ (x, y), z is stack_ ((x, y), z), but stack_ (x, y) :: z is (stack_ (x, y)) :: z? I find that strange - , and :: are very similiar - they are both infix operators that works as constructors.

No, that's not what we are saying.
* `stack_ (x, y), z` continues to be `(stack_ (x, y)), z`
* `stack_ (x, y) :: z` continues to be `(stack_ (x, y)) :: z`

As mentioned [here](https://github.com/ocaml-flambda/flambda-backend/pull/3106#issuecomment-2614072899), we would ideally like both of these to be `stack_ (...)`, but I found this more difficult to implement and would prefer to defer this for now, given the immediate benefit of the present PR (and the not-yet-implemented ability of ocamlformat to clearly indicate which parse is being chosen in both of these examples).

I think this PR implements exactly (2). The ideal state that we discussed in the meeting a few months ago is:

  2.5. `stack_` binds looser than `,` and `::`, and otherwise same as (2).

but again, (2.5) is not what's implemented here. (2) is.

ncik-roberts:
Thanks!

> do you think it's worthwhile to update pprintast? Currently it's probably printing extra parens (which is still valid).

I think it's not worthwhile right now &mdash; it doesn't seem like a trivial refactor, and the current behavior is correct.

PR #3107, opened 2024-10-02
Overwriting primitives in lambda

anfelor:
This adds the basic primitives needed to support overwriting to lambda.
 - In the middle end, we call `todo_overwrite_not_implemented`.
 - In `value_recursion_compiler`, we disallow overwriting
 - In `tmc` we allow overwriting.
 - In `bytecode`, we map reuses to fresh allocations.


ocaml/lambda/lambda.mli
```diff
@@ -332,6 +332,39 @@ type primitive =
...
+  | Preusemixedblock of
+      { tag : int;             (* The immutable tag of both the old and resulting block *)
+        mut : mutable_flag;                            (* Mutability of resulting block *)
+        shape : mixed_block_shape;                    (* Block shape of resulting block *)
+        resets : reset_field list;                  (* Reset-information for each field *)
+        (* Note: even if the mode is Alloc_local, the old allocation may be on the heap *)
+        mode : alloc_mode;                             (* Alloc mode of resulting block *)
+      }
+
+and reset_field =
```
goldfirere, line 365: As @anfelor and I discussed over Slack, I think it would be a significant win to change the representation here, including `reset_field` elements only when we are actually changing a field. (One way to understand this is the observation that `'a option list` is isomorphic to `(int * 'a) list`, ignoring trailing `None`s.) Not sure what the consumer of this is going to look like, so maybe we can do something even better than an `int` to count the missing fields, but storing an `int` is better than making a list with likely a lot of `None`s in it.

ocaml/lambda/lambda.mli
```diff
@@ -332,6 +332,39 @@ type primitive =
      if the value is locally allocated *)
   (* Fetching domain-local state *)
   | Pdls_get
+    (* In-place reuse of heap blocks.
+     Each of these takes the allocation to be reused as first argument.
+     The following arguments correspond to the fields that are Reuse_set. *)
+  | Preuseblock of
```
goldfirere, line 338: As @anfelor and I discussed over Slack, I don't currently see the need to have 4 separate primitives here. The overwriting action for all the different kinds of blocks should be the same, I think. This is really a question for @mshinwell or the folks at OCamlPro to answer than @anfelor, but I'm wondering if all of this could be collapsed to one `Preuseblock`. Do we need `mutable_flag`? Do we need `alloc_mode`? Without more knowledge of the middle/back ends, I can't be sure, but it's not obvious to me why we'd need these here.

ocaml/lambda/lambda.ml
```diff
@@ -1713,6 +1732,10 @@ let primitive_may_allocate : primitive -> alloc_mode option = function
   | Pmakefloatblock (_, m) -> Some m
   | Pmakeufloatblock (_, m) -> Some m
   | Pmakemixedblock (_, _, _, m) -> Some m
+  | Preuseblock { mode } -> Some mode
```
goldfirere, line 1735: But these new primitives can't allocate, can they? It's one of the reasons I don't think the `mode` is necessary.

ocaml/lambda/lambda.ml
```diff
@@ -1666,6 +1679,12 @@ let find_exact_application kind ~arity args =
           if arity <> List.length tupled_args
           then None
           else Some tupled_args
+      (* CR: this allocation is removed by the simplifier and
```
goldfirere, line 1682: Best to put something after the `CR` so we can find it later. Maybe `CR uniqueness`?

ocaml/parsing/location.ml
```diff
@@ -1106,3 +1106,7 @@ let () =
 
 let raise_errorf ?(loc = none) ?(sub = []) =
   Format.kdprintf (fun txt -> raise (Error (mkerror loc sub txt)))
+
+let todo_overwrite_not_implemented t =
```
goldfirere, line 1110:
```suggestion
(* CR uniqueness: remove this *)
let todo_overwrite_not_implemented t =
```

ocaml/lambda/translprim.ml
```diff
@@ -1064,8 +1064,22 @@ let specialize_primitive env loc ty ~has_constant_constructor prim =
         Some (Primitive (Pmakeblock(tag, mut, Some shape, mode),arity))
       else None
     end
+  | Primitive (Preuseblock { tag; mut; shape = None; resets; mode }, arity), fields ->
```
goldfirere, line 1067: It's not clear to me that this little optimization buys anything for memory re-use. @mshinwell ?

ocaml/lambda/tmc.ml
```diff
@@ -84,11 +88,16 @@ module Constr : sig
...
     flag: Lambda.mutable_flag;
     shape : block_shape;
+    is_reusing : is_reusing;
     before: lambda list;
     after: lambda list;
     loc : Debuginfo.Scoped_location.t;
   }
 
+  val makeblock :
+    (int * Lambda.mutable_flag * block_shape * is_reusing * Lambda.alloc_mode) ->
```
goldfirere, line 98: Why tuple these arguments?

ocaml/lambda/tmc.ml
```diff
@@ -763,11 +784,12 @@ let rec choice ctx t =
...
       }
 
-  and choice_makeblock ctx ~tail:_ (tag, flag, shape, mode) blockargs loc =
+  and choice_makeblock ctx ~tail:_ (tag, flag, shape, is_reusing, mode) blockargs loc =
     let choices = List.map (choice ctx ~tail:false) blockargs in
     match Choice.find_nonambiguous_tmc_call choices with
     | Choice.No_tmc_call args ->
-        Choice.lambda @@ Lprim (Pmakeblock (tag, flag, shape, mode), args, loc)
+        let block = Constr.makeblock (tag, flag, shape, is_reusing, mode) in
+        Choice.lambda @@ Lprim (block, args, loc)
```
goldfirere, line 792: It's not clear to me that `args` should be the same here for the two cases of `is_reusing`. That is, when `is_reusing` says to allocate, `args` must be all the fields of the constructor. But when `is_reusing` says to reuse, `args` is only the changed fields. But that's not what this code does.

ocaml/lambda/tmc.ml
goldfirere, line 1: I would expect some tests for the treatment here -- even if the tests just print the lambda code and then fall over. Or at least a loud comment somewhere reminding us to test this thoroughly before release. A mistake here could lead to miscompilation.

ocaml/bytecomp/bytegen.ml
```diff
@@ -1160,6 +1178,40 @@ and comp_expr_list stack_info env exprl sz cont = match exprl with
...
         (Kpush :: comp_expr_list stack_info env rem (sz+1) cont)
 
+(* Compile a list of reuse arguments [x; e1; ...; eN] and their resets
+   [None; Some _; ...] to a primitive operation. For each None, we push
+   a read on x on the stack and for each Some _, we push an ei on the stack.
+   As with regular arguments, we push in reversed order and
+   leave the first argument in the accumulator. *)
+
+and comp_reuse_args stack_info env argl resets getfield sz cont =
+  let rec zip_resets idx resets argl =
```
goldfirere, line 1188: This could be made tail-recursive -- and it would remove the need for the `List.rev` after it's done.

ocaml/bytecomp/bytegen.ml
```diff
@@ -1160,6 +1178,40 @@ and comp_expr_list stack_info env exprl sz cont = match exprl with
...
+      old getfield sz cont
+
+and comp_reuse_expr_list stack_info env exprl old getfield sz cont =
+  let comp_reuse_arg exp cont =
+    match exp with
+    | `ReadOld idx -> comp_expr stack_info env old sz (getfield idx :: cont)
+    | `PushNew exp -> comp_expr stack_info env exp sz cont
+  in
+  match exprl with
+  | [] -> cont
```
goldfirere, line 1209: Shouldn't this case be impossible?

ocaml/bytecomp/bytegen.ml
goldfirere, line 1: This, too, should get a bunch of testing.

ocaml/lambda/lambda.ml
anfelor: The `mode` will soon also contain uniqueness information that informs us whether the new block will itself be used uniquely. One worry then might be if the optimizer can lift out `Preuseblock`s that are unique? (In fact, it can't: it could only do that if the cell we reuse is statically known, but that cell will have been marked as unique and can not be lifted out). Furthermore, we sometimes handle `Pmakeblock` and `Preuseblock` similarly (eg. in `tmc.ml`) and it might be helpful if they have the same fields.

ocaml/lambda/lambda.mli
anfelor: See `bytegen.ml` for a consumer of this representation

anfelor: Note that each  `None` corresponds to an underscore in the source program (eg. `overwrite x with (_,3)`). We should thus expect the number of `None`s to be somewhat small and should focus just on what the best representation is for code quality.

ocaml/lambda/tmc.ml
anfelor: Yes, for the `tmc` analysis we only need to consider the changed fields, since only they can contain a recursive call. The same code works in both cases.

ocaml/bytecomp/bytegen.ml
anfelor: Because we never reuse blocks of size zero? I agree with this. However, this case is symmetric to the one in `comp_expr_list` which to my surprise can indeed happen, suggesting that the compiler creates `Pmakeblock`s of size 0.

anfelor:
I agree that the additions to `tmc` and `bytegen` should be tested before we merge this. How about we first merge #3089 and then add some code to the type-checker that maps the overwrites to these primitives? Then we can run a few simple programs (like the `unique_rbtree` or `unique_sort` from the old PR) to test this code.

ocaml/lambda/lambda.mli
anfelor: Which list representation do you mean? The `'a option list` in the PR or the `(int * 'a) list` proposed by Richard?

goldfirere: And I don't think the number of `None`s will be all that small, in the case of record updates: `{x with f}` can have an arbitrarily large number of `None`s.

ocaml/lambda/lambda.ml
goldfirere: This line is in the function `primitive_may_allocate`, and returning `Some mode` makes it sound like it might allocate. (Perhaps my comment about the `mode` buried my main point.)

One worry: it _does_ allocate in bytecode. Does that matter here?

goldfirere:
> How about we first merge https://github.com/ocaml-flambda/flambda-backend/pull/3089 and then add some code to the type-checker that maps the overwrites to these primitives? Then we can run a few simple programs (like the unique_rbtree or unique_sort from the old PR) to test this code.

OK. I'll focus there and then we can return to this one.

ocaml/lambda/lambda.mli
anfelor: As just discussed in our meeting, we want to keep all this information in since it will make the implementation in the backend easier.

ocaml/lambda/lambda.ml
anfelor: It would be sound to keep the `Some mode` here, since `primitive_may_allocate` is only used in `transl_primitive` where this value causes strictly more checks. Since we have decided to map `Preuseblock` to `Pmakeblock` in the backend for the moment, I think we should stick with this code for now. However, I will add a CR to make all of these return `None` once the backend supports actual overwriting. The bytecode interpreter does not use stack allocation so this check is irrelevant there.

ocaml/lambda/lambda.mli
anfelor: Since Mark seemed happy with the current design, I would propose we keep the representation as it is now.

ocaml/lambda/lambda.mli
```diff
@@ -334,6 +334,41 @@ type primitive =
      handlers, finalizers, memprof callbacks, etc, as well as GCs and
      GC slices, so should not be moved or optimised away. *)
   | Ppoll
+    (* In-place reuse of heap blocks.
+     Each of these takes the allocation to be reused as first argument.
+     The following arguments correspond to the fields that are Reuse_set. *)
+  | Preuseblock of
```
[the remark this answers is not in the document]
anfelor: Oh, right, that is the intended meaning. If `Preuseblock` was a function, we should have: `Preuseblock : memory cell -> overwritten args -> new block`. You should not refer to the overwritten memory cell after it was passed to `Preuseblock`. In fact, referring to the overwritten memory cell again would break lots of properties since the overwrite may change its type.

anfelor:
I have rebase the changes in this PR onto main and created a draft PR at https://github.com/ocaml-flambda/flambda-backend/pull/3350

PR #3108, opened 2024-10-02
Rename Lambda.alloc_mode to Lambda.locality_mode (no semantic change)

anfelor:
Following up on #3056 and preparing for adding uniqueness information to `Lambda.alloc_mode`, this PR fully removes all dependency on `Lambda.alloc_mode` and instead prefers `Lambda.locality_mode`. This will make it easy for us to see where uniqueness information is added to the IR once we add a `uniqueness` field to `Lambda.alloc_mode` in a separate PR.

Note that I made this PR against `main` rather than `overwriting`. There are three reasons for this:
 - #3056 was merged into main
 - This change touches many parts of the codebase which might change over time and make our merge of overwriting harder
 - The diff looks much cleaner since this relies on #3056 which is newer than the `overwriting` branch.

### DIFF
--- a/middle_end/flambda2/from_lambda/closure_conversion.ml
+++ b/middle_end/flambda2/from_lambda/closure_conversion.ml
@@ -556,7 +556,7 @@ let close_c_call acc env ~loc ~let_bound_ids_with_kinds
         k acc (List.map Named.create_var let_bound_vars))
   in
   let alloc_mode_app =
-    match Lambda.alloc_mode_of_primitive_description prim_desc with
+    match Lambda.locality_mode_of_primitive_description prim_desc with
     | None ->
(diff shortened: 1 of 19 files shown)


mshinwell:
To be sure I understand: the plan is to actually change some of these occurrences back to `Lambda.alloc_mode` once that is equipped with the uniqueness information as well, right?

The flambda2 parts of this are ok.

anfelor:
Yes, we want to change the `locality_mode` on `Pmake(.*)block`s back to be an `alloc_mode`; I have started working on this at https://github.com/anfelor/flambda-backend/pull/1. By making this change in two separate PRs, we can easily spot all the places where we add uniqueness to the IR.

goldfirere (approved):

PR #3110, opened 2024-10-03
Unboxed products in externals - flambda2

mshinwell:
This builds on #3092 to provide the necessary support in flambda2 for `external` declarations involving unboxed products in parameter and/or result types.  This fixes the calling convention in one way: such products are flattened in the obvious order, without reordering.  The remainder of the calling convention (which we intend to have match up with the x86-64/arm64 ABIs for return values from externals; the existing OCaml calling convention will be used for parameters) will be fixed by register assignments in the backend.

### DIFF
--- a/middle_end/flambda2/from_lambda/closure_conversion.ml
+++ b/middle_end/flambda2/from_lambda/closure_conversion.ml
@@ -767,17 +767,16 @@ let close_c_call acc env ~loc ~let_bound_ids_with_kinds
       let acc, body = keep_body acc in
       List.fold_left2
         (fun (acc, body) unarized_param (handler_param, let_bound_var') ->
-          match unarized_param.return_transformer with
-          | None -> acc, body
-          | Some return_transformer ->
-            let named =
-              Named.create_prim
-                (Unary (return_transformer, Simple.var handler_param))
-                dbg
-            in
-            Let_with_acc.create acc
-              (Bound_pattern.singleton let_bound_var')
-              named ~body)
+          let named =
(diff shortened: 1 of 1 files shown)


mshinwell:
(Another PR will follow for the backend change, @xclerc is looking at it)

middle_end/flambda2/from_lambda/closure_conversion.ml
```diff
@@ -697,27 +752,42 @@ let close_c_call acc env ~loc ~let_bound_ids_with_kinds
...
-          (Unary (box_return_value, Simple.var handler_param))
-          dbg
-      in
-      Let_with_acc.create acc
-        (Bound_pattern.singleton let_bound_var')
-        named ~body
+      List.fold_left2
+        (fun (acc, body) unarized_param (handler_param, let_bound_var') ->
+          match unarized_param.return_transformer with
+          | None -> acc, body
```
Ekdohibs, line 771: I think we need to bind `let_bound_var'` to `handler_param` here, or ensure we don't perform a rename of this `handler_param` in that case.

Ekdohibs (commented):
One small bug to fix but otherwise good. Maybe we should add a test returning an unboxed product, where one component needs a transformer and the other does not.

middle_end/flambda2/from_lambda/closure_conversion.ml
mshinwell: Well spotted, fixed.  I think we're going to add tests in the subsequent PR but will confirm.

PR #3115, opened 2024-10-04
Use pull_request_target for compatibility with forks

goldfirere:
@anfelor discovered that the old approach didn't work from a fork. This one should.


ncik-roberts (approved):
From talking to @anfelor it sounds like it was tested somewhere.

PR #3118, opened 2024-10-07
Add directory field to Debuginfo

gretay-js:
Save the value of `-directory` flag in `Debuginfo`.  It will only be used for printing error messages and debug info. 

1) Add `dinfo_dir` field to `Debuginfo.item` 
    This may increase cmx size:
    - The string for the current compilation's directory is shared by all `Debuginfo`. 
    - Debuginfo in code inlined from other libraries will have some duplication (need to measure). 
    Size increase is not expected to be significant, even in optimized builds.

2) Refactor printing of `Debuginfo` to use the new field to print directory. Existing `Debuginfo.print_compact` and `Debuginfo.print_compact_extended` are not affected.

(Prereq. for https://github.com/ocaml-flambda/flambda-backend/pull/3104)

### DIFF
--- a/lambda/debuginfo.ml
+++ b/lambda/debuginfo.ml
@@ -369,54 +369,44 @@ let compare { dbg = dbg1; assume_zero_alloc = a1; }
   let res = Dbg.compare dbg1 dbg2 in
   if res <> 0 then res else ZA.Assume_info.compare a1 a2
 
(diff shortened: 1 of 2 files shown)


ccasin (approved):
Looks good.  Seems fine to rely on normal roll-out process for evaluating cmx size impact.

PR #3122, opened 2024-10-07
Fix bug around coerce and modes

goldfirere:
This fixes a bug introduced in #3064.

Review suggestion: @riaqn

### DIFF
--- a/ocaml/testsuite/tests/typing-modes/coerce.ml
+++ b/ocaml/testsuite/tests/typing-modes/coerce.ml
@@ -1,0 +1,21 @@
+(* TEST
+   expect;
+*)
+
+(* We must put this in a module to stop the top-level from squeezing out
+   unconstrained mode variables *)
+
+module M = struct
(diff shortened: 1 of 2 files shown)


riaqn (approved):
LGTM - thanks.

Just to better understand: the issue was that `close_type_expr` removes mode variables but it shouldn't? It's called by `Typecore.type_coerce`.

goldfirere:
Yes. The `closed_type_expr` function is new. Previous to #3064, `type_coerce` called `free_variables` and checked that the returned list was empty. That's a bit inefficient, accumulating up a list of variables just to check for emptiness. Because I had to refactor `free_variables` for the l/r kinds (I wanted to return the lr-kinds in variables, instead of just returning a list of types from which I could get only an l-kind with `estimate_type_jkind`) anyway, I made this little improvement. But I missed the fact that `closed_type` also removed variables. This new patch separates the removing of variables from checking them.

PR #3129, opened 2024-10-10
Keep unused let-bindings for Simplify

mshinwell:
Currently, unused let-bindings that have no effects are deleted during closure conversion.  This inhibits Simplify from being able to turn them into `Invalid`.  For example:

```
external unsafe_get : 'a iarray -> int -> 'a
  = "%array_unsafe_get"

type ('a, 'b) t = Refl : ('a, 'a) t

let magic (type a b) (x : a) : b =
  let arr : (a, b) t iarray = [: :] in
  match unsafe_get arr 0 with
  | Refl -> x
```

This patch prevents this deletion if Simplify is going to be run.  Given that the majority of bindings are used, this should not produce any noticeable compile time changes.

(The above example still does not simplify to `Invalid` even with this patch, there will be a follow-up PR.)


lthls (approved):
I'm not overly fond of the double negation, but it's clear enough this way and I don't have anything obviously better to propose.

PR #3130, opened 2024-10-10
Refactoring of Flambda 2 array types

mshinwell:
This rearranges the representation of array types in the Flambda 2 type grammar a little.  It then exposes these directly through the provers rather than providing only a subset of information.  This will help to tighten up the array simplification functions to ensure that examples such as in https://github.com/ocaml-flambda/flambda-backend/pull/3129 produce `Invalid`.

### DIFF
--- a/middle_end/flambda2/simplify/simplify_binary_primitive.ml
+++ b/middle_end/flambda2/simplify/simplify_binary_primitive.ml
@@ -964,14 +964,26 @@ let simplify_array_load (array_kind : P.Array_kind.t)
     in
     (* CR mshinwell/vlaviron: if immutable array accesses were consistently
        setting [mutability] to [Immutable], we could restrict the following code
-       to immutable loads only and use [T.meet_is_immutable_array] instead. *)
-    match T.prove_is_immutable_array (DA.typing_env dacc) array_ty with
-    | Unknown -> contents_unknown ()
-    | Proved (elt_kind, fields, _mode) -> (
-      match elt_kind with
+       to immutable loads only and use [T.meet_is_array] instead. *)
+    match T.prove_is_array (DA.typing_env dacc) array_ty with
+    | Unknown
+    | Proved
+        { element_kind = _;
+          length = _;
+          contents = Known Mutable | Unknown;
+          alloc_mode = _
+        } ->
+      contents_unknown ()
+    | Proved
+        { element_kind;
+          length = _;
+          contents = Known (Immutable { fields });
+          alloc_mode = _
+        } -> (
+      match element_kind with
       | Unknown | Bottom -> contents_unknown ()
-      | Ok elt_kind -> (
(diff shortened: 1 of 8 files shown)


middle_end/flambda2/types/flambda2_types.ml
```diff
@@ -45,6 +45,22 @@ type typing_env = Typing_env.t
...
+
+module Array_type = struct
+  type nonrec t = array_type =
+    { element_kind : Flambda_kind.With_subkind.t Or_unknown_or_bottom.t;
+      length : t;
+      contents : array_contents Or_unknown.t;
+      alloc_mode : Alloc_mode.For_types.t
+    }
+end
+
```
lthls, line 63: I feel like these modules would fit better in `Type_grammar` directly.

middle_end/flambda2/simplify/simplify_unary_primitive.ml
```diff
@@ -657,12 +657,23 @@ let simplify_duplicate_array ~kind:_ ~(source_mutability : Mutability.t)
...
+    match T.meet_is_array (DA.typing_env dacc) arg_ty with
     | Invalid -> SPR.create_invalid dacc
-    | Need_meet ->
+    | Need_meet
+    | Known_result
+        { element_kind = _;
+          length = _;
+          contents = Known Mutable | Unknown;
+          alloc_mode = _
+        } ->
```
lthls, line 668: We could put the `Known Mutable` case with `Invalid` (this is what happened before).
Although I don't think we actually expose `Pduparray`, so this case likely never occurs at all.

middle_end/flambda2/simplify/simplify_binary_primitive.ml
```diff
@@ -964,14 +964,26 @@ let simplify_array_load (array_kind : P.Array_kind.t)
     in
     (* CR mshinwell/vlaviron: if immutable array accesses were consistently
        setting [mutability] to [Immutable], we could restrict the following code
-       to immutable loads only and use [T.meet_is_immutable_array] instead. *)
-    match T.prove_is_immutable_array (DA.typing_env dacc) array_ty with
-    | Unknown -> contents_unknown ()
-    | Proved (elt_kind, fields, _mode) -> (
-      match elt_kind with
+       to immutable loads only and use [T.meet_is_array] instead. *)
```
lthls, line 967: I think this comment needs to be updated. First, as discussed on Slack, we can use `T.meet_is_array` now, even for mutable loads.
The rest of the comment explains why it would be problematic to return `Invalid` on mismatched mutability (between the primitive and the type), so I propose to update it this way:
```ocaml
(* The frontend does not distinguish between mutable and immutable primitives yet,
   so we cannot assume that [mutability] must match the type of the array. *)
```

lthls (commented):
The PR looks correct, I wouldn't mind a bit of cleaning up before merging.

PR #3135, opened 2024-10-11
Fix printing of user-defined option type used for optional arguments

riaqn:
What the titles says. Please review by commits.

### DIFF
--- a/ocaml/typing/printtyp.ml
+++ b/ocaml/typing/printtyp.ml
@@ -1435,7 +1435,9 @@ let rec tree_of_typexp mode alloc_mode ty =
         let arg_mode = Alloc.zap_to_legacy marg in
         let t1 =
           if is_optional l then
-            match get_desc (tpoly_get_mono ty1) with
+            match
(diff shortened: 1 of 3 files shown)


ocaml/typing/printtyp.ml
```diff
@@ -1435,7 +1435,9 @@ let rec tree_of_typexp mode alloc_mode ty =
         let arg_mode = Alloc.zap_to_legacy marg in
         let t1 =
           if is_optional l then
-            match get_desc (tpoly_get_mono ty1) with
+            match
+              get_desc (Ctype.expand_head !printing_env (tpoly_get_mono ty1))
```
riaqn, line 1439: should we keep the original code as quick-path?

lpw25: I think you probably don't need it. In the common case this is basically just a single look up for the `option` path in the environment.

lpw25 (approved):
Not sure when this started being necessary, but this seems more robust and I think it should work fine.

PR #3140, opened 2024-10-14
Zero alloc: new payload "assume_unless_opt"

gretay-js:
On top of https://github.com/ocaml-flambda/flambda-backend/pull/3134.

The meaning of the new payload depends on the value of `-zero-alloc-check` flag (or the corresponding top-level `[@@@zero_alloc check]` attribute). 

- If checking of `opt` annotations is disabled, then `[@zero_alloc assume_unless_opt]` is treated as "[@zero_alloc assume]". 
- If checking `opt` annotations is enabled, then `[@zero_alloc assume_unless_opt]` is treated as `[@zero_alloc]` on function definitions. On applications, `[@zero_alloc assume_unless_opt]` is ignored, which corresponds to checking the callee as usual.
- The new payload is not allowed on signatures.

The intended use of this annotation is on function definitions that can only be shown as zero_alloc in an optimized build. As such, it is not intended to be inferred from the signatures.

Additional payloads (such as `strict` and `never_returns_normally`) are not supported.

The implementation is entirely during parsing, the new payload is interpreted as one of the the existing `zero_alloc` constructors.

### DIFF
--- a/flambda-backend/tests/backend/zero_alloc_checker/dune
+++ b/flambda-backend/tests/backend/zero_alloc_checker/dune
@@ -51,6 +51,14 @@
  (alias  runtest)
  (action (copy fail29_opt.mli fail29_opt2.mli )))
 
+(rule
+ (alias  runtest)
+ (action (copy test_assume_unless_opt_sig.ml test_assume_unless_opt_sig2.ml  )))
+
+(rule
+ (alias  runtest)
+ (action (copy test_assume_unless_opt_on_call.ml test_assume_unless_opt_on_call2.ml  )))
+
 ;; Tests whose outputs differ depending on stack_allocation configuration flag.
 ;; This condition is not expressible in "enable_if" clause
 ;; because dune does not support %{config:stack_allocation} yet.
--- a/flambda-backend/tests/backend/zero_alloc_checker/dune.inc
+++ b/flambda-backend/tests/backend/zero_alloc_checker/dune.inc
@@ -1311,7 +1311,7 @@
  (action (diff fail29_opt2.output fail29_opt2.output.corrected)))
 
 (rule
- (enabled_if (= %{context_name} "main"))
+ (enabled_if (and (= %{context_name} "main") (= %{architecture} "amd64")) )
  (targets test_assume_unless_opt.output.corrected)
  (deps (:ml test_assume_unless_opt.ml) filter.sh)
  (action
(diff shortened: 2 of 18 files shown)


ocaml/parsing/builtin_attributes.ml
```diff
@@ -935,20 +953,49 @@ let parse_zero_alloc_attribute ~is_arity_allowed ~default_arity attr =
...
+             else
+               (* Treat [@zero_alloc assume_unless_opt] as [@zero_alloc],
+                  forcing the function to be checked.
+                  Setting [opt = false] to satisfy [@zero_alloc]
+                  and not only [@zero_alloc opt] on the corresponding signatures. *)
+               empty arity)
+          else
+            (* Treat "assume_unless_opt" as "assume".
+               Reuse standard parsing for better error messages. *)
+            parse payload
```
ccasin, line 992: The logic here is subtle. I spent a while thinking this was wrong, because we reach this case if the opt check is enabled but `no_other_payload` is false, and the comment says we're going to treat `assume_unless_opt` as assume in that case even though the opt check is on. If I understand correctly, it's actually fine, because the table has no entries for `assume_unless_opt` that include other things, so actually the table lookup will fail and we'll get default, not assume.  

This seems a bit error prone - if we decide in the future to support `assume_unless_opt` in combination with some other modifier, we then have a subtle bug here.  I think it would be better to just make the behavior explicit by returning default and calling `warn` when `no_other_payload` is true.

ocaml/parsing/builtin_attributes.ml
```diff
@@ -935,20 +953,49 @@ let parse_zero_alloc_attribute ~is_arity_allowed ~default_arity attr =
...
+          (warn_payload loc txt
+             "The payload \"assume_unless_opt\" is not supported \
+              in signatures.";
+           (* Treat [@zero_alloc assume_unless_opt] as [@zero_alloc] in signatures. *)
+           parse rest)
+        else
+          let no_other_payload = List.compare_length_with rest 0 = 0 in
+          if is_zero_alloc_check_enabled ~opt:true && no_other_payload then
+            (if on_application then
+               (* Treat is if there is no attribute.
```
ccasin, line 980:
```suggestion
               (* Treat as if there is no attribute.
```

ccasin (commented):
Reviewed only the last four commits.  Looks good, but some suggestions below.

ocaml/parsing/builtin_attributes.ml
```diff
@@ -935,20 +953,49 @@ let parse_zero_alloc_attribute ~is_arity_allowed ~default_arity attr =
...
+      | None -> parse payload
+      | Some rest ->
+        if in_signature then
+          (warn_payload loc txt
+             "The payload \"assume_unless_opt\" is not supported \
+              in signatures.";
+           (* Treat [@zero_alloc assume_unless_opt] as [@zero_alloc] in signatures. *)
+           parse rest)
+        else
+          let no_other_payload = List.compare_length_with rest 0 = 0 in
```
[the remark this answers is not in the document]
gretay-js: We don't have `List.is_empty` in the "boot" context when building with `4.14.*` as the system compiler.

ocaml/parsing/builtin_attributes.ml
gretay-js: Done. (It's easier now thanks to the refactoring you did in #3134).

ccasin:
(Looks good to merge once #3134 is merged)

ccasin (approved):

PR #3141, opened 2024-10-14
Added the update level operation

ageorg29:
Added the `update_level` operation, which lowers the level of a variable. 

When changing the level, the mode variable invariant needs to be maintained: 
- relations are moved from vupper/vlower to maintain level invariant
- bounds are updated accordingly
- relations are added to maintain transitivity

### DIFF
--- a/ocaml/typing/solver.ml
+++ b/ocaml/typing/solver.ml
@@ -53,7 +53,7 @@ let rec find_error (f : 'x -> ('a, 'b) Result.t) : 'x list -> ('a, 'b) Result.t
 module Solver_mono (C : Lattices_mono) = struct
 
   type 'a var =
-    { level : int;
+    { mutable level : int;
       (** The level of the variable. This has the same meaning as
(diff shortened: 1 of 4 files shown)


ocaml/typing/solver.ml
```diff
@@ -520,8 +648,35 @@ module Solver_mono (C : Lattices_mono) = struct
...
+          update_level_v ~log obj level v)
+        mvs
+    | Amodemeet (_, mvs) ->
+      List.iter
+        (fun (Amorphvar (v, f)) ->
+          let obj = C.src obj f in
+          update_level_v ~log obj level v)
+        mvs
+
+  let cnt_debug = ref 0
```
lpw25, line 670: You should probably take this testing stuff out before we merge.

lpw25 (commented):
Looks good. A few small remarks.

lpw25 (approved):
LGTM

PR #3142, opened 2024-10-14
Squelch quotes with -error-style short

goldfirere:
As tin.

This is pretty gross. But it works.


goldfirere:
Closed in favor of #3146

PR #3143, opened 2024-10-15
Zero alloc: custom error message

gretay-js:
Specify a custom error message for `zero_alloc` assertions using a new payload of the form:
 `[@zero_alloc custom_error_message "This sentence will be appended to the default error message."]`.

This is useful for PPXes that generate `zero_alloc` annotations, such as [ppx_fields_conv](https://github.com/janestreet/ppx_fields_conv). 

There is some risk that these long string literals increase memory usage during compilation, especially when a PPX generates many copies of them. It's not clear to me if they would be shared. The overhead is one string per function symbol annotated with `zero_alloc`.

The PR is mostly mechanical propagating the new field all the way to the backend. I suggest to review by commit. 

- The new payload can be used in signatures too. 
- The new payload is not compatible with `assume`. 
- Parsing of the new payload is similar to `arity`. There is a small refactoring of parsing of arity and an improved error message for it.


typing/zero_alloc.ml
```diff
@@ -161,23 +163,35 @@ let sub_const_const_exn za1 za2 =
...
+      match msg1, msg2 with
+      | None, None -> msg1, false;
+      | None, Some _ -> msg2, true;
+      | Some _, None -> msg1, false;
+      | Some m1, Some m2 ->
+        let b = String.equal m1 m2 in
+        let msg =
+          if b then msg1 else Some (String.concat "\n" [m1; m2])
+        in
+        msg, b
```
ccasin, line 190: I think `b` here is flipped - it's true if we're not changing the message, but should be true if we are.  Here is a program you can add to the testsuite that should demonstrate the concatenated error message once this is fixed:
```ocaml
module type S1 = sig
  val f : 'a -> ('a * 'a) [@@zero_alloc custom_error_message "foo"]
end

module type S2 = sig
  val f : 'a -> ('a * 'a) [@@zero_alloc custom_error_message "bar"]
end

module M = struct
  let f x = (x,x)
end

module M' : S1 = M

module M'' : S2 = M
```

ccasin (approved):
Looks good, just a few minor things.

typing/zero_alloc.ml
ccasin: I guess one other point to make about this:  Because this concatenation behavior only happens in the case where the thing on the left is a variable, we'll often end up throwing away the string.  For example, if the custom error message appears in a signature but there is also a zero_alloc annotation in the structure (with or without a custom error message), we'll just throw away the string.

I think we could fix this by making the `custom_error_message` field itself a ref and just mutating that.  But you don't necessarily have to do that in this first version.

gretay-js: > I think b here is flipped - it's true if we're not changing the message, but should be true if we are.
thanks!! fixed.

I added the test you suggested to `typing-zero-alloc/signatures` but the `%expect` output it's not showing the custom error on `f`. I know the custom error gets propagated correctly the backend because when the new test is compiled, the zero alloc check fails with the combined string in the output.

PR #3148, opened 2024-10-16
Rename fields in `typedtree.signature` from `sig_` to `sg_`

riaqn:
EDIT: Changed approach. See comments below.

This PR renames `sig_` to `sigi_` for `typedtree.signature_item`; similar for `structure_item`.

# Motivation

Very soon we will have a field `psig_loc` in `parsetree.signature`, but there is already a field of this name in `parsetree.signature_item`. To avoid clashes, we will rename the second one to `psigi_loc` where `i` stands for "item". Then, for naming consistency between `parsetree` and `typedtree`, we should rename the `sig_*` fields in `typedtree.signature_item` as well. 

# Review

Please only review `typedtree.mli`. The rest is mechanical and boring.

### DIFF
--- a/ocaml/driver/compile_common.ml
+++ b/ocaml/driver/compile_common.ml
@@ -70,7 +70,7 @@ let typecheck_intf info ast =
     |> print_if info.ppf_dump Clflags.dump_typedtree Printtyped.interface
   in
   let alerts = Builtin_attributes.alerts_of_sig ~mark:true ast in
-  let sg = tsg.Typedtree.sig_type in
+  let sg = tsg.Typedtree.sg_type in
   if !Clflags.print_types then
     Printtyp.wrap_printing_env ~error:false info.env (fun () ->
         Format.(fprintf std_formatter) "%a@."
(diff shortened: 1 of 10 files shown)


lpw25:
I'm surprised that we renamed the exising one rather than picking a different name for the new one.

riaqn:
@lpw25 Do you mean we should rename `sig_` to `sg_` for `typedtree.signature` instead?

riaqn:
As discussed offline, this PR now renames `sig_*` in `typedtree.signature` to `sg_*`. @lpw25 could you look again?

goldfirere:
I don't love this change. It will introduce merge conflicts with upstream. I don't think the consistency in naming between parsetree and typedtree is worth introducing these conflicts.

riaqn:
I would hope to preserve the consistency naming between typedtree and parsetree, but I also understand your concern. This is orthogonal to https://github.com/ocaml-flambda/flambda-backend/pull/3049 so not blocking. Let's wait for other people's opinions.

riaqn:
It seems there aren't enough motivation to do this. Closing.

PR #3149, opened 2024-10-16
Move `backend/asm*` into `asmcomp/`

lukemaurer:
We're preparing to move upstream code out of `ocaml/` and realign this repo with the upstream directory structure to make dealing with merges and history easier. As part of this, we want to have files that we intend to keep in sync with upstream at the same path they have upstream. This includes driver code like `backend/asmgen.ml{,i}`, but not _true_ backend code, which has more of a clean break with upstream. With these moves, the files we consider upstreamable will get conflict markers as usual when merging from upstream, and the files we consider hard-forked will be untouched (producing modify/delete conflicts, which we can resolve trivially).


mshinwell (approved):

mshinwell:
@lukemaurer actually you need to make `asmcomp/.ocamlformat-enable` then move the `asmgen`-related lines from `backend/.ocamlformat-enable` into this one.  The `make fmt` rule and the CI check might need adjusting to look at `asmcomp/` too.

stedolan:
This change looks good to me. I like that `asmcomp` now means "the compiler modules that start with `asm`".

lukemaurer:
> @lukemaurer actually you need to make `asmcomp/.ocamlformat-enable` then move the `asmgen`-related lines from `backend/.ocamlformat-enable` into this one. The `make fmt` rule and the CI check might need adjusting to look at `asmcomp/` too.

Pushed a commit that does this (and adds `asmcomp/.ocamlformat` as well). Didn't need to update `make fmt` or CI since they operate on all directories nowadays

mshinwell (approved):

PR #3150, opened 2024-10-16
Added instantiate to mode solver

ageorg29:
Added the `instantiate` operation, which copies all (reachable) mode variables at the generic level. The copy is created by updating a new `optcopy` field in mode variables, which is set back to its original value once copying is done,

### DIFF
--- a/ocaml/typing/mode_intf.mli
+++ b/ocaml/typing/mode_intf.mli
@@ -114,6 +114,10 @@ module type S = sig
 
   val set_append_changes : (changes ref -> unit) -> unit
 
+  type copy_scope
+
+  val with_copy_scope : (copy_scope -> 'a) -> 'a
+
   type nonrec allowed = allowed
 
   type nonrec disallowed = disallowed
@@ -477,6 +481,24 @@ module type S = sig
 
     (** Returns the lower bound needed for [B -> C] in relation to [A -> B -> C] *)
     val partial_apply : (allowed * 'r) t -> l
+
+    val instantiate :
+      copy_scope:copy_scope ->
+      current_level:int ->
+      generic_level:int ->
+      ('l * 'r) t ->
+      ('l * 'r) t
+
+    val copy_generic :
+      copy_scope:copy_scope ->
+      generic_level:int ->
+      ('l * 'r) t ->
+      ('l * 'r) t
+
+    val duplicate :
+      copy_scope:copy_scope ->
+      ('l * 'r) t ->
+      ('l * 'r) t
   end
 
   (** The most general mode. Used in most type checking,
--- a/ocaml/typing/solver.ml
+++ b/ocaml/typing/solver.ml
@@ -93,6 +93,9 @@ module Solver_mono (C : Lattices_mono) = struct
          - for all [f u \in v.vlower] and [g w \in v.vupper] we
            have either [g'f \in w.vlower] or [f'g \in u.vupper]. *)
       id : int;  (** For identification/printing *)
+      mutable subst : 'a var option
+      (** Similar to Tsubst in a type, the [subst] field holds an optional copy used
+          during instantiation *)
     }
 
   and 'b lmorphvar = ('b, left_only) morphvar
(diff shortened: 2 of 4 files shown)


ocaml/typing/solver.ml
```diff
@@ -807,7 +836,107 @@ module Solver_mono (C : Lattices_mono) = struct
...
+            (fun (Amorphvar (u, f)) ->
+              let src = C.src obj f in
+              let ucopy = copy_under_condition_v ~copylog cond ?current_level src u in
+              (Amorphvar (ucopy, f)))
+            v.vlower
+          in
+          copy.level <- v.level;
+          copy.vupper <- vupper;
+          copy.vlower <- vlower;
+          Option.iter (fun level -> update_level_v ~log:None obj level copy) current_level;
```
lpw25, line 873: The semantics of `current_level` as a parameter appear to be that it places an optional upper bound on the level. This makes sense in the `copy_generic` case, but it's not obvious it's the right semantics for the `copy_all` case. Which values of `current_level` do we actually end up using for `copy_all` once we start using it in the type-checker?

lpw25 (commented):
Looks pretty good. A few comments.

ocaml/typing/mode_intf.mli
```diff
@@ -88,6 +92,19 @@ module type Common = sig
     ('l * 'r) t ->
     unit
 
+  val copy_generic :
+    copylog:mode_copy_scope ->
```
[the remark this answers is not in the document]
ageorg29: Changed it throughout to copy_scope

ocaml/typing/solver.ml
ageorg29: Indeed, copy_all should take `None` for the optional `current_level` parameter. I changed the exposed signatures in `mode_intf.mli` to enforce this.

ocaml/typing/solver.ml
```diff
@@ -128,6 +131,26 @@ module Solver_mono (C : Lattices_mono) = struct
   [l1]. *)
   let append_changes l0 l1 = l1 @ l0
 
+  type copy_change =
+    | Coptcopy : 'a var * 'a var option -> copy_change
```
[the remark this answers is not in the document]
ageorg29: This felt more "future proof" if we are using the field for different purposes that just copying

ocaml/typing/mode_intf.mli
```diff
@@ -88,6 +92,19 @@ module type Common = sig
...
+    ?current_level:int ->
```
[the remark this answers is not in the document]
ageorg29: Default is `None`, and means the copy's level is not changed

ocaml/typing/solver.ml
```diff
@@ -716,37 +727,57 @@ module Solver_mono (C : Lattices_mono) = struct
       generalize_topology ~log dst ~current_level ~generic_level u;
       update_level_v ~log dst generic_level u
 
-  type varcond = { b : 'a. 'a var -> bool; } [@@unboxed]
+  let copy_exists : type a. ?copy_to_level: int -> a C.obj -> a var -> (a var) option =
+    fun ?copy_to_level obj v ->
+      Option.bind copy_to_level (fun level ->
+        match v.vlower, v.vupper with
+        | Amorphvar (u1, f) :: _, Amorphvar (u2, _) :: _ ->
```
lpw25, line 734: This also needs to check that the morphisms are both the identity morphism. Possibly we need to ask for an `is_identity` function on morphisms to do this if we don't already.

ocaml/typing/solver_intf.mli
```diff
@@ -296,13 +298,25 @@ module type Solver_polarized = sig
...
+  val generalize_structure :
+    current_level:int ->
+    generic_level:int ->
+    'a obj ->
+    ('a, 'l * 'r) mode ->
+    log:changes ref option ->
+    unit
+
+  (** Copies all reachable variables whose level is at or above [copy_from_level],
+      and optionally puts the level of the copy to [copy_to_level].
```
lpw25, line 310: I think "puts" here is ambiguous, it would be to explicitly say that the level is forced to be less than or equal to [copy_to_level] to make clear that if something already has a level lower than that then it is left alone.

ocaml/typing/solver_intf.mli
```diff
@@ -296,13 +298,25 @@ module type Solver_polarized = sig
...
+      If no [copy_to_level] is given, the copy will have the same level as the original.
+      Returns the copied mode *)
+  val copy :
+    copy_scope:copy_scope ->
+    copy_from_level:int ->
+    ?copy_to_level:int ->
```
lpw25, line 316: Rather than have `copy_to_level` be optional, I think you can just pass `generic_level` or even `Sys.max_int` as it's value when using it to implement `duplicate`.

lpw25 (commented):
I few small comments then this is good to go.

PR #3151, opened 2024-10-16
Fix unsoundness that locks beyond `open` are ignored

riaqn:
Currently, the locks between `open M` and the definition of `M` are ignored, which is unsound. For example:
```
module M_nonportable = struct
    let f @ nonportable = fun () -> ()
end

let (foo @ portable) () =
    let open M_nonportable in
    let _ = f in
    ()
```
This should be rejected as `f` is `nonportable`. But currently this is accepted.

### DIFF
--- a/ocaml/testsuite/tests/typing-modes/val_modalities.ml
+++ b/ocaml/testsuite/tests/typing-modes/val_modalities.ml
@@ -775,3 +775,17 @@ let (_foo @ portable) () =
 [%%expect{|
 val _foo : unit -> unit @@ global many = <fun>
 |}]
+
+let () =
+  let open M_nonportable in
+  let (foo @ portable) () =
+    let _ = f in
+    ()
+  in
+  ()
+[%%expect{|
+Line 4, characters 12-13:
+4 |     let _ = f in
+                ^
+Error: The value "f" is nonportable, so cannot be used inside a function that is portable.
+|}]
--- a/ocaml/testsuite/tests/typing-modes/val_modalities.ml
+++ b/ocaml/testsuite/tests/typing-modes/val_modalities.ml
@@ -761,7 +761,10 @@ let (foo @ portable) () =
     let _ = f in
     ()
 [%%expect{|
-val foo : unit -> unit @@ global many = <fun>
+Line 3, characters 12-13:
+3 |     let _ = f in
+                ^
+Error: The value "f" is nonportable, so cannot be used inside a function that is portable.
 |}]
 
 let (_foo @ portable) () =
--- a/ocaml/testsuite/tests/typing-modes/val_modalities.ml
+++ b/ocaml/testsuite/tests/typing-modes/val_modalities.ml
@@ -741,3 +741,34 @@ module type S' =
       end
   end
 |}]
+
+
+(* interaction between open and locks *)
+module M_nonportable = struct
+    let f @ nonportable = fun () -> ()
+end
+
+module M_portable = struct
+    let f @ portable = fun () -> ()
+    end
+[%%expect{|
+module M_nonportable : sig val f : unit -> unit @@ global many end
+module M_portable : sig val f : unit -> unit @@ global many portable end
+|}]
+
+let (foo @ portable) () =
+    let open M_nonportable in
+    let _ = f in
+    ()
+[%%expect{|
+val foo : unit -> unit @@ global many = <fun>
+|}]
+
+let (_foo @ portable) () =
+    let open M_portable in
+    let _ = f in
+    ()
+
+[%%expect{|
+val _foo : unit -> unit @@ global many = <fun>
+|}]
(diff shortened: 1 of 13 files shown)


ocaml/typing/env.mli
```diff
@@ -36,7 +36,7 @@ type summary =
   | Env_modtype of summary * Ident.t * modtype_declaration
   | Env_class of summary * Ident.t * class_declaration
   | Env_cltype of summary * Ident.t * class_type_declaration
-  | Env_open of summary * Path.t
+  | Env_open of summary * Longident.t
```
lpw25, line 39: I don't like that the environment you get out will not always be exactly the same now if e.g. the available `cmi` files have changed. I'd probably prefer that this stay as a `Path.t` and include the locks data.

Also, it loks like we're not storing locks in this `summary`, which is definitely a (preexisting) bug.

ocaml/typing/env.mli
```diff
@@ -407,14 +407,14 @@ val add_signature_lazy: Subst.Lazy.signature_item list -> t -> t
...
-    ?loc:Location.t -> ?toplevel:bool ->
-    Asttypes.override_flag -> Path.t ->
-    t -> (t, [`Not_found | `Functor]) result
+    used_slot:bool ref ->
+    loc:Location.t -> toplevel:bool ->
+    Asttypes.override_flag -> Longident.t Location.loc ->
+    t -> Path.t * t
 
-val open_pers_signature: string -> t -> (t, [`Not_found]) result
+val open_pers_signature: loc:Location.t -> Longident.t -> t -> Path.t * t
```
lpw25, line 415: This should probably still take a `string` since e.g. `Ldot` cannot be a persistent identifier.

ocaml/testsuite/tests/typing-modules/pr9695.ml
```diff
@@ -8,5 +8,5 @@ module Test (S : sig module type S end) (M : S.S) =
 Line 2, characters 14-15:
 2 |   struct open M (* should not succeed silently *) end
                   ^
-Error: This module is not a structure; it has type "S.S"
+Error: The module "M" is abstract, it cannot have any components
```
lpw25, line 11: This error message is worse

ocaml/testsuite/tests/typing-modes/val_modalities.ml
```diff
@@ -741,3 +741,37 @@ module type S' =
...
+let (foo @ portable) () =
+    let open M_nonportable in
+    let _ = f in
+    ()
+[%%expect{|
+Line 3, characters 12-13:
+3 |     let _ = f in
+                ^
+Error: The value "f" is nonportable, so cannot be used inside a function that is portable.
+|}]
```
lpw25, line 768: Probably worth including a test where the `open` happens before entering the closure.

lpw25 (commented):
Mostly looks good. A couple of suggestions.

ocaml/typing/env.ml
```diff
@@ -226,9 +226,9 @@ module TycompTbl =
         opened = Some {using; components; root; next};
       }
 
-    let remove_last_open rt tbl =
+    let remove_last_open tbl =
       match tbl.opened with
-      | Some {root; next; _} when Path.same rt root ->
```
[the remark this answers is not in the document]
lpw25: Ah, I see. If we address my other comment and keep the path in summary then we can put this back too.

ocaml/typing/env.mli
riaqn: Adding locks to summary is tracked by internal ticket 1778.

Without doing that, `Env_summary` ruins locks anyway, I think for now `Env_open` can just contain `Path.t` without `lock_item list`. I will leave a CR to fix it in the future.

ocaml/testsuite/tests/typing-modules/Test.ml
```diff
@@ -123,7 +123,7 @@ module F : functor (X : sig end) -> sig val x : int end
 Line 2, characters 0-3:
 2 | F.x;; (* fail *)
     ^^^
-Error: The module "F" is a functor, it cannot have any components
+Error: The module "F" is a functor, it cannot be used as a structure
```
lpw25, line 126: Now I think you are making this error worse. Would you mind making:
```ocaml
type components_reason =
  | Projection
  | Open
```
and passing it down to `lookup_structure_components` and then putting it into `No_components_abstract` and `No_components_functor`, and using it to fine tune the message.

lpw25 (approved):
One small remaining point, then good to go.

ocaml/testsuite/tests/typing-modules/Test.ml
riaqn: Done - I did it slightly differently tho, because both `Project` and `Open` are specific to structure (as opposed to functors, or generic modules), so the `reason` shouldn't go into `No_components_abstract` and `No_components_alias` (as both are for generic modules).

(I think that's in fact what you meant anyway, modulo the typo)

PR #3152, opened 2024-10-16
Refactor Array_ref_kind and Array_set_kind in Lambda_to_flambda

mshinwell:
This implements a CR that was added in #3072.  A similar change was made in the original #3068, but this is neater.

### DIFF
--- a/middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
+++ b/middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
@@ -198,7 +198,8 @@ let convert_array_ref_kind_to_array_kind (array_ref_kind : Array_ref_kind.t) :
     match nfo with
     | Values -> Values
     | Immediates -> Immediates
-    | Naked_floats | Naked_float32s -> Naked_float32s
+    | Naked_floats -> Naked_floats
+    | Naked_float32s -> Naked_float32s
     | Naked_int32s -> Naked_int32s
     | Naked_int64s -> Naked_int64s
     | Naked_nativeints -> Naked_nativeints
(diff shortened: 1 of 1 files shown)


middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
```diff
@@ -872,25 +900,27 @@ let array_load_unsafe ~array ~index ~(mut : Lambda.mutable_flag) array_kind
...
-  | Naked_int32s ->
-    [Binary (Array_load (array_kind, Naked_int32s, mut), array, index)]
-  | Naked_int64s ->
-    [Binary (Array_load (array_kind, Naked_int64s, mut), array, index)]
-  | Naked_nativeints ->
-    [Binary (Array_load (array_kind, Naked_nativeints, mut), array, index)]
-  | Naked_vec128s ->
-    [Binary (Array_load (array_kind, Naked_vec128s, mut), array, index)]
+  | No_float_array_opt nfo -> (
+    match nfo with
```
Ekdohibs, line 908: Here we can share the different cases and only map between the kinds.

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
```diff
@@ -141,18 +141,19 @@ let convert_array_kind_for_length kind : P.Array_kind_for_length.t =
...
-    | Naked_floats_to_be_boxed of L.locality_mode
     | Naked_floats
     | Naked_float32s
     | Naked_int32s
     | Naked_int64s
     | Naked_nativeints
     | Naked_vec128s
+
+  type t =
+    | No_float_array_opt of no_float_array_opt
```
Ekdohibs, line 155: Could we actually use `P.Array_load_kind.t`? And likewise with `Array_set_kind`?

mshinwell: On the unboxed products dev branch these are different, but it might be possible to combine them.  I will leave a CR for now.

Ekdohibs (approved):

PR #3153, opened 2024-10-16
Fix problems with allocation of stacks

mshinwell:
The existing code to allocate stacks, when stack checks are off, erroneously asks for memory not backed by huge pages on Linux.  In addition it can inhibit the transparent huge page coalescing algorithm (or split existing huge pages), because the guard page that gets `mprotect`ed might be in the middle of a proposed (or existing) huge page partially occupied by something else at an address just below the stack.  (We only consider 2Mb huge pages as this seems like the common case.)

This PR aims to fix these problems.  It also removes the use of `MAP_STACK`, as it seems that certain versions of the kernel might interpret this as a request to disable huge pages.

### DIFF
--- a/runtime/caml/fiber.h
+++ b/runtime/caml/fiber.h
@@ -292,13 +292,9 @@ CAMLnoret CAMLextern void caml_raise_unhandled_effect (value effect);
 value caml_make_unhandled_effect_exn (value effect);
 
 #if defined(NATIVE_CODE) && !defined(STACK_CHECKS_ENABLED)
-// mmap does not seem to guarantee it will always return a page-aligned
-// address as per some man pages from a few years ago (more recent man
-// pages seem to indicate the alignment is guaranteed), hence the last
-// part of the expression below to add an offset guaranteeing alignment
+// We can assume that mmap returns page-aligned addresses.
 #define Protected_stack_page(block, page_size) \
-  (((char*) (block)) + (page_size) + (page_size) - \
-   ((uintnat) ((char*) (block)) % (page_size)))
+  (((char*) (block)) + (page_size))
 #endif
 
 #endif /* CAML_INTERNALS */
--- a/runtime/fiber.c
+++ b/runtime/fiber.c
@@ -110,6 +110,11 @@ struct stack_info** caml_alloc_stack_cache (void)
   return stack_cache;
 }
 
+#if defined(NATIVE_CODE) && !defined(STACK_CHECKS_ENABLED)
+// See [alloc_for_stack], below.
+static const size_t stack_extra_size_for_mmap = 2 * 1024 * 1024;
+#endif
+
 Caml_inline struct stack_info* alloc_for_stack (mlsize_t wosize)
 {
 #ifdef USE_MMAP_MAP_STACK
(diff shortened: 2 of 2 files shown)


stedolan:
For the record, the reason I'm glad to get rid of `MAP_STACK` despite it being documented as a no-op is that in very recent kernel versions it's been changed to imply `VM_NOHUGEPAGE`: https://github.com/torvalds/linux/commit/c4608d1bf7c6536d1a3d233eb21e50678681564e

(Arguably you want your stacks to not be hugepages, because you generally don't use much of them. The trouble we've seen is that breaking hugepages for the stack affects adjacent heap areas, since the stack is often not 2MB aligned. Maybe we should investigate manually aligning these at some stage, but I definitely don't want the behaviour suddenly changing with kernel version upgrades)

ocaml/runtime/fiber.c
```diff
@@ -175,23 +187,19 @@ Caml_inline struct stack_info* alloc_for_stack (mlsize_t wosize)
...
+  // -------------------- <- [stack], returned from [mmap]
+  char* stack;
+  stack = mmap(NULL, len + extra_size,
+    PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS,
+    -1, 0);
+  if (stack == MAP_FAILED) {
     return NULL;
   }
 
+  struct stack_info* block = (struct stack_info*) (stack + extra_size);
```
stedolan, line 201: It looks like you're throwing away the original `stack` value, which means the `munmap` in `caml_free_stack` isn't going to do the right thing.

ocaml/runtime/fiber.c
```diff
@@ -191,19 +195,18 @@ Caml_inline struct stack_info* alloc_for_stack (mlsize_t wosize)
...
 
-  struct stack_info* block = (struct stack_info*) (stack + extra_size);
+  struct stack_info* block =
+    (struct stack_info*) (stack + stack_extra_size_for_mmap);
 
   // Note that there is no assumption here on the alignment of the return
   // value from [mmap].  See the definition of [Protected_stack_page].
   if (mprotect(Protected_stack_page(block, page_size), page_size, PROT_NONE)) {
-    munmap(block, len);
+    caml_mem_unmap(block, len);
```
stedolan, line 209: This needs to munmap `stack`, not `block`

ocaml/runtime/fiber.c
```diff
@@ -947,7 +950,9 @@ void caml_free_stack (struct stack_info* stack)
     munmap(stack, stack->size);
 #else
 #if defined(NATIVE_CODE) && !defined(STACK_CHECKS_ENABLED)
-    munmap(stack, stack->size);
+    // See [alloc_for_stack].
+    char* mmap_base = ((char *) stack) - stack_extra_size_for_mmap;
+    caml_mem_unmap(mmap_base, stack->size);
```
stedolan, line 955: Is stack->size right here?

stedolan (changes requested):

stedolan:
This looks correct to me now (but it's late and I may have missed something).

Has it been tested? You can check what it's doing by looking at `/proc/self/smaps` before and after running a systhread.

mshinwell:
Will read this over and test tomorrow.

stedolan:
What testing has this had? In particular, does the current version have the right effect on mappings (as viewed by e.g. `smaps`) when threads are created/destroyed?

mshinwell:
This has been tested on a large multi-instance program with no crashes.  I also wrote a small program to inspect `smaps` before and after thread creation.  Here is an example of the situation after thread creation:
```
0x7fffb14f7000 = value returned by `caml_mem_map` for the thread stack

7fffb0cf7000-7fffb16f8000 rw-p 00000000 00:00 0
Size:              10244 kB
THPeligible:    1
7fffb16f8000-7fffb16f9000 ---p 00000000 00:00 0
Size:                  4 kB
THPeligible:    0
7fffb16f9000-7fffb56fa000 rw-p 00000000 00:00 0
Size:              65540 kB
THPeligible:    1
```
The value returned by `caml_mem_map` is in the middle of the first of these three mappings.  We can see that despite the presence of the guard page, the THP-eligibility of the first mapping (comprising something else's mapping at the lower addresses, followed by our unused 2Mb alignment mapping above that) is unaffected.

After this thread exited, the top mapping changed to:
```
7fffb0cf7000-7fffb14f7000 rw-p 00000000 00:00 0
Size:               8192 kB
```
We can see here that the value returned by `caml_mem_map` now forms the upper endpoint of this mapping, since our 2Mb unused alignment mapping has been unmapped.  The THP-eligibility of the `7fffb0cf7000` mapping remains unaffected.  The guard page mapping and the thread stack mapping itself have been freed.

lpw25 (approved):
Looks good to me

stedolan (commented):
Looks good to me too!

PR #3154, opened 2024-10-16
Rebased version of PR2488 (compaction for batch allocations)

mshinwell:
This is a rebased version of #2488 together with some compilation fixes.

The commits that correspond to those from #2488 contain some rebase fixes.  There is then one entirely new commit at the end.


mshinwell:
There is a further commit to re-enable a compaction test that was previously disabled until multicore markdelay had been merged.

mshinwell:
Superceded by #3179

PR #3160, opened 2024-10-18
Use map for vlower in mode solver (adopted from #1982)

anfelor:
This PR adopts https://github.com/ocaml-flambda/flambda-backend/pull/1982 for the new mode solver. As discussed in chat, we hope that this will reduce the overhead of unique barriers in auto-generated files.

The performance impact of this has not been measured yet.

### DIFF
--- a/ocaml/typing/solver.ml
+++ b/ocaml/typing/solver.ml
@@ -33,17 +33,37 @@ type 'a error =
     right : 'a
   }
 
-(** Map the function to the list, and returns the first [Error] found;
-    Returns [Ok ()] if no error. *)
-let rec find_error (f : 'x -> ('a, 'b) Result.t) : 'x list -> ('a, 'b) Result.t
-    = function
-  | [] -> Ok ()
-  | x :: rest -> (
-    match f x with Ok () -> find_error f rest | Error _ as e -> e)
-
 module Solver_mono (C : Lattices_mono) = struct
+  type any_morph = Any_morph : ('a, 'b, 'd) C.morph -> any_morph
+
+  module VarMap = Map.Make (struct
+    type t = int * any_morph
+
+    let compare (i1, m1) (i2, m2) =
+      match Int.compare i1 i2 with 0 -> compare m1 m2 | i -> i
+  end)
+
+  (** Map the function to the list, and returns the first [Error] found;
+      Returns [Ok ()] if no error. *)
+  let find_error (f : 'x -> ('a, 'b) Result.t) (t : 'x VarMap.t) :
+      ('a, 'b) Result.t =
+    let r = ref (Ok ()) in
+    let _ =
+      VarMap.for_all
+        (fun _ x ->
+          match f x with
+          | Ok () -> true
+          | Error _ as e ->
+            r := e;
+            false)
+        t
+    in
+    !r
+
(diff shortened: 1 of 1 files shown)


goldfirere:
What's the review plan for this patch? I could definitely look at the code and think about correctness, but I think that's not the crux of the issue. @riaqn do you know a good next step here? It looks like #1982 stalled waiting for benchmarks and testing.

riaqn:
I think we should close #1982 in preference of this.

I will mark this PR to be running on our benchmark infra. Then either me or someone in NYC will look at this together with Anton.

anfelor:
@ncik-roberts and I looked at the benchmarks this morning. The uniqueness patch has no big difference to the baseline either with this PR or without. However, this PR does get rid of the pathological behaviour in some files and therefore seems useful to have. There is no meaningful performance regression in other places.

riaqn (dismissed):
Thanks for looking into the benchmark - since it fixes some pathelogical while not causing significant overhead for normal cases, I think it's good to go.

ocaml/typing/solver.ml
```diff
@@ -33,17 +33,37 @@ type 'a error =
...
+            false)
+        t
+    in
+    !r
+
+  let var_map_to_list t = VarMap.fold (fun _ a xs -> a :: xs) t []
+
   type 'a var =
-    { mutable vlower : 'a lmorphvar list;
+    { mutable vlower : 'a lmorphvar VarMap.t;
```
riaqn, line 66: I'm slightly sad that this is a map instead of set. I can see the circular dependency that caused this. I wonder if we should use some `Obj.magic` to workaround the circle and improve performance.

ocaml/typing/solver.ml
```diff
@@ -33,17 +33,37 @@ type 'a error =
...
-(** Map the function to the list, and returns the first [Error] found;
-    Returns [Ok ()] if no error. *)
-let rec find_error (f : 'x -> ('a, 'b) Result.t) : 'x list -> ('a, 'b) Result.t
-    = function
-  | [] -> Ok ()
-  | x :: rest -> (
-    match f x with Ok () -> find_error f rest | Error _ as e -> e)
-
 module Solver_mono (C : Lattices_mono) = struct
+  type any_morph = Any_morph : ('a, 'b, 'd) C.morph -> any_morph
```
riaqn, line 37: I want to annotate this with `[@@unboxed]` for performance but got error because `C.morph` could be a float. @goldfirere How do I say in the interface of `C` that `C.morph` is not a float?

goldfirere: You can't. 😢

I don't know of a workaround here.

ocaml/typing/solver.ml
goldfirere: I would hope there's a way of doing this without `Obj.magic`. Maybe some recursive modules? But it's probably not worth going through extreme measures. I wouldn't expect a set to be much (or at all, really) faster than a map.

riaqn: Yeah I was also wondering about memory savings. But yeah, magic is maybe not worthwhile.

riaqn (approved):
:shipit:

PR #3161, opened 2024-10-19
Add converters to the capsule API

lpw25:
This PR extends the capsule API with a new `Converter.t` type that doesn't cross contention and, when `uncontended`, can freely wrap and unwrap `Data.t` values. They are made available by a `Capsule.access` function, which many of the `Capsule.Data` functions could be implemented in terms of. They should make some patterns slightly easier to write.

Converters replace the `expose` function and the use of global `Password.t`s for converting values, which allows us to safely mark `Password.t`s as crossing contention which makes it easier to nest their usage. Separating these two roles also helps to clarify the relationship between capsules and threads.

### DIFF
--- a/otherlibs/stdlib_alpha/capsule.ml
+++ b/otherlibs/stdlib_alpha/capsule.ml
@@ -12,34 +12,90 @@
 (*                                                                        *)
 (**************************************************************************)
 
+
+module Access : sig
+  (* CR layouts v5: this should have layout [void], but
+     [void] can't be used for function argument and return types yet. *)
+  type 'k t : value mod global portable many unique
+
+  type packed = P : 'k t -> packed
+
+  (* Can break soundness. *)
+  val unsafe_mk : unit -> 'k t @@ portable
+
+  val equality_witness : 'k t -> 'j t -> ('k, 'j) Type.eq @@ portable
+end = struct
+  type dummy
+
+  type 'k t = T : dummy t
+
+  type packed = P : 'k t -> packed
+
+  external unsafe_rebrand : 'k t -> 'j t @@ portable = "%identity"
+
+  let unsafe_mk (type k) () : k t = unsafe_rebrand T
+
+  let equality_witness (type k j) (T : k t) (T : j t) : (k, j) Type.eq =
+    Type.Equal
+
+end
+
+let current () = Access.P (Access.unsafe_mk ())
+
+type initial
+
+let initial = Access.unsafe_mk ()
+
 module Password : sig
   (* CR layouts v5: this should have layout [void], but
      [void] can't be used for function argument and return types yet. *)
-  type 'k t
+  type 'k t : value mod portable many unique uncontended
 
   (* Can break the soundness of the API. *)
   val unsafe_mk : unit -> 'k t @@ portable
+
+  module Shared : sig
+    (* CR layouts v5: this should have layout [void], but
+       [void] can't be used for function argument and return types yet. *)
+    type 'k t
+
+    (* Can break the soundness of the API. *)
+    val unsafe_mk : unit -> 'k t @@ portable
+  end
+
+  val shared : 'k t @ local -> 'k Shared.t @ local @@ portable
 end = struct
   type 'k t = unit
 
   let unsafe_mk () = ()
+
+  module Shared = struct
+    type 'k t = unit
+
+    let unsafe_mk () = ()
+  end
+
+  let shared () = ()
+
 end
 
-module ReaderPassword : sig
-  (* CR layouts v5: this should have layout [void], but
-     [void] can't be used for function argument and return types yet. *)
-  type 'k t
+(* Like [Stdlib.raise], but [portable], and the value
+   it never returns is also [portable] *)
+external reraise : exn -> 'a @ portable @@ portable = "%reraise"
 
-  (* Can break the soundness of the API. *)
-  val unsafe_mk : unit -> 'k t @@ portable
-end = struct
-  type 'k t = unit
+exception Contended of exn @@ contended
 
-  let unsafe_mk () = ()
-end
+let access (type k) (_ : k Password.t) f =
+  let c : k Access.t = Access.unsafe_mk () in
+  match f c with
+  | res -> res
+  | exception exn -> reraise (Contended exn)
 
-let weaken_password : 'k Password.t @ local -> 'k ReaderPassword.t @ local @@ portable =
-  fun _ -> ReaderPassword.unsafe_mk ()
+let access_shared (type k) (_ : k Password.Shared.t) f =
+  let c : k Access.t = Access.unsafe_mk () in
+  match f c with
+  | res -> res
+  | exception exn -> reraise (Contended exn)
 
 (* Like [Stdlib.Mutex], but [portable]. *)
 module M = struct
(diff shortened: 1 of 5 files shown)


ocaml/otherlibs/stdlib_alpha/capsule.mli
```diff
@@ -40,25 +40,68 @@
...
+(** Obtain a converter for the current capsule. Since we do not
+    know the brand for the current capsule, we receive a fresh
+    one. *)
+val current : unit -> Converter.packed @@ portable
+
+(** The brand for the initial capsule. *)
+type initial
+
+(** A converter for the initial capsule *)
+val initial : initial Converter.t
```
dkalinichenko-js, line 70: Perhaps we should have an `initial Password.t` instead?

ocaml/otherlibs/stdlib_alpha/capsule.mli
```diff
@@ -40,25 +40,68 @@
...
     with that brand. *)
 
-(** Passwords represent access to a capsule. *)
+(** Converters can wrap and unwrap [Data.t] values from the current
+    capsule. *)
+module Converter : sig
+
+    (* CR layouts v5: this should have layout [void], but
+       [void] can't be used for function argument and return types yet. *)
+    type 'k t : value mod global portable many unique
```
dkalinichenko-js, line 49: Why is this `portable`?

ocaml/otherlibs/stdlib_alpha/capsule.mli
```diff
@@ -40,25 +40,68 @@
...
+     Obtaining a ['k t] requires acquiring the mutex associated with
+     ['k], and modes prevent retaining the [t] after releasing the
+     mutex. This guarantees that uncontended access to the capsule is
+     only granted to a single domain at once. *)
 
 end
 
+val access :
+  'k Password.t @ local
+  -> ('k Converter.t -> 'a @ portable) @ local portable
```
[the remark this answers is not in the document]
dkalinichenko-js: I guess this shouldn't be able to leak a converter without marking it contended.

ocaml/otherlibs/stdlib_alpha/capsule.mli
```diff
@@ -40,25 +40,68 @@
...
-        the ability of the current domain to have [uncontended]
-        access to the capsule ['k]. They don't have a runtime representation,
-        and can't be passed between domains. Just as we don't share passwords
-        between people, we don't share [Password.t]s between domains.
-
-        Obtaining a ['k t] requires acquiring the mutex associated with ['k],
-        and modes prevent retaining the [t] after releasing the mutex. This
-        guarantees that uncontended access to the capsule is only granted
-        to a single domain at once. *)
+    type 'k t : value mod portable many unique uncontended
```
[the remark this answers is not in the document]
dkalinichenko-js: Same for contention.

ocaml/otherlibs/stdlib_alpha/capsule.mli
```diff
@@ -108,9 +152,21 @@ module Data : sig
         Operations on [('a, 'k) t] require a ['k Password.t],
         created from the ['k Mutex.t]. *)
 
-    exception Contended of exn @@ contended
-    (** If a function accessing the contents of the capsule raises an exception,
-        it is wrapped in [Contended] to avoid leaking access to the data. *)
+    val wrap :
```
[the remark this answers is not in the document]
dkalinichenko-js: Oh, I see, any way to send a converted across domains should mark it `contended`, which will prevent its usage.

ocaml/otherlibs/stdlib_alpha/capsule.mli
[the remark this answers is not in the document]
dkalinichenko-js: Locality is not sufficient to guarantee non-shared access in presence of `fork_join`.

ocaml/otherlibs/stdlib_alpha/capsule.mli
```diff
@@ -40,25 +40,68 @@
...
-(** Passwords represent access to a capsule. *)
+(** Converters can wrap and unwrap [Data.t] values from the current
+    capsule. *)
+module Converter : sig
+
+    (* CR layouts v5: this should have layout [void], but
+       [void] can't be used for function argument and return types yet. *)
+    type 'k t : value mod global portable many unique
+    (** ['k t] is the type of "converters" that can wrap and unwrap
+        [Data.t] values. An [uncontended] ['k t] indicates that ['k] is the
```
dkalinichenko-js, line 51: Does that mean `assert_current : 'a Converter.t -> 'b Converter.t -> ('a, 'b) Type.eq` is sound?

ocaml/otherlibs/stdlib_alpha/capsule.mli
```diff
@@ -40,25 +40,68 @@
...
+     [Password.t]s between fibers.
+
+     Obtaining a ['k t] requires acquiring the mutex associated with
+     ['k], and modes prevent retaining the [t] after releasing the
+     mutex. This guarantees that uncontended access to the capsule is
+     only granted to a single domain at once. *)
 
 end
 
+val access :
```
dkalinichenko-js, line 97: `access` should be `@@ portable`.

dkalinichenko-js:
Is it correct to think of `('k Converter.t @ local -> ...) @ portable` as a modality indexed by the capsule `'k` that has access to all variables from the background context guarded by `(..., 'k) Data.t` (in addition to all `contended` variables)?

ocaml/otherlibs/stdlib_alpha/capsule.mli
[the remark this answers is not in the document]
lpw25: Right, we aren't using locality to constrain converters. Instead they are treated like mutable data that is part of the capsule. So if they cross the capsule boundary they become contended and can no longer be used.

ocaml/otherlibs/stdlib_alpha/capsule.mli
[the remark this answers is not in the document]
lpw25: Yeah, with `fork_join` we will be relying on the `unyielding/yielding` modes to prevent passwords from being used by two threads at the same time. You can think of a password as an effect handler, and `yielding` is how we prevent those from being used inside a `fork_join` too.

ocaml/otherlibs/stdlib_alpha/capsule.mli
lpw25: As discussed offline, only because it safely can be not because it is actually useful. I prefer to reserve portability for controlling things that are like functions, and using contention for controlling things that are like mutable state, and I picture the converters as acting more like mutable state.

ocaml/otherlibs/stdlib_alpha/capsule.mli
lpw25: Yeah, I guess it does. Maybe we should include it?

lpw25:
I think the `local` isn't needed, but yes I think you can think of that as some form of modality.

ocaml/otherlibs/stdlib_alpha/capsule.mli
lpw25: I went ahead and added it

dkalinichenko-js (approved):

lpw25:
I've renamed `Converter.t` to `Access.t` which is a bit better. I've also updated the API to account for reader/writer locks. Notes that, rather than add a new `Access.Shared.t` to represent shared access, we just use having an `Access.t` at the `shared` mode to represent shared access.

dkalinichenko-js:
Happy with the updated PR. To be sure, we can't have `Password.t @ shared` because passwords should mode-cross contention, right?

PR #3162, opened 2024-10-20
Delete Jane Syntax

ncik-roberts:
This PR does a lot but each commit is reasonably self-contained.
  * Move remaining syntax forms to the parsetree (iarrays, comprehensions, module strengthening, module instances)
  * Move other misc things out from jane syntax modules:
    * The attribute for encoding curried functions moves out to `Builtin_attributes`
    * The function for checking enabledness of an extension moves out to `Language_extensions`
  * Delete `Jane_syntax` and `Jane_syntax_parsing`, and remove lingering references from the codebase

I think it's easier to get rid of it all in one go rather than split out the work into separate PRs.

The only important tests I add are ones that ensure we're checking for extension enabledness for the appropriate constructs. This is something that Jane Syntax gave us for free that we now need to manually check in the post-Jane Syntax world. (I think this is a fine tradeoff!)

Most of this is pretty mechanical. The main non-mechanical thing are the decisions in parsetree.mli, which are easy to modify if you disagree with my reasoning below:
  * Comprehensions: I add `Pcomp_`/`pcomp_` prefixes to comprehension records/variants &mdash; this is typical style in the parsetree to avoid naming collisions in the big recursive knot of types.
  * Comprehensions: I make `Pexp_comprehension` the top-level node for comprehensions instead of separate `Pexp_array_comprehension` and `Pexp_list_comprehension` &mdash; I don't feel strongly about this, but this more closely matched the choices we made when encoding Jane Syntax and was a more straightforward diff on top of existing code.
  * Iarrays: I add an argument to `Pexp_array`/`Ppat_array` instead of introducing new constructors for iarrays. This is what the original author said we should do in a comment; also, this more closely matches the current approach in the upstream PR (which just uses disambiguation instead of introducing dedicated syntax) and so maybe has less drift.
  * Module strengthening: I choose to add the constructor to `module_type_desc` instead of `with_constraint` even though the latter might be the better fit long-term. This choice was a smaller diff on top of the choices we made in the jane syntax encoding; we can easily revisit later, but this isn't the PR to make major changes in the structure of types.
  * Module instances: It's a bit odd that there's no dedicated syntax for this construct, but having a separate parsetree node does seem like the right move so that `ast_mapper`/`ast_iterator` don't mistake a bound name for a functor argument.


ocaml/parsing/depend.ml
```diff
@@ -605,6 +569,8 @@ and add_module_expr bv modl =
       add_expr bv e
   | Pmod_extension e ->
       handle_extension e
+  (* CR nroberts for lmaurer: what should happen here?  *)
+  | Pmod_instance _ -> ()
```
ncik-roberts, line 573: @lukemaurer How should we be treating module instances in ocamldep?

ocaml/parsing/parser.mly
```diff
@@ -822,6 +813,47 @@ let package_type_of_module_type pmty =
...
+    | { txt = _; loc } ->  raise_malformed_instance loc
+  in
+  let gather_args mexpr =
+    let rec loop mexpr acc =
+      match mexpr with
+      | { pmod_desc = Pmod_apply (f, v); _ } ->
+         (match f.pmod_desc with
+          | Pmod_apply (f, n) -> loop f ((n, v) :: acc)
+          | _ -> raise_malformed_instance f.pmod_loc)
+      | head -> head, acc
```
ncik-roberts, line 835: FYI @lukemaurer : this loop used to reverse the list at the end, but I think this was unintentional: the list is being built up in the proper order already bc functor application left-associates. I've changed this in this PR. Probably it doesn't matter anyway, as the order of instance args doesn't appear to mean anything.

lukemaurer: !! Yes, I think you're right on all counts (including that it doesn't matter much, though for errors we probably do want to stick to the user's syntax?).

ocaml/parsing/depend.ml
ncik-roberts: Luke tells me we should add a dependency on all module names that appear in the instance.

ccasin (approved):
This looks great.

One comment below.  Also: I think the code around `builtin_attr_prefixes` in `builtin_attributes.ml` can probably be deleted now?

typing/typecore.ml
```diff
@@ -2872,9 +2853,13 @@ and type_pat_aux
...
-  | Ppat_array spl ->
-      type_pat_array (Mutable Alloc.Comonadic.Const.legacy)
-         spl sp.ppat_attributes
+  | Ppat_array (mut, spl) ->
+      let mut =
+        match mut with
+        | Mutable -> Mutable Alloc.Comonadic.Const.legacy
+        | Immutable -> Immutable
+      in
+      type_pat_array mut spl sp.ppat_attributes
```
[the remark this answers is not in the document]
ncik-roberts: Thanks for catching that! Fixed.

ncik-roberts:
Indeed, that can be made to look more like upstream's. I've done this. (I've kept around `jane.non_erasable.instances` as a builtin attribute as that's needed for module instances.)

PR #3172, opened 2024-10-22
Module parameter names should have its own type

riaqn:
Based on #1726 

Uses the fact that parameters can't be parameterized, and encode that info in the types. This can help remove some `assert false` in the code.

This will be picked up by @lukemaurer when he has time.

### DIFF
--- a/ocaml/typing/persistent_env.ml
+++ b/ocaml/typing/persistent_env.ml
@@ -34,27 +34,27 @@ type error =
   | Direct_reference_from_wrong_package of
       CU.t * filepath * CU.Prefix.t
   | Illegal_import_of_parameter of Global_module.Name.t * filepath
-  | Not_compiled_as_parameter of Global_module.Name.t * filepath
+  | Not_compiled_as_parameter of Global_module.Parameter.t * filepath
   | Imported_module_has_unset_parameter of
       { imported : Global_module.Name.t;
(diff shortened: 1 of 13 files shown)


ocaml/typing/persistent_env.ml
```diff
@@ -97,14 +98,19 @@ type global_name_info = {
   gn_mentioned_by : Global_module.Name.t; (* For error reporting *)
 }
 
+type import_kind =
+  | Normal of {
+    imp_params : Global_module.Parameter.t list;
+    imp_arg_for : Global_module.Parameter.t option;
+    imp_impl : CU.t;
+    imp_raw_sign : Signature_with_global_bindings.t;
```
riaqn, line 106: Discussed offline - we think it's simpler to lift this out to be under `import` directly.

ocaml/typing/persistent_env.ml
```diff
@@ -116,17 +122,23 @@ type import_info =
   | Missing
   | Found of import
 
+type pers_name_kind =
+  | Parameter of Global_module.Parameter.t
+  | Normal of {
+    pn_global : Global_module.t;
+    pn_arg_for : Global_module.Parameter.t option;
```
riaqn, line 129: Discussed offline - this can be removed.

riaqn:
superseded by #3708

PR #3180, opened 2024-10-23
Documentation for Uniqueness

anfelor:
This PR adds some documentation for the uniqueness mode. This unblocks #3065.

I have followed the structure of the `local/` documentation. This is arguably too heavy weight for the current amount of content, but I would expect that we will add descriptions of borrowing and in-place overwrite to these files once those features are ready.

I have used the new syntax for modes throughout. I believe that this makes the documentation more clear and I would expect that most users only start to use uniqueness later on when the new syntax is stable. I have added a short description of the current syntax for users who read this earlier.

~~I have marked this PR as a draft because it should be merged simultanously with #3065.~~ Edit: @riaqn remarks below that this can be merged before #3065.

### DIFF
--- a/jane/doc/extensions/uniqueness/intro.md
+++ b/jane/doc/extensions/uniqueness/intro.md
@@ -103,9 +103,10 @@ let bad t =
     free t
 ```
 
-In the `bad` function above, the `free t` assumes that it gets all of `t`
-uniquely, but this is not true if `field` has already been freed. However, we
-can use parts of `t` twice if these uses happen in different branches:
+In the `bad` function above, `free_field` assumes that it gets the only
+reference to `field`. But `field` can still be referenced from `t`, which is
+itself passed to `free`. However, we are allowed to use parts of `t` twice if
+these uses happen in different branches:
 
 ```ocaml
 let okay t =
--- a/jane/doc/extensions/uniqueness/intro.md
+++ b/jane/doc/extensions/uniqueness/intro.md
@@ -1,0 +1,111 @@
+# Introduction to Uniqueness
+
+See also the full feature [reference](reference.md) and [common pitfalls](pitfalls.md).
+
+The `unique` mode designates values that have only a single reference pointing
+to them. For example, an externally allocated data structure might support a
+`free` operation:
+
+```ocaml
+val free : t @ unique -> unit
+```
+
+The compiler guarantees that this operation will be sound: as the argument is
+`unique`, there can be no further references to it that could lead to a
+use-after-free. The compiler tracks the uniqueness of values and marks values
+which have more than one reference pointing to them with the `aliased` mode:
+
+```ocaml
+let duplicate : t -> t * t @ aliased = fun t -> t, t
+```
+
+When a `unique` value is captured in a closure, this closure can be invoked at
+most once: if the closure was invoked more often, it could not use the value
+uniquely each time. Values that contain closures that can be invoked at most
+once are at mode `once`, while values with closures that can be invoked
+arbitrarily often are at mode `many`:
+
+```ocaml
+let delay_free : t @ unique -> (unit -> unit) @ once = fun t -> fun () -> free t
+```
+
+## Modalities and Mode Crossing
+
+These modes form two mode axis: the _uniqueness_ of a value is either `unique`
+or `aliased`, while the _affinity_ of a value is `once` or `many`. Similar to
+[locality](../local/intro.md), uniqueness and affinity are deep properties. If a
+value is at mode `unique` then all of its children are also `unique`.
+
+We make an exception to this rule for record fields that are annotated with a
+_modality_. Analogous to global fields, a record field that is annotated as `@@
+aliased`, can contain `aliased` values even if the record itself is `unique`.
+However, a read from such a field always returns an `aliased` value. If a record
+field is annotated by `@@ many`, then it can only store `many` values even if
+the record itself is `once`. In return, a read of that field always yields a
+`many` value.
+
+Analogous to the `globalize` function, it is possible to cast between the modes:
+
+```ocaml
+val alias : 'a @ unique -> 'a @ aliased
+val linearize : 'a @ many -> 'a @ once
+```
+
+However, the other directions are disallowed in general: if a value is `aliased`
+there is no way to get it back to mode `unique`; similarly a `once` value can
+not be made `many`.
+
+We make an exception to this rule for values of special types to which these
+modes do not apply. For example, `int`s and other immediates can be used as
+`unique` and `many`, no matter their mode. Similarly, most values that don't
+contain functions can be used as `many`. For example, an `int list @ once
+aliased` can be used as `many` but not as `unique`.
+
+## Checking for Uniqueness
+
+The compiler performs a sophisticated analysis to determine which values are
+aliased and which ones are unique. For example, it is fine to match on a value
+and then use it:
+
+```ocaml
+let okay t =
+  match t with
+  | Con { field } -> free t
+```
+
+But it would not be fine to use (parts of) `t` twice:
+
+```ocaml
+let bad t =
+  match t with
+  | Con { field } ->
+    free_field field;
+    free t
+```
+
+In the `bad` function above, the `free t` assumes that it gets all of `t`
+uniquely, but this is not true if the `field` as been freed previously. However,
+we can use parts of `t` twice if these uses happen in different branches:
+
+```ocaml
+let okay t =
+  match t with
+  | Con { field } ->
+    if flip_coin ()
+    then free_field field
+    else free t
+```
+
+or if the uses both accept `aliased` arguments:
+
+```ocaml
+val store : t @ aliased -> unit
+
+let okay t =
+  match t with
+  | Con { field } ->
+    store_field field;
+    store t
+```
+
+For more details, read [the reference](./reference.md).
--- a/jane/doc/extensions/uniqueness/intro.md
+++ b/jane/doc/extensions/uniqueness/intro.md
@@ -1,8 +1,7 @@
 # Introduction to Uniqueness
 
 See also the full feature [reference](reference.md) and [common pitfalls](pitfalls.md).
-In this document, we use the new [syntax for modes](../modes/syntax.md),
-which we expect to be stable soonish.
+In this document, we use the new [syntax for modes](../modes/syntax.md).
 
 The `unique` mode designates values that have only a single reference pointing
 to them. If an operation takes a `unique` argument, it will consume the only
--- a/jane/doc/extensions/uniqueness/intro.md
+++ b/jane/doc/extensions/uniqueness/intro.md
@@ -117,7 +117,8 @@ let okay t =
     else free t
 ```
 
-or if the uses both accept `aliased` arguments:
+or if the uses both accept `aliased` arguments (eg. by calling a function that
+only stores its argument somewhere):
 
 ```ocaml
 val store : t @ aliased -> unit
--- a/jane/doc/extensions/uniqueness/intro.md
+++ b/jane/doc/extensions/uniqueness/intro.md
@@ -130,14 +130,4 @@ let okay t =
     store t
 ```
 
-## Current syntax
-
-Until the new [syntax for modes](../modes/syntax.md) is stable, our syntax for
-uniqueness will be slightly different from the description above. Currently, you
-would mark an argument or return value in a type signature as unique by
-prepending it with `unique_` and as once by prepending it with `once_`. You can
-not mark arguments or return values as aliased or many. To mark a record field
-as `@@ aliased`, you would prepend `aliased_` and to mark it as `@@ many`, you
-would prepend `many_`.
-
 For more details, read [the reference](./reference.md).
--- a/jane/doc/extensions/uniqueness/intro.md
+++ b/jane/doc/extensions/uniqueness/intro.md
@@ -55,6 +55,8 @@ to make them unique, we wrap them in a record with the `aliased` modality:
 
 ```ocaml
 type 'a aliased = { a : 'a @@ aliased } [[@@unboxed]]
+
+val cons : 'a @ aliased -> 'a aliased list @ unique -> 'a aliased list @ unique
 ```
 
 ## Mode Crossing
--- a/jane/doc/extensions/uniqueness/intro.md
+++ b/jane/doc/extensions/uniqueness/intro.md
@@ -12,10 +12,11 @@ to them. For example, an externally allocated data structure might support a
 val free : t @ unique -> unit
 ```
 
-The compiler guarantees that this operation will be sound: as the argument is
-`unique`, there can be no further references to it that could lead to a
-use-after-free. The compiler tracks the uniqueness of values and marks values
-which have more than one reference pointing to them with the `aliased` mode:
+Normally, a `free` operation would be dangerous since any remaining references
+can lead to a use-after-free segfault. But with the `unique` mode, the compiler
+guarantees that no further references remain. To provide this guarantee, the
+compiler tracks the uniqueness of values and marks values which have more than
+one reference pointing to them with the `aliased` mode:
 
 ```ocaml
 let duplicate : t -> t * t @ aliased = fun t -> t, t
@@ -23,20 +24,20 @@ let duplicate : t -> t * t @ aliased = fun t -> t, t
 
 When a `unique` value is captured in a closure, this closure can be invoked at
 most once: if the closure was invoked more often, it could not use the value
-uniquely each time. Values that contain closures that can be invoked at most
-once are at mode `once`, while values with closures that can be invoked
-arbitrarily often are at mode `many`:
+uniquely each time. Such closures are at mode `once`, while closures that can be
+invoked arbitrarily often are at mode `many`:
 
 ```ocaml
 let delay_free : t @ unique -> (unit -> unit) @ once = fun t -> fun () -> free t
 ```
 
-## Modalities and Mode Crossing
+## Modalities
 
 These modes form two mode axis: the _uniqueness_ of a value is either `unique`
 or `aliased`, while the _affinity_ of a value is `once` or `many`. Similar to
 [locality](../local/intro.md), uniqueness and affinity are deep properties. If a
-value is at mode `unique` then all of its children are also `unique`.
+value is at mode `unique` then all of its children are also `unique`. If a value
+is `once` then all of the closures it contains are also at mode `once`.
 
 We make an exception to this rule for record fields that are annotated with a
 _modality_. Analogous to global fields, a record field that is annotated as `@@
@@ -46,22 +47,19 @@ field is annotated by `@@ many`, then it can only store `many` values even if
 the record itself is `once`. In return, a read of that field always yields a
 `many` value.
 
-Analogous to the `globalize` function, it is possible to cast between the modes:
+## Mode Crossing
 
-```ocaml
-val alias : 'a @ unique -> 'a @ aliased
-val linearize : 'a @ many -> 'a @ once
-```
+You can always cast a value to a mode that affords fewer guarantees: You can use
+a `unique` value like an `aliased` value, and a `many` value like a `once`
+value. However, the other directions are disallowed in general: if a value is
+`aliased` there is no way to get it back to mode `unique`; similarly a `once`
+value can not be made `many`.
 
-However, the other directions are disallowed in general: if a value is `aliased`
-there is no way to get it back to mode `unique`; similarly a `once` value can
-not be made `many`.
-
-We make an exception to this rule for values of special types to which these
-modes do not apply. For example, `int`s and other immediates can be used as
-`unique` and `many`, no matter their mode. Similarly, most values that don't
-contain functions can be used as `many`. For example, an `int list @ once
-aliased` can be used as `many` but not as `unique`.
+We make an exception to this rule for values of those types that cross these
+modes. For example, `int`s and other immediates can be used as `unique` and
+`many`, no matter their mode. Similarly, most values that don't contain
+functions can be used as `many`. For example, an `int list @ once aliased` can
+be used as `many` but not as `unique`.
 
 ## Checking for Uniqueness
 
@@ -86,7 +84,7 @@ let bad t =
 ```
 
 In the `bad` function above, the `free t` assumes that it gets all of `t`
-uniquely, but this is not true if the `field` as been freed previously. However,
+uniquely, but this is not true if the `field` has been freed previously. However,
 we can use parts of `t` twice if these uses happen in different branches:
 
 ```ocaml
--- a/jane/doc/extensions/uniqueness/intro.md
+++ b/jane/doc/extensions/uniqueness/intro.md
@@ -1,6 +1,8 @@
 # Introduction to Uniqueness
 
 See also the full feature [reference](reference.md) and [common pitfalls](pitfalls.md).
+In this document, we use the new [syntax for modes](../modes/syntax.md),
+which we expect to be stable soonish.
 
 The `unique` mode designates values that have only a single reference pointing
 to them. For example, an externally allocated data structure might support a
@@ -107,5 +109,15 @@ let okay t =
     store_field field;
     store t
 ```
+
+## Current syntax
+
+Until the new [syntax for modes](../modes/syntax.md) is stable, our syntax for
+uniqueness will be slightly different from the description above. Currently, you
+would mark an argument or return value in a type signature as unique by
+prepending it with `unique_` and as once by prepending it with `once_`. You can
+not mark arguments or return values as aliased or many. To mark a record field
+as `@@ aliased`, you would prepend `aliased_` and to mark it as `@@ many`, you
+would prepend `many_`.
 
 For more details, read [the reference](./reference.md).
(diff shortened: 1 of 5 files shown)


riaqn:
I think it's fine for this PR to be merged before #3065. :-)

jane/doc/extensions/uniqueness/intro.md
```diff
@@ -0,0 +1,123 @@
+# Introduction to Uniqueness
+
+See also the full feature [reference](reference.md) and [common pitfalls](pitfalls.md).
+In this document, we use the new [syntax for modes](../modes/syntax.md),
+which we expect to be stable soonish.
+
+The `unique` mode designates values that have only a single reference pointing
+to them. For example, an externally allocated data structure might support a
+`free` operation:
```
riaqn, line 9: Hmm how about something like:
> The `unique` mode designates a reference to a value that is the only reference to that value throughout the program.

Like, the `unique` is on the reference itself, not the value it points to.

jane/doc/extensions/uniqueness/intro.md
```diff
@@ -0,0 +1,123 @@
...
+
+```ocaml
+val free : t @ unique -> unit
+```
+
+The compiler guarantees that this operation will be sound: as the argument is
```
riaqn, line 15: Hmm is the operation unsound prior to uniqueness? I guess "sound" is a strong and vague word. Maybe we can just say "With the unique annotation, the compiler can now prevents use-after-free"?

jane/doc/extensions/uniqueness/intro.md
```diff
@@ -0,0 +1,123 @@
...
+`unique`, there can be no further references to it that could lead to a
+use-after-free. The compiler tracks the uniqueness of values and marks values
+which have more than one reference pointing to them with the `aliased` mode:
+
+```ocaml
+let duplicate : t -> t * t @ aliased = fun t -> t, t
+```
+
+When a `unique` value is captured in a closure, this closure can be invoked at
```
riaqn, line 24: Maybe clearer to say "if a closure captures a value and uses it as `unique`". Because capturing a `unique` value and using it as `aliased` is fine. I think an average user might be confused by the two.

jane/doc/extensions/uniqueness/intro.md
```diff
@@ -0,0 +1,123 @@
...
+most once: if the closure was invoked more often, it could not use the value
+uniquely each time. Values that contain closures that can be invoked at most
+once are at mode `once`, while values with closures that can be invoked
+arbitrarily often are at mode `many`:
```
riaqn, line 28: Instead of saying "values that contains closures that..", maybe just say "closures that ..."? I think users can infer the container-containee relation naturally (from their experience with local allocation)

jane/doc/extensions/uniqueness/intro.md
```diff
@@ -0,0 +1,123 @@
...
+
+We make an exception to this rule for record fields that are annotated with a
+_modality_. Analogous to global fields, a record field that is annotated as `@@
+aliased`, can contain `aliased` values even if the record itself is `unique`.
+However, a read from such a field always returns an `aliased` value. If a record
+field is annotated by `@@ many`, then it can only store `many` values even if
+the record itself is `once`. In return, a read of that field always yields a
+`many` value.
+
+Analogous to the `globalize` function, it is possible to cast between the modes:
```
riaqn, line 49: The `globalize` converts local value to global by copying to heap, which maybe is not what you mean here? Maybe just say:
> You can always use a value at a mode that's weaker than it actually at. For example, you can pass a `unique` value to a function that's expecting an `aliased` parameter, or pass a `many` value to a function that's expecting an `once` value.

jane/doc/extensions/uniqueness/intro.md
```diff
@@ -0,0 +1,123 @@
...
+```ocaml
+val alias : 'a @ unique -> 'a @ aliased
+val linearize : 'a @ many -> 'a @ once
+```
+
+However, the other directions are disallowed in general: if a value is `aliased`
+there is no way to get it back to mode `unique`; similarly a `once` value can
+not be made `many`.
+
+We make an exception to this rule for values of special types to which these
```
riaqn, line 60: I think we have popularized the term "mode-crossing" to the users, so it's probably more clarifying if we say the terms "crosses linearity axis", "crosses to many mode", etc

jane/doc/extensions/uniqueness/intro.md
```diff
@@ -0,0 +1,123 @@
...
+let okay t =
+  match t with
+  | Con { field } ->
+    store_field field;
+    store t
+```
+
+## Current syntax
+
+Until the new [syntax for modes](../modes/syntax.md) is stable, our syntax for
```
riaqn, line 115: Hmm I thought the new syntax is quite usable now. I will confirm.

jane/doc/extensions/uniqueness/intro.md
```diff
@@ -0,0 +1,123 @@
...
+```ocaml
+let bad t =
+  match t with
+  | Con { field } ->
+    free_field field;
+    free t
+```
+
+In the `bad` function above, the `free t` assumes that it gets all of `t`
+uniquely, but this is not true if the `field` as been freed previously. However,
```
riaqn, line 89: *has

jane/doc/extensions/uniqueness/intro.md
anfelor: I do think that uniqueness is on the value and find the current phrasing clearer than the proposed one. The subtle special case where you can use a value uniquely in one branch and share it in the other still makes sense thinking about references instead of usages and is only explained in the `reference.md` document.

anfelor: In particular, I would want readers to form a denotational semantics in their head here. "This value has only one reference" is such a semantics, while "This values usage in the program is the only usage" is more operational.

jane/doc/extensions/uniqueness/intro.md
anfelor: Right, "sound" is too PLy. I'll remove it

jane/doc/extensions/uniqueness/intro.md
```diff
@@ -0,0 +1,121 @@
+# Introduction to Uniqueness
+
+See also the full feature [reference](reference.md) and [common pitfalls](pitfalls.md).
+In this document, we use the new [syntax for modes](../modes/syntax.md),
+which we expect to be stable soonish.
+
+The `unique` mode designates values that have only a single reference pointing
+to them. For example, an externally allocated data structure might support a
```
TheNumbat, line 8: This isn't directly an example of a value with a single reference - something like 'uniqueness is useful in situations like freeing an externally allocated data structure...'

jane/doc/extensions/uniqueness/intro.md
```diff
@@ -0,0 +1,121 @@
...
+compiler tracks the uniqueness of values and marks values which have more than
+one reference pointing to them with the `aliased` mode:
+
+```ocaml
+let duplicate : t -> t * t @ aliased = fun t -> t, t
+```
+
+When a `unique` value is captured in a closure, this closure can be invoked at
+most once: if the closure was invoked more often, it could not use the value
+uniquely each time. Such closures are at mode `once`, while closures that can be
```
TheNumbat, line 27: I don't think it's obvious why this is true, since invoking the closure multiple times doesn't necessarily create additional references. Maybe also mention that using something uniquely is interpreted as 'consuming' it. (Though the example does make this clear)

jane/doc/extensions/uniqueness/intro.md
```diff
@@ -0,0 +1,121 @@
...
+invoked arbitrarily often are at mode `many`:
+
+```ocaml
+let delay_free : t @ unique -> (unit -> unit) @ once = fun t -> fun () -> free t
+```
+
+## Modalities
+
+These modes form two mode axis: the _uniqueness_ of a value is either `unique`
```
TheNumbat, line 36: axes

jane/doc/extensions/uniqueness/intro.md
```diff
@@ -0,0 +1,121 @@
...
+value is at mode `unique` then all of its children are also `unique`. If a value
+is `once` then all of the closures it contains are also at mode `once`.
+
+We make an exception to this rule for record fields that are annotated with a
+_modality_. Analogous to global fields, a record field that is annotated as `@@
+aliased`, can contain `aliased` values even if the record itself is `unique`.
+However, a read from such a field always returns an `aliased` value. If a record
+field is annotated by `@@ many`, then it can only store `many` values even if
+the record itself is `once`. In return, a read of that field always yields a
+`many` value.
```
TheNumbat, line 48: Could use a code example

jane/doc/extensions/uniqueness/intro.md
```diff
@@ -0,0 +1,121 @@
...
+a `unique` value like an `aliased` value, and a `many` value like a `once`
+value. However, the other directions are disallowed in general: if a value is
+`aliased` there is no way to get it back to mode `unique`; similarly a `once`
+value can not be made `many`.
+
+We make an exception to this rule for values of those types that cross these
+modes. For example, `int`s and other immediates can be used as `unique` and
+`many`, no matter their mode. Similarly, most values that don't contain
+functions can be used as `many`. For example, an `int list @ once aliased` can
+be used as `many` but not as `unique`.
```
TheNumbat, line 62: Could also use a code example, and a link to documentation about `with` kinds if/when that exists

jane/doc/extensions/uniqueness/intro.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Checking for Uniqueness
+
+The compiler performs a sophisticated analysis to determine which values are
+aliased and which ones are unique. For example, it is fine to match on a value
```
TheNumbat, line 67: fine to match on a unique value and then use its contents uniquely

jane/doc/extensions/uniqueness/intro.md
```diff
@@ -0,0 +1,121 @@
...
+```ocaml
+let bad t =
+  match t with
+  | Con { field } ->
+    free_field field;
+    free t
+```
+
+In the `bad` function above, the `free t` assumes that it gets all of `t`
+uniquely, but this is not true if the `field` has been freed previously. However,
```
TheNumbat, line 87: not true if `field` has already been freed.

jane/doc/extensions/uniqueness/pitfalls.md
```diff
@@ -0,0 +1,115 @@
...
+uniqueness, as well as some suggested workarounds. Over time, this list may grow
+(as experience discovers new things that go wrong) or shrink (as we deploy new
+compiler versions that ameliorate some issues).
+
+If you want an introduction to uniqueness, see the [introduction](intro.md).
+
+## Annotating unique return values
+
+When the compiler infers uniqueness, it will only mark values as `unique` if
+they are actually used as `unique`. For example, if create a smart constructor:
```
TheNumbat, line 13: For example, given a smart constructor:

jane/doc/extensions/uniqueness/pitfalls.md
```diff
@@ -0,0 +1,115 @@
...
+
+```ocaml
+let mk () = { x = 1 }
+```
+
+This value is not unique by default. This helps the compiler identify static
```
TheNumbat, line 19: The result is not...

jane/doc/extensions/uniqueness/pitfalls.md
```diff
@@ -0,0 +1,115 @@
...
+allocations: as long as `mk ()` returns an `aliased` value, the compiler can
+lift out the allocation so that `mk ()` returns the same pointer on each
+invokation. However, you might want to annotate this allocation as `unique`:
```
TheNumbat, line 22: invocation

jane/doc/extensions/uniqueness/pitfalls.md
```diff
@@ -0,0 +1,115 @@
...
+
+Most times, it will not be necessary to manually annotate a return value as
+`unique`. While the compiler defaults the return mode to `aliased`, it will make
+the return mode `unique` if there is any unique use of the result of `mk ()`:
+
+```ocaml
+let use () = free (mk ())
+```
+
+This code checks even if `mk` is not annotated, because, now that the need as
```
TheNumbat, line 45: typechecks even if `mk` is not annotated because, now that a need has...

jane/doc/extensions/uniqueness/pitfalls.md
```diff
@@ -0,0 +1,115 @@
...
+                  ^^^^^^^^^^
+Error: This value is "aliased" but expected to be "unique".
+```
+
+When abstracting functions that return unique values into a module, we need to
+provide an explicit mode annotation in the signature.
+
+## Threading unique values
+
+While unique values allow safe mutation, they require a coding style that is
```
TheNumbat, line 65: These docs don't include unique mutation / strong updates - is that supported yet? If so it should be covered in the intro

jane/doc/extensions/uniqueness/pitfalls.md
```diff
@@ -0,0 +1,115 @@
...
+much closer to purely functional code than mutating code. For example, you might
+imagine a function to set values in unique arrays:
+
+```ocaml
+val set : 'a array @ unique -> int -> 'a -> unit
```
TheNumbat, line 70: Should `'a` be unique too?

jane/doc/extensions/uniqueness/reference.md
```diff
@@ -0,0 +1,103 @@
...
+
+The uniqueness analysis makes it possible to consume some fields of an
+allocation without losing access to the allocation or its other fields. For
+example, you might write:
+
+```ocaml
+let okay r =
+  free r.field1;
+  match r with
+  | { field2 } -> free field2
```
TheNumbat, line 41: Is it valid to completely omit field1 here? Or should this be `{ field2; _ }`

jane/doc/extensions/uniqueness/pitfalls.md
anfelor: Overwrite is not yet supported -- not even as an alpha language extension.

jane/doc/extensions/uniqueness/intro.md
anfelor: No, it is `t` itself which is used here not its contents. The point I want to make here is that matching is not a use.

TheNumbat: ah right ok - I think this should still say 'match on a unique value and then use it' though

jane/doc/extensions/uniqueness/reference.md
anfelor: It is valid to do this in the testsuite of the compiler. Would the internal codeformatter complain about this?

TheNumbat: Yeah, it emits the `missing-record-field-pattern` warning

jane/doc/extensions/uniqueness/intro.md
anfelor: I think it is still in the `extension-universe alpha` though. And the pretty-printer for type-checked code still emits `unique_` and `once_` instead of `@ unique` and `@ once`, so users might well see that syntax.

jane/doc/extensions/uniqueness/intro.md
anfelor: I changed `captured` to `consumed`

jane/doc/extensions/uniqueness/intro.md
anfelor: Thanks, it now reads:

> The `unique` mode designates values that have only a single reference pointing
to them. If an operation takes a `unique` argument, it will consume the only
reference to the value. For example, an externally allocated data structure
might support a `free` operation:

jane/doc/extensions/uniqueness/intro.md
anfelor: Thanks, I changed `captured` to `consumed`

jane/doc/extensions/uniqueness/intro.md
anfelor: I added the `aliased` type and an explanation when it might come in handy.

jane/doc/extensions/uniqueness/intro.md
anfelor: Thanks, I missed that `with` kinds were not yet stable. I would assume that will edit the documentation again once they become available?

TheNumbat (approved):

jane/doc/extensions/uniqueness/reference.md
```diff
@@ -0,0 +1,103 @@
...
+
+The uniqueness analysis makes it possible to consume some fields of an
+allocation without losing access to the allocation or its other fields. For
+example, you might write:
+
+```ocaml
+let okay r =
+  free r.field1;
+  match r with
+  | { field2; _ } -> free field2
```
riaqn, line 41: We can strenghten the point by binding `field1` as well in the pattern. Of course, you can't use it. The point is that you can give a name to `field1` even if it's not usable.

jane/doc/extensions/uniqueness/reference.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+This is fine: we free the first field of `r`, then pattern match on `r` itself
+(which was not affected by the free of its child) and then free the second
+child. However, it would not be okay to free the first field and then free `r`
+itself (which has now partially been consumed):
```
riaqn, line 47: I wonder if instead of `free` which has certain connotation, we should just say `use_uniquely` which is more generic. For the current example, the reader might think: well I've freed `r.field1`, but I can still free the memory block of `r` itself (which doesn't contain `r.field1`).

jane/doc/extensions/uniqueness/reference.md
```diff
@@ -0,0 +1,103 @@
...
+
+```ocaml
+match x, y, z with
+| p, q, r -> ...
+```
+
+There is in fact no special syntax for this: as parentheses are optional in
+tuples, the above is actually a match on a single value, the tuple `(x, y, z)`,
+against a single pattern, the pattern `(p, q, r)`. Normally, this would mean
+that `x`, `y` and `z` need to have the same mode. However, just as for locals,
```
riaqn, line 91: I think you mean `p, q, r` would have the same mode.

The mode of the tuple `(x, y, z)` would be determined by the modes of all components, which doesn't have to be the same; the mode of the tuple pattern `(p, q, r)` would be the same as the mode of `(x, y, z)`. Then, the mode of each of `p, q, r` would be the same as `(p, q, r)`.

riaqn: I think that "tuple modes being treated specially" is generic to all modes. But on top of that, UA also treats tuple specially in terms of what consistute a usage. For example, here, `(x, y, z)` is not considered as using the three values to construct a fresh tuple, which is later consumed by pattern matching. Instead, we directly match `x` to `p`, and `y` to `q`, and `z` to `r`, without counting any usage. I think this is worth telling the users.

(I think it's unavoidable to familiarize the users with the concept of "usages" in uniqueness).

jane/doc/extensions/uniqueness/reference.md
```diff
@@ -0,0 +1,103 @@
...
+may be unrelated.
+
+However, this changes when the tuple is explicitly named:
+
+```ocaml
+match x, y, z with
+| p, q, r as t -> ...
+```
+
+In this case, the OCaml compiler may actually allocate a tuple `(x, y, z)`at
```
riaqn, line 102: Similar to the comment above, what's new here is that the `as t` is a usage of all of `x, y, z` and ruins the ergonomics we attemped in the last example.

jane/doc/extensions/uniqueness/intro.md
riaqn: Good point - we should switch uniqueness/linearity printing to the new syntax. Can you do that? If you are too busy I can do that.

jane/doc/extensions/uniqueness/intro.md
riaqn: That's even better. Thanks!

jane/doc/extensions/uniqueness/intro.md
riaqn: Consider the example:
```
let x = ... in
if b then use_as_aliased x
else use_as_unique x
```
we have some `x` in the environment, and some `Pexp_ident x` in the expression. The former is a "value" while the later is a "usage/reference" of that "value. My point is that, it's often more useful to talk about uniqueness for the latter, not the former. In this example, there is only one value, but it's `unique` sometimes, and `shared` sometimes?

jane/doc/extensions/uniqueness/intro.md
anfelor: Sure, will do that later today

jane/doc/extensions/uniqueness/reference.md
anfelor: We could, but we should probably not have code in the documentation that throws errors for unused variables.

jane/doc/extensions/uniqueness/reference.md
anfelor: Right, it is a bit confusing because in C `free` only applies to the current block rather than the whole structure. Whereas here it applies to the children as well. I'll rephrase to make that clearer.

`store` is just some function that takes an `aliased` argument (perhaps to store it in a global reference somewhere). Let me know if you prefer a different name.

jane/doc/extensions/uniqueness/reference.md
anfelor: Yes, this discussion of tuple modes mirrors the discussion in the locality reference. And you are right, it is an interesting point that you can still refer to `x` after the match.

jane/doc/extensions/uniqueness/intro.md
anfelor: I am not saying that `Pexp_ident x` is itself a value and I realized that you were thinking about the `unique_use` trick. But I still think it is helpful to think of modes using the denotational semantics of what it means when a value has that mode. 

This is also how we teach locality in the `local/intro.md` file: `These [local] values live on a separate stack`. Following your argument, we should expect users to be confused that a `global` value may be used as `global` sometimes and as `local` another time. Would that document be clearer if we wrote: "The `local` mode designates a usage of a value that the value may not escape from"? No, because the escape-restriction is just an incidental restriction that is necessary to ensure the soundness of the denotational meaning. Similarly here, the restriction that the value may only have one unique use in the control flow establishes the denotational meaning that the value only has one reference pointing to it.

jane/doc/extensions/uniqueness/reference.md
riaqn: Instead of `store`, maybe just `use_as_aliased`? I'm also fine with `store` as long as it's explained.

jane/doc/extensions/uniqueness/intro.md
riaqn: I think a constrast that was interesting to me was that, `unique_use` contains `Uniqueness.r` for uniqueness, but `Linearity.l` for linearity. This is probably hinting at the duality between monadic and comonadic axes. Locality being comonadic probably doesn't fall within my previous argument.

> Following your argument, we should expect users to be confused that a global value may be used as global sometimes and as local another time.

If we have a `global` variable in the environment, and we use it as `local` somewhere, the program will continue to type check if that usage is strenghthened to `global`. Therefore, the "global-ness` seems to be a property of the variable/value that we can use at will.

The same can't be said for monadic axes. If we have a `uncontended` variable in the environment, and we use it as `contended` somewhere, the program might fail to type check if the usage is strengthenened to `uncontended`. Here it's probably more useful to talk about the contention of the two usages, rather than that of the bound variable.

anfelor: Right, the uniqueness and contention of a value changes as time goes on. That means that you often reason backwards: this operation expects a unique value so you look at the code before it to see if the value is still unique. You ensure that the value has only one reference at this point by making sure that there is no use coming before it.

FWIW, the opening paragraph now reads:

> The `unique` mode designates values that have only a single reference pointing
to them. If an operation takes a `unique` argument, it will consume the only
reference to the value. For example, an externally allocated data structure
might support a `free` operation:

Are you okay with that phrasing? What would you write instead?

jane/doc/extensions/uniqueness/intro.md
anfelor: #3196

jane/doc/extensions/uniqueness/reference.md
anfelor: Right. I have added a sentence to explain that a possible reference to `field1` remains.

jane/doc/extensions/uniqueness/intro.md
riaqn: > That means that you often reason backwards

Yes I guess the backwardness is what makes me prefer "usage" over "value", as a value goes forward only: a global value can always be used as global, no matter what.

I still prefer to say "unique designates usages". But in the end it's just a matter of English which is subjective. So I'm happy to resolve.

anfelor:
Since we are now removing the old syntax for `unique_` and `once_` completely (see #3196), I will remove that section from the introduction and then merge into main.

PR #3182, opened 2024-10-24
CSE join: canonicalise the primitives in both environments

### LINKED ISSUE #3181: Missed CSE opportunity
On current `main` (ab2272bc17f140d400fe7e443d26c97eae5bb9a9) with the following code:

```ocaml
let always_true x =
  if match x with None -> true | Some _ -> false then
    match x with None -> true | Some _ -> false
  else true
```

I get the following output in `-dflambda`:

```
After simplify:
((module_symbol Test.camlTest) (return_continuation k3) (exn_continuation k2)
 (toplevel_my_region toplevel_my_region/0N)
 (toplevel_my_ghost_region toplevel_my_ghost_region/1N)
 (used_value_slots { })
 camlTest__always_true_0_0_code = (Deleted)
 camlTest__always_true_0_1_code =
  ((code_metadata
    ((newer_version_of camlTest__always_true_0_0_code) (stub false)
     (inline Default_inline) () (poll_attribute Default) (is_a_functor false)
     (is_opaque false)
     (params_arity 𝕍=Variant((consts ({ 0 })) (non_consts ({(0 [*])}))))
     (param_modes (Heap)) (first_complex_local_param 1)
     (result_arity 𝕍=tagged_ℕ𝕚) (result_types (Unknown)) (result_mode Heap)
     (contains_no_escaping_local_allocs true) (recursive Non_recursive)
     (cost_metrics
      size: 32 removed: {call: 0 alloc: 0 prim: 14 branch: 1 direct: 0
                         poly_cmp: 0 requested: 0})
     (inlining_arguments Oclassic) (dbg test.ml:1,16--133) (is_tupled false)
     (is_my_closure_used false)
     (inlining_decision (Function_body_too_large 10))
     (loopify Never_loopify)))
   (λ〈k13〉《k12》⟅my_region/27N⟆ ⟅my_ghost_region/26N⟆
      (x/29UV ∷ 𝕍=Variant((consts ({ 0 })) (non_consts ({(0 [*])})))) 
      (my_closure/28N ∷ 𝕍) my_depth/25N .
    (prim/32N = ((Is_int x/29UV) test.ml:2,18--22)
     (switch prim/32N test.ml:2,18--22
     | 0 ↦ apply_cont k16 #0
     | 1 ↦ apply_cont k16 #1)
     k16 (naked_immediate/37N ∷ ℕ𝕚) #More_than_one:
     ((switch naked_immediate/37N test.ml:2,18--22
      | 0 ↦ return k13 1
      | 1 ↦ goto k17 test.ml:3,17--21)
      k17 #One:
      prim/41N = ((Is_int x/29UV) test.ml:3,17--21)
      (switch prim/41N test.ml:3,17--21
      | 0 ↦ return k13 0
      | 1 ↦ return k13 1))))))
 Test.camlTest__always_true_1 ↤ (always_true/0 ∷ 𝕍) =
  (set_of_closures Heap
   ({((always_true/0 ∷ 𝕍) camlTest__always_true_0_1_code)}))
 Test.camlTest =
  (Immutable_block (tag 0) (shape Value_only) (Test.camlTest__always_true_1))
 module_init_end k3 Test.camlTest)
```

Both `prim/41N` and `prim/32N` are equal to `Is_int x/29UV`, and `prim/32N` dominates `prim/41N`, so I'd expect CSE to kick in and eliminate `prim/41N`, but evidently it does not.

lthls:
Fixes #3181.

As explained in a comment there, when joining CSE equations we want to transform equations at the use site into equations at the join site (in the env at fork). There are multiple ways to do that: all aliases of primitive arguments that are in scope in the env at fork could be considered, which would yield the most precise result but risks exploding the number of CSE equations.
Before this PR we canonicalise in the environment at use, which handles the cases where primitives were called on local aliases of a "global" variable.
But we're failing in cases like the one from #3181, where a CSE equation is common to all branches but some branches have equations that change which element is canonical.
So this PR proposes to canonicalise in the env at fork in addition to the env at use. This potentially doubles the number of primitives, but in the common case where the branch doesn't introduce new relevant aliases the primitives are the same (using `EP.Set.t` to represent these possibilities means we don't actually have to check that manually).

### DIFF
--- a/middle_end/flambda2/simplify/common_subexpression_elimination.ml
+++ b/middle_end/flambda2/simplify/common_subexpression_elimination.ml
@@ -161,11 +161,10 @@ let cse_with_eligible_lhs ~typing_env_at_fork ~cse_at_each_use ~params prev_cse
       in
       EP.Map.fold
         (fun prim bound_to eligible ->
-          let prim =
+          let canonicalise env prim =
             EP.filter_map_args prim ~f:(fun arg ->
                 match
-                  TE.get_canonical_simple_exn env_at_use arg
-                    ~min_name_mode:NM.normal
+                  TE.get_canonical_simple_exn env arg ~min_name_mode:NM.normal
                     ~name_mode_of_existing_simple:NM.normal
                 with
                 | exception Not_found -> None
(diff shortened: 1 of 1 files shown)


lthls:
@mshinwell You will probably want to look at this, although we should also discuss with @chambart next Tuesday.

mshinwell:
Let's add some more documentation about the subtleties of CSE as discussed.  I will approve now, just merge once that's done.

mshinwell (approved):

lthls:
After some more discussions, it looks like there is a deeper problem that we need to think about properly. Since the bug report doesn't come for real code, we're putting this PR on hold until we have a better idea of how to solve the problem correctly.

PR #3184, opened 2024-10-24
Make objects crosses modes

riaqn:
Objects are currently defined/produced/allocated at `legacy`, this allows an "actual"/left object to cross to `global` and `many` as needed.

Note that objects are also defined at `uncontended`, but that doesn't mean objects can cross to `uncontended`. For example:
```
let obj = new cla in
let (foo @ portable) () = .. obj#method ..
```
This shouldn't type check because (among other reasons) `obj` is used inside the closure as `uncontended`, which forces the closure to be `nonportable`.

In general, objects cannot cross monadic axes, because the mode crossing of objects are solely based on how they are defined at a fixed mode, which only affects comonadic axes.

### DIFF
--- a/testsuite/tests/typing-local/alloc.ml
+++ b/testsuite/tests/typing-local/alloc.ml
@@ -460,10 +460,15 @@ class cla = object
     val x = 42
 end
 
-let[@inline never] obj () =
+let obj () =
   ignore_local (new cla);
   ()
 
+let obj_direct () =
+  ignore_local (object
+    val x = 42
+  end);
+  ()
 
 let run name f x =
   let prebefore = Gc.allocated_bytes () in
@@ -519,7 +524,8 @@ let () =
   run "manylong" makemanylong 100;
   run "optionalarg" optionalarg (fun_with_optional_arg, 10);
   run "optionaleta" optionaleta ();
-  run "object" obj ()
+  run "object" obj ();
+  run "object_direct" obj_direct ()
 
   (* The following test commented out as it require more memory than the CI has
      *)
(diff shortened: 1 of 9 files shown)


testsuite/tests/typing-modes/class.ml
goldfirere, line 1: Is there a test anywhere that makes sure that objects are allocated on the heap? The correctness of mode-crossing locality depends on them always being on the heap, but I don't think that aspect is tested.

typing/typecore.ml
```diff
@@ -9362,7 +9365,8 @@ and type_immutable_array
 
 (* Typing of method call *)
 and type_send env loc explanation e met =
-  let obj = type_exp env mode_legacy e in
+  let expected_mode = expect_mode_cross_jkind Jkind.for_object mode_legacy in
```
goldfirere, line 9368: This is just a constant mode, right? If that's correct, maybe lift this outside of `type_send` to make the expected mode for an object in a send-expression more obviously constant.

testsuite/tests/typing-local/alloc.ml
```diff
@@ -456,6 +456,15 @@ let[@inline never] huge () =
   bytes_set b pos 'h';
   assert (bytes_get b pos = 'h')
 
+class cla = object
+    val x = 42
+end
+
+let[@inline never] obj () =
+  ignore_local (new cla);
+  ()
```
goldfirere, line 465: Could you also add a direct creation of an object with the `object ... end` syntax?

goldfirere (approved):

PR #3188, opened 2024-10-28
Add amd64 instrinsic for `CLDEMOTE`

Xyene:
`CLDEMOTE` is a hint to hardware that a cacheline should be demoted from L1/L2 to L3 (as opposed to `CLFLUSH(OPT)`, which demotes to RAM).

This can improve the performance of producer-consumer systems.

The instruction decodes as a no-op on hardware that does not support it.

Testing:

```
$ cat cldemote.ml
external cldemote : 'a -> unit = "caml_cldemote_ignore" "caml_cldemote" [@@noalloc] [@@builtin]
let () = cldemote "Hello, World!"

$ cat cldemote_stubs.c
CAMLprim value caml_cldemote_ignore(value v) { return Val_unit; }
CAMLprim value caml_cldemote(value v) { return Val_unit; }

$ ocamlopt.opt -c cldemote_stubs.c
$ ocamlopt.opt -c cldemote.ml
$ ocamlopt.opt -o cldemote cldemote_stubs.o cldemote.cmx

$ ./cldemote
; No SIGILL, even on machines that are too old to have `CLDEMOTE`.

$ objdump -d cldemote | grep '<camlCldemote__entry>:' -A4
000000000041e7a0 <camlCldemote__entry>:
  41e7a0:       48 8d 05 d9 0b 04 00    lea    0x40bd9(%rip),%rax        # 45f380 <camlCldemote+0x8>
  41e7a7:       0f 1c 00                cldemote (%rax)
  41e7aa:       b8 01 00 00 00          mov    $0x1,%eax
  41e7af:       c3                      retq
```

### DIFF
--- a/backend/amd64/CSE.ml
+++ b/backend/amd64/CSE.ml
@@ -41,6 +41,7 @@ method! class_of_operation op =
       | Pure -> Op_pure
       end
     | Ipause
+    | Icldemote _
     | Iprefetch _ -> Op_other
     end
   | Imove | Ispill | Ireload
@@ -84,6 +85,7 @@ class cfg_cse = object
       | Pure -> Op_pure
       end
     | Ipause
+    | Icldemote _
     | Iprefetch _ -> Op_other
       end
   | Move | Spill | Reload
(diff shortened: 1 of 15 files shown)


gretay-js (approved):
`CLDEMOTE` has its own cpu feature flag, but this PR doesn't add it to `Arch.Extension`. This is fine because `CLDEMOTE` instruction is decoded as a NOP on cpus that's have this feature.

PR #3190, opened 2024-10-28
Comprehensions layouts bug

ccasin:
(**Note**: This PR is based on the "delete jane syntax" branch to avoid conflicts and should be merged after that one.)

Right now the front-end prevents you from constructing an array/list of unboxed things with a comprehension, but not from iterating over one.  But `transl` assumes that the front-end does prevent this, making it possible to hit the scary "contact us" sanity check error in `value_kind`.

There are two commits here: the first demonstrates the bug with a test (as far as I can tell, there weren't any tests at all for unboxed things in comprehensions).  The second fixes it.

I have fixed this by adding the appropriate front-end check. It's quite possible this would just work if we instead threaded through the appropriate sort, but that should come with extensive tests that I don't think it is worth doing right now.

I noticed this because of a request to add a comprehensions test to #3139 

Review: @goldfirere

### DIFF
--- a/testsuite/tests/typing-layouts/comprehensions.ml
+++ b/testsuite/tests/typing-layouts/comprehensions.ml
@@ -1,0 +1,67 @@
+(* TEST
+ include stdlib_upstream_compatible;
+ {
+   flags = "-extension comprehensions";
+   expect;
+ }
+*)
+
+open Stdlib_upstream_compatible
+
+(* Unboxed types are banned in comprehensions: The output array must contain
+   things of layout value, as must any arrays you iterate over. *)
+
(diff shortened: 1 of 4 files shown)


ncik-roberts:
Jane Syntax's deletion is merged.

goldfirere:
It looks like this still needs a rebase or something, though...

ncik-roberts:
Indeed, that was meant to be a ping for @ccasin to rebase, not for you to review.

mshinwell:
@ccasin what is happening with this?

goldfirere (approved):

PR #3073, opened 2024-09-25
Selection: CFG pipeline

xclerc:
This pull request implements the missing
parts of the CFG pipeline for the selection
pass. More testing is needed, but we
should merge this version, given it is off
by default and behind a flag.

For review of the `Cfg_selectgen` module,
there are two kinds of functions and
methods: the functions which are just
straighforward translation of their Mach
counterparts, and the one which are indeed
CFG-specific.

In the first category:

- `maybe_emit_naming_op`;
- `join`;
- `join_array`;
- `insert_move`;
- `insert_return`;
- `emit_return`.

In the second category:
- `emit_expr_aux_op`;
- `emit_expr_aux_ifthenelse`;
- `emit_expr_aux_switch`;
- `emit_expr_aux_catch`;
- `emit_expr_aux_exit`;
- `emit_expr_aux_trywith`;
- `emit_tail_apply`;
- `emit_tail_ifthenelse`;
- `emit_tail_switch`;
- `emit_tail_catch`;
- `emit_tail_trywith`;
- `emit_fundecl`.

The translation is based on the notion of
a "sub cfg" (`Sub_cfg` module), which is
essentially a list of blocks, with one entry
point and one exit point. It is akind to
the list of instructions used by the existing
pipeline, and elements are added to the
exit block.

The pull request also adds a new CI job,
namely "cfg-selection", to test the new
pipeline (with CSE and zeroalloc also on
the CFG representation).

PR #3192, opened 2024-10-29
Minor refactoring of `Cfg_selectgen`

xclerc:
Follow-up to https://github.com/ocaml-flambda/flambda-backend/pull/3073.

This pull requests:

- adds a short description of the `Sub_cfg` module;
- changes `assert false` to more useful errors;
- removes the `XXX` "marker" from parameter names (leaving a CR instead);
- factors out the construction of blocks with a `Never` terminator.

### DIFF
--- a/backend/cfg_selectgen.ml
+++ b/backend/cfg_selectgen.ml
@@ -130,13 +130,17 @@ let next_instr_id () : int =
   incr next_instr_id;
   res
 
-(* A "sub" CFG is the counterpart of an instruction list in the original
+(* A "sub" CFG is the counterpart of an instruction list in the original Mach
    selection pass.
 
    It is essentially a collection of blocks (stored as a layout, i.e. as a
-   list), with two designated block: - an entry block; - an exit block.
+   list), with two designated blocks:
 
-   The exit block is where mere instructions are being added, which means that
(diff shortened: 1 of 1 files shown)


backend/cfg_selectgen.ml
```diff
@@ -130,6 +130,16 @@ let next_instr_id () : int =
   incr next_instr_id;
   res
 
+(* A "sub" CFG is the counterpart of an instruction list in the original
```
mshinwell, line 133: Suggestion:
```
(* A "sub" CFG is the counterpart of an instruction list in the original
   Mach selection pass.

   It is essentially a collection of blocks (stored as a layout, i.e. as a
   list), with two designated blocks:

   - an entry block;
   
   - an exit block.

   The exit block is where more instructions are being added, which means that
   the terminator of an in-construction "sub" CFG is `Never`, and will be
   changed only when no additional instructions will be inserted to the
   block. *)
```

mshinwell (approved):
OK apart from a few minor things in the first comment, I left a new version.  Fine to merge after that change.

backend/cfg_selectgen.ml
xclerc: (applied)

PR #3193, opened 2024-10-29
Delete new caml_string_set* primitives

dvulakh:
We treat `string`s as immutable, so we don't plan to expose the new UBT-related `string` setters in our libraries. We may as well remove them from the compiler as well.

We keep the string setters that exist in upstream (`caml_string_set{16,32,64}{,u}`) and the `caml_string_set` runtime function for campatibility reasons.

### Review
The first commit adds the string-setting primitives to the string `non_existent_primitives` test file. The second deletes them from the compiler. Their absence is then reflected in the `non_existent_primitives` test.


testsuite/tests/prim-bigstring/non_existent_primitives.ml
```diff
@@ -41,3 +43,499 @@ Lines 1-2, characters 0-26:
...
+external set_string : string -> int64# -> float32# -> unit
+  = "%caml_string_setf32u#_indexed_by_int64#"
+
+[%%expect
+{|
+Lines 1-2, characters 0-45:
+1 | external set_string : string -> int64# -> float32# -> unit
+2 |   = "%caml_string_setf32u#_indexed_by_int64#"
+Error: Unknown builtin primitive "%caml_string_setf32u#_indexed_by_int64#"
+|}]
```
ncik-roberts, line 541: I don't think we need 8 tests of this. (I didn't count.) 1 seems enough.

ncik-roberts:
This seems good to me, but I suggested the change and someone else should give their opinion. @mshinwell, could you take a look too, please?

mshinwell:
I haven't looked at the implementation, but the idea seems fine.

ncik-roberts (approved):
Thanks for seconding the idea. The implementation seems correct. I'm not a fan of the design of "generate a bunch of duplicates but then use the string  name to dedup", but this approach appears to predate this PR and probably now is not the right time to ask for a refactor.

PR #3196, opened 2024-10-29
Remove unique_ and once_ keywords from printer

anfelor:
This commit changes the printing of modes in the type checker so that the new mode syntax is printed if the extension is enabled (scroll down all the way to see the only code change in `printtyp.ml`). This will be helpful for the uniqueness documentation, so that we do not have to introduce the `unique_` and `once_` keywords.

~~A possible drawback of this PR is that it changes the printing of `local_` to be `@ local`. While we probably want to make that change at some point, it could change many expect-tests across the firm and we might not be ready for that yet. Do let me know if you want me to change this aspect.~~

### DIFF
--- a/testsuite/tests/lib-array/test_iarray.ml
+++ b/testsuite/tests/lib-array/test_iarray.ml
@@ -32,16 +32,16 @@ let rec list_map_local_input (local_ f) (local_ list) =
 module Iarray = Stdlib_stable.Iarray
 external ( .:() ) : 'a iarray -> int -> 'a = "%array_safe_get"
 val iarray : int iarray = [:1; 2; 3; 4; 5:]
-val iarray_local : unit -> local_ int iarray = <fun>
+val iarray_local : unit -> int iarray @ local = <fun>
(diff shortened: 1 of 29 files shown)


riaqn:
I think for now we only want to switch to the new syntax for uniqueness/linearity. We also want a follow-up PR that stop parsing `unique_` and `once_`, before announcing uniqueness extension to the users (so users never use the old syntax, which is headache to migrate).

For locality (which is widely used already) we need a delicate plan for transition to the new syntax; let's avoid that in this PR.

anfelor:
Okay, I have removed the `local` change for now. I agree that we should also remove `unique_` and `once_` from the syntax. Do you think we should only remove it from types (where we can now use the new syntax) or also from terms?

riaqn:
> Do you think we should only remove it from types (where we can now use the new syntax) or also from terms?

I think we want to remove `unique_` and `once_` from `parsing/lexer.mll`, which is to remove them completely from the language.
(we have the new syntax for terms as well. https://github.com/ocaml-flambda/flambda-backend/blob/main/jane/doc/extensions/modes/syntax.md)

riaqn (approved):
:shipit:

PR #3197, opened 2024-10-29
Improve comments of uniqueness analysis

riaqn:
Improves the comments on uniqueness analysis.

Review: @anfelor

### DIFF
--- a/typing/uniqueness_analysis.ml
+++ b/typing/uniqueness_analysis.ml
@@ -114,7 +114,7 @@ end = struct
   let mark_multi_use l =
     let force_one ((uni, lin), occ) =
       (* values being multi-used means two things:
-         - the expected mode must be higher than [aliased]
+         - the expected mode must be higher than Aliased
          - the access mode must be lower than [many] *)
       match Linearity.submode lin Linearity.many with
       | Error _ -> Error { occ; axis = Linearity }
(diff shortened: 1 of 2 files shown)


typing/uniqueness_analysis.ml
```diff
@@ -235,21 +235,41 @@ module Usage : sig
...
+      let x = .. in
+      use_as_unique x;
+      use_as_unique x
+      ]
+
+      the [Typedtree.unique_use] of both [Pexp_ident x] would say "need unique",
+      but the analysis infers that [x] is used multiple times and thus each use
+      site can only use it as shared. Error is then raised.
+
+      It is sound for the analysis to infer a [Pexp_ident] to have a
```
anfelor, line 267: s/[Pexp_ident]/use

(since we also treat field reads this way)

typing/uniqueness_analysis.ml
```diff
@@ -235,21 +235,41 @@ module Usage : sig
...
+      weaker/higher usage, which might result in false mode errors. It is useful
```
anfelor, line 268: > It is sound for the analysis to infer a [Pexp_ident] to have a weaker/higher usage, which might result in false mode errors.

I don't understand this sentence. It sounds to me like the analysis could always mark any use as `Unused`, which would be unsound.

typing/uniqueness_analysis.ml
```diff
@@ -264,14 +284,13 @@ end = struct
 
        and as a result, we can delay the projection at `x`.
 
-       The downside of relaxing is the loss of completeness: if we relax too
-       much the program will fail type check. In the extreme case we relax it to
-       Error which fails type check outright (and extremely sound, hehe).
+      On the other hand, the analysis shouldn't infer a [Pexp_ident] to have a
+      too strong usage, as that might be unsound.
```
anfelor, line 288: I don't understand this: it should always be sound to assume a use is unique instead of aliased? Or to assume that a use is aliased instead of borrowed? The old text seemed more correct there.

typing/uniqueness_analysis.ml
```diff
@@ -235,21 +235,41 @@ module Usage : sig
...
+      For example:
+      [
+      let x = .. in
+      use_as_unique x;
+      use_as_unique x
+      ]
+
+      the [Typedtree.unique_use] of both [Pexp_ident x] would say "need unique",
+      but the analysis infers that [x] is used multiple times and thus each use
+      site can only use it as shared. Error is then raised.
```
anfelor, line 265: I am not sure that this example adds much insight here. Perhaps we can say this in a more compressed way?

typing/uniqueness_analysis.ml
```diff
@@ -264,14 +284,13 @@ end = struct
...
-     - The purpose of this uniqueness analysis is to figure out the most relaxed
-     mode for each use, such that we get the best performance, while still
-     type-check. Currently there are really only two choices worth figuring out,
-     Namely
+     - Therefore, the analysis should try to figure out the strongest/lowest
+     usage of each [Pexp_ident], such that we get the best performance, while
+     still being sound. Currently there are really only two choices worth
+     figuring out, namely:
      - borrowed or aliased?
      - aliased or unique?
```
anfelor, line 296: Another comment I wrote over in #3157 , which you might be interested in:

```
(** Usage algebra
    In this file we track the usage of variables and tags throughout a source file.
    Information can be composed using three operators:
     - seq: sequential composition such as 'foo; bar'
     - par: parallel composition such as '(foo, bar)'
         where evaluation order is not specified.
     - choose: non-deterministic choice such as 'if b then foo else bar'
    subject to the following laws:
     - seq, par, choice are associative
     - par, choice are commutative
     - seq and par both distribute over choice
     - choice is idempotent
     - seq, par, choice have a common unit 'empty'
    Note: These laws do not apply regarding the concrete error messages reported
    to the user if the analysis fails. However, the laws do determine whether
    the analysis may fail in the first place.
    Aside: This structure is also known as a United Monoid (Mokhov et al.)
    It differs from, say, Kleene algebra in that seq and choice share a unit.
    To make this a Kleene algebra we would need a unit for choice that
    annihilates seq. In most presentations, this corresponds to a code path
    that can never be taken like 'assert false'. But we want to perform
    the uniqueness analysis even such branches, since it would be a bad
    user experience if a wrong use of uniqueness could lead to a segfault
    before reporting the failed assertion. *)
```

I include it here in case it inspires you and not asking you to include it -- feel free to resolve this as soon as you read it. In fact, I would prefer we just merge #3157 instead of adopting the comment now.

typing/uniqueness_analysis.ml
riaqn: Good point. How about 
> .. infer a usage (such as [Pexp_ident]) to have a

Because I think having examples are more readable.

typing/uniqueness_analysis.ml
riaqn: For example:
```
let x = .. in
use_uniquely x;
use_uniquely x;
```
The analysis should infer both usages of `x` to be aliased (which later conflicts with `unique_use` on `Texp_ident` as expected). It is unsound to infer the usages to be `unique` (which later agrees with `unique_use` on `Texp_ident`, and doesn't trigger mode error)

typing/uniqueness_analysis.ml
riaqn: Thanks - I will adopt that.

> In fact, I would prefer we just merge https://github.com/ocaml-flambda/flambda-backend/pull/3157 instead of adopting the comment now.

I feel it's better to have a separate PR to improve the comments that are unrelated to overwriting.

typing/uniqueness_analysis.ml
anfelor: I think I would prefer it if we explained at the top of the comment what a usage is and referred the reader to `[check_uniqueness_exp_as_value]`. Then we can simply write "usage" in the rest of the comment without extra examples.

typing/uniqueness_analysis.ml
anfelor: Okay, I see what you are saying. But I find it confusing: it would not be sound for the analysis to map both uses of `x` to `Aliased` because `Aliased` does not contain a unique barrier. Instead, it would be sound to infer both uses to be unique and then _cast_ them to aliased in a process that forces the unique barrier to be aliased. That is exactly what our `seq` function does.

On the other hand, it seems to me that it would always be sound to infer an aliased usage as unique. This rejects lots of valid programs but I don't see how it could be unsound?

typing/uniqueness_analysis.ml
anfelor: > If the analysis infer a [Pexp_ident] to be Unused, that will later be compared against the unique_use on the Pexp_ident

Where would this happen?

I think we should avoid discussing alternative designs and focus on the code that we have.

riaqn: This happens in `Maybe_unique.mark_multi_use`.

riaqn: > Where would this happen?

This doesn't happen - but just responding to your comment above.

> I think we should avoid discussing alternative designs and focus on the code that we have.

I agree - therefore I didn't go into details in the text. But I do think it's helpful to mention the direction of soundness/completeness.

typing/uniqueness_analysis.ml
anfelor: Okay!

typing/uniqueness_analysis.ml
riaqn: When the uniqueness analysis sees a `Texp_ident`, there is a duality that causes all the confusion.

- The `Texp_ident` contains a `unique_use` which contains a `Uniqueness.r` (we will ignore linearity for simplicity). If this mode (we call it `a`) is `Aliased`, it's sound to relax it to `Unique` - we are pretending that the consumer of this `Pexp_ident` is requiring a stricter mode, which is sound.
-  The lexical analysis, not seeing any `unique_use`, might decide that a `Pexp_ident` is the only usage and allow it to be `Unique` (we call it `b`). It's sound to relax it to `Aliased` - we are pretending that there are multiple uses of the value and thus less capability is granted to the usage.

The UA ensures that `b` is stronger than `a`. I.e., the uniqueness capability allowed by the lexical structure, is stronger than the uniqueness capability requested by the consumer of the value.

It's sound to make `b` weaker, or to make `a` stronger. It's the same thing.

I agree the text is currently confusing and I will clarify as above.

typing/uniqueness_analysis.ml
goldfirere: I like the example -- in general, I find examples really helpful in understanding comments such as these.

typing/uniqueness_analysis.ml
anfelor: How does the reasoning in the second bullet work when you relax `Error` to `Aliased` or when you relax `Aliased` to `Unused`?

riaqn: From the perspective of the lexical analysis:

If the analysis decides that a `Texp_ident` is `Aliased`, it's sound to relax that to `Unused`.

To better understand this, imagine we add `Unused` (weaker than `Aliased`) to `Mode.Aliasability`, and we have code:
```
let foo (a @ unused) = () in
foo x
``` 
The lexical analysis will decide the `Texp_ident x` should be `Aliased`. Furthermore, it's sound to relax further to `Unused`, which mean a submode call `x.uniqueness >= Unused`, which succeeds. Of course in general, the relaxing would cause submode failure - but it's always sound to do that.

riaqn: What's the business of `Error`? Here is my attempt. Imagine we add a bottom element to uniqueness and call it `Error`. 
Moreover, our UA is a trivial implementation that behaves like this:
1. its thinks all `Pexp_ident x`, by lexical structure that it doesn't even examine, can enjoy the best uniqueness `Error`. This means we will check `x.uniqueness >= Error` which always succeeds.
2. These `Error` usages combines (via `seq` and `par`, etc.) into a single `Error` in the end, and throws type error

This UA is sound. It's also sound if in step 1 we check `x.uniqueness >= Aliased` (by relaxing Error to Aliased). It would raise even more type error, but it's sound to do so.

riaqn:
We had an offline discussion. I've revised according to the discussion. @anfelor Can you take another look?

anfelor (approved):

PR #3198, opened 2024-10-29
Map unique_barrier information to Reads_vary (cherry-picked from #3066)

anfelor:
@goldfirere and I discussed today that we would like to have the full unique barrier information on `main` before we make uniqueness stable. While it is unlikely that users could actually exploit the bad behaviour around pushing down projections with just uniqueness (and not overwriting), it might be possible and we would like to be on the safe side.

This PR cherry-picks the merge commit of #3066, which was fully reviewed by @goldfirere.

### DIFF
--- a/chamelon/compat.jst.ml
+++ b/chamelon/compat.jst.ml
@@ -279,6 +279,18 @@ let view_texp (e : expression_desc) =
   | Texp_match (e, sort, cases, partial) -> Texp_match (e, cases, partial, sort)
   | _ -> O e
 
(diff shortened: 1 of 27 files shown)


goldfirere (approved):
Did anything interesting happen in the backport? Any conflicts that were interesting to resolve?

anfelor:
Only one: in `type_tuple_pat` of `typecore.ml` we added a uniqueness barrier to a pattern that is created after matching on `| Jpat_layout (Lpat_constant cst)`. On `main`, that branch was deleted.

PR #3200, opened 2024-10-30
Const_null

dkalinichenko-js:
Implement `Const_null` in Lambda and `Null` in Flambda. The kinds for the constant could be more precise.

### DIFF
--- a/middle_end/flambda2/types/expand_head.ml
+++ b/middle_end/flambda2/types/expand_head.ml
@@ -366,7 +366,10 @@ let expand_head_of_alias_type env kind
       | Naked_nativeint i ->
         ET.create_naked_nativeint (TG.Head_of_kind_naked_nativeint.create i)
       | Naked_vec128 i ->
-        ET.create_naked_vec128 (TG.Head_of_kind_naked_vec128.create i))
+        ET.create_naked_vec128 (TG.Head_of_kind_naked_vec128.create i)
+      | Null ->
+        ET.create_unknown
+          Flambda_kind.value (* CR layouts v3: assign a precise kind? *))
     ~name
 
 let expand_head0 env ty ~known_canonical_simple_at_in_types_mode =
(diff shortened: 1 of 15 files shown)


middle_end/flambda2/types/reify.ml
```diff
@@ -72,7 +72,7 @@ let try_to_reify_fields env ~var_allowed alloc_mode
               | Tagged_immediate _imm -> Some simple
               | Naked_immediate _ | Naked_float _ | Naked_float32 _
               | Naked_int32 _ | Naked_vec128 _ | Naked_int64 _
-              | Naked_nativeint _ ->
+              | Naked_nativeint _ | Null ->
```
lthls, line 75: I think that's wrong, but because this whole `const` case is wrong (independently of Null cases). I will submit a fix separately.

middle_end/flambda2/types/grammar/more_type_creators.ml
```diff
@@ -285,6 +285,7 @@ let type_for_const const =
   | Naked_int64 n -> TG.this_naked_int64 n
   | Naked_nativeint n -> TG.this_naked_nativeint n
   | Naked_vec128 n -> TG.this_naked_vec128 n
+  | Null -> TG.any_value (* CR layouts v3: assign a precise kind? *)
```
lthls, line 288: This is not great; I will push a commit with a proper type.

middle_end/flambda2/types/expand_head.ml
```diff
@@ -366,7 +366,10 @@ let expand_head_of_alias_type env kind
       | Naked_nativeint i ->
         ET.create_naked_nativeint (TG.Head_of_kind_naked_nativeint.create i)
       | Naked_vec128 i ->
-        ET.create_naked_vec128 (TG.Head_of_kind_naked_vec128.create i))
+        ET.create_naked_vec128 (TG.Head_of_kind_naked_vec128.create i)
+      | Null ->
+        ET.create_unknown
+          Flambda_kind.value (* CR layouts v3: assign a precise kind? *))
```
lthls, line 372: Same comment as in `more_type_creators.ml`

middle_end/flambda2/types/reify.ml
lthls: (It's in #3206)

PR #3201, opened 2024-10-30
Is_null

dkalinichenko-js:
Implement `Pisnull` in Lambda and `Is_null` in Flambda. This needs a proper simplifier.

### DIFF
--- a/lambda/translprim.ml
+++ b/lambda/translprim.ml
@@ -522,6 +522,7 @@ let lookup_primitive loc ~poly_mode ~poly_sort pos p =
     | "%floatarray_unsafe_set" ->
       Primitive ((Parraysetu (Pfloatarray_set, Ptagged_int_index)), 3)
     | "%obj_is_int" -> Primitive (Pisint { variant_only = false }, 1)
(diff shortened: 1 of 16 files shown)


lambda/translprim.ml
```diff
@@ -522,6 +522,7 @@ let lookup_primitive loc ~poly_mode ~poly_sort pos p =
     | "%floatarray_unsafe_set" ->
       Primitive ((Parraysetu (Pfloatarray_set, Ptagged_int_index)), 3)
     | "%obj_is_int" -> Primitive (Pisint { variant_only = false }, 1)
+    | "%is_null" -> Primitive (Pisnull, 1)
```
lthls, line 525: Do you really need to expose the primitive ? I suppose it makes sense to have it for testing, but I'm assuming that users should always the pattern-matching on the `or_null` type (or other nullable types in the future).

lthls (commented):
Everything looks fine, although I hope the `Translprim` patch isn't actually necessary.

lambda/translprim.ml
dkalinichenko-js: Yeah, it's just for testing.

PR #3204, opened 2024-10-30
Add partial rebuilding option to ART + some associated refactoring

Gbury:
(this is a pre-requisite for the "match-in-match" optimisation)

Instead of encoding a simple boolean, the [Are_rebuilding_terms.t], now also encodes whether we are rebuilding everything (as is almost always the case), or whether we are only rebuilding a small part of the unit (as will be the case for match-in-match/continuation specialization).

With this changes comes a small refactoring, removing the boolean conversions to/from [Are_rebuilding_terms] which would be misleading with the thirs option present, and renaming some functions for better clarity and consistency.

Finally, the [ART] in [uenv] is only there for printing, could we find a way to not need it ?

### DIFF
--- a/middle_end/flambda2/simplify/env/upwards_env.mli
+++ b/middle_end/flambda2/simplify/env/upwards_env.mli
@@ -18,6 +18,9 @@
 
 type t
 
+(** Create an upwards environment.
+
+    The [are_rebuilding_terms] provided is only used for printing. *)
 val create : Are_rebuilding_terms.t -> t
 
 val print : Format.formatter -> t -> unit
--- a/middle_end/flambda2/simplify_shared/are_rebuilding_terms.mli
+++ b/middle_end/flambda2/simplify_shared/are_rebuilding_terms.mli
@@ -14,21 +14,29 @@
 (*                                                                        *)
 (**************************************************************************)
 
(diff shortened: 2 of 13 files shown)


middle_end/flambda2/simplify/env/upwards_env.mli
```diff
@@ -18,6 +18,9 @@
 
 type t
 
+(** Create an upwards environment.
+
+    The [are_rebuilding_terms] provided is only used for printing. *)
```
mshinwell, line 23: I don't think this should be a specification in the interface.

middle_end/flambda2/simplify_shared/are_rebuilding_terms.mli
```diff
@@ -19,10 +19,16 @@
...
-val print : Format.formatter -> t -> unit
+val are_not_rebuilding : t -> bool
+
+val are_rebuilding_partially : t -> bool
+
+val rebuild_nothing : t
+
+val rebuild_everything : t
+
+val partial_rebuilding : t
```
mshinwell, line 34: We need some comments here

Gbury:
Marking this as a Draft since we have a better idea for how to do match-in-match at the moment.

PR #3205, opened 2024-10-30
Distinguish between "light" and "heavy" flambda2 invariants

bclement-ocp:
Light invariants can be checked in the CI, while heavy invariants are intended for local debugging. More precisely:

 - Light invariants must be fast to check and have low complexity. A guideline is that it is acceptable to call `find` or `mem` functions, but they should not use any kind of iterations iterations (`iter`, `fold`, `exists`, etc.). It must be viable to enable light invariants in regular user workflows without significantly impacting the performance of the compiler.

   Light invariants are enabled by the pre-existing `-dflambda-invariants` compiler flag, which is also enabled with `--enable-flambda-invariants` configure flag.

   To determine in code whether light invariants should be checked, call the function `Flambda_features.check_light_invariants ()`.

   Light invariants are checked on CI jobs, which should prevent drifts between the expressed invariants and the actual invariants.

 - Heavy invariants are allowed to be (almost) unbounded in complexity. They are not intended to be enabled during regular workflows, but rather to be used to help identify the source of a bug once it has been identified that one exists. It is fine if enabling heavy invariants drastically slow down the compiler on some inputs.

   Heavy invariants are enabled by a new `-dflambda-heavy-invariants` compiler flag. This flag implies `-dflambda-invariants`, so light invariants are always checked when heavy invariants are checked.

   To determine in code whether heavy invariants should be checked, call the function `Flambda_features.check_invariants ()`.

   Heavy invariants are *not* checked on CI jobs at the moment. They currently include invariants that are known to be broken and should be investigated before being moved to the light invariants.

The name of the compiler flags (invariants/heavy invariants) and feature checks (light invariants/invariants) is chosen to reflect that users should usually enable light checks only, but calling `check_light_invariants ()` is a reminder to make sure the invariant does not do any expensive computation.

A few invariants related to the typing env and aliases that currently hold and perform no iterations are moved to the set of light invariants. The goal is that invariants are slowly moved from the heavy set to the light set over time when it seems useful.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,7 +55,7 @@ jobs:
             ocamlparam: '_,O3=1'
 
           - name: flambda2_o3_advanced_meet_frame_pointers_runtime5_polling
-            config: --enable-middle-end=flambda2 --enable-frame-pointers --enable-runtime5 --enable-poll-insertion
+            config: --enable-middle-end=flambda2 --enable-frame-pointers --enable-runtime5 --enable-poll-insertion --enable-flambda-invariants
             os: ubuntu-latest
             build_ocamlparam: ''
             ocamlparam: '_,O3=1,flambda2-meet-algorithm=advanced,flambda2-expert-cont-lifting-budget=200'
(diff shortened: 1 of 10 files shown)


chambart (approved):
This looks good.

If those invariants hold, it will be a nice improvement. If they don't, it will be a very nice thing to know !

mshinwell (approved):

PR #3206, opened 2024-10-30
Fix reification of mixed blocks and float blocks

lthls:
See added example for a case where lifting fails.

The code used to only work for blocks full of values, when we added support for mixed blocks we failed to update the support for constants.


mshinwell:
cc @ncik-roberts

PR #3208, opened 2024-10-30
Fix SIMD build on arm64

mshinwell:
@TheNumbat please approve and merge


chambart:
That C code should properly work on mac. The `#ifdef` should be more general to be active on mac.
But those undefined function should stay there anyway

TheNumbat (approved):

TheNumbat:
We made the build system aware of these `CAMLprim`s in #3178, and `gen_primitives` ignores the `#ifdef` here, so now the arm64 build expects the symbols to exist. This solution seems fine, and should still work on x86 macos.

PR #3209, opened 2024-10-30
Port upstream 13584 (fixing issue 11865)

stedolan:
If a C finaliser tries to delete a root while we're iterating over the set of roots, then mark the root as to be deleted lazily rather than attempting to modify the roots data structure in place.


goldfirere (approved):
Approval based on @stedolan's comment above.

PR #3212, opened 2024-10-31
`Unique_barrier.Not_computed` defaults to `legacy`

riaqn:
Uniqueness analysis does not go into legacy language constructs such as objects; for those, we default to legacy, instead of raising.

### DIFF
--- a/testsuite/tests/typing-unique/unique_mod_class.ml
+++ b/testsuite/tests/typing-unique/unique_mod_class.ml
@@ -151,3 +151,15 @@ Line 7, characters 12-17:
                 ^^^^^
 Error: This value is "aliased" but expected to be "unique".
 |}]
+
+
+let foo (local_ x : string ref) =
+  let module M = struct
+    class c =
+      let y = !x in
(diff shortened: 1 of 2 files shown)


anfelor (approved):
I see, its annoying that this wasn't caught in CI. Does this break the nightly builds? I am happy to merge this right now to fix broken builds.

However, failing in the case of a non-computed unique barrier was one of our soundness conditions to ensure that all unique barriers are set properly. See eg. `comp_pattern_match` in the unique analysis where we also traverse exceptions for this reason. We should follow up with a PR that traverses objects in a similar way to ensure that this remains sound over time.

riaqn:
@anfelor Yeah it might be better to complete the UA instead of removing the invariant.

This is blocking #3065 . I do wish to merge that soonish.

goldfirere:
A middle ground is to accept this PR with a CR uniqueness that says to go back and finish the UA...

anfelor:
I added the CR. Let's merge this now to unblock #3065

PR #3214, opened 2024-10-31
PR#1810 again

xclerc:
As per title; #1810 revealed an issue not directly
related to its own code: by simplifying terminators,
we can produce dead code that will make later
passes fail.

(Merges into `cfg-avoid-deadcode` so that #3213
is included.)


xclerc:
(This diff has already been reviewed, so I
will just merge it into #3213 once that one
is reviewed.)

gretay-js (approved):

PR #3215, opened 2024-10-31
Add `-no-mach-ir` command-line switch

xclerc:
This pull request is arguably a bit gimmicky, but
slightly simplifies testing by providing a single
command-line switch (namely `-no-mach-ir`) to
switch all passes to their CFG implementations
(so that the Mach IR is not used).

At the time of writing, it is equivalent to passing:

- `-cfg-selection`;
- `-cfg-cse-optimize`;
- `-cfg-zero-alloc-checker`;
- `-regalloc cfg`.

### DIFF
--- a/driver/flambda_backend_args.ml
+++ b/driver/flambda_backend_args.ml
@@ -1348,12 +1348,7 @@ module Extra_params = struct
     match name with
     | "internal-assembler" -> set' Flambda_backend_flags.internal_assembler
     | "dgc-timings" -> set' Flambda_backend_flags.gc_timings
-    | "no-mach-ir" ->
-      Flambda_backend_flags.cfg_selection := true;
-      Flambda_backend_flags.cfg_cse_optimize := true;
-      Flambda_backend_flags.cfg_zero_alloc_checker := true;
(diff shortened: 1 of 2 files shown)


driver/flambda_backend_args.ml
```diff
@@ -1335,6 +1348,12 @@ module Extra_params = struct
     match name with
     | "internal-assembler" -> set' Flambda_backend_flags.internal_assembler
     | "dgc-timings" -> set' Flambda_backend_flags.gc_timings
+    | "no-mach-ir" ->
+      Flambda_backend_flags.cfg_selection := true;
+      Flambda_backend_flags.cfg_cse_optimize := true;
+      Flambda_backend_flags.cfg_zero_alloc_checker := true;
+      Flambda_backend_flags.regalloc := "cfg";
```
gretay-js, line 1355: factor out into a function, just in case there are more flags to set in the future.

gretay-js (approved):

driver/flambda_backend_args.ml
gretay-js: The polling pass on CFG has no special flag because it's implied by the register allocator selection.

PR #3217, opened 2024-10-31
Ensure consistent variable binding times

bclement-ocp:
When joining environments, we exclude variables that are no longer useful in the joined environment. This causes the remaining variables to have new (lower) binding times.

If we later join the resulting environment again with another environment that still has those variables, we end up having the same variable with different binding times in the joined level, and we define the variable twice in the resulting environment.

This can allow breaking the order between two variable's binding times, which means that different canonical elements can be chosen in different environments for the same pair of variables.

This patch fixes the issue by keeping track of `next_binding_time` *globally* across all related environments and keeping the binding time during join, ensuring that the binding time of a variable is globally unique and preventing any such re-orderings.

Longer explanation
------------------

Consider 4 environments that define variables `a`, `b`, `c`, and `d`.

```
      _________________________________
     /            |            |       \
  Env 1         Env 2        Env 3    Env 4
   /              |            |         \
define a       define a     define a  define a
   |              |            |         |
define b       define b     define b  define b
   |              |            |         |
define c       define c     define c  define c
   |              |            |         |
define d       define d     define d  define d

```

Suppose that we first join environments 3 and 4 (filtering out `a` and `b`) and environments 1 and 2 (filtering out `d`):

```
      ____________
     /            \
  Env 1 \/ 2      Env 3 \/ 4
   /                \
define a         define c
   |                |
define b         define d
   |
define c
```

Once we join the two resulting environments, if we filter out `a` and `b` but keep `c` and `d`, the current implementation creates a level which defines `c`, then `d`, then `c` again. Since we keep the last definition, we end up with the environment:

```
define d
   |
define c
```

We have now swapped the binding times of `c` and `d` which, if they were in the same alias set, can introduce subtle bugs.

### DIFF
--- a/middle_end/flambda2/types/env/typing_env.ml
+++ b/middle_end/flambda2/types/env/typing_env.ml
@@ -113,7 +113,10 @@ type t =
     (* [prev_levels] is sorted with the greatest scope at the head of the
        list *)
     current_level : One_level.t;
-    next_binding_time : Binding_time.t;
+    next_binding_time : Binding_time.t ref;
+    (* [next_binding_time] is a reference shared by all related environments to
+       the latest binding time ever used. Used to ensure consistent binding
+       times between variables defined in parallel branches. *)
     min_binding_time : Binding_time.t;
         (* Earlier variables have mode In_types *)
     is_bottom : bool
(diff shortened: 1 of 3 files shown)


lthls (approved):
I'm a bit sad that we still expose the `create` function from `Typing_env_level`; there is a single outside use in `Join_levels`, and part of the code there could probably be moved into `Typing_env_level`, but I consider it out of scope for this PR.

PR #3223, opened 2024-10-31
Support Modal module syntax

riaqn:
Add syntax for modal modules.

Please hold off merging, until I finish internal parsetree migration


parsing/parser.mly
```diff
@@ -1935,12 +1939,17 @@ module_type:
...
       { expecting $loc($1) "sig" }
   | FUNCTOR attrs = attributes args = functor_args
-    MINUSGREATER mty = module_type
+    MINUSGREATER mty = module_type mm = optional_at_mode_expr
       %prec below_WITH
       { wrap_mty_attrs ~loc:$sloc attrs (
-          List.fold_left (fun acc (startpos, arg) ->
-            mkmty ~loc:(startpos, $endpos) (Pmty_functor (arg, acc))
-          ) mty args
+          let mty, mm =
```
ncik-roberts, line 1945: I suggest a comment like "return modes go on the innermost functor arrow; all other return modes are empty."

parsing/parser.mly
```diff
@@ -2045,28 +2054,30 @@ signature_item:
...
+    let mty, modalities = body in
+    Md.mk name mty ~attrs ~loc ~docs ~modalities, ext
   }
 ;
 
 (* The body (right-hand side) of a module declaration. *)
-module_declaration_body:
-    COLON mty = module_type
-      { mty }
+module_declaration_body(mm):
```
ncik-roberts, line 2069: I suggest name the argument "modes" rather than "mm" to be a bit more self-documenting.

parsing/parser.mly
```diff
@@ -2045,28 +2054,30 @@ signature_item:
...
   | mkmty(
-      arg_and_pos = functor_arg body = module_declaration_body
+      arg_and_pos = functor_arg body = module_declaration_body(optional_atat_mode_expr)
         { let (_, arg) = arg_and_pos in
-          Pmty_functor(arg, body) }
+          let (ret, mret) = body in
+          Pmty_functor(arg, ret, mret) }
     )
-    { $1 }
+    { $1, [] }
```
ncik-roberts, line 2080: Should there be syntax to give a functor a mode? Something like this, to match functions?:

```ocaml
module (F @ modes) (Arg : Blah) : sig
end
```

parsing/ast_helper.mli
```diff
@@ -303,8 +303,8 @@ module Mod:
     val apply: ?loc:loc -> ?attrs:attrs -> module_expr -> module_expr ->
       module_expr
     val apply_unit: ?loc:loc -> ?attrs:attrs -> module_expr -> module_expr
-    val constraint_: ?loc:loc -> ?attrs:attrs -> module_expr -> module_type ->
-      module_expr
+    val constraint_: ?loc:loc -> ?attrs:attrs -> ?mode:modes -> ?ty:module_type ->
+      module_expr -> module_expr
```
ncik-roberts, line 307: It's a bit weird that the "default" way of calling this function&mdash;by omitting all optional arguments&mdash;lets you construct an invalid AST node. I have two alternative ideas:
   * Match what we do with `Exp.constraint_`: just have two required args.
   * Give `constraint_` the type `?loc:loc -> ?attrs:attrs -> ?mode:modes -> module_expr -> module_type -> module_expr`, and introduce a new function `mode_constraint` with optional module type and required modes.

typing/typemod.ml
```diff
@@ -1018,6 +1019,25 @@ and apply_modalities_module_type env modalities = function
...
+    | head' :: rest -> loc_end_of_modes head' rest
+  in
+  match modes with
+  | [] -> None
+  | head :: rest ->
+    let loc_start = head.loc.loc_start in
+    let loc_end = loc_end_of_modes head rest in
+    Some {loc_start; loc_end; loc_ghost=false}
+
+let check_no_modal_modules ~env modes =
```
ncik-roberts, line 1036: Probably need to check a language extension here.

typing/typemod.ml
```diff
@@ -1018,6 +1019,25 @@ and apply_modalities_module_type env modalities = function
...
+  let rec loc_end_of_modes (head : mode loc) = function
+    | [] -> head.loc.loc_end
+    | head' :: rest -> loc_end_of_modes head' rest
+  in
+  match modes with
+  | [] -> None
+  | head :: rest ->
+    let loc_start = head.loc.loc_start in
+    let loc_end = loc_end_of_modes head rest in
+    Some {loc_start; loc_end; loc_ghost=false}
```
ncik-roberts, line 1034: Does `Misc.last` simplify this at all?

typing/typemod.ml
```diff
@@ -2592,7 +2616,9 @@ and type_module_aux ~alias sttn funct_body anchor env smod =
       Shape.abs funct_shape_param body_shape
   | Pmod_apply _ | Pmod_apply_unit _ ->
       type_application smod.pmod_loc sttn funct_body env smod
-  | Pmod_constraint(sarg, smty) ->
+  | Pmod_constraint(sarg, smty, smode) ->
+      check_no_modal_modules ~env smode;
+      let smty = Option.get smty in
```
ncik-roberts, line 2621: We need a check in `ast_invariants.ml` to make sure that either `smty` is present or `smode` is non-empty. Otherwise it's possible to hit this exception case.

ncik-roberts (commented):
I can't comment on this inline, but can you delete `parsing/parser.cmly`? It seems like a build artifact.

jane/doc/extensions/modes/syntax.md
```diff
@@ -87,6 +87,21 @@ val foo : string -> string @ local @@ portable
 (expression : _ @@ local)
 ```
 
+## Modules
+```ocaml
+module M : S @@ portable = ..
```
goldfirere, line 92: But these syntaxes aren't really supported yet, right? That is, these parse, but they don't work yet, if I understand correctly. I don't think these should be in the user-facing documentation before they work.

parsing/parsetree.mli
```diff
@@ -1165,7 +1166,8 @@ and module_expr_desc =
       (** [functor(X : MT1) -> ME] *)
   | Pmod_apply of module_expr * module_expr  (** [ME1(ME2)] *)
   | Pmod_apply_unit of module_expr (** [ME1()] *)
-  | Pmod_constraint of module_expr * module_type  (** [(ME : MT)] *)
+  | Pmod_constraint of module_expr * module_type option * modes
+      (** [(ME : MT @@ modes)] *)
```
goldfirere, line 1170: and `(ME @ modes)`, presumably

goldfirere (approved):
I've reviewed the docs, source_jane_street.ml, and parsetree.mli

parsing/parsetree.mli
ncik-roberts: Good catch -- yes, we might as well list the 3 possible constructs

jane/doc/extensions/modes/syntax.md
riaqn: Ugh good point. I will remove.

typing/typemod.ml
riaqn: Currently we just outright don't support this by raising a type error; we will add the check when we support this feature.

typing/typemod.ml
riaqn: Hmm I think it would still require two pattern matching, so I'm not sure that's simpler.

PR #3224, opened 2024-11-01
Refactoring in jkind

goldfirere:
This is a bit of an omnibus PR, containing various refactorings that were necessary in order to integrate `with`-kinds.

The initial goal was to decrease the use of `Jkind.Const`, which previously was used in the output of `Jkind.get` -- but no longer. The reason to decrease this use is that, with `with`-kinds, `Jkind.Const` would need to have a number of complicated algorithms (notably, subjkinding) duplicated with normal `Jkind`s.

But it turned out that doing this was no mean feat, and I kept needing to do other preparatory work before the main event. Thus the omnibus.

This most certainly should be reviewed commit-by-commit. Each commit builds and passes the testsuite.

**Definition:** *Zonking* is the act of taking a filled-in variable and replacing it with its contents. Zonking transforms layout `Sort (Var { contents = Some s })` to `Sort s`. This term is taken from GHC, but I know of no other term for this action ("applying the ambient substitution" seems worse), and I propose to use it here. (It's in this commit message, but not in the source code.)

Here is a little guide, with checkboxes to click when a commit is considered fully reviewed (I'm happy to track this):

- [x] 1. 96ea53c3aeb5509308240cd3a604366b3937792c changes `Jkind.default_to_value_and_get` to just get the layout, instead of looking at the entire jkind. After all, `get` is all about zonking sort variables and so is really only about layouts, not full kinds. The patch also renames to `get_layout_defaulting_to_value` to make this clearer. Review suggestion: @ccasin 
- [x] 2. ce7b5f5517a135558454298153dd8654d344ac33 adds support for the `Option` monad. I've reached for this too many times not to just add it. Review suggestion: @dkalinichenko-js 
- [x] 3. 0a393236cae0810432ae15ef27c936adc9c31ccc removes an unused field from `Types.Extension`. (It was hard to accommodate this field in a later commit.) Review suggestion: @ccasin 
- [x] 4. 53da004879c6a4854a0e900a78c955d2981b06a2 stores the kind of a constructor arg in the `constructor_argument` record. It was previously stored in an array that lived alongside the constructor arguments (but was always the same length). The new treatment brings constructor arguments in line with record labels, which already stored the kind like this. This change simplifies a later patch. Review suggestion: @ccasin 
- [x] 5. af3266dd7830d7bbb8de3b02b8990a11bf8d6d4f is just a tiny change to the phrasing of an error message, because sort variables might get set to a sort even though they are not defaulted. This was necessary to avoid an error message regression in a later patch. Review suggestion: @ccasin 
- [x] 6. 2bf408cbed1ced5d5e1044be568652e1bf9bf92e changes the kind of the payload of an `[@@unboxed]` type to be a sort variable; previously it was `any`. Yet there was a check that the actual kind of the payload turns out to be `value` or `void`, so there is no loss in flexibility here. The kind can then be succesfully extracted in `constructor_descrs`. This was necessary to fix an inference problem that arose in a later commit (but seems to be a small improvement anyway). Review suggestion: @ccasin 
- [x] 7. ddc7127000be0271e379bc5fb5b36b16e26d9d45 changes `let rec` to work with `value_or_null`s. I thought this was necessary to avoid a regression in the inference of recursive `[@@unboxed]` types, but it turned out not to be. Yet this allows more programs, and so it seems fine to keep. Review suggestion: @dkalinichenko-js 
- [x] 8. 6aa4fb6acd314fef11a1e1bdebe70f49952536ee removes the printing of the layouts of `[@@unboxed]` types. Previously, we always printed the layout of an `[@@unboxed]` type, because the annotation is necessary in the case of recursive `[@@unboxed]` types. But now that we use a sort variable for the kind of an `[@@unboxed]` type, we can correctly infer that `type t = { x : t } [@@unboxed]` has layout `value` (needed for backward compatibility), and so we actually never need the kind annotation. Review suggestion: @ccasin 
- [x] 9. 83d86ecf76ddbc4b0f0ac418eca7b93da901bc59 removes the `Jkind.annotation` type, which previously stored both a `Jkind.Const.t` and a `Parsetree.jkind_annotation`. It was somewhat like what `Typedtree.jkind` should be like. But to achieve the goals of this PR, I needed to make `Jkind.Const.t` have l- and r-versions, and I didn't want to do that to `annotation`. So I killed `annotation`, storing the information in the top-level `Jkind.t` record instead (where e.g. the history is stored). It turns out that this information is used only in one place, in `Typetexp.TyVarEnv.ttyp_poly_arg`, and so could probably be gotten rid of. But one possibility for making jkind printing better is to store user-written annotations in jkinds, and so I haven't removed this. We might consider doing so after we have a final printing story. Review suggestion: @liam923 
- [x] 10. 5de6f75054b61de1e2f3d542dc2d4ba2b27806be ports the `Layout_and_axes` module from @lpw25's #3105. Review suggestion: @riaqn 
- [x] 11. 31d519ab3bf7e94af198232da30cbf19cf90888d gives the l/r treatment to `Jkind.Const`. This is necessary in order to write the memoization function in Subst without using unsafe features. Review suggestion: @riaqn 
- [x] 12. 69a60fd2b100d0d3b27117e6ee14a297be9528fd uses `Layout_and_axes` to define `Jkind.Const.t`, removing a little duplication. Review suggestion: @riaqn 
- [x] 13. d7761bb2617d08e7ab94d23fe3f3bb0cfc3e5f36 rewrites the memoization function in `Subst`. Review suggestion: @riaqn 
- [x] 14. c57bb226fa32b607bbbb25c3db171addcdd8b549 is (finally!) the main payload of the PR. This changes the output type of `Jkind.get` to be a `(Sort.Flat.t Layout.t, 'd) Layout_and_axes.t`. That is, some axes and a layout that stores a _flat_ sort: this is a new type that represents a sort where the variable is guaranteed not to be filled in (that is, zonked) and no products. This turned out to be harder than I expected, but I'm happy with the result. Review suggestion: @liam923 (or if the going gets tough there, @ccasin)
- [x] 15. 480fb2ee7fe897fbcfd92749a8e655d938848c2c removes some dead code. Review suggestion: @liam923

It feels as if this should be separate PRs, but with the amount of times I had to back into the earlier commits to shape them to better support the later ones, I think that would have been overall worse. I'm not happy with the omnibus nature of this, but I do think this is better than an alternative of many different PRs, several of which would have had to be redone. Our internal version control system would work much better here!


dkalinichenko-js:
I reviewed 2 and 7, LGTM.

Have you tried building the whole tree with this PR?

goldfirere:
> Have you tried building the whole tree with this PR?

Ah, good point. I wanted to do this but forgot. I'm a little worried about perf. I've set up a nightly build.

testsuite/tests/typing-layouts-or-null/test_beta.ml
```diff
@@ -6,7 +6,7 @@
 type t_any_non_null : any_non_null
 
 [%%expect{|
-type t_any_non_null : any_non_null
+type t_any_non_null : any
```
liam923, line 9: I don't know enough about or_null - is this desired?

liam923 (commented):
Reviewed 9 - 83d86ec

testsuite/tests/typing-layouts-or-null/test_beta.ml
dkalinichenko-js: `any` is parsed as `any_non_null` in alpha extensions aren't on, so I think this just changes the displayed name to `any` while still leaving it non-null. I should remove this whole hack, it's bad.

dkalinichenko-js: This change is fine for now.

typing/jkind.ml
```diff
@@ -975,31 +1018,42 @@ module Jkind_desc = struct
...
-      modes_upper_bounds = mode_ub;
-      externality_upper_bound = ext_ub;
-      nullability_upper_bound = null_ub
-    }
+    let annotations = Misc.Stdlib.Monad.Option.all annotations in
+    let annotation =
+      Option.map
+        (fun annotations ->
+          Parsetree.
+            { pjkind_loc = Location.none; pjkind_desc = Product annotations })
```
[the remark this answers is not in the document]
goldfirere: Great catch!

typing/jkind.ml
```diff
@@ -1886,15 +1786,26 @@ let score_reason = function
...
+    in
+    let choose_history k_a history_a k_b history_b =
+      match Jkind_desc.sub k_a k_b with
+      | Less -> history_a
+      | Not_le ->
+        (* CR layouts: this will be wrong if we ever have a non-trivial meet in
+           the kind lattice -- which is now! So this is actually wrong. *)
+        history_b
+      | Equal -> choose_better_history history_a history_b
+    in
```
liam923, line 1771: I don't love these names - maybe `higher_scored_history` for the first one?

liam923 (commented):
Revied 14 - c57bb22

liam923 (commented):
Reviewed 15 - 480fb2e

goldfirere:
Testing with nightly was successful. The refactoring here did cause at least one thing to break, but it's the usual breakage we see when core types change, and easily fixed. Performance tests are encouraging, leading to a possible 9.5% decrease in artifact sizes, though the test wasn't as head-to-head as I would like. In any case, I think we don't need to worry about that kind of impact.

goldfirere:
Thanks for the review @liam923. Here are my responses for your quick review:

- [ ] https://github.com/ocaml-flambda/flambda-backend/pull/3224/commits/66684b1dbeac4c32ced54cbc359e07178c990c77
- [ ] https://github.com/ocaml-flambda/flambda-backend/pull/3224/commits/4bfda6741b016c7fdc2d98ba41b5b653a37c1ea6
- [ ] https://github.com/ocaml-flambda/flambda-backend/pull/3224/commits/db754f884d03ef6808273218c97accefa1cbda1d

typing/jkind.mli
```diff
@@ -179,31 +179,41 @@ end
 
 module Const : sig
   (** Constant jkinds are used for user-written annotations *)
-  type t = Types.type_expr Jkind_types.Const.t
+  type 'd t = (Types.type_expr, 'd) Jkind_types.Const.t
```
riaqn: I'm not sure why we want to index constants over allowance. I thought constants are alwasy LR. Let's discuss offline.

goldfirere: We indeed discussed, and I believe @riaqn understands the motivation better. In short: `Jkind.Const.t` is *not* constant. It simply doesn't have any sort variables, but it still might have `with 'a` and such. @riaqn resolve if you're satisfied here.

typing/typedecl.ml
```diff
@@ -915,13 +903,14 @@ let transl_declaration env sdecl (id, uid) =
             env None true lbls (Record { unboxed = unbox })
           in
           let rep, jkind =
+            if unbox then
+              Record_unboxed,
+              Jkind.of_new_legacy_sort ~why:Old_style_unboxed_type
+            else
```
ccasin, line 909: I agree that in practice we were limited to value here, so we're not losing anything.  But it does seem a bit sad to restrict this to be representable - it goes against our general goal to have layouts be restricted by uses rather than here in typedecl.

I don't think we should hold up the PR for it, but can you say more about what inference problem leaving this as `any` causes?

ccasin (commented):
The commits I was asked to review look good, module a few small questions and suggestions below.

typing/datarepr.ml
```diff
@@ -107,10 +107,17 @@ let constructor_args ~current_unit priv cd_args cd_res path rep =
...
+    | Variant_extensible, _ -> assert false
+    | Variant_boxed x, _ -> x
+    | Variant_unboxed, [{ cd_args }] ->
+      begin match cd_args with
+      | Cstr_tuple [{ ca_jkind = jkind }]
+      | Cstr_record [{ ld_jkind = jkind }] ->
+        [| Constructor_uniform_value, [| jkind |] |]
+      | _ -> Misc.fatal_error "Multiple arguments in [@@unboxed] variant"
+      end
+    | _ -> Misc.fatal_error "Multiple constructors in [@@unboxed] variant"
```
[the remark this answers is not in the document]
goldfirere: This change is the major change in this patch, actually, driving the motivation behind it all. I don't remember all the details, but `decl.type_jkind` comes, I think, from the original estimate of the kind. We'll update the final `type_jkind` to be correct, but this one won't be right, yet, and so we have to use the kind from the constructor arg.

That said, of course you're right about the catch-all patterns. Will fix.

typing/typedecl.ml
goldfirere: No, sorry, I can't. I don't seem to remember. But I do remember not investing time into understanding this better, thinking that we would fix it in the sweep to delay checks for `any` to construction sites.

typing/datarepr.ml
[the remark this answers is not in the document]
ccasin: > I don't remember all the details, but decl.type_jkind comes, I think, from the original estimate of the kind. We'll update the final type_jkind to be correct, but this one won't be right, yet, and so we have to use the kind from the constructor arg.

I took a quick look through typedecl, and this doesn't yet makes sense to me. This function is reached by many paths, but I think the two relevant ones are probably the two calls to `add_types_to_env` in `Typedecl.transl_type_decl`. It's not obvious to me that the `ca_jkind`/`ld_jkind` should be more accurate than `decl.type_jkind` at either. At the first call site, we haven't yet reached `update_decl_jkind` yet so I think `ca_jkind`/`ld_jkind` will still be the "any" from `transl_labels` or `transl_types_gf`. At the second callsite, we're fully done checking the decl so I think the jkind of the field and the jkind of the overall type should be the same.

I'm sure I'm missing something basic here, but since it has eluded me it seems worth understanding and documenting with a comment on this function.

> This change is the major change in this patch, actually, driving the motivation behind it all.

If this change is the motivation, is there a test that demonstrates it?  Might be nice to point to from the comment.

typing/jkind.mli
riaqn: Yes it's constant in terms of sort variables (which is not described by allowance anyway). It's not constant in terms of mode crossing. We agree that further refactoring will clean up this bit, so it's fine.

Could you add a comment here to clarify that?

typing/jkind_types.ml
```diff
@@ -578,10 +578,11 @@ type ('type_expr, 'd) t =
   }
 
 module Const = struct
-  type 'type_expr t =
+  type ('type_expr, +'d) t =
     { layout : Layout.Const.t;
       modes_upper_bounds : Modes.t;
```
riaqn: I think it's worthwhile to pass down the `'d` parameter to `Modes.t`, to indicate that this is the part that allowance describes. The current `Modes.t` will probably just ignore this `'d`, but that's still more clearer.

(This code might have been moved around and renamed around "Layout_and_axes", but my point stays.

riaqn (approved):

typing/jkind_types.ml
goldfirere: I don't think I agree here. It's true that the allowance describes the modal component, but not in a way that really makes sense at this point in time: it's describing whether or not there are `with`-types, but that's not implemented yet. It does *not* describe left-modes vs right-modes, which is what attaching `'d` to `Modes.t` would look like it means (to me).

(The `'d` also equally applies to the non-mode axes.)

I don't see an easy way to express the valid point you're trying to make here -- that the `'d` is not about the layout! -- but it all becomes much better in the `with`-kinds PR, already posted. So I think it's just easiest to let this one lie, as it's very temporary.

ccasin (approved):

typing/jkind_types.ml
riaqn: > It does not describe left-modes vs right-modes, which is what attaching 'd to Modes.t would look like it means (to me).

Indeed, that should be renamed to `Mode_crossing.t`. Then, it makes sense to talk about left and right, in the following sense: we can compare two mode crossings, and `m0 <= m1` iff `m0` crosses more modes than `m1`.

There are, by my experience with left and right modes, two motivations to distinguish left and right:
- Internally, modes might be join or meet, which inherently forbids them to be on the LHS or RHS of submode, as that would be unsolvable.
- Externally, modes are used as actual or expected, in which case it's healthy to reflect that in the types, by mapping them to left and right modes.

Currently, without the `with` syntax, we don't have the internal motivation. But the external motivation is still valid.

On the other hand, I also understand this might be some work, with benenfits that are not obvious right now. So I'm happy to resolve this conversation if you can just add a CR there.

typing/datarepr.ml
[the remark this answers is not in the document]
goldfirere: OK. I've added the CR.

typing/jkind_types.ml
goldfirere: I think you're right to question this, but I don't wish to add the CR. This code changes a bunch in subsequent PRs and the issue you highlight goes away completely once we have `with`-kinds, which is already posted. So I think adding this CR will just cause a bit of work in dealing with conflicts and rebasing but without much benefit.

PR #3226, opened 2024-11-05
Fix bug in dataflow with [Invalid] extra_args

Ekdohibs:
This fixes a fatal error on main with this input:
```ocaml
let go input =
  let x, y =
    match input with
    | `A a -> 0, Some a
    | `B b -> b
  in
  x, y
;;

let make i =
  let list = [ i; 123 ] in
  go (`A list)
;;
```

What happened was that there was a continuation with two calling sites, one being the one taken at runtime, the other being invalid. However, this invalid was only seen due to (normal) unboxing, and thus ended in the extra args.
We thus had a continuation `k1` being called from two different continuations, `k2` and `k3`, with the calling site from `k3` being invalid.

Since we compute dataflow after normalisation and removing the invalid call site, we saw that a variable defined inside `k2` was the dominator of an argument to `k1`, and passed as an alias. However, `k3` was not in scope of this variable, but we tried to add it as we thought `k3` was calling `k1`, due to the `k1` being in the `apply_cont_args` map of `k3`, even though the list of call sites was empty! Since neither `k3` nor any continuation calling it in scope of the variable, the extra argument was added all the way up until the dummy toplevel continuation, crashing the compiler.

### DIFF
--- a/middle_end/flambda2/simplify/flow/flow_acc.ml
+++ b/middle_end/flambda2/simplify/flow/flow_acc.ml
@@ -445,6 +445,12 @@ let extend_args_with_extra_args (t : T.Acc.t) =
                   rewrite_ids)
             elt.apply_cont_args
         in
+        let apply_cont_args =
+          Continuation.Map.filter
+            (fun _ rewrite_ids ->
+              not (Apply_cont_rewrite_id.Map.is_empty rewrite_ids))
+            apply_cont_args
+        in
         { elt with apply_cont_args })
       t.map
   in
--- a/middle_end/flambda2/simplify/flow/flow_acc.ml
+++ b/middle_end/flambda2/simplify/flow/flow_acc.ml
@@ -445,7 +445,7 @@ let extend_args_with_extra_args (t : T.Acc.t) =
                     (add_extra_args_to_call ~extra_args)
                     rewrite_ids
               in
-              (* We must leave an empty [Apply_cont_rewrite_id] map here.
+              (* We must not leave an empty [Apply_cont_rewrite_id] map here.
                  Indeed, not having any actual continuation arguments will let
                  flow analysis find a dominator for parameters of [cont] that we
                  do not have access to. Since we then only look at which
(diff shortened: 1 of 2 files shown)


middle_end/flambda2/simplify/flow/flow_acc.ml
```diff
@@ -445,6 +445,12 @@ let extend_args_with_extra_args (t : T.Acc.t) =
                   rewrite_ids)
             elt.apply_cont_args
         in
+        let apply_cont_args =
```
Gbury, line 448: Could you add a comment here, resuming what you explain in the description of the PR ?

Gbury (approved):
Thanks for the quick debugging and fix !

middle_end/flambda2/simplify/flow/flow_acc.ml
Gbury: Btw, we should try and add some comments in flow<_analysis, at least to explain what each piece of code is supposed to do: re-reading the code is slightly painful at the moment, even though I wrote most of it, so I'll see if I have time to add some doc.

Gbury: Also, if we want to, we could try and fuse the `filter` added here to the `mapi` above, with some `Map.filter_mapi` (though I doubt that function exists) ?

middle_end/flambda2/simplify/flow/flow_acc.ml
```diff
@@ -434,15 +434,29 @@ let extend_args_with_extra_args (t : T.Acc.t) =
...
+              let rewrite_ids =
+                match Continuation.Map.find cont t.extra with
+                | exception Not_found -> rewrite_ids
+                | epa ->
+                  let extra_args = EPA.extra_args epa in
+                  Apply_cont_rewrite_id.Map.filter_map
+                    (add_extra_args_to_call ~extra_args)
+                    rewrite_ids
+              in
+              (* We must leave an empty [Apply_cont_rewrite_id] map here.
```
Gbury, line 448:
```suggestion
              (* We must not leave an empty [Apply_cont_rewrite_id] map here.
```

Gbury:
Oh, sorry to say this only now, but is there a way to add a test to ensure that this bug does not happen again ? Either the original example, or the one @lthls made.

Gbury (approved):
Seems good to me.

PR #3227, opened 2024-11-05
Ensure predecessor sets are correct after teminator simplification

xclerc:
This pull request fixes a bug in terminator
simplification. At the end of the pass, the
`register_predecessors_for_all_blocks` was
called, but this function can only add
missing predecessors, not remove ones.

This means that if predecessors would need
to be deleted, we would end up with an
invalid CFG. Interestingly enough, last
week's fix ensuring the deletion of dead
code was partially masking this issue.
In practice, when a predecessor needs to
be removed, it is often the case that the
portion of the CFG that would be invalid
is in fact dead code.

### DIFF
--- a/backend/cfg/simplify_terminator.ml
+++ b/backend/cfg/simplify_terminator.ml
@@ -199,4 +199,11 @@ let run cfg =
         let shortcircuit = block cfg b in
         registration_needed || shortcircuit)
   in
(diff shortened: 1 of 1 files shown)


gretay-js (approved):

xclerc:
(I have checked that the other uses of the
function were starting from empty sets.)

PR #3229, opened 2024-11-05
Unboxed records

rtjoa:
Add support for unboxed records (construction, functional updates, projection, pattern matching), whose types have product jkinds. For example:

```ocaml
type 'a with_idx : value & immediate = #{ data : 'a ; i : int }
let idx #{ data = _ ; i } = i
let inc r = #{ r with i = r.#i + 1 }
```

The fields of unboxed records have a separate namespace from record fields. While fields can be mutable, we don't yet support updating them.

## Implementation decisions

**Unboxed record types are distinct from record types, rather than being yet another `record_representation`.**
- Having separate codepaths is likely to be less error-prone than adding another record_representation, which we might not always match on in the right places.
- The majority of the duplication this causes is ameliorated by making `label_description` abstract over the type for record representations, which is now either `record_representation` or `record_unboxed_product_representation`. This allowed nearly all shared logic to be parameterized by the record "form," with the most notable exception being in [lambda/matching.ml](https://github.com/ocaml-flambda/flambda-backend/blob/rtjoa.basic_unboxed_records/lambda/matching.ml).

**A (temporary) jkind-checking algorithm that interleaves looking through unboxed types and type expressions constructors.**
- See the comment on `type_jkind_deep` in [typing/ctype.ml](https://github.com/ocaml-flambda/flambda-backend/blob/rtjoa.basic_unboxed_records/typing/ctype.ml).
- We could have instead updated `constrain_type_jkind` to return a jkind estimate, but this would require significant changes. Instead, we use this somewhat duplicative solution, as it will be removable once we support kind-polymorphic type declarations.

## Future work
- Declaring a boxed record `t` also declares its unboxed version `t#`, when doing so would be well-founded.
- The signature item `type t boxes value & value` declares the kinds of both `t` and `t#` and indicates that they are linked.
- Unboxed record fields can be mutated (only when the unboxed record is nested within a boxed record).
- Infinite-size unboxed records give better type errors (see [testsuite/tests/typing-layouts-unboxed-records/recursive.ml](https://github.com/ocaml-flambda/flambda-backend/blob/rtjoa.basic_unboxed_records/testsuite/tests/typing-layouts-unboxed-records/recursive.ml)).

Review: @ccasin


ccasin (commented):
I've only read the tests so far, and will continue on to the rest, but wanted to post what I have so you can get back to this.

testsuite/tests/typing-layouts-unboxed-records/unboxed_records.ml
```diff
@@ -0,0 +1,104 @@
+(* TEST
```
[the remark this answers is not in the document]
rtjoa: Updated to duplicate tests from the tuples version, and better testing of function update, projection, and partial patterns!

testsuite/tests/typing-layouts/basics_alpha.ml
```diff
@@ -636,15 +636,16 @@ module M9_4 = struct
...
-       but is mixed here with fields of type "('a : value)"
+           ^^^^^^^^^^^^^^
+Error: This pattern matches values of type "void_unboxed_record"
+       but a pattern was expected which matches values of type
+         "('a : value_or_null)"
        The layout of void_unboxed_record is void
          because of the definition of void_unboxed_record at line 12, characters 0-60.
        But the layout of void_unboxed_record must be a sublayout of value
-         because it's a boxed record type.
+         because it's the type of a tuple element.
```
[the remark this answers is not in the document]
rtjoa: I'm not sure...  I'll spend some time looking in case this is a symptom of an actual problem.

testsuite/tests/typing-layouts-unboxed-records/unique.ml
```diff
@@ -0,0 +1,92 @@
...
+ flags = "-extension layouts_alpha -extension unique";
+ {
+   expect;
+ }
+*)
+
+(* Uniqueness tests *)
+
+type t = #{ x : string ; y : string }
+let use_string (_s @ unique) = ()
```
anfelor, line 13: This function could be `unique_use` since it accepts non-strings as well. That also avoids confusion later when we may or may not make strings mode-cross uniqueness

testsuite/tests/typing-layouts-unboxed-records/unique.ml
```diff
@@ -0,0 +1,92 @@
...
+Line 4, characters 13-14:
+4 |   let t = #{ t with y = "fresh" } in
+                 ^
+Error: This value is used here, but it has already been used as unique:
+Line 3, characters 13-17:
+3 |   use_string t.#x;
+                 ^^^^
+
+|}]
+
```
anfelor, line 92: Can you add the following tests:
```
let () =
  let t = mk () in
  unique_use t;
  let t = #{ t with y = "fresh" } in
  unique_use t

let () =
  let t = mk () in
  unique_use t;
  let t = #{ t with x = "fresh"; y = "fresh" } in
  unique_use t.#x

type t = #{ x : unit; y : string }
val mk : unit -> t @ unique = <fun>

let () =
  let t = mk () in
  unique_use t;
  let t = #{ t with y = "fresh" } in
  unique_use t

let () =
  let t = mk () in
  unique_use t;
  let t = #{ t with x = (); y = "fresh" } in
  unique_use t.#x
```

I would guess that the output will be fail, success, success, success (this assumes that `unit` mode-crosses uniqueness which it might only do with same extra type annotations).

In contrast, normal records will fail all four tests. This tests the fact that we implicitly borrow the memory address of records for the functional update while we do not do that for unboxed records.

testsuite/tests/typing-layouts-unboxed-records/unique.ml
```diff
@@ -0,0 +1,92 @@
+(* TEST
```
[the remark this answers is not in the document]
anfelor: Both this file and `typing/uniqueness_analysis.ml` look good to me

testsuite/tests/typing-layouts-unboxed-records/unique.ml
rtjoa: Thanks! Updated.

testsuite/tests/typing-layouts-unboxed-records/unique.ml
rtjoa: Thanks very much for the tests - I added the first, second, and fourth.

I couldn't find a way to annotate the third test such that it passes. It seems like we want to annotate the unit component "produced" by the functional update as unique, but there's no piece of syntax in the functional update expression that refers. Is there a trick I'm missing?

ccasin (commented):
Have now read everything once - will double back to take a look at the remaining test questions.

testsuite/tests/typing-layouts-unboxed-records/unique.ml
anfelor: Thanks, I agree now that the third test will fail; this is due to how mode crossing is implemented in the uniqueness analysis and while we should fix it at some point in the future, this is not the place to do so. Can you still add the test with the error message and add the following CR?
```
(* CR uniqueness: this test should succeed since unboxed records have no memory address
    and the first field is of type unit and thus mode-crosses uniqueness. 
    We should fix this by allowing multiple unique uses for values that mode-cross uniqueness. *)
```

testsuite/tests/typing-layouts-unboxed-records/basics.ml
```diff
@@ -0,0 +1,493 @@
...
+
+type ('a : float64) t = #{ x : string; y : 'a }
+[%%expect{|
+type ('a : float64) t = #{ x : string; y : 'a; }
+|}];;
+
+type ('a : float64, 'b : immediate) t = #{ x : string; y : 'a; z : 'b }
+[%%expect{|
+type ('a : float64, 'b : immediate) t = #{ x : string; y : 'a; z : 'b; }
+|}];;
```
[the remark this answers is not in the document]
ccasin: (I think this probably doesn't get better with with kinds - @rtjoa is going to a closer look so we at least understand what is going wrong)

typing/typedecl_separability.ml
```diff
@@ -633,6 +632,12 @@ let check_def
       check_type env constructor.argument_type Sep
       |> msig_of_context ~decl_loc:def.type_loc
            ~parameters:constructor.result_type_parameter_instances
+  | Unboxed_product ->
+      (* CR layouts 7.3: we give products best_msig, as they are
+         not affected by the flat float array optimization.
+         See test [typing-layouts-unboxed-records/separability.ml].
+      *)
```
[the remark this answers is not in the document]
rtjoa: Per our meeting, changed this to be Algebraic, as it has the same behavior and is appropriate. (I think the fact that this was a CR was a mistake, and was just meant to be a normal comment)

typing/typedecl_separability.ml
```diff
@@ -46,30 +46,29 @@ type type_structure =
   | Open
   | Algebraic
   | Unboxed of argument_to_unbox
+  | Unboxed_product
```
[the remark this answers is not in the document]
rtjoa: Per our meeting, changed this to be Algebraic, as it has the same behavior and is appropriate.

typing/typedecl.ml
```diff
@@ -1000,7 +1020,9 @@ let transl_declaration env sdecl (id, uid) =
       let uid = decl.typ_type.type_uid in
       match decl.typ_type.type_kind with
       | Type_variant (cstrs, _) -> Shape.str ~uid (shape_map_cstrs cstrs)
-      | Type_record (labels, _) -> Shape.str ~uid (shape_map_labels labels)
+      | Type_record (labels, _)
+      | Type_record_unboxed_product (labels, _) ->
+        Shape.str ~uid (shape_map_labels labels)
```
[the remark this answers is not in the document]
rtjoa: (Updated with the suggested change.)

typing/shape.mli
```diff
@@ -79,6 +79,7 @@ module Sig_component_kind : sig
     | Type
     | Constructor
     | Label
+    | Unboxed_label
```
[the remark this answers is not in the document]
liam923: Merlin (and ocaml-index, but that's in the same codebase as Merlin) is the only tool on my radar that uses shapes. I don't anticipate any Merlin issues with this - I think the way we normally handle merging compiler changes into Merlin will handle this fine.

typing/typedecl.ml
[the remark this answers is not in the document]
liam923: That sounds right to me. I think there's two relevant locations to add tests for this:
- `testsuite/tests/shape-index/index_labels.ml`
- `testsuite/tests/shapes`. For whatever reason, it seems that non-unboxed labels are only indirectly tested at the moment, so I'm not sure where in there the right home would be. Probably just a new file?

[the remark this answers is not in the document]
rtjoa: Thanks @liam923. Added tests in fa009c1c059e2846ba12e218a0103b93bd7b31b7.

typing/ctype.ml
```diff
@@ -2087,11 +2097,18 @@ let unbox_once env ty =
...
       | Some ty2 ->
         let ty2 = match get_desc ty2 with Tpoly (t, _) -> t | _ -> ty2 in
-        Stepped (apply env decl.type_params ty2 args)
+        Stepped (apply ty2)
+      | None -> begin match decl.type_kind with
+        | Type_record_unboxed_product ([{ld_type = arg; _}], Record_unboxed_product) ->
+          Stepped (apply arg)
+        | Type_record_unboxed_product ((_::_::_ as lbls), Record_unboxed_product) ->
+          Stepped_record_unboxed_product (List.map (fun ld -> apply ld.ld_type) lbls)
+        | _ -> Final_result
```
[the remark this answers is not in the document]
ccasin: Sorry, I meant list the cases for the `_` on the previous line (so all the `type_kind`s).  It looks like instead you've listed all the type_descs a few lines down (which is probably not necessary, since this is really just looking for a couple specific ones.

testsuite/tests/typing-layouts-unboxed-records/basics.ml
```diff
@@ -31,6 +31,17 @@ type t = #{ i : int; j : int; }
...
 
+let bad_match (x : t) =
+  match x with
+  | _ -> .
+[%%expect{|
+Line 3, characters 4-5:
+3 |   | _ -> .
+        ^
+Error: This match case could not be refuted.
+       Here is an example of a value that would reach it: "{ _ }"
```
ccasin, line 42: This says `"{ _ }"` but I guess I expect it to say `#{ _ }`.  If it's hard to fix, just leave a CR, not worth spending a bunch of time investigating.

lambda/translcore.ml
```diff
@@ -696,18 +696,17 @@ and transl_exp0 ~in_new_scope ~scopes sort e =
           Lprim (Pmixedfield (lbl.lbl_pos, read, shape, sem), [targ],
                   of_location ~scopes e.exp_loc)
       end
-  | Texp_unboxed_field(arg, _id, lbl, _) ->
+  | Texp_unboxed_field(arg, arg_sort, _id, lbl, _) ->
     begin match lbl.lbl_repres with
     | Record_unboxed_product ->
-      let layouts, jkinds =
+      let layouts, _jkinds =
```
ccasin, line 702: I don't think there's any reason for this to compute `_jkinds` any more.

typing/typecore.ml
```diff
@@ -5941,12 +5941,19 @@ and type_expect_
       if Types.is_mutable label.lbl_mut then
         fatal_error
           "Typecore.type_expect_: unboxed record labels are never mutable";
+      let record_sort =
+        let jkinds =
+          Array.map (fun lbl -> lbl.lbl_jkind) label.lbl_all |> Array.to_list
+        in
+        let record_jkind = Jkind.Builtin.product ~why:Unboxed_record jkinds in
+        Jkind.sort_of_jkind record_jkind
+      in
```
ccasin, line 5961: I think we should get the sort from the type of the actual record value here, rather than computing a new sort from the kinds in the declaration.  In the future we'll allow record declarations like:

```ocaml
type ('a : any) t = #{ x : int; y : 'a }
```

Here, we don't actually know the sort of the field `y`, but we can know the sort of the fields of specific instantiations of this type.

There are a couple ways you could get the sort from the record.  You can call `type_sort` on the type in `record` here (using `~fixed:false` - this is exactly the kind of place we'll need to constrain type variables to be representable when we allow such record declarations).  Or you could change `type_label_access` to return the sort, possibly by changing it to use `type_expect` on the record rather than `type_exp`, and making the expected type a new type variable whose kind is a sort variable (using such a type as the expected type is essentially just checking the result is representable).

typing/typecore.ml
```diff
@@ -5453,9 +5453,20 @@ and type_expect_
...
+            | None -> None
+            | Some (exp, _) ->
+              let jkind =
+                Jkind.Builtin.product ~why:Unboxed_record
+                  (Array.map (fun lbl -> lbl.lbl_jkind) label_descriptions
+                   |> Array.to_list)
+              in
+              let sort = Jkind.sort_of_jkind jkind in
+              Some (exp, sort)
+          in
```
ccasin, line 5466: As in the previous commit, I think we should get the sort from `exp` rather than building a new one from the declaration here.  Probably the code that computes the typed version of `opt_exp` nearer the top of this function should be doing that - that's the right place to check the value we're overriding is representable (which we'll need to do anyway when we go from checking records are representable at declaration time to checking they are representable at construction time).

testsuite/tests/typing-layouts/basics_alpha.ml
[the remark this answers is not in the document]
rtjoa: Aha - the culprit is that we *no longer* assume that all `Maybe_a_record_type` record patterns are value. [Minimal working example on `main`.](https://github.com/ocaml-flambda/flambda-backend/commit/c903249d2297c9352e14123a8a690213d36e6288)

This assumption caused the "boxed records must be value" error to fire before the "tuple elements must be value" error had a chance to.

Our change should be useful anyway for when we allow non-value `[@@unboxed]` records. I don't think it can be harmful, because the sort variable is unified with the record labels' `lbl_res` anyway in `solve_Ppat_record_field`.

typing/typecore.ml
rtjoa: Thanks for the explanation, updated in 7ed17a022d34ae4d60f1a7205aba0f7c329b950.

typing/typecore.ml
rtjoa: Updated in 7ed17a022d34ae4d60f1a7205aba0f7c329b950!

ccasin: Sorry, I think this is still not quite right.  You are computing the sort from `ty_expected`, which I believe is the type of the result record.  But in a world where we allow any in declarations, these may not be the same. E.g.:
```ocaml
type 'a t = { x : 'a; y : string }
let update_sort (r : int64# t) : float# t = { r with x = Int64_u.of_float r.x }
```
You can use `exp.exp_type`, or whatever other copies of the type of the original record are lying around here.

typing/typecore.ml
```diff
@@ -10645,6 +10651,16 @@ let report_error ~loc env = function
         "@[Function arguments and returns must be representable.@]@ %a"
         (Jkind.Violation.report_with_offender
            ~offender:(fun ppf -> Printtyp.type_expr ppf ty)) violation
+  | Record_projection_not_rep (ty,violation) ->
+      Location.errorf ~loc
+        "@[Record projections must be representable.@]@ %a"
```
ccasin, line 10692: I think it's not exactly that the projection is unrepresentable, but rather that the record being projected from is unrepresentable.

typing/typecore.ml
rtjoa: Oops, thanks. Updated in c0382c6d35e4ec4593069c4b2da8a1ad15634dcf

ccasin (approved):

PR #3230, opened 2024-11-05
Reads of tags of unique allocations have to be Reads_agree

anfelor:
As part of #3066 (and its cherry-pick #3198), we added a mechanism to the compiler to detect reads from unique allocations. Since unique allocations can be overwritten, those reads may not be pushed down. We achieved this by mapping these reads to `Reads_vary` in lambda. To be safe, we also made reads of tags `Reads_vary`. While reads of tags can not be pushed down anyway (since they are used immediately to select a branch), we felt that we could make the code safer by enforcing this invariant by mapping them to `Reads_vary` as well. However, this lead to a compilation failure in flambda2 and might lead to a performance degradation since flambda2 does not propagate tag information to remove branches that can never be taken if the pattern match uses a `Reads_vary` to read the tag. Therefore, this PR rolls back the change to tag reads. In particular, after this PR there should be no change to generated code unless the code uses uniqueness.

To reproduce this bug in the compiler before this PR, compile this file with `ocamlopt -O3` (thanks to @ncik-roberts):
```
let go input =
  let x, y =
    match input with
    | `A a -> 0, Some a
    | `B b -> b
  in
  x, y
;;

let make i =
  let list = [ i; 123 ] in
  go (`A list)
;;
```

### DIFF
--- a/lambda/matching.ml
+++ b/lambda/matching.ml
@@ -3183,11 +3183,12 @@ let combine_constructor value_kind loc arg pat_env cstr partial ctx def
                       (Lprim (Pintcomp Ceq, [ Lvar tag; ext ], loc), act, rem, value_kind))
                   nonconsts default
               in
-              (* Since the tag is used directly in the tests, it would be sound
-                 to mark it Alias+Reads_agree but this would not change performance.
-                 We keep it this way to guarantee soundness of in-place overwriting. *)
+              (* CR uniqueness: This read might be on a unique allocation that could
+                 be overwritten. We should ensure that it is never pushed down.
+                 However, we can not use [Reads_vary] here, since this breaks
+                 flambda2. *)
               Llet (StrictOpt, Lambda.layout_block, tag,
-                    Lprim (Pfield (0, Pointer, Reads_vary), [ arg ], loc),
+                    Lprim (Pfield (0, Pointer, Reads_agree), [ arg ], loc),
                     tests)
         in
         List.fold_right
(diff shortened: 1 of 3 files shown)


lthls:
Also: I don't think it makes sense to support overwriting for polymorphic variants. The runtime memory representation is a bit unintuitive, with all polymorphic variants compiled to either raw integers (for constant variants) or blocks of size 2 (for non-constant ones). The payload is never inlined, so overwriting is of little interest compared to just unboxing.

lambda/matching.ml
```diff
@@ -3183,11 +3183,12 @@ let combine_constructor value_kind loc arg pat_env cstr partial ctx def
                       (Lprim (Pintcomp Ceq, [ Lvar tag; ext ], loc), act, rem, value_kind))
                   nonconsts default
               in
-              (* Since the tag is used directly in the tests, it would be sound
-                 to mark it Alias+Reads_agree but this would not change performance.
-                 We keep it this way to guarantee soundness of in-place overwriting. *)
+              (* CR uniqueness: This read might be on a unique allocation that could
+                 be overwritten. We should ensure that it is never pushed down.
+                 However, we can not use [Reads_vary] here, since this breaks
+                 flambda2. *)
```
ncik-roberts, line 3189: We should change this comment to discuss performance rather than the flambda2 issue

ncik-roberts (approved):

anfelor:
I have made a few more changes:

 - I now use a unique barrier to switch two of the tag reads between `Reads_agree` and `Reads_vary` depending on whether the allocation is unique. That means those reads are always going to be sound.
 - This means that the only `Reads_agree` that remains unprotected by a unique barrier is now in the `for_optional_arg_default` function. Since I don't think it is possible for users to refer to the allocation holding the optional argument, users should not be able to overwrite it or change it otherwise. That should make it sound for us to have a `Reads_agree` there.
 - After talking to @ncik-roberts it seems that we might lose some performance when using uniqueness due to the `Reads_vary`s. I have added two comments to `lambda.ml` detailing the reason for this and we will see in the future if we can improve that situation.

@lthls: I agree, we currently do not support overwriting polymorphic variants or extension constructors. However, a similar issue happens even without overwriting when users allocate off the OCaml heap. Since OCaml's pattern matching works on external allocations and users might at some point implement a `free` function for such values, it is better to be conservative and add unique barriers in that case as well.

ncik-roberts (approved):

PR #3231, opened 2024-11-05
Disable inlining for effect functions that manipulate stacks

TheNumbat:
All of the effect `continue` functions extract a stack pointer from the given continuation and hide it in an immediate.
If we enter the GC between extracing the stack and calling `resume`, the stack will not be marked, leading to memory corruption.
This situation can only occur due to re-ordering allocations after inlining, so we can work around this issue by disabling inlining.

Additionally, `Deep.try/match_with` and `Shallow.fiber` must not enter the GC between `alloc_stack` and `runstack` for the same reason.

A better fix would be to change the fiber layout to not rely on storing the stack/last_fiber pointers as immediates, or at least use a single primitive to implement atomic `continue` and `alloc_and_runstack` operations.

### DIFF
--- a/stdlib/effect.ml
+++ b/stdlib/effect.ml
@@ -76,7 +76,7 @@ module Deep = struct
   external reperform :
     'a t -> ('a, 'b) continuation -> last_fiber -> 'b = "%reperform"
 
-  let match_with comp arg handler =
+  let[@inline never] match_with comp arg handler =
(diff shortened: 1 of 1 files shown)


stedolan (approved):

mshinwell:
I think this probably should have had comments to explain the reason for those attributes in the code.

PR #3233, opened 2024-11-06
Improve hugepage support

stedolan:
Rounds up memory mapping sizes to a multiple of hugepage size, and ensures they are aligned on hugepage boundaries. Helps particularly for the minor heap on Linux before 6.7, where hugepages are important but the kernel does not align mmap addresses itself.

Currently x86_64 only.

This patch greatly improves the L1 TLB hit rate for accesses to the minor heap. I used the following benchmark:
```
let rec list_init len f =
  match len with
  | 0 -> []
  | len -> f () :: list_init (len - 1) f

let () =
  let xs =
    Sys.opaque_identity (list_init 100 (fun _ ->
      let _ = list_init 200 ignore in
      1))
  in
  for i = 0 to 10_000_000 do
    assert (List.length xs = 100)
  done
```
which allocates a list of length 100, with its cons cells spaced out, and iterates over it many times. (I reimplemented List.init because the current version in the stdlib is cleverer than this, doing TRMC and unrolling to go faster. This makes its memory layout harder to understand).

On x86_64 Linux, this benchmark usually runs 'slow' (about 3 seconds), but runs 'fast' (about 1 second) if the minor heap happens to be aligned enough to allow the kernel to allocate a hugepage, which happens sometimes by coincidence. This patch makes it reliably run 'fast'.

This patch is not necessary on very recent (2024 vintage) Linux kernels, which (finally!) default to allocating large mappings on hugepage boundaries without the manual alignment in this patch. However, this patch does not hurt on such kernels, and helps on all older ones.

### DIFF
--- a/runtime/caml/platform.h
+++ b/runtime/caml/platform.h
@@ -429,7 +429,7 @@ Caml_inline unsigned caml_plat_spin_step(unsigned spins,
 
 /* Memory management primitives (mmap) */
 
-uintnat caml_mem_round_up_pages(uintnat size);
+uintnat caml_mem_round_up_mapping_size(uintnat size);
 /* The size given to caml_mem_map and caml_mem_commit must be a multiple of
    caml_plat_pagesize. The size given to caml_mem_unmap and caml_mem_decommit
    must match the size given to caml_mem_map/caml_mem_commit for mem.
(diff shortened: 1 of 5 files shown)


mshinwell (approved):
I made one minor fix, improved a comment and made one piece of code slightly clearer.  All looks correct to me.

mshinwell (approved):
Looks good.  Let's do a follow-up PR to format simple_operation.ml{,i}.

mshinwell:
(sorry comment meant for another PR ended up on here)

PR #3234, opened 2024-11-07
Split the `Mach` module

xclerc:
This pull request splits the `Mach` module so
that elements shared with the CFG pipeline
are in a new module (thus `leaving` in `Mach`
only the elements which are not shared,
which eventually will make the deletion of
`Mach` slightly easier).


mshinwell:
ah and there is something to fix for `make runtest`, please just fix that and merge.

PR #3236, opened 2024-11-07
Extra protection against codegen pessimisation due to unique barriers

anfelor:
This PR changes the `transl_unique_barrier` function to always return `May_be_pushed_down` when uniqueness is not enabled. This makes it easy to reason about the absence of codegen changes for the next compiler roll. To review this, please look through #3066 and #3230 again and convince yourself that if all occurences of `ubr` are `May_be_pushed_down`, then we will emit exactly the same code as before.

### DIFF
--- a/lambda/translmode.ml
+++ b/lambda/translmode.ml
@@ -46,4 +46,10 @@ let transl_modify_mode locality =
 let transl_unique_barrier barrier =
   match Typedtree.Unique_barrier.resolve barrier with
   | Uniqueness.Const.Aliased -> May_be_pushed_down
-  | Uniqueness.Const.Unique -> Must_stay_here
+  | Uniqueness.Const.Unique ->
+    (* CR uniqueness: this is a temporary measure to ensure that we can roll
+       the compiler without breaking existing code. Once uniqueness is stable,
+       we can simplify this to always return [Must_stay_here]. *)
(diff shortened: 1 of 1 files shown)


lambda/translmode.ml
```diff
@@ -46,4 +46,10 @@ let transl_modify_mode locality =
 let transl_unique_barrier barrier =
   match Typedtree.Unique_barrier.resolve barrier with
   | Uniqueness.Const.Aliased -> May_be_pushed_down
-  | Uniqueness.Const.Unique -> Must_stay_here
+  | Uniqueness.Const.Unique ->
```
goldfirere, line 49: This is what we agreed when talking. But now looking at the code, I wonder: is it ever possible to have `Unique` with the extension off? That would make the statement about the uniqueness extension even stronger.

anfelor: Yes, we could probably return `assert false` here

anfelor:
As discussed in our meeting, this should be good to merge?

goldfirere:
I think we've since decided that it's best to roll back the changes to code-gen entirely, until we can be sure to avoid miscompilation.

anfelor:
Since we decided not to roll back the changes to code-gen, we should merge this. When we add the new flag, this PR ensures that the unique barriers are not enabled unless the user asked for it.

PR #3237, opened 2024-11-07
Add intrinsics for load/store of unboxed int8/int16

poechsel:
Add missing intrinsics to load store unboxed int8 and int16.

As they are the same as the one for normal int8 / int16 this is mainly for clarity, and this PR could be ignored


gretay-js:
These intrinsics can be exposed with better names as part of `ocaml_intrinsics` library.

PR #3240, opened 2024-11-07
Clarify reasoning about where unique barriers are necessary

anfelor:
This PR is to track the ongoing discussion on slack about which primitives other than field reads might need a unique barrier.

Notice that as of now the current implementation is probably unsafe around arrays since we have a remaining `Parraylength` call that may be pushed down past a free of the underlying array.


lambda/lambda.mli
```diff
@@ -603,12 +603,40 @@ type let_kind = Strict | Alias | StrictOpt
...
+    - Pattern matches ~> [Pfield]+co, [Parrayrefu], [Punboxed_product_field],
+        [Parraylength], [Lstringswitch]
+    where [Pfield]+co also includes [Pfloatfield], [Pufloatfield], [Pmixedfield].
+
+    If the unique_barrier is set in the typedtree, the generated lambda code
+    may not be pushed down. We achieve this as follows:
+    - [Pfield]+co: we use [add_barrier_to_read]
+    - [Punboxed_product_field]: an unboxed product is not an allocation,
+        so no protection is necessary
+    - [Pduprecord]: this one can not be pushed down
```
goldfirere, line 621: Do you mean that currently flambda2 does not ever push this down? Just not sure what the comment means here.

lambda/lambda.mli
```diff
@@ -603,12 +603,40 @@ type let_kind = Strict | Alias | StrictOpt
...
+    - [Parrayrefu]: we use [add_barrier_to_mutable_flag]
+    - [Parraylength]: TODO this one is currently unsafe
+    - [Lstringswitch]: we do not mind since strings mode-cross uniqueness
```
goldfirere, line 624: I find this unconvincing. Maybe it's a mistake to have string mode-cross uniqueness! What's the safety property? That strings are never updated and never freed?

lambda/lambda.mli
```diff
@@ -603,12 +603,40 @@ type let_kind = Strict | Alias | StrictOpt
...
+
+    There are many other primitives in lambda that can read from an allocation,
+    such as: [Parrayrefs], [Pduparray], [Pbyteslength], [Pbytesrefu],
+    [Pbytesrefs], [Pbigarrayref], [Pbigarraydim], [Pbytes_load*],
+    [P.*_array_load_128], [Patomic_load], [Patomic_exchange], [Patomic_cas],
+    [Patomic_fetch_add], [Punbox_vector], [Pbox_vector].
+    However, none of these can be generated by a one of the typedtree nodes. *)
```
goldfirere, line 631: > one of the typedtree nodes

You mean the ones above? I think there's a slightly more direct argument: These primitives are produced only by syntax that counts as a unique use. Is that right?

lambda/lambda.mli
```diff
@@ -603,12 +603,40 @@ type let_kind = Strict | Alias | StrictOpt
       we can discard e if x does not appear in e'
  *)
 
+(** A unique barrier is necessary on all references to a value that
```
goldfirere, line 606: This comment will grow stale quickly, because new primitives and syntax are being added all the time. At the least, it should include a date of validity and some instructions for the analysis you did so that it can be re-created in the future.

But how can it be kept up to date? Maybe leave some breadcrumbs in the right places in uniqueness_analysis.ml so that this is updated whenever a new bit of syntax doesn't count as a unique use?

goldfirere (commented):
Review only of comment in lambda.mli -- waiting for consensus with @mshinwell before proceeding.

anfelor:
Closed, since the comment is better addressed by #3251 and the `matching.ml` change is done better in #3243

PR #3242, opened 2024-11-08
Apply flambda2 switch table optimization to naked numbers and symbols

mshinwell:
This ensures that the optimization that turns an flambda2 `Switch` into a lookup table, in certain circumstances, applies to the naked number kinds as well, e.g.
```
type t = A | B | C

let f1 = function
  | A -> 8
  | B -> 4
  | C -> 2

let f2 = function
  | A -> #8l
  | B -> #4l
  | C -> #2l
```


mshinwell:
This needs a fix for floats, working on it.

mshinwell:
This also now deals with symbols, so it will work for things like boxed floats too.

middle_end/flambda2/identifiers/reg_width_const.mli
```diff
@@ -28,6 +28,29 @@ val is_naked_immediate : t -> Target_ocaml_int.t option
...
+
+val is_naked_vec128 : t -> Vector_types.Vec128.Bit_pattern.t option
+
+val is_naked_vec256 : t -> Vector_types.Vec256.Bit_pattern.t option
+
+val is_naked_vec512 : t -> Vector_types.Vec512.Bit_pattern.t option
+
+(** Create a numeric constant of the given kind ([Region] and [Rec_info] are
+    forbidden). *)
+
```
lthls, line 53: This comment seems to have been duplicated

middle_end/flambda2/simplify/simplify_switch_expr.ml
```diff
@@ -246,46 +269,110 @@ let recognize_switch_with_single_arg_to_same_destination0 machine_width ~arms =
...
-       the original [Switch]). *)
-    let[@inline] check_args prover must_untag_lookup_table_result =
-      let args' = List.filter_map prover args in
+    let[@inline] check_args prover must_untag_lookup_table_result wrapper =
+      let args' =
+        List.filter_map
+          (fun arg -> match arg with Const cst -> Some cst | Symbol _ -> None)
+          args
+      in
+      let args' = List.filter_map prover args' in
```
lthls, line 302: Can't you move the cal to `prover` to the previous call to `filter_map` ? E.g. `Const cst -> prover cst`

middle_end/flambda2/simplify/simplify_switch_expr.ml
```diff
@@ -328,37 +415,185 @@ let return ~added_code_size ~free_names expr uacc ~dacc_before_switch:_ =
...
   let block_sym =
     let var = Variable.create "switch_block" K.value in
     Symbol.create
       (Compilation_unit.get_current_exn ())
       (Linkage_name.of_string (Variable.unique_name var))
   in
-  let uacc =
-    let fields =
+  let uacc, array_kind, array_load_kind, loaded_kind =
+    let simples_from_consts make_simple numbers =
```
lthls, line 428: The name of this function (and `constify` below) are not great. I think that `fields_to_simples` and `fields_to_or_variables` would be better, although I don't think it's perfect.

bclement-ocp (commented):
I think the optimization is valuable, but I have two concerns with the code.

1. As far as I can tell, we are converting a list of `RWC.t` of the same kind into a `lookup_table_fields`, then back from the `lookup_table_fields` into a list of `RWC.t`, making things more convoluted than necessary.
    What do you think about incorporating all that logic in `recognize_switch_with_single_arg_to_same_destination0` and return something like this for `lookup_table_fields`:

    ```ocaml
    type lookup_table_fields =
      | Tagged_immediates of TI.t list
          (** All arms are tagged immediates (or, with [Must_untag], all arms are
              naked immediates, which are stored as tagged ones and untagged after
              loading). A value array specialised to [Immediates] can be used. *)
      | Static_arguments_of_single_kind of K.t * AK.t * ALK.t * Simple.t list
          (** Simples must all be symbols or constants of the same kind. *)
    ```

    I think it would simplify the code significantly.

2. I think the code is currently creating arrays of boxed floats with an array kind and array load kind marked as "Values", which is not supposed to happen with flat float arrays enabled. It might be OK because these are symbols and the runtime is (I believe) not involved, but I'd like your opinion — in any case it's probably worth a comment.

lthls (approved):
Some minor things to improve, otherwise looks correct.

mshinwell:
@bclement-ocp could you do another pass on this please?

middle_end/flambda2/simplify/simplify_switch_expr.ml
```diff
@@ -246,46 +260,101 @@ let recognize_switch_with_single_arg_to_same_destination0 machine_width ~arms =
...
+              let expected_discr =
+                TI.add (TI.one machine_width) expected_discr
+              in
+              Some (Some dest', arg :: args_rev, expected_discr))
+            ~const:(fun _const ->
               let expected_discr =
                 TI.add (TI.one machine_width) expected_discr
               in
-              Some (Some dest', const :: args_rev, expected_discr))))
+              Some (Some dest', arg :: args_rev, expected_discr))))
```
bclement-ocp: Nit (feel free to ignore): we could use `Simple.must_be_var` here now to avoid the duplication.

middle_end/flambda2/simplify/simplify_switch_expr.ml
```diff
@@ -246,46 +260,101 @@ let recognize_switch_with_single_arg_to_same_destination0 machine_width ~arms =
...
+    let module RWC = Reg_width_const in
+    let module ALK = P.Array_load_kind in
+    (* Symbols are always of kind [value]; they may be freely mixed with
+       [Const]s of kind [value] (i.e. tagged immediates). For all other kinds
+       symbols are not permitted and every arm must be a constant of the same
+       [Flambda_kind.t]. *)
+    let kind_of simple =
+      Simple.pattern_match' simple
+        ~var:(fun _ ~coercion:_ ->
+          (* Variables have already been ruled out above. *) assert false)
```
bclement-ocp: `Misc.fatal_error`? Especially since support for variables is a CR.

middle_end/flambda2/simplify/simplify_switch_expr.ml
```diff
@@ -246,46 +260,101 @@ let recognize_switch_with_single_arg_to_same_destination0 machine_width ~arms =
...
+        (* If no arm is a symbol and all [Const]s are tagged immediates, return
+           a [Tagged_immediates] variant so that the affine optimization can
+           apply below. Returns [None] otherwise. *)
+        List.fold_right
+          (fun simple acc ->
+            match acc with
+            | None -> None
+            | Some tagged_imms ->
+              Simple.pattern_match' simple
+                ~var:(fun _ ~coercion:_ -> assert false)
```
bclement-ocp: `None` or `Misc.fatal_error`?

bclement-ocp (approved):
Thanks for the refactoring, I think the code is easier to follow now.

I think it would be good to also have a `[%%expect_fexpr]` version of the test once #5837 is merged to make sure the optimisation keeps happening where we expect it to (i..e in the middle-end).

mshinwell:
@Keryan-dev I'm going to merge this but if you could incorporate some fexpr tests for it in #5837 or a follow-up PR it would be appreciated.

PR #3243, opened 2024-11-08
Add unique barriers for Pduprecord, Parraylength, Parrayrefu

anfelor:
This PR adds unique barriers for `Pduprecord`, `Parraylength` and `Parrayrefu`. It pushes them into flambda2 and adds coeffects if the unique barrier is enabled.

 - When reviewing this, keep in mind that our correctness condition is that a unique barrier is never forgotten. However, if a new instance of the primitive is generated, it will be fine to add a dummy barrier `May_be_pushed_down`.
 - There are three cases in `middle_end/flambda2/simplify/` where I did not know what to do. I have left CRs to be discussed in review.
 - I have added the unique barrier to `fexpr`, added the ability to print it and make a corresponding parser change. However, both `make regen-flambda2-tests` and `make runtest` currently fail for me (see slack).


anfelor:
Note that the CI currently only fails on `tests3.flt` as discussed on slack. I would appreciate a pointer about how to fix the failing tests and which test updates to include in this PR.

lpw25:
I think it is not clear to me why these constructs need unique barriers. I would not expect the uniqueness analysis to know anything about `Array.length`. Operations the uniqueness analysis does not know about should not need barriers.

lpw25:
Ok, I found the answer to my (implicit) question in #3240: array length is used to compile pattern matching on arrays. However, I think this PR is not the right initial solution to that problem: rather than changing compilation to support non-unique pattern matching on arrays, it is almost certainly much easier to make the uniqueness analysis stricter so that it considers such pattern matches as full uses of the array's address.

goldfirere:
> make the uniqueness analysis stricter so that it considers such pattern matches as full uses of the array's address.

I would worry that this would decrease the usefulness of uniqueness in DRF programming, because now we can't really have an array that holds things that hold keys.

lpw25:
You can still do that, and you can even pattern match on them, you just need to explicitly borrow the value before you match on it.

PR #3245, opened 2024-11-08
Detect when a kind makes being a function type impossible

goldfirere:
As reported by @tdelvecchio-jsc: `let f (x : ('a : bits64)) = x ()` produced an uncaught `Unify`.

First commit contains the tests; second fixes it.

Review: @dkalinichenko-js

### DIFF
--- a/testsuite/tests/typing-layouts/basics.ml
+++ b/testsuite/tests/typing-layouts/basics.ml
@@ -2857,3 +2857,20 @@ Error: This expression has type "float#" but an expression was expected of type
        But the layout of float# must be a sublayout of value
          because it's the type of the recursive variable x.
 |}]
+
+(**********************************************************)
+(* Test 46: errors when functions don't have layout value *)
+
+let f (x : ('a : bits64)) = x ()
+
+[%%expect{|
+Uncaught exception: Ctype.Unify(_)
+
+|}]
+
+let f (x : ('a : value mod portable)) = x ()
+
+[%%expect{|
+Uncaught exception: Ctype.Unify(_)
+
+|}]
--- a/testsuite/tests/typing-layouts/basics.ml
+++ b/testsuite/tests/typing-layouts/basics.ml
@@ -2864,13 +2864,21 @@ Error: This expression has type "float#" but an expression was expected of type
 let f (x : ('a : bits64)) = x ()
 
 [%%expect{|
-Uncaught exception: Ctype.Unify(_)
-
+Line 1, characters 28-29:
+1 | let f (x : ('a : bits64)) = x ()
+                                ^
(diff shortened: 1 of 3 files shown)


testsuite/tests/typing-layouts/basics.ml
```diff
@@ -2857,3 +2857,28 @@ Error: This expression has type "float#" but an expression was expected of type
...
+
+let f (x : ('a : bits64)) = x ()
+
+[%%expect{|
+Line 1, characters 28-29:
+1 | let f (x : ('a : bits64)) = x ()
+                                ^
+Error: This expression is used as a function, but its type "'a"
+       has kind "bits64", which cannot be the kind of a function.
+       (Functions always have kind "value mod unique uncontended".)
```
dkalinichenko-js, line 2872: Should we print the full `value mod unique uncontended` here?

dkalinichenko-js (commented):
I'm surprised functions mode-cross uniqueness. Why?

testsuite/tests/typing-layouts/basics.ml
dkalinichenko-js: Probably, there's nothing wrong with that, but maybe we can do some nice pretty-printing. I think it's not worth doing, but maybe you disagree.

goldfirere: I don't have a shorter way to print that. The error should be possible only with a kind annotation, so the programmer at least knows about kinds. I think it's OK.

dkalinichenko-js (approved):

tdelvecchio-jsc:
Note: this doesn't require the user to know about kinds, only that they use a library which involves kinds. In particular:
```
module Atomic : sig
  type ('a : value mod portable) t
  val get : 'a t -> 'a @ contended
end

let f x =
  let f = Atomic.get x in
  f ()
;;
```

(where the user obviously didn't implement `Atomic`)

goldfirere:
> I'm surprised functions mode-cross uniqueness. Why?

Why not? More mode-crossing is more good.

And good point @tdelvecchio-jsc. I still think it's OK to include the kind stuff in the output.

dkalinichenko-js:
> Why not? More mode-crossing is more good.

Why is it sound for them to mode-cross uniqueness?

goldfirere:
> Why is it sound for them to mode-cross uniqueness?

Because all of these are true:
1. There is no way to update a function in place, reusing existing memory to point to a new code pointer, for example.
2. Functions are atomic: they have no subpieces (which might support update-in-place).
3. We don't expect either of those qualtities to change (ever).

dkalinichenko-js:
I see. And if 3 changes, we'll make a new function type.

PR #3247, opened 2024-11-08
Fix soundness bug of uniqueness analysis on immutable array patterns

anfelor:
This PR fixes a soundness bug in the uniqueness analysis wrt immutable array patterns. The issue is that we used to make the elements of arrays their own roots in the `Ienv`. That used to be fine because mutable arrays can only hold many, aliased data. However, immutable arrays can hold unique elements and the old approach was unsound for them. I fixed this by correctly registering the elements of arrays as projections. To keep the old behaviour for mutable arrays, I added the `mutable_implied_modalities` to the projection.

### DIFF
--- a/typing/uniqueness_analysis.ml
+++ b/typing/uniqueness_analysis.ml
@@ -459,6 +459,7 @@ module Projection : sig
     | Record_field of string
     | Construct_field of string * int
     | Variant_field of label
+    | Array_index of int
     | Memory_address (* this is rendered as clubsuit in the ICFP'24 paper *)
 
   module Map : Map.S with type key = t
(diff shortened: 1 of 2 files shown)


typing/uniqueness_analysis.ml
```diff
@@ -820,6 +832,8 @@ end = struct
 
   let variant_field s t = child (Projection.Variant_field s) t
 
+  let array_index gf i t = modal_child gf (Projection.Array_index i) t
```
riaqn, line 835: Maybe better to let this function take `mut` instead of `gf`. Because currently the modalities is decided by solely `mut`, so the `gf` would have only two possible values here.

PR #3249, opened 2024-11-11
Flambda 2 CSE fixes

mshinwell:
WIP.  Currently:
1. Reduce the scope of `type t` in `common_subexpression.ml`, so it can be `private` most of the time.

Includes #3248


mshinwell:
This also adds a signature for `Eligible_for_cse` to avoid mistakes within `flambda_primitive.ml`.

bclement-ocp (approved):

PR #3250, opened 2024-11-11
Parallelise building `dynlink_compilerlibs`

lukemaurer:
This performs two weird tricks that together knock ~30 seconds off the build time in dev mode and ~40 seconds outside dev mode:

- To build `dynlink_compilerlibs`, copy both `parser.ml` and `parser.mli` from the main compiler source rather than just `parser.ml`. Copying just `parser.ml` means that the native build depends on the .cmi produced by the bytecode build.

- Explicitly add the .cma and .cmxa files for `dynlink` and `flambda2_from_lambda` to the `install` alias. It's not clear why this helps, but it appears that Dune is being excessively lazy in discovering installable targets, waiting until `ocamlcommon` is built before it even considers any targets in these libraries, including those that don't depend on `ocamlcommon` (and take quite a while to build). Redundantly adding targets to `install` seems to be the nudge Dune needs.

### DIFF
--- a/dune
+++ b/dune
@@ -866,3 +866,20 @@
  (package ocaml)
  (files
   (include compiler-libs-installation.sexp)))
+
+; This is entirely redundant, but without it, each target in these libraries
+; and their dependencies waits until ocamlcommon is built, including those
+; targets that don't actually depend on it. This produces severe bottlenecking,
+; costing about 15 seconds when building in dev mode. It's not immediately clear
+; why Dune fails to parallelise here. Going by the progress indicator, my best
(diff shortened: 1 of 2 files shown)


stedolan:
I am really looking forward to the cleansing fire of ocaml/ocaml#11996 that will arrive with 5.3, but in the meantime this looks great. (Might also be worth asking a Dune person to have a look at why adding `flambda2_from_lambda` helps? Seems like that might be a Dune bug).

stedolan (approved):

PR #3252, opened 2024-11-11
Documentation: Disclaimer for soundness of uniqueness

anfelor:


jane/doc/extensions/uniqueness/intro.md
```diff
@@ -5,8 +5,20 @@ In this document, we use the new [syntax for modes](../modes/syntax.md).
 
 The `unique` mode designates values that have only a single reference pointing
 to them. If an operation takes a `unique` argument, it will consume the only
-reference to the value. For example, an externally allocated data structure
-might support a `free` operation:
+reference to the value. This is particularly useful for the password of the
+mutual exclusion primitive in our design for data-race freedom.
+
+__Currently, uniqueness should only be used in the limited interface of
```
goldfirere, line 11: I'm a little worried about this wording. If you bought a car, and there was a button on the display -- next to all the other good buttons -- that might randomly blow up some other car on the same highway, you probably wouldn't buy the car.

But actually I'm worried about this whole approach (this is a new worry -- I had previously said I was in favor of this approach). What if we have a unique key, and try to store that in a structure? That will force the structure (an allocation!) to be unique. And do we have any reason to believe that doing so won't cause miscompilation?

So I continue to think that we should roll back all the changes to code-gen that interact with uniqueness. Then we can be confident of avoiding miscompilation. And we can then re-word this to say exactly what we mean: that the optimizer can reorder operations past unique usages of a value, and so any memory-affecting operation (freeing, overwriting) whose soundness depends on uniqueness is going to be unsound. But because we will offer no such operations, the feature as a whole is safe to use.

anfelor: Yes, the wording is quite strong. I agree that we can weaken it a bit if we revert the unique barrier patches on `main`. However, most of the documentation is currently written around a `free` operation which is currently unsafe (with or without unique barriers). That makes it hard to claim that the feature is safe to use.

goldfirere: Then we should probably restructure the documentation to downplay the `free` use-case.

I also worry that we've created false urgency here. Why do we need uniqueness to be stable in today's roll? I'm not sure what that buys us. If we can work out everything (rolling back the code-gen changes, fixing the docs), then I'm fine with making it stable, but I don't see a need to uniqueness to be stable today.

anfelor:
Following a discussion with @nmatschke, I have updated the documentation to discuss the current soundness issue and how it can be worked around. I have added this comment to the end, since #3271 now includes an error if a unique barrier should be installed but isn't. That makes uniqueness for DRF safe to use and the issue discussed here is merely an inconvenience where the uniqueness analysis may reject valid programs.

nmatschke (approved):

anfelor:
@nmatschke In #3271 we have now decided not to throw an error if a barrier would be required. I have changed the language to reflect this. I also removed the mention which constants mode-cross uniqueness since that depends on #3272 which is not yet merged.

nmatschke (approved):

PR #3253, opened 2024-11-11
Only run `float32#` tests on AMD64

dkalinichenko-js:
Only run `float32#` tests (and one problematic SIMD test) on AMD64, since ARM64 does not support `float32#` yet. Leave a CR to enable them on all platforms when supported.


dkalinichenko-js:
@TheNumbat `tests/mixed-blocks/generated_native_test.ml` is failing with this error in the CI, do you know what could be the reason?

```
> /usr/bin/ld: generated_native_test.o: in function `camlGenerated_native_test__immstring390':
> :(.data+0x6058): undefined reference to `caml_int64x2_low_to_int64'
> /usr/bin/ld: :(.data+0x6060): undefined reference to `caml_int64x2_low_of_int64'
> /usr/bin/ld: :(.data+0x6068): undefined reference to `caml_sse2_vec128_interleave_high_64'
> /usr/bin/ld: :(.data+0x6070): undefined reference to `caml_sse2_vec128_interleave_low_64'
> collect2: error: ld returned 1 exit status
> File "caml_startup", line 1:
> Error: Error during linking (exit code 1)
```

TheNumbat:
These symbols are provided by `stubs.c`. The test passes if I move `arch_amd64` after `modules`, so seems to be something weird about `ocamltest`.

testsuite/tests/flambda/incompatible_arrays.ml
```diff
@@ -1,5 +1,6 @@
 (* TEST
  flambda2;
+ arch_amd64;
```
TheNumbat, line 3: Tests for supported layouts should ideally still run - can you add CRs to re-enable the files that aren't specific to float32/vec128?

lthls:
The semantics of ocamltest actions is weird, indeed. Actions like `flambda2` or `arch_amd64` act as guards for the following actions only, meaning that if they're put after the "main" actions (usually `native`), they never prevent the main action from running in any circumstance (they will transform a passing result into a skipped result if the condition does not hold). That is why moving the condition earlier restores the expected behaviour.
To put it simply, `;` in ocamltest actions should usually be interpreted as short-cutting `&&` (with two possible `false` values, one for failure and one for skipping).

TheNumbat:
I see, so `arch_amd64` should go after `modules` but before `native`.

testsuite/tests/flambda/incompatible_arrays.ml
dkalinichenko-js: The particular test you commented on only makes sense with `float32#`, but I've added CRs to others.

testsuite/tests/lib-array/test_iarray_typeopt.ml
```diff
@@ -5,6 +5,10 @@
    expect;
 *)
 
+(* CR dkalinichenko: this test generates slightly different code
```
TheNumbat, line 8: Is the difference just that the `name/number` identifiers change? If it's more than that it seems concerning.

dkalinichenko-js: Yeah, that's right.

TheNumbat (approved):

PR #3254, opened 2024-11-11
Generate `META` files for otherlibs

dkalinichenko-js:
Generate `META` files for extension otherlibs that had them missing. Tested by running `make install` and inspecting the generated folder.

### DIFF
--- a/dune
+++ b/dune
@@ -791,10 +791,6 @@
 (install
  (files
   (compilerlibs/META as compiler-libs/META)
-  (otherlibs/stdlib_upstream_compatible/META as stdlib_upstream_compatible/META)
-  (otherlibs/stdlib_stable/META as stdlib_stable/META)
-  (otherlibs/stdlib_beta/META as stdlib_beta/META)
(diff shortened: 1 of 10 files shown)


TheNumbat (approved):
Looks fine. Are the meta files used in the opam installation?

dune
```diff
@@ -791,6 +791,10 @@
 (install
  (files
   (compilerlibs/META as compiler-libs/META)
+  (otherlibs/stdlib_upstream_compatible/META as stdlib_upstream_compatible/META)
```
TheNumbat, line 794: Should these go in the dune files for each lib?

dkalinichenko-js: Afaik, if a dune file has an `install` stanza, it won't automatically generate the list of files to install from `.ml`s. And I'd rather avoid unnecessarily listing all of those.

dkalinichenko-js:
Yeah, they're for opam and (external) dune.

dune
TheNumbat: `stdlib_alpha/dune` appears to already have an `install` block, though?

PR #3257, opened 2024-11-12
Support for Null constant and Is_null primitive in the middle-end

lthls:
This PR is the final piece for support of null-related features in the middle-end.
It includes support for the `Null` constant and the `Is_null` primitive in Lambda and Flambda2.

### DIFF
--- a/middle_end/flambda2/from_lambda/closure_conversion.ml
+++ b/middle_end/flambda2/from_lambda/closure_conversion.ml
@@ -175,7 +175,7 @@ let rec declare_const acc dbg (const : Lambda.structured_constant) =
             ~symbol:(fun _sym ~coercion:_ -> ())
             ~const:(fun cst ->
               match RWC.descr cst with
-              | Tagged_immediate _ -> ()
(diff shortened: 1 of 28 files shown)


middle_end/flambda2/from_lambda/closure_conversion.ml
```diff
@@ -250,7 +251,8 @@ let close_const0 acc dbg (const : Lambda.structured_constant) =
         | Naked_int32 _ -> KS.naked_int32
         | Naked_int64 _ -> KS.naked_int64
         | Naked_nativeint _ -> KS.naked_nativeint
-        | Naked_vec128 _ -> KS.naked_vec128)
+        | Naked_vec128 _ -> KS.naked_vec128
+        | Null -> KS.any_value (* CR layouts v3: give a more precise kind? *))
```
lthls, line 255: Note: this is correct, if not very precise, and it turns out these kinds aren't used anyway so I will delete this code in a follow-up PR.

Ekdohibs (approved):

PR #3258, opened 2024-11-12
Remove unused kind computations in From_lambda

lthls:
While working on #3257 I noticed that we compute kinds in `Closure_conversion` that are never used.
This PR removes these unused kinds.


lthls:
I think that `close_cont0` builds a `Simple.t`, while `close_cont` builds a `Named.t`, so they're not redundant.

lthls:
Ah, but `close_cont0` is just an alias for `declare_cont` now. I will do the renaming.

mshinwell (approved):

PR #3259, opened 2024-11-12
Make `Label.t` abstract

xclerc:
(as per title)


flambda-backend/tests/backend/regalloc_validator/check_regalloc_validation.ml
```diff
@@ -209,17 +209,20 @@ let entry_label =
...
-let move_tmp_res_label = entry_label + 3
+let call_label = label_add entry_label 2
 
-let add_label = entry_label + 4
+let move_tmp_res_label = label_add entry_label 3
 
-let return_label = entry_label + 5
+let add_label = label_add entry_label 4
 
-let new_label i = entry_label + 6 + i
```
gretay-js, line 222: This test just needs unique labels, they don't need to be consecutive, as far as I can tell, so `Label.new_label ()` is enough, and then `Label.of_int_for_testing` can be removed. It can be done as a follow-up PR.

gretay-js (approved):

PR #3261, opened 2024-11-12
Dedicated module to print registers

xclerc:
Follow-up to #3234.

To be able to eventually delete `Printmach`, we
have to move the functions related to registers
to another module. This cannot be `Reg` itself,
unfortunately, because that would cause circular
dependencies. This pull request thus creates a
dedicated module, namely `Printreg`.


gretay-js (approved):
- Testing: at least zero_alloc tests use `-dcse` and would have tested the new functions.
- `scripts/compare.sh` mentions `printmach.cmx` and needs to be updated to mention `printreg.cmx` and probably a few other files added recently. It can be done in a separate PR next time we use the script.

PR #3262, opened 2024-11-12
Fix case where parser drops attributes in packed module types.

tdelvecchio-jsc:
Parsing currently drops `[@annot1]` and `[@annot3]` in the code below:

```
module _ : sig
  val foo : (module T [@annot1] with type t = 'a [@annot2]) -> unit
end = struct
  let foo (type a) (module M : T [@annot3] with type t = a [@annot4]) = ()
end
```

There's no good place in the AST to attach these attributes to, so we should just raise a syntax error in this case.

### DIFF
--- a/parsing/parse.ml
+++ b/parsing/parse.ml
@@ -160,6 +160,8 @@ let prepare_error err =
             Format.fprintf ppf
               "only module type identifier and %a constraints are supported"
               Style.inline_code "with type"
+        | Misplaced_attribute ->
+            Format.fprintf ppf "an attribute cannot go here"
       in
       Location.errorf ~loc "invalid package type: %a" invalid ipt
   | Removed_string_set loc ->
--- a/parsing/parser.mly
+++ b/parsing/parser.mly
@@ -809,7 +809,12 @@ let package_type_of_module_type pmty =
   in
   match pmty with
   | {pmty_desc = Pmty_ident lid} -> (lid, [], pmty.pmty_attributes)
-  | {pmty_desc = Pmty_with({pmty_desc = Pmty_ident lid}, cstrs)} ->
+  | {pmty_desc = Pmty_with({pmty_desc = Pmty_ident lid; pmty_attributes = inner_attributes}, cstrs)} ->
+      begin match inner_attributes with
+      | [] -> ()
+      | attr :: _ ->
+        err attr.attr_loc Syntaxerr.Misplaced_attribute
+      end;
       (lid, List.map map_cstr cstrs, pmty.pmty_attributes)
   | _ ->
       err pmty.pmty_loc Neither_identifier_nor_with_type
--- a/parsing/syntaxerr.ml
+++ b/parsing/syntaxerr.ml
@@ -21,6 +21,7 @@ type invalid_package_type =
   | Private_types
   | Not_with_type
   | Neither_identifier_nor_with_type
+  | Misplaced_attribute
 
 type error =
     Unclosed of Location.t * string * Location.t * string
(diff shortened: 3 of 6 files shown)


tdelvecchio-jsc:
We should consider how to avoid issues like this in the future. One possibility is to register every attribute in a hashtbl as we parse it, and then check that the final parsed AST still contains all those attributes?

testsuite/tests/parsing/dropped_attributes.ml
```diff
@@ -0,0 +1,15 @@
+(* TEST
+   ocamlc_byte_exit_status = "2";
+   setup-ocamlc.byte-build-env;
+   ocamlc.byte;
+   check-ocamlc.byte-output;
+*)
+
+(* Attributes [[@annot1]] and [[@annot3]] have not node in the parsetree
```
goldfirere, line 8: I don't think this test covers `annot3`. Maybe we need another test to do so? But I think probably an expect-test will work nicely here.

testsuite/tests/parsing/dropped_attributes.ml
```diff
@@ -0,0 +1,15 @@
...
+   to attach to; make sure the parser doesn't drop them (which it used to do). *)
+
+module _ : sig
+  val foo : (module T [@annot1] with type t = 'a [@annot2]) -> unit
```
goldfirere, line 12: Are we testing anything with `annot2` and `annot4`?

goldfirere (commented):
This fix should probably be contributed upstream. I think the approach here is OK, but another possibility is to combine the lists of attributes -- my guess is that upstream would prefer the latter.

testsuite/tests/parsing/dropped_attributes.ml
tdelvecchio-jsc: How do you make an expect test work with something that's a syntax error?

testsuite/tests/parsing/dropped_attributes.ml
tdelvecchio-jsc: No, this was part of a compound bug where ocamlformat was dropping those two attributes. I'll remove them once.

tdelvecchio-jsc:
> This fix should probably be contributed upstream. I think the approach here is OK, but another possibility is to combine the lists of attributes -- my guess is that upstream would prefer the latter.

Is the process you're suggesting to make a PR upstream and then once it's accepted, cherry-pick it to our fork?

testsuite/tests/parsing/dropped_attributes.ml
goldfirere: Good point.

goldfirere:
> Is the process you're suggesting to make a PR upstream and then once it's accepted, cherry-pick it to our fork?

Not necessarily. The other way is to merge this now, but then also to make a PR upstream. The challenge there is that if upstream changes things, then you'd have to make yet another PR against ours. I'd honestly probably favor merging this here now but then posting it upstream and hoping for the best.

testsuite/tests/parsing/dropped_attributes.ml
tdelvecchio-jsc: Made an additional file to demonstrate both attributes.

testsuite/tests/parsing/dropped_attributes.ml
tdelvecchio-jsc: Removed the unrelated attributes.

testsuite/tests/parsing/dropped_attributes.ml
tdelvecchio-jsc: Combined the two tests back into one file using "toplevel", which lets you do this!

goldfirere (approved):

PR #3263, opened 2024-11-12
fix the autoconf script to correctly set CONFIGURE_ARGS

jvanburen:
as described in https://lists.gnu.org/archive/html/autoconf/2010-09/msg00093.html we need to set CONFIGURE_ARGS much earlier in the script, or it will get clobbered by macros later in the file


ccasin (approved):
Obviously correct, and matches upstream ocaml.

PR #3265, opened 2024-11-12
Runtime changes for `or_null`

dkalinichenko-js:
Runtime 5 changes for `or_null`, lifted from https://github.com/ocaml-flambda/flambda-backend/pull/2532. Tested by subsequent PRs.

@mshinwell is it worth porting those to runtime 4, or it's going to be gone soon anyways?


mshinwell:
I think we should port these to runtime4, yes.

dkalinichenko-js:
I took at try at porting the changes, would be great if someone more familiar with the runtime reviewed them.

runtime/caml/intext.h
```diff
@@ -99,6 +99,7 @@
 #define CODE_CUSTOM_LEN 0x18
 #define CODE_CUSTOM_FIXED 0x19
 #define CODE_UNBOXED_INT64 0x1a
+#define CODE_NULL 0x1b
```
stedolan, line 102: I'd prefer `0x1f` here. The available codes are `0x1b`, `0x1c`, `0x1d`, `0x1e`, `0x1f`. The first is the next one upstream will use, so I'd prefer us to start at the other end to minimise the likelihood of clashes.

runtime/caml/mlvalues.h
```diff
@@ -91,6 +91,9 @@ typedef opcode_t * code_t;
 #define Is_exception_result(v) (((v) & 3) == 2)
 #define Extract_exception(v) ((v) & ~3)
 
+/* or_null */
+#define Val_null ((value) 0)
```
stedolan, line 95: As far as I can tell, every bit of code in the runtime that handles an arbitrary OCaml value and checks `Is_block(v)` now needs to check `v != Val_nul && Is_block(v)`. Should we just integrate these into the `Is_block / Is_long` macros? `Val_null` is more like an immediate than a pointer, and making `Is_block(Val_null)` false breaks fewer invariants. (For instance, much code assumes that `Tag_val(v)` is defined if `Is_block(v)` holds).

stedolan (changes requested):

runtime/compare.c
```diff
@@ -140,9 +140,9 @@ static intnat do_compare_val(struct compare_stack* stk,
     poll_timer = COMPARE_POLL_PERIOD;
     while (--poll_timer > 0) {
       if (v1 == v2 && total) goto next_item;
-      if (Is_long(v1)) {
+      if (Is_long(v1) || v1 == Val_null) {
         if (v1 == v2) goto next_item;
-        if (Is_long(v2))
+        if (Is_long(v2) || v2 == Val_null)
           return Long_val(v1) - Long_val(v2);
```
stedolan, line 146: This looks wrong - it'll make Val_null compare equal to Val_int(0).

runtime/caml/intext.h
dkalinichenko-js: `CODE_UNBOXED_INT64` already seems to be a Jane Street extension, so we are already bound to clash. Should I also change that one?

runtime/compare.c
dkalinichenko-js: Right, fixed.

runtime/caml/mlvalues.h
mshinwell: I'm unsure about this, let's discuss next week.

dkalinichenko-js: I'm also somewhat worried about the performance impact. Perhaps we can skip the null pointer checks in some places.

stedolan: I think that having a gap between `Is_long` and `Is_block` is more error-prone, and I'd prefer that `Is_long(Val_null)` was true.

You should make `Is_block` a static inline function rather than a macro to avoid the issue you point out.

runtime/caml/intext.h
stedolan: Good point! Yeah, I'd prefer to see both of these at the high end.

dkalinichenko-js: Done.

runtime/caml/mlvalues.h
dkalinichenko-js: Done.

stedolan:
Current version looks good to me. I tried to find other places in the runtime that need special handling of nulls, and the only one I came across is `Obj.t`, which I think can now contain nulls because `Obj.field` can load them. I think `caml_obj_tag` should be updated to distinguish nulls from integers.

Reading the `obj.c` code made me happy with the change to make `Is_long(Val_null)` be true - `caml_lazy_update_to_forcing` is an example of some code that could have segfaulted with the prior version of the macros but is sound as written now, which we didn't notice before.

In user code, `Obj.is_int` now returns true if given a null `Obj.t`. It might be worth grepping for uses of this primitive, to see whether some uses need to think about nulls now.

I have some mild concerns about performance, because extra checks are being introduced into some of the hottest parts of the runtime code (the marking and scanning loops of the major and minor GC). However, these extra checks are very predictable, and so are unlikely to make things worse. Just to check, would you mind running [this prefetch GC benchmark](https://gist.github.com/stedolan/4369a0fa27820e27d6e56bee5e412896) before and after the patch, to verify that it is unaffected?

dkalinichenko-js:
> I think caml_obj_tag should be updated to distinguish nulls from integers.

What tag should I return for null pointers? Do I create a new one?

As is, the type system prevents `Obj.tag` from being called on `maybe_null` values, and maybe it should just stay that way. Same for `Obj.is_int` and the like.

dkalinichenko-js:
Compiling the benchmark with `-O3`, there does seem to be a slowdown, from ~1.11 s/gc to ~1.14 s/gc on my machine. I've bumped the n to 50, but the numbers are still pretty noisy, and the difference is about the size of the standard deviation.

dkalinichenko-js:
After investigation, the slowdown comes not from missed branches (those are essentially unchanged), but from the increased number of instructions in `Is_block`/`Is_long` (from 2 to 5). I'll try to optimize that.

dkalinichenko-js:
We can bring down the number of instructions required for `Is_block` back to 2 with some inline assembly (on AMD64):

```c
static inline uint64_t tzcnt_u64(uint64_t value) {
    uint64_t result;
    __asm__ volatile (
        "tzcnt %1, %0"
        : "=r" (result)
        : "r" (value)
    );
    return result;
}

Caml_inline int Is_block(intnat x) {
  return tzcnt_u64(x) & 0b111111;
}
```

This restores benchmarks back to previous numbers. We'd have to think about conditional compilation for different platforms, though.

dkalinichenko-js:
Given that `tzcnt` sets ZF to 1 if LSB equals 1 and CF to 1 is the input is 0, this can likely be optimized further.

dkalinichenko-js:
```c
Caml_inline bool Is_block(uint64_t value) {
    uint64_t result;
    __asm__ (
        "tzcnt %1, %0\n\t"
        "setnbe %b0"
        : "=r" (result)
        : "r" (value)
        : "cc"
    );
    return result;
}
```

This definition uses 2 instructions and returns either 0 or 1, but GCC can't figure out the last part for optimizations :(

It can't optimize the `_tzcnt_u64` intrinsic either.

stedolan:
Nice! Very cool trick using the ZF / CF flags like that.

Your idea might also work with the `shr reg, 1` instruction - it puts the low bit in `CF` and sets `ZF` if the other bits are zero, so `CF || ZF` is `Is_long`, and `shr` is a bit faster / more portable than `tzcnt`.

runtime/caml/intext.h
stedolan: I spoke to the people working on unboxed types, who are somewhat worried that there may be uses of `CODE_UNBOXED_INT64` in the wild in data that is expected to be portable across compiler versions. So, would you mind moving `CODE_UNBOXED_INT64` back to where it was? (Apologies for the back-and-forth here)

dkalinichenko-js:
After testing, I can't measure the slowdown from using `tzcnt`, the benchmark noise overwhelming the difference. It should, theoretically, be a bit slower, since the instruction latency goes from 1 to 3. 

`shr reg` sadly results in worse code generation, since it clobbers the input register.

dkalinichenko-js:
Updated with optimizations for `Is_block`/`Is_long` and `null_tag` added to `caml_obj_tag`.

stedolan (approved):

PR #3266, opened 2024-11-12
Use constant sorts in transl

goldfirere:
As discussed with @ccasin, all of the `transl` functions can really work over `Sort.Const.t`, not `Sort.t`. This PR makes this change.

Motivation: The second commit, which replaces `Jkind.t` with `Jkind.Sort.Const.t` in `variant_representation`, `label_description`, and `record_representation`. This actually was possible without the first commit, but there were a sad number of calls to `Jkind.Sort.of_const` to reconsistute the non-constant sorts for transl... so I wrote the first commit.

Review: commit-by-commit. The changes are very boring; it's the design change that's interesting.

* [ ] 1. f39bd0aef40d9b475118ee3c7199a6f35dcf7ab1 changes the transl functions to work over `Jkind.Sort.Const.t` instead of `Jkind.Sort.t` (aka `Jkind.sort`). The there are two sad bits. (a) There are *lots* more calls to `default_for_transl_and_get`. Previously, we called this function only in Typeopt when we actually needed the sort. Now this function has to be called for every sort stored in the typedtree. I think this is OK, though, because though the function call is written more, it is executed either less or the same amount (previously, sometimes a sort flowed through multiple paths, and it had to be `got`ten more than once). It might be reasonable to object to this commit based on the new calls. (b) There is a sad little dance in typedtree because a function needs to deal with both constant and non-constant sorts. The dance is local, but regrettable.
* [ ] 2. 540cc234e5cca4f47fd27949ea19b932b0a48345 changes the representations of structures to use `Jkind.Sort.Const.t`. This, in turn, is motivated by trying to use kinds as little as possible, because they're about to get a lot more complicated. But actually it's a nice change on its own.

If you like (2) but don't like (1) (reasonable! I'm not sure about (1) myself), you can check out https://github.com/goldfirere/flambda-backend/commit/ce8f169a7511a8c37cf04841048377d972a40014, which is a version of commit (2) without needing (1). It works. But see the number of `of_const` calls in translcore. These made me sad and is the concrete motivation for (1).

If you say to take (2) with the sad `of_const`s because that's better than (1), I won't push back.

Review: I'd love @ccasin to push back on the design. If he doesn't want to review the rest of the details, I'm sure @dkalinichenko-js could do so. I recommend looking at the commits separately.

### DIFF
--- a/typing/types.mli
+++ b/typing/types.mli
@@ -603,7 +603,7 @@ and record_representation =
   (* For an inlined record, we record the representation of the variant that
      contains it and the tag/representation of the relevant constructor of that
      variant. *)
-  | Record_boxed of jkind_l array
+  | Record_boxed of Jkind_types.Sort.Const.t array
   | Record_float (* All fields are floats *)
   | Record_ufloat
   (* All fields are [float#]s.  Same runtime representation as [Record_float],
@@ -623,7 +623,7 @@ and record_unboxed_product_representation =
 and variant_representation =
   | Variant_unboxed
   | Variant_boxed of (constructor_representation *
-                      jkind_l array) array
+                      Jkind_types.Sort.Const.t array) array
   (* The outer array has an element for each constructor. Each inner array
      has a jkind for each argument of the corresponding constructor.
 
@@ -650,7 +650,7 @@ and label_declaration =
     ld_mutable: mutability;
     ld_modalities: Mode.Modality.Value.Const.t;
     ld_type: type_expr;
-    ld_jkind : jkind_l;
+    ld_sort: Jkind_types.Sort.Const.t;
     ld_loc: Location.t;
     ld_attributes: Parsetree.attributes;
     ld_uid: Uid.t;
@@ -670,7 +670,7 @@ and constructor_argument =
   {
     ca_modalities: Mode.Modality.Value.Const.t;
     ca_type: type_expr;
-    ca_jkind: jkind_l;
+    ca_sort: Jkind_types.Sort.Const.t;
     ca_loc: Location.t;
   }
 
@@ -896,7 +896,7 @@ type 'a gen_label_description =
     lbl_mut: mutability;                (* Is this a mutable field? *)
     lbl_modalities: Mode.Modality.Value.Const.t;
                                         (* Modalities on the field *)
-    lbl_jkind : jkind_l;                (* Jkind of the argument *)
+    lbl_sort: Jkind_types.Sort.Const.t; (* Sort of the argument *)
     lbl_pos: int;                       (* Position in block *)
     lbl_num: int;                       (* Position in the type *)
     lbl_all: 'a gen_label_description array;   (* All the labels in this type *)
@@ -906,9 +906,6 @@ type 'a gen_label_description =
     lbl_attributes: Parsetree.attributes;
     lbl_uid: Uid.t;
   }
-(* CR layouts v5: once we allow [any] in record fields, change [lbl_jkind] to
-   be a [sort option].  This will allow a fast path for representability checks
-   at record construction, and currently only the sort is used anyway. *)
 
 type label_description = record_representation gen_label_description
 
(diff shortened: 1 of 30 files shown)


goldfirere:
I have rebased this to accommodate the fact that #3224 has been merged.

dkalinichenko-js:
I'm up to review this if you want.

ccasin:
Sorry - this has been behind review of Ryan's unboxed records PR on my stack.  I hope to finish that today.

I'm broadly happy for @dkalinichenko-js to review this.  But, question regarding the second commit: I believe we want, soon, to allow record declarations to have fields with layout `any`, and check for representability just when actually needed by the program. How will that look after this change?

goldfirere:
A good question. I think that moving to checking representability at construction would effectively remove the second commit. We won't be storing any sort information in these structures after that change, constant or no. Nothing in this patch makes the change you propose harder or easier, I believe.

ccasin:
> A good question. I think that moving to checking representability at construction would effectively remove the second commit. We won't be storing any sort information in these structures after that change, constant or no. Nothing in this patch makes the change you propose harder or easier, I believe.

Seems right.  Or maybe this field becomes an option to optimize the common case where it is known at declaration time.

Anyway, I recall our earlier conversation and still have no objections to this direction if it works (as it seems to).  So I think @dkalinichenko-js should review, including the questions about whether the first commit is a win.  Thanks in advance, Diana.

goldfirere:
> Seems right. Or maybe this field becomes an option to optimize the common case where it is known at declaration time.

Yes, that sounds likely.

dkalinichenko-js:
I don't think we should do (1). `default_for_transl_and_get` is a much more suspicious function than `of_const`, and we should keep the places it is called in to a minimum. Even if the number of calls stays the same, it becomes harder to keep track of them after (1).

goldfirere:
> Your thoughts on eventually eliminating `default_for_transl_and_get` from lambda?

Never? In theory, there could be another pass through the AST after type-checking and before translation; this pass would remove all sort variables. This would make great sense, and do a better job separating concerns. (GHC has this pass.) But it means reconstructing the entire AST! And it's so easy to do as a little initial step in translation.

One annoyance here is that the `lambda/` folder really contains at least two different things: the lambda language (part of the middle end) and the translation from typedtree into lambda (the interface between the front and the middle). I don't see a big problem with the translation being aware of aspects of the type-checker -- though I agree that having a separate pass to do defaulting would be cleaner in this regard.

So in some sense this patch brings the call to `default_for_transl_and_get` right to the boundary between the type-checker and the middle-end.... it's just perhaps on the wrong side of that boundary. But still better than deep in the bowels of Typeopt! (Recall that Typeopt is actually part of `lambda`, not the type-checker. It's just in the wrong directory.)

typing/types.mli
```diff
@@ -883,7 +883,7 @@ type label_description =
     lbl_mut: mutability;                (* Is this a mutable field? *)
     lbl_modalities: Mode.Modality.Value.Const.t;
                                         (* Modalities on the field *)
-    lbl_jkind : jkind_l;                (* Jkind of the argument *)
+    lbl_sort: Jkind_types.Sort.Const.t; (* Sort of the argument *)
```
dkalinichenko-js, line 886: I'm a little worried whether, in the future, we might want to ensure record fields or constructor arguments are non-null or are external. But for now, it's not an issue.

dkalinichenko-js (approved):

typing/types.mli
goldfirere: Yes! Very reasonable. This is part of why I think we've gotten the design of layouts and jkinds subtly wrong, because I think the details you mention (nullability and externality, specifically) might end up in the layout bit instead of the jkind bit. But we're not there in any case, and I'm pretty confident that `label_description` won't need a full jkinds (with mode-crossing behavior), so if we need externality/nullability, we'll just get it where it needs to be.

dkalinichenko-js: I'm fine with this then.

PR #3268, opened 2024-11-13
Fix Numbers.Int_base.compare

mshinwell:
There is an embarrassing error in `Numbers.Int_base.compare` which causes wrong results in the event of overflow.  This can lead to strange behaviour in Flambda 2, as this function is used for the comparison on e.g. `Simple.t`...

### DIFF
--- a/utils/numbers.ml
+++ b/utils/numbers.ml
@@ -17,7 +17,7 @@
 module Int_base = Identifiable.Make (struct
   type t = int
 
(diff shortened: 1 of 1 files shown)


lthls (approved):

goldfirere:
Should this fix also be propagated upstream?

Gbury:
It is, see  ocaml/ocaml#13606

PR #3269, opened 2024-11-13
Move `Language_extension` into parsing

liam923:
A recent change made `utils/language_extension.ml(i)` depend on `parsing/location.ml(i)`. There are other files in `parsing/` that depend on files in `utils/`, so there is a mutual dependency between files in `parsing/` and `utils/`. While this compiles fine in the compiler, this dependency structure is problematic in Merlin. To resolve this, this PR moves `Language_extension` to `parsing/`.

In the process, I also refactor parts of `Language_extension_kernel` into `Language_extension`. The parts of `Language_extension_kernel` that remain are used in `utils/profile_counters_functions.ml`, which prevents `utils/` from depending on `parsing/`.


liam923:
@goldfirere I agree this change is a bit awkward, but the alternatives @ccasin and I came up with aren't pretty either:
- Move `Location` into the `utils/` directory
- Refactor just the error-handing part of `Language_extensions` (the only part of it that uses `Location`) into `parsing/`
- Leave it as is, but in Merlin, move `Location` into `utils/`. This would require updating some scripts to special-case this.
- Leave it as is, but in Merlin, update build rules to tolerate this dependency structure. This would be difficult, especially when building internally.

ccasin (approved):
LGTM

PR #3270, opened 2024-11-13
Add check on code metadata return arity vs. Apply return arity for classic mode

mshinwell:
This mirrors a check in `Simplify_apply_expr`, and indeed, for some examples of malformed Lambda code this new check will go off before that one gets a chance.

The error message in Simplify has had its formatting improved in this PR too.


mshinwell:
(In case it isn't clear, line 1449 onwards is only hit when the application can be turned into a full direct call.)

lukemaurer (approved):

PR #3271, opened 2024-11-13
Put unique barrier creation behind a flag

anfelor:
This PR introduces a maturity level for uniqueness. I add two constants to `language_extension.ml` to describe the current maturity of uniqueness for DRF and for destruction. The uniqueness analysis will be run whenever uniqueness for DRF is enabled and the unique barriers will be inserted whenever uniqueness for destruction is enabled. ~~When uniqueness for DRF is enabled but not uniqueness for destruction, I throw an error whenever a unique barrier is required but not inserted. I added a testcase for this, but it requires that the maturities for DRF and destruction are different. I thus make uniqueness for DRF beta, which feels appropriate given the level of trust we already have in it and that we want to make it stable very soon.~~


lpw25:
>  When uniqueness for DRF is enabled but not uniqueness for destruction, I throw an error whenever a unique barrier is required but not inserted.

I would guess that this will go off in perfectly valid uses of uniqueness for DRF. If I have a record that holds a key and a function that needs that record uniquely -- so that it can get the key uniquely -- then I should be able to project out an `aliased` field of that record before calling the function. For full uniqueness we'd need to put a barrier on that projection in case the function destroys the record, so this check would presumably error.

goldfirere:
Yes I agree that this error is conservative. There is a workaround, by using a function to project out the needed field.

The question is: what’s worse? This over-eager check? Or the possibility that someone will use uniqueness in a way that requires the unique barriers but then will have miscompilation? On reflection, probably the check is worse, but it’s not completely obvious to me.

lpw25:
I think the check is likely to make uniqueness not usable for DRF. I think we should mostly not worry about people using uniqueness for things it isn't ready for without talking to us first.

goldfirere:
OK. Anton, is there a part of this that should be preserved, or do we close? Thanks.

lpw25:
> I think the check is likely to make uniqueness not usable for DRF

To illustrate why I think this. I expect the following to be a really common type with DRF:

```ocaml
type 'a owned =
  | Owned : 'k Key.t * ('a, 'k) Capsule.Data.t @ aliased -> 'a owned
```
and I think that will hit this barrier whenever you try to use it with the key-based `Data.t` functions.

lpw25:
Actually, I guess you wouldn't quite hit it for just using the key-based interface since you'd project out the key rather than use the whole value. Still, I think it wouldn't be long before people hit the error.

anfelor:
@lpw25 Yes, I agree that users could hit this error. For example, the following code is invalid using this PR:
```
match owned with
| Owned (key, capsule) -> do_something(capsule); owned
```
where `owned` is returned uniquely. That being said, you can always fix the error by writing:
```
match owned with
| Owned (key, capsule) -> do_something(capsule); Owned(key, capsule)
```
I think this inconvenience may be worth it for the extra protection. However, I don't have a strong opinion on this, since I would expect that we can make full uniqueness stable before either the above becomes too annoying or a user actually hits the bug. Let me know which version you prefer.

@goldfirere We should not close this PR, since it also splits the extension flag as you proposed.

typing/uniqueness_analysis.ml
```diff
@@ -619,9 +635,12 @@ end = struct
   let lift f t0 t1 =
     mapi2
       (fun first_is_of_second t0 t1 ->
-        try f t0 t1
-        with Usage.Error error ->
-          raise (Error (Usage { inner = error; first_is_of_second })))
+        try
```
nmatschke, line 638: Can't this be one `try ... with`? I.e.
```
try f t0 t1 with
| Usage.Error error -> raise (Error (Usage { inner = error; first_is_of_second }))
| Maybe_aliased.Error { occ } -> raise (Error (Unique_projection { occ }))
```

nmatschke (changes requested):
This is my first time using GitHub review, it seems I requested one change?

typing/uniqueness_analysis.ml
anfelor: I remove this error as requested above.

parsing/language_extension.ml
```diff
@@ -87,13 +87,17 @@ let is_erasable : type a. a t -> bool = function
   | Module_strengthening | SIMD | Labeled_tuples | Small_numbers | Instances ->
     false
 
+let maturity_of_unique_for_drf = Alpha
+
+let maturity_of_unique_for_destruction = Alpha
```
nmatschke, line 92: This is now unused; can we delete it and just have one `maturity_of_unique`?

typing/uniqueness_analysis.ml
```diff
@@ -163,12 +163,15 @@ end = struct
     | (_, access) :: _ -> access
 
   let add_barrier t uniq =
-    List.iter
-      (fun (_, access) ->
-        match access with
-        | Read barrier -> Unique_barrier.add_upper_bound uniq barrier
-        | _ -> ())
-      t
+    if Language_extension.is_at_least Unique Language_extension.Alpha
```
nmatschke, line 166: Why is this one a hardcoded `Alpha` rather than `maturity_of_unique...`?

nmatschke (changes requested):

parsing/language_extension.ml
anfelor: No, we later want to make `maturity_of_unique_for_drf` `Stable`, but want to keep `maturity_of_unique_for_destruction` as `Alpha`. That change has impacts on the production compiler though, so it shouldn't be done here.

typing/uniqueness_analysis.ml
anfelor: Thanks, that should be `maturity_of_unique_for_destruction`.

PR #3272, opened 2024-11-13
Ensure that all pattern-matching-constants mode-cross uniqueness

anfelor:
This PR makes all constants mode-cross uniqueness. This is necessary since we can not really ensure that the switcher code that is generated for `Tpat_construct` correctly handles the case where projections may not be pushed down. 

I create a new `jkind` for `constant_data` which includes strings, boxed floats and boxed integers. A value of kind `constant_data` is equivalent to a value of kind `immutable_data` except that it crosses uniqueness. In the testsuite, I adjusted the tests for uniqueness where I now generate a list of functions (which will never mode-cross uniqueness and linearity) to avoid wrong test-results arising from mode-crossing.


testsuite/tests/typing-layouts/jkinds.ml
```diff
@@ -507,15 +507,7 @@ type ('a : value mod unique) t = { unique_field : 'a }
...
-Line 2, characters 25-33:
-2 | let x = { unique_field = "string" }
-                             ^^^^^^^^
-Error: This expression has type "string" but an expression was expected of type
-         "('a : value mod unique)"
-       The kind of string is immutable_data
-         because it is the primitive type string.
-       But the kind of string must be a subkind of value mod unique
-         because of the definition of t at line 1, characters 0-54.
+val x : string t = {unique_field = "string"}
```
nmatschke, line 510: Should we have a new test demonstrating the error case here now?

typing/typedtree.mli
```diff
@@ -27,7 +27,10 @@ module Uid = Shape.Uid
 (* We define a new constant type that can represent unboxed values.
    This is currently used only in [Typedtree], but the long term goal
    is to share this definition with [Lambda] and completely replace the
-   usage of [Asttypes.constant] *)
+   usage of [Asttypes.constant].
+   For the soundness for the uniqueness extension it is important that all
```
nmatschke, line 31: I don't have a great sense for how this compares to other similar kinds of things in the compiler, but I'd be worried that this comment isn't scary enough. "Btw you might cause miscompilation if you ignore this comment". And it's easy to miss in review, when someone adds a new constant, if the diff doesn't include it.

It's better to break the build when someone needs to update something. For example, perhaps some file somewhere could have a test like:
```ocaml
type crosses_uniqueness = I_remembered_to_make_this_cross_uniqueness

let _all_constants_mode_cross_uniqueness : Typedtree.constant -> crosses_uniqueness = function
  | Const_int _ -> I_remembered_to_make_this_cross_uniqueness
  | Const_char _ -> I_remembered_to_make_this_cross_uniqueness
  (* ... *)
  | Const_unboxed_nativeint _ -> I_remembered_to_make_this_cross_uniqueness
;;
```

Ideally there's already some kind of convention for this sort of thing the compiler.

testsuite/tests/typing-unique/unique.ml
```diff
@@ -340,26 +349,29 @@ Line 1, characters 14-20:
...
 
-let inf1 (unique_ x : float) = unique_ let y = x in y
+let inf1 (unique_ x : non_mode_cross) = unique_ let y = x in y
 [%%expect{|
-val inf1 : float @ unique -> float = <fun>
+val inf1 : non_mode_cross @ unique -> non_mode_cross = <fun>
 |}]
 
-let inf2 (b : bool) (unique_ x : float) = unique_ let y = if b then x else 1.0 in y
+let inf2 (b : bool) (unique_ x : non_mode_cross) = unique_ let y = if b then x else [] in y
```
nmatschke, line 357: Iiuc `[]` should be coaliased someday, or something like that, so why not instead use `mk_non_mode_cross ()`?

testsuite/tests/typing-unique/unique_analysis.ml
```diff
@@ -180,21 +189,21 @@ Line 4, characters 21-23:
 
 
 
-let or_patterns1 : unique_ float list -> float list -> float =
+let or_patterns1 : unique_ non_mode_cross list -> non_mode_cross list -> non_mode_cross =
   fun x y -> match x, y with
   | z :: _, _ | _, z :: _ -> unique_ z
-  | _, _ -> 42.0
+  | _, _ -> []
```
nmatschke, line 195: Same comment as in unique.ml re: just use `mk_non_mode_cross ()`.

testsuite/tests/typing-unique/unique_analysis.ml
```diff
@@ -180,21 +189,21 @@ Line 4, characters 21-23:
...
                                          ^
 Error: This value is "aliased" but expected to be "unique".
 |}]
 
-let or_patterns2 : float list -> unique_ float list -> float =
+let or_patterns2 : non_mode_cross list -> unique_ non_mode_cross list -> non_mode_cross =
   fun x y -> match x, y with
   | z :: _, _ | _, z :: _ -> unique_ z
-  | _, _ -> 42.0
+  | _, _ -> []
```
nmatschke, line 206: Same comment as in unique.ml re: just use `mk_non_mode_cross ()`.

nmatschke (changes requested):
Though I can review the code, I feel ill-equipped to comment on whether this PR is a good idea.

First off, I do think `constant_data` vs `immutable_data` will be confusing for users, who mostly don't have the OCaml AST types burned into their brains. Even I was slightly surprised to be reminded that the language lets you match on `float`s, and I do a lot of PPX work!

But also, numbers I can sorta see, but `string` crossing uniqueness feels unintuitive to me. 

Is this our long-term solution to the problem, or something temporary until we can fix the code generation? I feel like I want someone else with a better understanding of our plans in this space to weigh in here.

lpw25:
I don't thing we should be making these types cross uniqueness. Being able to reuse an int64 as another int64 is definitely valuable. What is particularly difficult about pattern matching on constants?

lpw25:
If we think it will be a lot of work to get these patterns to have barriers associated with their operations, then it should be sufficient to just treat such patterns as uses of the associated values in the uniqueness analysis. For the types that do cross uniqueness that won't matter, and for the other types it will still be possible to do such pattern matching it will just require explicitly borrowing the value before matching on it.

PR #3273, opened 2024-11-14
Fix layouts bug for applications with omitted arguments

ccasin:
In the following code, we partially apply `f1`:
```ocaml
let f1 ~name:x y = unbox_float (x +. y)

let f2 (f : name:float -> float#) = f1 ~name:3.14

let f3 () = f2 (f1 3.15)
```
During translation to lambda, we translate `f1 3.15` to an eta-expanded version, roughly `fun param -> f1 ~name:param 3.15`.  Until this PR, that translation mistakenly assumes the layout of the return value of this eta-expanded function is the same as the layout of the original application.  This is wrong: the original application is partial so it is always a `value`, while the eta-expanded version in this case returns a `float#`.

I didn't find a particularly clean fix.  I considered two options: (1) having omitted arguments carry the sort that should be used when we construct this eta expansion, or (2) having the application node record the sort you'd get if you fully applied it to all the omitted arguments.

I went with (2), because this sort is really related to the whole application, not the individual omitted arguments.  But the result doesn't fill me with joy: applications now carry around this slightly weird sort that isn't actually the sort of the application, but rather the sort of some related compilation artifact.  And some functions in `transl` grow an extra parameter to pass this sort around.

Perhaps there is a third, better option - I'm open to suggestions.

I've added a test with several examples that previously caused the middle-end issues. One of the tests has two omitted arguments, to check a slight subtlety: we'll construct two eta expanded functions in that case, and only one of them should get the new return layout (the other returns a function).

Some existing layout error messages change slightly (for the better, I think) just because we now sometimes check that a function's return is representable a little earlier.

### DIFF
--- a/lambda/translcore.ml
+++ b/lambda/translcore.ml
@@ -1323,8 +1323,9 @@ and transl_apply ~scopes
        will occur exactly when all the arguments up to this parameter
        have been received.
   *)
-  let rec build_apply lam args loc pos ap_mode = function
-    | Omitted { mode_closure; mode_arg; mode_ret; sort_arg } :: l ->
+  let rec build_apply lam args loc pos ap_mode result_layout = function
+    | Omitted { mode_closure; mode_arg; mode_ret; sort_arg;
+                sort_eta_expansion_ret } :: l ->
         (* Out-of-order partial application; we will need to build a closure *)
         assert (pos = Rc_normal);
         let defs = ref [] in
@@ -1361,7 +1362,13 @@ and transl_apply ~scopes
           let mode = transl_alloc_mode_r mode_closure in
           let arg_mode = transl_alloc_mode_l mode_arg in
           let ret_mode = transl_alloc_mode_l mode_ret in
-          let body = build_apply handle [Lvar id_arg] loc Rc_normal ret_mode l in
+          let result_layout =
+            layout_of_sort (to_location loc) sort_eta_expansion_ret
+          in
+          let body =
+            build_apply handle [Lvar id_arg] loc Rc_normal ret_mode
+              result_layout l
+          in
           let nlocal =
             match join_locality_mode mode (join_locality_mode arg_mode ret_mode) with
             | Alloc_local -> 1
@@ -1388,7 +1395,8 @@ and transl_apply ~scopes
         List.fold_right
           (fun (id, layout, lam) body -> Llet(Strict, layout, id, lam, body))
           !defs body
-    | Arg (arg, _) :: l -> build_apply lam (arg :: args) loc pos ap_mode l
+    | Arg (arg, _) :: l ->
+      build_apply lam (arg :: args) loc pos ap_mode result_layout l
     | [] -> lapply lam (List.rev args) loc pos ap_mode result_layout
   in
   let args =
@@ -1400,7 +1408,7 @@ and transl_apply ~scopes
            Arg (transl_exp ~scopes sort_arg exp, layout_exp sort_arg exp))
       sargs
   in
-  build_apply lam [] loc position mode args
+  build_apply lam [] loc position mode result_layout args
 
 (* There are two cases in function translation:
     - [Tupled]. It takes a tupled argument, and we can flatten it.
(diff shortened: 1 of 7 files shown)


lpw25:
It's hard to tell without trying but I would expect option 1 to be cleaner. I don't really agree that the sort isn't related to the individual omitted parameters, they already carry around all the other information that is about the closure they are building, and the sort has no meaning for an application without omitted parameters.

typing/typedtree.mli
```diff
@@ -325,6 +325,12 @@ and expression_desc =
                          (Labelled "y", Some (Texp_constant Const_int 3))
                         ])
 
+            The sort here is _not_ the sort of this application.  It is the sort
+            of a _full_ application of the same function, and is used to
```
goldfirere, line 329: I don’t know what “full” means here. I think it means an application that includes all omitted arguments up to this point. But if you have an application that’s partial but omits no arguments (because the skipped arguments are after all the ones that are supplied), then the sort is the sort of the application. (This is basically implied below, but it still makes me worried about “full”.)

typing/typecore.ml
```diff
@@ -7548,6 +7548,7 @@ and type_argument ?explanation ?recarg env (mode : expected_mode) sarg
               |> Value.proj (Comonadic Areality)
               |> regional_to_global
               |> Locality.disallow_right,
+              ret_sort,
```
goldfirere, line 7551: I think this is right, because it looks like `args` cannot have any `Omitted`s. But this is a bit subtle (I had to check that fact), and so probably best to have a comment saying so. This would be easy to get wrong in the future.

typing/typecore.ml
```diff
@@ -7742,6 +7745,18 @@ and type_application env app_loc expected_mode position_and_mode
...
               (value_to_alloc_r2l funct_mode) sargs ret_tvar
           in
+          let sort_fully_applied =
+            (* [ty_ret] from [collect_apply_args] is the final return type of
+               the function after it has been applied to all arguments, even
+               those omitted in this application. This may be different from the
+               type of this application, if there are omitted arguments. We need
+               to know the sort of this final return type to compile such
+               applications. *)
+            match type_sort ~why:Function_result ~fixed:false env ty_ret with
```
goldfirere, line 7755: This call to `type_sort` makes me a little unhappy. Is there a way to get the information from `collect_apply_args`?

lambda/translcore.ml
```diff
@@ -1361,7 +1366,24 @@ and transl_apply ~scopes
...
           let arg_mode = transl_alloc_mode_l mode_arg in
           let ret_mode = transl_alloc_mode_l mode_ret in
-          let body = build_apply handle [Lvar id_arg] loc Rc_normal ret_mode l in
+          let result_layout =
+            (* If this is the last Omitted argument, then the layout of the
+               body of this eta expansion is the result layout of the original
+               function when fully applied.  If it's not the last omitted
+               argument, the body is another eta expanded function, so its
+               layout is value. *)
+            let last_omitted_arg =
```
goldfirere, line 1375: This calculation looks quadratic in the worst case.

goldfirere (approved):

ccasin:
> It's hard to tell without trying but I would expect option 1 to be cleaner. I don't really agree that the sort isn't related to the individual omitted parameters, they already carry around all the other information that is about the closure they are building, and the sort has no meaning for an application without omitted parameters.

You were correct.  I've pushed a version that does it the other way.  I thought it would be annoying to figure out the right value to put for sort in the `Omitted` - that we'd have to scan the remaining type and unprocessed arguments to figure out where the end is and what its sort is, basically duplicating the work already done by type_application.  But I was wrong - there is a place where this can be done very cleanly that I has missed

@goldfirere I've resolved all your comments because this has been rewritten.

typing/typedtree.mli
```diff
@@ -601,7 +601,8 @@ and omitted_parameter =
   { mode_closure : Mode.Alloc.r;
     mode_arg : Mode.Alloc.l;
     mode_ret : Mode.Alloc.l;
-    sort_arg : Jkind.sort }
+    sort_arg : Jkind.sort;
+    sort_eta_expansion_ret : Jkind.sort }
```
lpw25, line 605: I thing that this can be called just `sort_ret` because I think that `mode_ret` is the return mode for the same closure that this is the return sort of. Having them have very different names is a bit confusing.

lpw25 (approved):
LGTM. One small suggestion.

PR #3274, opened 2024-11-14
Unboxed records inside boxed records: Lambda changes

xclerc:
(very early draft, with only the changes to `lambda.mli`)


xclerc:
Actual work is happening in https://github.com/ocaml-flambda/flambda-backend/pull/3687.

PR #3275, opened 2024-11-14
Clarify the GC phase(s) for orphaning finalisers

NickBarnes:
Finalisers are currently orphaned, by calling `caml_orphan_finalisers()` from `domain_terminate()`, after `caml_finish_marking()` is called (which advances to the end of marking), so we will be in `Phase_sweep_and_mark_main`. We could assert that, but (as GC phases are internal to `major_gc.c`) it's an awkward pre-condition to impose on `domain_terminate()`. If the order or side-effects of calls from `domain_terminate()` into `major_gc.c` ever changes, then `caml_orphan_finalisers()` may break, as it will finish the current cycle (putting us into `Phase_sweep_main` and then `CAMLassert(caml_gc_phase == Phase_sweep_and_mark_main)`.
What we actually care about, to preserve invariants internal to `major_gc.c`, is that we orphan (and adopt) finalisers before we get into `Phase_mark_final` or `Phase_sweep_ephe`. So change the check and assertion here.

### DIFF
--- a/runtime/major_gc.c
+++ b/runtime/major_gc.c
@@ -456,13 +456,20 @@ void caml_orphan_finalisers (caml_domain_state* domain_state)
   if (f->todo_head != NULL || f->first.size != 0 || f->last.size != 0) {
     /* have some final structures */
     atomic_fetch_add(&num_domains_orphaning_finalisers, +1);
-    if (caml_gc_phase != Phase_sweep_and_mark_main) {
+    /* At present, we only call this function (during
+     * domain_terminate) after caml_finish_marking(), so should be in
+     * Phase_sweep_and_mark_main. However, we shouldn't rely on that;
+     * if we're not in a mark/sweep phase then we should force the
+     * current cycle into one. */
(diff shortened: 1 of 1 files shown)


NickBarnes:
Rebased.

stedolan:
This needs more of a comment explaining what the "constraints for orphaning finalisers" actually are. The new assertions suggest that it's fine to orphan finalisers both in `Phase_sweep_main` and `Phase_sweep_and_mark_main`, i.e. both before and after the roots (finaliser closures) have been marked.

I think the idea is that finalisers are always adopted unmarked and marked by the domain that adopts them. This can be accomplished either by handover before they're marked (in `Phase_sweep_main`), to be adopted at `caml_mark_roots_stw`, or by handover after they're marked (in `Phase_sweep_and_mark_main`), to be adopted during the next cycle when they have reverted to unmarked. Does that sound right? If so, could you replace the comment in `caml_orphan_finalisers` with something along those lines?

NickBarnes:
This is superseded by more recent change.

PR #3281, opened 2024-11-15
Disallow modes in `Pexp_coerce`

riaqn:
Currently modes in `Pexp_coerce` are parsed and dropped. This PR disallow modes in `Pexp_coerce`.

If in the future we decide to support it, we will add it (which requires parsetree change)

The second commit is the real change. The first commit is to restore a file accidentally removed by #3223


dvulakh (approved):
Looks good to me. This is also the easiest fix for the related issue in ocamlformat.

PR #3282, opened 2024-11-15
Zap modalities to identity for mli-less files

riaqn:
For `.ml` files without a corresponding `.mli`, previously we thought the generated `.cmi` will only be on the LHS of inclusion check, and therefore it's always better to infer the strongest modalities. This is obviously wrong, because another module could invoke `module type of` on the `.cmi` and use it on RHS of inclusion check. This has broken legacy code in practice.

This PR therefore zaps modalitites to identity, which is more legacy compatible.

Side note: In the future we should add two flags to override the default behavior:
- one flag on mli-less files, to indicate the direction of zapping when exporting `cmx`. It should error if `.mli` exists:
```
[@@@zap_modalities_to_floor]
(rest of the ml file)
``` 
- One flag on `module type of`, to indicate the direction of zapping when calculating the module type. It should error if the module already has a user-annotated module type:
```
module M = struct
  ...
end

module type S = module type of M [@@zap_modalities_to_floor]
```

### DIFF
--- a/testsuite/tests/typing-modes/def_nonportable.ml
+++ b/testsuite/tests/typing-modes/def_nonportable.ml
@@ -1,1 +1,0 @@
-let (f @ nonportable) x = x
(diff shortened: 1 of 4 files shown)


riaqn:
Supersededby #3462

PR #3283, opened 2024-11-17
`Yielding` mode axis

dkalinichenko-js:
Initial PR with the `yielding`/`unyielding` modes. The `Yielding` mode axis distinguishes between values that may yield (perform an effect, or do an equivalent operation like access a capsule with a password), from values that may not, with `unyielding < yielding`. Legacy OCaml values are `unyielding`.

Implements `Yielding` as a comonadic axis defaulting to `unyielding`. Closures over `yielding` values are `yielding`. This axis has no special handling otherwise.

Marks passwords in the Capsule API as yielding.

@lpw25 does this follow the design? Is the plan to have `@ local` secretly mean `@ local yielding`, and use `@ local unyielding` explicitly?

### DIFF
--- a/testsuite/tests/typing-modes/yielding.ml
+++ b/testsuite/tests/typing-modes/yielding.ml
@@ -1,0 +1,80 @@
+(* TEST
+ expect;
+*)
+
+(* CR dkalinichenko: allow [yielding] at toplevel? *)
+let my_effect : unit -> unit @@ yielding = print_endline "Hello, world!"
+[%%expect{|
+Line 1, characters 4-72:
+1 | let my_effect : unit -> unit @@ yielding = print_endline "Hello, world!"
+        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Error: This value is "yielding" but expected to be "unyielding".
+|}]
+
+let storage = ref ""
+
+let with_effect : ((string -> unit) @ local yielding -> 'a) -> 'a =
+  fun f -> f ((:=) storage)
+
+[%%expect{|
+val storage : string ref = {contents = ""}
+val with_effect : (local_ (string -> unit) @ yielding -> 'a) -> 'a = <fun>
+|}]
+
+let () = with_effect (fun k -> k "Hello, world!")
+
+let _ = !storage
+
+[%%expect{|
+- : string = "Hello, world!"
+|}]
+
+let run_yielding : (string -> unit) @ local yielding -> unit = fun f -> f "my string"
+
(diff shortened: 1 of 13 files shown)


lpw25:
I haven't looked at the code, but that sounds right assuming you have `unyielding < yielding`.

> Is the plan to have @ local secretly mean @ local yielding, and use @ local unyielding explicitly?

Yes, that's right.

Also, I think `global` will secretly mean `global unyielding` and people will have to write `global yielding` explicitly. That should only affect modalities and mode-crossing annotations.

It is probably also worth trying to change the behavior of the `zap_to_legacy` stuff so that, if the locality axis is already `local` it moves the `yielding` axis to `yielding` rather than `unyielding`. That would then allow us to have `data` types cross `yielding` without ending up printing `local unyielding` everywhere.

lpw25:
Probably not. It should probably only cross contention and the modes that are crossed by `'a`.

dkalinichenko-js:
Should `Mutex.t` mode-cross yielding? I think it should, as `immutable_data`?

typing/printtyp.ml
```diff
@@ -1419,7 +1419,9 @@ let tree_of_modes modes =
     tree_of_mode diff.uniqueness [Mode.Uniqueness.Const.Unique, Omd_new "unique"];
     tree_of_mode diff.portability [Mode.Portability.Const.Portable, Omd_new "portable"];
     tree_of_mode diff.contention [Mode.Contention.Const.Contended, Omd_new "contended";
-                                  Mode.Contention.Const.Shared, Omd_new "shared"]]
+                                  Mode.Contention.Const.Shared, Omd_new "shared"];
+    tree_of_mode diff.yielding [Mode.Yielding.Const.Yielding, Omd_new "yielding";
+                               Mode.Yielding.Const.Unyielding, Omd_new "unyielding"]]
```
riaqn, line 1424: Only the non-legacy modes need to be listed, because `a @ unyielding -> b` will be printed as `a -> b`.

riaqn (approved):
Reviewed everything except `yielding.ml` test.

dkalinichenko-js:
> Are we doing the "local implies yielding" in a future PR?
Yes.

testsuite/tests/typing-modes/yielding.ml
```diff
@@ -0,0 +1,80 @@
+(* TEST
+ expect;
+*)
+
+(* CR dkalinichenko: allow [yielding] at toplevel? *)
```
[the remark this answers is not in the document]
dkalinichenko-js: This makes sense, but we might want to allow more permissive top-level modes in the future.

typing/printtyp.ml
dkalinichenko-js: Fixed.

PR #3284, opened 2024-11-18
with-kinds

goldfirere:
This adds `with`-kinds to the compiler.

It is a *very* preliminary version. Importantly, you should not use an actual `with`-kind in your code and expect anything to work. Instead, this just provides *inferred* `with`-kinds, but that should be sufficient for looking through lists, `option`s, and arrays.

A rebase of this work (at https://github.com/ocaml-flambda/flambda-backend/tree/aspsmith/with-kinds-on-portable-stdlib) has been observed to build Base and Core, and this passes all the tests in the compiler testsuite. It is thus ready for review.

It is strongly suspected not to be performant enough for merging, though this fact has not been adequately assessed. I expect to follow this PR with another one to improve its performance. Search for an internal doc titled "Normalization for with-kinds" for more thoughts on how to do this. But nevertheless, I think this PR is now (Dec 24 '24) ready for correctness review.

Review is roughly by commit, as I enumerate:
* [x] 1. [8814e2c56173156b5ba4b55873d999458bab5c55](https://github.com/ocaml-flambda/flambda-backend/pull/3284/commits/8814e2c56173156b5ba4b55873d999458bab5c55) (first commit) and [25eadb9eb0a5a0ced02456283bd4222be1635eb7](https://github.com/ocaml-flambda/flambda-backend/pull/3284/commits/25eadb9eb0a5a0ced02456283bd4222be1635eb7) (last commit): @rtjoa (possibly in consultation with @ccasin). The first commit changes `narrow_to_manifest_jkind` to use `estimate_type_jkind` instead of `type_jkind_purely`. I don't remember my incentive for doing so, other than my general animosity toward `type_jkind_purely` (as documented in an internal doc titled "Let's kill type_jkind_purely"). You'll see some test case changes; please check these for reasonableness. The last commit undoes this change (restoring the use of `type_jkind_purely`) along with a long comment explaining why and how to do better. The last commit also contains several other small changes, mostly in response to unboxed-records work. (Some of the changes are not from that; feel free to delegate review of some files to @liam923 or @riaqn.) Some of the changes involve differences in behavior between principal-mode and non-principal-mode; these differences echo differences elsewhere and will be addressed later. (We basically don't care so much about principal mode.)
* [x] 2. [53b02d35b9e8513b1db08795d6b1b39cad3b25a3](https://github.com/ocaml-flambda/flambda-backend/pull/3284/commits/53b02d35b9e8513b1db08795d6b1b39cad3b25a3): @ccasin. This just deletes some code without affecting the testsuite. The code got in my way in a later patch.
* [x] 3. [0666ec87ec3b4d80e01da506718eb9647b8222f0](https://github.com/ocaml-flambda/flambda-backend/pull/3284/commits/0666ec87ec3b4d80e01da506718eb9647b8222f0): @ccasin. This removes the `from_subst` feature of `unify_var` in favor of the `in_subst_mode` check (which came temporally after `from_subst` but serves the same purpose and is propagated further). The change was necessary with `with`-kinds.
* [x] 4. [3274c88eaebec2ad1c62422c4748b750b77a9f63](https://github.com/ocaml-flambda/flambda-backend/pull/3284/commits/3274c88eaebec2ad1c62422c4748b750b77a9f63): @ccasin. We were doing a check on an uninstantiated type (essentially, the master copy), which caused trouble with `with`-kinds. Not anymore.
* [x] 5. [4ab10754812bce8aac26ff4436e625656f17dbdb](https://github.com/ocaml-flambda/flambda-backend/pull/3284/commits/4ab10754812bce8aac26ff4436e625656f17dbdb): @liam923. This is straightforward deletion of dead code.
* [x] 6. [8c85229c20e7c04b082f003c18e2d89ffd81a567](https://github.com/ocaml-flambda/flambda-backend/pull/3284/commits/8c85229c20e7c04b082f003c18e2d89ffd81a567): @liam923 and @riaqn. This is the main payload. I have broken it down below.
* [x] 7. [67700378aef9039e2a1079b1307605cab7b8d2da](https://github.com/ocaml-flambda/flambda-backend/pull/3284/commits/67700378aef9039e2a1079b1307605cab7b8d2da): @rtjoa (consulting with @ccasin or me as necessary). This adapts `with`-kinds to unboxed records and includes the deletion of the reimplementation of `estimate_type_jkind`.
* [x] 8. [259365684d7417e10e87e277033aac6f7b61eaa4](https://github.com/ocaml-flambda/flambda-backend/pull/3284/commits/259365684d7417e10e87e277033aac6f7b61eaa4): @liam923. This just updates test output. The change to the `templates` test has already been approved by @lukemaurer. (This test is about parameterized modules.)
* [x] 9. [c0a9d500bda0ce0c544d5d251d54180aecdad512](https://github.com/ocaml-flambda/flambda-backend/pull/3284/commits/c0a9d500bda0ce0c544d5d251d54180aecdad512): me, actually. @liam923 wrote this commit adding lots of tests, and I have already reviewed it.
* [x] 10. [f498428c3cfb7b4cd44eb93ebe0c447d5d5bdea9](https://github.com/ocaml-flambda/flambda-backend/pull/3284/commits/f498428c3cfb7b4cd44eb93ebe0c447d5d5bdea9): @liam923. This is the result of my review for the tests.
* [ ] 11. [57387ff66836df7ec6591372eb9c2e760bd104ef](https://github.com/ocaml-flambda/flambda-backend/pull/3284/commits/57387ff66836df7ec6591372eb9c2e760bd104ef) and [a20a171afec9f1e172f1b22d863327e656ef0203](https://github.com/ocaml-flambda/flambda-backend/pull/3284/commits/a20a171afec9f1e172f1b22d863327e656ef0203): @lpw25 (or possibly @ccasin). This is a test-case/fix pair that changes the use of the global mutable state of `univar_pairs` in ctype.
* [x] 12. [52da5d6bc7b4adec66b80bb3029ccc21d05c569b](https://github.com/ocaml-flambda/flambda-backend/pull/3284/commits/52da5d6bc7b4adec66b80bb3029ccc21d05c569b) and [97650466a699460a61cb42564ffd17245cc6818a](https://github.com/ocaml-flambda/flambda-backend/pull/3284/commits/97650466a699460a61cb42564ffd17245cc6818a): @liam923. This is another test-case/fix pair to fix a bug observed when building base and core.

And that's everything.

Now to break down the review of the main payload. The high-level overview is that much of this code was actually written by @liam923 and @riaqn in an early version of `with`-kinds. They should review. I break down where I have taken their code essentially as-is and where I have changed it (necessitating further review). The places where I took the code as-is do not need further review, as I reviewed the code as I adapted it. You can find your previous work at https://github.com/goldfirere/flambda-backend/pull/3/files.

* [x] In `jane/doc`: @riaqn. These are updates to the uniqueness documentation in light of `with`-kinds.
* [x] In `parsing`: @liam923. Just some comments.
* [x] In `testsuite`: @liam923. You're the most familiar with tests here -- as well as the expected use of `with`-kinds in base and core. I believe you probably already reviewed `with.ml` while you were writing new tests.
* [x] In `tools`: @liam923. This is a tiny change. (The reason for the eta-expansion is to avoid the value restriction.)
* [x] `ctype.ml`/`ctype.mli`: I think @riaqn. (I'm not sure exactly which of you did which pieces.)
  * [x] `instance_parameterized_kind`, `instance_declaration`, `jkind_subst`, and `is_principal` are unchanged from your previous work.
  * [x] `estimate_type_jkind` just has a small tweak compared to previous work (but would benefit from quick review)
  * [x] `type_jkind_purely_if_principal` is unchanged
  * [x] `constrain_type_jkind` just uses a different approach to getting `type_equal` than previously; pretty boring.
  * [x] `check_decl_jkind` and `constrain_decl_jkind` have been moved to the bottom of the file so that they could access `type_equal` (before I made the ref). Maybe this decision should be reversed.
  * [x] Changes in `intersect_type_jkind` and below are new; please review with care.
* [x] `includecore.ml`: @riaqn. The mode-cross stuff could use your expertise (though looks straightforward). The change in `type_declarations` is harder, but I already went over the basic idea with @ccasin (who agrees the approach is sound).
* [x] `jkind_axis.ml`/`jkind_axis.mli`: @riaqn. This is very directly inspired by previous work, but I added a bunch more functions to `Axis_collection`.
* [x] `jkind_types.ml`/`jkind_types.mli`: @riaqn. This is mostly boilerplate-y, but there is some stuff with left/right logic that you are the expert in.
* [x] `jkind.ml`/`jkind.mli`/`printtyp.ml`: Some combination of @liam923 and @riaqn (please work this out).
  * [x] The biggest piece needing careful scrutiny is `Bound.reduce_baggage`, which is a completely fresh reimplementation of `reduce_bound` from previous work. 
  * [x] Changes to `of_user_written_annotation_unchecked_level` are very close to prior work; compare the two to reduce review burden.
  * [x] Changes to `add_portability_and_contention_crossing` were a bit tricky in order to try to get decent behavior from this feature.
  * [x] Changes to `for_boxed_record`, `for_boxed_variant` are new and deserve scrutiny. These functions power the inference of `with`-kinds; an error here could easily lead to unsoundness.
  * [x] The `round_up` function is new; see usage sites for more context.
  * [x] The new `sub_jkind_l` function is horrid, but still needs to be sound. It will get improved later yet deserves a careful read-through here, too.
  * [x] Other changes in the file also deserve a read, of course, but they should hopefully be not too hard.
* [x] `mode.ml`/`mode_intf.mli`: @riaqn. This is just some refactoring.
* [x] `predef.ml`/`predef.mli`: Either @liam923 or @riaqn. This is basically all new (but easy).
* [x] `subst.ml`: Either @liam923 or @riaqn. This was easier than prior work due to #3266.
* [x] `typecore.ml`: Probably @riaqn. These changes are straightforward adaptations to the new types exposed by `Jkind`.
* [x] `typedecl.ml`/`typedecl.mli`: Not sure. I recommend either @liam923 or @riaqn take a spin, but if the going gets tough, let's loop in @ccasin. This file was hard, as you will see by the trail of comments I have left.
* [x] `typemode.ml`/`typemode.mli`: Whoever did this the first time around. I basically just adapted the prior work here. Only light review required.
* [x] `misc.ml`/`misc.mli`: Anyone. I guess this is what happens when Jane Street hires a Haskeller.
* [x] Other files (`datarepr.ml`, `typedecl_separability.ml`, `typemod.ml`, `typeopt.ml`, `types.ml`, `types.mli`): Anyone. These are all small changes.

Thanks, all!


goldfirere:
This is now ready for review. :tada: :gift: :christmas_tree: I have updated the top comment with a review plan, involving @rtjoa, @ccasin, @liam923, and @riaqn (and a tiny bit for @lpw25). @dkalinichenko-js could likely do some review, too, but given the fact that @liam923 and @riaqn have already taken a spin at the implementation here, it seems to make the most sense to have them review. But if the review burden here seems too high, looping in @dkalinichenko-js is a possible way to spread it out.

typing/typedecl.ml
```diff
@@ -1780,22 +1896,27 @@ let update_decl_jkind env dpath decl =
 
   (* check that the jkind computed from the kind matches the jkind
      annotation, which was stored in decl.type_jkind *)
-  if new_jkind != decl.type_jkind then
+  if new_jkind != decl.type_jkind then begin
     (* CR layouts v2.8: Consider making a function that doesn't compute
        histories for this use-case, which doesn't need it. *)
-    begin match Jkind.sub_jkind_l new_jkind decl.type_jkind with
+    (* The [decl.type_jkind] should never have [with]-types in it at this
+       point. *)
```
glittershark, line 1903: I suspect this is either wrong, or broken, and that's what's causing the failure in https://github.com/ocaml-flambda/flambda-backend/pull/3401#discussion_r1896882989

glittershark: in fact, if I replace this constantly-false `type_equal` with `Ctype.type_equal env`, a bunch of tests marked with "CR something: this should be accepted" or the like start being accepted. So I suspect that this should be actually using type equality. That said, what was the original idea behind this comment? I dug around and I couldn't figure out anything that might've been trying to strip baggage off the type_jkind of the decl, and in fact when constructing the type_jkind of the decl we explicitly add baggage from fields

glittershark: I've convinced myself enough that this was the right thing to do that I've pushed c03d79c09b4b93984ef60212dc765938345a7d63 making the change. feel free to revert if there's some subtlety I'm missing! but also the few branches I have based on this branch benefit from that change as well.

glittershark:
- [ ] [c03d79c](https://github.com/ocaml-flambda/flambda-backend/pull/3284/commits/c03d79c09b4b93984ef60212dc765938345a7d63) needs review, probably by @goldfirere

testsuite/tests/typing-layouts-products/basics_alpha.ml
```diff
@@ -66,7 +66,27 @@ type t2 : any_non_null
...
+         & any mod global with t1
+t2 unique with t1
+t2 many with t1
+t2
+                   uncontended with t1
+t2 portable with t1
+t2
+                   external_ with t1
+t2 non_null with t1
+t2
```
liam923, line 89: This formatting is pretty terrible. But I think it's okay to leave this for now because:
a) I don't think any users outside our team will see an error like this
b) We plan to fix printing soon

glittershark: > b) We plan to fix printing soon
 
extremely soon:

https://github.com/ocaml-flambda/flambda-backend/pull/3410

testsuite/tests/typing-jkind-bounds/basics.ml
```diff
@@ -1573,3 +1573,17 @@ type 'a t =
 type 'a u
 type 'a t = None | Some of ('a * 'a) t u
 |}]
+
+(*********************************)
+(* Test 13: Bug in class methods *)
```
liam923, line 1577: I don't see the bug here - did this used to fail?

testsuite/tests/typing-jkind-bounds/composite.ml
```diff
@@ -413,6 +414,8 @@ and 'a u = None | Some of 'a t
 |}]
 
 let foo (t : _ t @@ contended) = use_uncontended t
+(* CR reisenberg: This looks unsound, unless it's inferring a kind
+   of [immutable_data] for the argument but failing to print this.*)
```
liam923, line 411: I think this is sound because a `'a t` doesn't ever actually contain a `'a`. I believe it's kind should just be `immutable_data` (although I am pleasantly surprised that this gets inferred correctly).

testsuite/tests/typing-jkind-bounds/composite.ml
```diff
@@ -690,14 +693,22 @@ type 'a t = None | Some of ('a * 'a) t
...
 
+let foo (t : int t @@ contended) = use_uncontended t
+(* CR layouts v2.8: this should work when we get tuples working *)
+[%%expect {|
+Line 1, characters 51-52:
+1 | let foo (t : int t @@ contended) = use_uncontended t
+                                                       ^
+Error: This value is "contended" but expected to be "uncontended".
+|}]
+
```
liam923: I actually think that both these cases should be accepted, independently of tuples being accepted. Since a `'a t` can never contain a `'a`, its kind should always be `immutable_data`.

testsuite/tests/typing-jkind-bounds/composite.ml
```diff
@@ -740,7 +751,7 @@ type 'a t = None | Some of 'a t * 'a t
 |}]
 
 let foo (t : _ t @@ contended) = use_uncontended t
-(* CR layouts v2.8: fix this *)
+(* CR reisenberg: This looks unsound *)
```
liam923, line 747: Another instance of "I think this is okay because `'a t` can't contain a `'a`"

testsuite/tests/typing-jkind-bounds/records.ml
```diff
@@ -805,10 +798,131 @@ Error: Signature mismatch:
...
+end
+
+[%%expect{|
+module M : sig type 'a t : immutable_data end
+|}]
+
+(****************************************)
+
+type ('a : immutable_data) t : immutable_data = { x : 'a list }
+[%%expect {|
```
liam923, line 911: I think this should get a CR?

testsuite/tests/typing-jkind-bounds/variants.ml
```diff
@@ -526,14 +527,15 @@ Line 1, characters 61-62:
...
-(**** Test 4: Record types satisfy type constraints when appropriate ****)
+(**** Test 4: Variant types satisfy type constraints when appropriate ****)
 type t = Foo of int | Bar of string
 let t = Foo 10
 let () =
   cross_many t;
   cross_portable t;
   cross_uncontended t
 
+(* CR layouts v2.8: Fix this in the principal case. I think. *)
```
liam923, line 538: I agree

testsuite/tests/typing-layouts/illegal-across-modules/test.ml
```diff
@@ -0,0 +1,14 @@
...
+   flags = "-allow-illegal-crossing";
+   ocamlc.byte;
+   module = "shadow_stdlib.ml";
+   flags = "";
+   ocamlc.byte;
+   flags = "-allow-illegal-crossing";
+   module = "bug.ml";
+   ocamlc.byte;
+   check-ocamlc.byte-output;
+ *)
```
liam923, line 14: I'm surprised that the `-allow-illegal-crossing` flag has anything to do with this error since illegal crossing isn't actually used anywhere. But I do agree that the error only occurs when the flag is included. Do you know what's going on here? I had thought that maybe the check in `Type_decl.check_kind_coherence` that happens only when illegal crossing is enabled was responsible, but based on testing, that doesn't seem to be the case. This leads me to the conclusion that `add_portability_and_contention_crossing` isn't a no-op when no annotation is written, which I didn't expect.

utils/misc.ml
```diff
@@ -564,11 +579,20 @@ module Stdlib = struct
       let rec all_unit = function
         | [] -> return ()
         | t :: ts -> t >>= fun () -> all_unit ts
+
+      module Syntax = struct
+        let[@inline] (let+) t f = map f t
+        let[@inline] (and+) a b = a >>= fun a -> b >>= fun b -> return (a,b)
```
liam923, line 585: I think it's worth defining a `both` outside the `Syntax` module and implementing `and+` in terms of it.

utils/misc.ml
```diff
@@ -530,24 +530,39 @@ module Stdlib = struct
...
+      module Syntax : sig
+        val (let+) : ('a, 'e) t -> ('a -> 'b) -> ('b, 'e) t
+        val (and+) : ('a, 'e) t -> ('b, 'e) t -> ('a * 'b, 'e) t
+        val (let*) : ('a, 'e) t -> ('a -> ('b, 'e) t) -> ('b, 'e) t
+        val (and*) : ('a, 'e) t -> ('b, 'e) t -> ('a * 'b, 'e) t
+      end
     end
 
-    module Make (X : Basic) = struct
+    module[@inline] Make (X : Basic) = struct
```
liam923, line 562: It seems worthwhile to make `Make` and `Make2` share an implementation, a la `Base.Monad`.

liam923: And I think the same can be true of the signatures:
```
module type S = sig
  type 'a t
  include S2 with type ('a, 'e) t := 'a t
end
```

jane/doc/extensions/uniqueness/intro.md
```diff
@@ -76,9 +76,8 @@ type delayed_free = { id : int; callback : unit -> unit }
 let get_id : delayed_free @ once -> int @ many = fun t -> t.id
 ```
 
-We are working on a feature which will make it possible to use a value as `many`
-if its type prevents it from containing a function. For example, you will then
-be able to use an `int list @ once aliased` as `many` (but not as `unique`).
+This mode-crossing works through container types. For example, you can use an
```
riaqn, line 79: Instead of starting from immediate and extending to container types, we can be more general about types, but specific to uniqueness and linearity.

> We make an exception to this rule for values of those types that cross these
modes. A type crosses linearity if it doesn't contain functions; a type crosses uniqueness if it doesn't contain memory location subject to overwriting (even though overwriting hasn't been implemented yet). The mode crossing capability of a type is subject to explicit weakening (c.f. [proposals/kind-inference.md]).
 
> For example, `int`s and other immediates can be used as `unique` and
`many`, no matter their mode. For example:
```ocaml
type delayed_free = { id : int; callback : unit -> unit }
let get_id : delayed_free @ once -> int @ many = fun t -> t.id
```

> For another example, you can use an `int list @ once aliased` as `many` (but not as `unique`).

typing/jkind_axis.mli
```diff
@@ -12,29 +12,8 @@
...
-
-  val join : t -> t -> t
-
-  val print : Format.formatter -> t -> unit
-end
-
-(** Adapt a [Lattice] to be an [Axis_s] *)
-module Of_lattice (L : Mode_intf.Lattice) : Axis_s with type t = L.t
+(** Re-export *)
+module type Lattice = Mode_intf.Lattice
```
riaqn, line 16: The extra functions `equal` and `less_or_equal` are not used by `mode.ml` or `solver.ml`; they also require some extra invariants between `le` and `equal` and `less_or_equal` that are not guaranteed. Maybe we can instead have:
```
Misc.Le_result.less_or_equal : ('a -> 'a -> bool) -> 'a -> 'a -> Misc.Le_result.t
Misc.Le_result.equal : ('a -> 'a -> bool) -> 'a -> 'a -> bool
```
And then instead of `Mode.Uniqueness.less_or_equal a b` we write `Misc.Le_result.less_or_equal Mode.Uniqueness.le a b`.

typing/typedecl.ml
```diff
@@ -1075,7 +1113,9 @@ let rec check_constraints_rec env loc visited ty =
            *already* violate the constraints -- we need to report a problem with
            the unexpanded types, or we get errors that talk about the same type
            twice.  This is generally true for constraint errors. *)
-        match Ctype.matches ~expand_error_trace:false env ty ty' with
+        match Ctype.matches ~expand_error_trace:false env
+                (Ctype.instance ty) ty'
+        with
```
ccasin, line 1118: Probably this should be posted upstream, too.

typing/typedecl.ml
```diff
@@ -974,13 +1009,16 @@ let transl_declaration env sdecl (id, uid) =
...
          decompose the product of [any]s and recurse on the labels.
          See https://github.com/ocaml-flambda/flambda-backend/pull/3399. *)
-      match sdecl.ptype_kind with
-      | Ptype_record_unboxed_product lbls
-        when Jkind.has_layout_any (Jkind.terrible_relax_l jkind) ->
-        Jkind.Builtin.product ~why:Unboxed_record
-          (List.map (fun _ -> jkind) lbls)
-      | Ptype_abstract | Ptype_variant _ | Ptype_record _
-      | Ptype_record_unboxed_product _ | Ptype_open -> jkind
+      match kind with
```
rtjoa, line 1012: Nit: would be nice to keep the comment in sync as this is a confusing piece of typedecl (my fault :slightly_smiling_face:). A few lines above, I would change (GitHub won't let me comment there):
```diff
-      (* Hack: unboxed records are given a product-of-[any]s jkind
+      (* Hack: unboxed records are given a product-of-[any]s layout
```

testsuite/tests/typing-layouts-unboxed-records/recursive.ml
```diff
@@ -42,50 +42,30 @@ type bad = #{ bad : bad ; i : int}
...
-       The layout of bad is any & any
+Error: The layout of type "bad" is (value & value) & value
+         because it is an unboxed record.
+       But the layout of type "bad" must be a sublayout of value & value
          because it is an unboxed record.
-       But the layout of bad must be representable
-         because it is the type of record field bad.
 |}]
 
+(* It might be nice to reject, but it seems harmless to accept. *)
```
rtjoa, line 51: This also seems good to me (that this no longer fails the jkind check). And we'll explicitly disallow it with #3407 anyway.

rtjoa (commented):
The changes in 8814e2c56173156b5ba4b55873d999458bab5c55 and the non-test changes in 25eadb9eb0a5a0ced02456283bd4222be1635eb7 look good to me. I'd appreciate if someone more familiar with the guts of the PR could take a second look at the test changes in 25eadb9eb0a5a0ced02456283bd4222be1635eb7, which mostly seem to reflect changes from other commits.

typing/jkind.mli
```diff
@@ -371,17 +407,24 @@ val of_type_decl :
...
+val for_boxed_record : Types.label_declaration list -> jkind_l
 
-(** Choose an appropriate jkind for a boxed variant type, given whether
-    all of the fields of all of its constructors are [void]. *)
-val for_boxed_variant : all_voids:bool -> jkind_l
+(** Choose an appropriate jkind for an unboxed record type. Uses [jkind_of_type]
+    only in the singleton case, where the jkind of the unboxed record must match
+    that of the single field. *)
+val for_unboxed_record :
+  jkind_of_type:(Types.type_expr -> jkind_l) ->
```
rtjoa, line 422: Nit: It might be slightly more flexible and straightforward if this function took a `jkind_of_first_type`, like `Builtin.product`, instead of `jkind_of_type`.

rtjoa (commented):
67700378aef9039e2a1079b1307605cab7b8d2da looks good, glad to see `type_jkind_deep` gone.

typing/predef.ml
```diff
@@ -372,6 +386,8 @@ let build_initial_env add_type add_extension empty_env =
   |> add_type1 ident_lazy_t
        ~variance:Variance.covariant
        ~separability:Separability.Ind
+       (* CR layouts v2.8: Can [lazy_t] mode-cross at all? *)
```
riaqn, line 389: It can at least cross locality, because it's always heap-allocated.
It might also cross portability, linearity, uniqueness subject to its parameter. But I'm also fine not doing that for now (and wait until users complains).

typing/predef.ml
```diff
@@ -493,8 +519,9 @@ let add_or_null add_type env =
      For now, we mark the type argument as [Separability.Ind] to permit
      the most argument types, and forbid arrays from accepting [or_null]s.
      In the future, we will track separability in the jkind system. *)
+  (* CR layouts v2.8: Add baggage and more mode crossing here. *)
```
riaqn, line 522: This probably crosses all axes, subject to its parameter? Except maybe externality and nullability.

typing/typedecl.ml
goldfirere: Your change is clearly an improvement, and it's not worth figuring out why I thought with-kinds wouldn't be there.

testsuite/tests/typing-jkind-bounds/basics.ml
goldfirere: Yes, it did, but the exact details got lost because I wasn't careful about separating commits. I don't think it's worth teasing out.

testsuite/tests/typing-jkind-bounds/composite.ml
goldfirere: Right you are. I've updated the comment to read

```
(* CR layouts v2.8: this should work, but the recursive expansion
   of with-bounds presumably runs out of fuel and gives up. *)
```

testsuite/tests/typing-jkind-bounds/records.ml
goldfirere: Saying what? Looks ok to me. Note that `'a : immutable_data`, so no with-bounds are required for soundness.

testsuite/tests/typing-layouts/illegal-across-modules/test.ml
goldfirere: `type_jkind` sometimes gets stuff other than an annotation. I don't think it's worth the archaeology here, especially because `-allow-illegal-crossing` is on the chopping block.

utils/misc.ml
goldfirere: Good ideas.

jane/doc/extensions/uniqueness/intro.md
goldfirere: Yes, good idea.

typing/jkind_axis.mli
goldfirere: Ok

typing/typedecl.ml
goldfirere: https://github.com/ocaml/ocaml/pull/13725

testsuite/tests/typing-layouts-unboxed-records/basics.ml
```diff
@@ -752,3 +752,44 @@ Error:
...
+int -> q
+                             uncontended with int -> int
+int -> q
+                             portable with int -> int
+int -> q
+                             external_ with int -> int
+int -> q
+                             non_null with int -> int
+int -> q
+         because it is an unboxed record.
```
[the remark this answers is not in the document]
goldfirere: It's simpler: we just never try to simplify `with`-types. But you're right to be surprised... it's just that this first version is dump. @glittershark is making this better right now.

goldfirere:
Thanks for the review so far! I've handled the comments above. There is a little more review then:

* [ ] https://github.com/ocaml-flambda/flambda-backend/pull/3284/commits/a219496176fff945aee03e774f85a9fd15d81a6a @liam923 
* [ ] https://github.com/ocaml-flambda/flambda-backend/pull/3284/commits/5b8688122c0dfb283f9fec1aab48428f2f52919c @riaqn 
* [ ] https://github.com/ocaml-flambda/flambda-backend/pull/3284/commits/ad92250903334c44f220ef4c34e2c247ea929296 @riaqn 
* [ ] https://github.com/ocaml-flambda/flambda-backend/pull/3284/commits/2fd600cdb395577c15a06d8e60f3da1be4dbd621 @rtjoa 
* [ ] https://github.com/ocaml-flambda/flambda-backend/pull/3284/commits/1f6e7bfc47840281581f2fad7864e66f1c6ae7de @rtjoa 
* [ ] https://github.com/ocaml-flambda/flambda-backend/pull/3284/commits/65f2e6b2d575c9e7b354342070f3365421ef0bea @riaqn 
* [ ] https://github.com/ocaml-flambda/flambda-backend/pull/3284/commits/25eadb9eb0a5a0ced02456283bd4222be1635eb7 This was for @rtjoa but he wanted more help reviewing tests. @liam923 is our testing person, so please review the tests (only) in this commit.

Thanks!

goldfirere:
It looks like the checkmarks in my overview might not have gotten noticed. In particular, I believe @riaqn and @rtjoa have some check-marks to click. Thank you!

liam923:
@goldfirere Please review:
- [ ] 59ff14bd6cee2975ab6ba5b677d2d5ce55deb19d

testsuite/tests/typing-layouts/allow_illegal_crossing.ml
```diff
@@ -12,21 +12,35 @@ type a
 type b : value mod portable = { a : int; b : int }
 [%%expect {|
 type a
-type b : value mod portable = { a : int; b : int; }
+type b = { a : int; b : int; }
 |}]
```
liam923, line 16: This test is no longer testing the thing it was meant to test. `b` is meant to have a type that is normally inferred to not be portable. So I think the `int`s should be replaced with a type that doesn't cross portability.

The same is true for most of the tests in this file. But maybe this isn't worth the effort of fixing, because:
1. We plan to remove illegal crossing very soon, and I believe @glittershark plans to delete this file rather than incorporating it into the trust-me annotation tests.
2. I see that you've added some tests where the inferred type does not cross.

testsuite/tests/typing-layouts/jkinds.ml
```diff
@@ -1005,22 +982,15 @@ val v : int = 5
...
-Line 1, characters 0-70:
-1 | type ('a : immediate) t : value mod many portable = { mutable x : 'a }
-    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-Error: The kind of type "t" is value
-         because it's a boxed record type.
-       But the kind of type "t" must be a subkind of mutable_data
-         because of the annotation on the declaration of the type t.
+type ('a : immediate) t = { mutable x : 'a; }
 |}]
-(* CR layouts v2.8: this should be accepted *)
```
liam923, line 1016: It's very satisfying to see these crs go away

testsuite/tests/typing-layouts/modules.ml
```diff
@@ -340,10 +340,11 @@ type ('a : value) t4_val
 type t4 = M4.s t4_val;;
 [%%expect {|
 module F4 : functor (X : sig type t end) -> sig type s = Foo of X.t end
-module M4 : sig type s end
+module M4 : sig type s : immutable_data end
 type 'a t4_val
 type t4 = M4.s t4_val
 |}]
+(* CR layouts v2.8: The appearance of [immutable_data] there is just wrong. *)
```
liam923, line 347: I'm pleasantly surprised by this output - I think this is right? In `F4`, `s` should have kind `immutable_data with X.t`. And then after the functor application, `X.t` is `t = T`, which is immediate. So `M4.s` really is `immutable_data`

testsuite/tests/typing-layouts/modules_alpha.ml
```diff
@@ -283,11 +283,12 @@ type ('a : void) t4_void
 type t4 = M4.s t4_val;;
 [%%expect {|
 module F4 : functor (X : sig type t end) -> sig type s = Foo of X.t end
-module M4 : sig type s end
+module M4 : sig type s : immutable_data end
 type 'a t4_val
 type ('a : void) t4_void
 type t4 = M4.s t4_val
 |}]
+(* CR layouts v2.8: The appearance of [immutable_data] there is just wrong. *)
```
liam923, line 291: For the same reason as the test in modules.ml, I think this is right

typing/jkind.ml
```diff
@@ -1784,92 +2164,151 @@ let combine_histories reason (Pack k1) (Pack k2) =
...
+              | None ->
+                let (module Bound_ops) = Axis.get axis in
+                let baggage1 = Baggage.as_list bound1.baggage in
+                let baggage2 = Baggage.as_list bound2.baggage in
+                let modifiers =
+                  Bound_ops.equal bound1.modifier bound2.modifier
+                in
+                let baggages =
+                  List.compare_lengths baggage1 baggage2 = 0
+                  && List.for_all2 type_equal baggage1 baggage2
```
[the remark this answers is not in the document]
glittershark: yep

typing/jkind.ml
```diff
@@ -356,56 +363,332 @@ module Error = struct
...
+     if we have any baggage in either argument, the result is an l-Baggage, as
+     required. This might change once we have arrow kinds, but we'll deal with
+     that when we get there. *)
+  let join (type l r) (bag1 : (l * r) t) (bag2 : (l * r) t) : (l * r) t =
+    match bag1, bag2 with
+    | No_baggage, No_baggage -> No_baggage
+    | No_baggage, b -> b
+    | b, No_baggage -> b (* CR layouts v2.8: List concatentation is slow. *)
+    | Baggage (ty1, tys1), Baggage (ty2, tys2) ->
+      Baggage (ty1, tys1 @ (ty2 :: tys2))
```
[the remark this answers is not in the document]
glittershark: I did so in a child PR

typing/jkind_axis.ml
```diff
@@ -200,21 +174,34 @@ module Axis = struct
     | Modal Contention -> "contention"
     | Nonmodal Externality -> "externality"
     | Nonmodal Nullability -> "nullability"
+
+  let is_deep (type a) : a t -> bool = function
```
riaqn, line 176: Maybe rename this to `is_modal`? I.e., our modes are by default deep, and given the current axes as samples, the distinction seems to be between modal vs nonmodal? Do we have imaginary axes violating that?

This is also a good place to insert a CR saying that `Externality` is in fact `Modal` and will be made so in the future.

typing/ctype.ml
```diff
@@ -1828,6 +1836,20 @@ let subst env level priv abbrev oty params args body =
     undo_abbrev ();
     raise Cannot_subst
 
+let jkind_subst env level params args jkind =
```
riaqn, line 1839: I think @lpw25 should review this function.

typing/ctype.ml
```diff
@@ -6773,6 +6830,14 @@ let nondep_type_decl env mid is_covariant decl =
                 Private
             with Nondep_cannot_erase _ ->
               None, decl.type_private
+    and jkind =
```
riaqn, line 6779: I think @lpw25 should review these lines.

typing/ctype.ml
```diff
@@ -3396,11 +3444,13 @@ let add_jkind_equation ~reason uenv destination jkind1 =
         begin
           try
             let decl = Env.find_type p env in
-            if not (Jkind.equal (Jkind.terrible_relax_l jkind)
-                                (Jkind.terrible_relax_l decl.type_jkind))
-            then
-              let refined_decl = { decl with type_jkind = jkind } in
-              set_env uenv (Env.add_local_constraint p refined_decl env);
+            (* CR layouts v2.8: We might be able to do better here. *)
```
riaqn, line 3387: I think @lpw25 should review these lines.

typing/ctype.ml
lpw25: LGTM

typing/ctype.ml
lpw25: I think there is a bug here. In the general case, we would require a full meet operation on l-kinds. For example, if we have:
```
type t : value mod portable with X.t
type s : value mod portable with Y.t
```
and we learn that `t` equals `s` then I think we have no way to represent the resulting kind of `t` -- it crosses portability as long as either `X.t` or `Y.t` does. In such a case, we cannot add the equality without weakening the kind of either `t` or `s`. That might make type expressions in the environment ill-kinded, so I think that in such cases we are forced not to add the equality.

This is bad -- revealing more about a type can make fewer things type check -- but note that you already get this kind of issue when you expose the constructors of a type, so it's bad in a way we already have.

In the long term I expect us to move to a model that supports meets in l-kinds, which will allow us to fix this issue properly.

typing/ctype.ml
lpw25: I think it would be worth writing a proper `nondep_jkind` for l-kinds here, that weakened only the parts of the kind that referred to paths being removed when `is_covariant` is true. (This is ignoring the fact that [the whole `is_covariant` thing isn't really sound](https://github.com/ocaml/ocaml/issues/10491). But it's probably better to be consistent with the rest of the `nondep` machinery here.)

goldfirere:
* [ ] Just pushed [4473fb80b8ade4e09f4bb3a62d1e0ecd4f4d9ace](https://github.com/ocaml-flambda/flambda-backend/pull/3284/commits/4473fb80b8ade4e09f4bb3a62d1e0ecd4f4d9ace), which undoes a previous commit, as reviewed by @ccasin 

This reversion was motivated by trying to post the [change upstream](https://github.com/ocaml/ocaml/pull/13725) and pushback there. It turns out it's unnecessary.

typing/jkind.ml
```diff
@@ -356,56 +363,332 @@ module Error = struct
...
+      (l * allowed, axis) t ->
+      Misc.Le_result.t =
+   fun ~type_equal ~jkind_of_type ~axis { modifier = m1; baggage = b1 }
+       { modifier = m2; baggage = b2 } ->
+    let (module Axis_ops) = Axis.get axis in
+    match b1, b2 with
+    | No_baggage, No_baggage -> Axis_ops.less_or_equal m1 m2
+    (* CR layouts v2.8: This should expand types on the left. *)
+    | Baggage (ty, tys), No_baggage ->
+      if Axis_ops.le Axis_ops.max m2
```
liam923, line 578: This test is redundant since `reduce_baggage` terminates early when the bound is max

testsuite/tests/typing-layouts/allow_illegal_crossing.ml
```diff
@@ -45,50 +59,69 @@ Error: The kind of type "a" is value
...
 type t : value mod portable uncontended = { a : int; b : int }
 [%%expect {|
-type t : value mod uncontended portable = { a : int; b : int; }
+type t = { a : int; b : int; }
 |}]
 
 type ('a, 'b) t : value mod portable uncontended = { a : 'a; b : 'b }
 [%%expect {|
-type ('a, 'b) t : value mod uncontended portable = { a : 'a; b : 'b; }
+type ('a, 'b) t : immutable_data = { a : 'a; b : 'b; }
```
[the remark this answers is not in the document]
goldfirere: Ah yes, good point. It *is* just printing, because printing outside of alpha-mode drops the with-bounds. I've added a test and a comment.

* [ ] [ce95b1f5e167be9ee427a1effbe57b07c8c9884c](https://github.com/ocaml-flambda/flambda-backend/pull/3284/commits/ce95b1f5e167be9ee427a1effbe57b07c8c9884c) @liam923 to review

testsuite/tests/typing-layouts/allow_illegal_crossing.ml
goldfirere: Yes good point. I thought about this. MAybe I should have changed existing tests? But I think we still have good enough coverage here, so I think I won't do this. If you're ok with this, please Resolve.

typing/jkind.mli
```diff
@@ -522,27 +556,49 @@ type sub_or_intersect =
...
+    output. *)
+val round_up :
+  type_equal:(Types.type_expr -> Types.type_expr -> bool) ->
+  jkind_of_type:(Types.type_expr -> jkind_l option) ->
+  (allowed * 'r) t ->
+  ('l * allowed) t
+
+(** Map a function over types in [upper_bounds] *)
+val map_type_expr :
+  (Types.type_expr -> Types.type_expr) -> (allowed * 'r) t -> (allowed * 'r) t
```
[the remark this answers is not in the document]
goldfirere: I disagree (though not very strongly) here. Only l-kinds have with-bounds. If you're calling this function on an r-kind, you're probably doing something wrong, even though this function can successfully run (by doing nothing) in that scenario. I favor keeping this more restrictive type, but if you push back once more, I'll change it.

typing/jkind.mli
```diff
@@ -522,27 +556,49 @@ type sub_or_intersect =
...
   jkind_l ->
   jkind_l ->
   (jkind_l, Violation.t) result
 
+(** "round up" a [jkind_l] to a [jkind_r] such that the input is less than the
+    output. *)
+val round_up :
+  type_equal:(Types.type_expr -> Types.type_expr -> bool) ->
+  jkind_of_type:(Types.type_expr -> jkind_l option) ->
+  (allowed * 'r) t ->
```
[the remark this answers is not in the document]
goldfirere: See my response to line 600. I think the more restrictive type is more helpful to clients.

typing/jkind.ml
```diff
@@ -1144,47 +1434,112 @@ let of_type_decl ~context (decl : Parsetree.type_declaration) =
...
+    let has_gadt_constructor =
+      List.exists
+        (fun cstr -> match cstr.cd_res with None -> false | Some _ -> true)
+        cstrs
+    in
+    if has_gadt_constructor
+       (* CR layouts v2.8: This is sad, but I don't know how to account for
+          existentials in the baggage. See doc named "Existential
+          baggage". *)
+    then Builtin.value ~why:Boxed_variant
```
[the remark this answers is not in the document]
goldfirere: Yes, but even figuring out whether there are existentials seems not-quite-obvious to me. You'd have to take some set-difference between the free vars of the arguments and of the parameters, or something. Let's just return to this later.

typing/jkind.ml
```diff
@@ -1144,47 +1434,112 @@ let of_type_decl ~context (decl : Parsetree.type_declaration) =
...
+  List.for_all
+    (fun (lbl : Types.label_declaration) -> Sort.Const.(equal void lbl.ld_sort))
+    lbls
+
+let add_labels_as_baggage lbls jkind =
+  List.fold_right
+    (fun (lbl : Types.label_declaration) -> add_baggage ~baggage:lbl.ld_type)
+    lbls jkind
+
+(* CR layouts v2.8: This should take modalities into account. *)
```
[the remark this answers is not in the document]
goldfirere: Moot point, now that @glittershark has implemented support for modalities here. :) But yes I agree with your analysis.

typing/jkind.mli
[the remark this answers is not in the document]
liam923: Fair enough, I'm fine leaving it

typing/ctype.ml
```diff
@@ -2479,25 +2511,39 @@ let rec intersect_type_jkind ~reason env ty1 jkind2 =
...
     (* [intersect_type_jkind] is called rarely, so we don't bother with trying
        to avoid this call as in [constrain_type_jkind] *)
-    let ty1 = get_unboxed_type_approximation env ty1 in
-    Jkind.intersect_l_l ~reason (estimate_type_jkind env ty1) jkind2
+    let type_equal = !type_equal' env in
+    let jkind1 = type_jkind env ty1 in
+    let jkind_of_type = type_jkind_purely_if_principal env in
+    let jkind1 = Jkind.round_up ~type_equal ~jkind_of_type jkind1 in
+    let jkind2 = Jkind.round_up ~type_equal ~jkind_of_type jkind2 in
+    Jkind.intersection_or_error ~type_equal ~jkind_of_type ~reason jkind1 jkind2
```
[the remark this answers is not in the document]
goldfirere: Good alarm bells, but it's actually OK here. I've added a comment and will tag you for review of that comment.

goldfirere:
@liam923 

> As a meta-point, I think someone should review that the review tasks @goldfirere assigned are exhaustive. That is, make sure all changes are actually getting reviewed. I think jkind_intf, for example, did not get assigned anywhere (I have not reviewed that file).

Decent point. Anyone is welcome to double-check this. But I should not: jkind_intf did _not_ change in this patch, so there's no evidence I actually missed anything. :)

typing/jkind.ml
```diff
@@ -1784,92 +2164,151 @@ let combine_histories reason (Pack k1) (Pack k2) =
...
+      { sub with
+        history =
+          combine_histories ~type_equal ~jkind_of_type Subjkind (Pack sub)
+            (Pack super)
+      }
+  in
+  let failure = Error (Violation.of_ (Not_a_subjkind (sub, super))) in
+  (* First try normal subjkinding, if the right-hand jkind has the right
+     shape. *)
+  match try_allow_r super with
```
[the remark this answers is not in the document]
goldfirere: Yes I suppose so. I've refactored.

typing/jkind.ml
```diff
@@ -1784,92 +2164,151 @@ let combine_histories reason (Pack k1) (Pack k2) =
...
+              | Some bound2 ->
+                Misc.Le_result.is_le
+                  (Bound.less_or_equal ~axis ~type_equal ~jkind_of_type bound1
+                     bound2)
+              | None ->
+                let (module Bound_ops) = Axis.get axis in
+                let baggage1 = Baggage.as_list bound1.baggage in
+                let baggage2 = Baggage.as_list bound2.baggage in
+                let modifiers =
+                  Bound_ops.equal bound1.modifier bound2.modifier
```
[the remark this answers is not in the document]
goldfirere: Yes good point.

typing/jkind.ml
```diff
@@ -1784,92 +2164,151 @@ let combine_histories reason (Pack k1) (Pack k2) =
...
+                in
+                let baggages =
+                  List.compare_lengths baggage1 baggage2 = 0
```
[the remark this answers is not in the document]
goldfirere: Not an optimization. We really just want an equality comparison here. I think you mean to use `List.equal`. But type equality is expensive. I'll add a comment.

typing/jkind.ml
```diff
@@ -1784,92 +2164,151 @@ let combine_histories reason (Pack k1) (Pack k2) =
...
 
 (* CR layouts v2.8: Rewrite this to do the hard subjkind check from the
    kind polymorphism design. *)
-let sub_jkind_l sub super =
-  let super = terrible_relax_l super in
-  match check_sub sub super with
-  | Less | Equal ->
-    Ok { sub with history = combine_histories Subjkind (Pack sub) (Pack super) }
-  | Not_le -> Error (Violation.of_ (Not_a_subjkind (sub, super)))
+let sub_jkind_l ~type_equal ~jkind_of_type sub super =
```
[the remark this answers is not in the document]
goldfirere: ok

typing/jkind.ml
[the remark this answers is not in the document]
goldfirere: yep

typing/jkind.ml
```diff
@@ -356,56 +363,332 @@ module Error = struct
...
+              }
+          | Some (fuel, seen_args) ->
+            if List.for_all2 type_equal seen_args args
+            then Skip
+            else if fuel > 0
+            then
+              Continue
+                { t with constr = Path.Map.add p (fuel - 1, args) constr }
+            else Stop)
+        | Tvar _ | Tarrow _ | Tunboxed_tuple _ | Tobject _ | Tfield _ | Tnil
```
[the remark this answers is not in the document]
goldfirere: It doesn't have the same recursion issue. With `-rectypes`, boxed tuples can be recursive. Unboxed tuples cannot. I think this is covered by the existing comment.

typing/typedecl.ml
```diff
@@ -266,7 +267,21 @@ let enter_type ?abstract_abbrevs rec_flag env sdecl (id, uid) =
...
+         that [any] is the default).  Indeed the only reason (I think) we need a
+         non-[any] jkind here is to produce better error messages.
+
+         Doing better here will be annoying, because a type is in scope in its
+         own jkind... and yet we don't have an env that we can use at this
+         point. I think probably the solution will be to have
+         [Jkind.of_type_decl_default] just return [max] every time it sees a
+         [with]-kind... which basically just does this [type_exn] trick but much
+         more sanely. *)
+      ~transl_type:(fun _ -> Predef.type_exn)
```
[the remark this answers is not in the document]
goldfirere: It's a bit awkward and deserves at least a little careful thought, mostly because `of_type_decl_default` is also called in `approx_type_decl`, used in recursive module situations. I'd want some tests with recursive modules before assuming that the `max` behavior here is correct. It's also awkward to implement in jkind.ml. I think we should leave this until later.

typing/typedecl.ml
```diff
@@ -784,8 +799,28 @@ let transl_declaration env sdecl (id, uid) =
     | _ -> false, false (* Not unboxable, mark as boxed *)
   in
   verify_unboxed_attr unboxed_attr sdecl;
+  (* CR layouts v2.8: This next call to [transl_simple_type] probably can loop
+     because it will do perhaps-circular jkind checks. But actually I think the
+     same problem exists in e.g. record fields. We should probably look into this. *)
+  let transl_type sty =
+    (* CR layouts v2.8: The [~new_var_jkind:Any] is weird. The type is closed,
+       and so there shouldn't be any new vars. Investigate. *)
```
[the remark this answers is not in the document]
goldfirere: Not sure -- it's why there's a comment saying to investigate. But this investigation isn't needed now, and so I propose to defer.

typing/typedecl.ml
```diff
@@ -784,8 +799,28 @@ let transl_declaration env sdecl (id, uid) =
     | _ -> false, false (* Not unboxable, mark as boxed *)
   in
   verify_unboxed_attr unboxed_attr sdecl;
+  (* CR layouts v2.8: This next call to [transl_simple_type] probably can loop
+     because it will do perhaps-circular jkind checks. But actually I think the
+     same problem exists in e.g. record fields. We should probably look into this. *)
```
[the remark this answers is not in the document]
goldfirere: I think I disagree. The `Ptyp_constr` case necessarily unifies the supplied arguments with the expected parameters, which includes jkind checks. Maybe it's all OK? But I'd want to take a moment to think hard about this and write test cases.... but not this moment. :)

typing/ctype.ml
goldfirere: Worth it now, in this patch? Or something we can defer until later? If "now", I'll have to learn what `nondep` is all about, something I've hitherto avoided.

typing/ctype.ml
goldfirere: I agree with everything you've written, except your first sentence. Where's the bug? I think the consequence of all of this is that a value's mode-crossing behavior can depend on what type it's assigned (between your `s` and `t`), which is a pretty wiggly notion, and this could lead to hard-to-predict type inference. But I don't see a soundness problem here. Do you?

goldfirere:
* [ ] [5d07d15365b154a1d94379f134bffc1c2609b129](https://github.com/ocaml-flambda/flambda-backend/pull/3284/commits/5d07d15365b154a1d94379f134bffc1c2609b129) Review responses from @liam923 
* [ ] [d87cfbf7fe5fa674965cfd0c8b82e50860e1beae](https://github.com/ocaml-flambda/flambda-backend/pull/3284/commits/d87cfbf7fe5fa674965cfd0c8b82e50860e1beae) Review responses from @riaqn

typing/jkind.ml
```diff
@@ -356,56 +363,332 @@ module Error = struct
...
+        | Tvariant _ | Tunivar _ | Tpackage _ ->
+          (* these cases either cannot be infinitely recursive or their jkinds
```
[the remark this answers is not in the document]
goldfirere: I've commented; see if that's sufficient.

typing/jkind.ml
```diff
@@ -356,56 +363,332 @@ module Error = struct
...
+    *)
+    (* CR reisenberg: document seen_args *)
+    (* CR layouts v2.8: This would seem to eliminate the possibility of
+       mode-crossing for types like rose trees. There's a good chance
+       normalization will fix this. Once we have normalization, let's try this
+       out. *)
+    let module Loop_control = struct
+      type t =
+        { tuple_fuel : int;
+          constr : (int * type_expr list) Path.Map.t
```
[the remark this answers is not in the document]
goldfirere: If we see both (say) `int t` and `bool t`, we just give up. This is really all about just catching regular recursion, with `'a t`.

I do think there's lots of ways this could get improved, but it's likely @glittershark is already on that, and refining things here is probably not worthwhile.

goldfirere:
* [ ] [e0320004f375025d8d7985376a60ce628181fd15](https://github.com/ocaml-flambda/flambda-backend/pull/3284/commits/e0320004f375025d8d7985376a60ce628181fd15) @liam923

goldfirere:
OK. I believe I've responded to all comments up until now. Back to reviewers.

typing/ctype.ml
lpw25: I would probably do it now if it were me, because it is probably only a few minutes work to do. But if you think it will take you a while due to needing to pick up context first then maybe leave a note and come back to it later.

typing/ctype.ml
lpw25: We should be able to get an ill-formed context from this, and those can sometimes, but not always, be turned into a soundness bug using things like `module type of` and `moduel rec M : S = M`.

The ill-formedness probably needs you to have a functor which only accepts modules containing types with certain `with` bounds, then add equations to those types such that this code weakens them so that they no longer obey the requirements of the functor. Then you’ll have e.g. `Foo(M).t` in scope in an environment that doesn’t think `Foo(M).t` is well-typed.

typing/ctype.ml
goldfirere: I've now given myself enough context to understand what's going on here, but I still think it will be ~hours and not ~minutes, accounting for testing, bugs, review, etc. I'm punting, but I've clarified the CR.

typing/ctype.ml
goldfirere: It turns out that the call to `fatal_error` in `add_gadt_equation` falls over in this case. So at least there isn't a soundness problem. Punting for now.

liam923:
I found a valid program that gets rejected by these changes:
```
module rec Gadt_option : sig
  type 'a t = T : 'a option -> 'a t [@@unboxed]
end = Gadt_option

and Also_gadt_option : sig
  type 'a t = 'a Gadt_option.t
end = struct
  type 'a t = 'a Gadt_option.t
end
```

liam923:
As discussed in person, existential unboxed gadts allow a type to escape its scope as the existential will then appear in the with-bounds.

liam923:
> I found a valid program that gets rejected by these changes:
> 
> ```
> module rec Gadt_option : sig
>   type 'a t = T : 'a option -> 'a t [@@unboxed]
> end = Gadt_option
> 
> and Also_gadt_option : sig
>   type 'a t = 'a Gadt_option.t
> end = struct
>   type 'a t = 'a Gadt_option.t
> end
> ```

- [ ] I've resolved this in commits 0ca4ba46964188b99b99a9e094274e2e845214a4 and 1835b5d18e99fa0d3633c8a95963389d3cd1df69. @goldfirere please review.

typing/typedecl.ml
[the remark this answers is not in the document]
riaqn: Sorry, you are right - yes let's do this in the future.

testsuite/tests/typing-jkind-bounds/basics.ml
```diff
@@ -1581,9 +1581,22 @@ type t =
...
   | List of t list
 
-class sexp_of :
-  object
-    method array : ('a -> t) -> ('a array -> t)
-  end
+class type sexp_of =
+    object
+      method array : ('a -> t) -> ('a array -> t)
+    end
```
lpw25, line 1586: Why is the test being changed from a class to an array?

testsuite/tests/typing-jkind-bounds/basics.ml
```diff
@@ -1589,14 +1589,5 @@ class type sexp_of =
...
-Error: Class type declarations do not match:
-         class type sexp_of =
-           object method array : ('a -> t) -> 'a array -> t end
-       does not match
-         class type sexp_of =
-           object method array : ('a -> t) -> 'a array -> t end
-       The method array has type "'a. ('a -> t) -> 'a array -> t"
-       but is expected to have type "'a. ('a -> t) -> 'a array -> t"
-       Type "'a" is not compatible with type "'a0"
+class type sexp_of = object method array : ('a -> t) -> 'a array -> t end
```
lpw25, line 1591: The change in the commit is fine, but do we understand how it fixes the problem? I want to be sure we are missing another place where it should be resetting `univar_pairs`, or nested calls to functions that aren’t reentrant.

lpw25 (commented):
The change in a20a171afec9f1e172f is fine, but I would like to understand the bug it fixes better.

testsuite/tests/typing-jkind-bounds/basics.ml
goldfirere: I went through all uses of `univar_pairs` to use the protect-refs approach on all of them, not just the specific places that caused the bug. I sadly don't now remember which specific one it was. I suppose I could remove the fix and re-debug if you wanted more information.

typing/ctype.ml
```diff
@@ -2180,80 +2226,12 @@ let rec estimate_type_jkind ~expand_component env ty =
   | Tpoly (ty, _) -> estimate_type_jkind ~expand_component env ty
```
liam923, line 2226: As discussed offline, this allows type variables to escape their scope. For now, we can `Jkind.round_up`, and in the future we'll want something like `round_up_to_remove_these_variables`

goldfirere:
About to rebase en route to merging. State before the rebase is preserved at https://github.com/goldfirere/flambda-backend/tree/with-kinds-before-rebase

goldfirere:
Rebased.

glittershark (approved):

liam923:
:tada:

Issue #2421, opened 2024-04-08
Backend refactoring


xclerc:
(this is a list of ideas after reading the code under `backend/`; it has not
been curated / discussed / approved)

(some item below are moot if the Mach IR is deleted from the repository)

Registers:
- distinguish between pseudo-register/temporaries and hardware register (and unify the terminology)
- use a proper types rather than bare `int`s for register classes, stack classes, and register locations
- ensure that the stack location can be compared in a less error-prone way
- make the `Reg` module stateless
- store information about register using maps/tables rather than mutable `Reg.t` fields

Instructions:
- (at the "assembly" level) it is not clear the current encoding is better than one that would for instance have only one constructor for float addition, with a nested value encoding the width of the operation
- (at the "CFG" level) *expression problem*: since we are more often adding new instructions than we are adding new operations, a *record/object* representation might be better
- `Linear.t` should use the instructions from the CFG IR rather than from the Mach IR

CFG:
- to simplify (regalloc) code, we should identify critical edges and pre-emptively add blocks
- `Cfg.t` should hold information like the maximum instruction id (it is currently computed in several places)
- the representation of exception handlers is partially redundant
- the caches in `Cfg_with_infos` need to be manually invalidated
- all dead code should be deleted
- some optimizations use similar code
- `Cfg_equivalence` should be deleted
- `Cfg_loop_infos` should be optimized (and/or some computations made lazy)

Register allocators:
- simplify the logging logic (have only one rather than one per allocator, use indent/dedent)
- extract the construction of the interference graph to a dedicated module
- move the implementation of the priority queue to `utils/`
- move the implementation of the "array set" to `utils/`
- share the handling of ranges/intervals between linscan and greedy
- have another variant of `rewrite`, taking advantage of the fact that a block is never executed partially
- move the iterators over CFG values to `cfg/`
- move the handling of substitutions to a dedicated module

Miscellaneous:
- make `Label.t` abstract
- the list of instructions in `Linear` could re-use the doubly-linked list module
- the two versions of the stack operands are the dual of each other, which make it harder to keep them synchronized
- "finish" the peephole work, *i.e.* have the code generated from a higher-level representation (that can be used to produce data for other tools)
- document the invariants about various IRs
- collect per-function metrics/counters
- the use of objects may have performance implications
- there is quite a bit of code duplication in `Cmm_helpers` and in the binary emitter
- can we use SIMD instructions? - *e.g.* to use bitsets to implement sets
[...]

PR #3286, opened 2024-11-19
Make `Linear` independent of `Mach`

xclerc:
This pull reqesut makes `Linear` independent
of `Mach`, so that the latter is easier to delete.

The main dependency was the `Lop` constructor,
which is "split" by this PR into:

- `Lop` that will share its nested value with CFG
  (through a new module, namely `Operation`);
- `Lcall_op` that will get the constructors from
  the old `Lop` that cannot be mere operation
  in CFG.

The diff is large, but is mainly about moves and
renames. 

I think it is now handled automatically, but the
magic number for linear should definitely be
bumped.

### DIFF
--- a/backend/amd64/emit.mlp
+++ b/backend/amd64/emit.mlp
@@ -25,7 +25,6 @@ open Arch
 open Proc
 open Reg
 open Simple_operation
-open Mach
 open Linear
 open Emitaux
 
--- a/backend/amd64/emit.mlp
+++ b/backend/amd64/emit.mlp
@@ -1325,9 +1325,9 @@ let emit_instr ~first ~fallthrough i =
         cfi_adjust_cfa_offset n;
       end;
     end
-  | Lop(Imove | Ispill | Ireload) ->
+  | Lop(Move | Spill | Reload) ->
       move i.arg.(0) i.res.(0)
-  | Lop(Iconst_int n) ->
+  | Lop(Const_int n) ->
       if n = 0n then begin
         match i.res.(0).loc with
         | Reg _ ->
@@ -1346,7 +1346,7 @@ let emit_instr ~first ~fallthrough i =
           I.mov (nat n) (res i 0)
       end else
         I.mov (nat n) (res i 0)
-  | Lop(Iconst_float32 f) ->
+  | Lop(Const_float32 f) ->
       begin match f with
       | 0x0000_0000l ->       (* +0.0 *)
           I.xorpd (res i 0) (res i 0)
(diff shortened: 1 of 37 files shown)


backend/branch_relaxation.ml
```diff
@@ -63,11 +63,6 @@ module Make (T : Branch_relaxation_intf.S) = struct
         opt_branch_overflows map pc lbl0 max_branch_offset
           || opt_branch_overflows map pc lbl1 max_branch_offset
           || opt_branch_overflows map pc lbl2 max_branch_offset
-      | Lop (Ipoll { return_label = Some lbl }) ->
-        (* A poll-and-branch instruction can branch to the label lbl,
-           but also to an out-of-line code block. *)
-        code_size + max_out_of_line_code_offset - pc >= max_branch_offset
-        || branch_overflows map pc lbl max_branch_offset
```
gretay-js, line 70: `Emit` can still generate these shortcuts, but after this PR Linear cannot represent them. We stopped generating the shortcuts when we removed `Linearize` pass, so this is not changing any existing functionality, but the plan was to recover this optimization on the way out of Cfg. I think these shortcuts were added after some benchmarking upstream, and I hestitate to remove them without a trace. On the other hand, leaving a bunch of dead code in `Emit` is not great. What are you thoughts?

backend/amd64/emit.mlp
```diff
@@ -1762,8 +1753,8 @@ let emit_instr ~first ~fallthrough i =
        intervening wrapper that deals with getting the arguments to the probe
        in the correct place and managing the spilling/reloading of live
        registers.  See [emit_probe_handler_wrapper] below. *)
-    emit_call_probe_handler_wrapper i ~enabled_at_init ~probe_label
-  | Lop (Iprobe_is_enabled { name; }) ->
+                                    emit_call_probe_handler_wrapper i ~enabled_at_init ~probe_label
```
gretay-js, line 1756: extra spaces?

gretay-js (approved):
Not much is left in `cfg_to_linear_desc.ml`, do we still need it to be a separate file?

backend/amd64/emit.mlp
xclerc: Indeed - sorry.

backend/branch_relaxation.ml
gretay-js: Actually, `Emit` cannot generate these shortcuts any more after this PR, so I think a comment about it on https://github.com/ocaml-flambda/flambda-backend/issues/2421 would be enough.

PR #3287, opened 2024-11-19
Delete `Mach`

xclerc:
As per title; this pull request deletes
the *mach* pipeline. Most of the files
touched by this pull request are simply
deleted; only the `Asmgen` module
has a non-trivial diff.


mshinwell:
Are there any command-line flags that should be marked as "(ignored)" or similar in the help text as a result of this PR?

xclerc:
When #3294 and #3295 are in, the diff will
almost only contain deletions. Only the
`Asmgen` module will have a slightly more
complex diff.

xclerc:
(Squashed the commits for future rebases.
Although the pull request is essentially ready,
we do not want to merge it right now. We
will thus have to maintain it for a couple of
months.)

mshinwell (approved):

mshinwell:
OK to merge on Monday 2025-03-10.

mshinwell (approved):

PR #3290, opened 2024-11-19
Don't constrain kinds on Atomic.t

glittershark:
Rather than requiring the type argument on Atomic.t to cross portability, which prevents even /mentioning/ the type (eg if the user is doing unsafe stuff with non-portability-crossing values in atomics), just require all values that are *put* into an atomic to be portable at the call site. This allows eg using `Obj.magic_portable` to do unsafe things. Also, mark return values taken out of an atomic as portable too.

### DIFF
--- a/stdlib/camlinternalOO.ml
+++ b/stdlib/camlinternalOO.ml
@@ -132,7 +132,7 @@ let dummy_table =
       initializers = [];
       size = 0 }
 
-let table_count = Atomic.make 0
+let table_count = Atomic.make_safe 0
 
 (* dummy_met should be a pointer, so use an atom *)
 let dummy_met : item = Obj.magic_portable (of_repr (Obj.new_block 0 0))
@@ -174,8 +174,8 @@ let put array label element =
 
 (**** Classes ****)
 
-let method_count = Atomic.make 0
-let inst_var_count = Atomic.make 0
+let method_count = Atomic.make_safe 0
+let inst_var_count = Atomic.make_safe 0
 
 (* type t *)
 type meth = item
--- a/stdlib/capsule.ml
+++ b/stdlib/capsule.ml
@@ -26,7 +26,7 @@ end = struct
   type 'k t = int
   type packed = P : 'k t -> packed
 
-  let ctr = Atomic.make 0
+  let ctr = Atomic.make_safe 0
   let make () = P (Atomic.fetch_and_add ctr 1)
 
   type (_, _) eq = Eq : ('k, 'k) eq
--- a/stdlib/domain.ml
+++ b/stdlib/domain.ml
@@ -95,7 +95,7 @@ module Runtime_4 = struct
   (******** Callbacks **********)
 
   (* first spawn, domain startup and at exit functionality *)
-  let first_domain_spawned = Atomic.make false
+  let first_domain_spawned = Atomic.make_safe false
 
   let first_spawn_function = ref (fun () -> ())
 
@@ -225,7 +225,7 @@ module Runtime_5 = struct
       let to_capsule_password t = t
     end
 
-    let key_counter = Atomic.make 0
+    let key_counter = Atomic.make_safe 0
     let password_idx = Atomic.fetch_and_add key_counter 1
 
     let create_dls password =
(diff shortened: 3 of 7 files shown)


stdlib/atomic.mli
```diff
@@ -43,24 +43,23 @@ val make : 'a @ contended -> 'a t
...
-val make_contended : 'a @ contended -> 'a t
+val make_contended : 'a @ portable contended -> 'a t
 
 (** Get the current value of the atomic reference. *)
-val get_safe : 'a t -> 'a @ contended
+val get_safe : 'a t -> 'a @ portable contended
 
 (** Set a new value for the atomic reference. *)
-val set : 'a t -> 'a @ contended -> unit
+val set : 'a t -> 'a @ portable contended -> unit
```
tdelvecchio-jsc, line 52: Our current goal is that our version of Stdlib is a valid implementation of the upstream Stdlib. Adding `portable` here makes `set` a weaker definition than in upstream's. We should add `set_safe` and alert `set`.

stdlib/atomic.mli
```diff
@@ -43,24 +43,23 @@ val make : 'a @ contended -> 'a t
...
 
 (** Set a new value for the atomic reference, and return the current value. *)
-val exchange_safe : 'a t -> 'a @ contended -> 'a @ contended
+val exchange_safe : 'a t -> 'a @ portable contended -> 'a @ portable contended
```
tdelvecchio-jsc, line 55: ditto

stdlib/atomic.mli
```diff
@@ -43,24 +43,23 @@ val make : 'a @ contended -> 'a t
...
 
 (** [compare_and_set r seen v] sets the new value of [r] to [v] only
     if its current value is physically equal to [seen] -- the
     comparison and the set occur atomically. Returns [true] if the
     comparison succeeded (so the set happened) and [false]
     otherwise. *)
-val compare_and_set : 'a t -> 'a @ contended -> 'a @ contended -> bool
+val compare_and_set : 'a t -> 'a @ portable contended -> 'a @ portable contended -> bool
```
tdelvecchio-jsc, line 62: ditto

tdelvecchio-jsc (changes requested):

stdlib/atomic.mli
glittershark: this is the `_safe` variant (unless I'm missing something?

tdelvecchio-jsc: im silly

tdelvecchio-jsc:
Justification for why this is a good change: the new signature is strictly stronger than the previous one.
```
module type Atomic1 = sig
  type (!'a : value mod portable) t : value mod portable uncontended

  val make : 'a @ contended -> 'a t
  val get : 'a t -> 'a @ contended
  val set : 'a t -> 'a @ contended -> unit
  val exchange : 'a t -> 'a @ contended -> 'a @ contended
  val compare_and_set : 'a t -> 'a @ contended -> 'a @ contended -> bool
end

module type Atomic2 = sig
  type !'a t : value mod portable uncontended

  val make : 'a @ contended portable -> 'a t
  val get : 'a t -> 'a @ contended portable
  val set : 'a t -> 'a @ contended portable -> unit
  val exchange : 'a t -> 'a @ contended portable -> 'a @ contended portable
  val compare_and_set : 'a t -> 'a @ contended portable -> 'a @ contended portable -> bool
end

module F (A : Atomic2) : Atomic1 = struct
  type (!'a : value mod portable) t = 'a A.t

  let make a = A.make a
  let get t = A.get t
  let set t a = A.set t a
  let exchange t a = A.exchange t a
  let compare_and_set t a1 a2 = A.compare_and_set t a1 a2
end
```

stdlib/atomic.mli
glittershark: done

stdlib/atomic.mli
```diff
@@ -75,12 +74,25 @@ end @@ portable
...
 
-val exchange : 'a t -> 'a -> 'a
+val set : 'a t -> 'a @ contended -> unit
+[@@alert unsafe]
+
+val exchange : 'a t -> 'a @ contended -> 'a
 [@@alert unsafe]
 
+val compare_and_set : 'a t -> 'a @ contended -> 'a @ contended -> bool
+[@@alert unsafe]
```
tdelvecchio-jsc, line 93: Nit: can we remove the mode annotations from the unsafe ones? Stronger discouragement to use them in places where you actually care about modes.

tdelvecchio-jsc (approved):

PR #3291, opened 2024-11-20
Fix the kind of poison constants generated for unboxed value slots

lthls:
When unboxing a closure under a variant, the branches corresponding to the other tags get a poison constant for the unboxed values slots. The constant must be of the right kind.

### DIFF
--- a/middle_end/flambda2/simplify/unboxing/unboxers.ml
+++ b/middle_end/flambda2/simplify/unboxing/unboxers.ml
@@ -181,7 +181,10 @@ module Closure_field = struct
 
   let unboxer function_slot value_slot =
     { var_name = "closure_field_at_use";
-      poison_const = Const.const_zero;
+      poison_const =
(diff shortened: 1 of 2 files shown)


mshinwell (approved):

Gbury:
A quick question actually: so this bug appeared with the addition of non-value-kind value slots, where the kind was incorrect, which this PR fixes. That being said, is, now, or in the future, `0` always a correct value of any kind (including non-value kinds) ? of should we try and have a function `kind -> const` to generate poison values ?

lthls:
So far all the kinds that can end up unboxed allow numeric values, so `Const.of_int_of_kind` works. We should probably document that this function should only be used for dummy constants (the `Vec128` case in particular doesn't actually encode the given integer), and if we add kinds that don't allow numbers we can reinterpret it as "give me a constant of this kind taking this number as inspiration" and return an arbitrary constant of the right kind.

PR #3302, opened 2024-11-22
Add replay history to match names between two downwards traversals

Gbury:
[Note: This is a pre-requisite of match-in-match; the mechanism introduced is not yet used in this PR, but will be used by continuation specialization when the PR for match-in-match is opened]

In some cases (mostly the upcoming continuation specialization for match-in-match), we are interested in doing multiple downwards pass on a single term. In such cases, it might be necessary to relate the names of variables and continuations from the first pass to those from the subsequent passes. The predominant use of that for now should be to correctly handle continuations that have been lifted during the first pass, and whose calls should be rewritten in subsequent passes (including their lifted arguments).

This is done by storing the sequence of names generated when opening name abstractions during the first downwards pass. On subsequent passes, each new variable and continuation "consumes" an element of that sequence, allowing to establish the correspondance between names. As a safety measure, we check that correspondings names are in the same "renaming equivalence class", though note that this is not a guarantee (for isntance, by default all continuation have no names and therefore will all be in the same "renaming equivalence class").

Additionally, and this is more specific to continuation specialization but generalizable to other future uses, it is necessary to have some kind of replayability of inlining decisions. Indeed each inlining decision will change the sequence of bound names that are opened during the downwards pass, which would currently break the hypothesis of replay histories that exactly the same sequence of binders are opened during both downwards passes. In the case of match-in-match this is simple: the handler that we want to specialize ends with a switch, which means that any call inside the handler have been inlined, so the replay history has a boolean to denote that we want to inline everything while replaying the handler downwards pass. For other more complex uses, inlining decisions could be stored alongside the bound variables and continuations, so that they can be replayed, either as is, or within some notion of compatibility.

Finally, we also need to make it so that recursive continuations are bound in an order that is stable through renaming, so we change the map of continuation to an Lmap for recursive continuation bindings.

In the future, this replay mechanism would also be useful to do widening for recursive continuations (but that's a far away future).

### DIFF
--- a/middle_end/flambda2/simplify/env/replay_history.mli
+++ b/middle_end/flambda2/simplify/env/replay_history.mli
@@ -72,14 +72,14 @@ val define_continuations : Continuation.t list -> t -> t
 (** {2 Inspection API} *)
 
 (** Returns [true] is the replay history was created with the `always_inline` flag. (first pass
-    histories have the flage set to false). *)
+    histories have the flag set to false). *)
 val must_inline : t -> bool
 
 (** Type of results for inspection functions for which the result only makes sense
     when replaying a downwards pass. *)
 type 'a replay_result =
   | Still_recording
-  | Replayed of 'a (**)
+  | Replayed of 'a
 
 (** If called on a history created with {!replay}, returns the mapping from variables
     of the current pass to variables bound during the first pass.
--- a/middle_end/flambda2/simplify/env/replay_history.mli
+++ b/middle_end/flambda2/simplify/env/replay_history.mli
@@ -1,0 +1,95 @@
+(**************************************************************************)
+(*                                                                        *)
+(*                                 OCaml                                  *)
+(*                                                                        *)
+(*   Guillaume Bury, Pierre Chambart and Nathanaëlle Courant, OCamlPro    *)
+(*                                                                        *)
+(*   Copyright 2013--2024 OCamlPro SAS                                    *)
+(*   Copyright 2014--2024 Jane Street Group LLC                           *)
+(*                                                                        *)
+(*   All rights reserved.  This file is distributed under the terms of    *)
+(*   the GNU Lesser General Public License version 2.1, with the          *)
+(*   special exception on linking described in the file LICENSE.          *)
+(*                                                                        *)
+(**************************************************************************)
+
+(** {1 Replay Histories}
+
+    In some cases (e.g. continuation specialization for match-in-match), we are
+    interested in doing multiple downwards pass on a single term. In such
+    cases, it might be necessary to relate the names of variables and
+    continuations from the first pass to those from the subsequent passes. The
+    predominant use of that is currently to correctly handle continuations that
+    have been lifted during the first pass, and whose calls should be rewritten
+    in subsequent passes (including their lifted arguments).
+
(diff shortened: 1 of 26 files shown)


Ekdohibs (approved):

Gbury:
This has been rebased and is ready for review, ^^

middle_end/flambda2/simplify/env/replay_history.mli
```diff
@@ -0,0 +1,95 @@
...
+(** Define a new bound variable. *)
+val define_variable : Variable.t -> t -> t
+
+(** Define a new set of bound mutually recursive continuations. *)
+val define_continuations : Continuation.t list -> t -> t
+
+(** {2 Inspection API} *)
+
+(** Returns [true] is the replay history was created with the `always_inline` flag. (first pass
+    histories have the flage set to false). *)
```
Ekdohibs, line 75:
```suggestion
    histories have the flag set to false). *)
```

middle_end/flambda2/simplify/env/replay_history.mli
```diff
@@ -0,0 +1,95 @@
...
+val must_inline : t -> bool
+
+(** Type of results for inspection functions for which the result only makes sense
+    when replaying a downwards pass. *)
+type 'a replay_result =
+  | Still_recording
+  | Replayed of 'a (**)
```
Ekdohibs, line 82:
```suggestion
  | Replayed of 'a
```

middle_end/flambda2/simplify/env/replay_history.mli
```diff
@@ -0,0 +1,95 @@
...
+
+(** If called on a history created with {!replay}, returns the mapping from variables
+    of the current pass to variables bound during the first pass.
+
+    On a first pass replay, this will raise a fatal error. *)
```
Ekdohibs, line 87: Here and for the following function, this comment is incorrect as it returns `Still_recording`.

middle_end/flambda2/simplify_shared/call_site_inlining_decision_type.ml
```diff
@@ -164,6 +169,13 @@ let report_reason fmt t =
     Format.fprintf fmt "the@ call@ has@ an@ attribute@ forbidding@ inlining"
   | Attribute_always ->
     Format.fprintf fmt "the@ call@ has@ an@ [@@inline always]@ attribute"
+  | Replay_history_says_must_inline ->
+    (* CR gbury: We should probably not include in the inlining report inlining
+       decisions that were made during replays (e.G. continuation
+       specialization). *)
```
Ekdohibs, line 175: If we keep the inlining decisions themselves in the replay history, we wouldn't need a special reason here. That could remain for later, however.

Ekdohibs (approved):
Mostly good, see comments.

middle_end/flambda2/simplify_shared/call_site_inlining_decision_type.ml
Gbury: Comment updated to reflect that possibility.

middle_end/flambda2/simplify/env/replay_history.mli
Gbury: I removed this last sentence for both comments.

PR #3304, opened 2024-11-22
Fix thread stack sizes on runtime5

mshinwell:
A mistake with bytes and words means that thread stacks on runtime5 are currently eight times as large as intended...


stedolan:
`Wsize_bsize(size_in_bytes)` is the runtime idiom, rather than division by sizeof uintnat.

mshinwell:
Fixed

stedolan (approved):

PR #3305, opened 2024-11-23
arm64: add support for clz and ctz intrinsics

matthewelse:
As per the title, add support for `clz` and `ctz` intrinsics on arm64.

I held off on using the `ctz` instruction, since it's new enough that most ARMv8 CPUs don't support it. It would be easy enough to add some notion of extensions, similar to amd64/arch.ml, but given that ~no-one would be able to use it, I think it's probably fine to leave as is.

(I managed to build and run tests for ocaml_intrinsics with this patch with a bit of tweaking, and confirmed that the code gen for clz/ctz looks good)


matthewelse:
Any idea when you might have time to review this? Anything I can do to help the process?

gretay-js (approved):

gretay-js:
> Hi, any idea when you might have time to review this? Anything I can do to help the process?

sorry. approved now, but it needs a rebase please.

PR #3309, opened 2024-11-25
Peek and poke

mshinwell:
These are unsafe, but the idea is for them to be used via a specific library, if I understand correctly.

```
type ('a : any) t = nativeint#

external read : ('a : any). 'a t -> 'a = "%peek"
  [@@layout_poly]

external write : ('a : any). 'a t -> 'a -> unit = "%poke"
  [@@layout_poly]
```

cc @ccasin for an initial look (some frontend fixes required in here).

### DIFF
--- a/lambda/translprim.ml
+++ b/lambda/translprim.ml
@@ -28,6 +28,7 @@ module String = Misc.Stdlib.String
 type error =
   | Unknown_builtin_primitive of string
   | Wrong_arity_builtin_primitive of string
+  | Wrong_layout_for_peek_or_poke of string
   | Invalid_floatarray_glb
   | Product_iarrays_unsupported
   | Invalid_array_kind_for_uninitialized_makearray_dynamic
@@ -123,6 +124,10 @@ type prim =
   | Identity
   | Apply of Lambda.region_close * Lambda.layout
   | Revapply of Lambda.region_close * Lambda.layout
+  | Peek of Lambda.peek_or_poke option
+  | Poke of Lambda.peek_or_poke option
+    (* For [Peek] and [Poke] the [option] is [None] until the primitive
+       specialization code (below) has been run. *)
   | Unsupported of Lambda.primitive
 
 let units_with_used_primitives = Hashtbl.create 7
@@ -916,6 +921,8 @@ let lookup_primitive loc ~poly_mode ~poly_sort pos p =
       Primitive(Preinterpret_tagged_int63_as_unboxed_int64, 1)
     | "%reinterpret_unboxed_int64_as_tagged_int63" ->
       Primitive(Preinterpret_unboxed_int64_as_tagged_int63, 1)
+    | "%peek" -> Peek None
+    | "%poke" -> Poke None
     | s when String.length s > 0 && s.[0] = '%' ->
       (match String.Map.find_opt s indexing_primitives with
        | Some prim -> prim ~mode
@@ -1185,6 +1192,27 @@ let glb_array_set_type loc t1 t2 =
   (* Pfloatarray is a minimum *)
   | Pfloatarray_set, Pfloatarray -> Pfloatarray_set
 
+let peek_or_poke_layout_from_type ~prim_name error_loc env ty
+      : Lambda.peek_or_poke option =
+  match Ctype.type_sort ~why:Peek_or_poke ~fixed:true env ty with
+  | Error _ -> None
+  | Ok sort ->
+    let sort = Jkind.Sort.default_to_value_and_get sort in
+    let layout = Typeopt.layout env error_loc sort ty in
+    match layout with
+    | Punboxed_float Unboxed_float32 -> Some Ppp_unboxed_float32
+    | Punboxed_float Unboxed_float64 -> Some Ppp_unboxed_float
+    | Punboxed_int Unboxed_int32 -> Some Ppp_unboxed_int32
+    | Punboxed_int Unboxed_int64 -> Some Ppp_unboxed_int64
+    | Punboxed_int Unboxed_nativeint -> Some Ppp_unboxed_nativeint
+    | Pvalue { raw_kind = Pintval ; _ } -> Some Ppp_tagged_immediate
+    | Ptop
+    | Pvalue _
+    | Punboxed_vector _
+    | Punboxed_product _
+    | Pbottom ->
+      raise (Error (error_loc, Wrong_layout_for_peek_or_poke prim_name))
+
 (* Specialize a primitive from available type information. *)
 (* CR layouts v7: This function had a loc argument added just to support the void
    check error message.  Take it out when we remove that. *)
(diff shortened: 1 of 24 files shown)


lpw25:
We should have a `mod external` bound on the type of `write` to make it generally save, but IIUC Mark got an error when trying to add it. @goldfirere is that supposed to work yet?

goldfirere:
Well, it's spelled `mod external_`, but yes it should work.

mshinwell:
@TheNumbat maybe you could review this?  (I've asked @ccasin if he can fix up the frontend parts.)

stedolan:
`mod external_` alone doesn't make this safe - there's no typechecking here since `'a t = 'b t`. I'd prefer to type these as:
```
external read : ('a : any mod external_). nativeint# -> 'a = "%peek"
  [@@layout_poly]

external write : ('a : any mod external_). nativeint# -> 'a -> unit = "%poke"
  [@@layout_poly]
```
so at least we're not pretending there's some checking of `'a` here.

lpw25:
I agree that the current signature is not useful. I'd probably slightly prefer to just remove the equality of `'a t` to `nativeint#` though.

middle_end/flambda2/terms/flambda_primitive.ml
```diff
@@ -1767,6 +1784,7 @@ let result_kind_of_binary_primitive p : result_kind =
   | Phys_equal _ | Int_comp _ | Float_comp _ -> Singleton K.naked_immediate
   | Bigarray_get_alignment _ -> Singleton K.naked_immediate
   | Atomic_exchange | Atomic_fetch_and_add -> Singleton K.value
+  | Poke _ -> Singleton K.value
```
TheNumbat, line 1787: `Unit`?

middle_end/flambda2/terms/flambda_primitive.ml
```diff
@@ -1450,7 +1457,7 @@ let effects_and_coeffects_of_unary_primitive p : Effects_and_coeffects.t =
       Has_coeffects,
       Strict )
   | Get_header -> No_effects, No_coeffects, Strict
-  | Atomic_load _ -> Arbitrary_effects, Has_coeffects, Strict
+  | Atomic_load _ | Peek _ -> Arbitrary_effects, Has_coeffects, Strict
```
TheNumbat, line 1460: Why does peek need arbitrary effects?

middle_end/flambda2/to_cmm/to_cmm_primitive.ml
```diff
@@ -933,6 +939,13 @@ let binary_primitive env dbg f x y =
   | Bigarray_get_alignment align -> C.bigstring_get_alignment x y align dbg
   | Atomic_exchange -> C.atomic_exchange ~dbg x y
   | Atomic_fetch_and_add -> C.atomic_fetch_and_add ~dbg x y
+  | Poke kind ->
+    let memory_chunk =
+      K.Standard_int_or_float.to_kind_with_subkind kind
+      |> C.memory_chunk_of_kind
+    in
+    C.store ~dbg memory_chunk Assignment ~addr:x ~new_value:y
+    |> C.return_unit dbg
```
TheNumbat, line 948: Will the `return_unit` always be optimized out? If not it seems worthwhile to make poke return nothing (or unboxed unit) (or something like an uninitialized unboxed nativeint)

test_pp.ml
```diff
@@ -0,0 +1,19 @@
+type ('a : any) t = nativeint#
```
TheNumbat, line 1: I agree the real API shouldn't expose the equality with `nativeint#`

typing/primitive.ml
```diff
@@ -655,6 +655,20 @@ let prim_has_valid_reprs ~loc prim =
         is (Same_as_ocaml_repr C.value);
       ]
 
+    | "%peek" ->
+      (* XXX mshinwell: unsure why this doesn't work
```
TheNumbat, line 659: @ccasin ?

lambda/translprim.ml
```diff
@@ -887,6 +891,8 @@ let lookup_primitive loc ~poly_mode ~poly_sort pos p =
       Primitive(Preinterpret_tagged_int63_as_unboxed_int64, 1)
     | "%reinterpret_unboxed_int64_as_tagged_int63" ->
       Primitive(Preinterpret_unboxed_int64_as_tagged_int63, 1)
+    | "%peek" -> Peek None
```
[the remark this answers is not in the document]
lpw25: I think we normally don't include `unsafe` in primitive names, since they are all unsafe with the wrong type and can all be safe with the right type.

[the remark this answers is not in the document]
TheNumbat: True, but these are at least as unsafe as the unsafe array accessors, which do include unsafe or 'u' in their names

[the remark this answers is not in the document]
mshinwell: Yeah, but there are two options there, whereas these are always unsafe.  I think in the interests of getting this merged soon I'm going to leave the names as-is.

lambda/translprim.ml
```diff
@@ -1555,6 +1603,12 @@ let lambda_of_prim prim_name prim loc args arg_exps =
         ap_region_close = pos;
         ap_mode = alloc_heap;
       }
+  | Peek None, _ | Poke None, _ ->
+      Misc.fatal_error "Peek or Poke primitive without layout"
```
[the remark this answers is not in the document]
mshinwell: Yes

middle_end/flambda2/to_cmm/to_cmm_primitive.ml
mshinwell: I think we always need `return_unit` after stores, in case the result is bound to a variable, which is then expected to hold the unit value.  We've had bugs here before.

middle_end/flambda2/terms/flambda_primitive.ml
mshinwell: I think it's probably best if we don't allow it to be moved around, at least in the initial instance.  I've added a comment.

typing/primitive.ml
mshinwell: I don't think any check is needed here, `Translprim` should do enough.

testsuite/tests/peek_and_poke/peek_and_poke.ml
```diff
@@ -0,0 +1,85 @@
...
+external calloc :
+  count:(int[@untagged]) -> size:(int[@untagged]) -> nativeint# =
+  "caml_no_bytecode_impl" "calloc"
+  [@@noalloc]
+
+let test_read p1 p2 p3 p4 p5 p6
+  : #(int * int32# * int64# * nativeint# * float# * float#) =
+  #(read p1, read p2, read p3, read p4, read p5, read p6)
+
+(* CR mshinwell for gyorsh: the last number here should be #0.1234s *)
```
TheNumbat, line 54: Does this not work yet?

PR #3317, opened 2024-11-26
Implement %makearray_dynamic{,_uninit}

mshinwell:
This implements two primitives (exact signature of the second yet to be decided):
```
external[@layout_poly] makearray_dynamic : ('a : any). len:int -> init:('a [@local_opt]) -> ('a array [@local_opt]) =
  "%makearray_dynamic"

external[@layout_poly] makearray_dynamic_uninit : ('a : <TBD>). int -> ('a array [@local_opt]) =
  "%makearray_dynamic_uninit"
```
These provide array creation at runtime either on the heap or the local stack.  All layouts, including those involving unboxed products, are supported.  The arrays may be either initialized or uninitialized.  Any excess 32-bit word at the end of an initialized array (for arrays of `float32#` and `int32#`) is deterministically initialized to zero.

Uninitialized arrays may only be created if the layout completely prohibits valid OCaml values (which includes things of type `int` and equivalent, even if they occur in a context such as `#(int * int32#)` where the resulting block will be non-scannable).  This ensures soundness in the case where such array is inadvertently read before being initialized.  (Consider for example the case where the uninitialized data fails to have the bottom bit set.)

Arrays whose array kinds involve GC-ignorable unboxed products are always allocated as mixed blocks, even in cases such as `#(int32# * int32#)`, and 32-bit elements are never packed therein.  This means that the representation currently differs from e.g. `int32# array`, which uses a custom block, with packing.  As has been discussed, we intend to move everything to mixed blocks in the future, and there will no doubt be improvements to packing for unboxed products.

We should discuss whether it is undesirable for the representation of all-`int` arrays (which will end up as mixed blocks in this PR) to be different from e.g. that obtained from array literals, for the purposes of the polymorphic operations.

(This PR currently includes a test commit from https://github.com/ccasin/flambda-backend/commit/a55f456a2d1688848f9565621e35dca696ceeebb)


runtime/array.c
```diff
@@ -469,6 +488,139 @@ CAMLprim value caml_make_local_vect(value len, value init)
...
+    int move_init_to_major = 0;
+    for (mlsize_t i = 0; i < non_unarized_length; i++) {
+      if (Is_block(Field(v_init, i)) && Is_young(Field(v_init, i))) {
+        move_init_to_major = 1;
+      }
+    }
+    if (move_init_to_major) {
+      /* We don't want to create so many major-to-minor references,
+         so the contents of [v_init] are moved to the major heap by doing
+         a minor GC. */
```
TheNumbat, line 603: Why is this better than adding them to the remembered set with caml_initialize?

runtime/array.c
```diff
@@ -469,6 +488,139 @@ CAMLprim value caml_make_local_vect(value len, value init)
...
+      CAML_EV_COUNTER (EV_C_FORCE_MINOR_MAKE_VECT, 1);
+      caml_minor_collection ();
+    }
+    for (mlsize_t i = 0; i < non_unarized_length; i++) {
```
TheNumbat, line 607: `#ifdef DEBUG` ?

runtime/array.c
```diff
@@ -492,57 +644,111 @@ CAMLprim value caml_make_float_vect(value len)
...
+{
+  return caml_make_unboxed_nativeint_vect0(len, 1);
 }
 
 CAMLprim value caml_make_unboxed_nativeint_vect_bytecode(value len)
 {
   return caml_make_vect(len, caml_copy_nativeint(0));
 }
 
+CAMLprim value caml_make_unboxed_nativeint_vect_and_initialize(value len,
```
TheNumbat, line 742: Is this function used anywhere?

runtime/float32.c
```diff
@@ -863,8 +863,23 @@ CAMLprim value caml_make_unboxed_float32_vect(value len)
...
+  else
+    return caml_alloc_custom(ops, num_fields * sizeof(value), 0, 0);
+}
+
+CAMLprim value caml_make_unboxed_float32_vect(value len)
+{
+  return caml_make_unboxed_float32_vect0(len, 0);
+}
+
+CAMLprim value caml_make_unboxed_float32_vect_local(value len)
```
TheNumbat, line 880: `caml_make_local_unboxed_float32_vect`

runtime/simd.c
```diff
@@ -75,20 +75,37 @@ CAMLprim value caml_unboxed_vec128_vect_blit(value a1, value ofs1, value a2,
...
 
-    return caml_alloc_custom(&caml_unboxed_vec128_array_ops, num_fields * sizeof(value), 0, 0);
+CAMLprim value caml_make_local_unboxed_vec128_vect(value len)
+{
+  return caml_make_unboxed_vec128_vect0(len, 1);
 }
 
 CAMLprim value caml_make_unboxed_vec128_vect_bytecode(value len) {
-    caml_failwith("SIMD is not supported in bytecode mode.");
+  caml_failwith("SIMD is not supported in bytecode mode.");
```
TheNumbat, line 108: The other errors are `not supported on this platform`, might as well fix this one

testsuite/tests/typing-layouts/layout_poly.ml
```diff
@@ -721,19 +718,15 @@ Error: "[@layout_poly]" on this external declaration has no
 external[@layout_poly] makearray_dynamic : ('a : any). int -> 'a -> 'a array =
   "%makearray_dynamic"
 [%%expect{|
-Lines 1-2, characters 0-22:
-1 | external[@layout_poly] makearray_dynamic : ('a : any). int -> 'a -> 'a array =
-2 |   "%makearray_dynamic"
-Error: This construct requires the alpha version of the extension "layouts", which is disabled and cannot be used
+external makearray_dynamic : ('a : any). int -> 'a -> 'a array
```
TheNumbat, line 721: Could use a test that `%makearray_dynamic_uninit` requires alpha

typing/typeopt.ml
```diff
@@ -244,6 +244,14 @@ let array_kind_of_elt ~elt_sort env loc ty =
   | Unboxed_vector v -> Punboxedvectorarray v
   | Product c -> c
 
+let array_kind_of_array_type env loc ty =
```
TheNumbat, line 247: Can this be replaced/unified with `array_type_kind`? They do nearly the same thing

TheNumbat (commented):
Minor comments. I think it's fine that all-int unboxed tuple arrays end up with the mixed block representation, though it is a bit strange that this applies to singleton tuples too.

TheNumbat:
(Also as Chris noted, the tests are currently only running in bytecode)

mshinwell:
Note that the runtime4 CI tests are not yet expected to pass, I'll port the runtime5 changes to runtime4 once review has been completed.

middle_end/flambda2/from_lambda/lambda_to_lambda_transforms.ml
```diff
@@ -110,7 +110,321 @@ let rec_catch_for_for_loop env loc ident start stop
...
+  let external_call_desc =
+    Primitive.make ~name ~alloc:true (* the C stub may raise an exception *)
+      ~c_builtin:false ~effects:Arbitrary_effects ~coeffects:Has_coeffects
+      ~native_name:name
+      ~native_repr_args:
+        ([Primitive.Prim_global, L.Same_as_ocaml_repr (Base Value)]
+        @
+        match init with
+        | None -> []
+        | Some (init_extern_repr, _) -> [Primitive.Prim_global, init_extern_repr]
```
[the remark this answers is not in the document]
mshinwell: I don't think these values are used from this point onwards, but I'll set this to Alloc_local anyway.

lambda/lambda.mli
```diff
@@ -179,6 +179,9 @@ type primitive =
   (* Array operations *)
   | Pmakearray of array_kind * mutable_flag * locality_mode
   | Pmakearray_dynamic of array_kind * locality_mode
+  (** For [Pmakearray_dynamic], if the array kind specifies an unboxed
+      product, the float array optimization will never apply -- even if such
+      unboxed product is a singleton. *)
```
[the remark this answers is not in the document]
mshinwell: Actually, this case might be ruled out by the frontend.  Let's discuss with @ccasin.

lambda/translprim.ml
```diff
@@ -1228,18 +1231,38 @@ let specialize_primitive env loc ty ~has_constant_constructor prim =
...
+      else Some (Primitive (Pmakearray_dynamic (array_type, mode), 2))
+    end
+  | Primitive (Pmakearray_dynamic (at, mode), 1), _ :: [] -> begin
+      let loc = to_location loc in
+      match is_function_type env ty with
+      | None -> None
+      | Some (_, array_type) ->
+        match array_kind_of_array_type env loc array_type with
+        | None ->
+          (* XXX mshinwell: should this produce an error? *)
```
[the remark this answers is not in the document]
mshinwell: @ccasin could you please fix this one?

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
```diff
@@ -1922,6 +1924,8 @@ let convert_lprim ~big_endian (prim : L.primitive) (args : Simple.t list list)
     [Binary (Int_arith (I.Tagged_immediate, Div), arg1, arg2)]
   | Pdivint Safe, [[arg1]; [arg2]] ->
     [checked_arith_op ~dbg None Div None arg1 arg2 ~current_region]
+  | Pmodint Unsafe, [[arg1]; [arg2]] ->
```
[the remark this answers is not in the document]
mshinwell: Just wasn't needed.

typing/typeopt.ml
mshinwell: For @ccasin

runtime/array.c
mshinwell: Ah probably not any more.

runtime/array.c
mshinwell: Not sure, but I'm just following the existing caml_make_vect code.

TheNumbat: I assume it's because caml_initializing would create a reference for each array index instead of each element field. It sounds much better to explicitly remember each field instead of forcing a minor GC (and domain barrier), though, and same in caml_make_vect

TheNumbat: Though with safepoints the barrier is never supposed to be expensive, so maybe this is fine. (Not sure what our current plan is regarding them)

middle_end/flambda2/from_lambda/lambda_to_lambda_transforms.ml
```diff
@@ -228,7 +228,7 @@ let makearray_dynamic_unboxed_product_c_stub ~name (mode : L.locality_mode) =
     ~c_builtin:false ~effects:Arbitrary_effects ~coeffects:Has_coeffects
     ~native_name:name
     ~native_repr_args:
-      [ Prim_global, L.Same_as_ocaml_repr (Base Value);
+      [ Prim_local, L.Same_as_ocaml_repr (Base Value);
         Prim_global, L.Same_as_ocaml_repr (Base Value);
```
TheNumbat, line 232: The first argument is the length; the second is the initializer

middle_end/flambda2/from_lambda/lambda_to_lambda_transforms.ml
```diff
@@ -110,7 +110,417 @@ let rec_catch_for_for_loop env loc ident start stop
...
+      | Pintarray ->
+        Misc.fatal_errorf
+          "Cannot compile Pmakearray_dynamic at layout Pintarray without an \
+           initializer; otherwise it might be possible for values of type \
+           [int] having incorrect representations to be revealed, thus \
+           breaking soundness:@ %a"
+          Debuginfo.print_compact dbg
+      | Pgenarray | Paddrarray | Pfloatarray | Punboxedfloatarray _
+      | Punboxedintarray _ | Punboxedvectorarray _ | Pgcscannableproductarray _
+      | Pgcignorableproductarray _ ->
```
TheNumbat, line 377: `Pgcignorableproductarray` involving an int should also have a similar error to above

lambda/translprim.ml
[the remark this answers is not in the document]
ccasin: @mshinwell Not at a computer, but I think if you check your slack messages from earlier this week I sent you commits or instructions for how to resolve each xxxs

middle_end/flambda2/from_lambda/lambda_to_lambda_transforms.ml
```diff
@@ -110,7 +110,321 @@ let rec_catch_for_for_loop env loc ident start stop
...
+type initialize_array_element_width =
+  | Thirty_two of { zero_init : L.lambda }
+  | Sixty_four_or_more
+
+let initialize_array0 env loc ~length array_set_kind width ~(init : L.lambda)
+    creation_expr =
+  let array = Ident.create_local "array" in
+  (* If the element size is 32-bit, zero-initialize the last 64-bit word, to
+     ensure reproducibility. *)
+  (* CR mshinwell: why does e.g. caml_make_unboxed_int32_vect not do this? *)
```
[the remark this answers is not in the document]
mshinwell: Hmm indeed the stdlib `Marshal` is forbidden, but I still think this is probably a good idea.  It doesn't seem that unlikely that someone might want to marshal these in a different way; and maybe the forthcoming reinterpret casts could end up reading the last 32 bits?

testsuite/tests/typing-layouts/layout_poly.ml
mshinwell: I'm not quite sure what to do here now that we've moved these to beta - @ccasin can you confirm?  (See lines 715 onwards of `typing-layouts/layout_poly.ml`)

typing/typeopt.ml
mshinwell: I've removed the above function, but @ccasin should check `translprim.ml` around lines 1246 and 1260.

runtime/array.c
mshinwell: I've added a CR, we can discuss separately, since this matches existing well-tested code.

TheNumbat (approved):
Approving modulo discussion with Chris

lambda/lambda.mli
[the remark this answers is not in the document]
ccasin: This is ruled out by the front-end (and the array specialization code in typeopt).  Note the comment on `array_kind` in this file documenting the invariant that product array kinds have at least two elements.

It's still a little weird that `#(int32# * int32#) array` isn't packed tightly like `int32# array`, but I'm content to wait for other work on nesting before attempting that.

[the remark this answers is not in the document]
mshinwell: We can probably deal with the packing at the same time as allowing unboxed vec128s to be put in product arrays.  In both cases I think the indexing code needs adjusting, but this isn't completely straightforward.

testsuite/tests/typing-layouts/layout_poly.ml
ccasin: I've pushed a commit that updates this test appropriately.

ccasin (changes requested):
Looked at the bits I was tagged for - some minor suggestions.

lambda/lambda.mli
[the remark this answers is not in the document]
TheNumbat: In any case I think it would be even more surprising for `#(int32# * int32#)` to have different packing than `{ a : int32#; b : int32# }`, so we should do all of the packing at once.

lambda/translprim.ml
```diff
@@ -1861,6 +1909,11 @@ let report_error ppf = function
   | Product_iarrays_unsupported ->
       fprintf ppf
         "Immutable arrays of unboxed products are not yet supported."
+  | Invalid_array_kind_for_uninitialized_makearray_dynamic ->
+      fprintf ppf
+        "%%makearray_dynamic_uninit can only be used for GC-ignorable arrays@ \
+         not involving tagged immediates; and arrays of unboxed numbers.@ Use \
+         %%makearray instead, providing an initializer."
```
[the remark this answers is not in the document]
mshinwell: @rtjoa can presumably add this as part of his testsuite work

lambda/translprim.ml
```diff
@@ -1228,31 +1253,54 @@ let specialize_primitive env loc ty ~has_constant_constructor prim =
...
-      else Some (Primitive (Pmakearray_dynamic (array_type, mode), arity))
+      unboxed_product_iarray_check loc new_array_kind array_mut;
+      if array_kind = new_array_kind then None
+      else Some (Primitive (Pmakearray_dynamic (new_array_kind, mode), 2))
+    end
+  | Primitive (Pmakearray_dynamic (array_kind, mode), 1), _ :: [] -> begin
+      (* This is the version of the primitive that returns an uninitialized
+         array *)
+      let loc = to_location loc in
+      match is_function_type env ty with
```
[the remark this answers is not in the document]
mshinwell: Fixed

mshinwell:
I've also rebased this.

middle_end/flambda2/from_lambda/lambda_to_lambda_transforms.ml
```diff
@@ -110,7 +110,321 @@ let rec_catch_for_for_loop env loc ident start stop
...
+  let is_local =
+    L.of_bool (match mode with Alloc_heap -> false | Alloc_local -> true)
+  in
+  let external_call_desc =
+    makearray_dynamic_unboxed_product_c_stub
+      ~name:"caml_makearray_dynamic_scannable_unboxed_product" mode
+  in
+  (* Note that we don't check the number of unarized arguments against the
+     layout; we trust the front end. If we wanted to do this, it would have to
+     be done slightly later, after unarization. *)
```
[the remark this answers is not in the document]
mshinwell: I experimented for a while, but this really does seem non-trivial, and I think we should leave it for now.

middle_end/flambda2/from_lambda/lambda_to_lambda_transforms.ml
```diff
@@ -110,7 +110,422 @@ let rec_catch_for_for_loop env loc ident start stop
...
+  in
+  env, Transformed term
+
+let initialize_array env loc ~length array_set_kind width ~init creation_expr =
+  match init with
+  | None -> env, Transformed creation_expr
+  | Some init ->
+    initialize_array0 env loc ~length array_set_kind width ~init creation_expr
+
+let makearray_dynamic_singleton name (mode : L.locality_mode) ~length ~init loc
```
rtjoa, line 179: @mshinwell Still working on tests, but found this along the way. The following, when compiled with `ocamlopt`:
```ocaml
external[@layout_poly] make_vect : ('a : any_non_null) . int -> 'a -> 'a array =
  "%makearray_dynamic"

let _ = make_vect 100 #3.14
```

Gives this error:
```
:(.text+0x1f): undefined reference to `caml_make_unboxed_float64_vect'
collect2: error: ld returned 1 exit status
```

mshinwell: Thanks, fixed.

mshinwell:
Have rebased.

testsuite/tests/typing-layouts-arrays/test_ignorable_product_array_1.ml
```diff
@@ -0,0 +1,93 @@
+(* TEST
```
mshinwell, line 1: @rtjoa @ccasin These `test_*_product_array_*.ml` tests don't seem to be checking the locally-allocating versions of the primitives.  However the new autogenerated tests do, so maybe this is ok?

testsuite/tests/typing-layouts-arrays/generate_makearray_dynamic_tests.ml
```diff
@@ -0,0 +1,678 @@
...
+
+let mark_test_run test_id =
+  if not (List.mem ~set:!tests_run test_id) then
+    tests_run := test_id :: !tests_run
+
+(* Various interesting values *)
+
+let sizes = [ 0; 1; 2; 30; 31; 32 ]
+
+let bad_indices size = [ -100; -2; -1; size; size + 1; size + 100 ]
```
mshinwell, line 375: I think this list should include `min_int` and `max_int`.

testsuite/tests/typing-layouts-arrays/test_ignorable_product_array_1.ml
ccasin: I think it's OK for the reason you identified.  I believe @rtjoa looked at this and judged it easier to write those new tests than adapt this framework.

testsuite/tests/typing-layouts-arrays/generate_makearray_dynamic_tests.ml
rtjoa: Thanks, added a commit to [`rtjoa.makearray_dynamic_tests`](https://github.com/ocaml-flambda/flambda-backend/tree/rtjoa.makearray_dynamic_tests).

rtjoa:
@mshinwell Looks like `min_int` found some failures. I believe when we overflow by multiplying the index by the scalar width of the array kind.

(Confirming this: We fail on exactly the types whose scalar width is even. Small even numbers * int_min is 0, whereas small odd numbers * int_min are still negative.)

(I also added some more informative printing of failing tests on [`rtjoa.makearray_dynamic_tests`](https://github.com/ocaml-flambda/flambda-backend/tree/rtjoa.makearray_dynamic_tests).)

ccasin (approved):

PR #3318, opened 2024-11-26
Implement %arrayblit

mshinwell:
Implements something like:
```
external[@layout_poly] arrayblit :
  ('a : any). 'a array -> int -> 'a array -> int -> int -> unit =
  "%arrayblit"

external[@layout_poly] arrayblit_immut :
  ('a : any). 'a iarray -> int -> 'a array -> int -> int -> unit =
  "%arrayblit_immut"
```
This is done via a Lambda-to-Lambda transformation on the way into Flambda 2.

Actually, the signatures above are a bit of a lie, because this is not currently supported for any array of kind `Pgenarray`, `Paddrarray` or `Pfloatarray`.  @ccasin if we propagated the mode of the source array, we could remove this restriction I think.  (The problem is that we need to be able to convert an `array_set_kind` into an `array_ref_kind`.)

### DIFF
--- a/bytecomp/bytegen.ml
+++ b/bytecomp/bytegen.ml
@@ -688,6 +688,8 @@ let comp_primitive stack_info p sz args =
        blocks that can't be marshalled. We've decided to ignore that problem in
        the short term, as it's unlikely to cause issues - see the internal arrays
        epic for out plan to deal with it. *)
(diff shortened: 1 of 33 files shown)


mshinwell:
(Based on #3317)

mshinwell:
I'm going to move this into #3317 as it's easier for the tests.

PR #3326, opened 2024-11-29
Add inplace minimisation options for chamelon

Ekdohibs:
This adds two new options to chamelon, `--cmt` which is used to specify the path to an existing cmt file that chamelon will now use instead of generating it then deleting it; and `-i`, which minimizes the input file in-place (in which case, the compilation command should include the input files), useful when minimizing a file that depends on a `.cmi`, for instance.

### DIFF
--- a/chamelon/chamelon.ml
+++ b/chamelon/chamelon.ml
@@ -9,7 +9,8 @@ open Cmt_format
 let usage_msg =
   Format.asprintf
     "usage: %s <file1> [<file2>] ... -c \"<command>\" [-m <minimizers>] [-x \
-     <minimizers>] [-e <error>] [-t <typing command>] [-o <output>]"
+     <minimizers>] [-e <error>] [[-t <typing command>] | [--cmt <cmt file>]] \
+     [-i | [-o <output>]]"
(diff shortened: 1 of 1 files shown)


bclement-ocp:
Is inplace minimization going to destroy the original file? If so, that should probably be documented heavily (and maybe consider making a backup).

It also seems that with this option the input file needs to be provided twice (once for the underlying command, once for Chamelon). Is that right and if so what's the motivation?

Ekdohibs:
Yes, this destroys the original file (maybe we should consider doing a backup before this?).
The file has to be provided twice indeed, the objective is to be able to use more complex commands than a simpler compiler invocation (for instance, minimise a file `a.ml` which causes problems when compiling another file `b.ml`).

bclement-ocp:
I see, thanks for the clarification on why the file name needs to be provided twice, it makes sense.

Regarding the file destruction — I think either a backup or a more scary-looking flag name would be good. `-i` is easily confused as meaning "input" in this context, so something like `-inplace` or `-mutate-inplace` would reduce the risks.

lthls (approved):

PR #3328, opened 2024-11-29
Module inclusion check looks at module mode

riaqn:
Value description inclusion check now considers the mode of the modules, as opposed to looking at the value modalities only. 

This PR is surprisingly simple, because modules are currently fixed to be legacy. When we support non-legacy modules in the future, we will have to pipe those data through.

Based on this PR, I will shortly make another PR that simplifies inferred modalities.

### DIFF
--- a/testsuite/tests/typing-modes/val_modalities.ml
+++ b/testsuite/tests/typing-modes/val_modalities.ml
@@ -479,6 +479,35 @@ end
 module M : sig module N : sig val foo : 'a -> 'a @@ global many end end
 |}]
 
+(* CR zqian: modal inclusion check should cross modes *)
+module M : sig
+  module N : sig val foo : int @@ portable end
+end = struct
+  module N = struct let foo @ nonportable = 42 end
+end
+[%%expect{|
+Lines 3-5, characters 6-3:
+3 | ......struct
+4 |   module N = struct let foo @ nonportable = 42 end
+5 | end
+Error: Signature mismatch:
+       Modules do not match:
+         sig module N : sig val foo : int @@ global many end end
+       is not included in
+         sig module N : sig val foo : int @@ portable end end
+       In module "N":
+       Modules do not match:
+         sig val foo : int @@ global many end
+       is not included in
+         sig val foo : int @@ portable end
+       In module "N":
+       Values do not match:
+         val foo : int @@ global many
+       is not included in
+         val foo : int @@ portable
+       The second is portable and the first is nonportable.
+|}]
+
 (* module constraint is modal *)
 module F (M : sig val foo : 'a -> 'a end) = struct
   module M' : sig val foo : 'a -> 'a @@ global many end = M
--- a/testsuite/tests/typing-modes/val_modalities.ml
+++ b/testsuite/tests/typing-modes/val_modalities.ml
@@ -721,7 +721,10 @@ Error: Signature mismatch:
        The second is global_ and the first is not.
 |}]
 
-(* Similar for functor type declaration *)
+(* functor type inclusion: the following two functor types are equivalent,
+  because a functor of the first type at any mode, can be zero-runtime casted
+  to the second type at the same mode. Essentially, the parameter and return
+  mode is in the functor type, and doesn't depend on the mode of the functor. *)
 module M : sig
   module type F = (sig val foo : 'a @@ global many end) ->
     (sig end)
@@ -730,37 +733,8 @@ end = struct
     (sig end)
 end
 [%%expect{|
-Lines 4-7, characters 6-3:
-4 | ......struct
-5 |   module type F = (sig val foo : 'a end) ->
-6 |     (sig end)
-7 | end
-Error: Signature mismatch:
-       Modules do not match:
-         sig module type F = sig val foo : 'a end -> sig end end
-       is not included in
-         sig
-           module type F = sig val foo : 'a @@ global many end -> sig end
-         end
-       Module type declarations do not match:
-         module type F = sig val foo : 'a end -> sig end
-       does not match
-         module type F = sig val foo : 'a @@ global many end -> sig end
-       The second module type is not included in the first
-       At position "module type F = <here>"
-       Modules do not match:
-         functor $S1 -> ...
-       is not included in
-         functor $T1 -> ...
-       Module types do not match:
-         $S1 = sig val foo : 'a @@ global many end
-       does not include
-         $T1 = sig val foo : 'a end
-       Values do not match:
-         val foo : 'a
-       is not included in
-         val foo : 'a @@ global many
-       The second is global_ and the first is not.
+module M :
+  sig module type F = sig val foo : 'a @@ global many end -> sig end end
 |}]
 
 module M : sig
@@ -771,39 +745,7 @@ end = struct
     (sig end) -> (sig val foo : 'a @@ global many end)
 end
 [%%expect{|
-Lines 4-7, characters 6-3:
-4 | ......struct
-5 |   module type F =
-6 |     (sig end) -> (sig val foo : 'a @@ global many end)
-7 | end
-Error: Signature mismatch:
-       Modules do not match:
-         sig
-           module type F = sig end -> sig val foo : 'a @@ global many end
-         end
-       is not included in
-         sig module type F = sig end -> sig val foo : 'a end end
-       Module type declarations do not match:
-         module type F = sig end -> sig val foo : 'a @@ global many end
-       does not match
-         module type F = sig end -> sig val foo : 'a end
-       The second module type is not included in the first
-       At position "module type F = <here>"
-       Module types do not match:
-         sig end -> sig val foo : 'a end
-       is not equal to
-         sig end -> sig val foo : 'a @@ global many end
-       At position "module type F = <here>"
-       Modules do not match:
-         sig val foo : 'a end
-       is not included in
-         sig val foo : 'a @@ global many end
-       At position "module type F = <here>"
-       Values do not match:
-         val foo : 'a
-       is not included in
-         val foo : 'a @@ global many
-       The second is global_ and the first is not.
+module M : sig module type F = sig end -> sig val foo : 'a end end
 |}]
 
 module type T = sig @@ portable
--- a/testsuite/tests/typing-modes/val_modalities.ml
+++ b/testsuite/tests/typing-modes/val_modalities.ml
@@ -465,11 +465,27 @@ Error: Signature mismatch:
 |}]
 
 
-(* module inclusion check should look at the modes of the modules when comparing
-modalities. Currently modules are always global many, which allows more check to
-pass. *)
+(* module inclusion check should look at the modes of the modules, since some
+module type inclusion is only true for certain modes. Currently modules are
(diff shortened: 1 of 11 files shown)


testsuite/tests/typing-modes/val_modalities.ml
```diff
@@ -461,7 +461,18 @@ Error: Signature mismatch:
          val f : int -> int @@ global many
        is not included in
          val f : int -> int @@ portable
-       The second is portable and the first is not.
+       The second is portable and the first is nonportable.
+|}]
+
+module F (M : sig val f : 'a -> 'a @@ global many end) = struct
```
goldfirere, line 467: Were these not accepted before this patch?

typing/includecore.ml
```diff
@@ -92,6 +93,9 @@ let primitive_descriptions pd1 pd2 =
 let value_descriptions ~loc env name
     (vd1 : Types.value_description)
     (vd2 : Types.value_description) =
+  (* modules are now fixed to be legacy *)
+  let mm1 = Mode.Value.legacy in
```
goldfirere, line 97: I don't quite understand the semantics of these variables. The check here is a check between module *types*. How can we be talking about *modules*?

testsuite/tests/typing-modes/val_modalities.ml
riaqn: Yes - the first commit shows the pre-patch output.

typing/includecore.ml
riaqn: A naive module inclusion check first checks that that the mode of the first module is less than the second. It then checks that the type of the first module is included by the second, which involves submodality check. Sub-modality, `f <= g`, is defined as "f x <= g x for all x". This is too strong - we already know the modes of modules. So instead we check that `f x <= g y` where `x` and `y` are the mode of first and second module.

lpw25: I don't think you want to treat:
`sig val foo : t end` and `sig val foo : t @@ global end` as equal module types just because all modules are currently global, but that is what this will do as written. You should pipe in `mm1` from the point at which you know you are comparing the module type of an actual module. When comparing module types from module type definitions then you should instead use top (or bottom? whichever one gives the right behaviour) rather than legacy.

typing/includecore.ml
```diff
@@ -102,9 +106,17 @@ let value_descriptions ~loc env name
   | Ok () -> ()
   | Error e -> raise (Dont_match (Zero_alloc e))
   end;
-  begin match Mode.Modality.Value.sub vd1.val_modalities vd2.val_modalities with
-  | Ok () -> ()
-  | Error e -> raise (Dont_match (Modality e))
+  (* [wrap_constraint_with_shape] invokes inclusion check with idential inferred
+    modalities, which we need to workaround. *)
+  if not (vd1.val_modalities == vd2.val_modalities) then begin
```
lpw25, line 111: Is this workaround new or was it previously in `Modality.Value.sub`?

typing/includecore.ml
```diff
@@ -43,6 +43,7 @@ type value_mismatch =
   | Type of Errortrace.moregen_error
   | Zero_alloc of Zero_alloc.error
   | Modality of Mode.Modality.Value.error
```
lpw25, line 45: Is this `Modality` case still reachable?

lpw25 (commented):
I think the current behaviour isn't quite right for module type definitions.

typing/includecore.ml
riaqn: Yes it's also in `Modality.Value.sub`, which is probably unreachable now.

typing/includecore.ml
riaqn: No it's no longer reachable. But I think as we introduce non-legacy modules, comparing two module types will be done without knowing modes, so this will be reachable again.

typing/includecore.ml
goldfirere: First off, I think there are now two different threads in here.

1. About the semantics of this function (my question).
2. How to behave in the current world where all modules are at the legacy mode (Leo's question).

I'll go back to 1. I had been thinking that the inclusion check was checking a subsumption relationship between two module types. But as I think about it more, I think that's wrong -- we never need to know one module type subsumes another, I think. (Importantly, the inclusion check treats module types themselves as invariant.) Instead, we need to know that some module `M : S1` can be used at type `S2`. So this gives a nice meaning for `mm1` here: it's the mode of `M` in that check. But what's `mm2` then?

Do you agree that we're always checking to see if some module `M` can be used at a type `S2`? And then can we get rid of `mm2` maybe? (Or perhaps it's the case that we use submoding to go from `mm1` to `mm2`... and one might be more permissive than the other regarding modality inclusion? Not sure here.)

riaqn: `mm1` is the mode of the module on LHS, `mm2` is the mode of the module on RHS.

We want to check that a module of type `S1` at mode `mm1`, can be used as a module of type `S2` at mode `mm2`. This involves:

1. First check that `mm1 <= mm2`. 
2. Then for each value, `f1 mm1 <= f2 mm2`, where `f1` and `f2` are the modalities on values.

1 is not needed right now because they are all legacy. 2 is what the code in question is about.

goldfirere: OK. I'm satisfied now. I think the key thing for me is that this isn't a check about module types, despite the names of the functions that implement the check. It's a check to see whether a coercion on a particular module is safe. Thanks.

riaqn: Yeah, that's at least what I think. But maybe @lpw25 thinks differently.

lpw25: This function is used both for checking a relationship between two modules *and* for checking that two module types are equal (by checking the subsumption both ways). My point is that it is currently incorrect for that second use case. If you thread `mm1` and `mm2` in from the caller than you can get it to behave correctly for both uses:
- You can set them to the actual module modes in the first use case (which are currently both legacy)
- You can set them to top/bottom in the module type comparison case.

riaqn: Resolved since we had offline discussion.

riaqn:
Had an offline discussion - we now distinguish between module type declaration vs module/value declaration, when performing inclusion check. Only the latter is affected by this PR.

typing/includemod.ml
```diff
@@ -550,16 +551,27 @@ and try_modtypes ~in_eq ~loc env ~mark subst mty1 mty2 orig_shape =
...
       | Error e -> Error (Error.Signature e)
       end
 
   | Mty_functor (param1, res1), Mty_functor (param2, res2) ->
+      let modes_param, modes_res =
+        match modes with
+        | None ->
+            (* comparing two modules type decls, in which case we compare functor
+            parameter and return as module type decls. *)
+            None, None
```
lpw25, line 566: This actually doesn't look quite right to me. I don't think whether we are checking an inclusion as part of a `module type` check affects what has to happen here. For example, I think that this check should pass:
```ocaml
(struct
  module type S = functor (X : sig val x : string @@ global end) -> sig end
  end: sig
   module type S = functor (X : sig val x : string end) -> sig end
 end)
```
the two `S`s are mutually convertible without runtime coercion at any mode -- which is what the definition of module type equality is.

This is different from:
```ocaml
(struct
  module type S = sig val x : string @@ global
  end: sig
   module type S = sig val x : string
 end)
 ```
 because here the two `S`s are not mutually convertible at all modes: they would not be convertible at `local`. Of course we currently don't support `local` modules, but we don't want to make our type equality bake in that fact as we'd like to change it later.
 
(I think I got myself confused in a discussion with you the other day around this area, and strongly suspect that is what lead you to write it this way. Sorry about that).

File: typing/includemod.mli [hunk 6]
```diff
@@ -154,15 +162,15 @@ val is_runtime_component: Types.signature_item -> bool
 (* Typechecking *)
 
 val modtypes:
-  loc:Location.t -> Env.t -> mark:mark ->
+  loc:Location.t -> Env.t -> mark:mark -> modes:unit option ->
```
lpw25, line 165: In addition to changing these `modes` parameters to carry the modes, I think it might be worth making a specific type like:
```
type inclusion_mode =
  | All
  | Specific of { from; to_ }
```

lpw25 (commented):
I think it would be worth tidying up a few things, and I think the behaviour should be different for functors within module types.

typing/includemod.ml
```diff
@@ -79,7 +77,10 @@ module Error = struct
   and arg_functor_param_symptom =
     (functor_parameter, Ident.t) functor_param_symptom
 
-  and functor_params_diff = (functor_parameter list * module_type) core_diff
+  and functor_params_symptom = bool
```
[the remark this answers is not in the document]
riaqn: there is a comment in `includemod.mli`.

(The latest commit no longer uses boolean anyway)

typing/includemod.mli
riaqn: yeah that's what I did in the last commit. Although, the `Specific` constructor still doesn't carry modes.

typing/includemod.ml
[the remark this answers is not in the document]
riaqn: As discussed, this PR cares about the presense/absense of modes, not module vs. module type, and because functor parameters carries their own modes, we don't need to record this.

typing/includemod.ml
riaqn: As discussed offline, we agree this is the right way to go. Fixed in commit "correct semantics".

typing/includecore.mli
```diff
@@ -120,6 +121,9 @@ type type_mismatch =
 
 val value_descriptions:
   loc:Location.t -> Env.t -> string ->
+  mmodes:unit option ->
+  (* [unit] because modules are fixed to legacy. Will be replaced by
+     [Mode.Value.(l * r)] when allow modal modules. *)
```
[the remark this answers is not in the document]
riaqn: I adopted the other suggestion of having a dedicated type:

```
type modes = All | Legacy
```
which I think makes the code very sensible.

lpw25 (approved):
A couple of small suggestions, but this is good to go.

typing/includemod.ml
```diff
@@ -914,7 +924,12 @@ and module_declarations  ~in_eq ~loc env ~mark  subst id1 md1 md2 orig_shape =
   let p1 = Path.Pident id1 in
   if mark_positive mark then
     Env.mark_module_used md1.md_uid;
-  strengthened_modtypes  ~in_eq ~loc ~aliasable:true env ~mark subst
+  let modes =
+    match mmodes with
+    | All -> All
+    | Legacy -> Legacy
+  in
```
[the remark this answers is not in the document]
riaqn: In the future, with modalities on module declarations, we have to apply modalities on `mmodes` to get `modes`. If we pass them directly, we won't get any type error if I forgot to apply modalities.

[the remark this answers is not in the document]
lpw25: Yeah, that is what I had guessed, but I still wouldn't match on `mmodes` to make `modes`.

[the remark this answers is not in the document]
lpw25: You aren't going to forget to make that update, and if you do your tests will show you it.

PR #3330, opened 2024-12-02
Allow more minor custom blocks in runtime5

stedolan:
Runtime5 is currently overly aggressive in allocating bigarrays directly on the major heap, skipping the minor heap for anything over 70kB.

This patch changes the defaults to allow larger bigarrays to be collected by the minor collector, both by allowing more total minor bigarrays (up to 4x minor heap by default), and by allowing larger single bigarrays (10% of the amount that would trigger minor GC by default). There's some explanation of the new parameter defaults in `gc.mli`.

### DIFF
--- a/runtime/custom.c
+++ b/runtime/custom.c
@@ -117,7 +117,7 @@ CAMLexport value caml_alloc_custom_mem(const struct custom_operations * ops,
                                        mlsize_t mem)
 {
   mlsize_t max_minor = get_max_minor (); /* total allocs before minor GC */
-  mlsize_t max_minor_single;     /* largest single alloc before minor GC */
+  mlsize_t max_minor_single;      /* largest allowed alloc on minor heap */
   if (caml_custom_minor_max_bsz > 100) {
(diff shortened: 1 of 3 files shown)


runtime/custom.c
```diff
@@ -108,14 +109,22 @@ CAMLexport value caml_alloc_custom(const struct custom_operations * ops,
...
+  return alloc_custom_gen (ops, bsz, mem, max_major, max_minor, 1);
 }
 
 CAMLexport value caml_alloc_custom_mem(const struct custom_operations * ops,
                                        uintnat bsz,
                                        mlsize_t mem)
 {
-  value v = alloc_custom_gen (ops, bsz, mem, 0, get_max_minor());
+  mlsize_t max_minor = get_max_minor (); /* total allocs before minor GC */
+  mlsize_t max_minor_single;     /* largest single alloc before minor GC */
```
mshinwell, line 120: This second comment seems to be wrong, as discussed.

mshinwell (approved):

PR #3331, opened 2024-12-02
Fix GC pacing problem exacerbated by mark-delay

NickBarnes:
This fixes a problem detected by CI when upstreaming #2348 / #2358 / #3029 to https://github.com/ocaml/ocaml/pull/13580. It is a pacing problem in which major GC work performed can, in some circumstances, fall further and further behind the estimated work required. See the discussion on https://github.com/ocaml/ocaml/pull/13580 for more details. In a long-running process this could be problematic, leading to longer GC pauses. On 32-bit platforms the flawed estimation can lead to an overflow which stops GC altogether, fatally. The short-term fix in this PR applies a "catch-up" adjustment when it may be necessary. Work is in progress on replacing all the pacing code, which should result in a longer-term fix.

This is one of the three fixes in #3297, which @mshinwell asked to be separated out.

### DIFF
--- a/runtime/major_gc.c
+++ b/runtime/major_gc.c
@@ -755,6 +755,27 @@ static void update_major_slice_work(intnat howmuch,
   atomic_fetch_add (&work_counter, dom_st->major_work_done_between_slices);
   dom_st->major_work_done_between_slices = 0;
   atomic_fetch_add (&alloc_counter, new_work);
+
+  /* If the work_counter is falling far behind the alloc_counter,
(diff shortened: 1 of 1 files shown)


mshinwell (approved):

mshinwell:
This seems to be causing a flaky test failure e.g. https://github.com/ocaml-flambda/flambda-backend/actions/runs/12143046690/job/33858963003 .  Reverting for now, I've pinged Nick separately.

PR #3335, opened 2024-12-02
Consume patterns that do not have a working unique barrier yet

anfelor:
This PR changes the pattern-matching code in the uniqueness analysis so that the value we match on is consumed by certain patterns. This guarantees that the pattern-match can not be followed by a destructive write. This is necessary for those patterns where we do not propagate the unique barrier correctly yet, since for those patterns the read generated from the pattern-match could be pushed down a destructive write. This PR thus moves us further towards ensuring soundness of uniqueness for destruction.

### DIFF
--- a/typing/uniqueness_analysis.ml
+++ b/typing/uniqueness_analysis.ml
@@ -227,8 +227,10 @@ module Aliased : sig
   type t
 
   type reason =
-    | Forced  (** aliased because forced  *)
-    | Lazy  (** aliased because it is the argument of lazy forcing *)
+    | Forced  (** aliased because forced due to multiple usage *)
+    | Lazy  (** aliased because of a lazy pattern *)
+    | Array  (** aliased because of an array pattern *)
+    | Constant  (** aliased because of an constant pattern *)
     | Lifted of Maybe_aliased.access
         (** aliased because lifted from implicit borrowing, carries the original
           access *)
@@ -246,6 +248,8 @@ end = struct
   type reason =
     | Forced
     | Lazy
+    | Array
+    | Constant
     | Lifted of Maybe_aliased.access
 
   type t = Occurrence.t * reason
@@ -261,6 +265,8 @@ end = struct
     let print_reason ppf = function
       | Forced -> fprintf ppf "Forced"
       | Lazy -> fprintf ppf "Lazy"
+      | Array -> fprintf ppf "Array"
+      | Constant -> fprintf ppf "Constant"
       | Lifted ma -> fprintf ppf "Lifted(%a)" Maybe_aliased.print_access ma
     in
     fprintf ppf "(%a,%a)" Occurrence.print occ print_reason reason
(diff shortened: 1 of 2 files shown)


typing/uniqueness_analysis.ml
```diff
@@ -1058,117 +1062,133 @@ let rec pattern_match_tuple pat values =
...
+    updates can be performed on the memory.
+
+    Each pattern falls into one of three cases:
+    - If we do not read from the underlying memory,
+      we do not have to take an action.
+    - We can allow destructive updates later on by borrowing the
+      memory address. Then we have to protect the read from getting
+      pushed down using a unique barrier.
+    - We can disallow any destructive updates following the read
+      by consuming the memory address as aliased. *)
```
goldfirere, line 1078:
```suggestion
      by consuming the memory address as aliased.
      
    [pattern_match_single] recurs down the structure of the pattern, calling [pattern_match_barrier]
    at each step, so [pattern_match_barrier] itself does not need to recur into subpatterns.   
         *)
```

typing/uniqueness_analysis.ml
```diff
@@ -1058,117 +1062,133 @@ let rec pattern_match_tuple pat values =
...
-    no_borrow_memory_address ();
-    Ienv.Extension.singleton id paths, UF.unused
-  | Tpat_alias (pat', id, _, _, _) ->
-    no_borrow_memory_address ();
-    let ext0 = Ienv.Extension.singleton id paths in
-    let ext1, uf = pattern_match_single pat' paths in
-    Ienv.Extension.conjunct ext0 ext1, uf
+  | Tpat_or _ -> no_memory_access ()
+  | Tpat_any -> no_memory_access ()
+  | Tpat_var _ -> no_memory_access ()
```
goldfirere, line 1098: Why is this case correct? Binding a variable looks like a memory access. I'm having a hard time understanding what to do here. Maybe we're answering the question "Is a read of memory required to determine whether the match succeeds?" That sounds plausible. And then it's pretty clear that the answer is "no" here but "yes" in the constant case. But mostly I'm not sure.

typing/uniqueness_analysis.ml
```diff
@@ -1058,117 +1062,133 @@ let rec pattern_match_tuple pat values =
...
-    let uf_read = borrow_memory_address () in
-    Ienv.Extension.empty, uf_read
-  | Tpat_construct (lbl, cd, pats, _) ->
-    let uf_read = borrow_memory_address () in
-    let pats_args = List.combine pats cd.cstr_args in
-    let ext, uf_pats =
+    (* This is necessary since we can not guarantee that
+       the reads of constants in the pattern-matching code
+       are never pushed down. *)
+    consume_memory_address Constant
```
goldfirere, line 1104: Is this still the case for reading `int`s? Again, I'm not very clear on the exact semantics of this.

typing/uniqueness_analysis.ml
```diff
@@ -1058,117 +1062,133 @@ let rec pattern_match_tuple pat values =
...
+  | Tpat_construct _ -> borrow_memory_address ()
+  | Tpat_variant _ -> borrow_memory_address ()
+  | Tpat_record _ -> borrow_memory_address ()
```
goldfirere, line 1107: Now I'm confused again. Sorry! This doesn't have to read memory to tell whether the pattern succeeds, because a record pattern always succeeds. So I think I just don't understand the property we're trying to capture here.

goldfirere (commented):
I went to review, only to discover that I don't really understand what this function is meant to be doing. But instead of answering me here, maybe address this in the comment? Or we can talk about it later. I'll then take another pass.

typing/uniqueness_analysis.ml
anfelor: Your heuristic is exactly right: Binding a variable does not require dereferencing the pointer. That is why there is no memory access required here.

typing/uniqueness_analysis.ml
anfelor: Yeah, we might be able to do better for certain constants. But `consume_memory_address` is always safe.

typing/uniqueness_analysis.ml
anfelor: Yes, that is correct. We could create a special case here for record patterns that do not bind any names and those would not require a memory access. But part of the record pattern is that we can learn the memory addresses of its children by naming them which requires dereferencing the pointer to the record.

goldfirere (approved):
Thanks for explaining this to me -- all looks good now. And I really like the way the new code separates out concerns; good idea.

PR #3337, opened 2024-12-03
Unboxed pair returns from C externals

xclerc:
(joint work with @mshinwell)

This feature implements support for unboxed
pairs as return values.


mshinwell:
@Gbury could you look at the Cmm parts please?

mshinwell:
@TheNumbat could you review the `Proc` and the testsuite changes please?

backend/amd64/proc.ml
```diff
@@ -283,10 +323,28 @@ let max_arguments_for_tailcalls = 10 (* in regs *) + 64 (* in domain state *)
      Return value in rax or xmm0. *)
 
 let loc_external_results res =
-  let (loc, _ofs) = calling_conventions 0 0 100 100 not_supported 0 res in loc
+  let (loc, _ofs) =
+    calling_conventions
+      ~first_int:0
+      ~last_int:4
+      ~step_int:4
```
Gbury, line 330: it would be nice to have an explanation of why the step is `4` here...

middle_end/flambda2/to_cmm/to_cmm_expr.ml
```diff
@@ -265,15 +267,39 @@ let translate_apply0 ~dbg_with_inlined:dbg env res apply =
...
-      env,
-      res,
-      Ece.all )
+    let extcall =
+      C.extcall ~dbg ~alloc:needs_caml_c_call ~is_c_builtin ~effects ~coeffects
+        ~returns ~ty_args callee return_ty args
+    in
+    let wrap return_values =
+      let kinds = Flambda_arity.unarized_components return_arity in
+      assert (List.compare_length_with kinds (Array.length component_tys) = 0);
```
Gbury, line 276: Does that assertion match the comment above about `void` ? I.e. when exactly does `void` disappear from return arity & others unboxed tuples ?

middle_end/flambda2/to_cmm/to_cmm_expr.ml
```diff
@@ -238,25 +238,27 @@ let translate_apply0 ~dbg_with_inlined:dbg env res apply =
...
+
+         1. void has been erased in return arities by this point
+
+         2. All of the [machtype_component]s are singleton arrays. *)
+      Array.map (fun machtype -> [| machtype |]) return_ty
+    in
+    (* Returned int32 values need to be sign_extended because it's not clear
+       whether C code that returns an int32 returns one that is sign extended or
+       not. There is no need to wrap other return arities. Note that extcalls of
+       arity 0 are allowed (these never return). *)
```
Gbury, line 253: the part about arity 0 should be moved elsewhere

middle_end/flambda2/to_cmm/to_cmm_expr.ml
```diff
@@ -265,15 +267,39 @@ let translate_apply0 ~dbg_with_inlined:dbg env res apply =
...
+        C.make_tuple
+          (List.mapi
+             (fun i kind ->
+               maybe_sign_extend kind dbg
+                 (get_unarized_return_value return_values i))
+             kinds)
+      | _ ->
+        Misc.fatal_errorf
+          "C functions are currently limited to a single return value or a \
+           pair of return values"
```
Gbury, line 294: could you try and split off (most of) the code for treating unboxed returns in a separate function ? The code for the `C_call` case is starting to be quite long..

middle_end/flambda2/to_cmm/to_cmm_expr.ml
mshinwell: I think it should do.  Have added a comment.  We'll get tests for void in due course (unsure exactly when).

backend/amd64/proc.ml
xclerc: That is indeed a hack to get `rdx` as the second
register. Hopefully the comments will clarify it,
but we are also planning a refactoring for `Reg.t`
so that should not stay as ugly for too long.

middle_end/flambda2/to_cmm/to_cmm_expr.ml
mshinwell: I've moved the whole thing, indeed it is getting lengthy.

stedolan:
Which types is this expected to work for?

The x86 calling convention rules for structs of two elements are tricky when the element sizes are smaller than 64 bits. For instance, a struct of two `uint8_t` values gets packed into the low 16 bits of `rax`.

mshinwell:
I think just 64-bit ints and floats for the moment?  (A restriction in the frontend for the typing here will need to be added.)

Gbury (approved):

mshinwell:
Yeah I'll discuss with @xclerc

TheNumbat (commented):
Proc changes/test looks good. I agree it should assert that only bits64/float64 layouts are used.

mshinwell:
@TheNumbat I've added a check, please review the last changeset.

TheNumbat (approved):

PR #3338, opened 2024-12-03
Workaround semantics for `pmd_modalities`

riaqn:
This PR adds semantics for `pmd_modalities` (syntax introduced by #3223 ):
```
module M : S @@ portable
```

Lacking proper modal modules, we adopted a workaround semantics. Please see inlined comments for details.


typing/typemod.ml
```diff
@@ -1037,6 +1037,37 @@ let check_no_modal_modules ~env modes =
   | None -> ()
   | Some loc -> raise(Error(loc, env, Modal_module_not_supported))
 
+let apply_pmd_modalities env sig_modalities pmd_modalities mty =
+  let modalities =
+    match pmd_modalities with
+    | [] -> sig_modalities
+    | _ :: _ ->
+      Typemode.transl_modalities ~maturity:Alpha Immutable [] pmd_modalities
```
lpw25, line 1045: Maybe leave a comment explaining this is broken, but it is broken in the same way as `val` declarations.

lpw25 (approved):
I'm said about the hacky implementation of `nonportable` but it is apparently the same hacky implementation used for `nonportable` on `val` declarations, so I guess we can fix them both later.

PR #3340, opened 2024-12-04
Fix for PR2755 (ocamltest fail-if-test-does-nothing)

mshinwell:
#2755 caused ocamltest to produce test failures for stanzas that (often surprisingly!) do not actually run any tests.  For example before that PR, this:
```
(* TEST
   runtime5;
*)
```
caused nothing to be tested even on runtime5 (the test passes silently), but #2755 causes it to fail.  Unfortunately this functionality got lost in the 5.2 merge - I thought it had been merged correctly at the time but was mistaken.  This PR reworks the previous one and should restore things.

There only seems to be one test (fixed in this PR) which was silently failing to run, namely `lib-unix/common/bigarrays.ml`.

### DIFF
--- a/ocamltest/main.ml
+++ b/ocamltest/main.ml
@@ -107,12 +107,14 @@ let join_summaries sa sb =
   | No_failure, (No_failure | All_skipped)
   | All_skipped, No_failure -> No_failure
 
-let rec run_test_tree log common_prefix behavior env summ ast =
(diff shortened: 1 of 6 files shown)


mshinwell:
Ah, and `DLS_thread_safety.ml` was apparently testing nothing.

xclerc (approved):

PR #3343, opened 2024-12-05
Fix unboxed product returns for arm64

mshinwell:
This fixes a problem in #3337.  For the moment the middle end is going to disallow mixed int64/float unboxed product returns from C externals on arm64, as the calling convention is complicated.


xclerc (approved):

testsuite/tests/typing-layouts-return/unboxed_return2.ml
```diff
@@ -0,0 +1,32 @@
+(* TEST
+ modules = "stubs.c";
+ reference = "${test_source_directory}/unboxed_return.reference";
```
xclerc, line 3:
```suggestion
 reference = "${test_source_directory}/unboxed_return2.reference";
```

PR #3347, opened 2024-12-06
Add predefined types for tagged int8 and int16

jvanburen:
I also created modules for them in `Stdlib_beta`


jvanburen:
This will be a prerequisite for unboxed int8 and int16 types

testsuite/tests/lib-smallint/test_int8.ml
```diff
@@ -0,0 +1,143 @@
...
+let test_logical1 = test_conv1 ~equal:equal_logical
+
+let test_logical2 = test_conv2 ~equal:equal_logical
+
+let reference_shift_right_logical x i =
+  (* we need to ensure that we shift in zero bytes, which is incompatible with
+     sign-extension *)
+  Int.shift_right_logical (if i > 0 then x land mask else x) i
+
+let () =
```
ccasin, line 90: Things defined in the library that aren't tested here: `neg`, `min_int`, and `max_int`.

Testing `neg` seems straightforward.  For `min_int` and `max_int` I guess you can just compare them to the concrete expected values.

testsuite/tests/lib-smallint/test_int8.ml
```diff
@@ -0,0 +1,143 @@
...
+  test_arith1 Int8.pred Int.pred;
+  test_arith1 Int8.abs Int.abs;
+  test_logical2 Int8.logand Int.logand;
+  test_logical2 Int8.logor Int.logor;
+  test_logical2 Int8.logxor Int.logxor;
+  test_logical1 Int8.lognot Int.lognot;
+  for shift = 0 to int_size do
+    let apply_shift f x = f x shift in
+    test_logical1 (apply_shift Int8.shift_right) (apply_shift Int.shift_right);
+    test_conv1
```
ccasin, line 110: Can this be `test_logical`?

testsuite/tests/lib-smallint/test_int8.ml
```diff
@@ -0,0 +1,143 @@
...
+  test_arith1 Int8.succ Int.succ;
+  test_arith1 Int8.pred Int.pred;
+  test_arith1 Int8.abs Int.abs;
+  test_logical2 Int8.logand Int.logand;
+  test_logical2 Int8.logor Int.logor;
+  test_logical2 Int8.logxor Int.logxor;
+  test_logical1 Int8.lognot Int.lognot;
+  for shift = 0 to int_size do
+    let apply_shift f x = f x shift in
+    test_logical1 (apply_shift Int8.shift_right) (apply_shift Int.shift_right);
```
ccasin, line 109: Why is `test_logical` used for the arithmetic shift?

ccasin (commented):
Looks good, mostly minor suggestions.

otherlibs/stdlib_beta/int16.ml
```diff
@@ -0,0 +1,96 @@
...
+let[@inline always] rem x y = of_int (Int.rem (to_int x) (to_int y))
+
+let[@inline always] succ x = of_int (Int.succ (to_int x))
+
+let[@inline always] pred x = of_int (Int.pred (to_int x))
+
+let[@inline always] abs x = of_int (Int.abs (to_int x))
+
+let[@inline always] logand x y =
+  unsafe_of_int (Int.logand (to_int x) (to_int y))
```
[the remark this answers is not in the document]
jvanburen: Also I should check that abs works at min_int

otherlibs/stdlib_beta/int16.mli
```diff
@@ -0,0 +1,136 @@
+(**************************************************************************)
```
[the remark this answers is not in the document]
jvanburen: of_string is explicitly commented out of int.mli, so I didn't add that , but I added the hash functions and some others

testsuite/tests/lib-smallint/test_int8.ml
jvanburen: it's for the right shifts, not just the arithmetic one

otherlibs/stdlib_beta/int.ml
```diff
@@ -0,0 +1,319 @@
+(**************************************************************************)
```
ccasin, line 1: I think this module should be named something other than Int.  It's too confusing and error-prone to have two different modules named `Int` in the stdlib, even if one is in `stdlib_beta`.  One day we'd like all of this to be in the normal upstream stdlib!

otherlibs/stdlib_beta/.ocamlformat-enable
```diff
@@ -1,3 +1,7 @@
 stdlib_beta.mli
 stdlib_beta.ml
-
+int_intf.ml
```
ccasin, line 3: I don't see a file called `int_intf.ml` anywhere.

otherlibs/stdlib_beta/int.ml
```diff
@@ -0,0 +1,319 @@
...
+      (Container.shift_left (inject t) unused_bits)
+      unused_bits
+
+  let zero = sign_extend Container.zero
+
+  let one = sign_extend Container.one
+
+  let minus_one = sign_extend Container.minus_one
+
+  let[@inline available] neg x = sign_extend (Container.neg (inject x))
```
ccasin, line 228: I think these should just be marked "inline" - we proactively want them to be inlined, not just if the heuristic chooses it.  But probably you had some reason in mind I'm missing?

ccasin (approved):
A few more comments, but eagerly approving now.  Thanks!

otherlibs/stdlib_beta/int.ml
jvanburen: I was copying what int32.ml had, still inline always?

jvanburen: we'll ask mark and can change later, merging as-is

otherlibs/stdlib_beta/int.ml
jvanburen: Changed to `Stdlib_beta.Int_wrapper`

PR #3349, opened 2024-12-06
Add `fst` and `snd` to the capsule api

liam923:


dkalinichenko-js (approved):
LGTM. Could you add a test under `DIR=capsule-api` along with the test for `both`?

lpw25:
It might be slightly better to have `fst` and `snd` rather than `split` as they are probably more commonly used.

PR #3351, opened 2024-12-09
On Linux, use PR_SET_VMA_ANON_NAME to name memory mappings

stedolan:
This new Linux `prctl` call allows a name to be set for anonymous memory mappings. This is useful for debugging (especially performance debugging), as the names show up in `/proc/$PID/smaps` and are not coalesced with other mappings. (This means that questions like "how much of this process's RSS is OCaml heap" are answerable from `smaps`).

On a kernel that supports it, here's what part of smaps looks like:
```
7fa54de00000-7fa54e600000 rw-p 00000000 00:00 0                          [anon:OCaml: minor heap 0]
Size:               8192 kB
KernelPageSize:        4 kB
MMUPageSize:           4 kB
Rss:                4096 kB
Pss:                4096 kB
Pss_Dirty:          4096 kB
Shared_Clean:          0 kB
Shared_Dirty:          0 kB
Private_Clean:         0 kB
Private_Dirty:      4096 kB
Referenced:         4096 kB
Anonymous:          4096 kB
LazyFree:              0 kB
AnonHugePages:      4096 kB
ShmemPmdMapped:        0 kB
FilePmdMapped:         0 kB
Shared_Hugetlb:        0 kB
Private_Hugetlb:       0 kB
Swap:                  0 kB
SwapPss:               0 kB
Locked:                0 kB
THPeligible:    1
VmFlags: rd wr mr mw me ac sd 
```

### DIFF
--- a/runtime/fiber.c
+++ b/runtime/fiber.c
@@ -183,7 +183,7 @@ Caml_inline struct stack_info* alloc_for_stack (mlsize_t wosize)
   // 2Mb (= extra_size)
   // -------------------- <- [stack], returned from [mmap], page-aligned
   char* stack;
-  stack = caml_mem_map(len + stack_extra_size_for_mmap, 0);
+  stack = caml_mem_map(len + stack_extra_size_for_mmap, 0, "stack");
   if (stack == MAP_FAILED) {
     return NULL;
   }
--- a/runtime/fiber.c
+++ b/runtime/fiber.c
@@ -20,6 +20,7 @@
 
 #include "caml/config.h"
 #include <string.h>
+#include <stdio.h>
 #ifdef HAS_UNISTD
 #include <unistd.h>
 #endif
@@ -43,6 +44,11 @@
 #endif
 #if defined(USE_MMAP_MAP_STACK) || !defined(STACK_CHECKS_ENABLED)
 #include <sys/mman.h>
+#endif
+#ifdef __linux__
+/* for gettid */
+#include <sys/types.h>
+#include <sys/syscall.h>
 #endif
 
 #ifdef DEBUG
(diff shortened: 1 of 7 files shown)


runtime/shared_heap.c
```diff
@@ -211,7 +211,7 @@ static pool* pool_acquire(struct caml_heap_state* local) {
         caml_mem_round_up_mapping_size(Bsize_wsize(POOL_WSIZE) * new_pools);
       new_pools = mapping_size / Bsize_wsize(POOL_WSIZE);
 
-      void* mem = caml_mem_map(mapping_size, 0);
+      void* mem = caml_mem_map(mapping_size, 0, "major heap");
```
mshinwell, line 214: Maybe this could say how many pools it is?
Also how about some kind of sequence number here, so we could see the temporal distribution in `smaps`?

runtime/fiber.c
```diff
@@ -183,7 +183,7 @@ Caml_inline struct stack_info* alloc_for_stack (mlsize_t wosize)
   // 2Mb (= extra_size)
   // -------------------- <- [stack], returned from [mmap], page-aligned
   char* stack;
-  stack = caml_mem_map(len + stack_extra_size_for_mmap, 0);
+  stack = caml_mem_map(len + stack_extra_size_for_mmap, 0, "stack");
```
mshinwell, line 186: Would be nice if this said whether it was the main OCaml stack or a thread stack, with the tid in the latter case.

mshinwell (approved):
This is good and ok to merge, but a couple of suggestions for potentially useful improvements.

runtime/fiber.c
stedolan: I put the TID in, which sounds useful. Threading the bit for whether it's the main thread is more awkward and seems less helpful (main is when TID = PID, so it's not hard to find out), so I left it at that.

runtime/shared_heap.c
stedolan: I put in a sequence number - it sounds useful for debugging compaction, at least.

The number of pools is redundant with the size, so I left that one out.

mshinwell (approved):

PR #3352, opened 2024-12-09
Simplify inferred comonadic modality

riaqn:
This is based on #3328 

This simplifies the inferred comonadic modality when being zapped to constant. More specifically, if a module and a value therein are both `global many`, the value doesn't need the `global many` modality.

### DIFF
--- a/typing/mode.ml
+++ b/typing/mode.ml
@@ -2251,23 +2251,6 @@ module Modality = struct
 
       let print ppf = function
         | Join_const c -> Format.fprintf ppf "join_const(%a)" Mode.Const.print c
-
-      (** Given a modality and a guarantee that the modality will only be appled
-      on [x >= mm], we can find some lower modality that is equivalent on the
-      restricted range. This is similar to mode-crossing, where we can push a
-      mode lower given a restricted range of types. *)
-      let modality_cross_left ~mm = function
-        | Join_const c ->
-          (* We want to find the minimal [c'] such that [join c x <= join c' x]
-             for all [x >= mm]. By definition of join, this is equivalent to [c
-             <= join x c'] for all [x >= mm]. This is equivalent to [c <= join
-             mm c']. Equivalently [subtract c mm <= c']. Note that [mm] is a
-             mode variable, but we need a constant. Therefore, we conservatively
-             take its incomplete lower bound [mm.lower]. Also recall that we
-             want the smallest such [c']. So we take [c' = subtract c mm.lower].
-          *)
-          let mm = Mode.Guts.get_floor mm in
-          Join_const (Mode.Const.subtract c mm)
     end
 
     type t =
@@ -2322,10 +2305,13 @@ module Modality = struct
       | Const c -> c
       | Undefined -> Misc.fatal_error "modality Undefined should not be zapped."
       | Diff (mm, m) ->
-        let c = Mode.zap_to_floor m in
-        let m = Const.Join_const c in
-        (* To give the best modality, we try to cross modality. *)
-        Const.modality_cross_left ~mm m
+        let m = Mode.zap_to_floor m in
+        (* For soundness, we want some [c] such that [m <= join c mm], which
+           gives [subtract_mm m <= c]. Note that [mm] is a variable, but we need
+           a constant. Therefore, we take its floor [mm' <= mm], and we have
+           [subtract_mm m <= subtract_mm' m <= c]. *)
+        let mm' = Mode.Guts.get_floor mm in
+        Const.Join_const (Mode.Const.subtract m mm')
 
     let zap_to_id = zap_to_floor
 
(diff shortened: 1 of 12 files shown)


typing/mode.ml
```diff
@@ -2253,9 +2253,20 @@ module Modality = struct
         | Join_const c -> Format.fprintf ppf "join_const(%a)" Mode.Const.print c
     end
 
+    (* Similar to constant modalities, an inferred modality maps the mode of a
+       record/structure to the mode of a value therein. An inferred modality [f] is
+       inferred from the structure/record mode [mm] and the value mode [m].
+
+       Soundness: You should not get a value from a record/structure at a mode strictly
+       stronger than how it was put in. That is, [f mm >= m].
```
goldfirere, line 2261: Is this right? It seems OK for comonadic axes, but I don’t buy it for monadic ones.

typing/mode.ml
```diff
@@ -2460,8 +2460,26 @@ module Modality = struct
     let zap_to_floor = function
       | Const c -> c
       | Undefined -> Misc.fatal_error "modality Undefined should not be zapped."
-      | Exactly (_, m) ->
-        let c = Mode.zap_to_floor m in
+      | Exactly (mm, m) ->
+        let m = Mode.zap_to_floor m in
```
goldfirere, line 2462: I’m afraid that after staring at this comment for several minutes, I still can’t quite get to the bottom of it.

Here is a start at rewriting it that would resonate with me more:

```
The goal here is to find some [c] such that [meet_with c] can replace [Exactly(mm, m)]. For soundness, we require [m <= meet_with c mm]: that the result of [meet_with c mm] must be weaker than [m]. For completeness, we require [meet_with c mm <= m]: that the result of [meet_with c mm] is at least as useful as [m].

We will find the highest [c] that can satisfy completeness, and then prove that such a [c] is sound.

We thus start with [meet_with c mm <= m], which we rewrite to [c <= imply_with mm m]. [m] has already been zapped to its floor (this is in [zap_to_floor], after all), so that’s a constant. But [mm] is not. We must zap it. But in which direction? …
```

and this is where I lose the trail of the argument. There is some property about `imply_with` that you’re relying on… what is it? And then the comment says find a ceil with `mm’ >= mm`, but below is a call `mm’ = zap_to_floor mm`, which is in opposition.

My goal here isn’t a beautifully crafted comment, exactly. That’s not worth it. But I do think we should be able to create a shared understanding of what this function is meant to do and why, and this comment seems the best way of doing so.

typing/mode.ml
```diff
@@ -2322,10 +2305,13 @@ module Modality = struct
       | Const c -> c
       | Undefined -> Misc.fatal_error "modality Undefined should not be zapped."
       | Diff (mm, m) ->
-        let c = Mode.zap_to_floor m in
-        let m = Const.Join_const c in
-        (* To give the best modality, we try to cross modality. *)
-        Const.modality_cross_left ~mm m
+        let m = Mode.zap_to_floor m in
+        (* For soundness, we want some [c] such that [m <= join c mm], which
```
goldfirere, line 2320: This is all the dual of what was in the previous commit. Why does this change not change any tests?

goldfirere (commented):
I could be convinced to just let this one in. After all, if this change were part of a larger commit, I probably wouldn’t be zeroing in on the details here. But it would essentially be a “trust me” review, because I still don’t understand the essence of what’s going on here. Happy to chat more about it, but not sure that will be possible until Thursday at the earliest.

typing/mode.ml
riaqn: This is true for both axes (what you get out can't be stronger than what you put in).

Additionally, there is another soundness condition: for comonadic axes, upon construction the value must be stronger than the module. For monadic axes, upon destruction the value must be weaker than the module. The former is taken care of in `typemod.ml` `maybe_infer_modalities`. The latter is take care of by the definition of `apply` for `Diff`.

typing/mode.ml
riaqn: I'm not sure what you mean - this change is in its own commit, which is just a refactoring.

typing/mode.ml
riaqn: I improved the comment as beautiful as I can. 😄 
I also think again and decided to take `mm`'s ceil instead of zapping it to floor (reasons explained). The changes don't change behavior currently since all modules's mode are fixed already.

typing/mode.ml
goldfirere: Ah yes, not sure what I was thinking there.

typing/mode.ml
goldfirere: Oh yes I see now.

typing/mode.ml
goldfirere: Ah, yes, excellent. I now fully understand this and like the explanation. Thanks!

goldfirere (approved):

PR #3353, opened 2024-12-09
Fix soundness bug in capsule api exception handling

tdelvecchio-jsc:
In the current capsule api, if an exception crosses a capsule boundary, it gets wrapped in a `Contended` exception wrapper which applies the `contended` modality to the exception. This is used to prevent leaking access to data that is still in the capsule.

Leaking is still possible, however, as in this example:
```
type t = { mutable flag : bool }

exception E of (unit -> unit)

let () =
  let (P m) = create_with_mutex () in
  let ref_data = Data.create (fun () -> { flag = false }) in
  try
    Mutex.with_lock m (fun pw ->
      Data.iter pw (fun t -> raise (E (fun () -> t.flag <- true))) ref_data)
  with
  | E f -> f ()
;;
```

To guard against such examples, we take the approach of wrapping any escaping exceptions in a `Capsule.Encapsulated`, which carries the original `exn` in a `Capsule.Data.t` that corresponds to the capsule the exception originally came from. 

Since the `'k` type parameter of the capsule is made existential by the `exn` type constructor, we introduce the notion of a `Capsule.Name.t` to be able to re-associate the capsule with the mutex that guards it using a runtime check.

### DIFF
--- a/otherlibs/stdlib_alpha/capsule.ml
+++ b/otherlibs/stdlib_alpha/capsule.ml
@@ -85,6 +85,69 @@ external reraise : exn -> 'a @ portable @@ portable = "%reraise"
 
 exception Contended of exn @@ contended
 
+module Data = struct
+  type ('a, 'k) t : value mod portable uncontended
+
+  external unsafe_mk : 'a -> ('a, 'k) t @@ portable = "%identity"
+
+  external unsafe_get : ('a, 'k) t -> 'a @@ portable = "%identity"
+
+  let wrap _ t = unsafe_mk t
+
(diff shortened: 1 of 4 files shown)


lpw25 (approved):
LGTM. I have a small worry that we'll want to slightly change the definition of `name` -- not promising that there be only one name per-capsule -- when we add the `Key.t` interfaces, but we can cross that bridge when we come to it. We could also improve the performance by getting locks back down to just two words (or possible even 1 word with some effort), but again we can address that later.

PR #3354, opened 2024-12-09
Revert #11997

dkalinichenko-js:
Revert https://github.com/ocaml/ocaml/pull/11997, and translate structured constants into their representation at link time instead of compile time. This avoids a problem with the new structured constants introduced by JS extensions in bytecode, where previously they were marshaled with the upstream runtime.

Tested by existing tests + the rest of `or_null` PRs.

### DIFF
--- a/.depend
+++ b/.depend
@@ -2554,7 +2554,6 @@ bytecomp/dll.cmx : \
 bytecomp/dll.cmi :
 bytecomp/emitcode.cmo : \
     lambda/translmod.cmi \
-    bytecomp/symtable.cmi \
     typing/primitive.cmi \
     bytecomp/opcodes.cmi \
     utils/misc.cmi \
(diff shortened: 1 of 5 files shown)


TheNumbat (commented):
Confirmed this reverts 11997 (up to our changes). However, I don't think this fixes my problem on its own, because `initial_global_table`/`update_global_table` still translates float32 structured constants to an `Obj.t` at load time using the wrong version of `caml_float32_of_string`. 

I think we could now make the `caml_float32_of_string` in `float32_stubs.c` return a custom block that's compatible with the format in `float32.c`. That's kind of sketchy but should be ok since the target program uses a runtime/stdlib expecting that format.

(Also we will need to patch js_of_ocaml [here](https://github.com/ocsigen/js_of_ocaml/blob/master/compiler/lib/parse_bytecode.ml#L439))

TheNumbat (approved):

PR #3356, opened 2024-12-10
Ensure that all Bigarray memory managed by the GC is tracked by Memprof

stedolan:
There were two cases where this was not already the case:

  - Unix.map_file: Fixed by adding memprof calls to mmap_ba.c

  - User-allocated, GC-freed bigstrings: These are allocated by calls to caml_ba_alloc with CAML_BA_MANAGED, non-NULL data, and not CAML_BA_SUBARRAY (so new bigarrays, not slices)

    Runtime5 already handled this case correctly. The fix for runtime4 is to add a new call to memprof for the missing case.


lpw25 (approved):
LGTM

stedolan:
The build failure was #3358

PR #3360, opened 2024-12-10
Separate variants for boxed and unboxed kinds.

jvanburen:
This will help us once we have small "unboxed" integers that don't have a corresponding boxed integer type.

### DIFF
--- a/lambda/printlambda.ml
+++ b/lambda/printlambda.ml
@@ -19,6 +19,30 @@ open Primitive
 open Types
 open Lambda
 
+let unboxed_integer = function
+  | Unboxed_nativeint -> "unboxed_nativeint"
+  | Unboxed_int32 -> "unboxed_int32"
+  | Unboxed_int64 -> "unboxed_int64"
+
+let unboxed_float = function
+  | Unboxed_float64 -> "unboxed_float"
+  | Unboxed_float32 -> "unboxed_float32"
+
+let unboxed_vector = function
+  | Unboxed_vec128 -> "unboxed_vec128"
+
+let boxed_integer = function
+  | Boxed_nativeint -> "nativeint"
+  | Boxed_int32 -> "int32"
+  | Boxed_int64 -> "int64"
+
+let boxed_float = function
+  | Boxed_float64 -> "float"
+  | Boxed_float32 -> "float32"
+
+let boxed_vector = function
+  | Boxed_vec128 -> "vec128"
+
 let rec struct_const ppf = function
   | Const_base(Const_int n) -> fprintf ppf "%i" n
   | Const_base(Const_char c) -> fprintf ppf "%C" c
@@ -67,15 +91,6 @@ and struct_consts ppf (hd, tl) =
   in
   fprintf ppf "%a%a" struct_const hd sconsts tl
 
-let unboxed_float = function
-  | Pfloat64 -> "unboxed_float"
-  | Pfloat32 -> "unboxed_float32"
-
-let unboxed_integer = function
-  | Pint32 -> "unboxed_int32"
-  | Pint64 -> "unboxed_int64"
-  | Pnativeint -> "unboxed_nativeint"
-
 let rec scannable_product_element_kinds kinds =
   "[" ^ String.concat "; " (List.map scannable_product_element_kind kinds) ^ "]"
 
@@ -100,7 +115,7 @@ let array_kind = function
   | Pfloatarray -> "float"
   | Punboxedfloatarray f -> unboxed_float f
   | Punboxedintarray i -> unboxed_integer i
-  | Punboxedvectorarray Pvec128 -> "unboxed_vec128"
+  | Punboxedvectorarray v -> unboxed_vector v
   | Pgcscannableproductarray kinds ->
     "scannableproduct " ^ scannable_product_element_kinds kinds
   | Pgcignorableproductarray kinds ->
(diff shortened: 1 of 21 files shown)


middle_end/flambda2/to_cmm/to_cmm_primitive.ml
```diff
@@ -672,8 +672,8 @@ let binary_int_arith_primitive _env dbg (kind : K.Standard_int.t)
          if it doesn't support operations on 32-bit physical registers. There
          was a prototype developed of this but it was quite complicated and
          didn't get merged.) *)
-      C.sign_extend_32 dbg (C.safe_div_bi Unsafe x y Pint32 dbg)
-    | Mod -> C.sign_extend_32 dbg (C.safe_mod_bi Unsafe x y Pint32 dbg))
+      C.sign_extend_32 dbg (C.safe_div_bi Unsafe x y Boxed_int32 dbg)
+    | Mod -> C.sign_extend_32 dbg (C.safe_mod_bi Unsafe x y Boxed_int32 dbg))
```
mshinwell, line 676: These ones should be "Unboxed_" I think

mshinwell: (Here and below)

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
```diff
@@ -1487,11 +1501,15 @@ let convert_lprim ~big_endian (prim : L.primitive) (args : Simple.t list list)
     let arg1 = unbox_bint kind arg1 in
     let arg2 = unbox_bint kind arg2 in
     [ tag_int
-        (Binary (convert_boxed_integer_comparison_prim kind comp, arg1, arg2))
-    ]
+        (Binary
+           ( convert_unboxed_integer_comparison_prim
+               (Primitive.unbox_integer kind)
```
mshinwell, line 1506: I wonder if "Primitive.unboxed_integer" or something might be clearer - "unbox_integer" sounds like an operation on integers themselves.

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
```diff
@@ -1505,114 +1523,117 @@ let convert_lprim ~big_endian (prim : L.primitive) (args : Simple.t list list)
...
     [ tag_int
         (Binary
            ( Float_comp (Float64, Yielding_bool (convert_float_comparison comp)),
              arg1,
              arg2 )) ]
-  | Punbox_float Pfloat64, [[arg]] -> [Unary (Unbox_number Naked_float, arg)]
-  | Pbox_float (Pfloat64, mode), [[arg]] ->
+  | Punbox_float Boxed_float64, [[arg]] ->
+    [Unary (Unbox_number Naked_float, arg)]
+  | Pbox_float (Boxed_float64, mode), [[arg]] ->
```
mshinwell, line 1572: One could argue that this should be Pbox_float (Unboxed_float64, mode), maybe?

typing/primitive.ml
```diff
@@ -20,11 +20,13 @@ open Parsetree
 
 module String = Misc.Stdlib.String
 
-type boxed_integer = Pnativeint | Pint32 | Pint64
+type unboxed_integer =  Unboxed_int64 | Unboxed_nativeint | Unboxed_int32
```
mshinwell, line 23: Double space

typing/primitive.ml
```diff
@@ -330,38 +332,54 @@ let native_name p =
...
+  | Boxed_float32 -> Unboxed_float32
+  | Boxed_float64 -> Unboxed_float64
+
+let unbox_vector = function
+  | Boxed_vec128 -> Unboxed_vec128
+
+(* since these are just constant constructors, we can just use polymorphic
+   equality and comparison at no performance loss: *)
+let equal_unboxed_integer
+      ((Unboxed_int32 | Unboxed_nativeint | Unboxed_int64) as i1) i2 = i1 = i2
```
mshinwell, line 350: Or just: equal_unboxed_integer (i1 : unboxed_integer) i2 = i1 = i2?

lambda/printlambda.ml
```diff
@@ -627,15 +630,15 @@ let primitive ppf = function
...
-  | Pcompare_bints bi -> fprintf ppf "compare_bints %s" (boxed_integer_name bi)
+  | Pcompare_floats bf -> fprintf ppf "compare_floats %s" (boxed_float bf)
+  | Pcompare_bints bi -> fprintf ppf "compare_bints %s" (boxed_integer bi)
   | Poffsetint n -> fprintf ppf "%i+" n
   | Poffsetref n -> fprintf ppf "+:=%i"n
-  | Pfloatoffloat32 m -> print_boxed_float "float_of_float32" ppf Pfloat32 m
-  | Pfloat32offloat m -> print_boxed_float "float32_of_float" ppf Pfloat64 m
-  | Pintoffloat bf -> fprintf ppf "int_of_%s" (boxed_float_name bf)
+  | Pfloatoffloat32 m -> print_boxed_float "float_of_float32" ppf Boxed_float32 m
+  | Pfloat32offloat m -> print_boxed_float "float32_of_float" ppf Boxed_float64 m
```
mshinwell, line 638: Over 80 columns

lambda/translprim.ml
```diff
@@ -484,51 +484,51 @@ let lookup_primitive loc ~poly_mode ~poly_sort pos p =
         3)
     | "%array_safe_get_indexed_by_int64#" ->
       Primitive
-        ((Parrayrefs (gen_array_ref_kind mode, Punboxed_int_index Pint64, Mutable)), 2)
+        ((Parrayrefs (gen_array_ref_kind mode, Punboxed_int_index Unboxed_int64, Mutable)), 2)
```
mshinwell, line 487: There are some lines over 80 cols here but let's leave these for the moment, or they will conflict badly with one of the other array PRs in flight currently.

mshinwell (approved):

mshinwell:
I made a few suggestions, but this is ok to merge once some/all of those have been addressed as you see fit @jvanburen , without me having to review further.

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
jvanburen: I struggled with this for a bit, but I wanted to make sure we only apply these operation on values that can be boxed. otherwise we would be able to represent Pbox_int Unboxed_int16, which we don't want to be able to express. And we should keep it identical across int/float/vector

typing/primitive.ml
jvanburen: The match is so that you can verify that it's only constant constructors, and therefore polymorphic equality is appropriate

PR #3363, opened 2024-12-11
Static allocation / simplification for unboxed product arrays

mshinwell:
This allows arrays involving unboxed products to be statically allocated and in certain cases have their accesses simplified, just like for normal arrays.  This doesn't currently add to the testsuite; we can do that once #3317 has been merged with the augmented tests.  Previously for testing I used:
```
(* ... with the temporary used to communicate to the C runtime stub
   being constant, and statically allocated *)

let make_unboxed_tuple_vect_scannable () =
  make_unboxed_tuple_vect 42 #(1, 2.0, "3")

let make_unboxed_tuple_vect_ignorable () =
  make_unboxed_tuple_vect 42 #(#1.0, 2, #3L, true)

let _ = make_unboxed_tuple_vect_scannable ()
let _ = make_unboxed_tuple_vect_ignorable ()

(* ... with the temporary used to communicate to the C runtime stub
   being inconstant, and statically allocated *)

let[@inline] make_unboxed_tuple_vect_scannable_inconstant () =
  make_unboxed_tuple_vect 42 #(Sys.opaque_identity 1, 2.0, "3")

let[@inline] make_unboxed_tuple_vect_ignorable_inconstant () =
  make_unboxed_tuple_vect 42 #(#1.0, (Random.int [@inlined never]) 42, #3L, true)

let _ = make_unboxed_tuple_vect_scannable_inconstant ()
let _ = make_unboxed_tuple_vect_ignorable_inconstant ()

(* ... with the temporary used to communicate to the C runtime stub
   being dynamically allocated on the local stack. *)

let[@inline never] make_unboxed_tuple_vect_scannable_dynamic () =
  make_unboxed_tuple_vect 42 #(Sys.opaque_identity 1, 2.0, "3")

let[@inline never] make_unboxed_tuple_vect_ignorable_dynamic () =
  make_unboxed_tuple_vect 42 #(#1.0, (Random.int [@inlined never]) 42, #3L, true)

let _ = make_unboxed_tuple_vect_scannable_dynamic ()
let _ = make_unboxed_tuple_vect_ignorable_dynamic ()
```


middle_end/flambda2/types/provers.ml
```diff
@@ -583,20 +583,23 @@ let prove_unique_fully_constructed_immutable_heap_block env t :
...
-    | Region | Rec_info ->
-      Misc.fatal_errorf "Wrong element kind for array: %a" K.With_subkind.print
-        element_kind)
+  | Array { element_kinds = Ok element_kinds; _ } -> (
+    match element_kinds with
+    | [] | _ :: _ :: _ -> Invalid
+    | [element_kind] -> (
+      match K.With_subkind.kind element_kind with
+      | Value -> Proved false
+      | Naked_number Naked_float -> Proved true
```
lthls, line 598: That looks wrong. If you look at the comment in `flambda_primitive.mli`, on the `Unboxed_product` constructor of the `Array_kind.t` type, it says that arrays of unboxed products are never float arrays (in the sense of having `Double_array_tag`). So because here we can't distinguish between a legacy flat float array and an array of unboxed floats represented as unboxed products of size 1, we risk returning a wrong answer.

middle_end/flambda2/types/reify.ml
```diff
@@ -575,57 +577,99 @@ let reify ~allowed_if_free_vars_defined_in ~var_is_defined_at_toplevel
...
+          | [] -> Misc.fatal_error "Did not expect empty element kinds"
+          | _ :: _ :: _ -> (
+            let field_types_and_expected_kinds = List.combine fields kinds in
             match
               try_to_reify_fields env ~var_allowed alloc_mode
                 ~field_types_and_expected_kinds
             with
-            | Some fields -> Lift (Immutable_value_array { fields })
+            (* XXX mshinwell: We need to convince ourselves that there is no
+               problem with making different decisions based on immediacy
```
TheNumbat, line 611: I assume this means static arrays of immediates will become non-scannable blocks? I can't think of any reason that'd be a problem, but may be worth considering stuff like [this](https://github.com/ocaml-multicore/saturn/blob/306bea620cc0cfcc33639c45a56da59add9bdd92/src/ws_deque.ml#L45)

middle_end/flambda2/types/provers.ml
mshinwell: The new code here should also be better.

mshinwell:
Setting back to draft whilst work is in progress.

PR #3367, opened 2024-12-12
Implement %array_element_size_in_bytes

mshinwell:
Example use:
```
external[@layout_poly] array_element_size_in_bytes : ('a : any_non_null) . 'a array -> int =
  "%array_element_size_in_bytes"

let () =
  Printf.printf "%d\n" (array_element_size_in_bytes [| #(#1., #(#2l, #3.4, #5L), #(1, 2)) |])

let () =
  Printf.printf "%d\n" (array_element_size_in_bytes [| #1. |])

let () =
  Printf.printf "%d\n" (array_element_size_in_bytes [| #1.s |])

let () =
  Printf.printf "%d\n" (array_element_size_in_bytes [| #1l |])

let () =
  Printf.printf "%d\n" (array_element_size_in_bytes [| #1L |])
```

Output:
```
48
8
4
4
8
```

The new functions at the bottom of `lambda.ml` are taken from #3317 .


lpw25:
It's a shame that we need to create a dummy value for this. We could probably avoid that by making a predefined type `'a size` and then making the primitive be:
```ocaml
external[@layout_poly] array_element_size_in_bytes : ('a : any_non_null) .unit -> 'a size =
  "%array_element_size_in_bytes"
```
but then the primitive would have to inspect the type of it's return and only work if it was `'a size`. I don't think we do anything like that for any existing primitives.

We a lot more work we could give `size` a new kind `size 'k` where `'k` is the kind of it's input type. Then we would only be inspecting the kind of the result as we do in other cases. That is satisfyingly clean, but almost certainly a bad idea.

ccasin:
> It's a shame that we need to create a dummy value for this. We could probably avoid that by making a predefined type `'a size` and then making the primitive be:
> 
> ```ocaml
> external[@layout_poly] array_element_size_in_bytes : ('a : any_non_null) .unit -> 'a size =
>   "%array_element_size_in_bytes"
> ```
> 
> but then the primitive would have to inspect the type of it's return and only work if it was `'a size`. I don't think we do anything like that for any existing primitives.
> 
> We a lot more work we could give `size` a new kind `size 'k` where `'k` is the kind of it's input type. Then we would only be inspecting the kind of the result as we do in other cases. That is satisfyingly clean, but almost certainly a bad idea.

Indeed - this was debated a fair bit in advance of this PR (see the internal google doc).  We'll eventually want something like `'a size`, but it's nontrivially more work than this for the reasons you've outlined.  Also, with a type annotation or when the type is otherwise clear from inference, it should possible to use the empty array rather than constructing a dummy value of the relevant type.

ccasin (commented):
@mshinwell The approach here looks good - thanks for doing this.

I've pushed a few commits with minor adjustments and tests, please take a look through them.

The tests rely on makearray_dynamic and therefore do not yet work on native code.  We should wait for that before merging this, so it can be tested properly.  The test passes on bytecode, but since everything has the same size there that doesn't tell us much - I won't be surprised if there are some typos in the test that need fixing once we can run it on native code.

mshinwell:
@ccasin please approve

ccasin (approved):

PR #3368, opened 2024-12-12
Add Atomic.compare_exchange

TheNumbat:
The atomic interface has `compare_and_set`, which returns a bool indicating whether the operation was successful.

It's often the case that you want to know the previous value upon failure, so this PR adds `compare_exchange`, which returns the previous value. Its semantics are otherwise the same.

Only adds a simple test, since the implementation is now used by `compare_and_set`.

### DIFF
--- a/runtime/memory.c
+++ b/runtime/memory.c
@@ -400,6 +400,28 @@ CAMLprim value caml_atomic_cas (value ref, value oldv, value newv)
   }
 }
 
+CAMLprim value caml_atomic_compare_exchange (value ref, value oldv, value newv)
+{
+  if (caml_domain_alone()) {
+    value* p = Op_val(ref);
+    if (*p == oldv) {
+      *p = newv;
+      write_barrier(ref, 0, oldv, newv);
+      return oldv;
+    } else {
+      return *p;
+    }
+  } else {
+    atomic_value* p = &Op_atomic_val(ref)[0];
+    int cas_ret = atomic_compare_exchange_strong(p, &oldv, newv);
+    atomic_thread_fence(memory_order_release); /* generates `dmb ish` on Arm64*/
+    if (cas_ret) {
+      write_barrier(ref, 0, oldv, newv);
+    }
+    return oldv;
+  }
+}
+
 CAMLprim value caml_atomic_fetch_add (value ref, value incr)
 {
   value ret;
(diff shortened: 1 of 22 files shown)


stdlib/atomic.mli
```diff
@@ -61,6 +61,11 @@ val exchange : 'a t -> 'a -> 'a
     otherwise. *)
 val compare_and_set : 'a t -> 'a -> 'a -> bool
 
+(** [compare_exchange r seen v] sets the new value of [r] to [v] only
+    if its current value is physically equal to [seen] -- the comparison
+    and the set occur atomically. Returns the previous value. *)
```
glittershark, line 66: is this interface right? I guess I would have expected this to return a `Result.t` or something like that, based on the API of https://doc.rust-lang.org/std/sync/atomic/struct.AtomicUsize.html#method.compare_exchange and friends.

glittershark: maybe that's something we can just do at the call site, though.

runtime4/misc.c
```diff
@@ -252,11 +252,20 @@ CAMLprim value caml_atomic_load(value ref)
...
+    return oldv;
+  } else {
+    return *p;
+  }
+}
+
+CAMLprim value caml_atomic_cas(value ref, value oldv, value newv)
+{
+  if (caml_atomic_compare_exchange(ref, oldv, newv) == oldv) {
     return Val_int(1);
```
glittershark, line 269: not changed here, but... should this have been using `Val_bool` instead?

runtime/memory.c
```diff
@@ -376,27 +376,34 @@ CAMLprim value caml_atomic_exchange (value ref, value v)
...
-      return Val_int(0);
     }
+    return oldv;
+  }
+}
+
+CAMLprim value caml_atomic_cas (value ref, value oldv, value newv)
+{
+  if (caml_atomic_compare_exchange(ref, oldv, newv) == oldv) {
+    return Val_int(1);
```
glittershark, line 404: Should this be `Val_bool`?

glittershark (changes requested):
is this missing assembly for the primop in `backend/x86...`?

runtime4/misc.c
TheNumbat: sure

runtime/memory.c
TheNumbat: sure

TheNumbat:
This actually doesn't need a backend implementation - `compare_and_set` also always calls into the runtime so it can check domain_alone/write barrier/etc. The `lock_cmpxchg` in the x86 backend is only used for particular intrinsics. 

We should consider compiling the atomic primitives directly to instructions when their argument is an immediate, though. (In another PR)

glittershark (approved):
lgtm

PR #3369, opened 2024-12-12
Line pragmas and error reporting

ccasin:
When should a line number pragma affect an error message?  Here's an unfortunate example of how things work today:
```
$ cat foo.mli
# 1 "bar.mli"
val x : string
$ ocamlc -c -o bar.cmi foo.mli
$ cat foo.ml
# 1 "bar.ml"
let x = 42
$ ocamlc -c -cmi-file bar.cmi -o bar.cmo foo.ml
File "foo.ml", line 1:
Error: The implementation foo.ml does not match the interface bar.cmi: 
       Values do not match: val x : int is not included in val x : string
       The type int is not compatible with the type string
       File "bar.mli", line 1, characters 0-14: Expected declaration
       File "bar.ml", line 1, characters 4-5: Actual declaration
```

This error message mentions both "foo.ml" and "bar.ml" - which is it?  This is causing some problems for folks working on the build system, who would like to move files around in some new cases and use line number pragmas to get the compiler to report errors about the original files.

There is some logic to the current behavior - the error is about the whole file.  But I think it's basically a bug - the intent of the line number directive was for errors to be issued as if the file had a different name.

After discussion with @lukemaurer, a plausible path forward seemed to be: if the line number pragma appears on the first line of the file and says we're on the first line of another file, errors that are about the whole file should use the revised name.

The first commit here implements that.  Most whole-file errors refer to the ref [Location.input_name] to find the name of the file, so I now set that when we see a relevant line number directive during lexing. That handles the first reference to "foo.ml" in the above error.  For the second reference, I changed some uses of `Unit_info.source_file` to use this ref instead.

In the second commit, I looked through the other uses of `Unit_info.source_file` and changed the ones that are obviously just about errors or warnings.  In some cases I wasn't sure and left it alone - @lukemaurer, please advise about the call to `check_argument_type_if_given` in `typemod.ml`.

I'm far from certain this is the right fix to the build systems problem or the right semantics for line number directives - open to suggestions.  Marking this PR as a draft for now as it seems worth testing on our codebase before merging, which I'll do tomorrow.

Review: @lukemaurer


parsing/lexer.mll
```diff
@@ -851,6 +851,14 @@ and directive already_consumed = parse
             directive_error lexbuf "line number out of range"
               ~already_consumed ~directive
         | line_num ->
+           (* If we're at the beginning of the file and claiming to be at the
+              beginning of a different file, then update [Location.input_name]
+              for better errors. *)
```
stedolan, line 856: There are two reasonable definitions of "we're at the beginning of the file":
  1. It's the first byte of the file
  2. Either (1), or a line directive has just reset the line number to 1.

Which is intended here?

ccasin: Right, sorry, I meant to leave a note on the PR about this.

The former seems preferable (at least, if this heuristic is reasonable at all, which I'm on the fence about!).  But the latter was more obvious to me how to implement and seemed fine for testing this on the tree, which is my next step.

PR #3370, opened 2024-12-12
Add primitive types for unboxed small integers

jvanburen:
Native operations are not implemented on these values, but it should provide a good starting point for it.
Unboxed arrays for the new integers are also not implemented.


middle_end/flambda2/to_cmm/to_cmm_primitive.ml
```diff
@@ -556,67 +575,116 @@ let unary_float_arith_primitive _env dbg width op arg =
   | Float32, Abs -> C.float32_abs ~dbg arg
   | Float32, Neg -> C.float32_neg ~dbg arg
 
+module Cmm_kind = struct
```
Gbury, line 578: A few remarks: we already have a plethora of similar types/modules to express some notion of cmm type/kind, so 1) this name if not descriptive enough to know what it means and why we need it, 2) this needs more documentation, 3) do we really need yet again another cmm type/kind/machtype/exttype/whatever ? 4) assuming we really need this new notion, this code does not have its place in this file: smart constructors functions for cmm expression should be in `cmm_helpers.ml` or in a separate file.

Note: we need to do a pass on all of the cmm type/kinds that we currently have to see how we could simplify things a bit, this is currently getting out of hand and confusing: we're currently stuck in an intermediate situation between the origin/vanilla cmm / upstream compiler where there tends to be few different types, vs the flambda2 way where we define as many types as we want so that we can be really specific, and I think that in conjunction with all of the subtelty of cmm, this is does not help.

middle_end/flambda2/to_cmm/to_cmm_shared.ml
```diff
@@ -224,6 +232,12 @@ let const_static cst =
     (* Just in case of future big endian support, this is also written
        explicitly in two halves. *)
     [cint32 i; cint32 0l]
+  | Naked_int16 i ->
+    (* In keeping with Naked_float32, this is padded to 8 bytes, but it's not
+       obvious to me why we are doing this instead of requiring alignment on
+       other values that need it. *)
```
Gbury, line 238: I think it's less a question of other values requiring some kind of alignment, and more the fact that the assumption that most things take exactly one memory word is baked in so many places (well, except for cmm static consts, but they must be compatible with all of the rest of the code), that it's easier to be explicit here. I think it'll be easier when we want to try packing: we'll only have to remove the explicit padding that is inserted here.

middle_end/flambda2/to_cmm/to_cmm_shared.ml
```diff
@@ -389,12 +409,16 @@ let make_update env res dbg ({ kind; stride } : Update_kind.t) ~symbol var
         match kind with
         | Pointer -> Word_val
         | Immediate -> Word_int
+        | Naked_int8 ->
+          if stride = Arch.size_addr then Word_int else Byte_signed
```
Gbury, line 413: Considering the definitions above, the `else` branch of this condition is dead code. Since that code will then not be used and tested for a while, i'd rather see it fail at compile time, rather than have dead code that may work in the future if the correct changes are made (it will be simpler to write the correct code at that point).

middle_end/flambda2/to_cmm/to_cmm_shared.ml
```diff
@@ -389,12 +409,16 @@ let make_update env res dbg ({ kind; stride } : Update_kind.t) ~symbol var
...
+        | Naked_int16 ->
+          if stride = Arch.size_addr then Word_int else Sixteen_signed
         | Naked_int32 ->
           (* Cmm expressions representing int32 values are always sign extended.
              By using [Word_int] in the "fields" cases (see [Update_kind],
              above) we maintain the convention that 32-bit integers in 64-bit
              fields are sign extended. *)
-          if stride > 4 then Word_int else Thirtytwo_signed
+          if stride = Arch.size_addr then Word_int else Thirtytwo_signed
```
Gbury, line 421: this change does not look like it is related to the addition of `int8` and `int16`, could you explain why it's there ?

backend/cmm_helpers.ml
```diff
@@ -750,34 +753,41 @@ let mod_int c1 c2 is_safe dbg =
    occur, in which case we force x / -1 = -x and x mod -1 = 0. (PR#5513). *)
 
 let is_different_from x = function
-  | Cconst_int (n, _) -> n <> x
-  | Cconst_natint (n, _) -> n <> Nativeint.of_int x
+  | Cconst_int (n, _) -> Nativeint.of_int n <> x
+  | Cconst_natint (n, _) -> n <> x
```
Gbury, line 757: why is this change here ? does this actually fix a bug ? if so, that deserves a comment

backend/cmm_helpers.ml
```diff
@@ -750,34 +753,41 @@ let mod_int c1 c2 is_safe dbg =
...
-  bind "divisor" c2 (fun c2 ->
-      bind "dividend" c1 (fun c1 ->
-          let c = mkop c1 c2 is_safe dbg in
+let safe_divmod_bi mkop mkm1 ?(dividend_cannot_be_min_int = false) is_safe
+    dividend divisor dbg =
+  bind "divisor" divisor (fun divisor ->
+      bind "dividend" dividend (fun dividend ->
+          let dividend_cannot_be_min_int =
+            dividend_cannot_be_min_int
+            || dividend |> is_different_from Nativeint.min_int
```
Gbury, line 766: the use of `|>` here is gratuitous and reduces readability

backend/cmm_helpers.ml
```diff
@@ -1262,35 +1272,109 @@ let addr_array_initialize arr ofs newval dbg =
...
+    ->
+    Some (x, mask)
+  | Cop (Cand, ([x; Cconst_int (mask, _)] | [Cconst_int (mask, _); x]), _) ->
+    Some (x, Nativeint.of_int mask)
+  | _ -> None
+
+(* [low_bits ~bits x] is a value which agrees with x on at least the low [bits]
+   bits *)
+let rec low_bits ~bits dbg x =
+  let arch_bits = Arch.size_int * 8 in
```
Gbury, line 1286: Given all of the other assertions in this file, wouldn't it be simpler to have a hardcoded `arch_bits = 64` here ?

backend/cmm_helpers.ml
```diff
@@ -1403,6 +1487,31 @@ let get_field_computed imm_or_ptr mutability ~block ~index dbg =
...
 (* Getters for unboxed int fields *)
 
+let get_field_unboxed_int8 mutability ~block ~index dbg =
+  let memory_chunk = Byte_signed in
+  if Arch.big_endian
+  then
+    Misc.fatal_error
+      "Unboxed int8 fields only supported on little-endian architectures";
+  (* CR layouts v5.1: We'll need to vary log2_size_addr to efficiently pack
+   * int8s *)
```
Gbury, line 1497: To support packing, it won't be just varying `log2_size_addr` that will need to be done, more generally, we'll need to refactor most of the code to go from having field indexes everywhere as arguments, to have byte offsets (and the associated changes to `array_indexing`), but that's another topic.

mshinwell:
@jvanburen @Gbury is this ready to be approved?

typing/typeopt.ml
```diff
@@ -860,6 +870,10 @@ let[@inline always] rec layout_of_const_sort_generic ~value_kind ~error
     Lambda.Punboxed_float Unboxed_float64
   | Base Word when Language_extension.(is_at_least Layouts Stable) ->
     Lambda.Punboxed_int Unboxed_nativeint
+  | Base Bits8 when Language_extension.(is_at_least Layouts Stable) ->
```
dkalinichenko-js, line 873: At least `Layouts Beta`?

typing/typeopt.ml
```diff
@@ -155,6 +155,9 @@ let classify ~classify_product env loc ty sort : _ classification =
   | Tconstr (p, _args, _abbrev) ->
       if Path.same p Predef.path_float then Float
       else if Path.same p Predef.path_lazy_t then Lazy
+      else if Path.same p Predef.path_int
+           || Path.same p Predef.path_int8
+           || Path.same p Predef.path_int16 then Int
```
dkalinichenko-js, line 160: Why add those lines? Shouldn't those be always gc-ignorable and handled by the case above?

typing/jkind.ml
```diff
@@ -1390,6 +1396,28 @@ module Const = struct
         name = "word mod everything"
       }
 
+    let bits8 =
+      { jkind = mk_jkind (Base Bits8) ~mode_crossing:true ~nullability:Non_null;
```
dkalinichenko-js, line 1400: Shouldn't mode-cross.

typing/jkind.ml
```diff
@@ -1390,6 +1396,28 @@ module Const = struct
...
+      }
+
+    (* CR layouts v3: change to [Maybe_null] when separability is implemented. *)
+    let kind_of_unboxed_int8 =
+      { jkind = mk_jkind (Base Bits8) ~mode_crossing:true ~nullability:Non_null;
+        name = "bits8 mod everything"
+      }
+
+    let bits16 =
+      { jkind = mk_jkind (Base Bits16) ~mode_crossing:true ~nullability:Non_null;
```
dkalinichenko-js, line 1411: Shouldn't mode-cross.

typing/jkind.ml
```diff
@@ -1760,7 +1794,11 @@ module Const = struct
       (jkind : 'd t) =
     let rec scan_layout (l : Layout.Const.t) : Language_extension.maturity =
       match l, Mod_bounds.nullability jkind.mod_bounds with
-      | (Base (Float64 | Float32 | Word | Bits32 | Bits64 | Vec128) | Any), _
+      | ( ( Base
```
dkalinichenko-js, line 1797: `Beta`?

otherlibs/stdlib_beta/int16_u.ml
```diff
@@ -0,0 +1,87 @@
...
+
+let[@inline always] minus_one () = of_int16 Int16.minus_one
+
+let[@inline always] max_int () = of_int16 Int16.max_int
+
+let[@inline always] min_int () = of_int16 Int16.min_int
+
+let[@inline always] neg x = of_int16 (Int16.neg (to_int16 x))
+
+let[@inline always] add x y = of_int16 (Int16.add (to_int16 x) (to_int16 y))
```
dkalinichenko-js, line 41: Do those definitions get optimized to direct unboxed operations? Just want to make sure it's happening.

typing/predef.ml
```diff
@@ -543,7 +547,10 @@ let add_small_number_beta_extension_types add_type env =
   let _, add_type = mk_add_type add_type in
   env
   |> add_type ident_int8 ~jkind:Jkind.Const.Builtin.immediate
+       ~unboxed_jkind:Jkind.Const.Builtin.bits8
```
dkalinichenko-js, line 550: `kind_of_unboxed_int8`?

typing/predef.ml
```diff
@@ -543,7 +547,10 @@ let add_small_number_beta_extension_types add_type env =
...
   |> add_type ident_int16 ~jkind:Jkind.Const.Builtin.immediate
+       ~unboxed_jkind:Jkind.Const.Builtin.bits16
```
dkalinichenko-js, line 552: same here

typing/jkind.ml
```diff
@@ -1390,6 +1396,28 @@ module Const = struct
...
       }
 
+    let bits8 =
+      { jkind = mk_jkind (Base Bits8) ~mode_crossing:true ~nullability:Non_null;
+        name = "bits8"
+      }
+
+    (* CR layouts v3: change to [Maybe_null] when separability is implemented. *)
+    let kind_of_unboxed_int8 =
+      { jkind = mk_jkind (Base Bits8) ~mode_crossing:true ~nullability:Non_null;
```
dkalinichenko-js, line 1406: `~nullability:Non_null` is set on unboxed types so we can still put them into arrays. Perhaps it makes sense to define `bits8`/`bits16` as `Maybe_null` and thus rule out their arrays on the type level?

otherlibs/stdlib_beta/int16_u.ml
jvanburen: they do not, that's the responsibility of a future PR

dkalinichenko-js:
Could you add basic tests for the `bits8`/`bits16` layouts? Properties like (the absence of) sublayouts, mode-crossing of both the basic kinds and the kinds of `int8#`/`int16#`, etc.

utils/misc.ml
```diff
@@ -25,6 +25,9 @@ let fatal_errorf fmt =
 
 let fatal_error msg = fatal_errorf "%s" msg
 
+let unboxed_small_int_arrays_are_not_implemented () =
```
dkalinichenko-js, line 28: Usually we just inline the error into the usage site.

dkalinichenko-js:
The middle-end tests look fine for `beta` (don't take this as an approval, ideally @mshinwell or someone else more familiar with middle/backend parts would look at them), but you might want to look at `testsuite/tests/typing-layouts-bits64/basics.ml` and others to see what we want to test.

typing/jkind.ml
jvanburen: why not, doesn't  `int` mode-cross?

typing/jkind.ml
jvanburen: why not, doesn't  `int` mode-cross?

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
```diff
@@ -989,6 +1022,10 @@ let bigarray_box_or_tag_raw_value_to_read kind alloc_mode =
     fun arg -> H.Unary (Box_number (Naked_float32, alloc_mode), Prim arg)
   | Naked_number Naked_float ->
     fun arg -> H.Unary (Box_number (Naked_float, alloc_mode), Prim arg)
+  (* The following tagging operations for small integers are valid because they
```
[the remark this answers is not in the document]
jvanburen: how? is there even a way to emit this primitive?

lambda/lambda.ml
```diff
@@ -334,6 +334,8 @@ type primitive =
   | Pobj_magic of layout
   | Punbox_float of boxed_float
   | Pbox_float of boxed_float * locality_mode
+  | Puntag_int of unboxed_integer
```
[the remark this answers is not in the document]
jvanburen: puntag_int will logically truncate if the source doesn't fit into the destination. In the register it will end up being sign-extend it since that's how values are represented.

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
```diff
@@ -1711,6 +1756,20 @@ let convert_lprim ~big_endian (prim : L.primitive) (args : Simple.t list list)
             ( Naked_vec128,
               Alloc_mode.For_allocations.from_lambda mode ~current_region ),
           arg ) ]
+  | Puntag_int dst, [[arg]] ->
+    [ Unary
+        ( Num_conv
```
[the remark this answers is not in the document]
jvanburen: no, it truncates, which is intended. In the cmm it will end up being sign-extended since that's how small integer values are required to be represented in cmm.

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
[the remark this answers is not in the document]
dkalinichenko-js: There are primitives like `%caml_ba_set_1`, but I don't see any unboxed number tests using those. In that case, could you instead raise an error saying int8/int16 bigarrays are unsupported?

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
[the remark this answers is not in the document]
dkalinichenko-js: Could you add tests for this behavior?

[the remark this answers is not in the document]
jvanburen: added

lambda/lambda.ml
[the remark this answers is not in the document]
jvanburen: since this is translated to Num_conv, it should be tested by the same tests at the bottom of lib_smallint/test_int{8,16}_u.ml

jvanburen:
@dkalinichenko-js I added layout tests based on bits32 and 64

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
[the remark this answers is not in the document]
jvanburen: I changed it to Num_conv, which should be more robust / clearly correct. How's that?

middle_end/flambda2/terms/code_size.ml
```diff
@@ -350,8 +386,7 @@ let unary_prim_size prim =
         | Naked_vec128s | Unboxed_product _ ) ->
       array_length_size
     | Array_kind (Naked_int32s | Naked_float32s) ->
-      (* There is a dynamic check here to see if the array has an odd or even
-         number of elements *)
+      (* There is a dynamic check here to see the number of elements *)
```
lthls, line 389: Why was this comment updated ?

middle_end/flambda2/terms/static_const.ml
```diff
@@ -189,15 +189,15 @@ let [@ocamlformat "disable"] print ppf t =
       Flambda_colours.pop
       (Format.pp_print_list
         ~pp_sep:(fun ppf () -> Format.pp_print_string ppf "@; ")
-        (Or_variable.print (fun ppf n -> Format.fprintf ppf "%ld" n)))
+        (Or_variable.print Numeric_types.Int32.print))
```
lthls, line 192: I don't think this file needed to be modified, but the patch looks good so it's fine.

middle_end/flambda2/to_cmm/to_cmm_primitive.ml
```diff
@@ -53,19 +53,19 @@ let unbox_number ~dbg kind arg =
   | Naked_float -> C.unbox_float dbg arg
   | Naked_float32 -> C.unbox_float32 dbg arg
   | Naked_vec128 -> C.unbox_vec128 dbg arg
-  | Naked_int32 | Naked_int64 | Naked_nativeint ->
-    let primitive_kind = K.Boxable_number.primitive_kind kind in
-    C.unbox_int dbg primitive_kind arg
+  | Naked_int32 -> C.unbox_int dbg Boxed_int32 arg
+  | Naked_int64 -> C.unbox_int dbg Boxed_int64 arg
+  | Naked_nativeint -> C.unbox_int dbg Boxed_nativeint arg
```
lthls, line 58: Unless I'm mistaken, this change (and the one a few lines below) is not necessary.

middle_end/flambda2/types/flambda2_types.mli
```diff
@@ -606,6 +628,8 @@ type boxed_or_tagged_number = private
 val prove_is_a_boxed_or_tagged_number :
   Typing_env.t -> t -> boxed_or_tagged_number proof_of_property
 
+val prove_nothing : Typing_env.t -> t -> _ proof_of_property
```
lthls, line 631: I don't think it makes sense to expose this function. It doesn't seem to be used at all.

middle_end/flambda2/types/grammar/type_grammar.ml
```diff
@@ -3268,46 +3453,79 @@ let these_naked_vec128s vs =
...
+  | Rec_info _ | Region _ ->
     Misc.fatal_errorf "Type of wrong kind for [box_float]: %a" print t
 
+let tag_int8 (t : t) : t =
+  match t with
+  | Naked_int8 _ ->
+    non_null_value
+      (Variant
+         { is_unique = false;
+           immediates = Known t;
```
lthls, line 3475: This is wrong. The expected kind for `immediates` is `Naked_immediates`, which (as this PR stands) is incompatible with `Naked_int8`.

lthls (changes requested):
I've reviewed most of the Flambda2 changes.
In addition to the individual comments, I'd like to suggest to not create dedicated `Naked_int8` and `Naked_int16` kinds, but instead have them as particular cases of `Naked_immediates` (with a `width` field with possible values `Full | Int8 | Int16`).
It should not only make this patch simpler and easier to review, it also allows to remove almost all the changes to the `types` folder (it's fine to represent all small ints as `Naked_immediates` in the types).

typing/typedecl.ml
```diff
@@ -4483,15 +4494,19 @@ let report_error ppf = function
         Style.inline_code "[@@untagged]"
   | Cannot_unbox_or_untag_type Unboxed ->
       fprintf ppf "@[Don't know how to unbox this type.@ \
-                   Only %a, %a, %a, %a, vector primitives, and@ \
+                   Only %a, %a, %a, %a, %a, %a, vector primitives, and@ \
```
dkalinichenko-js, line 4497: I think we should pick one of either unboxed or untagged for int8/int16. Untagged seems like a better fit.

typing/typeopt.ml
```diff
@@ -860,6 +867,10 @@ let[@inline always] rec layout_of_const_sort_generic ~value_kind ~error
     Lambda.Punboxed_float Unboxed_float64
   | Base Word when Language_extension.(is_at_least Layouts Stable) ->
     Lambda.Punboxed_int Unboxed_nativeint
+  | Base Bits8 when Language_extension.(is_at_least Layouts Beta) ->
```
dkalinichenko-js, line 870: Shouldn't it be `small_numbers` and not layouts?

dkalinichenko-js: It's an important distinction, since `layouts` is marked as erasable in public release while those aren't erasable.

middle_end/flambda2/kinds/flambda_kind.ml
```diff
@@ -444,13 +464,6 @@ module Boxable_number = struct
     | Naked_nativeint -> Naked_number Naked_nativeint
     | Naked_vec128 -> Naked_number Naked_vec128
 
-  let primitive_kind t : Primitive.boxed_integer =
-    match t with
-    | Naked_vec128 | Naked_float | Naked_float32 -> assert false
-    | Naked_int32 -> Boxed_int32
-    | Naked_int64 -> Boxed_int64
-    | Naked_nativeint -> Boxed_nativeint
-
```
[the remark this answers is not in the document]
jvanburen: because it's unused now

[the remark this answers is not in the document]
jvanburen: it was only used in one place and this PR removed that call site

middle_end/flambda2/numbers/numeric_types.ml
```diff
@@ -37,41 +37,61 @@ module Int = struct
...
-    else i
+    let of_int i =
+      let extra_bits = Sys.int_size - num_bits in
+      (i lsl extra_bits) asr extra_bits
 
-  let lower_int64 = Int64.neg (Int64.shift_left Int64.one 15)
+    let to_int i = i
 
-  let upper_int64 = Int64.sub (Int64.shift_left Int64.one 15) Int64.one
+    let to_float = Float.of_int
```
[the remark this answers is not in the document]
jvanburen: it's needed by the signature in middle_end/flambda2/simplify/number_adjuncts.ml

middle_end/flambda2/terms/code_size.ml
```diff
@@ -103,27 +107,46 @@ let arith_conversion_size src dst =
...
+      ( Naked_int8 | Naked_int16 | Naked_int32 | Naked_int64 | Naked_nativeint
+      | Naked_immediate ) ) ->
+    1
+  | ( Naked_int8,
+      ( Naked_int16 | Naked_int32 | Naked_int64 | Naked_nativeint
+      | Naked_immediate ) ) ->
+    1
+  | Naked_int16, (Naked_int32 | Naked_int64 | Naked_nativeint | Naked_immediate)
+    ->
     1
```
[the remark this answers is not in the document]
jvanburen: I think I was wrong about this - since cmm values are all sign-extended (which is a weird choice but whatever), it should be 0 instructions to upconvert

middle_end/flambda2/to_cmm/to_cmm_primitive.ml
jvanburen: It gets rid of `K.Boxable_number.primitive_kind` which was a partial match with an assert in a function that doesn't even say it can raise.
I thought it was clearer this way than adding more to the `assert false` case

typing/typeopt.ml
jvanburen: both, right?

middle_end/flambda2/to_cmm/to_cmm_primitive.ml
lthls: But you wouldn't have to add anything to `K.Boxable_number.primitive_kind` anyway, since int8/int16 are not in `K.Boxable_number`. So while I can agree with you that this wasn't the most beautiful code, this part of the patch looks unrelated to the subject of this PR.

jvanburen: ah you're right. alright, i'll change it back

jvanburen: yes, you're right, i'll change it back

jvanburen:
For posterity, we decided that this isn't the right PR to start doing it differently, and we can come back to it later in a separate PR to reduce the verbosity of flambda2 kinds.

testsuite/tests/typing-layouts-bits16/test_int16_u.ml
```diff
@@ -0,0 +1,318 @@
...
+  test_binary    "logxor"              Int16.logxor              Int16_u.logxor;
+  test_unary     "lognot"              Int16.lognot              Int16_u.lognot;
+  test_shift     "shift_left"          Int16.shift_left          Int16_u.shift_left;
+  test_shift     "shift_right"         Int16.shift_right         Int16_u.shift_right;
+  test_shift     "shift_right_logical" Int16.shift_right_logical Int16_u.shift_right_logical;
+  test_unary_to  "of_int"              Int16.of_int              Int16_u.of_int               int_input;
+  test_unary_of  "to_int"              Int16.to_int              Int16_u.to_int               int_result;
+  test_unary_to  "of_float"            Int16.of_float            Int16_u.of_float             float_input;
+  test_unary_of  "to_float"            Int16.to_float            Int16_u.to_float             float_result;
+  (* test_unary_to  "of_string"           Int16.of_string           Int16_u.of_string            int16_string_input; *)
```
dkalinichenko-js, line 311: Why is `of_string` commented out?

dkalinichenko-js: Oh, it's just not implemented yet, nvm.

testsuite/tests/typing-layouts-bits8/unboxed_bits8s.ml
```diff
@@ -0,0 +1,421 @@
...
+let test3 () =
+  (* Test f3 *)
+  let steps = Array.init 10 (fun _ -> (Stdlib_beta.Int8.of_int 0)) in
+  let five_times_three = f3 5 (Stdlib_beta.Int8_u.of_int 3) steps in
+  print_int8u "Test 3, 5 * 3: " (five_times_three ());
+  Array.iteri (Printf.printf "  Test 3, step %d: %d\n") (Array.map Int8.to_int steps);
+
+  (* Test f3_manyargs
+
+          (1 + 2 + 3 + 5 + 8) = 19
```
dkalinichenko-js, line 265: Could you document the overflow happening here with int8?

testsuite/tests/typing-layouts-bits16/unboxed_bits16s.reference
```diff
@@ -0,0 +1,115 @@
...
+  Test 4, step 3: 6
+  Test 4, step 4: 3
+  Test 4, step 5: 0
+  Test 4, step 6: 0
+  Test 4, step 7: 0
+  Test 4, step 8: 0
+  Test 4, step 9: 0
+Test 5, 3 + 2 + 1: 6
+Test 6, 3: 3
+Test 6, 16: 32
```
dkalinichenko-js, line 113: Similarly this should be "Test 6, 32"

testsuite/tests/typing-layouts-bits16/c_api.reference
```diff
@@ -0,0 +1,6 @@
+int16# -> int16#, ~42: -43
+int16 -> int16#, ~(-100): 99
+int16# -> int16, ~255: -256
+(int16[@unboxed]) -> int16#, ~1024: -1025
```
dkalinichenko-js, line 4: Probably best to use non-overflowing values in this test.

tools/dune
```diff
@@ -151,7 +151,7 @@
 
 (executable
  (name merge_archives)
- (modes native)
+ (modes native byte)
```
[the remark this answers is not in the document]
jvanburen: probably accidental from a merge conflict or smth

lambda/lambda.mli
```diff
@@ -374,7 +376,7 @@ and extern_repr =
   | Same_as_ocaml_repr of Jkind.Sort.Const.t
   | Unboxed_float of boxed_float
   | Unboxed_vector of boxed_vector
-  | Unboxed_integer of boxed_integer
+  | Unboxed_integer of unboxed_integer
```
lthls, line 414: Is there a reason for this change ? It makes the diff in `Lambda_to_flambda` more complex than it should be.

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
```diff
@@ -34,6 +34,18 @@ let convert_integer_comparison_prim (comp : L.integer_comparison) :
...
+  | Unboxed_int8, Clt -> Int_comp (Naked_int8, Yielding_bool (Lt Signed))
+  | Unboxed_int8, Cgt -> Int_comp (Naked_int8, Yielding_bool (Gt Signed))
+  | Unboxed_int8, Cle -> Int_comp (Naked_int8, Yielding_bool (Le Signed))
+  | Unboxed_int8, Cge -> Int_comp (Naked_int8, Yielding_bool (Ge Signed))
+  | Unboxed_int16, Ceq -> Int_comp (Naked_int16, Yielding_bool Eq)
+  | Unboxed_int16, Cne -> Int_comp (Naked_int16, Yielding_bool Neq)
+  | Unboxed_int16, Clt -> Int_comp (Naked_int16, Yielding_bool (Lt Signed))
+  | Unboxed_int16, Cgt -> Int_comp (Naked_int16, Yielding_bool (Gt Signed))
+  | Unboxed_int16, Cle -> Int_comp (Naked_int16, Yielding_bool (Le Signed))
+  | Unboxed_int16, Cge -> Int_comp (Naked_int16, Yielding_bool (Ge Signed))
```
lthls: This function should have been factored into two independent matches. Maybe now would be the right time for that ?

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
```diff
@@ -984,6 +1017,12 @@ let bigarray_box_or_tag_raw_value_to_read kind alloc_mode =
     fun arg -> H.Unary (Box_number (Naked_float32, alloc_mode), Prim arg)
   | Naked_number Naked_float ->
     fun arg -> H.Unary (Box_number (Naked_float, alloc_mode), Prim arg)
+  | Naked_number Naked_int8 ->
+    fun arg ->
+      H.Unary (Num_conv { src = Naked_int8; dst = Tagged_immediate }, Prim arg)
+  | Naked_number Naked_int16 ->
+    fun arg ->
+      H.Unary (Num_conv { src = Naked_int16; dst = Tagged_immediate }, Prim arg)
```
lthls, line 1122: Why are we using `Num_conv` instead of a variant of `Tag_int`/`Untag_int` ?

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
```diff
@@ -1008,6 +1047,14 @@ let bigarray_unbox_or_untag_value_to_store kind =
...
   | Naked_number Naked_float ->
     fun arg -> H.Prim (Unary (Unbox_number Naked_float, arg))
+  | Naked_number Naked_int8 ->
+    fun arg ->
+      H.Prim
+        (Unary (Num_conv { src = Tagged_immediate; dst = Naked_int8 }, arg))
+  | Naked_number Naked_int16 ->
+    fun arg ->
+      H.Prim
+        (Unary (Num_conv { src = Tagged_immediate; dst = Naked_int16 }, arg))
```
lthls, line 1158: (Same comment as above)

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
```diff
@@ -1709,6 +1762,20 @@ let convert_lprim ~big_endian (prim : L.primitive) (args : Simple.t list list)
...
+              dst = standard_int_or_float_of_unboxed_integer dst
+            },
+          arg ) ]
+  | Ptag_int src, [[arg]] ->
+    [ Unary
+        ( Num_conv
+            { src = standard_int_or_float_of_unboxed_integer src;
+              dst = Tagged_immediate
+            },
+          arg ) ]
```
lthls, line 1927: Same as above: we have `Tag_int` and `Untag_int` primitives, it's strange to be converting `Ptag_int` and `Puntag_int` to something else.

middle_end/flambda2/kinds/flambda_kind.ml
```diff
@@ -1101,15 +1147,17 @@ module Standard_int_or_float = struct
...
       | Naked_nativeint -> Format.pp_print_string ppf "Naked_nativeint"
 
-    let compare = Stdlib.compare
+    let compare = (Stdlib.compare : t -> t -> int)
 
-    let equal t1 t2 = compare t1 t2 = 0
+    let equal = (Stdlib.( = ) : t -> t -> bool)
 
-    let hash = Hashtbl.hash
+    let hash = (Hashtbl.hash : t -> int)
```
lthls, line 1246: It's not clear why these changes are part of this PR, but these changes look good on their own so fine.

middle_end/flambda2/numbers/numeric_types.ml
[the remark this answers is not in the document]
lthls: I think that's a bad thing (we never convert `Int8` to `Float`, so we shouldn't need this function). But I see that it is a consequence of previous bad decisions around the `Num_conv` primitive, and not specific to this PR, so I will fix it later.

middle_end/flambda2/simplify/unboxing/build_unboxing_denv.ml
```diff
@@ -187,6 +188,12 @@ let rec denv_of_decision denv ~param_var (decision : U.decision) : DE.t =
       T.boxed_float_alias_to ~naked_float (Alloc_mode.For_types.unknown ())
     in
     denv_of_number_decision K.naked_float shape param_var naked_float denv
+  | Unbox (Number (Naked_int8, { param = naked_int8; args = _ })) ->
+    let shape = T.tagged_int8_alias_to ~naked_int8 in
+    denv_of_number_decision K.naked_int8 shape param_var naked_int8 denv
+  | Unbox (Number (Naked_int16, { param = naked_int16; args = _ })) ->
+    let shape = T.tagged_int16_alias_to ~naked_int16 in
+    denv_of_number_decision K.naked_int16 shape param_var naked_int16 denv
```
lthls, line 198: I will need @Gbury to confirm, but I'm not sure it makes sense to unbox/untag small integers. That does not necessarily need to be part of this PR, but in the future I think the `Number` case of `unboxing_decision` should take a `Boxable_number.t` and not a `Naked_number_kind.t`.

middle_end/flambda2/simplify/unboxing/unboxing_epa.ml
```diff
@@ -256,6 +256,8 @@ and compute_extra_args_for_one_decision_and_use_aux ~(pass : U.pass) rewrite_id
   | Unbox (Number (Naked_float, epa)) ->
     compute_extra_arg_for_number Naked_float Unboxers.Float.unboxer epa
       rewrite_id ~typing_env_at_use arg_being_unboxed
+  | Unbox (Number ((Naked_int8 | Naked_int16), _epa)) ->
+    U.Do_not_unbox U.Not_enough_information_at_use
```
lthls: I don't think `Not_enough_information_at_use` is right. What about `Not_beneficial` ?

middle_end/flambda2/to_cmm/to_cmm_primitive.ml
```diff
@@ -580,7 +594,15 @@ let unary_int_arith_primitive _env dbg kind op arg =
...
-      C.bswap16 arg dbg
+      C.Scalar_type.Integral.static_cast arg ~dbg ~src:naked_immediate
+        ~dst:naked_int16
+      |> (fun arg -> C.bbswap Unboxed_int16 arg dbg)
+      |> C.zero_extend ~bits:16 ~dbg
+    | Naked_int8 -> arg
+    | Naked_int16 ->
+      (* Byte swaps of small integers need a sign-extension in order to match
+         the Lambda semantics (where the swap might affect the sign). *)
+      C.sign_extend (C.bbswap Unboxed_int16 arg dbg) ~bits:16 ~dbg
```
lthls, line 744: I don't think all these changes are needed. `Swap_byte_endianness` is a weird primitive, that is only defined on a subset of the naked types with tricky semantics, so I would be in favour of not supporting the small integer versions and not changing the other ones.

middle_end/flambda2/types/grammar/type_grammar.ml
```diff
@@ -3268,46 +3453,79 @@ let these_naked_vec128s vs =
...
+  | Rec_info _ | Region _ ->
     Misc.fatal_errorf "Type of wrong kind for [box_float]: %a" print t
 
+let tag_int8 (t : t) : t =
+  match t with
+  | Naked_int8 _ ->
+    non_null_value
+      (Variant
+         { is_unique = false;
+           immediates = Unknown;
```
lthls: This could be improved by converting the underlying `Int8.Set.t` to a `Targetint_31_63.Set.t`, but I don't think it's worth the trouble. The real solution is to get rid of these redundant kinds anyway.

middle_end/flambda2/types/grammar/type_grammar.ml
```diff
@@ -3351,6 +3570,12 @@ let boxed_float32_alias_to ~naked_float32 =
 let boxed_float_alias_to ~naked_float =
   box_float (Naked_float (TD.create_equals (Simple.var naked_float)))
 
+let tagged_int8_alias_to ~naked_int8 =
+  tag_int8 (Naked_int8 (TD.create_equals (Simple.var naked_int8)))
+
+let tagged_int16_alias_to ~naked_int16 =
+  tag_int16 (Naked_int16 (TD.create_equals (Simple.var naked_int16)))
```
lthls: With the current implementation, it does not really make sense to call `TD.create_equals`, as the alias type created this way will be dropped by the `tag_intn` functions. In fact, because these functions do not (and cannot) actually create aliases, I don't think they should exist at all.

middle_end/flambda2/types/provers.ml
```diff
@@ -477,6 +502,8 @@ let prove_is_a_boxed_or_tagged_number_value _env
     Proved (Boxed (alloc_mode, Naked_vec128, contents_ty))
   | Mutable_block _ | Closures _ | String _ | Array _ -> Unknown
 
+let prove_nothing _env _ : _ proof_of_property = Unknown
```
lthls: I don't think we should expose this function. We could expose a way to construct `Unknown` proofs directly if needed, but I suspect we don't need this feature at all and can just tweak the places where it's used.

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
jvanburen: why not? What's the advantage to using Tag_int/Untag_int?

lthls: The semantics is well-defined, unlike `Num_conv`.

jvanburen: How is Num_conv is not well-defined? It's defined in to_cmm...

lambda/lambda.mli
jvanburen: It's so that people can use small integers when interfacing with C? It's also used in the follow up PR.

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
jvanburen: If anything I would think we could remove tag_int/Untag_int since they're totally subsumed by Num_conv

testsuite/tests/typing-layouts-bits16/c_api.reference
jvanburen: fixed in bits8, these don't overflow in bits16

middle_end/flambda2/to_cmm/to_cmm_primitive.ml
jvanburen: It is fully defined on all of the naked integer types in to_cmm

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
ccasin: FWIW, I tend to agree with @lthls here that being more explicit about what kind of conversion this is would be clearer

lthls (approved):

PR #3371, opened 2024-12-12
Convert float32 constants to int32 in first stage compiler

TheNumbat:
Currently, the stdlib is compiled using flambda-backend compiled with upstream. This compiler does not have access to `float32.c`, so represents float32s as int32s, and needs to convert to float32 when loading a constant. The installed compiler is then linked against the flambda-backend runtime, so can use the proper float32 functions.
This PR detects which stage we're in and decides whether to convert from int32.

### DIFF
--- a/bytecomp/bytegen.ml
+++ b/bytecomp/bytegen.ml
@@ -748,7 +748,8 @@ module Storer =
    Result = list of instructions that evaluate exp, then perform cont. *)
 
 (* We cannot use the [float32] type in the compiler. *)
-external float32_of_string : string -> int32 = "compiler_float32_of_string"
+external float32_is_stage1 : unit -> bool = "caml_float32_is_stage1"
+external float32_of_string : string -> Obj.t = "caml_float32_of_string"
 
 let rec comp_expr stack_info env exp sz cont =
   check_stack stack_info sz;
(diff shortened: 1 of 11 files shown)


bytecomp/bytegen.ml
```diff
@@ -768,6 +772,11 @@ let rec comp_expr stack_info env exp sz cont =
           Koffsetclosure(pos - env_pos) :: cont
         | exception Not_found -> not_found ()
       end
+  | Lconst (Const_base (Const_float32 f |
```
dkalinichenko-js, line 775: I think this will miss a more complicated constant that contains a `Const_float32` inside it. But perhaps it's impossible to come up with such a constant in bytecode.

dkalinichenko-js:
Could you add tests for this?

bytecomp/bytegen.ml
TheNumbat: Good point - const blocks may contain float32s, though `Stdlib_stable` doesn't currently create any. Added logic for it.

TheNumbat:
Added some tests to `float32_builtin.ml` and confirmed they use this code path. 
`float32_lib.ml` also tests the constants in `Stdlib_stable`, and I manually checked the jsoo output for it

dkalinichenko-js:
The test appears to work even before the other changes in this PR, even after comparing with 64-bit float constants.

TheNumbat:
Updated the tests so they fail without this PR - when run by the native bytecode interpreter, they need to check for the correct float32 custom ops rather than the numeric value.

dkalinichenko-js (approved):

PR #3372, opened 2024-12-12
Fix soundness bug in include functor

ccasin:
Include functor is expansive - see title and test.

Review: @lpw25 ?


lpw25 (approved):
LGTM

PR #3374, opened 2024-12-13
Allow retyping as-patterns that contain existentials

TheNumbat:
As discussed. Added test demonstrating a packed version vs. using `t` directly, which was previously rejected.


lpw25 (approved):
I'm fairly sure this is sound and there aren't any tests showing otherwise, so let's merge it. Probably worth submitting it upstream too to avoid diverging unnecessarily.

PR #3375, opened 2024-12-13
Unboxed product exhaustiveness

ccasin:
Exhaustiveness checking for unboxed product patterns is currently busted (the analysis thinks they are not exhaustive even when they are).

This PR fixes that, in three steps:
- The first commit adds tuple and record tests showing the issue.
- The second commit fixes the issue.
- The third commit improves another pattern matching function where I noticed we had missed similar cases. The function (`Parmatch.discr_pat`) is used for generating the counterexample in exhaustiveness warnings, and this commit adds a test showing it's still suboptimal in the unboxed records case, to be improved in a future PR.

Review: @rtjoa could you take a look at this?  Hopefully it's clear enough from the surrounding code, but if you want a reference for the complete exhaustiveness algorithm it's [here](http://moscova.inria.fr/~maranget/papers/warn/warn.pdf).

### DIFF
--- a/typing/parmatch.ml
+++ b/typing/parmatch.ml
@@ -177,9 +177,11 @@ let all_coherent column =
           | Const_unboxed_float32 _
           | Const_string _), _ -> false
       end
-    | Tuple l1, Tuple l2 -> l1 = l2
+    | Tuple l1, Tuple l2 ->
+      List.equal (Option.equal String.equal) l1 l2
     | Unboxed_tuple l1, Unboxed_tuple l2 ->
-      List.equal (fun (lbl1, _) (lbl2, _) -> lbl1 = lbl2) l1 l2
+      List.equal
+        (fun (lbl1, _) (lbl2, _) -> Option.equal String.equal lbl1 lbl2) l1 l2
     | Record (lbl1 :: _), Record (lbl2 :: _) ->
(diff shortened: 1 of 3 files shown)


typing/parmatch.ml
```diff
@@ -431,10 +431,16 @@ let simple_match d h =
   | Constant c1, Constant c2 -> const_compare c1 c2 = 0
   | Lazy, Lazy -> true
   | Record _, Record _ -> true
+  | Record_unboxed_product _, Record_unboxed_product _ -> true
   | Tuple len1, Tuple len2 -> len1 = len2
```
ncik-roberts, line 435: We check labels in the `Unboxed_tuple` case but use poly equal in this case.

ccasin: Indeed. I didn't want to duplicate the sad use of poly equal, and also didn't want to create unnecessary diffs with upstream by changing the existing code.  Though I guess the latter is pretty thinly motivated when we've changed the line before and the line after, and when the thing called `len` has had its type silently changed, so I'll push a similar improvement to the legacy code in a minute.  Thanks.

rtjoa (approved):
Looks good.

Noting the irony of a pattern matching exhaustiveness bug that could have been surfaced by the exhaustiveness check (had wildcard patterns been used in fewer places).

PR #3376, opened 2024-12-16
Avoid traversing C parts of frame pointer chain when reallocating stack

stedolan:
Port upstream 13635. In configuration `--enable-runtime5 --enable-stack-checks --enable-frame-pointers`, the new `qsort.ml` test segfaults before this patch and passes afterwards.

---

When the OCaml stack grows we need to rewrite frame pointers (if enabled) to point to the new stack.

However, when using a C library that was not compiled with frame pointers enabled, we cannot assume that there is an unbroken chain of frame pointers through both the OCaml and C parts of the stack. Doing so leads to segfaults.

Instead, we note that the only frame pointers that can point to OCaml stacks (the ones that need updating) are those already on OCaml stacks, plus the first ones pushed after any OCaml->C calls. These can be found by traversing the struct c_stack_link chain, without needing to traverse any intervening C frames.

This imposes a new constraint on the runtime assembly stubs: after switching to C they must not push anything to the stack before calling a C function. This was already true for all but caml_c_call_stack_args. Enforcing this invariant for caml_c_call_stack_args is straightforward enough, and simplifies the DWARF backtrace logic.

For arm64, a side-effect of this change is that DWARF backtraces now work on stacks containing calls to caml_c_call_stack_args, which were broken before. (Tested with macos lldb)

### DIFF
--- a/runtime/amd64.S
+++ b/runtime/amd64.S
@@ -186,7 +186,7 @@
      https://github.com/hjl-tools/x86-psABI/wiki/x86-64-psABI-1.0.pdf */
 
 #define DW_CFA_def_cfa_expression 0x0f
-#define DW_REG_rbx                3
+#define DW_REG_rbp                6
 #define DW_REG_rsp                7
 #define DW_REG_r13                13
 #define DW_OP_breg                0x70
@@ -232,7 +232,12 @@
 /* Stack switching operations */
 /******************************************************************************/
 
-/* Switch from OCaml to C stack. Clobbers %r10, %r11. */
+/* Switch from OCaml to C stack. Clobbers %r10, %r11.
+
+   If a C function is called which might call back into OCaml,
+   then nothing may be pushed to the C stack between SWITCH_OCAML_TO_C
+   and the next C call. (This is to ensure frame pointers are correctly
+   maintained if the stack is reallocated) */
 #ifdef ASM_CFI_SUPPORTED
 #define SWITCH_OCAML_TO_C_CFI                                   \
         CFI_REMEMBER_STATE;                                     \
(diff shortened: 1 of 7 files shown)


runtime/amd64.S
```diff
@@ -806,43 +811,54 @@ CFI_STARTPROC
...
-#ifdef ASM_CFI_SUPPORTED
-        .cfi_escape DW_CFA_def_cfa_expression, 5,           \
-          /* %rbp points to the c_stack_link structure */   \
-          DW_OP_breg + DW_REG_rbx, Cstack_sp, DW_OP_deref,  \
-          DW_OP_plus_uconst, RETADDR_ENTRY_SIZE
-#endif
     /* Make the alloc ptr available to the C code */
         movq    %r15, Caml_state(young_ptr)
+    /* Copy the arguments and call */
+        C_call  (GCALL(caml_c_call_copy_stack_args))
```
NickBarnes, line 817: Why do this in a separate function, rather than inline here?

runtime/amd64.S
```diff
@@ -232,7 +232,12 @@
 /* Stack switching operations */
 /******************************************************************************/
 
-/* Switch from OCaml to C stack. Clobbers %r10, %r11. */
+/* Switch from OCaml to C stack. Clobbers %r10, %r11.
+
+   If a C function is called which might call back into OCaml,
+   then nothing may be pushed to the C stack between SWITCH_OCAML_TO_C
```
NickBarnes, line 238: I don't understand this comment. In the new code (caml_c_call_copy_stack_args), new things are exactly "pushed to the C stack between SWITCH_OCAML_TO_C and the next C call". Do you mean the OCaml stack?
Same remark applies to the matching comment in arm64.S.

NickBarnes (approved):
This is OK; I know the ARM64 stubs less well than the AMD64, so not so certain there but it looks OK, modulo my one remark about comments (amd64.S:238). Hurrah it removes a nasty function from fiber.c.
Wishlist: we could do with a lot more diagrams of stack layouts, with comments about register contents, at various tricky points in the assembly stubs. But not now.

runtime/amd64.S
stedolan: > In the new code (caml_c_call_copy_stack_args), new things are exactly "pushed to the C stack between SWITCH_OCAML_TO_C and the next C call".

No, they aren't. The next C call after `SWITCH_OCAML_TO_C` is in fact the call to `caml_c_call_copy_stack_args`, which happens before any pushing. This is why that function is an out-of-line call, to ensure that the last frame pointer that points into the OCaml stack is at a constant offset from the stack position at `SWITCH_OCAML_TO_C` time.

NickBarnes (approved):

PR #3379, opened 2024-12-16
Reset domain lock in caml_thread_reinitialize

TheNumbat:
Systhreads overrides `caml_atfork_hook` with `caml_thread_reinitialize`, so it also needs to reinitialize the domain lock. This is necessary to release the master lock in the child after forking. Resolves the CR regarding this.


stedolan:
I don't think this is right - our locking scheme for systhreads is currently different from upstream. (See the big comment on `domain_lockmode`).

Possibly the right thing to do here is reset to LOCKMODE_STARTUP?

mshinwell (approved):
Approving based on discussion between Max and Stephen.

PR #3380, opened 2024-12-16
Add Capsule.Condition

TheNumbat:
Adds condition variables to the capsule API, which are already used in `tdelvecchio-portable-stdlib-5.2.0minus-4`. They have the same implementation as `Stdlib.Condition`, but work with `Capsule.Mutex`.


dkalinichenko-js (approved):
LGTM.

PR #3381, opened 2024-12-16
Add Capsule.protect

TheNumbat:
Adds `Capsule.protect`, which is already used in `tdelvecchio-portable-stdlib-5.2.0minus-4`. 

An equivalent function could be implemented with `key`s, but there's currently no way to defer creating the mutex until after the given function raises.


dkalinichenko-js (approved):
LGTM.

PR #3382, opened 2024-12-16
Allow Capsule mutexes in runtime4

TheNumbat:
Currently, `Capsule.Mutex` and `Capsule.Condition` are allowed in runtime4, as they defer to the upstream implementations of `caml_ml_mutex` and `caml_ml_condition`. However, they require initializing the systhreads library before creating either a mutex or a condition, which makes it difficult to provide a version of base/core that works in both 4 and 5. 

Instead, we may wrap the sync primitives in separate capsule functions that implement `Capsule.Mutex` as no-ops in runtime4. This is fine since there is only ever one domain. However, the capsule mutex will not provide synchronization across systhreads.

We don't need capsule conditions in runtime4, so their implementations raise, and same for rwlocks. I've updated the external symbol names for all three to indicate that they are specifically for capsules.


lpw25:
> However, they require initializing the systhreads library before creating either a mutex or a condition

Have we considered changing this instead? We could probably just move the whole code of the systhreads library into the runtime if we wanted at this point.

mshinwell:
I'm unsure it's worth the trouble of changing the way systhreads works given the current place on the timeline for moving to all-runtime5.  Instead I think it would be better to concentrate resources on the latter work.

TheNumbat:
Yeah, changing systhreads would probably be better overall, but I think this is fine given we're moving to all-runtime5 soon.

dkalinichenko-js (approved):
LGTM.

PR #3383, opened 2024-12-16
Portable Effects

TheNumbat:
Adds a mirror effect API that lets you work with portable continuations created from portable functions. 
Refer to the portable effects doc for discussion of why this is safe.

Draft since this relies on enabling alpha mode extensions. It will also likely be subsumed by @lpw25's new effect API, but we might need this version for the initial rollout.

Note that safely enabling portable effects will also require adding `unyielding` to the DLS/Capsule APIs.

### DIFF
--- a/stdlib/dune
+++ b/stdlib/dune
@@ -20,6 +20,8 @@
   (modules_before_stdlib camlinternalFormatBasics camlinternalAtomic))
  (flags
   (:standard
+   -extension-universe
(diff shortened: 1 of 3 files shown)


TheNumbat:
New effect-safe API merged; also this version probably wasn't sound since `perform` is not restricted to portable -> contended in a portable continuation.

PR #3384, opened 2024-12-17
Better hugepage alignment of stacks and heap

stedolan:
This patch improves the memory mapping of the runtime in two ways (which I noticed after #3351):

  - **Increase minimum pool allocation**: The major heap now grows in at least 8MB chunks. (Previously, it grew slowly at the start. This isn't very useful since the minor heap and default stack are both 8MB, so we may as well use larger heap chunks and get better hugepage alignment)
  - **Improve stack mapping logic**: Since `caml_mem_map` does hugepage alignment these days, there is no longer a need for custom 2MB alignment in `fiber.c`. (In passing, also fix a couple of other issues: error handling logic was wrong, and there's no need to mprotect memory you're about to unmap)

The result is that the OCaml parts of the memory map now have pleasant alignment. (Note the 8184 kB stack allocation, ending on a hugepage boundary. This means that the top of the stack will be hugepage-aligned).
```
$ cat /proc/1095532/smaps | grep -A1 OCaml
7f43b2200000-7f43b2a00000 rw-p 00000000 00:00 0                          [anon:OCaml: major heap (chunk 0)]
Size:               8192 kB
--
7f43b2a00000-7f43b2a01000 rw-p 00000000 00:00 0                          [anon:OCaml: stack (tid 1095532)]
Size:                  4 kB
--
7f43b2a01000-7f43b2a02000 ---p 00000000 00:00 0                          [anon:OCaml: stack (tid 1095532)]
Size:                  4 kB
--
7f43b2a02000-7f43b3200000 rw-p 00000000 00:00 0                          [anon:OCaml: stack (tid 1095532)]
Size:               8184 kB
--
7f43b3200000-7f43b3a00000 rw-p 00000000 00:00 0                          [anon:OCaml: minor heap 0]
Size:               8192 kB
--
7f43b3a00000-7f43f3200000 ---p 00000000 00:00 0                          [anon:OCaml: minor reservation]
Size:            1040384 kB
```

### DIFF
--- a/runtime/fiber.c
+++ b/runtime/fiber.c
@@ -140,6 +140,9 @@ static uintnat round_up_p2(uintnat x, uintnat p2)
    well-defined), but other fields are uninitialised */
 Caml_inline struct stack_info* alloc_for_stack (mlsize_t wosize)
 {
+  /* Ensure 16-byte alignment of the [struct stack_handler*] (e.g. for arm64) */
+  const int stack_alignment = 16;
+
 #ifdef USE_MMAP_MAP_STACK
   size_t len = sizeof(struct stack_info) +
                sizeof(value) * wosize +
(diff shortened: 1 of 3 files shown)


mshinwell (approved):

mshinwell:
@stedolan checked my code review commit.

PR #3385, opened 2024-12-17
Add attributes to (unsafely) skip jkind check 

glittershark:
Add a pair of attributes, `[@@unsafe_allow_any_kind_in_intf]` and
`[@@unsafe_allow_any_kind_in_impl]`, which if set on both the impl and the intf
respectively, skip checking the jkind of the type in a signature against the
jkind of the type in a struct entirely. This is a more-selective version of the
`--allow-illegal-crossing` flag, and likely eventually will subsume it.

Also, adds a warning if these attributes are set unnecessarily. This is plausibly a bad idea, since it can trigger in odd places (such as signature inclusion).

### DIFF
--- a/testsuite/tests/typing-layouts/allow_any.ml
+++ b/testsuite/tests/typing-layouts/allow_any.ml
@@ -43,7 +43,6 @@ module Mismatched_with_both_attrs : sig type t : float64 end
 |}]
 
 (* If we set the attributes but *don't* get a kind mismatch, we ought to be fine *)
-(* CR-soon aspsmith: this should be a warning *)
 module Matching : sig
   type t : value
   [@@unsafe_allow_any_kind_in_impl "I love segfaults"]
@@ -52,6 +51,12 @@ end = struct
   [@@unsafe_allow_any_kind_in_intf "I love segfaults"]
 end
 [%%expect{|
+Lines 2-3, characters 2-54:
+2 | ..type t : value
+3 |   [@@unsafe_allow_any_kind_in_impl "I love segfaults"]
+Warning 212 [unnecessary-allow-any-kind]: [@@allow_any_kind_in_intf] and [@@allow_any_kind_in_impl] set on a
+type, but the kind matches. The attributes can be removed.
+
 module Matching : sig type t end
 |}]
 
@@ -163,4 +168,23 @@ module M2 : S2 = M1
 [%%expect{|
 module M1 : sig type t = string end
 module M2 : S2
+|}]
+
+module type S3 = sig
+  type t : value
+  [@@unsafe_allow_any_kind_in_impl]
+end
+
+module M3 : S3 = M1
+(* CR aspsmith: This is somewhat unfortunate, if S3 and M1 are defined far away, but it's
+   unclear how to squash the warning *)
+[%%expect{|
+module type S3 = sig type t end
+Lines 2-3, characters 2-35:
+2 | ..type t : value
+3 |   [@@unsafe_allow_any_kind_in_impl]
+Warning 212 [unnecessary-allow-any-kind]: [@@allow_any_kind_in_intf] and [@@allow_any_kind_in_impl] set on a
+type, but the kind matches. The attributes can be removed.
+
+module M3 : S3
 |}]
--- a/testsuite/tests/typing-layouts/allow_any.ml
+++ b/testsuite/tests/typing-layouts/allow_any.ml
@@ -1,0 +1,166 @@
+(* TEST
+ flags = "-extension layouts_beta";
+ expect;
+*)
+
+(* Baseline: if the jkind doesn't match, we should get an error. *)
+module Mismatched_no_attrs : sig
+  type t : float64
+end = struct
+  type t = string
+end
+[%%expect{|
+Lines 3-5, characters 6-3:
+3 | ......struct
+4 |   type t = string
+5 | end
+Error: Signature mismatch:
+       Modules do not match:
+         sig type t = string end
+       is not included in
+         sig type t : float64 end
+       Type declarations do not match:
+         type t = string
+       is not included in
+         type t : float64
+       The layout of the first is value
+         because it is the primitive type string.
+       But the layout of the first must be a sublayout of float64
+         because of the definition of t at line 2, characters 2-18.
+|}]
+
+(* On the other hand, if we set the correct attributes on both the impl and the intf, we
+   shouldn't get an error (though, obviously, this is completely unsound!) *)
+module Mismatched_with_both_attrs : sig
+  type t : float64
+  [@@unsafe_allow_any_kind_in_impl "I love segfaults"]
+end = struct
+  type t = string
+  [@@unsafe_allow_any_kind_in_intf "I love segfaults"]
+end
+[%%expect{|
+module Mismatched_with_both_attrs : sig type t : float64 end
+|}]
+
+(* If we set the attributes but *don't* get a kind mismatch, we ought to be fine *)
+(* CR-soon aspsmith: this should be a warning *)
+module Matching : sig
+  type t : value
+  [@@unsafe_allow_any_kind_in_impl "I love segfaults"]
+end = struct
+  type t = string
+  [@@unsafe_allow_any_kind_in_intf "I love segfaults"]
+end
+[%%expect{|
+module Matching : sig type t end
+|}]
+
+(* If the attr is only on the signature we should get an error *)
+module Mismatched_with_attr_on_intf : sig
+  type t : float64
+  [@@unsafe_allow_any_kind_in_impl "I love segfaults"]
+end = struct
+  type t = string
+end
+[%%expect{|
+Lines 4-6, characters 6-3:
+4 | ......struct
+5 |   type t = string
+6 | end
+Error: Signature mismatch:
+       Modules do not match:
+         sig type t = string end
+       is not included in
+         sig type t : float64 end
+       Type declarations do not match:
+         type t = string
+       is not included in
+         type t : float64
+       The layout of the first is value
+         because it is the primitive type string.
+       But the layout of the first must be a sublayout of float64
+         because of the definition of t at lines 2-3, characters 2-54.
+|}]
+
+(* If the attr is only on the struct we should get an error *)
+module Mismatched_with_attr_on_impl : sig
+  type t : float64
+end = struct
+  type t = string
+  [@@unsafe_allow_any_kind_in_intf "I love segfaults"]
+end
+[%%expect{|
+Lines 3-6, characters 6-3:
+3 | ......struct
+4 |   type t = string
+5 |   [@@unsafe_allow_any_kind_in_intf "I love segfaults"]
+6 | end
+Error: Signature mismatch:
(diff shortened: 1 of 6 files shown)


testsuite/tests/typing-layouts/allow_any.ml
```diff
@@ -0,0 +1,180 @@
...
+
+module type S2 = sig
+  type t : float64
+  [@@unsafe_allow_any_kind_in_impl]
+end
+
+module F (M : S1) : S2 = struct
+  type t = M.t
+  [@@unsafe_allow_any_kind_in_intf]
+end
```
goldfirere, line 133: I don’t think this test is as interesting as others: it still tests a `struct` against a `sig`. Instead, I was thinking of something like

```
module type S1 = sig
  type t : value
  [@@unsafe_allow_any_kind_in_intf]
end

module F (X : S1) : S2 = X (* should work *)
module F (X : S2) : S1 = X (* shouldn’t work *)
```

testsuite/tests/typing-layouts/allow_any.ml
goldfirere, line 1: We decided this feature would be temporary, right? If it were permanent, I’d want to think more about subsumption of module types. (We want to make sure that if `S1 < S2`, then any `M : S1` also type-checks as `M : S2`.) But I think it’s OK to skip worrying about this for something temporary, as a missing subsumption check can’t make this (more) unsafe.

goldfirere, line 1: Might be good to check to make sure we get warnings for an unused attribute. For example, I’d want a warning for `struct type t = string [@unsafe_…] end`. Note the one `@`! This will attach to the `string` instead of the type declaration and will be ignored.

goldfirere (approved):

testsuite/tests/typing-layouts/allow_any.ml
```diff
@@ -0,0 +1,180 @@
...
+end = struct
+  type t = string
+  [@@unsafe_allow_any_kind_in_intf "I love segfaults"]
+end
+[%%expect{|
+module Mismatched_with_both_attrs : sig type t : float64 end
+|}]
+
+(* If we set the attributes but *don't* get a kind mismatch, we ought to be fine *)
+(* CR-soon aspsmith: this should be a warning *)
```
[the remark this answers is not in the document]
glittershark: oops!

testsuite/tests/typing-layouts/allow_any.ml
```diff
@@ -0,0 +1,180 @@
...
+   shouldn't get an error (though, obviously, this is completely unsound!) *)
+module Mismatched_with_both_attrs : sig
+  type t : float64
+  [@@unsafe_allow_any_kind_in_impl "I love segfaults"]
+end = struct
+  type t = string
+  [@@unsafe_allow_any_kind_in_intf "I love segfaults"]
+end
+[%%expect{|
+module Mismatched_with_both_attrs : sig type t : float64 end
```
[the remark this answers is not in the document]
glittershark: I figured out how to do that (breadcrumbs: we have to add something to `Outcometree.out_type_decl` to remember what we did) but along the way convinced myself somewhat that we shouldn't; specifically, I'm not sure it makes the most sense to *tell* clients of a module "we have lied about the kind", especially when the whole idea is to use this feature to expose safe interfaces to unsafe stuff.

I was also talking with @tdelvecchio-jsc about this feature, and he felt that maybe we should just make this an attribute on the impl (with a kind annotation) and leave it off the intf entirely, and do this when checking kinds on type declarations rather than when checking kinds during module inclusion. I was somewhat convinced, for the same reason, but don't feel terribly strongly about it

testsuite/tests/typing-layouts/allow_any.ml
glittershark: I think we should probably plan for this to be permanent? I'm not sure @tdelvecchio-jsc was convinced by our "everything we do here should be done via modalities and magic instead" argument.

glittershark:
summarizing our discussion about this this morning (please double check me @goldfirere / others to make sure I didn't get anything wrong here):

We want to rework this feature, such that:
- the attribute is only specified on the type declaration, not the type in the signature
- the attribute only allows "lying" about the parts of the jkind which we believe are at least _sometimes_ safe to lie about, specifically the thing we're now calling "modal bounds" and "with-bounds", and only for the mode axes which seem safe to lie about (eg, just portability and contention, not allocation-affecting things like locality, uniqueness, or externality)
- @ageorg29 thinks it should be possible to write down some "rules" for when it is ok to do this
- this morally ends up on the type_kind, not on the type_declaration (though, probably not syntactically)

I don't think we know what the surface syntax ought to look like on this, but just as a strawman I propose `[@@unsafe_allow_any_kind]` on the type declaration in the impl.

Additionally, since the above (which we still think there's a chance we'll want to keep around for a while) will take some time, I think we proposed to merge this PR as-is for now, and revert / remove this functionality down the line once the more-selective attribute is implemented.

@tdelvecchio-jsc I think is on the hook for examples (beyond what we already know about, which is the Capsule API) of what the attribute-on-the-type version of this looks like compared to what the modalities-and-magic version look like, to help us convince ourselves which approach is better long-term.

tdelvecchio-jsc:
This comment is about what the "long term" design should be; in particular, whether there should be any mechanism for giving a type a different kind than what is able to be inferred for it, even after kind inference becomes much stronger. The two sides are:
1. Keep some mechanism (for the sake of this comment, via attribute `[@@unsafe_allow_any_kind]`) available to use
2. Use modalities and various forms of `Obj.magic*`
Here's a particular example that I think demonstrates why, from a software engineering perspective, option 1 is sometimes a better choice than option 2.

Consider the following two implementations of `int iarray` (which can cross contention) backed by an `int array`:
```
module Obj = struct
  external magic_uncontended : 'a @ contended -> 'a = "%identity"
end

module type Iarray_of_ints = sig
  type t : value mod portable uncontended

  val of_list : int list -> t
  val get : t -> int -> int
end

module Using_illegal_crossing : Iarray_of_ints = struct
  type t : value mod portable uncontended = T of int array [@@unboxed] [@@unsafe_allow_any_kind]

  let of_list list = T (Array.of_list list)
  let get (T array) idx = array.(idx)
end

module Using_obj_magic : Iarray_of_ints = struct
  type t : value mod portable uncontended = T of int array @@ contended [@@unboxed]

  let of_list list = T (Array.of_list list)
  let get (T array) idx = (Obj.magic_uncontended array).(idx)
end
```

Both of these implementations seem reasonable enough: 
1. `Using_illegal_crossing` requires the programmer to write down some invariant that the type upholds in order to be safe, and to manually check that this invariant is preserved across all functions in the implementation.
2. `Using_obj_magic` requires the programmer to consider the safety of using `Obj.magic_uncontended` at that particular site; this reasoning is relatively straightforward, and is in fact safe.

Now, consider another dev coming along and wanting to add a new function `of_array` to the `iarray` library:
```
module type Iarray_of_ints = sig
  type t : value mod portable uncontended

  val of_list : int list -> t
  val get : t -> int -> int
  val of_array : int array -> t
end

module Using_illegal_crossing : Iarray_of_ints = struct
  type t : value mod portable uncontended = T of int array [@@unboxed] [@@unsafe_allow_any_kind]

  let of_list list = T (Array.of_list list)
  let get (T array) idx = array.(idx)
  let of_array array = T array
end

module Using_obj_magic : Iarray_of_ints = struct
  type t : value mod portable uncontended = T of int array @@ contended [@@unboxed]

  let of_list list = T (Array.of_list list)
  let get (T array) idx = (Obj.magic_uncontended array).(idx)
  let of_array array = T array
end
```

In both cases, the dev just had to add `let of_array array = T array`, and in both implementations, this is a bug: the original `array` might still have `uncontended` references, so sending the "`iarray`" to another domain and reading from it would be a data race.
1. In `Using_illegal_crossing`, however, it would be clear that `of_array` breaks the invariant attached to `type t`, and so its unsafety would be clear without having to read the entire implementation.
2. In `Using_obj_magic`, the *only* signal that this is unsafe is a use of `Obj.magic_uncontended` in a *different function*. In a more complicated data structure and a more complicated implementation, this could be *very* easy to miss, and requires every change to the implementation to consider whether it's compatible with the uses of `Obj.magic*` throughout the implementation.

Essentially: `Obj.magic*` should only be used in cases where the safety of its usage is a *local* property, while safety that requires some invariant to be upheld *globally* should be signaled by unsafe magic on the type.

testsuite/tests/typing-layouts/allow_any.ml
glittershark: (commented later, this version of this feature is temporary)

liam923:
In reply to @tdelvecchio-jsc :

You've partially swayed my opinion. I do agree that `Using_illegal_crossing` in your example is more maintainable than `Using_obj_magic`. But I think there's a third option that would be better than the two given:
```
module Using_obj_magic_safely : Iarray_of_ints = struct
  module T : sig
    type t : value mod portable uncontended
    (* these names could be better *)
    val unsafe_of_array_promise_no_more_mutation : int array -> t
    val unsafe_to_array_promise_i_won't_mutate : t -> int array
  end = struct
    type t : value mod portable uncontended = T of int array @@ contended [@@unboxed]
    let unsafe_of_array_promise_no_more_mutation t = T t
    let unsafe_to_array_promise_i_won't_mutate (T t) = Obj.magic_uncontended t
  end

  let of_list list = T.unsafe_of_array_promise_no_more_mutation (Array.of_list list)
  let get t idx = (T.unsafe_to_array_promise_i_won't_mutate t).(idx)
end
```
The advantage here is that the unsafety is encapsulated by the interface of `T`. This forces each location that does something unsafe to explicitly state what promises it is making. Whereas with `Using_illegal_crossing`, the unsafety is only stated at the type definition. Now to implement `of_array`, the user must explicitly type out the phrase `promise_i_won't_mutate`, which is a promise implicitly made in the `Using_illegal_crossing` example.

I will caveat this though by saying that I'm not sure this will scale well to more complex data structures. And in practice, I'm not sure people will actually write code like this, so maybe it'd be good to add a way to force types to have certain mode-crossing behavior so that users reach for `Using_illegal_crossing` over `Using_obj_magic`.

goldfirere:
I don’t have a great sense about choosing between these other approaches and am happy for others to debate this.

But I do have a sense for how the impl-only feature should look. I concretely propose

```
type ‘a t = { mutable x : int; y : ‘a }
  [@@unsafe_adjust_kind_bounds (portable : ‘a) (uncontended : ‘a)]
```

That is, I think the attribute should include the desired bounds directly, instead of extracting them from a kind annotation. This will make clearer that, for example, any re-export of the type needs to include the attribute, too.

testsuite/tests/typing-layouts/allow_any.ml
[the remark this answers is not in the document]
glittershark: per discussion further down-thread, I'm going to resolve this.

goldfirere:
Just in case there's any ambiguity, I'm fine with merging this as-is. Maybe we'll end up re-designing / doing better, but there's no great reason not to merge this patch.

PR #3391, opened 2024-12-18
Promote `val_modalities` to `Stable`

riaqn:
This PR promotes modalities on values descriptions and module declarations to `Stable`.

There are no longer behaviors controlled by `mode_alpha`, but maybe it's fine to keep it for future convenience.

I confirm that our internal code base builds without changes, with this PR.

### DIFF
--- a/jane/doc/extensions/modes/intro.md
+++ b/jane/doc/extensions/modes/intro.md
@@ -1,0 +1,9 @@
+# Introduction to the Mode System
+
+The locality mode used for [Stack allocation](../local/intro.md) has been
+extended to more axes, each tracking a property of values, in order to support
+features such as [Parallelism](../parallelism/intro.md) and
+[Uniqueness](../uniqueness/intro.md). This documentation gives a general
+introduction to the mode system.
+
+*This is a stub. Documentation will come soon.*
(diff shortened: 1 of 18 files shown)


tdelvecchio-jsc (approved):
Looks good.

tdelvecchio-jsc (changes requested):
I'm currently running into an issue where
```
(* x.ml, with no mli *)
type t = T of unit

let t = T ()
let u = lazy t

(* y.ml *)
let _ @ uncontended = X.u

(* Error: This value is contended but expected to be uncontended. *)
```

I hypothesize (but have not checked) that this PR introduces this bug. Can you check?

riaqn:
This is #2967 missing a mode crossing spot. Will fix.

Since we will push `u` to stongest modality `portable`, that makes the implicit lazy closure `fun () -> t` portable, which makes `t` only available as `contended` in the closure, which makes the returns of the closure `contended` and thus the whole `lazy` is at `contended`.

riaqn:
Well I just realized that the said mode crossing is already there for general variable lookup. What's missing is that the kind system don't know `type t` crosses contention. You can either try #3284, or maybe manually specify that it crosses `contention`?

Do you encounter similiar issues where such `type t` contains mutable fields? That might be more tricky to think about.

goldfirere (changes requested):
I think this should also update the documentation, now that we're putting the feature in stable.

goldfirere (approved):
Docs look fine. Is there an internal ticket to finish the job?

riaqn:
Yes internal JIRA-3903

PR #3392, opened 2024-12-18
Improve exhaustiveness counterexamples for unboxed records

rtjoa:
Fixes a CR left from #3375, which should be merged first. When reviewing this PR, see just the top commit.

(@ccasin since you're working from a fork, I thought I separate PR would be easier, but feel free to just cherry-pick/adapt this commit instead.)

### DIFF
--- a/testsuite/tests/typing-layouts-unboxed-records/exhaustiveness.ml
+++ b/testsuite/tests/typing-layouts-unboxed-records/exhaustiveness.ml
@@ -34,7 +34,7 @@ Lines 2-4, characters 2-26:
 4 |   | #{ y = B; _ } -> false
 Warning 8 [partial-match]: this pattern-matching is not exhaustive.
 Here is an example of a case that is not matched:
-#{y=A; _ }
+#{y=A; x=B}
 
 val g : t -> t -> bool = <fun>
 |}]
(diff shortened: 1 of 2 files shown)


testsuite/tests/typing-layouts-unboxed-records/exhaustiveness.ml
```diff
@@ -34,7 +34,7 @@ Lines 2-4, characters 2-26:
 4 |   | #{ y = B; _ } -> false
 Warning 8 [partial-match]: this pattern-matching is not exhaustive.
 Here is an example of a case that is not matched:
-#{y=A; _ }
+#{y=A; x=B}
```
ccasin, line 37: Can you change the comment above this test to not be a CR and explain why the test is here?

typing/parmatch.ml
```diff
@@ -793,7 +794,7 @@ let build_specialized_submatrices ~extend_row discr rows =
     let initial_constr_group =
       let open Patterns.Head in
       match discr.pat_desc with
-      | Record _ | Tuple _ | Lazy ->
```
ccasin, line 782: Should unboxed tuple also be added to this case? (I have not thought deeply about what this case is doing, and the comment suggests it's a non-critical improvement in any case, so don't spend too much time trying to answer this)

ccasin (approved):
Last commit looks good, module a couple minor things.  Thanks!

typing/parmatch.ml
rtjoa: I'm also not sure exactly what this case is doing, but I just added it because unboxed tuples/records should be treated analogously to their boxed versions in this file.

PR #3394, opened 2024-12-18
Test modalities in unboxed records

rtjoa:


ccasin (approved):
Thanks for checking this!

PR #3398, opened 2024-12-19
Module aliases save locks instead of walking them immediately

riaqn:
In this PR, module aliases don't walk locks immediately. Instead, they hold locks to be walked later when the alias is used.

Note: the following program is allowed by the current PR, but shouldn't when we have stack-allocated modules:
```
module M @ local = struct
end

let (f @ global) () = let module N = M in ()
```

### DIFF
--- a/typing/env.mli
+++ b/typing/env.mli
@@ -204,13 +204,7 @@ type shared_context =
   | Module
   | Probe
 
-type lock =
-  | Escape_lock of escaping_context
-  | Share_lock of shared_context
-  | Closure_lock of closure_context * Mode.Value.Comonadic.r
-  | Region_lock
-  | Exclave_lock
-  | Unboxed_lock (* to prevent capture of terms with non-value types *)
+type lock
 
 (** Items whose accesses are affected by locks *)
 type lock_item =
--- a/typing/env.mli
+++ b/typing/env.mli
@@ -403,7 +403,7 @@ val add_module: ?arg:bool -> ?shape:Shape.t ->
 val add_module_lazy: update_summary:bool ->
   Ident.t -> module_presence -> Subst.Lazy.module_type -> t -> t
 val add_module_declaration: ?arg:bool -> ?shape:Shape.t -> check:bool ->
-  Ident.t -> module_presence -> module_declaration -> t -> t
+  Ident.t -> module_presence -> module_declaration -> ?locks:lock list -> t -> t
 val add_module_declaration_lazy: ?arg:bool -> update_summary:bool ->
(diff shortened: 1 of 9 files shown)


typing/env.mli
```diff
@@ -204,6 +204,14 @@ type shared_context =
   | Module
   | Probe
 
+type lock =
```
riaqn, line 207: As disucssed, only expose abstract `type locks`.

typing/typemod.ml
```diff
@@ -2587,18 +2587,37 @@ let maybe_infer_modalities ~loc ~env ~md_mode ~mode =
...
+
+let is_alias = function
+  | No -> false
+  | Yes_walk_locks | Yes_hold_locks -> true
+
+let rec type_module_maybe_hold_locks ~alias sttn funct_body anchor env smod =
   Builtin_attributes.warning_scope smod.pmod_attributes
     (fun () -> type_module_aux ~alias sttn funct_body anchor env smod)
 
+and type_module ?(alias=false) sttn funct_body anchor env smod =
```
riaqn, line 2603: Define multiple versions of this function, and get rid of `fatal_error`.

typing/typemod.ml
```diff
@@ -2587,18 +2587,37 @@ let maybe_infer_modalities ~loc ~env ~md_mode ~mode =
...
+  let alias = if alias then Yes_walk_locks else No in
+  let md, shape, locks =
+    type_module_maybe_hold_locks ~alias sttn funct_body anchor env smod
+  in
+  begin match locks with
+  | [] -> ()
+  | _ :: _ -> Misc.fatal_error "locks should not be held"
+  end;
```
riaqn, line 2611: Use `assert` instead.

typing/typemod.ml
riaqn: I cannot define mutliple versions of the function, because that would create too much code duplication. I instead use GADT which maybe isn't so nice either (and I'm happy to revert)

typing/env.mli
riaqn: In the end I expose abstract `type lock`, but the user code still know `locks` is `lock list` because they need to be able to construct an empty lock list.

typing/typemod.ml
riaqn: This is now removed by GADT (which might be reverted later, see above)

typing/typemod.ml
lukemaurer: Yeah, it pains me to say it but I don't think the GADT is helping on balance.

lukemaurer (commented):
A few thoughts but I think this is basically there.

lukemaurer (approved):

PR #3399, opened 2024-12-19
Fix `constrain_type_jkind` for unboxed records with `any` and erroneous jkind annotations

rtjoa:
Currently, the below breaks an invariant of `Ctype.constrain_type_jkind`, causing a fatal error instead of a type error:
```ocaml
type a = B of b
and b : any = #{ i : int ; j : int }
```

Specifically, the `any` annotation obscures the fact that the unboxed record is a product, breaking the following invariant assumed by `Ctype.constrain_type_jkind`: a type with a product kind is given a product jkind of the correct arity by `estimate_type_jkind`. (`constrain_type_jkind` is called to check that `b` is representable as the argument to `B`.)

We fix this by giving unboxed records a product-of-`any`s jkind when they would otherwise be given `any`. This allows `estimate_type_jkind` to give an estimate that's just barely good enough, such that `constain_type_jkind` can always recurse on the labels.

Still, it's possible to get a "products jkinds don't line up" error if `b` is given a jkind annotation with an incorrect arity.
```ocaml
type a : B of b
and b : any & any & any = #{ i : int ; j : int }
```
We fix this simply returning a type error instead of a fatal error when product jkinds don't line up, as this should only reachable with incorrect annotations.

Review: @ccasin

### DIFF
--- a/testsuite/tests/typing-layouts-unboxed-records/basics.ml
+++ b/testsuite/tests/typing-layouts-unboxed-records/basics.ml
@@ -712,3 +712,11 @@ Error: The universal type variable 'a was declared to have kind any.
        But it was inferred to have kind value_or_null
          because of the definition of t at line 1, characters 0-40.
 |}]
+
+type a = B of b
+and b : any = #{ i : int ; j : int }
+[%%expect{|
+>> Fatal error: unboxed product jkinds don't line up
+Uncaught exception: Misc.Fatal_error
+
+|}]
--- a/testsuite/tests/typing-layouts-unboxed-records/basics.ml
+++ b/testsuite/tests/typing-layouts-unboxed-records/basics.ml
@@ -731,3 +731,15 @@ Line 1, characters 9-15:
 Error: Type "b" has layout "value & value".
        Variants may not yet contain types of this layout.
 |}]
+type a = B of b
+and b : any & any & any = #{ i : int ; j : int }
+[%%expect{|
+Line 2, characters 0-48:
+2 | and b : any & any & any = #{ i : int ; j : int }
+    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Error:
+       The layout of b is any & any & any
+         because of the annotation on the declaration of the type b.
+       But the layout of b must be representable
+         because it's the type of a constructor field.
+|}]
--- a/testsuite/tests/typing-layouts-unboxed-records/basics.ml
+++ b/testsuite/tests/typing-layouts-unboxed-records/basics.ml
@@ -713,10 +713,21 @@ Error: The universal type variable 'a was declared to have kind any.
          because of the definition of t at line 1, characters 0-40.
 |}]
 
+
+(************************************************************)
+
+(* This is a regression test. Previously, we hit a fatal error because the [any]
+   annotation obscured the fact that the unboxed record is a product, breaking
+   an invariant assumed by [Ctype.constrain_type_jkind].
+
+   [a] was necessary to trigger the error because it called
+   [check_representable] on [b], constraining its kind to be a sort variable. *)
 type a = B of b
 and b : any = #{ i : int ; j : int }
 [%%expect{|
->> Fatal error: unboxed product jkinds don't line up
-Uncaught exception: Misc.Fatal_error
-
+Line 1, characters 9-15:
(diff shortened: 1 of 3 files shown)


goldfirere:
I was intrigued so I took a look. I don't love the approach here, because it re-examines the kind, even after the kind has already been (in theory) fully processed.

And I think it's a bit wrong; I would expect it to fail with e.g. `type t : any mod portable = #{ ... }`, because that kind annotation isn't `max`.

I think an alternative approach would be to have the `product` helper in `constrain_type_jkind` take `ty's_jkind` as an argument, and then pass a product jkind in from the `Stepped_record_unboxed_product` case (after calling `estimate_type_jkind` on all the `tys`).... or maybe do this extra jkind construction only if the layout in `ty's_jkind` is `any` (which should happen only in the current scenario, I think).

------------------------------------------
Some other ramblings, before I came up with the solution suggested above (feel free to skip):

Why do we even look at the annotation at this point? I think it's just for error messages. That is, an alternative implementation would be to ignore the annotation in `transl_declaration` entirely (on anything but an abstract type with no manifest), and then just check the annotation in `update_decl_jkind` (which would have to take the annotation as an extra parameter, as it would no longer be stored in `type_jkind`). But I think doing this would make error messages worse, because (essentially) it means doing all the type-checking of mutually recursive types without ever peeking at the helpful information the user has written in annotations.

But: if the jkind inferred from the kind is less than the jkind written in the annotation, we can safely use the inferred jkind. That inferred one will have *better* error messages (I think), because it contains more information. And, by transitivity of subjkinding, the existing subjkind check at the end of `update_decl_jkind` still does the right thing. So then I thought that we put the lesser of `annot` or `jkind_default` in `type_jkind`.

Yet that's not great, because sometimes neither will be less than the other. Boo. That led me to want computing the intersection (or meet) between the two. That would work nicely.... except that we can't take the intersection of l-jkinds. (Concretely, if both have `with` bits, it's deeply unclear how to combine them.)

Happily, there is just barely a solution here: we really only care about the layout. And there is no distinction between an l-layout and an r-layout. So we can intersect the layouts of the annotation with the inferred jkind, and that should sidestep the problem.

Having written all that.... really, this is important only for unboxed records. So maybe this just amounts to a big explanation motivating the current patch... even though I don't like re-inspecting the kind.

ccasin:
> I was intrigued so I took a look. I don't love the approach here, because it re-examines the kind, even after the kind has already been (in theory) fully processed.
> 
> And I think it's a bit wrong; I would expect it to fail with e.g. type t : any mod portable = #{ ... }, because that kind annotation isn't max.
> 
> I think an alternative approach would be to have the product helper in constrain_type_jkind take ty's_jkind as an argument, and then pass a product jkind in from the Stepped_record_unboxed_product case (after calling estimate_type_jkind on all the tys).... or maybe do this extra jkind construction only if the layout in ty's_jkind is any (which should happen only in the current scenario, I think).

I agree the current approach feels a bit gross, and I too played around with other options involving modifying `constrain_type_jkind` itself when Ryan and I were discussing this earlier today.

I ended up prefer the current version. This avoids "polluting" `constrain_type_jkind`/`estimate_type_jkind` with a special case that is only necessary due to an implementation detail of type declaration checking.  That is, I think this case is only reachable due to the fact that the environment computed in `type_decl` is less accurate than it could be, so it's better to fix that environment rather than adding a weird case to `constrain_type_jkind` that shouldn't be reachable once we have checked type decls.

I'm less certain this is the best possible modification to type_decl - it feels like some restructuring earlier in `transl_declaration` might be possible that leads to this solution more naturally, avoiding the need to "re-examine the kind".  But even if there's no such simplification, this seems like the right solution to me, unless there is some way this case of `constrain_type_jkind` could be reached outside of typedecl.

goldfirere:
Fine. I buy the argument that this is Typedecl's problem, and so it should contain the solution. None of us likes the particular solution implemented, but anything more general is sad in a different way.

Maybe could I request a little comment on this check saying that it's a bit sad and pointing to this PR? This is the kind of thing I might see in a few years' time and think "why would someone do that?". We actually have a decent answer (in the comments in this PR), so we may as well allow the reader to find them.

ccasin (approved):
Approved, but please edit the comment to include a reference back to this PR before merging, as Richard requested.

Thanks for fixing this!

rtjoa:
Comment updated, auto-merging.

PR #3402, opened 2024-12-20
Allow values reachable from ephemeron keys to be collected by minor GC

mshinwell:
Port https://github.com/ocaml/ocaml/pull/13643

(cherry picked from commit fa197971cbda8290bca250ab6dda3628eaed73ae)

cc @stedolan


NickBarnes (changes requested):
I think this is semantically correct, but I have quite a few style points, some of which should be done now (some could be left to some future refactor).

NickBarnes:
I've made NickBarnes/nbarnes-minor-ephemerons with most of the minor style things.

stedolan:
Closed in favour of #3495

PR #3404, opened 2024-12-23
Refactor cmm functions to be more generic over operand width

jvanburen:


jvanburen:
This PR refactors cmm_helpers in preparation for unboxed 8 and 16-bit integers to be added to the compiler.
It replaces many functions that were copied for each width with a single more generic version, and then updates to_cmm to use it.

backend/cmm_helpers.ml
```diff
@@ -1280,35 +1281,104 @@ let addr_array_initialize arr ofs newval dbg =
...
+(** [get_const_bitmask x] returns [Some (y, mask)] if [x] is [y & mask] *)
+let get_const_bitmask = function
+  | Cop (Cand, ([x; Cconst_natint (mask, _)] | [Cconst_natint (mask, _); x]), _)
+    ->
+    Some (x, mask)
+  | Cop (Cand, ([x; Cconst_int (mask, _)] | [Cconst_int (mask, _); x]), _) ->
+    Some (x, Nativeint.of_int mask)
+  | _ -> None
+
+(** [low_bits ~bits x] is a (hopefully simplified) value which agrees with x on at least
```
Gbury, line 1293:
```suggestion
(** [low_bits ~bits x] is a (potentially simplified) value which agrees with x on at least
```

backend/cmm_helpers.ml
```diff
@@ -1280,35 +1281,104 @@ let addr_array_initialize arr ofs newval dbg =
...
+    the low [bits] bits. E.g., [low_bits ~bits x & mask = x & mask], where [mask] is a
+    bitmask of the low [bits] bits . *)
+let rec low_bits ~bits ~dbg x =
+  assert (bits > 0);
+  if bits >= arch_bits
+  then x
+  else
+    let unused_bits = arch_bits - bits in
+    let does_mask_ignore_low_bits test_mask =
```
Gbury, line 1302: Rather than ignore, maybe it's more of a `does_mask_keep_low_bits` ?

backend/cmm_helpers.ml
```diff
@@ -1280,35 +1281,104 @@ let addr_array_initialize arr ofs newval dbg =
...
+      (* If the mask has all the low bits set, then the low bits are unchanged.
+         This could happen from zero-extension. *)
+      let mask = Nativeint.pred (Nativeint.shift_left 1n bits) in
+      Nativeint.equal mask (Nativeint.logand test_mask mask)
+    in
+    (* Ignore sign and zero extensions, which do not affect the low bits *)
```
Gbury, line 1308:
```suggestion
    (* Ignore sign and zero extensions which do not affect the low bits *)
```

Since not all zero/sign-extensions do not affect the low bits and we only want to ignore those that indeed do not affect the low bits, I think that the phrase better reflect the code without the comma (but I may be wrong) ?

backend/cmm_helpers.ml
```diff
@@ -1280,35 +1281,104 @@ let addr_array_initialize arr ofs newval dbg =
...
+  if bits = arch_bits
+  then e
+  else
+    map_tail
+      (function
+        | Cop ((Casr | Clsr), [inner; Cconst_int (n, _)], _) as e
+          when 0 <= n && n < arch_bits ->
+          (* see middle_end/flambda2/z3/sign_extension.py for proof *)
+          if n > unused_bits
+          then (* already sign-extended *) e
```
Gbury, line 1366: Are we sure that it also applies to the `Clsr` case ? The `Casr` case is reasonable, but the logical shift strikes me as weird.

backend/cmm_helpers.ml
```diff
@@ -1680,20 +1747,16 @@ let alloc_generic_set_fn block ofs newval memory_chunk dbg =
     addr_array_initialize block ofs newval dbg
   | Word_int -> generic_case ()
   (* Generic cases that may differ under big endian archs *)
-  | Single _ | Double | Thirtytwo_unsigned | Thirtytwo_signed
-  | Onetwentyeight_unaligned | Onetwentyeight_aligned ->
+  | Byte_unsigned | Byte_signed | Sixteen_unsigned | Sixteen_signed | Single _
+  | Double | Thirtytwo_unsigned | Thirtytwo_signed | Onetwentyeight_unaligned
+  | Onetwentyeight_aligned ->
```
Gbury, line 1752: Is this change really necessary / part of this PR ? It seems to change the semantics of that function rather than be a refactoring.

backend/cmm_helpers.ml
```diff
@@ -1793,19 +1856,17 @@ let make_mixed_alloc ~mode dbg ~tag ~value_prefix_size args args_memory_chunks =
           (* regular scanned part of a block *)
           match memory_chunk with
           | Word_int | Word_val -> ok ()
-          | Byte_unsigned | Byte_signed | Sixteen_unsigned | Sixteen_signed ->
-            error "mixed blocks"
+          | Byte_unsigned | Byte_signed | Sixteen_unsigned | Sixteen_signed
```
Gbury, line 1859: Same remark as another change: this changes semantics /allow more things. It may be correct/reasonable, but it's surprising to see this as part of a refactor PR.

backend/cmm_helpers.ml
```diff
@@ -1793,19 +1856,17 @@ let make_mixed_alloc ~mode dbg ~tag ~value_prefix_size args args_memory_chunks =
...
         else
           (* flat suffix part of the block *)
           match memory_chunk with
           | Word_int | Thirtytwo_unsigned | Thirtytwo_signed | Double
-          | Onetwentyeight_unaligned | Onetwentyeight_aligned | Single _ ->
-            ok ()
+          | Onetwentyeight_unaligned | Onetwentyeight_aligned | Single _
           | Byte_unsigned | Byte_signed | Sixteen_unsigned | Sixteen_signed ->
-            error "mixed blocks"
+            ok ()
```
Gbury, line 1869: same remark

backend/cmm_helpers.ml
```diff
@@ -1280,35 +1281,104 @@ let addr_array_initialize arr ofs newval dbg =
...
+    asr_const (lsl_const e unused_bits dbg) unused_bits dbg
+  in
+  if bits = arch_bits
+  then e
+  else
+    map_tail
+      (function
+        | Cop ((Casr | Clsr), [inner; Cconst_int (n, _)], _) as e
+          when 0 <= n && n < arch_bits ->
+          (* see middle_end/flambda2/z3/sign_extension.py for proof *)
```
TheNumbat, line 1364: Doesn't seem to exist?

backend/cmm_helpers.ml
TheNumbat, line 1366: Perhaps note that `n` is strictly greater than `unused_bits` so it overwrote the sign

backend/cmm_helpers.ml
```diff
@@ -1448,7 +1519,7 @@ let set_field_unboxed ~dbg memory_chunk block ~index_in_words newval =
     let field_address =
       array_indexing log2_size_addr block index_in_words dbg
     in
-    let newval = if size_in_bytes = 4 then low_32 dbg newval else newval in
+    let newval = low_bits newval ~dbg ~bits:(8 * size_in_bytes) in
```
TheNumbat, line 1522: Not specific to this PR, but I noticed this does not write sign-extended int32#s into mixed blocks. Flambda2 tries to make sure the full content of blocks is deterministic (e.g. [in closures](https://github.com/ocaml-flambda/flambda-backend/blob/main/middle_end/flambda2/to_cmm/to_cmm_shared.ml#L398)), so I think [this code](https://github.com/ocaml-flambda/flambda-backend/blob/main/middle_end/flambda2/to_cmm/to_cmm_primitive.ml#L188) should probably be changed. But maybe this can never be used on static blocks so we don't care---we should check with the flambda2 devs.

TheNumbat (approved):
lgtm; some comments on comments

backend/cmm_helpers.ml
jvanburen: yes, it is proved. the trick is that the right shift either makes the sign-extend irrelevant, or all of the zero bits get cleared by sign-extension anyway

Gbury (approved):

PR #3407, opened 2024-12-24
Check for type recursion without boxing

rtjoa:
## Summary

We only allow recursion in unboxed types to occur through boxes, otherwise the type is uninhabitable and usually also infinite-size.

Because `check_well_founded` already ruled out recursion through structural types (in our case, we care about unboxed tuples and aliases), we just look for a cycle in nominal unboxed types (`[@@unboxed]` types and unboxed records), tracking the set of seen paths. This gives an algorithm that is simpler than `check_well_founded` (whose correctness relies on some subtle properties of `Tconstr`'s `abbrev_memo`).

We newly disallow recursive singleton `[@@unboxed]` types (`type t = { t : t } [@@unboxed]`), which are uninhabitable anyway.

## Specification

For each group of mutually recursive type declarations, we define the following "type contains" transitive relation on type expressions:
- Unboxed records and variants defined in the group contain their fields
  - If `type 'a t = #{ ...; lbl : u;  ... }`, `type 'a t = { lbl : u } [@@unboxed]`, or `type 'a t = U of u [@@unboxed]` is in the recursive group, then `'a t` contains `u`
- Abbreviations defined in the group contain their expansions
  - If `type 'a t = u` is in the recursive group then `'a t` contains `u`
- Types not in the recursive group contain the parameters indicated by their layout
  - `'a t` contains `'a` if `layout_of 'a` or `any` occurs in `'a t`'s layout
  - We don't yet have `layout_of`, so currently only consider `any`
- Unboxed tuples contain their components: `#(u_1 * u_2 * ...)` contains all `u_i`

(`'a` is a metavariable for zero or more type parameters; `t` is a metavariable for type constructors; `u` is a metavariable for type expressions)

If a path starting from the type expression on the LHS of a declaration contains two types with the same head type constructor, and that repeated type is an unboxed record or variant, then we give an error.

## Examples
```ocaml
type t = #{ t : t list }
(* Allowed, [t list] is not in this recursive group, and
   has layout [value] so it doesn't contain anything *)

type t = #{ x : #(int * u) }
and u = T of t [@@unboxed]
(* Banned, t -> #(int * u) -> u -> t *)

type 'a id = #{ a : 'a }
type t = int id id
(* Allowed, [int id] is not in this recursive group, and
   has layout [value] so it doesn't contain anything *)

type 'a id = #{ a : 'a }
and t = int id id
(* Sadly, banned. t -> int id id -> int id; id appears twice *)
```

See [`typing-layouts-unboxed-records/recursive.ml`](https://github.com/ocaml-flambda/flambda-backend/blob/rtjoa.check-type-recursion-without-indirection/testsuite/tests/typing-layouts-unboxed-records/recursive.ml) for more.

## Alternatives

**Layout variables.** We could introduce layout variables (for this check only), finding infinite-size types via the occurs check. Such an implementation might be more principled, accepting finite-size recursive types like `type t = #{ t : t }`. The current implementation gives more informative error traces, and leaves the flexiblity to switch to a less restrictive check in the future.

**The algorithm from the [Unboxed data constructors](https://arxiv.org/pdf/2311.07369) paper.** This would accept types like `type 'a id = #{ a : 'a } and t = int id id`. I think this could be an appealing final solution, but some details are a bit subtle. See discussion on https://github.com/ocaml/ocaml/pull/10479.

**Generalize `check_well_founded`.** My first implementation, which roughly generalizes `check_well_founded` over the "type contains" relation, accepts `type 'a id = #{ a : 'a } and t = int id id`. (On [`rtjoa.check-well-founded-unboxed-recursion`](https://github.com/ocaml-flambda/flambda-backend/tree/rtjoa.check-well-founded-unboxed-recursion), unpolished.) I'd hesitate to increase our reliance on `check_well_founded`, which I believe roughly implements the same algorithm as the paper above in a tricky way.

---

Review: @goldfirere

### DIFF
--- a/typing/ctype.mli
+++ b/typing/ctype.mli
@@ -581,6 +581,10 @@ val get_unboxed_type_approximation : Env.t -> type_expr -> type_expr
        [get_unboxed_type_representation], but doesn't indicate whether the type
        was fully expanded or not. *)
 
+val contained_without_boxing : Env.t -> type_expr -> type_expr list
+    (* Return all types that are directly contained without boxing
+      (or "without indirection" or "flatly") *)
+
 (* Given the row from a variant type, determine if it is immediate.  Currently
    just checks that all constructors have no arguments, doesn't consider
    void. *)
--- a/typing/typedecl.ml
+++ b/typing/typedecl.ml
@@ -89,6 +89,7 @@ type error =
   | Unboxed_mutable_label
   | Recursive_abbrev of string * Env.t * reaching_type_path
   | Cycle_in_def of string * Env.t * reaching_type_path
+  | Unboxed_recursion of string * Env.t * reaching_type_path
   | Definition_mismatch of type_expr * Env.t * Includecore.type_mismatch option
   | Constraint_failed of Env.t * Errortrace.unification_error
   | Inconsistent_constraint of Env.t * Errortrace.unification_error
@@ -2065,6 +2066,105 @@ let check_well_founded_decl  ~abs_env env loc path decl to_check =
        end)} in
   it.it_type_declaration it (Ctype.generic_instance_declaration decl)
 
+(* We only allow recursion in unboxed product types to occur through boxes,
+   otherwise the type is uninhabitable and usually also infinite-size.
+   See [typing-layouts-unboxed-records/recursive.ml].
+
+   Because [check_well_founded] already ruled out recursion through structural
+   types, we just look for a cycle in nominal unboxed types ([@@unboxed] types
+   and unboxed records), tracking the set of seen paths.
+
+   For each group of mutually recursive type declarations, we define the
+   following "type contains" transitive relation on type expressions:
+
+   1. Unboxed records and variants defined in the group contain their fields.
+
+      If [type 'a t = #{ ...; lbl : u;  ... }],
+      or [type 'a t = { lbl : u } [@@unboxed]],
+      or [type 'a t = U of u [@@unboxed]]
+      is in the recursive group, then ['a t] contains [u].
+
+   2. Abbreviations defined in the group contain their expansions.
+
+      If [type 'a t = u] is in the recursive group then ['a t] contains [u].
+
+   3. Unboxed tuples contain their components.
+
+      [#(u_1 * ...)] contains all [u_i].
+
+   4. Types not in the group contain the parameters indicated by their layout.
+
+      ['a t] contains ['a] if [layout_of 'a] or [any] occurs in ['a t]'s layout.
+      Note: We don't yet have [layout_of], so currently only consider [any].
+
+   If a path starting from the type expression on the LHS of a declaration
+   contains two types with the same head type constructor, and that repeated
+   type is an unboxed record or variant, then the check raises a type error.
+
+   CR layouts v7.2: accept safe types that expand the same path multiple times,
+   e.g. [type 'a t = #{ a : 'a } and x = int t t], either by using layouts
+   variables or the algorithm from "Unboxed data constructors - or, how cpp
+   decides a halting problem."
+   See https://github.com/ocaml-flambda/flambda-backend/pull/3407.
+*)
+type step_result =
+  | Contained of type_expr list
+  | Expanded_to of type_expr
+  | Is_cyclic
+let check_unboxed_recursion ~abs_env env loc path0 ty0 to_check =
+  let contained_parameters tyl layout =
+    (* A type whose layout has [any] could contain all its parameters.
+       CR layouts v11: update this function for [layout_of] layouts. *)
+    let rec has_any : Jkind_types.Layout.Const.t -> bool = function
+      | Any -> true
+      | Base _ -> false
+      | Product l -> List.exists has_any l
+    in
+    if has_any layout then tyl else []
+  in
+  let step_once parents ty =
+    match get_desc ty with
+    | Tconstr (path, tyl, _) ->
+      if to_check path then
+        if Path.Set.mem path parents then
+          Is_cyclic, parents
+        else
+          let parents = Path.Set.add path parents in
+          match Ctype.try_expand_safe_opt env ty with
+          | ty' ->
+            Expanded_to ty', parents
+          | exception Ctype.Cannot_expand ->
+            Contained (Ctype.contained_without_boxing env ty), parents
+      else
+        begin try
+          (* Determine contained types by layout for decls outside of the
+             recursive group *)
+          let jkind = (Env.find_type path env).type_jkind in
+          let layout = Option.get (Jkind.get_layout jkind) in
+          Contained (contained_parameters tyl layout), parents
+        with Not_found | Invalid_argument _ ->
+          (* Because [to_check path] is false, this decl has already been
+            typechecked, so it's already in [env] with a constant layout. *)
+          Misc.fatal_error "Typedecl.check_unboxed_recursion"
+        end
+    | _ -> Contained (Ctype.contained_without_boxing env ty), parents
+  in
+  let rec visit parents trace ty =
+    match step_once parents ty with
+    | Contained tys, parents ->
+      List.iter (fun ty' -> visit parents (Contains (ty, ty') :: trace) ty') tys
+    | Expanded_to ty', parents ->
+      visit parents (Expands_to(ty,ty') :: trace) ty'
+    | Is_cyclic, _ ->
+      raise (Error (loc, Unboxed_recursion (path0, abs_env, List.rev trace)))
+  in
+  Ctype.wrap_trace_gadt_instances env (visit Path.Set.empty []) ty0
+
+let check_unboxed_recursion_decl ~abs_env env loc path decl to_check =
+  let decl = Ctype.generic_instance_declaration decl in
+  let ty = Btype.newgenty (Tconstr (path, decl.type_params, ref Mnil)) in
+  check_unboxed_recursion ~abs_env env loc (Path.name path) ty to_check
+
 (* Check for non-regular abbreviations; an abbreviation
    [type 'a t = ...] is non-regular if the expansion of [...]
    contains instances [ty t] where [ty] is not equal to ['a].
@@ -2353,6 +2453,11 @@ let transl_type_decl env rec_flag sdecl_list =
     decls;
   List.iter (fun (tdecl, _shape) ->
     check_abbrev_regularity ~abs_env new_env id_loc_list to_check tdecl) tdecls;
+  List.iter (fun (id, decl) ->
+    check_unboxed_recursion_decl ~abs_env new_env (List.assoc id id_loc_list)
+      (Path.Pident id)
+      decl to_check)
+    decls;
   (* Now that we've ruled out ill-formed types, we can perform the delayed
      jkind checks *)
   List.iter (fun (checks,loc) ->
@@ -3438,6 +3543,7 @@ let check_recmod_typedecl env loc recmod_ids path decl =
      (path, decl) is the type declaration to be checked. *)
   let to_check path = Path.exists_free recmod_ids path in
   check_well_founded_decl ~abs_env:env env loc path decl to_check;
+  check_unboxed_recursion_decl ~abs_env:env env loc path decl to_check;
   check_regularity ~abs_env:env env loc path decl to_check;
   (* additional coherence check, as one might build an incoherent signature,
      and use it to build an incoherent module, cf. #7851 *)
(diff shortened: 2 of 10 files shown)


testsuite/tests/typing-layouts-unboxed-records/recursive.ml
```diff
@@ -106,92 +123,169 @@ type 'a bad = #{ bad : 'a bad ; u : 'a}
...
+type bad = #( s * s )
+and ('a : any) record_id2 = #{ a : 'a }
+and s = #{ u : u }
+and u = #(int * bad record_id2)
+[%%expect{|
+Line 1, characters 0-21:
+1 | type bad = #( s * s )
+    ^^^^^^^^^^^^^^^^^^^^^
+Error: The definition of "bad" is recursive without boxing:
+         "bad" = "#(s * s)",
```
goldfirere, line 167: Nice error message!

testsuite/tests/typing-layouts-unboxed-records/recursive.ml
```diff
@@ -106,92 +123,169 @@ type 'a bad = #{ bad : 'a bad ; u : 'a}
...
+         "#(Bad_rec1.s * Bad_rec1.s)" contains "Bad_rec2.u",
+         "Bad_rec2.u" = "Bad_rec1.t Bad_rec2.id",
+         "Bad_rec1.t Bad_rec2.id" = "Bad_rec1.t",
+         "Bad_rec1.t" = "#(Bad_rec1.s * Bad_rec1.s)",
+         "#(Bad_rec1.s * Bad_rec1.s)" contains "Bad_rec1.s"
+|}]
+
+(* When we allow records with elements of unrepresentable layout, this should
+   still be disallowed. *)
+module M : sig
```
goldfirere, line 226: This inspired a few more tests:

```ocaml
module M : sig
  type 'a t
end = struct
  type 'a t = 'a
end

type bad = #{ bad : bad M.t }
```

is accepted, even though it creates a bottomless type. This is harmless though because its size is finite.

But I also tried

```ocaml
module M : sig
  type ('a : value & value) t : value & value
end = struct
  type ('a : any) t = 'a
end

type bad : value & value = #{ x : int; bad : bad M.t }
```

which leads to

```
Error: The layout of type bad is value & (value & value)
         because it is an unboxed record.
       But the layout of type bad must be a sublayout of value & value
         because of the annotation on the declaration of the type bad.
```

This looks like the old message this patch is an attempt to improve. (Without the `value & value` annotation, the error message mentions `any & any`.)

I don't see an actual problem here, but I do think this example shows that we're still relying on the previous behavior for detecting these loops, at least for devious users.

testsuite/tests/typing-layouts-unboxed-records/recursive.ml
goldfirere, line 1: The new check is also conservative. It rejects

```ocaml
type 'a good = #{ good : 'a }
type x = int good good
```

with

```
Error: The definition of x is recursive without boxing:
         x = int good good,
         int good good contains int good
```

The error message states a true fact -- `int good good` _does_ contain `int good` -- but there's no problem here. I'm a little sad about this, but maybe it's OK? At least there should be a test case demonstrating the problem.

typing/ctype.mli
```diff
@@ -246,6 +246,9 @@ val apply:
 val try_expand_once_opt: Env.t -> type_expr -> type_expr
 val try_expand_safe_opt: Env.t -> type_expr -> type_expr
 
+val path_and_expansion: Env.t -> type_expr -> (Path.t * type_expr) option
+(** Return an option instead of raising, and gives the path of the input type *)
```
goldfirere, line 250: "raising": when? If the argument isn't a Tconstr?

goldfirere, line 250: The comment should say that it looks through `private` types, which I wouldn't otherwise expect.

typing/ctype.mli
```diff
@@ -581,6 +584,10 @@ val get_unboxed_type_approximation : Env.t -> type_expr -> type_expr
        [get_unboxed_type_representation], but doesn't indicate whether the type
        was fully expanded or not. *)
 
+val contained_without_boxing : Env.t -> type_expr -> type_expr list
+    (* Return all types that are contained without boxing
```
goldfirere, line 588: Not quite: this isn't recursive. Maybe

```suggestion
    (* Return all types that are directly contained without boxing
```

typing/typedecl.ml
```diff
@@ -2061,6 +2062,65 @@ let check_well_founded_decl  ~abs_env env loc path decl to_check =
        end)} in
   it.it_type_declaration it (Ctype.generic_instance_declaration decl)
 
+(* We only allow recursion in unboxed product types to occur through boxes,
+   otherwise the type is uninhabitable and usually also infinite-size.
+   See [typing-layouts-unboxed-records/recursive.ml].
+
+   Because `check_well_founded` already ruled out recursion through structural
```
goldfirere, line 2069:
```suggestion
   Because [check_well_founded] already ruled out recursion through structural
```

typing/ctype.mli
```diff
@@ -246,6 +246,9 @@ val apply:
 val try_expand_once_opt: Env.t -> type_expr -> type_expr
 val try_expand_safe_opt: Env.t -> type_expr -> type_expr
 
+val path_and_expansion: Env.t -> type_expr -> (Path.t * type_expr) option
```
goldfirere, line 249: Why have both path and expansion at the same time? I was expecting the implementation to take advantage of doing both operations at once... but it doesn't. Can this be removed in favor of other functions nearby?

goldfirere (commented):
I'm not quite convinced here. This mostly works. But it misses some cases (so it's not fully sound) and catches some cases it shouldn't (so it's not fully complete). The old check -- reporting when e.g. `value & (value & value)` doesn't subjkind with `value & value` is still in place, so the lack of soundness doesn't matter all that much. And maybe the lack of completeness doesn't bite in practice? But given these drawbacks, and the fact that the old check worked (but with a somewhat confusing error message), I think I lean mildly against merging.

goldfirere:
Actually, maybe I'm wrong about "mildly against merging". I agree that this improves the error message in the majority case, and that's pretty cool. And the implementation is simple enough. But at the least I'd leave an internal ticket about revisiting this in the future.

rtjoa (commented):
@goldfirere Thanks for the helpful review.

Here are my goals for this check:
1. Improve the programming experience for unboxed records
2. Prohibit all infinite-size types (types that would have a cyclic layout if we had layout variables without an occurs check)

It's because we are sound but not complete w.r.t. "no infinite-size types" that the error message complains about a stronger property, "no recursion without boxing," (it would be confusing to see a "no infinite-size types" error on `type t = #{ t : t }`).

So I'd agree that w.r.t. the stronger property of "no recursion without boxing," we are both unsound (your example with type recursion hidden behind an abstract signature, and your recursive module example) and incomplete (`int good good`). But I think that this is okay: the stronger property is only mentioned as it's a better error message.

I agree we'll eventually want to be more complete (e.g. allow `int good good`) - I'll create an internal ticket corresponding to the CR above `check_unboxed_recursion`.

lpw25:
It is a little unclear from the description what the definition of recursion is. Does the new check still allow more useful things like:
```
type t = { t : t list } [@@unboxed]
```

rtjoa:
@lpw25 The new check would allow your examples, including if `list` is abstract. A description of what exactly we allow (will also update comments):

For each group of mutually recursive type declarations, we define the following "type contains" relation on type expressions:
- Unboxed records and variants defined in the group contain their fields
  - If `type 'a t = #{ ...; lbl : u, ... }`, `type 'a t = { lbl : u } [@@unboxed]`, or `type 'a t = U of u [@@unboxed]` is in the recursive group, then `'a t` contains `u`
- Abbreviations defined in the group contain their expansions
  - If `type 'a t = u` is in the recursive group then `'a t` contains `u`
- Types not in the recursive group are treated as abstract
  - `'a t` contains `'a` if `layout_of 'a` occurs in `'a t`'s layout 
  - For now, as we don't have `layout_of`, types not in the recursive group are skipped because they all have a finite-size layout
- Unboxed tuples contain their components: `#(u_1 * u_2 * ...)` contains all `u_i`
- The relation is transitive

(`'a` is a metavariable for zero or more type parameters; `t` is a metavariable for type constructors; `u` is a metavariable for type expressions)

If a path starting from the type expression on the LHS of a declaration contains a two types with the same head type constructor, and that repeated type is an unboxed record or variant, then we give an error.

Examples (@goldfirere the third is newly allowed with a bug fix):
```ocaml
type t = #{ t : t list }
(* Allowed, [t list] is not in this recursive group, and
   has layout [value] so it doesn't contain anything *)

type t = #{ x : #(int * u) }
and u = T of t [@@unboxed]
(* Banned, t -> #(int * u) -> u -> t *)

type 'a id = #{ a : 'a }
type t = t id id
(* Allowed, [t id] is not in this recursive group, and
   has layout [value] so it doesn't contain anything *)

type 'a id = #{ a : 'a }
and t = int id id
(* Sadly, banned. t -> int id id -> int id; id appears twice *)
```

lpw25:
>  types not in the recursive group are skipped because they all have a finite-size layout

Presumably this is not the case for something like:
```ocaml
type ('a : any) t : any
```
is an `any` type just rejected directly currently?

lpw25:
> But I think it will be okay to accept this, since the type also won't be visibly cyclic to constrain_type_jkind (which is the cause of the bad errors this check is trying to avoid).

What about if your `Id` module were a functor parameter?

rtjoa:
> > But I think it will be okay to accept this, since the type also won't be visibly cyclic to constrain_type_jkind (which is the cause of the bad errors this check is trying to avoid).
> 
> What about if your `Id` module were a functor parameter?

If the functor parameter abstracts `t` as an `any` type, I think it should work similarly to the first example from my last comment: we'll accept it, but it should be fine since the type won't be visibly cyclic to `constrain_type_jkind`. (Another perspective: we *want* to accept this program, as rejecting it would break the abstraction boundary.)

```ocaml
module type S = sig
  type ('a : any) t : any
end

module F(Id : S) = struct
  type t = #{ t : t Id.t }
end

module Id : S = struct
  type ('a : any) t = 'a
end

module M = F(Id)
```

If the functor parameter exposes `t`'s definition or abstracts `t` as a type with `layout_of 'a`, it should work similarly to the second example from my last comment: we'll reject `F.t` by considering `t Id.t` to contain `t` via its `layout_of` layout.
```ocaml
module type S = sig
  type ('a : any) t = 'a (* changed line *)
end

module F(Id : S) = struct
  type t = #{ t : t Id.t }
end
(* Error: the definition of t contains recursion without boxing
     t contains t Id.t
     t Id.t contains t
*)
```
Does this sound right?

lpw25:
My concern was with your `M.t` above. It is cyclic in the way you are worried about and visibly so. Will that cause problems for `constrain_type_jkind`?

rtjoa:
I see the problem now, thanks. Current best guess for the fix: we consider an `any` type not in the recursive group to contain all of its type parameters.

rtjoa:
> Current best guess for the fix: we consider a type not in the recursive group whose layout contains `any` to contain all of its type parameters.

I made this change to future-proof against unboxed records with unrepresentable layouts, and refactored so we can more easily support `layout_of` later.

ccasin:
> I see the problem now, thanks. Current best guess for the fix: we consider a type not in the recursive group whose layout contains `any` to contain all of its type parameters.

I thought about it briefly and didn't come up with any examples where this heuristic misses anything or is overly conservative in a way that made me very sad (though it is certainly conservative).

I propose we (1) have Ryan check that this isn't so conservative that it rejects andything in our code base and then (2) merge this PR, unless @lpw25 can come up with another counter example. It will be much easier to figure out whether a better solution exists once we start working on `layout_of`, so as long as we're conservative I think we shouldn't invest a ton of time in perfecting this.

rtjoa:
> (1) have Ryan check that this isn't so conservative that it rejects anything in our code base

The check doesn't seem to.

I rebased this branch onto `5.2.0minus-4` as [`rtjoa.5.2.0minus-4-with-unboxed-recursion-check`](https://github.com/ocaml-flambda/flambda-backend/tree/rtjoa.5.2.0minus-4-with-unboxed-recursion-check), [very slightly modifying it](https://github.com/ocaml-flambda/flambda-backend/commit/7ab753476cb470ddc271fb7582ab9a081d47be58) because `5.2.0minus-4` doesn't have unboxed records. The check still works the same - many of the tests could be adapted to `[@@unboxed]` records, see 7e2da96e8a588847626c3f19db5a752fbd84a5f0 and 31b062e07011e79d5100c6437daff6920c909d7d. I then used this to build our code base (see `with-unboxed-recursion-check` internally).

typing/typedecl.ml
```diff
@@ -2065,6 +2066,105 @@ let check_well_founded_decl  ~abs_env env loc path decl to_check =
...
+
+      If [type 'a t = u] is in the recursive group then ['a t] contains [u].
+
+   3. Unboxed tuples contain their components.
+
+      [#(u_1 * ...)] contains all [u_i].
+
+   4. Types not in the group contain the parameters indicated by their layout.
+
+      ['a t] contains ['a] if [layout_of 'a] or [any] occurs in ['a t]'s layout.
```
goldfirere, line 2097: I find this heuristic to be unexpected. Maybe add a sentence explaining the need for it?

goldfirere (approved):
This looks all good from my perspective. I still think we'll eventually switch to use layout variables, but there's no reason not to merge this in the meantime.

PR #3410, opened 2024-12-27
Faithfully print with-kinds by reconstructing modalities for types

glittershark:
Reconstruct a more faithful representation of `with`-kinds when constructing outcome jkinds from `Jkind.t`s, using constant modalities on axes to omit types from axes which don't have those types in their baggage.

This approach is pretty dumb (we just walk the baggage of each axis building up a map tracking which types are mentioned on which axes), but from reading the existing test cases it seems to do a pretty good job of constructing the jkind that the user would have written down themselves - plus it's vastly better than the previous approach of just printing a kind that wouldn't even parse, let alone be correct.

Plausibly this approach would need to be rethought completely once we introduce non-constant/identity modalities, but those are far enough off that I claim this should serve us fine for the time being.

Review the whole diff, rather than individual commits.

### DIFF
--- a/testsuite/tests/parsetree/source_jane_street.ml
+++ b/testsuite/tests/parsetree/source_jane_street.ml
@@ -1117,19 +1117,30 @@ type 'a contended_with_int : immutable_data with 'a @@ contended with int
 
 [%%expect{|
 type 'a list
-  : value mod many with 'a uncontended with 'a portable with 'a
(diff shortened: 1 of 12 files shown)


glittershark:
not quite sure what's going on with the test failure here; hopefully this is reviewable regardless?

goldfirere (approved):

PR #3411, opened 2024-12-29
horrible quadratic baggage

glittershark:
let's see how horrible this actually is

### DIFF
--- a/parsing/ast_helper.ml
+++ b/parsing/ast_helper.ml
@@ -142,7 +142,7 @@ module Typ = struct
         | Default as x -> x
         | Abbreviation _ as x -> x
         | Mod (jkind, modes) -> Mod (loop_jkind jkind, modes)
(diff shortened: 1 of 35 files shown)


glittershark:
at least insofar as the compiler test suite is a reasonable performance benchmark for the compiler, this is... not actually terrible?

before:
```
user=697.35s system=219.18s cpu=1161% total=1:18.89
```

after:
```
user=812.91s system=217.03s cpu=891% total=1:55.51
```

stedolan:
> at least insofar as the compiler test suite is a reasonable performance benchmark for the compiler

for future reference: it really, really isn't a good perf benchmark for the compiler. It's dominated by weird things like e.g. regression tests for old obscure GC bugs that compact a million times. the compiler _build_ (main, when using the new compiler) is a better benchmark of compiler performance.

PR #3412, opened 2024-12-30
Vectorizer: check register compatibility

gretay-js:
When joining vectorized computations, check that registers used as address arguments in one computation are not vectorized in the other comptutation.

### DIFF
--- a/backend/cfg/vectorize.ml
+++ b/backend/cfg/vectorize.ml
@@ -2725,10 +2725,32 @@ end = struct
           t1.new_positions t2.new_positions
     }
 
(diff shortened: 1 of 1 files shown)


xclerc:
It looks correct, but would it be easy to add a test
showing how main is currently inadequate?

gretay-js:
> It looks correct, but would it be easy to add a test showing how main is currently inadequate?

I tried to minimize the example, but it's a bit hard to reproduce as it involves first class modules inlining and ppx.

xclerc (approved):
OK, let's merge it as-is then.

xclerc:
(Just for the record: the way the issue was
spotted was a rejection by the register
allocation validator.)

PR #3389, opened 2024-12-18
CI: simplify the regalloc jobs

xclerc:
As per title; the parameters removed
by this pull requests have been made
the defaults a while back.

PR #3414, opened 2024-12-30
Separate test for vectorizer in the CI

gretay-js:
Tests the vectorizer in the CI by building the compiler itself and running the tests with the vectorizer on. This is in addition to the dedicated vectorizer tests under `flambda-backend/tests/backend/vectorizer`.


xclerc:
I think it should be rebased onto #3389.

gretay-js:
This PR also removes the vectorizer from "gi" build.

xclerc:
Sure, but I think we should merge #3389 first.

xclerc (approved):

PR #3415, opened 2024-12-30
updated cmm_helpers interface to be more amenable to adding other int sizes

jvanburen:
This is to prepare for adding small integer types
This just changes the interface - it doesn't really update any code.

The only code change is in `safe_div_bi`, instead of taking an integer size, it explicitly takes an optional flag that tells us whether division could overflow. This is not only more clear, but the function is only used in a couple places so it's easy to verify that the new behavior is correct.


backend/cmm_helpers.ml
```diff
@@ -749,35 +751,46 @@ let mod_int c1 c2 is_safe dbg =
 (* Division or modulo on boxed integers. The overflow case min_int / -1 can
    occur, in which case we force x / -1 = -x and x mod -1 = 0. (PR#5513). *)
 
-let is_different_from x = function
-  | Cconst_int (n, _) -> n <> x
-  | Cconst_natint (n, _) -> n <> Nativeint.of_int x
-  | _ -> false
+(* Division or modulo on boxed integers. The overflow case min_int / -1 can
+   occur, in which case we force x / -1 = -x and x mod -1 = 0. (PR#5513). *)
```
Gbury, line 755: maybe add in a comment that the specified behavior is reached via the `mkop` and `mkm1` functions, which are used to build the correct cmm expressions respectively  in the usual case, and in the case where the divisor is `-1` (and the arch crashes on overflow).

backend/cmm_helpers.ml
```diff
@@ -749,35 +751,46 @@ let mod_int c1 c2 is_safe dbg =
...
+                  dbg,
+                  c,
+                  dbg,
+                  mkm1 dividend dbg,
+                  dbg,
+                  Any )))
 
-let safe_div_bi is_safe =
-  safe_divmod_bi div_int Any is_safe (fun c1 dbg ->
+let safe_div_bi =
```
Gbury, line 789: I wouldn't mind an eta-expension here for clarity (and it might also be generate slightly better code though that's not a real concern here I'd say).

backend/cmm_helpers.ml
```diff
@@ -749,35 +751,46 @@ let mod_int c1 c2 is_safe dbg =
...
+  safe_divmod_bi div_int (fun c1 dbg ->
       Cop (Csubi, [Cconst_int (0, dbg); c1], dbg))
 
-let safe_mod_bi is_safe =
-  safe_divmod_bi mod_int Any is_safe (fun _ dbg -> Cconst_int (0, dbg))
+let safe_mod_bi = safe_divmod_bi mod_int (fun _ dbg -> Cconst_int (0, dbg))
```
Gbury, line 793: same as remark for the eta-expansion

backend/cmm_helpers.ml
```diff
@@ -2010,6 +2054,40 @@ let zero_extend_63 dbg e =
...
+  then e
+  else
+    match bits with
+    | 63 -> low_63 dbg e
+    | 32 -> low_32 dbg e
+    | bits -> Misc.fatal_errorf "low_bits not implemented for %d bits" bits
+
+let ignore_low_bits ~bits ~dbg:(_ : Debuginfo.t) e =
+  assert (0 <= bits && bits <= arch_bits);
+  if bits = 0 then e else ignore_low_bit_int e
```
Gbury, line 2089: why not check that `bits` here is 1, as is done for the `zero_extend` and `sign_extend` functions ? I agree that the implementation is technically correct, since this is only used for optimizations, and thus "ignoring" less bits than requested is not wrong, but still, it would maybe be better in this iteration to check that we only ever ignore one bit ?

backend/cmm_helpers.mli
```diff
@@ -381,28 +380,18 @@ val bigarray_elt_size_in_bytes : Lambda.bigarray_kind -> int
...
 
-(** Zero extend from 32 bits to the word size *)
-val zero_extend_32 : Debuginfo.t -> expression -> expression
+(** Simplify the given expression knowing that bits other than the low [bits] bits will be
+    irrelevant *)
+val low_bits : bits:int -> dbg:Debuginfo.t -> expression -> expression
 
-(** Operations on 63-bit integers. These may only be used for compilation to
-    64-bit targets. *)
+val sign_extend : bits:int -> dbg:Debuginfo.t -> expression -> expression
```
Gbury, line 392: missing doc comment here and for `zero_extend` (especially considering that the "old" functions had doc comments)

backend/cmm_helpers.mli
```diff
@@ -719,13 +702,14 @@ val create_ccatch :
   body:Cmm.expression ->
   Cmm.expression
 
+(** Shift operations. take as first argument a tagged caml integer, and as
+    second argument an untagged machine intger which is the amount to shift the
+    first argument by. *)
```
Gbury, line 707: maybe also specify that the result is tagged ?

backend/cmm_helpers.mli
```diff
@@ -1198,57 +1182,26 @@ val unboxed_int64_or_nativeint_array_set :
...
-  Debuginfo.t ->
-  expression
-
-val get_field_unboxed_float32 :
-  Asttypes.mutable_flag ->
-  block:expression ->
-  index:expression ->
-  Debuginfo.t ->
-  expression
+    blocks} [immediate_or_pointer] is not needed as the layout is implied from the name,
```
Gbury, line 1185: maybe add full break lines (including a white line so that ocamlformat doesn't break everything) so that the doc title (in brackets) is more distinguishable from the explanation text ?

backend/cmm_helpers.mli
```diff
@@ -1198,57 +1182,26 @@ val unboxed_int64_or_nativeint_array_set :
...
+    and [initialization_or_assignment] is not needed as unboxed ints can always be
+    assigned without caml_modify (etc.). *)
```
Gbury, line 1187: Even though i understand who you reached this point, it'as a bit weird to specify why these function do not take some arguments (all the more since you don't specify the arguments they do take).

middle_end/flambda2/to_cmm/to_cmm_primitive.ml
```diff
@@ -130,66 +130,71 @@ let block_load ~dbg (kind : P.Block_access_kind.t) (mutability : Mutability.t)
...
+    C.float_array_set block index new_value dbg
+  | Mixed { field_kind = Flat_suffix field_kind; shape; _ } ->
+    let memory_chunk : Cmm.memory_chunk =
+      match field_kind with
+      | Tagged_immediate -> Word_int
+      | Naked_float -> Double
+      | Naked_float32 -> Single { reg = Float32 }
+      | Naked_int32 -> Thirtytwo_signed
+      | Naked_vec128 -> Onetwentyeight_unaligned
+      | Naked_int64 | Naked_nativeint -> Word_int
```
Gbury, line 190: this code could be factored out into a function from the `field_kind` to the `memory_chunk` and used for both the `get` and `set` primitives.

backend/cmm_helpers.mli
```diff
@@ -719,13 +696,17 @@ val create_ccatch :
   body:Cmm.expression ->
   Cmm.expression
 
+(** Shift operations.
+    Inputs: a tagged caml integer and an untagged machine integer.
+    Outputs: a tagged caml integer.
+    ake as first argument a tagged caml integer, and as
```
Gbury, line 702:
```suggestion
    Take as first argument a tagged caml integer, and as
```

backend/cmm_helpers.mli
```diff
@@ -1197,58 +1178,31 @@ val unboxed_int64_or_nativeint_array_set :
...
+    The [index_in_words] should be an untagged integer.
 
-val get_field_unboxed_float32 :
-  Asttypes.mutable_flag ->
-  block:expression ->
-  index:expression ->
-  Debuginfo.t ->
-  expression
+    In constrast to [setfield] and [setfield_computed], [immediate_or_pointer] is not
+    needed as the layout is implied from the name, and [initialization_or_assignment] is
```
Gbury, line 1187: the layout is not implied from the name anymore, but rather by the `memory_chunk` given as argument

utils/misc.mli
```diff
@@ -462,6 +462,10 @@ val log2: int -> int
        (** [log2 n] returns [s] such that [n = 1 lsl s]
            if [n] is a power of 2*)
 
+val log2_nativeint: nativeint -> int
+(** [log2_nativeint n] returns [s] such that [n = Nativeint.shift_left 1n s]
+    if [n] is a power of 2*)
```
Gbury, line 467: this is quite an under-specification (technically, the doc says nothing about the returned value if the input is not a power of 2, although I'm relatively certain the function is used on such arguments....), but since the other functions have the same style of doc, I guess it's okay-ish ?

Gbury (changes requested):

backend/cmm_helpers.ml
```diff
@@ -633,16 +635,54 @@ let raise_symbol dbg symb =
...
+    if n < 0n
     then
+      if n = Nativeint.min_int
+      then Cop (Ccmpi Ceq, [c1; Cconst_natint (Nativeint.min_int, dbg)], dbg)
+      else
+        neg_int
+          (div_int ?dividend_cannot_be_min_int c1
+             (Cconst_natint (Nativeint.neg n, dbg))
+             dbg)
+          dbg
```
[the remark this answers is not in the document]
jvanburen: to be honest, I don't actually know :disappointed: I'm preserving the behavior from before.

backend/cmm_helpers.ml
```diff
@@ -28,6 +30,12 @@ type arity =
 
 (* Local binding of complex expressions *)
 
+let sequence x y =
+  match x, y with
+  | (Cvar _ | Cconst_int _ | Cconst_natint _ | Ctuple []), y -> y
+  | x, Ctuple [] -> x
```
Gbury, line 36: I'm not sure this is correct: consider for instance something along the lines of `sequence (const_int 0) (Ctuple [])` (whose semantics is dubious, but still), this line would change the "result" of the cmm expression.

For what it's worth, I sure hope that we do not produce such expressions (whose semantics and usefulness is relatively unclear), and therefore this should not matter, but I'd rather remove this unless you have a compelling argument.

backend/cmm_helpers.ml
```diff
@@ -673,14 +681,18 @@ let rec div_int ?dividend_cannot_be_min_int c1 c2 dbg =
     if n < 0n
     then
       if n = Nativeint.min_int
-      then Cop (Ccmpi Ceq, [c1; Cconst_natint (Nativeint.min_int, dbg)], dbg)
+      then
+        (* integer division by min_int always returns 0 unless the dividend is
+           also min_int, in which case it's 1. This is the same as comparing
+           against min_int. *)
```
Gbury: So, this relies on the fact that comparisons are expected to return `0` or `1` for `false` and `true` resp. right ? this would be useful to document somewhere (let's say `cmm.ml` or `cmm.mli`), since even if we assume C-like conventions I think that the only requirements is `0` for false and anything else for `true` (which someone may be tempted to use at one point for some optimization).

backend/cmm_helpers.ml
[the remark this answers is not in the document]
Gbury: oh apologies, because of all of the re-ordering (and replacing substractions from zero with negation) I missed that this was not an actual change. I guess it's okay for now, though if you have time to dive into this and find out why, I wouldn't object to some documentation of why it's done like that, ^^

backend/cmm_helpers.ml
```diff
@@ -670,49 +678,39 @@ let make_safe_divmod operator ~if_divisor_is_negative_one
...
-          dbg
-    else if is_power_of_2 n
+      (* integer division by min_int always returns 0 unless the dividend is
+         also min_int, in which case it's 1. This is the same as comparing
+         against min_int. *)
+      Cop (Ccmpi Ceq, [c1; Cconst_natint (Nativeint.min_int, dbg)], dbg)
+    else if is_power_of_2 divisor
     then
-      let l = Misc.log2_nativeint n in
+      let l = Misc.log2_nativeint divisor in
```
Gbury, line 697: `log2_nativeint n` works correctly if `n > 0` according to its doc (and I admit that I don't have the motivation to check it works for negative integers right now, or what would be the exact spec), but with the new code, this is not guaranteed for `divisor` anymore here (at least not obviously like before), unless `is_power_of_2` also ensures that its argument is positive, in which case it might be useful to either make it part of the function name, or at least put a remainder in comments here.

backend/cmm_helpers.ml
```diff
@@ -722,16 +720,22 @@ let rec div_int ?dividend_cannot_be_min_int c1 c2 dbg =
          t = shift-right-signed(t, p)
 
          res = t + sign-bit(c1) *)
```
Gbury, line 722: you should update the algorithm to:
- match the new names below (particular the renamings such as `t -> q`, `c1 -> n`)
- make the algorithm reflect what is actually computed now as there are a few new cases (related to the handling of negative dividends if I'm not mistaken)

Gbury (changes requested):

backend/cmm_helpers.ml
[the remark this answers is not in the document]
jvanburen: It's because the original author didn't implement the signed division algorithm from the reference they cite, they only implmented unsigned division optimization. I'll add a comment.

backend/cmm_helpers.ml
jvanburen: Unless i'm mistaken, the previous code had this issue too, no?
```
let sequence x y =
  match x, y with
  | Ctuple [], _ -> y
  | _, Ctuple [] -> x
  | _, _ -> Csequence (x, y)

```

Gbury: indeed, and it probably means that this particular part of code is likely unused in practice, but since we're touching it, it makes sense to fix it

jvanburen: I changed it back and used bind instead

backend/cmm_helpers.ml
jvanburen: I can just use an if statement and rely on it getting optimized later

backend/cmm_helpers.ml
jvanburen: i added a comment - this is only ever called on positive powers of 2 where it clearly works

backend/cmm_helpers.ml
```diff
@@ -731,28 +677,33 @@ let div_int ?dividend_cannot_be_min_int c1 c2 dbg =
...
+  | _, Some 0n -> divide_by_zero c1 ~dbg
   | _, Some (1n | -1n) ->
     if_divisor_is_positive_or_negative_one ~dividend:c1 ~dbg
   | Some n1, Some n2 -> natint_const_untagged dbg (Nativeint.rem n1 n2)
   | _, Some n ->
     if n = Nativeint.min_int
     then
-      (* similarly to the division by min_int almost always being 0, modulo
+      (* [divisor] must be positive be here since we already handled zero and
+         min_int (the only negative power of 2).
```
Gbury, line 704: this comment is probably misplaced, given that this section of code deals with `n = min_int` (and `n` is `[divisor]` unless I'm mistaken) which is usually negative. This should instead be a few lines below.

testsuite/tests/basic/division_by_constant.ml
```diff
@@ -0,0 +1,366 @@
...
+  assert (remainder = rem dividend divisor);
+  assert (dividend = add (mul quotient divisor) remainder);
+  assert
+    (if divisor = min_int
+     then remainder <> min_int
+     else abs remainder < abs divisor);
+  let[@inline] sign x = if x < 0n then -1 else if x = 0n then 0 else 1 in
+  assert (remainder = 0n || sign remainder = sign dividend)
+
+let[@inline always] [@specialise always] check ~divisor =
```
Gbury, line 37: please use two different names for the `check` functions (e.g. `check_one` and `check` ?)

testsuite/tests/basic/division_by_constant.ml
```diff
@@ -0,0 +1,366 @@
...
+  ListLabels.iter test_cases ~f:(fun dividend ->
+    (Sys.opaque_identity check)
```
Gbury, line 39: No need for the `Sys.opaque_identity`, considering `check` is already `[@inline never]`

testsuite/tests/basic/division_by_constant.ml
Gbury: also, you can remove the `[@specialise always]` which should be useless/redundant given that the function is `[@inline always]` (also, specialisation is not yet done by flambda2 anyways for now...)

testsuite/tests/basic/division_by_constant.ml
```diff
@@ -0,0 +1,366 @@
...
+      ~dividend
+      ~divisor
+      ~quotient:(div dividend divisor)
+      ~remainder:(rem dividend divisor)
+  )
+
+(* manual tests *)
+let () = check ~divisor:3n
+let () = check ~divisor:(-3n)
```
Gbury, line 48: from what I read in Hacker's delight, `5` and `7` seem to also be good candidates for manual testing

testsuite/tests/basic/division_by_constant.ml
```diff
@@ -0,0 +1,366 @@
...
+let () = check ~divisor:65537n
+let () = check ~divisor:(-65537n)
+
+(* checks generated by test_cases. These get inlined so the division by a constant is
+   replaced *)
```
Gbury, line 53: this comment is quite confusing:
- you could specify that it was generated by a manual run of the `test_cases` function to have a hard-coded list of semi-random numbers (whereas all divisors are tested on the list generated by `test_cases`)
- the comment about `check` being inlined also applies to the "manual" tests above, so stating it here is confusing. You should instead add a comment to this effect on the definition of the `check` function explaining how the `[@inline]` annotations are used to ensure that the divisor is a compile-time known constant but not the dividend. That would also help specify exactly what is being tested by this file.

testsuite/tests/basic/division_by_constant.ml
Gbury: also, `1`, `-1`, `min_int` seems like good candidates since they have special code paths in the compiler.

In the same vein of interesting edge cases, we also have `max_int`, and `0` (which may be slightly more complex to check since it raises, but a good test nonetheless).

Lastly, you could add a few manual tests where the dividend and/or divisor expressions have side-effects (e.g. `((fun x -> (print_endline "foo"; x) / (print_endline "bar"; 3))[@inlined never]) 42`

Gbury (approved):

PR #3417, opened 2024-12-31
Refactor CMM unboxed field getters/setters

jvanburen:
This will be useful once we have unboxed integers of different sizes.

### DIFF
--- a/backend/cmm_helpers.ml
+++ b/backend/cmm_helpers.ml
@@ -924,7 +924,20 @@ let complex_im c dbg =
 
 (* Unit *)
 
-let return_unit dbg c = Csequence (c, Cconst_int (1, dbg))
+let return_unit dbg c =
+  match c with
+  | Csequence (_, Cconst_int (1, _)) as c -> c
+  | c -> Csequence (c, Cconst_int (1, dbg))
+
+let memory_chunk_width_in_bytes : memory_chunk -> int = function
+  | Byte_unsigned | Byte_signed -> 1
+  | Sixteen_unsigned | Sixteen_signed -> 2
+  | Thirtytwo_unsigned | Thirtytwo_signed -> 4
+  | Single { reg = Float64 | Float32 } -> 4
+  | Word_int -> size_int
+  | Word_val -> size_addr
+  | Double -> size_float
+  | Onetwentyeight_unaligned | Onetwentyeight_aligned -> size_vec128
 
 let strided_field_address ptr ~index ~stride dbg =
   if index * stride = 0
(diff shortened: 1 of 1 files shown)


Gbury:
Will review on Thursday after the new year, ^^

backend/cmm_helpers.ml
```diff
@@ -919,32 +919,29 @@ let complex_im c dbg =
 
 (* Unit *)
 
-let return_unit dbg c = Csequence (c, Cconst_int (1, dbg))
+let return_unit dbg c =
+  match c with
+  | Csequence (_, Cconst_int (1, _)) as c -> c
```
Gbury, line 924: the `as c` is unneeded since that's already the name of the value being pattern-matched.

Gbury (commented):
This PR seems fine however the test failure surprises me, so either I have missed something during my review or the test is a bit flaky (which might be the case since it's about finalisers).

Let's wait to see if @jvanburen can see why the test is failing and if it's related to this PR.

backend/cmm_helpers.ml
jvanburen: This feels kinda nit picky. I prefer the `as` pattern since it makes it clearer what we're returning

Gbury:
@jvanburen can you rebase (and fix the conflict) so that the github interface properly displays the actual changes of this PR ?

backend/cmm_helpers.ml
Gbury: alright, let's leave it like that then

backend/cmm_helpers.ml
```diff
@@ -1410,147 +1407,50 @@ let unboxed_int64_or_nativeint_array_set ~has_custom_ops arr ~index ~new_value
...
+    then
+      Misc.fatal_error
+        "Unboxed non-word-size fields are only supported on little-endian \
+         architectures";
+    (* CR layouts v5.1: We will need to vary log2_size_addr, among other things,
+       when small fields are efficiently packed. *)
+    let field_address =
+      array_indexing log2_size_addr block index_in_words dbg
+    in
+    let newval = if size_in_bytes = 4 then low_32 dbg newval else newval in
```
Gbury, line 1451: Maybe this deserves a CR for the future to use `low ~bits:..` (for all things such that `size_in_bytes < size_int`) when it's available ?

Gbury (approved):

backend/cmm_helpers.ml
jvanburen: I think i end up doing that in a descendant PR so i'm just gonna punt on that

PR #3420, opened 2025-01-02
Add `Capsule.with_password`

TheNumbat:
`Capsule.protect f` and the new `Capsule.with_password f` should provide `f` with a password representing the fresh capsule. 
This allows us to create local-password-protected data in `f` and pass it downward. 

Additionally, `protect` will now unwrap `Encapsulated` exceptions that are in the same capsule as the provided password.
`with_password` will do the same, but destroys the `'k` capsule before returning the exn directly.

The two functions are implemented separately to avoid the overhead of creating/destroying the mutex in `with_password`.

### DIFF
--- a/otherlibs/stdlib_alpha/capsule.ml
+++ b/otherlibs/stdlib_alpha/capsule.ml
@@ -85,6 +85,8 @@ module Password : sig
      [void] can't be used for function argument and return types yet. *)
   type 'k t : value mod portable many unique uncontended
 
+  type packed = P : 'k t -> packed
+
   (* Can break the soundness of the API. *)
   val unsafe_mk : 'k Name.t -> 'k t @@ portable
   val name : 'k t @ local -> 'k Name.t @@ portable
@@ -102,6 +104,8 @@ module Password : sig
   val shared : 'k t @ local -> 'k Shared.t @ local @@ portable
 end = struct
   type 'k t = 'k Name.t
+
+  type packed = P : 'k t -> packed
 
   let unsafe_mk name = name
   let name t = t
@@ -374,10 +378,10 @@ let create_with_rwlock () =
 
 exception Protected : 'k Mutex.t * (exn, 'k) Data.t -> exn
 
-let protect_local (f : 'k. 'k Password.t @ local -> _ @ local) = exclave_
+let protect_local f = exclave_
   let (P name) = Name.make () in
   let password = Password.unsafe_mk name in
-  try f password with
+  try f (Password.P password) with
   | Encapsulated (inner, data) as exn ->
     (match Name.equality_witness name inner with
      | Some Equal ->
@@ -386,10 +390,10 @@ let protect_local (f : 'k. 'k Password.t @ local -> _ @ local) = exclave_
   | exn ->
     reraise (Protected ({ name; mutex = M.create (); poisoned = false }, Data.unsafe_mk exn))
 
-let protect (f : 'k. 'k Password.t -> _) =
+let protect f =
   let (P name) = Name.make () in
   let password = Password.unsafe_mk name in
-  try f password with
+  try f (Password.P password) with
   | Encapsulated (inner, data) as exn ->
     (match Name.equality_witness name inner with
      | Some Equal ->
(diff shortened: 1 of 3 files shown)


otherlibs/stdlib_alpha/capsule.mli
```diff
@@ -416,15 +416,24 @@ exception Encapsulated : 'k Name.t * (exn, 'k) Data.t -> exn
...
+val protect : ('k. 'k Password.t @ local -> 'a) @ local portable -> 'a @@ portable
 (** [protect f] runs [f] in a fresh capsule. If [f] returns normally, [protect]
-    merges this capsule into the caller's capsule. If [f] raises, [protect]
-    raises [Protected], giving the caller access to the encapsulated exception. *)
+    merges this capsule into the caller's capsule. If [f] raises an [Encapsulated]
+    exception in the capsule ['k], [protect] unwraps the exception and re-raises
+    it as [Protected]. If [f] raises any other exception, [protect] re-raises
+    it as [Protected]. *)
+
+val with_password : ('k. 'k Password.t @ local -> 'a) @ local portable -> 'a @@ portable
```
tdelvecchio-jsc: Specify behavior of handling exceptions, given that this has the same exact type-signature as `protect`.

otherlibs/stdlib_alpha/capsule.ml
```diff
@@ -377,14 +374,46 @@ let create_with_rwlock () =
...
+    (match Name.equality_witness name inner with
+     | Some Equal ->
+       reraise (Protected ({ name; mutex = M.create (); poisoned = false }, data))
+     | None -> reraise exn)
+  | exn ->
+    reraise (Protected ({ name; mutex = M.create (); poisoned = false }, Data.unsafe_mk exn))
 
-let protect f =
-  try f () with
+let protect (f : 'k. 'k Password.t -> _) =
```
tdelvecchio-jsc: Implement `protect` as `(protect_local { global }).global`.

otherlibs/stdlib_alpha/capsule.ml
```diff
@@ -377,14 +374,46 @@ let create_with_rwlock () =
...
+  let (P name) = Name.make () in
+  let password = Password.unsafe_mk name in
+  try f password with
+  | Encapsulated (inner, data) as exn ->
+    (match Name.equality_witness name inner with
+     | Some Equal -> reraise (Data.unsafe_get data)
+     | None -> reraise exn)
+  | exn -> reraise exn
+
+let with_password (f : 'k. 'k Password.t -> _) =
```
tdelvecchio-jsc: Implement `with_password` as `(with_password_local { global }).global`.

otherlibs/stdlib_alpha/capsule.mli
```diff
@@ -416,15 +421,24 @@ exception Encapsulated : 'k Name.t * (exn, 'k) Data.t -> exn
...
-val protect
-  :  (unit -> 'a @ portable contended) @ local portable
-  -> 'a @ portable contended
-  @@ portable
+val protect : (Password.packed @ local -> 'a) @ local portable -> 'a @@ portable
 (** [protect f] runs [f] in a fresh capsule. If [f] returns normally, [protect]
-    merges this capsule into the caller's capsule. If [f] raises, [protect]
-    raises [Protected], giving the caller access to the encapsulated exception. *)
+    merges this capsule into the caller's capsule. If [f] raises an [Encapsulated]
+    exception in the capsule ['k], [protect] unwraps the exception and re-raises
```
tdelvecchio-jsc, line 433: `['k]` no longer references anything

tdelvecchio-jsc (changes requested):

otherlibs/stdlib_alpha/capsule.ml
TheNumbat: `Stdlib.Modes` doesn't exist in main yet, but done

tdelvecchio-jsc (approved):
Looks good.

otherlibs/stdlib_alpha/capsule.mli
lpw25: It should indeed do that, but the signatures shouldn’t be identical: this one doesn’t need the `portable` on the closure argument.

lpw25:
Why the change away from polymorphic parameters? They should be more convenient. Is there some public release issue?

lpw25:
I think the current primitives are a slightly awkward mix. Probably it would be better to start with a minimal set and then add some convenience ones on top. I think that is something like:
```
val access :
  'k Password.t @ local
  -> ('k Access.t -> 'a @ portable contended) @ local portable
  -> 'a @ contended
  @@ portable

val protect :
    ('k. 'k Password.t @ local -> 'a) @ local
    -> 'a @@ portable

val unprotect :
    (unit -> ‘a) @ local p
    -> ‘a
```
where `access` uses a password to access a capsule and wraps exceptions in `Encapsulated`, `protect` creates a fresh capsule and password and converts any `Encapsulted` exceptions for that password into `Protected` exceptions, `unprotect` unwraps any raised `Protected` exceptions into the current capsule by destroying the associated mutex.

Then `with_password` is a more efficient implementation of the composition of `protect` and `unprotect` — we should include it here for the improved performance but can document it as just a composition. And the old version of `protect` can be implemented by composing `protect` and `access`.

TheNumbat:
I removed the polymorphic parameters since base needs to use this API and can't have non-erasable extensions. (@tdelvecchio-jsc is there another way we should structure this?)

I'm not sure we need to expose `unprotect` - it does seem more primitive, but is there any reason to use it other than composing it with `protect` to get `with_password`? It's also simple to implement wherever needed (destroy + expose). Or maybe we should move protect/unprotect into the password module and only expose with_password at top level.

tdelvecchio-jsc:
If we want to keep it with universal instead of existential, we'd have to use a record like `type 'a protected = { protect : 'k. 'k Password.t @ local -> 'a }`.

lpw25:
I agree we probably don't need to expose `unprotect` since it can be implemented trivially. But I do think that we should expose the `protect` that gives you a password and takes `local` closure rather than the one that takes a `local portable` closure.

PR #3421, opened 2025-01-02
Make Capsule preserve wrapped exception backtraces

TheNumbat:
Reraising a newly constructed exception creates a fresh backtrace. 
We want to preserve the existing backtrace when encapsulating or protecting an exception raised from a capsule function.


tdelvecchio-jsc (approved):
Looks good.

lpw25:
We may need to be careful here. You might be getting quadratic performance by repeatedly copying the backtraces in and out of the backtrace buffer. It would probably be better to add a new raise primitive that does the right thing.

TheNumbat:
This sort of quadratic behavior is already present in `Fun.protect` and `Base.Exn.protect`, but I could see it being more of a problem when used in nested capsules/parallel tasks. In my use case the backtraces don't grow at each reraise, so I think it's okay to merge this for now, but I'll add a CR about adding a raise-with-current-backtrace primitive.

PR #3423, opened 2025-01-03
 Added Cmm_helpers.Scalar_type

jvanburen:
`Cmm_helpers.Scalar_type` provides utilities for converting between integers types of different widths and signedness. This is in preparation for adding unboxed small integer types.

### DIFF
--- a/backend/cmm_helpers.ml
+++ b/backend/cmm_helpers.ml
@@ -4691,3 +4691,270 @@ let reperform ~dbg ~eff ~cont ~last_fiber =
       dbg )
 
 let poll ~dbg = return_unit dbg (Cop (Cpoll, [], dbg))
+
+module Scalar_type = struct
+  module Float_width = struct
+    type t = Cmm.float_width =
+      | Float64
+      | Float32
+
+    let[@inline] static_cast ~dbg ~src ~dst exp =
+      match src, dst with
+      | Float64, Float64 -> exp
+      | Float32, Float32 -> exp
+      | Float32, Float64 -> float_of_float32 ~dbg exp
+      | Float64, Float32 -> float32_of_float ~dbg exp
+  end
+
+  module Signedness = struct
+    type t =
+      | Signed
+      | Unsigned
+
+    let equal (x : t) (y : t) = x = y
+
+    let print ppf t =
+      match t with
+      | Signed -> Format.pp_print_string ppf "signed"
+      | Unsigned -> Format.pp_print_string ppf "unsigned"
+  end
+
+  module Bit_width_and_signedness : sig
+    (** An integer with signedness [signedness t] that fits into a general-purpose
+        register. It is canonically stored in twos-complement representation, in the lower
+        [bits] bits of its container (whether that be memory or a register), and is sign-
+        or zero-extended to fill the entire container. *)
+    type t [@@immediate]
+
+    val create_exn : bit_width:int -> signedness:Signedness.t -> t
+
+    val bit_width : t -> int
+
+    val signedness : t -> Signedness.t
+
+    val equal : t -> t -> bool
+  end = struct
+    (* [signedness t] is stored in the low bit of [t], and [bit_width t] is
+       stored in the remaining high bits of [t]. We use this encoding to fit [t]
+       into an immediate value. This is worth trying since we expect to create
+       one of these for ~every integer operation, so it should cut down on
+       garbage *)
+    type t = { bit_width_and_signedness : int } [@@unboxed]
+
+    let[@inline] equal { bit_width_and_signedness = x }
+        { bit_width_and_signedness = y } =
+      Int.equal x y
+
+    let[@inline] bit_width { bit_width_and_signedness } =
+      bit_width_and_signedness lsr 1
+
+    let[@inline] signedness { bit_width_and_signedness } =
+      match bit_width_and_signedness land 1 with
+      | 0 -> Signedness.Signed
+      | 1 -> Signedness.Unsigned
+      | _ -> assert false
+
+    let[@inline] int_of_signedness : Signedness.t -> int = function
+      | Signed -> 0
+      | Unsigned -> 1
+
+    let[@inline] create_exn ~bit_width ~signedness =
+      assert (0 < bit_width && bit_width <= arch_bits);
+      { bit_width_and_signedness =
+          (bit_width lsl 1) + int_of_signedness signedness
+      }
+  end
+
+  module Integral_type = struct
+    include Bit_width_and_signedness
+
+    let[@inline] with_signedness t ~signedness =
+      create_exn ~bit_width:(bit_width t) ~signedness
+
+    let[@inline] signed t = with_signedness t ~signedness:Signed
+
+    let[@inline] unsigned t = with_signedness t ~signedness:Unsigned
+
+    (** Determines whether [dst] can represent every value of [src], preserving sign *)
+    let[@inline] can_cast_without_losing_information ~src ~dst =
+      match signedness src, signedness dst with
+      | Signed, Signed | Unsigned, Unsigned -> bit_width src <= bit_width dst
+      | Unsigned, Signed -> bit_width src < bit_width dst
+      | Signed, Unsigned -> false
+
+    let[@inline] static_cast ~dbg ~src ~dst exp =
+      if can_cast_without_losing_information ~src ~dst
+      then
+        (* Since [Bit_width_and_signedness] represents sign- or zero-extended
+           expressions, this is a no-op *)
+        exp
+      else
+        match signedness dst with
+        | Signed -> sign_extend ~bits:(bit_width dst) exp ~dbg
+        | Unsigned -> zero_extend ~bits:(bit_width dst) exp ~dbg
+
+    let[@inline] conjugate ~outer ~inner ~dbg ~f x =
+      x
+      |> static_cast ~src:outer ~dst:inner ~dbg
+      |> f
+      |> static_cast ~src:inner ~dst:outer ~dbg
+  end
+
+  module Integer = struct
+    include Integral_type
+
+    let print ppf t =
+      Format.fprintf ppf "%a int%d" Signedness.print (signedness t)
+        (bit_width t)
+
+    let nativeint = create_exn ~bit_width:arch_bits ~signedness:Signed
+  end
+
+  (** An {!Integer.t} but with the additional stipulation that its container must
+      reserve its lowest bit to be 1. The [bit_width] field includes this bit. *)
+  module Tagged_integer = struct
+    include Integral_type
+
+    let[@inline] create_exn ~bit_width_including_tag_bit:bit_width ~signedness =
+      assert (bit_width > 1);
+      create_exn ~bit_width ~signedness
+
+    let immediate =
+      create_exn ~bit_width_including_tag_bit:arch_bits ~signedness:Signed
+
+    let[@inline] bit_width_including_tag_bit t = bit_width t
+
+    let[@inline] bit_width_excluding_tag_bit t = bit_width t - 1
+
+    let[@inline] untagged t =
+      Integer.create_exn
+        ~bit_width:(bit_width_excluding_tag_bit t)
+        ~signedness:(signedness t)
+
+    let[@inline] untag ~dbg t exp =
+      match signedness t with
+      | Signed -> asr_const exp 1 dbg
+      | Unsigned -> lsr_const exp 1 dbg
+
+    let print ppf t =
+      Format.fprintf ppf "tagged %a int%d" Signedness.print (signedness t)
+        (bit_width_excluding_tag_bit t)
+  end
+
+  module Integral = struct
+    type t =
+      | Untagged of Integer.t
+      | Tagged of Tagged_integer.t
+
+    let nativeint = Untagged Integer.nativeint
+
+    let[@inline] untagged_or_identity = function
+      | Untagged t -> t
+      | Tagged t -> Tagged_integer.untagged t
+
+    let signedness = function
+      | Untagged t -> Integer.signedness t
+      | Tagged t -> Tagged_integer.signedness t
+
+    let with_signedness t ~signedness =
+      match t with
+      | Untagged t -> Untagged (Integer.with_signedness t ~signedness)
+      | Tagged t -> Tagged (Tagged_integer.with_signedness t ~signedness)
+
+    let[@inline] signed t = with_signedness t ~signedness:Signed
+
+    let[@inline] unsigned t = with_signedness t ~signedness:Unsigned
+
+    let[@inline] equal x y =
+      match x, y with
+      | Untagged x, Untagged y -> Integer.equal x y
+      | Untagged _, _ -> false
+      | Tagged x, Tagged y -> Tagged_integer.equal x y
+      | Tagged _, _ -> false
+
+    let print ppf t =
+      match t with
+      | Untagged untagged -> Integer.print ppf untagged
+      | Tagged tagged -> Tagged_integer.print ppf tagged
+
+    let[@inline] can_cast_without_losing_information ~src ~dst =
+      Integer.can_cast_without_losing_information
+        ~src:(untagged_or_identity src) ~dst:(untagged_or_identity dst)
+
+    let static_cast ~dbg ~src ~dst exp =
+      match src, dst with
+      | Untagged src, Untagged dst -> Integer.static_cast ~dbg ~src ~dst exp
+      | Tagged src, Tagged dst -> Tagged_integer.static_cast ~dbg ~src ~dst exp
+      | Untagged src, Tagged dst ->
+        tag_int
+          (Integer.static_cast ~dbg ~src ~dst:(Tagged_integer.untagged dst) exp)
+          dbg
+      | Tagged src, Untagged dst ->
+        Integer.static_cast ~dbg
+          ~src:(Tagged_integer.untagged src)
+          ~dst
+          (Tagged_integer.untag ~dbg src exp)
+
+    let[@inline] conjugate ~outer ~inner ~dbg ~f x =
+      x
+      |> static_cast ~src:outer ~dst:inner ~dbg
+      |> f
+      |> static_cast ~src:inner ~dst:outer ~dbg
+  end
+
+  type t =
+    | Integral of Integral.t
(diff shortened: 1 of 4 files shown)


middle_end/flambda2/to_cmm/to_cmm_primitive.ml
```diff
@@ -559,66 +592,23 @@ let unary_float_arith_primitive _env dbg width op arg =
...
+  else
+    let src = numeric_of_standard_int_or_float src in
+    let dst = numeric_of_standard_int_or_float dst in
+    let extra =
+      match src, dst with
+      | Integral (Tagged src), Integral (Untagged dst)
+        when C.Scalar_type.Integer.equal
+               (C.Scalar_type.Tagged_integer.untagged src)
+               dst ->
+        Some (Env.Untag arg)
```
TheNumbat, line 606: Should the reverse return `Env.Tag`?

middle_end/flambda2/to_cmm/to_cmm_primitive.ml
```diff
@@ -628,124 +618,127 @@ let phys_equal _env dbg op x y =
...
+      if requires_sign_extended_operands
+      then operand
+      else
+        let bits =
+          match kind with
+          | Untagged untagged -> C.Scalar_type.Integer.bit_width untagged
+          | Tagged tagged ->
+            C.Scalar_type.Tagged_integer.bit_width_including_tag_bit tagged
+        in
+        C.low_bits ~bits operand ~dbg
```
TheNumbat, line 674: I see the tagged case can't ignore the low bit since `*_int_caml` expects the tag to be one. Could add a cr-someday mentioning that.

TheNumbat (commented):
Looks good, I'm confident the conversions are correct. Have you checked that this doesn't hurt codegen for static casts? We don't have a good way to test that automatically, but spot-checking a few things seems sufficient.

backend/cmm_helpers.ml
```diff
@@ -4426,3 +4426,261 @@ let reperform ~dbg ~eff ~cont ~last_fiber =
...
+        Misc.fatal_errorf "static_cast: casting %a to float is not implemented"
+          Integral.print src
+      else
+        unary (Cstatic_cast (Float_of_int dst)) ~dbg
+          (Integral.static_cast exp ~dbg ~src ~dst:float_of_int_arg)
+    | Float src, Integral dst -> (
+      match Integral.signedness dst with
+      | Unsigned ->
+        Misc.fatal_errorf
+          "static_cast: casting floats to unsigned values is undefined"
```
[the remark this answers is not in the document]
jvanburen: yes, at some point...

middle_end/flambda2/to_cmm/to_cmm_primitive.ml
```diff
@@ -522,34 +522,67 @@ let dead_slots_msg dbg function_slots value_slots =
...
+      Misc.fatal_error "Not yet implemented"
+    | Naked_immediate ->
+      (* This case should not have a sign extension, confusingly, because it
+         arises from the [Pbswap16] Lambda primitive. That operation does not
+         affect the sign of the resulting value. *)
+      C.bswap16 arg dbg
+    | Naked_int32 ->
+      C.sign_extend (C.bbswap Unboxed_int32 arg dbg) ~bits:32 ~dbg
+    | Naked_int64 ->
+      C.sign_extend (C.bbswap Unboxed_int64 arg dbg) ~bits:64 ~dbg
```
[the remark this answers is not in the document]
jvanburen: yeah they for sure don't generate any code

middle_end/flambda2/to_cmm/to_cmm_primitive.ml
jvanburen: I don't think you can ever usefully ignore the tag bit?

TheNumbat: yeah, never mind

middle_end/flambda2/to_cmm/to_cmm_primitive.ml
jvanburen: It doesn't build anymore :/

TheNumbat: I guess not then

backend/cmm_helpers.ml
```diff
@@ -4426,3 +4426,261 @@ let reperform ~dbg ~eff ~cont ~last_fiber =
...
+    let[@inline] is_promotable ~src ~dst =
+      match signedness src, signedness dst with
+      | Signed, Signed | Unsigned, Unsigned -> bit_width src <= bit_width dst
+      | Unsigned, Signed -> bit_width src < bit_width dst
+      | Signed, Unsigned -> false
+
+    let[@inline] static_cast ~dbg ~src ~dst exp =
+      if is_promotable ~src ~dst
+      then
+        (* since the values are already stored sign- or zero-extended, this is a
```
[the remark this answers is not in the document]
TheNumbat: I mean the comment should say that instead of stored

TheNumbat (approved):

mshinwell:
Should we e.g. examine the actual generated assembly for a few executables before and after this change?  I think probably yes.

backend/cmm_helpers.ml
```diff
@@ -4436,3 +4457,263 @@ let reperform ~dbg ~eff ~cont ~last_fiber =
...
+    let[@inline] is_promotable ~src ~dst =
+      match signedness src, signedness dst with
+      | Signed, Signed | Unsigned, Unsigned -> bit_width src <= bit_width dst
+      | Unsigned, Signed -> bit_width src < bit_width dst
+      | Signed, Unsigned -> false
+
+    let[@inline] static_cast ~dbg ~src ~dst exp =
+      if is_promotable ~src ~dst
+      then
+        (* since cmm expressions representing int32#s are stored sign- or
```
Gbury, line 4556: true, but it applies to other bit-width, and it's actually also because of the definition given in the documentation for `Bit_width_and_signedness` (which admittedly must be kept in sync with how we handle "small" integers in cmm/to_cmm).

backend/cmm_helpers.ml
```diff
@@ -4436,3 +4457,263 @@ let reperform ~dbg ~eff ~cont ~last_fiber =
...
+
+  let static_cast ~dbg ~src ~dst exp =
+    match src, dst with
+    | Integral src, Integral dst -> Integral.static_cast ~dbg ~src ~dst exp
+    | Float src, Float dst -> Float_width.static_cast ~dbg ~src ~dst exp
+    | Integral src, Float dst ->
+      let float_of_int_arg = Integral.nativeint in
+      if not (Integral.is_promotable ~src ~dst:float_of_int_arg)
+      then
+        Misc.fatal_errorf "static_cast: casting %a to float is not implemented"
```
Gbury, line 4684: Is that because the expected semantics of casting from ints to floats currently says that we should not lose bits from the source integer, or for another reason ?

Actually, I think this code could benefit from some comments stating the expected semantics of the casts: currently the integers casts are happy to lose some bits if the destination is smaller than the source, but it's not clear what are the expected semantics for floats to/from ints conversions.

backend/cmm_helpers.ml
```diff
@@ -4436,3 +4457,263 @@ let reperform ~dbg ~eff ~cont ~last_fiber =
...
+      else
+        unary (Cstatic_cast (Float_of_int dst)) ~dbg
+          (Integral.static_cast exp ~dbg ~src ~dst:float_of_int_arg)
+    | Float src, Integral dst -> (
+      match Integral.signedness dst with
+      | Unsigned ->
+        Misc.fatal_errorf
+          "static_cast: casting floats to unsigned values is undefined"
+      | Signed ->
+        (* we can truncate, but we don't want to promote *)
```
Gbury, line 4695: why ?

Gbury (changes requested):

Gbury:
Also: there's a CI failure that seems related to this PR: `Fatal error: exception Stdlib__Domain.Encapsulated("Invalid_argument(\"Random.int32_in_range\")")` on `chi2.ml` in the upstream testsuite

backend/cmm_helpers.ml
```diff
@@ -4426,3 +4426,263 @@ let reperform ~dbg ~eff ~cont ~last_fiber =
...
+      match (Obj.magic (bit_width_and_signedness land 1) : Signedness.t) with
+      | (Signed | Unsigned) as signedness ->
+        (* If [Signedness.t] ever changes, adjust the representation of [t]
+           accordingly *)
+        signedness
+
+    (** This type annotation proves that [int_of_signedness] is valid *)
+    type signedness_is_immediate = Signedness.t [@@immediate]
+
+    external int_of_signedness : signedness_is_immediate -> int = "%identity"
```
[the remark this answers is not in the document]
mshinwell: Yeah I don't think we should be doing this, or using Obj.magic, in this code.

backend/cmm_helpers.ml
```diff
@@ -4436,3 +4457,263 @@ let reperform ~dbg ~eff ~cont ~last_fiber =
...
+
+  module Integral = struct
+    type t =
+      | Untagged of Integer.t
+      | Tagged of Tagged_integer.t
+
+    let nativeint = Untagged Integer.nativeint
+
+    let[@inline] untagged = function
+      | Untagged t -> t
```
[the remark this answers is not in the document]
jvanburen: Yes, this is so that you can cast to the untagged representation of a type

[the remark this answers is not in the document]
jvanburen: it's not for asserting that we know a type is untagged

middle_end/flambda2/to_cmm/to_cmm_primitive.ml
```diff
@@ -628,124 +619,127 @@ let phys_equal _env dbg op x y =
...
-    (* Operations on 32-bit integer arguments must return something in the range
-       of 32-bit integers, hence the sign_extensions here. The [C.low_32]
-       operations (see above in [sign_extend_32_can_delay_overflow]) are used to
-       avoid unnecessary sign extensions, e.g. when chaining additions together.
-       Also see comment below about [C.low_32] in the [Div] and [Mod] cases. *)
-    let sign_extend_32_can_delay_overflow f =
-      C.sign_extend ~bits:32 ~dbg
-        (f (C.low_bits ~bits:32 ~dbg x) (C.low_bits ~bits:32 ~dbg y) dbg)
+  let kind = integral_of_standard_int kind in
+  let[@local] wrap f =
```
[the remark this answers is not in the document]
jvanburen: I tried this and I don't think it was much cleaner. If you feel strongly about this could you give an example?

middle_end/flambda2/to_cmm/to_cmm_primitive.ml
```diff
@@ -522,34 +522,67 @@ let dead_slots_msg dbg function_slots value_slots =
 
 (* Arithmetic primitives *)
 
+let integral_of_standard_int : K.Standard_int.t -> C.Scalar_type.Integral.t =
+  let[@inline] untagged_int bit_width : C.Scalar_type.Integral.t =
+    Untagged (C.Scalar_type.Integer.create_exn ~bit_width ~signedness:Signed)
```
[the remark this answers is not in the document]
jvanburen: I disagree that we should create duplicate values per-type, that is a big part of what makes adding more types awful. nativeint is exposed because it gives some information about the target system, just like arch_bits. The allocations *should* already be lifted out by flambda2 since the constructors are all inlined.

middle_end/flambda2/to_cmm/to_cmm_primitive.ml
```diff
@@ -522,34 +522,67 @@ let dead_slots_msg dbg function_slots value_slots =
...
+  | Naked_nativeint -> Untagged C.Scalar_type.Integer.nativeint
+  | Naked_immediate ->
+    Untagged C.Scalar_type.Tagged_integer.(untagged immediate)
+  | Tagged_immediate -> Tagged C.Scalar_type.Tagged_integer.immediate
+
+let numeric_of_standard_int_or_float :
+    K.Standard_int_or_float.t -> C.Scalar_type.t =
+  let[@inline] untagged_int bit_width : C.Scalar_type.t =
+    Integral
+      (Untagged (C.Scalar_type.Integer.create_exn ~bit_width ~signedness:Signed))
```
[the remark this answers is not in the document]
jvanburen: same response as above

backend/cmm_helpers.ml
```diff
@@ -4436,3 +4457,263 @@ let reperform ~dbg ~eff ~cont ~last_fiber =
...
+      | Tagged x, Tagged y -> Tagged_integer.equal x y
+      | Tagged _, _ -> false
+
+    let print ppf t =
+      match t with
+      | Untagged untagged -> Integer.print ppf untagged
+      | Tagged tagged -> Tagged_integer.print ppf tagged
+
+    let[@inline] is_promotable ~src ~dst =
+      Integer.is_promotable ~src:(untagged src) ~dst:(untagged dst)
```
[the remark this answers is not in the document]
jvanburen: is_promotable is bad naming here. I renamed it to `can_cast_without_losing_information` which more accurately describes what we're computing here. That should also make it more obvious that it's correct.

backend/cmm_helpers.ml
jvanburen: These operations are just not implemented... (and in fact there's only one: float -> unsigned int64).
It just wasn't implemented before, and it remains unimplemented now, since that seemed unrelated to this PR.

I added a comment explaining the semantics to the `Scalar_type` module.

middle_end/flambda2/to_cmm/to_cmm_primitive.ml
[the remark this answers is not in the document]
Gbury: I don't understand what you mean by "create duplicate values per-type". I'm only advocating for creating some specific values **once** since we know that some value will be used very regularly (also, I think it helps with the code being more readable).

As for what makes adding more types awful, I think a big part of it is solved by this very PR: before it, the `to_cmm` code actually had to basically do a match on every numeric kind, or pairs of such kinds, or add conversions when the flambda2 operation kind didn't match the existing helper functions from cmm_helpers, and it is indeed tedious. With this PR, a lot of that work is transparently done via the `conjugate` function. Since this function operates on scalar_types that can be created with any bit-width, I don't see how exposing some commonly-used scalar types **in addition** to the generic scalar type creation functions will make adding more types awful ?

backend/cmm_helpers.ml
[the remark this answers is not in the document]
Gbury: From what I read, the only use of that function is in `cast_cast_without_losing_information` , where the actual information that is of interest is the number of bits actually used to store information, so I think it should be better to rearrange the code a bit to not need this function (and instead use a function that returns the actually use-for-information bit-width for instance).

backend/cmm_helpers.mli
```diff
@@ -1226,3 +1226,149 @@ val set_field_unboxed :
...
+
+    val signedness : t -> Signedness.t
+
+    val with_signedness : t -> signedness:Signedness.t -> t
+
+    val signed : t -> t
+
+    val unsigned : t -> t
+
+    val can_cast_without_losing_information : src:t -> dst:t -> bool
```
Gbury, line 1298: Some more documentation in general for these modules would be good, but at the very least for this function, you should document that even if it returns `[true]` it does not necessarily mean that the cast (that does not lose information) is the identity (since there may be the need to tag/untag).

backend/cmm_helpers.mli
```diff
@@ -1226,3 +1226,149 @@ val set_field_unboxed :
...
+      sign- or zero-extends the input value according to the sign of the result.
+
+      A static_cast from an integral type to a float is pretty self-explanatory.
+
+      A static_cast from a float to an integral type always rounds toward zero. If the
+      resulting integral does not fit in the destination type, the result is unspecified
+      (although it's generally zero).
+
+      Casting floats to/from unsigned register-width integers is not implemented and will
+      raise in the compiler. *)
```
Gbury, line 1245: This doc comment should be attached to that `static_cast` type declaration below, rather than the module itself, which would instead benefit from a high-level explanation of what is the purpose of this module (and its sub-modules).

Gbury (commented):
Should I wait until you fix the CI failures to review again ?

backend/cmm_helpers.ml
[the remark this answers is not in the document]
jvanburen: https://github.com/ocaml-flambda/flambda-backend/blob/67434f6c163cdbf77d94617c1c332cf9e957ab7c/middle_end/flambda2/to_cmm/to_cmm_primitive.ml#L721 

it's useful b/c you can static_cast or conjugate to untagged kind , and it's just a no-op if the integer is already untagged.

backend/cmm_helpers.ml
jvanburen: I clarified the comments

middle_end/flambda2/to_cmm/to_cmm_primitive.ml
```diff
@@ -557,193 +595,160 @@ let unary_float_arith_primitive _env dbg width op arg =
...
-       operations (see above in [sign_extend_32_can_delay_overflow]) are used to
-       avoid unnecessary sign extensions, e.g. when chaining additions together.
-       Also see comment below about [C.low_32] in the [Div] and [Mod] cases. *)
-    let sign_extend_32_can_delay_overflow f =
-      C.sign_extend ~bits:32 ~dbg
-        (f (C.low_bits ~bits:32 ~dbg x) (C.low_bits ~bits:32 ~dbg y) dbg)
+  let kind = integral_of_standard_int kind in
+  let[@local] wrap f =
+    (* We cast the operands to the width that the operator expects, apply the
+       operator, and cast the result back. *)
```
Gbury, line 656: The code below is correct, but every time I read it, it takes a long time to understand that calls to `wrap` will always be called with a `f` that operate on the `operator_kind` defined below, and the comment is not helping much. So either the comment should clearly clarify this, or **much preferably** `wrap` should take the `operator_kind` as argument alongside `f`.

Gbury (changes requested):
Good to go apart from one stylistic comment.

Gbury (approved):

PR #3436, opened 2025-01-08
`portable` lazy allows `nonportable` thunk

riaqn:
This PR continues #2967 and further relaxes the mode checking of `lazy`:

At present, thunk is only evaluated when `lazy` is forced which is required to be `uncontended`, which means the thunk is also at that time `uncontended`. Therefore, the thunk is fine to be `nonportable` upon construction.

### DIFF
--- a/jane/doc/extensions/modes/reference.md
+++ b/jane/doc/extensions/modes/reference.md
@@ -1,12 +1,34 @@
+The goal of this document is to be a reasonably complete reference to the mode system in
+OCaml.
+
+<!-- CR zqian: For a gentler introduction, see [the introduction](intro.md). -->
+
+The mode system in the compiler tracks various properties of values, so that certain
+performance-enhancing operations can be performed safely. For example:
+- Locality tracks escaping. See [the local allocations reference](../local/reference.md)
+- Uniqueness and linearity tracks aliasing. See [the uniqueness reference](../uniqueness/reference.md)
+- Portability and contention tracks inter-thread sharing.
+    <!-- CR zqian: reference for portability and contention -->
+
 # Lazy
 `lazy e` contains a thunk that evaluates `e`, as well as a mutable cell to store the
-result of `e`. Upon construction, `lazy e` follows the mode of `e` and the comonadic mode
-of the thunk. Upon destruction (forcing a lazy value), the result follows the mode of lazy
-value. Currently, the thunk, the result and the `lazy` are all heap-allocated.
+result of `e`. Upon construction, the mode of `lazy e` is weaker than `e`. For example, if
+`e` is `nonportable`, then `lazy e` cannot be `portable`. Upon destruction (forcing a lazy
+value), the result is weaker than the mode of lazy value. For example, forcing a
+`nonportable` lazy value cannot give a `portable` result. Additionally, forcing a lazy
+value involves accessing the mutable cell and thus requires the lazy value to be
+`uncontended`.
 
-Additionally, forcing a lazy value involves accessing the mutable cell and thus requires
-the lazy value to be `uncontended`. Therefore, one can construct a lazy value at
-`portable` even if the thunk is `nonportable` (e.g., closing over `uncontended` or
-`nonportable` values). Similarly, the thunk is guaranteed to run at most once even if the
-lazy value is forced multiple times. Therefore, one can construct the lazy value at `many`
-even if the thunk is `once` (e.g., closing over `unique` or `once` values).
+Currently, the above rules don't apply to the locality axis, because both the result and
+the lazy value are heap-allocated, so they are always `global`.
+
+Additionally, upon construction, the comonadic fragment of `lazy e` is weaker than the
+thunk. The thunk is checked as `fun () -> e`, potentially closing over variables, which
+weakens its comonadic fragment. This rule doesn't apply to several axes:
+- The thunk is always heap-allocated so always `global`.
+- Since the thunk is only evaluated if the lazy value is `uncontended`, one can construct
+a lazy value at `portable` even if the thunk is `nonportable` (e.g., closing over
+`uncontended` or `nonportable` values).
+- Since the thunk runs at most once even if the lazy value is forced multiple times, one
+can construct the lazy value at `many` even if the thunk is `once` (e.g., closing over
+`unique` or `once` values).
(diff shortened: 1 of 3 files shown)


lpw25 (approved):
LGTM

jane/doc/extensions/modes/reference.md
```diff
@@ -0,0 +1,12 @@
+# Lazy
```
goldfirere, line 1:
```suggestion
This document is a reference of how the modes interact with various other features of OCaml.

# Lazy
```

I just think it's nice to have a little intro.

jane/doc/extensions/modes/reference.md
```diff
@@ -0,0 +1,12 @@
...
+`lazy e` contains a thunk that evaluates `e`, as well as a mutable cell to store the
+result of `e`. Upon construction, `lazy e` follows the mode of `e` and the comonadic mode
```
goldfirere, line 3: I'm not sure what "follows" means here. Do you mean that `lazy e`'s monadic modes are that of `e` and comonadic ones are those of the thunk? But what modes does the thunk have? I think that, in order to be useful, this documentation needs an example -- I don't understand it as written.

jane/doc/extensions/modes/reference.md
```diff
@@ -0,0 +1,12 @@
...
+of the thunk. Upon destruction (forcing a lazy value), the result follows the mode of lazy
+value. Currently, the thunk, the result and the `lazy` are all heap-allocated.
+
+Additionally, forcing a lazy value involves accessing the mutable cell and thus requires
```
goldfirere, line 7: This paragraph I understand. But it would probably also be helped by an example.

goldfirere (commented):
I see that @lpw25 has already approved here, but we're going to have to write this documentation soon (as part of the OxCaml release), and now seems like a great time to get it right.

testsuite/tests/typing-modes/lazy.ml
```diff
@@ -43,8 +43,7 @@ let foo (x @ local) =
 val foo : local_ 'a lazy_t -> 'a = <fun>
 |}]
 
-(* one can construct portable lazy, if both the thunk and the result are
-   portable *)
+(* one can construct [portable] lazy if the result is [portable] *)
```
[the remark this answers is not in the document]
riaqn: It's really saying "one can construct [portable] lazy ONLY if the result is [portable]". Added "only".

jane/doc/extensions/modes/reference.md
```diff
@@ -0,0 +1,42 @@
...
+- Locality tracks escaping. See [the local allocations reference](../local/reference.md)
+- Uniqueness and linearity tracks aliasing. See [the uniqueness reference](../uniqueness/reference.md)
+- Portability and contention tracks inter-thread sharing.
+    <!-- CR zqian: reference for portability and contention -->
+
+# Lazy
+`lazy e` contains a thunk that evaluates `e`, as well as a mutable cell to store the
+result of `e`. Upon construction, the mode of `lazy e` is weaker than `e`. For example, if
+`e` is `nonportable`, then `lazy e` cannot be `portable`. Upon destruction (forcing a lazy
+value), the result is weaker than the mode of lazy value. For example, forcing a
```
goldfirere, line 17:
```suggestion
value), the result is no stronger than the mode of lazy value. For example, forcing a
```

jane/doc/extensions/modes/reference.md
```diff
@@ -0,0 +1,42 @@
...
+`nonportable` lazy value cannot give a `portable` result. Additionally, forcing a lazy
+value involves accessing the mutable cell and thus requires the lazy value to be
+`uncontended`.
+
+Currently, the above rules don't apply to the locality axis, because both the result and
```
goldfirere, line 22: This suggests, to me, that `force` can take a local value and return a global one. Is that true?

jane/doc/extensions/modes/reference.md
```diff
@@ -0,0 +1,42 @@
...
+the lazy value are heap-allocated, so they are always `global`.
+
+Additionally, upon construction, the comonadic fragment of `lazy e` is weaker than the
```
goldfirere, line 25:
```suggestion
Additionally, upon construction, the comonadic modes of `lazy e` are no stronger than the
```

jane/doc/extensions/modes/reference.md
```diff
@@ -0,0 +1,42 @@
...
+thunk. The thunk is checked as `fun () -> e`, potentially closing over variables, which
+weakens its comonadic fragment. This rule doesn't apply to several axes:
```
goldfirere, line 27:
```suggestion
may affect its comonadic modes. This rule doesn't apply to several axes:
```

goldfirere, line 27: It seems that this rule applies to exactly 0 of the comonadic axes. Maybe just remove and state more directly? That is, the expression `lazy e` is `global`, `portable`, `many`, and inherits its uniqueness and its contention from the `e` (if that's true).

riaqn: We just merged #3283 , and very soon #3098. Maybe they will interact with lazy in a special way as well, but it's certainly safe to have the current behavior.

jane/doc/extensions/modes/reference.md
riaqn: yes that's true.

goldfirere: But that's not true in the stdlib. Should it be? This sentence suggests that it is.

riaqn: Hmm my impression is that `stdlib` doesn't contain any mode stuff? If we want we should put a better `force` in `stdlib_alpha`?
(But then, I think many stuff is missing in stdlib_alpha)

jane/doc/extensions/modes/reference.md
riaqn: Hmm, "weaken" is more precise than "affect", right?

jane/doc/extensions/modes/reference.md
goldfirere: I think there is improvement to be found here, but we don't need to fix it now. Resolving.

jane/doc/extensions/modes/reference.md
goldfirere: Sure. But it looks like you've made this change.

PR #3437, opened 2025-01-08
Stub implementation of new custom memory API

stedolan:
This is a stub implementation of @damiendoligez's new custom memory allocation API. The new API tracks both allocation and deallocation of dependent memory using `caml_alloc_dependent_mem / caml_free_dependent_mem`, and provides a convenience function `caml_alloc_custom_dep` which combines a custom block allocation and a call to `caml_alloc_dependent_mem`.

The difference from `caml_alloc_custom_mem` is that a user of `caml_alloc_custom_dep` must be sure to call `caml_free_depedent_mem` when the custom block is finalised.

In this PR, the new API is added so that we can port code to it, but under the hood it just calls `caml_alloc_custom_mem` and the `alloc_dependent` / `free_dependent` functions are no-ops.

(The changes to the existing API functions `caml_alloc/free_dependent_mem` are not backwards compatible. However, I'm not aware of any actual users of these functions, and if we find one someday it's easy to adapt)

### DIFF
--- a/runtime/bigarray.c
+++ b/runtime/bigarray.c
@@ -251,7 +251,7 @@ caml_ba_alloc(int flags, int num_dims, void * data, intnat * dim)
   uses_resources =
     ((flags & CAML_BA_MANAGED_MASK) == CAML_BA_MANAGED)
     && !(flags & CAML_BA_SUBARRAY);
-  res = caml_alloc_custom_mem(&caml_ba_ops, asize, uses_resources ? size : 0);
+  res = caml_alloc_custom_dep(&caml_ba_ops, asize, uses_resources ? size : 0);
   b = Caml_ba_array_val(res);
   b->data = data;
   b->num_dims = num_dims;
@@ -291,9 +291,11 @@ CAMLexport void caml_ba_finalize(value v)
   case CAML_BA_MANAGED:
     if (b->proxy == NULL) {
       free(b->data);
+      caml_free_dependent_memory(v, caml_ba_byte_size(b));
     } else {
       if (caml_atomic_refcount_decr(&b->proxy->refcount) == 1) {
         free(b->proxy->data);
+        caml_free_dependent_memory(v, b->proxy->size);
         free(b->proxy);
       }
     }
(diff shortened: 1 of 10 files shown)


runtime4/caml/custom.h
```diff
@@ -65,6 +65,17 @@ CAMLextern value caml_alloc_custom_mem(struct custom_operations * ops,
...
 
+  /* [caml_alloc_custom_dep] allocates a custom block with dependent memory
+   (memory outside the heap that will be reclaimed when the block is
+   finalized). If [mem] is greater than [custom_minor_max_size] (see gc.mli)
+   the block is allocated directly in the major heap.
+   The program must call [caml_free_dependent_memory] when the memory is
+   reclaimed.
+  */
+CAMLextern value caml_alloc_custom_dep(struct custom_operations * ops,
+                                       uintnat size, /*size in bytes*/
```
NickBarnes, line 76: I'd like a clearer distinction in the comments between `size` and `mem`.

runtime/bigarray.c
```diff
@@ -251,7 +251,7 @@ caml_ba_alloc(int flags, int num_dims, void * data, intnat * dim)
   uses_resources =
     ((flags & CAML_BA_MANAGED_MASK) == CAML_BA_MANAGED)
     && !(flags & CAML_BA_SUBARRAY);
-  res = caml_alloc_custom_mem(&caml_ba_ops, asize, uses_resources ? size : 0);
+  res = caml_alloc_custom_dep(&caml_ba_ops, asize, uses_resources ? size : 0);
```
NickBarnes, line 254: Unclear to me why this is different in runtime4.

runtime/custom.c
```diff
@@ -130,6 +130,12 @@ CAMLexport value caml_alloc_custom_mem(const struct custom_operations * ops,
   return v;
 }
 
+CAMLexport value caml_alloc_custom_dep(const struct custom_operations * ops,
+                                       uintnat size, mlsize_t mem)
+{
+  return caml_alloc_custom_mem(ops, size, mem);
```
NickBarnes, line 136: Add a comment indicating what else will go in here in future.

NickBarnes (approved):
Broadly this is OK, and we should probably merge it as-is. Three minor points which we could address when we put the implementation in:
- I'm unclear why `bigarray.c` is so different between runtime 4 and runtime 5;
- I would find this stuff easier to navigate with a clearer distinction in variable names and comments between custom block sizes and dependent memory sizes. Maybe consistently using `dep_size`?
- Moar comments plzkthxbai.

PR #3440, opened 2025-01-08
More capsule API updates

TheNumbat:
- `with_password` doesn't need to take a portable `f`, as `f` is not run inside the fresh capsule.
- Deleted `Protected` in favor of the upcoming key API
- `Name.packed`, `Access.packed`, and `Password.packed` are now `[@@unboxed]`, which makes `with_password` zero-alloc. This currently requires the corresponding `t`s to be `value mod external_`, because otherwise the compiler doesn't know `t` does not contain floats. We may not want passwords to be `external_`  in the future, so this should be replaced with a non-float kind.
- Adds local versions of all `Capsule.Data` functions for working with local data. Most of the global functions can't be directly implemented using the local versions (and are ~1 line), so the implementations are duplicated to add `exclave_`.
- Adds local `access`/`access_shared` and implements the global versions with them. They are also able return a portable value, but this wasn't exposed.
- `Password.Shared.t` gets the same mode crossing behavior as `Password.t`
- Fixes some comment formatting

In total, this brings us closer to a zero-global-alloc scheduler--we're just missing the runtime changes for local fork/join.


otherlibs/stdlib_alpha/capsule.mli
```diff
@@ -427,21 +569,31 @@ exception Protected : 'k Mutex.t * (exn, 'k) Data.t -> exn
...
-    If [f] raises any other exception, [protect] re-raises it as [Protected]. *)
+    If [f] raises any other exception or returns normally, [protect] merges
+    the capsule into the caller's capsule. *)
+
+val protect_portable : (Password.packed @ local -> 'a) @ local portable -> 'a @@ portable
+(** [protect_portable f] runs [f password] in a fresh capsule represented by [password].
+    If [f] returns normally, [protect_portable] merges the capsule into the caller's capsule.
+    If [f] raises an [Encapsulated] exception in the capsule represented by [password],
+    [protect_portable] unwraps the exception and re-raises it as [Protected].
+    If [f] raises any other exception, [protect_portable] re-raises it as [Protected]. *)
```
tdelvecchio-jsc, line 584: We should probably document the practical different between `protect` and `protect_portable`; just from the types, `protect` looks strictly better, and it's hard to understand the implications of the difference in the exception behavior.

otherlibs/stdlib_alpha/capsule.mli
```diff
@@ -427,21 +569,31 @@ exception Protected : 'k Mutex.t * (exn, 'k) Data.t -> exn
...
 
-val with_password : (Password.packed @ local -> 'a) @ local portable -> 'a @@ portable
+val with_password : (Password.packed @ local -> 'a) @ local -> 'a @@ portable
 (** [with_password f] runs [f password] in a fresh capsule represented by [password].
     If [f] returns normally, [with_password] merges the capsule into the caller's capsule.
     If [f] raises an [Encapsulated] exception in the capsule represented by [password],
     [with_password] unwraps the exception and re-raises it directly. *)
```
tdelvecchio-jsc, line 590: Same here; a higher-level description of the difference between `protect` and `with_password` would be useful.

tdelvecchio-jsc (changes requested):

otherlibs/stdlib_alpha/capsule.mli
TheNumbat: We've concluded that we can just delete `protect` since the key API will be ready soon

tdelvecchio-jsc (approved):

PR #3442, opened 2025-01-09
Add function [DLL.for_all_i]

gretay-js:
I would like to add a slightly non-standard but convenient function `Doubly_link_list.for_all_i`.

### DIFF
--- a/utils/doubly_linked_list.ml
+++ b/utils/doubly_linked_list.ml
@@ -333,20 +333,20 @@ let exists t ~f =
   aux t f t.first
 
 let for_all t ~f =
-  let rec aux t f curr =
+  let rec aux f curr =
     match curr with
     | Empty -> true
-    | Node node -> if f node.value then aux t f node.next else false
+    | Node node -> if f node.value then aux f node.next else false
   in
-  aux t f t.first
+  aux f t.first
 
 let for_alli t ~f =
-  let rec aux t f i curr =
+  let rec aux f i curr =
     match curr with
     | Empty -> true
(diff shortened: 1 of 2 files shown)


utils/doubly_linked_list.ml
```diff
@@ -340,6 +340,14 @@ let for_all t ~f =
   in
   aux t f t.first
 
+let for_all_i t ~f =
+  let rec aux t f i curr =
```
xclerc, line 344: `t` does not seem to be used by `aux`
(except for the recursive call).

utils/doubly_linked_list.mli
```diff
@@ -78,6 +78,8 @@ val exists : 'a t -> f:('a -> bool) -> bool
 
 val for_all : 'a t -> f:('a -> bool) -> bool
 
+val for_all_i : 'a t -> f:(int -> 'a -> bool) -> bool
```
xclerc, line 81: For consistency with _e.g._ `mapi` [1],
the name should maybe be `for_alli`?

[1] https://github.com/ocaml-flambda/flambda-backend/blob/main/stdlib/list.mli#L186

utils/doubly_linked_list.ml
gretay-js: Removed here and in `for_all`.

utils/doubly_linked_list.mli
gretay-js: Renamed.

xclerc (approved):

PR #3448, opened 2025-01-09
Do not allow naked pointers (remove configure option)

gretay-js:
This PR proposes to remove the configure option of enabling and disabling naked pointers. It only matters for runtime4, and inconveniently the default currently is to allow naked pointers. This PR hardcodes `naked_pointers` to be disallowed in runtime4, to match runtime5 and upstream. With this PR, `flambda-backend` can no longer be used to compiler code that has naked pointers.


mshinwell (approved):

mshinwell:
For the record: we think the runtime4 GC is no longer correct in the presence of unboxed numbers in closures, except when in NNP mode.  (@stedolan is confirming this hypothesis.)  This was the cause of the segfaults seen on #3317 .

stedolan:
> For the record: we think the runtime4 GC is no longer correct in the presence of unboxed numbers in closures, except when in NNP mode. (@stedolan is confirming this hypothesis.) This was the cause of the segfaults seen on #3317 .

For the same record, here's a minimal example that segfaults on runtime4 with naked pointers enabled:
```
open Stdlib_upstream_compatible

let[@inline never] mk_closure (i : Int64_u.t) =
  Sys.opaque_identity (fun () -> i)

let () =
  let major_block = Array.make 1000 1 in
  let addr = Int64_u.(mul (of_int 2) (of_int (Obj.magic major_block))) in
  let clos = mk_closure (Int64_u.(add addr (of_int 4))) in
  Gc.full_major ();
  let _ = clos () in
  ()
```

The problem is [this check in runtime4's major GC](https://github.com/ocaml-flambda/flambda-backend/blob/main/runtime4/major_gc.c#L228-L242) - due to unboxed numbers, skipping the unscannable prefix per Closinfo is required even in naked-pointers mode. (I agree it is simpler to kill naked-pointers mode entirely than to try to fix this: naked-pointer bugs are extremely tricky to track down and there's no particular reason to believe this is the only one)

PR #3451, opened 2025-01-09
Vectorizer: propagate alignment of memory accesses

gretay-js:
Explicitly propagate alignment of memory operations to `simd_selection`.

This is just a small refactoring to help us think about alignment when we vectorizer new operations.
Currently, the alignment is always 8 but having this argument will help us consider alignment for new vector sequences and hopefully avoid bugs.

### DIFF
--- a/backend/amd64/simd_selection.ml
+++ b/backend/amd64/simd_selection.ml
@@ -485,7 +485,8 @@ let vectorize_operation (width_type : Vectorize_utils.Width_in_bits.t)
     match alignment_in_bytes with
     | None -> Misc.fatal_error "Unexpected memory operation"
     | Some alignment_in_bytes ->
-      Int.compare alignment_in_bytes vector_width_in_bytes >= 0
+      alignment_in_bytes mod vector_width_in_bytes = 0
+      && alignment_in_bytes / vector_width_in_bytes > 1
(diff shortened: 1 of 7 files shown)


backend/amd64/simd_selection.ml
```diff
@@ -474,12 +474,24 @@ let vector_width_in_bits = 128
...
   (* Assumes cfg_ops are isomorphic *)
   let width_in_bits = Vectorize_utils.Width_in_bits.to_int width_type in
   let length = List.length cfg_ops in
   assert (length * width_in_bits = vector_width_in_bits);
+  let vector_width_in_bytes = vector_width_in_bits / 8 in
+  let is_aligned_to_vector_width () =
+    match alignment_in_bytes with
+    | None -> Misc.fatal_error "Unexpected memory operation"
+    | Some alignment_in_bytes ->
+      Int.compare alignment_in_bytes vector_width_in_bytes >= 0
```
xclerc, line 488: I am a bit surprised the condition
is a mere comparison (as opposed
to something involving division or
modulo). I assume this is because
we want to only have powers of
two for these values, but wanted to
be sure that was the reason.

xclerc (approved):

PR #3452, opened 2025-01-09
Vectorize [Ifloatarithmem]

gretay-js:
On top of https://github.com/ocaml-flambda/flambda-backend/pull/3451.

There are two ways to vectorize `Ifloatarithmem, for example for `add` operation, we can either 
- two vector instructions:  vector load followed by vector add with register arguments (effectively reverting the choice made during instruction selection).
- emit one vector add instruction with a memory argument, but only when the memory is known to be 128-bit aligned. 

This PR supports both options, although the more efficient one instruction cannot currently be emitted because we don't have the alignment guarantees. This can be enabled in the future if we propagate sufficient informaton about bigarrays for example, but may need runtime changes to ensure alignment. 

To emit SIMD instructions with memory arguments, this PR adds a bunch of machinery to SIMD selection and handling to keep track of the addressing mode. In particular, it adds `ISimd_mem` to specific operations. The rest is trivial propagation of the new variant to all the right places. This is in a separate commit, which I am happy to split out into a separate PR if the design choices don't seem obvious. The new instructions are lightly tested manually (except for binary emitter), but are never emitted even when the vectorizer is enabled.

### DIFF
--- a/backend/amd64/simd_selection.ml
+++ b/backend/amd64/simd_selection.ml
@@ -857,7 +857,6 @@ let vectorize_operation (width_type : Vectorize_utils.Width_in_bits.t)
       | W32 -> None (* See previous comment *)
       | W16 -> None
       | W8 -> None)
-    | Ifloatarithmem _ -> None
     | Istore_int (_n, addressing_mode, is_assignment) -> (
       if not (Vectorize_utils.Width_in_bits.equal width_type W64)
       then None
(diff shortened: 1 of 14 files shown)


backend/amd64/simd_selection.ml
```diff
@@ -843,9 +856,77 @@ let vectorize_operation (width_type : Vectorize_utils.Width_in_bits.t)
...
+          [ { operation =
+                Operation.Specific (Isimd_mem (sse_op, addressing_mode));
+              arguments = Array.append results address_args;
+              results
+            } ]
+      else
+        (* Emit a load followed by an arithmetic operation, effectively
+           reverting the decision from Arch.selection. It will probably not be
+           beneficial with 128-bit accesses. *)
+        let sse_op : Simd.operation =
```
xclerc, line 895: Looks like a duplicate of line 874.

gretay-js: I don't know how to avoid it, the return types are different.

xclerc: Ah, sorry, I was not paying attention...

xclerc (approved):

PR #3455, opened 2025-01-09
Vectorizer: xmm register can hold ocaml values

gretay-js:
This PR adds a new variant `Valx2` to `Cmm.machtype_component` to represent a vector register that holds ocaml values and needs to be scanned by the GC, as opposed to the current `Vec128` machtype_component that is not scanned. 

- The first commit mechanically propagates the new variant through the backend. 
- The interesting part of recording these registers in the frame is in a separate commit. It requires https://github.com/ocaml-flambda/flambda-backend/pull/3453 to work correctly. 
- Finally, a separate commit fixes the vectorizer to correctly set the type of vector registers it creates. 

Currently, the vectorizer rejects candidates that would result in "mixed vectors". If we want to allow it, we can replace `Valx2`  and `Vec128` with `Vec128 of mask` where the mask indicates which parts are values.

### DIFF
--- a/backend/amd64/emit.mlp
+++ b/backend/amd64/emit.mlp
@@ -375,7 +375,7 @@ let record_frame_label live dbg =
           live_offset := encode n :: encode (n + 1) :: !live_offset
       | {typ = Valx2; loc = Stack s} as reg ->
           let n = slot_offset s (stack_slot_class reg.typ)  in
-          live_offset := n :: n + 1 :: !live_offset
+          live_offset := n :: n + Arch.size_addr :: !live_offset
       | {typ = Addr} as r ->
           Misc.fatal_error ("bad GC root " ^ Reg.name r)
       | { typ = (Val | Valx2); loc = Unknown ; } as r ->
@@ -2252,7 +2252,7 @@ let emit_probe_handler_wrapper p =
         (match r.typ with
         | Val -> k::acc
         | Int | Float | Vec128 | Float32 -> acc
-        | Valx2 -> k::k+1::acc
+        | Valx2 -> k::k+Arch.size_addr::acc
         | Addr -> Misc.fatal_error ("bad GC root " ^ Reg.name r))
       | _ -> assert false)
     saved_live
--- a/backend/amd64/emit.mlp
+++ b/backend/amd64/emit.mlp
@@ -364,13 +364,23 @@ let record_frame_label live dbg =
   let live_offset = ref [] in
   Reg.Set.iter
     (function
-      | {typ = Val; loc = Reg r} ->
+      | {typ = Val; loc = Reg r} as reg ->
+          assert (Proc.gc_regs_offset reg = r);
           live_offset := ((r lsl 1) + 1) :: !live_offset
       | {typ = Val; loc = Stack s} as reg ->
           live_offset := slot_offset s (stack_slot_class reg.typ) :: !live_offset
+      | {typ = Valx2; loc = Reg r} as reg ->
+          let n = Proc.gc_regs_offset reg in
+          let encode n = ((n lsl 1) + 1) in
+          live_offset := encode n :: encode (n + 1) :: !live_offset
+      | {typ = Valx2; loc = Stack s} as reg ->
+          let n = slot_offset s (stack_slot_class reg.typ)  in
+          live_offset := n :: n + 1 :: !live_offset
       | {typ = Addr} as r ->
           Misc.fatal_error ("bad GC root " ^ Reg.name r)
-      | _ -> ()
+      | { typ = (Val | Valx2); loc = Unknown ; } as r ->
+        Misc.fatal_error ("Unknown location " ^ Reg.name r)
+      | { typ = Int | Float | Float32 | Vec128; _ } -> ()
     )
     live;
   record_frame_descr ~label:lbl ~frame_size:(frame_size())
@@ -2242,6 +2252,7 @@ let emit_probe_handler_wrapper p =
         (match r.typ with
         | Val -> k::acc
         | Int | Float | Vec128 | Float32 -> acc
+        | Valx2 -> k::k+1::acc
         | Addr -> Misc.fatal_error ("bad GC root " ^ Reg.name r))
       | _ -> assert false)
     saved_live
--- a/backend/amd64/emit.mlp
+++ b/backend/amd64/emit.mlp
@@ -811,8 +811,7 @@ let move (src : Reg.t) (dst : Reg.t) =
   begin match src.typ, src.loc, dst.typ, dst.loc with
   | Float, Reg _, Float, Reg _
   | Float32, Reg _, Float32, Reg _
-  | Valx2, _, Valx2, _
-  | Vec128, _, Vec128, _ (* Vec128 stack slots are always aligned. *) ->
+  | (Vec128 | Valx2), _, (Vec128 | Valx2), _ (* Vec128 stack slots are always aligned. *) ->
     if distinct then I.movapd (reg src) (reg dst)
   | Float, _, Float, _ ->
     if distinct then I.movsd (reg src) (reg dst)
--- a/backend/amd64/emit.mlp
+++ b/backend/amd64/emit.mlp
@@ -811,7 +811,8 @@ let move (src : Reg.t) (dst : Reg.t) =
   begin match src.typ, src.loc, dst.typ, dst.loc with
   | Float, Reg _, Float, Reg _
   | Float32, Reg _, Float32, Reg _
-  | (Vec128 | Valx2), _, (Vec128 | Valx2), _ (* Vec128 stack slots are always aligned. *) ->
+  | Valx2, _, Valx2, _
+  | Vec128, _, Vec128, _ (* Vec128 stack slots are always aligned. *) ->
     if distinct then I.movapd (reg src) (reg dst)
   | Float, _, Float, _ ->
     if distinct then I.movsd (reg src) (reg dst)
--- a/backend/amd64/emit.mlp
+++ b/backend/amd64/emit.mlp
@@ -53,7 +53,7 @@ let float_reg_name = Array.init 16 (fun i -> XMM i)
 let register_name typ r =
   match (typ : machtype_component) with
   | Int | Val | Addr -> Reg64 (int_reg_name.(r))
-  | Float | Float32 | Vec128 -> Regf (float_reg_name.(r - 100))
+  | Float | Float32 | Vec128 | Valx2 -> Regf (float_reg_name.(r - 100))
 
 let phys_rax = phys_reg Int 0
 let phys_rdx = phys_reg Int 4
@@ -292,6 +292,7 @@ let emit_Llabel fallthrough lbl section_name =
 let x86_data_type_for_stack_slot : machtype_component -> data_type = function
   | Float -> REAL8
   | Vec128 -> VEC128
+  | Valx2 -> VEC128
   | Int | Addr | Val -> QWORD
   | Float32 -> REAL4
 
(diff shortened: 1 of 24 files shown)


gretay-js:
Rebasing it on top of https://github.com/ocaml-flambda/flambda-backend/pull/3453 so the vectorizer test in the CI doesn't segfault.

mshinwell:
Could you confirm an example code sequence where this is necessary, given @stedolan 's comments that it seems like the stores would have to be via `caml_modify`?

gretay-js:
> Could you confirm an example code sequence where this is necessary, given @stedolan 's comments that it seems like the stores would have to be via `caml_modify`?

Creating a new record, for example:
```
type t1 = { mutable d0 : int64; mutable d1: int64; }
let copy_fresh (a : t1) : t1 = { d0 = a.d0; d1 = a.d1; }
let copy_mutable (a : t1) (b: t1) : unit = b.d0 <- a.d0;  b.d1 <- a.d1 
```
- `copy_mutable` has `caml_modify`s and won't be vectorized.
- `copy_fresh` is vectorized and the vector instruction contains ocaml values that may be live across an allocation (of the record itself or an  unrelated instruction that may trigger a GC).

backend/amd64/emit.mlp
```diff
@@ -364,13 +365,23 @@ let record_frame_label live dbg =
...
           live_offset := ((r lsl 1) + 1) :: !live_offset
       | {typ = Val; loc = Stack s} as reg ->
           live_offset := slot_offset s (stack_slot_class reg.typ) :: !live_offset
+      | {typ = Valx2; loc = Reg r} as reg ->
+          let n = Proc.gc_regs_offset reg in
+          let encode n = ((n lsl 1) + 1) in
+          live_offset := encode n :: encode (n + 1) :: !live_offset
+      | {typ = Valx2; loc = Stack s} as reg ->
+          let n = slot_offset s (stack_slot_class reg.typ)  in
+          live_offset := n :: n + 1 :: !live_offset
```
TheNumbat, line 379: `slot_offset` for stack slots is in bytes, so this should be `n+8`

backend/amd64/emit.mlp
```diff
@@ -801,15 +812,15 @@ let move (src : Reg.t) (dst : Reg.t) =
   begin match src.typ, src.loc, dst.typ, dst.loc with
   | Float, Reg _, Float, Reg _
   | Float32, Reg _, Float32, Reg _
-  | Vec128, _, Vec128, _ (* Vec128 stack slots are always aligned. *) ->
+  | (Vec128 | Valx2), _, (Vec128 | Valx2), _ (* Vec128 stack slots are always aligned. *) ->
```
TheNumbat, line 815: We will never generate `Vec128 <-> Valx2` movs, so these should be separate cases

backend/amd64/emit.mlp
```diff
@@ -2242,6 +2253,7 @@ let emit_probe_handler_wrapper p =
         (match r.typ with
         | Val -> k::acc
         | Int | Float | Vec128 | Float32 -> acc
+        | Valx2 -> k::k+1::acc
```
TheNumbat, line 2256: I think this is also in bytes

backend/select_utils.ml
```diff
@@ -212,13 +212,18 @@ let oper_result_type = function
    dependencies, because it uses [Arch]. *)
 let size_component : machtype_component -> int = function
   | Val | Addr -> Arch.size_addr
-  | Int -> Arch.size_int
+  | Int ->
+    assert (Int.equal Arch.size_int Arch.size_addr);
```
TheNumbat, line 216: Why add the assert here?

backend/vectorize_utils.ml
```diff
@@ -95,3 +95,37 @@ module Vectorized_instruction = struct
...
+  | [] -> assert false
+  | hd :: tl -> (
+    let can_vectorize = List.for_all (vectorizable_machtypes hd) tl in
+    if not can_vectorize
+    then
+      Misc.fatal_errorf "register pack with incompatible mach types:"
+        Printreg.reglist pack;
+    match hd.typ, List.length pack with
+    | (Int | Addr | Float | Float32), _ ->
+      (* allows subregs, width should be correct by construction of [Group]. *)
```
TheNumbat, line 125: May as well check the length is 2 (or 4 for float32) since the val case does

backend/vectorize_utils.mli
```diff
@@ -81,3 +81,13 @@ module Vectorized_instruction : sig
...
 end
+
+(** Given two registers of non-vector types, return true iff there exist a vector type
+    that can contain both of them. Currently distinguishes between [Val] and other
+    types. Mixing [Val] with non-Val in a vector is not yet supported. *)
+val vectorizable_machtypes : Reg.t -> Reg.t -> bool
+
+val vectorize_machtypes : Reg.t list -> Cmm.machtype_component
+(* CR-someday gyorsh: [vectorizable_machtypes] should take a [pack] instead of a
+   pair, to handle longer vectors, and to present a uniform interface with
```
TheNumbat, line 92: Is taking a list sufficient?

TheNumbat (commented):
I'm confident this doesn't change behavior for the existing types, but I think scanning vectorized values will need some more testing.

backend/vectorize_utils.ml
gretay-js: Done

backend/vectorize_utils.mli
gretay-js: I was referring to generalizing `vectorizable_machtypes r1 r2` to `vectorizable_machtypes reg_list`. Why would it not be enough?

gretay-js:
@TheNumbat this is ready for another round of review. The CI failure is related to #3460 and will go away once we merge the fixes.

backend/amd64/emit.mlp
gretay-js: @xclerc can you please confirm if this is possible?

backend/cmm.ml
```diff
@@ -75,6 +77,8 @@ let lub_component comp1 comp2 =
     Printf.eprintf "%d %d\n%!" (Obj.magic comp1) (Obj.magic comp2);
     (* Float unboxing code must be sure to avoid this case. *)
     assert false
+  | Valx2, _ | _, Valx2 ->
```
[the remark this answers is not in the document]
xclerc: It is handled in the same way in
many other places, so I guess we
will update all the functions at
once if the middle-end starts
using such components.

backend/amd64/emit.mlp
TheNumbat: Yeah, we do get moves between `Int/Val/Addr`, but not just because they're the same register class (I think it's for optimization purposes). All other conversions are supposed to go through `reinterpret_cast`

backend/vectorize_utils.mli
TheNumbat: Sure, I mean the cr-someday can be deleted then

backend/cmm.ml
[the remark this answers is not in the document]
TheNumbat: Maybe another cleanup task for the sprint

TheNumbat (approved):

backend/amd64/emit.mlp
gretay-js: Here is my understanding of the situation from an offline discussion with @xclerc: 
- `types_are_compatible` determines what moves are allowed in the sense of "coaleasing", that is, what virtual registers can be allocated to the same hardware register. 
- For better register allocation, we should return `true` as often as possible, while still respecting the register class and stack class constraints. Therefore, `types_are_compatible` returns true for `Valx2` and `Vec128`.
- Since we need to keep `types_are_compatible` in sync with register and stack classes, I moved `types_are_compatible` to a target specific file, as suggested by @xclerc in another comment below. 
- It is still not clear to me if the register allocator can actual generate moves between registers with different types, even if `types_are_compatible` returns true. (If there are such moves, the "wrongly typed" register cannot be live across any instruction that may trigger a GC.) To be compatible with the way we handle `Val` and `Int` case in `Emit.move` I think we should allow moves between `Valx2` and `Vec128`.

backend/arm64/emit.mlp
```diff
@@ -164,7 +164,12 @@ let record_frame_label live dbg =
           live_offset := slot_offset s (stack_slot_class reg.typ) :: !live_offset
       | {typ = Addr} as r ->
           Misc.fatal_error ("bad GC root " ^ Reg.name r)
-      | _ -> ())
+      | { typ = Valx2; } as r ->
+          (* CR mslater: (SIMD) arm64 *)
+          Misc.fatal_error ("Unexpect Valx2 type of reg " ^ Reg.name r)
```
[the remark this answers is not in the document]
gretay-js: Fixed.

backend/reg.ml
```diff
@@ -345,4 +345,8 @@ let types_are_compatible left right =
   | Float32, Float32
   | Vec128, Vec128 ->
     true
-  | (Int | Val | Addr | Float | Float32 | Vec128), _ -> false
+  | Valx2, Valx2 ->
+    true
+  | Valx2, Vec128 | Vec128, Valx2 ->
+    true
```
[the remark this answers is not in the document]
gretay-js: Move to `Proc`.

backend/vectorize_utils.mli
gretay-js: The function `vectorize_machtypes` in the preview already uses a pack. The CR is talking about `vectorizable_machtypes` on Line 88, right above the preview, not visible in the preview, and does not take a list as input. This should really be the same function, and maybe should also live in a target-specific file. I prefer to clean this up in a follow up PR, if that's okay with you, @TheNumbat ?

backend/amd64/emit.mlp
xclerc: Thanks for the summary. As you wrote, the
function is used only by IRC to determine
whether a pseudo-reg can be an alias for
another one. One thing I noted is that it
will impact only the location of the pseudo-reg,
but crucially its `typ` is not affected.

xclerc (approved):

backend/cmm.ml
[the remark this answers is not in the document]
mshinwell-js: @xclerc please add this to the list

backend/vectorize_utils.ml
```diff
@@ -95,3 +95,37 @@ module Vectorized_instruction = struct
...
+  | (Val | Int | Addr | Float | Float32), (Vec128 | Valx2) ->
+    Misc.fatal_errorf "Unexpected vector machtype Vec128 or Valx2: %a %a"
+      Printreg.reg r1 Printreg.reg r2
+  | Val, Val -> true
+  | Val, (Int | Addr | Float | Float32) | (Int | Addr | Float | Float32), Val ->
+    false
+  | (Int | Addr | Float | Float32), (Int | Addr | Float | Float32) ->
+    (* It is safe to mix Float32, Float, and Int for the purpose of GC, because
+       they are not scannable. It may not be possible to vectorize the
+       operation. *)
```
stedolan, line 111: `Addr` here worries me a bit. An `Addr` value is something like an internal pointer into the middle of a heap block - not only are they not GC'able, the GC must not be allowed to run at all while they are live (as they can be invalidated by the GC moving things). So combining, say, an `Int` and an `Addr` into a `Vec128` sounds bad, as you'd lose the fact that the GC must be inhibited while this value is live.

gretay-js: For now we can just not vectorize instructions that use registers of type `Addr`. Would it be enough to address this concern? (A better solution would be to introduce `Addrx2` but I would like to do it as a follow after we find an example where this is useful and probably refactor `machtype_component` to make it easier to have a vector of components.)

mshinwell: Yes, I would just not vectorize those.

gretay-js: Done

backend/vectorize_utils.mli
gretay-js: slightly improved the CR before resolving this comment.

gretay-js:
> I think scanning vectorized values will need some more testing.

I added a test for it [here](https://github.com/ocaml-flambda/flambda-backend/pull/3456/commits/5630ea099f521884ad63c78cbb66ad5b5efbb2f0) in #3456.
This test occasionally segfaults without [this](https://github.com/ocaml-flambda/flambda-backend/pull/3455/commits/22e9a7733f0d1fc1514a44ea17b014489a2ca086) fix for live offsets of spilled Valx2 regs.

PR #3457, opened 2025-01-09
Restructure modal and with-bounds (fka baggage) in jkind

glittershark:
This commit contains a significant restructuring of the modal upper bounds and with-bounds (which is the new name for what was previously called "baggage") in the innards of `Jkind.t`

Previously, jkinds morally consisted of:

- a layout
- a map from axis, to a pair of upper bound and a list of types (the with-bounds
  for that axis)

now, we instead have, morally:
- a layout
- a map from axis to the upper bound for that axis
- a list of (type, modality) pairs, where the modality is effectively providing us the way that the type operates on the axis

This behavior is both much closer to the syntactic representation of jkinds, making it simpler both to construct and to print, and probably more efficient, since the usual behavior is that a type appears on the with-bounds for all axes (this is only not the case in the presence of modalities)

The test changes are mostly innocuous:

1. Printing changed slightly when illegal crossing is used - actually I'm reasonably sure that this change *fixed* printing of kind annotations for types with illegal crossing
2. We're no longer deduplicating types for printing kinds (we used to do this only when constructing outcometrees); this will come back later as we do deduplication of with-bounds at construction time.
3. We actually got better at printing kinds in general, eg we're now better at figuring out built-in aliases to print under some with-bounds.

Past the actual internals of with-bounds, we frequently /reconstruct/ the list-of-types-per-axis shape, eg in subsumption - this is mostly done to avoid this commit getting too large and hairy, and in follow-up I plan on refactoring to do more algorithms such as bound extension holistically rather than per-axis, which ought to also realize more performance gains here.

note for reviewers: the last commit is a cherry-pick from main to get CI passing, and can be ignored. The two other commits can easily be reviewed separately.

### DIFF
--- a/typing/jkind.ml
+++ b/typing/jkind.ml
@@ -1131,7 +1131,7 @@ module Jkind_desc = struct
     }
 
   let unsafely_set_upper_bounds t ~from =
-    { t with upper_bounds = from.upper_bounds }
+    { t with upper_bounds = from.upper_bounds; with_bounds = No_with_bounds }
 
   let add_with_bounds ~relevant_for_nullability ~type_expr ~modality t =
     { t with
@@ -1349,10 +1349,14 @@ end
 let add_nullability_crossing t =
   { t with jkind = Jkind_desc.add_nullability_crossing t.jkind }
 
-let unsafely_set_upper_bounds ~from t =
-  { t with
-    jkind = Jkind_desc.unsafely_set_upper_bounds t.jkind ~from:from.jkind
-  }
+let unsafely_set_upper_bounds (type l r) ~(from : (l * r) t) t =
+  match from.jkind.with_bounds with
+  | With_bounds _ -> Error ()
+  | No_with_bounds ->
+    Ok
+      { t with
+        jkind = Jkind_desc.unsafely_set_upper_bounds t.jkind ~from:from.jkind
+      }
 
 let add_with_bounds ~modality ~type_expr t =
   { t with
--- a/typing/jkind.ml
+++ b/typing/jkind.ml
@@ -401,9 +401,11 @@ module With_bounds = struct
       (l1 * allowed) t =
     match bag1, bag2 with No_with_bounds, No_with_bounds -> No_with_bounds
 
-  let add ~deep_only ~modality ~type_expr (t : (allowed * 'r) t) :
-      (allowed * 'r) t =
-    let type_info = Type_info.create ~type_expr ~modality ~deep_only in
+  let add ~relevant_for_nullability ~modality ~type_expr (t : (allowed * 'r) t)
+      : (allowed * 'r) t =
+    let type_info =
+      Type_info.create ~type_expr ~modality ~relevant_for_nullability
+    in
     match t with
     | No_with_bounds -> With_bounds (type_info, [])
     | With_bounds (ty, tys) -> With_bounds (type_info, ty :: tys)
@@ -1052,8 +1054,9 @@ module Const = struct
         { layout = base.layout;
           upper_bounds = base.upper_bounds;
           with_bounds =
-            With_bounds.add ~modality ~deep_only:true ~type_expr:type_
-              base.with_bounds
+            With_bounds.add ~modality
+              ~relevant_for_nullability:Irrelevant_for_nullability
+              ~type_expr:type_ base.with_bounds
         })
     | Default | Kind_of _ -> raise ~loc:jkind.pjkind_loc Unimplemented_syntax
 
@@ -1130,10 +1133,11 @@ module Jkind_desc = struct
   let unsafely_set_upper_bounds t ~from =
     { t with upper_bounds = from.upper_bounds }
 
-  let add_with_bounds ~deep_only ~type_expr ~modality t =
+  let add_with_bounds ~relevant_for_nullability ~type_expr ~modality t =
     { t with
       with_bounds =
-        With_bounds.add ~deep_only ~type_expr ~modality t.with_bounds
+        With_bounds.add ~relevant_for_nullability ~type_expr ~modality
+          t.with_bounds
     }
 
   let max = of_const Const.max
@@ -1227,7 +1231,8 @@ module Jkind_desc = struct
         let with_bounds =
           List.fold_right
             (fun (type_expr, modality) bounds ->
-              With_bounds.add ~deep_only:false ~type_expr ~modality bounds)
+              With_bounds.add ~relevant_for_nullability:Relevant_for_nullability
+                ~type_expr ~modality bounds)
             tys_modalities Jkind_types.With_bounds.No_with_bounds
         in
         { layout; upper_bounds; with_bounds }
@@ -1353,7 +1358,11 @@ let unsafely_set_upper_bounds ~from t =
 let add_with_bounds ~modality ~type_expr t =
   { t with
     jkind =
-      Jkind_desc.add_with_bounds ~deep_only:true ~type_expr ~modality t.jkind
+      Jkind_desc.add_with_bounds
+      (* We only care about types in fields of unboxed products for the nullability of
+         the overall kind *)
+        ~relevant_for_nullability:Irrelevant_for_nullability ~type_expr
+        ~modality t.jkind
   }
 
 let has_with_bounds (type l r) (t : (l * r) t) =
--- a/typing/jkind.ml
+++ b/typing/jkind.ml
@@ -919,32 +919,27 @@ module Const = struct
       (* Plausibly, once we have non-constant, non-identity modalities, this could be
          extended to track those as well *)
 
-      include Jkind_axis.Axis_collection (struct
-        type (+'type_expr, 'd, 'a) t = modality constraint 'd = 'l * 'r
-      end)
+      type t = modality Jkind_axis.Axis_collection.t
 
-      let to_modality =
-        Fold.f
+      let to_modality : t -> _ =
+        Jkind_axis.Axis_collection.fold
           ~combine:(fun m1 m2 -> Mode.Modality.Value.Const.concat m1 ~then_:m2)
-          { f =
-              (fun (type axis) ~(axis : axis Jkind_axis.Axis.t) include_modality ->
-                match axis, include_modality with
-                | Modal axis, Constant_modality -> (
-                  let (P axis) = Mode.Const.Axis.alloc_as_value (P axis) in
-                  match axis with
-                  | Monadic ax ->
-                    Mode.Modality.Value.Const.singleton
-                      (Atom
-                         ( Monadic ax,
-                           Join_with (Mode.Value.Monadic.Const.max_axis ax) ))
-                  | Comonadic ax ->
-                    Mode.Modality.Value.Const.singleton
-                      (Atom
-                         ( Comonadic ax,
-                           Meet_with (Mode.Value.Comonadic.Const.min_axis ax) ))
-                  )
-                | _ -> Mode.Modality.Value.Const.id)
-          }
+          ~f:(fun ~(axis : Jkind_axis.Axis.packed) include_modality ->
+            match axis, include_modality with
+            | Pack (Modal axis), Constant_modality -> (
+              let (P axis) = Mode.Const.Axis.alloc_as_value (P axis) in
+              match axis with
+              | Monadic ax ->
+                Mode.Modality.Value.Const.singleton
+                  (Atom
+                     ( Monadic ax,
+                       Join_with (Mode.Value.Monadic.Const.max_axis ax) ))
+              | Comonadic ax ->
+                Mode.Modality.Value.Const.singleton
+                  (Atom
+                     ( Comonadic ax,
+                       Meet_with (Mode.Value.Comonadic.Const.min_axis ax) )))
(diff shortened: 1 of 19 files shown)


typing/jkind_types.mli
```diff
@@ -106,53 +106,61 @@ module Layout : sig
   end
 end
 
-module Baggage : sig
+module With_bounds : sig
+  module Type_info : sig
+    type +'type_expr t =
+      { type_expr : 'type_expr;
+        modality : Mode.Modality.Value.Const.t;
+        nullability : bool
```
goldfirere, line 114: This should be a custom 2-element type.

typing/jkind_types.mli
```diff
@@ -106,53 +106,61 @@ module Layout : sig
...
+      }
+
+    val create :
+      type_expr:'type_expr ->
+      modality:Mode.Modality.Value.Const.t ->
+      deep_only:bool ->
```
goldfirere, line 120: Unclear what `deep_only` means now -- document?

typing/jkind_types.mli
```diff
@@ -106,53 +106,61 @@ module Layout : sig
...
+
   type (+'type_expr, 'd) t =
-    | No_baggage : ('type_expr, 'l * 'r) t
+    | No_with_bounds : ('type_expr, 'l * 'r) t
     (* There must always be at least one type. *)
-    | Baggage :
-        'type_expr * 'type_expr list
+    | With_bounds :
+        'type_expr Type_info.t * 'type_expr Type_info.t list
         -> ('type_expr, 'l * Allowance.disallowed) t
```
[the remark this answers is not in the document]
goldfirere: The whole idea of with-bounds needs documentation... but given that (user-facing) documentation, this encoding seems straightforward enough.

typing/jkind_types.mli
```diff
@@ -106,53 +106,61 @@ module Layout : sig
...
+  include Allowance.Allow_disallow with type ('a, _, 'd) sided = ('a, 'd) t
 
-  val try_allow_l :
-    ('type_expr, 'l * 'r, 'a) t -> ('type_expr, allowed * 'r, 'a) t option
+  val as_list : ('type_expr, 'l * 'r) t -> 'type_expr Type_info.t list
 
-  val try_allow_r :
-    ('type_expr, 'l * 'r, 'a) t -> ('type_expr, 'l * allowed, 'a) t option
-end
+  val has_with_bounds : ('type_expr, 'l * 'r) t -> bool
```
goldfirere, line 135:
```suggestion
  val has_with_bounds : ('type_expr, 'l * disallowed) t -> bool
```

? If the right is allowed, then the answer is always `false`.... and so we should probably never be asking the question.

typing/jkind_types.mli
```diff
@@ -106,53 +106,61 @@ module Layout : sig
...
-  type (+'type_expr, 'd, 'a) t =
-    { modifier : 'a;
-      baggage : ('type_expr, 'd) Baggage.t
-    }
-    constraint 'd = 'l * 'r
+  include Allowance.Allow_disallow with type ('a, _, 'd) sided = ('a, 'd) t
 
-  val try_allow_l :
-    ('type_expr, 'l * 'r, 'a) t -> ('type_expr, allowed * 'r, 'a) t option
+  val as_list : ('type_expr, 'l * 'r) t -> 'type_expr Type_info.t list
```
goldfirere, line 133: Like `has_with_bounds`, does this ever need to get called on an r-kind? Maybe.... but I kinda hope not.

typing/jkind_types.mli
```diff
@@ -106,53 +106,61 @@ module Layout : sig
...
 
-  val try_allow_r :
-    ('type_expr, 'l * 'r, 'a) t -> ('type_expr, 'l * allowed, 'a) t option
-end
+  val has_with_bounds : ('type_expr, 'l * 'r) t -> bool
 
-module Bounds : sig
-  include module type of Jkind_axis.Axis_collection (Bound)
+  val types_on_axis :
```
goldfirere, line 137: Also this one. Does this ever need to be called on an r-kind?

typing/jkind_types.ml
```diff
@@ -522,20 +522,58 @@ module Layout = struct
...
+        fprintf ppf "(@[%a@ @@@@ %a])" print_type_expr type_expr
+          Mode.Modality.Value.Const.print modality
+
+    let map_type_expr f ({ type_expr; _ } as t) =
+      { t with type_expr = f type_expr }
+
+    let is_on_axis (type a) ~(axis : a Jkind_axis.Axis.t) t =
+      match axis with
+      | Nonmodal Externality -> true (* All fields matter for externality *)
+      | Nonmodal Nullability -> t.nullability
```
goldfirere, line 547: Not sure what nullability means in this context.

typing/jkind_types.ml
```diff
@@ -522,20 +522,58 @@ module Layout = struct
...
+      | Modal axis ->
+        let (P axis) = Mode.Const.Axis.alloc_as_value (P axis) in
+        not
+          (Mode.Modality.Value.Const.proj axis t.modality
+          |> Mode.Modality.is_constant)
+
+    let compose_modality t ~then_ =
+      let modality = Mode.Modality.Value.Const.compose t.modality ~then_ in
+      assert (not (Mode.Modality.Value.Const.is_id modality));
```
goldfirere, line 556: Why this assertion?

typing/jkind_types.mli
```diff
@@ -106,53 +106,61 @@ module Layout : sig
...
+    axis:'a Jkind_axis.Axis.t -> ('type_expr, 'l * 'r) t -> 'type_expr list
 
-  include Allowance.Allow_disallow with type ('a, _, 'd) sided = ('a, 'd) t
+  val compose_modality :
```
goldfirere, line 140: This strange function can be deleted after we remove illegal crossing.

typing/jkind_types.ml
```diff
@@ -732,37 +717,34 @@ module Layout_and_axes = struct
...
-    | Some upper_bounds -> Some { layout; upper_bounds }
+  let try_allow_l :
+      type l r.
+      ('type_expr, 'layout, l * r) t ->
+      ('type_expr, 'layout, Allowance.allowed * r) t option =
+   fun { layout; upper_bounds; with_bounds } ->
+    match With_bounds.try_allow_l with_bounds with
     | None -> None
+    | Some with_bounds ->
+      Some { layout; upper_bounds = Obj.magic upper_bounds; with_bounds }
```
goldfirere, line 734: Yikes! Why the `Obj.magic`? At the least, this would need a loud comment. But I think it's not necessary?

typing/jkind_types.ml
```diff
@@ -732,37 +717,34 @@ module Layout_and_axes = struct
...
-    match
-      Bounds_map_option.f
-        { f = (fun ~axis:_ bound -> Bound.try_allow_r bound) }
-        upper_bounds
-    with
-    | Some upper_bounds -> Some { layout; upper_bounds }
+  let try_allow_r { layout; upper_bounds; with_bounds } =
+    match With_bounds.try_allow_r with_bounds with
+    | Some with_bounds ->
+      Some { layout; upper_bounds = Obj.magic upper_bounds; with_bounds }
```
goldfirere, line 739: Ditto here.

typing/jkind.ml
```diff
@@ -1288,72 +1119,113 @@ module Jkind_desc = struct
...
+      *)
+      Axis.all
+      |> List.map (fun (Axis.Pack axis) : Misc.Le_result.t ->
+             let (module Axis_ops) = Axis.get axis in
+             let bound1 = Bounds.get ~axis bounds1 in
+             let bound2 = Bounds.get ~axis bounds2 in
+             match with_bounds1, with_bounds2 with
+             | No_with_bounds, No_with_bounds ->
+               Axis_ops.less_or_equal bound1 bound2
+             (* CR layouts v2.8: This should expand types on the left. *)
```
goldfirere, line 1196: Seems like it does now, no?

typing/jkind.ml
```diff
@@ -2395,31 +2268,34 @@ let sub_jkind_l ~type_equal ~jkind_of_type sub super =
...
+              let bound1' =
+                With_bounds.extend_bound ~type_equal ~jkind_of_type ~axis
+                  ~bound:bound1 sub.jkind.with_bounds
+              in
+              Axis_ops.less_or_equal bound1' bound2 |> Misc.Le_result.is_le
+            | with_bounds2 ->
+              let with_bounds1 =
+                With_bounds.types_on_axis ~axis sub.jkind.with_bounds
+              in
+              let modifiers = Axis_ops.equal bound1 bound2 in
```
goldfirere, line 2292: You can do an `le` comparison here; equality is not necessary

typing/jkind_axis.ml
```diff
@@ -167,209 +167,269 @@ module Axis = struct
...
+        portability;
+        contention;
+        externality;
+        nullability
+      } =
+    combine (f ~axis:(Pack (Modal (Comonadic Areality))) locality)
+    @@ combine (f ~axis:(Pack (Modal (Comonadic Linearity))) uniqueness)
+    @@ combine (f ~axis:(Pack (Modal (Monadic Uniqueness))) linearity)
+    @@ combine (f ~axis:(Pack (Modal (Comonadic Portability))) contention)
+    @@ combine (f ~axis:(Pack (Modal (Monadic Contention))) portability)
```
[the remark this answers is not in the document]
glittershark: 😱

typing/jkind_types.mli
glittershark: I kinda wanna just rename it to `nullability_matters` or something like that. wdyt?

typing/jkind_types.ml
glittershark: oh, yikes indeed. I had intended to remove this entirely but missed it when reviewing.

typing/jkind_types.mli
glittershark: (did that)

goldfirere:
I have reviewed the latest four commits ("Apply suggestions from code review"..."remove pointless magic"), with no further comments.

glittershark:
[fb3f76c](https://github.com/ocaml-flambda/flambda-backend/pull/3457/commits/fb3f76c44795579e40950fd7f3b89bab7f3f1a45) needs review, after the rebase.

Also, likely,

glittershark:
[4f424dc](https://github.com/ocaml-flambda/flambda-backend/pull/3457/commits/4f424dc4f1b0a760cb9f40e6e4c205b77fa0d017) needs review (by @goldfirere )

typing/jkind_types.mli
glittershark: the place where this is called does in fact need to call it, or a test fails:

typing/jkind_types.mli
glittershark: it does (sadly?), for printing

typing/jkind_types.mli
glittershark: chased the implications of making this take an `(l * disallowed)` jkind and it ends up constraining round_up a bit. I think it's simpler to just make this return an empty list in that case.

typing/jkind_types.ml
glittershark: removed

typing/jkind.ml
glittershark: indeed

typing/jkind.ml
```diff
@@ -2272,31 +2272,36 @@ let sub_jkind_l ~type_equal ~jkind_of_type sub super =
   let bounds =
     List.for_all
       (fun (Axis.Pack axis) ->
-        let (module Axis_ops) = Axis.get axis in
+        let (module Bound_ops) = Axis.get axis in
         let bound1 = Bounds.get ~axis sub.jkind.upper_bounds in
         let bound2 = Bounds.get ~axis super.jkind.upper_bounds in
+        let with_bounds1 =
+          With_bounds.types_on_axis ~axis sub.jkind.with_bounds
```
goldfirere, line 2270: This does some work, and it will be wasted effort if the second shortcut below is taken.

typing/jkind.ml
```diff
@@ -2272,31 +2272,36 @@ let sub_jkind_l ~type_equal ~jkind_of_type sub super =
...
+        in
+        (* If bound1 is min and has no with-bounds, we're good. *)
+        (Bound_ops.le bound1 Bound_ops.min && List.length with_bounds1 = 0)
```
goldfirere, line 2273: `List.compare_length_with` is a tad more efficient. (Basically any use of `List.length` is probably wrong, in my book.)

goldfirere (approved):

glittershark:
https://github.com/ocaml-flambda/flambda-backend/pull/3457/commits/a485e229d8b0ee6900ee2cc11d4ee09d8fd0eb9a needs review from @goldfirere

typing/jkind.mli
```diff
@@ -331,7 +331,7 @@ val add_nullability_crossing : 'd t -> 'd t
 
 (** Forcibly change the modal upper bounds of a [t] based on the modal upper bounds of
```
goldfirere, line 332: This comment should explain the erroring condition

goldfirere (approved):

PR #3460, opened 2025-01-10
Reinstate %makearray_dynamic

mshinwell:
This reinstates #3317 .  There were only two merge conflicts: one in `build.yml` relating to the vectorizer, which isn't relevant now; and one relating to a test that has been deleted on `main`.

Two tests that were failing with very low probability on runtime4 have been restricted to runtime5.  Extensive investigations have failed so far to pinpoint the cause, although there is some indication it might be a problem in the GC (in particular one attempt was identified to mark a block with an invalid header and whose pointer was zero mod 2^32, which seems unlikely to be a coincidence).  There aren't as many CI jobs for runtime5, but these tests have been run many times during the investigation, and have never failed on runtime5 as far as I know.


mshinwell:
I should add: there is also some evidence that the previous segfaults may not be related directly to `%makearray_dynamic` - some core files were found on the Github CI from some of the unboxed float32 array tests as well.

mshinwell:
@gretay-js is going to look at the vectorizer failure here (this isn't a "required check").

PR #3462, opened 2025-01-10
Improve coherence of modality zapping

riaqn:
Given `c = zap m` where `m` is an inferred modality and `c` a constant modality, previously we ensure that `m <= c` will always hold, even after further mutation to `m`. This is to ensure the following example passes:
```
module type S = module type of M
..constrain modes in M by using it..
module N = M : S
```

This PR adds another condition:

> Given `c0 = zap m` and `c1 = zap' m`, where `zap` and `zap'` could be different zapping functions, we ensure `c0 <= c1` always holds, even after further mutation to `m`.

to ensure the following example passes:
```
module type S = module type of M
..constrain modes in M by using it..
module type S' = module type of M
..constrain modes in M by using it..
module N = (M : S) : S'
module N' = (M : S') : S
```

### DIFF
--- a/testsuite/tests/typing-modes/val_modalities.ml
+++ b/testsuite/tests/typing-modes/val_modalities.ml
@@ -67,24 +67,23 @@ module M : sig val x : string @@ portable contended end
 
 (* Testing the defaulting behaviour.
    "module type of" triggers the defaulting logic.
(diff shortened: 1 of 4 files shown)


lpw25 (approved):
LGTM

PR #3463, opened 2025-01-13
Reset the pacing of major collection after any synchronous major GC

NickBarnes:
During a full major collection (e.g. as part of a compaction), we do a lot of GC work and no allocation work. This causes the `work_counter` and `alloc_counter` used for GC pacing calculations to get out of sync with each other. This PR adds a function, `caml_reset_major_pacing()` which forces them into sync, and calls it after any forced major collection.

### DIFF
--- a/runtime/major_gc.c
+++ b/runtime/major_gc.c
@@ -635,6 +635,30 @@ static inline intnat diffmod (uintnat x1, uintnat x2)
   return (intnat) (x1 - x2);
 }
 
+/* Reset the work and alloc counters to be equal to each other, by
+ * setting them both equal to the "larger" (in the wrapping-around
+ * sense we are using here for work_counter and alloc_counter).
+ *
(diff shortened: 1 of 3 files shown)


runtime/major_gc.c
```diff
@@ -635,6 +635,17 @@ static inline intnat diffmod (uintnat x1, uintnat x2)
...
 }
 
+/* Reset the work counter to be equal to the alloc counter. For use at
+ * times when we have disturbed the major GC from its usual pacing and
+ * tempo, e.g. at the end of a forced major collection.
+ */
+
+void caml_reset_major_pacing(void)
+{
+  work_counter = alloc_counter;
```
stedolan, line 645: This is enough to test with, but before we merge this it needs a more domain-aware implementation (it would be bad to _decrease_ alloc_counter here, which could happen if some other domain has run ahead of us)

stedolan (approved):

PR #3465, opened 2025-01-14
Fix caml_obj_with_tag

mshinwell:
The existing `caml_obj_with_tag` had the following problems:
- Incorrect handling of the unboxed part of a closure environment.  Unboxed numbers may satisfy `Is_young`!
- No handling for `Infix_tag`.

Note that this function can be called via a `globalize` operation, so it's important to get right.

### DIFF
--- a/testsuite/tests/lib-obj/obj_dup_closures.ml
+++ b/testsuite/tests/lib-obj/obj_dup_closures.ml
@@ -1,0 +1,225 @@
+(* TEST *)
+
+external int_as_pointer : _ -> int = "%int_as_pointer"
+
+module Int64_u = struct
+  external to_int64 : int64# -> (int64[@local_opt]) = "%box_int64"
+    [@@warning "-187"]
+
+  external of_int64 : (int64[@local_opt]) -> int64# = "%unbox_int64"
+    [@@warning "-187"]
+
+  let equal x y = Int64.equal (to_int64 x) (to_int64 y)
+end
+
+let[@opaque] rand_near_minor_heap () =
+  let r = ref () in
+  let i : int = Obj.magic (int_as_pointer r) in
+  let minor_heap_size_in_bytes =
+    assert (Sys.word_size = 64);
+    (Gc.get ()).minor_heap_size * Sys.word_size / 8
+  in
+  let b = minor_heap_size_in_bytes / 2 in
+  let n = (Random.int b - (b / 2)) * 2 in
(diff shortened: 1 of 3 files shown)


testsuite/tests/lib-obj/obj_dup_closures.ml
```diff
@@ -0,0 +1,173 @@
...
+  let x = rand_string () in
+  let closures :
+      (unit -> int64#) * (unit -> int64#) * (unit -> int) *
+      (unit -> unit -> int64#) * (unit -> unit -> int64#) *
+      (unit -> unit -> int) * (unit -> int64#) * (unit -> int64#) *
+      (unit -> int) * (unit -> unit -> int64#) * (unit -> unit -> int64#) *
+      (unit -> unit -> int) =
+    make_closures i64_1 i64_2 i64_3 i64_4 x
+    |> Obj.repr  (* this includes opaqueness *)
+    |> Obj.dup
```
Ekdohibs, line 163: This will not copy the closures, only the block containing them I think?

mshinwell: Whoops, fixed

testsuite/tests/lib-obj/obj_dup_closures.ml
```diff
@@ -0,0 +1,225 @@
...
+        c3_2arg,
+        rec_c1_1arg,
+        rec_c2_1arg,
+        rec_c3_1arg,
+        rec_c1_2arg,
+        rec_c2_2arg,
+        rec_c3_2arg ) =
+    make_closures i64_1 i64_2 i64_3 i64_4 x
+  in
+  let c1_1arg : unit -> int64# = Obj.obj (Obj.dup (Obj.repr c1_1arg)) in
```
Ekdohibs, line 180: Could we replace all those (and the types annotations) by:
```ocaml
let dup (x : 'a) : 'a = Obj.(obj (dup (repr x)))
```

mshinwell: Need to use `(type a)`, but yes.

NickBarnes (approved):
I haven't properly reviewed the colossal test file, but `caml_obj_with_tag` LGTM, modulo style quibbles.

mshinwell:
@Ekdohibs looked at the first versions of the test file, and after that I just added the versions with large closures, for which I checked the flambda2 IR.  So I think that has had sufficient review.

PR #3469, opened 2025-01-14
Remove remnants of caml_obj_truncate

mshinwell:
The code for `caml_obj_truncate` (used to be exposed via `Obj.truncate`, which was removed some time ago) was still present in runtime4, and was broken in the presence of unboxed numbers in closures.  This PR also updates some comments, in both runtime4 and runtime5, notably removing a constraint that `Abstract_tag` has to be an odd number.

### DIFF
--- a/runtime/caml/misc.h
+++ b/runtime/caml/misc.h
@@ -534,7 +534,7 @@ int caml_runtime_warnings_active(void);
   00 -> free words in minor heap
   01 -> fields of free list blocks in major heap
   03 -> heap chunks deallocated by heap shrinking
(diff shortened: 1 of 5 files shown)


NickBarnes (approved):

NickBarnes:
Good riddance to `caml_obj_truncate`, which nobody will miss. Why was it important that `Abstract_tag` not be mistaken for a pointer?

mshinwell:
The truncation code used to add an `Abstract_tag` header at the start of the truncated portion, I think.

PR #3471, opened 2025-01-14
Support `stack_` for primitives that allocate

riaqn:
This PR allows writing `stack_ (ref "hello")`.

See my comments below for details.

### DIFF
--- a/lambda/translprim.ml
+++ b/lambda/translprim.ml
@@ -2067,7 +2067,7 @@ let report_error ppf = function
   | Invalid_stack_primitive Allocating_on_heap ->
       fprintf ppf
         "This primitive always allocates on heap@ \
-        (was it declared with %a or %a?)"
+        (maybe it should be declared with %a or %a?)"
         Style.inline_code "[@local_opt]" Style.inline_code "local_"
 
 let () =
--- a/lambda/translprim.ml
+++ b/lambda/translprim.ml
@@ -25,6 +25,11 @@ open Translmode
 
 module String = Misc.Stdlib.String
 
+type invalid_stack_primitive =
+  | Not_primitive
+  | Not_allocating
+  | Allocating_on_heap
+
 type error =
   | Unknown_builtin_primitive of string
   | Wrong_arity_builtin_primitive of string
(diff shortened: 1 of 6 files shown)


goldfirere:
I see why we might want this, but was this change motivated by a concrete use? In addition, a return mode of poly isn’t always right — for example, `%identity` has that property, I believe, but isn’t an allocation. If we wanted to do this, I think it would be better just to list out the primitives that optionally stack-allocate. But I probably favor not doing this at all, but instead to make the stack-allocation apparent in the named `external`, like `ref_stack`. If someone wants to ensure a stack allocation, they just use `ref_stack`.

goldfirere:
OK — I found the Slack thread that provides more context (and that already suggests not to use the “poly” criterion). But I still think that we should just have a `stack_ref` instead of allowing this. Let’s continue the design debate on Slack.

riaqn:
As discussed offline, we now introduce a new attribute `[@alloc_poly]` on `external`:
```
external[@alloc_poly] ref : 'a -> ('a ref[@local_opt]) = "%makemutable"
```
which indicates that the `[@local_opt]` is in fact an allocation. Both the behavior of `stack_`, and the `register_allocation` in `type_ident` are improved based on this.

**_This change needs to be noted in the release notes:_** users might need to update their `external` to use this attribute, to ensure the existing `register_allocation` optimization continues to happen.

Updated `stdlib` to use this, so that the tests pass. I might have missed some (and will be fixed in future PRs).

Currently `regions.ml` is still failing, which I will consult @mshinwell when he gets back.

riaqn:
Per offline discussion with @goldfirere , this is now implemented differently and no longer requires additional user annotations.

lambda/translprim.ml
```diff
@@ -2026,6 +2055,20 @@ let report_error ppf = function
...
+        because this primitive does not allocate."
+  | Invalid_stack_primitive Not_primitive ->
+      fprintf ppf
+        "This cannot be marked as stack_,@ \
+        because it is either not a primitive,@ \
+        or the primitive does not allocate."
+  | Invalid_stack_primitive Allocating_on_heap ->
+      fprintf ppf
+        "This primitive always allocates on heap@ \
+        (was it declared with %a or %a?)"
```
goldfirere, line 2070:
```suggestion
        (maybe it should be declared with %a or %a?)"
```

goldfirere (approved):
Looks great. Thanks

PR #3479, opened 2025-01-15
Update "Local allocation" documentation

riaqn:
Still draft.

This PR updates the local allocation documentation by decoupling modes from allocation.


goldfirere:
Instead of making suggestions here in the PR, I just went ahead and made some edits on top of what you started here, @riaqn. Could you please review my commit? If you're happy with that, then we can merge this.

jane/doc/extensions/stack/reference.md
```diff
@@ -121,51 +126,47 @@ type-checker ensures that it does not outlive the region that it lives in,
...
   in
   ...
 ```
 
 At the point marked `??` inside `g`, both `outer` and `inner` are
-stack-allocated values and thus local. However, only `inner` is any-region
-local, having been allocated in `g`'s region. The value `outer` is instead
-outer-region local: it is stack-allocated but from a region outer than `g`'s
-own.
+local values; they will both be allocated on the stack at runtime. However,
```
riaqn, line 168: I think we want to reduce the impression that "local values goes to stack"; here we can strictly stick to locality and escaping.

jane/doc/extensions/stack/reference.md
```diff
@@ -179,10 +180,25 @@ value `outer`, while being local, was definitely not local to the region of
...
+```ocaml
+let f (local_ x) =
+  let local_ y = 3 :: x in
+  ??
+```
+
+Both `x` and `y` are local and cannot, in general, escape a region. However,
+filling `??` in with `x` (but not `y`) is allowed. This is because we know that
+the value pointed to by `x` must have existed before `f` started and thus is
+guaranteed to be allocated in a pre-existing stack frame (or on the heap).
```
riaqn, line 194: thanks - this is an important addition.

reminder to myself: I think the interaction between allocation and locality is subtle. I will try to see if can be made clearer.

jane/doc/extensions/stack/reference.md
```diff
@@ -179,10 +180,25 @@ value `outer`, while being local, was definitely not local to the region of
...
+In contrast, `y` will point to a cons cell allocated in the current stack frame;
+returning `y` from `f` would be unsafe (and is a type error).
+
 ## Inference
 
 In fact, the allocations of the examples above will be on
-stack even without the `stack_` keyword, if it is safe to do
+stack even without the `stack_` or `local_` keywords, if it is safe to do
```
riaqn, line 201: (similarly, avoid using local_ for stack allocation)

jane/doc/extensions/stack/reference.md
```diff
@@ -316,10 +332,10 @@ let f () =
 
 Here, since `g` refers to the local value `outer`, the closure `g` must itself
 be stack-allocated. (As always, this is deduced by inference, and an explicit
-`stack_` annotation on `g` is not needed).
+`stack_` or `local_` annotation on `g` is not needed.)
```
riaqn, line 335: (Similar, avoid `local_`)

riaqn (commented):
Thanks for the edits - Could you look at my comments below? If you concur, I will make the edits.

jane/doc/extensions/stack/intro.md
```diff
@@ -2,20 +2,32 @@
...
+possible. Of course, not all values can be allocated on a stack: a value that is
+used beyond the scope of its introduction must be on the heap. Accordingly,
+the compiler uses the _locality_ of a value to determine where it will be
+allocated: _local_ values go on the stack, while _global_ ones must go on the
+heap.
+
+Though type inference will infer where stack allocation is possible, it is
+often wise to annotate places where you wish to enforce locality and stack
+allocation. This can be done in two ways, by either giving variables or
+values the `local` mode or by labeling an allocation as a `stack_` allocation:
```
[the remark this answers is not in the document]
goldfirere: I’m not sure I really agree here. To write a program that uses lots of stack allocation, you will have to write `local` in a bunch of places (mainly type signatures and on parameters). But you never have to write `stack_`. If I were teaching someone how to program with stack allocation, maybe I would start with `stack_` as the first example? But then very quickly I’d switch to talk almost exclusively about `local`, mentioning that `stack_` is simply a useful fail-safe.

I think that, to really get this right, we’ll have to do a larger-scale re-imagining of this text (which we should do, and possibly soon). But for me, that re-imagining starts with a general introduction to modes and the syntax around them, and then instantiates that general structure with locality. Probably with a page meant for users who just want to get to the important bits for stack allocation without going through the general stuff first…. yet even that page will, I think, have to emphasize `local` over `stack_`.

Bottom line for me: I think any guide to stack allocation has to mention plenty about `local`.

jane/doc/extensions/stack/intro.md
```diff
@@ -37,9 +49,10 @@ although this requires code changes to use the new `caml_alloc_local` instead of
 `caml_alloc`. A few types of allocation cannot be stack-allocated, though,
 including first-class modules, classes and objects, and exceptions. The contents
 of mutable fields (inside `ref`s, `array`s and mutable record fields) also
-cannot be stack-allocated. Annotating `stack_` on expressions that are not
-allocations is meaningless and triggers type errors.
+cannot be stack-allocated.
 
+The `stack_` keyword must immediately precede an allocation; putting it
+anywhere else (like on a function call) leads to a type error.
```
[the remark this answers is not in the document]
goldfirere: Oh yes, good point.

jane/doc/extensions/stack/reference.md
goldfirere: OK. Are you suggesting to drop the point about stack allocation? I’m fine with that.

jane/doc/extensions/stack/reference.md
goldfirere: As above, I don’t think we can avoid `local_` in stack-allocation programming. I *do* think we should avoid saying things like “stuff you label as `local` gets allocated on the stack”, because the reality is considerably more subtle. But that doesn’t mean `local_` is useless!

jane/doc/extensions/stack/intro.md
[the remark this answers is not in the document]
riaqn: The very specific point I want to make in this paragraph is that: don't use `local_` to force stack allocation - it won't always work as you expected. Use `stack_` instead.

For me, the story goes as follows. We use `stack_` to force stack allocation, and to ensure that is safe, we need to track the locality of values. As a result, some `local_` annotation needs to be added. The other direction cannot be relied upon: users can't expect stack allocation by using `local_` annotation.

I think `intro.md` can get away without saying the word "mode", but just talking about "this value is local", "the locality of this value", etc. The point is to get the readers started without much pre-requisite. The `reference.md` file can refer to `modes/intro.md` and be more systematic.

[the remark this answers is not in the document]
goldfirere: Yes I agree with that general plan.

jane/doc/extensions/stack/reference.md
```diff
@@ -121,51 +126,47 @@ type-checker ensures that it does not outlive the region that it lives in,
...
+values always live in the inner-most region, since at runtime one can
+allocate only in the current stack frame. There is an outermost region,
+including an entire file. This region never ends and is where global (heap-allocated)
+values live.
 
-  - **Any-region local**: lives in an unknown region, potentially the current
-    region.
+One subtlety is that we wish to treat `local` variables from the current region
+differently than `local` variables from an enclosing region. Variables from an
+enclosing region are called *regional*.
```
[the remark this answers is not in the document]
riaqn: I still think we should avoid using a word that sounds on par with local and global, but actually cannot be written. I think it's better to base on the local vs. global that we already have, and further split `local` into two.

I did a quick search on the slack channel, and it seems that it's more common to say "in the callers' region" etc. So maybe instead of "outer-region local" and "any-region local", we say "local to the outer region" and "local to the current region"?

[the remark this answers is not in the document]
goldfirere: I still think I favor just using the word “regional” (while explicitly saying that users can’t write that). But I’m fine with your “local to the outer region” phrasing if you disagree with “regional”.

riaqn:
This is blocked by #3471 , because I really want the complete story about stack and primitives.

riaqn:
We are also blocked by the resolution of #3106

riaqn:
We are now blocked by the ocamlformat support for #3106 - but it's also fine to tell users to use the old syntax, and fix ocamlformat later.

jane/doc/extensions/stack/reference.md
```diff
@@ -64,136 +50,121 @@ including:
...
+```
 
-Regions may nest, for instance when one function calls another. Local
-allocations always occur in the innermost (most recent) region.
+The `local_` keyword doesn't force stack allocation. However, it does weaken `l`
+to local, which prevents `l` from escaping the current region, and as a result
+the compiler will optimize `n :: x` to be stack-allocated in the current
+region. However, users may wish to use `stack_` to ensure stack allocation,
+as refactoring code can make an allocation that was previously on the stack
+silenetly move to the heap.
```
goldfirere, line 119:
```suggestion
silently move to the heap.
```

jane/doc/extensions/stack/reference.md
goldfirere: Please just go ahead and drop that bit of sentence to resolve this. Thanks

jane/doc/extensions/stack/reference.md
goldfirere: I continue to think it's better with `local_` there, because the last example above uses `local_`.

goldfirere (approved):
I think we should just try to get this out the door. I don't feel strongly on any of my points, and I think this is a nice improvement over the status quo. Happy with small changes you might make to resolve discussions. Thanks!

jane/doc/extensions/stack/intro.md
```diff
@@ -0,0 +1,166 @@
...
+although this requires code changes to use the new `caml_alloc_local` instead of
+`caml_alloc`. A few types of allocation cannot be stack-allocated, though,
+including first-class modules, classes and objects, and exceptions. The contents
+of mutable fields (inside `ref`s, `array`s and mutable record fields) also
+cannot be stack-allocated.
+
+The `stack_` keyword works shallowly: it only forces the immediately succeeding allocation
+ to be on stack. Putting it before an expression that is not an allocation (such as a
+ complete function application) leads to a type error. Stack allocating closures resulted
+ from partial applications will be supported in the future.
```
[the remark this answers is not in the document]
riaqn: Ahh yes good point. (it's the latter).

jane/doc/extensions/stack/intro.md
```diff
@@ -0,0 +1,166 @@
...
+
+## Local parameters
+
+Generally, OCaml functions can do whatever they like with their
+arguments: use them, return them, capture them in closures or store
+them in globals, etc. This is a problem when trying to pass around
+stack-allocated values, since we need to guarantee they do not
+escape.
+
+The remedy is that we allow the `local_` keyword to appear on
```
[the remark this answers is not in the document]
riaqn: I believe the majority is `local_`, and only some small amount of code written by compiler devs are `@ local`. I'm surprised this hasn't raised questions in the help channel. Maybe people outside of the compiler devs don't look at those code? Then maybe it's fine, in the sense that: people outside of compiler dev will only read and write `local_`.

jane/doc/extensions/stack/reference.md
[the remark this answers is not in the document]
riaqn: Resolved. I now say "local" and "outer-local", which I believe is a good balance.

jane/doc/extensions/stack/reference.md
riaqn: Yeah but that `local_` is used strictly for escaping analysis, not to do with stack allocation (even though the description still mentions stack allocation).

jane/doc/extensions/stack/intro.md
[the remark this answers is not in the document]
riaqn: I removed the whole sentence, since if we going towards no partial application, then it's not worth mentioning that we don't support `stack_` on them.

jane/doc/extensions/stack/reference.md
riaqn: Ahh good catch. It should say "allocations in general".

PR #3481, opened 2025-01-16
Fix error in caml_get_init_stack_wsize

mshinwell:
This function currently computes bogus stack size values when huge pages are not available.  (It is unclear why the macOS CI passed after the merge of #3384 .)


mshinwell:
Ah it's because there is no runtime5 macOS build, we should fix that.

NickBarnes (approved):
LGTM

PR #3482, opened 2025-01-16
Move two macOS CI controllers to runtime5

mshinwell:
We weren't previously checking macOS + runtime5, which has concealed at least one bug recently.


xclerc (approved):
Why not also `flambda2_macos_arm64_gi`? is it known to fail?

mshinwell:
Just because it was then half runtime5 and half 4 for the macOS ones.

PR #3483, opened 2025-01-16
Add "-ocamlrunparam" linker flag

mshinwell:
This adds a new option to `ocamlopt` called `-ocamlrunparam`.  It only has an effect when linking is taking place (this is not checked, but that is consistent with e.g. `-runtime-variant`).  It allows the specification of default `OCAMLRUNPARAM` values, which is very useful in systems where environment variables may be sanitized, or otherwise are difficult to have set correctly.  Any settings that are in the `OCAMLRUNPARAM` environment variables override any settings set via this new mechanism.

I wondered whether there might be difficulties with the shared linking mode, `-output-obj`, etc. but this follows what happens e.g. for the GC roots tables so I think it's ok.

### DIFF
--- a/asmcomp/asmlink.ml
+++ b/asmcomp/asmlink.ml
@@ -349,6 +349,16 @@ let sourcefile_for_dwarf ~named_startup_file filename =
   if named_startup_file then filename
   else ".startup"
 
+let emit_ocamlrunparam ~ppf_dump =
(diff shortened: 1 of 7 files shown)


NickBarnes (approved):
This is OK. I was itching to use the excuse to fix the horrible parsing code in `startup_aux.c` but let's do that another time. Goes well with #3468

PR #3486, opened 2025-01-17
Emit atomic compare and exchange

gretay-js:
Follow up of https://github.com/ocaml-flambda/flambda-backend/pull/3477, emit instructions for `%atomic_compare_exchange` and `%atomic_exchange`  on immediates.

### DIFF
--- a/backend/cmm.ml
+++ b/backend/cmm.ml
@@ -156,7 +156,7 @@ type rec_flag = Nonrecursive | Recursive
 
 type prefetch_temporal_locality_hint = Nonlocal | Low | Moderate | High
 
-type atomic_op = Fetch_and_add | Compare_and_swap
+type atomic_op = Fetch_and_add | Compare_and_swap | Exchange | Compare_exchange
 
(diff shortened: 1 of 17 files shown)


backend/cmm.ml
```diff
@@ -156,7 +156,7 @@ type rec_flag = Nonrecursive | Recursive
 
 type prefetch_temporal_locality_hint = Nonlocal | Low | Moderate | High
 
-type atomic_op = Fetch_and_add | Compare_and_swap
+type atomic_op = Fetch_and_add | Compare_and_swap | Exchange | Compare_exchange
```
TheNumbat, line 159: Can we rename `Compare_and_swap` to `Compare_and_set` so it matches the stdlib? It's not clear what the difference is vs. `Compare_exchange`

TheNumbat (approved):

backend/cmm.ml
gretay-js: I'll make separate PR for the change.

PR #3488, opened 2025-01-20
Rename [emit.mlp] to [emit.ml] on amd64

gretay-js:
We only need `.mlp` for arm64. Adjust `backend/dune` file accordingly. Now `emit.ml` on amd64 can benefit from dune's `copy_files#` and ocamlformat.


backend/dune
```diff
@@ -15,13 +15,13 @@
...
  (targets emit.ml)
  (mode    fallback)
- (deps    amd64/emit.mlp
-          arm64/emit.mlp)
+ (deps    arm64/emit.mlp)
  (action
    (progn
      (with-stdout-to contains-input-name
-       (bash "echo %{env:ARCH=amd64}/emit.mlp"))
+       (bash "echo %{env:ARCH=arm64}/emit.mlp"))
```
xclerc, line 25: Not really important, but this line
feels a bit weird, given it is now
guarded by the condition that
`env` is `"arm64"`.

gretay-js: yes, I can just remove the use of ARCH here (kept it for the case we add another target).

xclerc: My feeling is that it should either be
without `ARCH`, or with the original
fallback. But clearly I am nitpicking,
it should not be blocking.

xclerc (approved):

PR #3489, opened 2025-01-20
Support `-open Foo` where `Foo` is parameterised

lukemaurer:
The command line

```
ocamlopt -open Foo -parameter P -c bar.ml
```

should be fine, even if `Foo` is itself parameterised by `P`: as usual, we compile `bar.ml` as if it began with `open! Foo`, and by the subset rule, `Bar` can refer to `Foo` because it takes at least the same parameters. Unfortunately, currently we process `-open` before `-parameter`, so when we go to check the implicit reference to `Foo`, we think there are no parameters, and we report an error. (Confusingly, the error suggests that the user add `-parameter P` to the command line.)

The fix is simple: move the code that processes `-parameter` earlier so that the initial environment is constructed with the parameters already available.

### DIFF
--- a/driver/compile_common.ml
+++ b/driver/compile_common.ml
@@ -31,6 +31,7 @@ type compilation_unit_or_inferred =
 let with_info ~native ~tool_name ~source_file ~output_prefix
       ~compilation_unit ~dump_ext k =
   Compmisc.init_path ();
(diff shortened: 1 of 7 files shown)


riaqn (approved):
LGTM

PR #3490, opened 2025-01-20
Additional operations for int atomics

TheNumbat:
Adds `Atomic.{add, sub, logand, logor, logxor}` primitives. They are compiled to `lock op` instructions on x86. Note they are not exchanges, so do not return the previous value.

Added basic tests to `test_atomic.ml`.
- Duplicated the test to run in a separate domain, which triggers different behavior in the runtime. Tested on r4/r5 and with/without x86 specialization.
- Duplicated the test to run with the binary emitter.


lambda/value_rec_compiler.ml
```diff
@@ -349,6 +349,11 @@ let compute_static_size lam =
     | Patomic_compare_exchange _
     | Patomic_cas _
     | Patomic_fetch_add
+    | Patomic_add
+    | Patomic_sub
+    | Patomic_land
+    | Patomic_lor
+    | Patomic_lxor
```
lthls, line 356: Note that you could have placed these primitives in the unit-returning section earlier if you wanted (It wouldn't change anything in practice since `Value_rec_check` wouldn't allow these cases).

lthls (approved):
I'm approving the middle-end changes (plus `value_rec_compiler`).

gretay-js (approved):

flambda-backend/tests/intrinsics/atomics.ml
```diff
@@ -0,0 +1,67 @@
+
+let r = Atomic.make 1
```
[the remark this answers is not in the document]
TheNumbat: It is the same - this version only exists to test the binary emitter, though, so I think the static allocation is ok

lambda/value_rec_compiler.ml
TheNumbat: Sure, moved

PR #3491, opened 2025-01-20
Rename `atomic_cas`/`Compare_and_swap`

TheNumbat:
Currently, `Patomic_cas` / `%atomic_cas` / `Compare_and_swap` refer to the version of `cmpxchg` that returns a bool. This is confusing because we also have `Compare_exchange`, which returns the previous value. This PR renames the former to `Patomic_compare_set` / `%atomic_compare_set` / `Compare_set`, respectively.


lambda/translprim.ml
```diff
@@ -901,8 +901,8 @@ let lookup_primitive loc ~poly_mode ~poly_sort pos p =
         Primitive (Patomic_exchange {immediate_or_pointer=Pointer}, 2)
     | "%atomic_compare_exchange" ->
         Primitive (Patomic_compare_exchange {immediate_or_pointer=Pointer}, 3)
-    | "%atomic_cas" ->
```
gretay-js, line 904: I do not think we should change names of builtins that exist upstream, if possible. Adding new ones seems fine. Renaming existing `Lambda` primitive constractor here is not so bad (although I'd prefer not to), we have already changed it by adding an argument.

runtime/caml/memory.h
```diff
@@ -46,7 +46,7 @@ CAMLextern void caml_free_dependent_memory (value v, mlsize_t bsz);
 CAMLextern void caml_modify (volatile value *, value);
 CAMLextern void caml_modify_local (value obj, intnat i, value val);
 CAMLextern void caml_initialize (volatile value *, value);
-CAMLextern int caml_atomic_cas_field (value, intnat, value, value);
```
gretay-js, line 49: same here, I don't think we should rename this function.

runtime/obj.c
```diff
@@ -249,10 +249,10 @@ CAMLprim value caml_obj_add_offset (value v, value offset)
   return v + (unsigned long) Int32_val (offset);
 }
 
-CAMLprim value caml_obj_compare_and_swap (value v, value f,
```
gretay-js, line 252: same as above about renaming existing primitives.

TheNumbat: reverted

runtime/caml/memory.h
TheNumbat: reverted

runtime/obj.c
TheNumbat: reverted

lambda/translprim.ml
TheNumbat: Sad but alright

PR #3492, opened 2025-01-20
Remove wrong text, as pointed out by colleagues

goldfirere:
I think @stedolan would make a good reviewer here.

### DIFF
--- a/jane/doc/proposals/unboxed-types/index.md
+++ b/jane/doc/proposals/unboxed-types/index.md
@@ -1033,19 +1033,7 @@ get from the definition of `t0` or `t2` to a tightly packed representation; use
 
 A separate takeaway here is that memory layout in our design is *associative*:
 the memory layout does not depend on how the type definitions are
-structured. This is in contrast to C `struct`s, where each sub-`struct` must be
-appropriately aligned and padded. For example, the C translation of this example
-is
-
-```c
-struct t3 {
-  struct t1 { int32 x; int16 y; } t1;
-  int16 z;
-};
-```
-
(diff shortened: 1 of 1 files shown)


jane/doc/proposals/unboxed-types/index.md
```diff
@@ -1033,19 +1033,7 @@ get from the definition of `t0` or `t2` to a tightly packed representation; use
...
-
-```c
-struct t3 {
-  struct t1 { int32 x; int16 y; } t1;
-  int16 z;
-};
-```
-
-Yet this would take 2 words in memory, as `t1` would be padded in order to be
-32-bit aligned.
```
stedolan, line 1048: This text isn't very wrong and is helpful, so how about fixing it instead of deleting it?

All you need to do is change `2 words` to `12 bytes`: C is padding `t1` to 32-bit alignment, not 64-bit alignment. The analysis is otherwise correct, though.

stedolan (changes requested):

stedolan:
I don't really understand the layout algorithm presented here. What does it do on these examples?

```
type a = #{ p : int32#; q: int16#; r: int32# }
type b = #{ s: int16#; t: a }
```
From the description, I think we end up with 2 bytes of padding between `q` and `r` in an `a`, to ensure that `r` is well aligned. But when inlined into `b`, it looks like `a` will pack better, and the padding will be gone. Is this the intent? If so, it means we can never have anything like interior pointers, because the layout of `a` depends on what it's embedded into. (Maybe that's a reasonable decision, but I'd like to know we're not making it accidentally)

PR #3493, opened 2025-01-21
Always use a macro when checking for a promoted header

NickBarnes:
This introduces a macro `Is_promoted_hd` everywhere that checks a header is marking a promoted block. Every time we test for promotion, we do so using this predicate, so it could be broadened to allow (for instance) multiple promoted header values, rather than one single magic value. At present, it uses the constant value `Promoted_hd`, which can be zero (it is here, so this PR has no semantic effect at runtime) or non-zero (e.g. 0x200 - zero size, non-zero colour bits), which could be helpful for debugging.
This may be a small step towards using such named macros for other predicates, and also towards using distinct types for headers, tags, etc.
This was inspired by a hilarious bug in ocaml/ocaml#13643, see https://github.com/ocaml/ocaml/pull/13643#issuecomment-2602899055

### DIFF
--- a/runtime/caml/minor_gc.h
+++ b/runtime/caml/minor_gc.h
@@ -76,10 +76,10 @@ struct caml_minor_tables {
 };
 
 /* Setting this to a different value can be useful for debugging
- * although may cause some slowdown. I suggest 0x200 (zero size,
- * non-zero colour). Compare with In_progress_update_val in
- * minor_gc.c. */
-#define Promoted_hd ((header_t)0x0)
+ * although may cause some slowdown. It must be a header value not
+ * otherwise found on the minor heap, not using Infix_tag, and not
+ * equal to In_progress_hd.  I suggest Make_header(0 ,0, 0x200). */
+#define Promoted_hd (Make_header(0, 0, 0))
 #define Is_promoted_hd(hd) ((hd) == Promoted_hd)
 
 CAMLextern void caml_minor_collection (void);
--- a/runtime/caml/minor_gc.h
+++ b/runtime/caml/minor_gc.h
@@ -75,6 +75,13 @@ struct caml_minor_tables {
   struct caml_dependent_table dependent;
 };
 
+/* Setting this to a different value can be useful for debugging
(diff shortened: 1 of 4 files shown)


runtime/memprof.c
```diff
@@ -1499,7 +1499,7 @@ static bool entry_update_after_minor_gc(entry_t e, void *data)
   CAMLassert(Is_block(e->block)
              || e->deleted || e->deallocated || e->offset);
   if (!e->offset && Is_block(e->block) && Is_young(e->block)) {
-    if (Hd_val(e->block) == 0) {
+    if (Is_promoted_hd(caml_get_header_val(e->block))) {
```
stedolan: This is a nontrivial change, which is shaped like a bugfix. Is there actually a bug here to fix? (I thought this function ran after the minor GC barrier when `caml_get_header_val` was no longer necessary)

runtime/caml/minor_gc.h
```diff
@@ -75,6 +75,13 @@ struct caml_minor_tables {
   struct caml_dependent_table dependent;
 };
 
+/* Setting this to a different value can be useful for debugging
+ * although may cause some slowdown. I suggest 0x200 (zero size,
+ * non-zero colour). Compare with In_progress_update_val in
+ * minor_gc.c. */
+#define Promoted_hd ((header_t)0x0)
+#define Is_promoted_hd(hd) ((hd) == Promoted_hd)
```
stedolan, line 83:
```
(CAMLassert((hd) == Promoted_hd || Tag_hd(hd) != Infix_tag), (hd) == Promoted_hd)
```
might be a useful (if harder to read) definition of this. (Anyone asking whether an Infix_tag header is promoted is wrong)

stedolan (approved):

runtime/caml/minor_gc.h
NickBarnes: This turns out to be a bad idea, because in many places (such as `oldify_one`), we check `Promoted_hd` before anything else on a header. It seems unlikely that changing this to check `Infix_tag` first would improve overall performance.

PR #3495, opened 2025-01-21
Clear ephemerons which expire in a minor GC

NickBarnes:
This is a rework of @stedolan's work in #3402, to address style points raised in review there, and also fix the hilarious bug found upstream (see https://github.com/ocaml/ocaml/pull/13643#issuecomment-2602899055).


NickBarnes:
The CI failure of this PR is somewhat mysterious, and seems to result from recent changes to CI rather than anything in the PR itself.

NickBarnes:
More on the CI failure: `lib-runtime-events/test_fork.ml` is sometimes creating a core file but also succeeding. That core file is then confusing the CI run. Presumably the core file is made by the child process after it makes the `EV_FORK_CHILD` event. Hard to reproduce and probably unrelated to this actual PR, so we're disregarding it.

NickBarnes:
Rebased and turned off `unsafe_multidomain` alert on a test.

mshinwell:
Could someone please check the upstream PR as there are a couple of code review comments there which might be applicable?

stedolan:
The changes here look good. I'm going to have a look at the upstream comments next.

stedolan:
I've pushed a small change arising from the upstream comments

stedolan (approved):

PR #3498, opened 2025-01-22
Expose externals in the interface of Atomic

gretay-js:
To improve code generation, expose the `external` definitions of the following functions in the interface of `Atomic` in the standard library: `exchange`,`compare_set`,`compare_exchange`. 
It seems to be needed to get them specialized for immediates and emitted as instructions instead of C calls (even in an optimized build). Functions that operate directly on `int` do not need it.


TheNumbat (commented):
We should merge #3497 first and make `set` use `%atomic_set`, but otherwise lgtm

TheNumbat (approved):

PR #3500, opened 2025-01-23
Introduce automatic compaction to runtime 5

NickBarnes:
The OCaml 4 GC does automatic compaction at the end of a major GC when the heap overhead exceeds a given percentage of the live data (default 500%). This feature was removed from OCaml 5 because compaction overall was removed. Compaction was reintroduced in 2024; this PR reintroduces automatic compaction.
This uses the same parameters and criteria as OCaml 4, as far as possible.  Perhaps we should revisit those criteria (in particular, the segregated-fit allocator of OCaml 5 makes it quite hard for a non-pathological program to reach 500% overhead).


runtime/shared_heap.c
```diff
@@ -199,7 +199,27 @@ void caml_teardown_shared_heap(struct caml_heap_state* heap) {
               released, released_large);
 }
 
-/* this _must_ be called either with the pool_freelist.lock held or
+/* TODO: resurrect major_heap_increment */
+
+uintnat new_chunk_bsize(void)
```
stedolan, line 204: `alloc_pool` should now call this function rather than duplicating the logic?

(Also, it looks like the `caml_pool_min_chunk_bsz` logic from 3384 got lost in a merge, possibly of compaction. This restores it, which is another good reason to fix `alloc_pool`)

stedolan (changes requested):
This looks good to me. I'm somewhat doubtful it'll ever actually trigger, but let's see.

runtime/shared_heap.c
NickBarnes: Yes, that's an oversight. Fixed.

stedolan (approved):

PR #3501, opened 2025-01-23
Continuation specialization, aka match-in-match

Gbury:
This PR adds to flambda the capacity to specialize continuations.

For now, it's meant for testing. Once testing is done, code will be cleaned up, and some parts will be split up and submitted as independent PRs for an easier review.

# Context and overview

Following the continuation lifting introduced in #2295 , this PR (finally) adds continuation specialization, i.e. given one continuation `k` with more than one call site, generate on specialized version of ` k` for each call site. The most tricky part of that process is related to continuation lifting in order to ensure that the minimal amount of code is duplicated when specializing one continuation (more on that later).

Consider an expression of the form:
```ocaml
let cont k x =
  let y = untag_int x in
  let cont k' z =
    ..
  in
  switch y with
  | 0 -> k' ...
  | 1 -> k' ...
in
switch .. with
| 0 -> k 0
| 1 -> k 1
```
The process for continuation specialization is the following. 
1. when simplifying (downwards) the let_cont for `k` we first simplify the body (as usual), and see/record all the uses of `k` (as usual)
2. once the downwards pass on the body has been completed, we start the downwards pass on the handler of `k`
3. after some simplifications, we simplify the let_cont for `k'` and start with the downwards traversal of the body of the let_cont
4. we reach the switch at the end of the handler of `k` (i.e. the switch on `y`). At that point we have all the information that we need to know whether specializing `k` will eliminate the switch (for the match-in-match optimization), so we can decide whether to specialize or not. Note: the criterion for deciding when to specialize is not implemented in this PR and will be done later, for now we always specialize if the lifting budget allows for it.
5. let's assume we do decide to specialize `k`. We then also decide to lift any continuation inside of the handler of `k` to avoid duplicating those
6. We have now finished the downwards pass on the body of the let_cont of `k'`, and we now know that we want to specialize `k`, and thus also lift `k'`, so instead of simplifying the handler of `k'`, we lift it by storing it in the dacc (this is what #2295 does)
7. We now reach the down_to_up for the handler of `k`, we first pop the lifted continuations out of the `dacc` and craft a new `down_to_up` function that will simplify them. We can then start the specialization of `k`
8. We generate the specialized versions of `k` by copying the original handler (before simplification) for each callsite, and then doing the downwards pass on each of these "new" continuations.
9. For now, we always completely specialize (i.e. generate exactly one continuation for each callsite), so we adjust the free names and other metadata to completely erase information coming from the downwards traversal fo the original handler for `k` (in the future we might consider specializing only some callsites and keeping a "generic" version, but that will be for later)
10. During the downwards pass on the specialized versions of `k`, we must replace any bound continuation (whose name will be fresh as the let_cont will be freshly opened by the downwards traversal) by its lifted version, and compute adequate lifted arguments. This is where we use the notion of Replay_history to relate the freshly bound continuation (and the lifted arguments) with the names from the first downwards traversal on `k`.
11. We record the mapping from callsites to the name of the new continuations as that will be used later (in the data flow analysis and during rebuilding)
12.During data flow analysis, we rewrite the internal graphs using the mapping created at point 11 (to reflect the term that will be rebuilt)
12. When rebuilding terms we rewrite the calls of `k` to refer to the specialized version, using the mapping from point 10.

# Replay History

The notion of replay_history is introduced in #3302 (not merged yet) precisely to make it possible to lift a continuation during the first downwards pass on `k` (the specialized continuation), and then refer to the lifted continuation in the specialized versions of `k`.

Let's detail a bit more why that is required/necessary. Let's consider the same example as above, when we perform the downwards pass on the specialized versions of `k` (let's call them `k_0` and `k_1`), what the downwards traversals effectively see/observe is a term that would look like the following:

```ocaml
let cont k x =
  let y = untag_int x in
  let cont k' z =
    ..
  in
  switch y with
  | 0 -> k' ...
  | 1 -> k' ...
in
let cont k_0 x_0 =
  let y_0 = untag_int x_0 in
  let cont k'_0 z_0 =
    ..
  in
  switch y_0 with
  | 0 -> k'_0 ...
  | 1 -> k'_0 ...
in
let cont k_1 x_1 =
  let y_1 = untag_int x_1 in
  let cont k'_1 z_1 =
    ..
  in
  switch y_1 with
  | 0 -> k'_1 ...
  | 1 -> k'_1 ...
in
switch .. with
| 0 -> k_0 0
| 1 -> k_1 1
```

In particular because of the automatic name abstraction renaming, we have no easy way to relate the lifted continuation `k'` with its corresponding continuations `k'_0` and `k'_1`, except for the fact that all of those can be identified as the first continuation that is bound in the handler of `k`. Additionally, when lifting `k'`, we add to it the lifted parameter `y'` that was in scope at its original binding site and thus must be kept available in its handler: this brings another problem which is that we need to add the adequate argument for this new parameter at every callsite of `k'`, including those in the specialized continuations `k_0` and `k_1` where its value will be `y_0` and `y_1`, which can only be related to `y` (the value of the extra parameter `y'` in the original handler of `k`) because they are the first variable that is bound in the handler of `k` (and similarly for `x`, the first parameter of `k`).

Therefore, in order to correctly lift continuations out of the handler of specialized continuations, we need to be able to relate the names of parameters, variables and continuations that are bound in the handler of `k` beetween successive downwards traversal. More precisely, given a parameter/variable/continuation bound in the handler of a specialized version of `k`, we want to know what was the corresponding name generated during the first downwards pass (the one on the original handler of `k`).

That's where the notion of Replay_history comes from: basically, it's an ordered list of the successive parameters/variables/continuations that are bound in the handler of a given continuation. By creating and storing that list during the first pass (i.e. the downwards pass on the original handler of `k`), we can then on successive downwards passes establish a mapping from the names of the current (specialized) pass to the names of the original pass.

The replay history tries to be as "secure" as possible (i.e. prevent against establishing a mapping between two downwards pass on different expressions) by raising errors if the order or variables/continuations bound does not match, and by verifying that the corresponding names are renamed version of one another, which does **not** protect against any misuse since two unrelated variables/continuations can be seen as renamed version of one another if they have the same names in some situations, but at least offers some measure of safeguard.

# Lifted continuations

In the handler of the specialized continuations, we replace the handler of bound continuations by a call to the lifted continuation, and relay on the usual mechanism to correctly compute the adequate arguments for the lifted parameters. This works seamlessly because the lifted_cont_params structure interacts with the replay_history to correctly map the unique key of lifted params (i.e. the name of the variables in the original downwards pass) to their name in the current pass.

# Conclusion

General remarks/comments/questions about how everything works are welcome. Detailed remarks about the code are likely to not be of limited use until after testing (particularly once there are no bugs remaining), contrary to remarks about the general design and organisation of the feature, ^^

I'll ping the relevant persons when the PR is ready for reviews, but anyone should feel free to look at the code (keeping in mind that there's a fair bit of code cleanup to do).


Gbury:
This is ready for review, ^^

Gbury:
(ci failure appears to be due to the famous heisenbug, which will be solved by #3960 )

lukemaurer:
Shouldn't the budget for specialization be the amount of code we're willing to duplicate? In cases where we're not duplicating code, specialization is both quick (we'll just see a switch and pick a branch, no?) and highly desirable, so the cost should be zero (even negative in cases where we end up eliminating the switch).

Concretely, a very reasonable function might be doing a long series of `Result.bind`s, every one of which might be statically `Ok` or `Error`. Specialization there is all benefit and no cost and we'll be sad if we randomly stop doing it.

lukemaurer:
A nice example of something this optimizes that was asked about just today at Jane Street:

```ocaml
let[@inline never] f x =
  if x = 0 then This x else Null

let e = Error "aye"

let[@inline] bind x ~f = exclave_
  match x with
  | Error _ as e -> e
  | Ok x -> (f [@inlined hint]) x
;;

let g x = exclave_
  let y =
    match f x with
    | Null -> e
    | This x -> Ok x
  in
  bind y ~f:(fun y -> exclave_
    Ok (y + 1)) [@nontail]
```

The generated code is identical to writing

```ocaml
let g x = exclave_
  match f x with
  | Null -> e
  | This x -> Ok (x + 1)
```

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
```diff
@@ -1263,21 +1582,30 @@ and simplify_handlers ~simplify_expr ~down_to_up ~denv_for_join ~rebuild_body
...
       let dacc =
         DA.with_are_lifting_conts dacc
-          (Are_lifting_conts.think_about_lifting_out_of cont uses)
+          (Are_lifting_conts.think_about_lifting_out_of ~is_exn_handler cont
+             uses)
       in
+      if debug ()
+      then
+        Format.eprintf "^^^ CUE BODY %a@\n%a@\n@." Continuation.print cont
+          CUE.print body_continuation_uses_env;
```
Ekdohibs, line 1608:
```suggestion
```

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
```diff
@@ -1318,23 +1653,26 @@ and simplify_handlers ~simplify_expr ~down_to_up ~denv_for_join ~rebuild_body
...
-          let dacc =
-            (* Update the dacc with the new cont_uses_env, which removes the
-               uses of [cont] since it leaves scope. *)
-            let cont_uses_env = CUE.remove cont_uses_env_so_far cont in
-            DA.with_continuation_uses_env dacc ~cont_uses_env
-          in
+          (* let dacc = (* Update the dacc with the new cont_uses_env, which
+             removes the uses of [cont] since it leaves scope. *) let
+             cont_uses_env = CUE.remove cont_uses_env_so_far cont in
+             DA.with_continuation_uses_env dacc ~cont_uses_env in *)
```
Ekdohibs, line 1659:
```suggestion
```

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
```diff
@@ -1318,23 +1653,26 @@ and simplify_handlers ~simplify_expr ~down_to_up ~denv_for_join ~rebuild_body
...
           let dacc =
             DA.with_are_lifting_conts dacc previous_are_lifting_conts
           in
           let data : after_downwards_traversal_of_body_and_handlers_data =
             { rebuild_body;
+              dacc_after_body;
+              cont_uses_env_after_body = body_continuation_uses_env;
+              after_downwards_traversal_of_body = data;
               cont_uses_env = cont_uses_env_so_far;
+              (* CUE.remove cont_uses_env_so_far cont; *)
```
Ekdohibs, line 1669:
```suggestion
```

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
```diff
@@ -1410,13 +1750,20 @@ and after_downwards_traversal_of_body ~simplify_expr ~down_to_up
     let handlers =
       Original_handlers.add_params_to_lift data.handlers params_to_lift
     in
+    if debug ()
+    then
+      Format.eprintf "/// LIFTING:@\n%a@\n@." Original_handlers.print handlers;
```
Ekdohibs, line 1755:
```suggestion
```

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
```diff
@@ -1410,13 +1750,20 @@ and after_downwards_traversal_of_body ~simplify_expr ~down_to_up
...
     (* Restore lifted constants in dacc *)
     let dacc =
       DA.add_to_lifted_constant_accumulator dacc data.prior_lifted_constants
     in
     down_to_up dacc ~rebuild:rebuild_body
   | Not_lifting | Analyzing _ ->
+    if debug ()
+    then
+      Format.eprintf "... SIMPLIFY@\n%a@\n@." Original_handlers.print
+        data.handlers;
```
Ekdohibs, line 1766:
```suggestion
```

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
```diff
@@ -1438,6 +1785,7 @@ and down_to_up_for_lifted_continuations ~simplify_expr ~denv_for_join
     let data : after_downwards_traversal_of_body_data =
       { denv_for_join = actual_denv;
         prior_lifted_constants = LCS.empty;
+        (* TODO: is this correct ? *)
```
Ekdohibs, line 1788: Remaining TODO, I think this is correct

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
```diff
@@ -1482,10 +1830,60 @@ let simplify_let_cont0 ~(simplify_expr : _ Simplify_common.expr_simplifier) dacc
...
   in
+  (* During specialization, we must take care of correctly handling let-bound
+     continuations that have been lifted during the first downwards pass, and
+     whhose handlers must be replaced on subsequent passes, so that they can
+     refer to the lifted ones. *)
+  if debug ()
+  then
+    Format.eprintf "--- LET_CONT0@\n%a@\n%a@\n@." Replay_history.print
+      (DE.replay_history denv_for_body)
+      Original_handlers.print data.handlers;
```
Ekdohibs, line 1841:
```suggestion
```

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
```diff
@@ -1482,10 +1830,60 @@ let simplify_let_cont0 ~(simplify_expr : _ Simplify_common.expr_simplifier) dacc
     DE.add_lifting_cost lifting_cost
       (DE.define_continuations denv_for_body bound_continuations)
   in
+  (* During specialization, we must take care of correctly handling let-bound
+     continuations that have been lifted during the first downwards pass, and
+     whhose handlers must be replaced on subsequent passes, so that they can
```
Ekdohibs, line 1835:
```suggestion
     whose handlers must be replaced on subsequent passes, so that they can
```

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
```diff
@@ -1482,10 +1830,60 @@ let simplify_let_cont0 ~(simplify_expr : _ Simplify_common.expr_simplifier) dacc
...
+    with
+    | Still_recording -> data.handlers
+    | Replayed continuation_mapping -> (
+      match data.handlers with
+      | Non_recursive non_rec_handler ->
+        if debug ()
+        then
+          Format.eprintf "+++ LIFTED ? %a@\nreplay: %a@\n@." Continuation.print
+            non_rec_handler.cont Replay_history.print
+            (DE.replay_history denv_for_body);
```
Ekdohibs, line 1855:
```suggestion
```

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
```diff
@@ -1482,10 +1830,60 @@ let simplify_let_cont0 ~(simplify_expr : _ Simplify_common.expr_simplifier) dacc
...
+        then
+          Format.eprintf "*** LIFTED %a:@\n%a@\n@." Continuation.print
+            non_rec_handler.cont Flambda.Expr.print new_handler;
+        let new_non_rec_handler =
+          Non_recursive_handler.with_handler new_handler non_rec_handler
+        in
+        Original_handlers.create_non_recursive new_non_rec_handler
+      | Recursive _ ->
+        (* TODO: handle these *)
+        assert false)
```
Ekdohibs, line 1880: I think we don't specialize recursive continuations, we should at least put a nice `Misc.fatal_error` here

Ekdohibs (commented):
Mostly good, see comments. I'm not sure about removing all the debug printing code, but if we keep it, it should be behind a continuation-specialization–specific flag.

middle_end/flambda2/simplify/env/downwards_acc.ml
```diff
@@ -279,22 +294,62 @@ let get_continuation_lifting_budget t =
...
   if t.continuation_lifting_budget < 0
   then t
   else
-    { t with
-      continuation_lifting_budget = max 0 (t.continuation_lifting_budget - cost)
-    }
+    with_continuation_lifting_budget t
+      (max 0 (t.continuation_lifting_budget - cost))
+
+(* CR gbury: remove this code and use a proper heuristic for specialization *)
```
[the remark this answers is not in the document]
Gbury: Indeed.

middle_end/flambda2/simplify/flow/flow_acc.ml
```diff
@@ -450,16 +469,16 @@ let extend_args_with_extra_args (t : T.Acc.t) =
                  flow analysis find a dominator for parameters of [cont] that we
                  do not have access to. Since we then only look at which
                  continuation can call which to know which arguments to add,
-                 this causes the ‌adding of the extra parameters to assume we
+                 this causes the adding of the extra parameters to assume we
```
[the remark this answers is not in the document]
Ekdohibs: It looks like a ZWNJ was removed, even though I don't know how it ended here in the first place.

middle_end/flambda2/simplify/env/continuation_uses.mli
```diff
@@ -35,7 +35,7 @@ val add_use :
 
 val get_uses : t -> One_continuation_use.t list
 
-type arg_at_use = private
+type arg_at_use =
```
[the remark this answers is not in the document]
Gbury: It's because in `simplify_let_cont_expr` (around line 235), we actually create a map with type `arg_at_use Apply_cont_rewrite_id.Map.t` to represent the arguments for the lifted parameters when making (and computing extra args for) the unboxing decisions.

middle_end/flambda2/simplify/env/downwards_acc.ml
```diff
@@ -35,7 +35,11 @@ type t =
     are_lifting_conts : Are_lifting_conts.t;
     lifted_continuations : (DE.t * Original_handlers.t) list;
     (* head of the list is the innermost continuation being lifted *)
-    continuation_lifting_budget : int
+    continuation_lifting_budget : int;
+    continuation_specialization_budget : int;
+    continuations_to_specialize : Continuation.Set.t;
+        (* TODO* encode that into the map below as the keys of the map *)
```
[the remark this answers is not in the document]
Gbury: It might be, but I don't think it's urgent or needed right now, so i'll switch that to a CR I think.

middle_end/flambda2/simplify/specialization_cost.ml
```diff
@@ -0,0 +1,67 @@
...
+(*   All rights reserved.  This file is distributed under the terms of    *)
+(*   the GNU Lesser General Public License version 2.1, with the          *)
+(*   special exception on linking described in the file LICENSE.          *)
+(*                                                                        *)
+(**************************************************************************)
+
+type reason =
+  | At_toplevel
+  | Contain_static_consts
+  | Contain_set_of_closures
```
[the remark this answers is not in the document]
Gbury: Indeed, I changed it

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
```diff
@@ -966,71 +1054,280 @@ let sort_handlers data handlers =
...
+           information coming from the first traversal of the handler. We still
+           reset a few values that come from the dacc after simplification of
+           the handler, see comment above. *)
+        let dacc = data.dacc_after_body in
+        let dacc = DA.with_continuation_lifting_budget dacc lifting_budget in
+        let dacc =
+          DA.with_continuation_specialization_budget dacc spec_budget
+        in
+        (* We need to adjust the contination uses env with a few things: TODO:
+           continue this comment *)
```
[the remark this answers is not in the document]
Gbury: done

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
```diff
@@ -966,71 +1054,280 @@ let sort_handlers data handlers =
...
+            handlers = Continuation.Map.empty;
+            cont_uses_env;
+            cont_uses_env_after_body = cont_uses_env
+          }
+        in
+        compute_specialized_continuation_handlers ~simplify_expr ~replay
+          ~original_cont:cont ~handler dacc data uses k))
+  | _ -> k dacc data
+
 (* CR pchambart: Those functions don't need to be recursive, it would be nice to
```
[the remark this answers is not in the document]
Gbury: comment moved above the let rec (and slightly updated)

Ekdohibs (approved):

PR #3502, opened 2025-01-23
Track parent regions in Flambda2

lthls:
This PR fixes a bug in the following example (@mshinwell crafted the example a month or so ago):
```ocaml
type t = { mutable state : unit }
let[@inline never] [@local never] run (local_ f) = f ()
let[@inline never] [@local never] bad () =
  (fun () ->
    let ctx = { state = () } in
    run (fun () ->
      ignore (Sys.opaque_identity ctx : _);
      ())
    [@nontail])
    () [@nontail]
let () =
  while true do
    bad ()
  done
```
This example leaks local allocations, because the anonymous function passed to `run` is allocated locally, and should be cleaned up when exiting the anonymous function that contains it; but after inlining this anonymous function, its region closes at the same point as the region for `bad`, so we use the end of the region of `bad` to close both regions at once. However, since no allocation occurs directly in the region for `bar`, we remove this region, and the `End_region` that was supposed to clean up the inner allocation disappears.

We fix this in this PR by tracking relations between regions: a region that is created inside another one takes its parent as parameter, so a region can't be considered unused unless all its children are also unused.
We are considering improving this later with an analysis that can detect when a region never relies on its parent for cleaning up allocations, in which case we can remove the extra dependency and clean up the parent.

This PR consists of three meaningful commits (at the time of submission):

- `Make Begin_region Variadic` changes the `Begin_region` primitive from a nullary primitive to a variadic primitive. This should be a no-op: all occurrences of the primitives simply get an empty list of arguments.
- Add parent regions as parameters to Begin_region` builds on it to implement the fix discussed above: every `Begin_region` gets a single argument, which is the current region at the place where the primitive is inserted.
This commit fails to build, though, because of internal invariants requiring that functions containing no escaping local allocations should not contain any occurrence of their region parameter.
- So the last commiy, `Make my_region optional`, is a relatively big refactoring to track whether there is or not a meaningful region at a given place in the code. Functions that do not contain escaping local allocations no longer take any region parameters, and this propagates in a number of places. Since every local allocation in Flambda2 takes the region in which it must be allocated, checking for correctness wasn't too hard: if a local allocation (or local application) occurs in a context where no region is available, this is a fatal error at compile time. I had to tweak a few things here and there though, as some pieces of code around regions were just wrong. (For instance, the `region` parameter on Lambda function declarations is now ignored, as it doesn't correspond to anything useful in Flambda2. A function now is considered to contain escaping local allocations iff its return mode allows local allocations).


middle_end/flambda2/from_lambda/lambda_to_flambda.ml
```diff
@@ -1428,7 +1480,7 @@ and cps_function env ~fid ~(recursive : Recursive.t) ?precomputed_free_idents
     ~params_arity ~removed_params ~return ~calling_convention
     ~return_continuation:body_cont ~exn_continuation ~my_region ~my_ghost_region
     ~body ~attr ~loc ~free_idents_of_body recursive ~closure_alloc_mode:mode
-    ~first_complex_local_param ~contains_no_escaping_local_allocs:region
```
Ekdohibs, line 1431: I think we want to allow functions with `contains_no_escaping_local_allocs = true` but `ret_mode = Alloc_local`, when the result is not actually allocated on the stack.

lthls: In flambda2, maybe, but we can't rely on `region` for that (it only means that a region was inserted, but with `Lexclave` that doesn't mean that no allocations escape).
So we cannot set the `contains_no_escaping_local_allocations` flag to `true` when the return mode is `Alloc_local` unless we perform a dedicated analysis.

riaqn: Please note that #2985 which we will merge very soon, removes `contains_no_escaping_local_allocations` - would that simplify or complicate this discussion?

lthls: In a way, this PR re-introduces it under another name: the `my_region` parameter is now optional, if it is absent it means that the function doesn't contain any escaping local allocations. So we still conservatively put a `my_region` parameter whenever the return mode is `Local`, and could remove it with an analysis if it turns out to be not necessary.

Ekdohibs (approved):
Looks good, there's the question of performing a dedicated analysis for `contains_no_escaping_local_allocs` at some point, but that's outside the scope of this PR.

lthls:
You can merge #2985 first if it's ready and I will deal with rebasing.

riaqn:
@lthls cool - #2985 is now merged. :-)

PR #3503, opened 2025-01-23
Move (most of) the type definitions for jkind into Types

glittershark:
Jkinds are morally mutually recursive with types, since jkinds can contain
types (in the with-bounds) and types can be bounded on jkinds (eg for Tvar and
Tunivar). We previously "broke up" this recursion with a type parameter for
type_expr on the bits of Jkind.t that contain type_exprs, and by defining two
modules, Jkind_types and Jkind - the first with the type parameter, and the
second with the parameter instantiated to Types.type_expr. This has caused a lot
of headache and confusion, is hard to follow (you lose jump-to-definition, among
other things) and generally puts friction on algorithms for kinds that need to
do things with type_exprs (which we're about to do a lot more, for normalizing
of jkinds).

Instead, this commit starts the (multi-stage) process of splitting up jkind,
starting by just moving the type definitions (omitting sort and layout so far)
into types.ml, in the same recursive type declaration as type_expr. This
simplifies things a _lot_, without really losing much in terms of things being
broken up (the algorithms are still just defined in jkind.ml)!


goldfirere (approved):

goldfirere:
Last 2 commits looks good. I think this one's all set

PR #3504, opened 2025-01-24
Add environment variable FLAMBDA2_UNTRUSTED_KINDS

lthls:
When set to a non-empty string, this tells Flambda2 to ignore the subkind information from the type-checker. This can be used if the value kinds in lambda are known to be wrong.


lthls:
Unnecessary since #3507 has been merged.

PR #3505, opened 2025-01-24
Track quality of jkinds

glittershark:
Sometimes, inferred jkinds for a type are not exact, but instead *upper bounds*;
later, we might learn more about a type that causes us to lower that jkind
bound. For example, we might substitute a type variable, or perform `with`
substitution on an abstract type in a module type. It's important to know
whether this might happen, so that we can avoid normalizing away types with
jkinds that might get "better" later from with-bounds during jkind
normalization.

This commit adds a new "jkind_quality" type, and a "quality" field to jkind,
that tracks whether an inferred jkind is "best" or "not best", and sets this to
the correct value everywhere jkinds are constructed. Nothing actually /reads/
this field yet, but this will be used during normalization in a subsequent
commit.

### DIFF
--- a/typing/jkind.ml
+++ b/typing/jkind.ml
@@ -1576,6 +1576,9 @@ module Builtin = struct
         layouts
     in
     fresh_jkind_poly desc ~annotation:None ~why:(Product_creation why)
+    (* [mark_best] is correct here because the with-bounds of a product jkind include all
+       the components of the product. Accordingly, looking through the product, by one
+       step, never loses any information. *)
     |> mark_best
 
   let product_of_sorts ~why arity =
@@ -1587,6 +1590,9 @@ module Builtin = struct
       { layout; mod_bounds = Mod_bounds.max; with_bounds = No_with_bounds }
     in
     fresh_jkind_poly desc ~annotation:None ~why:(Product_creation why)
+  (* We do not [mark_best] here because the resulting jkind is used (only) in the middle of
+     type-checking mutually recursive type declarations. See Note [Default jkind in
+     transl_declaration] for more commentary on why we don't want [Best] jkinds there. *)
 end
 
 let add_nullability_crossing t =
--- a/typing/jkind.ml
+++ b/typing/jkind.ml
@@ -1510,10 +1510,7 @@ let mk_annot name =
 let mark_best (type l r) (t : (l * r) Types.jkind) =
   { (disallow_right t) with quality = Best }
 
-let is_best t =
-  match t.quality with
-  | Best -> true
-  | Not_best -> false
+let is_best t = match t.quality with Best -> true | Not_best -> false
 
 module Builtin = struct
   let any_dummy_jkind =
--- a/typing/jkind.ml
+++ b/typing/jkind.ml
@@ -1772,7 +1772,7 @@ let for_boxed_variant cstrs =
        (* CR layouts v2.8: This is sad, but I don't know how to account for
           existentials in the with_bounds. See doc named "Existential
           with_bounds". *)
-    then Builtin.value ~why:Boxed_variant |> mark_best
+    then Builtin.value ~why:Boxed_variant
     else
       let base =
         (if is_mutable then Builtin.mutable_data else Builtin.immutable_data)
--- a/typing/jkind.ml
+++ b/typing/jkind.ml
@@ -2471,7 +2471,8 @@ let intersection_or_error ~type_equal ~jkind_of_type ~reason t1 t2 =
           combine_histories ~type_equal ~jkind_of_type reason (Pack_jkind t1)
             (Pack_jkind t2);
         has_warned = t1.has_warned || t2.has_warned;
-        quality = Not_best (* As required by the fact that this is a [jkind_r] *)
+        quality =
+          Not_best (* As required by the fact that this is a [jkind_r] *)
       }
 
 let round_up (type l r) ~type_equal ~jkind_of_type (t : (allowed * r) jkind) :
--- a/typing/jkind.ml
+++ b/typing/jkind.ml
@@ -733,31 +733,92 @@ end
 
 (*********************************)
 
+module Quality = struct
+  include Allowance.Magic_allow_disallow (struct
+    type (_, _, 'd) sided = 'd jkind_quality constraint 'd = 'l * 'r
+
+    let disallow_left :
+        type l r. (l * r) jkind_quality -> (disallowed * r) jkind_quality =
+      function
+      | Not_best -> Not_best
+      | Best -> Best
+
+    let disallow_right :
+        type l r. (l * r) jkind_quality -> (l * disallowed) jkind_quality =
+      function
+      | Not_best -> Not_best
+      | Best -> Best
+
+    let allow_left :
+        type l r. (allowed * r) jkind_quality -> (l * r) jkind_quality =
+      function
+      | Not_best -> Not_best
+      | Best -> Best
+
+    let allow_right :
+        type l r. (l * allowed) jkind_quality -> (l * r) jkind_quality =
+      function
+      | Not_best -> Not_best
+  end)
+
+  let try_allow_r :
+      type l r. (l * r) jkind_quality -> (l * allowed) jkind_quality option =
+    function
+    | Not_best -> Some Not_best
+    | Best -> None
+end
+
 include Allowance.Magic_allow_disallow (struct
   type (_, _, 'd) sided = 'd jkind
 
   let disallow_right t =
-    { t with jkind = Layout_and_axes.disallow_right t.jkind }
+    { t with
+      jkind = Layout_and_axes.disallow_right t.jkind;
+      quality = Quality.disallow_right t.quality
+    }
 
-  let disallow_left t = { t with jkind = Layout_and_axes.disallow_left t.jkind }
+  let disallow_left t =
+    { t with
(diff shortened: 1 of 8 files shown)


typing/jkind.ml
```diff
@@ -1529,6 +1600,7 @@ module Builtin = struct
       { layout; mod_bounds = Bounds.max; with_bounds = No_with_bounds }
     in
     fresh_jkind_poly desc ~annotation:None ~why:(Product_creation why)
+    |> mark_best
```
goldfirere, line 1603: This is like the `mark_best` calls in typedecl, and is suspect -- note the `mod_bounds = Bounds.max` which doesn't look very best-like.

typing/jkind.ml
```diff
@@ -1703,11 +1781,12 @@ let for_boxed_variant cstrs =
        (* CR layouts v2.8: This is sad, but I don't know how to account for
           existentials in the with_bounds. See doc named "Existential
           with_bounds". *)
-    then Builtin.value ~why:Boxed_variant
+    then Builtin.value ~why:Boxed_variant |> mark_best
```
goldfirere, line 1784: As discussed, I think we shouldn't `mark_best` here. GADTs will reliably have jkind `value` now, but this will change. Marking this `not_best` for now will, I think, make things easier later when we improve the jkind.

typing/predef.ml
```diff
@@ -245,7 +246,7 @@ let mk_add_type add_type =
       {type_params = [];
        type_arity = 0;
        type_kind = kind;
-       type_jkind = jkind;
+       type_jkind = Jkind.mark_best jkind;
```
goldfirere, line 249: Should do something similar in `mk_add_type1`, which will lessen the diff here.

typing/jkind.ml
```diff
@@ -1574,19 +1646,24 @@ let of_new_legacy_sort_var ~why =
...
     history = Creation why;
-    has_warned = false
+    has_warned = false;
+    quality
   }
 
 let of_builtin ~why Const.Builtin.{ jkind; name } =
-  of_const ~annotation:(mk_annot name) ~why jkind |> allow_left |> allow_right
+  jkind |> Layout_and_axes.allow_left |> Layout_and_axes.disallow_right
+  |> of_const ~annotation:(mk_annot name) ~why ~quality:Best
```
goldfirere, line 1660: The use of `Best` here scared me. But it's OK. I would still leave a comment like

```
(* The [Best] is OK here because this function is used only in Predef. *)
```

typing/jkind.ml
```diff
@@ -1574,19 +1646,24 @@ let of_new_legacy_sort_var ~why =
...
 
 let of_annotated_const ~context ~annotation ~const ~const_loc =
   let context = Context_with_transl.get_context context in
-  of_const ~annotation ~why:(Annotated (context, const_loc)) const
+  of_const ~annotation
+    ~why:(Annotated (context, const_loc))
+    const ~quality:Not_best
```
goldfirere, line 1666: This choice of `Not_best` should be reflected in the mli, at least for `of_type_decl` and `of_type_decl_default`.

typing/jkind.ml
```diff
@@ -2387,7 +2468,8 @@ let intersection_or_error ~type_equal ~jkind_of_type ~reason t1 t2 =
         history =
           combine_histories ~type_equal ~jkind_of_type reason (Pack_jkind t1)
             (Pack_jkind t2);
-        has_warned = t1.has_warned || t2.has_warned
+        has_warned = t1.has_warned || t2.has_warned;
+        quality = Not_best
```
goldfirere, line 2472: This is a right-kind, so this `Not_best` carries no information, right? Then I would add a

```
(* as required by the fact that this is a [jkind_r] *)
```

to this line to tell the reader to look away.

typing/jkind.ml
```diff
@@ -2398,7 +2480,10 @@ let round_up (type l r) ~type_equal ~jkind_of_type (t : (allowed * r) jkind) :
           (fun ~axis -> upper_bound_for_axis ~type_equal ~jkind_of_type ~axis t)
       }
   in
-  { t with jkind = { t.jkind with mod_bounds; with_bounds = No_with_bounds } }
+  { t with
+    jkind = { t.jkind with mod_bounds; with_bounds = No_with_bounds };
+    quality = Not_best
```
goldfirere, line 2485: Similarly here. There is nothing interesting to see here.

goldfirere:
Reviewed two commits ("add some extra comments from review".."more comments, specifically documenting quality"). They look good to me.

typing/typedecl.ml
```diff
@@ -923,7 +925,7 @@ let transl_declaration env sdecl (id, uid) =
         let rep, jkind =
           if unbox then
             Variant_unboxed,
-            Jkind.of_new_legacy_sort ~why:Old_style_unboxed_type
+            (Jkind.of_new_legacy_sort ~why:Old_style_unboxed_type |> Jkind.mark_best)
```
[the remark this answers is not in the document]
glittershark: in fact it is fine to just never mark anything as best in typedecl. Did that.

typing/typedecl.ml
```diff
@@ -1838,7 +1843,10 @@ let update_decl_jkind env dpath decl =
   let new_decl, new_jkind = match decl.type_kind with
     | Type_abstract _ -> decl, decl.type_jkind
```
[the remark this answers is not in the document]
glittershark: added an assert (I believe this can never happen, especially now) but lmk if you prefer something else

typing/typedecl.ml
```diff
@@ -1838,7 +1843,10 @@ let update_decl_jkind env dpath decl =
...
     | Type_open ->
-      let type_jkind = Jkind.Builtin.value ~why:Extensible_variant in
+      let type_jkind =
+        Jkind.Builtin.value ~why:Extensible_variant
+        |> Jkind.mark_best
+      in
       { decl with type_jkind }, type_jkind
     | Type_record (lbls, rep) ->
```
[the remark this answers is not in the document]
glittershark: it happens later, in the normalization PR

typing/jkind.ml
glittershark: and in fact we don't need it.

typing/typedecl.ml
```diff
@@ -979,7 +979,8 @@ let transl_declaration env sdecl (id, uid) =
           Ttype_record_unboxed_product lbls,
           Type_record_unboxed_product(lbls', Record_unboxed_product), jkind
       | Ptype_open ->
-        Ttype_open, Type_open, Jkind.Builtin.value ~why:Extensible_variant
+        Ttype_open, Type_open,
+        (Jkind.Builtin.value ~why:Extensible_variant |> Jkind.mark_best)
```
goldfirere, line 983: Why this change? I think it's probably sound, because extensible types never get better kinds, but I think it's a nice property to have all jkinds used in the midst of type-checking a declaration group to have non-best kinds. At the least, I think there should be a nice comment here explaining why we do this. (And, come to think of it, additions to the big comment about jkind defaults describing how quality flows through the algorithm.)

typing/jkind.ml
```diff
@@ -1605,7 +1605,6 @@ module Builtin = struct
       { layout; mod_bounds = Bounds.max; with_bounds = No_with_bounds }
     in
     fresh_jkind_poly desc ~annotation:None ~why:(Product_creation why)
-    |> mark_best
```
goldfirere: It would be good to comment why `best` is ok in `product` but not here in `product_of_sorts`. I think it's this:

```ocaml
(* [mark_best] is correct here because the with-bounds of a product jkind include all the components
of the product. Accordingly, looking through the product, by one step, never loses any information. *)
```
in `product` and then
```
(* We do not [mark_best] here because the resulting jkind is used (only) in the middle of type-checking mutually recursive type declarations. See Note [Default jkind in transl_declaration] for more commentary on why we don't want [best] jkinds there. *)
```

goldfirere (approved):
Checked the most recent commits ("Define the jkind of or_null once, rather than redefining in typedecl".."simplify best jkinds in predef for type1s", "format").

typing/typedecl.ml
glittershark: It turns out we don't need this, so I just got rid of it.

typing/typedecl.ml
```diff
@@ -1839,10 +1839,7 @@ let update_decl_jkind env dpath decl =
       assert (not (Jkind.is_best decl.type_jkind));
       decl, decl.type_jkind
     | Type_open ->
-      let type_jkind =
-        Jkind.Builtin.value ~why:Extensible_variant
-        |> Jkind.mark_best
-      in
+      let type_jkind = Jkind.Builtin.value ~why:Extensible_variant in
```
goldfirere, line 1842: I agree with not marking the other one... but this one I think is important. It allows, I think, the following:

```
type open_ = ..
type t : immediate with open_

[%%expect {|
type open_ = ..
type t
|}]
```

where we normalize away the kind of `t` to just be `value`.

I know it's awkward because we can't test this in this patch, so I would either restore the `mark_best` here or leave a comment saying to do so in the later patch (where we can test and verify my suspicion).

Really, maybe what I'm saying is that the quality.hs test in the future patch should have tests with extensible types, which I think were left out. I will make a comment saying so.

goldfirere:
Reviewed latest commits. This one looks good

PR #3507, opened 2025-01-24
Track types of variables bound by `as`

lpw25:
Keep the types of variables bound by `as` because they may not equal the types of the `as` patterns themselves. This fixes some incorrect value kinds that can lead to miscompilation.

Currently the following is miscompiled with `-O3`:
```ocaml
let[@inline never] use_error _error = ()

let f x =
  match x with
  | Error ( `A as error )
  | Ok ( `B () as error ) -> use_error error
  | Ok `C -> ()

let () = f (Ok (`B ()))
```

### DIFF
--- a/chamelon/compat.jst.ml
+++ b/chamelon/compat.jst.ml
@@ -297,10 +297,10 @@ type tpat_var_identifier = Value.l
 let mkTpat_var ?id:(mode = dummy_value_mode) (ident, name) =
   Tpat_var (ident, name, Uid.internal_not_actually_unique, mode)
 
-type tpat_alias_identifier = Value.l
+type tpat_alias_identifier = Value.l * Types.type_expr
 
-let mkTpat_alias ?id:(mode = dummy_value_mode) (p, ident, name) =
-  Tpat_alias (p, ident, name, Uid.internal_not_actually_unique, mode)
+let mkTpat_alias ~id:(mode, ty) (p, ident, name) =
+  Tpat_alias (p, ident, name, Uid.internal_not_actually_unique, mode, ty)
 
 type tpat_array_identifier = mutability * Jkind.sort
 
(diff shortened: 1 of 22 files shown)


lpw25:
The following code miscompiles to give a segfault upstream due to the same issue:
```ocaml
type r = { foo : float }

type 'a t = Left of 'a | Right of r

type 'a ty =
  | Float : float ty
  | Anything : 'a ty

let print_if_b = function
  | `A -> ()
  | `B s -> print_endline s

let f (type a) (ty : a ty) (x : a t) =
  match ty, x with
  | Float, Right { foo = (((3.5 : a) as a) : float) }
  | _, Left a -> ignore a
  | _, _ -> ()

let f = Sys.opaque_identity f

let () = f Anything (Left 0)
```
It tries to unbox `a` as a float even though it can actually be any type.

(tested with 4.14.0)

goldfirere (commented):
Should this have tests with it?

lpw25:
> Should this have tests with it?

Ideally, but I'm not sure how to produce a reliable test for the issue. I could just put the repro case above in, but I'm not confident it would continue to test the underlying issue as flambda2 evolves.

goldfirere:
I think adding that would be better than nothing. And what's the nature of the fix? Giving a more accurate type to `error` (in the original report)? If so, then that should be visible as a change to some inferred type, I would think.

lthls:
> And what's the nature of the fix? Giving a more accurate type to `error` (in the original report)? If so, then that should be visible as a change to some inferred type, I would think.

I have written a slightly more detailed explanation in [the upstream PR](https://github.com/ocaml/ocaml/pull/13763).
If my understanding is correct, this will not change any of the inferred types, which were already the right ones. The difference is in the type returned by `Typedtree.{pat,let}_bound_idents_full`, and as far as I can tell this type is only used to generate the value kinds for `Lstaticcatch` handlers. So the differences only appear starting from Lambda.

lpw25:
@lthls Would you mind pushing your test from that PR to this PR? You said upstream that it was broken with both closure and flambda1, which convinces me that it is reasonably likely to continue to test the right thing as the code base evolves.

lthls:
Done

goldfirere (approved):

PR #3509, opened 2025-01-24
Normalize and deduplicate with-bounds in jkinds

glittershark:
This commit contains (somewhat unfortunately) two distinct, but tightly-related changes:

1. The internals of with-bounds have been changed from a list to a map (from types to sets of axes), with a "best-effort" semantic, structural comparison on type_exprs that falls back to pointer comparison on types which are "hard" to compare semantically
2. Full, recursive normalization of with-bounds on kinds is implemented, respecting quality.

Unfortunately, due to some issues I ran into with stack overflows when compiling large mutually recursive type declarations (such as the ASTs in the compiler itself during bootstrap), these ended up as the same commit which it's somewhat challenging to tease apart. Sorry!

With-bound deduplication
------------------------

For the sake of efficiency, with-bounds have been changed to be a (maybe non-empty) Map of type_expr, rather than a (maybe non-empty) list. This both deduplicates them (mostly), which has anywhere from a reasonable to a pretty significant impact on performance of things that need to deal with with-bounds (especially for deeply + widely mutually recursive type declarations, like ASTs), and also makes concatenation faster.

Comparing types semantically (eg, in a way that's compatible with Ctype.type_equal) is /hard/, though - especially in the context of things like polymorphic variants and objects. We might someday write an algorithm to do so, but today this feels like a bit of a monstrous undertaking, and we want to get something out and working. So instead, we do a hybrid approach: for all the types which /aren't/ hard to compare (type constructors, type variables, tuples, etc. - and trees of such) we compare structurally using a recursive lexicographic tree comparison. For all the remaining "hard" types, we compare them using pointer equality - but make all "hard" types compare strictly less than the "easy" ones. This gives us some false positives - we might compare two types as inequal which are actually semantically equal - but no false negatives, and also gets us a lot further towards the world we want for the most commonly used types.

Unfortunately, actually implementing this is a little tricky, because the Map in stdlib is a functor and the concrete type is not actually exposed, and with-bounds are mutually recursive (through jkind annotations on type vars) with type_exprs. We have, as I see it, three options:

1. Two mutually recursive modules, one containing type_expr and one containing the functor invocation for the map of types in with_bounds.
2. Define the map as actually just a map of Obj.t, with the comparator behind a ref that's map after the recursive type declaration for type_expr and with_bounds, and then define another module providing a safe interface
3. Define our own Map type that exposes a concrete type and accepts a comparator.

Option 1 is probably the nicest (and likely the only one to be accepted upstream!) but also involves a *lot* of changes and churn to types.ml, and seems like it'd make it really hard to pull in changes from upstream in the meantime due to this churn. Option 3 is also appealing, but a lot of work and finnicky to get right. So, this commit implements option 2 (with some regret about the Obj) since it felt like the best point in the tradeoff space

With-bound normalization
------------------------

This commit also implements normalization of with-bounds, adapting and replacing the existing `extend_bound` function to operate holistically on the whole kind rather than per-bound, and (optionally!) respect the best-ness of the kinds it looks up for types in with bounds. This is more or less obvious (I hope!) with a couple of notable caveats:

1. Different callers of normalize want different behavior when not-best kinds are encountered; some want to skip them and leave them in the with-bounds, and some want to omit them from the with-bounds but round up the bounds along the axes that type matters for to the max. Currently, this behavior is exposed via a `~require_best:bool` parameter, which is gross (booleans are sad) and really easy to get wrong. It would be preferable to have this parameter be a GADT with a phantom `l * r` parameter which controlled whether we're expected to return an l- or an lr-jkind, but doing this correctly (for... somewhat nuanced reasons) requires correct subsumption, so that's *currently* deferred until a follow-up pull request that also adds subsumption.

2. For reasons I don't yet entirely understand, I had to change the behavior with respect to fuel reuse when normalizing multiple fields of the same type, or else I got a stack overflow error when compiling `parsetree.mli` during bootstrapping. This is a little sad, but I spent multiple days trying to find a better approach and failed, and I'd like to push this forward and revisit the fuel usage within normalize later.


goldfirere:
Was it necessary to force-push here? This has made #3525 harder to review. I think it would be best to avoid force-pushing to any of these PRs until we're ready to start merging, unless I'm missing an aspect of the workflow.

glittershark:
@goldfirere ack, sorry, will stop force-pushing starting now

typing/jkind.ml
```diff
@@ -1417,37 +1363,188 @@ module Jkind_desc = struct
...
+                join_respecting_omit b_jkind.jkind.mod_bounds
+              in
+              (* Descend into the with-bounds of each of our with-bounds types'
+                  with-bounds *)
+              let bounds_so_far, nested_with_bounds =
+                loop ctl_after_unpacking_b bounds_so_far
+                  (With_bounds.to_list b_jkind.jkind.with_bounds)
+              in
+              (* Use *original* ctl here, so we don't fall over on
+                  a record with 20 lists with different payloads. *)
```
liam923, line 1504: this comment is stale

typing/types.ml
```diff
@@ -996,19 +1019,135 @@ end
...
+      | Tvar _
+      | Tobject (_, _)
+      | Tfield (_, _, _, _)
+      | Tnil
+      | Tlink _
+      | Tsubst (_, _)
+      | Tvariant _
+      | Tpackage (_, _)
+      | Tarrow (_, _, _, _)
+        -> -ty.id
```
goldfirere, line 1049: Why the negation? Well, I guess I can see why. But there should be a pretty loud comment about that one important character.

typing/types.ml
```diff
@@ -996,19 +1019,135 @@ end
...
+  else
+    let rank ty =
+      match ty.desc with
+      (* Types which must be compared by id *)
+      | Tvar _
+      | Tobject (_, _)
+      | Tfield (_, _, _, _)
+      | Tnil
+      | Tlink _
+      | Tsubst (_, _)
```
goldfirere, line 1045: These two type forms (`Tlink` and `Tsubst`) are used only transiently, in unification and substitution, respectively. (Well, `Tlink` isn't as transient, but we tend to call `repr` to avoid seeing it.) I think that, for this algorithm, you want to fatal-error on these forms.

typing/types.ml
```diff
@@ -996,19 +1019,135 @@ end
...
+      | Tvariant _
+      | Tpackage (_, _)
+      | Tarrow (_, _, _, _)
```
goldfirere, line 1048: `Tarrow` seems easy enough (and prevalent enough) to do a more proper treatment for -- unless there's a detail making it harder than it looks.

typing/types.ml
```diff
@@ -996,19 +1019,135 @@ end
...
+           if l = 0 then best_effort_compare_type_expr te1 te2 else l
+        )
+        elts1
+        elts2
+    | Tconstr (p1, args1, _), Tconstr (p2, args2, _) ->
+      let p = Path.compare p1 p2 in
+      if p = 0
+      then List.compare best_effort_compare_type_expr args1 args2
+      else p
+    | Tpoly (t1, ts1), Tpoly (t2, ts2) ->
```
goldfirere, line 1072: This case is hard, for a variety of reasons.

- The `ts` are binding sites for variables that may occur in the `t`s. If an in-scope variable is mentioned in a `ts`, it must be shadowed.
- The `ts` are set-like, in that order doesn't matter.
- Even worse, if a `ts` binds a variable that is then unused in a `t`, the `Tpoly` has no effect.
- If the `ts` is empty (which happens plenty of times in practice), then the `Tpoly` has no effect.

Maybe I'm missing some, too. All that said, given the "best effort" nature, maybe this is correct? At the least, there should be a loud comment saying this is mostly broken.

typing/types.ml
```diff
@@ -996,19 +1019,135 @@ end
...
+      | Tlink _
+      | Tsubst (_, _)
+      | Tvariant _
+      | Tpackage (_, _)
+      | Tarrow (_, _, _, _)
+        -> -ty.id
+      (* Types which we know how to compare structurally*)
+      | Ttuple _ -> 2
+      | Tunboxed_tuple _ -> 3
+      | Tunivar _ -> 4
```
goldfirere, line 1053: I'm surprised to see this here. Why isn't it like `Tvar`?

typing/types.ml
```diff
@@ -996,19 +1019,135 @@ end
...
+      : (type_expr -> info option -> info option -> info option) ->
+      t -> t -> t
+    val to_seq : t -> (type_expr * info) Seq.t
+    val map : (info -> info) -> t -> t
+  end
+end = struct
+  module U = With_bounds_type_map_unsafe
+  include U
+
+  let to_seq t = to_seq t |> Seq.map (fun (ty, ti) -> ((Obj.obj ty : type_expr), ti))
```
goldfirere, line 1117: I would abstract out the call to `Obj.obj` into a function with a type signature. I see it's only used twice, but it would be very easy to make a disastrous mistake around this.

goldfirere (commented):
Reviewed the tests and types.ml{,i}. Stopping as I'm properly more addled by illness than I should be for this.

I'm not thrilled with the use of `Obj` here, though I understand the motivation and don't see a better way. It would be great to get at least one more person's input on this choice of implementation strategy. In the end, I'm supportive of it, but only because the other options are worse.

Actually, wait, I think I have an idea.

What if the `nonempty_with_bounds_types` (the only type that needs to be used in the recursive blob) is just declared abstract? Then, later, when you can call `Map.Make` without issue, call it, and then have one spot (well, two functions) that cast between the real map type and the utterly-abstract `nonempty_with_bounds_types`? This approach is still unsafe (it will require `Obj.magic` to do that cast), but the approach has one footgun on a plinth -- and with obvious type signatures -- instead of inhabiting a room decorated with several footguns. Maybe I'm wrong somehow and this approach doesn't work. And even if I'm right, I'd still want a concurring opinion from someone else that it's a good idea.

testsuite/tests/shape-index/index_constrs_records.reference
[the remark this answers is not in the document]
glittershark: I am pretty sure that this is caused by https://github.com/ocaml-flambda/flambda-backend/pull/3509/commits/aa5de35b60e4436e0fdbca42dc7dcc36ad366bed, since that builds a bigger (intermediate) env.

typing/types.ml
```diff
@@ -114,21 +143,15 @@ and jkind_history =
...
-  | Irrelevant_for_nullability
-
-and with_bounds_type =
-  { type_expr : type_expr;
-    modality : Mode.Modality.Value.Const.t;
-    relevant_for_nullability: relevant_for_nullability
-  }
+and with_bounds_types =
+  With_bounds_type_info.t With_bounds_type_map_unsafe.t
+and nonempty_with_bounds_types = with_bounds_types
```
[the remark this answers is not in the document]
glittershark: yeah, I originally wrote this without the non-emptiness tracked at all, then added it back in. I kinda felt like the whole "you can only have with-bounds on the right" thing seemed vaguely worth tracking (though you'd know more about the particular ways that can go wrong than me, since you introduced this type originally), and without any nonempty guarantee here that's pretty easy to get wrong. Could happily go either way here.

typing/types.mli
```diff
@@ -308,6 +308,41 @@ and jkind_r = (disallowed * allowed) jkind  (* the jkind expected of a type *)
 and jkind_lr = (allowed * allowed) jkind    (* the jkind of a variable *)
 and jkind_packed = Pack_jkind : ('l * 'r) jkind -> jkind_packed
 
+(** This module provides the interface to construct, query, and destruct
```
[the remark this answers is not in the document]
glittershark: fleshed out the comment a bit.

typing/types.ml
glittershark: `commutable` seemed somewhat intimidating to get correct. Do you think it's easier than I'm expecting?

typing/types.ml
glittershark: oh, oops, indeed it should be.

typing/types.ml
glittershark: eep - making these fatal-error made us fail to build stdlib.

glittershark: I guess we're not calling `repr` somewhere we should be.

typing/types.ml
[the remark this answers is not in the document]
goldfirere: I don't think the nonempty guarantee helps there. (First off, I'm assuming you meant that with-bounds go on the left.) Because we don't have a guarantee that left-kinds have nonempty with-bounds -- just that right-kinds have empty ones. But we get that last property just by having the indexing on `with_bounds`, no nonemptiness required.

I introduced the nonemptiness check just to avoid the possibility of multiple equal representations.

Still not sure what's best -- but I don't yet buy the argument in this little thread.

ccasin:
> I'm not thrilled with the use of Obj here, though I understand the motivation and don't see a better way. It would be great to get at least one more person's input on this choice of implementation strategy. In the end, I'm supportive of it, but only because the other options are worse.

When I discussed this with @glittershark last week, I did not see a better alternative.  I haven't read the code in detail or thought hard about your tweak (which sounds fine).

typing/types.mli
```diff
@@ -408,6 +369,45 @@ end
...
+    distinct keys in the map - but two semantically {i inequal} types are guaranteed never
+    to occupy the same key.
+*)
+module With_bounds_types : sig
+  type info := With_bounds_type_info.t
+  type t := with_bounds_types
+
+  val empty : t
+  val is_empty : t -> bool
+  val to_seq : t -> (transient_expr * info) Seq.t
```
goldfirere, line 386: But `transient_expr` is meant to be transient. More precisely, a `transient_expr` becomes invalid once any mutation happens. `with_bounds` exist for a long time, and so they shouldn't hold `transient_expr`s.

Instead, I think using `type_expr` here was right, but we just need to `repr` more aggressively. This is probably best done by using `get_desc`, as `eqtype` does.

goldfirere:
I have reviewed the recent commits (though not really the last one, because I think the one before it was probably not right).

typing/mode_intf.mli
```diff
@@ -541,6 +541,10 @@ module type S = sig
             output list exactly once. *)
         val to_list : t -> atom list
 
+        (** Builds up a modality from a list of [atom], by composing each atom with
+            identity. *)
```
goldfirere, line 545:
```suggestion
            identity. The modalities are applied left to right. *)
```

Check me on this, but I think that's what the `fold_left` will do. Obviously different axes commute, but there's nothing stopping us from having multiple `atom`s on one axis (which won't necessarily commute).

typing/jkind_axis.ml
```diff
@@ -453,3 +437,47 @@ module Axis_collection = struct
     end
   end
 end
+
+module Axis_set = struct
+  (* each axis is true or false to indicate membership  *)
+  type t = bool Axis_collection.t
+
+  (* TODO: this could be represented with a unit8 since there's only 7 possible members *)
```
goldfirere, line 445:
```suggestion
  (* TODO: this could be represented with a uint8 since there's only 7 possible members *)
```

typing/typedecl.ml
```diff
@@ -1225,9 +1225,18 @@ let check_constraints env sdecl (_, decl) =
...
-  | None -> decl
-  | Some ty ->
+  match decl.type_manifest, decl.type_kind with
+  | None, _ -> decl
+  | Some _, (Type_record _ | Type_record_unboxed_product _ | Type_variant _ | Type_open)
+    when not (Builtin_attributes.has_or_null_reexport decl.type_attributes)
+    ->
+    (* If there's both a manifest and a kind, there's no reason to check that the jkind
+       of the manifest matches the annotation. This is because the manifest's jkind is
+       exactly the kind's jkind, which has already been checked against the annotation.
```
goldfirere, line 1235:
```suggestion
       exactly the kind's jkind, which has already been checked against the annotation in [update_decl_jkind].
```

(I think)

typing/typedecl.ml
```diff
@@ -1225,9 +1225,18 @@ let check_constraints env sdecl (_, decl) =
...
+       The annotation has also been narrowed based on the kind's jkind, so there's no
```
goldfirere, line 1236: I'm not sure what "annotation has been narrowed" means. I think you're trying to say that the `type_jkind` field of `type_declaration` has been filled in with the kind's jkind, which is good. But the annotation is part of the parsetree and is not in play here, I think.

typing/typedecl.ml
```diff
@@ -1225,9 +1225,18 @@ let check_constraints env sdecl (_, decl) =
    immediate, we should check the manifest is immediate). Also, update the
    resulting jkind to match the manifest. *)
 let narrow_to_manifest_jkind env loc decl =
-  match decl.type_manifest with
-  | None -> decl
-  | Some ty ->
+  match decl.type_manifest, decl.type_kind with
+  | None, _ -> decl
+  | Some _, (Type_record _ | Type_record_unboxed_product _ | Type_variant _ | Type_open)
+    when not (Builtin_attributes.has_or_null_reexport decl.type_attributes)
```
goldfirere, line 1231: This condition should be remarked on in the (helpful!) comment.

typing/typedecl.ml
```diff
@@ -1850,6 +1854,7 @@ let update_decl_jkind env dpath decl =
     | Type_record (lbls, rep) ->
       let lbls, rep, type_jkind = update_record_kind decl.type_loc lbls rep in
       let type_jkind, type_has_illegal_crossings = add_crossings type_jkind in
+      let type_jkind = Jkind.mark_best type_jkind in
```
goldfirere, line 1857: I think there needs to be a `Note [Quality of jkinds during inference]` explaining the slightly-delicate dance going on in this file. It actually doesn't need to be a long Note -- not_best during inference, best for concrete types after inference -- but the choices here are subtle and the code is distributed. Having a Note will make it much easier for us to understand what's going on in 3 years.

typing/typedecl.ml
```diff
@@ -1896,15 +1902,21 @@ let update_decl_jkind env dpath decl =
...
       type_jkind
   in
 
+
   (* check that the jkind computed from the kind matches the jkind
      annotation, which was stored in decl.type_jkind *)
   if new_jkind != decl.type_jkind then begin
+    (* Add the decl to a temporary env, so that normalization can look up the kind of the
+       type itself. *)
+    let env = add_type ~check:true ~shape id new_decl env in
```
goldfirere, line 1918: This seems pretty fragile, in that it will work only for simple recursion, not mutual recursion. Is this important for correctness? If so, then I fear we're not correct (because of the failure to account for mutual recursion). Or is this just an optimization to improve normalization? Then there should be a comment explaining this.

typing/typedecl.ml
```diff
@@ -1934,10 +1946,12 @@ let update_decls_jkind_reason env decls =
     )
     decls
 
-let update_decls_jkind env decls =
-  List.map
-    (fun (id, decl) -> (id, update_decl_jkind env (Pident id) decl))
+let update_decls_jkind env shapes decls =
+  List.map2
+    (fun (id, decl) shape ->
+       (id, update_decl_jkind env (Pident id) id decl shape))
```
goldfirere, line 1952: Probably don't need to pass both `Pident id` and `id`.

typing/typedecl.ml
```diff
@@ -2353,6 +2367,41 @@ let add_types_to_env decls shapes env =
       add_type ~check:true ~shape id decl env)
     decls shapes env
 
+let mark_decl_jkind_best decl =
+  match decl.type_kind with
+  | Type_abstract _ | Type_open -> decl
```
goldfirere, line 2372: Why not for `Type_open`?

typing/typedecl.ml
```diff
@@ -2353,6 +2367,41 @@ let add_types_to_env decls shapes env =
...
+  | Type_record (_, _) | Type_record_unboxed_product (_, _) | Type_variant (_, _) ->
+    { decl with type_jkind = Jkind.mark_best decl.type_jkind }
+
+(* Normalize the jkinds in a list of (potentially mutually recursive) type declarations *)
+let normalize_decl_jkinds env shapes decls =
+  (* Mark the jkinds in each decl as best (as we have now added all the information we
+     need to to mark them best) *)
+  let decls = List.map (fun (id, decl) -> (id, mark_decl_jkind_best decl)) decls in
```
goldfirere, line 2380: Is this step necessary? I think all the best markers should already be here, from `update_decls_jkind`.

typing/typedecl.ml
```diff
@@ -2353,6 +2367,41 @@ let add_types_to_env decls shapes env =
...
+           Jkind.normalize
+             ~require_best:true
+             ~jkind_of_type:(fun ty -> Some (Ctype.type_jkind env ty))
+             decl.type_jkind
+         in
+         let decl = { decl with type_jkind = normalized_jkind } in
+         (* Add the decl with the normalized kind back to the environment, so that later
+            kinds don't have to normalize this kind if they mention this type in their
+            with-bounds *)
+         let env = add_type ~check:false ~shape:shape id decl env in
```
goldfirere, line 2397: Similar to my comment above, this approach is not robust in the face of mutual recursion. That is, changing the order of declarations within a mutually recursive group will change the normalization behavior. Maybe that's OK? But at least it would be good to have a comment saying so. And if it affects compiler behavior (maybe requiring `module type of`?), there should be a test demonstrating the order-dependence.

typing/typedecl.ml
```diff
@@ -2353,6 +2367,41 @@ let add_types_to_env decls shapes env =
...
+(* Normalize the jkinds in a list of (potentially mutually recursive) type declarations *)
+let normalize_decl_jkinds env shapes decls =
+  (* Mark the jkinds in each decl as best (as we have now added all the information we
+     need to to mark them best) *)
+  let decls = List.map (fun (id, decl) -> (id, mark_decl_jkind_best decl)) decls in
+  (* Add the types, with non-normalized kinds, to the environment to start, so that eg
+     types can look up their own (potentially non-normalized) kinds *)
+  let env = add_types_to_env decls shapes env in
+  let (decls, env) =
+    List.fold_left2
```
goldfirere, line 2385: Might just be worth defining a `Misc.Stdlib.List.fold_left_map2` function to separate out thinking about iteration from thinking about what's actually going on.

goldfirere: Or you could use `fold_left_map` and `combine` pretty readily.

typing/typedecl.ml
```diff
@@ -2353,6 +2367,41 @@ let add_types_to_env decls shapes env =
...
+         (* Add the decl with the normalized kind back to the environment, so that later
+            kinds don't have to normalize this kind if they mention this type in their
+            with-bounds *)
+         let env = add_type ~check:false ~shape:shape id decl env in
+         ((id, decl) :: decls), env)
+      ([], env)
+      decls
+      shapes
+  in
+  List.rev decls, env
```
goldfirere, line 2403: If you use `fold_left_map{,2}`, remember to drop this `List.rev`.

goldfirere (commented):
Reviewed everything except for jkind.ml.

goldfirere:
I've made a small simplification.

* [ ] @glittershark please review [995992fda552d6e0a921675d9eb86754392f7bc1](https://github.com/ocaml-flambda/flambda-backend/pull/3509/commits/995992fda552d6e0a921675d9eb86754392f7bc1)

typing/types.mli
```diff
@@ -71,6 +71,12 @@ val is_mutable : mutability -> bool
 module Jkind_bounds :
```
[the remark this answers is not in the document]
glittershark: Renamed in https://github.com/ocaml-flambda/flambda-backend/pull/3503, which is where this was introduced.

typing/types.mli
glittershark: ah, yeah, good point. Updated to just call `get_desc` in the comparator.

lpw25:
Re the map issue, I’m surprised there is any actual mutual recursion between types and kinds. I would expect types to depend on right kinds and left kinds to depend on type, but I wouldn’t expect right kinds to depend on types or types to depend on left kinds.

goldfirere:
I think Leo's comment is likely accurate, and might provide a way out. But from where I stand, that way out looks like it requires swimming through a flooded cave, in that it would require some significant restructuring in how we think about left- vs right-jkinds. I propose that we not pursue that now, unless someone spots an above-ground route to Leo's proposed destination. If we were starting from scratch, I would strongly consider Leo's idea.

goldfirere:
I've reviewed the recent commits (up through "Fix modality printing bug"). All look good.

typing/jkind.ml
```diff
@@ -424,35 +404,54 @@ module With_bounds = struct
   let map_type_expr (type l r) f : (l * r) t -> (l * r) t = function
     | No_with_bounds -> No_with_bounds
     | With_bounds tys ->
-      With_bounds (Misc.Nonempty_list.map (Type_info.map_type_expr f) tys)
+      With_bounds
+        (tys |> With_bounds_types.Non_empty.to_seq
+        |> Seq.map (fun (ty, ti) -> f ty, ti)
+        |> List.of_seq |> With_bounds_types.of_list
+        |> With_bounds_types.Non_empty.of_maybe_empty |> Option.get)
```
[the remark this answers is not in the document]
glittershark: yeah, I'm convinced. Removed `with_bounds_types_nonempty`.

typing/types.ml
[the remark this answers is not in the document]
glittershark: yeah, I'm convinced. Removed `with_bounds_types_nonempty`.

typing/typedecl.ml
glittershark: yep, it should be best.

typing/jkind.ml
```diff
@@ -1845,6 +1936,11 @@ let for_object =
 (******************************)
 (* elimination and defaulting *)
 
+let[@inline] normalize ~require_best ~jkind_of_type t =
+  { (disallow_right t) with
+    jkind = Jkind_desc.normalize ~require_best ~jkind_of_type t.jkind
```
goldfirere, line 1939: If `not require_best`, then the updated jkind should be `not_best` I think

typing/jkind.ml
```diff
@@ -1869,26 +1965,20 @@ let get_layout jk : Layout.Const.t option = Layout.get_const jk.jkind.layout
...
-    upper_bound_for_axis ~type_equal ~jkind_of_type ~axis jk
-  in
-  { areality = upper_bound_for_axis (Modal (Comonadic Areality));
-    linearity = upper_bound_for_axis (Modal (Comonadic Linearity));
-    uniqueness = upper_bound_for_axis (Modal (Monadic Uniqueness));
-    portability = upper_bound_for_axis (Modal (Comonadic Portability));
-    contention = upper_bound_for_axis (Modal (Monadic Contention))
+let get_modal_upper_bounds (type l r) ~jkind_of_type (jk : (l * r) jkind) :
+    Alloc.Const.t =
+  let jk = normalize ~require_best:false ~jkind_of_type jk in
```
goldfirere, line 1968: I think soundness here requires that the result of `normalize` has no with-bounds. Should we assert that? Or maybe `normalize ~require_best:false` returns an lr-kind?

typing/jkind.ml
```diff
@@ -1869,26 +1965,20 @@ let get_layout jk : Layout.Const.t option = Layout.get_const jk.jkind.layout
...
+    uniqueness = get (Modal (Monadic Uniqueness));
+    portability = get (Modal (Comonadic Portability));
+    contention = get (Modal (Monadic Contention))
   }
 
-let get_externality_upper_bound ~type_equal ~jkind_of_type jk =
-  upper_bound_for_axis ~type_equal ~jkind_of_type ~axis:(Nonmodal Externality)
-    jk
+let get_externality_upper_bound ~jkind_of_type jk =
+  let jk = normalize ~require_best:false ~jkind_of_type jk in
```
goldfirere, line 1978: Ditto comment from `get_modal_upper_bounds`

typing/jkind.ml
```diff
@@ -2531,34 +2618,46 @@ let sub_jkind_l ~type_equal ~jkind_of_type sub super =
   let layouts =
     Misc.Le_result.is_le (Layout.sub sub.jkind.layout super.jkind.layout)
   in
+  let sub = normalize ~require_best:true ~jkind_of_type sub in
+  (* CR aspsmith: this weird double-normalization is poorly explained because it will go
```
goldfirere, line 2620: OK. I'm not reviewing this function.

testsuite/tests/typing-jkind-bounds/basics.ml
```diff
@@ -1631,3 +1631,18 @@ Error: This expression has type "int t" but an expression was expected of type
...
+(*********************************)
+(* Test 15: extensible variants *)
+
+(* The best kind an extensible variant can get is [value] *)
+type extensible : value = ..
+[%%expect{|
+type extensible = ..
+|}]
+
+(* Since the kind is [best], it should normalize away *)
```
goldfirere, line 1644: This doesn't test whether `extensible` normalizes away. I think you'd need something like

```
module M : sig
  type t : immediate with extensible
end = struct
  type t
end
```

typing/jkind.ml
glittershark: In the next PR (https://github.com/ocaml-flambda/flambda-backend/pull/3525), @liam923 refactors `normalize` to return an lr-kind in the not_best case. Plausibly it's still worth adding the assert, but I think it should happen there if so?

typing/jkind.ml
glittershark: ditto response :)

testsuite/tests/typing-jkind-bounds/basics.ml
glittershark: well, it does in this branch since there's no subsumption :p 

but sure, that's a more robust test anyway

typing/jkind.ml
glittershark: I'd like @liam923's take on this, since he shifts this stuff around a bit in the subsequent PR

typing/jkind.ml
liam923: In my PR, it using the `normalize_mode` gadt to guarantee this via types, and then it matches the result's with-bounds against `No_with_bounds` to assert this in the type system. But now that `With_bounds`'s map can be empty, this type assertion isn't as powerful as it used to be.

typing/jkind.ml
liam923: My pr makes it so that if it runs out of fuel then it gets `Not_best`. But it doesn't do what Richard suggests, and I think that'd be a good change?

testsuite/tests/parsetree/source_jane_street.ml
```diff
@@ -1068,8 +1068,8 @@ type 'a contended_with_int : immutable_data with 'a @@ contended with int
 [%%expect{|
 type 'a list : immutable_data with 'a
 type ('a, 'b) either : immutable_data with 'a * 'b
-type 'a contended : immutable_data with 'a @@ contended
-type 'a contended_with_int : immutable_data with 'a @@ contended with int
+type 'a contended : immutable_data with 'a
```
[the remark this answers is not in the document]
glittershark: it was, and @liam923 fixed it

typing/jkind.ml
glittershark: > now that `With_bounds`'s map can be empty

it can representationally... but it should never actually be empty

typing/typedecl.ml
glittershark: @liam923 wrote this - can you clarify, Liam?

typing/jkind.ml
```diff
@@ -1417,37 +1358,188 @@ module Jkind_desc = struct
...
-               if Axis_ops.le Axis_ops.max bound2
-               then Less
-               else
-                 let bound1_extended =
-                   With_bounds.extend_bound ~type_equal ~jkind_of_type ~axis
-                     ~bound:bound1 wbs
-                 in
-                 Axis_ops.less_or_equal bound1_extended bound2)
-      |> Misc.Le_result.combine_list
+    let { layout = lay1; mod_bounds = bounds1; with_bounds = _ } =
```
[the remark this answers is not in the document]
glittershark: this is in fact unsound here, but is fixed in @liam923's PR and difficult to fix here with the wrong subsumption check, so I'm resolving per our earlier discussion

typing/jkind.ml
```diff
@@ -1417,37 +1358,188 @@ module Jkind_desc = struct
...
+      normalize ~require_best:false ~jkind_of_type sub
```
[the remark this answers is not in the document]
glittershark: almost definitely, but I think this is probably sufficiently addressed in @liam923's PR that I'd like to not think about it here (since at least the tests are passing)

typing/jkind.ml
glittershark: Do you want to make that change on your branch, or should I?

typing/typedecl.ml
```diff
@@ -2367,31 +2375,43 @@ let add_types_to_env decls shapes env =
       add_type ~check:true ~shape id decl env)
     decls shapes env
 
+
+(* Note [Quality of jkinds during inference]
```
goldfirere, line 2379: This is great. Please refer to this Note (by name) from the calls to `mark_best`.

goldfirere (commented):
Reviewed through "Don't pass Pident id to update_decl_jkind" (and then one more commit whose message is not a good identifier)

goldfirere (approved):

typing/typedecl.ml
glittershark: indeed. the actual correct thing to do is a little more involved, but seems to work fine.

testsuite/tests/typing-jkind-bounds/composite.ml
```diff
@@ -520,13 +570,7 @@ Error: This value is "contended" but expected to be "uncontended".
...
 (* CR layouts v2.8: This should work once we get proper subsumption. *)
 [%%expect {|
-Line 1, characters 0-52:
-1 | type 'a t : immutable_data = Flat | Nested of 'a t t
-    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-Error: The kind of type "t" is value
-         because it's a boxed variant type.
-       But the kind of type "t" must be a subkind of immutable_data
-         because of the annotation on the declaration of the type t.
+type 'a t = Flat | Nested of 'a t t
```
liam923, line 573: Is this in-progress work? This test seems to be failing now (on my branch too) in favor of the old output

testsuite/tests/typing-jkind-bounds/composite.ml
```diff
@@ -563,13 +607,7 @@ Error:
...
 (* CR layouts v2.8: This should work once we get proper subsumption. *)
 [%%expect {|
-Line 1, characters 0-71:
-1 | type ('a : immutable_data) t : immutable_data = Flat | Nested of 'a t t
-    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-Error: The kind of type "t" is value
-         because it's a boxed variant type.
-       But the kind of type "t" must be a subkind of immutable_data
-         because of the annotation on the declaration of the type t.
+type ('a : immutable_data) t = Flat | Nested of 'a t t
```
liam923, line 610: Same deal with this test

testsuite/tests/typing-jkind-bounds/composite.ml
```diff
@@ -321,7 +365,13 @@ Error: This value is "aliased" but expected to be "unique".
 (***********************************************************************)
 type 'a t : immutable_data with 'a = { head : 'a; tail : 'a t option }
 [%%expect {|
-type 'a t = { head : 'a; tail : 'a t option; }
+Line 1, characters 0-70:
```
liam923, line 368: This one is failing on my branch in the happy way (being accepted)

typing/typedecl.ml
liam923: done

typing/typedecl.ml
liam923: I'm going to punt on this. How or_null is handle is going to have to change after the rebase, so this code is going to change

typing/jkind.ml
liam923: I've just made this change on this branch and pushed

typing/typedecl.ml
```diff
@@ -1955,11 +1937,51 @@ let update_decls_jkind_reason env decls =
...
 let update_decls_jkind env shapes decls =
-  List.map2
-    (fun (id, decl) shape ->
-       (id, update_decl_jkind env id decl shape))
+  let any_annots_need_checking, decls =
+    List.fold_left_map
+      (fun any_annots_need_checking (id, decl) ->
+         let new_decl = update_decl_jkind env decl in
+         let annot_to_check =
+           if new_decl.type_jkind != decl.type_jkind
```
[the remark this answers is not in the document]
goldfirere: Actually I think it's much simpler to do this check at the very end of `transl_type_decl`. We'll already have normalized (so jkinds are smaller) and we'll already have the right env. The only annoyance is getting the annotation back -- but even that's not too hard; just make sure not to shadow the decls before `update_decls_jkind`.

All that said, I think the check as written here is correct, and so I'm fine with delaying this proposed improvement until later.

goldfirere (commented):
Reviewed two commits from @glittershark, "Tests for jkind inference on composite types with many arrows" and "Check annotations in a separate pass from updating jkinds"

testsuite/tests/typing-jkind-bounds/composite.ml
```diff
@@ -520,13 +570,7 @@ Error: This value is "contended" but expected to be "uncontended".
 type 'a t : immutable_data = Flat | Nested of 'a t t
 (* CR layouts v2.8: This should work once we get proper subsumption. *)
```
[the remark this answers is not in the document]
goldfirere: Hmm.. The failure is back. OK.

testsuite/tests/typing-jkind-bounds/composite.ml
```diff
@@ -563,13 +607,7 @@ Error:
 type ('a : immutable_data) t : immutable_data = Flat | Nested of 'a t t
 (* CR layouts v2.8: This should work once we get proper subsumption. *)
```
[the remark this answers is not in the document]
goldfirere: Never mind; the failure is back.

goldfirere:
Reviewed "Clarify wording in comment in narrow_to_manifest_jkind" through "Fix build error"

typing/types.ml
```diff
@@ -1042,6 +1042,11 @@ let rec best_effort_compare_type_expr te1 te2 =
       | Tvariant _
       | Tpackage (_, _)
       | Tarrow (_, _, _, _)
+          (* NOTE: we can actually see Tsubst here in certain cases, eg during
+             [Ctype.copy] when copying the types inside of with_bounds. We also can't
+             compare Tsubst structurally, because the Tsubsts that are created in
+             Ctype.copy are cyclic (?). So the best we can do here is compare by id. *)
+      | Tsubst (_, _)
```
goldfirere, line 1049: As discussed today, I'm pretty sure this is wrong. (Just leaving this comment here so the wrongness doesn't get lost.)

goldfirere (commented):
Reviewed the last two commits

typing/types.ml
glittershark: added more commentary about the wrongness

typing/types.ml
```diff
@@ -1042,10 +1047,12 @@ let rec best_effort_compare_type_expr te1 te2 =
       | Tvariant _
       | Tpackage (_, _)
       | Tarrow (_, _, _, _)
-          (* NOTE: we can actually see Tsubst here in certain cases, eg during
+          (* CR layouts v3.8: we can actually see Tsubst here in certain cases, eg during
```
goldfirere, line 1050: 2.8

goldfirere (commented):
Reviewed 3 most recent commits.

glittershark:
>What if the nonempty_with_bounds_types (the only type that needs to be used in the recursive blob) is just declared abstract? Then, later, when you can call Map.Make without issue, call it, and then have one spot (well, two functions) that cast between the real map type and the utterly-abstract nonempty_with_bounds_types? This approach is still unsafe (it will require Obj.magic to do that cast), but the approach has one footgun on a plinth -- and with obvious type signatures -- instead of inhabiting a room decorated with several footguns. Maybe I'm wrong somehow and this approach doesn't work. And even if I'm right, I'd still want a concurring opinion from someone else that it's a good idea. 

done.

typing/typedecl.ml
```diff
@@ -1632,16 +1632,22 @@ let update_constructor_representation
...
+    decls shapes env
+
 (* This function updates jkind stored in kinds with more accurate jkinds.
    It is called after the circularity checks and the delayed jkind checks
    have happened, so we can fully compute jkinds of types.
 
-   This function is an important part
-   of correctness, as it also checks that the jkind computed from a kind
-   is consistent (i.e. a subjkind of) any jkind annotation.
-   See Note [Default jkinds in transl_declaration].
```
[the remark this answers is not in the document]
glittershark: I checked when I did this, and I don't think so.

typing/typedecl.ml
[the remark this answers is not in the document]
glittershark: we can do that for the mod+with bounds, but (it turns out) we need to check layouts as early as possible (in update_decls_jkind) in order to fill in sort variables, or else we end up defaulting things to `value` incorrectly.

pushed a commit which fixes the whole situation, but it's kinda sad. also added a paragraph to a Note to explain this all.

typing/typedecl.ml
glittershark: done in the commit that added the note.

typing/jkind.ml
```diff
@@ -1105,6 +1029,31 @@ module Const = struct
...
+                Atom
+                  (axis, Join_with (Mode.Value.Monadic.Const.max_axis monadic))
+              | Comonadic comonadic ->
+                Atom
+                  ( axis,
+                    Meet_with (Mode.Value.Comonadic.Const.min_axis comonadic) )
+            in
+            Modality.Value.Const.compose acc ~then_
+          | Nonmodal _ ->
+            (* TODO: don't know how to print *)
```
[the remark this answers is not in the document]
glittershark: without syntax for it, I'm having a hard time thinking up ways of even constructing a type which crosses everything but externality with some other type. Any suggestions?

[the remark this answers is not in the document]
glittershark: or nullability, come to think of it.

typing/jkind.ml
```diff
@@ -1417,37 +1358,188 @@ module Jkind_desc = struct
...
+          match Path.Map.find_opt p constr with
+          | None ->
+            Continue
+              { t with
+                constr = Path.Map.add p (initial_fuel_per_ty, args) constr
+              }
+          | Some (fuel, seen_args) ->
+            if List.for_all2
+                 (fun ty1 ty2 ->
+                   TransientTypeOps.equal (Transient_expr.repr ty1)
```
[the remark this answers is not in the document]
glittershark: My reasoning here was that we're already trying to deduplicate stuff semantically via the best-effort type map, and so it felt worth trading off a little bit of false-positives / overuse of fuel for better performance on type equality here.

goldfirere:
Today I learned that GitHub sometimes actually reorders my comments with respect to commits. Blargh. So I have to go back to naming the commits.

Reviewed a few recent commits, "Add comment about Tpoly" .. "correct version for layouts" (+ the merge commit directly thereafter)

typing/jkind.ml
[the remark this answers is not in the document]
liam923: I'm not sure about externality. But I think nullability actually almost always hits this case since it isn't deep. So if there was a modality for nullability, this would be a bug. And if there were other deep axes, those would have the same bug. Maybe we should put logic in here to handle this?

typing/jkind.ml
```diff
@@ -1417,37 +1358,188 @@ module Jkind_desc = struct
...
+        in
+        match Loop_control.check ctl ty with
+        | Stop ->
+          (* out of fuel *)
+          join_respecting_omit Mod_bounds.max, With_bounds_types.of_list bs
+        | Skip -> loop ctl bounds_so_far bs (* skip [b] *)
+        | Continue ctl_after_unpacking_b -> (
+          match jkind_of_type ty with
+          | Some b_jkind ->
+            if (match b_jkind.quality with Not_best -> false | Best -> true)
```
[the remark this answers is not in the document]
glittershark: I think we should look at what @liam923's branch does

typing/jkind.ml
```diff
@@ -1417,37 +1358,188 @@ module Jkind_desc = struct
...
+              in
+              (* Use *original* ctl here, so we don't fall over on
+                  a record with 20 lists with different payloads. *)
+              let bounds, bs' = loop ctl_after_unpacking_b bounds_so_far bs in
+              bounds, With_bounds.join_bounds_maybe_empty nested_with_bounds bs'
+            else
+              (* Don't know the best kind for this type, so skip it, leaving it in the
+                 (normalized) with_bounds. *)
+              let bounds_so_far =
+                if require_best
```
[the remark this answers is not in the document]
goldfirere: If `require_best` is false, then `not require_best` is true, and `blah || not require_best` is also true, and we won't be in the else branch. Unless I've missed something.

typing/jkind.ml
```diff
@@ -1529,15 +1529,16 @@ module Jkind_desc = struct
               let bounds_so_far, bs' =
                 loop ctl_after_unpacking_b bounds_so_far bs
               in
-              bounds_so_far, With_bounds.add_bound_maybe_empty ty ti bs'
+              ( bounds_so_far,
+                if require_best then With_bounds.add_bound ty ti bs' else bs' )
```
goldfirere, line 1533: Similarly here. I think `require_best` must be true.

typing/jkind.ml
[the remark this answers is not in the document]
glittershark: I'm going to leave this as-is, since this changes in the subsumption branch anyway

PR #3511, opened 2025-01-24
Support parameterised aliases with `-no-alias-deps`

lukemaurer:
Since aliases don't cause loads under `-no-alias-deps` and loads are what makes the type system aware that things are parameterised and that awareness drives substitution and substitution is what propagates instance arguments to their uses, `-no-alias-deps` aliases cause arguments to be dropped on the floor: for instance, `Foo.Bar.t` might fail and demand the user set `-parameter P` even if it's already set, if `Foo.Bar` is an alias for `Foo__bar` and `Foo__bar` is parameterised.


typing/persistent_env.ml
```diff
@@ -401,23 +401,74 @@ let find_import ~allow_hidden penv ~check modname =
...
+              gn_mentioned_by = first_mentioned_by }
+      | Error Inconsistent ->
           error (Inconsistent_global_name_resolution {
               name = global_name;
               old_global;
-              new_global = global;
+              new_global;
               first_mentioned_by;
               now_mentioned_by = mentioned_by;
             })
```
riaqn, line 439: I'm unsure of this error - in what case would an user error cause this? I didn't find it in the test cases. If this must indicate an internal bug, it should raise instead?

riaqn: As discussed offline, this is now made into an internal error. Resolved.

typing/global_module.ml
```diff
@@ -307,3 +313,111 @@ let has_arguments t =
...
+    | (glob1, Approximate), (glob2, Approximate) ->
+        meet_approximate glob1 glob2
+    | (glob1, Exact), (glob2, Exact) ->
+        begin match equal glob1 glob2 with
+        | true -> Ok t1
+        | false -> Error Inconsistent
+        end
+    | ((exact, Exact) as t_exact), (approx, Approximate)
+    | (approx, Approximate), ((exact, Exact) as t_exact) ->
+        begin match approximates ~exact ~approx with
```
riaqn, line 416: Just for the sake of reusing code, maybe we can `meet_approximate exact approx` and check that the result is equal to `exact`.

Also, maybe this function can simply raise internal error, instead of returning `Error Inconsistent`.

typing/env.ml
```diff
@@ -1255,7 +1254,18 @@ let find_ident_module id env =
   | Mod_unbound _ -> raise Not_found
   | Mod_persistent ->
       match Ident.to_global id with
-      | Some global_name -> find_pers_mod ~allow_hidden:true global_name
+      | Some global_name ->
+          let allow_excess_args =
+            (* This may be a global that arose by substituting instance arguments into
+               an overapproximated instance name, so we have to allow it to have more
+               arguments than expected. *)
```
riaqn, line 1261: An example would be helpful here. In general, throughout this PR there are functions that takes `Global_module.Name.t`, for which `alow_excess_args` doesn't make sense?

lukemaurer: They can indeed: if `foo.ml` says `module M = M` and we have to approximate `M` as `M{P}{Q}` in `foo.cmi`, then if someone needs the type of `Foo[P:Int][Q:Float]`, we'll do substitution on the approximation `M{P}{Q}` and come up with `M[P:Int][Q:Float]`. At that point we _still_ haven't necessarily loaded `m.cmi` so it's possible for one or both of `P` and `Q` to be excess arguments, and it's not in general possible to detect that fact before we call `Persistent_env.find` and such.

lukemaurer: Agreed that an example would be good, though, so I'll add one

riaqn: Thanks that makes sense.

typing/global_module.ml
```diff
@@ -307,3 +308,86 @@ let has_arguments t =
...
+    | Approximate -> Format.fprintf ppf "approx"
+
+  let output = Misc.output_of_print print
+
+  let equal t1 t2 =
+    match t1, t2 with
+    | Exact, Exact
+    | Approximate, Approximate -> true
+    | (Exact | Approximate), _ -> false
+end
```
riaqn, line 326: I wonder if this date type is needed. It seems currently it's only used to trigger internal errors? (Which I'm fine with. Just want to point out that this is not used for anthing that's visible to the users.)

typing/persistent_env.ml
```diff
@@ -467,18 +521,23 @@ let check_for_unset_parameters penv global =
...
+    pn.pn_global
+  in
   match Hashtbl.find penv.globals name with
-  | { gn_global; _ } -> gn_global
-  | exception Not_found ->
-      let pn = find_pers_name ~allow_hidden:true penv check name in
-      pn.pn_global
+  | { gn_global = (global, Exact); _ } -> global
+  | { gn_global = (_, Approximate); _ } -> load ()
+  | exception Not_found -> load ()
```
riaqn, line 534: This seems to be the only place where Exact/Approximate affects behavior other than raising internally. But, does `global_of_global_name` have to return exact global_module? I looked at a few call sites and ain't sure.

testsuite/tests/templates/dunelike/export_fancy_q_impl__.ml
```diff
@@ -0,0 +1,9 @@
+(* Parameters: P *)
+
+module Fancy_q_impl = Fancy(Q)(Q_impl) [@jane.non_erasable.instances]
+
+module No_direct_access_to_export_fancy_q_impl = struct
+  module Export_fancy_q_impl = No_such_module
```
riaqn, line 6: As discussed, would be good to test that using this alias gives error.

testsuite/tests/templates/dunelike/test_byte.reference
```diff
@@ -0,0 +1,5 @@
+Basic(4)
```
riaqn, line 1: As discussed, this can be merged with `test_natvie.reference`.

riaqn (approved):

PR #3512, opened 2025-01-24
Maintain stack alignment in OCaml callbacks

TheNumbat:
In runtime5, invoking OCaml code using `caml_callback` switches to the current fiber's OCaml stack. 
However, `caml_c_call` / `caml_c_call_stack_args` / `caml_call_gc` do not align the stack before saving the OCaml stack pointer and switching to the C stack. 
This results in the callback running on a misaligned OCaml stack, which will crash if it tries to spill a SIMD variable.

Instead, we now have:
1) Newly allocated stacks start with `sp = 8 mod 16` 
2) `caml_c_call` and friends still don't align the stack, so write `sp = 8 mod 16`
3) `caml_start_program` (and therefore `caml_callback`) expects `sp = 8 mod 16` and adds another padding word so that `sp = 0 mod 16` when entering OCaml.  

Note this is not necessary when frame pointers are enabled because `caml_c_call` and friends will align `sp` by pushing rbp. This is also the case on arm64.

### DIFF
--- a/runtime/amd64.S
+++ b/runtime/amd64.S
@@ -1365,7 +1365,7 @@ LBL(frame_runstack):
         leaq    32(%rsp), %r11 /* SP with exn handler popped */
 #if Stack_padding_word
     /* Pop the additional alignment padding word *and* the stack padding word,
-       so that $r11 points to the handler struct. */
+       so that %r11 points to the handler struct. */
         addq    $16, %r11
 #endif
         movq    Handler_value(%r11), %rbx
@@ -1396,7 +1396,7 @@ LBL(fiber_exn_handler):
         leaq    16(%rsp), %r11
 #if Stack_padding_word
     /* Pop the additional alignment padding word *and* the stack padding word,
-       so that $r11 points to the handler struct. */
+       so that %r11 points to the handler struct. */
         addq    $16, %r11
 #endif
         movq    Handler_exception(%r11), %rbx
--- a/runtime/amd64.S
+++ b/runtime/amd64.S
@@ -934,7 +934,7 @@ LBL(108):
         movq    %r11, Caml_state(gc_regs)
         addq    $16, %r10
 #if Stack_padding_word
-    /* Pop the additional alignment padding word. */
+    /* Pop the additional alignment padding word */
         addq    $8, %r10
 #endif
     /* Update alloc ptr */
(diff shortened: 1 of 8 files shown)


stedolan:
This is more complicated than I was expecting it to be and I haven't had time to go through it in detail today. One thing that would make me more confident in its correctness is if you adapt the `CHECK_STACK_ALIGNMENT` macros that upstream used to use (deleted in PR 11239) to trap using the `int3` instruction if the stack is not aligned at the point in `caml_start_program` where it is expected to be.

stedolan:
Also, do you have any idea why this doesn't currently break on arm64 builds? From the comments it looks like they have strict stack alignment requirements even without simd usage, so I'm surprised it works there at the moment.

TheNumbat:
Sure, added the alignment check in `caml_start_program`. I believe the problem doesn't apply to arm because the relevant functions push an even number of words ([e.g.](https://github.com/ocaml-flambda/flambda-backend/blob/main/runtime/arm64.S#L590)) on the stack regardless of frame pointers. Note the SIMD test I wrote only runs on x86, though.

runtime/amd64.S
```diff
@@ -1320,6 +1329,10 @@ CFI_STARTPROC
         movq    %rcx, Handler_parent(%r11)
         movq    %rax, Caml_state(current_stack)
         movq    Stack_sp(%rax), %r11
+#if Stack_padding_word
+    /* Skip stack alignment padding */
```
NickBarnes, line 1333: I found this comment confusing.

NickBarnes (approved):
This looks good to me.
Every time I look at the assembly stubs I shudder at the quantity of explicit constants (which should all be symbolic, computed, and statically checked) but this doesn't add significantly to that.

runtime/amd64.S
```diff
@@ -889,6 +889,10 @@ LBL(caml_start_program):
     /* Load the OCaml stack. */
         movq    Caml_state(current_stack), %r11
         movq    Stack_sp(%r11), %r10
+#if Stack_padding_word
+    /* Pad stack to preserve alignment */
+        subq    $8, %r10
```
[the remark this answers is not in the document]
TheNumbat: Yeah, seems fine for now

runtime/amd64.S
TheNumbat: Clarified

TheNumbat:
Turns out this did break effects (after rebasing on main) because `caml_runstack` was assuming the top of the OCaml stack is the address of its `stack_handler`. It now pops the padding/alignment words first.

I've added tests for SIMD spills during each part of an effect handler, and checked the other uses of the `Handler_*` macros, and I believe they're all OK because they're always directly applied to `stack_info->handler` values.

runtime/amd64.S
```diff
@@ -1353,15 +1353,21 @@ CFI_STARTPROC
...
 #endif
         movq    %rdi, %rax /* first argument */
+        CHECK_STACK_ALIGNMENT;
         callq   *(%rbx) /* closure in %rbx (second argument) */
 LBL(frame_runstack):
         leaq    32(%rsp), %r11 /* SP with exn handler popped */
+#if Stack_padding_word
+    /* Pop the additional alignment padding word *and* the stack padding word,
+       so that $r11 points to the handler struct. */
+        addq    $16, %r11
```
NickBarnes, line 1369: It took me a long time to grasp what was going on here. A diagram would really help!

NickBarnes (approved):
LGTM.

PR #3513, opened 2025-01-25
Unify scalar primitives

jvanburen:
This unifies the scalar primitives, which unlocks the ability to use unboxed primitives. More details forthcoming.


ccasin (commented):
I've started reading typing and lambda, but am going to pause because I think some documentation of the overall design goals and approach here (see my comment that suggests adding such documentation to `scalar.ml`, but possibly that's not the right place for it).

typing/predef.ml
```diff
@@ -95,8 +95,10 @@ and path_float64x2 = Pident ident_float64x2
 let path_unboxed_float = Path.unboxed_version path_float
 and path_unboxed_float32 = Path.unboxed_version path_float32
 and path_unboxed_nativeint = Path.unboxed_version path_nativeint
-and path_unboxed_int8 = Path.unboxed_version path_int8
-and path_unboxed_int16 = Path.unboxed_version path_int16
+and path_naked_char = Path.unboxed_version path_char
+and path_naked_int = Path.unboxed_version path_int
+and path_naked_int8 = Path.unboxed_version path_int8
+and path_naked_int16 = Path.unboxed_version path_int16
```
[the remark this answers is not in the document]
jvanburen: I used "naked" b/c the corresponding value type is not boxed, but rather tagged. The terminology is used in flambda2.

testsuite/tests/typing-layouts-bits8/basics.ml
```diff
@@ -472,7 +472,7 @@ external f10_7 : string -> int8# = "foo" "bar"
 
 external f10_8 : int8 -> int8#  = "foo" "bar" [@@unboxed];;
 [%%expect{|
-external f10_8 : (int8 [@unboxed]) -> int8# = "foo" "bar"
+external f10_8 : (int8 [@untagged]) -> int8# = "foo" "bar"
```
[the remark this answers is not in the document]
jvanburen: it's so that you can write `[@@unboxed]` on a function and get the hash versions of every type on the native-code version.

typing/scalar.mli
```diff
@@ -0,0 +1,323 @@
...
+  val negate : t -> t
+end
+
+module Maybe_naked : sig
+  type ('a, 'b) t =
+    | Value of 'a
+    | Naked of 'b
+end
+
+module type S := sig
```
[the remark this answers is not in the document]
jvanburen: It's just used as a shorthand for within the module. I also don't want to put that in the implementation too when it's just a shorthand for showing they have similar module types.

typing/predef.ml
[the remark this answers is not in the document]
jvanburen: okay fine i'll change it

typing/predef.ml
```diff
@@ -384,7 +388,7 @@ let unrestricted tvar ca_sort =
 
 (* CR layouts: Changes will be needed here as we add support for the built-ins
    to work with non-values, and as we relax the mixed block restriction. *)
-let build_initial_env add_type add_extension empty_env =
+let build_initial_env ~include_naked_small_integers add_type add_extension empty_env =
```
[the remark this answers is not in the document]
jvanburen: Okay, changed. 
Then why does `intial` in env.ml have types guarded on the flags?

lambda/lambda.mli
```diff
@@ -104,6 +102,131 @@ type region_close =
...
+  | Unboxed_float64
+  | Unboxed_float32
+
+and unboxed_integer = Primitive.unboxed_integer =
+  | Unboxed_int64
+  | Unboxed_nativeint
+  | Unboxed_int32
+  | Unboxed_int16
+  | Unboxed_int8
+  | Unboxed_int
```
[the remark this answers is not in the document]
jvanburen: added the comment to primitive.mli

lambda/lambda.ml
```diff
@@ -124,6 +120,135 @@ type region_close =
...
+  | Boxed_int32
+
+and boxed_vector = Primitive.boxed_vector =
+  | Boxed_vec128
+
+and nullable =
+  | Nullable
+  | Non_nullable
+
+and value_kind =
```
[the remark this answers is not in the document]
jvanburen: There had been a good reason, but it went away and I forgot to undo most of it.

typing/scalar.mli
```diff
@@ -0,0 +1,323 @@
...
+
+  type nonrec 'a t = ('a width, any_locality_mode width) Maybe_naked.t
+
+  val all : any_locality_mode t list
+
+  val map : 'a t -> f:('a -> 'b) -> 'b t
+
+  val ignore_locality : 'a t -> any_locality_mode t
+
+  val width : any_locality_mode t -> any_locality_mode width
```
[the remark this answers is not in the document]
jvanburen: no i think this was an old decision that I didn't clean up again

testsuite/tests/translprim/module_coercion.compilers.no-flat.reference
```diff
@@ -2,145 +2,152 @@
...
+        (function {nlocal = 0} prim[addrarray] prim[int] stub : float
+          (array.unsafe_get[addr indexed by int] prim prim))
         (function {nlocal = 0} prim[addrarray] prim[int] prim[float] stub
-          (array.set[addr] prim prim prim))
+          : int (array.set[addr indexed by int] prim prim prim))
         (function {nlocal = 0} prim[addrarray] prim[int] prim[float] stub
-          (array.unsafe_set[addr] prim prim prim))
-        (function {nlocal = 0} prim[float] prim[float] stub
-          ignore assert all zero_alloc : int
-          (compare_floats float prim prim))
```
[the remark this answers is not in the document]
jvanburen: I think this test just doesn't run anymore, and it hadn't been updated in a long time but it got caught in my crossfire...

typing/predef.ml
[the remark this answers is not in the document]
ccasin: Good question.  I think this is basically a combination of historical accident and bad implementation.

The fact that boxed and unboxed versions of things are linked in the environment is a recent changes - before that they were all just separate types, and one way to ensure the user gets an error if they write `float#` and have turned off the extension is to not add this to the environment.  But possibly we should always have done this by looking at the identifier to see if it ends in a hash.

typing/scalar.mli
```diff
@@ -0,0 +1,359 @@
...
+    that we want to expose. The submodules are organized to make it easy for use different
+    subsets of scalars in different places. Some examples:
+
+    - Primitive arguments don't depend on the locality of their arguments, but the results
+    do.
+    - Some primitives only take integers, some take only floats, and Three_way_compare
+    takes any scalar type.
+    - The bytecode compiler wants to easily map unboxed/untagged values to their [value]
+    equivalents
+    - The middle-end wants to easily cast between any integraal values using only certain
```
ccasin, line 33:
```suggestion
    - The middle-end wants to easily cast between any integral values using only certain
```

typing/scalar.mli
```diff
@@ -0,0 +1,359 @@
...
+    primitives.
+*)
+
+type any_locality_mode = Any_locality_mode
+
+module Integer_comparison : sig
```
ccasin, line 39: Encountering this module immediately after reading the above comment was surprising - the comment says `scalar.ml` is about representing types, but the presence of `Integer_comparison` suggests it's also about representing operations on those types. Which I thought was the job of `primitive.ml`.

I guess this file has the representations of the "intrinsics" and primitive has the other primitives?  My instinct is the comment above should explain that distinction, or this stuff go over to `primitive.ml` (or both).

typing/scalar.mli
```diff
@@ -0,0 +1,359 @@
...
+  val swap : t -> t
+
+  val negate : t -> t
+end
+
+module Maybe_naked : sig
+  type ('a, 'b) t =
+    | Value of 'a
+    | Naked of 'b
+end
```
ccasin, line 79: The term "naked" isn't otherwise used in the front-end, so this should get a comment explaining what it means (or be renamed in terms of unboxed/untagged)

typing/scalar.mli
```diff
@@ -0,0 +1,359 @@
...
+
+  type nonrec 'a t = ('a width, any_locality_mode width) Maybe_naked.t
+
+  val all : any_locality_mode t list
+
+  val map : 'a t -> f:('a -> 'b) -> 'b t
+
+  val ignore_locality : 'a t -> any_locality_mode t
+
+  val width : _ t -> any_locality_mode width
```
ccasin, line 92: Is the choice between `'a` and `_` intended to suggest some difference in the way these arguments are used?  If not, probably good to be consistent across the whole file.

typing/scalar.mli
```diff
@@ -0,0 +1,359 @@
...
+  val int32 : any_locality_mode t
+
+  val int64 : any_locality_mode t
+
+  val int : _ t
+
+  val nativeint : any_locality_mode t
+end
+
+module type Integral_constants := sig
```
ccasin, line 115: I don't understand the intent of these module names.

1) It looks like the difference is that one has just the boxed things and the other has boxed and unboxed things, but I'm not sure what that has to do with "width".

2) The front-end generally uses the word "constant" to mean "literal" (see, e.g., how it is used in parsetree and typedtree).  I don't think you're using it in that sense here, and am not sure in what sense these are constants.  I think this module type is meant to represent the built-in types, so maybe something like "Integral_builtin_types" would be a good name? (unsure)

Probably comments about what they represent / how they are used would be good.

ccasin (commented):
More questions about the `Scalar` interface.  Sorry for the very incremental review, though on the bright side I'm optimistic that once I understand the design of Scalar the rest of the review will go quickly.

typing/scalar.mli
```diff
@@ -0,0 +1,359 @@
...
+(** This module defines the scalar types intrinsic to the OCaml compiler, and the
+    primitive operations defined on them.
+
+    A [Scalar.t] represents a particular OCaml type that represents a scalar value. It
+    might be tagged, boxed, or neither. There is also a type parameter for the locality of
+    the scalar value, which represents the location in which that boxed values are
+    allocated.
+
+    The important consideration is for a [Scalar.t] to represent all of the primitives
+    that we want to expose. The submodules are organized to make it easy for use different
```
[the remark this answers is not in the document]
jvanburen: clarified

typing/primitive.mli
```diff
@@ -15,23 +15,35 @@
...
-type unboxed_integer =  Unboxed_int64 | Unboxed_nativeint | Unboxed_int32
+type unboxed_integer =
+  | Unboxed_int64
+  | Unboxed_nativeint
+  | Unboxed_int32
+  | Unboxed_int16
+  | Unboxed_int8
+  | Unboxed_int
+  (** We don't have an [int#] type (at least not yet), but for consistency, it's important
+      that we have unboxed types corresponding to all of our integral types *)
```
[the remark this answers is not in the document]
jvanburen: The Typeopt.array_kind_of_elt was just a bug, good catch.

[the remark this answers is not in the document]
jvanburen: The comment is just saying that we don't really expose support for `int#`, however it's still fully usable. I updated the comment to say this.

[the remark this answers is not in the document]
ccasin: If I understand correctly, you're saying there's a bunch of untested, dead code to support a type we don't expose.  My instinct we should either remove this support from the compiler or expose an `int#` type.  I guess I lean towards the former, as `int#` seems likely to be more of an attractive nuicance than something people really want, but we can get more opinions on that.

typing/scalar.mli
```diff
@@ -0,0 +1,373 @@
...
+
+module Floating : sig
+  module Width : sig
+    type 'mode t =
+      | Float32 of 'mode
+      | Float64 of 'mode
+
+    val map : 'a t -> f:('a -> 'b) -> 'b t
+
+    val to_string : any_locality_mode t -> string
```
goldfirere, line 168: This surprises me. We can't convert a general `t` to a string?

typing/scalar.mli
```diff
@@ -0,0 +1,373 @@
...
+
+    include Float_width_shorthands with type 'a t := 'a t
+  end
+
+  include S with type 'a width := 'a Width.t
+
+  include Float_shorthands with type 'a t := 'a t
+end
+
+module Width : sig
```
goldfirere, line 174: The word "width" in this file isn't working for me. It seems to be some kind of catch-all? I see how there are many widths represented here, but I don't know why this module is called `Width` or why some types are.

typing/scalar.mli
goldfirere, line 1: Reading this file through, I still don't think I really understand what's going on here. It might be helpful if there were a little diagram in the file comment showing the different relationships here. The rough structure looks plausible, but I'd be basically completely stuck if I had to use this interface without looking at other code nearby. A little map at the top and just a few examples would, I think, really unlock this.

typing/scalar.ml
```diff
@@ -0,0 +1,694 @@
...
+
+  let nativeint : _ t = Value Width.nativeint
+
+  let naked_int8 : _ t = Naked Width.int8
+
+  let naked_int16 : _ t = Naked Width.int16
+
+  let naked_int32 : _ t = Naked Width.int32
+
+  let naked_int64 : _ t = Naked Width.int32
```
goldfirere: Oops

goldfirere: Though I'm curious why this didn't come out in a test

typing/scalar.ml
```diff
@@ -0,0 +1,694 @@
...
+          ( width,
+            ( Add | Sub | Mul
+            | Div (Safe | Unsafe)
+            | Mod (Safe | Unsafe)
+            | And | Or | Xor ) ) ->
+        let sort = Integral.sort width in
+        sort, sort, sort
+      | Shift (width, (Lsl | Lsr | Asr), Int) ->
+        let sort = Integral.sort width in
+        sort, Jkind_types.Sort.Const.value, sort
```
goldfirere, line 542: We don't want to allow shifting by naked things? Or is that simply later?

typing/scalar.ml
```diff
@@ -0,0 +1,694 @@
...
+      | Floating (width, (Add | Sub | Mul | Div)) ->
+        let sort = Floating.sort width in
+        sort, sort, sort
+      | Icmp (width, (_ : Integer_comparison.t)) ->
+        let sort = Integral.sort width in
+        sort, sort, Jkind_types.Sort.Const.value
```
goldfirere, line 548: Similarly, I would expect things like this to be able to produce untagged booleans.

typing/typeopt.ml
```diff
@@ -155,6 +155,9 @@ let classify ~classify_product env loc ty sort : _ classification =
   | Tconstr (p, _args, _abbrev) ->
       if Path.same p Predef.path_float then Float
       else if Path.same p Predef.path_lazy_t then Lazy
+      else if Path.same p Predef.path_int
+           || Path.same p Predef.path_int8
+           || Path.same p Predef.path_int16 then Int
```
goldfirere: Are these cases covered by the `is_always_gc_ignorable` check?

typing/primitive.mli
```diff
@@ -21,6 +21,10 @@ type unboxed_integer =
   | Unboxed_int32
   | Unboxed_int16
   | Unboxed_int8
+  | Unboxed_int
+  (** We don't support an [int#] type in the stdlib. However, for consistency, it's
+      important that we have unboxed types corresponding to all of our integral types.
```
goldfirere:
```suggestion
      important that we have naked types corresponding to all of our integral types.
```

?

typing/primitive.ml
```diff
@@ -258,10 +257,10 @@ let print p osig_val_decl =
   let is_unboxed = function
     | _, Same_as_ocaml_repr (Base Value)
     | _, Repr_poly
-    | _, Untagged_immediate -> false
+    | _, Unboxed_integer (Unboxed_int | Unboxed_int8 | Unboxed_int16) -> false
```
goldfirere: The use of `Unboxed_integer` to carry both untagged and unboxed things makes me a little uneasy. I don't know as much about the code around here, so I could be wrong to feel uneasy. But I wanted to flag this choice as something I would want at least @mshinwell or @ccasin to agree with.

typing/primitive.ml
```diff
@@ -386,31 +389,27 @@ let equal_native_repr nr1 nr2 =
   match nr1, nr2 with
   | Repr_poly, Repr_poly -> true
   | Repr_poly, (Unboxed_float _ | Unboxed_integer _
-               | Untagged_immediate | Unboxed_vector _ | Same_as_ocaml_repr _)
```
goldfirere, line 397: Why are these changes OK?

typing/primitive.ml
```diff
@@ -774,6 +773,11 @@ let prim_has_valid_reprs ~loc prim =
         match String.Map.find_opt name stringlike_indexing_primitives with
         | Some reprs -> exactly reprs
         | None ->
+          let module I = Scalar.Intrinsic in
+          match I.With_percent_prefix.of_string name with
```
goldfirere: This call seems like it should replace some previous logic. Where's the logic that should be eliminated?

typing/predef.ml
```diff
@@ -384,7 +388,7 @@ let unrestricted tvar ca_sort =
 
 (* CR layouts: Changes will be needed here as we add support for the built-ins
    to work with non-values, and as we relax the mixed block restriction. *)
-let build_initial_env add_type add_extension empty_env =
+let build_initial_env  add_type add_extension empty_env =
```
goldfirere: oops extra space

typing/predef.ml
```diff
@@ -414,12 +418,15 @@ let build_initial_env add_type add_extension empty_env =
...
+       ~unboxed_jkind:Jkind.Const.Builtin.kind_of_unboxed_int8
   |> add_type ident_exn ~kind:Type_open ~jkind:Jkind.Const.Builtin.value
   |> add_type ident_extension_constructor ~jkind:Jkind.Const.Builtin.immutable_data
   |> add_type ident_float ~jkind:Jkind.Const.Builtin.immutable_data
       ~unboxed_jkind:Jkind.Const.Builtin.kind_of_unboxed_float
   |> add_type ident_floatarray ~jkind:Jkind.Const.Builtin.mutable_data
-  |> add_type ident_int ~jkind:Jkind.Const.Builtin.immediate
+  |> add_type ident_int
+       ~jkind:Jkind.Const.Builtin.immediate
+       ~unboxed_jkind:Jkind.Const.Builtin.kind_of_unboxed_nativeint
```
goldfirere: Is this the attractive nuisance that @ccasin doesn't want around?

testsuite/tests/typing-unique/overwriting_proj_push_down_bug.ml
```diff
@@ -184,7 +184,7 @@ let match_anf_aliased r =
         (non_consts ([0: [(consts ()) (non_consts ([0: *, *]))], *]))]
        (catch
          (let (y/340 =a (field_imm 1 r/338))
-           (if (== y/340 "") (let (*match*/347 =[int] 0) (exit 8 y/340))
+           (if (%eq y/340 "") (let (*match*/347 =[int] 0) (exit 8 y/340))
```
goldfirere: Does this change matter?

testsuite/tests/typing-layouts/hash_types.ml
```diff
@@ -310,19 +310,21 @@ Error: The type constructor "t#" expects 1 argument(s),
 type t = int
 and bad = t#
 [%%expect{|
-Line 2, characters 0-12:
-2 | and bad = t#
-    ^^^^^^^^^^^^
-Error: The type "t" has no unboxed version.
+type t = int
+and bad = t#
```
goldfirere, line 321: This should at least be renamed

testsuite/tests/typing-layouts/hash_types.ml
goldfirere, line 1: I want to take another look here after the question of `int#` is settled

typing/scalar.mli
jvanburen: We don't want to drop the local-ness when converting to a string, and we don't know how to turn an 'a into a string.

typing/scalar.mli
jvanburen: `Width.t` always represents the width of a scalar layout -- it has exactly one constuctor for each bit-width of the corresponding scalar. The type parameter is irrelevant within the `Width` module, that's why it's usually fixed to `any_locality_mode` when the corresponding function is expecting a simple sum type (e.g., `to_string`).

typing/typeopt.ml
jvanburen: yes, you're right, this change can be reverted.

typing/scalar.mli
jvanburen: Sorry, what do you mean by a "map"? 
Examples should be easy to source from elsewhere in the tree.

typing/scalar.ml
jvanburen: it's probably unused, perhaps I can move these shorthands to where they are used instead

testsuite/tests/typing-unique/overwriting_proj_push_down_bug.ml
jvanburen: no, it's a no-op. Just a change in how primitives are printed.

typing/predef.ml
jvanburen: yes

typing/primitive.ml
jvanburen: what do you mean? all 3 of those variants are untagged, not unboxed.

jvanburen: As in the name? I'd prefer to name it Naked_integer, but i was wary of changing things unnecessarily. Also curious to hear what they have to say.

typing/primitive.ml
jvanburen: They should be no-ops? `Untagged_immediate` is now just `Unboxed_integer Unboxed_int`.

typing/primitive.ml
jvanburen: It was previously handled by `if is_builtin_prim_name name then no_non_value_repr` since all of the primitives operated only on values, but this added code now handles the new unboxed primitives

typing/scalar.ml
jvanburen: I was intending to do that later, we want to be able to shift by naked int8.
I was trying to keep this PR smaller like the feedback you gave earlier.

typing/scalar.ml
jvanburen: same as above, left for later.

typing/scalar.mli
jvanburen: added a comment

typing/predef.ml
ccasin: I think I can be convinced this type is useful for writing fast code that has to interface with existing OCaml code, and more trouble to get rid of than to have.  But if this PR is going to add it, I think it also needs good tests of its runtime behavior.

typing/scalar.mli
```diff
@@ -0,0 +1,325 @@
...
+
+    - Primitive arguments don't depend on the locality of their arguments, but the results
+    do.
+    - Some primitives only take integers, some take only floats, and Three_way_compare
+    takes any scalar type.
+    - The bytecode compiler wants to easily map unboxed/untagged values to their [value]
+    equivalents
+    - The middle-end wants to easily cast between any integral values using only certain
+    primitives.
+*)
```
ccasin, line 80: This comment is super helpful - thanks for writing it!  I think it can be improved even further by making the last bit more concrete: the comment says the module is organized to make it easy to represent these four things, but does't actually say how they are represented. Can you show for each bullet how the types in this module are used for these purposes?

testsuite/tests/typing-layouts/hash_types.ml
```diff
@@ -308,21 +308,23 @@ Error: The type constructor "t#" expects 1 argument(s),
...
 type t = int
-and bad = t#
+and ok = t#
 [%%expect{|
-Line 2, characters 0-12:
-2 | and bad = t#
-    ^^^^^^^^^^^^
-Error: The type "t" has no unboxed version.
+type t = int
+and bad = t#
```
ccasin, line 321: something funny here - it looks like this test needs to be promoted?

typing/scalar.mli
ccasin: I think recording this intuition, along with an example or two, in the top-level comment would be helpful.

typing/typeopt.ml
ccasin: It looks like this change is still here.

typing/scalar.mli
ccasin: As mentioned in person, I think some of the types mentioned in the extended version don't exist (e.g., `Integral.Binary.Int_op.t`).

typing/scalar.mli
```diff
@@ -0,0 +1,339 @@
...
+      result : 'mode t
+    }
+
+  module Unary : sig
+    module Int_op : sig
+      type t =
+        | Neg
+        | Succ  (** add 1 *)
+        | Pred  (** subtract 1 *)
+        | Bswap
```
ccasin: I propose that `Bswap` is at least as deserving of a comment as `Succ` and `Pred`.

typing/scalar.mli
```diff
@@ -0,0 +1,342 @@
...
+    end
+
+    type nonrec 'mode t =
+      | Integral of 'mode Integral.t * Int_op.t
+      | Floating of 'mode Floating.t * Float_op.t
+      | Static_cast of
+          { src : any_locality_mode t;
+            dst : 'mode t
+          }
+      (** See scalar.md in the repo root for documentation on the semantics. *)
```
ccasin: I think this is now `jane/doc/scalars.md`

typing/scalar.mli
```diff
@@ -0,0 +1,342 @@
...
+    | CFnge
+
+  val to_string : t -> string
+
+  val swap : t -> t
+
+  val negate : t -> t
+end
+
+module Intrinsic : sig
```
ccasin: This module (`Intrinsic`) isn't mentioned in the top-level comment and I'm not immediately sure what it represents. I think it's different than thing we usually call an intrinsic (the `external`s that get a special implementation in the compiler exposed in the `ocaml_intrinsic` library), which makes me wonder if another name is better.  Maybe `Operation`?

typing/scalar.mli
```diff
@@ -0,0 +1,342 @@
...
+(*   the GNU Lesser General Public License version 2.1, with the          *)
+(*   special exception on linking described in the file LICENSE.          *)
+(*                                                                        *)
+(**************************************************************************)
+
+(** This module defines:
+    The scalar types intrinsic to the OCaml compiler, and all of the
+    primitive operations defined on them.
+
+    Overview: This module provides a comprehensive type system for scalar values in OCaml.
```
ccasin, line 19: Many of the comments in this file are too wide.  Unsure if ocamlformat is not working, or if it doesn't handle comments in the compiler and they need to be filled manually.

typing/primitive.ml
ccasin: I think your instinct to hesitate to make unnecessary changes is reasonable, and I've argued elsewhere that "Naked" is not something used by the front-end.  But seeing this confusing case, I think I agree we should change it.  And since that name is used by the middle-end, it seems fine here.

And should we actually have `| Naked_integer of any_locality_mode Scalar.Integral.Width.t` here rather than defining an additional `unboxed_integer` type?

typing/primitive.ml
```diff
@@ -308,8 +315,10 @@ let print p osig_val_decl =
      | Repr_poly -> []
      | Unboxed_float _
      | Unboxed_vector _
-     | Unboxed_integer _ -> if all_unboxed then [] else [oattr_unboxed]
-     | Untagged_immediate -> if all_untagged then [] else [oattr_untagged]
+     | Unboxed_integer (Unboxed_int32 | Unboxed_int64 | Unboxed_nativeint)
+       -> if all_unboxed then [] else [oattr_unboxed]
```
ccasin: I think the idiomatic thing is to put the `->` on the previous line (see the `Same_as_ocaml_repr` case below).

typing/primitive.ml
```diff
@@ -348,7 +357,10 @@ let unboxed_vector = function
    comparison at no performance loss. We still match on the variants to prove here that
    they are all constant constructors. *)
 let equal_unboxed_integer
-      ((Unboxed_int32 | Unboxed_nativeint | Unboxed_int64) as i1) i2 = i1 = i2
+      ((Unboxed_int | Unboxed_int8 | Unboxed_int16 | Unboxed_int32 | Unboxed_nativeint
+       | Unboxed_int64) as i1) i2
+  =
```
ccasin: Similarly, I think the `=` always goes on the line with the pattern, rather than it's own line.  (Also the line two above is too long, but I guess the 80ch check will catch that once this is rebased)

typing/predef.ml
```diff
@@ -543,7 +554,10 @@ let add_small_number_beta_extension_types add_type env =
   let _, add_type = mk_add_type add_type in
   env
   |> add_type ident_int8 ~jkind:Jkind.Const.Builtin.immediate
+       ~unboxed_jkind:Jkind.Const.Builtin.kind_of_unboxed_int8
   |> add_type ident_int16 ~jkind:Jkind.Const.Builtin.immediate
+       ~unboxed_jkind:Jkind.Const.Builtin.kind_of_unboxed_int16
+
```
ccasin, line 668: stray newline

typing/predef.mli
```diff
@@ -126,7 +134,7 @@ val list_argument_jkind : jkind_lr
    recursion between predef and env, we break it by parameterizing
    over Env.t, Env.add_type and Env.add_extension. *)
 
-val build_initial_env:
+val build_initial_env :
```
ccasin: stray space

typing/types.ml
```diff
@@ -937,18 +939,16 @@ let compare_tag t1 t2 =
...
     -> true
-  | (Value | Float64 | Float32 | Float_boxed | Word | Bits32 | Bits64 | Vec128), _
+  | (Value | Float64 | Float32 | Float_boxed | Word | Bits8 | Bits16 | Bits32 | Bits64 | Vec128), _
     -> false
 
 let compare_mixed_block_element e1 e2 =
   match e1, e2 with
   | Value, Value | Float_boxed, Float_boxed | Float64, Float64 | Float32, Float32
-  | Word, Word | Bits32, Bits32 | Bits64, Bits64 | Vec128, Vec128
+  | Word, Word | Bits8, Bits8| Bits16, Bits16 |Bits32, Bits32 | Bits64, Bits64 | Vec128, Vec128
```
ccasin: wonky spacing around the `|`s

typing/primitive.ml
mshinwell: I'm not really sure as I don't know the wider code here.  However in general I think it's probably a mistake to conflate untagged and unboxed things.  We also need to be careful because an untagged immediate has 63-bit integer semantics, not 64.

typing/env.ml
```diff
@@ -2953,7 +2931,24 @@ let add_language_extension_types env =
...
+      | false -> env
+    in
+    empty
+    |> Predef.build_initial_env
+         (add_type ~check:false)
+         (add_extension ~check:false ~rebind:false)
+    |> add SIMD Stable Predef.add_simd_stable_extension_types
+    |> add Small_numbers Stable Predef.add_small_number_extension_types
+    |> add Small_numbers Beta Predef.add_small_number_beta_extension_types
+    |> add Layouts Stable Predef.add_or_null)
```
ccasin: The reason for the change to this function (beyond adding small_numbers) isn't clear to me.

typing/typedecl.ml
```diff
@@ -4498,8 +4509,10 @@ let report_error ppf = function
         Style.inline_code "int64"
         Style.inline_code "nativeint"
   | Cannot_unbox_or_untag_type Untagged ->
-      fprintf ppf "@[Don't know how to untag this type. Only %a@ \
-                   and other immediate types can be untagged.@]"
+      fprintf ppf "@[Don't know how to untag this type. Only %a, %a, %a, \
+                   and@ other immediate types can be untagged.@]"
+        Style.inline_code "int8"
+        Style.inline_code "int16"
```
ccasin: I think I would leave this message unchanged.  It already says "and other immediate types", and I don't see a good reason to call out `int8` and `int16` when we don't call out `bool`, `char`, and `unit` - probably it's unnecessary and potentially confusing detail in most cases when this message is reached.

testsuite/tests/typing-layouts/hash_types.ml
ccasin: Also, it looks like the point of the parts of this test that have changed is to check that we get appropriate errors when trying to refer to an unboxed version of a type that doesn't have one. Rather than just promoting, I think we should update the test to use something other than `int`, to make sure we're still testing that appropriately - maybe `char` would do for now.

testsuite/tests/typing-layouts-bits16/basics_alpha.ml
```diff
@@ -0,0 +1,755 @@
+(* NOTEST
```
ccasin: Why aren't we running this?

testsuite/tests/typing-layouts-bits8/basics_alpha.ml
```diff
@@ -0,0 +1,755 @@
+(* NOTEST
```
ccasin: Also confused by `NOTEST` here.

testsuite/tests/typing-layouts-bits16/basics.ml
```diff
@@ -0,0 +1,760 @@
+(* TEST
```
ccasin: Comparing this directory to the `typing-layouts-bits32`, I see a couple missing tests:
- `alloc.ml`. This test describes itself as comparing allocations when using int32# vs int32. So I agree its behavior will change - but we should still be checking `int16#` versions of those functions don't allocate (and I guess just copying and slightly modifying the test, so that it is also showing `int16` doesn't allocate, is easiest and nice).
- `unboxed_bits16_alpha.ml`. Unsure why this one is missing.

(And the same for the bits8 tests)

typing/env.ml
```diff
@@ -2953,7 +2931,24 @@ let add_language_extension_types env =
...
+        f (add_type ?shape:None ~check:false) env
+      | false -> env
+    in
+    empty
+    |> Predef.build_initial_env
+         (add_type ~check:false)
+         (add_extension ~check:false ~rebind:false)
+    |> add SIMD Stable Predef.add_simd_stable_extension_types
+    |> add Small_numbers Stable Predef.add_small_number_extension_types
+    |> add Small_numbers Beta Predef.add_small_number_beta_extension_types
```
ccasin: On reflection, I think we should move small_numbers straight into stable with this PR. Mainly because of Leo's suggestion a few months ago that we should on the margins be moving more things that aren't especially risky straight into stable, and then just hold off on widely announcing them until we have good tests and tried some realistic uses in the tree. I do think there are some parts of this PR that are a bit risky (the large refactoring of primitives may introduce a subtle and hard-to-spot behavior change) but putting int8 and int16 behind this extension flag isn't giving us any help with that part of things.

If you agree, I think this, `jkind.ml` and `typeopt.ml` are the three places that have an extension check.

testsuite/tests/typing-layouts-bits16/basics.ml
```diff
@@ -0,0 +1,760 @@
...
+ {
+   flags = "-extension layouts_beta -extension small_numbers_beta";
+   expect;
+ }
```
ccasin: If you agree with the suggestion elsewhere to move this straight into stable, we'll need to make the test stanza here match the other similar layouts tests, where it tests both stable and beta.  (And also in several other tests in this directory, and the bits8 tests)

testsuite/tests/typing-layouts-bits16/c_api.ml
```diff
@@ -0,0 +1,73 @@
...
+ }{
+   flags = "-extension layouts_beta -extension small_numbers_beta";
+   bytecode;
+ }
+*)
+
+(* This file tests using external C functions with int16#. *)
+
+
+external to_int16 : int16# -> int16 = "%tag_int16" [@@warning "-187"]
```
ccasin: Something seems funny here - I think warning 187 is only supposed to fire when compiling with `-extension-universe upstream_compatible`, and this test doesn't use that flag. So either this isn't needed here or we've broken something about warning 187.

(And same thing in the `bits8` version)

testsuite/tests/typing-layouts-bits16/c_functions.c
```diff
@@ -0,0 +1,59 @@
...
+#include <caml/custom.h>
+
+
+static int16_t Int16_val(value v) {
+  return (int16_t)Long_val(v);
+}
+
+static value caml_copy_int16(int16_t x) {
+  return Val_long(x);
+}
```
ccasin: I think that for float32, we added `float32.{c,h}` to the runtime.  Should we do the same for int16 / int8 (and put these two functions in it)?

testsuite/tests/typing-layouts-bits16/test_int16_u.ml
```diff
@@ -0,0 +1,318 @@
...
+  test_unary_of  "to_int"              Int16.to_int              Int16_u.to_int               int_result;
+  test_unary_to  "of_float"            Int16.of_float            Int16_u.of_float             float_input;
+  test_unary_of  "to_float"            Int16.to_float            Int16_u.to_float             float_result;
+  (* test_unary_to  "of_string"           Int16.of_string           Int16_u.of_string            int16_string_input; *)
+  test_unary_of  "to_string"           Int16.to_string           Int16_u.to_string            string_result;
+  test_binary_of "compare"             Int16.compare             Int16_u.compare              int_result;
+  test_binary_of "unsigned_compare"    Int16.unsigned_compare    Int16_u.unsigned_compare     int_result;
+  test_binary_of "equal"               Int16.equal               Int16_u.equal                bool_result;
+  test_binary    "min"                 Int16.min                 Int16_u.min;
+  test_binary    "max"                 Int16.max                 Int16_u.max;
```
ccasin: Relative to the `int32` version of this test, I see four changes: `unsigned_to_int`, `bits_of_float` and `float_of_bits` are missing, and `of_string` has been commented out.

* `unsigned_to_int` does appear in `Int16.ml` and `Int16_u.ml`, so I'm a confused about its absence here.
* The float functions are missing from the stdlib module. That seems OK fo now, we can add these for int16 later if people want them.  (for int8, I think there isn't even an ieee standard for 8 bit floats, though we could also consider adding that following whatever the nvidia spec is).
* `of_string` is also missing from the stdlib module, but I think we should probably implement it?  Maybe it was hard for some reason?

testsuite/tests/typing-layouts-bits16/unboxed_bits16s.ml
```diff
@@ -0,0 +1,421 @@
...
+  print_int16u "Test 1, minus_five" minus_five;
+
+  let twice_minus_five = minus_five + ((Stdlib_beta.Int16_u.of_int (-5))) in
+  print_int16u "Test 1, twice_minus_five" twice_minus_five;
+
+  let thrice_minus_five = (Stdlib_beta.Int16_u.of_int 3) * minus_five in
+  print_int16u "Test 1, thrice_minus_five" thrice_minus_five;
+
+  let twice_minus_five_again = thrice_minus_five - minus_five in
+  print_int16u "Test 1, twice_minus_five_again" twice_minus_five;
```
ccasin: It looks like you fixed this to print `twice_minus_five_again` in the int32 version (nice catch!), but this and the int8 version also need to be updated.

testsuite/tests/typing-layouts-bits16/unboxed_bits16s.ml
```diff
@@ -0,0 +1,421 @@
...
+
+  (* (30 - 20) / 3 = 3 *)
+  let o = (Sys.opaque_identity f6_1) () in
+  print_int16u "Test 6, 3"
+    (o#f6_m1 (Stdlib_beta.Int16_u.of_int 30) (Stdlib_beta.Int16_u.of_int 20) (Stdlib_beta.Int16_u.of_int 3));
+
+  (* 4 * 8 = 16 *)
+  let o = (Sys.opaque_identity f6_2) (4,7) in
+  let result = o#f6_m2 8 (Stdlib_beta.Int16_u.of_int 4) (fun x -> x * (Stdlib_beta.Int16_u.of_int 2)) in
+  print_int16u "Test 6, 16" result;
```
ccasin: Looks like an overzealous find-and-replace changed the reference answer here from 32 to 16

(And similarly in the int8 file)

testsuite/tests/typing-layouts-bits16/unboxed_bits16s.ml
```diff
@@ -0,0 +1,421 @@
...
+
+let _ = test6 ()
+
+(*****************************************)
+(* Test 7: int16# and assert false joins *)
+
+module M = struct
+  open Int16_u
+  let[@inline never] f () = assert false
+  let g () = if Sys.opaque_identity true then (Stdlib_beta.Int16_u.of_int 16) else f ()
```
ccasin: Another spot where a `32` literal was changed to `16` (and in the print statement just below, as well).

(And similarly in the int8 file)

testsuite/tests/typing-layouts-bits8/unboxed_bits8s.ml
```diff
@@ -0,0 +1,422 @@
...
+  print_int8u "Test 3, 5 * 3: " (five_times_three ());
+  Array.iteri (Printf.printf "  Test 3, step %d: %d\n") (Array.map Int8.to_int steps);
+
+  (* Test f3_manyargs
+
+          (1 + 2 + 3 + 5 + 8) = 19
+      3 * (1 + 2 + 3 + 5 + 8) = 57
+      6 * (1 + 2 + 3 + 5 + 8) = 114
+      9 * (1 + 2 + 3 + 5 + 8) = 171
+      12 * (1 + 2 + 3 + 5 + 8) = 228
```
ccasin: It looks like you found and fixed a bug here that also occurs in the other versions of this test, which do in fact go to 228, but the comment doesn't reflect that and the printing a few lines below (on line 285) is also wrong in those versions (claiming 171 will be printed where 228 appears).  Well spotted!

Can you fix the other versions too?  (16, 32 and 64 all have this typo)

It might also be good to change the string we print in this version to reflect that we expect to see `-28` in the output rather than `228` (because it's a signed int8).

testsuite/tests/lib-smallint/test_int16_u.ml
```diff
@@ -0,0 +1,194 @@
...
+  assert (
+    Smallint.equal (Smallint.succ (Smallint.max_int ())) (Smallint.min_int ()));
+  assert (
+    Smallint.equal (Smallint.pred (Smallint.min_int ())) (Smallint.max_int ()));
+  test_arith2 Smallint.add Int.add;
+  test_arith2 Smallint.sub Int.sub;
+  test_arith2 Smallint.mul Int.mul;
+  test_arith2 Smallint.div Int.div;
+  (* test_arith2 Smallint.unsigned_div Int.unsigned_div ~unsigned:true;
+   * test_arith2 Smallint.unsigned_rem Int.unsigned_rem ~unsigned:true; *)
```
ccasin: Why are these tests commented out?  The functions seem to exist in the library.

(And same question for the int8 version)

testsuite/tests/lib-smallint/test_int16_u.ml
```diff
@@ -0,0 +1,194 @@
...
+      (apply_shift reference_shift_right_logical);
+    test_conv1
+      (apply_shift' Smallint.shift_left)
+      (apply_shift Int.shift_left)
+      ~equal:(if shift = 0 then equal_logical else equal_arith)
+  done;
+  test2 (fun x y ->
+      assert (Smallint.equal (of_int x) (of_int y) = Int.equal x y));
+  test2 (fun x y ->
+      assert (Smallint.compare (of_int x) (of_int y) = Int.compare x y));
```
ccasin: There's a test here for `unsigned_compare` in the tagged version that seems to be missing here.

(And same question for the int8 version)

otherlibs/stdlib_beta/int16.ml
```diff
@@ -14,17 +14,92 @@
 
 [@@@ocaml.flambda_o3]
 
+open! Stdlib
+
 type t = int16
 
-include
-  Int_wrapper.Make
```
ccasin, line 20: Is `Int_wrapper` now dead code? Should it be deleted?

jane/doc/scalars.md
```diff
@@ -0,0 +1,280 @@
+# Scalar Types and Static Casting
```
ccasin: I know I asked for this documentation, and do appreciate you adding it.  But something feels off here, to me - it's surprising that it takes 280 lines to explain the semantics of the one `Static_cast` operation I was asking about. I think this means that one operation is orders of magnitude more complex than any other primitive or scalar operation we have at the lambda level. This suggests to me that we should instead have a larger number of simpler operations with names or parameters that make their semantics clear.

lambda/lambda.mli
```diff
@@ -374,15 +334,14 @@ and extern_repr =
...
+  | Unboxed_integer of unboxed_integer
 
 and external_call_description = extern_repr Primitive.description_gen
 
-and integer_comparison =
+and integer_comparison = Scalar.Integer_comparison.t =
     Ceq | Cne | Clt | Cgt | Cle | Cge
 
-and float_comparison =
+and float_comparison = Scalar.Float_comparison.t =
```
ccasin: These types are no longer mentioned in this file or elsewhere in `lambda/` - I think it's probably clearer to delete them and update references elsewhere in the compiler to mention `Scalar.foo`.

lambda/lambda.mli
```diff
@@ -394,6 +353,7 @@ and array_kind =
   | Pgcignorableproductarray of ignorable_product_element_kind list
   (* Invariant: the product element kind lists have length >= 2 *)
 
+
```
ccasin: stray newline

lambda/lambda.mli
```diff
@@ -508,6 +470,10 @@ and unboxed_integer = Primitive.unboxed_integer =
   | Unboxed_int64
   | Unboxed_nativeint
   | Unboxed_int32
+  | Unboxed_int16
+  | Unboxed_int8
+  | Unboxed_int
+
```
ccasin: stray newline

lambda/lambda.mli
```diff
@@ -523,11 +489,12 @@ and boxed_integer = Primitive.boxed_integer =
 
 and boxed_vector = Primitive.boxed_vector =
   | Boxed_vec128
-
```
ccasin, line 574: accidentally removed newline

lambda/lambda.mli
```diff
@@ -578,6 +545,8 @@ val layout_of_extern_repr : extern_repr -> layout
 
 type structured_constant =
     Const_base of constant
+  | Const_naked_immediate of int * Scalar.Integral.Taggable.Width.t
+  (** this is a stopgap until naked immediate values exist in [constant] *)
```
ccasin: I think I am surprised this is needed, since everything we have literals for is already represented by `constant`.  Looking quickly, it seems to only be created by `const_int`, but as far as I can see that is only called with the `size` parameter of `Lambda.int`, which won't hit that case.  So maybe dead code?

lambda/lambda.mli
```diff
@@ -1224,3 +1188,33 @@ val primitive_can_raise : primitive -> bool
 val count_initializers_array_kind : array_kind -> int
 val ignorable_product_element_kind_involves_int :
   ignorable_product_element_kind -> bool
+
+(** construction helpers *)
+
+val int : _ Scalar.Integral.t
```
ccasin, line 1250: I think this could use a more specific name - I found reading various code that mentions it (e.g., simplif) a bit confusing.  Also maybe it belongs in `Scalar` rather than here?

lambda/lambda.mli
```diff
@@ -1224,3 +1188,33 @@ val primitive_can_raise : primitive -> bool
 val count_initializers_array_kind : array_kind -> int
 val ignorable_product_element_kind_involves_int :
   ignorable_product_element_kind -> bool
+
+(** construction helpers *)
```
ccasin: Maybe it's not worth splitting these out into a separate file, but I suggest creating a submodule called `Helpers` for them (and maybe submodules of that for `Unary_op` and `Binary_op`, though either way is fine). I think "helpers" is the idiomatic name for this kind of thing in the compiler - see `parsing/ast_helpers.mli` and `backend/cmm_helpers.mli` for example.

(Edit: but see comment on this part of the ml before making a change - I may be confused)

lambda/lambda.ml
```diff
@@ -124,6 +120,15 @@ type region_close =
   | Rc_nontail
   | Rc_close_at_apply
 
+
```
ccasin: extra newline

lambda/lambda.ml
```diff
@@ -215,25 +200,7 @@ type primitive =
...
-  | Pmodbint of { size : boxed_integer; is_safe : is_safe; mode: locality_mode }
-  | Pandbint of boxed_integer * locality_mode
-  | Porbint of boxed_integer * locality_mode
-  | Pxorbint of boxed_integer * locality_mode
-  | Plslbint of boxed_integer * locality_mode
-  | Plsrbint of boxed_integer * locality_mode
-  | Pasrbint of boxed_integer * locality_mode
-  | Pbintcomp of boxed_integer * integer_comparison
-  | Punboxed_int_comp of unboxed_integer * integer_comparison
+
```
ccasin: extra newline

lambda/lambda.ml
```diff
@@ -351,19 +311,18 @@ type primitive =
   (* Poll for runtime actions *)
   | Ppoll
 
+
```
ccasin: extra newline

lambda/lambda.ml
```diff
@@ -351,19 +311,18 @@ type primitive =
...
 and extern_repr =
   | Same_as_ocaml_repr of Jkind.Sort.Const.t
   | Unboxed_float of boxed_float
   | Unboxed_vector of boxed_vector
-  | Unboxed_integer of boxed_integer
-  | Untagged_int
+  | Unboxed_integer of unboxed_integer
 
 and external_call_description = extern_repr Primitive.description_gen
-
```
ccasin, line 401: accidentally removed newline

lambda/lambda.ml
```diff
@@ -427,6 +388,7 @@ and array_kind =
   | Pgcscannableproductarray of scannable_product_element_kind list
   | Pgcignorableproductarray of ignorable_product_element_kind list
 
+
```
ccasin: extra newline

lambda/lambda.ml
```diff
@@ -493,6 +458,8 @@ and peek_or_poke =
   | Ppp_tagged_immediate
   | Ppp_unboxed_float32
   | Ppp_unboxed_float
+  | Ppp_unboxed_int8
+  | Ppp_unboxed_int16
```
ccasin: I don't recall seeing tests for these new peek/poke options - I think `testsuite/tests/peek_and_poke/peek_and_poke.ml` is probably the right place, if I didn't miss them elsewhere.

lambda/lambda.ml
```diff
@@ -1820,9 +1775,16 @@ let locality_mode_of_primitive_description (p : external_call_description) =
...
+    (match (Scalar.Intrinsic.info op).result with
+     | Naked _ -> None
+     | Value (Integral (Taggable (Int8 | Int16 | Int))) -> None
+     | Value (Integral (Boxable (Int32 mode | Int64 mode | Nativeint mode))
+             | Floating (Float64 mode | Float32 mode)) -> Some mode)
+  | Pphys_equal _
   | Pbytes_to_string | Pbytes_of_string
   | Parray_to_iarray | Parray_of_iarray
-  | Pignore -> None
+  | Pignore  -> None
```
ccasin: stray space

lambda/lambda.ml
```diff
@@ -2143,13 +2058,12 @@ let primitive_can_raise prim =
...
+  | Pctconst _ | Pint_as_pointer _ | Popaque _
   | Pprobe_is_enabled _ | Pobj_dup | Pobj_magic _
-  | Pbox_float (_, _)
-  | Punbox_float _
   | Pbox_vector (_, _)
-  | Punbox_vector _ | Punbox_int _ | Pbox_int _ | Pmake_unboxed_product _
-  | Punboxed_product_field _ | Pget_header _ ->
+  | Punbox_vector _ | Pmake_unboxed_product _
+  | Punboxed_product_field _ | Pget_header _
+  | Pmakelazyblock _ ->
```
ccasin: I think `Pmakelazyblock` belongs up next to the other `Pmake*block` primitives

lambda/lambda.ml
```diff
@@ -2197,10 +2119,12 @@ let rec layout_of_const_sort (c : Jkind.Sort.Const.t) : layout =
...
 
 let layout_of_extern_repr : extern_repr -> _ = function
-  | Untagged_int ->  layout_int
   | Unboxed_vector v -> layout_boxed_vector v
   | Unboxed_float bf -> layout_boxed_float bf
-  | Unboxed_integer bi -> layout_boxed_int bi
+  | Unboxed_integer (Unboxed_int | Unboxed_int8 | Unboxed_int16) ->  layout_int
+  | Unboxed_integer (Unboxed_int64) -> layout_boxed_int Boxed_int64
+  | Unboxed_integer (Unboxed_int32) -> layout_boxed_int Boxed_int32
+  | Unboxed_integer (Unboxed_nativeint) -> layout_boxed_int Boxed_nativeint
```
ccasin:
```suggestion
  | Unboxed_integer Unboxed_int64 -> layout_boxed_int Boxed_int64
  | Unboxed_integer Unboxed_int32 -> layout_boxed_int Boxed_int32
  | Unboxed_integer Unboxed_nativeint -> layout_boxed_int Boxed_nativeint
```

lambda/lambda.ml
```diff
@@ -2267,56 +2211,33 @@ let primitive_result_layout (p : primitive) =
...
-  | Pmulbint (bi, _) | Pdivbint {size = bi} | Pmodbint {size = bi}
-  | Pandbint (bi, _) | Porbint (bi, _) | Pxorbint (bi, _)
-  | Plslbint (bi, _) | Plsrbint (bi, _) | Pasrbint (bi, _)
-  | Pbbswap (bi, _) | Pbox_int (bi, _) ->
-      layout_boxed_int bi
-  | Punbox_int bi -> Punboxed_int (Primitive.unboxed_integer bi)
   | Pstring_load_32 { boxed = true; _ } | Pbytes_load_32 { boxed = true; _ }
   | Pbigstring_load_32 { boxed = true; _ } ->
-      layout_boxed_int Boxed_int32
+    layout_boxed_int Boxed_int32
```
ccasin: Here and in the rest of this function there's a bunch of whitespace-only diff.  I am also frustrated about the inconsistent whitespace conventions in pattern matches in the compiler, but I think it's better to undo this, because it creates friction when navigating the history.

lambda/lambda.ml
```diff
@@ -2617,3 +2540,25 @@ let rec ignorable_product_element_kind_involves_int
   | Punboxedfloat_ignorable _ | Punboxedint_ignorable _ -> false
   | Pproduct_ignorable kinds ->
     List.exists ignorable_product_element_kind_involves_int kinds
+
+(* Functions for simulating naked int primitives. This is used for bytecode
+   compilation which doesn't support naked int primitives directly *)
```
ccasin: The mli just says these are construction helpers. If they are not supposed to be used in native code compilation, I think the `mli` comment needs to be explicit about that (and also in this case the comment I left in the mli suggesting to move them into a module called `Helpers` is wrong - that module should get a clearer name, like `Bytecode_helpers` or something).

It's not obvious to me what here is "simulating", though - to my untrained eye, these do just look like normal construction helpers that work equally on naked and boxed/tagged representations.

lambda/matching.ml
```diff
@@ -2669,13 +2668,33 @@ let rec do_tests_nofail value_kind loc tst arg = function
           do_tests_nofail value_kind loc tst arg rem,
           act, value_kind )
 
-let make_test_sequence value_kind loc fail tst lt_tst arg const_lambda_list =
+let make_test_sequence value_kind loc fail size arg const_lambda_list =
+  let icmp size cmp =
+    match (size : _ Scalar.Integral.t), cmp with
+    | Value (Taggable Int), Ceq -> Pphys_equal Eq
+    | Value (Taggable Int), Cne -> Pphys_equal Noteq
```
ccasin: I don't understand why we need these two special cases.

lambda/matching.ml
```diff
@@ -2669,13 +2668,33 @@ let rec do_tests_nofail value_kind loc tst arg = function
...
+  let lt_tst =
+    cmp Clt CFlt
+  in
   let const_lambda_list = sort_lambda_list const_lambda_list in
   let hs, const_lambda_list, fail =
     share_actions_tree value_kind const_lambda_list fail
   in
   let rec make_test_sequence const_lambda_list =
-    if List.length const_lambda_list >= 4 && lt_tst <> Pignore then
+    if List.length const_lambda_list >= 4 then
```
ccasin: I guess the && was already dead code? Fun.

I think it has been dead since upstream ocaml/ocaml#6269 ([this commit](https://github.com/ocaml/ocaml/commit/c2a88c27fd3867e022c8ae4363bb24bca864619b#diff-574a9ed85084e8419a9fdc29c147bd50cf338c63200992a97d330344062fba94))

lambda/matching.ml
```diff
@@ -2694,18 +2713,21 @@ let make_test_sequence value_kind loc fail tst lt_tst arg const_lambda_list =
...
   type primitive = Lambda.primitive
+  let int_scalar = Scalar.Maybe_naked.Value (Scalar.Integral.Width.Taggable Scalar.Integral.Taggable.Width.Int)
+  let pintcomp cmp =
+    Pscalar (Binary (Icmp (int_scalar, cmp)))
 
-  let eqint = Pintcomp Ceq
+  let eqint = Pphys_equal Eq
 
-  let neint = Pintcomp Cne
+  let neint = Pphys_equal Noteq
```
ccasin: Are these two things actually different, particularly when using the `int_scalar` scalar?  I think it is pretty confusing that we have these special cases for equal in this file and elsewhere - it's surprising to me that `Ceq` could be wrong where `Cle` is correct.

lambda/lambda.mli
```diff
@@ -1224,3 +1188,33 @@ val primitive_can_raise : primitive -> bool
...
+
+type 'a binop := 'a -> lambda -> lambda -> loc:scoped_location -> lambda
+val add : locality_mode Scalar.Integral.t binop
+val sub : locality_mode Scalar.Integral.t binop
+val and_ : locality_mode Scalar.Integral.t binop
+
+(** This is ONLY for comparing scalar integral values.
+
+    It may NOT be applied to arbitrary arguments of kind value, even if they
+    are both known to be immediates. Use [phys_equal] for that. *)
```
ccasin: I think it would be good if this comment explained also _why_ it can't be used for immediates.  Just from reading the front end, it's not obvious to me how `Ceq` for the `int` scalar is going to get compiled differently than `phys_equal` (or maybe this isn't the perfect place for that explanation and it should go in `translprim` or the `scalars`, with a pointer to there from here).

lambda/transl_comprehension_utils.ml
```diff
@@ -31,8 +31,12 @@ module Let_binding = struct
 end
 
 module Lambda_utils = struct
+  let int_scalar =
+    Scalar.Maybe_naked.Value
+      (Scalar.Integral.Width.Taggable Scalar.Integral.Taggable.Width.Int)
```
ccasin: I think this is `Lambda.int` (which as noted elsewhere should probably move to `Scalars`).

lambda/translprim.ml
```diff
@@ -588,60 +612,55 @@ let lookup_primitive loc ~poly_mode ~poly_sort pos p =
...
+    | "%int64_lsl" -> binary (Shift (int64, Lsl, Int))
+    | "%int64_lsr" -> binary (Shift (int64, Lsr, Int))
+    | "%int64_asr" -> binary (Shift (int64, Asr, Int))
+    | "%nativeint_of_int32" -> static_cast ~dst:(i nativeint) ~src:(i int32)
+    | "%nativeint_to_int32" -> static_cast ~src:(i nativeint) ~dst:(i int32)
+    | "%int64_of_int32" -> static_cast ~dst:(i int64) ~src:(i int32)
+    | "%int64_to_int32" -> static_cast ~src:(i int64) ~dst:(i int32)
+    | "%int64_of_nativeint" -> static_cast ~dst:(i int64) ~src:(i nativeint)
+    | "%int64_to_nativeint" -> static_cast ~src:(i int64) ~dst:(i nativeint)
+
```
ccasin: extra newline

lambda/translprim.ml
```diff
@@ -890,10 +909,10 @@ let lookup_primitive loc ~poly_mode ~poly_sort pos p =
...
     | "%array_to_iarray" -> Primitive (Parray_to_iarray, 1)
     | "%array_of_iarray" -> Primitive (Parray_of_iarray, 1)
-    | "%unbox_float" -> Primitive(Punbox_float Boxed_float64, 1)
-    | "%box_float" -> Primitive(Pbox_float (Boxed_float64, mode), 1)
-    | "%unbox_float32" -> Primitive(Punbox_float Boxed_float32, 1)
-    | "%box_float32" -> Primitive(Pbox_float (Boxed_float32, mode), 1)
+    | "%unbox_float" -> static_cast ~src:(f float) ~dst:(naked (f float))
+    | "%box_float" -> static_cast ~src:(naked (f float)) ~dst:(f float)
+    | "%unbox_float32" -> static_cast ~src:(f float32) ~dst:(naked (f float32))
+    | "%box_float32" -> static_cast ~src:(naked (f float32)) ~dst:(f float32)
```
ccasin, line 971: I think it's surprising for something called a "cast" to do allocation, making me further believe `Static_cast` is many different operations currently disguised as one.

Maybe it's six operations: `box`, `unbox`, `tag`, `untag`, `truncate` and `extend`.  Do we ever use it for something that isn't one of those, and do those always have simple, easy to document semantics?  I guess the float size conversions aren't really well described by "truncate" and "extend".

lambda/translprim.ml
```diff
@@ -923,23 +942,37 @@ let lookup_primitive loc ~poly_mode ~poly_sort pos p =
...
-       | None -> raise (Error (loc, Unknown_builtin_primitive s)))
-    | _ -> External lambda_prim
+    | s ->
+      if String.length s = 0 || s.[0] <> '%' then
+        External lambda_prim
+      else
+        (match String.Map.find_opt s indexing_primitives with
+         | Some prim -> prim ~mode
+         | None ->
+           match Scalar.Intrinsic.With_percent_prefix.of_string s with
```
ccasin: It's not very obvious to me why some scalar intrinsics appear explicitly in the match in this file, and others are instead found by lookup here.  If there's not a reason for this split, probably everything should move to `scalar.ml`.  But probably there is, and it's just not clear to me, in which case a comment on `of_string` or here would be helpful.

lambda/translprim.ml
```diff
@@ -1214,10 +1247,13 @@ let peek_or_poke_layout_from_type ~prim_name error_loc env ty
     match layout with
     | Punboxed_float Unboxed_float32 -> Some Ppp_unboxed_float32
     | Punboxed_float Unboxed_float64 -> Some Ppp_unboxed_float
+    | Punboxed_int Unboxed_int8 -> Some Ppp_unboxed_int8
+    | Punboxed_int Unboxed_int16 -> Some Ppp_unboxed_int16
     | Punboxed_int Unboxed_int32 -> Some Ppp_unboxed_int32
     | Punboxed_int Unboxed_int64 -> Some Ppp_unboxed_int64
     | Punboxed_int Unboxed_nativeint -> Some Ppp_unboxed_nativeint
     | Pvalue { raw_kind = Pintval ; _ } -> Some Ppp_tagged_immediate
+    | Punboxed_int Unboxed_int
```
ccasin: Can this one be implemented as `Some Ppp_tagged_immediate`?

lambda/translprim.ml
```diff
@@ -1955,12 +2011,12 @@ let lambda_primitive_needs_event_after = function
...
   | Patomic_compare_set _ | Patomic_fetch_add | Patomic_add | Patomic_sub
   | Patomic_land | Patomic_lor | Patomic_lxor | Patomic_load _ | Patomic_set _
-  | Pintofbint _ | Pctconst _ | Pbswap16 | Pint_as_pointer _ | Popaque _
+  | Pctconst _ | Pint_as_pointer _ | Popaque _
   | Pdls_get
-  | Pobj_magic _ | Punbox_float _ | Punbox_int _ | Punbox_vector _
+  | Pobj_magic _ | Punbox_vector _
   | Preinterpret_unboxed_int64_as_tagged_int63 | Ppeek _ | Ppoke _
   (* These don't allocate in bytecode; they're just identity functions: *)
-  | Pbox_float (_, _) | Pbox_int _ | Pbox_vector (_, _)
```
ccasin, line 2121: Just to check: I think we're newly going to be sticking events after some `box` primitives as a result of this change. That's probably fine - I don't think this affects native code at all.  But if that's right, the `Pscalar` above should probably get a comment noting it's conservative relative to the comment at the beginning of the function.

ccasin (commented):
I've read everything except `backend/`, `bytecomp/` and `middle_end/` (which unfortunately github says means I've only looked at 105/171 files, but @mshinwell and I can fight about how to divide the rest next week).

This is looking good overall - going to be a big improvement in things.  Many comments and questions below, mostly minor, but a few about design and correctness mixed in.

typing/scalar.mli
mshinwell: I think in interfaces they need to be filled manually (this is a regression from earlier behaviour, I will investigate at some point).

testsuite/tests/typing-layouts-bits16/unboxed_bits16s.ml
mshinwell: This seems to have been fixed as part of my conflict resolution for a second merge with oxcaml/main.

lambda/translprim.ml
jvanburen: Yes, it is a bunch of operations in one, would "convert" be a better name?

ccasin: I think it would be better to have separate operations - the semantics of this one is too complicated (see also my comment on the doc for it)

lambda/translprim.ml
mshinwell: I'm not sure either, but let's fix this in a subsequent PR if we need to.

lambda/lambda.mli
jvanburen: I don't remember why... maybe flambda2 doesn't like it?

lambda/matching.ml
jvanburen: I don't recall either anymore

jvanburen: okay it's similar to the comment from lambda.ml, it is because otherwise flambda2 will complain. I'll add an example / test case

lambda/matching.ml
jvanburen: yes

lambda/translprim.ml
jvanburen: it already says "may" raise an exception or allocate, is that not enough?

lambda/translprim.ml
jvanburen: I don't think so, because it's used to determine the kind of the expression, which will make flambda2 mad

lambda/lambda.mli
jvanburen: alright i was just wrong, I changed the comment and replaced the phys_equals with Ceq and all the tests still pas

lambda/lambda.mli
jvanburen: It would need to go in Scalar.Integral.int, which is a little verbose... and it only appears on one line in simplif? I would think that merlin jump to makes it pretty clear. I'm going to add a doc comment on Lambda.int and resolve this since it's not blocking merge :/

typing/primitive.ml
```diff
@@ -521,6 +520,7 @@ module Repr_check = struct
     check checks prim
 end
 
+
```
ccasin: extra newline

typing/primitive.mli
```diff
@@ -21,6 +21,12 @@ type unboxed_integer =
   | Unboxed_int32
   | Unboxed_int16
   | Unboxed_int8
+  | Unboxed_int
+  (* CR mshinwell: I think [Unboxed_int31_63] might be better *)
+  (** We don't support an [int#] type in the stdlib. However, for consistency,
+      it's important that we have naked types corresponding to all of our
+      integral types.  This is also used when you write [int[@untagged]] in
+      [external] functions. *)
```
ccasin: I think the comment above is out of date because `int#` has been added (iiuc)

typing/value_rec_check.ml
```diff
@@ -787,8 +787,8 @@ let rec expression : Typedtree.expression -> term_judg =
           | Record_mixed mixed_shape ->
             (match mixed_shape.(i) with
              | Value | Float_boxed -> Guard
-             | Float64 | Float32 | Bits8 | Bits16 | Bits32 | Bits64
-             | Vec128 | Vec256 | Vec512 | Word | Void | Product _ ->
```
ccasin, line 791: Looks like an unintentional diff.

typing/scalar.mli
mshinwell: I've added a comment, any renaming should happen later I think.  The Intrinsic module is now mentioned in the comment at the top of the file, though, which clarifies matters.

testsuite/tests/typing-layouts/hash_types.ml
mshinwell: "string" is used now

typing/primitive.ml
mshinwell: I've done a big rename to fix this, see the current HEAD.

mshinwell: And I've left a CR about the suggestion about using `Width`.

typing/scalar.mli
mshinwell: I've decided to draw the line at adding the above plus another sentence in the documentation, I think that should be clear enough.

testsuite/tests/typing-layouts/hash_types.ml
mshinwell: This has probably been looked at enough now (I very much hope)

jane/doc/scalars.md
mshinwell: As discussed, this doc is more for the semantics through the backend, I'll add something separately which is much shorter.

typing/typeopt.ml
mshinwell: Fixed

typing/env.ml
mshinwell: Let's do this in a separate PR please

jane/doc/scalars.md
mshinwell: I've expanded the Static_cast comment, I think it's enough now especially in conjunction with the type having been changed to make it more explicit that only conversions (and (un)boxing/(un)tagging ops) are involved in these casts, no actual operations.  I was confused about this before.

testsuite/tests/typing-layouts-bits8/unboxed_bits8s.ml
jvanburen: credit goes to diana :) i'll change the others

typing/scalar.mli
mshinwell: We think this has been done

testsuite/tests/typing-layouts-bits16/c_functions.c
jvanburen: no, these functions are just for testing the api, they're already covered by `Val_long` and `Long_val`

testsuite/tests/typing-layouts-bits16/test_int16_u.ml
jvanburen: There is no float16 or float8 type so I don't think there _can_ be a bits_of_float or its inverse
I can add the other things...

ccasin: In the int32 case, `bits_of_float` works on `float`, not `float32`.  So I don't think we need the `float16` or `float8` types.  But I think it's fine to leave them off for now anyway.

jvanburen: What also doesn't help is that there's no consensus between ieee 754 binary16 and bfloat16, which are different formats! marking as resolved.

jvanburen: done, can't resolve for some reason

testsuite/tests/lib-smallint/test_smallint.ml
```diff
@@ -1,4 +1,164 @@
...
+  (** A seeded hash function for ints, with the same output value as
+      {!Hashtbl.seeded_hash}. This function allows this module to be passed as
+      argument to the functor {!Hashtbl.MakeSeeded}. *)
+  val seeded_hash : int -> t -> int
+
+  (** An unseeded hash function for ints, with the same output value as
+      {!Hashtbl.hash}. This function allows this module to be passed as argument
+      to the functor {!Hashtbl.Make}. *)
+  val hash : t -> int
+end
```
ccasin, line 161: This feels sad - I think it would be better to use the intf pattern or even module type of (but I prefer the former).  We do use the intf pattern a bit in our extensions.

PR #3514, opened 2025-01-25
Implicit source position: allow caller to explicitly pass an argument as `[%call_pos]`

ncik-roberts:
Allow users to write wrapper functions that call call-pos functions:
```ocaml
let wrap b f g ~here x = if b then f ~(here : [%call_pos]) x else g ~(here : [%call_pos]) x
```
If you drop the `[%call_pos]` annotations, `f` and `g` are inferred as having a normal label, not a `%call_pos` label. Prior to this PR, the easiest way of getting the proper types was with a type annotation:
```ocaml
let wrap b (f: here:[%call_pos] -> _) (g: here:[%call_pos] -> _) ~here x = ...
```
which works out ok in this small example, but becomes more painful if there are many args before the `%call_pos` arg.

The new feature feels fairly similar to the lightweight way that you can tell the compiler to infer an optional label for a function arg: `let f g x = g ?label:(Some x)`, as opposed to `let f g x = g ~label:x`.

This is upstream-compatible. You can erase `f ~(here : [%call_pos])` to `f ?here:(Some here)`. I haven't written the ocamlformat feature, but I could do this.

### DIFF
--- a/testsuite/tests/typing-implicit-source-positions/expressions.ml
+++ b/testsuite/tests/typing-implicit-source-positions/expressions.ml
@@ -52,6 +52,35 @@ let _ = wrap `G (fun ~here:_ -> assert false)
 {pos_fname = ""; pos_lnum = 4; pos_bol = 1164; pos_cnum = 1186}
 |}]
 
+(* call_pos type annotations not permitted anywhere other than labelled arg
+   position *)
+let error1 x = (x : [%call_pos])
+
+[%%expect{|
+Line 1, characters 22-30:
+1 | let error1 x = (x : [%call_pos])
+                          ^^^^^^^^
+Error: [%call_pos] can only exist as the type of a labelled argument
+|}]
+
+let error2 f x = f (x : [%call_pos])
+
+[%%expect{|
+Line 1, characters 24-35:
+1 | let error2 f x = f (x : [%call_pos])
+                            ^^^^^^^^^^^
+Error: A position argument must not be unlabelled.
+|}]
+
+let error3 f x = f ?x:(x : [%call_pos])
+
+[%%expect{|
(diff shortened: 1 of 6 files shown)


testsuite/tests/typing-implicit-source-positions/expressions.ml
```diff
@@ -27,6 +27,31 @@ let _ = f ~call_pos:[%src_pos] () ;;
 {pos_fname = ""; pos_lnum = 2; pos_bol = 438; pos_cnum = 458}
 |}]
 
+(* passing an argment explicitly as call_pos *)
```
goldfirere, line 30: I would want a test showing that `[%call_pos]` is not accepted as a type in other positions

typing/typecore.ml
```diff
@@ -8040,6 +8040,8 @@ and type_apply_arg env ~app_loc ~funct ~index ~position_and_mode ~partial_app (l
         (* CR layouts v5: relax value requirement *)
         unify_exp env arg
           (type_option(newvar Predef.option_argument_jkind));
+      if is_position lbl then
```
goldfirere, line 8043: Style nit: maybe use a `match lbl with` here instead of two `if`s?

goldfirere (changes requested):
cc @Enoumy, who may have opinions here.

I'm happy both with the design and implementation.

Please add documentation of the new feature to `jane/doc/implicit_source_code_positions/readme.md` and an example to `testsuite/tests/parsetree/source_jane_street.ml`.

Thanks for contributing this!

ncik-roberts:
Thanks @goldfirere for the review. I've incorporated both points of feedback.

goldfirere (approved):

PR #3517, opened 2025-01-27
`-dparsetree` should print kind annotations

ccasin:
This makes `-dparsetree` print kind annotations in type declarations.

There is also a test for a particular case we were interested in, confirming it parses the way we believe.


dvulakh (approved):
LGTM, thanks for fixing this

PR #3520, opened 2025-01-28
Fix under-reporting of custom block sizes in runtime4 Gc.Memprof

stedolan:
The custom block sizes should be rounded *up* to words, not down.

(Already correct in runtime5)


NickBarnes (approved):
Obviously correct. Maybe `Wsize_bsize` should round up automatically (or have a better name).

PR #3522, opened 2025-01-28
Refactor new mode syntax

riaqn:
This PR refactors the new mode syntax (`@` and `@@`).

- With this PR, `@` always means modes, and `@@` always mean modalities. Previously, `@@` can mean modes in some position for disambiguition. For example, `x : a -> a @@ local` says `x` is local, and `x : b -> a @ local` says `a` is local). After this PR, the first example will write `x : (a -> a) @ local`. The second example doesn't change. You can also write `x : a @ local`.
- To ensure certain parsing-printing roundtrip, we print modes in legacy position only if the whole mode annotation contains old modes only. That means `local_ a @ portable -> b` will print back `a @ local portable -> b`, while `local_ a -> b` prints back the same. The idea is that "if the user already knows about and uses the new mode syntax, then we should print back new mode syntax". This is done in both `pprintast.ml` and `printtyp.ml`.
- `fun a b : ty @@ modes -> ..` was introduced by #3327 , and would be changed to `fun a b : ty @ modes -> ..`. Note the ambiguity:  the arrow could be arrow type. We decide to remove this for simplicity.
- Refactoring and fixes in `parser.mly` and `pprintast.ml`, so that some missing corner cases are covered, and diff to the upstream is minimized. For example, `labeled_simple_pattern` is now almost identical to upstream. This allows systematic enumeration of all possible syntax, which helps with user documentation, and test coverage.
- The mode section in `source_jane_street.ml` is rewritten to cover all possible syntax.

# Suggestions to reviewers

- First look at the mode section of `source_jane_street.ml`, to get a sense of the new syntax (don't look at the type errors - they are fine).
- Then look at `parser.mly`. 
- Then look at `pprintast.ml` and `printtyp.ml`. The former should be functionally "most fine" as tested by `source_jane_street.ml`, but other aspects such as coding style should be inspected.
- Changes in tests, except `source_jane_street.ml`, are all mechanical and can be skipped.

### DIFF
--- a/jane/doc/extensions/modes/syntax.md
+++ b/jane/doc/extensions/modes/syntax.md
@@ -1,12 +1,3 @@
-# Motivation
-Previously we have a fixed set of mode names in the parser. Everytime we want to
-add more modes we need to update the parser (and correspondingly ocamlformat).
-This is too costy in the long term.
-
-We introduce a new mode syntax that allows mode names (and mode expressions in
-the future) to enjoy a dedicated name space. As a result, any identifier can be
-used as a mode name.
-
 # Modes and modalities
 Currently a mode expression is simply a space-delimited list of modes.
 
@@ -23,55 +14,23 @@ local
 local unique
 ```
 
-Currently modalities have  the same syntax as modes:
+Modes are in a dedicated namespace separated from variable names or type names,
+which means you can continue to use `local` or `unique` as variable or type
+names.
+
+Currently a modality expression is simply a space-delimited list of modalities.
 
 ```
 modality := local | global | ..
 modalities := separated_nonempty_list(SPACE, modality)
 ```
+Similarly, modalities are in a dedicated namespace.
 
 # Where can they appear
 
-Mode expressions and modalities can appear after either `@` or `@@`. The former
-is used in most cases, while the latter is used for disambiguition. For example,
-in `val f : a -> b @ global`, the `@ global` could be annotating either `b` or
-`f`. To avoid the ambiguity, we will write `val f : a -> b @ local @@ global`
-where `local` is the mode of `b` and `global` is the modality on `f`.
-
-## Function parameter
-```ocaml
-let foo ?(label_let_pattern = default) = ..
-let foo ~(label_let_pattern) = ..
-let foo (x @ modes) = ..
-let foo (x : ty @@ modes) = ..
-```
-where `label_let_pattern` could be
-```ocaml
-x @ modes
-x : ty @@ modes
-x : a b c. a -> b -> c @@ modes
-```
-Patterns that are not directly function parameters can’t have modes. For
-example:
-```
-let foo ((x, y) @ modes) = .. (error)
-```
-
-## Let bindings
-```ocaml
-let x @ modes = ..
-let x : ty @@ modes = ..
-let x : type a b. ty @@ modes = ..
-```
-
-## Functions and their return
-```ocaml
-let (foo @ modes) x y = ..
-let foo x y @ modes = ..
-let foo x y : ty @@ modes = ..
-fun foo x y @ modes -> ..
-fun foo x y : ty @@ modes -> ..
-```
+To write a mode expression in program, it has to be prefixed by an `@` symbol.
+Similarly, a modality expression has to be prefixed by an `@@` symbol. They can
+appear in several places in a program as described below.
 
 ## Arrow types
 ```ocaml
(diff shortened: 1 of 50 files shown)


parsing/pprintast.ml
```diff
@@ -802,11 +790,41 @@ and labeled_tuple_pattern ctxt f ~unboxed l closed =
     (list ~sep:",@;" (labeled_pattern1 ctxt)) l
     closed_flag closed
 
+and pattern2 ctxt f p =
```
tdelvecchio-jsc, line 793: This naming convention seems awful as is, but if we're going to keep it, we should update the comments accordingly (there's a comment above about "pattern" vs "pattern1"; should we use "pattern2" instead?).

parsing/pprintast.ml
```diff
@@ -802,11 +790,41 @@ and labeled_tuple_pattern ctxt f ~unboxed l closed =
...
+        optional_legacy_modes m
+        (simple_pattern ctxt) p
+    | None, false ->
+        pp f "@[<2>%a%a@]"
+        (simple_pattern ctxt) p
+        optional_at_modes m
+    end
+  | _ -> pattern1 ctxt f p
+
+and simple_pattern1 ctxt f p =
```
tdelvecchio-jsc, line 817: Ditto

parsing/pprintast.ml
```diff
@@ -1429,24 +1431,34 @@ and kind_abbrev ctxt f name jkind =
     string_loc name
     (jkind_annotation ctxt) jkind
 
+and module_type_with_optional_modes ctxt f (mty, mm) =
+  match mm with
+  | [] -> module_type ctxt f mty
+  | _ :: _ -> pp f "%a%a" (module_type1 ctxt) mty optional_at_modes mm
```
tdelvecchio-jsc, line 1437: Confirming: this is correct in calling `module_type1` instead of `module_type` here?

testsuite/tests/parsetree/source_jane_street.ml
```diff
@@ -687,13 +724,14 @@ Error: Mode annotations on modules are not supported yet.
...
   module M : S @@ portable
-  module M = N @ portable
-  module (M @ portable) = N
-  module M0 (_ : S @@ portable) (X : S @@ portable) : S @@ portable
-  (* The above [@@ portable] is the return mode of the functor, not the modality on the
-  module declaration; To do that, one must write in the following ways. *)
-  module (M0 @ portable) (_ : S @@ portable) (X : S @@ portable) : S @@ portable
-  module M0 : functor (_ : S @@ portable) (X : S @@ portable) -> S @ portable @@ portable
+  module M = N @@ portable
+  module (M @@ portable) = N
```
tdelvecchio-jsc, line 728: Any examples of `module (M @ portable) = N` or `module (F @ portable) () = N` in impls?

testsuite/tests/typing-modes/syntax-error.compilers.reference
```diff
@@ -148,8 +152,8 @@ Line 2, characters 16-19:
...
                     ^^^
   This "sig" might be unmatched
-Line 2, characters 10-33:
-2 | let foo = ("bar" :> int @@ local);;
-              ^^^^^^^^^^^^^^^^^^^^^^^
-Error: Syntax error: "mode annotations" not expected.
+Line 2, characters 31-32:
+2 | let foo = ("bar" :> int @ local);;
+                                   ^
+Error: Syntax error
```
tdelvecchio-jsc, line 158: Error message regression?

typing/typemode.ml
```diff
@@ -139,7 +139,8 @@ let untransl_mode_annots ~loc (modes : Mode.Alloc.Const.Option.t) =
   in
   let yielding = print_to_string_opt Mode.Yielding.Const.print modes.yielding in
   List.filter_map
-    (fun x -> Option.map (fun s -> { txt = Parsetree.Mode s; loc }) x)
+    (fun x ->
+      Option.map (fun s -> { txt = Parsetree.Mode s; loc = Location.none }) x)
```
tdelvecchio-jsc, line 143: Why this change?

tdelvecchio-jsc (changes requested):
It's a bit sad that this PR is 3 different changes in one, but all in heavily overlapping code:
- switching `local_ x @ portable` to `x @ local portable`
- switching `(x : t @@ local)` to `(x : t @ local)`
- changes some logic related to curry syntax? not sure what this is about.

Alas, thing at least looked reasonable on my first pass. Left some comments.

riaqn:
Thanks for the review. 

> It's a bit sad that this PR is 3 different changes in one, but all in heavily overlapping code

Yeah I agree this is sad; but also due to the overlapping, it wouldn't make sense to separate them into PRs.

jane/doc/extensions/modes/syntax.md
```diff
@@ -80,55 +39,87 @@ foo:a * b @ modes -> b @ modes
...
+Again, one should pay attention to the precedences. In the following example, the first
+`modes` annotates `'b`, while the second annotates `x`.
 ```ocaml
-type r = { x : string -> string @@ (local -> unique) global}
+let foo (x : 'a -> 'b @ modes) = ..
+let foo (x : ('a -> 'b) @ modes) = ..
 ```
-which is equivalent to
+
+## Let bindings
```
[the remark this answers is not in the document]
riaqn: > How is this different than function parameters?

Function parameters can have tilda and question mark, which complicates things. Also, the two are handled separately in the parser as well, so it's probably easier to document them correspondingly.

riaqn:
Note to myself: add default modalities to `syntax.md`.

parsing/pprintast.ml
riaqn: it seems that `pattern1` is still the most commonly used. `pattern2` is only used in two specific cases. So I think not updating the comments is fine.

parsing/pprintast.ml
riaqn: Ditto. 😄

testsuite/tests/parsetree/source_jane_street.ml
riaqn: added

testsuite/tests/typing-modes/syntax-error.compilers.reference
riaqn: fixed.

typing/typemode.ml
riaqn: because none of the callers of this function has a sensible `loc` to pass.

parsing/pprintast.ml
tdelvecchio-jsc: I think it'd be useful for anyone reading pprintast to know why there's three different `pattern` functions. What does `pattern2` do differently than `pattern1` and `pattern`?

tdelvecchio-jsc (changes requested):

parsing/pprintast.ml
riaqn: added.

tdelvecchio-jsc (approved):

PR #3523, opened 2025-01-29
Shifts with unboxed count

gretay-js:
Add intrinsics for shifts of `int32` and `int64` that differ from `Stdlib` as follows:
(a) Use the same type for count as for the operand
(b) Don't explicitly mask the count

Note that (b) means we probably shouldn't use %builtins because the operation is unsafe and hardware dependent (although both arm64 and amd64 mask the 64-bit shift).

### DIFF
--- a/backend/cmm_builtins.ml
+++ b/backend/cmm_builtins.ml
@@ -60,11 +60,18 @@ let int_of_value arg dbg = Cop (Creinterpret_cast Int_of_value, [arg], dbg)
 
 let value_of_int arg dbg = Cop (Creinterpret_cast Value_of_int, [arg], dbg)
 
-let shift32 op arg count dbg =
+let shift32 make_op arg count dbg =
   assert (size_int = 8);
   let mask = 32 - 1 in
-  let count = Cop (Cand, [count; Cconst_int (mask, dbg)], dbg) in
-  Some (Cop (op, [arg; count], dbg))
+  let count =
+    match count with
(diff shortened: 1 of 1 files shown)


backend/cmm_builtins.ml
```diff
@@ -60,6 +60,12 @@ let int_of_value arg dbg = Cop (Creinterpret_cast Int_of_value, [arg], dbg)
 
 let value_of_int arg dbg = Cop (Creinterpret_cast Value_of_int, [arg], dbg)
 
+let shift32 op arg count dbg =
+  assert (size_int = 8);
+  let mask = 32 - 1 in
+  let count = Cop (Cand, [count; Cconst_int (mask, dbg)], dbg) in
```
xclerc, line 66: Should we avoid the `Cand` if
`count` is a known value?

xclerc (approved):

backend/cmm_builtins.ml
gretay-js: I have just push a commit that adds some optimizations of constants, using existing `Cmm_helpers`. What do you think? Tested manually on a few small examples of int64 and int32.

xclerc: That's what I had in mind. The functions in
`Cmm_helpers` should optimize further if
the count is `0`. By looking at them, it seems
to happen for `lsr_int` and `asr_int`, but not
for `lsl_int`. Should we add it (in another PR)?

PR #3525, opened 2025-01-29
jkind subsumption

liam923:
Improve the implementation of `Jkind.sub_jkind_l` to give a proper subsumption check. The implementation is based on the rules in `jane/doc/proposals/kind-inference.md`. At the moment, this check is quadratic because we don't have a proper comparison function on types. But based on my anecdotal experience building and running tests, this doesn't seem to hurt performance too bad.

### DIFF
--- a/testsuite/tests/parsetree/source_jane_street.ml
+++ b/testsuite/tests/parsetree/source_jane_street.ml
@@ -1116,10 +1116,12 @@ type 'a contended : immutable_data with 'a @@ contended
 type 'a contended_with_int : immutable_data with 'a @@ contended with int
 
 [%%expect{|
-type 'a list : immutable_data with 'a
-type ('a, 'b) either : immutable_data with 'a * 'b
-type 'a contended : immutable_data with 'a @@ contended
-type 'a contended_with_int : immutable_data with 'a @@ contended
+type 'a list : immutable_data with 'a @@ global aliased
+type ('a, 'b) either : immutable_data with 'a * 'b @@ global aliased
+type 'a contended : immutable_data with 'a @@ global aliased contended
+type 'a contended_with_int
+  : immutable_data
+  with 'a @@ global aliased contended
 |}]
 
 (* not yet supported *)
--- a/testsuite/tests/typing-jkind-bounds/composite.ml
+++ b/testsuite/tests/typing-jkind-bounds/composite.ml
@@ -568,7 +568,7 @@ type 'a t : immutable_data = Flat | Nested of 'a t t
 Line 1, characters 0-52:
 1 | type 'a t : immutable_data = Flat | Nested of 'a t t
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-Error: The kind of type "t" is value
+Error: The kind of type "t" is immutable_data
          because it's a boxed variant type.
        But the kind of type "t" must be a subkind of immutable_data
          because of the annotation on the declaration of the type t.
@@ -611,7 +611,7 @@ type ('a : immutable_data) t : immutable_data = Flat | Nested of 'a t t
 Line 1, characters 0-71:
 1 | type ('a : immutable_data) t : immutable_data = Flat | Nested of 'a t t
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-Error: The kind of type "t" is value
+Error: The kind of type "t" is immutable_data
          because it's a boxed variant type.
        But the kind of type "t" must be a subkind of immutable_data
          because of the annotation on the declaration of the type t.
--- a/testsuite/tests/typing-jkind-bounds/subsumption/basics.ml
+++ b/testsuite/tests/typing-jkind-bounds/subsumption/basics.ml
@@ -236,7 +236,7 @@ Error: Signature mismatch:
 module M : sig
   type 'a t : immutable_data with 'a ref
 end = struct
-  type 'a t : mutable_data with 'a @@ many
+  type 'a t : mutable_data with 'a @@ many unyielding
 end
 [%%expect {|
 module M :
--- a/testsuite/tests/typing-jkind-bounds/subsumption/basics.ml
+++ b/testsuite/tests/typing-jkind-bounds/subsumption/basics.ml
@@ -79,8 +79,19 @@ end
 
 module type T = S with type ('a, 'b) t = ('a, 'b) t
 [%%expect {|
-type ('a, 'b) t : immutable_data with 'a with 'b
-module type S = sig type ('a, 'b) t : immutable_data with 'a with 'b end
+type ('a, 'b) t
+  : immutable_data
+  with 'a @@ global aliased
+
+  with 'b @@ global aliased
+module type S =
+  sig
+    type ('a, 'b) t
+      : immutable_data
+      with 'a @@ global aliased
+
+      with 'b @@ global aliased
+  end
 module type T = sig type ('a, 'b) t = ('a, 'b) t end
 |}]
 
@@ -90,7 +101,14 @@ end = struct
   type ('a, 'b) t : immutable_data with 'b
 end
 [%%expect {|
-module M : sig type ('a, 'b) t : immutable_data with 'a with 'b end
+module M :
+  sig
+    type ('a, 'b) t
+      : immutable_data
+      with 'a @@ global aliased
+
+      with 'b @@ global aliased
+  end
 |}]
 
 module M : sig
@@ -105,30 +123,47 @@ Lines 3-5, characters 6-3:
 5 | end
 Error: Signature mismatch:
        Modules do not match:
-         sig type ('a, 'b) t : immutable_data with 'a with 'b end
+         sig
+           type ('a, 'b) t
+             : immutable_data
+             with 'a @@ global aliased
+
+             with 'b @@ global aliased
+         end
        is not included in
-         sig type ('a, 'b) t : immutable_data with 'a end
+         sig type ('a, 'b) t : immutable_data with 'a @@ global aliased end
        Type declarations do not match:
-         type ('a, 'b) t : immutable_data with 'a with 'b
+         type ('a, 'b) t
+           : immutable_data
+           with 'a @@ global aliased
+
+           with 'b @@ global aliased
        is not included in
-         type ('a, 'b) t : immutable_data with 'a
-       The kind of the first is immutable_data with 'a with 'b
+         type ('a, 'b) t : immutable_data with 'a @@ global aliased
+       The kind of the first is immutable_data with 'a @@ global aliased
+         with 'b @@ global aliased
          because of the definition of t at line 4, characters 2-50.
-       But the kind of the first must be a subkind of immutable_data with 'a
+       But the kind of the first must be a subkind of immutable_data
+         with 'a @@ global aliased
          because of the definition of t at line 2, characters 2-42.
 |}]
 
 type ('a, 'b) u : immutable_data with 'b with 'a
 type ('a, 'b) t : immutable_data with 'a = ('a, 'b) u
 [%%expect {|
-type ('a, 'b) u : immutable_data with 'a with 'b
+type ('a, 'b) u
+  : immutable_data
+  with 'a @@ global aliased
+
+  with 'b @@ global aliased
 Line 2, characters 0-53:
 2 | type ('a, 'b) t : immutable_data with 'a = ('a, 'b) u
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-Error: The kind of type "('a, 'b) u" is immutable_data with 'a with 'b
+Error: The kind of type "('a, 'b) u" is immutable_data
+         with 'a @@ global aliased with 'b @@ global aliased
          because of the definition of u at line 1, characters 0-48.
        But the kind of type "('a, 'b) u" must be a subkind of immutable_data
-         with 'a
+         with 'a @@ global aliased
          because of the definition of t at line 2, characters 0-53.
 |}]
 
@@ -140,15 +175,21 @@ end
 
 module type T = S with type ('a, 'b) t = ('a, 'b) t
 [%%expect {|
-type ('a, 'b) t : immutable_data with 'a with 'b
-module type S = sig type ('a, 'b) t : immutable_data with 'a end
+type ('a, 'b) t
+  : immutable_data
+  with 'a @@ global aliased
+
+  with 'b @@ global aliased
+module type S =
+  sig type ('a, 'b) t : immutable_data with 'a @@ global aliased end
 Line 7, characters 23-51:
 7 | module type T = S with type ('a, 'b) t = ('a, 'b) t
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-Error: The kind of type "('a, 'b) t" is immutable_data with 'a with 'b
+Error: The kind of type "('a, 'b) t" is immutable_data
+         with 'a @@ global aliased with 'b @@ global aliased
          because of the definition of t at line 1, characters 0-48.
        But the kind of type "('a, 'b) t" must be a subkind of immutable_data
-         with 'a
+         with 'a @@ global aliased
          because of the definition of t at line 4, characters 2-42.
 |}]
 
@@ -158,7 +199,8 @@ end = struct
   type 'a t : immutable_data with 'a ref
 end
 [%%expect {|
-module M : sig type 'a t : mutable_data with 'a end
+module M :
+  sig type 'a t : mutable_data with 'a @@ global aliased contended end
 |}]
 
 module M : sig
@@ -174,16 +216,20 @@ Lines 3-5, characters 6-3:
 5 | end
 Error: Signature mismatch:
        Modules do not match:
-         sig type 'a t : mutable_data with 'a end
+         sig type 'a t : mutable_data with 'a @@ global aliased contended end
        is not included in
-         sig type 'a t : mutable_data with 'a end
+         sig
+           type 'a t : mutable_data with 'a @@ global many aliased contended
+         end
        Type declarations do not match:
-         type 'a t : mutable_data with 'a
+         type 'a t : mutable_data with 'a @@ global aliased contended
        is not included in
-         type 'a t : mutable_data with 'a
-       The kind of the first is mutable_data with 'a
+         type 'a t : mutable_data with 'a @@ global many aliased contended
+       The kind of the first is mutable_data
+         with 'a @@ global aliased contended
          because of the definition of t at line 4, characters 2-34.
-       But the kind of the first must be a subkind of mutable_data with 'a
+       But the kind of the first must be a subkind of mutable_data
+         with 'a @@ global many aliased contended
          because of the definition of t at line 2, characters 2-40.
 |}]
 
@@ -193,7 +239,8 @@ end = struct
   type 'a t : mutable_data with 'a @@ many
 end
 [%%expect {|
-module M : sig type 'a t : mutable_data with 'a end
+module M :
+  sig type 'a t : mutable_data with 'a @@ global many aliased contended end
 |}]
 
 (* CR layouts v2.8: 'a u's kind should get normalized to just immutable_data *)
@@ -204,7 +251,7 @@ end
 [%%expect {|
 module M :
   sig
-    type ('a : immutable_data) u : immutable_data with 'a
+    type ('a : immutable_data) u : immutable_data with 'a @@ global aliased
     type ('a : immutable_data) t = 'a u
   end
 |}]
@@ -227,7 +274,7 @@ type t : value mod portable with u
 type q : value mod portable with t = { x : t }
 [%%expect {|
 type u
-type t : value mod portable with u
+type t : value mod portable with u @@ global many aliased contended
 type q = { x : t; }
 |}]
 
@@ -237,8 +284,8 @@ type v : value mod portable with t
 type q : value mod portable with t = { x : v }
 [%%expect {|
 type u
-type t : value mod portable with u
-type v : value mod portable with t
+type t : value mod portable with u @@ global many aliased contended
+type v : value mod portable with t @@ global many aliased contended
 type q = { x : v; }
 |}]
 
@@ -252,7 +299,7 @@ type t = private u
 Line 3, characters 0-46:
 3 | type v : immutable_data with u = { value : t }
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-Error: The kind of type "v" is immutable_data with t
+Error: The kind of type "v" is immutable_data with t @@ global aliased
          because it's a boxed record type.
        But the kind of type "v" must be a subkind of immutable_data with u
          because of the annotation on the declaration of the type v.
@@ -274,7 +321,7 @@ Line 2, characters 0-44:
 Error: The kind of type "[ `foo of t ]" is value
          because it's a polymorphic variant type.
        But the kind of type "[ `foo of t ]" must be a subkind of immutable_data
-         with t
+         with t @@ global aliased
          because of the definition of u at line 2, characters 0-44.
 |}]
 
@@ -310,12 +357,12 @@ type t1
 type t2
 module M :
   sig
-    type a : immutable_data with t2
-    type b : immutable_data with t1
+    type a : immutable_data with t2 @@ global aliased
+    type b : immutable_data with t1 @@ global aliased
     val eq : (a, b) eq
   end
 >> Fatal error: Abstract kind with [with]: immutable_data
-with t1
+with t1 @@ global aliased
 Uncaught exception: Misc.Fatal_error
 
 |}]
@@ -335,13 +382,14 @@ Error: Signature mismatch:
        Modules do not match:
          sig type 'a t = 'a end
        is not included in
-         sig type 'a t : immutable_data with 'a end
+         sig type 'a t : immutable_data with 'a @@ global aliased end
        Type declarations do not match:
          type 'a t = 'a
        is not included in
-         type 'a t : immutable_data with 'a
+         type 'a t : immutable_data with 'a @@ global aliased
        The kind of the first is value
          because of the definition of t at line 2, characters 2-36.
-       But the kind of the first must be a subkind of immutable_data with 'a
+       But the kind of the first must be a subkind of immutable_data
+         with 'a @@ global aliased
          because of the definition of t at line 2, characters 2-36.
 |}]
--- a/testsuite/tests/typing-jkind-bounds/subsumption/basics.ml
+++ b/testsuite/tests/typing-jkind-bounds/subsumption/basics.ml
@@ -1,0 +1,347 @@
+(* TEST
+    flags = "-extension layouts_alpha";
+    expect;
+*)
+
+module M : sig
+  type t
+end = struct
+  type t : value with int
+end
+[%%expect {|
+module M : sig type t end
+|}]
+
+module M : sig
+  type _ t
+end = struct
+  type 'a t : value with 'a
+end
+[%%expect {|
+module M : sig type _ t end
+|}]
+
+module M : sig
+  type t : immediate
+end = struct
+  type t : immediate with int
+end
+[%%expect {|
+module M : sig type t : immediate end
+|}]
+
+module M : sig
+  type t : immediate
+end = struct
+  type t : immediate with string
+end
+[%%expect {|
+Lines 3-5, characters 6-3:
+3 | ......struct
+4 |   type t : immediate with string
+5 | end
+Error: Signature mismatch:
+       Modules do not match:
+         sig type t : immutable_data end
+       is not included in
+         sig type t : immediate end
+       Type declarations do not match:
+         type t : immutable_data
+       is not included in
+         type t : immediate
+       The kind of the first is immutable_data
+         because of the definition of t at line 4, characters 2-32.
+       But the kind of the first must be a subkind of immediate
+         because of the definition of t at line 2, characters 2-20.
+|}]
+
+module M : sig
+  type t : float64
+end = struct
+  type t : float64 with int
+end
+[%%expect {|
+module M : sig type t : float64 end
+|}]
+
+type u : immutable_data
+type t : immutable_data with int = u
+[%%expect {|
+type u : immutable_data
+type t = u
+|}]
+
+type ('a, 'b) t : immutable_data with 'b with 'a
+
+module type S = sig
+  type ('a, 'b) t : immutable_data with 'a with 'b
+end
+
+module type T = S with type ('a, 'b) t = ('a, 'b) t
+[%%expect {|
+type ('a, 'b) t : immutable_data with 'a with 'b
+module type S = sig type ('a, 'b) t : immutable_data with 'a with 'b end
+module type T = sig type ('a, 'b) t = ('a, 'b) t end
+|}]
+
+module M : sig
+  type ('a, 'b) t : immutable_data with 'a with 'b
+end = struct
+  type ('a, 'b) t : immutable_data with 'b
+end
+[%%expect {|
+module M : sig type ('a, 'b) t : immutable_data with 'a with 'b end
+|}]
+
+module M : sig
+  type ('a, 'b) t : immutable_data with 'a
+end = struct
+  type ('a, 'b) t : immutable_data with 'b with 'a
+end
+[%%expect {|
+Lines 3-5, characters 6-3:
+3 | ......struct
+4 |   type ('a, 'b) t : immutable_data with 'b with 'a
+5 | end
+Error: Signature mismatch:
+       Modules do not match:
+         sig type ('a, 'b) t : immutable_data with 'a with 'b end
+       is not included in
+         sig type ('a, 'b) t : immutable_data with 'a end
+       Type declarations do not match:
+         type ('a, 'b) t : immutable_data with 'a with 'b
+       is not included in
+         type ('a, 'b) t : immutable_data with 'a
+       The kind of the first is immutable_data with 'a with 'b
+         because of the definition of t at line 4, characters 2-50.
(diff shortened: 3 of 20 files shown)


typing/typedecl.ml
```diff
@@ -1233,9 +1233,18 @@ let check_constraints env sdecl (_, decl) =
    immediate, we should check the manifest is immediate). Also, update the
    resulting jkind to match the manifest. *)
 let narrow_to_manifest_jkind env loc decl =
-  match decl.type_manifest with
-  | None -> decl
-  | Some ty ->
+  match decl.type_manifest, decl.type_kind with
+  | None, _ -> decl
+  | Some _, (Type_record _ | Type_record_unboxed_product _ | Type_variant _ | Type_open)
+    when not (Builtin_attributes.has_or_null_reexport decl.type_attributes)
```
liam923, line 1239: Revisit this after rebase

typing/jkind.mli
```diff
@@ -524,8 +524,17 @@ val get_annotation : 'd Types.jkind -> Parsetree.jkind_annotation option
 (*********************************)
 (* normalization *)
 
+type 'd normalize_mode =
```
goldfirere, line 527: The indices on the two constructors look backwards. I would expect `Require_best` to be happy with either a left or a right, but `Ignore_best` should only work on the left. I'm worried I'm missing something if everything has type-checked with things as written.

goldfirere: Ah, well, this type is used only on the left, anyway (in `normalize`, which takes a `jkind_l`). But then I think that means it doesn't need to be indexed.

typing/jkind.mli
```diff
@@ -639,7 +648,7 @@ val sub_jkind_l :
   jkind_of_type:(Types.type_expr -> Types.jkind_l option) ->
   Types.jkind_l ->
   Types.jkind_l ->
-  (Types.jkind_l, Violation.t) result
+  (unit, Violation.t) result
```
goldfirere, line 651: Please update the comment, which says this returns an updated jkind.

testsuite/tests/typing-jkind-bounds/subsumption/functors.ml
```diff
@@ -0,0 +1,95 @@
...
+|}]
+
+module F (M : sig type 'a t end) = struct
+  type 'a t : immutable_data with 'a M.t
+end
+
+module Ref = struct
+  type 'a t = 'a ref
+end
+type 'a t : mutable_data with 'a = 'a F(Ref).t
```
goldfirere, line 43: Maybe try this test with `immutable_data`, which should fail. And also missing the `with 'a`, which should also fail.

testsuite/tests/typing-jkind-bounds/subsumption/functors.ml
```diff
@@ -0,0 +1,95 @@
...
+type t = F(Int_abstract).t
+type t = F(Int_abstract).t
+|}]
+
+type t : value mod portable = F(Int_abstract).t
+[%%expect {|
+Line 1, characters 0-47:
+1 | type t : value mod portable = F(Int_abstract).t
+    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Error: The kind of type "F(Int_abstract).t" is immediate with Int_abstract.t
```
goldfirere, line 89: Not for this patch, but this printing is not helpful for users, who might (quite easily) worry about the difference between `immediate` and `value` between the parts of this error message... even though that bit is completely fine.

testsuite/tests/typing-jkind-bounds/subsumption/quality.ml
```diff
@@ -0,0 +1,251 @@
...
+         type 'a t
+       is not included in
+         type 'a t : immutable_data with 'a
+       The kind of the first is value
+         because of the definition of t at line 4, characters 2-11.
+       But the kind of the first must be a subkind of immutable_data with 'a
+         because of the definition of t at line 2, characters 2-36.
+|}]
+
+(* This appears sound to accept. But it isn't. See the following test. *)
```
goldfirere, line 33: This doesn't appear sound to be. In the struct, we have `t : value`, with no mode-crossing ability whatsoever. But not so in the sig. Maybe there is a different relationship you meant to test?

goldfirere (commented):
Partial review. More to come.

typing/jkind.mli
liam923: I think you have the meaning of the indexing backwards. It's specifying the result of normalization, not restricting the input type. There are a few places that require there to be no with-bounds after `Ignore_best` normalization, at this gives us a type assurance that this is the case.

testsuite/tests/typing-jkind-bounds/subsumption/quality.ml
liam923: This is indeed testing what I wanted to - I think maybe this is just more obvious to you, and I should update the comment. I was temporarily under the impression that abstract types could be given `best` if the signature is immediately implemented. But this is clearly wrong.

utils/misc.ml
```diff
@@ -293,6 +293,18 @@ module Stdlib = struct
...
+      let rec find seen_rev = function
+        | [] -> None
+        | hd :: tl ->
+          (match f hd with
+           | true -> Some (hd, seen_rev, tl)
+           | false -> find (hd :: seen_rev) tl)
+      in
+      match find [] l with
+      | Some (e, prefix_rev, suffix) -> Some (e, (List.rev prefix_rev) @ suffix)
+      | None -> None
```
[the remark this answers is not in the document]
liam923: I can actually just delete this function. This was an artifact of an earlier attempt at the subsumption check and is unused now.

testsuite/tests/typing-jkind-bounds/subsumption/functors.ml
liam923: I agree. I have a cr in `sub_jkind_l` that suggests we report which axes disagree, and what the bounds are along those axes.

typing/jkind.mli
liam923: In fact, making this type indexed helped us discover a bug in normalization.

testsuite/tests/parsetree/source_jane_street.ml
```diff
@@ -1068,8 +1068,8 @@ type 'a contended_with_int : immutable_data with 'a @@ contended with int
 [%%expect{|
 type 'a list : immutable_data with 'a
 type ('a, 'b) either : immutable_data with 'a * 'b
-type 'a contended : immutable_data with 'a @@ contended
-type 'a contended_with_int : immutable_data with 'a @@ contended with int
+type 'a contended : immutable_data with 'a
```
goldfirere: This looks wrong -- I think we need the `@@ contended` to remain.

testsuite/tests/typing-jkind-bounds/composite.ml
```diff
@@ -523,7 +523,7 @@ type 'a t : immutable_data = Flat | Nested of 'a t t
 Line 1, characters 0-52:
 1 | type 'a t : immutable_data = Flat | Nested of 'a t t
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-Error: The kind of type "t" is value
+Error: The kind of type "t" is immutable_data
```
goldfirere, line 526: This error message is pretty bad now. And now that we have normalization and subsumption, it's probably time to remove the "don't print on alpha" trick. (Though I could be wrong there.)

goldfirere:
Note to self: I've reviewed through "Promote tests" (except for the files that GitHub knows I haven't finished yet)

testsuite/tests/typing-jkind-bounds/subsumption/modalities.ml
```diff
@@ -0,0 +1,320 @@
...
+  type 'a t : immutable_data with 'a with 'a @@ portable
+end
+[%%expect {|
+Lines 3-5, characters 6-3:
+3 | ......struct
+4 |   type 'a t : immutable_data with 'a with 'a @@ portable
+5 | end
+Error: Signature mismatch:
+       Modules do not match:
+         sig type 'a t : immutable_data with 'a end
```
goldfirere, line 198: This is a pretty terrible error message. I think there might be a bug here

testsuite/tests/typing-jkind-bounds/subsumption/modalities.ml
```diff
@@ -0,0 +1,320 @@
...
+2 | type 'a t : value mod global = 'a u
+    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Error: The kind of type "'a u" is value mod uncontended with 'a
+         because of the definition of u at line 1, characters 0-51.
+       But the kind of type "'a u" must be a subkind of value mod global
+         because of the definition of t at line 2, characters 0-35.
+|}]
+
+module M : sig
+  type 'a t : immutable_data with 'a @@ contended portable
```
goldfirere, line 234: Can we try this the other way around? That way should work.

testsuite/tests/typing-jkind-bounds/subsumption/modalities.ml
```diff
@@ -0,0 +1,320 @@
...
+  type 'a t : immutable_data with 'a @@ contended with 'a @@ portable
+end
+[%%expect {|
+Lines 3-5, characters 6-3:
+3 | ......struct
+4 |   type 'a t : immutable_data with 'a @@ contended with 'a @@ portable
+5 | end
+Error: Signature mismatch:
+       Modules do not match:
+         sig type 'a t : immutable_data with 'a end
```
goldfirere, line 245: Another poor error message

testsuite/tests/typing-jkind-bounds/subsumption/modalities.ml
```diff
@@ -0,0 +1,320 @@
...
+  type ('a, 'b) t : immutable_data with 'a @@ contended with 'b @@ portable
+end
+[%%expect {|
+Lines 3-5, characters 6-3:
+3 | ......struct
+4 |   type ('a, 'b) t : immutable_data with 'a @@ contended with 'b @@ portable
+5 | end
+Error: Signature mismatch:
+       Modules do not match:
+         sig type ('a, 'b) t : immutable_data with 'a with 'b end
```
goldfirere, line 286: More bad error message

goldfirere (commented):
A bunch more review. Still more to go! (I have not yet reviewed constraint.ml, fuel.ml, quality.ml, recursive.ml, or the main payload in jkind.ml.)

testsuite/tests/typing-jkind-bounds/subsumption/constraint.ml
goldfirere, line 1: None of these tests introduce new variables in the constraint. I would want to see a few tests with things like `type 'a t = ... constraint 'a = 'b option` or something, where the `'b` is introduced only in the constraint.

testsuite/tests/typing-jkind-bounds/subsumption/fuel.ml
```diff
@@ -0,0 +1,119 @@
...
+[%%expect {|
+type 'a my_list = 'a list = [] | (::) of 'a * 'a my_list
+type t = int list my_list list my_list list
+|}]
+type t : immutable_data = int list list list list list list
+[%%expect {|
+Line 1, characters 0-59:
+1 | type t : immutable_data = int list list list list list list
+    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Error: The kind of type "int list list list list list list" is immutable_data
```
goldfirere, line 23: As elsewhere, I'm surprised this doesn't print out more information about the kinds

testsuite/tests/typing-jkind-bounds/subsumption/fuel.ml
```diff
@@ -0,0 +1,119 @@
+(* TEST
+    flags = "-extension layouts_alpha";
+    expect;
+*)
+
+type 'a my_list = 'a list = [] | ( :: ) of 'a * 'a my_list
+
+(* At the time this test is written, "fuel" is used to deal with recursive types. Two
```
goldfirere, line 8: I think I don't understand this comment. The "below types"... what types specifically? The two types `t`, one of which is accepted and one of which isn't? They don't look equal to me: the first has 5 `list`s while the second has 6.

I think mostly this file is evidence that the fuel mechanism is not working out so well and that maybe we need to rethink. Or maybe instead that we shouldn't care so much. Not sure.

goldfirere: If we do decide to keep this around, it might be worth getting this file to be generated, because we want the same types to be used throughout. There exist other tests that are generated. I could try to do some sleuthing to find them, but Chris probably knows off the top of his head.

testsuite/tests/typing-jkind-bounds/subsumption/quality.ml
goldfirere: I'll admit being completely confused about what this file is testing, because there were lots of subsumption checks between e.g. `value <: value mod global with a` -- which to my eyes clearly shouldn't work. But of course it only doesn't work because we (rightly) refuse to expand `a`. This only became apparent to me later when the type in the with-bounds is more complex, so I had to think harder.

Bottom line: your response above is right. Abstract types "obviously" don't expand, and so I didn't see what was even being tested. But this was my mistake.

testsuite/tests/typing-jkind-bounds/subsumption/quality.ml
```diff
@@ -0,0 +1,286 @@
...
+         type t : value mod uncontended portable with M.u
+       The kind of the first is value mod portable
+         because of the definition of t at line 15, characters 2-29.
+       But the kind of the first must be a subkind of
+         value mod uncontended portable with M.u
+         because of the definition of t at line 13, characters 2-50.
+|}]
+
+module M : sig
+  type a = [`a of string | `b]
```
goldfirere, line 204: I actually think it would be fine to accept this: there is no row variable here (no `<` or `>` in the poly-variant). But I would expect to reject for now. Maybe leave a comment saying it would be fine to accept -- and then also test the cases with row variables.

testsuite/tests/typing-jkind-bounds/subsumption/quality.ml
```diff
@@ -0,0 +1,286 @@
...
+       is not included in
+         type t : value mod global with a
+       The kind of the first is value
+         because of the definition of t at line 6, characters 2-8.
+       But the kind of the first must be a subkind of value mod global with a
+         because of the definition of t at line 3, characters 2-34.
+|}]
+
+module M : sig
+  type a = < value : string >
```
goldfirere, line 235: Another case where I think it's safe to accept. No substitution is going to make `< value : string >` not be a `value`.

testsuite/tests/typing-jkind-bounds/subsumption/quality.ml
```diff
@@ -0,0 +1,286 @@
...
+  type t : value mod portable with gadt
+end = struct
+  type t
+end
+[%%expect {|
+type gadt = Foo : int -> gadt
+module M : sig type t end
+|}]
+
+(* CR layouts v2.8: gadts shouldn't be "best". But maybe they should track quality along
```
goldfirere, line 275: I agree that it is both sound and additionally expressive to track quality along axes separately. (This is actually a shortcoming in the original typing rules, which did not reflect on this possibility.) But let's not, until we are sufficiently motivated.

testsuite/tests/typing-jkind-bounds/subsumption/quality.ml
```diff
@@ -0,0 +1,286 @@
...
+         type t : value mod global with a
+       The kind of the first is value
+         because of the definition of t at line 6, characters 2-8.
+       But the kind of the first must be a subkind of value mod global with a
+         because of the definition of t at line 3, characters 2-34.
+|}]
+
+(* CR layouts v2.8: gadts shouldn't be "best" because we intend to give them more refined
+   jkinds in the future. So this program will error in the future. *)
+type gadt = Foo : int -> gadt
```
goldfirere, line 264: Would be cool to repeat this test with `[@@unboxed]`.

testsuite/tests/typing-jkind-bounds/subsumption/quality.ml
goldfirere, line 1: This file tests a bunch of cases where we have `not_best`. There should probably be at least one test per type form, even the `best` ones. (So a test with an arrow, a test with a tuple, etc.) Can we engineer a test with a `Tpoly`? I bet so, maybe using an `[@@unboxed]` record.

I'd also want a few tests in here with `private`. Right now, `private` is transparent to the kind system. But it won't always be the case, and when we make `private` more opaque, we'll want to make sure that `private` abbreviations have `not_best` kinds.

testsuite/tests/typing-jkind-bounds/subsumption/recursive.ml
```diff
@@ -0,0 +1,223 @@
...
+3 |   | Cons of 'a * 'a my_list my_list my_list my_list my_list my_list my_list my_list my_list my_list my_list my_list
+                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Error: Layout mismatch in final type declaration consistency check.
+       This is most often caused by the fact that type inference is not
+       clever enough to propagate layouts through variables in different
+       declarations. It is also not clever enough to produce a good error
+       message, so we'll say this instead:
+         The layout of 'a my_list my_list my_list my_list my_list my_list
+                       my_list my_list my_list my_list my_list is any
+           because the .cmi file for my_list is missing.
```
goldfirere, line 38: Zoiks. This is pretty awful.

testsuite/tests/typing-jkind-bounds/subsumption/recursive.ml
```diff
@@ -0,0 +1,223 @@
...
+         But the layout of 'a my_list my_list my_list my_list my_list my_list
+                           my_list my_list my_list my_list my_list must be a sublayout of value
+           because it instantiates an unannotated type parameter of my_list,
+           chosen to have layout value.
+         No .cmi file found containing my_list.
+       A good next step is to add a layout annotation on a parameter to
+       the declaration where this error is reported.
+|}]
+
+type 'a mutable_list : mutable_data with 'a = Nil | Cons of 'a ref * 'a mutable_list
```
goldfirere, line 48: Would be good to check that this also works with a `mutable` field (instead of via `ref`)

testsuite/tests/typing-jkind-bounds/subsumption/recursive.ml
goldfirere, line 1: Is there a test like this here?

```ocaml
type 'a t : immutable_data = Leaf | Node of int * 'a t
```

? The `'a` never appears in a field.

Another test worth having:

```ocaml
type t1
type 'a t2 : immutable_data with 'a with t1 = Leaf of 'a | Node of 'a * t1 t2
```

That's (rather benign) irregular recursion, and so maybe we reject somehow, but it's worth adding the test.

goldfirere (commented):
All tests reviewed 🎉

goldfirere:
After reviewing the tests, I'm pleased to report that I didn't spot any soundness bugs! 🔉 Now onto the actual implementation, which I can happily lower my level of scrutiny, as I'm pretty sure it's basically correct via the tests.

typing/jkind.ml
```diff
@@ -1457,94 +1466,105 @@ module Jkind_desc = struct
...
-          join_respecting_omit Bounds.max, With_bounds_types.of_list bs
-        | Skip -> loop ctl bounds_so_far bs (* skip [b] *)
-        | Continue ctl_after_unpacking_b -> (
-          match jkind_of_type ty with
-          | Some b_jkind ->
-            if (match b_jkind.quality with Not_best -> false | Best -> true)
-               || not require_best
-            then
+        match Axis_set.is_empty ti.relevant_axes with
+        | true -> loop ctl bounds_so_far bs
```
goldfirere, line 1490:
```suggestion
        | true -> loop ctl bounds_so_far bs (* [ty] matters only for axes that are already [max]; drop from the with-bounds *)
```

typing/jkind.ml
```diff
@@ -1457,94 +1466,105 @@ module Jkind_desc = struct
...
-                if require_best
-                then bounds_so_far
-                else join_respecting_omit Bounds.max
+              let bounds_so_far, nested_with_bounds, fuel_result1 =
+                loop new_ctl bounds_so_far (With_bounds.to_list b_with_bounds)
               in
-              let bounds_so_far, bs' =
-                loop ctl_after_unpacking_b bounds_so_far bs
+              (* CR: which ctl to use?
+                 (* Use *original* ctl here, so we don't fall over on
```
goldfirere, line 1518: Especially with the stress on `original`, it's a little confusing using `new_ctl` (the `new` bit). At this point in my read, I don't fully understand this yet so can't suggest a concrete improvement.

typing/jkind.ml
```diff
@@ -1457,94 +1466,105 @@ module Jkind_desc = struct
...
+          | Skip -> loop ctl bounds_so_far bs (* skip [b] *)
+          | Continue ctl_after_unpacking_b -> (
+            match jkind_of_type ty with
+            | Some b_jkind ->
+              found_jkind_for_ty ctl_after_unpacking_b b_jkind.jkind.mod_bounds
+                b_jkind.jkind.with_bounds b_jkind.quality
+            | None ->
+              (* kind of b is not principally known, so we treat it as having the max
+                 bound (only along the axes we care about for this type!) *)
+              found_jkind_for_ty ctl_after_unpacking_b Bounds.max No_with_bounds
```
goldfirere, line 1544: Might be hard to do, but I don't think there are any tests for this case.

typing/jkind.ml
```diff
@@ -2608,70 +2645,97 @@ let sub_or_error ~type_equal ~jkind_of_type t1 t2 =
...
+  let find t ty ~type_equal =
+    (* CR layouts v2.8: maybe it's worth memoizing using a Best_effort_type_map.t? *)
+    Best_effort_type_map.to_seq t
+    |> Seq.fold_left
+         (fun acc (ty2, data) ->
+           match type_equal ty ty2 with true -> data :: acc | false -> acc)
+         []
+end
+
 (* CR layouts v2.8: Rewrite this to do the hard subjkind check from the
```
goldfirere, line 2674: That's what this patch does!

typing/jkind.ml
```diff
@@ -2608,70 +2645,97 @@ let sub_or_error ~type_equal ~jkind_of_type t1 t2 =
...
+  let ( ({ layout = _;
+           mod_bounds = sub_upper_bounds;
+           with_bounds = No_with_bounds
+         } :
+          Allowance.right_only jkind_desc),
+        _ ) =
+    (* CR layouts v2.8: don't expand along bounds where the upper bound is max on the
+       right. Normalization *)
+    (* MB_EXPAND_L *)
+    Jkind_desc.map_normalize ~jkind_of_type ~mode:Ignore_best
```
goldfirere, line 2720: This step could use a little commentary. I don't quite see what it's doing.

goldfirere:
OK. I have reviewed everything in this patch up to now. Looking pretty good!

typing/jkind.ml
```diff
@@ -1363,8 +1344,32 @@ module Jkind_desc = struct
...
+      | Sufficient_fuel
+
+    let both a b =
+      match a, b with
+      | Ran_out_of_fuel, _ | _, Ran_out_of_fuel -> Ran_out_of_fuel
+      | Sufficient_fuel, Sufficient_fuel -> Sufficient_fuel
+  end
+
+  (* Normalize the jkind. If mode is Require_best, only jkinds that are best will be used.
+     If mode is Ignore_best, then Not_best will be used. Since Ignore_best can use
```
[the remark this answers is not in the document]
liam923: The `map_type_info` parameter enables us to do the stepping that you're referring to - I'll expand on this on your commend at the usage site.

typing/jkind.ml
liam923: The comment stating "original" is an old comment from when we really did use the original one. I think aspen's pr might also make this same change, so I think the discussion over which ctl to use should happen there.

typing/jkind.ml
liam923: I'll add a comment. But because I'm in a hurry right now, a brief explanation: This does the incremental stepping. The left is `Ignore_best` normalized, but before each type is expanded, we look for occurrences of it on the right. If it does occur on the right, then we skip over it.

To go into some more detail on "skipping over": We only skip over axes that are "relevant" to the rhs. ie, if `@@ portable` is on the type on the rhs, we don't skip it. This is why we take the diff of the axes relevant to the left - the axes relevant to the right. And if the resulting set is empty, `map_normalize` is smart enough to not bother expanding the type.

testsuite/tests/typing-jkind-bounds/subsumption/quality.ml
goldfirere, line 1: I don't think there are any tests here showing that the kind of an extensible type (e.g. `type t = ..`) is `best`. Please add some.

testsuite/tests/parsetree/source_jane_street.ml
liam923: Yes I agree

testsuite/tests/typing-jkind-bounds/subsumption/modalities.ml
liam923: I think this might be the same as the printing bug where `@@ contended` was dropped.

testsuite/tests/typing-jkind-bounds/subsumption/fuel.ml
liam923: I don't think I follow what information you expected to see here. Although I do think the error message is bad due to the `because of the definition of t at line 1`. I think it should instead say something about the kind annotation on `t`.

testsuite/tests/typing-jkind-bounds/subsumption/fuel.ml
liam923: The numbers are different because I messed up - they're both meant to have 5 `list`s. I'll fix.

typing/jkind.mli
liam923: Ah, I see what you mean. Yes, you're right that this shouldn't leak into the mli. I'll fix.

testsuite/tests/typing-jkind-bounds/subsumption/basics.ml
```diff
@@ -0,0 +1,213 @@
...
+  type 'a t : value with 'a
+end
+[%%expect {|
+module M : sig type _ t end
+|}]
+
+module M : sig
+  type t : immediate
+end = struct
+  type t : immediate with int
```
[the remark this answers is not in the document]
liam923: done

testsuite/tests/typing-jkind-bounds/subsumption/basics.ml
```diff
@@ -0,0 +1,213 @@
...
+         because of the definition of t at line 1, characters 0-48.
+       But the kind of type "('a, 'b) t" must be a subkind of immutable_data
+         with 'a
+         because of the definition of t at line 4, characters 2-42.
+|}]
+
+module M : sig
+  type 'a t : mutable_data with 'a
+end = struct
+  type 'a t : immutable_data with 'a ref
```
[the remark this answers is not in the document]
liam923: Done. The other way was actually pretty interesting. Just flipping this actually causes a rejection. This is because `'a ref` is always `many`, because at the moment a `mutable` field has an implied `@@ many` modality.

testsuite/tests/typing-jkind-bounds/subsumption/modalities.ml
```diff
@@ -0,0 +1,311 @@
...
+         because of the definition of t at line 1, characters 0-77.
+       But the kind of type "('a, 'b) t" must be a subkind of
+         value mod portable
+         because of the definition of t at line 4, characters 2-38.
+|}]
+
+module M : sig
+  type 'a t : immutable_data with 'a @@ portable
+end = struct
+  type 'a t : immutable_data with 'a @@ portable with 'a
```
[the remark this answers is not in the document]
liam923: done

testsuite/tests/typing-jkind-bounds/subsumption/modalities.ml
liam923: done

testsuite/tests/typing-jkind-bounds/subsumption/constraint.ml
liam923: done

testsuite/tests/typing-jkind-bounds/subsumption/quality.ml
liam923: done

testsuite/tests/typing-jkind-bounds/subsumption/quality.ml
liam923: done

testsuite/tests/typing-jkind-bounds/subsumption/quality.ml
liam923: done

testsuite/tests/typing-jkind-bounds/subsumption/quality.ml
liam923: done

testsuite/tests/typing-jkind-bounds/subsumption/recursive.ml
liam923: Yeah. I'm tempted to make error messages say something about running out of fuel. But I think that's outside the scope of this PR. I'll leave a CR here about it

testsuite/tests/typing-jkind-bounds/subsumption/recursive.ml
liam923: done

testsuite/tests/typing-jkind-bounds/subsumption/recursive.ml
liam923: I seem to have neglected the former. I've added both

typing/jkind.ml
liam923: I've added a comment that expands on this

typing/jkind.ml
liam923: removed

typing/jkind.ml
liam923: I've committed a slightly different comment that I find to be more clear

typing/jkind.ml
```diff
@@ -352,25 +352,21 @@ module With_bounds = struct
...
 
     let join { relevant_axes = axes1 } { relevant_axes = axes2 } =
       { relevant_axes = Axis_set.union axes1 axes2 }
   end
 
-  let of_with_bounds_types tys =
-    match With_bounds_types.Non_empty.of_maybe_empty tys with
-    | None -> No_with_bounds
-    | Some tys -> With_bounds tys
+  let to_be_map = function
```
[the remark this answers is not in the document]
liam923: I've renamed to `to_best_eff_map`

testsuite/tests/typing-jkind-bounds/subsumption/quality.ml
liam923: done

testsuite/tests/typing-jkind-bounds/subsumption/quality.ml
liam923: agreed

typing/jkind.ml
liam923: I've done this in @glittershark's normalization pr in commit e8190bd6f921f1cc5ac2b9d7cda6738afac61a7d : https://github.com/ocaml-flambda/flambda-backend/pull/3509/commits/e8190bd6f921f1cc5ac2b9d7cda6738afac61a7d

testsuite/tests/parsetree/source_jane_street.ml
liam923: I just fixed this in @glittershark 's normalization pr. The bugfix should also resolve the other modality printing issues you commented on once it gets merged into this branch.

testsuite/tests/typing-jkind-bounds/subsumption/basics.ml
```diff
@@ -0,0 +1,213 @@
...
+end
+[%%expect {|
+module M : sig type 'a t : mutable_data with 'a end
+|}]
+
+module _ = struct
+  type ('a : immutable_data) u : immutable_data with 'a
+  type 'a t : immutable_data = 'a u
```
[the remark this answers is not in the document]
goldfirere: Yes I agree. I hadn't accounted for the `immutable_data` annotation on `'a`.

goldfirere:
I've reviewed commits up through "Improve subsumption comments"

testsuite/tests/typing-jkind-bounds/subsumption/functors.ml
```diff
@@ -0,0 +1,132 @@
...
+module Int_abstract = struct
+  type t = int
+  type u : value mod global
+end
+type t : value mod global = F(Int_abstract).t
+type t : value mod portable with Int_abstract.u = F(Int_abstract).t
+[%%expect {|
+module Int_abstract : sig type t = int type u : value mod global end
+type t = F(Int_abstract).t
+>> Fatal error: Tlink or TSubst encountered in With_bounds_types
```
liam923, line 112: @glittershark do you know what's going on here?

glittershark: urgh, yes, I'll take a look at this. Forgot to call `repr`/ `get_desc` somewhere I'm sure.

goldfirere:
Reviewed most recent commits.

goldfirere:
Reviewed most recent commits.

testsuite/tests/typing-jkind-bounds/subsumption/functors.ml
glittershark: fixed (badly) in my branch.

goldfirere:
Discussed through "Revert best_effort_type_map" with @glittershark. All looks good.

goldfirere (approved):

PR #3532, opened 2025-01-31
`local` implies `yielding`

dkalinichenko-js:
Make `yielding` mode the default if `local` is set. The change is only limited to syntax parsing and does not change the inference code. To preserve consistency, the `global` modality/modifier now also implies `unyielding` by default, and `[@local_opt]` also varies along the `yielding` axis.

Subsequent PRs will add `yielding` annotations to the Capsule API and to the Effects API.

The printer now displays `stack_` allocation return values with `@ unyielding`, as intended.

This change sometimes causes principal and non-principal type inference to diverge when the old syntax for `local_` annotations is used. I'm surprised this compiler option affects syntax parsing, but the issue seems benign.

### DIFF
--- a/testsuite/tests/parsetree/source_jane_street.ml
+++ b/testsuite/tests/parsetree/source_jane_street.ml
@@ -493,6 +493,23 @@ let f ~(x1 @ many)
 
 [%%expect{|
 val f :
+  ('b : value_or_null) ('c : value_or_null) ('d : value_or_null) 'e.
+    x1:'b ->
+    x2:local_ string -> local_
+    (x3:local_ (string -> string) ->
+     x4:local_ ('a. 'a -> 'a) ->
+     x5:local_ 'c ->
+     x6:local_ bool ->
+     x7:local_ bool ->
+     x8:local_ unit ->
+     string ->
+     local_ 'd -> local_
+     'b * string * (string -> string) * ('e -> 'e) * 'c * string * string *
+     int array * string * (int -> local_ (int -> int) @ unyielding) *
+     (int -> local_ (int -> int) @ unyielding) @ contended) @ unyielding =
+  <fun>
+|}, Principal{|
+val f :
   x1:'b ->
   x2:local_ string ->
   x3:local_ (string -> string) ->
@@ -504,8 +521,8 @@ val f :
   string ->
   local_ 'd -> local_
   'b * string * (string -> string) * ('e -> 'e) * 'c * string * string *
-  int array * string * (int -> local_ (int -> int)) *
-  (int -> local_ (int -> int)) @ contended = <fun>
+  int array * string * (int -> local_ (int -> int) @ unyielding) *
+  (int -> local_ (int -> int) @ unyielding) @ contended = <fun>
 |}]
 
 let f1 (_ @ local) = ()
@@ -585,7 +602,14 @@ Error: This value escapes its region.
 let f2 (x @ local) (f @ once) : t2 = exclave_ { x; f }
 
 [%%expect{|
-val f2 : local_ float -> (float -> float) @ once -> local_ t2 @ once = <fun>
+val f2 :
+  local_ float -> local_
+  ((float -> float) @ once -> local_ t2 @ once unyielding) @ unyielding =
+  <fun>
+|}, Principal{|
+val f2 :
+  local_ float -> (float -> float) @ once -> local_ t2 @ once unyielding =
+  <fun>
 |}]
 
 
--- a/testsuite/tests/parsetree/source_jane_street.ml
+++ b/testsuite/tests/parsetree/source_jane_street.ml
@@ -494,22 +494,6 @@ let f ~(x1 @ many)
 [%%expect{|
 val f :
   x1:'b ->
-  x2:local_ string -> local_
-  (x3:local_ (string -> string) ->
-   x4:local_ ('a. 'a -> 'a) ->
-   x5:local_ 'c ->
-   x6:local_ bool ->
-   x7:local_ bool ->
-   x8:local_ unit ->
-   string ->
-   local_ 'd -> local_
-   'b * string * (string -> string) * ('e -> 'e) * 'c * string * string *
-   int array * string * (int -> local_ (int -> int) @ unyielding) *
-   (int -> local_ (int -> int) @ unyielding) @ contended) @ unyielding =
-  <fun>
-|}, Principal{|
-val f :
-  x1:'b ->
   x2:local_ string ->
   x3:local_ (string -> string) ->
   x4:local_ ('a. 'a -> 'a) ->
@@ -520,8 +504,8 @@ val f :
   string ->
   local_ 'd -> local_
   'b * string * (string -> string) * ('e -> 'e) * 'c * string * string *
-  int array * string * (int -> local_ (int -> int) @ unyielding) *
-  (int -> local_ (int -> int) @ unyielding) @ contended = <fun>
+  int array * string * (int -> local_ (int -> int)) *
+  (int -> local_ (int -> int)) @ contended = <fun>
 |}]
 
 let f1 (_ @ local) = ()
@@ -601,14 +585,7 @@ Error: This value escapes its region.
 let f2 (x @ local) (f @ once) : t2 = exclave_ { x; f }
 
 [%%expect{|
-val f2 :
-  local_ float -> local_
-  ((float -> float) @ once -> local_ t2 @ once unyielding) @ unyielding =
-  <fun>
-|}, Principal{|
-val f2 :
-  local_ float -> (float -> float) @ once -> local_ t2 @ once unyielding =
-  <fun>
+val f2 : local_ float -> (float -> float) @ once -> local_ t2 @ once = <fun>
 |}]
 
 
--- a/testsuite/tests/parsetree/source_jane_street.ml
+++ b/testsuite/tests/parsetree/source_jane_street.ml
@@ -493,20 +493,19 @@ let f ~(x1 @ many)
 
 [%%expect{|
 val f :
-  ('b : value_or_null) ('c : value_or_null) ('d : value_or_null) 'e.
-    x1:'b ->
-    x2:local_ string -> local_
-    (x3:local_ (string -> string) ->
-     x4:local_ ('a. 'a -> 'a) ->
-     x5:local_ 'c ->
-     x6:local_ bool ->
-     x7:local_ bool ->
-     x8:local_ unit ->
-     string ->
-     local_ 'd -> local_
-     'b * string * (string -> string) * ('e -> 'e) * 'c * string * string *
-     int array * string * (int -> local_ (int -> int) @ unyielding) *
-     (int -> local_ (int -> int) @ unyielding) @ contended) @ unyielding =
+  x1:'b ->
+  x2:local_ string -> local_
+  (x3:local_ (string -> string) ->
+   x4:local_ ('a. 'a -> 'a) ->
+   x5:local_ 'c ->
+   x6:local_ bool ->
+   x7:local_ bool ->
+   x8:local_ unit ->
+   string ->
+   local_ 'd -> local_
+   'b * string * (string -> string) * ('e -> 'e) * 'c * string * string *
+   int array * string * (int -> local_ (int -> int) @ unyielding) *
+   (int -> local_ (int -> int) @ unyielding) @ contended) @ unyielding =
   <fun>
 |}, Principal{|
 val f :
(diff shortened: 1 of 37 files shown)


testsuite/tests/typing-unique/overwriting.ml
```diff
@@ -176,7 +176,7 @@ let disallowed_by_regionality (local_ unique_ r) x =
 Line 3, characters 16-17:
 3 |   let ref = ref r in
                     ^
-Error: This value escapes its region.
+Error: This value is "yielding" but expected to be "unyielding".
```
dkalinichenko-js, line 179: This error message got worse. I'm not sure what to do about it, since the axis on which we fail submoding is produced by the mode solver.

lpw25: Something looks wrong to me here. The code in `modes.ml` prefers to tell you that the locality is wrong rather than the yielding mode. I think you should debug why it isn't doing that.

typing/ctype.ml
```diff
@@ -1644,22 +1644,37 @@ let instance_label ~fixed lbl =
...
-  | Primitive.Prim_local, _ -> Locality.allow_right Locality.local
+(* CR dkalinichenko: we must vary yieldingness together with locality to get
+   sane behavior around defaults. Remove once we have mode polymorphism. *)
+let prim_mode mvar mvar' = function
+  | Primitive.Prim_global, _ ->
+    Locality.allow_right Locality.global,
+    Yielding.allow_right Yielding.unyielding
+  | Primitive.Prim_local, _ ->
+    Locality.allow_right Locality.local,
+    Yielding.allow_right Yielding.yielding
```
riaqn, line 1655: Is it meaningful to allow users to write `global yielding` or `local unyielding` on primitives?

typing/ctype.ml
```diff
@@ -1778,11 +1795,12 @@ let instance_prim_mode (desc : Primitive.description) ty =
...
-    let mode = Locality.newvar () in
-    let finalret = prim_mode (Some mode) desc.prim_native_repr_res in
+    let mode_l = Locality.newvar () in
+    let mode_y = Yielding.newvar () in
+    let finalret = prim_mode (Some mode_l) (Some mode_y) desc.prim_native_repr_res in
     instance_prim_locals desc.prim_native_repr_args
-      mode (Alloc.disallow_right Alloc.legacy) finalret ty,
-    Some mode
+      mode_l mode_y (Alloc.disallow_right Alloc.legacy) finalret ty,
+    Some mode_l
```
riaqn, line 1803: Here we are dropping `mode_y`, which is unsafe if we allow users to write `global yielding` and `local unyielding` on primitives. (This `instance_prim_mode` is used in `includecore.ml` for value inclusion)

testsuite/tests/typing-unique/overwriting.ml
riaqn: This is probably a mode bug in `Pexp_overwrite` - I'm looking.

typing/ctype.ml
dkalinichenko-js: `local unyielding`, yes. For example, `%obj_dup` on a `local unyielding` value should also return a `local unyielding` value. `global yielding` I don't think makes much sense.

typing/ctype.ml
dkalinichenko-js: Will fix.

testsuite/tests/typing-unique/overwriting.ml
riaqn: I pushed a dirty fix, but probably not ideal. Let's discuss in the slack channel wrt uniqueness and overwriting.

riaqn: As discussed offline, overwriting will not be avaialble to the users for some time, and we will re-do the type checking at some point. So the dirty fix is fine. Resolved.

typing/ctype.ml
dkalinichenko-js: After double-checking, I don't think we need that -- those constructors are only used in the middle-end, which doesn't care about `yielding`.

typing/ctype.ml
dkalinichenko-js: Fixed. I'm not very concerned about unsoundness here (regarding yieldingness, not locality), but can add tests to be sure.

dkalinichenko-js:
@riaqn I'd appreciate help debugging the divergence between principal and non-principal typing here, I can't figure out why it happens.

riaqn:
The principality issue is because some types (for example, string) is `immutable_data` and crosses yielding. Let's discuss offline.

testsuite/tests/parsetree/source_jane_street.ml
```diff
@@ -444,6 +444,22 @@ let f ~(x1 @ many)
   stack_ (x1, x2, x3, x4, x5, x9, x10, x11, (* x12, *) x13, x14, x15)
 
 [%%expect{|
+val f :
+  x1:'b ->
+  x2:local_ string -> local_
```
lpw25, line 449: This is the part that looks wrong. It is printing `a -> b @ local -> (c -> d) @ local` which should always be equivalent to `a -> b @ local -> c -> d` and so should be printed as such.

lpw25: Actually, I see now that it is rather doing printing `a -> b @ local -> (c -> d) @ local unyielding`. (@riaqn We should really not print `local_` when we also need to print `@ foo` for some other `foo`, it is really confusing).

So I think the problem is rather the defaulting logic. It should be defaulting the yielding in the return mode to `yielding` since it cannot default the locality to `global`.

riaqn: > We should really not print local_ when we also need to print @ foo for some other foo, it is really confusing).

Yes #3522 already fixes that.

typing/ctype.ml
```diff
@@ -1663,14 +1663,12 @@ let instance_label ~fixed lbl =
   )
 
 (* CR dkalinichenko: we must vary yieldingness together with locality to get
-   sane behavior around defaults. Remove once we have mode polymorphism. *)
+   sane behavior around [@local_opt]. Remove once we have mode polymorphism. *)
 let prim_mode mvar mvar' = function
   | Primitive.Prim_global, _ ->
-    Locality.allow_right Locality.global,
-    Yielding.allow_right Yielding.unyielding
+    Locality.allow_right Locality.global, Yielding.newvar ()
```
riaqn, line 1669: Hmm I still think we should introduce `Primitive.Prim_global_unyielding` and alikes. But maybe what you wrote is sufficient - let's discuss offline.

typing/jkind.ml
```diff
@@ -1439,6 +1439,24 @@ module Const = struct
...
+      |> function
+      | None -> None
+      | Some modes -> (
+        match List.mem "global" modes, List.mem "unyielding" modes with
+        | true, true ->
+          (* We default [mod global] to [mod global unyielding],
+             so we don't want to print the latter. *)
+          Some (List.filter (fun m -> m <> "unyielding") modes)
+        | true, false ->
+          (* Otherwise, print [mod global yielding] to indicate [yielding]. *)
```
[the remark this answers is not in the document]
dkalinichenko-js: Hm, we don't really need too. Reverted this, now I just add it to the end. Ideally, we'd print in some canonical order?

typing/includecore.ml
```diff
@@ -169,37 +169,37 @@ let value_descriptions ~loc env name
...
-         let ty2_local =
-           let ty2, mode2, _ = Ctype.instance_prim p2 vd2.val_type in
-           Option.iter
-             (fun m -> Mode.Locality.submode_exn Mode.Locality.local m)
-             mode2;
-           ty2
-         in
-         (try Ctype.moregeneral env true ty1_local ty2_local
-          with Ctype.Moregen err -> raise (Dont_match (Type err)));
+         let restrict_locality = [
```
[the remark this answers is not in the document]
dkalinichenko-js: Done.

dkalinichenko-js:
I changed legacy zapping behavior of `yielding` to vary based on areality too. This fixed the differences between principal and non-principal typing, so I'm happy to merge this now.

dkalinichenko-js:
I also changed contents of `lazy`s to always be `unyielding` in addition to `global`. Otherwise, existing code that relies on the contents of a local `lazy_t` being at the legacy mode would break.

typing/jkind.ml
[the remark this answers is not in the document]
riaqn: Yeah ideally this function `get_modal_bounds` should return (instead of a list) a record, each field representing an axis. Until then, I don't see a clean way to maintain a canonical order.

typing/includecore.ml
```diff
@@ -169,26 +169,20 @@ let value_descriptions ~loc env name
...
-           (fun m -> Mode.Locality.submode_exn Mode.Locality.local m)
-         ] in
-         let restrict_yielding = [
-           (fun m -> Mode.Yielding.submode_exn m Mode.Yielding.unyielding);
-           (fun m -> Mode.Yielding.submode_exn Mode.Yielding.yielding m)
-         ] in
-         List.iter (fun restrict_loc ->
-           List.iter (fun restrict_yield ->
+         let locality = [ Mode.Locality.global; Mode.Locality.local ] in
+         let yielding = [ Mode.Yielding.unyielding; Mode.Yielding.yielding ] in
```
riaqn, line 173: Sorry to keep pushing this, then now I think about this, it might be even better to have e.g. `Mode.Locality.Const.all` which is a list of all the constants? I'm slightly worried that we might introduce more modes in the axis and we forgot to update the check here.

dkalinichenko-js:
Fixed the issue you pointed out + print order.

testsuite/tests/typing-modes/yielding.ml
```diff
@@ -78,3 +72,197 @@ Line 2, characters 45-46:
...
+[%%expect{|
+external id : ('a [@local_opt]) -> ('a [@local_opt]) = "%identity"
+val f1 : 'a -> 'a = <fun>
+val f2 : local_ 'a -> local_ 'a = <fun>
+val f3 : 'a @ yielding -> 'a @ yielding = <fun>
+val f4 : local_ 'a @ unyielding -> local_ 'a = <fun>
+|}]
+
+(* Test [instance_prim] + mixed mode annots. *)
+external requires_unyielding : 'a @ local unyielding -> (unit [@local_opt]) = "%ignore"
```
riaqn, line 191: Another test case to add is: `'a @ local yielding [@local_opt]` - both `local` and `yielding` are overwritten by the `local_opt` without warning. We don't need to fix it since it was already broken, but should add a test case showing that.

typing/ctype.ml
```diff
@@ -1796,18 +1816,19 @@ let instance_prim_mode (desc : Primitive.description) ty =
   let is_poly = function Primitive.Prim_poly, _ -> true | _ -> false in
   if is_poly desc.prim_native_repr_res ||
        List.exists is_poly desc.prim_native_repr_args then
-    let mode = Locality.newvar () in
-    let finalret = prim_mode (Some mode) desc.prim_native_repr_res in
+    let mode_l = Locality.newvar () in
+    let mode_y = Yielding.newvar () in
```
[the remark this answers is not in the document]
dkalinichenko-js: Yes, that's intended. In cases where `local_opt` is fine, varying `yielding` is fine too -- none of the primitives we use it for right now care about `yielding`.

typing/ctype.mli
```diff
@@ -222,11 +222,13 @@ val instance_label:
         _ gen_label_description -> type_expr list * type_expr * type_expr
         (* Same, for a label *)
 val prim_mode :
-        (Mode.allowed * 'r) Mode.Locality.t option -> (Primitive.mode * Primitive.native_repr)
+        (Mode.allowed * 'r) Mode.Locality.t option
+        -> (Primitive.mode * Primitive.native_repr)
```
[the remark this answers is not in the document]
dkalinichenko-js: Fixed.

typing/ctype.ml
```diff
@@ -1662,22 +1662,40 @@ let instance_label ~fixed lbl =
     (vars, ty_arg, ty_res)
   )
 
-let prim_mode mvar = function
-  | Primitive.Prim_global, _ -> Locality.allow_right Locality.global
-  | Primitive.Prim_local, _ -> Locality.allow_right Locality.local
+(* CR dkalinichenko: we must vary yieldingness together with locality to get
+   sane behavior around [@local_opt]. Remove once we have mode polymorphism. *)
+let prim_mode' mvar mvar' = function
```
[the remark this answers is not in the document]
dkalinichenko-js: Done.

testsuite/tests/typing-modes/yielding.ml
dkalinichenko-js: Done.

typing/ctype.ml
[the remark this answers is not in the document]
dkalinichenko-js: (See tests in `yielding.ml` that test this behavior).

PR #3533, opened 2025-02-02
fix build on arm64 linux

matthewelse:
arm64 linux builds were previously failing due to a mismatching END_FUNCTION call. This patch trivially fixes this.

Example of a previous build failure:

```
# /tmp/build_65ff2e_dune/cc9ksxcR.s: Assembler messages:
# /tmp/build_65ff2e_dune/cc9ksxcR.s: Error: .size expression for caml_call_gc does not evaluate to a constant
```


gretay-js (approved):

matthewelse:
Can we merge this? (I don't have permissions to)

PR #3536, opened 2025-02-03
Prefer generic mode error message for `stack_`

riaqn:
`stack_ exp` could trigger mode error for two reasons:
- The surroudning expression expects it to be `global`. For example, `ref (stack_ (42, 42))`. This should give the same error message as a typical mode error.
- The `exp` itself cannot be stack allocated for reasons of its own. There are two cases this can happen as follows. Ideally we want the error message to be more detailed, but that's difficult. Currently it just gives `Cannot stack allocate` for both cases.
  -  Uncurried functions. `stack_ (fun x y -> x : 'a -> 'a -> 'a)` (the inner function is `global`, therefore the outer function has to be `global` too).
  - Explicit mode annotation. E.g., `stack_ ((42, 42) : _ @@ global)`.

**_What this PR does:_**
It's possible that `stack_ exp` fails for both reasons, in which case we prefer the first reason for error messages.

### DIFF
--- a/typing/typecore.ml
+++ b/typing/typecore.ml
@@ -6821,11 +6821,17 @@ and type_expect_
       | Texp_record {alloc_mode = Some alloc_mode; _}
       | Texp_array (_, _, _, alloc_mode)
       | Texp_field (_, _, _, Boxing (alloc_mode, _), _) ->
-        begin match Locality.submode Locality.local
-          (Alloc.proj (Comonadic Areality) alloc_mode.mode) with
-        | Ok () -> ()
-        | Error _ -> raise (Error (e.pexp_loc, env,
-            Cannot_stack_allocate alloc_mode.locality_context))
+        begin
+          submode ~loc ~env
+            (Value.min_with (Comonadic Areality) Regionality.local)
+            expected_mode;
+          match
+            Locality.submode Locality.local
+              (Alloc.proj (Comonadic Areality) alloc_mode.mode)
+          with
+          | Ok () -> ()
+          | Error _ -> raise (Error (e.pexp_loc, env,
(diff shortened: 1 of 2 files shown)


typing/typecore.ml
```diff
@@ -6786,18 +6785,28 @@ and type_expect_
...
-          (Alloc.proj (Comonadic Areality) alloc_mode.mode) with
-        | Ok () -> ()
-        | Error _ -> raise (Error (e.pexp_loc, env,
-            Cannot_stack_allocate alloc_mode.locality_context))
-        end
+          submode ~loc ~env
+            (Value.min_with (Comonadic Areality) Regionality.local)
+            expected_mode;
+          Locality.submode_exn Locality.local
+            (Alloc.proj (Comonadic Areality) alloc_mode.mode)
```
lpw25, line 6809: Are there any tests of this path? It looks like you've changed the error message, but I don't see a corressponding test output change.

lpw25 (approved):
Looks good to me, although maybe it indicates we're missing a test.

typing/typecore.ml
riaqn: The original code type check `Texp_tuple` with expected_mode being local. That should make `Texp_tuple.alloc_mode` extracted here to be `local` and therefore the original `Cannot_stack_allocate` will never trigger. It's the other submode below that might fail.

The new code change the always-successful `submode` to `submode_exn`. The behavior doesn't change.

riaqn: Ugh, nvm, my assumption is wrong - I will fix it.

riaqn: Yeah my previously assumption was wrong. I've rewritten the PR and it's now much smaller. Please see the updated description and review by commit.

Issue #472, opened 2022-01-18
Join loses sharing, causes OOM


lukemaurer:
In the Jane Street tree, we have a large functor (Merlin reports an expanded module type that's 5400 lines long). It has several sub-modules that reference each other heavily.
```
module Make(Config : Config.S) : S = struct
  module Config = Config
  module My_foo = Foo(Config)
  module My_bar = Bar(My_foo)
  
  (* ... many, many more things ... *)
end
```
`Config.S` is basically a record type with 18 small fields, but `Foo` and `Bar` are pretty substantial functors in their own right. Then the driver comes along and says:

```
module Implementations = struct
  module I1 = Make(Config1)
  module I2 = Make(Config2)
  module I3 = Make(Config3)
end

let run n =
  let module Implementation =
    (val
      match n with
      | 1 -> (module Implementations.I1 : S)
      | 2 -> (module Implementations.I2)
      | _ -> (module Implementations.I3))
  in
  ...
```

Up to `run`, all of this compiles fine: with aggressive inlining, the three functor calls in `Implementations` generate _a lot_ of code, but nothing we can't handle (in about 30 s). The `let module`, however, is trouble: It produces a join point whose argument type is `I1 ∨ I2 ∨ I3`, which is a three-way join over the _entire_ signature `S`. Worse, join operations produce expanded types that have no sharing: every reference to `Config` in `Foo` becomes a new record type with 18 fields, and then every reference to `My_foo` in `Bar` becomes a _giant_ new record type _in which `Config` has been expanded many times_. The result of all this (with `Foo` doing something similar on its own) is an OOM on a 96G machine.

(For reasons I haven't quite pinned down, the OOM only happens if the max inlining depth is at least 9, even though I'm running this with functor return types, so in principle the types should get huge even without inlining. My working theory is that inlining often causes functions to have more free variables, and thus bigger closures: a function where `I1` occurs free, for instance, might become one where `I1` and `I1.My_bar` _separately_ occur free, probably ~doubling the size of the fully-expanded type of the closure.)

[...]

Issue #3300, opened 2024-11-22
Improve join performance


bclement-ocp:
Turns out we spent quite some time joining environments (when that is enabled, to `-O2` and up), up to 30% of compilation times (even 60% on rare outliers), which is a lot. It is expected that joining would be expensive, but this sounds like a bit much. #3299 should be able to give us more data.

PR #474, opened 2022-01-19
Limit depth of recursion for join of types sharing structure

lthls:
This PR is aimed at limiting the amount of duplication one can get when joining two structures that were created in a similar way from different sets of input values, containing several references to these inputs.

For instance:
```ocaml
let  make_struct v =
  (v, (v, v))

let v1 = (0, 1)
let v2 = (2, 3)
let r = if foo then make_struct v1 else make_struct v2
```

Setting the new `join_depth` parameter to 0 should give `r` a type looking like `(T, T)`, with depth 1 it should get `((T, T), (T, T))`, with depth 2 it should get `(({0, 2}, {1, 3}), ((T, T), (T, T)))` and with depth 3 or more it should get the full `(({0, 2}, {1, 3}), (({0, 2}, {1, 3}), ({0, 2}, {1, 3})))` that is computed without this PR.

I still need to test that it works as expected, and I'm not sure yet what the default parameters should be.

PR #3538, opened 2025-02-04
flambda2-types: New n-way join algorithm

bclement-ocp:
The existing join algorithm suffers from several drawbacks:

 - It can be slow due to the use of a quadratic algorithm, taking up to 60% of the total compilation time in -O3 mode in pathological cases (lambda_to_flambda_primitives.ml). See also #3300.

 - It is inefficient as it computes the join of all types appearing in *any* joined environment prior to filtering out the types that are not needed, instead of first computing the types whose join will be needed.

 - It is sensitive to the names of local variables that only exist in some of the joined environments but not in the target environment.

 - It relies on a global binding time of variables across all joined environments and the target environment that does not exist, as figured in #3278. Subsequently, it can lose aliasing information, and breaks typing env invariants by recording the same variable as defined multiple times (with dubious semantics).

This patch implements a new join algorithm, based on a n-way join of types. The new algorithm is:

 - Faster, as it avoids quadratic complexity (outside of complex nesting of env extensions). Compared to the existing join algorithm (with advanced meet), on my machine, the new join algorithm is 30x faster on the pathological lambda_to_flambda_primitives.ml, taking only around 10% of the total compilation time and speeding up the compilation of the file by 3.5x. On camlinternalFormat.ml, the new join is about 2.5-3x faster, reducing the time spent in the join from 20% to less than 10% and speeding up the total compilation time by about 20%.

 - More efficient, as it only computes a join if it can possibly result in a more precise type, i.e. if the variable has been assigned a new type in all joined environments (otherwise the existing type in the target environment is already the most precise).

 - Independent of the names of local variables.

 - Only depends on a consistent binding time *order* of the shared variables (defined in both the target environment and all joined environments), which is respected. Since the result is independent of the binding times of local / existential variables, the typing env invariants are respected.

## Reviewing guide

The actual changes in this PR are in the `Meet_and_join_new` and `Join_env` modules, as well as `Join_levels` (which is now just a small wrapper over `Join_env`). The changes in both modules are mostly independent: changes in `Meet_and_join_new` make the join of types a n-way join, while the `Join_env` module implements the new join algorithm for environments, and in particular takes care of joining aliases. The main thing to know to relate these two modules is that the join of (canonical) alias types should go through `Join_env.n_way_join_simple`, which will return an appropriate alias (creating a local variable if required) in the target env.

(As an aside: I think the organization of the `Join_env` module is roughly right, but am not too satisfied with the *names* of things and happily take suggestions)

The rest of the changes are temporary, and mostly intended for further debugging of the PR. They should not need to be reviewed, and will be removed before merging the PR. In particular, the `Meet_new_and_join_old` and `Join_levels_old` modules are (almost) identical copies of the `Meet_and_join_new` and `Join_levels` modules from `main`. They are used in conjunction with the environment variable `FLAMBDA2_JOIN_ALGORITHM=checked` to compare the results of the new and old join algorithm. This comparison uses the auxiliary `Equal_types_for_debug` module to print out the result of joins that are different when computed using the new and old join algorithms. When visually inspecting such differences, I recommend also setting `FLAMBDA2_JOIN_DEBUG_IGNORE_NAMES=cse_param` -- the new join is better at preserving equations between `cse_param` variables and this creates noise.

From my own inspection, the main differences between the old and new join is that the new join is better at preserving aliases (so sometimes the new join will have an alias where the old join will have duplicated the type), and it is also better at preserving the `alloc_mode` of blocks (the old join considers the alloc mode of Bottom blocks, while the new join ignores them -- which I think is correct).

## Symbol projections

I am not sure I understand how symbol projections are supposed to be joined. The old join seems to preserve symbol projections as long as they are present in *one* of the joined envs (and the variable is still accessible in the target env), which might make sense from the point of view of Simplify but seems dubious from the point of view of the typing env. The new join preserve symbol projections if they are present in all of the joined envs (up to demotions) because that is what seems sensible for a join, but I am not sure if this is the intended semantics. I don't really understand how symbol projections are created/used; maybe @mshinwell can clarify.

## Basic meet

The PR is currently only compatible with the new (advanced) meet, which will always be used when joining environments. Before merging the PR, we should decide how we want to proceed w.r.t. the basic meet. If it goes away, there is nothing to do; if we still keep it for a bit, we can either keep the `Join_levels_old` module for the basic meet; force the join to go through the advanced meet (even if the basic meet is otherwise used); or adapt the n-way join code for the basic meet as a last resort.

### DIFF
--- a/middle_end/flambda2/types/meet_and_n_way_join.ml
+++ b/middle_end/flambda2/types/meet_and_n_way_join.ml
@@ -44,7 +44,7 @@ type 'a n_way_join_result = 'a Or_unknown.t * Join_env.t
 
 let map_join_result ~f (v, env) = Or_unknown.map ~f v, env
 
-let ( let>+ ) x f = map_join_result ~f x
+let ( let>>+ ) x f = map_join_result ~f x
 
 let add_equation (simple : Simple.t) ty_of_simple env ~meet_type :
     unit meet_result =
@@ -1698,7 +1698,7 @@ and n_way_join_expanded_head env kind (expandeds : ET.t Join_env.join_arg list)
                   | Value head -> head | _ -> assert false)
                 expandeds
             in
-            let>+ head =
+            let>>+ head =
               n_way_join_head_of_kind_value env ((id1, head1) :: heads)
             in
             ET.create_value head
@@ -1709,7 +1709,7 @@ and n_way_join_expanded_head env kind (expandeds : ET.t Join_env.join_arg list)
                   | Naked_immediate head -> head | _ -> assert false)
                 expandeds
             in
-            let>+ head =
+            let>>+ head =
               n_way_join_head_of_kind_naked_immediate env ((id1, head1) :: heads)
             in
             ET.create_naked_immediate head
@@ -1720,7 +1720,7 @@ and n_way_join_expanded_head env kind (expandeds : ET.t Join_env.join_arg list)
                   | Naked_float32 head -> head | _ -> assert false)
                 expandeds
             in
-            let>+ head =
+            let>>+ head =
               n_way_join_head_of_kind_naked_float32 env (head1, id1) heads
             in
             ET.create_naked_float32 head
@@ -1731,7 +1731,7 @@ and n_way_join_expanded_head env kind (expandeds : ET.t Join_env.join_arg list)
                   | Naked_float head -> head | _ -> assert false)
                 expandeds
             in
-            let>+ head =
+            let>>+ head =
               n_way_join_head_of_kind_naked_float env (head1, id1) heads
             in
             ET.create_naked_float head
@@ -1742,7 +1742,7 @@ and n_way_join_expanded_head env kind (expandeds : ET.t Join_env.join_arg list)
                   | Naked_int32 head -> head | _ -> assert false)
                 expandeds
             in
-            let>+ head =
+            let>>+ head =
               n_way_join_head_of_kind_naked_int32 env (head1, id1) heads
             in
             ET.create_naked_int32 head
@@ -1753,7 +1753,7 @@ and n_way_join_expanded_head env kind (expandeds : ET.t Join_env.join_arg list)
                   | Naked_int64 head -> head | _ -> assert false)
                 expandeds
             in
-            let>+ head =
+            let>>+ head =
               n_way_join_head_of_kind_naked_int64 env (head1, id1) heads
             in
             ET.create_naked_int64 head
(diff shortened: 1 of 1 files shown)


bclement-ocp:
(Looking at the failures — we should probably detect Bottom prior to "Cannot add alias between two consts: 1, 0")

bclement-ocp:
> (Looking at the failures — we should probably detect Bottom prior to "Cannot add alias between two consts: 1, 0")

Yup, this one seems ultimately unrelated — we have an invariant asserting that we do not add aliases between two constants that we check *before* the code that deals with aliases between constants. Will fix in a separate PR.

The other is a real bug in the n-way join — I was misled to the semantics of the "other" field for row-like by the code for the meet, and do not deal properly with the join of a tag that is known on one side and "other" in an other side. Fix incoming.

bclement-ocp:
I forgot to mention in the PR description that there is a subtle case with env extensions where we can lose equations.

Consider that we are performing the join between `x : (tagged_imm (= a))` and `x : (tagged_imm (= b))`. We will create a new existential variable `ab` with type the join of the types of `a` and `b` and set `x : (tagged_imm (= ab))`.

If we later have an env extension with `b : (= a)` on the right, we will include `ab : (= a)` in the joined extension. But if we process the env extension first, we will not have the equation `ab : (= a)` in the joined extension.

I don't think we are likely to hit this case frequently in practice, so it's probably fine, but ideally we'd want to process env extensions after the non-extension types to avoid this corner case.

flambda-backend/tests/backend/vectorizer/filter.sh
```diff
@@ -1,3 +1,3 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
lthls, line 1: You should make an independent PR for that

bclement-ocp:
Note: the new join implements a variation of @lthls's [suggestion](https://github.com/ocaml-flambda/flambda-backend/issues/472#issuecomment-1016253875) in #472 and fixes #472 even though it does not use the `join_depth` parameter from #474.

lthls (commented):
I've gone through `joined_env.ml`, which should be the most complex part of the PR. I'm posting my review notes on this file, and will move on to the rest of the PR.

bclement-ocp:
Rebased on top of #3689 and added proper `-flambda2-join-algorithm` flag.

This makes the review of `Meet_and_n_way_join` a bit more annoying, especially for `meet_disjunction` and `meet_row_like`, since it's no longer in the same file. Sorry :(

lthls (approved):

PR #3541, opened 2025-02-04
Portable continuations

dkalinichenko-js:
Support portable continuations in the effects API. Also, annotate every function there as portable. Add a test.

Unsafe: calling the continuation from two different domains at once may lead to data races due to shared mutable state. A subsequent PR will require the continuations to be `unique` to prevent that.


otherlibs/stdlib_alpha/effect.mli
```diff
@@ -156,6 +192,12 @@ module type S = sig
       is passed a [t Handler.t] so that it can perform operations from effect
       [t]. *)
 
+  val portable_fiber :
+    (t Handler.t @ local -> 'a -> 'b) @ portable
+    -> ('a, ('b, unit) Result.Portable.t, unit) Continuation.t @ portable
+  (** [portable_fiber f] works the same as [fiber], but takes a [portable]
+      closure and creates a [portable] continuation. *)
+
```
lpw25, line 200: Rather than have `foo`, `portable_foo`, `Portable.foo` and `Portable.portable_foo`, I think it would be better to follow the pattern you used within result: `foo`, `Portable.foo`, `Contended.foo` and `Portable.Contended.foo`. With `Portable` referring to portable continuations and `Contended` referring to functions that where the `Handler.t` can/has crossed a capsule boundary.

I think it would then be better to also have `Portable.Result.t` rather than `Result.Portable.t`.

otherlibs/stdlib_alpha/effect.ml
```diff
@@ -529,18 +533,16 @@ let fiber (type a b e)
...
 let fiber_with (type a b e es) (local_ l : es Handler.List.Length.t)
     (f : local_ (e * es) Handler.List.t -> a -> b) =
-  let module Effs = struct
+  let module H = Handler.Create(struct
       type nonrec e = e
       type nonrec es = es
-    end
-  in
-  let module H = Handler.Create(Effs) in
+    end) in
```
lpw25, line 545: This change seems superfluous?

otherlibs/stdlib_alpha/effect.ml
```diff
@@ -577,10 +579,45 @@ let run_with (type a e es) (local_ hs : es Handler.List.t)
...
+    -> (('e * 'es) Handler.List.t @ local -> 'a)
+    -> ('a, 'e, 'es) res
+end
+
+module type PortableAux = sig include Aux @@ portable end
+let magic_aux (aux : (module Aux)) : (module PortableAux) = Obj.magic aux
+
+(* CR dkalinichenko: [fiber], [fiber_wth], [run] and [run_with] are not
+   [portable] because they use first-class modules, even though it's safe
+   to declare them [portable] becuase the first-class module is created locally.
```
lpw25, line 603: I think they actually can't be portable due to the use of the `Handler.Create` functor, since functors cannot yet be portable. That functor can be replaced with a function that takes a first-class module and returns another first-class mode. Such a function can be portable, so I think that might be enough to remove this magic.

otherlibs/stdlib_alpha/effect.ml
```diff
@@ -896,18 +1295,105 @@ module Make_rec (Ops : Operations_rec)
...
+      (* CR dkalinichenko: here and below, we use [Obj.magic_uncontended]
+         to mark a [contended] handler safe to use with [portable]
+         operations. 
+         
+         We may be able to rewrite this to avoid magic. *)
+      perform (op, Obj.magic_uncontended h.h)
+  end
+
+  let portable_fiber (type a b) f =
+    let k : (a, b, t, unit) continuation = Obj.magic (fiber f) in
```
lpw25, line 1368: This line looks identical to 1373 but has an `Obj.magic`. Is it definitely needed?

otherlibs/stdlib_alpha/effect.ml
```diff
@@ -896,18 +1295,105 @@ module Make_rec (Ops : Operations_rec)
...
+  (* Here and below, [portable] versions of functions are defined before
+     non-[portable] ones to avoid shadowing the original functions. *)
+
+  module Portable = struct
+    let portable_fiber (type a b) f =
+    (* CR dkalinichenko: here and below, we use [magic_portable]
+       to indicate that the handler is [portable] when it receives
+       [contended] values.
+      
+       We may be able to rewrite this to avoid magic. *)
```
lpw25, line 1307: Rather than have lots of repeated `Obj.magic_portable`s, could you instead use just one to create a toplevel `fiber_portable` function that is equal to the toplevel `fiber` function.

testsuite/tests/effects-api/cross_capsule.ml
```diff
@@ -0,0 +1,47 @@
...
+  type 'a t = 'a op
+end)
+
+let store = Atomic.Safe.make 0
+
+let rec handle = function
+  | Eff.Result.Contended.Value v -> v
+  | Exception e -> raise e
+  | Operation (Set n, k) ->
+    Atomic.Safe.set store n;
```
lpw25, line 26: Why is this using `Atomic.t` rather than just `ref`? I don't think you are relying on `store` itself being `contended` anywhere.

lpw25 (commented):
Looks good. Various suggestions

otherlibs/stdlib_alpha/effect.mli
```diff
@@ -56,9 +58,12 @@ end
...
 module Continuation : sig
 
-  type (-'a, +'b, 'es) t
+  type (-'a, +'b, 'es) t : value mod contended
   (** [('a, 'b, 'es) continuation] is a delimited continuation that expects an ['a]
-      value and returns a ['b] value. It requires handlers for the effects ['es]. *)
+      value and returns a ['b] value. It requires handlers for the effects ['es].
+
+      Like closures, continuations cross contention and are [portable] if the
+      variables they capture are [portable]. *)
```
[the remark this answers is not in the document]
dkalinichenko-js: Done.

otherlibs/stdlib_alpha/effect.ml
dkalinichenko-js: Done. Unfortunately, creating the first-class module directly in the function leads to inefficient compilation, making one of the tests fail running out of memory. But with the functor private to the implementation, we can limit the magic to one function.

otherlibs/stdlib_alpha/effect.ml
dkalinichenko-js: Removed redundant `Obj.magic` calls.

otherlibs/stdlib_alpha/effect.ml
dkalinichenko-js: Done, now everything lives in the `DRF` module.

testsuite/tests/effects-api/cross_capsule.ml
dkalinichenko-js: Fixed.

dkalinichenko-js:
Updated the interface and streamlined the implementation. In particular:
1. Interfaces now follow the `S`, `S.Contended`, `S.Portable` and `S.Portable.Contended` format.
2. `portable` or `contended` upgraded to `portended` everywhere it's not superfluous.
3. `Portable.run` functions no longer return `Result.t @ portable`, that was unsound.

otherlibs/stdlib_alpha/effect.ml
lpw25: The allocation semantics of the functor and the function working on first-class modules should actually be equivalent, so I don't think that should actually happen. Functors do get more aggressive inlining, maybe annotating the function `[@inline]` would fix it. Either way we shouldn't be able to cause anything to run out of memory -- any such allocation would be extremely short-lived. I think it would be worth working out what was going on there.

dkalinichenko-js: The exact error is `Fatal error: allocation failure during minor GC`, and it happens only in bytecode in the `evenodd.ml` test, which creates 100000 handlers in succession.

dkalinichenko-js: Seems like it doesn't happen in the test runner, but I can consistently reproduce it on my machine.

dkalinichenko-js: Looks like the value handler pointer in the top ~25000 handlers is for some reason not updated after a major GC in bytecode, which leads to a crash when trying to interpret them as pointers to live blocks. I'll try to reproduce this using upstream effects.

dkalinichenko-js: Max pointed me to https://github.com/ocaml-flambda/flambda-backend/pull/3231, which allowed me to resolve the issue. I'll create a separate PR for it.

lpw25 (commented):
Looks good. I think there are a couple changes that we should make across the interface. I left comments showing examples of them both.

otherlibs/stdlib_alpha/effect.mli
```diff
@@ -182,6 +188,136 @@ module type S = sig
...
+
+    val run_with :
+      'es Handler.List.t @ local portable
+      -> ((t * 'es) Handler.List.t @ local portable -> 'a)
+      -> ('a, 'es) Result.t
+    (** [run_with hs f] works as the non-contended version, but provides
+        [portable] handlers and returns a [Contended.Result.t]. *)
+
+    val perform :
+      t Handler.t @ local portable contended ->
```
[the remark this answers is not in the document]
dkalinichenko-js: That would mean the handler will receive the `('a, t) ops @ portable contended` at the legacy modality. This seems unsound to allow for all handlers?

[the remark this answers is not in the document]
lpw25: I think that is already unsound. You can call both `perform` and `Contended.perform` on the `Handler.t` that comes from a `Contended.fiber`, so you already don't know that the `ops` are portable.

The portable should be removed from the `ops` part of the `Contended.Result.t` type, and the `contended` should be removed from the `ops` parameter here.

This is all related to my point 2 in the other comment.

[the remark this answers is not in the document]
dkalinichenko-js: Done.

otherlibs/stdlib_alpha/effect.mli
```diff
@@ -182,6 +188,136 @@ module type S = sig
...
+        | Exception : exn @@ global -> ('a, 'es) t
+        | Operation :
+            ('o, eff) ops @@ global portable contended
+            * ('o Modes.Portended.t, ('a, 'es) t, 'es) Continuation.t
+            -> ('a, 'es) t
+    end
+
+    val fiber :
+      (t Handler.t @ local portable -> 'a @ portable contended -> 'b)
+      -> ('a Modes.Portended.t, ('b, unit) Result.t, unit) Continuation.t
```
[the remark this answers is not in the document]
dkalinichenko-js: Done.

otherlibs/stdlib_alpha/effect.mli
```diff
@@ -182,6 +188,140 @@ module type S = sig
...
+    module Result : sig
+      type eff := t
+
+      type ('a, 'es) t =
+        | Value : 'a @@ global -> ('a, 'es) t
+        | Exception : exn @@ global -> ('a, 'es) t
+        | Operation :
+            ('o, eff) ops @@ global
+            * ('o, ('a, 'es) t, 'es) Continuation.t @@ portable
+            -> ('a, 'es) t
```
lpw25, line 251: Suddenly, I'm worried about the soundness of this version of things. A portable closure is clearly necessary for a portable continuation, but I'm worried it is not sufficient. The values passing in and out of the closure probably need to be considered to be crossing a capsule boundary too. For instance, I think we can currently write this:
```ocaml
module Op = struct
  type 'a t =
    | Fetch : int ref t
    | Yield : unit t
end
module Eff = Effect.Make(Op)

let comp h =
  let r = Eff.Portable.perform h Fetch in
  Eff.Portable.perform h Yield;
  while true do
    incr r
  don

let foo () =
  let r = ref 0 in
  match Eff.Portable.run comp with
  | Value _ | Exception _ | Operation(Yield, _) -> assert false
  | Operation(Fetch, k) ->
      match Effect.continue k r with
      | Value _ | Exception _ | Operation(Field, _) -> assert false
      | Operation(Yield, k) ->
          Domain.spawn (fun () -> ignore (Effect.continue k ()));
          !r
```
which has a data race.

I think that we should make the `Portable.Contended` version be the `Portable` version, and I think that version should also be adjusted slightly so that the `Handler.t` starts out `contended`.

lpw25 (commented):
I was all set to approve, but then I realized that there was a soundness issue.

otherlibs/stdlib_alpha/effect.mli
lpw25: Actually, I'm not even sure we need to have the `Portable` version. Can't we just use a capsule paired with a key to make a portable continuation? Now that we require uniqueness anyway it isn't any harder to use.

dkalinichenko-js: Yeah, you're right, the above example works, even with `Domain.Safe.spawn`.

How would a capsule with a key help? The continuation is given by pattern-matching inside the handler, and, if it's not portable, we can't move it into another capsule, even a new one, in the first place.

lpw25: Here is an example of a handler to create portable int generators:

```ocaml
module Ops = struct
  type 'a t = Yield : int -> unit t
end
module Eff = Effect.Make(Ops)

type state =
  | Used
  | Cont : { key : 'k Capsule.Key.t;
                 cont : ((unit, (unit, unit) Eff.Result.t) Effect.Continuation.t, 'k) Capsule.Data.t  } -> state

type t = state Unique_atomic.t

let create ~f =
  let Packed key = Capsule.Key.create () in
  let cont = Capsule.Data.create (Eff.fiber (fun h () -> f h)) in
  Unique_atomic.create (Cont { key; cont })

exception Already_used

let next t =
  match Unique_atomic.swap t Used with
  | Used -> raise Already_used
  | Cont { key; cont } ->
      let res, key = 
        Capsule.Key.with_access key (fun ax ->
          let cont = Capsule.Data.unwrap ax cont in
          match Effect.continue cont () with
          | Eff.Value () -> None
          | Eff.Exception e -> raise e
          | Eff.Operation(Yield i, cont) ->
              Some (i, Capsule.Data.wrap ax cont)
      in
      match res with
      | None -> None
      | Some (i, cont) -> Some (i, Cont { key; cont })
```
Note that `create` and `next` both produce `portable` `t`s that you can pass between domains.

dkalinichenko-js: This works. Added a test, removed `Portable` modules from the interface.

lpw25 (approved):
LGTM

PR #3543, opened 2025-02-05
Reduce default space overhead from 160 to 80

NickBarnes:
160 is definitely too large a value for this parameter now. 100 is the default value for runtime4 (whereas upstream ocaml/ocaml has 120). Also runtime4 undershoots this parameter by quite a lot in the presence of big arrays. With our new pacing policy, runtime 5 does not. Some basic eyeballing of `memtrace-viewer` charts suggests that the runtime5 behaviour at 80 is pretty similar to runtime4 at 100.


stedolan:
This change looks good to me but I'm going to hold off on merging until after we've merged the upcoming work on GC pacing, which rewrites the calculations that depend on this value.

mshinwell (approved):

PR #3547, opened 2025-02-05
Add a flag to control inference of with-bounds on type decls

glittershark:
Add a new flag, -infer-with-bounds, which enables inference of
with-bounds (called "baggage" as of this commit, but to be renamed to
with-bounds later) on type declarations. This defaults to false, and has to be
turned on per-compilation-unit, because we are concerned about the performance
impact of rolling out with-kind inference and want to do so gradually.

If the flag is off, we restore the original inferred kinds prior to the
introduction of with-bounds.

### DIFF
--- a/testsuite/tests/typing-jkind-bounds/no-infer-across-modules/define_with_kinds.ml
+++ b/testsuite/tests/typing-jkind-bounds/no-infer-across-modules/define_with_kinds.ml
@@ -12,7 +12,7 @@ and 'a rec2 =
   | Foo2 of 'a rec1 * int
   | Bar2 of 'a
 
-type ('a : value & value) unboxed_rec = #{ x : 'a ; y : string }
+type ('a : value & value) unboxed_record = #{ x : 'a ; y : string }
 
 module type Optionish = sig
   type 'a my_option =
--- a/testsuite/tests/typing-jkind-bounds/no-infer-across-modules/define_with_kinds.ml
+++ b/testsuite/tests/typing-jkind-bounds/no-infer-across-modules/define_with_kinds.ml
@@ -1,0 +1,27 @@
+type 'a my_list =
+  | Nil
+  | Cons of 'a * 'a my_list
+
+type 'a my_ref = { mutable contents : 'a }
+
+type 'a rec1 =
+  | Foo1 of 'a rec2 * int
+  | Bar1 of 'a
+
+and 'a rec2 =
+  | Foo2 of 'a rec1 * int
+  | Bar2 of 'a
+
+type ('a : value & value) unboxed_rec = #{ x : 'a ; y : string }
+
+module type Optionish = sig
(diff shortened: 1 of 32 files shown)


testsuite/tests/typing-jkind-bounds/no-infer-across-modules/define_with_kinds.ml
```diff
@@ -0,0 +1,27 @@
...
+
+type 'a rec1 =
+  | Foo1 of 'a rec2 * int
+  | Bar1 of 'a
+
+and 'a rec2 =
+  | Foo2 of 'a rec1 * int
+  | Bar2 of 'a
+
+type ('a : value & value) unboxed_rec = #{ x : 'a ; y : string }
```
liam923, line 15: I object to this name since the previous type uses `rec` to mean "recursive" and here it means "record"

testsuite/tests/typing-jkind-bounds/no-infer-across-modules/test.ml
```diff
@@ -0,0 +1,275 @@
...
+       The kind of #((int -> int) * int) Define_with_kinds.unboxed_rec is
+         (immediate & immediate) & immediate.
+       But the kind of #((int -> int) * int) Define_with_kinds.unboxed_rec must be a subkind of
+         (value & value) & value
+         because of the definition of use_portable_three_values at line 2, characters 84-95.
+|}]
+
+(******)
+
+module Option : Optionish = struct
```
liam923, line 259: I'd also like to see:
```
module rec Foo : Optionish = Foo
```
```
type 'a my_option = 'a Optionish2.my_option = ...
```
Maybe there's some others along this line I'm not thinking of?

testsuite/tests/typing-jkind-bounds/no-infer-across-modules/define_with_kinds.ml
```diff
@@ -0,0 +1,27 @@
...
+
+module type Optionish = sig
+  type 'a my_option =
+    | Nothing
+    | Just of 'a
+end
+
+module Optionish2 = struct
```
liam923, line 23: Fun fact: these can both be `Optionish`. Module types are a separate namespace from modules (they can never appear in the same syntactic location)

liam923 (commented):
Besides those couple test cases I requested, lgtm. As I mentioned offline, I'm pessimistic about this being a silver bullet than will fix our performance woes. But either way, I think this is still a good change.

testsuite/tests/typing-jkind-bounds/no-infer-across-modules/define_with_kinds.ml
glittershark: yeah, but that's just unnecessarily confusing :p

liam923 (approved):

PR #3549, opened 2025-02-06
Update documentation for `or_null`

dkalinichenko-js:
Title.

### DIFF
--- a/jane/doc/proposals/unboxed-types/inference.md
+++ b/jane/doc/proposals/unboxed-types/inference.md
@@ -67,53 +67,27 @@ modules or doing arithmetic on IDs.
 ## Defaults
 
 When inferring the kind of a type, we must consider what we use as the
-default layout. In general, we want make a choice low in the lattice for
-covariant positions and high in the lattice for contravariant positions, as
-these defaults will allow for more expressiveness. That is, if we say
-`type 'a t` (with no `=`), a low-in-the-lattice layout for `t` means that `t`
-can be used in more contexts, while a high-in-the-lattice layout for `'a` means
-that `t` can be applied to more types. With this in mind, we will use the
-following defaults:
-
-```ocaml
-type t : <<here>>                 (* default: non_null_value *)
-type ('a : <<here>>) t            (* default: non_null_value *)
-fun (type (a : <<here>>)) -> ...  (* default: value *)
-let f : ('a : <<here>>). ...      (* default: value *)
-```
-
-Abstract types `t` should have a default layout of `non_null_value`, as this is
-both low in the lattice and also the layout of any type declared with record,
-variant, extensible variant, or object type definition. It thus seems likely
-that the default will work with both usages of `t` and the definition of `t`.
-
-A parameter to an abstract type `'a` in `type 'a t` (ditto the type variables in
-`type ('a, 'b) t`) also defaults to have layout `non_null_value` -- even though
-this choice is low in our lattice. The reason we must do this is to maintain
-backwards compatibility. Consider the following module:
+default layout. Backwards compatibility forces us to use the same defaults
+for abstract types and type parameters:
 
 ```ocaml
 module M : sig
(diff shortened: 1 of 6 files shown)


jane/doc/proposals/unboxed-types/inference.md
goldfirere, line 1: I don't understand the changes here. Well, I understand that we're using one default -- `value` -- everywhere. But the new "assumed" comments don't hit home for me. Maybe we just state that we have to use `value` everywhere? The `module M : sig type 'a t end = struct type 'a t = 'a end` example is good to illustrate why `'a`'s default needs to match `t`'s.

jane/doc/proposals/unboxed-types/null.md
```diff
@@ -25,120 +25,110 @@ We thus want `t or_null` to be just like `t`, except now with a meaning
...
-In addition, we will allow `'a or_null` to mode-cross the same way that
-`'a` does -- effectively allowing `or_null` to work nicely with `immediate`s.
-This can be implemented by "looking through" the
-`or_null` type constructor during mode-cross-checking. (The alternative would be to
-have some degree of kind polymorphism, where we have
-`type ('a : non_null_value & 'modes) or_null : value & 'modes`
-or similar, but looking through
-`or_null` is easy enough so that we can avoid this extra complexity.)
-This permits, e.g. `int or_null` to still be subject to
+With-kinds will allow `'a or_null` to mode-cross the same way that
```
goldfirere, line 59:
```suggestion
The with-kinds feature allows `'a or_null` to mode-cross the same way that
```

jane/doc/proposals/unboxed-types/null.md
```diff
@@ -25,120 +25,110 @@ We thus want `t or_null` to be just like `t`, except now with a meaning
...
 ```ocaml
-type ('a : value) pair_with_default = { contents : 'a or_null; default : 'a }
+module M : sig
+  type 'a t (* Assumed non-null. *)
+end = struct
+  type 'a t = 'a (* [’a] must be non-null. *)
+end
+
+module F (X : sig
+  type t (* Assumed maybe-null. *)
```
goldfirere, line 85: Is this really true? I didn't know we had a different default in a functor sig. That seems hard.

jane/doc/proposals/unboxed-types/null.md
```diff
@@ -25,120 +25,110 @@ We thus want `t or_null` to be just like `t`, except now with a meaning
...
 
-## Sub-typing
+## Arrays
 
-Because any valid `t` is also a valid `t or_null`, we have a natural sub-typing
-relationship: `t` is a sub-type of `t or_null` for all `t` (such that `t
-or_null` is well formed). Note that this relationship is entirely notionally
-separate from the sub-layout relationships described above. Here, we're talking
-about term-level conversions such as in
+Flat-float array optimization demands that all types in OCaml are *separable*:
```
goldfirere, line 101:
```suggestion
The flat-float-array optimization demands that all types in OCaml are *separable*:
```

jane/doc/proposals/unboxed-types/null.md
goldfirere, line 1: `non_null` is also allowed after `mod`, correct? Or maybe that's only with the with-kinds patch, which this predates? And it probably belongs with the with-kinds documentation. Which doesn't exist. Ok, so maybe we don't include this detail here.

goldfirere, line 1: This is now implemented; I think it should be in the `extensions` folder.

goldfirere (approved):

jane/doc/proposals/unboxed-types/null.md
dkalinichenko-js: No, we don't. I was illustrating the reasoning why the defaults must agree. I will rewrite this to be clearer.

PR #3550, opened 2025-02-06
Ensure that field_set always returns unit

jvanburen:
Fixes a regression from #3415 where field stores without `C.return_unit` can cause the register allocator to fail.


xclerc (approved):

Gbury:
How come this was not caught by the CI tests ?

xclerc:
There does not seem to be a test mutating
a naked float in mixed block. Assuming that's
true, it should be added, but possibly in a
follow-up pull request. Indeed,  the bug is
breaking builds, so it would be nice to merge
this pull request soon.

@Gbury would you have the time to review?

gretay-js (approved):
This seems like a straightforward fix. We can merge it to fix some builds that are failing, and add a test as a follow-up PR.

Gbury (approved):

PR #3553, opened 2025-02-06
Only run `memset` when stack checks are enabled

dkalinichenko-js:
When stack checks are disabled, `memset` tries to access the guarded page of the stack, causing a segfault.

This issue was introduced in #3384 and only shows ups when running effects with debug runtime.

### DIFF
--- a/runtime/fiber.c
+++ b/runtime/fiber.c
@@ -939,7 +939,7 @@ void caml_free_stack (struct stack_info* stack)
     stack->exception_ptr =
       (void*)(cache[stack->cache_bucket]);
     cache[stack->cache_bucket] = stack;
-#ifdef DEBUG
+#if defined(DEBUG) && defined(STACK_CHECKS_ENABLED)
     memset(Stack_base(stack), 0x42,
            (Stack_high(stack)-Stack_base(stack))*sizeof(value));
 #endif
--- a/testsuite/tests/effects/test1.ml
+++ b/testsuite/tests/effects/test1.ml
@@ -1,9 +1,11 @@
 (* TEST
- skip;
+  runtime5;
+  { bytecode; }
+  { native; }
 *)
 
-open Effect
-open Effect.Deep
+open Stdlib__Effect
+open Stdlib__Effect.Deep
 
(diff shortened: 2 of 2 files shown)


runtime/fiber.c
```diff
@@ -939,7 +939,7 @@ void caml_free_stack (struct stack_info* stack)
     stack->exception_ptr =
       (void*)(cache[stack->cache_bucket]);
     cache[stack->cache_bucket] = stack;
-#ifdef DEBUG
+#if defined(DEBUG) && defined(STACK_CHECKS_ENABLED)
```
stedolan, line 942: Instead of entirely disabling this debug feature, I'd prefer to make it work properly in this case, with something like the following (untested):
```c
char* base;
#if defined(NATIVE_CODE) && !defined(STACK_CHECKS_ENABLED)
  base = Protected_stack_page(stack, caml_plat_pagesize) + caml_plat_pagesize;
#else
  base = (char*)Stack_base(stack);
#endif
memset(base, 0x42, (char*)Stack_high(stack) - base); 
```

stedolan (changes requested):

runtime/fiber.c
dkalinichenko-js: Got it. I'm not sure why this is enabled even without stack checks while a similar check just a few lines below requires `STACK_CHECKS_ENABLED`, but will do.

stedolan: On second thoughts, let's go with your original implementation instead of my suggestion.

Stacks allocated when stack checks are disabled are huge, and we hope not to actually fault all that memory in but instead rely on lazy OS allocation. It would be a shame to fault it all in just to free it! On programs with lots of effects / systhreads, this could plausibly cause out-of-memory errors in debug builds.

(This means we'll lose the debug check, but oh well)

dkalinichenko-js: I see, should we merge or there's something else to do?

stedolan: Let's merge!

stedolan (approved):

PR #3554, opened 2025-02-06
Finer typing for closing over module coercion

riaqn:
This PR improves the type checking of a `Pmod_ident` that's immediately coerced. Specifically:
- When it's the argument to a functor application.
- When it's being packed into a first class module
- When it's in `Pmod_constraint`.

`Pmod_unpack` is not dealt with because it's doesn't coerce module. See comments in test for details.

# Review
- Please review by commits.
- Some of the changes interact with #2431 #3398 , care is taken to minimize the diff with upstream. The reviewer is welcome to spot further clean-ups possible.

### DIFF
--- a/testsuite/tests/typing-modes/module.ml
+++ b/testsuite/tests/typing-modes/module.ml
@@ -229,5 +229,5 @@ let (bar @ portable) () =
 Line 3, characters 19-20:
 3 |         module L = M
                        ^
-Error: Modules are nonportable, so cannot be used inside a function that is portable.
+Error: "M" is a module, and modules are always nonportable, so cannot be used inside a function that is portable.
 |}]
--- a/testsuite/tests/typing-modes/module.ml
+++ b/testsuite/tests/typing-modes/module.ml
@@ -29,20 +29,6 @@ module M : sig type 'a t = int val x : 'a -> unit end
 module F : functor (X : S) -> sig type t = int val x : 'a -> unit end
 |}]
 
-(* Closing over modules affects closure's modes *)
-let u =
-    let foo () =
-        let _ = (module M : S) in
-        ()
-    in
-    portable_use foo
-[%%expect{|
-Line 6, characters 17-20:
-6 |     portable_use foo
-                     ^^^
-Error: This value is "nonportable" but expected to be "portable".
-|}]
-
 let u =
     let foo () =
         let module X = struct
@@ -57,21 +43,6 @@ let u =
 Line 10, characters 17-20:
 10 |     portable_use foo
                       ^^^
-Error: This value is "nonportable" but expected to be "portable".
-|}]
-
-(* File-level modules are looked up differently and need to be tested
-separately. *)
-let u =
-    let foo () =
-        let _ = (module List : SL) in
-        ()
-    in
-    portable_use foo
-[%%expect{|
-Line 6, characters 17-20:
-6 |     portable_use foo
-                     ^^^
 Error: This value is "nonportable" but expected to be "portable".
 |}]
 
--- a/testsuite/tests/typing-modes/val_modalities.ml
+++ b/testsuite/tests/typing-modes/val_modalities.ml
@@ -926,7 +926,7 @@ module type S = sig
 end
 
 module type Module = sig
-  module M : sig include S end (* to prevent shallow_equal *)
+  module M : S
 end
 
 module type S' = sig
@@ -952,29 +952,14 @@ module type S =
     val baz : 'a -> 'a
     class cla : object  end
   end
-module type Module =
-  sig
-    module M :
-      sig
-        val x : int
-        val foo : 'a -> 'a @@ portable
-        val baz : 'a -> 'a
-        class cla : object  end
-      end
-  end
+module type Module = sig module M : S end
 module type S' =
   sig
     val x : int
     val foo : 'a -> 'a @@ portable
     val baz : 'a -> 'a
     class cla : object  end
-    module M :
-      sig
-        val x : int
-        val foo : 'a -> 'a @@ portable
-        val baz : 'a -> 'a
-        class cla : object  end
-      end
+    module M : S
   end
 module M : S
 |}]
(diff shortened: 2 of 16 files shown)


testsuite/tests/typing-modes/val_modalities.ml
```diff
@@ -901,59 +901,242 @@ let () =
...
+let (bar @ portable) () =
+  let k = (module M_Func_portable : Func_portable) in
+  k
+[%%expect{|
+module M_Func_portable : Func_portable
+Line 4, characters 18-33:
+4 |   let k = (module M_Func_portable : Func_portable) in
+                      ^^^^^^^^^^^^^^^
+Error: "M_Func_portable" is a module, and modules are always nonportable, so cannot be used inside a function that is portable.
+|}]
```
riaqn, line 1051: This might be less satisfactory. I would like @tdelvecchio-jsc 's opinion on this.

tdelvecchio-jsc: Does this get fixed if we change it to `module M_Func_portable : sig include Func_portable end = M`?

riaqn: Yes then the modules types are no longer shallow_equal. If you think this workaround is not too bad coding style, then we can go with it. We can fix it in the future if we still find it counter-intuitive.

typing/includecore.ml
```diff
@@ -127,17 +137,27 @@ let value_descriptions ~loc env name
...
       in
       let mode1 = Mode.Modality.Value.apply vd1.val_modalities mmode1 in
-      let mode2 =
-        Mode.Modality.Value.(Const.apply (to_const_exn vd2.val_modalities) mmode2)
+      let mode2 = Mode.Modality.Value.(Const.apply val2_modalities mmode2) in
+      let mode1 =
+        match close_over_coercion with
+        | Some held_locks ->
+          (* Similiar to [Typecore.type_ident] *)
+          let mode1 = left_mode_cross env vd2.val_type mode1 in
```
lukemaurer, line 155: Can you give an intuition for why `vd2.val_type` is used here rather than `vd1.val_type`?

testsuite/tests/typing-modes/val_modalities.ml
riaqn: Discussed with @lukemaurer and we decide to fix it in the last commit.

typing/typemod.ml
```diff
@@ -3500,9 +3487,20 @@ let type_toplevel_phrase env sig_acc s =
...
-  type_module_maybe_hold_locks ~alias:Yes_hold_locks true false None
+let type_module_alias env smod =
+  let md, shape, held_locks =
+    type_module_maybe_hold_locks ~alias:true ~hold_locks:true true false
+      None env smod
+  in
+  let locks =
+    match held_locks with
+    | None -> Env.locks_empty
+    | Some (locks, _, _) -> locks
```
lukemaurer, line 3498: What are we potentially throwing away here? Not sure what the `Longident.t` represents here in particular

testsuite/tests/typing-modes/module.ml
```diff
@@ -29,20 +29,6 @@ module M : sig type 'a t = int val x : 'a -> unit end
...
-        ()
-    in
-    portable_use foo
-[%%expect{|
-Line 6, characters 17-20:
-6 |     portable_use foo
-                     ^^^
-Error: This value is "nonportable" but expected to be "portable".
-|}]
-
```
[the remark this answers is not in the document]
riaqn: Yes this file is about legacy behaviors of module.

testsuite/tests/typing-modes/val_modalities.ml
```diff
@@ -1098,7 +1082,7 @@ module F : functor (X : sig end) -> sig end
 Line 4, characters 18-19:
 4 |   let k = (module F : F) in
                       ^
-Error: Modules are nonportable, so cannot be used inside a function that is portable.
+Error: "F" is a module, and modules are always nonportable, so cannot be used inside a function that is portable.
```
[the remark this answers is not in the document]
riaqn: All modules are always nonportable, not just functors.

typing/typemod.ml
riaqn: We are throwing away `lid` and `loc`, both of which are passed to `env.walk_locks` only used for error printing. This function is only used by `typecore` which doesn't need those info.

typing/env.ml
```diff
@@ -3307,29 +3304,20 @@ and lookup_apply ~errors ~use ~loc lid0 env =
...
+      let md = Subst.Lazy.force_module_decl data.mda_declaration in
+      path, md, locks
+  | Ldot(l, s) ->
+      let path, locks, data = lookup_dot_module ~errors ~use ~loc l s env in
+      let md = Subst.Lazy.force_module_decl data.mda_declaration in
+      path, md, locks
+  | Lapply _ as lid ->
+      let path_f, comp_f, path_arg = lookup_apply ~errors ~use ~loc lid env in
+      let md = md (modtype_of_functor_appl comp_f path_f path_arg) in
+      Papply(path_f, path_arg), md, locks_empty
```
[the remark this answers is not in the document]
riaqn: There are three such occurrences - do we put a comment for each of them?

testsuite/tests/typing-modes/val_modalities.ml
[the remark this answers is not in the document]
lukemaurer: Okay, right, but lots of modules will be usable in a `nonportable` function anyway. So “`F` is a module and is therefore nonportable” is now an insufficient explanation: it's being a functor that makes the reference forbidden

typing/env.ml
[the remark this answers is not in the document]
lukemaurer: This struck me as the one where it's particularly alarming that we're just returning `locks_empty`, and if you don't know that `Lapply` can never have runtime behavior then you'll think it's wrong. If the other two have the same kind of weirdness then sure

lukemaurer (approved):

PR #3558, opened 2025-02-07
Allow non-values in `[@@unboxed]` types

rtjoa:
Allow non-values in `[@@unboxed]` types. This was made straightforward by prior work by @goldfirere, so most of this PR is filling gaps in the existing testsuite (mainly non-value GADTs).

Review: @ccasin


testsuite/tests/typing-layouts-products/separability.ml
```diff
@@ -81,11 +81,17 @@ and bad = F : 'a r -> bad [@@unboxed]
...
+       This is most often caused by the fact that type inference is not
+       clever enough to propagate layouts through variables in different
+       declarations. It is also not clever enough to produce a good error
+       message, so we'll say this instead:
+         The layout of 'a r is any & any
+           because it is an unboxed record.
+         But the layout of 'a r must be representable
+           because it's the type of a constructor field.
+       A good next step is to add a layout annotation on a parameter to
+       the declaration where this error is reported.
```
ccasin, line 94: What happens here if we add a layout annotation on `'a`?  I guess `value` is the scary case where we want this to be rejected.

(But `void` is imaginary so even if this is allowed let's ship this, after writing some CRs/tickets)

testsuite/tests/typing-layouts-products/basics.ml
```diff
@@ -1412,78 +1412,48 @@ Error: This expression has type "letrec_simple"
          because it's the type of the recursive variable _x.
 |}]
 
-(**********************************************************)
-(* Test 11: not allowed in [@@unboxed] declarations (yet) *)
+(************************************************)
+(* Test 11: allowed in [@@unboxed] declarations *)
```
ccasin, line 1416: 🎉

ccasin (approved):
Looks great.  Let's make sure this gets into the next release.

testsuite/tests/typing-unboxed-types/unboxed_gadts.ml
```diff
@@ -0,0 +1,278 @@
+(* TEST
```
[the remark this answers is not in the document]
rtjoa: Renamed!

testsuite/tests/typing-layouts-products/separability.ml
rtjoa: I got this to typecheck by annotating `r` as `value & any` and leaving `'a` as a `value`! I think this still safe, but I left a CR and will make sure this is tracked internally.

testsuite/tests/typing-unboxed-types/test.ml
```diff
@@ -561,3 +561,124 @@ type ('a, 'kind) tree =
...
+       declarations. It is also not clever enough to produce a good error
+       message, so we'll say this instead:
+         The layout of 'a is value
+           because it instantiates an unannotated type parameter of tag,
+           chosen to have layout value.
+         But the layout of 'a must overlap with value & value
+           because of the annotation on 'a in the declaration of the type tag.
+       A good next step is to add a layout annotation on a parameter to
+       the declaration where this error is reported.
+|}]
```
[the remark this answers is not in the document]
rtjoa: Annotating `'c` as `value & value` does eliminate the error; added this test case and will track this internally.

testsuite/tests/typing-layouts-products/separability.ml
ccasin: The reason this is unsafe is because an array of `bad`s can have some elements that are `float`s and some elements that are not.  If you can use the [standard array creation function](https://github.com/ocaml-flambda/flambda-backend/blob/93e294b2972997eca06e6ea04223e065864c3e65/runtime4/array.c#L430) to make one, it might make it a float array (if the first element happens to be a float) or not - easily leading to segfaults.

We can probably handle this by teaching the compiler to treat `value & void` differently than `value`, and in particular use an array creation operation that never creates flat float arrays for this layout (which is is a little sad, but you could have used `float64 & void` if that's what you wanted).

But, in any case, there's no way to make an array for this type at all right now, since `void` is not supported past lambda.  So this is fine to merge.

rtjoa: Makes sense, thanks! Assumed we'd treat `value & void` differently by default so this is good to watch out for.

PR #3559, opened 2025-02-07
Initial stab at documenting the kind system

ccasin:
As title.

This is a draft PR because at a minimum there are some bits left to be filled in. But, also, I think we need some discussion about how various bits of the system should be presented. Here, I've tried to present things uniformly as they really are (or will be in a few weeks), so that a whole bunch of special cases aren't needed. But in practice our system obscures this reality a bit (for example, around nullability), and I'm not sure how best to handle that.

I've tried to write this more as useful documentation than as a complete reference.


jane/doc/extensions/kinds.md
```diff
@@ -0,0 +1,335 @@
...
+```
+And on locally abstract types:
+```ocaml
+let foo (type (a : value mod portable) (b : bits32 & float64)) (x : a) (y : b)
+  = ...
+```
+
+It's legal to omit the `mod ...` portion of the kind when you only want to
+specify the layout (as we did in a couple of these examples). The bounds of
+unspecified axes are the maximum on those axes.
```
goldfirere, line 124: This is subtle. When considered as a mode, `aliased` is the maximum on the uniqueness axis. But when considered as a modal bound, `unique` is the maximum: it mode-crosses the least. This means that `value mod aliased <= value mod unique` (because the latter is identically `value`), even though `unique <= aliased`.

This is sad and confusing, yet true of all past axes.

jane/doc/extensions/kinds.md
```diff
@@ -0,0 +1,335 @@
...
+
+## Kind abbreviations
+
+Kinds can get long, so we have built-in abbreviations for several common kinds,
+and there is a mechanism to define aliases of your own.  The builtin
+abbreviations include kinds like "immediate", which is secretly shorthand for
+the kind we showed for `int` above.  Here is a list of the built-in abbreviations:
+
+| Alias            | Meaning                                                                |
+|------------------|------------------------------------------------------------------------|
```
goldfirere, line 155: Add `unyielding` to all of these.

jane/doc/extensions/kinds.md
```diff
@@ -0,0 +1,335 @@
...
+  (* ... *)
+end
+```
+
+and still pass `Array` to functors expecting a `type 'a t`, which assumes `('a :
+value)`.
+
+CR ccasinghino: I just stole the above from the layouts docs, verbatim, except
+changing the word layout to kind.  Probably we should delete it from there and
+point here?
```
goldfirere, line 218: Yup. I think we can delete it there.

jane/doc/extensions/kinds.md
```diff
@@ -0,0 +1,335 @@
...
+
+The solution to this problem is "with kinds": kinds that record dependencies on
+types.  The actual kind of list in our.
+```ocaml
+type 'a list : value mod non_null (contended portable many with 'a)
+```
+
+CR ccasinghino: I'm giving up on explaining with-kinds here for now. I think
+there is no way to write the kind of list in the current system? (The above is
+not legal syntax)
```
goldfirere, line 256: It's just `value mod contended portable many with 'a`. That works actually for two different reasons:

1. `value` already has `mod non_null` in it.
2. `with`-bounds don't affect nullability, by special decree. (This is in fact implemented.)

One possible way forward is to move nullability into the layout. But I think we probably won't do that in time for release.

jane/doc/extensions/kinds.md
```diff
@@ -0,0 +1,335 @@
...
+# Non-modal kind axes
+
+## Externality
+
+The externality axis records whether all a type's values may safely be ignored
+by the GC.  This may be because they are OCaml "immediates" (values represented
+by a tagged integer), because they are unboxed types like `float#` or `int32#`,
+or because they are allocated elsewhere.
+
+The axis has two possible values, with `external < internal`.  Here, `external`
```
goldfirere, line 267: Just remembered: it's `external_`, not `external`.

goldfirere, line 267: It actually has 3 possible values: `external_ < external64 < internal`, so that `immediate64` can be `mod external64`.

jane/doc/extensions/kinds.md
```diff
@@ -0,0 +1,335 @@
...
+In the future, we plan to make externality a mode, rather than just a property
+of types.
+
+## Nullability
+
+The nullability axis records whether `NULL` (the machine word 0) is a possible
+value of a type, and is used to support the non-allocating option `'a or_null`
+type. The axis has two possible values, with `non_null < maybe_null`.
+
+A type may be `non_null` if none of its value values are `NULL`.  Such types are
```
goldfirere, line 284:
```suggestion
A type may be `non_null` if none of its values are `NULL`.  Such types are
```

jane/doc/extensions/kinds.md
```diff
@@ -0,0 +1,335 @@
...
+compatible with `or_null`, whose definition is:
+```ocaml
+type ('a : value mod non_null) or_null : value mod maybe_null =
```
goldfirere, line 287: `value mod maybe_null` doesn't work. Instead, I think it's `value_maybe_null`, but @dkalinichenko-js knows best.

jane/doc/extensions/kinds.md
```diff
@@ -0,0 +1,335 @@
...
+Instead, for past axes, a modal bound is _lower bound_  on expectations: `mod
+shared` means that even if you're expecting to get an `uncontended` thing of
+this type, a `shared` thing is just as good.
+
+## Kind inference
+
+CR ccasinghino: I'm imagining the "with kinds" section above to be a practical
+introduction to why we have with kinds and how to write them. It probably should
+not cover the details of how we compute the kind of type. Is it worth explaining
+that down here?
```
goldfirere, line 335: Not in great detail. But I do think there should be some commentary mentioning at least these highlights:

* The default kind for `type t` is `value`
* The kind for an extensible type is `value`
* The kind for a boxed variant with no fields is `immediate`
* The kind for a boxed record or variant with no mutable fields (but at least one field) is `immutable_data with <<all the fields>>`
* The kind for a boxed record or variant with at least one mutable field is `mutable_data with <<all the fields>>`
* The kind for an unboxed record is `<<product of the fields' layouts>> mod <<everything>> with <<all the fields>>`
* The kind for an `[@@unboxed]` record or variant is `<<layout of the payload>> mod <<everything>> with <<the field>>` (by further special decree, this `with` *does* take nullability into account!)
* The kind for an unboxed tuple is `<<layout of the components>> mod <<everything>> with <<the components>>`
* The kind for a closed polymorphic variant with no fields is `immediate`
* The kind of a function is `value mod aliased contended`
* The kind for other types is `value`
* Kinds for other types may improve over time. For example, we expect to change the kind of a boxed tuple to be `immutable_data with <<the components>>`. Similarly, we expect to improve on the kinds of polymorphic variants and first-class modules.
* The kind of a private abbreviation is the kind of its right-hand side. We expect to allow users to give a private abbreviation a superkind of its right-hand side (thus improving abstraction) soon.

A full specification of how with-bounds interact with subsumption is probably beyond the scope of this document. I can produce some hand-wavy text about how we try our best here if you like.

goldfirere (commented):
I think this is fantastic.

jane/doc/extensions/kinds.md
```diff
@@ -0,0 +1,335 @@
...
+This kind indicates that `int` mode crosses on all five of our modal axes. The
+non-modal bounds `external` and `non_null` capture two other properties of
+`int`s: `external` describes types that can safely be ignored by the OCaml
+garbage collector, and `non_null` describes types that do not have `NULL` as a
+valid member. These non-modal axes are described in more detail below.
+
+# The meaning of kinds
+
+In additional to `value`, Oxcaml supports layouts like `float64` (unboxed
+floating point numbers that are passed in int registers), `bits64` and `bits32`
```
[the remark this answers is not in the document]
ccasin: Going with "general-purpose registers" for maximum precision.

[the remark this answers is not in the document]
lthls: Isn't "general-purpose registers" a bit misleading ? Usually it refers to the registers used for integers and pointers, not the SIMD registers that we use (at least on x86) for `float64`.

jane/doc/extensions/kinds.md
ccasin: 🤦‍♂️

I added a note about external64 and jsoo/wasm - you may want to check its accuracy.

jane/doc/extensions/kinds.md
ccasin: I've added a paragraph explaining what "maximum" means here - let me know if you think that covers it.

jane/doc/extensions/kinds.md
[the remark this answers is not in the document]
ccasin: Yes, this was obviously wrong, sorry.  A comedy of errors really - writing `int` registers here initially was just a typo/brain-fart.  Then Richard presumably suggested being more precise without noticing the error.  Then I tried to be even more precise but still didn't notice the error.

Anyway, fixed now, thanks for noticing.

ccasin:
OK, I've integrated most of the feedback here - thanks for catching all those typos.  A couple notes:

**Nullability**: I've reworked the doc to just not mention nullability at all until the section on it.  Despite my earlier whinging, I think this works out fine, and everything remains accurate, likely due to the good design choice to have things as they are.  @dkalinichenko-js may want to check the nullability section for accuracy.

**With-kinds**: This is the remaining thing. I've realized I have no idea what the current syntax is.  I propose we ask someone familiar with the current implementation (@glittershark , maybe?) to fill in that part of this doc.  (You can just push to this branch)

glittershark:
> **With-kinds**: This is the remaining thing. I've realized I have no idea what the current syntax is. I propose we ask someone familiar with the current implementation (@glittershark , maybe?) to fill in that part of this doc. (You can just push to this branch)
I'll gladly do this, ideally sometime next week

liam923:
> I'll gladly do this, ideally sometime next week (or if you want it sooner, @liam923 or @goldfirere might have time before then?)

I don't think I'd be able to get around to it before next week.

ccasin (commented):
Some small comments on @goldfirere's additions

jane/doc/extensions/kinds/syntax.md
```diff
@@ -0,0 +1,508 @@
...
+module M : sig
+  type 'a t
+end = struct
+  type 'a t = 'a
+end
+```
+
+The kind chosen for `'a t` in the signature is `value` (as described below), and
+thus inferring `'a : any` in the signature would not be safe: it would allow us
+to smuggle, say, a `float64` type where only `value`s are expected.
```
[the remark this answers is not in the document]
goldfirere: Better now?

jane/doc/extensions/kinds/syntax.md
```diff
@@ -0,0 +1,508 @@
...
+`non_null` -- unless you write a kind annotation saying otherwise (like `type
+('a : any) array`). However, even if you write a kind annotation, we still
+perform inference. For example, if you write `type ('a : any) t = K of 'a
+t_port`, then we'll infer `'a : value mod portable`. The annotation is not
+ignored, but instead causes the initial assumed kind of `'a` to be `any`,
+allowing that kind to be lowered by usages of `'a`.
+
+If type inference ends and we still do not know what layout should be used for
+the parameter (this can happen only without an annotation), we default the
+layout to `value_non_null`.
```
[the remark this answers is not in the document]
goldfirere: Indeed. I don't know what I was thinking.

jane/doc/extensions/kinds/syntax.md
```diff
@@ -0,0 +1,508 @@
...
+  
+let bar (type a : value mod portable) (x : a) = ... 
+  (* no extra parentheses needed when there is only one locally abstract type *)
+```
+
+It's legal to omit the `mod ...` or `with ...` portion of the kind when you only
+want to specify the layout (as we did in a couple of these examples). The bounds
+of unspecified axes are the maximum on those axes.
+
+## Syntax reference
```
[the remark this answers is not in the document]
goldfirere: Done. Good suggestion

PR #3563, opened 2025-02-10
Per-axis jkind errors

liam923:
This PR adds a section to subjkind errors when either kind's with-bounds requires a modality to be printed. The section lists each axis along which there was disagreement, and what the jkinds are along those axes. The section is only printed when the layouts extension is in alpha in order to match the behavior of printing with-bounds. (If no with-bounds are ever printed, then no modalities ever are.)


typing/jkind.ml
```diff
@@ -2464,11 +2518,89 @@ module Violation = struct
...
+              (With_bounds.to_list jkind.jkind.with_bounds)
+          in
+          jkind_has_modalities sub || jkind_has_modalities super
+        in
+        match disagreeing_axes, has_modalities with
+        | None, _ | _, false -> ()
+        | Some disagreeing_axes, true ->
+          (* CR: @\n is discouraged by the documentation, but @;@; seems to emit one newline
+             and then one space rather than two newlines *)
+          fprintf ppf "@\n@\nThe first mode-crosses less than the second along:";
```
liam923, line 2564: I couldn't figure out the "right" way to get a blank line separating the stanzas.

glittershark: I think this is probably fine for now.

glittershark (approved):

PR #3564, opened 2025-02-10
handle unsupported null constants like float32 constants

dkalinichenko-js:
Like float32 constants, null constants are not supported by the upstream bytecode marshalling functions. Since we run `flambda_backend/tests` using the stage 1/bootstrap compiler using the upstream runtime, moving `or_null` to stable runs into this problem again.

Eventually, we should just build those tests using the final compiler. But for now, it's easier to do the same trick for `or_null`.

Tested by https://github.com/ocaml-flambda/flambda-backend/pull/3565.

### DIFF
--- a/bytecomp/bytegen.ml
+++ b/bytecomp/bytegen.ml
@@ -768,23 +768,29 @@ module Storer =
    cont = list of instructions to execute afterwards
    Result = list of instructions that evaluate exp, then perform cont. *)
 
-(* We cannot use the [float32] type in the compiler. *)
+(* CR dkalinichenko: this error happens because we run tests
+   under [flambda_backend/tests] with the stage 1 compiler instead of the
+   final compiler. Run them using the final compiler.*)
+(* We cannot use the [float32] or [or_null] types in the compiler. *)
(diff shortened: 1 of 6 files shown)


bytecomp/bytegen.ml
```diff
@@ -768,23 +768,29 @@ module Storer =
    cont = list of instructions to execute afterwards
    Result = list of instructions that evaluate exp, then perform cont. *)
 
-(* We cannot use the [float32] type in the compiler. *)
+(* CR dkalinichenko: this error happens because we run tests
+   under [flambda_backend/tests] with the stage 1 compiler instead of the
+   final compiler. Run them using the final compiler.*)
+(* We cannot use the [float32] or [or_null] types in the compiler. *)
 external float32_is_stage1 : unit -> bool = "caml_float32_is_stage1"
```
TheNumbat, line 775: Perhaps this should be renamed to not imply it's specific to float32

TheNumbat (approved):
Are there particular tests in 3565 that this fixes?

dkalinichenko-js:
> Are there particular tests in 3565 that this fixes?

All tests under `flambda-backend/tests` fail when trying to load `stdlib_stable` otherwise.

PR #3569, opened 2025-02-12
Remove `build-resolved-files-for-ci`, the script that checked the build during the ocaml 5 merge

riaqn:
Remove `build-resolved-files-for-ci` (introduced in https://github.com/ocaml-flambda/ocaml-jst/pull/189) since we no longer need it.


ncik-roberts (approved):
Looks good &mdash; I think the deleted technique might be useful for merges we do in the future, so I've SEO'ed the PR title a bit more.

PR #3571, opened 2025-02-12
Easy compare script

jvanburen:
Added a script to allow you to diff old versions of the compiler with new versions.
This also involved adding `boot-install` to the makefile, so that the bootstrapped compiler can be installed.

### DIFF
--- a/Makefile.common-jst
+++ b/Makefile.common-jst
@@ -441,13 +441,13 @@ debug-printers: runtime-stdlib # required for $(ws_main) to work
 	@echo To load into ocamldebug, use:
 	@echo source \"$$(realpath _build/main/$(ocamldir)/tools/debug_printers)\"
 
-# boot-install installs the bootstrapped compiler into _install. It is mostly
+# boot-_install installs the bootstrapped compiler into _install. It is mostly
 # just copied from the install phrase.
 # This allows tools/compare.sh to function as expected, which is useful because
 # it allows us to easily run the bootstrapped compiler, especially with
 # tools/compare.sh
-.PHONY: boot-install
-boot-install: runtime-stdlib
+.PHONY: boot-_install
+boot-_install: runtime-stdlib
 	rm -rf _install
 	mkdir -p _install/{bin,lib/ocaml}
 	$(cpl) _build/runtime_stdlib_install/bin/* _install/bin/
@@ -460,9 +460,10 @@ boot-install: runtime-stdlib
 	rm -f _install/lib/ocaml/compiler-libs/*.cmo
 	mkdir _install/lib/stublibs
 	find _build/default -name "dll*.so" \
-          -exec cp -f -t _install/lib/stublibs {} \+
+       -print0 | xargs -0 -I '{}' cp -f '{}' _install/lib/stublibs
 	mkdir -p _install/lib/ocaml/compiler-libs
 	cp _build/default/parser.cmly _install/lib/ocaml/compiler-libs/
 	find _build/default/ \( -name "flambda2*.cmi" \
           -or -name "flambda2*.cmti" -or -name "flambda2*.cmt" \) \
-          -exec cp -f -t _install/lib/ocaml/compiler-libs {} \+
+          -print0 | xargs -0 -I '{}' cp -f '{}' _install/lib/ocaml/compiler-libs
+
--- a/Makefile.common-jst
+++ b/Makefile.common-jst
@@ -440,3 +440,29 @@ debug-printers: runtime-stdlib # required for $(ws_main) to work
 	@echo
 	@echo To load into ocamldebug, use:
 	@echo source \"$$(realpath _build/main/$(ocamldir)/tools/debug_printers)\"
+
+# boot-install installs the bootstrapped compiler into _install. It is mostly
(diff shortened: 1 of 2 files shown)


Makefile.common-jst
```diff
@@ -440,3 +440,29 @@ debug-printers: runtime-stdlib # required for $(ws_main) to work
 	@echo
 	@echo To load into ocamldebug, use:
 	@echo source \"$$(realpath _build/main/$(ocamldir)/tools/debug_printers)\"
+
+# boot-install installs the bootstrapped compiler into _install. It is mostly
+# just copied from the install phrase.
+# This allows tools/compare.sh to function as expected, which is useful because
+# it allows us to easily run the bootstrapped compiler, especially with
+# tools/compare.sh
+.PHONY: boot-install
```
dkalinichenko-js, line 449: The target name `boot-install` is inconsistent with the target that installs the normal compiler into the `_install` directory, which is named `_install`.

Makefile.common-jst
```diff
@@ -440,3 +440,29 @@ debug-printers: runtime-stdlib # required for $(ws_main) to work
...
+	$(cpl) _build/_bootinstall/bin/* _install/bin/
+	$(cpl) -R _build/runtime_stdlib_install/lib/ocaml_runtime_stdlib/* \
+    _install/lib/ocaml/
+	rm -f _install/lib/ocaml/{META,dune-package,Makefile.config,dynlink.cmxa}
+	$(cpl) -R _build/_bootinstall/lib/ocaml/* _install/lib/ocaml/
+	rm -f _install/lib/ocaml/{META,dune-package}
+	rm -f _install/lib/ocaml/compiler-libs/*.cmo
+	mkdir _install/lib/stublibs
+	find _build/default -name "dll*.so" \
+          -exec cp -f -t _install/lib/stublibs {} \+
```
dkalinichenko-js, line 463: The `-t` flag is unavailable on MacOS.

tools/compare.sh
```diff
@@ -0,0 +1,87 @@
...
+
+# Use this script to compare the output of an existing compiler
+# with one installed to _install
+#
+# E.g.,
+# make boot-install
+# export OLD='opam exec --switch=MY_OLD_SWITCH ocamlopt.opt --'
+# tools/compare.sh testsuite/tests/lib-int/test.ml -c -dcmm -dump-into-file
+# diff {old,new}.cmx.dump
+
```
dkalinichenko-js, line 11:
```suggestion

set -eu -o pipefail
```

tools/compare.sh
```diff
@@ -0,0 +1,87 @@
...
+    shift
+
+    if [[ ${#LIB[@]} -gt 0 || ${#INCLUDE[@]} -gt 0 ]]; then
+        if [[ -z ${stdlib+x} ]]; then
+            stdlib="$("${cmd[@]}" -config-var standard_library)" || return 1
+        fi
+
+        local include
+        for include in "${INCLUDE[@]}"; do
+            cmd+=( "${stdlib}/${lib}/${lib}.cmxa" )
```
dkalinichenko-js, line 63: This looks wrong, it should reference `${include}`?

dkalinichenko-js (changes requested):

tools/compare.sh
```diff
@@ -0,0 +1,87 @@
...
+        done
+
+        local lib
+        for lib in "${LIB[@]}"; do
+            cmd+=( '-I' "${stdlib}/${lib}/${lib}.cmxa" )
```
dkalinichenko-js, line 68: This also looks wrong, `-I` takes a directory argument, not a file.

tools/compare.sh
```diff
@@ -0,0 +1,87 @@
...
+    # Set the binary as the first part of the command
+    read -a cmd <<<"$1"
+    shift
+
+    if [[ ${#cmd[@]} -eq 0 ]]; then
+        echo 'no compiler specified' >&2
+        return 1
+    fi
+
+    local stdlib="$2"
```
dkalinichenko-js, line 53: Shouldn't it be `$1` after `shift`?

tools/compare.sh
```diff
@@ -0,0 +1,149 @@
...
+        elif [[ ${#parsed[@]} -eq 1 && -d ${parsed[0]} ]]; then
+             # automagically add BIN if we've specified a directory
+            parsed=( "${parsed[0]}/${BIN}" )
+        fi
+    else
+        parsed=("$(type -P "${BIN}")") || fail
+    fi
+
+    # overwrite VAR_NAME with the parsed version
+    declare -a -g "$1"=( "${parsed[@]}" )
```
dkalinichenko-js, line 64: I'm getting `tools/compare.sh: line 64: syntax error near unexpected token '('` when trying to run 
```bash
make boot-_install
export OLD='opam exec --switch=5.2.0 ocamlopt.opt --'
tools/compare.sh testsuite/tests/lib-int/test.ml -c -dcmm -dump-into-file
diff {old,new}.cmx.dump
```
from a clean install.

jvanburen: fixed!

dkalinichenko-js (approved):
LGTM.

PR #3572, opened 2025-02-12
removed unary negation from flambda

jvanburen:
Unary negation just adds complexity - it's generally translated to `0 - arg` in cmm anyway, so there's no real gain.

The only potential loss is performace on tagged integer negation, but that will be made up for soon, and I don't think it's that big of a deal...

I'm removing this now to make adding small integers easier soon.


mshinwell:
I'll review this

Gbury (approved):

PR #3576, opened 2025-02-13
Remove the Unnecessary_allow_any_kind warning

glittershark:
This became unused when the attributes went away, but I forgot to remove the
warning.


goldfirere (approved):
I am unreasonably sad by losing a number to this very temporary warning. I will console myself with the realization that there are lots of numbers.

PR #3583, opened 2025-02-17
Fix `stack_` precedence

riaqn:
Currrently `stack_ a#b` does not parse, but it should because `a#b` is a `simple_expr`. This is because `stack_` has too high precedence.

This PR fixes that by some light refactor. 
- The diff to the [upstream parser](https://github.com/ocaml/ocaml/blob/trunk/parsing/parser.mly) in order to support `stack_ fun` etc. is now more local. The changes being local also allows easier porting to `ocamlformat`'s extended parser, which has a different structure (https://github.com/janestreet/ocamlformat/pull/103)
- `stack_ fun ..` has the same precedence as `fun`. Similar for `stack_ function ..` and `stack_ A a b`  This is done implicitly by menhir, because each rule has its precedence equal to its rightmost terminal symbol.
- `stack_ simple_expr` has the same precedence as constructor application.


testsuite/tests/parsetree/source_jane_street.ml
```diff
@@ -775,6 +776,25 @@ Line 7, characters 17-18:
 Error: This expression is not an allocation site.
 |}]
 
+let unary_minus_plus () =
+  let a = stack_ (-x) in
```
dvulakh, line 780: Can we make `x` an `int` so that we get a type error indicating we correctly parsed `stack_`?

dvulakh (changes requested):

dvulakh (approved):
Looks good, thanks

PR #3585, opened 2025-02-17
Runtime events should never be conditional on debug log level

NickBarnes:
While working on making the runtime events system more useful, I discovered that major GC cycle runtime event recording is conditional on (among other things) the GC log level. This can't be right.
Digging into git history I find that the runtime events used to be [string-based and very slow](https://github.com/ocaml-flambda/flambda-backend/blob/2897e91b5fc6758a2ff5d35d916ea5e635016f14/byterun/eventlog.c#L225-L229). Now they are very whizzy.

### DIFF
--- a/runtime/major_gc.c
+++ b/runtime/major_gc.c
@@ -1924,37 +1924,26 @@ static void major_collection_slice(intnat howmuch,
   uintnat saved_major_cycle = caml_major_cycles_completed;
   intnat budget;
 
+  /* shortcut out if there is no opportunistic work to be done */
+  if (mode == Slice_opportunistic &&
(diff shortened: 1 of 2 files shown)


NickBarnes:
Rebased and tidied up following in-person input from @stedolan, but leaving this in "draft" for now, as we need to measure, and be aware of, the performance effect of runtime events in opportunistic slices before merging it.

kayceesrk:
IIRC @sadiqj observed that this particular event can be quite spammy and hence had it behind a log-level.

NickBarnes:
Superseded by #7085.

PR #3586, opened 2025-02-17
fix probe handler empty parameters

riaqn:
Currently, when one write a probe without free variables, the probe handler will be a function without parameters. That might be bad - while I can't find specification directly forbidding that, the definitionf of `nlocal` assumes non-empty parameters and would raise in #2985 .

This PR fixes this issue by inserting a dummy parameter when there is no parameter.

### DIFF
--- a/lambda/translcore.ml
+++ b/lambda/translcore.ml
@@ -1152,6 +1152,21 @@ and transl_exp0 ~in_new_scope ~scopes sort e =
           (* Might be a module, which are all values.  Otherwise raise. *)
           ignore (Env.find_module_lazy path e.exp_env)
       ) arg_idents;
+      let params, ap_args =
+        match param_idents with
+        | [] ->
+            [{name = Ident.create_local "unit";
+              layout = layout_probe_arg;
+              attributes = Lambda.default_param_attribute;
+              mode = alloc_local}]
+          , [Lconst (Const_base (Const_int 0))]
+        | _ :: _ ->
+            List.map (fun name ->
+              { name; layout = layout_probe_arg;
+                attributes = Lambda.default_param_attribute;
+                mode = alloc_local }) param_idents
+          , List.map (fun id -> Lvar id) arg_idents
+      in
       let body = Lambda.rename map lam in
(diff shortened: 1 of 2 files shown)


lambda/translcore.ml
```diff
@@ -1152,6 +1152,21 @@ and transl_exp0 ~in_new_scope ~scopes sort e =
...
           ignore (Env.find_module_lazy path e.exp_env)
       ) arg_idents;
+      let params, ap_args =
+        match param_idents with
+        | [] ->
+            [{name = Ident.create_local "unit";
+              layout = layout_probe_arg;
+              attributes = Lambda.default_param_attribute;
+              mode = alloc_local}]
+          , [Lconst (Const_base (Const_int 0))]
```
gretay-js, line 1162:
```suggestion
          , [lambda_unit]
```

lambda/translcore.ml
```diff
@@ -1152,6 +1152,21 @@ and transl_exp0 ~in_new_scope ~scopes sort e =
           (* Might be a module, which are all values.  Otherwise raise. *)
           ignore (Env.find_module_lazy path e.exp_env)
       ) arg_idents;
+      let params, ap_args =
+        match param_idents with
+        | [] ->
+            [{name = Ident.create_local "unit";
+              layout = layout_probe_arg;
+              attributes = Lambda.default_param_attribute;
+              mode = alloc_local}]
```
gretay-js, line 1161: It'd be nice to factor this part out of the match to avoid duplication:

```
let make_param name = { 
   name; 
   layout = layout_probe_arg;               
   attributes = Lambda.default_param_attribute;
   mode = alloc_local }
in
...
```

lthls:
Note that the upstream compiler, with Flambda, can generate function with zero parameters (when all parameters are specialised). So the native backend at least should be able to handle it.

This PR deals with Lambda though, and I think ensuring all Lambda functions have at least one parameter seems reasonable (I don't think bytecode handles functions without parameters very well).

gretay-js (approved):

PR #3597, opened 2025-02-19
Separate the mechanisms and APIs for dependent memory and custom blocks

stedolan:
Previously, allocating a block with caml_alloc_custom or caml_alloc_custom_mem would both register a custom block finaliser and accelerate the GC.

Now, caml_alloc_custom(_mem) has no effect on GC, and the functions caml_adjust_gc_speed and caml_adjust_minor_gc_speed become noops. The GC speed increase for dependent memory happens only through caml_alloc_dependent_memory and caml_free_dependent_memory.

The function caml_alloc_custom_dep is available to call both caml_alloc_custom and caml_alloc_dependent_memory. However, the user must ensure to call caml_free_dependent_memory in their finaliser. (Bigarrays have already been updated to use this API in a previous patch)

This is to pave the way for a new GC pacing policy. However, no change to pacing is made by this patch - the GC should perform as previously on e.g. bigarray values, which have been ported to use the new API.

(Most of the code in this patch was written by @damiendoligez )

### DIFF
--- a/runtime/caml/domain_state.tbl
+++ b/runtime/caml/domain_state.tbl
@@ -142,7 +142,7 @@ DOMAIN_STATE(int, unique_id)
 DOMAIN_STATE(value, dls_root)
 /* Domain-local state */
 
-/* How much external memory is currenty held by the minor and major heap. */
+/* How much external memory is currenty held by the minor heap. */
 DOMAIN_STATE(uintnat, minor_dependent_bsz)
 
 /* How much work needs to be done (by all domains) before we stop this slice. */
--- a/runtime/caml/domain_state.tbl
+++ b/runtime/caml/domain_state.tbl
@@ -90,6 +90,10 @@ DOMAIN_STATE(uintnat, allocated_words_direct)
 /* Number of words allocated directly to the major heap since the latest
    slice. (subset of allocated_words) */
 
+DOMAIN_STATE(uintnat, allocated_dependent_bytes)
+/* Number of external bytes whose pointers were promoted or allocated in
+   the major heap since latest slice. */
+
 DOMAIN_STATE(uintnat, swept_words)
 
 DOMAIN_STATE(uintnat, major_slice_epoch)
@@ -138,11 +142,8 @@ DOMAIN_STATE(int, unique_id)
 DOMAIN_STATE(value, dls_root)
 /* Domain-local state */
 
-DOMAIN_STATE(double, extra_heap_resources)
-DOMAIN_STATE(double, extra_heap_resources_minor)
-
-DOMAIN_STATE(uintnat, dependent_size)
-DOMAIN_STATE(uintnat, dependent_allocated)
+/* How much external memory is currenty held by the minor and major heap. */
+DOMAIN_STATE(uintnat, minor_dependent_bsz)
 
 /* How much work needs to be done (by all domains) before we stop this slice. */
 DOMAIN_STATE(intnat, slice_target)
@@ -165,6 +166,9 @@ DOMAIN_STATE(struct caml_intern_state*, intern_state)
 DOMAIN_STATE(uintnat, stat_minor_words)
 DOMAIN_STATE(uintnat, stat_promoted_words)
 DOMAIN_STATE(uintnat, stat_major_words)
+DOMAIN_STATE(uintnat, stat_minor_dependent_bytes)
+DOMAIN_STATE(uintnat, stat_promoted_dependent_bytes)
+DOMAIN_STATE(uintnat, stat_major_dependent_bytes)
 DOMAIN_STATE(intnat, stat_forced_major_collections)
 DOMAIN_STATE(uintnat, stat_blocks_marked)
 
--- a/runtime/custom.c
+++ b/runtime/custom.c
@@ -51,22 +51,8 @@ mlsize_t caml_custom_get_max_major (void)
          * caml_custom_major_ratio;
 }
 
-/* [mem] is an amount of out-of-heap resources, in the same units as
-   [max_major] and [max_minor]. When the cumulated amount of such
-   resources reaches [max_minor] (for resources held by the minor
-   heap) we do a minor collection; when it reaches [max_major] (for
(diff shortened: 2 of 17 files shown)


runtime/custom.c
```diff
@@ -108,23 +86,12 @@ Caml_inline mlsize_t get_max_minor (void)
...
-  return alloc_custom_gen (ops, bsz, mem, max_major, max_minor, 1, local);
-}
-
 CAMLexport value caml_alloc_custom(const struct custom_operations * ops,
                                    uintnat bsz,
                                    mlsize_t mem,
                                    mlsize_t max)
 {
-  return caml_alloc_custom0(ops, bsz, mem, max, 0);
+  return alloc_custom_gen(ops, bsz, /* minor_ok: */ 1, /* local: */ 0);
```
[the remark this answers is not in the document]
NickBarnes: Ah, no. The proxy block is already sampled, and this is now the path for custom blocks without foreign memory.

runtime/caml/domain_state.tbl
```diff
@@ -138,11 +142,8 @@ DOMAIN_STATE(int, unique_id)
 DOMAIN_STATE(value, dls_root)
 /* Domain-local state */
 
-DOMAIN_STATE(double, extra_heap_resources)
-DOMAIN_STATE(double, extra_heap_resources_minor)
-
-DOMAIN_STATE(uintnat, dependent_size)
-DOMAIN_STATE(uintnat, dependent_allocated)
+/* How much external memory is currenty held by the minor and major heap. */
```
NickBarnes, line 145: s/and major//

runtime/major_gc.c
```diff
@@ -664,21 +665,22 @@ static void update_major_slice_work(intnat howmuch,
                                     int may_access_gc_phase)
 {
   double heap_words;
-  intnat alloc_work, dependent_work, extra_work, new_work;
-  intnat my_alloc_count, my_alloc_direct_count, my_dependent_count;
+  intnat alloc_work, extra_work, new_work;
+  intnat my_alloc_count, my_alloc_direct_count;
```
NickBarnes, line 669: I'm only skimming this function as I review this PR because I know it's about to change radically (in the next week or so).

runtime/memory.c
```diff
@@ -238,50 +238,47 @@ CAMLexport CAMLweakdef void caml_modify (volatile value *fp, value val)
...
+  if (Is_young (v)){
+    Caml_state->stat_minor_dependent_bytes += nbytes;
+    add_to_dependent_table (&Caml_state->minor_tables->dependent, v, nbytes);
+    Caml_state->minor_dependent_bsz += nbytes;
+    if (Caml_state->minor_dependent_bsz >
+          Bsize_wsize (Caml_state->minor_heap_wsz)
+          / 100 * caml_custom_minor_ratio){
+      caml_request_minor_gc ();
+    }
+  }else{
```
NickBarnes, line 252: horrible whitespace in this function.

runtime/memory.c
```diff
@@ -238,50 +238,47 @@ CAMLexport CAMLweakdef void caml_modify (volatile value *fp, value val)
...
+    caml_add_dependent_bytes (Caml_state->shared_heap, nbytes);
+    Caml_state->allocated_dependent_bytes += nbytes;
+    /* FIXME sdolan: what's the right condition here? */
```
NickBarnes, line 255: An excellent question. I don't really believe in any of this caml_custom_get_max_major() stuff. "2/3 of 44% of the heap size, divided by 5". I keep expecting to subtract the number I first thought of and tada: it's the playing card hidden in the orange.

NickBarnes (changes requested):
LGTM, modulo the pacing code which is about to change. One comment is just plain wrong, and the FIXME asks the important question (which I'm sure will be answered by the next PR).

runtime/custom.c
```diff
@@ -51,22 +51,8 @@ mlsize_t caml_custom_get_max_major (void)
...
-   If [max_major] is 0, then [mem] is a number of bytes and the actual
-   limit is [caml_custom_get_max_major ()] computed at the
-   time when the custom block is promoted to the major heap.
-*/
 static value alloc_custom_gen (const struct custom_operations * ops,
                                uintnat bsz,
-                               mlsize_t mem,
-                               mlsize_t max_major,
-                               mlsize_t max_minor,
                                int minor_ok,
```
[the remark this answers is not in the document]
stedolan: Yeah, good idea.

runtime/major_gc.c
stedolan: The only claim I'm making here is that I haven't changed the policy, at least in the steady-state case.

Previously, `dependent_allocated` was always zero and `extra_work` went from `0` to `1`, growing by `s / get_custom_max_major()` on off-heap allocation of size `s`. Now instead `allocated_dependent_bytes` is an absolute byte count, and the division by `get_custom_max_major()` happens here instead.

(In the non-steady-state case there's a slight difference: now we're sampling `get_custom_max_major()` at slice time rather than at allocation time, at which point the heap size may be slightly different. I don't think this makes any difference in practice)

runtime/memory.c
stedolan: Hmm, yep

runtime/memory.c
stedolan: Yeah, I think this criterion is wrong, and I believe so does @damiendoligez who changed it to something more sensible in his branch. I'm using the `caml_custom_get_max_major() / 5` number here to match the `Caml_state->extra_heap_resources > 0.2` logic it replaces, in order to make this PR preserve GC policy - I'd like to separate the rearrange-code and the change-policy parts of this work.

(The orange will be peeled in the next PR)

NickBarnes (approved):
Sorted.

NickBarnes:
> Sorted.

Lol, not sorted in fact.

stedolan:
I believe it is sorted now? The build failure was a bug introduced by #3376 and fixed in #3614

PR #3601, opened 2025-02-19
Reimplement Axis_set in terms of a bitfield

glittershark:
Reimplement Jkind_axis.Axis_set.t in terms of a bitfield, rather than a record
with 7 booleans, for improved performance. This is showing up as rather hot in a
profile, so it's worth microoptimizing.

### DIFF
--- a/typing/jkind_axis.ml
+++ b/typing/jkind_axis.ml
@@ -131,7 +131,7 @@ module Axis = struct
     (* A functor to add some convenient functions to modal axes *)
     include M
 
-    let less_or_equal a b : Misc.Le_result.t =
+    let[@inline] less_or_equal a b : Misc.Le_result.t =
       match le a b, le b a with
       | true, true -> Equal
       | true, false -> Less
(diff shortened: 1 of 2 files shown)


typing/jkind_axis.ml
```diff
@@ -407,55 +407,71 @@ module Axis_collection = struct
...
-  (* each axis is true or false to indicate membership  *)
-  type t = bool Axis_collection.t
+  (* This could be [bool Axis_collection.t], but instead we represent it as a bitfield for
+     performance (this matters, since these are hammered on quite a bit during with-bound
+     normalization) *)
 
-  (* TODO: this could be represented with a uint8 since there's only 7 possible members *)
+  type t = int
+  (* CR layouts: if we had unboxed types in the compiler, this could be represented with a
+     uint8 since there are only 7 bits that we care about *)
```
liam923, line 416: *8 bits, now that we have yielding

typing/jkind_axis.ml
```diff
@@ -407,55 +407,71 @@ module Axis_collection = struct
...
+    | false -> t land lnot (axis_bit axis)
 
-  let singleton axis = add empty axis
+  let empty = 0
 
-  let remove t axis = Axis_collection.set ~axis t false
+  let[@inline] add t axis = set ~axis ~to_:true t
 
-  let mem t axis = Axis_collection.get ~axis t
+  let[@inline] create ~f =
```
liam923, line 437: This doesn't seem to be getting optimized the way we want; let's unroll it?

typing/jkind_axis.ml
```diff
@@ -407,55 +407,79 @@ module Axis_collection = struct
...
+  let[@inline] diff t1 t2 = t1 land lnot t2
+
+  let[@inline] is_subset t1 t2 = Int.equal (t1 land t2) t1
+
+  let[@inline] is_empty t = Int.equal t 0
+
+  let[@inline] complement t = diff all t
+
+  let[@inline] to_seq t =
+    Axis.all |> List.to_seq |> Seq.filter (fun (Axis.Pack axis) -> mem t axis)
```
goldfirere, line 476: Is this function used much? As written, it's calling `List.to_seq` on a constant list every time; this could probably be improved.

typing/jkind_axis.ml
```diff
@@ -407,55 +407,79 @@ module Axis_collection = struct
...
+
+  let[@inline] to_list t = List.of_seq (to_seq t)
```
goldfirere, line 478: This similarly seems less efficient than it could be.

typing/jkind_axis.ml
glittershark: I think these functions are basically only used in printing.

PR #3602, opened 2025-02-19
Fix runtime events deadlock

NickBarnes:
Ported from ocaml/ocaml#13714
Taking more care in the runtime events system to avoid a deadlock.


NickBarnes:
Superseded by #3603

PR #3603, opened 2025-02-19
Fix runtime deadlocks with judicious use of caml_plat_non_blocking

NickBarnes:
Ported from ocaml/ocaml#13227 and ocaml/ocaml#13714.
(note that #13714 reverted parts of #13227 and then took a different path; this combines both upstream PRs).

### DIFF
--- a/runtime/callback.c
+++ b/runtime/callback.c
@@ -447,14 +447,15 @@ static unsigned int hash_value_name(char const *name)
 
 CAMLprim value caml_register_named_value(value vname, value val)
 {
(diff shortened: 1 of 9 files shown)


mshinwell (approved):

stedolan:
I think this introduces a bug into `caml_register_named_value`, since `String_val(name)` is  not valid across `caml_plat_lock_non_blocking`

NickBarnes:
Good catch. The perils of taking upstream bugfixes as read. I'll make a patch tomorrow.

PR #3605, opened 2025-02-19
Define inlined, more-dumb versions of Mod_bounds ops

glittershark:
The mod-bounds ops that used polymorphic functions and first-class modules were
not only a little bit over-abstracted for some peoples' taste, they actually had
noticeably worse performance, due to suboptimal inlining behavior (and
allocation!) of the `Accent_lattice` FCM. This changes them to be dumber and
more repetitive, but more direct, which also gets us a few percent performance
win.


typing/jkind.ml
```diff
@@ -432,40 +438,122 @@ module Mod_bounds = struct
...
+        uniqueness;
+        portability;
+        contention;
+        yielding;
+        externality;
+        nullability
+      } t2 =
+    let locality = Locality.Const.join locality t2.locality in
+    let linearity = Linearity.Const.join linearity t2.linearity in
+    let uniqueness = Uniqueness.Const_op.join uniqueness t2.uniqueness in
```
liam923, line 453: I'm very worried about these subtle `_op`s. You and I both wrote a few bugs in the past couple days due to getting them backwards. Maybe we should do some sort of aliasing so that there's more consistency between axes?

liam923 (commented):
As we discussed in person, I disagree with the thesis that this code is "better". I think it is harder to read and maintain. As of now, I think it is also harder to reason about due to some of the axes being backwards (but I think we can fix this). However, it is clearly more performant, and these are hot paths. So I think this is worth it.

typing/jkind.ml
liam923: Maybe make them all `Jkind_axis.<axis-name>`?

goldfirere (commented):
I don't really love the hybrid approach in this PR: keeping `Mod_bounds` as an `Axis_collection`, but then just inlining a few function calls. Would we get these same performance gains with `[@inline]`? Then we should do that. Or if the goal is to remove the `Axis_collection` abstraction, then go all the way and do it -- possibly recouping even more perf gains.

goldfirere:
@liam923 

> As we discussed in person, I disagree with the thesis that this code is "better". 

I had a long convesation with @lpw25 about this last week -- because he thinks this new simplified code is better, and I disagreed. I continue to find the old more abstract code _easier_ to reason about than the new code, because I find the abstraction of `Axis_collection` to be conceptually straightforward, even with its parameterization with `Indexed`. I can learn that abstraction once, and then each use of a function in it is easy to understand in context.

**But:** this abstraction has been evolving. I know because I spent a bunch of time adding capabilities to it (e.g. `Map2` and `Fold`). And I believe you recently added `Indexed` and the `mono` and `poly` variants. Folks who come across it for the first time have to spend a not-quite-trivial amount of time understanding what's going on and tracing through the meanings of the various pieces. On the other hand, if we had done the simple stuff to begin with, I think the net time spent would be significantly lower, even accounting for several times spending several minutes routinely updating the fields being affected. (Tuning the abstraction is _much_ more fun than repetitively editing field names. We're not optimizing for fun, sadly.)

Until Leo made the point to me about this time cost, I pushed back about which code was better. I continue to think that, if this code were more at rest, `Axis_collection` makes it better. I think some of this just comes down to personal taste: at one point I used `Option.bind` because that just makes more sense to me than a direct match, but Leo found the code more confusing. Perhaps if the majority of the team prefers the simpler code, that really does make it better. In any case, though, in this case, I have to agree that going on without `Axis_collection` is the right call, just because of the time-cost of its maintenance.

(And then, of course, there's the performance aspect!)

glittershark:
I've pushed a (trivial) rebase followed by another commit, https://github.com/ocaml-flambda/flambda-backend/pull/3605/commits/90881e19a972b9b5101eb807d00b482e0888ecbd, which gets rid of `Axis_collection` entirely, at @goldfirere's suggestion. perf seems basically equivalent. @goldfirere and @liam923 PTAL

glittershark:
and another rebase, to fix conflicts; the commit to review is now https://github.com/ocaml-flambda/flambda-backend/pull/3605/commits/4002964ebc99078a605452e6784660c8bf7ae08c

typing/jkind.ml
```diff
@@ -383,89 +383,117 @@ let raise ~loc err = raise (Error.User_error (loc, err))
...
+            ~axis:(Pack (Modal (Comonadic Yielding))) (yielding t1)
+            (yielding t2))
+    @@ Sub_result.combine
+         (axis_less_or_equal ~le:Externality.le
+            ~axis:(Pack (Nonmodal Externality)) (externality t1)
+            (externality t2))
+    @@ axis_less_or_equal ~le:Nullability.le ~axis:(Pack (Nonmodal Nullability))
+         (nullability t1) (nullability t2)
+
+  let equal t1 t2 =
```
goldfirere, line 476: Another good place for a pattern-match

typing/jkind.ml
```diff
@@ -856,16 +884,24 @@ module Layout_and_axes = struct
...
-                    then
-                      let (module Bound_ops) = Axis.get axis in
-                      Bound_ops.join b1 b2
-                    else b1)
-              }
-              b1 b2
+            let value_for_axis (type a) ~(axis : a Axis.t) : a =
+              if Axis_set.mem relevant_axes axis
+              then
+                let (module Bound_ops) = Axis.get axis in
```
goldfirere, line 890: Can we do better here? I thought we identified this as an allocate-y operation which might be slow. And it's in a loop here. It's a bit more verbose, but https://github.com/goldfirere/flambda-backend/blob/rae/dismantle-axis-collection/typing/jkind.ml#L1654 has another way.

typing/jkind.ml
```diff
@@ -383,89 +383,117 @@ let raise ~loc err = raise (Error.User_error (loc, err))
...
+    let uniqueness = Uniqueness.meet (uniqueness t1) (uniqueness t2) in
+    let portability = Portability.meet (portability t1) (portability t2) in
+    let contention = Contention.meet (contention t1) (contention t2) in
+    let yielding = Yielding.meet (yielding t1) (yielding t2) in
+    let externality = Externality.meet (externality t1) (externality t2) in
+    let nullability = Nullability.meet (nullability t1) (nullability t2) in
+    create ~locality ~linearity ~uniqueness ~portability ~contention ~yielding
+      ~externality ~nullability
+
+  let less_or_equal t1 t2 =
```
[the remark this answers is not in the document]
goldfirere: Hmm... I now see why this is hard. I don't have a great idea here. Maybe the best is to point here from the definition of `Mod_bounds` saying that any addition needs to be reflected? Ditto `equal`.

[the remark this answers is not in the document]
glittershark: I realized I forgot to mention this in the commit message, but the choice to make these records abstract and only expose getters and setters was intentional, to pave the road for having the more efficient representation (bit fields) next.

typing/jkind.ml
glittershark: see also https://github.com/ocaml-flambda/flambda-backend/pull/3605#discussion_r1970422090

typing/jkind.ml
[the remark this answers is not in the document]
liam923: I'm also concerned about forgetting new axes here. Here's a not-fun suggestion:
```
val match_axes : t -> (locality:Locality.t -> linearity:Linearity.t -> ... -> 'a) -> 'a
```
Hopefully this would get inlined the way I'd hope.

typing/jkind.ml
liam923: We could also make `value_for_axis` take the join operation as an argument.

typing/jkind.ml
```diff
@@ -1504,19 +1539,26 @@ module Const = struct
...
-                  Typemode.Transled_modifiers.get ~axis parsed_modifiers
-                in
-                let base_bound = Mod_bounds.get ~axis base.mod_bounds in
-                match parsed_modifier, base_bound with
-                | None, base_bound -> base_bound
-                | Some parsed_modifier, base_modifier ->
-                  A.meet base_modifier parsed_modifier.txt)
-          }
+        let value_for_axis (type a) ~(axis : a Axis.t) : a =
+          let (module A) = Axis.get axis in
```
liam923, line 1543: This seems analogous to the situation in normalize

typing/jkind.ml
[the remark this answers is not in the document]
glittershark: I think the eliminator approach is probably the nicest choice here, if it does in fact get inlined - but I am wary of doing it until after we do the bitfield refactor, because it's only then that we know if it's introduced noticeable pessimization. Do you mind if we wait until after that change to do this?

typing/jkind.ml
glittershark: this doesn't end up showing up on a profile, fwiw, for whatever reason. We could make it worse for consistency's sake, but I'm hesitant to?

typing/jkind.ml
glittershark: analogously, this doesn't really show up on a profile, so I'm hesitant to make it more verbose without good reason.

typing/jkind.ml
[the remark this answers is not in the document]
goldfirere: Fine by me, yes

typing/jkind.ml
glittershark: `Accent_lattice` was being referenced from `get_max_axes`, not here - that's gone now, and this is now below threshold on a profile.

typing/jkind.ml
```diff
@@ -497,10 +497,34 @@ module Mod_bounds = struct
 
   (** Get all axes that are set to max *)
   let get_max_axes t =
-    Axis_set.create ~f:(fun ~axis:(Pack axis) ->
-        let (module Axis_ops) = Axis.get axis in
-        let bound = get ~axis t in
-        Axis_ops.le Axis_ops.max bound)
+    let add_if b ax axis_set =
```
goldfirere, line 499: `[@inline]`?

goldfirere (approved):

PR #3609, opened 2025-02-20
Remember with bounds in unboxed products without `-infer-with-bounds`

goldfirere:
This corrects a small oversight. The first commit includes a test that shows that there previously was a soundness bug; the second commit fixes things.

Review: @glittershark 

(The bug was introduced in [this commit](https://github.com/ocaml-flambda/flambda-backend/commit/b7f5a58dd9ed7c951f16fdcae2a1d3dbbc5d16ee#diff-8b6d4083f9de4c12c6a1774cca3f746ae96583c6b2545655c90f809fe41fa413R1240), presumably as we were rebasing.)

### DIFF
--- a/testsuite/tests/typing-jkind-bounds/no_infer_with_bounds.ml
+++ b/testsuite/tests/typing-jkind-bounds/no_infer_with_bounds.ml
@@ -17,14 +17,20 @@ let f (x : #( (int -> int) list * int ref list ) @@ nonportable contended) =
   use_portable x
 
 [%%expect{|
-val f : #((int -> int) list * int ref list) @ contended -> unit = <fun>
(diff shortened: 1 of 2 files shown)


testsuite/tests/typing-jkind-bounds/no_infer_with_bounds.ml
```diff
@@ -0,0 +1,36 @@
+(* TEST
```
glittershark, line 1: you probably need to promote this test now that you've (hopefully) fixed it.

glittershark (approved):
oops!

lgtm, but I'd like to double check the promoted test once you do that.

glittershark (approved):
LGTM!

PR #3610, opened 2025-02-20
Control -infer-with-bounds from an attribute

goldfirere:
This is an idea from @mshinwell to deal with older dunes. To really finish the job, we also need to add the attribute to a bunch of the stdlib. But @glittershark wonders about just removing `-infer-with-bounds` instead, so we'll wait to check perf numbers tonight before proceeding.


goldfirere:
This one has been overtaken by events.

PR #3611, opened 2025-02-21
Remove base kind mode crossing and add `mod everything`

ccasin:
Two commits, best reviewed individually.
- The first makes it so kinds like `bits64`, when written in the source program, no longer imply mode crossing.
- The second adds the ability to write "`mod everything`" to conveniently get the full mode crossing kinds of the built-in types like `int64#`.

I think the only surprise is that there are a couple tests that used to pass only in non-principal mode that now always fail.  I think these are just standard cases where we expect type annotations to be needed for mode crossing to work (even if we are sad about it).

Review: @glittershark or @liam923 ?

### DIFF
--- a/testsuite/tests/typing-modal-kinds/everything.ml
+++ b/testsuite/tests/typing-modal-kinds/everything.ml
@@ -285,3 +285,20 @@ Line 1, characters 22-32:
                           ^^^^^^^^^^
 Error: Unrecognized modality everything.
 |}]
+
+(* Printing *)
+type t : immediate & immediate
+[%%expect{|
+type t : immediate & immediate
+|}]
+
+type t : value & float64 mod everything
+[%%expect{|
+type t : immediate & float64 mod everything
+|}]
+
+type t : value & (immediate & bits64) & float32 mod everything
+[%%expect{|
+type t
+  : immediate & (immediate & bits64 mod everything) & float32 mod everything
+|}]
--- a/testsuite/tests/typing-modal-kinds/everything.ml
+++ b/testsuite/tests/typing-modal-kinds/everything.ml
@@ -1,0 +1,287 @@
+(* TEST
+   expect;
+*)
+
+(* The kinds [value] and [float64] don't mode cross, but the kinds [value mod
+   everything] and [float64 mod everything] cross everything (ignoring
+   nullability) *)
+type t_value : value
+type t_value_mod_e : value mod everything
+type t_float64 : float64
+type t_float64_mod_e : float64 mod everything
+[%%expect{|
+type t_value
+type t_value_mod_e : immediate
+type t_float64 : float64
+type t_float64_mod_e : float64 mod everything
+|}]
+
+(*** locality ***)
+type bad : value mod global = t_value
+[%%expect{|
+Line 1, characters 0-37:
+1 | type bad : value mod global = t_value
+    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Error: The kind of type "t_value" is value
+         because of the definition of t_value at line 1, characters 0-20.
+       But the kind of type "t_value" must be a subkind of value mod global
+         because of the definition of bad at line 1, characters 0-37.
+|}]
+
+type good : value mod global = t_value_mod_e
+[%%expect{|
+type good = t_value_mod_e
+|}]
+
+type bad : float64 mod global = t_float64
+[%%expect{|
+Line 1, characters 0-41:
+1 | type bad : float64 mod global = t_float64
+    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Error: The kind of type "t_float64" is float64
+         because of the definition of t_float64 at line 3, characters 0-24.
+       But the kind of type "t_float64" must be a subkind of float64 mod global
+         because of the definition of bad at line 1, characters 0-41.
+|}]
+
+type good : float64 mod global = t_float64_mod_e
+[%%expect{|
+type good = t_float64_mod_e
+|}]
+
+(*** linearity ***)
+type bad : value mod many = t_value
+[%%expect{|
+Line 1, characters 0-35:
+1 | type bad : value mod many = t_value
+    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Error: The kind of type "t_value" is value
+         because of the definition of t_value at line 1, characters 0-20.
+       But the kind of type "t_value" must be a subkind of value mod many
+         because of the definition of bad at line 1, characters 0-35.
+|}]
+
+type good : value mod many = t_value_mod_e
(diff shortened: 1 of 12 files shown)


typing/jkind.ml
```diff
@@ -972,41 +972,82 @@ module Const = struct
...
-          mk_jkind (Base Float64) ~mode_crossing:true ~nullability:Non_null;
+          mk_jkind (Base Float64) ~mode_crossing:false ~nullability:Non_null;
         name = "float64"
       }
 
+    (* CR layouts v3: change to [Maybe_null] when separability is implemented. *)
+    let kind_of_unboxed_float =
+      { jkind =
+          mk_jkind (Base Float64) ~mode_crossing:true ~nullability:Non_null;
+        name = "float64 mod everything"
```
glittershark, line 983: I don't think names like this are the right thing to do here; these are used in printing verbatim, and the usual thing to do when printing `<layout> mod <modes>` is to put a break hint instead of spaces.

It kinda seems like we might want to not use Builtin for these (the `mod everything` kinds) and pull up a level of abstraction, so we avoid weirdness in printing.

testsuite/tests/typing-modal-kinds/everything.ml
```diff
@@ -0,0 +1,291 @@
...
+|}]
+
+(* mod everything does not affect nullability *)
+type t : value_or_null mod everything
+type bad : value = t
+
+[%%expect{|
+type t
+  : value_or_null mod global aliased many contended portable unyielding
+                      external_
```
glittershark, line 262: I think we should try to print `mod everything` here too

testsuite/tests/typing-modal-kinds/everything.ml
```diff
@@ -0,0 +1,291 @@
...
+|}]
+
+type good : float64 mod external_ = t_float64_mod_e
+[%%expect{|
+type good = t_float64_mod_e
+|}]
+
+(* mod everything does not affect nullability *)
+type t : value_or_null mod everything
+type bad : value = t
```
glittershark, line 257: how does this test that `mod everything` doesn't affect nullability?

typing/typemode.ml
```diff
@@ -95,24 +98,37 @@ module Transled_modifiers =
...
+          linearity = Some { txt = Linearity.Const.min; loc };
+          uniqueness = Some { txt = Uniqueness.Const.max; loc };
+          portability = Some { txt = Portability.Const.min; loc };
+          contention = Some { txt = Contention.Const.max; loc };
+          yielding = Some { txt = Yielding.Const.min; loc };
+          externality = Some { txt = Externality.min; loc };
+          nullability =
+            Transled_modifiers.get ~axis:(Nonmodal Nullability) modifiers_so_far
+        }
   in
```
glittershark, line 132: this seems worth floating out to a top-level declaration? maybe `Transled_modifiers.everything`?

glittershark (changes requested):

testsuite/tests/typing-modal-kinds/everything.ml
ccasin: Sorry, I meant to write a comment about printing: my understanding is that printing jkinds remains a minefield that is due for a substantial rework.  I propose putting off any thought about how to print this well until that rework, I suspect it's not trivial.  Can we compromise on a CR?

typing/jkind.ml
ccasin: I think you've made two points:
* It's not obvious these definitions belong here. I thought about putting them elsewhere, but concluded they want to be in the `all` list so that `subst` memoizes them appropriately. Open to other suggestions, though.
* The names are weird.  Agreed!  But they do sometimes get printed, and this means they get printed as things that parse.  The lack of line breaks is sad, but I'm not sure I see a better alternative short of a larger refactoring of `Builtin`?

goldfirere: I think we should not try to make this great here. `Builtin` was a cheap-and-cheerful (well not so cheerful) approach from its birth, and is mostly sad. It should be removed. So I think if the name with spaces works ok in practice, it's fine for now.

testsuite/tests/typing-modal-kinds/everything.ml
ccasin: I think if `mod everything` set the bound of `t` to `non_null` then the second line would typecheck.

testsuite/tests/typing-modal-kinds/everything.ml
goldfirere: It's less of a minefield now than it was. I think detecting the everything case and using `everything` there would be easy and valuable. I do think we'll revisit this, but this bit of fruit seems low enough to pluck now.

typing/typemode.ml
ccasin: I spent a few minutes on this, but have been defeated by the quantity of abstraction between here and there.  In this module, I can use the concrete options I've written here.

It looks like `Transled_modfiers` is an application of `Jkind_axis.Axis_Collection.Indexed`, which is itself an application of another functor `Indexed_gen`, and in the body of `Indexed_gen` the type of the individual fields of this record are not `const loc option`s but instead are `T.t` where `T` is a parameter whose type is the helpfully named `Misc.T2`, and taht signature defines `t` as abstract.  So, how do I actually build one of these over in that functor?  Also, my head hurts.

Perhaps this is clear to a with-kinds expert - feel free to modify this code and push.

testsuite/tests/typing-modal-kinds/everything.ml
ccasin: OK, I realized the dumbest possible thing works: The reason this one doesn't print as `value_or_null mod everything` is just that that is not one of the kinds in the list `Builtin.all`.  And you might think we can't add all the possible layout/nullability pairs to that list - they are infinite.  But because of how hilariously broken printing for product kinds is (it pushes the modal bounds down to the base layouts), actually it's enough to just enumerate the base ones.

So, I've added the remaining few missing valid base layout/nullability pairs to the list, and now it works.

testsuite/tests/typing-modal-kinds/everything.ml
goldfirere: I agree with Chris

typing/typemode.ml
goldfirere: I went ahead and did this. But actually I think it's better not to: we can't pre-allocated the `everything` node because it contains the location. And I think it's a bit cleaner just to have the definition of everything right where it is. The one change I would make is to use `Const_op.min` for the past axes. (At some point, `Const_op.t` and `Const.t` will become not equal, forcing this change anyway.) I'll push this change.

ccasin: @goldfirere you mentioned pushing a change here.  If you still have that around, could you push it?  I think this is the last piece missing for this PR and it would be good to get it in the next roll.

glittershark (approved):

PR #3613, opened 2025-02-24
Inline ENTER_FUNCTION and LEAVE_FUNCTION on arm64

stedolan:
PR #3376 was just merged but broke the build on arm64, because it used new macros from upstream which have not yet been backported. This patch expands those macros in arm64.S.


NickBarnes (approved):
"WORKSFORME"

PR #3614, opened 2025-02-24
Add EV_MINOR_EPHE_CLEAN to runtime_events

stedolan:
This fixes a bug in the recently-merged #3376, which incorrectly added a new runtime event without updating `otherlibs/runtime_events`.


NickBarnes (approved):
Obviously correct.

PR #3615, opened 2025-02-24
Strip jkind_axis functions out of modules

liam923:
This is a small change that I believe makes `Axis_collection` a bit less cryptic. Rather than writing `.Create.f`, you now write `.create`.


goldfirere:
I think this has been overtaken by #3605, which gets rid of `Axis_collection` entirely.

(Maybe it will come back someday.)

liam923:
Agreed, closing

PR #3616, opened 2025-02-24
Use the same threshold for out-of-cycle work for major and minor dependent allocs

stedolan:
This replaces the strange `caml_custom_get_max_major` computation with something more sensible. The function `caml_custom_get_max_major` is now inlined into its only callsite (which is about to be replaced)


NickBarnes (approved):
An obvious improvement (although looking forward to the imminent deletion of more of this code).

PR #3617, opened 2025-02-24
added type-directed disambiguation of integer literals

jvanburen:
People have complained about the verbosity of unboxed types. This is a quick example attempt at fixing it by using type annotations.

### DIFF
--- a/testsuite/tests/typing-core-bugs/const_int_hint_with_tydi.ml
+++ b/testsuite/tests/typing-core-bugs/const_int_hint_with_tydi.ml
@@ -1,0 +1,104 @@
+(* TEST
+ flags = "-extension type_directed_disambiguation";
+ expect;
+*)
+
+let _ = Int32.(add 1 2);;
+[%%expect{|
+- : int32 = 3l
+|}]
+
+let _ : int32 * nativeint = 42l, 43;;
+[%%expect{|
+- : int32 * nativeint = (42l, 43n)
+|}]
+
+let _ = min 6L 7m;;
+[%%expect{|
+Line 1, characters 15-17:
+1 | let _ = min 6L 7m;;
+                   ^^
+Error: The constant "7m" has type "int" but an expression was expected of type
+         "int64"
+Hint: Did you mean "7L"?
+|}]
+
+let _ = min 6L 7;;
+[%%expect{|
+- : int64 = 6L
+|}, Principal{|
+Line 1, characters 15-16:
+1 | let _ = min 6L 7;;
+                   ^
+Warning 18 [not-principal]: this coercion to int64 is not principal.
+
+- : int64 = 6L
+|}]
+
+(* pattern *)
+let _ : int32 -> int32 = function
+  | 0m -> 0l
+  | x -> x
+[%%expect{|
+Line 2, characters 4-6:
+2 |   | 0m -> 0l
+        ^^
+Error: This pattern matches values of type "int"
+       but a pattern was expected which matches values of type "int32"
+Hint: Did you mean "0l"?
+|}]
+
+let _ : int32 -> int32 = function
+  | 0 -> 0l
+  | x -> x
+[%%expect{|
+- : int32 -> int32 = <fun>
+|}]
+
+let _ : int64 -> int64 = function
+  | 1L | 2m -> 3L
+  | x -> x;;
+[%%expect{|
+Line 2, characters 9-11:
+2 |   | 1L | 2m -> 3L
+             ^^
+Error: This pattern matches values of type "int"
+       but a pattern was expected which matches values of type "int64"
+Hint: Did you mean "2L"?
+|}]
+
+let _ : int64 -> int64 = function
+  | 1L | 2 -> 3L
+  | x -> x;;
+[%%expect{|
+- : int64 -> int64 = <fun>
+|}]
+
+let _ : nativeint -> nativeint = function
+  | 12 -> 0n
+  | x -> x;;
+[%%expect{|
+- : nativeint -> nativeint = <fun>
+|}]
+
+let _ = function
+  | 1L | 2 -> 3L
+  | x -> x;;
+[%%expect{|
+- : int64 -> int64 = <fun>
+|}, Principal{|
+Line 2, characters 9-10:
+2 |   | 1L | 2 -> 3L
+             ^
+Warning 18 [not-principal]: this coercion to int64 is not principal.
+
+- : int64 -> int64 = <fun>
+|}]
+
+let _ : int8 = 129;;
+[%%expect{|
+Line 1, characters 15-18:
+1 | let _ : int8 = 129;;
+                   ^^^
+Error: Integer literal exceeds the range of representable integers of type "int8"
+|}]
--- a/testsuite/tests/typing-layouts/literals.ml
+++ b/testsuite/tests/typing-layouts/literals.ml
@@ -221,6 +221,17 @@ let () = test_char "untagged decimal char" #'\222'
 untagged decimal char: '\222'
 |}]
 
+let () = test_int32 "type_directed_int32" 0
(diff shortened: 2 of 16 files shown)


goldfirere:
I have not read the implementation, but I'm generally in support of type-based disambiguation for literals. Literals are basically just the constructors for built-in types, and so the idea of type-based disambiguation of constructors carries over quite naturally.

For what it's worth, I think upstream is open to this idea, too.

typing/typecore.ml
```diff
@@ -2912,6 +2906,28 @@ let forbid_atomic_field_patterns loc penv (label_lid, label, pat) =
   if Types.is_atomic label.lbl_mut && not (wildcard pat) then
     raise (Error (loc, !!penv, Atomic_in_pattern label_lid.txt))
 
+let disambiguate_integer_literal ~loc env ty_expected i =
+  (* Terrible hack for integer literals *)
```
jra4: As funny as it is, I think we should remove this comment. Disambiguating integer literals is less hacky than for format strings since:
* The integer types are all similar to each other, unlike format strings, which are nothing like strings.
* The disambiguated constant is simple, unlike for format strings, which are disambiguated into a large tree of nested constructors.
* The types that are disambiguated are predefs, unlike for format strings, which disambiguate into a Stdlib type.

jvanburen: Agreed!

utils/language_extension_kernel.mli
```diff
@@ -31,6 +31,7 @@ type _ t =
   | Let_mutable : unit t
   | Layout_poly : maturity t
   | Runtime_metaprogramming : unit t
+  | Type_directed_disambiguation : unit t
```
dkalinichenko-js, line 34: It's a little unclear what this means, since OCaml already has type-directed disambiguation for constructors and arrays. Perhaps `Type_directed_literal_disambiguation`?

testsuite/tests/typing-core-bugs/const_int_hint_with_tydi.ml
dkalinichenko-js, line 1: Should those two files be automatically generated from the same template?

testsuite/tests/typing-layouts/literals.ml
dkalinichenko-js, line 1: I think we need a little more tests. In particular, I'd like to see that:
- Disambiguation doesn't work on literals with explicit suffices
- Boxed vs unboxed versions

typing/typecore.ml
```diff
@@ -936,86 +936,69 @@ let type_constant: Typedtree.constant -> type_expr = function
...
-  | Int16_literal_overflow
-  | Int32_literal_overflow
-  | Int64_literal_overflow
-  | Nativeint_literal_overflow
-  | Int_literal_overflow
-  | Unknown_constant_literal of char
-
-let constant_integer i ~suffix :
-    (constant_integer_result, constant_integer_error) result =
+let parse_integer (scalar : 'a Scalar.Integral.t) i : _ result =
```
dkalinichenko-js, line 939: I think this function should be named something other than `parse_`; the literal is already parsed, what we're doing is converting it into the right constant. Maybe `constant_of_scalar`? Or `constant_of_integer_literal`?

testsuite/tests/typing-layouts/literals.ml
dkalinichenko-js: + interactions of type aliases and type-driven disambiguation

typing/typecore.ml
```diff
@@ -1381,6 +1348,38 @@ let disambiguate_array_literal ~loc env expected_ty =
...
     { ty_elt = None; mut = Mutable }
 
+let disambiguate_integer_literal ~loc env ty_expected i =
+  if not (Language_extension.is_enabled Type_directed_disambiguation)
+  then None
+  else
+    let ty_expected = expand_head env (protect_expansion env ty_expected) in
+    match get_desc ty_expected with
+    | Tconstr (ty_expected_path, _, _)
+      when not (Path.same ty_expected_path Predef.path_int) ->
```
dkalinichenko-js, line 1358: Why `when not (Path.same ty_expected_path Predef.path_int)`? I'd think the solution should be uniform.

typing/typecore.ml
```diff
@@ -1381,6 +1348,38 @@ let disambiguate_array_literal ~loc env expected_ty =
...
+      List.find_map
+        (fun (scalar : _ Scalar.Integral.t) ->
+          let candidate = Predef.path_of_scalar (Scalar.integral scalar) in
+          if Path.same ty_expected_path candidate then (
+            if not (is_principal ty_expected)
+            then
+              Location.prerr_warning loc
+                (not_principal "this coercion to %s" (Path.name candidate));
```
dkalinichenko-js, line 1366: This warning is somewhat cryptic, but this is a pre-existing problem and not worth addressing here.

typing/typecore.ml
```diff
@@ -1381,6 +1348,38 @@ let disambiguate_array_literal ~loc env expected_ty =
...
+  else
+    let ty_expected = expand_head env (protect_expansion env ty_expected) in
+    match get_desc ty_expected with
+    | Tconstr (ty_expected_path, _, _)
+      when not (Path.same ty_expected_path Predef.path_int) ->
+      List.find_map
+        (fun (scalar : _ Scalar.Integral.t) ->
+          let candidate = Predef.path_of_scalar (Scalar.integral scalar) in
+          if Path.same ty_expected_path candidate then (
+            if not (is_principal ty_expected)
```
dkalinichenko-js, line 1363: Let's factor out those two ifs into a helper function.

typing/typecore.ml
```diff
@@ -1381,6 +1348,38 @@ let disambiguate_array_literal ~loc env expected_ty =
   else
     { ty_elt = None; mut = Mutable }
 
+let disambiguate_integer_literal ~loc env ty_expected i =
+  if not (Language_extension.is_enabled Type_directed_disambiguation)
+  then None
+  else
+    let ty_expected = expand_head env (protect_expansion env ty_expected) in
```
dkalinichenko-js, line 1355: Why `protect_expansion`?

dkalinichenko-js (commented):
Thanks @jra4!

typing/typecore.ml
jra4: Without that, `-principal` will emit warnings in cases like these:
```
1 | let _ = [5; 10]
                ^^
Warning 18 [not-principal]: this coercion to int is not principal.
```
I'll move the condition to `if not (is_principal ty_expected)` to make it clearer.

typing/typecore.ml
jra4: As usual, the answer is GADTs. 
```ocaml
(* a.ml *)
type _ t = T : int t

let f (type a) (x : a t) : a = match x with T -> 5
```
Without `protect_expansion`:
```
$ ocamlopt -principal a.ml
File "a.ml", line 3, characters 31-50:
3 | let f (type a) (x : a t) : a = match x with T -> 5
                                   ^^^^^^^^^^^^^^^^^^^
Error: This match expression has type a
       but an expression was expected of type a
       This instance of a is ambiguous:
       it would escape the scope of its equation
```
I don't understand the typechecker well enough to explain why this is the case (or why `-principal` is needed for this example). JVB mimicked the other cases of literal disambiguation when he wrote this: format-string disambiguation also calls `protect_expansion`, and callers of `disambiguate_array_literal` call `generic_instance` on `ty_expected`.

utils/language_extension_kernel.mli
jra4: I'm not too fond of that name either since the OCaml manual calls them array literals [1]. Maybe `Scalar_literal_type_directed_disambiguation`?

\[1]: https://ocaml.org/manual/5.5/arrayliterals.html

PR #3621, opened 2025-02-25
Avoid re-marking ephemerons with trivial data 

stedolan:
Port of ocaml #13827

---

In general, ephemerons may need to be marked more than once, if new marking has occurred since ephemeron marking began. This patch optimises this logic: re-marking is never necessary if the data field of the ephemeron is trivial. This optimisation was present in OCaml 4, and is important for Stdlib.Weak, which only uses ephemerons with trivial data.

### DIFF
--- a/runtime/major_gc.c
+++ b/runtime/major_gc.c
@@ -1271,7 +1271,7 @@ static intnat ephe_mark (intnat budget, uintnat for_cycle,
   mlsize_t size, i;
   caml_domain_state* domain_state = Caml_state;
   int alive_data;
-  intnat marked = 0, made_live = 0;
+  intnat marked = 0, trivial_data = 0, made_live = 0;
 
   CAMLassert(caml_marking_started());
   if (domain_state->ephe_info->cursor.cycle == for_cycle &&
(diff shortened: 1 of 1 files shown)


runtime/major_gc.c
```diff
@@ -1322,28 +1322,42 @@ static intnat ephe_mark (intnat budget, uintnat for_cycle,
...
     budget -= Whsize_wosize(i);
 
-    if (force_alive || alive_data) {
-      if (data != caml_ephe_none && Is_block(data)) {
-        caml_darken (domain_state, data, 0);
-      }
-      Ephe_link(v) = domain_state->ephe_info->live;
-      domain_state->ephe_info->live = v;
-      *prev_linkp = todo;
+    bool keep;
```
NickBarnes, line 1325: I guess I'd initialize this to `true` here, with the comment from line 1339, then lose the `else` block at 1338-1341. Also maybe call it `leave`?

NickBarnes (approved):
LGTM; naming quibble as always.

runtime/major_gc.c
stedolan: Since C compiler warnings these days will reliably complain about use of uninitialised variables, I prefer to declare this kind of thing uninitialised so that the compiler verifies I picked a value for it along every reachable path.

PR #3622, opened 2025-02-25
Add more information to `ocamlobjinfo` output

lukemaurer:
**Recommended to review commit-by-commit.**

There are a couple of useful bits of data missing from `ocamlobjinfo` output, which may be of interest if one's build rules are using it to compute dependency information:

* When an alias of an instance name is declared (with `-no-alias-deps` on), only the head appears in the interface imports of the module, even though resolving the alias will require loading all parts of the instance name.
* Implementation imports are missing instance arguments altogether.

This PR addresses both issues.


testsuite/tests/templates/dunelike/gen-native.sh
```diff
@@ -9,4 +26,8 @@ s/\.cmo/.cmx/g
 s/\.bc/.exe/g
 # Do this last to avoid replacing the _byte:
 s/TEST/TEST (* DO NOT EDIT. Instead edit test_byte.ml and run gen-native.sh. *)/
-' test_byte.ml > test_native.ml
+' test_byte.ml > "$OUTPUT"
+
+if [ -n "$CHECKING" ]; then
+  diff -q test_native.ml "$OUTPUT" && rm "$OUTPUT"
```
riaqn, line 32: This should return non-zero to indicate error?

riaqn (approved):

testsuite/tests/templates/dunelike/gen-native.sh
lukemaurer: It does; I've added a comment since it's not obvious.

PR #3624, opened 2025-02-25
Fix bug in pprintast around type parameters with jkind annotations

goldfirere:
The fix is a bit ugly, but doesn't seem worth making better.

Review: @liam923

### DIFF
--- a/parsing/pprintast.ml
+++ b/parsing/pprintast.ml
@@ -432,11 +432,11 @@ let include_kind f = function
   | Structure -> ()
 
 (* c ['a,'b] *)
-let rec class_params_def ctxt f =  function
+let rec class_params_def f =  function
   | [] -> ()
   | l ->
       pp f "[%a] " (* space *)
-        (list (type_param ctxt) ~sep:",") l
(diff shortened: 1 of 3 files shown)


parsing/pprintast.ml
```diff
@@ -2070,11 +2084,7 @@ and type_extension ctxt f x =
     pp f "@\n|@;%a" (extension_constructor ctxt) x
   in
   pp f "@[<2>type %a%a += %a@ %a@]%a"
-    (fun f -> function
-       | [] -> ()
-       | l ->
-           pp f "%a@;" (list (type_param ctxt) ~first:"(" ~last:")" ~sep:",") l)
```
liam923, line 2076: The line-break behavior is different here - the break comes before the comma, whereas in the other case it comes after. But with this PR, it's now always after. Maybe this is fine, but could you add a test case that shows what it looks like with line breaks?

goldfirere: Looks OK to me.

goldfirere:
Back to you @liam923

liam923 (approved):

PR #3625, opened 2025-02-26
Move `or_null` back to beta

dkalinichenko-js:
Title. Only add the `or_null` type when layouts are at least beta. Layouts like `any_non_null` or `value_or_null` are still available, since the bug only happens when working with `Null` values incorrectly considered impossible by the simplifier.


testsuite/tests/typing-layouts-or-null/stable.ml
```diff
@@ -0,0 +1,14 @@
+(* TEST
+ flags = "-extension-universe stable";
+ expect;
+*)
+
+(* CR dkalinichenko: [or_null] is not available in [stable]. *)
+let x : int or_null = This 3
```
goldfirere, line 7: Let's also have tests to make sure `This` and `Null` are not available

goldfirere:
What about re-exports of `or_null`? Is that something to worry about?

testsuite/tests/typing-layouts-or-null/stable.ml
dkalinichenko-js: Done.

dkalinichenko-js:
Exports won't work either since `or_null` is not added to the environment. I added a test for this.

testsuite/tests/typing-layouts-or-null/basics.ml
```diff
@@ -1,4 +1,5 @@
 (* TEST
+ flags = "-extension layouts_alpha";
```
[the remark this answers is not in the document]
dkalinichenko-js: Fixed.

PR #3626, opened 2025-02-26
Nullability in value_kind

goldfirere:
There should be more tests, but we're moving `or_null` back to beta and will write those tests as the move out of beta.

This includes promotion of two test changes. I'm not sure what's going on here, so we need to make sure this gets reviewed by someone who understands these changes, perhaps @lthls.

### DIFF
--- a/testsuite/tests/syntactic-arity/max_arity_locals.compilers.reference
+++ b/testsuite/tests/syntactic-arity/max_arity_locals.compilers.reference
@@ -19,7 +19,7 @@
        x100 x101 x102 x103 x104 x105 x106 x107 x108 x109 x110 x111 x112 x113
        x114 x115 x116 x117 x118 x119 x120 x121 x122 x123 x124 x125 x126
        (function {nlocal = 1} x127 x128 x129 x130 x131
-         [(consts ()) (non_consts ([0: *, *]))](makelocalblock 0 (*,*) x1 x2)))
+         [(consts ()) (non_consts ([0: *, *]))](makelocalblock 0 x1 x2)))
    local_param_after_split =
      (function {nlocal = 0} x1 x2 x3 x4 x5 x6 x7 x8 x9 x10 x11 x12 x13 x14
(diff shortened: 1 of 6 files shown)


lthls (commented):
The changes in the testsuite are expected: a few blocks went from having shapes looking like `Non_nullable Pgenval` to `Nullable Pgenval`, and there is a piece of code in `Printlambda.block_shape` that avoids printing shapes where all fields are generic (nullable) values which now triggers.

The change in `Typeopt` looks correct (assuming that the mechanism to get the nullability is correct; I don't know enough about that part to review it).
However, I think it would make sense to split `value_kind` into `value_kind_non_null`, which would keep the current code but return values kinds without nullability (of type `Lambda.value_kind_non_null`), and `value_kind` which would be a small wrapper that computes nullability from the type and adds it to the raw kind returned by `value_kind_non_null`. It depends on having a reliable way to get nullability from an arbitrary type expression (and environment), but would simplify the code and in particular remove the concern about using type declarations.

goldfirere:
I don't think I agree about splitting the function. In many cases, we know the nullability within `value_kind`. (There are only three calls to `add_nullability_from_jkind`, in fact.) If we split the function, we will effectively forget this, and will have to recompute it, which can be expensive.

goldfirere:
I'm glad that you both agree that the new implementation is correct. Given that we're trying to roll out the fix today, could one of you approve? If need be, we can address the refactoring later.

dkalinichenko-js (approved):
Done. You might want to rebase to pick up the change moving `or_null` to beta.

lthls:
> I don't think I agree about splitting the function. In many cases, we know the nullability within `value_kind`. (There are only three calls to `add_nullability_from_jkind`, in fact.) If we split the function, we will effectively forget this, and will have to recompute it, which can be expensive.

I'm fine with your argument, but I'm not very confident that the current version, where nullability is hardcoded in each case, is future-proof enough. So I would welcome a way to at least check that the nullability we compute is consistent with what we would get by looking at the type's jkind independently (that could be turned on if/when weird bugs involving `or_null` start appearing again).

PR #3629, opened 2025-02-26
Define lattice ops in terms of conversion to an integer

glittershark:
From some digging around in godbolt, this generates *way* nicer code than the explicit match - plus it's (imo) clearer and easier to get right.

This does not, in the current state of the world, show up on a quick benchmark, but the generated assembly is so obviously better - and the code not obviously worse (and arguably nicer) that I somewhat think we should make this change anyway.

### DIFF
--- a/typing/mode.ml
+++ b/typing/mode.ml
@@ -51,7 +51,7 @@ module Lattices = struct
 
     let legacy = L.legacy
 
-    let le a b = L.le b a
+    let[@inline] le a b = L.le b a
 
     let join = L.meet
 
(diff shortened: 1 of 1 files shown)


typing/mode.ml
```diff
@@ -111,23 +138,17 @@ module Lattices = struct
...
         | Global -> Format.fprintf ppf "global"
         | Local -> Format.fprintf ppf "local"
-    end)
+
+      let[@inline] index = function Global -> 0 | Local -> 1
+
+      let[@inline] of_index_exn = function
+        | 0 -> Global
+        | 1 -> Local
+        | _ -> invalid_arg "Locality.of_index_exn"
```
goldfirere, line 150: Does this check for invalidity matter? It means that the conversion isn’t a no-op. Might be interesting to benchmark with and without this line.

goldfirere (commented):
I’m on the fence on this one. If it offered a demonstrated benefit, that would be great. But without that, I’m left to consider whether I like the new code more or less than the old. Pros: the new code doesn’t define easy-to-get-wrong comparison functions. Cons: the new code feels indirect, and has an easy-to-get-wrong conversion. I think the conversion functions are harder to get wrong than the comparison functions, but it’s still not a no-brainer for me.

Maybe a way to make this more palatable to me is to just have the `index` function and implement `join` and `meet` as `argmax` and `argmin`. That can probably be made to have as good code without needing the possibly-inconsistent and unnecessarily-branching back-conversion functions.

glittershark:
probably not worth it, tbh

PR #3632, opened 2025-02-27
Fix uniqueness error of variables bound internal to a module

riaqn:
Currently we treat the boundary between a module and its surrouding environment coarsely: references in both directions are always `aliased`.

To do that, we find all open variables in the module and mark them as `aliased`. We rely on `ienv` (containing only the surrouding variables, but not variables internal to the module) to decide if a variable is bound. If the variable is not found in `ienv`, then we think it's bound locally, and shouldn't be forced to `aliased`.

However, a subtle bug makes it not behaving properly. This PR fixes that.

### DIFF
--- a/testsuite/tests/typing-unique/unique_mod_class.ml
+++ b/testsuite/tests/typing-unique/unique_mod_class.ml
@@ -173,9 +173,4 @@ let () =
         end
     in ()
 [%%expect{|
-Line 6, characters 28-29:
-6 |           let _ = unique_id x in
-                                ^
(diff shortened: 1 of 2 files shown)


goldfirere (commented):
Clearly the code fixes the bug, but it would be greatly improved with some comments. What's `force_missing`? Why should it be `false` in the `Texp_ident` case? (Couldn't the ident be captured from outside the module?)

goldfirere (approved):

goldfirere:
Thanks for the helpful comments!

PR #3633, opened 2025-02-27
More `or_null` tests

dkalinichenko-js:
More `or_null` tests, similar to those in `typing-layouts-products/unboxed_tuples.ml`. I'm not sure how thoroughly I covered all the cases, so please point out what I missed.

### DIFF
--- a/testsuite/tests/typing-layouts-or-null/optimized.ml
+++ b/testsuite/tests/typing-layouts-or-null/optimized.ml
@@ -6,7 +6,7 @@
    native;
  } {
    flags = "-extension-universe beta -O3";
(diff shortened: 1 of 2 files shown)


lthls:
I think a test with a variant payload would be useful too (checking how `Is_null` interacts with `Is_int` and `Get_tag`). So at least having something like `int option or_null`, ideally a type with more constructors (`| A | B | C of int | D of int`).

dkalinichenko-js:
Done, with both `int option or_null` and a custom variant type.

lthls (approved):
Looks good to me

PR #3635, opened 2025-02-27
Make instruction identifiers abstract

xclerc:
As per title.

### DIFF
--- a/backend/arm64/cfg_selection.ml
+++ b/backend/arm64/cfg_selection.ml
@@ -180,5 +180,5 @@ class selector =
   end
 
 let fundecl ~future_funcnames f =
-  Cfg.reset_next_instr_id ();
+  Cfg.reset_instr_id ();
   (new selector)#emit_fundecl ~future_funcnames f
--- a/backend/fdo_info.mli
+++ b/backend/fdo_info.mli
@@ -33,12 +33,12 @@
 type info = private
   {
     dbg: Debuginfo.t;
-    discriminator: int;
+    discriminator: InstructionId.t;
   }
 type t = info option
 val none : t
 val is_none : t -> bool
-val create : dbg:Debuginfo.t -> discriminator:int -> t
+val create : dbg:Debuginfo.t -> discriminator:InstructionId.t -> t
 
 val equal_info : info -> info -> bool
 val equal : t -> t -> bool
--- a/backend/fdo_info.mli
+++ b/backend/fdo_info.mli
@@ -33,12 +33,12 @@
 type info = private
   {
     dbg: Debuginfo.t;
-    discriminator: InstructionId.t;
+    discriminator: int;
   }
 type t = info option
 val none : t
 val is_none : t -> bool
-val create : dbg:Debuginfo.t -> discriminator:InstructionId.t -> t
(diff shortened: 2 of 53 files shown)


backend/fdo_info.mli
```diff
@@ -33,12 +33,12 @@
 type info = private
   {
     dbg: Debuginfo.t;
-    discriminator: int;
+    discriminator: InstructionId.t;
```
gretay-js, line 36: The `discriminator` should remain as an int (fdo puts block label id there).

backend/arm64/cfg_selection.ml
```diff
@@ -180,5 +180,5 @@ class selector =
   end
 
 let fundecl ~future_funcnames f =
-  Cfg.reset_next_instr_id ();
+  Cfg.reset_instr_id ();
```
gretay-js, line 183: why do we need to reset here, instead of creating a fresh cfg?

gretay-js (approved):
Looks good, I only have two minor comments.

Later, it would be nice to find a better name for "generators" of type `InstructionId.sequence` to distinguish them from `InstructionId.t` fields. It can be done as part of a follow up to clean the remaining duplication around these generators. (Maybe each CFG can have its own generator to avoid exposing `last_used` and ordering requirements related to it.)

backend/fdo_info.mli
xclerc: So the discriminator is either a label or an instruction id?

backend/regalloc/regalloc_utils.ml
```diff
@@ -76,7 +76,7 @@ let make_log_function : verbose:bool -> label:string -> log_function =
   { log; enabled = verbose }
 
 module Instruction = struct
-  type id = int
+  type id = int (* XXX *)
```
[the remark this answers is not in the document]
xclerc: Sorry, forgot about that. Fixed.

backend/regalloc/regalloc_irc_state.ml
```diff
@@ -83,14 +83,15 @@ let[@inline] make ~initial ~stack_slots ~next_instruction_id () =
   let coalesced_nodes = RegWorkList.make ~original_capacity in
   let colored_nodes = Doubly_linked_list.make_empty () in
   let select_stack = [] in
-  let original_capacity = Int.min max_capacity (pred next_instruction_id) in
+  let original_capacity = 1024 (* XXX *) in
```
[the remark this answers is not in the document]
xclerc: Yes, sorry, forgot to revisit that line.

backend/arm64/cfg_selection.ml
xclerc: I did that to keep the pull request
obviously semantics-preserving,
but indeed I would like to eventually
revove that kind of state.

xclerc:
I agree; eventually each Cfg (or "with infos")
shoud hold its own generatoe, but as you
wrote it is probably better to do so in a
follow-up pull request.

backend/fdo_info.mli
gretay-js: yes

PR #3636, opened 2025-02-27
Some `value_or_null` support in Stdlib

dkalinichenko-js:
Annotate modules in `Stdlib` as working with `value_or_null`.

A notable omission are the `Hashtbl`, `Map` and `Set` modules. All of those can be made to work with `or_null`, but the changes to the key interface would be nontrivial.

Another omission is the `Array` module -- supporting `or_null` values in arrays requires separability. `Weak` hash tables also store their data in arrays.

`Atomic` is omitted pending minor changes to the middle end to propagate kind information correctly.

### DIFF
--- a/flambda-backend/tests/backend/zero_alloc_checker/test_custom_error_msg.opt.output
+++ b/flambda-backend/tests/backend/zero_alloc_checker/test_custom_error_msg.opt.output
@@ -22,7 +22,7 @@ Error: Annotation check for zero_alloc strict failed on function Test_custom_err
 This annotation is for testing custom error messages.
 
 File "test_custom_error_msg.ml", line 8, characters 22-39:
-Error: called function may allocate on a path to exceptional return (external call to caml_format_int) (test_custom_error_msg.ml:8,22--39;stdlib.ml:273,2--19)
+Error: called function may allocate on a path to exceptional return (external call to caml_format_int) (test_custom_error_msg.ml:8,22--39;stdlib.ml:275,2--19)
 
 File "test_custom_error_msg.ml", line 8, characters 13-40:
 Error: allocation of 24 bytes on a path to exceptional return
--- a/flambda-backend/tests/backend/zero_alloc_checker/test_custom_error_msg.output
+++ b/flambda-backend/tests/backend/zero_alloc_checker/test_custom_error_msg.output
@@ -22,7 +22,7 @@ Error: Annotation check for zero_alloc strict failed on function Test_custom_err
 This annotation is for testing custom error messages.
 
 File "test_custom_error_msg.ml", line 8, characters 22-39:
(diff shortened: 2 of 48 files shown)


goldfirere:
I'm glad to see this happening! But the actual PR makes me a little sad: it adds a lot of clutter, and will likely cause merge conflicts. And it's just the beginning of a lot of similar changes.

We had talked about having some kind of

```
[@@@default_kind: ('a : value_or_null)]
```

annotation that we could put in files that would make this a much smaller diff. Does it make sense to pause now and make this a reality? Or is that too slow and we should just blast forward?

goldfirere:
> I thought we hadn't reached much consensus about it last time it came up.

I agree with this. But I also don't remember consensus against. (Maybe I'm remembering selectively.) I see a feature like this to be quite like the `@@ portable` feature we've added, for much similar reasons.

(To be clear, I don't even have consensus with myself on the feature. Language features that change defaults can often have spooky action at a distance, and they make the semantics of code context-dependent in a way that means that a code snippet shared with a friend can subtly change meaning. But I continue to balk at the thought of changing ~every type signature everywhere to accommodate `or_null`.)

lpw25:
> But I continue to balk at the thought of changing ~every type signature everywhere to accommodate or_null

FWIW I think this is slightly the wrong thing to balk at. I'm pretty sure everywhere we would like to change now for `or_null` is a place we would like to change in future for layout polymorphism. I think that is worth bearing in mind when considering potential solutions. We should probably still balk at writing `('a : any). ...` everywhere, but an annotation that switches the default from `value` to `any` feels somewhat less ad-hoc.

It might also be possible to combine the default switch with marking values as being `static`. So maybe we have `sig static ... end` and it makes all the values `static` and thus changes the default for unannotated variables in those type schemes from `value` to `any`.

dkalinichenko-js:
While this is definitely a problem we need to solve for Base and Core, I think it's fine for Stdlib to be like that, and we don't need to resolve it to merge this.

dkalinichenko-js:
List of changed primitives to make sure they work with `value_or_null`:

Basics: `%identity`, `%obj_magic`, `%ignore`, `%revapply`, `%apply` -- clearly fine.
Exceptions: `%raise`, `%raise_notrace`, `%raise_with_backtrace` -- clearly fine.
Location: `%loc_LOC`, `%loc_LINE`, `%loc_POS` -- clearly fine.
Comparison: `%equal`, `%notequal`, ... -- changed to support `or_null` in https://github.com/ocaml-flambda/flambda-backend/pull/3265, tested in `runtime.ml`.
Block: `%makemutable`, `%field0`, ... -- tested in `runtime.ml` and `optimized.ml`.
Marshaling: `caml_output_value`, `caml_input_value`, ... -- changed to support `or_null` in https://github.com/ocaml-flambda/flambda-backend/pull/3265, tested in `runtime.ml`.

PR #3640, opened 2025-02-28
Mode crossing `mutable(m0)`

riaqn:
Based on #3639 

Currently `mutable` means `mutable(legacy)`. In general, `mutable(m0)` means new values for mutation will have mode `m0`. As a result, a record being constructed cannot have comonadic modes stronger than `m0`. When `m0 = legacy = nonportable`, that specifically means constructed record  cannot be `portable`.

This PR allows mode crossing `m0` depending on the field type and the modalities on it. Please see tests for examples.

### DIFF
--- a/typing/typecore.ml
+++ b/typing/typecore.ml
@@ -1012,14 +1012,14 @@ let mutable_mode m0 =
 
 (** Takes the mutability, the type and the modalities of a field, and expected
     mode of the record (adjusted for allocation), check that the construction
-    would be allowed. *)
-let check_construct_mutability ~loc ~env mutability ty ?modalities argument_mode =
+    would be allowed. This applies to mutable arrays similarly. *)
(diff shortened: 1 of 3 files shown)


typing/typecore.ml
```diff
@@ -5455,7 +5460,8 @@ and type_expect_
           None, expected_mode
       in
       let type_label_exp overwrite ((_, label, _) as x) =
-        check_construct_mutability ~loc ~env label.lbl_mut argument_mode;
+        check_construct_mutability ~loc ~env label.lbl_mut label.lbl_arg
+          ~modalities:label.lbl_modalities argument_mode;
```
lukemaurer, line 5464: As discussed, want to rename this `argument_mode` to `record_mode`

lukemaurer (approved):
LGTM with a couple of renames

PR #3641, opened 2025-02-28
Add fast-path to get_nullability

liam923:
When `-infer-with-bounds` is on, we currently spend a lot of time in `get_externality` and `get_nullability`, even though usually none of the with-bounds are relevant.

The branch https://github.com/goldfirere/flambda-backend/tree/rae/optimize-normalize fixes this for the `get_externality` case by adding a fast-path to `normalize` for when all relevant axes are max. But this doesn't work for `get_nullability` because most types are `Non_null`, which is not max.

One way to speed this up would be to add a fast-path to `normalize` (ie, avoid calling `loop`) for when none of the with-bounds are relevant. But this actually seems to slow things down. Empirically, the time overall spent in `normalize` ticked up slightly by adding this check.

Rather, this PR simply adds a special case to nullability where it avoids calling `normalize` when no with-bounds are relevant. When this patch is applied to https://github.com/goldfirere/flambda-backend/tree/rae/optimize-normalize (as of time of writing this), it drops the share of time spent in `get_nullability` when compiling `typecore.ml` from 5.8% to 0.8% (on main, it drops from 3.8% to 0.5%).

### DIFF
--- a/typing/jkind.ml
+++ b/typing/jkind.ml
@@ -568,9 +568,10 @@ module With_bounds = struct
     | No_with_bounds -> With_bounds_types.empty
     | With_bounds bounds -> bounds
 
-  let to_seq : type d. d with_bounds -> _ = function
-    | No_with_bounds -> Seq.empty
-    | With_bounds tys -> With_bounds_types.to_seq tys
+  let for_all (type l r) f (t : (l * r) t) =
+    match t with
+    | No_with_bounds -> true
+    | With_bounds tys -> With_bounds_types.for_all f tys
 
   let to_list : type d. d with_bounds -> _ = function
     | No_with_bounds -> []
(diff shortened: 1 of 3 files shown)


liam923:
> Can we ever have with-bounds relevant to nullability?

I think right now no, but I can easily imagine a future where we do. Specifically, if the implementation of `[@@unboxed]` changes. Also, I think when we add `kind_of` we will be able to.

typing/jkind.ml
```diff
@@ -2246,18 +2250,28 @@ let set_externality_upper_bound jk externality_upper_bound =
...
-      ~map_type_info:(fun _ ti ->
-        { relevant_axes =
-            (* Optimization: We only care about the nullability axis *)
-            Axis_set.intersection ti.relevant_axes only_nullability
-        })
-      jk.jkind
+  (* Optimization: Usually, no with-bounds are relevant to nullability. If we check for
+     this case, we can avoid calling normalize. *)
+  let all_with_bounds_are_irrelevant =
+    jk.jkind.with_bounds |> With_bounds.to_seq
```
goldfirere, line 2256: We can probably even avoid the `to_seq` here by defining a `With_bounds.for_all`. But this is also good as-is.

goldfirere (approved):

typing/jkind.ml
liam923: That shaved off enough to be worthwhile - down from 0.5% to 0.2% now.

liam923: Inspired by this comment, I also did a similar change for `map_type_expr`, which also shaved off `0.7%`.

goldfirere (approved):

PR #3642, opened 2025-03-03
Avoid polymorphic comparison in `backend/{cfg,peephole,regalloc}`

xclerc:
(As per title.)

### DIFF
--- a/backend/regalloc/regalloc_gi.ml
+++ b/backend/regalloc/regalloc_gi.ml
@@ -5,12 +5,6 @@ open! Regalloc_utils
 open! Regalloc_gi_utils
 module State = Regalloc_gi_state
 
-module List = struct
-  include List
-
-  let is_empty = function [] -> true | _ :: _ -> false
-end
-
 module Utils = struct
   include Regalloc_gi_utils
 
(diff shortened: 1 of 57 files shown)


backend/regalloc/regalloc_gi.ml
```diff
@@ -1,9 +1,16 @@
 [@@@ocaml.warning "+a-4-30-40-41-42"]
 
+open! Int_replace_polymorphic_compare
 open! Regalloc_utils
 open! Regalloc_gi_utils
 module State = Regalloc_gi_state
 
+module List = struct
```
gretay-js, line 8: can you please add a CR-someday here saying that we don't need it when we swtich to build with opam's ocaml 5.1?

backend/regalloc/regalloc_validate.ml
```diff
@@ -495,7 +496,8 @@ end = struct
            interference graph, which can lead to cycles. *)
         ()
       | _ ->
-        if instr.desc <> old_instr.desc
+        (* CR-soon xclerc for xclerc: avoid polymorphic equality. *)
```
gretay-js, line 499: why did you leave this for later?

backend/regalloc/regalloc_irc_state.ml
```diff
@@ -1,9 +1,16 @@
 [@@@ocaml.warning "+a-4-30-40-41-42"]
 
+open! Int_replace_polymorphic_compare
 open! Regalloc_utils
 open! Regalloc_irc_utils
 module Doubly_linked_list = Flambda_backend_utils.Doubly_linked_list
 
+module List = struct
```
gretay-js, line 8: should we add it back to `Misc.List`?

gretay-js (approved):

backend/regalloc/regalloc_irc_state.ml
xclerc: Done

backend/regalloc/regalloc_validate.ml
xclerc: Laziness, mainly, the comparison function
will be annoying to write.

backend/regalloc/regalloc_gi.ml
xclerc: Moved to `Misc` as suggested in
another comment. I reckon we
should have a recurring task to
remove no-longer needed functions
from `Misc`.

PR #3646, opened 2025-03-03
Avoid polymorphic comparison in `backend/{asm_targets,debug,internal_assembler}`

xclerc:
(As per title.)


mshinwell:
Might it be easier to use the build system to provide `-open Int_replace_polymorphic_compare`?

xclerc:
Yes, but I think it is annoying to do so on
a per-file/module basis, and not all are
ready.

PR #3648, opened 2025-03-03
Locals x Effects II

TheNumbat:
Rebased version of #2215 with the following additions:
- I reviewed the code
- Deletes local state fields from `caml_thread_struct` that became redundant
- Frees local arenas in `caml_free_stack` (previously leaked for every fiber). This also applies to domains and systhreads.
- All effects tests are enabled
- Deleted CR about comparing `stack->id`; there's nothing to compare it with
- Changed XXX about the initial fiber stack size to a CR

If you run out of memory by creating too many fibers, you can get `Fatal error: allocation failure during minor GC` when the minor GC tries to `malloc` a block. This seems unfortunate.


runtime/caml/memory.h
```diff
@@ -231,8 +231,18 @@ enum caml_alloc_small_flags {
 #define Alloc_small(result, wosize, tag, GC) \
   Alloc_small_with_reserved(result, wosize, tag, GC, (uintnat)0)
 
-CAMLextern caml_local_arenas* caml_get_local_arenas(caml_domain_state*);
-CAMLextern void caml_set_local_arenas(caml_domain_state*, caml_local_arenas* s);
+// Retrieve the local arenas for the given stack.
+// If the stack is the current stack, the copy of [local_sp] at the
+// root of [Caml_state] is saved in the current [stack_info]
+// structure, as it may have been updated by OCaml code.
+CAMLextern caml_local_arenas* caml_get_local_arenas_and_save_local_sp(struct stack_info*);
```
stedolan, line 238: I don't like the name of this function!

  - it's long and describes the implementation rather than what it's for
  - it sounds like the getter corresponding to `caml_set_local_arenas` but it's not (that one only sets the current stack's arenas, but this one reads the arenas of the given stack)

runtime/memory.c
```diff
@@ -526,40 +528,64 @@ CAMLexport void caml_modify_local (value obj, intnat i, value val)
...
-  caml_local_arenas* s = dom->local_arenas;
-  if (s != NULL)
-    s->saved_sp = dom->local_sp;
+  caml_local_arenas* s = stack->local_arenas;
+
+  // OCaml code may have updated [Caml_state->local_sp], so sync it to
+  // the [stack_info] structure, in the case where we are working with
+  // the current stack.
+
+  if (stack == Caml_state->current_stack && s != NULL) {
```
stedolan, line 540: I don't think the `s != NULL` check is needed any more? (The `current_stack->local_sp` field always exists and is always initialised, even if you haven't used any locals)

otherlibs/systhreads/st_stubs.c
```diff
@@ -294,8 +292,8 @@ static void save_runtime_state(void)
   th->gc_regs_buckets = Caml_state->gc_regs_buckets;
   th->exn_handler = Caml_state->exn_handler;
   th->async_exn_handler = Caml_state->async_exn_handler;
+  caml_get_local_arenas_and_save_local_sp(th->current_stack);
```
stedolan, line 295: I think this is being executed purely for the side-effect of setting:
```
th->current_stack->local_sp = Caml_state->local_sp;
```
and if so, it would probably clearer to write that directly.

testsuite/tests/backtrace/backtrace.byte.reference
```diff
@@ -0,0 +1,26 @@
...
+Raised at Backtrace.f in file "backtrace.ml", line 11, characters 16-32
+Called from Backtrace.f in file "backtrace.ml", line 11, characters 42-53
+Called from Backtrace.f in file "backtrace.ml", line 11, characters 42-53
+Called from Backtrace.f in file "backtrace.ml", line 11, characters 42-53
+Called from Backtrace.f in file "backtrace.ml", line 11, characters 42-53
+Called from Backtrace.f in file "backtrace.ml", line 11, characters 42-53
+Called from Backtrace.g in file "backtrace.ml", line 15, characters 4-11
+Called from Backtrace in file "backtrace.ml", line 21, characters 9-25
+Fatal error: exception Invalid_argument("index out of bounds")
+Raised at Backtrace in file "backtrace.ml", line 21, characters 12-24
```
stedolan, line 26: `backtrace.byte.reference` and `backtrace.reference` appear to be identical, so I don't think we need separate files here.

testsuite/tests/backtrace/backtrace_effects.byte.reference
```diff
@@ -0,0 +1,12 @@
...
+Called from Backtrace_effects.foo in file "backtrace_effects.ml", line 27, characters 12-17
+Called from Backtrace_effects in file "backtrace_effects.ml", line 43, characters 8-14
+(** get_continuation_callstack **)
+Raised by primitive operation at Backtrace_effects.bar in file "backtrace_effects.ml", line 22, characters 4-13
+Called from Backtrace_effects.foo in file "backtrace_effects.ml", line 27, characters 12-17
+(** raise **)
+Fatal error: exception Stdlib.Exit
+Raised at Backtrace_effects.bar in file "backtrace_effects.ml", line 17, characters 4-14
+Re-raised at Backtrace_effects.baz.(fun) in file "backtrace_effects.ml", line 33, characters 21-28
+Called from Backtrace_effects in file "backtrace_effects.ml", line 43, characters 8-14
```
stedolan, line 12: Is this file used? It doesn't seem referenced from `backtrace_effects.ml`

runtime/amd64.S
```diff
@@ -1341,6 +1363,9 @@ CFI_STARTPROC
         movq    Caml_state(exn_handler), %r10
         movq    %rsp, Stack_sp(%rcx)
         movq    %r10, Stack_exception(%rcx)
+    /* Save local allocations state that the OCaml code might have modified */
+        movq    Caml_state(local_sp), %r10
+        movq    %r10, Stack_local_sp(%rcx)
```
stedolan, line 1368: Why is this necessary here?

stedolan:
This looks good, thanks! Will be good to finally get this patch merged.

runtime/amd64.S
TheNumbat: We're about to switch `Caml_state->current_stack` to the new fiber, so we need to preemptively update the current stack's `local_sp` (same as `SWITCH_OCAML_STACKS`)

testsuite/tests/backtrace/backtrace_effects.byte.reference
TheNumbat: Oops, edited the wrong test at some point. Fixed.

testsuite/tests/backtrace/backtrace.byte.reference
TheNumbat: Oops, fixed

runtime/memory.c
TheNumbat: Agreed

runtime/caml/memory.h
TheNumbat: How about `caml_refresh_locals`? Also changed `set` to `use`.

runtime/major_gc.c
```diff
@@ -1127,6 +1127,7 @@ Caml_noinline static intnat do_some_marking(struct mark_stack* stk,
     }
 
     value_ptr scan_end = me.end;
+    CAMLassert(scan_end != NULL);
```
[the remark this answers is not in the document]
TheNumbat: I'm not sure either, but I agree it can never fire. Maybe left over from @mshinwell debugging something?

testsuite/tests/callback/stack_overflow.ml
```diff
@@ -1,7 +1,6 @@
 (* TEST
- modules = "stack_overflow_.c";
- reason = "CR ocaml 5 effects: re-enable this test";
- skip;
+ runtime5;
+ { modules = "nested_fiber_.c"; }
```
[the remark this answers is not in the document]
TheNumbat: Huh, ok. Apparently this is only the case when the test stanza has a predicate. It indeed was complaining the test didn't do anything without the extra `{ ... }`, so I thought that fixed it.

testsuite/tests/parallel/test_issue_11094.ml
```diff
@@ -55,7 +56,9 @@ let run =
   in
   let domains =
     Array.init num_domains (fun _ ->
-        Domain.spawn (fun () -> spawn (work 100000)))
+        (* By default, each fiber allocates an 8MB stack, so we can only
+           create ~1000 of them without running out of memory in CI. *)
+        Domain.spawn (fun () -> spawn (work 1000)))
```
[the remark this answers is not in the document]
TheNumbat: Changed

testsuite/tests/callback/stack_overflow.ml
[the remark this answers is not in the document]
mshinwell: There must be a mistake in the check in obscure cases, I will look at this.

runtime/major_gc.c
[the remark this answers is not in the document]
mshinwell: Yeah probably, let's just remove it.

runtime/amd64.S
```diff
@@ -252,6 +255,10 @@
     /* Fill in Caml_state->current_stack->sp */            \
         movq    Caml_state(current_stack), %r10;           \
         movq    %rsp, Stack_sp(%r10);                      \
+    /* No need to update Stack_local_sp: if the runtime    \
+       needs this value, it will copy it out of            \
+       Caml_state first (see                               \
+       caml_get_local_arenas_and_update_local_sp). */      \
```
mshinwell, line 261: This name needs updating

mshinwell: (also elsewhere)

testsuite/tests/typing-local/effects.ml
```diff
@@ -0,0 +1,215 @@
+(* TEST
+   runtime5;
+   arch_amd64;
```
mshinwell, line 3: Please leave a comment at the top of the arm64 emitter that it needs updating to support effects.  I'll probably do the updates themselves in due course.

runtime/caml/memory.h
mshinwell: Names seem fine now.

mshinwell (approved):

PR #3650, opened 2025-03-04
Remove dependency of trie.ml and leapfrog.ml on `Named_ref`.

Ekdohibs:
This cleans up the code and should make it possible to cleanly move these files in `algorithms/`.

### DIFF
--- a/middle_end/flambda2/datalog/cursor.ml
+++ b/middle_end/flambda2/datalog/cursor.ml
@@ -17,18 +17,24 @@ open Heterogenous_list
 
 type vm_action =
   | Unless :
-      ('t, 'k, 'v) Trie.is_trie * 't Named_ref.t * 'k Option_ref.hlist
+      ('t, 'k, 'v) Trie.is_trie
+      * 't ref
+      * 'k Option_ref.hlist
+      * string
+      * string list
       -> vm_action
 
 type action =
-  | Bind_iterator : 'a option Named_ref.t * 'a Trie.Iterator.t -> action
+  | Bind_iterator : 'a option Named_ref.t * 'a Trie.Iterator.with_name -> action
   | VM_action : vm_action -> action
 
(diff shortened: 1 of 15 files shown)


middle_end/flambda2/datalog/cursor.ml
```diff
@@ -14,21 +14,31 @@
...
+open With_name
 
 type vm_action =
   | Unless :
-      ('t, 'k, 'v) Trie.is_trie * 't Named_ref.t * 'k Option_ref.hlist
+      ('t, 'k, 'v) Trie.is_trie
+      * 't ref
+      * 'k Option_ref.hlist
+      * string
+      * string list
```
bclement-ocp, line 24: It would be useful to have a short comment mentioning that we don't use `with_name` here to avoid an extra indirection during execution (otherwise I could see someone making that change).

middle_end/flambda2/datalog/datalog.ml
```diff
@@ -92,33 +93,45 @@ let bind_iterator actions var iterator =
...
- fun ~order post_level args iterators ->
-  match args, iterators with
-  | [], [] -> ()
-  | this_arg :: other_args, this_iterator :: other_iterators -> (
+    type a.
+    order:_ -> _ -> a Term.hlist -> a Trie.Iterator.hlist -> string list -> unit
+    =
+ fun ~order post_level args iterators iterator_names ->
+  match args, iterators, iterator_names with
+  | [], [], _ -> ()
```
bclement-ocp, line 100: Let's be robust and do

```ocaml
| [], [], [] -> ()
| [], [], _ :: _ -> Misc.fatal_error "Too many names in [bind_atom]"
```

bclement-ocp (approved):
Looks good overall, thanks! Just two minor comments.

I am also wondering if we would want to make the `cell` field of `Parameter.t` a `'a option ref with_name` directly (and remove the `name` field entirely), but that's probably too nitpicky.

PR #3651, opened 2025-03-04
Experimental new GC pacing policy

stedolan:
Enable with Xgc_pacing_policy=1, and maybe tweak with Xpercent_sweep_per_mark=N


NickBarnes (approved):
LGTM. Needs a banner comment adjacent to the calculation, but I think that can wait until we have further refinements (the time/space trade-off control).

PR #3653, opened 2025-03-04
Specialize comparisons in presence of `immediate_or_null` correctly.

dkalinichenko-js:
Code handling comparisons in `translprim.ml` assumes that all immediate values are tagged integers. But, with `or_null`, this is no longer true. Check that a type is both `external` and `non_null` before specializing to `Compare_ints`.

I reviewed other uses of `maybe_pointer_type`, and they seem unproblematic to allow on null pointers.

Add a test. On `main`, the test fails to compile since Flambda recognizes that `Compare_ints` doesn't work on null pointers.

Tagging @glittershark for the front-end review, @lthls for the middle-end.

### DIFF
--- a/lambda/translprim.ml
+++ b/lambda/translprim.ml
@@ -1246,7 +1246,7 @@ let specialize_primitive env loc ty ~has_constant_constructor prim =
   in
   match prim, param_tys with
   | Primitive (Psetfield(n, Pointer, init), arity), [_; p2] -> begin
-      match maybe_pointer_type env p2 with
+      match fst (maybe_pointer_type env p2) with
       | Pointer -> None
       | Immediate -> Some (Primitive (Psetfield(n, Immediate, init), arity))
     end
@@ -1254,7 +1254,7 @@ let specialize_primitive env loc ty ~has_constant_constructor prim =
       (* try strength reduction based on the *result type* *)
       let is_int = match is_function_type env ty with
         | None -> Pointer
-        | Some (_p1, rhs) -> maybe_pointer_type env rhs in
+        | Some (_p1, rhs) -> fst (maybe_pointer_type env rhs) in
       Some (Primitive (Pfield (n, is_int, mut), arity))
   | Primitive (Parraylength t, arity), [p] -> begin
       let loc = to_location loc in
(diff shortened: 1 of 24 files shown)


typing/ctype.ml
```diff
@@ -2462,6 +2462,14 @@ let check_type_externality env ty ext =
   | Ok () -> true
   | Error _ -> false
 
+let check_type_nullability env ty nullability =
+  let upper_bound =
+    Jkind.set_nullability_upper_bound (Jkind.Builtin.any ~why:Dummy_jkind) nullability
+  in
+  match check_type_jkind env ty upper_bound with
```
glittershark, line 2469:
```suggestion
  (* CR layouts v2.8: We should have a better way to check the upper bound of a jkind along a particular relevant axis than this (also for [check_type_externality], above) *)
  match check_type_jkind env ty upper_bound with
```

glittershark (approved):
I'm sad about this, but no more sad than I already am about `check_type_externality`. changes in `typing` LGTM

lambda/translprim.ml
```diff
@@ -1432,7 +1432,9 @@ let specialize_primitive env loc ty ~has_constant_constructor prim =
       Some (Comparison(comp, Compare_ints))
     end else if (is_base_type env p1 Predef.path_int
         || is_base_type env p1 Predef.path_char
-        || (maybe_pointer_type env p1 = Immediate)) then begin
+        || ((* Non-null external types are always represented by tagged integers. *)
+            maybe_pointer_type env p1 = Immediate
```
lthls, line 1436: I think I would prefer if the return type of `maybe_pointer` was tweaked to carry nullability information. In my mind, `Immediate` implies `Non_null`, so if this is not the case I'd like to be more explicit about it

dkalinichenko-js: Good idea. This caught a case I missed with `[@untagged]` annotations on externals, fixed and added a test.

dkalinichenko-js:
Merging to include the fix in the compiler roll, @lthls please note if you had something else in mind by changing the return type of `maybe_pointer_type`. I also thought about changing `immediate_or_pointer` to include nullability information, but that introduces too much noise in places where we don't care about nullability, e.g. when working with `Pfield`/`Psetfield`.

PR #3656, opened 2025-03-05
Ensure root safety in caml_register_named_value.

NickBarnes:
Fix a bug identified by @stedolan in [this comment](https://github.com/ocaml-flambda/flambda-backend/pull/3603#issuecomment-2696730631) on #3603: the `caml_plat_lock_non_blocking` might conceal a GC and/or compaction which moves `vname` and/or `val`, and so invalidate `name`.

### DIFF
--- a/runtime/callback.c
+++ b/runtime/callback.c
@@ -447,16 +447,18 @@ static unsigned int hash_value_name(char const *name)
 
 CAMLprim value caml_register_named_value(value vname, value val)
 {
+  CAMLparam2(vname, val);
   const char * name = String_val(vname);
   size_t namelen = strlen(name);
   unsigned int h = hash_value_name(name);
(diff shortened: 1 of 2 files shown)


runtime/callback.c
```diff
@@ -447,6 +447,7 @@ static unsigned int hash_value_name(char const *name)
 
 CAMLprim value caml_register_named_value(value vname, value val)
 {
+  CAMLparam2(vname, val);
   const char * name = String_val(vname);
```
mshinwell, line 451: I suggest adding a comment about `name` as this seems like a footgun (unsafe to use after line 456).

NickBarnes: commented (and set to NULL for good measure).

mshinwell (approved):

PR #3657, opened 2025-03-05
Remove `-infer-with-bounds` flag, defaulting to `true`

glittershark:
It now looks like performance of with-bounds is satisfactory to release enabled
by default, so let's just remove this flag entirely to clean things up.


goldfirere (approved):

goldfirere (approved):
Some of the testsuite changes aren't great, but when we unconditionally print with-bounds, I think they'll improve. Still isn't worth holding this up.

PR #3659, opened 2025-03-05
Drop `jkind_of_first_type`

glittershark:
Per the CR in `product`, this can go away entirely now that we have proper
subsumption, and simplify things a bit.

### DIFF
--- a/testsuite/tests/typing-local/local.ml
+++ b/testsuite/tests/typing-local/local.ml
@@ -1228,6 +1228,8 @@ let foo y =
   gbl
 [%%expect{|
 val foo : 'a -> 'a = <fun>
+|}, Principal{|
+val foo : '_weak1 -> '_weak1 = <fun>
 |}]
 let foo (local_ gbl) =
   let _ = #{ gbl } in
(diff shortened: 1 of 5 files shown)


testsuite/tests/typing-local/local.ml
```diff
@@ -1228,6 +1228,8 @@ let foo y =
   gbl
 [%%expect{|
 val foo : 'a -> 'a = <fun>
+|}, Principal{|
+val foo : '_weak1 -> '_weak1 = <fun>
```
glittershark, line 1232: is this ok? what's going on here?

goldfirere: Don't know and don't care. We don't use non-principal mode within Jane Street. (Other than within the compiler.) But we should definitely leave a CR to come back here when we take a run at fixing principal-mode problems.

typing/typedecl.ml
```diff
@@ -1861,12 +1861,6 @@ let update_decl_jkind env id decl =
           in
           let type_jkind =
             Jkind.for_unboxed_record
```
goldfirere, line 1863: Could tighten up the spacing here

goldfirere (approved):

testsuite/tests/typing-local/local.ml
glittershark: sgtm

PR #3662, opened 2025-03-05
Remove as many magics and unsafe crossings as possible

glittershark:
Now that with-kinds has landed and inference is enabled by default, remove as many `[@@unsafe_allow_any_mode_crossing]` annotations and `Obj.magic`s which were annotated with "CR with-kinds" as possible

The correct mode crossing of the types in `Modes` is tested (extensively) in https://github.com/ocaml-flambda/flambda-backend/blob/aspsmith/fix-with-kinds-for-unboxed-records/testsuite/tests/typing-modal-kinds/unboxed.ml

### DIFF
--- a/stdlib/domain.ml
+++ b/stdlib/domain.ml
@@ -227,7 +227,6 @@ module Runtime_5 = struct
 
     let init () = create_dls ()
 
-    (* CR with-kinds: Remove [Key] wrapper. *)
     type 'a key = int * (Access.t -> 'a) Modes.Portable.t
 
     let key_counter = Atomic.Safe.make 0
--- a/stdlib/domain.ml
+++ b/stdlib/domain.ml
@@ -395,9 +395,7 @@ module Runtime_5 = struct
 
   let spawn' f =
     do_before_first_spawn ();
-    let pk = DLS.access (fun access -> DLS.get_initial_keys access
-      |> Obj.magic_portable (* CR with-kinds: Unnecessary magic. *))
-    in
+    let pk = DLS.access (fun access -> DLS.get_initial_keys access) in
 
     (* [term_sync] is used to synchronize with the joining domains *)
     let term_sync =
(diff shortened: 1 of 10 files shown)


tdelvecchio-jsc (changes requested):
Can you link (in the PR description) to tests that demonstrate these types all still have the desired crossing behavior?

glittershark:
> Can you link (in the PR description) to tests that demonstrate these types all still have the desired crossing behavior?

Done (they didn't, so thanks for the push to check that 😄)

stdlib/sys.mli
```diff
@@ -419,13 +417,12 @@ type extra_prefix = Plus | Tilde
 type extra_info = extra_prefix * string
```
glittershark, line 417: this needs to wait until https://github.com/ocaml-flambda/flambda-backend/pull/3664 also merges

stdlib/random.ml
```diff
@@ -375,9 +375,7 @@ let bits32 = apply0 State.bits32
 let bits64 = apply0 State.bits64
 let nativebits = apply0 State.nativebits
 
-let full_init seed =
-  (* CR with-kinds: Unnecessary magic. *)
-  let seed = Obj.magic_portable seed in
+let full_init (seed : int array @@ contended) =
```
tdelvecchio-jsc, line 378: This should probably actually explicitly require it `@ uncontended`; this magic is actually a bit sloppy and should be documented a bit more carefully...

tdelvecchio-jsc (approved):
Looks good, thanks for the tests!

glittershark:
@tdelvecchio-jsc mind taking another look?

stdlib/domain.ml
```diff
@@ -228,9 +228,7 @@ module Runtime_5 = struct
     let init () = create_dls ()
 
     (* CR with-kinds: Remove [Key] wrapper. *)
```
tdelvecchio-jsc, line 230: Delete cr

tdelvecchio-jsc (approved):
Delete the cr I commented on, then looks good.

PR #3663, opened 2025-03-05
Fix untypeast/pprintast bug (backport upstream 13845)

ccasin:
This tiny PR backports ocaml/ocaml#13845 to fix a bug in untypeast/pprintast.

It has already been reviewed upstream so low scrutiny is appropriate. The only question on our end is whether we need to do anything special about modes, and I think the answer is no because both `let x @ local = ...` and `let (x @ local) = ...` are valid syntax.

Request review from @tdelvecchio-jsc

### DIFF
--- a/typing/untypeast.ml
+++ b/typing/untypeast.ml
@@ -437,9 +437,16 @@ let case : type k . mapper -> k case -> _ = fun sub {c_lhs; c_guard; c_rhs} ->
 let value_binding sub vb =
   let loc = sub.location sub vb.vb_loc in
   let attrs = sub.attributes sub vb.vb_attributes in
-  Vb.mk ~loc ~attrs
-    (sub.pat sub vb.vb_pat)
-    (sub.expr sub vb.vb_expr)
+  let pat = sub.pat sub vb.vb_pat in
+  let pat, value_constraint, modes =
+    match pat.ppat_desc with
+    | Ppat_constraint (pat, Some ({ ptyp_desc = Ptyp_poly _; _ } as cty),
+                       modes) ->
+      let constr = Pvc_constraint {locally_abstract_univars = []; typ = cty } in
+      pat, Some constr, modes
+    | _ -> pat, None, []
+  in
+  Vb.mk ~loc ~attrs ?value_constraint ~modes pat (sub.expr sub vb.vb_expr)
 
 let comprehension sub comp =
   let iterator = function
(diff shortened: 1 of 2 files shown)


tdelvecchio-jsc (changes requested):
It seems worthwhile to test modes here? In particular, a weird thing is
```
let (f @ mode) = fun x -> x
```
is valid, but
```
let (f @ mode) : 'a. 'a -> 'a = fun x -> x
```
is not.

ccasin:
> It seems worthwhile to test modes here? In particular, a weird thing is
> 
> ```
> let (f @ mode) = fun x -> x
> ```
> 
> is valid, but
> 
> ```
> let (f @ mode) : 'a. 'a -> 'a = fun x -> x
> ```
> 
> is not.

@tdelvecchio-jsc Good question. It turns out that the `untypeast/pprintast` combo is just totally broken on value bindings involving modes. This PR neither fixes nor causes this.

I don't have time to debug it today - what I've done is to add a test demonstrating it's broken in the first commit. The second commit doesn't fix it, but at least we know it's not the cause. I propose merging this as is and fixing this another day.

typing/untypeast.ml
```diff
@@ -437,9 +437,16 @@ let case : type k . mapper -> k case -> _ = fun sub {c_lhs; c_guard; c_rhs} ->
...
+  let pat = sub.pat sub vb.vb_pat in
+  let pat, value_constraint, modes =
+    match pat.ppat_desc with
+    | Ppat_constraint (pat, Some ({ ptyp_desc = Ptyp_poly _; _ } as cty),
+                       modes) ->
+      let constr = Pvc_constraint {locally_abstract_univars = []; typ = cty } in
+      pat, Some constr, modes
+    | _ -> pat, None, []
+  in
+  Vb.mk ~loc ~attrs ?value_constraint ~modes pat (sub.expr sub vb.vb_expr)
```
tdelvecchio-jsc, line 449: This currently drops `vb.pvb_modes`. If you are fine with that since modes are broken anyways then that's fine, but wanted to point it out.

tdelvecchio-jsc (approved):
See comment, but otherwise fine with me.

typing/untypeast.ml
ccasin: Thanks.  I agree this is busted.  I'm going to merge this now in the interest of making incremental improvement and take a look to see if I can fix modes.

PR #3664, opened 2025-03-05
Infer with-bounds for tuples

glittershark:


glittershark:
still need to do performance testing on this, but unless things change drastically this should be roughly ~final

glittershark:
fwiw, the change in timing for typecore is within noise threshold.

glittershark:
full tree build seems roughly within noise threshold, so I'm gonna say this is open for review.

goldfirere (approved):

PR #3668, opened 2025-03-06
Shape UIDs for parameter modules

lukemaurer:
This will fix a Merlin bug where it's not possible to jump to interface when the
cursor is on a parameter module.


lpw25 (approved):
LGTM

PR #3671, opened 2025-03-07
Fix mode crossing inference for `[@@unboxed]` types

glittershark:
Kinds are inferred for `[@@unboxed]` types by first popping off as many layers
of `[@@unboxed]` data representations, then inferring the kind of the resulting
outermost non-`[@@unboxed]` type. This is the correct thing to do for _layouts_,
but importantly wrong for mod- and with-bounds, because those `[@@unboxed]`
types that we pop off might have had a modality for their (single) field, which
we need to respect by inferring the minimum for all axes for which that modality
is constant. This (I believe, if I'm reading the notation right) is what's
specified by the `TK_UNBOXED` typing rule in the [kind inference
doc](https://github.com/ocaml-flambda/flambda-backend/blob/main/jane/doc/proposals/kind-inference.md):

``` 
Γ ⊢ σ : κ
δ represents the type_kind
---------------------------------------------- TK_UNBOXED
Γ ⊢tk {{ private }} (( { ℓ : σ @@ modalities }
                     | K : σ @@ modalities -> τs t
                     | K : { ℓ : σ @@ modalities } -> τs t )) [@@unboxed] ↠
  δ : lay(κ); ⟪⊥_Ξ with types_for(Ξ, σ @@ modalities)⟫
```

This PR implements that, by returning the modality in addition to the innermost
`type_expr` from `get_unboxed_type_representation` and
`get_unboxed_type_approximation`, and then "applying" that modality to the
inferred jkind, by setting the mod bounds to min for all the axes along which
the modality is constant, and intersecting the relevant axes for each with-bound
with the relevant axes implied by the modality.

### DIFF
--- a/testsuite/tests/typing-jkind-bounds/basics.ml
+++ b/testsuite/tests/typing-jkind-bounds/basics.ml
@@ -1160,9 +1160,9 @@ type 'a t : value mod global portable contended many aliased unyielding =
 Lines 1-2, characters 0-66:
 1 | type 'a t : value mod global portable contended many aliased unyielding =
 2 |   { x : 'a @@ global portable contended many aliased } [@@unboxed]
-Error: The kind of type "t" is value
+Error: The kind of type "t" is value mod global aliased many contended portable
          because it instantiates an unannotated type parameter of t,
-         chosen to have kind value.
+         chosen to have kind value mod global aliased many contended portable.
        But the kind of type "t" must be a subkind of
          immutable_data mod global aliased
          because of the annotation on the declaration of the type t.
@@ -1202,16 +1202,8 @@ type ('a : value mod global) t = Foo of 'a [@@unboxed]
 type ('a : value mod contended many) t : value mod contended many aliased =
   { x : 'a @@ aliased } [@@unboxed]
 [%%expect {|
-Lines 1-2, characters 0-35:
-1 | type ('a : value mod contended many) t : value mod contended many aliased =
-2 |   { x : 'a @@ aliased } [@@unboxed]
-Error: The kind of type "t" is value mod many contended
-         because of the annotation on 'a in the declaration of the type t.
-       But the kind of type "t" must be a subkind of
-         value mod aliased many contended
-         because of the annotation on the declaration of the type t.
+type ('a : value mod many contended) t = { x : 'a @@ aliased; } [@@unboxed]
 |}]
-(* CR layouts v2.8: this should be accepted *)
 
 type ('a : value mod external_) t : immediate =
   Foo of 'a @@ global portable contended many aliased [@@unboxed]
--- a/testsuite/tests/typing-jkind-bounds/basics.ml
+++ b/testsuite/tests/typing-jkind-bounds/basics.ml
@@ -1167,7 +1167,19 @@ Error: The kind of type "t" is value mod global aliased many contended portable
          immutable_data mod global aliased
          because of the annotation on the declaration of the type t.
 |}]
-(* CR layouts v2.8: this should be accepted *)
+(* CR layouts v2.8: this could be accepted, if we infer ('a : value mod
+   unyielding). We do not currently do this, because we finish inference of the
+   type declaration before ever consulting the jkind annotation. *)
+(* CR layouts v2.8: In addition, the error message is a little sad, in that it
+   reports the jkind of t imprecisely. Really, its jkind should have "mod
+   unyielding with 'a @@ stuff" -- because if 'a mode-crossing yielding, then so
+   does 'a t (and this is true in practice). What's going on here is that the
+   algorithm in typedecl uses the jkind of 'a (which is value) as the jkind of
+   'a t (after taking modalities into account). This is misleading, though
+   understandable. In the end, though, this bug manifests only as a confusing
+   error message, not deeper misbehavior, and so is low priority. When we have
+   [layout_of], we'll be able to give a better jkind to [@@unboxed] types, and
+   this will likely improve. *)
 
 type 'a t : value mod global portable contended many aliased unyielding =
   Foo of 'a @@ global portable contended many aliased [@@unboxed]
--- a/testsuite/tests/typing-modal-kinds/unboxed.ml
+++ b/testsuite/tests/typing-modal-kinds/unboxed.ml
@@ -52,30 +52,35 @@ Error: This value is "nonportable" but expected to be "portable".
 |}]
 
 let foo (x : ((int -> int) portable_unboxed) @@ nonportable) = use_portable x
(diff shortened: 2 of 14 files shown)


typing/ctype.ml
```diff
@@ -2229,14 +2240,16 @@ let rec estimate_type_jkind ~(expand_component : type_expr -> open_type_expr) en
...
+            let ({ ty; is_open = is_open2 }, modality) = expand_component ty in
+            (is_open1 || is_open2), (ty, modality))
          false ltys
      in
      (* CR layouts v2.8: This pretty ridiculous use of [estimate_type_jkind]
         just to throw most of it away will go away once we get [layout_of]. *)
      let jkinds =
-       List.map (fun (ty, _) -> estimate_type_jkind ~expand_component env ty)
+       List.map (fun (ty,
+                      (* CR aspsmith: use this modality? *)
```
glittershark, line 2251: I wasn't able to find anything that failed due to us throwing this away. Plausibly this is just fine to ignore?

typing/ctype.ml
```diff
@@ -2440,7 +2455,11 @@ let constrain_type_jkind ~fixed env ty jkind =
              let recur ty's_jkinds jkinds =
                let results =
                  Misc.Stdlib.List.map3
-                   (loop ~fuel ~expanded:false ~is_open) tys ty's_jkinds jkinds
+                   (fun (ty,
+                         (* CR aspsmith: use this? *)
```
glittershark, line 2459: I wasn't able to find anything that failed due to us throwing this away. Plausibly this is just fine to ignore?

typing/ctype.ml
```diff
@@ -2484,7 +2503,7 @@ let constrain_type_jkind ~fixed env ty jkind =
                  Error
                    (Jkind.Violation.of_ ~jkind_of_type
                       (Not_a_subjkind (ty's_jkind, jkind, sub_failure_reasons)))
-               | Stepped { ty; is_open = is_open2 } ->
+               | Stepped ({ ty; is_open = is_open2 }, (* CR aspsmith: use this? *) _modality) ->
```
glittershark, line 2506: I wasn't able to find anything that failed due to us throwing this away. Plausibly this is just fine to ignore?

goldfirere:
The rule from the design doc isn't really implementable without `lay`, which we also call `layout_of`. So I wouldn't say you've given an interpretation for that rule.... but I agree with your alternative rule.

glittershark:
force pushed to rebase, no new changes

typing/ctype.ml
goldfirere: This one seems important. I think it will probably come up with tests with unboxed records.

typing/ctype.ml
goldfirere: This one also seems important. I think it will be triggered with something like

```
type ('a : value mod portable) require_portable
type t = { x : (int -> int) @@ portable } [@@unboxed]
type t2 = t require_portable
```

typing/jkind.ml
```diff
@@ -2258,6 +2265,24 @@ let set_nullability_upper_bound jk nullability_upper_bound =
 
 let set_layout jk layout = { jk with jkind = { jk.jkind with layout } }
 
+let apply_modality modality jk =
```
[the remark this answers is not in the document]
goldfirere: Actually I figured this out. It is correct.

testsuite/tests/typing-jkind-bounds/basics.ml
```diff
@@ -1160,9 +1160,9 @@ type 'a t : value mod global portable contended many aliased unyielding =
 Lines 1-2, characters 0-66:
 1 | type 'a t : value mod global portable contended many aliased unyielding =
 2 |   { x : 'a @@ global portable contended many aliased } [@@unboxed]
-Error: The kind of type "t" is value
+Error: The kind of type "t" is value mod global aliased many contended portable
```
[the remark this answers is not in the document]
goldfirere: I've figured out what's going on here and added some commentary. It's confusing (to the point of being misleading, I think), but not a sign of a deeper problem. Resolving.

typing/jkind.mli
```diff
@@ -565,10 +565,20 @@ val set_nullability_upper_bound :
...
-  Mode.Modality.Value.Const.t -> 'd Types.jkind -> 'd Types.jkind
+  Mode.Modality.Value.Const.t -> Types.jkind_l -> Types.jkind_l
+
+(** Change a jkind to be appropriate for an expectation of a type under
+    a modality. This means that the jkind's axes affected by the modality
+    will all be top. The with-bounds are left unchanged. *)
+val apply_modality_to_expected :
+  Mode.Modality.Value.Const.t ->
+  ('l * allowed) Types.jkind ->
+  ('l * allowed) Types.jkind
```
glittershark: idea: maybe `apply_modality_l` and `apply_modality_r`? or `_left` and `_right` if you want to be more verbose

glittershark: (the "expected" suffix threw me for a bit of a loop originally)

testsuite/tests/typing-modal-kinds/unboxed.ml
```diff
@@ -186,3 +186,80 @@ Line 1, characters 62-63:
...
+
+(* unboxed records *)
+
+type 'a t : value & value mod portable =
+  #{ x : 'a @@ portable; y : 'a @@ portable }
+[%%expect{|
+type 'a t = #{ x : 'a @@ portable; y : 'a @@ portable; }
+|}]
+
+let f (x : (int -> int) t @@ nonportable) = use_portable_vv x
```
glittershark, line 198: can you also do a check of subkinding (using `require_portable_vv`)?

testsuite/tests/typing-modal-kinds/unboxed.ml
```diff
@@ -186,3 +186,80 @@ Line 1, characters 62-63:
...
+       The kind of (int -> int) t is
+         immediate with int -> int @@ portable & immediate
+         with int -> int @@ portable
+         because of the definition of t at lines 1-2, characters 0-45.
+       But the kind of (int -> int) t must be a subkind of value & value
+         because of the definition of use_portable_vv at line 6, characters 64-75.
+
+       The first mode-crosses less than the second along:
+         nullability: mod non_null with int -> int ≰ mod non_null
+|}]
```
glittershark, line 217: maybe leave a CR for this?

glittershark:
@goldfirere's commits all LGTM

goldfirere (approved):
Ah, I forgot @glittershark can't approve, because she opened the PR. This covers both of us.

PR #3675, opened 2025-03-11
Restore the `-no-mach-ir` command-line switch

xclerc:
As per title; the switch was wrongly
removed in #3287. While it is no
longer useful, it should be kept as
a no-op for some time, to make the
transition easier.

### DIFF
--- a/driver/flambda_backend_args.ml
+++ b/driver/flambda_backend_args.ml
@@ -22,6 +22,9 @@ let mk_flambda2_debug f =
 let mk_no_flambda2_debug f =
   "-no-flambda2-debug", Arg.Unit f, " Disable debug output for the Flambda2 pass"
 
+let mk_no_mach_ir f =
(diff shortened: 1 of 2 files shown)


driver/flambda_backend_args.ml
```diff
@@ -22,6 +22,9 @@ let mk_flambda2_debug f =
 let mk_no_flambda2_debug f =
   "-no-flambda2-debug", Arg.Unit f, " Disable debug output for the Flambda2 pass"
 
+let mk_no_mach_ir f =
+  "-no-mach-ir", Arg.Unit f, " Avoid using the Mach IR"
```
ccasin, line 26: Maybe update the string to indicate this is a no-op kept for compatibility reasons?

ccasin (approved):

driver/flambda_backend_args.ml
xclerc: Good idea, thanks for the review.

PR #3681, opened 2025-03-11
Extend the Xgc_pacing_policy tweak to allow a "full 2025" option.

NickBarnes:
Lifted the actual pacing code out into a separate function, used a switch, and added a third option, here identical to the second, in preparation for @stedolan's square-root pacing extension.

### DIFF
--- a/runtime/major_gc.c
+++ b/runtime/major_gc.c
@@ -77,7 +77,13 @@ uintnat caml_percent_free = Percent_free_def;
 uintnat caml_max_percent_free = Max_percent_free_def;
 uintnat caml_percent_sweep_per_mark = 120; /* TODO: benchmark this value */
 
-uintnat caml_gc_pacing_policy = 0;
+/* Allowable values of caml_gc_pacing_policy */
+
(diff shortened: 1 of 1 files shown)


runtime/major_gc.c
```diff
@@ -77,7 +77,13 @@ uintnat caml_percent_free = Percent_free_def;
 uintnat caml_max_percent_free = Max_percent_free_def;
 uintnat caml_percent_sweep_per_mark = 120; /* TODO: benchmark this value */
 
-uintnat caml_gc_pacing_policy = 0;
+/* Allowable values of caml_gc_pacing_policy */
+
+#define GC_PACING_OCAML_53        0 /* Pacing as in OCaml 5.3 */
```
stedolan, line 82: Maybe say OCaml 5.3 + markdelay? The markdelay patch has quite a large effect on how pacing works (by changing the total work to do, rather than changing the work computation)

NickBarnes: I've updated the comment. I think the identifier is good enough.

mshinwell (approved):

PR #3682, opened 2025-03-12
Support kind annotations on existential type variables in pattern matches

ccasin:
If you have a type like this
```ocaml
type packed = T : ('a : float64) . 'a -> packed
```
there is currently no way to match on the constructor and give a name to the existential variable, because there's no way to write a kind annotation on the variable and without that annotation it is defaulted to value
```ocaml
let f p =
  match p with
  | T (type a) (_ : a) -> ()
```

```
Line 5, characters 20-21:
5 |   | T (type a) (_ : a) -> ()
                        ^
Error: This pattern matches values of type "a"
       but a pattern was expected which matches values of type "('a : float64)"
       The layout of a is value
         because it's an unannotated existential type variable.
       But the layout of a must be a sublayout of float64
         because of the definition of packed at line 1, characters 0-47.
```

This PR adds support for the syntax:
```ocaml
let f p =
  match p with
  | T (type (a : float64)) (_ : a) -> ()
```

Two commits - the first adds this test case with the error, the second adds support for the annotation and fixes the test. It's 95% boilerplate.

Review: @rtjoa

### DIFF
--- a/testsuite/tests/typing-jkind-bounds/annots.ml
+++ b/testsuite/tests/typing-jkind-bounds/annots.ml
@@ -1714,17 +1714,9 @@ type packed = T : ('a : float64) . 'a -> packed
 
 let f p =
   match p with
-  | T (type a) (_ : a) -> ()
+  | T (type (a : float64)) (_ : a) -> ()
 
 [%%expect{|
 type packed = T : ('a : float64). 'a -> packed
-Line 5, characters 20-21:
-5 |   | T (type a) (_ : a) -> ()
-                        ^
-Error: This pattern matches values of type "a"
-       but a pattern was expected which matches values of type "('a : float64)"
-       The layout of a is value
-         because it's an unannotated existential type variable.
-       But the layout of a must be a sublayout of float64
-         because of the definition of packed at line 1, characters 0-47.
+val f : packed -> unit = <fun>
 |}]
--- a/testsuite/tests/typing-jkind-bounds/annots.ml
+++ b/testsuite/tests/typing-jkind-bounds/annots.ml
@@ -1720,3 +1720,18 @@ let f p =
 type packed = T : ('a : float64). 'a -> packed
 val f : packed -> unit = <fun>
 |}]
+
+let bad p =
+  match p with
+  | T (type (a : float64)) (x : a) -> Some x
+[%%expect{|
+Line 3, characters 43-44:
+3 |   | T (type (a : float64)) (x : a) -> Some x
+                                               ^
+Error: This expression has type "a" but an expression was expected of type
+         "('a : value_or_null)"
+       The layout of a is float64
+         because of the annotation on the existential variable a.
+       But the layout of a must be a sublayout of value
+         because the type argument of option has layout value_or_null.
+|}]
--- a/testsuite/tests/typing-jkind-bounds/annots.ml
+++ b/testsuite/tests/typing-jkind-bounds/annots.ml
@@ -1707,3 +1707,24 @@ type t = int
 type t = int
 type t = nativeint#
 |}]
+
+(*********************************************************)
+(* Test 14: annotation on existential variable in unpack *)
+type packed = T : ('a : float64) . 'a -> packed
+
+let f p =
+  match p with
+  | T (type a) (_ : a) -> ()
+
+[%%expect{|
+type packed = T : ('a : float64). 'a -> packed
+Line 5, characters 20-21:
+5 |   | T (type a) (_ : a) -> ()
+                        ^
+Error: This pattern matches values of type "a"
+       but a pattern was expected which matches values of type "('a : float64)"
+       The layout of a is value
+         because it's an unannotated existential type variable.
+       But the layout of a must be a sublayout of float64
+         because of the definition of packed at line 1, characters 0-47.
+|}]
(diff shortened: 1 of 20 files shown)


testsuite/tests/typing-jkind-bounds/annots.ml
```diff
@@ -1707,3 +1707,16 @@ type t = int
...
 type t = nativeint#
 |}]
+
+(*********************************************************)
+(* Test 14: annotation on existential variable in unpack *)
+type packed = T : ('a : float64) . 'a -> packed
+
+let f p =
+  match p with
+  | T (type (a : float64)) (_ : a) -> ()
```
goldfirere, line 1717: In other contexts (e.g. after `fun`), we allow e.g. `fun (type a : float64)`, without nesting parentheses, when binding only one variable. This isn’t quite compositional (nesting the parentheses is more natural in some sense), but it’s easier on the eyes. Does this new feature support omitting the parens here? I think it would be nice if it did, for consistency. (I didn’t actually check — just zooming straight to this test and don’t see it tested.)

typing/printtyped.ml
```diff
@@ -349,7 +349,13 @@ and pattern : type k . _ -> _ -> k general_pattern -> unit = fun i ppf x ->
...
         (fun i ppf (vl,ct) ->
-          let names = List.map (fun {txt} -> "\""^Ident.name txt^"\"") vl in
+          let names =
+            List.map
+              (fun ({txt}, jk) ->
+                 asprintf "\"%s\", %a"
+                   (Ident.name txt) (option i jkind_annotation) jk)
+              vl
+          in
           line i ppf "[%s]\n" (String.concat "; " names);
```
rtjoa, line 359:
```suggestion
          line i ppf "vars%a\n"
            (fun ppf ->
              List.iter (fun ({txt}, jk) ->
                typevar_jkind ~print_quote:false ppf (Ident.name txt, jk)))
            vl;
```
Nit: `option` prints a newline, so the spacing gets weird here.

Before the suggested change, `-dtypedtree` of
```ocaml
type packed = T : 'a * 'b -> packed
let f (T (type a b) (_ : a * b)) = ()
```
contains
```
                Tpat_construct "T"
                [
                  pattern (../testu.ml[2,36+21]..../testu.ml[2,36+22])
                    Tpat_any
                  pattern (../testu.ml[2,36+21]..../testu.ml[2,36+22])
                    Tpat_any
                ]
                Some
                  ["a",                   None
; "b",                   None
]
                  core_type (../testu.ml[2,36+25]..../testu.ml[2,36+30])
```
With the suggested change, this snippet becomes:
```
                Tpat_construct "T"
                [
                  pattern (../testu.ml[2,36+21]..../testu.ml[2,36+22])
                    Tpat_any
                  pattern (../testu.ml[2,36+21]..../testu.ml[2,36+22])
                    Tpat_any
                ]
                Some
                  vars a b
                  core_type (../testu.ml[2,36+25]..../testu.ml[2,36+30])
```

testsuite/tests/typing-jkind-bounds/annots.ml
```diff
@@ -1707,3 +1707,16 @@ type t = int
...
+
+[%%expect{|
+type packed = T : ('a : float64). 'a -> packed
+val f : packed -> unit = <fun>
+|}]
```
rtjoa, line 1722:
```suggestion
|}]

let bad p =
  match p with
  | T (type (a : float64)) (x : a) -> Some x
[%%expect{|
Line 3, characters 43-44:
3 |   | T (type (a : float64)) (x : a) -> Some x
                                               ^
Error: This expression has type "a" but an expression was expected of type
         "('a : value_or_null)"
       The layout of a is float64
         because of the annotation on the existential variable a.
       But the layout of a must be a sublayout of value
         because the type argument of option has layout value_or_null.
|}]
```
(Testing the mismatch case + the new `annotation_context`.)

testsuite/tests/typing-jkind-bounds/annots.ml
rtjoa: I agree this would be nice to have.

For completeness of discussion, it's also worth thinking about the following, where we don't allow an annotation without parentheses around `(a : value)`.
```ocaml
let BAD : type a : value . a -> a = fun x -> x
```
But this one is different because we cannot write `(type a)`.

rtjoa (approved):

testsuite/tests/typing-jkind-bounds/annots.ml
rtjoa: Hah, I found a fourth case in a method overrides. Apparently this is "exotic syntax."
https://github.com/ocaml-flambda/flambda-backend/blob/08b57b3fc5e7021309c4cb8628528e8b5b50f60d/testsuite/tests/exotic-syntax/exotic.ml#L63-L68

typing/printpat.ml
```diff
@@ -99,7 +99,7 @@ let rec pretty_val : type k . _ -> k general_pattern -> _ = fun ppf v ->
       | (_, _, Some ([], _t)) ->
           fprintf ppf "@[<2>%s@ @[(%a : _)@]@]" name (pretty_vals ",") vs
       | (_, _, Some (vl, _t)) ->
-          let vars = List.map (fun x -> Ident.name x.txt) vl in
+          let vars = List.map (fun (x, _) -> Ident.name x.txt) vl in
```
rtjoa, line 102: I think it's fine to discard the annotation here because this file is just used to print exhaustive match counterexamples, which won't need them - is that right?

I tried putting an `assert false` in this match case, and also in the one above (with `Some ([], _t)`) and the testsuite still passed...

typing/printtyped.ml
ccasin: You're a saint for bothering to check this.

typing/printpat.ml
ccasin: I think it's the case that we'll never generate a counterexample with an annotation here, yes.

PR #3683, opened 2025-03-12
Eliminate some memory allocations in the register allocator

seliopoulos-js:
This pull request eliminates some memory allocations in the register allocator by introducing a `Cursor.t` type to the doubly-linked list implementation.

Previously, some util functions used the doubly-linked list's `cell` API to iterate. This API allocates when the intiial cell is created, regardless of whether the list is empty, and then further allocates each time a pointer in the list is traversed.

The cursor API on the other hand:
* Does not allocate if the list is empty
* Allocates exactly once upon creation from a non-empty list
* Does not allocate when advancing the cursor.

### DIFF
--- a/backend/regalloc/regalloc_ls_utils.ml
+++ b/backend/regalloc/regalloc_ls_utils.ml
@@ -95,24 +95,35 @@ module Range = struct
 
   let print ppf r = Format.fprintf ppf "[%d,%d]" r.begin_ r.end_
 
-  let rec overlap_cell : t DLL.cell option -> t DLL.cell option -> bool =
-   fun left right ->
-    match left, right with
-    | Some left_cell, Some right_cell ->
-      let left_value = DLL.value left_cell in
-      let right_value = DLL.value right_cell in
+  let rec overlap_cursor : t DLL.Cursor.t -> t DLL.Cursor.t -> bool =
+    fun left right ->
+      let left_value = DLL.Cursor.value left in
+      let right_value = DLL.Cursor.value right in
       if left_value.end_ >= right_value.begin_
          && right_value.end_ >= left_value.begin_
       then true
       else if left_value.end_ < right_value.end_
-      then overlap_cell (DLL.next left_cell) right
+      then (
+        match DLL.Cursor.next left with
+        | Error `End_of_list -> false
+        | Ok () -> overlap_cursor left right)
       else if left_value.end_ > right_value.end_
-      then overlap_cell left (DLL.next right_cell)
-      else overlap_cell (DLL.next left_cell) (DLL.next right_cell)
-    | None, _ | _, None -> false
+      then (
+        match DLL.Cursor.next right with
+        | Error `End_of_list -> false
+        | Ok () -> overlap_cursor left right)
+      else (
+        match DLL.Cursor.next left, DLL.Cursor.next right with
+        | Error `End_of_list, _ | _, Error `End_of_list -> false
+        | Ok (), Ok ()  -> overlap_cursor left right)
+  ;;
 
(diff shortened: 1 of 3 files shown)


utils/doubly_linked_list.mli
```diff
@@ -86,3 +86,14 @@ val to_list : 'a t -> 'a list
...
+module Cursor : sig
+  type 'a t
+
+  val value : 'a t -> 'a
+
+  val next : 'a t -> (unit, [`End_of_list]) result
+  val delete_and_next : 'a t -> (unit, [`End_of_list]) result
+end
+
+val create_hd_cursor : 'a t -> ('a Cursor.t, [`Empty]) result
```
xclerc, line 99: (I don't think we use that error-handling
style that much in the compiler, and since
there is only one possible error in all
case it would probably be ok to use a
bare option.)

utils/doubly_linked_list.ml
```diff
@@ -375,3 +375,51 @@ let map t ~f =
...
+    match t.node with
+    | Empty ->
+      (* internal invariant: cell's nodes are not empty *)
+      assert false
+    | Node content ->
+      (match content.next with
+      | Empty -> Error `End_of_list
+      | Node _ ->
+        t.node <- content.next;
+        Ok ())
```
xclerc, line 418: Should it be `next t`?

backend/regalloc/regalloc_ls_utils.ml
```diff
@@ -95,55 +95,67 @@ module Range = struct
...
-      if left_value.end_ >= right_value.begin_
-         && right_value.end_ >= left_value.begin_
-      then true
-      else if left_value.end_ < right_value.end_
-      then overlap_cell (DLL.next left_cell) right
-      else if left_value.end_ > right_value.end_
-      then overlap_cell left (DLL.next right_cell)
-      else overlap_cell (DLL.next left_cell) (DLL.next right_cell)
-    | None, _ | _, None -> false
+    let left_value = DLL.Cursor.value left in
```
xclerc, line 100: The code is slightly longer with that version.
Do you think we could keep the same code
structure by having `DLL.Cursor.next` return
a `(_ DLL.Cursor.t, ...) result` value? (It would
return the same value, just enabling "chaining".)

xclerc (commented):
Apologies for the delayed review.
I think the code is correct and could
be merged, but posted a couple of
(ignorable) comments.

backend/regalloc/regalloc_ls_utils.ml
seliopoulos-js: This will unfortunately allocate in the straightforward implementation. One could do something like:

```
type nonrec t = 
  { outer : 'a outer
  ; mutable node : 'a node
  ; ok_result = lazy t
  }

let create_hd_cursor t =
  match t.first with
  | Empty -> Error `Empty
  | Node _ ->
    let rec l = lazy (Ok { outer = t; node = t.first; ok_result = l })
    force l
```

and use `t.ok_result` as the return value in that case. This does allocate more but at least the allocation is constant.

utils/doubly_linked_list.ml
seliopoulos-js: Fixed.

backend/regalloc/regalloc_ls_utils.ml
xclerc: I was assuming `Ok ()`  (in `DLL.Cursor.next`)
would allocate, but you are implying it will
actually be lifted, right?

If that's the reason why my version would
be worse, I think we can forget what I said.

xclerc (approved):

PR #3684, opened 2025-03-12
Remove unnecessary existential tracking

goldfirere:
This removes some code that was pointless. (I wrote it after fixing other bugs nearby... but this bit wasn't necessary.)

But we might not want to merge, as a discussion on Slack suggests a new use for the infrastructure being removed here.


goldfirere:
Closing in favor of #3686

PR #3685, opened 2025-03-12
Accept shared modality

goldfirere:
As title.

### DIFF
--- a/testsuite/tests/typing-modal-kinds/basics.ml
+++ b/testsuite/tests/typing-modal-kinds/basics.ml
@@ -835,3 +835,81 @@ let arrow_left () =
 [%%expect{|
 val arrow_left : unit -> unit -> unit = <fun>
 |}]
+
+(*******************)
+(* shared modality *)
+
+type t = { x : int ref @@ shared }
+
+[%%expect{|
+>> Fatal error: Don't yet know how to interpret non-constant, non-identity modalities, but got join_with(shared) along axis contention.
+
+If you see this error, please contant the Jane Street compiler team.
+Uncaught exception: Misc.Fatal_error
+
+|}]
+
+type t : value mod contended = { x : int ref @@ shared }
+
+[%%expect{|
+>> Fatal error: Don't yet know how to interpret non-constant, non-identity modalities, but got join_with(shared) along axis contention.
+
+If you see this error, please contant the Jane Street compiler team.
+Uncaught exception: Misc.Fatal_error
+
+|}]
+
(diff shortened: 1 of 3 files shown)


testsuite/tests/typing-modal-kinds/basics.ml
```diff
@@ -835,3 +835,114 @@ let arrow_left () =
...
+Line 1, characters 10-15:
+1 | type t2 = int t require_contended
+              ^^^^^
+Error: This type "int t" should be an instance of type
+         "('a : value mod contended)"
+       The kind of int t is immutable_data with int
+         because of the definition of t at line 1, characters 0-32.
+       But the kind of int t must be a subkind of value mod contended
+         because of the definition of require_contended at line 1, characters 0-49.
+|}]
```
glittershark, line 899: this seems wrong; maybe we should leave a CR?

typing/jkind.ml
```diff
@@ -396,21 +396,11 @@ let raise ~loc err = raise (Error.User_error (loc, err))
...
-        | true, _ -> false
-        | _, true -> true
-        | false, false ->
-          Misc.fatal_errorf
-            "Don't yet know how to interpret non-constant, non-identity \
-             modalities, but got %a along axis %a.\n\n\
-             If you see this error, please contant the Jane Street compiler \
-             team."
-            Mode.Modality.print modality Mode.Value.print_axis axis)
+        not is_constant
```
glittershark, line 403: nit: I think we can probably get away with just skipping the extra `is_constant` variable, tbh

typing/jkind.ml
```diff
@@ -396,21 +396,11 @@ let raise ~loc err = raise (Error.User_error (loc, err))
 let relevant_axes_of_modality ~relevant_for_nullability ~modality =
   Axis_set.create ~f:(fun ~axis:(Pack axis) ->
       match axis with
-      | Modal axis -> (
+      | Modal axis ->
         let (P axis) = Mode.Const.Axis.alloc_as_value (P axis) in
         let modality = Mode.Modality.Value.Const.proj axis modality in
         let is_constant = Mode.Modality.is_constant modality in
```
glittershark, line 402: this is a place where we're diverging (at least a little bit) from `kind-inference.md`. Can we either:

- update that to document that it's ok to treat non-constant modalities the same as identity modalities for the purposes of mode crossing in kinds
- add a comment here to the same effect?

goldfirere: I've added a comment. Resolve if you're happy with it

PR #3686, opened 2025-03-12
Track existentials in constrain_type_jkind

goldfirere:
The first commit shows the bad case, the second commit adds some helpful debugging support, and the third commit fixes the problem.

A few notes:
* The extra line in for debugging must grossly be all on one line, because of the way the debug_printers.ml file is fed into the debugger via a [very very dumb `awk` invocation](https://github.com/ocaml-flambda/flambda-backend/blob/70a438e2518ecbbcf807c5db6f528ec4a88ff9c0/tools/dune#L137). The alternative is to define this function in the compiler, but dependencies within the Btype module make this slightly annoying. Writing one very long line seemed the best option, unless you (for any value of "you") want to improve that `awk` invocation.
* Interestingly, the fix required two mitigations: 1) track existentials in `constrain_type_jkind` (this is pretty straightforward, in the end), and 2) round up with-bounds in `update_decl_jkind`. Step 2 surprised me, but is clearly the right thing to do. Before this patch, we had `type_jkind`s in the wild with unbound variables in them. Horrors.
* Sadly, the `apply` step in `unbox_once` copies the unboxed type's field, which means that the variable changes identity. I worked around this by including the existentials within the same copy scope, but this makes me very sad. Either the copy mechanism should be taught not to copy existentials at generic-level (which seems hard and probably wrong), or maybe some cleverer interaction with the `copy_scope` could simplify this. Staring at this a bit showed no obvious way to get what I want, but I'm still a little mystefied by e.g. `Tsubst`. I'm confident that what I did is correct, but not confident that it's as good as it could be (from both a coding clarity and performance standpoint). Very happy for pointers here, probably from @lpw25 or @stedolan.

Review: @ccasin 

This is stacked on #3685, which is stacked on #3671. This patch is actually separable from those others, but it will cause annoying merge conflicts with them, and so I decided to stack them, given that the others are burning-hot priority.


lpw25:
> Step 2 surprised me, but is clearly the right thing to do. Before this patch, we had type_jkinds in the wild with unbound variables in them.

In the long run, I think it would be better if we could inject kinds back into the space of types used for with-kinds. That would give more precise with-kinds here. Alternatively, with-kinds could bind type variables, but that seems worse to me.

> Sadly, the apply step in unbox_once copies the unboxed type's field, which means that the variable changes identity. I worked around this by including the existentials within the same copy scope, but this makes me very sad.

I think that probably the right solution here is to leave apply as it is and use:
```ocaml
let apply ty2 existentials =
   apply env (decl.type_params @ existentials) ty2 (args @ existentials)
in
```
in `unbox_once`. Treating `ty` as a type expression with both `decl.type_params` and `existentials` as parameters, which it essentially is, and then filling in the `existentials` parameters with themselves.

goldfirere:
> I think that probably the right solution here is to leave apply as it is and use:

Yes that seems simpler. Thanks for taking a look. I will try.

typing/ctype.mli
```diff
@@ -579,12 +579,23 @@ val package_subtype :
 (* Raises [Incompatible] *)
 val mcomp : Env.t -> type_expr -> type_expr -> unit
 
+(* A set of bound variables *)
+module Bound_vars : sig
+  type t
+
+  val is_empty : t -> bool
+  val mem : t -> type_expr -> bool
```
liam923, line 587: @goldfirere are these exposed for debugging reasons? They don't seem to be used anywhere.

typing/ctype.ml
```diff
@@ -2220,13 +2220,54 @@ let expand_head_opt env ty =
...
+
+  val empty : t
+  val of_list : type_expr list -> t
+  val union : t -> t -> t
+  val mem : t -> type_expr -> bool
+  val is_empty : t -> bool
+end = struct
+  (* It is rare to care about which variables have been bound. This
+     implementation prioritizes the common case of caring only about the
+     presence of variables. *)
```
liam923, line 2234: I'm unconvinced that this is actually a perf improvement. In the common case, the sets are empty, so the set operations would be fast anyways and the laziness would just add overhead (I think it would add an allocation for the lazy?). I'm going to kick off some benchmarks to check.

liam923: Ok I've done some measuring (I built core 5 times with each version), and I'm unable to observe a performance difference the two approaches. So I propose we scrap the laziness to take the simpler approach.

liam923: (I also don't see any substantial perf changes compared to 5.2.0minus-23)

liam923:
I no longer think this is the way to go about this. I'm closing in favor of #4995

PR #3687, opened 2025-03-13
Unboxed records inside boxed records

xclerc:
This pull request is a follow-up to #3581,
and introduces a new construct in Lambda
shapes: a product, which recursively embed
a shape.

This pull request handles such products by
flattening them, and then applying the
reordering already present in #3581, so
that the prefix contains only value while
the suffix contains only unboxed elements.

This pull request also changes the index
(a single integer) of `Pmixedfield` and
`Psetmixedfield` to a path (a list of integers).

Like the previous pull request, this is joint
work with @ccasin and @mshinwell.


xclerc:
Here is a small example giving an
idea of the impact this pull request
can have:

```
type sub = #{a:int32#; b: string; c: float#}

type sub' = #{ d: char; e: int64#; f: int }

type record = {
  x : sub;
  y : sub';
}

let f r =
  r.x

let g r =
  let y = r.y in
  y.#f

external box_int32 : int32# -> (int32[@local_opt]) = "%box_int32"
external box_float : float# -> (float[@local_opt]) = "%box_float"

let () =
  let v = Sys.opaque_identity {
    x = #{ a = #12l; b = "abc"; c = #3.14; };
    y = #{ d = 'p'; e = #34L; f = 56; };
  } in
  let sub = f v in
  let z = g v in
  Printf.printf "%ld %S %g -- %d\n%!"
    (box_int32 sub.#a)
    sub.#b
    (box_float sub.#c)
    z

```

with `f` and `g` compiled to (linear):

```
camlTest__f_0_2_code: {test.ml:10,6-15}
  r:V/61[%rax] := R:I/0[%rax]
  F/62[%xmm0] := float64 [r:V/61[%rax] + 32] {test.ml:11,2-5}
  V/63[%rbx] := val [r:V/61[%rax]] {test.ml:11,2-5}
  I/64[%rax] := signed int32 [r:V/61[%rax] + 24] {test.ml:11,2-5}
  R:I/0[%rax] := I/64[%rax]
  R:I/1[%rbx] := V/63[%rbx]
  R:F/13[%xmm0] := F/62[%xmm0]
  return R:I/0[%rax] R:I/1[%rbx] R:F/13[%xmm0]

camlTest__g_1_3_code: {test.ml:13,6-33}
  r:V/61[%rax] := R:I/0[%rax]
  V/62[%rax] := val [r:V/61[%rax] + 16] {test.ml:14,10-13}
  R:I/0[%rax] := V/62[%rax]
  return R:I/0[%rax]
```

mshinwell:
I've been through a pass of review on `Mixed_block_shape` with @xclerc

rtjoa (commented):
Reviewed typing changes and relevant tests.

mshinwell (approved):

PR #3688, opened 2025-03-13
Remove Cassign and Cmut_let

mshinwell:
This means that all variables in `Cmm` are immutable.  WIP.


mshinwell:
@lthls You might be best placed to look at this, given the delayed handlers work.

mshinwell:
(There is a separate commit to improve Cfg printing that will be split out in due course.)

backend/cfg_selectgen.ml
```diff
@@ -742,43 +753,75 @@ class virtual selector_generic =
...
+      let env_body =
+        env_add_regs_for_exception_extra_args exn_cont extra_arg_regs env_body
+      in
       let r1, s1 = self#emit_sequence env_body e1 ~bound_name in
-      let rv = self#regs_for typ_val in
+      let rv_list =
+        List.map
+          (fun machtype -> self#regs_for machtype)
+          (typ_val :: List.map snd extra_args)
+      in
```
lthls, line 780: I think that we could use `let rv_list = (self#regs_for typ_val) :: extra_args_regs_split`, and remove the extra moves at the beginning of the handler (only keeping the move from `Proc.loc_exn_bucket` to `List.hd rv_list`.
What this PR currently does is reasonable, creating one set of registers for the calling convention of the handler and one set for the arguments of the handler, but since they are both local to the current function I think it's going to be simpler to just use the same set for both.

backend/cfg/cfgize_utils.ml
```diff
@@ -188,9 +188,8 @@ module Stack_offset_and_exn = struct
...
   let update_cfg : Cfg.t -> unit =
-   fun cfg ->
-    update_block cfg cfg.entry_label ~stack_offset:0 ~traps:[];
-    Cfg.iter_blocks cfg ~f:(fun _ block ->
-        if block.stack_offset = invalid_stack_offset then block.dead <- true;
-        assert (not (block.is_trap_handler && block.dead)))
+   fun cfg -> update_block cfg cfg.entry_label ~stack_offset:0 ~traps:[]
+  (*; Cfg.iter_blocks cfg ~f:(fun _ block -> if block.stack_offset =
+    invalid_stack_offset then block.dead <- true; assert (not
+    (block.is_trap_handler && block.dead))) *)
```
lthls, line 194: It would be nice to know why this code was removed

exn.ml
```diff
@@ -0,0 +1,16 @@
+let [@inline never] print_endline x = print_endline x
+
+let g () = failwith "foo"
+
+let f () =
+  let r = ref "foo" in
+  let s = ref "bar" in
+  try
+    (g [@inlined never]) ()
```
lthls, line 9: Maybe you could try to modify the references before raising, to check that the correct value gets picked up ?

backend/cfg/cfgize_utils.ml
mshinwell: Ah whoops, this was for debugging

exn.ml
mshinwell: I'll turn this into a proper test case too, and add one involving async exns as discussed.

middle_end/flambda2/to_cmm/to_cmm_expr.ml
```diff
@@ -428,49 +428,64 @@ let translate_apply0 ~dbg_with_inlined:dbg env res apply =
...
     in
-    let call, env, res, free_vars =
-      List.fold_left2 aux (call, env, res, free_vars) extra_args mut_vars
+    let extra_args = List.rev extra_args_rev in
+    let exn_var = Backend_var.create_local "*exn*" in
+    let handler_cont = Lambda.next_raise_count () in
+    let push_cont = Lambda.next_raise_count () in
+    let body =
+      C.create_ccatch ~rec_flag:false
+        ~body:(C.cexit push_cont [] [Push handler_cont])
```
lthls, line 470: I think there's a bug here: this `Push` doesn't have a corresponding `Pop`. I think you will have to add another wrapper continuation, for the normal return continuation, that pops the trap properly.

flambda-backend/testsuite/tests/asmgen/quicksort.cmm
```diff
@@ -2,8 +2,12 @@
  readonly_files = "main.c asan_report_wrappers.c";
  arguments = "-DSORT -DFUN=quicksort main.c asan_report_wrappers.c";
  asmgen;
+ skip;
+ reason="Needs rewriting to not use Cassign";
```
lthls, line 6: You have to put `skip` before any meaningful action if you want it to work properly (and you probably want to set the reason before skipping too, although that is less important).

mshinwell: Should be fixed

PR #3689, opened 2025-03-14
Remove basic meet

bclement-ocp:
This patch removes the old/basic meet and makes the new/advanced meet the only and default meet algorithm.

The configuration options (`-flambda2-basic-meet`, `-flambda2-advanced-meet`, and the `flambda2-meet-algorithm` OCAMLPARAM) are kept to avoid breaking scripts, but are now silently ignored.

Note that the `Meet_and_join_new` module is *NOT* renamed to `Meet_and_join`, and the `Meet_and_join` module simply forwards to `Meet_and_join_new`. This is intentional in order to re-use the same dispatch mechanism for the new join algorithm in #3538.

This PR also takes the opportunity to clean up the interfaces:

 - Remove the now unused `meet_env_extension`;
 - Remove `join` from the `Flambda2_types` interface


lthls (approved):
Looks good to me.

bclement-ocp:
Closing in favor of #3726 and #3727

PR #3690, opened 2025-03-14
Don't forget about modalities for fields of arrow types

glittershark:


typing/jkind.ml
```diff
@@ -1712,7 +1712,13 @@ module Jkind_desc = struct
       (* Optimization: all arrow types have the same (with-bound-free) jkind, so
          we can just eagerly do a join on the mod-bounds here rather than having
          to add them to our with bounds only to be normalized away later. *)
-      { t with mod_bounds = Mod_bounds.join t.mod_bounds Mod_bounds.for_arrow }
+      { t with
+        mod_bounds =
+          Mod_bounds.join t.mod_bounds
+            (Mod_bounds.set_min_in_set Mod_bounds.for_arrow
```
goldfirere, line 1718: Use `apply_modality_l` here? (I know that function is further down in this file, but I'd rather not repeat this subtle logic.)

goldfirere (approved):

typing/jkind.ml
glittershark: well, that applies to an entire jkind, whereas here we're only modifying the mod_bounds

glittershark: gonna merge for now since you approved, but feel free to push back and I'll make a follow-up PR.

PR #3691, opened 2025-03-14
Improve the [Atomic] interface to allow non-portable atomics and atomics of uncontended data.

tdelvecchio-jsc:
We were over-eager in making [Atomic.t] mode-cross, when the desired mode-crossing behavior is achievable with a combination of with-kinds and modalities. The new interface is simple, allows the existing upstream functions to be safe, and is more expressive. In particular, it allows us to generalize over portability more easily (e.g. in ppx_template).

### DIFF
--- a/stdlib/atomic.ml
+++ b/stdlib/atomic.ml
@@ -14,27 +14,27 @@
 
 type !'a t : mutable_data with 'a
 
-external make : 'a -> 'a t @@ portable = "%makemutable"
-external make_contended : 'a -> 'a t @@ portable = "caml_atomic_make_contended"
-external get : 'a t -> 'a @@ portable = "%atomic_load"
-external set : 'a t -> 'a -> unit @@ portable = "%atomic_set"
-external exchange : 'a t -> 'a -> 'a @@ portable = "%atomic_exchange"
-external compare_and_set : 'a t -> 'a -> 'a -> bool @@ portable = "%atomic_cas"
-external compare_exchange : 'a t -> 'a -> 'a -> 'a @@ portable = "%atomic_compare_exchange"
-external fetch_and_add : int t @ contended -> int -> int @@ portable = "%atomic_fetch_add"
-external add : int t @ contended -> int -> unit @@ portable = "%atomic_add"
-external sub : int t @ contended -> int -> unit @@ portable = "%atomic_sub"
-external logand : int t @ contended -> int -> unit @@ portable = "%atomic_land"
-external logor : int t @ contended -> int -> unit @@ portable = "%atomic_lor"
-external logxor : int t @ contended -> int -> unit @@ portable = "%atomic_lxor"
+external make : 'a -> ('a t[@local_opt]) @@ portable = "%makemutable"
+external make_contended : 'a -> ('a t[@local_opt]) @@ portable = "caml_atomic_make_contended"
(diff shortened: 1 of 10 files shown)


stdlib/atomic.ml
```diff
@@ -12,36 +12,29 @@
...
-
-external fetch_and_add : int t -> int -> int @@ portable = "%atomic_fetch_add"
-external add : int t -> int -> unit @@ portable = "%atomic_add"
-external sub : int t -> int -> unit @@ portable = "%atomic_sub"
-external logand : int t -> int -> unit @@ portable = "%atomic_land"
-external logor : int t -> int -> unit @@ portable = "%atomic_lor"
-external logxor : int t -> int -> unit @@ portable = "%atomic_lxor"
+external make : 'a -> 'a t @@ portable = "%makemutable"
+external make_contended : 'a -> 'a t @@ portable = "caml_atomic_make_contended"
+external get : 'a t -> 'a @@ portable = "%atomic_load"
```
polytypic, line 19: While we are at it, could we also make it so that the operations take the atomic as local?

IOW, `s/'a t ->/'a t @ local ->/g`.

tdelvecchio-jsc: Do we want `make_local` while we're at it? Or alternatively, `external make : 'a -> ('a t [@local_opt]) = "%makemutable"`?

stdlib/gc.ml
```diff
@@ -113,21 +113,21 @@ external finalise_last : (unit -> unit) -> 'a -> unit =
 external finalise_release : unit -> unit @@ portable = "caml_final_release"
 
 
-type alarm = bool Atomic.t
+type alarm = bool Atomic.t Modes.Immutable.t
```
TheNumbat, line 116: `Immutable` doesn't seem like a good name. Maybe `Immutable_data` would be better because it has the same (weird) semantics as the kind. But can you just use portended instead?

stdlib/printexc.ml
```diff
@@ -22,7 +22,9 @@ open Printf
 
 type t = exn = ..
 
-let printers = Atomic.Safe.make []
+type printers : immutable_data = (exn -> string option) Modes.Immutable.t list
```
TheNumbat, line 25: Here the argument is immutable but you're using the modality to enforce the _data part of Immutable_data.

stdlib/gc.ml
tdelvecchio-jsc: Got rid of immutable entirely.

stdlib/printexc.ml
```diff
@@ -22,7 +22,7 @@ open Printf
 
 type t = exn = ..
 
-type printers : immutable_data = (exn -> string option) Modes.Immutable.t list
+type printers = (exn -> string option) Modes.Portended.t list
```
TheNumbat, line 25: This could just use Portable

TheNumbat (approved):

stdlib/atomic.ml
TheNumbat: Make local_opt seems preferable

stdlib/printexc.ml
tdelvecchio-jsc: Ah, thought that wouldn't work cause of a with-kinds bug, but it does.

PR #3693, opened 2025-03-17
datalog: Clear binders after rule evaluation

bclement-ocp:
Clear out references after evaluating a Datalog rule or query to avoid keeping tables alive longer than needed.


bclement-ocp:
Actually this probably should be done inside `Fun.protect`, updating the PR.

Ekdohibs (approved):

PR #3694, opened 2025-03-17
Annotate `Unix` and `Thread` as (mostly) portable

polytypic:
This PR annotates the `Unix` and `Thread` modules (under `otherlibs`) as (mostly) portable.

### DIFF
--- a/otherlibs/systhreads/thread.mli
+++ b/otherlibs/systhreads/thread.mli
@@ -13,14 +13,35 @@
 (*                                                                        *)
 (**************************************************************************)
 
+@@ portable
+
 (** Lightweight threads for Posix [1003.1c] and Win32. *)
 
-type t
+type t : value mod contended portable
 (** The type of thread handles. *)
 
 (** {1 Thread creation and termination} *)
 
-val create : ('a -> 'b) -> 'a -> t
+module Portable : sig
+  val create : ('a -> 'b) @ portable -> 'a -> t
+  (** [Thread.Portable.create funct arg] creates a new thread of control,
+     in which the function application [funct arg]
+     is executed concurrently with the other threads of the domain.
+     The application of [Thread.create]
+     returns the handle of the newly created thread.
+     The new thread terminates when the application [funct arg]
+     returns, either normally or by raising the {!Thread.Exit} exception
+     or by raising any other uncaught exception.
+     In the last case, the uncaught exception is printed on standard error,
+     but not propagated back to the parent thread. Similarly, the
+     result of the application [funct arg] is discarded and not
+     directly accessible to the parent thread.
+
+     See also {!Domain.spawn} if you want parallel execution instead.
+     *)
+end
+
+val create : ('a -> 'b) -> 'a -> t @@ nonportable
 (** [Thread.create funct arg] creates a new thread of control,
    in which the function application [funct arg]
    is executed concurrently with the other threads of the domain.
@@ -166,7 +187,7 @@ val default_uncaught_exception_handler : exn -> unit
 (** [Thread.default_uncaught_exception_handler] will print the thread's id,
     exception and backtrace (if available). *)
 
-val set_uncaught_exception_handler : (exn -> unit) -> unit
+val set_uncaught_exception_handler : (exn -> unit) @ portable -> unit
 (** [Thread.set_uncaught_exception_handler fn] registers [fn] as the handler
     for uncaught exceptions.
 
--- a/otherlibs/systhreads4/thread.mli
+++ b/otherlibs/systhreads4/thread.mli
@@ -14,14 +14,32 @@
 (*                                                                        *)
 (**************************************************************************)
 
+@@ portable
+
 (** Lightweight threads for Posix [1003.1c] and Win32. *)
 
-type t
+type t : value mod contended portable
 (** The type of thread handles. *)
 
 (** {1 Thread creation and termination} *)
 
-val create : ('a -> 'b) -> 'a -> t
+module Portable : sig
+  val create : ('a -> 'b) @ portable -> 'a -> t
+  (** [Thread.Portable.create funct arg] creates a new thread of control,
+     in which the function application [funct arg]
+     is executed concurrently with the other threads of the program.
+     The application of [Thread.create]
+     returns the handle of the newly created thread.
+     The new thread terminates when the application [funct arg]
+     returns, either normally or by raising the {!Thread.Exit} exception
+     or by raising any other uncaught exception.
+     In the last case, the uncaught exception is printed on standard error,
+     but not propagated back to the parent thread. Similarly, the
+     result of the application [funct arg] is discarded and not
+     directly accessible to the parent thread. *)
+end
+
+val create : ('a -> 'b) -> 'a -> t @@ nonportable
 (** [Thread.create funct arg] creates a new thread of control,
    in which the function application [funct arg]
    is executed concurrently with the other threads of the program.
@@ -165,7 +183,7 @@ val default_uncaught_exception_handler : exn -> unit
 (** [Thread.default_uncaught_exception_handler] will print the thread's id,
     exception and backtrace (if available). *)
 
-val set_uncaught_exception_handler : (exn -> unit) -> unit
+val set_uncaught_exception_handler : (exn -> unit) @ portable -> unit
 (** [Thread.set_uncaught_exception_handler fn] registers [fn] as the handler
     for uncaught exceptions.
 
--- a/otherlibs/unix/unix.mli
+++ b/otherlibs/unix/unix.mli
@@ -28,6 +28,8 @@
    manual/src/library/libunix.etex
 *)
 
+@@ portable
+
 (** Interface to the Unix system.
 
    To use the labeled version of this module, add [module Unix][ = ][UnixLabels]
@@ -131,7 +133,7 @@ exception Unix_error of error * string * string
 val error_message : error -> string
 (** Return a string describing the given error code. *)
 
-val handle_unix_error : ('a -> 'b) -> 'a -> 'b
+val handle_unix_error : ('a -> 'b) -> 'a -> 'b @@ nonportable
 (** [handle_unix_error f x] applies [f] to [x] and returns the result.
    If the exception {!Unix_error} is raised, it prints a message
    describing the error and exits with code 2. *)
@@ -322,7 +324,7 @@ val nice : int -> int
 (** {1 Basic file input/output} *)
 
 
-type file_descr
+type file_descr : value mod contended portable
 (** The abstract type of file descriptors. *)
 
 val stdin : file_descr
@@ -1411,7 +1413,7 @@ val getgrgid : int -> group_entry
 (** {1 Internet addresses} *)
 
 
-type inet_addr
+type inet_addr : value mod contended portable
 (** The abstract type of Internet addresses. *)
 
 val inet_addr_of_string : string -> inet_addr
@@ -1683,7 +1685,7 @@ val shutdown_connection : in_channel -> unit
    connection is over. *)
 
 val establish_server :
-  (in_channel -> out_channel -> unit) -> sockaddr -> unit
+  (in_channel -> out_channel -> unit) -> sockaddr -> unit @@ nonportable
 (** Establish a server on the given address.
    The function given as first argument is called for each connection
    with two buffered channels connected to the client. A new process
--- a/otherlibs/unix/unix_unix.ml
+++ b/otherlibs/unix/unix_unix.ml
@@ -96,7 +96,7 @@ exception Unix_error of error * string * string
 let _ = Callback.Safe.register_exception "Unix.Unix_error"
                                     (Unix_error(E2BIG, "", ""))
 
-external error_message : error -> string = "caml_unix_error_message"
+external error_message : error -> string @@ portable = "caml_unix_error_message"
 
 let () =
   Printexc.Safe.register_printer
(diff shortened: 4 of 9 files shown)


otherlibs/systhreads/thread.mli
```diff
@@ -166,7 +168,7 @@ val default_uncaught_exception_handler : exn -> unit
 (** [Thread.default_uncaught_exception_handler] will print the thread's id,
     exception and backtrace (if available). *)
 
-val set_uncaught_exception_handler : (exn -> unit) -> unit
+val set_uncaught_exception_handler : (exn -> unit) @ portable -> unit
```
polytypic, line 171: [According to sherlocode](https://sherlocode.com/?q=set_uncaught_exception_handler) this seems rarely used.

otherlibs/unix/unix_unix.ml
```diff
@@ -962,13 +965,13 @@ end = struct
     | Some mutex -> Mutex.protect mutex f
 end
 
-let popen_mutex = lazy (Mutex.create ())
+let popen_mutex = Mutex.create ()
```
polytypic, line 968: I removed the `Lazy` wrapper from here.  A mutex should not be very expensive to allocate.

otherlibs/unix/unix_unix.ml
```diff
@@ -962,13 +965,13 @@ end = struct
...
 
 let open_proc prog args envopt proc input output error =
   let pid =
     create_process_gen prog args envopt input output error in
-  Mutex.protect (Lazy.force popen_mutex) (fun () ->
-    Hashtbl.add popen_processes proc pid)
+  Mutex.protect popen_mutex (fun () ->
+    Hashtbl.add (magic_uncontended popen_processes) proc pid)
```
polytypic, line 974: The `popen_mutex` is used to protect accesses to `popen_processes`.  Instead of using magic here, another alternative would perhaps be to use the capsule API.  I guess that might complicate things here as the capsule API is currently in `stdlib_alpha` &mdash; perhaps we should revisit this once the capsule API is in `stdlib`.

otherlibs/unix/unix_unix.ml
mshinwell: The lazy wrapper is needed because of a restriction under our runtime4 that systhreads must be initialized prior to using `Mutex`.

polytypic: Hmm… Good point, but I think in this case a real mutex is only created on runtime5. Does that make a difference?

mshinwell: The code as-is requires the lazy to be there, I don't think it will work correctly otherwise.

polytypic: See #3698

otherlibs/systhreads/thread.mli
polytypic: We found only a couple of uses of this function and one of them was incorrect.  So, going with the backwards incompatible change seems appropriate here.

otherlibs/unix/unix_unix.ml
polytypic: See #3698

otherlibs/unix/unix_unix.ml
```diff
@@ -395,7 +395,7 @@ external map_internal:
    file_descr -> ('a, 'b) Stdlib.Bigarray.kind
```
tdelvecchio-jsc, line 395: I think there's a real concern about `map_internal` here. If we allow `file_descr` to cross contention, then two domains can have simultaneous access to the same `file_descr`, can both `map_internal` on it, and then thus have two `Bigarray`s to the same underlying memory, which is a data race. Arguably, `map_internal` is inherently data-racey in nature, and so we shouldn't even bother, but I think it would be nice if we can prevent accidentally sharing the same `file_descr` across domains and using them in this unsafe way.

Proposal: Make `file_descr` cross portability but *not* contention; make most of the functions in this interface accept `file_descr @ contended`, but have `map_internal` require `file_descr @ uncontended`.

otherlibs/unix/unix_unix.ml
```diff
@@ -937,18 +937,21 @@ type popen_process =
   | Process_out of out_channel
   | Process_full of in_channel * out_channel * in_channel
 
-let popen_processes = (Hashtbl.create 7 : (popen_process, int) Hashtbl.t)
+let popen_processes @ portable contended =
+  Obj.magic_portable (Hashtbl.create 7 : (popen_process, int) Hashtbl.t)
```
tdelvecchio-jsc, line 941: Add comment that this is protected by a mutex below.

otherlibs/unix/unix_unix.ml
```diff
@@ -208,21 +208,21 @@ type wait_flag =
...
-external waitpid : wait_flag list -> int -> int * process_status
+external fork : unit -> int @@ portable = "caml_unix_fork"
+external wait : unit -> int * process_status @@ portable = "caml_unix_wait"
+external waitpid : wait_flag list -> int -> int * process_status @@ portable
    = "caml_unix_waitpid"
-external _exit : int -> 'a = "caml_unix_exit"
-external getpid : unit -> int = "caml_unix_getpid"
-external getppid : unit -> int = "caml_unix_getppid"
-external nice : int -> int = "caml_unix_nice"
+external _exit : int -> 'a @@ portable = "caml_unix_exit"
```
tdelvecchio-jsc, line 222: I can't find the implementation of this; does it skip over the `at_exit` logic from `Stdlib`?

otherlibs/unix/unix.mli
```diff
@@ -1683,7 +1685,7 @@ val shutdown_connection : in_channel -> unit
    connection is over. *)
 
 val establish_server :
-  (in_channel -> out_channel -> unit) -> sockaddr -> unit
+  (in_channel -> out_channel -> unit) -> sockaddr -> unit @@ nonportable
```
tdelvecchio-jsc, line 1688: Why is this nonportable?

tdelvecchio-jsc (changes requested):
Left some comments. I definitely don't have a deep enough understanding of low-level unix stuff to catch if some of these are unsafe, so it might be worth having another person take a look. (In particular, I blindly assume a lot of the c code is written in a thread-safe way; not sure how carefully we want to check that.)

Do we not care about making windows build?

mshinwell:
Windows isn't supported at present.

otherlibs/unix/unix_unix.ml
polytypic: Yes, it should skip the `at_exit` logic.  That is how it is [documented](https://ocaml.org/manual/5.3/api/Unix.html#VAL_exit).

otherlibs/unix/unix_unix.ml
polytypic: On the other hand, multiple processes may map the same object and already cause various issues &mdash; which I suspect is likely far more common than having data race issues within a single process using a single file descriptor.  Furthermore, it is also possible to have multiple file descriptors refer to the same object within a single process.  I feel like this is not a problem we can solve with modes.  (In another context one could design an API for more safely sharing memory, but that is different thing entirely.)

otherlibs/unix/unix.mli
polytypic: It calls `exit`, i.e. the version that does `at_exit` processing.

otherlibs/unix/unix_unix.ml
polytypic: Added comment.

otherlibs/unix/unix_unix.ml
tdelvecchio-jsc: I think the main thing here is that we are introducing a new axis of parallelism, and it's one where it's much easier to neglect the sharing that's happening. Anyone using this function is surely already thinking about 1) that another process likely has its own file descriptor to this memory location, and 2) if the current process creates the same file descriptor twice, then there is aliasing happening. This, however, is 3) a place where someone might share a single reference across domains without considering the fact that this is, implicitly, just an array.

I'm not hard-set on this at all, but I do think there is some benefit to the (slight) extra mode safety, even if it's just communicating to the user "hey, don't forget to consider this" when doing multi-domain programs. If they're already doing something unsafe, they can just reach for `Obj.magic_uncontended` to assert that they are aware of the unsafety.

otherlibs/systhreads/thread.mli
```diff
@@ -13,14 +13,33 @@
...
 (** Lightweight threads for Posix [1003.1c] and Win32. *)
 
-type t
+type t : value mod contended portable
 (** The type of thread handles. *)
 
 (** {1 Thread creation and termination} *)
 
-val create : ('a -> 'b) -> 'a -> t
+val create_portable : ('a -> 'b) @ portable -> 'a -> t
```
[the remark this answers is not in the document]
polytypic: Like this?
```ocaml
module Portable : sig
  val create : ('a -> 'b) @ portable -> 'a -> t
end
```
:thinking:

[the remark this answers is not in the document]
tdelvecchio-jsc: I think so, yeah. Reach out to dkalinichenko if you think this feels wrong.

otherlibs/unix/unix_unix.ml
polytypic: What about returning the bigarray as contended?

I think that would better match the actual semantics.  File descriptors pretty much cross portability and contention, but the memory mapped object may or may not be contended (accessed by multiple threads or multiple processes).

otherlibs/systhreads/thread.mli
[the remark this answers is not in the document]
polytypic: I don't have a strong opinion on whether this is wrong or not, but the mention made me curious.  Has there been some similar cases where a submodule was ultimately deemed the right approach, @dkalinichenko-js?

otherlibs/unix/unix_unix.ml
tdelvecchio-jsc: Returning it as contended is sad because that means that people not working in multi-domain programs now have to interact with the `contended` mode where they didn't before. Maybe this is desirable, but it definitely is a more invasive change.

By the way, atomics will soon *not* cross portability and contention, and instead will have exactly the behavior I described: there will be functions which accept `contended` atomics.

polytypic: > people not working in multi-domain programs now have to interact with the contended mode where they didn't before

OTOH, the possibility of contention exists even without multiple domains &mdash; they have to deal with that possibility regardless.  (Aside from multiple processes accessing the shared memory in parallel, it is also possible to have multiple systhreads accessing the same shared memory.)

Also, with atomics, the referred object might be of an immutable type.  In this case it is not.

However, we could also have a separate portable version:

```ocaml
val map_file :
     file_descr
  -> ?pos:int64
  -> ('a, 'b) Bigarray.kind
  -> 'c Bigarray.layout
  -> bool
  -> int array
  -> ('a, 'b, 'c) Bigarray.Genarray.t @@ nonportable

module Portable : sig
  val map_file :
     file_descr
  -> ?pos:int64
  -> ('a, 'b) Bigarray.kind
  -> 'c Bigarray.layout
  -> bool
  -> int array
  -> ('a, 'b, 'c) Bigarray.Genarray.t @ contended
end
```

Perhaps we should just mark `map_file` as nonportable for now?

polytypic: I've been sleeping on this for a few days as I felt I exhausted the basic arguments I had in mind.

To add one more bit to the previous arguments, I spent a moment looking through uses of `map_file` with sherlocode, and a typical pattern to using `map_file` looks roughly like this:

```ocaml
let buffer =
  let fd = Unix.openfile path [Unix.O_RDONLY] 0 in
  let len = Unix.lseek fd 0 Unix.SEEK_END in
  let map = Unix.map_file fd Bigarray.int8_unsigned Bigarray.c_layout false [|len|] in
  Unix.close fd;
  map
in
(* ... *)
```

What happens above, is that a file is opened creating a `fd`, the mapping is created by calling `map_file`, and then the `fd` is closed (and discarded).

As I see it, a file descriptor is just not a good proxy for managing contention on the result of `map_file`.  The mapping (i.e. the bigarray) is independent of the file descriptor.  The underlying object can be mapped or mutated otherwise by multiple processes, or by a single process using multiple file descriptors.  Furthermore, perhaps the most common way to obtain a mapping does not keep the file descriptor around (a fd is opened, mapping is created, the fd is closed, mapping is returned).

In light of all this, I feel that we should just make `file_descr` cross contention, because it makes the API slightly more convenient and easier to use.  Higher level APIs can and should build safer APIs on top of `map_file` to support memory mapping, but the `Unix.map_file` API is just not designed to be safe (i.e. uncontended) at this level.

tdelvecchio-jsc: Sounds good, I'm happy sticking with your solution since it's sound and in practice seems like it'll be more ergonomic.

otherlibs/systhreads4/thread.mli
```diff
@@ -14,14 +14,30 @@
...
 (** Lightweight threads for Posix [1003.1c] and Win32. *)
 
-type t
+type t : value mod contended portable
 (** The type of thread handles. *)
 
 (** {1 Thread creation and termination} *)
 
-val create : ('a -> 'b) -> 'a -> t
+val create_portable : ('a -> 'b) @ portable -> 'a -> t
```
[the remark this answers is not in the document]
polytypic: Now wrapped with `Portable` module.

tdelvecchio-jsc (approved):
Great, thank you!

PR #3695, opened 2025-03-17
Allow (most) kind checks to be turned off

mshinwell:
We have inadvertently ended up in a situation where kind checks in Flambda 2 can prevent valid code from being compiled.  Before it was possible for GADT matches to refine to different layouts, kind errors in Flambda 2 were always compiler errors; however, that is no longer the case.  For example:
```
type _ t =
| A : (float# -> int) t
| B : (int -> int) t

let[@inline] f (type a) (x : a t) (g : a) =
  match x with
  | A -> g #3.14
  | B -> g 42

let n =
  f (Sys.opaque_identity B) (fun i -> succ i)
```
We are trying to find ways of reinstating the checks in situations where we know they will be safe, but for now this PR disables the majority of them, guarded by a flag (defaulting to the checks being absent, `-no-flambda2-kind-checks`).  Some of the kind checks are still in place as they are either pre-beta-reduction or at places where there still should never be any mismatch.

When the kind checks are disabled, code exhibiting kind mismatches (e.g. the `A` branch in the above example) is replaced by `Invalid`.

### DIFF
--- a/driver/flambda_backend_args.ml
+++ b/driver/flambda_backend_args.ml
@@ -583,6 +583,14 @@ let mk_flambda2_unicode f =
     \     Flambda 2 code"
 ;;
 
+let mk_flambda2_kind_checks f =
+  "-flambda2-kind-checks", Arg.Unit f, " Perform kind checks on Flambda 2\n\
+    \     code (may cause fatal errors with layout-poly GADT code)"
+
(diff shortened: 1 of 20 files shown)


lthls (approved):
The PR looks fine. I've left two comments about comments that I think are slightly wrong.

PR #3706, opened 2025-03-18
Arm64 SIMD DSL

gretay-js:
~(On top of #3703)~

- Adds `Arm64_ast` that mimics `X86_*` structure.
- Adds `Isimd` to `Arch.Specific` and `backend/arm64/simd*` files that mimic `amd64` structure. Supports a small subset of intructions needed for most compiler tests.

Emit is updated but there is currently no change to code genereation because the relevant types for Float32 and Vec128 are not yet enabled, so we should be able to merge this PR without breaking arm64 builds.


backend/arm64/simd.ml
```diff
@@ -18,4 +18,93 @@
...
+      | Subq_i64 ),
+      _ ) ->
+    false
+
+let class_of_operation op =
+  match op with
+  | Round_f32 _ | Round_f32_i64 | Min_scalar_f32 | Max_scalar_f32 | Fmin_f32
+  | Fmax_f32 | Zip1_f32 | Zip1q_f32 | Zip1q_f64 | Zip2q_f64 | Addq_i64
+  | Subq_i64 ->
+    Pure
```
xclerc, line 108: I see there is a "current" mode for
`Round_f32`; does that amount to
a co-effect?

backend/arm64/simd_selection.ml
```diff
@@ -11,14 +11,55 @@
...
+  | "caml_neon_float32_fmin" -> Some (Fmin_f32, args)
+  | "caml_neon_float32_fmax" -> Some (Fmax_f32, args)
+  | "caml_neon_float32x2_zip1" -> Some (Zip1_f32, args)
+  | "caml_simd_vec128_interleave_low_32" | "caml_neon_float32x4_zip1" ->
+    Some (Zip1q_f32, args)
+  | "caml_simd_vec128_interleave_low_64" | "caml_neon_float64x2_zip1" ->
+    Some (Zip1q_f64, args)
+  | "caml_simd_vec128_interleave_high_64" | "caml_neon_float64x2_zip2" ->
+    Some (Zip2q_f64, args)
+  | "caml_simd_int64x2_add" | "caml_neon_int64x2_add" -> Some (Addq_i64, args)
```
xclerc, line 56: (Out of curiosity, why do we have two
names for some instructions?)

backend/arm64/isa/arm64_ast.ml
```diff
@@ -0,0 +1,597 @@
...
+  end
+
+  module Extend = struct
+    module Kind = struct
+      type t =
+        | UXTB (* 000 *)
+        | UXTH (* 001 *)
+        | UXTW (* 010 *)
+        | UXTX
+        (* 011 *)
```
xclerc, line 308:
```suggestion
        | UXTX (* 011 *)
```

backend/arm64/isa/arm64_ast.ml
```diff
@@ -0,0 +1,597 @@
...
+    | Comment of string
+    | Global of string
+    | Protected of string
+    | Hidden of string
+    | Weak of string
+    | Long of constant
+    | NewLabel of string
+    | NewLine
+    | Quad of constant
+    | Section of string list * string option * string list * bool
```
xclerc, line 447: (I understand the other version may
be defined like that, but an inline
record would be nicer.)

backend/arm64/isa/arm64_ast.ml
```diff
@@ -0,0 +1,597 @@
...
+    | Word of constant
+    (* gas only (the masm emitter will fail on them) *)
+    | Cfi_adjust_cfa_offset of int
+    | Cfi_endproc
+    | Cfi_startproc
+    | Cfi_remember_state
+    | Cfi_restore_state
+    | Cfi_def_cfa_register of string
+    | Cfi_def_cfa_offset of int
+    | File of int * string (* (file_num, file_name) *)
```
xclerc, line 460: (I would probably remove the
comment, and use a nested
record.)

backend/arm64/isa/arm64_ast.ml
```diff
@@ -0,0 +1,597 @@
...
+    | Type (_, _)
+    | Direct_assignment (_, _)
+    | Align _ ->
+      Misc.fatal_error "Directive not implemented yet"
+
+  let print_program ppf lines = List.iter (print_line ppf) lines
+
+  let generate_asm oc lines =
+    let ppf = Format.formatter_of_out_channel oc in
+    print_program ppf lines
```
xclerc, line 499: I might be midly worried `Format` could
insert line break - is that a valid concern?

backend/arm64/isa/arm64_ast.ml
```diff
@@ -0,0 +1,597 @@
...
+
+    let ins name operands = append (ins name operands)
+
+    let ins_cond name cond operands = append (ins_cond name cond operands)
+
+    let generate_asm oc =
+      let asm_code = List.rev !asm_code in
+      reset_asm_code ();
+      Asm.generate_asm oc asm_code
+  end
```
xclerc, line 596: (Should we use a doubly-linked list,
to avoid the need for a `rev`. I am a
bit worried we could end up with
very long lists.)

backend/arm64/isa/arm64_ast.mli
```diff
@@ -0,0 +1,194 @@
+(* The same physical register has different names in the assembly encoding of
```
xclerc, line 1: (For consistency, maybe enable some
warnings at the top of the file.)

backend/arm64/isa/arm64_ast.mli
```diff
@@ -0,0 +1,194 @@
...
+end
+
+module Reg : sig
+  type t
+
+  val create : Reg_name.t -> int -> t
+end
+
+module Instruction_name : sig
+  (** mnemonic *)
```
xclerc, line 58: (Should it be moved to line 87?)

xclerc (commented):
This is really a detail, but the asymetry
between the x86 modules (in `backend/`)
and the arm modules (in `backend/arm64/isa`).
Is the plan to eventually move the x86
module to a similar `isa` folder?

backend/arm64/isa/arm64_ast.ml
gretay-js: hmm, this was supposed to be auto-formatted. I am moving the file to `backend` to address your other comment, it'll help with autoformatting too.

backend/arm64/isa/arm64_ast.ml
gretay-js: I won't address this and the next comment to make it easier to identify the duplication during the upcoming refactoring of this area.

backend/arm64/isa/arm64_ast.ml
gretay-js: I don't know but it doesn't look like it, I'm not emitting any break hints or boxes, and the assembler will complain. Would you `Printf` instead throuhgout this module? Using `Format` in this mordule because it has more helper functions, but it's easy to change. In any case `generate_asm` is not used at the moment, the entire `Asm.Acc` is here to get rid of `.mlp`.

backend/arm64/isa/arm64_ast.ml
gretay-js: I'm adding this task to fix later. We should do it for both targets, this is copied from somewhere in amd64 code.

backend/arm64/isa/arm64_ast.mli
gretay-js: done.

backend/arm64/isa/arm64_ast.mli
gretay-js: Fixed.

backend/arm64/simd.ml
gretay-js: I think it's fine to just assume it's never changed during the program's execution. We also assume somewhere that it's a particular rounding mode (nearest, I think). It doesn't have co-effect in amd64 backend either. As long as it's consistent between the two backends, it's fine for now.

backend/arm64/simd_reload.ml
```diff
@@ -16,6 +16,16 @@
...
 (* SIMD instruction reload for ARM64 *)
 
-open! Int_replace_polymorphic_compare [@@ocaml.warning "-66"]
+open! Int_replace_polymorphic_compare [@@warning "-66"]
 
-let reload_operation _ _ arg res = arg, res
+let reload_operation makereg op arg res =
+  let stackp r =
+    match r.Reg.loc with Stack _ -> true | Reg _ | Unknown -> false
+  in
```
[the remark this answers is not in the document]
gretay-js: fixed.

backend/arm64/simd_selection.ml
gretay-js: This is discussed on [PR3704](https://github.com/ocaml-flambda/flambda-backend/pull/3704#issue-2928757611)

gretay-js (commented):
> This is really a detail, but the asymetry
between the x86 modules (in backend/)
and the arm modules (in backend/arm64/isa).
Is the plan to eventually move the x86
module to a similar isa folder?

I moved the new file under `backend` for now. It seemed better to put arm64 related code under one folder, but it causes trouble to formatting at the moment, and as you point out inconsistent with x86_* files.

backend/arm64/isa/arm64_ast.ml
xclerc: Ah, yes, sorry, since there are no boxes
that should be fine indeed.

backend/arm64/isa/arm64_ast.ml
xclerc: Sure, but please add an item specifically
for this case - the generic one is not very
likely to be addressed soon.

backend/amd64/simd_reload.ml
```diff
@@ -21,7 +21,8 @@ open! Int_replace_polymorphic_compare [@@ocaml.warning "-66"]
 let reload_operation makereg (register_behavior : Simd_proc.register_behavior)
     arg res =
   let stackp r =
-    match r.Reg.loc with Stack _ -> true | Reg _ | Unknown -> false
+    Reg.is_stack r
+    (* match r.Reg.loc with Stack _ -> true | Reg _ | Unknown -> false *)
```
xclerc, line 25: Sorry, but why keep the comment?

backend/arm64/simd_reload.ml
```diff
@@ -16,6 +16,17 @@
...
 (* SIMD instruction reload for ARM64 *)
 
-open! Int_replace_polymorphic_compare [@@ocaml.warning "-66"]
+open! Int_replace_polymorphic_compare [@@warning "-66"]
 
-let reload_operation _ _ arg res = arg, res
+let reload_operation makereg op arg res =
+  let stackp r =
+    Reg.is_stack r
+    (* match r.Reg.loc with Stack _ -> true | Reg _ | Unknown -> false *)
```
xclerc, line 24: (Same as under `amd64`.)

xclerc (approved):

backend/arm64/isa/arm64_ast.ml
gretay-js: Added a CR too.

PR #3707, opened 2025-03-18
Arm64: Float32 support

gretay-js:
~On top of #3706, only the last 2 commits are new.~

Support for `Float32` on `arm64` in `Proc` and `Emit`. 

Thanks a lot to @matthewelse for providing most of the code in this PR (the bugs are all mine).

### DIFF
--- a/backend/arm64/proc.ml
+++ b/backend/arm64/proc.ml
@@ -67,8 +67,8 @@ let float32_reg_name =
 
 let num_register_classes = 2
 
-let register_class r =
-  match (r.typ : Cmm.machtype_component) with
+let register_class_of_machtype_component typ =
+  match (typ : Cmm.machtype_component) with
   | Val | Int | Addr  -> 0
   | Vec128 ->
     (* CR mslater: (SIMD) arm64 *)
@@ -77,6 +77,9 @@ let register_class r =
     (* CR mslater: (SIMD) arm64 *)
     fatal_error "arm64: got valx2 register"
   | Float | Float32 -> 1
+
+let register_class r =
+  register_class_of_machtype_component r.typ
 
 let num_stack_slot_classes = 2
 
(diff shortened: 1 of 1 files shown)


backend/arm64/emit.mlp
```diff
@@ -1046,9 +1045,17 @@ let emit_instr i =
         end
     | Lop(Const_int n) ->
         emit_intconst i.res.(0) n
-    | Lop (Const_float32 _) ->
-        (* CR mslater: (float32) arm64 *)
-        Misc.fatal_error "float32 is not supported on this architecture"
+    | Lop (Const_float32 f) ->
+        DSL.check_reg Float32 i.res.(0);
+        if f = 0l then
```
xclerc, line 1050: (I am surprised this comparison does
not trigger an error - should the
polymorphic comparison be shadowed
in this file?)

backend/arm64/proc.ml
```diff
@@ -129,12 +126,13 @@ let register_name ty r =
...
-    (* CR mslater: (float32) arm64 *)
-    fatal_error "arm64: got float32 register"
+    float32_reg_name.(r - first_available_register.(1))
   | Valx2 ->
     (* CR mslater: (SIMD) arm64 *)
     fatal_error "arm64: got valx2 register"
 
+(* CR gyorsh for xclerc: [rotate_registers] used in [coloring] on Mach,
+   but not in IRC on CFG. Are we dropping an optimization here? *)
 let rotate_registers = true
```
xclerc, line 136: Possibly; I would remove the binding,
and create an issue to be sure we do
something about it.

backend/arm64/proc.ml
```diff
@@ -146,15 +144,17 @@ let hard_int_reg =
   done;
   v
 
-let hard_float_reg =
+let hard_float_reg_gen kind =
   let v = Array.make 32 Reg.dummy in
   for i = 0 to 31 do
-    v.(i) <- Reg.at_location Float (Reg(100 + i))
+    v.(i) <- Reg.at_location kind (Reg(100 + i))
```
xclerc, line 150: Shouldn't `100` depend on the kind,
or are they synonyms?

gretay-js: I've refactored it slightly, see what you think. 

Constants like this `100` are used about a hunder times all over the backend as you know and we should clean it up separately. In this particular place, `hard_float_reg_gen` was only used with machtype that falls into the `register_class` of float, as the name indicates. I can generalize it slightly for `hard_int_reg` and avoid magic constants.

xclerc: Yes, that should go when we refactor
the classes indeed. You are right, we
should probably not spend more time
on it right now.

xclerc (approved):

PR #3709, opened 2025-03-18
Arm64: Vec128 support

gretay-js:
(on top of #3707, only the last 2 commits are new) 

This PR adds support for `Vec128` in `Proc` and `Emit`.


backend/arm64/cfg_selection.ml
```diff
@@ -173,6 +173,52 @@ class selector =
       (* Other operations are regular *)
       | _ -> super#select_operation op args dbg ~label_after
 
+    method! emit_stores env dbg data regs_addr =
+      let offset = ref (-Arch.size_int) in
```
xclerc, line 177: I think it deserves a comment.
While the reason for the override
can be inferred from the code, it
is probably useful to have a couple
of lines explaining why it is needed.

xclerc (approved):

backend/arm64/arch.ml
```diff
@@ -93,10 +93,8 @@ let division_crashes_on_overflow = false
 
 let identity_addressing = Iindexed 0
 
-let offset_addressing addr delta =
-  match addr with
-  | Iindexed n -> Iindexed(n + delta)
-  | Ibased(s, n) -> Ibased(s, n + delta)
+let offset_addressing _addr _delta =
+  Misc.fatal_error "Arch.offset_addressing not supported"
```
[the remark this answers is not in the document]
gretay-js: No, this will remain `fatal_error` but we can't delete `offset_addressing` function as far as I can see because it's referenced in `Select_utils`.

backend/arm64/cfg_selection.ml
gretay-js: Added comments.

backend/arm64/proc.ml
```diff
@@ -521,9 +514,8 @@ let assemble_file infile outfile =
 let init () = ()
 
 let operation_supported : Cmm.operation -> bool = function
+  | Cprefetch _ | Catomic _ -> false
   | Cpopcnt
```
TheNumbat, line 518: I see `popcnt` was already supported by emit, but is it tested anywhere?

TheNumbat (approved):

PR #3710, opened 2025-03-18
Arm64: emit casts with float32 and vec128

gretay-js:
~(on top of #3709 , only the last commit is new)~

Emit reinterpret cast and static cast that involve Float32 and Vec128. Refactor existing cast and moves to use the new DSL.

### DIFF
--- a/backend/arm64/emit.mlp
+++ b/backend/arm64/emit.mlp
@@ -1031,16 +1031,12 @@ let emit_reinterpret_cast (cast : Cmm.reinterpret_cast) i =
        if distinct then (
         DSL.check_reg Float src;
         DSL.check_reg Float32 dst;
-        (* CR melse: I think this might be broken - maybe we need to do fmov dN,
-           dM without referring to single-precision registers? *)
-          DSL.ins I.FMOV [| DSL.emit_reg dst; DSL.emit_reg src |])
+        DSL.ins I.FCVT [| DSL.emit_reg dst; DSL.emit_reg src |])
     | Float_of_float32 ->
       if distinct then (
         DSL.check_reg Float32 src;
         DSL.check_reg Float dst;
-        (* CR melse: I think this might be broken - maybe we need to do fmov dN,
-           dM without referring to single-precision registers? *)
-        DSL.ins I.FMOV [| DSL.emit_reg dst; DSL.emit_reg src |])
+        DSL.ins I.FCVT [| DSL.emit_reg dst; DSL.emit_reg src |])
     | V128_of_v128 ->
       if distinct then (
         DSL.check_reg Vec128 src;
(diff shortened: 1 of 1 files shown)


backend/arm64/emit.mlp
```diff
@@ -987,21 +981,139 @@ let emit_load_literal dst lbl =
...
-    | _ ->
-      assert false
-  end[@@warning "-4"]
+let move (src : Reg.t) (dst : Reg.t) =
+  let distinct = not (Reg.same_loc src dst) in
+  if distinct then
+  match src.typ, src.loc, dst.typ, dst.loc with
+  | Float, Reg _, Float, Reg _
+  | Float32, Reg _, Float32, Reg _
+  | (Vec128|Valx2), Reg _, (Vec128|Valx2), Reg _ ->
```
TheNumbat, line 990: Hmm, why has `Valx2` shown up already? 
Should this case be split into `Vec128, Vec128 | Valx2, Valx2` or do we allow moves between these types like with `Int/Val/Addr`?

backend/arm64/emit.mlp
```diff
@@ -987,21 +981,139 @@ let emit_load_literal dst lbl =
...
+    | Int32_of_float32 ->
+      DSL.check_reg Float32 src;
+      DSL.ins I.FMOV [| DSL.emit_reg_w dst; DSL.emit_reg src |]
+    | Float32_of_float ->
+       if distinct then (
+        DSL.check_reg Float src;
+        DSL.check_reg Float32 dst;
+        (* CR melse: I think this might be broken - maybe we need to do fmov dN,
+           dM without referring to single-precision registers? *)
+          DSL.ins I.FMOV [| DSL.emit_reg dst; DSL.emit_reg src |])
```
TheNumbat, line 1035: yeah apparently fmov 32<->64 is a static cast

backend/arm64/emit.mlp
gretay-js: yes, it's the same for moves between `Int` and `Val`. `Vec128` and `Valx2` are treated the [same](https://github.com/ocaml-flambda/flambda-backend/blame/3bfc7a86bc18d8cab8e86bb83c0689be1485e3ba/backend/amd64/emit.ml#L829) in amd64 backend, and we had a similar discussion there when `Valx2` was added, can't find the comment now. IIRC register allocation can introduce these moves when coalescing.

backend/arm64/emit.mlp
gretay-js: fixed to use FCVT not FMOV and removed the addressed CR.

TheNumbat: Wrong way! it needs fmov 64->64

TheNumbat: We should probably also add some tests for reinterpret float/float32

gretay-js: oops, i got it mixed up with static cast, thanks for the catch. the tests will be enabled in a follow-up pr.

TheNumbat (approved):

PR #3711, opened 2025-03-18
Refactor simd tests

gretay-js:
(requires #3705 and #3704 to be merged first)

Factor out target specific intrinsics. The tests are still disabled on arm64 in this patch and arm64 intrinsics are not implemented.


gretay-js:
superseded by https://github.com/oxcaml/oxcaml/pull/4140

PR #3712, opened 2025-03-18
Annotate Unix.file_descr as immediate

glittershark:
It's an int under the hood, so annotating it as immediate lets it mode cross
appropriately.

### DIFF
--- a/otherlibs/unix/unix.mli
+++ b/otherlibs/unix/unix.mli
@@ -322,7 +322,7 @@ val nice : int -> int
 (** {1 Basic file input/output} *)
 
 
-type file_descr
+type file_descr : immediate
 (** The abstract type of file descriptors. *)
 
 val stdin : file_descr


tdelvecchio-jsc (approved):
Also checked whether operations on file descriptors are synchronized. Technically this isn't necessary since none of the functions here are marked as portable, but yes, they all seem synchronized.

lthls:
This is wrong on Windows, where `Unix.file_descr` values are custom blocks.

tdelvecchio-jsc (dismissed):
See comment from lthls.

glittershark:
> This is wrong on Windows, where `Unix.file_descr` values are custom blocks.

are they at least `immutable_data`?

lthls:
> are they at least `immutable_data`?

I don't know the semantics of `immutable_data`, but as custom blocks they behave in a very similar way to bigarrays, if that helps.

mshinwell:
I would suggest just leaving a CR about this in the Windows code.

mshinwell (approved):
Rebased and CR added

PR #3714, opened 2025-03-20
Delete `backend/amd64/selection.ml`

xclerc:
As per title, this is dead code that should
have been deleted in #3287.


xclerc:
(The CI failure is due to the long-standing evasive arm64 bug.)

gretay-js (approved):

PR #3716, opened 2025-03-20
Add `Modes.Aliased` wrapper for aliased values in a unique context

polytypic:
This adds a `Modes.Aliased.t` wrapper for aliased values in a unique context that has turned out to be useful.

### DIFF
--- a/stdlib/modes.ml
+++ b/stdlib/modes.ml
@@ -12,6 +12,11 @@
 (*                                                                        *)
 (**************************************************************************)
 
+module Aliased = struct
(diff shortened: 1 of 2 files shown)


lpw25:
@riaqn Does this need to include `global` to get the long-term behaviour of the `aliased` modality of implying `global`? Also, can we fix that soon?

riaqn:
@lpw25 Do you mean "global implying aliased"? We don't have that yet. I can fix that, but the changes will be in `typemode.ml` which I recall will be refactored by a future PR of yours?

polytypic:
Seems that `Aliased` was already added via some other PR.

PR #3722, opened 2025-03-20
Improve runtime clarity with Scannable_* macros

NickBarnes:
The runtime has a lot of comparisons of header tag fields against `No_scan_tag`. Newcomers to the runtime may be confused about the sign of the comparison, or about whether it is inclusive or exclusive. Replacing it with a set of macros `Scannable_tag`, `Scannable_hd`, `Scannable_val` clarifies the code and makes bugs like [this](https://github.com/ocaml/ocaml/blob/979ce41050ddaf772cd1f372b571b924d35a4325/runtime/weak.c#L293) less likely.


NickBarnes:
Rebased and tested.

PR #3725, opened 2025-03-24
arm64.S changes for locals x effects

mshinwell:
This provides the necessary support in the runtime for effects on arm64 in the presence of locals, following on from #3648 .


TheNumbat (approved):
LGTM

PR #3728, opened 2025-03-24
No allocation between stack creation and usage

dkalinichenko-js:
See https://github.com/ocaml-flambda/flambda-backend/pull/3231. No tests here, since the issue is very hard to replicate, but 
this fixes the issue in https://github.com/ocaml-flambda/flambda-backend/pull/3541.

@TheNumbat should `last_fiber` be subject to the same treatment? Garbage collection code in the runtime doesn't seem to touch the second field of the continuation.


otherlibs/stdlib_alpha/effect.ml
```diff
@@ -392,8 +392,14 @@ type ('a, 'e) perform = ('a, 'e) op * 'e Handler.t'
 
 external perform : ('a, 'e) perform -> 'a = "%perform"
 
+(* Stacks are represented as tagged pointers, and garbage collection
```
TheNumbat: It's not _quite_ true that GC invalidates them, just that they do not keep their contents alive, so if they are the only remaining reference they will be invalidated. This is the case when we allocate a fresh stack or swap a stack out of a cont.

otherlibs/stdlib_alpha/effect.ml
```diff
@@ -392,8 +392,14 @@ type ('a, 'e) perform = ('a, 'e) op * 'e Handler.t'
...
+   invalidates them. Therefore, no allocations must happen between
+   creation and use of a [stack]. *)
 type (-'a, +'b) stack : immediate
 
+(* Stacks are represented as tagged pointers, and garbage collection
```
TheNumbat: ...and `last_fiber`s are not relied upon to keep fibers alive, since there should always be another reference via `cont.fiber(->parent)*`

TheNumbat (approved):
I've moved the unsafe parts into a module to be sure nothing else calls those externals.
Looks good now (you can resolve comments)

PR #3729, opened 2025-03-24
Remove allocation near unsafe effect primitive

stedolan:
The `alloc_stack` primitive used in effects is completely unsafe, and the runtime may crash if its result is held live across an allocation. That happens in one place, causing the `evenodd.ml` test to fail in bytecode on my machine (depending on, among other things, the length of the path to the test directory).

The right fix is to change this primitive to be less unsafe (e.g. by returning a continuation, not a `stack`). That's left for another PR, this one just removes the allocation.


mshinwell (approved):

dkalinichenko-js:
https://github.com/ocaml-flambda/flambda-backend/pull/3728 refactors code in effect.ml to be more robust to such issues.

PR #3730, opened 2025-03-24
Dynamic adjustment of space_overhead

stedolan:
This patch adds a new GC tweak, `Xgc_overhead_adjustment`, which dynamically adjusts the current space_overhead according to the promotion rate.

With this patch, when the new GC policy is enabled, the space_overhead specified by the user is in fact the space_overhead setting at a nominal promotion rate of 10%. The actual space_overhead used by the GC will be larger or smaller than this one, when the promotion rate is respectively above or below 10%.

Roughly, the idea is that when large numbers of allocations into the major heap are being done, extra GC work to deal with those allocations is unusually expensive. So at such times, we should be willing to use more space to save time, which corresponds to a larger space_overhead.

The actual space_overhead calculation is to scale it by the promotion rate (normalised so that 10% = 1), raised to the power of `gc_overhead_adjustment` (in percent). This means that `Xgc_overhead_adjustment=0` is the current behaviour, and `Xgc_overhead_adjustment=50` adjusts overhead by the square root of promotion rate (which some theory suggests is optimal, but let's see).

I tried this mechanism out on a simple benchmark with a heavy-allocation and a light-allocation phase:
```ocaml
let () =
  (* Phase I: build a big data structure (high promotion rate) *)
  let thing = List.init 1_000 (fun _ -> List.init 50_000 ref) in
  (* Phase II: allocate mostly garbage (low promotion rate) *)
  for _i = 1 to 50_000 do ignore (List.init 20_000 ref) |> Sys.opaque_identity done;
  ignore (Sys.opaque_identity thing);
```

On this benchmark, the patch improves the space-time tradeoff at all values of space_overhead. The highlighted point is the default `o=80`:
download

### DIFF
--- a/runtime/major_gc.c
+++ b/runtime/major_gc.c
@@ -40,6 +40,7 @@
 #include "caml/startup_aux.h"
 #include "caml/weak.h"
 #include "caml/custom.h"
+#include "caml/minor_gc.h"
 
 /* NB the MARK_STACK_INIT_SIZE must be larger than the number of objects
    that can be in a pool, see POOL_WSIZE */
@@ -80,10 +81,15 @@ uintnat caml_percent_sweep_per_mark = 120; /* TODO: benchmark this value */
 /* Allowable values of caml_gc_pacing_policy */
 
 #define GC_PACING_OCAML_53        0 /* Pacing as in OCaml 5.3 (plus mark-delay) */
-#define GC_PACING_2025_SIMPLE     1 /* simple variant of 2025 pacing */
-#define GC_PACING_2025_FULL       2 /* full 2025 pacing (TODO) */
+#define GC_PACING_2025            1 /* 2025 pacing */
 
 uintnat caml_gc_pacing_policy = GC_PACING_OCAML_53;
+
+/* The degree to which space_overhead should be dynamically adjusted according
+   to the promotion rate.
+
+   (0 = disabled, 50 = theoretically optimal space/time tradeoff) */
+uintnat caml_gc_overhead_adjustment = 0;
 
 /* This variable is only written with the world stopped, so it need not be
    atomic */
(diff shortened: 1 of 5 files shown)


runtime/major_gc.c
```diff
@@ -749,12 +757,32 @@ static uintnat gc_slice_work(uintnat heap_words,
...
+      /* In theory, we should adjust according to the allocation rate.
+
+         However, this produces a dependence on real time, making the GC
+         nondeterministic even for single-threaded programs and doing weird
+         things for programs which pause in I/O for a long time.
+
+         Instead, we use minor heap allocation as a proxy for time, so instead
+         of measuring the allocation rate we measure the promotion rate.
+
+         The promotion rate is scaled so that the nominal promotion rate of 10%
```
NickBarnes, line 774: Perhaps measure; 10% is really a guess.

runtime/major_gc.c
```diff
@@ -749,12 +757,32 @@ static uintnat gc_slice_work(uintnat heap_words,
...
+         comes out as 1.0, and direct-to-major allocations are deemed to have
+         this nominal promotion rate. */
+      double scaled_prom_rate =
+        (10.0 * allocated_words) /
+        (10.0 * allocated_direct_words + minor_words);
```
NickBarnes, line 779: Can this divide by zero?

NickBarnes (changes requested):
This looks OK as long as it's turned off! Modulo the possible divide-by-zero I think we can merge this and then improve it later.

runtime/major_gc.c
stedolan: Right, agreed.

But I'd point out that this is not a statement that "most programs have 10% promotion", but just a specification of what the space_overhead parameter does. (The 10% is effectively just a choice of units for space_overhead).

So if the average promotion rate across many programs turns out to be 7.35%, then I think we should still keep this at 10%, to keep the specification of what it does memorable.

(In other words, I agree we should measure more, but then we should pick this number to be whichever of 1%, 2%, 5%, 10%, 20% is closest to most programs)

runtime/major_gc.c
stedolan: Hmm, probably. (Hard to trigger this code without doing some allocation, so maybe not. Worth fixing regardless)

runtime/major_gc.c
NickBarnes: Yes, but we might want to change the default `space_overhead` accordingly.

mshinwell (approved):

PR #3734, opened 2025-03-25
Improve simplification of `is_null` in Flambda

dkalinichenko-js:
Copy the approach taken for `is_int`.

### DIFF
--- a/middle_end/flambda2/simplify/simplify_unary_primitive.ml
+++ b/middle_end/flambda2/simplify/simplify_unary_primitive.ml
@@ -900,8 +900,8 @@ let simplify_lazy ~original_prim dacc ~original_term ~arg:_ ~arg_ty:_
     (P.result_kind' original_prim)
     ~original_term
 
-let simplify_is_null dacc ~original_term ~arg:_ ~arg_ty:scrutinee_ty
-    ~result_var =
+let simplify_is_null dacc ~original_term ~arg:_ ~arg_ty:scrutinee_ty ~result_var
+    =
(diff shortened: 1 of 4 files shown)


lthls:
I think the description ("Copy the approach taken for `Is_int`") is misleading.
There are two cases for `Is_int`: the variant-only case is for primitives inserted by the compiler (during pattern-matching compilation or later for splitting switches), while the other case is for `Obj.is_int`.
`Is_null` is currently handled like the first case, and this PR proposes to handle it like the second case instead.

For `Is_int`, the variant-only case focuses on relational information: the result of the primitive is linked to properties on the argument, so later matching on the result allows us to also learn more about the value.
`Obj.is_int` cannot be handled like that for various reasons (it would require some changes in how we represent Flambda2 types), so we go instead for constant propagation only (i.e. check if the argument allows us to know the result is constant).

So at first glance, this looks like a regression: instead of being able to link the result of the primitive with the statis of its argument, we know only propagate forward. This regression should show up in cases where we match twice on the same `or_null` value, in a nested way:
```ocaml
let f x =
  match x with
  | This _ ->
    begin match x with
      | This y -> ...
      | Null -> ... (* This branch should disappear after Flambda *)
    end
  | Null -> ...
```

lthls:
> Oh, I see. Is `(* CR layouts v3: implement a real simplifier. *)` outdated then?

Probably, yes. It comes from #3257, which contains commits from both of us, and I think you had a placeholder implementation with that CR, and I wrote the current implementation but forgot to remove the CR.

PR #3737, opened 2025-03-25
Linscan: use doubly-linked lists for intervals

xclerc:
This pull request change the list implementation
used to stored intervals from `list` to
`Doubly_linked_list.t`.

The existing fields storing such interval list are
renamed from `xyz` to `xyz_list`, and additional
fields named `xyz_dll` are added. The `xyz_dll`
contain a copy of the list, simply using the new
representation. The functions updating the lists
are modified to check whether the contents of
`xyz_list` and `xyz_dll` are equal.

A follow-up pull request will delete the `xyz_list`
fields.

### DIFF
--- a/backend/regalloc/regalloc_ls.ml
+++ b/backend/regalloc/regalloc_ls.ml
@@ -140,18 +140,21 @@ let allocate_free_register : State.t -> Interval.t -> spilling_reg =
     | num_available_registers ->
       let available = Array.make num_available_registers true in
       let num_still_available = ref num_available_registers in
-      let set_not_available r =
+      let set_not_available (r : int) : unit =
         let idx = r - first_available in
         if available.(idx) then decr num_still_available;
         available.(idx) <- false;
         if !num_still_available = 0 then raise No_free_register
       in
-      List.iter intervals.active ~f:(fun (interval : Interval.t) ->
-          match interval.reg.loc with
-          | Reg r ->
-            if r - first_available < num_available_registers
-            then set_not_available r
-          | Stack _ | Unknown -> ());
+      let set_not_available_if_valid_phys_reg (interval : Interval.t) : unit =
+        match interval.reg.loc with
+        | Reg r ->
+          if r - first_available < num_available_registers
+          then set_not_available r
+        | Stack _ | Unknown -> ()
+      in
+      List.iter intervals.active_list ~f:set_not_available_if_valid_phys_reg;
+      DLL.iter intervals.active_dll ~f:set_not_available_if_valid_phys_reg;
       let remove_bound_overlapping (itv : Interval.t) : unit =
         match itv.reg.loc with
         | Reg r ->
@@ -161,8 +164,10 @@ let allocate_free_register : State.t -> Interval.t -> spilling_reg =
           then set_not_available r
         | Stack _ | Unknown -> ()
       in
-      List.iter intervals.inactive ~f:remove_bound_overlapping;
-      List.iter intervals.fixed ~f:remove_bound_overlapping;
+      List.iter intervals.inactive_list ~f:remove_bound_overlapping;
+      DLL.iter intervals.inactive_dll ~f:remove_bound_overlapping;
+      List.iter intervals.fixed_list ~f:remove_bound_overlapping;
+      DLL.iter intervals.fixed_dll ~f:remove_bound_overlapping;
       let rec assign idx =
         if idx >= num_available_registers
         then Misc.fatal_error "No_free_register should have been raised earlier"
@@ -170,8 +175,8 @@ let allocate_free_register : State.t -> Interval.t -> spilling_reg =
         then (
           reg.loc <- Reg (first_available + idx);
           reg.spill <- false;
-          intervals.active
-            <- Interval.List.insert_sorted intervals.active interval;
+          intervals.active_list
+            <- Interval.List.insert_sorted intervals.active_list interval;
           if debug
           then (
             indent ();
@@ -188,7 +193,7 @@ let allocate_blocked_register : State.t -> Interval.t -> spilling_reg =
   let reg = interval.reg in
   let reg_class = Proc.register_class reg in
   let intervals = State.active state ~reg_class in
-  match intervals.active with
+  match intervals.active_list with
   | hd :: tl ->
     let chk r =
       assert (same_reg_class r.Interval.reg hd.Interval.reg);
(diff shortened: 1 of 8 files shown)


utils/doubly_linked_list.mli
```diff
@@ -87,6 +91,13 @@ val transfer : to_:'a t -> from:'a t -> unit -> unit
 
 val map : 'a t -> f:('a -> 'b) -> 'b t
 
+val copy : 'a t -> 'a t
+
+(* Deletes the passed cell and all the cells after it. *)
+val cut : 'a cell -> unit
```
spiessimon, line 97: Perhaps this should be moved to the top of the interface with all the other cell functions. Also, at first, from the name, I thought it cuts out a single cell. Perhaps it might be worth renaming the function to, for example, `cut_from`, to clarify this.

utils/doubly_linked_list.ml
```diff
@@ -358,24 +363,50 @@ let transfer ~to_ ~from () =
...
+    let rec aux eq left right =
+      match left, right with
+      | None, None -> true
+      | (None, (Some _)) | ((Some _), None) -> false
+      | Some left_cell, Some right_cell ->
+        let left_value = value left_cell in
+        let right_value = value right_cell in
+        eq left_value right_value && aux eq (next left_cell) (next right_cell)
+    in
+    aux eq (hd_cell left) (hd_cell right)
```
spiessimon, line 409: I think this iteration will create various cells along the way. Since the comparison is based only on the value, I think it would make sense to iterate over the nodes directly instead. (I have not compiled or tested the code below.)

```suggestion
  let equal eq left right =
    let rec aux eq left right =
      match left, right with
      | Empty, Empty -> true
      | (Empty, Node _) | (Node _, Empty) -> false
      | Node { value = left_value; next = left_next; prev = _ }, 
        Node { value = right_value; next = right_next; prev = _ } ->
        eq left_value right_value && aux eq left_next right_next
    in
    aux eq left.first right.first
```

backend/regalloc/regalloc_ls.ml
```diff
@@ -159,17 +162,20 @@ let allocate_free_register : State.t -> Interval.t -> spilling_reg =
           then set_not_available r
         | Stack _ | Unknown -> ()
       in
-      List.iter intervals.inactive ~f:remove_bound_overlapping;
-      List.iter intervals.fixed ~f:remove_bound_overlapping;
+      List.iter intervals.inactive_list ~f:remove_bound_overlapping;
```
spiessimon, line 165: Same as above. These don't seem to change the list.

backend/regalloc/regalloc_ls_utils.ml
```diff
@@ -207,65 +227,204 @@ module Interval = struct
...
+    in
+    let consistent_active =
+      equal_list_dll Interval.equal t.active_list t.active_dll
+    in
+    let consistent_inactive =
+      equal_list_dll Interval.equal t.inactive_list t.inactive_dll
+    in
+    if not (consistent_fixed && consistent_active && consistent_inactive)
+    then (
+      print Format.err_formatter t;
```
spiessimon, line 302: Out of curiosity, did this ever trigger while you were working on adding the linked list implementation?

backend/regalloc/regalloc_ls_state.ml
```diff
@@ -3,54 +3,136 @@
...
+type class_interval_array =
+  { elements : Interval.t array;
+    mutable length : int
+  }
+
+let dummy_interval =
+  { Interval.reg = Reg.dummy;
+    begin_ = -1;
+    end_ = -1;
+    ranges = DLL.make_empty ()
```
spiessimon, line 61: It is quite important here that the result of `DLL.make_empty ()` returns a list that can be duplicated without duplicating references. Might be worth a comment. 
```suggestion
    ranges = DLL.make_empty () (* this list can be safely duplicated below, because the empty list is stateless *)
```

backend/regalloc/regalloc_ls_state.ml
```diff
@@ -3,54 +3,136 @@
...
+  | c -> c
 
 let[@inline] update_intervals state map =
-  let active = state.active in
+  check_consistency state "State.update_intervals/begin";
+  let active : ClassIntervals.t array = state.active in
   Array.iter active ~f:ClassIntervals.clear;
-  state.intervals
+  let active' : class_interval_array array =
+    Array.init (Array.length active) ~f:(fun _ -> make_class_interval_array ())
```
spiessimon, line 96: This seems to not have the entries from `state.active` but empty entries instead. I don't understand the code here enough to tell whether that makes a difference.

backend/regalloc/regalloc_ls_state.ml
```diff
@@ -3,54 +3,136 @@
...
+  let class_intervals = extract_class_interval_array class_intervals in
+  Array.sort ~cmp:compare_asc_begin class_intervals;
+  DLL.clear state.interval_dll;
+  DLL.add_array state.interval_dll class_intervals;
+  if ls_debug
+  then log_interval_list ~indent:1 ~kind:"regular" state.interval_list;
+  Array.iteri active' ~f:(fun i (intervals : class_interval_array) ->
+      let extracted = extract_class_interval_array intervals in
+      Array.sort extracted ~cmp:compare_desc_end;
+      DLL.clear active.(i).fixed_dll;
```
spiessimon, line 123: I did not understand completely what is going on here. This is duplicating the code for the regular lists, because we are maintaining the two representations in this PR, right? It seems to sort the same as for lists below.

backend/regalloc/regalloc_ls_state.ml
spiessimon: Ah, I saw in the other PR that `active` will stay around.

utils/doubly_linked_list.mli
xclerc: Done

utils/doubly_linked_list.ml
xclerc: Done

backend/regalloc/regalloc_ls_utils.ml
xclerc: Yes, it did catch a couple of bugs...

backend/regalloc/regalloc_ls_state.ml
xclerc: Done

backend/regalloc/regalloc_ls.ml
```diff
@@ -138,18 +138,21 @@ let allocate_free_register : State.t -> Interval.t -> spilling_reg =
...
-            then set_not_available r
-          | Stack _ | Unknown -> ());
+      let set_not_available_if_valid_phys_reg (interval : Interval.t) : unit =
+        match interval.reg.loc with
+        | Reg r ->
+          if r - first_available < num_available_registers
+          then set_not_available r
+        | Stack _ | Unknown -> ()
+      in
+      List.iter intervals.active_list ~f:set_not_available_if_valid_phys_reg;
```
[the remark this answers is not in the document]
xclerc: I agree this is misleading; I thought it would
made the follow-up more obvious, but that
was quite possibly a mistake.

backend/regalloc/regalloc_ls_state.ml
xclerc: Yes, it is, https://github.com/ocaml-flambda/flambda-backend/pull/3751 will keep only the "new" representation.

backend/regalloc/regalloc_ls.ml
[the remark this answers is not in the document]
spiessimon: Since the other PR removes the list operations, I think in the end the semantics is preserved. Probably a good idea to merge the two PRs in short succession in any case.

spiessimon (approved):
Looks good to me! Thank you for addressing my comments.

PR #3738, opened 2025-03-25
Don't forget about the yielding axis in modality

glittershark:
The `Yielding` axis was forgotten about in Modality.to_list


goldfirere (approved):
Is there a way we can think of for avoiding this problem in the future? I can't think of anything quickly other than a comment on the appropriate type definitions, but maybe you can. We should merge regardless.

riaqn:
>Is there a way we can think of for avoiding this problem in the future?

We should replace `to_list` with a record type, so that a missing field would be type error.

PR #3743, opened 2025-03-25
Merge Cmm Ctrywith with Ccatch

lthls:
The `Ctrywith` constructor in `Cmm` used to encode both declaring an exception handler and wrapping the body with the proper pushtrap/poptrap instructions.
Flambda2 needed a way to dissociate these two concepts, with the introduction of the handler being similar to `Ccatch` but the trap instructions being encoded elsewhere. For what I remember as upstreaming reasons, we chose to encode these handlers as special "delayed" cases of `Ctrywith` rather than `Ccatch`.
Since then, we have removed the support for non-Flambda2 middle-ends, so all the `Ctrywith` we generate are delayed. It seems now might be a good moment to finally merge these delayed exception handlers back into the `Ccatch` case where they fit better.

I need to add some comments here and there for some tricky parts of the patch, I will try to do this either tomorrow or on Friday.

### DIFF
--- a/backend/cfg_selectgen.ml
+++ b/backend/cfg_selectgen.ml
@@ -1082,8 +1082,9 @@ module Make (Target : Cfg_selectgen_target_intf.S) = struct
     let sub_handlers =
       List.map
         (fun ((rs, label), (_, sub_handler)) ->
-          begin match flag with
-          | Normal | Recursive -> Sub_cfg.add_empty_block_at_start sub_handler ~label
+          (match flag with
+          | Normal | Recursive ->
+            Sub_cfg.add_empty_block_at_start sub_handler ~label
           | Exn_handler ->
             Sub_cfg.mark_as_trap_handler sub_handler ~exn_label:label;
             let exn_bucket_in_handler =
@@ -1092,8 +1093,7 @@ module Make (Target : Cfg_selectgen_target_intf.S) = struct
               | r :: _ -> r
             in
             Sub_cfg.add_instruction_at_start sub_handler (Cfg.Op Move)
-              [| Proc.loc_exn_bucket |] exn_bucket_in_handler Debuginfo.none
-          end;
+              [| Proc.loc_exn_bucket |] exn_bucket_in_handler Debuginfo.none);
           sub_handler)
         l
     in
@@ -1330,9 +1330,9 @@ module Make (Target : Cfg_selectgen_target_intf.S) = struct
       in
       (* CR vlaviron: I hit an issue in the non-tail case where this
          [add_empty_block_at_start] call interfered with the
-         [mark_as_trap_handler] call for the exception case (creating two
-         blocks with the same label). I suspect the issue should occur here
-         too but happens to never be triggered in the compiler itself and its
+         [mark_as_trap_handler] call for the exception case (creating two blocks
+         with the same label). I suspect the issue should occur here too but
+         happens to never be triggered in the compiler itself and its
          testsuite. *)
       Sub_cfg.add_empty_block_at_start seq ~label;
       label, rs, seq
@@ -1365,21 +1365,20 @@ module Make (Target : Cfg_selectgen_target_intf.S) = struct
     let s_handlers =
       List.map
         (fun (label, rs, s) ->
-           let () =
-             match flag with
-             | Normal | Recursive -> ()
-             | Exn_handler ->
-               Sub_cfg.mark_as_trap_handler s ~exn_label:label;
-               let exn_bucket_in_handler =
-                 match rs with
-                 | [] ->
-                   Misc.fatal_error "Exception handler with no parameters"
-                 | r :: _ -> r
-               in
-               Sub_cfg.add_instruction_at_start s (Cfg.Op Move)
-                 [| Proc.loc_exn_bucket |] exn_bucket_in_handler Debuginfo.none
-           in
-           s)
+          let () =
+            match flag with
+            | Normal | Recursive -> ()
+            | Exn_handler ->
+              Sub_cfg.mark_as_trap_handler s ~exn_label:label;
+              let exn_bucket_in_handler =
+                match rs with
+                | [] -> Misc.fatal_error "Exception handler with no parameters"
+                | r :: _ -> r
+              in
+              Sub_cfg.add_instruction_at_start s (Cfg.Op Move)
+                [| Proc.loc_exn_bucket |] exn_bucket_in_handler Debuginfo.none
+          in
+          s)
         new_handlers
     in
     Sub_cfg.join_tail ~from:(s_body :: s_handlers) ~to_:sub_cfg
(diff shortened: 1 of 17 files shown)


flambda-backend/testsuite/tools/dune
```diff
@@ -13,7 +13,7 @@
 
 (executable
  (name codegen_main)
- (modes byte)
+ (modes native)
```
lthls, line 16: Note: this unrelated change is here because one of the asmgen test failed, and the `codegen` binary used in this test proved completely useless for debugging. Using `native` mode means that the generated executable contains enough debug info to actually print backtraces, and can be debugged with a native debugger.

lthls:
This PR has been rebased onto the recent big changes to `Cfg_selectgen`.

backend/cfg_selectgen.ml
```diff
@@ -1085,8 +1081,19 @@ module Make (Target : Cfg_selectgen_target_intf.S) = struct
...
     let sub_handlers =
       List.map
-        (fun ((_, _, _, label), (_, sub_handler)) ->
-          Sub_cfg.add_empty_block_at_start sub_handler ~label;
+        (fun ((rs, label), (_, sub_handler)) ->
+          (match flag with
+          | Normal | Recursive ->
+            Sub_cfg.add_empty_block_at_start sub_handler ~label
+          | Exn_handler ->
+            Sub_cfg.mark_as_trap_handler sub_handler ~exn_label:label;
```
mshinwell, line 1089: This block can be shared with the tail case, I think?

backend/cfg_selectgen.ml
```diff
@@ -1414,8 +1328,14 @@ module Make (Target : Cfg_selectgen_target_intf.S) = struct
                     [||] [||])
               ids_and_rs)
       in
+      (* CR vlaviron: I hit an issue in the non-tail case where this
```
mshinwell, line 1331: Was there a reason this wasn't fixed at this point?  Can we add a test case?

backend/select_utils.ml
```diff
@@ -116,8 +106,29 @@ let print_traps ppf traps =
   in
   Format.fprintf ppf "(%a)" print_traps traps
 
-let set_traps nfail traps_ref base_traps exit_traps =
-  let traps = combine_traps base_traps exit_traps in
+let rec set_traps env nfail traps_ref base_traps exit_traps =
```
mshinwell, line 109: Could we try implementing the "proper" solution as well, as it seems like it might not be too much code, and easier to understand?

mshinwell (changes requested):

backend/select_utils.ml
lthls: I wrote this PR before #3768, it might not be necessary now. I'll check and update the PR as needed.

backend/cfg_selectgen.ml
lthls: I didn't fix it because I didn't know whether I had wrongly mixed `add_empty_block_at_start` and `mark_as_trap_handler`, or if that was supposed to work but didn't for other reasons.
By leaving this comment I was hoping that a reviewer would be able to enlighten me.

xclerc: I am not sure I follow; `mark_as_trap_handler`
does not create a new block, so I am not sure
how it would interfere. `emit_tail_trywith` used
to call `Cmm.new_label`; what would be the
equivalent in the new version?

lthls: The call to `Cmm.new_label ()` should still be there, when computing `handlers_map`

xclerc: Ah, yes, sorry. And I was misrimembering
what `mark_as_trap_handler` was supposed
to do. While it does not create a new block,
it changes the label. In the new implementation,
shouldn't `mark_as_trap_handler` only set
`is_trap_handler` to `true`?

lthls: That looks correct, thanks. I'll implement that.

backend/select_utils.ml
lthls: It is still necessary, as `Cfg_simplify` (which is supposed to handle the removal of dead exception handlers) runs too late: `Select_utils.Stack_offset_and_exn.update_cfg` fails if it encounters any `Pushtrap` operations for which the handler does not exist, and `Cfg_simplify` only runs after that.

backend/cfg_selectgen.ml
lthls: Yes, I have done that now.

backend/select_utils.ml
gretay-js: - Is this "weird piece of code" replacing what we used to have in `env_enter_trywith`?  
- calling `set_traps` recursively with an empty exit list is confusing here. the recurisve call is guaranteed to skip `List.fold_left` in L120-130. and go directly to update `traps_ref` in L132-142. If this is right, factoring out L132-142 into a little helper function and just calling that would make this code easier to understand and no need for `rec`.

lthls: I have done this refactoring and updated the comment.

gretay-js: This is much clearer, thanks!

backend/select_utils.ml
```diff
@@ -131,13 +120,36 @@ let set_traps nfail traps_ref base_traps exit_traps =
...
+       instead, we make every Push trap action count as a use of the handler. *)
+    List.fold_left
+      (fun trap_stack trap ->
+        (match trap with
+        | Push lbl -> (
+          match env_find_static_exception lbl env with
+          | { traps_ref; _ } -> set_traps_for_handler lbl traps_ref trap_stack
+          | exception Not_found ->
+            Misc.fatal_errorf "Trap %d not registered in env" lbl)
+        | Pop _ -> ());
```
gretay-js, line 140: Which iterator are you referring to in this comment? 

I am guessing that it's [this line](https://github.com/ocaml-flambda/flambda-backend/blob/9f948e09581b0387f427da353b15776251104ff6/backend/cfg_selectgen.ml#L1037) in `Cfg_selectgen.translate_one_handler` that fails when we have an unreachable trap handler.

In the previous version (before this PR) we had special handling for these cases: see L1222 onwards in `Cfg_selectgen` fabricating a `segfault` handler. Perhaps we should have something simiilar in the new version? In `translate_one_handler` we can check the `flag` in the `Unreachable` case, and either remove the handler on the fly (maybe as part of `build_all_reachable_handlers` ?) or fabricate a `segfault` handler as before.

By the way, `Select_utils.Stack_offset_and_exn` may not be needed any more, because we can directly compute stack_offset during selection when constructing the traps of each block, instead of having to recover this information from push/poptrap instructions the way it was done when CFG was constructed later in the pipeline.

cc @xclerc

lthls: The iterator that is failing is `Select_utils.Stack_offset_and_exn`.

Regarding the segfault handler version, I missed that when I wrote this PR; I think it can apply with this PR too. I will try it and let you know if it works.

lthls: I have pushed a version where unreachable exception handlers are kept with a dummy expression (and reverted the changes to `set_traps`).

mshinwell (approved):

PR #3746, opened 2025-03-26
arm64: popcnt/ctz without CSSC

gretay-js:
Emit a sequence for `Ipopcnt` and  `Ictz` on older version of `armv8` that don't support CSSC. Add a flag to enable the short sequence (off by default). Tests coming up in a separate PR.

(needs #3747)

### DIFF
--- a/backend/arm64/proc.ml
+++ b/backend/arm64/proc.ml
@@ -86,6 +86,7 @@ let phys_reg ty n =
   | Float -> hard_float_reg.(n - 100)
   | Float32 -> hard_float32_reg.(n - 100)
   | Vec128 | Valx2 -> hard_vec128_reg.(n - 100)
+
 let reg_x8 = phys_reg Int 8
 
 let stack_slot slot ty =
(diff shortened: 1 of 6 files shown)


backend/arm64/emit.mlp
```diff
@@ -255,7 +255,7 @@ let name_for_int_operation = function
   | Ilsr -> "lsr"
   | Iasr -> "asr"
   | Iclz { arg_is_non_zero = _ } -> "clz"
```
TheNumbat, line 257: The `Iclz` case is also unused now

backend/arm64/proc.ml
```diff
@@ -393,6 +393,11 @@ let destroyed_at_basic (basic : Cfg_intf.S.basic) =
     -> [||]
   | Op (Static_cast
           (V128_of_scalar _|Scalar_of_v128 _))
+  | Op (Intop (Ipopcnt | Ictz _ )) ->
```
TheNumbat, line 396: Only popcnt needs this right?

backend/arm64/emit.mlp
```diff
@@ -1357,12 +1361,25 @@ let emit_instr i =
...
         (* `ctz Rd, Rn` is optionally supported from Armv8.7, but rbit and clz
            are supported in all ARMv8 CPUs. *)
-        ` rbit {emit_reg i.res.(0)}, {emit_reg i.arg.(0)}\n`;
-        ` clz  {emit_reg i.res.(0)}, {emit_reg i.res.(0)}\n`
-    | Lop(Intop (Iclz _ as op)) ->
-        let instr = name_for_int_operation op in
-        `	{emit_string instr}	{emit_reg i.res.(0)}, {emit_reg i.arg.(0)}\n`
-    | Lop(Intop ((Iadd|Isub|Imul|Idiv|Iand|Ior|Ixor|Ilsl|Ilsr|Iasr|Ipopcnt) as op)) ->
+        if !Arch.feat_cssc then
+           `	ctz	{emit_reg i.res.(0)}, {emit_reg i.res.(0)}\n`
```
TheNumbat, line 1365: These are both `i.res` 🤔

gretay-js: oh, good catch. fixed.

TheNumbat (approved):

PR #3747, opened 2025-03-26
Arm64: more simd intrinsics and instructions

gretay-js:
Add a few more arm64 intrinsics that are needed for some of simd tests.


backend/arm64_ast.ml
```diff
@@ -42,7 +42,7 @@ module Neon_reg_name = struct
       | Q
 
     let to_string t =
-      match t with B -> "B" | H -> "H" | S -> "S" | D -> "D" | Q -> "Q"
+      match t with B -> "b" | H -> "h" | S -> "s" | D -> "d" | Q -> "q"
```
gretay-js, line 45: Using lowercase here to match assembly that comes out of `emit.mlp`.

gretay-js:
The build failure in `tests/lib-systhreads/threadsigmask.ml` is not related to this patch, it was seen in other branches.

xclerc (approved):

PR #3751, opened 2025-03-26
Linscan: keep only the doubly-linked lists.

xclerc:
(Only CI at this point.)


backend/regalloc/regalloc_ls.ml
```diff
@@ -189,28 +184,28 @@ let allocate_blocked_register : State.t -> Interval.t -> spilling_reg =
...
        && not
-            (List.exists ~f:chk intervals.fixed_list
-            || List.exists ~f:chk intervals.inactive_list)
+            (DLL.exists ~f:chk intervals.fixed_dll
+            || DLL.exists ~f:chk intervals.inactive_dll)
     then (
       (match hd.reg.loc with Reg _ -> () | Stack _ | Unknown -> assert false);
       interval.reg.loc <- hd.reg.loc;
-      intervals.active_list <- Interval.List.insert_sorted tl interval;
       (match DLL.hd_cell intervals.active_dll with
```
spiessimon, line 201: This check is a left over from the previous PR. Unless I'm missing something, I think the match can be removed and `hd_cell` can be used in place of `cell` below.

xclerc: Indeed - thanks

xclerc:
The implementations of the register allocators
are "guarded" by the validator, which is expected
to catch most if not all possible miscompilations.
However, there are other regressions that could
be caught by the script.

xclerc:
(The comparison script was run successfully.)

PR #3752, opened 2025-03-26
Ensure that caml_check_pending_signals does not report blocked signals

stedolan:
Since #3740, we're seeing frequent deadlocks in a test that tests `Thread.sigmask`. The issue is that it loops until all pending signals are handled, which never happens if there are signals that are both pending and blocked.

The fix is to make `caml_check_pending_signals` return false when all pending signals are blocked: this function reports whether there are signal handlers to run, and in this case there are not. That way, `caml_process_pending_signals_exn` will result in caml_check_pending_signals becoming false. Before this patch it does not if there are signals that are pending yet blocked.


mshinwell:
Yeah I think this is correct, and it fixes the problem on my machine.

PR #3755, opened 2025-03-27
Remove example file from #3688

xclerc:
I think this is a leftover, that was
turned into a proper test case (in
#3688).


mshinwell:
Guilty

mshinwell (approved):

PR #3760, opened 2025-03-28
Use caml_array_blit for %arrayblit on Pgenarray and Paddrarray

mshinwell:
This allows the primitive to take advantage of various GC-related optimisations that are implemented in the runtime's version of blit.


mshinwell:
@ccasin would you mind looking at this?

PR #3744, opened 2025-03-26
github ci jobs with debug runtime for macos

gretay-js:
Add two new jobs on macos with debug runtime on. Hopefully it'll help us debug the recurring failure in `warnings.ml`.

PR #3761, opened 2025-03-28
Use caml_modify_local on runtime4 for atomics

mshinwell:
The C implementations of atomics on runtime4 (as used on arm64, and on x86-64 in cases where specialization has not happened) incorrectly call `caml_modify` rather than `caml_modify_local`.  This causes failures as seen on #3744.  The structure of the code on runtime5 looks correct already.

### DIFF
--- a/runtime4/misc.c
+++ b/runtime4/misc.c
@@ -256,7 +256,7 @@ CAMLprim value caml_atomic_compare_exchange(value ref, value oldv, value newv)
 {
   value* p = Op_val(ref);
   if (*p == oldv) {
-    caml_modify(p, newv);
+    caml_modify_local(ref, 0, newv);
     return oldv;
   } else {
     return *p;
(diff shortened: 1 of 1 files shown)


mshinwell:
I have checked manually on a macOS arm64 machine that the failing example in #3744 works correctly after this PR.

mshinwell:
@spiessimon it was suggested by @xclerc that maybe vicuna could spot this sort of error?

gretay-js (approved):

spiessimon:
Hmm, interesting. So `caml_modify_local` can be used on values of type `Atomic.t`, but `caml_modify` cannot? We would have to augment the externals extraction with information about atomics for that. I think they are currently treated either abstractly or as a block.

mshinwell:
The problem is that in `stdlib/atomic.mli` the primitives are declared with the atomics potentially being at local mode.

stedolan:
> Hmm, interesting. So `caml_modify_local` can be used on values of type `Atomic.t`, but `caml_modify` cannot? We would have to augment the externals extraction with information about atomics for that. I think they are currently treated either abstractly or as a block.

No, `caml_modify_local` cannot be used on values of type `Atomic.t`, same as `caml_modify`.

Runtime4 does not have multicore support and so doesn't need atomics, so runtime4's `Atomic.t` is secretly just a `ref` and can be accessed and modified as such. But this is not something that should be relied upon outside of the implementation of the `caml_atomic_foo` stubs.

PR #3769, opened 2025-03-31
Use the same verbose parameter for all register allocators

xclerc:
As per title; it is no longer useful to be
able to enable logging for only a usbset
of allocators.

### DIFF
--- a/backend/regalloc/regalloc_utils.ml
+++ b/backend/regalloc/regalloc_utils.ml
@@ -52,6 +52,8 @@ let validator_debug = bool_of_param "VALIDATOR_DEBUG"
 
 let block_temporaries = bool_of_param "BLOCK_TEMPORARIES"
 
+let verbose : bool Lazy.t = bool_of_param "VERBOSE"
+
 type liveness = Cfg_with_infos.liveness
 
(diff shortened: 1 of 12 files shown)


backend/regalloc/regalloc_utils.ml
```diff
@@ -52,6 +52,8 @@ let validator_debug = bool_of_param "VALIDATOR_DEBUG"
 
 let block_temporaries = bool_of_param "BLOCK_TEMPORARIES"
```
spiessimon, line 54: It might be a good idea to leave a comment how verbose is enabled. Looking at the code, I believe it is the following:

```suggestion

(* to enable pass -regalloc-param VERBOSE:true *)
```

spiessimon:
I looked over the changes, and they look good to me. In particular, the verbose logging is off by default.

backend/regalloc/regalloc_utils.ml
spiessimon: Thanks for updating the docs.

xclerc:
I have updated the documentation in `Notes.md`,
where the regalloc parameters are described.

PR #3770, opened 2025-03-31
Add a maximum depth to best_effort_compare_type_expr

glittershark:
Track how deep we've recursed in `best_effort_compare_type_expr`, and if we
recurse too deep, bail out to comparing by ID. I haven't looked deeply into
exactly /what/ is happening, but it seems like at least some of the time we're
recursing infinitely here on real-world code.

This ought to be deterministic as the structure of types is deterministic, but
if that ends up not being the case an alternative approach would be to /raise/
here, catch at the top level, and only then bail out to comparing by id


goldfirere (approved):
Would be cool to have this exercised by a test case, but if we're in a rush, that could come later.

PR #3772, opened 2025-03-31
Refactor `Reg` names & creation

TheNumbat:
- Register names are no longer mutable.
- Pre-assignment is now indicated by a separate immutable field.
- Creation functions have been centralized and have more descriptive names
- `all_hw_regs` is deleted (was unused)
- Functions on all-non-pre-assigned regs have more descriptive names 
- Selection and bind-let no longer attempt to coalesce unshared temps


TheNumbat:
With IRC, `ocamlopt.opt` actually contains fewer `mov` instructions after this patch, but LS is ~2.2% worse. This seems fine since IRC is the default now.

IRC: `1311649` -> `1307670`
LS: `1468857` -> `1501551`

backend/reg.mli
```diff
@@ -63,17 +64,21 @@ and stack_location =
...
+val createv_with_typs: t array -> t array
+val createv_with_typs_and_id: id:Ident.t -> t array -> t array
+
 val typv: t array -> Cmm.machtype
 val is_preassigned : t -> bool
 val is_unknown : t -> bool
-
-(* Name for printing *)
-val name : t -> string
+val print : t -> string
```
xclerc, line 81: (Arguably nitpicking, but `print`
seems a bit unexpected for an
accessor.)

backend/reg.ml
```diff
@@ -126,26 +127,24 @@ let is_reg t =
   | Reg _ -> true
   | Stack _ | Unknown -> false
 
-let reset() =
-  (* When reset() is called for the first time, the current stamp reflects
-     all hard pseudo-registers that have been allocated by Proc, so
-     remember it and use it as the base stamp for allocating
-     soft pseudo-registers *)
+let restart () =
```
xclerc, line 130: (We should maybe consider a
longer name, more in line with
the other renames. Typically
with the substring `relocatable`
in it.)

backend/reg.mli
```diff
@@ -15,16 +15,17 @@
...
+
+  val to_string : t -> string
 end
 
 type t =
-  { mutable raw_name: Raw_name.t;         (* Name *)
+  { name: Name.t;                         (* Name *)
     stamp: int;                           (* Unique stamp *)
     typ: Cmm.machtype_component;          (* Type of contents *)
+    preassigned: bool;                    (* Pinned to a specific location *)
```
xclerc, line 28: (I would maybe add something like
"(either hardware register or stack)",
as I have a feeling one might think
"preassigned" means "hardware
register".)

backend/reg.mli
```diff
@@ -63,17 +64,21 @@ and stack_location =
...
+val create_with_typ_and_name: ?prefix_if_var:string -> t -> t
+val create_at_location: Cmm.machtype_component -> location -> t
+
 val createv: Cmm.machtype -> t array
-val createv_like: t array -> t array
-val clone: t -> t
-val at_location: Cmm.machtype_component -> location -> t
+val createv_with_id: id:Ident.t -> Cmm.machtype -> t array
+val createv_with_typs: t array -> t array
+val createv_with_typs_and_id: id:Ident.t -> t array -> t array
```
xclerc, line 76: (I` wonder whether the `createv`
functions could use a longer
name, to avoid the `v`.)

xclerc (approved):

PR #3773, opened 2025-03-31
Configurable fiber stack sizes

TheNumbat:
- Adds `caml_init_fiber_stack_wsz`, which defaults to the pthread stack size
- Refactors the logic in `caml_get_init_stack_wsize` to choose the right value. The pthread stack size is now queried at startup instead of in systhreads. This seems fine since it already links/includes pthreads
- Cleans up the various ifdefs about guard pages to one `STACK_GUARD_PAGES` defintion
- Fixes `Stack_base` so that it correctly skips the info/guard pages when enabled
- Because `caml_callback` checks if it needs to reallocate the stack, `caml_maybe_resize_stack` with guard pages needs to raise stack overflow instead of aborting


TheNumbat:
Seems to have exposed a problem with stack reallocation in callbacks, looking into it...

runtime/fiber.c
```diff
@@ -613,12 +619,18 @@ void caml_maybe_expand_stack (void)
     (value*)stk->sp - Stack_base(stk);
   uintnat stack_needed =
     Stack_threshold / sizeof(value)
-    + 10 /* for words pushed by caml_start_program */;
-  /* XXX does this "8" need updating?  Provisionally changed to 10 */
+    /* for words pushed by caml_start_program */
+    + 8 + Stack_padding_word;
```
NickBarnes, line 623: In light of the deleted comment: is this `8` definitely right now? Can we make it a defined constant and check it somehow? Is there a matching comment in caml_start_program to keep things in sync? Are there any other places which care about this number?

NickBarnes (changes requested):
Basically fine. A few minor stylistic quibbles.

runtime/fiber.c
TheNumbat: `caml_start_program` only pushes 4 words (plus the padding word) on the OCaml stack, so 8 is more than sufficient unless there's some other constraint. @mshinwell do you know why it was changed to 10 at one point?

Technically `alloc_for_stack` also cares about this number as `wosize` must not be smaller than it, but that can't happen since the min stack size is either a page or 64 words. Agreed it'd be a lot better to keep this in sync automatically...

NickBarnes (approved):

PR #3775, opened 2025-04-01
Tweak logging code of register allocators

xclerc:
Based on https://github.com/ocaml-flambda/flambda-backend/pull/3769.

Removes the `indent:int` parameter passed
to every single logging function, and uses
`indent` / `dedent` functions to increment /
decrement the indentation.

(I have tried to be very careful, although
not aiming for a 1:1 change, but even if
everything is not exactly right, it is only
about logging code.)

### DIFF
--- a/backend/regalloc/regalloc_gi.ml
+++ b/backend/regalloc/regalloc_gi.ml
@@ -43,7 +43,10 @@ let rewrite : State.t -> Cfg_with_infos.t -> spilled_nodes:Reg.t list -> bool =
 
 let update_register_locations : State.t -> unit =
  fun state ->
-  if gi_debug then log ~indent:0 "update_register_locations";
+  if gi_debug
+  then (
+    log "update_register_locations";
+    indent ());
   let update_register (reg : Reg.t) : unit =
     match reg.Reg.loc with
     | Reg _ -> ()
@@ -56,11 +59,12 @@ let update_register_locations : State.t -> unit =
       | Some location ->
         if gi_debug
         then
-          log ~indent:1 "updating %a to %a" Printreg.reg reg
+          log "updating %a to %a" Printreg.reg reg
             Hardware_register.print_location location;
         reg.Reg.loc <- Hardware_register.reg_location_of_location location)
   in
-  List.iter (Reg.all_registers ()) ~f:update_register
+  List.iter (Reg.all_registers ()) ~f:update_register;
+  if gi_debug then dedent ()
 
 module Prio_queue = Make_max_priority_queue (Int)
 
@@ -74,7 +78,10 @@ let priority_heuristics : Reg.t -> Interval.t -> int =
 
 let make_hardware_registers_and_prio_queue (cfg_with_infos : Cfg_with_infos.t) :
     Hardware_registers.t * prio_queue =
-  if gi_debug then log ~indent:0 "creating registers and queue";
+  if gi_debug
+  then (
+    log "creating registers and queue";
+    indent ());
   let intervals = build_intervals cfg_with_infos in
   let hardware_registers = Hardware_registers.make () in
   let prio_queue =
@@ -87,8 +94,10 @@ let make_hardware_registers_and_prio_queue (cfg_with_infos : Cfg_with_infos.t) :
       | Reg _ -> (
         if gi_debug
         then (
-          log ~indent:1 "pre-assigned register %a" Printreg.reg reg;
-          log ~indent:2 "%a" Interval.print interval);
+          log "pre-assigned register %a" Printreg.reg reg;
+          indent ();
+          log "%a" Interval.print interval;
+          dedent ());
         match Hardware_registers.of_reg hardware_registers reg with
         | None -> ()
         | Some hardware_reg ->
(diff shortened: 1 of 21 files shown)


backend/regalloc/regalloc_split_state.ml
```diff
@@ -343,41 +374,51 @@ end = struct
...
+            then (
+              dedent ();
+              log "(ignored as a half-destruction point)");
             existing
           | None ->
-            if split_debug then log ~indent:2 "(not a destruction point)";
+            if split_debug
+            then (
+              dedent ();
+              log "(not a destruction point)");
```
spiessimon, line 407: I believe this is at the wrong indentation, but I'm not completely sure. At the beginning of the function in line 347, the indentation is increased to 2. I think if we dedent here, then we would be at 1 and not at 2 as before.

backend/regalloc/regalloc_gi.ml
```diff
@@ -132,37 +146,38 @@ let rec main : round:int -> flat:bool -> State.t -> Cfg_with_infos.t -> unit =
...
     incr step;
-    if gi_debug
-    then log ~indent:1 "step #%d (size=%d)" !step (Prio_queue.size prio_queue);
+    if gi_debug then log "step #%d (size=%d)" !step (Prio_queue.size prio_queue);
     let { Prio_queue.priority; data = reg, interval } =
       Prio_queue.get_and_remove prio_queue
     in
-    if gi_debug
-    then log ~indent:2 "got register %a (prio=%d)" Printreg.reg reg priority;
+    if gi_debug then log "got register %a (prio=%d)" Printreg.reg reg priority;
```
spiessimon, line 175: I think this is missing an indentation. It was 2 before, but I think it is currently one. Note that there is no indentation in between line 171 (which is at level 1) and 175 (which was at level 2).

backend/regalloc/regalloc_gi.ml
```diff
@@ -132,37 +146,38 @@ let rec main : round:int -> flat:bool -> State.t -> Cfg_with_infos.t -> unit =
...
     match Hardware_registers.find_available hardware_registers reg interval with
     | For_assignment { hardware_reg } ->
       if gi_debug
       then
-        log ~indent:3 "assigning %a to %a" Printreg.reg reg
+        log "assigning %a to %a" Printreg.reg reg
```
spiessimon, line 180: I think this is missing an indentation. It was 3 before, but I think it is currently one. Note that there is no indentation in between line 171 (which is at level 1) and 180 (which was at level 3).

backend/regalloc/regalloc_gi.ml
```diff
@@ -171,7 +186,7 @@ let rec main : round:int -> flat:bool -> State.t -> Cfg_with_infos.t -> unit =
     | For_eviction { hardware_reg; evicted_regs } ->
       if gi_debug
       then
-        log ~indent:3 "evicting %a from %a" Printreg.regs
+        log "evicting %a from %a" Printreg.regs
```
spiessimon, line 189: Same as above.

backend/regalloc/regalloc_gi.ml
```diff
@@ -205,23 +220,30 @@ let rec main : round:int -> flat:bool -> State.t -> Cfg_with_infos.t -> unit =
                          Reg.same r r')))
     | Split_or_spill ->
       (* CR xclerc for xclerc: we should actually try to split. *)
-      if gi_debug then log ~indent:3 "spilling %a" Printreg.reg reg;
+      if gi_debug then log "spilling %a" Printreg.reg reg;
```
spiessimon, line 223: Same as above.

backend/regalloc/regalloc_gi_utils.ml
```diff
@@ -689,8 +696,8 @@ module Hardware_registers = struct
                   let overlap = Interval.overlap interval itv in
                   if gi_debug
                   then
-                    log ~indent:5 "%a is assigned / overlap=%B" Printreg.reg
-                      pseudo_reg overlap;
+                    log "%a is assigned / overlap=%B" Printreg.reg pseudo_reg
```
spiessimon, line 699: I think this is at the same indentation as line 681. Before, it seems to have been 4 there vs 5 here.

backend/regalloc/regalloc_gi_utils.ml
```diff
@@ -715,8 +722,8 @@ module Hardware_registers = struct
                 if gi_debug
                 then
                   List.iter overlaping ~f:(fun assigned ->
-                      log ~indent:5 "evicting %a"
-                        Hardware_register.print_assigned assigned);
+                      log "evicting %a" Hardware_register.print_assigned
```
spiessimon, line 725: Same as above.

spiessimon:
I think getting rid of the absolute values of `ident` is a very good idea. There seem to be various functions that have hardcoded the indentation level to, for example, 3. 

I think the approach with `indent` and `dedent` works, but I'm a bit worried that it is very easy to forget a matching `dedent` or mess up the number of indents/detents. There were a couple of places where I got quite confused or had some difficulty following the indentation level (see my comments above). I wonder whether a slightly different interface would allow us to let the compiler do a lot of the well-bracketedness tracking. For example, I think the following two functions would cover a lot of the use of indent/dendent I have seen:

```ocaml
let with_log_indented (f: unit -> unit) = indent(); f(); dendent()
let with_log_conditionally_indented (b: bool) (f: unit -> unit) = if b then with_log_indented f else f ()
```

xclerc:
I definitely agree a `with_xyz` is a better approach.
The only reason I did not use it is because I wanted
to keep the `if debug` logic so that a non-debug
compiler does not pay any price for logging.
We have something similar for profiling though,
but I felt there are significantly more logging.

spiessimon:
Hmm, that's fair. I guess even the conditional version would pay the price of wrapping the rest of the code in a closure.

xclerc:
As discusssed on another channel, I was
wondering whether what I wrote above
might be wrong / too conservative.
@mshinwell would flambda2 inlining
clever enough here?

backend/regalloc/regalloc_split_state.ml
xclerc: Thanks, I have added another `dedent`
at the end of the function.

backend/regalloc/regalloc_gi.ml
xclerc: (I think it is fine to not indent here; we
indent after each new step, but the
messages related to a given step can
all be at the same level.)

PR #3776, opened 2025-04-01
Avoid uncaught `Not_found` arising from free vars in probes

lukemaurer:
We have a check that the free variables in a probe are all values. Unfortunately, this check runs _after_ translation to lambda, which means it considers the free variables of the lambda code. This is bad if the probe has a reference to a compilation unit that takes a parameter, since that will get translated to a local variable that doesn't appear in the environment and therefore we can't check its jkind.

Also, the error one gets in this situation is an uncaught `Not_found` with a backtrace, which is unpleasant and hard to debug.

Other improvements are possible, but for the moment we can just remember which lambda-level locals refer to runtime parameters, since those will always have value jkinds. Additionally, I've added a nicer error message for any other `Not_found`s that might arise from this check.

### DIFF
--- a/lambda/translcore.ml
+++ b/lambda/translcore.ml
@@ -30,6 +30,7 @@ type error =
     Free_super_var
   | Unreachable_reached
   | Bad_probe_layout of Ident.t
+  | Unknown_probe_layout of Ident.t
   | Illegal_void_record_field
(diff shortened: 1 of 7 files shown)


ccasin (approved):
LGTM

PR #3777, opened 2025-04-01
Simplify the logging logic (register allocators)

xclerc:
Based on https://github.com/ocaml-flambda/flambda-backend/pull/3775.

This pull request simplifies the register
allocator logging logic, only using a
single couple of value for `debug` /
`invariants` rather than one per allocator.

### DIFF
--- a/backend/regalloc/regalloc_gi.ml
+++ b/backend/regalloc/regalloc_gi.ml
@@ -8,13 +8,9 @@ module State = Regalloc_gi_state
 module Utils = struct
   include Regalloc_gi_utils
 
-  let debug = gi_debug
+  let debug = debug
 
-  let invariants = gi_invariants
-
-  let log = log
-
-  let log_body_and_terminator = log_body_and_terminator
+  let invariants = invariants
 
   let is_spilled reg = reg.Reg.spill
 
@@ -43,7 +39,7 @@ let rewrite : State.t -> Cfg_with_infos.t -> spilled_nodes:Reg.t list -> bool =
 
 let update_register_locations : State.t -> unit =
  fun state ->
-  if gi_debug
+  if debug
   then (
     log "update_register_locations";
     indent ());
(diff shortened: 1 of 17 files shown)


backend/regalloc/regalloc_gi.ml
```diff
@@ -8,13 +8,9 @@ module State = Regalloc_gi_state
 module Utils = struct
   include Regalloc_gi_utils
 
-  let debug = gi_debug
+  let debug = debug
```
spiessimon, line 11: Is there a reason we shadow `debug` and `invariants` here, but no longer do so for `log` and `log_body_and_terminator`? Perhaps because they come from `regalloc_utils.ml`?

backend/regalloc/regalloc_gi_utils.ml
```diff
@@ -4,24 +4,16 @@ open! Int_replace_polymorphic_compare
 open! Regalloc_utils
 module DLL = Flambda_backend_utils.Doubly_linked_list
 
-let gi_debug = true
```
spiessimon, line 7: It seems `gi_debug` was true by default, whereas we now use `debug` which is false in place of `gi_debug`.

spiessimon (changes requested):
This PR looks good to me, except for the switch from `true` to `false` (assuming that one was not intentional).

backend/regalloc/regalloc_gi.ml
xclerc: It was unintentional, but is in the spirit of
this chain of pull request: simply the
allocators and control their verbosity
through shared mechanisms.

backend/regalloc/regalloc_gi_utils.ml
xclerc: It was unintentional, but is in the spirit of
this chain of pull request: simply the
allocators and control their verbosity
through shared mechanisms.

spiessimon (approved):
If flipping the `gi_debug` verbosity is not a problem, then I think this PR is ready to go.

backend/regalloc/regalloc_gi.ml
xclerc: (Indeed, `debug` and `invariants` come from
`Regalloc_utils` with the other functions now
come from `Regalloc_gi_utils`.)

PR #3778, opened 2025-04-01
Reset logging at the start of register allocation

xclerc:
Based on https://github.com/ocaml-flambda/flambda-backend/pull/3777.

This is defensive, to be sure the indentation
level is not unbounded if there is a bug
in the refactoring of the pull requests in
this chain.


spiessimon (approved):
Looks good to me.

PR #3780, opened 2025-04-01
Refactor Reg stamps & machine regs

TheNumbat:
Every temp and every physical register has a unique stamp, but now physical registers aliased at different types share stamps.

I think this is not too scary, because I've hidden the stamp from the rest of the compiler:
- In `Reg`, `same/compare/Set/Map/Tbl` all have the same behavior as before the patch, because they now take into account both the type and the stamp
- I've replaced all direct usage of stamps with these functions, except one
- The one place that still uses stamps directly is the IRC interference adjacency graph, which needs to remember interference between aliased hardware registers

This is effectively just a more principled fix to #2573, since it doesn't require iterating the interference lists in `coalesce`. 
It would also be nice to remove the type-duplication entirely (e.g. in `Proc.destroyed_at`), but I don't see a way to do this without entirely reworking how types are tracked in the backend.

I've confirmed the compiler binary is identical before and after this patch using #3779, and `mixed-blocks/constructor_args.ml` still works properly without the fix from #2573, and the validator succeeds on the testsuite.


backend/regalloc/regalloc_irc.ml
```diff
@@ -62,7 +62,7 @@ let build : State.t -> Cfg_with_infos.t -> unit =
       if is_move_instruction instr
          && (not (Reg.is_stack instr.arg.(0)))
          && (not (Reg.is_stack instr.res.(0)))
-         && same_reg_class instr.arg.(0) instr.res.(0)
+         && Proc.types_are_compatible instr.arg.(0) instr.res.(0)
```
xclerc, line 65: (I think this is slightly more constrained,
which is fine - probably even better.)

backend/regalloc/regalloc_ls_state.ml
```diff
@@ -25,7 +25,7 @@ let[@inline] make ~stack_slots ~last_used =
   { interval_dll; active; stack_slots; instruction_id }
 
 type class_interval_array =
-  { elements : Interval.t array;
+  { mutable elements : Interval.t array;
```
xclerc, line 28: I am not against the changes in
this file (far from it), but out of
curiosity why did you have to
do this change?

backend/regalloc/regalloc_ls_state.ml
```diff
@@ -39,12 +39,15 @@ let dummy_interval =
   }
 
 let make_class_interval_array () =
-  (* CR-soon xclerc for xclerc: this will essentially use twice the required
-     memory *)
-  { elements = Array.make (Reg.total_registers ()) dummy_interval; length = 0 }
+  { elements = Array.make 32 dummy_interval; length = 0 }
```
xclerc, line 42: I think it is a welcome change, but
(as in other places) we should
probably leave a CR to consider
switching to `Dynarray` in due
course.

xclerc (approved):

backend/regalloc/regalloc_ls_state.ml
TheNumbat: The current stamp is no longer the total reg count, so I deleted `Reg.total_regs` and did this instead

PR #3781, opened 2025-04-02
Revert "Merge atomic counter inc/dec functions and use them consistently"

mshinwell:
Reverts ocaml-flambda/flambda-backend#3644

This caused trouble in two ways:
* some things looking for `caml/atomic_refcount.h` which has gone, this should be an easy fix
* it seems like some code gets `atomic_uintnat` via the `camlatomic.h` include in `misc.h`, which has now gone.

cc @NickBarnes

### DIFF
--- a/runtime/bigarray.c
+++ b/runtime/bigarray.c
@@ -27,7 +27,7 @@
 #include "caml/memory.h"
 #include "caml/mlvalues.h"
 #include "caml/signals.h"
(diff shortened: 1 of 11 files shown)


stedolan (approved):
Approved as a temporary workaround until someone has time to fix the build problems it caused, but I'd like to get this patch back in soon.

PR #3782, opened 2025-04-02
Remove objects from instruction selection code

mshinwell:
Removing the objects from `Selectgen` and associated modules produces a version of the instruction selection code which is easier to understand, and should indeed run faster.  In addition we can take full advantage of constructor disambiguation, which doesn't work in all cases with methods.

At first sight the open recursion in the old code made it look as if a simple functor-based approach, as in this PR, would be intractable.  However, upon closer inspection, it can be seen that the recursion that actually happens at runtime is quite restricted.  A small set of types to communicate values from the target-specific code back to the generic code can be defined instead, and used to eliminate the "backwards" recursive edges.

One question is whether the `sub_cfg` parameters in `cfg_selectgen.ml` could actually go in `env`.


xclerc:
> One question is whether the sub_cfg parameters in cfg_selectgen.ml could actually go in env.

As you said offline, it is probably better to keep
the two apart, since they flow differently.

mshinwell:
One note: the previous default implementation of `select_store` ignored the machtype and always used `Word_val`.  The new version respects `Word_int` (see the `Cstore` case).

backend/amd64/cfg_selection.ml
```diff
@@ -119,15 +119,15 @@ let one_arg name args =
    [effects_of], below. *)
 let inline_ops = ["sqrt"]
 
-let is_immediate n = n <= 0x7FFF_FFFF && n >= -0x8000_0000
+let is_immediate_int n = n <= 0x7FFF_FFFF && n >= -0x8000_0000
```
xclerc, line 122: (Not really important, but the `arm64`
uses `int_is_immediate`.)

backend/amd64/cfg_selection.ml
```diff
@@ -224,241 +224,226 @@ let pseudoregs_for_operation op arg res =
...
+  | Ctrywith (_, _, _, _, _, _, _) ->
+    Use_default
+
+let is_store_out_of_range _chunk ~byte_offset:_ :
+    Cfg_selectgen_target_intf.is_store_out_of_range_result =
+  Within_range
+
+let insert_move_extcall_arg _exttype _src _dst :
+    Cfg_selectgen_target_intf.insert_move_extcall_arg_result =
+  Use_default
```
xclerc, line 304: (I prefer this non-raising version, but
before you changed it I was wondering
whether the interface should have
optional functions. The version of this
pull request is probably less noisy.)

mshinwell:
We have compared the binary `ocamlopt` executables on both x86-64 and arm64, together with all the .S files involved for good measure, before and after this change.  They are identical.

xclerc (approved):
This looks equivalent to the previous version,
just easier to read... (Except maybe the
`Select_operation_then_rewrite` case, but
there are few occurrences.)

PR #3787, opened 2025-04-02
Clean up regalloc jobs (CI)

xclerc:
Now that CFG is the only pipeline, a couple
of CI jobs can be slightly simplified. This
pull request also renames the `cfg-selection`
job to `cfg-invariants`, since CFG selection
is the only alternative anyway...


mshinwell:
I've removed `cfg_selection` as a required check but not yet added `cfg_invariants` because it will cause trouble for people who haven't rebased recently.  We can add that in due course, @xclerc please make a note

PR #3795, opened 2025-04-03
Remove the current_sub_cfg instance variable

mshinwell:
This makes the `Cfg_selectgen` object empty, and means that the `Sub_cfg.t` values have to be passed around in the same way as in the version without objects.

Please read commit by commit.  There are some other small changes bundled in here to reduce CI load:
* Removal of the `tailrec_label` instance variable; this is constant during any pass of selection and can just be stored in the environment
* The `entry` and `exit` fields of `Sub_cfg` are now mutable, to make everything work nicely (otherwise new `Sub_cfg` values have to be created unnecessarily at certain points)
* The type parameter has been removed from `Select_utils.environment`.

### DIFF
--- a/backend/cfg/sub_cfg.mli
+++ b/backend/cfg/sub_cfg.mli
@@ -41,9 +41,9 @@ val exit_has_never_terminator : t -> bool
 
 val make_empty : unit -> t
 
-val add_empty_block_at_start : t -> label:Label.t -> t
+val add_empty_block_at_start : t -> label:Label.t -> unit
 
-val add_never_block : t -> label:Label.t -> t
+val add_never_block : t -> label:Label.t -> unit
 
 (** Use [add_instruction] in preference to this function. *)
 val add_instruction_at_start :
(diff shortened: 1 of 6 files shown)


backend/cfg/sub_cfg.mli
```diff
@@ -64,9 +64,9 @@ val iter_basic_blocks : t -> f:(Cfg.basic_block -> unit) -> unit
 
 val exists_basic_blocks : t -> f:(Cfg.basic_block -> bool) -> bool
 
-val join : from:t list -> to_:t -> t
+val join : from:t list -> to_:t -> unit
 
-val join_tail : from:t list -> to_:t -> t
+val join_tail : from:t list -> to_:t -> unit
```
xclerc, line 69: (Maybe change a label to make it
clear "which" is modified? That
could be `to_` ~> `into`?)

xclerc (approved):

backend/cfg/sub_cfg.mli
mshinwell: Noted, I will do this in the next PR to avoid yet another CI run

PR #3802, opened 2025-04-03
Cfg_selectgen: move code around

mshinwell:
No code changes, but this matches the order necessary when not using objects.  At present the order of the methods in the fine doesn't matter...

Based on #3801


backend/select_utils.ml
```diff
@@ -81,7 +88,7 @@ let env_find_regs_for_exception_extra_args id env =
   try Int.Map.find id env.regs_for_exception_extra_args
   with Not_found ->
     Misc.fatal_errorf
-      "Could not find exception label args registers for continuation %d" id
+      "Could not find exception extra args registers for continuation %d" id
 
 let _env_find_with_provenance id env = V.Map.find id env.vars
```
xclerc, line 93: (Not really about this pull request,
but maybe that could be deleted.)

xclerc (approved):
(Not in this pull request, but when the
dust settles, we should grep for
`Misc.fatal` and double check the error
messages do not refer to modules
which are no longer there.)

PR #3803, opened 2025-04-03
New version of Cfg_selectgen.emit_stores

mshinwell:
The aim here is to remove the arm64-specific version of this, as discussed with @gretay-js and @xclerc earlier this week.

Only the newest changeset here pertains to this PR, the rest are from other PRs.

### DIFF
--- a/backend/cfg_selectgen.ml
+++ b/backend/cfg_selectgen.ml
@@ -611,45 +611,104 @@ class virtual selector_generic =
          instruction for int32 arguments. *)
       self#insert_moves env sub_cfg src dst
 
-    method emit_stores env sub_cfg dbg data regs_addr =
-      let a =
+    method virtual select_store_new
+        : bool ->
+          Arch.addressing_mode ->
+          Cmm.expression ->
+          Cfg_selectgen_target_intf.select_store_result
+
+    method virtual is_store_out_of_range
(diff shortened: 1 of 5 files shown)


backend/cfg_selectgen.ml
```diff
@@ -861,90 +611,142 @@ class virtual selector_generic =
...
+          match operation_replacing_store with
+          | None ->
+            for i = 0 to Array.length regs - 1 do
+              let r = regs.(i) in
+              let chunk =
+                match r.Reg.typ with
+                | Float -> Double
+                | Float32 -> Single { reg = Float32 }
+                | Vec128 ->
+                  (* 128-bit memory operations are default unaligned. Aligned
```
xclerc, line 660:
```suggestion
                  (* 128-bit memory operations are by default unaligned. Aligned
```

xclerc (approved):
I find the logic of `emit_stores` a
bit complex, but think it has the
same behaviour as before.

mshinwell:
@gretay-js I've set this to merge when CI passes since it's had one review, the testsuite seems likely to catch any errors in the new function, and there is some time pressure.  However please review as well, thanks.

mshinwell:
Ah, except there is a build failure, so not quite ready.

PR #3804, opened 2025-04-03
Made asan compatible with gcc

jvanburen:


jvanburen:
it's not safe to get rid of preserve_all. We'd need gcc to save *all* the registers too.

PR #3807, opened 2025-04-04
 `value_or_null` atomics with correct result kind inference

dkalinichenko-js:
Correctly infer the results of `%atomic_load`, `%atomic_exchange` and `%atomic_compare_exchange` to be nullable immediates where previously they were assumed to be non-null (see https://github.com/ocaml-flambda/flambda-backend/pull/3731). This hopefully should be sufficient to still infer a tagged integer result in cases where the kind is known to be non-null.

Add `prove_is_immediate` to handle the case where we know that a value is either a tagged immediate or a null pointer, but don't know which. Use it in `simplify_atomic_compare_and_set_or_exchange_args`.

Annotate `Stdlib.Atomic` with `value_or_null`. Add tests.

### DIFF
--- a/middle_end/flambda2/simplify/simplify_ternary_primitive.ml
+++ b/middle_end/flambda2/simplify/simplify_ternary_primitive.ml
@@ -87,9 +87,9 @@ let simplify_atomic_compare_and_set_or_exchange_args
        also because the old value cannot be a pointer if the operation is to
        perform a write. *)
     let is_immediate ty =
-      match T.prove_is_a_tagged_immediate (DA.typing_env dacc) ty with
-      | Proved () -> true
-      | Unknown -> false
+      match T.prove_is_immediate (DA.typing_env dacc) ty with
+      | Proved true -> true
+      | Proved false | Unknown -> false
     in
     if is_immediate comparison_value_ty && is_immediate new_value_ty
     then Immediate
--- a/middle_end/flambda2/simplify/simplify_ternary_primitive.ml
+++ b/middle_end/flambda2/simplify/simplify_ternary_primitive.ml
@@ -139,7 +139,7 @@ let simplify_atomic_compare_exchange
   in
   let result_var_ty =
     match atomic_kind (* N.B. not [args_kind] *) with
-    | Immediate -> T.any_tagged_immediate
+    | Immediate -> T.any_tagged_immediate_or_null
     | Any_value -> T.any_value
   in
   let dacc = DA.add_variable dacc result_var result_var_ty in
--- a/middle_end/flambda2/simplify/simplify_ternary_primitive.ml
+++ b/middle_end/flambda2/simplify/simplify_ternary_primitive.ml
@@ -87,7 +87,7 @@ let simplify_atomic_compare_and_set_or_exchange_args
        also because the old value cannot be a pointer if the operation is to
        perform a write. *)
     let is_immediate ty =
-      match T.prove_is_immediate (DA.typing_env dacc) ty with
+      match T.prove_is_not_a_pointer (DA.typing_env dacc) ty with
       | Proved true -> true
       | Proved false | Unknown -> false
     in
--- a/middle_end/flambda2/types/flambda2_types.mli
+++ b/middle_end/flambda2/types/flambda2_types.mli
@@ -649,7 +649,7 @@ val prove_unique_fully_constructed_immutable_heap_block :
 val prove_is_int : Typing_env.t -> t -> bool proof_of_property
 
 (* Either a tagged integer or a null poitner. *)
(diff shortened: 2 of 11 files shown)


dkalinichenko-js:
Tested that examples

```
module Atomic' = struct
  open Stdlib.Atomic

  type nonrec 'a t = 'a t

  module Set_here_or_modified_elsewhere = struct
    type t =
      | Set_here
      | Modified_elsewhere
  end

  external get : ('a t[@local_opt]) -> 'a @ portable contended @@ portable = "%atomic_load"

  external compare_and_set_safe
    :  ('a t[@local_opt])
    -> 'a @ portable contended
    -> 'a @ portable contended
    -> bool
    @@ portable
    = "%atomic_cas"

  let compare_and_set t ~if_phys_equal_to ~replace_with =
    if compare_and_set_safe t if_phys_equal_to replace_with
    then Set_here_or_modified_elsewhere.Set_here
    else Modified_elsewhere
  ;;
end

let rec incr' (x : int Atomic'.t) =
  let if_phys_equal_to = Atomic'.get x in
  let replace_with = if_phys_equal_to + 1 in
  match Atomic'.compare_and_set x ~if_phys_equal_to ~replace_with with
  | Set_here -> ()
  | Modified_elsewhere -> incr' x
```

and `let cas_nones t = Atomic.compare_and_set t None None` still compile to processor instructions after this PR.

lthls:
I have looked at this PR. The code itself looks reasonable, but I'm worried about misunderstandings around terminology. We used to call an `immediate`, in Flambda2, an OCaml integer. It can be a tagged immediate or a naked immediate, depending on how it is represented, but with our model `Null` would not be an immediate.
But from what I see here, in the frontend `Immediate` seems to stand for anything that is a value but does not need to be scanned, including `Null`. I suspect that we will need to change naming conventions somewhere to avoid misunderstandings; Flambda2 probably needs to reason about both cases, so we will probably find a new, different name for OCaml integers and keep `Immediate` only for the cases where `Null` is also allowed, like the ones from this PR.
I will discuss a bit with the other Flambda2 devs about this and post the conclusions here.

dkalinichenko-js:
Yeah, makes sense.

middle_end/flambda2/types/flambda2_types.mli
```diff
@@ -646,6 +648,9 @@ val prove_unique_fully_constructed_immutable_heap_block :
 
 val prove_is_int : Typing_env.t -> t -> bool proof_of_property
 
+(* Either a tagged integer or a null poitner. *)
+val prove_is_immediate : Typing_env.t -> t -> bool proof_of_property
```
lthls, line 652: I don't think the comment is enough: changing this to `prove_is_not_a_pointer` would be a safe alternative until we sort out naming issues.

middle_end/flambda2/types/provers.ml
```diff
@@ -153,6 +153,25 @@ let prove_is_int env t =
 let meet_is_int_variant_only env t =
   gen_value_to_meet (prove_is_int_generic_value ~variant_only:true) env t
 
+let prove_is_immediate_generic_value env t =
```
lthls, line 156: See comment in flambda2_types.mli

middle_end/flambda2/types/provers.mli
```diff
@@ -123,6 +123,9 @@ val prove_unique_tag_and_size :
 
 val prove_is_int : Typing_env.t -> Type_grammar.t -> bool proof_of_property
 
+val prove_is_immediate :
```
lthls, line 126: See comment in flambda2_types.mli

middle_end/flambda2/simplify/simplify_ternary_primitive.ml
```diff
@@ -87,9 +87,9 @@ let simplify_atomic_compare_and_set_or_exchange_args
...
     let is_immediate ty =
-      match T.prove_is_a_tagged_immediate (DA.typing_env dacc) ty with
-      | Proved () -> true
-      | Unknown -> false
+      match T.prove_is_immediate (DA.typing_env dacc) ty with
+      | Proved true -> true
+      | Proved false | Unknown -> false
     in
     if is_immediate comparison_value_ty && is_immediate new_value_ty
     then Immediate
```
lthls, line 95: In this piece of code, renaming `T.prove_is_immediate` will make things a bit more confusing, but I think it's fine. Renaming the `is_immediate` function may improve things a bit, but changing the `Immediate` constructor should be left for later.

lthls (approved):
I'm approving the PR, because I think naming issues run a bit too deep to be fixed in this PR. I am suggesting a few changes that would make sense right now, feel free to follow or ignore these suggestions, as we will likely revisit this soon-ish anyway.

PR #3811, opened 2025-04-04
Specific instructions cannot raise

gretay-js:
Remove terminator `Cfg.Specific_can_raise` and `Arch.can_raise_operation`.

Currently arm64 and amd64 don't have specific instructions that can raise. Removing this terminator from CFG simplifies some passes.

### DIFF
--- a/backend/arm64/arch.ml
+++ b/backend/arm64/arch.ml
@@ -320,23 +320,6 @@ let operation_is_pure : specific_operation -> bool = function
 
 (* Specific operations that can raise *)
 
-let operation_can_raise = function
-  | Ifar_alloc _
-  | Ifar_poll -> true
-  | Imuladd
-  | Imulsub
-  | Inegmulf
-  | Imuladdf
-  | Inegmuladdf
-  | Imulsubf
-  | Inegmulsubf
-  | Isqrtf
-  | Imove32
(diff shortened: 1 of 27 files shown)


mshinwell:
I'll review this

backend/arm64/arch.ml
```diff
@@ -52,6 +52,7 @@ type cmm_label = Label.t
 
 type bswap_bitwidth = Sixteen | Thirtytwo | Sixtyfour
 
+(* Specific operations, include [Simd], must not raise. *)
```
mshinwell, line 55: "including" - same in the other file

backend/arm64/cfg_selection.ml
```diff
@@ -114,9 +111,8 @@ let select_addressing chunk (expr : Cmm.expression) :
   | arg -> Iindexed 0, arg
 
 let select_operation ~generic_select_condition:_ (op : Cmm.operation)
-    (args : Cmm.expression list) _dbg ~label_after :
+    (args : Cmm.expression list) _dbg ~label_after:_ :
```
mshinwell, line 114: Can this parameter be deleted?  If so let's do it

mshinwell: Agreed we will do this in a follow-on PR

PR #3812, opened 2025-04-04
Fix bug around persisting unsafe_mode_crossing

goldfirere:
First commit is the test; second commit is the fix.


glittershark (approved):
urgh...

we should just make this use `map_kind`; there are too many traversal functions for types and this is too easy to get wrong.

PR #3814, opened 2025-04-04
Infer with-kinds for GADTs, using `Tof_kind` for existentials

glittershark:
Infer full with-kinds for GADTs, using the new `Tof_kind` type to stand-in for existential type variables.

There's a big `Note` in `jkind.ml` explaining the approach here; see that for an overview.

Review all in one go, commit history is mostly not useful.

### DIFF
--- a/testsuite/tests/typing-jkind-bounds/gadt.ml
+++ b/testsuite/tests/typing-jkind-bounds/gadt.ml
@@ -61,12 +61,8 @@ type t = Foo : ('a : immutable_data). 'a -> t
 |}]
 
 let foo (t : t @ contended) = use_uncontended t
-(* CR layouts v2.8: fix this *)
 [%%expect {|
-Line 1, characters 46-47:
-1 | let foo (t : t @ contended) = use_uncontended t
-                                                  ^
-Error: This value is "contended" but expected to be "uncontended".
+val foo : t @ contended -> unit = <fun>
 |}]
 
 let foo (t : t @ local) = use_global t [@nontail]
@@ -85,21 +81,16 @@ type 'a t = Foo : 'a -> 'a t
 |}]
 
 let foo (t : int t @ once) = use_many t
-(* CR layouts v2.8: fix this *)
 [%%expect {|
-Line 1, characters 38-39:
-1 | let foo (t : int t @ once) = use_many t
-                                          ^
-Error: This value is "once" but expected to be "many".
+val foo : int t @ once -> unit = <fun>
 |}]
 
-let foo (t : t @ aliased) = use_unique t
+let foo (t : int t @ aliased) = use_unique t
 [%%expect {|
-Line 1, characters 13-14:
-1 | let foo (t : t @ aliased) = use_unique t
-                 ^
-Error: The type constructor "t" expects 1 argument(s),
-       but is here applied to 0 argument(s)
+Line 1, characters 43-44:
+1 | let foo (t : int t @ aliased) = use_unique t
+                                               ^
+Error: This value is "aliased" but expected to be "unique".
 |}]
 
 (***********************************************************************)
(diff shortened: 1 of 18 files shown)


goldfirere (commented):
I have done some light review and have concerns, sadly. I just don't think we're going to get away this easily. Might be worth checking in with Leo or Stephen to see if they have wisdom we do not.

typing/datarepr.ml
```diff
@@ -315,3 +315,32 @@ let unboxed_labels_of_type ty_path decl =
...
+        |> Seq.map Transient_expr.repr
+        |> TypeSet.of_seq
+      | _ -> Misc.fatal_error "cd_res must be Tconstr"
+    in
+    let arg_vars_set =
+      free_vars (newgenty (Ttuple (List.map (fun ty -> None, ty) tyl)))
+    in
+    TypeSet.diff arg_vars_set bound_vars
+
+let unbound_type_vars decl =
```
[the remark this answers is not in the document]
glittershark: I talked to @stedolan, and it sounds like we *do* have the invariant that type variables in different constructors are always distinct. So I think this is fine?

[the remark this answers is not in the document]
stedolan: As evidence for my assertion, we'd have a hard time typing this if the variables weren't disjoint:
```
# type 'a t = A : (int as 'a) -> 'a t | B : (string as 'a) -> 'a t;;
type _ t = A : int -> int t | B : string -> string t
```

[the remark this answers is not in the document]
goldfirere: Are we sure that the invariant holds even after, say, serializing and deserializing the type definition? Though maybe that doesn’t matter, because we’ll be doing this step only on a fresh declaration, in `update_decl_jkind`?

I’m not particularly convinced by Stephen’s example: I don’t see any type variables in constructors there. That is, by the time we’re done with an individual constructor, there won’t be a type variable in it. In any case, any example would be a proof that there exists a definition for which type variables are distinct, not a proof that all definitions have type variables that are distinct (but of course you know this).

However, I am encouraged by Stephen’s general assertion about this — long-term experience with the type checker is informative here. And maybe we don’t quite need the broad invariant I was angling for: we need it to be true only in the code path where we call these functions. If we can identify that code path carefully and observe that the invariant is true there, I’d be satisfied with a detailed comment.

typing/datarepr.mli
```diff
@@ -46,3 +46,6 @@ val constructor_existentials :
     - the types of the constructor's arguments
     - the existential variables introduced by the constructor
  *)
+
+val constructor_unbound_type_vars_excluding_row_variables
```
glittershark, line 50: comment

typing/jkind.ml
```diff
@@ -762,6 +765,21 @@ module With_bounds = struct
...
+
+let normalize' =
+  ref
+    (fun
+      ~mode:(_ : normalize_mode)
+      ~jkind_of_type:_
+      (_ : Types.jkind_l)
+      :
+      Types.jkind_l
+    -> assert false)
```
glittershark, line 781: probably not necessary anymore?

typing/jkind.ml
```diff
@@ -2278,32 +2296,134 @@ let for_boxed_variant cstrs =
...
-            (fun arg ->
-              add_with_bounds ~modality:arg.ca_modalities ~type_expr:arg.ca_type)
-            args jkind
-        | Cstr_record lbls -> add_labels_as_with_bounds lbls jkind
-      in
-      List.fold_right add_cstr_args cstrs base
+    let has_gadt_constructors =
+      List.exists (fun cstr -> Option.is_some cstr.cd_res) cstrs
+    in
+    let module Variant_parts = struct
```
glittershark, line 2307: add a big comment explaining what's going on

typing/jkind.ml
```diff
@@ -2278,32 +2283,207 @@ let for_boxed_variant cstrs =
...
+                List.fold_left2
+                  (fun (`Args ret_args, `Params params, seen) arg param ->
+                    if Btype.TypeSet.mem arg seen
+                    then
+                      (* We've already seen this type parameter, so don't add it again.
+                         See Step 2a from Note [With-bounds for GADTs] *)
+                      `Args ret_args, `Params params, seen
+                    else
+                      match Types.get_desc arg, Types.get_desc param with
+                      | Tvar _, Tvar _ ->
```
goldfirere, line 2444: The param technically doesn't have to be a var here.... but if it's not, a type error will be reported elsewhere. Test case:

```
type 'a t =
  K : 'b -> 'b t
  constraint 'a = int
```

Here, the arg will be a Tvar but the param will not be. I think it's just cleaner not to look at the param at all.

typing/jkind.ml
```diff
@@ -2278,32 +2283,207 @@ let for_boxed_variant cstrs =
...
+        { cstr_arg_tys = [];
+          cstr_arg_modalities = [];
+          params = [];
+          ret_args = [];
+          seen = Btype.TypeSet.empty
+        }
+    end in
+    let { Variant_parts.cstr_arg_tys; cstr_arg_modalities; ret_args; params; _ }
+        =
+      (* This left fold is step 1 from Note [With-bounds for GADTs] *)
```
goldfirere, line 2372: Now I see why the Note is the way it is: the algorithm builds up the information from all constructors into one big pile. I guess that works, according to the invariant Stephen has averred. But I see no need to do this (and doing it is wasteful, making for a harder unification problem in `apply`). Instead, just `apply` once per constructor (well, build up the arguments to `apply` once per constructor, and then actually apply those to each field). This also means that e.g. `seen` is maintained only within one constructor, instead of over the whole type.

typing/jkind.ml
```diff
@@ -2278,32 +2283,207 @@ let for_boxed_variant cstrs =
...
+    in
+    let cstr_arg_tys =
+      if Misc.Stdlib.List.is_empty params
+      then cstr_arg_tys
+      else
+        match
+          (* Step 2 from Note [With-bounds for GADTs] *)
+          type_apply ret_args
+            (Btype.newgenty
+               (Ttuple (List.map (fun ty -> None, ty) cstr_arg_tys)))
```
goldfirere, line 2476: I don't think it's necessary to build a tuple here. Why not just map `type_apply`? All of the free variables in `cstr_arg_tys` should be listed in `ret_args` and so there is no worrying about sharing. (except maybe row variables?)

typing/datarepr.ml
```diff
@@ -315,3 +317,29 @@ let unboxed_labels_of_type ty_path decl =
         labels rep decl.type_private
   | Type_record _
   | Type_variant _ | Type_abstract _ | Type_open -> []
+
+let constructor_unbound_type_vars_excluding_row_variables cstr =
```
goldfirere, line 321: Much of this computation duplicates work in `for_boxed_variant`. Maybe just inline this? That is, you traverse the result args to find all the bare Tvars, stored in a list. This list will be matched with the params in the call to `apply`. But it also can be turned into a set and subtracted from the free variables of the entire constructor. This approach has the nice property of making it even clearer that every variable free in a constructor's fields gets substituted away by the `apply`, either to a param or to a `Tof_kind`.

goldfirere: I think you'd probably need to pass `Ctype.free_vars` into `for_boxed_variant`. That's a small shame.... but certainly no worse than the forward-declaration of this function.

typing/datarepr.ml
```diff
@@ -315,3 +317,29 @@ let unboxed_labels_of_type ty_path decl =
...
+        |> List.to_seq
+        |> Seq.filter is_Tvar
+        |> Seq.map Transient_expr.repr
+        |> TypeSet.of_seq
+      | _ -> Misc.fatal_error "cd_res must be Tconstr"
+    in
+    let arg_vars_set =
+      free_vars
+        ~include_row_variables:false
+        (newgenty (Ttuple (List.map (fun ty -> None, ty) tyl)))
```
goldfirere, line 339: Tupling here is, I believe, more efficient than mapping and set-unioning. Might be nice to have a comment explaining why do this (because, to me, mapping and set-unioning is a more direct statement of what you're trying to do).

typing/ctype.ml
```diff
@@ -2271,7 +2271,12 @@ let type_jkind_purely_if_principal' =
    don't. *)
 let rec estimate_type_jkind ~expand_component env ty =
   match get_desc ty with
-  | Tvar { jkind } | Tof_kind jkind -> Jkind.disallow_right jkind
+  | Tvar { jkind } -> Jkind.disallow_right jkind
+  | Tof_kind jkind  ->
+    jkind
+    |> Jkind.disallow_right
+    (* [Tof_kind] stands in for an existential type, which always have best kinds *)
+    |> Jkind.mark_best
```
goldfirere, line 2279: Can't we arrange just to mark these best upon construction? I think that's better than doing it here.

typing/btype.mli
```diff
@@ -137,6 +138,9 @@ val fold_type_expr: ('a -> type_expr -> 'a) -> 'a -> type_expr -> 'a
 val iter_row: (type_expr -> unit) -> row_desc -> unit
         (* Iteration on types in a row *)
 val fold_row: ('a -> type_expr -> 'a) -> 'a -> row_desc -> 'a
+val iter_row_fields: (type_expr -> unit) -> row_desc -> unit
+        (* Iteration on fields in a row, excluding row variable *)
+val fold_row_fields: ('a -> type_expr -> 'a) -> 'a -> row_desc -> 'a
```
goldfirere, line 143: Don't need this (anymore)

typing/datarepr.ml
```diff
@@ -21,20 +21,22 @@ open Types
...
-          iter_row loop row;
-          if not (static_row row) then begin
-            match get_desc (row_more row) with
-            | Tvar _ when param -> ret := TypeSet.add ty !ret
-            | _ -> loop (row_more row)
-          end
+        if include_row_variables
+        then iter_row loop row
+        else iter_row_fields loop row;
+        if not (static_row row) && include_row_variables then begin
```
goldfirere, line 35: This looks wrong to me. I think that fields can be spread across many `row_more`s, in a little linked list, even for closed rows.

testsuite/tests/typing-jkind-bounds/gadt.ml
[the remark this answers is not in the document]
glittershark:
```ocaml
type 'a t =
  | K : 'b -> 'b option t
  constraint 'a = 'c option
[%%expect{|
Line 3, characters 13-27:
3 |   constraint 'a = 'c option
                 ^^^^^^^^^^^^^^
Warning 62 [constraint-on-gadt]: Type constraints do not apply to GADT cases of variant types.

type 'a t = K : 'b -> 'b option t constraint 'a = 'c option
|}]
```
and also
```ocaml
type 'c t2 =
  | K : 'd t -> 'd t2
  constraint 'c = 'e option
[%%expect{|
Line 3, characters 13-27:
3 |   constraint 'c = 'e option
                 ^^^^^^^^^^^^^^
Warning 62 [constraint-on-gadt]: Type constraints do not apply to GADT cases of variant types.

type 'a t2 = K : 'b option t -> 'b option t2 constraint 'a = 'e option
|}]
```

which is nice.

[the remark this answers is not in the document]
goldfirere: But... what are the kinds?

testsuite/tests/typing-jkind-bounds/gadt.ml
```diff
@@ -158,4 +149,374 @@ Uncaught exception: Misc.Fatal_error
...
+    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Error: The kind of type "box" is immutable_data with _
+         because it's a boxed variant type.
+       But the kind of type "box" must be a subkind of immediate
+         because of the annotation on the declaration of the type box.
+|}]
+
+(* Existential row variables *)
+
+type exist_row1 = Mk : ([< `A | `B of int ref] as 'a) -> exist_row1
```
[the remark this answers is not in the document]
glittershark: well, that was a good call, since
```
Error: The kind of type "exist_row1" is
         Uncaught exception: Typedecl.Error(_, _)

```

testsuite/tests/typing-jkind-bounds/gadt.ml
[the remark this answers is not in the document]
glittershark: seemingly the same as if the constraints weren't there. Do you still want a test for it?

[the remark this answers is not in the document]
goldfirere: I added one.

typing/jkind.ml
```diff
@@ -2259,7 +2264,7 @@ let for_unboxed_record lbls =
   in
   Builtin.product ~why:Unboxed_record tys_modalities layouts
 
-let for_boxed_variant cstrs =
+let for_boxed_variant ~decl_params ~type_apply cstrs =
   let open Types in
   if List.for_all
```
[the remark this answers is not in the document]
goldfirere: Added a CR.

testsuite/tests/typing-jkind-bounds/gadt.ml
```diff
@@ -577,16 +577,33 @@ type 'a exist_row3 =
...
+
+let foo (x : [`A | `B of int ref] option exist_row3 @ contended) = use_uncontended x
+[%%expect{|
+Line 1, characters 83-84:
+1 | let foo (x : [`A | `B of int ref] option exist_row3 @ contended) = use_uncontended x
+                                                                                       ^
+Error: This value is "contended" but expected to be "uncontended".
+|}]
+
+let foo (x : [`A | `B of int ref] option exist_row3 @ nonportable) = use_portable x
```
goldfirere, line 601: It would be nice if we could accept this, but that's beyond our expressiveness. I'll add a CR.

testsuite/tests/typing-jkind-bounds/gadt.ml
```diff
@@ -158,4 +149,374 @@ Uncaught exception: Misc.Fatal_error
...
+type 'a abstract : value mod portable
+Lines 2-3, characters 0-70:
+2 | type existential_abstract : immediate =
+3 |   | P : ('a : value mod portable). 'a abstract -> existential_abstract
+Error: The kind of type "existential_abstract" is immutable_data
+         with (type : value mod portable) abstract
+         because it's a boxed variant type.
+       But the kind of type "existential_abstract" must be a subkind of
+         immediate
+         because of the annotation on the declaration of the type existential_abstract.
```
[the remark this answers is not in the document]
goldfirere: I'm pursuing this outside of this PR.

goldfirere:
This is ready for final review. I've reviewed everything up to my 4 commits, which should garner someone else's attention. @glittershark ?

glittershark (commented):
Last four commits LGTM!

goldfirere (approved):

PR #3820, opened 2025-04-07
Unboxing in the reaper

Ekdohibs:
Not ready for merging yet. Extends the reaper to make possible unboxing of values cross-function, including unboxing the value slots of closures.


mshinwell:
For removing closure allocations in cases like:
```
let f x y = x + y

let[@inline never] foo f i = f i

let g z =
  let half = f z in
  foo (fun x -> (half [@inlined]) x) (z * 2)
```
(here, the closure for `half` will be removed, and `z` will instead end up in the closure of the lambda on the last line).

mshinwell (approved):
We will continue code review on the reaper itself after merging, the other parts have been reviewed.

PR #3821, opened 2025-04-07
Remove all objects from ocamltest

lukemaurer:
A simple refactor. Replaces all objects with first-class modules. Inheritance is mimicked using `include`. There is exactly once instance where that doesn't work, since a method gets overridden and one one of the remaining methods calls it, but all that's necessary is to “override” both “methods” (and they're not even in different files so it's easy).

The formatting is pretty wonky as I was hoping to keep a small diff in case changes are made upstream. (Obviously this can only help so much, especially given the repeated uses of `method`.)

### DIFF
--- a/ocamltest/ocaml_actions.ml
+++ b/ocamltest/ocaml_actions.ml
@@ -210,11 +210,12 @@ let cmas_need_dynamic_loading directories libraries =
   List.find_map loads_c_code (String.words libraries)
 
 let compile_program (compiler : Ocaml_compilers.compiler) log env =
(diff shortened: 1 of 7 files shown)


goldfirere:
Why do this? I worry that this will create merge conflicts with upstream.

mshinwell:
Following on from the recent experience in the backend this seems like a good idea to me, with the aim of increasing maintainability.  I am happy to discuss with upstream to try to get this merged there too, although I'm not worried about conflicts here.  I will review.

mshinwell (approved):

PR #3822, opened 2025-04-08
Tweak logging code of register allocators

xclerc:
This pull request is essentially a copy of #3775
(not sure what went wrong), where the discussion
will still happen in order to not lose the state
of the conversation.

### DIFF
--- a/backend/regalloc/regalloc_split_state.ml
+++ b/backend/regalloc/regalloc_split_state.ml
@@ -353,17 +353,18 @@ end = struct
             let destroyed : Reg.Set.t =
               Reg.Set.filter
                 (fun (reg : Reg.t) ->
-                  if split_debug
-                  then (
-                    log "register %a" Printreg.reg reg;
-                    indent ());
+                  if split_debug then log "register %a" Printreg.reg reg;
                   let keep =
                     match Reg.Tbl.find_opt num_sets reg with
                     | None | Some Maybe_more_than_once -> true
                     | Some At_most_once -> (
                       match Reg.Map.find_opt reg !already_spilled with
                       | None ->
-                        if split_debug then log "case/2";
+                        if split_debug
+                        then (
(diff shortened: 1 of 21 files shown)


backend/regalloc/regalloc_split_state.ml
```diff
@@ -408,11 +408,15 @@ end = struct
...
-        remove_dominated_spills doms child ~num_sets
-          ~already_spilled:!already_spilled ~destructions_at_end)
+    let res =
+      List.fold_left tree.children ~init:destructions_at_end
+        ~f:(fun destructions_at_end (child : Cfg_dominators.dominator_tree) ->
+          if split_debug then log "child %a" Label.format child.label;
+          remove_dominated_spills doms child ~num_sets
+            ~already_spilled:!already_spilled ~destructions_at_end)
+    in
+    if split_debug then dedent ();
```
spiessimon, line 418: I think that was the wrong direction. The function now has two indents and four dedents.

spiessimon (approved):
I think this looks good! Thank you for addressing my comments from the other PR. I did not carefully read over all of the changes here again, given the other PR (and in particular did not do indent/dedent counting).

I talked to @mshinwell, and flambda2 with 03 enabled (but perhaps already as is) should be able to optimize away the `with_` constructs that we discussed the other day. If we want to force O3, it can be enabled at the beginning of the file with `[@@@ocaml.flambda_o3]`. (But feel free to keep the indent/dedent if refactoring it would be too much of a hassle.)

backend/regalloc/regalloc_split_state.ml
xclerc: I have refactored that part to make
it clear it is will parenthesized.

xclerc:
(I have made a note to revisit the logging
during the next sprint.)

PR #1719, opened 2023-08-10
Emit all sections as before except for debug_line and debug_frames which are delayed

poechsel:
#1643 caused problem with mdx as it's involved a big changed in signatures.

This PR does the same but explicitly. All sections - but .debug_line, and later .debug_frames are emitted directly. .debug_line and .debug_frames are now considered as "delayed" and will only be emitted once [Dwarf.emit_delayed] is called.

PR #3826, opened 2025-04-08
Import original Asm_directives module

spiessimon:
This PR adds a new module to emit assembly directives based on [an old branch](https://github.com/mshinwell/ocaml/tree/gdb-names-gpr) of @mshinwell. The file itself should have no impact on the existing compiler, but the changes in the existing files could affect other parts of the compiler (e.g., if the they iterate over the list of available sections in order).

### DIFF
--- a/backend/asm_targets/asm_section.ml
+++ b/backend/asm_targets/asm_section.ml
@@ -42,6 +42,11 @@ type dwarf_section =
 
 type t =
   | DWARF of dwarf_section
+  | Data
+  | Read_only_data
+  | Eight_byte_literals
+  | Sixteen_byte_literals
+  | Jump_tables
   | Text
 
 type section_details =
@@ -70,8 +75,14 @@ let is_delayed = function
   | DWARF
       ( Debug_info | Debug_abbrev | Debug_aranges | Debug_str | Debug_loclists
       | Debug_rnglists | Debug_addr | Debug_loc | Debug_ranges )
-  | Text ->
-    false
+  (* CR sspies: Decide which of these are delayed. *)
+  | Data ->
+    Misc.fatal_error "Not yet implemented"
+  | Read_only_data -> Misc.fatal_error "Not yet implemented"
+  | Eight_byte_literals -> Misc.fatal_error "Not yet implemented"
+  | Sixteen_byte_literals -> Misc.fatal_error "Not yet implemented"
+  | Jump_tables -> Misc.fatal_error "Not yet implemented"
+  | Text -> false
 
 let details t ~first_occurrence =
   let names, flags, args =
@@ -118,6 +129,12 @@ let details t ~first_occurrence =
         | false, _ -> []
       in
       [name], flags, args
+    (* CR sspies: Determine what the details should be for these. *)
+    | Data, _ -> Misc.fatal_error "Not yet implemented"
+    | Read_only_data, _ -> Misc.fatal_error "Not yet implemented"
+    | Eight_byte_literals, _ -> Misc.fatal_error "Not yet implemented"
+    | Sixteen_byte_literals, _ -> Misc.fatal_error "Not yet implemented"
+    | Jump_tables, _ -> Misc.fatal_error "Not yet implemented"
     | Text, _ -> Misc.fatal_error "Not yet implemented"
   in
   { names; flags; args }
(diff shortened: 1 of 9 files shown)


backend/asm_targets/asm_section.ml
```diff
@@ -70,7 +75,10 @@ let is_delayed = function
...
       ( Debug_info | Debug_abbrev | Debug_aranges | Debug_str | Debug_loclists
       | Debug_rnglists | Debug_addr | Debug_loc | Debug_ranges )
-  | Text ->
+  (* CR sspies: Decide which of these are delayed. *)
+  (* FIXME: Except for [Text], these are just a guess. *)
+  | Data | Read_only_data | Eight_byte_literals | Sixteen_byte_literals
+  | Jump_tables | Text ->
     false
 
 let details t ~first_occurrence =
```
mshinwell, line 84: I'm not quite sure what's happened here, but this seems to be an older version of "flags"?  Maybe we can delete this one?

backend/asm_targets/asm_symbol.ml
```diff
@@ -26,30 +26,59 @@
...
-  type t = string
+  type t =
+    { name : string;
+      without_prefix : bool
+    }
 
-  let compare = String.compare
+  let compare { name = name1; without_prefix = without_prefix1 }
+      { name = name2; without_prefix = without_prefix2 } =
+    if String.compare name1 name2 = 0
```
mshinwell, line 53: Add a let-binding for the result of String.compare, to avoid repeating it.

backend/asm_targets/asm_symbol.ml
```diff
@@ -26,30 +26,59 @@
...
+    | OpenBSD | Generic_BSD | Solaris | BeOS | GNU | Dragonfly | Unknown ->
+      "" (* checked ok. *)
+    | MacOS_like -> "_" (* checked ok. *))
+  | ARM | POWER | Z | Riscv -> ""
+
 let should_be_escaped = function
   | 'A' .. 'Z' | 'a' .. 'z' | '0' .. '9' | '_' | '.' -> false
   | _c -> true
 
+(* CR sspies: This hack with the prefix is not great. *)
```
mshinwell, line 44: Seems ok to me?

utils/target_system.mli
```diff
@@ -41,3 +41,38 @@ type assembler =
...
+
+type machine_width =
+  | Thirty_two
+  | Sixty_four
+
+
+(** The natural machine width of the target system. *)
+val machine_width : unit -> machine_width
+
+
```
mshinwell, line 54: Remove double blank lines.  I would also leave a CR saying that we should cut down this list of targets, as some of them are obscure.  OK for now though.

utils/target_system.ml
```diff
@@ -120,3 +120,61 @@ let assembler () =
...
+  | Dragonfly -> Dragonfly
+  | GNU -> GNU
+  | BeOS -> BeOS
+  | Unknown -> Unknown
+
+let windows () =
+  match system () with
+  | Windows _ -> true
+  | _ -> false
+let is_macos () =
```
mshinwell, line 177: Poor formatting

mshinwell (changes requested):

backend/asm_targets/asm_section.ml
```diff
@@ -70,7 +75,10 @@ let is_delayed = function
   | DWARF
       ( Debug_info | Debug_abbrev | Debug_aranges | Debug_str | Debug_loclists
       | Debug_rnglists | Debug_addr | Debug_loc | Debug_ranges )
-  | Text ->
+  (* CR sspies: Decide which of these are delayed. *)
```
spiessimon, line 78: @mshinwell, is it correct that none of these sections are delayed?

backend/asm_targets/asm_section.ml
spiessimon: As discussed offline, I've merged `flags` and `details` now. In the dwarf cases, we use the new section flags that were introduced in PR #3078.

backend/asm_targets/asm_section.ml
spiessimon: This PR (#1719) suggests that this is correct and only `.debug_line` and `.debug_frame` are delayed.

PR #3827, opened 2025-04-08
Small cleanup changes in emit.ml

spiessimon:
This PR makes the following small changes to the ARM assembly emission:

- It removes an alias instruction, because `bcc` is the same as `b.cc`.
- It removes the comparison alias `lo`, because it is the same as `cc`.
- It adds hashes for immediates in several instructions that were missing them before.
- It removes the condition `al`, and clarifies why we do not implement `al` and `nv` (both mean always and should be replaced by a regular branch).

### DIFF
--- a/backend/arm64_ast.mli
+++ b/backend/arm64_ast.mli
@@ -81,10 +81,11 @@ module Instruction_name : sig
       | LT
       | GT
       | LE
-      (* The following are not supported, because NV means AL, but has a different encoding. *)
-      (* Use unconditional branching instead. *)
-      (* | AL *)
-      (* | NV *)
+    (* The following are not supported, because NV means AL, but has a different
+       encoding. *)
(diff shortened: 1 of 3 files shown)


backend/arm64_ast.mli
```diff
@@ -82,8 +82,7 @@ module Instruction_name : sig
       | GT
       | LE
       | AL
-      (* CR sspies: Remove [LO], since it is an alias of [CC] *)
-      | LO
+      | NV
```
gretay-js, line 85: We should not  emit `NV`, it's behavior is equivalent to `AL` but it has a different encoding. I prefer to not have it in this list (add a comment maybe to explain why it's not here?). An alternative is to assert that it's not printed.

gretay-js (approved):

gretay-js (approved):

PR #3828, opened 2025-04-08
Downstreams for renaming

voodoos:
1. Distinguish uids from interfaces and implementations: https://github.com/ocaml/ocaml/pull/13286
> Note: There should be a cleaner way to store the additional information. Upstream this info is part of a `Unit_info.t` record that is stored as the "current unit" in the environement. But here in flambda-backend it is `Compilation_unit.t` instead. This looks like conflicting refactorings that it would be useful to reconcile.
Here, for the prototype, we use a tuple to add that information.

2. Linked related declarations together: https://github.com/ocaml/ocaml/pull/13308


voodoos:
@lukemaurer what do you think would be the best way to handle the additionnal information in [b826fd7](https://github.com/ocaml-flambda/flambda-backend/pull/3828/commits/b826fd7e5430eef78ac1208f013721a6c9828903) ?

In the upstream compiler, we now store a `Unit_info.t` as the "current unit", but in flambda backend it is a `Compilation_unit.t` which is less obviously extendable.

mshinwell:
@lukemaurer I'd like to discuss the `Unit_info` part of this

lukemaurer:
Hmm. I think it would make more sense to do what we've been intending to do for a while and have `Unit_info.t` store a `Compilation_unit.t` rather than a `modname`, and then `Env.{get,set}_unit_name` can just do exactly what they do upstream. The only reason we didn't do that before is because it didn't make sense in the middle of a big merge.

(Ultimately I'm not sure that `Compilation_unit.t` is the right type here as opposed to `Compilation_unit.Name.t`, but that can be another PR.)

lukemaurer:
Discussed this with @mshinwell. I think putting a `Compilation_unit.t` in `Unit_info.t` makes the most sense. The alternative would be to combine `Compilation_unit` and `Unit_info`, but I don't think they're really combinable: a `Compilation_unit.t` is a (complicated) identifier and a `Unit_info.t` also has things like the source file and (in 5.3) the `kind` field. It makes a lot of sense for `Env.get_unit_name` to return a rich `Unit_info.t` and not much sense for an identifier to be aware of whether it's from the intf or the impl.

Concretely, what I propose is:
* `Unit_info.t` adds the `kind` field from upstream 5.3
* `Unit_info.t` stores a `Compilation_unit.t` instead of a `modname`. Awkwardly, it will need two exported constructors (or some fiddly refactoring in `compile_common.ml` and elsewhere):
```ocaml
(* the existing constructor, which will need the pack prefix to create a [Compilation_unit.t] *)
val make: ?check_modname:bool -> source_file:filename -> for_pack_prefix:Compilation_unit.Prefix.t -> intf_or_impl -> file_prefix -> t

(* a new constructor, which already knows the [Compilation_unit.t] *)
val of_compilation_unit: source_file:filename -> output_prefix:file_prefix -> kind:intf_or_impl -> Compilation_unit.t -> t
```
* `Env.{get,set}_unit_name` and the like operate on `Unit_info.t`, like upstream
* `Shape.Uid.mk` takes a `Unit_info.t option`, like upstream

I think this gets us to where things make good sense now and are not very different from upstream. There's still a bit of overlap in functionality between `Unit_info` and `Compilation_unit` but that can be cleaned up later.

typing/shape.ml
```diff
@@ -16,7 +16,10 @@
 module Uid = struct
   type t =
     | Compilation_unit of string
-    | Item of { comp_unit: string; id: int }
+    | Item of {
+        comp_unit: string;
+        id: int;
+        from: Unit_info.intf_or_impl }
```
lukemaurer: Should keep this on one line just to match upstream

lukemaurer (approved):
LGTM (with one formatting concern) as we're just concerned with differences from the upstream patches (most of which was smoothed over by the other PR).

typing/shape.ml
voodoos: Done. Thanks for the review !

PR #3829, opened 2025-04-09
Annotate `'a Domain.t` as `mod portable contended with 'a`

polytypic:
This crucially makes it possible to `Domain.join domain` from any other domain without having to unsafely coerce the `domain` to portable and uncontended.

### DIFF
--- a/stdlib/domain.mli
+++ b/stdlib/domain.mli
@@ -27,7 +27,7 @@
 
     @since 5.0 *)
 
-type !'a t
+type !'a t : value mod portable contended with 'a
 (** A domain of type ['a t] runs independently, eventually producing a
     result of type 'a, or an exception *)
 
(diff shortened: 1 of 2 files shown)


polytypic:
Do we need portable exceptions for this to be sound?

stdlib/domain.mli
```diff
@@ -27,7 +27,7 @@
 
     @since 5.0 *)
 
-type !'a t
+type !'a t : value mod portable contended with 'a
```
lpw25, line 30: Is this safe? It looks like it would let you call `join` from different capsules, exposing the `'a` into multiple capsules simultaneously even though it might be mutable.

polytypic: If `'a` is mutable then it shouldn't cross contention and then `'a t` also shouldn't cross contention and calling
```ocaml
val join : 'a t -> 'a @@ portable
```
from multiple capsules should be prevented.

lpw25: Sorry, I just missed the `with 'a` part. This is clearly fine.

lpw25 (approved):
This will be sound once we merge portable exceptions. As we are generally accepting some unsoundness until that is merged, we might as well go ahead and merge this now.

PR #3830, opened 2025-04-09
runtime events: disable flaky test

gretay-js:
The test for dropped runtime events has been failing intermittently on arm64 linux builds with runtime 5. It's already disabled on "native". This PR also disables it on bytecode, after a failure was [observed](https://github.com/ocaml-flambda/flambda-backend/actions/runs/14356072948/job/40245697583?pr=3813) on a different PR. There is a separate tracker to look into this issue.


NickBarnes (approved):
LGTM. I need to fix this bug but let's disable the test for now.

PR #3832, opened 2025-04-09
Fix a soundness bug in Simplif

stedolan:
`Alias` bindings, as generated by the pattern-match compiler, are pure computations that can be delayed by substituting the definition to its use site. However, they should not be substituted under lambdas, as this risks delaying a field projection past the region in which the projection is valid (if the value being projected from is local)

This used to be the case, but the recent refactoring of Lletrec introduced a bug: Lletrec now no longer has a Lfunction constructor underneath it (instead, using `Lambda.lfunction` directly), so the Simplif check to avoid substituting under lambda is defeated.

This is a soundness issue - the attached testcase hits an assertion failure in the GC under debug mode, but other variants of it can segfault.

The fix is straightforward: Lletrec should be treated as though it still contained an Lfunction.


lthls (approved):
The patch to Simplif looks correct. It is not completely obvious why the tests would fail without this patch, but I assume that both involve local allocations that end up cleaned up before calling a function, triggering assertions when accessed in the function.
(I'm wondering why we don't use `StrictOpt` instead of `Alias` for reads from stack-allocated values though)

PR #3834, opened 2025-04-09
Reintroduce support for OCAMLRUNPARAM=H=1 (force hugepages)

stedolan:
This adds support for passing the `MAP_HUGETLB` flag to `mmap` when expanding the heap, to use configured static huge pages (on Linux). This is a feature of OCaml 4 currently missing in OCaml 5.


NickBarnes (approved):
This looks OK. Semantics feel a bit different from runtime4, so I had to convince myself that this only affects the same memory (for example, on both runtime 4 and runtime 5, local allocation stacks aren't mmaped).

PR #3835, opened 2025-04-10
Remove IRC-related fields from `Reg.t`

xclerc:
As per title, IRC will use maps (in the state
to store the values).

### DIFF
--- a/backend/regalloc/regalloc_invariants.ml
+++ b/backend/regalloc/regalloc_invariants.ml
@@ -42,37 +42,16 @@ let precondition : Cfg_with_layout.t -> unit =
       : unit =
     ArrayLabels.iter regs ~f:(register_must_not_be_on_stack id)
   in
-  (* CR xclerc for xclerc: the check below should not be in this function, since
-     it is IRC-specific *)
-  let register_must_be_on_unknown_list (id : InstructionId.t) (reg : Reg.t) :
-      unit =
-    match reg.Reg.irc_work_list with
-    | Unknown_list -> ()
-    | Precolored -> ()
-    | Initial | Simplify | Freeze | Spill | Spilled | Coalesced | Colored
-    | Select_stack ->
-      fatal "instruction %a has a register (%a) already in a work list (%S)"
-        InstructionId.format id Printreg.reg reg
-        (Reg.string_of_irc_work_list reg.Reg.irc_work_list)
-  in
-  let register_must_be_on_unknown_list (id : InstructionId.t)
-      (regs : Reg.t array) : unit =
-    ArrayLabels.iter regs ~f:(register_must_be_on_unknown_list id)
-  in
   Cfg_with_layout.iter_instructions cfg_with_layout
     ~instruction:(fun instr ->
       let id = instr.id in
       desc_is_neither_spill_or_reload id instr.desc;
       registers_must_not_be_on_stack id instr.arg;
-      registers_must_not_be_on_stack id instr.res;
-      register_must_be_on_unknown_list id instr.arg;
-      register_must_be_on_unknown_list id instr.res)
+      registers_must_not_be_on_stack id instr.res)
     ~terminator:(fun term ->
       let id = term.id in
       registers_must_not_be_on_stack id term.arg;
-      registers_must_not_be_on_stack id term.res;
-      register_must_be_on_unknown_list id term.arg;
-      register_must_be_on_unknown_list id term.res);
+      registers_must_not_be_on_stack id term.res);
   let fun_num_stack_slots =
     (Cfg_with_layout.cfg cfg_with_layout).fun_num_stack_slots
   in
--- a/backend/regalloc/regalloc_irc.ml
+++ b/backend/regalloc/regalloc_irc.ml
@@ -394,7 +394,9 @@ module Utils = struct
   let invariants = invariants
 
   let is_spilled state reg =
-    WorkList.equal (State.work_list state reg) WorkList.Spilled
+    match State.work_list_opt state reg with
+    | None -> false
+    | Some work_list -> WorkList.equal work_list WorkList.Spilled
 
   let set_spilled _state reg = reg.Reg.spill <- true
 end
--- a/backend/regalloc/regalloc_irc.ml
+++ b/backend/regalloc/regalloc_irc.ml
@@ -395,7 +395,11 @@ module Utils = struct
 
   let is_spilled state reg =
     match State.work_list_opt state reg with
-    | None -> false
+    | None ->
+      (* Freshly-created may not have been added to the map yet; such registers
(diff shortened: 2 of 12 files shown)


xclerc:
(There is a bit of information duplication in
the IRC state, but that should be addressed
in a follow-up to keep this pull request easy
to review.)

backend/regalloc/regalloc_ls.ml
```diff
@@ -10,13 +10,15 @@ let snapshot_for_fatal = ref None
...
+
   let debug = debug
 
   let invariants = invariants
 
-  let is_spilled reg = reg.Reg.spill
+  let is_spilled _state reg = reg.Reg.spill
 
-  let set_spilled _reg = ()
+  let set_spilled _state _reg = ()
```
spiessimon, line 21: Where is `spill` actually set? (This is not a change of this PR, I'm just wondering.)

backend/regalloc/regalloc_irc_state.ml
```diff
@@ -49,33 +49,44 @@ type t =
...
+    reg_color : int option Reg.Tbl.t;
+    reg_alias : Reg.t option Reg.Tbl.t;
+    reg_interf : Reg.t list Reg.Tbl.t;
+    reg_degree : int Reg.Tbl.t
   }
 
 let max_capacity = 1024
 
 let[@inline] make ~initial ~stack_slots ~last_used () =
+  let num_registers = List.length initial in
```
spiessimon, line 63: Is this larger than the size of `Reg.all_registers ()`? If not, it could make sense to preallocate the table with the larger one. We iterate over all registers anyway below to initialize their respective entry with the initial value.

backend/regalloc/regalloc_irc_state.ml
```diff
@@ -49,33 +49,44 @@ type t =
     stack_slots : Regalloc_stack_slots.t;
     instruction_id : InstructionId.sequence;
     mutable inst_temporaries : Reg.Set.t;
-    mutable block_temporaries : Reg.Set.t
+    mutable block_temporaries : Reg.Set.t;
+    reg_work_list : WorkList.t Reg.Tbl.t;
```
spiessimon, line 53: Instead of 5 tables that each track one field, would it make sense to have one table that tracks each of the fields in a mutable/immutable record? Then the type checker would enforce that all tables track the same collection of registers.

backend/regalloc/regalloc_irc_state.ml
```diff
@@ -190,69 +210,83 @@ let[@inline] reset state ~new_inst_temporaries ~new_block_temporaries =
   RegisterStamp.PairSet.clear state.adj_set;
   Reg.Tbl.clear state.move_list
 
-let[@inline] is_precolored _state reg =
-  Reg.equal_irc_work_list reg.Reg.irc_work_list Reg.Precolored
+let[@inline] work_list state reg = Reg.Tbl.find state.reg_work_list reg
```
spiessimon, line 213: In the representation before, there was always a worklist when one had a register. I think this is still the case, because this function can only be called after `make` has been called, and registers are never deleted from the map (only reset back to the initial values). Is that correct?

spiessimon: Actually, after looking at the code for `is_spilled`, I no longer think this is the case. It might make sense to add a comment somewhere here explaining when a register is in these maps and when it is not. (At first glance, I thought registers are always in these maps after `make`.) A lot of the code below seems to rely on the registers being in the respective maps (except for `work_list_opt`), since otherwise an exception is thrown. I guess this is somewhat different from before, where merely having a register meant one could access the IRC fields.

spiessimon (approved):
I left a few comments below, but this PR looks good to me. It might be worth clarifying when a register is in the respective maps and when the functions that access the maps without checks can be called safely.

backend/regalloc/regalloc_irc.ml
```diff
@@ -377,23 +377,32 @@ let assign_colors : State.t -> Cfg_with_layout.t -> unit =
...
   include Regalloc_irc_utils
 
+  type state = State.t
+
   let debug = debug
 
   let invariants = invariants
 
-  let is_spilled reg = Reg.equal_irc_work_list reg.Reg.irc_work_list Reg.Spilled
+  let is_spilled state reg =
```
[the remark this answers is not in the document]
spiessimon: I just saw that this is changed in 3838. Never mind.

backend/regalloc/regalloc_invariants.ml
```diff
@@ -42,37 +42,16 @@ let precondition : Cfg_with_layout.t -> unit =
       : unit =
     ArrayLabels.iter regs ~f:(register_must_not_be_on_stack id)
   in
-  (* CR xclerc for xclerc: the check below should not be in this function, since
-     it is IRC-specific *)
-  let register_must_be_on_unknown_list (id : InstructionId.t) (reg : Reg.t) :
```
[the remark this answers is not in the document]
xclerc: I think it is indeed no longer needed:
it was used to be sure that when entering
the allocator all registers were on the
unknown list; this is now the case because
the map holding the information is created
when initializing the state of the allocator.

backend/regalloc/regalloc_irc_state.ml
xclerc: I thought it was easier to review this way,
but indeed we should consider doing that.
I would rather do that in a follow-up,
partially because some maps may happen
to be redundant.

gretay-js (approved):

backend/regalloc/regalloc_irc_state.ml
spiessimon: Sounds good :)

backend/regalloc/regalloc_invariants.ml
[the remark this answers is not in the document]
spiessimon: Thanks for clarifying!

backend/regalloc/regalloc_irc_state.ml
xclerc: (I have made a node to revisit.)

backend/regalloc/regalloc_irc_state.ml
xclerc: In practice, I think they will most of
the time have the same size, but
`initial` could be smaller in some
corner cases. (Changed as suggested.)

backend/regalloc/regalloc_irc_state.ml
xclerc: Registers are indeed never deleted from the map.
All registers should be in the maps, but I have
changed the code to have proper failures rather
the dreaded `Not_found` in case I am wrong.

xclerc:
(The failure is due to the infamous arm64 bug.)

PR #3836, opened 2025-04-10
Simplify the `Utils` module (register allocation)

xclerc:
This pull request simplifies the `Utils`
module used in register allocation by
tweaking the `rewrite_gen` function.
Instead of storing whether a register
is spilled or not, the function now uses
the passed `spilled_nodes` parameter.


gretay-js (commented):
looks good to me.

gretay-js (approved):

PR #3840, opened 2025-04-11
Generate and propagate never returns from emit_tuple

Gbury:
In some cases, particularly in dead code (which happens to be a bit more easily to produce with the upcoming match-in-match in some cases, but could happen anyway), `to_cmm` can produce code that makes Selection fails, because emit_tuple refuses to emit code that never returns (e.g. code that raises while evaluating the argument of another raise). This PR simply allows emit_tuple to return `Never_returns`, and the required change to propagate that value upwards until it can be handled.

This PR introduces and uses some monadic lets to make the diff much smaller (most notably, without them, there would need to be dozens of lines of code re-indented because of added `match .. with`). If accepted, a further PR could use these monadic lets at other existing places in the code for more coherence.

### DIFF
--- a/backend/cfg_selectgen.ml
+++ b/backend/cfg_selectgen.ml
@@ -28,6 +28,7 @@ module Or_never_returns = Select_utils.Or_never_returns
 module SU = Select_utils
 module V = Backend_var
 module VP = Backend_var.With_provenance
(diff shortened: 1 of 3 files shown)


mshinwell (approved):
Looks great

PR #3843, opened 2025-04-11
Comment how type_application works

goldfirere:
As tin.

I have two questions that could be addressed in review:

1. Is my summary correct?

2. Is this comment helpful?

For (2), I think it would have saved me at least an hour to have this comment written. I know @lpw25 prefers smaller comments on the individual pieces, but I pretty strongly prefer overview comments like this, where I can understand an overall flow of an algorithm -- usually individual pieces of code can speak for themselves in light of the overview.

Review: @lpw25 

Motivation: I needed to understand this all in order to port https://github.com/ocaml/ocaml/pull/285, and so I thought I'd write down what I learned.

### DIFF
--- a/typing/typecore.ml
+++ b/typing/typecore.ml
@@ -3710,6 +3710,7 @@ let list_labels env ty =
 
 (* Collecting arguments for function applications *)
 
+(* See also Note [Type-checking applications] *)
 type untyped_apply_arg =
   | Known_arg of
       { sarg : Parsetree.expression;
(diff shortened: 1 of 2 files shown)


lpw25:
I do indeed dislike comments like this. It is consists almost entirely of descriptions of the meanings of constructors in some types, and a series of steps each of which correspond to a specific function call within `type_application`. But it is 200 lines away from those type definitions and 5000 lines away from `type_application`. I think that makes it less useful and less likely to stay up-to-date than if those pieces of information were moved to where they are relevant.

lpw25:
(I do think adding comments for this purpose is good though. Although some light refactoring of the code would probably make some of the information unnecessary by making the code clearer.)

goldfirere:
I'm not quite sure how to proceed then. I'm confident that the existence of this comment would have saved me ~2 hours last week. For me, breaking this up into pieces diminishes its value -- it's precisely the forest that I'm trying to understand here, and breaking it into pieces would give me more trees. (Perhaps you and I are different in this way. I do not suggest that breaking it up would be worse for you, just for me.) There is always a risk of a comment like this going out of date; my solution to that would be more `(* See Note [Type-checking applications], Step 2 *)` comments.... but I was afraid you'd like that even less.

Comments like this are everywhere in GHC. They tend to be well-maintained and I can attest that they are invaluable. Newcomers to the codebase frequently adopt this style in their own projects (anecdotally). (Admittedly, there is social pressure against saying "I don't like this", but I don't think social pressure drives people to adopt the style in their own projects.)

Well structured code that avoids the needs for comments like this would definitely be an improvement. But it's really hard to get code that good! And probably not worth the time polishing it to a shine.

I'm curious what others would think about this. Thus paging @ccasin and @ncik-roberts. Do comments like this help you? How could it be better?

ccasin:
Seems like a clear improvement to me.  I've had to work out what this code is doing before and a comment to check my understanding would have been nice.

I share Leo's distaste for this kind of thing - it would be better if the comment were broken up and put in the places it is describing - but this seems obviously net positive to me in its current form.

ccasin (approved):

PR #3844, opened 2025-04-11
Some improvements to mode handling in type_omitted_parameters

goldfirere:
I need to reorder the calls to `type_omitted_parameters` and `type_apply_arg` in order to port https://github.com/ocaml/ocaml/pull/285, so I need to understand the former pretty well. In so doing, I spotted some opportunities for improvement.

Review: @riaqn

### DIFF
--- a/typing/typecore.ml
+++ b/typing/typecore.ml
@@ -568,14 +568,6 @@ let mode_tailcall_argument mode =
   { (mode_default mode) with
     locality_context = Some Tailcall_argument }
 
-let mode_partial_application expected_mode =
-  let expected_mode =
-    mode_morph (fun mode -> alloc_as_value (value_to_alloc_r2g mode))
-      expected_mode
-  in
-  { expected_mode with
-    locality_context = Some Partial_application }
-
 let mode_trywith expected_mode =
   { expected_mode with position = RNontail }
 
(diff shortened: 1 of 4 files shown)


riaqn (approved):
Thanks - I think this looks correct.

Some notes on what this is doing: Consider the example `f e0 _ e1 _ e2` where `_` denotes omitted parameters. This expands to:
```
let a = e0 in
let b = e1 in
let c = e2 in
let f' = f a in
fun x ->
let f'' = f' x b in
fun y ->
f'' y c
```
The inner closure closes over `c` and partially apply `f''` (it actually fully applies `f''`, but still contains a pointer to `f''` whose mode is described by `parital_apply`). The outer closure closes over `b` and `c`, and partially apply `f'`.

The deleted code `submode marg expected_mode` looks indeed redundant.

typing/typecore.ml
```diff
@@ -4050,21 +4045,10 @@ let type_omitted_parameters expected_mode env loc ty_ret mode_ret args =
...
-                 open_args
-             in
-             let closed_args = new_closed_args @ closed_args in
-             let open_args = [] in
-             let mode_closed_args = List.map Alloc.close_over closed_args in
              let mode_partial_fun = Alloc.partial_apply mode_fun in
              let mode_closure, _ =
                Alloc.newvar_above (Alloc.join
-                (mode_partial_fun:: mode_closed_args))
+                [ mode_partial_fun; closed_args_mode ])
```
riaqn, line 4051: I suspect it's clearer to use `Value` (instead of `Alloc`) throughout this function, since many of those modes are not allocation and thus not suitable for `Alloc`. Then, the following call to `register_allocation_mode` can be replaced by `register_allocation`.

lpw25:
I’m pretty sure it was done this way to get the better error location and hint. It seems a shame to lose that.

riaqn:
Yeah I think we can still pass in the `expected_mode`, while also fix the O(n^2) complexity.

goldfirere:
The new error messages still mention partial application, but I agree they’re not as precise as the old ones. The challenge here is that this is meant to be preparation for a patch that calls `type_omitted_parameters` *before* type-checking the `Arg`s. I think all of this can be worked out though. I will take a spin when I return to this, but that is unlikely to be for a few weeks, sadly.

typing/typecore.ml
riaqn: After second thought, I think the current code is fine - since we are in the same region anyway (since those psudo-functions don't have region).

goldfirere:
I don't expect to return here for a little while. Closing to reduce clutter. But this has been linked from internal ticket 5077, so it is not lost.

PR #3845, opened 2025-04-12
Infer with-kinds for easy polymorphic variants

glittershark:
In the case that a polymorphic variant is *static*, meaning it's closed and its row variable isn't bound, it's easy enough to infer a more precise kind for it than `value` - we just treat it like a regular (boxed) variant, adding all its mentioned types (given by `Btype.fold_row`) to the with-bounds. Since it's static, we can also mark it best

### DIFF
--- a/testsuite/tests/typing-jkind-bounds/row.ml
+++ b/testsuite/tests/typing-jkind-bounds/row.ml
@@ -137,3 +137,35 @@ Error: The kind of type "[< `A of 'a | `B of 'a * 'a | `C ]" is value
          immutable_data with 'a
          because of the definition of t at line 2, characters 2-83.
 |}]
+
+(* Tunivar-ified row variables *)
+
+type t1 = { f : ('a : value mod portable). ([> `Foo of int] as 'a) -> unit }
+(* CR layouts v2.8: This should be accepted *)
+[%%expect{|
+Line 1, characters 64-65:
+1 | type t1 = { f : ('a : value mod portable). ([> `Foo of int] as 'a) -> unit }
+                                                                    ^
+Error: This alias is bound to type "[> `Foo of int ]"
+       but is used as an instance of type "('a : value mod portable)"
+       The kind of [> `Foo of int ] is value
+         because it's a polymorphic variant type.
+       But the kind of [> `Foo of int ] must be a subkind of
+         value mod portable
+         because of the annotation on the universal variable 'a.
+|}]
+
+type t2 = { f : ('a : value mod portable). ([< `Foo of int] as 'a) -> unit }
+(* CR layouts v2.8: This should be accepted *)
+[%%expect{|
+Line 1, characters 64-65:
+1 | type t2 = { f : ('a : value mod portable). ([< `Foo of int] as 'a) -> unit }
+                                                                    ^
+Error: This alias is bound to type "[< `Foo of int ]"
+       but is used as an instance of type "('a : value mod portable)"
+       The kind of [< `Foo of int ] is value
+         because it's a polymorphic variant type.
+       But the kind of [< `Foo of int ] must be a subkind of
+         value mod portable
+         because of the annotation on the universal variable 'a.
+|}]
--- a/testsuite/tests/typing-jkind-bounds/row.ml
+++ b/testsuite/tests/typing-jkind-bounds/row.ml
@@ -1,0 +1,139 @@
+(* TEST
+    expect;
+*)
+
+let use_global : 'a @ global -> unit = fun _ -> ()
+let use_unique : 'a @ unique -> unit = fun _ -> ()
+let use_uncontended : 'a @ uncontended -> unit = fun _ -> ()
+let use_portable : 'a @ portable -> unit = fun _ -> ()
+let use_many : 'a @ many -> unit = fun _ -> ()
+[%%expect{|
+val use_global : 'a -> unit = <fun>
+val use_unique : 'a @ unique -> unit = <fun>
+val use_uncontended : 'a -> unit = <fun>
+val use_portable : 'a @ portable -> unit = <fun>
+val use_many : 'a -> unit = <fun>
+|}]
+
+(* Simple cases : closed polymorphic variants with fields *)
+
+type 'a t : immutable_data with 'a = [ `A of 'a ]
+
+[%%expect{|
+type 'a t = [ `A of 'a ]
+|}]
+
+type 'a u : mutable_data with 'a = [ `B of 'a ref ]
+[%%expect{|
+type 'a u = [ `B of 'a ref ]
+|}]
+
+type 'a v : value mod contended with 'a = [ `C of 'a | `D of 'a -> 'a | `E of 'a option ]
+[%%expect{|
+type 'a v = [ `C of 'a | `D of 'a -> 'a | `E of 'a option ]
+|}]
+
+type 'a w : value mod contended with 'a = [ 'a t | 'a v ]
+[%%expect{|
+type 'a w = [ `A of 'a | `C of 'a | `D of 'a -> 'a | `E of 'a option ]
+|}]
+
+let cross_contention (x : int t @ contended) = use_uncontended x
+let cross_portability (x : int t @ nonportable) = use_portable x
+let cross_linearity (x : int t @ once) = use_many x
+[%%expect{|
+val cross_contention : int t @ contended -> unit = <fun>
+val cross_portability : int t -> unit = <fun>
+val cross_linearity : int t @ once -> unit = <fun>
+|}]
+
+let don't_cross_unique (x : int t @ aliased) = use_unique x
+[%%expect{|
+Line 1, characters 58-59:
+1 | let don't_cross_unique (x : int t @ aliased) = use_unique x
+                                                              ^
+Error: This value is "aliased" but expected to be "unique".
+|}]
+
+let don't_cross_locality (x : int t @ local) = use_global x
+[%%expect{|
+Line 1, characters 58-59:
+1 | let don't_cross_locality (x : int t @ local) = use_global x
+                                                              ^
+Error: This value escapes its region.
+  Hint: This argument cannot be local,
+  because it is an argument in a tail call.
+|}]
+
+
+let cross_contention (x : int w @ contended) = use_uncontended x
+[%%expect{|
+val cross_contention : int w @ contended -> unit = <fun>
+|}]
+
+let don't_cross_portability (x : int w @ nonportable) = use_portable x
+[%%expect{|
+Line 1, characters 69-70:
+1 | let don't_cross_portability (x : int w @ nonportable) = use_portable x
+                                                                         ^
+Error: This value is "nonportable" but expected to be "portable".
+|}]
+
+(* Quality *)
+
+type 'a record : immutable_data with [ `A of 'a ] = { inner : 'a }
+[%%expect{|
+type 'a record = { inner : 'a; }
+|}]
+
+(* Harder cases: row variables *)
+
+(* CR layouts v2.8: These are both correct, but we could probably infer a more precise kind for both. *)
+type ('a, 'b) t : immediate = [< `X | `Y of 'a] as 'b
+[%%expect{|
+Line 1, characters 0-53:
+1 | type ('a, 'b) t : immediate = [< `X | `Y of 'a] as 'b
+    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Error: The kind of type "[< `X | `Y of 'a ]" is value
+         because it's a polymorphic variant type.
+       But the kind of type "[< `X | `Y of 'a ]" must be a subkind of immediate
+         because of the definition of t at line 1, characters 0-53.
+|}]
+type ('a, 'b) u : immediate = [> `X | `Y of 'a] as 'b
+[%%expect{|
+Line 1, characters 0-53:
+1 | type ('a, 'b) u : immediate = [> `X | `Y of 'a] as 'b
+    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Error: The kind of type "[> `X | `Y of 'a ]" is value
+         because it's a polymorphic variant type.
+       But the kind of type "[> `X | `Y of 'a ]" must be a subkind of immediate
+         because of the definition of u at line 1, characters 0-53.
+|}]
+
+(* less-than rows *)
+
+let f (x : [< `A of int | `B of string] @ contended) =
+  use_uncontended x
+(* CR layouts v2.8: This should probably be accepted *)
+[%%expect{|
+Line 2, characters 18-19:
+2 |   use_uncontended x
+                      ^
+Error: This value is "contended" but expected to be "uncontended".
+|}]
+
+module M : sig
+  type 'a t : immutable_data with 'a = private [< `A of 'a | `B of ('a * 'a) | `C ]
+end = struct
+  type 'a t = [ `C ]
+end
+[%%expect{|
+Line 2, characters 2-83:
+2 |   type 'a t : immutable_data with 'a = private [< `A of 'a | `B of ('a * 'a) | `C ]
+      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Error: The kind of type "[< `A of 'a | `B of 'a * 'a | `C ]" is value
+         because it's a polymorphic variant type.
+       But the kind of type "[< `A of 'a | `B of 'a * 'a | `C ]" must be a subkind of
+         immutable_data with 'a
+         because of the definition of t at line 2, characters 2-83.
+|}]
--- a/testsuite/tests/typing-jkind-bounds/row.ml
+++ b/testsuite/tests/typing-jkind-bounds/row.ml
@@ -86,6 +86,54 @@ type 'a record : immutable_data with [ `A of 'a ] = { inner : 'a }
 type 'a record = { inner : 'a; }
 |}]
 
+module type S = sig
+  type 'a polyvar = [ `A of 'a ]
+  type 'a record : immutable_data with 'a polyvar = { inner : 'a }
+end
+[%%expect{|
+module type S =
+  sig type 'a polyvar = [ `A of 'a ] type 'a record = { inner : 'a; } end
+|}]
+
+(* This is why we're able to give closed polymorphic variants best kinds *)
+module type S2 = S with type 'a polyvar = [ `B of int ref ]
+[%%expect{|
+Line 1, characters 17-59:
+1 | module type S2 = S with type 'a polyvar = [ `B of int ref ]
+                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Error: In this "with" constraint, the new definition of "polyvar"
+       does not match its original definition in the constrained signature:
+       Type declarations do not match:
+         type 'a polyvar = [ `B of int ref ]
+       is not included in
+         type 'a polyvar = [ `A of 'a ]
+       The type "[ `B of int ref ]" is not equal to the type "[ `A of 'a ]"
+       The second variant type does not allow tag(s) "`B"
+|}]
+
+(* When we give open polymorphic variants more precise kinds, we should make sure to give them not-best quality *)
+module type S = sig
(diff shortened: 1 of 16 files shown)


glittershark:
urk, stack overflow :(

testsuite/tests/typing-jkind-bounds/row.ml
```diff
@@ -0,0 +1,278 @@
...
+module type S = sig
+  type 'a polyvar = [ `A of 'a ]
+  type 'a record : immutable_data with 'a polyvar = { inner : 'a }
+end
+[%%expect{|
+module type S =
+  sig type 'a polyvar = [ `A of 'a ] type 'a record = { inner : 'a; } end
+|}]
+
+(* This is why we're able to give closed polymorphic variants best kinds *)
```
goldfirere, line 98: I don't think this comment is quite accurate: the `with` fails because the type is different, never mind the kinds. If you want to do something like this, make a private open polymorphic variant and then substitute it with a closed one.

testsuite/tests/typing-jkind-bounds/row.ml
```diff
@@ -0,0 +1,278 @@
...
+(* When we give open polymorphic variants more precise kinds, we should make sure to give them not-best quality *)
+module type S = sig
+  type 'a polyvar = private [< `A of 'a | `B of int ref ]
+  type 'a record : immutable_data with 'a polyvar = { inner : 'a }
+end
+module type S2 = S with type 'a polyvar = [ `A of 'a ]
+module F(M : S2) = struct
+  let cross (x : int M.record @ contended) = use_uncontended x
+end
+(* CR layouts v2.8: This should be accepted *)
```
goldfirere, line 123: And it is indeed accepted. Delete the comment?

testsuite/tests/typing-jkind-bounds/row.ml
```diff
@@ -0,0 +1,278 @@
...
+       does not match its original definition in the constrained signature:
+       Type declarations do not match:
+         type 'a polyvar = [ `B of int ref ]
+       is not included in
+         type 'a polyvar = [ `A of 'a ]
+       The type "[ `B of int ref ]" is not equal to the type "[ `A of 'a ]"
+       The second variant type does not allow tag(s) "`B"
+|}]
+
+(* When we give open polymorphic variants more precise kinds, we should make sure to give them not-best quality *)
```
goldfirere, line 114: Very true! I had not thought about this. I do not see a test for bestness of this example, though. (I actually hope it would be sound without effort, though. Probably will end up with an error message mentioning `#row` somewhere, which is regrettable, but good enough for a first release.)

testsuite/tests/typing-jkind-bounds/row.ml
```diff
@@ -0,0 +1,278 @@
...
+module type S2 =
+  sig type 'a polyvar = [ `A of 'a ] type 'a record = { inner : 'a; } end
+module F :
+  functor (M : S2) -> sig val cross : int M.record @ contended -> unit end
+|}]
+
+
+(* Harder cases: row variables *)
+
+(* CR layouts v2.8: These are both correct, but we could probably infer a more precise kind for both. *)
```
goldfirere, line 139: Not clear whether we should even try to do better here. I mean, we could infer better kinds for the parameters, but I think that would probably make the error message worse.

Instead, I would love to see tests that could plausibly be accepted -- basically just replace the `immediate` here with `immutable_data` or even `immutable_data with 'a`.

testsuite/tests/typing-jkind-bounds/row.ml
```diff
@@ -0,0 +1,278 @@
...
+         because it's a polymorphic variant type.
+       But the kind of type "[> `X | `Y of 'a ]" must be a subkind of immediate
+         because of the definition of u at line 1, characters 0-53.
+|}]
+
+(* less-than rows *)
+
+let f (x : [< `A of int | `B of string] @ contended) =
+  use_uncontended x
+(* CR layouts v2.8: This should probably be accepted *)
```
goldfirere, line 165:
```suggestion
(* CR layouts v2.8: This should be accepted *)
```

But not in a first version.

testsuite/tests/typing-jkind-bounds/row.ml
```diff
@@ -0,0 +1,278 @@
...
+[%%expect{|
+Line 2, characters 18-19:
+2 |   use_uncontended x
+                      ^
+Error: This value is "contended" but expected to be "uncontended".
+|}]
+
+module M : sig
```
goldfirere, line 173: This one should also be accepted. But with a not-best kind.

testsuite/tests/typing-jkind-bounds/row.ml
```diff
@@ -0,0 +1,278 @@
...
+         because it's a polymorphic variant type.
+       But the kind of type "[< `A of 'a | `B of 'a * 'a | `C ]" must be a subkind of
+         immutable_data with 'a
+         because of the definition of t at line 2, characters 2-83.
+|}]
+
+(* Tunivar-ified row variables *)
+
+type t1 = { f : ('a : value mod portable). ([> `Foo of int] as 'a) -> unit }
+(* CR layouts v2.8: This should be accepted *)
```
goldfirere, line 192: No, I don't think so. Not until we have kind-restricted row extensions, which we have not planned out.

On the other hand, it would be much more plausible to accept this with a `<` instead of a `>`.

testsuite/tests/typing-jkind-bounds/row.ml
```diff
@@ -0,0 +1,278 @@
...
+|}]
+
+type trec_succeeds : value mod immutable = [ `C | `D of 'a * unit -> 'a ] as 'a
+
+[%%expect{|
+type trec_succeeds = [ `C | `D of 'a * unit -> 'a ] as 'a
+|}]
+
+type trec_rec_fails : immutable_data =
+  [ `X of 'b | `Y of [ `Z of ('a -> 'b) | `W of 'a | `Loop of 'b ] as 'b ] as 'a
```
goldfirere, line 253: This type makes me cry.

testsuite/tests/typing-jkind-bounds/subsumption/quality.ml
```diff
@@ -240,33 +237,11 @@ end = struct
...
-           constraint 'a = [< `a of string | `b ]
-       The kind of the first is value
-         because of the definition of t at line 6, characters 2-49.
-       But the kind of the first must be a subkind of value mod global
-         with [< `a of string | `b ] u
-         because of the definition of t at line 3, characters 2-40.
+module M :
+  sig
+    type 'a u = 'a constraint 'a = [< `a of string | `b ]
+    type 'a t constraint 'a = [< `a of string | `b ]
```
goldfirere, line 243: This one is ok.... but I'm worried we have a bestness problem here. Set up the same test with portability and replace the `string` with a function. I'd want to make sure that we just expand `'a u` to `'a` and not make any assumptions about `u`'s kind.

testsuite/tests/typing-modules/nondep_private_abbrev.ml
```diff
@@ -80,20 +80,20 @@ module IndirectPub : sig type t = [ `Foo of 'a ] as 'a end
...
-module IndirectPriv : sig type t end
+module IndirectPriv : sig type t : immutable_data end
 |}]
 
 (* These two behave as though a functor was defined *)
 module DirectPrivEta =
   (functor (X : sig end) -> Priv(X))(struct end);;
 [%%expect{|
-module DirectPrivEta : sig type t end
+module DirectPrivEta : sig type t : immutable_data with t end
```
goldfirere, line 90: That is not a very helpful kind. It seems sound, but we should leave a CR to examine the interaction between kinds and nondep. Urgh.

testsuite/tests/typing-modules/nondep_private_abbrev.ml
```diff
@@ -149,5 +149,5 @@ module I : functor (X : sig end) -> sig type t = Priv(X).t end
 
 module IndirectPriv = I(struct end);;
 [%%expect{|
-module IndirectPriv : sig type t end
+module IndirectPriv : sig type t : value mod immutable with int end
```
goldfirere, line 152: Yikes. `with int` is ugly. Definitely leave a CR for this one.

typing/ctype.ml
```diff
@@ -2319,7 +2319,7 @@ let rec estimate_type_jkind ~expand_component env ty =
   | Tlink _ | Tsubst _ -> assert false
   | Tvariant row ->
      if tvariant_not_immediate row
```
goldfirere, line 2321: Seems odd to have this check outside of `for_boxed_row`.

typing/jkind.ml
```diff
@@ -957,8 +960,18 @@ module Layout_and_axes = struct
                 Continue
                   { t with constr = Path.Map.add p (fuel - 1, args) constr }
               else Stop { t with fuel_status = Ran_out_of_fuel })
+          | Tvariant row -> (
+            let id = get_id (row_more row) in
```
goldfirere, line 964: I think it is more customary to use `Btype.proxy` here instead of `row_more`.

typing/jkind.ml
```diff
@@ -2312,6 +2325,19 @@ let for_boxed_tuple elts =
     elts
     (Builtin.immutable_data ~why:Tuple |> mark_best)
 
+let for_boxed_row row =
+  (let base = Builtin.immutable_data ~why:Polymorphic_variant in
+   if not (Btype.static_row row)
+   then
+     (* CR layouts v2.8: We can probably do a fair bit better here in most cases *)
+     Builtin.value ~why:Polymorphic_variant
```
goldfirere, line 2333: I don't think this case is `best`.

goldfirere: As I've thought about this (and as my understanding of `row_more` has [increased](https://github.com/ocaml/ocaml/pull/14023)), I think we might be able to make this cleaner: unconditionally add the `row_more` to the with-bounds. This is wild and crazy. As I like to say, `row_more` is not a type; it's a row. But I think putting it in the with-bounds has just the right behavior:
* For static rows, the `row_more` is `Tnil`. Assuming we give `Tnil` a bottom best kind (the layout doesn't matter), this works.
* For non-static rows, the `row_more` will be either a variable or an abstract type, none of which are best. Good! But when they get substituted, they might become best -- also good.

The challenge in this approach will be printing. Maybe we can look at existing handling of printing for `row_more` for inspiration.

But before making any changes here, we should write some tests. Here are some good ones:

```
type 'a t1 = [< `A of string | `B of int ] as 'a
type 'a t2 : immutable_data with 'a t1 = C of string  (* should be rejected, at least until we sort out closed-but-not-static bestness *)
type t3 : immutable_data with [ `A of string] t1 = C of string  (* should be accepted *)

type 'a t1 = [> `A of string | `B of int ] as 'a
type 'a t2 : immutable_data with 'a t1 = C of string  (* should be rejected *)
type t3 : immutable_data with [ `A of string | `B of int | `C ] t1 = C of string  (* should be accepted *)

module type S = sig
  type t = private [< `A of string | `B of int ]
end
module M1 : S = struct
  type t = [ `A of string ]
end
type t2 : immutable_data with M1.t = C of string  (* should be rejected, at least until we sort out closed-but-not-static bestness *)
module M2 : S with type t = [ `A of string ] = struct
  type t = [ `A of string ]
end
type t3 : immutable_data with M2.t = C of string (* should be accepted *)

type (_, _) eq = Refl : ('a, 'a) eq

(* I'm not sure whether module substitution over a non-static private row type preserves the Tvariant structure; so I made this harder case, too *)
let sneaky (x : (M1.t, [ `A of string ]) eq) = match x with
  | Refl -> let open struct
    type t4 : immutable_data with M1.t = C of string  (* not sure what will happen, but we should eventually accept *)
  end in ()

module type S = sig
  type t = private [> `A of string | `B of int ]
end
module M1 : S = struct
  type t = [ `A of string | `B of int | `C of (int -> int) ref ]
end
type t2 : immutable_data with M1.t = C of string  (* should be rejected *)
module M2 : S with type t = [ `A of string | `B of int ] = struct
  type t = [ `A of string | `B of int ]
end
type t3 : immutable_data with M2.t = C of string (* should be accepted *)

let sneaky (x : (M1.t, [ `A of string | `B of int ]) eq) = match x with
  | Refl -> let open struct
    type t4 : immutable_data with M1.t = C of string  (* not sure what will happen, but we should eventually accept *)
  end in ()
```

testsuite/tests/typing-jkind-bounds/row.ml
dkalinichenko-js: It's a good testcase then :)

goldfirere: Yes, it is.

testsuite/tests/typing-jkind-bounds/row.ml
```diff
@@ -0,0 +1,219 @@
...
+                                                              ^
+Error: This value is "aliased" but expected to be "unique".
+|}]
+
+let don't_cross_locality (x : int t @ local) = use_global x
+[%%expect{|
+Line 1, characters 58-59:
+1 | let don't_cross_locality (x : int t @ local) = use_global x
+                                                              ^
+Error: This value escapes its region.
```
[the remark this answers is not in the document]
dkalinichenko-js: Done.

testsuite/tests/typing-jkind-bounds/row.ml
dkalinichenko-js: I edited the tests, does it do the right thing now?

testsuite/tests/typing-jkind-bounds/row.ml
dkalinichenko-js: Added CRs that hopefully get the message across.

testsuite/tests/typing-jkind-bounds/row.ml
dkalinichenko-js: Deleted the test since I don't quite see the point.

testsuite/tests/typing-jkind-bounds/row.ml
dkalinichenko-js: Done.

testsuite/tests/typing-jkind-bounds/row.ml
dkalinichenko-js: Added a CR.

testsuite/tests/typing-jkind-bounds/row.ml
dkalinichenko-js: Commented.

testsuite/tests/typing-jkind-bounds/subsumption/quality.ml
dkalinichenko-js: Added a test, but not sure that I got it right.

testsuite/tests/typing-modules/nondep_private_abbrev.ml
dkalinichenko-js: Done.

testsuite/tests/typing-modules/nondep_private_abbrev.ml
dkalinichenko-js: Done.

typing/ctype.ml
dkalinichenko-js: Moved `tvariant_not_immediate` into `Btype` and the check into `Jkind.for_boxed_row`. `tvariant_not_immediate` could also live in `Jkind`, but for now I feel like it fits more into `Btype`. We can move it into `Jkind` when we start considering `void` arguments.

typing/jkind.ml
dkalinichenko-js: Done.

typing/jkind.ml
dkalinichenko-js: I added the tests, but I don't quite understand why any of those should be rejected. All are accepted now.

dkalinichenko-js (commented):
Had to rebase for this one, sorry.

dkalinichenko-js:
Sorry! 5e114cdbb4387a8dc4563f2248e61ac4a0643609 to b0753ec5a688f680f268d2b5b3726be170c78bf5 (latest) are new.

testsuite/tests/typing-jkind-bounds/row.ml
```diff
@@ -291,3 +291,81 @@ type trec_rec_succeeds =
     [ `X of [ `Loop of 'b | `W of 'a | `Z of 'a -> 'b ] as 'b | `Y of 'b ]
     as 'a
 |}]
+
+(* Future tests for when we start adding row variables to with-bounds. *)
+
+type 'a t1 = [< `A of string | `B of int ] as 'a
+type 'a t2 : immutable_data with 'a t1 = C of string  (* should be rejected, at least until we sort out closed-but-not-static bestness *)
```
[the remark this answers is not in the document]
goldfirere: Oh, this one's OK. `'a t1` doesn't mode cross at all, and so it supplies an upper bound of "top", which is accepted. But good to update the comment.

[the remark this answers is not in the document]
goldfirere: Oh wait, no it's not OK. Because a non-static polymorphic variant should have a _best_ kind. So this seems no good.

testsuite/tests/typing-jkind-bounds/row.ml
```diff
@@ -291,3 +291,81 @@ type trec_rec_succeeds =
...
+type t3 : immutable_data with [ `A of string] t1 = C of string  (* should be accepted *)
+[%%expect{|
+type 'a t1 = 'a constraint 'a = [< `A of string | `B of int ]
+type 'a t2 = C of string constraint 'a = [< `A of string | `B of int ]
+type t3 = C of string
+|}]
+
+type 'a t1 = [> `A of string | `B of int ] as 'a
+type 'a t2 : immutable_data with 'a t1 = C of string  (* should be rejected *)
```
goldfirere, line 307: But it's not rejected. Eek!

testsuite/tests/typing-jkind-bounds/row.ml
```diff
@@ -291,3 +291,81 @@ type trec_rec_succeeds =
...
+val sneaky : (M1.t, [ `A of string ]) eq -> unit = <fun>
+|}]
+
+module type S = sig
+  type t = private [> `A of string | `B of int ]
+end
+module M1 : S = struct
+  type t = [ `A of string | `B of int | `C of (int -> int) ref ]
+end
+type t2 : immutable_data with M1.t = C of string  (* should be rejected *)
```
goldfirere, line 352: Agreed -- but it's not!

testsuite/tests/typing-jkind-bounds/row.ml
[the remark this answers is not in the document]
dkalinichenko-js: I think all of those tests succeed because the typechecker sees through the `= C of string` and normalize `'a t2` to `immutable_data` without any bounds. Then, all checks trivially succeed. If you want to test quality here, I think you'll need to do something more involved.

testsuite/tests/typing-jkind-bounds/subsumption/quality.ml
```diff
@@ -244,6 +244,21 @@ module M :
...
+  type 'a t : value mod portable with 'a u
+end = struct
+  type 'a u = [< `a of (int -> int) | `b] as 'a
+  type 'a t constraint 'a = [< `a of (int -> int) | `b]
+end
+[%%expect {|
+module M :
+  sig
+    type 'a u = 'a constraint 'a = [< `a of int -> int | `b ]
+    type 'a t constraint 'a = [< `a of int -> int | `b ]
```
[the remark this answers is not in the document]
dkalinichenko-js: Isn't this how `as 'a` works?

testsuite/tests/typing-jkind-bounds/subsumption/quality.ml
goldfirere: Left some `CR reisenberg`s around this (on call)

typing/jkind.ml
goldfirere: As discussed on the call, some of my "should be rejected" might be OK.

typing/jkind.ml
```diff
@@ -2420,21 +2420,34 @@ let for_boxed_tuple elts =
...
-       (* CR layouts v2.8: We can probably do a fair bit better here in most cases *)
-       Builtin.value ~why:Polymorphic_variant
-     else
-       Btype.fold_row
-         (fun jkind type_expr ->
-           add_with_bounds ~modality:Mode.Modality.Value.Const.id ~type_expr
-             jkind)
-         base row)
-    |> mark_best
+    let base = Builtin.immutable_data ~why:Polymorphic_variant in
```
goldfirere, line 2439: Just put this in the `else`; no reason to compute it out here

goldfirere (approved):

PR #3847, opened 2025-04-14
Cleanup of the Arm directives code 

spiessimon:
This PR makes small changes to the emission of the Arm directives to match a version using the new, shared directives code. The resulting assembly files should be semantically equivalent. The changes include:

- Adding various new lines and consistent spacing using tabs.
- Replacing `balign` with `align`.
- Adding comments to the emission of float constants.
- Printing all numbers as decimals instead of a mix of decimals and hex.
- Replacing `asciz` with `ascii`.
- Printing column information for locations (also affects x86).
- Reordering the expression for emitting distances relative to a label.

### DIFF
--- a/backend/arm64/emit.ml
+++ b/backend/arm64/emit.ml
@@ -2101,7 +2101,10 @@ let emit_item (d : Cmm.data_item) =
   | Csymbol_address s -> emit_printf "\t.quad\t%a\n" femit_symbol s.sym_name
   | Csymbol_offset (s, o) ->
     emit_printf "\t.quad\t%a+%a\n" femit_symbol s.sym_name femit_int o
-  | Cstring s -> if String.length s = 0 then emit_string "\n" else emit_string_directive "\t.ascii\t" s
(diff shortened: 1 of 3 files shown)


backend/arm64/emit.ml
```diff
@@ -562,7 +562,7 @@ let local_realloc_sites = ref ([] : local_realloc_call list)
 
 let emit_local_realloc lr =
   emit_printf "%a:\n" femit_label lr.lr_lbl;
-  emit_printf "\t%a\n" (femit_debug_info ~discriminator:0) lr.lr_dbg;
+  emit_debug_info ~discriminator:0 lr.lr_dbg;
```
gretay-js, line 565: I think this shouldn't set the discriminator.

gretay-js (approved):

backend/arm64/emit.ml
spiessimon: I think I erroneously added the discriminator in a previous PR. I've removed it.

PR #3850, opened 2025-04-14
Move CFG iterators

xclerc:
This pull request moves a couple of Cfg
iterators to their natural modules, for better
discoverability.

### DIFF
--- a/backend/cfg/cfg.ml
+++ b/backend/cfg/cfg.ml
@@ -253,6 +253,27 @@ let entry_label t = t.entry_label
 
 let iter_blocks t ~f = Label.Tbl.iter f t.blocks
 
+let iter_blocks_dfs : t -> f:(Label.t -> basic_block -> unit) -> unit =
+ fun cfg ~f ->
+  let marked = ref Label.Set.empty in
+  let rec iter (label : Label.t) : unit =
+    if not (Label.Set.mem label !marked)
+    then (
+      marked := Label.Set.add label !marked;
+      let block = get_block_exn cfg label in
+      f label block;
+      Label.Set.iter
+        (fun succ_label -> iter succ_label)
+        (successor_labels ~normal:true ~exn:true block))
+  in
+  iter cfg.entry_label;
+  (* note: some block may not have been seen since we currently cannot remove
+     all non-reachable blocks. *)
+  if Label.Set.cardinal !marked <> Label.Tbl.length cfg.blocks
+  then
+    iter_blocks cfg ~f:(fun label block ->
+        if not (Label.Set.mem label !marked) then f label block)
+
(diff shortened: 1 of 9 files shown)


backend/cfg/cfg.ml
```diff
@@ -253,6 +253,27 @@ let entry_label t = t.entry_label
 
 let iter_blocks t ~f = Label.Tbl.iter f t.blocks
 
+let iter_blocks_dfs : t -> f:(Label.t -> basic_block -> unit) -> unit =
```
spiessimon, line 256: Compared to the old version, this adds a label argument to `f`. The three occurrences where this function is called seem to discard the label. Are there currently any uses of the label?

xclerc:
Sorry, merged by mistake. Will address the
comment in [move-block-insertion](https://github.com/ocaml-flambda/flambda-backend/tree/move-block-insertion) / https://github.com/ocaml-flambda/flambda-backend/pull/3849.

backend/cfg/cfg.ml
spiessimon: Did you mean "do not use"? Otherwise, I'm a bit confused. But effectively triggering a warning when the label is not used seems fine by me :)

xclerc: I have a feeling we may be referring to
different things... The calls sites seem
to be:
- https://github.com/ocaml-flambda/flambda-backend/pull/3850/files#diff-e0f55d223b9d69c824106ea755935ef72e8b6fdf70c1ac789ccbafc9fc3320dbR98;
- https://github.com/ocaml-flambda/flambda-backend/pull/3850/files#diff-e0f55d223b9d69c824106ea755935ef72e8b6fdf70c1ac789ccbafc9fc3320dbR296;
- https://github.com/ocaml-flambda/flambda-backend/pull/3850/files#diff-37c2c960631af8f9db37fabd4b43b50a235a5e1e7f4b9f043dec0e9f5618207fR56.

All three have a label, but there was
also a label before: https://github.com/ocaml-flambda/flambda-backend/pull/3850/files#diff-37c2c960631af8f9db37fabd4b43b50a235a5e1e7f4b9f043dec0e9f5618207fL49.

spiessimon: Those are also the call sites that I was looking at. The anonymous functions passed at the call sites all seem to drop/ignore the label. And it seems the label was not passed to the function `f` previously in `iter_cfg_dfs`. Hence, I was wondering why the label argument is introduced and then dropped at all of the call sites in this PR. But I think you explained it is to trigger a warning if the label is now dropped.

xclerc: (Discussed off-line, there was a misunderstanding
around "label" as `Label.t` vs `~f`. My motivation
was to provide the same interface as `iter_blocks`.)

PR #3854, opened 2025-04-15
Separability axis basics

dkalinichenko-js:
Add the separability jkind axis. Separability is a shallow, non-modal axis going `non_float < separable < non_separable`, with `non_float` denoting types whose values are not pointers to `Double_tag` blocks, `separable` denoting types with either all values being pointers to `Double_tag` blocks or none, and `non_separable` denoting all types.

Add jkinds `non_float_value`, for non-float values, and `immutable_separable_value`, for floats. I'm not sold on the second jkind and will likely drop it from the final version of this PR.

Change other jkinds to `non_float` or `separable` where appropriate. In particular, default `immutable_data` to `non_float`. This will exclude floats from `immutable_data`, but will mark records, variants and other `immutable_data` types as `non_float`. Since `immutable_data` almost always appears in a covariant positions, this seems like the best choice.

Separability is a shallow axis, so it should not affect with-bounds. Semantics of separability for product layouts are to be determined, but it seems like all products should be non-float?

`('a : non_float_value) or_null` will be made `non_float` in a future PR.

`t : non_separable_value` will provide an escape hatch from existing OCaml separability checks (e.g. for existential) in a future PR. Those checks will also themselves take into account separability annotations.

`any_non_null` will be changed to `any_separable`, with arrays accepting `any_separable` elements, in a future PR.

### DIFF
--- a/testsuite/tests/typing-jkind-bounds/basics.ml
+++ b/testsuite/tests/typing-jkind-bounds/basics.ml
@@ -213,7 +213,7 @@ Error: The layout of type "a" is value
          because of the definition of b at line 2, characters 0-30.
 |}]
 
-type a : value mod global aliased many immutable stateless external_ unyielding
+type a : value mod global aliased many immutable stateless external_ unyielding non_float
 type b : value mod local unique once contended nonportable internal = a
 [%%expect{|
 type a : immediate
@@ -273,8 +273,8 @@ Error: Layout void is more experimental than allowed by the enabled layouts exte
 |}]
 
 type a : immediate
-type b : value mod global aliased many immutable stateless unyielding external_ = a
-type c : value mod global aliased many immutable stateless unyielding external_
+type b : value mod global aliased many immutable stateless unyielding external_ non_float = a
+type c : value mod global aliased many immutable stateless unyielding external_ non_float
 type d : immediate = c
 [%%expect{|
 type a : immediate
@@ -284,8 +284,8 @@ type d = c
 |}]
 
 type a : immediate64
-type b : value mod global aliased many immutable stateless unyielding external64 = a
-type c : value mod global aliased many immutable stateless unyielding external64
+type b : value mod global aliased many immutable stateless unyielding external64 non_float = a
+type c : value mod global aliased many immutable stateless unyielding external64 non_float
 type d : immediate64 = c
 [%%expect{|
 type a : immediate64
@@ -1173,11 +1173,11 @@ type 'a t = { x : 'a @@ global many portable aliased contended; } [@@unboxed]
    [layout_of], we'll be able to give a better jkind to [@@unboxed] types, and
    this will likely improve. *)
 
-type 'a t : value mod global immutable stateless many aliased unyielding =
+type 'a t : value mod global immutable stateless many aliased unyielding non_float =
   Foo of 'a @@ global immutable stateless many aliased [@@unboxed]
 [%%expect {|
 Lines 1-2, characters 0-66:
-1 | type 'a t : value mod global immutable stateless many aliased unyielding =
+1 | type 'a t : value mod global immutable stateless many aliased unyielding non_float =
 2 |   Foo of 'a @@ global immutable stateless many aliased [@@unboxed]
 Error: The kind of type "t" is value
          because it instantiates an unannotated type parameter of t,
@@ -1271,7 +1271,7 @@ Error: The kind of type "t" is immutable_data with 'a @@ unyielding
 type ('a : value mod aliased) t = ('a : value mod global)
 type ('a : immediate) t = ('a : value)
 type ('a : value) t = ('a : immediate)
-type ('a : value mod external_ stateless many unyielding) t = ('a : value mod immutable global aliased)
+type ('a : value mod external_ stateless many unyielding non_float) t = ('a : value mod immutable global aliased)
 type ('a : value) t = ('a : any)
 type ('a : value) t = ('a : value)
 type ('a : bits32 mod aliased) t = ('a : any mod global)
@@ -1666,8 +1666,8 @@ Error: This expression has type "int t" but an expression was expected of type
 (*********************************)
 (* Test 15: extensible variants *)
 
-(* The best kind an extensible variant can get is [value] *)
-type extensible : value = ..
+(* The best kind an extensible variant can get is [non_float_value] *)
+type extensible : non_float_value = ..
 [%%expect{|
 type extensible = ..
 |}]
@@ -1676,10 +1676,10 @@ type extensible = ..
 module M : sig
   type t : immediate with extensible
 end = struct
-  type t
+  type t : non_float_value
 end
 [%%expect{|
-module M : sig type t end
+module M : sig type t : non_float_value end
 |}]
 
 (*********************************)
@@ -1721,8 +1721,8 @@ Error: This expression has type "int t" but an expression was expected of type
 (*********************************)
 (* Test 17: extensible variants *)
 
-(* The best kind an extensible variant can get is [value] *)
-type extensible : value = ..
+(* The best kind an extensible variant can get is [non_float_value] *)
+type extensible : non_float_value = ..
 [%%expect{|
 type extensible = ..
 |}]
@@ -1731,8 +1731,8 @@ type extensible = ..
 module M : sig
   type t : immediate with extensible
 end = struct
-  type t
+  type t : non_float_value
 end
 [%%expect{|
-module M : sig type t end
+module M : sig type t : non_float_value end
 |}]
--- a/testsuite/tests/typing-jkind-bounds/basics.ml
+++ b/testsuite/tests/typing-jkind-bounds/basics.ml
@@ -1438,7 +1438,7 @@ Line 1, characters 41-51:
                                              ^^^^^^^^^^
 Error: This expression has type "<  >" but an expression was expected of type
          "('a : value mod aliased)"
-       The kind of <  > is value mod global many
+       The kind of <  > is value mod global many non_float
          because it's the type of an object.
        But the kind of <  > must be a subkind of value mod aliased
          because of the annotation on the wildcard _ at line 1, characters 19-36.
@@ -1451,7 +1451,7 @@ Line 1, characters 42-52:
                                               ^^^^^^^^^^
 Error: This expression has type "<  >" but an expression was expected of type
          "('a : value mod portable)"
-       The kind of <  > is value mod global many
+       The kind of <  > is value mod global many non_float
          because it's the type of an object.
        But the kind of <  > must be a subkind of value mod portable
          because of the annotation on the wildcard _ at line 1, characters 19-37.
@@ -1464,7 +1464,7 @@ Line 1, characters 43-53:
                                                ^^^^^^^^^^
 Error: This expression has type "<  >" but an expression was expected of type
          "('a : value mod contended)"
-       The kind of <  > is value mod global many
+       The kind of <  > is value mod global many non_float
          because it's the type of an object.
        But the kind of <  > must be a subkind of value mod contended
          because of the annotation on the wildcard _ at line 1, characters 19-38.
@@ -1477,7 +1477,7 @@ Line 1, characters 43-53:
                                                ^^^^^^^^^^
 Error: This expression has type "<  >" but an expression was expected of type
          "('a : value mod external_)"
-       The kind of <  > is value mod global many
+       The kind of <  > is value mod global many non_float
          because it's the type of an object.
        But the kind of <  > must be a subkind of value mod external_
          because of the annotation on the wildcard _ at line 1, characters 19-38.
@@ -1666,8 +1666,8 @@ Error: This expression has type "int t" but an expression was expected of type
 (*********************************)
 (* Test 15: extensible variants *)
 
-(* The best kind an extensible variant can get is [non_float_value] *)
-type extensible : non_float_value = ..
+(* The best kind an extensible variant can get is [value mod non_float] *)
+type extensible : value mod non_float = ..
 [%%expect{|
 type extensible = ..
 |}]
@@ -1676,10 +1676,10 @@ type extensible = ..
 module M : sig
   type t : immediate with extensible
 end = struct
-  type t : non_float_value
+  type t : value mod non_float
 end
 [%%expect{|
-module M : sig type t : non_float_value end
+module M : sig type t : value mod non_float end
 |}]
 
 (*********************************)
@@ -1721,8 +1721,8 @@ Error: This expression has type "int t" but an expression was expected of type
 (*********************************)
 (* Test 17: extensible variants *)
 
-(* The best kind an extensible variant can get is [non_float_value] *)
-type extensible : non_float_value = ..
+(* The best kind an extensible variant can get is [value mod non_float] *)
+type extensible : value mod non_float = ..
 [%%expect{|
 type extensible = ..
 |}]
@@ -1731,8 +1731,8 @@ type extensible = ..
 module M : sig
   type t : immediate with extensible
 end = struct
-  type t : non_float_value
+  type t : value mod non_float
 end
 [%%expect{|
-module M : sig type t : non_float_value end
+module M : sig type t : value mod non_float end
 |}]
--- a/testsuite/tests/typing-layouts-or-null/separability.ml
+++ b/testsuite/tests/typing-layouts-or-null/separability.ml
@@ -394,7 +394,7 @@ Line 1, characters 13-18:
 1 | type fails = float accepts_nonfloat
                  ^^^^^
 Error: This type "float" should be an instance of type "('a : any mod non_float)"
-       The kind of float is immutable_separable_value
+       The kind of float is value mod many unyielding stateless immutable
          because it is the primitive type float.
        But the kind of float must be a subkind of any mod non_float
          because of the definition of accepts_nonfloat at line 3, characters 0-46.
--- a/testsuite/tests/typing-layouts-or-null/separability.ml
+++ b/testsuite/tests/typing-layouts-or-null/separability.ml
@@ -1,0 +1,705 @@
+(* TEST
+ expect;
+*)
+
+(* Tests for the separability jkind axis. *)
+
+(* Basic subkinding relation. *)
+
+type t_nonsep : any mod non_separable
+type t_sep : any mod separable
+type t_nonfloat : any mod non_float
+
+[%%expect{|
+type t_nonsep : any
+type t_sep : any mod separable
+type t_nonfloat : any mod non_float
+|}]
+
+type ('a :  any mod non_separable) accepts_nonsep
+type ('a : any mod separable) accepts_sep
+type ('a : any mod non_float) accepts_nonfloat
+
+[%%expect{|
+type ('a : any) accepts_nonsep
+type ('a : any mod separable) accepts_sep
+type ('a : any mod non_float) accepts_nonfloat
+|}]
+
+type succeeds = t_nonsep accepts_nonsep
+type succeeds = t_sep accepts_nonsep
+type succeeds = t_nonfloat accepts_nonsep
+
+[%%expect{|
+type succeeds = t_nonsep accepts_nonsep
+type succeeds = t_sep accepts_nonsep
+type succeeds = t_nonfloat accepts_nonsep
+|}]
+
+type fails = t_nonsep accepts_sep
+[%%expect{|
+Line 1, characters 13-21:
+1 | type fails = t_nonsep accepts_sep
+                 ^^^^^^^^
+Error: This type "t_nonsep" should be an instance of type
+         "('a : any mod separable)"
+       The kind of t_nonsep is any
+         because of the definition of t_nonsep at line 1, characters 0-37.
+       But the kind of t_nonsep must be a subkind of any mod separable
+         because of the definition of accepts_sep at line 2, characters 0-41.
+|}]
+
+type succeeds = t_sep accepts_sep
+type succeeds = t_nonfloat accepts_sep
+
+
+[%%expect{|
+type succeeds = t_sep accepts_sep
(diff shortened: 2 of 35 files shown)


goldfirere:
Nice design write-up. For these comments, I have not looked at all at the implementation — just working through consequences of the design.

Should `int64#` be non-separable? After all, the type contains both pointers to heap-allocated floats, among other values. According to your definition, it probably should be non-separable, but my guess is that this would cause trouble. We could modify the definition to exclude non-`value`s, and that would fix this little problem. But it also suggests we should never query the separability of non-`value`s in the implementation. (I think this last sentence would be true without effort.)

So I continue to wonder about making separability a part of the `value` layout somehow. What currently happens with unboxed products of floats in arrays? My guess is that the constructing function for arrays does not do the float-check for non-`value`s. This means that we could get an array full of heap-allocated `float`s by storing an unboxed product `float`s…. but my guess is that this is just fine. All that said, I’m not against going ahead with the design here: it’s probably easier to implement than making `value` more interesting. But maybe we should briefly discuss alternatives today.

dkalinichenko-js:
Note on mod bounds for `immutable_data`: we may rename `immutable_data` to `plain_data` or `simple_data` in a subsequent PR for more clarity.

typing/jkind.ml
goldfirere, line 1: Is there a concern about maturity level for separability? I think "no". It strikes me like many other type system features that we don't need a staged release for. But it might be a good idea to go through all the different things we plan to build on top of this to make sure that none of them are too risky. Maybe you and I do this together at some point, and then we go with the results to Slack?

goldfirere, line 1: I went through this file looking for every time nullability comes up, just to make sure that the equivalent treatment for separability is right next to it. It is. Good. But is there a structural way we can ensure this? It's almost as if I want to create some kind of shallow-bounds structure and just manipulate that everywhere. What do you think?

typing/jkind.ml
```diff
@@ -1250,49 +1277,65 @@ module Const = struct
...
+          mk_jkind Any ~mode_crossing:false ~nullability:Maybe_null
+            ~separability:Non_separable;
         name = "any"
       }
 
     let any_mod_everything =
-      { jkind = mk_jkind Any ~mode_crossing:true ~nullability:Maybe_null;
+      { jkind =
+          mk_jkind Any ~mode_crossing:true ~nullability:Maybe_null
+            ~separability:Non_separable;
```
goldfirere, line 1301: This disagrees with the meaning of `everything` in the documentation. I'm not sure what's better, but we need to be consistent. Do you have a sense for what we should do?

typing/jkind.ml
```diff
@@ -1250,49 +1277,65 @@ module Const = struct
...
+            ~separability:Separable;
         name = "any_non_null"
       }
 
+    (* CR layouts v3: replace with [any_separable] when or_null arrays are implemented. *)
     let any_non_null_mod_everything =
-      { jkind = mk_jkind Any ~mode_crossing:true ~nullability:Non_null;
+      { jkind =
+          mk_jkind Any ~mode_crossing:true ~nullability:Non_null
+            ~separability:Separable;
```
goldfirere, line 1317: This also disagrees with the doc for `everything`

typing/jkind.ml
```diff
@@ -1250,49 +1277,65 @@ module Const = struct
...
+          mk_jkind (Base Value) ~mode_crossing:false ~nullability:Maybe_null
+            ~separability:Non_separable;
         name = "value_or_null"
       }
 
     let value_or_null_mod_everything =
       { jkind =
-          mk_jkind (Base Value) ~mode_crossing:true ~nullability:Maybe_null;
+          mk_jkind (Base Value) ~mode_crossing:true ~nullability:Maybe_null
+            ~separability:Non_separable;
```
goldfirere, line 1331: Another `everything` discrepancy

typing/jkind.ml
```diff
@@ -1250,49 +1277,65 @@ module Const = struct
...
         name = "value_or_null mod everything"
       }
 
     let value =
-      { jkind = mk_jkind (Base Value) ~mode_crossing:false ~nullability:Non_null;
+      { jkind =
+          mk_jkind (Base Value) ~mode_crossing:false ~nullability:Non_null
+            ~separability:Separable;
```
goldfirere, line 1338: The docs need to be updated to say that `value` is `separable`.

typing/jkind.ml
```diff
@@ -1307,12 +1350,20 @@ module Const = struct
                 ~contention:Contention.Const_op.min
                 ~statefulness:Statefulness.Const.min
                 ~visibility:Visibility.Const_op.min ~externality:Externality.max
-                ~nullability:Nullability.Non_null;
+                ~nullability:Nullability.Non_null
+                ~separability:Jkind_axis.Separability.Non_float;
```
goldfirere, line 1354:
```suggestion
                ~separability:Separability.Non_float;
```

(here and elsewhere)

typing/jkind.ml
```diff
@@ -1957,22 +2043,27 @@ module Jkind_desc = struct
...
   let map_type_expr f t = Layout_and_axes.map_type_expr f t
 
-  let of_new_sort_var nullability_upper_bound =
+  let of_new_sort_var nullability_upper_bound separability_upper_bound =
     let layout, sort = Layout.of_new_sort_var () in
     ( { layout;
         mod_bounds =
-          Mod_bounds.set_nullability nullability_upper_bound Mod_bounds.max;
+          Mod_bounds.set_nullability nullability_upper_bound Mod_bounds.max
+          |> Mod_bounds.set_separability separability_upper_bound;
```
goldfirere, line 2051:
```suggestion
          Mod_bounds.max
          |> Mod_bounds.set_nullability nullability_upper_bound
          |> Mod_bounds.set_separability separability_upper_bound;
```

typing/jkind.ml
```diff
@@ -1993,7 +2084,7 @@ module Jkind_desc = struct
       List.fold_right
         (fun (type_expr, modality) bounds ->
           With_bounds.add_modality ~relevant_for_nullability:`Relevant
-            ~type_expr ~modality bounds)
+            ~relevant_for_separability:`Irrelevant ~type_expr ~modality bounds)
```
goldfirere, line 2087: I'm surprised the two shallow axes are different. I suspect this is wrong.

typing/jkind.ml
```diff
@@ -2337,11 +2435,29 @@ let for_object =
       mod_bounds =
         Mod_bounds.create ~linearity ~locality ~uniqueness ~portability
           ~contention ~yielding ~statefulness ~visibility
-          ~externality:Externality.max ~nullability:Non_null;
+          ~externality:Externality.max ~nullability:Non_null
+          ~separability:Separability.Separable;
```
goldfirere, line 2439: non_float?

typing/typecore.ml
```diff
@@ -6920,7 +6920,8 @@ and type_expect_
       in
       let cell_type =
         (* CR uniqueness: this could be the jkind of exp2 *)
-        mk_expected (newvar (Jkind.Builtin.value ~why:Boxed_record))
+        (* CR layouts: should this be (im)mutable_data with ...? *)
```
goldfirere, line 6923: I don't think we need this CR. See the comment directly above.

typing/typedecl.ml
```diff
@@ -955,7 +955,8 @@ let transl_declaration env sdecl (id, uid) =
                    Constructor_uniform_value, sorts)
                 (Array.of_list cstrs)
             ),
-            Jkind.Builtin.value ~why:Boxed_variant
+          (* CR layouts: should this be (im)mutable_data with ...? *)
+          Jkind.Builtin.non_float_value ~why:Boxed_variant
```
goldfirere, line 959: I don't think so. This is essentially a provisional kind, to be updated later. I think probably only the layout actually matters. It's a little swimmy in my head, but I'm pretty sure we don't need to be very accurate here.

typing/typedecl.ml
```diff
@@ -973,7 +974,8 @@ let transl_declaration env sdecl (id, uid) =
                [Record_mixed].  Those cases are fixed up after we can get
                accurate sorts for the fields, in [update_decl_jkind]. *)
               Record_boxed (Array.make (List.length lbls) Jkind.Sort.Const.void),
-              Jkind.Builtin.value ~why:Boxed_record
+              (* CR layouts: should this be (im)mutable_data with ...? *)
```
goldfirere, line 977: Similar to above. I think it's fine as-is.

typing/typedecl.ml
```diff
@@ -1978,7 +1980,9 @@ let rec update_decl_jkind env dpath decl =
       decl
     | Type_open ->
       let type_jkind =
-        Jkind.Builtin.value ~why:Extensible_variant
+        Jkind.Builtin.non_float_value ~why:Extensible_variant
+        (* CR layouts: below is no longer true, and will be even less true
+           when exceptions start crossing portable/contended. Rethink? *)
```
goldfirere, line 1985: I think it's fine. The problem would be if we lower the kind without user intervention, but I don't think that's on the horizon. And getting this wrong isn't so bad -- it just means that it's more likely we have some work to do in our codebase when changing the language.

goldfirere (approved):
Looks good! Sorry this took so long to review.

typing/jkind.ml
```diff
@@ -347,6 +347,7 @@ end
 
 module Externality = Externality
 module Nullability = Nullability
+module Separability = Jkind_axis.Separability
```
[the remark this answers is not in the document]
dkalinichenko-js: This doesn't work, because there's also a `Types.Separability` module.

typing/jkind.ml
dkalinichenko-js: Makes sense, and then we'd want to move externality into modes. I think we should leave that for later PRs, though.

typing/jkind.ml
dkalinichenko-js: I think separability should be consistent with nullability, and we'll need to update the docs. `everything` really means everything modal.

typing/jkind.ml
dkalinichenko-js: Separability for products is always non-float.

typing/jkind_axis.ml
```diff
@@ -156,12 +212,14 @@ module Axis = struct
...
+      Pack (Nonmodal Nullability);
+      Pack (Nonmodal Separability) ]
 
   let name (type a) : a t -> string = function
     | Modal axis -> Format.asprintf "%a" Mode.Alloc.print_axis axis
     | Nonmodal Externality -> "externality"
     | Nonmodal Nullability -> "nullability"
+    | Nonmodal Separability -> "separability"
 
   let is_modal (type a) : a t -> bool = function
```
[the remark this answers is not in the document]
dkalinichenko-js: This function was never used anywhere, so I deleted it.

typing/jkind.ml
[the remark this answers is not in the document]
goldfirere: Blargh. OK.

testsuite/tests/typing-jkind-bounds/basics.ml
```diff
@@ -1666,8 +1666,8 @@ Error: This expression has type "int t" but an expression was expected of type
 (*********************************)
 (* Test 15: extensible variants *)
 
-(* The best kind an extensible variant can get is [value] *)
-type extensible : value = ..
+(* The best kind an extensible variant can get is [non_float_value] *)
```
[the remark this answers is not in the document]
dkalinichenko-js: Removed `non_float_value`.

testsuite/tests/typing-layouts-or-null/separability.ml
```diff
@@ -0,0 +1,705 @@
+(* TEST
+ expect;
+*)
+
+(* Tests for the separability jkind axis. *)
+
+(* Basic subkinding relation. *)
+
+type t_nonsep : any mod non_separable
```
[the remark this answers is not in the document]
dkalinichenko-js: Done.

testsuite/tests/typing-layouts-or-null/separability.ml
```diff
@@ -0,0 +1,705 @@
...
+[%%expect{|
+type 'a succeeds = 'a array
+|}]
+
+(* CR layouts v3: Arrays should accept [value_or_null mod separable] elements. *)
+
+type ('a : value_or_null mod separable) should_succeed = 'a array
+
+[%%expect{|
+type 'a should_succeed = 'a array
```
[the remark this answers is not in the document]
dkalinichenko-js: No, that's not what I meant, thanks for pointing this out.

testsuite/tests/typing-layouts-or-null/separability.ml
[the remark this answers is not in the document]
dkalinichenko-js: Done.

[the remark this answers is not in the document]
dkalinichenko-js: Thanks for pointing this out, I didn't realize 1-element unboxed records inherit the kind of their elements. This means we do need to consider separability for product jkind, and for now we have suboptimal bounds on 2+ element product jkinds. Maybe that's fine. 

The interactions between product jkinds, nullability and separability are all terrible. We should come up with something better.

typing/jkind.ml
```diff
@@ -393,7 +394,8 @@ let raise ~loc err = raise (Error.User_error (loc, err))
 
 (* Returns the set of axes that is relevant under a given modality. For example,
    under the [global] modality, the locality axis is *not* relevant. *)
-let relevant_axes_of_modality ~relevant_for_nullability ~modality =
+let relevant_axes_of_modality ~relevant_for_nullability
```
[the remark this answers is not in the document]
goldfirere: But... why?

typing/jkind.ml
goldfirere: I think it meshes fine with the inference we have (combining `value` and `word` would not). You'll have seen my little proposal doc about this. But I agree that we don't have to fix this now. But I might urge us to fix it if we add any more shallow axes.

typing/jkind.ml
goldfirere: Then sadly `immediate` won't be `value mod everything`. Instead it would be `value mod everything non_null non_float`. Seems a little sad. But if we're going to go with this, maybe update the doc in this PR?

typing/jkind_axis.ml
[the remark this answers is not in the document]
goldfirere: Even better. Thanks.

typing/jkind.ml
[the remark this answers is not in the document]
dkalinichenko-js: Separability is only needed due to the floating point array optimization, which is only relevant for basic sort jkinds. For product jkinds, it's never applied, so we can always consider them non-float.

testsuite/tests/typing-layouts-or-null/separability.ml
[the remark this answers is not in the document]
goldfirere: I certainly agree with your final sentences, but I don't think we need to consider separability for the product jkind, because there is no unary product. Maybe the function that makes products sometimes has only one argument? But the `&` is never applied to only one thing.

typing/jkind.ml
dkalinichenko-js: `value mod everything non_float` since `value` is already `non_null`.

I'll update the docs in a subsequent PR now. I've noticed that they get a few things wrong even without separability, I'll fix that too.

typing/jkind.ml
```diff
@@ -2335,6 +2319,22 @@ let for_unboxed_record lbls =
...
+        Mod_bounds.create ~locality:Locality.Const.max
+          ~linearity:Linearity.Const.max ~portability:Portability.Const.max
+          ~yielding:Yielding.Const.max ~uniqueness:Uniqueness.Const_op.max
+          ~contention:Contention.Const_op.max
+          ~statefulness:Statefulness.Const.max
+          ~visibility:Visibility.Const_op.max ~externality:Externality.max
+          ~nullability:Nullability.Non_null ~separability:Separability.Non_float;
+      with_bounds = No_with_bounds
+    }
+    ~annotation:(mk_annot "non_float_value")
```
goldfirere, line 2352: I think we don't use this annotation field much (at all?), but it should probably be `value mod non_float`.

testsuite/tests/typing-layouts-or-null/separability.ml
```diff
@@ -736,3 +736,164 @@ Error: This type "int or_null" should be an instance of type
...
+7 | type works = int or_null smth bounded
+                 ^^^^^^^^^^^^^^^^
+Error: This type "int or_null smth" should be an instance of type
+         "('a : immediate)"
+       The kind of int or_null smth is immediate with int or_null
+         because of the definition of smth at line 1, characters 0-50.
+       But the kind of int or_null smth must be a subkind of immediate
+         because of the definition of bounded at line 3, characters 0-29.
+|}]
+
```
goldfirere, line 776: I'd still want a more direct test:

```ocaml
type 'a t : value mod non_float with 'a
type ('a : value mod non_float) req_non_float
type test = float t req_non_float
```

testsuite/tests/typing-layouts-or-null/separability.ml
```diff
@@ -736,3 +736,164 @@ Error: This type "int or_null" should be an instance of type
...
+(* Separability and [@@unboxed]. *)
+
+type unbx = { unbx : t_nonsep_val } [@@unboxed]
+
+(* CR layouts v3.4: non-separable unboxed records should be allowed. *)
```
goldfirere, line 781: Interesting. Doesn't have to be fixed in this PR.

testsuite/tests/typing-layouts-or-null/separability.ml
```diff
@@ -736,3 +736,164 @@ Error: This type "int or_null" should be an instance of type
...
+
+
+type b = #{ a : int; b: t_nonsep_val; c: float# }
+
+type ('b : value & value & float64 mod everything non_float) fails = unit constraint 'b = b
+
+(* CR layouts v3.4: separability of 2+ element unboxed records should always be
+   non-float.
+
+   Also, this error is horrible. *)
```
goldfirere, line 839: Because the with-bounds are repeated in every factor? Otherwise, I don't think it's that bad, really. And it's right -- this definition should be rejected.

testsuite/tests/typing-layouts-or-null/separability.ml
```diff
@@ -736,3 +736,164 @@ Error: This type "int or_null" should be an instance of type
...
+         & float64 mod everything with t_nonsep_val
+         because of the definition of b at line 1, characters 0-49.
+       But the kind of b must be a subkind of
+         immediate & immediate & float64 mod everything
+         because of the annotation on 'b in the declaration of the type fails.
+|}]
+
+type c = #( float * float or_null * float# )
+
+(* CR layouts v3.4: separability of unboxed tuples should always be
```
goldfirere, line 860: Again, I think the error here is accurate.

typing/jkind.ml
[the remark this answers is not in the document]
goldfirere: Sure.... but isn't that also just as true for nullability? Or is this because we can't put maybe_null things in arrays, but we want to put non-values in arrays? Anyway, I think the two are aligned now.

typing/jkind.ml
goldfirere: OK

testsuite/tests/typing-layouts-or-null/separability.ml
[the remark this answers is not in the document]
dkalinichenko-js: Ok, we don't need to consider it for the jkind, but we do need to consider it for unboxed records.

typing/jkind.ml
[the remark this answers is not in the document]
dkalinichenko-js: We might want to support `t or_null` where `t : value & value`. I'm ambivalent about this, but it was the plan.

testsuite/tests/typing-layouts-or-null/separability.ml
dkalinichenko-js: It's an accurate error, it's just very verbose, displaying a lot of unnecessary information.

testsuite/tests/typing-layouts-or-null/separability.ml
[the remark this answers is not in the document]
goldfirere: OK. But I realize now that this is just about adding tests, which you've done. Should we resolve?

typing/jkind.ml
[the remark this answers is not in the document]
goldfirere: We could debate that if we want... but I think these are aligned now, so it's probably ok just to Resolve, I think.

typing/jkind.ml
dkalinichenko-js: Oops. I think it actually should be `None`, since this jkind doesn't come from annotations. Fixed.

typing/jkind.ml
goldfirere: Does this need to hold up merging this patch? I think "no": this patch, on its own, can't cause anything to go wrong, I think. It's just future patches that would. If you agree, please Resolve, and we can make sure to address this soon.

goldfirere (commented):
Reviewed most recent commit

dkalinichenko-js:
Ack, sorry! I usually rebase when going back to the feature to avoid another review cycle for fixes due to the rebase, but I didn't consider the effect on your review.

I think GitHub can track which commits you reviewed though? At least I recall it keep tracking the new additions for me when I'm reviewing PRs. Maybe rebase breaks that?

typing/jkind.ml
dkalinichenko-js: Fixed in https://github.com/ocaml-flambda/flambda-backend/pull/3854.

goldfirere:
Yes rebase breaks that. GitHub also tracks files that I've seen, but at a per-file granularity, which is only marginally useful.

(Rebasing the way you have does make sense! It's just too bad that GitHub doesn't accommodate that good workflow so well.)

PR #3855, opened 2025-04-15
Move Capply to Cmm.expression and harmonise with Cfg

mshinwell:
This moves `Capply` from being a Cmm `operation` and makes it an `expression`, matching up with what happens in Flambda 2.  A further refinement is made: applications corresponding to extcalls which do not require `caml_c_call` and which always return remain as `operation`s (this is not yet reflected in Flambda 2 but we will probably change this at some point, by adding an extcall case to `Flambda_primitive`).

This simplifies various code and introduces a nice property: target-specific instruction selection cannot introduce control flow (this is seen by the removal of the `Cfg.basic_or_terminator` type).  This may become important for potential future changes/replacement of the Cmm language, e.g. by a representation taking some inspiration from the "sea of nodes" approaches.  In addition, after this PR and #3743, we are close to having the control flow constructs in Cmm match those of Flambda 2.

While matching this up with what is provided by `Cfg` (which also required a change to permit the above special case of extcalls to be an `Operation`) I found the existing type confusing, and have refactored it to match up with `Cmm`.  As noted in a CR I think we should also move the tailcall cases under the new `Call` terminator, which will then match up with `Linear`.  It is possible this part could be split into a separate PR but I'm unsure that it's worth the work and the risk of introducing bugs in instruction selection.

I haven't made the necessary changes for the x86-64 backend yet, pending a first pass of review.  (I will get in touch with potential reviewers directly.)

### DIFF
--- a/backend/arm64/emit.ml
+++ b/backend/arm64/emit.ml
@@ -859,7 +859,7 @@ let num_call_gc_points instr =
         | Intop_atomic _
         | Floatop (_, _)
         | Csel _ | Reinterpret_cast _ | Static_cast _ | Probe_is_enabled _
-        | Name_for_debugger _ | Extcall _ )
+        | Name_for_debugger _ | External _ )
     | Lprologue | Lreloadretaddr | Lreturn | Lentertrap | Lpoptrap | Lcall_op _
     | Llabel _ | Lbranch _
     | Lcondbranch (_, _)
@@ -926,7 +926,7 @@ module BR = Branch_relaxation.Make (struct
           | Intop_atomic _
           | Floatop (_, _)
           | Csel _ | Reinterpret_cast _ | Static_cast _ | Probe_is_enabled _
-          | Name_for_debugger _ | Extcall _ )
+          | Name_for_debugger _ | External _ )
       | Lprologue | Lend | Lreloadretaddr | Lreturn | Lentertrap | Lpoptrap
       | Lcall_op _ | Llabel _ | Lbranch _ | Lswitch _ | Ladjust_stack_offset _
       | Lpushtrap _ | Lraise _ | Lstackcheck _ ->
@@ -972,7 +972,7 @@ module BR = Branch_relaxation.Make (struct
           { alloc; stack_ofs; func = _; ty_res = _; ty_args = _; returns = _ })
       ->
       if Config.runtime5 && stack_ofs > 0 then 5 else if alloc then 3 else 5
-    | Lop (Extcall _) ->
+    | Lop (External _) ->
       (* Never allocates *)
       5
     | Lop (Stackoffset _) -> 2
@@ -1514,7 +1514,7 @@ let emit_instr i =
   | Lcall_op (Lextcall { func; alloc; stack_ofs; _ }) ->
     emit_extcall i ~func ~alloc ~stack_ofs
   | Lop
-      (Extcall
+      (External
         { func_symbol = func;
           effects = _;
           ty_args = _;
--- a/backend/arm64/emit.ml
+++ b/backend/arm64/emit.ml
@@ -859,7 +859,7 @@ let num_call_gc_points instr =
         | Intop_atomic _
         | Floatop (_, _)
         | Csel _ | Reinterpret_cast _ | Static_cast _ | Probe_is_enabled _
-        | Name_for_debugger _ )
+        | Name_for_debugger _ | Extcall _ )
     | Lprologue | Lreloadretaddr | Lreturn | Lentertrap | Lpoptrap | Lcall_op _
     | Llabel _ | Lbranch _
     | Lcondbranch (_, _)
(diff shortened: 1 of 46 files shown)


mshinwell:
I should also have clarified: the cases for probes and extcalls that do not satisfy the above condition also had to be moved into `Cmm.expression`.

mshinwell:
I've realised that `Craise` needs to be treated as an expression, not an operation, to ensure no operation has control flow effects.

backend/cfg/cfg.ml
```diff
@@ -710,3 +743,22 @@ let equal_irc_work_list left right =
...
+  in
+  DLL.iter ~f:(pp_with_id ppf ~pp:print_basic) block.body;
+  pp_with_id ppf ~pp:print_terminator block.terminator;
+  fprintf ppf "\npredecessors:";
+  Label.Set.iter (fprintf ppf " %a" Label.format) block.predecessors;
+  fprintf ppf "\nsuccessors:";
+  Label.Set.iter
+    (fprintf ppf " %a" Label.format)
+    (successor_labels ~normal:true ~exn:false block);
+  fprintf ppf "\nexn-successors:";
```
xclerc, line 760:
```suggestion
  fprintf ppf "\nexn-successor:";
```

backend/cfg/cfg_comballoc.ml
```diff
@@ -38,7 +38,7 @@ let rec find_next_allocation : cell option -> allocation option =
         | Intop _ | Intop_imm _ | Intop_atomic _ | Floatop _ | Csel _
         | Reinterpret_cast _ | Static_cast _ | Probe_is_enabled _ | Opaque
         | Begin_region | End_region | Specific _ | Name_for_debugger _ | Dls_get
-        | Poll )
+        | Poll | Extcall _ )
```
xclerc, line 41: Should we comballoc across `Extcall`?

backend/cfg/cfg_intf.ml
```diff
@@ -40,18 +40,29 @@ module S = struct
     { func_symbol : string;
       alloc : bool;
       (* CR mshinwell: rename [alloc] -> [needs_caml_c_call] *)
+      returns : Label.t option;
+          (** At least one of [alloc] and [returns] must be set.  Otherwise,
```
xclerc, line 44: > At least one of [alloc] and [returns] must be set

Not necessarily in this pull request, but shouldn't
we enforce that through types?

backend/cfg/cfg_comballoc.ml
xclerc: (Discussed off-line, this is thought
to be fine because the callee should
not be able to see the uninitialized
bits of the "combined" allocation.)

backend/arm64/emit.ml
```diff
@@ -971,7 +971,10 @@ module BR = Branch_relaxation.Make (struct
         (Lextcall
           { alloc; stack_ofs; func = _; ty_res = _; ty_args = _; returns = _ })
       ->
+      (* XXX shouldn't this be "if alloc then 5 else 3"? *)
       if Config.runtime5 && stack_ofs > 0 then 5 else if alloc then 3 else 5
+    | Lop (Extcall { stack_ofs; _ }) ->
+      if Config.runtime5 && stack_ofs > 0 then 5 else 5
```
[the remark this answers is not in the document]
mshinwell: I was wrong here

backend/cfg/cfg_intf.ml
mshinwell: See cmm.mli

backend/cfg/cfg_intf.ml
```diff
@@ -40,18 +40,29 @@ module S = struct
...
+              use [Op]. *)
+      (* CR mshinwell: same comment as in cmm.mli about [effects], [alloc] and
```
gretay-js, line 46: Which comment you are referring to?

gretay-js (commented):
I've reviewed the Cfg changes and they look ok. Can you please split them into a separate PR?

- The property "Cmm operation has no control-flow effects" is confusing because Alloc and Poll do have control flow in the generated assembly (the jump to call gc, not only async exception in the out of memory case). 
- Consider renaming `Cextcall` to make it clear that it's a "no-alloc" call. It'll also help avoid bugs in rebase and general confusion with the established terminology.
- Cextcall and External fields are almost the same, should the record be shared? also it would make it easier to enforce the constraint on `alloc` and `returns` fields of `External`.
- An explicit terminator for `Call_external_no_return` would be nicer, I think. It would avoid the special cases that now need to `match` on terminators just for this purpose like this
`Call (External { returns = None; _ }` . It would also avoid the constraint "At least one of [alloc] and [returns] must be set".

backend/cfg/cfg_intf.ml
mshinwell: Clarified

PR #3859, opened 2025-04-15
Remove cifthenelse

mshinwell:
This is to bring the Cmm language further in line with Flambda 2.

I think this should be ok, but will look at the generated code.  Two problems might be:
* if instruction selection failed to recognise the pattern generated by `Cmm_helpers.ifthenelse`, what will happen?  This should be fine if the results of e.g. `Iinttest` are in `{0, 1}`, but is that guaranteed?
* if an existing `Cswitch` on `{0, 1}` gets caught by the new pattern in instruction selection, will the code be worse?  It's unclear such a switch would ever be generated anyway (see `To_cmm_expr.switch`).


mshinwell:
This is based on #3855

mshinwell:
Ah but the first bullet point isn't quite right: the "`test`"s won't be generated unless we go through the if-then-else instruction selection code.  So we would go via `select_arith_comp` and it should be fine.

PR #3864, opened 2025-04-15
Introduce a new flavor of type, `Tof_kind`, for representing existentials that end up in with-bounds

glittershark:
Add a new flavor of type `Tof_kind`, which will (in a forthcoming PR) be used to stand-in for existential types that end up in with-bounds (eg because they're an argument to an abstract type constructor). The syntax looks like:
```ocaml
type 'a abstract
type existential_abstract : immutable_data with (type : value mod portable) abstract =
  | Mk : ('a : value mod portable). 'a abstract -> existential_abstract
```

These types are uninhabited, and can never be inferred as the type of an expression - their only purpose is to show up in with-bounds. Another way of doing this would be to put a TConstr of every possible combination of mod-bounds in the initial environment, but this felt moderately cleaner and didn't end up being much more work

Note that this is basically untested as of this PR, except for the parsing - a subsequent PR will actually use this in kind inference for GADTs, and there more testing is performed.

Review all in one go; the commits aren't useful.

### DIFF
--- a/parsing/pprintast.ml
+++ b/parsing/pprintast.ml
@@ -525,7 +525,7 @@ and core_type ctxt f x =
                           l)
           sl (core_type ctxt) ct
     | Ptyp_of_kind jkind ->
-      pp f "(type : %a)" (jkind_annotation reset_ctxt) jkind
+      pp f "@[(type@ :@ %a)@]" (jkind_annotation reset_ctxt) jkind
     | _ -> pp f "@[<2>%a@]" (core_type1 ctxt) x
 
 and core_type1 ctxt f x =
--- a/parsing/pprintast.ml
+++ b/parsing/pprintast.ml
@@ -524,6 +524,8 @@ and core_type ctxt f x =
                           (tyvar_loc_jkind tyvar) ~sep:"@;")
                           l)
           sl (core_type ctxt) ct
+    | Ptyp_of_kind jkind ->
+      pp f "(type : %a)" (jkind_annotation reset_ctxt) jkind
     | _ -> pp f "@[<2>%a@]" (core_type1 ctxt) x
 
 and core_type1 ctxt f x =
@@ -608,7 +610,7 @@ and core_type1 ctxt f x =
     | Ptyp_open(li, ct) ->
        pp f "@[<hov2>%a.(%a)@]" longident_loc li (core_type ctxt) ct
     | Ptyp_extension e -> extension ctxt f e
-    | (Ptyp_arrow _ | Ptyp_alias _ | Ptyp_poly _) ->
+    | (Ptyp_arrow _ | Ptyp_alias _ | Ptyp_poly _ | Ptyp_of_kind _) ->
        paren true (core_type ctxt) f x
 
 and core_type2 ctxt f x =
--- a/toplevel/genprintval.ml
+++ b/toplevel/genprintval.ml
@@ -589,6 +589,8 @@ module Make(O : OBJ)(EVP : EVALPATH with type valu = O.t) = struct
               tree_of_val (depth - 1) obj ty
           | Tpackage _ ->
               Oval_stuff "<module>"
+          | Tof_kind _jkind ->
+              Oval_stuff "of_kind_ <kind>"
         end
 
       and tree_of_record_fields depth env path type_params ty_list
--- a/toplevel/genprintval.ml
+++ b/toplevel/genprintval.ml
@@ -583,14 +583,12 @@ module Make(O : OBJ)(EVP : EVALPATH with type valu = O.t) = struct
                 find (row_fields row)
           | Tobject (_, _) ->
               Oval_stuff "<obj>"
-          | Tsubst _ | Tfield(_, _, _, _) | Tnil | Tlink _ ->
+          | Tsubst _ | Tfield(_, _, _, _) | Tnil | Tlink _ | Tof_kind _  ->
               fatal_error "Printval.outval_of_value"
           | Tpoly (ty, _) ->
               tree_of_val (depth - 1) obj ty
           | Tpackage _ ->
               Oval_stuff "<module>"
-          | Tof_kind _jkind ->
-              Oval_stuff "of_kind_ <kind>"
         end
 
       and tree_of_record_fields depth env path type_params ty_list
--- a/typing/btype.ml
+++ b/typing/btype.ml
@@ -260,7 +260,10 @@ let fold_row f init row =
       (row_fields row)
   in
   match get_desc (row_more row) with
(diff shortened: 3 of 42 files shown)


parsing/pprintast.ml
```diff
@@ -524,6 +524,8 @@ and core_type ctxt f x =
                           (tyvar_loc_jkind tyvar) ~sep:"@;")
                           l)
           sl (core_type ctxt) ct
+    | Ptyp_of_kind jkind ->
+      pp f "(type : %a)" (jkind_annotation reset_ctxt) jkind
```
goldfirere, line 528:
```suggestion
      pp f "@[(type :@ %a)@]" (jkind_annotation reset_ctxt) jkind
```

I think will line-break better. Maybe I'm wrong, though.

parsing/pprintast.ml
```diff
@@ -608,7 +610,7 @@ and core_type1 ctxt f x =
     | Ptyp_open(li, ct) ->
        pp f "@[<hov2>%a.(%a)@]" longident_loc li (core_type ctxt) ct
     | Ptyp_extension e -> extension ctxt f e
-    | (Ptyp_arrow _ | Ptyp_alias _ | Ptyp_poly _) ->
+    | (Ptyp_arrow _ | Ptyp_alias _ | Ptyp_poly _ | Ptyp_of_kind _) ->
```
goldfirere, line 613: The `paren true` here makes me think this case will cause double parens.

toplevel/genprintval.ml
```diff
@@ -589,6 +589,8 @@ module Make(O : OBJ)(EVP : EVALPATH with type valu = O.t) = struct
               tree_of_val (depth - 1) obj ty
           | Tpackage _ ->
               Oval_stuff "<module>"
+          | Tof_kind _jkind ->
+              Oval_stuff "of_kind_ <kind>"
```
goldfirere, line 593: This really should never happen, I think -- it would be a value of type `Tof_kind blah`. I think this could `fatal_error`.

typing/btype.ml
```diff
@@ -260,7 +260,7 @@ let fold_row f init row =
       (row_fields row)
   in
   match get_desc (row_more row) with
-  | Tvar _ | Tunivar _ | Tsubst _ | Tconstr _ | Tnil ->
+  | Tvar _ | Tunivar _ | Tsubst _ | Tconstr _ | Tnil | Tof_kind _ ->
```
goldfirere, line 263: This scares me a little. I don't have a semantics for when `Tof_kind` appears in a `row_more` field. Do we need this? If so, I'd like to understand the case better.

typing/types.mli
```diff
@@ -234,6 +234,15 @@ and type_desc =
   | Tpackage of Path.t * (Longident.t * type_expr) list
   (** Type of a first-class module (a.k.a package). *)
 
+  | Tof_kind of jkind_lr
+  (** [Tof_kind jkind] ==> [(type : jkind)]
+
+      The "canonical" type of a particular kind.
+
+      These types are uninhabited, and any appearing in translation will cause an error.
+      They are only used to represent the kinds of existentially-quantified types
```
goldfirere, line 243: This comment should either have an example or point to a test case. This is subtle.

typing/typetexp.ml
```diff
@@ -1066,6 +1066,10 @@ and transl_type_aux env ~row_context ~aliased ~policy mode styp =
       in
       let cty = transl_type new_env ~policy ~row_context mode t in
       ctyp (Ttyp_open (path, mod_ident, cty)) cty.ctyp_type
+  | Ptyp_of_kind jkind ->
+    let tjkind = jkind_of_annotation (Type_of_kind loc) styp.ptyp_attributes jkind in
+    let ty = Btype.newgenty (Tof_kind tjkind) in
```
goldfirere, line 1071: I wouldn't use `newgenty` here. Just use `newty`. Using `newgenty` will violate the invariant that levels always go down as you descend through a type.

typing/outcometree.mli
```diff
@@ -154,6 +154,7 @@ and out_type =
   | Otyp_jkind_annot of out_type * out_jkind
       (* Currently only introduced with very explicit code in [Printtyp] and not
          synthesized directly from the [Typedtree] *)
+  | Otyp_canonical of out_jkind
```
goldfirere, line 157: Rename, please

typing/ctype.ml
```diff
@@ -2271,7 +2271,7 @@ let type_jkind_purely_if_principal' =
    don't. *)
 let rec estimate_type_jkind ~expand_component env ty =
   match get_desc ty with
-  | Tvar { jkind } -> Jkind.disallow_right jkind
+  | Tvar { jkind } | Tof_kind jkind -> Jkind.disallow_right jkind
```
goldfirere, line 2274: This is a bit too quiet for my tastes. Just make a new case, please, like `Tunivar`.

goldfirere (approved):
I actually think the design here is slightly off. We don't need to store an entire jkind here. Instead, we just need some mod-bounds (and only the modal mod-bounds at that). This might matter in normalization, where we can safely deduplicate `(type : value mod portable) abstract` and `(type : float64 mod portable) abstract`. So maybe this should be e.g. `(type mod portable)`. I'd be OK merging this patch before revising (I don't think this is unsound or anything), but simplicity is good. And it means I don't have to worry about whether e.g. unification might affect sort variables.

typing/btype.ml
```diff
@@ -260,7 +260,10 @@ let fold_row f init row =
       (row_fields row)
   in
   match get_desc (row_more row) with
-  | Tvar _ | Tunivar _ | Tsubst _ | Tconstr _ | Tnil | Tof_kind _ ->
+  | Tvar _ | Tunivar _ | Tsubst _ | Tconstr _ | Tnil
+    (* Tof_kind can appear in [row_more] in case the row's row variable was existentially
+       quantified in a GADT *)
```
goldfirere, line 265: This is helpful, but still doesn't have all of the detail I would want. So I have made #3934 so that there is a good place to put those details.

I *think* that any `Tof_kind` in a row variable will always be of kind `value`, because we are not yet supporting having an open polymorphic variant whose fields are limited to those that mode-cross a certain way.

parsing/pprintast.ml
goldfirere: I do not understand why my comment above is wrong, but it evidently is. I've added a test. Moving on.

typing/btype.ml
goldfirere: Let's not hold this up on resolving this.

glittershark (commented):
last two commits LGTM

PR #3865, opened 2025-04-15
modalities zap to id for mli-less files

dkalinichenko-js:
Rebasing and re-opening https://github.com/ocaml-flambda/flambda-backend/pull/3282 to fix the issue with mli-less modules inferring non-legacy modalities for `module type of`.

### DIFF
--- a/testsuite/tests/typing-modes/val_modalities_floor.ml
+++ b/testsuite/tests/typing-modes/val_modalities_floor.ml
@@ -1,12 +1,8 @@
-(* We test that cmi generated without mli has the modalities pushed to the
-strongest instead of legacy *)
-
 (* TEST
  readonly_files = "\
    def_portable.ml \
-   def_nonportable.ml \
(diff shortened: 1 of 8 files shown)


testsuite/tests/typing-modes/val_modalities_floor.ml
goldfirere, line 1: There's a lot of changes / deletions here. What's going on?

goldfirere (commented):
Sadly, I'm not quite comfortable here without @riaqn's review, especially about the changes in the locations tests, which seem to keep more mode variables than before.

testsuite/tests/typing-modes/val_modalities_floor.ml
dkalinichenko-js: `def_nonportable.ml` explicitly annotated `f` as non-portable. But now that's unnecessary, since with the absent interface the function is defaulted to non-portable anyway.

PR #3866, opened 2025-04-16
Use Typing_env_extension for CSE join result

bclement-ocp:
This is a longstanding CR and reduces the exposure of types' internals to simplify.

(Groundwork for the match-in-match heuristic)


lthls (approved):
I was slightly worried about the title (I don't think we want to use typing env extensions inside `CSE.t` structures), but the PR itself only impacts the join of CSE equations.

bclement-ocp:
> I don't think we want to use typing env extensions inside CSE.t structures

Agreed — it's not about CSE equations, it's about type equations created by CSE; will change the title.

PR #3867, opened 2025-04-16
flambda2: Do not expose relations as types 

bclement-ocp:
Add dedicated functions for recording relational primitives (Is_int, Get_tag, Is_null) instead. This makes it easier to change the internal representation of relations in the typing env.

(This is groundwork for the match-in-match heuristic and improved relations)


lthls (approved):
Looks good to me

PR #3868, opened 2025-04-16
Param libs errors as cmi warnings

lukemaurer:
With `-no-alias-deps`, we actually do look at .cmi files for aliases, but we catch most errors and turn them into warnings. Only currently we assert `assert false` for any of the errors relating to parameterized modules. Instead, we should be turning each possible exception into a short blurb that gets tacked onto the warning.

I included some test cases where they were easy to write. Note that there are a few cases that currently aren't raising warnings at all, which is odd but not in scope for this PR.


riaqn:
@lukemaurer CI is failing - could you fix that?

PR #3871, opened 2025-04-16
Format more backend files

gretay-js:
Disabled formatting of a couple of comments where vertical alignment is useful. The rest is `make fmt`. Might conflict with open PRs but should be easy to rebase by cherry-pick the changes in `backend/.ocamlformat-enable`.


backend/cmm.mli
```diff
@@ -24,7 +24,7 @@ type machtype_component = Cmx_format.machtype_component =
   | Float32
   | Valx2
 
-(* - [Val] denotes a valid OCaml value: either a pointer to the beginning
+(*= - [Val] denotes a valid OCaml value: either a pointer to the beginning
```
spiessimon, line 27: The indentation is now a bit off between this item and the next one.

```suggestion
(*=- [Val] denotes a valid OCaml value: either a pointer to the beginning
```

backend/cmmgen_state.ml
```diff
@@ -36,40 +38,40 @@ and uconstant =
...
 let compare_constants c1 c2 =
   match c1, c2 with
-  | Const_ref(lbl1, _c1), Const_ref(lbl2, _c2) -> String.compare lbl1 lbl2
-      (* Same labels -> same constants.
-         Different labels -> different constants, even if the contents
-           match, because of string constants that must not be
-           reshared. *)
+  | Const_ref (lbl1, _c1), Const_ref (lbl2, _c2) ->
+    String.compare lbl1 lbl2
+    (* Same labels -> same constants. Different labels -> different constants,
```
spiessimon, line 60: Here, perhaps the spacing was intentional. But should also be fine like this.

spiessimon (approved):
Looks good to me with or without the nits.

backend/cmmgen_state.ml
gretay-js: yes, I think it's readable after formating.

backend/cmm.mli
gretay-js: fixed

PR #3873, opened 2025-04-16
flambda2(types): Refactor `add_equation`

bclement-ocp:
Previously, the `add_equation` logic handled multiple concerns at the same time in a relatively convoluted way, in part due to the need to accomodate both basic and advanced meet algorithms in the same code. With the removal of the basic meet (#3726), we can considerably simplify the code.

This commit refactors `add_equation` by decomposing it into smaller, well-defined steps and recognizing that concrete types and aliases need to be handled differently.

This restructuring clarifies the process of adding an equation:

1. Compute the canonical of the LHS with `get_canonical_simple_ignoring_name_mode`

2. Check if the type is a concrete type or an alias, which is then resolved to its current canonical (`add_equation_on_canonical`).

3. If an alias, merge the two canonical simples using `merge_distinct_canonicals` (which updates the Aliases structure), then call `record_demotion` to update the typing env, transferring the existing concrete type from the demoted element to its new canonical element.

4. If a concrete type, perform the `meet` with the existing concrete type, then call `replace_concrete_equation` if a more precise concrete type is found.  `replace_concrete_equation` automatically demotes to a constant instead if the concrete type is a singleton (without going through `record_demotion`, since we have already computed the `meet`).

Additionally, `Type_grammar.recover_const_alias` is renamed to `must_be_singleton`, which better captures its behavior.

Overall, this should be easier to read and maintain. It also facilitates extensions; for instance, the improved relations tracking for the match-in-match heuristics will rely on `record_demotion` and `add_alias_between_canonicals`.

### DIFF
--- a/middle_end/flambda2/types/env/typing_env.ml
+++ b/middle_end/flambda2/types/env/typing_env.ml
@@ -722,25 +722,9 @@ let invariant_for_new_equation (t : t) name ty =
       Misc.fatal_errorf "New equation@ %a@ =@ %a@ has unbound names@ (%a):@ %a"
         Name.print name TG.print ty Name_occurrences.print unbound_names print t)
 
-let pattern_match_equation ~name ~const simple ty =
-  (* We have [(coerce <bare_lhs> <coercion>) : <ty>]. Thus [<bare_lhs> : (coerce
-     <ty> <coercion>^-1)]. *)
-  let bare_lhs = Simple.without_coercion simple in
-  let coercion_from_bare_lhs_to_ty = Simple.coercion simple in
-  let coercion_from_ty_to_bare_lhs =
-    Coercion.inverse coercion_from_bare_lhs_to_ty
-  in
-  let ty = TG.apply_coercion ty coercion_from_ty_to_bare_lhs in
-  let name eqn_name ~coercion =
-    (* true by definition *)
-    assert (Coercion.is_id coercion);
-    name eqn_name ty
-  in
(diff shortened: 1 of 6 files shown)


middle_end/flambda2/types/env/typing_env.ml
```diff
@@ -792,33 +776,70 @@ let rec replace_equation (t : t) name ty =
...
-and replace_equation_or_add_alias_to_const t name ty =
-  match TG.recover_const_alias ty with
-  | None -> replace_equation t name ty
+type merge_distinct_canonicals_result =
+  { demoted : Name.t;
+    canonical : Simple.t;
+    t : t
+  }
+
+let merge_distinct_canonicals t ~canonical_element1 ~canonical_element2 :
```
bclement-ocp, line 785: Note: I'm not too happy about the name of this one. The goal is to wrap `Aliases.add` and massage its return type to return a `Name.t` -> `Simple.t` demotion, but I don't want to call it `add_alias` because I think that should be the name of a function to add an alias between two arbitrary names (which should resolve the canonicals and call `merge_distinct_canonicals` if needed). Happy to take suggestions.

bclement-ocp: Actually it makes sense to just moving this to `Aliases.add`, doing it.

bclement-ocp: Moved this directly into `Aliases.add`, sorry for the noise!

lthls (approved):
Thanks, this makes the control flow of `add_equation` much easier to follow.

PR #3874, opened 2025-04-16
Move priority queue to utils folder

gretay-js:


utils/priority_queue.mli
```diff
@@ -0,0 +1,40 @@
+(** Max priority queue  *)
+
+[@@@ocaml.warning "+a-40-41-42"]
+
```
xclerc, line 4: (Like on #3875, I think we should leave
a comment/CR about the use of a stdlib
module - `Pqueue`.)

xclerc (approved):

utils/priority_queue.mli
gretay-js: I put a CR in `priority_queue.ml`

PR #3877, opened 2025-04-17
Nuke bogus test file

NickBarnes:
The file `test_register_compatible_vectorized.cmx.dump.expected` at the top-level was introduced in error a few months ago. It's annoying (hence this PR) but otherwise harmless.


NickBarnes:
This file got removed by another change.

PR #3883, opened 2025-04-17
Fix kind checks for partial/over applications

Ekdohibs:
The check is moved behind `Flambda_features.kind_checks ()`, as the error can be triggered in valid code with GADTs (see test files). I had to change a bit more code in the over-application case, as it otherwise produced another fatal error if the check was removed.

### DIFF
--- a/middle_end/flambda2/simplify/simplify_common.ml
+++ b/middle_end/flambda2/simplify/simplify_common.ml
@@ -241,13 +241,34 @@ let split_direct_over_application apply
         ~free_names_of_body:(Known perform_over_application_free_names)
   in
   let after_full_application = Continuation.create () in
+  let full_apply_result_arity =
+    Code_metadata.result_arity callee's_code_metadata
+  in
   let after_full_application_handler =
-    let func_param = BP.create func_var K.With_subkind.any_value in
-    Continuation_handler.create
-      (Bound_parameters.create [func_param])
-      ~handler:perform_over_application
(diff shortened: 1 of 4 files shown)


middle_end/flambda2/simplify/simplify_common.ml
```diff
@@ -241,13 +241,34 @@ let split_direct_over_application apply
...
+             (fun kind ->
+               Bound_parameter.create (Variable.create "over_app_result") kind)
+             (Flambda_arity.unarized_components full_apply_result_arity))
+      in
+      Continuation_handler.create params
+        ~handler:
+          (Expr.create_invalid
+             (Message
+                "Over-application of a function that does not return a \
+                 singleton value"))
```
mshinwell, line 262: Let's put the whole Apply_expr in the message.  Can we add a new Invalid constructor too?

mshinwell (changes requested):

lthls:
@mshinwell I've added a commit to handle the case that you reported today. If you want you can review it, although it might be simpler if @Ekdohibs reviews it instead.

mshinwell:
@lthls probably best if @Ekdohibs reviews that, thanks

Ekdohibs:
@lthls Your changes are good! (But I can't approve them, as I can't approve my own pull request)

lthls (approved):
Adding the official seal of approval (I've reviewed the earlier commits too).

mshinwell (approved):

PR #3884, opened 2025-04-17
Use [Are_rebuilding_terms.t] in DE instead of bool + other small changes to `ART` API

Gbury:
Also, remove the `of_bool` function in favour of dedicated constructors for values of type [ART.t], since they were only called with constant literals for now, and introduce a non-negated function to know whether we are rebuilding terms.

As a comment/CR states, now that we have functions./values to both create and inspect values of type `ART.t`, it could make more sens to have the `{do|do_not}_rebuild_terms` names used to create new values, and the `{are|are_not}_rebuilding_terms` names used to inspect the value to know what is the status of term rebuilding, but I did not change it as that would have make the diff bigger. That being said, if people agree that swapping the names makes sense, I'd be glad to do it.

Not a necessary patch, but seems cleaner (for the curious, this patch originally comes from the time when we considered a partial rebuilding upwards pass for match-in-match, which required `ART.t` to become a proper enum with 3 cases; that attempt was eventually abandoned in favor of the current approach, but some changes/patches to `ART` seemed interesting to keep).

### DIFF
--- a/middle_end/flambda2/simplify_shared/are_rebuilding_terms.mli
+++ b/middle_end/flambda2/simplify_shared/are_rebuilding_terms.mli
@@ -19,10 +19,12 @@
 
 type t
 
-val of_bool : bool -> t
-
-val do_not_rebuild_terms : t -> bool
-
-val are_rebuilding : t -> bool
-
 val print : Format.formatter -> t -> unit
+
+(* CR gbury: the terms/names used for creating values and for inspecting them
+    would make more sens if they were swapped. *)
+val are_rebuilding : t
+val are_not_rebuilding : t
+
+val do_rebuild_terms : t -> bool
+val do_not_rebuild_terms : t -> bool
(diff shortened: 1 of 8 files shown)


middle_end/flambda2/simplify_shared/are_rebuilding_terms.mli
```diff
@@ -19,10 +19,14 @@
 
 type t
 
-val of_bool : bool -> t
+val print : Format.formatter -> t -> unit
 
-val do_not_rebuild_terms : t -> bool
+(* CR gbury: the terms/names used for creating values and for inspecting them
+   would make more sens if they were swapped. *)
```
Ekdohibs, line 25: I second this comment, maybe do it?

Gbury: cc @mshinwell ?

Ekdohibs (approved):

PR #3886, opened 2025-04-17
Parallelism tutorial

lukemaurer:
Basic introduction to data-race-free parallelism, focusing on fork/join parallelism and data parallelism provided by the `parallel` library as a way to introduce the `portable` and `contended` modes.

Currently pretty drafty:
- [x] Needs an introduction
- [ ] Should probably be reorganized a bit—the section on `portable` and `contended` was supposed to be sort of an intermission but it's the biggest part
- [ ] Possibly a bit too formal in places? I think having numbered rules for things is good but then I've written conference papers
- [ ] Some of the example code could be more motivated: a `Thing.t` is just something with an immutable `price` and a mutable `mood`. (If nothing else, if my time at a trading firm has taught me anything, it's that prices are in fact mutable.)
- [ ] Other not-quite-finished bits


jane/doc/parallelism-tutorial.md
```diff
@@ -0,0 +1,1030 @@
...
+(We use `fold'` rather than `fold` to get the version that passes `par` down
+into our `f` and `combine` functions.)
+
+Another approach would be to convert the entire tree into a
+`Parallel.Sequence.t` at once rather than rely on nested parallelism. This is
+possible using `Parallel.Sequence.unfold`, though it does take more work,
+requiring a definition of an unfolding state with an operation that splits it in
+half.
+
+**[I made a solution for this but unless there's something I missed that would
```
goldfirere, line 1027: Hard for me to see exactly where this is going, so I don't have much of an opinion here. I do like the idea of generalizing the initial example to something a little harder.

jane/doc/parallelism-tutorial.md
goldfirere, line 1: I think folks reading this may also want to be able to observe the speedup. Though I don't think you should spend a lot of time explaining benchmarking or anything, I think it would be a valuable addition to show an actual executable that can be run to get timing measurements -- and then to show that parallelism can speed things up.

goldfirere, line 1: Might be nice to have a little summary at the end. This would include tips like "look at the Parallel library" or "benchmark after parallelizing". It could also maybe have all your Rules stated together in a little table.

goldfirere (commented):
Nice tutorial. I enjoyed reading it.

I wondered in one of my inline comments about whether we want all the general portability/contention stuff in this tutorial. But after reflecting about this with @ccasin, I think it's good. We'll have general documentation for this stuff. But pointing readers to it in the middle of the tutorial is annoying, and the general documentation will inevitably have more detail than is needed at this point. So the stuff here might be duplicative, but I think on balance it's good duplication.

Happy to chat about anything here if you'd like.

jane/doc/parallelism-tutorial.md
```diff
@@ -0,0 +1,1030 @@
...
+
+Of course, in this case what saves us is that it doesn't matter what order the
+updates happen in: addition is commutative. However, now that we've made `mood`
+atomic as well, suppose we wanted to take the average of all the nodes whose
+`mood` is `Happy`? Nothing in the system stops us now: if someone is calling
+`bum_out` on the entire tree in parallel, then our average will reflect an
+unpredictable number of those changes. The only way to stop _that_ is to use
+something more sophisticated like a lock over the whole tree, which grants a
+function `uncontended` access while the lock is held. That's what the capsule
+API is for, but it's beyond the scope of this tutorial.
```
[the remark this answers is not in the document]
TheNumbat: I wouldn't say memory corruption exactly since the memory model already prevents undefined behavior. Maybe state corruption?

jane/doc/parallelism-tutorial.md
```diff
@@ -0,0 +1,1030 @@
...
+
+<a id="code-average"></a>
+```ocaml
+module Tree = struct
+  type 'a t =
+    | Leaf of 'a
+    | Node of 'a t * 'a t
+end
+
+let average (tree : float Tree.t) =
```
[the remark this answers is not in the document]
lukemaurer: Adding a link to the paper (seems like the canonical documentation at the moment)

jane/doc/parallelism-tutorial.md
```diff
@@ -0,0 +1,1030 @@
...
+```
+
+All we have to do to sum over the prices in a `Thing.t Tree.t` is change the
+`Tree.Leaf` case:
+
+```ocaml
+  let average_things_par (par : Parallel.t) tree =
+    let rec total par tree =
+      match tree with
+      | Tree.Leaf x -> ~total:(Thing.price x), ~count:1 (* <== new code *)
```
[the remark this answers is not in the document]
lukemaurer: Trying it as a diff with two lines of context. Seems good

jane/doc/parallelism-tutorial.md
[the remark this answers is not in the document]
goldfirere: What paper? This is about labeled tuples, which I don't think are in a paper...

jane/doc/parallelism-tutorial.md
```diff
@@ -0,0 +1,1030 @@
...
+      ~total:(total_l +. total_r), ~count:(count_l + count_r)
+  in
+  let ~total, ~count = total tree in
+  total /. (count |> Float.of_int)
+;;
+```
+
+Now suppose we want to take advantage of multicore OCaml and get those sweet,
+sweet parallel speedups. This is an obvious candidate for _fork/join
+parallelism,_ as provided by the `parallel` library’s `fork_join*` functions.
```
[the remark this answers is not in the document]
goldfirere: As on Slack: sadly, no. But we can link to the mli file.

jane/doc/parallelism-tutorial.md
```diff
@@ -0,0 +1,1030 @@
...
+
+Adding `portable` and `contended` to your .mli files can take some work, but
+over the next few sections we'll cover what exactly the modes mean, what the
+most important rules for using them are, and some workarounds and shortcuts. In
+return, these two modes suffice to run even large amounts of code in fork/join
+style without fear of data races.
+
+## How `portable` and `contended` work
+
+Firstly, `portable` and `contended` are _modes_. Like a type, a mode describes
```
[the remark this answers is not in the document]
lukemaurer: Yeah, I've tried to keep this section lean (emphasis on “tried”) in part for this reason. I do think there's value in going over this stuff even if there's overlap with other documentation.

jane/doc/parallelism-tutorial.md
```diff
@@ -0,0 +1,1030 @@
...
+let price_and_mood (t @ uncontended) = price t, mood t
+```
+
+Can we really treat `t` as both `contended` (to call `price`) and `uncontended`
+(to call `mood`)? Actually, we can:
+
+<a id="rule-contended-submode"></a>
+> **Rule 3.** An `uncontended` value may be used as though it is `contended`.
+
+This is entirely consistent with rules 1 and 2: having an `uncontended` value
```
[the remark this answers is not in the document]
lukemaurer: Hmm. Really there are two audiences, though, and for the Jane Street audience the chances are much higher that they'll have seen it, and it seems making the connection. Maybe if I move the parenthetical sentence here to a footnote? That would more strongly suggest that it's not important to understand. Or I could keep it in place so that people see it but phrase it as “If you've worked with `local`, you may have seen that it works the same way.”

jane/doc/parallelism-tutorial.md
```diff
@@ -0,0 +1,1030 @@
...
+notation called a _modality_ to make a field `contended` instead:
+
+[Rule 4]: #rule-contended-deep
+[rule 3]: #rule-contended-submode
+
+```ocaml
+module State = struct
+  type t = {
+    mutable most_expensive_thing : Thing.t @@ contended
+    mutable my_favorite_thing : Thing.t
```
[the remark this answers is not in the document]
lukemaurer: Urk, I think I must have had this realization in my test code but failed to transfer that knowledge to the tutorial. (My test code has the record but neither field is `mutable`.)

jane/doc/parallelism-tutorial.md
```diff
@@ -0,0 +1,1030 @@
...
+causing the same data race we've been trying to avoid all along.
+
+In fact, rule 4 is pretty well required to make [rule 1] work at all. The
+interested reader is invited to work out why the text of rule 1 demands that,
+for instance, we say that `t_in_a_trenchcoat.inner_t` is `contended` when
+`t_in_a_trenchcoat` is `contended`.
+
+[rule 1]: #rule-contended-parallel
+
+#### Modalities
```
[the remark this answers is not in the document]
lukemaurer: Sorry, not sure what general treatment you're referring to? I tend to agree, we can probably lose it. The one thing that makes me think otherwise is that I explain at some length why, say, a `contended` thing can _only ever_ contain `contended` things, and then it's valuable to say that `uncontended` things only contain `uncontended` things _by default._ But we probably don't need the whole song and dance, especially if I can link somewhere else for the details.

jane/doc/parallelism-tutorial.md
```diff
@@ -0,0 +1,1030 @@
...
+`immutable_data`. Also like types, kinds are inferred where possible, so in
+fact the `: immutable_data` in this case is redundant. However, in an .mli
+they are often useful.
+
+The three most important kinds for data-race freedom are `immutable_data`,
+`mutable_data`, and `value`.
+
+| Kind | Requirements | Crosses |
+| ---- | ------------ | ------- |
+| `immutable_data` | no functions or mutable fields, deeply | portability, contention |
```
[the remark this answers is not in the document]
lukemaurer: Ah, good point. I've added a blurb about not being exhaustive—we could tease the names of other axes here, but the nice thing about the table as is is that the reader might hope to explain the whole table to themselves. (Also if I make the table exhaustive it will become outdated every month or two.)

Am I right that `value` truly crosses nothing? If there's something it doesn't cross, than I should write “neither” rather than “none”.

jane/doc/parallelism-tutorial.md
```diff
@@ -0,0 +1,1030 @@
...
+  let mood { mood; _ } = Atomic.get mood
+  let cheer_up { mood; _ } = Atomic.set mood Happy
+  let bum_out { mood; _ } = Atomic.set mood Sad
+end
+```
+
+As you can see, there's a bit of syntactic overhead, but in return we get to
+access `mood` even if a `Thing.t` is `contended`, and in fact we can mark `t` as
+`immutable_data` so that it ignores `contended` and `uncontended` altogether
+(that is, it [crosses contention]). Do go over the documentation for the `Atomic`
```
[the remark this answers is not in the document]
lukemaurer: Urk. Well, I can link to the stdlib documentation, which doesn't have the right modes and isn't quite the same as Base/Core. Leaving this dangling for the moment

jane/doc/parallelism-tutorial.md
```diff
@@ -0,0 +1,1030 @@
...
+
+If we kept `total` and `count` in `ref`s, then `go` would not be able to access
+them. (Exercise: What rules combine to stop us? Remember, a `ref` is just a
+record whose only field is `mutable`. Don't rely on rule 1 of `contended` or
+rule 1 of `portable` for your answer—those only tell us that we _must be
+stopped,_ not what actually stops us.)
+
+Now for the caveats: Firstly, there are performance penalties, since an
+`Atomic.t` is a pointer and atomic operations are more expensive. Secondly,
+`Atomic.t` frees us from concerns about data races, but it absolutely does not
```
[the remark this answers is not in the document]
lukemaurer: Yeah, this is def one of the things that needs to be in the introduction

jane/doc/parallelism-tutorial.md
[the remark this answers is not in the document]
lukemaurer: Hmm. I'm learning that my belief in what DRF prevents was imprecise. I'll ask Stephen when he's in tomorrow

jane/doc/parallelism-tutorial.md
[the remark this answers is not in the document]
goldfirere: Today, `value` crosses nothing. But it's still `mod non_null separable` (where the `separable` part comes with the next compiler roll). But those things aren't modes, so your statement that `value` crosses nothing is still valid.

jane/doc/parallelism-tutorial.md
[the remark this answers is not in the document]
goldfirere: Just link to the Base/Core mli file, I think.

jane/doc/parallelism-tutorial.md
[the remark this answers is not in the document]
goldfirere: Good point about memory corruption. I had been neglecting that fact about the memory model. Maybe our type system only prevents memory corruption in the presence of unboxed products like internal pointers? That is, a data race on a mutable unboxed product -- where one element in the product determines the validity of another -- can cause arbitrary memory corruption. But OCaml does not have such products natively. We're about to add them -- in the next week or two, according to @ccasin. So once we have such things, despite the guarantees the memory model gives us, then we'll have the possibility for memory corruption. And our type system prevents it. We both giveth and taketh away.

Neglecting this corner case (around dependent unboxed products), I'm honestly not sure what we gain with DRF.

jane/doc/parallelism-tutorial.md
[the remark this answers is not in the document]
lukemaurer: Well, there's at least this paper (workshop paper I think?): https://tyconmismatch.com/papers/ml2024_labeled_tuples.pdf

jane/doc/parallelism-tutorial.md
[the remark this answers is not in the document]
lukemaurer: Removed both “Modalities” sections for now.

jane/doc/parallelism-tutorial.md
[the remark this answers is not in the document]
lukemaurer: Added a whole section to the beginning about this, having talked with Stephen. The headline is we get sequential consistency (but of course I have to explain what that is).

jane/doc/parallelism-tutorial.md
```diff
@@ -1,38 +1,162 @@
...
+includes many parts, from common primitives like atomics and locks to newer
+features like modes and capsules. Fortunately, not every parallel program will
+need all of this machinery, and following this tutorial, you'll see how far you
+can get with just two modes and the `Parallel.fork_join2` function.
+
+## What is a data race?
+
+A data race is a race condition in which two accesses to the same memory
+location conflict. Specifically, there are four elements of the crime: **[[Oops,
+I forgot to define “domain” first]]**
```
goldfirere, line 24: Yes, maybe write a little section before this one entitle "A very brief introduction to multicore OCaml" or something

jane/doc/parallelism-tutorial.md
```diff
@@ -1,38 +1,162 @@
...
+`initialised` is set first, _while `price_of_gold` is still zero,_ leading
+domain B to make an expensive mistake.
+
+[atomic]: #atomics
+
+It turns out our intuition assumes what the literature calls
+_sequential consistency:_ essentially, the principle that things happen in order
+they're written, and parallelism just means we're not sure which domain will do
+its next thing. This principle is so essential that it's hard to imagine
+reasoning about a multicore program without it.
```
goldfirere, line 127:
```suggestion
reasoning about a multicore program without it. And yet, when a data race occurs,
sequential consistency goes out the window: we can observe mutations (and other
side effects) *completely out of order*.
```

Or something. I would love to incorporate the vision in my head of an oxcaml (the animal) wrestling the dog-prince of chaos to the ground and putting a leash on it, but I don't know how to do so. Maybe AI can help? Or maybe James Somers can, if he enjoys a little creative writing.

jane/doc/parallelism-tutorial.md
```diff
@@ -1,38 +1,162 @@
...
+
+Of course, what's happened here is two data races: each of `price_of_gold` and
+`initialised` is being modified by domain A and read by domain B in parallel.
+So the moral of the story is that data races violate sequential consistency.
+
+The good news is that the inverse is also true: any program without data races
+is sequentially consistent. Data-race freedom gives us back the power to reason
+intuitively about our code, no matter how buggy it might get. In our example,
```
goldfirere, line 135:
```suggestion
is sequentially consistent. *Data-race freedom gives us back the power to reason
intuitively about our code, no matter how buggy it might get.* In our example,
```

jane/doc/parallelism-tutorial.md
```diff
@@ -1,38 +1,162 @@
...
+that means that once we fix the data races (easy to do with `Atomic.t`), we know
+that fixes our bug: sequential consistency means that it's impossible for domain
```
goldfirere, line 137:
```suggestion
our bug is gone: sequential consistency means that it's impossible for domain
```

jane/doc/parallelism-tutorial.md
```diff
@@ -1,38 +1,162 @@
...
+B to observe a state where `initialised` is `true` and `price_of_gold` is
+`0.0`.[^price-zero]
+
+[^price-zero]: Assuming, of course, `calculate_price_of_gold ()` doesn't
+return `0.0`. The compiler can't catch everything.
+
+The even _better_ news is that, in code using OxCaml's extensions, the data
```
goldfirere, line 144: Great para.

jane/doc/parallelism-tutorial.md
```diff
@@ -1,38 +1,162 @@
...
+to understand the type system, the system for data-race freedom has a learning
+curve. This tutorial aims to get you moving with a practical framework that
+hopes to achieve significant parallel speedups without much fuss. You'll still
+need to convince the compiler that your code has no data races, but we'll walk
+through the rules and even some [shortcuts](#amenities) for common special
+cases.
+
+# Fork/join parallelism
+
+Any problem that can be neatly subdivided is a good candidate for
```
goldfirere, line 156: "subdivided into parallel pieces"? But that makes the sentence a little circular. I mean, *any* problem can be subdivided. So I think this isn't exactly what you mean (but fail to have a better phrasing at the moment).

jane/doc/parallelism-tutorial.md
```diff
@@ -1,38 +1,162 @@
...
+_fork/join parallelism,_ as provided by the `parallel` library’s `fork_join*`
+functions. For example, a very expensive way to add four integers (I promise the
+other examples are more substantial) would be this:
```
goldfirere, line 159: I love the parenthetical. But it also begs the question: who is "I"? Are we planning on putting your name at the top of this? Perhaps we should; I see no reason not to. If not, though, maybe switch to "we".

jane/doc/parallelism-tutorial.md
```diff
@@ -77,36 +197,71 @@ let%expect_test "add4 in parallel" =
...
+  in
+  let ~total, ~count = total tree in
+  total /. (count |> Float.of_int)
+;;
+```
+
+(Note that we've made use of the new [labeled tuples] feature for the values
+returned by `total`.)
+
+[labeled tuples]: https://tyconmismatch.com/papers/ml2024_labeled_tuples.pdf
```
goldfirere, line 236: I think it's better to point to the documentation at oxcaml.org.

jane/doc/parallelism-tutorial.md
```diff
@@ -77,36 +197,71 @@ let%expect_test "add4 in parallel" =
...
+      let ~total:total_r, ~count:count_r = total r in
+      ~total:(total_l +. total_r), ~count:(count_l + count_r)
+  in
+  let ~total, ~count = total tree in
+  total /. (count |> Float.of_int)
+;;
+```
+
+(Note that we've made use of the new [labeled tuples] feature for the values
+returned by `total`.)
```
goldfirere, line 234: I would move this note above the code. Otherwise, readers' OCaml parsers will fail and they may not make it here.

jane/doc/parallelism-tutorial.md
```diff
@@ -77,36 +197,71 @@ let%expect_test "add4 in parallel" =
...
+<a id="code-average"></a>
+```ocaml
+module Tree = struct
+  type 'a t =
+    | Leaf of 'a
+    | Node of 'a t * 'a t
+end
+
+let average (tree : float Tree.t) =
+  let rec total tree : total:float * count:int =
```
goldfirere, line 220:
```suggestion
  let rec total tree : (total:float * count:int) =
```

The parens are redundant, of course, but given the other `:` on the line, I think it makes the code easier to read.

jane/doc/parallelism-tutorial.md
```diff
@@ -133,27 +288,17 @@ end
...
-            (fun par -> total par l)
-            (fun par -> total par r)
-        in
-        ~total:(total_l +. total_r), ~count:(count_l + count_r)
-    in
-    let ~total, ~count = total par tree in
-    total /. (count |> Float.of_int)
-  ;;
+```diff
+  let rec total par tree : total:float * count:int =
```
goldfirere, line 292: Did you mean the change the name of the function too?

jane/doc/parallelism-tutorial.md
```diff
@@ -394,90 +563,40 @@ for instance, we say that `t_in_a_trenchcoat.inner_t` is `contended` when
...
-Note that there is actually an `@@ uncontended` modality but it's not useful—if
-the record is `contended`, the field has to be `contended` no matter what (see
-[rule 4]), so the modality doesn't change anything there. Fortunately, by rule
-3, you can _always_ put an `uncontended` value in a `contended` record, so the
-modality isn't even necessary there.
+Shouldn't the returned value be `contended`? In other circumstances it would: if
+you try changing the `float` in the definition of `t` to a `float ref` then
+you'll find you indeed have to add `contended` to the return type of `price`.
+We'll cover the details in the section on [mode crossing], but the short answer
+is that since `float`s are immutable, no data race can arise from them, so the
```
goldfirere, line 583:
```suggestion
is that since `float`s are deeply immutable, no data race can arise from them, so the
```

jane/doc/parallelism-tutorial.md
```diff
@@ -394,90 +563,40 @@ for instance, we say that `t_in_a_trenchcoat.inner_t` is `contended` when
...
+compiler simply ignores `contended` and `uncontended` on `float`s (and you can
+make use of this for your own immutable data as well).
```
goldfirere, line 585:
```suggestion
make use of this for your own deeply immutable data as well).
```

jane/doc/parallelism-tutorial.md
```diff
@@ -609,47 +729,33 @@ And also for similar reasons as before, we need `portable` to be deep the way
...
-requirements altogether for values of `t`. We'll cover the specifics when we get
-to [mode crossing].
+You may have been wondering why rules 1 and 2a restrict all values, when really
+it's only functions that we care about being `portable` or
+not.[^other-code-types] This is the reason: many values can _contain_ functions,
+and if it weren't for rule 4, we could use, say, a record to smuggle the
+function around (in other words, the record could be a function [in a
+trenchcoat]). So we have to forbid _all accesses_ of `nonportable` values from
+other domains, not just function calls. The good news is that many types (in
+fact, most types) obviously _can't_ cause a problem—you'll never make an `(int *
```
goldfirere, line 739: Wrong kind of dash?

jane/doc/parallelism-tutorial.md
```diff
@@ -609,47 +729,33 @@ And also for similar reasons as before, we need `portable` to be deep the way
...
+string option) list` that contains _any_ function, much less one that will cause
+a data race—and if the compiler is aware that a type `t` can't have a function
```
goldfirere, line 741: Other end of "wrong kind of dash?"

jane/doc/parallelism-tutorial.md
```diff
@@ -711,7 +817,10 @@ fact the `: immutable_data` in this case is redundant. However, in an .mli
 they are often useful.
 
 The three most important kinds for data-race freedom are `immutable_data`,
-`mutable_data`, and `value`.
+`mutable_data`, and `value`. Some kinds constrain what types they describe, and
+in return types with those kinds get to cross certain modes. In summary (this
+table isn't nearly exhaustive—in particular, there are many modes we haven't
+covered, and `immutable_data` and `mutable_data` cross some of them as well):
```
goldfirere, line 823: Link to the now-extant kinds documentation somewhere around here.

jane/doc/parallelism-tutorial.md
```diff
@@ -738,70 +847,44 @@ let always_portable : int_or_string @ nonportable -> int_or_string @ portable =
...
-  :  t @ local
-  -> (t @ local -> 'a @ contended portable) @ portable
-  -> (t @ local -> 'b @ contended portable) @ portable
-  -> 'a * 'b @ contended portable
-```
+In fact, we've already seen mode crossing in action. We mentioned before when
+talking about [rule 4 of `contended`](#rule-contended-deep) that when
+`Thing.price` projects the `price` field out of a `contended` `Thing.t`, the
+only reason it gets to return an `uncontended` `float` is that `float` crosses
+contention (which is to say, it has kind `immutable_data`).
```
goldfirere, line 854: Groan. Very sadly, `float : immutable_data` no longer holds. The problem is that `immutable_data` expands to `value mod ... non_float`. So, yeah, not `float`.

jane/doc/parallelism-tutorial.md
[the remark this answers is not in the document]
lukemaurer: Ah, that would be better.

jane/doc/parallelism-tutorial.md
```diff
@@ -0,0 +1,1030 @@
...
+
+> <a id="rule-contended-parallel"></a>
+> **Rule 1.** If multiple accesses of the same value may occur in parallel, at
+> most one of them may consider the value `uncontended`. The others must
+> consider it `contended`.
+
+> <a id="rule-contended-mutable"></a>
+> **Rule 2.** Reading or modifying a `mutable` field is not allowed if the
+> record is `contended`. The same goes for any element of a `contended` array
+> (unless it's an `iarray`, since those are immutable).
```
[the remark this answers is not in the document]
lukemaurer: Decided to let the typography speak for itself and add a short subsection called “Immutable Arrays” after “Atomics.” (Obviously iarrays aren't DRF-specific, but they're still a feature that makes DRF life easier.)

jane/doc/parallelism-tutorial.md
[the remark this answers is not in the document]
lukemaurer: Downgrading the passage here to a footnote. I doubt writing for the non-JS audience would annoy the JS audience, I'd just hate to lose the opportunity to leverage something the reader might already know

jane/doc/parallelism-tutorial.md
```diff
@@ -0,0 +1,1030 @@
...
+```
+
+Once you're familiar with the `portable` and `contended` modes, you'll see that
+this can be read as saying:
+
+> The `price` function is safe to call from any domain, and it won't produce a
+> data race even if its argument is mutated in parallel.
+
+That's enough to let `average_par_things` go through: from there, the compiler
+infers the safety properties that `Parallel.fork_join2` demands.
```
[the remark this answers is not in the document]
lukemaurer: Rewrote this to be more specific about the safety properties (while still being vague about how `contended` and `portable` get us there). For the counterexample I change `price` to be `mutable` and show that the accessor no longer typechecks.

jane/doc/parallelism-tutorial.md
```diff
@@ -1,38 +1,162 @@
...
+# Simple Parallelism Without Data Races
 
 # Introduction
 
-# Fork/join parallelism
+OCaml 5 unleashed parallel programming for the OCaml ecosystem. As is often the
+case when something is unleashed, this has quite a bit of potential to cause
+chaos. **[Need to quit fussing with this so I'm leaving it as is for the moment,
+but am I saying we're putting the leash back on? Leashes are good, actually? Or
+maybe we're putting up a fence so the the multicore dog can run around without
```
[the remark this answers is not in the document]
lukemaurer: Decided in the end to keep the line but not extend the metaphor.

jane/doc/parallelism-tutorial.md
```diff
@@ -1,38 +1,162 @@
...
+data races biting people??]**
 
-Suppose we're working with binary trees, and we want to take an average over all
-the values in the tree. Here's a basic implementation:
+OxCaml aims to provide an infrastructure for multicore OCaml programming that
+exploits the full power of parallelism while _guaranteeing_ freedom from an
+especially pernicious threat, that of the _data race._ That infrastructure
+includes many parts, from common primitives like atomics and locks to newer
+features like modes and capsules. Fortunately, not every parallel program will
+need all of this machinery, and following this tutorial, you'll see how far you
```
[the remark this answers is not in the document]
lukemaurer: Hmm, I should be more precise here: I mean that some parallel programs only need _some_ of this machinery.

jane/doc/parallelism-tutorial.md
lukemaurer: Hmm. It's actually not quite that bad, though? The data races are, after all, bounded in time and space. Maybe I don't need to split those hairs for this, though?

jane/doc/parallelism-tutorial.md
lukemaurer: Oh heh yeah, switching to “we.”

jane/doc/parallelism-tutorial.md
lukemaurer: Yeah, I struggled a bit too. The word _neatly_ is sort of standing in for the idea that the pieces have to be separate but it's not very precise.

jane/doc/parallelism-tutorial.md
lukemaurer: I've changed the references to keep using the old name.

jane/doc/parallelism-tutorial.md
lukemaurer: I thinks “floats have no mutable parts” is a bit more obvious (keeping “deeply immutable” for the next one, hoping the meaning is clear from the first part)

jane/doc/parallelism-tutorial.md
lukemaurer: I think an em dash is correct here? Used parenthetically?

jane/doc/parallelism-tutorial.md
lukemaurer: That's absolutely _tragic._

lukemaurer: Thought about hand-waving this a bit but I think it's easiest just to delete the parenthetical. There's already the blurb after the table that introduces the idea that some special types don't have these exact kinds but do cross those modes.

jane/doc/parallelism-tutorial.md
```diff
@@ -0,0 +1,1093 @@
...
+provided by a library and determines the policy by which tasks get doled out and
+run in domains. For this tutorial, we'll use `parallel_scheduler_work_stealing`,
+which implements the popular [work-stealing] strategy.
+
+[work-stealing]: https://en.wikipedia.org/wiki/Work_stealing
+
+```ocaml
+let test_add4 par = add4 par 1 10 100 1000
+
+let%expect_test "add4 in parallel" =
```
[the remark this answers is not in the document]
lukemaurer: Ah, good point. Probably fine to just use a top-level printf here and put the result of execution in the next text box?

jane/doc/parallelism-tutorial.md
```diff
@@ -0,0 +1,1093 @@
...
+val create : price:float -> mood:Mood.t -> t
+val price : t -> float
+val mood : t -> Mood.t
+```
+
+But now we get an error from the compiler:
+
+```
+The value Thing.price is nonportable, so cannot be used inside a function that
+is portable.
```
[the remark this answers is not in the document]
lukemaurer: Whoops, yeah I lost track that I'd already added a `@ portable` in my test code

jane/doc/parallelism-tutorial.md
```diff
@@ -0,0 +1,1093 @@
...
+  module Mood = struct
+    type t =
+      | Happy
+      | Neutral
+      | Sad
+  end
+
+  type t =
+    { price : float
+    ; mutable mood : Mood.t
```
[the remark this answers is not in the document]
lukemaurer: Ah, thanks—we seem to have that warning turned off

jane/doc/parallelism-tutorial.md
```diff
@@ -0,0 +1,1093 @@
...
+
+[in a trenchcoat]: #code-cheer_up_sneakily
+[mode crossing]: #mode-crossing
+
+We can summarize the rules of `portable` and `contended` like so:
+
+|        | `contended` | `portable` |
+|--------|-------------|------------|
+| Rule 1 | No parallel `uncontended` accesses | Only `portable` values may cross domains |
+| Rule 2 | No accessing `contended` `mutable` state | A `portable` function only sees `portable` `contended` values |
```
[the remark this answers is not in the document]
lukemaurer: The only analogy really is that rule 1 is a global soundness property and rule 2 is a typing rule. I wasn't sure there's a good way to get that across without straying into POPL territory, though

jane/doc/parallelism-tutorial.md
```diff
@@ -0,0 +1,1093 @@
...
+let beat_the_system par =
+  let t = { price = 42.0; mood = Neutral }
+  let (), () =
+    Parallel.fork_join2 par (fun _par -> cheer_up t) (fun _par -> bum_out t)
+  in
+  ()
+```
+
+```
+Error: This value is contended but expected to be uncontended.
```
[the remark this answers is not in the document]
lukemaurer: Oh! I never did say to add `@@ portable` to those functions, so that makes sense.

jane/doc/parallelism-tutorial.md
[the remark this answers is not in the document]
goldfirere: I think a footnote makes sense. And actually I wouldn't assume the JS audience knows about locality either. Most devs I speak with don't. They've vaguely heard of it, but have no idea what it means.

jane/doc/parallelism-tutorial.md
```diff
@@ -0,0 +1,1228 @@
...
+```
+
+(We use `fold'` rather than `fold` to get the version that passes `par` down
+into our `f` and `combine` functions.)
+
+Another approach would be to convert the entire tree into a
+`Parallel.Sequence.t` at once rather than rely on nested parallelism. This is
+possible using `Parallel.Sequence.unfold`, though it does take more work,
+requiring a definition of an unfolding state with an operation that splits it in
+half.
```
goldfirere: A conclusion -- reminding the reader of what they've learned -- would be nice.

jane/doc/extensions/_01-tutorials/01-intro-to-parallelism-part-1.md
```diff
@@ -0,0 +1,1353 @@
...
+[defaulting to `portable`]: #defaulting-to-portable
+
+Since `Thing.t` is now `mutable_data`, every `Thing.t` is implicitly assumed to
+be `portable`.
+
+Any type variable can be given a kind, so we can write a version of
+`always_portable` that works for _any_ `mutable_data` type:
+
+```ocaml
+let always_portable' (a : ('a : mutable_data) @ nonportable)
```
mdelvecchio-jsc, line 1130: I think we generally prefer to use
```
let always_portable' (type a : mutable_data) (a : a @ nonportable) : a @ portable
```
and 
```
val always_portable' : ('a : mutable_data). 'a @ nonportable -> 'a @ portable
```

jane/doc/extensions/_01-tutorials/01-intro-to-parallelism-part-1.md
```diff
@@ -0,0 +1,1353 @@
...
+ignoring those modes. In summary (this table isn't nearly exhaustive—see the
+[documentation on kinds] for many more modes and what kinds cross them):
+
+[documentation on kinds]: ../_06-kinds/intro
+
+| Kind | Requirements | Crosses |
+| ---- | ------------ | ------- |
+| `immutable_data` | no functions or mutable fields, deeply | portability, contention |
+| `mutable_data` | no functions, deeply | portability |
+| `value` | none | none |
```
mdelvecchio-jsc, line 1066: Are we purposefully omitting `value mod contended`, `value mod portable`, and `value mod contended portable` here?

jane/doc/extensions/_01-tutorials/01-intro-to-parallelism-part-1.md
lukemaurer: That makes sense, but I'm not sure I want to introduce explicit quantification syntax here.

jane/doc/extensions/_01-tutorials/01-intro-to-parallelism-part-1.md
lukemaurer: Yes. I don't think we want to go into the full generality here, just the most useful few kinds to know about.

jane/doc/parallelism-tutorial.md
lukemaurer: Ran out of time for this. Keeping in mind for another PR

lukemaurer: Did end up making a table. A conclusion would be good but can be added later.

PR #3887, opened 2025-04-17
Stable ordering for bound continuations

Gbury:
This PR changes the representation of bound continuations to use a list (`Lmap`) instead of a regular map. This is done so that we can ensure that an (arbitrary) order of the bound continuations is kept even through renaming and other transformations. This is a pre-requisite for the replay_history (which itself is a pre-requisite of match-in-match).

### DIFF
--- a/middle_end/flambda2/compare/compare.ml
+++ b/middle_end/flambda2/compare/compare.ml
@@ -1300,7 +1300,7 @@ and let_cont_exprs env (let_cont1 : Let_cont.t) (let_cont2 : Let_cont.t) :
       ~f:(fun ~invariant_params ~body1 ~body2 cont_handlers1 cont_handlers2 ->
         pairs ~f1:exprs ~f2:compare_handler_maps
           ~subst2:(fun env map ->
-            Continuation.Lmap.map (subst_cont_handler env) map)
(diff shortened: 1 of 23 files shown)


middle_end/flambda2/compare/compare.ml
```diff
@@ -1292,15 +1292,15 @@ and let_cont_exprs env (let_cont1 : Let_cont.t) (let_cont2 : Let_cont.t) :
...
+        (map1 |> Continuation.Lmap.bindings)
+        (map2 |> Continuation.Lmap.bindings)
+      |> Comparison.map ~f:Continuation.Lmap.of_list
     in
     Recursive_let_cont_handlers.pattern_match_pair handlers1 handlers2
       ~f:(fun ~invariant_params ~body1 ~body2 cont_handlers1 cont_handlers2 ->
         pairs ~f1:exprs ~f2:compare_handler_maps
           ~subst2:(fun env map ->
-            Continuation.Map.map_sharing (subst_cont_handler env) map)
+            Continuation.Lmap.map (subst_cont_handler env) map)
```
Ekdohibs, line 1303:
```suggestion
            Continuation.Lmap.map_sharing (subst_cont_handler env) map)
```

Ekdohibs (approved):

PR #3896, opened 2025-04-18
Refactor amd64 simd

gretay-js:
Follow-up to #3853 refactoring. I think it will make the code easier to read and maintain:
- move simdgen to `backend/`
- remove simd_proc from amd64
- refactor `Simd.operation` and `Simd.Seq` to make selection and emit clearer.
- format another file
- enable more warnings


gretay-js:
superseded by https://github.com/ocaml-flambda/flambda-backend/pull/3907

PR #3898, opened 2025-04-18
Register classes

xclerc:
(Only for CI at this point.)


backend/arm64/reg_class.ml
```diff
@@ -0,0 +1,265 @@
...
+       95
+    |]
+
+  let dwarf_register_numbers reg_class =
+    match reg_class with
+    | Int64 -> int_dwarf_reg_numbers
+    | Float128 -> float_dwarf_reg_numbers
+
+  let int_reg_name =
+    [| "x0";
```
gretay-js, line 96: Looks like the formatting of these arrays was lost in the move. It'd be nice ot keep the formatting for not, it helps understand the mapping of indexes.

gretay-js (approved):
Looks good to me. 

For later: we should consider renaming `Reg_class` to something like `Reg_classes` or splitting it up to make improve things like `Reg_class.register_name typ reg` . Also there are a couple of copies of the same functionality in dwarf parts that compare registers using types and indexes that can be factored out.

backend/arm64/reg_class.ml
xclerc: (locally disabled formatting in a couple of places.)

PR #3899, opened 2025-04-18
Ensure register location equality is correct

xclerc:
This pull request changes the `Reg.same_loc`
function so that it is no longer error-prone, by
comparing not only the location, but also the
stack and register classes.

I am not 100% sure it is exactly the right
implementation, but I think it is sound and
can potentially void mistakes.


gretay-js (commented):
Looks like it would be useful to have another version of `Reg.same_loc` that raises on `Unknown`.

gretay-js (approved):

PR #3905, opened 2025-04-21
Fix use of -eliminate-dead-trap-handlers

gretay-js:
Missing `not` when the flag is used.


xclerc (approved):

gretay-js:
The ci failure is another instance of early failure in `warnings.ml`, unlikely caused by this PR.

PR #3768, opened 2025-03-31
Cfg_simplify pass

gretay-js:
- Add a new Cfg_simplify pass and move into it all passes that simplify CFG control flow:
  - eliminate_fallthrough_blocks
  - merge_straightline_blocks
  - eliminate_dead_blocks
  - disconnect_block
  
  There is no change in the pass that removes unused pure instructions (confusingly named `cfg_deadcode`).

- Delete dead trap handlers from CFG and remove the associated Pushtrap/Poptrap instructions.
- Delete unused `dead` field from the CFG and unused tracking of new labels.

PR #3909, opened 2025-04-22
Fix stack offsets

gretay-js:
Fix a bug introduced in PR #3768: `stack_offset` of blocks and individual instructions needs to be updated after eliminating dead trap handlers and removing the corresponding push/pop trap instructions. 

The new functionality is not reusing the code of `Selection_utils.Stack_offset_and exn.update_cfg` which makes special assumption about the input CFG not yet fully constructed (missing `exn` successorts and `stack_offset` fields are still `invalid`). It's unfortunate to have duplication of this subtle code, but the plan is to simplify the code in selection or remove entirely.  

This PR contains minor cleanup and refactoring of the code in selection that can be split into a separate PR or skipped entirely.

### DIFF
--- a/backend/cfg/cfg.ml
+++ b/backend/cfg/cfg.ml
@@ -575,6 +575,8 @@ let reset_instr_id () = InstructionId.reset instr_id
 
 let next_instr_id () = InstructionId.get_and_incr instr_id
 
+let invalid_stack_offset = -1
+
 let make_instr desc arg res dbg =
   { desc;
     arg;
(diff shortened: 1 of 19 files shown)


backend/cfg/cfg.ml
```diff
@@ -703,3 +667,98 @@ let equal_irc_work_list left right =
...
+    | Call_no_return _ | Call _ | Prim _ ->
+      ());
+    update_instruction terminator ~stack_offset
+  and update_block label ~stack_offset =
+    let block = get_block_exn t label in
+    if Label.Set.mem label !visited
+    then assert (block.stack_offset = stack_offset)
+    else visited := Label.Set.add label !visited;
+    if not (Int.equal block.stack_offset stack_offset)
+    then block.stack_offset <- stack_offset;
```
xclerc, line 731: (For symmetry, shouldn't it go
through `set_stack_offset`?)

xclerc (approved):

PR #3913, opened 2025-04-22
Tweak condition forcing linscan (register allocation)

xclerc:
In addition to the existing condition, this pull
request adds one related to the number of
registers.

If the number of registers is above a threshold
(defaulting to `100_000`, configurable through
a new `-regalloc-linscan-threshold` command-line
flag) then linscan will be used.


asmcomp/asmgen.ml
```diff
@@ -156,8 +156,13 @@ let write_ir prefix =
 
 let should_emit () = not (should_stop_after Compiler_pass.Linearization)
 
-let should_use_linscan fun_codegen_options =
-  !use_linscan || List.mem Cmm.Use_linscan_regalloc fun_codegen_options
+let should_use_linscan fd =
+  !use_linscan
+  || List.mem Cmm.Use_linscan_regalloc fd.fun_codegen_options
+  || List.compare_length_with
+       (Reg.all_relocatable_regs ())
```
gretay-js, line 163: this is a bit tricky:
- if this is called too early, it won't give the right results (e.g., right after reset, before selection)
- current implementation of `Reg.all_relocatable_regs` may be traversing a rather long list, even when using `List.compare_length_with`. It doesn't matter because any register allocator will traverse this list again anyway.

Both of the above are worth a note somewhere at least.

driver/flambda_backend_flags.ml
```diff
@@ -17,6 +17,7 @@ let use_ocamlcfg = ref true             (* -[no-]ocamlcfg *)
 let dump_cfg = ref false                (* -dcfg *)
 let cfg_invariants = ref false          (* -dcfg-invariants *)
 let regalloc = ref ""                   (* -regalloc *)
+let regalloc_linscan_threshold = ref 100_000 (* -regalloc-linscan-threshold *)
```
gretay-js, line 20: It'd be nice to have `Flambda_backend_flags.default_regalloc_linscan_threshold` and print it in the command line option description because the units are somewhat mysterious (similar to `default_heap_reduction_threshold` for example).

gretay-js (approved):

driver/flambda_backend_flags.ml
xclerc: done

asmcomp/asmgen.ml
xclerc: I have added a note just before the function,
but it is "safe" in the sense that if called between
reset and selection, the "old" result will be returned.

xclerc:
Not really, on the couple of cases I looked at
(where IRC was significantly slower, we are
talking more than one order of magniture)
the number of register was above 100,000.

gretay-js (approved):

PR #3915, opened 2025-04-22
No unboxed versions for mixed-float records

rtjoa:
As mixed-block records containing only `float` and`float#` store floats flatly, they shouldn't get implicit unboxed versions if we want to support layout-directed `%box` in the future. (For this reason, we already chose not to give float records unboxed versions.)


ccasin (approved):
LGTM. Thanks!

PR #3916, opened 2025-04-23
Revert "Disable macOS on x86"

spiessimon:
Reverts ocaml-flambda/flambda-backend#3904


mshinwell:
(we'd forgotten there was still at least one use of this)

PR #3920, opened 2025-04-23
Unsigned Comparison Primitives

jvanburen:
Adds unsigned comparison primitives to lambda and pipe them all the way through to the backend.


mshinwell:
The z3 proof is by @xclerc

mshinwell (approved):

PR #3774, opened 2025-04-01
Statefulness and visibility axes

dkalinichenko-js:
Rebase of https://github.com/ocaml-flambda/flambda-backend/pull/3098 with updated naming: statefulness going `stateless < observing < stateful` and access going `immutable < read < read_write`.  Similarly to the defaulting for `yielding`, `stateless` implies `portable`, `immutable` implies `contended`, and `read` implies `shared`. Also adds a new kind `atomic_immutable_data`.

Subsequent PRs will update Stdlib, Capsule API, etc, to use those.

PR #3922, opened 2025-04-24
"module type of" zap modalities to floor

riaqn:
`module type of M` should zap modalities in `M` to the strongest. This might cause type error where user uses `module type of M` to constrain another implementation that's weaker than `M`, which is just bad code. I did some quick check in our internal code base and found such bad code to be rare.

### DIFF
--- a/typing/typemod.ml
+++ b/typing/typemod.ml
@@ -3562,11 +3562,9 @@ let type_module_type_of env smod =
   let mty = Mtype.scrape_for_type_of ~remove_aliases env tmty.mod_type in
   (* PR#5036: must not contain non-generalized type variables *)
   check_nongen_modtype env smod.pmod_loc mty;
-  (* for [module type of], we zap to identity modality for best legacy
-  compatibility *)
   let mty =
     remove_modality_and_zero_alloc_variables_mty env
-      ~zap_modality:Mode.Modality.Value.zap_to_id mty
+      ~zap_modality:Mode.Modality.Value.zap_to_floor mty
   in
   tmty, mty
 


typing/typemod.ml
```diff
@@ -3562,11 +3562,9 @@ let type_module_type_of env smod =
   let mty = Mtype.scrape_for_type_of ~remove_aliases env tmty.mod_type in
   (* PR#5036: must not contain non-generalized type variables *)
   check_nongen_modtype env smod.pmod_loc mty;
-  (* for [module type of], we zap to identity modality for best legacy
-  compatibility *)
   let mty =
     remove_modality_and_zero_alloc_variables_mty env
-      ~zap_modality:Mode.Modality.Value.zap_to_id mty
+      ~zap_modality:Mode.Modality.Value.zap_to_floor mty
```
goldfirere, line 3567: This line is obviously correct. But please add some tests to observe this behavior, so that changes around this show up in the testsuite. Maybe one pretty basic test that just prints out the module type, and then another that shows the backward-incompatibility.

We also need to document this. Normally, this kind of low-level detail gets left off the documentation, for decent reasons. But because it’s not backward compatible, I think it deserves careful description. I would say that the documentation should probably be phrased around the new `-ox` flag I proposed, and so maybe that step belongs as part of the public release plan. @ccasin or @dkalinichenko-js, do we have any project tracking for that release or its documentation? I would want to add this there, but not sure where to put it.

riaqn:
@dkalinichenko-js can you look at the updated tests (those with added `stateless`) if they are correct?

typing/typemod.ml
riaqn: tests added. I suggest we move the documentation discussion elsewhere since it's out of the scope of the current PR.

testsuite/tests/typing-modes/val_modalities.ml
```diff
@@ -73,88 +73,121 @@ module Module_type_of_comonadic = struct
...
     let _ = portable_use M.x (* The original inferred modality is zapped *)
 end
 [%%expect{|
-Line 10, characters 25-28:
-10 |     let _ = portable_use M.x (* The original inferred modality is zapped *)
-                              ^^^
-Error: This value is "nonportable" but expected to be "portable".
+module Module_type_of_comonadic :
+  sig
+    module M : sig val x : 'a -> 'a @@ stateless end
```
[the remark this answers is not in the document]
goldfirere: (but we do not have to resolve that in this PR!)

testsuite/tests/typing-modes/val_modalities.ml
```diff
@@ -73,88 +73,121 @@ module Module_type_of_comonadic = struct
...
+         val x : 'a -> 'a
+       is not included in
+         val x : 'a -> 'a @@ stateless
+       The second is portable and the first is nonportable.
 |}]
 
 module Module_type_of_monadic = struct
     module M = struct
-        let x @ uncontended = ref "hello"
+        let x : string ref @ uncontended = ref "hello"
```
goldfirere, line 122: I was intrigued by the need to add this and so checked what would happen without. There is a different output in `-principal` mode. Do we know why? Is the deviation of principal mode here tracked by a CR or ticket? I think I'd prefer not adding the type here so that we see the discrepancy (so we can fix it some day).

testsuite/tests/typing-zero-alloc/signatures.ml
```diff
@@ -865,7 +865,8 @@ end
 
 [%%expect{|
 module M_base_for_mto : sig val f : int -> int [@@zero_alloc] end
-module type S_base_mto = sig val f : int -> int [@@zero_alloc] end
+module type S_base_mto =
+  sig val f : int -> int @@ portable [@@zero_alloc] end
```
goldfirere, line 869: Hrm. But now why does _this_ `portable` show up?

testsuite/tests/typing-modes/val_modalities.ml
riaqn: Mode crossing is guarded by a `is_principal` check, which means in principal mode some mode crossing are no longer availab.e

testsuite/tests/typing-modes/val_modalities.ml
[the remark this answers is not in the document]
riaqn: My impression from https://github.com/ocaml-flambda/flambda-backend/pull/3774 is that `stateless` implies `portable`, hence here `portable` is omitted (but morally still there). 
I think we can wait @dkalinichenko-js to elaborate.

testsuite/tests/typing-modes/val_modalities.ml
goldfirere: Is it just different printing? I seem to remember seeing different inferred modalities. I think that, even in principal mode, there should be no difference here, because the type `string ref` is principally determined (by generalization) before any zapping happens.

To be clear, I'm fine with not fixing this now, but I think adding the type signatures has hidden (rather than delayed) the bug.

testsuite/tests/typing-modes/val_modalities.ml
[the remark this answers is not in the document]
goldfirere: Urgh. Implications bite again. Is there any documentation for the implications? Perhaps once I see them cleanly documented, I can implement them in my brain's compiler. But I have been so far unable to do so.

testsuite/tests/typing-modes/val_modalities.ml
riaqn: I don't remember the details, but the mode crossing in question is probably 1. argument to `ref` is mode crossed 2. result of `ref` is mode crossed. Both are before the generalization.

goldfirere: I don’t want this patch caught here. Maybe just remove the type signature, accept the strange `Principal` output, and leave a CR to return later?

testsuite/tests/typing-zero-alloc/signatures.ml
```diff
@@ -1047,8 +1051,8 @@ module type S = module type of M_infer2
 [%%expect{|
 module type S =
   sig
-    val f : 'a -> 'a [@@zero_alloc strict]
-    val g : 'a -> 'a [@@zero_alloc]
+    val f : 'a -> 'a @@ stateless [@@zero_alloc strict]
+    val g : 'a -> 'a @@ stateless [@@zero_alloc]
```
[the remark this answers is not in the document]
dkalinichenko-js: I believe it's due to the stdlib, which is annotated with portability, but not statelessness. So functions that use something from stdlib are inferred to be `portable`, while functions that don't use anything are inferred to be `stateless`.

testsuite/tests/typing-modes/val_modalities.ml
```diff
@@ -73,88 +73,121 @@ module Module_type_of_comonadic = struct
...
+         val x : string ref @@ stateless
        The second is uncontended and the first is contended.
 |}]
 
 module Module_type_nested = struct
     module M = struct
         let x @ portable = fun t -> t
         module N = struct
-            let y @ uncontended = ref "hello"
+            let y : string ref @ uncontended = ref "hello"
```
dkalinichenko-js, line 152: Why do we need the type ascription here now?

testsuite/tests/typing-modes/val_modalities.ml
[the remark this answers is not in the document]
dkalinichenko-js: Yeah, this seems correct.

The implications are:

- `global => unyielding`
- `local => yielding`
- `stateless => portable`
- `observing => nonportable`
- `stateful => nonportable`
- `immutable => contended`
- `read => shared`
- `read_write => uncontended`

testsuite/tests/typing-modes/val_modalities.ml
goldfirere: I still think leaving a CR would be good. The CR makes it more likely we'll actually get back here someday. Include a link to this discussion in the CR, so we can remember this context later. Thanks.

goldfirere (approved):

PR #3923, opened 2025-04-24
Capsule tests should only run on runtime5

gretay-js:
a minor change in testsuite. Do we expect these tests to pass on runtime4 too?


riaqn:
As discussed, this is probably intentional since mutex is still available on runtime4.

PR #3926, opened 2025-04-24
Use `Unit_info.t` more like upstream

lukemaurer:
Cleans up some of the awkwardness around `Compilation_unit.t` and `Unit_info.t`, the later of which was recently introduced upstream. This PR is meant to take on a slice of what #3828 is doing (and as such it's admittedly a little scattered).

Specifically:

* Adds the `kind` field to `Unit_info.t` to track whether the unit is an interface or implementation (aiming toward pulling in ocaml/ocaml#13286)
* Uses `Compilation_unit.t` in place of `modname` in `Unit_info.t` (see [this comment] for rationale and alternatives)
* Stores the current unit as a `Unit_info.t` rather than a `Compilation_unit.t`, with the `Env` and `Shape` APIs changing accordingly (again this helps with ocaml/ocaml#13286).
 
[this comment]: https://github.com/ocaml-flambda/flambda-backend/pull/3828#issuecomment-2824008311

Meant to be reviewed commit by commit.


typing/env.mli
```diff
@@ -518,8 +518,8 @@ val reset_cache: preserve_persistent_env:bool -> unit
 val reset_cache_toplevel: unit -> unit
 
 (* Remember the name of the current compilation unit. *)
-val set_unit_name: Compilation_unit.t option -> unit
-val get_unit_name: unit -> Compilation_unit.t option
+val set_unit_name: Unit_info.t option -> unit
+val get_unit_name: unit -> Unit_info.t option
```
voodoos, line 522: Is there a reason here why you don't stick closer to the upstream signature ? 
Having a specialized `get_unit_name` that returns the `Compilation_unit.t` directly might reduce the diff a bit ?

Upstream signature is:

```ocaml
(* Remember the current compilation unit. *)
val set_current_unit: Unit_info.t -> unit
val get_current_unit : unit -> Unit_info.t option
val get_current_unit_name: unit -> string (* Compilation_unit.t *)
```

parsing/unit_info.mli
```diff
@@ -108,11 +117,11 @@ module Artifact: sig
    val filename: t -> filename
 
    (** [modname a] is the module name of the compilation artifact.*)
-   val modname: t -> modname
+   val modname: t -> Compilation_unit.t
 
    (** [from_filename filename] reconstructs the module name
        [modname_from_source filename] associated to the artifact [filename]. *)
-   val from_filename: filename -> t
+   val from_filename: for_pack_prefix:Compilation_unit.Prefix.t -> filename -> t
```
riaqn, line 130: update comment?

typing/env.ml
```diff
@@ -1007,18 +1007,22 @@ let rec address_head = function
...
 end = struct
+  let current_unit : Unit_info.t option ref =
+    ref None
   let get () =
-    Compilation_unit.get_current ()
-  let set comp_unit =
-    Compilation_unit.set_current comp_unit
+    !current_unit
+  let set cu =
+    current_unit := cu
```
riaqn, line 1021: Probably rename to `unit_name` instead of `cu`?

riaqn:
Someone else should check alignment with upstream.

goldfirere:
@voodoos You’re probably best poised to check about upstream alignment — do you think this achieves that goal?

mshinwell:
I don't understand why there is a constraint with upstream alignment.

goldfirere:
No constraint. But my understanding is that the goal of this PR is to increase upstream alignment, and so we may want to check that we've done so. If we can't do that today, then I think we can skip that step.

voodoos (commented):
I think this does improve alignment with upstream and more importantly it reconciles the parallel evolution of `Compilation_unit.t` and `Unit_info.t`.

PR #3928, opened 2025-04-24
stack checks in github ci

gretay-js:
This PR proposes to enable stack checks in one of the existing actions for runtime5.


mshinwell (approved):
(worth noting that I think at least some of the arm64 builds e.g. for macOS use stack checks too)

PR #3929, opened 2025-04-25
Use new directives on Arm

spiessimon:
This PR replaces the existing code for emitting directives on Arm with the new directives. The changes that were observed in the emitted assembly code are summarized in this PR ().


backend/asm_targets/asm_directives_new.mli
```diff
@@ -424,3 +442,6 @@ val debug_header : get_file_num:(string -> int) -> unit
 
 (** Reinitialize the emitter before compiling a different source file. *)
 val reset : unit -> unit
+
+(** Directly set the internal section ref. Use this function with caution. It only makes sense when you manually switch directly to a section. *)
+val unsafe_set_interal_section_ref : Asm_section.t -> unit
```
mshinwell, line 447: Typo in the function name

backend/arm64/emit.ml
```diff
@@ -758,15 +742,20 @@ let function_name = ref ""
...
 let add_literal p f =
   try List.assoc f !p
   with Not_found ->
-    let lbl = Cmm.new_label () in
+    (* CR sspies: The [Text] section here is incorrect. We should be in the
+       respective section of the literal type (i.e., 16 or 8 bytes). The code
+       below uses the [Text] section, because that is the section that we are in
+       when we emit literals in the function body. Only macOS currently switches
+       to a dedicated section. *)
+    let lbl = L.create Text in
```
mshinwell, line 758: I'm unclear if something has to be fixed here

mshinwell: Ah no I see, never mind

backend/arm64/emit.ml
```diff
@@ -2068,47 +2083,59 @@ let fundecl fundecl =
...
-  | Csingle f -> emit_float32_directive_with_comment (Int32.bits_of_float f)
-  | Cdouble f -> emit_float64_directive_with_comment (Int64.bits_of_float f)
+      D.global sym;
+    D.define_symbol_label ~section:Data sym
+  (* [Cint8] mirrors mshinwell/ocaml#gdb-names-gpr x86 emitter *)
+  | Cint8 n -> D.int8 (Numbers.Int8.of_int_exn n)
+  (* [Cint16] mirrors mshinwell/ocaml#gdb-names-gpr x86 emitter *)
+  | Cint16 n -> D.int16 (Numbers.Int16.of_int_exn n)
+  (* [Cint32] mirrors mshinwell/ocaml#gdb-names-gpr x86 emitter *)
+  | Cint32 n -> D.int32 (Nativeint.to_int32 n)
```
mshinwell, line 2117: It's a bit strange that this case doesn't have a bounds check, whereas the 8/16 cases do.

backend/arm64/emit.ml
```diff
@@ -2068,47 +2083,59 @@ let fundecl fundecl =
...
-    emit_printf "%a:\n" femit_symbol s.sym_name
-  | Cint8 n -> emit_printf "\t.byte\t%d\n" n
-  | Cint16 n -> emit_printf "\t.2byte\t%d\n" n
-  | Cint32 n -> emit_printf "\t.4byte\t%Ld\n" (Int64.of_nativeint n)
-  | Cint n -> emit_printf "\t.8byte\t%Ld\n" (Int64.of_nativeint n)
-  | Csingle f -> emit_float32_directive_with_comment (Int32.bits_of_float f)
-  | Cdouble f -> emit_float64_directive_with_comment (Int64.bits_of_float f)
+      D.global sym;
+    D.define_symbol_label ~section:Data sym
+  (* [Cint8] mirrors mshinwell/ocaml#gdb-names-gpr x86 emitter *)
```
mshinwell, line 2112: I don't think we need these "mirroring" comments.

backend/arm64/emit.ml
```diff
@@ -2068,47 +2083,59 @@ let fundecl fundecl =
...
+     macOS. The asm directives take care of correctly handling this distinction.
+     For the size, they automatically emit the size [. - symbol], meaning "this
+     minus symbol definition". *)
+  D.type_symbol ~ty:Function fun_sym;
+  D.size fun_sym;
   emit_literals ()
 
 (* Emission of data *)
 
 let emit_item (d : Cmm.data_item) =
```
mshinwell, line 2100: This function should be able to be shared with amd64/emit.ml and moved to emitaux.ml; let's leave a CR.

backend/arm64/emit.ml
```diff
@@ -2068,47 +2083,59 @@ let fundecl fundecl =
...
-    emit_printf "\t.8byte\t%a+%a\n" femit_symbol s.sym_name femit_int o
-  | Cstring s ->
-    if String.length s = 0
-    then emit_string "\n"
-    else emit_string_directive "\t.ascii\t" s
-  | Cskip n -> if n > 0 then emit_printf "\t.space\t%a\n" femit_int n
-  | Calign n -> emit_printf "\t.align\t%a\n" femit_int (Misc.log2 n)
+    let sym = S.create s.sym_name in
+    D.symbol_plus_offset ~offset_in_bytes:(Targetint.of_int o) sym
+  | Cstring s -> D.string s
```
mshinwell, line 2132: Looks like we have lost the zero-length check, although there is a CR in Asm_directives_new.  Maybe fix this in this PR.

backend/arm64/emit.ml
```diff
@@ -2068,47 +2083,59 @@ let fundecl fundecl =
...
+  | Cskip n -> if n > 0 then D.space ~bytes:n
```
mshinwell, line 2133: I think the conditional can go inside D.space

backend/arm64/emit.ml
```diff
@@ -2127,72 +2154,93 @@ let begin_assembly _unix =
...
+  D.int64 0L;
+  D.align ~bytes:8;
   (* #7887 *)
-  let lbl = Cmm_helpers.make_symbol "frametable" in
-  emit_printf "\t.globl\t%a\n" femit_symbol lbl;
-  emit_printf "%a:\n" femit_symbol lbl;
+  let frametable = Cmm_helpers.make_symbol "frametable" in
+  let frametable_sym = S.create frametable in
+  D.global frametable_sym;
+  D.define_symbol_label ~section:Data frametable_sym;
```
mshinwell, line 2199: Let's add a CR about sharing this emit_frames code with amd64/emit.ml, as it should be easy.

backend/arm64/emit.ml
```diff
@@ -2127,72 +2154,93 @@ let begin_assembly _unix =
...
   emit_frames
     { efa_code_label =
         (fun lbl ->
-          emit_symbol_type femit_label lbl "function";
-          emit_printf "\t.8byte\t%a\n" femit_label lbl);
+          let lbl = label_to_asm_label ~section:Text lbl in
```
mshinwell, line 2203: I wonder if the Linear language should use Asm_label, which would remove a lot of label_to_asm_label calls (maybe all of them)?  Maybe leave a CR in linear.mli.

backend/arm64/emit.ml
```diff
@@ -2127,72 +2154,93 @@ let begin_assembly _unix =
...
+          D.define_label lbl);
+      efa_string = (fun s -> D.string (s ^ "\000"))
     };
-  emit_symbol_type femit_symbol lbl "object";
-  emit_symbol_size lbl;
+  D.type_symbol ~ty:Object frametable_sym;
+  D.size frametable_sym;
   if not !Flambda_backend_flags.internal_assembler
   then Emitaux.Dwarf_helpers.emit_dwarf ();
   match Config.system with
```
mshinwell, line 2241: This section should move to e.g. D.mark_stack_as_non_executable.

backend/arm64/emit.ml
```diff
@@ -2127,72 +2154,93 @@ let begin_assembly _unix =
...
         (fun lbl ofs ->
-          emit_printf "\t.4byte\t(%a - .) + %ld\n" femit_label lbl ofs);
-      efa_def_label = (fun lbl -> emit_printf "%a:\n" femit_label lbl);
-      efa_string = (fun s -> emit_string_directive "\t.ascii\t" (s ^ "\000"))
+          let lbl = label_to_asm_label ~section:Data lbl in
+          D.between_this_and_label_offset_32bit_expr ~upper:lbl
+            ~offset_upper:(Targetint.of_int32 ofs));
+      efa_def_label =
+        (fun lbl ->
+          (* The frametable lives in the [.data] section on Arm, but in the
```
mshinwell, line 2231: There should be a CR to change this, I don't think there is any need for this to differ

backend/arm64/emit.ml
```diff
@@ -2127,72 +2154,93 @@ let begin_assembly _unix =
...
-      efa_8 = (fun n -> emit_printf "\t.byte\t%d\n" n);
-      efa_16 = (fun n -> emit_printf "\t.2byte\t%d\n" n);
-      efa_32 = (fun n -> emit_printf "\t.4byte\t%ld\n" n);
-      efa_word = (fun n -> emit_printf "\t.8byte\t%d\n" n);
-      efa_align =
-        (fun n -> emit_printf "\t.align\t%a\n" femit_int (Misc.log2 n));
+          let lbl = label_to_asm_label ~section:Data lbl in
+          D.type_label ~ty:Object lbl;
+          D.label lbl);
+      (* [efa_8] is not part of mshinwell/ocaml#gdb-names-gpr *)
```
mshinwell, line 2211: We can remove the gdb-names-gpr comments I think.

mshinwell (changes requested):
This looks great in general, a few minor things to fix.

backend/arm64/emit.ml
spiessimon: I agree. It seems `Nativeint` does not have an exception version. Should we add an explicit check? We could go up to `int64`, check its in range, and then down to `int32`.

backend/arm64/emit.ml
spiessimon: I explicitly added the zero length check to the Arm code in a previous PR, because the `print_gas` function of the assembly directives checks for zero length. That's why removing it was not necessary in the patches.

backend/arm64/emit.ml
spiessimon: Removed.

backend/arm64/emit.ml
spiessimon: Done.

backend/arm64/emit.ml
spiessimon: Done.

backend/arm64/emit.ml
spiessimon: Done. I changed the implementation in `asm_directives_new.ml` to match the code that was here.

backend/arm64/emit.ml
spiessimon: I turned the comment into a CR.

backend/arm64/emit.ml
spiessimon: Removed.

backend/arm64/emit.ml
spiessimon: I added a conversion function that performs the check. Please have a look that I did not accidentally mess up the bounds checks. Should this function go somewhere else?

backend/arm64/emit.ml
```diff
@@ -2097,6 +2097,13 @@ let fundecl fundecl =
 
 (* Emission of data *)
 
+let int64_to_int32_exn n =
+  if Int64.compare n 0x80000000L (* Int32.int_min *) < 0
+     || Int64.compare n 0x7FFFFFFFL (* Int32.int_max *) > 0
+  then Misc.fatal_error "int64_to_int32_exn: out of range";
+  Int64.to_int32 n
```
mshinwell, line 2106: I would move this to the Numbers module, maybe in an Int64 submodule or something.

mshinwell: (also you shouldn't need to hardcode the constants, the min/max values are defined in Int32 I think)

mshinwell: + use Misc.fatal_errorf with the offending `n`

backend/arm64/emit.ml
```diff
@@ -2238,9 +2236,4 @@ let end_assembly () =
   D.size frametable_sym;
   if not !Flambda_backend_flags.internal_assembler
   then Emitaux.Dwarf_helpers.emit_dwarf ();
-  match Config.system with
-  | "linux" ->
-    (* Mark stack as non-executable *)
-    D.switch_to_section_raw ~names:[".note.GNU-stack"] ~flags:(Some "")
-      ~args:["%progbits"]
-  | _ -> ()
+  D.mark_stack_non_executable ()
```
mshinwell, line 2241: This can presumably be done for amd64 too?

backend/arm64/emit.ml
```diff
@@ -505,6 +505,8 @@ let record_frame_label live dbg =
         Misc.fatal_errorf "Unknown location %a" Printreg.reg r
       | { typ = Int | Float | Float32 | Vec128; _ } -> ())
     live;
+  (* CR sspies: Consider chanting [record_frame_descr] to [Asm_label.t] instead
```
mshinwell, line 508: Chanting?

backend/arm64/emit.ml
spiessimon: Yes, amd64 has the same code here. I'll replace it once this PR is merged.

backend/arm64/emit.ml
spiessimon: Typo. Should be fixed.

mshinwell (approved):

backend/arm64/emit.ml
spiessimon: As discussed in the other comment, this now lives in `Numbers` and we use the bounds from `Int32`.

PR #3931, opened 2025-04-25
Use new directives on x86

spiessimon:
This PR replaces the existing code for emitting directives on x86 with the new directives. The changes that were observed in the emitted assembly code are summarized in this PR (#3932).

### DIFF
--- a/backend/amd64/emit.ml
+++ b/backend/amd64/emit.ml
@@ -41,11 +41,12 @@ module Simd_instrs = Amd64_simd_instrs
 
 open! Branch_relaxation
 module ND = Asm_targets.Asm_directives_new
+module S = Asm_targets.Asm_symbol
+module L = Asm_targets.Asm_label
 
 let rec to_x86_constant (c : ND.Directive.Constant.t) : X86_ast.constant =
   match c with
   | Signed_int i -> Const i
-  (* CR sspies: Is this cast safe? Seems to be just the identity. *)
   | Unsigned_int i -> Const (Numbers.Uint64.to_int64 i)
   | This -> ConstThis
   | Named_thing s ->
@@ -76,10 +77,11 @@ let to_x86_directive (dir : ND.Directive.t) : X86_ast.asm_line list =
     else []
   in
   match dir with
-  | Align { bytes } ->
-    [X86_ast.Align (false, bytes)]
-    (* The data field is currently ignored by both assembler backends. The bytes
-       field is only converted to the final value when printing. *)
+  | Align { bytes; data_section } ->
+    [X86_ast.Align (data_section, bytes)]
+    (* The data field is currently ignored by GAS and MASM, but used in the
+       binary emitter. The bytes field is only converted to the final value when
+       printing. *)
   | Bytes { str; comment } -> comment_lines comment @ [X86_ast.Bytes str]
   | Comment s -> comment_lines (Some s)
   | Const { constant; comment } ->
@@ -98,9 +100,15 @@ let to_x86_directive (dir : ND.Directive.t) : X86_ast.asm_line list =
     (* Behavior differs for negative column values. x86 will not output
        anything, but new directives will output 0. *)
     [X86_ast.Loc { file_num; line; col; discriminator }]
-  | New_label (s, _typ) ->
-    [X86_ast.NewLabel (s, NONE)]
-    (* typ is ignored on x86 and in the new directives*)
+  (* CR sspies: The [typ] matters only for MASM. The convention (implemented in
+     asm directives) is that in the text section, we use Code (NONE) and in the
+     data section, we use Machine_width_data (QWORD for amd64). The two will be
+     emitted differently by MASM. Because some code such as the frame tables
+     have moved from the data section to the text section (but were previously
+     still emitted with QUAD), using the new directives below changes this
+     behavior. *)
+  | New_label (s, Code) -> [X86_ast.NewLabel (s, NONE)]
+  | New_label (s, Machine_width_data) -> [X86_ast.NewLabel (s, QWORD)]
   | New_line -> [X86_ast.NewLine]
   | Private_extern s -> [X86_ast.Private_extern s]
   | Section { names; flags; args } ->
@@ -126,8 +134,22 @@ let to_x86_directive (dir : ND.Directive.t) : X86_ast.asm_line list =
   | Cfi_restore_state -> [X86_ast.Cfi_restore_state]
   | Cfi_def_cfa_register r -> [X86_ast.Cfi_def_cfa_register r]
   | Protected s -> [X86_ast.Protected s]
+  | Hidden s -> [X86_ast.Hidden s]
+  | Weak s -> [X86_ast.Weak s]
+  | External s -> [X86_ast.External (s, NEAR)]
+  (* All uses of [.extrn] use NEAR as the type. *)
+  | Reloc { offset; name = R_X86_64_PLT32; expr } ->
+    [ X86_ast.Reloc
+        { offset = to_x86_constant offset;
+          name = R_X86_64_PLT32;
+          expr = to_x86_constant expr
+        } ]
 
-let _label s = D.label ~typ:QWORD s
+(** Turn a Linear label into an assembly label. The section is checked against the
+    section tracked by [D] when emitting label definitions. *)
+let label_to_asm_label (l : label) ~(section : Asm_targets.Asm_section.t) : L.t
+    =
+  L.create_int section (Label.to_int l)
 
 (* Override proc.ml *)
 
@@ -149,26 +171,11 @@ let phys_rcx = phys_reg Int 5
 
 let phys_xmm0v () = phys_reg Vec128 100
 
-(* CFI directives *)
-
-let cfi_startproc () = if Config.asm_cfi_supported then D.cfi_startproc ()
-
-let cfi_endproc () = if Config.asm_cfi_supported then D.cfi_endproc ()
-
-let cfi_adjust_cfa_offset n =
-  if Config.asm_cfi_supported then D.cfi_adjust_cfa_offset n
-
-let cfi_remember_state () =
-  if Config.asm_cfi_supported then D.cfi_remember_state ()
-
-let cfi_restore_state () =
-  if Config.asm_cfi_supported then D.cfi_restore_state ()
-
-let cfi_def_cfa_register reg =
-  if Config.asm_cfi_supported then D.cfi_def_cfa_register reg
+let file_emitter ~file_num ~file_name =
+  ND.file ~file_num:(Some file_num) ~file_name
 
 let emit_debug_info ?discriminator dbg =
-  emit_debug_info_gen ?discriminator dbg D.file D.loc
+  emit_debug_info_gen ?discriminator dbg file_emitter ND.loc
 
 let emit_debug_info_linear i =
   match i.fdo with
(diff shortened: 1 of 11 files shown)


backend/amd64/emit.ml
```diff
@@ -2109,45 +2145,56 @@ let fundecl fundecl =
...
-      _label (emit_symbol s.sym_name);
-      _label (label_name (emit_symbol s.sym_name)))
-  | Cint8 n -> D.byte (const n)
-  | Cint16 n -> D.word (const n)
-  | Cint32 n -> D.long (const_nat n)
-  | Cint n -> D.qword (const_nat n)
-  | Csingle f -> D.long (Const (Int64.of_int32 (Int32.bits_of_float f)))
-  | Cdouble f -> D.qword (Const (Int64.bits_of_float f))
+      ND.define_symbol_label ~section:Data sym;
+      ND.define_label (L.create_string_unchecked Data (S.encode sym)))
```
mshinwell, line 2167: Seems odd that this is having to use define_label, but if it's giving the correct result leave a CR and we can fix it in a subsequent PR

backend/amd64/emit.ml
```diff
@@ -2179,52 +2226,53 @@ let begin_assembly unix =
...
-    D.extrn "caml_call_gc" NEAR;
-    D.extrn "caml_c_call" NEAR;
-    D.extrn "caml_allocN" NEAR;
-    D.extrn "caml_alloc1" NEAR;
-    D.extrn "caml_alloc2" NEAR;
-    D.extrn "caml_alloc3" NEAR;
-    D.extrn "caml_ml_array_bound_error" NEAR;
-    D.extrn "caml_raise_exn" NEAR);
+    (* These symbols are emitted without additional encoding.*)
+    ND.extrn (S.create ~already_encoded:true "caml_call_gc");
```
mshinwell, line 2233: I think we should just define values of type S.t for these (in module S itself)

mshinwell: (Can be a subsequent PR)

backend/amd64/emit.ml
```diff
@@ -2109,45 +2145,56 @@ let fundecl fundecl =
...
+    match emit_cmm_symbol s with
+    | `Symbol s -> ND.symbol s
+    | `Label l -> ND.label l)
+  | Csymbol_offset (s, o) -> (
     add_used_symbol s.sym_name;
-    D.qword (ConstLabelOffset (emit_cmm_symbol s, o))
-  | Cstring s -> D.bytes s
-  | Cskip n -> if n > 0 then D.space n
-  | Calign n -> D.align ~data:true n
+    match emit_cmm_symbol s with
```
mshinwell, line 2186: Seems like this symbol/label match could be factored out, it occurs elsewhere too

backend/amd64/emit.ml
```diff
@@ -2556,47 +2600,35 @@ let emit_trap_notes () =
...
   let emit_labels list =
-    List.iter (fun l -> D.qword (ConstLabel (emit_label l))) list;
-    D.qword (const 0)
+    List.iter (fun l -> ND.label l) list;
+    ND.int64 0L
   in
   let emit_desc () =
-    D.qword (ConstLabel "_.stapsdt.base");
-    emit_labels (Label.Set.elements traps.enter_traps);
+    ND.symbol (S.create ~already_encoded:true "_.stapsdt.base");
```
mshinwell, line 2607: These could also be predefined in module S.  I wonder if that means we could remove "already_encoded" from the interface, even if it still exists internally within S.

backend/amd64/emit.ml
```diff
@@ -2606,64 +2638,59 @@ let end_assembly () =
...
+  ND.int64 0L;
   (* PR#6329 *)
-  emit_global_label "data_end";
-  D.qword (const 0);
-  D.text ();
-  D.align ~data:true 8;
+  emit_global_label ~section:Data "data_end";
+  ND.int64 0L;
+  ND.text ();
+  ND.align ~data_section:true ~bytes:8;
```
mshinwell, line 2649: This data_section:true deserves a comment

backend/amd64/emit.ml
```diff
@@ -2606,64 +2638,59 @@ let end_assembly () =
...
-            incr setcnt;
-            let s = Printf.sprintf "L$set$%d" !setcnt in
-            D.setvar (s, c);
-            D.long (ConstLabel s))
-          else D.long c);
-      efa_def_label = (fun l -> _label (emit_label l));
-      efa_string = (fun s -> D.bytes (s ^ "\000"))
+          let lbl = label_to_asm_label ~section:Text lbl in
+          let ofs = Targetint.of_int32 ofs in
+          (* CR sspies: On x86 macOS, this changes the name of the variables,
```
mshinwell, line 2671: This CR can be removed probably

backend/asm_targets/asm_directives_new.ml
```diff
@@ -884,9 +932,11 @@ let between_labels_16_bit ?comment:_ ~upper:_ ~lower:_ () =
   (* CR poechsel: use the arguments *)
   Misc.fatal_error "between_labels_16_bit not implemented yet"
 
-let between_labels_32_bit ?comment:_ ~upper:_ ~lower:_ () =
-  (* CR poechsel: use the arguments *)
-  Misc.fatal_error "between_labels_32_bit not implemented yet"
+let between_labels_32_bit ?comment:_comment ~upper ~lower () =
+  let expr = const_sub (const_label upper) (const_label lower) in
+  (* CR sspies: Following the x86 implementation, we *do not* force an assembly
+     time constant here. *)
```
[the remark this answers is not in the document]
spiessimon: I don't know either. Here's one example, where we explicitly emit the expression: https://github.com/ocaml-flambda/flambda-backend/blob/54d64677f66920099264c718c8e0207048818ff3/backend/amd64/emit.ml#L618

I would not change the implementation in this PR, but it could be addressed in a small (easily revertable) subsequent PR. I'll change the comment to indicate this.

backend/amd64/emit.ml
spiessimon: My understanding is that we currently always define two labels. For example, the code `let g (x: int) = x` currently leads to the following, I think,
```asm
camlFile__g_0_1_code:
.LcamlFile__g_0_1_code:
    ...
```

backend/amd64/emit.ml
spiessimon: Not sure what you have in mind here. We currently do 3 of these matches: once to produce a string for the label/symbol, here for a directive storing the symbol/label, and below for a version with an additional offset.

backend/amd64/emit.ml
spiessimon: I agree. However, I don't know what the reason is. Note that in the original code it is also set to true.

backend/asm_targets/asm_section.ml
```diff
@@ -185,6 +195,18 @@ let details t ~first_occurrence =
     | Read_only_data, _, (MinGW_32 | Win32) -> data ()
     | Read_only_data, _, (MinGW_64 | Cygwin) -> [".rdata"], Some "dr", []
     | Read_only_data, _, _ -> rodata ()
+    (* CR sspies: Is this one really possible on all systems? *)
```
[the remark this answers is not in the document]
mshinwell: I would check with @gretay-js

backend/asm_targets/asm_directives_new.ml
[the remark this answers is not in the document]
mshinwell: Yeah that seems fine.

backend/amd64/emit.ml
mshinwell: ok - this would be useful as a comment, but leave a CR as we can probably improve the emitter code to make this clearer

backend/amd64/emit.ml
mshinwell: Ah I thought some of those were the same, but maybe not, so never mind

backend/amd64/emit.ml
spiessimon: I've removed it.

backend/asm_targets/asm_directives_new.ml
```diff
@@ -436,6 +464,11 @@ module Directive = struct
     | Uleb128 _ -> unsupported "Uleb128"
     | Direct_assignment _ -> unsupported "Direct_assignment"
     | Protected _ -> unsupported "Protected"
+    | Hidden _ -> unsupported "Hidden"
+    | Weak _ -> unsupported "Weak"
+    | External s -> bprintf buf "\tEXTRN\t%s: NEAR" s
+    (* The only supported "type" on EXTRN declarations i NEAR. *)
```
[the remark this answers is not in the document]
spiessimon: Fixed.

backend/amd64/emit.ml
```diff
@@ -288,14 +291,19 @@ let emit_jump s = I.jmp (rel_plt s)
...
 let emit_cmm_symbol (s : Cmm.symbol) =
+  let sym = S.create s.sym_name in
   match (s.sym_global : Cmm.is_global) with
-  | Global -> emit_symbol s.sym_name
-  | Local -> label_name (emit_symbol s.sym_name)
+  | Global -> `Symbol sym
+  (* This label is special in that it is not of the form "Lnumber". Instead, we
+     take the symbol, encode it, and turn the resulting string into a label. The
+     label will still be prefixed by ".L"/"L" when emitting. *)
+  | Local -> `Label (L.create_string_unchecked Text (S.encode sym))
```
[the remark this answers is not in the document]
spiessimon: I left a CR.

backend/amd64/emit.ml
```diff
@@ -845,17 +866,17 @@ let global_maybe_protected sym =
       (* Global symbols can be marked as being protected. Unlike in C we don't
          want them to be preempted as we're doing a lot of cross module
          inlining. *)
-      D.protected sym
+      ND.protected sym
 
-let emit_global_label_for_symbol lbl =
+let emit_global_label_for_symbol ~section lbl =
```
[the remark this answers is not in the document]
spiessimon: I left a CR.

backend/amd64/emit.ml
```diff
@@ -1793,15 +1816,15 @@ let emit_instr ~first ~fallthrough i =
       (* We need (63 - result_of_bsr), which can be done with xor. *)
       I.xor (int 63) (res i 0))
     else
-      let lbl_z = Cmm.new_label () in
-      let lbl_nz = Cmm.new_label () in
+      let lbl_z = L.create Text in
+      let lbl_nz = L.create Text in
       I.bsr (arg i 0) (res i 0);
-      I.je (label lbl_z);
+      I.je (sym (L.encode lbl_z));
```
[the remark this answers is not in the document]
spiessimon: Done.

backend/amd64/emit.ml
spiessimon: I left a CR. I think you're right, we can then make `already_encoded` internal.

backend/amd64/emit.ml
spiessimon: I added a CR.

backend/asm_targets/asm_section.ml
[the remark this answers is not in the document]
spiessimon: For platforms other than Linux, the compiler currently does not support Stapsdt and in Arm they are currently not emitted. I've added a check for linux and raised a fatal error for other target systems.

backend/amd64/emit.ml
spiessimon: I talked to @stedolan. I've renamed the parameter into `fill_x86_bin_emitter`, because it controls what the x86 binary emitter fills the space with for align directives. I've also documented this place in the code. In short, while we are in the text section, this code will never be executed so (I believe) it does not matter what we choose.

mshinwell (approved):

PR #3934, opened 2025-04-25
Document [row_more]

goldfirere:
I believe I now understand `row_more`, and this commit invites others into my recently-entered [elysium](https://www.merriam-webster.com/dictionary/Elysium).

Review: @lpw25

### DIFF
--- a/typing/types.mli
+++ b/typing/types.mli
@@ -523,6 +523,24 @@ val create_row:
   name:(Path.t * type_expr list) option -> row_desc
 
 val row_fields: row_desc -> (label * row_field) list
+
+(** [row_more] returns a [type_expr] with one of the following [type_desc]s:
+
+    * [Tvar]: This is a standard row variable; it would occur in e.g. [val f :
+    [> `A | `B] -> int]. When/if we learn more about a polymorphic variant, this
+    variable might get unified with one of the other [type_desc]s listed here,
+    or a [Tvariant] that represents a new set of constructors to add to the row.
+
+    * [Tunivar]: This is a universally quantified row variable; it would occur
+    in e.g. [type t = { f : 'a. ([> `A | `B ] as 'a) -> int }].
+
+    * [Tconstr]: This is always an abstract [#row] type created by a [private]
+    row type, as in [type t = private [> `A | `B]]. (After a substitution, this
+    type might have a manifest; expanding the type will yield either a
+    [Tvariant] or another [type_desc] listed here.
+
+    * [Tnil]: Used to denote a closed polymorphic variant.
+*)
 val row_more: row_desc -> type_expr
 val row_closed: row_desc -> bool
 val row_fixed: row_desc -> fixed_explanation option


goldfirere:
Discussed with Jacques and submitted https://github.com/ocaml/ocaml/pull/14023 upstream. Will update this PR with the result of that review in due course; changing this to a draft in the meantime.

typing/types.mli
```diff
@@ -523,6 +523,24 @@ val create_row:
   name:(Path.t * type_expr list) option -> row_desc
 
 val row_fields: row_desc -> (label * row_field) list
+
+(** [row_more] returns a [type_expr] with one of the following [type_desc]s:
+
+    * [Tvar]: This is a standard row variable; it would occur in e.g. [val f :
+    [> `A | `B] -> int]. When/if we learn more about a polymorphic variant, this
+    variable might get unified with one of the other [type_desc]s listed here,
+    or a [Tvariant] that represents a new set of constructors to add to the row.
```
lpw25, line 532: I don't think you can describe this as a standard row variable. See [Jacques' paper](https://caml.inria.fr/pub/papers/garrigue-polymorphic_variants-ml98.pdf) for a description of roughly how the typing of polymorphic variants works. I could definitely be wrong, but I think the `row_more` "contains" *both* a row variable (written as a rho in that paper) and a presence variable (written as an i in that paper and bound above and below by sets of labels, which are only indicated in the other parts of the `Tvariant`).

typing/types.mli
```diff
@@ -523,6 +523,24 @@ val create_row:
...
+
+    * [Tunivar]: This is a universally quantified row variable; it would occur
+    in e.g. [type t = { f : 'a. ([> `A | `B ] as 'a) -> int }].
+
+    * [Tconstr]: This is always an abstract [#row] type created by a [private]
+    row type, as in [type t = private [> `A | `B]]. (After a substitution, this
+    type might have a manifest; expanding the type will yield either a
+    [Tvariant] or another [type_desc] listed here.
```
lpw25, line 540: This could also be an abstract row coming from an existential

lpw25 (commented):
I think this only describes some of what is going on with `row_more`. Describing what it actually means seems hard, so I'd suggest having the comment only describe the 3 possible cases and which types would produce them, without commenting on what `row_more` actually is beyond that.

typing/types.mli
goldfirere: That doesn't quite make sense to me. If the `row_more` is a `Tvar`, how could it contain all that information? I'm happy to drop the word "standard" from the comment.

typing/types.mli
goldfirere: Yes, Jacques pointed this out, too. See https://github.com/ocaml/ocaml/pull/14023

typing/types.mli
lpw25: There is an invariant that each `Tvar` can only appear within a single `Tvariant`, so they are in a one-to-one relationship. The data about what the `Tvar` is allowed to be unified with is contained in the `Tvariant`. (Personally, I think this is a bad implementation choice and makes the whole thing hard to think about, but I believe it is the right mental model for thinking about how they behave).

goldfirere: Ah, yes, I independently documented that invariant last night. And your comment makes sense in light of that invariant. In the upstream patch, I've dropped the word "standard" (nothing here is standard). But I don't feel able to add text about what you're saying, because I don't yet understand it well enough. If you'd like to suggest specific text to add, I'm happy to. But maybe it's best at this point just to comment on the [upstream patch](https://github.com/ocaml/ocaml/pull/14023)?

My original motivation here is this: In the OCaml type-checker in my head, the type of `row_more` is not a `type_expr`, but Something Else. I had a draft version of the Something Else worked out, and I wanted both to share what I had found and to seek feedback. At the moment, it's just the type definition of the Something Else that I need, hence the shape of this comment. (I'm open to the idea that the best real implementation re-uses `type_expr` for `row_more`. But any formalism of the language would not do so.)

goldfirere:
The version here was accepted upstream. Shall we merge here too?

typing/types.mli
lpw25: I'm fine with just dropping the word standard.

FWIW I actually think that `type_expr` is the right thing to use. I think many formalisms would use a type variable here -- rows can very reasonably be considered types of a particular kind and that is my preferred approach. The problems in this particular implementation come from the kind information then being in a really weird place: in the `Tvariant` that contains the `Tvar`. (There are other problems too: polymorphic variants should really be considered to have two separate components: a set of present tags and a row mapping tags to types, but that separation is completely blurred in the current representation).

lpw25 (approved):

typing/types.mli
goldfirere: Sure -- a kind system to keep rows apart from proper types also works.

PR #3937, opened 2025-04-29
Support for phantom bindings

lthls:
This PR restores support for generating phantom bindings.

I wrote it by enabling phantom lets, and fixing errors as I find them. Given the number of bugs I had to fix, it is likely that some remain in some less-tested corner cases.

Here is a summary of the changes:
- The Name_mode lattice is now a straight line. The typing env requires all the variables in it to have a name mode at least `In_types`, so I made `Phantom` greater than `In_types` instead of not comparable. It would also be possible to prevent the typing env from handling variables in `Phantom` mode, and that should make the environments smaller, but I think it should be a separate PR.
- The logic in `Expr_builder` to decide whether to keep a binding, delete it or phantomise it has been moved to `Simplify_let_expr.rebuild_let`. This makes decisions more consistent (now even bindings removed by mutable unboxing get phantomised if needed)
- Slot offset computations now ignore slots not in normal mode, not only in projections. Phantom closures also get `Deleted` code IDs to avoid keeping old code around.
- No CSE equations are created for phantom bindings

It is possible that this PR still contains changes that were made redundant by a more general fix. Do not hesitate to point these out during review.

### DIFF
--- a/middle_end/flambda2/simplify/expr_builder.ml
+++ b/middle_end/flambda2/simplify/expr_builder.ml
@@ -25,21 +25,13 @@ module UE = Upwards_env
 module DA = Downwards_acc
 module VB = Bound_var
 
-type let_creation_result =
-  | Defining_expr_deleted_at_compile_time
-  | Defining_expr_deleted_at_runtime
-  | Nothing_deleted_at_runtime
-
-let equal_let_creation_results r1 r2 =
-  match r1, r2 with
-  | Defining_expr_deleted_at_compile_time, Defining_expr_deleted_at_compile_time
-  | Defining_expr_deleted_at_runtime, Defining_expr_deleted_at_runtime
-  | Nothing_deleted_at_runtime, Nothing_deleted_at_runtime ->
-    true
-  | ( ( Defining_expr_deleted_at_compile_time | Defining_expr_deleted_at_runtime
-      | Nothing_deleted_at_runtime ),
-      _ ) ->
-    false
+type binding_to_place =
+  | Keep_binding of
+      { let_bound : Bound_pattern.t;
+        simplified_defining_expr : Simplified_named.t;
+        original_defining_expr : Named.t option
+      }
+  | Delete_binding of { original_defining_expr : Named.t option }
 
 let add_set_of_closures_offsets ~is_phantom named uacc =
   let add_offsets_from_set uacc set_of_closures =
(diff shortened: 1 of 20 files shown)


middle_end/flambda2/simplify/simplify_unary_primitive.ml
```diff
@@ -153,7 +153,8 @@ let simplify_unbox_number (boxable_number_kind : K.Boxable_number.t) dacc
                   ( Box_number
                       (boxable_number_kind, Alloc_mode.For_allocations.heap),
                     Simple.var result_var' )))
-            ~bound_to:arg)
+            ~bound_to:arg
+            ~name_mode:(Bound_var.name_mode result_var))
```
chambart, line 157: There are a few cases like that where the `name_mode` argument does not directly come from the `bound_to`. I am not really able to tell in those cases whether the thing should have the same name mode, or if there is a contrieved way for them not to.

middle_end/flambda2/simplify/simplify_let_expr.ml
```diff
@@ -76,26 +76,156 @@ let rebuild_let simplify_named_result removed_operations ~rewrite_id
...
+    | Singleton bound_var ->
+      (* We avoid the closure allocation (below) in this case. *)
+      Name_occurrences.greatest_name_mode_var free_names_of_body
+        (VB.var bound_var)
+    | Set_of_closures _ ->
+      Bound_pattern.fold_all_bound_vars bound_vars
+        ~init:Name_mode.Or_absent.absent ~f:(fun greatest_name_mode bound_var ->
+          Name_occurrences.greatest_name_mode_var free_names_of_body
+            (VB.var bound_var)
+          |> Name_mode.Or_absent.join_in_terms greatest_name_mode)
```
chambart: I don't find that this `|>` operator increase lisibility here (but it doesn't matter much)

middle_end/flambda2/simplify/simplify_let_expr.ml
```diff
@@ -76,26 +76,156 @@ let rebuild_let simplify_named_result removed_operations ~rewrite_id
...
+              (Name_mode.Or_absent.present declared_name_mode)
+            > 0
+          in
+          let { Simplified_named.named = defining_expr;
+                free_names = _;
+                cost_metrics = _
+              } =
+            simplified_defining_expr
+          in
+          let defining_expr = Simplified_named.to_named defining_expr in
```
chambart:
```suggestion
          let defining_expr = Simplified_named.to_named
            simplified_defining_expr.Simplified_named.named
          in
```

middle_end/flambda2/simplify/simplify_let_expr.ml
```diff
@@ -76,26 +76,156 @@ let rebuild_let simplify_named_result removed_operations ~rewrite_id
...
+            | None -> false, false
+            | Some region ->
+              let is_used =
+                Name.Set.mem (Name.var region) (UA.required_names uacc)
+              in
+              not is_used, is_used
+          in
+          if is_end_region_for_used_region
+             || (not is_end_region_for_unused_region)
+                && not (Named.at_most_generative_effects defining_expr)
```
chambart:
```suggestion
          let must_be_kept_for_its_effects =
             is_end_region_for_used_region
             || (not is_end_region_for_unused_region)
                && not (Named.at_most_generative_effects defining_expr)
          in
          if must_be_kept_for_its_effets
```

middle_end/flambda2/simplify/simplify_set_of_closures.ml
```diff
@@ -797,8 +797,49 @@ let simplify_non_lifted_set_of_closures0 dacc bound_vars ~closure_bound_vars
...
   in
   let { set_of_closures; dacc } =
-    simplify_set_of_closures0 dacc context set_of_closures ~closure_bound_names
-      ~closure_bound_names_inside ~value_slots ~value_slot_types
+    if Name_mode.is_normal (Bound_pattern.name_mode bound_vars)
+    then
+      simplify_set_of_closures0 dacc context set_of_closures
+        ~closure_bound_names ~closure_bound_names_inside ~value_slots
+        ~value_slot_types
+    else
```
chambart, line 817:
```suggestion
    else
      (* If the closure is to be phantomized, mark all code as deleted *)
```

middle_end/flambda2/simplify/expr_builder.ml
```diff
@@ -285,72 +157,57 @@ let create_coerced_singleton_let uacc var defining_expr
...
-  { let_bound : Bound_pattern.t;
-    simplified_defining_expr : Simplified_named.t;
-    original_defining_expr : Named.t option
-  }
+      generate_outer_binding (Bound_var.name_mode var)
 
 let make_new_let_bindings uacc ~bindings_outermost_first ~body =
   (* The name occurrences component of [uacc] is expected to be in the state
      described in the comment at the top of [Simplify_let.rebuild_let]. *)
+  let delete_binding uacc ~original_defining_expr =
```
chambart: This is used only once. I'm not sure it is better as a function rather than inlined.
If you want to keep the function, I'd prefer a better name since this does not delete the binding.
Like `notify_removed`

middle_end/flambda2/simplify/simplify_named.ml
```diff
@@ -81,65 +81,95 @@ let simplify_named0 dacc (bound_pattern : Bound_pattern.t) (named : Named.t)
...
-                    simplified_defining_expr = defining_expr;
-                    original_defining_expr = Some named
-                  } ])))
+    if not (Name_mode.is_normal (Bound_pattern.name_mode bound_pattern))
+    then
+      let min_name_mode = Bound_pattern.name_mode bound_pattern in
+      let defining_expr =
+        Named.create_prim
+          (P.map_args
+             (fun simple -> snd (S.simplify_simple dacc simple ~min_name_mode))
```
chambart, line 103: When we phantomize an expression, should we `simplify_simple` its arguments ? This is going to replace phantom variables with constants, but is this what we want.
It's really not obvious either way, since when debugging, it might be better to know the value or to know the relation depending on the situation.
If we were not to do it, we certainly need to do something like some parts of simplify simple for the, but maybe don't do the simplifications to constants.

middle_end/flambda2/simplify/simplify_unary_primitive.ml
lthls: `add_cse` will ignore all equations when the mode is not `Normal`. Here we use the mode of the result var to know whether we have a real equation (both `bound_to` and the primitive arguments have normal mode) or not.

middle_end/flambda2/simplify/simplify_let_expr.ml
lthls: I'm always happy to remove these operators (I don't know why I thought it was a good idea).

middle_end/flambda2/simplify/expr_builder.ml
lthls: I went with `notify_removed`.

middle_end/flambda2/simplify/simplify_named.ml
lthls: I don't remember debuggers being able to actually show the relations. I think knowing the value is always going to be better, at least when the source is available.

lthls: @mshinwell do you know if it would be useful for a debugger to know that, for example, `z` is `x + y` instead of (or in addition to) `x + 1`, when `y` is known to be `1` ?

mshinwell: Maybe, but for the moment it's probably ok to do the simplification here.  Please leave a CR.

PR #3938, opened 2025-04-29
Add comments clarifying domain/finaliser guarantees

NickBarnes:
It has always been true, in multicore OCaml, that finalisers are run by the domain which registered them, unless that domain has terminated in which case they are adopted by some other domain. This is important for mode safety (as values captured by a finaliser closure may otherwise escape to some other domain), so we're should document it in the `.mli`.


tdelvecchio-jsc:
Do we have a guarantee (or can we make it a guarantee) that the domain that inherits the finaliser is the domain that joins the terminated domain? This would enable finalising nonportable values with nonportable finalisers on non-initial domains.

NickBarnes:
We do not have such a guarantee, and I think it would be non-trivial (although clearly not impossible) to make it. At present, the "orphaned" finalizer data structures from the terminated domain are adopted by the first domain which notices that they exist (while GCing), which might happen long before any domain joins the terminated domain.
To add this we would have to add a mechanism to attach those finalizer structures to the `term_sync` value associated with the running (or terminated) domain. Until the domain is joined, somebody still has to do the GC work of scanning those structures looking for dead values and transferring finalizer closures from one table to another etc, while nobody is permitted to actually run the finalizers. So I guess you'd have a "fostering" mode, in addition to "adoption". Clearly doable, but note that the "orphaned finalizers" (and "orphaned ephemerons") are a rich source of nasty GC bugs, and this would add to an already twisty heap of code.

NickBarnes:
Furthermore, multiple domains can call `Domain.join` on the same domain. It would be possible to provide a guarantee like: "Finalizers which have not yet run may be inherited from a terminating domain by some other domain which has called `Domain.join` on it. This inheritance may be successive as inheriting domains themselves terminate and are joined. Finalisers will only ever be run by the registering domain or by some other domain which has inherited them in this way." But as noted above, that would require work in the runtime and in the `Domain` module implementation.

PR #3940, opened 2025-04-29
Fix miscompilation / liveness errors for string operations

mshinwell:
This fixes the random CI failures seen on arm64.

It turns out that the Cmm machtypes given to "pointer + offset" calculations used for all of the different varieties of string reads and writes are wrong.  Some of them have been wrong for a very long time (30 years).  They are currently given type `Int`, but in the case where the pointers may be into the OCaml heap (e.g. for `string` and `bytes` rather than `Bigstring.t`) they must have type `Addr`.  Otherwise, CSE may apply, causing pointers into the middle of values to be live across a GC.  For example in `warnings.ml`:
```ocaml
  and loop_letter_num tokens modifier i =
    if i >= String.length s then error () else
    match s.[i] with  (* first string access *)
    | '0' .. '9' ->
        let i, n1, n2 = get_range i in
        loop (Num(n1,n2,modifier)::tokens) i
    | 'A' .. 'Z' | 'a' .. 'z' ->
       loop (Letter(s.[i],Some modifier)::tokens) (i+1)  (* repeated string access after potential GC *)
    | _ -> error ()
```

This patch affects the system compiler as well as the flambda-backend compiler, hence the use of a patch file to fix the former.  I will deal with getting this fixed upstream.  The system compiler patch isn't currently as good as the flambda-backend patch since it doesn't special-case out-of-heap accesses.

There are a couple of other changes currently in here too, which can be discussed and maybe split:
- fixes to stop the system compiler applying `Comballoc` across `Ccheckbound`, which can raise an exception.  I need to discuss this with @stedolan though, maybe this is only required if the allocation is on the major heap?  (This problem does not arise with the flambda-backend compiler since bounds checks are expanded before Flambda 2.)
- fixes to silence the debug runtime with the system compiler, which I think was causing excessive CI logs in some cases.

This bug was very troublesome to track down and @xclerc must be thanked for his help with this, especially since we thought there was a second bug, which was actually caused by a mistake in my hack for tracking down the first bug!  Some of the CI changes in this PR were written by @xclerc .

Once the dust has settled I will re-enable the arm64 builds as required.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -200,12 +200,14 @@ jobs:
       if: matrix.os == 'macos-latest'
       run: HOMEBREW_NO_INSTALL_CLEANUP=TRUE brew install autoconf
 
+      # NB. The "rev" number in the cache key below must be updated each time
+      # the patch file is changed!
     - name: Cache OCaml 4.14, dune and menhir
       uses: actions/cache@v4
       id: cache
       with:
         path: ${{ github.workspace }}/ocaml-414/_install
-        key: ${{ matrix.os }}-cache-ocaml-414-patched-dune-3152-menhir-20231231
+        key: ${{ matrix.os }}-cache-ocaml-414-patched-dune-3152-menhir-20231231-rev13
 
     - name: Checkout OCaml 4.14
       uses: actions/checkout@master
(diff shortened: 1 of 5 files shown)


mshinwell:
Something is wrong with the CI here, re-opening.

mshinwell:
See https://github.com/ocaml-flambda/flambda-backend/pull/3960

PR #3941, opened 2025-04-29
Remove upstream tests that also exist in flambda-backend tests

jvanburen:


goldfirere:
Having two copies of the tests is definitely bad. But is there a good reason to favor the flambda-backend testsuite over the upstream one? I expect we'll have an easier time upstreaming flambda2 if we don't have to overhaul the testsuite at the same time. I don't really know what I'm talking about here, and maybe you all have conferred (in which case just tell me to wonder off), but I saw this go by and was curious.

Gbury:
I'm with @goldfirere on this one: as much as possible, we should use the existing upstream testsuite harness for tests (because upstreaming). Only for specific tests that cannot be made compatible should we keep separate tests (and even then, I suspect it would be possible to adapt these to ocamltest, since it allows to use arbitrary scripts).

jvanburen:
someone else will have to confirm but i believe this *does* use the upstream test harness, the tests are copied into the upstream directory by the makefile before running!

bclement-ocp:
Something slightly fishy seems to be going on with these tests because I think we are not just using new versions of existing tests but also dropping a bunch of them. The files that are there (I just looked at `poll_attr_both` and `poll_attr_user`) also seem to be identical in both directories. Might be an artifact of the pivot root @mshinwell @lukemaurer?

```console
$ ls flambda-backend/testsuite/tests/asmcomp/
poll_attr_both.compilers.reference  poll_attr_inserted.compilers.reference  poll_attr_prologue.compilers.reference  poll_attr_user.compilers.reference
poll_attr_both.ml                   poll_attr_inserted.ml                   poll_attr_prologue.ml                   poll_attr_user.ml

$ ls testsuite/tests/asmcomp/
0001-test.compilers.reference      evaluation_order.reference   is_static.ml                            poll_attr_prologue.ml               regression_value_kinds.ml      static_float_array_flambda_opaque.ml
0001-test.ml                       func_sections.arm.reference  lift_mutable_let_flambda.ml             poll_attr_user.compilers.reference  run.ml                         try_checkbound.ml
bind_tuples.ml                     func_sections.ml             optargs.ml                              poll_attr_user.ml                   select_addr.ml                 unrolling_flambda2.ml
compare.ml                         func_sections.reference      poll_attr_both.compilers.reference      polling.c                           select_addr.reference          unrolling_flambda.ml
compare.reference                  func_sections.run            poll_attr_both.ml                       polling_insertion.ml                simple_float_const.ml
evaluation_order_broken.ml         is_in_static_data.c          poll_attr_inserted.compilers.reference  prevent_fma.ml                      simple_float_const_opaque.ml
evaluation_order_broken.reference  is_static_flambda_dep.ml     poll_attr_inserted.ml                   register_typing.ml                  staticalloc.ml
evaluation_order.ml                is_static_flambda.ml         poll_attr_prologue.compilers.reference  register_typing_switch.ml           static_float_array_flambda.ml
```

Gbury:
> someone else will have to confirm but i believe this _does_ use the upstream test harness, the tests are copied into the upstream directory by the makefile before running!

Right, but in that case, why not have/commit the files at their correct place from the start ? Having separate copies that we have to copy and replace the original with seem like an really unnecessary complication (compared to just adding/changing the files in the `testsuites/tests/**` directory).

lukemaurer:
> Something slightly fishy seems to be going on with these tests because I think we are not just using new versions of existing tests but also dropping a bunch of them. The files that are there (I just looked at `poll_attr_both` and `poll_attr_user`) also seem to be identical in both directories. Might be an artifact of the pivot root @mshinwell @lukemaurer?

I'd be very surprised if this happened because of or after pivot-root—there was a lot of stuff copied from one tree to another in the early days of flambda-backend (that was a major motivation for pivot-root, in fact). As I recall, I basically punted on doing anything with the downstream ocamltest tests besides move them into `flambda-backend/`.

I agree with @goldfirere and @Gbury, though: we should definitely delete the _downstream_ test where there is truly a duplicate. Actually I would go a bit further: Where there is a collision, we should actually be running _both tests_ since the upstream and downstream have probably diverged from when the test was copied. So the test harness should be copying `flambda-backend/testsuite/tests/*` into `_runtest/testsuite/tests/flambda-backend/` rather than straight into `_runtest/testsuite/tests/`. (But that's maybe a separate PR.)

PR #3943, opened 2025-04-29
Clarify and expand modes syntax docs

goldfirere:
I was reading through this on my way to somewhere else and made some edits.

Review: @riaqn, who I have also requested to add an entry that I don't feel qualified to do myself.

### DIFF
--- a/jane/doc/extensions/_04-modes/syntax.md
+++ b/jane/doc/extensions/_04-modes/syntax.md
@@ -142,6 +142,10 @@ though it is conceivable we will have modalities other than modes in the future.
 modalities ::= modes
 ```
 
+<!-- CR reisenberg: This should be moved to a page about the semantics
+of modalities, instead of here in the syntax page. But we don't have
+such a page now, so it's here for the time being. -->
+
 Modalities are used to describe the relationship between a container and an
 element in that container; for example, if you have a record field `x` with
 a `portable` modality, then `r.x` is `portable` even if `r` is `nonportable`.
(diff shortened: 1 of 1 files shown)


jane/doc/extensions/_04-modes/syntax.md
```diff
@@ -24,19 +36,19 @@ Modes are in a dedicated namespace separated from variable names or type names,
...
 names.
 
-Currently a modality expression is simply a space-delimited list of modalities.
+A mode expression actually contains a specification for each modal axis, whether
+you have written a choice for that axis or not. For axes that are omitted, the
+so-called *legacy* modes are used instead. The legacy modes are as follows:
 
+```ocaml
+global aliased many nonportable uncontended unyielding stateful read_write
 ```
```
riaqn, line 45: This is only true for modes on arrow types. But I think maybe better to move this to another document about semantics.

jane/doc/extensions/_04-modes/syntax.md
```diff
@@ -100,40 +115,126 @@ You can also specify the mode of the function body:
...
+
+<!-- CR reisenberg for zqian: There are other signature items that can have
+modalities, according to the parser:
+  include S @@ modalities
+  module (M @@ modalities) : module_type   (* but this does not work with [rec] *)
+  module M : module_type @@ modalities  (* this form *does* work with [rec] *)
+  module (M @@ modalities) = M2
+  module M = M2 @@ modalities
+I don't exactly know what these mean, so I have not documented them.
+-->
```
riaqn, line 230: The first is kinda hacky right now, and I'd rather not mention it until #3759 .

The rest are modules which are not supported yet.

jane/doc/extensions/_04-modes/syntax.md
```diff
@@ -100,40 +115,126 @@ You can also specify the mode of the function body:
...
+Any axis left out of a modality expression is assumed to be the identity
+modality. (When a modality is the identity, then the mode of a field is the same
+as the mode of the record.) For future axes, this would be the top mode; for
+past axes, this would be the bottom mode. These are the identity modalities:
+
+```ocaml
+local unique once nonportable uncontended unyielding stateless immutable
+```
+
+Note that a legacy mode might or might not be the same as the identity modality.
```
riaqn, line 168: While the added text is very informative, it probably belongs to another page dedicated to the *semantics* of modes and modalities. This document was intended as a quick lookup for where and how mode/modalities can show up in syntax.

goldfirere: Yes, quite right, thanks. I've added a CR to move it once we have a place to move it to.

jane/doc/extensions/_04-modes/syntax.md
riaqn: Yeah when you write `(exp : _ @ local)` you are only specifying the locality; the rest are not constrained.

goldfirere: Ah yes. I've revised. Thanks for catching this.

riaqn (approved):

Issue #3914, opened 2025-04-22
Equality via pattern-matching compiles to slow code


jvanburen:
example: 
```
  type t =
    | Naked_immediate
    | Naked_float32
    | Naked_float
    | Naked_int8
    | Naked_int16
    | Naked_int32
    | Naked_int64
    | Naked_nativeint
    | Naked_vec128

  let equal t1 t2 =
    match t1, t2 with
    | Naked_immediate, Naked_immediate -> true
    | Naked_float32, Naked_float32 -> true
    | Naked_float, Naked_float -> true
    | Naked_int8, Naked_int8 -> true
    | Naked_int16, Naked_int16 -> true
    | Naked_int32, Naked_int32 -> true
    | Naked_int64, Naked_int64 -> true
    | Naked_nativeint, Naked_nativeint -> true
    | Naked_vec128, Naked_vec128 -> true
    | ( ( Naked_immediate | Naked_float32 | Naked_float | Naked_int8
        | Naked_int16 | Naked_int32 | Naked_int64 | Naked_nativeint
        | Naked_vec128 ),
        _ ) ->
      false
```
leads to 
```
camlExample.equal_1:
        .quad   caml_curry2
        .quad   0x280000000000007
        .quad   camlExample.equal_0_1_code
        .quad   0xffc
        .ascii  "unreachable switch case"
        .byte   0
camlExample.equal_0_1_code:
        sar     rax
        lea     rdx, [rip + .L185]
        movsxd  rax, dword ptr [rdx + 4*rax]
        add     rdx, rax
        jmp     rdx
.L117:
        cmp     rbx, 1
        je      .L121
        mov     eax, 1
        ret
.L121:
        mov     eax, 3
        ret
.L124:
        cmp     rbx, 3
        je      .L128
        mov     eax, 1
        ret
.L128:
        mov     eax, 3
        ret
.L131:
        cmp     rbx, 5
        je      .L135
        mov     eax, 1
        ret
.L135:
        mov     eax, 3
        ret
.L138:
        cmp     rbx, 7
        je      .L142
        mov     eax, 1
        ret
.L142:
        mov     eax, 3
        ret
.L145:
        cmp     rbx, 9
        je      .L149
        mov     eax, 1
        ret
.L149:
        mov     eax, 3
        ret
.L152:
        cmp     rbx, 11
        je      .L156
        mov     eax, 1
        ret
.L156:
        mov     eax, 3
        ret
.L159:
[...]

PR #3946, opened 2025-04-29
Add PPX support to the compiler

jvanburen:
Adapt ppx_base to work in the compiler codebase. This will produce faster and smaller equality/comparison code than the hand-rolled ones, and it will make the codebase smaller and more readable.

### DIFF
--- a/typing/primitive.ml
+++ b/typing/primitive.ml
@@ -20,13 +20,19 @@ open Parsetree
 
 module String = Misc.Stdlib.String
 
-type unboxed_integer =  Unboxed_int64 | Unboxed_nativeint | Unboxed_int32
+type unboxed_integer = Unboxed_int64 | Unboxed_nativeint | Unboxed_int32
+[@@deriving compare, enumerate, equal]
 type unboxed_float = Unboxed_float64 | Unboxed_float32
+[@@deriving compare, enumerate, equal]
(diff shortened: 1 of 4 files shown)


jvanburen:
this adds ppx_base as a dependency though, which we'll need to mention + update github actions...

mshinwell:
I don't think we should do this, for at least the following reasons:
* there will never be acceptance of ppx upstream (correctly in my opinion) and so using it in flambda-backend is going to increase obstacles to upstreaming
* ppx is likely to make it harder to work on the compiler when it itself needs to be debugged, e.g. for the current arm64 `warnings.ml` problem.  I believe this outweighs any concern about time spent manually maintaining these functions
* it will mean that the semantics of the compiler starts depending on the semantics of ppx
* the argument about the legibility and performance of the boilerplate functions isn't really a strong one in my opinion: in particular, it seems unlikely these are serious performance bottlenecks.  Algorithms with poor time complexity or repeated IR traversals seem more likely suspects
* AI systems can now easily be used to write the boilerplate functions, which significantly reduces effort for new ones.

This seems likely to be related to #3914, but we should just fix the code generation if it is actually a problem.

Gbury:
I agree with @mshinwell : I don't think the benefits of using/addping ppx in the compiler outweighs the (many) drawbacks it would inflict on us.

PR #3947, opened 2025-04-29
Add multidomain config option & CI jobs

TheNumbat:
To get multiple domains, you now need to configure with `--enable-multidomain`.
Multidomain requires also enabling `--enable-stack-checks` and `--enable-poll-insertion`.

- Adds a `multidomain` test predicate and adds it to all tests that spawn domains.
- Adds CI jobs for stack-checks-only, poll-insertion-only, and multidomain.

We have a lot of CI jobs now and some of them are mostly redundant, so we should consider removing some.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,6 +20,18 @@ jobs:
             config: --enable-middle-end=flambda2 --enable-runtime5
             os: ubuntu-latest
 
+          - name: flambda2_stack_checks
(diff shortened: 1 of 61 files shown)


TheNumbat:
Note `test_issue_11094.ml` is now able to create the full 100k fibers, matching the upstream test. This is because it only runs with `--enable-multidomain`, which implies `--enable-stack-checks`, which implies small (non-guard-page) fibers.

PR #3949, opened 2025-04-30
Refactor lattice flipping for modes

riaqn:
Currently in `solver.ml` there is `Solver_mono` which can handle monotone morphisms. Based on that, there are two polarized solvers `Positive` and `Negative`, used for comonadic and monadic axes respectively. The intention is such that user code (`mode.ml`) doesn't need to think about the flipping. This has two issues:

- It turns out that `mode.ml` still needs to think about the flipping in some definitions but not all of them. This mixture, and the fact that both `mode.ml` and `solver.ml` handles flipping, makes the code more confusing.
- `Positive.t` and `Negative.t` are different types and complicates type polymorphism. In particular, `Value.Axis.t` is a complicated GADT.

So we remove both `Positive` and `Negative`, and let `mode.ml` to be the only file that cares about flipping. We move some code such that the flipping are concentrated to fewer places.

The downside is that code that uses `mode` might see `(Uniqueness.Const.t, allowed * disallowed) mode` instead of `(disallowed * allowed) Uniqueness.t`. The two types are the same, but ideally we want to show the latter.

# Review

Please review by commit.

### DIFF
--- a/typing/mode.ml
+++ b/typing/mode.ml
@@ -1753,6 +1753,14 @@ module Comonadic_gen (Obj : Obj) = struct
 
   let of_const : type l r. const -> (l * r) t = fun a -> Solver.of_const obj a
 
+  let meet_const c m = Solver.apply obj (Meet_with c) m
+
+  let join_const c m = Solver.apply obj (Join_with c) m
+
+  let imply c m = Solver.apply obj (Imply c) (Solver.disallow_left m)
+
+  let subtract c m = Solver.apply obj (Subtract c) (Solver.disallow_right m)
+
   module Guts = struct
     let get_floor m = Solver.get_floor obj m
 
@@ -1826,6 +1834,14 @@ module Monadic_gen (Obj : Obj) = struct
   let zap_to_floor m = with_log (Solver.zap_to_ceil obj m)
 
   let of_const : type l r. const -> (l * r) t = fun a -> Solver.of_const obj a
+
+  let meet_const c m = Solver.apply Obj.obj (Join_with c) m
+
+  let join_const c m = Solver.apply Obj.obj (Meet_with c) m
+
+  let imply c m = Solver.apply obj (Subtract c) (Solver.disallow_right m)
+
+  let subtract c m = Solver.apply obj (Imply c) (Solver.disallow_left m)
 
   module Guts = struct
     let _get_floor m = Solver.get_ceil obj m
@@ -2057,6 +2073,20 @@ module type Areality = sig
   val zap_to_legacy : (Const.t, allowed * 'r) Solver.mode -> Const.t
 end
 
+module BiHeyting_Product (L : BiHeyting) = struct
+  include L
+
+  type 'a axis = (t, 'a) Axis.t
+
+  let min_with ax c = Axis.update ax c min
+
+  let max_with ax c = Axis.update ax c max
+
+  let min_axis ax = Axis.proj ax min
+
+  let max_axis ax = Axis.proj ax max
+end
+
 module Comonadic_with (Areality : Areality) = struct
   module Obj = struct
     type const = Areality.Const.t C.comonadic_with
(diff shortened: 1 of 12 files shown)


typing/mode.ml
```diff
@@ -2928,11 +2973,11 @@ let value_to_alloc_r2l m =
   { comonadic; monadic }
 
 module Modality = struct
-  type ('m, 'a) raw =
-    | Meet_with : 'a -> (('a, 'l * 'r) mode_comonadic, 'a) raw
-    | Join_with : 'a -> (('a, 'l * 'r) mode_monadic, 'a) raw
```
riaqn, line 2933: This GADT was intended to encode the fact that `Meet_with` can be only applied on comonadic axes, to remove the assertion at L2988. But it doesn't seem to be work, so I'm simplify it.

typing/mode.ml
```diff
@@ -2295,37 +2335,29 @@ module Value_with (Areality : Areality) = struct
...
-      P (Monadic Contention);
-      P (Comonadic Portability);
-      P (Comonadic Statefulness);
-      P (Monadic Visibility) ]
-
-  let proj_obj : type m a d. (m, a, d) axis -> a C.obj = function
+  module Axis = struct
+    type ('a, 'd0, 'd1) t =
+      | Monadic : (Monadic.Const.t, 'a) Axis.t -> ('a, 'd, 'd neg) t
+      | Comonadic : (Comonadic.Const.t, 'a) Axis.t -> ('a, 'd, 'd pos) t
```
riaqn, line 2341: TBH the new GADT doesn't seem much simpler than before, but at least it no longer mentions `mode_monadic` and `mode_comonadic`, and instead uses allowance to encode flipping.

typing/untypeast.ml
```diff
@@ -274,11 +274,11 @@ let mutable_ (mut : Types.mutability) : mutable_flag =
   match mut with
   | Immutable -> Immutable
   | Mutable m ->
-      if Mode.Alloc.Comonadic.Const.eq m Mode.Alloc.Comonadic.Const.legacy then
+      let open Mode.Alloc.Comonadic.Const in
+      if Misc.Le_result.equal ~le:le m legacy then
```
goldfirere, line 278:
```suggestion
      if Misc.Le_result.equal ~le m legacy then
```

typing/printtyp.ml
```diff
@@ -1837,7 +1837,8 @@ let tree_of_label l =
     match l.ld_mutable with
     | Mutable m ->
         let mut =
-          if Alloc.Comonadic.Const.eq m Alloc.Comonadic.Const.legacy then
+          let open Alloc.Comonadic.Const in
+          if Misc.Le_result.equal ~le m Alloc.Comonadic.Const.legacy then
```
goldfirere, line 1841:
```suggestion
          if Misc.Le_result.equal ~le m legacy then
```

goldfirere (approved):

typing/mode.ml
```diff
@@ -2197,31 +2249,19 @@ module Monadic = struct
...
+    Solver.apply Obj.obj (Min_with ax) (Solver.disallow_right m)
 
   let min_with ax m =
-    Obj.Solver.via_monotone Obj.obj (Max_with ax) (Obj.Solver.disallow_right m)
-
-  let join_with ax c m = join_const (C.max_with Obj.obj ax c) m
+    Solver.apply Obj.obj (Max_with ax) (Solver.disallow_left m)
 
-  let meet_with ax c m = meet_const (C.min_with Obj.obj ax c) m
+  let join_with ax c m = join_const (Const.min_with ax c) m
```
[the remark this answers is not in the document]
riaqn: because both `join_const` and `min_with` are already flippeed. Essentially, only when we call `Solver.*` we need to worry about flipping.

typing/solver_intf.mli
```diff
@@ -304,70 +292,10 @@ module type S = sig
...
-    and Negative :
-      (Solver_polarized
-        with type 'd polarized = 'd neg
-         and type ('a, 'd) mode_op = ('a, 'd) Positive.mode)
-
-    (* The following definitions show how this solver works over a category by
-       defining objects and morphisms. These definitions are not used in
-       practice. They are put into a module to make it easy to spot if we end up
-       using these in the future. *)
-    module Category : sig
```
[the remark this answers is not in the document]
riaqn: Yeah after we removed the `positive` and `negative` solver, it's hard to connect this virtual category to what we are doing. Instead, I will reorganize `mode.ml` and add comments.

PR #3952, opened 2025-04-30
Cleanup after using the new directives

spiessimon:
This PR contains additional cleanup like removing the code for the old asm directives. It should only be considered after #3931 and #3948 are merged. (The changes are all in the last commit.) There is no corresponding patches PR for this one.

One change that is not done yet in this PR to avoid merge conflicts when rebasing is renaming the asm directives from `asm_directives_new` into simply `asm_directives`.


mshinwell (approved):
Excellent, although I'd like to double check the reason for the removal of the following CR:
`(* CR sspies: Potentially enable range check again here. We currently emit
         numbers like 158 as an 8-bit constant, which breaks int8 range checks,
         but make the assembler perfectly happy. *)`
I think commits could be added on this PR to remove the old directives files and do the renames, although fine in a separate PR too.

backend/asm_targets/asm_label.mli
```diff
@@ -55,6 +55,11 @@ val create_string : Asm_section.t -> string -> t
 (** Create a textual label. Argument string is not checked, so use with caution. *)
 val create_string_unchecked : Asm_section.t -> string -> t
 
+(* CR sspies: Can we say more about where this special label comes from? *)
```
spiessimon, line 58: Any suggestions on better naming here?

spiessimon:
> Excellent, although I'd like to double check the reason for the removal of the following CR: `(* CR sspies: Potentially enable range check again here. We currently emit numbers like 158 as an 8-bit constant, which breaks int8 range checks, but make the assembler perfectly happy. *)` I think commits could be added on this PR to remove the old directives files and do the renames, although fine in a separate PR too.

As discussed offline, I removed the constructor `Constant_with_width.create` from the `.mli`, since it was not used in the backend.

PR #3954, opened 2025-04-30
Convert flambda2 indexing operations to nativeints

jvanburen:


backend/cmm_helpers.ml
```diff
@@ -682,7 +681,10 @@ let rec mul_int c1 c2 dbg =
 
 let tag_int i dbg =
   match i with
-  | Cconst_int (n, _) -> int_const dbg n
+  | Cop (Clsr, [c; Cconst_int (n, _)], _) when n > 0 && is_defined_shift n ->
```
jvanburen, line 684: we need to re-optimize this case for array_length, which does a tag right after a right shift.

PR #3955, opened 2025-04-30
Initial stab at mode documentation

ccasin:
This adds some high-level documentation for our modes. The idea is to have a page that gives a brief overview of the whole suite of modes, without getting too far into the details of any one of them (for which it refers to the documentation about that particular feature).

I haven't attempted to write docs about the following axes: statefulness, visibility, yielding. Either someone else can write those and push them here, or they can be added in a future PR.

### DIFF
--- a/jane/doc/extensions/_04-modes/intro.md
+++ b/jane/doc/extensions/_04-modes/intro.md
@@ -4,12 +4,184 @@ collectionName: Modes
 title: Intro
 ---
 
+<style>
+.table {
+    width: fit-content;
+    margin-left: auto;
+    margin-right: auto;
+    margin-bottom: 20px;
+    border-style: solid;
+    border-color: blue;
+    border-radius: 25px;
+    padding: 15px;
+    text-align: center;
+}
+</style>
+
 # Introduction to the Mode System
 
-The locality mode used for [Stack allocation](../stack/intro) has been
-extended to more axes, each tracking a property of values, in order to support
-features such as [Parallelism](../parallelism/intro) and
-[Uniqueness](../uniqueness/intro). This documentation gives a general
-introduction to the mode system.
+Modes are deep properties of values that are tracked by the OxCaml type
+system. Each mode is associated with a particular operation that may be
+performed on a value. The mode may be a *past* mode, which tracks whether the
+operation has happened to this value in the past, or a *future* mode, which
+tracks whether the operation is allowed to happen to this value in the future.
 
-*This is a stub. Documentation will come soon.*
+This page shows the modes that are currently supported. Each mode belongs to a
+modal *axis* determined by the operation it tracks and whether it is a past or
+future mode. The axes on this page are arranged with the least mode at the
+bottom and the greatest mode at the top. The type system supports *submoding*:
+values may move freely to greater modes (which typically restrict what can be
+done with those values) but not to lesser modes.
+
+Each axis has a *legacy* mode, shown in bold. This is the "default" mode, and is
+chosen to make the modal type system backwards compatible with legacy OCaml
+programs.
+
+* [Modes for scope (locality)](#locality)
+* [Modes for moving between threads (portability and
+  contention)](#portability-contention)
+* [Modes for aliasing (uniqueness and linearity)](#uniqueness-linearity)
+
+# Modes for scope {#locality}
+
+## Future modes: Locality
+
+|------------|
+| local      |
+| `|`        |
+| **global** |
+{: .table }
+
+Locality is a future axis that tracks whether a value is allowed to escape its
+*region*. Regions are scopes created by standard OCaml language constructs: each
+function's body is a region, as are loop bodies.
+
+The type checker does not allow values that are *local* to escape their scope
+(for example, by being returned from a function or stored in a global
+ref). Values that are *global* may freely escape their scope. The compiler can
+stack allocate values that are local.
+
+Locality is irrelevant for types that never cause allocation on the OCaml heap,
+like int. Values of such types *mode cross* on the locality axis; they may be
+used as global even when they are local.
+
+More documentation for locality and stack allocation is available
+[here](../stack-allocation/intro).
+
+# Modes for moving between threads {#portability-contention}
+
+## Past modes: Contention
+
+|-----------------|
+| contended       |
+| `|`             |
+| shared          |
+| `|`             |
+| **uncontended** |
+{: .table }
+
+Contention is a past axis that tracks whether a value has been shared between
+threads. A value is *contended* if multiple threads have full access to it,
+*shared* if multiple threads have read-only access to it, and *uncontended* if
+only one thread has access to it.
+
+To enforce data race freedom, the typechecker does not permit reading or writing
+the mutable portions of contended values. The mutable portions of shared values
+may be read, but not written to. Uncontended values may be accessed and mutated
+freely.
+
+Contention is irrelevant for types that are deeply immutable. Values of such
+types *mode cross* on the contention axis; they may be used as uncontended even
+when they are contended.
+
+## Future modes: Portability
+
+|-----------------|
+| **nonportable** |
+| `|`             |
+| portable        |
+{: .table }
+
+Portability is a future axis that tracks whether a value is permitted to be
+shared with another thread. The type checker does not allow *nonportable* values
+to move across thread boundaries, while *portable* values may move freely.
+
+Portability is about functions: functions that capture uncontended mutable state
+are not portable.
+
+Notably, it is generally safe to send mutable data *itself* to other threads,
+because it will then be *contended*, so the mutable portions will be
+inaccessible. What is scary is to send a function that *captures* uncontended
+mutable data to another thread, because the captured data would remain
+uncontended even when the function is shared. When the second thread runs the
+funtion, both threads would be accessing the same uncontended mutable state (a
+data race!).
+
+Portability is irrelevant for types that do not contain functions. Values of
+such types *mode cross* on the portability axis; they may be used as portable
+even when they are nonportable.
+
+# Modes for aliasing {#uniqueness-linearity}
+
+## Past modes: Uniqueness
+
+|-------------|
+| **aliased** |
+| `|`         |
+| unique      |
+{: .table }
+
+Uniqueness is a past axis that tracks whether there are multiple references to a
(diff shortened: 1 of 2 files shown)


jane/doc/extensions/_04-modes/intro.md
```diff
@@ -4,12 +4,184 @@ collectionName: Modes
...
+## Future modes: Portability
+
+|-----------------|
+| **nonportable** |
+| `|`             |
+| portable        |
+{: .table }
+
+Portability is a future axis that tracks whether a value is permitted to be
+shared with another thread. The type checker does not allow *nonportable* values
```
goldfirere, line 106:
```suggestion
shared with another thread. OxCaml's parallelism API does not allow *nonportable* values
```

jane/doc/extensions/_04-modes/intro.md
```diff
@@ -4,12 +4,184 @@ collectionName: Modes
...
+| once     |
+| `|`      |
+| **many** |
+{: .table }
+
+Linearity is a future axis that tracks whether a function is permitted to be
+aliased.  Values that are *many* may used multiple times, while values that are
+*once* may only be used once.
+
+Linearity is about functions: its purpose is to track unique values in
```
goldfirere, line 174:
```suggestion
Like portability, linearity is about functions: its purpose is to track unique values in
```

goldfirere (approved):

ccasin:
This is ready to merge, but let's wait until ccasin/flambda-backend#3 is merged into it before hitting the button.

goldfirere:
@cuihtlauac should review [Add paragraph on the relationship between modes and types](https://github.com/ocaml-flambda/flambda-backend/pull/3955/commits/70974ff4757da2a7895308fa4fe35574a2ec94af).

jane/doc/extensions/_04-modes/intro.md
```diff
@@ -26,6 +26,16 @@ performed on a value. The mode may be a *past* mode, which tracks whether the
...
 
+Just like a value has a type, a value in OxCaml also has a mode. Types do *not*
+have a mode. That is, we do not have a type `string @ local` (say), but rather
+if we have `(x : string @ local)`, then `x` has type `string` and is at mode
+`local`. Modes also appear in the argument and return slots of a function type,
+so we can have `string @ local -> string option @ global` to describe a function
+whose argument will have the `local` mode and whose return will have the
+`global` mode. These modes are properly associated with the function type, not
+the argument or result types (that is, there is still no `string @ local` or
+`string option @ global`).
```
cuihtlauac, line 37:
```suggestion
`global` mode. These modes are associated with the function's behavior, not
the argument or result types (that is, there is still no `string @ local` or
`string option @ global`).
```

jane/doc/extensions/_04-modes/intro.md
```diff
@@ -4,12 +4,198 @@ collectionName: Modes
...
-The locality mode used for [Stack allocation](../stack/intro) has been
-extended to more axes, each tracking a property of values, in order to support
-features such as [Parallelism](../parallelism/intro) and
-[Uniqueness](../uniqueness/intro). This documentation gives a general
-introduction to the mode system.
+Modes are deep properties of values that are tracked by the OxCaml type
+system. Each mode is associated with a particular operation that may be
+performed on a value. The mode may be a *past* mode, which tracks whether the
+operation has happened to this value in the past, or a *future* mode, which
+tracks whether the operation is allowed to happen to this value in the future.
```
cuihtlauac, line 27:
```suggestion
Modes are properties of values and are tracked by the OxCaml compiler. They are inferred or read from annotations attached to declarations. Modes are checked. Modes have similarities and relationships with types, but they are not part of them. Types describe what the data is, while modes describe how it is used. Each mode is associated with a particular operation that may be performed on a value. The mode may be a *past* mode, which tracks whether the operation has happened to this value in the past, or a *future* mode, which tracks whether the operation is allowed to happen to this value in the future. Modes are deep; when attached to structured data, they apply to components, all the way through. Not all modes apply to all values. Some modes are irrelevant for some data types. A function's type may also be mode-annotated, in parameter or return position. However, this does not mean modes are attached to the function's type; it means the function respects some mode-relevant behavior. 
```

jane/doc/extensions/_04-modes/intro.md
```diff
@@ -4,12 +4,198 @@ collectionName: Modes
...
+so we can have `string @ local -> string option @ global` to describe a function
+whose argument will have the `local` mode and whose return will have the
+`global` mode. These modes appear in the function's type but are associated with the function's behavior, not the argument or result types (that is, there is still no `string @ local` or `string option @ global`).
+
+This page shows the modes that are currently supported. Each mode belongs to a
+modal *axis* determined by the operation it tracks and whether it is a past or
+future mode. The axes on this page are arranged with the least mode at the
+bottom and the greatest mode at the top.
+
+The type system supports *submoding*: values may move freely to greater modes
```
cuihtlauac, line 42:
```suggestion
The mode checking system supports *submoding*: values may move freely to greater modes
```

jane/doc/extensions/_04-modes/intro.md
```diff
@@ -4,12 +4,198 @@ collectionName: Modes
...
+(which typically restrict what can be done with those values) but not to lesser
+modes. Additionally, the type system knows that some types don't have
```
cuihtlauac, line 44:
```suggestion
modes. Additionally, the mode checking system knows that some types don't have
```

jane/doc/extensions/_04-modes/intro.md
```diff
@@ -4,12 +4,198 @@ collectionName: Modes
...
+interesting interactions with some modes. Such types are said to *mode cross* on
+those axes, which means values of these types may freely move in either
+direction on the axis. The sections for each axis below describe which types can
+mode cross on that axis.
+
+Each axis has a *legacy* mode, shown in bold. This is the "default" mode, and is
+chosen to make the modal type system backwards compatible with legacy OCaml
```
cuihtlauac, line 51:
```suggestion
chosen to make the modal system backwards compatible with legacy OCaml
```

jane/doc/extensions/_04-modes/intro.md
```diff
@@ -4,12 +4,198 @@ collectionName: Modes
...
+| local      |
+| `|`        |
+| **global** |
+{: .table }
+
+Locality is a future axis that tracks whether a value is allowed to escape its
+*region*. Regions are scopes created by standard OCaml language constructs: each
+function's body is a region, as are loop bodies.
+
+The type checker does not allow values that are *local* to escape their scope
```
cuihtlauac, line 73:
```suggestion
The mode checker does not allow values that are *local* to escape their scope
```

jane/doc/extensions/_04-modes/intro.md
```diff
@@ -4,12 +4,198 @@ collectionName: Modes
...
+| shared          |
+| `|`             |
+| **uncontended** |
+{: .table }
+
+Contention is a past axis that tracks whether a value has been shared between
+threads. A value is *contended* if another thread can write to it, *shared* if
+multiple threads have read-only access to it, and *uncontended* otherwise.
+
+To enforce data race freedom, the typechecker does not permit reading or writing
```
cuihtlauac, line 101:
```suggestion
To enforce data race freedom, the mode checker does not permit reading or writing
```

jane/doc/extensions/_04-modes/intro.md
goldfirere: Hmm.... I don't think we should break out "mode checker" from "type checker". Because of the modes on arrows, it would be impossible to implement without the other. (Well, maybe that's not true. It might be possible to do all type checking followed by mode checking, but it would be awkward in some cases, I think.) However, we should clarify that we use the term "type checker" inclusively to refer to checking both modes and types. (We could rename it the "type-and-mode-checker" or something, but I think that makes things worse.)

jane/doc/extensions/_04-modes/intro.md
goldfirere: > Modes are deep properties of values that are tracked by the OxCaml compiler. Like types, they are inferred from definitions and checked for consistency. (We use the term "type checking" to include both traditional type checking/inference and mode checking/inference.) Modes have similarities and relationships with types, but remain distinct: types are not modes, modes are not types, types do not have modes, and modes do not have types. Types describe what the data *is*, while modes describe how it is *used*.
> 
> Each mode is associated with a particular operation that may be performed on a value. The mode may be a *past* mode, which tracks whether the operation has happened to this value in the past, or a *future* mode, which tracks whether the operation is allowed to happen to this value in the future. Modes are deep; when attached to structured data, they apply to components, recursively. (OxCaml's *modality* feature cuts off the deepness of a mode; modalities can be placed on record and constructor fields.)

?

I've removed the last couple of sentences. It's true that some modes do not apply to some types, but I think that's covered below at the introduction of mode crossing. And modes *are* attached to a function's type: we should think of `->` has having four arguments: the argument type, the argument mode, the return type, and the return mode. I tried to get at this in the end of the next paragraph; do you see a way of making this clearer?

cuihtlauac: I find “irrelevant” easier to understand than “crossing.” But I mostly think this concept should be stressed more. Right now, it feels like a detail, while, to my understanding, there is a strong relationship between types and mode-controlled operations.

> we should think of -> has having four arguments: the argument type, the argument mode, the return type, and the return mode.

Aren't modes in function types behavior underspecifications? I see this as a striking feature of OxCaml.

goldfirere: I'm not sure what you mean by underspecifications. If you mean that there's some kind of subtyping relationship, that's true, but it's not robust. Maybe we can have a comment saying that we try to convert among arrow types when passing a function to another function, but it's not really a designed feature of the type system.

Maybe "irrelevant" is better, but we use the term "crossing" pretty widely. I think if we want to move away from "mode crossing", we should make that a more consistent change. Also, the modes are relevant for all types. For example, `(int @ nonportable -> bool) -> string` is distinct from `(int @ portable -> bool) -> string`. (Test it!) It's just that when we have a `int @ nonportable`, we can use it in a context that's expecting something portable -- that's mode crossing.

(It would be safe to treat those two types as equal, but that's not what we have today.)

PR #3960, opened 2025-05-02
Fix miscompilation / liveness errors for string operations

mshinwell:
This fixes the random CI failures seen on arm64.

It turns out that the Cmm machtypes given to "pointer + offset" calculations used for all of the different varieties of string reads and writes are wrong.  Some of them have been wrong for a very long time (30 years).  They are currently given type `Int`, but in the case where the pointers may be into the OCaml heap (e.g. for `string` and `bytes` rather than `Bigstring.t`) they must have type `Addr`.  Otherwise, CSE may apply, causing pointers into the middle of values to be live across a GC.  For example in `warnings.ml`:
```ocaml
  and loop_letter_num tokens modifier i =
    if i >= String.length s then error () else
    match s.[i] with  (* first string access *)
    | '0' .. '9' ->
        let i, n1, n2 = get_range i in
        loop (Num(n1,n2,modifier)::tokens) i
    | 'A' .. 'Z' | 'a' .. 'z' ->
       loop (Letter(s.[i],Some modifier)::tokens) (i+1)  (* repeated string access after potential GC *)
    | _ -> error ()
```

This patch affects the system compiler as well as the flambda-backend compiler, hence the use of a patch file to fix the former.  I will deal with getting this fixed upstream.  The system compiler patch isn't currently as good as the flambda-backend patch since it doesn't special-case out-of-heap accesses.

There are a couple of other changes currently in here too, which can be discussed and maybe split:
- fixes to stop the system compiler applying `Comballoc` across `Ccheckbound`, which can raise an exception.  I need to discuss this with @stedolan though, maybe this is only required if the allocation is on the major heap?  (This problem does not arise with the flambda-backend compiler since bounds checks are expanded before Flambda 2.)
- fixes to silence the debug runtime with the system compiler, which I think was causing excessive CI logs in some cases.

This bug was very troublesome to track down and @xclerc must be thanked for his help with this, especially since we thought there was a second bug, which was actually caused by a mistake in my hack for tracking down the first bug!  Some of the CI changes in this PR were written by @xclerc .

Once the dust has settled I will re-enable the arm64 builds as required.

### DIFF
--- a/backend/cmm_helpers.ml
+++ b/backend/cmm_helpers.ml
@@ -404,6 +404,15 @@ let rec sub_int c1 c2 dbg =
         add_const (sub_int c1 c2 dbg) n1 dbg
       | c1, c2 -> Cop (Csubi, [c1; c2], dbg))
 
+let add_int_addr c1 c2 dbg = Cop (Cadda, [c1; c2], dbg)
+
+let add_int_ptr ~ptr_out_of_heap c1 c2 dbg =
+  (* The [add_int_addr] case is only used for string access, and it seems
+     unlikely that the optimizations done for [add_int] will apply.
+
+     For out-of-heap accesses, we use [add_int], thus allowing more CSE. *)
+  if ptr_out_of_heap then add_int c1 c2 dbg else add_int_addr c1 c2 dbg
+
 let neg_int c dbg = sub_int (Cconst_int (0, dbg)) c dbg
 
 (** This function conservatively approximates the number of significant bits in its signed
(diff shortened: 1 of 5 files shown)


backend/cmm_helpers.ml
```diff
@@ -404,6 +404,15 @@ let rec sub_int c1 c2 dbg =
         add_const (sub_int c1 c2 dbg) n1 dbg
       | c1, c2 -> Cop (Csubi, [c1; c2], dbg))
 
+let add_int_addr c1 c2 dbg = Cop (Cadda, [c1; c2], dbg)
+
+let add_int_ptr ~ptr_out_of_heap c1 c2 dbg =
+  (* The [add_int_addr] case is only used for string access, and it seems
+     unlikely that the optimizations done for [add_int] will apply.
```
Gbury, line 411: Even if it's maybe not that frequent, I think that at least the optimization for when the index / offset added is `0` might be useful ?

backend/cmm_helpers.ml
```diff
@@ -2333,34 +2355,55 @@ let unaligned_set_16 ptr idx newval dbg =
...
+              (cconst_int 1) dbg ],
           dbg )
     in
     let v3 =
       Cop
         ( mk_load_mut Byte_unsigned,
-          [add_int (add_int ptr idx dbg) (cconst_int 2) dbg],
+          [ add_int_addr
+              (add_int_ptr ~ptr_out_of_heap ptr idx dbg)
+              (cconst_int 2) dbg ],
```
Gbury, line 2398: maybe for a later PR, but maybe we could instead do:
```ocaml
(add_int_ptr ~ptr_out_of_heap ptr (add_int idx 2 dbg) dbg)
```
which would at least optimize the case where `idx` is a constant.

Gbury (approved):
Oh I see, that's a very tricky bug, congratulations for tracking it down !

That actually reminds me of earlier work I had done on automatically computing the machtypes of additions, and related suggestions I've had for more precise machtype and automated checking in cmm, to ensure that this doesn't happen again.

backend/cmm_helpers.ml
mshinwell: Ah yes, good point!

backend/cmm_helpers.ml
mshinwell: Ah there's a mistake here as well, the `add_int_addr` was supposed to be `add_int_ptr`.

mshinwell: I've left a CR for the rearrangement though.

mshinwell:
Upstream PR: https://github.com/ocaml/ocaml/pull/14010

mshinwell:
@Gbury Agreed we should discuss the machtype checking (etc) more.

PR #3961, opened 2025-05-02
`Modality.Value.Const` should behave like a product

riaqn:
Based on #3949 

This PR refactors the interface of `Modality.Value.Const` to make it behave like a product of axes (that you can project and update specific axis) instead of a sequence of atoms (that you compose).

This allows other incoming PRs that do the following:
- remove `[@no_mutable_implied_modalities]`, since we can spell out modalities to override those
- fix the behavior of default modalities, so the overriding is per-axis.
- (done in this PR) warnings for specifying the same modality axis multiple times in a single modality expression. 
- warnings for superfluous overriding (e.g., overriding `portable` with `portable`)

# Review

@dkalinichenko-js , could you look at the changes in `typemode.ml` and tests. In particular, could you check that the implication and the print back is still as expected. Feel free to add more tests.

### DIFF
--- a/typing/mode.ml
+++ b/typing/mode.ml
@@ -929,6 +929,18 @@ module Lattices_mono = struct
       | Visibility : (Monadic_op.t, Visibility_op.t) t
       | Contention : (Monadic_op.t, Contention_op.t) t
 
+    let to_int : type a b. (a, b) t -> int = function
+      | Areality -> 0
+      | Yielding -> 1
+      | Linearity -> 2
+      | Statefulness -> 3
+      | Portability -> 4
+      | Uniqueness -> 5
+      | Visibility -> 6
+      | Contention -> 7
+
+    let compare a b = to_int a - to_int b
+
     let print : type p r. _ -> (p, r) t -> unit =
      fun ppf -> function
       | Areality -> Format.fprintf ppf "locality"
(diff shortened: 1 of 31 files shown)


riaqn:
Thank you @d-kalinichenko ! I reviewed your fixes and they look great!
I did some further clean up, and it should be ready for review now.

typing/mode.ml
```diff
@@ -2349,10 +2369,11 @@ module Value_with (Areality : Areality) = struct
 
     let all =
```
dkalinichenko-js, line 2370: What is the intended order here? Should it match the order in `compare`?

typing/mode_intf.mli
```diff
@@ -373,16 +373,18 @@ module type S = sig
 
   module Axis : sig
     (** ('p, 'r) t represents a projection from a product of type ['p] to an
-    element of type ['r]. *)
+    element of type ['r].
+
+    NB: must listed in the order of axis implication. See [typemode.ml]. *)
```
dkalinichenko-js, line 378: Why?

dkalinichenko-js (approved):
LGTM modulo the order comments.

typing/mode.ml
riaqn: I think it currently doesn't have to, but yes it doesn't hurt to have it ordered.

typing/mode_intf.mli
riaqn: because `transl_modalities` relies on this fact to have implied modalities correct. If user write `yielding global`, we first reorder it to `global yielding`, so we see `yielding` later than `global` and override the `unyielding` implied by `global`.

dkalinichenko-js: Oh, right, I see.

PR #3962, opened 2025-05-02
Remove `@no_mutable_implied_modalities`

riaqn:
Based on #3961, we no longer need this attribute - we can write out modalities to override the `mutable`-implied ones.

### DIFF
--- a/typing/typemode.ml
+++ b/typing/typemode.ml
@@ -462,11 +462,10 @@ let untransl_modality (a : Modality.t) : Parsetree.modality loc =
 
 (* For now, mutable implies legacy modalities for both comonadic axes and
    monadic axes. In the future, implications on the comonadic axes will be
-   removed (and can be experimented currently with using
-   @no_mutable_implied_modalities). The implications on the monadic axes will
-   stay. *)
+   removed. The implications on the monadic axes will stay. Implied modalities
+   can be overriden. *)
 (* CR zqian: decouple mutable and comonadic modalities *)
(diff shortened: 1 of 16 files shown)


goldfirere:
In the process of reviewing this, I realized I wanted documentation around the semantics of modalities -- mostly so that I could check that the implementation met that specification. But we don't have any documentation! However, I don't think it makes sense to add this as part of this patch. Instead, I have added a note within #3943 to add more documentation when we fix a CR in that PR. (No action needed in response, just noting the connection to #3943.)

typing/typemode.ml
```diff
@@ -462,11 +462,10 @@ let untransl_modality (a : Modality.t) : Parsetree.modality loc =
 
 (* For now, mutable implies legacy modalities for both comonadic axes and
    monadic axes. In the future, implications on the comonadic axes will be
-   removed (and can be experimented currently with using
-   @no_mutable_implied_modalities). The implications on the monadic axes will
-   stay. *)
+   removed. The implications on the monadic axes will stay. Implied modalities
```
goldfirere, line 465: Is this still the plan (to remove the implied modalities)? I suppose this patch makes it no harder to do that, and it simplifies things in the meantime. Just checking to make sure this comment is still accurate; just Resolve this if it is.

goldfirere (approved):
🎉

typing/typemode.ml
riaqn: Yeah it's still our plan to remove the implied modalities and this PR doesn't make that harder.

PR #3964, opened 2025-05-02
Implement `let mutable`

jra4:
Implement `let mutable`.

This feature makes explicit an existing optimization which puts `ref`s in registers or the call stack (which eliminates an allocation). See `jane/doc/extensions/_08-miscellaneous-extensions/let-mutable.md` for more.

Most relevant functionality is tested in `testsuite/tests/typing-local/let_mutable.ml`, and other tests have been modified (e.g. `typing-local/alloc.ml` now makes sure `let mutable` does not allocate). Additional testing with unboxed types is done in `testsuite/tests/typing-layouts/let_mutable.ml`. Parse error tests have been added to `testsuite/tests/parse-errors`.

### DIFF
--- a/testsuite/tests/typing-layouts/let_mutable_alpha.ml
+++ b/testsuite/tests/typing-layouts/let_mutable_alpha.ml
@@ -20,12 +20,12 @@ let () =
   v <- u;
   print_endline "Hello, world!"
 
-type t = #{ x: int; v: void; y: int }
+type t = #{ x: int; v: void; y: int32# }
 
 let () =
-  let mutable r = #{ x = 10; v = unbox_unit (); y = 20 } in
-  r <- #{ x = 50; v = unbox_unit (); y = 60 };
+  let mutable r = #{ x = 10; v = unbox_unit (); y = #20l } in
+  r <- #{ x = 50; v = unbox_unit (); y = #60l };
   print_int r.#x;
   print_string " ";
-  print_int r.#y;
+  print_int (Stdlib_upstream_compatible.Int32_u.to_int r.#y);
   print_endline ""


lambda/translcore.ml
```diff
@@ -948,11 +952,14 @@ and transl_exp0 ~in_new_scope ~scopes sort e =
       let self = transl_value_path loc e.exp_env path_self in
       let var = transl_value_path loc e.exp_env path in
       Lprim(Pfield_computed Reads_vary, [self; var], loc)
+  | Texp_mutvar id -> Lmutvar id.txt (* jra: is this right? *)
```
ccasin, line 955: This looks right to me.

parsing/parsetree.mli
```diff
@@ -468,6 +476,7 @@ and expression_desc =
          *)
   | Pexp_send of expression * label loc  (** [E # m] *)
   | Pexp_new of Longident.t loc  (** [new M.c] *)
+    (* jra: should this be renamed? Both mutable vars and instance vars use this syntax *)
```
ccasin, line 479: Sounds like a good idea.  Maybe just `Pexp_setvar`?  Probably the comment on this case should note the two things it works on.

typing/typemod.ml
```diff
@@ -3100,7 +3100,8 @@ and type_structure ?(toplevel = None) funct_body anchor env sstr =
         Tstr_eval (expr, sort, attrs), [], shape_map, env
     | Pstr_value (rec_flag, sdefs) ->
         let (defs, newenv) =
-          Typecore.type_binding env rec_flag ~force_toplevel sdefs in
+          (* jra: should this be hard-coded? *)
+          Typecore.type_binding env Immutable rec_flag ~force_toplevel sdefs in
```
ccasin, line 3104: This looks correct to me - you've ruled out the possibility that structure-level bindings are mutable in the parser.

typing/env.mli
```diff
@@ -257,6 +257,7 @@ type lookup_error =
   | Non_value_used_in_object of Longident.t * type_expr * Jkind.Violation.t
   | No_unboxed_version of Longident.t
   | Error_from_persistent_env of Persistent_env.error
+  | Mutable_value_used_in_closure of string (* jra: Maybe rename this error/add other errors? *)
```
ccasin, line 260: Agreed - I think we'll want to generalize this now that we have more complicated locks - can discuss tomorrow.

typing/env.ml
```diff
@@ -4031,27 +4035,49 @@ let lookup_all_labels_from_type ?(use=true) ~record_form ~loc usage ty_path env
...
-      | Val_ivar(mut, cl_num) ->
+      match desc.val_kind, path with
+      | Val_ivar(mut, cl_num), _ ->
           use_value ~use ~loc path vda;
-          path, mut, cl_num, Subst.Lazy.force_type_expr desc.val_type
-      | _ ->
-          lookup_error loc env (Not_an_instance_variable name)
+          Instance_variable (path, mut, cl_num, Subst.Lazy.force_type_expr desc.val_type)
+      | Val_mut, Pident id ->
+          let rec mode_of_locks mode = function (* jra: surely this is incorrect *)
```
ccasin, line 4051: There's an ongoing discussion with @riaqn about this, but my guess is we want to do something like the inverse of the `walk_locks` function above.  Note how that function is used by `type_ident` in `Typecore`: basically, it tells you how you need to change the mode of something to push it down through locks from its declaration site to a use site. Here, we're doing this in the other direction - we want to know what mode we should require something to have so that it is safe to push it up through the locks from the place where the value is to the place where the mutable variable we're sticking it in is declared.

typing/typeclass.ml
```diff
@@ -1434,7 +1434,8 @@ and class_expr_aux cl_num val_env met_env virt self_scope scl =
          }
   | Pcl_let (rec_flag, sdefs, scl') ->
       let (defs, val_env) =
-        Typecore.type_let In_class_def val_env rec_flag sdefs in
+        (* jra: should this be hard-coded? *)
+        Typecore.type_let In_class_def val_env Immutable rec_flag sdefs in
```
ccasin, line 1438: This one looks good for a similar reason as the one in typemod.

typing/typecore.ml
```diff
@@ -1151,6 +1157,8 @@ type pattern_variable =
     pv_id: Ident.t;
     pv_uid: Uid.t;
     pv_mode: Value.l;
+    (* jra: I don't fully understand the difference between mutable_flag and mutability *)
+    pv_mutable: mutable_flag;
```
ccasin, line 1161: Not quite certain what you're referring to by "mutability" here, but I suspect there isn't a difference - we can discuss.

typing/typecore.ml
```diff
@@ -1954,7 +1974,9 @@ let type_for_loop_index ~loc ~env ~param =
             let pv_id = Ident.create_local txt in
             let pv_uid = Uid.mk ~current_unit:(Env.get_unit_name ()) in
             let pv =
-              { pv_id; pv_uid; pv_mode=Value.disallow_right pv_mode; pv_type; pv_loc; pv_as_var; pv_attributes }
+              { pv_id; pv_uid; pv_mode=Value.disallow_right pv_mode;
+                pv_mutable=Immutable; pv_type; pv_loc; pv_as_var; pv_attributes }
+              (* jra: is this right? See add_pattern_variables# *)
```
ccasin, line 1979: Seems right - we won't support mutable for loop index variables.

ccasin (commented):
Just a quick pass over some of the places where you had questions - we'll chat more tomorrow.

typing/typecore.ml
```diff
@@ -5692,6 +5746,11 @@ and type_expect_
                          match lid.txt with
                              Longident.Lident txt -> { txt; loc = lid.loc }
                            | _ -> assert false)
+        | Val_mut -> begin
+            match path with
+            | Path.Pident id -> Texp_mutvar {loc = lid.loc; txt = id}
+            | _ -> assert false
```
ccasin, line 5752: I know I wrote this, but I think it would be better to use `Misc.fatal_error` here.  A good rule of thumb, which I didn't understand at the time, is that `assert false` is fine when the invariant is enforced locally to the relevant code, but `Misc.fatal_error` is preferrable when the invariant in question relies on other functions or files, because someone might change those other places to violate the invariant in the future and then at least we get a helpful error message. (This rule is not consistently followed in older code in the compiler, but it's good to use in new code)

typing/typecore.ml
```diff
@@ -5749,7 +5808,7 @@ and type_expect_
         exp_type = type_constant cst;
         exp_attributes = sexp.pexp_attributes;
         exp_env = env }
-  | Pexp_let(Nonrecursive,
+  | Pexp_let(_, Nonrecursive,
```
ccasin, line 5811: I think we should match explicitly on `Immutable` here.  It's already true that this case will never fire for let mutable, because of invariants on what patterns we allow with `let mutable` and the definition of `turn_let_into_match`.  But it would be a bug if this transformation `did` happen to let mutable, so better to be explicit about it.

typing/typecore.ml
```diff
@@ -5826,8 +5890,15 @@ and type_expect_
           (* The "body" component of the scope escape check. *)
           unify_exp new_env body (newvar (Jkind.Builtin.any ~why:Dummy_jkind)))
       in
+      let exp =
+        match mutable_flag, pat_exp_list with
+        | Immutable, _ -> Texp_let(rec_flag, pat_exp_list, body)
+        | Mutable, [vb] -> Texp_letmutable(vb, body)
+        | Mutable, _ -> fatal_error "Typecore.type_expect"
+          (* Unreachable: should be prevented by [value_bindings_mutability] *)
```
ccasin, line 5898: (Just a note to update this comment if you do change the name of `value_bindings_mutability`)

typing/typecore.ml
```diff
@@ -6411,28 +6482,48 @@ and type_expect_
...
-            exp_type = instance Predef.type_unit;
-            exp_attributes = sexp.pexp_attributes;
-            exp_env = env }
-      | _ ->
-          raise(Error(loc, env, Instance_variable_not_mutable lab.txt))
-    end
+  | Pexp_setvar (lab, snewval) ->
+      let desc =
+        match Env.lookup_settable_variable ~loc lab.txt env with
+        | Instance_variable (path,Mutable,cl_num,ty) ->
```
ccasin, line 6488: As a general style rule, please put a space after the `,`s in tuples.

typing/typecore.ml
```diff
@@ -6411,28 +6482,48 @@ and type_expect_
...
+            let newval =
+              type_expect env
+                (mode_default (Mode.Value.max_with (Comonadic Areality)
+                  Mode.Regionality.global))
```
ccasin, line 6492: I expected the instance variable case to be ~unchanged by this PR - why has the mode here changed?

typing/typecore.ml
```diff
@@ -6411,28 +6482,48 @@ and type_expect_
...
+            let lid = {txt = id; loc} in
+            let sort =
+              match
+                Ctype.type_sort ~why:Jkind.History.Mutable_var_assignment
+                  ~fixed:false env newval.exp_type
+              with
+              | Ok sort -> sort
+              | Error _ ->
+                (* unreachable since [x <- e] was already type-checked,
+                   so [e] is representable *)
```
ccasin, line 6515: I think this isn't quite right. We're in the middle of checking `x <- e`, here.  It's unreachable because we checked `ty` is representable when we created the binding for the mutable variable.

Which, actually, gives me an idea: could we just remember the sort of this variable in the environment rather than recomputing it here?

typing/typecore.ml
```diff
@@ -9249,7 +9340,10 @@ and type_let ?check ?check_strict ?(force_toplevel = false)
         (fun (_, pat, _) (exp, _) ->
           if maybe_expansive exp then lower_contravariant env pat.pat_type)
         mode_pat_typ_list exp_list;
-      iter_pattern_variables_type generalize pvs;
+      iter_pattern_variables_type_mut
+        ~f_immut:generalize
+        ~f_mut:(unify_var env (newvar (Jkind.Builtin.any ~why:Mutable_value)))
```
ccasin, line 9345: As I mentioned elsewhere, I think we can use `Dummy_jkind` for the reason here.

ccasin (commented):
OK - I've now read everything except the modes bits we're going to discuss with @riaqn.

testsuite/tests/typing-local/let_mutable.ml
```diff
@@ -0,0 +1,385 @@
...
+   include stdlib_upstream_compatible;
+   expect; *)
+
+(* Test 1.1: basic usage in a for loop *)
+let foo1 y =
+  let mutable x = y in
+  for i = 1 to 10 do
+    x <- x + i
+  done;
+  (x : int)
```
ccasin: I think this type annotation, and the ones like it on similar examples below, are no longer needed.

jane/doc/extensions/_08-miscellaneous-extensions/let-mutable.md
```diff
@@ -0,0 +1,35 @@
+# The `let mutable` extension
+
+The `let mutable` extension provides a new type of `let` statement which
+declares a stack-local variable. It can be thought of as an unboxed `ref`.
```
ccasin: I think describing it as a "stack-local variable" is probably confusing in the presence of `local`.  I also don't love describing it as an "unboxed ref", because `ref#` is a different thing we're going to have very soon.

I think I would say something like:

> The `let mutable` extension provides a mechanism for creating mutable variables.  This codifies a pre-existing optimization, where the compiler attempts to eliminate actually allocating a box for a `ref` when it can see the `ref` is only used locally in a given scope, instead simply storing the value in a register.

jane/doc/extensions/_08-miscellaneous-extensions/let-mutable.md
```diff
@@ -0,0 +1,35 @@
...
+let triangle n =
+  let mutable total = 0 in
+  for i = 1 to n do
+    total <- total + i
+  done;
+  total
+```
+
+Mutable variables must not escape their scope. For example, you can't
+return a closure that closes over a mutable variable. At the moment, the mode
```
ccasin: It's true you can't return a closure that closes over a mutable variable, but the restriction is more severe than that: you can't capture a mutable variable in a closure, period.  I think I would just say that, e.g., "For example, you can't capture a mutable variable in a closure".

ccasin: I think it would also be good to clarify that you _can_ store local things in a mutable variable. You could do something like this, for example:

```ocaml
let triangle_list n =
  let mutable to_sum = [] in
  for i = 1 to n do exclave_
    to_sum <- stack_ (i :: to_sum)
  done;
  List.sum (module Int) to_sum
```

Or maybe you can come up with an example where the intermediate data structure is more useful.

jane/doc/extensions/_08-miscellaneous-extensions/let-mutable.md
```diff
@@ -0,0 +1,35 @@
...
+checker is, sadly, not sophisticated enough to allow some constructions which
```
ccasin: I think the restriction here is not really about the mode checker (and I don't think it would be hard to implement the mode checking behavior to allow mutable variables to be captured).  The problem is just that we have no way to compile that - no notion of a captured mutable variable.  The mutable variable would need to go on the stack, or something, and the closure would have to contain a reference to it.

We should acknowledge this limitation, but not link it to the mode checker.  I also think we can probably move it in to the "restrictions" section.

> Because closures may not capture mutable variables, some uses that are apparently safe from a scope perspective are not possible.  For example, the following program is rejected: ...

testsuite/tests/ast-invariants/test.ml
```diff
@@ -63,7 +63,9 @@ let kind fn =
 (* some test directories contain files that intentionally violate the
    expectations of ast-invariants *)
 let is_ok_dir dir =
-  not (String.ends_with ~suffix:"tests/jane-modular-syntax" dir)
+  (* CR jrayman: this first directory doesn't seem to exist? *)
+  not (String.ends_with ~suffix:"tests/jane-modular-syntax" dir) &&
+  not (String.ends_with ~suffix:"tests/parse-errors" dir)
```
ccasin, line 66: Indeed - I tihnk this test is out of date and refers to a directory we deleted a while ago.  But I don't think it should be necessary to add `parse-errors` here, because this test is just for files that parse, and those files don't.  Does something go wrong if you leave this unchanged?

testsuite/tests/typing-layouts/let_mutable.ml
```diff
@@ -0,0 +1,84 @@
...
+let triangle_f64 n =
+  let mutable sum = #0.0 in
+  for i = 1 to n do
+    sum <- Float_u.add sum (Float_u.of_int i)
+  done;
+  sum
+
+let () = Printf.printf "%.2f\n" (triangle_f64 10 |> Float_u.to_float)
+
+(* CR jrayman: [Float32_u] is wrong. What is it supposed to be? *)
```
ccasin: I think the issue is just that you've opened `Stdlib_upstream_compatible`, but `float32` isn't upstream compatible - it is in `Stdlib_stable`

testsuite/tests/typing-layouts/let_mutable.ml
```diff
@@ -0,0 +1,84 @@
...
+  let mutable sum = #0l in
+  for i = 1 to n do
+    sum <- Int32_u.add sum (Int32_u.of_int i)
+  done;
+  sum
+
+let () = Printf.printf "%d\n" (triangle_i32 10 |> Int32_u.to_int)
+
+
+(* CR jrayman: how do you create a vec128? *)
```
ccasin: See the simd tests, e.g., https://github.com/oxcaml/oxcaml/blob/main/oxcaml/tests/simd/basic.ml

testsuite/tests/typing-local/let_mutable.ml
```diff
@@ -0,0 +1,394 @@
...
+  done; x
+
+[%%expect{|
+Line 3, characters 13-29:
+3 |   while x <- (local_ (x + 1)); x <= 100 do
+                 ^^^^^^^^^^^^^^^^
+Error: This value escapes its region.
+|}]
+
+(* exclave_ closes one region, not two *)
```
ccasin, line 206: Let's also test that if you put `exclave_` in both places, it does work.

testsuite/tests/typing-local/let_mutable.ml
```diff
@@ -0,0 +1,394 @@
...
+    done
+  done;
+  x
+;;
+[%%expect{|
+Line 5, characters 11-30:
+5 |       x <- local_ ((i*j) :: x)
+               ^^^^^^^^^^^^^^^^^^^
+Error: This value escapes its region.
+|}]
```
ccasin, line 237: Let's also test these examples from slack the other day:

```ocaml
let f () =
  let mutable x = [] in
  (x <- stack_ 1 :: []);
  x
```
and
```ocaml
let f b =
  let mutable x = [] in
  (x <- if b then 2 :: [] else stack_ 1 :: []);
  x
```

These should both be rejected because the stack allocated thing potentially escapes

testsuite/tests/typing-local/let_mutable.ml
```diff
@@ -0,0 +1,394 @@
...
+let x_13_3 = ref 0
+let y_13_3 =
+  let mutable x @ local = ref (ref 0) in
+  x := x_13_3;
+  x <- ref x_13_3;
+  !x
+[%%expect{|
+val x_13_3 : int ref = {contents = 0}
+val y_13_3 : int ref = {contents = 0}
+|}]
```
ccasin, line 452: I'd like to test some other interactions with modes here.  In particular, I think if you have `let mutable x @ m = e in ...` you get a check that `e` has mode `m`, but not that other values `x` is updated with in the future have this mode.  And the mode of `x` is correctly updated for values put in it in the future.

This behavior seems fine, but we should have a test documenting it.  For example, consider:
```ocaml
let require_portable (f : (int -> unit) @ portable) = ()
let x = ref 42

(* let _ : int -> unit = require_portable (fun z -> x := z) *)

let allowed1 =
  let mutable f @ portable = fun _ -> () in
  (f <- fun z -> ());
  require_portable f

let allowed2 =
  let mutable f @ portable = fun _ -> () in
  (f <- fun z -> x := z)

let disallowed1 =
  let mutable f = fun _ -> () in
  (f <- fun z -> x := z);
  require_portable f

let disallowed2 =
  let mutable f = fun _ -> () in
  (f <- fun z -> x := z);
  (f <- fun z -> ());
  require_portable f
```

here, `allowed1` shows we know that an update with a portable function preserves `f`'s portability, `allowed2` shows we _do_ allow updating `f` with a non-portable function, `disallowed1` shows we do track correctly when something non-portable has been put in, and `disallowed2` shows the mutable variable's mode just reflects the worst thing put in it.

typing/env.ml
```diff
@@ -3300,9 +3301,49 @@ let walk_locks ~errors ~loc ~env ~item ~lid mode ty locks =
           vmode
     ) vmode locks
 
+(** Take the parameter of [mutable(m0)] at declaration site,  *)
```
ccasin: I find this comment a little confusing.  This function is called at places where the variable is updated, not its declaration site.  Maybe the comment is just referring to the `mode` argument?  But I think it would be good to have a comment that more completely describes the function's behavior. (cc @riaqn)

testsuite/tests/typing-local/let_mutable.ml
```diff
@@ -0,0 +1,394 @@
...
+
+let foo5_4 y = (* Assign of local works in _local_ while cond region *)
+  let mutable x = y in
+  while exclave_ x <- (local_ (x + 1)); x <= 100 do
+    x <- x + x
+  done; (x : int)
+
+[%%expect{|
+val foo5_4 : int -> int = <fun>
+|}]
```
ccasin, line 328: I think we should also have some tests for "regional".  If I understand correctly, this example is allowed, because we allow the expected mode of the updated value to be regional when the update happens one region away from the declaration site.
```ocaml
let allowed =
  let mutable x = [] in
  let y = stack_ (1 :: []) in
  for i = 0 to 1 do
    x <- y
  done
```

On the other hand, this example is (correctly) rejected - the list is regional, but because the assignment is two regions from the declaration site we require the list to be global:
```ocaml
let disallowed =
  let mutable x = [] in
  for i = 0 to 1 do
    let z = stack_ (2 :: []) in
    for j = 0 to 1 do
      x <- z
    done
  done
```

And this next example is rejected too - it's conservative (it would be safe to allow) but it's rejected for the same reason as the previous one (which seems fine):
```ocaml
let disallowed2 =
  let mutable x = [] in
  let y = stack_ (1 :: []) in
  for i = 0 to 1 do
    for j = 0 to 1 do
      x <- y
    done
  done
```
@riaqn , please check my reasoning here.

typing/types.ml
```diff
@@ -20,12 +20,17 @@ open Asttypes
...
+  | Mutable of Mode.Value.Comonadic.lr
 
 let is_mutable = function
   | Immutable -> false
   | Mutable _ -> true
 
+let mutable_mode m0 : _ Mode.Value.t =
+  { comonadic = m0
+  ; monadic = Mode.Value.Monadic.(min |> allow_left |> allow_right)
+  }
```
ccasin, line 34: @riaqn it's not obvious to me why this is right - maybe a comment?

typing/typecore.ml
```diff
@@ -1301,7 +1320,7 @@ let add_module_variables env module_variables =
   ) env module_variables_as_list
 
 let enter_variable ?(is_module=false) ?(is_as_variable=false) tps loc name mode
-    ty attrs =
+    ?(kind = Val_reg) ty attrs =
```
ccasin: I think in general optional parameters like this are pretty error prone and best avoided unless there's a good specific reason for them.  This function only has a few call sites, so I propose we just specify this at all of them.

typing/typecore.ml
```diff
@@ -2811,8 +2856,28 @@ and type_pat_aux
...
       in
+      let mode, kind =
+        match mutable_flag with
+        | Immutable -> alloc_mode, Val_reg
+        | Mutable ->
+            let m0 = Value.Comonadic.newvar () in
+            let mode = mutvar_mode ~loc ~env:!!penv m0 alloc_mode in
+            let kind =
+              Val_mut
+              (* CR-someday let_mutable: move the sort calculation elsewhere *)
```
ccasin: I think we should probably delete this comment - if we left a comment for every small thing we wish there were another way to do, the typechecker would be 90% comments.

It's not crazy to leave it, but I'd suggest then making it not a CR, just a comment, and being more specific: the thing that is a little sad is that we have to check all pattern variables are representable, so we're already calculating the sort of this elsewhere, but it's too far removed to easily pass that information here.

typing/typecore.ml
```diff
@@ -2811,8 +2856,28 @@ and type_pat_aux
...
+              (m0,
+              match
```
ccasin, line 2886: I think having a large expression like this as an element of a pair makes this a bit hard to read, and suggest pulling it out into a let binding.

jane/doc/extensions/_08-miscellaneous-extensions/let-mutable.md
jra4: I think your example is fine. Even though it is a bit nonsensical, it demonstrates the functionality clearly. The smallest "useful" example I could think of is this Forth interpreter, where it's hard to see the locals being used.

```ocaml
let eval program =
  let mutable program = program in
  let mutable operands = [] in
  while not (List.is_empty program) do exclave_
    (match List.hd program with
     | Add -> (match operands with
               | lhs :: rhs :: tl -> operands <- stack_ ((lhs + rhs) :: tl)
               | _ -> raise Not_enough_operands)
     | Push x -> operands <- stack_ (x :: operands));
    program <- List.tl program
  done;
  match operands with
  | x :: _ -> x
  | _ -> raise Not_enough_operands
;;
```

Also, does your example compile? In the last line you pass `to_sum` which is local, to `List.sum`, which does not have an `@ local` annotation on its arguments. Shouldn't you get a `This value escapes its region.` error?

jane/doc/extensions/_08-miscellaneous-extensions/let-mutable.md
jra4: I've updated `let-mutable.md` according to the 3 comments you left. Feel free to mark these discussions as resolved if you are satisfied with the changes.

testsuite/tests/ast-invariants/test.ml
jra4: Yes. [See this test failure](https://github.com/oxcaml/oxcaml/pull/3964/commits/324f6b221370a288acab2b68b3786759e1431d54#diff-1d6f93d1ab3ee9d2c26938e4a6d7bc863ae064bc309c4187aabc1dab8c8a582b)

typing/typecore.ml
jra4: Changed the comment

testsuite/tests/typing-local/let_mutable.ml
jra4: Added these tests. Currently, the compiler implements the behavior you described

testsuite/tests/typing-local/let_mutable.ml
jra4: Added these tests

testsuite/tests/typing-local/let_mutable.ml
jra4: Also, should we add `disallowed2` to the "Limitations" section of `let-mutable.md`?

lthls:
Don't forget monadic operators:
```ocaml
let (let*) x f = f x in
let mutable x = 42 in
let* y = 0 in
x + y
```
(This example should be forbidden as `x` is captured by the implicit closure for `y`)

typing/env.ml
riaqn: Marking this as resolved, as this has been addressed in the newer commits. (further comments should go there)

testsuite/tests/typing-local/let_mutable.ml
riaqn: @ccasin I agree with all of what you said.

typing/env.ml
riaqn: My apologies - the comment is just incomplete...

jane/doc/extensions/_08-miscellaneous-extensions/let-mutable.md
ccasin: Good point about `List.sum` and locality.  Probably we should make the example work, which I think you can do by just writing a very simple sum function that takes the list locally and including it in the example.

jane/doc/extensions/_11-miscellaneous-extensions/let-mutable.md
```diff
@@ -0,0 +1,51 @@
...
+following program is rejected:
+
+```ocaml
+let sum xs =
+  let mutable total = 0 in
+  List.iter xs ~f:(fun x -> total <- total + x);
+  total
+```
+
+Unboxed products are not yet supported.
```
ccasin: Maybe say explicitly you mean mutable variables may not yet contain unboxed products.

mshinwell:
The unboxed product support is now in https://github.com/oxcaml/oxcaml/pull/4184 (which will merge into this PR).

PR #3966, opened 2025-05-05
Implement vec256/512

TheNumbat:


backend/x86_binary_emitter.ml
```diff
@@ -868,12 +868,12 @@ let emit_MOV b dst src =
 
 let emit_vex3 buf ~rexr ~rexx ~rexb ~vex_m ~vex_w ~vex_v ~vex_l ~vex_p =
   buf_int8 buf 0xC4; (* We only emit 3-byte VEX instructions. *)
-  buf_int8 buf (((lnot rexr) lsl 7) lor
-                ((lnot rexx) lsl 6) lor
-                ((lnot rexb) lsl 5) lor
+  buf_int8 buf (((rexr lxor 1) lsl 7) lor
```
TheNumbat, line 871: This is a bugfix

TheNumbat:
Review:
- @ccasin - `typing`, `toplevel`, `testsuite`, `parsing`, `lambda`, `.github`, `file_formats`, `flambda-backend`
- @mshinwell - `middle_end`
- @jvanburen - `backend`, `tools`
- @NickBarnes - `runtime`, `runtime4`

backend/x86_binary_emitter.ml
TheNumbat: No, I'll make another pr

ccasin (commented):
I've reviewed the directories assigned to me and they look fine, modulo my comments above - mainly I think we need to document clearly somewhere all the bits that are missing before we can consider moving this out of beta.

mshinwell:
The `middle_end/` changes are ok, I pushed a few small fixes.

mshinwell:
New review assignment: I will read the Cmm parts, @xclerc will read the remainder of the backend.

backend/cmmgen_state.ml
```diff
@@ -27,8 +27,24 @@ type ustructured_constant =
   | Const_int64 of int64
   | Const_nativeint of nativeint
   | Const_vec128 of
-      { high : int64;
-        low : int64
+      { word0 : int64;
```
xclerc: (Maybe duplicate the comment about
`word0` being the least significant?)

(If feels like the new names are slightly
worse, but I have unfortunately no
suggestions.)

backend/cmmgen_state.ml
```diff
@@ -95,13 +113,38 @@ let compare_structured_constants c1 c2 =
     if c <> 0 then c else compare_constant_lists l1 l2
   | Const_float_array l1, Const_float_array l2 -> compare_float_lists l1 l2
   | Const_string s1, Const_string s2 -> String.compare s1 s2
-  | Const_vec128 { high = l0; low = l1 }, Const_vec128 { high = r0; low = r1 }
-    ->
-    let cmp = Int64.compare l0 r0 in
-    if cmp = 0 then Int64.compare l1 r1 else cmp
+  | ( Const_vec128 { word0 = l0; word1 = l1 },
+      Const_vec128 { word0 = r0; word1 = r1 } ) ->
+    List.compare Int64.compare [l0; l1] [r0; r1]
```
xclerc: (wouldn't that be a tiny bit better
with an array? Not that it really
matters...)

backend/printcmm.ml
```diff
@@ -287,7 +295,14 @@ let operation d = function
...
   | Cconst_natint (n, _dbg) -> fprintf ppf "%s" (Nativeint.to_string n)
-  | Cconst_vec128 ({ low; high }, _dbg) -> fprintf ppf "%016Lx:%016Lx" high low
+  | Cconst_vec128 ({ word0; word1 }, _dbg) ->
+    fprintf ppf "%016Lx:%016Lx" word0 word1
+  | Cconst_vec256 ({ word0; word1; word2; word3 }, _dbg) ->
+    fprintf ppf "%016Lx:%016Lx:%016Lx:%016Lx" word0 word1 word2 word3
+  | Cconst_vec512
+      ({ word0; word1; word2; word3; word4; word5; word6; word7 }, _dbg) ->
+    fprintf ppf "%016Lx:%016Lx:%016Lx:%016Lx:%016Lx:%016Lx:%016Lx:%016Lx" word0
+      word1 word2 word3 word4 word5 word6 word7
```
xclerc: It feels like the reverse order would
be more natural.

backend/printcmm.ml
```diff
@@ -450,8 +465,16 @@ let data_item ppf = function
...
+  | Cvec128 { word0; word1 } ->
+    fprintf ppf "vec128 %s:%s" (Int64.to_string word0) (Int64.to_string word1)
+  | Cvec256 { word0; word1; word2; word3 } ->
+    fprintf ppf "vec256 %s:%s:%s:%s" (Int64.to_string word0)
+      (Int64.to_string word1) (Int64.to_string word2) (Int64.to_string word3)
+  | Cvec512 { word0; word1; word2; word3; word4; word5; word6; word7 } ->
+    fprintf ppf "vec512 %s:%s:%s:%s:%s:%s:%s:%s" (Int64.to_string word0)
+      (Int64.to_string word1) (Int64.to_string word2) (Int64.to_string word3)
+      (Int64.to_string word4) (Int64.to_string word5) (Int64.to_string word6)
+      (Int64.to_string word7)
```
xclerc: (Same remark.)

backend/printoperation.ml
```diff
@@ -15,7 +15,12 @@ let operation ?(print_reg = Printreg.reg) (op : Operation.t) arg ppf res =
   | Const_float32 f -> fprintf ppf "%Fs" (Int32.float_of_bits f)
   | Const_float f -> fprintf ppf "%F" (Int64.float_of_bits f)
   | Const_symbol s -> fprintf ppf "\"%s\"" s.sym_name
-  | Const_vec128 { high; low } -> fprintf ppf "%016Lx:%016Lx" high low
+  | Const_vec128 { word0; word1 } -> fprintf ppf "%016Lx:%016Lx" word0 word1
+  | Const_vec256 { word0; word1; word2; word3 } ->
+    fprintf ppf "%016Lx:%016Lx:%016Lx:%016Lx" word0 word1 word2 word3
+  | Const_vec512 { word0; word1; word2; word3; word4; word5; word6; word7 } ->
+    fprintf ppf "%016Lx:%016Lx:%016Lx:%016Lx:%016Lx:%016Lx:%016Lx:%016Lx" word0
+      word1 word2 word3 word4 word5 word6 word7
```
xclerc, line 23: (Same remark as in `Printcmm`.)

backend/x86_ast.mli
```diff
@@ -61,7 +61,9 @@ type data_type =
   | NONE
   | REAL4 | REAL8 (* floating point values *)
   | BYTE | WORD | DWORD | QWORD (* integer values *)
-  | VEC128 (* vector values (float & integer) *)
+  | VEC128
```
xclerc: IS the deletion of the comment intentional?

backend/amd64/reg_class.mli
```diff
@@ -1,8 +1,8 @@
 [@@@ocaml.warning "+a-40-41-42"]
 
 type t =
-  | Int64
-  | Float128
+  | GPR
+  | SIMD
```
xclerc: (I wonder whether we should comment
the constructors; maybe leave a note
about the sizes?)

backend/amd64/reg_class.ml
```diff
@@ -67,8 +90,8 @@ module T = struct
     let reg_class = of_machtype typ in
     let index = reg_index - first_available_register reg_class in
     match reg_class with
-    | Int64 -> index
-    | Float128 ->
+    | GPR -> index
+    | SIMD ->
       let slot_size_in_vals = 2 in
```
xclerc: This looks suspicious, or deserves
a comment since `SIMD` is no longer
guaranteed to be 128 bits.)

backend/amd64/reg_class.ml
```diff
@@ -48,16 +47,40 @@ module T = struct
          "%xmm8"; "%xmm9"; "%xmm10"; "%xmm11";
          "%xmm12"; "%xmm13"; "%xmm14"; "%xmm15" |]
 
+  let[@ocamlformat "disable"] ymm_name =
+    match Config.ccomp_type with
+    | "msvc" ->
+      [| "ymm0"; "ymm1"; "ymm2"; "ymm3"; "ymm4"; "ymm5"; "ymm6"; "ymm7";
```
xclerc: (It was already a bit unpleasant
with only `xmm`, and having it
three times makes it worse.
Could we avoid these array
literals? - quite possibly in another
pull request.)

backend/amd64/emit.ml
```diff
@@ -747,18 +788,59 @@ let emit_float_constant f lbl =
...
+    let lbl = L.create Thirtytwo_byte_literals in
+    vec256_constants := (bits, lbl) :: !vec256_constants;
+    lbl
+
+let add_vec512_constant bits =
+  try List.assoc bits !vec512_constants
+  with Not_found ->
+    let lbl = L.create Sixtyfour_byte_literals in
+    vec512_constants := (bits, lbl) :: !vec512_constants;
+    lbl
```
xclerc, line 863: Is it worth factoring out the three
`add_vecXYZ_constant` functions?
(I am not positive it is.)

backend/amd64/emit.ml
```diff
@@ -1442,13 +1561,25 @@ let emit_instr ~first ~fallthrough i =
     | _ ->
       let lbl = add_float_constant f in
       I.movsd (mem64_rip REAL8 (L.encode lbl)) (res i 0))
-  | Lop (Const_vec128 { high; low }) -> (
-    match high, low with
+  | Lop (Const_vec128 { word0; word1 }) -> (
+    match word0, word1 with
```
xclerc: not matter, but it feels likw
`word1; word0` might be
slightly more natural?)

backend/amd64/emit.ml
```diff
@@ -2078,9 +2246,23 @@ let emit_item : Cmm.data_item -> unit = function
   | Csingle f -> D.float32 f
   | Cdouble f -> D.float64 f
   (* SIMD vectors respect little-endian byte order *)
-  | Cvec128 { high; low } ->
-    D.float64_from_bits low;
-    D.float64_from_bits high
+  | Cvec128 { word0; word1 } ->
```
xclerc, line 2282: (Again, I am not sure it is the
natural order for the fields, but
obviously does not matter / can
be ignored.)

xclerc (approved):
(I have read the files under `backend/`,
except the cmm ones.)

flambda-backend/tests/simd/basic256.ml
```diff
@@ -0,0 +1,354 @@
+open Stdlib
```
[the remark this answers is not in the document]
TheNumbat: There's enough features for `probes` and `callback` -- added those.

typing/typedecl.ml
```diff
@@ -1826,8 +1834,8 @@ let rec update_decl_jkind env dpath decl =
            match repr with
            | Float_element -> repr_summary.floats <- true
            | Unboxed_element Float64 -> repr_summary.float64s <- true
-           | Unboxed_element ( Float32 | Bits32 | Bits64 | Vec128 | Word
-                             | Product _ ) ->
```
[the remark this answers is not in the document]
TheNumbat: There is a bit of testing in `basic`, but updated the generated tests to use int64x4

backend/cmmgen_state.ml
TheNumbat: Added everywhere

backend/cmmgen_state.ml
TheNumbat: It would, but `Array.compare` is not provided by the stdlib >_>

backend/printcmm.ml
TheNumbat: Flipped

backend/x86_ast.mli
TheNumbat: no

backend/amd64/reg_class.mli
TheNumbat: Added a comment

backend/amd64/reg_class.ml
TheNumbat: This is wrong, good catch. It needs to know about the live set now, since the offset depends on whether vec256/512s are in use, so I made it specific to amd64 reg_class.

backend/amd64/reg_class.ml
TheNumbat: Fixed

backend/amd64/emit.ml
TheNumbat: I think it's not unless we want to unify the constant lists into a better data structure. Could do that in a future pr.

backend/amd64/emit.ml
TheNumbat: Done

backend/amd64/emit.ml
TheNumbat: Added comments

typing/typedecl.ml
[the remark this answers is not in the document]
TheNumbat: This requires making the test only run on amd64, but I think that's ok?

TheNumbat:
Responded to review - now need to fix the asan job & stack checks + effects

flambda-backend/tests/simd/basic256.ml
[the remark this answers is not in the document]
TheNumbat: Added a comment in predef

backend/amd64/reg_class.ml
TheNumbat: (Also note this is only used for computing the offset of `Valx2` registers arising from autovectorization)

backend/cmmgen_state.ml
xclerc: But it is in `Misc.Stdlib.Array`.

backend/amd64/emit.ml
xclerc: I think it would be worth thinking
about it, it a future pr as you point
out.

backend/cmmgen_state.ml
TheNumbat: ah right, swapped to that

backend/cmmgen_state.ml
```diff
@@ -82,7 +98,9 @@ let rank_structured_constant = function
   | Const_float_array _ -> 5
   | Const_string _ -> 6
   | Const_vec128 _ -> 8
-  | Const_float32 _ -> 9
+  | Const_vec256 _ -> 9
```
xclerc: (I suspect we don't care, but having
a hole in the returned values is very
likely to generate worse code.)

TheNumbat: huh, seems it was already like that. fixed.

xclerc: Thanks - indeed that was not part of the diff.

runtime/caml/simd.h
```diff
@@ -31,45 +32,91 @@
...
+#define Store_vec128_vali(v, x) _mm_storeu_si128((__m128i *)Bp_val(v), x)
 
-#define Vec128_val(v)  _mm_loadu_ps((const float*)Bp_val(v))
-#define Vec128_vald(v) _mm_loadu_pd((const double*)Bp_val(v))
-#define Vec128_vali(v) _mm_loadu_si128((const __m128i*)Bp_val(v))
-#define Store_vec128_val(v,x)  _mm_storeu_ps((float*)Bp_val(v), x)
-#define Store_vec128_vald(v,x) _mm_storeu_pd((double*)Bp_val(v), x)
-#define Store_vec128_vali(v,x) _mm_storeu_si128((__m128i*)Bp_val(v), x)
+Caml_inline value caml_copy_vec128(__m128 v) {
+    value res = caml_alloc_small(2, Abstract_tag);
```
NickBarnes: I would prefer a symbolic constant in place of the "2" here (and in the matching places nearby).

runtime/caml/simd.h
```diff
@@ -31,45 +32,91 @@
...
+
+#define Vec256_val(v) _mm256_loadu_ps((const float *)Bp_val(v))
+#define Vec256_vald(v) _mm256_loadu_pd((const double *)Bp_val(v))
+#define Vec256_vali(v) _mm256_loadu_si256((const __m256i *)Bp_val(v))
+#define Store_vec256_val(v, x) _mm256_storeu_ps((float *)Bp_val(v), x)
+#define Store_vec256_vald(v, x) _mm256_storeu_pd((double *)Bp_val(v), x)
+#define Store_vec256_vali(v, x) _mm256_storeu_si256((__m256i *)Bp_val(v), x)
+
+Caml_inline value caml_copy_vec256(__m256 v) {
+    value res = caml_alloc_small(4, Abstract_tag);
```
NickBarnes: Ditto ditto 4

runtime/caml/simd.h
```diff
@@ -31,45 +32,91 @@
...
+#define Store_vec128_vald(v, x) _mm_storeu_pd((double *)Bp_val(v), x)
+#define Store_vec128_vali(v, x) _mm_storeu_si128((__m128i *)Bp_val(v), x)
 
-#define Vec128_val(v)  _mm_loadu_ps((const float*)Bp_val(v))
-#define Vec128_vald(v) _mm_loadu_pd((const double*)Bp_val(v))
-#define Vec128_vali(v) _mm_loadu_si128((const __m128i*)Bp_val(v))
-#define Store_vec128_val(v,x)  _mm_storeu_ps((float*)Bp_val(v), x)
-#define Store_vec128_vald(v,x) _mm_storeu_pd((double*)Bp_val(v), x)
-#define Store_vec128_vali(v,x) _mm_storeu_si128((__m128i*)Bp_val(v), x)
+Caml_inline value caml_copy_vec128(__m128 v) {
```
NickBarnes, line 46: These functions aren't apparently used anywhere in the compiler tree. Are they utilities for third-party code? Could do with some comments anyway.

runtime/simd.c
```diff
@@ -25,29 +25,12 @@
 
 // SIMD is only supported on 64-bit targets
 #define Max_unboxed_vec128_array_wosize    (Max_custom_array_wosize / 2)
+#define Max_unboxed_vec256_array_wosize    (Max_custom_array_wosize / 4)
```
NickBarnes: I'm curious why there is no vec512 code in the runtime, given that there is in backend.

runtime/simd.c
```diff
@@ -25,29 +25,12 @@
...
-}
-
-CAMLexport value caml_copy_vec128d(simd_float64x2_t v) {
-    value res = caml_alloc_small(2, Abstract_tag);
-    Store_vec128_vald(res, v);
-    return res;
-}
-
-/* Defined in array.c */
+#define Words_per_vec128 2
```
NickBarnes: Move to simd.h?

runtime/simd.c
```diff
@@ -105,26 +88,60 @@ CAMLprim value caml_make_local_unboxed_vec128_vect(value len)
...
+            ((uintnat *)a1 + 1) + Long_val(ofs1) * Words_per_vec256,
+            Long_val(n) * sizeof(uintnat) * Words_per_vec256);
+    return Val_unit;
 }
 
-CAMLprim value caml_make_unboxed_vec128_vect(value len) {
-  caml_failwith("SIMD is not supported on this platform.");
+static value caml_make_unboxed_vec256_vect0(value len, int local)
+{
+  /* This is only used on 64-bit targets. */
```
NickBarnes: Why is this important?

runtime4/startup_nat.c
```diff
@@ -48,6 +48,9 @@ extern int caml_parser_trace;
 extern char caml_system__code_begin, caml_system__code_end;
 extern uintnat caml_prelinking_in_use;
 
+/* Implemented in isa.c. */
```
NickBarnes: It is? At least in the compiler tree, this is neither defined nor called.

runtime/amd64.S
```diff
@@ -493,9 +495,149 @@ G(caml_system__code_begin):
 /* Allocation */
 /******************************************************************************/
 
+#define SAVE_XMM_REGS                         \
+        movupd  %xmm0,  (0*16 + 13*8)(%r15);  \
```
NickBarnes: Would prefer a symbolic identifier (possibly computed?) for `13*8` throughout this file.

runtime/amd64.S
```diff
@@ -694,63 +839,127 @@ CFI_STARTPROC
...
+CFI_ENDPROC
+ENDFUNCTION(G(caml_alloc3_avx))
+
+FUNCTION(G(caml_alloc3_avx512))
+CFI_STARTPROC
+Make_alloc(32,_avx512)
+CFI_ENDPROC
+ENDFUNCTION(G(caml_alloc3_avx512))
+
+FUNCTION(G(caml_allocN))
```
NickBarnes: It feels like these could be macro-ized too.

runtime/amd64.S
```diff
@@ -694,63 +839,127 @@ CFI_STARTPROC
...
-        subq    $16, %r15
-        cmpq    Caml_state(young_limit), %r15
-        jb      LBL(caml_call_gc)
-        LEAVE_FUNCTION
-        ret
+Make_alloc(16,)
 CFI_ENDPROC
 ENDFUNCTION(G(caml_alloc1))
 
+FUNCTION(G(caml_alloc1_avx))
```
NickBarnes: Another layer of macrology here, to unify the 1/2/3 cases? Or would that just obfuscate?

NickBarnes: (and yet more macrology here, for `/_avx/_avx512`? Looks doable but does it make it all too opaque?)

runtime/caml/simd.h
TheNumbat: Moved constants from `simd.c`

runtime/caml/simd.h
TheNumbat: Yeah, these are intended for user C stubs that include `caml/simd.h`. I realized they needed to be defined inline in `simd.h` rather than `simd.c` because we don't compile the runtime with avx support (but C stubs that use avx will enable it). Updated the comment at the top

runtime/simd.c
TheNumbat: No particular reason, just that 512-bit vectors are not actually supported yet in the backend (and restricted to the simd_alpha extension)

runtime/simd.c
TheNumbat: Strictly speaking it's not, it's only important in that Words_per_vec128/256 is a fixed 2/4, which is used to compute the max array size. Deleted the comment here.

runtime4/startup_nat.c
TheNumbat: Sorry, forgot to delete from an earlier draft

runtime/amd64.S
TheNumbat: Yeah, added your macro changes

runtime/amd64.S
TheNumbat: lgtm

runtime/amd64.S
TheNumbat: Added

NickBarnes (approved):
The runtime stuff looks good to me now.

ccasin (approved):

PR #3967, opened 2025-05-05
Propagate Debugging Identifiers through the Middle End 

spiessimon:
This PR builds on top of #3942. It propagates the debugging identifiers from Lambda through Flambda.


middle_end/flambda2/types/env/join_env.ml
```diff
@@ -1192,7 +1192,12 @@ let cut_and_n_way_join ~n_way_join_type ~meet_type ~cut_after target_env
       Variable.Map.fold
         (fun var kind target_env ->
           TE.add_definition target_env
-            (Bound_name.create_var (Bound_var.create var Name_mode.in_types))
+            (Bound_name.create_var
+               (Bound_var.create var Flambda_debug_uid.none
+                  (* CR sspies: We can probably find a better
+                     [Flambda_debug_uid.t] for these extra variables. However,
+                     propagating it here seems non-trivial. *)
```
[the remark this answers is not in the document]
spiessimon: Ah, that's very useful to know! I've updated the comment here and in a few other places to indicate that `Flambda_debug_uid.none` is correct here.

middle_end/flambda2/simplify/unboxing/unboxing_types.mli
```diff
@@ -31,10 +31,14 @@ type do_not_unbox_reason =
 module Extra_param_and_args : sig
   type t = private
     { param : Variable.t;
+      param_debug_uid : Flambda_debug_uid.t;
+      (* CR sspies: This is [Flambda_debug_uid.none] at all the creation points
+         that I found. I still added the field, because that is non-obvious at
+         the usage points of the these extra params. *)
```
Gbury: this comment seems to miss a few words ?

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
```diff
@@ -305,11 +305,15 @@ let extra_params_for_continuation_param_aliases cont uacc rewrite_ids =
           (fun _id -> EPA.Extra_arg.Already_in_scope (Simple.var var))
           rewrite_ids
       in
+      let var_duid = Flambda_debug_uid.none in
+      (* CR sspies: [extra_params] sounds generated and not user visible. If
+         this is wrong, we should bother to propagate the right
+         [Flambda_debug_uid.t] values here. *)
```
Gbury: Actually, extra params can sometimes be generated and not user visible (e.g. create during unboxing), and sometimes can be user-visible, or at least aliases of user-visible variable (e.g. CSE, loop invariants, etc...).

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
```diff
@@ -555,8 +559,13 @@ let add_lets_around_handler cont at_unit_toplevel uacc handler =
   let handler, uacc =
     Variable.Map.fold
       (fun var bound_to (handler, uacc) ->
+        let var_duid = Flambda_debug_uid.none in
+        (* CR sspies: This seems very generic. Are these ever user visible? If
+           so, we should propagate their [Flambda_debug_uid.t] values to this
+           point. *)
```
Gbury: as stated in another of my comments, some of these may be aliases of user-visible variables, so I'd guess we should try and propagate their debug uids (that will likely require updating the flow analysis).

middle_end/flambda2/simplify/lifted_cont_params.ml
```diff
@@ -35,10 +35,11 @@ let new_param t ~replay_history bound_param =
...
     | Replayed variable_mapping ->
-      let original_var =
-        Variable.Map.find (BP.var bound_param) variable_mapping
-      in
-      BP.create original_var (BP.kind bound_param)
+      let param_var, param_duid = BP.var_and_uid bound_param in
+      let original_var = Variable.Map.find param_var variable_mapping in
+      BP.create original_var (BP.kind bound_param) param_duid
+    (* CR sspies: Is it correct that we copy the [Flambda_debug_uid.t] here?
+       This requires a closer look. *)
```
Gbury: In this case, i'd say it doesn't matter: we use Bound_parameters as key for a Map, so all we care about is that they are compared equal to their original value, but otherwise the keys of the map are never returned, so their debug uid will not leak/appear anywhere I'd say.

middle_end/flambda2/simplify/flow/mutable_unboxing.ml
```diff
@@ -476,7 +476,11 @@ module Fold_prims = struct
               (fun i kind ->
                 let name = Variable.unique_name block_needed in
                 let var = Variable.create (Printf.sprintf "%s_%i" name i) in
-                Bound_parameter.create var kind)
+                let var_duid = Flambda_debug_uid.none in
+                (* CR sspies: While this is an internally generated variable,
+                   would it make sense to try to propagate a
+                   [Flambda_debug_uid.t] here? *)
```
Gbury: these variables correspond to the fiels of the block being unboxed (for instance the contents of a ref), I'm not sure we have a debug uid available here (and in the case where this variable ends up as an alias of another variable, i'd say this is not here that we handle to handle it but more generally correctly handle aliases and debug uids).

middle_end/flambda2/to_cmm/to_cmm.ml
```diff
@@ -91,7 +93,10 @@ let unit0 ~offsets ~all_code ~reachable_names flambda_unit =
   (* See comment in [To_cmm_set_of_closures] about binding [my_region] *)
   let env, toplevel_region_var =
     Env.create_bound_parameter env
-      (Flambda_unit.toplevel_my_region flambda_unit)
+      ( Flambda_unit.toplevel_my_region flambda_unit,
+        Flambda_debug_uid.none
+        (* CR sspies: Do we have a better [Flambda_debug_uid.t] available
+           here? *) )
```
Gbury: region variables are not user visible (and don't appear in the source code), so I'm not sure if there can be any debug uid available ?

middle_end/flambda2/to_cmm/to_cmm_set_of_closures.ml
```diff
@@ -448,8 +448,13 @@ let params_and_body0 env res code_id ~result_arity ~fun_dbg
     if not is_my_closure_used
     then params
     else
+      let my_closure_duid = Flambda_debug_uid.none in
+      (* CR sspies: Not sure whethere these closures can ever be user visible.
+         Popagating a [Lambda_debug_uid.t] here is nontrivial, so I picked
+         [Flambda_debug_uid.none] for now. *)
```
Gbury: The closures should indeed not be user visible.

mshinwell:
Hold on, I'm not sure all uses of variables should be carrying the debug uid.  Will have to think about this further.

spiessimon:
When we emit the final DWARF information, the debugging uids are needed for the binders but not for the uses of the variable. Hence, I would be inclined to keep them with the binders. It might be that for some intermediate transformations, it is better if we also track them with the variables. However, switching from binding occurrences to uses is a somewhat larger change. It probably makes sense to only do that if we notice we are losing important debugging uids along the way.

middle_end/flambda2/simplify/unboxing/unboxing_types.mli
spiessimon: Yeah, this was confusing. I've clarified it.

middle_end/flambda2/simplify/flow/mutable_unboxing.ml
spiessimon: I've removed the comment for now. Taking a field out of a block should be a type changing operation, so I don't think we would even have type information available typically for this case (and it would need a separate uid).

middle_end/flambda2/simplify/lifted_cont_params.ml
spiessimon: Thanks for clarifying!

middle_end/flambda2/to_cmm/to_cmm.ml
spiessimon: Great! I've removed the comment.

middle_end/flambda2/to_cmm/to_cmm_set_of_closures.ml
spiessimon: Great! I've removed the comment.

middle_end/flambda2/identifiers/flambda_debug_uid.mli
```diff
@@ -0,0 +1,33 @@
...
+(*   special exception on linking described in the file LICENSE.          *)
+(*                                                                        *)
+(**************************************************************************)
+
+(** Augmented version of [Lambda.debug_uid] that can track variables forming parts
+    of unboxed products. *)
+
+type t = private
+  | Uid of Lambda.debug_uid
+  | Proj of Lambda.debug_uid * int
```
mshinwell: This should probably use an inline record, so the integer can be labelled.

middle_end/flambda2/identifiers/flambda_debug_uid.ml
```diff
@@ -0,0 +1,69 @@
...
+(*                                                                        *)
+(*   All rights reserved.  This file is distributed under the terms of    *)
+(*   the GNU Lesser General Public License version 2.1, with the          *)
+(*   special exception on linking described in the file LICENSE.          *)
+(*                                                                        *)
+(**************************************************************************)
+
+module Uid = Shape.Uid
+(* Lambda debugging uids of type [Lambda.debug_uid] are actually values of type
+   Shape.Uid.t, so we use the functions from [Shape.Uid] below. *)
```
mshinwell, line 17: It's best not to be sharing types across IRs like this: for one thing, in the future they invariably need changing for particular IRs.  In a separate PR maybe we could add Lambda.Debug_uid as a module, together with e.g. the comparison operation, and a creation function that takes the type checker's Uid.t.

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
spiessimon: I've updated the comment in the code for this one. I think improving the propagation would be great in a subsequent step.

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
spiessimon: Same as above.

middle_end/flambda2/bound_identifiers/bound_var.ml
```diff
@@ -16,19 +16,22 @@
...
 type t =
   { var : Variable.t;
+    debug_uid : Flambda_debug_uid.t;
     name_mode : Name_mode.t
   }
 
-let [@ocamlformat "disable"] print ppf { var; name_mode = _; } =
-  Variable.print ppf var
+let [@ocamlformat "disable"] print ppf { var; debug_uid; name_mode = _; } =
+  Format.fprintf ppf "%a,uid=%a" Variable.print var Flambda_debug_uid.print debug_uid
```
mshinwell: Printing the duids should probably be guarded by a flag

middle_end/flambda2/bound_identifiers/bound_parameter.ml
```diff
@@ -18,34 +18,45 @@ module Simple = Int_ids.Simple
...
+  let hash { param; kind; uid } =
+    Hashtbl.hash
+      ( Variable.hash param,
+        Flambda_kind.With_subkind.hash kind,
+        Flambda_debug_uid.hash uid )
 
-  let [@ocamlformat "disable"] print ppf { param; kind; } =
-    Format.fprintf ppf "@[(%t%a%t @<1>\u{2237} %a)@]"
+  let [@ocamlformat "disable"] print ppf { param; kind; uid } =
+    Format.fprintf ppf "@[(%t%a,uid=%a%t @<1>\u{2237} %a)@]"
```
mshinwell: Same comment as elsewhere about guarding the printing of duids

middle_end/flambda2/from_lambda/closure_conversion.ml
```diff
@@ -1663,7 +1694,10 @@ let close_switch acc env ~condition_dbg scrutinee (sw : IR.switch) :
     Expr_with_acc.t =
   let scrutinee = find_simple_from_id env scrutinee in
   let untagged_scrutinee = Variable.create "untagged" in
-  let untagged_scrutinee' = VB.create untagged_scrutinee Name_mode.normal in
+  let untagged_scrutinee_duid = Flambda_debug_uid.none in
+  let untagged_scrutinee' =
+    VB.create untagged_scrutinee untagged_scrutinee_duid Name_mode.normal
+  in
```
mshinwell, line 1727: This is interesting.  We should probably have something like a phantom let here, even though the target variable isn't user-visible.  Then in the debugger, if only the untagged scrutinee were available (and not the original variable), we could still show a value for the original variable.  Moreover I wonder if we should do this systematically in To_cmm for tagging/untagging/boxing/unboxing operations.

mshinwell: This seems related to tracking of relational information in the Flambda type system.  It could even maybe be extended to things like conditionals: if a register contains the result of a conditional, we could show that in the debugger together with what was being compared...

middle_end/flambda2/from_lambda/closure_conversion.ml
```diff
@@ -1777,20 +1814,28 @@ let close_switch acc env ~condition_dbg scrutinee (sw : IR.switch) :
 
 let variables_for_unboxing boxed_variable_name (k : Function_decl.unboxing_kind)
     =
+  (* CR sspies: This function currently picks [.none] for all the debugging uids
```
mshinwell: We can deal with this later; I think this code is used for the unboxed attributes (allowing calling convention overhead to be eliminated without changing layouts in the source code), but that's not being used yet as iirc there were some outstanding problems related to locals.

middle_end/flambda2/from_lambda/closure_conversion.ml
```diff
@@ -1869,19 +1915,26 @@ let compute_body_of_unboxed_function acc my_region my_closure
...
               (Alloc_mode.For_types.to_lambda param_mode)
           in
+          let _, param_duid = BP.var_and_uid param in
           Let_with_acc.create acc
             (Bound_pattern.singleton
-               (Bound_var.create (Bound_parameter.var param) Name_mode.normal))
+               (Bound_var.create
+                  (Bound_parameter.var param)
+                  param_duid Name_mode.normal))
+            (* CR sspies: Is this debugging uid correct? *)
```
mshinwell: For the moment maybe just set it to "none", and leave a CR

middle_end/flambda2/from_lambda/closure_conversion.ml
```diff
@@ -1957,13 +2014,15 @@ let compute_body_of_unboxed_function acc my_region my_closure
...
+          (List.map (fun (_, _, kind) -> kind) vars_with_kinds),
         unboxed_return_continuation )
   in
   let my_unboxed_closure = Variable.create "my_unboxed_closure" in
   let acc, unboxed_body =
     Let_with_acc.create acc
-      (Bound_pattern.singleton (Bound_var.create my_closure Name_mode.normal))
+      (Bound_pattern.singleton
+         (Bound_var.create my_closure my_closure_duid Name_mode.normal))
+      (* CR sspies: Should this be [my_unboxed_closure]? *)
```
mshinwell: I'm unsure what this CR is referring to

middle_end/flambda2/from_lambda/closure_conversion.ml
```diff
@@ -2899,8 +2970,11 @@ let close_let_rec acc env ~function_declarations
     List.fold_left
       (fun (fun_vars_map, ident_map) decl ->
         let ident = Function_decl.let_rec_ident decl in
+        let ident_duid = Function_decl.let_rec_debug_uid decl in
         let fun_var =
-          VB.create (fst (Env.find_var env ident)) Name_mode.normal
+          VB.create (fst (Env.find_var env ident)) ident_duid Name_mode.normal
+          (* CR sspies: Does it make sense here to take the debugging uid of the
+             function? *)
```
mshinwell: Yeah this seems like it should be correct

middle_end/flambda2/from_lambda/closure_conversion_aux.mli
```diff
@@ -14,6 +14,8 @@
 (*                                                                        *)
 (**************************************************************************)
 
+module Uid = Shape.Uid
```
mshinwell: This seems to be unused?

middle_end/flambda2/from_lambda/lambda_to_flambda.ml
```diff
@@ -360,7 +374,14 @@ let wrap_return_continuation acc env ccenv (apply : IR.apply) =
       let params =
         List.map2
           (fun return_value_component kind ->
-            return_value_component, IR.Not_user_visible, kind)
+            ( return_value_component,
+              Flambda_debug_uid.none,
+              (* CR sspies: There are some debug uids for the extra arguments
+                 that we could propagate here. But it seems these are never user
+                 visible, so there is currently no point in propagate the debug
+                 uids. (They are probably [.none] anyways, I guess.) *)
```
mshinwell: I'm not sure it's correct that these are never user visible.  I think they could have arisen from mutable variables in the Lambda code or maybe other transformations such as @gbury mentioned

middle_end/flambda2/from_lambda/lambda_to_flambda.ml
```diff
@@ -432,8 +454,11 @@ let name_if_not_var acc ccenv name simple kind body =
   | IR.Var id -> body id acc ccenv
   | IR.Const _ ->
     let id = Ident.create_local name in
+    let id_duid = Flambda_debug_uid.none in
+    (* CR sspies: I think the [name] is always a constant string (so these are
+       not user visible variables). As such, using [.none] should be fine. *)
```
mshinwell: Yes this is correct

middle_end/flambda2/from_lambda/lambda_to_flambda.ml
```diff
@@ -488,9 +511,13 @@ let rec cps acc env ccenv (lam : L.lambda) (k : cps_continuation)
           ap_loc ap_inlined ap_probe ap_result_layout k k_exn)
   | Lfunction func ->
     let id = Ident.create_local (name_for_function func) in
+    let id_duid = Flambda_debug_uid.none in
+    (* CR sspies: Is there a better [Flambda_debug_uid.t] available here? *)
```
mshinwell: Hmm we should discuss, but we maybe need a predefined uid for "arbitrary closure".  It would be nice to propagate the uid properly though - maybe it could be added to Lambda.lfunction somehow?

middle_end/flambda2/from_lambda/lambda_to_flambda.ml
```diff
@@ -653,16 +686,15 @@ let rec cps acc env ccenv (lam : L.lambda) (k : cps_continuation)
   | Llet
       ((Strict | Alias | StrictOpt), _layout, id, _duid, defining_expr, Lvar id')
     when Ident.same id id' ->
-    (* CR sspies: dropping [debug_uid]; address in subsequent PR. *)
+    (* CR sspies: Is there any use for the debugging uid in this case? *)
```
mshinwell: This is a case where a phantom let should be inserted; that could carry the duid.

middle_end/flambda2/from_lambda/lambda_to_flambda.ml
```diff
@@ -854,12 +892,14 @@ let rec cps acc env ccenv (lam : L.lambda) (k : cps_continuation)
...
           k_exn)
       k_exn
-  | Ltrywith (body, id, _duid, handler, kind) ->
-    (* CR sspies: dropping [debug_uid]; address in subsequent PR. *)
+  | Ltrywith (body, id, duid, handler, kind) ->
     let dbg = Debuginfo.none (* CR mshinwell: fix [Lambda] *) in
     let body_result = Ident.create_local "body_result" in
+    let body_result_lambda_duid = L.debug_uid_none in
     let region = Ident.create_local "try_region" in
+    let region_duid = Flambda_debug_uid.none in
```
mshinwell, line 899: It would be kind of fun to be able to see in the debugger that register %rbx or whatever contains the address of the local stack pointer at foo.ml:42 in the code...

middle_end/flambda2/reaper/rebuild.ml
```diff
@@ -47,6 +47,10 @@ type param_decision =
   | Delete
   | Unbox of Variable.t DS.unboxed_fields Field.Map.t
 
+(* CR sspies: Throughout this file, we create bound paramters and variables
+   without corresponding debugging uids. Does it make sense to properly
+   propagate debugging uids there? If so, where should they come from? *)
```
mshinwell, line 52: As discussed (re. wiki or CR labelling)

middle_end/flambda2/simplify/env/downwards_env.ml
```diff
@@ -343,7 +347,10 @@ let define_name t name kind =
   Name.pattern_match (Bound_name.name name)
     ~var:(fun [@inline] var ->
       (define_variable [@inlined hint]) t
-        (Bound_var.create var (Bound_name.name_mode name))
+        (Bound_var.create var Flambda_debug_uid.none
+           (* CR sspies: Unclear whether bound names should have a
+              [Flambda_debug_uid.t]. For now, I just left it as [.none]. *)
+           (Bound_name.name_mode name))
```
mshinwell, line 353: I think this will need fixing, but ok in a subsequent PR

middle_end/flambda2/simplify/flow/mutable_unboxing.ml
```diff
@@ -484,7 +484,8 @@ module Fold_prims = struct
                         "[Mutable Unboxing] Cannot unbox constants")
                 in
                 let var = Variable.create (Printf.sprintf "%s_%i" name i) in
-                Bound_parameter.create var kind)
+                let var_duid = Flambda_debug_uid.none in
```
mshinwell, line 487: This could probably be improved in the future.  Either leave a CR or we can just grep for Flambda_debug_uid.none

middle_end/backend_var.ml
```diff
@@ -35,29 +35,33 @@ module Provenance = struct
...
     let printf fmt = Format.fprintf ppf fmt in
     printf "@[<hov 1>(";
     printf "@[<hov 1>(module_path@ %a)@]@ "
       Path.print module_path;
     if !Clflags.locations then
       printf "@[<hov 1>(location@ %a)@]@ "
         Debuginfo.print_compact location;
-    printf "@[<hov 1>(original_ident@ %a)@]"
-      Ident.print original_ident;
+    printf "@[<hov 1>(original_ident@ %a,uid=%a)@]"
```
mshinwell: "duid"?  (and maybe in other printing functions)

middle_end/flambda2/simplify/inlining/inlining_transforms.ml
```diff
@@ -44,8 +44,13 @@ let make_inlined_body ~callee ~called_code_id ~unroll_to ~params ~args
         in
         Some callee, unrolled_rec_info)
   in
+  let my_closure_duid = Flambda_debug_uid.none in
+  (* CR sspies: Not sure whethere these closures can ever be user visible.
+     Popagating a [Lambda_debug_uid.t] here is nontrivial, so I picked
+     [Flambda_debug_uid.none] for now. *)
```
mshinwell: We could maybe have a distinguished "closure" uid as mentioned elsewhere.  OTOH now I think about that, maybe it isn't that useful, as we can always distinguish these at runtime by observing Closure_tag / Infix_tag.

middle_end/flambda2/simplify/expr_builder.ml
```diff
@@ -264,6 +264,9 @@ let create_coerced_singleton_let uacc var defining_expr
         let name = "uncoerced_" ^ Variable.unique_name (VB.var var) in
         Variable.create name
       in
+      let uncoerced_var_duid = Flambda_debug_uid.none in
+      (* CR sspies: Would it make sense here to propagate the
+         [Flambda_debug_uid.t] of the variable [var]? *)
```
mshinwell: I think so, but needs checking - for later.

middle_end/flambda2/simplify/expr_builder.ml
```diff
@@ -569,7 +573,9 @@ let create_let_symbols uacc lifted_constant ~body =
       let free_names_of_defining_expr = Named.free_names defining_expr in
       let expr, uacc, _ =
         create_coerced_singleton_let uacc
-          (VB.create var Name_mode.normal)
+          (VB.create var Flambda_debug_uid.none Name_mode.normal)
+          (* CR sspies: I have no idea whether this could ever be a user visible
+             variable. *)
```
mshinwell: I think the answer is no

middle_end/flambda2/simplify/lifted_cont_params.ml
```diff
@@ -35,10 +35,12 @@ let new_param t ~replay_history bound_param =
...
     | Still_recording -> bound_param
     | Replayed variable_mapping ->
-      let original_var =
-        Variable.Map.find (BP.var bound_param) variable_mapping
-      in
-      BP.create original_var (BP.kind bound_param)
+      let param_var, param_duid = BP.var_and_uid bound_param in
+      let original_var = Variable.Map.find param_var variable_mapping in
+      BP.create original_var (BP.kind bound_param) param_duid
+    (* It is fine to copy the debugging UID here, because the bound parameters
```
mshinwell: This CR can be removed now

middle_end/flambda2/simplify/simplify_apply_expr.ml
```diff
@@ -124,6 +124,9 @@ let simplify_direct_tuple_application ~simplify_expr dacc apply
   let vars_and_fields =
     List.init tuple_size (fun field ->
         ( Variable.create "tuple_field",
+          Flambda_debug_uid.none,
+          (* This internally created varibale does not get a
```
mshinwell: Typo

mshinwell: It should probably have a phantom let though

middle_end/flambda2/simplify/simplify_apply_expr.ml
```diff
@@ -318,8 +321,9 @@ let simplify_direct_full_application ~simplify_expr dacc apply function_type
               let denv =
                 List.fold_left2
                   (fun denv kind result ->
+                    let result_var, result_uid = BP.var_and_uid result in
                     DE.add_variable denv
-                      (VB.create (BP.var result) NM.in_types)
+                      (VB.create result_var result_uid NM.in_types)
```
mshinwell, line 328: Since this is in-types mode, the uid could be dropped, but maybe better to propagate since it's right here

middle_end/flambda2/simplify/simplify_apply_expr.ml
```diff
@@ -604,7 +610,13 @@ let simplify_direct_partial_application ~simplify_expr dacc apply
...
               | Const _ | Symbol _ -> expr, cost_metrics, free_names
               | In_closure { var; value_slot; value = _ } ->
-                let arg = VB.create var Name_mode.normal in
+                let arg =
+                  VB.create var Flambda_debug_uid.none
+                    (* CR sspies: I'm unsure whether these are user visible
+                       variables. At least at one call site, this is an
+                       internally defined variable. *)
+                    Name_mode.normal
+                in
```
mshinwell, line 620: Let's look at this later

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
mshinwell: As discussed, let's copy this text into the CR, but we'll deal with the duid propagation later.

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
mshinwell: For later though (will have to check where this is used)

mshinwell (changes requested):
First pass of review complete

middle_end/flambda2/from_lambda/lambda_to_flambda.ml
mshinwell: (also see comment below about this)

middle_end/flambda2/from_lambda/closure_conversion.ml
spiessimon: I was wondering whether this is a typo in the original code and it should be `my_unboxed_closure` instead of `my_closure`, since the variable is created right above.

middle_end/flambda2/identifiers/flambda_debug_uid.mli
spiessimon: Turned it into an inline record.

middle_end/flambda2/identifiers/flambda_debug_uid.ml
spiessimon: I left a comment to address this in the future.

middle_end/flambda2/bound_identifiers/bound_var.ml
spiessimon: I've disabled the printing for now. I had already made it conditional on the `-ddebug-uids` flag in PR #4255. We can deal with the printing of the UIDs in that one.

middle_end/flambda2/bound_identifiers/bound_parameter.ml
spiessimon: Same as above. Removed the printing for now.

middle_end/flambda2/from_lambda/closure_conversion.ml
spiessimon: I've left a comment to address this in the future.

middle_end/flambda2/from_lambda/closure_conversion.ml
spiessimon: Removed the comment.

middle_end/flambda2/from_lambda/closure_conversion_aux.mli
spiessimon: Indeed. I've removed it.

middle_end/flambda2/from_lambda/lambda_to_flambda.ml
spiessimon: Removed the comment.

middle_end/flambda2/from_lambda/closure_conversion.ml
spiessimon: Done.

middle_end/flambda2/from_lambda/lambda_to_flambda.ml
spiessimon: Is that what the region tracks here? I wasn't quite sure where to add a CR.

middle_end/flambda2/from_lambda/lambda_to_flambda.ml
spiessimon: To map this UID always to a generic "closure" DIE? We do capture the functions arising from `let f x = ...` currently as a binding for `f`.

middle_end/flambda2/from_lambda/lambda_to_flambda.ml
spiessimon: Removed the comment.

middle_end/flambda2/simplify/inlining/inlining_transforms.ml
spiessimon: Yes, I think that's true. It might still make sense to have a distinguished debugging UID, because it tells us what the layout of the variable is (`value`). That's quite helpful, because without layout information, it makes little sense to emit a dwarf die.

middle_end/backend_var.ml
spiessimon: Removed the printing here as for the other cases for now.

middle_end/flambda2/simplify/simplify_apply_expr.ml
spiessimon: Fixed the typo and left a comment.

middle_end/flambda2/simplify/simplify_apply_expr.ml
spiessimon: I left this as is, since it is quite easy to propagate here.

middle_end/flambda2/simplify/expr_builder.ml
spiessimon: Removed the comment.

middle_end/flambda2/simplify/expr_builder.ml
spiessimon: Turned it into a CR to look at in the future.

middle_end/flambda2/simplify/flow/mutable_unboxing.ml
spiessimon: I left a CR to look at this again.

middle_end/flambda2/from_lambda/lambda_to_flambda.ml
spiessimon: I left a CR here that we should add a phantom let.

middle_end/flambda2/simplify/env/downwards_env.ml
spiessimon: I left a CR to look at this in the future.

middle_end/flambda2/simplify/lifted_cont_params.ml
spiessimon: Removed.

middle_end/flambda2/simplify/simplify_apply_expr.ml
spiessimon: Left a CR to look at this in the future.

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
spiessimon: I've copied the text into the CR and left a reference to this PR.

middle_end/flambda2/simplify/simplify_let_cont_expr.ml
spiessimon: I've updated the comment and left a pointer to this PR.

middle_end/flambda2/from_lambda/closure_conversion.ml
spiessimon: I've left a CR with part of the text in your comment and a pointer to this PR.

middle_end/flambda2/from_lambda/lambda_to_flambda.ml
spiessimon: Changed the comment to say we should try to improve the debug UIDs here in the future.

middle_end/flambda2/from_lambda/lambda_to_flambda.ml
mshinwell: Yeah but in fact we can just use the runtime block tags to identify closures, so I'm not sure there is anything to do here.  If we have a proper uid though that's good, as it means we could show the function type in the debugger.

middle_end/flambda2/reaper/rebuild.ml
mshinwell: We decided just to grep for `Flambda_debug_uid.none` when we want to improve things

middle_end/flambda2/from_lambda/lambda_to_flambda.ml
mshinwell: I'm not sure we need a CR - but yes, the region variables hold values of the local stack pointer.

middle_end/flambda2/from_lambda/closure_conversion.ml
mshinwell: Think the original code is ok

spiessimon: I've removed the CR.

PR #3968, opened 2025-05-06
Fix "regalloc" profile category

xclerc:
The "regalloc" category (profiling) was
incorrectly set, and would account for
other CFG operations. This pull request
ensures that only the following is part
of the "regalloc" category:

- validator snapshot;
- register allocation;
- computation of available registers;
- allocation validation.

@gretay-js I am tempted to move
the computation of available registers
to after validation, and hence not to
have it part of the "regalloc" category.
Does that sound right?


gretay-js (approved):
> @gretay-js I am tempted to move
the computation of available registers
to after validation, and hence not to
have it part of the "regalloc" category.
Does that sound right?

yes, swapping the order of `available_reg` and `regalloc_validate` passes sound good to me.

PR #3972, opened 2025-05-06
Fix incorrect sort in `type_argument`

ccasin:
This fixes the sort of a let binding constructed in the case where `type_argument` is coercing a function with missing optional arguments to meet a type without those arguments. As a result, an flambda2 assertion failure is tripped.

The case is pretty hard to hit: I believe you have to be doing the above, the function's next argument has to be a non-value, and you also must avoid an optimization in `simplif` that removes the relevant let binding when the function being used is a variable.

I left a comment that hopefully explains what was wrong in more detail. There is also a test case.  I couldn't easily do a commit with just the failing test, because it doesn't fail on bytecode, but you can try running the new test through the compiler before this patch to witness the flambda2 failure.


goldfirere:
Oops, sorry, didn't see review was assigned to @rtjoa. (Yes, I remember when @ccasin advised me to do better about this.)

PR #3973, opened 2025-05-06
`non_float` arrays are `addrarray`s

dkalinichenko-js:
Make arrays of `non_float` values always at least `addrarray`s.

This seems like a both valuable and a relatively safe change. This reuses an existing optimization we have for records, variants and other non-float types, but keeps it functioning even when the element type is abstract and we only have a kind annotation.

### DIFF
--- a/testsuite/tests/typing-layouts-or-null/non_float_array.ml
+++ b/testsuite/tests/typing-layouts-or-null/non_float_array.ml
@@ -1,56 +1,60 @@
 (* TEST
- flags = "-dlambda";
+ flags = "-dlambda -dno-unique-ids";
  expect;
 *)
 
-(* Normal arrays are [genarray]s. *)
+(* Normal arrays are [genarray]s. Due to the float array optimization,
+   they must check whether their elements are float or non-float on
+   creation, [get] and [set].
+
+   We can see what kind of array we are getting by looking at Lambda. *)
 
 let mk_gen (x : 'a) = [| x |]
 [%%expect{|
-(let
-  (mk_gen/283 =
-     (function {nlocal = 0} x/285 : genarray (makearray[gen] x/285)))
-  (apply (field_imm 1 (global Toploop!)) "mk_gen" mk_gen/283))
+(let (mk_gen = (function {nlocal = 0} x : genarray (makearray[gen] x)))
+  (apply (field_imm 1 (global Toploop!)) "mk_gen" mk_gen))
 val mk_gen : 'a -> 'a array = <fun>
 |}]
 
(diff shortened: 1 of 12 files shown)


testsuite/tests/typing-layouts-or-null/non_float_array.ml
goldfirere, line 1: I really like looking at `-dlambda` here. I'm a little worried, though, about churn; a lot of those unique identifiers will change for uninteresting reasons. Is there a better way? I'd ask a middle-end person, or maybe we can just request a review from one.

typing/typeopt.ml
```diff
@@ -243,8 +243,16 @@ let array_kind_of_elt ~elt_sort env loc ty =
     else
       Pgcscannableproductarray (scannable_product_array_kind loc sorts)
   in
+  (* CR dkalinichenko: layout, externality and separability should be
```
goldfirere, line 246: This isn't enough to detect `float array`, I think. But I like the general idea here.

goldfirere:
I've read through this and it looks good to me... but I think I'd like further opinions before merging this potentially dangerous patch. I think it's worth some discussion. Maybe post on Slack eliciting opinions?

typing/typeopt.ml
dkalinichenko-js: Oh, right.

testsuite/tests/typing-layouts-or-null/non_float_array.ml
dkalinichenko-js: We already have tests printing `-dlambda`. I think the churn is worth clearly seeing what we're getting here.

ccasin: I tend to think it's better to find a way to test whether the generated program does or doesn't do the thing you want, precisely because this style of test introduces a ton of churn.  But if there's no way to test whatever that property is (it can be tricky!) please use `-dno-unique-ids`.

ccasin: Additionally, if this can't be eliminated, it would be good to write a comment explaining "what we're getting here".  If I don't know anything about this test, but I introduce a change in lambda that causes me to accept a diff here, what do I need to check to make sure the test is still working?

dkalinichenko-js: I think the specific lambda (or flambda) output is what we really care about here. Generated code would be even better, but we can't display that in expect tests. But lambda is almost just as good.

Thanks for pointing out `-dno-unique-ids`!

PR #3980, opened 2025-05-08
Newly created type variables get exactly the jkind from the annotation

dkalinichenko-js:
Change type annotations on new variables from being upper bounds to being exact values. This makes examples like `val f : ('a : value_or_null) -> 'a` check `'a : value_or_null` properly instead of confusingly inferring `'a : value`.

Unfortunately, this makes typechecking dependent on the order annotations: `val f2 : ('a : value_or_null) -> 'a -> 'a` infers `'a : value_or_null`, but `val f2' : 'a -> ('a : value_or_null) -> 'a` infers `'a : value`. We can probably fix this too, by defaulting only at the end when we know there are no annotations on `'a`. 

It also would be nice to require equality instead of subkinding here. Perhaps it's not OCaml-y, but it's way less confusing.

### DIFF
--- a/testsuite/tests/typing-layouts-or-null/variables.ml
+++ b/testsuite/tests/typing-layouts-or-null/variables.ml
@@ -282,8 +282,7 @@ Error: Signature mismatch:
          because of the definition of f at line 4, characters 6-7.
 |}]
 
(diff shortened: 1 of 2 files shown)


goldfirere:
As discussed on Slack, I think the order-dependence here is pretty bad. I'd like to search for a different solution.

PR #3981, opened 2025-05-09
Remove subkinds from value slots

Ekdohibs:
When the reaper replaces some values by poison, the subkind information in value slots can become incorrect. This fixes the problem by completely removing the subkind information in value slots, which is otherwise only used for putting tagged immediates in the non-scanned area of closures.

### DIFF
--- a/middle_end/flambda2/simplify/simplify_apply_expr.ml
+++ b/middle_end/flambda2/simplify/simplify_apply_expr.ml
@@ -417,7 +417,7 @@ let simplify_direct_partial_application ~simplify_expr dacc apply
   in
   let applied_unarized_args, _ =
     Misc.Stdlib.List.map2_prefix
-      (fun arg kind -> arg, K.With_subkind.kind kind)
+      (fun arg kind -> arg, kind)
       args
       (Flambda_arity.unarize param_arity)
   in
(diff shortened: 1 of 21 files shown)


lthls (approved):
This patch has the following negative consequences:
- All value slots of kind `Value`, including immediates, are now stored in the scannable part of the closure
- All value slots of kind `Value`, for large closures and static closures, will be initialised using `caml_initialize`. For immediates, before this PR a direct write could be used.
- In some very uncommon cases, when simplifying a value slot projection the flambda2 type for the result will be less precise than it was before.

These are expected consequences, and I am in favour of merging this PR now to allow progress on unboxing of free variables, and work on restoring the lost functionality in a later PR (at least the first two points; the third one is likely irrelevant).

mshinwell:
I'm not sure actually.  How much work is it to restore the old behaviour for immediates here?

middle_end/flambda2/simplify/simplify_apply_expr.ml
```diff
@@ -417,15 +417,15 @@ let simplify_direct_partial_application ~simplify_expr dacc apply
   in
   let applied_unarized_args, _ =
     Misc.Stdlib.List.map2_prefix
-      (fun arg kind -> arg, kind)
+      (fun arg kind -> arg, K.With_subkind.kind kind)
```
lthls, line 420: I think you should revert this part of the patch, and instead update `mk_value_slot` below to use the subkind information to set `is_always_immediate`.

PR #3986, opened 2025-05-09
New computation for signal stack sizes

mshinwell:
The existing computation for signal stack sizes can be wrong on Intel machines with AMX extensions, causing `sigaltstack` to fail.  On Linux at least it is necessary to take into account the "dynamic" minimum signal stack size returned by `getauxval` in addition to any values for e.g. `SIGSTKSZ` (which may or may not be constant) in the C library.  For example, with a C library that has a constant `SIGSTKSZ` and `MINSIGSTKSZ`, the required minimum that becomes enforced by `sigaltstack` after AMX context initialisation can actually be larger than both of these quantities. A similar problem is described here: https://bugs.python.org/issue46968

For the moment this is conditionalised for Linux only, it can be extended in the future if necessary.

This PR also adds some reporting of `errno` in case of these failures.

### DIFF
--- a/otherlibs/systhreads/st_stubs.c
+++ b/otherlibs/systhreads/st_stubs.c
@@ -134,6 +134,7 @@ struct caml_thread_struct {
   char * async_exn_handler;  /* saved value of Caml_state->async_exn_handler */
   memprof_thread_t memprof;  /* memprof's internal thread data structure */
   void * signal_stack;       /* this thread's signal stack */
+  size_t signal_stack_size;  /* size of this thread's signal stack in bytes */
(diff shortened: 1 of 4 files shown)


mshinwell:
One problem with the `4 *` is that it seems newer glibc versions set the user-facing `MINSIGSTKSZ` to `SIGSTKSZ`, so we will end up multiplying by 16.  This is probably fine for now, but we may want to revisit.

jvanburen (approved):

mshinwell:
I've cherry-picked this onto `main` (`2b7f4e22cc950311d2450e57bcf8a2920a05dbea`)

PR #3987, opened 2025-05-09
Add bounds checks to the frame table emission code

spiessimon:
In preparation of re-enabling #3931, this PR adds bounds checks to the emission code for frame tables. Due to recent compiler changes, some integers that used to be non-negative (the live offsets in the frame table) can now be negative. The changes in this PR make it clear which ranges are expected for various constants in the frame table emission code. 

Notably, I have left `efa_word` alone for now. I believe it is a signed integer of target-platform width, but I wasn't completely sure.

### DIFF
--- a/backend/emitaux.ml
+++ b/backend/emitaux.ml
@@ -198,9 +198,12 @@ let record_frame_descr ~label ~frame_size ~live_offset debuginfo =
 type emit_frame_actions =
   { efa_code_label : Label.t -> unit;
     efa_data_label : Label.t -> unit;
-    efa_8 : int -> unit;
-    efa_16 : int -> unit;
-    efa_32 : int32 -> unit;
+    efa_i8 : Numbers.Int8.t -> unit;
+    efa_i16 : Numbers.Int16.t -> unit;
+    efa_i32 : Int32.t -> unit;
+    efa_u8 : Numbers.Uint8.t -> unit;
+    efa_u16 : Numbers.Uint16.t -> unit;
+    efa_u32 : Numbers.Uint32.t -> unit;
     efa_word : int -> unit;
     efa_align : int -> unit;
     efa_label_rel : Label.t -> int32 -> unit;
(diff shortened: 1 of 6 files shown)


runtime/caml/frame_descriptors.h
```diff
@@ -60,6 +60,14 @@
 #define FRAME_RETURN_TO_C 0xFFFF
 #define FRAME_LONG_MARKER 0x7FFF
 
+
+/* CR sspies: The frame descriptions below are no longer accurate with respect to
+   whether the integers in the description are signed or unsigned. Due to recent compiler
+   changes, the live offsets can now be negative. To make the integer ranges explicit,
```
gretay-js, line 66: only on runtime4

spiessimon: Oh, then I'm confused. I thought the files in `runtime/` are for runtime 5.

spiessimon: I've updated the comment and moved it to the relevant place in runtime 4.

backend/emitaux.ml
```diff
@@ -389,7 +435,7 @@ let emit_frames a =
       a.efa_label_rel
         (label_defname d.dinfo_file defname loc)
         (Int64.to_int32 info);
-      a.efa_32 (Int64.to_int32 (Int64.shift_right info 32));
+      emit_i32 (Int32.to_int (Int64.to_int32 (Int64.shift_right info 32)));
```
gretay-js, line 438: same as above if target is 32-bit.

backend/emitaux.ml
```diff
@@ -198,9 +198,12 @@ let record_frame_descr ~label ~frame_size ~live_offset debuginfo =
 type emit_frame_actions =
   { efa_code_label : Label.t -> unit;
     efa_data_label : Label.t -> unit;
-    efa_8 : int -> unit;
-    efa_16 : int -> unit;
-    efa_32 : int32 -> unit;
+    efa_i8 : Numbers.Int8.t -> unit;
```
gretay-js, line 201: `efa_i8` is currently used. maybe remove or is should have `emit_i8`  to go with it.

backend/emitaux.ml
```diff
@@ -209,6 +212,41 @@ type emit_frame_actions =
...
+      Misc.fatal_errorf
+        "attempting to emit signed 16-bit integer %d out of range" n
+    else a.efa_i16 (Numbers.Int16.of_int_exn n)
+  in
+  let emit_u16 n =
+    if n < 0 || n > 0xFFFF
+    then
+      Misc.fatal_errorf
+        "attempting to emit unsigned 16-bit integer %d out of range" n
+    else a.efa_u16 (Numbers.Uint16.of_nonnegative_int_exn n)
```
gretay-js, line 234: `of_int_exn` and `of_nonnegative_int_exn` in `Numbers` already has this check, why having another one here?

gretay-js (commented):
I have a few small questions about the change.

backend/emitaux.ml
```diff
@@ -209,6 +212,41 @@ type emit_frame_actions =
...
+  in
+  let emit_i32 n =
```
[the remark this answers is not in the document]
spiessimon: I've changed the functions to rely on, where available, the bounds checks in the functions from the standard library. As discussed offline, such a function does not seem to exist for `int32`, so I've implemented one here.

runtime/caml/frame_descriptors.h
```diff
@@ -60,6 +60,7 @@
 #define FRAME_RETURN_TO_C 0xFFFF
 #define FRAME_LONG_MARKER 0x7FFF
 
+
```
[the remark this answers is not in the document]
spiessimon: Removed.

backend/emitaux.ml
spiessimon: I've changed it to use `a.efa_i32` directly. I left a comment why it differs from the other emission sites (because of the 32-bit target issue that would truncate the possible range that we support here).

gretay-js (approved):

PR #3989, opened 2025-05-09
Add a check for >80 char lines to the CI

jra4:


ccasin (approved):
LGTM

As noted elsewhere, we should one day switch to using upstream's `check-typo` script (in `tools/`) but that will require fixing up all the existing 81+ character lines.  This is a good stopgap to prevent the problem from getting worse.

PR #3992, opened 2025-05-12
Fix emit_stores

gretay-js:
Fix special handling of initializing stores in selection: the `byte_offset` was incorrectly initialized. As a result, the base address register was not reset as needed in some cases, resulting in an invalid assembly.   This PR also fixes the machtype of the base register used in case of reset.

Re-enable one of the tests that was failing on arm64.

### DIFF
--- a/backend/cfg_selectgen.ml
+++ b/backend/cfg_selectgen.ml
@@ -654,31 +654,28 @@ module Make (Target : Cfg_selectgen_target_intf.S) = struct
               | Maybe_out_of_range ->
                 Target.is_store_out_of_range chunk ~byte_offset:!byte_offset
             in
-            let _ =
-              match is_out_of_range with
-              | Within_range -> ()
-              | Out_of_range ->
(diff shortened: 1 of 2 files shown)


backend/cfg_selectgen.ml
```diff
@@ -654,11 +654,12 @@ module Make (Target : Cfg_selectgen_target_intf.S) = struct
               | Maybe_out_of_range ->
                 Target.is_store_out_of_range chunk ~byte_offset:!byte_offset
             in
-            let new_addressing_mode =
+            let _ =
```
xclerc, line 657: Is the binding needed?
(If we want to keep it, I
think it would be useful
to have a type annotation.)

xclerc (approved):

backend/cfg_selectgen.ml
gretay-js: Fixed.

PR #4000, opened 2025-05-14
Provide mechanism to save CFG before register allocation

xclerc:
It is already possible to save the
CFG into a `cmir` file right before
before emission. This pull request
adds support for saving the CFG
right before register allocation.

This feature is not for end users,
but rather for register allocation
development. The pull request
adds `tools/regalloc/regalloc.ml`
that can be used to run register
allocation on a `cmir` file.


tools/regalloc/regalloc.ml
```diff
@@ -0,0 +1,82 @@
...
+  in
+  let all_regs =
+    Cfg_with_layout.fold_instructions cfg_with_layout ~init:Reg.Set.empty
+      ~instruction:add_regs ~terminator:add_regs
+  in
+  let relocatable_regs = Reg.Set.filter Reg.is_unknown all_regs in
+  Printf.eprintf "    %d register(s)...\n%!" (Reg.Set.cardinal relocatable_regs);
+  Reg.For_testing.set_state
+    ~stamp:(succ (Reg.Set.max_elt relocatable_regs).stamp)
+    ~relocatable_regs:(List.of_seq (Reg.Set.to_seq relocatable_regs));
```
gretay-js, line 35: Is this part intended to guarantee that the "shared state" is the same as in the original compilation? Instead of trying to reverse-engineer this information, can we just save it alongside the cfg everything we need it? Do we need to save the label too somewhere?  It's saved as part of Cfg file format, for the purpose of "start from emit" it only need to not overlap with existing labels, but maybe we want something more precise here for regalloc?

tools/regalloc/regalloc.ml
```diff
@@ -0,0 +1,82 @@
...
+  let cfg_with_infos = Cfg_with_infos.make cfg_with_layout in
+  let start_time = cpu_time () in
+  let (_ : Cfg_with_infos.t) =
+    match register_allocator with
+    | GI -> Regalloc_gi.run cfg_with_infos
+    | IRC -> Regalloc_irc.run cfg_with_infos
+    | LS -> Regalloc_ls.run cfg_with_infos
+  in
+  let end_time = cpu_time () in
+  Printf.eprintf "  register allocation took %gs...\n%!" (end_time -. start_time);
```
gretay-js, line 45: This is probably just a placeholder, but it should have some `GC` stats too, right?  At least to explain the noise.

gretay-js (approved):

PR #4001, opened 2025-05-14
Optionally disable hugepages for stacks

stedolan:
Hugepages are not necessarily a good idea for stacks: they mean that every stack takes 2MB of RAM, even though most programs confine themselves to a few k of stacks. This patch adds a new GC tweak to optionally disable hugepages for stacks (`Xnohugepage_stacks=1`), by passing `MADV_NOHUGEPAGE` to `madvise`.

On this program:
```
let n = Atomic.make 0
let () =
  let _ = List.init 100 (fun _ -> Thread.create (fun () -> Atomic.incr n; Unix.sleep 100) ()) in
  while Atomic.get n < 100 do Thread.yield () done;
  ()
```

the new flag makes the memory consumption go from 400 MB to 200 MB. (I think the remaining 2MB per thread is the glibc pthread stack).

### DIFF
--- a/runtime/fiber.c
+++ b/runtime/fiber.c
@@ -66,6 +66,8 @@ uintnat caml_init_main_stack_wsz = 0;   /* -Xmain_stack_size= */
 uintnat caml_init_thread_stack_wsz = 0; /* -Xthread_stack_size= */
 uintnat caml_init_fiber_stack_wsz = 0;  /* -Xfiber_stack_size= */
 
+uintnat caml_nohugepage_stacks = 0;
+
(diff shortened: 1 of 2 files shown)


NickBarnes:
I concur that the remaining 2MiB Rss per thread is glibc pthread stacks (which are incidentally mapped at 9.56 MiB); we could trim these if we cared, by mapping them ourselves and using `pthread_attr_setstack`. Each thread also gets a 132k mapping (with 8k of RSS) from the GlibC malloc arena, which we don't care about.

NickBarnes (approved):
LGTM. Manually tested.

PR #4006, opened 2025-05-15
Skip the 80ch check on autoformatted files

jra4:
See (https://github.com/ocaml-flambda/flambda-backend/pull/4005) for an example

### DIFF
--- a/.github/workflows/80ch.sh
+++ b/.github/workflows/80ch.sh
@@ -49,8 +49,13 @@ do
   # (a more sophisticated script would instead remove %%expect blocks)
   [[ "$changed_file" == testsuite/tests/* ]] && continue
 
+  # Don't check autoformatted files (there's a bug in `ocamlformat` that
+  # sometimes produces lines over 80 characters)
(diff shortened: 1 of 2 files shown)


.github/workflows/80ch.sh
```diff
@@ -49,8 +49,17 @@ do
...
   [[ "$changed_file" == testsuite/tests/* ]] && continue
 
+  # Don't check autoformatted files (there's a bug in `ocamlformat` that
+  # sometimes produces lines over 80 characters)
+  ocamlformat -m 1 "$changed_file" &> /dev/null || continue
+  #           |^^^
+  #           This setting causes `ocamlformat` to crash, but
+  #           only if the formatter is run. Thus, we can use
+  #           the return code to determine whether the file
+  #           is autoformatted
```
ccasin, line 59: This is clever, but feels pretty janky/brittle.  I think it wouldn't be too hard to have the script make a list of all the files listed in `.ocamlformat-enable` files, and just not check those? (Or have an AI make the script do that - they are pretty good at bash)

mshinwell (approved):

PR #4012, opened 2025-05-16
Use OxCaml instead of OCaml

cuihtlauac:
When relevant, according to my understanding. And I may have missed
some

### DIFF
--- a/jane/doc/extensions/_01-stack-allocation/reference.md
+++ b/jane/doc/extensions/_01-stack-allocation/reference.md
@@ -6,7 +6,7 @@ title: Reference
 # Stack Allocations Reference
 
 The goal of this document is to be a reasonably complete reference to stack
-allocations in OCaml. For a gentler introduction, see [the
+allocations in OxCaml. For a gentler introduction, see [the
 introduction](../intro).
 
 The stack allocations language extension allows the compiler to
@@ -44,7 +44,7 @@ let f = ref (stack_ `Foo)
 Error: This expression is not an allocation site.
 ```
 
-Most OCaml types can be stack-allocated, including records, variants,
+Most OxCaml types can be stack-allocated, including records, variants,
 polymorphic variants, closures, boxed numbers and strings. However, certain
 values cannot be stack-allocated, and will always be on the GC heap,
 including:
@@ -60,7 +60,7 @@ including:
 ### Runtime behavior
 
 At runtime, stack allocations do not take place on the function call stack, but on a
-separately-allocated stack that follows the same layout as the OCaml
+separately-allocated stack that follows the same layout as the stock OCaml
 minor heap. In particular, this enables local-returning functions
 without the need to copy returned values. See "Use `exclave_` to return a local value"
 below for more details.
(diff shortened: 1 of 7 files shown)


jane/doc/extensions/_01-stack-allocation/reference.md
goldfirere, line 1: I think I favor keeping the "stock" qualifier. Except maybe if that word is less familiar, change it to "mainstream"? The appearance (or lack) of the `x` is a bit subtle.

jane/doc/extensions/_05-kinds/syntax.md
```diff
@@ -308,7 +308,7 @@ about its implementation.)
 
 ## Kind annotations on variables
 
-Type variables come in two flavors in OCaml: rigid and unifiable. Rigid type
+Type variables come in two flavors in OxCaml: rigid and unifiable. Rigid type
```
goldfirere, line 311: Well, the statement is true of OCaml just as much as OxCaml. I'm happy for the change here, but I just wanted to point out that this is not new.

jane/doc/extensions/_08-miscellaneous-extensions/polymorphic-parameters.md
```diff
@@ -76,7 +76,7 @@ let r = create forty_two
 ## Limitations
 
 All polymorphic parameter require an annotation. Without an annotation
```
goldfirere, line 78:
```suggestion
Each polymorphic parameter requires an annotation. Without an annotation,
```

goldfirere (approved):

jane/doc/extensions/_01-stack-allocation/reference.md
```diff
@@ -60,7 +60,7 @@ including:
 ### Runtime behavior
 
 At runtime, stack allocations do not take place on the function call stack, but on a
-separately-allocated stack that follows the same layout as the OCaml
+separately-allocated stack that follows the same layout as the stock OCaml
```
goldfirere, line 63: Sorry to be picky here, but I don't think we should add "stock" to all uses of OCaml. Putting the word in here suggests (to me) that there is a difference between the minor heap layout in OCaml vs OxCaml. We could be pedantic and write `O(x)Caml` here, but I'd prefer not so. Maybe "as the OCaml minor heap (unchanged in OxCaml)"?

jane/doc/extensions/_01-stack-allocation/reference.md
```diff
@@ -834,7 +834,7 @@ the local or global status of parts of a value, but rather tracks this
 only once per variable or expression. There is one exception to this
 rule, as follows.
 
-In OCaml, it is possible to simultaneously match on multiple values:
+In stock OCaml, it is possible to simultaneously match on multiple values:
```
goldfirere, line 837: Here too.

jane/doc/extensions/_01-stack-allocation/reference.md
```diff
@@ -862,7 +862,7 @@ let a, b, c =
   x, y, z
 ```
 
-Again, there is no actual syntax for this in OCaml: that's a binding
+Again, there is no actual syntax for this in stock OCaml: that's a binding
```
goldfirere, line 865: And here. Maybe the best way to address this is some kind of introduction explaining our terminology? But I worry no one would actually read that introduction.

PR #4013, opened 2025-05-16
Add [Thread.use_domains] to force us into the domains locking mode

lpw25:
Currently, you cannot start your first domain from a thread other than the main one. Working around this restriction requires starting a dummy domain on the many thread first. However, this can be harder than it seems because you also can't fork once you've started any domains.

This PR adds a function that avoids this restriction: as long as you call [Thread.use_domains] on the main thread first, you are allowed to start your first domain from any thread you like.

This restriction will go away once we fix the locking scheme. At that point this function should be removed.

### DIFF
--- a/otherlibs/systhreads/st_stubs.c
+++ b/otherlibs/systhreads/st_stubs.c
@@ -611,6 +611,27 @@ static void caml_thread_domain_stop_hook(void) {
 
 static atomic_bool threads_initialized = false;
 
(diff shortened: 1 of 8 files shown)


NickBarnes (changes requested):
This LGTM, apart from the highlighted error messages. Reviewing this area of code led to #4065.

lpw25:
Fixed the error message. Should be good to go now

NickBarnes (approved):

PR #4017, opened 2025-05-16
Block indices

rtjoa:
Add explicit *indices* into a block, gated by `layouts_beta`. A quick example follows below. For full documentation, see [jane/doc/extensions/_03-unboxed-types/03-block-indices.md](https://github.com/oxcaml/oxcaml/blob/rtjoa.block-indices/jane/doc/extensions/_03-unboxed-types/03-block-indices.md).

```ocaml
open Stdlib_beta

type pt = { x : int; y : int }
type line = { p : pt#; q : pt# }

let mk_idx () : (line, int) idx_imm = (.q.#y)

let get_coord (line : line) (i : (line, int) idx_imm) : int =
  Idx_imm.unsafe_get line i

let first_x (): (pt# array, int) idx_mut = (.(0).#x)

let inc_coord (pts : 'a) (i : ('a, int) idx_mut) =
  Idx_mut.unsafe_set pts i (Idx_mut.unsafe_get pts i + 1)
```

## Testing

- Generated tests by enumerating records/arrays supporting block indices in [testsuite/tests/records-and-block-indices](https://github.com/ocaml-flambda/flambda-backend/tree/rtjoa.block-indices/testsuite/tests/records-and-block-indices)
  * Compared get/set using block indices against normal get/set
  * Checked that each block index `(<prefix><suffix>)` is equivalent to the deepening operation `(.idx(<prefix>)<suffix>)`
 - Typing and handwritten runtime tests in [testsuite/tests/typing-layouts-block-indices](https://github.com/ocaml-flambda/flambda-backend/tree/rtjoa.block-indices/testsuite/tests/typing-layouts-block-indices)

## Reviewing
- @ccasin to "whole-feature" review. Review commit-by-commit.
- @mshinwell to do targeted review of the middle-end, runtime, and runtime tests (`backend/`, `bytecomp/`, `middle_end/`, `testsuite/tests/typing-layouts-block-indices/block_indices.ml`, `testsuite/tests/records-and-block-indices/test_generation.ml`)
- Note that much of the diff is due to generated tests (beginning with `generated_`), which generates tests of the same shape for many types. The best way to review these is to review the file that does the generation, then look at the tests for just a few types.


testsuite/tests/typing-layouts-block-indices/basics_block_indices.ml
```diff
@@ -0,0 +1,852 @@
...
+
+(* Disambiguation of block access field *)
+
+type t1 = { mutable a : string; b : int }
+type t2 = { mutable a : string; b : int; c : string }
+
+let a2 () = (.a)
+let b2 () = (.b)
+let a1 () : (t1, _) idx_mut = (.a)
+let b2 () : (t1, _) idx_imm = (.b)
```
ccasin: I think you meant to call this `b1`

testsuite/tests/typing-layouts-block-indices/basics_block_indices.ml
```diff
@@ -0,0 +1,852 @@
...
+
+(* Unboxed access disambiguates the next unboxed access *)
+type wrap_r = { r : int r# }
+let f () = (.r.#u.#x)
+[%%expect{|
+type wrap_r = { r : int r#; }
+val f : unit -> (wrap_r, int) idx_imm = <fun>
+|}]
+
+(* Disambiguation causes earlier error while typechecking block access *)
```
ccasin, line 116: I don't quite understand this comment. I think there is no disambiguation going on here - the labels are all in the same scope as the function, and deleting the type annotation doesn't change the error.

jane/doc/extensions/_03-unboxed-types/block-indices.md
```diff
@@ -0,0 +1,215 @@
...
+let inc_coord (pts : 'a) (i : ('a, int) idx_mut) =
+  (* Equivalent to [pts.(i) <- pts.(i) + 1] when [i] is in bounds and [pts] is
+     an [int array]. But this function could also be used update a mutable
+     record containing an [int]. *)
+  Idx_mut.unsafe_set pts i (Idx_mut.unsafe_get pts i + 1)
+```
+
+# Overview
+
+A block index is an opaque, explicit index to an element. The language feature
```
ccasin:
```suggestion
A block index is an opaque, explicit index to an element in a larger structure. The language feature
```

jane/doc/extensions/_03-unboxed-types/block-indices.md
```diff
@@ -0,0 +1,215 @@
...
+```ocaml
+module Stack : sig
+  type 'a t
+  val empty : 'a t
+  val push : 'a t -> 'a -> unit
+  val pop : 'a t -> 'a
+  val update_top : 'a t -> ('a, 'b) idx_mut -> 'b -> unit
+  (* [update_top] normally isn't possible without exposing the representation.
+     One could use ['a t -> ('a -> unit) -> unit], but the closure is less
+     efficient. *)
```
ccasin: I confess I don't find this argument about the closure particularly compelling. The closure version is also nicer in that it gives you access to the current value of the element, which you probably usually want.  Of course you can do that with the block index version too, but it'll have to take a closure (or I guess add a `peek` function to the API).

Probably fine to leave it in, but maybe put the other use case first.

jane/doc/extensions/_03-unboxed-types/block-indices.md
```diff
@@ -0,0 +1,215 @@
...
+# Representation of block indices
+
+Consider the following type:
+
+```ocaml
+type a = #{ s : string; i : int64# }
+type b = #{ i : int64#; a : a; s : string }
+type c = { mutable b : b; s : string }
+```
+The layout of `c` has the shape
```
ccasin: I found this explanation slightly confusing.  Mainly, (a) I think it should telegraph up front the problem that `a` and `b`'s fields do not appear contiguously in `c`'s runtime representation, and (b) I found it a little confusing to say "the layout of c has the shape ..." (because the layout of `c` is `value`).

Here's another stab:

> The record `c` presents an interesting problem for block indices: the fields of the `b` and `a` records it contains are not actually contiguous at runtime, when using the native code compiler.  This problem is caused by the mixed block represntation (link), which mandates that we must reorder fields so that values come before unboxed types.
>
> Its actual representation looks like this: ...

And then carry on with your picture.

testsuite/tests/records-and-block-indices/generated_array_idx_access_bytecode_test.ml
```diff
@@ -0,0 +1,2988 @@
+(* TEST
+ include stdlib_stable;
+ include stdlib_beta;
+ include stdlib_upstream_compatible;
+ modules = "idx_repr.ml stubs.c";
+ flags = "-extension layouts_alpha";
+ bytecode;
+*)
```
ccasin, line 8: for several of these new tests, including this one, the only difference between the bytecode and native code versions is the definition of `int64x2_u_equal`.  I think this probably suggests we should either have just one copy of the test or that there are missing test cases.

testsuite/tests/records-and-block-indices/generated_array_idx_access_bytecode_test.ml
```diff
@@ -0,0 +1,2988 @@
...
+    if not test then failwithf "test 132 failed %d %d" size i;
+    (* Path: [.#b27] *)
+    let el = #{ el with b27 = next_el.#b27 } in
+    Idx_mut.unsafe_set a ((.(i).#b27) : (t27 array, _) idx_mut) next_el.#b27;
+    mark_test_run 133;
+    let test = eq (Idx_mut.unsafe_get a (.(i))) el in
+    if not test then failwithf "test 133 failed %d %d" size i;
+    (* Paths of depth 2 *)
+    let next_el = mk_value (i + 100 * 2) in
+    (* We update [a.(i)] to become [next_el] by setting all paths of depth 2 *)
```
ccasin: Skimming these generated tests, I find this comment a little confusing in cases like this one - I think we're not quite updating `a.(i)` to become `next_el` because it has paths of depth one whose elements aren't changed.

middle_end/flambda2/parser/flambda_to_fexpr.ml
```diff
@@ -652,7 +652,8 @@ let ternop env (op : Flambda_primitive.ternary_primitive) : Fexpr.ternop =
     let ask = fexpr_of_array_set_kind env ask in
     Array_set (ak, ask)
   | Bytes_or_bigstring_set (blv, saw) -> Bytes_or_bigstring_set (blv, saw)
-  | Bigarray_set _ | Atomic_compare_and_set _ | Atomic_compare_exchange _ ->
+  | Bigarray_set _ | Atomic_compare_and_set _ | Atomic_compare_exchange _
+  | Write_offset _ ->
```
ccasin: It's fine to do it in a follow-up PR, but I think it would be good to implement `fexpr` support for the block index primitives.  It probably takes just 10 minutes, and in my experience can be a real annoyance for minimizing problems and making reproducable tests when things aren't supported.  (I defer to @mshinwell though).

testsuite/tests/records-and-block-indices/idx_repr.ml
```diff
@@ -0,0 +1,47 @@
...
+  | Bytecode of { path : int list }
+  | Native of { offset : int; gap : int }
+
+external magic_box_bits64 : ('a : bits64) 'b . 'a -> 'b =
+  "%box_int64"
+external lessthan_if_bytecode : int -> int -> bool =
+  "caml_lessthan" "caml_greaterthan"
+
+let of_idx idx =
+  let is_bytecode = lessthan_if_bytecode 0 1 in
```
ccasin: Delightful, but I think you can use `Sys.backend_type`.

jane/doc/extensions/_03-unboxed-types/block-indices.md
```diff
@@ -0,0 +1,215 @@
...
+
+In-memory representation:
+- In the native compiler, the offset and gap are packed into
+  a single `bits64`. There are two subcases:
+  * The index is to product containing both values and non-values. In this
+    case, the offset is the lower 48 bits and the gap is the upper 16 bits.
+  * The index is to all values/non-values. In this case, all 64 bits are used
+    for the offset.
+- In the bytecode compiler, the field positions are stored as tagged integers
+  in single block with tag 0.
```
ccasin: I think we should spell out more explcitiy somewhere (here or in a comment in an appropriate file) why this is safe.  A couple considerations:
* Having a different representation on native code and bytecode is scary - in general the compiler promises that if you marshal data on bytecode and unmarshal it on native code, you get the same thing back , and vice versa (though there are some known existing ways this is false).  I think this OK, though, because the Marshal module will raise when attempting to marshall/unmarshall a `bits64` or a block containing a `bits64`, at least on native code, so there is no way to witness the difference here.
* On bytecode, block indices have layout `bits64` but are in fact values.  I spent a little bit trying to see if I could convince the bytecode compiler to skip a needed `caml_modify` as a result of this mismatch, but ultimately I think it's fine - the story we already have in place for, e.g., `int64#` also works here.

testsuite/tests/records-and-block-indices/do_test_generation.ml
```diff
@@ -13,6 +13,102 @@
   include stdlib_upstream_compatible;
   ocamlopt.opt;
   set test_name = "";
+  { test_name = "record_idx_access";
```
ccasin, line 16: There are a lot of tests here.  Which is great!  But also I'm a little concerned about the impact on the time it takes to run tests.  I did a little experiment, and found that this and #4174 combined are a pretty substantial increase on wall-clock test running time (about 100% - from ~1:12 to ~2:20 in actual human waiting time).  Here is what I observed on the box I develop on, just running `/usr/bin/time make runtest-upstream` a couple times for each version:

```
* After block indices:

773.83user 202.60system 2:25.41elapsed 671%CPU (0avgtext+0avgdata 1058636maxresident)k
4097072inputs+7187952outputs (4830major+39429238minor)pagefaults 0swaps

764.06user 178.25system 2:19.83elapsed 673%CPU (0avgtext+0avgdata 1058800maxresident)k
2945688inputs+7186976outputs (1051major+33549462minor)pagefaults 0swaps

* After 4174:

696.52user 150.39system 1:19.64elapsed 1063%CPU (0avgtext+0avgdata 1190852maxresident)k
808inputs+6980784outputs (834major+26119058minor)pagefaults 0swaps

696.94user 150.79system 1:19.48elapsed 1066%CPU (0avgtext+0avgdata 1190860maxresident)k
8784inputs+6980040outputs (803major+26112044minor)pagefaults 0swaps

* Before 4174:

662.58user 147.90system 1:11.22elapsed 1138%CPU (0avgtext+0avgdata 874032maxresident)k
12448inputs+6955688outputs (778major+26121226minor)pagefaults 0swaps

677.06user 150.21system 1:12.03elapsed 1148%CPU (0avgtext+0avgdata 874032maxresident)k
21672inputs+6955976outputs (748major+26115883minor)pagefaults 0swaps
```

It makes sense to add a bunch of tests, but we should spend a little time thinking about whether we can trim them or parallelize them so the impact is more modest (I think a 20% increase in compiler test time for a new language feature like this would be just fine, for example).

testsuite/tests/typing-layouts-block-indices/basics_block_indices.ml
```diff
@@ -0,0 +1,852 @@
...
+
+let idx_imm x = (.idx_imm(x))
+let idx_mut x = (.idx_mut(x))
+[%%expect{|
+val idx_imm : ('a, 'b) idx_imm -> ('a, 'b) idx_imm = <fun>
+val idx_mut : ('a, 'b) idx_mut -> ('a, 'b) idx_mut = <fun>
+|}]
+
+(**************)
+(* Modalities *)
```
ccasin, line 479: As an aside, I think we could support these if we put the modality box types in predef.  E.g., if the typechecker had a name for a type like this:
```ocaml
type 'a gel = { x : 'a @@ global } [@@unboxed]
```

Then I think we could type an index to this field:
```ocaml
type t  = { s : string @@ global }
````
as a `(t, string gel) idx_imm`

(I know we've discussed that we can do this once we rework the language to treat modalities as the primitive thing. Probably we've also discussed this workaround, and I've just forogotten. In any case, definitely not for this PR)

File: testsuite/tests/typing-layouts-block-indices/basics_block_indices.ml [hunk 13]
```diff
@@ -0,0 +1,852 @@
...
+
+type 'a id = { id : 'a }
+type 'a id = { mutable mut : 'a }
```
ccasin: I think it makes the test a little confusing that these two types have the same name - maybe call the second `id_mut` or something.  Or really `box` would be a better name than `id` for both.

testsuite/tests/typing-layouts-block-indices/basics_block_indices.ml
```diff
@@ -0,0 +1,852 @@
...
+type 'a global = { global : 'a @@ global }
+type 'a aliased = { aliased : 'a @@ aliased }
+type 'a many = { many : 'a @@ many }
+type 'a unyielding = { unyielding : 'a @@ unyielding }
+type 'a portable = { portable : 'a @@ portable }
+type 'a contended = { contended : 'a @@ contended }
+type 'a mut_not_global = { mutable mut_not_global : 'a @@ local }
+type 'a mut_not_many = { mutable mut_not_many : 'a @@ once }
+type 'a mut_not_global_nor_aliased =
+  (* mut_not_aliased not possible bc global implies aliased *)
```
ccasin: I think you can override the default modality, including just part of it, like `type t = { mutable x : string @@ unique }`.  Here `unique` is really the identity modality, but it overrides the default aliased modality.

testsuite/tests/typing-layouts-block-indices/basics_block_indices.ml
```diff
@@ -0,0 +1,852 @@
...
+[%%expect{|
+val ok :
+  ('a : value mod non_float).
+    unit -> ('a unyielding# aliased# many# global# array, 'a) idx_mut =
+  <fun>
+|}]
+let ok () = (.mut.#mut_not_global.#id)
+[%%expect{|
+val ok : unit -> ('a id/2# mut_not_global# id/1, 'a) idx_mut = <fun>
+|}]
```
ccasin: Maybe they are elsewhere and I haven't spotted them yet, but it would also be good to test that values read through block indices have the modes expected due to the default modalities (so the mode of the block in the immutable case, and the mode of the block @@ global aliased ... in the mutable case).  And that values we write through mutable block indices must have the modes expected based on the mode of the block and the default modalities.

testsuite/tests/typing-layouts-block-indices/basics_block_indices.ml
```diff
@@ -0,0 +1,852 @@
...
+
+(************)
+(* Variance *)
```
ccasin: You've written another similar variance test [above](https://github.com/oxcaml/oxcaml/blob/b90d7c3ebed2e03fc1a493aa503bbaa92cde1064/testsuite/tests/typing-layouts-block-indices/basics_block_indices.ml#L269) and it would probably be good to put the two next to each other.

testsuite/tests/typing-layouts-block-indices/basics_block_indices.ml
```diff
@@ -0,0 +1,852 @@
...
+2 | let bad () : (abstract array, _) idx_mut = (.(0))
+                                               ^^^^^^
+Error: This expression has type "('a array, 'a) idx_mut"
+       but an expression was expected of type "(abstract array, 'b) idx_mut"
+       The kind of abstract is value
+         because of the definition of abstract at line 1, characters 0-13.
+       But the kind of abstract must be a subkind of
+         value_or_null mod non_float
+         because it's the element type (the second type parameter) for a
+         block index (idx or mut_idx).
```
ccasin, line 705: This error is a bit sad, in a couple ways.
1) I think it's pretty hard for the user to spot the actual problem (the `non_float`).  I wonder if it's worth extending the explanation in the second bit with something like "(Note: block indices are not supported into float arrays)". I'm unsure (that text could then probably also be reached in other scenarios where it's not relevant).
2) The second bit doesn't seem like a fully accurate explanation of the kind.  It implies that the second parameter of a block index must have layout `value`, for example, but that isn't true.

Obviously at some point we will do a bit re-imagining of mode and kind errors, so don't spend too much time thinking about this.

testsuite/tests/typing-layouts-block-indices/basics_block_indices.ml
```diff
@@ -0,0 +1,852 @@
...
+                                                ^^^^^^^
+Error: This expression has type "('a iarray, 'a) idx_imm"
+       but an expression was expected of type "(abstract iarray, 'b) idx_imm"
+       The kind of abstract is value
+         because of the definition of abstract at line 1, characters 0-13.
+       But the kind of abstract must be a subkind of
+         value_or_null mod non_float
+         because it's the element type (the second type parameter) for a
+         block index (idx or mut_idx).
+|}]
```
ccasin, line 760: I suppose this one is a good reason not to like my suggestion in point 1 just above.

testsuite/tests/typing-layouts-block-indices/basics_block_indices.ml
```diff
@@ -0,0 +1,852 @@
...
+
```
ccasin, line 761: I do think it's possible to hit even sadder errors.  E.g., for this program:
```ocaml
let f (x : float array) =
  let y = (.(42)) in
  Idx_mut.unsafe_get x y
```

I get the error:
```
File "/home/ccasinghino/tmp/foo.ml", line 5, characters 23-24:
5 |   Idx_mut.unsafe_get x y
                           ^
Error: This expression has type ('a array, 'a) idx_mut
       but an expression was expected of type (float array, 'b) idx_mut
       The kind of float is value mod many unyielding stateless immutable
         because it is the primitive type float.
       But the kind of float must be a subkind of value_or_null mod non_float
         because of the definition of y at file "/home/ccasinghino/tmp/foo.ml", line 4, characters 10-17.
```

Here I'm not even seeing the explanation that this has something to do with block indices.  I think this is probably hard to fix in the short term, but maybe record it as something we can improve in the future, here.

parsing/parsetree.mli
```diff
@@ -450,6 +450,8 @@ and expression_desc =
       (** [E1.l <- E2] *)
   | Pexp_array of mutable_flag * expression list
       (** [[| E1; ...; En |]] or [[: E1; ...; En :]] *)
+  | Pexp_idx of block_access * unboxed_access list
+      (** [(BA1 UA1 UA2 ...)] e.g. [(.foo.#bar.#baz)] *)
```
ccasin: I think this comment is slightly unidiomatic - most of the comments in this file either:
* Just show the concrete syntax (like the case just above), or
* Say something like `[Pexp_idx(ba, [ua1; ...; uan])] represents ...` (as in the `pexp_record` case above).

parsing/parsetree.mli
```diff
@@ -620,6 +622,20 @@ and function_constraint =
   }
 (** See the comment on {{!expression_desc.Pexp_function}[Pexp_function]}. *)
 
+and block_access =
+  | Baccess_field of Longident.t loc
+      (** [.foo] *)
+  | Baccess_array of mutable_flag * index_kind * expression
+      (** Mutable block accesses: [.(E)], [.L(E)], [.l(E)], [.n(E)]
```
ccasin:
```suggestion
      (** Mutable array accesses: [.(E)], [.L(E)], [.l(E)], [.n(E)]
```

parsing/parsetree.mli
```diff
@@ -620,6 +622,20 @@ and function_constraint =
...
+          Immutable array accesses: [.:(E)], [.:L(E)], [.:l(E)], [.:n(E)]
+
+          Indexed by [int], [int64#], [int32#], or [nativeint#], respectively.
+      *)
+  | Baccess_block of mutable_flag * expression
```
ccasin, line 644: Comment missing here

parsing/parsetree.mli
```diff
@@ -620,6 +622,20 @@ and function_constraint =
...
+
+and unboxed_access =
+  | Uaccess_unboxed_field of Longident.t loc
```
ccasin, line 650: and here

parsing/parse.ml
```diff
@@ -178,6 +178,14 @@ let prepare_error err =
...
       Location.errorf ~loc
         "Syntax error: Unexpected in module instance"
+  | Block_access_bad_paren loc ->
+      Location.errorf ~loc
+        "Syntax error: A parenthesis here can only follow one of: \n  \
+         %a, %a, %a, %a, %a, %a, %a, %a, %a, %a."
+        Style.inline_code "." Style.inline_code ".L" Style.inline_code ".l"
+        Style.inline_code ".n" Style.inline_code ".:" Style.inline_code ".:L"
+        Style.inline_code ".:l" Style.inline_code ".:n"
+        Style.inline_code ".idx_imm" Style.inline_code ".idx_mut"
```
ccasin, line 206: Can you add a test case demonstrating the ways to hit this error?  From the parser, it looks like there are three.

typing/jkind.mli
```diff
@@ -449,6 +449,11 @@ val of_new_legacy_sort_var :
 val of_new_legacy_sort :
   why:History.concrete_legacy_creation_reason -> 'd Types.jkind
 
+(** Same as [of_new_sort_var], but the jkind is lowered to [Non_float].
+    Defaulting the sort variable produces exactly [value].  *)
```
ccasin: I think the last line here is a bit ambiguous, as `value` is also the name of the kind.  Maybe clearer to say that defaulting the sort variable produces `[value mod non_float]`.

typing/typemode.ml
```diff
@@ -478,14 +478,47 @@ let mutable_implied_modalities (mut : Types.mutability) =
...
+      (* If this list is updated, the external bindings in the [Idx_imm] and
+         [Idx_mut] modules in [Stdlib_beta] may also have to be updated. *)
+      modality_of_list
+        [ Atom (Comonadic Areality, Meet_with Regionality.Const.legacy);
+          Atom (Comonadic Linearity, Meet_with Linearity.Const.legacy);
+          Atom (Comonadic Yielding, Meet_with Yielding.Const.legacy);
+          Atom (Monadic Uniqueness, Join_with Uniqueness.Const.legacy) ]
+    else Mode.Modality.Value.Const.id
+  in
+  match Mode.Modality.Value.Const.equate expected1 expected2 with
```
ccasin, line 452: If I understand correctly, this check is a sanity check that `mutable_implied_modalities` hasn't been changed in a way that's incompatible with the types we give to the block index primitives.  That's a great idea!

But there are really only two checks to do here (one for immutable and one for mutable) and it's a little sad from a compiler performance perspective that we run one of those every time this function is called. I think we should try to refator this to fix that. One possibility is to arrange for them to run at compiler startup, though I don't know of anything else that works like that. Another possibility is to pull out the check into a lazy, something like:
```ocaml
let block_index_modality_sanity_check : mut:bool -> lazy ... =
  let imm_check = lazy (... (* fatal_error in here if check fails *) ...) in
  let mut_check = lazy (... (* fatal_error in here if check fails *)...) in
  if mut then mut_check else imm_check
```
And then force this in `idx_expected_modalities`.

Or I'm sure you can imagine various other possibilities, perhaps therre is something more elegant.

typing/uniqueness_analysis.ml
```diff
@@ -2290,6 +2290,22 @@ let rec check_uniqueness_exp ~overwrite (ienv : Ienv.t) exp : UF.t =
...
+            index;
+            base_ty = _;
+            elt_ty = _;
+            elt_sort = _
+          } ->
+        check_uniqueness_exp ~overwrite:None ienv index
+      | Baccess_block (_, idx) -> check_uniqueness_exp ~overwrite:None ienv idx
+    in
+    let unboxed_access = function Uaccess_unboxed_field _ -> UF.unused in
+    UF.pars (block_access ba :: List.map unboxed_access uas)
```
ccasin: This is the same as leaving off the `uas` bit, right?

The instinct to be explicit and guard against future changes is good, but this does feel a little wasteful.  Maybe do something like:
```
    let _unboxed_access = function Uaccess_unboxed_field _ -> UF.unused in
    block_access ba
```
This way we are forced to revisit if another `Uaccess` case is added, but we don't actually bother allocating a new list and `par`ing it. You can write a comment claritying that if other non-`unused` cases are added it will be necessary to `par` over the list.

typing/value_rec_check.ml
```diff
@@ -720,6 +721,24 @@ let rec expression : Typedtree.expression -> term_judg =
...
+            ; elt_ty = _
+            ; elt_sort = _ } ->
+          expression index << Dereference
+        | Baccess_block (_, idx) ->
+          expression idx << Dereference
+      in
+      let unboxed_access = function
+        | Uaccess_unboxed_field _ -> empty
+      in
+      join (block_access ba :: List.map unboxed_access uas)
```
ccasin: Same note here as in `uniqueness_analysis` (if I understand your intent correctly)

typing/typecore.ml
```diff
@@ -4237,6 +4246,23 @@ let rec is_nonexpansive exp =
...
+            ; base_ty = _
+            ; elt_ty = _
+            ; elt_sort = _ } ->
+          is_nonexpansive index
+        | Baccess_block (_, idx) -> is_nonexpansive idx
+      in
+      let unboxed_access = function
+        | Uaccess_unboxed_field _ -> true
+      in
+      block_access ba && List.for_all unboxed_access uas
```
ccasin: Another spot to consider skipping some work.

typing/typecore.ml
```diff
@@ -5691,6 +5733,123 @@ and type_expect_
         exp_attributes = sexp.pexp_attributes;
         exp_env = env }
   in
+  let type_block_access expected_base_ty principal
```
ccasin: Leo recently pointed out that we've been bad citizens by letting `type_pat_aux` and `type_expect_` grow these big blocks of internal helper functions that really obscure the purpose and control flow of the functions they are in. We should do a pass to pull them all out into stand-alone helpers at some point. That's beyond the scope of this PR, but can you just check if it's easy to pull these new ones out, to avoid growing the problem?

typing/typecore.ml
```diff
@@ -5280,6 +5306,22 @@ let may_lower_contravariant_then_generalize env exp =
   if maybe_expansive exp then lower_contravariant env exp.exp_type;
   generalize exp.exp_type
 
+let generalize_type_block_access_result { ba; base_ty; el_ty; flat_float = _ } =
```
ccasin: I think this should be called `generalize_structure_type_block_access_result`.

typing/typecore.ml
```diff
@@ -6232,6 +6391,74 @@ and type_expect_
...
+      | _ ->
+        newgenvar (Jkind.Builtin.value ~why:Idx_base)
+    in
+    let expected_base_ty = expected_base_ty ty_expected in
+    let principal = is_principal ty_expected in
+    let { ba; base_ty; el_ty; flat_float; modality } =
+      with_local_level_if_principal ~post:generalize_type_block_access_result
+        (fun () ->
+           let res = type_block_access expected_base_ty principal ba in
+           (try unify env res.base_ty expected_base_ty with Unify _ -> ());
```
ccasin, line 6475: This is surprising to me - why are we throwing away unification errors here?

typing/typecore.ml
```diff
@@ -6232,6 +6391,74 @@ and type_expect_
...
+      | Baccess_array { mut = Immutable; _ } | Baccess_block (Immutable, _) ->
+        false
+      | Baccess_field (_, { lbl_mut = Mutable _; _ })
+      | Baccess_array { mut = Mutable; _ } | Baccess_block (Mutable, _) ->
+        true
+    in
+    let (el_ty, modality), uas =
+      List.fold_left_map
+        (fun (el_ty, modality) ua ->
+           with_local_level_if_principal
```
ccasin, line 6492: It's tempting to suggest we should skip all these intermediate level adjustments and generalization steps, just putting the bulk of this case in one `with_local_level_if_principal` and `generalize_structure`ing its results at the end.  My guess is you've done it this way because the intermediate generalizations are necessary for accurate principality warnings on the label declarations? Maybe a comment, if so.

(It's further tempting to suggest doing the efficient thing if we're not in principal mode, but I think the perf benefit would be modest and outweighed by the additional code complexity)

typing/typecore.ml
```diff
@@ -5691,6 +5733,123 @@ and type_expect_
...
+      { ba; base_ty; el_ty; flat_float = false; modality }
+  in
+  let type_unboxed_access el_ty ua =
+    match ua with
+    | Parsetree.Uaccess_unboxed_field lid ->
+      let expected_record_type =
+        match extract_concrete_typedecl_protected env el_ty with
+        | Typedecl(p0, p, {type_kind=Type_record_unboxed_product _}) ->
+          (* we treat this as principal because [ty_expected] only
+              affects disambiguation for the block access *)
```
ccasin: I don't understand this comment - can you explain a little more?  (In person is fine, then I'll probably suggest updating it)

lambda/lambda.mli
```diff
@@ -149,6 +149,14 @@ type primitive =
   | Punboxed_product_field of int * (layout list)
       (* the [layout list] is the layout of the whole product *)
   | Parray_element_size_in_bytes of array_kind
+  (* Block indices *)
+  | Pidx_field of int
+  | Pidx_mixed_field of mixed_block_shape * int * int list
+    (** The lone int is the index into the mixed_block_shape, the int list is
+        the path into that mixed_block_element *)
+  | Pidx_array of
+      array_kind * array_index_kind * unit mixed_block_element * int list
```
ccasin, line 162: I am tempted to suggest that `Pidx_field`, and `Pidx_mixed_field` should be constants, not primitives, since they are just numbers that are fully determined at compile time.  But I guess that's not strictly true for bytecode, where we might need to allocate a block?  Though I guess even then the block could conceivably be statically allocated.

Maybe this distinction isn't so important - I'll leave it up to you. I also wonder, if they stay primitives, if names like `Pmake_idx_field` would be more appropriate (the idea being that the primitive operation is to make a block index, and it just so happens that the compiler can sometimes optimize this into a static number).

lambda/lambda.mli
```diff
@@ -1215,6 +1225,14 @@ val mod_setfield: int -> primitive
 
 val structured_constant_layout : structured_constant -> layout
 
+val mixed_block_element_of_layout : layout -> unit mixed_block_element
+
+val mixed_block_element_leaves
+  : 'a mixed_block_element -> 'a mixed_block_element list
+
+(** Whether these exists a non-value before a value *)
```
ccasin:
```suggestion
(** Whether there exists a non-value before a value *)
```

lambda/lambda.ml
```diff
@@ -2087,7 +2095,13 @@ let primitive_may_allocate : primitive -> locality_mode option = function
   | Pdls_get
   | Preinterpret_unboxed_int64_as_tagged_int63
   | Parray_element_size_in_bytes _
-  | Ppeek _ | Ppoke _ -> None
+  | Pidx_field _
+  | Pidx_mixed_field _
+  | Pidx_array _
+  | Pidx_deepen _
+  | Pget_idx _ | Pset_idx _
```
ccasin, line 2048: I think some of these can allocate on bytecode.

jane/doc/extensions/_03-unboxed-types/block-indices.md
```diff
@@ -0,0 +1,215 @@
+---
```
ccasin: To arrange things so that the `intro` page still shows up first, we're going to have to number the files in this directory now - see comprehensions as an example.  Probably some links will then need to be updated too.  Sorry, the website infrastructure could use some work.

(Or you can come up with a title later in the alphabet :)

lambda/mixed_product_bytes.ml
```diff
@@ -0,0 +1,63 @@
...
+let zero = { value = 0; flat = 0 }
+
+let add { value; flat } { value = value'; flat = flat' } =
+  { value = value + value'; flat = flat + flat' }
+
+let rec count (el : _ Lambda.mixed_block_element) : t =
+  match el with
+  | Value _ -> { value = 8; flat = 0 }
+  | Float_boxed _ | Float64 | Float32 | Bits32 | Bits64 | Word ->
+    (* In a record, [float32]s and [bits32]s aren't packed tightly *)
```
ccasin: This could change - we should find a way to create a loud failure when it does.  That could be as simple as a compiler test in one of the block indices folders that checks the size of some allocated records.

lambda/mixed_product_bytes_wrt_path.ml
```diff
@@ -0,0 +1,104 @@
+(******************************************************************************
```
ccasin: These two new modules are pretty small, and this one is only used in one place. Maybe this can just be a submodule of the other?  Possibly that also allows exposing a slightly smaller interface from the other module, as I think some things are only used here.

lambda/mixed_product_bytes.mli
```diff
@@ -0,0 +1,55 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,   *
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL    *
+ * THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER *
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING    *
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER        *
+ * DEALINGS IN THE SOFTWARE.                                                  *
+ ******************************************************************************)
+
+(** Counts the number of bytes belonging to values and flat (non-value) of a
+    mixed block element *)
```
ccasin: I think it would be useful for readers to note that this module is used for compiling block indices (I would even be tempted to changes its name to reflect that, to better distinguish it from `mixed_block_shape.ml`, but I don't feel strongly).

testsuite/tests/typing-layouts-block-indices/basics_block_indices.ml
```diff
@@ -0,0 +1,852 @@
...
+(* A gap of 2^16 bytes is not allowed *)
+let bad_idx () = (.si)
+[%%expect{|
+Line 1, characters 17-22:
+1 | let bad_idx () = (.si)
+                     ^^^^^
+Error: Block indices into records that contain both values and non-values,
+       and occupy over 2^16 bytes, cannot be created.
```
ccasin: I think it's important for error messages to be accurate, even if the inaccuracy is subtle - maybe this could say: "This block index can not be created because it refers to values and non-values that are separated by more than 2^16 bytes in their block, or could be deepened to such an index" (or feel free to reword to your taste).

lambda/mixed_product_bytes_wrt_path.mli
```diff
@@ -0,0 +1,70 @@
...
+
+val all : t -> Mixed_product_bytes.t
+
+type offset_and_gap_bytes =
+  { offset_bytes : Mixed_product_bytes.Byte_count.t;
+    gap_bytes : Mixed_product_bytes.Byte_count.t
+  }
+
+(** Compute the offset and gap in bytes for an index to [here]. Returns [None]
+    if the index could lead an "illegal" gap of 2^16 or greater bytes, which
```
ccasin:
```suggestion
    if the index could lead to an "illegal" gap of 2^16 or greater bytes, which
```

lambda/translcore.ml
```diff
@@ -2201,6 +2205,89 @@ and transl_record_unboxed_product ~scopes loc env fields repres opt_init_expr =
       let exp = transl_exp ~scopes init_expr_sort init_expr in
       Llet(Strict, layout, init_id, init_id_duid, exp, lam)
 
+(* See [jane/doc/extensions/_02-unboxed-types/block-indices.md] *)
+and transl_idx ~scopes loc env ba uas =  (*  *)
```
ccasin:
```suggestion
and transl_idx ~scopes loc env ba uas =
```

lambda/translcore.ml
```diff
@@ -2201,6 +2205,89 @@ and transl_record_unboxed_product ~scopes loc env fields repres opt_init_expr =
...
+      (* Preserve the invariant that products have at least two elements *)
+      let base_sort =
+        if Int.equal (Array.length lbl.lbl_all) 1 then
+          lbl.lbl_sort
+        else
+          Jkind.Sort.Const.Product
+            (Array.to_list (Array.map (fun lbl -> lbl.lbl_sort) lbl.lbl_all))
+      in
+      let base_layout = layout env lbl.lbl_loc base_sort lbl.lbl_res in
+      let mbe = mixed_block_element_of_layout base_layout in
```
ccasin, line 2341: The approach here makes sense and seems correct.  Something about it does feel a little funny to me, though - like we're rebuilding something pretty similar to the machinery that already exists for mixed block compilation.  I wonder if, rather than computing the `layout` based on `lbl_res`, we can just look at the declaration of the record and see if it's a `Record_boxed`, in which case we have all values, or a `Record_mixed`, in which case we have a `mixed_block_shape` available and can use that rather than computing a `Lambda.layout` and calling `mixed_block_element_of_layout` on it. I'm also concerned that this approach ends up computing a bunch of `value_kind`s that we're going to end up throwing away (for example, if we deepen past them), and `value_kind` is expensive.

(Happy to discuss in person if useful)

lambda/translcore.ml
```diff
@@ -2201,6 +2205,89 @@ and transl_record_unboxed_product ~scopes loc env fields repres opt_init_expr =
...
+      let shape =
+        Lambda.transl_mixed_product_shape
+          ~get_value_kind:(fun _ -> Lambda.generic_value) shape
+      in
+      (* Check to make sure the gap never overflows.
+         See [jane/doc/extensions/_02-unboxed-types/block-indices.md]. *)
+      let cts =
+        Mixed_product_bytes_wrt_path.count_shape shape lbl.lbl_pos uas_path in
+      if Option.is_none
+          (Mixed_product_bytes_wrt_path.offset_and_gap cts) then
```
ccasin: Two tiny style nits, I think the usual thing is to put `in` on its own line if it does not fit on the same line as `let`, and to put `then` at the beginning of a line if it does not fit on the same line as `if`.

lambda/translprim.ml
```diff
@@ -380,13 +380,15 @@ let lookup_primitive loc ~poly_mode ~poly_sort pos p =
...
-       the middle-end. *)
+       [prim_has_valid_reprs] in [typing/primitive.ml]. *)
     let (_, repr) = lambda_prim.prim_native_repr_res in
     Lambda.layout_of_extern_repr repr
   in
+  let arg_layouts =
+    List.map
+      (fun (_, repr) -> Lambda.layout_of_extern_repr repr)
+      lambda_prim.prim_native_repr_args
+  in
```
ccasin, line 582: It feels slightly wasteful to build this list when only one element of it is used and in only one case (if I understand correctly).  Consider moving the computation of the needed layout to the case where it's used.

lambda/translprim.ml
```diff
@@ -1968,6 +1981,7 @@ let lambda_primitive_needs_event_after = function
   | Pdls_get
   | Pobj_magic _ | Punbox_float _ | Punbox_int _ | Punbox_vector _
   | Preinterpret_unboxed_int64_as_tagged_int63 | Ppeek _ | Ppoke _
+  | Pget_idx _ | Pset_idx _
```
ccasin: I think these are analagous to the unsafe array operations - they are compiled to C calls that have camlasserts in them, on bytecode - so we should return `true` for them.

testsuite/tests/typing-layouts-block-indices/basics_block_indices.ml
rtjoa: Good catch. Changing the test to the following, and now the annotation does change the error given.
```ocaml
type y = { y : int }
type 'a t = { a : 'a }
let bad c = if c then
    ((.a) : (y# t, _) idx_imm)
  else
    (.a.#a)
```

jane/doc/extensions/_03-unboxed-types/block-indices.md
rtjoa: Makes sense, swapped with the other use case

jane/doc/extensions/_03-unboxed-types/block-indices.md
rtjoa: Thanks, this reads much better

testsuite/tests/records-and-block-indices/generated_array_idx_access_bytecode_test.ml
rtjoa: Updated to be
```ocaml
(* We update [a.(i)] at all paths of depth 2 to be the corresponding item *)
(* in [next_el].*)
```

parsing/parsetree.mli
rtjoa: `(BA1 UA1 UA2 ...)` was an attempt write the concrete syntax, though I see how looks a bit out of place as we don't use the metavariables `BA` and `UA` elsewhere (this prompted me to add the example). (Is there a better way to write the concrete syntax?)

testsuite/tests/records-and-block-indices/generated_array_idx_access_bytecode_test.ml
rtjoa: Thanks for catching this! At the time of writing, we didn't support vec128 arrays; we do now so I updated `Metaprogramming.array_element` and they now differ.

jane/doc/extensions/_03-unboxed-types/block-indices.md
rtjoa: Added the following (mixed blocks cannot be marshaled on either bytecode or native):
```
While marshaling data should be consistent between bytecode and native, this
case is safe because mixed blocks (the only `value`s that can contain `bits64`s)
cannot be marshaled.
```

testsuite/tests/typing-layouts-block-indices/basics_block_indices.ml
rtjoa: Good point - filing this.

testsuite/tests/typing-layouts-block-indices/basics_block_indices.ml
rtjoa: Renamed to `box` and `box_mut`.

lambda/lambda.ml
rtjoa: Joined the index creation primitives into the below case:
```ocaml
  | Pidx_field _
  | Pidx_mixed_field _
  | Pidx_array _
  | Pidx_deepen _
  | Preinterpret_tagged_int63_as_unboxed_int64 ->
    if !Clflags.native_code then None
    else
      (* We don't provide a locally-allocating version of this primitive
         since it would only apply to bytecode, and code requiring performance
         at a level where these primitives are necessary is very likely going
         to be native. *)
      Some alloc_heap
```

jane/doc/extensions/_03-unboxed-types/block-indices.md
rtjoa: Numbered these pages

lambda/mixed_product_bytes.ml
rtjoa: This should already be tested by `testsuite/tests/records-and-block-indices/generated_record_size_native_test.ml` and `testsuite/tests/records-and-block-indices/generated_record_size_bytecode_test.ml`.

And I added a comment in `Metaprogramming.Type_structure.size` that this is coupled.

lambda/mixed_product_bytes_wrt_path.ml
rtjoa: Inlined that module

lambda/mixed_product_bytes.mli
rtjoa: I couldn't think of a nice name that succinctly mentioned block indices (maybe `Mixed_product_idx_byte_offsets`? but this is a bit verbose, and I think the current name is a clearer description for its weight), so I updated the comment to state "for the purpose of compiling block indices."

lambda/translprim.ml
rtjoa: Changed this to a thunk (`get_arg_layouts ()`)

testsuite/tests/typing-layouts-block-indices/basics_block_indices.ml
rtjoa: Added a CR above

testsuite/tests/typing-layouts-block-indices/basics_block_indices.ml
rtjoa: Added a CR here

testsuite/tests/typing-layouts-block-indices/basics_block_indices.ml
rtjoa: Added this test and another CR

testsuite/tests/typing-layouts-block-indices/basics_block_indices.ml
rtjoa: As index get and set are "just externals," I think the compiler check is just of the signatures given to those externals, so there isn't new behavior to test. (But lmk if you disagree)

lambda/translcore.ml
rtjoa: (Notes for in-person discussion: we might not be deepening from a boxed record, and might not have a `record_representation`)

typing/typecore.ml
rtjoa: Pulled these out

lambda/lambda.mli
rtjoa: Renamed these:
```
Pidx_field -> Pmake_idx_field
Pidx_mixed_field -> Pmake_idx_mixed_field
Pidx_array -> Pmake_idx_array
```

typing/jkind.mli
rtjoa: Changed to "exactly the sort [value]" (note there are two very similar functions above that say [value], so avoiding diverging too much)

testsuite/tests/records-and-block-indices/do_test_generation.ml
rtjoa: I reduced the number of block indices tests (this change brought real time from 1m26->1m3s for me), and propose changing existing tests in a follow-up PR.

typing/typemode.ml
rtjoa: Agreed, though I think the cost here won't be noticeable, so I added a CR for now.

middle_end/flambda2/parser/flambda_to_fexpr.ml
rtjoa: Sounds good - I'll do this in a follow-up PR.

typing/typecore.ml
rtjoa: Added this comment:
```ocaml
(* This unification is to get a better [base_ty], and is not
   necessary for safety (thus we discard the failure). For example, if the
   initial expected base type is [t array], this makes the next base type
   [t] (which may be critical to disambiguate the unboxed accesses). *)
```

typing/typecore.ml
rtjoa: This is an outdated/wrong comment (below, we used to pass `true` instead of `is_principal el_ty`) - deleting.

For fun, this is the test case that would be wrong if we always pass `true`:
```ocaml
type u = #{ x : int }
type u2 = #{ x : string }

(* First unboxed index disambiguated non-principally through the expected array
   type *)
let f c =
  if c then
    ((.(0).#x) : (u array, _) idx_mut)
  else
    (.(1).#x)
[%%expect{|
val f : bool -> (u array, int) idx_mut = <fun>
|}, Principal{|
Line 5, characters 11-12:
5 |     (.(1).#x)
               ^
Warning 18 [not-principal]: this type-based unboxed record field disambiguation is not principal.

val f : bool -> (u array, int) idx_mut = <fun>
|}]
```

typing/typecore.ml
rtjoa: Yep, added such a comment.

lambda/translcore.ml
rtjoa: Per discussion, decided to leave a CR that we may be able to compute a mixed block element from the sort instead of the layout, as we don't need the value_kind

parsing/parsetree.mli
rtjoa: (updated comment)

mshinwell (approved):

PR #4032, opened 2025-05-21
Use affine instead of linear

cuihtlauac:


goldfirere:
This is actually something we've debated on and off, and we've always ended up concluding "just keep 'linearity'". If we make this change, we should also change the codebase, I think. Maybe we bring this up in the coordination meeting tomorrow?

PR #4044, opened 2025-05-27
Add an OCaml equivalent of `Assert_mixed_block_layout_v#`

ccasin:
See documentation changes for an explanation.  People are increasingly using `Obj.magic` to depend on the representation of mixed blocks, so this feels prudent to do now.


dkalinichenko-js (approved):
LGTM. It'll be hard to get people to use it, but I don't see better solutions.

You should also change `lib/janestreeet_shims` in the tree to export this value. There are tests I can point you to.

PR #4046, opened 2025-05-27
Fix yet another Infix_tag bug, this one in oldify_one.

NickBarnes:
When we promote a shared closure, before we update the header of the "from" block, we must copy across any infix header words. Otherwise, anyone oldifying an infix pointer to the same shared closure will create a pointer on the heap which doesn't (yet) have a header word.
On more than one occasion this caused SEGVs in multi-domain application code, when a domain was doing opportunistic marking while waiting for other domains to finish the minor GC. Marking such a pointer caused a bogus range to be pushed to the mark stack, including ASCII words which are then treated as pointers to mark, which crashed. Although we only saw it in marking, possibly there are other times when such a bogus pointer might be used (it's potentially disastrous at any point until the domain which promotes the shared closure completes its minor GC).
As it happens, this was only seen in a minor GC when `mark_requested` was set. It's unclear whether that was essential to the SEGV, or makes it more likely, or is just coincidence.

### DIFF
--- a/runtime/minor_gc.c
+++ b/runtime/minor_gc.c
@@ -313,6 +313,23 @@ static void oldify_one (void* st_v, value v, volatile value *p)
     st->live_bytes += Bhsize_hd(hd);
     result = alloc_shared(st->domain, sz, tag, Reserved_hd(hd));
     field0 = Field(v, 0);
+    if (tag == Closure_tag) {
+      /* Copy unscannable prefix, which will include all infix tags,
+       * so that any domain can oldify infix pointers to `v` into
+       * pointers which also have Infix_tag to a working header (so
+       * that, e.g., major GC marking can work while the block is
+       * still on our oldify todo list). Have to do this here, before
+       * we update the object header.
+       */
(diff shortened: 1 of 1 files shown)


runtime/minor_gc.c
```diff
@@ -313,6 +313,23 @@ static void oldify_one (void* st_v, value v, volatile value *p)
     st->live_bytes += Bhsize_hd(hd);
     result = alloc_shared(st->domain, sz, tag, Reserved_hd(hd));
     field0 = Field(v, 0);
+    if (tag == Closure_tag) {
+      /* Copy unscannable prefix, which will include all infix tags,
+       * so that any domain can oldify infix pointers to `v` into
+       * pointers which also have Infix_tag to a working header (so
+       * that, e.g., major GC marking can work while the block is
+       * still on our oldify todo list). Have to do this here, before
+       * we update the object header.
```
stedolan, line 322: This looks good but I don't understand the last sentence. Why is there an ordering requirement with the object header?

stedolan (approved):

PR #4048, opened 2025-05-28
Linscan: use skip lists

xclerc:
(Only for testing/benchmarking at this point.)

### DIFF
--- a/backend/amd64/reg_class.ml
+++ b/backend/amd64/reg_class.ml
@@ -94,10 +94,12 @@ module T = struct
 
   let hash : t -> int = function Int64 -> 0 | Float128 -> 1
 
+  let to_string : t -> string = function
+    | Int64 -> "int64"
+    | Float128 -> "float128"
+
   let print : Format.formatter -> t -> unit =
-   fun ppf reg_class ->
-    Format.fprintf ppf "%s"
-      (match reg_class with Int64 -> "int64" | Float128 -> "float128")
+   fun ppf reg_class -> Format.fprintf ppf "%s" (to_string reg_class)
 end
 
 include T
--- a/backend/arm64/reg_class.ml
+++ b/backend/arm64/reg_class.ml
@@ -91,10 +91,12 @@ module T = struct
 
   let hash : t -> int = function Int64 -> 0 | Float128 -> 1
 
+  let to_string : t -> string = function
+    | Int64 -> "int64"
+    | Float128 -> "float128"
+
   let print : Format.formatter -> t -> unit =
-   fun ppf reg_class ->
-    Format.fprintf ppf "%s"
-      (match reg_class with Int64 -> "int64" | Float128 -> "float128")
+   fun ppf reg_class -> Format.fprintf ppf "%s" (to_string reg_class)
 end
 
 include T
--- a/backend/regalloc/regalloc_ls.ml
+++ b/backend/regalloc/regalloc_ls.ml
@@ -145,6 +145,8 @@ let allocate_free_register : State.t -> Interval.t -> spilling_reg =
         | Stack _ | Unknown -> ()
       in
       DLL.iter intervals.active_dll ~f:set_not_available_if_valid_phys_reg;
+      Interval.DescEndList.iter intervals.active_sl
+        ~f:set_not_available_if_valid_phys_reg;
       let remove_bound_overlapping (itv : Interval.t) : unit =
         match itv.reg.loc with
         | Reg r ->
@@ -156,6 +158,10 @@ let allocate_free_register : State.t -> Interval.t -> spilling_reg =
       in
       DLL.iter intervals.inactive_dll ~f:remove_bound_overlapping;
       DLL.iter intervals.fixed_dll ~f:remove_bound_overlapping;
+      (* note: already removed, for transition *)
+      Interval.DescEndList.iter intervals.inactive_sl
+        ~f:remove_bound_overlapping;
+      Interval.DescEndList.iter intervals.fixed_sl ~f:remove_bound_overlapping;
       let rec assign idx =
         if idx >= num_available_registers
         then Misc.fatal_error "No_free_register should have been raised earlier"
@@ -163,6 +169,7 @@ let allocate_free_register : State.t -> Interval.t -> spilling_reg =
         then (
           reg.loc <- Reg (first_available + idx);
           Interval.DLL.insert_sorted intervals.active_dll interval;
+          Interval.DescEndList.insert intervals.active_sl interval;
           if debug
           then (
             indent ();
@@ -179,25 +186,54 @@ let allocate_blocked_register : State.t -> Interval.t -> spilling_reg =
   let reg = interval.reg in
   let reg_class = Reg_class.of_machtype reg.typ in
   let intervals = State.active state ~reg_class in
+  (* will change to `match Interval.DescEndList.hd intervals.active_sl with` *)
   match DLL.hd_cell intervals.active_dll with
   | Some hd_cell ->
+    let hd_cell' = Interval.DescEndList.hd_cell intervals.active_sl in
+    if Option.is_none hd_cell'
+    then fatal "Regalloc_ls.allocate_blocked_register: missing skip list head";
+    let hd_cell' = Option.get hd_cell' in
     let hd = DLL.value hd_cell in
+    let hd' = Interval.DescEndList.value hd_cell' in
+    if not (Interval.equal hd hd')
+    then
+      fatal
+        "Regalloc_ls.allocate_blocked_register: inconsistent heads (dll=%a vs \
+         sl=%a)"
+        Interval.print hd Interval.print hd';
     let chk r =
       assert (same_reg_class r.Interval.reg hd.Interval.reg);
       Reg.same_loc r.Interval.reg hd.Interval.reg && Interval.overlap r interval
     in
-    if hd.end_ > interval.end_
-       && not
-            (DLL.exists ~f:chk intervals.fixed_dll
-            || DLL.exists ~f:chk intervals.inactive_dll)
+    let dll_exists =
+      DLL.exists ~f:chk intervals.fixed_dll
+      || DLL.exists ~f:chk intervals.inactive_dll
+    in
+    let sl_exists =
+      Interval.DescEndList.exists ~f:chk intervals.fixed_sl
+      || Interval.DescEndList.exists ~f:chk intervals.inactive_sl
+    in
+    if not (Bool.equal dll_exists sl_exists)
+    then
+      fatal
+        "Regalloc_ls.allocate_blocked_register: inconsistent lists \
+         (dll_exists=%B sl_exists=%B)"
+        dll_exists sl_exists;
+    if hd.end_ > interval.end_ && not dll_exists
     then (
       (match hd.reg.loc with Reg _ -> () | Stack _ | Unknown -> assert false);
       interval.reg.loc <- hd.reg.loc;
       DLL.delete_curr hd_cell;
       Interval.DLL.insert_sorted intervals.active_dll interval;
+      Interval.DescEndList.delete_curr hd_cell';
+      Interval.DescEndList.insert intervals.active_sl interval;
       allocate_stack_slot hd.reg)
     else allocate_stack_slot reg
-  | None -> allocate_stack_slot reg
+  | None ->
+    if Option.is_some (Interval.DescEndList.hd_cell intervals.active_sl)
+    then
+      fatal "Regalloc_ls.allocate_blocked_register: unexpected skip list head";
+    allocate_stack_slot reg
 
 let reg_reinit () =
   List.iter (Reg.all_relocatable_regs ()) ~f:(fun (reg : Reg.t) ->
@@ -218,7 +254,9 @@ let rec main : round:int -> State.t -> Cfg_with_infos.t -> unit =
     log "main, round #%d" round;
     indent ());
   reg_reinit ();
+  State.check_consistency state "main/start";
   build_intervals state cfg_with_infos;
+  State.check_consistency state "main/after build_intervals";
   State.invariant_intervals state cfg_with_infos;
   State.invariant_active state;
   if debug then snapshot_for_fatal := Some (State.for_fatal state);
@@ -229,18 +267,22 @@ let rec main : round:int -> State.t -> Cfg_with_infos.t -> unit =
   let spilled =
     State.fold_intervals state ~init:Reg.Set.empty
       ~f:(fun acc (interval : Interval.t) ->
+        State.check_consistency state "main/before release";
         (* Equivalent to [walk_interval] in "backend/linscan.ml".*)
         let pos = interval.begin_ land lnot 1 in
         if debug
         then log_interval ~kind:(Printf.sprintf "<pos=%d>" pos) interval;
         State.release_expired_intervals state ~pos;
+        State.check_consistency state "main/after release";
         let spilled =
           match allocate_free_register state interval with
-          | spilled -> spilled
+          | spilled ->
+            State.check_consistency state "main/after allocate_free_register";
+            spilled
           | exception No_free_register ->
             allocate_blocked_register state interval
         in
-        State.invariant_active state;
+        State.check_consistency state "main/after allocation";
         match spilled with
         | Not_spilling -> acc
         | Spilling reg -> Reg.Set.add reg acc)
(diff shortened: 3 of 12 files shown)


backend/amd64/reg_class.ml
```diff
@@ -94,10 +94,12 @@ module T = struct
...
 
+  let to_string : t -> string = function
+    | Int64 -> "int64"
+    | Float128 -> "float128"
+
   let print : Format.formatter -> t -> unit =
-   fun ppf reg_class ->
-    Format.fprintf ppf "%s"
-      (match reg_class with Int64 -> "int64" | Float128 -> "float128")
+   fun ppf reg_class -> Format.fprintf ppf "%s" (to_string reg_class)
```
spiessimon, line 102: This could be the following, but I don't think it matters in terms of performance.

```suggestion
   fun ppf reg_class -> Format.pp_print_string ppf (to_string reg_class)
```

backend/arm64/reg_class.ml
```diff
@@ -91,10 +91,12 @@ module T = struct
...
 
+  let to_string : t -> string = function
+    | Int64 -> "int64"
+    | Float128 -> "float128"
+
   let print : Format.formatter -> t -> unit =
-   fun ppf reg_class ->
-    Format.fprintf ppf "%s"
-      (match reg_class with Int64 -> "int64" | Float128 -> "float128")
+   fun ppf reg_class -> Format.fprintf ppf "%s" (to_string reg_class)
```
spiessimon, line 99: Same here:

```suggestion
   fun ppf reg_class -> Format.pp_print_string ppf (to_string reg_class)
```

utils/skip_list.ml
```diff
@@ -0,0 +1,345 @@
+[@@@ocaml.warning "+a-30-40-41-42"]
+
+[@@@ocaml.flambda_o3]
+
+(* CR-soon xclerc: add `open! Int_replace_polymorphic_compare`. *)
+
+(* CR xclerc for xclerc: refactor. *)
+let init_random : unit Lazy.t = lazy (Random.self_init ())
```
spiessimon, line 8: Would it make sense to always use the same seed here? In either case, the seed will have no effect on the output of the compiler, but it would make performance more deterministic. 

Moreover, this could be a separate instance of the RNG instead of one that shares state with (potentially) other parts of the compiler.

utils/skip_list.ml
```diff
@@ -0,0 +1,345 @@
...
+  (* CR-soon xclerc for xclerc: consider removing the `level` field. *)
+  type node =
+    { data : elem;
+      level : int;
+      prev : pillar;
+      next : pillar
+    }
+
+  and pillar = node option array
+
```
spiessimon, line 68: I tried to summarize the invariant that I understood that this implementation maintains below. Please take a look whether I misunderstood something. (It breaks the 80ch rule at the moment, sorry.)

```suggestion
  (* Invariant: A pillar must always be non-empty. A node at level `n` has pillars of length `n + 1`. To indicate that there are no successors (e.g., for the last element of the list), this array stores `None` instead of being empty. The immediate successor/predecessor is always at index 0. *)
```

utils/skip_list.ml
```diff
@@ -0,0 +1,345 @@
...
+  val next : cell -> cell option
+
+  val cut : cell -> unit
+
+  val delete_curr : cell -> unit
+
+  type t
+
+  val make_empty : max_skip_level:int -> skip_factor:float -> unit -> t
+
```
spiessimon, line 32:
```suggestion
  (** Create a new skip list. `max_skip_level` must be at least 1, and `skip_factor` must be between 0 and 1. The `max_skip_level` determines the maximal number of forward (resp. backward) references in a node besides the immediate forward/backward reference, and the `skip_factor` determines the skip level of intermediate nodes: the skip level of these nodes is chosen randomly with distribution `P(n) = skip_factor^n * (1 - skip_factor)`, except for the last one, where it is `P(max_skip_level) = skip_factor^max_skip_level`. *)

```

utils/skip_list.ml
```diff
@@ -0,0 +1,345 @@
...
+  type elem
+
+  type cell
+
+  val value : cell -> elem
+
+  val next : cell -> cell option
+
+  val cut : cell -> unit
+
```
spiessimon, line 26:
```suggestion
  (** Removes everything after the given cell. *)
```

utils/skip_list.ml
```diff
@@ -0,0 +1,345 @@
...
+    match cell.node.prev.(0) with
+    | None ->
+      for level = 0 to pred (Array.length cell.t.head) do
+        cell.t.head.(level) <- None
+      done
+    | Some node ->
+      let curr_node : node ref = ref node in
+      let curr_pillar : pillar ref = ref node.next in
+      for level = 0 to cell.t.max_skip_level do
+        while level >= Array.length !curr_pillar do
```
spiessimon, line 102: If I understand this code correctly, this is the case for "we want to cut after a cell of level n and must also remove all forward references of cells >n". It took me a while to figure this out. Perhaps we could leave a comment here to explain that. 

Also, if this is the case, I believe it would be correct to travel back at level `Array.length !curr_pillar - 1` instead of `0`, which might be a bit faster.

utils/skip_list.ml
```diff
@@ -0,0 +1,345 @@
...
+
+  let delete_curr : cell -> unit =
+   fun cell ->
+    let node = cell.node in
+    for level = 0 to pred (Array.length node.prev) do
+      match node.prev.(level) with
+      | None -> cell.t.head.(level) <- node.next.(level)
+      | Some prev -> prev.next.(level) <- node.next.(level)
+    done;
+    let level = ref 0 in
```
spiessimon, line 120: I wonder whether this code would be a little easier to read if the second loop would simply mirror the first one. "Aborting early" in the case that there are no more elements coming is of course a little more efficient. Or the similarity and subtle different could be highlighted in a comment.

utils/skip_list.ml
```diff
@@ -0,0 +1,345 @@
...
+      (* previous node for each level *)
+      Array.make (succ t.max_skip_level) None
+    in
+    let update : pillar array =
+      (* pillar to update for each level *)
+      Array.make (succ t.max_skip_level) t.head
+    in
+    let curr : pillar ref = ref t.head in
+    let prev : node option ref = ref None in
+    for i = t.max_skip_level downto 0 do
```
spiessimon, line 187:
```suggestion
    (* For each level, we search for the right spot to insert a new entry; remember the predecessor and successor  (pillar) at this level, and then descend one level lower. *)
    for i = t.max_skip_level downto 0 do
```

utils/skip_list.ml
```diff
@@ -0,0 +1,345 @@
...
+        level;
+        prev = Array.init (succ level) (fun i -> prev_nodes.(i));
+        next = Array.init (succ level) (fun i -> update.(i).(i))
+      }
+    in
+    for i = 0 to level do
+      (match update.(i).(i) with
+      | None -> ()
+      | Some n -> n.prev.(i) <- Some node);
+      update.(i).(i) <- Some node
```
spiessimon, line 207: I don't understand what this update is updating. Are these, effectively, the `next` pillars of all of the predecessors, which now need to point to the new node? This is a bit subtle.

utils/skip_list.ml
```diff
@@ -0,0 +1,345 @@
...
+    let rec aux (node : node option) f =
+      match node with
+      | None -> false
+      | Some node -> f node.data || aux node.next.(0) f
+    in
+    fun t ~f -> aux t.head.(0) f
+
+  let to_list : t -> elem list =
+   fun t ->
+    (* CR xclerc for xclerc: avoid the `List.rev`. *)
```
spiessimon, line 249: Perhaps it would make sense to use some of the other iteration operations above, such as `fold_left` with cons and then a reversal. Or, to avoid the reversal, one could implement `fold_right` as (1) find the last element by traversing the list from the highest level down and (2) traversing the list backwards using `prev.(0)` to find the previous element until the head is reached.

spiessimon (commented):
I've looked mostly at the implementation of the skip lists. Will look at the other parts of this PR a bit later. The implementation of the skip lists is quite subtle. I didn't spot any bugs, but I think thorough testing would make sense, especially for things like: after a bunch of insertions, cuts, removals, etc. are all of the invariants still maintained and does a forward traversal and a backward traversal yield the same elements in the correct order.

backend/regalloc/regalloc_ls_utils.ml
```diff
@@ -286,10 +392,73 @@ module ClassIntervals = struct
       aux t ~pos (DLL.hd_cell l)
   end
 
+  module SL = struct
+    let release_expired_fixed (l : Interval.DescEndList.t) ~pos =
```
spiessimon, line 396: Why does this function not have a counter part in the DLL module?

utils/skip_list.mli
```diff
@@ -0,0 +1,49 @@
...
+module type T = sig
+  type elem
+
+  type cell
+
+  val value : cell -> elem
+
+  val next : cell -> cell option
+
+  val cut : cell -> unit
```
spiessimon, line 18: This operation is called `cut_from` for doubly linked lists. I guess either is fine. A comment here or in the implementation could be a good idea to describe what the operation does.

backend/regalloc/regalloc_ls_utils.ml
```diff
@@ -286,10 +392,73 @@ module ClassIntervals = struct
...
+          let value = Interval.DescEndList.value cell in
+          if value.Interval.end_ >= pos
+          then (
+            Interval.remove_expired value ~pos;
+            aux (Interval.DescEndList.next cell) ~pos)
+          else Interval.DescEndList.cut cell
+      in
+      aux (Interval.DescEndList.hd_cell l) ~pos
+
+    let release_expired_active (t : t) ~(pos : int) (l : Interval.DescEndList.t)
```
spiessimon, line 410: Would it make sense to (a) remove the argument `l`, since it seems to be `t.active_sl`; or (b) make the inactive list an argument? It took me a minute to see that when we insert the value in `t.inactive_sl` and then remove it from `l`, these two cannot be the same list. (I was wondering whether this could lead to an finite loop.) I guess the DLL implementation does the same, so perhaps not in this PR.

utils/skip_list.ml
```diff
@@ -0,0 +1,345 @@
...
+    done
+
+  let hd_cell : t -> cell option =
+   fun t -> match t.head.(0) with None -> None | Some node -> Some { node; t }
+
+  let should_advance (i : int) (node : node option array) (value : elem) : bool
+      =
+    match node.(i) with
+    | None -> false
+    | Some (node : node) -> OT.compare node.data value < 0
```
spiessimon, line 167: For skip lists over intervals descending by the end, insertion for equal `.end_` values will order the two based on Reg.compare---without considering the `.start_` value. Is this intentional/correct?

backend/regalloc/regalloc_ls_state.ml
```diff
@@ -53,51 +81,49 @@ let add_class_interval_array ci x =
...
   Reg.Tbl.iter
     (fun reg interval ->
       match reg.loc with
       | Reg _ ->
         let reg_class = Reg_class.of_machtype reg.typ in
-        add_class_interval_array (Reg_class.Tbl.find active' reg_class) interval
-      | Stack _ | Unknown -> add_class_interval_array class_intervals interval)
+        add_class_interval_array (Reg_class.Tbl.find active' reg_class) interval;
+        Interval.DescEndList.insert
+          (Reg_class.Tbl.find state.active reg_class).fixed_sl interval
```
spiessimon, line 101: Just to double check, this step corresponds to the following iteration, right?
```
Reg_class.Tbl.iter active'
    ~f:(fun reg_class (intervals : class_interval_array) ->
      let extracted = extract_class_interval_array intervals in
      Array.sort extracted ~cmp:compare_desc_end;
      DLL.clear (Reg_class.Tbl.find active reg_class).fixed_dll;
      DLL.add_array (Reg_class.Tbl.find active reg_class).fixed_dll extracted)
 ```
If so, maintaining sorted lists is quite nice and will make the code considerably easier to understand.

File: backend/regalloc/regalloc_ls.ml [hunk 17]
```diff
@@ -145,6 +145,8 @@ let allocate_free_register : State.t -> Interval.t -> spilling_reg =
         | Stack _ | Unknown -> ()
       in
       DLL.iter intervals.active_dll ~f:set_not_available_if_valid_phys_reg;
+      Interval.DescEndList.iter intervals.active_sl
+        ~f:set_not_available_if_valid_phys_reg;
       let remove_bound_overlapping (itv : Interval.t) : unit =
```
spiessimon, line 150: As in the previous PR that added doubly linked lists, I think this does not actually do the same as the first iteration (since that one will already set available registers to false), but it also does not hurt. Perhaps it would already be a good idea to comment out the old iteration to see that things are working as expected.

backend/regalloc/regalloc_ls.ml
```diff
@@ -156,13 +158,18 @@ let allocate_free_register : State.t -> Interval.t -> spilling_reg =
       in
       DLL.iter intervals.inactive_dll ~f:remove_bound_overlapping;
       DLL.iter intervals.fixed_dll ~f:remove_bound_overlapping;
+      (* note: already removed, for transition *)
+      Interval.DescEndList.iter intervals.inactive_sl
```
spiessimon, line 162: Same here.

utils/skip_list.ml
spiessimon: We basically always access `update.(i).(i)`. I wonder whether that is the same as `.next.(i)` of `prev_nodes.(i)` (where `t.head.(i)` is used in the `None` case). Considering that the current version works, I guess it doesn't matter, but I still don't have a good intuition for the `update` array.

utils/skip_list.mli
xclerc: I will rename it; even though doubly-linked
lists and skip lists are not drop-in replacements
for each other, I meant to use the same names.

backend/regalloc/regalloc_ls_utils.ml
xclerc: The function was misplaced, in `Interval.DLL`:
https://github.com/oxcaml/oxcaml/blob/529d5438b47bbef8f03ab778387c551509336b00/backend/regalloc/regalloc_ls_utils.ml#L202

utils/skip_list.ml
xclerc: I am not sure about using a seed, but
indeed using a separate source of
randomness is what I meant by "refactor".

utils/skip_list.ml
xclerc: I think it is correct to only use `end_`,
and `Reg.compare` is being used only
to be sure both implementations have
the intervals/ranges in the very same
order so that they can be compared.

backend/regalloc/regalloc_ls.ml
xclerc: Yes, that's right; like in the other PR,
I didn't trust myself so preferred to
have a redundant operation in the
first pull request of the chain.

utils/skip_list.ml
xclerc: I will copy/paste the docstring from
the doubly-linked list implementation
(note that it also deletes the passed
cell).

backend/regalloc/regalloc_ls_utils.ml
xclerc: Indeed, I suspect it took this form
initially to match upstream's
implementation, but it would be
clearer to only pass `t` to all three
functions in the module. I will leave
a CR-soon to address in a future
pull request.

backend/regalloc/regalloc_ls_state.ml
xclerc: Yes, we were previously populating a
dynamic array, sorting it and then copying
it to a list after clearing it...

backend/amd64/reg_class.ml
xclerc: (I don't think we would gain much
indeed, and I slightly prefer the
format string, as it makes changes
easier.)

backend/arm64/reg_class.ml
spiessimon: I guess the same applies as above here.

backend/amd64/reg_class.ml
spiessimon: Fair point :)

backend/regalloc/regalloc_ls_utils.ml
spiessimon: Ah, I see. I've now compared it with the old implementation and the two seem to match up. Thanks for clarifying.

backend/regalloc/regalloc_ls.ml
spiessimon: Perhaps it would make sense to add a consistency check here to ensure that the two lists store the same intervals (since the first iteration "overshadows" the second iteration).

PR #4056, opened 2025-05-28
Fix product layouts and shallow axes

dkalinichenko-js:
Make product kinds be always `non_null` and `non_float`. Separability doesn't make sense for product kinds, as we don't apply the float array optimization for arrays of product kinds. Nullability could make sense, but if we decide to implement that, we could look at the jkinds of the product components directly.

See tests. Note that the behavior of one-element unboxed records is unchanged, see the full file at `testsuite/tests/typing-layouts-or-null/separability.ml`.

### DIFF
--- a/testsuite/tests/typing-layouts-or-null/separability.ml
+++ b/testsuite/tests/typing-layouts-or-null/separability.ml
@@ -841,7 +841,7 @@ type a = #{ a : t_maybesep_val; }
 
 type b = #{ a : int; b: t_maybesep_val; c: float# }
 
-type ('b : value & value & float64 mod everything non_float) fails = unit constraint 'b = b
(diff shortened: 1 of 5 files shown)


testsuite/tests/typing-layouts-or-null/separability.ml
```diff
@@ -841,7 +841,7 @@ type a = #{ a : t_maybesep_val; }
 
 type b = #{ a : int; b: t_maybesep_val; c: float# }
 
-type ('b : value & value & float64 mod everything non_float) fails = unit constraint 'b = b
+type ('b : value & value & float64 mod non_null non_float) fails = unit constraint 'b = b
 
 (* CR layouts v3.4: separability of 2+ element unboxed records should always be
```
goldfirere, line 846: Delete this CR now?

goldfirere (approved):
🎉

PR #4062, opened 2025-05-29
`exn` crosses portable

riaqn:
This PR is a temporary workaround that allows `exn` to cross portability. To make it safe, exception constructors are assigned a portability, and are portable iff all its arguments are portable.

`exn` doesn't cross contention - I think that's not needed.


testsuite/tests/typing-modes/exn.ml
```diff
@@ -0,0 +1,196 @@
...
+
+let (foo @ portable) () =
+    match x with
+    | Contended _ -> ()
+    | _ -> ()
+[%%expect{|
+Line 3, characters 6-15:
+3 |     | Contended _ -> ()
+          ^^^^^^^^^
+Error: This value is "contended" but expected to be "uncontended".
```
riaqn, line 135: as discussed, this should be about portability, not contention.

typing/env.ml
```diff
@@ -269,23 +281,50 @@ module TycompTbl =
                 opened
               @ rest
 
+    let add_lock lock next =
+      { current = Ident.empty; layer = Lock {lock; next} }
+
+    let rec find_all_and_locks ~mark name tbl acc =
```
dkalinichenko-js, line 287: Maybe reuse this as the implementation of `find_all`? Also, I think `find_all_with_locks` would be a better name.

testsuite/tests/typing-modes/exn.ml
dkalinichenko-js, line 1: I'd add tests that all built-in exceptions that can be raised implicitly, such as `Assertion_failure` and `Match_failure`, are portable, since they always can be raised from portable functions.

testsuite/tests/typing-modes/exn.ml
```diff
@@ -0,0 +1,196 @@
...
+Error: This constructor is at mode "contended", but expected to be at mode "uncontended".
+|}, Principal{|
+Line 3, characters 31-40:
+3 |         exception Contended' = Contended
+                                   ^^^^^^^^^
+Error: The constructor "Contended" is nonportable, so cannot be used inside a function that is portable.
+|}]
+
+
+(* other extensible types are not affected *)
```
dkalinichenko-js, line 178: Let's add a CR for when we support other extensible variants.

dkalinichenko-js (commented):
The core change looks good.

testsuite/tests/typing-modes/exn.ml
goldfirere, line 1: I'd want to see some tests for scenarios specific to exceptions:
* [ ] `match something with exception Portable -> ...`
* [ ] `try something with Portable -> ...`
* Defining a new exception (without using an existing constructor) (I think this shouldn't affect portability)

goldfirere, line 1: I know I'm a broken record.... but documentation? We have documentation for modes now -- can we add this?

goldfirere (commented):
Just took a look at the tests

stdlib/effect.ml
```diff
@@ -27,7 +27,8 @@ let () =
         Some msg
     | _ -> None
   in
-  Printexc.Safe.register_printer printer
+  (* need magic because [Unhandled] is uncontended *)
+  Printexc.Safe.register_printer (Obj.magic_portable printer)
```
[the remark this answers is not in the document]
riaqn: Maybe it's better to have this `magic` as a big giant warning (that this needs to be improved)?

typing/env.ml
riaqn: We already have `IdTbl.find_all_and_locks`, so I followed the naming.
Also, "find all with locks" can be interpreted as "(find all) with locks", which is not what the function does. Adding to that confusion, this function takes `acc : locks`.

testsuite/tests/typing-modes/exn.ml
```diff
@@ -0,0 +1,196 @@
...
+    SemiPortable (s, _) -> print_endline s
+[%%expect{|
+exception SemiPortable of string * (unit -> unit)
+Line 5, characters 4-16:
+5 |     SemiPortable (s, _) -> print_endline s
+        ^^^^^^^^^^^^
+Error: The constructor "SemiPortable" is nonportable, so cannot be used inside a function that is portable.
+|}]
+
+(* [exn] also crosses contention. To make it safe, exception constructors
```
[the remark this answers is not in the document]
riaqn: It's probably similar to the `int ref` case? But I added them anyway.

typing/env.ml
riaqn: I'm also not sure about reuse this for `find_all`, because that require iterating the returned list, which might hurt performance.

dkalinichenko-js (approved):

PR #4063, opened 2025-05-29
Clean up "CR ocaml 5 domains" comments

NickBarnes:
There were four comments with the text "CR ocaml 5 domains," indicating things which had to be cleaned up, or at least checked, before multi-domain programs go live. This PR addresses all of them:

- `otherlibs/unix/signals.c` was basically the runtime4 version of this file. I've updated it to match (fairly well) the version from upstream. Semantically this just amounts to adding signals which are pending in the runtime to the list of signals returned by `caml_unix_sigpending()`. There are a couple of annoying hacks to make this still compile with runtime4, but it does.
- `runtime/extern.c` had a comment on `add_to_long_value` noting that it's multicore-unsafe. In fact this is fine because the function is only applied to fields of an array recently allocated by the same thread; no other core is going to race on it. I've augmented the comment and trivially refactored the function.
- `runtime4/caml/camlatomic.h` had a comment on some typedefs. This file (being `runtime4`) isn't a problem for runtime 5; the matching runtime 5 file already has the matching typedefs. I've removed the comment.
- `testsuite/tests/c-api/test_c_thread_has_lock_systhread.ml` had a comment remarking on the change in `caml_state` when using systhreads. This comment remains true; I have rewritten it to emphasize the distinction, and added a contrasting comment to the other (without `_systhread`) test.


otherlibs/unix/signals.c
```diff
@@ -90,11 +79,23 @@ CAMLprim value caml_unix_sigprocmask(value vaction, value vset)
 CAMLprim value caml_unix_sigpending(value unit)
 {
   sigset_t pending;
-  int i;
+  uintnat curr;
   if (sigpending(&pending) == -1) caml_uerror("sigpending", Nothing);
-  for (i = 1; i < NSIG; i++)
+#ifdef CAML_RUNTIME_5
+  /* Add signals which are "pending" in the runtime */
+  for (int i = 1; i < NSIG_WORDS; i++) {
```
mshinwell, line 86: Why does this loop start at 1, rather than 0, as it does upstream?

otherlibs/unix/signals.c
```diff
@@ -90,11 +79,23 @@ CAMLprim value caml_unix_sigprocmask(value vaction, value vset)
...
+    curr = atomic_load(&caml_pending_signals[i]);
+    if (curr == 0) continue;
+    for (int j = 0; j < BITS_PER_WORD; j++) {
+      if (curr & ((uintnat)1 << j))
+      sigaddset(&pending, i * BITS_PER_WORD + j + 1);
```
mshinwell, line 91: Missing indentation, let's indent even though this is wrong upstream

otherlibs/unix/signals.c
```diff
@@ -90,11 +79,23 @@ CAMLprim value caml_unix_sigprocmask(value vaction, value vset)
...
+    }
+  }
+#else
+  for (int i = 1; i < NSIG; i++)
```
mshinwell, line 95: Please check this `1` is correct as well, for good measure

otherlibs/unix/signals.c
NickBarnes: This is a bug, well-spotted.

otherlibs/unix/signals.c
NickBarnes: Fixed.

otherlibs/unix/signals.c
NickBarnes: 1 is not incorrect here, as there is no signal 0, but the relevant loops in runtime4/signals.c all start at zero so let's start at zero too, for consistency.

PR #4065, opened 2025-05-30
Record that we have switched locking schemes

NickBarnes:
It was possible to spawn domains after switching locking schemes, because we weren't setting the obvious variable. Also added a test. Found while reviewing #4013.


lpw25 (approved):
LGTM

Issue #3219, opened 2024-10-31
New treatment of relational information in Flambda 2 types


mshinwell:

bclement-ocp:
In light of the discussions related to the relations and the realization that this could also be useful to make CSE more robust, I have done some more reseach on the topic and I think that I have a more promising approach than my prototype. This approach takes inspiration from [egglog](https://arxiv.org/abs/2304.04332), which combines the best of [congruence closure](https://link.springer.com/article/10.1007/s11424-019-8377-8) and [deductive databases](https://arxiv.org/abs/1210.0481).

This has somewhat grown in scope compared to the initial goal we set with @lthls to extract `Is_int` and `Get_tag` relations from the typing env. However, I believe the approach presented below (already discussed with @lthls) would provide a principled and generic solution to many of the issues around Flambda2 losing informations during joins or otherwise due to the aliasing stars not being properly aligned. It also provides a generic interface to introduce additional abstract domains and provides control on how to schedule inter-reductions between domains.

In addition, the approach and underlying data structures can be extended (in a second step) to also provide a principled, and generic implementation of env extensions that would remove the need of threading support for extensions at specific places as was done in #1324.

## Details

Roughly, the idea would be to implement a proper deductive database in the
typing environment, that can be extended and queried as needed.

We want to store a combination of:

 - An union-find data structure that associates identifiers to a canonical
   representative (adapted into the existing `Aliases` module), and

 - A set of equations of the form `f(x1, ..., xn) = y`, where the `xi` are
   identifiers (`Simple`s) and `y` is a value in an arbitrary abstract domain,
   and `f` is a function name such as `Is_int` or `Get_tag`. Equations are kept 
   in canonical form wrt the union-find, and stored in a logic database (which
   I will not describe the data-structures here, but it is essentially a lot of
   tries).

I mention `y` being an arbitrary abstract domain above because that is one of
the powerful aspects of the approach; I will note that one particular domain of
interest is the alias domain, in which case `y` is a (canonical) identifier.

The database is kept canonical using a rebuilding procedure similar to what is
used in egglog, wherein we use (database) indices to update the corresponding
entries to canonize them. This can cause us to have duplicate entries for a
given set of arguments, in which case we perform a `meet` of the underlying
abstract domain (for the alias domain, we record equalities in the union-find
and perform a second rebuild to obtain a congruence closure algorithm).

New facts (such as `x = Is_int(y)`) are added to the database by canonicalizing
[...]

Issue #926, opened 2022-10-28
Improved compilation of match-in-match


mshinwell:

Gbury:
After some discussions with @lthls and @bclement-ocp here are some interesting things to do (and/or consider) for match-in-match (the first three points are the basic requirements for match-in-match to work as expected, the other points are improvements):
- [x] the base implementation of continuation specialization (cf #3501 )
- [x] the heuristic of match-in-match to specialize when all callsites of a continuation know the value of the scrutinee
- [ ] improve the heuristic to also handle when some callsites know the value, but some others do not
- [ ] Extend the partitioning algorithm from the typing env (i.e. the code that states whether all callsites of a continuation have a known value for the scrutinee) to handle more relations, e.g. code_ids, and possibly others (to discuss)
- [ ] consider if we want to "group" the specialization (i.e. if/when many callsite have the same value for the scrutinee) -> this may not be always good, as often continuations in the same "group" may have different values for other variables, and these may actually benefit from the specialization; we may want to use the existence of knowledge about other values for each of the callsites in order to determine whether groups are beneficial or not (depends on the extension of the partitioning algorithm to other primitives/values)
- [ ] Some switch optimizations can hinder match-in-match, notably the transformation into `Boolean_not` and into an `Array_load` -> add flags to control these optimizations (so that at least users can disable them locally so that match-in-match can trigger)
- [ ] Consider specialization of continuations whose handler does not end with a `switch`, e.g. an `apply_expr` may benefit from specialization by turning from an indirect to a direct call (if the code_id is know at all callsites)
- [ ] More generally, we may also want to specialize primitives (e.g. an `Boolean_not`, or an `Array_load` of a known immutable array), in which case to avoid duplicating too much code (particularly the code after the primitive), we'd "split" the current handler by wrapping the body of the let-expr into a new continuation, and then specialize the current continuation
- [ ] Look at what happens with **lifted** set of closures that occur in the handlers of specialized continuations. This is allowed currently, because it can work and we expect reasonable use cases, but the current approach is not ideal as it will duplicate the set of closures. We should investigate ways to at least not duplicate lifted set of closures when specializing the continuation does not change the lifted set of closures, and even look at heuristics for when it is actually beneficial to specialize such set of closures (instead of just lifting it once using the `denv_at_fork` and using the same lifted soc in all specialized handlers)

Gbury:
[...]

PR #4073, opened 2025-06-02
Exploit sharing in patricia tree combine operations

bclement-ocp:
This patch introduces a new `union`-like function to the patricia tree (and containers) interface, `diff`. It also provides `_shared` and `_sharing` variants of both `diff` and `union` to exploit internal sharing, and a typed interface to help ensure sharing is not accidentally lost.

The `diff` function is an `union`-like generalization of the existing `diff_domains`. `diff f m1 m2` is similar to `merge f m1 m2` with `f None (Some _) = None` and `f (Some v) None = Some v`.

The `_shared` variants (`diff_shared`, really) are the main motivation for this patch, and have a transformative impact on performance: `diff_shared` and `union_shared` exploit physical equality to simplify `diff_shared f m m = empty` and `union_shared f m m = m` in constant time. They also ensure maximal sharing of the result with their first argument.

The `_sharing` variants do not perform this optimisation but still ensure maximal sharing of the result with their first argument. They are useful to add or remove entries into a map while preserving sharing.

The `Shared_set` and `Shared_map` are thin wrapper modules that only expose sharing-preserving functions to avoid accidental loss of sharing.

These functions are currently unused, but will be used in the implementation #3219 and for the match-in-match heuristic (see #926), as discussed offline with @chambart and @lthls.

### DIFF
--- a/middle_end/flambda2/algorithms/container_types.mli
+++ b/middle_end/flambda2/algorithms/container_types.mli
@@ -48,3 +48,138 @@ module Make_pair (T1 : S) (T2 : S) : sig
 
   val create_from_cross_product : T1.Set.t -> T2.Set.t -> Set.t
 end
+
+module Shared_set (T : S) : sig
+  type t = private T.Set.t
+
+  type elt = T.t
+
+  val create : T.Set.t -> t
+
+  val print : Format.formatter -> t -> unit
+
+  val empty : t
+
+  val is_empty : t -> bool
+
+  val singleton : elt -> t
+
+  val add : elt -> t -> t
+
+  val remove : elt -> t -> t
+
+  val mem : elt -> t -> bool
+
+  (** [diff ss1 ss2] computes the difference of the shared sets [ss1] and [ss2].
+
+      {b Sharing}: Sharing is guaranteed with sub-trees of [ss1] that are
+      disjoint from [ss2].
+
+      {b Complexity}: The complexity of [diff (union ss1 ss2) ss1] is linear
+      in the size of [ss2]. The complexity of [diff (add k ss) ss] is the
+      same as that of [mem k (add k ss)]. *)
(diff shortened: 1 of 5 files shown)


middle_end/flambda2/algorithms/container_types.mli
```diff
@@ -48,3 +48,138 @@ module Make_pair (T1 : S) (T2 : S) : sig
...
+
+  (** [diff ss1 ss2] computes the difference of the shared sets [ss1] and [ss2].
+
+      {b Sharing}: Sharing is guaranteed with sub-trees of [ss1] that are
+      disjoint from [ss2].
+
+      {b Complexity}: The complexity of [diff (union ss1 ss2) ss1] is linear
+      in the size of [ss2]. The complexity of [diff (add k ss) ss] is the
+      same as that of [mem k (add k ss)]. *)
+  val diff : t -> t -> t
```
bclement-ocp, line 81: I made this return a `t` because it felt more general, but in practice I end up always converting to a non-shared set / non-shared map.

I'll expand on the way these are intended to be used, because it might help clarify things. In the context where we are accumulating a "main" set/map (e.g. in a typing env), we use `diff` and `diff_set` / `diff_map` for two fundamentally different scenarios:

 - `diff` is used to efficiently compute the difference between an old version of the set/map and a new version of the set/map, in order to do some processing (e.g. a join) of the changes. It is crucial that sharing can be used to reduce the complexity (the size of the changes is typically small compared to the size of the main set), however sharing of the output is not important.
- `diff_set` / `diff_map` is used to remove one (or more) elements from the main set/map; the result will be used as the new main set/map and sharing must be preserved.

Considering this, I think it makes more sense to return `T.Set.t` here, and probably also to make the `_shared` version *not* try to preserve the sharing of their output. If the reviewers agree, I'll make those changes.

PR #4074, opened 2025-06-02
Document modalities

goldfirere:
As title.

Review: @riaqn

### DIFF
--- a/jane/doc/extensions/_04-modes/modalities.md
+++ b/jane/doc/extensions/_04-modes/modalities.md
@@ -1,0 +1,137 @@
+---
+layout: documentation-page
+collectionName: Modes
+title: Modalities
+---
+
+As described in the [introduction to modes](../intro), modes are deep properties
+of values: if you have a `portable` record and you project a field, that field
+will also be `portable`. However, it is sometimes desirable to have an escape
+hatch: some nonportable data might have fields which are nevertheless portable.
+OxCaml's *modalities* feature allows you to change modes as you extract a
+component of some structure.
+
+Here is an example:
+
+```ocaml
+type tribble = {
+  spawn : unit -> tribble;
+  scurry : direction -> unit @@ portable;
+}
+```
+
+If you have a variable `trib : tribble @ nonportable` (that is, a variable named
+`trib`, whose type is `tribble` and mode is `nonportable`), then `trib.spawn`
+will be `nonportable` (this is the usual deep behavior), while `trib.scurry`
+will be `portable` (this is the new behavior, enabled by the `@@ portable`
+modality).
+
+Modalities can appear on elements of structures (records, constructors, modules,
+etc). The full description is in our [syntax page](../syntax), but modalities
+are always introduced with a `@@` herald. (This is in contrast to *modes* which
+are introduced with a `@` herald.) A modality describes the relationship between
+the mode of the structure itself and the mode of its element. We can think of a
+modality as a function from a mode to a mode: the input is the mode of the
+structure and the output is the mode of the element. Right now, all modalities
+are modes (whose interpretation is described in the next paragraph), but that
+might not always be the case.
+
+Modalities work differently on future axes vs. past axes. On a future axis (like
+portability), the modality imposes an upper bound on the mode (thus always
+lowering that mode). Thus applying the `portable` modality to a `nonportable`
+record yields a `portable` field, because `portable < nonportable`. On a past
+axis (like contention), the modality imposes a lower bound (thus always raising
+that mode). Accordingly, a `contended` modality applied to an `uncontended`
+record yields a `contended` field, because `uncontended < contended`.
+
+Because modalities impose upper and lower bounds, some modalities have no
+effect. If we have a `type t = { x : (int -> int) @@ nonportable }`, the
+`nonportable` modality has no effect: it is an upper bound on the mode of `x`,
+but `nonportable` is the top element of the portability axis, and so using it as
+an upper bound is uninformative.
+
+The difference between future axes and past axes is because this is the only
+way modalities are safe. It would be bad to have a `nonportable` element within
+a `portable` structure: you might send that structure to another thread, and
+the `nonportable` element would come along for the ride. It would similarly be
+bad to have an `uncontended` element within a `contended` structure: maybe that
+structure came from another thread and might be exposed to parallel reads and
+writes.
+
+## Modality types
+
+It is sometimes convenient to apply a modality at a time other than when you are
+defining a record. The `base` library thus provides *modality types*, such as
+[`Modes.Portable.t`](https://github.com/janestreet/base/blob/26c2f4df29a76e792cabfceb53963b3538ba6dc1/src/modes_intf.ml#L519):
+
+```ocaml
+type 'a t = { portable : 'a @@ portable } [@@unboxed]
+```
+
+This type (which does not appear at runtime -- that's what the `[@@unboxed]` does)
+allows you to store a `portable` component of an arbitrary structure.
+
+## Modalities and mode-crossing
+
+Some types are allowed to *mode cross* along certain mode axes. For example,
+portability affects only functions; a type that contains no functions (even after
+looking deeply through it) does not care about its portability. We say it
+*mode-crosses portability*. A type like `string option list` has this property,
+but a type like `(int -> int) list` does not.
+
+However, a modality allows more mode-crossing to happen. For example:
+
+```ocaml
+type t = { 
+  f : (int -> int) @@ portable;
+  value : int 
+}
+```
+
+Without the modality, it would be unsound for `t` to mode-cross portability: it
+might contain a nonportable function. However, the modality means that `f` will
+always be portable. Because the other field of `t` contains no functions, we can
+be sure that all values of type `t` do not have functions in them, and thus it
+is safe for `t` to mode-cross portability.
+
+As a consequence of this, the modality types all mode-cross along the relevant
+axis. That is, `Mode.Portability.t` mode-crosses portability. If you have a
+scenario where you need a type to mode-cross but it contains a field whose type
+does not cross that mode, a modality type might be the answer to your problem.
+
+## Modalities and mutable fields
+
+Mutable fields require special consideration. Consider this example:
+
+```ocaml
+type t = { mutable f : unit -> unit }
+
+let set (t : t @ nonportable) =
+  t.f <- (fun () -> incr global_counter)
+  
+let bad_news () =
+  let t @ portable = { f = (fun () -> ()) } in
+  let non_portable_t @ nonportable) = t in
+  set non_portable_t;
+  spawn t.f
+```
+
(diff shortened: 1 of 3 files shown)


goldfirere:
> My understanding is the following. For a modal axis, an expression's mode is a _compositional_ property; the mode of the expression depends on the modes of its subexpressions. Is it correct to say that the mode of an expression is the maximum (resp. minimum) of the modes of its subexpressions, for future-oriented (resp. past-oriented) axes?

Sadly, no. The mode of `f x` comes from the type of `f`, which may have any relationship you like with the modes of `f` and `x`. The property you describe holds true of construction expressions (record construction / constructor use) of types without past-axis modalities -- except that all axes take "maximum". In effect, this is "just" submoding.

> 
> In future-oriented modes, no subexpression can have a greater mode than its encompassing expression.

This is true for construction expressions.

> Respectively, in past-oriented modes, no subexpression can have a lower mode than its encompassing expression.

Even for construction expressions, this is not true if there is a modality.

> 
> In this view, is saying “modes are deep” a description of the mode inference mechanism? Subexpressions' modes are determined in a top-down way.

Not really. "Modes are deep" is a description of how modes affect structures in memory. It's not really about the syntax.

> 
> I'm actually under the impression that a clarification is needed. On one hand, there's the relationship between the modes inside an expression. What's the mode at each level? On the other hand, syntax and inference mechanisms allow for specifying modes without annotating every subexpression. Aren't these two related but distinct things?

I don't understand this point, sorry!

jane/doc/extensions/_04-modes/modalities.md
```diff
@@ -0,0 +1,137 @@
...
+
+The difference between future axes and past axes is because this is the only
+way modalities are safe. It would be bad to have a `nonportable` element within
+a `portable` structure: you might send that structure to another thread, and
+the `nonportable` element would come along for the ride. It would similarly be
+bad to have an `uncontended` element within a `contended` structure: maybe that
+structure came from another thread and might be exposed to parallel reads and
+writes.
+
+## Modality types
```
riaqn, line 61: I wonder if it's better to not mention `Modes.Portable.t`, since that's outside of the compiler (hence ideally shouldn't be in the compiler's doc). The URL also points to a specific git revision, which might be outdated in the future.

How about keep the type declaration, and trust that the users will find `modes.Portable.t` by themselves? (Even if they didn't, it's not the end of the world to redefine their own version).

jane/doc/extensions/_04-modes/modalities.md
```diff
@@ -0,0 +1,137 @@
...
+}
+```
+
+If you have a variable `trib : tribble @ nonportable` (that is, a variable named
+`trib`, whose type is `tribble` and mode is `nonportable`), then `trib.spawn`
+will be `nonportable` (this is the usual deep behavior), while `trib.scurry`
+will be `portable` (this is the new behavior, enabled by the `@@ portable`
+modality).
+
+Modalities can appear on elements of structures (records, constructors, modules,
```
riaqn: I think it's better to be more specific by saying:

> Modalities can appear in places such as record fields, constructor arguments, and values/submodules in structures, etc. The full description ...

(I also avoid saying "elements of structures" because that already means modules).

jane/doc/extensions/_04-modes/modalities.md
```diff
@@ -0,0 +1,137 @@
...
+etc). The full description is in our [syntax page](../syntax), but modalities
+are always introduced with a `@@` herald. (This is in contrast to *modes* which
+are introduced with a `@` herald.) A modality describes the relationship between
+the mode of the structure itself and the mode of its element. We can think of a
```
riaqn: Similarly, I'd avoid saying "structure" here. How about "container and its element"? It's probably worth going through the whole documentation to make sure we are consistent in this terminology. This can probably be done by AI (In fact, I will try to do that if you agree).

jane/doc/extensions/_04-modes/modalities.md
```diff
@@ -0,0 +1,137 @@
...
+type t = { 
+  f : (int -> int) @@ portable;
+  value : int 
+}
+```
+
+Without the modality, it would be unsound for `t` to mode-cross portability: it
+might contain a nonportable function. However, the modality means that `f` will
+always be portable. Because the other field of `t` contains no functions, we can
+be sure that all values of type `t` do not have functions in them, and thus it
```
riaqn: This sentence is not quite right. How about:

> Because the other field of `t` contains no functions, a value of type `t` at `nonportable` is just as good as a value of type `t` at `portable`. In other words, `t` crosses portability.

(BTW, I think it's slightly more concise to say "cross portability" instead of "mode-cross portability" - if you agree, I will run an AI pass over the whole documentation).

jane/doc/extensions/_04-modes/modalities.md
```diff
@@ -0,0 +1,137 @@
...
+[`Modes.Portable.t`](https://github.com/janestreet/base/blob/26c2f4df29a76e792cabfceb53963b3538ba6dc1/src/modes_intf.ml#L519):
+
+```ocaml
+type 'a t = { portable : 'a @@ portable } [@@unboxed]
+```
+
+This type (which does not appear at runtime -- that's what the `[@@unboxed]` does)
+allows you to store a `portable` component of an arbitrary structure.
+
+## Modalities and mode-crossing
```
riaqn, line 74: I think mode crossing probably deserves a page of its own, which should contain this section. For now, maybe just add a CR for that?

jane/doc/extensions/_04-modes/modalities.md
```diff
@@ -0,0 +1,137 @@
...
+}
+```
+
+Without the modality, it would be unsound for `t` to mode-cross portability: it
+might contain a nonportable function. However, the modality means that `f` will
+always be portable. Because the other field of `t` contains no functions, we can
+be sure that all values of type `t` do not have functions in them, and thus it
+is safe for `t` to mode-cross portability.
+
+As a consequence of this, the modality types all mode-cross along the relevant
```
riaqn, line 96: This isn't quite right (due to axes containing more than two levels).

I wonder if we should replace the example above with a middle mode as the general case, and derive the special case of full-crossing for monadic and comonadic axes separately?

jane/doc/extensions/_04-modes/modalities.md
```diff
@@ -0,0 +1,137 @@
...
+  t.f <- (fun () -> incr global_counter)
+  
+let bad_news () =
+  let t @ portable = { f = (fun () -> ()) } in
+  let non_portable_t @ nonportable) = t in
+  set non_portable_t;
+  spawn t.f
+```
+
+This creates a `portable` `t`, uses submoding to turn it into a `nonportable`
```
riaqn: Better to say `t @ portable`, which aligns better with reading-out-aloud and the concrete syntax.

jane/doc/extensions/_04-modes/modalities.md
```diff
@@ -0,0 +1,137 @@
...
+Horrors! The problem here is that, in `set`, just because `t` is `nonportable`,
+we cannot assume that we can write a `nonportable` function into `t.f`.
+
+To prevent examples like this from being allowed, mutable fields have
+*implied modalities*. Every mutable field has the modalities derived from
+the legacy modes (the modes that describe OCaml values that existed before
+we had modes), as given here:
+
+```
+global aliased many nonportable uncontended unyielding stateful read_write
```
riaqn: I wonder if the legacy modes should be spelled in the `intro.md`, and referred to from here.

I also think mutable fields is complex enough that deserves its own page. For now, maybe just leave a CR?

jane/doc/extensions/_04-modes/modalities.md
goldfirere: Yes I see your point here. I've made the link point to the `base` library generally (github.com/janestreet/base, which should be reliable) and agree that folks can find the actual definition if they need to.

jane/doc/extensions/_04-modes/modalities.md
goldfirere: Sure. I don't think I'd spend that much time on it, but AI might make quick work of it.

I've taken your suggestion here.

jane/doc/extensions/_04-modes/modalities.md
goldfirere: I'm torn here. "mode-cross" is an easier thing to search for (and find). I think I'd prefer keeping mode-cross. But I don't think it's strictly necessary for this to be the usage everywhere. (Maybe you disagree and find the inconsistency jarring... and you might be right.)

goldfirere: (But I took your suggestion.)

jane/doc/extensions/_04-modes/modalities.md
goldfirere: Why should this be a separate page? It seems to fit under "modalities" to me.

jane/doc/extensions/_04-modes/modalities.md
goldfirere: Hm... I think all (non-identity) modalities allow mode crossing. (Or at least they should?) It's just that middle modalities allow only partial mode crossing. Maybe this should be drawn out in the docs? Actually now I think I understand your comment about separate page. Let me take a spin at this.

jane/doc/extensions/_04-modes/modalities.md
goldfirere: Yes I agree this should just point to existing docs on legacy modes.

But I don't yet see why mutable fields deserves its own page -- at least not until it becomes more glorious in the future.

jane/doc/extensions/_04-modes/modalities.md
goldfirere: I ended up not making a new page here, but added stuff to the (rather short!) reference page. Not sure what's best. This stuff is confusing.

jane/doc/extensions/_04-modes/modalities.md
riaqn: I think there are a bunch of things to talk about for mode-crossing: the formal definition of mode crossing; sources of mode crossing (modality is only one of them); the syntax point where the compiler checks for mode crossing, etc.

I guess I'm inclined towards having a page for each substantial topic, instead of putting everything under a giant "reference" page.

(Again, I'm perfectly happy to leave this to future work)

jane/doc/extensions/_04-modes/modalities.md
riaqn: Yeah previously I misread your phrase "mode-cross along the relevant axis" as "crossing the whole axis". But I'm afraid users might get the same wrong impression too, so I think it's better to be explicit.

jane/doc/extensions/_04-modes/modalities.md
riaqn: I was thinking about the `no-mutable-implied-modalities` attribute which causes subtle mode checking. This attribute will be the default in the glorious future, and I think it's good to inform and encourage users to migrate to that future from NOW.
(But again, I don't expect us to do that in this PR).

PR #4076, opened 2025-06-03
Add the ability to emit module entry functions into a separate section

Svetlitski:
Module entry functions are ice-cold after startup, so for the sake of improving iTLB usage it is desirable to place all of them together in a separate section so that they aren't mixed in with hot code. This behavior is opt-in via the new command line flag `-module-entry-functions-section`.

### DIFF
--- a/backend/amd64/emit.ml
+++ b/backend/amd64/emit.ml
@@ -232,6 +232,12 @@ let load_symbol_addr (s : Cmm.symbol) arg =
    supported on the target system. *)
 
 let emit_named_text_section ?(suffix = "") func_name =
+  (* CR-someday ksvetlitski: In the future we should consider extending this to
+     also place other known-cold functions in a separate section (specifically
+     [.text.unlikely.caml]). This would necessitate adding a new constructor to
+     [Cfg.codegen_option], and at that point we could add a constructor for
+     module-entry-functions too so that we don't have to inspect [func_name]
+     like we do now. *)
   if !Flambda_backend_flags.module_entry_functions_section
      && String.ends_with func_name ~suffix:"__entry"
   then (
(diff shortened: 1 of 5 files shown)


backend/amd64/emit.ml
```diff
@@ -232,7 +232,29 @@ let load_symbol_addr (s : Cmm.symbol) arg =
...
+    | S_macosx
+    (* Names of section segments in macosx are restricted to 16 characters, but
+       function names are often longer, especially anonymous functions. *)
+    | S_win64 | S_mingw64
+    | S_cygwin
+      (* Win systems provide named text sections, but configure on these systems
+         does not support function sections. *) ->
+      assert false
+    | _ ->
+      D.switch_to_section_raw ~names:[".text.entry"] ~flags:(Some "ax")
```
gretay-js: It would be nice to have `caml` in the section name, and we should probably use one of the specially-named sections such as `.text.unlikely` that the default linker scripts know about.
```suggestion
      D.switch_to_section_raw ~names:[".text.unlikely.caml.entry"] ~flags:(Some "ax")
```

driver/flambda_backend_args.ml
```diff
@@ -1069,6 +1084,8 @@ module Flambda_backend_options_impl = struct
     Flambda_backend_flags.reorder_blocks_random := Some seed
   let basic_block_sections () =
     set' Flambda_backend_flags.basic_block_sections ()
+  let module_entry_functions_section () =
```
gretay-js, line 1088: For completeness, the flag should also be available via `OCAMLPARAM`, similarly to `basic-block-sections` around L1436.

backend/amd64/emit.ml
Svetlitski: I agree in general, but `.text.unlikely` feels somewhat wrong, since there's a 100% chance of the module entry functions being executed. Restricting ourselves to just [section name prefixes linkers know about by default](https://github.com/llvm/llvm-project/blob/6c1091ea3fb7ab31aa9b8406cb32b9a549c4e7cd/lld/ELF/LinkerScript.cpp#L80-L104), how about `.text.startup.caml` instead?

gretay-js: sounds good, thanks!

PR #4077, opened 2025-06-03
Mode cross modalities even for non-principal type

riaqn:
A non-principal type wrapped in modalities should still cross modes by the modalities.

### DIFF
--- a/typing/ctype.ml
+++ b/typing/ctype.ml
@@ -5010,14 +5010,16 @@ let crossing_of_jkind env jkind =
   Jkind.get_mode_crossing ~jkind_of_type jkind
 
 let crossing_of_ty env ?modalities ty =
-  if not (is_principal ty)
-  then Crossing.top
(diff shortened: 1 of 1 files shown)


goldfirere:
Saw this tiny PR so I thought I'd review. The code looks good, but is there a test?

riaqn:
This is currently not observable, because this function is never called with non-principal type and `Some modalities`.

goldfirere:
Oh, well, that's true. I still think there's something not quite right here (because we're handling a case that can't come up in practice without documenting this fact) but it's not worth fixing. I'll merge.

lpw25:
I actually think that the thing that is wrong here is just that we don't have any support for doing unit tests of functions in the compiler, so the only way to test their behaviour is via running the compiler.

It's still better to have a function behave correctly on all of its input range rather than only on the parts we currently happen to use.

PR #4078, opened 2025-06-03
Restore gc_regs_buckets after stack overflow

TheNumbat:
`caml_call_realloc_stack` pops a `gc_regs` bucket to save registers before calling `caml_try_realloc_stack`.
If the reallocation fails, it does not restore the bucket before raising the stack overflow async exn.
If the program then needs to spill gc_regs (e.g. tries to grow the stack again), it will crash due the bucket list being empty.

It now restores the bucket in the stack overflow case. I've updated `callback.ml` to fail without this change.

Also, I noticed in `arm64.S`, `caml_call_realloc_stack` does not switch back to the C stack before raising the async exn -- is that wrong? It also restores the registers, which means it doesn't have the gc_regs issue.


mshinwell (approved):
Let's do the arm64 fix separately.

PR #4079, opened 2025-06-04
Shortcut in overlap computation (linscan & greedy allocators)

xclerc:
As per title, use the information about
interval bounds to avoid useless iteration
over lists of ranges.

### DIFF
--- a/backend/regalloc/regalloc_gi_utils.ml
+++ b/backend/regalloc/regalloc_gi_utils.ml
@@ -193,6 +193,7 @@ module Range = struct
 
   let rec overlap : t list -> t list -> bool =
    fun left right ->
+    (* CR-soon xclerc for xclerc: use the same version as linscan (cursors). *)
     match left, right with
     | left_hd :: left_tl, right_hd :: right_tl ->
       if left_hd.end_ >= right_hd.begin_ && right_hd.end_ >= left_hd.begin_
@@ -255,9 +256,10 @@ module Interval = struct
     List.iter t.ranges ~f:(fun r -> Format.fprintf ppf " %a" Range.print r)
 
   let overlap : t -> t -> bool =
-   (* CR xclerc for xclerc: short-cut to avoid iterating over the lists using
-      the Interval.{begin_in_,end_} fields *)
-   fun left right -> Range.overlap left.ranges right.ranges
+   fun left right ->
+    if left.end_ < right.begin_ || right.end_ < left.begin_
+    then false
+    else Range.overlap left.ranges right.ranges
 
   (* CR xclerc for xclerc: assumes no overlap *)
   let add_ranges : t -> from:t -> unit =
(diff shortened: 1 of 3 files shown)


backend/regalloc/regalloc_gi_utils.ml
```diff
@@ -255,15 +256,16 @@ module Interval = struct
     List.iter t.ranges ~f:(fun r -> Format.fprintf ppf " %a" Range.print r)
 
   let overlap : t -> t -> bool =
-   (* CR xclerc for xclerc: short-cut to avoid iterating over the lists using
-      the Interval.{begin_in_,end_} fields *)
-   fun left right -> Range.overlap left.ranges right.ranges
+   fun left right ->
+    if left.end_ < right.begin_ || right.end_ < left.begin_
```
spiessimon: This looks correct even with `begin_ = end_ = max_int` as initialized by `make_empty`.

backend/regalloc/regalloc_gi_utils.ml
```diff
@@ -255,15 +256,16 @@ module Interval = struct
...
+    then false
+    else Range.overlap left.ranges right.ranges
 
   (* CR xclerc for xclerc: assumes no overlap *)
   let add_ranges : t -> from:t -> unit =
    fun t ~from ->
     t.begin_ <- Int.min t.begin_ from.begin_;
-    t.end_ <- Int.min t.end_ from.end_;
+    t.end_ <- Int.max t.end_ from.end_;
```
spiessimon: `Int.min` was definitely wrong. However, I think this will now lead to quite imprecise bounds: `make_empty` uses `max_int` as the upper bound. Thus, if we ever add an empty range, the upper bound goes up to `max_int`. Perhaps the representation could be `start_, length_` instead of `begin_,end_` or `make_empty` could set `end_ = -1` and `begin_ = 0`.

spiessimon (commented):
I think adding the short cut is a good idea. It would be nice to fix the bounds for the empty interval if it's not too much trouble.

backend/regalloc/regalloc_gi_utils.ml
```diff
@@ -236,34 +237,55 @@ module Range = struct
...
 
 module Interval = struct
+  (* CR-soon xclerc for xclerc: use a doubly-linked list for `ranges`, and do
+     not store bounds. *)
   type t =
-    { mutable begin_ : int;
-      mutable end_ : int;
+    { mutable begin_ : int option;
+      mutable end_ : int option;
       mutable ranges : Range.t list
```
spiessimon, line 248:
```suggestion
      (* The `begin_` and `end_` fields should always either both be `None`, or they should both be `Some`. *)
      mutable ranges : Range.t list
```

xclerc:
Thanks for review; as discussed off-line,
it is likely we will eventually only store
a

PR #4080, opened 2025-06-04
Simplify modality atoms

riaqn:
Currently `Modality.Atom.raw` is a variant (join or meet), but this information is duplicated by the axis being monadic or comonadic. This PR removes the variants and some `assert false`.

### DIFF
--- a/typing/jkind.ml
+++ b/typing/jkind.ml
@@ -1721,12 +1721,9 @@ module Const = struct
               let (P axis) = Mode.Const.Axis.alloc_as_value (P axis) in
               match axis with
               | Monadic monadic ->
-                Atom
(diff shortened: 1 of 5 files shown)


goldfirere:
I'm willing to be convinced, but I think I prefer the old code over the new. The old code is more explicit, in an important way. I do see how the old code had more degrees of freedom than it should have, but if we want to eliminate that, maybe put another type parameter on `raw` that indicates monadic-vs-comonadic or something? It just feels like the new code is more likely to lead to mistakes than the old.

goldfirere:
No, I'm worried that someone will misinterpret the mode in the raw modality and do the wrong thing.

PR #4081, opened 2025-06-04
Linscan: remove leftovers from the skip list transition

xclerc:
As per title, this pull request only deletes
code that happens to be dead.


spiessimon (approved):
LGTM!

PR #4088, opened 2025-06-04
Void function arguments and returns

rtjoa:
Supports `void` function arguments and returns. Also adds `%unbox_unit` to create values of layout `void`. We still gate `void` by `-extension layouts_alpha`.

## Implementation

Through typechecking, void is represented as a distinct base sort (as before). In Lambda and Flambda2, terms of layout void are represented as nullary unboxed products, which Flambda was updated to handle in #4083. (Products containing void are products containing empty products, i.e. the tree structure is preserved rather than flattened.)
 
## Testing
The runtime behavior test [testsuite/tests/typing-layouts-void/void.ml](https://github.com/ocaml-flambda/flambda-backend/blob/rtjoa.void-args-returns/testsuite/tests/typing-layouts-void/void.ml) is adapted from https://github.com/mshinwell/flambda-backend/blob/133dfea83b7eb87dabdb12d2f36bf3836a3dd4bc/up.ml.

Existing tests of `void` should suffice for typing. 

## Reviewing
Several errors change from
```
Error: Non-value layout void detected in [Typeopt.layout] as sort for type
       void_rec. Please report this error to the Jane Street compilers team.
```
to another "Please report this error to the Jane Street compiler team" error. These should all be fixed when we allow `void` in blocks, and should not be problematic because `void` is still in alpha with this PR. 

## Stack
Depends on:
- https://github.com/ocaml-flambda/flambda-backend/pull/4083

Full dependency graph:
```
Support nullary
unboxed products ──> Void args/returns #4088 ──┐
in flambda #4083                               ├──> Void in blocks,
                                               │    move from alpha->stable
                     Delete lbl_pos #4086 ─────┘
```


testsuite/tests/typing-layouts-void/void.ml
dkalinichenko-js, line 1: Would be good to include tests for C stubs, even if they're not implemented yet.

dkalinichenko-js: Also, a more exhaustive test file similar to e.g. https://github.com/oxcaml/oxcaml/blob/main/testsuite/tests/typing-layouts-float32/unboxed_float32s.ml would be good, though it's not necessary yet.

testsuite/tests/typing-layouts-void/void.ml
```diff
@@ -0,0 +1,192 @@
...
+  Printf.printf "%d (expected 42)\n%!" (p (void_from_product_inlined ()));
+  let p = Sys.opaque_identity (two_voids_const_side2 (void_from_product_inlined ())) in
+  Printf.printf "after definition\n%!";
+  Printf.printf "%d (expected 42)\n%!" (p (void_from_product_inlined ()))
+
+(* version without printing to declutter Cmm *)
+let[@inline never] returns_unboxed_pair_of_voids_not_inlined0 x =
+  #(void (), void ())
+
+let[@inline never] returns_unboxed_pair_of_voids_not_inlined x =
```
dkalinichenko-js, line 167: What about a pair of `void` and something else? What about a pair of `void & void` and something else?

dkalinichenko-js: What about an unboxed record with 1) 1 void field 2) 2 void fields 3) non-void and void fields?

dkalinichenko-js (commented):
Looks good overall, just needs a few more tests.

testsuite/tests/typing-layouts-void/void.ml
rtjoa: Explicitly disallowed C stubs and tested this in ef4e27523f07084c688421c6ccdd6bcbd9863ba7, thanks for catching.

Thanks for the pointer to https://github.com/oxcaml/oxcaml/blob/main/testsuite/tests/typing-layouts-float32/unboxed_float32s.ml - I propose adapting it for the PR adding void in blocks.

testsuite/tests/typing-layouts-void/void.ml
rtjoa: Added all of these in a4ec0456b7e742253fd1a610fe6778da51ba2959.

rtjoa:
@dkalinichenko-js thanks for the review! See responses to your comments and the two most recent commits (ef4e27523f07084c688421c6ccdd6bcbd9863ba7, a4ec0456b7e742253fd1a610fe6778da51ba2959). I also rebased but the conflicts were trivial.

testsuite/tests/typing-layouts-void/void.ml
dkalinichenko-js: Yep, that works.

dkalinichenko-js (approved):

PR #4089, opened 2025-06-04
Split use tracking for mutable variables into uses and mutations

jra4:
Add a warning for mutable variables that are never mutated and stop counting a mutable variable mutation as a "use". The `unused-var` and `unused-var-strict` warnings are updated to take a `mutated` flag to give better messages when a mutable variable is mutated but not used.

Note that use tracking is done on variables, not values. Consider the example:
```
let mutable x = 5 in
x <- 10;
print_int x
```
`x` is mutated and used, so no errors are generated. However, the analogous program in Rust generates an error since the value `5` is never used.

See Tests 18.1, 18.2, and 18.3 in `testsuite/tests/typing-local/let_mutable.ml` for examples.

### DIFF
--- a/typing/typecore.ml
+++ b/typing/typecore.ml
@@ -9478,7 +9478,7 @@ and type_let_def_wrap_warnings
     ?(check = fun name mutated -> Warnings.Unused_var { name; mutated })
     ?(check_strict = fun name mutated ->
       Warnings.Unused_var_strict { name; mutated } )
-    ?(check_mutable = fun name -> Warnings.Unused_mutable name)
+    ?(check_mutable = fun name -> Warnings.Unmutated_mutable name)
     ~is_recursive ~entirely_functions ~exp_env ~new_env ~spat_sexp_list
     ~attrs_list ~mode_pat_typ_list ~pvs
     type_def =
--- a/utils/warnings.ml
+++ b/utils/warnings.ml
@@ -122,7 +122,7 @@ type t =
   | Unused_tmc_attribute                    (* 71 *)
   | Tmc_breaks_tailcall                     (* 72 *)
   | Generative_application_expects_unit     (* 73 *)
-  | Unused_mutable of string                (* 186 *)
+  | Unmutated_mutable of string             (* 186 *)
   | Incompatible_with_upstream of upstream_compat_warning (* 187 *)
   | Unerasable_position_argument            (* 188 *)
   | Unnecessarily_partial_tuple_pattern     (* 189 *)
@@ -217,7 +217,7 @@ let number = function
   | Unused_tmc_attribute -> 71
   | Tmc_breaks_tailcall -> 72
   | Generative_application_expects_unit -> 73
-  | Unused_mutable _ -> 186
+  | Unmutated_mutable _ -> 186
   | Incompatible_with_upstream _ -> 187
   | Unerasable_position_argument -> 188
   | Unnecessarily_partial_tuple_pattern -> 189
@@ -1226,7 +1226,7 @@ let message = function
   | Generative_application_expects_unit ->
       "A generative functor\n\
        should be applied to '()'; using '(struct end)' is deprecated."
-  | Unused_mutable v -> "mutable variable " ^ v ^ " was never mutated."
+  | Unmutated_mutable v -> "mutable variable " ^ v ^ " was never mutated."
   | Incompatible_with_upstream (Immediate_erasure id)  ->
       Printf.sprintf
       "Usage of layout immediate/immediate64 in %s \n\
--- a/utils/warnings.mli
+++ b/utils/warnings.mli
@@ -130,7 +130,7 @@ type t =
   | Tmc_breaks_tailcall                     (* 72 *)
   | Generative_application_expects_unit     (* 73 *)
 (* Oxcaml specific warnings: numbers should go down from 199 *)
-  | Unused_mutable of string             (* 186 *)
+  | Unmutated_mutable of string             (* 186 *)
   | Incompatible_with_upstream of upstream_compat_warning (* 187 *)
   | Unerasable_position_argument            (* 188 *)
   | Unnecessarily_partial_tuple_pattern     (* 189 *)


typing/env.mli
```diff
@@ -343,6 +346,8 @@ type settable_variable =
   | Instance_variable of Path.t * Asttypes.mutable_flag * string * type_expr
   | Mutable_variable of Ident.t * Mode.Value.r * type_expr * Jkind.Sort.t
 
+(** For a mutable variable, [use] means mark as mutated. For an instance
+    variable, it means mark as used (unclear if this is desired). *)
```
jra4: In particular, if an instance variable is mutated but never accessed, should that throw a warning? The current compiler's answer is no, but if we are adding for a warning for `unmutated mutable variable`, might as well make it a more general `unmutated settable variable` warning

typing/typecore.ml
```diff
@@ -9533,7 +9534,19 @@ and type_let_def_wrap_warnings
...
+                | Val_mut _ ->
+                  let mutated = ref false in
+                  if not (name = "" || name.[0] = '_' || name.[0] = '#') then
+                    add_delayed_check
+                      (fun () ->
+                        if not !mutated then
+                          Location.prerr_warning vd.Subst.Lazy.val_loc
+                            (check_mutated name)
+                      );
+                  Env.set_value_mutated_callback vd (fun () -> mutated := true)
```
jra4: The analogous callback passed to `Env.set_value_used_callback` (a few lines above) is longer, and I don't fully understand the stuff with `slot` and `current_slot`. My callback is probably not correct.

testsuite/tests/typing-local/let_mutable.ml
```diff
@@ -568,6 +568,32 @@ Error: This expression has type "int" but an expression was expected of type
...
+
+(* Test 18.2: mutation doesn't count as use *)
+let x_18_2 =
+  let mutable x = 3 in x <- 4; 4
+;;
+[%%expect{|
+Line 2, characters 14-15:
+2 |   let mutable x = 3 in x <- 4; 4
+                  ^
+Warning 26 [unused-var]: unused variable x.
```
ccasin: I think this warning is too confusing - the user can see what looks like a use right there on the line.  I think we should instead do something like the situation for record fields, where there are separate warnings for "never mutated" and "mutated but never read".  See how warning 69 is implemented, from warnings.ml:

```ocaml
type field_usage_warning =
  | Unused
  | Not_read
  | Not_mutated

type t =
  ...
  | Unused_field of
      { form : string; field : string; complaint : field_usage_warning }(* 69 *)
```

utils/warnings.ml
```diff
@@ -570,6 +572,13 @@ let descriptions = [
     description = "A generative functor is applied to an empty structure \
                    (struct end) rather than to ().";
     since = since 5 1 };
+  { number = 186;
+    names = ["unused-mutable"];
+    description =
+    "Suspicious mutable variable is never mutated: mutable variable that is\n\
+    \    bound with \"let\" or \"as\", and doesn't start with an\n\
+    \    underscore (\"_\") character.";
+    since = since 5 2 };
```
ccasin, line 580: This is called "unused-mutable", but it looks like it's specifically for mutable variables that are never mutated, which I think should get a more explicit name.  (This is probably subsumed by my above suggestion to rework let-mutable warnings by analogy to warning 69)

typing/typecore.ml
ccasin: The point of this code is to deal with recursive groups.  If I have a `let rec ... and ... and` and all of the bindings are defined in terms of each other, but none of them are used from outside the recursive group, then we still want to issue unused warnings even though there seem to be uses within the recursive group itself.

I think the situation for mutable variables is simpler: since they can never appear within recursive group, any use is a genuine use. So I think maybe we can get away with not adding anything here, and instead just change `Env.lookup_settable_variable` to take as an argument whether you are looking the variable up for use or mutation, and do something in env similar to what is done for warning 69?

typing/env.mli
ccasin: I agree it's possible we can make improvements to instance variables, but I think it's probably not worth investing time in considering the limited use of the class system and the fact that it will be annoying for upstrream compatibility.

typing/typecore.ml
```diff
@@ -9549,12 +9555,27 @@ and type_let_def_wrap_warnings
...
                 let used = ref false in
+                let mutable_ =
+                  (match vd.val_kind with Val_mut _ -> true | _ -> false)
+                in
+                let unused_mutable = ref mutable_ in
                 if not (name = "" || name.[0] = '_' || name.[0] = '#') then
                   add_delayed_check
                     (fun () ->
-                      if not !used then
+                      if not !used || !unused_mutable then
```
ccasin: I think the below match is pretty hard to understand, as a reader.  I suggest the following changes:
* `unused_mutable`'s meaning is a bit unclear - does it mean "not used" or "not mutated".  Suggest changing it to "unmutated" or similar.
* Maybe instead of having the big pattern match, you can have an if then else like:
```
if not !used then
  ...
else if !unmutated then
  ...
;
```

And have comments in each case explaining it.

testsuite/tests/typing-local/let_mutable.ml
```diff
@@ -568,6 +568,45 @@ Error: This expression has type "int" but an expression was expected of type
...
 
+(* Test 18.1: unmutated mutable variable warning *)
+let x_18_1 =
+  let mutable x = 3 in x + 1
+;;
+[%%expect{|
+Line 2, characters 14-15:
+2 |   let mutable x = 3 in x + 1
+                  ^
+Warning 186 [unused-mutable]: variable x does not need to be mutable.
```
ccasin: I think that, just to help people pattern match, we should rephrase this to be closer to warning 69:
```
File "/home/ccasinghino/tmp/mut.ml", line 6, characters 13-28:
6 |   type t = { mutable x : int }
                 ^^^^^^^^^^^^^^^
Warning 69 [unused-field]: mutable record field x is never mutated.
```

utils/warnings.ml
ccasin: To be clearer: my concern is that "unused-mutable" could be read to mean "mutable variable that is never used" or "mutable variable that is never mutated", and we should pick a name that more clearly means the latter.

ccasin (commented):
This looks close to good now.  Per discussion, I'm OK with the choice to leave this in typecore, but think we should rephrase one of the errors (and there's still another outstanding comment in `warnings.ml` about picking a clearer name).

utils/warnings.ml
jra4: Changed to "unmutated-mutable"

testsuite/tests/typing-local/let_mutable.ml
jra4: This has been changed to [unmutated-mutable]

ccasin (approved):
Looks great now - thanks!

PR #4091, opened 2025-06-05
fix: genprintval should not call `is_null` on a `value` layout

bclement-ocp:
Genprintval uses calls to `is_null` on values of type `Obj.t` in order to be able to deal with nullable values at toplevel. In the native version, flambda2 will happily simplify these calls to `false` in certain circumstances, since `Obj.t` has layout `value` and values of type `Obj.t` cannot be null.

This patch first converts to an `or_null` value (using %opaque) before calling `is_null` to ensure this does not happen.


dkalinichenko-js (approved):
LGTM.

PR #4098, opened 2025-06-06
`Typeopt.array_type_kind`: when element is non-value or unknown, error instead of defaulting to `Pgenvalue`

rtjoa:
Since `5.1.1minus-12`, compiling this program with `ocamlopt` causes an flambda fatal error:
```ocaml
module M : sig
  type t
  val x : t
  external unsafe_get : t -> int -> int64# = "%array_unsafe_get"
end = struct
  type t = int64# array
  let x = [| #0L |]
  external unsafe_get : t -> int -> int64# = "%array_unsafe_get"
end

let get () : int -> int64# = M.unsafe_get M.x
```

The cause is that when `Typeopt.array_type_kind` calculates an array kind from an array type, if it's not visible that the type is an array type (e.g., it's abstract), it defaults the kind to `Pgenarray`, which is not safe for non-value arrays.

Two possible fixes:

1. Give an error when an array operation would otherwise default to `Pgenarray` but the element sort (for `get`, the element is the return type, for `set`, it's the third type parameter) is not actually `value`. (And always give an error if we can't tell the element sort from the typechecked application of the array operation.)
2. Don't default to `Pgenarray`, and choose a different array kind, based on the element jkind.

(1) is safer and good if we want to get this fix in soon, as it simply makes more programs error. (2) is probably more useful, as it makes more programs typecheck, but less safe. This PR implements (1). See the new tests in `testsuite/tests/typing-layouts-arrays/array_type_kind_errors.ml` for examples.

## Stack
1. https://github.com/ocaml-flambda/flambda-backend/pull/4101
2. https://github.com/ocaml-flambda/flambda-backend/pull/4098

### DIFF
--- a/lambda/translprim.ml
+++ b/lambda/translprim.ml
@@ -1273,7 +1273,8 @@ let specialize_primitive env loc ty ~has_constant_constructor prim =
          fixed. To do that, we will need more checking of primitives
          in the front end. *)
       let array_type =
-        glb_array_type loc t (array_type_kind ~elt_sort:None env loc p)
+        glb_array_type loc t
+          (array_type_kind ~elt_sort:None ~elt_ty:None env loc p)
       in
       if t = array_type then None
       else Some (Primitive (Parraylength array_type, arity))
@@ -1281,16 +1282,18 @@ let specialize_primitive env loc ty ~has_constant_constructor prim =
   | Primitive (Parrayrefu (rt, index_kind, mut), arity), p1 :: _ -> begin
       let loc = to_location loc in
       let array_ref_type =
-        glb_array_ref_type loc rt (array_type_kind ~elt_sort:None env loc p1)
+        glb_array_ref_type loc rt
+          (array_type_kind ~elt_sort:None ~elt_ty:(Some rest_ty) env loc p1)
       in
       let array_mut = array_type_mut env p1 in
       if rt = array_ref_type && mut = array_mut then None
       else Some (Primitive (Parrayrefu (array_ref_type, index_kind, array_mut), arity))
     end
-  | Primitive (Parraysetu (st, index_kind), arity), p1 :: _ -> begin
+  | Primitive (Parraysetu (st, index_kind), arity), p1 :: _ :: p3 :: _ -> begin
       let loc = to_location loc in
       let array_set_type =
-        glb_array_set_type loc st (array_type_kind ~elt_sort:None env loc p1)
+        glb_array_set_type loc st
+          (array_type_kind ~elt_sort:None ~elt_ty:(Some p3) env loc p1)
       in
       if st = array_set_type then None
       else Some (Primitive (Parraysetu (array_set_type, index_kind), arity))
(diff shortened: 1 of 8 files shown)


testsuite/tests/typing-layouts-arrays/array_type_kind_errors.ml
```diff
@@ -0,0 +1,77 @@
+(* TEST
+ include stdlib_upstream_compatible;
+ flambda2;
+ expect;
+*)
+
+type ('a : any_non_null) t
+external unsafe_get : 'a t -> int -> 'a = "%array_unsafe_get"
```
bclement-ocp: Taking a quick look since I investigated this from the flambda side; shouldn't this be:

```ocaml
external unsafe_get : ('a : any_non_null) t -> int -> 'a = "%array_unsafe_get"
[@@layout_poly]
```

Otherwise the error for `let f a i : int64# = unsafe_get a i` seems unrelated.

rtjoa: Thanks! Updated these tests.

typing/typeopt.ml
```diff
@@ -296,8 +299,33 @@ let array_type_kind ~elt_sort env loc ty =
...
+      | Ok _ -> Pgenarray
+      | Error e ->
+        (* CR layouts v4: rather than constraining [elt_ty]'s jkind to be value,
+           we could instead use its jkind to determine a non-value array kind.
+
+           We are choosing to error in this case for now because it is safer,
+           and because it could be potentially confusing that there is a second
+           source of information used to determine array type kinds (in addition
+           to the type kind of the array parameter). See PR #4098.
+        *)
```
ccasin, line 320: Can you note here that it also wouldn't help in the cases we looked at where this doesn't arise from an explicit user-written primitive (case 3 in your notes)? I'm not particularly worried about users being confused whether the array or element type matters when writing the primitive out explicitly (if it's different, the problem is on them). But the point that this doesn't work in other cases where this primitve appears is more compelling as a reason to be OK with this stop gap.

lambda/translprim.ml
```diff
@@ -1273,41 +1273,46 @@ let specialize_primitive env loc ty ~has_constant_constructor prim =
...
       in
       if t = array_type then None
       else Some (Primitive (Parraylength array_type, arity))
     end
   | Primitive (Parrayrefu (rt, index_kind, mut), arity), p1 :: _ -> begin
       let loc = to_location loc in
       let array_ref_type =
-        glb_array_ref_type loc rt (array_type_kind ~elt_sort:None env loc p1)
+        glb_array_ref_type loc rt
+          (array_type_kind ~elt_sort:None ~elt_ty:(Some rest_ty) env loc p1)
```
ccasin, line 1286: I think this one is a little tricky.  In the pattern match case, we're checking just that we found one arrow (`p1 :: _`).  Imagine:

```
type foo = int -> float#

external array_get : abstract_ufloat_array -> foo = "%array_unsafe_get"
```

I think our heuristic will incorrectly allow this since `rest` is a value.

I think we want to check for at least two arguments.  Not exactly two arguments, because it could be specialized to arrays of functions.

lambda/translprim.ml
```diff
@@ -1273,41 +1273,46 @@ let specialize_primitive env loc ty ~has_constant_constructor prim =
...
       let loc = to_location loc in
       let array_set_type =
-        glb_array_set_type loc st (array_type_kind ~elt_sort:None env loc p1)
+        glb_array_set_type loc st
+          (array_type_kind ~elt_sort:None ~elt_ty:(Some p3) env loc p1)
       in
       if st = array_set_type then None
       else Some (Primitive (Parraysetu (array_set_type, index_kind), arity))
     end
   | Primitive (Parrayrefs (rt, index_kind, mut), arity), p1 :: _ -> begin
```
ccasin, line 1301: Similar issue here as above

lambda/translprim.ml
rtjoa: This example seems to be caught by the primitive arity check; added a test case in
31fa2642197f6d46655b83b88dc95a658d9271c2.

typing/typeopt.ml
rtjoa: Noted.

PR #4099, opened 2025-06-06
Move systhreads4 stubs into runtime4 itself

mshinwell:
This removes the requirement that the systhreads library be linked before using e.g. `Mutex`, which is problematic e.g. in contexts where code may also be compiled with jsoo (where the systhreads lib is absent).

The moved files should be diffed during review against the ones at the base revision, as there are some minor changes.

### DIFF
--- a/runtime4/Makefile
+++ b/runtime4/Makefile
@@ -118,8 +118,6 @@ endif
 
 OC_CPPFLAGS += -DCAMLDLLIMPORT=
 
-#OC_NATIVE_CPPFLAGS += -Isysthreads
-
 # merge5: runtime4 requires -DCAML_NAME_SPACE and doesn't compile cleanly with two
 # new warnings added to the build in ocaml5
 OC_CPPFLAGS += -DCAML_NAME_SPACE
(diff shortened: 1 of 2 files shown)


runtime4/Makefile
```diff
@@ -117,6 +118,8 @@ endif
 
 OC_CPPFLAGS += -DCAMLDLLIMPORT=
 
+#OC_NATIVE_CPPFLAGS += -Isysthreads
```
TheNumbat: Delete?

runtime4/systhreads/st_stubs.c
```diff
@@ -15,7 +15,7 @@
 
 #define CAML_INTERNALS
 
-#define CAML_NAME_SPACE
+// CAML_NAME_SPACE is already defined at this point
```
TheNumbat: could use ifndef, define instead

TheNumbat (commented):
Changes look fine, but I don't think this is sufficient - `caml_thread_initialize` is called at top level in `systhreads4/thread.ml`, but now needs to happen in the stdlib.

TheNumbat (approved):
(nvm, deleting the versions in domain.c that error is sufficient)

mshinwell:
We discussed this - full thread initialization still requires use of the systhreads lib.  However for `Mutex` and `Condition` such initialization isn't required.

PR #4100, opened 2025-06-06
Ban $ from being interpreted as an infix operator; <[, ]> delimit quotations

Dreian:
These operators are being reserved for use in runtime metaprogramming (`<[` for opening a quotation, `]>` for closing, `$` for splicing, `<[:` for creating a quotation of a type).

Infix operators with such banned names are removed or renamed.


Dreian:
> We have quite a few uses of `<<` in the tree, and removing them all could be painful. Could we use another operator for metaprogramming?

We won't be using `<<` and `>>` in the end; instead, we'll be using `<[` and `]>`, which are not valid operators.

PR #4101, opened 2025-06-06
Don't specialize `%obj_field`, `%obj_setfield`, and `%obj_size`

rtjoa:
The primitives `%obj_field`, `%obj_setfield`, and `%obj_size` all masquerade as array primitives, whose array type kinds are specialized. This never changes the obj primitives in practice, as they are never bound to array types.

However, it's useful to stop specializing these primitives, as it allows us to add extra checks to array primitive specialization in the following PR. (Eventually, we should just make separate obj primitives for these in lambda, and a CR is left stating as such, but we'd like to get the following PR through.)

## Stack
1. https://github.com/ocaml-flambda/flambda-backend/pull/4101
2. https://github.com/ocaml-flambda/flambda-backend/pull/4098

### DIFF
--- a/lambda/translprim.ml
+++ b/lambda/translprim.ml
@@ -1225,6 +1225,16 @@ let peek_or_poke_layout_from_type ~prim_name error_loc env ty
     | Pbottom ->
       raise (Error (error_loc, Wrong_layout_for_peek_or_poke prim_name))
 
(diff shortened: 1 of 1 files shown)


ccasin (approved):
I looked through our code base and confirmed these are never used in a way where their behavior would change as a result of this PR.

PR #4105, opened 2025-06-06
More informative error for mixed product arrays

rtjoa:
Give a more informative error message for product arrays whose elements are neither scannable nor external.

Old:
```
1 | let f_bad (x : #(string * float#)) = make_vect 42 x
                                         ^^^^^^^^^^^^^^
Error: Unboxed product array elements must be external or contain all gc
       scannable types. The product type this function is applied at is
       not external but contains an element of sort float64.
```
New:
```
1 | let f_bad (x : #(string * float#)) = make_vect 42 x
                                         ^^^^^^^^^^^^^^
Error: An unboxed product array element must be formed from all
       external types (which are ignored by the gc) or all gc-scannable types.
       But this array operation is peformed for an array whose
       element type is #(string * float#), which is an unboxed product
       that is not external and contains a type with the non-scannable
       layout float64.
       Hint: if the array contents should not be scanned, annotating
       contained abstract types as [mod external] may resolve this error.
```

### DIFF
--- a/testsuite/tests/typing-layouts-products/product_arrays.ml
+++ b/testsuite/tests/typing-layouts-products/product_arrays.ml
@@ -125,8 +125,9 @@ Line 1, characters 37-51:
 1 | let f_bad (x : #(string * float#)) = make_vect 42 x
                                          ^^^^^^^^^^^^^^
 Error: Unboxed product array elements must be external or contain all gc
-       scannable types. The product type this function is applied at is
-       not external but contains an element of sort float64.
+       scannable types. But this array operation is peformed for an array
+       whose element type is #(string * float#), which is an unboxed
+       product that is not external and contains the non-scannable sort float64.
 |}]
 
 (* And similarly if we specialize it at declaration time. *)
@@ -166,8 +167,9 @@ Line 3, characters 21-29:
 3 | let make_bad_app x = make_bad x
                          ^^^^^^^^
 Error: Unboxed product array elements must be external or contain all gc
-       scannable types. The product type this function is applied at is
-       not external but contains an element of sort float64.
+       scannable types. But this array operation is peformed for an array
+       whose element type is #(string * float#), which is an unboxed
+       product that is not external and contains the non-scannable sort float64.
 |}]
 
 (* Unboxed vectors are also rejected. *)
@@ -217,8 +219,9 @@ Line 1, characters 43-48:
 1 | let f_bad (x : #(string * float#) array) = len x
                                                ^^^^^
 Error: Unboxed product array elements must be external or contain all gc
-       scannable types. The product type this function is applied at is
-       not external but contains an element of sort float64.
+       scannable types. But this array operation is peformed for an array
+       whose element type is #(string * float#), which is an unboxed
+       product that is not external and contains the non-scannable sort float64.
 |}]
 
 (* And similarly if we specialize it at declaration time. *)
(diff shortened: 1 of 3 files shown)


testsuite/tests/typing-layouts-products/product_arrays.ml
```diff
@@ -125,8 +125,9 @@ Line 1, characters 37-51:
 1 | let f_bad (x : #(string * float#)) = make_vect 42 x
                                          ^^^^^^^^^^^^^^
 Error: Unboxed product array elements must be external or contain all gc
-       scannable types. The product type this function is applied at is
-       not external but contains an element of sort float64.
+       scannable types. But this array operation is peformed for an array
+       whose element type is #(string * float#), which is an unboxed
+       product that is not external and contains the non-scannable sort float64.
```
goldfirere:
```
Error: An unboxed product array element must be formed from all non-gc
       types or all gc-scannable types.
       But this array operation is peformed for an array
       whose element type is #(string * float#), which is an unboxed
       product that mixes the two, containing both scannable types and a non-scannable type with layout float64.
```

? My goals here:
* Avoid "external", which might not be something readers are familiar with
* Avoid contrasting "external" with "gc-scannable", which don't look like opposites but mostly are
* Not need more concrete info than the existing error message (that is, my message doesn't say that `string` is the gc type, but if we can do so easily, that would be cool)
* Preserve the info that's already in the message

I think the message proposed here is an improvement over the status quo, but I think we can do even better. Happy for more revisions on my suggestion.

rtjoa: Incorporated this suggestion, editing slightly to further avoid contrasting "external" with "gc-scannable":
```
Error: An unboxed product array element must be formed from all
       gc-ignorable types or all gc-scannable types. But this array
       operation is peformed for an array whose element type is
       #(string * float#), which is an unboxed product that contains both
       a must-scan type and a non-scannable type of sort float64.
```

rtjoa: On second thought, I actually think we might want to mention "external," because a common way to fix this error is to annotate a type as `mod external` (especially now that non-values are no longer `mod everything` by default).

Here's the latest version:
```
Error: An unboxed product array element must be formed from all
       gc-ignorable types or all gc-scannable types. But this array
       operation is peformed for an array whose element type is
       #(string * float#), which is an unboxed product that cannot be ignored
       (as it is not [external]) and contains the non-scannable sort float64.
```

goldfirere: I see your point about `external`. OK.

Do we need to use the word "sort"? I think we don't otherwise use this. I think we can use "layout" there.

Also, the product doesn't contain `float64`. It contains a type whose layout is `float64`.

So maybe

```
Error: An unboxed product array element must be formed from all
       gc-ignorable types or all gc-scannable types. But this array
       operation is peformed for an array whose element type is
       #(string * float#), which is an unboxed product that cannot be ignored
       (as it is not [external]) but contains a type with non-scannable layout float64.
```

?

testsuite/tests/typing-layouts-products/product_arrays.ml
```diff
@@ -124,9 +124,12 @@ let f_bad (x : #(string * float#)) = make_vect 42 x
 Line 1, characters 37-51:
 1 | let f_bad (x : #(string * float#)) = make_vect 42 x
                                          ^^^^^^^^^^^^^^
-Error: Unboxed product array elements must be external or contain all gc
-       scannable types. The product type this function is applied at is
-       not external but contains an element of sort float64.
+Error: An unboxed product array element must be formed from all
+       gc-ignorable types or all gc-scannable types. But this array
```
goldfirere:
```suggestion
       external types (which are ignored by the gc) or all gc-scannable types. But this array
```

@ccasin might like this use of "external" more. But I'm fine with this as-is, also.

goldfirere (approved):

goldfirere (approved):

mshinwell:
The whole point of my original comment here was that "external" seems confusing.  I'd like to discuss this further.

PR #4106, opened 2025-06-06
Add OxCaml Template Docs

nmatschke:
Add `ppx_template` documentation for OxCaml.

### DIFF
--- a/jane/doc/extensions/_09-templates/01-intro.md
+++ b/jane/doc/extensions/_09-templates/01-intro.md
@@ -44,10 +44,12 @@ let id : 'a. 'a @ global -> 'a @ global = fun x -> x
 
 The implementation of these two functions is obviously identical, so it would be nice if
 we could just write it once, but get two copies of the function with different modes.
-Conveniently, we've written a preprocessor to do exactly that: `ppx_template`. Much like
-templates in C++, this allows us to define the function once, polymorphic over some mode
-variable, while the compiler will instantiate the template for each value of the variable.
-Thus our two identity functions can be written as:
+Conveniently, we've written a 
+[preprocessor](https://dune.readthedocs.io/en/stable/reference/preprocessing-spec.html#using-ppx-rewriters) 
+to do exactly that: `ppx_template`. Much like templates in C++, this allows us to define 
+the function once, polymorphic over some mode variable, while the compiler will 
+instantiate the template for each value of the variable. Thus our two identity functions 
+can be written as:
 
 ```ocaml
 let%template[@mode m = (global, local)] id : 'a. 'a @ m -> 'a @ m = fun x -> x
--- a/jane/doc/extensions/_09-templates/01-intro.md
+++ b/jane/doc/extensions/_09-templates/01-intro.md
@@ -8,11 +8,11 @@ title: Intro
 
 ## Polymorphism and its Limitations
 
-Stack allocation, unboxed types, and soon multicore are exciting new(ish) features of this
-language we love, and each in some way depend on the _mode_ and/or _kind_ systems the
-OxCaml team has developed. However, there is still much work to be done, and while these
-systems will eventually be comparable in their expressivity to the type system, they are
-at present notably constrained in their capacity for polymorphism.
+OxCaml comes with _mode_ and _kind_ systems which enable many of its key features, 
+including stack allocation, unboxed types, and data-race-free parallelism. However, there 
+is still work to be done, and while these systems will eventually be comparable in their
+expressivity to the type system, they are at present notably constrained in their capacity
+for polymorphism.
 
 For example, consider the identity function:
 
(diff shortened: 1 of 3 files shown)


jane/doc/extensions/_09-templates/01-intro.md
```diff
@@ -0,0 +1,200 @@
...
+
+# OCaml with Templates
+
+## Polymorphism and its Limitations
+
+Stack allocation, unboxed types, and soon multicore are exciting new(ish) features of this
+language we love, and each in some way depend on the _mode_ and/or _kind_ systems the
+OxCaml team has developed. However, there is still much work to be done, and while these
+systems will eventually be comparable in their expressivity to the type system, they are
+at present notably constrained in their capacity for polymorphism.
```
goldfirere: This para doesn't quite make sense in this new context -- too time-specific

jane/doc/extensions/_09-templates/01-intro.md
```diff
@@ -0,0 +1,200 @@
...
+For example, consider the identity function:
+
+```ocaml
+let id : 'a. 'a -> 'a = fun x -> x
+```
+
+In upstream OCaml, this is about as polymorphic as you can get! I can call `id x` on any
+value `x`, regardless of its type, and get the same value back out.
+
+But in `OxCaml`, things are more complex. If my value is stack-allocated, then it will be
```
goldfirere:
```suggestion
But in OxCaml, things are more complex. If my value is stack-allocated, then it will be
```

jane/doc/extensions/_09-templates/01-intro.md
```diff
@@ -0,0 +1,200 @@
...
+
+```ocaml
+let id : 'a. 'a @ global -> 'a @ global = fun x -> x
+```
+
+## Templates
+
+The implementation of these two functions is obviously identical, so it would be nice if
+we could just write it once, but get two copies of the function with different modes.
+Conveniently, we've written a preprocessor to do exactly that: `ppx_template`. Much like
```
goldfirere: How do users invoke this? I know how we do it here at Jane Street, but how would users do this outside of Jane Street? Maybe @dkalinichenko-js can help here?

jane/doc/extensions/_09-templates/02-reference.md
```diff
@@ -0,0 +1,739 @@
...
+     mangle the name, but makes it clear that the alloc variable isn't used.
+     Side note: single [_] doesn't work due to how the syntax is implemented. *)
+end
+]
+```
+
+As with mode-polymorphism, the `[@@zero_alloc_if_stack a args...]` attribute conditionally
+attaches a `[@@zero_alloc args...]` attribute to a syntax node when `a = stack`. This is the
+preferred attribute over `[@@zero_alloc_if_local]` as conditionally being zero-alloc is
+a property of allocation behavior, not of modes.
```
goldfirere: We don't have `zero_alloc` documented elsewhere. Maybe strike this bit?

jane/doc/extensions/_09-templates/02-reference.md
```diff
@@ -0,0 +1,739 @@
...
+```ocaml
+module%template.portable [@modality p] F (T1 : S1) (T2 : S2) : S3 = struct
+  include Other_f1 [@modality p] (T1)
+  include Other_f2 [@modality p] (T2)
+end
+```
+
+## Mangling
+
+**WE DO NOT GUARANTEE STABILITY OF THE MANGLING ALGORITHM AT THIS TIME.**
```
goldfirere:
```suggestion
**WE DO NOT GUARANTEE STABILITY OF THE MANGLING ALGORITHM.**
```

goldfirere (approved):
Spotted a few small things to change.

jane/doc/extensions/_09-templates/01-intro.md
nmatschke: Reworded - what do you think?

jane/doc/extensions/_09-templates/01-intro.md
nmatschke: Oops, typo.

jane/doc/extensions/_09-templates/01-intro.md
nmatschke: I think this is covered by the `Instantiation` section below?

jane/doc/extensions/_09-templates/02-reference.md
nmatschke: Done.

jane/doc/extensions/_09-templates/02-reference.md
nmatschke: Done.

jane/doc/extensions/_09-templates/01-intro.md
goldfirere: I mean in a `dune` file.

nmatschke: I added a link to the dune documentation. I would add a link to the opam package too, but we haven't made it public yet.

goldfirere (approved):

PR #4107, opened 2025-06-07
runtime: add macros to handle `or_null` values alike those available for `option`

NicolasT:
A first commit adds `Val_this`, `This_val`, `Is_null` and `Is_this` (next to the existing `Val_null`) to make the API to work with `or_null` values alike the API available to work with `option` values. Granted, `Val_this` and `This_val` are essentially "the identity macro", but I believe it to be useful to "document" the current (supposed) type of some `value` in C code.

A second commit uses these new macros in the runtime implementation. It uses `Is_null`, since that's the only applicable one, switching `== Val_null` comparisons with, well, `Is_null`. If this is deemed an unnecessary change, the commit can easily be dropped.

While doing so, I noticed there's some existing code calling

```
if (Is_long(v) || Is_null(v))
```

which, I believe, is a bit superfluous since `Is_long(v)` is true for `v == Val_null` already. I imagine a C compiler optimizing this away when the macro-version of `Is_long` is used, but maybe not when using the `inline __asm__` version (though I didn't check).

### DIFF
--- a/runtime/caml/mlvalues.h
+++ b/runtime/caml/mlvalues.h
@@ -69,8 +69,13 @@ typedef opcode_t * code_t;
 
 #include "domain_state.h"
 
-/* The null pointer value. */
+/* Or_null constructors. */
+
 #define Val_null ((value) 0)
+#define Val_this(v) (v)
+#define This_val(v) (v)
+#define Is_null(v) ((v) == Val_null)
+#define Is_this(v) ((v) != Val_null)
 
 /* Longs vs blocks. */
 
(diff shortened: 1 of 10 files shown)


NickBarnes:
That case of `Is_long(v) || Is_null(v)` is interesting. I'd prefer to have a more clearly-named test (`Is_imm` ?) to test for "long or null": the name as it stands is misleading. But then many (all?) of our `Is_long` instances would have to become `Is_imm`. Sigh.

PR #4108, opened 2025-06-09
Allow --enable-multidomain without --enable-stack-checks

mshinwell:
After discussion with @lpw25 we'd like to relax the restrictions here for now.


lpw25 (approved):
LGTM

PR #4109, opened 2025-06-10
Add externals information to .cms files

spiessimon:
This PR adds information about the external declarations in a file to the `.cms` files. This allows one to use the externals extraction tool in the compiler not just on `.cmt` files (which require `-bin-annot` when compiling) but also on the smaller `.cms` files.


extract_externals/extract_externals.ml
```diff
@@ -102,31 +100,57 @@ let extract_shapes_from_cmt ~verbose file =
...
-      if not (String.ends_with file ~suffix:".cmt")
-      then Misc.fatal_errorf "File %s is not a .cmt file; aborting\n" file)
-    files;
-  List.concat_map (extract_shapes_from_cmt ~verbose) files
+  let files =
+    List.map
+      (fun file ->
+        match file with
+        | _ when String.ends_with file ~suffix:".cms" -> CMS file
+        | _ when String.ends_with file ~suffix:".cmt" -> CMT file
```
gretay-js: Consider using `Filename.extension`

extract_externals/extract_externals.ml
```diff
@@ -102,31 +100,57 @@ let extract_shapes_from_cmt ~verbose file =
...
+  | { cms_externals; _ } -> Array.to_list cms_externals
+
+type externals_source_file =
+  | CMT of string
+  | CMS of string
+
+let extract_shapes_from_file ~verbose file =
+  match file with
+  | CMS file -> extract_shapes_from_cms ~verbose file
+  | CMT file -> extract_shapes_from_cmt ~verbose file
```
gretay-js, line 125: Are you plannig to use this type for anything else?

gretay-js (approved):
It's be nice to keep formatting enabled on the new `vicuna_*` files: add to `typing/.ocamlformat-enable`.

extract_externals/extract_externals.ml
spiessimon: Ah, thanks for the pointer to `Filename`. I used `Filename.check_suffix` now, since it seems to consider upper/lowercase file extensions on different systems.

extract_externals/extract_externals.ml
spiessimon: No, it could just as well have been a polymorphic variant. I just wanted something to avoid confusing .cms and .cmt files.

PR #4111, opened 2025-06-10
Add documentation of zero alloc and probe

gretay-js:
Added under misc extentions rather than a new folder because this is not of interest for typical users.

### DIFF
--- a/jane/doc/extensions/_08-miscellaneous-extensions/zero_alloc_check.md
+++ b/jane/doc/extensions/_08-miscellaneous-extensions/zero_alloc_check.md
@@ -477,17 +477,16 @@ error message will show all functions that failed the check in the same compilat
 
 The following settings are mostly for testing and debugging of the analysis.
 
-- `-zero-alloc-check` compiler flag and `[@@@zero_alloc check*]` attributes
 The check can be controlled by the compiler flag `-zero-alloc-check` with the following
 arguments:
 
 * `default` : check attributes without `opt` payload (default)
 * `check all` : covers both `opt` and default payloads and is intended to be used for
-      optimized builds `CHECK_ZERO_ALLOC_OPT` environment variable above.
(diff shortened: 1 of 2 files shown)


jane/doc/extensions/_08-miscellaneous-extensions/zero_alloc_check.md
```diff
@@ -0,0 +1,623 @@
...
+for situations of unexpectedly high overhead of the analysis on
+specific files, for example due to precise analysis of recursive
+functions. It will cause all uses of these functions to be
+conservatively treated and likely fail the check.
+`-zero-alloc-checker-join` compilation flag provides a more fine-grained
+control of precision.
+
+## Annotation `[@@@zero_alloc all]` and `[@zero_alloc ignore]`
+
+Some files consists mostly of non-allocating code. To reduce the annotation burden, we
```
ccasin:
```suggestion
Some files consist mostly of non-allocating code. To reduce the annotation burden, we
```

jane/doc/extensions/_08-miscellaneous-extensions/zero_alloc_check.md
```diff
@@ -0,0 +1,623 @@
...
+
+When the check fails, it is a compile error (not a warning) and compilation stops. The
+error message will show all functions that failed the check in the same compilation unit
+(not just the first function that failed the check).
+
+## Other ways to control the analysis
+
+The following settings are mostly for testing and debugging of the analysis.
+
+- `-zero-alloc-check` compiler flag and `[@@@zero_alloc check*]` attributes
```
ccasin: It looks like this nested list isn't rendering correctly.  Probably it can be fixed with indendation, or alternatively you can just ditch the outer list, since it only has one element anyway (and probably slightly rephrase it as a complete sentence)

PR #4112, opened 2025-06-10
Better treatment of relations in flambda2 types 

bclement-ocp:
We currently deal with relational information through two (mostly,
because the CSE environment adds backwards information to the typing
environment) separate mechanisms:

 - The CSE environment keeps track of a map from arguments to values and
   effectively implements congruence closure over relations (i.e. it
   ensures that if `Get_tag(x) = y` and `Get_tag(x) = z` then `y = z`).

 - The typing env keeps track of a map from values to arguments though
   the relational types (if we have `Get_tag(x) = y` then `y` has type
   `(= Get_tag x)`)

This is not ideal, for at least two reasons: the CSE environment
and the canonicals in the typing env can get out of sync, causing missed
optimisation opportunities such as https://github.com/oxcaml/oxcaml/issues/3181; and the representation of
relational types is restrictive (see e.g. https://github.com/oxcaml/oxcaml/pull/2259, which this is can be
considered a rewrite of).

This patch introduces a new module in the flambda2 typing environment,
the "types database". This module is intended to provide a single and
principled source of relational informations between *canonical
elements* (or equivalently of non-alias relational information). The
design is loosely inspired from (a simplified version of) the
description in https://github.com/oxcaml/oxcaml/issues/3219. The types database internally maintains the same
maps that the CSE environment and the typing environment maintained
previously, except that it is wholly inside the typing environment now.
In particular, the typing environment can make sure that the canonicals
cannot get out of sync as with the CSE environment, and we also have
access to the more precise n-way join when computing the join of the
types database.

Having a single source of truth for relational information, instead of
it being distributed across the CSE environment and the typing
environment, also makes it easier to implement a robust heuristic for
match-in-match, which will come in a follow-up patch

More precisely, the current patch implements the following:

 - The types database itself, with support for `Is_int`, `Get_tag` and
   `Is_null` relations ;
 - Integration in the typing env in order to properly maintain
   canonicals ;
 - Join of databases with the new n-way join algorithm

Caveats:

 - This is currently gated behind the `-flambda2-types-database`
   command-line option or the `flambda2-types-database=1` OCAMLPARAM.
   This is done in order to facilitate testing, and we need to discuss
   during review whether to keep this flag or make it the default.

 - Enabling the types database currently enables the relational
   primitives but *does not* disable CSE for the corresponding
   primitives. This is because there are places (e.g. simplification of
   switch expressions) where we look up directly in the CSE environment,
   but these should be easy to replace with a lookup in the types
   database (it was not implemented in order to avoid multiplying the
   code paths, but will be implemented before merging).

 - Join is only implemented for the new n-way join algorithm.
   Implementing the join properly for the old binary join algorithm
   should not be complicated but requires a bit of thought and I'd like
   to wait until at least the first round of review first.

 - Support for env extension is currently partial. We discussed with
   @lthls on using identifiers for extensions in order to avoid
   potential infinite loops during meet, and this enables to store
   arbitrary information on extensions. Currently, the implementation is
   incorporating a database extension inside the `typing_env_extension`
   type, and allows attaching a full `typing_env_extension` to the
   database extension points. I believe that the first part is not
   necessary, and there are multiple places where database information
   is simply skipped in env extensions. Depending on the first round(s)
   of review, I'd be happy to either (a) keep things as is, or (b) more
   aggressively store database information in extensions (e.g. after
   a join) or (c) don't have database information in the
   `typing_env_extension` type at all (I personally favor (c); I believe
   this only has an effect on the join in practice and that we don't
   need to join such deep extensions).

 - Although the database supports `Untag_imm` and `Tag_imm` relations,
   they are not exposed through the typing env or used during
   simplification. This is in order to maintain feature parity with the
   existing implementation and facilitate comparisons; I plan on
   enabling them in a follow-up PR.

Note: The implementation relies heavily on the shared patricia tree
functions introduced in https://github.com/oxcaml/oxcaml/pull/4073, and is rebased on https://github.com/oxcaml/oxcaml/pull/4073; changes to the
`middle_end/flambda2/algorithms/` directory in this PR do not need to be
reviewed here.

### DIFF
--- a/middle_end/flambda2/types/env/database.ml
+++ b/middle_end/flambda2/types/env/database.ml
@@ -1,0 +1,1186 @@
+module RWC = Reg_width_const
+
+(** {2 Meet}
+
+    We use a simplified version of [TE.meet_return_value] for the database: we
+    only care about whether the result of a [meet] is identical to the value
+    that was previously stored in the database, or if it needs to be updated.
+
+    We choose a left-biased approach: the existing value should always be passed
+    as the first argument of the [meet] function. *)
+
+type 'a left_meet_return_value =
+  | Left_input
+  | New_result of 'a
+
+let left_meet_return_value res left =
+  if res == left then Left_input else New_result res
+
+let extract_value res left =
+  match res with Left_input -> left | New_result value -> value
+
+(** {2 Row-like}
+
+    Generic row-like functor, used for disjunctions. *)
+
+module Row_like = struct
+  module Make (Branch : Container_types.S) = struct
+    type 'a t =
+      { known : 'a Branch.Map.t;
+        other : 'a Or_bottom.t
+      }
+
+    let create ?default known =
+      let other : _ Or_bottom.t =
+        match default with None -> Bottom | Some other -> Ok other
+      in
+      { known; other }
+
+    let print_known pp ppf known = Branch.Map.print pp ppf known
+
+    let print pp ppf { known; other } =
+      Format.fprintf ppf
+        "@[<hov 1>(@[<hov 1>(known@ %a)@]@ @[<hov 1>(other@ %a)@]@]"
+        (print_known pp) known (Or_bottom.print pp) other
+
+    let is_bottom { other; known } =
+      match other with Bottom -> Branch.Map.is_empty known | Ok _ -> false
+
+    let merge f { known = known1; other = other1 }
+        { known = known2; other = other2 } =
+      let known =
+        Branch.Map.merge
+          (fun _ case1 case2 ->
+            let case1 : _ Or_bottom.t =
+              match case1 with None -> other1 | Some case1 -> Ok case1
+            in
+            let case2 : _ Or_bottom.t =
+              match case2 with None -> other2 | Some case2 -> Ok case2
+            in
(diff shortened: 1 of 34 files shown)


middle_end/flambda2/types/env/database.ml
```diff
@@ -0,0 +1,1186 @@
...
+        Branch.Map.merge
+          (fun _ case1 case2 ->
+            let case1 : _ Or_bottom.t =
+              match case1 with None -> other1 | Some case1 -> Ok case1
+            in
+            let case2 : _ Or_bottom.t =
+              match case2 with None -> other2 | Some case2 -> Ok case2
+            in
+            match (f case1 case2 : _ Or_bottom.t) with
+            | Bottom -> None
```
Ekdohibs: It seems to me that in the case where `f other1 other2` is not bottom, this loses information.

bclement-ocp: Ah, good catch — I initially replicated the behavior of the `meet` on types and noticed this but then forgot about it. I agree, it makes sense to check `other` here.

Gbury:
I think that this PR should add (or change) one run of CI to actually test the new code.

PR #4113, opened 2025-06-10
Fix Rec_info for imported classic mode approximations

mshinwell:
Functions generated in classic mode are currently failing to inline at higher optimization levels unless there is an attribute at the call site.  This is because the rec-info expression that has to be constructed (in order to form a full Flambda 2 type) from the classic mode closure approximation is currently set to "unknown", which translates to "infinity", namely prevent all inlining.

@lukemaurer does this look correct to you?


lukemaurer (approved):
I believe it's correct, though a bit fragile: it assumes that this function continues to be only used from `create_from_closure_conversion_approx` and not with any symbol that could have arisen from unrolling.

PR #4117, opened 2025-06-10
Fix soundness bug in `Stdlib.Atomic`

mdelvecchio-jsc:
Argument for correctness: We consider an atomic to live in a capsule, and for that capsule to have uncontended access to it. If someone has the value `@ contended`, then they are either inside that capsule and just have a weakened reference to it, or are in another capsule and interacting across a capsule boundary.

For `Contended.get`, the contents of the atomic are crossing a capsule boundary, so we need to return it `@ contended`. (This is better than requiring `'a : value mod contended` since one can clearly implement the other but vice versa requires changing the type of the `'a` in `'a t` in a way that's not always desired.) We don't need to worry about portability here because the only way to get `'a` in another capsule is if `'a t` were in another capsule, and `'a t` is only portable if `'a` crosses portability.

For `Contended.set`, we are sending a value across a capsule boundary to the one that has the atomic, so the set value needs to be `@ portable`, but doesn't need to cross portability. It does need to cross contention, however, because after it crosses the capsule boundary, it needs to be `@ uncontended` to put it in the atomic.

The other types are derived from these, leaving out unnecessary `@ contended` annotations when `value mod contended` is required.

### DIFF
--- a/stdlib/atomic.ml
+++ b/stdlib/atomic.ml
@@ -42,19 +42,19 @@ let incr r = add r 1
 let decr r = sub r 1
 
 module Contended = struct
-  external get : ('a : value_or_null mod contended).
-    'a t @ contended local -> 'a @@ portable =
+  external get : ('a : value_or_null).
+    'a t @ contended local -> 'a @ contended @@ portable =
     "%atomic_load"
-  external set : ('a : value_or_null mod portable).
-    'a t @ contended local -> 'a -> unit @@ portable =
+  external set : ('a : value_or_null mod contended).
+    'a t @ contended local -> 'a @ portable -> unit @@ portable =
     "%atomic_set"
-  external exchange : ('a : value_or_null mod contended portable).
-    'a t @ contended local -> 'a -> 'a @@ portable =
+  external exchange : ('a : value_or_null mod contended).
+    'a t @ contended local -> 'a @ portable -> 'a @@ portable =
     "%atomic_exchange"
-  external compare_and_set : ('a : value_or_null mod portable).
-    'a t @ contended local -> 'a -> 'a -> bool @@ portable =
+  external compare_and_set : ('a : value_or_null mod contended).
+    'a t @ contended local -> 'a -> 'a @ portable -> bool @@ portable =
     "%atomic_cas"
-  external compare_exchange : ('a : value_or_null mod contended portable).
-    'a t @ contended local -> 'a -> 'a -> 'a @@ portable =
+  external compare_exchange : ('a : value_or_null mod contended).
+    'a t @ contended local -> 'a -> 'a @ portable -> 'a @ contended @@ portable =
     "%atomic_compare_exchange"
 end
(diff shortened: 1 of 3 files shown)


stdlib/atomic.ml
```diff
@@ -14,47 +14,121 @@
...
+  = "%atomic_cas"
+
+external compare_exchange
+  : ('a : value_or_null).
+  'a t @ local -> 'a -> 'a -> 'a
+  @@ portable
+  = "%atomic_compare_exchange"
+
+external fetch_and_add
+  :  int t @ contended local
```
TheNumbat, line 60: It's a bit strange that these allow `contended` when the other contended functions are relegated to the module, but we don't need two versions since int crosses portability and contention...I guess it's fine

stdlib/atomic.ml
```diff
@@ -14,47 +14,121 @@
...
-    'a t @ contended local -> 'a -> 'a @@ portable =
-    "%atomic_exchange"
-  external compare_and_set : ('a : value_or_null mod portable).
-    'a t @ contended local -> 'a -> 'a -> bool @@ portable =
-    "%atomic_cas"
-  external compare_exchange : ('a : value_or_null mod contended portable).
-    'a t @ contended local -> 'a -> 'a -> 'a @@ portable =
-    "%atomic_compare_exchange"
+  external get
+    : ('a : value_or_null).
```
TheNumbat: I agree this is already safe because the only way to get a portable atomic is via crossing (I think they're equivalent to one-element parallel arrays now), but it might be easier to understand if we write `mod portable` on read, `mod contended` on write, and `mod portable contended` on exchange.

- `get` returns a contended value since it's crossing a capsule boundary. But it also requires future users of the atomic only write portable things into it. Hence `'a` must cross portability.
- `set` requires a portable argument since it's crossing a capsule boundary. But it also requires past users of the atomic to get the value out at contended. Hence `'a` must cross contention.
- `exchange` does both

So I like to think of the crossing requirement as applying the mode axis' constraint on the opposite co/contravariant position / the "wrong" temporal direction as the mode. (This actually came up for the same reason when I wrote unsound to/of_array for parallel arrays :)

TheNumbat (approved):

stdlib/atomic.ml
mdelvecchio-jsc: I agree that's easier to understand (that's what I actually originally wrote), but I haven't convinced myself that there's no occasions where what you suggested is too weak. I also can't come up with any examples where it is too weak, though. For now, I'd rather keep the one that is obviously at least as strong for at least the thing in stdlib since, if we realize later it is too weak, I'd like to avoid having to make another PR.

TheNumbat: Weak as in not expressive enough? Sure, please add a comment explaining this though

mshinwell:
@mdelvecchio-jsc what's the status of this PR?

stdlib/atomic.ml
mdelvecchio-jsc: Nevermind, let's just do the simpler more clearly correct thing for now. If something does come up that demonstrates this version to be not expressive enough, then we can change it then.

TheNumbat (approved):

PR #4118, opened 2025-06-11
Dedicated engine for Cmm Peephole optimisations

lthls:
This draft PR is a proposal for handling phantom lets (or debug info in general) in Cmm without interfering with the various existing peephole optimisations.

Here I propose to rewrite the existing optimisations using a dedicated DSL, provided by the `Cmm_peephole_engine` module, then in a second phase improve the engine itself to handle phantom lets (and maybe regular lets, if we want to).

Right now (at time of PR creation), the `add_int` Cmm helper function has been rewritten to use this engine.

I am going to wait for feedback before committing fully to this approach (as alternative proposals have also been put forward).

### DIFF
--- a/backend/cmm_helpers.ml
+++ b/backend/cmm_helpers.ml
@@ -19,6 +19,9 @@
    Int_replace_polymorphic_compare *)
 module V = Backend_var
 module VP = Backend_var.With_provenance
+module P = Cmm_peephole_engine
+open P.Syntax
+open P.Default_variables
 open Cmm
 open Arch
 
@@ -393,6 +396,17 @@ let rec add_int c1 c2 dbg =
       | c1, Cop (Caddi, [c2; Cconst_int (n2, _)], _) ->
         add_const (add_int c1 c2 dbg) n2 dbg
       | _, _ -> Cop (Caddi, [c1; c2], dbg))
+
+let rec add_int' arg1 arg2 dbg =
+  let res = Cop (Caddi, [arg1; arg2], dbg) in
+  P.run res [
+    Add (Const_int_var i, Var c), (fun e -> add_const e#.c e#.i dbg);
+    Add (Var c, Const_int_var i), (fun e -> add_const e#.c e#.i dbg);
+    Add (Add (Var c1, Const_int_var i1), Var c2), (fun e -> add_const (add_int' e#.c1 e#.c2 dbg) e#.i1 dbg);
+    Add (Var c1, Add (Var c2, Const_int_var i2)), (fun e -> add_const (add_int' e#.c1 e#.c2 dbg) e#.i2 dbg);
+  ]
+
+let add_int = if false then add_int else add_int'
 
 let rec sub_int c1 c2 dbg =
   map_tail2 c1 c2 ~f:(fun c1 c2 ->
--- a/backend/cmm_peephole_engine.ml
+++ b/backend/cmm_peephole_engine.ml
@@ -1,0 +1,143 @@
+(* Variable kinds *)
+
+type _ pattern_kind =
+| Expr : Cmm.expression pattern_kind
+| Int : int pattern_kind
+| Natint : Nativeint.t pattern_kind
+
+type 'a pattern_var =
+{ id : int;
+  name : string;
+  kind : 'a pattern_kind;
+}
+
+let var_counter = ref 0
+
+let create_var kind name =
+  incr var_counter;
+  { id = !var_counter;
+    name;
+    kind;
+  }
+
+module IM = Numbers.Int.Map
+
+module Env : sig
+  type t
+  val empty : t
+  val add : t -> 'a pattern_var -> 'a -> t
+  val find_exn : t -> 'a pattern_var -> 'a
+end = struct
+  type t = {
+    exprs : Cmm.expression IM.t;
+    ints : int IM.t;
+    natints : Nativeint.t IM.t;
+  }
+  let empty = {
+    exprs = IM.empty;
+    ints = IM.empty;
+    natints = IM.empty;
+  }
+  let add (type a) env (var : a pattern_var) (expr : a) =
+    match var.kind with
+    | Expr ->
+        if IM.mem var.id env.exprs
+        then Misc.fatal_errorf "Duplicate binding for var %s" var.name
+        else { env with exprs = IM.add var.id expr env.exprs }
+    | Int ->
+        if IM.mem var.id env.ints
+        then Misc.fatal_errorf "Duplicate binding for var %s" var.name
+        else { env with ints = IM.add var.id expr env.ints }
+    | Natint ->
+        if IM.mem var.id env.natints
+        then Misc.fatal_errorf "Duplicate binding for var %s" var.name
+        else { env with natints = IM.add var.id expr env.natints }
+  let find_exn (type a) env (var : a pattern_var) : a =
+    match var.kind with
+    | Expr -> IM.find var.id env.exprs
+    | Int -> IM.find var.id env.ints
+    | Natint -> IM.find var.id env.natints
+end
+
+type cmm_pattern =
+  | Any
+  | Var of Cmm.expression pattern_var
+  | Const_int_fixed of int
+  | Const_int_var of int pattern_var
+  | Const_natint_fixed of Nativeint.t
+  | Const_natint_var of Nativeint.t pattern_var
+  | Add of cmm_pattern * cmm_pattern
+  | When of cmm_pattern * (Env.t -> bool)
+
+type 'a clause =
+  cmm_pattern * (Env.t -> 'a)
+
+let match_clauses_in_order clauses expr =
+  let rec match_one_pattern env pat (expr : Cmm.expression) =
+    match pat with
+    | Any -> Some env
+    | Var v -> Some (Env.add env v expr)
+    | Const_int_fixed n1 -> begin
+        match expr with
+        | Cconst_int (n2, _) -> if Int.equal n1 n2 then Some env else None
+        | _ -> None
+      end
+    | Const_int_var v -> begin
+        match expr with
+        | Cconst_int (n, _) -> Some (Env.add env v n)
+        | _ -> None
+      end
+    | Const_natint_fixed n1 -> begin
+        match expr with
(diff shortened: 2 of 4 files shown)


backend/cmm_helpers.ml
```diff
@@ -441,6 +521,24 @@ let rec max_signed_bit_length e =
...
 
+let rec max_signed_bit_length' e =
+  P.run_default ~default:(fun _ -> arch_bits) (prefer_or e) [
+    Binop (Comparison, Any c1, Any c2) => (fun _e -> 1);
+    When (Binop (And, Any c, Const_int n), (fun e -> e#.n > 0))
+    => (fun e -> 1 + Misc.log2 e#.n);
+    When (Binop (Lsl, Any c, Const_int n), is_defined_shift' n)
+    => (fun e -> Int.min arch_bits (max_signed_bit_length' e#.c + e#.n));
+    When (Binop (Asr, Any c, Const_int n), is_defined_shift' n)
+    => (fun e -> Int.min arch_bits (max_signed_bit_length' e#.c - e#.n));
```
Ekdohibs:
```suggestion
    => (fun e -> Int.max 0 (max_signed_bit_length' e#.c - e#.n));
```

lthls: Thanks

backend/cmm_helpers.ml
```diff
@@ -324,6 +354,8 @@ let add_no_overflow n x c dbg =
 
 let is_defined_shift n = 0 <= n && n < arch_bits
 
+let is_defined_shift' n e = is_defined_shift e#.n
```
Gbury: I'd argue for naming the env variables `env`, or really anything else than `e` since `e` is already used in quite a few places in this file to denote a cmm expression, and it always confuses me for a while when reading this code where `e` is instead an `env` (or really a substitution resulting from some pattern matching, maybs `subst` ? although that's longer).

backend/cmm_helpers.ml
```diff
@@ -19,9 +19,39 @@
    Int_replace_polymorphic_compare *)
 module V = Backend_var
 module VP = Backend_var.With_provenance
+module P = Cmm_peephole_engine
+open P.Syntax
+open P.Default_variables
 open Cmm
 open Arch
 
+let check_equal_1 name f1 f2 arg1 =
```
Gbury, line 51: These function should at least add some type annotation to ensure that they are only used to compare cmm `expressions`s which are (currently) safe(-ish) to compare with the polymorphic comparison, and a comment to state that this is fine because it's temporary to ensure that the new peephole optimisation engine is correct and complete (i.e. performs the same optimisations that are currently done).

backend/cmm_helpers.ml
```diff
@@ -19,9 +19,39 @@
    Int_replace_polymorphic_compare *)
 module V = Backend_var
 module VP = Backend_var.With_provenance
+module P = Cmm_peephole_engine
+open P.Syntax
+open P.Default_variables
```
Gbury: I'm a bit unsure about opening this module and shadowing identifiers that are used a lot throughout the file already (e.g. `c1`, `c2`, etc...). Maybe opening it locally for each use of peephole rules would be better for now ?

backend/cmm_helpers.ml
```diff
@@ -380,6 +412,35 @@ let rec add_const c n dbg =
           add_const c (n - x) dbg
         | _ -> Cop (Caddi, [c; Cconst_int (n, dbg)], dbg))
 
+let rec add_const' arg const dbg =
+  map_tail1 arg ~f:(fun arg ->
+      let res = Cop (Caddi, [prefer_add arg; Cconst_int (const, dbg)], dbg) in
+      let x = P.create_var Int "x" in
+      P.run res
+        [ (Binop (Add, Any c, Const_int_fixed 0) => fun e -> e#.c);
+          ( When
```
Gbury: Maybe some kind of `when_` infix operator would be a bit more readable here than a Prefix variant constructor ?

backend/cmm_helpers.ml
```diff
@@ -380,6 +412,35 @@ let rec add_const c n dbg =
...
+              ( Binop (Add, Binop (Sub, Const_int x, Any c), Const_int n),
+                fun e -> Misc.no_overflow_add e#.n e#.x )
+          => fun e -> Cop (Csubi, [Cconst_int (e#.n + e#.x, dbg); e#.c], dbg) );
+          ( When
+              ( Binop (Add, Binop (Sub, Any c, Const_int x), Const_int n),
+                fun e -> Misc.no_overflow_sub e#.n e#.x )
+          => fun e -> add_const' e#.c (e#.n - e#.x) dbg ) ])
+
+let add_const = check_equal_3 "add_const" add_const add_const'
```
Gbury, line 497: maybe it's a baseless performance concern, but wouldn't it be better to eta-expand here ? (though I guess flambda2 may do it automatically ?)

backend/cmm_peephole_engine.mli
```diff
@@ -0,0 +1,60 @@
...
+type binop =
+  | Add
+  | Sub
+  | Lsl
+  | Lsr
+  | Asr
+  | Or
+  | And
+  | Comparison
+  | Bitwise_op
```
Gbury: I think that for future rules, we'll likely need more precise binops for comparisons and bitwise operations, what's your plan for then ? add new binops ?

backend/cmm_peephole_engine.ml
```diff
@@ -0,0 +1,187 @@
+(* Variable kinds *)
+
+type _ pattern_kind =
+  | Expr : Cmm.expression pattern_kind
+  | Int : int pattern_kind
+  | Natint : Nativeint.t pattern_kind
```
Gbury, line 43: Maybe having a `Pattern` submodule (and maybe others also ?) would probably be cleaner for most of the naming around pattern-related things ?

backend/cmm_peephole_engine.mli
```diff
@@ -0,0 +1,60 @@
...
+
+type 'a clause
+
+val run : Cmm.expression -> Cmm.expression clause list -> Cmm.expression
+
+val run_default :
+  default:(Cmm.expression -> 'a) -> Cmm.expression -> 'a clause list -> 'a
+
+module Syntax : sig
+  val ( => ) : cmm_pattern -> (Env.t -> 'a) -> 'a clause
```
Gbury, line 134: Could we expose a regular (aka non-infix) function to create a clause ? It's confusing reading the interface of this module and having to wait until the end to see the only way to construct a clause.

backend/cmm_peephole_engine.mli
```diff
@@ -0,0 +1,60 @@
...
+
+  val ( #. ) : Env.t -> 'a pattern_var -> 'a
```
Gbury, line 137: Same remark, exposing `Env.find_exn` would be good I think.

backend/cmm_peephole_engine.mli
```diff
@@ -0,0 +1,60 @@
+type _ pattern_kind =
+  | Expr : Cmm.expression pattern_kind
+  | Int : int pattern_kind
+  | Natint : Nativeint.t pattern_kind
```
Gbury, line 51: This module needs some doc comments, ^^

backend/cmm_peephole_engine.mli
lthls: Yes, those types are expanded as needed. `Bitwise_op` already subsumes `Or` and `And`; the constructors do not have to match distinct expressions.

backend/cmm_peephole_engine.mli
lthls: I prefer if the only way to construct a clause is through the `Syntax` module. Besides, the definition for `clause` is about 8 lines above the function to create a clause; I don't think you had to wait for very long.

Gbury: I really dislike the idea that **the only (reasonable) way** to create clauses is to open a module and then use an infix operator, but it's more a matter of style, so I won't fight too hard on that.

backend/cmm_helpers.ml
lthls: I don't think it matters. All backends (Closure, Flambda, flambda2) will see this as a partial application and eta-expand.

backend/cmm_peephole_engine.ml
lthls: I don't see how it would help. `Pattern.kind` doesn't strike me as more readable than `pattern_kind`, and almost all of the module is about patterns anyway.

Gbury: ok

lthls:
(I mean, I will wait for your review before rebasing; no need to rush)

backend/cmm_peephole_engine.ml
```diff
@@ -216,3 +216,113 @@ module Default_variables = struct
...
+  (* Checks equivalence of expressions. At the moment this ignores debuginfo and
+     phantom expressions. *)
+  let rec equivalent (x : Cmm.expression) (y : Cmm.expression) =
+    let module V = Backend_var in
+    let module VP = Backend_var.With_provenance in
+    match x, y with
+    | Cconst_int (n1, _), Cconst_int (n2, _) -> Int.equal n1 n2
+    | Cconst_natint (n1, _), Cconst_natint (n2, _) -> Nativeint.equal n1 n2
+    | Cconst_float32 (f1, _), Cconst_float32 (f2, _) -> Float.equal f1 f2
+    | Cconst_float (f1, _), Cconst_float (f2, _) -> Float.equal f1 f2
```
Gbury: Do we want simple float equality or bitwise equality here ?

lthls: Bitwise equality would be more robust, but given that we don't have peephole optimisations involving float constants in the engine at the moment I felt that regular float equality was acceptable.

Gbury: I'd argue for bitwise equality, if only to avoid errors in the future, particularly around `+ 0.` and `. 0.`, which might arise in optimization one day (and it doesn't require much more effort to do right now).

lthls: Feel free to send a PR :)

Gbury: done in https://github.com/lthls/flambda-backend/pull/4

PR #4121, opened 2025-06-11
Fix recursive modules' treatment of destructive substitution

rtjoa:


typing/typemod.ml
```diff
@@ -1163,6 +1167,48 @@ let rec approx_modtype env smty =
...
+      let md =
+        { md with
+          md_type = Mty_signature (remove_from_sg namespace path env sg) }
+      in
+      Some (Sig_module (id, presence, md, rs, priv))
+    | _ ->
+      Some item
+  in
+  (* Removing destructively-substituted items approximates [transl_with],
+     as all approximated types are treated as abstract. *)
```
ccasin: The obvious worry here is that we could get weird dangling references.  If you consider extending `S` in your first example such that it contains references to the module `A` that's about to be removed, do we end up getting a bad approximate signature with references to items that no longer exist, when we approximate `S with A := ...`?

I think the answer is no: the comment above `approx_modtype` says that the only things it keeps are type, module, and module type elements.  Since all the types are made abstract, as you say, there is no place in the approximated signature for a dangling reference to `A` to appear.

Can you flesh out this comment to explain in more detail why this is OK?

typing/typemod.ml
```diff
@@ -1163,6 +1167,48 @@ let rec approx_modtype env smty =
...
+  match constr with
+  | Pwith_typesubst (lid, _) ->
+    remove_from_sg Type (Longident.flatten lid.txt) env sg
+  | Pwith_modsubst (lid, _) ->
+    remove_from_sg Module (Longident.flatten lid.txt) env sg
+  | Pwith_modtypesubst (lid, _) ->
+    remove_from_sg Module_type (Longident.flatten lid.txt) env sg
+  | _ ->
```
ccasin: Please make this an or-pattern that lists all the remaining cases, so we're forced to revisit if we add more `with` clauses (which we will, soon, for abstract kinds).

ccasin (approved):

typing/typemod.ml
rtjoa: I looked further into this, and found it actually is possible to get a dangling reference in the case of destructive module substitution. See 17841aa98d6d0d0120d9ab5794a149855084e98f for a regression test of this and the fix (which is to actually do the substitution). Thanks for highlighting this.

ccasin:
> May I request some time to review this PR? At first glance, I tend to prefer the approach in [ocaml/ocaml#13829](https://github.com/ocaml/ocaml/pull/13829) which fixes a different problem with `with` constraints by reusing the main `merge_constraint` on approximation to avoid creating invalid module type.

Thanks for pointing out that PR, @Octachron!  We didn't realize this had already been fixed upstream (I guess we tested on an old compiler).  We definitely considered the approach of using the main merge_constraint, and don't have any particular objections to it.

We don't want to diverge from upstream here, so I think we should try to take that solution if it works.  @rtjoa  can you take a look at that PR and check if it also handles the layouts version of the problem (and whether a small change will make it so, of not)?

typing/typemod.ml
```diff
@@ -1163,6 +1167,100 @@ let rec approx_modtype env smty =
...
+    | Some path ->
+      let path', _, _ = Env.lookup_module ~loc l'.txt env in
+      let sub = Subst.add_module_path path path' Subst.identity in
+      Subst.signature Make_local sub sg
+    end
+  | Pwith_modtypesubst (l, _) ->
+    let _replaced_path, sg =
+      remove_from_sg Module_type (Longident.flatten l.txt) env sg
+    in
+    sg
```
Octachron: Removing the abstract module type is not sufficient, consider:
```ocaml
module type P = sig
  module type T
  module A:T
end

module rec X: P with module type T := sig type t end = struct
  module A = struct type t end
end
and Y : sig type t = X.A.t end = struct
   type t = X.A.t
end
```

rtjoa: Thanks - fixed this in 3d4f5a94c198574d8e479685ddb72a5156b78fb0.

typing/typemod.ml
```diff
@@ -1163,6 +1167,103 @@ let rec approx_modtype env smty =
...
+    in
+    replaced_path,
+    List.filter_map Fun.id filtered_sg
+  and remove_from_sig_item namespace path env sg_for_env item =
+    match path, namespace, item with
+    | [s], Type, Sig_type (id, _, _, _)
+    | [s], Module, Sig_module (id, _, _, _, _)
+    | [s], Module_type, Sig_modtype (id, _, _)
+        when Ident.name id = s ->
+      Some (Pident id), None
```
Octachron, line 1219: Note that in the `Type` case, this can create invalid module types by removing a ghost companion item. 
I am not sure if it can be exploited in the context of recursive modules, but this can lead to misleading error messages. For instance, in
```ocaml
module rec X: sig type t = Y.c end = X
and Y: sig class c:object end end with type c := < > = struct
   class c = object end
end
```
the current code will yield an error on a missing type `Y.c` while the root error is that the `with` constraint
```ocaml
sig class c:object end end with type c := < > 
```
is invalid, because a `with` constraint should not be able to remove the type `c` derived from the class `c` from the module type.

PR #4125, opened 2025-06-12
Update toplevel LICENSE file

mshinwell:
We still need to fix headers in files, which I'm working on separately, but this should clarify matters in the toplevel licensing file.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,3 +1,6 @@
+OxCaml licensing
+================
+
 In the following, "the OCaml Core System" refers to all files marked
 "Copyright INRIA" in this distribution.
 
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -18,6 +21,15 @@ clause 2 of the GNU Lesser General Public License.  This exception
 does not however invalidate any other reasons why the executable file
 might be covered by the GNU Lesser General Public License.
 
+Files that are marked "Copyright Jane Street" and not "Copyright INRIA"
+are distributed under the MIT license.
+
+Some files are marked both "Copyright Jane Street" and "Copyright INRIA"
+because they were derived from files in the OCaml Core System.  In this
+case the original OCaml Core System licensing, as above, is in effect.
```
xclerc, line 29:
```suggestion
case, the original OCaml Core System licensing, as above, is in effect.
```
(not sure)

mshinwell: Too many commas

PR #4132, opened 2025-06-12
Add support for the `[@atomic]` attribute on mutable fields

glittershark:
Add support to only the frontend for specifying the `[@atomic]` attribute on (only) mutable fields. This is a no-op currently, but is forwarded down through the ld_mutable on the label description in the typedtree.

This is the first of many commits pulling in upstream support[0] for atomic record fields. I've slightly diverged from the design there, by adding atomic_flag to mutability only under Mutable rather than as a new ld_atomic field, since that feels like a better design (it avoids immutable atomic fields by construction, rather than having to assert that they never happen)

[0] https://github.com/ocaml/ocaml/pull/13404

### DIFF
--- a/typing/typedecl.ml
+++ b/typing/typedecl.ml
@@ -2117,18 +2117,22 @@ let update_decls_jkind env decls =
     (fun (id, decl) ->
        Builtin_attributes.warning_scope decl.type_attributes (fun () ->
          let allow_any_crossing =
-           Builtin_attributes.has_unsafe_allow_any_mode_crossing decl.type_attributes
+           Builtin_attributes.has_unsafe_allow_any_mode_crossing
+             decl.type_attributes
          in
 
-         (* Check that the attribute is valid, if set (unconditionally, for consistency). *)
+         (* Check that the attribute is valid, if set (unconditionally, for
+            consistency). *)
          if allow_any_crossing then begin
            match decl.type_kind with
            | Type_abstract _ | Type_open ->
-             raise(Error(decl.type_loc, Unsafe_mode_crossing_on_invalid_type_kind))
+             raise(Error(
+               decl.type_loc, Unsafe_mode_crossing_on_invalid_type_kind))
            | _ -> ()
          end;
 
-         (id, decl, allow_any_crossing, update_decl_jkind env (Pident id) decl)))
+         (id, decl, allow_any_crossing,
+          update_decl_jkind env (Pident id) decl)))
     decls
 
 (* See Note [Typechecking unboxed versions of types]. *)


testsuite/tests/atomic-locs/record_fields.ml
```diff
@@ -0,0 +1,116 @@
...
+[%%expect{|
+Lines 1-3, characters 17-3:
+1 | .................struct
+2 |   type t = { mutable x : int }
+3 | end......
+Error: Signature mismatch:
+       Modules do not match:
+         sig type t = { mutable x : int; } end
+       is not included in
+         sig type t = { mutable x : int; } end
```
goldfirere: The `[@atomic]` should print here. (and elsewhere)

testsuite/tests/atomic-locs/record_fields.ml
```diff
@@ -0,0 +1,116 @@
+(* TEST
+   flags = "-dlambda -dno-locations -dno-unique-ids";
+   expect;
+*)
+
+(* Check module interface checking: it is not allowed to remove or add
+   atomic attributes. *)
+
+module Wrong1 = (struct
```
goldfirere, line 21: Is there a reason to use this syntax rather than `module M : sig ... end = struct ... end`? I got the struct and sig backwards on a first reading.

testsuite/tests/atomic-locs/record_fields.ml
goldfirere, line 1: There are lots of relevant tests in https://github.com/ocaml/ocaml/pull/13404/files#diff-ab614a7953e51fff5c47bd357217b5c774fce4f9a9bd9a58eb60725b87604d0e

Maybe those get added in a future patch?

typing/types.mli
```diff
@@ -26,14 +26,22 @@ open Allowance
...
+  | Mutable of
+      { modal_upper_bound : Mode.Alloc.Comonadic.Const.t
+        (** The upper bound of the new field value upon mutation. *)
+      ; atomic : atomic
+      }
 
-(** Returns [true] is the [mutable_flag] is mutable. Should be called if not
-    interested in the payload of [Mutable]. *)
+(** Returns [true] is the [mutable_flag] is mutable or atomic. Should be called
+    if not interested in the payload of [Mutable] or [Atomic]. *)
```
goldfirere:
```suggestion
    if not interested in the payload of [Mutable]. *)
```

typing/typedecl.ml
```diff
@@ -481,19 +482,32 @@ let transl_labels (type rep) ~(record_form : rep record_form) ~new_var_jkind
...
       (fun () ->
+         let is_atomic = Builtin_attributes.has_atomic attrs in
          let mut : mutability =
-          match mut with
-          | Immutable -> Immutable
-          | Mutable ->
+          match mut, is_atomic with
+          | Immutable, false -> Immutable
+          | Immutable, true ->
+            raise (Error (loc, Atomic_field_must_be_mutable name.txt))
```
goldfirere, line 490: Would be good to have a test for this case.

parsing/asttypes.mli
```diff
@@ -40,6 +40,8 @@ type private_flag = Private | Public
 
 type mutable_flag = Immutable | Mutable
 
+type atomic_flag = Nonatomic | Atomic
```
goldfirere: I think this is unused.

typing/typedecl.ml
```diff
@@ -481,19 +482,32 @@ let transl_labels (type rep) ~(record_form : rep record_form) ~new_var_jkind
...
+                atomic = if is_atomic then Atomic else Nonatomic
+              }
               | Unboxed_product -> raise(Error(loc, Unboxed_mutable_label))
          in
          let modalities =
           Typemode.transl_modalities ~maturity:Stable mut modalities
          in
          let arg = Ast_helper.Typ.force_poly arg in
          let cty = transl_simple_type ~new_var_jkind env ?univars ~closed Mode.Alloc.Const.legacy arg in
+         let is_atomic = Builtin_attributes.has_atomic attrs in
```
goldfirere: duplicate?

typing/typedecl.ml
```diff
@@ -481,19 +482,32 @@ let transl_labels (type rep) ~(record_form : rep record_form) ~new_var_jkind
...
+         begin match is_atomic, mut with
+         | true, Mutable _
+         | false, _  -> ()
+         | true, Immutable ->
+           raise (Error (loc, Atomic_field_must_be_mutable name.txt));
```
goldfirere: duplicate?

typing/untypeast.ml
```diff
@@ -273,7 +273,7 @@ let constructor_declaration sub cd =
 let mutable_ (mut : Types.mutability) : mutable_flag =
   match mut with
   | Immutable -> Immutable
-  | Mutable m ->
+  | Mutable { modal_upper_bound = m; atomic = _ } ->
```
goldfirere:
```suggestion
  (* ignore [atomic], as it's already in the attributes *)
  | Mutable { modal_upper_bound = m; atomic = _ } ->
```

typing/typecore.ml
```diff
@@ -3041,7 +3041,11 @@ and type_pat_aux
   | Ppat_array (mut, spl) ->
       let mut =
         match mut with
-        | Mutable -> Mutable Alloc.Comonadic.Const.legacy
+        | Mutable  -> Mutable {
+          modal_upper_bound = Alloc.Comonadic.Const.legacy;
+          (* CR aspsmith: Revisit once we support atomic arrays *)
```
goldfirere, line 3116: What is the status of atomic array elements, anyway? Is there upstream support? Is it planned? I'm wondering if there should be an internal epic for this... and if there is, then we can label this comment with the epic somehow so we can find all such comments.

typing/typecore.ml
```diff
@@ -6223,7 +6228,11 @@ and type_expect_
   | Pexp_array(mut, sargl) ->
       let mutability =
         match mut with
-        | Mutable -> Mutable Alloc.Comonadic.Const.legacy
+        | Mutable -> Mutable {
+          modal_upper_bound = Alloc.Comonadic.Const.legacy;
+          (* CR aspsmith: Revisit once we support atomic record fields *)
```
goldfirere:
```suggestion
          (* CR aspsmith: Revisit once we support atomic arrays *)
```

goldfirere (approved):

testsuite/tests/atomic-locs/record_fields.ml
glittershark: this test is taken from upstream verbatim, I'm kinda reticent to change it at all from there (just to make sure we maintain upstream compatibility). feel free to push back on that (or just resolve if you agree)

testsuite/tests/atomic-locs/record_fields.ml
glittershark: yep!

glittershark: we need actual support for stuff before most of those can be added.

typing/typedecl.ml
glittershark: done

typing/typecore.ml
glittershark: I think we can basically support it with just some stdlib changes after this series of PRs? And I plan on doing that, since I have some use cases for them.

glittershark:
@goldfirere can you take a look at the last commit, "Disable flat float array if float fields are atomic"? Everything else is relatively-trivial rebasing, but that adds critical new functionality (from upstream).

typing/types.mli
```diff
@@ -35,15 +35,19 @@ type atomic =
 type mutability =
   | Immutable
   | Mutable of
-      { modal_upper_bound : Mode.Alloc.Comonadic.Const.t
-        (** The upper bound of the new field value upon mutation. *)
+      { mode : Mode.Value.Comonadic.lr
```
goldfirere, line 38: I'm a little surprised this isn't a `Const`. Do we need to store variables here?

glittershark: oh, so am I. I think maybe this was bad merge conflict resolution with some changes that were (maybe?) undone in `main`. I'll double check...

glittershark:
> I'm a little dubious of the float-record change. This means that adding `[@atomic]` changes the performance characteristics of nearby fields, something that is probably not widely known. I had thought that we forbade `[@atomic]` in float-records, which would seem to be preferable over the treatment here. If you disagree, maybe post in the team channel; I'd want to consult the wisdom of the crowd here.

I only really disagree because upstream disagrees (I also would have just forbidden this, personally, but the change here is the change made by upstream)

typing/types.mli
glittershark: oh, I think perhaps you're looking at the diff of a merge commit from `main` - `main` has `Mode.Value.Comonadic.lr` here (I think due to `let mutable`, but I'm not really sure).

goldfirere:
We don't have to do it just because they do -- especially because this makes us strictly less expressive (so there's no upstream-public-release concern). I would ask around before doing this.

testsuite/tests/atomic-locs/record_fields.ml
```diff
@@ -159,3 +168,154 @@ module Float_records :
...
+3 |   mutable a : float [@atomic];
+4 | }
+Warning 214 [atomic-float-record-boxed]: This record contains atomic
+float fields, which prevents the float record optimization. The
+fields of this record will be boxed instead of being
+represented as a flat float array.
+0
+
+type not_suppressed_atomic_float = { mutable a : float [@atomic]; }
+|}];;
```
goldfirere: I'd also want to see that e.g. `type t = { mutable x : float [@atomic] } [@@warning "-214"]` avoids the warning. This is more convenient in practice than `[@@@warning]` which affects a whole (suffix of a) module (signature).

typing/typedecl.ml
```diff
@@ -1845,33 +1845,41 @@ let rec update_decl_jkind env dpath decl =
...
            | Unboxed_element Float64 -> repr_summary.float64s <- true
            | Unboxed_element ( Float32 | Bits8 | Bits16 | Bits32 | Bits64
                              | Vec128 | Vec256 | Vec512 | Word | Product _ ) ->
                repr_summary.non_float64_unboxed_fields <- true
-           | Value_element -> repr_summary.values <- true
+           | Value_element ->
+               repr_summary.values <- true;
+               (* Check if this value field is atomic and float *)
+               if Types.is_atomic lbl.Types.ld_mutable
+               && is_float env lbl.Types.ld_type
```
goldfirere: Can this happen? Wouldn't this be a `Float_element`?

typing/typedecl.ml
```diff
@@ -1907,22 +1915,35 @@ let rec update_decl_jkind env dpath decl =
...
+            float64s = false; non_float64_unboxed_fields = false }
           [@warning "+9"] ->
           Misc.fatal_error "Typedecl.update_record_kind: empty record"
       in
+      (* Emit warning if atomic float fields prevented float record
+         optimization *)
+      if repr_summary.atomic_floats then begin
+        (* Check if all fields are floats (atomic or not) *)
+        let all_float_fields =
+          List.for_all (fun lbl -> is_float env lbl.Types.ld_type) lbls
```
goldfirere: Instead of using `is_float`, can you search for `Float_element`? I think that would be faster.

goldfirere: Or, even better, maybe do this check within the cases above that result in `Record_float` and `Record_ufloat`? That seems to be a great point to decide that an atomic field is the only reason the optimization doesn't apply.

testsuite/tests/atomic-locs/record_fields.ml
glittershark: good call, this actually didn't work because we were missing a `warning_scope`

typing/typedecl.ml
glittershark: indeed, sorry, this was left over from a previous version that considered non-atomic floats values. good catch!

typing/typedecl.ml
glittershark: ah, yeah, that's much better, thanks. Done!

goldfirere (approved):

glittershark:
The change to wrap `update_decl_jkind` in a warning scope appears to have inadvertently caused a change to how warnings work in conjunction with attribute attachment via `[@@ppwarning]`. I don't know much of the background of these features, but just from looking at the code I think this looks like I just accidentally fixed a bug - but I'd appreciate you taking a look to see if you agree @goldfirere. I tried poking around a bit in the manual to see if it could help give us a spec here but wasn't able to find anything useful.

dkalinichenko-js (approved):

PR #4138, opened 2025-06-13
Make all atomic primitives take a field index

glittershark:
To pave the road to atomic record fields, make all atomic primitives from lambda onwards operate on an arbitrary field index of a block, rather than always operating on the first field. The previous primitives (%atomic_load, etc) are left over for upstream compatibility, but are translated to field-based operations hardcoded to field 0 as soon as they hit lambda.

Some of this was taken from the upstream PR[0] for atomic fields, but diverges significantly from that in that upstream doesn't have flambda2, and also in that upstream only has C stubs for atomic primitives, not codegen.

[0] https://github.com/ocaml/ocaml/pull/13404


lambda/translprim.ml
```diff
@@ -130,6 +150,7 @@ type prim =
   | Identity
   | Apply of Lambda.region_close * Lambda.layout
   | Revapply of Lambda.region_close * Lambda.layout
+  | Atomic of atomic_op * atomic_kind
```
TheNumbat, line 153: It seems inconsistent to add this to `prim` since they could be covered by the `Primitive` case. Can you just do `Primitive (lambda_of_atomic, atomic_arity)`?

lambda/lambda.mli
```diff
@@ -315,17 +315,18 @@ type primitive =
...
-  | Patomic_exchange of {immediate_or_pointer : immediate_or_pointer}
-  | Patomic_compare_exchange of {immediate_or_pointer : immediate_or_pointer}
-  | Patomic_compare_set of {immediate_or_pointer : immediate_or_pointer}
-  | Patomic_fetch_add
-  | Patomic_add
-  | Patomic_sub
-  | Patomic_land
-  | Patomic_lor
-  | Patomic_lxor
+  | Patomic_load_field of { immediate_or_pointer : immediate_or_pointer }
```
TheNumbat, line 319: Note that these operations must not be used on fields of float blocks

lambda/translprim.ml
```diff
@@ -901,22 +922,28 @@ let lookup_primitive loc ~poly_mode ~poly_sort pos p =
...
-        Primitive (Patomic_compare_exchange {immediate_or_pointer=Pointer}, 3)
-    | "%atomic_cas" ->
-        Primitive (Patomic_compare_set {immediate_or_pointer=Pointer}, 3)
-    | "%atomic_fetch_add" -> Primitive (Patomic_fetch_add, 2)
-    | "%atomic_add" -> Primitive (Patomic_add, 2)
-    | "%atomic_sub" -> Primitive (Patomic_sub, 2)
-    | "%atomic_land" -> Primitive (Patomic_land, 2)
-    | "%atomic_lor" -> Primitive (Patomic_lor, 2)
-    | "%atomic_lxor" -> Primitive (Patomic_lxor, 2)
+    | "%atomic_load" -> Atomic(Load, Ref)
```
TheNumbat, line 925: Add a CR noting what else we need to do before exposing these in the stdlib (atomic field attribute that rejects float blocks + tests?)

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
```diff
@@ -2664,7 +2678,7 @@ let convert_lprim ~big_endian (prim : L.primitive) (args : Simple.t list list)
       | Punbox_vector _
       | Pbox_vector (_, _)
       | Punbox_int _ | Pbox_int _ | Punboxed_product_field _ | Pget_header _
-      | Pufloatfield _ | Patomic_load _ | Pmixedfield _
+      | Pufloatfield _ | Patomic_load_field _ | Pmixedfield _
```
TheNumbat: This is no longer unary (+ the failure cases below)

middle_end/flambda2/simplify/simplify_ternary_primitive.ml
```diff
@@ -68,83 +68,33 @@ let simplify_bigarray_set ~num_dimensions:_ _bigarray_kind _bigarray_layout dacc
...
     Named.create_prim
-      (Ternary
-         (Atomic_compare_and_set args_kind, atomic, comparison_value, new_value))
+      (Ternary (Atomic_field_int_arith op, atomic, field, value))
       dbg
   in
-  let dacc = DA.add_variable dacc result_var T.any_tagged_bool in
+  let result_kind = P.result_kind' original_prim in
+  let dacc =
+    if Flambda_kind.equal result_kind Flambda_kind.value
```
TheNumbat: Atomic int arith wasn't previously specialized; is this necessary? And isn't the result always an immediate?

middle_end/flambda2/terms/code_size.ml
```diff
@@ -400,11 +400,7 @@ let binary_prim_size prim =
     binary_float_comp_primitive width cmp
   | Float_comp (_width, Yielding_int_like_compare_functions ()) -> 8
   | Bigarray_get_alignment _ -> 3 (* load data + add index + and *)
-  | Atomic_int_arith _ -> 1
-  | Atomic_set Immediate -> 1
-  | Atomic_exchange Immediate -> 1
-  | Atomic_exchange Any_value | Atomic_set Any_value ->
-    does_not_need_caml_c_call_extcall_size
+  | Atomic_load_field _ -> 1
```
TheNumbat, line 403: Potentially 2 for an address computation and then the load? Maybe we don't care since it will usually be inlined into the addressing mode. (Same below)

lambda/lambda.ml
```diff
@@ -2484,27 +2485,27 @@ let primitive_result_layout (p : primitive) =
...
   | Prunstack | Presume | Pperform | Preperform -> layout_any_value
-  | Patomic_load { immediate_or_pointer = Immediate } ->
+  | Patomic_load_field { immediate_or_pointer = Immediate } ->
     layout_int_or_null
-  | Patomic_load { immediate_or_pointer = Pointer } ->
+  | Patomic_load_field { immediate_or_pointer = Pointer } ->
     layout_any_value
-  | Patomic_set _ -> layout_unit
-  | Patomic_exchange { immediate_or_pointer = Immediate } ->
+  | Patomic_set_field _ -> layout_unit
```
TheNumbat, line 2500: We say `%atomic_set` returns unit, but in bytecode it compiles to a call to `caml_atomic_exchange`, which returns the exchanged value. Not sure if that could directly break anything but it should be fixed. (preexisting bug, also note it's correct in native as to_cmm_primitive adds a return_unit)

middle_end/flambda2/terms/flambda_primitive.ml
```diff
@@ -1865,7 +1816,7 @@ let args_kind_of_binary_primitive p =
   | Float_arith (Float32, _) | Float_comp (Float32, _) ->
     K.naked_float32, K.naked_float32
   | Bigarray_get_alignment _ -> bigstring_kind, K.naked_immediate
-  | Atomic_set _ | Atomic_exchange _ | Atomic_int_arith _ -> K.value, K.value
+  | Atomic_load_field _ -> K.value, K.value
```
TheNumbat: Should field offset arguments be untagged? Not folding in the untag may improve codegen, but idk if it matters

middle_end/flambda2/terms/flambda_primitive.ml
```diff
@@ -1888,9 +1839,7 @@ let result_kind_of_binary_primitive p : result_kind =
   | Float_arith (Float32, _) -> Singleton K.naked_float32
   | Phys_equal _ | Int_comp _ | Float_comp _ -> Singleton K.naked_immediate
   | Bigarray_get_alignment _ -> Singleton K.naked_immediate
-  | Atomic_exchange _ | Atomic_int_arith Fetch_add -> Singleton K.value
-  | Atomic_set _ -> Unit
-  | Atomic_int_arith (Add | Sub | And | Or | Xor) -> Unit
+  | Atomic_load_field _ -> Singleton K.value
```
TheNumbat: It would be nice if all of the wildcards for these wrote out the block access kinds so we remember to update them all when adding atomic ops on unboxed fields

runtime/memory.c
```diff
@@ -337,163 +311,221 @@ CAMLprim value caml_atomic_make(value v)
...
   } else {
-    value v;
     /* See Note [MM] above */
     atomic_thread_fence(memory_order_acquire);
-    v = atomic_load(Op_atomic_val(ref));
-    return v;
+    return atomic_load(&Op_atomic_val(obj)[field]);
   }
 }
+CAMLprim value caml_atomic_load (value ref)
```
TheNumbat, line 326: newline

runtime/memory.c
```diff
@@ -337,163 +311,221 @@ CAMLprim value caml_atomic_make(value v)
...
+  value obj, value vfield, value oldv, value newv)
 {
+  intnat field = Long_val(vfield);
   if (caml_domain_alone()) {
-    value* p = Op_val(ref);
+    value* p = &Op_val(obj)[field];
     if (*p == oldv) {
       *p = newv;
-      write_barrier(ref, 0, oldv, newv);
+      write_barrier(obj, 0, oldv, newv);
```
TheNumbat: `field` & below

runtime4/misc.c
```diff
@@ -247,12 +247,19 @@ CAMLprim value caml_atomic_make(value v)
...
+
+
 CAMLprim value caml_atomic_load(value ref)
 {
-  return Field(ref, 0);
+  return caml_atomic_load_field(ref, Val_long(0));
 }
 
-CAMLprim value caml_atomic_compare_exchange(value ref, value oldv, value newv)
+CAMLprim value caml_atomic_compare_exchange_field(value ref, value vfield, value oldv, value newv)
```
TheNumbat, line 274: `vfield` unused

runtime4/misc.c
```diff
@@ -263,7 +270,11 @@ CAMLprim value caml_atomic_compare_exchange(value ref, value oldv, value newv)
   }
 }
 
-CAMLprim value caml_atomic_cas(value ref, value oldv, value newv)
+CAMLprim value caml_atomic_compare_exchange(value ref, value oldv, value newv) {
+  return caml_atomic_compare_exchange_field(ref, Val_long(0), oldv, newv);
+}
+
+CAMLprim value caml_atomic_cas_field(value ref, value vfield, value oldv, value newv)
```
TheNumbat, line 289: `vfield` unused

lambda/translprim.ml
glittershark: I felt the same way, but this is what upstream did and I don't want to diverge too much from that (mostly for merge conflict reasons)

middle_end/flambda2/simplify/simplify_ternary_primitive.ml
glittershark: yeah, good point. Removed.

lambda/lambda.ml
glittershark: fixed in https://github.com/oxcaml/oxcaml/pull/4265, I'll rebase this PR onto that and change the function to be field based.

middle_end/flambda2/terms/code_size.ml
glittershark: What do you think we should do? I'm not familiar enough with this code to have an opinion (is this... size in instructions?)

middle_end/flambda2/terms/flambda_primitive.ml
glittershark: I'd rather keep it simple for now

lambda/translprim.ml
glittershark: I've got PRs already open that address that CR, so I'd rather leave it off just to avoid churn, do you mind?

middle_end/flambda2/terms/code_size.ml
TheNumbat: Leaving it as `1` is fine

TheNumbat: It is basically size in instructions, but it's approximating the actual code size :shrug:

lambda/translprim.ml
TheNumbat: This file may already be a lost cause for upstream merge conflicts, but alright

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
```diff
@@ -2712,8 +2726,10 @@ let convert_lprim ~big_endian (prim : L.primitive) (args : Simple.t list list)
             _,
             _ )
       | Pcompare_ints | Pcompare_floats _ | Pcompare_bints _
-      | Patomic_exchange _ | Patomic_set _ | Patomic_fetch_add | Patomic_add
-      | Patomic_sub | Patomic_land | Patomic_lor | Patomic_lxor | Ppoke _ ),
+      | Patomic_exchange_field _ | Patomic_set_field _ | Patomic_fetch_add_field
```
TheNumbat: These are no longer binary

runtime/memory.c
TheNumbat: as in the write barrier needs `field`

glittershark: hmm, I'm not sure what you're saying? The code is now:
```      
write_barrier(obj, Long_val(vfield), oldv, newv);
```

TheNumbat: oops, must have looked at an old version

runtime/caml/memory.h
```diff
@@ -46,7 +46,7 @@ CAMLextern void caml_free_dependent_memory (value v, mlsize_t bsz);
 CAMLextern void caml_modify (volatile value *, value);
 CAMLextern void caml_modify_local (value obj, intnat i, value val);
 CAMLextern void caml_initialize (volatile value *, value);
-CAMLextern int caml_atomic_cas_field (value, intnat, value, value);
+CAMLextern value caml_atomic_cas_field (value, value, value, value);
```
TheNumbat, line 49: I realized this changes an exposed part of the runtime API in a way that's not compatible with upstream, and it seems particularly scary since you won't get an error if you pass an int here. We should rename this version and probably also provide the old version.

glittershark: upstream actually made the same change: https://github.com/ocaml/ocaml/blob/trunk/runtime/caml/memory.h#L46

TheNumbat: Technically this compiler still supports upstream 5.2 packages, but seems fine given it will reconverge

PR #4140, opened 2025-06-16
(Arm64) implement simd intrinsics and enable simd tests

gretay-js:
(on top of https://github.com/oxcaml/oxcaml/pull/4139)
~(conflicts with #3966)~

This PR add arm64-equivalents of most SSE intrinsics and enables the corresponding tests. 
The PR also re-erranges the tests:
- splitting out the `external` declarations into a separate target-specific file.
- splitting the big `ops.ml` test into separate files per type, to make disabling individual tests easier.
- splititng out tests that are amd64 specific and do not current have an equivalent implementation on arm64.


backend/arm64/emit.ml
```diff
@@ -359,9 +379,21 @@ end = struct
     | Rf64_to_Rf64 ->
       check_reg Float i.arg.(0);
       check_reg Float i.res.(0)
-    | Rf32_to_Ri64 ->
+    | Rf32_to_Rs64 ->
       check_reg Float32 i.arg.(0);
       check_reg Int i.res.(0)
+    | Rs32x4_to_Rs32 _ | Rs64x2_to_Rs64 _ | Rs16x8_to_Rs16 _ | Rs8x16_to_Rs8 _
+      ->
+      check_reg Vec128 i.arg.(0)
```
TheNumbat: Also check `res`?

backend/arm64/emit.ml
```diff
@@ -359,9 +379,21 @@ end = struct
...
+    | Rs64x2_Rs64x2_to_First _ | Rs16x8_Rs32x4_to_First | Rs8x16_Rs16x8_to_First
+    | Rs32x4_Rs64x2_to_First ->
+      check_reg Vec128 i.arg.(0);
+      check_reg Vec128 i.arg.(1);
+      assert (Reg.same_loc i.res.(0) i.arg.(0))
+    | Rs32x4_Rs32_to_First _ | Rs64x2_Rs64_to_First _ | Rs16x8_Rs16_to_First _
+    | Rs8x16_Rs8_to_First _ ->
+      check_reg Vec128 i.arg.(0);
```
TheNumbat, line 391: Also check `arg.(1)`?

backend/arm64/emit.ml
```diff
@@ -775,29 +981,52 @@ let emit_literals p align emit_literal =
...
+    then
       D.switch_to_section_raw
-        ~names:["__TEXT,__literal" ^ Int.to_string align]
+        ~names:["__TEXT"; "__literal" ^ Int.to_string align]
         ~flags:None
         ~args:[Int.to_string align ^ "byte_literals"]
+        ~is_delayed:false
+    else
+      D.switch_to_section_raw
+        ~names:[".rodata.cst" ^ Int.to_string align]
```
TheNumbat, line 989: Did we not need this change previously?

backend/arm64/emit.ml
```diff
@@ -775,29 +981,52 @@ let emit_literals p align emit_literal =
...
+    D.unsafe_set_internal_section_ref Text;
     D.align ~fill_x86_bin_emitter:Nop ~bytes:align;
     List.iter emit_literal !p;
     p := [])
 
+let emit_float32_literal (f, lbl) =
+  D.define_label lbl;
+  let comment =
+    if !Clflags.keep_asm_file
+    then Some (Printf.sprintf "%.12f" (Int32.float_of_bits f))
```
TheNumbat: This should be `%.9g`, and same in `asm_directives.ml`. Also shouldn't the double case use `%.17g` rather than `%.12g`?

backend/arm64/emit.ml
```diff
@@ -775,29 +981,52 @@ let emit_literals p align emit_literal =
...
+    else None
+  in
+  (* Do not use [D.float32] to aviod conversion [Int32.float_of_bits f] that
```
TheNumbat: It looks like `D.float32` does the same thing though?

backend/arm64/emit.ml
```diff
@@ -1596,11 +1825,27 @@ let emit_instr i =
...
+        assert (not !Clflags.dlcode);
+        (* see selection_utils.ml *)
+        let s = S.create s in
+        DSL.ins I.ADRP [| DSL.emit_reg reg_tmp1; DSL.emit_symbol ~offset s |];
+        DSL.ins I.ADD
+          [| DSL.emit_reg reg_tmp1;
+             DSL.emit_reg reg_tmp1;
+             DSL.emit_symbol ~reloc:LOWER_TWELVE ~offset s
+          |]);
+      DSL.ins I.LDR [| DSL.emit_reg dst; DSL.emit_mem reg_tmp1 |]
```
TheNumbat, line 1837: Is the reason we have to do this so that `LDR` doesn't get an alignment hint? If so can we add a flag to `emit_addressing` instead?

backend/arm64/emit.ml
```diff
@@ -1637,10 +1882,29 @@ let emit_instr i =
...
+        assert (not !Clflags.dlcode);
+        (* see selection_utils.ml *)
+        let s = S.create s in
+        DSL.ins I.ADRP [| DSL.emit_reg reg_tmp1; DSL.emit_symbol ~offset s |];
+        DSL.ins I.ADD
+          [| DSL.emit_reg reg_tmp1;
+             DSL.emit_reg reg_tmp1;
+             DSL.emit_symbol ~reloc:LOWER_TWELVE ~offset s
+          |]);
+      DSL.ins I.STR [| DSL.emit_reg src; DSL.emit_mem reg_tmp1 |]
```
TheNumbat, line 1893: Same

backend/arm64/emit.ml
```diff
@@ -1637,10 +1882,29 @@ let emit_instr i =
     | Single { reg = Float32 } ->
       DSL.check_reg Float32 src;
       DSL.ins I.STR [| DSL.emit_reg src; DSL.emit_addressing addr base |]
-    | Onetwentyeight_aligned | Onetwentyeight_unaligned ->
-      (* CR gyorsh: check alignment *)
+    | Onetwentyeight_aligned ->
       DSL.check_reg Vec128 src;
       DSL.ins I.STR [| DSL.emit_reg src; DSL.emit_addressing addr base |]
+    (* CR gyorsh: should the above line be:
```
TheNumbat: Seems like no

backend/arm64/simd_proc.ml
```diff
@@ -37,41 +37,155 @@ type register_behavior =
...
+    Rs64x2_Rs64x2_to_Rs64x2
+  | Cvtq_s32_f32 -> Rf32x4_to_Rs32x4
+  | Cvtq_f32_s32 -> Rs32x4_to_Rf32x4
   | Cvt_f64_f32 ->
     (* Input should be in Vec128 register but only the bottom f32x2 is used by
        this instruction. *)
     Rf32x2_to_Rf64x2
-  | Cmp_f32 _ -> Rf32x4_Rf32x4_to_Ri32x4
-  | Cmpz_s32 _ -> Ri32x4_to_Ri32x4
+  | Cvt_f32_f64 ->
```
TheNumbat, line 121: Slightly confusing that the operations are x_of_y but the registers are x_to_y

backend/arm64/simd_selection.ml
```diff
@@ -35,62 +61,357 @@ open! Int_replace_polymorphic_compare [@@ocaml.warning "-66"]
    Some intrinsics have both names to make it easier to correlate with both
    amd64 intrinsics and arm64 instructions, depending on context. *)
 
-let select_simd_instr op args =
+let select_simd_instr op args dbg =
   match op with
   | "caml_simd_float32_round_neg_inf" -> Some (Round_f32 Neg_inf, args)
+  | "caml_simd_float64_round_neg_inf" -> Some (Round_f64 Neg_inf, args)
```
TheNumbat: I think all of the `caml_simd` cases should also accept `caml_neon` (unless the neon version is different of course)

backend/arm64/simd_selection.ml
```diff
@@ -35,62 +61,357 @@ open! Int_replace_polymorphic_compare [@@ocaml.warning "-66"]
...
+  select_simd_instr op args dbg
   |> Option.map (fun (op, args) -> Operation.Specific (Isimd op), args)
 
-let pseudoregs_for_operation _ arg res = arg, res
+let pseudoregs_for_operation (simd_op : Simd.operation) arg res =
+  match Simd_proc.register_behavior simd_op with
+  | Rs32x4_Rs32_to_First _ | Rs64x2_Rs64_to_First _ | Rs16x8_Rs16_to_First _
+  | Rs8x16_Rs8_to_First _ | Rs64x2_Rs64x2_to_First _ | Rs16x8_Rs32x4_to_First
+  | Rs8x16_Rs16x8_to_First | Rs32x4_Rs64x2_to_First ->
+    let arg = Array.copy arg in
```
TheNumbat, line 412: Does this need to copy?

backend/arm64_ast.ml
```diff
@@ -41,24 +52,75 @@ module Neon_reg_name = struct
...
+      check_index 0 last t.lane
+
+    let name t index =
+      let suffix =
+        match t.r with V v -> Vector.to_string v | S s -> Scalar.to_string s
+      in
+      Printf.sprintf "V%d.%s[%d]" index suffix t.lane
+  end
+
+  module Struct_lane = struct
```
TheNumbat: The only use of this is `emit_struct_reglane_d`, which is unused. Should we delete it?

oxcaml/tests/simd/amd64/sse_other_builtins.ml
```diff
@@ -0,0 +1,109 @@
+(* CR gyorsh: Add arm64 support for intrinsics below. This file contains amd64
+   intrinsics that don't have an equivalent arm64 neon intrinsic. They can be
+   implemented using a very short sequence of arm64 instructons. *)
```
TheNumbat: If they can be implemented in terms of existing intrinsics, they can go in the `ocaml_simd_neon` library instead

oxcaml/tests/simd/arm64/builtins.ml
```diff
@@ -0,0 +1,1233 @@
+(* CR gyorsh: all instructions in this file can be renamed from "caml_neon" to
+   "caml_simd" because they have the corresponding implementation on amd64. If
+   we do it, [builtins.ml] in target specific folders will be identical, and we
+   can move them up into the parent folder. If we keep support for both
+   "caml_neon" and "caml_simd" in the compiler, we should add some tests for
```
TheNumbat, line 5: I think it's fine to only add `caml_simd` names for the intrinsics we need in base/ocaml_intrinsics_kernel, everything else can delegate to `ocaml_simd_sse` or `ocaml_simd_neon` respectively.
But agree we should also test the `caml_simd` names

oxcaml/tests/simd/arm64/builtins.ml
```diff
@@ -0,0 +1,1233 @@
...
+
+  external cvt_float32x4 : t -> float32x4
+    = "caml_vec128_unreachable" "caml_neon_cvt_float64x2_to_float32x2"
+    [@@noalloc] [@@unboxed] [@@builtin]
+
+  external round_near : (t[@unboxed]) -> (t[@unboxed])
+    = "caml_vec128_unreachable" "caml_neon_float64x2_round_near"
+    [@@noalloc] [@@builtin]
+
+  (* CR gyorsh: which coversion function should be used? *)
```
TheNumbat, line 539: Is that the right one?

oxcaml/tests/simd/arm64/builtins.ml
```diff
@@ -0,0 +1,1233 @@
...
+  external interleave_high_32 : t -> t -> t
+    = "caml_vec128_unreachable" "caml_simd_vec128_interleave_high_32"
+    [@@noalloc] [@@unboxed] [@@builtin]
+
+  external interleave_low_32 : t -> t -> t
+    = "caml_vec128_unreachable" "caml_simd_vec128_interleave_low_32"
+    [@@noalloc] [@@unboxed] [@@builtin]
+
+  (* CR gyorsh: looks like there is no single instruction on arm64 that
+     corresponds to Intel's [shuffle_32], TBL and TBX are the closest. *)
```
TheNumbat, line 943: This and the movemasks and shuffles can also be implemented in `caml_simd_neon`

oxcaml/tests/simd/gen/gen_dune.ml
```diff
@@ -0,0 +1,150 @@
...
+      "arrays", enabled_if_main;
+      "arrays_u", enabled_if_main;
+      "scalar_ops", enabled_if_main;
+      "consts", enabled_if_main;
+      "consts_u", enabled_if_main;
+      "callback", enabled_if_main;
+      "callback256", enabled_if_main_amd64_not_macos;
+      "test_callee_save_neon_regs", enabled_if_main;
+      "probes", enabled_if_main_amd64_not_macos;
+      "probes256", enabled_if_main_amd64_not_macos ]
```
TheNumbat, line 173: Thanks, updating the dune file was getting annoying lol

oxcaml/tests/simd/scalar_ops.ml
```diff
@@ -22,27 +22,6 @@ let eq' x y = if x <> y then Printf.printf "%016Lx <> %016Lx\n" x y
 
 let eqi x y = if x <> y then Printf.printf "%d <> %d\n" x y
 
-module Int64x2 = struct
-  type t = int64x2
-
-  external clmul :
```
TheNumbat, line 28: Do these tests still run anywhere?

backend/arm64/simd_proc.ml
gretay-js: Yes, sorry, the operations are named after the intrinsics in `arm_neon.h` which have a very confusing convention. I am going to refactor this similarly to what you did in `amd64` backend and generate these intrinsics automatically, so `simd_proc` will go away along with its confusing naming.

backend/arm64_ast.ml
gretay-js: Removed for now.

oxcaml/tests/simd/amd64/sse_other_builtins.ml
gretay-js: yes, I'm just not sure we have all the intrinsics in the compiler that are needed for it, updated the comment.

oxcaml/tests/simd/scalar_ops.ml
gretay-js: Test for `clmul` moved to `oxcaml/tests/simd/amd64/sse_other_ops.ml` but I forgot to sse-only tests to `gen_dune`-  fixed. thanks for the catch!

backend/arm64/emit.ml
gretay-js: Added. I intentionally didn't have checks for "int" registers because I wasn't sure if their machtype would always be `Int` or may end up being `Val` or `Addr` after regalloc (similarly to the issue we have in `Emit.move`). We will see what happens :)

backend/arm64/emit.ml
gretay-js: We do it to handle larger offsets that need to be a multiple of the access size (see `Cfg_selection.is_offset` for arm64). I guess this form is not friendly for prefetching (in addition to being a longer code sequence)... is that what you mean by "alignement **hint**"?

backend/arm64/emit.ml
gretay-js: Without it, arm64 ubuntu CI was failing (with a weirld linker error about relocations, actually) only when the constants where used, not when they were emitted, and the array tests that triggered it are only enabled on arm64 by this PR. For arm64, when we had non-macos CI jobs, we didn't have vec128 I think. We have similar 16-byte rodata sections on amd64 for vec128.

backend/arm64/emit.ml
gretay-js: sorry, removed. this was leftover from some debugging. 

it may not be doing exactly the same thing when going through the C stub for `caml_int32_bits_of_float_unboxed`    / `caml_int32_float_of_bits_unboxed` which use `double` for arg / return types, resp.

backend/arm64/emit.ml
gretay-js: Fixed. I was also wondering where the "%.12f" came from. It's just a comment though so it doesn't matter. However, in `Arm64_ast` we are using "%.7f" to emit a constant operand of `mov`. Looks like [7 was enough](https://github.com/ocaml/ocaml/pull/1904#issuecomment-458505815). I don't know what the downside is of printing the constant with its full precision using `%.17g`. Is it worth trying to find the minimal number for float32 in the case it's not the same? Actually, I am tempted to change it to .17 for float and .9 for float32 (in a separate PR).

backend/arm64/simd_selection.ml
gretay-js: Not sure, there used to be sharing of these arrays and amd64 version copies them too. Seems safe and it doesn't happen often enough to worry about for now.

backend/arm64/simd_selection.ml
gretay-js: Done.

oxcaml/tests/simd/arm64/builtins.ml
gretay-js: I'll do it as a follow-up with auto-gen the rest of the intrinsics.

backend/arm64/emit.ml
TheNumbat: Hmm yeah, it will preserve the value but potentially not nan patterns. Let's add `float32_from_bits` instead.

backend/arm64/emit.ml
TheNumbat: nvm, I read someone addressing modes can be annotated with an alignment but we're not doing that

backend/arm64/simd_selection.ml
TheNumbat: I've also updated amd64 selection to support the same set of simd names

oxcaml/tests/simd/arm64/builtins.ml
TheNumbat: I think it should be round_current - the amd64 version is `cvtps2dq`, which uses the current rounding mode

oxcaml/tests/simd/scalar_ops.ml
TheNumbat: This also needs `pext/pdep`

gretay-js: This is covered under the name `Int64.bit_deposit` and `Int64.bit_extract`  in `sse_other_ops.ml` which is now enabled on amd64 (see `dune.inc`), but "run_if_not_under_rosetta2" doesn't seem to be working and is not needed any more I think. I'll make a separate PR to discuss it.

oxcaml/tests/simd/arm64/builtins.ml
gretay-js: There seems to be some difference with the conversion between diffferent widths and number of elements. I'll make a separate PR to try to re-enable this test.

PR #4144, opened 2025-06-16
Optimize if-then-else when condition uses boolean operations

Gbury:
This PR optimizes the translation from lambda to flambda of if-then-elses whose condition is a sequential boolean operator.

This optimizes code such as the following:
```ocaml
let foo (x : int) (y : int) (z : int) =
  if x < y && y < z then 42 else 0
```

### DIFF
--- a/middle_end/flambda2/from_lambda/lambda_to_lambda_transforms.ml
+++ b/middle_end/flambda2/from_lambda/lambda_to_lambda_transforms.ml
@@ -22,48 +22,57 @@ type primitive_transform_result =
   | Primitive of L.primitive * L.lambda list * L.scoped_location
   | Transformed of L.lambda
 
+let mk_switch ~cond ~ifso ~ifnot ~kind =
+  let switch : L.lambda_switch =
+    { sw_numconsts = 2;
+      sw_consts = [0, ifnot; 1, ifso];
+      sw_numblocks = 0;
+      sw_blocks = [];
+      sw_failaction = None
+    }
+  in
+  L.Lswitch (cond, switch, L.try_to_find_location cond, kind)
+
+let optimize_boolean_sequop ~cond ~ifso ~ifnot ~kind =
+  let ifso_cont = Lambda.next_raise_count () in
+  let ifso_jump = L.Lstaticraise (ifso_cont, []) in
+  let ifnot_cont = Lambda.next_raise_count () in
+  let ifnot_jump = L.Lstaticraise (ifnot_cont, []) in
+  let rec aux ~cond ~ifso ~ifnot =
+    match[@warning "-4"] cond with
+    (* CR gbury: should we try to use the locs here, or is it better to keep
+       using the locs from each individual condition ? *)
+    | L.Lprim (Psequand, [a; b], _loc) ->
+      aux ~cond:a ~ifnot ~ifso:(aux ~cond:b ~ifso ~ifnot)
+    | L.Lprim (Psequor, [a; b], _loc) ->
+      aux ~cond:a ~ifso ~ifnot:(aux ~cond:b ~ifso ~ifnot)
+    | L.Lprim (Pnot, [c], _loc) -> aux ~cond:c ~ifso:ifnot ~ifnot:ifso
+    | L.Lprim ((Psequand | Psequor), _, _) ->
+      Misc.fatal_error "Psequand / Psequor must have exactly two arguments"
+    | _ -> mk_switch ~cond ~ifso ~ifnot ~kind
+  in
+  L.Lstaticcatch
+    ( L.Lstaticcatch
+        ( aux ~cond ~ifso:ifso_jump ~ifnot:ifnot_jump,
+          (ifnot_cont, []),
+          ifnot,
+          Same_region,
+          kind ),
+      (ifso_cont, []),
+      ifso,
+      Same_region,
+      kind )
+
+let optimize_boolean_sequop_expr ~cond =
+  let const_true = L.Lconst (Const_base (Const_int 1)) in
+  let const_false = L.Lconst (Const_base (Const_int 0)) in
+  optimize_boolean_sequop ~cond ~ifso:const_true ~ifnot:const_false
+    ~kind:Lambda.layout_int
+
 let switch_for_if_then_else ~cond ~ifso ~ifnot ~kind =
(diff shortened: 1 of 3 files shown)


middle_end/flambda2/from_lambda/lambda_to_lambda_transforms.ml
```diff
@@ -23,15 +23,48 @@ type primitive_transform_result =
...
+    let ifso_cont = Lambda.next_raise_count () in
+    let ifso_jump = L.Lstaticraise (ifso_cont, []) in
+    let ifnot_cont = Lambda.next_raise_count () in
+    let ifnot_jump = L.Lstaticraise (ifnot_cont, []) in
+    let rec aux ~cond ~ifso ~ifnot =
+      match[@warning "-4"] cond with
+      (* CR gbury: should we try to use the locs here, or is it better to keep
+         using the locs from each individual condition ? *)
+      | L.Lprim (Psequand, [a; b], _loc) ->
+        aux ~cond:a ~ifnot ~ifso:(aux ~cond:b ~ifso ~ifnot)
```
lthls:
```suggestion
        aux ~cond:a ~ifso:(aux ~cond:b ~ifso ~ifnot) ~ifnot
```

middle_end/flambda2/from_lambda/lambda_to_lambda_transforms.ml
```diff
@@ -33,6 +33,53 @@ let switch_for_if_then_else ~cond ~ifso ~ifnot ~kind =
...
+(* This function helps bind expression to avoid duplicating them in the
+   generated code. Currently, this is only used for if-then-else optimization,
+   which guarantees that even if expressions are duplicated, they are still only
+   evaluated once (but could have been duplicated among multiple code paths), so
+   this is for code size optimization mainly.
+
+   One could want to duplicate constants, however the backend does not
+   deduplicate them, especially when they are the return value of the function,
+   so for now constants returned are shared via a static catch/raise. *)
+let bind_expr ~kind ~expr k =
```
lthls: I think the name `bind_expr` is not ideal. Usually `bind` helpers generate `let`-bindings; for continuations, the various helpers in `Lambda` tend to use `share` in the name rather than `bind`.

middle_end/flambda2/from_lambda/lambda_to_lambda_transforms.ml
```diff
@@ -33,6 +33,53 @@ let switch_for_if_then_else ~cond ~ifso ~ifnot ~kind =
...
+  | _ ->
+    let cont = L.next_raise_count () in
+    let jump = L.Lstaticraise (cont, []) in
+    L.Lstaticcatch (k jump, (cont, []), expr, Same_region, kind)
+
+let optimize_boolean_sequop ~cond ~ifso ~ifnot ~kind =
+  let rec aux ~kind ~cond ~ifso ~ifnot =
+    match[@warning "-4"] cond with
+    (* CR gbury: should we try to use the locs here, or is it better to keep
+       using the locs from each individual condition ? *)
```
lthls, line 61: `mk_switch` contains a call to `L.try_to_find_location`. I suspect that we may want to instead optionally pass a location, only using `try_to_find_location` if we don't have one. But I'm fine with this PR keeping the CR, and coming back to this later.

middle_end/flambda2/from_lambda/lambda_to_lambda_transforms.ml
```diff
@@ -33,6 +33,53 @@ let switch_for_if_then_else ~cond ~ifso ~ifnot ~kind =
...
+    | L.Lprim (Psequand, [a; b], _loc) ->
+      bind_expr ~kind ~expr:(aux ~kind ~cond:b ~ifso ~ifnot) (fun ifso ->
+          aux ~kind ~cond:a ~ifso ~ifnot)
+    | L.Lprim (Psequor, [a; b], _loc) ->
+      bind_expr ~kind ~expr:(aux ~kind ~cond:b ~ifso ~ifnot) (fun ifnot ->
+          aux ~kind ~cond:a ~ifso ~ifnot)
+    | L.Lprim (Pnot, [c], _loc) -> aux ~kind ~cond:c ~ifso:ifnot ~ifnot:ifso
+    | L.Lprim ((Psequand | Psequor), _, _) ->
+      Misc.fatal_error "Psequand / Psequor must have exactly two arguments"
```
lthls: It's a bit weird to see this error here, in particular since I believe that it is too late: the code in `transform_primitive0` already matches on lists with two elements, so if we end up with a wrong number of elements this case will not even be reached (unless it is deep in a sequence). Instead, we will generate an invalid `Primitive` which will get caught later (possibly with a different error message).

middle_end/flambda2/from_lambda/lambda_to_lambda_transforms.ml
Gbury: renamed to `share_expr`

middle_end/flambda2/from_lambda/lambda_to_lambda_transforms.ml
Gbury: Right, that code was originally in `transform_primitive0` and I moved it here at one point because I had changed the match in `transform_primitive0`,, but now that I've went back to matching on the two arguments, I'll move it back where it was originally.

middle_end/flambda2/from_lambda/lambda_to_lambda_transforms.ml
Gbury: I'd say let's keep the CR here then ?

Gbury:
Indeed, I'll add some more cases, though, since we do this at the level of lambda to flambda, we might not have as much substitution of let-bound values that we would have at the `to_cmm` level, so we might have less practical ocurrences of the `true` and `false` constants (but this is only for classic mode).

Gbury:
@lpw25 the latest version should cover all of the cases that `transl_if` handles.

lthls (approved):
Looks correct.
I think `aux` would deserve a better name but I can't think of a good one at the moment.
I also think that you should post in this PR the examples you used to check that all the cases re reachable and produce the expected code. It is likely not possible to put it into the testsuite (at least not without a lot of work), but if we think the optimisation may be broken it would be nice to be able to reproduce the checks.

PR #4150, opened 2025-06-17
Add `Multicore` to systhreads and `Backoff` to stdlib

TheNumbat:
Merges `Multicore` into systhreads and `Backoff` into stdlib. `Multicore` is intended as the only interface for spawning domains, so `Domain.spawn` now has a more aggressive alert. I've also updated all of the tests that spawn domains to silence the alert.

This code has been reviewed internally, but it has some changes for compilation in the compiler tree:
- No longer uses `ctz` intrinsic 
- No longer uses `Uopt`
- No longer uses `iarray` (mode crossing replaced with magic)
- No longer uses await-based sync primitives (replaced with mutex+condition)
- The manager thread for the initial domain is created lazily (otherwise breaks systhreads tests)
- 80 char limit :angry:

### DIFF
--- a/runtime4/signals.c
+++ b/runtime4/signals.c
@@ -351,7 +351,7 @@ value caml_process_pending_actions_with_root_exn(value extra_root)
   return process_pending_actions_with_root_exn(extra_root);
 }
 
-value caml_process_pending_actions_with_root(value extra_root)
+CAMLprim value caml_process_pending_actions_with_root(value extra_root)
 {
   value res = process_pending_actions_with_root_exn(extra_root);
   caml_raise_async_if_exception(res, "");
--- a/testsuite/tests/basic/patmatch_for_multiple.ml
+++ b/testsuite/tests/basic/patmatch_for_multiple.ml
@@ -26,15 +26,15 @@ match (3, 2, 1) with
 | _ -> false
 ;;
 [%%expect{|
-(let (*match*/283 =[int] 3 *match*/284 =[int] 2 *match*/285 =[int] 1)
+(let (*match*/284 =[int] 3 *match*/285 =[int] 2 *match*/286 =[int] 1)
   (catch
     (catch
-      (catch (if (!= *match*/284 3) (exit 3) (exit 1)) with (3)
-        (if (!= *match*/283 1) (exit 2) (exit 1)))
+      (catch (if (!= *match*/285 3) (exit 3) (exit 1)) with (3)
+        (if (!= *match*/284 1) (exit 2) (exit 1)))
      with (2) 0)
    with (1) 1))
-(let (*match*/283 =[int] 3 *match*/284 =[int] 2 *match*/285 =[int] 1)
-  (catch (if (!= *match*/284 3) (if (!= *match*/283 1) 0 (exit 1)) (exit 1))
+(let (*match*/284 =[int] 3 *match*/285 =[int] 2 *match*/286 =[int] 1)
+  (catch (if (!= *match*/285 3) (if (!= *match*/284 1) 0 (exit 1)) (exit 1))
    with (1) 1))
 - : bool = false
 |}];;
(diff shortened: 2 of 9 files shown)


TheNumbat:
Actually I think `Backoff` should go in stdlib and `Multicore` in systhreads, since we want to depend on backoff in jsoo libs. Will move it tomorrow.

otherlibs/systhreads/Makefile
```diff
@@ -45,12 +45,12 @@ LIBNAME=threads
 BYTECODE_C_OBJS=st_stubs.b.$(O)
 NATIVECODE_C_OBJS=st_stubs.n.$(O)
 
-THREADS_SOURCES = thread.ml event.ml
+THREADS_SOURCES = thread.ml event.ml backoff.ml multicore.ml
```
polytypic: I could see `Backoff`, much like `Atomic`, `Mutex`, and `Condition`, ..., in the Stdlib rather than with systhreads, because it doesn't depend on systhreads and is also useful without systhreads and it is one thing you quickly want to have when you start to use `Atomic`.

(Heh... I noticed your comment after writing mine.)

otherlibs/systhreads/multicore.ml
```diff
@@ -0,0 +1,367 @@
...
+            then (
+                let bit_index = count_trailing_zeros before in
+                let bit_mask = 1 lsl bit_index in
+                let after = before - bit_mask in
+                match
+                Atomic.compare_and_set word before after
+                with
+                | true ->
+                    let i = bit_index + (word_index * bits_per_word) in
+                    This i
```
polytypic: Particularly in the absence of `Int.ctz` there is a better way to do this.  The problem here is that `count_trailing_zeros` is now relatively expensive and is done inside the critical section between `Atomic.get` and `Atomic.compare_and_set`.

The code above finds the index of the lowest non-zero bit and clears it.  The index is really only needed for the return value.  We can make it so that we directly eliminate the lowest set bit and use that to try to update the state.  We only compute the index in case we succeeded:

```ocaml
(* ... *)
let after = before land (before - 1) in
match Atomic.compare_and_set word before after with
| true ->
  let bit_mask = before - after in
  let bit_index = bit_index_of bit_mask in
  let i = bit_index + (word_index * bits_per_word) in
  This i
(* ... *)
```

Also, `count_leading_zeros` can be replaced with [code that finds the highest bit set](https://github.com/ocaml-multicore/multicore-magic/blob/c87d608d18d09cb7f4b84249f27bf217aa69ce10/src/index.ocaml5.ml#L7-L20).

This suggestion could be added as an alternative in `CR-someday` or just implemented right away.

otherlibs/systhreads4/backoff.ml
```diff
@@ -0,0 +1,73 @@
...
+(* ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES   *)
+(* WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN    *)
+(* ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF  *)
+(* OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.           *)
+(*                                                                          *)
+(****************************************************************************)
+
+type t = int
+
+let cpu_relax = Thread.yield
```
polytypic: I believe we only need something that ensures there is a poll point.  I know that `yield` was used in the original code, but I believe now that it is not the best thing to do here.

otherlibs/systhreads/Makefile
TheNumbat: Yeah, moved

otherlibs/systhreads/multicore.ml
TheNumbat: Sounds good, switched to `bit_index_of`

otherlibs/systhreads4/backoff.ml
TheNumbat: Switched back to poll

polytypic (approved):
LGTM

PR #4151, opened 2025-06-18
Take specialization into account in shortcuts

Gbury:
The shortcuting code "forgets" any specialization if/when it triggers and there are no rewrites (or an empty rewrite). This was unfortunately missed when i reviewed the shortcut PR, and this PR fixes that.

### DIFF
--- a/middle_end/flambda2/simplify/expr_builder.ml
+++ b/middle_end/flambda2/simplify/expr_builder.ml
@@ -758,22 +758,23 @@ let rewrite_fixed_arity_continuation0 uacc cont_or_apply_cont ~use_id arity :
     | Continuation cont -> cont
     | Apply_cont apply_cont -> Apply_cont.continuation apply_cont
   in
-  let[@local] shortcut_this_continuation_if_possible () :
-      rewrite_fixed_arity_continuation0_result =
-    (* Apply the shortcut if we can, but not if we are rewriting a
-       [Continuation] since we would need to introduce a new wrapper anyway. *)
-    match cont_or_apply_cont with
-    | Continuation cont ->
-      This_continuation (apply_continuation_aliases uenv cont)
-    | Apply_cont apply_cont ->
-      Apply_cont (apply_continuation_shortcuts uenv apply_cont)
-  in
   let cont =
     match
       Continuation_callsite_map.find cont use_id (UA.specialization_map uacc)
     with
     | exception Not_found -> cont
     | specialized -> specialized
+  in
+  let[@local] shortcut_this_continuation_if_possible () :
+      rewrite_fixed_arity_continuation0_result =
+    (* Apply the shortcut if we can, but not if we are rewriting a
+       [Continuation] since we would need to introduce a new wrapper anyway. *)
+    match cont_or_apply_cont with
+    (* We must be careful to use the specialized continuation *)
+    | Continuation _ -> This_continuation (apply_continuation_aliases uenv cont)
+    | Apply_cont apply_cont ->
+      let apply_cont = Apply_cont.with_continuation apply_cont cont in
+      Apply_cont (apply_continuation_shortcuts uenv apply_cont)
   in
   match UE.find_apply_cont_rewrite uenv cont with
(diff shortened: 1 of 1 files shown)


middle_end/flambda2/simplify/expr_builder.ml
```diff
@@ -758,23 +758,24 @@ let rewrite_fixed_arity_continuation0 uacc cont_or_apply_cont ~use_id arity :
     | Continuation cont -> cont
     | Apply_cont apply_cont -> Apply_cont.continuation apply_cont
   in
+  let cont =
+    match
+      Continuation_callsite_map.find cont use_id (UA.specialization_map uacc)
+    with
+    | exception Not_found -> cont
+    | specialized -> specialized
+  in
```
bclement-ocp: I am a bit worried of having `cont` be the specialised version and `cont_or_apply_cont` the non-specialised version that needs to be specialised later; it causes some subtle control flow in `shortcut_this_continuation_if_possible`.

What do you think of specializing both, e.g.:

```ocaml
let specialize_if_needed cont =
  match Continuation_callsite_map.find cont use_id (UA.specialization_map uacc)
in
let cont, cont_or_apply_cont =
  match cont_or_apply_cont with
  | Continuation cont ->
    let cont = specialize cont in
    cont, Continuation cont
  | Apply_cont apply_cont ->
    let cont = specialize (Apply_cont.continuation apply_cont) in
    let apply_cont = Apply_cont.with_continuation apply_cont cont in
    cont, Apply_cont apply_cont
in
```

Gbury: Indeed, that's cleaner (although slightly less efficient since we allocate some more, but I don't think it matters very much).

Gbury: done in the last commit

bclement-ocp: We can inline `specialize_if_needed` and avoid the allocation in the `Not_found` case if we are worried about allocation

Gbury: with flambda2, it should get unboxed anyway, so that should not matter (well, as long as flambda2 works as expected, ^^)

bclement-ocp: Hmmm I don't think we are able to avoid the allocation of `Continuation cont` / `Apply_cont apply_cont` in the `Not_found` case (without continuation specialization after `specialize`)?

Gbury: in this case, we don't need specialization, and this is independant of what `specialize_if_needed` does (and therefore the `Not_found` case), just look at that shape of code:
```ocaml
let _, cont_or_apply_cont =
  match .. with
  | _ ->
    ..
    _, Continuation cont
  | _ ->
    ..
    _, Apply_cont apply_cont
```

that's textbook unboxing, where we have two jumps (because the let binding will become a continuation binding), each allocating the variant right before the jump, so we can unbox `cont_or_apply_cont`.

bclement-ocp: I was thinking about comparison with the following code, which can reuse the original `cont_or_apply_cont`:

```
let cont, cont_or_apply_cont =
  match cont_or_apply_cont with
  | Continuation cont -> (
    match Continuation_callsite_map.find cont use_id (UA.specialization_map uacc) with
    | cont -> cont, Continuation cont
    | exception Not_found -> cont, cont_or_apply_cont
  )
  | Apply_cont apply_cont ->(
    let cont = Apply_cont.continuation apply_cont in
    match Continuation_callsite_map.find cont use_id (UA.specialization_map uacc) with
    | cont ->
        let apply_cont = Apply_cont.with_continuation apply_cont cont in
        cont, Apply_cont apply_cont
    | exception Not_found -> cont, cont_or_apply_cont
in
```

But I didn't take into consideration that since the returned `cont_or_apply_cont` is only used locally in a subsequent match, it doesn't matter here. Anyway, I think we can merge, as you say it is unlikely to have any impact.

PR #4152, opened 2025-06-18
Disable the "for all or none" check for variable quantification

dkalinichenko-js:
Title. To maintain compatibility with upstream OCaml, accept unbound type variables after quantification only if the extension universe is `stable` or higher. If the extension universe is `upstream_compatible`, emit an error and a hint instead.

### DIFF
--- a/typing/typetexp.ml
+++ b/typing/typetexp.ml
@@ -1019,7 +1019,8 @@ and transl_type_aux env ~row_context ~aliased ~policy mode styp =
       end;
       let name = !name in
       let make_row more =
-        create_row ~fields ~more ~closed:(closed = Asttypes.Closed) ~fixed:None ~name
+        create_row ~fields ~more ~closed:(closed = Asttypes.Closed)
+          ~fixed:None ~name
       in
       let more =
         if Btype.static_row
@@ -1313,7 +1314,8 @@ and transl_fields env ~policy ~row_context o fields =
   let ty_init =
      match o with
      | Asttypes.Closed -> newty Tnil
-     | Asttypes.Open -> TyVarEnv.new_var (Jkind.Builtin.value ~why:Row_variable) policy
+     | Asttypes.Open ->
+        TyVarEnv.new_var (Jkind.Builtin.value ~why:Row_variable) policy
   in
   let ty = List.fold_left (fun ty (s, ty') ->
       newty (Tfield (s, field_public, ty', ty))) ty_init fields in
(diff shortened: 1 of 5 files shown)


typing/typetexp.ml
```diff
@@ -177,7 +177,8 @@ module TyVarEnv : sig
   (* see mli file *)
 
   type policy
-  val make_fixed_policy : jkind_initialization_choice -> policy
+  val make_fixed_policy :
+    ?hint_for_unbound_variables:string -> jkind_initialization_choice -> policy
```
goldfirere: This parameter is passed at the one call-site of this function; it should not be optional.

typing/typetexp.mli
```diff
@@ -114,7 +114,8 @@ val transl_label_from_expr :
    of a constraint is typed using [Any] while the right hand side uses [Sort]. *)
 val transl_simple_type:
         Env.t -> new_var_jkind:jkind_initialization_choice
-        -> ?univars:TyVarEnv.poly_univars -> closed:bool -> Alloc.Const.t
+        -> ?univars:TyVarEnv.poly_univars
+        -> ?hint_for_unbound_variables:string -> closed:bool -> Alloc.Const.t
```
goldfirere: This hint is never needed outside of the module; do not expose this parameter. Instead, shadow `transl_simple_type` with the version without this parameter.

typing/typetexp.ml
```diff
@@ -1339,11 +1347,12 @@ let make_fixed_univars ty =
   make_fixed_univars ty;
   Btype.unmark_type ty
 
-let transl_simple_type env ~new_var_jkind ?univars ~closed mode styp =
+let transl_simple_type env ~new_var_jkind ?univars ?hint_for_unbound_variables
```
goldfirere: `hint_for_unbound_variables` is used only when `closed` is `true`. Maybe make `closed` a `string option`? Actually, that's not quite right, either. Instead, I think you want something like

```ocaml
type transl_simple_type_openness =
  | Open
  | Closed
  | Closed_for_upstream_compatibility
```

and then compose the text for the last case in the error-printing code. (Regardless of the relationship with `closed`, it would be better to have error strings in the error-printing code, instead of buried in other logic.) You'll probably want to change the extra field in the error structures to be a `bool` (or perhaps get another type -- types are cheap! -- that just captures the two "closed" conditions).

dkalinichenko-js: Yeah, I originally wrote a version that did this, then removed it to minimize the diff to other files, given that `Closed_for_upstream_compatibility` is only used once. But maybe that's not so bad.

goldfirere: But it doesn't need to affect other files -- the exported `transl_simple_type` can still take a `bool` and then just call the unexported (shadowed) `transl_simple_type`.

dkalinichenko-js: Yeah, that's the better option. Done.

typing/typetexp.ml
dkalinichenko-js: Removed the parameter.

typing/typetexp.mli
dkalinichenko-js: Removed the parameter.

PR #4153, opened 2025-06-19
Free stack cache at domain termination

stedolan:
Domain termination leaks memory because the stack cache is not freed. Since Oxcaml currently uses large main stacks by default, this is quite a lot of memory (and worse, distinct mappings) in programs that create and destroy very many domains.

(There aren't many such programs, but multicoretests is one: this bug prevents many multicoretests tests from completing)

### DIFF
--- a/runtime/domain.c
+++ b/runtime/domain.c
@@ -775,6 +775,7 @@ static void domain_create(uintnat initial_minor_heap_wsize,
   goto domain_init_complete;
 
 alloc_main_stack_failure:
+  caml_free_stack_cache(domain_state->stack_cache);
 create_stack_cache_failure:
   caml_remove_generational_global_root(&domain_state->dls_root);
 reallocate_minor_heap_failure:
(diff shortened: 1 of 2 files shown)


runtime/domain.c
```diff
@@ -2145,6 +2145,7 @@ static void domain_terminate (void)
   if(domain_state->current_stack != NULL) {
     caml_free_stack(domain_state->current_stack);
   }
+  caml_free_stack_cache(domain_state->stack_cache);
```
NickBarnes, line 2149: We should also call this at the `alloc_main_stack_failure` label in `domain_create()`.

runtime/fiber.c
```diff
@@ -942,6 +955,20 @@ struct stack_info* caml_alloc_main_stack (uintnat init_wsize)
   return stk;
 }
 
+static void free_stack_memory(struct stack_info* stack)
+{
+#if defined(DEBUG) && defined(STACK_CHECKS_ENABLED)
+  memset(stack, 0x42, (char*)stack->handler - (char*)stack);
+#endif
+#ifdef USE_MMAP_MAP_STACK
```
NickBarnes: I see that this is consistent with the directives in `alloc_for_stack`, but for more local consistency I'd prefer `#if defined(USE_MMAP_MAP_STACK)`

NickBarnes (approved):

PR #4158, opened 2025-06-19
`('a : value mod non_float) or_null` is `non_float`

dkalinichenko-js:
Change `or_null` to be `non_float` when its type argument is itself `non_float`. We can't express this behavior with
mod-bounds, so hardcode it in `estimate_type_jkind` for `or_null` itself and types marked as re-exporting `or_null` through the magic `[@@or_null_reexport]` attribute.

This change has no risk on its own, since `'a or_null` can't be put into arrays regardless. However, a subsequent PR will allow `'a or_null array`.

### DIFF
--- a/otherlibs/stdlib_stable/or_null.mli
+++ b/otherlibs/stdlib_stable/or_null.mli
@@ -24,9 +24,9 @@
 (* CR layouts: enable ocamlformat for this module when it starts supporting
    jkind annotations. *)
 
-type 'a t : immediate_or_null with 'a = 'a or_null [@@or_null_reexport]
+type 'a t : value_or_null = 'a or_null [@@or_null_reexport]
       (** The type of nullable values. Either [Null] or a value [This v].
-          ['a or_null] has a non-standard [immediate_or_null with 'a] layout,
+          ['a or_null] has a non-standard [value_or_null] layout,
           preventing the type constructor from being nested. *)
 
 val null : 'a t
--- a/otherlibs/stdlib_stable/or_null.mli
+++ b/otherlibs/stdlib_stable/or_null.mli
@@ -26,8 +26,8 @@
 
 type 'a t : value_or_null = 'a or_null [@@or_null_reexport]
       (** The type of nullable values. Either [Null] or a value [This v].
-          ['a or_null] has a non-standard [value_or_null] layout,
-          preventing the type constructor from being nested. *)
+          ['a or_null] accepts arguments of kind [value] while itself
+          having kind [value_or_null], forbidding [_ or_null or_null]. *)
 
 val null : 'a t
 (** [null] is [Null]. *)
--- a/testsuite/tests/typing-layouts-or-null/containers.ml
+++ b/testsuite/tests/typing-layouts-or-null/containers.ml
@@ -63,8 +63,8 @@ Line 1, characters 21-25:
                          ^^^^
 Error: This expression has type "'a or_null"
        but an expression was expected of type "('b : value)"
-       The kind of 'a or_null is immediate_or_null with 'a
-         because it is the primitive immediate_or_null type or_null.
+       The kind of 'a or_null is value_or_null
+         because it is the primitive value_or_null type or_null.
        But the kind of 'a or_null must be a subkind of value
          because it's the type of an array element,
          chosen to have kind value.
@@ -129,8 +129,8 @@ Line 1, characters 28-32:
                                 ^^^^
 Error: This expression has type "'a or_null"
        but an expression was expected of type "('b : value)"
-       The kind of 'a or_null is immediate_or_null with 'a
-         because it is the primitive immediate_or_null type or_null.
+       The kind of 'a or_null is value_or_null
+         because it is the primitive value_or_null type or_null.
        But the kind of 'a or_null must be a subkind of value
          because it's the type of an array element,
          chosen to have kind value.
@@ -260,7 +260,8 @@ let should_work_option3 = None
 
 [%%expect{|
 val should_work_option1 : float or_null option = Some (This 3.4)
-val should_work_option2 : 'a or_null option = Some Null
+val should_work_option2 :
+  ('a : value_or_null mod non_null). 'a or_null option = Some Null
 val should_work_option3 : 'a option = None
 |}]
 
--- a/testsuite/tests/typing-layouts-or-null/immediate.ml
+++ b/testsuite/tests/typing-layouts-or-null/immediate.ml
@@ -85,21 +85,9 @@ type should_work = int_or_null accept_immediate_or_null
 type should_work = int_or_null accept_immediate_or_null
 |}]
 
-(* CR layouts v2.8: this is a bug in principal inference with with-kinds. *)
-
 type should_work = int or_null accept_immediate_or_null
 [%%expect{|
 type should_work = int or_null accept_immediate_or_null
-|}, Principal{|
-Line 1, characters 19-30:
-1 | type should_work = int or_null accept_immediate_or_null
-                       ^^^^^^^^^^^
-Error: This type "int or_null" should be an instance of type
-         "('a : immediate_or_null)"
-       The kind of int or_null is immediate_or_null mod non_float with int
-         because it is the primitive immediate_or_null type or_null.
-       But the kind of int or_null must be a subkind of immediate_or_null
-         because of the definition of accept_immediate_or_null at line 1, characters 0-54.
 |}]
 
 (* Values. *)
@@ -118,8 +106,6 @@ external write_imm : ('a : immediate_or_null). 'a myref -> 'a -> unit
 external equal : ('a : immediate_or_null). 'a -> 'a -> bool = "%equal"
 |}]
 
-(* CR layouts v2.8: this is a bug in principal inference with with-kinds. *)
-
 let () =
   let r = { v = (Null : int or_null) } in
   let x = read_imm r in
@@ -129,14 +115,4 @@ let () =
 ;;
 
 [%%expect{|
-|}, Principal{|
-Line 2, characters 16-36:
-2 |   let r = { v = (Null : int or_null) } in
-                    ^^^^^^^^^^^^^^^^^^^^
-Error: This expression has type "int or_null"
-       but an expression was expected of type "('a : immediate_or_null)"
-       The kind of int or_null is immediate_or_null mod non_float with int
-         because it is the primitive immediate_or_null type or_null.
-       But the kind of int or_null must be a subkind of immediate_or_null
-         because of the definition of myref at line 1, characters 0-56.
 |}]
--- a/testsuite/tests/typing-layouts-or-null/immediate.ml
+++ b/testsuite/tests/typing-layouts-or-null/immediate.ml
@@ -116,3 +116,14 @@ let () =
 
 [%%expect{|
 |}]
+
+(* [immediate_or_null] is [non_float]: *)
+
+type ('a : value_or_null mod non_float) accepts_nonfloat
+
+type succeeds = t_immediate_or_null accepts_nonfloat
+
+[%%expect{|
+type ('a : value_or_null mod non_float) accepts_nonfloat
+type succeeds = t_immediate_or_null accepts_nonfloat
+|}]
--- a/testsuite/tests/typing-layouts-or-null/immediate.ml
+++ b/testsuite/tests/typing-layouts-or-null/immediate.ml
@@ -96,7 +96,7 @@ Line 1, characters 19-30:
                        ^^^^^^^^^^^
 Error: This type "int or_null" should be an instance of type
          "('a : immediate_or_null)"
-       The kind of int or_null is immediate_or_null with int
+       The kind of int or_null is immediate_or_null mod non_float with int
          because it is the primitive immediate_or_null type or_null.
        But the kind of int or_null must be a subkind of immediate_or_null
          because of the definition of accept_immediate_or_null at line 1, characters 0-54.
@@ -135,7 +135,7 @@ Line 2, characters 16-36:
                     ^^^^^^^^^^^^^^^^^^^^
 Error: This expression has type "int or_null"
        but an expression was expected of type "('a : immediate_or_null)"
-       The kind of int or_null is immediate_or_null with int
+       The kind of int or_null is immediate_or_null mod non_float with int
          because it is the primitive immediate_or_null type or_null.
        But the kind of int or_null must be a subkind of immediate_or_null
          because of the definition of myref at line 1, characters 0-56.
--- a/testsuite/tests/typing-layouts-or-null/reexport.ml
+++ b/testsuite/tests/typing-layouts-or-null/reexport.ml
@@ -18,6 +18,7 @@ Error: This variant or record definition does not match that of type
          "'a or_null"
        Their internal representations differ:
        the original definition has a null constructor.
+       Hint: add [@@or_null_reexport].
 |}]
 
 module Or_null = struct
--- a/testsuite/tests/typing-layouts-or-null/reexport.ml
+++ b/testsuite/tests/typing-layouts-or-null/reexport.ml
@@ -160,7 +160,7 @@ Error: This expression has type "'a Or_null.t" = "'a or_null"
        The kind of 'a Or_null.t is value_or_null
          because it is the primitive value_or_null type or_null.
        But the kind of 'a Or_null.t must be a subkind of
-         value_or_null mod non_null
+           value_or_null mod non_null
          because of the definition of t at line 2, characters 2-63.
 |}]
 
--- a/testsuite/tests/typing-layouts-or-null/reexport.ml
+++ b/testsuite/tests/typing-layouts-or-null/reexport.ml
@@ -5,23 +5,23 @@
 (* CR layouts v3.5: ['a or_null] can't be re-exported normally,
    because users can't define their own [Null]-like constructors. *)
 module Or_null = struct
-  type ('a : value) t : immediate_or_null with 'a = 'a or_null =
+  type ('a : value) t : value_or_null = 'a or_null =
     | Null
     | This of 'a
 end
 [%%expect{|
 Lines 2-4, characters 2-16:
-2 | ..type ('a : value) t : immediate_or_null with 'a = 'a or_null =
+2 | ..type ('a : value) t : value_or_null = 'a or_null =
 3 |     | Null
 4 |     | This of 'a
-Error: The kind of type "t" is immutable_data with 'a
-         because it's a boxed variant type.
-       But the kind of type "t" must be a subkind of immediate_or_null with 'a
-         because of the annotation on the declaration of the type t.
+Error: This variant or record definition does not match that of type
+         "'a or_null"
+       Their internal representations differ:
+       the original definition has a null constructor.
 |}]
 
 module Or_null = struct
-  type ('a : value) t : immediate_or_null with 'a = 'a or_null
+  type ('a : value) t : value_or_null = 'a or_null
 end
 [%%expect{|
 module Or_null : sig type 'a t = 'a or_null end
@@ -50,7 +50,7 @@ Error: Unbound constructor "Or_null.This"
 (* [@@or_null_reexport] re-exports those constructors. *)
 
 module Or_null = struct
-  type ('a : value) t : immediate_or_null with 'a = 'a or_null [@@or_null_reexport]
+  type ('a : value) t : value_or_null with 'a = 'a or_null [@@or_null_reexport]
 end
 let n = Or_null.Null
 let t v = Or_null.This v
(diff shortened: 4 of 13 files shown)


testsuite/tests/typing-layouts-or-null/immediate.ml
```diff
@@ -96,7 +96,7 @@ Line 1, characters 19-30:
                        ^^^^^^^^^^^
 Error: This type "int or_null" should be an instance of type
          "('a : immediate_or_null)"
-       The kind of int or_null is immediate_or_null with int
+       The kind of int or_null is immediate_or_null mod non_float with int
```
goldfirere: Hmm. `immediate` should imply `non_float`. It's impossible to have a non-pointer float.

testsuite/tests/typing-layouts-or-null/separability.ml
```diff
@@ -856,3 +840,122 @@ type ('c : value & value & float64 mod non_null non_float) fails = unit constrai
 type c = #(float * float or_null * float#)
 type 'a fails = unit constraint 'a = c
 |}]
+
+(* Separability and [@@or_null_reexport]. *)
+
+(* Re-exported [or_null] types preserve separability behavior. *)
+
+module Or_null_reexport = struct
```
goldfirere, line 783: Do we allow something like

```ocaml
module M : sig
  type 'a t [@@or_null_reexport]
end = struct
  type 'a t = 'a or_null [@@or_null_reexport]
end
```

? Before this patch, there was no reason to allow this, because `type ('a : value) t : value_or_null mod everything with 'a` was sufficient to specify the extensional behavior of `or_null`. But now, there is new kinding magic, and one could imagine wanting to support this without exposing the equality to `or_null` itself.

I don't think this needs to be in this patch (or ever, really, unless/until there is a demand), but perhaps we should make a ticket to consider doing this. (Maybe at the minimum we should document that we've left this tiny bit of expressiveness on the table.)

typing/ctype.ml
```diff
@@ -2303,15 +2323,24 @@ let rec estimate_type_jkind ~expand_component env ty =
...
          intersection types sometimes do not unify with themselves. Removing
          this check causes typing-misc/pr7937.ml to fail. *)
-      if Jkind.has_with_bounds jkind && List.compare_length_with args 0 <> 0
-      then
-        let level = get_level ty in
-        (* CR layouts v2.8: We could possibly skip this substitution if we're
-           called from [constrain_type_jkind]; the jkind returned without
-           substing is just weaker than the one we would get by substing. *)
-        jkind_subst env level type_decl.type_params args jkind
+      let substituted_jkind =
```
goldfirere: I don't think this is quite the right implementation strategy.

Instead, I think we should view `or_null` as closer to an `[@@unboxed]` type. Interestingly, `estimate_type_jkind` does not treat `[@@unboxed]` types specially. The special treatment for `[@@unboxed]` types comes in `constrain_type_jkind`, which knows how to look inside them. This is all ok, because `estimate_type_jkind` can just return an upper bound. This means that the separability of `ty or_null` can be `maybe_separable` for any `ty` -- as returned from `estimate_type_jkind`. But we must work harder in `constrain_type_jkind` to make up for this.

The way I think would work better is to teach `unbox_once` about `or_null`. `unbox_once` could return a new constructor `Stepped_or_null` (or something) in the `or_null` case. Critically, in `constrain_type_jkind`, before recurring on the type in the `or_null`, we have to adjust the `jkind` passed to `loop`, much like the existing use of `Jkind.apply_modality_r` call on line 2572.

I think this new implementation strategy is more "with the grain" of current things, and it also should empower more inference: if we have `val f : 'a or_null accepts_non_float -> ...`, we'll be able to infer that `'a` is `mod non_float` too. (I don't think this works with the current implementation.)

testsuite/tests/typing-layouts-or-null/immediate.ml
dkalinichenko-js: Fixed. Before, `immediate_or_null` had to be maybe-separable because `'a or_null` had kind `immediate_or_null with 'a`, but now the kind is `value_or_null` with magic.

testsuite/tests/typing-layouts-or-null/separability.ml
dkalinichenko-js: I think this is the same behavior as the magic behavior of unboxed types. I don't expect us to need re-exposing it like that.

typing/ctype.ml
dkalinichenko-js: Yeah, you're right. We want to eventually unify `Variant_with_null` with `Variant_unboxed` anyway.

Done.

testsuite/tests/typing-layouts-or-null/separability.ml
goldfirere: How so? An `[@@unboxed]` type has the same kind as its inner type, but that's easy to express via a kind annotation. Or maybe you're saying that we can't give an abstract kind to `type 'a t = { x : 'a } [@@unboxed]` ? Yes that's true today, but we'll eventually be able to do this. In any case, I'm ok if you don't see anything worth doing here and just Resolve.

otherlibs/stdlib_stable/or_null.mli
```diff
@@ -24,9 +24,9 @@
 (* CR layouts: enable ocamlformat for this module when it starts supporting
    jkind annotations. *)
 
-type 'a t : immediate_or_null with 'a = 'a or_null [@@or_null_reexport]
+type 'a t : value_or_null = 'a or_null [@@or_null_reexport]
       (** The type of nullable values. Either [Null] or a value [This v].
-          ['a or_null] has a non-standard [immediate_or_null with 'a] layout,
+          ['a or_null] has a non-standard [value_or_null] layout,
           preventing the type constructor from being nested. *)
```
goldfirere: Not new in this patch, but I don't understand this comment.

testsuite/tests/typing-layouts-or-null/containers.ml
```diff
@@ -63,8 +63,8 @@ Line 1, characters 21-25:
                          ^^^^
 Error: This expression has type "'a or_null"
        but an expression was expected of type "('b : value)"
-       The kind of 'a or_null is immediate_or_null with 'a
-         because it is the primitive immediate_or_null type or_null.
+       The kind of 'a or_null is value_or_null
```
goldfirere, line 66: Why the change from `immediate_or_null with 'a`? I think `immediate_or_null with 'a` is more accurate. Is the kind of `'a` getting defaulted to `value`?

testsuite/tests/typing-layouts-or-null/reexport.ml
```diff
@@ -5,23 +5,23 @@
...
 3 |     | Null
 4 |     | This of 'a
-Error: The kind of type "t" is immutable_data with 'a
-         because it's a boxed variant type.
-       But the kind of type "t" must be a subkind of immediate_or_null with 'a
-         because of the annotation on the declaration of the type t.
+Error: This variant or record definition does not match that of type
+         "'a or_null"
+       Their internal representations differ:
+       the original definition has a null constructor.
```
goldfirere: This is a terrible error message. Let's at least leave a CR to return.

testsuite/tests/typing-layouts-or-null/reexport.ml
```diff
@@ -129,7 +129,8 @@ type ('a : float64) t = 'a or_null [@@or_null_reexport]
 Line 1, characters 24-26:
 1 | type ('a : float64) t = 'a or_null [@@or_null_reexport]
                             ^^
-Error: This type "('a : float64)" should be an instance of type "('b : value)"
+Error: This type "('a : float64)" should be an instance of type
+         "('b : value_or_null mod non_null)"
```
goldfirere, line 134: This is very strange.

testsuite/tests/typing-layouts-or-null/separability.ml
```diff
@@ -549,19 +561,10 @@ Line 1, characters 13-26:
...
-Line 1, characters 13-26:
-1 | type fails = float or_null accepts_sep
-                 ^^^^^^^^^^^^^
-Error: This type "float or_null" should be an instance of type
-         "('a : any mod separable)"
-       The kind of float or_null is immediate_or_null with float
-         because it is the primitive immediate_or_null type or_null.
-       But the kind of float or_null must be a subkind of any mod separable
+         value mod many unyielding stateless immutable
+         because it is the primitive type float.
```
goldfirere: No it's not...

testsuite/tests/typing-layouts-or-null/separability.ml
```diff
@@ -600,18 +603,8 @@ Line 1, characters 22-33:
...
-         because it's the type argument to the array type.
-|}, Principal{|
-Line 1, characters 22-33:
-1 | type should_succeed = int or_null array
-                          ^^^^^^^^^^^
-Error: This type "int or_null" should be an instance of type
-         "('a : any_non_null)"
-       The kind of int or_null is immediate_or_null mod non_float with int
-         because it is the primitive immediate_or_null type or_null.
+       The kind of int or_null is value_or_null
```
goldfirere, line 603: But `int or_null` should be `immediate_or_null`. I think.

testsuite/tests/typing-layouts-or-null/separability.ml
```diff
@@ -736,14 +684,22 @@ type ('a : value_or_null) smth : immediate with 'a
...
-(* CR layouts v2.8: broken mode crossing with [-principal]. *)
+(* CR layouts v2.8: with-bounds don't peek through unboxed-like types. *)
 
 type works = int or_null smth bounded
 
 [%%expect{|
 type 'a smth : immediate with 'a
 type ('a : immediate) bounded
-type works = int or_null smth bounded
+Line 7, characters 13-29:
```
goldfirere, line 691: Why is this accepted in principal mode? It should be rejected, as far as I can tell. This is a little worrying.

testsuite/tests/typing-layouts-or-null/separability.ml
```diff
@@ -915,21 +854,10 @@ Line 1, characters 13-37:
...
-|}, Principal{|
-Line 1, characters 13-37:
-1 | type fails = float Or_null_reexport.t accepts_nonfloat
-                 ^^^^^^^^^^^^^^^^^^^^^^^^
-Error: This type "float Or_null_reexport.t" = "float or_null"
-       should be an instance of type "('a : any mod non_float)"
-       The kind of float Or_null_reexport.t is immediate_or_null with float
-         because it is the primitive immediate_or_null type or_null.
+         value mod many unyielding stateless immutable
+         because it is the primitive type float.
```
goldfirere: No it's not.

typing/ctype.ml
```diff
@@ -2207,6 +2209,14 @@ let unbox_once env ty =
                                    modality = ld.ld_modalities }) lbls)
         | Type_record_unboxed_product ([], _, _) ->
           Misc.fatal_error "Ctype.unboxed_once: fieldless record"
+        | Type_variant ([_; cd2], Variant_with_null, _) ->
+          begin match cd2.cd_args with
+          | Cstr_tuple [arg] ->
+            Stepped_or_null { ty = apply arg.ca_type;
+                              is_open = false;
+                              modality = arg.ca_modalities }
+          | _ -> Final_result
```
goldfirere: This is impossible, right?

typing/ctype.ml
```diff
@@ -2207,6 +2209,14 @@ let unbox_once env ty =
                                    modality = ld.ld_modalities }) lbls)
         | Type_record_unboxed_product ([], _, _) ->
           Misc.fatal_error "Ctype.unboxed_once: fieldless record"
+        | Type_variant ([_; cd2], Variant_with_null, _) ->
+          begin match cd2.cd_args with
+          | Cstr_tuple [arg] ->
+            Stepped_or_null { ty = apply arg.ca_type;
+                              is_open = false;
+                              modality = arg.ca_modalities }
```
goldfirere, line 2219: This is always the identity, right? (but maybe not tomorrow)

typing/ctype.ml
```diff
@@ -2521,6 +2532,41 @@ let constrain_type_jkind ~fixed env ty jkind =
                   (Not_a_subjkind (ty's_jkind, jkind, sub_failure_reasons)))
              end
           in
+          let or_null ~fuel ty is_open modality =
```
goldfirere, line 2539: This still isn't quite right, I think. Closer! But not quite there.

* It should be implemented within jkind.ml, much like `apply_modality_r`.
* It needs to work only on r-kinds. These don't have with-bounds, and so you won't need to pass in `estimate_type_jkind` (which actually should probably be `type_jkind_purely_if_principal`... but you won't need it).
* Within jkind, you can just look directly at the mod_bounds.
* Actually, `get_nullability` and `get_separability` should probably never be called on an r-kind. The temptation to do so is probably a bug. The operations make sense... but almost certainly will lead to strange type inference. And indeed if I change the types of these functions to accept only l-kinds, they work in every scenario except this new one.
* This will be *much* shorter in jkind!

typing/jkind.ml
```diff
@@ -2887,6 +2908,26 @@ let set_separability_upper_bound jk separability_upper_bound =
       }
   }
 
+let get_separability ~jkind_of_type jk =
```
goldfirere: With the change suggested above, this function can be removed.

typing/predef.ml
```diff
@@ -678,12 +678,8 @@ let or_null_kind tvar =
...
   Type_variant (cstrs, Variant_with_null, None)
 
-let or_null_jkind param =
-  Jkind.Builtin.immediate_or_null ~why:(Primitive ident_or_null) |>
-  Jkind.add_with_bounds
-    ~modality:Mode.Modality.Value.Const.id
-    ~type_expr:param |>
-  Jkind.mark_best
+let or_null_jkind _ =
+  Jkind.Builtin.value_or_null ~why:(Primitive ident_or_null)
```
goldfirere, line 682: This kind surprises me a bit, as I suggest in my comments on tests. I suppose it doesn't matter, though, because we'll always look through an `or_null`. Maybe there's no good answer here?

testsuite/tests/typing-layouts-or-null/containers.ml
dkalinichenko-js: `immediate_or_null with 'a` does not reflect that `'a or_null` is `non_float` only if `'a` itself is `non_float`. Either we make `immediate_or_null` `maybe_separable` (confusing), or we assign a more general kind to `or_null` (also confusing, but reflects existing magic rules).

testsuite/tests/typing-layouts-or-null/reexport.ml
dkalinichenko-js: `value_or_null mod non_null` really should be `value_maybe_separable`.

dkalinichenko-js: Maybe we should come up with a name to describe the layout like `pointer`? So `pointer mod non_null` makes sense.

testsuite/tests/typing-layouts-or-null/separability.ml
dkalinichenko-js: It's rejected in the principal mode? Just with a different error message.

typing/ctype.ml
dkalinichenko-js: Yep, yep.

typing/ctype.ml
dkalinichenko-js: Thanks for suggestions, implemented.

typing/ctype.ml
dkalinichenko-js: Yep, added an error if we ever reach this.

testsuite/tests/typing-layouts-or-null/separability.ml
dkalinichenko-js: Thanks for catching this one, fixed error reporting for it.

testsuite/tests/typing-layouts-or-null/separability.ml
dkalinichenko-js: We could print a better error here.

testsuite/tests/typing-layouts-or-null/reexport.ml
dkalinichenko-js: Added a hint, is it better now?

otherlibs/stdlib_stable/or_null.mli
dkalinichenko-js: Is it better now?

typing/predef.ml
dkalinichenko-js: Yeah, I'm not sure what to do here.

testsuite/tests/typing-layouts-or-null/containers.ml
goldfirere: OK. Fine.

testsuite/tests/typing-layouts-or-null/reexport.ml
goldfirere: Ah. At least now I understand why this output is this way.

I don't love `pointer`, because `int` is not a pointer. `scannable`? Or maybe `gen_value` (for "generalized")? That is, we want it to have the layout of `value` but otherwise be top.

testsuite/tests/typing-layouts-or-null/separability.ml
goldfirere: Oh, whoops

testsuite/tests/typing-layouts-or-null/reexport.ml
goldfirere: A bit. Still pretty confusing. I mean, the definition as given has a `Null` constructor, after all.

otherlibs/stdlib_stable/or_null.mli
goldfirere: Yes, thanks

typing/predef.ml
goldfirere: Yeah, fine, let's move on.

typing/ctype.ml
goldfirere: I think it's helpful to leave a comment saying this. It's a bit surprising to have a data field that (today) never carries any information.

typing/ctype.ml
```diff
@@ -2521,6 +2532,31 @@ let constrain_type_jkind ~fixed env ty jkind =
...
+            let error =
```
goldfirere: OCaml is strict! So this structure gets allocated unconditionally, even on the happy path. Boo. Better to make this into a thunk (e.g. `let error () = ...`).

typing/ctype.ml
```diff
@@ -2521,6 +2532,31 @@ let constrain_type_jkind ~fixed env ty jkind =
...
+              Error (Jkind.Violation.of_ ~jkind_of_type
+                (Not_a_subjkind (ty's_jkind, jkind, sub_failure_reasons)))
+            in
+            let jkind = Jkind.apply_modality_r modality jkind in
+            match
+              Jkind.apply_or_null jkind
+            with
+            | Ok jkind ->
+              (match
+                loop ~fuel:(fuel - 1) ~expanded:false ty ~is_open
```
goldfirere: This fuel deduction is redundant with the one at the call site of `or_null`. (Actually, we don't really need to deduct fuel at all in this case; the kinds stop us from doing it forever. But using fuel is ok.)

goldfirere (approved):

testsuite/tests/typing-layouts-or-null/separability.ml
goldfirere: Not sure how to interpret your comment....

testsuite/tests/typing-layouts-or-null/reexport.ml
dkalinichenko-js: Will add in a subsequent PR.

goldfirere: OK.

testsuite/tests/typing-layouts-or-null/separability.ml
dkalinichenko-js: I think I know the place to change it to have a better error. Left a CR there.

testsuite/tests/typing-layouts-or-null/reexport.ml
dkalinichenko-js: Expanded to make it a bit clearer.

typing/ctype.ml
dkalinichenko-js: Fixed.

typing/ctype.ml
dkalinichenko-js: Fixed.

typing/ctype.ml
dkalinichenko-js: Fixed.

PR #4159, opened 2025-06-20
Fix table syntax

cuihtlauac:


goldfirere:
Is there something wrong with the old syntax? The original syntax works with table-mode in emacs, but the new syntax won't.

cuihtlauac:
Current syntax isn't rendered properly on the web site

goldfirere (approved):
Ah, indeed.

PR #4160, opened 2025-06-20
Improve newlines in kind errors

goldfirere:
After having trouble parsing kind errors (and I should be in the set of people with the most context for these), I wanted to adjust their newlines. This patch does it.

Review: @liam923.

Review recommendation:
* Spot-check several error message changes
* Look at the code changes to see that all it does is mess with newlines and spacing


testsuite/tests/typing-modal-kinds/unboxed.ml
```diff
@@ -234,11 +234,11 @@ Line 1, characters 10-24:
 Error: This type "(int -> int) t" should be an instance of type
          "('a : value mod portable & value mod portable)"
        The kind of (int -> int) t is
-         immediate with int -> int @@ portable & immediate
-         with int -> int @@ portable
+           immediate with int -> int @@ portable &
```
liam923: I have a (weak) preference for having the ampersand at the start of the second line, rather than the end of the first. In support of this view, I'll point out that this is what ocamlformat does, and I think there's value in having consistency between error message formatting and ocamlformat formatting.

typing/oprint.ml
```diff
@@ -621,19 +621,29 @@ and print_out_label ppf (name, mut, arg, gbl) =
...
+        let fix_indentation ppf =
+          let { out_newline; out_indent } as out_functions =
+            pp_get_formatter_out_functions ppf ()
+          in
+          let out_newline () =
+            out_newline ();
+            out_indent 18  (* this works well in practice *)
+          in
+          pp_set_formatter_out_functions ppf { out_functions with out_newline }
+        in
```
liam923: I don't follow what's going on here. Is this setting the indentation to be 18 characters?

testsuite/tests/typing-jkind-bounds/modalities.ml
```diff
@@ -137,11 +137,11 @@ Line 1, characters 64-65:
 Error: This expression has type "(string -> string) t"
        but an expression was expected of type "('a : value & value)"
        The kind of (string -> string) t is
-         immediate with string -> string @@ global & immediate
-         with string -> string @@ global
+           immediate with string -> string @@ global &
+           immediate with string -> string @@ global
```
liam923: This is probably outside the scope of this PR, but I think this kind is wrong. iirc, `&` binds tighter than `with`, so this is `(immediate with string -> string @@ global & immediate) with string -> string @@ global`. But what we actually mean is `(immediate with string -> string @@ global) & (immediate with string -> string @@ global)`. (ocamlformat seems to align with my memory.)

testsuite/tests/typing-modal-kinds/unboxed.ml
goldfirere: I liked the vertical alignment of my new output. But I'll change it back, because I agree with your reasoning.

typing/oprint.ml
goldfirere: I've expanded the comment

goldfirere: Resolve if you like it

testsuite/tests/typing-jkind-bounds/modalities.ml
goldfirere: I've made internal ticket 4547 to track this; well spotted. But I don't want to derail the current PR.

liam923 (approved):

PR #4161, opened 2025-06-20
Add hint to challenging error message around kinds

goldfirere:
As requested by a colleague.

The main body of the error message is actually accurate: if there were no kind annotation on the type involved, the code would be accepted (with a bizarre type that the user probably did not intend). But it is easy to detect this case and issue a hint, as I've done here.

Review: @rtjoa (pretty arbitrarily)

Recommend review by commits.

### DIFF
--- a/testsuite/tests/typing-layouts/basics.ml
+++ b/testsuite/tests/typing-layouts/basics.ml
@@ -2916,9 +2916,9 @@ let g t = M.f t ~key:0
 [%%expect{|
 type ('v : immediate) t
 module M : sig val f : ('v : immediate). 'v t -> int -> 'v end
-Line 9, characters 10-13:
+Line 9, characters 10-15:
 9 | let g t = M.f t ~key:0
-              ^^^
+              ^^^^^
 Error: This expression is used as a function, but its type "'a"
        has kind "immediate", which cannot be the kind of a function.
        (Functions always have kind "value mod aliased immutable non_float".)
--- a/testsuite/tests/typing-layouts/basics.ml
+++ b/testsuite/tests/typing-layouts/basics.ml
@@ -2897,3 +2897,29 @@ let f (x : ('a : value mod uncontended)) = x ()
 val f : (unit -> 'a) -> 'a = <fun>
 val f : (unit -> 'a) -> 'a = <fun>
 |}]
+
+(***************************************)
+(* Test 47: Error message on bad label *)
+
+(* reduced from a test case in the wild *)
+
+type ('v : immediate) t
+
+module M : sig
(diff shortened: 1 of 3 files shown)


typing/typecore.ml
```diff
@@ -3955,8 +3956,20 @@ let collect_unknown_apply_args env funct ty_fun mode_fun rev_args sargs ret_tvar
...
-                raise(Error(funct.exp_loc, env,
-                            Impossible_function_jkind (ty_fun, jkind)))
+                (* need to calculate a location containing the function
+                   and any arguments already processed *)
+                let get_loc (_, arg) = match arg with
+                  | Arg (Known_arg { sarg }
+                         | Unknown_arg { sarg }) -> Some sarg.pexp_loc
+                  | Arg (Eliminated_optional_arg _) -> None
+                  | Omitted _ -> None
+                in
```
rtjoa: Nit: the same function also appears to be used ~45 lines below, in the calculation of `previous_arg_loc`. Consider pulling out into a helper? (But, it's not much code - up to you.)

testsuite/tests/typing-layouts/basics.ml
```diff
@@ -2897,3 +2897,30 @@ let f (x : ('a : value mod uncontended)) = x ()
...
+[%%expect{|
+type ('v : immediate) t
+module M : sig val f : ('v : immediate). 'v t -> int -> 'v end
+Line 9, characters 10-15:
+9 | let g t = M.f t ~key:0
+              ^^^^^
+Error: This expression is used as a function, but its type "'a"
+       has kind "immediate", which cannot be the kind of a function.
+       (Functions always have kind "value mod aliased immutable non_float".)
+       Hint: Perhaps you have over-applied the function or used an incorrect label.
```
rtjoa: When I move the labeled argument before the unlabeled one, the locations get merged to include all of `M.f ~key:0 t`:
```
17 | let g t = M.f ~key:0 t
               ^^^^^^^^^^^^
Error: This expression is used as a function, but its type 'a ...
```

I can't think of a good way to say that just `M.f` and `t` comprise the expression used as a function. But maybe we can include the current argument, and say something like the following?
```
17 | let g t = M.f ~key:0 t
               ^^^^^^^^^^^^
Error: This function application uses an expression with type 'a
as a function, but that type has kind "immediate", which cannot
be the kind of a function.
```

goldfirere: Yes good point.

goldfirere:
Thanks for the review @rtjoa ; back to you.

rtjoa (approved):

PR #4166, opened 2025-06-23
Restore unboxing of closures in the reaper

Ekdohibs:
This was removed from the unbox-fv-closures PR to have a minimum working version of it. Now that the reaper is merged, we can look at this separately. However, it currently doesn't pass the testsuite, that looks to be caused by a problem with SIMD vectors passed on the stack.


mshinwell:
The segfault should be fixed by #4171.
In this context, "unboxing of closures" is what is usually called "lambda lifting".

PR #4167, opened 2025-06-23
Use consistent control flow throughout domain_create

NickBarnes:
This fixes a couple of hard-to-hit bugs in domain_create, and more importantly makes the control flow there wholly consistent:
- domain lock is unlocked on creation failure;
- minor heap is freed on creation failure;
- interrupt word is cleared on creation failure;
- every failure mode is `goto fail_<name>`;
- resource-allocating calls which which will exit rather than return a fail are commented.

I was provoked into this by having to check through the code for review of #4153.

### DIFF
--- a/runtime/domain.c
+++ b/runtime/domain.c
@@ -601,8 +601,9 @@ static void domain_create(uintnat initial_minor_heap_wsize,
 
   d = next_free_domain();
 
-  if (d == NULL)
-    goto domain_init_complete;
+  if (d == NULL) {
+    goto fail_domain;
+  }
 
   s = &d->interruptor;
   CAMLassert(!s->running);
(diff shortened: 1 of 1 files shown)


runtime/domain.c
```diff
@@ -772,30 +774,34 @@ static void domain_create(uintnat initial_minor_heap_wsize,
...
   add_next_to_stw_domains();
-  goto domain_init_complete;
+  CAML_GC_MESSAGE(DOMAIN, "Creation complete.\n");
+  caml_plat_unlock(&all_domains_lock);
+  return;
 
-alloc_main_stack_failure:
-create_stack_cache_failure:
+fail_main_stack:
+  // caml_free_stack_cache()
```
stedolan: Should this be uncommented?

NickBarnes: It should now that the function exists! Rebased and uncommented.

stedolan (approved):

PR #4168, opened 2025-06-23
Add missing locks for generative functor and functor of anonymous parameter

riaqn:
What the title says.


dkalinichenko-js (approved):
LGTM.

PR #4171, opened 2025-06-23
Unaligned domainstate args

TheNumbat:


backend/amd64/emit.ml
```diff
@@ -919,6 +919,12 @@ let emit_global_label ~section s =
   let lbl = Cmm_helpers.make_symbol s in
   emit_global_label_for_symbol ~section lbl
 
+let loc_is_domainstate = function
+  | { loc = Stack (Domainstate _); _ } -> true
+  | { loc = Stack (Incoming _ | Outgoing _ | Reg.Local _) | Reg _ | Unknown; _ }
+    ->
+    false
+
```
mshinwell: This should move to `Reg`.

mshinwell (approved):
Just one thing to fix

PR #4174, opened 2025-06-23
Generate unboxed types tests

rtjoa:
Generate tests exercising interesting unboxed types. This was already used for nested records (#3687), and will be used for block indices (#4017) and void-in-blocks (to come). We commit this infrastructure now so we can decouple the latter two and reduce their marginal diff.

### Reviewing
A cursory review should be fine for this PR as it only adds testing; the ultimate responsibility for ensuring future changes are well-tested falls to their PRs. That said:
- Review commit-by-commit.
- The large diff in 98360aa426ce8b0e4c15a58c25485c4d7d9cb839 is because this PR adds types to test (records containing only `float` and `float#`) and generated comments.


ccasin (approved):
Per discussion, I did not read the generation code in detail, but I did skim some of the test diff and it looks fine.

PR #4178, opened 2025-06-24
Arm64: remove rosseta check

gretay-js:
`Test_helpers.run_if_not_under_rosetta2` doesn't do anything on amd64 (the check returns `Unknown` when I test it locally on an amd64 machine, and the checks don't run). Now that we have a separate arm64 implementation of these intrinsics, we can removed this check from  both shared and amd64-only tests.

### DIFF
--- a/oxcaml/tests/simd/amd64/sse_other_ops.ml
+++ b/oxcaml/tests/simd/amd64/sse_other_ops.ml
@@ -39,7 +39,7 @@ module Float32x4 = struct
         else eq_float32x4 ~result ~expect)
 
   let () =
-    Test_helpers.run_if_not_under_rosetta2 ~f:(fun () ->
+
         Float32.check_floats (fun f0 f1 ->
             (failmsg
                := fun () ->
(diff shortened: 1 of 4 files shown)


xclerc:
I am not saying it should not be removed,
but it is expected to do nothing if the box
is a physical amd64. The only use of this
code was when asking a physical arm64
to behave as an amd64 box.

mshinwell:
@xclerc tests on Rosetta, please discuss with him first.

xclerc:
I have just checked that, with Rosetta2 on,
I get the same output for `make ci` before
and after this pull request.

There are a handful or failures, but they are
(i) the same and (ii) not related to the tests
this pull request touches.

(The failures are about duplicate symbols in
cmm tests, and probes.)

I have also checked that with Rosetta2 off,
`make ci` does not report any failure.

xclerc:
Yes, I think it is. It vaguely remember it being
useful at the start of the Apple Silicon rollout
in GitHub CI, and I think I read somewhere
that Rosetta2 will be deprecated in about a year.

PR #4179, opened 2025-06-24
Do not include fail_if_called_indirectly in data section for shared outputs

Gbury:
The current scheme always generates the `fail_if_called_indirectly` function, which makes the linker fail when some shared object includes it and is later linked with a binary that also includes it. This PR should ensure that we only generate the function once when we emit a binary (we may later want to see whether we can include it in the cache for `caml_apply` functions, which already does something along these lines).


mshinwell:
Not sure this is correct yet, as discussed with @Gbury

PR #4180, opened 2025-06-24
Fix test order for inlining conditions

chambart:
This prevents any option from overriding the main environment condition.

This made inlining possible when simplifying stub function which is a massive compilation performance problem.


lthls (approved):

Gbury (approved):

mshinwell:
Superceded by #4188

PR #4181, opened 2025-06-24
Add missing inlining annotations to Name.pattern_match

lthls:
We noticed some anonymous functions passed to `Name.pattern_match`  in perf profiles. This is likely not a big problem, but given the typical usage of `Name.pattern_match` it seems better to always inline.

### DIFF
--- a/middle_end/flambda2/identifiers/int_ids.ml
+++ b/middle_end/flambda2/identifiers/int_ids.ml
@@ -831,7 +831,7 @@ module Code_id_or_symbol = struct
 
   let create_symbol symbol = symbol
 
-  let pattern_match t ~code_id ~symbol =
+  let[@inline always] pattern_match t ~code_id ~symbol =
     let flags = Table_by_int_id.Id.flags t in
     if flags = Code_id_data.flags
     then (code_id [@inlined hint]) t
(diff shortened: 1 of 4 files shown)


middle_end/flambda2/identifiers/int_ids.ml
```diff
@@ -549,9 +549,9 @@ module Name = struct
   let[@inline always] pattern_match t ~var ~symbol =
     let flags = Id.flags t in
     if flags = var_flags
-    then var t
+    then (var[@inline hint]) t
```
Gbury:
```suggestion
    then (var[@inlined hint]) t
```

middle_end/flambda2/identifiers/int_ids.ml
```diff
@@ -549,9 +549,9 @@ module Name = struct
...
     else if flags = symbol_flags
-    then symbol t
+    then (symbol[@inline hint]) t
```
Gbury:
```suggestion
    then (symbol[@inlined hint]) t
```

Ekdohibs:
There are several other `pattern_match` functions in the same file which would benefit from this as well, I think.

Gbury (approved):

PR #4183, opened 2025-06-24
Sharing on patricia trees

chambart:
Extends the https://github.com/oxcaml/oxcaml/pull/4073 PR.

This adds some important inlining annotations, and add a new helper function.

This removes Shared_set and map as the type is not obviously the good one.

It also add the update_and_extend function that is useful to optimize a quadratic case in dataflow


middle_end/flambda2/algorithms/patricia_tree.ml
```diff
@@ -449,48 +461,193 @@ end = struct
...
       else t
 
-  (* CR pchambart: union x x is expensive, while it could be O(1). This would
-     require that we demand f x x = x *)
-  (* CR-someday lmaurer: Generalize [merge] so that we can implement this in
-     terms of it. Rather than one callback, it should take three, along the
-     lines of Haskell's implementation; this would let us preserve sharing (see
-     Haskell's [Data.IntMap] for details). *)
-  let rec union f t0 t1 =
+  let[@inline always] universal_merge ?shortcut ?check0 ?check1 ~only_left
```
bclement-ocp, line 464: We agreed on renaming `check0` and `check1` to something more sensible — maybe `phys_eq_left` and `phys_eq_right`?

middle_end/flambda2/algorithms/patricia_tree.ml
```diff
@@ -941,6 +1101,16 @@ end = struct
     | Branch (prefix, bit, t0, t1) ->
       branch prefix bit (filter_map iv f t0) (filter_map iv f t1)
 
+  let combine_update_many iv f t0 t1 =
+    combine_union iv (fun k v0 v1 -> f k (Some v0) v1) t0 t1
+
+  let update_many f t0 t1 =
+    let iv = is_value_of t0 in
+    universal_merge
```
bclement-ocp, line 1109: Following offline discussion, we should add `check0` here for memory optimisation, and either document it as a guarantee or add `update_many_sharing` with that guarantee.

Ekdohibs:
Subsumed by #4262.

PR #4184, opened 2025-06-24
Allow non-singleton layouts for Lambda mutable variables

mshinwell:
Based on #3964 .  This doesn't update any of the tests.  Minimal testing thus far, only the following:
```
let triangle n =
  let mutable total = #(1, 2) in
  for i = 1 to n do
    let #(x, y) = total in
    total <- #(x + i, y + i + 3)
  done;
  total
```


mshinwell:
@ccasin @jra4 I mentioned internally there was some weird problem with `+`, but that was just a workflow problem, I needed to recompile the stdlib.

mshinwell:
@jra4 you should be able to push now

PR #4188, opened 2025-06-25
More precise notion of inlining being disabled

mshinwell:
An improved version of #4180 which should be more precise.


Ekdohibs (approved):

Gbury (approved):

chambart (approved):
This seems good

PR #4189, opened 2025-06-25
Remove caml_get_header_val, as not needed.

NickBarnes:
This arose during review of #3493 : `caml_get_header_val` only differs from `Hd_val` in that it will spin if it sees an in-progress header, i.e. if some domain is in the process of promoting the block. This can only happen between the barriers of a minor collection; there are very few places outside `minor_gc.c` where it might be possible, and certainly not while updating finalizer data structures after a minor GC.


stedolan (approved):

mshinwell:
@NickBarnes can we finish this and merge?

PR #4192, opened 2025-06-25
Add `[@expose]` annotation to trigger return value env extension on a function

chambart:
This clearly need a new name

```ocaml
let[@expose][@inline never] plop x = Some x
let v = match plop 42 with
   | None -> 1
   | Some x -> x * 2
```

Will compile to the expected `let v = let _ = plop 42 in 84`

### DIFF
--- a/lambda/translattribute.ml
+++ b/lambda/translattribute.ml
@@ -311,11 +311,9 @@ let add_expose_attribute expr loc attributes =
           | Default_expose -> ()
           | Always_expose | Never_expose ->
             Location.prerr_warning loc
-              (Warnings.Duplicated_attribute "inline")
+              (Warnings.Duplicated_attribute "expose")
         end;
         let attr = { attr with expose } in
-        check_local_inline loc attr;
-        check_poll_inline loc attr;
         lfunction_with_attr ~attr funct
     end
   | _ -> expr


lambda/translattribute.ml
```diff
@@ -280,6 +300,26 @@ let add_inline_attribute expr loc attributes =
...
+  | Lfunction({ attr = { stub = false } as attr } as funct) ->
+    begin match get_expose_attribute attributes with
+      | Default_expose -> expr
+      | (Always_expose | Never_expose)
+          as expose ->
+        begin match attr.expose with
+          | Default_expose -> ()
+          | Always_expose | Never_expose ->
+            Location.prerr_warning loc
+              (Warnings.Duplicated_attribute "inline")
```
lthls:
```suggestion
              (Warnings.Duplicated_attribute "expose")
```

lambda/translattribute.ml
```diff
@@ -280,6 +300,26 @@ let add_inline_attribute expr loc attributes =
...
+        end;
+        let attr = { attr with expose } in
+        check_local_inline loc attr;
+        check_poll_inline loc attr;
```
lthls:
```suggestion
```

PR #4195, opened 2025-06-25
Arrays accept any `separable` arguments, including non-float `or_null`s

dkalinichenko-js:
Remove the `any_non_null` layout. Instead, arrays now accept all `any mod separable` layouts, which included non-float `or_null`s. Add tests for `or_null` arrays.

Prevent `immediate_or_null` arrays from being optimized to `Intarray` in Flambda, since `Intarray`s expect their elements to be non-null tagged integers. This should be fixed in a subsequent PR.

Non-value layouts are still `Non_null` to avoid issues with type inference defaulting to non-null sort variables in some places. Really, we should merge the nullability axis into the layout, but I leave that for later.

This change carries some risk: we now allow `or_null` arrays, which can theoretically lead to miscompilations or runtime errors. I think this risk is small. Tests for `or_null` arrays succeed; we have also previously tested `or_null` in `Uniform_array.t`.

Best reviewed commit by commit.

### DIFF
--- a/typing/jkind.ml
+++ b/typing/jkind.ml
@@ -2636,7 +2636,7 @@ let for_float ident =
 
 let for_array_argument =
   let mod_bounds =
-    Mod_bounds.create ~locality:Locality.Const.max
+    Mod_bounds.create ~areality:Regionality.Const.max
       ~linearity:Linearity.Const.max ~portability:Portability.Const.max
       ~yielding:Yielding.Const.max ~uniqueness:Uniqueness.Const_op.max
       ~contention:Contention.Const_op.max ~statefulness:Statefulness.Const.max


dkalinichenko-js:
Yeah, that's right, support for the latter will be done in a future PR.

testsuite/tests/parsetree/source_jane_street.ml
```diff
@@ -927,7 +927,7 @@ let f x =
   | _ -> assert false;;
 
 [%%expect{|
-val f : 'a iarray -> 'a iarray = <fun>
+val f : ('a : value_or_null mod separable). 'a iarray -> 'a iarray = <fun>
```
rtjoa, line 930: It may be confusing for users unfamiliar with the kind system that `let f x = [| x |]` no longer typechecks as `'a -> 'a array` but as `('a : value_or_null mod separable). 'a -> 'a array`. This seems inevitable, and I can't think of a nice abbreviation for that kind, but I'm leaving this comment in case you have better ideas.

typing/jkind.ml
```diff
@@ -1501,8 +1479,8 @@ module Const = struct
         name = "immediate64"
       }
 
-    (* CR layouts v3: change to [Maybe_null] when
-       [or_null array]s are implemented. *)
+    (* CR or_null: nullability here should be [Maybe_null], but is set
+       to [Non_null] for now due to inference limitations. *)
```
rtjoa, line 1482: For my own understanding, could you elaborate a bit on what these inference limitations are?

typing/typeopt.ml
```diff
@@ -150,7 +150,10 @@ let classify ~classify_product env ty sort : _ classification =
   let ty = scrape_ty env ty in
   match (sort : Jkind.Sort.Const.t) with
   | Base Value -> begin
-  if is_always_gc_ignorable env ty then Int
+  (* CR or_null: [immediate_or_null] arrays are intarrays. *)
```
rtjoa: Why are they not currently? This is my guess:
```suggestion
  (* CR or_null: [immediate_or_null] arrays can be intarrays once that is
     supported by the middle-end *)
```

typing/typeopt.ml
```diff
@@ -259,7 +262,7 @@ let array_kind_of_elt ~elt_sort env loc ty =
     | Some s -> s
     | None ->
       Jkind.Sort.default_for_transl_and_get
-        (type_legacy_sort ~why:Array_element env loc ty)
+        (type_sort ~why:Array_element env loc ty)
```
rtjoa, line 266: `type_sort`'s kind bound is `Maybe_separable`; should this instead use `Separable` as it's an array element?

testsuite/tests/typing-layouts-products/product_arrays.ml
```diff
@@ -102,14 +102,14 @@ type t8a = t8 array
 (* Test 4: makearray_dynamic *)
```
rtjoa, line 102: Could you add some tests for the runtime behavior of product arrays containing null? Ideally:
1. A few basic handwritten tests
2. Another instance of the functorized product array tests, as in `testsuite/tests/typing-layouts-arrays/test_scannable_product_array_1.ml`

rtjoa (commented):
Excited for this change!

While experimenting on this branch, I noticed there seems to be an issue with constraining types to be `any mod separable`. I think we should first fix this in a separate PR, and make sure this PR's testsuite covers this case. 
```ocaml
external[@layout_poly] make_vect : ('a : any mod separable) . int -> 'a -> 'a array =
  "%makearray_dynamic"
external[@layout_poly] get : ('a : any mod separable) . 'a array -> int -> 'a =
  "%array_safe_get"
external[@layout_poly] set :
  ('a : any mod separable) . 'a array -> int -> 'a -> unit = "%array_safe_set"

let () =
  let a = make_vect 1 (This 0.) in
  set a 0 Null

(* with [ocamlopt]:
[ocaml] [flambda2] Invalid code:
(Defining_expr_of_let (bound_pattern Parraysets/25N)
 (defining_expr
  (((Array_set Values (Values Assign Heap)) a/20N 0 #null)
   ../testu.ml:10,2--14)))
*)
```

typing/typeopt.ml
dkalinichenko-js: Doesn't matter here because this function only gets the sort, and array elements are checked to be separable in other places.

testsuite/tests/parsetree/source_jane_street.ml
dkalinichenko-js: Yeah, I don't really have a better idea either.

typing/jkind.ml
dkalinichenko-js: Specifically, we want to default type variables and abstract types in signatures to `value`. Therefore, we must initialize them to a sort variable constrained to be in the same position as `value` on all axes. Therefore, other kinds like `bits64` also ought to start at that position or they would need extra annotations like `value_or_null` does (for the former `val f : ('a : bits64) -> unit` works but the latter requires `val g : ('a : value_or_null) . 'a -> unit`. 

I'm sure we could fix this or find a workaround, but that's how it currently is.

testsuite/tests/typing-layouts-products/product_arrays.ml
dkalinichenko-js: Added those tests.

rtjoa: Thanks!

typing/jkind.ml
rtjoa: Got it, thanks for the explanation.

testsuite/tests/peek_and_poke/peek_and_poke_forbidden_for_value.ml
```diff
@@ -23,13 +23,29 @@ let bad_read p : string = read p
 Line 1, characters 26-32:
 1 | let bad_read p : string = read p
                               ^^^^^^
-Error: Unsupported layout for the peek primitive
+Error: This expression has type "('a : value_or_null mod external_)"
```
rtjoa: Is it possible for this test to still exercise the original error (maybe by changing `string` to `int`?)

rtjoa (approved):

testsuite/tests/peek_and_poke/peek_and_poke_forbidden_for_value.ml
rtjoa: Actually, changing to `int` gives the following error on bytecode. I couldn't get it to give the original error in native either (in 5 minutes of trying) - if you find a way to get this to work, great, otherwise I think it's fine to leave a CR explaining what this used to test.
```
>> Fatal error: Blambda_of_lambda: (peek
tagged_immediate) is not supported in bytecode
Uncaught exception: Misc.Fatal_error
```

PR #4196, opened 2025-06-25
Allow `or_null` in `[@@unboxed]` types

dkalinichenko-js:
There isn't any reason to disallow it, we just didn't get to allowing it before. Add tests.

@rtjoa could you double-check that the `void` tests are fine?

### DIFF
--- a/testsuite/tests/typing-layouts-products/separability.ml
+++ b/testsuite/tests/typing-layouts-products/separability.ml
@@ -75,33 +75,33 @@ Error: This type cannot be unboxed because
        You should annotate it with "[@@ocaml.boxed]".
 |}]
 
-(* CR layouts v12: Once we allow products containing void in unboxed GADTs,
-   we'll have to make sure the below fails separability checking: *)
+(* #(value & void) and similar kinds are always considered separable,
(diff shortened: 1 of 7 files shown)


testsuite/tests/typing-layouts-products/separability.ml
```diff
@@ -81,42 +81,27 @@ type t_void : void
...
-3 | and bad = F : 'a r -> bad [@@unboxed]
-    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-Error: The kind of type "bad" is value_or_null & void
-         because it is an unboxed record.
-       But the kind of type "bad" must be a subkind of value & void
-         because it's an [@@unboxed] type,
-         chosen to have kind value & void.
+type t_void : void
+and 'a r = #{ a : 'a; v : t_void; }
+and bad = F : 'a r -> bad [@@unboxed]
```
rtjoa: These tests were written because we thought `#(float * void)` would be subject to the float array hack, but it's not, so it's actually okay to allow these types. Could you update the CR above, and rename the type from `bad` to reflect this?

rtjoa (approved):

PR #4198, opened 2025-06-25
Remove with-bounds for polymorphic variants

glittershark:
This appears to have caused a performance regression, so we're just reverting it until we can investigate further.


ccasin (approved):
The diff to ctype is correct and the test diff looks like the one I expect.  @glittershark please describe in a little more detail the relationship between this and #3845 in a comment or the PR description.

PR #4200, opened 2025-06-25
New design for portable exceptions

dkalinichenko-js:
Update https://github.com/oxcaml/oxcaml/pull/4062 with the new design for portable exceptions.
Instead of always requiring the argument types to cross portability and contention, on creation under a contention lock, arguments are required to be portable and cross contention. Similarly, on pattern-matching under a contention lock, arguments must cross portability and are marked as contended. Rebinding still requires full mode-crossing.

### DIFF
--- a/jane/doc/extensions/_05-modes/reference.md
+++ b/jane/doc/extensions/_05-modes/reference.md
@@ -47,3 +47,50 @@ can construct the lazy value at `many` even if the thunk is `once` (e.g., closin
 let r = { x = 0 } in
 let l @ many = lazy (overwrite_ r with { x = 42 })
 ```
+
+# Exceptions
+
+The exception type `exn` crosses portability and contention.
+
+For backwards compatibility with OCaml, we don't require exception constructor argument
+types to cross portability and contention themselves. Instead, we treat each instance
+of an exception constructor as belonging to the capsule it originally was defined in.
+
+When the constructor is instantiated outside the original capsule
+(i.e. in a `portable` function), its arguments are required to cross contention
+and be portable. Likewise, when pattern-matched on outside the original capsule,
+its arguments must cross portability and are marked as contended.
+
+```ocaml
+exception Foo of (unit -> unit)
+exception Bar of int ref
+
+
+let (foo @ portable) f =
+  raise (Foo f) (* Here, [f] is required to be portable and must cross contention. *)
+
+let (bar @ portable) g =
+  try g () with
+  | Bar x -> ... (* And here [x] is marked as contended and must cross portaibility. *)
+```
+
+Rebinding exception constructors "resets" its originating capsule.
+It's permitted only if all its argument types cross portability and contention:
+
+```ocaml
+exception Crossing of int list
+exception Noncrossing of (string -> unit)
+
+let (cross @ portable) () =
+    let module M = struct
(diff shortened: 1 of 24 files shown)


dkalinichenko-js:
The current design is unsound, to be fixed.

typing/mode_intf.mli
```diff
@@ -803,6 +816,12 @@ module type S = sig
     (** The mode crossing that crosses everything. *)
     val bot : t
 
+    (** The mode crossing that's weaker than both inputs *)
+    val join : t -> t -> t
+
+    (** The mode crossing that's stronger than both inputs *)
+    val meet : t -> t -> t
+
```
riaqn: You can say `include Lattice with type t := t` instead.

typing/mode_intf.mli
```diff
@@ -409,18 +409,30 @@ module type S = sig
     module Areality : Common
 
     module Monadic : sig
+      type 'a axis := (monadic, 'a) Axis.t
+
```
riaqn: I'm not sure about this - I personally prefer saying `Axis.t` over `axis` in the following signature, also I think this change is unrelated to the PR anyway.

typing/ctype.ml
```diff
@@ -7356,3 +7356,82 @@ let constrain_decl_jkind env decl jkind =
         match decl.type_manifest with
         | None -> err
         | Some ty -> constrain_type_jkind env ty jkind
+
+type 'res constructor_crossing_kind =
+  | Creation : Mode.Value.r constructor_crossing_kind
+  | Destruction : Mode.Value.l constructor_crossing_kind
+  | Rebinding : unit constructor_crossing_kind
+
+let check_exn_constructor_crossing (type res)
```
riaqn: looking at this function, the code sharing between `Creation` and `Destruction` is quite limited - maybe better to just have two functions. And `Rebinding` just calls both.

jane/doc/extensions/_05-modes/reference.md
```diff
@@ -47,3 +47,50 @@ can construct the lazy value at `many` even if the thunk is `once` (e.g., closin
...
 let l @ many = lazy (overwrite_ r with { x = 42 })
 ```
+
+# Exceptions
+
+The exception type `exn` crosses portability and contention.
+
+For backwards compatibility with OCaml, we don't require exception constructor argument
+types to cross portability and contention themselves. Instead, we treat each instance
+of an exception constructor as belonging to the capsule it originally was defined in.
```
riaqn, line 57: We should mention the current soundness bug: they can be packed in a first-class module and sent to other threads.

Also, while I appreciate the capsule analogy, I think we need to either be more precise (by at least pointing to `project` and `inject` and explain the correspondence), or just remove the analogy and state the concrete behavior.

typing/ctype.ml
dkalinichenko-js: Done.

jane/doc/extensions/_05-modes/reference.md
dkalinichenko-js: Done.

typing/mode_intf.mli
dkalinichenko-js: Changed to `Axis.t`.

PR #4201, opened 2025-06-26
Vec256 casts & constants

TheNumbat:
Adds constants, static casts, and reinterpret cast builtins for vec256 & vec512. The structure is the same as vec128.
Adds tests for all vec256 builtins; vec512 is still unimplemented.  (Most of the diff is tests)

### DIFF
--- a/oxcaml/tests/simd/gen/gen_dune.ml
+++ b/oxcaml/tests/simd/gen/gen_dune.ml
@@ -101,6 +101,7 @@ let make_ops_u name =
   (with-stdout-to
    ${name}_u.ml
    (progn
+   (echo "[@@@ocaml.warning \"-unused-module\"]")
    (echo "module Builtins = Builtins_u\n\n")
    (echo "module Utils = Utils_u\n\n")
    (cat "${name}.ml")))))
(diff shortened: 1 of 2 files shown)


oxcaml/tests/simd/gen/gen_dune.ml
```diff
@@ -170,7 +170,14 @@ let () =
       "callback256", enabled_if_main_amd64_not_macos;
       "test_callee_save_neon_regs", enabled_if_main;
       "probes", enabled_if_main_amd64_not_macos;
-      "probes256", enabled_if_main_amd64_not_macos ]
+      "probes256", enabled_if_main_amd64_not_macos;
+      "utils256", enabled_if_main_amd64_not_macos;
+      "ops_float32x8", enabled_if_main_amd64_not_macos;
```
gretay-js, line 186: do you want to test unboxed versions of these new operations too? it needs to be added manually above.

gretay-js (approved):

oxcaml/tests/simd/gen/gen_dune.ml
TheNumbat: ah, yeah added

PR #4206, opened 2025-06-26
Initial `CLAUDE.md`

dkalinichenko-js:
Create a `CLAUDE.md` file for Claude Code. It's pretty barebones, but it's still helpful for describing what commands the AI should run, and so seems worthwhile to share. I've been using this file for a few weeks now.


dkalinichenko-js:
There is no syntax :) You can write whatever you want, and it'll be added to the LLM prompt.

TheNumbat (approved):
Haven't actually tried using this file with Claude yet, but looks good for now. We can adjust as we use it more.

PR #4207, opened 2025-06-26
Revert "Improve the inclusion check around identity abbreviations"

glittershark:
Reverts oxcaml/oxcaml#4157

This appears to have ommitted modalities somehow, I'm seeing eg:
```
Signature mismatch:
...
Type declarations do not match:
  type 'a t__portable = { contents : 'a @@ portable; } [@@unboxed]
is not included in
  type 'a t__portable : value mod portable
The kind of the first is immediate with 'a
  because it is the expansion of a type abbreviation.
But the kind of the first must be a subkind of value mod portable
  because of the definition of t__portable at file "<file"
```


goldfirere:
I think this can be closed in favor of #4208. Do you agree @glittershark ?

glittershark:
yep

PR #4209, opened 2025-06-26
Use proper equality comparison to avoid duplicating work

goldfirere:
This fixes a terrible performance regression

### DIFF
--- a/typing/ctype.ml
+++ b/typing/ctype.ml
@@ -7198,7 +7198,12 @@ let print_global_state fmt global_state =
    can't be written in Jkind. It's possible that, after jkind.ml is broken up,
    this problem goes away, because the dependency from Env to Jkind is pretty
    minimal. *)
-let type_equal env ty1 ty2 = is_equal env false [ty1] [ty2]
+let type_equal env ty1 ty2 =
+  let snap = Btype.snapshot () in
+  let result = is_equal env false [ty1] [ty2] in
+  Btype.backtrack snap;
+  result
+
 let () = type_equal' := type_equal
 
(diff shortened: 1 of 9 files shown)


goldfirere:
> Are we sure that `Ctype.equal` doesn't call `Jkind.normalize` and we avoid looping?

Good question. But I think it's fine: `Ctype.equal` should call back into Jkind only for variables, which have r-kinds, which don't have with-bounds. (I checked the code on this point, but only briefly)

typing/ctype.ml
```diff
@@ -7196,7 +7196,12 @@ let print_global_state fmt global_state =
    can't be written in Jkind. It's possible that, after jkind.ml is broken up,
    this problem goes away, because the dependency from Env to Jkind is pretty
    minimal. *)
-let type_equal env ty1 ty2 = is_equal env false [ty1] [ty2]
+let type_equal env ty1 ty2 =
```
dkalinichenko-js, line 7201: I'm suspicious of this change, since it affects everything, not just jkind normalization, and so is potentially breaking. Perhaps we should only do this inside `Jkind.normalize`?

glittershark: agreed

goldfirere: No, I think that function is used only around jkind stuff. Look at the comment. I'm a little suspicious, too, but I think it should be fine.

goldfirere: I just confirmed that it's used only for jkind stuff.

goldfirere: Please Resolve if that satisfies you...

PR #4211, opened 2025-06-26
Void in blocks; move void to stable

rtjoa:
## Summary

Allow `void` in records and variants. Variant constructors whose arguments are all void or products of void are represented as tagged integers (as if they have no arguments). We store other variants and non-empty records containing void as mixed blocks.

This PR also moves void from alpha to stable. 

Explicitly not yet supported, to be added in future PRs:
- Arrays containing void.
- Records whose fields are all void/products of void (including inline records).
  * These records will have the same representation as `unit`.
- Extensible variants containing void
  * This is disallowed already because non-values are not supported in extensible variants.
- Void at the top level of modules
  * This is disallowed already because non-values are not supported at the top level of modules.
- Void in C bindings.

## Context
* **Void.** Void is represented as a nullary unboxed product in lambda and flambda2. See #4088.
* **Mixed blocks.** The compilation of mixed blocks is relevant. For details, see #3687, or in the source, `lambda/mixed_block_shape.mli` and the translation of `Pmixedfield` in `middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml`.

## Review

- @dkalinichenko-js to "whole-feature" review. Reviewing commit-by-commit is recommended. Note that much of the diff is due to test generation infrastructure from #4211. Note also that I have set [`patient = true`](https://github.com/oxcaml/oxcaml/blob/6b2bd51072e0eb38a59d3d2c0c3ca121e66b9590/testsuite/tests/records-and-block-indices/test_generation.ml#L20-L23) (generates extra tests) for now so those tests can run in CI, but I will set it to `false` before merging.
- @xclerc to review changes to `lambda/`.

## Stack
Depends on:
- https://github.com/oxcaml/oxcaml/pull/4205


testsuite/tests/typing-layouts-void/basics_void.ml
```diff
@@ -0,0 +1,212 @@
+(* TEST
+ expect;
+*)
+
+type unit_u : void
+
+(* Variants whose constructor arguments are all void are immediates *)
+
+type v : immediate = A of unit_u
```
dkalinichenko-js, line 12: This seems wrong; the jkind of such a type should be `immediate with unit_u`. Otherwise, we can store types like `Capsule.Key.t` in such a wrapper and get them to cross mode axes they don't cross natively.

lambda/matching.ml
```diff
@@ -1926,30 +1903,61 @@ let get_expr_args_constr ~scopes head (arg, _mut, sort, layout) rem =
...
-              ~get_mode:(fun _i ->
-                Misc.fatal_error
-                  "unexpected flat float of layout value in \
-                    constructor field")
-              shape
-          in
-          Pmixedfield ([pos], shape, sem)
+  let make_void_access binding_kind sort pos =
+    (* Constructors whose arguments are all void, e.g.
+       [A of #(void * void) * void], are immediate. Accesses to their arguments
```
dkalinichenko-js, line 1910: Why isn't emitting `Pmixedfield` sufficient here? I'd expect `Pmixedfield` on a `Void` field to be translated into a no-op later, and it's surprising if that isn't the case.

testsuite/tests/typing-layouts-void/basics_void.ml
rtjoa: Thanks for catching, fixed in e4d90880dd876d7427d66f2eb43529e29cb45f9b (note most of the GitHub-rendered diff here is just an indentation change.)

testsuite/tests/typing-layouts-void/basics_void.ml
```diff
@@ -2,13 +2,15 @@
  expect;
 *)
 
-type unit_u : void
+type unit_u : void mod everything
+[%%expect{|
+type unit_u : void mod global aliased many stateless immutable external_
```
dkalinichenko-js: I'm mildly surprised this isn't printed as `void mod everything`, I think it should be? Feel free to turn this into a CR.

testsuite/tests/typing-layouts-void/basics_void.ml
```diff
@@ -28,6 +30,34 @@ Error: The kind of type "bad" is immutable_data with unit_u
          because of the annotation on the declaration of the type bad.
 |}]
 
+(* With-bounds for all-void variants *)
```
dkalinichenko-js, line 33: Would be good to add a test that a variant with a `void mod global` field fits into `value mod global`.

typing/jkind.ml
```diff
@@ -2554,115 +2554,115 @@ let for_abbreviation ~type_jkind_purely ~modality ty =
...
-        (fun cstr ->
-          match cstr.cd_args with
-          | Cstr_tuple _ -> false
-          | Cstr_record lbls -> has_mutable_label lbls)
-        cstrs
-    in
-    let base =
+  let base =
+    if List.for_all
+         (* CR layouts v12: This code assumes that all voids mode-cross. I
```
dkalinichenko-js: Does it assume that now? I think we can remove this CR.

lambda/translcore.ml
```diff
@@ -525,10 +506,13 @@ and transl_exp0 ~in_new_scope ~scopes sort e =
...
       | Ordinary {runtime_tag}, _ when cstr.cstr_constant ->
-          assert (args_with_sorts = []);
-          (* CR layouts v5: This could have void args, but for now we've ruled
-             that out by checking that the sort list is empty *)
-          Lconst(const_int runtime_tag)
+          assert (
+            List.for_all
+              (fun (_, s) -> Jkind.Sort.Const.all_void s) args_with_sorts);
+          List.fold_left
+            (fun (acc : lambda) (e : lambda) -> Lsequence (e, acc))
```
dkalinichenko-js, line 513: Why emit `Lsequence` here?

typing/typedecl.ml
```diff
@@ -1787,7 +1782,10 @@ let rec update_decl_jkind env dpath decl =
   in
   let decl = { decl with type_unboxed_version } in
   let open struct
-    (* For tracking what types appear in record blocks. *)
+    (* For tracking what types appear in record blocks. All product layouts
+       count only as a [non_float64_unboxed_field], even if its a
+       [float64 & float64] or [void & void].
```
dkalinichenko-js, line 1802: Or, presumably, `float64 & void`?

typing/typedecl.ml
```diff
@@ -1838,27 +1838,27 @@ let rec update_decl_jkind env dpath decl =
                              | Vec128 | Vec256 | Vec512 | Word | Product _ ) ->
                repr_summary.non_float64_unboxed_fields <- true
            | Value_element -> repr_summary.values <- true
-           | Element_without_runtime_component _ -> ())
+           | Void ->
+               repr_summary.voids <- true)
         reprs;
       let rep =
```
dkalinichenko-js, line 1859: Aside: you don't have to fix this right now, but I find this pattern-match very confusing without warning 9/`missing-record-field-pattern`.

typing/typedecl.ml
```diff
@@ -1838,27 +1838,27 @@ let rec update_decl_jkind env dpath decl =
...
         match repr_summary with
         (* We store mixed float/float64 records as flat if there are no
```
dkalinichenko-js: Should it be no non-float and non-void fields now?

typing/typedecl.ml
```diff
@@ -1880,19 +1883,19 @@ let rec update_decl_jkind env dpath decl =
...
+            non_float64_unboxed_fields = false; voids = false } ->
           Record_float
         | { values = false; floats = false; float64s = true;
-            non_float64_unboxed_fields = false } ->
+            non_float64_unboxed_fields = false; voids = false } ->
           Record_ufloat
         | { values = false; floats = false; float64s = false;
-            non_float64_unboxed_fields = false }
-          [@warning "+9"] ->
+            non_float64_unboxed_fields = false; voids = _ }  ->
```
dkalinichenko-js, line 1914: Could you leave a CR about all-void records here?

testsuite/tests/typing-layouts-void/basics_void.ml
```diff
@@ -1,5 +1,5 @@
 (* TEST
- flags = "-extension layouts_alpha";
+ flags = "-extension layouts_beta";
```
dkalinichenko-js: No flag needed?

typing/typeopt.ml
```diff
@@ -1219,6 +1218,9 @@ let report_error ppf = function
       fprintf ppf
         "Unboxed vector types are not yet supported in arrays of unboxed@ \
          products."
+  | Unsupported_void_in_array ->
+      fprintf ppf
+        "Types whose layout contains [void] are yet supported in arrays."
```
dkalinichenko-js: Are not yet supported?

testsuite/tests/typing-layouts-void/behavior_test_based_on_old_basics_void.ml
```diff
@@ -0,0 +1,353 @@
+(* TEST
+ reference = "${test_source_directory}/behavior_test_based_on_old_basics_void.reference";
+ include stdlib_upstream_compatible;
+ flambda2;
+ {
+   flags = "-extension layouts_alpha";
```
dkalinichenko-js: No need for `-extension layouts_alpha`?

testsuite/tests/typing-layouts-void/behavior_test_based_on_old_basics_void.ml
```diff
@@ -0,0 +1,353 @@
...
+let b' = id1' b'
+
+let _ =
+  assert (List.for_all2 (=) !r [12;11;10;9;8;7;6;5;4;3;2;1]);
+  assert (b'.x = 3 && b'.z = 42)
+;;
+
+(***************************************************)
+(* Test 2: evaluation order of variants with voids *)
+type void_variant =
```
dkalinichenko-js, line 96: Would be good to add a similar test for an all-void variant.

testsuite/tests/typing-layouts-void/basics_void.ml
dkalinichenko-js: And also a test with two different constructors and with-bounds.

testsuite/tests/typing-layouts-void/voids.ml
```diff
@@ -0,0 +1,595 @@
+(* TEST
+ reference = "${test_source_directory}/voids.reference";
+ include stdlib_stable;
+ flambda2;
+ {
+   native;
+ }{
+   bytecode;
+ }{
+   flags = "-extension layouts_alpha";
```
dkalinichenko-js: O3 and Oclassic?

testsuite/tests/typing-layouts-void/voids.ml
```diff
@@ -0,0 +1,595 @@
...
+
+let _ = test10 ()
+
+(****************************************)
+(* Test 11: void in recursive functions *)
+
+let test11 () =
+  start_test "void in recursive functions";
+
+  let[@inline never] rec countdown n v acc =
```
dkalinichenko-js, line 407: I think this one is tail-recursive, we should test both tail and non-tail recursion.

testsuite/tests/typing-layouts-void/voids.ml
dkalinichenko-js, line 1: Great testsuite, thanks! This is really nice.

Still, I'm going to ask you to test some more things, since this PR is going to move `void` into `stable`:
1. Runtime conversions between blocks with void and non-void fields and same blocks with all void fields omitted, using `%opaque`.
2. As above, but for immediates.
3. Equality, comparison, hashing, and marshalling for blocks and immediates containing voids. If some of those e.g. don't work in bytecode, we should record that.
4. Higher-order functions with `void` arguments or returns.
5. Object methods with `void` arguments or returns.
6. Local continations (`let[@local]`) with `void` arguments or returns.
7. Loopification (`let[@loop]`) with `void` arguments or returns.
8. Immediates with void fields (use `void mod external_` here) are `value mod external_` and thus are optimized correspondingly.
9. Local closures (`@ local`) with void arguments or returns.
10. Stack-allocating blocks containing voids.

You might want to split some of those into separate files.

dkalinichenko-js:
Is `void` in C bindings still unsupported?

testsuite/tests/typing-layouts-void/voids.ml
dkalinichenko-js: Another one: `Obj.tag`, `Obj.field`, etc on records and variants with void fields.

lambda/translcore.ml
mshinwell: Is this a question about whether `Lsequence` permits void-returning things rather than just unit?  We should discuss

dkalinichenko-js: Yeah.

lambda/matching.ml
rtjoa: This is necessary for correctness in bytecode, where `Pmixedfield`s that access `void` are not erased but translated into field access(es) (as unboxed products are boxed in bytecode). I updated the comment to state this.

testsuite/tests/typing-layouts-void/basics_void.ml
rtjoa: Fixed - this is now printed as `void mod everything`.

testsuite/tests/typing-layouts-void/basics_void.ml
rtjoa: Added in efc2a7c694da9b13c7420dad406bc7715b49030e.

typing/jkind.ml
rtjoa: Right, removed

typing/typedecl.ml
rtjoa: Yeah - I highlighted those two layouts because I think they are particularly ambiguous (one might think they count for `float64s` and `voids`, respectively).

typing/typedecl.ml
rtjoa: Left a CR to improve this.

typing/typedecl.ml
rtjoa: Right, updated this comment to simply say `We store floats flatly in mixed records if all fields are float/float64/void.`

typing/typedecl.ml
rtjoa: Added.

typing/typeopt.ml
rtjoa: Thanks, fixed.

testsuite/tests/typing-layouts-void/behavior_test_based_on_old_basics_void.ml
rtjoa: Removed, thanks.

testsuite/tests/typing-layouts-void/voids.ml
rtjoa: Added, thanks.

testsuite/tests/typing-layouts-void/voids.ml
rtjoa: Updated in f23b14da6977ba29cbc92b0cfaee1f1c0509c106.

testsuite/tests/typing-layouts-void/behavior_test_based_on_old_basics_void.ml
rtjoa: Added in efc2a7c694da9b13c7420dad406bc7715b49030e (I think this is already mostly covered by the test for `C`, whose arguments are all void, but I added a test for a variant whose constructors are all all-void).

rtjoa:
Thanks for the review @dkalinichenko-js. I still have to add the list of extra tests you enumerated, but have addressed other comments. As for your questions:

> Is `void` in C bindings still unsupported?

That's right - updated the PR description to state this.

> How is `void` going to be represented in JavaScript and WASM?

I expect this to just work because we made no changes to the bytecode IR or runtime, but I'll test this before releasing.

> What's the public release erasure story for void? I see a potential problem here. Some types (like variants with only void arguments) are now `[@@immediate]` with `void` and non-immediate when layout annotations are erased. Should we warn of that in upstream-compatible code? This may be too complicated.
> 
> We do have to allow `void` in public release, since `Capsule` is going to use `void`. At least, it would be good to have tests for what happens to `void` with `-extension-universe upstream_compatible`.

Great point. I think a warning seems the most reasonable here, but let's chat about this.

lambda/translcore.ml
mshinwell: I checked and it's fine for `Lsequence` to be used with expressions of any layout, so this use should be ok.  Please add a comment in lambda.mli to this effect.

testsuite/tests/typing-layouts-void/voids.ml
rtjoa: Thanks for the suggestions!

I added all of these (with the help of Claude), modulo these exceptions:
- I didn't add runtime conversion tests using `%opaque`: We don't support runtime conversions between blocks with and without void, as these have different representations on bytecode. Even on native, adding void can make something a mixed block when it wasn't before.
- Note that structural equality, hashing, and marshalling don't work for blocks containing void, as those are mixed blocks (this is reflected in the tests).
- I did not add tests of `Obj.field`, as it will differ on bytecode (and I assert there isn't much interesting to test here beyond what we already have, as void causes "normal" mixed blocks).

rtjoa:
@dkalinichenko-js I added an upstream compatibility warning, as well as more tests. Also see `Jkind.for_boxed_variant`, as it had a non-trivial rebase conflict with the GADT with-kind rollback.

testsuite/tests/typing-layouts-void/voids.ml
dkalinichenko-js: Thanks, makes sense.

PR #4212, opened 2025-06-27
fix(types): Do not call `TE.find` on imported variables in n-way join

bclement-ocp:
The new join algorithm can fail with a "missing cmx and kind" error when creating a new existential variable that is equal in at least one of the branches to an imported variable for which we do not have type information.

This patch threads enough kind information to the places where this can occur, making sure that we only call `TE.find` with an unknown kind on variables that have been demoted and hence cannot have been imported from another compilation unit.

### DIFF
--- a/middle_end/flambda2/types/env/join_env.ml
+++ b/middle_end/flambda2/types/env/join_env.ml
@@ -92,17 +92,39 @@ let get_nth_joined_env index joined_envs =
    the target environment if they have been demoted. *)
 
 module Thing_in_env (Thing : Container_types.S) () : sig
-  include Container_types.S with type t = private Thing.t
+  include
+    Container_types.S
+      with type t = private Thing.t
+       and type Set.t = private Thing.Set.t
+       and type +!'a Map.t = private 'a Thing.Map.t
 
   val create : Thing.t -> t
+
+  val create_set : Thing.Set.t -> Set.t
+
+  val create_map : 'a Thing.Map.t -> 'a Map.t
 end = struct
   include Thing
 
   let create thing = thing
+
+  let create_set s = s
+
+  let create_map m = m
+end
+
+module Variable_in_target_env = struct
+  include Thing_in_env (Variable) ()
 end
 
 module Name_in_target_env = struct
   include Thing_in_env (Name) ()
+
+  let var (var : Variable_in_target_env.t) : t =
+    create (Name.var (var :> Variable.t))
+
+  let var_map (type a) (m : a Variable_in_target_env.Map.t) : a Map.t =
+    create_map (Name.var_map (m :> a Variable.Map.t))
 end
 
 module Simple_in_target_env : sig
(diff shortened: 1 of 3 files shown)


middle_end/flambda2/types/env/join_env.ml
```diff
@@ -1090,8 +1112,14 @@ let n_way_join_levels ~n_way_join_type t all_levels : _ Or_bottom.t =
           let canonicals =
             Simples_in_joined_envs.in_envs all_levels canonicals
           in
+          (* Passing [None] to [TE.find] here is OK, because [name] has been
+             demoted in at least one environment thus cannot be imported. *)
```
lthls, line 1116: This will not hold if we start simplifying physical equality:
```ocaml
(* a.ml *)
let var1 = Sys.opaque_identity 0
let var2 = Sys.opaque_identity 0

(* b.ml, compiled with a.cmx available *)
let symbol = (var1, var2)

(* c.ml, compiled with b.cmx available but a.cmx unavailable *)
let _ =
  let x, y = B.symbol in
  if x == y then ...
```

bclement-ocp: I believe even if we simplified physical equality this is OK: we could never demote neither `var1` nor `var2`, because we would raise a [`Binding_time_resolver_failure`](https://github.com/oxcaml/oxcaml/blob/6f2a91b78af6ccd6b2cb64169ea898b8a49aa77e/middle_end/flambda2/types/env/typing_env.ml#L339) when trying to figure which of `var1` and `var2` would be canonical, and would [drop the equality](https://github.com/oxcaml/oxcaml/blob/6f2a91b78af6ccd6b2cb64169ea898b8a49aa77e/middle_end/flambda2/types/env/typing_env.ml#L891-L895) instead.

bclement-ocp: But I agree the comment is misleading — what matters is not that the variable "cannot be imported" but that we know its kind information (because we currently can't demote a variable without knowing its kind information). I will fix the comment.

bclement-ocp: Ah actually I think that's not strictly true either — I think we can demote a variable to a symbol even without knowing the kind of the variable if we do know the kind of the symbol.

lthls: We can actually demote two variables coming from different units, even if one (or both) of the units is missing.
But this is all speculation anyway (we don't generate arbitrary demotions at the moment), so I'll resolve this conversation.

lthls (approved):
Looks ok for a temporary fix. Hopefully we can solve this for good by storing kinds on variables, before we hit the corner cases where kinds are missing.

middle_end/flambda2/types/env/join_env.ml
bclement-ocp: Ah, that's fair, good point. 

I was thinking that the case that is not speculative would be demotion to a constant (which we can trigger through a switch), which we are dangerously close to being able to trigger as it could happen if:

 - We had an unboxed int type, or
 - We were a bit more aggressive with recovering `y : (= 0)` from `(y : (Variant (naked_immediate (= x))` and `(x : (= #0))`, or
 - We propagated joined equations on missing variables (both old and new join currently drop them)

PR #4214, opened 2025-06-27
fix(flambda2): Deterministic order for lifted symbols

bclement-ocp:
Ensure that the order in which lifted constants are defined does not depend on the `Code_id_or_symbol` ordering by using an `Lmap` and sorting the connected components.

This helps making comparisons between different runs (e.g. with different options) using the `flambda2-compare` library.

### DIFF
--- a/middle_end/flambda2/tests/mlexamples/unroll4.flt
+++ b/middle_end/flambda2/tests/mlexamples/unroll4.flt
@@ -81,28 +81,12 @@ in
 let code odd_2 deleted and code even_1 deleted in
 let code parity_is_0 deleted in
 let code k_3 deleted in
-let code rec loopify(never) size(26) newer_version_of(even_1)
+let code rec loopify(never) size(26) newer_version_of(odd_2)
       even_1_1 (n : imm tagged)
         my_closure my_region my_ghost_region my_depth
         -> k * k1 =
   let k = %project_value_slot even.k my_closure in
-  let odd = %project_function_slot (even -> odd) my_closure in
-  let prim = %phys_eq (n, 0) in
-  switch prim
-    | 0 -> k2
-    | 1 -> k3
-    where k3 =
-      apply inlined(hint) k (1) -> k * k1
-    where k2 =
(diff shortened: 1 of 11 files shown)


Ekdohibs (approved):

utils/strongly_connected_components.mli
```diff
@@ -73,10 +73,18 @@ module type S = sig
...
+  type numbered_graph
+
   type component =
     | Has_loop of Id.t list
     | No_loop of Id.t
 
+  val stable_number : (Id.t * Id.Set.t) list -> numbered_graph
+
+  val numbered_connected_components_sorted_from_roots_to_leaf
+     : numbered_graph
```
Ekdohibs: Maybe these two functions could be exposed as a single function doing the composition?

bclement-ocp: I was initially thinking of exposing only `numbered_xxx` but it requires more code changes; this suggestion looks like a good intermediate solution.

bclement-ocp: Done.

middle_end/flambda2/tests/mlexamples/unroll4.flt
```diff
@@ -81,38 +81,38 @@ in
 let code odd_2 deleted and code even_1 deleted in
 let code parity_is_0 deleted in
 let code k_3 deleted in
-let code rec loopify(never) size(26) newer_version_of(even_1)
+let code rec loopify(never) size(26) newer_version_of(odd_2)
       even_1_1 (n : imm tagged)
```
bclement-ocp: Hmm this looks suspicious…

bclement-ocp: Some order changed and things got garbled when applying the substitution in compare. I regenerated `unroll4.flt` from the actual generated code, which incidentally also fixed the `unroll4.flt: FAIL` output in the tests.

Ekdohibs (approved):

PR #4216, opened 2025-06-27
Mode cheatsheet tables

cuihtlauac:
Not meant to be included as-is

### DIFF
--- a/jane/doc/extensions/_05-modes/cheatsheet.md
+++ b/jane/doc/extensions/_05-modes/cheatsheet.md
@@ -1,0 +1,121 @@
+---
+layout: documentation-page
+collectionName: Modes
+title: Cheat Sheet
+---
+
+<style>
+.table {
+    width: fit-content;
+    margin-left: auto;
+    margin-right: auto;
+    margin-bottom: 20px;
+    border-style: solid;
+    border-color: blue;
+    border-radius: 25px;
+    padding: 15px;
+    text-align: center;
+}
+</style>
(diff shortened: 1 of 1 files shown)


goldfirere:
What kind of feedback are you looking for here? The PR says that it's not to be included as-is... but then what parts are WIP? As is, I'm not sure what action to take here.

cuihtlauac:
I believe having a single-page view of all the modes is helpful. It highlights resemblances and differences. It also builds a structured big picture of the modes landscape. This PR is meant to share my take on this matter and hopefully start a discussion on what could be added, whether it should be included in OxCaml docs and where it could go.

cuihtlauac:
> I think I prefer bold over italic for the legacy mode. (And there will need to be a little key somewhere saying that's what the bold means.)

Ok.

> I find the rotated text hard to read. Maybe use a diagonal if we want the space?

Ok.


> I do not understand the table at the bottom.

The idea is to express whether a definition at mode Y can be captured in a closure at mode X. The correspondence is line-wise, 

many can capture aliased (first line in X and Y cells)
observing can capture read (third line in X and Y cells)
nonportable can capture uncontended (second line in X and Y cells)

Since the topology is the same for aliasing, threads and purity, it's actually three tables in one. I agree it packs a lot of information.

I prefer drawing arrows in the main table. Each arrow points from the closure's mode to the lowest mode it can capture. But that's a bit heavy on plain HTML.

image

PR #4222, opened 2025-06-27
Dump detailed information about .cms files in `ocamlobjinfo`

lukemaurer:
Currently there's a bunch of data in common between .cmt files and .cms files but `ocamlobjinfo` doesn't include it for .cms files. This adds that data, and also exercises it in a few tests for parameterised libraries.

**Note for reviewing:** You may want to look at the diff offline with `git diff --color-moved` as a fair amount of code has been factored out.


mshinwell:
@riaqn would you be able to review this, should be straightforward?

riaqn:
@lukemaurer can you fix the CI?

PR #4224, opened 2025-06-27
local module type substitution can fail (backport ocaml/ocaml#13540)

rtjoa:
Backports https://github.com/ocaml/ocaml/pull/13540.

### Stack
1. https://github.com/oxcaml/oxcaml/pull/4224
2. https://github.com/oxcaml/oxcaml/pull/4240
3. https://github.com/oxcaml/oxcaml/pull/4241
4. https://github.com/oxcaml/oxcaml/pull/4246

### DIFF
--- a/typing/subst.ml
+++ b/typing/subst.ml
@@ -40,7 +40,7 @@ type additional_action =
   | Duplicate_variables
   | No_action
 
-type t =
+type s =
   { types: type_replacement Path.Map.t;
     modules: Path.t Path.Map.t;
     modtypes: module_type Path.Map.t;
@@ -48,8 +48,15 @@ type t =
     additional_action: additional_action;
 
     loc: Location.t option;
-    mutable last_compose: (t * t) option  (* Memoized composition *)
+    mutable last_compose: (s * s) option  (* Memoized composition *)
   }
+
+type 'a subst = s
+type safe = [`Safe]
+type unsafe = [`Unsafe]
+type t = safe subst
+exception Module_type_path_substituted_away of Path.t * Types.module_type
+
 let identity =
   { types = Path.Map.empty;
     modules = Path.Map.empty;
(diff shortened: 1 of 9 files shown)


typing/subst.ml
```diff
@@ -84,26 +91,23 @@ let add_type_replacement types id replacement =
...
     | _ -> types
 
+let unsafe x = x
+
 let add_type_path id p s =
   let types = add_type_replacement s.types id (Path p) in
   { s with types; last_compose = None }
 
 let add_type id p s =
   add_type_path (Pident id) p s
```
ccasin: I think I would copy upstream by (if I understand correctly) inlining `add_type_path` into its only caller, here.  Our `add_type_path` is a little bigger and a little different, but this will still make the diff a little smaller and clearer.

typing/subst.ml
```diff
@@ -84,26 +91,23 @@ let add_type_replacement types id replacement =
...
 
-let add_type_function id ~params ~body s =
-  let types =
-    add_type_replacement s.types id (Type_function { params; body })
-  in
-  { s with types; last_compose = None }
-
 let add_module_path id p s =
   { s with modules = Path.Map.add id p s.modules; last_compose = None }
```
ccasin: and similarly here

typing/subst.ml
```diff
@@ -1062,6 +1067,33 @@ let modtype_declaration sc s decl =
 let module_declaration scoping s decl =
   Lazy.(decl |> of_module_decl |> module_decl scoping s |> force_module_decl)
 
+module Unsafe = struct
```
ccasin, line 1066: I think the functions in this module need to update `last_compose` - setting it to `None` or (in the case of `compose`) to the last composition.

typing/typemod.ml
```diff
@@ -943,7 +928,7 @@ let merge_constraint initial_env loc sg lid constr =
             by the caller. So what we do with the scope doesn't really matter. But
             making it local makes it unlikely that we will ever use the result of
             this function unfreshened without issue. *)
```
ccasin: I think you should delete this comment.  It explains the reason for `Make_local` on the next line, which no longer appears (and has also been deleted upstream, though it was in a slightly different place there because of our "lazy strengthening" rework).

ccasin (changes requested):

ccasin (approved):
Thanks!

PR #4228, opened 2025-06-29
Make setting and getting atomic fields use atomic prims 

glittershark:
Based on https://github.com/oxcaml/oxcaml/pull/4132

When transl'ing Texp_setfield and Texp_field to lambda, use Patomic_set_field and Patomic_load_field if the fields are atomic.
Note this _only_ changes the runtime behavior - changing typing to have relaxed modes will (and can) come later

### DIFF
--- a/lambda/translcore.ml
+++ b/lambda/translcore.ml
@@ -650,14 +650,13 @@ and transl_exp0 ~in_new_scope ~scopes sort e =
           Record_boxed _
         | Record_inlined (_, Constructor_uniform_value, Variant_boxed _) ->
           let immediate_or_pointer, _ = maybe_pointer e in
-          begin match Types.atomic lbl.lbl_mut with
-          | Atomic ->
+          if Types.is_atomic lbl.lbl_mut
+          then
             Some
               (Patomic_load_field { immediate_or_pointer },
                [targ; Lconst (Const_base (Const_int lbl.lbl_pos))])
-          | Nonatomic ->
+          else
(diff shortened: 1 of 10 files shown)


TheNumbat (commented):
LGTM but I think a frontend person should also look over translcore

glittershark:
> LGTM but I think a frontend person should also look over translcore

I nominate @goldfirere

glittershark:
@rtjoa could probably also review

glittershark:
> Well then who reviewed the primitives? 
@TheNumbat did

glittershark:
Alright, the diff here now has what I expect.

dkalinichenko-js (changes requested):
`Translcore` looks good.

I'd like some more tests, particularly:
1. Tests for optimized compilation like `-O3`.
2. Tests that the runtime behavior is still right when floats are involved.
3. Tests involving `or_null` atomic fields.

Also would be good to document why we need no frame pointers on the cmm test.

dkalinichenko-js (approved):
I think we should also have runtime tests with native and optimizations (instead of just `expect` which is bytecode). But it's somewhat covered by other tests.

glittershark:
> Something like `typing-layouts-or-null/optimized.ml`.

Done.

glittershark:
Force-pushed to remove merge of https://github.com/oxcaml/oxcaml/pull/4364, but otherwise the diff is unchanged.

dkalinichenko-js (approved):

PR #4229, opened 2025-06-29
Add support for a new [%atomic.loc r.field] construct

glittershark:
Add support for a new [%atomic.lock r.field] contstruct, which gets typed into a
new Texp_atomic_loc typedtree expression node, which in turn turns into a block
allocation containing the record `r`, and the field index of the field `field`
in that record, with type `'a atomic_loc`, where `'a'` is the type of that
field:

```
Γ ⊢ e₁ : τ₁ @ m,
     τ₁ = { ... f : τ₂ ... }
——————————————————————————————————————————– ATOMIC_LOC
Γ ⊢ [%atomic.loc e₁.f] : τ₂ atomic_loc @ m
```

### DIFF
--- a/lambda/translcore.ml
+++ b/lambda/translcore.ml
@@ -643,6 +643,19 @@ and transl_exp0 ~in_new_scope ~scopes sort e =
         {fields; representation; extended_expression } ->
       transl_record_unboxed_product ~scopes e.exp_loc e.exp_env
         fields representation extended_expression
+  | Texp_atomic_loc (arg, arg_sort, id, lbl, alloc_mode) ->
+      let shape =
+        Some
+          [ Typeopt.value_kind arg.exp_env arg.exp_loc arg.exp_type;
+            { raw_kind = Pintval; nullable = Non_nullable }
+          ]
+      in
+      let arg_sort = Jkind.Sort.default_for_transl_and_get arg_sort in
+      check_record_field_sort id.loc lbl.lbl_sort;
+      let (arg, lbl) = transl_atomic_loc ~scopes arg arg_sort lbl in
+      let loc = of_location ~scopes e.exp_loc in
+      Lprim (Pmakeblock (0, Immutable, shape, transl_alloc_mode alloc_mode),
+             [arg; lbl], loc)
   | Texp_field(arg, arg_sort, id, lbl, float, ubr) ->
       let arg_sort = Jkind.Sort.default_for_transl_and_get arg_sort in
       let targ = transl_exp ~scopes arg_sort arg in
@@ -2210,6 +2223,11 @@ and transl_record ~scopes loc env mode fields repres opt_init_expr =
         Llet(Strict, Lambda.layout_block, init_id, init_id_duid,
              transl_exp ~scopes init_expr_sort init_expr, lam)
     end
+
+and transl_atomic_loc ~scopes arg arg_sort lbl =
+  let arg = transl_exp ~scopes arg_sort arg in
+  let lbl = Lconst (Const_base (Const_int (lbl.lbl_pos))) in
+  (arg, lbl)
 
 and transl_record_unboxed_product ~scopes loc env fields repres opt_init_expr =
   match repres with
--- a/testsuite/tests/typing-modes/atomic_loc.ml
+++ b/testsuite/tests/typing-modes/atomic_loc.ml
@@ -1,0 +1,30 @@
+(* TEST
+   expect;
+   flags += "-no-mutable-implied-modalities";
+*)
+
+type 'a atomic = { mutable contents : 'a [@atomic] }
+[%%expect{|
+type 'a atomic = { mutable(<non-legacy>) contents : 'a; }
+|}]
+
+let atomic_loc_contended (t @ contended) = [%atomic.loc t.contents]
+[%%expect{|
+val atomic_loc_contended : 'a atomic @ contended -> 'a atomic_loc @ contended =
+  <fun>
+|}]
+
+let atomic_loc_portable (t @ portable) : _ @ portable = [%atomic.loc t.contents]
+[%%expect{|
+val atomic_loc_portable : 'a atomic @ portable -> 'a atomic_loc @ portable =
+  <fun>
+|}]
+
+let uses_unique (t @ unique) : _ @ unique =
+  [%atomic.loc t.contents], [%atomic.loc t.contents]
+[%%expect{|
+Line 2, characters 2-26:
+2 |   [%atomic.loc t.contents], [%atomic.loc t.contents]
+      ^^^^^^^^^^^^^^^^^^^^^^^^
+Error: This value is "aliased" but expected to be "unique".
+|}]
--- a/typing/typecore.ml
+++ b/typing/typecore.ml
@@ -244,6 +244,8 @@ type error =
   | Probe_name_undefined of string
   | Probe_is_enabled_format
   | Extension_not_enabled : _ Language_extension.t -> error
+  | Invalid_atomic_loc_payload
+  | Label_not_atomic of Longident.t
   | Literal_overflow of string
   | Unknown_literal of string * char
   | Float32_literal of string
@@ -4348,6 +4350,7 @@ let rec is_nonexpansive exp =
            | Kept _ -> true)
         fields
       && is_nonexpansive_opt (Option.map fst extended_expression)
+  | Texp_atomic_loc(exp, _, _, _, _) -> is_nonexpansive exp
   | Texp_field(exp, _, _, _, _, _) -> is_nonexpansive exp
   | Texp_unboxed_field(exp, _, _, _, _) -> is_nonexpansive exp
   | Texp_ifthenelse(_cond, ifso, ifnot) ->
@@ -4849,6 +4852,7 @@ let check_partial_application ~statement exp =
             | Texp_ident _ | Texp_constant _ | Texp_tuple _
             | Texp_unboxed_tuple _
             | Texp_construct _ | Texp_variant _ | Texp_record _
+            | Texp_atomic_loc _
             | Texp_record_unboxed_product _ | Texp_unboxed_field _
             | Texp_overwrite _ | Texp_hole _
             | Texp_field _ | Texp_setfield _ | Texp_array _
@@ -7015,6 +7019,41 @@ and type_expect_
     end
   | Pexp_extension ({ txt = "src_pos"; _ }, _) ->
       rue (src_pos loc sexp.pexp_attributes env)
+  | Pexp_extension ({ txt = ("ocaml.atomic.loc"
+                             |"atomic.loc"); _ },
+                    payload) ->
+      begin match payload with
+      | PStr [ { pstr_desc =
+                  Pstr_eval (
+                    { pexp_desc = Pexp_field (srecord, lid); _ } as sexp, _
+                  )
+               } ] ->
+          let (record, record_sort, rmode, label, _) =
+            type_label_access Legacy env srecord Env.Mutation lid
+          in
+          Env.mark_label_used Env.Projection label.lbl_uid;
+          begin match (Types.atomic label.lbl_mut) with
+          | Nonatomic -> raise (Error (loc, env, Label_not_atomic lid.txt))
+          | Atomic -> ()
+          end;
+          let (_, ty_arg, ty_res) = instance_label ~fixed:false label in
+          unify_exp env record ty_res;
+          let alloc_mode, argument_mode = register_allocation expected_mode in
+          let mode = Modality.Value.Const.apply label.lbl_modalities rmode in
+          let mode = cross_left env ty_arg mode in
+          submode ~loc ~env mode argument_mode;
+          (* let uu = unique_use ~loc ~env mode (as_single_mode expected_mode) In *)
+          rue {
+            exp_desc =
+              Texp_atomic_loc
+                (record, record_sort, lid, label, alloc_mode);
+            exp_loc = loc; exp_extra = [];
+            exp_type = instance (Predef.type_atomic_loc ty_arg);
+            exp_attributes = sexp.pexp_attributes;
+            exp_env = env }
+      | _ ->
+          raise (Error (loc, env, Invalid_atomic_loc_payload))
+      end
   | Pexp_extension ext ->
     raise (Error_forward (Builtin_attributes.error_of_extension ext))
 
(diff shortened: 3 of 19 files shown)


goldfirere:
I'm not familiar with this construct. Could you describe it? Adding some documentation may be the best way...

glittershark:
> I'm not familiar with this construct. Could you describe it? Adding some documentation may be the best way...

ah, sure, sorry! Added some links to the PR description to the upstream manual source, but https://github.com/ocaml/ocaml/blob/trunk/manual/src/tutorials/parallelism.etex#L745-L791 is the relevant bit

typing/typecore.ml
```diff
@@ -6995,6 +6999,40 @@ and type_expect_
...
+          begin match (Types.atomic label.lbl_mut) with
+          | Nonatomic -> raise (Error (loc, env, Label_not_atomic lid.txt))
+          | Atomic -> ()
+          end;
+          let (_, ty_arg, ty_res) = instance_label ~fixed:false label in
+          unify_exp env record ty_res;
+          let alloc_mode, argument_mode = register_allocation expected_mode in
+          let mode = Modality.Value.Const.apply label.lbl_modalities rmode in
+          let mode = cross_left env ty_arg mode in
+          submode ~loc ~env mode argument_mode;
```
riaqn: My impression is that, at runtime, this will be a tuple containing a pointer to the record, and an integer representing some offset? In that sense, the locality of the tuple would be constrained by the record alone. For example, if the record is local, the tuple cannot be global (otherwise you have heap-to-stack pointer).

However, in the abstract sense, this is some `Atomic.t` containing the field? And that's why the modality are taken into account?

I think we should follow the runtime sense (let's discuss offline)

riaqn: After offline discussion, we agreed on certain design. However, I think that design is unsound. Let's discuss again.

Specifically, the modalities here are still tricky. If it's `@@ portable`, we will construct a `Atomic.loc.t` that we can write `nonportable` value into. Now, if we access the original field, we will get that `nonportable` value back, but typed as `portable`.

I think the safest way to type this construct is simply require everything is legacy (legacy mode, identity modality). There might be smarter ways to type this, but it will take some thinking.

To achieve what was intended by this feature, it might be easier to just support some attribute on `Pexp_field` and `Pexp_setfield`:
```
foo (r.x [@atomic_contended])
r.x <- "hello" [@atomic_contended]
```

glittershark: After discussing in person we decided to just forbid modalities entirely, which is now implemented at https://github.com/oxcaml/oxcaml/pull/4364

typing/typecore.ml
```diff
@@ -7001,6 +7005,42 @@ and type_expect_
...
+          | Nonatomic -> raise (Error (loc, env, Label_not_atomic lid.txt))
+          | Atomic -> ()
+          end;
+          let (_, ty_arg, ty_res) = instance_label ~fixed:false label in
+          unify_exp env record ty_res;
+          let alloc_mode, argument_mode = register_allocation expected_mode in
+          (* NOTE: we intentionally ignore lbl.ld_modalities here, since we
+             force it to always be legacy in typecore. The mode of the
+             atomic_loc is then exactly the mode of the record *)
+          let mode = cross_left env ty_arg rmode in
```
riaqn: If the record is local, the field is integer, the mode crossing would allow us to have `atomic.loc` at global. 

Maybe other axes are safe to cross? But I'm inclined toward doing nothing here, until we find some use cases that need this mode crossing.

glittershark: seems sensible to me.

typing/typecore.ml
```diff
@@ -11079,6 +11121,18 @@ let report_error ~loc env =
...
+        "Invalid %a payload, a record field access is expected."
+        Style.inline_code "[%atomic.loc]"
+  | Label_not_atomic lid ->
+      Location.errorf ~loc "The record field %a is not atomic"
+        (Style.as_inline_code longident) lid
+  | Modalities_on_atomic_field lid ->
+    Location.errorf ~loc
+      "Modalities are not allowed on fields given to %a (here, %a)"
+      Style.inline_code  "[%atomic.loc]"
+      (Style.as_inline_code longident) lid
```
riaqn, line 11434: We should have tests to demonstrate this error?
- for comonadic axes, we require identity modalities 
- For monadic axes, we require legacy modalities.

riaqn: We also need test to demonstrate that `[@atomic]` attribute on field doesn't affects modalities - users will have to write `@local once` etc. to override the legacy modalities to identity modalities for comonadic axes.

typing/uniqueness_analysis.ml
```diff
@@ -2294,6 +2294,11 @@ let rec check_uniqueness_exp ~overwrite (ienv : Ienv.t) exp : UF.t =
     let uf_write = Value.mark_implicit_borrow_memory_address Write value in
     let uf_tag = Value.invalidate_tag value in
     UF.pars [uf_rcd; uf_arg; uf_write; uf_tag]
+  | Texp_atomic_loc (rcd, _, _, _, _) ->
+    let value, uf_rcd = check_uniqueness_exp_as_value ienv rcd in
+    let uf_write = Value.mark_implicit_borrow_memory_address Write value in
+    let uf_tag = Value.invalidate_tag value in
+    UF.pars [uf_rcd; uf_write; uf_tag]
```
riaqn: This should be probably handled in `check_uniqueness_exp_as_value` instead (since we are making an alias of a record field, not "consuming" anything).
I will make a commit tomorrow.

lambda/translcore.ml
```diff
@@ -2270,6 +2282,11 @@ and transl_record ~scopes loc env mode fields repres opt_init_expr =
              transl_exp ~scopes init_expr_sort init_expr, lam)
     end
 
+and transl_atomic_loc ~scopes arg arg_sort lbl =
+  let arg = transl_exp ~scopes arg_sort arg in
+  let lbl = Lconst (Const_base (Const_int (lbl.lbl_pos))) in
+  (arg, lbl)
```
riaqn, line 2294: Can we just inline this function?

lambda/translcore.ml
```diff
@@ -638,6 +638,18 @@ and transl_exp0 ~in_new_scope ~scopes sort e =
         {fields; representation; extended_expression } ->
       transl_record_unboxed_product ~scopes e.exp_loc e.exp_env
         fields representation extended_expression
+  | Texp_atomic_loc (arg, arg_sort, _id, lbl, alloc_mode) ->
+      let shape =
+        Some
+          [ Typeopt.value_kind arg.exp_env arg.exp_loc arg.exp_type;
+            { raw_kind = Pintval; nullable = Non_nullable }
```
riaqn, line 648: Can you say a bit about why this is `Pintval`? (I'm not familiar with this)

typing/typecore.ml
glittershark: i ended up (in commit titled "Default atomic fields to id for comonadic axes") just defaulting to identity for comonadic axes, since it's really confusing to default to one thing but require another. See internal thread for a longer discussion (and we might want to change our minds here, so I'm leaving this thread open for now)

lambda/translcore.ml
glittershark: Upstream doesn't, and I kinda wanted to keep things similar to upstream.

lambda/translcore.ml
glittershark: The second field is just an integer representing the logical index of the field.

glittershark:
force push to rebase and resolve conflicts, only the latest commit, "Default atomic fields to id for comonadic axes", is new.

typing/uniqueness_analysis.ml
riaqn: I couldn't tell what `invalidate_tag` means. @anfelor could you kindly explain?

anfelor: The `invalidate_tag` function assigns the value the `Tag_was_mutated` state. In particular, the value can no longer be overwritten in-place since the analysis can not be sure that it has the correct tag.

riaqn: Hmm looking at the case above `Texp_setfield` - setting field shouldn't change the tag (in the GC sense)?

testsuite/tests/typing-modes/atomic_loc.ml
```diff
@@ -20,10 +20,15 @@ val atomic_loc_portable : 'a atomic @ portable -> 'a atomic_loc @ portable =
...
-(* This is allowed because atomic mutable implies aliased *)
 [%%expect{|
-val uses_unique :
-  'a atomic @ unique -> 'a atomic_loc * 'a atomic_loc @ unique = <fun>
+Line 2, characters 41-42:
+2 |   [%atomic.loc t.contents], [%atomic.loc t.contents]
+                                             ^
+Error: This value is used here, but it is already being used as unique:
+Line 2, characters 15-16:
+2 |   [%atomic.loc t.contents], [%atomic.loc t.contents]
```
goldfirere, line 31: This surprises me. Reading the test, I expected this one to work fine, by the fact that `unique < aliased`. That is, the `t` is used twice, but that's generally ok for a unique value. And the uniqueness of the `atomic.loc`s themselves is irrelevant.

testsuite/tests/typing-unique/unique_analysis.ml
```diff
@@ -806,3 +806,67 @@ Line 3, characters 20-21:
...
+  let _ = r.y in
+  ()
+[%%expect{|
+type 'a r = { mutable x : 'a [@atomic]; y : 'a; }
+Line 6, characters 10-11:
+6 |   let _ = r.y in
+              ^
+Error: This value is read from here, but it has already been used as unique:
+Line 4, characters 23-24:
+4 |   let x = [%atomic.loc r.x] in
```
goldfirere, line 824: I would expect this to pass as well: the uniqueness of the atomic.loc should be immaterial

testsuite/tests/typing-unique/unique_analysis.ml
```diff
@@ -806,3 +806,67 @@ Line 3, characters 20-21:
...
+                           ^
+
+|}]
+
+let foo (r : 'a r) =
+  let x = [%atomic.loc r.x] in
+  let _ = aliased_id x in
+  let _ = r.y in
+  ()
```
goldfirere, line 833: Does this test anything interesting?

testsuite/tests/typing-unique/unique_analysis.ml
```diff
@@ -806,3 +806,67 @@ Line 3, characters 20-21:
...
+|}]
+
+
+let foo (r : 'a r) =
+  let _ = r.y in
+  let x = [%atomic.loc r.x] in
+  let _ = unique_id x in
+  ()
+[%%expect{|
+val foo : 'a r @ unique -> unit = <fun>
```
goldfirere, line 845: I would not expect this to impose a uniqueness requirement on the original record

testsuite/tests/typing-unique/unique_analysis.ml
```diff
@@ -806,3 +806,67 @@ Line 3, characters 20-21:
...
+  let _ = r.y in
+  let _ = unique_id x in
+  ()
+[%%expect{|
+Line 3, characters 10-11:
+3 |   let _ = r.y in
+              ^
+Error: This value is read from here, but it has already been used as unique:
+Line 2, characters 23-24:
+2 |   let x = [%atomic.loc r.x] in
```
goldfirere, line 860: I would expect this to be accepted

testsuite/tests/typing-unique/unique_analysis.ml
```diff
@@ -806,3 +806,67 @@ Line 3, characters 20-21:
...
+                           ^
+
+|}]
+
+let foo (r : 'a r) =
+  let x = [%atomic.loc r.x] in
+  let _ = r.y in
+  let _ = aliased_id x in
+  ()
```
goldfirere, line 869: What is this testing?

testsuite/tests/typing-unique/unique_analysis.ml
goldfirere, line 1: I think I'd also like to see tests around linearity here: the memory block for the record mode-crosses linearity (I think), and linearity's submoding relation is backwards from uniqueness. These two facts might make linearity behave somewhat differently from uniqueness in cases like these.

I'd also like to see some tests with unique usages of record fields. That is, the tests as written seem to focus on the uniqueness of the atomic.loc, which I think is not all that interesting. On the other hand, the uniqueness of the record itself _is_ interesting, and so there should be some inspection of that.

goldfirere (commented):
Reviewed tests in last two commits ("fix UA" and "promote changes to uniqueness") only. Did not review the (very brief) implementation, as we're still converging on the specification.

testsuite/tests/typing-modes/atomic_loc.ml
riaqn: My impression of the offline resolution: `Atomic.loc` might cross uniqueness in the future, but for simplicity let's not do that in this PR.

Marked as resolved.

testsuite/tests/typing-unique/unique_analysis.ml
riaqn: DIto.

testsuite/tests/typing-unique/unique_analysis.ml
riaqn: Dito.

testsuite/tests/typing-unique/unique_analysis.ml
riaqn: Subsumed by later discussion. Marking as resolved.

testsuite/tests/typing-unique/unique_analysis.ml
riaqn: I'm not sure what to see about unique usages of record fields- if they are the same field as the `atomic.loc`, it will be under `many aliased` modalities and thus not tracked by UA. If they are not the same field, it won't interact with `atomic.loc`.

testsuite/tests/typing-unique/unique_analysis.ml
riaqn: Just some positive test - not anything in particular.

testsuite/tests/typing-unique/unique_analysis.ml
riaqn: Just some positive test, nothing in particular.

testsuite/tests/typing-unique/unique_analysis.ml
riaqn: Added tests for linerity

PR #4231, opened 2025-06-29
Forbid pattern matching on atomic fields

glittershark:


testsuite/tests/atomic-locs/record_fields.ml
goldfirere, line 1: Maybe just add a test with the `; _` syntax that covers an atomic?

goldfirere (approved):

PR #4233, opened 2025-06-30
chore: Use `kind_checks` for fallible kind-related invariant

bclement-ocp:
The invariant in `closure_conversion` that asserts that we are not trying to call the `Block_load` primitive on something that is not a block can fail in the presence of lazy values or GADTs, as noted by a CR. In fact, it *does* fail when compiling `parser.mly`.

Move this invariant to the newly introduced `kind_checks` instead, since it is (sub)kind-related and the `kind_checks` already contain many fallible checks, whereas invariants should always hold.


lthls (approved):
LGTM

PR #4234, opened 2025-06-30
Deterministic order for CSE parameters based on binding times

bclement-ocp:
This patch ensures that the `cse_param` variables introduced by the CSE pass are introduced in a stable order that does not depend on the `Int_ids` assigned to variables.

This helps making comparisons between different runs (e.g. with different options) using the `flambda2-compare` library.

The order is determined lexicographically depending on the primitive under consideration and its arguments, which are guaranteed to be in scope in the target environment (see `cse_with_eligible_lhs`). The order for the argument uses the new `stable_compare_simples` function which is a total order on `Simple.t` values defined in a given typing environment, which coincides with the binding time order and does not depend on `Int_ids` hashing.

### DIFF
--- a/middle_end/flambda2/simplify/common_subexpression_elimination.ml
+++ b/middle_end/flambda2/simplify/common_subexpression_elimination.ml
@@ -326,39 +326,6 @@ module Join_result = struct
     }
 end
 
-let stable_compare_primitives tenv (prim1 : P.t) (prim2 : P.t) =
-  match prim1, prim2 with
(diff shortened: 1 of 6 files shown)


lthls (commented):
The PR looks correct, but I'm wondering if you considered having a version of `Flambda_primitive.compare` that is generic on the simple comparison primitive ?
This would avoid having what is essentially a copy of `Flambda_primitive.compare` in `Common_subexpression_elimination`.

PR #4236, opened 2025-06-30
Change how extension level affects inferred modalities

riaqn:
Currently, inferred modalities are fixed to be identity if `mode` extension is disabled. This causes some incompatibility between modules that has `mode` enabled and those not. For example, if we have `a.ml` that has `mode` enabled, and we have `b.ml` with `mode` disabled:
```
include A
```
and `b.mli`:
```
include module type of A
```
The later will have non-identity modalities while the former doesn't.

This is currently not testable, since with modules fixed at `legacy`, the type checker simply copies the modalities from `A` into `B` without infering new modalities. But #3759 will be affected.

The solution is to always infer modalties, but use extension level to control what the users can write and read.

### DIFF
--- a/typing/ctype.ml
+++ b/typing/ctype.ml
@@ -5012,6 +5012,11 @@ let relevant_pairs pairs v =
   | Contravariant -> pairs.contravariant_pairs
   | Bivariant -> pairs.bivariant_pairs
 
+let zap_modalities_to_floor_if_at_least level =
+  if Language_extension.(is_at_least Mode level)
+    then Mode.Modality.Value.zap_to_floor
+    else Mode.Modality.Value.zap_to_id
+
 let crossing_of_jkind env jkind =
   let jkind_of_type = type_jkind_purely_if_principal env in
   Jkind.get_mode_crossing ~jkind_of_type jkind
--- a/typing/ctype.ml
+++ b/typing/ctype.ml
@@ -5012,7 +5012,7 @@ let relevant_pairs pairs v =
   | Contravariant -> pairs.contravariant_pairs
   | Bivariant -> pairs.bivariant_pairs
 
-let zap_modalities_to_floor_if_at_least level =
+let zap_modalities_to_floor_if_modes_enabled_at level =
   if Language_extension.(is_at_least Mode level)
     then Mode.Modality.Value.zap_to_floor
     else Mode.Modality.Value.zap_to_id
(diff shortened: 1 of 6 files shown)


typing/ctype.ml
```diff
@@ -5012,6 +5012,11 @@ let relevant_pairs pairs v =
   | Contravariant -> pairs.contravariant_pairs
   | Bivariant -> pairs.bivariant_pairs
 
+let zap_modalities_to_floor_if_at_least level =
```
lukemaurer: Oof. This is the first I'm realizing that the maturity `Alpha` is _greater than_ the maturity `Stable`. That's so confusing that “at least” is probably not the best phrasing here. Maybe `zap_modalities_to_floor_if_modes_enabled_at`? It's easier to reason that if `Alpha` is enabled then so is `Stable` than to try and remember which way around maturities get compared

typing/typemod.ml
```diff
@@ -3565,7 +3557,7 @@ let type_module_type_of env smod =
   check_nongen_modtype env smod.pmod_loc mty;
   let mty =
     remove_modality_and_zero_alloc_variables_mty env
-      ~zap_modality:Mode.Modality.Value.zap_to_floor mty
+      ~zap_modality:(Ctype.zap_modalities_to_floor_if_at_least Stable) mty
```
lukemaurer: I'm surprised we want this change? What happens if we just zap to floor here?

typing/typemod.ml
```diff
@@ -3916,7 +3909,8 @@ let type_implementation target modulename initial_env ast =
             (* Generating [cmi] without [mli]. This [cmi] will only be on the
                LHS of inclusion check, so we zap to floor (strongest). *)
             remove_modality_and_zero_alloc_variables_sg finalenv
-              ~zap_modality:Mode.Modality.Value.zap_to_floor simple_sg
+              ~zap_modality:(Ctype.zap_modalities_to_floor_if_at_least Stable)
```
lukemaurer: Similarly here, what happens if we just always zap to floor? (Definitely worth updating the comment in any case)

typing/typemod.ml
riaqn: If we just zap to floor, the obtained module type might be stronger than what the users want. Now, we've established that user code that are affected by this are just bad code, but it doesn't hurt to provide a workaround?

typing/typemod.ml
riaqn: Similarly, it's giving the user a workaround in case they don't like the stronger module type.

lukemaurer: Makes sense. I would just tweak the comment to match the code more closely

typing/typemod.ml
lukemaurer: Discussed offline; the user is manually asking for a module type here so it makes sense to zap to their expectation

PR #4239, opened 2025-06-30
First cut of runtime dynamic bindings

NickBarnes:
This is the first draft of a runtime implementation of dynamic variable binding.

### DIFF
--- a/otherlibs/systhreads/st_stubs.c
+++ b/otherlibs/systhreads/st_stubs.c
@@ -136,6 +136,7 @@ struct caml_thread_struct {
   void * signal_stack;       /* this thread's signal stack */
   size_t signal_stack_size;  /* size of this thread's signal stack in bytes */
   int is_main;               /* whether this is the main thread of its domain */
+  dynamic_thread_t dynamic;  /* dynamic value bindings */
 
 #ifndef NATIVE_CODE
   intnat trap_sp_off;      /* saved value of Caml_state->trap_sp_off */
@@ -260,6 +261,7 @@ static void caml_thread_scan_roots(
     do {
       (*action)(fdata, th->descr, &th->descr);
       (*action)(fdata, th->backtrace_last_exn, &th->backtrace_last_exn);
+      caml_dynamic_scan_thread_roots(th->dynamic, action, fflags, fdata);
       /* Don't rescan the stack of the current thread, it was done already */
       if (th != active) {
         if (th->current_stack != NULL)
@@ -331,6 +333,7 @@ static void restore_runtime_state(caml_thread_t th)
   Caml_state->external_raise_async = th->external_raise_async;
 #endif
   caml_memprof_enter_thread(th->memprof);
+  caml_dynamic_enter_thread(th->dynamic);
 }
 
 CAMLexport void caml_thread_restore_runtime_state(void)
@@ -395,7 +398,7 @@ static void caml_thread_leave_blocking_section(void)
 /* Create and setup a new thread info block.
    This block has no associated thread descriptor and
    is not inserted in the list of threads. */
-static caml_thread_t caml_thread_new_info(void)
+static caml_thread_t caml_thread_new_info(caml_thread_t parent)
 {
   caml_thread_t th = NULL;
   caml_domain_state *domain_state = Caml_state;
@@ -415,10 +418,12 @@ static caml_thread_t caml_thread_new_info(void)
      must be initialized to a valid value, as in [domain_create]. */
 
   th->current_stack = caml_alloc_main_stack(stack_wsize);
-  if (th->current_stack == NULL) goto out_err1;
+  if (th->current_stack == NULL) goto fail_alloc_stack;
 
   th->memprof = caml_memprof_new_thread(domain_state);
-  if (th->memprof == NULL) goto out_err2;
+  if (th->memprof == NULL) goto fail_memprof;
+  th->dynamic = caml_dynamic_new_thread(parent->dynamic);
+  if (th->dynamic == NULL) goto fail_dynamic;
 
   th->c_stack = NULL;
   th->local_roots = NULL;
@@ -439,9 +444,11 @@ static caml_thread_t caml_thread_new_info(void)
 #endif
   return th;
 
- out_err2:
+ fail_dynamic:
+  caml_memprof_delete_thread(th->memprof);
+ fail_memprof:
   caml_free_stack(th->current_stack);
- out_err1:
+ fail_alloc_stack:
   caml_stat_free(th);
   return NULL;
 }
@@ -466,6 +473,7 @@ void caml_thread_free_info(caml_thread_t th)
      init_mask: stack-allocated
   */
   caml_memprof_delete_thread(th->memprof);
+  caml_dynamic_delete_thread(th->dynamic);
   caml_free_stack(th->current_stack);
   caml_free_backtrace_buffer(th->backtrace_buffer);
 
@@ -497,7 +505,7 @@ static value caml_thread_new_descriptor(value clos)
 /* Allocate a thread info block and add it to the list of threads */
 static caml_thread_t thread_alloc_and_add(void)
 {
-  caml_thread_t th = caml_thread_new_info();
+  caml_thread_t th = caml_thread_new_info(Active_thread);
 
   if (th == NULL) return NULL;
 
@@ -684,12 +692,15 @@ static void caml_thread_domain_initialize_hook(void)
   new_thread->prev = new_thread;
   new_thread->backtrace_last_exn = Val_unit;
   new_thread->memprof = caml_memprof_main_thread(Caml_state);
+  new_thread->dynamic = Caml_state->dynamic_bindings;
+  CAMLassert(new_thread->dynamic);
   new_thread->is_main = 1;
   new_thread->signal_stack = NULL;
 
   This_thread = new_thread;
   Active_thread = new_thread;
   caml_memprof_enter_thread(new_thread->memprof);
+  caml_dynamic_enter_thread(new_thread->dynamic);
 }
 
 static void thread_yield(void);
--- a/otherlibs/systhreads/st_stubs.c
+++ b/otherlibs/systhreads/st_stubs.c
@@ -261,12 +261,12 @@ static void caml_thread_scan_roots(
     do {
       (*action)(fdata, th->descr, &th->descr);
       (*action)(fdata, th->backtrace_last_exn, &th->backtrace_last_exn);
-      caml_dynamic_scan_thread_roots(th->dynamic, action, fflags, fdata);
       /* Don't rescan the stack of the current thread, it was done already */
       if (th != active) {
         if (th->current_stack != NULL)
           caml_do_local_roots(action, fflags, fdata,
-                              th->local_roots, th->current_stack, th->gc_regs);
+                              th->local_roots, th->current_stack, th->gc_regs,
+                              th->dynamic);
       }
       th = th->next;
     } while (th != active);
--- a/runtime/amd64.S
+++ b/runtime/amd64.S
@@ -154,6 +154,14 @@
 #define Handler_effect(REG)     16(REG)
 #define Handler_parent          24
 
+/* struct dynamic_thread_s */
+
+/* If you change Dynamic_cache_entries, you must also update FLUSH_DYNAMIC_CACHE */
+#define Dynamic_cache            0 /* offset of cache in dynamic_thread_t */
+#define Dynamic_binding_size     (2*8) /* dyn, val */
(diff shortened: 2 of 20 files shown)


NickBarnes:
Back out of draft, with `testsuite/tests/effects/dynamic.ml` as a working set of tests defining a `Dynamic.t`.

otherlibs/systhreads/st_stubs.c
```diff
@@ -136,6 +136,7 @@ struct caml_thread_struct {
   void * signal_stack;       /* this thread's signal stack */
   size_t signal_stack_size;  /* size of this thread's signal stack in bytes */
   int is_main;               /* whether this is the main thread of its domain */
+  dynamic_thread_t dynamic;  /* dynamic value bindings */
```
TheNumbat, line 139: If I understand correctly, the dynamic value can be bound at three levels: the initial value (set by `Dynamic.create`), the current thread table (set by `Dynamic.set_root`, copied by threads) and the nearest fiber binding (set by `Dynamic.with_temporarily`, inherited by fibers) -- this should be written down somewhere (maybe in fiber.h)

otherlibs/systhreads/st_stubs.c
```diff
@@ -334,6 +336,7 @@ static void restore_runtime_state(caml_thread_t th)
   Caml_state->external_raise_async = th->external_raise_async;
 #endif
   caml_memprof_enter_thread(th->memprof);
+  caml_dynamic_enter_thread(th->dynamic);
```
TheNumbat, line 336: I assume we don't need to save `Caml_state->dynamic_bindings` in `save_runtime_state` because it never changes between thread swaps?

runtime/amd64.S
```diff
@@ -154,6 +154,11 @@
 #define Handler_effect(REG)     16(REG)
 #define Handler_parent          24
 
+/* struct dynamic_thread_s */
+#define Dynamic_cache            0
+#define Dynamic_binding_size     (2*8) /* dyn, val */
+#define Dynamic_cache_entries    8 /* must match DYNAMIC_CACHE_SIZE in fiber.c */
```
TheNumbat, line 163: Can we define these in `config.h` instead?

runtime/amd64.S
```diff
@@ -284,6 +289,19 @@
...
+#define FLUSH_DYNAMIC_CACHE(reg)                    \
+        movq    Caml_state(dynamic_bindings), reg; \
+        movq    $0, (Dynamic_cache + 0*Dynamic_binding_size)(reg) ; \
+        movq    $0, (Dynamic_cache + 1*Dynamic_binding_size)(reg) ; \
+        movq    $0, (Dynamic_cache + 2*Dynamic_binding_size)(reg) ; \
+        movq    $0, (Dynamic_cache + 3*Dynamic_binding_size)(reg) ; \
+        movq    $0, (Dynamic_cache + 4*Dynamic_binding_size)(reg) ; \
+        movq    $0, (Dynamic_cache + 5*Dynamic_binding_size)(reg) ; \
+        movq    $0, (Dynamic_cache + 6*Dynamic_binding_size)(reg) ; \
+        movq    $0, (Dynamic_cache + 7*Dynamic_binding_size)(reg)
```
TheNumbat: Any way to assert this matches `Dynamic_cache_entries`?

runtime/caml/fiber.h
```diff
@@ -67,14 +67,20 @@ struct stack_info {
   intnat local_sp;
   void* local_top;
   intnat local_limit;
+
+  /* Dynamic binding */
+  /* TODO: Consider making this a table of bindings */
```
TheNumbat, line 72: I don't need this right away, but seems important to support soon

runtime/caml/fiber.h
```diff
@@ -67,14 +67,20 @@ struct stack_info {
...
+  /* TODO: improve the names of these slots */
```
TheNumbat: Maybe dyn_key dyn_value?

runtime/fiber.c
```diff
@@ -733,6 +742,10 @@ void caml_scan_stack(
       }
     }
 
+    /* Scan dynamic binding, both key and value */
+    f(fdata, stack->dyn, &stack->dyn);
+    f(fdata, stack->bind, &stack->bind);
```
TheNumbat: Should these check is_scannable?

runtime/fiber.c
```diff
@@ -1117,3 +1133,318 @@ CAMLexport void caml_raise_unhandled_effect (value effect)
...
+
+extern value caml_fresh_oo_id(value v);
+
+/* Making a dynamic value */
+
+CAMLprim value caml_dynamic_make(value val)
+{
+  CAMLparam1(val);
+  /* TODO: consider other hash functions. This one is ~unique, which is nice */
+  value hash = caml_fresh_oo_id(Val_unit);
```
TheNumbat, line 1160: Is there any way to give it a stable hash without adding an indirection to cache lookup? 
Alternatively could we use its address as the cache hash (but not the table hash) and just lose track of the cache entry when it moves? Or clear the entry when moved?

runtime/fiber.c
```diff
@@ -1117,3 +1133,318 @@ CAMLexport void caml_raise_unhandled_effect (value effect)
...
+  /* Copy existing bindings */
+  for (size_t i = 0; i < old_capacity; ++ i) {
+    if (Is_bound(&old_bindings[i])) {
+      dynamic_table_add(table, old_bindings[i].dyn, old_bindings[i].val);
+    }
+  }
+  if (old_bindings) {
+    caml_stat_free(old_bindings);
+  }
+  table->bindings = new_bindings;
```
TheNumbat: Repeated above

runtime/fiber.c
```diff
@@ -1117,3 +1133,318 @@ CAMLexport void caml_raise_unhandled_effect (value effect)
...
+
+CAMLexport dynamic_thread_t caml_dynamic_new_thread(dynamic_thread_t parent)
+{
+  dynamic_thread_t res = caml_stat_alloc_noexc(sizeof(dynamic_thread_s));
+  if (!res) {
+    goto fail_alloc;
+  }
+  /* Not even going to think about the semantics of copying cache entries */
+  dynamic_flush_cache(res);
+ dynamic_table_fresh(&res->base);
```
TheNumbat: Indent

runtime/fiber.c
```diff
@@ -1117,3 +1133,318 @@ CAMLexport void caml_raise_unhandled_effect (value effect)
...
+  bool res = dynamic_table_bind(&thread->base, dyn, val);
+  if (!res) {
+    caml_raise_out_of_memory();
+  }
+
+  /* invalidate cache */
+  uintnat hash = Hash_dyn(dyn);
+  uintnat index = hash & (DYNAMIC_CACHE_SIZE - 1);
+  binding_t entry = thread->cache + index;
+  if (entry->dyn == dyn) {
```
TheNumbat, line 1412: Note this doesn't need to probe since the cache doesn't do chaining (or should it?)

runtime/fiber.c
```diff
@@ -1117,3 +1133,318 @@ CAMLexport void caml_raise_unhandled_effect (value effect)
...
+  value val;
+  /* Not in cache; let's look at the fiber */
+  CAMLassert(Caml_state->current_stack);
+  struct stack_info *stack = Caml_state->current_stack;
+  while (stack) {
+    if (dyn == stack->dyn) {
+      val = stack->bind;
+      goto found;
+    }
+    stack = Stack_parent(stack);
```
TheNumbat, line 1443: Probably doesn't matter, but could guard against finding a cycle

testsuite/tests/effects/dynamic.ml
```diff
@@ -0,0 +1,254 @@
...
+   Thread.join (Thread.create (fun () -> Printf.printf "get on child thread created after set_root [expect 2]: %d\n" (Dynamic.get d)) ());
+   r := d;
+   go ();
+   Thread.join (Thread.create (fun () -> Dynamic.set_root d 4) ());
+   Printf.printf "get after set_root on child thread [expect 2]: %d\n" (Dynamic.get d))
+
+(* `with_temporarily` should change the value seen:
+   - within its dynamic extent;
+   - but not on some other thread while it is running;
+   - or after it returns;
```
TheNumbat, line 126: What about a thread created inside of `with_temporarily` (presumably no - is that what we want?)

testsuite/tests/effects/dynamic.ml
```diff
@@ -0,0 +1,254 @@
...
+      check_other_thread();
+      Dynamic.set_root d (n+3);
+      Effect.Deep.continue k ();
+      Printf.printf "after continuation [expect %d]: %d\n" (n+5) (Dynamic.get d))
+    | e -> None in
+
+  (Dynamic.set_root d n;
+   Effect.Deep.match_with f ()
+     { retc = (fun () -> (Printf.printf "after fiber [expect %d]: %d\n" (n+4) (Dynamic.get d);
+                          check_other_thread();
```
TheNumbat, line 210: Don't think it'd be any different, but we could also check `with_temporarily` works in `retc`/`exnc`

otherlibs/systhreads/st_stubs.c
NickBarnes: Yes. The value of the `dynamic_thread_t` for any given thread never changes (although its contents do).

runtime/amd64.S
NickBarnes: Possibly (well, DYNAMIC_CACHE_BITS at least, from which DYNAMIC_CACHE_SIZE is derived). I'm going to take this as a process improvement suggestion and think about better ways to share this structure-layout information between asm and C (and OCaml)

runtime/amd64.S
NickBarnes: Iterating in the C preprocessor is hard (although not impossible) to get right. I'm punting to our future selves. I've added a comment adjacent to the `DYNAMIC_CACHE_BITS` definition in `fiber.c`.

runtime/caml/fiber.h
NickBarnes: Easy to fix this, but I don't know what the requirement is. Leaving to later.

otherlibs/systhreads/st_stubs.c
NickBarnes: Yes. The `.mli` interface should cover this in more detail (see my separate draft of that), but we should have something in this header file for C coders. Doing that.

runtime/fiber.c
NickBarnes: I decided not. Bring me real code where it makes a difference and I'll revisit.

runtime/fiber.c
NickBarnes: I thought about putting Floyd's tortoise-and-hare algorithm in here, but decided against: if we ever get a cycle in the stacks graph then everything will get stuck in an infinite loop (e.g. the GC) and that should be easy to find and fix. If we do want to guard against it, we should probably do so in `caml_scan_stack` rather than here.

testsuite/tests/effects/dynamic.ml
NickBarnes: Do people want nasally-inserted fire? I genuinely don't know. This could be done by copying fiber binding values into the per-thread root bindings, at thread creation time. Leaving this as-is until somebody clarifies the requirement.

NickBarnes:
Addressed all review comments and rebased.

runtime/caml/fiber.h
NickBarnes: I've settled on `dyn` and `val`.

runtime/fiber.c
NickBarnes: I think it's implicit in the API that the scanner `f` is a nop unless is_scannable. For example, `oldify_one` begins with `Is_block` and `Is_young` tests; `caml_darken` begins with `Is_markable` (which is `Is_block` and not `Is_young`); `compact_update_value` begins with `Is_block`. Of course this should be written down somewhere....

runtime/fiber.c
NickBarnes: There are various possibilities here, but I am deliberately punting performance hacks to later work, and focusing here on getting the semantics right. The same goes for @stedolan's notion to use a vector unit hack for the cache check.
If we used some bits of the address as the hash, then on a minor GC we'd not only have to flush young items from the per-thread cache, but also rehash all the per-thread root-binding tables (which use the same hash value).

stedolan (changes requested):
I think this is missing a marking of `Caml_state->dynamic_bindings` - it's covered by the one in systhreads when systhreads is in use (almost always), but it's still possible to not have st_stubs linked in.

stedolan (approved):

PR #4242, opened 2025-06-30
Remove with-bounds for GADTs

glittershark:
This caused a major performance regression in some edge cases that we don't yet understand, so we're removing it temporarily until we can track down the issue.


typing/jkind.mli
```diff
@@ -513,16 +513,7 @@ val for_unboxed_record : Types.label_declaration list -> Types.jkind_l
 
     [decl_params] is the parameters in the head of the type declaration. [type_apply]
     should be [Ctype.apply] partially applied to an [env]. *)
```
goldfirere: This comment is now out of date

goldfirere (approved):

PR #4020, opened 2025-05-20
Allow for unboxed elements at the top-level of modules

jra4:
Lift the value-only restriction on module fields.

New types `Types.module_representation` and `Lambda.module_representation` have been added to track which modules have non-value elements. These module representations (or similar) now have to be passed around to compile modules properly. Some notable examples:

  - In `typing/`:
      - Thanks to an earlier feature, `Tpat_var` and `Tpat_alias` (the two patterns that bind idents) contain the sort of the pattern. This puts sort information into structure items.
      - The constructor `Val_reg sort : value_kind` has a new element `sort : Jkind.Sort.t`. This puts sort information into signature items.
      - The constructor `Env.Adot(root, module_repr, pos) : address` has a new element `module_repr : Types.module_representation`, representing the fields in the module `root`.
      - `Tcoerce_structure`, `Tincl_functor`, and `Tincl_gen_functor` now store the module representation of their inputs.
      - While it is possible to convert a signature into a module representation, it is more ergonomic for some data structures to store a module representation alongside a signature. For example: `open_infos` and `include_infos` in `typedtree.mli`.
  - In `lambda/`:
      - `mod_field` now requires a module representation to switch between `Pfield` and `Pmixedfield` when accessing a module field.
      - A new function, `block_of_module_representation`, is used to switch between `Pmakeblock` and `Pmakemixedblock` when allocating a module.
      - Many places which previously represented a module as a `size : int` now use `repr : module_representation`. Examples include `main_module_block_format` and `transl_structure`.
  - In `middle_end/flambda2/`:
      - Like in `lambda/`, modules can no longer be represented by their size. Now, a module representation is threaded though a bunch of functions until it is used in `wrap_final_module_block`. This path is only used for translating file-level modules.

We chose not to resolve sort variables until after typing is completed (see `sort_variable_defaulting.ml`). Thus, `Types.module_representation` is an array of sorts, and `Lambda.module_representation` is either `Module_value_only size` or `Module_mixed mixed_block_shape`. We default sort variables in `transl_module_representation : Types.module_representation -> module_representation`. We check that the scannable value prefix is not too big in `block_of_module_representation`, the latest possible place. This means it is possible to define uninhabited module types and functors that take uninhabited module types as input.

[...]

PR #4243, opened 2025-06-30
Flambda2: Store kind on variables

lthls:
This PR updates the data associated to each Flambda2 variable to also include the kind (but not the subkind).
The rationale for this PR is that several parts of Flambda2 need to know the kinds of the variables (or simples), and tracking this independently requires some work (typically, the typing env stores the kind for every variable).
This is particularly problematic in the case where a cmx might be missing: currently we can load a cmx containing occurrences of a variable coming from another compilation unit, whose cmx is now unavailable, and in this case we cannot know the kind of the variable (at least not directly). This happens in practice with the new n-way join, and #4212 is a workaround for this issue.
With this PR, the variable would have its kind stored in all cmx files where it occurs, so even with a missing cmx we would be able to access the kind.
Currently the PR doesn't change the existing code to make use of the stored kinds, but a follow-up PR could clean the typing environment to avoid storing kinds completely.

### DIFF
--- a/middle_end/flambda2/from_lambda/closure_conversion.ml
+++ b/middle_end/flambda2/from_lambda/closure_conversion.ml
@@ -76,7 +76,7 @@ let register_const0 acc constant name =
   match Static_const.Map.find constant (Acc.shareable_constants acc) with
   | exception Not_found ->
     (* Create a variable to ensure uniqueness of the symbol. *)
-    let var = Variable.create name in
+    let var = Variable.create name K.value in
     let acc, symbol =
       manufacture_symbol acc
         (* CR mshinwell: this Variable.rename looks to be redundant *)
@@ -404,7 +404,7 @@ module Inlining = struct
     let inlined_debuginfo =
       Inlined_debuginfo.create ~called_code_id ~apply_dbg
     in
-    let inlined_dbg_var = Variable.create "inlined_dbg" in
+    let inlined_dbg_var = Variable.create "inlined_dbg" K.value in
     let inlined_dbg_var_duid = Flambda_debug_uid.none in
     Let_with_acc.create acc
       (Bound_pattern.singleton
@@ -753,13 +753,15 @@ let close_c_call acc env ~loc ~let_bound_ids_with_kinds
           when Stdlib.( = ) src src_kind && Stdlib.( = ) dst dst_kind -> (
           match args with
           | [arg] ->
-            let result = Variable.create "reinterpreted" in
+            let prim = P.Unary (Reinterpret_64_bit_word op, arg) in
+            let result =
+              Variable.create "reinterpreted" (P.result_kind' prim)
+            in
             let result_duid = Flambda_debug_uid.none in
             let result' =
               Bound_var.create result result_duid Name_mode.normal
             in
             let bindable = Bound_pattern.singleton result' in
-            let prim = P.Unary (Reinterpret_64_bit_word op, arg) in
             let acc, return_result =
               Apply_cont_with_acc.create acc return_continuation
                 ~args:[Simple.var result]
@@ -808,13 +810,14 @@ let close_c_call acc env ~loc ~let_bound_ids_with_kinds
         | None -> fun args acc -> call (arg :: args) acc
         | Some named ->
           fun args acc ->
-            let unboxed_arg = Variable.create "unboxed" in
+            let prim = P.Unary (named, arg) in
+            let unboxed_arg = Variable.create "unboxed" (P.result_kind' prim) in
             let unboxed_arg_duid = Flambda_debug_uid.none in
             let unboxed_arg' =
               VB.create unboxed_arg unboxed_arg_duid Name_mode.normal
             in
             let acc, body = call (Simple.var unboxed_arg :: args) acc in
-            let named = Named.create_prim (Unary (named, arg)) dbg in
+            let named = Named.create_prim prim dbg in
             Let_with_acc.create acc
               (Bound_pattern.singleton unboxed_arg')
               named ~body)
@@ -1716,7 +1719,7 @@ let close_apply_cont acc env ~dbg cont trap_action args : Expr_with_acc.t =
 let close_switch acc env ~condition_dbg scrutinee (sw : IR.switch) :
     Expr_with_acc.t =
   let scrutinee = find_simple_from_id env scrutinee in
-  let untagged_scrutinee = Variable.create "untagged" in
+  let untagged_scrutinee = Variable.create "untagged" K.naked_immediate in
   let untagged_scrutinee_duid = Flambda_debug_uid.none in
   (* CR sspies: We should probably have something like a phantom let here, even
      though the target variable isn't user-visible. Then in the debugger, if
@@ -1762,7 +1765,7 @@ let close_switch acc env ~condition_dbg scrutinee (sw : IR.switch) :
              Simple.const (Reg_width_const.naked_immediate case) ))
         condition_dbg
     in
-    let comparison_result = Variable.create "eq" in
+    let comparison_result = Variable.create "eq" K.naked_immediate in
     let comparison_result_duid = Flambda_debug_uid.none in
     let comparison_result' =
       VB.create comparison_result comparison_result_duid Name_mode.normal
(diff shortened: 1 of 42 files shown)


middle_end/flambda2/to_cmm/to_cmm_env.ml
```diff
@@ -653,7 +653,9 @@ and split_in_env env res var binding =
     let env, res =
       List.fold_left
         (fun (env, res) new_binding ->
-          let flambda_var = Variable.create "to_cmm_tmp" in
+          let flambda_var =
+            Variable.create "to_cmm_tmp" Flambda_kind.value (*TODO:fix*)
```
Gbury: This kind should never be used, so any kind would work here I think.

I tried to see what it would take to propagate correct kinds here, and it would involve a lot of code everywhere in `to_cmm_env.ml`, and it didn't seem worth it. Ideally, it would be great if we had an 'invalid'/error if ever found kind for this situation, but if there is'nt one, anything should do here I'd say.

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives_helpers.ml
```diff
@@ -345,7 +351,11 @@ let rec bind_recs acc exn_cont ~register_const0 (prim : expr_primitive)
       bind_recs acc exn_cont ~register_const0 ifso_or_ifnot dbg
       @@ fun acc ifso_or_ifnot ->
       let result_vars =
-        List.map (fun _ -> Variable.create (name ^ "_result")) ifso_or_ifnot
+        List.map
+          (fun k ->
+            Variable.create (name ^ "_result")
+              (Flambda_kind.With_subkind.kind k))
+          result_kinds
```
bclement-ocp, line 391: Should this be using the `Named.kind` from `ifso_or_ifnot` instead? Although I think these are guaranteed to be the `result_kinds`.

middle_end/flambda2/types/env/join_env.ml
```diff
@@ -245,6 +247,11 @@ end = struct
     in
     match shared_name with Some raw_name -> raw_name | None -> "join_var"
 
+  let kind (t : t) =
+    match Index.Map.min_binding_opt (t :> Simple.t Index.Map.t) with
```
bclement-ocp: Using `min_binding_opt` here make it look like we care about which binding we use, but we don't, so `choose_opt` would be better.

middle_end/flambda2/reaper/traverse.ml
```diff
@@ -69,7 +74,8 @@ let prepare_code ~denv acc (code_id : Code_id.t) (code : Code.t) =
       (fun i ->
         Code_id_or_name.var
           (Variable.create
-             (Printf.sprintf "witness_%d_for_%s" i (Code_id.name code_id))))
+             (Printf.sprintf "witness_%d_for_%s" i (Code_id.name code_id))
+             Flambda_kind.value))
```
bclement-ocp, line 81: I understand that these variables (and also `not_pure` below) are just a device used by the reaper and don't exist in the code (so they have "type" `unit`), is that right?

middle_end/flambda2/to_cmm/to_cmm_set_of_closures.ml
```diff
@@ -727,7 +727,7 @@ let let_dynamic_set_of_closures0 env res ~body ~bound_vars set
       ~mode:(C.alloc_mode_for_allocations_to_cmm closure_alloc_mode)
       dbg ~tag l memory_chunks
   in
-  let soc_var = Variable.create "*set_of_closures*" in
+  let soc_var = Variable.create "*set_of_closures*" Flambda_kind.value in
```
bclement-ocp, line 745: I assume this is not strictly speaking correct, but that we won't actually need the kind afterwards (like `"to_cmm_tmp"`)?

middle_end/flambda2/terms/flambda.mli
```diff
@@ -187,6 +187,9 @@ module Named : sig
       necessarily type-correct, at the given kind. *)
   val dummy_value : Flambda_kind.t -> t
 
+  (** Return the kind of the expression *)
+  val kind : t -> Flambda_kind.t
```
bclement-ocp, line 192: This might deserve a comment stating that it should not be called on set of closures or static consts as those don't have a kind.

middle_end/flambda2/from_lambda/closure_conversion.ml
```diff
@@ -3539,7 +3563,9 @@ let wrap_final_module_block acc env ~program ~prog_return_cont
     let field_vars =
       List.init module_block_size_in_words (fun pos ->
           let pos_str = string_of_int pos in
-          pos, Variable.create ("field_" ^ pos_str))
+          ( pos,
+            Variable.create ("field_" ^ pos_str)
+              K.value (* TODO: handle non-value layouts in modules *) ))
```
bclement-ocp: Is this a TODO for the PR or should it be a CR? Looking at the code below (using a `block_access_kind` of `Any_value`) it doesn't seem we support this anyway.

middle_end/flambda2/identifiers/int_ids.ml
```diff
@@ -203,25 +203,29 @@ module Variable_data = struct
     { compilation_unit : Compilation_unit.t;
       name : string;
       name_stamp : int;
+      kind : Flambda_kind.t;
```
bclement-ocp, line 231: Not saying that we should do this, but I think if we are ever concerned about cmx sizes, we could encode the kind in the tag of the `Variable_data` and still (barely) be a small block.

lthls: Noted, but I think that if cmx size was a problem I'd rather merge the kind with the stamp and visibility than store the kind in the tag.

middle_end/flambda2/from_lambda/closure_conversion.ml
lthls: I think I was afraid that by the time this PR lands, support for non-value fields in modules would be there (I think that's being worked on in #4020). I'll switch to a CR.

middle_end/flambda2/terms/flambda.mli
lthls: Right, I'll expand the documentation comment.

middle_end/flambda2/to_cmm/to_cmm_set_of_closures.ml
Gbury: Actually, this should be correct: the variable refers to a whole set of closures, which **must** be a block since it will be scanned by the GC.

lthls: Yes, exactly. This corresponds to something that has no Flambda2 kind, but since it would be correct to give it the `typ_val` machtype I don't see anything that could go wrong even if the kind wasn't ignored.

middle_end/flambda2/types/env/join_env.ml
lthls: Indeed, I will change it.

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives_helpers.ml
lthls: I would rather keep iterating on `result_kinds`, because it was more obvious to me that `result_vars` should have kinds matching `result_kinds`. As far as I can tell they should match.

bclement-ocp: Got it, thanks. My comment was poorly worded — I agree iterating on `result_kinds` is better if they do match, but we were iterating on `ifso_or_ifnot` before.

middle_end/flambda2/to_cmm/to_cmm_env.ml
lthls: I'm replacing the TODO with a proper comment.

middle_end/flambda2/reaper/traverse.ml
lthls: I don't remember. I think I must have asked @chambart or @Ekdohibs when I wrote the PR. I *think* that these variables never end up in the rebuilt code so their kinds don't matter, but it would be nice to get confirmation.

lthls: I got confirmation from @chambart; I'll add a comment for future readers.

PR #4245, opened 2025-06-30
Fix a bug in calling `caml_callbackN` on functions that use local arguments / intermediate local closures

stedolan:
The function type:
```
a -> b @ local -> c -> d
```
actually means:
```
a -> (b @ local -> (c -> d) @ local)
```
That is, after the first local argument `b`, any intermediate closures created by partial application might be on the stack.

Using `caml_callbackN` with more than three arguments internally groups arguments in threes and does partial application of them. If some of the arguments are local, this means that some of the intermediate closures may be stack allocated.

When calling a function that may return a stack-allocated result, you need to use `caml_apply3L` instead of `caml_apply3` (the latter resets the local stack after the call, which is not correct if you're expecting a local value). This wasn't happening, causing crashes.

The fix is in three commits:

  1. Add a test (repro case by @TheNumbat, slightly tweaked)
  2. Fix the bug by using `caml_apply(2/3)L` instead of `caml_apply(2/3)` in callbacks. However, this introduces a leak: the intermediate closures now might never be popped off if someone calls `caml_callbackN` in a loop.
  3. Fix the leak introduced above, by resetting the stack around each call to `caml_callback*`.

We end up with the right behaviour, where the local stack is reset once per call to `caml_callbackN`, even if that does many `caml_apply3L` calls internally.


testsuite/tests/callback/test_local_bug_.c
```diff
@@ -0,0 +1,16 @@
+#include <caml/memory.h>
+#include <caml/callback.h>
+
+value local_args_repro_call(value fn) {
+  CAMLparam1(fn);
+
+  int nargs = 7;
```
NickBarnes: I'd prefer a `#define` here, so that it can be used on the next line (in the absence of VLA).

NickBarnes (approved):
This is OK.

mshinwell:
@stedolan @NickBarnes could one of you please fix the above and merge this?

PR #4246, opened 2025-06-30
Refactor merge constraints on signatures (backport ocaml/ocaml#13911)

rtjoa:
Backports https://github.com/ocaml/ocaml/pull/13911/


ccasin (commented):
I've read this and didn't spot any issues with the backport.  But, because it's a large refactor of a performance-sensitive bit of the typechecker, I think it would be prudent to do some performance testing on our internal code base before approving/merging.  I'll message you via internal channels about that.

ccasin (approved):
Performance numbers look good - appears to be in the noise, possibly a tiny improvement.

Unfortunately looks like there is a conflict.

PR #4247, opened 2025-06-30
Vec256 array load/store prims

TheNumbat:
Adds all types of array load/store primitives for 256 and 512-bit vectors. 

- The lambda primitives have been renamed, as they are now also parameterized by the vector width.
- Adds a 256 and 512 bit accessor width in flambda2. Regenerates the fl2 parser.

Translates the arrays tests for 128-bit to 256-bit (most of the diff is tests). The 512-bit primitives are still unimplemented and untested (they're rejected in the backend).


lambda/lambda.ml
```diff
@@ -244,37 +244,37 @@ type primitive =
...
       mode : locality_mode; boxed : bool }
   | Pstring_load_f32 of { unsafe : bool; index_kind : array_index_kind;
       mode : locality_mode; boxed : bool }
   | Pstring_load_64 of { unsafe : bool; index_kind : array_index_kind;
       mode : locality_mode; boxed : bool }
-  | Pstring_load_128 of
-      { unsafe : bool; index_kind : array_index_kind;
-      mode : locality_mode; boxed: bool }
+  | Pstring_load_vec of
+      { size : boxed_vector; unsafe : bool; index_kind : array_index_kind;
```
rtjoa, line 256: Nit: It's a bit odd to me that `boxed_vector` is used for the size, as these primitives can be boxed or unboxed.

What do you think about replacing both `boxed_vector` and `unboxed_vector` with something like the below, to make things a bit simpler/clearer?
```ocaml
type vector = Vec128 | Vec256 | Vec512
```

This is similar to how things were before https://github.com/oxcaml/oxcaml/pull/3360, which points out that a joined type only makes sense if there will be both unboxed/boxed versions of all vectors.

Feel free to disagree or decide this isn't worth the diff to change.

lambda/printlambda.ml
```diff
@@ -747,9 +747,10 @@ let primitive ppf = function
      fprintf ppf "string.%sget64%s%s[indexed by %a]"
        (if unsafe then "unsafe_" else "") (if boxed then "" else "#")
        (locality_kind mode) array_index_kind index_kind
-  | Pstring_load_128 {unsafe; index_kind; mode; boxed} ->
-     fprintf ppf "string.%sunaligned_get128%s%s[indexed by %a]"
-       (if unsafe then "unsafe_" else "") (if boxed then "" else "#")
+  | Pstring_load_vec {size; unsafe; index_kind; mode; boxed} ->
+     fprintf ppf "string.%sunaligned_get%s%s%s[indexed by %a]"
```
rtjoa, line 756: Nit:

This is another place where I feel a type for vector width that is not necessarily boxed or unboxed could be appropriate; this prints `getvec128` where it previously printed `get128` (although I think this change is inconsequential).

(If there was a function from vec size to number-as-a-string, then this would also mean that `vec_sizes` in `translprim.ml` could be a `vector list` instead of a `(boxed_vector * string) list`)

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
```diff
@@ -1009,7 +1046,7 @@ let check_array_vector_access ~dbg ~size_int ~array array_kind ~index ~vec_kind
     ~dbg
 
 let array_like_load_vec ~dbg ~size_int ~unsafe ~mode ~boxed ~current_region
-    ~(vec_kind : Vector_types.Kind.t) ~index_kind array_kind array index =
+    ~(vec_kind : Vector_types.Kind.t) array_kind array ~index_kind index =
```
rtjoa, line 1049: Just curious, why was this argument reordered?

oxcaml/tests/simd/arrays256.ml
```diff
@@ -0,0 +1,1591 @@
...
+
+  let set_unaligned_unsafe first second third fourth offset =
+    let set = int8x32_of_int64s first second third fourth in
+    set_int8x32_unaligned_unsafe data offset set;
+    let v = get_int8x32_unaligned data offset in
+    eq first second third fourth (int8x32_first_int64 v) (int8x32_second_int64 v) (int8x32_third_int64 v) (int8x32_fourth_int64 v);
+  ;;
+
+  let () =
+    set_unaligned 0x1010101010101010L 0x1010101010101010L 0x1010101010101010L 0x1010101010101010L 0;
```
rtjoa, line 112: Would this test be more robust if these values weren't all the same?

lambda/lambda.ml
TheNumbat: Agreed this is odd, but I'd rather keep it consistent with floats/ints. Maybe we should unify both the float and vector types, but not in this pr.

lambda/printlambda.ml
TheNumbat: Added a `vector_width` function just for printing

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
TheNumbat: Just seemed like it should precede `index`

oxcaml/tests/simd/arrays256.ml
TheNumbat: It would, but it uses random args afterward, which seems sufficient

PR #4249, opened 2025-07-01
arm64: fix reinterpret cast for Float/Float32

gretay-js:
Use FMOV instead of MOV. Fixes assembler error. 

Draft for the CI first, and still needs a test.

### DIFF
--- a/backend/arm64/emit.ml
+++ b/backend/arm64/emit.ml
@@ -1552,13 +1552,13 @@ let emit_reinterpret_cast (cast : Cmm.reinterpret_cast) i =
     then (
       DSL.check_reg Float src;
       DSL.check_reg Float32 dst;
-      DSL.ins I.MOV [| DSL.emit_reg_d dst; DSL.emit_reg_d src |])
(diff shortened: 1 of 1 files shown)


gretay-js:
I haven't managed to make a test for this without functions in PR #4247 or adding intrinsics to expose the Float64 <-> Float32 reinterpret cast. The only place we generate such a cast is in unboxed float32 array of odd size. We should probably merge this bug fix PR without a test and then add the intrinsics in a separate PR. @TheNumbat what do you think?

```
module _ = struct
  external unboxed_float32_array_get_float32x4 : float32# array -> int32# -> float32x4 = "%caml_unboxed_float32_array_get128_indexed_by_int32#"

  let unboxed_float32_array_get_float32x4 arr i = unboxed_float32_array_get_float32x4 arr (Stdlib_upstream_compatible.Int32_u.of_int i)

  let test () =
    let a = [|#0.0s;#0.0s;#0.0s|] in
    let i = 0 in
    let _ = unboxed_float32_array_get_float32x4 a i in
    ()

  let () =
    test () ;
    test () 
end
```

PR #4250, opened 2025-07-01
Even more moves of symbols initializations in to_cmm

Gbury:
This builds upon #4238  and goes further to move symbol initializations, where the field value is a cmm variable `v`, up right after the binding for `v`. #4238 moves symbol inits up, but cannot cross points where the environment is flushed. This PR propagates symbol initializations (i.e. the map from cmm variables to cmm expressions that initializes symbols) up when rebuilding cmm terms, similarly to what is done for the free vars of generated cmm expressions, which allows symbols inits to cross environment flushes.

We can look at a more complex version of the example from #4238 , where we inline some functions with non-trivial control flow (i.e. functions whose body contain continuations that will not be inlined by to_cmm):
```ocaml
let[@inline never] foo () = 42

let[@inline always] bar () =
  if Sys.opaque_identity true then 13 else 42

let[@inline always] loop () =
  let r = ref 0 in
  for i = 0 to Sys.opaque_identity 5 do
    r := !r + i
  done;
  !r

let field1 = foo ()
let field2 = bar ()
let field3 = loop ()
```

With current main, we get the following cmm code for the module init code:
```
(function no_cse reduce_code_size linscan camlFoo__entry ()
 (catch
   (let field1/422 (app{foo.ml:14,13-19} G:"camlFoo__foo_0_3_code" 1 int)     [ call to `foo` (not inlined)
     (catch                                                                   ┌
       (let Popaque/432 (opaque 3)                                            │ inlined body of `bar`
         (if (== 1 Popaque/432) (exit 8 85) (exit 8 27)))                     │
     with(8 field2/423: int)                                                  └
       (catch                                                                 ┌
         (let for_stop/425 (opaque 11)                                        │ inlined body of `loop`
           (if (<= 1 for_stop/425)                                            │
             (catch rec (exit 7 1 1) with(7 i/427: int                        │
               r_365_unboxed0/428: int)                                       │
               (let Paddint/429 (+ (+ r_365_unboxed0/428 i/427) -1)           │
                 (if (!= i/427 for_stop/425) (exit 7 (+ i/427 2) Paddint/429) │
                   (exit 6 Paddint/429))))                                    │
             (exit 6 1)))                                                     │
       with(6 field3/424: int)                                                └
         (extcall "caml_initialize" (+a G:"camlFoo" 24) field1/422 ->unit) 1  [ init with the result of `foo`
         (extcall "caml_initialize" (+a G:"camlFoo" 32) field2/423 ->unit) 1  [ init with the result of `bar`
         (extcall "caml_initialize" (+a G:"camlFoo" 40) field3/424 ->unit) 1  [ init with the result of `loop`
         (exit 4 G:"camlFoo"))))
 with(4 *ret*/408: val) 1))
```
where we once again get the problem that the lifetime of `field1/422` and  `field2/423` are needlessly long.

With #4238, we get the same output, since the symbol inits cannot cross above the flush at the start of the cmm handler `6`.

With this PR we get
```
(function no_cse reduce_code_size linscan camlFoo__entry ()
 (catch
   (let field1/422 (app{foo.ml:14,13-19} G:"camlFoo__foo_0_3_code" 1 int)       [ call to `foo` (not inlined)
     (extcall "caml_initialize" (+a G:"camlFoo" 24) field1/422 ->unit)          [ init with the result of `foo`
     (catch                                                                     ┌
       (let Popaque/432 (opaque 3)                                              │ inlined body of `bar`
         (if (== 1 Popaque/432) (exit 8 85) (exit 8 27)))                       │
     with(8 field2/423: int)                                                    └
       (extcall "caml_initialize" (+a G:"camlFoo" 32) field2/423 ->unit)        [ init with the result of `bar`
       (catch                                                                   ┌
         (let for_stop/425 (opaque 11)                                          │ inlined body of `loop`
           (if (<= 1 for_stop/425)                                              │
             (catch rec (exit 7 1 1) with(7 i/427: int r/428: int)              │
               (let Paddint/429 (+ (+ r/428 i/427) -1)                          │
                 (if (!= i/427 for_stop/425) (exit 7 (+ i/427 2) Paddint/429)   │
                   (exit 6 Paddint/429))))                                      │
             (exit 6 1)))                                                       │
       with(6 field3/424: int)                                                  └
         (extcall "caml_initialize" (+a G:"camlFoo" 40) field3/424 ->unit)      [ init with the result of `loop`
         (exit 4 G:"camlFoo"))))
 with(4 *ret*/408: val) 1))
```
Resulting in an optimal placement of symbol initializations, and the shorted lifetimes for the values used in the initialization (this also applies in classic mode).

Note that while in this kind of simple examples, this PR achieves the best we can, it will not always apply: if the value used for the symbol initialization is not a cmm variable (but e.g. a substituted primitive), then nothing will change compared to main (though I'd expect a "sea of nodes" approach would be able to optimize this case), similarly if the variable used for symbol init is aliased, the process implemented in this PR may not move it up past the aliasing point (it may depends on some of the choices made by to_cmm).

Finally, while most of the changes may seem simple and mechanical, actually great care must be taken to correctly thread the symbol inits upwards, so all changes should be reviewed carefully to ensure that no symbol inits are dropped, and that they are propagated to the right places.

### DIFF
--- a/middle_end/flambda2/to_cmm/to_cmm_expr.ml
+++ b/middle_end/flambda2/to_cmm/to_cmm_expr.ml
@@ -535,7 +535,7 @@ let translate_raise ~dbg_with_inlined:dbg env res apply exn_handler args =
     let free_vars = Backend_var.Set.union exn_free_vars extra_free_vars in
     let wrap, _, res = Env.flush_delayed_lets ~mode:Branching_point env res in
     let cmm = C.raise_prim raise_kind exn ~extra_args:extra dbg in
-    let cmm, free_vars, symbol_inits = wrap cmm free_vars Env.empty_symbol_inits in
+    let cmm, free_vars, symbol_inits = wrap cmm free_vars Env.Symbol_inits.empty in
     cmm, free_vars, symbol_inits, res
   | [] ->
     Misc.fatal_errorf "Exception continuation %a has no arguments:@ \n%a"
@@ -558,7 +558,7 @@ let translate_jump_to_continuation ~dbg_with_inlined:dbg env res apply types
     let args = C.remove_skipped_args args types in
     let args, free_vars, env, res, _ = C.simple_list ~dbg env res args in
     let wrap, _, res = Env.flush_delayed_lets ~mode:Branching_point env res in
-    let cmm, free_vars, symbol_inits = wrap (C.cexit cont args trap_actions) free_vars Env.empty_symbol_inits in
+    let cmm, free_vars, symbol_inits = wrap (C.cexit cont args trap_actions) free_vars Env.Symbol_inits.empty in
     cmm, free_vars, symbol_inits, res
   else
     Misc.fatal_errorf "Types (%a) do not match arguments of@ %a"
@@ -579,12 +579,12 @@ let translate_jump_to_return_continuation ~dbg_with_inlined:dbg env res apply
     let wrap, _, res = Env.flush_delayed_lets ~mode:Branching_point env res in
     match Apply_cont.trap_action apply with
     | None ->
-      let cmm, free_vars, symbol_inits = wrap return_value free_vars Env.empty_symbol_inits in
+      let cmm, free_vars, symbol_inits = wrap return_value free_vars Env.Symbol_inits.empty in
       cmm, free_vars, symbol_inits, res
     | Some (Pop { exn_handler; _ }) ->
       let cont = Env.get_cmm_continuation env exn_handler in
       let cmm, free_vars, symbol_inits =
-        wrap (C.trap_return return_value [Cmm.Pop cont]) free_vars Env.empty_symbol_inits
+        wrap (C.trap_return return_value [Cmm.Pop cont]) free_vars Env.Symbol_inits.empty
       in
       cmm, free_vars, symbol_inits, res
     | Some (Push _) ->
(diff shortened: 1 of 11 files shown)


middle_end/flambda2/to_cmm/to_cmm_expr.ml
```diff
@@ -1170,22 +1220,29 @@ and switch env res switch =
...
+        (0, else_, else_free_vars, else_inits, else_dbg) ) ->
       let free_vars =
         Backend_var.Set.union scrutinee_free_vars
           (Backend_var.Set.union else_free_vars then_free_vars)
       in
-      let cmm, free_vars =
-        wrap (C.ite ~dbg scrutinee ~then_dbg ~then_ ~else_dbg ~else_) free_vars
+      (* CR gbury: technically, these symbol inits should be empty since neither
+         of the branches should be at top-level, but it should not hurt to do
+         this. *)
```
Ekdohibs: Here and in the other cases, I'd prefer if this caused an error, to prevent from moving code that was under a branch outside it (even if it should not happen).

Gbury: Oh so apparently, having a non-empty symbol inits in these cases does seem to occur frequently... I'm investigating.

Gbury: And indeed as I was suspecting, my comment was actually false (and the code correct). Indeed consider the following example:
```ocaml
let tuple =
  let x = Sys.opaque_identity 13 in
  if Sys.opaque_identity false then raise Exit;
  let y = Sys.opaque_identity 42 in
  (x, y)
```

The tuple allocation is correctly lifted, and as it happens, it occurs in a continuation handler whose only call is as a branch in the switch for the if-then-else; `to_cmm` will happily inline that continuation, and end up with symbol inits coming from one of the branch of a switch. This is normal and expected I'd say.

I adjusted the comment to make it clear that it is expected that this can happen (although in theory only one branch at most should have such symbol inits, I'd argue it would complexify the code needlessly to check this).

Ekdohibs (approved):

PR #4253, opened 2025-07-01
Type Shapes and Type Declaration Shapes for Debugger

spiessimon:
This PR is part of the chain of PRs for propagating debugging information through the compiler (e.g., see also #3967). It captures the typing information required for emitting DWARF information about the variables from the type system. More specifically, it captures *type shapes* (see `type_shape.ml`), simplified, abstract representations of type expressions and type declarations. These type shapes will---in subsequent PRs---be used to emit DWARF type entries into the resulting assembly. (The code for emitting the DWARF type entries is not part of this PR.)

**Note 1:** The implementation of type shapes is only a first version. It currently omits unboxed types, and it is orthogonal to the current shape mechanism in the compiler. In subsequent PRs, it will move into the shape type from `shape.ml` and use the lookup machinery of those shapes to enable emitting DWARF information for type declarations spread across files.

**Note 2:** We currently compute type declaration shapes *for every type declaration* and type expression shapes *for every binding variable*. The performance effect of this computation has not yet been measured.

The code in this PR is based on the type shapes and the capturing in #1852.

### DIFF
--- a/lambda/translcore.ml
+++ b/lambda/translcore.ml
@@ -1590,34 +1590,31 @@ and transl_tupled_function
         in
         let region = region || not (may_allocate_in_region body) in
         add_type_shapes_of_cases arg_sort cases;
-        (* CR sspies: Unsure whether we need to add this here. *)
+        (* CR sspies: Unsure whether this is the right place to grab the type
+           expressions. *)
         Some
           ((Tupled, tparams, return_layout, region, return_mode), body)
     with Matching.Cannot_flatten -> None
       end
   | _ -> None
 
-(* CR sspies: I think for the three functions below to work correctly, we must maintain
-   the following invariant:
+(* For the functions [add_type_shape_of_cases], [add_type_shapes_of_params], and
+   [add_type_shapes_of_patterns] to be correct, we must ensure that at the type
+   tree level, a [debug_uid] is never associated with more than one type
+   expression, because the type expressions determine the debug information we
+   emit for the bound variable associated with the debug uid.
 
-      At the typed_tree level, each debug uid is associated with at most one variable.
-
-   The reason is that here, we associate debug UIDs with type expressions from the
-   variable declaration. For this to make sense, there should only be one type expression
-   for a UID. For example, for:
+   For example, for:
 
       let f (x: int list) = x
 
-   The functions below will associate the UID of [x] with [int list] as the type
-   expression. We will the separately associate the debugging UID of [x] with the type
-   declaration of [list], such that in the backend we have both the type expression
-   (here [int list]) and the type declaration (here [type 'a list = ...]) available for
-   the variable [x].
+   the functions below will associate the UID of [x] with [int list] as the type
+   expression.
 *)
 
-(** [add_type_shapes_of_cases] iterates through a given list of cases and associates
-    for each case, the debugging UID of the variable with the type expression of
-    the variable and its sort. *)
+(** [add_type_shapes_of_cases] iterates through a given list of cases and
+    associates for each case, the debugging UID of the variable with the type
+    expression of the variable and its sort. *)
 and add_type_shapes_of_cases sort cases =
   let add_case (case : Typedtree.value Typedtree.case) =
     let var_list = Typedtree.pat_bound_idents_full sort case.c_lhs in
@@ -1628,12 +1625,15 @@ and add_type_shapes_of_cases sort cases =
   in
   List.iter add_case cases
 
-(** [add_type_shapes_of_params] iterates through the variables in a function parameter
-    and, for each variable, associates the debugging UID of the variable with the type
-    expression of the variable. *)
+(** [add_type_shapes_of_params] iterates through the variables in a function
+    parameter and, for each variable, associates the debugging UID of the
+    variable with the type expression of the variable. *)
 and add_type_shapes_of_params params =
     let add_param (param : Typedtree.function_param) =
-      let pattern = match param.fp_kind with Tparam_pat p -> p | Tparam_optional_default (p, _, _) -> p in
+      let pattern = match param.fp_kind with
+                    | Tparam_pat p -> p
+                    | Tparam_optional_default (p, _, _) -> p
+      in
       let sort = Jkind.Sort.default_for_transl_and_get param.fp_sort in
       let var_list = Typedtree.pat_bound_idents_full sort pattern in
       List.iter (fun (_ident, _loc, type_expr, var_uid, var_sort) ->
@@ -1643,9 +1643,9 @@ and add_type_shapes_of_params params =
     in
     List.iter add_param params
 
-(** [add_type_shapes_of_patterns] iterates through the variables in a value binding
-  and, for each variable, associates the debugging UID of the variable with the type
-  expression of the variable. *)
+(** [add_type_shapes_of_patterns] iterates through the variables in a value
+    binding and, for each variable, associates the debugging UID of the variable
+    with the type expression of the variable. *)
 and add_type_shapes_of_patterns patterns =
   let add_case (value_binding : Typedtree.value_binding) =
     let sort = Jkind.Sort.default_for_transl_and_get value_binding.vb_sort in
--- a/lambda/translcore.ml
+++ b/lambda/translcore.ml
@@ -1589,11 +1589,71 @@ and transl_tupled_function
             (transl_tupled_cases ~scopes return_sort pats_expr_list) partial
         in
         let region = region || not (may_allocate_in_region body) in
+        add_type_shapes_of_cases cases;
+        (* CR sspies: Unsure whether we need to add this here. *)
         Some
           ((Tupled, tparams, return_layout, region, return_mode), body)
     with Matching.Cannot_flatten -> None
       end
   | _ -> None
+
+(* CR sspies: I think for the three functions below to work correctly, we must maintain
+   the following invariant:
+
+      At the typed_tree level, each debug uid is associated with at most one variable.
+
+   The reason is that here, we associate debug UIDs with type expressions from the
+   variable declaration. For this to make sense, there should only be one type expression
+   for a UID. For example, for:
+
+      let f (x: int list) = x
+
+   The functions below will associate the UID of [x] with [int list] as the type
+   expression. We will the separately associate the debugging UID of [x] with the type
+   declaration of [list], such that in the backend we have both the type expression
+   (here [int list]) and the type declaration (here [type 'a list = ...]) available for
+   the variable [x].
+*)
+
+(** [add_type_shapes_of_cases] iterates through a given list of cases and associates
+    for each case, the debugging UID of the variable with the type expression of
+    the variable. *)
+and add_type_shapes_of_cases cases =
+  let add_case (case : Typedtree.value Typedtree.case) =
+    let var_list = Typedtree.pat_bound_idents_full Jkind.Sort.Const.value case.c_lhs in
+    List.iter (fun (_ident, _loc, type_expr, var_uid, _mode) ->
+      Type_shape.add_to_type_shapes var_uid type_expr
+        (Typedecl.uid_of_path ~env:case.c_lhs.pat_env))
+      var_list
+  in
+  List.iter add_case cases
+
+(** [add_type_shapes_of_params] iterates through the variables in a function parameter
+    and, for each variable, associates the debugging UID of the variable with the type
+    expression of the variable. *)
+and add_type_shapes_of_params params =
+    let add_param (param : Typedtree.function_param) =
+      let pattern = match param.fp_kind with Tparam_pat p -> p | Tparam_optional_default (p, _, _) -> p in
+      let var_list = Typedtree.pat_bound_idents_full Jkind.Sort.Const.value pattern in
+      List.iter (fun (_ident, _loc, type_expr, var_uid, _mode) ->
+        Type_shape.add_to_type_shapes var_uid type_expr
+          (Typedecl.uid_of_path ~env:pattern.pat_env))
+        var_list
+    in
+    List.iter add_param params
+
+(** [add_type_shapes_of_patterns] iterates through the variables in a value binding
+  and, for each variable, associates the debugging UID of the variable with the type
+  expression of the variable. *)
+and add_type_shapes_of_patterns patterns =
+  let add_case (value_binding : Typedtree.value_binding) =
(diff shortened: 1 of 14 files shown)


lambda/translcore.ml
```diff
@@ -1589,12 +1589,74 @@ and transl_tupled_function
...
+(** [add_type_shapes_of_cases] iterates through a given list of cases and
+    associates for each case, the debugging UID of the variable with the type
+    expression of the variable and its sort. *)
+and add_type_shapes_of_cases sort cases =
+  let add_case (case : Typedtree.value Typedtree.case) =
+    let var_list = Typedtree.pat_bound_idents_full sort case.c_lhs in
+    List.iter (fun (_ident, _loc, type_expr, var_uid, var_sort) ->
+      Type_shape.add_to_type_shapes var_uid type_expr var_sort
+        (Env.find_uid_of_path case.c_lhs.pat_env))
+      var_list
```
lukemaurer: These lines are largely the same in all three `add_` functions. Can we pull it out into `add_type_shapes_of_pattern`?

typing/env.ml
```diff
@@ -1660,6 +1660,14 @@ let find_shape env (ns : Shape.Sig_component_kind.t) id =
   | Class_type ->
       (IdTbl.find_same id env.cltypes).cltda_shape
 
+let find_uid_of_path env path =
+  (* We currently only support lookup up debugging uids in the current
```
lukemaurer:
```suggestion
  (* We currently only support looking up debugging uids in the current
```

typing/env.ml
```diff
@@ -2929,10 +2937,25 @@ let save_signature_with_imports ~alerts sg modname cu cmi imports =
...
+     environment in a table. *)
+  let added_types = Ident.Tbl.create 16 in
+  let add_type_and_remember_decl (type_ident : Ident.t) decl env =
+    Ident.Tbl.add added_types type_ident decl;
+    add_type type_ident decl env ~check:false
+  in
+  let initial_env = Predef.build_initial_env
+                      add_type_and_remember_decl
+                      (add_extension ~check:false ~rebind:false)
+                      empty
```
lukemaurer: Not that the formatting in the compiler is especially consistent, but I do think this would be better with a line break after the equal sign.

typing/type_shape.ml
```diff
@@ -0,0 +1,401 @@
...
+
+  (* Similarly to [value_kind], we track a set of visited types to avoid cycles
+     in the lookup and we, additionally, carry a maximal depth for the recursion.
+     We allow a deeper bound than [value_kind]. *)
+  (* CR sspies: Consider additionally adding a max size for the set of visited types.
+     Also consider reverting to the original value kind depth limit (although 2
+     seems low). *)
+  let rec of_type_expr_go ~visited ~depth (expr : Types.type_expr) uid_of_path =
+    let[@inline] cannot_proceed () =
+      Numbers.Int.Set.mem (Types.get_id expr) visited || depth >= 10
```
lukemaurer: This is picky, but I think it would be slightly more intuitive to compare using `>` here and start the depth at 0 rather than -1.

typing/type_shape.ml
```diff
@@ -0,0 +1,401 @@
...
+    else
+      let visited = Numbers.Int.Set.add (Types.get_id expr) visited in
+      let depth = depth + 1 in
+      let desc = Types.get_desc expr in
+      let map_expr_list (exprs : Types.type_expr list) =
+        List.map
+          (fun expr -> of_type_expr_go ~depth ~visited expr uid_of_path)
+          exprs
+      in
+      match[@warning "-4"] desc with
```
lukemaurer: I'd rather we use the name of the warning here. (I can guess, of course, but the less time I spend guessing the better.)

typing/type_shape.ml
```diff
@@ -0,0 +1,401 @@
...
+    let[@inline] cannot_proceed () =
+      Numbers.Int.Set.mem (Types.get_id expr) visited || depth >= 10
+    in
+    if cannot_proceed ()
+    then Ts_other
+    else
+      let visited = Numbers.Int.Set.add (Types.get_id expr) visited in
+      let depth = depth + 1 in
+      let desc = Types.get_desc expr in
+      let map_expr_list (exprs : Types.type_expr list) =
```
lukemaurer: This would be better called `of_type_expr_list` or something: `map_expr_list` sounds like a general `map` operation over `expr` lists

typing/type_shape.ml
```diff
@@ -0,0 +1,401 @@
...
+          | Some uid -> Ts_constr ((uid, path), map_expr_list constrs)
+          | None -> Ts_other))
+      | Ttuple exprs -> Ts_tuple (map_expr_list (List.map snd exprs))
+      | Tvar { name; _ } -> Ts_var name
+      | Tpoly (type_expr, []) ->
+        of_type_expr_go ~depth ~visited type_expr uid_of_path
+      | _ -> Ts_other
+
+  let of_type_expr (expr : Types.type_expr) uid_of_path =
+    of_type_expr_go ~visited:Numbers.Int.Set.empty ~depth:(-1) expr uid_of_path
```
lukemaurer: (see above: depth should start at 0)

typing/type_shape.ml
```diff
@@ -0,0 +1,401 @@
...
+           ~pp_sep:(fun ppf () -> Format.pp_print_string ppf ", ")
+           print)
+        shapes
+    | Ts_constr ((uid, path), shapes) ->
+      Format.fprintf ppf "Ts_constr uid=%a path=%a (%a)" Uid.print uid
+        Path.print path
+        (Format.pp_print_list
+           ~pp_sep:(fun ppf () -> Format.pp_print_string ppf ", ")
+           print)
+        shapes
```
lukemaurer, line 114: Worth being thoughtful about how we're going to wrap long shapes here (the pretty-printing API is truly awful, though; I can help with this)

typing/type_shape.ml
```diff
@@ -0,0 +1,401 @@
...
+        (fun ppf opt -> Format.pp_print_option Format.pp_print_string ppf opt)
+        name
+    | Ts_other -> Format.fprintf ppf "Ts_other"
+
+  (* CR sspies: This is a hacky "solution" to do type variable substitution in
+     type expression shapes. In subsequent PRs, this code should be changed to
+     use the shape mechanism instead. *)
+  let rec replace_tvar t ~(pairs : (t * t) list) =
+    match
+      List.filter_map
```
lukemaurer: Can use `List.find_map` for this. (Or `List.assoc` but that implicitly uses polymorphic equality so it's best avoided.)

typing/type_shape.ml
```diff
@@ -0,0 +1,401 @@
...
+    | new_type :: _ -> new_type
+    | [] -> (
+      match t with
+      | Ts_constr (uid, shape_list) ->
+        Ts_constr (uid, List.map (replace_tvar ~pairs) shape_list)
+      | Ts_tuple shape_list ->
+        Ts_tuple (List.map (replace_tvar ~pairs) shape_list)
+      | Ts_var name -> Ts_var name
+      | Ts_predef (predef, shape_list) -> Ts_predef (predef, shape_list)
+      | Ts_other -> Ts_other)
```
lukemaurer: These can be a single or-pattern that returns `t`.

typing/type_shape.ml
```diff
@@ -0,0 +1,401 @@
...
+  end)
+end
+
+module Type_decl_shape = struct
+  type 'a complex_constructor =
+    { name : string;
+      args : 'a complex_constructor_arguments list
+    }
+
+  and 'a complex_constructor_arguments =
```
lukemaurer: This represents a single argument, so the name should be singular (i.e., `complex_constructor_argument`).

typing/type_shape.ml
```diff
@@ -0,0 +1,401 @@
...
+
+    let compare = Stdlib.compare
+
+    let print = print
+
+    let hash = Hashtbl.hash
+
+    let equal (x : t) y = x = y
+
+    let output _oc _t = Misc.fatal_error "unimplemented"
```
lukemaurer: You can use `Misc.output_of_print` to define this from `print` (it's almost but not quite trivial).

typing/type_shape.ml
```diff
@@ -0,0 +1,401 @@
...
+  and 'a complex_constructor_arguments =
+    { field_name : string option;
+      field_value : 'a
+    }
+
+  type tds =
+    | Tds_variant of
+        { simple_constructors : string list;
+          complex_constructors : Type_shape.t complex_constructor list
+        }
```
lukemaurer: Is it necessary to distinguish the no-argument case like this? Why not just have a simple constructor be a constructor whose list of arguments is empty?

typing/type_shape.ml
```diff
@@ -0,0 +1,401 @@
...
+          })
+        list
+
+  let is_empty_constructor_list (cstr_args : Types.constructor_declaration) =
+    let length =
+      match cstr_args.cd_args with
+      | Cstr_tuple list -> List.length list
+      | Cstr_record list -> List.length list
+    in
+    length = 0
```
lukemaurer: Since you're already matching, you can just match against `Cstr_tuple []` for the `true` case. (Records aren't allowed to be empty so you don't need to handle the `Cstr_record` case closely.)

typing/type_shape.ml
```diff
@@ -0,0 +1,401 @@
...
+        | Type_variant (cstr_list, _variant_repr, _unsafe_mode_crossing) ->
+          let simple_constructors, complex_constructors =
+            List.partition_map
+              (fun (cstr : Types.constructor_declaration) ->
+                let name = Ident.name cstr.cd_id in
+                match is_empty_constructor_list cstr with
+                | true -> Left name
+                | false ->
+                  Right
+                    { name; args = get_variant_constructors cstr uid_of_path })
```
lukemaurer: Hmm. Does this mean the debugger will display
```
type t = A of int * int
```
and
```
type t = A of { x : int; y : int }
```
the same? I don't think that'll sit well with users.

typing/type_shape.ml
```diff
@@ -0,0 +1,401 @@
...
+  let print_tds ppf = function
+    | Tds_variant { simple_constructors; complex_constructors } ->
+      Format.fprintf ppf
+        "Tds_variant simple_constructors=%a complex_constructors=%a"
+        (Format.pp_print_list ~pp_sep:Format.pp_print_space
+           Format.pp_print_string)
+        simple_constructors
+        (Format.pp_print_list ~pp_sep:Format.pp_print_space
+           (print_complex_constructor Type_shape.print))
+        complex_constructors
```
lukemaurer, line 303: We're definitely going to want to think about boxes here

typing/type_shape.ml
```diff
@@ -0,0 +1,401 @@
...
+  let type_shape = Type_shape.of_type_expr type_expr uid_of_path in
+  Uid.Tbl.add all_type_shapes var_uid { type_shape; type_sort = sort }
+
+let tuple_to_string (strings : string list) =
+  match strings with
+  | [] -> ""
+  | hd :: [] -> hd
+  | _ :: _ :: _ -> "(" ^ String.concat " * " strings ^ ")"
+
+let shapes_to_string (strings : string list) =
```
lukemaurer: This should be called `type_arg_list_to_string` since it's very specifically doing that job (otherwise the parens and trailing space make no sense).

lambda/translcore.ml
spiessimon: Great idea! I pulled them out into a separate function.

typing/type_shape.ml
spiessimon: That's a good point. I've been going back and forth on the type shape constructors across different PRs (e.g., I'm thinking about moving them into the `shape.ml` type itself, which removes constructors like `Ts_var`). I think I'd prefer figuring out the pretty printing once the type shapes/type declaration shapes are in a more stable state. I left comments in the file so we don't forget.

typing/type_shape.ml
spiessimon: While we're at this point. Knowing more about the type system now, I think we should use physical equality `==` instead of polymorphic equality here to detect the type parameters inside of a type declaration. Is that correct? (I changed it as part of the review commit.)

typing/type_shape.ml
spiessimon: I think consolidating these two is a good idea. As this representation is currently also used in several branches down the line, I think I would like to keep it like this for now and then improve it once more of the code is merged. I left a comment.

typing/type_shape.ml
spiessimon: Nope, not quite the `complex_constructor_argument` type carries an optional label for the field name. For tuples, this is none (for all fields) and for records this is some. The DWARF output/LLDB will factor in whether there are field names or not.

typing/type_shape.ml
spiessimon: Same as above. I agree, but would like to hold off on that for now.

spiessimon:
I made a pass and, I think, addressed all of your comments. For the minor ones, I just implemented the change and didn't separately comment above. Please resolve as appropriate.

typing/type_shape.ml
lukemaurer: Sounds good. Yeah, the issue's only really forced once these are appearing in `objinfo` output (though I would encourage doing that eagerly once type shapes are in .cms files)

typing/type_shape.ml
lukemaurer: Hmm. It's not crazy, but it would mean being very careful of any transformations that could destroy sharing. I would think if this is a temporary solution then probably it's better off being correct but a bit slow (i.e., use `=`)? Or is the sharing easier to guarantee than I know?

typing/type_shape.ml
lukemaurer: Oh heh, I'd just been looking at that code too :laughing: Actually, can we rename `get_variant_constructors` to `get_constructor_args` since that's what it's operating on?

lukemaurer (approved):
Looks good with one or two small changes (feel free to merge using `==` if my worries about sharing are unwarranted)

typing/type_shape.ml
spiessimon: Good point. I renamed it.

typing/type_shape.ml
spiessimon: Ah, I confused myself there with type expressions. I changed it back to `=`. Thanks!

PR #4257, opened 2025-07-02
Add -disable-builtin-check flag

gretay-js:
Currently, compilations fails if an `external` annotated with `[@@builtin]` is not recongnized.  This is a "safety" mechanism to prevent silent performance regressions. This PR adds a flag that provides a way to disable this behavior and fall-back to the corresponding C stub. This is useful for testing new targets.

- Add a flag and use it in `Cfg_selection`.
- Improve user error message about `[@@builtin]` not recognized in the default case.
- Add a test.
- Sneak in a fix to an unrelated typo in Oxcaml_args


driver/oxcaml_args.ml
```diff
@@ -210,6 +210,11 @@ let mk_function_layout f =
   (Printf.sprintf " Order of functions in the generated assembly (default: %s)"
      default)
 
+let mk_disable_builtin_check f =
+  "-disable-builtin-check", Arg.Unit f,
+  " If an external annotated with [@builtin] is not recognized, \
```
xclerc:
```suggestion
  " If an external annotated with [@@builtin] is not recognized, \
```

PR #4258, opened 2025-07-02
Unboxed indexing for array vector load/store 

TheNumbat:
Updates the vector load/store array primitives to support indexing by unboxed integers.
Converts the hardcoded matches to a similar pre-generated map as used for string-like load/store primitives.
Now runs the tests in `simd/arrays{_u}.ml` for all kinds of indexes.

(The formatting in lambda is a bit off, but it's replaced in the following pr)


lambda/translprim.ml
```diff
@@ -351,6 +351,85 @@ let indexing_primitives =
...
+     (Printf.sprintf "%%caml_%s_get%s%s%s%s" array_type size_str unsafe_sigil
+        boxed_sigil index_kind_sigil,
+      fun ~mode ->
+        Primitive (load_prim ~unsafe ~index_kind ~mode ~boxed, 2));
+     (Printf.sprintf "%%caml_%s_set%s%s%s%s" array_type size_str unsafe_sigil
+        boxed_sigil index_kind_sigil,
+      fun ~mode:_ -> Primitive (set_prim ~unsafe ~index_kind ~boxed, 3))
+   ])
+  |> List.to_seq
+  |> fun seq -> String.Map.add_seq seq String.Map.empty
```
rtjoa: Nit (although I see isn't used for the stringlike indexing enumeration). This could also apply to `primitive.ml`.
```suggestion
  |> String.Map.of_seq
```

typing/primitive.ml
```diff
@@ -586,6 +586,62 @@ let prim_has_valid_reprs ~loc prim =
       |> List.to_seq
       |> fun seq -> String.Map.add_seq seq String.Map.empty
     in
+    let vector_array_indexing_primitives =
```
rtjoa, line 591: I think it could be useful to add a comment that this is coupled with the enumeration in `translprim.ml`, and a backpointer there as well. But I don't see such a comment for `stringlike_indexing_primitives`, so up to you.

PR #4261, opened 2025-07-03
Zero alloc: missing summary hint

gretay-js:
Improve error messages about missing allocation summaries when the zero_alloc check fails due to build-system heuristic that "hide" cmx files of some transitive dependencies.

The two new tests are identical except in one of them the ".cmx" file is deleted after compilation.

### DIFF
--- a/backend/zero_alloc_checker.ml
+++ b/backend/zero_alloc_checker.ml
@@ -49,18 +49,41 @@ module Witness = struct
         }
     | Widen
 
+  type hint =
+    | No_hint
+    | Missing_summary
+    | Conservative
+
   type t =
     { dbg : Debuginfo.t;
-      kind : kind
+      kind : kind;
+      hint : hint
     }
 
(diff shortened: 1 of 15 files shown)


backend/zero_alloc_checker.ml
```diff
@@ -49,18 +49,41 @@ module Witness = struct
...
+      let c = Stdlib.compare kind1 kind2 in
+      if c <> 0 then c else Stdlib.compare h1 h2
+
+  let print_hint ppf hint =
+    match hint with
+    | No_hint -> ()
+    | Missing_summary ->
+      Format.fprintf ppf
+        "@.Hint: Build artifacts for the library containing the callee are not \
+         available.@.Try adding the library as an explicity dependency.@."
```
ccasin:
```suggestion
         available.@.Try adding the library as an explicit dependency.@."
```

PR #4262, opened 2025-07-03
Rewrite patricia tree merge functions with `pattern_match_pair`

bclement-ocp:
This is a rewrite of both #4073 and #4183.

Instead of writing a `universal_merge` function in the style of Haskell's `Data.IntMap`, we instead use a `pattern_match_pair` function encapsulating the logic for pattern-matching on pairs of trees. This is both more flexible, and avoids leaving unneeded value slots around after simplification as with the `universal_merge` function.

Each of the function (`union`, etc.) is itself recursive, but the code remains small and readable:

```ocaml
  let rec union f t0 t1 =
    let iv = is_value_of t0 in
    pattern_match_pair_merge
      ~only_left:(fun t0 -> t0)
      ~only_right:(fun t1 -> t1)
      ~both_sides:(fun t0 t1 -> union f t0 t1)
      iv f t0 t1
```

For `union` specifically, I have carefully made sure that the generated code is identical to the current code in `main` (in fact, a tiny bit better, because we are able to avoid one allocation we failed to eliminate before).

This function includes both the `diff` function from #4073 (which is used in #4112) and the `update_many` function from #4183 (which will be used in a separate PR to optimize the quadratic use case in dataflow mentioned in #4183).


bclement-ocp:
Two additional notes I forgot in the PR description:

 - There is a separation between `pattern_match_pair` (which does the actual parallel pattern matching between two trees) and `pattern_match_pair_merge` (which provides an easy-to-use wrapper around `pattern_match_pair` for use cases that involve reconstructing a tree). This is because `pattern_match_pair` can be useful in order to implement functions that do not reconstruct a tree; for instance, we discussed with @Ekdohibs  a `diff_iterator` function that would implement `iterator (diff t0 t1)` more efficiently and could be useful in both to further improve performance of the new join and the datalog engine.

 - I did not rewrite the `inter` family of functions because the current implementation calls out to loopified `find` (or `mem`) and I couldn't figure out how to reproduce that without complexifying the `pattern_match_pair` API.

middle_end/flambda2/algorithms/patricia_tree.ml
```diff
@@ -449,48 +462,253 @@ end = struct
...
+    | None ->
+      pattern_match_pair t0 t1
+        ~leaf:(fun i d0 d1 ->
+          match combine i d0 d1 with
+          | None -> empty iv
+          | Some d -> (
+            match phys_eq_check_leaf_left ~orig_t:t0 ~orig_d:d0 d with
+            | Some t' -> t'
+            | None -> (
+              match phys_eq_check_leaf_right ~orig_t:t1 ~orig_d:d1 d with
```
Ekdohibs:
```suggestion
            match (phys_eq_check_leaf_left [@inlined hint]) ~orig_t:t0 ~orig_d:d0 d with
            | Some t' -> t'
            | None -> (
              match (phys_eq_check_leaf_right [@inlined hint]) ~orig_t:t1 ~orig_d:d1 d with
```

middle_end/flambda2/algorithms/patricia_tree.ml
```diff
@@ -449,48 +462,253 @@ end = struct
...
+    let iv = is_value_of t0 in
+    pattern_match_pair_merge_sharing
+      ~only_left:(fun t0 -> t0)
+      ~only_right:(fun _ -> empty iv)
+      ~both_sides:(fun t0 t1 -> diff_sharing f t0 t1)
+      iv f t0 t1
+
+  let rec diff_shared f t0 t1 =
+    (* Using [pattern_match_pair_merge_sharing] here is an optimisation, it is
+       not guaranteed by the interface of [diff_shared]. *)
```
Ekdohibs: The interface says this, which leads me to think that sharing with the first argument is guaranteed:
```ocaml
(** [diff_shared f m1 m2] is a version of [diff_sharing f m1 m2] that also
      exploits sharing of [m1] and [m2] to avoid calling [f] when possible,
      assuming that [f x x] always returns [None]. *)
```

bclement-ocp: Ah, good catch — I initially wanted it to be, but then realised that it was not really useful (we want sharing when we grow the maps, but `diff_shared` shrinks them). `diff_shared` is intended to be used when we expect sharing between `m1` and `m2`, and in this case `diff_shared f m1 m2 == m1` is meaningless. I will update the interface.

bclement-ocp: That said, I guess there could be situations where we'd want to do `diff_shared g (diff_shared f m1 m2) m1`, so maybe we should keep it documented.

Ekdohibs (approved):

PR #4263, opened 2025-07-03
Zero alloc: improve error message for arch-specific

gretay-js:
This is a minor fix for an uncommon interaction between inlining and `assume` annotations. 
The test uses an external simd instruction for simplicity to get the same output on both amd64 and arm64.


ccasin (approved):
Looks good to me!  (I did not carefully assess whether the instructrions are named correctly in both cases, but we can always fix that later if not).

PR #4264, opened 2025-07-03
Add atomic prims that take a loc

glittershark:


lambda/translcore.ml
```diff
@@ -64,6 +64,29 @@ let check_record_field_sort loc : Jkind.Sort.Const.t -> _ = function
...
+let field_offset_for_label lbl =
+  match lbl.lbl_repres with
+  | Record_boxed _
+  | Record_inlined (_, Constructor_uniform_value, Variant_boxed _)
+  | Record_inlined (_, Constructor_uniform_value, Variant_with_null) ->
+      lbl.lbl_pos
+  | Record_inlined (_, Constructor_uniform_value, Variant_extensible) ->
+      lbl.lbl_pos + 1
+  | Record_unboxed | Record_inlined (_, _, Variant_unboxed) ->
+    (* For unboxed records, no offset calculation needed in regular field
```
TheNumbat: CR layouts 5.1

lambda/translcore.ml
```diff
@@ -2226,7 +2249,14 @@ and transl_record ~scopes loc env mode fields repres opt_init_expr =
 
 and transl_atomic_loc ~scopes arg arg_sort lbl =
   let arg = transl_exp ~scopes arg_sort arg in
-  let lbl = Lconst (Const_base (Const_int (lbl.lbl_pos))) in
+  begin match lbl.lbl_repres with
+  | Record_unboxed | Record_inlined (_, _, Variant_unboxed) ->
+      (* Atomic fields not allowed here *)
```
TheNumbat, line 2320: What's the plan for these?

lambda/translprim.ml
```diff
@@ -1686,14 +1696,21 @@ let atomic_arity op (kind : atomic_kind) =
   in
   let extra_kind_arity =
     match kind with
-    | Ref (* | Loc  *)-> 0
+    | Ref | Loc-> 0
```
TheNumbat: Format

TheNumbat (approved):

lambda/translcore.ml
glittershark: well, currently you can't have a mutable field of any sort here. I think @rtjoa is planning on relaxing that restriction.

lambda/translcore.ml
glittershark: I think I did what you wanted, but please double check?

lambda/translcore.ml
rtjoa: (We may add different types of fields to unboxed records, but I don't think there are plans to change `[@@unboxed]` records. So erroring here seems reasonable.)

PR #4269, opened 2025-07-07
flambda2(types): Extract meet-related functions to a separate module

bclement-ocp:
Extract the meet-related functions (`add_equation` and the like) from the typing env to a separate `Meet_env` module.

For now the `Meet_env` module is just providing functions that operate on a `Typing_env.t`, but this is a first step towards a separate `Meet_env.t` type to be used during meets. This serves multiple purposes:

 - It makes it easier to record per-meet statistics ;
 - It makes the implementation of the `reducer`s in #4112 less awkward/fragile ;
 - It frees up the `Typing_env` module from meet-related concerns, making it closer to a dumb store of information. This provides a better separation of concerns and is also a motivation to extract these functions in a separate module rather than in a nested `Typing_env.Meet_env` module.

### DIFF
--- a/middle_end/flambda2/types/env/typing_env.ml
+++ b/middle_end/flambda2/types/env/typing_env.ml
@@ -245,14 +245,6 @@ end = struct
             now_meeting_or_joining_names t name1 name2))
 end
 
-type 'a meet_return_value =
-  | Left_input
-  | Right_input
-  | Both_inputs
-  | New_result of 'a
-
-type meet_type = t -> TG.t -> TG.t -> (TG.t meet_return_value * t) Or_bottom.t
-
 module Join_env : sig
   type t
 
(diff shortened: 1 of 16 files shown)


middle_end/flambda2/types/env/typing_env.ml
```diff
@@ -844,214 +834,22 @@ let aliases_add t ~canonical_element1 ~canonical_element2 =
...
-let add_env_extension t env_extension ~meet_type =
-  try add_env_extension ~raise_on_bottom:true t env_extension ~meet_type
-  with Bottom_equation -> make_bottom t
+let add_alias t ~canonical_element1 ~canonical_element2 :
+    _ Or_unknown_or_bottom.t =
+  match aliases_add t ~canonical_element1 ~canonical_element2 with
+  | Bottom -> Bottom
+  | exception Binding_time_resolver_failure ->
+    (* Addition of aliases between names that are both in external compilation
+       units failed, e.g. due to a .cmx file. Simply drop the equation. *)
```
lthls:
```suggestion
       units failed, e.g. due to a missing .cmx file. Simply drop the equation. *)
```

PR #4270, opened 2025-07-07
Generic Optional Arguments Parser

zchenb:


.vscode/settings.json
```diff
@@ -1,7 +1,7 @@
 {
         "ocaml.sandbox": {
                 "kind": "opam",
-                "switch": "4.14.1"
+                "switch": "4.14.2"
```
dkalinichenko-js: As discussed, could you make a separate PR for this, straight to `main`?

zchenb: Sure. See #4276.

extract_externals/traverse_typed_tree.ml
```diff
@@ -356,6 +356,8 @@ let rec split_external_type (ct : core_type) :
     match lab with
     | Nolabel | Labelled _ -> (arg, false) :: args, ret
     | Optional _ -> (arg, true) :: args, ret
+    (* CR generic-optional : CHECK *)
```
dkalinichenko-js: This change looks fine as is.

parsing/parser.mly
```diff
@@ -5171,6 +5212,10 @@ optlabel:
    | OPTLABEL                                   { $1 }
    | QUESTION LIDENT COLON                      { $2 }
 ;
+genoptlabel:
+   | GENOPTLABEL                                   { $1 }
+   | QUESTION QUOTE LIDENT COLON                      { $3 }
```
dkalinichenko-js: Nit: best to align the right column.

typing/typeclass.ml
```diff
@@ -1283,6 +1284,10 @@ and class_expr_aux cl_num val_env met_env virt self_scope scl =
         | Optional _ ->
           Location.prerr_warning pat.pat_loc
             Warnings.Unerasable_optional_argument;
+        | Generic_optional _ ->
+          (*CR generic-optional : CHECK *)
```
dkalinichenko-js: This line seems good.

parsing/ast_mapper.ml
```diff
@@ -150,6 +157,7 @@ module T = struct
 
   let map_labeled_tuple sub tl = List.map (map_snd (sub.typ sub)) tl
 
+
```
dkalinichenko-js: Nit: we try to avoid unnecessary whitespace changes in the diff.

parsing/ast_mapper.ml
```diff
@@ -467,7 +477,10 @@ module E = struct
     let desc =
       match desc with
       | Pparam_val (label, def, pat) ->
-          Pparam_val (label, Option.map (sub.expr sub) def, sub.pat sub pat)
+          Pparam_val (map_arg_label sub label
```
dkalinichenko-js, line 479: Nit: usually, we style this like:
```ocaml
Pparam_val (map_arg_label sub label,
  Option.map (sub.expr sub) def, sub.pat sub pat)
```

testsuite/tests/generic-optional-arguments/missing_type_annotation_2.ml
```diff
@@ -0,0 +1,12 @@
+(* TEST
+ flags = "-extension-universe alpha";
+
+ setup-ocamlc.byte-build-env;
+ (* This test should fail after type checking is implemented. *)
```
dkalinichenko-js: Same comment.

testsuite/tests/generic-optional-arguments/test.ml
```diff
@@ -0,0 +1,29 @@
...
+module M : S = struct
+let concat Stdlib.Option.?'(sep : string = " ") xs =
+  String.concat sep xs
+end
+
+(* Usage *)
+let default_concat ys = M.concat ys
+let comma_concat zs = M.concat ~sep:"," zs
+let chain_call Stdlib.Option.?'(sep : string option) arg =
+  M.concat Stdlib.Option.?'sep arg
```
dkalinichenko-js, line 25: This should work without the `Stdlib.Option.` prefix, make sure the parser supports that. We can always parse this into `Stdlib.Option.` this PR, and change it to actually infer it when introducing different option types like `or_null`.

typing/typetexp.ml
```diff
@@ -1628,7 +1647,13 @@ let report_error env ppf =
         (match arg_label with
         | Nolabel -> "unlabelled"
         | Optional _ -> "optional"
+        | Generic_optional _ -> "generic optional"
         | Labelled _ -> assert false )
+  | Invalid_generic_optional_argument_module_path lid ->
+      fprintf ppf
+        "Invalid generic optional argument module path: %a"
+        (Style.as_inline_code longident) lid
```
dkalinichenko-js, line 1658: Let's leave a CR to give a better error message, like this:
```
Hint: the specified module path should contain a submodule named <NAME>.
```

testsuite/tests/generic-optional-arguments/missing_type_annotation_1.ml
```diff
@@ -0,0 +1,20 @@
...
+*)
+
+(* Interface *)
+module type S = sig
+  val concat : Stdlib.Option.?'sep:string -> string list -> string
+end
+
+(* Implementation *)
+module M : S = struct
+let concat Stdlib.Option.?'(sep=" ") xs =
```
dkalinichenko-js: I don't think it should? We should always require a module name here, sure, because we can't infer it. But we don't need to require the type annotation.

testsuite/tests/generic-optional-arguments/test.ml
```diff
@@ -0,0 +1,29 @@
...
+
+
+(* Interface *)
+module type S = sig
+val concat : Stdlib.Option.?'sep:string -> string list -> string
+end
+
+(* Implementation *)
+module M : S = struct
+let concat Stdlib.Option.?'(sep : string = " ") xs =
```
dkalinichenko-js, line 14: This should also parse without the `: string` annotation.

testsuite/tests/generic-optional-arguments/test.ml
```diff
@@ -0,0 +1,29 @@
...
+  String.concat sep xs
+end
+
+(* Usage *)
+let default_concat ys = M.concat ys
+let comma_concat zs = M.concat ~sep:"," zs
+let chain_call Stdlib.Option.?'(sep : string option) arg =
```
dkalinichenko-js, line 23: Likewise this shouldn't require an annotation.

testsuite/tests/generic-optional-arguments/test.ml
```diff
@@ -0,0 +1,29 @@
+(* TEST
+ flags = "-extension-universe alpha";
+
```
dkalinichenko-js: Nit: unnecessary whitespace.

testsuite/tests/generic-optional-arguments/test.ml
```diff
@@ -0,0 +1,29 @@
...
+let default_concat ys = M.concat ys
+let comma_concat zs = M.concat ~sep:"," zs
+let chain_call Stdlib.Option.?'(sep : string option) arg =
+  M.concat Stdlib.Option.?'sep arg
+
+let () =
+  print_endline (default_concat ["x"; "y"; "z"]);
+  print_endline (comma_concat ["x"; "y"; "z"]);
+  print_endline (chain_call ["x"; "y"; "z"]);
+  ()
```
dkalinichenko-js, line 39: I'd like to have some more tests in this while, just to make sure to have more examples. E.g., you can try to have the parameters be annotated with modes like `@ local`.

testsuite/tests/parsetree/source_jane_street.ml
```diff
@@ -1485,3 +1485,79 @@ let triangle_10 = let mutable x = 0 in
...
 val triangle_10 : int = 55
 |}]
+
+(******************************)
+(* generic optional arguments *)
+(*
+The following syntaxes are tested
+
+1. in type declaration [?'lbl:tp -> ...]
+2. function parameter with default argument [?'(lbl: tp <- val)]
```
dkalinichenko-js: Update the comment to `=`.

dkalinichenko-js: Also, generally update the comment to include the module name.

testsuite/tests/generic-optional-arguments/test.ml
zchenb: Discussed in person.

testsuite/tests/generic-optional-arguments/missing_type_annotation_1.ml
zchenb: Discussed in person.

testsuite/tests/parsetree/source_jane_street.ml
zchenb: Not sure I understand the first point

> Update the comment to `=`.

testsuite/tests/generic-optional-arguments/test.ml
zchenb: Can we do this in a child feature? This change is somewhat significant to the parser. 

We discussed yesterday that we should probably start working on the type checker, and so I recommend coming back to this later.

dkalinichenko-js: That's fine, let's leave a CR (and also a CR for trying the `Module_name?` syntax).

PR #4272, opened 2025-07-07
Use vex encoding for SSE operations

TheNumbat:
When `AVX` is enabled, makes the compiler emit VEX-encoded instructions for all existing SSE operations. This includes SSE intrinsics as well as normal float and float32 operations.

There are a few reasons why this is a good idea (and why gcc/clang do so):
- When the VEX encoded instructions write to an XMM register, they clear the bits above 128. The legacy instructions preserve the upper bits, which can introduce false data dependencies.
- Many of the VEX instructions have three operands, which can improve register allocation. This is currently only taken advantage of by the `simd_selection` code path. The float/float32 operations are converted in #4300 
- Switching between decoding VEX and legacy instructions has some cost 

This can increase code size, but it looks like a negligible amount:
```
vec-ext-refactor:
   text    data     bss     dec     hex filename
    32106916 3724712  28432 35860060 2232e5c _install/bin/ocamlopt.opt
    1636147  105428   28392 1769967  1b01ef ./_build/main/oxcaml/tests/small_numbers/float32_lib.exe
    1576769  101044   28392 1706205  1a08dd ./_build/main/oxcaml/tests/small_numbers/float32_u_lib.exe
    1321091  83292    28232 1432615  15dc27 ./_build/main/oxcaml/tests/simd/basic.exe
    1360998  86140    28232 1475370  16832a ./_build/main/oxcaml/tests/simd/basic_u.exe

vex-encodings:
   text    data     bss     dec     hex filename
    32133346 3728872  28432 35890650 223a5da _install/bin/ocamlopt.opt
    1636307  105428   28392 1770127  1b028f ./_build/main/oxcaml/tests/small_numbers/float32_lib.exe
    1576865  101044   28392 1706301  1a093d ./_build/main/oxcaml/tests/small_numbers/float32_u_lib.exe
    1321091   83292   28232 1432615  15dc27 ./_build/main/oxcaml/tests/simd/basic.exe
    1360998   86140   28232 1475370  16832a ./_build/main/oxcaml/tests/simd/basic_u.exe
```

### DIFF
--- a/backend/amd64/emit.ml
+++ b/backend/amd64/emit.ml
@@ -959,7 +959,7 @@ let movd src dst =
   match Arch.Extension.enabled AVX with
   | false -> I.movd src dst
   | true -> (
-    match X86_dsl.is_regf src with
+    match is_regf src with
     | false -> I.simd vmovd_X_r32m32 [| src; dst |]
     | true -> I.simd vmovd_r32m32_X [| src; dst |])
 
@@ -968,7 +968,7 @@ let movq src dst =
   match Arch.Extension.enabled AVX with
   | false -> I.movq src dst
   | true -> (
-    match X86_dsl.is_regf src with
+    match is_regf src with
     | false -> I.simd vmovq_X_r64m64 [| src; dst |]
     | true -> I.simd vmovq_r64m64_X [| src; dst |])
 
@@ -977,7 +977,7 @@ let movss src dst =
   match Arch.Extension.enabled AVX with
   | false -> I.movss src dst
   | true -> (
-    match X86_dsl.is_mem src, X86_dsl.is_mem dst with
+    match is_mem src, is_mem dst with
     | false, false -> I.simd vmovss_X_X_X [| src; dst; dst |]
     | false, true -> I.simd vmovss_m32_X [| src; dst |]
     | true, false -> I.simd vmovss_X_m32 [| src; dst |]
@@ -988,7 +988,7 @@ let movsd src dst =
   match Arch.Extension.enabled AVX with
   | false -> I.movsd src dst
   | true -> (
-    match X86_dsl.is_mem src, X86_dsl.is_mem dst with
+    match is_mem src, is_mem dst with
     | false, false -> I.simd vmovsd_X_X_X [| src; dst; dst |]
     | false, true -> I.simd vmovsd_m64_X [| src; dst |]
     | true, false -> I.simd vmovsd_X_m64 [| src; dst |]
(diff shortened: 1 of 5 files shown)


typing/predef.ml
```diff
@@ -614,14 +614,10 @@ let add_simd_stable_extension_types add_type env =
 
 (* CR-soon mslater:
   Remaining work before these can be moved to stable:
-    - Static & reinterpet casts (tests: see ops.ml)
```
gretay-js, line 617: These are fixed in other PRs, right?

tools/simdgen/amd64/amd64.csv
```diff
@@ -1081,8 +1081,8 @@ MONITOR,0F 01 C8,,NA,NA,NA,NA
 "MOVQ r/m64,xmm",66 REX.W 0F 7E /r,SSE2,ModRM:r/m (r),ModRM:reg (w),NA,NA
 "VMOVD xmm1,r32/m32",VEX.128.66.0F.W0 6E /r,AVX,ModRM:reg (w),ModRM:r/m (r),NA,NA
 "VMOVQ xmm1,r64/m64",VEX.128.66.0F.W1 6E /r,AVX,ModRM:reg (w),ModRM:r/m (r),NA,NA
-"VMOVD r32/m32,xmm1",VEX.128.66.0F.W0 7E /r,AVX,ModRM:r/m (r),ModRM:reg (w),NA,NA
-"VMOVQ r64/m64,xmm1",VEX.128.66.0F.W1 7E /r,AVX,ModRM:r/m (r),ModRM:reg (w),NA,NA
+"VMOVD r32/m32,xmm1",VEX.128.66.0F.W0 7E /r,AVX,ModRM:r/m (w),ModRM:reg (r),NA,NA
```
gretay-js: A few other cases seem to have the same bug, should they be fixed too? e.g. L1081

typing/predef.ml
TheNumbat: Yeah

tools/simdgen/amd64/amd64.csv
TheNumbat: Yeah, just fixed the other moves. Not sure if there are more, though, since they're not used by the compiler (yet).

TheNumbat: They're now used in #4300

backend/x86_gas.ml
```diff
@@ -229,10 +231,16 @@ let print_instr b = function
       i2 b ("cmp" ^ string_of_float_condition_imm imm ^ "ps") arg1 arg2
     | Cmppd, [| imm; arg1; arg2 |] ->
       i2 b ("cmp" ^ string_of_float_condition_imm imm ^ "pd") arg1 arg2
+    (* The assembler needs a suffix to disambiguate the memory argument. *)
```
gretay-js, line 234: There are a few more instructions that may need to suffix (for example sarx, bzhi, pdep, pext), is there a way to detect them based on the encoding? Or always put "q" for 64-bit memory accesses? Sounds like it was not straightforward to identify this when it was causign test failures, so may be worth doing the generic thing as a followup.

gretay-js (approved):
I don't fully understand the changes in the binary emitter but the rest looks good to me.

backend/x86_gas.ml
TheNumbat: I considered adding this but it didn't seem worth it - the only ambiguous case is when an instruction has no register-only arguments that determine the access size and the mnemonic doesn't indicate the size (like movd movq etc). So the only ambiguous instructions I can find are the int->float conversions

PR #4273, opened 2025-07-08
Use `Mode.Crossing` in jkinds

riaqn:
Based on #4619 

**_Please review by commit_**

This PR enriches `Mode.Crossing` and and use the new types in places where suitable.

This PR is bigger than ideal - I attempted to split this into smaller PRs but couldn't. In particular `Jkind_axis` and `types` and `jkind` are all interwined, and to split the changes would require some placeholder code that will be replaced by subsequent changes, which I think is not worthwhile.

# Performance concern

The changes to `types.ml` add indirection and might affect performance.
UPDATE:
Running `-dtimings` on `typecore.ml` , and the `typing` part is `1.2768s`  (10 runs average), compared to `1.2652s` in the `main` branch.

### DIFF
--- a/dune
+++ b/dune
@@ -102,6 +102,7 @@
    -w
    -67-47
    ; remove -w -67 by adding the camlinternalMenhirLib hack like the Makefile
+   ; -47 needed because [@inline available] is not recognized by the system compiler
    ))
  (ocamlopt_flags
   (:include %{project_root}/ocamlopt_flags.sexp))
--- a/dune
+++ b/dune
@@ -100,7 +100,7 @@
    -safe-string
    -strict-formats
    -w
-   -67
+   -67-47
    ; remove -w -67 by adding the camlinternalMenhirLib hack like the Makefile
    ))
  (ocamlopt_flags
--- a/typing/mode.ml
+++ b/typing/mode.ml
@@ -2518,7 +2518,6 @@ end
 
 module Visibility = struct
   module Const = C.Visibility
-  module Const_op = C.Visibility_op
 
   module Obj = struct
     type const = Const.t
@@ -2560,7 +2559,6 @@ end
 
 module Uniqueness = struct
   module Const = C.Uniqueness
-  module Const_op = C.Uniqueness_op
 
   module Obj = struct
     type const = Const.t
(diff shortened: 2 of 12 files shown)


riaqn:
This is now almost working, but I need to split this into smaller PRs.

dune
```diff
@@ -100,7 +100,7 @@
    -safe-string
    -strict-formats
    -w
-   -67
+   -67-47
```
glittershark, line 103: can you add a comment about why we're disabling 47?

typing/mode.ml
```diff
@@ -2829,24 +2817,24 @@ module Monadic = struct
...
-        let obj = proj_obj ax in
-        C.min obj
+        let obj = (proj_obj [@inlined hint]) ax in
+        (C.min [@inlined hint]) obj
 
       let min ax =
-        let obj = proj_obj ax in
-        C.max obj
+        let obj = (proj_obj [@inlined hint]) ax in
+        (C.max [@inlined hint]) obj
```
glittershark, line 2840: can you comment on the flipping here too? I think it's pretty insidious

typing/mode.ml
```diff
@@ -4146,79 +4119,216 @@ module Crossing = struct
...
+
+      let le ax (Modality (Join_with c0)) (Modality (Join_with c1)) =
+        (Mode.Const.Per_axis.le [@inlined hint]) ax c1 c0
+
+      let join ax (Modality (Join_with c0)) (Modality (Join_with c1)) =
+        Modality
+          (Join_with ((Mode.Const.Per_axis.meet [@inlined hint]) ax c0 c1))
+
+      let meet ax (Modality (Join_with c0)) (Modality (Join_with c1)) =
+        Modality (Join_with (Mode.Const.Per_axis.join ax c0 c1))
```
glittershark: same comment here re: pointing to the note on flipping

glittershark (approved):
some places where I'd like more comments, but otherwise lgtm!

typing/mode.ml
riaqn: The flipping here is unrelated to the previous flipping. I added comments to the interface as well as the implementation.

PR #4274, opened 2025-07-08
Add hints to modes

ofsouzap:
Adding hints to the solver for modes so that mode errors can have more descriptive error messages.

Hints are used in the typing solver module by having some `Hint` module implementation provided, with two types, `const` and `morph`, for hints about constant values, and hints for morphism applications, respectively.

In `typing/solver.ml`, the implementation of the solver, hints are now passed around along with their respective `mode` values throughout most of the code (for efficiency reasons, they are passed separately, not as a tuple). For example, the `'a var` record type now stores a hint for each of the `upper` and `lower` bounds (one for each).

In `typing/mode.ml`, optional hint parameters are now provided for various functions that create mode variables and functions that apply morphims to mode variables.

For printing, `typing.mode.ml` contains functions to convert solver errors, which contain information in multiple axes, to errors that only contain information for a single axis, the offending axis, and which can then be printed using the `report_axerror` function.

### DIFF
--- a/typing/typecore.ml
+++ b/typing/typecore.ml
@@ -253,7 +253,6 @@ type error =
   | Expr_not_a_record_type of record_form_packed * type_expr
   | Expr_record_type_has_wrong_boxing of record_form_packed * type_expr
   | Submode_failed of Value.error * submode_reason * Env.shared_context option
-  | Submode_failed_alloc of Alloc.error
   | Curried_application_complete of
       arg_label * Mode.Alloc.error * [`Prefix|`Single_arg|`Entire_apply]
   | Param_mode_mismatch of Alloc.equate_error
@@ -6356,7 +6355,7 @@ and type_expect_
   | Pexp_while(scond, sbody) ->
       let env = Env.add_share_lock While_loop env in
       let cond_env = Env.add_region_lock env in
-      let mode = mode_region Value.(of_const Const.max) in
+      let mode = mode_region Value.max in
       let wh_cond =
         type_expect cond_env mode scond
           (mk_expected ~explanation:While_loop_conditional Predef.type_bool)
@@ -6381,11 +6380,11 @@ and type_expect_
         exp_env = env }
   | Pexp_for(param, slow, shigh, dir, sbody) ->
       let for_from =
-        type_expect env (mode_region Value.(of_const Const.max)) slow
+        type_expect env (mode_region Value.max) slow
           (mk_expected ~explanation:For_loop_start_index Predef.type_int)
       in
       let for_to =
-        type_expect env (mode_region Value.(of_const Const.max)) shigh
+        type_expect env (mode_region Value.max) shigh
           (mk_expected ~explanation:For_loop_stop_index Predef.type_int)
       in
       let env = Env.add_share_lock For_loop env in
@@ -9802,7 +9801,7 @@ and type_n_ary_function
       | (Check _ | Assume _ | Ignore_assert_all) ->
         Zero_alloc.create_const zero_alloc
     in
-    let alloc_mode : alloc_mode = Mode.Alloc.disallow_left fun_alloc_mode in
+    let alloc_mode = Mode.Alloc.disallow_left fun_alloc_mode in
     re
       { exp_desc =
           Texp_function
@@ -11060,8 +11059,6 @@ let report_error ~loc env =
       | _ -> []
     in
     Location.errorf ~loc ~sub "%a" (Value.report_error ?target:None) fail_reason
-  | Submode_failed_alloc(fail_reason) ->
-    Location.errorf ~loc "%a" (Alloc.report_error ?target:None) fail_reason
   | Curried_application_complete (lbl, Error (ax, {left; _}), loc_kind) ->
       let sub =
         match loc_kind with
@@ -11103,8 +11100,7 @@ let report_error ~loc env =
           Location.errorf ~loc
             "This function or one of its parameters escape their region@ \
             when it is partially applied."
-      | Error (ax, err) ->
-          let left, right = err.left, err.right in
+      | Error (ax, {left; right; _}) ->
           Location.errorf ~loc
             "This function when partially applied returns a value which is %a,@ \
               but expected to be %a."
--- a/typing/env.ml
+++ b/typing/env.ml
@@ -326,6 +326,7 @@ module TycompTbl =
 
   end
 
+
 type empty = |
 
 type mode_with_locks = Mode.Value.l * locks
@@ -3303,7 +3304,7 @@ let share_mode ~errors ~env ~loc ~item ~lid vmode shared_context =
     {mode; context = Some shared_context}
 
 let closure_mode ~errors ~env ~loc ~item ~lid
-  ({mode = {Mode.monadic; comonadic}; _} as vmode) closure_context (comonadic0 : (_ * Allowance.allowed) Mode.Value.Comonadic.t) =
+  ({mode = {Mode.monadic; comonadic}; _} as vmode) closure_context comonadic0 =
   (* [mode] is the mode of the value being closed over, [comonadic0] is the mode of the closure *)
   begin
     match
@@ -4572,7 +4573,7 @@ let sharedness_hint ppf : shared_context -> _ = function
         "@[Hint: This identifier cannot be used uniquely,@ \
           because it is defined outside of the probe.@]"
 
-let print_lock_item_reason ppf (item, lid) =
+let print_lock_item ppf (item, lid) =
   let open Mode.Hint in
   match item with
   | Module ->
@@ -4786,20 +4787,20 @@ let report_lookup_error _loc env ppf = function
       fprintf ppf
         "@[%a local, so cannot be used \
           inside %s.@]"
-        print_lock_item_reason (item, lid)
+        print_lock_item (item, lid)
         (string_of_escaping_context context);
   | Once_value_used_in (item, lid, context) ->
       fprintf ppf
         "@[%a once, so cannot be used \
             inside %s@]"
-        print_lock_item_reason (item, lid)
+        print_lock_item (item, lid)
         (string_of_shared_context context)
   | Value_used_in_closure (item, lid, err) ->
       Mode.Value.Comonadic.report_error ppf ~target:(item, lid) err
   | Local_value_used_in_exclave (item, lid) ->
       fprintf ppf "@[%a local, so it cannot be used \
                   inside an exclave_@]"
-        print_lock_item_reason (item, lid)
+        print_lock_item (item, lid)
   | Non_value_used_in_object (lid, typ, err) ->
       fprintf ppf "@[%a must have a type of layout value because it is \
                    captured by an object.@ %a@]"
--- a/typing/includecore.ml
+++ b/typing/includecore.ml
@@ -335,8 +335,7 @@ let report_modality_sub_error first second ppf e =
   let print_modality id ppf m =
     Printtyp.modality ~id:(fun ppf -> Format.pp_print_string ppf id) ppf m
   in
-  let Modality.Value.Error(ax, err) = e in
-  let left, right = err.left, err.right in
+  let Modality.Value.Error(ax, {left; right; _}) = e in
   Format.fprintf ppf "%s is %a and %s is %a."
     (String.capitalize_ascii second)
     (print_modality "empty") (Atom (ax, right) : Modality.t)
@@ -344,8 +343,7 @@ let report_modality_sub_error first second ppf e =
     (print_modality "not") (Atom (ax, left) : Modality.t)
 
 let report_mode_sub_error got expected ppf e =
-  let Mode.Value.Error(ax, err) = e in
-  let left, right = err.left, err.right in
+  let Mode.Value.Error(ax, {left; right; _}) = e in
   match ax with
   | Comonadic Areality -> Format.fprintf ppf "This escapes its region."
   | _ ->
--- a/typing/typecore.mli
+++ b/typing/typecore.mli
@@ -301,7 +301,6 @@ type error =
   | Expr_not_a_record_type of record_form_packed * type_expr
   | Expr_record_type_has_wrong_boxing of record_form_packed * type_expr
   | Submode_failed of Mode.Value.error * submode_reason * Env.shared_context option
-  | Submode_failed_alloc of Mode.Alloc.error
   | Curried_application_complete of
       arg_label * Mode.Alloc.error * [`Prefix|`Single_arg|`Entire_apply]
   | Param_mode_mismatch of Mode.Alloc.equate_error
--- a/typing/typemod.ml
+++ b/typing/typemod.ml
@@ -4811,8 +4811,7 @@ let report_error ~loc _env = function
         (Style.as_inline_code (Mode.Value.Const.print_axis ax)) left
         (Style.as_inline_code (Mode.Value.Const.print_axis ax)) right
         d
-  | Submode_failed (Error (ax, err)) ->
-  let left, right = err.left, err.right in
+  | Submode_failed (Error (ax, {left; right; _})) ->
       Location.errorf ~loc
         "This is %a, but expected to be %a."
         (Style.as_inline_code (Mode.Value.Const.print_axis ax)) left


typing/solver_intf.mli
```diff
@@ -271,20 +294,29 @@ module type Solver_mono = sig
...
+  type morph
 
+  val morph_none : morph
+
+  val left_adjoint : morph -> morph
+
+  val right_adjoint : morph -> morph
+
+  val compose : morph -> morph -> morph
+end
```
riaqn: Can we give some comments to these items? That will be helpful for users of the solver.

typing/solver_intf.mli
```diff
@@ -293,9 +325,10 @@ module type S = sig
     Equal with type ('a, 'b, 'c) t = ('a, 'b, 'c) X.t
 
   (** Solver that supports lattices with monotone morphisms between them. *)
-  module Solver_mono (C : Lattices_mono) :
+  module Solver_mono (C : Lattices_mono) (Hint : Hint) :
```
riaqn: Nit pick: maybe put `Hint` before `C`? That would match the order how current solver functions take parameters. `?hint -> 'a -> 'a mode`.

typing/solver.ml
```diff
@@ -28,13 +28,18 @@ module Magic_equal (X : Equal) :
...
+    | Morph : Hint.morph * ('b, 'a, 'd) C.morph * 'b hint -> 'a hint
+    | Const : Hint.const -> 'a hint
+    | Branch : 'a * 'a hint * 'a * 'a hint -> 'a hint
+
+  type 'a error =
+    { left : 'a;
+      left_hint : 'a hint;
+      right : 'a;
+      right_hint : 'a hint
+    }
```
riaqn: These two types are also defined in `solver_intf.mli` - can we avoid that duplication? I vaguely recall that a newer version of the definition can do that?

typing/solver.ml
```diff
@@ -128,16 +141,16 @@ module Solver_mono (C : Lattices_mono) = struct
...
   type ('a, 'd) mode =
-    | Amode : 'a -> ('a, 'l * 'r) mode
+    | Amode : 'a * 'a hint -> ('a, 'l * 'r) mode
     | Amodevar : ('a, 'd) morphvar -> ('a, 'd) mode
     | Amodejoin :
-        'a * ('a, 'l * disallowed) morphvar VarMap.t
+        'a * 'a hint * ('a, 'l * disallowed) morphvar VarMap.t
         -> ('a, 'l * disallowed) mode
-        (** [Amodejoin a [mv0, mv1, ..]] represents [a join mv0 join mv1 join ..] *)
+        (** [Amodejoin a c [mv0, mv1, ..]] represents [a join mv0 join mv1 join ..] with hint [c]. *)
```
riaqn: worth commenting that the hint `c` only explains the constant; the `morphvar`s have their own hints.
(same below)

typing/solver.ml
```diff
@@ -208,97 +221,142 @@ module Solver_mono (C : Lattices_mono) = struct
...
+  let max (type a) (obj : a C.obj) = Amode (C.max obj, Const Hint.const_none)
 
-  let of_const _obj a = Amode a
+  let of_const _obj ?(hint = Hint.const_none) a = Amode (a, Const hint)
 
-  let apply_morphvar dst morph (Amorphvar (var, morph')) =
-    Amorphvar (var, C.compose dst morph morph')
+  let apply_morphvar dst morph morph_hint
+      (Amorphvar (var, morph', _morph'_hint)) =
+    Amorphvar (var, C.compose dst morph morph', morph_hint)
```
riaqn: `_morph'_hint` shouldn't be ignored - it should be composed with `morph_hint`.

typing/solver.ml
```diff
@@ -208,97 +221,142 @@ module Solver_mono (C : Lattices_mono) = struct
...
       type a b l r.
-      b C.obj -> (a, b, l * r) C.morph -> (a, l * r) mode -> (b, l * r) mode =
-   fun dst morph m ->
+      b C.obj ->
+      ?hint:Hint.morph ->
+      (a, b, l * r) C.morph ->
+      (a, l * r) mode ->
+      (b, l * r) mode =
+   fun dst ?(hint = Hint.morph_none) morph m ->
+    let create_const_morph_hint h = Morph (hint, morph, h) in
```
riaqn: This function doesn't seem worthwhile - probably just inline it?

typing/solver.ml
```diff
@@ -208,97 +221,142 @@ module Solver_mono (C : Lattices_mono) = struct
...
+    | Amodemeet (a, h, vs) ->
+      Amodemeet
+        ( C.apply dst morph a,
+          create_const_morph_hint h,
+          VarMap.map (apply_morphvar dst morph hint) vs )
+
+  let choose_hint_for_join_cc obj a a_hint b b_hint =
+    (* This just provides a minor optimization to avoid having unnecessarily-nested Branch hints.
+       When a join, [join a b], is performed, the hint we use for the result should be a [Branch]
+       if both arguments are relevant, but if we have that [a < b] in the lattice structure then
```
riaqn: should be `a <= b` (less or equal)

typing/solver.ml
```diff
@@ -208,97 +221,142 @@ module Solver_mono (C : Lattices_mono) = struct
...
-        Amodemeet (c, VarMap.map Morphvar.disallow_right mvs)
+      | Amodejoin (c, h, mvs) ->
+        Amodejoin (c, h, VarMap.map Morphvar.disallow_right mvs)
+      | Amodemeet (c, h, mvs) ->
+        Amodemeet (c, h, VarMap.map Morphvar.disallow_right mvs)
   end)
 
-  let mlower dst (Amorphvar (var, morph)) = C.apply dst morph var.lower
+  let mlower dst (Amorphvar (var, morph, hint)) =
+    C.apply dst morph var.lower, Morph (hint, morph, var.lower_hint)
```
riaqn: This function is allocating a tuple, which we want to avoid. I think we can keep the original `mlower` and add another `mlower_hint`. Also note that some call sites don't need the returned hint.
(same for `mupper`)

typing/solver.ml
```diff
@@ -417,78 +492,87 @@ module Solver_mono (C : Lattices_mono) = struct
          2. f v.lower <= g u
          3. adding g' (f v) to the u.vlower, where g' is the left adjoint of g.
       *)
-      match submode_mvc ~log dst mv (mupper dst mu) with
-      | Error a -> Error (a, mupper dst mu)
+      match submode_mvc ~log dst mv muupper muupper_hint with
+      | Error (a, a_hint) -> Error (a, a_hint, muupper, muupper_hint)
```
riaqn, line 960: I think only `mupper dst mu` needs to be bound to a variable (as cache) - all others can be calculated on the fly.

(similar for `mlower dst mv` below)

(then we can remove the bindings at the begining of this function.

typing/solver.ml
```diff
@@ -208,97 +221,142 @@ module Solver_mono (C : Lattices_mono) = struct
...
+        ( C.apply dst morph a,
+          create_const_morph_hint h,
+          VarMap.map (apply_morphvar dst morph hint) vs )
+    | Amodemeet (a, h, vs) ->
+      Amodemeet
+        ( C.apply dst morph a,
+          create_const_morph_hint h,
+          VarMap.map (apply_morphvar dst morph hint) vs )
+
+  let choose_hint_for_join_cc obj a a_hint b b_hint =
```
riaqn: I think some callsites of this function ensures that `a <= b` never happens (or maybe `b <= a` never happens?). Maybe we can have another function `hint_biased_join_const` that assumes `a <= b` is false. Then this function (let's rename to `hint_join_const` checks `a <= b`: if so, return `b_hint`; otherwise call `hint_biased_join_const`.

(Then update the callsites to use the appropriate function)

typing/solver.ml
```diff
@@ -594,54 +715,64 @@ module Solver_mono (C : Lattices_mono) = struct
       *)
   let zap_to_floor_morphvar_aux (type a r) (obj : a C.obj)
```
riaqn: this function doesn't need to return hint?

typing/solver.ml
```diff
@@ -594,54 +715,64 @@ module Solver_mono (C : Lattices_mono) = struct
...
    fun obj m ~log ->
     match m with
-    | Amode a -> a
-    | Amodevar mv -> zap_to_floor_morphvar obj mv ~commit:log
-    | Amodejoin (a, mvs) ->
-      let floor =
+    | Amode (a, _a_hint) -> a
+    | Amodevar mv -> zap_to_floor_morphvar obj mv ~commit:log |> fst
+    | Amodejoin (a, a_hint, mvs) ->
+      let floor, floor_hint =
```
riaqn: We don't need this `floor_hint`.

typing/solver.ml
```diff
@@ -654,43 +785,58 @@ module Solver_mono (C : Lattices_mono) = struct
...
+      then Amode (a, a_hint), false
+      else
+        ( Amodevar
+            (Amorphvar
+               (fresh ~lower:a ~lower_hint:a_hint obj, C.id, Hint.morph_none)),
+          true )
     | Amodevar mv ->
       (* [~lower] is not precise (because [mlower mv] is not precise), but
          it doesn't need to be *)
+      let mv_lower, mv_lower_hint = mlower obj mv in
```
riaqn: once `mlower` is splitted into two functions, we don't need the bindings here.

typing/solver_intf.mli
ofsouzap: Will do

typing/solver_intf.mli
ofsouzap: Sure, will do

typing/solver.ml
ofsouzap: Have discussed offline, won't try change this

typing/solver.ml
ofsouzap: Will do

typing/solver.ml
ofsouzap: Whoops, will fix

typing/solver.ml
ofsouzap: Will change

typing/solver.ml
ofsouzap: Sure, will do

typing/solver.ml
ofsouzap: Fixed

typing/solver.ml
ofsouzap: Have moved the bindings to later in the function, so that they are only computed when they'll definitely be later needed

typing/solver.ml
ofsouzap: Done

typing/solver.ml
ofsouzap: Will do this

typing/solver.ml
ofsouzap: Will change

typing/solver.ml
ofsouzap: Will remove

typing/solver.ml
```diff
@@ -593,17 +788,23 @@ module Solver_mono (C : Lattices_mono) = struct
...
+      | Error (a, _a_hint) ->
+        (* Note, the hint here, [_a_hint], isn't useful as it is just for explaining why
+           [a] is too high, but this is not useful as all [a] is then used for is as an
+           upper bound (not a lower bound) in the next iteration. *)
         undo_changes !log;
         loop (C.join obj a lower)
     in
-    loop (mlower obj mv)
+    let lower = mlower obj mv in
+    loop lower
```
riaqn: Since we have gone back and forth, before merging we should remove unnecessary changes like this one. 

(Just a reminder - we can do this later :smiley: )

typing/mode.ml
```diff
@@ -1592,6 +1589,249 @@ type 'a comonadic_with = 'a C.comonadic_with =
...
+      | Statefulness -> Axis Visibility
+    in
+    match m, ax with
+    | Compose (g, f), ax -> (
+      (* Operates similarly to the equivalent branch in [find_responsible_axis_single] *)
+      match find_responsible_axis_prod g ax with
+      | NoneResponsible -> NoneResponsible
+      | SourceIsSingle -> find_responsible_axis_single f
+      | Axis c_ax -> find_responsible_axis_prod f c_ax)
+    | Id, _ | Meet_with _, _ | Imply _, _ | Map_comonadic _, _ -> SourceIsSingle
```
riaqn: I think these shouldn't be `SourceIsSingle`; they should be just `Axis ax`.

For certain cases of `Meet_with` and `Imply`, we want to give `NonResponsible`. To detect that properly, this function needs to take another parameter of type `b` or `b_ax`. I'm happy to leave that for future work. In general, this function (same for the `_single` version) should lead us to one of the three cases:
- Part or the whole of the input is responsible for the output. If nothing from the source is responsible, we skip the rest of the chain.
- The morphism is responsible for the output. The morphism might behave like `Id` for specific input. For example: Applying `Meet_with_c` on `m` where `m <= c` is no-op, in which case we will skip printing the hint for the morph.
- Both the morphism and the input are responsible for the output.
- 
(again, I'm happy to leave this as future work).

typing/mode.ml
```diff
@@ -1592,6 +1589,249 @@ type 'a comonadic_with = 'a C.comonadic_with =
...
+  axerror_get_left_const err, axerror_get_right_const err
+
+(** Description of an input axis responsible for an output axis of a morphism *)
+type 'a responsible_axis =
+  | NoneResponsible : 'a responsible_axis
+      (** None of the input axes are responsible for the output axis *)
+  | SourceIsSingle : 'a responsible_axis
+      (** The input of the morphism is a single axis object that is responsible for the output axis *)
+  | Axis : ('a, 'a_x) Axis.t -> 'a responsible_axis
+      (** The specified axis of the input object is responsible for the output axis *)
```
riaqn: In retrospect, I think we should frame the type slightly differently:
```
(** Description the part of input that is responsible for some output 
N.B. [Axis] is more precise than [Full] and is prefered. *)
type 'a responsible_axis =
  | None: 'a responsible_axis
      (** The input is not responsible for the output *)
  | Full: 'a responsible_axis
      (** The whole input is responsible for the output *)
  | Axis : ('a, 'a_x) Axis.t -> 'a responsible_axis
      (** The specified axis of the input object is responsible for the output *)
```

Then, the functions should probably be called `find_responsible_input_for_output` and `find_responsible_input_for_axis_of_output`.

typing/mode.ml
```diff
@@ -1592,6 +1589,249 @@ type 'a comonadic_with = 'a C.comonadic_with =
...
+    | Proj (_a_obj, ax) -> Axis ax
+    | Compose (g, f) -> (
+      match find_responsible_axis_single g with
+      | NoneResponsible -> NoneResponsible
+      | SourceIsSingle -> find_responsible_axis_single f
+      | Axis c_ax -> find_responsible_axis_prod f c_ax)
+    | Id | Meet_with _ | Imply _ -> SourceIsSingle
+    | Max_with _ | Min_with _ | Map_comonadic _ | Monadic_to_comonadic_min
+    | Comonadic_to_monadic _ | Monadic_to_comonadic_max ->
+      assert false
```
riaqn: Following the above reframing, we will return `Full` here.

The idea is as follows. Sure, we want to ensure that printed hints are about single axis. The best way to ensure that is via type system, which we apparantly can't do. The second best is by careful implementation and reviewing, which might miss. But missing is fine, because it's an UX improvement anyway (instead of soundness). But in case of missing, it's still better to print some cumbersome (but correct) error message than raising exceptions.

typing/mode.ml
```diff
@@ -1592,6 +1589,249 @@ type 'a comonadic_with = 'a C.comonadic_with =
...
+      (ax : (r, a) Axis.t) side ->
+    (* This function is for when we have a solver hint (aka "shint") for a product lattice and
+       wish to project the hint to be for a single axis and convert it to a mode "axhint". *)
+    let a_obj = proj_obj ax r_obj in
+    let proj = Proj (r_obj, ax) in
+    let a = apply a_obj proj r in
+    let a_shint : (a, left1 * right1) S.hint =
+      let rec compose_shint_with_proj = function
+        | S.Morph (morph_hint, morph, b_shint) ->
+          S.Morph (morph_hint, compose a_obj proj morph, b_shint)
```
riaqn: Instead of calling `compose`, simply pass additional argument `ax` to `single_axis_shint_to_axhint`.
After all, the meaning of `Proj` is already captured by `find_responsible_axis_prod`?

typing/mode.ml
ofsouzap: Have discussed offline, we have a better solution, will handle in future

typing/mode.ml
```diff
@@ -2188,49 +2571,63 @@ module Monadic = struct
...
-              contention = contention1;
-              visibility = visibility1
-            };
-          right =
-            { uniqueness = uniqueness2;
-              contention = contention2;
-              visibility = visibility2
-            }
+  type axis_with_proj_pair =
+    | Axis_with_proj_pair : 'a axis * 'a * 'a -> axis_with_proj_pair
```
riaqn: I think we should remove this type. `flipped_axis_of_error` can just be inlined into `flipped_axis_of_solver_error`.
Another call site of `flipped_axis_of_error` should call `flipped_axis_of_solver_error` instead (and drop the hints for now).

typing/mode_intf.mli
```diff
@@ -14,6 +14,30 @@
 
 open Allowance
 
+type empty = |
+
+(** Hints for the mode solvers. These are axis-specific hints that contain a trace
+of the values in a single axis from an error. *)
+type ('a, 'morph, 'const) axhint =
+  | Morph :
+      'a
```
riaqn: In retrospect, it might still be better if we pull this `'a` out of the type. (And then `axerror` will contain four fields).

Then we don't need `axhint_get_const`.

typing/mode_intf.mli
```diff
@@ -14,6 +14,30 @@
...
+      * (Format.formatter -> 'a -> unit)
```
riaqn: As discussed, we will have a `'b obj` in this constructor, then we don't need this function.

typing/mode_intf.mli
```diff
@@ -126,21 +154,89 @@ end
...
+  end
+
+  val axhint_get_const : ('a, 'morph, 'const) axhint -> 'a
+
+  val axerror_get_left_const : ('a, 'lmorph, 'rmorph, 'const) axerror -> 'a
+
+  val axerror_get_right_const : ('a, 'lmorph, 'rmorph, 'const) axerror -> 'a
+
+  val axerror_get_consts_pair :
+    ('a, 'lmorph, 'rmorph, 'const) axerror -> 'a * 'a
```
riaqn: We probably don't need to expose these.

typing/mode_intf.mli
```diff
@@ -126,21 +154,89 @@ end
...
+      | Compose : ('l * 'r) morph * ('l * 'r) morph -> ('l * 'r) morph
+      constraint 'd = _ * _
+    [@@ocaml.warning "-62"]
+
+    (* This is needed for the destructive substitutions in [Common_axis] for
+       the monadic axis modules, as we can't use [neg] within the substitution due
+       to type checker limitations *)
+    type 'd neg_morph = 'd neg morph constraint 'd = _ * _
+
+    val morph_none : _ morph
```
riaqn: We probably don't need to expose this.

typing/mode_intf.mli
```diff
@@ -126,21 +154,89 @@ end
...
 
-  include Common with type error := error and module Const := Const
+  val apply_hint : ('l * 'r) hint_morph -> ('l * 'r) t -> ('l * 'r) t
 end
 
 module type S = sig
+  module Hint : sig
+    type const = None
+
+    val const_none : const
```
riaqn: We probably don't need to expose this.

typing/mode_intf.mli
```diff
@@ -70,6 +94,10 @@ type equate_step =
 module type Common = sig
   module Const : Lattice
 
+  type 'd hint_morph constraint 'd = 'l * 'r
+
+  type hint_const
+
```
riaqn: I don't understand why we need these here. Is it to for declaring `apply_hint`? I think it's fine to just declare it in the specific module that needs it.

typing/mode.ml
```diff
@@ -1570,6 +1639,277 @@ type 'a comonadic_with = 'a C.comonadic_with =
...
+let axerror_get_left_const { left; right = _ } = axhint_get_const left
+
+let axerror_get_right_const { left = _; right } = axhint_get_const right
+
+let axerror_get_consts_pair err =
+  axerror_get_left_const err, axerror_get_right_const err
+
+type submode_exn_error =
+  | SubmodeError :
+      Longident.t option
```
riaqn: It's easier to move `Longident.t` back to the exception from `env.`

typing/solver.ml
ofsouzap: Will change

typing/mode.ml
ofsouzap: Have changed that `SourceIsSingle` to `Axis ax`, this was a bug.

Will add note to do the rest in the future

typing/mode.ml
ofsouzap: Have fixed this

typing/mode.ml
ofsouzap: Have changed

typing/mode_intf.mli
ofsouzap: Have removed

typing/mode_intf.mli
ofsouzap: Have removed

typing/mode_intf.mli
ofsouzap: Have removed

typing/mode_intf.mli
ofsouzap: Have changed

typing/mode_intf.mli
ofsouzap: Have changed

typing/mode_intf.mli
ofsouzap: Have changed

typing/mode.ml
ofsouzap: Have changed

typing/typecore.ml
```diff
@@ -483,15 +460,15 @@ let check_tail_call_local_returning loc env ap_mode {region_mode; _} =
     end
   | None -> ()
 
-let meet_regional mode =
+let meet_regional hint mode =
```
riaqn: Maybe make `hint` a labeled argument for consistency?

typing/solver_intf.mli
```diff
@@ -271,20 +296,44 @@ module type Solver_mono = sig
...
+  val const_none : const
+
+  (** The type of hints for mode variables with morphisms *)
+  type 'd morph constraint 'd = 'l * 'r
+
+  (** The empty mode morphism hint *)
+  val morph_none : 'd morph
+
+  (** The skip mode morphism hint *)
+  val morph_skip : 'd morph
```
riaqn: Probably want more comments so the user knows the difference between the two.

typing/mode.ml
ofsouzap: We will implement this once we do the aforementioned reframing

testsuite/tests/typing-modes/class.ml
```diff
@@ -172,7 +172,10 @@ let u =
 Line 3, characters 17-20:
 3 |     portable_use foo
                      ^^^
-Error: This value is "nonportable" but expected to be "portable".
+Error: This value is "nonportable"
+       because it closes over a function (at Line 2, characters 21-24)
+       which is "nonportable" because it is used in a class, and classes are always nonportable.
+       However, it is expected to be "portable".
```
riaqn: As discussed, this error message needs to be fixed.

testsuite/tests/typing-modes/lazy.ml
```diff
@@ -15,8 +15,8 @@ let foo () =
 Line 2, characters 37-38:
 2 |     lazy (let x @ local = "hello" in x)
                                          ^
-Error: This value escapes its region.
-  Hint: It is the result of a lazy expression.
+Error: This value is "local"
+       but expected to be "global" because it is the result of a lazy expression.
```
riaqn: As discussed, we need to distinguish between "lazy expression" and "the result of lazy expression".

testsuite/tests/typing-modes/lazy.ml
```diff
@@ -109,9 +113,8 @@ let foo (x @ contended) =
 Line 3, characters 6-12:
 3 |     | lazy _ -> ()
           ^^^^^^
-Error: This value is "contended" but expected to be "uncontended".
-  Hint: In order to force the lazy expression,
-  the lazy needs to be uncontended.
+Error: This value is "contended"
+       but expected to be "uncontended" because it forces a lazy expression.
```
riaqn: maybe say "because this is a lazy expression being forced"

testsuite/tests/typing-modes/module.ml
```diff
@@ -42,7 +42,12 @@ let u =
 Line 10, characters 17-20:
 10 |     portable_use foo
                       ^^^
-Error: This value is "nonportable" but expected to be "portable".
+Error: This value is "nonportable"
+       because it closes over a module (at Line 7, characters 23-24)
+       which is "nonportable"
+       because it closes over a value (at Line 15, characters 12-15)
+       which is "nonportable".
+       However, it is expected to be "portable".
```
riaqn: We will print the code at the locations in a future PR. For now, let's leave some CR somewhere in the code (probably the place where the location is printed) suggesting that. Also, comment here explain what these locations point to.

testsuite/tests/typing-modes/portable-contend.ml
```diff
@@ -128,7 +124,10 @@ let foo () =
 Line 4, characters 23-26:
 4 |     let _ @ portable = bar in
                            ^^^
-Error: This value is "nonportable" but expected to be "portable".
+Error: This value is "nonportable"
+       because it closes over a value (at Line 3, characters 24-34)
+       which is "nonportable".
+       However, it is expected to be "portable".
```
riaqn: Reminder to myself: The new hint indicates the test is not written properly.
(The test should be abount mutation, not about `best_bytes`)

testsuite/tests/typing-modes/portable-contend.ml
```diff
@@ -141,7 +140,10 @@ let foo () =
 Line 4, characters 23-26:
 4 |     let _ @ portable = bar in
                            ^^^
-Error: This value is "nonportable" but expected to be "portable".
+Error: This value is "nonportable"
+       because it closes over a value (at Line 3, characters 25-26)
+       which is of some unknown mode.
```
riaqn: Maybe just say "which is of some mode"? It's unknown to the hint system, but not to the type checker.

testsuite/tests/typing-modes/stack.ml
```diff
@@ -32,7 +34,8 @@ let f () =
 Line 2, characters 17-41:
 2 |   let g = stack_ ((42, 42) : _ @ global ) in
                      ^^^^^^^^^^^^^^^^^^^^^^^^
-Error: This allocation cannot be on the stack.
+Error: This value is "local" because it is in a stack expression.
+       However, it is expected to be "global".
```
riaqn: Let's discuss this offline.

testsuite/tests/typing-modes/stack.ml
```diff
@@ -14,15 +14,17 @@ let f = ref (stack_ fun x -> x)
 Line 1, characters 12-31:
 1 | let f = ref (stack_ fun x -> x)
                 ^^^^^^^^^^^^^^^^^^^
-Error: This value escapes its region.
+Error: This value is "local" because it is in a stack expression.
```
riaqn: Should say "it is a stack allocation" (no "in")

testsuite/tests/typing-modes/statefulness-visibility.ml
```diff
@@ -586,7 +585,7 @@ let override : 'a @ contended -> ('a @ immutable shared -> 'b) -> 'b = fun x f -
 Line 1, characters 84-85:
 1 | let override : 'a @ contended -> ('a @ immutable shared -> 'b) -> 'b = fun x f -> f x
                                                                                         ^
-Error: This value is "contended" but expected to be "shared" or "uncontended".
+Error: This value is "contended" but expected to be "shared".
```
riaqn: We might want to preserve the original behavior (for easier understanding)

testsuite/tests/typing-modes/lazy.ml
ofsouzap: Done. Have split `Lazy` const hint to two, one for the lazy expression, one for the result of the lazy expression

testsuite/tests/typing-modes/stack.ml
ofsouzap: Have fixed

typing/typecore.ml
ofsouzap: Have changed

typing/solver_intf.mli
ofsouzap: Done

testsuite/tests/typing-modes/portable-contend.ml
ofsouzap: Have changed

testsuite/tests/typing-modes/class.ml
ofsouzap: Have fixed

testsuite/tests/typing-modes/lazy.ml
ofsouzap: Have fixed

testsuite/tests/typing-modes/module.ml
ofsouzap: Have done

testsuite/tests/typing-modes/stack.ml
ofsouzap: Have discussed offline, we will revert this back to the old functionality for now, and leave a CR comment to improve this in the future

testsuite/tests/typing-modes/statefulness-visibility.ml
ofsouzap: Have changed this. I've made a special case for when we are printing an expected mode that is "shared" that we will also print "uncontended". This doesn't apply to when we are printing the actual mode, as this would be confusing to the user

typing/solver_intf.mli
```diff
@@ -168,10 +168,40 @@ module type Solver_mono = sig
...
+
+  type hint_const
+
   (** The object type from the [Lattices_mono] we're working with *)
   type 'a obj
 
-  type 'a error
+  (** ['a hint] explains a bound of type ['a], but doesn't include the bound itself *)
+  type ('a, 'd) hint =
+    | Morph :
```
riaqn: In retrospect, this constructor should be called "Apply".

typing/solver.ml
```diff
@@ -503,71 +690,91 @@ module Solver_mono (C : Lattices_mono) = struct
...
         (a, allowed * r) mode list ->
         (a, allowed * disallowed) mode =
-     fun a mvs rest ->
+     fun a a_hint mvs rest ->
       if C.le obj (C.max obj) a
-      then Amode (C.max obj)
+      then
+        (* In this case, [a] is the maximum element, and we can use [a_hint]
+           as the output's hint *)
+        Amode (C.max obj, a_hint)
```
riaqn: maybe just `Amode (a, a_hint)`? Reads better.

typing/solver.ml
```diff
@@ -503,71 +690,91 @@ module Solver_mono (C : Lattices_mono) = struct
...
         (a, l * allowed) mode list ->
         (a, disallowed * allowed) mode =
-     fun a mvs rest ->
+     fun a a_hint mvs rest ->
       if C.le obj a (C.min obj)
-      then Amode (C.min obj)
+      then
+        (* In this case, [a] is the minimum element, and we can use [a_hint]
+           as the output's hint *)
+        Amode (C.min obj, a_hint)
```
riaqn: Similiar to above.

typing/solver.ml
```diff
@@ -503,71 +690,91 @@ module Solver_mono (C : Lattices_mono) = struct
...
+              (hint_meet obj a a_hint mvupper (mupper_hint mv))
+              (cons_dedup mv mvs) xs
+          | Amodemeet (b, b_hint, mvs') ->
+            loop (C.meet obj a b)
+              (hint_meet obj a a_hint b b_hint)
+              (union_prefer_left mvs' mvs)
+              xs)
     in
-    loop (C.max obj) VarMap.empty l
+    loop (C.max obj) (Const H.const_none) VarMap.empty l
```
riaqn: I think, in addition to `H.const_none`, we also need `H.const_skip`, which says "this hint can never be printed, because it's explaining a right-bound that's max", and the dual case of left-bound that's min (below).

This is different to `const_none`.

typing/solver.ml
```diff
@@ -597,10 +804,15 @@ module Solver_mono (C : Lattices_mono) = struct
       (mv : (a, allowed * r) morphvar) =
     let rec loop lower =
       let log = ref empty_changes in
-      let r = submode_mvc ~log:(Some log) obj mv lower in
+      (* Since this call to [submode_mvc] should always succeed,
+         we provide a throwaway, arbitrary hint *)
+      let r = submode_mvc ~log:(Some log) obj mv lower (Const H.const_none) in
```
riaqn: it's not that it always succeed, but that even if it fails, we don't use the hint.

typing/solver.ml
```diff
@@ -655,43 +870,57 @@ module Solver_mono (C : Lattices_mono) = struct
...
+  let newvar obj = Amodevar (Amorphvar (fresh obj, C.id, H.morph_skip))
 
   let newvar_above (type a r) (obj : a C.obj) (m : (a, allowed * r) mode) =
     match disallow_right m with
-    | Amode a ->
+    | Amode (a, a_hint) ->
       if C.le obj (C.max obj) a
-      then Amode a, false
-      else Amodevar (Amorphvar (fresh ~lower:a obj, C.id)), true
+      then Amode (a, Const H.const_none), false
```
riaqn: Should probably use `a_hint` as hint.

typing/solver.ml
```diff
@@ -655,43 +870,57 @@ module Solver_mono (C : Lattices_mono) = struct
...
+        true )
 
   let newvar_below (type a l) (obj : a C.obj) (m : (a, l * allowed) mode) =
     match disallow_left m with
-    | Amode a ->
+    | Amode (a, a_hint) ->
       if C.le obj a (C.min obj)
-      then Amode a, false
-      else Amodevar (Amorphvar (fresh ~upper:a obj, C.id)), true
+      then Amode (a, Const H.const_none), false
```
riaqn: Should probably use `a_hint` as hint.

typing/solver.ml
ofsouzap: Have an allowance problem here. Will discuss offline

typing/solver.ml
ofsouzap: Have an allowance problem here. Will discuss offline

typing/solver.ml
ofsouzap: Have corrected

typing/solver.ml
ofsouzap: Have changed

typing/solver.ml
ofsouzap: Have changed

typing/solver_intf.mli
ofsouzap: Changed

typing/solver.ml
ofsouzap: Have an allowance problem here. Will discuss offline

typing/solver.ml
ofsouzap: Will discuss offline

ofsouzap: Have added

typing/solver.ml
ofsouzap: Have solved this by having two hints for `Amode`, one "lower" one and one "upper" one

typing/solver.ml
ofsouzap: Have solved this by having two hints for `Amode`, one "lower" one and one "upper" one

typing/mode.ml
```diff
@@ -1570,6 +1769,474 @@ type 'a comonadic_with = 'a C.comonadic_with =
...
+module Axerror = struct
+  module Printing = struct
+    type print_hint_res =
+      | HintPrinted
+      | NothingPrinted
+
+    (** Print out the text for a constant hint. Either prints nothing when there is
+  no hint and returns [NothingPrinted] or prints " because {hint}" where {hint}
+  is text for the specific constant hint and returns [HintPrinted]. *)
+    let print_const_hint a_obj a ppf : Hint.const -> unit =
```
riaqn, line 1794: this function should be moved to `module Hint`.

testsuite/tests/typing-modes/crossing.ml
```diff
@@ -129,7 +129,7 @@ let cross_local (x : cross_local @ local) : _ @ global = x
 Line 1, characters 57-58:
 1 | let cross_local (x : cross_local @ local) : _ @ global = x
                                                              ^
-Error: This value escapes its region.
+Error: This value is local to the parent region but expected to be "global".
```
riaqn, line 132: @goldfirere suggests that we could print both meesages (specifically for locality)

riaqn: Instead of "parent region", use the hint system to specify either "caller region" or "outer region". (future work. Created ticket in JIRA)

testsuite/tests/typing-modes/currying.ml
```diff
@@ -15,8 +15,8 @@ let apply1 x = g x
 Line 1, characters 15-18:
 1 | let apply1 x = g x
                    ^^^
-Error: This value escapes its region.
-  Hint: Cannot return a local value without an "exclave_" annotation.
+Error: This value is "local" but expected to be local to the parent region
+       because it is a function return value without an exclave annotation.
```
riaqn, line 19: Instead of just "regional", say "regional or global" to avoid confusing the user.

riaqn: "without an exclave annotation" should be an additional sentence, not as an appendix.

testsuite/tests/typing-modes/currying.ml
```diff
@@ -87,33 +87,35 @@ let app1 (f : a:int -> b:local_ int ref -> unit -> unit) = f ~b:(local_ ref 42)
 Line 1, characters 64-79:
 1 | let app1 (f : a:int -> b:local_ int ref -> unit -> unit) = f ~b:(local_ ref 42) ()
                                                                     ^^^^^^^^^^^^^^^
-Error: This value escapes its region.
-  Hint: It is captured by a partial application.
+Error: This value is "local" but expected to be "global"
+       because it is captured by a partial application which is "global"
+       because it is a function return value without an exclave annotation.
```
riaqn, line 92: This "it is" might be understood as refering to the underlined code, which is not right.

Maybe have a new line for "captured by a partial application" to make it more obvious what "it is" is refering to.

Or maybe, instead of "it", say "this partial appllication".

testsuite/tests/typing-modes/currying.ml
riaqn: both are resolved.

PR #4276, opened 2025-07-08
Bump sandbox version

zchenb:


dkalinichenko-js:
Hi, can you disable the Copilot review check? So far, it seems like it just adds noise to the PR.

dkalinichenko-js (approved):

zchenb:
Okay, it seems that once copilot review is requested, it cannot be removed.

PR #4280, opened 2025-07-09
Detect bit manipulation instruction (BMI) set

xclerc:
(Tentative fix for https://github.com/oxcaml/oxcaml/issues/4267.)


xclerc:
(I have also checked that the code of
this pull request behaves as expected
on Apple Silicon, with and without the
Rosetta2 layer. It is noteworthy that
for the BMI set to be reported as
available, one should set
`ROSETTA_ADVERTISE_AVX=1`.)

PR #4281, opened 2025-07-09
Refactor simd extension checks

TheNumbat:
Refactors the flags/checks for enabling amd64 SIMD extensions. Should not change behavior.

- Gets rid of `allow/require` in favor of checking the base extension. The flag to enable AVX2 now also implies AVX, so this is fine.
- `simdgen` now records the extension of the instruction, and we check that it's enabled upon emission.
- Distinguishes between instructions with a specialization immediate and another register encoded as an immediate (for use in #4272)

### DIFF
--- a/backend/amd64/arch.ml
+++ b/backend/amd64/arch.ml
@@ -90,8 +90,8 @@ module Extension = struct
 
   let enabled_by_default = function
     | SSE3 | SSSE3 | SSE4_1 | SSE4_2
-    | POPCNT | CLMUL | LZCNT | BMI | BMI2 | AVX | AVX2 -> true
-    | PREFETCHW | PREFETCHWT1 | AVX512F -> false
+    | POPCNT | CLMUL | LZCNT | BMI | BMI2 | AVX -> true
+    | AVX2 | PREFETCHW | PREFETCHWT1 | AVX512F -> false
 
   let all =
     Set.of_list
@@ -103,16 +103,21 @@ module Extension = struct
   let enabled t = Set.mem t !config
   let disabled t = not (enabled t)
 
-  let allow_vec256 () = List.exists enabled [AVX; AVX2; AVX512F]
-  let allow_vec512 () = List.exists enabled [AVX512F]
+  let implies t =
+    match t with
+    | SSE3 | SSSE3 | SSE4_1 | SSE4_2
+    | POPCNT | CLMUL | LZCNT | BMI | BMI2 | AVX
+    | PREFETCHW | PREFETCHWT1 -> [t]
+    | AVX2 -> [t; AVX]
+    | AVX512F -> [t; AVX; AVX2]
 
-  let require_vec256 () =
-    if not (allow_vec256 ())
-    then Misc.fatal_error "AVX or AVX512 is required for 256-bit vectors"
-
-  let require_vec512 () =
-    if not (allow_vec512 ())
-    then Misc.fatal_error "AVX512 is required for 512-bit vectors"
+  let implied_by t =
+    match t with
+    | SSE3 | SSSE3 | SSE4_1 | SSE4_2
+    | POPCNT | CLMUL | LZCNT | BMI | BMI2
+    | PREFETCHW | PREFETCHWT1 | AVX512F -> [t]
+    | AVX -> [t; AVX2; AVX512F]
+    | AVX2 -> [t; AVX512F]
 
   let args =
     let y t = "-f" ^ (name t |> String.lowercase_ascii) in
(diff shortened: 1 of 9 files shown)


backend/amd64/arch.ml
```diff
@@ -121,13 +126,34 @@ module Extension = struct
...
+      | SSE3 -> enabled SSE3
+      | SSSE3 -> enabled SSSE3
+      | SSE4_1 -> enabled SSE4_1
+      | SSE4_2 -> enabled SSE4_2
+      | PCLMULQDQ -> enabled CLMUL
+      | BMI2 -> enabled BMI2
+      | AVX -> enabled AVX
+      | AVX2 -> enabled AVX2
+    in
+    if not (Array.for_all enabled instr.ext)
```
gretay-js, line 174: What's the meaning of `instr.ext` when it has more than one element? I am guessing that `instr` is supported by all extension in `instr.ext`.. if so, the condition should be:
```suggestion
    if not (Array.exists enabled instr.ext)
```

backend/amd64/arch.ml
```diff
@@ -103,16 +103,21 @@ module Extension = struct
...
+    | SSE3 | SSSE3 | SSE4_1 | SSE4_2
+    | POPCNT | CLMUL | LZCNT | BMI | BMI2 | AVX
+    | PREFETCHW | PREFETCHWT1 -> [t]
+    | AVX2 -> [t; AVX]
+    | AVX512F -> [t; AVX; AVX2]
 
-  let require_vec512 () =
-    if not (allow_vec512 ())
-    then Misc.fatal_error "AVX512 is required for 512-bit vectors"
+  let implied_by t =
```
gretay-js: It would be nice to automatically keep `implied_by` and `implies` in sync with each other in the unlikely case that anything else is added. At least worth a comment about it.  Also, ensure that the default set is closed under `implies`.

(note to self: this is (reflexive) transitive closure of the simple partial order relation: AVX<=AVX2<=AVX512F but it's probably not worth implementing the machinery to compute it from  
`let less_than t1 t2 = function | AVX, AVX2 | AVX2, AVX512F -> true| _ -> false`).

gretay-js: ugh, we only have `AVX512F` so it's actually quite likely we are going to add its friends `AVX512*`

backend/amd64/emit.ml
```diff
@@ -329,10 +337,10 @@ let x86_data_type_for_stack_slot : Cmm.machtype_component -> X86_ast.data_type =
   | Float -> REAL8
   | Vec128 -> VEC128
   | Vec256 ->
-    Arch.Extension.require_vec256 ();
+    Arch.Extension.require AVX;
     VEC256
   | Vec512 ->
-    Arch.Extension.require_vec512 ();
+    Arch.Extension.require AVX512F;
```
gretay-js: If other `AVX512*` extensions are added, this and other places will need to be updated or the new extensions have to imply `AVX512F`. They probably always do, but may be worth a comment in `arch`. I kinda like the abstraction provided by `require_vec256` and `require_vec512`.

gretay-js (commented):
Is it possible to sort `Amd64_simd_instrs.t` lexicographically (or by their order in the other original csv)? Why is the diff in that file so big?

backend/amd64/arch.ml
TheNumbat: Sure, added comments. We will be adding more AVX512 cases but I'm not too concerned since these invariants don't rely on anything outside of this module - what do you think?

backend/amd64/emit.ml
TheNumbat: Yeah, all of the other extensions will imply AVX512F. I could add back `require_vec256`/512 that just checks AVX/AVX512F if you want.

backend/amd64/arch.ml
TheNumbat: It actually requires all extensions - added a comment to simd_defs.
(This is only used for `vpclmulqdq`)

backend/amd64/emit.ml
gretay-js: It would be nice to hide the fact that AVX and AVX512F are "minimal" in some sense for vec256 abd vec512, resp).

tools/simdgen/simdgen.ml
```diff
@@ -443,19 +443,16 @@ let %s = {
...
-      match Hashtbl.find_opt constructors ctr with
-      | Some () -> ()
-      | None ->
-        Hashtbl.add constructors ctr ();
-        printf "  | %s\n" ctr)
-    all_instructions;
+  Map.iter (fun binding _ -> printf "  | %s\n" binding) all;
   print_endline "\ntype nonrec instr = id instr";
-  Hashtbl.iter (fun instr () -> print_one instr) all_instructions
+  Map.iter (fun _ instr -> print_one instr) all
```
gretay-js: minor suggestion: pass `binding` to `print_one` to avoid recreatign it.

backend/amd64/arch.ml
gretay-js: as discussed offline, comments are fine for now, but if there is a simple way to define `implied_by` from `implies` it would be nice, maybe in a separate PR / when adding another extension. The order could also include `sse*` extensions.

TheNumbat: I've made it (inefficiently) compute these from `directly_implied_by`

backend/amd64/emit.ml
TheNumbat: Switched back to those

tools/simdgen/simdgen.ml
TheNumbat: Done

gretay-js (approved):

PR #4283, opened 2025-07-09
Greedy allocator: minor optimizations

xclerc:
This pull request:

- computes spilling costs only when needed;
- avoids filtering a list when evicting;
- avoids unconditionally computing both overlapping conditions.

### DIFF
--- a/backend/regalloc/regalloc_gi_utils.ml
+++ b/backend/regalloc/regalloc_gi_utils.ml
@@ -389,13 +389,23 @@ module Hardware_register = struct
   type t =
     { location : location;
       interval : Interval.t;
-      mutable assigned : assigned list
+      assigned : assigned Reg.Tbl.t
     }
 
   let add_non_evictable t reg interval =
+    assert (not (Reg.Tbl.mem t.assigned reg));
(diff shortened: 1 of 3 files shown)


backend/regalloc/regalloc_gi_utils.ml
```diff
@@ -456,27 +466,30 @@ module Hardware_registers = struct
...
-    let overlap_assigned =
-      List.exists hardware_reg.assigned
-        ~f:(fun
-             { Hardware_register.pseudo_reg = _; interval = itv; evictable = _ }
-           -> Interval.overlap itv interval)
-    in
-    let overlap = overlap_hard || overlap_assigned in
-    if debug
-    then (
-      log "overlap=%B (hard=%B, assigned=%B)" overlap overlap_hard
```
TheNumbat, line 476: Is this log intentionally deleted?

TheNumbat (approved):

backend/regalloc/regalloc_gi_utils.ml
xclerc: Yes, it was not really useful,
and it is wasteful to evaluate
both sub-conditions every
time.

PR #4285, opened 2025-07-09
Implicit kinds

dkalinichenko-js:
Introduce implicit kinds for variables, similar to a Rocq feature called [implicit types](https://rocq-prover.org/doc/V9.0.0/refman/language/extensions/implicit-arguments.html#implicit-types-of-variables).

Introduce the `[@@@implicit_kind: ('var1 : jkind1) * ('var2 : jkind2) * ...]` syntax declaring the implicit (j)kinds of all variables named `'var1`, `'var2` and so on. In the scope of an implicit kind declaration, all variables with those names are assumed to and must have the corresponding jkind. Trying to constrain the jkind with a type annotation or to override it with explicit quantification raises an error. Implicit kinds follow usual scoping rules and allow overriding.

For now, implicit kinds only work in signatures, but they will eventually work in structures. This extension is controlled by an `-extension implicit-types` flag, and we plan to also introduce implicit types for term variables with similar behavior.

### DIFF
--- a/typing/typetexp.ml
+++ b/typing/typetexp.ml
@@ -686,7 +686,8 @@ let transl_type_param env path jkind_default styp =
           Jkind.of_annotation ~context:(Type_parameter (path, name)) jkind_annot
         in
         jkind, Some jkind_annot
-    | Some _, None, Some _ -> assert false  (* no implicit jkinds for underscores *)
+    | Some _, None, Some _ -> assert false
+      (* no implicit jkinds for underscores *)
   in
   let attrs = styp.ptyp_attributes in
   match styp.ptyp_desc with


lpw25:
> Trying to constrain the jkind with a type annotation or to override it with explicit quantification raises an error

I think these behaviors surprise me for this form of the feature. I would expect using a type annotation to get you the intersection of the kinds (and an error only if that intersection does not exist. I would explicit quantification to be allowed but to be given the kind from the implicit_kind.

>  controlled by an `-implicit-types` flag

Why is this not an ordinary `-extension` flag?

dkalinichenko-js:
Oops, I meant a regular `-extension` flag.

dkalinichenko-js:
> Trying to constrain the jkind with a type annotation or to override it with explicit quantification raises an error

I think this was the conclusion of the planning meeting, at least going by the meeting notes. @goldfirere also confirmed to me that this is the intended design after.

I'm in favor of attempts to constrain the jkind raising errors, as this prevents users from e.g. declaring `[@@@implicit_kind ('elt : value_or_null)]` and then accidentally restricting the kind to `value`; since this happens for restrictions induced by type constructors, it's logical that it also applies to annotations. I don't have strong opinions on explicit overrides. Richard argued that it's nice to have `implicit_kind` mean that all variables with the specified name have the same jkind whatever else happens.

testsuite/tests/implicit-types/implicit_kinds.ml
```diff
@@ -0,0 +1,551 @@
...
+(* CR implicit-variables: implicit kinds don't work in structures. *)
+
+module M = struct
+  [@@@implicit_kind: ('elt : bits64)]
+
+  let f : 'elt -> 'elt array = fun x -> [| x |]
+end
+
+[%%expect{|
+module M : sig val f : 'elt -> 'elt array end
```
goldfirere: I would expect a warning for an unused attribute here.

testsuite/tests/implicit-types/implicit_kinds.ml
```diff
@@ -0,0 +1,551 @@
...
+module type S6 = sig
+  [@@@implicit_kind: ('no : immutable_data with int -> int)]
+
+  val x : unit -> 'no
+end
+[%%expect{|
+Line 2, characters 48-58:
+2 |   [@@@implicit_kind: ('no : immutable_data with int -> int)]
+                                                    ^^^^^^^^^^
+Error: 'with' syntax is not allowed on a right mode.
```
goldfirere, line 121: This is a poor error message (not sure if that's new in this patch), for two reasons:
* This isn't a mode.
* Users don't know about left- vs right-modes.

Instead, maybe

```
'with' syntax is allowed only on kind annotations on type declarations
```

?

testsuite/tests/implicit-types/implicit_kinds.ml
```diff
@@ -0,0 +1,551 @@
...
+      functor (X : sig val x : ('elem : immediate). 'elem end) ->
+        sig val get : ('elem : immediate). unit -> 'elem end
+  end
+|}]
+
+(* Multiple variable attributes - latter overrides former. *)
+
+module type S12 = sig
+  [@@@implicit_kind: ('a : immediate)]
+  [@@@implicit_kind: ('a : bits64)]
```
goldfirere, line 239: I think this should be an error. It's very confusing. But: it's OK in a nested signature. And if it's hard to eliminate this, then it's not worth making the error.

testsuite/tests/implicit-types/implicit_kinds.ml
```diff
@@ -0,0 +1,551 @@
...
+  type ('b : value_or_null mod immutable) t
+  val get : 'a t -> 'a
+  val geti : 'c t -> 'c
+end
+
+[%%expect{|
+Line 5, characters 12-14:
+5 |   val get : 'a t -> 'a
+                ^^
+Error: The universal type variable 'a was declared to have kind value_or_null.
```
goldfirere, line 381: I could see how this could get annoying in practice. I continue to like the "if you declare the kind for a variable with this spelling, then anything else is an error". But there's a chance I'm wrong. Maybe just change the error message to include

```
If you think we should allow this, please contact the Jane Street OCaml Language team.
```

or something. Because I imagine allowing this is actually easier than rejecting it.

testsuite/tests/implicit-types/implicit_kinds.ml
```diff
@@ -0,0 +1,551 @@
...
+       But the layout of [> `ignored ] must be a sublayout of bits32
+         because of the annotation on the implicit kind of type variables named a.
+|}]
+
+(* Implicits work in constraints. *)
+
+module type S22 = sig
+  [@@@implicit_kind: ('b : bits32)]
+
+  type 'a t constraint 'a = #('b * 'b)
```
goldfirere, line 429: Ooh. An actual increase in expressiveness. 🎉

testsuite/tests/implicit-types/implicit_kinds.ml
```diff
@@ -0,0 +1,551 @@
...
+
+  val f : 'a t -> 'a
+end
+
+[%%expect{|
+module type S22 =
+  sig
+    type 'a t constraint 'a = #('b * 'b)
```
goldfirere, line 438: This sadly has dropped some information. I don't really know what to suggest though. Maybe a kind annotation? Or maybe just ignore and move on with our lives... though "ignore" might be better read as "make a comment here saying that better output would be better".

testsuite/tests/implicit-types/implicit_kinds.ml
```diff
@@ -0,0 +1,551 @@
...
+module type S24 = sig
+  [@@@implicit_kind: ('null : value_or_null)]
+
+  type 'a t constraint 'a = int
+
+  val f : 'null t -> unit
+end
+
+[%%expect{|
+module type S24 = sig type 'a t constraint 'a = int val f : int t -> unit end
```
goldfirere, line 470: Ah, this surprises me. I would expect this to be an error. It seems strange to allow this but not allow a narrowing of the kind -- both feel like kinda the same thing but just to different degrees.

testsuite/tests/implicit-types/implicit_kinds.ml
```diff
@@ -0,0 +1,551 @@
...
+  [@@@implicit_kind: ('w : word)]
+
+  val f : ('w . 'w -> 'w) -> unit
+end
+
+[%%expect{|
+module type S27 = sig val f : (('w : word). 'w -> 'w) -> unit end
+|}]
+
+(* CR implicit-kinds: inherit the annotation here? *)
```
goldfirere: This was `CR implicit-variables` earlier.

testsuite/tests/implicit-types/implicit_kinds.ml
```diff
@@ -0,0 +1,551 @@
...
+
+(* Possible constraints. *)
+module type S31 = sig
+  [@@@implicit_kind: ('v : value_or_null)]
+
+  type 'v t constraint 'v = int
+end
+
+[%%expect{|
+module type S31 = sig type 'a t constraint 'a = int end
```
goldfirere, line 581: I think we should reject this, in sympathy with the fact that we reject narrowing of a kind.

testsuite/tests/implicit-types/implicit_kinds.ml
goldfirere, line 1: Other tests I would want:
* `implicit_kind` without the `:` (that is, a payload of the wrong type)
* What about inheriting in a module sig in a `module type of struct`? I don't really care what the answer is, but may as well test (and, thus, specify) it.
* `class type`
* some declarations above the `implicit_kind` and then others below
* would be cool if an unused `implicit_kind` declaration caused a warning, as it could be a typo (this seems hard and annoying and probably not worth it, but maybe I'm wrong)
* `external` declarations
* `external` declarations with an implicit `any` and `[@layout_poly]`
* conflicting kind annotation in a type parameter to a type (there is some special handling there, especially around `any`)

testsuite/tests/implicit-types/implicit_kinds_upstream_compatible.ml
```diff
@@ -0,0 +1,16 @@
+(* TEST
+ flags = "-extension-universe upstream_compatible";
+ expect;
+*)
+
+(* Implicit kinds are upstream-compatible. *)
```
goldfirere, line 6: What about around `immediate`? Or maybe that's caught by existing mechanisms.

parsing/language_extension.ml
goldfirere, line 1: Why do we need a new extension here? I don't see this feature as needing beta protections. So I think I'd just fold it into `layouts`. Also: when this feature expands to support types (not just kinds), I think the upstream-compatibility story might be subtle, and so we probably would want that to have a separate extension.

goldfirere: After discussing: I believe we agreed to remove this extension.

testsuite/tests/implicit-types/implicit_kinds.ml
goldfirere: After discussion, I believe I failed to convince you that this should be an error. I think it's OK to accept this, though I am unconvinced that doing so is the best design. But since neither of us has convinced the other -- and this is a dark corner of the language anyway -- I think it's ok to let whatever behavior is easier remain.

typing/typemod.ml
```diff
@@ -3330,6 +3344,8 @@ and type_open_decl_aux ?used_slot ?toplevel funct_body names env od =
 
 and type_structure ?(toplevel = None) funct_body anchor env ?expected_mode
   sstr =
+  (* CR implicit-types: implement implicit variable jkinds in structures. *)
+  let env = Env.clear_implicit_jkinds env in
```
goldfirere, line 3390: This seems perhaps wrong. It makes me wonder about this:

```
module type S = sig
  [@@@implicit_kind ('n : value_or_null)]
  module type S2 = module type of struct
    let id x = x
  end
  val f : 't -> 't
end
```

Oh.... it's probably OK because we're not actually mutating the env. Maybe? But it should still be tested.

typing/typemod.ml
```diff
@@ -2288,7 +2288,21 @@ and transl_signature env {psg_items; psg_modalities; psg_loc} =
         typedtree, tsg, newenv
     | Psig_attribute attr ->
         Builtin_attributes.parse_standard_interface_attributes attr;
-        mksig (Tsig_attribute attr) env loc, [], env
+        let newenv =
```
goldfirere, line 2369: Would be quite easy just pass the whole list to the env function and avoid rebuilding the env for every variable. Probably doesn't matter, but it's an easy change and would make things a tiny bit more efficient.

testsuite/tests/implicit-types/implicit_kinds.ml
goldfirere, line 1: Also, if this feature is going straight into stable, there needs to be user-facing documentation for it.

typing/typetexp.ml
```diff
@@ -377,31 +410,40 @@ end = struct
...
+    | Tvar { jkind } | Tunivar { jkind } when
+        not (Jkind.equate jkind jkind_info.original_jkind) ->
+      let reason =
+        Bad_univar_jkind { name; jkind_info; inferred_jkind = jkind }
+      in
+      raise (Error (loc, env, reason))
+    | _ -> ()
+
+  let quantify env loc name ty =
+    let v = Btype.proxy ty in
```
goldfirere: It's a bit sad to call this twice...

typing/typetexp.ml
```diff
@@ -594,9 +649,14 @@ let transl_type_param_var env loc attrs name_opt
...
-    | None -> jkind_default, None
-    | Some jkind_annot ->
+    match jkind_annot, name with
+    | None, None -> jkind_default, None
+    | None, Some var_name ->
+        begin match Env.find_implicit_jkind var_name env with
+        | Some jkind -> jkind, None
+        | None -> jkind_default, None
+        end
+    | Some jkind_annot, _ ->
```
goldfirere: I would expect a check for jkind consistency here. Or do we allow overrides for type parameters but not in `val`s? That's strange.

typing/typetexp.ml
```diff
@@ -1135,6 +1199,26 @@ and transl_type_poly env ~policy ~row_context mode loc vars st =
 
 and transl_type_alias env ~row_context ~policy mode attrs styp_loc styp name_opt
       jkind_annot_opt =
+  let jkind_for_fresh_var env alias alias_loc attrs jkind_annot_opt =
```
goldfirere, line 1254: This stretch seems a little strange, too. I would expect this code to accept

```
[@@@implicit_kind: ('a : value_or_null)]
val f : (int as 'a) -> 'a
val g : ('b : immediate). 'b as 'a -> 'b
```
but reject
```
val h : int as ('a : immediate) -> 'a
val j : ('b : immediate). 'b as ('a : immediate) -> 'b
```

I'd prefer a design where we just always reject using kinded variables as aliases. Or at least required an exact kind match when aliasing.

typing/typetexp.ml
```diff
@@ -1372,6 +1451,7 @@ let transl_simple_type_univars env styp =
         let policy = TyVarEnv.univars_policy in
         let typ = transl_type env policy Alloc.Const.legacy styp in
         TyVarEnv.globalize_used_variables policy env ();
+        TyVarEnv.check_rigid_variables env;
```
goldfirere: How do we know that this is called in all the places it needs to be called?

typing/typetexp.ml
```diff
@@ -258,6 +269,16 @@ end = struct
   *)
   let pre_univars = ref ([] : type_expr list)
 
+  type rigid_info = {
```
goldfirere: I think this can usefully be merged with `jkind_info`.

testsuite/tests/implicit-types/implicit_kinds.ml
dkalinichenko-js: To be clear -- I'm myself not convinced of either position here, I'm just implemented what I think we had converged on in the meeting.

testsuite/tests/implicit-types/implicit_kinds.ml
dkalinichenko-js: Unused attribute warnings do not show up in expect-tests. I added a separate test for this.

testsuite/tests/implicit-types/implicit_kinds.ml
dkalinichenko-js: This is old, unfortunately.

testsuite/tests/implicit-types/implicit_kinds.ml
dkalinichenko-js: After some thought, I decided to forbid this at the cost of forbidding nested signatures too. We can allow this to work in nested signatures later if users ask us to.

testsuite/tests/implicit-types/implicit_kinds.ml
dkalinichenko-js: It's hard to carve out specifically type constructors, if that's what you mean. I don't think we should add this message to every error, but we should listen if users complain.

testsuite/tests/implicit-types/implicit_kinds.ml
dkalinichenko-js: Added a CR.

testsuite/tests/implicit-types/implicit_kinds.ml
dkalinichenko-js: Added a CR to maybe rethink this.

testsuite/tests/implicit-types/implicit_kinds.ml
dkalinichenko-js: Fixed.

testsuite/tests/implicit-types/implicit_kinds.ml
dkalinichenko-js: Added those tests.

testsuite/tests/implicit-types/implicit_kinds_upstream_compatible.ml
dkalinichenko-js: Added a test.

typing/typemod.ml
dkalinichenko-js: Added a test.

typing/typemod.ml
dkalinichenko-js: I think that would be premature optimization. Other functions in `Env` add items to the env one-by-one. Let's be uniform.

typing/typetexp.ml
dkalinichenko-js: It's called once now.

typing/typetexp.ml
dkalinichenko-js: Fixed.

typing/typetexp.ml
dkalinichenko-js: Fixed.

typing/typetexp.ml
dkalinichenko-js: Merged this into `globalize_used_variables` so we don't miss it.

typing/typetexp.ml
dkalinichenko-js: Not `jkind_info`, but I merged this with used variables (and reorganized the code a bit).

jane/doc/extensions/_06-kinds/implicit.md
```diff
@@ -0,0 +1,74 @@
...
+
+Trying to re-declare an implicit kind will fail too:
+
+```ocaml
+module type Oiter = sig
+  [@@@implicit_kind: ('t : bits64)]
+
+  val outer : 't -> 't
+
+  module Inner : sig
```
goldfirere, line 82: Though implied by this example, the doc doesn't say that implicit kinds are inherited within subsignatures. I think this should be made explicit.

jane/doc/extensions/_06-kinds/implicit.md
```diff
@@ -0,0 +1,74 @@
...
+
+```ocaml
+type ('elt : word) collection
+
+val singleton : ('elt : word) . 'elt -> 'elt collection
+val lenght : ('elt : word) . 'elt collection -> int
+```
+
+Implicit kinds can't be overridden -- a variable declared with
+an implicit kind must always have that kind. Attempts to narrow it will fail:
```
goldfirere:
```suggestion
an implicit kind must always have that kind. Attempts to narrow or change it will fail:
```

jane/doc/extensions/_06-kinds/implicit.md
```diff
@@ -0,0 +1,74 @@
...
+
+```ocaml
+module type S = sig
+  [@@@implicit_kind: ('a : value_or_null)]
+  val i : ('a : value mod external_) -> 'a
+end
```
goldfirere, line 47:
```suggestion
  val i : ('a : value mod external_) -> 'a
  val j : ('a : bits64) -> 'a
end
```

Well not quite because we produce only one error at a time. But I'd like to clarify that narrowing and changing both cause trouble.

jane/doc/extensions/_06-kinds/implicit.md
goldfirere, line 1: Also maybe comment that implicit kinds aren't `include`d with a signature. It's just syntactic over the sig.

goldfirere, line 1: Ooh, another idea: include an example with a `constraint`, mostly because implicit kinds are the only way to interact with constraints.

testsuite/tests/implicit-types/implicit_kinds.ml
goldfirere: Is there a ticket or a CR to fix this? If not, please leave a CR here and point to a fresh internal ticket. Thanks!

testsuite/tests/implicit-types/implicit_kinds.ml
```diff
@@ -0,0 +1,729 @@
...
+2 |   [@@@implicit_kind: ('no : immutable_data with int -> int)]
+                                                    ^^^^^^^^^^
+Error: 'with' syntax is not allowed on a right mode.
+|}]
+
+
+(* Multiple implicit kinds. *)
+
+module type S7 = sig
+  [@@@implicit_kind: ('a : immediate) * ('b : bits64)]
```
goldfirere, line 128: This syntax should be documented, too.

testsuite/tests/implicit-types/implicit_kinds.ml
```diff
@@ -0,0 +1,729 @@
...
+  [@@@implicit_kind: ('a : value mod immutable)]
+
+  val i : ('a : value) -> 'a
+end
+
+[%%expect{|
+module type S18 = sig val i : ('a : value mod immutable). 'a -> 'a end
+|}]
+
+(* Quantification overrides the default. *)
```
goldfirere: This comment is stale.

testsuite/tests/implicit-types/implicit_kinds.ml
goldfirere: Sorry to poke again here, but could you also create an internal ticket and point to it from here? We just have far too many of these CRs!

testsuite/tests/implicit-types/implicit_kinds.ml
goldfirere: The current behavior (allowing a concrete unification here) corresponds with what we've planned in e.g. internal ticket 5122. Still pretty unsure of what's best (but we can merge with this).

typing/typetexp.ml
```diff
@@ -446,15 +481,21 @@ end = struct
       associate row_context p;
       p.univar, s
     with Not_found ->
-      let (univar, _, s) = TyVarMap.find name !used_variables in
-      univar, s
+      let info = TyVarMap.find name !used_variables in
+      instance info.ty, info.stage
```
goldfirere, line 491: Is this call to `instance` new?

typing/typetexp.ml
```diff
@@ -516,7 +557,12 @@ end = struct
       { flavor; unbound_variable_policy; _ } env =
     let r = ref [] in
     TyVarMap.iter
-      (fun name (ty, loc, s) ->
+      (fun name { ty; rigid; loc; stage = s } ->
+        (match rigid with
+        | Some original_jkind ->
+          let v = Btype.proxy ty in
```
goldfirere: Why this call to `proxy`? Does it change the behavior of a test if this is removed?

typing/typetexp.ml
```diff
@@ -611,11 +657,31 @@ let transl_type_param_var env loc attrs name_opt
...
+            Mismatched_jkind_annotation
+              { name = var_name; explicit_jkind = jkind;
+                implicit_jkind }));
+        jkind, Some jkind_annot
+    | Some jkind_annot, _, None ->
+        let jkind =
+          Jkind.of_annotation ~context:(Type_parameter (path, name)) jkind_annot
+        in
+        jkind, Some jkind_annot
+    | Some _, None, Some _ -> assert false
```
goldfirere:
```suggestion
    | Some _, None, Some _ -> assert false  (* no implicit jkinds for underscores *)
```

testsuite/tests/implicit-types/implicit_kinds.ml
dkalinichenko-js: Done, see internal ticket 6216.

typing/typetexp.ml
dkalinichenko-js: Looks like the call was removed by https://github.com/oxcaml/oxcaml/pull/4395 but was picked up by the rebase. The comment below wasn't removed, so this change doesn't look intentional. I say, let's restore it.

typing/typetexp.ml
dkalinichenko-js: No, it doesn't. I don't remember why I did this, I think it was mirroring some other call to `Btype.proxy`, but this one doesn't make sense now when I think about it.

PR #4286, opened 2025-07-10
Remove specific operation [Ipause] from amd64 backend

gretay-js:
`Ipause` specific operation is the same as the generic `Pause`  in #4251.
Keep `caml_pause_hint` builtin, because it's not the same as "%cpu_relax": `cpu_relax` which also adds a poll point when poll insertion is dsiabled.


backend/amd64/cfg_selection.ml
```diff
@@ -353,7 +353,6 @@ let select_operation
     match func with
     | "caml_rdtsc_unboxed" -> Rewritten (specific Irdtsc, args)
     | "caml_rdpmc_unboxed" -> Rewritten (specific Irdpmc, args)
-    | "caml_pause_hint" -> Rewritten (specific Ipause, args)
```
TheNumbat, line 356: I think we should keep this intrinsic since `%cpu_relax` also acts as a polling point when poll insertion is disabled. It can be recognized in `cmm_builtins.ml` and mapped to `Cpause` now.

gretay-js: ah, I missed that part, thanks! fixed.

PR #4290, opened 2025-07-11
Placeholder and flags for llvm backend prototype

usuyus:
This adds the flags `-llvm-backend` (enable LLVM backend), `-dllvm` (dump LLVM IR), and `-llvm-path` (specify which LLVM compiler to use) along with a placeholder pass to convert CFG into textual LLVM IR. It currently implements a very small subset of instructions and data declarations.

### DIFF
--- a/backend/cfg/llvmize.ml
+++ b/backend/cfg/llvmize.ml
@@ -46,7 +46,7 @@ end = struct
   module Gen = struct
     type ident = t
 
-    type nonrec t = { mutable next : ident }
+    type t = { mutable next : ident }
 
     (* Local identifiers are only valid within function scope, so we can reset
        it to 0 every time *)


driver/oxcaml_args.ml
```diff
@@ -251,6 +251,15 @@ let mk_internal_assembler f =
 let mk_gc_timings f =
   "-dgc-timings", Arg.Unit f, "Output information about time spent in the GC"
 
+let mk_llvm_backend f =
+  "-llvm-backend", Arg.Unit f, "Enable LLVM backend (...does nothing for now)"
+
+let mk_dllvmir f =
+  "-dllvmir", Arg.Unit f, "Dump LLVM IR"
```
gretay-js: Lets make this debug  flag undocuments, similarly to other `-d...` flags, and add a separate PR for saving llvm-ir to `.ll` file (i.e., keeping the generated llvm-ir file, instead of deleting it).  `-save-ir-after cfg-to-llvm-ir`.
```suggestion
  "-dllvmir", Arg.Unit f, " (undocumented)"
```

driver/oxcaml_args.ml
```diff
@@ -1507,6 +1528,7 @@ module Extra_params = struct
     | "gstartup" -> set' Debugging.dwarf_for_startup_file
     | "gdwarf-max-function-complexity" ->
       set_int' Debugging.dwarf_max_function_complexity
+    | "llvm-backend" -> set' Oxcaml_flags.llvm_backend
```
gretay-js, line 1531: Add `llvm-path` here too.

dune
```diff
@@ -181,7 +181,7 @@
   ast_invariants
   depend
   parser_types ; manual update: mli only files asttypes parsetree
-  ;; TYPING
+               ;; TYPING
```
gretay-js: can you remove this unrelated whitespace change please?

driver/oxcaml_flags.ml
```diff
@@ -124,6 +124,12 @@ let flags_by_opt_level ~opt_level ~default ~oclassic ~o2 ~o3 =
   | Set O2 -> o2
   | Set O3 -> o3
 
+let llvm_backend = ref false (* -llvm-backend *)
+
+let dump_llvmir = ref false (* -dllvmir *)
+
+let llvm_path = ref Default (* -llvm-path *)
```
gretay-js: `Default` is used for flambda2 flags, it's better to keep the backend flags separate. For now, using `option` here is enough. If we need more, we can introduce a separate type for it.

asmcomp/asmgen.ml
```diff
@@ -352,66 +364,95 @@ let available_regs ~stack_slots ~f x =
...
+    ++ cfg_with_layout_profile ~accumulate:true "cfg_polling"
+         (Cfg_polling.instrument_fundecl ~future_funcnames:funcnames)
+    ++ cfg_with_layout_profile ~accumulate:true "cfg_zero_alloc_checker"
+         (Zero_alloc_checker.cfg ~future_funcnames:funcnames ppf_dump)
+    ++ cfg_with_layout_profile ~accumulate:true "cfg_comballoc"
+         Cfg_comballoc.run
+    ++ Compiler_hooks.execute_and_pipe Compiler_hooks.Cfg_combine
+    ++ pass_dump_cfg_if ppf_dump Oxcaml_flags.dump_cfg "After comballoc"
+    ++ Profile.record ~accumulate:true "save_cfg" save_cfg
+    ++ Profile.record ~accumulate:true "llvmize" Llvmize.cfg
```
gretay-js: Let's make a separate function for it. It'll make it easier to see that the new pass doesn't affect normal compilation.

backend/cfg/llvmize.ml
```diff
@@ -0,0 +1,574 @@
...
+
+let begin_assembly () =
+  let t = Option.get !current_compilation_unit in
+  (* CR yusumez: Source filename needs to get emitted here (it won't work at the
+     end), but we don't have access to it here. It isn't required for now. *)
+  (* Option.iter (F.source_filename t) sourcefile; *)
+  F.line t.ppf "target triple = \"x86_64-redhat-linux-gnu\"";
+  (* CR yusumez: We need to emit these since the linker expects them to exist.
+     Remove once properly implemented. *)
+  F.line t.ppf "@camlA__data_begin = global i64 37";
```
gretay-js: This should use `Cmm_helpers.make_symbol` to get the string. Similarly for other "camlA" here.

backend/cfg/llvmize.ml
```diff
@@ -0,0 +1,574 @@
...
+  F.line t.ppf "@camlA__data_end = global i64 37";
+  F.line t.ppf "@camlA__code_begin = global i64 37";
+  F.line t.ppf "@camlA__code_end = global i64 37";
+  F.line t.ppf "@camlA__frametable = global i64 37";
```
gretay-js: Emit a valid empty frametable.

backend/cfg/llvmize.ml
```diff
@@ -0,0 +1,574 @@
...
+  let t = Option.get !current_compilation_unit in
+  (* CR yusumez: Source filename needs to get emitted here (it won't work at the
+     end), but we don't have access to it here. It isn't required for now. *)
+  (* Option.iter (F.source_filename t) sourcefile; *)
+  F.line t.ppf "target triple = \"x86_64-redhat-linux-gnu\"";
+  (* CR yusumez: We need to emit these since the linker expects them to exist.
+     Remove once properly implemented. *)
+  F.line t.ppf "@camlA__data_begin = global i64 37";
+  F.line t.ppf "@camlA__data_end = global i64 37";
+  F.line t.ppf "@camlA__code_begin = global i64 37";
```
gretay-js: Emit `_end` symbols and `frametable` in `end_assembly` instead. Try definiting `code_begin/end` as a function that does nothing and see if it ends up in .text section. data_begin/end can be zero sized, but if it doesn't work, it is fine to keep them as i64.
I don't know how to ensure correct placement of these before/after the other definitions in this compilation unit. There is a CR below, just add a note about this for now to the CR.

backend/cfg/llvmize.ml
```diff
@@ -0,0 +1,574 @@
...
+  F.line t.ppf "@camlA__code_end = global i64 37";
+  F.line t.ppf "@camlA__frametable = global i64 37";
+  Format.pp_print_newline t.ppf ()
+
+let end_assembly ~sourcefile =
+  let t = Option.get !current_compilation_unit in
+  (* Emit source_file *)
```
gretay-js:
```suggestion
```

backend/cfg/llvmize.ml
```diff
@@ -0,0 +1,574 @@
...
+  (* Call clang to compile .ll to .s *)
+  let ret_code = llvmir_to_assembly t in
+  if ret_code <> 0
+  then
+    raise
+      (Error
+         (Asm_generation
+            ( Option.value ~default:"(no source file specified)" sourcefile,
+              ret_code )));
+  if not !Oxcaml_flags.dump_llvmir then remove_file t.llvmir_filename;
```
gretay-js, line 562: This should be a separate flag, `save_llvmir` or something like that, similar to `-save-ir-after`, see handling of `should_save_ir_after` in `asmgen`.  `dump_<ir>` is used elsewhere in the backend for assorted debug output from the pass. it's fine for now, can be fixed in a follow-up pr.

backend/cfg/llvmize.ml
```diff
@@ -0,0 +1,574 @@
...
+
+let current_compilation_unit = ref None
+
+let llvmir_to_assembly t =
+  (* CR-someday gyorsh: add other optimization flags and control which passes to
+     perform. *)
+  let cmd =
+    match !Oxcaml_flags.llvm_path with
+    | Set path -> [path]
+    | Default -> Config.asm :: Misc.debug_prefix_map_flags ()
```
gretay-js: Do we need this `debug_prefix_map_flags` for the other case? Does clang support `--debug-prefix-map` or is it gcc specific?

backend/cfg/llvmize.ml
```diff
@@ -0,0 +1,574 @@
...
+  let print fmt t = Format.fprintf fmt "%d" t
+
+  module Gen = struct
+    type ident = t
+
+    type nonrec t = { mutable next : ident }
+
+    (* Local identifiers are only valid within function scope, so we can reset
+       it to 0 every time *)
+    (* let create () = { next = Label.to_int (Cmm.new_label ()) } *)
```
gretay-js:
```suggestion
```
(I thought about trying to keep block labels the same between cfg and llvmir, but it's too fragile)

backend/cfg/llvmize.ml
```diff
@@ -0,0 +1,574 @@
...
+      let res = t.next in
+      t.next <- succ res;
+      res
+  end
+end
+
+module Llvm_typ = struct
+  (** Type representing LLVM types *)
+  type t =
+    | Int of int  (** Arbitrary-sized (in bits) integer *)
```
gretay-js:
```suggestion
    | Int of int  (** Argument is the bitwidth of the integer *)
```

backend/cfg/llvmize.ml
```diff
@@ -0,0 +1,574 @@
...
+  F.line t.ppf "@camlA__code_begin = global i64 37";
+  F.line t.ppf "@camlA__code_end = global i64 37";
+  F.line t.ppf "@camlA__frametable = global i64 37";
+  Format.pp_print_newline t.ppf ()
+
+let end_assembly ~sourcefile =
+  let t = Option.get !current_compilation_unit in
+  (* Emit source_file *)
+  (* Emit data declarations *)
+  collect_data t ~declare:(F.data_decl t);
```
gretay-js: F.data_decl and t are both available inside `collect_data`.
```suggestion
  emit_data t;
```

backend/cfg/llvmize.ml
```diff
@@ -0,0 +1,574 @@
...
+  collect_regs t cfg;
+  F.define t ~fun_name ~fun_args ~fun_res_type:Cmm.typ_val ~fun_dbg ~fun_attrs
+    pp_body
+
+(* CR yusumez: Implement this *)
+let data ds =
+  let t = Option.get !current_compilation_unit in
+  t.data := List.append !(t.data) ds
+
+let collect_data t ~declare =
```
gretay-js: Please add a comment here explaining what this is doing and why it is needed. Add a CR to extend/modify `Cmm.data_item` so that it can express the structure instead of having to reconstructed it in the backend, as we discussed .

backend/cfg/llvmize.ml
```diff
@@ -0,0 +1,574 @@
...
+    List.fold_left
+      (fun (cur_sym, ds) (d : Cmm.data_item) ->
+        match d with
+        | Cdefine_symbol { sym_name; sym_global = _ } ->
+          Option.iter (fun cur_sym -> declare cur_sym ds) cur_sym;
+          Some sym_name, []
+        | Cint _ | Csymbol_address _ | Cint8 _ | Cint16 _ | Cint32 _ | Csingle _
+        | Cdouble _ | Cvec128 _ | Cvec256 _ | Cvec512 _ | Csymbol_offset _
+        | Cstring _ | Cskip _ | Calign _ ->
+          cur_sym, ds @ [d])
```
gretay-js: Add `fatal_error` here too for things not handled in `declare`. In particular, `Calign` should be associated with `Cdefine` that appears after it. Ideally, this `collect_data` would create a new representation that is passed to `declare`.

backend/cfg/llvmize.ml
```diff
@@ -0,0 +1,574 @@
...
+
+(* Mark regs we need to allocate *)
+let collect_regs t (cfg : Cfg.t) =
+  let body_regs =
+    Cfg.fold_blocks cfg
+      ~f:(fun _ block regs ->
+        let body_regs =
+          DLL.fold_left block.body
+            ~f:(fun regs (instr : Cfg.basic Cfg.instruction) ->
+              Reg.Set.add_seq (Array.to_seq instr.res) regs)
```
gretay-js: To be on the safe side, `instr.arg` should be explicitly added to `temp_regs` too. Same for terminator.

backend/cfg/llvmize.ml
```diff
@@ -0,0 +1,574 @@
...
+    data : Cmm.data_item list ref
+  }
+
+let create_fun_info () =
+  { ident_gen = Ident.Gen.create ();
+    reg2ident =
+      Reg.Tbl.create 37 (* CR yusumez: change this to be more reasonable *);
+    label2ident = Label.Tbl.create 37;
+    arg_regs = Reg.Set.empty;
+    temp_regs = Reg.Set.empty
```
gretay-js: I don't think we need to keep `temp_regs` after `alloca_regs`, and if so, `collect_regs` can be done as part of `alloca_regs`: just alloca them as soon as you see them. Then, avoid overwriting previously `alloca`d identifier for fun_arg but checking if something is arg before looking it up.

backend/cfg/llvmize.ml
```diff
@@ -0,0 +1,574 @@
...
+    F.block_label_with_predecessors t label preds;
+    DLL.iter ~f:(F.basic t) block.body;
+    F.terminator t block.terminator
+  in
+  let pp_body () =
+    (* last argument id + 1 is reserved, so we cannot use it. we explicitly skip
+       it here *)
+    let (_ : Ident.t) = Ident.Gen.get_fresh !(t.current_fun_info).ident_gen in
+    alloca_regs t;
+    F.ins t "br %a" (F.pp_label t) entry_label;
```
gretay-js, line 493: I think this branch is not needed (because entry_label is always first in the layout), but no harm.

backend/cfg/llvmize.ml
```diff
@@ -0,0 +1,574 @@
...
+      Misc.fatal_error
+        (asprintf "Unimplemented instruction: %a" Cfg.print_instruction
+           (`Basic i))
+
+  let basic_op t (i : Cfg.basic Cfg.instruction) (op : Operation.t) =
+    match op with
+    | Move ->
+      let temp = load_reg_to_temp t i.arg.(0) in
+      ins_store t temp i.res.(0)
+    | Const_int n -> ins_store' t (Nativeint.to_string n) i.res.(0)
```
gretay-js: `ins_store_nativeint` and hide the converstion inside `ins_store_nativeint`?

backend/cfg/llvmize.ml
```diff
@@ -0,0 +1,574 @@
...
+    | Const_symbol { sym_name; sym_global = _ } ->
+      ins_store_global t sym_name i.res.(0)
```
gretay-js: I think it may be a locally symbol. Maybe rename to `ins_store_symbol`?

backend/cfg/llvmize.ml
```diff
@@ -0,0 +1,574 @@
...
+       after it. *)
+    (* CR yusumez: Multiline comments don't work...? *)
+    fprintf ppf "  ; !dbg <id> %a" Debuginfo.print_compact dbg
+
+  let pp_dbg_instr ppf instr =
+    let dbg =
+      match instr with
+      | `Basic (i : Cfg.basic Cfg.instruction) -> i.dbg
+      | `Terminator (i : Cfg.terminator Cfg.instruction) -> i.dbg
+    in
```
gretay-js: This polymorphic variant should be avoidable.

backend/cfg/llvmize.ml
usuyus: No, clang doesn't have that flag, so we can remove it.

backend/cfg/llvmize.ml
usuyus: Done

backend/cfg/llvmize.ml
usuyus: I don't think LLVM likes it when it sees a label without seeing a terminator before it, so this is necessary.

backend/cfg/llvmize.ml
usuyus: Done and put the last bit in a comment.

asmcomp/asmgen.ml
usuyus: Done.

backend/cfg/llvmize.ml
```diff
@@ -0,0 +1,583 @@
...
+  end
+end = struct
+  type t = int
+
+  let print fmt t = Format.fprintf fmt "%d" t
+
+  module Gen = struct
+    type ident = t
+
+    type nonrec t = { mutable next : ident }
```
xclerc:
```suggestion
    type t = { mutable next : ident }
```

backend/cfg/llvmize.ml
```diff
@@ -0,0 +1,583 @@
...
+      let res = t.next in
+      t.next <- succ res;
+      res
+  end
+end
+
+module Llvm_typ = struct
+  (** Type representing LLVM types *)
+  type t =
+    | Int of int  (** Argument is the bitwidth of the integer *)
```
xclerc:
```suggestion
    | Int of { width_in_bits : int }
```
(maybe)

backend/cfg/llvmize.ml
```diff
@@ -0,0 +1,583 @@
...
+        (** Temporary registers mentioned in the body of the function. *)
+  }
+
+type t =
+  { llvmir_filename : string;
+    oc : Out_channel.t;
+    ppf : Format.formatter;
+    asm_filename : string;
+    current_fun_info : fun_info ref; (* should be reset for every function *)
+    data : Cmm.data_item list ref
```
xclerc: It looks like all accesses to these fields are
through `t.xyz`, so it is not immediately
clear why `ref` is preferred to `mutable`.

(Feel free to ignore and resolve if you know
future code will benefit from the reference.)

backend/cfg/llvmize.ml
```diff
@@ -0,0 +1,583 @@
...
+    ident
+
+let get_ident_for_reg t reg =
+  let fun_info = !(t.current_fun_info) in
+  match Reg.Tbl.find_opt fun_info.reg2ident reg with
+  | Some ident -> ident
+  | None ->
+    let ident = Ident.Gen.get_fresh fun_info.ident_gen in
+    Reg.Tbl.add fun_info.reg2ident reg ident;
+    ident
```
xclerc: (Could be -partially- factored out.)

backend/cfg/llvmize.ml
xclerc: (Indeed, if/when you need the sum
type, `Cfg_intf.basic_or_terminator`
can be used.)

backend/cfg/llvmize.ml
```diff
@@ -0,0 +1,583 @@
...
+        let body_regs =
+          DLL.fold_left block.body
+            ~f:(fun regs (instr : Cfg.basic Cfg.instruction) ->
+              Reg.Set.add_seq
+                (Seq.append (Array.to_seq instr.res) (Array.to_seq instr.arg))
+                regs)
+            ~init:Reg.Set.empty
+        in
+        let terminator_regs = Array.to_seq block.terminator.res in
+        Reg.Set.add_seq terminator_regs body_regs |> Reg.Set.union regs)
```
xclerc: It looks like `terminator.arg` is not
used - is it on purpose?

backend/cfg/llvmize.ml
```diff
@@ -0,0 +1,583 @@
...
+      ~init:Reg.Set.empty
+  in
+  let arg_regs = Array.to_seq cfg.fun_args |> Reg.Set.of_seq in
```
xclerc:
```suggestion
  let arg_regs = Reg.set_of_array cfg.fun_args in
```
(I think other computations in this function could
be simplified by using the helpers from [`Reg`](https://github.com/oxcaml/oxcaml/blob/main/backend/reg.mli#L102).)

asmcomp/asmgen.ml
```diff
@@ -423,15 +435,31 @@ let compile_fundecl ~ppf_dump ~funcnames fd_cmm =
...
+           cfg_with_layout
+           ++ compile_cfg ppf_dump ~funcnames fd_cmm
+           ++ pass_dump_linear_if ppf_dump dump_linear "Linearized code"
+           ++ Compiler_hooks.execute_and_pipe Compiler_hooks.Linear
+           ++ Profile.record ~accumulate:true "save_linear" save_linear
+           ++ (fun (fd : Linear.fundecl) ->
+                match !Oxcaml_flags.cfg_stack_checks with
+                | false -> Stack_check.linear fd
+                | true -> fd)
+           ++ Profile.record ~accumulate:true "emit_fundecl" emit_fundecl)
```
xclerc: (I think we should maybe consider
extracting the pipelines to two
auxiliary functions.)

xclerc (commented):
I have made a first -arguably superficial-
pass, and did not read all of `Llvmize`.

It should quite possibly be left to a future
pull request, but I think it is worth considering
having a `CFG` -> `LLVM representation`
pass followed by a `LLVM representation`
->`LLVM text IR` pass. It will be more verbose,
but will give us more flexibility than having to
convert a CFG into IR/text in one go.

backend/cfg/llvmize.ml
```diff
@@ -0,0 +1,584 @@
...
+    ~f:(fun _ block regs ->
+      let body_regs =
+        DLL.fold_left block.body
+          ~f:(fun regs (instr : Cfg.basic Cfg.instruction) ->
+            Reg.Set.add_seq
+              (Seq.append (Array.to_seq instr.res) (Array.to_seq instr.arg))
+              regs)
+          ~init:Reg.Set.empty
+      in
+      let terminator_regs = Array.to_seq block.terminator.res in
```
gretay-js: Add `terminator.arg` here to. Also, this function may be easier to understand if you use existing `Reg.add_set_array`.

backend/cfg/llvmize.ml
```diff
@@ -0,0 +1,584 @@
...
+(* CR yusumez: We do this cumbersome list wrangling since we receive data
+   declarations as a flat list. Ideally, [data_item]s would be represented in a
+   more structured manner which we can directly pass on to [declare]. *)
+let emit_data t =
+  let declare = F.data_decl t in
+  let cur_sym, ds =
+    List.fold_left
+      (fun (cur_sym, ds) (d : Cmm.data_item) ->
+        match d with
+        | Cdefine_symbol { sym_name; sym_global = _ } ->
```
gretay-js: check `sym_global` here too. fix the other occurrence too.

gretay-js (approved):
Looks good! Address the remaining minor comments and it can be merged.

PR #4292, opened 2025-07-11
build: Support `make boot-_install` with ocamllex on another device

bclement-ocp:
The `boot-_install` targets use hard links from the `_boot/_bootinstall/bin` directory to the `_install/bin` directory. However, the `_boot/_bootinstall/bin` directory contains a symbolic to the system's `ocamllex` executable, which might not be on the same device as the oxcaml repository.

Use a symbolic link for `ocamllex` in the `boot-_install` target instead.

### DIFF
--- a/Makefile.common-ox
+++ b/Makefile.common-ox
@@ -479,7 +479,7 @@ boot-_install: runtime-stdlib
 	mkdir -p _install/{bin,lib/ocaml}
 	$(cpl) _build/runtime_stdlib_install/bin/* _install/bin/
 	# can't use a hard link here because ocamllex comes from the system and
-	# is possibly on a difference device
+	# is possibly on a different device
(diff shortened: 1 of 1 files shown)


Makefile.common-ox
```diff
@@ -478,7 +478,11 @@ boot-_install: runtime-stdlib
 	rm -rf _install
 	mkdir -p _install/{bin,lib/ocaml}
 	$(cpl) _build/runtime_stdlib_install/bin/* _install/bin/
-	$(cpl) _build/_bootinstall/bin/* _install/bin/
+	# can't use a hard link here because ocamllex comes from the system and
+	# is possibly on a difference device
```
Ekdohibs:
```suggestion
	# is possibly on a different device
```

jvanburen (approved):

PR #4293, opened 2025-07-11
Add a field for function return type to Cmm and Cfg

usuyus:
This PR propagates the return `machtype` of functions from flambda2 to cfg.  This information is needed to emit llvm-ir. To this end, a field `fun_ret_type` was added to `Cmm.fundecl` and `Cfg.t`.


gretay-js (commented):
This change looks good to me.

oxcaml/testsuite/tools/parsecmm.mly
```diff
@@ -177,7 +177,8 @@ fundecl:
            ]
            else [ Reduce_code_size ];
          fun_poll = Lambda.Default_poll;
-         fun_dbg = debuginfo ()} }
+         fun_dbg = debuginfo ();
+         fun_ret_type = Cmm.typ_val } }
```
xclerc, line 183: I think we should probably do the same thing
we do for parameters. I reckon this would lead
to a large-ish diff because the syntax would
have to be tweaked, so this should probably
happen in a follow-up (just leaving a CR in
this pull request).

oxcaml/tests/backend/regalloc_validator/check_regalloc_validation.ml
```diff
@@ -95,6 +95,7 @@ module Cfg_desc = struct
         ~fun_num_stack_slots:(Stack_class.Tbl.make 0)
         ~fun_poll:Lambda.Default_poll
         ~next_instruction_id:(InstructionId.make_sequence ())
+        ~fun_ret_type:Cmm.typ_val
```
xclerc: I does not matter that much at this
point, but it feels like `fun_ret_type`
should be part of the `t` type above,
so that this line would copy the value
from the passed `t` value.

backend/cfg/cfg.mli
```diff
@@ -104,7 +104,8 @@ type t =
     fun_num_stack_slots : int Stack_class.Tbl.t;
         (** Precomputed at register allocation time *)
     fun_poll : Lambda.poll_attribute; (* Whether to insert polling points. *)
-    next_instruction_id : InstructionId.sequence (* Next instruction id. *)
+    next_instruction_id : InstructionId.sequence; (* Next instruction id. *)
+    fun_ret_type : Cmm.machtype  (** Function return type *)
```
xclerc: Should it have a comment similar to the
one for `fun_args`, essentially stating
that the value will not be used if starting
from linear?

backend/cmm_helpers.ml
```diff
@@ -4067,7 +4073,8 @@ let fail_if_called_indirectly_function () =
       fun_body;
       fun_codegen_options = [];
       fun_poll = Default_poll;
-      fun_dbg = Debuginfo.none
+      fun_dbg = Debuginfo.none;
+      fun_ret_type = typ_val
```
xclerc: (Maybe add a comment explaining
why `typ_val` is fine to use here.)

xclerc (approved):
I think something might be missing
(but can reasonably be added in a
follow-up): should we also update
the printing functions for the Cmm
and CFG representations.

gretay-js (approved):

PR #4294, opened 2025-07-11
Unsatisfactory yet impactful micro-optimizations

xclerc:
While trying to understand why https://github.com/oxcaml/oxcaml/pull/4283 was indeed making compilation
slower, I stumbled on a couple of micro-optimizations which are not
particularly elegant but seem to have an interesting effect.

`6be5c52` changes the definitions of the functions in `Int_replace_polymorphic_compare`
to externals, in order to avoid calls to compare integers (a couple of
profiles I was looking at show several of these functions taking -each-
more than 1%). I reckon it is not the proper fix, and we should rather
fix cross-module inlining in the compiler build.

`072943e` (which builds on `6be5c52`) tweaks the comparison of Mach
types to ensure `Reg.compare` and `Reg.same` contain no calls at all.
Again, it is not the proper fix, and in addition to cross-module inlining
we should investigate why the identity function is always properly
identified by the middle-end.

The table below compares the different revisions for the execution of
`make install` from a clean repository. `time` is used to get the
real/user/sys measures, and each revision is built 5 times, the slowest
run being ignored.

| Revision  | Real      | User        | Sys       |
| --------- | --------- | ----------- | --------- |
| `e0ba824` | 3m11-3m15 | 12m16-12m26 | 1m53-1m54 |
| `6be5c52` | 2m53-3m00 | 11m26-11m51 | 1m53-1m54 |
| `072943e` | 2m53-2m54 | 11m27-11m35 | 1m53-1m54 |

`072943e` is about 10% faster (real) / 7% faster (user) than the base line.

The table below compares the different revisions on IRC (using the
`regalloc.exe` tool); "IRC total" is the amount of time needed to run
IRC on all functions from the compiler and "IRC worst function" is the
time it takes to run IRC on the "largest" function.

| Revision  | IRC total   | IRC worst function |
| --------- | ----------- | ------------------ |
| `e0ba824` | 35.3s-35.8s | 1.3s-1.3s          |
| `6be5c52` | 31.6s-32.0s | 1.1s-1.3s          |
| `072943e` | 29.8s-30.4s | 1.2s-1.3s          |

`072943e` is about 15% faster than the base line for all IRC runs, and
the "worst" column shows that this is not due to a (few) pathological
cases.

### DIFF
--- a/backend/cmm.ml
+++ b/backend/cmm.ml
@@ -724,16 +724,7 @@ let map_shallow f = function
     | Cvar _ ) as c ->
     c
 
-let rank_machtype_component : machtype_component -> int = function
-  | Val -> 0
-  | Addr -> 1
-  | Int -> 2
-  | Float -> 3
-  | Vec128 -> 4
-  | Vec256 -> 5
-  | Vec512 -> 6
-  | Float32 -> 7
-  | Valx2 -> 8
+external rank_machtype_component : machtype_component -> int = "%identity"
 
 let compare_machtype_component
     ((Val | Addr | Int | Float | Vec128 | Vec256 | Vec512 | Float32 | Valx2) as
--- a/backend/reg.ml
+++ b/backend/reg.ml
@@ -63,7 +63,7 @@ module For_testing = struct
   let get_stamp () = !currstamp
   let set_state ~stamp ~relocatable_regs =
     currstamp := stamp;
-    all_relocatable_regs := relocatable_regs  
+    all_relocatable_regs := relocatable_regs
 end
 
 let create_gen ~name ~typ ~loc =
@@ -148,13 +148,15 @@ let reinit_relocatable_regs () = List.iter (fun r -> r.loc <- Unknown) !all_relo
 
 let all_relocatable_regs () = !all_relocatable_regs
 
+external rank_machtype_component : machtype_component -> int = "%identity"
+
 let compare r1 r2 =
-  let c = Int.compare r1.stamp r2.stamp in
+  let c = r1.stamp - r2.stamp in
   if c <> 0 then c
-  else Cmm.compare_machtype_component r1.typ r2.typ
+  else (rank_machtype_component r1.typ) - (rank_machtype_component r2.typ)
 
 let same r1 r2 =
-  r1.stamp = r2.stamp && Cmm.equal_machtype_component r1.typ r2.typ
+  r1.stamp = r2.stamp && (rank_machtype_component r1.typ) = (rank_machtype_component r2.typ)
 
 module RegOrder = struct
   type t = reg


backend/cmm.ml
```diff
@@ -724,16 +724,7 @@ let map_shallow f = function
...
-let rank_machtype_component : machtype_component -> int = function
-  | Val -> 0
-  | Addr -> 1
-  | Int -> 2
-  | Float -> 3
-  | Vec128 -> 4
-  | Vec256 -> 5
-  | Vec512 -> 6
-  | Float32 -> 7
-  | Valx2 -> 8
```
bclement-ocp, line 736: This should be always identified as the identity since #4058 and it seems to be when I compile it in its own file — are you saying this is not getting simplified? If so, that's definitely a bug we should investigate.

xclerc: Indeed, I also tried in a standalone file and it
was properly simplified. However, when I use
`objdump` on `cmm.o`, I get:

```
0000000000000000 <camlCmm__rank_machtype_component_1929>:
   0:   48 d1 f8                sar    %rax
   3:   48 8d 44 00 01          lea    0x1(%rax,%rax,1),%rax
   8:   c3                      retq
```

while when properly simplified, I only get `retq`.
I tried enabling O3, but that did not help.

Later in the file I get:

```
0000000000000000 <camlCmm__compare_machtype_component_1931>:
   0:   48 83 ec 18             sub    $0x18,%rsp
   4:   48 89 04 24             mov    %rax,(%rsp)
   8:   48 89 d8                mov    %rbx,%rax
   b:   e8 00 00 00 00          callq  10 <camlCmm__compare_machtype_component_1931+0x10>
  10:   48 89 44 24 08          mov    %rax,0x8(%rsp)
  15:   48 8b 04 24             mov    (%rsp),%rax
  19:   e8 00 00 00 00          callq  1e <camlCmm__compare_machtype_component_1931+0x1e>
  1e:   48 8b 5c 24 08          mov    0x8(%rsp),%rbx
  23:   48 29 d8                sub    %rbx,%rax
  26:   48 ff c0                inc    %rax
  29:   48 83 c4 18             add    $0x18,%rsp
  2d:   c3                      retq
```

which seems to indicate the rank function is not
inlined even though it is very small. That can be
fixed by an `inline` attribute though, leading to:

```
0000000000000000 <camlCmm__compare_machtype_component_1931>:
   0:   48 d1 fb                sar    %rbx
   3:   48 d1 e3                shl    %rbx
   6:   48 d1 f8                sar    %rax
   9:   48 d1 e0                shl    %rax
   c:   48 29 d8                sub    %rbx,%rax
   f:   48 ff c0                inc    %rax
  12:   c3                      retq
```

PR #4295, opened 2025-07-11
Add Generic_optional arg_label to type tree

zchenb:


typing/typeclass.ml
```diff
@@ -1566,6 +1569,7 @@ let rec approx_declaration cl =
       let arg =
         match l with
         | Optional _ -> Ctype.instance var_option
+        | Generic_optional _ -> Ctype.instance var_option
```
dkalinichenko-js, line 1573: CR: change from `var_option`.

dkalinichenko-js: Feel free to just resolve this if you already changed this in your subsequent PR.

dkalinichenko-js (approved):
A thing that surprises me here is that there's no change in any comparison functions: how would the compiler tell the difference between an `Optional` and a `Generic_optional`? You might want to look at functions involving labels like `Ctype.equivalent_with_nolabels`, and change them correspondingly.

typing/typeclass.ml
zchenb: CR Added

zchenb:
> A thing that surprises me here is that there's no change in any comparison functions: how would the compiler tell the difference between an `Optional` and a `Generic_optional`? You might want to look at functions involving labels like `Ctype.equivalent_with_nolabels`, and change them correspondingly.

Discussed in person. Will do this in the next PR.

PR #4297, opened 2025-07-11
Setup ocamlj executable

smjleo:
Mostly a lot of copying from `ocamlc` and setting up `dune` rules.


driver/main_args.ml
```diff
@@ -1703,6 +1709,137 @@ module Make_opttop_options (F : Opttop_options) = struct
   ]
 end;;
 
+module Make_jscomp_options (F : Jscomp_options) =
+struct
+  let list = [
+    (* CR selee: A lot of these are probability useless in our setting. Go through and
```
smjleo, line 1715: I'm leaving this until later because I can't tell a priori which ones we'll end up needing

Makefile
```diff
@@ -4,6 +4,7 @@ export ARCH
 
 boot_ocamlc = main_native.exe
 boot_ocamlopt = boot_ocamlopt.exe
+boot_ocamlj = jsmain_native.exe
```
jvanburen: I think we can use a better name like boot_ocamlj

driver/jscompile.ml
```diff
@@ -0,0 +1,153 @@
...
+(*   All rights reserved.  This file is distributed under the terms of    *)
+(*   the GNU Lesser General Public License version 2.1, with the          *)
+(*   special exception on linking described in the file LICENSE.          *)
+(*                                                                        *)
+(**************************************************************************)
+
+open Misc
+open Compile_common
+
+let tool_name = "ocamlc"
```
jvanburen: ocamlj?

driver/jscompile.mli
```diff
@@ -0,0 +1,58 @@
...
+  compilation_unit:Compilation_unit.t ->
+  runtime_args:Translmod.runtime_arg list ->
+  main_module_block_size:int ->
+  arg_descr:Lambda.arg_descr option ->
+  keep_symbol_tables:bool ->
+  unit
+
+(** {2 Internal functions} **)
+
+(** [to_bytecode info typed] takes a typechecked implementation
```
jvanburen, line 42: should the below functions be deleted?

driver/jsmaindriver.ml
```diff
@@ -0,0 +1,47 @@
...
+(*                                                                        *)
+(*   All rights reserved.  This file is distributed under the terms of    *)
+(*   the GNU Lesser General Public License version 2.1, with the          *)
+(*   special exception on linking described in the file LICENSE.          *)
+(*                                                                        *)
+(**************************************************************************)
+
+module Options = Main_args.Make_jscomp_options (Main_args.Default.Jsmain)
+
+let main argv ppf =
```
jvanburen, line 18: what happened to ` Symbol.this_is_ocamlj ();`

jvanburen, line 18: I think you'll want to copy/factor out some of the code from optmaindriver.ml also, which sets up some terminal shenanigans.

driver/jsmaindriver.ml
```diff
@@ -0,0 +1,47 @@
...
+  Clflags.add_arguments __LOC__ Options.list;
+  match
+    Compenv.readenv ppf Before_args;
+    Compenv.parse_arguments (ref argv) Compenv.anonymous program;
+    Compmisc.read_clflags_from_env ();
+    if !Clflags.plugin
+    then Compenv.fatal "-plugin is only supported up to OCaml 4.08.0";
+    try
+      Compenv.process_deferred_actions
+        (ppf, Jscompile.implementation, Jscompile.interface, ".cmo", ".cma")
```
jvanburen: These are ocaml bytecode files, you'll want to pick your own file extensions!

dune
```diff
@@ -336,6 +336,36 @@
...
+ (flags
+  (:standard -strict-sequence -w +67 -bin-annot -safe-string -strict-formats))
+ (ocamlopt_flags
+  (:include %{project_root}/ocamlopt_flags.sexp))
+ (libraries ocamlcommon ocamlbytecomp)
+ (modules
+   jsmaindriver
+   jscompile))
+
+(executable
```
jvanburen, line 353: you can delete this, we don't need the bytecode version of ocamlj

dune
```diff
@@ -380,6 +410,13 @@
  (section bin)
  (package ocaml))
 
+(install
+ (files
+  (main.bc as ocamlj.byte)
```
jvanburen: no need for bytecode

driver/main_args.ml
```diff
@@ -2187,4 +2323,18 @@ third-party libraries such as Lwt, but with a different API."
     let _vmthread () = Compenv.fatal vmthread_removed_message
   end
 
+  module Jsmain = struct
+    include Core
+    include Compiler
+    let _output_complete_obj () =
+      output_c_object := true;
+      output_complete_object := true;
+      (* CR selee: ??? *)
```
jvanburen, line 2332: lol

driver/jscompile.mli
smjleo: I chose to leave it here for now because I'm modifying them later to make `to_jsir` etc

driver/jsmaindriver.ml
smjleo: `Symbol.this_is_ocamlc` seems to prevent `Symbol` from running on `ocamlc`, but I saw it being used in some places in flambda2, e.g. `Closure_conversion.close_primitive`. Do you think we shouldn't be running this still?

driver/main_args.ml
smjleo: I promise to get rid of these when I'm cleaning up flags (L1714) - I'm leaving them for now because I don't know a priori which will be useful before implementing the pass

driver/jscompile.mli
jvanburen: ok, then just leave a CR that says why you're leaving them there

PR #4299, opened 2025-07-11
Add Stdlib.Or_null to generic optional arguments

zchenb:
Add type checking support for `Stdlib.Or_null` as module path in generic optional arguments.

- Type checking code that inserts option/Some/None now takes an additional parameter `mpath` that specifies whether it is `Option` or `Or_null`
- Added test cases `for` calling compatibilities between vanilla options and generic options

### DIFF
--- a/ocamldoc/odoc_info.mli
+++ b/ocamldoc/odoc_info.mli
@@ -794,7 +794,8 @@ val create_index_lists : 'a list -> ('a -> string) -> 'a list list
 (** Take a type and remove the option top constructor. This is
    useful when printing labels, we then remove the top option constructor
    for optional labels.*)
-val remove_option : Types.arg_label -> Types.type_expr -> Types.type_expr
+val remove_option :
+  Btype.optional_module_path -> Types.type_expr -> Types.type_expr
 
 (** Return [true] if the given label is optional.*)
 val is_optional : Types.arg_label -> bool
--- a/ocamldoc/odoc_info.mli
+++ b/ocamldoc/odoc_info.mli
@@ -794,7 +794,7 @@ val create_index_lists : 'a list -> ('a -> string) -> 'a list list
 (** Take a type and remove the option top constructor. This is
    useful when printing labels, we then remove the top option constructor
    for optional labels.*)
-val remove_option : Types.type_expr -> Types.type_expr
+val remove_option : Types.arg_label -> Types.type_expr -> Types.type_expr
 
 (** Return [true] if the given label is optional.*)
 val is_optional : Types.arg_label -> bool
--- a/ocamldoc/odoc_misc.ml
+++ b/ocamldoc/odoc_misc.ml
@@ -486,7 +486,7 @@ let create_index_lists elements string_of_ele =
 
 (*** for labels *)
 
-let is_optional = Btype.is_optional_arg
+let is_optional = Btype.is_optional
 let label_name = Btype.label_name
 
 let remove_option lbl typ =
@@ -497,8 +497,8 @@ let remove_option lbl typ =
       when Path.same path
             (match Btype.classify_optionality lbl with
             | Required_or_position_arg -> assert false
-            | Optional_arg path -> (
-                match Btype.classify_module_path path with
+            | Optional_arg mpath ->
+                (match mpath with
                 | Stdlib_option -> Predef.path_option
                 | Stdlib_or_null -> Predef.path_or_null))
         -> get_desc ty
--- a/ocamldoc/odoc_misc.ml
+++ b/ocamldoc/odoc_misc.ml
@@ -486,15 +486,23 @@ let create_index_lists elements string_of_ele =
 
 (*** for labels *)
 
-let is_optional = Btype.is_optional
+let is_optional = Btype.is_optional_arg
 let label_name = Btype.label_name
 
-let remove_option typ =
+let remove_option lbl typ =
   let open Types in
   let rec trim t =
     match t with
     | Tconstr(path, [ty], _)
-      when Path.same path Predef.path_option -> get_desc ty
+        when Path.same path
+              (match Btype.classify_optionality lbl with
+              | Not_optional_arg -> assert false
+              | Optional_arg path -> (
+                  match Btype.classify_module_path path with
+                  | Stdlib_option -> Predef.path_option
+                  | Stdlib_or_null -> Predef.path_or_null
+              ))
+        -> get_desc ty
     | Tconstr _
     | Tvar _
     | Tunivar _
--- a/ocamldoc/odoc_misc.ml
+++ b/ocamldoc/odoc_misc.ml
@@ -489,18 +489,12 @@ let create_index_lists elements string_of_ele =
 let is_optional = Btype.is_optional
 let label_name = Btype.label_name
 
-let remove_option lbl typ =
+let remove_option mpath typ =
   let open Types in
   let rec trim t =
     match t with
     | Tconstr(path, [ty], _)
-      when Path.same path
-            (match Btype.classify_optionality lbl with
-            | Required_or_position_arg -> assert false
-            | Optional_arg mpath ->
-                (match mpath with
-                | Stdlib_option -> Predef.path_option
-                | Stdlib_or_null -> Predef.path_or_null))
+      when Path.same path (Ctype.predef_path_of_optional_module_path mpath)
         -> get_desc ty
     | Tconstr _
     | Tvar _
--- a/ocamldoc/odoc_misc.ml
+++ b/ocamldoc/odoc_misc.ml
@@ -494,14 +494,13 @@ let remove_option lbl typ =
   let rec trim t =
     match t with
     | Tconstr(path, [ty], _)
-        when Path.same path
-              (match Btype.classify_optionality lbl with
-              | Not_optional_arg -> assert false
-              | Optional_arg path -> (
-                  match Btype.classify_module_path path with
-                  | Stdlib_option -> Predef.path_option
-                  | Stdlib_or_null -> Predef.path_or_null
-              ))
+      when Path.same path
+            (match Btype.classify_optionality lbl with
+            | Required_or_position_arg -> assert false
+            | Optional_arg path -> (
+                match Btype.classify_module_path path with
+                | Stdlib_option -> Predef.path_option
+                | Stdlib_or_null -> Predef.path_or_null))
         -> get_desc ty
     | Tconstr _
     | Tvar _
--- a/ocamldoc/odoc_str.ml
+++ b/ocamldoc/odoc_str.ml
@@ -144,7 +144,7 @@ let string_of_class_params c =
           (if parent then "(" else "")
           (Odoc_print.string_of_type_expr
              (if Odoc_misc.is_optional label then
-               Odoc_misc.remove_option t
+               Odoc_misc.remove_option label t
              else
                t
              )
--- a/ocamldoc/odoc_str.ml
+++ b/ocamldoc/odoc_str.ml
@@ -143,11 +143,9 @@ let string_of_class_params c =
           )
           (if parent then "(" else "")
           (Odoc_print.string_of_type_expr
-             (if Odoc_misc.is_optional label then
-               Odoc_misc.remove_option label t
-             else
-               t
-             )
+             (match Btype.classify_optionality label with
+             | Optional_arg mpath -> Odoc_misc.remove_option mpath t
+             | Required_or_position_arg -> t)
           )
           (if parent then ")" else "");
         iter ctype
--- a/testsuite/tests/generic-optional-arguments/calling_compatibility.ml
+++ b/testsuite/tests/generic-optional-arguments/calling_compatibility.ml
@@ -1,0 +1,221 @@
+(* TEST
+flags = "-extension-universe alpha ";
+
+ expect;
+*)
+
+(* Comprehensive test of calling compatibility for optional arguments *)
+
+(* Three function definitions *)
+let f_vanilla ?(x : int option) () = match x with None -> 0 | Some v -> v
+let f_option Stdlib.Option.?'(x = 0) () = x
+let f_or_null Stdlib.Or_null.?'(x: int or_null) () = x
+(* CR generic-optional: This seg faults *)
+(*= let f_or_null Stdlib.Or_null.?'(x = 0) () = x *)
+
+[%%expect{|
+val f_vanilla : ?x:int -> unit -> int = <fun>
+val f_option : Stdlib.Option.?'x:int -> unit -> int = <fun>
+val f_or_null : Stdlib.Or_null.?'x:int -> unit -> int or_null = <fun>
+|}]
+
+(* ===================================================================== *)
+(* Testing f_vanilla with 7 calling styles                               *)
+(* ===================================================================== *)
+
+(* 1. f_vanilla with ~x:1 *)
+let _ = f_vanilla ~x:1 ()
+[%%expect{|
+- : int = 1
+|}]
+
+(* 2. f_vanilla with ?x:(Some 1) *)
+let _ = f_vanilla ?x:(Some 1) ()
+[%%expect{|
+- : int = 1
+|}]
+
+(* 3. f_vanilla with ?x:(This 1) *)
+let _ = f_vanilla ?x:(This 1) ()
+[%%expect{|
+Line 1, characters 22-26:
+1 | let _ = f_vanilla ?x:(This 1) ()
+                          ^^^^
+Error: This variant expression is expected to have type "int option"
+       There is no constructor "This" within type "option"
+|}]
+
+(* 4. f_vanilla with Stdlib.Option.?'x:(Some 1) *)
+let _ = f_vanilla Stdlib.Option.?'x:(Some 1) ()
+[%%expect{|
+- : int = 1
+|}]
+
+(* 5. f_vanilla with Stdlib.Option.?'x:(This 1) *)
+let _ = f_vanilla Stdlib.Option.?'x:(This 1) ()
+[%%expect{|
+Line 1, characters 37-41:
+1 | let _ = f_vanilla Stdlib.Option.?'x:(This 1) ()
+                                         ^^^^
+Error: This variant expression is expected to have type "int option"
+       There is no constructor "This" within type "option"
+|}]
+
+(* 6. f_vanilla with Stdlib.Or_null.?'x:(Some 1) *)
+let _ = f_vanilla Stdlib.Or_null.?'x:(Some 1) ()
+[%%expect{|
+Line 1, characters 37-45:
+1 | let _ = f_vanilla Stdlib.Or_null.?'x:(Some 1) ()
+                                         ^^^^^^^^
+Error: The function applied to this argument has type ?x:int -> int
+This argument cannot be applied with label "Stdlib.Or_null.?'x"
+|}]
+
+(* 7. f_vanilla with Stdlib.Or_null.?'x:(This 1) *)
+let _ = f_vanilla Stdlib.Or_null.?'x:(This 1) ()
+[%%expect{|
+Line 1, characters 37-45:
+1 | let _ = f_vanilla Stdlib.Or_null.?'x:(This 1) ()
+                                         ^^^^^^^^
+Error: The function applied to this argument has type ?x:int -> int
+This argument cannot be applied with label "Stdlib.Or_null.?'x"
+|}]
+
+(* ===================================================================== *)
+(* Testing f_option with 7 calling styles                                *)
+(* ===================================================================== *)
+
+(* 1. f_option with ~x:1 *)
+let _ = f_option ~x:1 ()
+[%%expect{|
+- : int = 1
+|}]
+
+(* 2. f_option with ?x:(Some 1) *)
+let _ = f_option ?x:(Some 1) ()
+[%%expect{|
+- : int = 1
+|}]
+
+(* 3. f_option with ?x:(This 1) *)
+let _ = f_option ?x:(This 1) ()
+[%%expect{|
+Line 1, characters 21-25:
+1 | let _ = f_option ?x:(This 1) ()
+                         ^^^^
+Error: This variant expression is expected to have type "int option"
+       There is no constructor "This" within type "option"
+|}]
+
+(* 4. f_option with Stdlib.Option.?'x:(Some 1) *)
+let _ = f_option Stdlib.Option.?'x:(Some 1) ()
+[%%expect{|
+- : int = 1
+|}]
+
+(* 5. f_option with Stdlib.Option.?'x:(This 1) *)
+let _ = f_option Stdlib.Option.?'x:(This 1) ()
+[%%expect{|
+Line 1, characters 36-40:
+1 | let _ = f_option Stdlib.Option.?'x:(This 1) ()
+                                        ^^^^
+Error: This variant expression is expected to have type "int option"
+       There is no constructor "This" within type "option"
+|}]
+
+(* 6. f_option with Stdlib.Or_null.?'x:(Some 1) *)
+let _ = f_option Stdlib.Or_null.?'x:(Some 1) ()
+[%%expect{|
+Line 1, characters 36-44:
+1 | let _ = f_option Stdlib.Or_null.?'x:(Some 1) ()
+                                        ^^^^^^^^
+Error: The function applied to this argument has type
+         Stdlib.Option.?'x:int -> int
+This argument cannot be applied with label "Stdlib.Or_null.?'x"
+|}]
+
+(* 7. f_option with Stdlib.Or_null.?'x:(This 1) *)
+let _ = f_option Stdlib.Or_null.?'x:(This 1) ()
+[%%expect{|
+Line 1, characters 36-44:
+1 | let _ = f_option Stdlib.Or_null.?'x:(This 1) ()
+                                        ^^^^^^^^
+Error: The function applied to this argument has type
+         Stdlib.Option.?'x:int -> int
+This argument cannot be applied with label "Stdlib.Or_null.?'x"
+|}]
+
+(* ===================================================================== *)
+(* Testing f_or_null with 7 calling styles                               *)
+(* ===================================================================== *)
+
+(*= CR generic-optional : These test cases will cause a segfault. *)
+
+(* 1. f_or_null with ~x:1 *)
+let _ = f_or_null ~x:1 ()
+[%%expect{|
+- : int or_null = This 1
+|}]
+
+(* 2. f_or_null with ?x:(Some 1) *)
+let _ = f_or_null ?x:(Some 1) ()
+[%%expect{|
+Line 1, characters 21-29:
+1 | let _ = f_or_null ?x:(Some 1) ()
+                         ^^^^^^^^
+Error: The function applied to this argument has type
+         Stdlib.Or_null.?'x:int -> int or_null
+This argument cannot be applied with label "?x"
+|}]
+
+(* 3. f_or_null with ?x:(This 1) *)
+let _ = f_or_null ?x:(This 1) ()
+[%%expect{|
+Line 1, characters 21-29:
+1 | let _ = f_or_null ?x:(This 1) ()
+                         ^^^^^^^^
+Error: The function applied to this argument has type
+         Stdlib.Or_null.?'x:int -> int or_null
+This argument cannot be applied with label "?x"
+|}]
+
+(* 4. f_or_null with Stdlib.Option.?'x:(Some 1) *)
+let _ = f_or_null Stdlib.Option.?'x:(Some 1) ()
+[%%expect{|
+Line 1, characters 36-44:
+1 | let _ = f_or_null Stdlib.Option.?'x:(Some 1) ()
+                                        ^^^^^^^^
+Error: The function applied to this argument has type
+         Stdlib.Or_null.?'x:int -> int or_null
+This argument cannot be applied with label "Stdlib.Option.?'x"
+|}]
+
+
+(* 5. f_or_null with Stdlib.Option.?'x:(This 1) *)
+let _ = f_or_null Stdlib.Option.?'x:(This 1) ()
+[%%expect{|
+Line 1, characters 36-44:
+1 | let _ = f_or_null Stdlib.Option.?'x:(This 1) ()
+                                        ^^^^^^^^
+Error: The function applied to this argument has type
+         Stdlib.Or_null.?'x:int -> int or_null
+This argument cannot be applied with label "Stdlib.Option.?'x"
+|}]
+
+
+(* 6. f_or_null with Stdlib.Or_null.?'x:(Some 1) *)
+let _ = f_or_null Stdlib.Or_null.?'x:(Some 1) ()
+[%%expect{|
+Line 1, characters 38-42:
+1 | let _ = f_or_null Stdlib.Or_null.?'x:(Some 1) ()
+                                          ^^^^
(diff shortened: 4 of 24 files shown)


typing/btype.mli
```diff
@@ -260,16 +260,26 @@ val backtrack: snapshot -> unit
 (**** Utilities for labels ****)
 
 val is_optional_parsetree : Parsetree.arg_label -> bool
-val is_optional : arg_label -> bool
+type optionality = Optional_arg of Longident.t
+(* The reason this is called [Not_optional_arg] instead of [Required_arg] is
+that [Position] are omittable (thus not required), but is not optional *)
+                 | Not_optional_arg
```
dkalinichenko-js: `Required_or_position_arg` would be a better name in this case.

typing/btype.ml
```diff
@@ -624,12 +624,25 @@ let is_optional_parsetree : Parsetree.arg_label -> bool = function
...
+
+let classify_optionality : Types.arg_label -> optionality = function
+  | Optional _ -> Optional_arg (Ldot (Lident "Stdlib", "Option"))
+  | Generic_optional(path, _) -> Optional_arg (path.txt)
+  | _ -> Not_optional_arg
+
+(* is_optional is really buggy when generic optionals are involved.
+Renaming it to prevent its use
+*)
+let is_optional_arg arg = not (classify_optionality arg = Not_optional_arg)
```
dkalinichenko-js: I think, in such a scenario, it's best to delete `is_optional` completely, requiring the use of `classify_optionality`. Then, future callers of this function would be explicitly reminded to handle generic optional arguments.

typing/btype.ml
```diff
@@ -638,16 +651,49 @@ let label_name = function
...
+  | _ -> failwith "Only expected Stdlib.Option and Stdlib.Or_null"
 
 let prefixed_label_name = function
     Nolabel -> ""
   | Labelled s | Position s -> "~" ^ s
   | Optional s -> "?" ^ s
-
-let rec extract_label_aux hd l = function
+  | Generic_optional (path, s) ->
+    (
```
dkalinichenko-js: Nit: we try not to take excessive vertical space. It would be better to write this with parentheses on the same line as everything else.

Also, why not print out the longident directly instead of this back-and-forth?

typing/ctype.ml
```diff
@@ -3312,8 +3312,18 @@ let rec expands_to_datatype env ty =
...
   | _ -> false
 
+let arg_label_equal l1 l2 =
+  match l1, l2 with
+  | Generic_optional(path1, str1), Generic_optional(path2, str2) ->
+      str1 = str2 && (
+        classify_module_path path1.txt = classify_module_path path2.txt
+      )
+  | _ -> l1 = l2
+
```
dkalinichenko-js: Nit: excessive whitespace.

typing/ctype.ml
```diff
@@ -3312,8 +3312,18 @@ let rec expands_to_datatype env ty =
       end
   | _ -> false
 
+let arg_label_equal l1 l2 =
```
dkalinichenko-js, line 3317: When is this comparison different from standard equality?

typing/ctype.ml
```diff
@@ -4537,25 +4547,59 @@ let filter_arrow env t l ~force_tpoly =
...
-            newty2 ~level
-              (* CR layouts v5: Change the Jkind.Builtin.value when option can
-                 hold non-values. *)
-              (Tconstr(Predef.path_option,
-                       [newvar2 level Predef.option_argument_jkind],
-                       ref Mnil))
-          else if is_position l then
-            newty2 ~level (Tconstr (Predef.path_lexing_position, [], ref Mnil))
-          else
+      match force_tpoly with
```
dkalinichenko-js: In abstract, I do prefer pattern-matches to if-then-else statements, but I don't think changing `if not force_tpoly` to pattern-match here was a good idea. It's good practice to minimize the diff of your PR, and to include only changes that are relevant to your PR. Additionally, in the compiler, we try to minimize the diff with upstream, so you should be even more biased towards keeping existing code as is.

typing/ctype.ml
```diff
@@ -4537,25 +4547,59 @@ let filter_arrow env t l ~force_tpoly =
...
+                {v
+                  match l with
+                  | Optional _ ->
+                    newty2 ~level
+                      (Tconstr(Predef.path_option,
+                              [newvar2 level Predef.option_argument_jkind],
+                              ref Mnil))
+                v}
+
+                zhichen: This CR is from outdated code for handling optional
```
dkalinichenko-js: Feel free to remove it.

typing/printtyp.ml
```diff
@@ -1509,14 +1509,21 @@ let rec tree_of_typexp mode alloc_mode ty =
            at legacy, we will be able to omit printing them. *)
         let arg_mode = Alloc.zap_to_legacy marg in
         let t1 =
-          if is_optional l then
+          if is_optional_arg l then
```
dkalinichenko-js: `classify_optionality` would work better here.

typing/printtyp.ml
```diff
@@ -1509,14 +1509,21 @@ let rec tree_of_typexp mode alloc_mode ty =
...
               when Path.same path Predef.path_option ->
                 tree_of_typexp mode arg_mode ty
-            | _ -> Otyp_stuff "<hidden>"
+            | Generic_optional (genopt_path, _), Tconstr(path, [ty], _)
+              when Path.same path
+                    (match Btype.classify_module_path genopt_path.txt with
+                    | Stdlib_option -> Predef.path_option
+                    | Stdlib_or_null -> Predef.path_or_null) ->
+                tree_of_typexp mode arg_mode ty
+            | (Optional _ | Generic_optional _), _ -> Otyp_stuff "<hidden>"
```
dkalinichenko-js, line 1526: Do we ever hit this case?

typing/typeclass.ml
```diff
@@ -20,6 +20,7 @@ open Types
 open Typetexp
 open Format
 open Mode
+open Btype
```
dkalinichenko-js: Why? Opening `Btype` here seems bad.

typing/typeclass.ml
```diff
@@ -453,9 +454,16 @@ and class_type_aux env virt self_scope scty =
       in
       let ty = cty.ctyp_type in
       let ty =
-        if Btype.is_optional l
-        then Ctype.newty (Tconstr(Predef.path_option,[ty], ref Mnil))
-        else ty in
+        match l with
```
dkalinichenko-js: Wouldn't `classify_optionality` work better here?

typing/typecore.ml
```diff
@@ -884,23 +884,44 @@ let constant_or_raise env loc cst =
       c
   | Error err -> raise (Error (loc, env, err))
 
+
+let predef_path_of_arg_label arg_label =
+  match classify_optionality arg_label with
+  | Optional_arg path ->
+    (
```
dkalinichenko-js: Nit: best to put parens on the same line.

typing/typecore.ml
```diff
@@ -7479,7 +7518,12 @@ and type_function
             in
             let ty_default_arg = newvar default_arg_jkind in
             begin
-              try unify env (type_option ty_default_arg) ty_arg_mono
+              try unify env (type_option (
+                match arg_label with
```
dkalinichenko-js: Did you mean to put this in the final PR? Looks a debug assertion.

typing/ctype.ml
zchenb: `Generic_optional` has a `Location.t`,  which needs to be ignored during equality comparison

typing/typecore.ml
zchenb: I think this case is for handling default arg, which only occurs when things are optional. If you look at line immediately below, `assert false` is used.

typing/btype.ml
zchenb: is_optional_arg is often used in pattern match, e.g. 

```
| ... when is_optional_arg arg -> ...
```

inlining this will require writing `when not (classify_optionality arg = Not_optional_arg)`, which is counter-intuitive in my opinion

typing/printtyp.ml
zchenb: Here `tpoly_get_mono` cannot be called (will throw an exception) unless l is optional. So you need 2 levels of match anyways. I think the top-level `if` in this case is a better style.

typing/printtyp.ml
zchenb: I think I did hit this case when I forgot to add the case for Generic_optional, e.g. `?(x : int or_null)` will display as `?(x: <hidden>)`

ocamldoc/odoc_str.ml
```diff
@@ -144,7 +144,7 @@ let string_of_class_params c =
           (if parent then "(" else "")
           (Odoc_print.string_of_type_expr
              (if Odoc_misc.is_optional label then
```
dkalinichenko-js: This should be a pattern-match on `Btype.classify_optionality` and pass the module path to `remove_option`.

ocamldoc/odoc_info.mli
```diff
@@ -794,7 +794,7 @@ val create_index_lists : 'a list -> ('a -> string) -> 'a list list
 (** Take a type and remove the option top constructor. This is
    useful when printing labels, we then remove the top option constructor
    for optional labels.*)
-val remove_option : Types.type_expr -> Types.type_expr
+val remove_option : Types.arg_label -> Types.type_expr -> Types.type_expr
```
dkalinichenko-js: This should accept the module path instead of the label.

typing/btype.mli
```diff
@@ -259,17 +259,29 @@ val backtrack: snapshot -> unit
 
 (**** Utilities for labels ****)
 
-val is_optional_parsetree : Parsetree.arg_label -> bool
+(* CR generic-optional: temporary function, to remove *)
+type optional_module_path = Stdlib_option | Stdlib_or_null
+val classify_module_path : Longident.t -> optional_module_path
+
+type optionality = Optional_arg of optional_module_path
+(* The reason this is called [Required_or_position_arg] instead of
```
dkalinichenko-js, line 267: Nit: move this comment past the definition.

typing/ctype.ml
```diff
@@ -3312,8 +3312,15 @@ let rec expands_to_datatype env ty =
       end
   | _ -> false
 
+let arg_label_equal l1 l2 =
+  match l1, l2 with
+  | Generic_optional(path1, str1), Generic_optional(path2, str2) ->
+      str1 = str2 &&
+        (classify_module_path path1.txt = classify_module_path path2.txt)
```
dkalinichenko-js: Nit: I think parens are unnecessary here.

ocamldoc/odoc_misc.ml
```diff
@@ -489,12 +489,19 @@ let create_index_lists elements string_of_ele =
...
   let open Types in
   let rec trim t =
     match t with
     | Tconstr(path, [ty], _)
-      when Path.same path Predef.path_option -> get_desc ty
+      when Path.same path
+            (match Btype.classify_optionality lbl with
+            | Required_or_position_arg -> assert false
+            | Optional_arg mpath ->
+                (match mpath with
```
dkalinichenko-js: Multiple places in the code convert an `optional_module_path` to a type constructor `Path.t` like this. Let's abstract it into its own function.

typing/ctype.ml
dkalinichenko-js: Let's leave a comment explaining why we need `arg_label_equal` here then.

typing/ctype.ml
```diff
@@ -3312,8 +3312,15 @@ let rec expands_to_datatype env ty =
...
+  | _ -> l1 = l2
```
dkalinichenko-js: It's best to write out the equality comparison for other label constructors too, in case we ever add another new constructor.

typing/ctype.ml
```diff
@@ -4538,21 +4545,39 @@ let filter_arrow env t l ~force_tpoly =
...
+              match mpath with
+              | Stdlib_option ->
+                  Predef.path_option, Predef.option_argument_jkind
+              | Stdlib_or_null ->
+                  Predef.path_or_null,
+                    Jkind.for_or_null_argument Predef.ident_or_null
+            in
+            newty2 ~level (Tconstr(t_cons,
+              [newvar2 level arg_jkind], ref Mnil))
+          | _ ->
```
dkalinichenko-js: Write `| Required_or_position_arg` here.

typing/btype.ml
```diff
@@ -624,7 +624,28 @@ let is_optional_parsetree : Parsetree.arg_label -> bool = function
...
+  | Ldot(Lident "Stdlib", "Or_null") -> Stdlib_or_null
+  | _ -> failwith "Only expected Stdlib.Option and Stdlib.Or_null"
+
+type optionality = Optional_arg of optional_module_path
+                 | Required_or_position_arg
+
+let classify_optionality (lbl: Types.arg_label) = match lbl with
+  | Optional _ -> Optional_arg Stdlib_option
+  | Generic_optional(path, _) -> Optional_arg (classify_module_path path.txt)
+  | _ -> Required_or_position_arg
```
dkalinichenko-js: In general, it's best to avoid wildcard patterns in pattern-matches, in case we ever add another constructor. Use `| Nolabel | Labelled _ | Position _` here.

dkalinichenko-js: Same for the rest of this PR.

typing/btype.ml
```diff
@@ -624,7 +624,28 @@ let is_optional_parsetree : Parsetree.arg_label -> bool = function
...
+
+let classify_optionality_parsetree (lbl : Parsetree.arg_label) = match lbl with
+  | Optional _ -> Optional_arg Stdlib_option
+  | Generic_optional(path, _) -> Optional_arg (classify_module_path path.txt)
+  | _ -> Required_or_position_arg
+
+let is_optional arg =
+  not (classify_optionality arg = Required_or_position_arg)
```
dkalinichenko-js: Nit: use an exhaustive pattern-match here.

typing/btype.ml
```diff
@@ -641,16 +662,32 @@ let label_name = function
...
+let prefixed_label_name ppf l =
+  let open Format in
+  match l with
+    Nolabel -> fprintf ppf  ""
+  | Labelled s | Position s -> fprintf ppf "~%s" s
+  | Optional s -> fprintf ppf "?%s" s
+  | Generic_optional (path, s) ->
+      fprintf ppf "%a.?'%s" Pprintast.longident path.txt s
+
+let arg_label_compatible param_label arg_label =
```
dkalinichenko-js, line 677: Nit: rewrite this pattern-match to avoid wildcards.

typing/ctype.ml
```diff
@@ -4538,21 +4545,39 @@ let filter_arrow env t l ~force_tpoly =
...
+
+              {v
+              let type_ident : Longident.t
+                = Ldot (Ldot (path.txt, "Opt_syntax"), "t") in
+              let (path, _) = Env.lookup_type ~loc:path.loc type_ident env in
+              v}
+
+            *)
+            let t_cons, arg_jkind =
+              match mpath with
```
dkalinichenko-js: I think this pattern-match should also be abstracted into a separate function, although it may be harder to justify if this is the only place it comes up.

typing/printtyp.ml
```diff
@@ -1511,12 +1511,21 @@ let rec tree_of_typexp mode alloc_mode ty =
...
               when Path.same path Predef.path_option ->
                 tree_of_typexp mode arg_mode ty
-            | _ -> Otyp_stuff "<hidden>"
+            | Generic_optional (genopt_path, _), Tconstr(path, [ty], _)
+              when Path.same path
+                    (match Btype.classify_module_path genopt_path.txt with
+                    | Stdlib_option -> Predef.path_option
+                    | Stdlib_or_null -> Predef.path_or_null) ->
+                tree_of_typexp mode arg_mode ty
+            (* this case will show up when there is a bug in the compiler
```
dkalinichenko-js: If there is a bug, not when 😅.

Probably best to phrase this as "This case is normally impossible, indicating a compiler bug due to an incorrect type assigned to an optional argument."

typing/printtyp.ml
```diff
@@ -2386,12 +2395,18 @@ let rec tree_of_class_type mode params =
...
+       | Optional _, Tconstr(path, [ty], _)
+         when Path.same path Predef.path_option ->
+           tree_of_typexp mode ty
+       | Generic_optional (genopt_path, _), Tconstr(path, [ty], _)
+         when Path.same path
+                (match Btype.classify_module_path genopt_path.txt with
+                | Stdlib_option -> Predef.path_option
+                | Stdlib_or_null -> Predef.path_or_null) ->
+           tree_of_typexp mode ty
+       | (Optional _ | Generic_optional _), _ -> Otyp_stuff "<hidden>"
```
dkalinichenko-js: Same comment here.

typing/typeclass.ml
```diff
@@ -453,9 +453,15 @@ and class_type_aux env virt self_scope scty =
       in
       let ty = cty.ctyp_type in
       let ty =
-        if Btype.is_optional l
-        then Ctype.newty (Tconstr(Predef.path_option,[ty], ref Mnil))
-        else ty in
+        match Btype.classify_optionality l with
+        | Optional_arg mpath ->
+            let t_cons =
+              match mpath with
```
dkalinichenko-js: As commented, should be abstracted into a separate function.

typing/typeclass.ml
```diff
@@ -1331,26 +1337,36 @@ and class_expr_aux cl_num val_env met_env virt self_scope scl =
         match ty_fun, ty_fun0 with
         | Cty_arrow (l, ty, ty_fun), Cty_arrow (_, ty0, ty_fun0)
           when sargs <> [] ->
-            let name = Btype.label_name l
-            and optional = Btype.is_optional l in
+            (* CR generic-optional:
```
dkalinichenko-js: "CR generic-optional: test generic optional arguments in class methods".

typing/typeclass.ml
```diff
@@ -1331,26 +1337,36 @@ and class_expr_aux cl_num val_env met_env virt self_scope scl =
...
-                  Typecore.type_option_some val_env sarg ty ty0,
-                  (* CR layouts v5: Change the sort when options can hold
-                     non-values. *)
-                  Jkind.Sort.value
+                  (match Btype.classify_optionality l with
+                  | Optional_arg mpath ->
+                      Typecore.type_option_some val_env mpath sarg ty ty0,
+                      (* CR layouts v5: Change the sort when options can hold
+                        non-values. *)
+                      Jkind.Sort.value
```
dkalinichenko-js, line 1351: Change this to `CR generic-optional`, we're going to allow non-values soon.

typing/typeclass.ml
```diff
@@ -1331,26 +1337,36 @@ and class_expr_aux cl_num val_env met_env virt self_scope scl =
...
               )
             in
-            let eliminate_optional_arg () =
-              Arg (Typecore.type_option_none val_env ty0 Location.none,
-                   (* CR layouts v5: Change the sort when options can hold
-                      non-values. *)
-                   Jkind.Sort.value
-                  )
+            let eliminate_optional_arg lbl =
+              (match Btype.classify_optionality lbl with
```
dkalinichenko-js: Nit: I think the parens are unnecessary here.

typing/typeclass.ml
```diff
@@ -1331,26 +1337,36 @@ and class_expr_aux cl_num val_env met_env virt self_scope scl =
...
+              | Optional_arg mpath ->
+                  Arg
+                    (Typecore.type_option_none val_env mpath ty0 Location.none,
+                      (* CR layouts v5: Change the sort when options can hold
+                          non-values. *)
+                      Jkind.Sort.value)
+              | Required_or_position_arg -> assert false)
```
dkalinichenko-js: Annoying to leave an `assert false`, but I think here it's really the least inconvenient option. Feel free to resolve this.

typing/typeclass.ml
```diff
@@ -1592,8 +1615,13 @@ let rec approx_description ct =
     Pcty_arrow (l, core_type, ct) ->
       let l = transl_label l (Some core_type) in
       let arg =
-        if Btype.is_optional l then Ctype.instance var_option
-        else Ctype.newvar (Jkind.Builtin.value ~why:Class_term_argument)
+        match Btype.classify_optionality l with
+        | Optional_arg mpath ->
+            (match mpath with
```
dkalinichenko-js: Duplicate pattern-match, abstract into its own function.

typing/typecore.ml
```diff
@@ -884,23 +884,36 @@ let constant_or_raise env loc cst =
       c
   | Error err -> raise (Error (loc, env, err))
 
+
+let predef_path_of_optional_module_path = function
```
dkalinichenko-js: Ah, I see those now. Should really be defined somewhere earlier and used everywhere, not just in typecore.

typing/typecore.ml
```diff
@@ -884,23 +884,36 @@ let constant_or_raise env loc cst =
...
+
+let predef_jkind_of_optional_module_path = function
+  | Stdlib_option -> Predef.option_argument_jkind
+  | Stdlib_or_null -> Jkind.for_or_null_argument Predef.ident_or_null
+
 (* Specific version of type_option, using newty rather than newgenty *)
 
-let type_option ty =
-  newty (Tconstr(Predef.path_option,[ty], ref Mnil))
+let type_option (mpath : optional_module_path) ty =
```
dkalinichenko-js: Is the annotation necessary here?

typing/typecore.ml
```diff
@@ -3286,9 +3299,12 @@ let type_class_arg_pattern cl_num val_env met_env l spat =
       end;
       List.iter (fun f -> f()) tps.tps_pattern_force;
       (* CR layouts v5: value restriction here to be relaxed *)
```
dkalinichenko-js: CR is unnecessary now.

typing/typecore.ml
```diff
@@ -8001,15 +8026,18 @@ and type_format loc str env =
...
+  let ty0' = extract_option_type env mpath ty0 in
   let alloc_mode, argument_mode = register_allocation expected_mode in
   let arg = type_argument ~overwrite:No_overwrite env argument_mode sarg ty' ty0' in
-  let lid = Longident.Lident "Some" in
-  let csome = Env.find_ident_constructor Predef.ident_some env in
+  let lid, csome = (match mpath with
+    | Stdlib_option -> (Longident.Lident "Some",
+      Env.find_ident_constructor Predef.ident_some env)
+    | Stdlib_or_null -> (Longident.Lident "This",
+      Env.find_ident_constructor Predef.ident_this env)) in
```
dkalinichenko-js: Nit: `in` in multi-line let expressions should be on its own line, with the same indentation as `let`.

typing/typetexp.ml
```diff
@@ -671,23 +671,21 @@ let transl_label (label : Parsetree.arg_label)
...
-      match mod_ident with
-      | { txt = Longident.Ldot (Longident.Lident "Stdlib", "Option")
-        ; _ } -> Optional l
-      | { loc ; txt } ->  (
-        raise
-          (Error (loc,
-                  Env.empty,
-                  Invalid_generic_optional_argument_module_path txt))
-      )
+        if path.txt = Longident.Ldot (Lident "Stdlib", "Option") ||
```
dkalinichenko-js, line 683: Leave a CR to change this here.

typing/ctype.ml
zchenb: There will still be a catch all case at the end, for unequal things. I am not sure this is better than `_ -> l1 = l2`:

```
let arg_label_equal l1 l2 =
  match l1, l2 with
  | Generic_optional(path1, str1), Generic_optional(path2, str2) ->
      str1 = str2 &&
        classify_module_path path1.txt = classify_module_path path2.txt
  | Optional str1, Optional str2
  | Position str1, Position str2
  | Labelled str1, Labelled str2 -> str1 = str2
  | Nolabel, Nolabel -> true
  | _ -> false
```

Please comment.

typing/btype.ml
zchenb: I am not sure how to do this properly. Let's discuss this in person.

typing/ctype.ml
zchenb: Fixed.

typing/typecore.ml
zchenb: Indeed not necessary once we got rid of pattern matching.

typing/typecore.ml
zchenb: I removed this CR but I don't understand why this is unnecessary. I need some explanation.

typing/btype.ml
zchenb: I made changes for a majority of the cases. However, there are still some cases where I think `_` makes more sense. We should look at the remaining cases and decide.

typing/ctype.ml
dkalinichenko-js: We use `| (Generic_optional _ | Optional _ | Position _ | Labelled _ | Nolabel), _ -> false` in such cases. The invariant to maintain is that adding a new variant case should cause a compiler warning/error.

typing/typecore.ml
dkalinichenko-js: I believe "value restriction" here refers to `Predef.option_argument_jkind`, which is `value`. Now we can be generic over jkinds.

testsuite/tests/generic-optional-arguments/calling_compatibility.ml
```diff
@@ -0,0 +1,221 @@
+(* TEST
```
dkalinichenko-js, line 1: Since we touched code handling function argument order, you should add a bunch of tests for generic optional parameters behaving sanely in functions of many arguments of multiple kinds (positional/labeled/optional).

dkalinichenko-js: Not necessarily in this file.

testsuite/tests/generic-optional-arguments/test.ml
```diff
@@ -25,7 +29,11 @@ let chain_call Stdlib.Option.?'(sep : string option) arg =
   M.concat Stdlib.Option.?'sep arg
 
 let chain_call_2 Stdlib.Option.?'(sep) arg =
-  String.concat (match sep with None -> " " | Some s -> s) arg
+  M.concat_2 Stdlib.Or_null.?'sep:(
+    match sep with
+    | None -> Null
+    | Some x -> This x
+  ) arg
```
dkalinichenko-js, line 37: I'd also like some more testing for module inclusion checks, both between 2 signatures and between signatures and structures, with multiple arguments.

dkalinichenko-js: It also would be good to add CRs for testing objects/classes, or even make a separate test file with just CRs.

typing/ctype.ml
```diff
@@ -4532,27 +4541,49 @@ type filtered_arrow =
     ret_mode : Mode.Alloc.lr
   }
 
+let predef_path_of_optional_module_path = function
```
dkalinichenko-js, line 4550: Nit: `predef_` prefix is unnecessary here, especially since we are going to soon use those functions for non-predef paths or jkinds.

typing/typeclass.ml
```diff
@@ -453,9 +453,11 @@ and class_type_aux env virt self_scope scty =
...
       let ty = cty.ctyp_type in
       let ty =
-        if Btype.is_optional l
-        then Ctype.newty (Tconstr(Predef.path_option,[ty], ref Mnil))
-        else ty in
+        match Btype.classify_optionality l with
+        | Optional_arg mpath ->
+            let t_cons = Ctype.predef_path_of_optional_module_path mpath in
+            Ctype.newty (Tconstr(t_cons,[ty], ref Mnil))
+        | Required_or_position_arg -> ty in
```
dkalinichenko-js: Nit: move `in` to a new line.

typing/typecore.ml
```diff
@@ -884,23 +884,28 @@ let constant_or_raise env loc cst =
       c
   | Error err -> raise (Error (loc, env, err))
 
+
```
dkalinichenko-js: Nit: unnecessary whitespace.

typing/typecore.ml
```diff
@@ -4157,7 +4164,9 @@ let collect_apply_args env funct ignore_labels ty_fun ty_fun0 mode_fun sargs ret
             match sargs with
             | [] -> assert false
             | (l', sarg) :: remaining_sargs ->
-                if name = label_name l' || (not omittable && l' = Nolabel) then
+                if (arg_label_compatible l l')
```
dkalinichenko-js: Nit: I think parens are unnecessary here.

typing/typecore.ml
```diff
@@ -7466,20 +7483,19 @@ and type_function
...
-              match arg_label with
-              | Optional arg_label -> arg_label
-              (* CR generic-optional : CHECK THIS *)
-              | Generic_optional (_, arg_label) -> arg_label
-              | Nolabel | Labelled _ ->
+            let arg_label, mpath =
+              match arg_label, classify_optionality_parsetree arg_label with
+              | (Optional arg_label | Generic_optional (_, arg_label)),
+                 Optional_arg mpath -> arg_label, mpath
+              | _ ->
```
dkalinichenko-js: I think here it's still best to write out all possible cases.

typing/typecore.ml
```diff
@@ -8405,18 +8431,16 @@ and type_apply_arg env ~app_loc ~funct ~index ~position_and_mode ~partial_app (l
...
+          let arg = type_option_none env mpath
+                      (instance ty_arg) Location.none
+          in
           (lbl, Arg (arg, Mode.Value.legacy, sort_arg))
-      | Position _ ->
+      | Position _, Required_or_position_arg ->
           let arg = src_pos (Location.ghostify funct.exp_loc) [] env in
           (lbl, Arg (arg, Mode.Value.legacy, sort_arg))
-      | Labelled _ | Nolabel -> assert false)
+      | _ -> assert false)
```
dkalinichenko-js: Also best to write out all cases.

dkalinichenko-js (commented):
This mostly looks good now, but needs more tests.

typing/ctype.ml
zchenb: I left a CR to remove the prefix when we add non-predef paths. Let's remove this in a child feature.

zchenb: Renaming here would also introduce merge conflicts with child PRs.

testsuite/tests/generic-optional-arguments/calling_compatibility.ml
zchenb: I think this testing is a bit more involved. I have created a CR file. Can we move the actual creation of tests to a child PR?

testsuite/tests/generic-optional-arguments/test.ml
zchenb: Same here. I agree we need more comprehensive testing. Can we do this in a child PR? [I have added an empty CR-only testing file].

testsuite/tests/generic-optional-arguments/calling_compatibility.ml
dkalinichenko-js: Sure, this works.

testsuite/tests/generic-optional-arguments/test.ml
dkalinichenko-js: Sure.

typing/ctype.ml
dkalinichenko-js: Ok.

dkalinichenko-js (approved):

PR #4303, opened 2025-07-14
Add Partial Support for External Mode

alpha-convert:
This PR makes the following changes on the way to supporting an Externality mode.

1. It changes the externality jkind axis from a non-modal one to a modal one, implementing the requisite lattice operations for a modal axis.
2. It adds a new minimum point Byte_external to the axis, for values that are safe to hide from the GC, even in bytecode-compiled code.
3. It ensures that all allocation sites have externality at least Internal by modifying register_allocation to constrain the mode of allocated values.
4. Adds some tests for the externality mode checking, and promotes changes in some other tests

### DIFF
--- a/typing/typecore.ml
+++ b/typing/typecore.ml
@@ -8058,7 +8058,8 @@ and type_label_exp
     overwrite:_ -> create:_ -> _ -> _ -> _ -> _ ->
     _ * rep gen_label_description * _ -> rep record_form ->
     _ * rep gen_label_description * _
-  = fun ~overwrite ~create env arg_mode loc ty_expected (lid, label, sarg) record_form ->
+  = fun ~overwrite ~create env arg_mode loc ty_expected
+        (lid, label, sarg) record_form ->
   (* Here also ty_expected may be at generic_level *)
   let separate = !Clflags.principal || Env.has_local_constraints env in
   (* #4682: we try two type-checking approaches for [arg] using backtracking:


jane/doc/extensions/_05-modes/intro.md
```diff
@@ -69,6 +69,7 @@ programs.
 * [Modes for moving between threads (portability and
   contention)](#portability-contention)
 * [Modes for aliasing (uniqueness and linearity)](#uniqueness-linearity)
+* [Modes ] (#externality)
```
ccasin:
```suggestion
* [Modes for garbage collection] (#externality)
```

jane/doc/extensions/_05-modes/intro.md
```diff
@@ -277,3 +278,43 @@ while observing closures capture all values at visibility *read*.
...
+
+|--------------|
+| **internal** |
+| `|`          |
+| external64   |
+| `|`          |
+| external_    |
+{: .table}
+
+Externality is a future axis wich records whether a value may safely be ignored
```
ccasin:
```suggestion
Externality is a future axis which records whether a value may safely be ignored
```

jane/doc/extensions/_05-modes/intro.md
```diff
@@ -277,3 +278,43 @@ while observing closures capture all values at visibility *read*.
...
+by the GC.  This may be because they are OCaml "immediates" (values represented
+by a tagged integer), because they are unboxed types like `float#` or `int32#`,
+or because they are allocated elsewhere.
+
+The compiler uses the externality axis for certain runtime optimizations. In
+particular, updating a mutable reference to a value that is `external_` can skip
+the write barrier (i.e., it does not need a call to `caml_modify`).
+
+The axis has three possible values, with `external_ < external64 < internal`.
+* `external_` means that the value may safely ignored by the
```
ccasin:
```suggestion
* `external_` means that the value may be safely ignored by the
```

jane/doc/extensions/_05-modes/intro.md
```diff
@@ -277,3 +278,43 @@ while observing closures capture all values at visibility *read*.
...
+  types like `float#`, or other values represented by a tagged integer like `None`
+  or `[]`.
+* `external64` means the value can be safely ignored by the GC
+  _on 64-bit systems_. The only 32-bit target currently supported by the OxCaml
+  compiler is bytecode. Note that, although JavaScript and WASM are 32-bit
+  platforms and the compiler goes through bytecode to reach them, they still
+  count as 64-bit systems for the purpose of this axis because of their unique
+  data models.
+* `internal` means values of the type may need to be scanned. Examples of internal
+  values include those that
```
ccasin: This sentence is missing its conclusion, but I think I would say instead something like "This includes all standard OCaml values that are represented by blocks and managed by the OCaml GC (like lists and records)."

jane/doc/extensions/_06-kinds/intro.md
```diff
@@ -69,12 +69,11 @@ the bounds:
 
 ```
 value mod global contended portable aliased many unyielding immutable stateless
-          non_float external_
+          non_float
```
ccasin: I think `int` still does have an `external_` bound in its kind (well, technically `byte_external`, but I agree with the choice to not acknowledge `byte_external` in the docs). That's now a modal axis rather than a non-modal axis, but it should still appear here (though you could move it before `non_float` if you like).

jane/doc/extensions/_06-kinds/intro.md
```diff
@@ -69,12 +69,11 @@ the bounds:
...
 ```
 
 This kind indicates that `int` mode crosses on all eight of our modal axes. In
-addition, `int`s are not `float`s and they do not need to be garbage-collected
-(they are `external_` to the garbage collector).
+addition, `int`s are not `float`.
```
ccasin: As above, probably keep this.

testsuite/tests/typing-jkind-bounds/basics.ml
```diff
@@ -216,7 +216,7 @@ Error: The layout of type "a" is value
 type a : value mod global aliased many immutable stateless external_ unyielding non_float
 type b : value mod local unique once contended nonportable internal = a
 [%%expect{|
-type a : immediate
+type a : immediate64 mod external_
```
ccasin: Hmm.  I guess this is printing this way because `immediate` is `mod byte_external`?  I think we should hack the printing code to print both `immediate` and this kind as `immediate`, as the goal is to hide the difference from the user (it exists just so we don't skip needed caml_modifys when compiling to bytecode).

testsuite/tests/typing-jkind-bounds/basics.ml
```diff
@@ -301,7 +307,7 @@ type d : float64 = c
 [%%expect{|
 type a = float#
 type b = a
-type c : float64 mod everything
+type c : float64 mod global aliased many stateless immutable external_
```
ccasin: As above, I suppose this doesn't print as `mod everything` any more because `mod everything` actually means `byte_external`.  But I think we should lie and this kind as `float64 mod everything` anyway.

Also, we should probably arrange it so that if you _write_ `immediate` or `mod everything` in your program, we understand that to mean the `external` version, rather than the` byte_external` version.  Otherwise, people will be surprised when they write `type t : float64 mod everything = float#` and it fails to typecheck.

testsuite/tests/typing-jkind-bounds/basics.ml
```diff
@@ -482,13 +488,7 @@ Error: The kind of type "t_value" is value
...
 type t : any mod portable = t_value
 [%%expect{|
-Line 1, characters 0-35:
-1 | type t : any mod portable = t_value
-    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-Error: The kind of type "t_value" is value
-         because of the definition of t_value at line 1, characters 0-20.
-       But the kind of type "t_value" must be a subkind of any mod portable
-         because of the definition of t at line 1, characters 0-35.
+type t = t_value
```
ccasin: This looks like a bug - I don't understand what would have changed here.

testsuite/tests/typing-modes/externality.ml
```diff
@@ -0,0 +1,151 @@
+(* TEST
+ expect;
+*)
+
+
+let immediate_is_external () =
```
ccasin, line 21: I think we should have similar tests for `byte_external`, showing e.g., that `3` is `byte_external` and `#3.14` is not.  We should arrange things so that you can type `byte_external` but it still prints as `external`.  This will mean the error messages are sad - we can think a little about whether it's easy to print `byte_exernal` explicitly when it's relevant in the error, but I suspect it's not, and since its existence is temporary and we don't plain to expose it to users, we'll just accept the sad error.

testsuite/tests/typing-modes/externality.ml
```diff
@@ -0,0 +1,151 @@
...
+Line 3, characters 36-37:
+3 |   let bar (x : 'a @ internal) = foo x in
+                                        ^
+Error: This value is "internal" but expected to be "external64".
+|}]
+
+
+
+(*
+   CR jcutler: The code below is the expected behavior. For future axes where
```
ccasin: This comment is definitely helpful, but I don't think it needs to be a CR, as there's nothing we expect to change here in the future.

testsuite/tests/typing-modes/externality.ml
```diff
@@ -0,0 +1,151 @@
...
+type 'a t = {mutable f : 'a}
+let foo (t : 'a t @ external_) (x : 'a @ internal) =
+  t.f <- x
+[%%expect{|
+type 'a t = { mutable f : 'a; }
+val foo : 'a t -> 'a -> unit = <fun>
+|}]
+
+(* CR jcutler: In the above type's case, creating an @external_ record of this type
+   should be disallowed. *)
```
ccasin: Two suggestions:
* I think it's useful if a CR has a slightly more actionable instruction. What do I do to address this CR? In this case, I think you left the CR to suggest such a test should be written, so I'd probably have added a second sentence that say "Write a test demonstrating this."
* In this specific case, rather than just leaving a CR to write such a test, I'd write the test now, and leave a CR to fix it if it doesn't have the right output. (Though I think we should also probably fix it to work correctly in this PR, and write similar tests for the ref and array cases below)

testsuite/tests/typing-modes/externality.ml
```diff
@@ -0,0 +1,151 @@
...
+let into_external () =
+  let _ : int option external_ = {ext = Some 3} in
+  ()
+[%%expect {|
+type 'a external_ = { ext : 'a @@ external_; }
+Line 3, characters 40-46:
+3 |   let _ : int option external_ = {ext = Some 3} in
+                                            ^^^^^^
+Error: This value is "internal" but expected to be "external_".
+|}]
```
ccasin, line 191: It would be good to have a test here that the above does work if the initial value is in fact external (either because it mode crosses on the externality axis, or because it's another argument we've taken `@ external`).

Also to test the CR just above.

testsuite/tests/typing-modes/externality.ml
```diff
@@ -0,0 +1,151 @@
...
+  ()
+
+[%%expect{|
+Line 3, characters 36-37:
+3 |   let bar (x : 'a @ internal) = foo x in
+                                        ^
+Error: This value is "internal" but expected to be "external64".
+|}]
+
+
```
ccasin, line 133: It would be good to have some tests about the deepness of external, which is hard for boxed datastructures because the boxes themselves make them internal, but we can probably test a bit in unboxed data structures.  e.g.:
```
let f (x : string @ internal) : #(string * string) @ external_ = #(x, x) (* should be rejected *)
let f (x : string @ external_) : #(string * string) @ external_ = #(x, x) (* should be allowed *)
let f (x : int @ internal) : #(int * int) @ external_ = #(x, x) (* also allowed *)
```

And probably you can come up with more (e.g., unboxed records with modalities).

ccasin: Another test case for deepness that I don't think we can write now at all, but you might leave a comment about: when we support external allocations of functions, we'll need to check they don't close over anything `internal`.

testsuite/tests/typing-unboxed-types/test.ml
```diff
@@ -549,7 +549,13 @@ type 'a t [@@immediate];;
...
 [%%expect{|
 type 'a t : immediate
-type u = U : 'a t -> u [@@unboxed]
+Line 2, characters 0-34:
+2 | type u = U : 'a t -> u [@@unboxed];;
+    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Error: This type cannot be unboxed because
+       it might contain both float and non-float values,
+       depending on the instantiation of the existential variable "'a".
+       You should annotate it with "[@@ocaml.boxed]".
```
ccasin, line 559: I don't understand why this one changed - I think it suggests we changed the separability of `immediate` somehow?

typing/ctype.ml
```diff
@@ -2577,14 +2577,6 @@ let constrain_type_jkind env ty jkind =
...
   Env.constrain_type_jkind := constrain_type_jkind
 
-let check_type_externality env ty ext =
-  let upper_bound =
-    Jkind.set_externality_upper_bound (Jkind.Builtin.any ~why:Dummy_jkind) ext
-  in
-  match check_type_jkind env ty upper_bound with
-  | Ok () -> true
-  | Error _ -> false
-
```
ccasin: It looks like you've inlined this function at its call-sites in `typeopt` and `vicuna_traverse_typed_tree`, but I think it would be better to leave it here - `ctype` is morally the right place for it and it's next to other similar functions (like check_type_nullability, below).

It would be cool if some uses of this could change from checking the kind of a type to instead checking the mode of an expression, and then possibly this even goes away entirely, but until then I think leaving it here is the right thing, unless there's some complication I haven't understood.

typing/env.ml
```diff
@@ -4767,6 +4767,7 @@ let report_lookup_error _loc env ppf = function
         | Error (Statefulness, {left; right}) ->
           asprintf "%a" Mode.Statefulness.Const.print left,
           asprintf "is %a" Mode.Statefulness.Const.print right
+        | Error (Externality, _) -> failwith "unimplemented" (* CR jcutler: fixme *)
```
ccasin: I think this error is only reachable once we have external closures (but let's ask Zesen next week)

typing/jkind.ml
```diff
@@ -1379,7 +1379,8 @@ module Const = struct
                 ~uniqueness:Uniqueness.Const_op.max
                 ~contention:Contention.Const_op.min
                 ~statefulness:Statefulness.Const.min
-                ~visibility:Visibility.Const_op.min ~externality:Externality.max
+                ~visibility:Visibility.Const_op.min
+                ~externality:Mod_bounds.Externality.max
```
ccasin, line 1387: While it amounts to the same thing (I think), this should probably be `Externality.Const.max`, to match the other comonadic cases (and similarly in several other spots - just search for `Mod_bounds.Externality`)

typing/jkind.ml
```diff
@@ -2416,8 +2419,27 @@ let for_non_float ~(why : History.value_creation_reason) =
...
+      { parent_path = Path.Pident ident; position = 1; arity = 1 }
+  in
+  let mod_bounds =
+    Mod_bounds.create ~locality:Locality.Const.max
+      ~linearity:Linearity.Const.max ~portability:Portability.Const.max
+      ~yielding:Yielding.Const.max ~uniqueness:Uniqueness.Const_op.max
+      ~contention:Contention.Const_op.max ~statefulness:Statefulness.Const.max
+      ~visibility:Visibility.Const_op.max ~externality:Externality.Const.max
+      ~nullability:Nullability.Non_null
+      ~separability:Separability.Maybe_separable
```
ccasin, line 2455: I think the above definition should't be here and is probably the result of resolving a merge conflict incorrectly.  It looks like the code comes from #4298, and probably got picked up as we were trying to rebase onto the correct commit.

typing/jkind.ml
```diff
@@ -2649,16 +2679,17 @@ let to_unsafe_mode_crossing jkind =
   }
 
 let all_except_externality =
-  Axis_set.singleton (Nonmodal Externality) |> Axis_set.complement
+  Axis_set.singleton (Modal (Comonadic Externality)) |> Axis_set.complement
 
+(* CR jcutler: can we get rid of this? *)
```
ccasin: I think we could get rid of it if we replaced every check that a type crosses the externality axis with a check that some particular value is `external`, but I suspect there are som places the former is actually what we want, or is hard to get rid of without a bigger refactoring.

typing/jkind.mli
```diff
@@ -530,6 +530,9 @@ val for_float : Ident.t -> Types.jkind_l
 (** The jkind for values that are not floats. *)
 val for_non_float : why:History.value_creation_reason -> 'd Types.jkind
 
+(** The jkind for [or_null] type arguments. *)
+val for_or_null_argument : Ident.t -> 'd Types.jkind
+
```
ccasin: (Another diff from merge conflict resolution that should be removed, I think)

typing/jkind.mli
```diff
@@ -610,15 +613,17 @@ val get_mode_crossing :
 
 val to_unsafe_mode_crossing : Types.jkind_l -> Types.unsafe_mode_crossing
 
+(* CR jcutler: these two functions should probably go away, since this
+   is now covered by the existing functiosn to get and set the axis.*)
```
ccasin: I'm not actually sure any such functions exist for kinds except these - probably fine to leave them.

typing/mode.ml
```diff
@@ -45,7 +45,7 @@ module type CoHeyting = sig
   val subtract : t -> t -> t
 end
 
-(* Even though our lattices are all bi-heyting algebras, that knowledge is
+(* Even thoudgh our lattices are all bi-heyting algebras, that knowledge is
```
ccasin: typo

typing/mode.ml
```diff
@@ -491,6 +491,80 @@ module Lattices = struct
...
+        | External, Byte_external -> Byte_external
+        | External, (External | External64 | Internal) -> External
+        | External64, (External | Byte_external | External64) -> b
+        | External64, Internal -> External64
+        | Internal, (Byte_external | External | External64) -> b
+        | Internal, Internal -> Internal
+
+      let print ppf = function
+        | Byte_external ->
+          Format.fprintf ppf "external_" (* CR jcutler: do we want this??*)
```
ccasin: I think we do want this, but I think it would be good to write a comment somewhere explaining what we're doing with `byte_external` (I can help) and refer to that here and in other places where our hackery comes up.

typing/mode_intf.mli
```diff
@@ -380,6 +406,8 @@ module type S = sig
       | Linearity : ('areality comonadic_with, Linearity.Const.t) t
       | Statefulness : ('areality comonadic_with, Statefulness.Const.t) t
       | Portability : ('areality comonadic_with, Portability.Const.t) t
+      (* CR jcutler: is this in the right order?? *)
```
ccasin: `externality` doesn't participate in any implications, so this is fine (as would be any position, I think)

typing/typecore.ml
```diff
@@ -696,22 +696,34 @@ let allocations : Alloc.r list ref = Local_store.s_ref []
...
 
-let register_allocation_mode alloc_mode =
+let register_allocation_mode ~loc ~env alloc_mode =
+  let externality = Alloc.proj (Comonadic Externality) alloc_mode in
+  let res = Externality.submode Externality.internal externality in
+  (match res with
+  | Ok () -> ()
+  | Error failure_reason ->
+    (* CR jcutler: do we need to pass values for these contexts? I don't think
+       they have anything to do with the type error we're raising here... *)
```
ccasin: I agree this is fine.  We can consider adding an externality context later if it seems helpful for errors.

typing/typecore.ml
```diff
@@ -6320,7 +6333,8 @@ and type_expect_
           submode ~loc:record.exp_loc ~env rmode mode_mutate_mutable;
           let mode = mutable_mode m0 |> mode_default in
           let mode = mode_modality label.lbl_modalities mode in
-          type_label_exp ~overwrite:No_overwrite_label false env mode loc ty_record
+          (* This is the mode of the field we're updating. Should be internal. *)
```
ccasin: I'm not sure this comment is accurate - might it not be `external` if I've written an external modality?

Separately, I think `mutable_mode` and `mode_modality` end up having an affect on several axes, and calling out externality alone in a comment could be confusing here - I would probably delete this comment (and consider fleshing out comments on those functions) or more fully describe the situation.

typing/typecore.ml
```diff
@@ -7297,6 +7311,7 @@ and type_ident env ?(recarg=Rejected) lid =
   where [m] is the mode of a closure lock. Since join is commutative and
   associative, the order of which we apply those join does not matter.
   *)
+  (* CR jcutler: ensure this argument holds *)
```
ccasin, line 7338: We should discuss with Zesen, but I think we have a problem here, once we have external closures.  If we allow such a closure to hold a reference to an internal module through which we reference some external value, that module might be freed by the GC before we use it.

typing/typecore.ml
```diff
@@ -8013,16 +8028,17 @@ and type_option_some env expected_mode sarg ty ty0 =
...
     _ * rep gen_label_description * _
-  = fun ~overwrite create env arg_mode loc ty_expected (lid, label, sarg) record_form ->
+  = fun ~overwrite ~create env arg_mode loc ty_expected (lid, label, sarg) record_form ->
   (* Here also ty_expected may be at generic_level *)
   let separate = !Clflags.principal || Env.has_local_constraints env in
   (* #4682: we try two type-checking approaches for [arg] using backtracking:
      - first try: we try with [ty_arg] as expected type;
      - second try; if that fails, we backtrack and try without
   *)
+    (* CR jcutler: ty_Arg is the expected? real? type of the argument *)
```
ccasin: I believe `ty_arg` is the actual type of the argument (or at least as much as we know about it at this point in inference)

(I think this is unrelated to this PR, though it's a fine question)

typing/typecore.ml
```diff
@@ -11165,6 +11181,7 @@ let report_error ~loc env =
         | Error (Comonadic Portability, _ ) -> []
         | Error (Comonadic Yielding, _) -> []
         | Error (Comonadic Statefulness, _) -> []
+        | Error (Comonadic Externality, _) -> [] (* CR jcutler: fixme *)
```
ccasin: This seems fine - we can do more here in the future if it seems useful for error message.

typing/typedecl_separability.ml
```diff
@@ -485,13 +485,8 @@ let msig_of_external_type env decl =
...
   in
-  let jkind_of_type = Ctype.type_jkind_purely_if_principal env in
-  let is_external =
-    match Jkind.get_externality_upper_bound ~jkind_of_type decl.type_jkind with
-    | Internal -> false
-    | External | External64 -> true
-  in
-  if is_not_value_or_null || is_external then best_msig decl else worst_msig decl
+  (* CR jcutler: this is now slightly more pessimistic. Unclaer if it's the right choice. *)
+  if is_not_value_or_null then best_msig decl else worst_msig decl
```
ccasin, line 490: I agree this is tricky - we can now indeed have `external` types that are not separable.  That's potentially a problem for us: lots of code in the Jane Street codebase surely relies on "immediates" being separable.  But there's hope - we should discuss with @dkalinichenko-js , but I think we should be able to replace this with a check on the new separability axis, which I think will catch all the cases this check used to catch.

typing/typemode.ml
```diff
@@ -116,7 +118,7 @@ module Transled_modifiers = struct
       yielding : Mode.Yielding.Const.t Location.loc option;
       statefulness : Mode.Statefulness.Const.t Location.loc option;
       visibility : Mode.Visibility.Const.t Location.loc option;
-      externality : Jkind_axis.Externality.t Location.loc option;
+      externality_mod : Mode.Externality.Const.t Location.loc option;
```
ccasin: I think we can rename this back to `externality` now.

typing/typemode.ml
```diff
@@ -456,7 +461,24 @@ let untransl_modality (a : Modality.t) : Parsetree.modality loc =
     | Atom (Monadic Visibility, Join_with Visibility.Const.Read) -> "read"
     | Atom (Monadic Visibility, Join_with Visibility.Const.Read_write) ->
       "read_write"
-    | _ -> failwith "BUG: impossible modality atom"
+    | Atom
+        (Comonadic Externality, Meet_with Mode.Externality.Const.Byte_external)
+      ->
+      "external_"
+    (* CR jcutler: think about correct printing of byte external, do we
```
ccasin: I think this is the right thing for now.  I think I left a comment elsewhere saying we should have a detailed comment that explains the situation and point to it from the relevant places - this is one of those places.

typing/typemode.ml
```diff
@@ -465,13 +487,15 @@ let untransl_modality (a : Modality.t) : Parsetree.modality loc =
    removed. The implications on the monadic axes will stay. Implied modalities
    can be overriden. *)
 (* CR zqian: decouple mutable and comonadic modalities *)
+(* CR jcutler: why is this? Why aliased? Can we get away without *)
```
ccasin: Thanks for leaving this comment to remind us to come back here.  I think we understand the situation well enough now, and have seen how it plays out in `use` and `free`, that we can drop the CR.

typing/typeopt.ml
```diff
@@ -400,7 +405,7 @@ let value_kind_of_value_jkind env jkind =
   let jkind_of_type ty = Some (Ctype.type_jkind_purely env ty) in
   let externality_upper_bound = Jkind.get_externality_upper_bound ~jkind_of_type jkind in
   match layout, externality_upper_bound with
-  | Base Value, External -> Pintval
+  | Base Value, (External | Byte_external) -> Pintval
```
ccasin: This is tricky - I think this is, at least in part, what controls whether we're going to get `caml_modify`s later.  This looks correct for native code, but probably we should check here if we're targeting native code (`!Clflags.native_code`) and, if not, map `Byte_external` to `Pintval` but `external` to `Pgenval`.  We can also possibly write a test case - let's discuss.

typing/types.ml
```diff
@@ -230,10 +230,11 @@ module Jkind_mod_bounds = struct
       else t.visibility
     in
     let externality =
-      if mem min_axes (Nonmodal Externality)
-      then Externality.min
+      if mem min_axes (Modal (Comonadic Externality))
+      then Externality.max
```
ccasin: Do we want `min` here?

typing/types.ml
```diff
@@ -230,10 +230,11 @@ module Jkind_mod_bounds = struct
...
       else t.externality
     in
+
```
ccasin: stray newline

typing/types.ml
```diff
@@ -292,7 +293,7 @@ module Jkind_mod_bounds = struct
         yielding = Yielding;
         statefulness = Stateful;
         visibility = Read_write;
-        externality = External;
+        externality = Internal;
```
ccasin: Wild - I guess this is just a live bug?

typing/typetexp.ml
```diff
@@ -779,6 +779,7 @@ and transl_type_aux env ~row_context ~aliased ~policy mode styp =
                 (newconstr Predef.path_option [Btype.tpoly_get_mono arg_ty])
             end
           in
+
```
ccasin: stray newline

ccasin (commented):
A bunch of mostly minor comments, but this is exciting!

typing/types.ml
alpha-convert: Seems like it! I had originally implemented it in the same buggy way, but I actually managed to trigger the bug with a test case (not sure which one).

testsuite/tests/typing-unboxed-types/test.ml
alpha-convert: Yes — this is because of [this imprecision](https://github.com/oxcaml/oxcaml/pull/4303#discussion_r2208946151), I believe.

testsuite/tests/typing-modes/externality.ml
alpha-convert: As discussed earlier today, constants get lifted so it's sort of hard to show that `#3.14` is not `byte_external`. Any ideas for how to show this otherwise?

typing/typedecl_separability.ml
dkalinichenko-js: The separability axis check also has a problem: it may fail in public release. I think we should replace this with some sort of "is immediate" check.

testsuite/tests/typing-modes/externality.ml
ccasin: Sure - I think a more complex constant like `#3.14 + #1.0` or `#(3, 3)` might work, but if that fails you can write a test like:
```ocaml
let add_pi (x : float# @ byte_external) : float# @ byte_external = Float_u.add (#3.14 + x)
```
Which I think should fail, while a similar test for ints will pass.

(See [this test](https://github.com/oxcaml/oxcaml/blob/main/testsuite/tests/typing-layouts-float64/stdlib_float_u.ml) for an example of how to update your test stanza to give you access to `Stdlib_upstream_compatible`, which has the `Float_u` module.)

testsuite/tests/typing-unboxed-types/test.ml
ccasin: Ack, per Diana's comment elsewhere I think we need to fix this by making some core changes to the jkind system (adding a non_pointer layout).  We're not going to do that in this PR, so I think you should leave a CR here and also where that change was made.  The CR should be clear that we can't actually merge this into the compiler until we fix that.

testsuite/tests/typing-modes/externality.ml
```diff
@@ -149,3 +167,76 @@ Line 3, characters 40-46:
...
+[%%expect {|
+type t = #{ x : int; y : string; }
+val f : int -> string -> t = <fun>
+|}]
+
+(* CR jcutler: When we support externally allocating functions, write a test to show that
+You can only allocate functions that don't close over internals.
+*)
+
+(* Byte externality vs externality *)
```
ccasin: I think it would be good to leave a comment here noting that this error message is bad by design.  We expect users never to write `byte_external` and indeed not to be aware of is existence.  And the badness will go away in the glorious future where `byte_external` is deleted.

I'd also like to test that unboxed products of `byte_external` things (`#(3, 4)`) do not get treated as `byte_external`.

typing/ctype.ml
```diff
@@ -2577,6 +2577,17 @@ let constrain_type_jkind env ty jkind =
...
+  let upper_bound =
+    Jkind.set_externality_upper_bound
+      (Jkind.Builtin.any ~why:Dummy_jkind)
+      externality
+  in
+  match check_type_jkind env ty upper_bound with
+  | Ok () -> true
+  | Error _ -> false
+
+
```
ccasin: extra newline

typing/ctype.mli
```diff
@@ -659,6 +659,12 @@ val check_type_jkind :
 val constrain_type_jkind :
   Env.t -> type_expr -> ('l * allowed) jkind -> (unit, Jkind.Violation.t) result
 
+  (* Check whether a type's externality's upper bound is less than some target.
```
ccasin: It looks like the first line of this comment got indented by two extra spaces when it was restored.  In general it's good to take care that the full PR doesn't have any unecessary diffs, as they create noise when trying to look through the history (e.g., via git blame).

typing/jkind.ml
```diff
@@ -1305,14 +1305,19 @@ module Const = struct
         name : string
       }
 
-    let mk_jkind ~mode_crossing ~nullability ~separability
+    let mk_jkind ?externality ~mode_crossing ~nullability ~separability
```
ccasin: Optional arguments are tempting, but mostly should be avoided.  The convenience of having to write the argument less often is ~always outweighed by the risk someone forgets to write it when they need it.  Let's make this a normal named argument (and then probably move it next to nullability and separability).

typing/jkind.ml
```diff
@@ -1493,6 +1498,7 @@ module Const = struct
 
     (* CR layouts v3: change to [Maybe_null] when
        [or_null array]s are implemented. *)
+    (* CR jcutler: ensure all the kind/kind_of_unboxed pairs have this pattern.*)
```
ccasin: This sounds like something we should do in this PR, rather than leaving a CR for.

typing/typedecl_separability.ml
ccasin: We've resolved this for now by adding a CR noting we can't merge this until we have a better resolution.  Our plan is to add a new layout, or a new mode, to track this, and is being discussed elsewhere.

typing/env.ml
ccasin: Per discussion with Zesen, this is indeed unreachable now, but will be reachable in the next PR - we should update the comment to clarify this.

Seperately, as a matter of convention, we tend to use "failwith" or "assert false" for things that are obviously unreachable as a matter of some very local check within the function, and otherwise use "Misc.fatal_error" with text that prints the name of the module and function we're in for things that are unreachable due to a broader compiler invariant that we might be wrong about (you can search the compiler codebase for some examples).  This can be helpful for debugging when we get such invariants wrong.  This seems like the latter case.

ccasin (approved):
Looks good - just a few more minor things to address before merging.

PR #4309, opened 2025-07-15
Annotate non-thread-safe `Unix` functions as `nonportable`

polytypic:
This annotates a number of functions in the unix library as `nonportable`.  This includes functions that access the environment or the user database and others.


otherlibs/unix/unix.mli
```diff
@@ -232,19 +232,19 @@ val execv : string -> string array -> 'a
...
-val execvpe : string -> string array -> string array -> 'a
+val execvpe : string -> string array -> string array -> 'a @@ nonportable
 (** Same as {!execve}, except that
    the program is searched in the path. *)
 
-val fork : unit -> int
+val fork : unit -> int @@ nonportable
 (** Fork a new process. The returned integer is 0 for the child
    process, the pid of the child process for the parent process. It
    fails if the OCaml process is multi-core (any domain has been
```
Skepfyr, line 252: My man pages claim `system` is safe, but it does fork so may be `nonportable` for the same reason fork is.

otherlibs/unix/unix.mli
```diff
@@ -1233,13 +1233,13 @@ val time : unit -> float
...
 (** Convert a time in seconds, as returned by {!time}, into a date and
    a time. Assumes UTC (Coordinated Universal Time), also known as GMT.
    To perform the inverse conversion, set the TZ environment variable
    to "UTC", use {!mktime}, and then restore the original value of TZ. *)
 
-val localtime : float @ local -> tm
+val localtime : float @ local -> tm @@ nonportable
 (** Convert a time in seconds, as returned by {!time}, into a date and
    a time. Assumes the local time zone.
    The function performing the inverse conversion is {!mktime}. *)
```
Skepfyr: For `mktime`, "The [tm] argument is interpreted in the local timezone" make me think it's `nonportable`.

otherlibs/unix/unix.mli
```diff
@@ -1389,24 +1389,24 @@ type group_entry =
...
+val getpwuid : int -> passwd_entry @@ nonportable
 (** Find an entry in [passwd] with the given user id.
 
    @raise Not_found if no such entry exists, or always on Windows. *)
 
-val getgrgid : int -> group_entry
+val getgrgid : int -> group_entry @@ nonportable
 (** Find an entry in [group] with the given group id.
 
    @raise Not_found if no such entry exists, or always on Windows. *)
```
Skepfyr: `gethostbyname` and friends are thread unsafe, as is `getaddrinfo` and `getnameinfo`. ;(

otherlibs/unix/unix.mli
```diff
@@ -217,7 +217,7 @@ type wait_flag =
   | WUNTRACED (** Report also the children that receive stop signals. *)
 (** Flags for {!waitpid}. *)
 
-val execv : string -> string array -> 'a
+val execv : string -> string array -> 'a @@ nonportable
```
Skepfyr: The non-`p` versions of exec appear to be thread safe (depending on why `fork` isn't)

otherlibs/unix/unix.mli
polytypic: I believe `Unix.fork` is considered particularly multithread unsafe, because any user and system level shared state might be left in a bad state in the new process.  The situation with `Unix.system` is slightly different as it basically `fork`s and then more or less immediately `exec`s.

otherlibs/unix/unix.mli
polytypic: Good catch!

As discussed, we plan to mark `Unix` functions, such as `putenv`, that modify global shared state as `nonportable` and `[@@alert unsafe_multidomain]` (the intention being that one should not use them), and also annotate functions that read global state (environment, locale) as `portable`.  We considered adding extra locking around functions that access global shared state, but it seems libc is already doing some amount of locking and the set of affected functions (that read global shared state) is essentially open ended, which means adding our own locking would only go so far.

otherlibs/unix/unixLabels.mli
```diff
@@ -1386,24 +1390,24 @@ type group_entry = Unix.group_entry =
   }
```
Skepfyr: `gmtime`, `localtime` as implemented here is MT-unsafe (it could call the `_r` variants but doesn't).

otherlibs/unix/unixLabels.mli
```diff
@@ -1386,24 +1390,24 @@ type group_entry = Unix.group_entry =
...
 (** Structure of entries in the [groups] database. *)
```
Skepfyr: `sleep` and `sleepf` are kinda unsafe, usually they use `nanosleep` or `select` but it can fall back to actual `sleep` which is unsafe in the presence of `SIGALRM`.... I'm not sure about this one, somewhat tempted to just pretend it's fine.

otherlibs/unix/unixLabels.mli
```diff
@@ -1386,24 +1390,24 @@ type group_entry = Unix.group_entry =
...
+val getpwuid : int -> passwd_entry @@ nonportable
 (** Find an entry in [passwd] with the given user id.
 
    @raise Not_found if no such entry exists, or always on Windows. *)
 
-val getgrgid : int -> group_entry
+val getgrgid : int -> group_entry @@ nonportable
 (** Find an entry in [group] with the given group id.
 
    @raise Not_found if no such entry exists, or always on Windows. *)
```
Skepfyr: `gethostbyname`, `gethostbyaddr` is also kinda unsafe, if the `_r` versions exist they're fine but not otherwise.
`getprotobyname`, `getprotobynumber`, `getservbyname`, `getservbyport` are all racy.

Skepfyr (changes requested):

otherlibs/unix/unixLabels.mli
polytypic: I believe it is reasonable to assume `select` exists.

otherlibs/unix/unixLabels.mli
polytypic: I added implementations using the `_r` versions.

otherlibs/unix/unixLabels.mli
polytypic: I marked `getprotobyname`, `getprotobynumber`, `getservbyname`, and `getservbyport` as nonportable.

I believe we can assume the `_r` versions exists.

To make `getaddrinfo` and `getnameinfo` portable, I made it so that, in case the safe versions don't exist, the "emulation" using thread-unsafe functions is allowed only when running on main domain.

otherlibs/unix/getlogin.c
```diff
@@ -22,8 +22,16 @@ extern char * getlogin(void);
 
 CAMLprim value caml_unix_getlogin(value unit)
 {
+#ifdef HAS_GETLOGIN_R
+  char name[_POSIX_LOGIN_NAME_MAX];
```
Skepfyr: Unfortunately, `_POSIX_LOGIN_NAME_MAX` is `9`, `LOGIN_NAME_MAX` is the number you want (`256` on Linux).

polytypic: :sweat_smile: &mdash; Good catch! I believe tried using `LOGIN_NAME_MAX`, but it seemed like isn't always defined in `<limits.h>`. I think I'll switch this to dynamically allocating a buffer similar to various other functions.

I'll also try and write some trivial smoke tests for the functions that this PR changes.  There are no existing tests for these functions. :astonished:

otherlibs/unix/getgr.c
```diff
@@ -41,6 +41,37 @@ static value alloc_group_entry(struct group *entry)
...
+#ifdef HAS_GETGRNAM_R
+CAMLprim value caml_unix_getgrnam(value name)
+{
+  struct group * entry, entry_val;
+  size_t len;
+  char *buf;
+  int error;
+  value result;
+  if (! caml_string_is_c_safe(name)) caml_raise_not_found();
+  len = 1024;
```
glittershark, line 53: the man page for `getgrnam_r` says:

```
     The call

         sysconf(_SC_GETGR_R_SIZE_MAX)

     returns  either  -1,  without   changing  errno,  or  an  initial
     suggested size  for buf.  (If  this size  is too small,  the call
     fails with  ERANGE, in  which case  the caller  can retry  with a
     larger buffer.)
```

Maybe we should do that in addition to retrying on `ERANGE`?

though, retrying on `ERANGE` is plausibly good enough; if you think so feel free to ignore this comment.

polytypic: I feel retry on `ERANGE` should be good enough &mdash; I assume 1k is usually large enough (maybe I'm wrong?) and using `sysconf` adds some complexity as you need to check for errors and you still need the logic to grow the buffer.   Should be easy to address later if the current implementation proves problematic.

PR #4313, opened 2025-07-16
Add llvmize tests + QoL tweaks

usuyus:
This PR adds the following:

- Two types of tests under oxcaml/tests/backend/llvmize
    - Only compile down to LLVM IR and check the output (done in `id_fn`)
    - Compile down to an executable and run it (done in `const_val`)
- New flags
    - `-stop-after llvmize`: only emits the `-ll` file and doesn't compile it into asm
    - `-keep-llvmir` keeps the `.ll` file (analog of `-S`)
    - `-dllvmir` now does nothing
- Some tweaks in `Llvmize`
    - Take `sourcefile` in begin_assembly
    - Pass `ppf_dump` to llvmize to implement `-dllvmir` in the future
    - Move init logic out of `open_out` into `init`
    - Specify llvmize in error messages
    - Touch up debug comments


asmcomp/asmgen.ml
```diff
@@ -202,25 +202,28 @@ let should_use_linscan fd =
...
 let if_emit_do f x = if should_emit () then f x else ()
 
-let emit_begin_assembly unix =
-  if_emit_do
-    (fun () ->
-      if !Oxcaml_flags.llvm_backend
-      then Llvmize.begin_assembly ()
-      else Emit.begin_assembly unix)
-    ()
+(* CR yusumez: [Llvmize] uses [begin_assembly] and [end_assembly] to emit extra
```
gretay-js: Please move this comment to llvmize.ml .

driver/oxcaml_args.ml
```diff
@@ -252,11 +252,14 @@ let mk_gc_timings f =
...
 
 let mk_llvm_backend f =
-  "-llvm-backend", Arg.Unit f, " Enable LLVM backend (...does nothing for now)"
+  "-llvm-backend", Arg.Unit f, " Enable LLVM backend (experimental)"
 
 let mk_dllvmir f =
   "-dllvmir", Arg.Unit f, " (undocumented)"
 
+let mk_keep_llvmir f =
+  "-keep-llvmir", Arg.Unit f, " Keep the LLVM IR file"
```
gretay-js:
```suggestion
  "-keep-llvmir", Arg.Unit f, " Keep the LLVM IR file produce by -llvm-backend"
```

backend/cfg/llvmize.mli
```diff
@@ -17,10 +17,12 @@ val cfg : Cfg_with_layout.t -> unit
 
 val data : Cmm.data_item list -> unit
 
-val begin_assembly : unit -> unit
+val begin_assembly : sourcefile:string option -> unit
```
gretay-js, line 20: Add sourcefile as a field of `Llvmize.t` so we don't need to pass it to `end_assembly` and it's available to use throughout the compilation unit for example for user error messages.

oxcaml/tests/backend/llvmize/gen/gen_dune.ml
```diff
@@ -0,0 +1,97 @@
...
+(rule
+ ${enabled_if}
+ (alias runtest)
+ (deps ${output} ${output}.corrected)
+ (action
+  (diff ${output} ${output}.corrected)))
+|}
+  in
+  print_test_llvmir "id_fn";
+  print_test_run ~bootstrap:"const_val_bootstrap" "const_val"
```
gretay-js: Seems redundant to provide the same string twice. Each test will need its own `main`, the name can be generated, for example `const_val_test.ml` and `const_val_main.ml`  can be generated from `const_val`.

oxcaml/tests/backend/llvmize/id_fn.output
```diff
@@ -0,0 +1,68 @@
+source_filename = "id_fn.ml"
+target triple = "x86_64-redhat-linux-gnu"
+
+@camlId_fn__data_begin = global i64 0
+define void @camlId_fn__code_begin() { ret void }
+
+; camlId_fn__f_0_1_code [ id_fn.ml:1,21--26 ]
+define i64 @camlId_fn__f_0_1_code(i64 %0)  {
```
gretay-js: The substring `_0_1_` is not stable, in the sense that the numbers can change with configure options and unrelated compiler changes. This substring  should be filtered out here and in the references to it.  Same problem for per-instruction comments that mention registers, but it'd be easier to just not print the comments. Piggyback on `-dno-asm-comments` for now to determine if comments should be emitted or not.

oxcaml/tests/backend/llvmize/filter.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+# remove trailing whitespace at the ends of lines
```
gretay-js, line 3: is it necessary?

oxcaml/tests/backend/llvmize/dune.inc
```diff
@@ -0,0 +1,45 @@
...
+(rule
+ (enabled_if (and (= %{context_name} "main") (= %{architecture} "amd64")))
+ (targets id_fn.output.corrected)
+ (deps (:ml id_fn.ml) filter.sh)
+ (action
+  (with-outputs-to
+   id_fn.output.corrected
+   (pipe-outputs
+    (run
+     %{bin:ocamlopt.opt} %{ml} -c -llvm-backend -stop-after llvmize -keep-llvmir)
```
gretay-js: Add `-g -O3` compilation flags too.

oxcaml/tests/backend/llvmize/dune.inc
```diff
@@ -0,0 +1,45 @@
...
+(rule
+ (enabled_if (and (= %{context_name} "main") (= %{architecture} "amd64")))
+ (alias runtest)
+ (deps id_fn.output id_fn.output.corrected)
+ (action
+  (diff id_fn.output id_fn.output.corrected)))
+
+(rule
+ (enabled_if (and (= %{context_name} "main") (= %{architecture} "amd64")))
+ (targets const_val.output.exe)
```
gretay-js, line 44: `const_val` should also have llvmir output test.

oxcaml/tests/backend/llvmize/dune.inc
```diff
@@ -0,0 +1,45 @@
...
+ (deps (:ml const_val_bootstrap.ml const_val.ml))
+ (action
+  (pipe-outputs
```
gretay-js: what does `pipe-outputs` do here? should it be `progn` instead to run a sequence of commands?

gretay-js: this may be the cause of ci test failures.

oxcaml/tests/backend/llvmize/id_fn.output
xclerc: (`target triple` above should
probably be redacted too.)

oxcaml/tests/backend/llvmize/const_val_ir.output
```diff
@@ -0,0 +1,34 @@
+source_filename = "const_val_ir.ml"
+
+
+@camlConst_val_ir.data_begin = global i64 0
+define void @camlConst_val_ir.code_begin() { ret void }
+
+define i64 @camlConst_val_ir.entry()  {
```
gretay-js, line 6: It should be _entry, not .entry I think.

gretay-js (approved):

oxcaml/tests/backend/llvmize/const_val_ir.output
gretay-js: oh, this is the difference in rt4 vs rt5 identifiers.

oxcaml/tests/backend/llvmize/id_fn.output
gretay-js: actually, the triple is hardcoded at the moment in Llvmize, which is obvously wrong the target is not amd64. amd also causing a warning from llvm. We don't have to emit target triple at all for now.

oxcaml/tests/backend/llvmize/filter.sh
gretay-js: some IDE interaction

PR #4314, opened 2025-07-16
add `Mode.Crossing.Atom`

riaqn:
Based on #4436 

This PR adds the per-axis version of `Mode.Crossing`.

It's correctness is demonstrated in the last commit, where we print mode crossing using the utilities just introduced.

### DIFF
--- a/typing/mode.ml
+++ b/typing/mode.ml
@@ -3485,69 +3485,72 @@ module Crossing = struct
   *)
 
   module Monadic = struct
-    module Modality = Modality.Monadic.Const
     module Mode = Value.Monadic
 
-    type t = Modality.t
+    type t = Modality of Modality.Monadic.Const.t [@@unboxed]
 
-    let of_bounds c : t = Join_const c
+    let of_bounds c = Modality (Join_const c)
 
-    let modality m t = Modality.concat ~then_:t m
+    let modality m (Modality t) =
+      Modality (Modality.Monadic.Const.concat ~then_:t m)
 
-    let apply_left : t -> _ -> _ = function
-      | Join_const c -> fun m -> Mode.subtract c (Mode.join_const c m)
+    let apply_left (Modality (Join_const c)) m =
+      Mode.subtract c (Mode.join_const c m)
 
-    let apply_right : t -> _ -> _ = function
-      | Join_const c ->
-        fun m ->
-          (* The right adjoint of join is a restriction of identity *)
-          Mode.join_const c m
+    let apply_right (Modality (Join_const c)) m =
+      (* The right adjoint of join is a restriction of identity *)
+      Mode.join_const c m
 
-    let le (t0 : t) (t1 : t) =
-      match t0, t1 with Join_const c0, Join_const c1 -> Mode.Const.le c1 c0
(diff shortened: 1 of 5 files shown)


typing/mode.ml
```diff
@@ -2170,13 +2193,40 @@ module Monadic = struct
...
+        let obj = proj_obj ax in
+        C.le obj b a
+
+      let join ax a b =
+        let obj = proj_obj ax in
+        C.meet obj a b
+
+      let meet ax a b =
+        let obj = proj_obj ax in
+        C.join obj a b
```
glittershark: maybe comment on why the joins and meets here are backwards? I got spooked initially

typing/mode.ml
```diff
@@ -2170,13 +2193,40 @@ module Monadic = struct
...
+
+      let max ax =
+        let obj = proj_obj ax in
+        C.min obj
+
+      let min ax =
+        let obj = proj_obj ax in
+        C.max obj
```
glittershark: and max and min I suppose

typing/mode.ml
```diff
@@ -3357,6 +3407,56 @@ module Crossing = struct
...
+
+    let print ax ppf t = Modality.print ppf (Atom (ax, t))
+
+    let min : _ Value.Axis.t -> _ t = function
+      | Monadic ax -> Join_with (Value.Monadic.Const.Per_axis.max ax)
+      | Comonadic ax -> Meet_with (Value.Comonadic.Const.Per_axis.min ax)
+
+    let max : _ Value.Axis.t -> _ t = function
+      | Monadic ax -> Join_with (Value.Monadic.Const.Per_axis.min ax)
+      | Comonadic ax -> Meet_with (Value.Comonadic.Const.Per_axis.max ax)
```
glittershark: same here probably.

typing/mode.ml
```diff
@@ -3357,6 +3407,56 @@ module Crossing = struct
...
+      [@@ocaml.warning "-4"]
+
+    let join (ax : _ Value.Axis.t) (a : _ t) (b : _ t) : _ t =
+      match ax, a, b with
+      | Monadic ax, Join_with a, Join_with b ->
+        Join_with (Value.Monadic.Const.Per_axis.meet ax a b)
+      | Comonadic ax, Meet_with a, Meet_with b ->
+        Meet_with (Value.Comonadic.Const.Per_axis.join ax a b)
+      | _ -> assert false
+      [@@ocaml.warning "-4"]
```
glittershark: do we actually need to squelch this warning? These matches look total to me.

riaqn: I've now make those matches total.

typing/mode.ml
riaqn: I think it's not right to think "comonadic fragment has it in the expected order, monadic fragment has it flipped and thus requires documentation". 

Rather, there is no "expected order": the ordering of mode crossing is a fresh thing that requires fresh definitions (which can be found above in this file). Currently all our mode crossing can be expressed using modalities, but they are not the same thing, and we could introduce more mode crossing (that has absolutely nothing to do with modalities) in the future.

PR #4315, opened 2025-07-16
Regalloc.exe: miscellaneous enhancements

xclerc:
This pull request extends the `regalloc.exe` tools by:

- adding columns to the CSV format (number of blocks, destruction points, etc);
- iterating over the paths provided on the command line rather than expecting
  the caller to provide a list of files to process.


backend/regalloc/regalloc_split_utils.mli
```diff
@@ -2,6 +2,8 @@
 
 val split_live_ranges : bool Lazy.t
 
+val split_more_destruction_points : bool Lazy.t
```
TheNumbat, line 5: Should this be an argument of regalloc.exe? It's only used there

TheNumbat (approved):

backend/regalloc/regalloc_split_utils.mli
xclerc: Yes, sorry, I forgot to include that part.
I have added a `-param` command-line
option which is the counterpart of
of `regalloc-param` in the compiler.

PR #4318, opened 2025-07-17
Extend the Type Shapes

spiessimon:
This PR is part of the chain of PRs for propagating debugging information through the compiler (e.g., see also https://github.com/oxcaml/oxcaml/pull/3967). This PR extends the recently introduced notion of type shapes (see #4253) to support several additional type formers (e.g., unboxed types, exceptions, float32, unboxed variants and records, etc.).

### DIFF
--- a/typing/type_shape.mli
+++ b/typing/type_shape.mli
@@ -79,11 +79,26 @@ module Type_shape : sig
 
   type 'a ts =
     | Ts_constr of (Uid.t * Path.t * 'a) * without_layout ts list
+        (** [Ts_constr ((uid, p, ly), args)] is the shape of the type constructor
+        [p] applied to the arguments [args], resulting in data of layout [ly].
+        Unlike for the resulting data, we do not definitely know the layout
+        of the arguments yet. This is only fully determined once we look at the
+        declaration behind [p]. Thus, even when the type constructor application
+        [Ts_constr((uid, p, ly), args)] itself carries a layout, the layout of
+        the arguments is still to be determined.  *)
     | Ts_tuple of 'a ts list
     | Ts_unboxed_tuple of 'a ts list
     | Ts_var of string option * 'a
     | Ts_predef of Predef.t * without_layout ts list
+        (** [Ts_predef(p, args)] is the shape for predefined type shapes.
+          Analogously to [Ts_constr], its arguments do not carry a layout yet.
+        *)
     | Ts_arrow of without_layout ts * without_layout ts
+        (** [Ts_arrow(arg, ret)] is the shape for the function type
+            [arg -> ret]. When emitting DWARF information for this type, there
+            is no need to inspect the type arguments, and we never find out
+            their layout. As such, these remain [without_layout] even after
+            substituting in layouts.   *)
     | Ts_variant of 'a ts poly_variant_constructors
     | Ts_other of 'a
 


typing/type_shape.mli
```diff
@@ -1,84 +1,209 @@
...
-  include Identifiable.S with type t := t
-end
+    val unboxed_type_to_layout : unboxed -> Jkind_types.Sort.base
 
-module Type_decl_shape : sig
-  type 'a complex_constructor =
-    { name : string;
-      args : 'a complex_constructor_argument list
-    }
+    val predef_to_layout : t -> Layout.t
```
lukemaurer: Since this is already the `Predef` module, this should just be called `to_layout`.

typing/type_shape.mli
```diff
@@ -1,84 +1,209 @@
...
-    { field_name : string option;
-      field_value : 'a
+  (* Type shapes are abstract representations of type expressions. We define
+     them with a placeholder 'a for the layout inside. This allows one to
+     first create shapes without a type by picking [without_layout] for 'a
+     and then later substituting in a layout of type [Layout.t]. *)
+  type without_layout = Layout_to_be_determined
+
+  type 'a ts =
+    | Ts_constr of (Uid.t * Path.t * 'a) * without_layout ts list
```
lukemaurer, line 81: Not clear to me why so many (but not all) of the components are always `without_layout`? Worth some comments in any event

typing/type_shape.mli
```diff
@@ -91,6 +216,6 @@ val add_to_type_shapes :
   (Path.t -> Uid.t option) ->
   unit
 
-val find_in_type_decls : Uid.t -> Type_decl_shape.t option
+val find_in_type_decls : Uid.t -> Type_decl_shape.tds option
 
-val type_name : Type_shape.t -> string
+val type_name : 'a. 'a Type_shape.ts -> string
```
lukemaurer: Probably best to just say `_ Type_shape.ts` here (the `'a.` makes me think there's something fancy going on).

typing/type_shape.ml
```diff
@@ -1,60 +1,265 @@
...
+      match b with
+      | Unboxed_float -> Float64
+      | Unboxed_float32 -> Float32
+      | Unboxed_nativeint -> Word
+      | Unboxed_int64 -> Bits64
+      | Unboxed_int32 -> Bits32
+      | Unboxed_simd s -> simd_vec_split_to_layout s
+
+    let predef_to_layout : t -> Layout.t = function
+      | Array -> Layout.Base Value
```
lukemaurer: Should be possible to leave off the `Layout.` prefix in these if I'm not mistaken.

typing/type_shape.ml
```diff
@@ -77,35 +282,132 @@ module Type_shape = struct
...
-  let rec print ppf = function
-    (* CR sspies: We should figure out pretty printing for shapes. For now, this
-       verbose non-boxed version should be fine. *)
+  let rec shape_with_layout ~(layout : Layout.t) (sh : without_layout ts) :
+      Layout.t ts =
+    match sh, layout with
+    | Ts_constr ((uid, path, Layout_to_be_determined), shapes), _ ->
+      Ts_constr ((uid, path, layout), shapes)
+    | Ts_tuple shapes, Base Value ->
+      let layouted_shapes =
```
lukemaurer: The word _layouted_ feels pretty awkward. Can we just say `shapes_with_layout`?

typing/type_shape.ml
```diff
@@ -77,35 +282,132 @@ module Type_shape = struct
...
 
-  let rec print ppf = function
-    (* CR sspies: We should figure out pretty printing for shapes. For now, this
-       verbose non-boxed version should be fine. *)
+  let rec shape_with_layout ~(layout : Layout.t) (sh : without_layout ts) :
+      Layout.t ts =
+    match sh, layout with
+    | Ts_constr ((uid, path, Layout_to_be_determined), shapes), _ ->
+      Ts_constr ((uid, path, layout), shapes)
+    | Ts_tuple shapes, Base Value ->
```
lukemaurer, line 345: This makes me think `Ts_tuple` should take `without_layout` arguments? We never use the layouts

typing/type_shape.ml
```diff
@@ -77,35 +282,132 @@ module Type_shape = struct
...
+    | Ts_predef (predef, shapes), _
+      when Layout.equal (Predef.predef_to_layout predef) layout ->
+      Ts_predef (predef, shapes)
+    | Ts_predef (predef, _), _ ->
+      Misc.fatal_errorf
+        "predef %s has layout %a, but is expected to have layout %a"
+        (Predef.to_string predef) Layout.format
+        (Predef.predef_to_layout predef)
+        Layout.format layout
+    | Ts_variant fields, Base Value ->
```
lukemaurer, line 388: Similarly here, seems like `Ts_variant` should take `without_layout` arguments

typing/type_shape.ml
```diff
@@ -168,49 +520,103 @@ module Type_decl_shape = struct
...
+              field_value =
+                Type_shape.of_type_expr lbl.ld_type uid_of_path, lbl.ld_sort
+            })
+          list
+    in
+    let constructor_repr =
+      match constructor_repr with
+      | Constructor_mixed shapes ->
+        let shapes_and_fields = List.combine (Array.to_list shapes) args in
+        List.iter
```
lukemaurer: Can use `List.iter2` here

typing/type_shape.ml
```diff
@@ -168,49 +520,103 @@ module Type_decl_shape = struct
...
-    | Tds_variant of
-        { simple_constructors : string list;
-          (* CR sspies: Deduplicate these cases once type shapes have reached
-             a more stable form. *)
-          complex_constructors : Type_shape.t complex_constructor list
-        }
-    | Tds_record of (string * Type_shape.t) list
-    | Tds_alias of Type_shape.t
-    | Tds_other
+  and constructor_representation = Constructor_mixed of mixed_product_shape
```
lukemaurer: If you're unifying mixed and non-mixed constructor representations, I'd rather avoid `Constructor_mixed` as the name here. If you're not planning on adding more constructors, we can of course just say `type constructor_representation = mixed_product_shape`.

typing/type_shape.mli
spiessimon: Done.

typing/type_shape.mli
spiessimon: Done.

typing/type_shape.ml
spiessimon: Done.

typing/type_shape.ml
spiessimon: Done.

typing/type_shape.ml
spiessimon: Done.

typing/type_shape.ml
spiessimon: Done.

typing/type_shape.mli
spiessimon: I've gone back and forth on the treatment of the layouts in type shapes. This version with layouts inside the shape type is designed for a DWARF emission function that is simply recursive on a `Layout.t Type_shape.ts`. The alternative is an emission function that threads through the layout and the type as separate arguments and basically performs the kind of recursive matching on both that we now do in `shape_with_layout`. 

For this approach, when we insert the layout into the type shape, there are several places for which we do not yet have a layout. For example,  in the `Ts_constr` case, the layout of the type arguments cannot be derived from the layout of the constructor application, because we do not yet know how the arguments will be used. That's where all the uses of `without_layout` come from. 

That said, I do think the version with `'a` is confusing. In the latest version (several changes down the road), I've gone back to keeping them entirely separate, and emitting the final DWARF by simultaneous recursion over type shape and type layout. How do you feel about a CR-comment indicating that this will be changed in subsequent PRs instead of a careful explanation?

typing/type_shape.ml
spiessimon: Hmm, there are no uses of the version with layouts in this PR. They are in the final emission of the type shapes to DWARF. There, it is nice to recursive over a `Layout.t Type_shape.ts` without running into a `Type_shape.without_layout Type_shape.ts` along the way. My proposal would be: keep it for now and then move the layout out in subsequent PRs (see the message above).

typing/type_shape.ml
spiessimon: Same as above.

typing/type_shape.mli
lukemaurer: That makes some sense. I would still like to have code that makes the most sense it can in the meantime, though, and as is it's not obvious why certain cases are hard-wired to `without_layout` and some aren't. (And actually `Layout_to_be_determined` is the wrong value to store inside a `Layout.t ts` since that's no longer waiting to determine the layout—there simply isn't a layout being stored there. So it's not clear that `without_layout` is the right type to hard-wire things to.)

typing/type_shape.ml
lukemaurer: Ah, is it that the hard-wired `without_layout`s are layouts that don't appear in the final DWARF info? That would be good to capture in a comment. (Even better, you could have separate `layout_pending` and `no_layout` types, and the comment on `no_layout` can explain that it's a layout that doesn't appear in the final DWARF info.)

lukemaurer: Discussed offline; my suggestion doesn't work.

typing/type_shape.mli
lukemaurer: As discussed offline, this is fine to leave with a comment about why some recursive occurrences say `'a` and some say `without_layout`.

spiessimon: I've added comments to the cases of `ts`. Explaining why we don't have a layout there. Please have a look.

typing/type_shape.ml
spiessimon: As discussed offline, this is not quite right. The layout of, for example, type constructor arguments is determined when we look up the declaration and insert them into the declaration (where we find layouts in, for example, the declaration of a mixed record).

PR #4319, opened 2025-07-17
Printing Support for the Debugging UID Tables

spiessimon:
This PR adds support for printing the debugging UIDs collected in the type shape tables (introduced in #4253). To do so, it adds two new flags, namely `-ddebug-uids` and `-ddebug-uid-tables`. The former adds uid information to binders and the latter prints the tables with debug uids in `type_shape.ml`. At the moment, it only supports printing the tables for the lambda and rawlambda representations. In #4255, this is expanded with support for the other intermediate representations.

### DIFF
--- a/driver/compile.ml
+++ b/driver/compile.ml
@@ -46,6 +46,8 @@ let raw_lambda_to_bytecode i raw_lambda ~as_arg_for =
            arg_block_idx } ->
        Builtin_attributes.warn_unused ();
        lambda
+       |> print_if i.ppf_dump Clflags.dump_debug_uid_tables
+          (fun ppf _ -> Type_shape.print_debug_uid_tables ppf)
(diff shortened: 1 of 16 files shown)


mshinwell:
Looks ok but I'm unsure we should always print the uid tables when also printing uids.  In addition we should have some way of printing the uids without dumping the lambda code, as they will also show up in the Flambda IR, right?  Maybe we could solve both of these problems by printing the uid tables iff a particular (new) command-line flag is set, irrespective of what IR is being dumped?

Will need rebasing.

spiessimon:
This PR currently adds `-ddebug-uids`. We could add `-ddebug-uid-tables`, with the semantics the former adds them only to the IR (but no table printing) and the latter only dumps the tables. Perhaps we then want one central place to dump the tables. Not sure where in the compiler the right place for that would be, though. (For tests, it would be good if we can also dump the tables during non-native compilation.)

PR #4321, opened 2025-07-17
Simplify Value.Axis.t GADT

riaqn:
Currently, `Value.Axis.t` is a GADT with three parameters `a, d0, d1`. The `d0` and `d1` are to enable a polymorphic `proj` function that works for both monadic and comonadic modes. However, callers always know exactly which axis they are projecting, and thus never need this polymorphism. The `d0` and `d1` also complicates code elsewhere.

This PR removes these extra parameters.

### DIFF
--- a/typing/mode.ml
+++ b/typing/mode.ml
@@ -1992,8 +1992,6 @@ end
 module Lattice_Product (L : Lattice) = struct
   open L
 
-  type 'a axis = (t, 'a) Axis.t
-
   let min_with ax c = Axis.set ax c min
 
   let max_with ax c = Axis.set ax c max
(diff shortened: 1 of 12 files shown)


typing/mode.ml
```diff
@@ -2301,38 +2317,34 @@ module Value_with (Areality : Areality) = struct
...
+      | Comonadic : 'a Comonadic.Axis.t -> 'a t
 
-    let compare : type a d0 d1 b e0 e1. (a, d0, d1) t -> (b, e0, e1) t -> int =
+    let compare : type a b. a t -> b t -> int =
      fun t0 t1 ->
       match t0, t1 with
       | Monadic t0, Monadic t1 -> Axis.compare t0 t1
-      | Monadic _, Comonadic _ -> 1
-      | Comonadic _, Monadic _ -> -1
+      | Monadic t0, Comonadic t1 -> Axis.compare t0 t1
```
dkalinichenko-js, line 2278: What is happening in those two lines? Why are we comparing monadic and comonadic axes?

dkalinichenko-js (commented):
Looks good to me overall.

typing/mode.ml
riaqn: I guess you mean "monadic and comonadic axes are not comparable, because they don't have implication order among them", which I agree. But currently we do implement a single type `('r, 'a) Axis.t` that contains all the axes.

dkalinichenko-js: Oh, I see. Yeah. Let's simplify this to `| (Monadic t0 | Comonadic t0), (Monadic t1 | Comonadic t1) -> Axis.compare t0 t1` so it looks more intentional though.

dkalinichenko-js (approved):

typing/mode.ml
riaqn: That wouldn't work due to GADT. :cry:

PR #4324, opened 2025-07-17
Add missing SSE intrinsics

TheNumbat:
Implements intrinsics for `ptest`, `pmuldq` / `pmuludq` / `pmulhrsw`, and `cvttpd2dq` / `cvttps2dq`.
The latter are also implemented for arm.

Adds tests for the new intrinsics.


oxcaml/tests/simd/gen/gen_dune.ml
```diff
@@ -142,6 +142,7 @@ let () =
       "ops_int16x16";
       "ops_int8x32";
       "sse_other_ops";
+      "sse41_test_ops";
```
gretay-js, line 145: this file is not used, it needs to be added to the list below as well.

oxcaml/tests/simd/amd64/sse41_test_ops.ml
```diff
@@ -0,0 +1,80 @@
...
+  let result = Vec128.testnzc v1 v2 in
+  if result <> 0
+  then Printf.printf "testnzc failed: expected 0, got %d\n" result
+
+let () =
+  (* Test with different vector types *)
+  let f1 = Float32x4.of_float32s 1.0s 2.0s 3.0s 4.0s in
+  let f2 = Float32x4.of_float32s 0.0s 0.0s 0.0s 0.0s in
+  let i1 = (Obj.magic f1 : int64x2) in
+  let i2 = (Obj.magic f2 : int64x2) in
```
gretay-js: Can we use `bits_of_float` to avoid magic?

oxcaml/tests/simd/amd64/sse_other_ops.ml
```diff
@@ -157,6 +157,41 @@ module Int32x4 = struct
...
   let () =
+    Int32s.check_ints (fun l r ->
+        (failmsg := fun () -> Printf.printf "%08lx|%08lx mul_even\n%!" l r);
+        let v0 = Int32s.of_int32s l 1l r 3l in
+        let v1 = Int32s.of_int32s r 2l l 1l in
+        let result = mul_even v0 v1 in
+        let expect =
+          int64x2_of_int64s
+            Stdlib.Int64.(mul (of_int32 l) (of_int32 r))
+            Stdlib.Int64.(mul (of_int32 l) (of_int32 r))
```
gretay-js:
```suggestion
            Stdlib.Int64.(mul (of_int32 l) (of_int32 r))
            Stdlib.Int64.(mul (of_int32 r) (of_int32 l))
```
more consistent but should change the result

oxcaml/tests/simd/stubs.c
```diff
@@ -94,6 +95,12 @@ static inline simd_float32x4_t simd_float32x4_round_near(simd_float32x4_t v)
   return vrndnq_f32(v);
 }
 
+static inline simd_int32x4_t simd_float64x2_to_int32x2_trunc(simd_float64x2_t v)
+{
+  int32x2_t i32 = vqmovn_s64(vcvtq_s64_f64(v));
+  return vcombine_s32(i32, i32);
```
gretay-js: should the high part be 0?
```suggestion
  return vcombine_s32(i32,0);
```

gretay-js (approved):

oxcaml/tests/simd/stubs.c
TheNumbat: sure, shouldn't make a difference since the x86 version has unspecified high bits though

oxcaml/tests/simd/amd64/sse41_test_ops.ml
TheNumbat: deleted this bit, it's not testing anything interesting

oxcaml/tests/simd/gen/gen_dune.ml
TheNumbat: oops, thanks, that exposed that `ptest` has the wrong result encoding - fixed by changing the defintion in `amd64.csv`

PR #4325, opened 2025-07-17
Correct Max Mod Bounds

alpha-convert:
Fixes a bug in a function `is_max` for checking if a set of mod-bounds is the max possible. In particular, `Portable` is not the max of portability, `External` is not the max of externality. This function is used as part of the early termination condition for the normalization in `constrain_type_jkind`.


glittershark (changes requested):
Can we write at least one test case to cover this (`testsuite/typing-jkind-bounds` might be a reasonable place)? I think you had an example, maybe there are more?

glittershark (approved):
I'll defer to @goldfirere here.

PR #4331, opened 2025-07-18
Output marshaled .cmj files after compiling to Js_of_ocaml IR

smjleo:
Output to a file instead of dumping to stdout.

### DIFF
--- a/middle_end/flambda2/to_jsir/to_jsir_env.ml
+++ b/middle_end/flambda2/to_jsir/to_jsir_env.ml
@@ -32,9 +32,9 @@ type continuation =
   | Block of Jsir.Addr.t
 
 let get_continuation_exn t cont =
-  if cont = t.return_continuation
+  if Continuation.equal cont t.return_continuation
   then Return
-  else if cont = t.exn_continuation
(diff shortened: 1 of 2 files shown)


driver/jscompile.ml
```diff
@@ -68,6 +68,24 @@ let raw_lambda_to_jsir i raw_lambda ~as_arg_for =
...
+    ~exceptionally:(fun () ->
+      Misc.remove_file (Unit_info.Artifact.filename cmj))
+    (fun () ->
+      (* CR selee: it's fairly likely that we will have to include more metadata
+         - for example [Cmo_format] includes [Compilation_unit] to get the linkage
+         name etc. *)
+      output_string oc Config.cmj_magic_number;
+      (* We include the highest used variable in the translation, so that Js_of_ocaml
+         can read this number and update its own state accordingly. *)
+      output_binary_int oc
```
jvanburen: output a single marshaled record instead of multiple independent fields (like we discussed)

jvanburen:
I think this project is a good opportunity to clean up things like top-level refs in JSIR, but I won't block progress on it

PR #4335, opened 2025-07-19
AVX/AVX2 intrinsics

TheNumbat:
Adds AVX & AVX2 intrinsics according to the Intel intrinsics guide. Sequences are excluded other than `vptest`. 
Load/store operations are also excluded for now, as we need a better abstraction for providing memory locations.

Almost all of the diff is tests. They're pretty messy and many were generated by claude code, but I've verified they check all intrinsics basically work. I'll write better tests for the `ocaml_simd_avx` library.


oxcaml/tests/simd/amd64/utils256.ml
```diff
@@ -36,17 +38,115 @@ external int64x4_third_int64 : int64x4 -> int64 = "" "vec256_third_int64"
...
+
+external float64x4_of_float64x2s : float64x2 -> float64x2 -> float64x4
+  = "" "vec256_of_vec128s"
+  [@@noalloc] [@@unboxed]
+
+external int32x8_of_int32x4s : int32x4 -> int32x4 -> int32x8
+  = "" "vec256_of_vec128s"
+  [@@noalloc] [@@unboxed]
+
+external extract_128_f32 :
```
spiessimon, line 113: What happens for these if one passes an index that is neither `0` nor `1`? Should these perhaps be two different functions like for the projection operations above?

oxcaml/tests/simd/amd64/stub_builtins.c
```diff
@@ -208,3 +208,211 @@ BUILTIN(caml_sse42_vec128_cmpistrz);
...
+BUILTIN(caml_avx_vec256_permute2_128);
+BUILTIN(caml_avx_vec128_permutev_64);
+BUILTIN(caml_avx_vec128x2_permutev_64);
+BUILTIN(caml_avx_vec128_permutev_32);
+BUILTIN(caml_avx_vec128x2_permutev_32);
+BUILTIN(caml_avx_float32x8_rcp);
+BUILTIN(caml_avx_float64x4_round);
+BUILTIN(caml_avx_float32x8_round);
+BUILTIN(caml_avx_float64x4_round_current);
+BUILTIN(caml_avx_float32x8_round_current);
```
spiessimon: I wasn't able to find an implementation for these primitives:
```
caml_avx_float32x8_round_current
caml_avx_float32x8_round_near
caml_avx_float32x8_round_neg_inf
caml_avx_float32x8_round_pos_inf
caml_avx_float32x8_round_towards_zero
caml_avx_float64x4_round_current
caml_avx_float64x4_round_near
caml_avx_float64x4_round_neg_inf
caml_avx_float64x4_round_pos_inf
caml_avx_float64x4_round_towards_zero
```
I looked, in particular, in `simd_selection.ml`. At the same time, it seems that there are the builtins 
```
caml_simd_float32_round_current
caml_simd_float32_round_near
caml_simd_float32_round_neg_inf
caml_simd_float32_round_pos_inf
caml_simd_float32_round_towards_zero
caml_simd_float64_round_current
caml_simd_float64_round_near
caml_simd_float64_round_neg_inf
caml_simd_float64_round_pos_inf
caml_simd_float64_round_towards_zero
```
that are currently not defined in this file. Do the latter have the intended semantics?

oxcaml/tests/simd/amd64/avx_test_ops.ml
```diff
@@ -0,0 +1,114 @@
+[@@@ocaml.warning "-unused-module"]
+
+open Utils256
+
+module Vec256 = struct
+  external testz :
```
spiessimon: Is there any reason that these are not defined with the other builtins? If not, I think we should probably move them to the other file.

oxcaml/tests/simd/amd64/stub_builtins.c
```diff
@@ -208,3 +208,211 @@ BUILTIN(caml_sse42_vec128_cmpistrz);
...
+BUILTIN(caml_avx2_int16x16_max_unsigned);
+BUILTIN(caml_avx2_int32x8_max_unsigned);
+BUILTIN(caml_avx2_int8x32_min);
+BUILTIN(caml_avx2_int16x16_min);
+BUILTIN(caml_avx2_int32x8_min);
+BUILTIN(caml_avx2_int8x32_min_unsigned);
+BUILTIN(caml_avx2_int16x16_min_unsigned);
+BUILTIN(caml_avx2_int32x8_min_unsigned);
+BUILTIN(caml_avx2_vec256_movemask_8);
+BUILTIN(caml_avx2_int8x16x2_multi_sad_unsigned);
```
spiessimon, line 350: This one is not declared in `oxcaml/tests/simd/amd64/builtins.ml`, I think.

oxcaml/tests/simd/amd64/stub_builtins.c
```diff
@@ -208,3 +208,211 @@ BUILTIN(caml_sse42_vec128_cmpistrz);
...
+BUILTIN(caml_avx2_int32x8_mul_even);
+BUILTIN(caml_avx2_int32x8_mul_even_unsigned);
+BUILTIN(caml_avx2_int16x16_mul_high);
+BUILTIN(caml_avx2_int16x16_mul_high_unsigned);
+BUILTIN(caml_avx2_int16x16_mul_round);
```
spiessimon, line 355: This one is not declared in `oxcaml/tests/simd/amd64/builtins.ml`, I think.

oxcaml/tests/simd/amd64/stub_builtins.c
spiessimon: They are also not declared in `oxcaml/tests/simd/amd64/builtins.ml` as far as I can tell.

oxcaml/tests/simd/amd64/builtins.ml
```diff
@@ -928,3 +928,855 @@ module Sse_other_builtins = struct
...
+
+  external blendv_8 : int8x32 -> int8x32 -> int8x32 -> int8x32
+    = "caml_vec256_unreachable" "caml_avx2_vec256_blendv_8"
+    [@@noalloc] [@@unboxed] [@@builtin]
+
+  external broadcast_8_128 : int8x16 -> int8x16
+    = "caml_vec128_unreachable" "caml_avx2_vec128_broadcast_8"
+    [@@noalloc] [@@unboxed] [@@builtin]
+
+  external broadcast_8 : int8x16 -> int8x32
```
spiessimon, line 1227: Is it clear that `256` will be the maximum? If not, it could make sense to annotate the `256` function with the suffix `_256`.

oxcaml/tests/simd/amd64/builtins.ml
```diff
@@ -928,3 +928,855 @@ module Sse_other_builtins = struct
...
+
+  external cmpeq : t -> t -> t
+    = "caml_vec256_unreachable" "caml_avx2_int8x32_cmpeq"
+    [@@noalloc] [@@unboxed] [@@builtin]
+
+  external cmpgt : t -> t -> t
+    = "caml_vec256_unreachable" "caml_avx2_int8x32_cmpgt"
+    [@@noalloc] [@@unboxed] [@@builtin]
+
+  external cvtsx_int16x16 : int8x16 -> int16x16
```
spiessimon: This one (and the ones below) don't seem to be specific to `int8x32`. Can we move them to a different module?

oxcaml/tests/simd/amd64/builtins.ml
```diff
@@ -928,3 +928,855 @@ module Sse_other_builtins = struct
...
+
+  external srav : t -> t -> t
+    = "caml_vec256_unreachable" "caml_avx2_int32x8_srav"
+    [@@noalloc] [@@unboxed] [@@builtin]
+
+  external srlv : t -> t -> t
+    = "caml_vec256_unreachable" "caml_avx2_int32x8_srlv"
+    [@@noalloc] [@@unboxed] [@@builtin]
+
+  external sllv_128 : int32x4 -> int32x4 -> int32x4
```
spiessimon: Can we move this function and the `_128` ones below to their own module?

oxcaml/tests/simd/amd64/builtins.ml
```diff
@@ -928,3 +928,855 @@ module Sse_other_builtins = struct
...
+  external sllv : t -> t -> t
+    = "caml_vec256_unreachable" "caml_avx2_int64x4_sllv"
+    [@@noalloc] [@@unboxed] [@@builtin]
+
+  external srlv : t -> t -> t
+    = "caml_vec256_unreachable" "caml_avx2_int64x4_srlv"
+    [@@noalloc] [@@unboxed] [@@builtin]
+
+  external sllv_128 : int64x2 -> int64x2 -> int64x2
+    = "caml_vec128_unreachable" "caml_avx2_int64x2_sllv"
```
spiessimon: Same here.

oxcaml/tests/simd/utils.ml
```diff
@@ -652,22 +656,19 @@ module Int16 = struct
     = "caml_vec128_unreachable" "int16_su8"
     [@@noalloc]
 
-  let of_ints a b c d e f g h =
+  let pack4 a b c d =
```
spiessimon: How about `to_int64`, which would match the other functions?

oxcaml/tests/simd/utils.ml
```diff
@@ -775,7 +776,7 @@ module Int8 = struct
     = "caml_vec128_unreachable" "int8_zxi64"
     [@@noalloc]
 
-  let of_ints a b c d e f g h =
+  let pack8 a b c d e f g h =
```
spiessimon: Same here.

oxcaml/tests/simd/gen/gen_dune.ml
```diff
@@ -187,7 +194,6 @@ let () =
       "consts256", enabled_if_main_amd64_not_macos;
       "consts256_u", enabled_if_main_amd64_not_macos;
       "probes256", enabled_if_main_amd64_not_macos;
-      "utils256", enabled_if_main_amd64_not_macos;
```
spiessimon, line 190: Why is that one no longer needed? Because it is now explicitly bound at the top?

backend/amd64/simd.ml
```diff
@@ -82,95 +82,85 @@ module Seq = struct
...
+    | Vptestnzc_Y
 
   type nonrec t =
     { id : id;
-      instr : Amd64_simd_instrs.instr
+      instr : instr
     }
 
-  let sqrtss = { id = Sqrtss; instr = Amd64_simd_instrs.sqrtss }
+  open Amd64_simd_instrs
```
spiessimon: Can we move this `open` up?

oxcaml/tests/simd/amd64/avx_test_ops.ml
```diff
@@ -0,0 +1,114 @@
...
+    = "caml_vec256_unreachable" "caml_avx_vec256_testc"
+    [@@noalloc] [@@builtin]
+
+  external testnzc :
+    (int64x4[@unboxed]) -> (int64x4[@unboxed]) -> (int[@untagged])
+    = "caml_vec256_unreachable" "caml_avx_vec256_testnzc"
+    [@@noalloc] [@@builtin]
+end
+
+let () =
```
spiessimon, line 6: We are not testing that many combinations for these. That's probably not a problem, but if it's not too much trouble, it would be nice to add more test cases with expected results.

oxcaml/tests/simd/ops256.ml
```diff
@@ -0,0 +1,422 @@
...
+  let () =
+    (failmsg := fun () -> Printf.printf "interleave_32");
+    let _00000011 = Int32s.to_int32x8 0l 0l 0l 0l 0l 0l 1l 1l in
+    let _00110000 = Int32s.to_int32x8 0l 0l 1l 1l 0l 0l 0l 0l in
+    let _11000000 = Int32s.to_int32x8 1l 1l 0l 0l 0l 0l 0l 0l in
+    let _00001100 = Int32s.to_int32x8 0l 0l 0l 0l 1l 1l 0l 0l in
+    let _01010101 = Int32s.to_int32x8 0l 1l 0l 1l 0l 1l 0l 1l in
+    let _10101010 = Int32s.to_int32x8 1l 0l 1l 0l 1l 0l 1l 0l in
+    let res = interleave_high_32 _00110000 _00000011 in
+    eq4 (word0_32 res) (word1_32 res) (word2_32 res) (word3_32 res)
```
spiessimon, line 58: I stared a long time at this one. Could we create two `int64`s `v01` and `v10` and then use them for the projections instead. It's a bit subtle that we are projecting from different numbers below. (Perhaps we could also rename them to not start with underscore, due to the special handling of underscore in the compiler.)

oxcaml/tests/simd/ops256.ml
```diff
@@ -0,0 +1,422 @@
...
+      make64 0xffffffffffffffffL 0x8000000000000000L 0x7fffffffffffffffL 0x0L
+    in
+    let i0 = movemask_64 v0 in
+    eqi i0 0 0b0011 0;
+    ()
+
+  let () =
+    (failmsg := fun () -> Printf.printf "movemask_32");
+    let v0 =
+      Int32s.to_int32x8 (-1l) 1l 2l 0x80000000l (-1l) 1l 2l 0x80000000l
```
spiessimon: Can we use the same literal notation for all of them here?

oxcaml/tests/simd/ops256.ml
```diff
@@ -0,0 +1,422 @@
...
+    in
+    let i0 = movemask_32 v0 in
+    eqi i0 0 0b10011001 0
+
+  let () =
+    (failmsg := fun () -> Printf.printf "shuffle_64");
+    let _1234 = make64 1L 2L 3L 4L in
```
spiessimon: It confused me that `make64` reverses the order compared to `Int32s.to_int32x8`. Can we make them consistent in this file? I think I prefer the order of `make64`, because it matches the order of the projections.

oxcaml/tests/simd/ops256.ml
spiessimon: After looking at the `64`-bit version, I would say that one is much easier to understand.

oxcaml/tests/simd/ops256.ml
```diff
@@ -0,0 +1,422 @@
...
+  let () =
+    (failmsg := fun () -> Printf.printf "dup_64");
+    let v0 = make64 1L 2L 3L 4L in
+    let d0 = dup_even_64 v0 in
+    eq4 (word0 d0) (word1 d0) (word2 d0) (word3 d0) 1L 1L 3L 3L;
+    let v0 = Int32s.to_int32x8 0l 1l 2l 3l 4l 5l 6l 7l in
+    let d0 = dup_odd_32 v0 in
+    let d1 = dup_even_32 v0 in
+    eq4 (word0_32 d0) (word1_32 d0) (word2_32 d0) (word3_32 d0)
+      0x0000000100000001L 0x0000000300000003L 0x0000000500000005L
```
spiessimon, line 134: Instead of using `word0_32` ..., perhaps we could project out the `32`-bit integers here.

oxcaml/tests/simd/ops256.ml
```diff
@@ -0,0 +1,422 @@
...
+      0x1122334455667788L 0x1122334455667788L 0x1122334455667788L
+      0x1122334455667788L;
+    let v128_32 =
+      Int32s.to_int32x4 0x11223344l 0x55667788l 0x99aabbccl 0xddeeff00l
+    in
+    let v256_32 = broadcast_32 v128_32 in
+    eq4 (word0_32 v256_32) (word1_32 v256_32) (word2_32 v256_32)
+      (word3_32 v256_32) 0x1122334411223344L 0x1122334411223344L
+      0x1122334411223344L 0x1122334411223344L;
+    let v128_32_res = broadcast_32_128 v128_32 in
```
spiessimon, line 170: Probably that one should go to 128-bit tests, right?

oxcaml/tests/simd/ops256.ml
```diff
@@ -0,0 +1,422 @@
...
+    let v1 = insert_128 0 v256 new_low in
+    let v2 = insert_128 1 v256 new_high in
+    eq4 (word0 v1) (word1 v1) (word2 v1) (word3 v1) 0xaaaaaaaabbbbbbbbL
+      0xccccccccddddddddL 0x1111222233334444L 0x5555666677778888L;
+    eq4 (word0 v2) (word1 v2) (word2 v2) (word3 v2) 0x0123456789abcdefL
+      0xfedcba9876543210L 0xeeeeeeeeffffffffL 0x0000000011111111L
+
+  let () =
+    (failmsg := fun () -> Printf.printf "permute2_128");
+    let v0 = make64 0L 1L 2L 3L in
```
spiessimon: I think the instruction, for reasons that are beyond me, will zero out the result if the topmost control bit is set of the respective 4-bit segment. Perhaps the tests should also include an example that exercises this functionality.

oxcaml/tests/simd/ops256.ml
```diff
@@ -0,0 +1,422 @@
...
+  let () =
+    (failmsg := fun () -> Printf.printf "vzeroall/vzeroupper");
+    let vec256 =
+      make64 0x1122334455667788L 0x99aabbccddeeff00L 0xfedcba9876543210L
+        0x0123456789abcdefL
+    in
+    let vec128 = int64x2_of_int64s 0xaaaaaaaabbbbbbbbL 0xccccccccddddddddL in
+    let f32 = 1.25s in
+    let f64 = 2.5 in
+    vzeroupper ();
```
spiessimon, line 221: It seems this instruction should set the upper half to zero, but it doesn't  seem to have an effect. Is this intended? Analogously for `vzeroall` below.

oxcaml/tests/simd/ops256.ml
```diff
@@ -0,0 +1,422 @@
...
+    eqf f64 f64 2.5 2.5
+end
+
+module AVX2_Util = struct
+  include Builtins.AVX2
+
+  let () =
+    (failmsg := fun () -> Printf.printf "movemask_8");
+    let v0 =
+      Int8.to_int8x32 0xff 0x7f 0x80 0x0 0x1 0xcc 0x33 0x55 0xff 0x7f 0x80 0x0
```
spiessimon, line 247: Same here, reversing the order such that the bitstring below has the same order as the arguments would be helpful.

oxcaml/tests/simd/ops256.ml
```diff
@@ -0,0 +1,422 @@
...
+    (failmsg := fun () -> Printf.printf "shift_left_bytes");
+    let v0 =
+      Int8.to_int8x32 0x0 0x1 0x2 0x3 0x4 0x5 0x6 0x7 0x8 0x9 0xa 0xb 0xc 0xd
+        0xe 0xf 0x10 0x11 0x12 0x13 0x14 0x15 0x16 0x17 0x18 0x19 0x1a 0x1b 0x1c
+        0x1d 0x1e 0x1f
+    in
+    let v1 = shift_left_bytes 1 v0 in
+    eq4 (int8x32_first_int64 v1) (int8x32_second_int64 v1)
+      (int8x32_third_int64 v1) (int8x32_fourth_int64 v1) 0x060504030201000fL
+      0x0e0d0c0b0a090800L 0x161514131211101fL 0x1e1d1c1b1a191800L;
```
spiessimon: It seems `0x07` is lost in this shift left. I found that a bit counterintuitive, and I'm still not completely convinced that this is correct. Could we adjust this test such that it is a bit easier to see that we really shift left here? Same for the right shift below.

oxcaml/tests/simd/ops256.ml
```diff
@@ -0,0 +1,422 @@
...
+    (failmsg := fun () -> Printf.printf "blend_16");
+    let v0 = Int16.to_int16x16 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 in
+    let v1 =
+      Int16.to_int16x16 16 17 18 19 20 21 22 23 24 25 26 27 28 29 30 31
+    in
+    let b0 = blend_16 0b00000000 v0 v1 in
+    let b1 = blend_16 0b01010101 v0 v1 in
+    let b2 = blend_16 0b10101010 v0 v1 in
+    let b3 = blend_16 0b11111111 v0 v1 in
+    eq4 (int16x16_first_int64 b0) (int16x16_second_int64 b0)
```
spiessimon: I think this is another case where it would be good if the order in the declaration of the inital vectors was the same one as in the equality checks. And perhaps we could use the same literal notation for the 16-bit integers a in the equality comparison.

spiessimon (commented):
I've made my way through about half of this PR. It took me a long time to review the tests in `avx_test_ops.ml`. In many cases, I had a much harder time seeing that the transformation in the test is correct than just checking that the intrinsic that is used performs the desired effect. I wonder whether we could do something to make it easier to see in the tests which effect the operation has and that it is the desired one (e.g., converting the pseudo code from the reference manual to a reference implementation and comparing the two on a couple of inputs).

oxcaml/tests/simd/amd64/utils256.ml
TheNumbat: You get a compile error from `extract_constant` in `simd_selection`. Most of the intrinsics that require immediates are wrapped in `ocaml_simd` such that users can't provide invalid values, but insert/extract aren't. Eventually we'll have a static mode, which will let us provide better (frontend) errors here.

oxcaml/tests/simd/amd64/stub_builtins.c
TheNumbat: Ah right, the scalar round-with-mode intrinsics were added for use in base, which needs them to be portable (i.e. not rely on the amd64 rounding mode immediate). I guess they weren't tested in the compiler when added.
The vector rounding intrinsic isn't portable, so we don't need the rounding modes; deleted.

oxcaml/tests/simd/amd64/avx_test_ops.ml
TheNumbat: It's ported from `sse_test_ops.ml`, but both should be moved (done)

oxcaml/tests/simd/amd64/stub_builtins.c
TheNumbat: good catch, the intrinsic was actually wrong!

oxcaml/tests/simd/amd64/stub_builtins.c
TheNumbat: added

oxcaml/tests/simd/amd64/builtins.ml
TheNumbat: For AVX, yes - the 512-bit version will go under AVX512

oxcaml/tests/simd/amd64/builtins.ml
TheNumbat: yeah, moved to int8x16 (+ int16x8 + int32x4)

oxcaml/tests/simd/amd64/builtins.ml
TheNumbat: Moved to int32x4

oxcaml/tests/simd/amd64/builtins.ml
TheNumbat: Moved to int64x2

oxcaml/tests/simd/gen/gen_dune.ml
TheNumbat: Yeah, it's now included in everything else (same as `utils`)

oxcaml/tests/simd/amd64/avx_test_ops.ml
TheNumbat: I've added more tests in `ocaml_simd`

oxcaml/tests/simd/ops256.ml
TheNumbat: Removed the reversal, but yeah it's still confusing. It'd be better to reverse in `stubs256.c`, but I don't really want to change that everywhere

oxcaml/tests/simd/ops256.ml
TheNumbat: This uses AVX despite being 128-bit, so it can't go in ops (which also runs on arm)

oxcaml/tests/simd/ops256.ml
TheNumbat: Would be nicer, but ¯\\\_(ツ)\_/¯

oxcaml/tests/simd/ops256.ml
TheNumbat: strange, indeed

oxcaml/tests/simd/ops256.ml
TheNumbat: Yeah, these shouldn't have any visible effects - they force spill/reload for live variables. I think the only use of `vzeroupper` would be when mixing sse & avx instructions, and I'm not sure `vzeroall` is ever useful. Maybe we shouldn't provide them.

oxcaml/tests/simd/ops256.ml
TheNumbat: This instruction shifts within 2 x 128 bit lanes rather than the entire register - I'll make that clear in `ocaml_simd`.
Agree the test is very confusing, but not really sure how to make it readable short of reading out each byte individually

oxcaml/tests/simd/ops256.ml
TheNumbat: Swapped

oxcaml/tests/simd/ops256.ml
TheNumbat: Yeah, agreed but this is particularly annoying since we need a bit-packed result. Could add a bit-reverse function though.

TheNumbat: I at least made it more obvious that the bytes are the same

oxcaml/tests/simd/amd64/utils256.ml
spiessimon: Ah, okay, a compile error doesn't hurt much. I was more worried about weird behavior at runtime.

oxcaml/tests/simd/utils.ml
```diff
@@ -1,34 +1,38 @@
...
 let eql lv hv l h =
   if l <> lv then Printf.printf "%016lx <> %016lx\n" lv l;
   if h <> hv then Printf.printf "%016lx <> %016lx\n" hv h;
-  if l <> lv || h <> hv then !failmsg ()
+  if l <> lv || h <> hv then failure ()
 
 let eqi lv hv l h =
   if l <> lv
-  then Printf.printf "low:  expected = %016x <> %016x = actual\n" lv l;
+  then Printf.printf "low:  actual = 0x%016x <> 0x%016x = expected\n" lv l;
```
spiessimon, line 21: Was the order here wrong before? The order of arguments of the printing function did not change.

oxcaml/tests/simd/ops_int8x32.ml
```diff
@@ -12,15 +9,253 @@ external low_to : (t[@unboxed]) -> (int[@untagged])
...
+        Int8.to_int8x32 l l r r l l r r l l r r l l r r l l r r l l r r l l r r
+          l l r r
+      in
+      let v2 =
+        Int8.to_int8x32 l r l r l r l r l r l r l r l r l r l r l r l r l r l r
+          l r l r
+      in
+      let result = mul_unsigned_hadd_saturating_int16x16 v1 v2 in
+      let sum0 = Int16.adds (Int8.mulu_i16 l l) (Int8.mulu_i16 l r) in
+      let sum1 = Int16.adds (Int8.mulu_i16 r l) (Int8.mulu_i16 r r) in
```
spiessimon, line 208: I was a bit surprised that it was `Int8.mulu_i16 r l` instead of `Int8.mulu_i16 l r`, but I guess they are commutative, so it doesn't matter.

oxcaml/tests/simd/ops_int8x32.ml
```diff
@@ -12,15 +9,253 @@ external low_to : (t[@unboxed]) -> (int[@untagged])
...
+        expect expect expect expect);
+  Int8.check_ints (fun l r ->
+      (failmsg := fun () -> Printf.printf "%02x|%02x msadu\n%!" l r);
+      let v0 =
+        Int8.to_int8x32 l l r r l l r r l l r r l l r r l l r r l l r r l l r r
+          l l r r
+      in
+      let v1 =
+        Int8.to_int8x32 l r l r l r l r l r l r l r l r l r l r l r l r l r l r
+          l r l r
```
spiessimon: I think here only a subset of the integers are actually relevant, right? We could zero out the others, since as I understand it, the integer `0` down below selects which quadruplet of `v1` should be used for the operation.

oxcaml/tests/simd/ops_int64x4.ml
```diff
@@ -12,15 +9,187 @@ external low_to : t -> int64
...
+        (int64x4_third_int64 result)
+        (int64x4_fourth_int64 result)
+        (int64x4_first_int64 expect)
+        (int64x4_second_int64 expect)
+        (int64x4_third_int64 expect)
+        (int64x4_fourth_int64 expect));
+  Int64s.check_ints (fun l r ->
+      (failmsg := fun () -> Printf.printf "%016Lx|%016Lx << 7\n%!" l r);
+      let v = Utils256.int64x4_of_int64s l r l r in
+      let result = slli 7 v in
```
spiessimon: Why by `7` specifically?

oxcaml/tests/simd/ops_int64x4.ml
```diff
@@ -12,15 +9,187 @@ external low_to : t -> int64
...
+      eq4
+        (int64x4_first_int64 result)
+        (int64x4_second_int64 result)
+        (int64x4_third_int64 result)
+        (int64x4_fourth_int64 result)
+        (int64x4_first_int64 expect)
+        (int64x4_second_int64 expect)
+        (int64x4_third_int64 expect)
+        (int64x4_fourth_int64 expect));
+  Int64s.check_ints (fun l r ->
```
spiessimon, line 170: This test and the one below should probably be in a different place.

oxcaml/tests/simd/ops_int32x8.ml
```diff
@@ -12,15 +9,455 @@ external low_to : t -> int32
...
+        (int32x8_first_int64 expect)
+        (int32x8_second_int64 expect)
+        (int32x8_third_int64 expect)
+        (int32x8_fourth_int64 expect));
+  Int32s.check_ints (fun l r ->
+      (failmsg := fun () -> Printf.printf "%08lx|%08lx mul_even\n%!" l r);
+      let v1 = Int32s.to_int32x8 l r l r l r l r in
+      let v2 = Int32s.to_int32x8 r l r l r l r l in
+      let result = mul_even v1 v2 in
+      let expect = Int64.mul (Int64.of_int32 l) (Int64.of_int32 r) in
```
spiessimon, line 129: Multiplication should be commutative, so if I'm not mistaken, this test would also pass if the wrong part of the vector would be multiplied. Perhaps we can zero out the parts that will not be multiplied. Same below.

oxcaml/tests/simd/ops_int32x8.ml
```diff
@@ -12,15 +9,455 @@ external low_to : t -> int32
...
+        (int16x16_first_int64 expect)
+        (int16x16_second_int64 expect)
+        (int16x16_third_int64 expect)
+        (int16x16_fourth_int64 expect))
+
+let () =
+  Int32s.check_ints (fun l r ->
+      (failmsg := fun () -> Printf.printf "%08lx << %08lx\n%!" l r);
+      let v = Int32s.to_int32x8 l r l r l r l r in
+      let shift = Int32.logand r 0x1fl in
```
spiessimon: In the 64-bit tests, this is `63` so perhaps we should use `31` here instead of `0x1f`.

oxcaml/tests/simd/ops_int32x8.ml
```diff
@@ -12,15 +9,455 @@ external low_to : t -> int32
...
+        (int32x8_fourth_int64 result)
+        (int32x8_first_int64 expect)
+        (int32x8_second_int64 expect)
+        (int32x8_third_int64 expect)
+        (int32x8_fourth_int64 expect));
+  Int32s.check_ints (fun l r ->
+      (failmsg := fun () -> Printf.printf "%08lx|%08lx sllv\n%!" l r);
+      let v = Int32s.to_int32x8 l r l r l r l r in
+      let shifts =
+        Int32s.to_int32x8 (Int32.logand r 0x1fl) (Int32.logand l 0x1fl)
```
spiessimon: Same here. Alternatively, instead of sprinkling `31` everywhere, we could also make it a constant defined at the top of the file or wrap the operation in a function.

oxcaml/tests/simd/ops_int32x8.ml
```diff
@@ -12,15 +9,455 @@ external low_to : t -> int32
...
+      eq4
+        (int32x8_first_int64 result)
+        (int32x8_second_int64 result)
+        (int32x8_third_int64 result)
+        (int32x8_fourth_int64 result)
+        (int32x8_first_int64 expect)
+        (int32x8_second_int64 expect)
+        (int32x8_third_int64 expect)
+        (int32x8_fourth_int64 expect));
+  Int32s.check_ints (fun l r ->
```
spiessimon, line 416: Same here as for the int64 tests. These three should probably go to another file.

oxcaml/tests/simd/ops_int16x16.ml
```diff
@@ -12,15 +9,413 @@ external low_to : (t[@unboxed]) -> (int[@untagged])
...
+  (failmsg := fun () -> Printf.printf "%s: %04x | %04x\n%!" msg i0 i1);
+  let r0 = scalar i0 i1 in
+  let r1 = scalar i1 i0 in
+  let expect =
+    Int16.to_int16x16 r0 r1 r0 r1 r0 r1 r0 r1 r0 r1 r0 r1 r0 r1 r0 r1
+  in
+  let v1 = Int16.to_int16x16 i0 i1 i0 i1 i0 i1 i0 i1 i0 i1 i0 i1 i0 i1 i0 i1 in
+  let v2 = Int16.to_int16x16 i1 i0 i1 i0 i1 i0 i1 i0 i1 i0 i1 i0 i1 i0 i1 i0 in
+  let result = vector v1 v2 in
+  eq4
```
spiessimon, line 33: These kinds of equality functions show up several times, where we compare two larger vectors `64`-bit wise. Perhaps we could move them into the respective vector module. Doesn't have to be in this PR, but perhaps you could leave a CR.

oxcaml/tests/simd/ops_int16x16.ml
```diff
@@ -12,15 +9,413 @@ external low_to : (t[@unboxed]) -> (int[@untagged])
...
+        (int16x16_second_int64 expect)
+        (int16x16_third_int64 expect)
+        (int16x16_fourth_int64 expect));
+  Int16.check_ints (fun l r ->
+      (failmsg := fun () -> Printf.printf "%04x|%04x hadd\n%!" l r);
+      let v0 = Int16.to_int16x16 l l r r l l r r l l r r l l r r in
+      let v1 = Int16.to_int16x16 r r l l r r l l r r l l r r l l in
+      let result = hadd v0 v1 in
+      let expect =
+        Int16.to_int16x16 (Int16.add r r) (Int16.add l l) (Int16.add r r)
```
spiessimon: I'm confused. Why doesn't this start with `Int16.add l l`?

oxcaml/tests/simd/ops_int16x16.ml
```diff
@@ -12,15 +9,413 @@ external low_to : (t[@unboxed]) -> (int[@untagged])
...
+        (int16x16_second_int64 expect)
+        (int16x16_third_int64 expect)
+        (int16x16_fourth_int64 expect));
+  Int16.check_ints (fun l r ->
+      (failmsg := fun () -> Printf.printf "%04x|%04x hsub\n%!" l r);
+      let v0 = Int16.to_int16x16 l r l r l r l r l r l r l r l r in
+      let v1 = Int16.to_int16x16 r l r l r l r l r l r l r l r l in
+      let result = hsub v0 v1 in
+      let expect =
+        Int16.to_int16x16 (Int16.sub r l) (Int16.sub r l) (Int16.sub r l)
```
spiessimon: Same confusion with `Int16.sub l r` vs `Int16.sub r l`.

oxcaml/tests/simd/ops_int16x16.ml
spiessimon: I double checked with the `32x8`-version and that one starts with `l l`.

oxcaml/tests/simd/ops_int16x16.ml
```diff
@@ -12,15 +9,413 @@ external low_to : (t[@unboxed]) -> (int[@untagged])
...
+        (int32x8_third_int64 expect)
+        (int32x8_second_int64 expect)
+        (int32x8_first_int64 expect));
+  Int16.check_ints (fun l r ->
+      (failmsg := fun () -> Printf.printf "%04x|%04x hadd_saturating\n%!" l r);
+      let v0 = Int16.to_int16x16 l l r r l l r r l l r r l l r r in
+      let v1 = Int16.to_int16x16 r r l l r r l l r r l l r r l l in
+      let result = hadd_saturating v0 v1 in
+      let expect =
+        Int16.to_int16x16 (Int16.adds r r) (Int16.adds l l) (Int16.adds r r)
```
spiessimon: Same here and for subtraction below.

oxcaml/tests/simd/ops_int16x16.ml
```diff
@@ -12,15 +9,413 @@ external low_to : (t[@unboxed]) -> (int[@untagged])
...
+        (int16x16_second_int64 expect)
+        (int16x16_third_int64 expect)
+        (int16x16_fourth_int64 expect));
+  Int16.check_ints (fun l r ->
+      (failmsg := fun () -> Printf.printf "%04x|%04x mul_hadd\n%!" l r);
+      let v0 = Int16.to_int16x16 l l r r l l r r l l r r l l r r in
+      let v1 = Int16.to_int16x16 r r l l r r l l r r l l r r l l in
+      let result = mul_hadd v0 v1 in
+      let sum1 = Int32.add (Int16.mul_i32 l r) (Int16.mul_i32 l r) in
+      let sum2 = Int32.add (Int16.mul_i32 r l) (Int16.mul_i32 r l) in
```
spiessimon, line 287: I think commutativity here again has the effect that we are basically testing the same on both sides. Probably fine (so feel free to just resolve the comment), but if you have an idea how we could test more patterns there that could be a good idea.

oxcaml/tests/simd/ops_float64x4.ml
```diff
@@ -12,17 +11,216 @@ external low_to : t -> float
...
-
 let () =
   let v1 = low_of 1. in
   let v2 = low_of 2. in
-  let i1 = float64x4_low_int64 v1 in
-  let i2 = float64x4_low_int64 v2 in
-  eq i1 i2 0x3ff0000000000000L 0x4000000000000000L;
+  let i1 = float64x4_fourth_int64 v1 in
+  let i2 = float64x4_fourth_int64 v2 in
+  eq4 i1 i1 i1 i1 0x3ff0000000000000L 0x3ff0000000000000L 0x3ff0000000000000L
```
spiessimon: I guess this is to have better error messages compared to `Int64.equal`? Perhaps we could just add an `eq` function to `utils.ml` with this behavior. It's a bit strange to do the same comparison 4 times. I was initially searching for the difference.

oxcaml/tests/simd/ops_float64x4.ml
```diff
@@ -12,17 +11,216 @@ external low_to : t -> float
...
+    then Int64.float_of_bits 0xFFFFFFFFFFFFFFFFL, 1
+    else Int64.float_of_bits 0L, 0
+  in
+  let r1, m1 =
+    if scalar f1 f0
+    then Int64.float_of_bits 0xFFFFFFFFFFFFFFFFL, 1
+    else Int64.float_of_bits 0L, 0
+  in
+  let expect = Float64.to_float64x4 r0 r1 r0 r1 in
+  let expect_mask = m0 lor (m1 lsl 1) lor (m0 lsl 2) lor (m1 lsl 3) in
```
spiessimon, line 40: Can't say I understand why this is the right way of comparing floats. But it does match what we do for `float32x4`. Just to double check, the alternation between `m1` and `m0` is intended here, right?

oxcaml/tests/simd/ops_float64x4.ml
```diff
@@ -12,17 +11,216 @@ external low_to : t -> float
...
+  let result_i = Vector256_casts.int64x4_of_float64x4 result in
+  let mask = Builtins.AVX.movemask_64 result_i in
+  let expect_i = Vector256_casts.int64x4_of_float64x4 expect in
+  eqi mask mask expect_mask (Builtins.AVX.movemask_64 expect_i);
+  eq_float64x4 ~result ~expect
+
+let () =
+  let remove_nan p l r = p l r && not (Float.is_nan l || Float.is_nan r) in
+  let add_nan p l r = p l r || Float.is_nan l || Float.is_nan r in
+  Float64.check_floats
```
spiessimon, line 53: These tests have more descriptive names for `float32x4`. Can we use the same names here?

oxcaml/tests/simd/ops_float64x4.ml
```diff
@@ -12,17 +11,216 @@ external low_to : t -> float
...
+       (add_nan (fun l r -> Float.compare l r >= 0))
+       (fun l r -> cmp 5 l r));
+  Float64.check_floats
+    (check_cmp "6"
+       (add_nan (fun l r -> Float.compare l r = 1))
+       (fun l r -> cmp 6 l r));
+  Float64.check_floats
+    (check_cmp "7"
+       (fun l r -> (not (Float.is_nan l)) && not (Float.is_nan r))
+       (fun l r -> cmp 7 l r))
```
spiessimon, line 83: Perhaps adding the corresponding operations to Float would not be a bad idea. We do that for `Float32`.

oxcaml/tests/simd/ops_float64x4.ml
```diff
@@ -12,17 +11,216 @@ external low_to : t -> float
...
+
+let check_unop msg scalar vector f =
+  (failmsg := fun () -> Printf.printf "check_unop %s  %f\n%!" msg f);
+  let r = scalar f in
+  let expect = Float64.to_float64x4 r r r r in
+  let v = Float64.to_float64x4 f f f f in
+  let result = vector v in
+  eq_float64x4 ~result ~expect
+
+let () = Float64.check_floats (fun f _ -> check_unop "sqrt" Float.sqrt sqrt f)
```
spiessimon, line 111: Is there a reason we do not have `rcp` and `rsqrt` here as well? We do for the 32-bit version. Also, there's a typo in the `float32x4` version where it is called `rqrt` instead of `rsqrt`.

backend/amd64/simd_selection.ml
```diff
@@ -577,6 +582,291 @@ let select_operation_sse42 ~dbg:_ op args =
...
+    | "caml_avx_float64x2x2_hsub" -> instr vhsubpd_Y_Y_Ym256 args
+    | "caml_avx_float32x4x2_hsub" -> instr vhsubps_Y_Y_Ym256 args
+    | "caml_avx_vec256_insert_128" ->
+      let i, args = extract_constant args ~max:1 op in
+      instr vinsertf128 ~i args
+    | "caml_avx_float64x4_max" -> instr vmaxpd_Y_Y_Ym256 args
+    | "caml_avx_float32x8_max" -> instr vmaxps_Y_Y_Ym256 args
+    | "caml_avx_float64x4_min" -> instr vminpd_Y_Y_Ym256 args
+    | "caml_avx_float32x8_min" -> instr vminps_Y_Y_Ym256 args
+    | "caml_avx_vec256_dup_even_64" -> instr vmovddup_Y_Ym256 args
```
spiessimon, line 641: I'm a bit surprised that there is no odd version of this.

backend/amd64/simd_selection.ml
```diff
@@ -577,6 +582,291 @@ let select_operation_sse42 ~dbg:_ op args =
...
+    | "caml_avx_vec256_permute2_128" ->
+      let i, args = extract_constant args ~max:0xff op in
+      instr vperm2f128 ~i args
+    | "caml_avx_vec128_permutev_64" -> instr vpermilpd_X_X_Xm128 args
+    | "caml_avx_vec128x2_permutev_64" -> instr vpermilpd_Y_Y_Ym256 args
+    | "caml_avx_vec128_permutev_32" -> instr vpermilps_X_X_Xm128 args
+    | "caml_avx_vec128x2_permutev_32" -> instr vpermilps_Y_Y_Ym256 args
+    | "caml_avx_float32x8_rcp" -> instr vrcpps_Y_Ym256 args
+    | "caml_avx_float64x4_round" ->
+      let i, args = extract_constant args ~max:15 op in
```
spiessimon, line 670: I think we should probably consistently use hex or base-10 notation here (see `0x1f` above).

oxcaml/tests/simd/ops256.ml
spiessimon: I see.

oxcaml/tests/simd/ops256.ml
spiessimon: Hmm, yeah, sounds like providing them just causes confusion.

oxcaml/tests/simd/ops256.ml
spiessimon: Already quite a bit better. Perhaps we could just leave a comment like "read left to right" or something to that extent.

oxcaml/tests/simd/ops256.ml
spiessimon: I guess it's fine if the the tests in `ocaml_simd` are a bit clearer.

spiessimon (approved):
Looks good to me! Once the remaining comments are addressed, this should be good to go.

oxcaml/tests/simd/amd64/stub_builtins.c
spiessimon: Great, thanks!

oxcaml/tests/simd/amd64/avx_test_ops.ml
spiessimon: Thanks!

oxcaml/tests/simd/amd64/builtins.ml
spiessimon: Thank you!

oxcaml/tests/simd/amd64/avx_test_ops.ml
spiessimon: Okay, that sounds good :+1:

oxcaml/tests/simd/ops256.ml
TheNumbat: Let's leave them in the compiler but not expose them in `ocaml_simd` unless someone has an actual use case.

oxcaml/tests/simd/utils.ml
TheNumbat: Yeah, it was wrong

oxcaml/tests/simd/ops_int8x32.ml
TheNumbat: Yeah, `sum1` is used for the result of both orders

oxcaml/tests/simd/ops_int64x4.ml
TheNumbat: No particular reason

oxcaml/tests/simd/ops_int64x4.ml
TheNumbat: Would be better to put it in ops_int64x2, but those also run on arm. I've just added a comment

oxcaml/tests/simd/ops_int64x4.ml
mshinwell: Add a comment!

oxcaml/tests/simd/ops_int16x16.ml
TheNumbat: This was correct but even more confusing than usual. I ended up switching the order in `stubs256.c` and correcting everything. The results are much easier to verify now.

oxcaml/tests/simd/ops_int16x16.ml
TheNumbat: fixed

oxcaml/tests/simd/ops_int16x16.ml
TheNumbat: fixed

oxcaml/tests/simd/ops_float64x4.ml
TheNumbat: it is intended. the extra logic is just checking movemask on the result works properly.

oxcaml/tests/simd/ops_float64x4.ml
TheNumbat: These are using the stdlib Float, actually, not utils. But it could be added ¯\_(ツ)_/¯

oxcaml/tests/simd/ops_float64x4.ml
TheNumbat: yeah, they don't exist :^)

backend/amd64/simd_selection.ml
TheNumbat: indeed there isn't

backend/amd64/simd_selection.ml
TheNumbat: good point, fixed

oxcaml/tests/simd/ops256.ml
TheNumbat: It's clearer now that I switched the creation order

PR #4336, opened 2025-07-19
Sanitize simd load/store

TheNumbat:
First, fixes ASAN address checks for SIMD load/store. 
Clang generates the following checks for an unaligned 16-byte load:
```
        push    rax
        mov     rax, rdi
        shr     rax, 3
        movzx   eax, byte ptr [rax + 2147450880]
        test    al, al
        jne     .LBB1_1
.LBB1_2:
        lea     rax, [rdi + 15]
        mov     rcx, rax
        shr     rcx, 3
        movzx   ecx, byte ptr [rcx + 2147450880]
        test    cl, cl
        jne     .LBB1_3
.LBB1_4:
        vmovups xmm0, xmmword ptr [rdi]
        pop     rax
        ret
.LBB1_1:
        mov     ecx, edi
        and     cl, 7
        cmp     cl, al
        jl      .LBB1_2
        mov     esi, 16
        call    __asan_report_load_n@PLT
.LBB1_3:
        mov     edx, eax
        and     dl, 7
        cmp     dl, cl
        jl      .LBB1_4
        mov     esi, 16
        mov     rdi, rax
        call    __asan_report_load_n@PLT
```

We now generate this:
```	movq	%r12, %rdi
	movq	%rdi, %r11
	shrq	$3, %r11
	movzbq	2147450880(%r11), %r11
	testb	%r11b, %r11b
	je	.L152
	movq	%rdi, %r10
	andq	$7, %r10
	cmpb	%r11b, %r10b
	jl	.L152
	call	caml_asan_report_load16_noabort
.L152:
	leaq	15(%r12), %rdi
	movq	%rdi, %r11
	shrq	$3, %r11
	movzbq	2147450880(%r11), %r11
	testb	%r11b, %r11b
	je	.L153
	movq	%rdi, %r10
	andq	$7, %r10
	cmpb	%r11b, %r10b
	jl	.L153
	addq	$-15, %rdi
	call	caml_asan_report_load16_noabort
.L153:
	vmovupd	(%r12), %xmm0
```

This checks that the first and last bytes are valid. Both cases need the slow path because the address may not be 8-byte aligned. 
Aligned and unaligned 32 and 64 byte accesses do the same thing. I believe it'd be safe for aligned 32/64 to use the fast path for the first test, but I just went with what clang does.

Second, calls report function wrappers that save YMM / ZMM registers if any are in use. This required generating 3 copies of each function with sse/avx/avx512 target attributes.


backend/amd64/emit.ml
```diff
@@ -1290,24 +1299,58 @@ end = struct
...
-    | 8 -> Sym "caml_asan_report_load16_noabort"
-    | 9 -> Sym "caml_asan_report_store16_noabort"
-    | _ ->
-      (* CR-soon mslater: this is wrong, 32/64-byte operations should be routed
-         to [__asan_report_load_n_noabort]. Also, unaligned 16-byte operations
-         need a second check that the last byte's address is valid. *)
-      if access = 0
-      then Sym "caml_asan_report_load16_noabort"
-      else Sym "caml_asan_report_store16_noabort"
+    match simd_regs with
```
Svetlitski, line 1302: It seems like it would be much cleaner to multiversion these symbols in C so that we can have just one set of names here on the OCaml side. Since we're only supporting ASAN under clang, you can use [the `target_clones` attribute ](https://clang.llvm.org/docs/AttributeReference.html#target-clones) on the stubs and clang will do the multiversioning for you!

TheNumbat: Good idea, but apparently `target_clones` is entirely incompatible with `no_caller_saved_registers`, so we'd have to generate the cpuid check ourselves. We could do so but I don't think it'd be much of an improvement over using different names in OCaml.

runtime/memory.c
```diff
@@ -1127,22 +1127,41 @@ __asan_default_options(void) {
...
-CREATE_ASAN_REPORT_WRAPPER(store, 1)
-CREATE_ASAN_REPORT_WRAPPER(store, 2)
-CREATE_ASAN_REPORT_WRAPPER(store, 4)
-CREATE_ASAN_REPORT_WRAPPER(store, 8)
-CREATE_ASAN_REPORT_WRAPPER(store, 16)
+// The point of these wrappers is to spill all caller-saved registers
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wexcessive-regsave"
+
+void __asan_report_load_n_noabort(const void* addr, size_t size);
```
Svetlitski, line 1134: Why do you have to explicitly declare this, whereas none of the others have to be declared? Are you sure this isn't just a holdover from having compiled with GCC locally or something?

TheNumbat: They're now all using `load/store_n` instead of the size-specific symbols

Svetlitski (approved):

PR #4337, opened 2025-07-21
Llvmize simple operations

usuyus:
This PR adds the following to `Llvmize`
- Integer operations
- Float operations
- Load, store, and address calculations
- Tests for all the above


backend/amd64/cfg_selection.ml
```diff
@@ -334,102 +334,107 @@ let select_operation
...
-    let locality : Arch.prefetch_temporal_locality_hint =
-      match select_locality locality with
-      | Moderate when is_write && not (Arch.Extension.enabled PREFETCHWT1) ->
-        High
-      | l -> l
-    in
-    let addr, eloc = select_addressing Word_int (one_arg "prefetch" args) in
-    Rewritten (specific (Iprefetch { is_write; addr; locality }), [eloc])
-  | _ -> Use_default
+  if !Oxcaml_flags.llvm_backend
```
mshinwell, line 444: I suggest renaming the original function here, then moving this conditional into a new function which calls the old one.  This will avoid indenting everything and the big diff.

usuyus: Is it OK now?

mshinwell: Yeah that looks better, thanks

oxcaml/tests/backend/llvmize/array_rev.ml
```diff
@@ -0,0 +1,10 @@
+let arr = Sys.opaque_identity Array_rev_data.arr
+
+let len = Sys.opaque_identity Array_rev_data.len
```
gretay-js: Do we need `Sys.opaque_identity` here even if `Array_rev_data` is compiled with `-opaque`?

oxcaml/tests/backend/llvmize/dune.inc
```diff
@@ -63,3 +52,105 @@
...
+ (action
+  (diff gcd.output gcd.output.corrected)))
+
+(rule
+ (enabled_if (and (= %{context_name} "main") (= %{architecture} "amd64")))
+ (targets array_rev.output.exe array_rev_ir.output.corrected)
+ (deps array_rev_data.ml array_rev_main.ml array_rev.ml)
+ (action
+  (progn
+   (run %{bin:ocamlopt.opt} array_rev_data.ml -g -O3 -opaque)
```
gretay-js:
```suggestion
   (run %{bin:ocamlopt.opt} array_rev_data.ml -c -g -O3 -opaque)
```

oxcaml/tests/backend/llvmize/gcd_main.ml
```diff
@@ -0,0 +1,12 @@
+let try_gcd x y =
+  Gcd_data.x := x;
+  Gcd_data.y := y;
+  Gcd.gcd ();
+  Format.printf "gcd(%d, %d) = %d\n" x y !Gcd_data.res
+
+let () =
+  try_gcd 216 250 (* 2 *);
+  try_gcd 37 37 (* 30 *);
```
gretay-js:
```suggestion
  try_gcd 37 37 (* 37 *);
```

backend/arm64/arch.ml
```diff
@@ -115,6 +115,10 @@ let num_args_addressing = function
   | Iindexed _ -> 1
   | Ibased _ -> 0
 
+let addressing_displacement_for_llvmize = function
+  | Iindexed d -> d
+  | Ibased (_, d) -> d
```
gretay-js: This case should be fatal_error, like in amd64.

backend/amd64/arch.ml
```diff
@@ -318,6 +318,15 @@ let num_args_addressing = function
   | Iscaled _ -> 1
   | Iindexed2scaled _ -> 2
 
+let addressing_displacement_for_llvmize = function
```
gretay-js: Check at the beginning that `Oxcaml.llvm_backend` is set.

backend/amd64/cfg_selection.ml
```diff
@@ -431,6 +433,15 @@ let select_operation
     Rewritten (specific (Iprefetch { is_write; addr; locality }), [eloc])
   | _ -> Use_default
 
+let select_operation
```
gretay-js, line 439: Please make similar changes to arm64. An alternative would be to make these changes in `cfg_selectgen` but there are a few calls and we may need to do something special for builtins, so it currently seems easier to do this check per target.

oxcaml/tests/backend/llvmize/int_ops_main.ml
```diff
@@ -0,0 +1,19 @@
...
+(-15) lor 4 = %d
+(-15) lxor 4 = %d
+lnot (-15) = %d
+(-15) lsl 4 = %d
+(-15) lsr 4 = %d
+(-15) asr 4 = %d
+  |}
+    (Int_ops.add ()) (Int_ops.sub ()) (Int_ops.mul ()) (Int_ops.div ())
+    (Int_ops.mod_ ()) (Int_ops.land_ ()) (Int_ops.lor_ ()) (Int_ops.lxor_ ())
+    (Int_ops.lnot_ ()) (Int_ops.lsl_ ()) (Int_ops.lsr_ ()) (Int_ops.asr_ ())
```
gretay-js: nitpicking: please arrange it similarly to float op test.

oxcaml/tests/backend/llvmize/int_ops.ml
```diff
@@ -0,0 +1,29 @@
...
+let[@inline never] div () = if y = 0 then 0 else x / y
+
+let[@inline never] mod_ () = x mod y
+
+let[@inline never] land_ () = x land y
+
+let[@inline never] lor_ () = x lor y
+
+let[@inline never] lor_ () = x lor y
+
```
gretay-js: These two look the same.
```suggestion
let[@inline never] lor_ () = x lor y

```

backend/cfg/llvmize.ml
```diff
@@ -105,8 +137,12 @@ type t =
     mutable current_fun_info : fun_info;
         (* Maintains the state of the current function (reset for every
            function) *)
-    mutable data : Cmm.data_item list
+    mutable data : Cmm.data_item list;
         (* Collects data items as they come and processes them at the end *)
+    mutable defined_symbols : String.Set.t;
+        (* Keeps track of all function symbols so far *)
```
gretay-js: Is this only for function symbols or symmetric to `referenced_symbols`?
```suggestion
        (* Keeps track of all global symbols so far *)
```

backend/cfg/llvmize.ml
```diff
@@ -15,19 +15,22 @@
...
 
 type error = Asm_generation of (string * int)
 
 exception Error of error
 
 module Ident : sig
-  (** Unnamed LLVM identifiers that start with "%".  This includes
-      basic blocks, unnamed function parameters, and temporaries
-      (virtual registers).  *)
+  (** LLVM identifiers that start with "%".  This includes
```
gretay-js: Have you considered having a separate module for `Named_ident`?

backend/cfg/llvmize.ml
gretay-js: Discussed offline, this is for function symbols defined so far in the current compilation unit. Data symbols are not defined until the end. My suggestion is to change it to `defined_function_symbols`. A followup PR will add another data structure for `external` functions that needed to be declared because they are used in `extcall`. The LLVM type of the extcall can be obtained from `Cfg.external_call_operation.ty_args` and `ty_res`.

oxcaml/tests/backend/llvmize/int_ops.ml
```diff
@@ -1,29 +1,25 @@
-let x = Sys.opaque_identity (-15)
+module D = Int_ops_data
```
gretay-js, line 1: It would be nice to have tests that exercise `Intop_imm`.

oxcaml/tests/backend/llvmize/float_ops.ml
```diff
@@ -0,0 +1,26 @@
...
+  let[@inline always] mul x y = of_float (Float.mul (to_float x) (to_float y))
+
+  let[@inline always] div x y = of_float (Float.div (to_float x) (to_float y))
+
+  let[@inline always] neg x = of_float (Float.neg (to_float x))
+
+
+  let[@inline always] abs x = of_float (Float.abs (to_float x))
+
+  let[@inline always] compare x y = (Float.compare (to_float x) (to_float y))
```
gretay-js: remove `[@inline always]`  or change it to `[@inline never]` would makes more sense even though it is not needed because this file is compiled with `-opaque` .

oxcaml/tests/backend/llvmize/float_ops_main.ml
```diff
@@ -0,0 +1,26 @@
...
+let print_int_newline n = print_int n; print_newline ()
+
+let () =
+  float_u_test "3.7 + 2.3" (F.add (#3.7) (#2.3));
+  float_u_test "3.7 - 2.3" (F.sub (#3.7) (#2.3));
+  float_u_test "3.7 * 2.3" (F.mul (#3.7) (#2.3));
+  float_u_test "3.7 / 2.3" (F.div (#3.7) (#2.3));
+  float_u_test "abs(-3.7)" (F.abs (F.of_float (-3.7)));
+  float_u_test "neg(3.7)" (F.neg (#3.7));
+  compare_test 3.7 2.3;
```
gretay-js, line 24: It may be worth having tests for compare with some of the special values (NaNs and +Inf, -Inf, +0, -0), I'm not sure.

oxcaml/tests/backend/llvmize/dune.inc
```diff
@@ -154,3 +158,38 @@
...
+(rule
+ (enabled_if (and (= %{context_name} "main") (= %{architecture} "amd64")))
+ (targets float_ops.output.exe float_ops_ir.output.corrected)
+ (deps  float_ops_main.ml float_ops.ml)
+ (action
+  (progn
+   
+   (run %{bin:ocamlopt.opt} float_ops.ml -c -g -O3 -opaque -llvm-backend -llvm-path clang -keep-llvmir -dno-asm-comments)
+   (run %{bin:ocamlopt.opt} float_ops_main.ml -c -g -O3 -opaque)
+   (run %{bin:ocamlopt.opt}  float_ops.cmx float_ops_main.cmx -opaque -o float_ops.output.exe)
```
gretay-js: Consider adding more options here to help debugging, such as "-S -dump-into-file -dcmm -dcfg -dlinear".

backend/arm64/arch.ml
```diff
@@ -115,9 +115,18 @@ let num_args_addressing = function
...
-  | Iindexed d -> d
-  | Ibased (_, d) -> d
+let addressing_displacement_for_llvmize addr =
+  if not !Oxcaml_flags.llvm_backend
+  then
+    Misc.fatal_error
+      "Arch.displacement_addressing_for_llvmize: should only be called with \
+        -llvm-backend"
+  else
+    match addr with
```
gretay-js, line 125: Update `Cfg_selection.select_operation`  and `select_addressing` in arm64 backend to check `Oxcaml.llvm_backend` flag, similarly to the way it is done in `amd64`.

backend/cfg/llvmize.ml
gretay-js: Discussed offline. Looks like `Ident` is used in the same way so far, whether it's named or not, and the difference is only in creation, so it's not worth having a separate module at this time.  Splitting changes related to identifiers out of this PR will make it easier to review.

gretay-js (commented):
I reviewed everything except the changes in `llvmize.ml`. I'll review them after rebase over #4346 .

backend/amd64/cfg_selection.ml
```diff
@@ -255,6 +255,8 @@ let select_addressing (_chunk : Cmm.memory_chunk) exp :
   (* PR#4625: displacement must be a signed 32-bit immediate *)
   if not (int_is_immediate d)
   then Iindexed 0, exp
+  else if !Oxcaml_flags.llvm_backend (* Llvmize only expects [Iindexed] *)
```
gretay-js: Please make the change to `select_addressing` in the same way that you made the change to `select_operation` below, by adding another function just for the check. It will make the normal compilation completely independent. Same for arm64.

backend/cfg/llvmize.ml
```diff
@@ -292,22 +326,132 @@ module F = struct
 
   let ins_branch t label = ins t "br %a" (pp_label t) label
 
-  let ins_alloca t (reg : Reg.t) =
-    ins t "%a = alloca %a" (pp_reg_ident t) reg pp_machtyp_component reg.typ
+  (* [dbg] is used to print the [Reg.t] the allocated identifier corresponds to
+     in the preamble *)
+  let ins_alloca ?dbg t (reg : Reg.t) =
```
gretay-js: Very minor, could be done in another PR: `dbg` is a name commonly used for `Debuginfo.t`, it had an unexpected meaning here. 
```suggestion
  let ins_alloca ?regname  t (reg : Reg.t) =
```

backend/cfg/llvmize.ml
```diff
@@ -292,22 +326,132 @@ module F = struct
...
+      src Llvm_typ.pp_t dst_typ
+
+  let ins_binop t op arg1 arg2 res typ =
+    ins t "%a = %s %a %a, %a" pp_ident res op Llvm_typ.pp_t typ pp_ident arg1
+      pp_ident arg2
+
+  let ins_binop_imm t op arg imm res typ =
+    ins t "%a = %s %a %a, %s" pp_ident res op Llvm_typ.pp_t typ pp_ident arg imm
+
+  let ins_monop t op arg res typ =
```
gretay-js: Terminology used more commonly is "unary" operations.
```suggestion
  let ins_unary_op t op ~arg ~res typ =
```

backend/cfg/llvmize.ml
```diff
@@ -292,22 +326,132 @@ module F = struct
...
+  let not_implemented_basic = not_implemented_aux Cfg.print_basic
+
+  let not_implemented_terminator = not_implemented_aux Cfg.print_terminator
+
   (* == Cfg instructions == *)
   (* CR-soon yusumez: Add implementations for missing basic and terminator
      instructions *)
 
   let terminator t (i : Cfg.terminator Cfg.instruction) =
+    pp_dbg_instr_terminator t.ppf i;
```
gretay-js: No need to print twice
```suggestion
```

backend/cfg/llvmize.ml
```diff
@@ -274,14 +303,19 @@ module F = struct
...
+    ins t "store %a %a, ptr %a" Llvm_typ.pp_t typ pp_ident src pp_ident dst
 
-  let ins_store t ident (reg : Reg.t) =
-    ins t "store %a %a, ptr %a" pp_machtyp_component reg.typ pp_ident ident
-      (pp_reg_ident t) reg
+  let ins_load_reg t ident (reg : Reg.t) =
+    ins_load t ~src:(get_ident_for_reg t reg) ~dst:ident
+      (Llvm_typ.of_machtyp_component reg.typ)
+
+  let ins_store_reg t ident (reg : Reg.t) =
```
gretay-js: It'd be nice to have labeled `~src` and `~dst` args for the other store/load functions, to avoid confusion.

backend/cfg/llvmize.ml
```diff
@@ -292,22 +326,132 @@ module F = struct
...
+
+  let ins_conv t op ~src ~dst ~src_typ ~dst_typ =
+    ins t "%a = %s %a %a to %a" pp_ident dst op Llvm_typ.pp_t src_typ pp_ident
+      src Llvm_typ.pp_t dst_typ
+
+  let ins_binop t op arg1 arg2 res typ =
+    ins t "%a = %s %a %a, %a" pp_ident res op Llvm_typ.pp_t typ pp_ident arg1
+      pp_ident arg2
+
+  let ins_binop_imm t op arg imm res typ =
```
gretay-js, line 355:
```suggestion
  let ins_binop_imm t op ~arg imm ~res typ =
```

backend/cfg/llvmize.ml
```diff
@@ -292,22 +326,132 @@ module F = struct
...
+
   let load_reg_to_temp t reg =
     let temp = fresh_ident t in
-    ins_load t temp reg;
+    ins_load_reg t temp reg;
     temp
 
+  let load_addr t (addr : Arch.addressing_mode) (i : 'a Cfg.instruction) n =
+    let offset = Arch.addressing_displacement_for_llvmize addr in
+    let arg = load_reg_to_temp t i.arg.(n) in
```
gretay-js, line 380: Looks like `i` and `n` are only used to in `i.arg.(n)` so the caller can do it:
```suggestion
  let load_addr t (addr : Arch.addressing_mode) (reg : Reg.t) =
    let offset = Arch.addressing_displacement_for_llvmize addr in
    let arg = load_reg_to_temp t reg in
```

gretay-js: I guess it was needed to handle more interesting addressing modes.. ok to keep as is, just in case there is some reason to do it again later.

gretay-js (approved):

backend/cfg/llvmize.ml
usuyus: I wanted to keep it like that for consistency with `ins_binop`

backend/cfg/llvmize.ml
usuyus: I didn't do that since we always load from and store to registers. Maybe I can change their names to reflect that instead?

PR #4340, opened 2025-07-21
Correctly handle the compilation of default parameters in generic optional arguments

zchenb:
1. Add a module path to `Tparam_optional_default`. 
2. The translation (`for_optional_arg_default`) now calls the correct match based on this path.

### DIFF
--- a/typing/typecore.ml
+++ b/typing/typecore.ml
@@ -884,7 +884,6 @@ let constant_or_raise env loc cst =
       c
   | Error err -> raise (Error (loc, env, err))
 
-
 (* Specific version of type_option, using newty rather than newgenty *)
 
 let type_option mpath ty =


lambda/matching.ml
```diff
@@ -4496,6 +4496,15 @@ let for_optional_arg_default
...
   in
   let supplied_or_default =
+    match mpath with
+    | Stdlib_or_null ->
+    transl_match_on_or_null
+      default_arg_layout
+      (Lvar param)
+      Loc_unknown
+      ~if_null:default_arg
+      ~if_this:(Lvar param) (* CR generic-optional: CHECK THIS *)
```
nmatschke: This should all be indented two characters past the start of `Stdlib_or_null`, as with other `match` cases in this file. The same applies for the case below, even though you didn't modify that code.

lambda/matching.ml
```diff
@@ -4484,7 +4484,7 @@ let for_multiple_match ~scopes ~return_layout loc paraml mode pat_act_list parti
     (do_for_multiple_match ~scopes ~return_layout loc paraml mode pat_act_list
        partial)
 
-let for_optional_arg_default
+let for_optional_arg_default (mpath : Btype.optional_module_path)
```
nmatschke: I think this should be a labeled argument (since e.g. `default_arg` is) and likely on the next line or the following. Look at how other functions in this file with many arguments are formatted.

chamelon/compat.ox.ml
```diff
@@ -144,7 +144,13 @@ let mkTexp_function ?(id = texp_function_defaults)
                 (match optional_default with
                 | None -> Tparam_pat pattern
                 | Some default ->
-                    Tparam_optional_default (pattern, default, id.param_sort));
+                    let mpath =
+                      match Btype.classify_optionality arg_label with
+                      | Optional_arg mpath -> mpath
+                      | Required_or_position_arg -> assert false
```
nmatschke: It's sad to repeat this code in several places. How about adding a function to `Btype` which returns an option and then using `Option.value_exn`? Or a function in `Btype` which does the raising itself, though that's a bit subtler for the reader.

typing/typedtree.mli
```diff
@@ -562,11 +562,12 @@ and function_param =
...
-  | Tparam_optional_default of pattern * expression * Jkind.sort
-  (** [Tparam_optional_default (p, e, sort)] is an optional argument [p] with
-      default value [e], i.e. [?x:(p = e)]. If the parameter is of type
+  | Tparam_optional_default of
+      pattern * expression * Jkind.sort * Btype.optional_module_path
+  (** [Tparam_optional_default (p, e, sort, mpath)] is an optional argument [p]
+      with default value [e], i.e. [?x:(p = e)]. If the parameter is of type
       [a option], the pattern and expression are of type [a]. [sort] is the
-      sort of [e]. *)
+      sort of [e]. [mpath] is the generic optioal module path *)
```
nmatschke: `optioal` spelling

nmatschke: Also this sentence should end in a period.

nmatschke: Perhaps give an example of how such code would be written by a human, as in the `[?x:(p = e)]` example given above.

typing/typecore.ml
```diff
@@ -7522,7 +7522,8 @@ and type_function
             let default_arg =
               type_expect env mode_legacy default (mk_expected ty_default_arg)
             in
-            ty_default_arg, Some (default_arg, arg_label, default_arg_sort)
+            ty_default_arg,
+              Some (default_arg, arg_label, default_arg_sort, mpath)
```
nmatschke: I think this should probably not be indented.

typing/tast_iterator.ml
```diff
@@ -302,7 +302,7 @@ let function_param sub { fp_loc; fp_kind; fp_newtypes; _ } =
     fp_newtypes;
   match fp_kind with
   | Tparam_pat pat -> sub.pat sub pat
-  | Tparam_optional_default (pat, default_arg, _) ->
+  | Tparam_optional_default (pat, default_arg, _, _) ->
```
nmatschke, line 305: If this becomes a regular module path someday, I imagine we'll want to recur on it here. Maybe worth leaving a note so we don't forget.

nmatschke (changes requested):

lambda/matching.ml
zchenb: Just to clarify, is indentation-only change not considered diff by the upstream? [Otherwise by the "minimizing diff" principle, I should not change indentations].

chamelon/compat.ox.ml
zchenb: This is a good suggestion. I added a function `get_optional_module_path_exn` in `Btype`.

lambda/matching.ml
dkalinichenko-js: Yeah, here I'd say you shouldn't change the other cases. I would argue for this even to make the review easier, but minimizing the git diff with the upstream OCaml compiler is an additional consideration. And git diffs do consider whitespace.

nmatschke: That's pretty surprising to me. Will the whitespace diffs even show up in review? If they do, won't it be obvious they're whitespace diffs? It makes reading the code harder in the future, which is typically more valuable than making this one PR easier to review. Also, there must necessarily be a big diff in the below case relative to upstream, due to the two comments.

nmatschke: (Is there really not a merge tool that can handle whitespace diffs?)

dkalinichenko-js: There are, but merges from upstream run into a lot of problems, so we err towards trying keeping the existing compiler code unchanged. I'm not sure if this convention is still relevant, but for now I think it's best not to touch this.

dkalinichenko-js: Oh, wait, nvm, here I agree that you should change this due to the pattern-match.

dkalinichenko-js: We do want to get a conflict here if something's changed.

zchenb: I am still confused on reasoning. Is the principle that "minimize diff unless you are in a pattern match, in which case you should change indentation instead of minimizing diff"?

dkalinichenko-js: I wouldn't say there are hard rules here. Instead, there are multiple counterbalancing considerations:

1. We want our code to look nice and have a consistent style. This points towards identing the pattern-match.
2. We want the PR diff to be easily reviewable. Normally, this points towards only doing changes related to the goal of the PR, and leaving unrelated code untouched. I'd say the code in the pattern-match is related to this change, since it's now only called conditionally instead of always. This points towards identing the pattern-match.
3. We want to make merging future OCaml compiler releases less painful. Normally, this points towards not changing the existing compiler code, to avoid merge conflicts. However, we only want to avoid merge conflicts in situations where we didn't change the semantic meaning of the original code. Here, we did change the semantic meaning of the code, so we actually do want to get a merge conflict if upstream also changes it, and resolve the conflict manually. So this also points towards identing the pattern-match.

Sorry for sometimes confusing and contradictory review comments! I should have explained the reasoning behind them in more detail. I don't think we have a written style guide for the compiler, and so most of this is based on implicit understanding which may diverge between different reviewers and applies differently depending on the situation.

lambda/matching.ml
```diff
@@ -4496,22 +4496,31 @@ let for_optional_arg_default
...
-            [ Lvar param ],
-            Loc_unknown))
+    match mpath with
+    | Stdlib_or_null ->
+        transl_match_on_or_null
+          default_arg_layout
+          (Lvar param)
+          Loc_unknown
+          ~if_null:default_arg
+          ~if_this:(Lvar param) (* CR generic-optional: CHECK THIS *)
```
dkalinichenko-js: This line seems good, as discussed, I think.

lambda/translcore.ml
```diff
@@ -1656,9 +1656,15 @@ and transl_curried_function ~scopes loc repr params body
...
           | Tparam_pat pat ->
               pat.pat_env, pat.pat_type, Translattribute.transl_param_attributes pat
-          | Tparam_optional_default (pat, expr, _) ->
-              expr.exp_env, Predef.type_option expr.exp_type, Translattribute.transl_param_attributes pat
+          | Tparam_optional_default (pat, expr, _, mpath) ->
+              expr.exp_env,
+              (match mpath with
+              | Stdlib_option -> Predef.type_option expr.exp_type
+              | Stdlib_or_null -> Predef.type_or_null expr.exp_type),
+              (* CR generic-optional : CHECK THIS *)
```
dkalinichenko-js: This seems good.

lambda/translcore.ml
```diff
@@ -1678,13 +1684,15 @@ and transl_curried_function ~scopes loc repr params body
                 fp_partial
                 ~arg_sort:fp_sort ~arg_layout
                 ~return_layout
-          | Tparam_optional_default (pat, default_arg, default_arg_sort) ->
+          | Tparam_optional_default
+              (pat, default_arg, default_arg_sort, mpath) ->
+              (* CR generic-optional : CHECK THIS *)
               let default_arg_sort = Jkind.Sort.default_for_transl_and_get default_arg_sort in
```
dkalinichenko-js, line 1688: I think we'll need to change this once we introduce truly generic optional arguments, but for now it's fine.

typing/uniqueness_analysis.ml
```diff
@@ -2156,7 +2156,8 @@ let rec check_uniqueness_exp ~overwrite (ienv : Ienv.t) exp : UF.t =
               let value = Match_single (Paths.fresh ()) in
               let ext, uf_pat = pattern_match pat value in
               ext, (UF.unused, uf_pat)
-            | Tparam_optional_default (pat, default, _) ->
+            | Tparam_optional_default (pat, default, _, _) ->
+              (* CR generic-optional : CHECK THIS *)
```
dkalinichenko-js: This seems fine.

typing/untypeast.ml
```diff
@@ -535,7 +535,8 @@ let expression sub exp =
                let pat, default_arg =
                  match fp.fp_kind with
                  | Tparam_pat pat -> pat, None
-                 | Tparam_optional_default (pat, expr, _) -> pat, Some expr
+                 (* CR generic-optional : Check This *)
+                 | Tparam_optional_default (pat, expr, _, _) -> pat, Some expr
```
dkalinichenko-js, line 538: I think this looks fine? In any case, `Untypeast` is not used much.

lambda/matching.ml
zchenb: Thank you very much! This explanation is now very clear on when I should change/not change the existing code.

dkalinichenko-js (approved):

PR #4342, opened 2025-07-21
Avoid closure in the mode solver

### LINKED ISSUE #4323: Typer performance regression
While trying to investigate the (non)-linear performance behavior of some flambda2 algorithms, I wrote the a script (see below) to generate a very dumb file with two branches, each containing a bunch of calls to the `%setfield0` primitive.

I did find non-linearities however to my surprise they were not in flambda2 (according to the output of `-dprofile`):

| Size | OCaml 5.3 (typing, s) | OxCaml 5.2.0minus-2 (typing, s) | OxCaml bd19d0d91a3cac79132a721a686f87f7325fad08 (typing, s) |
| -- | -- | -- | -- |
| 1024 | 0.019 | 0.039 | 0.104 |
| 2048 | 0.041 | 0.079 | 0.342 |
| 4096 | 0.087 | 0.169 | 1.135 |
| 8192 | 0.176 | 0.343 | 4.569 |

I'd expect typing on this file to take time proportional to the number of calls to `%setfield0`, and both stock OCaml and OxCaml 5.2.0minus-2 (I had this one lying around) seem to agree. However current `main` seems to be showing some pretty bad non-linearities.

```python
#!/usr/bin/env python3

def generate(*, size: int = 1024):
    print("type t")
    print("external setfield0 : t -> int -> unit = \"%setfield0\"")
    print("")
    print("let f g b x =")
    print("  let y =")
    print("    if b then")
    for i in range(size):
        print("      let _u = setfield0 x 0 in")
    print("      x")
    print("    else")
    for i in range(size):
        print("      let _u = setfield0 x 0 in")
    print("      x")
    print("  in")
    print("  g y")

def main():
    import argparse

    parser = argparse.ArgumentParser()
    parser.add_argument('size', nargs='?', default=1024, type=int)

    ns = parser.parse_args()
    generate(size=ns.size)

if __name__ == "__main__":
    main()
```

riaqn:
Fixes #4323 

# Benchmark
Use the script in #4323 to generate `test.ml` with size 8192.

Runs `_build/_bootinstall/bin/ocamlopt.opt test.ml -nostdlib -nopervasives -c -dtimings`

Before PR, `typing` takes 5.927s.
After this PR, `typing` takes 0.661s.

### DIFF
--- a/typing/solver.ml
+++ b/typing/solver.ml
@@ -430,8 +430,9 @@ module Solver_mono (C : Lattices_mono) = struct
           let src = C.src dst g in
           let g'f = C.compose src g' (C.disallow_right f) in
           let x = Amorphvar (v, g'f) in
-          if not (VarMap.exists (fun _ mv -> eq_morphvar mv x) u.vlower)
-          then set_vlower ~log u (VarMap.add (get_key x) x u.vlower);
+          let key = get_key x in
+          if not (VarMap.mem key u.vlower)
+          then set_vlower ~log u (VarMap.add key x u.vlower);
           Ok ())
 
   let cnt_id = ref 0


dkalinichenko-js (approved):
Looks good to me, although I agree that this doesn't reach O(n).

dkalinichenko-js:
Would be good to have a benchmark of this on the whole tree before rolling – I'm not sure, but could there be scenarios where this fix is even slower?

riaqn:
@dkalinichenko-js I think this can't be slower than the original code.

PR #4344, opened 2025-07-21
Reinterpret vector widths

TheNumbat:
Extends vector reinterpret-cast to allow converting between widths. When converting a smaller vector to a larger vector, the upper bits are left unspecified (they'll be zero if a move is generated, or the existing bits if coalesced).

### DIFF
--- a/backend/cmm_builtins.ml
+++ b/backend/cmm_builtins.ml
@@ -335,11 +335,11 @@ let transl_vec_builtin name args dbg _typ_res =
   | "caml_vec512_low_to_vec128" ->
     let op = Creinterpret_cast (V128_of_vec Vec512) in
     if_operation_supported op ~f:(fun () -> Cop (op, args, dbg))
-  | "caml_vec512_low_of_vec512" ->
-    let op = Creinterpret_cast (V512_of_vec Vec512) in
+  | "caml_vec512_low_of_vec256" ->
+    let op = Creinterpret_cast (V512_of_vec Vec256) in
     if_operation_supported op ~f:(fun () -> Cop (op, args, dbg))
-  | "caml_vec512_low_to_vec512" ->
-    let op = Creinterpret_cast (V512_of_vec Vec512) in
+  | "caml_vec512_low_to_vec256" ->
+    let op = Creinterpret_cast (V256_of_vec Vec512) in
     if_operation_supported op ~f:(fun () -> Cop (op, args, dbg))
   (* Scalar casts. These leave the top bits of the vector unspecified. *)
   | "caml_float64x2_low_of_float" ->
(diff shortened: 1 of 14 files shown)


backend/printcmm.ml
```diff
@@ -213,9 +218,9 @@ let to_string msg =
     ppf msg
 
 let reinterpret_cast : Cmm.reinterpret_cast -> string = function
-  | V128_of_v128 -> "vec128 as vec128"
-  | V256_of_v256 -> "vec256 as vec256"
-  | V512_of_v512 -> "vec512 as vec512"
+  | V128_of_vec w -> Printf.sprintf "vec128 as %s" (vector_width w)
+  | V256_of_vec w -> Printf.sprintf "vec256 as %s" (vector_width w)
+  | V512_of_vec w -> Printf.sprintf "vec512 as %s" (vector_width w)
```
xclerc:
```suggestion
  | V128_of_vec w -> Printf.sprintf "%s as vec128" (vector_width w)
  | V256_of_vec w -> Printf.sprintf "%s as vec256" (vector_width w)
  | V512_of_vec w -> Printf.sprintf "%s as vec512" (vector_width w)
```

backend/cmm_builtins.ml
```diff
@@ -315,13 +315,31 @@ let transl_vec_builtin name args dbg _typ_res =
...
+    let op = Creinterpret_cast (V512_of_vec Vec128) in
+    if_operation_supported op ~f:(fun () -> Cop (op, args, dbg))
+  | "caml_vec512_low_to_vec128" ->
+    let op = Creinterpret_cast (V128_of_vec Vec512) in
+    if_operation_supported op ~f:(fun () -> Cop (op, args, dbg))
+  | "caml_vec512_low_of_vec512" ->
+    let op = Creinterpret_cast (V512_of_vec Vec512) in
+    if_operation_supported op ~f:(fun () -> Cop (op, args, dbg))
+  | "caml_vec512_low_to_vec512" ->
+    let op = Creinterpret_cast (V512_of_vec Vec512) in
```
xclerc:
```suggestion
  | "caml_vec512_low_of_vec256" ->
    let op = Creinterpret_cast (V512_of_vec Vec256) in
    if_operation_supported op ~f:(fun () -> Cop (op, args, dbg))
  | "caml_vec512_low_to_vec256" ->
    let op = Creinterpret_cast (V256_of_vec Vec512) in
```

xclerc (commented):
I was initially going to write that there are
few tests, but more are assumed to be in
the library that would expose the builtins.

However, if my suggestion in `cmm_builtins.ml`
is correct, there is a bug that may have been
caught with more testing (I am unsure because
of the "unspecified" status of upper bits).

PR #4345, opened 2025-07-21
Stabilize vec256

TheNumbat:
Moves the vec256 types into the stable extension. 

We haven't finished TSAN support and OCaml stack slots are not yet aligned, but I don't think either of these are required for stable. TSAN isn't deployed and is broken for other reasons (and seems non-trivial to support, see `simd-tsan` branch), and stack alignment doesn't effect correctness so may be added later.

Using `vec256` types without enabling AVX will raises a fatal error rather than a structured error message. It might be better to only add the `vec256` types when AVX is enabled, but we need `ocamlc` to accept SIMD types (despite not supporting them) in order to release the `ocaml_simd` libraries.


TheNumbat:
I'll work on improving the error messages. It's also possible we should fully support the types without avx (+ in bytecode) by compiling to sse load/store, but that's not urgent.

TheNumbat:
The error messages now print the function name, which seems acceptable. For consistency, we only check for support during `emit`, so usage of 256-bit types that gets dead-code-eliminated won't raise. That's a bit weird, but we can eliminate the errors entirely if we choose to support the types without AVX.

mshinwell:
@xclerc could you please do that?  should be quick

backend/amd64/emit.ml
```diff
@@ -43,11 +43,49 @@ module D = Asm_targets.Asm_directives
...
+      | Extension_disabled msg -> Some ("extension_disabled: " ^ msg)
+      | _ -> None)
+
+  let require_vec256 () =
+    if not (Arch.Extension.enabled_vec256 ())
+    then
+      raise
+        (Extension_disabled
+           (Printf.sprintf
+              "found 256-bit register during %s, but AVX is not enabled."
```
xclerc: "... during <function_name>..."
feels a bit weird, what about
"... during <function_name> emission..."?

backend/amd64/emit.ml
```diff
@@ -43,11 +43,49 @@ module D = Asm_targets.Asm_directives
...
 module I = struct
   include I
 
-  let simd simd args =
-    Arch.Extension.require_instruction simd;
+  exception Extension_disabled of string
+
+  let () =
+    Printexc.register_printer (function
+      | Extension_disabled msg -> Some ("extension_disabled: " ^ msg)
```
xclerc:
```suggestion
      | Extension_disabled msg -> Some ("extension disabled: " ^ msg)
```

backend/amd64/emit.ml
TheNumbat: swapped to `while emitting`

backend/amd64/emit.ml
```diff
@@ -43,11 +43,49 @@ module D = Asm_targets.Asm_directives
...
+      | Extension_disabled msg -> Some ("extension disabled: " ^ msg)
+      | _ -> None)
+
+  let require_vec256 () =
+    if not (Arch.Extension.enabled_vec256 ())
+    then
+      raise
+        (Extension_disabled
+           (Printf.sprintf
+              "found 256-bit register while emitting %s, but AVX is not enabled."
```
xclerc: (for consistency, that should be "whilst"...
not that I could explain the difference...)

TheNumbat: ah, updated. I think it's just a british-ism

PR #4346, opened 2025-07-22
Add named identifiers + track undeclared symbols for Llvmize

usuyus:
This PR adds the following:

- Identifiers can now also be named. This was necessary since labels do not have to occur in order in `Cfg_with_layout.t` while LLVM expects that for unnamed identifiers.
- Referenced but not declared (via a `data_item`) symbols are declared as external symbols. This is different from extern function declarations, which will be added in a future PR.

### DIFF
--- a/backend/cfg/llvmize.ml
+++ b/backend/cfg/llvmize.ml
@@ -15,18 +15,21 @@
 
 module CL = Cfg_with_layout
 module DLL = Oxcaml_utils.Doubly_linked_list
+module String = Misc.Stdlib.String
 
 type error = Asm_generation of (string * int)
 
 exception Error of error
 
 module Ident : sig
-  (** Unnamed LLVM identifiers that start with "%".  This includes
-      basic blocks, unnamed function parameters, and temporaries
-      (virtual registers).  *)
+  (** LLVM identifiers that start with "%".  This includes
+      basic blocks, function parameters, and temporaries
+      (virtual registers). They can be unnamed or named.  *)
   type t
 
   val print : Format.formatter -> t -> unit
+
+  val named : string -> t
 
   module Gen : sig
     (** per-function counter for generating identifiers *)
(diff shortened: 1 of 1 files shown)


backend/cfg/llvmize.ml
```diff
@@ -383,10 +403,7 @@ module F = struct
     match d with
     | Cdefine_symbol _ -> assert false (* cannot happen *)
     | Cint n -> fprintf ppf "%s" (Nativeint.to_string n)
-    | Csymbol_address { sym_name; sym_global } -> (
-      match sym_global with
-      | Global -> pp_global ppf sym_name
-      | Local -> Misc.fatal_error "Llvmize.pp_const_data_item: not implemented")
+    | Csymbol_address { sym_name; sym_global = _ } -> pp_global ppf sym_name
```
gretay-js, line 406: Do we have examples where we get  `Local` case? If not, then we should keep the fatal_error case here.  Same below for other uses. Thanks for adding a CR below about it, I just want to understand when this arises.

backend/cfg/llvmize.ml
```diff
@@ -518,29 +541,64 @@ let data ds =
...
+        match d with
+        | Csymbol_address { sym_name; _ } -> Some sym_name
+        | Cdefine_symbol _ | Cint _ | Cint8 _ | Cint16 _ | Cint32 _ | Csingle _
+        | Cdouble _ | Cvec128 _ | Cvec256 _ | Cvec512 _ | Csymbol_offset _
+        | Cstring _ | Cskip _ | Calign _ ->
+          None)
+      t.data
+    |> String.Set.of_list
+    |> String.Set.union t.referenced_symbols
+  in
```
gretay-js: Can you try to merge these two matches with the one below that already traverses t.data? Looks like you can just add the symbols to the right sets at line L587 and L591.

gretay-js (changes requested):

backend/cfg/llvmize.ml
usuyus: I renamed `defined_function_symbols` back to `defined_symbols` since now it also keeps track of data.

backend/cfg/llvmize.ml
usuyus: This happens in the float test (with something named `Pmakeblock`, I'm assuming related to there being a module there), so we do need to ignore the `Local` case at those places as well.

PR #4350, opened 2025-07-22
Remove SA_NODEFER from the `sa_flags` of the SIGSEGV handler

stedolan:
SA_NODEFER allows signal handlers to recurse. This is generally a bad idea - if we segfault in our segfault handler, nothing good will come from attempting to handle it.

We have it because this code has been copied around [since 1999](https://github.com/oxcaml/oxcaml/commit/7afdd2ae849a093effe543307e17a0cda69826db) and it's unclear why it was added even at the time.


NickBarnes (approved):
Obvious improvement.

PR #4353, opened 2025-07-22
Make Multicore.spawn take a unique value, and give it back on exn

glittershark:
If spawning a function @ once as a thread fails, all the unique values that the thread closed over have semantically not been used by that @ once function - but currently, our API doesn't allow using them again in the exceptional case. This commit changes the Multicore API to expose this capability, by making the spawn functions take a ['a] uniquely, and give it back along with the exception if spawning the thread fails. A small amount of magic is used internally to make this work, plus an extension to the systhreads library to allow the argument passed to the thread closure to be [value_or_null & value_or_null].

### DIFF
--- a/otherlibs/systhreads/multicore.ml
+++ b/otherlibs/systhreads/multicore.ml
@@ -211,6 +211,9 @@ let () =
   done
 ;;
 
+external magic_unique__portable
+  : 'a @ portable -> 'a @ portable unique @@ portable
+  = "%identity"
 
 (** Run some function on a new thread. *)
 let thread (Request { action; argument; _ })=
@@ -223,14 +226,8 @@ let thread (Request { action; argument; _ })=
          We must wakeup the manager to potentially allow it to exit. *)
       wakeup_manager t
   in
-  let open struct
-    (* SAFETY: We know each value is only popped from the request stack
-       once *)
-    external magic_unique_argument
-      : 'a @ portable -> 'a @ portable unique
-      = "%identity"
-  end in
-  match action (magic_unique_argument argument) with
+  (* SAFETY: We know each value is only popped from the request stack once *)
+  match action (magic_unique__portable argument) with
   | () -> decr ()
   | exception exn ->
     let bt = Printexc.get_raw_backtrace () in
@@ -348,7 +345,7 @@ let spawn_on ~domain:i f a =
   Mutex.unlock mutex;
   (* SAFETY: We know that if we got an error here, the thread failed to spawn
      and hence no longer has a reference to [a] *)
-  Obj.magic_unique request_inner.result
+  magic_unique__portable request_inner.result
 ;;
 
 let spawn f =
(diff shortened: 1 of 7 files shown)


otherlibs/systhreads/multicore.ml
```diff
@@ -240,8 +248,12 @@ let rec manager_loop t =
       manager_loop t
     | _ ->
       let requests = Atomic.exchange t.incoming [] in
-      List.iter (fun (request : request) ->
-        (match Thread.Portable.create thread request.action with
+      List.iter (fun (Request request : request) ->
+        (match
+           Thread.Portable.create2
+             thread
+             #(request.action, request.argument)
```
polytypic: Hmm... Perhaps just pass the whole `request` to the `thread` function?
```ocaml
Thread.Portable.create thread request
```
No need to extend the `Thread` interface and should be efficient.  (Allocating a pair at this point should be an insignificant cost as well.)

otherlibs/systhreads/multicore.ml
```diff
@@ -317,14 +338,18 @@ let spawn_on ~domain:i f =
...
-  | None -> ()
-  | Some { portable = exn, bt } -> Printexc.raise_with_backtrace exn bt
+  Mutex.unlock mutex;
+  match request with
+  | Request { error = None; _ } -> Spawned
+  | Request { error = Some { portable = exn, bt }; _ } ->
+    Failed (
+      (* SAFETY: We know that if we got an error here, the thread failed to spawn and
+         hence no longer has a reference to [a] *)
+      Obj.magic_unique a, exn, bt)
```
polytypic: If you want to reduce allocations, then, instead of an `error` field, the request could have a `result` field.  It would be initialized with `Spawned` and updated to `Failed ...` in case of an error before setting the `ready` flag.

otherlibs/systhreads/multicore.ml
glittershark: Yep, that's obviously nicer. thanks!

otherlibs/systhreads/multicore.ml
glittershark: nice! Done.

otherlibs/systhreads/multicore.mli
```diff
@@ -54,4 +62,7 @@ val spawn_on : domain:int -> (unit -> unit) @ once portable unyielding -> unit
     so as to balance the load per domain.
 
     @raise Sys_error in case the system fails to create a new thread. *)
-val spawn : (unit -> unit) @ once portable unyielding -> unit
+val spawn
+  :  ('a @ unique portable -> unit) @ once portable unyielding
+  -> 'a @ unique portable
+  -> 'a spawn_result @ unique
```
polytypic: Should the return value be `portable`?  (At least the `'a` input is.  The same also applies to `spawn_on` above.)

glittershark: ah, indeed it should be, thank you!

polytypic:
Note that there are two copies of the `Multicore` module, under `systhreads` and `systhreads4`.

glittershark:
> Note that there are two copies of the `Multicore` module, under `systhreads` and `systhreads4`.

oh, gross. Fixed.

polytypic (approved):

PR #4354, opened 2025-07-22
Closure translation from flambda2 -> jsoo IR

smjleo:
Should be reviewed by commit (sorry they're not very granular)
1. Bulk of the work, implementing the actual translation of code blocks and `Set_of_closures`, handling [Project_*_slot] in the environment, etc.
2. (plan) Do two passes over batched lets: code first then everything else
3. (plan) Function applications

### DIFF
--- a/middle_end/flambda2/to_jsir/to_jsir_env.ml
+++ b/middle_end/flambda2/to_jsir/to_jsir_env.ml
@@ -22,6 +22,9 @@ let create ~module_symbol ~return_continuation ~exn_continuation =
     value_slots = Value_slot.Map.empty
   }
 
+let enter_function_body t ~return_continuation ~exn_continuation =
+  { t with return_continuation; exn_continuation }
+
 let module_symbol t = t.module_symbol
 
 let add_continuation t cont addr =
@@ -59,9 +62,10 @@ let get_symbol_exn t symbol = Symbol.Map.find symbol t.symbols
 
 let get_code_id_exn t code_id = Code_id.Map.find code_id t.code_ids
 
-let get_function_slot t fslot = Function_slot.Map.find fslot t.function_slots
+let get_function_slot_exn t fslot =
+  Function_slot.Map.find fslot t.function_slots
 
-let get_value_slot t vslot = Value_slot.Map.find vslot t.value_slots
+let get_value_slot_exn t vslot = Value_slot.Map.find vslot t.value_slots
 
 let add_alias_of_var_exn t ~var ~alias_of =
   let jvar = get_var_exn t alias_of in
--- a/middle_end/flambda2/to_jsir/to_jsir_env.ml
+++ b/middle_end/flambda2/to_jsir/to_jsir_env.ml
@@ -4,7 +4,10 @@ type t =
     exn_continuation : Continuation.t;
     continuations : Jsir.Addr.t Continuation.Map.t;
     vars : Jsir.Var.t Variable.Map.t;
-    symbols : Jsir.Var.t Symbol.Map.t
+    symbols : Jsir.Var.t Symbol.Map.t;
+    code_ids : (Jsir.Addr.t * Jsir.Var.t list) Code_id.Map.t;
+    function_slots : Jsir.Var.t Function_slot.Map.t;
+    value_slots : Jsir.Var.t Value_slot.Map.t
   }
 
 let create ~module_symbol ~return_continuation ~exn_continuation =
@@ -13,7 +16,10 @@ let create ~module_symbol ~return_continuation ~exn_continuation =
     exn_continuation;
     continuations = Continuation.Map.empty;
     vars = Variable.Map.empty;
-    symbols = Symbol.Map.empty
+    symbols = Symbol.Map.empty;
+    code_ids = Code_id.Map.empty;
+    function_slots = Function_slot.Map.empty;
+    value_slots = Value_slot.Map.empty
   }
 
 let module_symbol t = t.module_symbol
@@ -25,6 +31,15 @@ let add_var t fvar jvar = { t with vars = Variable.Map.add fvar jvar t.vars }
 
 let add_symbol t symbol jvar =
   { t with symbols = Symbol.Map.add symbol jvar t.symbols }
+
+let add_code_id t code_id ~addr ~params =
+  { t with code_ids = Code_id.Map.add code_id (addr, params) t.code_ids }
+
+let add_function_slot t fslot jvar =
+  { t with function_slots = Function_slot.Map.add fslot jvar t.function_slots }
+
+let add_value_slot t vslot jvar =
+  { t with value_slots = Value_slot.Map.add vslot jvar t.value_slots }
 
 type continuation =
   | Return
@@ -41,6 +56,12 @@ let get_continuation_exn t cont =
 let get_var_exn t fvar = Variable.Map.find fvar t.vars
 
 let get_symbol_exn t symbol = Symbol.Map.find symbol t.symbols
+
+let get_code_id_exn t code_id = Code_id.Map.find code_id t.code_ids
+
+let get_function_slot t fslot = Function_slot.Map.find fslot t.function_slots
+
+let get_value_slot t vslot = Value_slot.Map.find vslot t.value_slots
 
 let add_alias_of_var_exn t ~var ~alias_of =
   let jvar = get_var_exn t alias_of in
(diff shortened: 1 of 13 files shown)


middle_end/flambda2/to_jsir/to_jsir_set_of_closures.ml
```diff
@@ -0,0 +1,63 @@
...
+        with
+        | Deleted _ -> env, res
+        | Code_id { code_id; only_full_applications = _ } ->
+          (* CR selee: thread through debug information *)
+          let addr, params = To_jsir_env.get_code_id_exn env code_id in
+          let expr : Jsir.expr = Closure (params, (addr, []), None) in
+          (* If this function slot is used, it should've already been added when
+             the code using it was defined *)
+          let env, fn_var =
+            match To_jsir_env.get_function_slot_exn env slot with
```
smjleo: This is where (among others) we make sure that the correct variable (bound to the function slot encountered earlier) is used for binding closures

middle_end/flambda2/to_jsir/to_jsir_static_const.ml
```diff
@@ -88,3 +88,85 @@ let block_like ~env ~res (const : Static_const.t) :
...
+  let free_names = Code0.free_names code in
+  let function_slots = Name_occurrences.all_function_slots free_names in
+  let value_slots = Name_occurrences.all_value_slots free_names in
+  let symbols = Name_occurrences.symbols free_names in
+  (* We create new variables that represent each function slot, value slot and
+     symbol (for static mutually-recursive code blocks) if they don't exist
+     already, and use them everywhere that the corresponding slot is used. We
+     will make sure later (when translating [Set_of_closures]) that the closures
+     or values representing these slots are bound to the correct variables. *)
+  let env =
```
smjleo, line 106: and this is where unseen function slots/value slots/symbols are assigned fresh variables

middle_end/flambda2/to_jsir/to_jsir_env.ml
```diff
@@ -42,10 +64,33 @@ let get_var_exn t fvar = Variable.Map.find fvar t.vars
...
 let add_alias_of_var_exn t ~var ~alias_of =
   let jvar = get_var_exn t alias_of in
   { t with vars = Variable.Map.add var jvar t.vars }
 
 let add_alias_of_symbol_exn t ~var ~alias_of =
   let jvar = get_symbol_exn t alias_of in
   { t with vars = Variable.Map.add var jvar t.vars }
+
+let add_if_not_found t item ~get ~add =
+  match get t item with
```
jvanburen: Map.mem?

middle_end/flambda2/to_jsir/to_jsir_set_of_closures.ml
```diff
@@ -0,0 +1,63 @@
+let set_of_closures ~env ~res ~bindings ~add_to_env soc =
+  let fun_decls = Set_of_closures.function_decls soc in
+  let decls =
+    Function_declarations.funs_in_order fun_decls |> Function_slot.Lmap.to_seq
+  in
+  let env, res =
+    Seq.fold_left2
```
jvanburen: i've never seen anyone actually use seq over list, is there a reason?

middle_end/flambda2/to_jsir/to_jsir_set_of_closures.ml
jvanburen: consider returning an option?

middle_end/flambda2/to_jsir/to_jsir_set_of_closures.ml
```diff
@@ -0,0 +1,63 @@
...
+              let var = Jsir.Var.fresh () in
+              To_jsir_env.add_function_slot env slot var, var
+          in
+          let res = To_jsir_result.add_instr_exn res (Let (fn_var, expr)) in
+          add_to_env ~env ~res binding fn_var)
+      (env, res) bindings decls
+  in
+  Value_slot.Map.fold
+    (fun slot simple (env, res) ->
+      match To_jsir_env.get_value_slot_exn env slot with
```
jvanburen: return an option since you expect this to sometimes fail?

middle_end/flambda2/to_jsir/to_jsir_set_of_closures.ml
```diff
@@ -0,0 +1,63 @@
...
+    ~add_to_env:(fun ~env ~res var fn_var ->
+      To_jsir_env.add_var env var fn_var, res)
+    soc
+
+let static_set_of_closures ~env ~res ~closure_symbols soc =
+  let add_to_env ~env ~res symbol fn_var =
+    (* We may have already encountered the symbol when translating the code, so
+       we should check first whether a variable already exists; if so, we should
+       make sure that the variable also points to the function variable *)
+    match To_jsir_env.get_symbol_exn env symbol with
```
jvanburen: same here, option-ify

middle_end/flambda2/to_jsir/to_jsir_primitive.ml
```diff
@@ -85,12 +83,10 @@ let unary ~env ~res (f : Flambda_primitive.unary_primitive) x : result =
     primitive_not_supported ()
   | Untag_immediate -> primitive_not_supported ()
   | Tag_immediate -> primitive_not_supported ()
-  | Project_function_slot { move_from; move_to } ->
-    ignore (move_from, move_to);
-    primitive_not_supported ()
-  | Project_value_slot { project_from; value_slot } ->
-    ignore (project_from, value_slot);
-    primitive_not_supported ()
+  | Project_function_slot { move_from = _; move_to } ->
```
jvanburen, line 86: should these check move_from/project_from?

jvanburen (changes requested):

middle_end/flambda2/to_jsir/to_jsir_set_of_closures.ml
smjleo: I used `Seq` because `Lmap` only has a `to_seq` and no `to_list`

middle_end/flambda2/to_jsir/to_jsir_set_of_closures.ml
smjleo: Here and the below couple of comments I might add non-exn variants of these so that the normal exn variants can be used elsewhere where we don't expect to fail, is that alright?

middle_end/flambda2/to_jsir/to_jsir_primitive.ml
smjleo: Seems to be for optimisations rather than codegen. Though I found out that apparently value_slot is used to call external functions too? I'll worry about that later and add a CR here

middle_end/flambda2/to_jsir/to_jsir_static_const.ml
```diff
@@ -88,3 +88,75 @@ let block_like ~env ~res (const : Static_const.t) :
...
+         ~is_my_closure_used:_
+         ~my_region:_
+         ~my_ghost_region:_
+         ~my_depth:_
+         ~free_names_of_body:_
+       ->
+      (* CR selee: A hack to get things to work, should figure out what
+         [my_closure] is actually used for *)
+      let var = Jsir.Var.fresh () in
+      let env0 = To_jsir_env.add_var env my_closure var in
```
smjleo, line 138: Now we don't keep return the bound parameters (same issue as in #4378)

middle_end/flambda2/to_jsir/to_jsir_static_const.ml
```diff
@@ -88,3 +88,75 @@ let block_like ~env ~res (const : Static_const.t) :
...
+         ~my_region:_
+         ~my_ghost_region:_
+         ~my_depth:_
+         ~free_names_of_body:_
+       ->
+      (* CR selee: A hack to get things to work, should figure out what
+         [my_closure] is actually used for *)
+      let var = Jsir.Var.fresh () in
+      let env = To_jsir_env.add_var env my_closure var in
+      let env_with_params, fn_params =
```
smjleo, line 139: Now we don't keep return the bound parameters (same issue as in #4378 )

middle_end/flambda2/to_jsir/to_jsir_set_of_closures.ml
lthls: You might want to use `Lmap.bindings`. `Lmap` uses lists under the hood, so this function is actually the identity.

smjleo: Ah thank you! Will do

jvanburen (approved):

PR #4355, opened 2025-07-23
Reduce nearby address collisions in [Hash_retaddr]

dvulakh:
Currently the `Hash_retaddr` that populates the frame description hashtable is just `x >> 3`. Consequently, registering a large number of symbols with nearby addresses causes many collisions and substantially degrades the performance of programs that rely on exceptions for control flow / error handling.

This PR changes the hash function to `x * 715827883`, which is good enough to fix the regression we encountered internally and still much cheaper than e.g. a proper FNV hash.

### DIFF
--- a/runtime/caml/frame_descriptors.h
+++ b/runtime/caml/frame_descriptors.h
@@ -151,7 +151,7 @@ Caml_inline bool frame_has_debug(frame_descr *d) {
   (void*)(((uintnat)(p) + sizeof(ty) - 1) & -sizeof(ty))
 
 #define Hash_retaddr(addr, mask)                          \
-  (((uintnat)(addr) * 715827883) & (mask))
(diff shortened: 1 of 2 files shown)


stedolan:
Looks good, but can I propose the following tweak: try something like `(x * N) >> 5`, where `N` is some large random odd number like 715827883. (This is the [multiply-shift universal hashing scheme](https://en.wikipedia.org/wiki/Universal_hashing#Avoiding_modular_arithmetic))

The shift is important because for constant `N`, the low k bits of `x * N` depend only on the low k bits of `x`. On e.g. arm64 instructions are 4-byte aligned, so the low 2 bits of `x` are always `0`, and only 1/4 of the hash values are possible if the hash function is `x * N`. Even on x86, there are enough alignment directives in various places that the distribution of the low 3-4 bits is unlikely to be uniform. (I think this is why the old "hash" had a shift)

PR #4356, opened 2025-07-23
Add Syntax and Typing Support for External Allocations

alpha-convert:
Adds syntax and typechecking support for malloc_.

* Currently, (boxed) records, (boxed) tuples, and constructors/variants that allocate can be malloc_'d.
* Typing-wise, all components of a malloc_'d block must be external_. malloc_ itself returns a 'a mallocd, a new builtin of kind bits64.
* This syntax is preliminary and subject to change.
* In this PR, malloc_ is semantically a no-op. No off-heap allocation occurs, the block is heap-allocated as usual.

### DIFF
--- a/parsing/pprintast.ml
+++ b/parsing/pprintast.ml
@@ -27,7 +27,7 @@ open Location
 open Longident
 open Parsetree
 
-let prefix_symbols  = [ '!'; '?'; '~']
+let prefix_symbols  = [ '!'; '?'; '~' ]
 let infix_symbols = [ '='; '<'; '>'; '@'; '^'; '|'; '&'; '+'; '-'; '*'; '/';
                       '$'; '%'; '#' ]
 
--- a/parsing/pprintast.ml
+++ b/parsing/pprintast.ml
@@ -27,7 +27,7 @@ open Location
 open Longident
 open Parsetree
 
-let prefix_symbols  = [ '!'; '?'; '~' ]
+let prefix_symbols  = [ '!'; '?'; '~']
 let infix_symbols = [ '='; '<'; '>'; '@'; '^'; '|'; '&'; '+'; '-'; '*'; '/';
                       '$'; '%'; '#' ]
 
@@ -1051,6 +1051,8 @@ and expression ctxt f x =
     | Pexp_stack e ->
         (* Similar to the common case of [Pexp_apply] *)
         pp f "@[<hov2>stack_@ %a@]" (expression2 reset_ctxt)  e
+    | Pexp_malloc e ->
+        pp f "@[<hov2>malloc_@ %a@]" (expression2 reset_ctxt)  e
     | Pexp_construct (li, Some eo)
       when not (is_simple_construct (view_expr x))-> (* Not efficient FIXME*)
         (match view_expr x with
--- a/parsing/printast.ml
+++ b/parsing/printast.ml
@@ -451,6 +451,9 @@ and expression i ppf x =
   | Pexp_stack e ->
       line i ppf "Pexp_stack\n";
       expression i ppf e
+  | Pexp_malloc e ->
+      line i ppf "Pexp_malloc\n";
+      expression i ppf e
   | Pexp_comprehension c ->
       line i ppf "Pexp_comprehension\n";
       comprehension_expression i ppf c
--- a/testsuite/tests/typing-modes/malloc.ml
+++ b/testsuite/tests/typing-modes/malloc.ml
@@ -260,8 +260,8 @@ type t = int or_null mallocd
 type t = int or_null mallocd
 |}]
 
-(* Basic type behavior: does malloc_ returrn a mallocd, and does it
-   requires that its arguments are passed @ external *)
+(* Basic type behavior: does malloc_ return a mallocd, and does it
+   require that its arguments are passed @ external *)
 let foo () = malloc_ (1,2)
 [%%expect {|
 val foo : unit -> (int * int) mallocd = <fun>
--- a/testsuite/tests/typing-modes/malloc.ml
+++ b/testsuite/tests/typing-modes/malloc.ml
@@ -250,11 +250,6 @@ Error: This type "float#" should be an instance of type "('a : value_or_null)"
          because the type argument of mallocd has layout value_or_null.
 |}]
 
-type t = #(int * int)
-[%%expect {|
-type t = #(int * int)
-|}]
-
 type t = int or_null mallocd
 [%%expect {|
 type t = int or_null mallocd
--- a/testsuite/tests/typing-modes/malloc.ml
+++ b/testsuite/tests/typing-modes/malloc.ml
@@ -41,6 +41,35 @@ let foo () = malloc_ (1,2)
 val foo : unit -> (int * int) mallocd = <fun>
 |}]
 
+let f (x @ external_) (y @ external_) @ unique =
+  malloc_ (x,y)
+[%%expect {|
+val f : 'a @ unique -> 'b @ unique -> ('a * 'b) mallocd @ unique = <fun>
+|}]
+
+let f (x @ external_ aliased) (y @ external_) @ unique =
+  malloc_ (x,y)
+[%%expect {|
+Line 2, characters 11-12:
+2 |   malloc_ (x,y)
+               ^
+Error: This value is "aliased" but expected to be "unique".
+|}]
+
+let f (x @ external_ unique) @ unique =
+  malloc_ (x,x)
+[%%expect {|
+Line 2, characters 13-14:
+2 |   malloc_ (x,x)
+                 ^
+Error: This value is used here, but it is already being used as unique:
+Line 2, characters 11-12:
+2 |   malloc_ (x,x)
+               ^
+
+|}]
+
+
 let foo () = malloc_ (1,"Asdf")
 [%%expect {|
 val foo : unit -> (int * string) mallocd = <fun>
@@ -370,4 +399,33 @@ let glorb2 () = malloc_ (Foo (1,2))
 [%%expect {|
 type 'a t = Foo of 'a * 'a
 val glorb2 : unit -> int t mallocd = <fun>
+|}]
+
+type t = Pack : 'k. 'k -> t
+let f (x @ external_) = malloc_ (Pack x)
+[%%expect{|
+type t = Pack : 'k -> t
+val f : 'a -> t mallocd = <fun>
+|}]
+
+let f (x @ internal) = malloc_ (Pack x)
+[%%expect{|
+Line 1, characters 37-38:
+1 | let f (x @ internal) = malloc_ (Pack x)
+                                         ^
+Error: This value is "internal" but expected to be "external_".
+|}]
+
+type 'a t =
+| Foo : int * 'a -> int t
+| Bar : { x : char ; y : 'a } -> char t
+let f (x @ external_) = malloc_ (Foo (3,x))
+[%%expect{|
+type 'a t = Foo : int * 'a -> int t | Bar : { x : char; y : 'a; } -> char t
+val f : 'a -> int t mallocd = <fun>
+|}]
+
+let f (y @ external_) = malloc_ (Bar {x = 'c'; y })
+[%%expect{|
+val f : 'a -> char t mallocd = <fun>
 |}]
--- a/testsuite/tests/typing-modes/malloc.ml
+++ b/testsuite/tests/typing-modes/malloc.ml
@@ -1,0 +1,370 @@
+(* TEST
+ expect;
+*)
+
+type t : bits64 = int mallocd
+[%%expect {|
+type t = int mallocd
+|}]
+
+type t : bits64 mod external_ = int mallocd
+[%%expect {|
+Line 1, characters 0-43:
+1 | type t : bits64 mod external_ = int mallocd
+    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Error: The kind of type "int mallocd" is bits64
+         because it is the primitive type mallocd.
+       But the kind of type "int mallocd" must be a subkind of
+           bits64 mod external_
+         because of the definition of t at line 1, characters 0-43.
+|}]
+
+type t = float# mallocd
+[%%expect {|
+Line 1, characters 9-15:
+1 | type t = float# mallocd
+             ^^^^^^
+Error: This type "float#" should be an instance of type "('a : value_or_null)"
+       The layout of float# is float64
+         because it is the unboxed version of the primitive type float.
+       But the layout of float# must be a sublayout of value
+         because the type argument of mallocd has layout value_or_null.
+|}]
+
+type t = int or_null mallocd
+[%%expect {|
+type t = int or_null mallocd
+|}]
+
+let foo () = malloc_ (1,2)
+[%%expect {|
+val foo : unit -> (int * int) mallocd = <fun>
+|}]
+
+let foo () = malloc_ (1,"Asdf")
+[%%expect {|
+val foo : unit -> (int * string) mallocd = <fun>
+|}]
+
+let foo (x @ external_) = malloc_ (x,1)
+[%%expect {|
+val foo : 'a -> ('a * int) mallocd = <fun>
+|}]
+
+let asdf (x @ internal) = malloc_ (x,2)
+[%%expect {|
+Line 1, characters 35-36:
+1 | let asdf (x @ internal) = malloc_ (x,2)
+                                       ^
+Error: This value is "internal" but expected to be "external_".
+|}]
+
+type t = {x : int; y : int}
+let f () = malloc_ {x = 3; y = 4}
+[%%expect{|
+type t = { x : int; y : int; }
+val f : unit -> t mallocd = <fun>
+|}]
+
+let f (x @ external_) = malloc_ {x = x; y = x}
+[%%expect{|
+val f : int -> t mallocd = <fun>
+|}]
+
+type 'a t = {x : 'a ; y : 'a option mallocd }
+let f x =
+  let y = malloc_ (Some x) in
+  malloc_ {x;y}
+[%%expect {|
+type 'a t = { x : 'a; y : 'a option mallocd; }
+val f : 'a -> 'a t mallocd = <fun>
+|}]
+
+
+
+
+let foo () = malloc_ []
+[%%expect {|
+Line 1, characters 21-23:
+1 | let foo () = malloc_ []
+                         ^^
+Error: This expression is not an allocation site.
+|}]
+
+let foo () = malloc_ [1]
+[%%expect {|
+val foo : unit -> int list mallocd = <fun>
+|}]
+
+let foo () = malloc_ [1;2;3]
+[%%expect {|
+Line 1, characters 24-28:
+1 | let foo () = malloc_ [1;2;3]
+                            ^^^^
+Error: This value is "internal" but expected to be "external_".
+|}]
+
+let foo (x @ external_) (xs @ external_) =
+  malloc_ (x :: xs)
+[%%expect {|
+val foo : 'a -> 'a list -> 'a list mallocd = <fun>
+|}]
+
+let foo (x @ internal) (xs @ internal) =
+  malloc_ (x :: xs)
+[%%expect {|
+Line 2, characters 11-12:
+2 |   malloc_ (x :: xs)
+               ^
+Error: This value is "internal" but expected to be "external_".
+|}]
+
+let foo (x @ external_) (xs @ internal) =
+  malloc_ (x :: xs)
+[%%expect {|
+Line 2, characters 16-18:
+2 |   malloc_ (x :: xs)
+                    ^^
+Error: This value is "internal" but expected to be "external_".
+|}]
+
+
+type t1 = Foo of int | Bar
+type t2 = Baz of t1
+
+let f () = malloc_ (Baz (Foo 3))
+[%%expect {|
+type t1 = Foo of int | Bar
+type t2 = Baz of t1
+Line 4, characters 24-31:
+4 | let f () = malloc_ (Baz (Foo 3))
+                            ^^^^^^^
+Error: This value is "internal" but expected to be "external_".
+|}]
+
+
+
+let foo1 () = malloc_ [|1;2;3|]
+[%%expect {|
+Line 1, characters 22-31:
+1 | let foo1 () = malloc_ [|1;2;3|]
+                          ^^^^^^^^^
+Error: Externally allocating arrays is not supported yet.
+|}]
+
+
+
+(* CR external-mode: The following error correctly catches an error: to malloc a
+   record with mutable fields, all mutable fields must be marked @@ external_.
+   However, it doesn't really explain that this is *why*, and worse, it's
+   actively misleading. High priority to fix this. *)
+type t = {mutable x : string; y : bool}
+let bar1 x y = malloc_ {x;y}
+[%%expect {|
+type t = { mutable x : string; y : bool; }
+Line 2, characters 23-28:
+2 | let bar1 x y = malloc_ {x;y}
+                           ^^^^^
+Error: This value is "internal" but expected to be "external_".
+|}]
+
+type t = {mutable x : string @@ external_ ; y : bool}
+let bar2 x y = malloc_ {x;y}
+[%%expect {|
+type t = { mutable x : string @@ external_; y : bool; }
+val bar2 : string -> bool -> t mallocd = <fun>
+|}]
+
+
+
+type 'a t = Foo | Bar of 'a | Baz of 'a * string
+let baz1 () = malloc_ Foo
+[%%expect {|
+type 'a t = Foo | Bar of 'a | Baz of 'a * string
+Line 2, characters 22-25:
+2 | let baz1 () = malloc_ Foo
+                          ^^^
+Error: This expression is not an allocation site.
+|}]
+
+let baz2 () = malloc_ (Bar 1)
+[%%expect {|
+val baz2 : unit -> int t mallocd = <fun>
+|}]
+
+let baz3 () = malloc_ (Bar "asdf")
+[%%expect {|
+val baz3 : unit -> string t mallocd = <fun>
+|}]
+
+let baz4 (x @ external_) = malloc_ (Baz (1,x))
+[%%expect {|
+val baz4 : string -> int t mallocd = <fun>
+|}]
+
+type 'a t = FooBar of {mutable x : 'a @@ external_ ; mutable y : 'a @@ external_}
+let foobar (x @ external_) (y @ external_ ) = malloc_ (FooBar {x ; y})
+[%%expect {|
+type 'a t =
+    FooBar of { mutable x : 'a @@ external_; mutable y : 'a @@ external_; }
+val foobar : 'a -> 'a -> 'a t mallocd = <fun>
+|}]
+
+type 'a r = {mutable x : 'a @@ external_ ; mutable y : 'a @@ external_}
+type 'a t = FooBar of 'a
+let flop (x @ external_) (y @ external_ ) = malloc_ {x ; y}
+[%%expect {|
+type 'a r = { mutable x : 'a @@ external_; mutable y : 'a @@ external_; }
+type 'a t = FooBar of 'a
+val flop : 'a -> 'a -> 'a r mallocd = <fun>
+|}]
+let fleep (x @ external_) (y @ external_ ) = malloc_ (FooBar {x ; y})
+[%%expect {|
+Line 1, characters 61-68:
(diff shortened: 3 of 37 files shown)


parsing/printast.ml
```diff
@@ -451,6 +451,9 @@ and expression i ppf x =
   | Pexp_stack e ->
       line i ppf "Pexp_stack\n";
       expression i ppf e
+  | Pexp_malloc e ->
+      line i ppf "Pexp_malloc\n";
+      expression i ppf e
```
ccasin, line 456: (Not a particularly relevant place to leave this comment)

I think the other printer `pprintast.ml`, also needs to be updated for `Pexp_malloc`.  I guess it has only sad pattern matches with catch-all cases so it's easy to miss. You'll notice that a bot left a comment asking you to add a test in `parsetree/source_jane_street.ml`, and you'll probably need to update `Pprintast` to make that work.

testsuite/tests/typing-modes/malloc.ml
```diff
@@ -0,0 +1,431 @@
+(* TEST
+ expect;
+*)
```
ccasin, line 3: As a general rule, I think it's useful to put comments in test files explaining what you're testing.  You don't have to put a comment on every individual test, but see other modes tests in the testsuite (e.g., [here](https://github.com/oxcaml/oxcaml/blob/main/testsuite/tests/typing-modes/portable-contend.ml) or [here](https://github.com/oxcaml/oxcaml/blob/main/testsuite/tests/typing-modes/val_modalities.ml)) for a decent balance of providing some sign-posting and making the purpose of the tests clear, or look at the layouts test for tests that go farther in this direction (e.g., [here](https://github.com/oxcaml/oxcaml/blob/main/testsuite/tests/typing-layouts/modules.ml) or [here](https://github.com/oxcaml/oxcaml/blob/main/testsuite/tests/typing-layouts-products/basics.ml)).  This makes it much easier to try to find tests that check specific things, to assess how complete the testsuite is, and to understand what a test is doing when working on a compiler PR that has changed its output.

testsuite/tests/typing-modes/malloc.ml
```diff
@@ -0,0 +1,431 @@
...
+[%%expect {|
+Line 1, characters 0-43:
+1 | type t : bits64 mod external_ = int mallocd
+    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Error: The kind of type "int mallocd" is bits64
+         because it is the primitive type mallocd.
+       But the kind of type "int mallocd" must be a subkind of
+           bits64 mod external_
+         because of the definition of t at line 1, characters 0-43.
+|}]
```
ccasin, line 186: I think it's safe to let `'a malloced` mode cross on this axis (at least, to `external_`, but not to `byte_external`).

ccasin, line 186: Also: I think we should have a thorough test of the mode crossing behavior for all axes (showing it doesn't cross on the axes it doesn't and does on the axes it does).

typing/predef.ml
```diff
@@ -532,6 +535,16 @@ let build_initial_env add_type add_extension empty_env =
...
+  |> add_type1 ident_mallocd
+       ~variance:Variance.covariant
+       ~separability:Separability.Sep
+       ~param_jkind:(Jkind.Builtin.value_or_null
+                      ~why:(Type_argument
+                             {parent_path = path_mallocd;
+                              position = 1; arity = 1}))
+       ~jkind:(fun _param ->
+         Jkind.of_builtin Jkind.Const.Builtin.bits64
+          ~why:(Primitive ident_mallocd))
```
ccasin: I think `'a malloced` can mode cross more than this. I mentioned `external` in the tests, but I also think   it can cross many other axes _with_ its argument (every modal axis except unqueness?). We can discuss.

testsuite/tests/typing-modes/malloc.ml
```diff
@@ -0,0 +1,431 @@
...
+
+let foo () = malloc_ (1,2)
+[%%expect {|
+val foo : unit -> (int * int) mallocd = <fun>
+|}]
+
+let f (x @ external_) (y @ external_) @ unique =
+  malloc_ (x,y)
+[%%expect {|
+val f : 'a @ unique -> 'b @ unique -> ('a * 'b) mallocd @ unique = <fun>
```
ccasin: It seems like the printing isn't quite right here - `internal` is the legacy mode, so we should be printing `external_` where that is the relevant mode, as we do for unique here.

testsuite/tests/typing-modes/malloc.ml
```diff
@@ -0,0 +1,431 @@
...
+|}]
+
+let f (x @ external_ aliased) (y @ external_) @ unique =
+  malloc_ (x,y)
+[%%expect {|
+Line 2, characters 11-12:
+2 |   malloc_ (x,y)
+               ^
+Error: This value is "aliased" but expected to be "unique".
```
ccasin: It wasn't obvious to me, from looking at the test of uniqueness here, whether (a) `malloc_` requires its argument uniquely, or (b) `malloc_` is mode polymorphic - returning the `'a malloced` with the same uniqueness as the argument.  I think we should have clear tests for this.

Based on this test further below, I believe the answer is the latter:
```ocaml
let foo (x @ external_) (xs @ external_) =
  malloc_ (x :: xs)
[%%expect {|
val foo : 'a -> 'a list -> 'a list mallocd = <fun>
```

I think this is a fine design - it's probably rare to want to malloc something you can never free, but certainly not unreasonable (e.g., for data that lives for the whole program).  But if this is the case, we'll need also to provide a `use` function that doesn't require the `'a malloced` uniquely.  I guess actually it's safe to just have `val deref : 'a malloced -> 'a` for such cases, since you're guaranteed it can never be freed?  We will also have to be careful that we don't rely on the idea that an `'a malloced` can only be used uniquely when reasoning about how much it can mode cross.  (It's possible it actually makes for a better user experience to have a separate `malloc_` function and `malloced` type for this use case, unsure).

Alternatively, it's also fine to require uniqueness for a first go.

testsuite/tests/typing-modes/malloc.ml
```diff
@@ -0,0 +1,431 @@
...
+let f () = malloc_ {x = 3; y = 4}
+[%%expect{|
+type t = { x : int; y : int; }
+val f : unit -> t mallocd = <fun>
+|}]
+
+let f (x @ external_) = malloc_ {x = x; y = x}
+[%%expect{|
+val f : int -> t mallocd = <fun>
+|}]
```
ccasin: A test showing this also works if you take `x` internally because ints mode cross would be good.

testsuite/tests/typing-modes/malloc.ml
```diff
@@ -0,0 +1,431 @@
...
+
+
+
+let foo () = malloc_ []
+[%%expect {|
+Line 1, characters 21-23:
+1 | let foo () = malloc_ []
+                         ^^
+Error: This expression is not an allocation site.
+|}]
```
ccasin, line 317: I think we should have a test of every `Pexp_*` form showing which `malloc_` works with and which it does not.

(It looks like this file tests many of them throughout, but I'm unsure if it's all, and it would be good to have a test of all of them in one place)

testsuite/tests/typing-modes/malloc.ml
```diff
@@ -0,0 +1,431 @@
...
+
+
+
+let foo1 () = malloc_ [|1;2;3|]
+[%%expect {|
+Line 1, characters 22-31:
+1 | let foo1 () = malloc_ [|1;2;3|]
+                          ^^^^^^^^^
+Error: Externally allocating arrays is not supported yet.
+|}]
```
ccasin, line 370: I think it would be good to have a test of something like:
```ocaml
type 'a malloced_list = Nil | Cons of 'a * 'a malloced malloced_list
```
For which I think it is possible to do things like:
```ocaml
let foo () = malloc (Cons 0 (malloc_ Cons 1 (malloc_ Cons 2 Nil)))
```

testsuite/tests/typing-modes/malloc.ml
```diff
@@ -0,0 +1,431 @@
...
+2 | let bar1 x y = malloc_ {x;y}
+                           ^^^^^
+Error: This value is "internal" but expected to be "external_".
+|}]
+
+type t = {mutable x : string @@ external_ ; y : bool}
+let bar2 x y = malloc_ {x;y}
+[%%expect {|
+type t = { mutable x : string @@ external_; y : bool; }
+val bar2 : string -> bool -> t mallocd = <fun>
```
ccasin: I expected this to be:
```ocaml
val bar2 : string @ external -> bool -> t mallocd @ external = <fun>
```
But I guess it's the same printing bug as elsewhere?

typing/typecore.ml
```diff
@@ -7078,6 +7127,60 @@ and type_expect_
...
+       | Pexp_new _ -> unsupported Object
+       | Pexp_lazy _ -> unsupported Lazy
+       | Pexp_object _ -> unsupported Object
+       | Pexp_pack _ -> unsupported Module
+       | Pexp_array _ -> unsupported Array
+       | Pexp_record _ -> ()
+       | Pexp_tuple _ -> ()
+       | Pexp_construct (_,Some _) -> ()
+       | Pexp_variant (_,Some _) -> ()
+       | _ -> not_alloc ()
```
ccasin: We should explicitly list all the `Pexp_*` cases in an or-pattern here, so that when adding a new one we're forced to consider explicitly how it interacts with `malloc_` (like whether it should get the `unsuppoted` error or the `not_alloc` error). Right now a new case would be silently handled by the catch-all pattern, but if we list all the cases then we'll instead get an "incomplete pattern match warning" when a new constructor is added. In general catch-all patterns are good to avoid, to make the compiler robust to that kind of change.

typing/typecore.ml
```diff
@@ -7078,6 +7127,60 @@ and type_expect_
...
+      let unsupported cat =
+        raise (Error (e.pexp_loc, env, Unsupported_external_allocation cat))
+      in
+      let not_alloc () =
+        raise (Error (e.pexp_loc, env, Not_allocation))
+      in
+      (* If we do the same thing as Pexp_stack here and typecheck *anything*
+         before checking if it's actually malloc-able, we get worse error
+         messages. Better to prune based on syntax first, then type errors. *)
+       (match e.pexp_desc with
```
ccasin: The indentation here looks wonky to me - this line is indended one more than the previous bits, but not for any any reason I can see.

typing/typecore.ml
```diff
@@ -7078,6 +7127,60 @@ and type_expect_
...
+       | Pexp_object _ -> unsupported Object
+       | Pexp_pack _ -> unsupported Module
+       | Pexp_array _ -> unsupported Array
+       | Pexp_record _ -> ()
+       | Pexp_tuple _ -> ()
+       | Pexp_construct (_,Some _) -> ()
+       | Pexp_variant (_,Some _) -> ()
+       | _ -> not_alloc ()
+       );
+        let unify_as_mallocd ty_expected =
```
ccasin: Here again it looks like we've unintentionally indented more than the previous parts.

typing/typecore.ml
```diff
@@ -7078,6 +7127,60 @@ and type_expect_
...
+          let inner_ty =
+            newvar (Jkind.Builtin.value_or_null ~why:(Type_argument
+              { parent_path = Predef.path_mallocd ; position = 1; arity = 1}))
+          in
+          let to_unify = Predef.type_mallocd inner_ty in
+          with_explanation (fun () ->
+            unify_exp_types loc env to_unify ty_expected);
+          inner_ty
+        in
+      let inner_ty = unify_as_mallocd ty_expected in
```
ccasin: I think it would be more in keeping with the style of the typechecker not to define a helper function for this, and just do:
```suggestion
      let inner_ty =
        newvar (Jkind.Builtin.value_or_null ~why:(Type_argument
          { parent_path = Predef.path_mallocd ; position = 1; arity = 1}))
      in
      let to_unify = Predef.type_mallocd inner_ty in
      with_explanation (fun () ->
        unify_exp_types loc env to_unify ty_expected)
```
(It looks like there are some other cases that do something like what you've done, but they all need to call the helper more than once)

testsuite/tests/typing-modes/malloc.ml
```diff
@@ -0,0 +1,431 @@
...
+|}]
+
+type t = {a : int; b : int}
+let f () = malloc_ (`Apple {a = 1; b = 3})
+[%%expect {|
+type t = { a : int; b : int; }
+Line 2, characters 27-41:
+2 | let f () = malloc_ (`Apple {a = 1; b = 3})
+                               ^^^^^^^^^^^^^^
+Error: This value is "internal" but expected to be "external_".
```
ccasin, line 550: Does a `t#` work here?  I think yes - maybe good to test.

typing/typedtree.mli
```diff
@@ -518,6 +523,8 @@ and expression_desc =
        Position argument in function application *)
   | Texp_overwrite of expression * expression (** overwrite_ exp with exp *)
   | Texp_hole of unique_use (** _ *)
+  (* CR external-mode: at the moment, this is only ever Allocator_malloc. *)
+  | Texp_alloc of expression * allocator
```
ccasin: This line should get a comment showing what syntax leads to this case - see the other cases above.

typing/printtyped.ml
```diff
@@ -679,6 +679,9 @@ and expression i ppf x =
     expression i ppf e2
   | Texp_hole _ ->
     line i ppf "Texp_hole"
+  | Texp_alloc (e,_) ->
+      line i ppf "Texp_alloc";
+      expression i ppf e;
```
ccasin, line 684: I think we should print the allocator here.  It's too easy to forget to add that printing once the other values are reachable.

typing/typecore.ml
```diff
@@ -409,11 +431,17 @@ type expected_mode =
...
-    strictly_local : bool;
-    (** Indicates that the expression was directly annotated with [local], which
-    should force any allocations to be on the stack. No invariant between this
-    field and [mode]: this field being [true] while [mode] being [global] is
-    sensible, but not very useful as it will fail all expressions. *)
+    (* CR-someday jcutler: This really should not be a part of the
+       [expected_mode]. There are a bunch of places where tying the current
+       allocator to the expected mode is a potential source of confusion or even
+       bugs.
+    *)
```
ccasin: I suggest just folding this into the below comment, doens't really need to be a CR, mentioning that it's a bit semantically surprising for this to be part of expected  mode.  Of course, this was already true about the bool, but it's good to document it.

typing/typecore.ml
```diff
@@ -142,6 +142,23 @@ let print_unsupported_stack_allocation ppf = function
   | List_comprehension -> Format.fprintf ppf "list comprehensions"
   | Array_comprehension -> Format.fprintf ppf "array comprehensions"
 
+
```
ccasin: stray newline

typing/typecore.ml
```diff
@@ -486,21 +514,22 @@ let meet_regional mode =
   let mode = Value.disallow_left mode in
   Value.meet_with Areality Regionality.Const.Regional mode
 
-let mode_default mode =
+(* CR jcutler: optional bad *)
```
ccasin: I think this comment is left over from an intermediate version.

typing/typecore.ml
```diff
@@ -516,31 +545,33 @@ let mode_morph f expected_mode =
 let mode_modality modality expected_mode =
   as_single_mode expected_mode
   |> Modality.Value.Const.apply modality
-  |> mode_default
+  (* CR external-mode: This is another place where the expected mode was getting
+     clobbered in unfortunate ways. *)
```
ccasin: I'm not sure I understand this comment, and it doesn't seem like it needs to be a CR - maybe it's left over from an intermediate version?

I also think we've changed the behavior of this function - it use to make the bool always `false`, now it preserves the allocator (and in particular will preserve `Alloc_stack`). Is this correct?

typing/typecore.ml
```diff
@@ -713,14 +742,27 @@ let lower_bound_externality ~loc ~env lower_bound alloc_mode =
...
+    let externality = Alloc.proj_comonadic Externality alloc_mode in
+    let res = Externality.submode Externality.internal externality in
+    (match res with
+    | Ok () -> ()
+    | Error reason ->
+        let error =
+          Submode_failed(Value.Error
+            (Comonadic Externality,reason), Other, None, None, None, None)
+        in
+        raise (Error(loc, env, error)));
```
ccasin: It looks like the definition of `lower_bound_externality` has been inlined here, but I am not seeing why.

typing/typecore.ml
```diff
@@ -729,13 +771,14 @@ let register_allocation_value_mode ~loc ~env mode =
     of potential subcomponents. *)
 let register_allocation ~loc ~env (expected_mode : expected_mode) =
   let alloc_mode, mode =
-    register_allocation_value_mode ~loc ~env (as_single_mode expected_mode)
+    register_allocation_value_mode ~loc ~env expected_mode.mode
```
ccasin: It looks like the call to `as_single_mode` has been dropped here, which I expect to be a bug for some examples involving tuples (though I'm surprised no such examples surfaced in the test suite).

typing/typecore.ml
```diff
@@ -729,13 +771,14 @@ let register_allocation_value_mode ~loc ~env mode =
...
+      expected_mode.allocator
   in
   let alloc_mode : alloc_mode =
     { mode = alloc_mode;
       locality_context = expected_mode.locality_context }
   in
-  alloc_mode, mode_default mode
+  alloc_mode, mode_default mode ~allocator:Allocator_heap
```
ccasin, line 767: It's surprising to me that we throw away `expected_mode.allocator` here - maybe a comment would be helpful?

typing/typecore.ml
```diff
@@ -5264,9 +5308,9 @@ let split_function_ty
          function deserves a separate allocation mode.
       *)
       let mode, _ = Value.newvar_below (as_single_mode expected_mode) in
-      fst (register_allocation_value_mode ~loc ~env mode)
+      fst (register_allocation_value_mode ~loc ~env mode Allocator_heap)
   in
-  if expected_mode.strictly_local then
+  if expected_mode.allocator = Allocator_stack then
```
ccasin: It's not insane to use polymorphic equality here - it's reasonably defined for enumerations, and this is done a fair bit in older code in the compiler. But I would still mildly prefer replacing this with a pattern match or defining and equality function for allocators.

typing/typecore.ml
```diff
@@ -7078,6 +7127,60 @@ and type_expect_
...
+      let inner_ty = unify_as_mallocd ty_expected in
+      let expected_mode =
+        mode_coerce (Value.max_with_comonadic Externality Externality.external_)
+          expected_mode
+      in
+      let exp =
+        type_expect env {expected_mode with allocator = Allocator_malloc} e
+          {ty = inner_ty; explanation}
+      in
+
```
ccasin: stray newline

typing/typecore.mli
```diff
@@ -207,6 +207,15 @@ type unsupported_stack_allocation =
...
 
+type unsupported_external_allocation =
+  | Lazy
+  | Module
+  | Object
+  | Comprehension
+  | Function
+  | Array
+
+
```
ccasin: extra newline

typing/types.ml
```diff
@@ -234,6 +234,7 @@ module Jkind_mod_bounds = struct
       then Externality.min
       else t.externality
     in
+
```
ccasin: stray newline

typing/typetexp.ml
```diff
@@ -779,6 +779,7 @@ and transl_type_aux env ~row_context ~aliased ~policy mode styp =
                 (newconstr Predef.path_option [Btype.tpoly_get_mono arg_ty])
             end
           in
+
```
ccasin: stray newline

typing/uniqueness_analysis.ml
```diff
@@ -2386,6 +2386,7 @@ let rec check_uniqueness_exp ~overwrite (ienv : Ienv.t) exp : UF.t =
   | Texp_probe { handler } -> check_uniqueness_exp ~overwrite:None ienv handler
   | Texp_probe_is_enabled _ -> UF.unused
   | Texp_exclave e -> check_uniqueness_exp ~overwrite:None ienv e
+  | Texp_alloc (e, _) -> check_uniqueness_exp ~overwrite:None ienv e
```
ccasin, line 2389: Regarding the earlier comment about `malloc_` being mode polymorphic, I suspect here is where we are making that choice.

typing/untypeast.ml
```diff
@@ -726,6 +726,9 @@ let expression sub exp =
     | Texp_overwrite (exp1, exp2) ->
         Pexp_overwrite(sub.expr sub exp1, sub.expr sub exp2)
     | Texp_hole _ -> Pexp_hole
+    | Texp_alloc (exp,Allocator_malloc) -> Pexp_malloc (sub.expr sub exp)
+    (* CR external-mode: Fix this when other allocators are supported. *)
+    | Texp_alloc (_,_) -> assert false
```
ccasin: This is a place where I'd use `Misc.fatal_error` rather than `assert false` because the fact it is unreachable is an invariant of the whole typechecker, not local to this function.

typing/value_rec_check.ml
```diff
@@ -257,7 +257,7 @@ let classify_expression : Typedtree.expression -> sd =
           (* other cases compile to a lazy block holding a function *)
           Static
       end
-
+    | Texp_alloc _ -> Static
```
ccasin: I think it's possible this is right because of our restrictions on what you can malloc, but I think `classify_expression e` would be more future-proof.  And I more concretely propose we make it `Dynamic` and just not support recursive values involving `malloc_`.  Either way, we should have test cases showing what's allowed and what isn't.

typing/printtyped.ml
alpha-convert: Sure, we can -- I didn't take a ton of care around Texp_alloc since it goes away in the next PR.

typing/predef.ml
alpha-convert: I think I agree with this, except for locality... It technically crosses with its argument on locality, but these are disjoint sets. Anything that crosses on locality isn't `malloc_`able, afacit. So I think having it cross on locality is just a bit confusing.

typing/value_rec_check.ml
alpha-convert: Will need some help figuring out how to make a type-correct recursive value using `malloc_` for testing purposes. It's a tough type puzzle

typing/uniqueness_analysis.ml
alpha-convert: If we get rid of `Texp_alloc` and instead make it an `exp_extra` (as we'll do in the next PR), this choice is forced on us.

typing/value_rec_check.ml
alpha-convert: (Also -- this issue disappears in the next PR)

lthls: I usually test with the following snippets:
```ocaml
let rec x =
  expr (* where expr doesn't mention x *)

let rec x =
  let _ = x in
  expr (* where expr doesn't mention x *)
```

This is usually enough to trigger all pieces of code related to recursive value compilation. Trying to generate an actually recursive value isn't usually relevant, and is often easier to do with mutable values than with recursive definitions.

testsuite/tests/typing-modes/malloc.ml
alpha-convert: I think polymorphic variants can only contain types of kind `value`

testsuite/tests/typing-modes/malloc.ml
alpha-convert: Did you mean to have the non-regular recursion here? Or instead

```
type 'a malloced_list = Nil | Cons of 'a * 'a malloced_list malloce
```

alpha-convert: In the first version, the `malloced`s pile up and the `n`th element of the list is `'a malloced^n`. The second version has only infinite inhabitants, since you can't `malloc_ Nil`.

typing/uniqueness_analysis.ml
alpha-convert: (We should figure out how to ensure malloc returns uniquely)

parsing/pprintast.ml
```diff
@@ -27,7 +27,7 @@ open Location
 open Longident
 open Parsetree
 
-let prefix_symbols  = [ '!'; '?'; '~' ]
+let prefix_symbols  = [ '!'; '?'; '~']
```
ccasin: stray change

typing/jkind.mli
```diff
@@ -310,6 +310,8 @@ module Const : sig
     (** The jkind of unboxed 32-bit native-sized integers with no mode crossing. *)
     val word : t
 
+    val kind_of_mallocd : t
```
ccasin: Probably a comment here, like the ones above.

~~And, I think, unlike those, this kind is a little weird - it's not actually the kind of `mallocd`, because it still needs the with-bounds added - and that difference should be documented.~~

Actually, that last bit is dumb of me.  We should just make this a function that takes a `type_expr` for the parameter and compute the jkind _including_ the with bounds, moving the  with bounds logic from predef to here.  (Or, alternatively, don't define this here at all and built it from `word` fully in `predef`, but I recall that was a bit annoying to do for some reason)

typing/predef.ml
```diff
@@ -532,6 +535,25 @@ let build_initial_env add_type add_extension empty_env =
...
+       ~separability:Separability.Sep
+       ~param_jkind:(Jkind.Builtin.value_or_null
+                      ~why:(Type_argument
+                             {parent_path = path_mallocd;
+                              position = 1; arity = 1}))
+       ~jkind:(fun param ->
+                let modality =
+                  Mode.Modality.Value.Const.id
+                  |> Mode.Modality.Value.Const.set (Monadic Uniqueness)
+                      (Join_with Jkind_mod_bounds.Uniqueness.min)
```
ccasin: Something looks funny here - the comment on `Join_with` says that `Join_with min` is the identity, which is what we started with.

I think it shouldn't be necessary to change the with bound for uniqueness because the underlying mod bound doesn't claim we cross on this axis.

typing/typecore.ml
```diff
@@ -5310,9 +5296,11 @@ let split_function_ty
       let mode, _ = Value.newvar_below (as_single_mode expected_mode) in
       fst (register_allocation_value_mode ~loc ~env mode Allocator_heap)
   in
-  if expected_mode.allocator = Allocator_stack then
-    Locality.submode_exn Locality.local
-      (Alloc.proj_comonadic Areality alloc_mode);
+  begin match expected_mode.allocator with
+        | Allocator_stack -> Locality.submode_exn Locality.local
+                              (Alloc.proj_comonadic Areality alloc_mode);
+        | _ -> ()
```
ccasin: Better not to use a catch-all pattern, certainly possible we could add other allocators that need interesitng behavior here in the future.

testsuite/tests/typing-modes/malloc.ml
```diff
@@ -0,0 +1,927 @@
...
+         contention: mod contended with 'a ≰ mod contended
+         visibility: mod immutable with 'a ≰ mod immutable
+|}]
+
+type 'a t : word mod immutable with 'a = 'a mallocd
+[%%expect{|
+type 'a t = 'a mallocd
+|}]
+
+(* Layotu checks: [t mallocd] is ok only if t is value_or_null *)
```
ccasin:
```suggestion
(* Layout checks: [t mallocd] is ok only if t is value_or_null *)
```

testsuite/tests/typing-modes/malloc.ml
```diff
@@ -31,158 +250,130 @@ Error: This type "float#" should be an instance of type "('a : value_or_null)"
          because the type argument of mallocd has layout value_or_null.
 |}]
 
+type t = #(int * int)
+[%%expect {|
+type t = #(int * int)
+|}]
+
```
ccasin: I think this is a stray definition probably left over from an intermediate state.

testsuite/tests/typing-modes/malloc.ml
```diff
@@ -31,158 +250,130 @@ Error: This type "float#" should be an instance of type "('a : value_or_null)"
...
 type t = int or_null mallocd
 [%%expect {|
 type t = int or_null mallocd
 |}]
 
+(* Basic type behavior: does malloc_ returrn a mallocd, and does it
+   requires that its arguments are passed @ external *)
```
ccasin:
```suggestion
(* Basic type behavior: does malloc_ return a mallocd, and does it
   require that its arguments are passed @ external *)
```

ccasin (approved):
Looks good!

PR #4360, opened 2025-07-23
Regalloc tool: record spill/reload costs

xclerc:
This pull request simply records two
additional elements when producing
CSV data using the `regalloc.exe`
tool: the spill and reload costs. The
costs are using a common heuristics:
an instruction costs `10**d` where
`d` is the depth of the loop.

(Ideally the code should be shared
with the allocator, but I think it is OK
to experiment.)

### DIFF
--- a/tools/regalloc/regalloc.ml
+++ b/tools/regalloc/regalloc.ml
@@ -60,11 +60,27 @@ let dummy_in_stats =
 
 type out_stats =
   { spills : int;
-    reloads : int
+    reloads : int;
+    spill_cost : int;
+    reload_cost : int
   }
 
 let row_header =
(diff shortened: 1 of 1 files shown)


tools/regalloc/regalloc.ml
```diff
@@ -157,14 +173,41 @@ let collect_in_stats (cfg_with_infos : Cfg_with_infos.t)
     num_high_pressure_points
   }
 
+let safe_pow10 n =
```
gretay-js, line 176: Consider adding a comment that this is intended to be the same heuristic as in regalloc_utils.

gretay-js (approved):

tools/regalloc/regalloc.ml
xclerc: Thanks, I meant to have such a note and then forgot about it.

PR #4361, opened 2025-07-23
Register allocators: move spills/reloads outside of loops when possible

xclerc:
This pull request is adding an optimization pass to
the split part of the register allocators (the split
part being shared by all allocators).

This pass moves spills and reloads outside of loops
when possible: when a pseudo-register is _live_ at a
destruction point inside a loop but never _used_ inside
the loop, it is indeed possible to spill before the
loop and reload after the loop (hence once) rather
than do so inside the loop (hence at each and every
iteration).

The code has been tested (using the validator) and
benchmarked on the compiler distribution using the
`regalloc.exe` tool. The tables below show, for each
allocator the effect of enabling the optimization,
first in absolute numnbers and then using percentages.

`spills` and `reloads` are simply counting the numbers
of such instructions, while the costs are computed
using a common heuristics: each instruction is given
a weight of `10**d`, where `d` is the depth of the loop.

|               | `irc` / `off` | `irc` / `on` | `ls` / `off` | `ls` / `on` | `gi` / `off` | `gi` / `on` |
| ------------- | ------------- | ------------ | ------------ | ----------- | ------------ | ----------- |
| `spills`      |       206,011 |      208,581 |      211,626 |     213,096 |      210,281 |     212,397 |
| `reloads`     |       283,098 |      284,808 |      294,438 |     295,225 |      287,819 |     288,979 |
| `spill_cost`  |       743,527 |      698,325 |      766,548 |     708,807 |      757,499 |     705,165 |
| `reload_cost` |       942,924 |      857,442 |      991,956 |     889,549 |      957,617 |     865,636 |
| `rounds`      |          1.04 |         1.05 |         1.04 |        1.05 |         1.05 |        1.05 |
| `time`        |   19.30~19.45 |  18.61~18.77 |  12.04~12.29 | 12.15~12.37 |  14.48~14.69 | 14.57~14.81 |

|               | `irc` change | `ls` change | `gi` change |
| ------------- | ------------ | ----------- | ----------- |
| `spills`      |        +1.3% |       +0.7% |       +1.0% |
| `reloads`     |        +0.6% |       +0.3% |       +0.4% |
| `spill_cost`  |        -6.1% |       -7.5% |       -6.9% |
| `reload_cost` |        -9.1% |      -10.3% |       -9.6% |
| `rounds`      |        +1.0% |       +1.0% |        0.0% |
| `time`        |        -3.5% |       +0.9% |       +0.8% |

(caveat: benchmarked on Apple Silicon using Rosetta, so
the `time` metric could be a bit off)


xclerc:
Simple example to show the effects of
the optimization:

```ocaml
let[@inline never] sink _ = ()

let a_loop x y =
  for i = 0 to y do
    sink i;
  done;
  x
```

and the diff (optimization is disabled on
the left an enabled on the right):

```diff
 L_camlT00__a_loop_1_3_code:
 	.cfi_startproc
 	subq	$24, %rsp
 	.cfi_adjust_cfa_offset 24
 	cmpq	$1, %rbx
 	jl	L121
-	movl	$1, %edi
+	movq	%rax, 16(%rsp)
+	movl	$1, %eax
 L111:
-	movq	%rdi, 16(%rsp)
-	movq	%rbx, 8(%rsp)
-	movq	%rax, (%rsp)
-	movq	%rdi, %rax
+	movq	%rax, 8(%rsp)
+	movq	%rbx, (%rsp)
 	call	_camlT00__sink_0_2_code
 L124:
-	movq	(%rsp), %rax
-	movq	8(%rsp), %rbx
-	movq	16(%rsp), %rdi
-	cmpq	%rbx, %rdi
+	movq	(%rsp), %rbx
+	movq	8(%rsp), %rax
+	cmpq	%rbx, %rax
 	je	L117
-	addq	$2, %rdi
+	addq	$2, %rax
 	jmp	L111
 	.align	2
 L117:
+	movq	16(%rsp), %rax
 	addq	$24, %rsp
 	.cfi_adjust_cfa_offset -24
 	ret
 	.cfi_adjust_cfa_offset 24
 	.align	2
 L121:
 	addq	$24, %rsp
 	.cfi_adjust_cfa_offset -24
 	ret
 	.cfi_adjust_cfa_offset 24
 	.cfi_adjust_cfa_offset -24
```

backend/regalloc/regalloc_split_state.ml
```diff
@@ -51,6 +51,157 @@ let log_renaming_info : t -> unit =
...
+        let destroyed_here =
+          match Label.Map.find_opt label destructions_at_end with
+          | None -> Reg.Set.empty
+          | Some (_kind, regs) -> regs
+        in
+        let destroyed = Reg.Set.union destroyed destroyed_here in
+        let block = Cfg.get_block_exn cfg label in
+        let[@inline] update_occuring :
+            type a. Reg.Set.t -> a Cfg.instruction -> Reg.Set.t =
+         fun occurring instr ->
```
TheNumbat, line 80: This doesn't need to add regs destroyed by the instruction to `destroyed` because those are always hardware regs, right? Could use a comment.

backend/regalloc/regalloc_split_state.ml
```diff
@@ -51,6 +51,157 @@ let log_renaming_info : t -> unit =
...
+      in
+      let destructions_at_end : destructions_at_end =
+        Label.Set.fold
+          (fun label acc ->
+            if debug then log "destructions now happen at %a" Label.print label;
+            Label.Map.update label
+              (function
+                | None -> Some (Destruction_on_all_paths, to_move)
+                | Some (_kind, regs) ->
+                  (* CR xclerc for xclerc: double-check the kind is correct. *)
```
TheNumbat, line 151: Is there a reason this can't be `Destruction_only_on_exceptional_path`?

TheNumbat (approved):
Cool, thanks. I agree it'll need performance testing in the tree but looks good to merge.

backend/regalloc/regalloc_split_state.ml
xclerc: I just double-checked; `Destruction_only_on_exceptional_path`
is for when the normal edges but exceptional one do. So, I think
that makes the kind correct here. But it is not optimal: if the
predecessor has multiple successors, we want to have the
destruction only on the edge leading to the loop. The easiest
solution is to insert blocks where needed.

backend/regalloc/regalloc_split_state.ml
xclerc: My description was a bit imprecise:
the split pass considers registers
"destroyed" only at destruction points
(basically, calls). This is because we
nnow before allocation that such
destructions will result in spills/reloads
(so, we decide to add them before
allocation itself). The registers in the
destroyed list of an instruction don't
have this property in general. (I think
there might be something interesting
to do when the pressure is high, but
I have not spend enough time on it.)

backend/regalloc/regalloc_split_state.ml
TheNumbat: I see, but is it ok to replace an existing `Destruction_only_on_exceptional_path` kind with all paths?
Or is that never the case when entering this optimization pass?

mshinwell: @xclerc ping on this comment

xclerc: Sorry, I missed the notification. I think it is
always safe to change `only_on_exception`
to `on_all_paths`, but it might create more
spills.

In this particular case, I think it is (very?)
unlikely we will switch from one kind to the
other because that would mean "entering"
a loop through an exceptional edge.

That being said, as alluded to in a previous
comment, this pull is suboptimal (indepen-
dently of the exception case) because it
adds a destruction point on every prede-
cessor. I will work on a patch to address
that, and -depending on its complexity-
either push it to this branch, or create a
new pull request.

xclerc: I don't think it is possible with the current
compiler to have a loop predecessor whose
terminator is not `Always _`, because of
the way loops are being compiled and the
handling of critical edges at the start of the
"split" pass. (I have double-checked that on
the compiler distribution by instrumenting
the code in this pull request.)

However, it would be better to not depend
on the way loops are being compiled
(basically, always entering from the head).
I think it should be done in a follow-up pull
request.

PR #4362, opened 2025-07-23
Add Pgettag primitive with atomic loads for variant_only=false

mshinwell:
## Summary

This PR adds a new `Pgettag` primitive to Lambda, exposes it to users as `%get_tag`, and implements atomic acquire loads for `Get_tag { variant_only = false }` cases, following the synchronization requirements noted in `runtime/obj.c`.

### Lambda Layer
- Added `Pgettag of { variant_only : bool }` primitive to `lambda.ml{,i}` 
- Wired `Pgettag` through `lambda_to_flambda_primitives.ml` to convert to Flambda2's `Get_tag`
- Exposed `%get_tag` %-primitive in `translprim.ml` with `variant_only = false`
- Updated `matching.ml` to use `Pgettag` instead of `caml_obj_tag` calls
- Fixed integer handling crash by checking `Pisint` first, then using appropriate tag values

### Flambda2 Layer  
- Updated `Get_tag` primitive to include `variant_only` parameter (already done in previous work)
- Modified `to_cmm_primitive.ml` to distinguish between:
  - `Get_tag { variant_only = true }`: Uses regular `get_tag` (non-atomic)  
  - `Get_tag { variant_only = false }`: Uses new `get_tag_atomic` (atomic)

### CMM Layer
- Added `get_header_atomic` function using `mk_load_atomic Word_int` 
- Added `get_tag_atomic` function that uses atomic header loads + bitwise AND
- Both functions kept internal to `cmm_helpers.ml` (not exposed in interface)

### Pattern Matching Updates
- Updated all files with `Get_tag` pattern matches across the codebase
- Added `Pgettag` pattern matches to Lambda processing files
- Fixed compilation errors and warnings

## Rationale

The atomic acquire loads are needed for `variant_only = false` because `runtime/obj.c` explicitly uses `atomic_load_acquire` for reading headers, noting this is required for Forward_tag block synchronization in lazy evaluation.

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/backend/cmm_helpers.ml
+++ b/backend/cmm_helpers.ml
@@ -1287,8 +1287,8 @@ let get_header ptr dbg =
       dbg )
 
 let get_header_atomic ptr dbg =
-  (* Atomic header load for Get_tag with variant_only=false, as per obj.c comment
-     about Forward_tag synchronization needing atomic acquire loads *)
+  (* Atomic header load for Get_tag with variant_only=false, as per obj.c
+     comment about Forward_tag synchronization needing atomic acquire loads *)
   Cop
     ( mk_load_atomic Word_int,
       [Cop (Cadda, [ptr; Cconst_int (-size_int, dbg)], dbg)],
--- a/backend/cmm_helpers.ml
+++ b/backend/cmm_helpers.ml
@@ -1286,6 +1286,14 @@ let get_header ptr dbg =
       [Cop (Cadda, [ptr; Cconst_int (-size_int, dbg)], dbg)],
       dbg )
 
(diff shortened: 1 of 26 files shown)


mshinwell:
Fixed the test to handle both flat and non-flat float array configurations. The test now checks Config.flat_float_array to determine the expected tag:
- When flat float arrays are enabled: expects tag 254 (Obj.double_array_tag)
- When disabled: expects tag 0 (regular array tag)

This ensures the test passes correctly with or without the float array optimization.

@stedolan

PR #4363, opened 2025-07-23
Primitives for raw OCaml block access

mshinwell:
## Summary

This PR extracts the Flambda2 parts of the block indices work from PR #4017 (rtjoa.block-indices). It adds two new primitives that will enable faster field access in unusual use cases, similar to Obj.raw_field but with better performance.

## Changes

- **Read_offset**: Binary primitive that reads from a memory location at a given offset
- **Write_offset**: Ternary primitive that writes to a memory location at a given offset

Both primitives include:
- Proper type kinds and mutability/allocation mode tracking
- Placeholder CMM translations (add offset to base pointer, then load/store)
- Code size estimates
- Basic simplification support

This is a draft PR as these primitives will need user-facing wrappers before they can be used.

Co-Authored-By: Claude <noreply@anthropic.com>


mshinwell:
(Not yet rebased onto `main`)

mshinwell:
I've added two new primitives to support raw field access that bypasses some safety checks:

- `%raw_field`: Takes a block and a field number (as a tagged immediate), returns the value at that field
- `%set_raw_field`: Takes a block, field number, and new value, sets the field

These Lambda primitives (`Prawfield` and `Psetrawfield`) translate to the Flambda2 `Read_offset` and `Write_offset` primitives. The implementation handles the conversion from field numbers (in words) to byte offsets during the Lambda to Flambda2 translation.

The primitives assume:
- Values are of kind `Value` (corresponding to `Pgenval` in Lambda)  
- Blocks are mutable
- Field numbers are provided as tagged immediates

These primitives provide faster field access for cases where the normal safety guarantees are not needed.

Co-Authored-By: Claude <noreply@anthropic.com>

mshinwell:
Fixed the primitives - %raw_field now returns nativeint# and %set_raw_field now takes nativeint# for the new value.

-- Claude

mshinwell:
I've added comprehensive tests for the new `%raw_field` and `%set_raw_field` primitives:

- `testsuite/tests/basic/raw_field_primitives.ml` - Tests for tuples, variants, and records (both mutable and immutable) that work in both bytecode and native modes
- `testsuite/tests/basic/raw_field_primitives_nativeint.ml` - Native-only tests for records containing nativeint# fields

The tests verify that the primitives correctly read and write field values as unboxed nativeint# values.

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
```diff
@@ -2727,6 +2727,72 @@ let convert_lprim ~big_endian (prim : L.primitive) (args : Simple.t list list)
     [Ternary (Atomic_field_int_arith Or, atomic, field, i)]
   | Patomic_lxor_field, [[atomic]; [field]; [i]] ->
     [Ternary (Atomic_field_int_arith Xor, atomic, field, i)]
+  | Prawfield, [[block]; [field]] ->
+    (* Convert field number (in words) to byte offset *)
+    (* First untag the field index *)
+    let untagged_field = H.Prim (Unary (Untag_immediate, field)) in
+    (* Word size is 8 on 64-bit systems, 4 on 32-bit systems *)
+    let word_size = if Targetint_32_64.size = 64 then 8 else 4 in
```
rtjoa, line 2735: I think this is dead code, as the native compiler only supports 64-bit architectures

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
```diff
@@ -2727,6 +2727,72 @@ let convert_lprim ~big_endian (prim : L.primitive) (args : Simple.t list list)
...
+      H.Simple
+        (Simple.const
+           (Reg_width_const.naked_immediate (Targetint_31_63.of_int word_size)))
+    in
+    let byte_offset =
+      H.Prim
+        (Binary
+           (Int_arith (Naked_immediate, Mul), untagged_field, word_size_const))
+    in
+    (* Convert untagged immediate to naked int64 for Read_offset *)
```
rtjoa, line 2746: Infinitesimal nit - IMO this is over-commented, and it might be good to avoid the precedent for future AI changes

middle_end/flambda2/from_lambda/lambda_to_flambda_primitives.ml
```diff
@@ -2727,6 +2727,72 @@ let convert_lprim ~big_endian (prim : L.primitive) (args : Simple.t list list)
...
+    let byte_offset_int64 =
+      H.Prim
+        (Unary
+           (Num_conv { src = Naked_immediate; dst = Naked_int64 }, byte_offset))
+    in
+    (* Wrap block in Opaque_identity to prevent Flambda2 from analyzing it *)
```
rtjoa, line 2752: I think this is necessary specifically because we are reinterpreting things as native ints - if so would be nice to note that in the comment.

testsuite/tests/basic/raw_field_primitives.ml
```diff
@@ -0,0 +1,113 @@
+(* TEST *)
```
rtjoa, line 1: I think these particular tests would be good to have:
1. Accesses of mixed blocks that get reordered
2. Accesses of values in all-float and mixed-float-float# records (if supported - see top-level comment)
3. Accesses of values before and after `bits32`, `void`, and `vec128` fields
4. Accesses of {scannable, non-scannable} {product, non-product} arrays

middle_end/flambda2/to_cmm/to_cmm_primitive.ml
```diff
@@ -1185,6 +1189,20 @@ let ternary_primitive _env dbg f x y z =
     bytes_or_bigstring_set ~dbg kind width ~bytes:x ~index:y ~new_value:z
   | Bigarray_set (_dimensions, kind, _layout) ->
     bigarray_store ~dbg kind ~bigarray:x ~index:y ~new_value:z
+  | Write_offset (kind, mode) ->
```
rtjoa, line 1192: Someone else should check changes to this file, to avoid self-review. In early drafts, this was also a source of bugs, e.g. the appropriate choice of {`store`,`caml_modify`,`caml_modify_local`}, so it would be good to read closely.

middle_end/flambda2/parser/flambda_to_fexpr.ml
```diff
@@ -655,7 +655,7 @@ let ternop env (op : Flambda_primitive.ternary_primitive) : Fexpr.ternop =
     Array_set (ak, ask)
   | Bytes_or_bigstring_set (blv, saw) -> Bytes_or_bigstring_set (blv, saw)
   | Bigarray_set _ | Atomic_field_int_arith _ | Atomic_set_field _
-  | Atomic_exchange_field _ ->
+  | Atomic_exchange_field _ | Write_offset _ ->
```
rtjoa, line 658: Forwarding a [comment](https://github.com/oxcaml/oxcaml/pull/4017/files#r2202669877) from @ccasin on #4017 so it doesn't get dropped:
> It's fine to do it in a follow-up PR, but I think it would be good to implement `fexpr` support for the block index primitives.  It probably takes just 10 minutes, and in my experience can be a real annoyance for minimizing problems and making reproducable tests when things aren't supported.  (I defer to @mshinwell though).

rtjoa (commented):
Looks good overall. Comments not corresponding to a particular source location:
- I think it would be good to comment somewhere the assumptions of the primitives, in particular that the fields they access are values.
- Are float records supported? Mixed float/float# records?

(edited this comment after noticing your comment above).

middle_end/flambda2/parser/flambda_to_fexpr.ml
rtjoa: (Of course, I can just do this when I get back to #4017 if you think it's necessary to include.)

PR #4364, opened 2025-07-23
Forbid (non-legacy) modalities on atomic fields

glittershark:
See internal doc titled "Modalities for atomic fields" for more context and background.


typing/typedecl.ml
```diff
@@ -4778,6 +4792,12 @@ let report_error ppf = function
       fprintf ppf
         "@[The label %a must be mutable to be declared atomic.@]"
         Style.inline_code name
+  | Modalities_on_atomic_field name ->
+    fprintf ppf
+      "@[Non-legacy modalities are not allowed on atomic fields (here, %a).\n\
+       Use one of the modality types from the Modes module in the type of the\n\
+       field instead]"
```
riaqn: "Legacy modalities" is confusing - it could mean many things.
Specifically here, I think we will say: "Modalities of comonadic axes are not allowed on atomic fields"

(Similarly for other changes in the PR)

riaqn (approved):

typing/typedecl.ml
glittershark: I didn't say "comonadic" since I think we're trying not to say that in documentation currently, instead I went with "future". Can you check that you're happy with the new state?

testsuite/tests/atomic-locs/record_fields.ml
```diff
@@ -82,6 +82,32 @@ end)
...
+module Modalities_forbidden = struct
+  type t = { mutable x : (int -> int) @@ portable [@atomic] }
+end
+[%%expect{|
+Line 2, characters 13-59:
+2 |   type t = { mutable x : (int -> int) @@ portable [@atomic] }
+                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Error: Modalities of future axes are not allowed on atomic fields
+(here, "x"). Use one of the modality types from the Modes module in the
+type of the field instead]
```
goldfirere:
```suggestion
type of the field instead
```

typing/typedecl.ml
```diff
@@ -499,6 +500,17 @@ let transl_labels (type rep) ~(record_form : rep record_form) ~new_var_jkind
          let modalities =
           Typemode.transl_modalities ~maturity:Stable mut modalities
          in
+         (if is_atomic
+          then
+            (* Forbid comonadic modalities on atomic fields, as we can't force
+               them to be respected when setting them via eg [%atomic.loc]. *)
+            match
+              Mode.Modality.Value.Const.equate modalities
```
goldfirere, line 508: The error message calls out only future axes... but this looks like it checks past axes too. Is that intentional?

typing/typedecl.ml
```diff
@@ -4778,6 +4790,12 @@ let report_error ppf = function
       fprintf ppf
         "@[The label %a must be mutable to be declared atomic.@]"
         Style.inline_code name
+  | Modalities_on_atomic_field name ->
+    fprintf ppf
+      "@[Modalities of future axes are not allowed on atomic fields\n\
+       (here, %a). Use one of the modality types from the Modes module in the\n\
+       type of the field instead]"
```
goldfirere:
```suggestion
       type of the field instead@]"
```

testsuite/tests/atomic-locs/record_fields.ml
goldfirere, line 1: I think there should also be a test showing that the default modalities for atomic fields are identity, not legacy.

typing/typemode.ml
```diff
@@ -479,10 +479,10 @@ let mutable_implied_modalities ~for_mutable_variable (mut : Types.mutability) =
   in
   match mut with
   | Immutable -> []
-  | Mutable _ -> if for_mutable_variable then monadic else monadic @ comonadic
+  | Mutable _ -> if include_comonadic then monadic @ comonadic else monadic
```
goldfirere: There's something funny here: we check whether this is atomic before getting here, but then also case on the mutability. I think we never include the comonadic ones when the mutability is atomic, and this could easily be encoded right here instead of distributed across the call sites.

goldfirere (approved):

typing/typedecl.ml
glittershark: ech, yeah, good point. I think maybe we should just make the error say "modalities", and leave the fact that identity modalities are in fact allowed somewhat undocumented (I'm not sure anyone is likely to care, as it's mostly an academic point). WDYT?

goldfirere: I'm fine with that.

typing/typemode.ml
glittershark: yeah, good point. I restored the previous `for_mutable_variable` flag and just match on that and atomicity here now.

glittershark: ah, yeah, that's much better. Good call

testsuite/tests/atomic-locs/record_fields.ml
glittershark: Added a bunch of default modality tests, would love a second pair of eyes on them.

typing/typemode.ml
```diff
@@ -477,9 +477,16 @@ let mutable_implied_modalities ~for_mutable_variable (mut : Types.mutability) =
       Atom (Monadic Contention, Join_with Contention.Const.legacy);
       Atom (Monadic Visibility, Join_with Visibility.Const.legacy) ]
   in
-  match mut with
-  | Immutable -> []
-  | Mutable _ -> if for_mutable_variable then monadic else monadic @ comonadic
+  (* We only need to include comonadic mutable implied modalities for now for
+     backwards-compatibility with legacy code which expects them. Atomic record
+     fields and mutable variables are a features, and so don't have this
```
riaqn, line 482: "are new features"?

testsuite/tests/atomic-locs/record_fields.ml
```diff
@@ -334,3 +360,56 @@ type suppressed_via_mnemonic = { mutable a : float [@atomic] }
 0
 type suppressed_via_mnemonic = { mutable a : float [@atomic]; }
 |}]
+
+(* Tests for default modalities: atomic fields imply legacy modalities for
+   comonadic axes only *)
```
riaqn, line 365: "for monadic axes only"

testsuite/tests/atomic-locs/record_fields.ml
```diff
@@ -82,6 +82,32 @@ end)
...
+Line 2, characters 13-59:
+2 |   type t = { mutable x : (int -> int) @@ portable [@atomic] }
+                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Error: Modalities are not allowed on atomic fields (here, "x"). Use one of
+the modality types from the Modes module in the type of the field
+instead
+|}]
+
+(* Legacy modalities are allowed because they're the same thing as writing
+   nothing at all *)
```
riaqn, line 98: "Legacy modalities on monadic axes are allowed because ..."

PR #4367, opened 2025-07-24
CI: fix path to `regalloc.exe`

xclerc:
In #4301, I forgot to update the path
to `regalloc.exe` in the CI jobs.
Unfortunately the error exit code was
swallowed by (I think) either the bash
`for` loop or `find`.


xclerc:
(I have checked on the log that this
pull request fixes the issue.)

mshinwell (approved):

PR #4369, opened 2025-07-24
flambda2: More maintainable implementation of n-way join

bclement-ocp:
The n-way join algorithm introduced in #3538 is written in a somewhat confusing way (see also: #4279) and is full of subtleties. This is partly due to the way the `Join_aliases` module is dealing with more responsibilities than necessary, and partly because we try to do too much in a single pass over the environments. This makes the implementation deviate somewhat from the underlying algorithm it is supposed to implement; for instance, it was not clear how to correctly add the information required for the match-in-match heuristic even though it should theoretically have been straightforward.

This patch rewrites the implementation to more closely follow the underlying algorithm, and should greatly improve readability and maintainability. It also improves performance slightly due to using the efficient leapfrog join from the Datalog engine, but I don't expect this to matter too much in practice (although it does mean that the join can now have sub-linear performance when the joined levels have little variables in common).

I suggest not trying to review the diff on the `Join_env` module and to just read the new implementation. The implementation consists in four parts:

 1) Some things related to iterators and the computation of an
    incremental join using Datalog's seminaive algorithm. This should be
    shared with the `Flambda_datalog` module at some point, which also
    has an implementation of that algorithm, but it's not trivial to use for
    our purpose here.

 2) A bunch of modules for type-safety to represent simples and types in
    each of the environments involved in a join.  This is essentially
    lifted from the previous implementation, but with some more methods.

 3) Supporting modules, split off from the responsibilities of the old
    `Join_aliases` module. These are the `Bindings_in_source_env`,
    `Bindings_in_target_env`, and `Joined_envs` modules.

 4) The join itself (and supporting functions) at the toplevel.

Part 1) can be reviewed independently of the rest. Part 2 should be wholly unsurprising and self-explanatory. I suggest reading part 4 first, looking at part 3 when necessary, before reviewing part 3 proper.

Reviewers: @Ekdohibs  for part 1, @lthls for the rest.

### DIFF
--- a/middle_end/flambda2/types/env/join_env.ml
+++ b/middle_end/flambda2/types/env/join_env.ml
@@ -13,6 +13,77 @@
 (*                                                                        *)
 (**************************************************************************)
 
+(* Implement the join of typing envs, or more precisely of typing env levels.
+
+   Most of the code here is actually concerned with the join of *aliases*
+   specifically (keeping track of how names change between the different
+   environments), and delegates the actual join of types to the
+   [Meet_and_n_way_join] module.
+
+   The join involves multiple environments that are known under different names.
+   Within this file, we standardise on the following names:
+
+   - The {b source environment} is the initial value (before the join) of the
+   environment that we will extend. This is also called the "definition typing
+   env" in join_levels.ml; in the context of [Simplify], this is the environment
+   we would use to simplify the handler when not doing a join. This is different
+   from the "env at fork" in that the source environment is expected to already
+   have definitions for the params (and extra params) of the current handler.
+
+   - The {b target environment} is the final value (after the join) of the
+   source environment. This is also called the "handler env" in [Simplify]. This
+   environment does not exist until the join is completed, but it is still
+   helpful to refer to things that will exist there (those either also exist in
+   the source environment, or are existential variables added during the join).
+
+   - The {b joined environments} are each of the individual environments that we
+   are joining. In the context of [Simplify], these are the environments at each
+   use. The joined environments are uniquely identified (within the current
+   join) by an {!Index.t}.
+
+   {1:assumptions Assumptions}
+
+   We make the following assumptions on the input environments.
+
+   {2:scope_of_names Scope of variables and symbols}
+
+   We assume that any name (variable or symbol) defined in the source
+   environment is also be defined in all the joined environments.
+
+   Any name defined in the source environment is also necessarily defined in the
+   target environment by definition.
+
+   {2:lifted_constants Lifted constants}
+
+   We further assume that any symbol defined in one of the joined environments
+   is also defined in the source environment (and hence the target environment).
+   In the context of [Simplify], this means that we expect lifted constants from
+   the joined environments to have already been inserted into the source
+   environment with a suitable type.
+
+   In practice, this means that any of the symbols we manipulate can be assumed
+   to exist in both the source environment and in the target environemnt (but
+   not in the joined environments, as they could be lifted constants from
+   another branch).
+
+   {2:coherent_binding_times Coherent binding times}
+
+   We assume that a {b the relative order of variables defined in the source
+   environment is preserved across all the joined environments}.
+
+   More precisely, if [a] is defined before (resp. strictly before) [b] in the
+   source environment, then [a] is also defined before (resp. strictly before)
+   [b] in all of the joined environments. In the context of [Simplify], this
+   means that the continuation parameters must be added in the same order in the
+   handler and at all uses.
+
+   Note that this assumption does not impose any restriction on the relative
+   binding times of variables that don't exist in the source environment, even
+   if they exist in all the joined environments.
+
+   This assumption is used in [get_possible_canonical_in_source_env] and allows
+   an efficient (linear) implementation of this function. *)
+
 module K = Flambda_kind
 module TG = Type_grammar
 module TE = Typing_env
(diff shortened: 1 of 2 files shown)


middle_end/flambda2/types/env/join_env.ml
```diff
@@ -13,6 +13,77 @@
...
+   environment with a suitable type.
+
+   In practice, this means that any of the symbols we manipulate can be assumed
+   to exist in both the source environment and in the target environemnt (but
+   not in the joined environments, as they could be lifted constants from
+   another branch).
+
+   {2:coherent_binding_times Coherent binding times}
+
+   We assume that a {b the relative order of variables defined in the source
```
lthls:
```suggestion
   We assume that {b the relative order of variables defined in the source
```

middle_end/flambda2/types/env/join_env.ml
```diff
@@ -264,1031 +612,1132 @@ end = struct
...
+      canonicals_in_joined_envs No_simples_in_joined_envs
 end
 
-module Join_aliases : sig
+module Bindings_in_target_env : sig
+  (* This module is only concerned with providing a consistent name to represent
+     a set of simples in the joined environments.
+
+     Names in the target environment are either names that exist in the source
+     environment, and local variables that are created in the target environment
```
lthls:
```suggestion
     environment, or local variables that are created in the target environment
```

middle_end/flambda2/types/env/join_env.ml
```diff
@@ -264,1031 +612,1132 @@ end = struct
...
                 | Some _ -> raise Not_found))
-          (t :> Simple.t Index.Map.t)
+          (t : t :> Simple.t Index.Map.t)
           None
       with Not_found -> None
     in
     match shared_name with Some raw_name -> raw_name | None -> "join_var"
+end
+
+module Bindings_in_source_env : sig
```
lthls: I'm not completely convinced by this name. The implementation only contains the source environment, and looking at the exposed values I think it would work as well if the module was named just `Source_env`.

middle_end/flambda2/types/env/join_env.ml
```diff
@@ -264,1031 +612,1132 @@ end = struct
...
+     and [x: (= b)] in env 1, where [x] exists in the source environment but not
+     [a] and [b]. Then we can use [x] to represent [((0 a) (1 b))], we do not
+     have to create a local variable. Note that in the case of imported
+     variables, this effectively mean that we can rename variables as we import
+     them. *)
+
   type t
 
-  val empty : t
+  val create : TE.t -> TE.t Index.Map.t -> t
```
lthls: I'm fine passing the joined envs even though they're not actually used, but I think naming the first argument would be nice. I believe that `~source_env` would be a suitable label.
I think the name could be changed to make it more obvious that it merely sets up an initial value to use as a starting point, although I don't have a good suggestion either (I've thought of `create_empty` or `init`, but they don't feel much better).

middle_end/flambda2/types/env/join_env.ml
```diff
@@ -264,1031 +612,1132 @@ end = struct
...
+  val alias_types_in_target_env :
+    t -> Type_in_target_env.t Name_in_source_env.Map.t
+
+  (* Assuming that [since] derives from [t], returns the definitions of local
+     variables that have been added to [t] after [since]. *)
+  val new_bindings :
+    t -> since:t -> definition_in_joined_envs Name_in_target_env.Map.t
+
+  (* val alias_types_in_target_env : t -> Type_in_target_env.t
+     Name_in_target_env.Map.t *)
```
lthls: Left-over from a previous version ?

middle_end/flambda2/types/env/join_env.ml
```diff
@@ -264,1031 +612,1132 @@ end = struct
...
+
+         {b Note}: This is not necessarily the variable itself. For instance, if
+         there is a variable [x] in the source environment that gets demoted to
+         a local variable [y] in all joined environments, then [x] might be used
+         as a name for [y]. *)
+      created_variables : K.t Variable_in_target_env.Map.t;
+      (* This contains all the local variables (existentials or imported) that
+         exist in the target environment but not in the source environment. *)
+      definitions_in_joined_envs :
+        definition_in_joined_envs Name_in_target_env.Map.t;
```
lthls, line 890: I assume that using `Name_` is more convenient here, but do we agree that all the names used as keys in this map are actually variables ?

middle_end/flambda2/types/env/join_env.ml
```diff
@@ -264,1031 +612,1132 @@ end = struct
...
+        Misc.fatal_errorf
+          "Cannot add alias of %a to itself with a non-identity coercion@ %a"
+          Simple_in_target_env.print canonical_element Coercion.print
+          coercion_to_canonical
+    else
+      let alias_types_in_target_env =
+        Name_in_source_env.Map.add name_to_be_demoted
+          (Type_in_target_env.alias_type_of kind
+             (Simple_in_target_env.apply_coercion_exn canonical_element
+                (Coercion.inverse coercion_to_canonical)))
```
lthls: I think it would be slightly simpler if we computed directly the coercion from the canonical to the name to be demoted (it would remove all occurrences of `Coercion.inverse` in this function).

middle_end/flambda2/types/env/join_env.ml
```diff
@@ -264,1031 +612,1132 @@ end = struct
...
+            Name_in_target_env.Map.add name_in_target_env
+              coercion_from_joined_to_target aliases)
+      in
+      ( canonical_in_target_env,
+        { t with
+          equations_for_local_vars;
+          existential_for_these_simples;
+          aliases_of_names_in_joined_envs;
+          definitions_in_joined_envs
+        } )
```
lthls: This whole section is a bit messy. I think it would make sense to move the (small) part about updating `existential_for_these_simples` at the top, and add a comment that the rest is about nested joins.

middle_end/flambda2/types/env/join_env.ml
```diff
@@ -264,1031 +612,1132 @@ end = struct
...
+    Printexc.raise_with_backtrace Misc.Fatal_error bt
+
+(** {2:extensions} Join of extensions *)
+
+let prepare_nested_join ~meet_type ~joined_envs ~bindings extensions =
+  let extensions =
     List.fold_left
-      (fun (joined_levels, joined_envs) (index, extension) ->
-        let parent_env = get_nth_joined_env index t.joined_envs in
+      (fun joined_envs_and_levels (index, extension) ->
```
lthls: The name `joined_envs_and_levels` was a bit confusing (particularly since it is bound to a variable named `extensions` after the fold finishes).  I think that using `joined_envs_and_extensions` for both the accumulator and the final result would be better.

middle_end/flambda2/types/env/join_env.ml
```diff
@@ -264,1031 +612,1132 @@ end = struct
...
+          : Coercion.t Name_in_target_env.Map.t Name_in_one_joined_env.Map.t
+          :> Coercion.t Name_in_target_env.Map.t Name.Map.t)
+        ~init:equations
+        ~f:(fun [@inline] name ty aliases equations ->
+          let kind = Type_in_one_joined_env.kind ty in
+          let ty = TG.alias_type_of kind (Simple.name name) in
+          Name_in_target_env.Map.fold
+            (fun alias coercion equations ->
+              (* alias = coercion(name) *)
+              let ty = TG.apply_coercion ty coercion in
```
lthls: Not a big deal, but I would appreciate if this was `TG.alias_type_of (Simple.with_coercion (Simple.name name) coercion)`. I'm not particularly worried about performance, but I think that `TG.apply_coercion` is a dangerous function and I'd rather not use it if we don't need to.

middle_end/flambda2/types/env/join_env.ml
```diff
@@ -1300,112 +1749,263 @@ let n_way_join_env_extension ~n_way_join_type ~meet_type t envs_with_extensions
...
-    Ok (Simple.var var), { t with existential_vars; join_aliases; pending_vars }
+  Index.Map.mapi
+    (fun index (env, diff_ext) ->
+      let previous_equations =
+        Joined_envs.equations_in_nth_joined_env joined_envs index
+      in
+      let diff_equations =
+        Type_in_one_joined_env.create_equations (TEE.to_map diff_ext)
+      in
+      (* The call below to [reintroduce_demotions_in_one_joined_env] is only
```
lthls: There are multiple references to `reintroduce_demotions_in_one_join_env` in comments; if I'm not mistaken, these should be renamed to `replay_definition_of_aliases_in_target_env`.

lthls (commented):
I've (finally) gone through all the code. I've left various comments, in particular pointing out when the name of a variable made it harder to understand something.

I'm pleased with the code in general. There are various pieces of code that are similar to one another, and I think this could deserve more corss-references (i.e. if one modifies `join_aliases_into_bindings` then one may need to update `join_aliases_in_env_extension` too), but I haven't seen cases where factoring the code wouldn't make it hard to understand.

middle_end/flambda2/types/env/join_env.ml
bclement-ocp: I agree, I will rename it to `Source_env`. I used `Bindings_in_source_env` for symmetry with `Bindings_in_target_env`, but I don't think there is actually any symmetry there.

middle_end/flambda2/types/env/join_env.ml
bclement-ocp: I don't remember why the joined envs are here, they could be removed. Maybe simply  `val from_source_env : Source_env.t -> t`?

middle_end/flambda2/types/env/join_env.ml
bclement-ocp: Yes, removed.

middle_end/flambda2/types/env/join_env.ml
bclement-ocp: That is correct — any name that gets demoted can be passed as the `name_in_source_env` for `add_imported_var` / `add_existential_for_these_simples` and end up in `definitions_in_joined_envs`, and that is the only place where symbols could be added here, but symbols can't (currently) get demoted.

middle_end/flambda2/types/env/join_env.ml
bclement-ocp: I think that might be true — IIRC I simply reproduced the logic of the `Aliases` module because I didn't want to think about the arguments of `Coercion.compose`.

middle_end/flambda2/types/env/join_env.ml
bclement-ocp: Ah, that's true, the one thing that is actually important outside of env extensions is sneakily hidden in the middle..

middle_end/flambda2/types/env/join_env.ml
bclement-ocp: Sure, will fix this and other places to remove `TG.apply_coercion`.

middle_end/flambda2/types/env/join_env.ml
bclement-ocp: Indeed, good catch. Sadly the ocamllsp server does not check references in comments yet (:

middle_end/flambda2/types/env/join_env.ml
bclement-ocp: That's fair, especially since `joined_envs_and_levels` does not contain levels! I forgot to rename the variable.

PR #4371, opened 2025-07-24
Rename function to avoid confusion (peephole rules)

xclerc:
As per title; the name made sense when
the function was first introduced, but
after switching to doubly-linked lists we
don't need to "reverse" elements anymore.


backend/peephole/peephole_rules.ml
```diff
@@ -37,12 +37,12 @@ let remove_overwritten_mov (cell : Cfg.basic Cfg.instruction DLL.cell) =
           | Const_vec256 _ | Const_vec512 _ ) ) ->
       (* Removing the second instruction is okay here since it doesn't change
```
TheNumbat:
```suggestion
      (* Removing the first instruction is okay here since it doesn't change
```

TheNumbat (approved):

PR #4374, opened 2025-07-24
Match in match heuristic

bclement-ocp:


driver/oxcaml_flags.ml
```diff
@@ -310,9 +314,16 @@ module Flambda2 = struct
...
       fallback_inlining_heuristic = false;
+      cont_lifting_budget = 100;
+      cont_spec_threshold = 0.;
     }
 
-    let o3 = default
+    let o3 = {
+      default with
+      cont_lifting_budget = 1_000;
+      (* in the worss case : 1_000 budget -> ~+18% compilation time *)
```
Ekdohibs:
```suggestion
      (* in the worst case : 1_000 budget -> ~+18% compilation time *)
```

middle_end/flambda2/simplify/simplify_switch_expr.ml
```diff
@@ -667,6 +667,169 @@ let simplify_arm ~typing_env_at_use ~scrutinee_ty arm action (arms, dacc) =
...
+           parameters, which increase the work done by the typing env, as well
+           as the flow analysis. We then only do the lifting if the cost is
+           within the budget for the current function. *)
+        let lifting_budget = DA.get_continuation_lifting_budget dacc in
+        let lifting_cost =
+          DE.cost_of_lifting_continuations_out_of_current_one denv
+        in
+        (* is_lifting_allowed_by_budget ? *)
+        if not (lifting_budget > 0 && lifting_cost <= lifting_budget)
+        then dacc, cont, `Insuficient_lifting_budget
```
Ekdohibs:
```suggestion
        then dacc, cont, `Insufficient_lifting_budget
```

middle_end/flambda2/simplify/simplify_switch_expr.ml
```diff
@@ -667,6 +667,169 @@ let simplify_arm ~typing_env_at_use ~scrutinee_ty arm action (arms, dacc) =
...
+            | None -> `Not_enough_join_info
+            | Some join_analysis -> (
+              match
+                Join_analysis.simple_refined_at_join join_analysis
+                  (DE.typing_env denv) scrutinee
+              with
+              | Not_refined_at_join -> `Not_enough_join_info
+              | Invariant_in_all_uses _ ->
+                (* in this case, we don't need to specialize to know the
+                   scrutinee, or to simplify the switch, it will hapen without
```
Ekdohibs:
```suggestion
                   scrutinee, or to simplify the switch, it will happen without
```

middle_end/flambda2/simplify/simplify_switch_expr.ml
```diff
@@ -667,6 +667,169 @@ let simplify_arm ~typing_env_at_use ~scrutinee_ty arm action (arms, dacc) =
...
+                    var_analysis
+                    ( Apply_cont_rewrite_id.Set.empty,
+                      Apply_cont_rewrite_id.Set.empty )
+                in
+                match Apply_cont_rewrite_id.Set.cardinal generic with
+                | 0 | 1 -> `Spec (join_analysis, specialized, generic)
+                | _ ->
+                  if Apply_cont_rewrite_id.Set.is_empty specialized
+                  then `All_unknown
+                  else `Too_many_unknown_uses))
```
Ekdohibs, line 929: In the future, I'd like if we were able to specialize even in that case, by making a single continuation that is used in all unknown cases. However, that's outside the scope of this PR.

middle_end/flambda2/simplify/simplify_switch_expr.ml
```diff
@@ -667,6 +667,169 @@ let simplify_arm ~typing_env_at_use ~scrutinee_ty arm action (arms, dacc) =
...
+          in
+          match join_analysis_result with
+          | ( `No_reason_to_spec | `Too_many_unknown_uses | `All_unknown
+            | `Not_enough_join_info ) as res ->
+            dacc, cont, res
+          | `Spec (join_analysis, specialized, generic) ->
+            (* Specialization benefit estimation: we use heuristics similar to
+               that of inlining to estimate the benefit based on code size and
+               removed operations (note that we use the join info the the typing
```
Ekdohibs:
```suggestion
               removed operations (note that we use the join info in the typing
```
?

middle_end/flambda2/simplify/simplify_switch_expr.ml
```diff
@@ -667,6 +667,169 @@ let simplify_arm ~typing_env_at_use ~scrutinee_ty arm action (arms, dacc) =
     let arms = TI.Map.add arm (action, rewrite_id, arity, env_at_use) arms in
     arms, dacc
 
+let decide_continuation_specialization0 ~dacc ~switch ~scrutinee =
```
Ekdohibs, line 830: Could we go a little further and make `decide_continuation_specialization0` not return the continuation and `dacc`, and do that work outside this function (by changing the `Specialized` case to return the lifting cost as well)? This would allow completely splitting the logic of the decision from the actual modification of the `dacc`.

Ekdohibs (commented):
There are still 6 example files at toplevel, could we move them to the testsuite? (maybe checking the fexpr output, but at least just checking they compile correctly).

middle_end/flambda2/simplify/simplify_switch_expr.ml
Gbury: Agreed, we probably want to be able to do that, and we'll probably try and do it in a later PR.

middle_end/flambda2/simplify/simplify_switch_expr.ml
Gbury: That seems like a good idea, I'll do that.

Ekdohibs (approved):

driver/oxcaml_flags.ml
```diff
@@ -230,7 +233,7 @@ module Flambda2 = struct
 
   let o3 = {
     o2 with
-    function_result_types = Functors_only
+    function_result_types = Functors_only;
```
Ekdohibs:
```suggestion
    function_result_types = Functors_only
```
Stray change?

Ekdohibs (approved):

driver/oxcaml_flags.ml
Gbury: No idea, I'd guess it's an ocamlformat thing ?

Gbury: So apparently, ocamlformat keeps the original formatting for this very particular change, because reasons I guess ?

driver/oxcaml_flags.ml
```diff
@@ -328,9 +332,16 @@ module Flambda2 = struct
...
 
-    let o3 = default
+    let o3 = {
+      default with
+      cont_lifting_budget = 1_000;
+      (* in the worst case : 1_000 budget -> ~+18% compilation time *)
+      cont_spec_threshold = 0.;
+    }
 
     let o4 = default
```
bclement-ocp: This should be `let o4 = o3` I think.

Gbury: done (as well as slightly raising the lifting budget for O4)

PR #4375, opened 2025-07-24
Add profiling pass for speculative inlining

bclement-ocp:
Note: it looks like #4188 enabled nested speculative inlining with `DE.enter_set_of_closures denv Do_not_disable_inlining`, which may not have been intended.


bclement-ocp:
This will split other flambda2 profiling passes (e.g. join, make_unboxing_decisions) into the speculative and non-speculative case, but I think that's fine.

Gbury (approved):

Gbury:
Seems useful to better understand what happens. I'm in favour of merging, though the CI seems to need some fixing ?

bclement-ocp:
I think this was an unrelated error. Rebased, let's see if it works.

PR #4376, opened 2025-07-24
format everything in the tree

jvanburen:
We hadn't formatted everything earlier because of difficulties rebasing onto upstream. 
However, we can attempt to ocamlformat upstream before rebasing ourselves.

This PR ignores everything that can't be formatted with upstream with `.ocamlformat-ignore` files


jvanburen:
The first commit of this feature is manual changes, it's all that should need review.
The second commit is just the output of running `make fmt`

goldfirere:
I've wondered about doing this, too.

Others may know better, but I think it might be best to try this in the context of an upstream merge, just to make sure that we can still merge easily (enough). Also, I think there's a way of labeling a commit as blame-free, so that `git blame` can look through it. We should do that here.

testsuite/tests/typing-jkind-bounds/.ocamlformat-ignore
goldfirere, line 1: Why ignore all these test files? And is there a way to ignore a whole directory?

testsuite/tests/backtrace-multifiles/main.reference
goldfirere, line 1: Given that many tests weren't formatted, why these changes? Should more tests be excluded?

testsuite/tests/typing-jkind-bounds/.ocamlformat-ignore
jvanburen: they use features not supported by upstream ocamlformat

jvanburen: we can ignore a whole directory by just putting .ocamlformat with `disable=true`

stedolan:
I think this is a bad idea. Merges will become much, much harder.

Git does a good job of merging changes from upstream because it is currently able to do a three-way merge due to the shared history - it is not just comparing oxcaml vs trunk, it's comparing oxcaml vs. trunk vs. shared ancestor. Merging by formatting upstream and attempting to merge will not work this way, since there is no formatted shared ancestor to compare against.

lthls:
My experience with auto-formatting is that the gains in development speed are negative. I've lost a lot of time by dealing with formatting-related merge conflicts and formatting-related CI failures, and the gains to me are very limited.

PR #4378, opened 2025-07-24
Implement recursive continuations for JSIR translation

smjleo:
Works in a similar manner to recursive functions: we make fresh addresses for all the continuation symbols and use them later.


middle_end/flambda2/to_jsir/to_jsir.ml
```diff
@@ -124,8 +124,45 @@ and let_cont ~env ~res (e : Flambda.Let_cont_expr.t) =
...
+              env, res)
+            (env, res) domain
+        in
+        let env, res =
+          Continuation.Lmap.fold
+            (fun k handler (env, res) ->
+              Continuation_handler.pattern_match handler
+                ~f:(fun params ~handler:cont_body ->
+                  let params, env =
+                    To_jsir_shared.bound_parameters ~env
```
jvanburen, line 145: I think these parameters are being bound to every subsequent expression in the fold, am I right/is that intentional?

jvanburen (changes requested):

middle_end/flambda2/to_jsir/to_jsir.ml
smjleo: Yes good catch, thanks

smjleo: Actually a similar function `To_jsir_static_const.code` has the same issue. I'll change this in that PR too

jvanburen (approved):

PR #4381, opened 2025-07-25
Change module-based syntax to type-based syntax for generic optional arguments

zchenb:
Changes the syntax of generic optional arguments from `Module.Name.?'x:` to `(?x ...`

For to make the PR not so big, `Generic_optional` takes a `Longident.t` which indicates the constructor name. This needs to be updated in the future.

### DIFF
--- a/testsuite/tests/generic-optional-arguments/complex_pattern.ml
+++ b/testsuite/tests/generic-optional-arguments/complex_pattern.ml
@@ -1,0 +1,38 @@
+(* TEST
+flags = "-extension-universe alpha ";
+
+ expect;
+*)
+
+
+module type T = sig
+  val fst : (?x): (int * string) or_null -> unit -> int
+  val snd : (?x): (int * string) or_null -> unit -> string
+end
+
+module M : T = struct
+  let fst (?x:(((y, _) : int * string) = 5, "six") : _ or_null) () = y
+  let snd (?x:(((y, z) : int * string) = 5, "six") : _ or_null) () = z
+end
+
+[%%expect {|
+module type T =
+  sig
+    val fst : (?x):(int * string) or_null -> unit -> int
+    val snd : (?x):(int * string) or_null -> unit -> string
+  end
+module M : T @@ stateless
+|}]
+
+let v = M.fst ()
+
+[%%expect{|
+val v : int = 5
+|}]
+
+
+let v = M.snd ()
+
+[%%expect{|
+val v : string = "six"
+|}]
--- a/testsuite/tests/generic-optional-arguments/type_aliases.ml
+++ b/testsuite/tests/generic-optional-arguments/type_aliases.ml
@@ -4,6 +4,17 @@ flags = "-extension-universe alpha ";
  expect;
 *)
 
+
+module type T = sig
+  val g : (?x): int -> unit -> int
+end
+[%%expect {|
+Line 2, characters 16-19:
+2 |   val g : (?x): int -> unit -> int
+                    ^^^
+Error: Unknown generic optional argument type
+|}]
+
 (* CR: generic-optional: This should succeed *)
 type int_option = int option
 type int_or_null = int option


typing/btype.ml
```diff
@@ -628,22 +628,40 @@ let is_optional_parsetree : Parsetree.arg_label -> bool = function
...
   | Optional _ -> Optional_arg Stdlib_option
-  | Generic_optional(path, _) -> Optional_arg (classify_module_path path.txt)
+  | Generic_optional(path, _) -> Optional_arg (classify_module_path path)
   | Labelled _ | Position _ | Nolabel -> Required_or_position_arg
 
-let classify_optionality_parsetree (lbl : Parsetree.arg_label) = match lbl with
+let extract_optional_tp_from_pattern_constraint pat = match pat with
+  | {Parsetree.ppat_desc = Ppat_constraint (under_pat, ty, modes); _} ->
+    (match ty with
+    | Some {ptyp_desc = Ptyp_constr ({ txt = Lident ident_name}, [arg]); _} ->
```
dkalinichenko-js, line 648: This pattern-match can be merged with the previous by expanding `ty` correspondendgly.

typing/btype.ml
```diff
@@ -628,22 +628,40 @@ let is_optional_parsetree : Parsetree.arg_label -> bool = function
...
+        Some (ident_name, arg,
```
dkalinichenko-js: Here, I think we should convert the identifier into a proper path to handle code like `type 'a option = 'a not_option` . Feel free to leave a CR and do it in the next PR.

typing/ctype.ml
```diff
@@ -3312,19 +3312,10 @@ let rec expands_to_datatype env ty =
...
-  | Generic_optional(path1, str1), Generic_optional(path2, str2) ->
-      str1 = str2 &&
-        classify_module_path path1.txt = classify_module_path path2.txt
-  | Optional str1, Optional str2
-  | Position str1, Position str2
-  | Labelled str1, Labelled str2 -> str1 = str2
-  | Nolabel, Nolabel -> true
-  | (Generic_optional _ | Optional _ | Position _ | Labelled _ | Nolabel), _
-      -> false
+(* CR generic-optinoal: This function was needed because [Generic_optional]
```
dkalinichenko-js: I think we still need this function in the future if we want to handle type aliases like `type 'a my_option = 'a option` correctly.

typing/typecore.ml
```diff
@@ -7500,20 +7502,34 @@ and type_function
...
+                  (match pat.ppat_desc with
+                  | Ppat_constraint (_, Some sty, _) ->
+                      let gloc = { default.pexp_loc with loc_ghost = true } in
+                      Ast_helper.Exp.constraint_ default (Some sty) ~loc:gloc []
+                      , pat
+                  | _ -> default, pat)
+              | Generic_optional _ ->
+                  (match extract_optional_tp_from_pattern_constraint pat with
+                  | None ->
+                      failwith
```
dkalinichenko-js: `failwith` is ok here for now, but I think we should raise a proper error if the type is not `something option_like`.

dkalinichenko-js: If this error is impossible at this stage, we should change the return type of `extract_optional_tp_from_pattern_constraint`.

dkalinichenko-js (commented):
Looks great, thank you!

I'd like some tests for when there's no type annotation, some tests for when the annotated type is not a type constructor (including something like `type int_option = int option`), and some more tests for edge cases when generic optional syntax is combined with pattern-matching, like what was discussed in the meeting.

typing/ctype.ml
dkalinichenko-js: Also, typo in the CR.

typing/btype.ml
dkalinichenko-js: Feel free to just leave this alone if my suggestion makes the code too awkward though.

typing/btype.ml
zchenb: I would like to point out that folding ty makes `type int_option = int option` somehow problematic to handle. I have added CRs to more explicit error messages.

typing/ctype.ml
zchenb: Left a CR.

typing/btype.ml
zchenb: Left a CR.

typing/ctype.ml
zchenb: I see. My plan is to eventually remove `mpath` from `Generic_optional`. The `mpath` is in the arg_label is an unfortunate consequence of previous design. But I do not think that is the best way of implementing the current design. I left a CR anyways to investigate.

dkalinichenko-js:
Oh, nevermind, we can't drop the type annotation in the interface completely. At most, we can do something like `val f : (?x): int -> unit -> int`.

typing/btype.ml
dkalinichenko-js: Ack.

typing/btype.ml
```diff
@@ -628,22 +628,43 @@ let is_optional_parsetree : Parsetree.arg_label -> bool = function
...
+    (match ty with
+    (* CR generic-optional:  convert the identifier into a proper path to
+       handle code like type 'a option = 'a not_option *)
+    | Some {ptyp_desc = Ptyp_constr ({ txt = Lident ident_name}, [arg]); _} ->
+        (ident_name, arg,
+        {pat with
+          Parsetree.ppat_desc = Ppat_constraint (under_pat, Some arg, modes)})
+    (* CR generic-optional: handle the case for type int_option = int option *)
+    | Some _  -> failwith "Expected optional type"
+    | None  -> failwith "Missing type annotation"
```
dkalinichenko-js, line 654: Is this error ever raised? If it is, could you add a test for it?

typing/btype.ml
```diff
@@ -628,22 +628,43 @@ let is_optional_parsetree : Parsetree.arg_label -> bool = function
...
+  | {Parsetree.ppat_desc = Ppat_constraint (under_pat, ty, modes); _} ->
+    (match ty with
+    (* CR generic-optional:  convert the identifier into a proper path to
+       handle code like type 'a option = 'a not_option *)
+    | Some {ptyp_desc = Ptyp_constr ({ txt = Lident ident_name}, [arg]); _} ->
+        (ident_name, arg,
+        {pat with
+          Parsetree.ppat_desc = Ppat_constraint (under_pat, Some arg, modes)})
+    (* CR generic-optional: handle the case for type int_option = int option *)
+    | Some _  -> failwith "Expected optional type"
```
dkalinichenko-js, line 653: Same here.

typing/btype.ml
zchenb: This error is never raised. Instead, all cases that may belong here have been ruled out during `transl_label`, which raises `Unknown_generic_optional_argument_type`. Most tests hit that case.

typing/btype.ml
zchenb: Same as above

zchenb:
> Re: no type annotations, I also meant testing that case in the interface, at least to document the error we're raising in this case. For non-trivial pattern-matches, I meant examples like (?x:((y, _) : int * string = 5, "six") : _ option) we discussed in the meeting (or have I missed the test for this?).

No type annotations have been added in `type_aliases.ml`. I added this specific example to `complex_pattern.ml`. But pattern matching is already present in module `M_15` in `syntax_comparison.ml`.

typing/btype.ml
dkalinichenko-js: I see. We usually use `Misc.fatal_error` (or `assert false`) to indicate an error case in the compiler that should never be reached. Since you're changing this anyway later, feel free to just keep `failwith` for now, though.

PR #4382, opened 2025-07-25
Translate switch expressions to JSIR

smjleo:
Now
```
let f x =
  match x with
  | -100 -> 0
  | 0 -> 1
  | 100 -> 2
  | _ -> 3
;;
```
gets translated to
```
==== 0 () ====
  v10 = fun(v2){1 ()}
  v11 = imm{tag=0; 0 = v10}
  v12 = "caml_register_global"(0, v11, "Switch_simple"j)
  stop

==== 1 () ====
  v3 = !(v2 === -100)
  v9 = CONST{0}
  v14 = 0 === v3
  if v14 then 7 () else 2 ()

==== 2 () ====
  v4 = !(v2 === 0)
  v8 = CONST{1}
  v15 = 0 === v4
  if v15 then 6 () else 3 ()

==== 3 () ====
  v5 = !(v2 === 100)
  v6 = CONST{2}
  v7 = CONST{3}
  v16 = 0 === v5
  if v16 then 4 () else 5 ()

==== 4 () ====
  return v6

==== 5 () ====
  return v7

==== 6 () ====
  return v8

==== 7 () ====
  return v9
  ```

### CODE REVIEW

File: middle_end/flambda2/to_jsir/to_jsir.ml [hunk 1]
```diff
@@ -190,74 +190,110 @@ and apply_expr ~env ~res e =
...
+          let res, addr = To_jsir_result.new_block res ~params:[] in
+          let res = To_jsir_result.end_block_with_last_exn res last in
+          res, (addr, []))
+      res
+      (Array.init (max + 1) Targetint_31_63.of_int)
+  in
+  let last : Jsir.last =
+    match Array.length arms with
+    | 2 ->
+      (* Special case: turn it into if/then/else *)
```
jvanburen, line 292: is this always guaranteed to be 0 and 1?

jvanburen: maybe we could add an assert and a comment if so

middle_end/flambda2/to_jsir/to_jsir_result.ml
```diff
@@ -104,3 +99,15 @@ let to_program_exn
        used";
   let free_pc = (Jsir.Addr.Map.max_binding complete_blocks |> fst) + 1 in
   { Jsir.start = Jsir.Addr.zero; blocks = complete_blocks; free_pc }
+
+let invalid_switch_block t =
```
jvanburen, line 103: what's the benefit of making this one shared block instead of emitting the extern call inline everywhere?

middle_end/flambda2/to_jsir/to_jsir.ml
smjleo: Sorry I'm not sure I follow, it's an array so if it has length 2 then indices 0 and 1 should be populated (we're matching on the Jsir arms not the flambda ones)

middle_end/flambda2/to_jsir/to_jsir_result.ml
smjleo: We need a new block since switch requires us to specifically jump to a block (so can't inline in the same block), and if we need separate blocks for this anyway then I thought it'd probably be better to have one shared one to avoid code bloat for something we don't expect to trigger

smjleo: Though it's entirely possible jsoo does CSE to get rid of them anyway

jvanburen (approved):

PR #4383, opened 2025-07-25
Add DWARF debugging information compression support

mshinwell:
## Summary
- Adds `-gdwarf-compression=VALUE` flag to ocamlopt (default: zlib)
- Passes compression options to assembler and linker when DWARF info is generated
- Compression is only applied when either `-gdwarf-inlined-frames` or `-gno-upstream-dwarf` are passed

## Implementation Details
The implementation uses the `-gz=` flag which is portable between GCC and Clang, unlike `--compress-debug-sections`. See: https://maskray.me/blog/2022-01-23-compressed-debug-sections

The compression flag is passed to the C toolchain in three scenarios:
1. When assembling .s files
2. When compiling C files from command line  
3. During linking

Without `-gdwarf-inlined-frames` or `-gno-upstream-dwarf`, no compression flag is passed to maintain compatibility with upstream behaviour.

## Testing
Manual testing has been performed to verify the flag is correctly passed in all scenarios. No ocamltest tests are included as they are difficult to set up for this feature.


backend/debug/dwarf/dwarf_flags/dwarf_flags.mli
```diff
@@ -74,3 +74,14 @@ val gdwarf_may_alter_codegen : bool ref
     [restrict_to_upstream_dwarf] is set to [false], although that implies
     this variable being set to [true]). *)
 val dwarf_inlined_frames : bool ref
+
+val default_gdwarf_compression : string
+
+val gdwarf_compression : string ref
+
+(** Get the DWARF compression flag to pass to the C toolchain.
+    Returns a flag of the form "-gz=<compression>".
```
spiessimon:
```suggestion
    Returns a flag of the form " -gz=<compression>".
```
The space here is actually quite important, because we do not add one at some usage sites. I'm wondering whether we actually should and remove it here.

driver/optmaindriver.ml
```diff
@@ -61,6 +61,9 @@ let main unix argv ppf ~flambda2 =
     Clflags.Opt_flag_handler.set Oxcaml_flags.opt_flag_handler;
     Compenv.parse_arguments (ref argv) Compenv.anonymous "ocamlopt";
     Compmisc.read_clflags_from_env ();
+    (* Set up DWARF compression for C compiler invocations *)
+    if !Clflags.debug && !Clflags.native_code then
+      Clflags.dwarf_c_toolchain_flag := Dwarf_flags.get_dwarf_c_toolchain_flag ();
```
spiessimon: Why is this a flag? It doesn't seem like one can set it from the command line. I guess because it is set lazily? A bit sneaky. I guess this is forwarding from the `Dwarf_flags` to the `Clflags`. Do we need it in `Clflags` because of utils/ccomp.ml?

backend/debug/dwarf/dwarf_flags/dwarf_flags.ml
```diff
@@ -172,3 +172,18 @@ let gdwarf_self_tail_calls = ref default_gdwarf_self_tail_calls
...
+
+let gdwarf_compression = ref default_gdwarf_compression
+
+let get_dwarf_compression_flag () =
+  if !dwarf_inlined_frames || not !restrict_to_upstream_dwarf then
+    Some !gdwarf_compression
+  else
+    None
+
+let get_dwarf_c_toolchain_flag () =
```
spiessimon, line 185: I think this should mention compression somewhere in the name, or are we planning to use this for other DWARF flags as well?

backend/arm64/proc.ml
```diff
@@ -443,8 +443,15 @@ let slot_offset (loc : Reg.stack_location) ~stack_class ~stack_offset
...
 let assemble_file infile outfile =
+  let dwarf_flag =
+    if !Clflags.native_code && !Clflags.debug then
+      Dwarf_flags.get_dwarf_c_toolchain_flag ()
+    else
+      ""
+  in
   Ccomp.command (Config.asm ^ " " ^
                  (String.concat " " (Misc.debug_prefix_map_flags ())) ^
+                 dwarf_flag ^
```
spiessimon, line 456: I'm a bit surprised we don't provide the assembler with the `-g` flag, but it seems we also didn't before.

backend/x86_proc.ml
```diff
@@ -370,9 +370,16 @@ let compile infile outfile =
...
+    let dwarf_flag =
+      if !Clflags.native_code && !Clflags.debug then
+        Dwarf_flags.get_dwarf_c_toolchain_flag ()
+      else
+        ""
+    in
     Ccomp.command
       (Config.asm ^ " "
       ^ String.concat " " (Misc.debug_prefix_map_flags ())
+      ^ dwarf_flag
```
spiessimon: This is where we do not manually add a space and it is important that `get_dwarf_c_toolchain_flag` is prefixed by a space if not empty.

utils/ccomp.ml
```diff
@@ -93,6 +93,12 @@ let compile_file ?output ?(opt="") ?stable_name name =
     | Some stable when Config.c_has_debug_prefix_map ->
       Printf.sprintf " -fdebug-prefix-map=%s=%s" name stable
     | Some _ | None -> "" in
+  let debug_flags =
+    if !Clflags.debug && Config.ccomp_type <> "msvc" then
+      "-g" ^ !Clflags.dwarf_c_toolchain_flag
```
spiessimon: Is this necessarily called after `optmaindriver` or is there another way to call this?

driver/optmaindriver.ml
mshinwell: Yeah, we can fix all this by merging the flags files together as discussed, I hope.

backend/debug/dwarf/dwarf_flags/dwarf_flags.ml
mshinwell: Potentially yes

xclerc:
The linked issue is basically what I was
worried about, I remember having
reproducibility issues with ztd / tar+gzip.
If memory serves, it was not only
about the versions, zstd output was
at some point non deterministic.

backend/debug/dwarf/dwarf_flags/dwarf_flags.mli
```diff
@@ -74,3 +74,20 @@ val gdwarf_may_alter_codegen : bool ref
...
 val dwarf_inlined_frames : bool ref
+
+val default_gdwarf_compression : string
+
+val gdwarf_compression : string ref
+
+(** Get the DWARF compression flag to pass to the C toolchain.
+    Returns a flag of the form " -gz=<compression>" (note the leading space).
+    Returns an empty string if compression is not enabled.
+    Note: The -gz= option is common between GCC and Clang, unlike
```
spiessimon, line 85: Now that the flags can be configured, I think this comment should be updated. In particular, the note should be removed and the reference, since it does not matter anymore which flags the C toolchain accepts.

driver/oxcaml_args.ml
```diff
@@ -844,6 +844,18 @@ let mk_gdwarf_max_function_complexity f =
       \     will not be emitted, to improve compilation time (default %d)"
       !Dwarf_flags.dwarf_max_function_complexity )
 
+let mk_gdwarf_compression f =
+  ( "-gdwarf-compression",
+    Arg.String f,
+    Format.sprintf " Set the DWARF compression format (default %s)"
+      !Dwarf_flags.gdwarf_compression )
+
+let mk_gdwarf_compression f =
```
spiessimon: Is this function duplicated?

utils/ccomp.ml
mshinwell: See latest commit

backend/debug/dwarf/dwarf_flags/dwarf_flags.mli
mshinwell: Would prefer to leave this for our own reference

PR #4385, opened 2025-07-28
Ignore alloc mode when comparing types

bclement-ocp:
This is a source of spurious differences between the old and new join (the new join being better about tracking this).

### DIFF
--- a/middle_end/flambda2/types/equal_types_for_debug.ml
+++ b/middle_end/flambda2/types/equal_types_for_debug.ml
@@ -19,6 +19,10 @@ module TG = Type_grammar
 module ET = Expand_head.Expanded_type
 module ME = Meet_env
 
+type config = { ignore_alloc_mode : bool }
+
+let create_config ?(ignore_alloc_mode = true) () = { ignore_alloc_mode }
+
 type renaming =
   { mutable left_renaming : Variable.t Variable.Map.t;
     mutable right_renaming : Variable.t Variable.Map.t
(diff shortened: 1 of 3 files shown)


bclement-ocp:
The old join does not seem to track allocation mode in variants with constant constructors (to be fair I did not investigate why, it might be easy to fix).

```ocaml
(* amode.ml *)
let f b x c f1 f2 =
    let[@local] k o = if c then f1 o else f2 o in
    if b then k None else k (Some x)
```

```console
$ ./_build/_bootinstall/bin/ocamlopt.opt -nostdlib -nopervasives -O3 -c -flambda2-join-algorithm=checked amode.ml
====================== Distinct joins ======================
 -- Old join --
  ((equations
    (o/31UV :
      (Val
       ( ((Variant
           (blocks
            ((alloc_mode Heap_or_local) (known
              {(tag_0 => (Known 1), ((Val ((= x/26UV)))))}) (other Bottom)))
           (tagged_imms (Naked_immediate ((= #0))))))!))
     b/25UV :
      (Val
       ( ((Variant (blocks ⊥) (tagged_imms (Naked_immediate (({ 0 1 }))))))!))))))
 -- New join --
  ((defined_vars { join_var/36N })
   (equations
    (join_var/36N : (Naked_immediate (({ 0 1 })))
     o/31UV :
      (Val
       ( ((Variant
           (blocks
            ((alloc_mode Heap) (known
              {(tag_0 => (Known 1), ((Val ((= x/26UV)))))}) (other Bottom)))
           (tagged_imms (Naked_immediate ((= #0))))))!))
     b/25UV :
      (Val
       ( ((Variant (blocks ⊥)
           (tagged_imms (Naked_immediate ((= join_var/36N))))))!))))))
 Names with distinct types: { o/31UV }
============================================================
```

bclement-ocp:
Independently of fixing the old join or not, I think having a way to configure how loose we want to be when comparing types is useful.

lthls:
I agree about configuration of the comparison.
I looked at the reason for the difference, and it looks like the n-way join filters out bottom branches in `n_way_join_row_like_for_blocks` while the old join doesn't, and bottom blocks have their alloc mode set to Top instead of Bottom (or `Heap_or_local` instead of `Heap`).
So we could fix the old join if we wanted to (although the change would be in `Type_grammar`, not in the join itself), but I think we should merge this PR anyway, and maybe change the default config back to not ignoring alloc modes when (or if) we fix the mismatch.

PR #4386, opened 2025-07-28
Just call meet_naked_immediates in reify

bclement-ocp:
Instead of duplicating it.

Note that this introduces a subtle difference: if we are reifying a naked immediate variable that is the `Get_tag` of a value, we now call `meet_get_tag` on that value instead of `prove_get_tag`.

Semantically, this is OK: we can only use the `Get_tag` of a value in contexts where the value is known to be a block.

In practice, this should also be fine, because the only place where a `Get_tag` variable can be introduced without being in the scope of a `%get_tag` is by unboxing, and such variables can only be used through the CSE of a `%get_tag` primitive.


lthls (approved):
We discussed it offline before, and I agree that it's simpler and better this way.

PR #4390, opened 2025-07-28
flambda2-types: Relations are projections

bclement-ocp:
This patch stores relations (specifically, `Is_null`, `Is_int` and `Get_tag`) as fields of the corresponding types of `value` kind.

This enables tracking them precisely during the (n-way) join, which is a requirement of the match-in-match heuristic.

Note that we do do any forward propagation: instead, all propagations are still backward from the naked immediates. This is to ensure we do not introduce loops during the meet, although it could mean we lose aliases in some rare cases where we have two distinct variables for the tags, likely due to env extension shenanigans.

### DIFF
--- a/middle_end/flambda2/types/meet_and_n_way_join.ml
+++ b/middle_end/flambda2/types/meet_and_n_way_join.ml
@@ -2081,7 +2081,7 @@ and n_way_join_expanded_head env kind (expandeds : ET.t Join_env.join_arg list)
 and n_way_join_head_of_kind_value env
     (heads : TG.head_of_kind_value Join_env.join_arg list) :
     TG.head_of_kind_value n_way_join_result =
-  let (is_null : TG.is_null), env =
+  let is_null, env =
     let machine_width = Join_env.machine_width env in
     let exception Cannot_track in
     match
@@ -2093,12 +2093,12 @@ and n_way_join_head_of_kind_value env
           | Maybe_null { is_null = Some is_null } -> id, Simple.var is_null)
         heads
     with
-    | exception Cannot_track -> Maybe_null { is_null = None }, env
+    | exception Cannot_track -> TG.Maybe_null { is_null = None }, env
     | is_null_simples -> (
       (* Note: we ideally would use [n_way_join_relation_simples] here, but we
          need to store a [Not_null] constructor if the join is [false]. *)
       match n_way_join_simples env K.naked_immediate is_null_simples with
-      | Bottom, env -> Maybe_null { is_null = None }, env
+      | Bottom, env -> TG.Maybe_null { is_null = None }, env
       | Ok simple, env ->
         let is_null =
           Simple.pattern_match' simple


middle_end/flambda2/types/meet_and_join.ml
```diff
@@ -911,22 +932,60 @@ and meet_array_contents env (array_contents1 : TG.array_contents Or_unknown.t)
...
+              (* Distinct constants: this is a bottom result *)
+              Bottom (New_result ()))
+            ~name:(fun _ ~coercion:_ : _ meet_result -> Ok (Right_input, env)))
+        ~name:(fun _ ~coercion:_ ->
+          Simple.pattern_match simple2
+            ~const:(fun _ : _ meet_result -> Ok (Left_input, env))
+            ~name:(fun _ ~coercion:_ ->
+              (* CR bclement: We want to add an equality here, but it can
+                 currently cause loops due to the reductions that could be
+                 stored on the relation variables. *)
```
bclement-ocp: I was quite happy to notice that not adding this equality seemed to work and prevent infinite loops without any other infrastructure, but actually we need to add it for some of the match-in-match examples (: 

I will resurrect the `Meet_env` PR that prevents loops.

bclement-ocp: It looks like #4394 *might* be fixing the issue (at least, I can build the compiler without stack overflow when adding an equality here), but it could also just be making it less rare.

bclement-ocp: #4397 

Although I'm thinking maybe we want to guard adding an equality between names behind a --use-match-in-match flag, potentially.

bclement-ocp:
Rebased.

Note: I think this PR is in a reviewable state, although we will need to discuss what to do with the reductions.

middle_end/flambda2/types/grammar/type_grammar.ml
```diff
@@ -45,6 +45,7 @@ end
 type is_null =
   | Not_null
   | Maybe_null
+  | Is_null of Variable.t
```
lthls: The `Maybe_null` and `Is_null _` cases should be merged. I'm not sure which name is best (the name of the type itself, `is_null`, is not very good either), but it should be more obvious that the semantics of `Is_null` is the same as `Maybe_null`.
Given that the variant case below has `Variable.t option` for the `is_int` and `get_tag` fields, it would be natural to use `Maybe_null of { is_null : Variable.t option }`.

middle_end/flambda2/types/grammar/type_grammar.ml
```diff
@@ -1373,13 +1403,21 @@ and ids_for_export_head_of_kind_value { non_null; is_null = _ } =
   | Unknown | Bottom -> Ids_for_export.empty
   | Ok non_null -> ids_for_export_head_of_kind_value_non_null non_null
```
lthls: You need to add the variable for `Is_null`

middle_end/flambda2/types/grammar/type_grammar.ml
```diff
@@ -363,15 +366,23 @@ and free_names_head_of_kind_value0 ~follow_value_slots { non_null; is_null = _ }
   | Ok non_null ->
     free_names_head_of_kind_value_non_null ~follow_value_slots non_null
```
lthls: You need to add the variable for `Is_null`

middle_end/flambda2/types/grammar/type_grammar.ml
```diff
@@ -2068,10 +2106,32 @@ and remove_unused_value_slots_and_shortcut_aliases_head_of_kind_value head
     then head
     else { non_null = Ok non_null'; is_null = head.is_null }
```
lthls: You need to handle variables in the `Is_null _` case

middle_end/flambda2/types/grammar/type_grammar.ml
```diff
@@ -2785,9 +2847,27 @@ and project_head_of_kind_value ~to_project ~expand head =
...
```
lthls: You need to handle the variable in `Is_null _`

middle_end/flambda2/types/meet_and_join.ml
lthls: #4397 is now merged, so we can consider adding the equality here.

middle_end/flambda2/types/meet_and_join.ml
```diff
@@ -698,23 +708,34 @@ and meet_head_of_kind_value_non_null env
...
+          is_int = is_int2;
+          get_tag = get_tag2
         } ) ->
     (* Uniqueness tracks whether duplication/lifting is allowed. It must always
        be propagated, both for meet and join. *)
     let is_unique = is_unique1 || is_unique2 in
-    map_result
-      ~f:(fun (blocks, immediates, extensions) ->
+    combine_results2 env
+      ~rebuild:(fun is_int (get_tag, blocks, immediates, extensions) ->
```
lthls: It feels a bit strange to leave `is_int` out of `meet_variant`. I understand why we want to handle `is_int` and `get_tag` at different points, but I'm wondering if pushing this `combine_results2` into `meet_variant` would make the code easier to understand.

middle_end/flambda2/types/meet_and_join.ml
```diff
@@ -943,17 +1004,17 @@ and meet_variant env ~(blocks1 : TG.Row_like_for_blocks.t Or_unknown.t)
...
         | Ext { when_a = when_immediate; when_b = when_block } ->
           Ext { when_immediate; when_block }
       in
-      blocks, imms, extensions)
+      get_tag, blocks, imms, extensions)
     (meet_disjunction ~meet_a ~meet_b ~bottom_a ~bottom_b ~meet_type
-       ~join_env_extension env imms1 blocks1 extensions1 imms2 blocks2
-       extensions2)
+       ~join_env_extension env imms1 (get_tag1, blocks1) extensions1 imms2
+       (get_tag2, blocks2) extensions2)
```
lthls: I think we should add a comment explaining why the `get_tag` meet has to be under `meet_disjunction`.

middle_end/flambda2/types/meet_and_join.ml
```diff
@@ -1785,13 +1846,23 @@ and join_head_of_kind_value env (head1 : TG.head_of_kind_value)
   in
   let is_null : TG.is_null =
     match head1.is_null, head2.is_null with
-    | Maybe_null, _ | _, Maybe_null -> Maybe_null
+    | Is_null var1, Is_null var2 when Variable.equal var1 var2 ->
+      (* CR bclement: Look for a shared alias. *)
+      Is_null var1
```
lthls: If you follow my suggestion to have a single `Maybe_null of { is_null : Variable.t option }` case then this can call `join_relation`.

middle_end/flambda2/types/meet_and_n_way_join.ml
```diff
@@ -1944,12 +2066,15 @@ and n_way_join_head_of_kind_value env
     match heads with
     | [] -> Not_null
     | (_, head1) :: heads -> (
+      (* CR bclement: preserve [is_null] relation *)
```
lthls: I think this should be done (using `n_way_join_relatino_simples` if possible)

middle_end/flambda2/types/grammar/type_grammar.ml
bclement-ocp: Placeholder for GitHub

middle_end/flambda2/types/grammar/type_grammar.ml
bclement-ocp: I agree -- I don't remember why I wanted to keep the two constructors but it's clearly misguided, confusing and error-prone (easier to miss cases). I'll use the definition you suggest.

middle_end/flambda2/types/grammar/type_grammar.ml
bclement-ocp: Good catch, done.

middle_end/flambda2/types/grammar/type_grammar.ml
bclement-ocp: Done.

middle_end/flambda2/types/grammar/type_grammar.ml
bclement-ocp: Done.

middle_end/flambda2/types/grammar/type_grammar.ml
bclement-ocp: Done.

middle_end/flambda2/types/meet_and_n_way_join.ml
bclement-ocp: Done, but not using `n_way_join_relation_simples`. It is a bit tricky because in the case of `is_int` / `get_tag` variables the constant information is replicated elsewhere (either in the `immediates` / `blocks` fields of `head_of_kind_value_non_null`, or the tags themselves), but for `is_null` we must preserve it.

middle_end/flambda2/types/meet_and_join.ml
bclement-ocp: Done.

middle_end/flambda2/types/meet_and_join.ml
bclement-ocp: Thought I had done so, fixed.

middle_end/flambda2/types/meet_and_join.ml
bclement-ocp: Done.

middle_end/flambda2/types/meet_and_join.ml
```diff
@@ -676,9 +676,16 @@ and meet_head_of_kind_value env
...
     | Not_null, Not_null -> Bottom Both_inputs
-    | Maybe_null, Maybe_null -> Ok (Both_inputs, env)
-    | Not_null, Maybe_null -> Bottom Left_input
-    | Maybe_null, Not_null -> Bottom Right_input
+    | Not_null, Maybe_null _ ->
+      (* We rely on the inverse relations for reductions. *)
+      Ok (Right_input, env)
+    | Maybe_null _, Not_null ->
+      (* We rely on the inverse relations for reductions. *)
+      Ok (Left_input, env)
```
lthls: This is a change from the previous version (which returned `Bottom _` in those two cases). Is that intended ?
The two comments were attached to the `Maybe_null, Is_null _` cases in the previous version, and those cases are now part of the last case and handled by `meet_relation`.

middle_end/flambda2/types/meet_and_n_way_join.ml
```diff
@@ -818,19 +818,16 @@ and meet_head_of_kind_value env
...
     | Not_null, Not_null -> Bottom Both_inputs
-    | Maybe_null, Maybe_null -> Ok (Both_inputs, env)
-    | Maybe_null, Is_null _ ->
+    | Not_null, Maybe_null _ ->
       (* We rely on the inverse relations for reductions. *)
       Ok (Right_input, env)
-    | Is_null _, Maybe_null ->
+    | Maybe_null _, Not_null ->
       (* We rely on the inverse relations for reductions. *)
       Ok (Left_input, env)
```
lthls: (Same comment as in `meet_and_join.ml`)

middle_end/flambda2/types/meet_and_join.ml
bclement-ocp: No, that's me getting confused because `Not_null` ~~ `Bottom`.

This reminds me that I was thinking that maybe we want to define `type is_null = Variable.t option Or_bottom.t`, or maybe even `type is_null = Variable.t Or_unknown_or_bottom` (where `Bottom` means "not null", `Unknown` means "maybe null", and `Ok var` means "maybe null ; `var` holds the result of `%is_null`) — it would make code such as this less confusing IMO.

bclement-ocp: This is the source of the change to the output of `local.flt`.  But it is still surprising that we generate worse code than we should.

(Good catch by the way, thanks!)

middle_end/flambda2/types/meet_and_join.ml
```diff
@@ -949,10 +949,17 @@ and meet_relation env var1 var2 =
...
+              (* Note: This equality can (rarely?) cause loops due to reductions
+                 that could be stored on the relation variables (if the relation
+                 variable has a [Is_int] / [Get_tag] / [Is_null] type). This is
+                 caught by the safeguard in [Meet_env]. *)
+              match
+                add_equation simple1
+                  (TG.alias_type_of K.naked_immediate simple2)
+                  env ~meet_type
+              with
+              | Ok (_, env) -> Ok (Both_inputs, env)
```
lthls, line 961: If I'm not mistaken, this could be `Ok ((), env) ->`

lthls: (Although it seems that there are other occurrences of this pattern in this file, so maybe it's not a big deal)

bclement-ocp: Do you mean it could be `Ok (New_result (), env)`? I think it needs to be `Both_inputs` for the stopping criterion to work (so that we just learn the alias without needlessly updating the typing env).

lthls: Sorry, I forgot that the payload of `Ok` was `'a meet_return_value * ME.t` and not `'a * ME.t`. So the code is fine.

lthls: (I'm always a bit suspicious when part of a pattern is ignored)

bclement-ocp: To be fair, the result of `add_equation` seems to always be ignored, I'm not sure why it doesn't just return a `ME.t Or_bottom.t` (I can make a follow-up PR to clean it up).

bclement-ocp: (We never use any of the `meet_result` combinators on the result of `add_equation`, we just match it directly)

PR #4394, opened 2025-07-28
Improved meet for naked immediates

bclement-ocp:
This is one part of #4112, with information stored in the types in the style of #2259 (which it replaces).

One crucial difference with #2259 is that during the meet, we only reduce new information. For instance, if we are meeting:

  (Naked_immediate {0, 1, 2} (Get_tag x))

and

  (Naked_immediate {1, 2} (Get_tag y))

then we will propagate:

 - Nothing if `x` and `y` have the same canonical;

 - The set of tags `{1, 2}` on `x` otherwise.

This provides a way to solve potential reduction loops with forward reductions from the value itself: if the set of possible tags of `x` shrinks to a new set `S`, and we know that `get_tag(x) = get_tag_x`, then we can add the equation

	get_tag_x : (Naked_immediate S (Get_tag x))

If we already know this set of possible values for `get_tag_x`, then nothing happens. If we know (or learn through intersection with `S`) a more precise set of values for `get_tag_x`, the reduction on `get_tag_x` will add a type on `x` with tags `S'` and `get_tag` variable `get_tag_x`. Since `S'` is a subset of `S`, we will then not propagate things back onto `get_tag_x`.

### DIFF
--- a/middle_end/flambda2/types/env/join_env.ml
+++ b/middle_end/flambda2/types/env/join_env.ml
@@ -1553,7 +1553,8 @@ let rec add_inverse_relation_to_env_extension ?(seen = Name.Set.empty)
   let empty_descr : TG.Head_of_kind_naked_immediate.descr =
     { naked_immediates = Unknown; inverse_relations = TG.Relation.Map.empty }
   in
-  let[@inline] type_from_descr (descr : TG.Head_of_kind_naked_immediate.descr) =
+  let[@inline] updated_type_from_descr
+      (descr : TG.Head_of_kind_naked_immediate.descr) =
     let inverse_relations =
       TG.Relation.Map.update relation
         (function
@@ -1567,7 +1568,8 @@ let rec add_inverse_relation_to_env_extension ?(seen = Name.Set.empty)
   in
   match Name.Map.find_opt name (TEE.to_map env_extension) with
   | None ->
-    TEE.add_or_replace_equation env_extension name (type_from_descr empty_descr)
+    TEE.add_or_replace_equation env_extension name
+      (updated_type_from_descr empty_descr)
   | Some existing_ty -> (
     match TG.descr existing_ty with
     | Naked_immediate Bottom ->
@@ -1578,12 +1580,13 @@ let rec add_inverse_relation_to_env_extension ?(seen = Name.Set.empty)
       (* This should not happen, as we would usually only only store non-obvious
          types in extensions -- but it's also harmless. *)
       TEE.add_or_replace_equation env_extension name
-        (type_from_descr empty_descr)
+        (updated_type_from_descr empty_descr)
     | Naked_immediate (Ok (No_alias head)) ->
       (* There is a concrete type for this name in the extension; augment it
          with the reverse relation. *)
       let descr = TG.Head_of_kind_naked_immediate.descr head in
-      TEE.add_or_replace_equation env_extension name (type_from_descr descr)
+      TEE.add_or_replace_equation env_extension name
+        (updated_type_from_descr descr)
     | Naked_immediate (Ok (Equals simple)) ->
       (* Usually we expect that the name we are adding an alias for would be
          canonical in the env extension, but it could (rarely) happen that it is
@@ -1598,7 +1601,7 @@ let rec add_inverse_relation_to_env_extension ?(seen = Name.Set.empty)
                break the loop to store the non-alias type anywhere, so we might
                as well do it when we detect the loop. *)
             TEE.add_or_replace_equation env_extension name
-              (type_from_descr empty_descr)
+              (updated_type_from_descr empty_descr)
           else
             add_inverse_relation_to_env_extension ~seen:(Name.Set.add name seen)
               env_extension name' relation ~scrutinee)
--- a/middle_end/flambda2/types/env/join_env.ml
+++ b/middle_end/flambda2/types/env/join_env.ml
@@ -1548,10 +1548,174 @@ let join_aliases_into_bindings ~joined_envs ~bindings equations_to_join =
           in
           equations_to_join, bindings))
 
+let rec add_inverse_relation_to_env_extension ?(seen = Name.Set.empty)
+    env_extension name relation ~scrutinee =
+  let empty_descr : TG.Head_of_kind_naked_immediate.descr =
+    { naked_immediates = Unknown; inverse_relations = TG.Relation.Map.empty }
+  in
+  let[@inline] type_from_descr (descr : TG.Head_of_kind_naked_immediate.descr) =
+    let inverse_relations =
+      TG.Relation.Map.update relation
+        (function
+          | None -> Some (Name.Set.singleton scrutinee)
+          | Some existing_args -> Some (Name.Set.add scrutinee existing_args))
+        descr.inverse_relations
+    in
+    TG.create_from_head_naked_immediate
+      (TG.Head_of_kind_naked_immediate.from_descr_non_empty
+         { descr with inverse_relations })
+  in
+  match Name.Map.find_opt name (TEE.to_map env_extension) with
+  | None ->
+    TEE.add_or_replace_equation env_extension name (type_from_descr empty_descr)
+  | Some existing_ty -> (
+    match TG.descr existing_ty with
+    | Naked_immediate Bottom ->
+      (* If we already know that we are bottom, we don't need to store anything
+         more precise. *)
+      env_extension
+    | Naked_immediate Unknown ->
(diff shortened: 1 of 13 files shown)


middle_end/flambda2/types/grammar/type_grammar.ml
```diff
@@ -46,6 +46,88 @@ type is_null =
...
+  let is_null = 0
+
+  let is_int = 1
+
+  let get_tag = 2
+
+  type descr =
+    | Is_null
+    | Is_int
+    | Get_tag
```
lthls, line 84: I'm assuming that not using `descr` as the main type is required to use patricia tres for maps and sets ?

middle_end/flambda2/types/grammar/type_grammar.ml
```diff
@@ -3005,18 +3160,51 @@ and project_head_of_kind_value_non_null ~to_project ~expand head =
...
+            Name.pattern_match name
+              ~symbol:(fun _ -> Name.Set.add name names)
+              ~var:(fun var ->
+                if not (Variable.Set.mem var to_project)
+                then Name.Set.add name names
+                else
+                  match get_alias_opt (expand var) with
+                  | None ->
+                    changed := true;
+                    names
```
lthls, line 3179: If I'm not mistaken, we could also update the set of known naked immediates here in some cases (I'm not suggesting to actually implement it, but maybe we could leave a comment here like for the constant cases)

middle_end/flambda2/types/traversals.ml
```diff
@@ -1040,11 +1040,12 @@ struct
...
   and rewrite_head_of_kind_naked_immediate
       (head : TG.head_of_kind_naked_immediate) : _ Or_unknown.t =
-    match head with
-    | Naked_immediates _ -> Or_unknown.Known head
-    | Is_int _ | Get_tag _ | Is_null _ ->
-      (* CR bclement: replace with prove. *)
-      Or_unknown.Unknown
+    match TG.Head_of_kind_naked_immediate.descr head with
+    | { naked_immediates; inverse_relations = _ } ->
+      (* CR bclement: keep inverse_relations *)
```
lthls, line 1045: How hard would it be to fix this CR now ? (It probably doesn't matter in practice, but might help in a few cases with return types)

middle_end/flambda2/types/meet_and_n_way_join.ml
```diff
@@ -638,6 +629,107 @@ let join_array_element_kinds (element_kind1 : _ Or_unknown_or_bottom.t)
...
+    any_input : 'a meet_return_value;
+    right_inputs : 'a
+  }
+
+let meet_inverse_relations inverse_relations1 inverse_relations2 =
+  if inverse_relations1 == inverse_relations2
+  then
+    { left_inputs = TG.Relation.Map.empty;
+      any_input = Both_inputs;
+      right_inputs = TG.Relation.Map.empty
```
lthls, line 699: I'm having trouble understanding why it's correct to put empty maps here. From reading the rest of the code, I suspect that the `left_input` and `right_inputs` maps mean the relations that only occur on the corresponding input. If that is the case, I think it deserves a comment (or renaming the fields to `left_input_only` and `right_input_only`).

middle_end/flambda2/types/meet_and_n_way_join.ml
```diff
@@ -2555,53 +2666,46 @@ and n_way_join_head_of_kind_naked_immediate env
...
+        | Ok head2 ->
+          let descr2 = TG.Head_of_kind_naked_immediate.descr head2 in
+          let immediates : _ Or_unknown.t =
+            match immediates, descr2.naked_immediates with
+            | Unknown, _ | _, Unknown -> Unknown
+            | Known imms, Known imms2 -> Known (I.Set.union imms imms2)
+          in
+          let relations =
+            (* CR bclement: The inverse relations should be rebuilt from the
+               direct relations during join. *)
```
lthls, line 2687: I think this code should be either removed (its result is ignored anyway) or updated as described by the comment.

middle_end/flambda2/types/meet_and_join.ml
```diff
@@ -2061,69 +2172,54 @@ and join_head_of_kind_naked_immediate env
...
-          (I.Set.add (I.zero machine_width)
-             (I.Set.add (I.one machine_width) is_null))
-      in
-      match head with
-      | Ok head -> Known head
-      | Bottom ->
-        Misc.fatal_error
-          "Did not expect [Bottom] from [create_naked_immediates]"))
-  | (Is_int _ | Get_tag _ | Is_null _), (Is_int _ | Get_tag _ | Is_null _) ->
+  let only_in_target_join_env inverse_relations =
```
lthls, line 2175: I'm not completely sure that this function is necessary (the old join is supposed to re-introduce with mode `In_types` the variables that occur free in the joined types), but I agree that it's correct to do that and it may even help by keeping the environment reasonably sized.

lthls (commented):
I have reviewed everything except `meet_and_join.ml`, for which I only reviewed the join (I'm assuming the meet is identical to the one in `meet_and_n_way_join.ml`).
It looks correct, but I'd like to see the n-way join resolved properly if possible.

middle_end/flambda2/types/grammar/type_grammar.ml
bclement-ocp: Yes, that is the case — I will add a comment explaining this.

middle_end/flambda2/types/grammar/type_grammar.ml
bclement-ocp: I think that's right -- we could filter based on the expanded type of `var`. I will also add a comment (however since writing this code, I'm not sure it is a good idea to do any kind of reductions here)

middle_end/flambda2/types/meet_and_join.ml
bclement-ocp: I wrote this code with the assumption that keeping a relation to something that doesn't exist in the `target_join_env` would be useless, but it is right that we can have names that exist in all (or multiple) joined environments but not in the target env and would get imported — I'll remove it (it could also hide some bugs).

middle_end/flambda2/types/meet_and_n_way_join.ml
bclement-ocp: Yeah, `left_inputs` and `read_inputs` are poorly named — they indeed contain the relations that only hold for one side but not the other. I'll rename the fields/variables and add a comment.

middle_end/flambda2/types/meet_and_n_way_join.ml
bclement-ocp: It's a bit tricky to actually do the rebuilding from the direct relations currently, because we can only put the reverse relation in the appropriate env extension, but we don't have the relevant information when we do the join of the env extensions, so we'd need to process the extension twice. I need to think about it a bit more if we want to do it; maybe we can do an OK underapproximation from here.

middle_end/flambda2/types/traversals.ml
bclement-ocp: I need to think about it — it might be easier to do with #5133 (the main issue is that the API to communicate with the reaper about the names we would put here needs to be designed).

bclement-ocp (commented):
@lthls Sorry, i forgot about this one for a while.

I have pushed a new patch taking into account the review and adding a computation of the reverse relation from the direct relations in the n-way join, if you can take another look!

middle_end/flambda2/types/env/join_env.ml
```diff
@@ -1507,10 +1507,174 @@ let join_aliases_into_bindings ~joined_envs ~bindings equations_to_join =
           in
           equations_to_join, bindings))
 
+let rec add_inverse_relation_to_env_extension ?(seen = Name.Set.empty)
+    env_extension name relation ~scrutinee =
+  let empty_descr : TG.Head_of_kind_naked_immediate.descr =
+    { naked_immediates = Unknown; inverse_relations = TG.Relation.Map.empty }
+  in
+  let[@inline] type_from_descr (descr : TG.Head_of_kind_naked_immediate.descr) =
```
lthls: I would find it a bit easier to follow if this function was called `updated_type_from_descr` or something like that. `type_from_descr` looks like the name of a generic descr-to-type function.

lthls (approved):
I've read the last commit and it looks great, although I have made a small suggestion to help future readers.

middle_end/flambda2/types/env/join_env.ml
bclement-ocp: Renamed as suggested.

PR #4395, opened 2025-07-28
Runtime quotations: standard library, parsing messages, typing rules

Dreian:
The pull request introduces runtime quotations to OxCaml. They are introduced using `<[`, `]>` brackets (`LESSLBRACKET`, `RBRACKETGREATER` tokens), and can be eliminated ("spliced") inside quotations using `$` (`DOLLAR` token).

Quotations are represented as an AST-like structure defined in stdlib in the `camlInternalQuote` module.

The notion of stage is added to the type system ("how many laters of quotations is an expression under").

Currently there is no support for opening modules inside quotations, which is detected by the type system.


otherlibs/dynlink/dune
```diff
@@ -181,6 +182,12 @@
...
 
+(rule
+ (action
+   (progn
+     (copy# ../../stdlib/camlinternalQuote.ml camlinternalQuote_bootstrap.ml)
+     (copy# ../../stdlib/camlinternalQuote.mli camlinternalQuote_bootstrap.mli))))
+
 (copy_files ../../utils/runtimetags.ml)
 
 (copy_files ../../utils/binutils.ml)
```
Skepfyr, line 192: I can't leave a comment on the correct lines but I think:
`.dynlink_compilerlibs.objs/byte/dynlink_compilerlibs__CamlinternalQuote_bootstrap.cmo`
needs to be added in the dynlink.cma rule and
`.dynlink_compilerlibs.objs/native/dynlink_compilerlibs__CamlinternalQuote_bootstrap.cmx`
needs to be added to the dynlink.cmxa rule.

Skepfyr: (It appears it can be added just above Oprint, I think that's the first thing that depends on it.)

testsuite/tests/quotation/build_quotation.ml
```diff
@@ -0,0 +1,764 @@
...
+[%%expect {|
+Line 1, characters 3-30:
+1 | <[ let x::xs = [1; 2; 3] in xs ]>;;
+       ^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Warning 8 [partial-match]: this pattern-matching is not exhaustive.
+Here is an example of a case that is not matched:
+[]
+
+- : <[ int list ]> expr =
+<[ match (::) (1, ((::) (2, ((::) (3, []))))) with | (::) (x, xs) -> xs ]>
```
riaqn: Why is the parsetree changed?

testsuite/tests/quotation/quotation_typing.ml
```diff
@@ -0,0 +1,252 @@
+(* TEST
```
riaqn, line 1: I think this PR is lacking a bunch of runtime tests (i.e., some quote that gets compiled and run at runtime,  and gives expected output)

typing/typecore.ml
```diff
@@ -1877,7 +1882,9 @@ let solve_Ppat_variant ~refine loc env tag no_arg expected_ty =
   let make_row more =
     create_row ~fields ~closed:false ~more ~fixed:None ~name:None
   in
-  let row = make_row (newgenvar (Jkind.Builtin.value ~why:Row_variable)) in
+  let row =
+    make_row (newgenvar (Jkind.Builtin.value ~why:Row_variable))
+  in
```
riaqn: superfluous diff.

typing/env.ml
```diff
@@ -3437,39 +3488,103 @@ let walk_locks_for_mutable_mode ~errors ~loc ~env locks m0 =
...
+      | Some n ->
+          may_lookup_error errors loc env
+            (Incompatible_stage (Lident s, loc, env.stage,
+                                 tda.tda_declaration.type_loc,
+                                 env.stage - n))
+    end
+  | exception Not_found | Error _ ->
       may_lookup_error errors loc env (Unbound_type (Lident s))
 
+let lookup_ident_module (type a) (load : a load) ~errors ~use ~loc s env =
```
riaqn: The moving of this function creates huge diff. If this is to access some dependencies (`quotation_locks_offset`?), we should move the dependencies instead.

typing/env.ml
```diff
@@ -3395,8 +3426,28 @@ let walk_locks ~errors ~env ~loc ~item ~lid mode ty locks =
...
+let quotation_locks_offset locks =
+  List.fold_right
+    (fun lock rel_stage ->
+       match rel_stage with
+       | None -> None
+       | Some n ->
+         match lock with
+         | Quotation_lock -> Some (n + 1)
+         | Splice_lock -> Some (n - 1)
+         | Local_env_lock -> None
```
riaqn: Not sure why we need `Local_env_lock`. Is it ever possible that `Local_env_lock` is later (inner) than `Quotation_lock/Splice_lock`? Note that this is currently only used in `Typemod.initial_env`.

stdlib/camlinternalQuote.mli
```diff
@@ -0,0 +1,630 @@
+type 'a lam
```
riaqn, line 31: Why do we need this file? Can't we just copy `typedtree`?

testsuite/tests/quotation/quotation_typing.ml
Dreian: This PR doesn't contain evaluation. That comes next.

typing/typecore.ml
```diff
@@ -8892,8 +8933,8 @@ and type_tuple ~overwrite ~loc ~env ~(expected_mode : expected_mode) ~ty_expecte
   (* CR layouts v5: non-values in tuples *)
   let unify_as_tuple ty_expected =
     let labeled_subtypes =
-      List.map (fun (label, _) -> label,
-                                  newgenvar (Jkind.Builtin.value_or_null ~why:Tuple_element))
+      List.map
+        (fun (label, _) -> label, newgenvar (Jkind.Builtin.value_or_null ~why:Tuple_element))
```
riaqn: superfluous diff
(some others below)

typing/typecore.ml
```diff
@@ -7357,6 +7369,35 @@ and type_expect_
...
             exp_env = env }
+  | Pexp_quotation exp ->
+      let expected_mode' = expected_mode in
+      let jkind = Jkind.Builtin.value ~why:Quotation_result in
+      let new_env = Env.add_quotation_lock env in
+      let ty = newgenvar jkind in
+      let quoted_ty = newgenty (Tquote ty) in
+      let to_unify = Predef.type_code quoted_ty in
+      with_explanation (fun () ->
+        unify_exp_types loc new_env to_unify (generic_instance ty_expected));
```
riaqn: probably doesn't matter, but should be `env` instead of `new_evn`? We don't want the quotation_lock?

stdlib/camlinternalQuote.mli
Dreian: No -- as discussed, we're keeping these representations separate.

testsuite/tests/quotation/quotation_typing.ml
```diff
@@ -0,0 +1,254 @@
...
+[%%expect {|
+type ('a, 'b, 'c) s4 = <[ $ ('a) list -> $ ('b) option -> $ ('c) ]> expr
+|}];;
+
+type ('a, 'b) s5 = <[$'a -> [`A of 'b]]> expr;;
+[%%expect {|
+Line 32, characters 35-37:
+32 | type ('a, 'b) s5 = <[$'a -> [`A of 'b]]> expr;;
+                                        ^^
+Error: Type variable "b" is used at Line 32, characters 35-37
```
riaqn: I think we don't need to print the location, because it's already pointed at.

typing/env.ml
riaqn: Resolved offline.

typing/env.ml
```diff
@@ -4870,6 +4968,27 @@ let report_lookup_error _loc env ppf = function
       in
       fprintf ppf
         "@[Mutable variable cannot be used inside %s.@]" ctx
+  | Incompatible_stage (lid, usage_loc, usage_stage, intro_loc, intro_stage) ->
+      fprintf ppf
+        "@[Identifier %a is used at %a@ \
+         in a context with %a;@ \
```
riaqn: Instead of "in a context with stage..." maybe just say "in stage ..."

typing/env.ml
```diff
@@ -4870,6 +4968,27 @@ let report_lookup_error _loc env ppf = function
...
+        Location.print_loc usage_loc
+        print_stage usage_stage
+        Location.print_loc intro_loc
+        print_stage intro_stage
+  | No_constructor_in_stage (lid, usage_loc, usage_stage) ->
+      fprintf ppf
+        "@[Constructor %a used at %a@ \
+         is unbound in this context;@ \
+         identifier %a is unbound@ \
+         in a context with %a.@]"
```
riaqn: (note that we don't need the location - they are already pointed at).
Maybe just say "The constructor %a is unbound in stage ... since it was declared in stage ...".

typing/typecore.ml
Dreian: You are right!

testsuite/tests/quotation/quotation_typing.ml
```diff
@@ -0,0 +1,254 @@
+(* TEST
+ expect;
+*)
+
+#syntax quotations on
+
+type s0 = <[int]>;;
+[%%expect {|
+type s0 = <[ int ]>
```
lpw25: I think that probably we don't want the type printer to print spaces around `int` here.

testsuite/tests/quotation/quotation_typing.ml
```diff
@@ -0,0 +1,254 @@
...
+|}];;
+
+type s1 = <[string]>;;
+[%%expect {|
+type s1 = <[ string ]>
+|}];;
+
+type 'a s2 = <[$'a -> int]> expr;;
+[%%expect {|
+type 'a s2 = <[ $ ('a) -> int ]> expr
```
lpw25: If we don't need parentheses in the parser then we probably shouldn't print them here either. We certainly shouldn't print a space after the `$`.

testsuite/tests/quotation/quotation_typing.ml
```diff
@@ -0,0 +1,254 @@
...
+|}];;
+
+type ('a, 'b) s5 = <[$'a -> [`A of 'b]]> expr;;
+[%%expect {|
+Line 32, characters 35-37:
+32 | type ('a, 'b) s5 = <[$'a -> [`A of 'b]]> expr;;
+                                        ^^
+Error: Type variable "b" is used at Line 32, characters 35-37
+       in a context with one layer of quotation (<[ ... ]>);
+       it should only be used in a context with no quotations or splices.
```
lpw25: I think this error message is a little convoluted. I would aim for something like:
```
Error: Type variable "'b" cannot be used directly within a quotation,
       as it has already occurred outside of any quotations.
Hint: Perhaps you meant "$'b"?
```
Also note that your "b" in the message is missing it's `'`.

testsuite/tests/quotation/quotation_typing.ml
```diff
@@ -0,0 +1,254 @@
...
+
+type 'a t3 = $'a -> $'a -> 'a expr;;
+[%%expect {|
+Line 1, characters 27-29:
+1 | type 'a t3 = $'a -> $'a -> 'a expr;;
+                               ^^
+Error: Type variable "'a" is used at Line 1, characters 27-29
+       in a context with no quotations or splices;
+       it should only be used in a context with one layer of splicing ($).
+|}];;
```
lpw25, line 80: I'm worried about the fact this didn't error until `'a expr`. Please add a test for just:
```ocaml
type 'a t = $'a -> $'a
```
which I think should be an error because you can't just start splicing outside of any quotations.

testsuite/tests/quotation/quotation_typing.ml
```diff
@@ -0,0 +1,254 @@
...
+
+fun (x: 'a) -> <[fun (y : $'a) -> 1]>;;
+[%%expect {|
+- : 'a -> <[ $ ('a) -> int ]> expr = <fun>
+|}];;
```
lpw25, line 110: You should add a comment indicating that this example should fail once we track `staged` in kinds.

testsuite/tests/quotation/quotation_typing.ml
```diff
@@ -0,0 +1,254 @@
...
+
+fun (x: 'a) -> <[fun (y : 'a) -> 1]>;;
+[%%expect {|
+Line 1, characters 26-28:
+1 | fun (x: 'a) -> <[fun (y : 'a) -> 1]>;;
+                              ^^
+Error: Type variable "a" is used at Line 1, characters 26-28
+       in a context with one layer of quotation (<[ ... ]>);
+       it should only be used in a context with no quotations or splices.
+|}];;
```
lpw25, line 125: Could you also add an example where it will find the variable within the quotation first? Something like:
```ocaml
(<[ fun (y : 'a) -> 1 ]>, fun (x : 'a) -> ())
```

testsuite/tests/quotation/quotation_typing.ml
```diff
@@ -0,0 +1,254 @@
...
+87 | fun (x: 'a) -> <[fun (y : <['a]>) -> 1]>;;
+                                 ^^
+Error: Type variable "a" is used at Line 87, characters 28-30
+       in a context with 2 layers of quotation (<[ ... ]>);
+       it should only be used in a context with no quotations or splices.
+|}];;
+
+fun (x: <['a]> expr) -> <[fun (y : 'b) -> ($x, y)]>;;
+[%%expect {|
+- : <[ $ ('a) ]> expr -> <[ $ ('b) -> $ ('a) * $ ('b) ]> expr = <fun>
```
lpw25: We should try to get the printer to normalize away `<[$'a]>` to `'a`

testsuite/tests/quotation/quotation_typing.ml
```diff
@@ -0,0 +1,254 @@
...
+|}];;
+
+fun (x: <['a]> expr) -> <[fun (y : 'a) -> ($x, y)]>;;
+[%%expect {|
+Line 1, characters 35-37:
+1 | fun (x: <['a]> expr) -> <[fun (y : 'a) -> ($x, y)]>;;
+                                       ^^
+Error: Type variable "a" is used at Line 1, characters 35-37
+       in a context with one layer of quotation (<[ ... ]>);
+       it should only be used in a context with no quotations or splices.
```
lpw25: This error appears incorrect. It looks like both uses of the `'a` are under a single layer of quotation.

testsuite/tests/quotation/quotation_typing.ml
```diff
@@ -0,0 +1,254 @@
...
+fun (type a) (type b) x -> <[fun (y : a) -> y]>;;
+[%%expect {|
+Line 1, characters 38-39:
+1 | fun (type a) (type b) x -> <[fun (y : a) -> y]>;;
+                                          ^
+Error: Identifier "a" is used at Line 1, characters 38-39
+       in a context with one layer of quotation (<[ ... ]>);
+       it is introduced at Line 1, characters 10-11
+       in a context with no quotations or splices.
+|}];;
```
lpw25, line 155: This error message is somewhat convoluted. I would aim for something more like:
```
Error: Type "a" cannot be used directly within a quotation,
       as it was declared outside of any quotations.
Hint: Perhaps you meant "$a"?
```

testsuite/tests/quotation/quotation_typing.ml
```diff
@@ -0,0 +1,254 @@
...
+[%%expect {|
+Line 1, characters 27-29:
+1 | type 'a t3 = $'a -> $'a -> 'a expr;;
+                               ^^
+Error: Type variable "'a" is used at Line 1, characters 27-29
+       in a context with no quotations or splices;
+       it should only be used in a context with one layer of splicing ($).
+|}];;
+
+fun (x: 'a) -> <[fun (y : $'a) -> 1]>;;
```
lpw25: I think that it is better to use `let` rather than a top-level expression for these tests. Just:
```ocaml
let foo (x : 'a) = <[fun (y : $'a) -> 1]>
```
is fine.

testsuite/tests/quotation/quotation_typing.ml
```diff
@@ -0,0 +1,254 @@
...
+Error: Identifier "a" is used at Line 1, characters 38-39
+       in a context with one layer of quotation (<[ ... ]>);
+       it is introduced at Line 1, characters 10-11
+       in a context with no quotations or splices.
+|}];;
+
+(fun (type a) (type b) x -> <[fun (y : $a) -> y]>) 42;;
+[%%expect {|
+- : <[ $ ('_a) -> $ ('_a) ]> expr = <[ fun (y : _) -> y ]>
+|}];;
```
lpw25, line 165: I think it would be better to first see the type of the function definition -- which should be fully generalized -- before seeing the result of applying it.

testsuite/tests/quotation/quotation_typing.ml
```diff
@@ -0,0 +1,254 @@
...
+<[fun (type a) (type b) (x : a) (y : b) -> (x, y)]>;;
+[%%expect {|
+- : <[ $ ('a) -> $ ('b) -> $ ('a) * $ ('b) ]> expr =
+<[ fun (type a) (type b) (x : a) (y : b) -> (x, y) ]>
+|}];;
+
+type t4 = A | B;;
+[%%expect {|
+type t4 = A | B
+|}];;
```
lpw25, line 190: I think it is better to not have an expect block here, and instead include the output in the output of the next test, since this definition is only there for use in that test.

testsuite/tests/quotation/quotation_typing.ml
```diff
@@ -0,0 +1,254 @@
...
+
+<[A]>;;
+[%%expect {|
+Line 1, characters 2-3:
+1 | <[A]>;;
+      ^
+Error: Constructor "A" used at Line 1, characters 2-3
+       is unbound in this context; identifier "A" is unbound
+       in a context with one layer of quotation (<[ ... ]>).
```
lpw25: I think this error is a bit hard for a user to understand. Rather than talking about the constructor being unbound I think it would be better to say that `A` cannot be used within quotations as it was defined outside of any quotations.

testsuite/tests/quotation/quotation_typing.ml
```diff
@@ -0,0 +1,254 @@
...
+   splices and quotations are present. *)
+
+let mk_pair x = <[$x, $x]>;;
+[%%expect {|
+val mk_pair : <[ $ ('a) ]> expr -> <[ $ ('a) * $ ('a) ]> expr = <fun>
+|}];;
+
+mk_pair <[123]>;;
+[%%expect {|
+- : <[ $ (<[ int ]>) * $ (<[ int ]>) ]> expr = <[ (123, 123) ]>
```
lpw25: Similar to one of the earlier examples: we should probably make sure that the printer normalizes `$<[t]>` to `t`.

testsuite/tests/quotation/quotation_typing.ml
```diff
@@ -0,0 +1,254 @@
...
+|}];;
+
+fun (x: <[<[<[$($'a)]>]>]>) -> (x: 'a);;
+[%%expect {|
+Line 1, characters 32-33:
+1 | fun (x: <[<[<[$($'a)]>]>]>) -> (x: 'a);;
+                                    ^
+Error: This expression has type "<[ 'a ]>"
+       but an expression was expected of type "'a"
+       The type variable "'a" occurs inside "<[ 'a ]>"
```
lpw25: I don't think you should be able to get to an occurs check like this: there should already have been an error complaining that a variable was occuring at multiple stages.

testsuite/tests/quotation/build_quotation.ml
lpw25: I think this is the type-checker's fault. It sometimes replaces `let` with `match` and then type-checks the `match`. Not much that the quotation stuff can do about it, but it also shouldn't matter in practice.

testsuite/tests/quotation/build_quotation.ml
```diff
@@ -0,0 +1,767 @@
...
```
lpw25: We shouldn't put too much effort into nice printing, but it is probably worth including a case to print list literals using the nice syntax.

testsuite/tests/quotation/quotation_typing.ml
```diff
@@ -0,0 +1,254 @@
...
+|}];;
```
lpw25, line 304: There don't seem to be any examples of ordinary identifiers being used at the wrong stage. It would also be good to include examples of issues we know we haven't addressed yet, e.g. bad GADT usages and type-based disambiguation that won't work in the generated code.

lpw25 (commented):
I did a pass of review on the testsuite changes

testsuite/tests/quotation/quotation_typing.ml
Dreian: Sure, I will add this test. I agree it is a bit non-intuitive that it errors here.

testsuite/tests/quotation/build_quotation.ml
Dreian: Sure, I can add a special case to the printer.

testsuite/tests/quotation/quotation_typing.ml
Dreian: Sure, I agree GADT and type-based disambiguation tests should be included. What do you mean by "ordinary identifiers" in this context?

testsuite/tests/quotation/build_quotation.ml
riaqn: I think this is printed as `CamlinternalQuote_bootstrap.Code.t`? So I guess my real question is: why do we need another `Code.t` type instead of just `Typedtree.t`. I will discuss this in internal channel.

typing/btype.mli
```diff
@@ -63,6 +63,8 @@ val generic_level: int
 
 val newgenty: type_desc -> type_expr
         (* Create a generic type *)
+val newgenconstr: Path.t -> type_expr list -> type_expr
+        (* Create a generic type constructor *)
```
riaqn: The only callsite of this function should be replaced by call to `Predef.type_code` instead; then we don't need to add this function.

typing/btype.ml
```diff
@@ -281,7 +282,7 @@ let fold_row f init row =
       (row_fields row)
   in
   match get_desc (row_more row) with
-  | Tvar _ | Tunivar _ | Tsubst _ | Tconstr _ | Tnil
+  | Tvar _ | Tunivar _ | Tsubst _ | Tconstr _ | Tquote _ | Tsplice _ | Tnil
```
riaqn: Can you try to move this to the `asssert false` branch instead? I think `Tquote` and `Tsplice` shouldn't appear as row variable.

typing/env.ml
```diff
@@ -163,9 +163,19 @@ type lock =
...
 
 type locks = lock list
 
+type empty = |
+
+type mode_with_locks = Mode.Value.l * locks
+
+let locks_empty = []
+let locks_is_empty l = l = locks_empty
+
```
riaqn: This diff only moves those definitions, and should be reverted.

typing/env.ml
```diff
@@ -559,7 +560,7 @@ module IdTbl =
             (Pdot (root, name), desc) :: find_all wrap name next
           with Not_found ->
             find_all wrap name next
-          end
+        end
```
riaqn: superfluous diff.

typing/env.ml
```diff
@@ -471,7 +472,7 @@ module IdTbl =
         | Nothing -> raise exn
         end
 
-    let find_same id (tbl : (empty, _, _) t) =
+    let find_same id (tbl : (_, _, _) t) =
```
riaqn: The diff makes the code worse. (we used to ensure that the locks must be of type `empty` and thus safe to ignore)

typing/env.ml
```diff
@@ -163,9 +163,19 @@ type lock =
   | Region_lock
   | Exclave_lock
   | Unboxed_lock (* to prevent capture of terms with non-value types *)
+  | Local_env_lock (* to allow reference of top-level names inside quotations *)
```
riaqn: This is only used once, after creating the initial environment, which is kind of wasteful as a constructor.

Instead, it's more straightforward to do this: after looking up, you get its path and locks, etc. You extract the head component of the path (special case of `Papply` cannot be a top-level name), which is an ident, then you call `Ident.is_global_or_predef`. 

I think you want to make a function `if_cross_quotation` that checks `path` and potentially call `quotation_locks_offset` (which can be simplified now). Then you will use that function instead of `quotation_locks_offset`.

Then make another function `check_cross_quotation` as described in another comment below.

typing/env.ml
```diff
@@ -2846,6 +2853,36 @@ let add_exclave_lock env = add_lock Exclave_lock env
 
 let add_unboxed_lock env = add_lock Unboxed_lock env
 
+let add_local_env_lock env = add_lock Local_env_lock env
+
+let add_quotation_lock env =
+  add_lock Quotation_lock {env with stage = env.stage + 1}
```
riaqn: Better call this `enter_quotation`, since it does more than just adding a quotation lock.

typing/env.ml
```diff
@@ -2846,6 +2853,36 @@ let add_exclave_lock env = add_lock Exclave_lock env
...
+
+let add_splice_lock env = add_lock Splice_lock {env with stage = env.stage - 1}
```
riaqn: Similarly, this should be called `enter_splice`.

Moreover, having seen its only call site, I think this function should check that `stage > 0`, and raise error otherwise. (that means it should take a `~loc` parameter). Then we no longer need `without_open_quotations`.

typing/typecore.ml
```diff
@@ -7356,6 +7368,35 @@ and type_expect_
...
+        exp_desc = Texp_quotation arg;
+        exp_loc = loc; exp_extra = [];
+        exp_type = instance ty_expected;
+        exp_attributes = sexp.pexp_attributes;
+        exp_env = new_env }
+  | Pexp_splice exp ->
+      if Env.without_open_quotations env then
+        raise (Error (exp.pexp_loc, env, Toplevel_splice));
+      let argument_mode = mode_legacy in
+      let new_env = Env.add_splice_lock env in
```
riaqn: If we do what I suggested for `add_splice_lock`, here we can simply call `enter_splice`. Then we don't need the function `without_open_quotations`.

typing/env.ml
```diff
@@ -3290,7 +3327,14 @@ let lookup_global_name_module_no_locks
...
     match find_name_module ~mark:use s env.modules with
-    | res -> res
+    | path, locks, data -> begin
+        match quotation_locks_offset locks with
+        | None | Some 0 -> path, locks, data
+        | Some n ->
+          may_lookup_error errors loc env
+            (Incompatible_stage (Lident s, loc, env.stage,
+                                 Location.none, env.stage - n))
+    end
```
riaqn: This is a common pattern and should be factored out into something like:
```
check_cross_quotation path locks ~loc ~lid:(Lident s) env.stage;
path, locks, data
```
where `check_cross_quotation` calls `if_cross_quotation` (also see my comment on `Local_env_lock`), and might raise `Incompatible_stage`.

This should help greatly reduce the diffs below.

typing/typetexp.ml
```diff
@@ -1079,9 +1097,19 @@ and transl_type_aux env ~row_context ~aliased ~policy mode styp =
...
+        jkind_of_annotation (Type_of_kind loc) styp.ptyp_attributes jkind
+      in
+      let ty = newty (Tof_kind tjkind) in
+      ctyp (Ttyp_of_kind jkind) ty
+  | Ptyp_quote t ->
+      let new_env = Env.add_quotation_lock env in
+      let cty = transl_type new_env ~policy ~row_context mode t in
+      ctyp (Ttyp_quote cty) (newty (Tquote cty.ctyp_type))
+  | Ptyp_splice t ->
+      let new_env = Env.add_splice_lock env in
```
riaqn: Bug here: this means writing `$(int -> int)` (outside of any quotation) would be allowed - but it shouldn't.

If we adopt my comment on `add_split_lock`, this bug should go away.

Can you add a test to prevent future regression?

typing/env.ml
```diff
@@ -4009,12 +4100,12 @@ let lid_without_hash = function
 let lookup_type ~errors ~use ~loc lid env =
   match lid_without_hash lid with
   | None ->
-    let path, tda = lookup_type_full ~errors ~use ~loc lid env in
+    let path, _, tda = lookup_type_full ~errors ~use ~loc lid env in
```
riaqn: Bug here: I think you shouldn't just ignore the locks (why modify `lookup_type_full` to return it, otherwise?). This would allow you to write:
```
<[ type t = ...
     $( x : t) ]>
```
which is wrong.

To fix this, you should call `check_cross_quotation`.

Please add test to prevent future regression.

(Similar for such ignoring of locks below; please all add tests)

typing/predef.ml
```diff
@@ -51,6 +51,7 @@ and ident_extension_constructor = ident_create "extension_constructor"
 and ident_floatarray = ident_create "floatarray"
 and ident_lexing_position = ident_create "lexing_position"
 and ident_atomic_loc = ident_create "atomic_loc"
+and ident_code = ident_create "expr"
```
riaqn, line 56: Not super vital, but why different names? code vs. expr?

typing/jkind.ml
```diff
@@ -3937,6 +3941,10 @@ module Debug_printers = struct
     | Class_term_argument -> fprintf ppf "Class_term_argument"
     | Debug_printer_argument -> fprintf ppf "Debug_printer_argument"
     | Recmod_fun_arg -> fprintf ppf "Recmod_fun_arg"
+    | Quotation_result -> fprintf ppf "Quotation_result"
+    | Antiquotation_result -> fprintf ppf "it's the result type of splicing"
+    | Tquote -> fprintf ppf "it's a staged type"
+    | Tsplice -> fprintf ppf "it's a splice of a staged type"
```
riaqn: This is debug printer, not pretty printer.

typing/typecore.ml
```diff
@@ -4482,8 +4490,8 @@ let rec is_nonexpansive exp =
   | Texp_override _
   | Texp_letexception _
   | Texp_letop _
-  | Texp_extension_constructor _ ->
-    false
+  | Texp_extension_constructor _
+  | Texp_antiquotation _ -> false
```
riaqn: You can slightly reduce the diff by putting `Texp_antiquotation` before `Texp_extension_constructor`.

typing/env.ml
```diff
@@ -2846,6 +2853,36 @@ let add_exclave_lock env = add_lock Exclave_lock env
...
 
+let add_local_env_lock env = add_lock Local_env_lock env
+
+let add_quotation_lock env =
+  add_lock Quotation_lock {env with stage = env.stage + 1}
+
+let add_splice_lock env = add_lock Splice_lock {env with stage = env.stage - 1}
+
+let without_open_quotations env = env.stage = 0
+let has_open_quotations env = env.stage <> 0
```
riaqn: Having seen all the call sites of this function, I think we should instead expose a function `val Env.check_no_open_quotations : no_open_quotation_context -> ...` that will raise error if `stage > 0`

where `type no_open_quotation_context = Open | Object | Structure | ...`.

typing/typecore.ml
```diff
@@ -7356,6 +7368,35 @@ and type_expect_
             exp_type = exp2.exp_type;
             exp_attributes = sexp.pexp_attributes;
             exp_env = env }
+  | Pexp_quotation exp ->
+      let expected_mode' = expected_mode in
```
riaqn: Are you explicitly stating that the mode of the quotation will simply follow the mode of the body? I find that requires lots of convincing. It seems (from the `Pexp_splice` case) for now we just take the interface between stages as `legacy`, which I'm happy with. If so, we should do that here too. That means:

- We will check that `legacy <= expected_mode`
- We will type check `exp` with `mode_legacy`.

typing/typecore.ml
```diff
@@ -7356,6 +7368,35 @@ and type_expect_
...
+      let to_unify = Predef.type_code quoted_ty in
+      with_explanation (fun () ->
+        unify_exp_types loc env to_unify (generic_instance ty_expected));
+      let arg = type_expect new_env expected_mode' exp (mk_expected ty) in
+      re {
+        exp_desc = Texp_quotation arg;
+        exp_loc = loc; exp_extra = [];
+        exp_type = instance ty_expected;
+        exp_attributes = sexp.pexp_attributes;
+        exp_env = new_env }
```
riaqn: `exp_env` should be `env` instead. (similar for the case below).

typing/typecore.ml
```diff
@@ -7356,6 +7368,35 @@ and type_expect_
...
+  | Pexp_splice exp ->
+      if Env.without_open_quotations env then
+        raise (Error (exp.pexp_loc, env, Toplevel_splice));
+      let argument_mode = mode_legacy in
```
riaqn: I think we are missing a mode check that `legacy <= expected_mode`.

typing/typecore.ml
```diff
@@ -8890,8 +8931,9 @@ and type_tuple ~overwrite ~loc ~env ~(expected_mode : expected_mode) ~ty_expecte
   (* CR layouts v5: non-values in tuples *)
   let unify_as_tuple ty_expected =
     let labeled_subtypes =
-      List.map (fun (label, _) -> label,
-                                  newgenvar (Jkind.Builtin.value_or_null ~why:Tuple_element))
+      List.map
+        (fun (label, _) ->
+           label, newgenvar (Jkind.Builtin.value_or_null ~why:Tuple_element))
```
riaqn: superfluous diff.

typing/typecore.ml
```diff
@@ -9023,7 +9065,8 @@ and type_construct ~overwrite env (expected_mode : expected_mode) loc lid sarg
   let constr, locks =
     wrap_disambiguate "This variant expression is expected to have"
       ty_expected_explained
-      (Constructor.disambiguate Env.Positive lid env expected_type) constrs
+      (Constructor.disambiguate Env.Positive lid env expected_type)
+      constrs
```
riaqn: superfluous diff.

typing/typecore.ml
```diff
@@ -9679,7 +9722,7 @@ and type_let ?check ?check_strict ?(force_toplevel = false)
               type_pattern_list Value existential_context env mutable_flag spatl
                 nvs sorts allow_modules
             ) ~post:(fun (_, _, _, pvs, _) ->
-                       iter_pattern_variables_type generalize pvs)
+                iter_pattern_variables_type generalize pvs)
```
riaqn: superfluous diff.

typing/typecore.ml
```diff
@@ -9823,7 +9866,7 @@ and type_let ?check ?check_strict ?(force_toplevel = false)
                - : 'a array -> int -> 'a = <fun>
              ]}
              so we do it anyway. *)
-              generalize exp.exp_type
+            generalize exp.exp_type
```
riaqn: superfluous diff.

typing/typecore.ml
```diff
@@ -235,6 +235,7 @@ type error =
   | Unrefuted_pattern of pattern
   | Invalid_extension_constructor_payload
   | Not_an_extension_constructor
+  | Invalid_quote_payload
```
riaqn: This error is never constructed.

typing/typecore.ml
```diff
@@ -295,6 +296,10 @@ type error =
       { some_args_ok : bool; ty_fun : type_expr; jkind : jkind_lr }
   | Overwrite_of_invalid_term
   | Unexpected_hole
+  | Toplevel_splice
+  | Quotation_object
+  | Open_inside_quotation
+  | Unsupported_quotation_construct
```
riaqn: This error is never constructed.

typing/typedecl.ml
```diff
@@ -3292,7 +3292,8 @@ let transl_type_extension extend env loc styext =
       List.iter Ctype.generalize type_params;
       List.iter
         (fun (ext, _shape) ->
-          Btype.iter_type_expr_cstr_args Ctype.generalize ext.ext_type.ext_args;
+          Btype.iter_type_expr_cstr_args
+             Ctype.generalize ext.ext_type.ext_args;
```
riaqn: superfluous diff.

typing/typedecl.ml
```diff
@@ -3351,7 +3352,8 @@ let transl_exception env sext =
         transl_extension_constructor ~scope env
           Predef.path_exn [] [] Asttypes.Public sext)
       ~post: begin fun (ext, _shape) ->
-        Btype.iter_type_expr_cstr_args Ctype.generalize ext.ext_type.ext_args;
+        Btype.iter_type_expr_cstr_args
+          Ctype.generalize ext.ext_type.ext_args;
```
riaqn: superfluous diff.

typing/typedecl.ml
```diff
@@ -3808,7 +3810,7 @@ let transl_value_decl env loc ~modalities valdecl =
           ~on_application:false
           ~default_arity valdecl.pval_attributes
       in
-      let zero_alloc =
+      let zero_alloc  =
```
riaqn: superfluous diff.

typing/typedecl_separability.ml
```diff
@@ -147,6 +147,7 @@ let rec immediate_subtypes : type_expr -> type_expr list = fun ty ->
       (* these should only occur under Tobject and not at the toplevel,
          but "better safe than sorry" *)
       immediate_subtypes_object_row [] ty
+  | Tquote _ | Tsplice _ -> []
```
riaqn: This case seems less important, but I think `Tquote x` can return `[x]`. Similar for `Tsplice`.

typing/value_rec_check.ml
```diff
@@ -273,7 +273,9 @@ let classify_expression : Typedtree.expression -> sd =
     | Texp_assert _
     | Texp_try _
     | Texp_override _
-    | Texp_letop _ ->
+    | Texp_letop _
+    | Texp_quotation _
```
riaqn, line 278: Intuitively, both quotation and antiquotation feel like static?

typing/uniqueness_analysis.ml
```diff
@@ -2485,6 +2524,12 @@ let rec check_uniqueness_exp ~overwrite (ienv : Ienv.t) exp : UF.t =
       Paths.mark
         (Usage.maybe_unique use occ)
         Learned_tags.empty Overwrites.empty p)
+  | Texp_quotation e ->
+    let uf = check_uniqueness_exp ~overwrite:None ienv e in
+    UF.quote uf
+  | Texp_antiquotation e ->
+    let uf = check_uniqueness_exp ~overwrite:None ienv e in
+    UF.antiquote uf
```
riaqn, line 2534: what if I just treat `Texp_quotation t` as `t`? Do you have examples where the current approach is better?

riaqn:
I've reviewed everything in `typing/` except those to do with unification and generalization (mostly in `ctype.ml`)

typing/value_rec_check.ml
Dreian: Hm, you might be right about this. @lpw25 thoughts?

typing/typetexp.ml
Dreian: I agree with this.

testsuite/tests/quotation/build_quotation.ml
Dreian: I have made updates to the printer so that lists are printed in a somewhat more readable way.

typing/env.ml
Dreian: This has now been moved.

typing/predef.ml
Dreian: We initially wanted to use `t code`, but `expr` is more accurate because it's only expressions being quoted and not patterns/types. There are a few references to `code` internally and I could replace all of them with `expr`. Do you think it would be a worthwhile change to make?

typing/jkind.ml
Dreian: ah, copy issue; will fix

typing/env.ml
Dreian: I like this idea, thank you! I will give it a go after addressing other comments.

typing/btype.ml
Dreian: I think so too.

testsuite/tests/quotation/quotation_typing.ml
Dreian: Do you mean the type printer?

testsuite/tests/quotation/quotation_typing.ml
Dreian: This actually exposes a bug/assumption! I will look into this.

lambda/translcore.ml
```diff
@@ -1330,6 +1331,10 @@ and transl_exp0 ~in_new_scope ~scopes sort e =
       Location.todo_overwrite_not_implemented ~kind:"Translcore" e.exp_loc
   | Texp_hole _ ->
       Location.todo_overwrite_not_implemented ~kind:"Translcore" e.exp_loc
+  | Texp_quotation exp -> transl_quote (transl_exp ~scopes sort) exp e.exp_loc
+  (* TODO: update scopes *)
+  | Texp_antiquotation _ ->
+      failwith "Cannot unqoute outside of a quotation context."
```
Skepfyr:
```suggestion
      failwith "Cannot unquote outside of a quotation context."
```

typing/typecore.ml
Dreian: Thank you! My knowledge here is not great, so I appreciate the suggestion.

typing/env.ml
Dreian: @lpw25 left a similar comment -- I've simplified the messages somewhat!

typing/env.ml
Dreian: While I agree with you in principle, it is (currently) impossible to write a type declaration inside a quotation.

typing/env.ml
Dreian: I agree about adding the `stage > 0` check and getting rid of `has_open_quotations`.

However, I am not sure about the naming it `enter_splice`, because I prefer naming consistency.

typing/env.ml
Dreian: Agree. Honestly don't know how and why this happened. Undoing.

testsuite/tests/quotation/quotation_typing.ml
lpw25: I just mean identifiers for ordinary values, as opposed to types or variant constructors. You have an example of using the `A` constructor above at the wrong stage, but nothing like:
```ocaml
let foo x = <[ x ]>
```

typing/uniqueness_analysis.ml
Dreian: @lpw25 was the author of this originally and can provide context.

testsuite/tests/quotation/quotation_typing.ml
Dreian: By the way, the reason why this is not reported earlier is that the "no top-level splices" logic was not added to types. Will fix this.

typing/env.ml
riaqn: What consistency are you refering to? I think `enter_quotation` and `enter_splice` is consistent naming?

Or do you mean `add_closure_locks` and such? Note that in those cases, the function really only adds the lock, so their names are accurate.

typing/env.ml
riaqn: Sure - then how about:
```
<[ let foo (type a) () = $(42 : a) ]>
```
The main point is that (IIUC) looking up types should walk the quotation locks.

typing/predef.ml
riaqn: I actually think `t code` is unambiguous, because (I think?) language categories don't mix even across stages? That is, a type `t` in stage 1 can be embedded into stage 0 as `<[ t ]>`, which is type-like (in the sense of having a jkind). We will never have an image of `t` in stage 0 available as an expression? That means, whenever in stage 0 we see an **_expression_** of type `_ code`, we know it must corresponds to an **_expression_** in stage 1 or so.

typing/env.ml
```diff
@@ -1462,7 +1462,9 @@ let rec find_type_data path env seen =
     }
   | exception Not_found -> begin
       match path with
-      | Pident id -> IdTbl.find_same id env.types
+      | Pident id ->
+          let ty, _locks =  IdTbl.find_same_and_locks id env.types in
```
riaqn: You can use `find_same_without_locks`, similar for a bunch more below.

typing/jkind.ml
```diff
@@ -3943,8 +3943,8 @@ module Debug_printers = struct
     | Recmod_fun_arg -> fprintf ppf "Recmod_fun_arg"
     | Quotation_result -> fprintf ppf "Quotation_result"
     | Antiquotation_result -> fprintf ppf "it's the result type of splicing"
```
riaqn: this line should be a debug printer.

typing/typetexp.ml
```diff
@@ -1501,11 +1501,22 @@ let report_unbound_variable_reason ppf = function
...
-  if stage = 0 then fprintf ppf "no quotations or splices"
-  else if stage = 1 then fprintf ppf "one layer of quotation (<[ ... ]>)"
-  else if stage = -1 then fprintf ppf "one layer of splicing ($)"
-  else if stage > 1 then fprintf ppf "%d layers of quotation (<[ ... ]>)" stage
-  else fprintf ppf "%d layers of splicing ($)" stage
+  if stage = 0 then fprintf ppf "outside quotations"
+  else if stage = 1 then fprintf ppf "within a quotation (<[ ... ]>)"
+  else if stage > 1 then
+    fprintf ppf "within %d layers of quotation (<[ ... ]>)" stage
+  else assert false
```
riaqn: This function duplicates `Env.print_stage`. We should let `Env` expose that and use that here.

testsuite/tests/quotation/quotation_typing.ml
```diff
@@ -51,141 +56,186 @@ type 'a t1 = 'a expr
...
 Line 1, characters 27-29:
 1 | type 'a t3 = $'a -> $'a -> 'a expr;;
                                ^^
-Error: Type variable "'a" is used at Line 1, characters 27-29
-       in a context with no quotations or splices;
-       it should only be used in a context with one layer of splicing ($).
+Error: Type variable "'a" is used at Line 1, characters 27-29,
+       outside quotations;
+       it already occurs Uncaught exception: Typetexp.Error(_, _, _)
+
```
riaqn: I think the exception here is because typetexp directly `add_splice_lock` without checking? That makes type `'a` successfully parsed at stage `-1`, and only triggers exception later when you tries to print it.

I think if you use the `enter_splice` function as commented elsewhere, we should get the correct type error.

typing/typetexp.ml
```diff
@@ -506,35 +515,44 @@ end = struct
...
-              add name v2 jkind
-            | Closed, true ->
-              raise(Error(loc, env,
-                          Unbound_type_variable (Pprintast.tyvar_of_name name,
-                                                 get_in_scope_names (),
-                                                 None)))
-            | Closed_for_upstream_compatibility, true ->
-              raise(Error(loc, env,
-                          Unbound_type_variable (Pprintast.tyvar_of_name name,
+      (fun name (ty, loc, s) ->
```
riaqn, line 516: A large part of the diff below are due to indentation - I think they should be reverted to highlight the real changes.

typing/typetexp.ml
```diff
@@ -506,35 +515,44 @@ end = struct
...
+           then try
+               let (type_expr, stage) = lookup_global name in
+               if stage <> s then
+                 raise
+                   (Error (loc, env, (Invalid_variable_stage
+                                        {name = Pprintast.tyvar_of_name name;
+                                         intro_stage = stage;
+                                         usage_loc = loc;
+                                         usage_stage = s})));
+               r := (loc, v, type_expr, stage) :: !r
```
riaqn: It seems that the `stage` stored in `r` is never used, so it's better to never store it.

typing/typetexp.ml
```diff
@@ -165,15 +170,15 @@ module TyVarEnv : sig
   (* evaluate with a locally extended set of univars *)
 
   val ttyp_poly_arg :
-    poly_univars -> (string * Parsetree.jkind_annotation option) list
+    poly_univars -> (string * Parsetree.jkind_annotation option * int) list
```
riaqn: All callsites of this function ignores the new `int` value, so I think it's better to not add this.

typing/typetexp.ml
```diff
@@ -165,15 +170,15 @@ module TyVarEnv : sig
...
 
-  val make_poly_univars : string Location.loc list -> poly_univars
+  val make_poly_univars : (string Location.loc * int) list -> poly_univars
   (* a version of [make_poly_univars_jkinds] that doesn't take jkinds *)
 
   val make_poly_univars_jkinds :
-    context:(string -> Jkind.History.annotation_context_lr) ->
-    (string Location.loc * Parsetree.jkind_annotation option) list
+    context:(string -> Jkind.History.annotation_context_lr)
+    -> (string Location.loc * Parsetree.jkind_annotation option * int) list
```
riaqn: We can reduce diff by moving the `->` back to the last line.

typing/typetexp.ml
```diff
@@ -372,15 +380,15 @@ end = struct
            match get_desc v with
            | Tvar { name; jkind } when get_level v = Btype.generic_level ->
                set_type_desc v (Tunivar { name; jkind });
-               v :: acc
+               v::acc
```
riaqn: superfluous diff.

typing/typetexp.ml
```diff
@@ -419,7 +427,7 @@ end = struct
     vs |> List.iter (fun v ->
       match get_desc v with
       | Tunivar { name; jkind } ->
-         set_type_desc v (Tvar { name; jkind })
+          set_type_desc v (Tvar { name; jkind })
```
riaqn: superfluous diff

typing/typetexp.ml
```diff
@@ -1403,7 +1440,7 @@ let transl_type_scheme_mono env styp =
       TyVarEnv.reset ();
       transl_simple_type ~new_var_jkind:Sort env ~closed:false Alloc.Const.legacy styp
     end
-    ~post:generalize_ctyp
+      ~post:generalize_ctyp
```
riaqn: superfluous diff.

typing/ctype.ml
```diff
@@ -4648,6 +4747,8 @@ let filter_arrow env t l ~force_tpoly =
       else raise (Filter_arrow_failed
                     (Label_mismatch
                        { got = l; expected = l'; expected_type = t }))
+  | Tsplice ty -> filter_arrow env ty l ~force_tpoly
+  | Tquote ty -> filter_arrow env ty l ~force_tpoly
```
riaqn: These two cases should raise error (same as the next case); example:
```
let foo (f : <[ int -> int ]>) =
  f 42
```
This should error - but I suspect that it would currently pass.

Also please add tests to avoid future regression.

typing/ctype.ml
```diff
@@ -7318,7 +7439,6 @@ let print_global_state fmt global_state =
     print_field fmt "global_level" global_level;
   in
   Format.fprintf fmt "@[<1>{@;%a}@]" print_fields global_state
-
               (*******************************)
```
riaqn, line 7431: superfluous diff

.github/workflows/build.yml
```diff
@@ -145,7 +145,7 @@ jobs:
             dwarf_tests_only: true
             oxcaml_llvm_tag: oxcaml-lldb-16.0.6-minus1
             oxcaml_llvm_source_branch: lldb
-
+
```
mshinwell: Superfluous whitespace (same below).

stdlib/camlinternalQuote.ml
```diff
@@ -0,0 +1,2766 @@
+type 'a lam = 'a
```
mshinwell, line 31: Missing MIT licence header - needed for all new files.  See `dwarf_inlined_frames.ml`

lambda/translcore.ml
```diff
@@ -24,6 +24,7 @@ open Typedtree
 open Typeopt
 open Lambda
 open Translmode
+open Translquote
```
mshinwell: There's no need to `open` here.  Just use `Translquote.transl_quote` below.

lambda/translcore.ml
```diff
@@ -1330,6 +1331,10 @@ and transl_exp0 ~in_new_scope ~scopes sort e =
       Location.todo_overwrite_not_implemented ~kind:"Translcore" e.exp_loc
   | Texp_hole _ ->
       Location.todo_overwrite_not_implemented ~kind:"Translcore" e.exp_loc
+  | Texp_quotation exp -> transl_quote (transl_exp ~scopes sort) exp e.exp_loc
+  (* TODO: update scopes *)
+  | Texp_antiquotation _ ->
+      failwith "Cannot unquote outside of a quotation context."
```
mshinwell: This message should use `Misc.fatal_errorf` and provide more context (you can use `%a` and a `Format`-using printer to dump the expression, for example).

driver/oxcaml_flags.ml
```diff
@@ -37,7 +37,7 @@ let cfg_eliminate_dead_trap_handlers = ref false  (* -cfg-eliminate-dead-trap-ha
 
 let cfg_prologue_validate = ref false    (* -[no-]cfg-prologue-validate *)
 let cfg_prologue_shrink_wrap = ref false    (* -[no-]cfg-prologue-shrink-wrap *)
-let cfg_prologue_shrink_wrap_threshold = ref 16384 
+let cfg_prologue_shrink_wrap_threshold = ref 16384
```
mshinwell: Spurious whitespace change

file_formats/cmt_format.ml
```diff
@@ -262,9 +262,9 @@ let iter_on_occurrences
       | Texp_object _ | Texp_pack _ | Texp_letop _ | Texp_unreachable
       | Texp_list_comprehension _ | Texp_array_comprehension _ | Texp_probe _
       | Texp_probe_is_enabled _ | Texp_exclave _
-      (* CR-someday let_mutable: maybe iterate on mutvar? *)
       | Texp_mutvar _ | Texp_setmutvar _
```
mshinwell, line 266: CR comment has been lost here

typing/typetexp.ml
```diff
@@ -528,8 +545,8 @@ end = struct
             | Closed_for_upstream_compatibility, true ->
               raise(Error(loc, env,
                           Unbound_type_variable (Pprintast.tyvar_of_name name,
-                                                 get_in_scope_names (),
-                                                 Some Upstream_compatibility))))
+                                                get_in_scope_names (),
+                                                Some Upstream_compatibility))))
```
riaqn: superfluous diff.

typing/env.ml
```diff
@@ -3459,39 +3515,44 @@ let walk_locks_for_mutable_mode ~errors ~loc ~env locks m0 =
...
             (Mutable_value_used_in_closure `Closure)
-      | Unboxed_lock -> mode
+      | Unboxed_lock | Quotation_lock | Splice_lock | Local_env_lock ->
+          mode
     ) mode locks
 
 let lookup_ident_value ~errors ~use ~loc name env =
   match IdTbl.find_name_and_locks wrap_value ~mark:use name env.values with
-  | Ok (path, locks, Val_bound vda) ->
+  | Ok (path, locks, Val_bound vda) -> begin
```
riaqn: This `begin` ... `end` pair is superfluous.

typing/typetexp.ml
```diff
@@ -506,20 +514,29 @@ end = struct
...
+              let (type_expr, stage) = lookup_global name in
+              if stage <> s then
+                raise
+                  (Error (loc, env, (Invalid_variable_stage
+                                       {name = Pprintast.tyvar_of_name name;
+                                        intro_stage = stage;
+                                        usage_loc = loc;
+                                        usage_stage = s})));
+              r := (loc, v, type_expr) :: !r
+            with Not_found ->
```
riaqn, line 530: this diff line is due to added indentation.

typing/env.ml
```diff
@@ -3459,39 +3515,44 @@ let walk_locks_for_mutable_mode ~errors ~loc ~env locks m0 =
...
-  | (path, data) as res ->
-      use_type ~use ~loc path data;
-      res
-  | exception Not_found ->
+  match IdTbl.find_name_and_locks wrap_identity ~mark:use s env.types with
+  | Ok (path, locks, tda) ->
+      check_cross_quotation errors loc tda.tda_declaration.type_loc env (Lident s) locks;
+      use_type ~use ~loc path tda;
+      path, locks, tda
+  | exception Not_found | Error _ ->
```
riaqn: I think `exception Not_found` won't happen here.

typing/ctype.ml
```diff
@@ -218,6 +218,22 @@ let increase_global_level () =
 let restore_global_level gl =
   global_level := gl
 
+(**** Control variable stage in inference *)
+
+let rec update_variable_stage stage_offset ty name jkind =
+  if stage_offset = 0 then ()
+  else if stage_offset < 0 then begin
+    let v = newgenvar ?name jkind in
+    let ty' = newgenty (Tquote v) in
```
riaqn: As discussed, it's better if we express that "the level of `v` and `ty'` follows that of `ty`".

testsuite/tests/quotation/quotation_typing.ml
Dreian: I don't think it hurts, though.

typing/env.ml
Dreian: I think you're right, especially now that we have locks associated with types.

typing/env.ml
Dreian: This is already handled in `typetexp.ml`, in `TyVarEnv.globalize_used_variables`: all types have their identifier validities verified.

I will add a check here, too, as it is worthwhile to report errors earlier.

testsuite/tests/quotation/quotation_typing.ml
Dreian: Fixed.

typing/typetexp.ml
Dreian: The diff is now gone.

typing/typetexp.ml
Dreian: Gone.

typing/ctype.ml
Dreian: Done. To express this, `update_variable_stage` had to be moved a bit lower.

lambda/translcore.ml
Dreian: Dumped some context.

stdlib/camlinternalQuote.ml
Dreian: Done in all 6 new non-test files.

testsuite/tests/quotation/quotation_typing.ml
Dreian: Fixed.

testsuite/tests/quotation/quotation_typing.ml
Dreian: I think what is there now is better?

typing/env.ml
Dreian: Done! Not quite in the way you suggested, but `quotation_locks_offset` is now simpler and does not need to be exposed.

typing/env.ml
Dreian: Done.

typing/typetexp.ml
Dreian: Done.

typing/env.ml
Dreian: While it is indeed nicer, I would like to add this slightly later.

testsuite/tests/quotation/quotation_typing.ml
riaqn: I think it's better to avoid duplicated information to the user.

typing/env.ml
riaqn: Anything making it difficult? I think it's not much work and improve code quality.

typing/env.ml
```diff
@@ -368,6 +370,7 @@ module TycompTbl =
 
 type empty = |
 
+
```
riaqn: superfluous white space.

typing/env.ml
```diff
@@ -3165,6 +3206,28 @@ let may_lookup_error report_errors loc env err =
...
+    Pident id -> Ident.is_global_or_predef id
+  | Pdot(p, _) | Pextra_ty (p, _) -> path_head_is_global_or_predef p
+  | Papply _ -> false
+
+let crosses_quotation path locks =
+  path_head_is_global_or_predef path
+  ||
+  (match quotation_locks_offset locks with
+   | 0 -> false
+   | _ -> true)
```
riaqn: The function should in fact be called `does_not_cross_quotation`?

Also, you can replace the pattern match with some integer operator.

typing/env.ml
```diff
@@ -3165,6 +3206,28 @@ let may_lookup_error report_errors loc env err =
...
+
+let check_cross_quotation report_errors loc_use loc_def env path lid locks =
+  if path_head_is_global_or_predef path
+  then ()
+  else
+    match quotation_locks_offset locks with
+    | 0 -> ()
+    | n ->
```
riaqn: it's better to just call `crosses_quotation` instead of duplicating its code - maybe you can make `crosses_quotation` return `(unit, int) Result.t` where the integer means the offset.

typing/env.mli
```diff
@@ -526,6 +528,11 @@ val add_closure_lock : Mode.Hint.pinpoint_desc
 val add_region_lock : t -> t
 val add_exclave_lock : t -> t
 val add_unboxed_lock : t -> t
+val enter_quotation : t -> t
+val enter_splice : loc:Location.t -> t -> t
+
+val has_open_quotations : t -> bool
+val stage : t -> int
```
riaqn: I think it's sligntly better if `Env` exposes `type stage` as an abstract type. This can prevent downstream code from doing strange stuff with integers.

This can also make downstream code much cleaner (currently they just are a bunch of `int` without comment).

typing/ctype.ml
```diff
@@ -806,29 +822,43 @@ let duplicate_class_type ty =
...
     begin match get_desc ty with
-      Tconstr (_, _, abbrev) ->
-        iter_abbrev generalize !abbrev
-    | _ -> ()
+    | Tvar name -> update_variable_stage stage_offset ty name.name name.jkind
+    | Tvariant row ->
+        if stage_offset <> 0 && is_Tvar (row_more row) then
+          lower_all ty
+        else
+          iter_type_expr (generalize stage_offset) ty
```
riaqn, line 849: The treatment for `Tvariant` here is different than what's in `iter_type_expr`(the previous treatment) - why the change?

typing/ctype.ml
```diff
@@ -2143,9 +2199,11 @@ let safe_abbrev_opt env ty =
     Btype.backtrack snap;
     false
 
-let try_expand_once_opt env ty =
+let rec try_expand_once_opt env ty =
   match get_desc ty with
     Tconstr _ -> expand_abbrev_opt env ty
+  | Tsplice t -> ignore (try_expand_once_opt env t); quote_splice_cancel ty
+  | Tquote t -> ignore (try_expand_once_opt env t); quote_splice_cancel ty
```
riaqn, line 2207: Why ignore the result of expansion? I.e., what side effects does this achieve?

N.B.: there is a comment on `expand_abbrev_opt` saying it's memorized - so it's probably mostly "pure".

typing/ctype.ml
```diff
@@ -2055,12 +2085,36 @@ let safe_abbrev env ty =
...
       false
 
+(* Cancel out all pairs of $ and <[_]>, or <[_]> and $.
+   This ensures type unification works correctly. *)
+let rec quote_splice_cancel ty =
+  match get_desc ty with
+  | Tquote t -> begin
+      match get_desc t with
+      | Tsplice t' -> t'
+      | Tquote _ -> set_type_desc ty (Tquote (quote_splice_cancel t)); ty
```
riaqn, line 2096: I feel uneasy about this mutation - we probably shouldn't mutate `type_desc`. Can you make it work without mutation?

Also, this function is doing nested (and overlapping) pattern matching. Instead, we can have three mutually recursive function `cancel_inside_quote` and `cancel_inside_splice` and `cancel_inside_other` that calls each other. For example, `cancel_inside_quote` is only called when we are already inside a `Tquote`, so when we see a `Tsplice`, that should cancel the `Tquote` on the outside.

typing/predef.ml
Dreian: We should be able to eventually support quoted patterns reasonably well.

typing/ctype.ml
```diff
@@ -3912,7 +3985,15 @@ let rec unify uenv t1 t2 =
...
-      (Tvar _, Tconstr _) when deep_occur t1 t2 ->
+    | (Tvar _, Tquote _) when deep_occur t1 t2 ->
+        unify2 uenv t1 t2
+    | (Tquote _, Tvar _) when deep_occur t2 t1 ->
+        unify2 uenv t1 t2
+    | (Tvar _, Tsplice _) when deep_occur t1 t2 ->
+        unify2 uenv t1 t2
+    | (Tsplice _, Tvar _) when deep_occur t2 t1 ->
+        unify2 uenv t1 t2
+    | (Tvar _, Tconstr _) when deep_occur t1 t2 ->
```
riaqn, line 3999: we can move this line back to the first case and remove this line of diff.

typing/ctype.ml
Dreian: The semantics are the same when `stage_offset = 0`.
The other case deals with how we handle variant generalization/specialization, mainly related to what happens if we encounter pattern-matches on polymorphic variants inside quotations.
The relevant test for this is `build_quotation.ml`.

typing/predef.ml
Dreian: Leaving a CR comment here, to address in the future.

testsuite/tests/quotation/quotation_typing.ml
Dreian: Removed.

typing/env.mli
Dreian: Done.

typing/ctype.ml
Dreian: The point is to possibly raise the `Cannot_expand` exception. The cancellation of quotes and splices only happens if no exception is raised.

typing/ctype.ml
Dreian: Let's look into it in the future.

typing/value_rec_check.ml
Dreian: Leaving a CR here, to discuss later.

typing/uniqueness_analysis.ml
Dreian: Leaving a CR to discuss later.

typing/env.ml
Dreian: Done.

testsuite/tests/quotation/quotation_typing.ml
lpw25: This doesn't seem to have been fixed. I am still seeing it in error messages.

PR #4396, opened 2025-07-28
Properly look up generic optional argument types

zchenb:
1. Remove `Longident.t` from `Types.arg_label`, fully reverting previous module-based design
2. Add path look up for generic optional parameters, check for `[@@option_like]`
3. Enforce syntactic constraint on generic optionals, disallowing type aliases like `type 'a t = int option`.

### DIFF
--- a/typing/ctype.ml
+++ b/typing/ctype.ml
@@ -4602,8 +4602,8 @@ let filter_arrow env t l ~force_tpoly ~generic_optional_info =
                       newty2 ~level (Tconstr(t_cons,
                         [newvar2 level arg_jkind], ref Mnil))
                   | _ ->
-                      failwith "Cannot create fresh generic optional without \
-                                existing type"))
+                      Misc.fatal_error "Cannot create fresh generic optional \
+                        without existing type"))
           | Required_or_position_arg ->
               if is_position l then
                 newty2 ~level


testsuite/tests/generic-optional-arguments/test_custom_datatypes.ml
dkalinichenko-js, line 1: As discussed, since we want to support them would be good to have tests for GADTs like:

```
type (_, _) t =
  | A : ('a, 'b) t
  | B : 'b -> (int, 'b) t
```

dkalinichenko-js, line 1: Let's add tests for custom datatypes with non-value constructor arguments, like:
```
type ('b : bits32) bits32_option =
  | None32
  | Some32 of 'b
```

I think supporting those will require some changes to `matching.ml`, since instead of normal field accesses we're going to have mixed block field accesses. Let's discuss this offline.

testsuite/tests/generic-optional-arguments/test_jkind.ml
dkalinichenko-js, line 1: Why is this named `test_jkind.ml`?

testsuite/tests/generic-optional-arguments/test_custom_datatypes.ml
dkalinichenko-js, line 1: Let's add tests for checking structures against signatures and signature subsumption.

dkalinichenko-js, line 1: Let's add tests for modes and modalities. In particular, a type like below ought to have mode-correct behavior when used as an optional argument:
```
type 'a globalized_option = 
  | None
  | Some of 'a @@ global
```

dkalinichenko-js, line 1:
```
type 'a reexported = 'a or_null [@@or_null_reexport]
```
should work as well as just `or_null` . This would mean changing existing checks for `or_null` as a path to checking for either exactly `or_null` or `[@@or_null_reexport]`.

dkalinichenko-js, line 1: Let's add tests for datatypes defined in. modules like `My_option.t`.

dkalinichenko-js (changes requested):
I added those comments to a test file, but if they need non-trivial code changes, feel free to split them off into another PR.

lambda/matching.ml
```diff
@@ -4496,16 +4496,14 @@ let for_optional_arg_default ~(mpath : Btype.optional_module_path)
...
-    | Stdlib_or_null ->
-        transl_match_on_or_null
-          default_arg_layout
-          (Lvar param)
-          Loc_unknown
-          ~if_null:default_arg
-          ~if_this:(Lvar param)
-    | Stdlib_option ->
-        transl_match_on_option
+    if Path.same path Predef.path_or_null then
```
dkalinichenko-js: I already mentioned this re `[@@or_null_reexport]`, but also, having this gated by a `Path.same` check here is pretty fragile. It would be better to factor this in a separate function like `val is_or_null_variant` which lives somewhere else. 

Also, even with bools, it's often nice to use pattern-matching,

dkalinichenko-js: Now that I'm thinking, the best way to check whether a variant is `or_null` is to look at its kind, which then will be `Variant_with_null`.

typing/ctype.ml
```diff
@@ -4539,16 +4539,31 @@ type filtered_arrow =
...
 (* CR generic-optional: Remove predef_ prefix when adding non-predef jkinds *)
-let predef_jkind_of_optional_module_path = function
-  | Stdlib_option -> Predef.option_argument_jkind
-  | Stdlib_or_null -> Jkind.for_or_null_argument Predef.ident_or_null
+let jkind_of_generic_optional_type_path (decl : type_declaration) =
+  match decl.type_params with
+  | param :: _ -> (
+    match get_desc param with
+    | Tvar {jkind; _} -> jkind
+    | _ -> Misc.fatal_error "Expecting a type variable in type params"
```
dkalinichenko-js, line 4550: Do those fatal errors never happen?

typing/ctype.mli
```diff
@@ -311,10 +311,16 @@ type filtered_arrow =
     ret_mode : Mode.Alloc.lr
   }
 
-val predef_path_of_optional_module_path: Btype.optional_module_path -> Path.t
-val predef_jkind_of_optional_module_path: Btype.optional_module_path -> jkind_lr
+val path_of_generic_optional_type_path:
+        Btype.generic_optional_type_path -> Path.t
```
dkalinichenko-js, line 315: Nit: I think we usually would put less spaces here. But this file has indentation all over the place, so I'm not sure what's the correct decision.

testsuite/tests/generic-optional-arguments/test_jkind.ml
zchenb: The reason is that we are currently retrieving `jkind` from the first type parameter, and it won't work if the first type is not a type-var. But somehow this test did not trigger this case. 

Now that I think about it -- I think leaving a CR rather than having a test would be enough.

testsuite/tests/generic-optional-arguments/test_custom_datatypes.ml
zchenb: Left a CR (to be addressed in a child feature)

zchenb: Left a CR in `test_reexport.ml`.

typing/ctype.ml
```diff
@@ -4560,29 +4575,43 @@ let filter_arrow env t l ~force_tpoly =
...
-            newty2 ~level (Tconstr(t_cons,
-              [newvar2 level arg_jkind], ref Mnil))
-          | Required_or_position_arg ->
-            if is_position l then
+          | Vanilla_optional_arg ->
               newty2 ~level
-                (Tconstr (Predef.path_lexing_position, [], ref Mnil))
-            else
-              newvar2 level k_arg
+                (* CR layouts v5: Change the Jkind.Builtin.value when option can
```
dkalinichenko-js: Since we don't refer to `Jkind.Builtin.value` here, we can drop this CR.

typing/ctype.ml
```diff
@@ -4560,29 +4575,43 @@ let filter_arrow env t l ~force_tpoly =
...
+                        extract_optional_tp_from_type_exn env ty_arg
+                      in
+                      let t_cons, arg_jkind =
+                        path_of_generic_optional_type_path path,
+                          jkind_of_generic_optional_type_path decl
+                      in
+                      newty2 ~level (Tconstr(t_cons,
+                        [newvar2 level arg_jkind], ref Mnil))
+                  | _ ->
+                      failwith "Cannot create fresh generic optional without \
```
dkalinichenko-js: Do we ever hit this error?

typing/typecore.ml
```diff
@@ -8371,12 +8471,15 @@ and type_apply_arg env ~app_loc ~funct ~index ~position_and_mode ~partial_app (l
...
        | Labelled _ | Nolabel -> ()
-       | Optional _ | Generic_optional _ ->
+       | Optional _ ->
            (* CR layouts v5: relax value requirement *)
-           let mpath = get_optional_module_path_exn lbl in
            unify_exp env arg
-            (type_option mpath
-              (newvar (predef_jkind_of_optional_module_path mpath)))
+             (type_option(newvar Predef.option_argument_jkind))
+           (* We always infer the vanilla option argument, even potentially
```
dkalinichenko-js: Nit: *even if

testsuite/tests/generic-optional-arguments/test_custom_datatypes.ml
zchenb: Added to the end of `test_custom_datatypes.ml`. Not sure how to test signature subsumption, please check.

dkalinichenko-js (commented):
Some more comments.

testsuite/tests/generic-optional-arguments/test_jkind.ml
zchenb: I have renamed this file to `test_gadt.ml`

testsuite/tests/generic-optional-arguments/test_custom_datatypes.ml
zchenb: Added a test file with CRs.

typing/ctype.ml
zchenb: I am struggling to get this to show up. In most cases, a prior check will just rule out this error. 

I am expecting GADTs to trigger this case, but somehow it is still difficult. Perhaps we can discuss?

typing/ctype.mli
zchenb: I tried to make my indentation align with the rest of the file. I would recommend 8 spaces in this case.

lambda/matching.ml
zchenb: I changed this to a match and left a CR.

However, if we want to have access to the type declaration here, we need to pass the type declaration information through `Tparam_optional_default`, which is somewhat significant. We can do it in a child PR if necessary.

typing/ctype.ml
zchenb: We only hit this error if there is a bug in the compiler.

I fixed such a bug in #4439.

testsuite/tests/generic-optional-arguments/test_non_value_constructors.ml
```diff
@@ -0,0 +1,37 @@
+(* TEST
+ flags = "-extension-universe alpha";
+ expect;
```
dkalinichenko-js: This needs native tests since unboxed types are represented as boxed in bytecode.

lambda/matching.ml
dkalinichenko-js: This is fine. Let's do it in a subsequent PR.

typing/ctype.ml
dkalinichenko-js: Ack. Let's change it to `Misc.fatal_error` just to be slightly nicer.

typing/ctype.ml
dkalinichenko-js: Hm, I looked at the code and I wasn't sure how to trigger this either. I think it's fine to leave this aside until we support GADTs.

PR #4397, opened 2025-07-28
 flambda2-types: Prevent accidentally recursive meet 

bclement-ocp:
There are some situations (particularly with extensions stored on a
variable `x` that add new equations for that same variable `x`) that
causes the current (and the previous) meet algorithm to go into an
infinite loop.  The rest of the code is carefully written to avoid these
situations, but this is hampering recent work in the typing env (https://github.com/oxcaml/oxcaml/pull/4112
and https://github.com/oxcaml/oxcaml/pull/3538 are carefully working around this).

The proper long term fix for the issue is not clear. It might be similar
to what is done in https://github.com/oxcaml/oxcaml/pull/4112, where extensions are associated with
identifiers to avoid adding extensions recursively, if we want to allow
recursive extensions. Or it might be some clever mechanism to extract
recursive equations from extensions and lift them to the point of the
disjunction.

In the short term, this patch uses the `meet_env` from https://github.com/oxcaml/oxcaml/pull/4269 and https://github.com/oxcaml/oxcaml/pull/4278
in order to keep track of variables we are currently adding an equation
for, and simply dropping a second occurence of the same variable: if
this happens, we are likely entering an infinite loop of trying to add
the same equation recursively, so we simply drop the inner equation.

This should not happen currently, however it is going to start happening
when we start tracking the `Is_int` and `Get_tag` values of blocks for
the match-in-match heuristics.

Note: this PR includes (and is rebased on top of) #4278 ; only the latest commit is new in this PR.

### DIFF
--- a/middle_end/flambda2/types/env/meet_env.ml
+++ b/middle_end/flambda2/types/env/meet_env.ml
@@ -19,7 +19,10 @@ module TG = Type_grammar
 module TE = Typing_env
 module TEL = Typing_env_level
 
-type t = { typing_env : TE.t } [@@unboxed]
+type t =
+  { typing_env : TE.t;
+    adding_equations_for_names : Name.Set.t
+  }
 
 type 'a meet_return_value =
   | Left_input
@@ -33,11 +36,12 @@ type meet_type =
   Type_grammar.t ->
   (Type_grammar.t meet_return_value * t) Or_bottom.t
 
-let create typing_env = { typing_env }
+let create typing_env =
+  { typing_env; adding_equations_for_names = Name.Set.empty }
 
-let typing_env { typing_env } = typing_env
+let typing_env { typing_env; _ } = typing_env
 
-let with_typing_env { typing_env = _ } typing_env = { typing_env }
+let with_typing_env t typing_env = { t with typing_env }
 
 let use_meet_env t ~f = typing_env (f (create t))
 
@@ -50,6 +54,27 @@ let use_meet_env_strict t ~f : _ Or_bottom.t =
     if TE.is_bottom tenv then Bottom else Ok tenv
 
 let map_typing_env t ~f = with_typing_env t (f (typing_env t))
+
+let adding_equation_for_name t name ~f =
+  (* If we were to add an equation on [x] while already adding an equation on
+     [x], either the inner equation would get overriden by the outer equation
+     and would not be visible, or it would get captured in an env extension.
+
+     That env extension would then likely end up stored on [x] itself, which
+     currently would cause an infinite loop next time we try to add an equation
+     on [x].
+
+     Instead, we simply ignore such recursive equations. *)
+  (* CR bclement: Implement support for recursive extensions (i.e. extensions
+     stored on [x] that can contain equations on [x]), then get rid of this. *)
+  if Name.Set.mem name t.adding_equations_for_names
+  then t
+  else
+    let adding_equations_for_names =
+      Name.Set.add name t.adding_equations_for_names
+    in
+    let t' = f { t with adding_equations_for_names } in
+    { t' with adding_equations_for_names = t.adding_equations_for_names }
 
 let replace_concrete_equation t name ty =
   match TG.must_be_singleton ty ~machine_width:(TE.machine_width t) with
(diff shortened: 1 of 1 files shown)


middle_end/flambda2/types/env/meet_env.ml
```diff
@@ -19,17 +19,62 @@ module TG = Type_grammar
...
+
+     That env extension would then likely end up stored on [x] itself, which
+     currently would cause an infinite loop next time we try to add an equation
+     on [x].
+
+     Instead, we simply ignore such recursive equations. *)
+  (* CR bclement: Implement support for recursive extensions (i.e. extensions
+     stored on [x] that can contain equations on [x]), then get rid of this. *)
+  if Name.Set.mem name t.adding_equations_for_names
+  then Misc.fatal_error "Looping!"
```
bclement-ocp: This is obviously a remnant of some testing that was not intended to be included in the diff. However it did reveal that there is a situation other than env extensions where we could hit this condition, which is recursive types.

With recursive types (I don't think we have a way to actually construct recursive types from simplify though, so this is mostly a theoretical concern) we lose equations (but not their side conditions) already (notice how we have `z/2N: (Val ((= x/0N)))` but the equation on the second field is still `Naked_immediate {0 1}`; the `(= 0)` from `z/2N` has been dropped):

```
Environment: ((defined_symbols { }) (code_age_relation {})
              (levels
               ((scope 1) (defined_vars
                {(z/2N :: 𝕍 Normal) (x/0N :: 𝕍 Normal) (y/1N :: 𝕍 Normal)}))
               ((defined_vars { z/2N x/0N y/1N })
                (equations
                 (z/2N :
                   (Val
                    ( ((Variant
                        (blocks
                         ((alloc_mode Heap) (known
                           {(tag_0 => (Known 2), ((Val ⊤) (Val ((= 0)))))})
                          (other Bottom))) (tagged_imms (Naked_immediate ⊥))))!))
                  x/0N :
                   (Val
                    ( ((Variant
                        (blocks
                         ((alloc_mode Heap) (known
                           {(tag_0 => (Known 2),
                             ((Val ((= y/1N)))
                              (Val
                               ( ((Variant (blocks ⊥)
                                   (tagged_imms
                                    (Naked_immediate (({ 0 1 2 }))))))!))))})
                          (other Bottom))) (tagged_imms (Naked_immediate ⊥))))!))
                  y/1N :
                   (Val
                    ( ((Variant
                        (blocks
                         ((alloc_mode Heap) (known
                           {(tag_0 => (Known 2),
                             ((Val ((= z/2N)))
                              (Val
                               ( ((Variant (blocks ⊥)
                                   (tagged_imms
                                    (Naked_immediate (({ 0 1 }))))))!))))})
                          (other Bottom))) (tagged_imms (Naked_immediate ⊥))))!)))))))
              (aliases
               ((canonical_elements {}) (aliases_of_canonical_names {})
                (aliases_of_consts {}))))
Meet of: (Val ((= x/0N))) and (Val ((= y/1N)))
Result type: (Val ((= x/0N)))
New environment:
((defined_symbols { }) (code_age_relation {})
 (levels
  ((scope 1) (defined_vars
   {(z/2N :: 𝕍 Normal) (x/0N :: 𝕍 Normal) (y/1N :: 𝕍 Normal)}))
  ((defined_vars { z/2N x/0N y/1N })
   (equations
    (z/2N : (Val ((= x/0N)))
     x/0N :
      (Val
       ( ((Variant
           (blocks
            ((alloc_mode Heap) (known
              {(tag_0 => (Known 2),
                ((Val ((= z/2N)))
                 (Val
                  ( ((Variant (blocks ⊥)
                      (tagged_imms (Naked_immediate (({ 0 1 }))))))!))))})
             (other Bottom))) (tagged_imms (Naked_immediate ⊥))))!))
     y/1N : (Val ((= x/0N))))))))
 (aliases
  ((canonical_elements
    {(z/2N (x/0N (coercion id))) (y/1N (x/0N (coercion id)))})
   (aliases_of_canonical_names {(x/0N {(Normal {(z/2N id) (y/1N id)})})})
   (aliases_of_consts {}))))
```

PR #4398, opened 2025-07-28
WIP: Merge Type Shapes into Shapes for Functor Support in DWARF Emission

spiessimon:
This PR merges the recently introduced type shapes (#4253, #4318, and #4343) into the shape type of the compiler. This will allow debug information to leverage the shape mechanism of the compiler to resolve functor application and produce debug information based on `.cms` files of other `.ml` files.

Some aspects of this PR are still unfinished:

- [ ] Evaluating the impact on the size of `.cms` files now that the shapes store more information.
- [ ] Resolve the [name emission PR](https://github.com/oxcaml/oxcaml/pull/4391) and rebase on top of it; this PR currently uses `unknown/n` as type names.
- [ ] Evaluate the performance impact (and potentially improve the implementation of) the unfolding of type shapes.
- [ ] Potentially split the PR into two to simplify reviewing.
- [ ] Fix the support for Merlin. (We currently lose UIDs during the shape construction.)

### DIFF
--- a/backend/debug/dwarf/dwarf_ocaml/dwarf_type.ml
+++ b/backend/debug/dwarf/dwarf_ocaml/dwarf_type.ml
@@ -212,7 +212,7 @@ let rec field_name_with_path base path =
   | [] -> base
   | i :: path -> field_name_with_path base path ^ ".#" ^ Int.to_string i
 
-type 'layout projected_field = string option * S.without_layout S.ts * 'layout
+type 'layout projected_field = string option * S.t * 'layout
 
 let project_field_given_path (fields : Layout.t projected_field array) path :
     base_layout projected_field =
@@ -232,7 +232,7 @@ let project_field_given_path (fields : Layout.t projected_field array) path :
     let field_name = Option.value ~default:("." ^ Int.to_string i) field_name in
     let field_name_with_projection = field_name_with_path field_name subpath in
     ( Some field_name_with_projection,
-      Ts_other Layout_to_be_determined,
+      Shape.leaf' None,
       (* CR sspies: To properly support unboxed records in mixed records, we we
          need to propagate the right shape information here. *)
       project_layout field_layout subpath )
@@ -271,13 +271,8 @@ let flatten_fields_in_mixed_record ~(mixed_block_shapes : Layout.t array)
 (* This function deals with variants that are annotated with [@@unboxed]. They
    are only allowed to have a single constructor. *)
 let create_attribute_unboxed_variant_die ~reference ~parent_proto_die ~name
-    ~constr_name ~arg_name ~(arg_layout : Layout.t) ~arg_die =
-  let base_layout =
-    match arg_layout with
-    | Base base_layout -> base_layout
-    | Product _ -> Misc.fatal_error "Not a base layout"
-  in
-  let width = base_layout_to_byte_size base_layout in
+    ~constr_name ~arg_name ~arg_layout ~arg_die =
+  let width = base_layout_to_byte_size arg_layout in
   let structure_ref = reference in
   let variant_part_ref = Proto_die.create_reference () in
   let variant_member_ref = Proto_die.create_reference () in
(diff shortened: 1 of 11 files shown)


backend/debug/dwarf/dwarf_ocaml/dwarf_type.ml
```diff
@@ -1517,108 +1505,142 @@ let rec flatten_to_base_sorts (sort : Layout.t) : base_layout list =
...
+  let max_number_of_cms_files = ref 20
+  (* A single compilation may not read more than 20 .cms files. *)
+  (* CR sspies: Investigate the performance some more and the balance between
+     different variables. *)
+
+  let read_unit_shape ~unit_name =
+    match String.Tbl.find_opt cms_file_cache unit_name with
+    | Some shape -> shape
+    | None ->
+      if !max_number_of_cms_files <= 0
```
mshinwell, line 1714: This needs a command-line flag

backend/debug/dwarf/dwarf_ocaml/dwarf_type.ml
```diff
@@ -1517,108 +1505,142 @@ let rec flatten_to_base_sorts (sort : Layout.t) : base_layout list =
...
+  let fuel = 10
+
+  let cms_file_cache = String.Tbl.create 10
+
+  let max_number_of_cms_files = ref 20
+  (* A single compilation may not read more than 20 .cms files. *)
+  (* CR sspies: Investigate the performance some more and the balance between
+     different variables. *)
+
+  let read_unit_shape ~unit_name =
```
mshinwell, line 1710: Does this need a command-line flag too?

backend/debug/dwarf/dwarf_ocaml/dwarf_type.ml
```diff
@@ -1517,108 +1505,142 @@ let rec flatten_to_base_sorts (sort : Layout.t) : base_layout list =
...
+    match String.Tbl.find_opt cms_file_cache unit_name with
+    | Some shape -> shape
```
mshinwell, line 1712: I would make the default size bigger

backend/debug/dwarf/dwarf_ocaml/dwarf_type.ml
```diff
@@ -1517,108 +1505,142 @@ let rec flatten_to_base_sorts (sort : Layout.t) : base_layout list =
...
+        decr max_number_of_cms_files;
+        let filename = String.uncapitalize_ascii unit_name in
+        match Load_path.find_normalized (filename ^ ".cms") with
+        | exception Not_found -> None
+        | fn -> (
+          match Cms_format.read fn with
+          | exception Cms_format.Error _ ->
+            (* CR sspies: We could consider throwing a louder error here, since
+               there must be something like a [.cms] version mismatch here. For
+               now, since it's only debugging information, we fail silently. *)
```
mshinwell: I think you should only do this if the read succeeds

backend/debug/dwarf/dwarf_ocaml/dwarf_type.ml
```diff
@@ -1517,108 +1505,142 @@ let rec flatten_to_base_sorts (sort : Layout.t) : base_layout list =
...
+  let read_unit_shape ~unit_name =
+    match String.Tbl.find_opt cms_file_cache unit_name with
+    | Some shape -> shape
+    | None ->
+      if !max_number_of_cms_files <= 0
+      then None
+      else (
+        decr max_number_of_cms_files;
+        let filename = String.uncapitalize_ascii unit_name in
+        match Load_path.find_normalized (filename ^ ".cms") with
```
mshinwell, line 1720: This shouldn't be a string.  It should probably be Compilation_unit.t

backend/debug/dwarf/dwarf_ocaml/dwarf_type.ml
```diff
@@ -1517,108 +1505,142 @@ let rec flatten_to_base_sorts (sort : Layout.t) : base_layout list =
...
+      let shapes_with_layout = List.combine shapes layouts in
+      List.concat_map (fun (sh, ly) -> flatten_shape sh ly) shapes_with_layout
+    | Layout.Product _ -> Misc.fatal_error "unboxed tuple field mismatch"
+    | Layout.Base _ -> Misc.fatal_error "unboxed tuple must have product layout"
+    )
+  | Constr _, Base b -> [Known (type_shape, b)]
+  | Constr _, _ -> unknown_base_layouts type_layout
+  (* CR sspies: These should not happen with support for recursive types. For
+     now, we conservatively give back defaults. *)
+  | Predef _, Base base_layout -> [Known (type_shape, base_layout)]
```
mshinwell, line 1587: All of these fatal errors should print the shape involved

typing/type_shape.ml
```diff
@@ -336,85 +385,143 @@ module Type_decl_shape = struct
...
+      of_type_declaration_go type_declaration type_param_shapes shape_for_constr
+    in
+    let decl_shape = Shape.abs_list definition type_param_idents in
+    Shape.set_uid_if_none decl_shape type_declaration.type_uid
+
+  let of_type_declarations
+      (type_declarations : (Ident.t * Types.type_declaration) list)
+      shape_for_constr =
+    let decl_lookup_map = Ident.Map.of_list type_declarations in
+    (* We unbind all declarations, to avoid accidental recursive cycles. *)
```
mshinwell, line 466: I didn't follow this comment, what is it referring to?

typing/type_shape.ml
```diff
@@ -81,85 +83,121 @@ module Type_shape = struct
...
+     types.  Also consider reverting to the original value kind depth limit
+     (although 2 seems low). *)
+  let rec of_type_expr_go ~visited ~depth (expr : Types.type_expr)
+      (subst : (Types.type_expr * Shape.t) list) shape_for_constr : Shape.t =
     let open Shape in
+    let unknown_shape = Shape.leaf' None in
+    (* Leaves indicate we do not know. *)
     let[@inline] cannot_proceed () =
-      Numbers.Int.Set.mem (Types.get_id expr) visited || depth > 10
+      Numbers.Int.Map.mem (Types.get_id expr) visited || depth > 10
```
mshinwell, line 133: This parameter should have a command-line flag associated with it, I think

typing/type_shape.ml
```diff
@@ -81,85 +83,121 @@ module Type_shape = struct
...
-              | Types.Rabsent -> [] (* we filter out absent constructors *)
-              | Types.Reither (_, args, _) ->
-                [{ pv_constr_name = name; pv_constr_args = of_expr_list args }])
-            row_fields
+      match
+        List.find_opt (fun (p, _) -> Types.get_id p == Types.get_id expr) subst
+      with
+      (* CR sspies: Physical equality is also how printing in [printtyp.ml]
+         works. It seems to be the way to substitute type parameters (after
+         type inference has already made them more precise). *)
```
mshinwell: Please ask a frontend dev about this, maybe @ccasin

backend/debug/dwarf/dwarf_ocaml/dwarf_type.ml
spiessimon: I left a comment here for now. The command line flag will be added in #4535.

backend/debug/dwarf/dwarf_ocaml/dwarf_type.ml
spiessimon: I think this is the same one as the one in 1608 (i.e., it's a counter counting down and we should have a flag for the initial value).

backend/debug/dwarf/dwarf_ocaml/dwarf_type.ml
spiessimon: I think this is for the table. I've increased the initial size to 264.

backend/debug/dwarf/dwarf_ocaml/dwarf_type.ml
spiessimon: Changed it to fail loudly when `Cms_format.read` raises the error it raises for a magic number mismatch. In all other cases, we fail silently.

backend/debug/dwarf/dwarf_ocaml/dwarf_type.ml
spiessimon: Done.

typing/type_shape.ml
spiessimon: The function `shape_for_constr'` deletes all bindings for identifiers that will be bound in the current mutually recursive declaration block. I added it a while ago as a precaution when I experimented with a recursive version of `shape_for_constr_with_declarations`. The current version is much simpler and should not lead to accidental recursion. In fact, I believe all identifiers in the block should be unbound at this point, so at the moment `shape_for_constr'` is largely a no-op as far as correctness is concerned. It is not a no-op in case declarations can use the same identifier multiple times.

backend/debug/dwarf/dwarf_ocaml/dwarf_type.ml
spiessimon: Seems like Luke stumbled over the same issue at some point: https://github.com/oxcaml/oxcaml/blob/ad36b5e490554fe5195ea0fbc7d84d36c93457ef/typing/shape.mli#L353. I would tackle this in another PR, since this conceivably has some effect on Merlin (although technically I think Merlin currently always uses the version that never looks up `.cms` files in the first place).

typing/type_shape.ml
spiessimon: Turned it into a flag in #4535 and left a comment here.

typing/type_shape.ml
spiessimon: I checked with @goldfirere. It seems physical equality on type expressions is the most canonical form to test this. I removed the CR and changed it to physical equality directly on the types (instead of the IDs) in 04c8f95.

lpw25: I think it needs to be `Types.eq_type` rather than physical equality, which is equivalent to `(repr t1) == (repr t2)` which is the correct way to test this. Physical equality of `type_expr` on its own doesn't ensure that `repr` has been called. 

Often in the type-checker you know that `repr` has been called already so frequently you do just use physical equality, but it's not obvious that you know that here and even if you do its easier to reason about if you just call `eq_type`.

spiessimon: Oh, I didn't know that function exists. I've changed it to `Types.eq_type` in 6c5f8d4.

.vscode/settings.json
```diff
@@ -1,7 +1,7 @@
 {
         "ocaml.sandbox": {
                 "kind": "opam",
-                "switch": "4.14.2"
+                "switch": "4.14.1"
```
mshinwell: Looks like a spurious change

oxcaml/tests/backend/oxcaml_dwarf/test_datatypes_dwarf.output
```diff
@@ -36,7 +36,7 @@
     frame #N: <ADDRESS> test_datatypes_dwarf.exe`Test_datatypes_dwarf.f_open_poly_variant(x=`Alpha : [> `Alpha | `Beta of int ] @ value)
     frame #N: <ADDRESS> test_datatypes_dwarf.exe`Test_datatypes_dwarf.f_open_poly_variant(x=(`Beta 123) : [> `Alpha | `Beta of int ] @ value)
     frame #N: <ADDRESS> test_datatypes_dwarf.exe`Test_datatypes_dwarf.f_nested_variant(x=(Leaf 42) : nested_variant @ value)
-    frame #N: <ADDRESS> test_datatypes_dwarf.exe`Test_datatypes_dwarf.f_nested_variant(x=(Node { left = (Leaf 1); right = (Leaf 2); value = 3.14 }) : nested_variant @ value)
+    frame #N: <ADDRESS> test_datatypes_dwarf.exe`Test_datatypes_dwarf.f_nested_variant(x=(Node { left = [1]; right = [2]; value = 3.14 }) : nested_variant @ value)
```
mshinwell, line 39: Why has this regressed?

.vscode/settings.json
spiessimon: This was added in d36e7564a24c1fe175d57fbdd048f1a3c89db0d5. I was wondering whether that was intentional or not. I will revert it.

typing/type_shape.mli
```diff
@@ -3,25 +3,25 @@ module Layout = Jkind_types.Sort.Const
 
 type base_layout = Jkind_types.Sort.base
```
mshinwell, line 33: This file and its implementation are missing the MIT licence header.  See dwarf_inlined_frames.ml for an example.

oxcaml/tests/backend/oxcaml_dwarf/test_datatypes_dwarf.output
spiessimon: This has regressed, because recursive types are not handled in this PR. They are handled in #4492. Specifically, this line [is coming back here](https://github.com/oxcaml/oxcaml/pull/4492/commits/ce8f0372bfc752cc20ac880016abe93d46627ba2#diff-c01fed6fdf982a53f23438247e4d2b1919dcd3febb9cf7cf349c86d432521d5a).

typing/type_shape.mli
spiessimon: Added.

backend/debug/dwarf/dwarf_ocaml/dwarf_type.ml
mshinwell: Ah actually I meant that we shouldn't decrement the counter unless the read succeeds, but maybe it's fine as-is, too.

backend/debug/dwarf/dwarf_ocaml/dwarf_type.ml
mshinwell: Added a CR

backend/debug/dwarf/dwarf_ocaml/dwarf_type.ml
spiessimon: Done in 6f865d0.

